PR #10495, opened 2022-05-28
Allow 256 colors for terminal vim on Windows

### LINKED ISSUE #9498: [Win32] t_Co=16 until termguicolors, then t_Co=256
### Steps to reproduce

In `cmd`, `powershell`, `pwsh`, or `wt` on Windows run `vim --clean -c "echom 'Before tgc: '.&t_Co | set termguicolors | echom 'With tgc: '.&t_Co | set notermguicolors | echom 'After tgc: '.&t_Co"`

`:messages` contains:

    Before tgc: 16
    With tgc: 256
    After tgc: 256

### Expected behaviour

`:messages` contains:

    Before tgc: 256
    With tgc: 256
    After tgc: 256

### Version of Vim

8.2.4014

### Environment

OS: Windows 10 Version 10.0.19043.1415
Terminal: ConHost (`cmd`, `powershell`, `pwsh`) or Windows Terminal
`$TERM`: not set

### Logs and stack traces

_No response_

chrisbra:
I am not sure if the following makes sense, so feel free to discard if not wanted.

Enable 256 colors when run on recent enough windows in terminal mode.
Since there is a glitch with default 256 colors, also define Normal highlighting color as a black one.

Also fixes vim/vim#9498

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -1570,7 +1570,7 @@ parse_builtin_tcap(char_u *term)
 		else
 		{
 		    term_strings[p->bt_entry] = (char_u *)p->bt_string;
-#ifdef MSWIN
+#ifdef FEAT_VTP
 		    // Console has 256 Colors since Windows 1703
 		    if (is_term_win32() && has_vtp_working() && p->bt_entry == (int)KS_CCO)
 			term_strings[p->bt_entry] = (char_u *)"256";
--- a/src/term.c
+++ b/src/term.c
@@ -1570,6 +1570,11 @@ parse_builtin_tcap(char_u *term)
 		else
 		{
 		    term_strings[p->bt_entry] = (char_u *)p->bt_string;
+#ifdef MSWIN
+		    // Console has 256 Colors since Windows 1703
+		    if (is_term_win32() && has_vtp_working() && p->bt_entry == (int)KS_CCO)
+			term_strings[p->bt_entry] = (char_u *)"256";
+#endif
 #ifdef FEAT_EVAL
 		    opt_idx = get_term_opt_idx(&term_strings[p->bt_entry]);
 #endif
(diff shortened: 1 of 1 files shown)


LemonBoy:
What is exactly the problem? A description and some steps to reproduce it would be great.

chrisbra:
hm, i thought this is clear from the linked issues. Basically, this will make vim use 256 colors  in terminal mode on modern Windows (e.g. on cmd.exe)

brammool:
What is the change at the end of termcapinit() needed for?  A comment would help.

LemonBoy:
> hm, i thought this is clear from the linked issues. Basically, this will make vim use 256 colors in terminal mode on modern Windows (e.g. on cmd.exe)

I was referring to this bit
> Since there is a glitch with default 256 colors, also define Normal highlighting color as a black one.

What's the problem? The code in `termcapinit` looks like a workaround for some sort of problem as calling the highlight stuff is definitely out of place in that function.

From what I can see the virtual terminal processing is unconditionally enabled at startup, we could add a new builtin terminfo entry called `win32-vt` with all the VT escape sequences and `t_Co` set to 256 and use that whenever `vtp` is detected at startup (similarly to what already happens for eg. 256color terminals on Linux). The old `win32` using the legacy Win32 API can be then left alone and a lot of code can be simplified (eg. getting rid of all the tcap swapping machinery).

brammool:
> > hm, i thought this is clear from the linked issues. Basically, this
> > will make vim use 256 colors in terminal mode on modern Windows
> > (e.g. on cmd.exe)
> 
> I was referring to this bit
> > Since there is a glitch with default 256 colors, also define Normal
> > highlighting color as a black one.
> 
> What's the problem? The code in `termcapinit` looks like a workaround
> for some sort of problem as calling the highlight stuff is definitely
> out of place in that function.
> 
> From what I can see the virtual terminal processing is unconditionally
> enabled at startup, we could add a new builtin terminfo entry called
> `win32-vt` with all the VT escape sequences and `t_Co` set to 256 and
> use that whenever `vtp` is detected at startup (similarly to what
> already happens for eg. 256color terminals on Linux). The old `win32`
> using the legacy Win32 API can be then left alone and a lot of code
> can be simplified (eg. getting rid of all the tcap swapping
> machinery).

This solution possibly also helps with colorschemes adjusting to the
terminal.  We should at least give this a try and evaluate the resulting
situation.

-- 
"A mouse can be just as dangerous as a bullet or a bomb."
             (US Representative Lamar Smith, R-Texas)

chrisbra:
I'll rework it completly. But I'll need some time

brammool:
Still reworking this?

chrisbra:
yes, did not have much time the past week, since I was very busy with my job

Issue #10411, opened 2022-05-12
feature request: leading "multispace"-like option


bugsbugsbux:
hi, the new listchars option 'multispace' is really cool and i use it to show indent levels of space-indented lines, however, it also marks multiple spaces after the first visible character. this may be what the feature was intended to do (#8834), but for my usecase it is slightly annoying.

it would be great to have a similar feature for just the leading spaces. the 'lead' option could take multi-character strings like 'multispace' does, which wouldn't break peoples configs because setting a single-character string would result in the same behaviour as the current 'lead' behaviour (if this is triggered by leading spaces instead of multiple ones).

regards

zeertzjq:
Related comment: <https://github.com/vim/vim/pull/7772#issuecomment-772561146>
> For an indentation guide another field would be more appropriate. You might also want to show it when not showing leading space.

zeertzjq:
This should also be closed by patch 8.2.5066

chrisbra:
closing then. Thanks!

marcusatiliusregulus:
yay! thanks so much!

PR #10496, opened 2022-05-28
listchars: allows to specify leading multiple spaces

chrisbra:
As requested in vim/vim#10411 add a new suboption for the 'listchars'
setting, that allows to specify multiple leading spaces. This basically
works like 'multispace' suboption, but only for leading whitespace, e.g.
before any non-whitespace characters.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -2126,16 +2126,11 @@ win_line(
 			if (wp->w_lcs_chars.leadmultispace[multispace_pos] == NUL)
 			    multispace_pos = 0;
 		    }
-#if 0
-		    else
-			c = (ptr > line + trailcol) ? wp->w_lcs_chars.trail
-			     : (wp->w_lcs_chars.lead ? wp->w_lcs_chars.lead : ' ');
-#endif
 
 		    else if (ptr > line + trailcol && wp->w_lcs_chars.trail)
 			c = wp->w_lcs_chars.trail;
 
-		    else if (ptr < line + trailcol && wp->w_lcs_chars.lead)
+		    else if (ptr < line + leadcol && wp->w_lcs_chars.lead)
 			c = wp->w_lcs_chars.lead;
 
 		    else if (leadcol != 0 && c == ' ' && wp->w_lcs_chars.space)
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -767,6 +767,7 @@ win_line(
     {
 	if (wp->w_lcs_chars.space
 		|| wp->w_lcs_chars.multispace != NULL
+		|| wp->w_lcs_chars.leadmultispace != NULL
 		|| wp->w_lcs_chars.trail
 		|| wp->w_lcs_chars.lead
 		|| wp->w_lcs_chars.nbsp)
@@ -781,7 +782,7 @@ win_line(
 	    trailcol += (colnr_T) (ptr - line);
 	}
 	// find end of leading whitespace
-	if (wp->w_lcs_chars.lead)
+	if (wp->w_lcs_chars.lead || wp->w_lcs_chars.leadmultispace != NULL)
 	{
 	    leadcol = 0;
 	    while (VIM_ISWHITE(ptr[leadcol]))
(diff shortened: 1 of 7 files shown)


src/screen.c
```diff
@@ -4835,11 +4835,13 @@ get_encoded_char_adv(char_u **p)
     char *
 set_chars_option(win_T *wp, char_u **varp)
 {
-    int		round, i, len, entries;
+    int		round, i, len, len2, entries;
     char_u	*p, *s;
     int		c1 = 0, c2 = 0, c3 = 0;
     char_u	*last_multispace = NULL; // Last occurrence of "multispace:"
+    char_u	*last_lmultispace = NULL; // Last occurrence of "lead-multispace:"
```
zeertzjq:
```suggestion
    char_u	*last_lmultispace = NULL; // Last occurrence of "leadmultispace:"
```

src/screen.c
```diff
@@ -4835,11 +4835,13 @@ get_encoded_char_adv(char_u **p)
...
     int		multispace_len = 0;	 // Length of lcs-multispace string
+    int		lead_multispace_len = 0; // Length of lcs-lead-multispace string
```
zeertzjq:
```suggestion
    int		lead_multispace_len = 0; // Length of lcs-leadmultispace string
```

src/screen.c
```diff
@@ -5012,6 +5023,43 @@ set_chars_option(win_T *wp, char_u **varp)
...
+
+		else if ((varp == &p_lcs || varp == &wp->w_p_lcs)
+			&& STRNCMP(p, "leadmultispace", len2) == 0
+			&& p[len2] == ':'
+			&& p[len2 + 1] != NUL)
+		{
+		    s = p + len2 + 1;
+		    if (round == 0)
+		    {
+			// Get length of lcs-multispace string in first round
```
zeertzjq:
```suggestion
			// Get length of lcs-leadmultispace string in first round
```

src/screen.c
```diff
@@ -5012,6 +5023,43 @@ set_chars_option(win_T *wp, char_u **varp)
...
+			lead_multispace_len = 0;
+			while (*s != NUL && *s != ',')
+			{
+			    c1 = get_encoded_char_adv(&s);
+			    if (char2cells(c1) > 1)
+				return e_invalid_argument;
+			    ++lead_multispace_len;
+			}
+			if (lead_multispace_len == 0)
+			    // lcs-multispace cannot be an empty string
```
zeertzjq:
```suggestion
			    // lcs-leadmultispace cannot be an empty string
```

src/drawline.c
```diff
@@ -2118,8 +2119,29 @@ win_line(
...
+#if 0
+		    else
+			c = (ptr > line + trailcol) ? wp->w_lcs_chars.trail
+			     : (wp->w_lcs_chars.lead ? wp->w_lcs_chars.lead : ' ');
+#endif
+
+		    else if (ptr > line + trailcol && wp->w_lcs_chars.trail)
+			c = wp->w_lcs_chars.trail;
+
+		    else if (ptr < line + trailcol && wp->w_lcs_chars.lead)
```
zeertzjq:
```suggestion
		    else if (ptr < line + leadcol && wp->w_lcs_chars.lead)
```

brammool:
In the test, instead of using "ABCD" using something easier to recognize makes it easier to see what happens.  Something like ".-+*".

src/screen.c
chrisbra: fixed

src/screen.c
chrisbra: fixed

src/screen.c
chrisbra: fixed

src/screen.c
chrisbra: fixed

src/drawline.c
chrisbra: fixed

chrisbra:
> In the test, instead of using "ABCD" using something easier to recognize makes it easier to see what happens. Something like ".-+*".

Changed the test

lacygoill:
For those of you following the PR, it was merged in patch [8.2.5066](https://github.com/vim/vim/releases/tag/v8.2.5066).  The latter has a misleading message, which is a duplicate of [8.2.5067](https://github.com/vim/vim/releases/tag/v8.2.5067).  Thank you @chrisbra for the patch.

PR #10504, opened 2022-05-31
syntax/rust: add "dyn" keyword

Soveu:
I noticed `dyn` doesn't get highlighted, it was kinda surprising, considering vim has syntax highlighting even for non-official keywords (like `box`).

It is placed in the same group as `mut` so both words in `&mut dyn` are identically highlighted.


brammool:
The file header suggests to request changes here: https://github.com/rust-lang/rust.vim

Soveu:
Well, since using rust.vim is recommended instead, I will close this PR.

PR #10494, opened 2022-05-28
Use a more accurate clock source when possible

LemonBoy:
Kill every single (mis)use of gettimeofday() and replace them with a
faster non-wall-clock timer.

cc @paul-ollis

PR #10505, opened 2022-05-31
Timeout

paul-ollis:
FYI: Use of OS timers to timeout syntax highlighting and searching.

Removes the need to regularly read any system timers while executing the RE engine.

### DIFF
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -8276,7 +8276,7 @@ static int timer_created = FALSE;
  * Callback for when the timer expires.
  */
     static void
-set_flag(union sigval _unused)
+set_flag(union sigval _unused UNUSED)
 {
     timeout_flag = TRUE;
 }
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -24,6 +24,12 @@
 # include "if_mzsch.h"
 #endif
 
+#ifdef FEAT_RELTIME
+#include <errno.h>
+#include <signal.h>
+#include <sys/time.h>
+#endif
+
 #include "os_unixx.h"	    // unix includes for os_unix.c only
 
 #ifdef USE_XSMP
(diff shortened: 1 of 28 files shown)


src/os_unix.c
```diff
@@ -8256,3 +8262,129 @@ xsmp_close(void)
     }
 }
 #endif // USE_XSMP
+
+#ifdef FEAT_RELTIME
+static struct itimerval prev_interval;
+static struct sigaction prev_sigaction;
+static int timeout_flag         = FALSE;
```
puremourning, line 8352: is it risky that this is global? i.e. if start_timeout is called-reentrantly? 

is it possible for the regex code to be re-entrant? A lot of vim code has to handle this sort of thing.

puremourning:
> `getitimer` and `setitimer` are deprecated in favor of `timer_gettime` and `timer_settime` which, in the spirit of #10494, allow the coder to select a monotonic clock source.

not on macos FYI - would require feature test probably.

paul-ollis:
@LemonBoy wrote:

> getitimer and setitimer are deprecated in favor of timer_gettime and timer_settime which, in the spirit of https://github.com/vim/vim/pull/10494, allow the coder to select a monotonic clock source.

I know. And I would like use them in preference because they have other advantages as well. I already have implementation (on a separate branch on my PC); it works well and the code is a bit simpler.

I switched to setitimer on this branch for pragmatic reasons; namely the code ports to OS X and so stood a chance of actually passing macos CI checks. I am planning to revive oldish Mac laptop (dead hard drive) so I am in a position to develop an OS X specific solution.

brammool:
Thanks for working on this alternative solution.  The idea looks good, we need to check that this works on all systems where SYN_TIME_LIMIT is defined.

The alarm signal is set for redrawing a window.  That matches with what 'redrawtime' is for.  The functions should be called to reflect that:  init_redraw_timeout() instead of init_regexp_timeout().

I'm not sure if it ever gets used recursively.  For a start I would produce an internal error when the alarm is tried to set when it already is.  If we find a case where this happens we'll have to see how we need to handle it.
I suspect it might happen when a timer or channel callback is invoked.  We might then have to stop the running alarm, set one for whatever happens in the callback, then restore the previous one for the remaining time.  That's a bit complicated, thus we should figure out if it is really needed.

LemonBoy:
> I switched to setitimer on this branch for pragmatic reasons; namely the code ports to OS X and so stood a chance of actually passing macos CI checks. I am planning to revive oldish Mac laptop (dead hard drive) so I am in a position to develop an OS X specific solution.

On macOS you can emulate the `timer_` API using the GCD, [here's](https://gist.github.com/lundman/731d0d7d09eca072cd1224adb00d9b9e) a small example of how to do so. According to the API [docs](https://developer.apple.com/documentation/dispatch/1385606-dispatch_source_set_timer) there's a way to avoid using the wall clock timer.

LemonBoy:
One more thing to worry about is the flag atomicity given it's written from within a signal handler, check out [this article](https://wiki.sei.cmu.edu/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers) for more info.

paul-ollis:
> One more thing to worry about is the flag atomicity given it's written from within a signal handler, check out [this article](https://wiki.sei.cmu.edu/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers) for more info.

Definitely not an issue for this code. Possibly should use one of more 'volatile' declarations though for correctness.

brammool:
> > The alarm signal is set for redrawing a window. That matches with
> > what 'redrawtime' is for. The functions should be called to reflect
> > that: init_redraw_timeout() instead of init_regexp_timeout().
> 
> Actually the timeout logic (existing and on this branch) is also used
> for the vim search() function; I added a test for it on this branch.

Ah, OK.  Also used by searchpair().

-- 
The question is:  What do you do with your life?
The wrong answer is: Become the richest guy in the graveyard.
				(billionaire and Oracle founder Larry Ellison)

src/os_unix.c
paul-ollis: A valid question.

I did review the code to check for such issues and found none. Not really surprising since this is replacing one, single global timeout mechanism for another.

brammool:
Starting to look much better, only one relevant test failure on Windows:
Found errors in Test_search_timeout():

paul-ollis:
@brammool wrote.

> Starting to look much better, only one relevant test failure on Windows:
> Found errors in Test_search_timeout():

I think I may have reached a point where the tests should be stable. I now plan to switch to timer_settime for Unix and some equivalent for OS X (now I have an OS X platform).

Is there any way to manually trigger a CI run (or better still selective parts of a CI run)? It is the only way I can get some evidence of stability of the timeout tests, since I could not reproduce the occasional failures on my computers.

brammool:
All tests are passing now.  CI is only triggered if you push a change.
Let me know when it's ready to include, then we can give it to a wider audience to see if there are problems on some system.

paul-ollis:
@brammool I think this is ready for a wider audience now.

Is there some definitive list of the platforms that we support?
Even better, is there some matrix or platforms and supported features?
I am now set up to test on "the big 3", but would be interested to expand to other platforms where possible.

brammool:
I have included this now.  I made some cosmetic adjustments.

Please check, if someone has a problem with it we may fix that, or roll back this change.

PR #10507, opened 2022-06-01
Add additional autocmd tests

yegappan:


brammool:
Not sure why you remove that check for NULL, theoretically it can happen if the string value is NULL.

PR #10509, opened 2022-06-01
add name for checkout step

pheiduck:
Add name for actions/checkout@v3
Problem: CI step actions/checkout@v3 does not look nice
Solution: make it look nicer


brammool:
How do I see the difference in the output?

pheiduck:
withou_my_changes

with_my_changes

PR #10511, opened 2022-06-02
Add `breakundo()` function (like `i_CTRL-G_u`)

andmis:


src/evalfunc.c
```diff
@@ -3185,6 +3188,15 @@ get_buf_arg(typval_T *arg)
...
+/*
+ * "breakundo()" function
+ */
+    static void
+f_breakundo(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
+{
+    u_sync(1);
+}
+
 /*
```
yegappan, line 3200: Can you add a test for this new function?

brammool:
Do we really need a function for this?  You can do ```&undolevels = &undolevels```.

andmis:
I think having a function makes intent more explicit. Its presence in the lists of builtin functions will make a higher percentage of plugin authors aware that (1) this is something they should be mindful of, if their plugins modify the buffer, and (2) that there is this simple and explicit way of breaking the undo sequence from a script context.

src/evalfunc.c
andmis: Done

brammool:
I doubt adding a function does much to visibility.  One still needs to search the help to find this.
Also, I don't think it is needed very often.  Only in few cases does a script need to break an undo block in two.
For the intent, adding a comment is usually sufficient.  This is done in tests in various places.
I'll add an example in the help, to make it a bit easier to spot.

lacygoill:
> Do we really need a function for this? You can do &undolevels = &undolevels.

We can also force a redraw of the statusline by executing `:let &ro = &ro` as documented at [`:help 'statusline'`](https://vimhelp.org/options.txt.html#%27statusline%27).  That did not stop us from introducing `:redrawstatus`, which makes the code easier to understand.

brammool:
> > Do we really need a function for this? You can do &undolevels = &undolevels.
> 
> We can also force a redraw of the statusline by executing `:let &ro =
> &ro` as documented at [`:help
> 'statusline'`](https://vimhelp.org/options.txt.html#%27statusline%27).
> That did not stop us from introducing `:redrawstatus`, which makes the
> code easier to understand.

There are several plugins that add various things to the statusline,
thus this is much more common.
Also, setting 'readonly' to get the statusline to update is really
indirect.  Without a comment almost nobody would guess why that is done.

Anyway, thanks for pointing out that outdated remark, it should mention
using :redrawstatus instead.

-- 
A poem:                read aloud:

<> !*''#               Waka waka bang splat tick tick hash,
^"`$$-                 Caret quote back-tick dollar dollar dash,
!*=@$_                 Bang splat equal at dollar under-score,
%*<> ~#4               Percent splat waka waka tilde number four,
&[]../                 Ampersand bracket bracket dot dot slash,
|{,,SYSTEM HALTED      Vertical-bar curly-bracket comma comma CRASH.

Fred Bremmer and Steve Kroese (Calvin College & Seminary of Grand Rapids, MI.)

lacygoill:
> There are several plugins that add various things to the statusline,
   > thus this is much more common.

That's true.  I've just never liked writing/reading `let &undolevels = &undolevels`.  We need to add a comment, and comments can lie (after a refactoring, or just moving some lines).  And even if the comment does not lie, we still have to read one more line.  I would prefer reading this:

    breakundo()

Than this:

    # break undo sequence
    &undolevels = &undolevels

Small changes like this might look minor, but I think they can add up.  At some point, they can make a big difference in how easy a script is to read.

andmis:
- Probably most plugins that modify the buffer should be breaking the undo sequence more often than they are -- for example, anytime they modify the buffer in response to a user Insert mode mapping.
- I think adding a function would certainly increase visibility -- it's very natural to search `builtins.txt` for the string "undo" when you want to break the undo sequence from a script, which is exactly what I did.
- Even many expert Vim users are unaware that it is possible to break the undo sequence from a script at all.
- +1 to @lacygoill 's example above.
- Does adding this function add any maintenance overhead? "Breaking the undo sequence" is functionality that will need to be supported forever anyway and the Vim script function is just a thin wrapper around the C function.

I hope you'll reconsider...

lacygoill:
> I think adding a function would certainly increase visibility -- it's very natural to search builtins.txt for the string "undo" when you want to break the undo sequence from a script, which is exactly what I did.

We don't even need to look for undo in the right file.  Write this on the command-line:

    :h *keyword*(

Then, instead of pressing Enter to execute the command, press CTRL-D.

Right now, if we replace "keyword" with "undo", we get:

    undofile()  undotree()

With `breakundo()` (or `undobreak()`?), we would get:

    breakundo()  undofile()  undotree()

---

> Shouldn't the function be called undobreak() to be more consistent with the already existing command :undojoin?

Yes, a command would be more consistent.  Although, nowadays, I tend to prefer functions over commands.

src/evalfunc.c
```diff
@@ -3185,6 +3188,15 @@ get_buf_arg(typval_T *arg)
     return buf;
 }
 
+/*
+ * "breakundo()" function
+ */
+    static void
+f_breakundo(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
+{
+    u_sync(1);
```
zeertzjq:
```suggestion
    u_sync(TRUE);
```

PR #10517, opened 2022-06-03
Inserting space in the message

charlesduarte019:
Alert message is showing username concatenating with date message. To make the correction, I put 3 space in front of the message.

image


brammool:
Thanks, I'll just include it.

PR #10213, opened 2022-04-17
Fix file type of pacman.conf and pacman hook files.

chaoren:
They're just generic configuration files with # comments.

pacman.conf example:
https://archlinux.org/pacman/pacman.conf.5.html#_example

hook file example:
https://man.archlinux.org/man/alpm-hooks.5#EXAMPLES

dosini has ; comments, so that's not correct.

PR #10518, opened 2022-06-03
Create new filetype for ini style config files

Mudskipper875:
Several programs have config files with a syntax similar to dosini.
The only difference is that they only have '#' comments and don't
support semicolon comments.

 This is the config file format for such programs
 ```ini
 [section]
 option = value
 option2=value2

 # Comment
 ```

This filetype can temporarily source `cfg.vim` until a proper filetype
is created.

This pull request is based on the discussion in #10213


brammool:
Included with Patch 8.2.5054 (which referred to the wrong issue).

Issue #10231, opened 2022-04-20
i3config syntax isn't quite compatible with sway


rpigott:
The new i3config syntax isn't quite compatible with sway's config syntax.

Although we try to retain compatibility with i3, there are some cases that compel us to introduce new keywords and features in sway's configuration, mostly to replace features that used to be part of X configuration that must now be handled in the compositor.

So the highlighter doesn't understand `bindswitch` or `input`/`output` sway config directives, for example. It also doesn't understand the `include` directive, even though i3 has since added support for this originally sway-specific feature in i3/i3#4420. The highlight path */.sway/config also is not appropriate for sway since sway does not check that path for its configuration (https://man.archlinux.org/man/community/sway/sway.1.en#CONFIGURATION).

Do you think we should try to keep an i3 and sway compatible syntax definition or make two separate syntaxes? I really wouldn't mind if not for the obnoxious error highlighting, so that's what I'm trying to correct. If preferable, maybe we can just be less strict about highlighting errors.

dkearns:
I'm not really familiar with either but in my experience it's almost always easier to create a new filetype.  It sounds like you could just source the i3 syntax file and make additions.

brammool:
If two filetypes are similar but not the same, then it's usually easier to have two filetypes, and have the syntax perhaps share the common part.  If one is a superset of the other, then include that other and add on top of it.

hiqua:
@RPigott:
1. please see the [PR](https://github.com/vim/vim/pull/7969) for the context, basically it's a best-effort thing, with the hope that vim is better with than without these syntax files
2. given this, I should probably remove the error highlighting stuff given that it's bound to break too often to be a real gain. This should solve the main issues with sway
3. until somebody chimes in to support sway config properly in vim, I think it's fine to have some minimal support in the same files
4. `The highlight path */.sway/config also is not appropriate for sway since sway does not check that path for its configuration (https://man.archlinux.org/man/community/sway/sway.1.en#CONFIGURATION).`: I think you meant that sway doesn't check every path ending with `.sway/config`, but I think this is fine given the rest of the filetype.vim file. It's unlikely to have a path ending with this which wouldn't be a sway config file. I think the other paths mentioned in your link are already covered, so there shouldn't be any change required for this, but let me know if I've missed something.

I'll create a PR for 2. Somebody already added the `include` directive. If somebody wants to create a PR for missing sway directives feel free to do so!

habamax:
[...]

PR #10522, opened 2022-06-04
Do not highlight errors in i3 config files

hiqua:
See #10231.


brammool:
Thanks.  I'll include it and also update the "Last Change" date in the header.

PR #10524, opened 2022-06-04
MANPAGER: make buffer nomodifiable and hide line numbers

rvighne:
The current version has a bug where it doesn't matter whether you invoke `vim -M` because it does `setl modifiable`. This fixes that, and makes a bunch of other options the same as what you get if you do `:Man`; most notably, disabling line numbering so that the preformatted text from `man` doesn't wrap at every line.

Also some general cleanup for clarity and updating docs to suggest `--not-a-term` so that the "Reading from stdin..." message is hidden, which is more in line with plain `man`.

### DIFF
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -642,15 +642,15 @@ manpages and follows linked manpages on hitting CTRL-]).
 
 For bash,zsh,ksh or dash, add to the config file (.bashrc,.zshrc, ...)
 
-	export MANPAGER="vim -M +MANPAGER -"
+	export MANPAGER="vim +MANPAGER --not-a-term -"
 
 For (t)csh, add to the config file
 
-	setenv MANPAGER "vim -M +MANPAGER -"
+	setenv MANPAGER "vim +MANPAGER --not-a-term -"
 
 For fish, add to the config file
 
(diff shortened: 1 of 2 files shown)


chrisbra:
have you contacted the maintainer?

rvighne:
Contacted the maintainer, waiting to hear back. Also it looks like someone incorporated some of these changes into master already, so after resolving conflicts this diff has reduced in scope a bit (setting nomodifiable in particular). Still, this change makes it behave more similar to the `:Man` command, e.g. removing line numbers, setting buffer type, setting fewer global options.

rvighne:
Haven't heard from the maintainer yet. Is there anyone else who can review this change? What usually happens in cases like this? If needed, I can become the maintainer of this file.

chrisbra:
Hm, not 100% sure, but I believe: ping @Konfekt

Konfekt:
Thank you @rvighne ! Looks good to me. Regarding `setlocal iskeyword+=:` it was probably added to jump to manpages containing colons by `K`;  I'm unsure about the benefit in removing this.

brammool:
Thanks, I'll include it.  Let's keep the 'iskeyword' change, I haven't heard complaints about it.

PR #10526, opened 2022-06-05
Prevent block detection after a comment

### LINKED ISSUE #10508: Comments highlight doesn't work properly if there's no space after the leading #
https://github.com/vim/vim/blob/b5912e0643dc21b1b88c517d3ac1282eba57293d/runtime/syntax/i3config.vim#L27

i3config

hiqua:
Fixes #10508.

@Aster89 can you play with this a bit and check if things still break somehow?


Aster89:
Kind of works, but it breaks elsewhere. Here's a minimal example:
```
bar {
    colors {
    }
}
```
the last `}` is marked as `i3ConfigError`.

hiqua:
`i3ConfigError` will be removed (already created a PR, should be merged soon) so that should be fine.

brammool:
Let me include this.  If there are remaining problems please open a new issue.

PR #10529, opened 2022-06-06
Update ex_cmds.c

H4niz:
Update length check before doing `mch_memmove()` in `ex_substitute()` function at ex_cmds.c to prevent heap-based buffer overflow.

### DIFF
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -4480,8 +4480,18 @@ ex_substitute(exarg_T *eap)
 		    nmatch_tl += nmatch - 1;
 		}
 		copy_len = regmatch.startpos[0].col - copycol;
-		needed_len = copy_len + ((unsigned)STRLEN(p1)
+
+		/*
+		 * Make sure needed_len always greater than copy_len
+		 */
+		if((unsigned)STRLEN(p1) == 0)	{
+			needed_len = copy_len + sublen + 1;
+		}		
+		else {
+			needed_len = copy_len + ((unsigned)STRLEN(p1)
 				       - regmatch.endpos[0].col) + sublen + 1;
+		}
+
 		if (new_start == NULL)
 		{
 		    /*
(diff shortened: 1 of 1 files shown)


src/ex_cmds.c
```diff
@@ -4480,8 +4480,18 @@ ex_substitute(exarg_T *eap)
 		    nmatch_tl += nmatch - 1;
 		}
 		copy_len = regmatch.startpos[0].col - copycol;
-		needed_len = copy_len + ((unsigned)STRLEN(p1)
+
+		/*
+		 * Make sure needed_len always greater than copy_len
+		 */
+		if((unsigned)STRLEN(p1) == 0)	{
```
dpelle: STRLEN is O(n). To check whether it's an empty string, you can do it in O(1) with:
```
if (*p1 == NUL)
```
Although the compiler might optimize it anyway.

brammool:
How do you reproduce that heap-based buffer overflow?

H4niz:
> How do you reproduce that heap-based buffer overflow?

I reported this vulnerability on huntr.dev, POC as well. Please contact with them for more informations. It's maybe my mistake because of first time I report a vulneribility on huntr.dev, in order that, I commited both. Sorry about that! Let follow huntr.dev!

brammool:
Can you please turn the POC into a test case?  You can check previous fixes for substitute for examples.  The test would cause a valgrind error before the fix.

H4niz:
[Updated]
I could not run the testcase on v8.2.5062. The vulnerability was fixed!

Full git log below:
```
commit 1f89abf69d2c485c38da9a45161e806de0f86f2f (HEAD -> master, tag: v8.2.5062, origin/master, origin/HEAD)
Author: Bram Moolenaar <Bram@vim.org>
Date:   Mon Jun 6 10:07:01 2022 +0100

    patch 8.2.5062: Coverity warns for dead code
    
    Problem:    Coverity warns for dead code.
    Solution:   Remove the dead code.
```

chrisbra:
closing, since it is fixed.

PR #10538, opened 2022-06-08
Fix clang compiler warnings in the MS-Windows build

yegappan:


src/dosinst.c
```diff
@@ -212,7 +212,7 @@ check_unpack(void)
 	printf("ERROR: Cannot find filetype.vim in \"%s\"\n", installdir);
 	printf("It looks like you did not unpack the runtime archive.\n");
 	printf("You must unpack the runtime archive \"vim%srt.zip\" before installing.\n",
```
k-takata: If line 215 is changed from `VIM_VERSION_NODOT + 3` to `VIM_VERSION_NODOT`, then this line should be changed as follows:

```suggestion
	printf("You must unpack the runtime archive \"%srt.zip\" before installing.\n",
```

vim-ml:
Hi,

On Tue, Jun 7, 2022 at 7:52 PM K.Takata ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/dosinst.c
> <https://github.com/vim/vim/pull/10538#discussion_r891868223>:
>
> > @@ -212,7 +212,7 @@ check_unpack(void)
>
>  	printf("ERROR: Cannot find filetype.vim in \"%s\"\n", installdir);
>
>  	printf("It looks like you did not unpack the runtime archive.\n");
>
>  	printf("You must unpack the runtime archive \"vim%srt.zip\" before installing.\n",
>
>
> If line 215 is changed from VIM_VERSION_NODOT + 3 to VIM_VERSION_NODOT,
> then this line should be changed as follows:
>

VIM_VERSION_NODOT is defined as a string. The current code is trying to add
a number to the string.


> ⬇️ Suggested change
>
> -	printf("You must unpack the runtime archive \"vim%srt.zip\" before installing.\n",
>
> +	printf("You must unpack the runtime archive \"%srt.zip\" before installing.\n",
>
>
>
>
Updated the PR with this change.

- Yegappan

k-takata:
> VIM_VERSION_NODOT is defined as a string. The current code is trying to add
a number to the string.

Yes, it was intended. `VIM_VERSION_NODOT` is "vim82". Adding 3 to it means that the pointer points to the third character of the string. So, `VIM_VERSION_NODOT + 3` is the same as "82".

PR #10550, opened 2022-06-09
Update Turkish translations

bitigchi:


src/po/tr.po
```diff
@@ -21,15 +21,13 @@ msgstr ""
 msgid "ERROR: "
 msgstr "HATA: "
 
-#, c-format
```
k-takata, line 24: These `, c-format` comments should be kept.

bitigchi: If I'm not mistaken, they were automatically removed via the `cleanup.vim` script. Restored for now.

brammool:
@k-takata commented on this pull request.
> 
> > @@ -21,15 +21,13 @@ msgstr ""
>  msgid "ERROR: "
>  msgstr "HATA: "
>  
> -#, c-format
> 
> These `, c-format` comments should be kept.

Can you tell me why?  I was discussing this with the Italian translator,
and we thought it would be OK to remove.  I added a line in the cleanup
script for this.  If they should be kept, I need to update that script.

-- 
I wonder, do vegetarians eat fruit bats?

chrisbra:
I think they need to be there for xgettext to check the order of the `%` expandos at runtime. At least that's how I understood them. I usually have only removed them manually, if the `%` was clearly not meant as a `printf` formatting atom.

See also here: https://www.gnu.org/software/gettext/manual/html_node/c_002dformat-Flag.html

brammool:
I'll include the translations and apply the cleanup script, without removing c-format for now.

If we leave out those c-format comments, I don't think anything goes wrong.  It's mainly for the translator to avoid mistakes.  But we also have the check script do this, I recall this is needed because the tooling doesn't always catches the mistakes.

You could try to see what happens with the German translations, intentionally making a mistake in a printf string and see what happens (when using the translations, or when running msgfmt.

PR #10552, opened 2022-06-10
Reduce duplicate code

zeertzjq:


zeertzjq:
I'll also remove a condition that is always true.

PR #10563, opened 2022-06-12
Update colorschemes

habamax:
- Add missing 8c support
- Bring back bolds as in originals

@romainl, fyi


brammool:
I will include it and send out a runtime files update soon.  Thanks!

PR #10565, opened 2022-06-13
Update Perl runtime files

dkearns:
@petdance, can you run your eyes over this please.


petdance:
Notes:

* I think we should get the http://github.com/ URLs in the headers of the files we are submitting to be https:.  I see that some headers had https: and they went back to http:.  Is there a reason we'd want http:?

* I see some date formats changed from "2015-04-25" to "2018 Mar 28".  Does Bram prefer that?

dkearns:
I've updated the headers to use https: URLs.  I'll update them in the vim-perl repo too.

I don't think he particularly minds but that's the date format that Bram uses and as the date header isn't updated in the vim-perl repo we may as well use this format.  E.g., https://github.com/vim/vim/blob/db77c49401145d76441fbb3d22a1d7d987681c13/runtime/filetype.vim#L4

brammool:
The date format "2022 Mar 4" gets update automatically when I edit a file.  That usually is good, but sometimes it gets updated unintentionally, e.g. when fixing a typo.  You can complain if it's wrong...

I'll include the patch, thanks.

Issue #10530, opened 2022-06-06
test_terminal.vim Failed / ArchLinux PKGBUILD


pheiduck:
### Steps to reproduce

use Arch Linux PKGBUILD
build it with makepkg -src or chrootbuild

### Expected behaviour

no test failed like on upstream (here)

### Version of Vim

8.2.5062-1

### Environment

gh Action ubuntu-latest
manjaro-chrootbuild

### Logs and stack traces

```shell
Failures: 
	From test_terminal.vim:
	Found errors in Test_terminal_qall_prompt():
	Run 1:
	command line..script /build/vim/src/vim-8.2.5062/src/testdir/runtest.vim[459]..function RunTheTest[44]..Test_terminal_qall_prompt[6]..WaitForAssert[2]..<SNR>6_WaitForCommon[11]..<lambda>133 line 1: Pattern '\\[running]' does not match '<r@fv-az82-486:/build/vim/src/vim-8.2.5062/src/testdir]  0,0-1          All'
	Run 2:
	command line..script /build/vim/src/vim-8.2.5062/src/testdir/runtest.vim[492]..function RunTheTest[44]..Test_terminal_qall_prompt[6]..WaitForAssert[2]..<SNR>6_WaitForCommon[11]..<lambda>138 line 1: Pattern '\\[running]' does not match '<r@fv-az82-486:/build/vim/src/vim-8.2.5062/src/testdir]  0,0-1          All'
	Run 3:
	command line..script /build/vim/src/vim-8.2.5062/src/testdir/runtest.vim[492]..function RunTheTest[44]..Test_terminal_qall_prompt[6]..WaitForAssert[2]..<SNR>6_WaitForCommon[11]..<lambda>143 line 1: Pattern '\\[running]' does not match '<r@fv-az82-486:/build/vim/src/vim-8.2.5062/src/testdir]  0,0-1          All'
	Run 4:
	command line..script /build/vim/src/vim-8.2.5062/src/testdir/runtest.vim[492]..function RunTheTest[44]..Test_terminal_qall_prompt[6]..WaitForAssert[2]..<SNR>6_WaitForCommon[11]..<lambda>148 line 1: Pattern '\\[running]' does not match '<r@fv-az82-486:/build/vim/src/vim-8.2.5062/src/testdir]  0,0-1          All'
	Run 5:
[...]

PR #10575, opened 2022-06-14
:test_override instead of :call test_override

pheiduck:
found here: https://github.com/vim/vim/issues/10530
Fix error in 99f4b6e -- it does :test_override instead of :call test_override.


pheiduck:
already in 5096

PR #10586, opened 2022-06-17
Enable limit option on fuzzy matching a list of dictionaries

yuzu3886:
When passing a list of dictionaries to matchfuzzy(), limit option is ignored


mattn:
@yuzu3886 Thank you

PR #10592, opened 2022-06-18
Do not prepend a dot to files in ~/.vim as per FHS 3.0.

imageck:
Filesystem Hierarchy Standard Version 3.0 advises against this practice. 

From 3.8.2. Requirements

> User specific configuration files for applications are stored in the user's home directory in a file that starts with the '.' character (a "dot file"). If an application needs to create more than one dot file then they should be placed in a subdirectory with a name starting with a '.' character, (a "dot directory"). In this case the configuration files **should not** start with the '.' character.


brammool:
You cannot just change that, it causes lots of trouble.

Issue #10562, opened 2022-06-12
8.2.5057 allows searchpair() to run much longer than the timeout


tomtomjhj:
### Steps to reproduce

```
vim --clean
:e src/evalfunc.c
:so repro.vim
:call Slow()
:q
```

repro.vim
```vim
profile start profile.log
profile func *
profile file *

function! Skip()
    let id = synID(line('.'), col('.'), 0)
    let attr = synIDattr(id, 'name')
    return attr !~ 'comment'
endfunction
function! Slow()
    echo searchpair('(', '', ')', 'crnm', 'Skip()', 0, 1)
endfunction
```

result (profile.log): `searchpair()` runs much longer than the timeout (1ms)
```
FUNCTION  Slow()
    Defined: ~/.vim/repro.vim:10
Called 1 time
Total time:   0.505885
 Self time:   0.076826

count  total (s)   self (s)
    1   0.505883   0.076824     echo searchpair('(', '', ')', 'crnm', 'Skip()', 0, 1)

FUNCTION  Skip()
    Defined: ~/.vim/repro.vim:5
Called 7255 times
Total time:   0.429059
 Self time:   0.429059

count  total (s)   self (s)
 7255              0.397128     let id = synID(line('.'), col('.'), 0)
 7255              0.015111     let attr = synIDattr(id, 'name')
 7255              0.014219     return attr !~ 'comment'

FUNCTIONS SORTED ON TOTAL TIME
count  total (s)   self (s)  function
    1   0.505885   0.076826  Slow()
 7255   0.429059             Skip()

FUNCTIONS SORTED ON SELF TIME
count  total (s)   self (s)  function
 7255              0.429059  Skip()
    1   0.505885   0.076826  Slow()
```

### Expected behaviour

`searchpair()` returns after 1ms. (tested on 8.2.5056)
```
FUNCTION  Slow()
    Defined: ~/.vim/repro.vim:10
Called 1 time
Total time:   0.001057
 Self time:   0.000293

count  total (s)   self (s)
    1   0.001055   0.000291     echo searchpair('(', '', ')', 'crnm', 'Skip()', 0, 1)

FUNCTION  Skip()
    Defined: ~/.vim/repro.vim:5
Called 10 times
Total time:   0.000764
 Self time:   0.000764

count  total (s)   self (s)
   10              0.000677     let id = synID(line('.'), col('.'), 0)
   10              0.000042     let attr = synIDattr(id, 'name')
[...]

PR #10595, opened 2022-06-19
Issue 10562

paul-ollis:
A fix for #10562.

Not polished, I had in mind to add some more tests, but Bram asked to see it.

### DIFF
--- a/src/testdir/test_search.vim
+++ b/src/testdir/test_search.vim
@@ -1584,11 +1584,20 @@ endfunction
 
 func Test_searchpair_timeout()
   edit ../evalfunc.c
+  if has('win32')
+    " Windows timeouts are rather coarse grained, about 16ms.
+    let ms = 20
+    let min_time = 0.016
+  else
+    let ms = 1
+    let min_time = 0.001
+  endif
+  let max_time = min_time * 10.0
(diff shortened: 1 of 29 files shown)


src/testdir/test_search.vim
```diff
@@ -1576,6 +1576,31 @@ func Test_search_timeout()
...
+  else
+    let ms = 1
+    let min_time = 0.001
+  endif
+  let max_time = min_time * 10.0
+  let start = reltime()
+  let found = searchpair('(', '', ')', 'crnm', 'Skip()', 0, ms)
+  let elapsed = reltimefloat(reltime(start))
+  call assert_true(elapsed >= min_time)
+  call assert_true(elapsed < max_time)
```
zeertzjq, line 1600:
```suggestion
  call assert_inrange(min_time, max_time, elapsed)
```

brammool:
I have included a few parts of this, but not the change to handle nested calls.  If it's possible to write a test for that (even if only sketched) I'll have another look.

I made the autoconf change for MacOS.  After enabling timer_delete() it seems to work, at least the tests pass.

paul-ollis:
Closing this since other changes cover this now.

PR #10599, opened 2022-06-20
xxd: Add -n[ame] option to set the variable name used with -i

sulix:
At present, ``xxd -i`` will always create an output file either with no variables (if stdin is used as input), or with variables with names
based on the input filename/path. In particular:
- An array of bytes with the name of the input file (with special characters removed), and
- An integer length of the output.

It'd be very convenient to allow overriding these names (or providing them at all in the case of stdin), particularly since they're based on the exact argument passed to xxd, which may be an absolute path. (No one would want to have to access variables based on the absolute path name in their code!)

This patch adds a ``-n`` option, which accepts a new variable name as an argument. This variable name is used for the data, and ``_len`` is appended to it for the length variable. Both still have special characters stripped and replaced with '_', and are capitalised if ``-C`` is used.

For example:
```
xxd -i -n myvar ~/data.dat
```
produces the following variables:
- ``unsigned char myvar[]``
- ``unsigned int myvar_len``
instead of something like:
- ``unsigned char _home_david_data_dat``
- ``unsigned int _home_david_data_dat_len``

### DIFF
--- a/runtime/doc/xxd.1
+++ b/runtime/doc/xxd.1
@@ -113,6 +113,10 @@ Stop after writing
 .RI  < len >
 octets.
 .TP
+.I "\-n name " | " \-name name"
+Override the variable name output when \-i is used. The array is named
+\fIname\fP and the length is named \fIname\fP_len.
+.TP
 .I \-o offset
 Add
 .RI < offset >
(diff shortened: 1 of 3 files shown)


k-takata:
`runtime/doc/xxd.man.1` should be also updated.

sulix:
> `runtime/doc/xxd.man.1` should be also updated.

Thanks — I've updated `runtime/doc/xxd.1`. I haven't regenerated `xxd.man`, as it's adding a bunch of extra formatting changes here. I can look into whatever settings need changing and do so if that's desired, though. Similarly, I haven't attempted to update the translated pages.

brammool:
Please add a test for this in src/testdir/test_xxd.vim

PR #10600, opened 2022-06-20
i3config syntax with better keyword/type distinction

litoj:
This syntax uses a couple more colour highlights to match the actual function of used keywords in the config file. This way it is easier to figure out what does what.
- distinctly highlights key combos to clearly differentiate it from the rest of the command
- all keywords that can be used as command line commands are highlighted as Functions
- conditions are now properly highlighted with all operators clearly visible
+ some bug fixes

file updated at: https://github.com/JosefLitos/vim-i3config


hiqua:
Thanks! Will review.

runtime/syntax/i3config.vim
```diff
@@ -1,9 +1,9 @@
 " Vim syntax file
 " Language: i3 config file
 " Original Author: Mohamed Boughaba <mohamed dot bgb at gmail dot com>
-" Maintainer: Quentin Hibon (github user hiqua)
-" Version: 0.4
-" Last Change: 2022 Jun 05
+" Maintainer: Josef Litoš (github JosefLitos)
```
hiqua: Sure about that? :smile:

runtime/syntax/i3config.vim
```diff
@@ -1,9 +1,9 @@
...
+" Version: 4.20
```
hiqua: I guess we can remove the version line altogether, it makes little sense IMHO.

runtime/syntax/i3config.vim
```diff
@@ -17,39 +17,39 @@ endif
 
 scriptencoding utf-8
 
+" Error
+syn match i3ConfigError /.*/
```
hiqua: I've removed that already in a separate commit, can you remove it as well? It poses problem whenever a new keyword is introduced e.g. in sway, and we're not fully up to date, which is bound to happen regularly.

hiqua (changes requested):
I know it's a long shot, but, before I deep-dive into the code to review it, any chance you could try to reduce the diff? That would make my life way easier.

Otherwise I'll make do with this, but it might take me more time.

JosefLitos:
I will get to it as soon as I can (in August).

JosefLitos:
I reorganized it a little. But the color binds I left untouched as I find it a little clearer this way.

hiqua:
Thanks! I'll have a look.

hiqua:
Hi @JosefLitos, sorry I took some time, and now your PR seems to have conflicts and failing the CI, can you rebase it before I delve into it? I promise I won't be that long after that.

JosefLitos:
@hiqua I am a little afraid I did the rebase procedure incorrectly as all the new commits have been made. But it shouldn't interfere with master anymore.

PR #10603, opened 2022-06-21
getchar(2) waits for character without moving the cursor

ii14:
`getchar()` is useful for mapping dynamic key sequences, but the downside is that it moves the cursor to the command line when waiting for input. If the action depends on where the cursor is, you can lose the visual feedback on where you're about to make an edit. `getchar(2)` behaves just like `getchar()`, but bypasses the move of the cursor.

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -2055,7 +2055,7 @@ char_avail(void)
 getchar_common(typval_T *argvars, typval_T *rettv)
 {
     varnumber_T		n;
-    int			error = FALSE;
+    varnumber_T		arg = -1;
 
     if (in_vim9script() && check_for_opt_bool_arg(argvars, 0) == FAIL)
 	return;
@@ -2067,20 +2067,31 @@ getchar_common(typval_T *argvars, typval_T *rettv)
     parse_queued_messages();
 #endif
 
-    // Position the cursor.  Needed after a message that ends in a space.
-    windgoto(msg_row, msg_col);
+    if (argvars[0].v_type != VAR_UNKNOWN) {
+	int error = FALSE;
+	arg = tv_get_number_chk(&argvars[0], &error);
+	if (error || arg < 0) {
+	    arg = 0;
+	}
+    }
+
+    // getchar(2): blocking wait without moving the cursor
+    if (arg != 2) {
+	// Position the cursor.  Needed after a message that ends in a space.
+	windgoto(msg_row, msg_col);
+    }
 
     ++no_mapping;
     ++allow_keys;
     for (;;)
     {
-	if (argvars[0].v_type == VAR_UNKNOWN)
-	    // getchar(): blocking wait.
+	if (arg == -1 || arg == 2)
+	    // getchar() or getchar(2): blocking wait.
 	    n = plain_vgetc();
-	else if (tv_get_bool_chk(&argvars[0], &error))
+	else if (arg == 1)
 	    // getchar(1): only check if char avail
 	    n = vpeekc_any();
-	else if (error || vpeekc_any() == NUL)
+	else if (vpeekc_any() == NUL)
 	    // illegal argument or getchar(0) and no char avail: return zero
 	    n = 0;
 	else
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -2057,7 +2057,7 @@ getchar_common(typval_T *argvars, typval_T *rettv)
     varnumber_T		n;
     varnumber_T		arg = -1;
 
-    if (in_vim9script() && check_for_opt_bool_arg(argvars, 0) == FAIL)
+    if (in_vim9script() && check_for_opt_bool_or_number_arg(argvars, 0) == FAIL)
 	return;
 
 #ifdef MESSAGE_QUEUE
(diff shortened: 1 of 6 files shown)


src/testdir/test_vim9_builtin.vim
```diff
@@ -1683,8 +1683,7 @@ def Test_getchar()
   endwhile
   getchar(true)->assert_equal(0)
   getchar(1)->assert_equal(0)
-  v9.CheckDefAndScriptFailure(['getchar(2)'], ['E1013: Argument 1: type mismatch, expected bool but got number', 'E1212: Bool required for argument 1'])
-  v9.CheckDefAndScriptFailure(['getchar("1")'], ['E1013: Argument 1: type mismatch, expected bool but got string', 'E1212: Bool required for argument 1'])
+  v9.CheckDefAndScriptFailure(['getchar("1")'], ['E1013: Argument 1: type mismatch, expected bool but got string', 'E1290: Bool or Number required for argument 1'])
```
ii14, line 1686: These tests are probably wrong. I don't get how to change what types vim9script expects.

src/getchar.c
```diff
@@ -2067,20 +2067,31 @@ getchar_common(typval_T *argvars, typval_T *rettv)
     parse_queued_messages();
 #endif
 
-    // Position the cursor.  Needed after a message that ends in a space.
-    windgoto(msg_row, msg_col);
+    if (argvars[0].v_type != VAR_UNKNOWN) {
+	int error = FALSE;
+	arg = tv_get_number_chk(&argvars[0], &error);
```
ii14, line 2072: note to self: vim9script throws `Using a Bool as a Number` here

ii14:
@brammool Hey, before I proceed, do you have any comments on this? Would you be okay with this feature? It works right now with the old vim script, vim9script is broken at the moment.

Here's an example script that adds a window pseudo-mode. `CTRL-W` to enter it, and then `CTRL-{W,P,H,J,K,L}` will sustain that mode until you escape, `CTRL-C` or press some other key.
```vim
let s:remap = {
  \ "\<c-w>": "\<c-w>w",
  \ "\<c-p>": "\<c-w>p",
  \ "\<c-h>": "\<c-w>h",
  \ "\<c-j>": "\<c-w>j",
  \ "\<c-k>": "\<c-w>k",
  \ "\<c-l>": "\<c-w>l",
  \ "w":      "\<c-w>w",
  \ "p":      "\<c-w>p",
  \ "h":      "\<c-w>h",
  \ "j":      "\<c-w>j",
  \ "k":      "\<c-w>k",
  \ "l":      "\<c-w>l",
  \ }

function Getch(...)
  let ch = getcharstr(2)
  if has_key(s:remap, ch)
    call feedkeys(s:remap[ch], 'n')
    call timer_start(0, function('Getch'))
  else
    call feedkeys("\<C-W>"..ch, 'n')
    echo ''
  end
endfunction

nno <nowait> <C-W> <cmd>echo '-- WINDOW --'<bar>call timer_start(0, function('Getch'))<CR>
```

zeertzjq:
There is an entry on `todo.txt`:
https://github.com/vim/vim/blob/4e0fc8956649d3208aeaa1642c5efc44e385d77a/runtime/doc/todo.txt#L213

brammool:
> @brammool Hey, before I proceed, do you have any comments on this?
> Would you be okay with this feature? It works right now with the old
> vim script, vim9script is broken at the moment.

Yes, it appears to be useful.  I'm holding off until after the 9.0
release, no new features right now.

I would think that we need an extra argument for the flag to not
position the cursor.  Then the first, existing argument can be used as
before.

The extra argument can be just a flag, or, if we think there might be
other flags in the future, a dictionary with options.

-- 
It's totally unfair to suggest - as many have - that engineers are socially
inept.  Engineers simply have different objectives when it comes to social
interaction.

ii14:
> I would think that we need an extra argument for the flag to not
position the cursor.  Then the first, existing argument can be used as
before.

@brammool The problem is that `getchar()` has the blocking behavior when the argument is skipped. I might be wrong, but I don't think you can skip the first argument (leaving it as `VAR_UNKNOWN`?) and just supply the second in vim script?

ii14:
@brammool any decision on how you'd like arguments for getchar to work?

brammool:
Looks like we can use -1 for the first argument to get the same behavior as when it is omitted (blocking wait).

One other option that can be useful is to always return a string.  Currently it can return a number or a string, which can be clumsy to handle.  Thus it's best to make the second argument a dictionary, with an entry for each flag: "move" for moving the cursor (default true), "number" for possibly returning a number (default true).  How about that?

ii14:
So with number=false it would basically be `getcharstr()`?

brammool:
Hmm, it might be better to also add the extra argument to getcharstr() and not have the "number" flag.  More consistent.

ii14:
What about `getchar()` accepting either a number or a dictionary as the first argument? A number to be backwards compatible and a dictionary for all the potential new features.

brammool:
> What about `getchar()` accepting either a number or a dictionary as
> the first argument? A number to be backwards compatible and a
> dictionary for all the potential new features.

An argument that can be a number or a dictionary is a bit strange, I
rather not do it like that.

-- 
You got to work at a mill?  Lucky!  I got sent back to work in the
acid-mines for my daily crust of stale bread... which not even the
birds would eat.

chrisbra:
This seems to have stalled and the tests seems to be wrong. So I am closing this for now. If you have a working version, please consider opening a new PR. thanks!

PR #10611, opened 2022-06-23
Move f_hasmapto() to map.c

zeertzjq:
Also rename two function to better match what they do:
- Rename map_clear() to do_mapclear() as it also does the command line
  parsing in :mapclear in :abclear.
- Rename map_clear_int() to map_clear_mode() to be more descriptive.


zeertzjq:
CI seems to be having some internet problems.

PR #10613, opened 2022-06-24
Handle more edge cases in Win32 GUI

### LINKED ISSUE #10579: CTRL+key combinations don't work in gvim with alternative input language after 8.2.4807 in Windows
OS: Windows 10, Windows 11.
gvim version: 8.2.4807 or newer

The problem is with the keyboard behavior in insert mode in gvim when keyboard layout is switched to another input language (Russian, in my case).

When trying completion while switched to Russian (switching system's keyboard input language, which is left_ALT+SHIFT by default, just in case), C-n just prints т, and C-p prints з. This affects all CTRL+key commands (corresponding letters are printed instead).

Everything works fine in command-line vim. 

Just in case, I tried gvim from https://tuxproject.de/projects/vim/ , as well as the one from https://github.com/vim/vim-win32-installer/releases

Tried to bisect (using installers from https://github.com/vim/vim-win32-installer/releases ), and now I'm pretty sure https://github.com/vim/vim/commit/77fc0b02e53bdc49213900e0055cb6d678ce180d
(patch 8.2.4807: processing key eveints in Win32 GUI is not ideal) is the one that broke it.
I.e. completion (and other CTRL+key combinations) works fine in INSERT mode in https://github.com/vim/vim-win32-installer/releases/tag/v8.2.4803 , but does not in more recent versions.

### LINKED ISSUE #10602: different behavier between vim and gvim, about `<C-_>`
### Steps to reproduce

1. open vim.exe, type `Ctrl` and `-`, wait sometimes, it will show like this:
image

2. open gvim.exe, type as before, it will show like this:
image

3. The different behavier will make some bugs, such as tcomment.vim use `<c-_><c-_>` to comment line, but now it counld not works well.

### Expected behaviour

gvim should output as vim.
image

### Version of Vim

8.2.5136

### Environment

windows 10 x64

LemonBoy:
Closes #10602
Closes #10579

### DIFF
--- a/src/gui_w32.c
+++ b/src/gui_w32.c
@@ -2055,21 +2055,21 @@ process_message(void)
 	    int		i;
 	    UINT	scan_code;
 
-	    // Construct the state table with only a few modifiers, we don't
-	    // really care about the presence of Ctrl/Alt as those modifiers are
-	    // handled by Vim separately.
+	    // Construct the keyboard state table, the modifiers can and will
+	    // affect the character translation performed by ToUnicode.
+	    // Eg. With a Russian keyboard layout pressing 'n' produces 'т' but
+	    // Ctrl+p produces 'p', this is essential for the keybindings to
+	    // work.
 	    memset(keyboard_state, 0, 256);
+	    if (GetKeyState(VK_CONTROL) & 0x8000)
+		keyboard_state[VK_CONTROL] = 0x80;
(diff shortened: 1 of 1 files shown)


LemonBoy:
@dantezhu @ladayaroslav let me know if this commit fixes the problem for you.

PR #10619, opened 2022-06-25
add gui_gtk4 to test_gui

pheiduck:


brammool:
has("gui_gtk4") is always false.  Is something missing?

PR #10621, opened 2022-06-26
Added Python runtime syntax soft keywords

dzfrias:
## Problem
In Python's current syntax highlighting, the words `match` and `case` are always highlighted, which could give the wrong idea about their usage. `match` and `case` are only given special meaning when used in a specific context, so it would be misleading to highlight them under the same conditions of `if` and `else`.

## Fix
Added `match` and `case` as soft keywords, as described [here](https://peps.python.org/pep-0634/#the-match-statement). Modeled after Python 3.10's own IDLE behavior when dealing with soft keywords.

I contacted the maintainer of the file about this pull request.

### DIFF
--- a/runtime/syntax/python.vim
+++ b/runtime/syntax/python.vim
@@ -95,8 +95,8 @@ syn keyword pythonAsync		async await
 " See https://peps.python.org/pep-0634/#the-match-statement for more on this.
 " In Python 3.10's IDLE for example, 'match' and 'case' are only highlighted
 " when in the right context
-syn match   pythonConditional   "\v^\s*case(.*:$)@="
-syn match   pythonConditional   "\v^\s*match(.*:$)@="
+syn match   pythonConditional   "\v^\s*case(.*:\s*$)@="
+syn match   pythonConditional   "\v^\s*match(.*:\s*$)@="
 
 " Decorators
 " A dot must be allowed because of @MyClass.myfunc decorators.
--- a/runtime/syntax/python.vim
+++ b/runtime/syntax/python.vim
@@ -1,7 +1,7 @@
 " Vim syntax file
 " Language:	Python
 " Maintainer:	Zvezdan Petkovic <zpetkovic@acm.org>
-" Last Change:	2021 Dec 10
+" Last Change:	2022 Jun 25
 " Credits:	Neil Schemenauer <nas@python.ca>
 "		Dmitry Vasiliev
 "
@@ -84,12 +84,19 @@ syn keyword pythonStatement	as assert break continue del global
 syn keyword pythonStatement	lambda nonlocal pass return with yield
 syn keyword pythonStatement	class def nextgroup=pythonFunction skipwhite
 syn keyword pythonConditional	elif else if
-syn keyword pythonConditional	case match
 syn keyword pythonRepeat	for while
 syn keyword pythonOperator	and in is not or
 syn keyword pythonException	except finally raise try
 syn keyword pythonInclude	from import
 syn keyword pythonAsync		async await
+
+" Soft keywords
+" These keywords do not mean anything unless used in the right context
+" See https://peps.python.org/pep-0634/#the-match-statement for more on this.
+" In Python 3.10's IDLE for example, 'match' and 'case' are only highlighted
+" when in the right context
+syn match   pythonConditional   "\v^\s*case(.*:$)@="
+syn match   pythonConditional   "\v^\s*match(.*:$)@="
 
 " Decorators
 " A dot must be allowed because of @MyClass.myfunc decorators.
--- a/runtime/syntax/python.vim
+++ b/runtime/syntax/python.vim
@@ -94,8 +94,8 @@ syn keyword pythonAsync		async await
 " These keywords do not mean anything unless used in the right context
 " See https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords 
 " for more on this.
-syn match   pythonConditional   "^\s*\zscase\%(\s.*:.*$\)\@="
-syn match   pythonConditional   "^\s*\zsmatch\%(\s.*:\s*\%(#.*\)\?$\)\@="
+syn match   pythonConditional   "^\s*\zscase\%(\s\+.*:.*$\)\@="
+syn match   pythonConditional   "^\s*\zsmatch\%(\s\+.*:\s*\%(#.*\)\=$\)\@="
 
 " Decorators
 " A dot must be allowed because of @MyClass.myfunc decorators.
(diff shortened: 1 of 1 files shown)


runtime/syntax/python.vim
```diff
@@ -84,13 +84,20 @@ syn keyword pythonStatement	as assert break continue del global
...
 syn keyword pythonInclude	from import
 syn keyword pythonAsync		async await
 
+" Soft keywords
+" These keywords do not mean anything unless used in the right context
+" See https://peps.python.org/pep-0634/#the-match-statement for more on this.
+" In Python 3.10's IDLE for example, 'match' and 'case' are only highlighted
+" when in the right context
+syn match   pythonConditional   "\v^\s*case(.*:$)@="
+syn match   pythonConditional   "\v^\s*match(.*:$)@="
```
dkearns: It looks like you need to allow trailing whitespace after the colons.

dzfrias: I think this should be fixed now. Thanks for letting me know

brammool:
@zvezdan

runtime/syntax/python.vim
```diff
@@ -84,13 +84,20 @@ syn keyword pythonStatement	as assert break continue del global
...
 syn keyword pythonInclude	from import
 syn keyword pythonAsync		async await
 
+" Soft keywords
+" These keywords do not mean anything unless used in the right context
+" See https://peps.python.org/pep-0634/#the-match-statement for more on this.
+" In Python 3.10's IDLE for example, 'match' and 'case' are only highlighted
+" when in the right context
+syn match   pythonConditional   "\v^\s*case(.*:\s*$)@="
+syn match   pythonConditional   "\v^\s*match(.*:\s*$)@="
```
zvezdan: A few issues with these regular expressions:

* Both `case` and `match` must be followed by a `\s` before any other content -- match subject or case pattern(s). These regular expressions are not enforcing this space separation and would accept `matchX` or `caseY`. There needs to be a `\s` as the first part of the content inside the parentheses for zero-width look-ahead match (`@=`).
* According to grammar, the `match subject:` must be followed by a newline, indented block, and dedent, which allows for the check of `$` at the end of the regex. The `case pattern:` is followed by a _block_ according to grammar. This means that besides "newline, indented block, dedent" it can be followed by a simple statement. So a check for `$` is not necessarily correct in that case. It can be a `\s*$`, but it could also be other characters that could make a simple statement between the colon and the end of the line.
* The patterns in this file _never_ use `\v` (very magic) and I'd keep the prevailing style. The non-capturing parentheses could be used `\%( ... \)@=` and are the only part of the regex that requires a slightly more complex writing without `\v` which doesn't seem too bad.
* The `^\s*` makes the space in front of `match` or `case` the part of this match, while only the keywords should be matched. See some examples in the other regular expressions in this file that set `\zs` for the start of the match after finding the pre-requisite part in front of the match.

runtime/syntax/python.vim
```diff
@@ -84,13 +84,20 @@ syn keyword pythonStatement	as assert break continue del global
...
-syn keyword pythonConditional	case match
 syn keyword pythonRepeat	for while
 syn keyword pythonOperator	and in is not or
 syn keyword pythonException	except finally raise try
 syn keyword pythonInclude	from import
 syn keyword pythonAsync		async await
 
+" Soft keywords
+" These keywords do not mean anything unless used in the right context
+" See https://peps.python.org/pep-0634/#the-match-statement for more on this.
```
zvezdan: I would refer to the Python documentation section on soft keywords instead: https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords

That should be enough without any further explanation except the heading _Soft keywords_.

runtime/syntax/python.vim
```diff
@@ -84,13 +84,20 @@ syn keyword pythonStatement	as assert break continue del global
...
+" In Python 3.10's IDLE for example, 'match' and 'case' are only highlighted
+" when in the right context
```
zvezdan: This is irrelevant. Grammar is the relevant reason why we need this change, not what IDLE does. I would remove this part of the comment.

zvezdan (changes requested):
Thank you for noticing this issue and working on the fix.
I asked for some changes.
If they are not clear or you don't feel like working out all the complexities, feel free to let me know and I'll add the changes to a later update of the syntax file.

dzfrias:
Thanks for reviewing this! I added the changes requested, tell me if I did something wrong. I also added the ability to make inline comments after a `match` statement.

runtime/syntax/python.vim
```diff
@@ -84,13 +84,19 @@ syn keyword pythonStatement	as assert break continue del global
...
 syn keyword pythonOperator	and in is not or
 syn keyword pythonException	except finally raise try
 syn keyword pythonInclude	from import
 syn keyword pythonAsync		async await
 
+" Soft keywords
+" These keywords do not mean anything unless used in the right context
+" See https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords 
+" for more on this.
+syn match   pythonConditional   "^\s*\zscase\%(\s.*:.*$\)\@="
```
zvezdan: It's probably not frequent, but one could write `case   4` which conforms to the grammar as much as `case 4` -- multiple spaces after `case` vs. a single space. So, I think we should have `\s\+` after `case` and before `.*:`.

runtime/syntax/python.vim
```diff
@@ -84,13 +84,19 @@ syn keyword pythonStatement	as assert break continue del global
...
+syn match   pythonConditional   "^\s*\zsmatch\%(\s.*:\s*\%(#.*\)\?$\)\@="
```
zvezdan: Same comment about `\s\+` after `match` and before `.*:`.
Additionally, the regular expressions in this file use `\=` instead of `\?` consistently. Perhaps we can get that replaced too, together with the multi-space fix?

Again, thank you, and sorry for the long delay for this second review. I was unavailable, but next review should be quick. This is looking practically ready. In fact, if you're tired of changing it, we might commit as is and I can change in the subsequent commit. If you don't mind making these two small changes, even better. 🙂

dzfrias: I added those changes. No problem at all, hopefully they look okay now

dzfrias:
Do these look good now @zvezdan? Let me know if I need to do more or messed something up.

zvezdan (approved):
Looks good!
Thanks!

zvezdan:
> Do these look good now @zvezdan? Let me know if I need to do more or messed something up.

Yes. I just tested them also and the highlighting looks correct.
Thanks for changing this.

brammool:
I'll include it, thanks

PR #10629, opened 2022-06-28
Session file (tabs section): switch to existing buffers instead of re-editing files

jamescherti:
Problem: Session files (tabs section) use `:edit` to open files even when the file buffer is loaded.

Solution: Switch to the buffer with `:buffer` if it exists.

### DIFF
--- a/src/session.c
+++ b/src/session.c
@@ -827,9 +827,14 @@ makeopens(
 		    goto fail;
 		need_tabnext = FALSE;
 
-		if (fputs("edit ", fd) < 0
-			      || ses_fname(fd, wp->w_buffer, &ssop_flags, TRUE)
-								       == FAIL)
+		if (fputs("if bufexists(fnamemodify('", fd) < 0
+		  || ses_fname(fd, wp->w_buffer, &ssop_flags, FALSE) == FAIL
+		  || fputs("', ':p')) | buffer ", fd) < 0
+		  || ses_fname(fd, wp->w_buffer, &ssop_flags, FALSE) == FAIL
+		  || fputs(" | else | edit ", fd) < 0
+		  || ses_fname(fd, wp->w_buffer, &ssop_flags, FALSE) == FAIL
+		  || fputs(" | endif", fd) < 0
+		  || put_eol(fd) == FAIL)
 		    goto fail;
 		if (!wp->w_arg_idx_invalid)
 		    edited_win = wp;


jamescherti:
> In which situation does this make a difference?

In my case, loading session files reset all folds when a file is open in more than one tab.

One of the disadvantages of `:edit` is that it will execute the autocmds that are supposed to be executed a file is loaded (e.g. several plugins and configuration files use autocmds like BufReadPost).

I do not see any disadvantage in using `:buffer` over `:edit` when the buffer is already loaded. How about you, Bram?

> Can you give an example?

I could not reproduce the issue with session files generated with `vim -u NONE`. 

**Steps to reproduce it manually (without a session file):**

1. Open the same file in more than one tab:
```
:tabnew file1
:tabnew file1
:tabnew file1
:tabnew file1
:set foldmethod=indent
```

2. Open/close folds in each tab.

3. Re-edit file1 one or two times:
```
:edit file1
```

4. Vim will reset the folds in all tabs.

jamescherti:
> OK, so this is for when you already have been editing some files and then load the session. 

Not necessarily. The issue also happens when the session is loaded before editing files.

> The question then is what is the intention? Keep most of the current state, or reset it to how it was when the session was created?

The intention is to reset the state to how it was when `:mksession` created the session. Does using `:buffer` instead of multiple `:edit` prevent us from achieving that?

> When the session was created, 'foldmethod' was not set to "indent"?
> Then a user might use the session to get back that state.

The fold method was set to "indent" before the session was created.

> Thus it's not clear to me if this changes does what most users expect.

There is a similar code in `put_view()` (session.c):
```
	    // Editing a file in this buffer: use ":edit file".
	    // This may have side effects! (e.g., compressed or network file).
	    //
	    // Note, if a buffer for that file already exists, use :badd to
	    // edit that buffer, to not lose folding information (:edit resets
	    // folds in other buffers)
	    if (fputs("if bufexists(fnamemodify(\"", fd) < 0
		    || ses_fname(fd, wp->w_buffer, flagp, FALSE) == FAIL
		    || fputs("\", \":p\")) | buffer ", fd) < 0
		    || ses_fname(fd, wp->w_buffer, flagp, FALSE) == FAIL
		    || fputs(" | else | edit ", fd) < 0
		    || ses_fname(fd, wp->w_buffer, flagp, FALSE) == FAIL
		    || fputs(" | endif", fd) < 0
		    || put_eol(fd) == FAIL)
		return FAIL;
```

Just out of curiosity, why was the `bufexists()` check `:buffer` added to session.c in 2019?

Is there any reason Vim should not do the same in makeopens()? (this pull request)

jamescherti:
I couldn't reproduce this issue anymore. I am going to close this pull request.

PR #10630, opened 2022-06-28
Recognise *.ha files as `hare`

WhyNotHugo:
See: https://harelang.org/


WhyNotHugo:
On 2022-06-28 03:06, Bram Moolenaar wrote:
> 
> Please also add an entry in the test, src/testdir/test_filetype.vim
> 

Done

-- 
Hugo Osvaldo Barrera

WhyNotHugo:
Thanks!

PR #10634, opened 2022-06-28
feat: enable matchit for html tags in vue files

bwskin:
Problem:
In vue files `%` key (mapped to matchit) is not working for html tags
because of missing `b:match_words` value

Solution:
- Create `runtime/ftplugin/vue.vim` file.
- Copy `b:match_words` value from `runtime/ftplugin/html.vim`

### DIFF
--- a/runtime/ftplugin/vue.vim
+++ b/runtime/ftplugin/vue.vim
@@ -1,0 +1,15 @@
+" Vim filetype plugin file
+" Language:	vue
+
+if exists("b:did_ftplugin") | finish | endif
+let b:did_ftplugin = 1
+
+" Copied from ftplugin/html.vim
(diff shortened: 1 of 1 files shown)


brammool:
Please add the check for "b:did_ftplugin", like in other ftplugin files.

bwskin:
Thank you for reply. Added that check.
It is my first contribution, so I would like to ask: Why is that necessary? I know that it prevents running more than one time per buffer, but why this could happen? Running on `BufEnter` event or something simmilar?

brammool:
When a filetype plugin is already loaded, it sets b:did_ftplugin to avoid that further plugins, later in 'runtimepath' change the settings.
I'll include it with a small change: The dot for string concatenation at the end of a line changed to a double dot at the start of the next line.  Just reads better that way.

PR #10642, opened 2022-07-01
Convert the cfilter plugin to use vim9script

yegappan:


brammool:
Thanks.  I suppose this can go with the runtime file updates.

PR #10648, opened 2022-07-01
runtime/syntax/fetchmail.vim: add "pidfile"

jwilk:


jwilk:
@dkearns

dkearns:
Thanks. LGTM

brammool:
I'll include it then.  And update the date in the header.

Issue #10503, opened 2022-05-30
Inconsistent timeoutlen behavior between mappings and built-in commands


andmis:
### Steps to reproduce

```
set timeout
set timeoutlen=1000
map gx <Cmd>echom "gx"<CR>
map gxy <Cmd>echom "gxy"<CR>

" Press gxy → prints "gxy"
" Press gx → wait 1 second → prints "gx"
" This works as expected.

map g <Cmd>echom "g"<CR>
" Press g → wait 1 second → prints "g"
" Press gg → prints "g" twice, instead of executing the gg built-in.
```

### Expected behaviour

`gg` goes to the first line of the file, rather than printing "g" twice.

### Version of Vim

Latest

### Environment

macOS 12

brammool:
This is working as intended.
Not sure what you were trying to do. If you want "g" to do one thing and "gg" to do something else, you must also map "gg".
And you probably want to use :noremap instead of :map.

andmis:
I want `g` to do something custom, and `gg` to do the default, built-in command after `timeoutlen`, along with every other `g`-prefixed command in `:help g`.

Right now I can `:map g <MyCustomThing>` and `:noremap gg gg`, but then I need to `:noremap g<Key> g<Key>` for every single `g<Key>` command in `:help g`.

andmis:
In case you care, this is to enable a plugin similar to https://github.com/liuchengxu/vim-which-key to work with built-in commands.

The natural way to implement this is to map `g` to show a pop-up, so that the pop-up appears after `timeoutlen`, but if the user completes a command by typing `gg` quickly, then there is no need for the pop-up to appear.

brammool:
The letter "g" is the prefix for many commands, so perhaps it's not a great idea to map it.

andmis:
I agree in general, but in my use case (a plugin that pops up a hint window showing the mappings having the prefix entered so far), the whole reason to map it is that there are commands that start with the letter.

zeertzjq:
This may also be achieved using `getchar()` and `feedkeys()` in the mapping.

andmis:
How can you get the popup to appear when pressing `g` after a timeout without creating a user mapping for `g` and thus disabling the built-ins?

You can emulate all of the built-ins by using `feedkeys` without remap from plugin code, but that really seems like a hack. For example, wouldn't I need to manually forward registers and ranges?

zeertzjq:
That's why I said "this specific purpose". In some other cases it is not very convenient.

[...]

PR #10652, opened 2022-07-03
Add a way to know if <expr> mapping is triggered on timeout

zeertzjq:
I've currently chosen `v:timedout` as the name, as this name may be reused for other timeout-related things in future.

### DIFF
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -2544,6 +2544,10 @@ v:testing	Must be set before using `test_garbagecollect_now()`.
 		Also, when set certain error messages won't be shown for 2
 		seconds. (e.g. "'dictionary' option is empty")
 
+					*v:timedout* *timedout-variable*
+v:timedout	Set to |TRUE| when inside an <expr> mapping triggered on
+		timeout and |FALSE| otherwise.
+
 				*v:this_session* *this_session-variable*
 v:this_session	Full filename of the last loaded or saved session file.  See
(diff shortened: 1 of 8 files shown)


zeertzjq:
The initial use case I thought of is to enable a way to make a mapping prefix "never time out", for example: 
```vim
nmap <expr> g v:timedout ? '<Ignore>g' : 'g'
```
Makes mapping starting with `g` behave as if they never time out. Only using `nmap g <Ignore>g` may also achieve this, but will lead to high CPU usage, whereas using this example the CPU usage is fine as long as `'timeoutlen'` is non-zero.

Another use case I noticed is to possibly achieve the purpose mentioned in #10503. Using this mapping (`g<Plug>` is a dummy mapping to avoid triggering `g` mapping immediately):
```
nnoremap <expr> g v:timedout ? '<Cmd>echo 123<CR>' : 'g'
nnoremap g<Plug> <Nop>
```
Make pressing `gg` within `'timeoutlen'` behave like builtin `gg` command, while a timed out `g` prints `123`.

brammool:
> I've currently chosen `v:timedout` as the name, as this name may be
> reused for other timeout-related things in future.

What do you need this for?

The help says "mapping triggered on timeout".  What does that mean
exactly?  I suppose 'timeoutlen' and 'ttimeoutlen' matter.


-- 
Nothing is fool-proof to a sufficiently talented fool.

zeertzjq:
This isn't very useful and can also be implemented by calculating the time difference between two mappings. Closing.

Issue #10656, opened 2022-07-03
Different behavior in map command for imported vim9 functions


dvogel:
### Steps to reproduce

In `demo9.vim` place:
```
vim9script
export def ReturnA(): string
    return "A"
enddef
defcompile
```

In `main.vim` place:
```
set nocompatible
import "./demo9.vim" as demo9
function s:ReturnB()
    return "B"
endfunction
echo s:demo9.ReturnA()
let s:ReturnA = s:demo9.ReturnA
nmap <F3> :echo <SID>ReturnB()<CR>
nmap <F4> :echo <SID>ReturnA()<CR>
nmap <F5> :echo <SID>demo9.ReturnA()<CR>
```

Run vim as:
```
vim --clean -u main.vim
```

On startup it will echo A
F3 will echo B
F4 will echo A
F5 will yield E121 Undefined variable <SNR>1_demo9

### Expected behaviour

F3 should echo B
F4 should echo A
F5 should echo A

The `vim9.txt` help says:

```
`:import` can also be used in legacy Vim script.  The imported items still
become script-local, even when the "s:" prefix is not given.
```

This seems to say that the import alias will create a local variable `s:demo9`. That seems to be verified by the call:

```
echo s:demo9.ReturnA()
```

The ability reference script locals in a `map` command seems to be verified by:

```
nmap <F3> :echo <SID>ReturnB()<CR>
```

The ability to call compiled functions via script local references seems to be verified by:

```
let s:ReturnA = s:demo9.ReturnA
nmap <F4> :echo <SID>ReturnA()<CR>
```

The F5 map command combines all of these but has a different result.

### Version of Vim

9.0.0030

### Environment

debian/sid, TERM=xterm-kitty

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jul  2 2022 16:32:10)
Included patches: 1-30
Compiled by dvogel@bitbox
Huge version with GTK3 GUI.  Features included (+) or not (-):
+acl               +file_in_path      +mouse_urxvt       -tag_any_white
+arabic            +find_in_path      +mouse_xterm       -tcl
+autocmd           +float             +multi_byte        +termguicolors
+autochdir         +folding           +multi_lang        +terminal
-autoservername    -footer            -mzscheme          +terminfo
+balloon_eval      +fork()            +netbeans_intg     +termresponse
+balloon_eval_term +gettext           +num64             +textobjects
+browse            -hangul_input      +packages          +textprop
[...]

PR #10657, opened 2022-07-03
Problem: docs elide import semantics when used in vim9script vs legacy script

dvogel:
Problem: docs elide import semantics when used in vim9script vs legacy script.
Solution: Provide clarifying examples.

This updates the documentation based on my experience as reported in https://github.com/vim/vim/issues/10656. I've condensed the import syntax explanation so the generic syntax and specific examples are closer together. I then moved the legacy import points to their own section with tags identifying the section. I've moved this below the exceptions to the main vim9script behavior in order to more clearly delineate what is specific to legacy imports.


brammool:
Thanks.  I'll include it and tweak it a bit more.  I don't like comments in the examples explaining what happens.

Issue #10454, opened 2022-05-20
Ctrl-[ don't escape from insert mode (like Esc) on Belgian keyboard


vds2212:
### Steps to reproduce

On a computer with a Belgian keyboard using gVim

1. Start Vim
2. Enter in insert mode
3. Type: Ctrl-[

### Expected behaviour

You should be back in Normal mode but you stay in insert mode.

### Version of Vim

8.2.4943 on Windows 10 8.2.3995 on Ubuntu 22.04

### Environment

Operating System: Windows 10 or Ubuntu 22.04

I can only reproduce the problem on gVim (on Vim the problem do not occur)

vds2212:
I have compiled different tags of Vim on Windows 10.
I don't have the problem at version 8.2.4006
But I do have the problem at version 8.2.4007 and later

chrisbra:
I think this is a duplicate of https://github.com/vim/vim/issues/10324

brammool:
@LemonBoy was working on a solution, but it didn't get completed yet.

vds2212:
Thanks! Let me know if I can contribute in any ways.

brammool:
Patch 9.0.0033 fixes this for GTK, but it is still a problem for MS-Windows

vds2212:
Thanks for the update :-)

tonymec:
WFM on gvim (Big, GTK3) 9.0.32 i.e. without the patch. Don't forget, however, that on a Belgian keyboard, [ is AltGr+dead-circumflex, therefore Ctrl+[ is Ctrl+AltGr+dead-circumflex.

Best regards,
Tony.

vds2212:
Hi Tony,

I have tested version 9.0.35 and the problem is still present in gVim.

On gVim 8.2 I can escape using: Ctrl+DeadCircumflex
On gVim 9.0 I can't escape neither Ctrl+DeadCircumflex nor Ctrl+AltGr+DeadCircumflex

Remark: The problem is not present on Vim 9.0

Best regards,
Vivian

PR #10658, opened 2022-07-03
gtk GUI: Fix Ctrl+[ (Belgian,German keyb. layout)

haron13-2019:
Problem: issue #10454 reported troubles with leaving insert mode on Belgian keyboard in GTK GUI. Handling of Ctrl+ü on German keyboard is lost over the time as well in GTK GUI.

This PR partly fixes issue #10454 (only GTK part of it, not Win32). Additionally it provides back missing Ctrl+ü handling on German keyboard (ü is on same place in german layout where [ is on english keyboard). Fix is probably not very nice and cannot be yet considered a final solution (?) but it seems to work. It was made possible due to the fact that Motif GUI still handles both Belgian and German CTRL+[ nicely. Tracing/comparing of the key codes feed to background Vim engine by Motif GUI and by GTK GUI via add_to_input_buf() was the key.

### DIFF
--- a/src/gui_gtk_x11.c
+++ b/src/gui_gtk_x11.c
@@ -1160,6 +1160,7 @@ key_press_event(GtkWidget *widget UNUSED,
     int		key;
     guint	state;
     char_u	*s, *d;
+    int         ctrl_prefix_added = 0;
 
     gui.event_time = event->time;
     key_sym = event->keyval;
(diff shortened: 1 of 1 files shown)


vds2212:
I have tested release 9.00.35 and build release 9.00.48 and I still have the problem.
I'm afraid the patch doesn't solves the problem and issue should be reopen.

vim-ml:
Hi, Vivian,

Did you made your tests on Linux or on windows, because my patch is only
for unix-like GTK gvim. What does your :version command in gvim reports?

(Windows version of the patch is in pipeline yet ))))

Vivian De Smedt ***@***.***> schrieb am Sa., 9. Juli 2022,
15:34:

vds2212:
Hi Vim,

I do my test on Windows 10 on Vim version 9.00.48.

Best regards,
Vivian.

haron13-2019:
Hi Vivian,

Just created a PR https://github.com/vim/vim/pull/10687 which hopefully solves your issue on windows.

If you can build your own windows binary you can try my branch directly:

https://github.com/haron13-2019/vim/tree/azerty_ctrl_dead

With best regards, Anton

vds2212:
Hi Anton,

Thanks for the branch (https://github.com/haron13-2019/vim.git).
I only realize that you wrote to me few minutes ago (sorry for that).

I have tested your branch but unfortunately it doesn't solve my problem.

Best regards,
Vivian.

Issue #1516, opened 2017-02-26
deleted text is stored in PRIMARY even when clipboard=unnamedplus


tertium:
I would like to configure vim to only store the yanked/deleted text in CLIPBOARD and never touch PRIMARY, unless maybe via modeless selection. I would expect that "set clipboard=unnamedplus" or "set clipboard=unnamedplus,autoselectml" would provide the desired configuration, since neither "unnamed" nor "autoselect" are included in the option. However, this does not work as vim modifies both registers '*' and '+' when I delete text using d, x, s, c, etc.

A probably related issue is that deletion still modifies both registers when I set clipboard=unnamed,unnamedplus. This contradicts the documentation for "clipboard-unnamedplus" which says the following: «When "unnamed" is also included to the option, yank operations (but not delete, change or put) will additionally copy the text into register '*'.»

This is observed in Vim 8.0.322 as distributed in Debian's vim-gtk (huge version).

makmanalp:
Possibly related? https://github.com/vim/vim/issues/264

nuko8:
Reason?  I'm afraid it has nothing to do with #264.

makmanalp:
@nuko8 My fault!

inkarkat:
I just stumbled across this issue. It's still reproducible in Vim 8.2.2560. It gets triggered when the `a` (_autoselect_) option is removed from `'guioptions'`:

###### good: Both yanks and deletes only affect the clipboard, but not the primary selection.

```
vim -g --clean --cmd 'set clipboard=unnamedplus' -c 'let @*="" | let @+="" | call append(0, [1, 2])'
:1yank
:reg *+
Type Name Content
  c  "*
  l  "+   1^J
:2delete
Type Name Content
  c  "*
  l  "+   2^J
```

###### bad: Yanks still only affect the clipboard, but the primary selection is updated on deletes or changes.

```
vim -g --clean --cmd 'set clipboard=unnamedplus,guioptions-=a' -c 'let @*="" | let @+="" | call append(0, [1, 2])'
:1yank
:reg *+
Type Name Content
  c  "*
  l  "+   1^J
:2delete
Type Name Content
  l  "*   2^J
  l  "+   2^J
```

So, after `:set guioptions-=a`, the documented behavior of `:set clipboard=unnamedplus` is broken. Likewise, with `:set clipboard=unnamed,unnamedplus`, the **following** is violated:

[...]

PR #10669, opened 2022-07-06
Do not modify "* reg if op_delete and both unnamed,unnamedplus.

errael:
There's either a bug in the doc for `unnamedplus` or in the implementation. Since the doc looked intentional, I went with the code.

### DIFF
--- a/src/register.c
+++ b/src/register.c
@@ -1409,7 +1409,8 @@ op_yank(oparg_T *oap, int deleting, int mess)
 # ifdef FEAT_X11
     // If we were yanking to the '+' register, send result to selection.
     // Also copy to the '*' register, in case auto-select is off.  But not when
-    // 'clipboard' has "unnamedplus" and not "unnamed".
+    // 'clipboard' has "unnamedplus" and not "unnamed"; and not when
+    // deleting and both "unnamedplus" and "unnamed".
     if (clip_plus.available
 	    && (curr == &(y_regs[PLUS_REGISTER])
 		|| (!deleting && oap->regname == 0
@@ -1425,6 +1426,8 @@ op_yank(oparg_T *oap, int deleting, int mess)
 	if (!clip_isautosel_star()
 		&& !clip_isautosel_plus()
 		&& !((clip_unnamed | clip_unnamed_saved) == CLIP_UNNAMED_PLUS)
+		&& !(deleting && (clip_unnamed | clip_unnamed_saved)
+					== (CLIP_UNNAMED|CLIP_UNNAMED_PLUS))
 		&& !did_star
 		&& curr == &(y_regs[PLUS_REGISTER]))
 	{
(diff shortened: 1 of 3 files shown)


errael:
The test I added for this PR doesn't work on windows (no surprise). I don't see any testing for platform in test_registers.vim, and I'm not certain of the proper test for supporting both `"*` and `"+`.

brammool:
On MS-Windows there is only one clipboard, "* and "+ are the same thing.
You have to handle that in the test.  Could be better explained in the help.

errael:
> On MS-Windows there is only one clipboard, "* and "+ are the same thing

Right. But what I don't know is how to check in a test if both `"*` and `"+` are supported on the platform where the test is running. Or is the idea to check if running on windows, and if not then assume both clipboards are supported?

errael:
Put the both clipboard unnamed test into it's own function; don't run it on windows.

errael:
`macos` fails. Does `macos` also have a single clipboard, or other clipboard issue to consider?

errael:
Added `checkTwoClipboards` to enable execution of the new test.

errael:
#1516 mentions this problem, but the fix does not handle this case.

@inkarkat your fix is possibly more `autoselect` related (I'm fuzzy on all the `clipboard` options). Appreciate any comment or insight you might have about this PR; the included test is straightforward.

errael:
AFAICT, this is ready to go. It passes the test I added (the test fails without the change). I'm not familiar with this area at all, and there may be considerations I'm missing. Is there someone familiar with the clipboard and using both unnamed,unnamedplus who can review it?

brammool:
> Added `checkTwoClipboards` to enable execution of the new test.

Hmm, I think just checking for X11 is more reliable.  The clipboard is
shared with other processes, it may change any time.  This already makes
tests flaky, adding more changes to the clipboard makes it a bit less
reliable.  Using has('x11') is expected to just work.

I actually often have this problem, since running tests takes quite a
bit of time.  When I copy something to put it somewhere else a test may
have changed it and I have to try again...

-- 
83. Batteries in the TV remote now last for months.

brammool:
> `macos` fails. Does `macos` also have a single clipboard, or other
> clipboard issue to consider?

Yes, I think that only X11 supports both a selection and a clipboard.
Thus checking with has('x11') is probably the most straightforward.

The "+ register is explained in the gui_x11.txt help file.  That sort-of
implies it is only supported for X11.  But not only the GUI.

-- 
69. Yahoo welcomes you with your own start page

errael:
> > Added `checkTwoClipboards` to enable execution of the new test.

> Hmm, I think just checking for X11 is more reliable. 

(not sure how a recent comment came about to the 3 month old PR that was integrated)

I think this comment was made, maybe on email list, I saw it and changed `CheckTwoClipboards` to

```
" Command to check that there are two clipboards
command CheckTwoClipboards call CheckTwoClipboards()
func CheckTwoClipboards()
  " avoid changing the clipboard here, only X11 supports both
  if !has('X11')
    throw 'Skipped: requires two clipboards'
  endif
endfunc
```

The idea is to make it clear what the check is for.

PR #7969, opened 2021-03-15
Runtime: add i3 config from https://github.com/mboughaba/i3config.vim

hiqua:
I'm not the author of this code.

Taken from: https://github.com/mboughaba/i3config.vim

Suggested in this issue where the author approved of the idea: https://github.com/mboughaba/i3config.vim/issues/34

License: MIT, so relicensing under the Vim license should be ok.

PR #10672, opened 2022-07-07
Runtime: add better sway support as extension of i3config.vim

jamespeapen:
This based on my fork of the current i3config [parent repo](https://github.com/mboughaba/i3config.vim) that was added here through #7969. I have added support for sway specific keywords and extended options based on conversations in https://github.com/vim/vim/issues/10231#issuecomment-1105485718. I am actively maintaining my fork and can maintain the sway syntax file here.

In this PR, I've updated the i3config syntax, both with i3 options (`smart_gaps off`), and for getting it to support the sway syntax.


hiqua:
LGTM for the i3 syntax files. Regarding the error highlighting that you've enabled again, we can see if we manage to keep up with new keywords, hopefully that shouldn't be too frequent, so I'd say it's fine.

Issue #10571, opened 2022-06-14
msgfunc - Allow the user to add a custom message sink (e.g. for :echo)


vimpostor:
Similar to `thesaurusfunc`, we could have an option `msgfunc` that allows the user to define a custom function that displays messages that are printed with `:echo`, `:echom`, etc...
I think this would also be a good first step to make implementing #940 easier.

A usecase scenario might be that the user wants to see messages in their statusline, or they might want to show them in the corner via the popup api similar to [this plugin](https://github.com/rcarriga/nvim-notify).

[Neovim first came up with this idea](https://github.com/neovim/neovim/pull/16480), but they abandoned it because they want to use a more generic Lua implementation (which doesn't apply to Vim because it doesn't have external UI API).
I for one disagree with their reasoning, and think `msgfunc` would be a good idea.

So before I start writing patches, I wanted to get input whether you all (and especially Bram) agree that this is a good idea.

chrisbra:
I'd like the idea.

Shougo:
notify plugin for Vim is here.

https://github.com/mattn/vim-notification

vimpostor:
> I am the patch author.

Yes, thank you very much for your neovim patch. I am also a fan of your `cmdheight=0` work, I appreciate what you are doing!
I don't want to "steal" your idea, but from the looks of it, neovim rejected the idea of `msgfunc`...

I also assumed that you didn't want to port this to Vim, because you opened the patch in Neovim first whereas usually cross-codebase patches are opened in Vim first and then ported to Neovim.

If my assumption was wrong, and you are interested in adapting your patch to Vim yourself, then of course I will give you the right of way to post the patch!

brammool:
> Similar to `thesaurusfunc`, we could have an option `msgfunc` that
> allows the user to define a custom function that displays messages
> that are printed with `:echo`, `:echom`, etc...
> I think this would also be a good first step to make implementing #940
> easier.
> 
> An usecase scenario might be that the user wants to see messages in
> their statusline, or they might want to show them in the corner via
> the popup api similar to [this
> plugin](https://github.com/rcarriga/nvim-notify).
> 
> [Neovim first came up with this
> idea](https://github.com/neovim/neovim/pull/16480), but they abandoned
> it because they want to use a more generic Lua implementation (which
> doesn't apply to Vim because it doesn't have external UI API).
> I for one disagree with their reasoning, and think `msgfunc` would be
> a good idea.
> 
[...]

PR #10675, opened 2022-07-08
feat: set cmdheight=0

### LINKED ISSUE #940: Feature Request: add option for collapsing cmdline when empty.
I would like to be able to hide the cmdline when not in use:

screen shot 2016-07-21 at 1 28 38 pm-1

Shougo:
Fix #940 

It is the port of https://github.com/neovim/neovim/pull/16251, https://github.com/neovim/neovim/pull/18961, https://github.com/neovim/neovim/pull/19185.

It almost works, but more testers and help are needed.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1766,6 +1766,9 @@ A jump table for the options with a short description can be found at |Q_op|.
 	The value of this option is stored with the tab page, so that each tab
 	page can have a different value.
 
+	When 'cmdheight' is zero, it disables echo area and all outputs need
+	|hit-enter| prompt.
+
 						*'cmdwinheight'* *'cwh'*
 'cmdwinheight' 'cwh'	number	(default 7)
 			global
@@ -6446,9 +6449,11 @@ A jump table for the options with a short description can be found at |Q_op|.
 		45%	relative position in the file
 	If 'rulerformat' is set, it will determine the contents of the ruler.
 	Each window has its own ruler.  If a window has a status line, the
-	ruler is shown there.  Otherwise it is shown in the last line of the
-	screen.  If the statusline is given by 'statusline' (i.e. not empty),
-	this option takes precedence over 'ruler' and 'rulerformat'
+	ruler is shown there.  If a window doesn't have a status line and
+	'cmdheight' is 0, the ruler is not shown.  Otherwise it is shown in
+	the last line of the screen.  If the statusline is given by
+	'statusline' (i.e. not empty), this option takes precedence over
+	'ruler' and 'rulerformat'
 	If the number of characters displayed is different from the number of
 	bytes in the text (e.g., for a TAB or a multibyte character), both
 	the text column (byte number) and the screen column are shown,
@@ -7098,6 +7103,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 			|+cmdline_info| feature}
 	Show (partial) command in the last line of the screen.  Set this
 	option off if your terminal is slow.
+	The option is disabled if 'cmdheight' is 0.
 	In Visual mode the size of the selected area is shown:
 	- When selecting characters within a line, the number of characters.
 	  If the number of bytes is different it is also displayed: "2-6"
@@ -7147,6 +7153,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 	If in Insert, Replace or Visual mode put a message on the last line.
 	Use the 'M' flag in 'highlight' to set the type of highlighting for
 	this message.
+	The option is disabled if 'cmdheight' is 0.
 	When |XIM| may be used the message will include "XIM".  But this
 	doesn't mean XIM is really active, especially when 'imactivatekey' is
 	not set.
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -2506,7 +2506,8 @@ win_update(win_T *wp)
 	    // Past end of the window or end of the screen. Note that after
 	    // resizing wp->w_height may be end up too big. That's a problem
 	    // elsewhere, but prevent a crash here.
-	    if (row > wp->w_height || row + wp->w_winrow >= Rows)
+	    if (row > wp->w_height ||
+		    row + wp->w_winrow >= (p_ch > 0 ? Rows : Rows + 1))
 	    {
 		// we may need the size of that too long line later on
 		if (dollar_vcol == -1)
@@ -2560,7 +2561,7 @@ win_update(win_T *wp)
 
 	// Safety check: if any of the wl_size values is wrong we might go over
 	// the end of w_lines[].
-	if (idx >= Rows)
+	if (idx >= (p_ch > 0 ? Rows : Rows + 1))
 	    break;
     }
 
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -2947,24 +2947,26 @@ redraw_asap(int type)
     schar_T	*screenline2 = NULL;	// copy from ScreenLines2[]
 
     redraw_later(type);
-    if (msg_scrolled || (State != MODE_NORMAL && State != MODE_NORMAL_BUSY)
-								    || exiting)
+    if (msg_scrolled
+	    || (State != MODE_NORMAL && State != MODE_NORMAL_BUSY)
+	    || exiting
+	    || p_ch < 1)
 	return ret;
 
     // Allocate space to save the text displayed in the command line area.
     rows = screen_Rows - cmdline_row;
-    screenline = p_ch < 1 ? NULL : LALLOC_MULT(schar_T, rows * cols);
-    screenattr = p_ch < 1 ? NULL : LALLOC_MULT(sattr_T, rows * cols);
+    screenline = LALLOC_MULT(schar_T, rows * cols);
+    screenattr = LALLOC_MULT(sattr_T, rows * cols);
     if (screenline == NULL || screenattr == NULL)
 	ret = 2;
     if (enc_utf8)
     {
-	screenlineUC = p_ch < 1 ? NULL : LALLOC_MULT(u8char_T, rows * cols);
+	screenlineUC = LALLOC_MULT(u8char_T, rows * cols);
 	if (screenlineUC == NULL)
 	    ret = 2;
 	for (i = 0; i < p_mco; ++i)
 	{
-	    screenlineC[i] = p_ch < 1 ? NULL : LALLOC_MULT(u8char_T, rows * cols);
+	    screenlineC[i] = LALLOC_MULT(u8char_T, rows * cols);
 	    if (screenlineC[i] == NULL)
 		ret = 2;
 	}
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -649,8 +649,8 @@ win_redr_ruler(win_T *wp, int always, int ignore_pum)
     int		off = 0;
     int		width;
 
-    // If 'ruler' off or redrawing disabled, don't do anything
-    if (!p_ru)
+    // If 'ruler' off or messages area disabled, don't do anything
+    if (!p_ru || p_ch < 1)
 	return;
 
     /*
@@ -671,7 +671,7 @@ win_redr_ruler(win_T *wp, int always, int ignore_pum)
 	return;
 
 #ifdef FEAT_STL_OPT
-    if (*p_ruf)
+    if (*p_ruf && p_ch > 0)
     {
 	int	called_emsg_before = called_emsg;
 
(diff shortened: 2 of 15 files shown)


vimpostor:
Nice work, this is awesome and I hope it can be merged soon.

There is a slight problem and inconsistency with the neovim implementation though. If you enter the cmdline while the cursor is in the lower half of the window, the entire buffer content will shift one line up while the cmdline is open.

Minimal reproducable example:
```vim
src/vim --clean -Nu <(cat<<EOF
call append(1, range(0,105))
set cmdheight=0
norm 100j
echom "Buffer will shift up as soon as you enter cmdline with ":"!"
EOF
)
```

Shougo:
> There is a slight problem and inconsistency with this implementation though. If you enter the cmdline while the cursor is in the lower half of the window, the entire buffer content will shift one line up while the cmdline is open.

It is expected behavior.  Please see https://github.com/neovim/neovim/issues/19112.

vimpostor:
> It is expected behavior. Please see [neovim/neovim#19112](https://github.com/neovim/neovim/issues/19112).

It's ok to fix this problem later, but I wouldn't call it expected behavior. If the buffer content moves up while I am in the lower half of the window and doesn't move up just because I happen to move my cursor a few lines up, then I would certainly not call that  behavior expected for any reasonable definition of "expected".

But of course that small problem is not a blocker right now and for now it is more important to get Bram's opinion on the entirety of this patch.
If you don't want to fix this problem, I can fix it myself in a later patch.
Right now it's more important to get this to a stable state, and I already got a segmentation fault while testing this patch, but I haven't been able to reliably reproduce it yet.

Shougo:
> If you don't want to fix this problem, I can fix it myself in a later patch.

I don't fix it, but you can fix it of course.

> Right now it's more important to get this to a stable state, and I already got a segmentation fault while testing this patch, but I haven't been able to reliably reproduce it yet.

I need the reproduce ways.

vimpostor:
> I need the reproduce ways.

The following triggers a deadly `SIGABRT`, if vim is started in a terminal window that has `56` or more rows:
```vim
src/vim --clean -Nu <(cat<<EOF
call append(1, range(0,105))
set cmdheight=0
call feedkeys("Ggg")
call feedkeys(":set lines=5\<CR>")
EOF
)
```

As soon as you resize the terminal window (this is simulated above using `:set lines=5`), vim crashes.
Note that you need the `feedkeys()`, the crash does not happen with `:norm`.

Shougo:
I don't reproduce the SEGV, but lines seems broken.

> Nice work, this is awesome and I hope it can be merged soon.

If you need the merge, please help the bug fix.
The fix is not easy.

Shougo:
I have fixed the behavior.
`cmdheight` must be positive if `lines` is smaller.

Shougo:
I think the tests are almost covered...

vimpostor:
> I have fixed the behavior. `cmdheight` must be positive if `lines` is smaller.

The `:set lines` call was just there to simulate the window resize. The crash still happens if you resize the window with the mouse instead of using `:set lines`. Also you can't just "fix" a problem by disabling the way to reproduce the problem. The problem is obviously more deeply rooted, as we are dealing with a memory corruption.
We need to fix the real problem and then we can remove your workaround to set `ch` back to `1` too.

Besides, vim still crashes after your supposed "fix", now we have a double free:

```c
Thread 1 "vim" received signal SIGABRT, Aborted.
0x00007ffff671b36c in ?? () from /usr/lib/libc.so.6
(gdb) backtrace
#0  0x00007ffff671b36c in  () at /usr/lib/libc.so.6
#1  0x00007ffff66cb838 in raise () at /usr/lib/libc.so.6
#2  0x00007ffff66b5535 in abort () at /usr/lib/libc.so.6
#3  0x00007ffff670f45e in  () at /usr/lib/libc.so.6
#4  0x00007ffff67250cc in  () at /usr/lib/libc.so.6
#5  0x00007ffff672725c in  () at /usr/lib/libc.so.6
#6  0x00007ffff67298f3 in free () at /usr/lib/libc.so.6
#7  0x00005555555fb7e5 in vim_free (x=<optimized out>) at alloc.c:625
#8  0x0000555555832b7a in win_free_lsize (wp=wp@entry=0x5555559abb90) at window.c:5445
#9  0x0000555555782770 in screenalloc (doclear=doclear@entry=0) at screen.c:2689
#10 0x0000555555784b79 in screenclear () at screen.c:2979
#11 0x00005555557d02c3 in set_shellsize (width=118, height=5, mustset=mustset@entry=1) at term.c:3521
#12 0x00005555557d0663 in set_shellsize (width=<optimized out>, height=<optimized out>, mustset=mustset@entry=1)
    at term.c:3487
#13 0x000055555571fb08 in set_num_option
    (opt_idx=opt_idx@entry=188, varp=varp@entry=0x5555559737e0 <Rows> "\005", value=5, errbuf=errbuf@entry=0x7fffffffcd60 "", errbuflen=errbuflen@entry=80, opt_flags=opt_flags@entry=0) at option.c:3718
#14 0x0000555555721d52 in do_set (arg_start=0x555555a2a6f4 "lines=5", opt_flags=0) at option.c:1699
#15 0x0000555555681ceb in do_one_cmd
     (cookie=0x0, fgetline=0x55555568e990 <getexline>, cstack=0x7fffffffd140, flags=0, cmdlinep=0x7fffffffcef0) at ex_docmd.c:2570
#16 do_cmdline (cmdline=cmdline@entry=0x0, fgetline=0x55555568e990 <getexline>, cookie=cookie@entry=0x0, flags=0)
    at ex_docmd.c:992
#17 0x0000555555704c96 in nv_colon (cap=0x7fffffffd810) at normal.c:3203
#18 0x0000555555709ce1 in normal_cmd (oap=oap@entry=0x7fffffffd8d0, toplevel=toplevel@entry=1) at normal.c:939
#19 0x000055555589777a in main_loop (cmdwin=cmdwin@entry=0, noexmode=noexmode@entry=0) at main.c:1527
#20 0x0000555555898a13 in vim_main2 () at main.c:885
#21 0x00007ffff66b6290 in  () at /usr/lib/libc.so.6
#22 0x00007ffff66b634a in __libc_start_main () at /usr/lib/libc.so.6
#23 0x00005555555fb2e5 in _start () at ../sysdeps/x86_64/start.S:115
```

I can try debugging the problem later, if you can't reproduce the SIGABRT.

Shougo:
Oh, no.  The window behavior is broken when `cmdheight=0`.
Because Vim doesn't support `cmdheight=0` behavior.
It is very difficult to fix.

> I can try debugging the problem later, if you can't reproduce the SIGABRT.

Please help me.  The window height seems broken.

Shougo:
If the statusline exists, it seems OK.  But statusline does not exists, it is broken.

Shougo:
I think `screenalloc()` needs to be fixed.

Shougo:
@vimpostor I have fixed the errors.  Please check the changes.

vimpostor:
> @vimpostor I have fixed the errors. Please check the changes.

Not fixed for me, my [reproduce instructions from above](https://github.com/vim/vim/pull/10675#issuecomment-1181665395) still crash vim, if vim is started in a terminal with more than 60 rows.

Shougo:
Really?  I cannot reproduce the error.

> It's not easy to debug, because I think the memory corruption happens somewhere else long before the crash.

I cannot debug this.  Because it is not reproduced for me.

Shougo:
Please test it for more testers.
If not, it cannot be merged.

brammool:
Have some users tried out this patch?  Any problems?

Shougo:
In neovim, many people have reviewed and tried.  But in Vim it is not yet...

https://github.com/vim/vim/issues/940

The feature is not desired??

Shougo:
> Have some users tried out this patch? Any problems?

One problem is reported.  I don't reproduce the problem yet.
I cannot find other problems.

Can you review the changes?

Shougo:
I have tested it with valgrind.

The leak is reported, but it is same with HEAD.

brammool:
> In neovim, many people have reviewed and tried.  But in Vim it is not yet...
> 
> https://github.com/vim/vim/issues/940
> 
> The feature is not desired??

Vim users and NeoVim users are different.  We have seen that a feature
added to NeoVim was appreciated there, but once added to Vim there are
comments.  The more users, the more opinions and the more different
expectations.

I'll look at the patch once I have time, but I currently have long todo
list.

-- 
125. You begin to wonder how often it REALLY is necessary to get up
     and shower or bathe.

Shougo:
> Vim users and NeoVim users are different.  We have seen that a feature
added to NeoVim was appreciated there, but once added to Vim there are
comments. 

Yes.  I know this.

> I'll look at the patch once I have time, but I currently have long todo
list.

Thanks.

brammool:
One thing that I notice is that after using the command line, the screen is redrawn with CLEAR.  It's predictable that some users are going to complain about flicker...
Anyway, it looks like this change isn't too big, and users who don't like it can simply not used it.
And I'll assume that when any problems are reported, they will be fixed.

bfrg:
Just tried it out. I find it a little bit annoying that every time I press `:`, the window is scrolled up one line to make room for the command-line. Maybe this will be fixed, once https://github.com/vim/vim/pull/10682 gets merged?

bfrg:
> I find it a little bit annoying that every time I press :, the window is scrolled up one line to make room for the command-line. Maybe this will be fixed, once https://github.com/vim/vim/pull/10682 gets merged?

Okay, this seems to happen only when the cursor is in the lower half of the window.

vimpostor:
> > I find it a little bit annoying that every time I press :, the window is scrolled up one line to make room for the command-line. Maybe this will be fixed, once #10682 gets merged?
> 
> Okay, this seems to happen only when the cursor is in the lower half of the window.

Yes, I already reported that here: https://github.com/vim/vim/pull/10675#issuecomment-1179765521
But it got ignored, or called "expected behaviour"

brammool:
> > I find it a little bit annoying that every time I press :, the
> > window is scrolled up one line to make room for the command-line.
> > Maybe this will be fixed, once https://github.com/vim/vim/pull/10682
> > gets merged?
> 
> Okay, this seems to happen only when the cursor is in the lower half
> of the window.

This is because of "scroll to fraction".  I don't think we need to rely
on that option, let me try to avoid it.  Oh, also have to reset
'scrolloff'.

-- 
199. You read this entire list of symptoms, looking for something
     that doesn't describe you.

h-east:
@brammool 

Undo/Redo messages are not displayed.

- How tp repro.
`$ vim -Nu NONE +"set cmdheight=0" +"norm iHello"` after type `u`

- Expected behavior
image

- Actual behavior
image

---
P.S
I would never use `set cmdheight=0`.

Shougo:
@h-east The messages are skipped by `messaging()`.

```c
    int
messaging(void)
{
    return (!(p_lz && char_avail() && !KeyTyped)) && p_ch > 0;
}
```

If you remove `p_ch` check, the messages are displayed.
But it is not simple problem.  The messages are disappeared instantly.

The root cause is the command line is used as messaging area.
msgfunc feature is needed for it.
https://github.com/vim/vim/issues/10571

h-east:
@Shougo 
In the first place, increasing a buffer line instead of suppressing the command line, which is also used as a message area, is too much of a disadvantage.

This problem was foreseen because you adopted "Press Enter" to switch from the command line to the buffer line.
On the other hand, it is also subtle to make the transition from the command line to the buffer line after a certain amount of time.

In other words, this patch's specification is distorted.
A message area is always needed.

I hope Bram reverts this patch.

Shougo:
`set cmdheight=0` patch means user can choose hide the message area or not.
Yes, the feature is not perfect and it has side effects like you have pointed out problems.

I think the message area should be splitted out from command line.
But it should be another PR.

brammool:
I think that adding a "msgfunc" functionality basically means the problem is handed over to some plugin, that you can keep tuning forever.
For messages that are informative, where pressing Enter would be very annoying, showing them briefly (couple of seconds) on top of the window, is probably the best solution.  It could be done with a popup window, and it could be placed at the bottom of the current window.  In case the user didn't manage to read it, ":mes" should show it again.
This requires work, to separate messages in two categories: one that can show briefly and one that does require pressing Enter (e.g. error messages).  And handling longer messages (wrapping, truncation).  And adding all these messages to the message history.

Shougo:
> This requires work, to separate messages in two categories: one that can show briefly and one that does require pressing Enter (e.g. error messages). And handling longer messages (wrapping, truncation). And adding all these messages to the message history.

The builtin message popup feature is better for me.

Shane-XB-Qian:
pls just keep if kept cmdheight=1 then everything would not be impacted, then would be fine...
// some plugins e.g show symbol info at cmdline (by echo), then definitily wish cmdheight=1 kept working always.. :-)
// plus some plugins req enter something e.g passwd

Shane-XB-Qian:
vs if cmdheight=0, then no info output, 
unless some setup like a popup at top or bottom, that is something like gnu-screen currently looks-like, its msg showing there.
// well, it required some extra setup to make it., but so far, looks not direct way to make it. right?
// some user perhaps not like so, if so.

Shougo:
> // well, it required some extra setup to make it., but so far, looks not direct way to make it. right?

Unfortunately no direct way for it...
We need to define new message UI.
`set cmdheight=0` is not perfect solution.  It just provide the start line for UI.

h-east:
At this stage, it should be noted in the documentation of `set cmdheight=0` that there are restrictions (*1).

*1
Some messages are not displayed, including:
- undo/redo message
- 'redrawtime' exceeded, syntax highlighting disabled

I03D:
I need this feature.

Shougo:
The feature is reverted.
Please see other issues.

PR #10676, opened 2022-07-08
Correct translation item

adaext:
Correct translation items in `shellmenu.vim`


brammool:
I'll include this, thanks.

PR #10677, opened 2022-07-08
Unify menu items to include shortcut key always

adaext:
According to Vim official document, that characters like "&" need to be included for translated names to be found.


brammool:
@jamessan

runtime/ftplugin/debchangelog.vim
```diff
@@ -182,8 +182,8 @@ function <SID>FinaliseMenu()
     amenu enable Changelog.New\ Version
```
jamessan: @adaext Was there a reason this wasn't changed to `&New\ Version`?

runtime/ftplugin/debchangelog.vim
```diff
@@ -203,17 +203,17 @@ function <SID>MakeMenu()
     menu Changelog.-sep-				<nul>
 
     amenu Changelog.Set\ &Distribution.&unstable	:call Distribution("unstable")<CR>
```
jamessan: @adaext Since `&Changelog` is used on line 200, should it be used for all the menu definitions and enable/disable commands?

jamessan (commented):
From my understanding of the very brief mention in the documentation, the general idea behind this PR is that the menupath needs to consistently use `&` anytime a menupath is defined with one.  However, the PR doesn't do that.

Was the PR just trying to show the pattern of what changes need to be done, or was it intended to be complete, @adaext?

runtime/ftplugin/debchangelog.vim
adaext: > @adaext Since `&Changelog` is used on line 200, should it be used for all the menu definitions and enable/disable commands?

Thanks, you are right. I've fixed it in the new commit.

runtime/ftplugin/debchangelog.vim
adaext: > @adaext Was there a reason this wasn't changed to `&New\ Version`?

Thanks, I've also fixed it.

jamessan:
I've sent the updated file to Bram.

adaext:
&nbsp;Thank you very much!
&nbsp;
------------------&nbsp;Original&nbsp;------------------
From: ***@***.***&gt;;
Date: &nbsp;Tue, Jul 26, 2022 10:30 AM
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; ***@***.***&gt;; 
Subject: &nbsp;Re: [vim/vim] Unify menu items to include shortcut key always (PR #10677)

&nbsp;



 
I've sent the updated file to Bram.

brammool:
James has sent me updated files with these changes merged in

adaext:
> James has sent me updated files with these changes merged in

Ok, thank you very much.

PR #10678, opened 2022-07-08
Overlay "whitespace" syntax over other syntax correctly

adaext:
The current implementation replaces the whole syntax with only "whitespace", instead of overlaying it over existing one.


brammool:
I'll include it.  I think we can add a check for whether ".whitespace" is already included, but as you mention, it's unlikely the user would do this twice.

PR #10681, opened 2022-07-09
Change EndOfBuffer color for desert.vim

EmptiedSoul:


brammool:
This should most likely go to https://github.com/vim/colorschemes

chrisbra:
ping @romainl @habamax

habamax:
I don't like endofbuffer having different color...

Having said that, legacy desert has the same `EndOfBuffer` (linked to `NonText`)

New Desert has a different `NonText` which some thinks is better than original.

legacy
image

new
image

habamax:
So if the fix is needed, it should be done by changing `NonText`

romainl:
@EmptiedSoul, the remade colorschemes are all generated from templates so it is best to keep eventual changes in the templates.

Could you open an issue in https://github.com/vim/colorschemes/issues so that we work this out over there?

chrisbra:
in that case closing it here then.

PR #10682, opened 2022-07-09
Add new option splitscroll to avoid scrolling

luukvbaal:
This PR originates from https://github.com/neovim/neovim/pull/19243.

Adds a new option to avoid scrolling when opening/closing splits. Perceived by users of the [plugin](https://github.com/luukvbaal/stabilize.nvim) mentioned in the original PR as less visually jarring. Especially so when using the quickfix window for automatic linting(opening and closing automatically as a buffers' linting errors are resolved).


brammool:
> This PR originates from https://github.com/neovim/neovim/pull/19243.
> 
> Adds a new option to avoid scrolling when opening/closing splits.
> Perceived by users of the
> [plugin](https://github.com/luukvbaal/stabilize.nvim) mentioned in the
> original PR as less visually jarring. Especially so when using the
> quickfix window for automatic linting(opening and closing
> automatically as a buffers&#39; linting errors are resolved).

I find it hard to understand the explanation of the option.
It probably helps to separately explain what happens when splitting and
when closing a window.

It is also about terminology: what is "the current scroll position"?
Is that what line the first line of the window is displaying?

Centering the cursor in the window does not mean "keep centered".  What
if the cursor was not centered to start with?

If we can agree on what this does, it will need a test.

-- 
From "know your smileys":
 y:-)	Bad toupee

luukvbaal:
Thanks for considering the option. I'll try to improve the wording in the docs and look into making a test. 

After some discussion under the neovim devs the current proposal is to make the option non-boolean and have the following options:

1. scroll to maintain relative position (current behavior)
2. don't scroll except as needed to keep cursor row visible (and subject to 'scrolloff')
3. never scroll, even if cursor row will be hidden or 'scrolloff' would be violated

luukvbaal:
I was thinking tests could go in `testdir/test_window_cmd.vim` looking something like:
```vim
func Test_winscroll_with_splits()
  set splitbelow
  set scrolloff=0
  set splitscroll=1
  call setline(1, range(1,256))
  norm Mk "move cursor to line above where new split will begin
  wincmd s
  wincmd p
  call assert_equal(1, line("w0"))
  wincmd q

  similar sequence when cursor is behind new split and for splitscroll={0,2} and scrolloff > 0
  %bw!
endfunc
```

Does that seem right?

src/testdir/test_window_cmd.vim
```diff
@@ -1557,4 +1557,47 @@ func Test_window_alloc_failure()
...
+
+  set scrolloff=0
+  set splitscroll=2
+  norm ggL
+  split
+  quit
+  call assert_equal(winheight(0), line("."))
+  split
+  wincmd p
+  call assert_equal(winheight(0), line("."))
```
luukvbaal: I made the option non-boolean in the latest commit as per https://github.com/neovim/neovim/pull/19243#issuecomment-1179607771. I also added some tests but I'm running into an issue.

I cant get this test to pass even though manually running the commands yields the expected results. I expect this has something to do with https://github.com/vim/vim/blob/34be31ebaa1ed3f139928d0f7a1a6d1993621e31/src/window.c#L4943-L4945
Where we move the cursor to the last visible line when moving back to the original window in the case where `'splitscroll' == 2` and the cursor was behind the newly opened split. I suspect this is not the correct way to set the cursor position and is what causes the assertion to fail when run inside a VimL function.

I suspected some function from `move.c` should be called to update/invalidate but I didn't find it yet. What would be the correct way to update the cursor, or am I on the wrong track?

luukvbaal: I figured out `w_botline` is not the expected value here when running the test, hence the cursor is not set and the test fails. Is `w_botline` being outdated expected when running VimL functions?

Exchanging `w_botline` for `w_topline + w_height` resolves the issue and makes the test pass. Would this be a reasonable workaround or should I be calling some sort of function from `move.c` to update `w_botline`?

brammool:
> I figured out `w_botline` is not the expected value here when running
> the test, hence the cursor is not set and the test fails. Is
> `w_botline` being outdated expected when running VimL functions?
> 
> Exchanging `w_botline` for `w_topline + w_height` resolves the issue
> and makes the test pass. Would this be a reasonable workaround or
> should I be calling some sort of function from `move.c` to update
> `w_botline`?

No, when lines wrap then w_topline + w_height can be much higher than
the actual w_botline.

If it is really only for testing, a ":redraw" should work.
Otherwise you can use validate_botline().  This is an expensive function
though.

-- 
From "know your smileys":
 =):-)	Uncle Sam

luukvbaal:
> No, when lines wrap then w_topline + w_height can be much higher than the actual w_botline.

Right, can confirm this change yielded the wrong outcome when wrapped lines are present.

> If it is really only for testing, a ":redraw" should work.

I think it is, and indeed `redraw` makes the test pass.

src/window.c
```diff
@@ -6379,9 +6379,10 @@ win_new_height(win_T *wp, int height)
 	    if (curwin->w_cursor.lnum
 		    > (curwin->w_botline - 1 - get_scrolloff_value()))
 	    {
-	        curwin->w_fraction = (height - get_scrolloff_value() - 0.0) / height
-								    * FRACTION_MULT;
+	        curwin->w_fraction = (float)(height - get_scrolloff_value())
+						     / height * FRACTION_MULT;
 	        scroll_to_fraction(wp, prev_height);
+		curwin->w_valid &= ~VALID_TOPLINE;
```
luukvbaal: Without this invalidation, it was possible to scroll to invalid cursor positions disobeying `'scrolloff'`. I guess a consequence of setting `w_fraction` manually.

luukvbaal:
I think the latest commit is a valid implementation of the proposed option. Requesting review I guess.

brammool:
Who has tried out this patch?  Is it useful and does it work properly?

nullchilly:
I have tried it out, `set splitscroll=2` helps stabilize your buffer content on window open/close event (In my case terminal toggling)

`set splitscroll=1` also works correctly

Video preview:

https://user-images.githubusercontent.com/56817415/180603725-8fe33f94-150e-4bff-b251-83c0a64c6f92.mp4

bfrg:
Instead of numbers, wouldn't it be better to use strings for the `splitscroll` option? It's hard to remember what which number does.

bfrg:
I have tried it out and it seems to work as intentional. Until now I had my own small script that does the same thing using autocommands with `winsaveview()` and `winrestview()`. Having a built-in option would be nice.

@luukvbaal  There's one issue though. With `set splitscroll=2` the buffer is still scrolled when the terminal window is resized such that it would hide the cursor. If there are multiple vertical split windows, only the currently focused window is scrolled.

bfrg:
It should probably also be mentioned somewhere that this option only works when `splitbelow` is set.

luukvbaal:
> Instead of numbers, wouldn't it be better to use strings for the splitscroll option?

Numbers seems to be the convention for similar options. See e.g. `'conceallevel'` and `'laststatus'`.

> If there are multiple vertical split windows, only the currently focused window is scrolled.

I'm not sure this is an issue, the option claims to determine the scroll behavior when splitting, not when resizing. And when `set splitscroll=2`, to delay setting the cursor position setting the cursor to a valid position(`botline`) until you again focus the window. If you are actively resizing the splits, the currently active window will thus keep its cursor on botline(and scroll).
Would you expect all windows to scroll, or no windows to scroll including the currently focused window?

> It should probably also be mentioned somewhere that this option only works when `splitbelow` is set.

Yeah fair. Or to improve the option to also "work" with `'nosplitbelow'`(as in stabilize the window content, it currently just doesn't consider the option but you're right that different logic would be required to stabilize buffer content with `'nosplitbelow'`).

luukvbaal:
Latest commit handles `nosplitbelow` where feasible. Stabilizing is possible when the cursor position is not hidden behind lower window but doesn't make sense when it would be.

brammool:
> > Instead of numbers, wouldn't it be better to use strings for the
> > splitscroll option?
> 
> Numbers seems to be the convention for similar options. See e.g.
> `'conceallevel'` and `'laststatus'`.

Those options were added long ago.  These days we don't care so much for
the extra memory use, and care more for easy of use (since there are so
many options now).  If we can come up with good, meaningful names that
would be preferred.

> > If there are multiple vertical split windows, only the currently
> > focused window is scrolled.
> 
> I'm not sure this is an issue, the option claims to determine the
> scroll behavior when splitting, not when resizing. And when `set
> splitscroll=2`, to delay setting the cursor position setting the
> cursor to a valid position(`botline`) until you again focus the
> window. If you are actively resizing the splits, the currently active
> window will thus keep its cursor on botline(and scroll).
> Would you expect all windows to scroll, or no windows to scroll
> including the currently focused window?
> 
> > It should probably also be mentioned somewhere that this option only
> > works when `splitbelow` is set.
> 
> Yeah fair. Or to improve the option to also "work" with
> `'nosplitbelow'`(as in stabilize the window content, it currently just
> doesn't consider the option but you're right that different logic
> would be required to stabilize buffer content with `'nosplitbelow'`).

So when the 'splitscroll' option is non-zero (active) then 'splitbelow'
would be ignored?  Or the other way around?  That doesn't make sense.
If the new window is above the current window, it could still do the
same thing, right?  It's only a matter of time until users ask for this.
Ah, it appears you already took care of it after this message.

-- 

luukvbaal:
> Those options were added long ago. These days we don't care so much for the extra memory use, and care more for easy of use (since there are so many options now). If we can come up with good, meaningful names that would be preferred.

I see, I guess I missed options like `'ambiwidth'`: https://github.com/vim/vim/blob/cb398d1a5b189261b32658e180d13081761b8d24/runtime/doc/options.txt#L743-L754?

Something like `'splitscroll'` with these values then?
* "relative": keep same relative cursor position(current default)
* "stable": do not scroll if cursor is still visible
* "force-stable": never scroll and set new cursor position when re-focusing window

> So when the 'splitscroll' option is non-zero (active) then 'splitbelow' would be ignored? Or the other way around? That doesn't make sense. If the new window is above the current window, it could still do the same thing, right? It's only a matter of time until users ask for this. Ah, it appears you already took care of it after this message.

No nothing was ignored in earlier commits, window position would still honor `'splitbelow'` but the logic required some mirroring to make sense for `'splitbelow'` vs `'nosplitbelow'`.

brammool:
> > Those options were added long ago. These days we don't care so much
> > for the extra memory use, and care more for easy of use (since there
> > are so many options now). If we can come up with good, meaningful
> > names that would be preferred.
> 
> I see, I guess I missed options like `'ambiwidth'`: https://github.com/vim/vim/blob/cb398d1a5b189261b32658e180d13081761b8d24/runtime/doc/options.txt#L743-L754?
> 
> Something like `'splitscroll'` with these values then?
> * "relative": keep same relative cursor position(current default)
> * "stable": do not scroll if cursor is still visible
> * "force-stable": never scroll and set new cursor position when re-focusing window

Reading the explanation gives a hint about what to call them:
how about: "normal", "noscroll" and "neverscroll"?
"stable" doesn't really give a clue.

-- 
"New York, the city that never sleeps".  "There is so much noise?"

bfrg:
>  Would you expect all windows to scroll in this case, or no windows at all, including the currently focused window?

I think the current behavior is inconsistent. The inactive windows are not scrolled but the currently focused one is. I would have expected all windows to stay "stable" when the Vim terminal window is resized.

And there's another inconsistency. Let's say you have two vertical splits and then you run `:botright copen`. While in the quickfix window, run `:resize +5`. You can run the command multiple times and none of the vertical splits will be scrolled (assuming `set splitscroll=2`). I would have assumed the same behavior when the entire screen is resized (think of a quick `tmux split-window`, or quickly opening a new terminal window with a tiling window manager).

bfrg:
@luukvbaal There's another case when the currently focused window is scrolled but the other split windows are kept stable.

Open two vertical splits, place the cursor at the bottom in each window. Then open the command-line window with `q:`, and the close it with `:q`. When the command-line window is opened, the windows are not scrolled, but when it's closed, the last focused window is scrolled.

luukvbaal:
> Reading the explanation gives a hint about what to call them: how about: "normal", "noscroll" and "neverscroll"?


Yeah sure, sounds good to me. I have a commit ready implementing the change to stringopt. 

Sorry for not getting back to some of the replies here. I think the option implementation requires some further consideration. It works for the simple case splitting a single window into two but I'm not sure isolating the change to `win_new_height()` is sufficient. I think we might require some of the logic in `win_split_ins()` handling all the edge cases. Will have to look at it again. Consider draft for now.

luukvbaal:
I reworked the option to truly obey the description in all cases (I think). Any and all opened splits will maintain their current scroll position regardless of how they were opened (`topleft`/`belowright`/`(no)splitbelow` etc.

In the process I reverted back to a boolean option and got rid of the non-forcing `noscroll` option, only keeping the (now more comprehensive) `neverscroll`. Always setting the cursor position instead of scrolling. I'm not sure adding back the `noscroll` option will be worth the hassle/code.

Another round of testing would be appreciated. Tests still need updating.

luukvbaal:
[asciicast](https://asciinema.org/a/DCMQin2PcjdXTj7zTsOo81dEc)

luukvbaal:
tabline/winbar could use handling as well, not sure if feasible or worth it.

luukvbaal:
RE the failing tests: right, the `topfrp` argument is a remnant of when I tried to reduce the handling of unchanged windows by only passing the frame where the window heights have changed. Couldn't get it working right away properly, not sure if it's still possible but maybe the current check is sufficient for performance considerations and we can get rid of the frame arg:

https://github.com/vim/vim/blob/1aea71e8c3800409fbce66850a6bfd60c1bf97f0/src/window.c#L6371-L6373

luukvbaal:
I wasn't sure if it's always possible to know the frame(s) that contain changed windows and pass it to the scroll correct function from where it's called. I think looping over all the windows in  to be sure is fine and unnecessary work is avoided by the `prev_height` check.

src/window.c
```diff
@@ -1327,6 +1333,11 @@ win_split_ins(
 	msg_col = 0;	// put position back at start of line
     }
 
+    // If we have there is a winbar, win_equal invalidates w_botline
+    // before we get to spsc_correct_scroll (why only with winbar?).
+    if (!p_spsc && WINBAR_HEIGHT(curwin))
```
luukvbaal: @brammool with the latest commit we also avoid scrolling in all cases when a winbar is present. Without this check `w_botline` was invalidated due to some recursive calls to `win_new_width()` in `equal_win()`, before we get to `spsc_correct_scroll()`. Resulting in scrolling in `curwin`(actually `prevwin` after `win_enter_ext(wp)`.

Would a check like this be okay? It's not really checking against the root cause of what causes scrolling but it seems to work and I don't see any side effects.

brammool:
> @luukvbaal commented on this pull request.
> 
> > @@ -1327,6 +1333,11 @@ win_split_ins(
>  	msg_col = 0;	// put position back at start of line
>      }
>  
> +    // If we have there is a winbar, win_equal invalidates w_botline
> +    // before we get to spsc_correct_scroll (why only with winbar?).
> +    if (!p_spsc && WINBAR_HEIGHT(curwin))
> 
> @brammool with the latest commit we also avoid scrolling in all cases when a winbar is present. Without this check `w_botline` was invalidated due to some recursive calls to `win_new_width()` in `equal_win()`, before we get to `spsc_correct_scroll()`. Resulting in scrolling in `curwin`(actually `prevwin` after `win_enter_ext(wp)`.
> 
> Would a check like this be okay? It's not really checking against the root cause of what causes scrolling but it seems to work and I don't see any side effects.

I'm not quite sure what your strategy is.  I thought that when
'splitscroll' is off then in every window w_topline remains unchanged.
If the window heigh changes, this means w_botline will become invalid.
Then the cursor may have to be moved to keep it in the visible text.
w_botline will be updated automatically, although you may want to do it
before deciding where to put the cursor.
I have no idea why the window bar matters in any of this.

-- 
       Bravely bold Sir Robin, rode forth from Camelot,
       He was not afraid to die, Oh Brave Sir Robin,
       He was not at all afraid to be killed in nasty ways
       Brave, brave, brave, brave Sir Robin.

luukvbaal:
The strategy is to keep all lines on the same actual terminal line, updating the cursor to a valid position rather then scrolling when necessary. To do this, indeed `topline` remains unchanged a window is reduced in size from below(`botline` shifts upward). 

However, when a window is reduced in size from above(i.e. when `winrow` != `prev_winrow`), to keep the window content "stable", we have to update `topline` instead. To do this, I am setting the cursor line to where the new `botline` would have been, followed by a `scroll_to_fraction()` with fraction set to `FRACTION_MULT`, such that this cursor position will be the new `botline`.

This `botline` value depends on the value of `laststatus`, the number of windows in a `frame/tabpage`, and the `winbar` height. I think I covered all possible cases now but I'm not quite sure why the recursive calls to `win_new_width()` only happen when a winbar is open. It feels like I'm working around the problem but if it works it works.

brammool:
> The strategy is to keep all lines on the same actual terminal line,
> updating the cursor to a valid position rather then scrolling when
> necessary. To do this, indeed `topline` remains unchanged a window is
> reduced in size from below(`botline` shifts upward). 
> 
> However, when a window is reduced in size from above(i.e. when
> `winrow` != `prev_winrow`), to keep the window content "stable", we
> have to update `topline` instead. To do this, I am setting the cursor
> line to where the new `botline` would have been, followed by a
> `scroll_to_fraction()` with fraction set to `FRACTION_MULT`, such that
> this cursor position will be the new `botline`.
> 
> This `botline` value depends on the value of `laststatus`, the number
> of windows in a `frame/tabpage`, and the `winbar` height. I think I
> covered all possible cases now but I'm not quite sure why the
> recursive calls to `win_new_width()` only happen when a winbar is
> open. It feels like I'm working around the problem but if it works it
> works.

Hmm, I think what you are trying to do is not 100% possible.
The topline can only be the start of a buffer line, thus when this line
wraps it occupies two screen lines.  If the window moves down one screen
line, then either you keep topline, in which case the window contents
scrolls down one screen line, or you increment topline, in which case
the window contents scrolls up one screen line.  See what I mean?

There is a todo item somewhere to have the first line possible start
halfway the line.  That would have to be implemented first.  The reason
it's not done yet is that it will require a lot of work to get right.

-- 
Error:015 - Unable to exit Windows.  Try the door.

luukvbaal:
> Hmm, I think what you are trying to do is not 100% possible.
The topline can only be the start of a buffer line, thus when this line wraps it occupies two screen lines. 

Yeah true, I noticed this during testing. I saw it as an acceptable loss for now, I use `nowrap` my self and it will still lead to less scrolling, even when lines are wrapped.

> There is a todo item somewhere to have the first line possible start halfway the line.  That would have to be implemented first.  The reason it's not done yet is that it will require a lot of work to get right.

I can imagine this will be tricky.

luukvbaal:
@barmmool How would I go about adding topfill lines to a window? I'm trying to stabilize one (final hopefully) case where the curwin has a winbar and is split above. The winbar will move to the with the window to the lower split, leaving more available toplines for the upper split.
I can account for this by scrolling up, but when `w_topline < WINBAR_HEIGHT(curwin)`, topfill lines would have to be added instead.

luukvbaal:
See https://asciinema.org/a/zKuzBLQttMYGsjvnx7Yi6DTwL for visualization.

brammool:
> @barmmool How would I go about adding topfill lines to a window? I'm trying to stabilize one (final hopefully) case where the curwin has a winbar and is split above. The winbar will move to the with the window to the lower split, leaving more available toplines for the upper split.
> I can account for this by scrolling up, but when `w_topline < WINBAR_HEIGHT(curwin)`, topfill lines would have to be added instead.

filler lines are only for when using diffmode when two windows are to be
kept in sync.  Their use is very complicated, using them also in another
situation is not a good idea.
Implementing the other solution, having the first line start with some
offset, is much more useful.  Also complicated, but there is a reason to
spend time on it.


-- 
Facepalm statement #1: "I'm going to New York tomorrow, hopefully I have time
to visit the White House"

luukvbaal:
I see, thanks. I think this is it for now then. Still waiting for some feedback on whether the `normal/noscroll/neverscroll` option is deemed necessary, also over at the neovim PR.

luukvbaal:
Not sure why the added test fails on appveyor with gvim, can't seem to replicate it failing.

bfrg:
Is it possible to add the old cursor position to the jump list before it is changed? Does it even make sense?

Just a thought: imagine, I carefully placed the cursor somewhere in the buffer, then open a new split, and it happens that the cursor is moved up to avoid scrolling the window. If I focus the previous window before I opened the split, the cursor will be on a different line. Someone might need the old position, in that case my first thought would be to hit `<C-o>` to jump back to the previous buffer position. But this doesn't work at the moment.

I tested the latest version. There are still some cases when the window is scrolled. 

Open two vertical splits, place the cursor at the bottom in each window. Then open the command-line window with `q:`. When the command-line window is opened, the last window is scrolled.

With `set nosplitscroll`, when a vertical split is opened with `<c-w>v`, the cursor is not at the same position in the split. This is unexpected. Why is the cursor moved in the new vertical split? I think it should stay at the old position, since everything is copied from the old window.

This is illustrated in the following video. Before I open the vertical split, the cursor is at the very bottom of the buffer on the word "Atari". By default the old cursor position is copied when a new split is opened. But when I hit `<c-w>v`, the cursor is moved 5 lines above (below the line "OS/2 support...").

https://user-images.githubusercontent.com/6266600/183931234-cd4e3197-aee8-4e5d-a3d3-a305cd6b410a.mp4

It looks like that for vertical splits, it is never possible to keep the cursor at the very bottom of the window. Open two vertical splits, place the cursor at the bottom and then hit `<c-w>p`. The cursor is not at the old position anymore. Even if no windows have been opened.

**Note:** for consistency, there should probably also be help tags for `nosplitscroll` and `nospsc`.

luukvbaal:
> Is it possible to add the old cursor position to the jump list before it is changed? Does it even make sense?

Yeah my plugin did this as well. It does make sense I guess although I can't say I ever used it. Would have to figure out how to populate the jumplist.

Thanks for testing, I'll look into the problems you mentioned. You're right that vertical splits should be skipped altogether. I had this in previous versions but got lost somewhere along the way. Although I wonder if it is always enough to know that a vertical split was spawned. I'm not familiar with all the possible ways splits can be opened. Can there ever be opened a vertical split where other windows in the tab might change in height? That would complicate things.

The `q:` one is interesting, not sure why it isn't handled currently. Is it not a regular window? I'll check it out.

luukvbaal:
@bfrg I think I addressed everything you mentioned, thanks again for testing it out.

luukvbaal:
Alright, I'm happy with how this looks and works now. I would consider it mergeable and would put it up for review @brammool but I haven't had any feedback yet from neovim side on the omission of the `noscroll` option(or here for that matter, but there were some opinions shared on the option earlier and it originated over at the neovim PR). The addition of the jumplist entry with the old cursor position makes it even more obsolete imo.

The failing test on appveyor with gvim still requires fixing I guess but I haven't been able to reproduce it yet. It seems to be running the test on windows, perhaps I can reproduce it there.

bfrg:
For vertical splits, the cursor is still moved even though it's not necessary.

To reproduce, place the cursor on the very last line of the window and the open a new vertical split with `<c-w>v`. In the new window, the cursor remains in the same line as in the previous window, but when I jump back with `<c-w>p`, the cursor has been moved up. If I press `<c-w>p` again, it has been moved in the other window as well. The cursors should remain at their previous positions since no window was scrolled. You can try to move the cursor to the lowest line in each window and then focus the previous window. The cursor will be moved up again. It looks like the amount it is moved up depends on `scrolloff` which is set to 5 by default.

Pressing `<c-o>` works as expected :+1: .

luukvbaal:
> To reproduce, place the cursor on the very last line of the window and the open a new vertical

Oh right, though this is only true and "not necessary" when the last line of the window is also the last line of the buffer/file. I can't reproduce it on any arbitrary position in the buffer when placing the the cursor on the last line(also not possible when scrolloff is set). Is that what you mean, or are you seeing this without scrolloff set/not on the last line of the buffer?

luukvbaal:
@brammool I would like to do some refactoring and look into the failing test(seems it fails on gvim although I can't reproduce), perhaps hold off on review if you were planning to.

brammool:
Can some people try this out and give feedback?

spsc_correct_cursor() and spsc_correct_scroll() are  declared static but the function are not.

puremourning:
I gave this a quick spin. It seems to work reasonably well and the results were unsurprising, which has to be the goal.

However, I was sort of confused by the result of `:1spl` I was expecting the 1-line split to cover the first line in the buffer in the same way that in `:split` or `rightbelow 1split` we sort of "steal" the screen lines at the bottom of the buffer.

here's some playing including the result of a number of 1spl actions: https://asciinema.org/a/oA68f6K9CKHRV04HMfLYiZe0p

luukvbaal:
Yes, I would agree the result of `:1split` is surprising. It seems to work as expected for `2:split`. I doubt many people ever run `:1split` but I'll look into it thanks.

puremourning:
reason I tested 1split - CSV files. Useful to "freeze" the first row, given this feature.

luukvbaal:
Should be fixed now, the culprit was an erroneous empty buffer guard. I think there was a reason why I had it at `botline < 3` but the tests don't seem to fail anymore with it at `< 2`.

luukvbaal:
Added some minor fixes and comments.

luukvbaal:
Added a todo note regarding the topline wrap offset.

luukvbaal:
> There is a todo item somewhere to have the first line possible start
halfway the line.

I wanted to reference the existing todo not but couldn't find it btw.

brammool:
> > There is a todo item somewhere to have the first line possible start
> > halfway the line.
> 
> I wanted to reference the existing todo not but couldn't find it btw.

Search for "smoothscroll".

-- 
Q: How does a UNIX Guru do Sex ?
A: unzip;strip;touch;finger;mount;fsck;more;yes;umount;sleep

luukvbaal:
Sorry this is like the fourth time I said this but I kept on finding more edge cases. I think I'm done for now...

luukvbaal:
I think I found some (roundabout) condition that handles the winbar offset. 

Comment from neovim PR: It's worth noting that some edge case handling in the vim patch regarding winbar/tabline doesn't seem to be necessary in neovim due to it's inner window size concept. On the other hand, for Neovim we require some checks for floating windows that don't seem to be necessary for vims popup windows. I'll update accordingly once the vim patch is finalized and merged (or rejected).

runtime/doc/options.txt
```diff
@@ -7511,6 +7511,17 @@ A jump table for the options with a short description can be found at |Q_op|.
 	When on, splitting a window will put the new window right of the
 	current one. |:vsplit|
 
+			*'splitscroll'* *'spsc'* *'nosplitscroll'* *'nospsc'*
+'splitscroll' 'spsc'	boolean	(default 1)
```
```suggestion
'splitscroll' 'spsc'	boolean	(default on)
```

luukvbaal:
After multiple refactors I think this is as clean as I'm gonna get this PR in it's current form. Feedback/review is welcome.

One thing I still want to look into is determining a better scroll position for the splits in between the top and bottom window in the case of 3 horizontal splits or more. Maintaining topline/botline works great for the top/bottom window, but any additional splits should also try to keep the same buffer visible as much as possible (when splitting the current buffer). Will require changing the topline with some offset related to winrow/winheight for these middle windows presumably. This shouldn't change the shape of the PR much though, just add another case in the FOR_ALL_WINDOWS loop.

luukvbaal:
> One thing I still want to look into is determining a better scroll position for the splits in between the top and bottom window in the case of 3 horizontal splits or more. 

Nearly got this working, I'll update soon.

brammool:
> We now keep the same screenlines when splitting a window multiple
> times in all new opened verital/horizontal splits. Not sure how useful
> it is but it better adheres to the option description, and is pleasing
> to the(my) eyes. Of course we also still avoid scrolling when not
> splitting the current window but opening e.g. the quickfix window:
> [asciicast](https://asciinema.org/a/ZfjX6YmExZcXV0bVZSUprg9cS)

I've been holding off on including the patch, since there are still
improvements being made.  Let me know when it's ready.

-- 
ARTHUR:       Now stand aside worthy adversary.
BLACK KNIGHT: (Glancing at his shoulder) 'Tis but a scratch.
ARTHUR:       A scratch?  Your arm's off.

src/window.c
```diff
@@ -6314,6 +6334,104 @@ set_fraction(win_T *wp)
...
+	// buffer has less lines than the window height.
+	if (wp->w_height != wp->w_prev_height
+		&& wp->w_height < wp->w_buffer->b_ml.ml_line_count)
+	{
+	    lnum = wp->w_cursor.lnum;
+	    // Determine cursor position required to keep the same screenlines.
+	    if (wp->w_winrow != wp->w_prev_winrow)
+	    { // For bottom windows, keep the same botline except for 'ls' == 1.
+		if (!close && ((wp->w_winrow + VISIBLE_HEIGHT(wp)
+					+ wp->w_status_height) == cmdline_row))
```
luukvbaal: Is there any better way to determine whether a window is a bottom window i.e. above the cmdline? I would imagine something with vertical fr_layout and NULL fr_next but I'm not sure that's better and or correct. I guess a vertical frame can stop half-way the terminal window no? Then it wouldn't work, I haven't fully looked into the frame structure.

src/window.c
```diff
@@ -6314,6 +6334,104 @@ set_fraction(win_T *wp)
...
+		    wp->w_cursor.lnum = wp->w_botline - 1 - ls_offset;
+		else // Otherwise determine based on winrow and height.
+		    wp->w_cursor.lnum = MIN(wp->w_botline - 1
```
luukvbaal: I think this can evaluate to negative values, do we need to guard that here? I think it is caught in some other function. Don't see any problems at least.

src/window.c
```diff
@@ -6314,6 +6334,104 @@ set_fraction(win_T *wp)
...
+	    wp->w_fraction = FRACTION_MULT;
+	    scroll_to_fraction(wp, wp->w_prev_height);
+	}
+	// Change cursor position to avoid scrolling. Store and temporarily
+	// set curwin to save cursor position in wp with setmark().
+	else
+	{
+	    curwp = curwin;
+	    curwin = wp;
+	    setmark('\'');
```
luukvbaal: Is it worth parameterizing setmark() with a win_T argument to avoid mis-using curwin here?

luukvbaal (commented):
@brammool I think this is ready for review now, I've done all the improving I can and don't see any more problems. I'll pose some review questions.

src/window.c
luukvbaal: Alternatively, it feels like there should be a more efficient way to determine whether this offset is possible in win_split_ins() such that we can get rid of this if/else clause. I see there is a need_status in win_split_ins() already. If we can use the pre-existing logic in win_split_ins() to add the need_status offset to w_prev_height, that might be enough. I haven't figured out the correct logic in win_split_ins() though, feels like it should be possible with flags/WSP_BOT/before variable and whatnot.

brammool:
> @luukvbaal commented on this pull request.
> 
> @brammool I think this is ready for review now, I've done all the improving I can and don't see any more problems. I'll pose some review questions.
> 
> > +		if (!close && ((wp->w_winrow + VISIBLE_HEIGHT(wp)
> +					+ wp->w_status_height) == cmdline_row))
> 
> Is there any better way to determine whether a window is a bottom
> window i.e. above the cmdline? I would imagine something with vertical
> fr_layout and NULL fr_next but I'm not sure that's better and or
> correct. I guess a vertical frame can stop half-way the terminal
> window no? Then it wouldn't work, I haven't fully looked into the
> frame structure.

Using cmdline_row for this seems a bit brittle.  I see one other place
where this is done.  Perhaps comparing with topframe->fr_height works
better?

> 
> > +    FOR_ALL_WINDOWS(wp)
> +    {
> +	// Skip when window height has not changed or when
> +	// buffer has less lines than the window height.
> +	if (wp->w_height != wp->w_prev_height
> +		&& wp->w_height < wp->w_buffer->b_ml.ml_line_count)
> +	{
> +	    lnum = wp->w_cursor.lnum;
> +	    // Determine cursor position required to keep the same screenlines.
> +	    if (wp->w_winrow != wp->w_prev_winrow)
> +	    { // For bottom windows, keep the same botline except for 'ls' == 1.
> +		if (!close && ((wp->w_winrow + VISIBLE_HEIGHT(wp)
> +					+ wp->w_status_height) == cmdline_row))
> +		    wp->w_cursor.lnum = wp->w_botline - 1 - ls_offset;
> +		else // Otherwise determine based on winrow and height.
> +		    wp->w_cursor.lnum = MIN(wp->w_botline - 1
> 
> I think this can evaluate to negative values, do we need to guard that
> here? I think it is caught in some other function. Don't see any
> problems at least.

It is important that the cursor is on a valid line.  check_cursor_lnum()
can be used.

> 
> > +	    curwp = curwin;
> +	    curwin = wp;
> +	    setmark('\'');
> 
> Is it worth parameterizing setmark() with a win_T argument to avoid
> mis-using curwin here? 

It also uses "curbuf".  You could call setmark_pos() directly.

-- 
If your company is not involved in something called "ISO 9000" you probably
have no idea what it is.  If your company _is_ involved in ISO 9000 then you
definitely have no idea what it is.

luukvbaal:
> Using cmdline_row for this seems a bit brittle. I see one other place where this is done. 

Yes, I thought so too but I indeed copied it from the other use.

> Perhaps comparing with topframe->fr_height works better?

Not sure, still using winrow + height but instead comparing against topframe height instead of cmdline row you mean? I guess if there is no more idiomatic way to identify a bottom window in vim it might be better than cmdline.

> It is important that the cursor is on a valid line. check_cursor_lnum() can be used.

Yeah, I think we correct the cursor position in all possible control flows. Either directly by calling win_fix_cursor() in win_fix_scroll() for the current window, or later in win_enter_ext(). This temporary "ghost" cursor position is only being used in the scroll_to_fraction() call. My point was that it can evaluate to a negative number here but I guess we're fine because of:
https://github.com/vim/vim/blob/91a874eb88eff2df9b8fdd64ccf98eed337823cf/src/window.c#L6378-L6379


> It also uses "curbuf". You could call setmark_pos() directly.

Right, missed that. But setmark_pos() does not take a win/buf arg either. Do you mean it would be acceptable to add those?

brammool:
> > Using cmdline_row for this seems a bit brittle. I see one other
> > place where this is done. 
> 
> Yes, I thought so too but I indeed copied it from the other use.
> 
> > Perhaps comparing with topframe->fr_height works better?
> 
> Not sure, still using winrow + height but instead comparing against
> topframe height instead of cmdline row you mean? I guess if there is
> no more idiomatic way to identify a bottom window in vim it might be
> better than cmdline.

cmdline_row changes when messages cause the screen to scroll up.  While
the frame height isn't changed unless the window layout changes.

> > It also uses "curbuf". You could call setmark_pos() directly.
> 
> Right, missed that. But setmark_pos() does not take a win/buf arg
> either. Do you mean it would be acceptable to add those?

setmark_pos(c, &win->w_cursor, win->w_buffer->b_fnum)

Could add a setmark_win() for this, and also use it in setmark().

-- 
If the Universe is constantly expanding, why can't I ever find a parking space?

luukvbaal:
> Could add a setmark_win() for this, and also use it in setmark().

Sure, done in latest commit @brammool. Let me know if you think this is a justifiable change and preferred over (ab)using curwin/curbuf in win_fix_scroll(). To reiterate: the reason why we need to call setmark() for a window other than curwin, is to save and change the now invalid cursor position(due to a height change), before entering the window, which would cause scrolling to occur.

brammool:
> > Could add a setmark_win() for this, and also use it in setmark().
> 
> Sure, done in latest commit @brammool. Let me know if you think this
> is a justifiable change and preferred over (ab)using curwin/curbuf in
> win_fix_scroll(). To reiterate: the reason why we need to call
> setmark() for a window other than curwin, is to save and change the
> now invalid cursor position(due to a height change), before entering
> the window, which would cause scrolling to occur.

Yes, this is one of the disadvantages of keeping the view, then the
cursor may have to move.  Setting a mark is probably best.  Did you
check that when doing this twice you can go back to both old positions
with CTRL-O ?

There is a long standing todo item to allow for the cursor to be outside
of the view.  Especially useful when scrolling the text around.  Then
when doing something with the cursor, the view would jump back.  Or we
would have some command to make the cursor jump inside the view (like
what you did in this patch).  This is one of those changes that will
affect lots of code (most of which assumes the cursor position is valid
and visible), which is why it wasn't done yet.  Once it would be, it
could be used for this too.

It's hard to tell from staring at the code if this works properly, we'll
have to hear from users.  Unless there is still a problem I'll include
it in a few days.

-- 
It doesn't really matter what great things you are able to do if
you don't do any of them.  (Bram Moolenaar)

luukvbaal:
> Did you check that when doing this twice you can go back to both old positions with CTRL-O ? Unless there is still a problem I'll include it in a few days.

Yes, this seems to work. However, it currently does not work for next_curwin. Introducing setmark_win() still does not apply to that case, I think due to the usage of curwin in setmark_pos() (and setpcmark()) here:
https://github.com/vim/vim/blob/c1eb131c9eb38e00e08109f50e3e5337c072b71e/src/mark.c#L61-L72

I think a win_T argument to setmark_pos() and setpcmark() instead of using curwin directly would be necessary to make this work. Just a matter of replacing curwin/curbuf usage with wp/wp->w_buffer but perhaps keeping the solution isolated to win_fix_scroll() would be preferred? We can get around the problem there by setting curwin/curbuf temporarily and simply calling setmark().

src/window.c
```diff
@@ -6314,6 +6334,104 @@ set_fraction(win_T *wp)
...
+    if (nlnum)
+    {   // Ensure cursor stays visible if we are not in normal mode.
+	if (normal)
+	{
+	    wp->w_fraction = FRACTION_MULT;
+	    scroll_to_fraction(wp, wp->w_prev_height);
+	}
+	else
+	{
+	    curwin = wp;
```
luukvbaal: @brammool I think this workaround is fine. Keeps the problem/solution isolated and it works properly. I guess if there ever is a need to add a setmark_win() in the future it could be used here too but for now I would do it like this.

luukvbaal:
> There is a long standing todo item to allow for the cursor to be outside of the view.

Yes, that sounds like it could be useful in some cases. Not sure it directly applies to this patch though. I can imagine some overlap but it's hard to tell if it could be re-used for this patch.

> Unless there is still a problem I'll include it in a few days.

The only remaining "problem" is the bottom window detection. Not sure if it is a problem though. I'm open to a better way of identifying the bottom windows, but I could not find one.

luukvbaal:
We now also keep the same scroll position when resizing windows as @bfrg had suggested earlier. I think it fits the option.

luukvbaal:
@brammool is there any way I can use screenrow() inside in a test? I would like to get rid of the winbar/tabline compensation and just test for the screenrow being equal to the line the cursor is on: screenrow() == line("."). call assert_equal(screenrow(), line(".")) does not work presumably due to the same reason as described by the limitation in the docs. I tried the mapping workaround as done in Test_conceal_cursor_pos() by setting a variable but could not get it to work.

luukvbaal:
> Not sure, since you are sourcing a script the cursor is not on the command line. Perhaps you need to :redraw ? Alternatively use a screendump test. It is slower, but for checking the window layout it's very easy to see what fails.

Got the resize tests working. I presume setting 'lines' to resize the shell/gui is not really recommended or well defined. It seems whether the test passes or not depends on the actual window size vs. 'lines' value used in the test. I found values that work for now. Might be brittle but at least we are testing the behavior.

Writing a screendump test for each state that is tested currently seems cumbersome(granted we are probably testing redundantly with these nested loops). Or is there some way to keep some of the nested logic/iterative window splitting while doing screendumps? Anyhow, I think all the behavior added in the patch is tested now. Accounting for the winbar/tabline offset was annoying while adding new tests but converting the test seems unnecessary now..

brammool:
> > Not sure, since you are sourcing a script the cursor is not on the
> > command line. Perhaps you need to :redraw ? Alternatively use a
> > screendump test. It is slower, but for checking the window layout
> > it's very easy to see what fails.
> 
> Got the resize tests working. I presume setting 'lines' to resize the
> shell/gui is not really recommended or well defined. It seems whether
> the test passes or not depends on the actual window size vs. 'lines'
> value used in the test. I found values that work for now. Might be
> brittle but at least we are testing the behavior.
> 
> Writing a screendump test for each state that is tested currently
> seems cumbersome(granted we are probably testing redundantly with
> these nested loops). Or is there some way to keep some of the nested
> logic/iterative window splitting while doing screendumps? Anyhow, I
> think all the behavior added in the patch is tested now. Accounting
> for the winbar/tabline offset was annoying while adding new tests but
> converting the test seems unnecessary now..

When testing a lot of situations you would end up with a lot of
screendumps, that is inconvenient.  Especially if you can calculate the
expected result.

-- 
38. You wake up at 3 a.m. to go to the bathroom and stop and check your e-mail
    on the way back to bed.

src/testdir/test_window_cmd.vim
```diff
@@ -1631,5 +1631,140 @@ func Test_win_equal_last_status()
...
+              " No scroll when equalizing windows
+              wincmd =
+              call assert_equal(1, line("w0"))
+              wincmd j
+              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
+              wincmd k
+              call assert_equal(1, line("w0"))
+
+              " No scroll in windows split multiple times
+              if ea || !so " Windows too small to guarantee no scroll with 'scrolloff'
```
luukvbaal: Sorry for another barrage of commits @brammool. I know you said you would include it a fey days ago but I cannot help myself iterating on the patch if I see room for improvement.
I think the only remaining caveat now is the one that is avoided in this test. I'm not sure how or if it can be fixed. We correctly avoid scrolling in the small window height/scrolloff case when splitting the window by setting the cursor in the middle of the window. If it would not be in the middle of the window vim would place it there when scrolloff > window height / 2, causing scrolling to occur. However subsequent win_fix_cursor() calls in win_enter_ext() can cause the cursor to move by 1. I think because the middle of the window we are calculating might be floored. Not sure how to determine whether the floored or ceiled value is needed to avoid scrolling:
[asciicast](https://asciinema.org/a/eAKo7yBe6Ld1uP656SbE08Pag)

brammool:
> @luukvbaal commented on this pull request.
> 
> 
> 
> > +              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
> +              set lines-=2
> +              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
> +              wincmd k
> +              call assert_equal(1, line("w0"))
> +
> +              " No scroll when equalizing windows
> +              wincmd =
> +              call assert_equal(1, line("w0"))
> +              wincmd j
> +              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
> +              wincmd k
> +              call assert_equal(1, line("w0"))
> +
> +              " No scroll in windows split multiple times
> +              if ea || !so " Windows too small to guarantee no scroll with 'scrolloff'
> 
> Sorry for another barrage of commits @brammool. I know you said you
> would include it a fey days ago but I cannot help myself iterating on
> the patch if I see room for improvement.
> I think the only remaining caveat now is the one that is avoided in
> this test. I'm not sure how or if it can be fixed. We correctly avoid
> scrolling in the small window height/scrolloff case when splitting the
> window by setting the cursor in the middle of the window. If it would
> not be in the middle of the window vim would place it there when
> scrolloff > window height / 2, causing scrolling to occur. However
> subsequent win_fix_cursor() calls in win_enter_ext() can cause the
> cursor to move by 1. I think because the middle of the window we are
> calculating might be floored. Not sure how to determine whether the
> floored or ceiled value is needed to avoid scrolling:
> [asciicast](https://asciinema.org/a/eAKo7yBe6Ld1uP656SbE08Pag)

Does this really matter or are we now splitting hairs?

What happens when 'scrolloff' can't be applied is a bit unpredictable,
the round-off can go either way.  It tries to avoid scrolling unless
it's necessary.

-- 
Those who live by the sword get shot by those who don't.

luukvbaal:
> Does this really matter or are we now splitting hairs?
Yeah I agree it doesn't matter much. I just thought I'd mention the shortcoming and the reason for skipping that part of the test.

src/window.c
```diff
@@ -6316,6 +6346,98 @@ set_fraction(win_T *wp)
...
+	    && (wp->w_botline - wp->w_buffer->b_ml.ml_line_count) != 1)
+	nlnum = MAX(wp->w_botline - so - 1, 1);
+    // If cursor was invalid scroll or change cursor.
+    if (nlnum)
+    {
+	if (normal)
+	{   // Make sure cursor is closer to topline than botline.
+	    if (so == wp->w_height / 2
+		    && nlnum - wp->w_topline > wp->w_botline - 1 - nlnum)
+		nlnum--;
```
luukvbaal, line 6428: Turns out that when 'scrolloff' cannot be applied due to the window height being too small, we just need to make sure that the cursor we set to avoid scrolling is closer to topline than botline. Was able to identify the condition to make sure this happens. Setting 'lines' and skipping part of the test for 'noequalalways' and 'scrolloff' is now no longer required to make the test pass.

luukvbaal:
@brammool this patch is finished as far as I'm concerned. I might look into making it work with wrapped/folded lines but I suspect that will be non-trivial and could be done in a follow-up patch.

brammool:
> @brammool this patch is finished as far as I'm concerned. I might look
> into making it work with wrapped/folded lines but I suspect that will
> be non-trivial and could be done in a follow-up patch.

Thanks for persistently working on this feature.  I'll include it now.

Perhaps I'll work on 'smoothscroll' one of these days, then it can be
made to work better.

-- 
George:  "I just got a new set of golf clubs for my wife!"
  John:  "Great trade!"

brammool:
Included as patch 9.0.0445

ychin:
Nice change! Just trying it out and my initial reaction was that I only wanted this behavior when splitting on the bottom, but not on the top (the setting doesn't allow you to pick), but so far I think I have been won over to have it work for both situations.

> Hmm, I think what you are trying to do is not 100% possible. The topline can only be the start of a buffer line, thus when this line wraps it occupies two screen lines. If the window moves down one screen line, then either you keep topline, in which case the window contents scrolls down one screen line, or you increment topline, in which case the window contents scrolls up one screen line. See what I mean? There is a todo item somewhere to have the first line possible start halfway the line. That would have to be implemented first. The reason it's not done yet is that it will require a lot of work to get right.

> Search for "smoothscroll".

Just for reference @brammool I think there is a WIP PR on Neovim to handle this: https://github.com/neovim/neovim/pull/11014. I'm not sure if you want to take a look or coordinate, since once it gets merged to Neovim I'm sure you will have to deal with this anyway as people try to submit patches to port it over. (It's been in works since 2019 though and traction on it seems to have slowed down)

brammool:
> > Search for "smoothscroll".
> 
> Just for reference @brammool I think there is a WIP PR on Neovim to
> handle this: https://github.com/neovim/neovim/pull/11014. I'm not sure
> if you want to take a look or coordinate, since once it gets merged to
> Neovim I'm sure you will have to deal with this anyway as people try
> to submit patches to port it over. (It's been in works since 2019
> though and traction on it seems to have slowed down)

It is not surprising it takes a long time, it's a complicated feature.
It's the kind of thing where the first change to make it show isn't so
difficult, but then there are lots and lots of side effects to take
care of.

I have to balance fixing bugs (several of them in recently added
features) and adding cool features.  Of course it helps a lot if someone
makes a PR and others give feedback.

-- 
47. You are so familiar with the WWW that you find the search engines useless.

ychin:
Back to `splitscroll`, after using this for a little bit, I have a bit of mixed reaction to it (as I alluded to in the last comment). I feel like this new option is essentially two arguably different features in one:

1. *Stability when split on bottom*: **Don't scroll the window when splitting.** If the cursor would be out of bounds, just move it. It's a pretty simple idea and one that I quite like, and actually seems like the original motivation behind the plugin (since it makes a bottom split not affect the top window's stability)
2. *Stability when split on top*: **When splitting on top, *do* counter-scroll the window just enough** to make it look visually stable on the bottom. This feature seems a bit more complicated, and there are edge cases where it's impossible to maintain the visual stability.

I quite like (1) but I find (2) a little weird and inconsistent with my intuitive feel of how this option should work. In short, it doesn't seem to me that when splitting on top, the bottom window should "preserve stability". Let me attach this video to illustrate my point, of me dragging the status bar up and down normally:

https://user-images.githubusercontent.com/1217449/189979819-0857a03f-d4f2-48cb-8480-38d3b96276e4.mov

As you can see, the top window stays put, whereas the bottom window moves along and doesn't stay "stable", because the bottom window is always subject to the top window's sizing, as Vim splits are top-aligned.

For (2), there are also edge cases where you can't maintain the stability. That happens when you have wrapped lines (as mentioned above), and also when the file just isn't long enough (since you can't scroll too far down).

Furthermore, I think the option name "splitscroll" is a misnomer. For (1), it's actually accurate, as it prevents the window from scrolling. For (2), we are *still* scrolling the window, albeit just the right amount to make it look "stable" by counter-scrolling the exact amount against the split window's size. The bottom window is inherent subject to the top window's size, so "not scrolling" should just mean it gets offset by the same amount.

Personally, I would love it if we can just have (1), and (2) being a nice extra that I think is not as desired as a feature, even though it's more complicated to implement.

---

I do realize it's probably better if I filed these feedbacks earlier, but I didn't see this PR before.

ychin:
Note that this is unrelated to the original `splitscroll=1` and `splitscroll=2` differentiation for those who didn't read through the history. That was for whether to keep the cursor in the window or not.

But in general I just feel like watching videos of people using this feature, the vast  majority of users who want this are trying to fix (1) as I mentioned above: having a bottom split not cause the other splits to scroll which a lot of people find unintuitive. It seems to me that just addressing (1) separately would result in a most intuitive setting (by basically having Vim to be as "not trying to be smart" as possible)

luukvbaal:
> I feel like this new option is essentially two arguably different features in one

Depends on the frame of reference you take. Whether you consider scrolling the movement of the topline, or the screenlines. The point of this option is to reduce the visually jarring change text that is displayed on each line when opening or closing splits. To keep them stable and avoid "scrolling" of screenlines, we indeed "scroll" the topline if the location of the start of a split changes. Whether this happens on the top or the bottom or in between two splits makes no difference from that point of reference and is how I designed the option. Indeed this feedback would have been better earlier on. I see no reason to discern the two cases on the moment of splitting a window. If you are just talking about actively resizing a window, e.g. by dragging the statusline, that is a feature that was added later and indeed could be considered two features in one. We could separate that out if desired.

ychin:
I think “scroll position”, both in implementation and common usage, usuallt refers to the offset from the top. To “scroll” is to change the scroll position. That’s what I meant in the video I attached: when a window doesn’t scroll under normal circumstances, the bottom window *will* shift when the top one changes.

You can see this in basically any other split-enabled text editors. When you open a split at the bottom, the top doesn’t scroll. When you open a top split, the bottom gets moved to account for it but the editor wouldn’t randomly scroll the bottom split to account for this. I think people opening a top split actually *expects* the jump rather than forcing a stable counter-scroll.

it feels to me that (1) is what the most intuitive option would be and most consistent with common UX, as it just eliminates the scroll that Vim used to do. (2) is actually a little counter to what most other editors do.

ychin:
> If you are just talking about actively resizing a window, e.g. by dragging the statusline, that is a feature that was added later and indeed could be considered two features in one. We could separate that out if desired.

No just to be clear what I meant is actively resizing the window is doing the right thing. When you resize windows it doesn’t scroll, and hence the bottom window is moved down. I don’t think anyone wants a different option for that to be different. What I was trying to say with that video is that that was what “no scrolling” means and what I first thought the option would do, instead of the additional feature (2) that tries to scroll the window for “stability”.

luukvbaal:
I can see where you're coming from when the split opened at the top is a different buffer. But the behavior is more so designed for what happens when splitting the current buffer, i.e. pressing <kbd>Ctrl-S W</kbd>. We are then counter-scrolling as you call it to just show the same screen, with a horizontal separator line. This then also happens when opening a different buffer in the top split, and I prefer it like that, but it could be considered a by-product of the option. The point is, the text that was already on the screen will not move when opening or closing a split, whether it be above or below, that was the entire point. Feel free to separate out the behavior if you feel so inclined, I won't be doing so.

luukvbaal:
Also how would this look in the situation you describe?
https://asciinema.org/a/LgB40B5GFNncp8EDnzlMPEzTS
Which window will scroll and which will not?

ychin:
> But the behavior is more so designed for what happens when splitting the current buffer,

This feature affects all splitting though, including using `:help` etc.

> Also, how would this look in the situation you describe? Which window will scroll and which will not? You say this is counter intuitive and works differently in other split enabled editors. Do you have an example? I've never used a different editor that allows you to split the windows like so:

None of the windows would scroll. This means the bottom windows would get pushed down visually rather than remaining stable. As I mentioned, "scrolling" as a term means the offset from the top. "Not scrolling" means you don't change the top line number, meaning the bottom window gets pushed down.

Here's an example from VSCode:

https://user-images.githubusercontent.com/1217449/190060555-13468b98-be53-4ee3-97cc-7597bbedf720.mov

As we can see, the top split doesn't change in "stability", but the bottom split got pushed down. Both splits are still at line 20 at the top.

> Feel free to separate out the behavior if you feel so inclined, I won't be doing so.

That's fair. I mostly wanted to post what I think is an unintuitive behavior while the feature is fresh, especially when compared with what I think "not scrolling" should mean and also when comparing with other text editors. Now that this feature is in there should be a higher bar to change things, but I want to see what other what other users think instead of just me vs you.

ychin:
Here's how Xcode does it. It's the same situation. The top line number at the window (2000) stays constant when you split. They don't dynamically re-calculate the bottom split to make it stay visually constant. My take is that either way it may or may not be what the user wants, and hence they do the simpler thing by just not changing the scroll offset of the window.

https://user-images.githubusercontent.com/1217449/190061853-53881d4a-1e03-4894-b9d8-af3a956c8d8a.mov

Sorry to hammer on this, but I just wanted to get my point across I guess, since I think the way this PR (which does have its uses!) goes against the grain of how other editors usually implement this.

brammool:
> Here's how Xcode does it. It's the same situation. The top line number
> at the window (2000) stays constant when you split. They don't
> dynamically re-calculate the bottom split to make it stay visually
> constant. My take is that either way it may or may not be what the
> user wants, and hence they do the simpler thing by just not changing
> the scroll offset of the window.
> 
> https://user-images.githubusercontent.com/1217449/190061853-53881d4a-1e03-4894-b9d8-af3a956c8d8a.mov
> 
> Sorry to hammer on this, but I just wanted to get my point across I
> guess, since I think the way this PR (which does have its uses!) goes
> against the grain of how other editors usually implement this.

Keeping the topline at the same buffer line is an alternative that some
users might like.

For a window where the position of the window doesn't change, only the
height, the same thing  happens.

For a window that vertically moves, the current implementation tries to
keep the text at the same screen line, therefore the "no scrolling"
name.  However, the screen lines that are now part of a different window
will change anyway.  Thus it's not that the screen stays the same.

Adding the option to keep the topline the same might actually be simpler
to implement.  The question is, now that 'splitscroll' is a boolean
option, how do we let the user make the choice?  It's not too late to
change the type of the option or rename it.  The "split" part isn't
quite right, since it also affects what happens when a window is closed.

-- 
65. The last time you looked at the clock it was 11:30pm, and in what
    seems like only a few seconds later, your sister runs past you to
    catch her 7am school bus.

luukvbaal:
> Adding the option to keep the topline the same might actually be simpler to implement.

I think it might amount to simply guarding this clause behind the option. If users really want this I could look into it but the option shouldn't get needlessly complicated imo.
https://github.com/vim/vim/blob/febe13892ede37d2a8462b9921e05a55e059db0c/src/window.c#L6373-L6383

> The "split" part isn't quite right, since it also affects what happens when a window is closed.

If you consider the help doc "The value of this option determines the scroll behavior when opening, closing or resizing horizontal splits." I think it does fit. "split" meaning "split windows" not the verb "splitting a window". However I am not attached to the name.

luukvbaal:
If we keep the same name and option behavior, I suppose the option's short description needs updating: "determines scroll behavior when splitting windows" -> "determines scroll behavior for split windows".

brammool:
> If we keep the same name and option behavior, I suppose the option's
> short description needs updating: "determines scroll behavior when
> splitting windows" -> "determines scroll behavior for split windows".

Yes, that works better.  I can't think of a better name for "opening,
closing, resizing windows".

Let's await opinions about the idea to just keep the topline the same.

-- 
72. Somebody at IRC just mentioned a way to obtain full motion video without
    a PC using a wireless protocol called NTSC, you wonder how you never
    heard about it

luukvbaal:
> Adding the option to keep the topline the same might actually be simpler to implement.

> >I think it might amount to simply guarding this clause behind the option. If users really want this I could look into it but the option shouldn't get needlessly complicated imo.

So yeah circling back to this after some initial user bug reports/improvements to the feature, I really do think providing this option is as simple as guarding the `winrow != prev_winrow` clause behind an alternate option value. We are then simply (in)validating botline for each window, and correcting the cursor position for curwin.

To reiterate, instead of having the previously proposed options for scroll behavior depending on whether or not the cursor would be valid ({"normal", "noscroll", "neverscroll"}, we dropped this because changing the cursor position was considered not to be that egregious when splitting a window, and as a compromise we add the previous cursor position to the jumplist).

Now the proposal is the have 3 separate scroll behaviors for 'splitscroll' because user(s)(@ychin) might prefer one over the other:
* "normal": default behavior, keep the same relative cursor position in both new and old split.
* "counterscroll": keep the text that was already visible in a window on the same screenlines for each window that is resized/moved by a split command, effectively "counter-scrolling" the topline to keep the text stable. (current 'nosplitscroll' behavior).
* "noscroll": keep the same topline in all windows, effectively only keeping the text stable in the uppermost split.

Not sure "counterscroll" is very telling, can anyone think of a better name for these option values? Do any more users see the need for the third option? I'll cross-post in the (as of yet unmerged) neovim PR.

brammool:
> > Adding the option to keep the topline the same might actually be
> > simpler to implement.
> 
> > >I think it might amount to simply guarding this clause behind the
> > >option. If users really want this I could look into it but the
> > >option shouldn't get needlessly complicated imo.
> 
> So yeah circling back to this after some initial user bug
> reports/improvements to the feature, I really do think providing this
> option is as simple as guarding the `winrow != prev_winrow` clause
> behind an alternate option value. We are then simply (in)validating
> botline for each window, and correcting the cursor position for
> curwin.
> 
> To reiterate, instead of having the previously proposed options for
> scroll behavior depending on whether or not the cursor would be valid
> ({"normal", "noscroll", "neverscroll"}, we dropped this because
> changing the cursor position was considered not to be that egregious
> when splitting a window, and as a compromise we add the previous
> cursor position to the jumplist).
> 
> Now the proposal is the have 3 separate scroll behaviors for
> 'splitscroll' because ***@***.***) might prefer one over the other:
> * "normal": default behavior, keep the same relative cursor position
> in both new and old split.
> * "counterscroll": keep the text that was already visible in a window
> on the same screenlines for each window that is resized/moved by a
> split command, effectively "counter-scrolling" the topline to keep the
> text stable. (current 'nosplitscroll' behavior).
> * "noscroll": keep the same topline in all windows, effectively only
> keeping the text stable in the uppermost split.
> 
> Not sure "counterscroll" is very telling, can anyone think of a better
> name for these option values? Do any more users see the need for the
> third option? I'll cross-post in the (as of yet unmerged) neovim PR.

If we call the option 'splitkeep' the value can say what we keep:
- "cursor" - relative cursor position == 'splitscroll'
- "screen" - keep the same screen lines == 'nosplitscroll'
- "topline" - keep the same line at the top of the window

One misnomer is that it would also happen when closing windows, not only
when splitting.  What is a short name for split + close ?

I wonder if we would ever have something equivalent for vertical splits.
If we do, we probably can use a different option and not worry about it
now.

-- 
225. You sign up for free subscriptions for all the computer magazines

luukvbaal:
> If we call the option 'splitkeep' the value can say what we keep: - "cursor" - relative cursor position == 'splitscroll' - "screen" - keep the same screen lines == 'nosplitscroll' - "topline" - keep the same line at the top of the window One misnomer is that it would also happen when closing windows, not only when splitting. What is a short name for split + close ?

Yes, that could work. Not sure there is a better name for split + close (+resize).

> I wonder if we would ever have something equivalent for vertical splits. If we do, we probably can use a different option and not worry about it now.

The default behavior has never bothered me there before, I guess it's only an issue for long lines anyways. But if at any point we want to add it I guess it would make sense to have a complementary option "vsplitkeep". There it would be just "cursor" vs "screen" I guess.

bew:
> One misnomer is that it would also happen when closing windows, not only when splitting.

Have you considered `scrollkeep` ?

PR #10688, opened 2022-07-10
Update Chinese menu items translations

adaext:
Since "shellmenu.vim", "debchangelog.vim", and "synmenu.vim" have updated, the template menu items should also be updated.


brammool:
I'll include it, thanks.

PR #10689, opened 2022-07-10
Remove comment about orig_regname.

errael:
This part of the comment should have been remove when
`orig_regname != 0` was removed in `v8.1.0999`.


brammool:
I'll include it, thanks.

Issue #10690, opened 2022-07-10
:match highlight of NS throws an exception


haron13-2019:
### Steps to reproduce

1.   Create shell script with following content (cat ./killer_3mat_ns.sh):   

    # script which defines Todo highlight for non-breakable space
    # and move cursor to opening brace of the braces pair
    echo -e "\n{\na\xc2\xa0b c\n}\n" > ./killer.txt
    echo -e "set nocompatible|syn on|color default\n\nautocmd BufReadPost *\n  \\\ exec \"3mat Todo /\\\xc2\\\xa0/\"" > ./killer_vimrc.vim
    # with that you will need move down one line yourself to expose the bug
    #vim -Nu ./killer_vimrc.vim -c 'normal gg' ./killer.txt
     vim -Nu ./killer_vimrc.vim -c 'normal ggj'  ./killer.txt
    # ^^^ this will move down for you with "j" and trigger the bug

2.  Execute it:

    bash ./killer_3mat_ns.sh

3.  Error message is shown:

    "./killer.txt" 5L, 13B
    Error detected while processing CursorMoved Autocommands for "*"..function <SNR>14_Highlight_Matching_Pair:
    line  148:
    E801: ID already taken: 3
    Press ENTER or type command to continue

### Expected behaviour

There should be no error message

### Version of Vim

9.0.0049 huge with GTK

### Environment

OS: openSUSE_Thumbleweed, cygwin
Terminal: xterm (xterm -v -> Xterm(372))
value of $TERM: xterm-256color
Shell: /bin/bash 5.1.16(1)-releaseOS: openSUSE_Thumbleweed, cygwin
Terminal: xterm (xterm -v -> Xterm(372))
$TERM: xterm-256color
Shell: /bin/bash (5.1.16(1)-release

### Logs and stack traces

```shell
:scriptnames
    1: ~/data/killer_3mat_ns/killer_vimrc.vim
    2: /usr/local/share/vim/vim90/syntax/syntax.vim
    3: /usr/local/share/vim/vim90/syntax/synload.vim
    4: /usr/local/share/vim/vim90/syntax/syncolor.vim
    5: /usr/local/share/vim/vim90/colors/lists/default.vim
    6: /usr/local/share/vim/vim90/filetype.vim
    7: /usr/local/share/vim/vim90/scripts.vim
    8: /usr/local/share/vim/vim90/autoload/dist/script.vim
    9: /usr/local/share/vim/vim90/colors/default.vim
[...]

PR #10691, opened 2022-07-10
remove misleading :3mat from the documentation

haron13-2019:
When skimming over documentation it is easy to erroneously believe one can use ":3mat" command - see my recent bug report #10690 instantly closed by @chrisbra   ))))

Thinking about it a bit, may be this is good idea to officially remove ":3mat" from the documentation. For normal use cases it is anyway not (really) available, it can be only used (as far as I understood) only in some exotic circumstances - for example that standard matchit plugin is deactivated?

Another "improvement"  - documentation encourages plugin authors to use ":2mat". Is it not better to leave ":mat" and ":2mat" for interactive usage entirely and ask plugin authors to use matchadd() call directly?

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -2316,7 +2316,6 @@ exists({expr})	The result is a Number, which is |TRUE| if {expr} is defined,
 					To check for a supported command
 					always check the return value to be 2.
 			:2match		The |:2match| command.
-			:3match		The |:3match| command.
 			#event		autocommand defined for this event
 			#event#pattern	autocommand defined for this event and
 					pattern (the pattern is taken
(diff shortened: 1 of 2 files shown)


vim-ml:
On 2022-07-10, haron13-2019 wrote:
> When skimming over documentation it is easy to erroneously believe one can use
> ":3mat" command - see my recent bug report #10690 instantly closed by @chrisbra
> ))))
> 
> Thinking about it a bit, may be this is good idea to officially remove ":3mat"
> from the documentation. For normal use cases it is anyway not (really)
> available, it can be only used (as far as I understood) only in some exotic
> circumstances - for example that standard matchit plugin is deactivated?
> 
> Another "improvement" - documentation encourages plugin authors to use ":2mat".
> Is it not better to leave ":mat" and ":2mat" for interactive usage entirely and
> ask plugin authors to use matchadd() call directly?

I think your change goes too far.  The :3match command _does_ exist
and it _can_ be used.  In fact, it _is_ used, by the matchparen
plugin.  I don't think that hiding it or pretending that it doesn't
exist benefits the user.  I certainly would not remove the mentions
of the matchparen plugin.

For example, matcharg(3) returns a different result than matcharg(4)
and :3match will work while :4match will fail.  This could be
confusing to the user without an discoverable explanation.

The semi-reserved status of match ID 3 should be mentioned in a few
more places, such as ":help matchadd()", but I would not remove any
of the existing documentation on it.

Regards,
Gary

tonymec:
If someone reads the source of the matchparen (not matchit) plugin and tries to understand what it does when and how, he'll find that the :3match instruction is used. IMHO it runs totally against the whole Vim philosophy to have such a feature intentionally entirely undocumented: in Vim, **everything** is documented (or if not, it is an unintentional oversight).

Best regards,
Tony.

brammool:
As mentioned, removing the :3match command and other info goes against the intention of having help for everything.
I'll include some remarks where a match is added that the ID three is being used by matchparen.

PR #10692, opened 2022-07-11
Synchronize translation files with different encodings but with the same content.

adaext:
"menu_sk_sk.iso_8859-2.vim" should be generated from "menu_slovak_slovak_republic.1250.vim" automatically,
"menu_hu_hu.iso_8859-2.vim" should be generated from "menu_hu_hu.utf-8.vim" automatically.


runtime/lang/Makefile
```diff
@@ -194,3 +197,15 @@ menu_uk_ua.koi8-u.vim: menu_uk_ua.utf-8.vim
...
+menu_hu_hu.iso_8859-2.vim: menu_hu_hu.utf-8.vim
+	rm -f menu_hu_hu.iso_8859-2.vim
+	iconv -f utf-8 -t iso8859-2 menu_hu_hu.utf-8.vim | \
+		sed -e 's/scriptencoding utf-8/scriptencoding iso8859-2/' -e 's/" Original translations/" Generated from menu_hu_hu.utf-8.vim, DO NOT EDIT/' > menu_hu_hu.iso_8859-2.vim
+
+# Convert menu_slovak_slovak_republic.1250.vim to create menu_sk_sk.iso_8859-2.vim.
+menu_sk_sk.iso_8859-2.vim: menu_slovak_slovak_republic.1250.vim
+	rm -f menu_sk_sk.iso_8859-2.vim
+	iconv -f cp1250 -t iso8859-2 menu_slovak_slovak_republic.1250.vim | \
+		sed -e 's/scriptencoding cp1250/scriptencoding iso8859-2/' -e 's/" Original translations/" Generated from menu_slovak_slovak_republic.1250.vim, DO NOT EDIT/' > menu_sk_sk.iso_8859-2.vim
```
k-takata: The last LF is missing.

adaext: > The last LF is missing.

Ok, I've added it. Thanks.

brammool:
Thanks, I'll include it.  I prefer using "iso-8859" instead of "iso8859", since it's more similar to the file name.

PR #7861, opened 2021-02-17
Update French translation of GUI menus

gdupras:
Note: I did not update the file type translations.

@dblanchet (previous modification)

@adrienbeau (maintainer, not sure if this is the right username)

PR #10708, opened 2022-07-12
WIP: Update menu translations for Francais

adaext:
Developers who know the language are now needed to help complete this PR.  The menu translation items template is generated by https://github.com/adaext/vim-menutrans-helper. I've scanned all the menu items and created a fairly completed template. It haven't been updated for years, many new items would be added except items in`menu.vim`.

### DIFF
--- a/runtime/lang/menu_fr_fr.latin1.vim
+++ b/runtime/lang/menu_fr_fr.latin1.vim
@@ -1,486 +1,1267 @@
-" Menu Translations:	Francais
-" Maintainer:		Adrien Beau <version.francaise@free.fr>
-" First Version:	Francois Thunus <thunus@systran.fr>
-" Contributors:		David Blanchet <david.blanchet@free.fr>
-" 			Gabriel Dupras <gabrieldupras@gmail.com>
-" Last Change:		2021 Feb 16
+" Menu Translations:	  Francais
+" Maintainer:           Ada (Haowen) Yu <me@yuhaowen.com>
+" Previous Maintainer:  Adrien Beau <version.francaise@free.fr>
+" Contributors:         Francois Thunus <thunus@systran.fr>, David Blanchet <david.blanchet@free.fr>, Gabriel Dupras <gabrieldupras@gmail.com>
+" Last Change:          2022 July 10
 " Original translations
+"
+" Generated with the scripts from:
+"
+"       https://github.com/adaext/vim-menutrans-helper
 
 " Quit when menu translations have already been done.
+
 if exists("did_menu_trans")
   finish
 endif
 let did_menu_trans = 1
-let s:keepcpo= &cpo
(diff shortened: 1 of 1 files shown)


adaext:
> What should be reviewed? Everything? Only the items that say "TRANSLATION\ MISSING"?

Everything. Could you please kindly download the file, modify that "TRANSLATION\ MISSING" to translations, then send the final result to me? Please check the translated items at the same time, it may have some errors. Since you can't commit in my PR, it's also inconvenient for me to add the comments into the file one by one.

vds2212:
My mother tong is French. If in the future you need someone to translate into French I'll be glad to do so.

chrisbra:
If you could review the translations and update to latest master, that would be appreciated.

vds2212:
It seems to me that there is an encoding problem.
The new translation appear odd in the GUI after I switch to fr_FR
```
set langmenu=fr_FR
let $LANG = 'fr_FR'
source $VIMRUNTIME/delmenu.vim
source $VIMRUNTIME/menu.vim
```

image

Apart from that:
- I don't see in the new version some real new translations
- When I use the Vim menu I don't find menu that are not translated into French (but maybe some are hidden)

gdupras:
@adaext has submitted a similar PR for every language. See these search results:

https://github.com/vim/vim/pulls?q=is%3Apr+is%3Aopen+WIP%3A+Update+menu+translations+for

I’ve only skimmed the changes and as far as I can tell the only real difference is the organization of the code. The translations are not updated.

To me it seems @adaext wants us to use their template they’ve created for GUI translations but they’ve not really explained why we should.

It’s a lot of work to review so many changed lines but if the community thinks it’s worth it to switch to using this template then I could take a closer look at the translations.

By the way I’m the last person to have reviewed the translations, a couple of years ago: https://github.com/vim/vim/pull/7861

runtime/lang/menu_fr_fr.latin1.vim
```diff
@@ -1,486 +1,1267 @@
...
-if &enc != "cp1252" && &enc != "iso-8859-15" && &enc != "iso-8859-1"
-  scriptencoding latin1
+scriptencoding latin1
+
+" Help menu
+menutrans &Help &Aide
+" Help menuitems and dialog {{{1
+menutrans &Overview<Tab><F1> &Sommaire<Tab><F1>
+menutrans &User\ Manual &Manuel\ utilisateur
+menutrans &How-to\ Links &T�ches\ courantes
```
k-takata, line 28: `scriptencoding` is set to "latin1", but actually this file is saved with UTF-8 with many U+FFFD marks (replacement character), which means that this file is broken.

chrisbra:
Thanks all, I guess we can close it then as it currently is not that useful.

PR #10727, opened 2022-07-12
WIP: Update menu translations for Traditional Chinese

adaext:
Developers who know the language are now needed to help complete this PR.  The menu translation items template is generated by https://github.com/adaext/vim-menutrans-helper. I've scanned all the menu items and created a fairly completed template. It haven't been updated for years, many new items would be added except items in`menu.vim`.

### DIFF
--- a/runtime/lang/menu_zh_tw.utf-8.vim
+++ b/runtime/lang/menu_zh_tw.utf-8.vim
@@ -1,292 +1,1266 @@
-" Menu Translations:	Traditional Chinese
-" Translated By:	Hung-Te Lin	<piaip@csie.ntu.edu.tw>
-" Last Change:		2012 May 01
+" Menu Translations:	  Traditional Chinese
+" Maintainer:           Ada (Haowen) Yu <me@yuhaowen.com>
+" Previous Maintainer:  Hung-Te Lin	<piaip@csie.ntu.edu.tw>
+" Last Change:          2022 July 10
 " Original translations
+"
+" Generated with the scripts from:
+"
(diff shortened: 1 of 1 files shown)


ychin:
I just stumbled across this, and the Traditional Chinese translations are looking pretty sad :(. I don't mind helping out in getting some translations in (but not sure if I want to be the maintainer as I don't use Chinese daily these days as primary work in English), but are we hoping to get all the untranslated ones in before committing? I counted ~900 missing translations but a lot of those are file types (e.g. GNU) which don't usually need translations. I wonder if we can modify https://github.com/adaext/vim-menutrans-helper to take those into account @adaext ?

adaext:
> I just stumbled across this, and the Traditional Chinese translations are looking pretty sad :(. I don't mind helping out in getting some translations in (but not sure if I want to be the maintainer as I don't use Chinese daily these days as I primarily work in English), but are we hoping to get all the untranslated ones in before committing? I counted ~900 missing translations but a lot of those are file types (e.g. GNU) which don't usually need translations. I wonder if we can modify https://github.com/adaext/vim-menutrans-helper to take those into account and not count them as needing translations @adaext ?

Thanks for your kind help! Sure, we don't need to translate all these items and I commented them in the translation template. You could first commit the items' translations uncommented in the template. Then if you find other items also need to be translated, you could uncomment, translate and commit again. So that it would be clear to review.

chrisbra:
This seems to be maschine translated with some errors. I am not including this. Closing.

PR #10733, opened 2022-07-13
Use LineNr and CursorLineNr on folded line number column, and use Cur…

ghost:
This is a modification of the Folded line highlight rules:
- use of LineNr and CursorLineNr on folded line number column
- use of CursorLine on folded line

This enables a more consistent look and feel when extensively using folds (e.g. when displaying a tree structure).

(edit)
I also added the display of the sign of the first folded line.
Still looking for a way to make all this optional.

### DIFF
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -1087,6 +1087,9 @@ fold_line(
     int		ri;
     int		line_highlight;
     int		linenr_highlight;
+    int		sign_highlight;
+    sign_attrs_T sattr;
+    char_u	*sign_text;
 
     // Build the fold line:
     // 1. Add the cmdwin_type for the command-line window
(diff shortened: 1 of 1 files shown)


brammool:
How about a screenshot to show the difference?
I wonder if this will have backwards compatibility problems.

You can't declare a variable halfway a block in ANSI C, which we use.

There should be a test for this.  Perhaps an existing test will fail?  I just triggered CI.

Zhern:
image

In both instances I used `fdm=marker` and I focus on line 3 (`b`).
On the left is the modified version: the line 3 highlight is different from line 4 (also a fold) because it is the current one, and the current column number line (in blue) is using the CursorLineNR highlight pattern.
On the right is the standard version: the line 3 highlight is the same as line 4 (both are folds) even if I am focusing on line 3, and the current column number line is not using the CursorLineNR highlight pattern.

I admit that it is not possible to achieve the old highlight behavior with my version, maybe would it be preferable to use new highlight pattern names like `FoldedCursorLineNR` and `FoldedCursorLine`?

Zhern:
I also added the display of the sign of the first folded line: on the left is the new behavior, on the right the standard version.

- Not folded:
image
- Folded:
image

Zhern:
However I am not sure how to make a test for this fold-highlight-sign stuff, I will look at existing ones when I have time.

brammool:
There is a failing test that seems related to the change.  Also perhaps a hint on how to test the change.

Zhern:
I don't think I will have time to continue this. I close the PR and will perhaps reopen it later when I have progressed.

PR #10736, opened 2022-07-14
Correct name of third argument of inputdialog() in builtin function list

zeertzjq:


brammool:
I'll include it, thanks.

PR #10744, opened 2022-07-16
Fix a typo in tutor.zh_cn.utf-8

alick:
Thanks 武岳 <wuyue983163 at 163 dot com> for reporting!


brammool:
I'll include it, thanks.

I notice that tutor.zh.big5 is outdated.  Perhaps we can drop it?  Or can someone update it?

alick:
Thanks @brammool !

zh.big5 will be only useful on Traditional Chinese environment with the legacy encoding (Windows, very old Linux distro) IIUC. Someone who uses Traditional Chinese daily may have better knowledge :)

PR #10747, opened 2022-07-17
feature: tab closing with confirming

regomne:


brammool:
This is not backwards compatible.  Is this fixing a bug?

regomne:
From the perspective of UI experience, I think if a user selects "Close Tab" in the menu, then he must want to close the tab, no matter whether it is currently modified or not. Therefore, I think it is more friendly to pop up a confirm dialog box at this time than to directly prompt that it cannot be closed.
I know this is an incompatible change and would like to discuss the feasibility of this change.

vimpostor:
The user can just use `set confirm`.
We should have consistent behaviour for all operations, not hardcode `confirm` for just tabclosing.

I tend to agree that `set confirm` by default would be a more sane default, but indeed it would be a breaking change.

brammool:
In this case being backwards compatible is more relevant than introducing behavior which can be what the user wanted, or perhaps not.
Therefore I prefer not to make this change.

PR #10749, opened 2022-07-17
doc: fix bad example markers

c4rlo:
Fix some missing or bad example markers (`>`, `<`) that were interfering with correct syntax highlighting for subsequent `*`-enclosed tag anchors.


brammool:
I'll include it, thanks.

PR #10752, opened 2022-07-18
Check `has('patch-x.y.z')` more strictly

### LINKED ISSUE #10751: bug: 'has()' return 1 for invalid patch string
### Steps to reproduce

echo has('patch-hogera')

### Expected behaviour

return value as 0.

### Version of Vim

9.0.0055

### Environment

Windows 10 Pro (64bit)
gVim

### Logs and stack traces

nothing

### Note

Currently return value as 1.

When version 8.2.1287, return 0

k-takata:
Close #10751

Since 9.0.0010, `has('patch-X')` returned 1.

The following patterns also returned 1:
* `patchX`
* `patch0X`
* `patch-9.0.0X`
* `patch-9X0X0`
* `patch-9_0_0`

Also there was a potential issue that `has('patch-10.0.0')` will not
return 1, even after Vim 10 will be released in the future.

This PR fixes them.

As a side effect, `has('patch-09.00.0')` is now allowed.


k-takata:
But, there are still remaining issues:

* Invalid patterns like `patch-9X0X0` are still accepted.
* `patch-10.0.0` doesn't work in the future when Vim 10 is released.

I updated my patch.

brammool:
> But, there are still remaining issues:
> 
> * Invalid patterns like `patch-9X0X0` are still accepted.
> * `patch-10.0.0` doesn't work in the future when Vim 10 is released.
> 
> I updated my patch.

Thanks.  I haven't checked,  but patch-9.0.0001 should work.

-- 
From "know your smileys":
 [:-)	Frankenstein's monster

k-takata:
Yes, the correct pattern like "patch-9.0.0001" works as expected regardless of this PR.

PR #9607, opened 2022-01-23
Add detection for various filetypes

dundargoc:

PR #10755, opened 2022-07-18
fix: recognize *.tfvars as terraform-vars, not terraform

radeksimko:
`*.tfvars` entry was initially introduced as part of https://github.com/vim/vim/pull/9607

In the context of Terraform, it's acceptable to _highlight_ both `*.tf` and `*.tfvars` the same, if the highlighting engine treats them as HCL (i.e. doesn't imply any particular keywords). It is however impossible to provide relevant autocompletion or any other IntelliSense via the language server.

See https://github.com/hashicorp/terraform-ls/blob/main/docs/language-clients.md#language-ids

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1982,8 +1982,8 @@ au BufRead,BufNewFile *.ttl
 " Terminfo
 au BufNewFile,BufRead *.ti			setf terminfo
 
-" Terraform
-au BufRead,BufNewFile *.tfvars			setf terraform
+" Terraform variables
+au BufRead,BufNewFile *.tfvars			setf terraform-vars
 
 " TeX
 au BufNewFile,BufRead *.latex,*.sty,*.dtx,*.ltx,*.bbl	setf tex
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -547,7 +547,7 @@ let s:filename_checks = {
     \ 'template': ['file.tmpl'],
     \ 'teraterm': ['file.ttl'],
     \ 'terminfo': ['file.ti'],
-    \ 'terraform': ['file.tfvars'],
+    \ 'terraform-vars': ['file.tfvars'],
     \ 'tex': ['file.latex', 'file.sty', 'file.dtx', 'file.ltx', 'file.bbl'],
     \ 'texinfo': ['file.texinfo', 'file.texi', 'file.txi'],
     \ 'texmf': ['texmf.cnf'],


brammool:
Are we going to get terraform and terraform-vars runtime files?  If the highlighting is the same, the terraform-vars syntax file could include the terraform syntax file.

radeksimko:
@brammool 

> Are we going to get terraform and terraform-vars runtime files? If the highlighting is the same, the terraform-vars syntax file could include the terraform syntax file.

I'm not sure what exactly do you mean by _runtime files_?

For the highlighting - this depends on how the highlighting grammar is written. If it expects any particular names of blocks or attributes, then it's schema-specific - in which case this should be separate for `terraform` and separate for `terraform-vars`, since they don't have the same schema (e.g. `resource "foo" "bar" {}` is **not** valid within `*.tfvars`).

If the grammar is simple enough and does _not_ care about names, then I'd call that HCL grammar (not Terraform) and using such a grammar is totally acceptable and correct for both `terraform`/`*.tf` and `terraform-vars`/`*.tfvars` as well as `*.hcl`.

Here is an example of two different grammars:

 - https://github.com/hashicorp/syntax/blob/main/syntaxes/hcl.tmGrammar.json
 - https://github.com/hashicorp/syntax/blob/9d47edcacc7fb5918459196cef5cf237b2146372/syntaxes/terraform.tmGrammar.json#L93

brammool:
> > Are we going to get terraform and terraform-vars runtime files? If the highlighting is the same, the terraform-vars syntax file could include the terraform syntax file.
> 
> I'm not sure what exactly do you mean by _runtime files_?
> 
> For the highlighting - this depends on how the highlighting grammar is written. If it expects any particular names of blocks or attributes, then it's schema-specific - in which case this should be separate for `terraform` and separate for `terraform-vars`, since they don't have the same schema (e.g. `resource "foo" "bar" {}` is **not** valid within `*.tfvars`).

Well, you mentioned needing a different filetype name, thus I assume you
have different settings for each filetype.  These would then go in a
filetype plugin.

> If the grammar is simple enough and does _not_ care about names, then
> I'd call that HCL grammar (not Terraform) and using such a grammar is
> totally acceptable and correct for both `terraform`/`*.tf` and
> `terraform-vars`/`*.tfvars` as well as `*.hcl`.
> 
> Here is an example of two different grammars:
> 
>  - https://github.com/hashicorp/syntax/blob/main/syntaxes/hcl.tmGrammar.json
>  - https://github.com/hashicorp/syntax/blob/9d47edcacc7fb5918459196cef5cf237b2146372/syntaxes/terraform.tmGrammar.json#L93

That is up to whoever wants to write the syntax file.

Another runtime file is for automatic indenting.

-- 
I wish there was a knob on the TV to turn up the intelligence.
There's a knob called "brightness", but it doesn't seem to work. 

radeksimko:
> Well, you mentioned needing a different filetype name, thus I assume you
have different settings for each filetype.  These would then go in a
filetype plugin.

Please bear in mind I'm not as familiar with the vim terminology myself 😅 , I just work on the (Terraform) language server and know that we need different language IDs there to be able to provide relevant completion, hover, go-to-definition and other IntelliSense which a language server typically provides.

For the settings, I guess it depends on what kind of settings do you mean - formatting (which is probably what you mean by "automatic indenting") would make sense to potentially configure separately for each language ID, but in practice there is nothing to configure, other than turning it on or off. Both `terraform`/`*.tf` and `terraform-vars`/`*.tfvars` are practically formatted using the same logic under the hood and the same rules, which are opinionated and _not_ configurable (on purpose).

PR #10760, opened 2022-07-20
Do not use utfc_ptr2char_len() when printing until NUL

zeertzjq:
I haven't found a way to test this change.


zeertzjq:
Also use ternary operators to reduce a bit of indent.

PR #10773, opened 2022-07-23
Add runtime files for bitbake

gpanders:
These are (very slightly) modified versions of the runtime files
maintained by the [upstream bitbake project][1]. They are licensed MIT
and I got the ok from the [bitbake developer mailing list][2] to upstream
them to Vim.

[1]: https://github.com/openembedded/bitbake/tree/master/contrib/vim
[2]: https://lists.openembedded.org/g/bitbake-devel/message/13820


brammool:
The indent file bails out when BitbakeIndent exits.  However, it hasn't set 'indentexpr' yet.  It would bail out after the options are set, before the functions are defined (which are shared between buffers).

brammool:
Also, it should not define global functions, unless the name is unique.  GetPythonIndent() is used elsewhere.

gpanders:
I'll take care of both of those issues. The bitbake indent file uses a slightly modified version of `GetPythonIndent`, though most of the code is identical to the one in `indent/python.vim`. So there are a few options:

1. Keep the (slightly modified) duplicated function, but make it script-local
2. Use the existing function from indent/python.vim without the bitbake-specific modifications. Might result in minor indenting quirks for bitbake files, but removes the duplicated code
3. Add the modifications to the indent/python.vim file (namely, the `s:is_bb_python_func_def` function and associated checks). This couples the bitbake syntax definitions to the Python indent file, which doesn't feel very clean to me.

I lean toward option 2 but would like to hear your thoughts.

brammool:
What are the "bitbake-specific modifications" ?   It is a large function, duplicating it isn't great.
Perhaps we can make GetPythonIndent() take an argument for what is different in bitbake.

gpanders:
They define this function:

    function s:is_bb_python_func_def(lnum)
        let stack = synstack(a:lnum, 1)
        if len(stack) == 0
            return 0
        endif


        let top = synIDattr(stack[0], "name")
        echo top


        return synIDattr(stack[0], "name") == "bbPyFuncDef"
    endfunction

and use it in GetPythonIndent in two places. Example:

    if parlnum > 0
      " We may have found the opening brace of a BitBake Python task, e.g. 'python do_task {'
      " If so, ignore it here - it will be handled later.
      if s:is_bb_python_func_def(parlnum)
        let parlnum = 0
        let plindent = indent(plnum)
        let plnumstart = plnum
      else
        let plindent = indent(parlnum)
        let plnumstart = parlnum
      endif
    else
      let plindent = indent(plnum)
      let plnumstart = plnum
    endif

gpanders:
It looks like it's very specific to just a few syntax items. The default Python indent function likely works well enough as is.

brammool:
> They define this function:
> 
>     function s:is_bb_python_func_def(lnum)
>         let stack = synstack(a:lnum, 1)
>         if len(stack) == 0
>             return 0
>         endif
> 
> 
>         let top = synIDattr(stack[0], "name")
>         echo top
> 
> 
>         return synIDattr(stack[0], "name") == "bbPyFuncDef"
>     endfunction
> 
> and use it in GetPythonIndent in two places. Example:
> 
>     if parlnum > 0
>       " We may have found the opening brace of a BitBake Python task, e.g. 'python do_task {'
>       " If so, ignore it here - it will be handled later.
>       if s:is_bb_python_func_def(parlnum)
>         let parlnum = 0
>         let plindent = indent(plnum)
>         let plnumstart = plnum
>       else
>         let plindent = indent(parlnum)
>         let plnumstart = parlnum
>       endif
>     else
>       let plindent = indent(plnum)
>       let plnumstart = plnum
>     endif

How about moving GetPythonIndent() to an autoload script,
and adding an argument that is a funcref that for Python is empty and
for BitBake is s:is_bb_python_func_def().
This would add some BitBake specific lines, but moving it out of
indent/python.vim makes clear it's used for more than one filetype.

-- 
Corduroy pillows: They're making headlines!

gpanders:
Sounds good, I’ll take care of that soon.

gpanders:
>How about moving GetPythonIndent() to an autoload script,
>and adding an argument that is a funcref that for Python is empty and
>for BitBake is s:is_bb_python_func_def().
>This would add some BitBake specific lines, but moving it out of
>indent/python.vim makes clear it's used for more than one filetype.

All right done, let me know if there are any other changes you'd like to see.

brammool:
> >How about moving GetPythonIndent() to an autoload script,
> >and adding an argument that is a funcref that for Python is empty and
> >for BitBake is s:is_bb_python_func_def().
> >This would add some BitBake specific lines, but moving it out of
> >indent/python.vim makes clear it's used for more than one filetype.
> 
> All right done, let me know if there are any other changes you'd like
> to see.

I haven't tried it, but it it looks like the mechanism is OK.

Optionally you could add an indent test in runtime/indent/testdir.

Note that I removed the check for 'compatible' from the syntax file.
I don't think that is needed and it breaks the script that I use to
check runtime files.

Oh, the runtime/autoload/python.vim file uses line continuation, it
needs to set the 'cpo' option.

-- 
Nobody will ever need more than 640 kB RAM.
		-- Bill Gates, 1983
Windows 98 requires 16 MB RAM.
		-- Bill Gates, 1999
Logical conclusion: Nobody will ever need Windows 98.

gpanders:
>Oh, the runtime/autoload/python.vim file uses line continuation, it
>needs to set the 'cpo' option.

Done, also added an indent test for bitbake.

brammool:
Thanks, I'll include it now.  Please check after the next runtime files update.

Issue #10694, opened 2022-07-11
syntax/make.vim: highlight grouped targets


matteocoder:
GNU Make 4.3 introduced 'Grouped targets', which allow to specify that two or more targets are produced by a single recipe simultaneously. Currently, the make syntax file does not support this feature, so I modified the regex to correctly detect it.
The following patch works for me on Vim 9.0.44:
```diff
--- "/c/Program Files/Vim/vim90/syntax/make.vim"	2022-07-02 00:01:18.000000000 +0200
+++ make.vim	2022-07-03 21:32:43.424931600 +0200
@@ -45,11 +45,11 @@
 syn match makeImplicit		"^\.[A-Za-z0-9_./\t -]\+\s*:[^=]"me=e-2
 
 syn region makeTarget transparent matchgroup=makeTarget
-	\ start="^[~A-Za-z0-9_./$()%-][A-Za-z0-9_./\t $()%-]*:\{1,2}[^:=]"rs=e-1
+	\ start="^[~A-Za-z0-9_./$()%-][A-Za-z0-9_./\t $()%-]*\(&\)\?:\{1,2}[^:=]"rs=e-1
 	\ end=";"re=e-1,me=e-1 end="[^\\]$"
 	\ keepend contains=makeIdent,makeSpecTarget,makeNextLine,makeComment,makeDString
 	\ skipnl nextGroup=makeCommands
-syn match makeTarget		"^[~A-Za-z0-9_./$()%*@-][A-Za-z0-9_./\t $()%*@-]*::\=\s*$"
+syn match makeTarget		"^[~A-Za-z0-9_./$()%*@-][A-Za-z0-9_./\t $()%*@-]*\(&\)\?::\=\s*$"
 	\ contains=makeIdent,makeSpecTarget,makeComment
 	\ skipnl nextgroup=makeCommands,makeCommandError
 
```
I tried contacting the maintainer at the address provided in the file,
but so far he has not responded to my mail.

jamessan:
@rohieb

rohieb:
Mh yes sorry, my inbox is an unstructured mess… But the patch looks good to me.

rohieb:
Ah, they also added the `&::` syntax, see https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html at the bottom. Can you add another `:\?` to your patch?

matteocoder:
@rohieb Do you mean like this?
```diff
[...]

Issue #6943, opened 2020-09-12
Bug in makefile syntax highlighting


haizaar:
**Describe the bug**
If [target-specific variable](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html) definition contains a semi-colon, e.g. a shell command, the highlighting breaks:
![image](https://user-images.githubusercontent.com/58201/92997967-6e067480-f55a-11ea-9248-c89488b5ac08.png)

Even more evident in Zenburn color scheme:
![image](https://user-images.githubusercontent.com/58201/92998038-d7868300-f55a-11ea-8ef3-0fde9a413e8d.png)

I think it treats semicolon as a definition breaker and disregards the fact that we got quotes surrounding it.

**To Reproduce**
Here is the makefile in question:
```makefile
foo:
    echo "$(VAR)"
    echo 1
    echo 2

bar: VAR = $(shell python -c "import sys; print(sys)")
bar:
    echo "$(VAR)"
    echo 2
    echo 3
```

**Environment (please complete the following information):**
 - Vim version 8.1.2269
 - OS: Ubuntu 20.04
 - Terminal: GNOME Terminal

brammool:
Is that valid Makefile syntax?  "bar: VAR = " seems to mix a dependency with an assignment.  ah, that's a GNU make extension.
I haven't seen that used anywhere.
Anyway, this is for @rohieb

haizaar:
@rohieb Any chance you'll have time to look at it?

rohieb:
Huh, I'm learning much more about interesting features of the different make implementations…

Anyway, the culprit is [line 49 of make.vim](https://github.com/vim/vim/blob/59d8e56e048eb5d384649284fb35363931fc3697/runtime/syntax/make.vim#L49) which defines `end=";"re=e-1,me=e-1` – and I don't know enough about different make implementations to know why this is there… it's certainly not a GNU extension 😁 I think I would just remove it and see if anyone complains.

jamessan:
> it's certainly not a GNU extension grin I think I would just remove it and see if anyone complains.

Maybe it's trying to handle defining part of the recipe on the same line as the target -- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html#tag_20_76_13_04

PR #10774, opened 2022-07-23
runtime/syntax/make.vim updates

rohieb:
* highlight grouped targets (solves #10694, thanks to @matteocoder)
* add ONESHELL as a special command (thanks to @averms)
* hopefully solve #6943


brammool:
I'll include it, thanks.

PR #10776, opened 2022-07-24
po/Makefile: Avoid race over LINGUAS file

rpurdie:
The creation of the LINGUAS file is duplicated for each desktop file
which can lead the commands to race against each other. One target might
remove it before another has been able to use it. Rework the makefile to
avoid this as the expense of leaving the file on disk.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>

### DIFF
--- a/src/po/Makefile
+++ b/src/po/Makefile
@@ -213,17 +213,16 @@ $(PACKAGE).pot: $(PO_INPUTLIST) $(PO_VIM_INPUTLIST)
 	# Delete the temporary files
 	rm *.js
 
-vim.desktop: vim.desktop.in $(POFILES)
+LINGUAS:
 	echo $(LANGUAGES) | tr " " "\n" |sed -e '/\./d' | sort > LINGUAS
+
+vim.desktop: vim.desktop.in $(POFILES) LINGUAS
 	$(MSGFMT) --desktop -d . --template vim.desktop.in -o tmp_vim.desktop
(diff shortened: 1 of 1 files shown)


brammool:
Well, after this change it leaves the LINGUAS file behind.  The documentation isn't very clear, but it seems that the file can't be given another name, so that we would be able to use a different file in each block.
Any way to delete LINGUAS when done?

dpelle:
@brammool wrote:

> Well, after this change it leaves the LINGUAS file behind.

`LINGUAS` is removed in the `clean` target. So I think the change is fine.

brammool:
> @brammool wrote:
> 
> > Well, after this change it leaves the LINGUAS file behind.
> 
> `LINGUAS` is removed in the `clean` target. So I think the change is fine.

I normally don't type "make clean", as it also deltes the *.mo files.

Perhaps there is a way to make the targets run sequential.
Perhaps by making gvim.desktop depend on vim.desktop.
Or by joining the two blocks.

It's a bit weird that a tool has an input file for which the name can't
be specified.

-- 
Send $25.00 for handy leaflet on how to make money by selling leaflets

brammool:
Can you check if the problem still exists after the recent change to src/po/Makefile ?

rpurdie:
We haven't seen the issue again since that change so I think we're good, thanks!

PR #10777, opened 2022-07-24
configure.ac: Fix create_timer solaris test for cross compiling

rpurdie:
A runtime test was added for create_timer however this meant cross compiling
would no longer work. Allow a cache value to be specified to allow cross
compiling again.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -3807,7 +3807,7 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM(
 dnl Check for timer_create. It probably requires the 'rt' library.
 dnl Run the program to find out if timer_create(CLOCK_MONOTONIC) actually
 dnl works, on Solaris timer_create() exists but fails at runtime.
-AC_MSG_CHECKING([for timer_create])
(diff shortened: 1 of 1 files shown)


shak5009:
Still having issues with vim timer  at 9.0.0065+

checking for timer_create without -lrt... configure: error: cross-compiling: please set 'vim_cv_timer_create'

brammool:
> Still having issues with vim timer  at 9.0.0065+
> 
> checking for timer_create without -lrt... configure: error:
> cross-compiling: please set 'vim_cv_timer_create'

The docs for AC_RUN_IFELSE are a bit unclear, but it looks like it
assumes cross-compiling when it cannot link the program.  I cannot think
of a way to make a difference, failing to compile because timer_create
is not available and actually cross-compiling cannot be distinguished.
On my system I did not manage to make it give the cross-compiling
message even when passing a non-existing library.

I think we can work around this by adjusting the message to say "if
cross-compiling, then...".  And use AC_MSG_WARN instead of AC_MSG_ERROR,
so that configure continues.

-- 
195. Your cat has its own home page.

PR #10778, opened 2022-07-24
docs: fix a couple of tag links

c4rlo:


brammool:
The "statusline" tag does refer to the 'statusline' option, as the example shows, it sets that option.
But the tag should have single quotes instead of bars.

For the change in pi_netrw.txt: @cecamp

c4rlo:
Sorry, you're right about 'statusline'.

I see these were now fixed in b529cfbd04c02e31cfa88f2c8d88b5ff532d4f7d.

PR #10783, opened 2022-07-25
Correct filetype detection for bsdl

arp242:
In PR #5945 it's added as "*.bsd", but in commit 0fc1288ae it got added
as "*bsd", without a dot.

I guess that's probably an error? So correct that, as it highlights
files named "freebsd", "openbsd", etc. wrong.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -259,7 +259,7 @@ au BufNewFile,BufRead *.bb,*.bbappend,*.bbclass,*/build/conf/*.conf,*/meta{-*,}/
 au BufNewFile,BufRead */etc/blkid.tab,*/etc/blkid.tab.old   setf xml
 
 " BSDL
-au BufNewFile,BufRead *bsd,*.bsdl		setf bsdl
+au BufNewFile,BufRead *.bsd,*.bsdl		setf bsdl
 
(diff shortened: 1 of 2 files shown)


brammool:
The test for this also uses the files "bsd" and "some-bsd".  It looks like the missing dot was intentional.
But it catches too many names.  Perhaps "bsd" would be OK and a file ending in "-bsd"?

arp242:
Oh right; looks like the test was added in 10a1a534d1 later by someone else who added a bunch of filetype tests; I think they just "copied" the typo.

I tried looking up if these files are being used, but I can't really find all that much info about this BSDL thing beyond an ungodly amount of SEO linkspam, but as near as I can determine from the git history you "applied" the small patch by just editing the file, with this small typo.

I just changed the test now.

CC: @danielkho

arp242:
CI failure is unrelated to this patch:

> curl: (35) schannel: next InitializeSecurityContext failed: SEC_E_ILLEGAL_MESSAGE (0x80090326) - This error usually occurs when a fatal SSL/TLS alert is received (e.g. handshake failed). More detail may be available in the Windows System event log.

brammool:
OK, if someone really wants "bsd" and "some-bsd" to be recognized as bsdl we'll await comments.

PR #10787, opened 2022-07-25
Fixed a few compiler errors in MSVC. (size_t typecast related)

wbpayne22902:
I added a typecast from size_t to int in file charset.c at line 1095 and file drawline.c at line 1527.
This fixes a few C4627 errors when compiling with Microsoft Visual C++.


brammool:
I included the (identical) patch from Mike Williams

PR #10788, opened 2022-07-25
add runtime/ftplugin/desktop.vim

### LINKED ISSUE #2495: Add correct 'commentstring' for *.desktop files
Add correct `'commentstring'` to `runtime/ftplugin/desktop.vim` accordingly [freedesktop specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html).
```vim
setlocal commentstring=#%s
```

e-kwsm:
close #2495


brammool:
Thanks.  Setting and restoring 'cpo' isn't needed here, those lines can be removed.  I'll do that and include it.

PR #10789, opened 2022-07-26
Add a test for a crash fixed by patch 8.1.1424

zeertzjq:
An expression mapping can also reproduce the crash.


zeertzjq:
Hangs on MS-Windows

zeertzjq:
Oh it needs `CheckNotGui`

PR #10795, opened 2022-07-26
Fix compiler warnings for assigning size_t to int

mrdubya:
FWIW VIM is now at 60% of calls to STRLEN() being cast to int.  Just sayin'


mrdubya:
Ah, already reported. Apologies..

brammool:
Mike Williams wrote:

> FWIW VIM is now at 60% of calls to STRLEN() being cast to int.  Just sayin'

We could add STRLEN_INT() but it won't really help.

-- 
134. You consider bandwidth to be more important than carats.

PR #10801, opened 2022-07-27
Modifiable v:completed_item

Shougo:
I think `v:completed_item` should be modifiable.

Some plugins(ddc.vim, nvim-cmp, coc.nvim etc) plugins emulates native popup completion system, but it cannot support `CompleteDone` autocmd behavior.
Because `v:completed_item` is read only.

Related issue: https://github.com/neoclide/coc.nvim/pull/3862


Shougo:
Thank you for the merge.

PR #10818, opened 2022-07-30
Fix ffs_filerray_cur type to not overflow in directories with more than 255 entries

jdrouhard:
`ffs_filearray_cur` is used as an index up to `ffs_filearray_size` but the types don't match. Specifically, `ffs_filearray_cur` is currently a `char_u` (while `ffs_filearray_size` and index variable `i` are ints), so it overflows when searching a directory that contains more than 255 entries. This is technically undefined behavior, and likely causes the find loop at around line 971 to do more work than necessary.

Add this assert:
```diff
diff --git a/src/findfile.c b/src/findfile.c
index 881eef156..9e59346e9 100644
--- a/src/findfile.c
+++ b/src/findfile.c
@@ -1046,6 +1046,7 @@ vim_findfile(void *search_ctx_arg)
 #endif

                                // push dir to examine rest of subdirs later
+                               assert(i < UCHAR_MAX);
                                stackp->ffs_filearray_cur = i + 1;
                                ff_push(search_ctx, stackp);
```

Test case:
```sh
$ mkdir test; cd test
$ for i in {1..300}; do mkdir -p $i; touch $i/$i.h; done
$ cat << EOF > test.h
#include "99.h"
EOF
$ cat << EOF > test.in
:set path=\$PWD/**
Wgf
EOF
$ vim -u NONE test.h -s test.in
```

The fix is simply change `ffs_filearray_cur` to an int to match the loop index `i` and `ffs_filearray_size`.

Related: https://github.com/neovim/neovim/issues/13284

### DIFF
--- a/src/findfile.c
+++ b/src/findfile.c
@@ -75,7 +75,7 @@ typedef struct ff_stack
     // of wc_part
     char_u		**ffs_filearray;
     int			ffs_filearray_size;
-    char_u		ffs_filearray_cur;   // needed for partly handled dirs
+    int			ffs_filearray_cur;   // needed for partly handled dirs
 
     // to store status of partly handled directories
     // 0: we work on this directory for the first time
@@ -1046,6 +1046,7 @@ vim_findfile(void *search_ctx_arg)
 #endif
 
 				// push dir to examine rest of subdirs later
+				assert(i < INT_MAX);
 				stackp->ffs_filearray_cur = i + 1;
 				ff_push(search_ctx, stackp);
 
--- a/src/testdir/test_gf.vim
+++ b/src/testdir/test_gf.vim
@@ -250,4 +250,26 @@ func Test_includeexpr_scriptlocal_func()
   bw!
 endfunc
 
+func Test_gf_subdirs_wildcard()
+  let cwd = getcwd()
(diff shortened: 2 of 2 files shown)


brammool:
Thanks for pinpointing this problem. Could you turn the test case into a test? 
Can be added to src/testdir/test_gf.vim

jdrouhard:
> Thanks for pinpointing this problem. Could you turn the test case into a test?
> Can be added to src/testdir/test_gf.vim

Done. I verified the test fails with `assert (i < UCHAR_MAX)` added without the fix. I did add an assert that `i < INT_MAX` since it's still possible for this value to overflow but should be much less likely (since this value is always based on `ffs_filearray_size` which is populated by `expand_wildcards()`).

To be fair, I realized that _unsigned_ values that overflow when adding beyond the type's max value is not undefined in the C standard--it's defined to wrap around to 0--so this isn't actually UB. Though fixing this will still improve performance since the loop won't be searching more than necessary as it shouldn't ever overflow/wrap, now.

brammool:
Thanks.  I'll include the patch, but without that assert the test passes, even though the problem is still there.
I wonder if it's possible to change the test that it would fail (without the assert).

PR #10833, opened 2022-08-01
treat {} like () and [] when indenting lisp

benknoble:
Racket (https://racket-lang.org) treats all 3 as synonymous (with
openers paired correctly with closers): see the section "Reading Pairs
and Lists"[1].

[1]: https://docs.racket-lang.org/reference/reader.html#%28part._parse-pair%29

### DIFF
--- a/src/indent.c
+++ b/src/indent.c
@@ -1975,7 +1975,12 @@ lisp_match(char_u *p)
     int
 get_lisp_indent(void)
 {
-    pos_T	*pos, realpos, paren;
+    typedef struct {
+	int found;
+	pos_T pos;
+    } search;
+    pos_T	*pos, realpos;
+    search	paren, square, curly;
     int		amount;
     char_u	*that;
     colnr_T	col;
@@ -1989,15 +1994,37 @@ get_lisp_indent(void)
     realpos = curwin->w_cursor;
     curwin->w_cursor.col = 0;
 
-    if ((pos = findmatch(NULL, '(')) == NULL)
-	pos = findmatch(NULL, '[');
-    else
+    // Find starting point
+    // Post-condition:
+    // - pos is NULL if no opening bracket is found, OR
+    // - pos points to the "greatest" (by LT_POSP) non-NULL position of an
+    // opening bracket
+    // The temporary pos_T values are to hold the value returned by findmatch,
+    // which is a static global overwritten by subsequent calls to findmatch.
+    paren.found = square.found = curly.found = 0;
+    if ((pos = findmatch(NULL, '(')) != NULL)
     {
-	paren = *pos;
-	pos = findmatch(NULL, '[');
-	if (pos == NULL || LT_POSP(pos, &paren))
-	    pos = &paren;
+	paren.pos = *pos;
+	paren.found = 1;
     }
+    if ((pos = findmatch(NULL, '[')) != NULL)
+    {
+	square.pos = *pos;
+	square.found = 1;
+    }
+    if ((pos = findmatch(NULL, '{')) != NULL)
+    {
+	curly.pos = *pos;
+	curly.found = 1;
+    }
+    pos = NULL;
+    if (paren.found)
+	pos = &paren.pos;
+    if (square.found && (pos == NULL || LT_POSP(pos, &square.pos)))
+	pos = &square.pos;
+    if (curly.found && (pos == NULL || LT_POSP(pos, &curly.pos)))
+	pos = &curly.pos;
+
     if (pos != NULL)
     {
 	// Extra trick: Take the indent of the first previous non-white
@@ -2042,9 +2069,9 @@ get_lisp_indent(void)
 		    if (*that == NUL)
 			break;
 		}
-		if (*that == '(' || *that == '[')
+		if (*that == '(' || *that == '[' || *that == '{')
 		    ++parencount;
-		else if (*that == ')' || *that == ']')
+		else if (*that == ')' || *that == ']' || *that == '}')
 		    --parencount;
 	    }
 	    if (parencount == 0)
@@ -2085,7 +2112,7 @@ get_lisp_indent(void)
 		// (let ((a 1))    instead    (let ((a 1))
 		//   (...))	      of	   (...))
 
-		if (!vi_lisp && (*that == '(' || *that == '[')
+		if (!vi_lisp && (*that == '(' || *that == '[' || *that == '{')
 						      && lisp_match(that + 1))
 		    amount += 2;
 		else
@@ -2112,7 +2139,8 @@ get_lisp_indent(void)
 		    {
 			// test *that != '(' to accommodate first let/do
 			// argument if it is more than one line
-			if (!vi_lisp && *that != '(' && *that != '[')
+			if (!vi_lisp && *that != '(' && *that != '['
+						     && *that != '{')
 			    firsttry++;
 
 			parencount = 0;
@@ -2131,17 +2159,20 @@ get_lisp_indent(void)
 					|| quotecount
 					|| parencount)
 				    && (!((*cts.cts_ptr == '('
-							|| *cts.cts_ptr == '[')
+							|| *cts.cts_ptr == '['
+							|| *cts.cts_ptr == '{')
 					    && !quotecount
 					    && !parencount
 					    && vi_lisp)))
 			    {
 				if (*cts.cts_ptr == '"')
 				    quotecount = !quotecount;
-				if ((*cts.cts_ptr == '(' || *cts.cts_ptr == '[')
+				if ((*cts.cts_ptr == '(' || *cts.cts_ptr == '['
+							 || *cts.cts_ptr == '{')
 							       && !quotecount)
 				    ++parencount;
-				if ((*cts.cts_ptr == ')' || *cts.cts_ptr == ']')
+				if ((*cts.cts_ptr == ')' || *cts.cts_ptr == ']'
+							 || *cts.cts_ptr == '}')
 							       && !quotecount)
 				    --parencount;
 				if (*cts.cts_ptr == '\\'
@@ -2167,7 +2198,7 @@ get_lisp_indent(void)
 	}
     }
     else
-	amount = 0;	// no matching '(' or '[' found, use zero indent
+	amount = 0;	// no matching '(', '[', or '{' found, use zero indent
 
     curwin->w_cursor = realpos;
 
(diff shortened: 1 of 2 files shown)


benknoble:
Looks like initial CI passed, and I need approval to run the rest of the checks.

brammool:
The built-in Lisp support was only done to be Vi compatible.  For Racket indenting should be done in a plugin.

We don't have a separate racket filetype.  Perhaps that's the first thing to take care of, if it is in any way different from original Lisp.

benknoble:
> The built-in Lisp support was only done to be Vi compatible. For Racket indenting should be done in a plugin.

Yet Scheme uses Lisp indenting, probably because the easiest way to indent a Lisp is to use the internal Lisp indenter: https://github.com/vim/vim/blob/0b0ccbbfb014e1c9682b86a7a41ff2837b0b8047/runtime/ftplugin/scheme.vim#L17

Racket descends from Scheme. This is further evidenced by Vim setting the `scheme` filetype for Racket files: https://github.com/vim/vim/blob/0b0ccbbfb014e1c9682b86a7a41ff2837b0b8047/runtime/filetype.vim#L1786

I suggest that if Vim wants to support Racket (via Scheme), which uses `'lisp'` for indentation, then it should support indenting in `{}` correctly.

If it were easy to re-implement and maintain the Lisp indenting algorithm in a plugin (or every plugin that needed it), I would do it. But the shortest path to Lisp indenting is `'lisp'` and `'lispwords'`, and thus the change to `get_lisp_indent` is the easiest way to adjust Lisp-like indentation.

Clojure also uses `[]` and `{}`, though specifically for data-literals (vectors and sets/maps, respectively). Clojure has to work harder with its own indent file, though it falls back on `'lisp'` if `searchpairpos` is not available.

> We don't have a separate racket filetype. Perhaps that's the first thing to take care of, 

There were efforts by me and @dkearns to get this started based on my plugin (https://github.com/benknoble/vim-racket/pull/8, https://github.com/benknoble/vim-racket/issues/7), but this has stalled.

In addition, we would need to discuss how to actually handle Racket filetypes: my plugin builds on the "Racket is a language and an ecosystem of languages" idea to make different `#lang`s different filetypes (they may have different syntax altogether, necessitating different indent and syntax files). Some people appear prefer to have them all be `racket` even when those indent and syntax files don't work well for the particular `#lang`. One suggestion was to include just the Racket filetype and related things to start.

> if it is in any way different from original Lisp.

Well, it's not a Lisp in the sense of Common Lisp. It's a Scheme descendant. The keywords are different; some groupings are different. But it's a Lisp with parentheses and prefix notation all the same. Scheme is also different from "original Lisp" if I understand you correctly to mean Common Lisp, and Scheme is currently using Lisp support—see above.

brammool:
> > The built-in Lisp support was only done to be Vi compatible. For
> > Racket indenting should be done in a plugin.
> 
> Yet Scheme uses Lisp indenting, probably because the easiest way to
> indent a Lisp is to use the internal Lisp indenter:
> https://github.com/vim/vim/blob/0b0ccbbfb014e1c9682b86a7a41ff2837b0b8047/runtime/ftplugin/scheme.vim#L17
> 
> Racket descends from Scheme. This is further evidenced by Vim setting the `scheme` filetype for Racket files: https://github.com/vim/vim/blob/0b0ccbbfb014e1c9682b86a7a41ff2837b0b8047/runtime/filetype.vim#L1786

Sometimes people just use the filetype that comes closest and live with
the incorrect indenting/highlighting.  But we should try to do it
properly, at some point those incorrect things become annoying.

> I suggest that if Vim wants to support Racket (via Scheme), which uses
> `'lisp'` for indentation, then it should support indenting in `{}`
> correctly.

But we don't want to break the good old lisp indenting.  There is the
'lispwords' options.  Perhaps we need a 'lispoptions' option?  I rather
not add yet another option, but we also have 'cinoptions'.

> If it were easy to re-implement and maintain the Lisp indenting
> algorithm in a plugin (or every plugin that needed it), I would do it.
> But the shortest path to Lisp indenting is `'lisp'` and `'lispwords'`,
> and thus the change to `get_lisp_indent` is the easiest way to adjust
> Lisp-like indentation.

As mentioned, we should be able to do this properly.  Not hack something
that works now and will come back to us when someone wants it to work
better.

> Clojure also uses `[]` and `{}`, though specifically for data-literals
> (vectors and sets/maps, respectively). Clojure has to work harder with
> its own indent file, though it falls back on `'lisp'` if
> `searchpairpos` is not available.
> 
> > We don't have a separate racket filetype. Perhaps that's the first
> > thing to take care of, 
> 
> There were efforts by me and @dkearns to get this started based on my plugin (https://github.com/benknoble/vim-racket/pull/8, https://github.com/benknoble/vim-racket/issues/7), but this has stalled.
> 
> In addition, we would need to discuss how to actually handle Racket
> filetypes: my plugin builds on the "Racket is a language and an
> ecosystem of languages" idea to make different `#lang`s different
> filetypes (they may have different syntax altogether, necessitating
> different indent and syntax files). Some people appear prefer to have
> them all be `racket` even when those indent and syntax files don't
> work well for the particular `#lang`. One suggestion was to include
> just the Racket filetype and related things to start.

Is Racket really a filetype or a collection of languages?  Wikipedia
mentions several.  We should use a different filetype name if the
indenting, highlighting or option values need to be different.
> 
> > if it is in any way different from original Lisp.
> 
> Well, it's not a Lisp in the sense of Common Lisp. It's a Scheme
> descendant. The keywords are different; some groupings are different.
> But it's a Lisp with parentheses and prefix notation all the same.
> Scheme is also different from "original Lisp" if I understand you
> correctly to mean Common Lisp, and Scheme is currently using Lisp
> support—see above.

I'm not familiar with these languages and it requires someone actually
using them to say how relevant the differences are.

-- 
232. You start conversations with, "Have you gotten a fiber connection?"

benknoble:
> Sometimes people just use the filetype that comes closest and live with the
> incorrect indenting/highlighting. But we should try to do it properly, at some
> point those incorrect things become annoying.

> But we don't want to break the good old lisp indenting. There is the
> 'lispwords' options. Perhaps we need a 'lispoptions' option? I rather not add
> yet another option, but we also have 'cinoptions'.

> As mentioned, we should be able to do this properly. Not hack something that
> works now and will come back to us when someone wants it to work better.

If you're comfortable with some kind of option to control this, which I had
considered, I can try to make that work (something like `lispoptions` or maybe a
flag in `cpoptions` since it's compatibility-related).

If not, I will probably close this and try to take on the task of writing a
Racket `indentexpr` (which will approximate the `lisp` indenter and its
accounting for `lispwords`, though with fixes for multi-line strings, heredocs,
brackets, `for/fold`, and anything else I find as I go). This can stay scoped to
my plugin, though, since there is no Racket filetype yet. Any advice on how to
tackle this is welcome.

> > > We don't have a separate racket filetype. Perhaps that's the first thing
> > > to take care of.
>
> > There were efforts by me and @dkearns to get this started based on my plugin
> > ([benknoble/vim-racket#8](https://github.com/benknoble/vim-racket/pull/8),
> > [benknoble/vim-racket#7](https://github.com/benknoble/vim-racket/issues/7)),
> > but this has stalled. In addition, we would need to discuss how to actually
> > handle Racket filetypes: my plugin builds on the "Racket is a language and
> > an ecosystem of languages" idea to make different `#lang`s different
> > filetypes (they may have different syntax altogether, necessitating
> > different indent and syntax files). Some people appear prefer to have them
> > all be `racket` even when those indent and syntax files don't work well for
> > the particular `#lang`. One suggestion was to include just the Racket
> > filetype and related things to start.
>
> Is Racket really a filetype or a collection of languages? Wikipedia mentions
> several. We should use a different filetype name if the indenting,
> highlighting or option values need to be different.

Er, both. Racket is a language platform: that there are many `#lang`s that run
on it, capable of interoperating if they choose, with any syntax and semantics
they want (and you can write your own). In that sense, an arbitrary Racket file
may have its own syntax and therefore indent, ftplugin, or syntax files (etc.).

However, one core language on the platform is `#lang racket`: most languages
that want to interoperate expand into some version of this language. This is the
predominant Lisp-like Scheme-descended language, with the prefix-parenthesised
syntax. In some sense it was the first `#lang` and continues to be the most
widely used one (citations needed). In _this_ sense, there is a `racket`
filetype that can establish this syntax, indent, ftplugin, etc.

My plugin makes the decision that `#lang racket` files have filetype `racket`,
and (for example) `#lang scribble` files have filetype `scribble`, while `#lang
datalog` files have filetype `datalog`. These are all languages in the racket
ecosystem with varied syntax; to me this makes the most sense in terms of having
a maintainable ecosystem of Vim support. (Notable exceptions: some languages
modify others, so `#lang at-exp racket` has filetype `racket`, and currently
lacks support for the special features given by `at-exp`; other languages are
`racket` syntax with more things piled on top, so `#lang web-server/insta` is
also filetype `racket`.)

> > > if it is in any way different from original Lisp.
>
> > Well, it's not a Lisp in the sense of Common Lisp. It's a Scheme descendant.
> > The keywords are different; some groupings are different. But it's a Lisp
> > with parentheses and prefix notation all the same. Scheme is also different
> > from "original Lisp" if I understand you correctly to mean Common Lisp, and
> > Scheme is currently using Lisp support—see above.
>
> I'm not familiar with these languages and it requires someone actually using
> them to say how relevant the differences are.

The differences are mostly solved by `lispwords`, though some things remain
imperfect. It is true that creating my own `indentexpr` would allow me to fix
these things for Racket; I just don't relish the thought of implementing the
`lisp` indenting algorithm in Vimscript :) but it looks like that's what I'll be
doing regardless.

brammool:
> > Sometimes people just use the filetype that comes closest and live
> > with the incorrect indenting/highlighting. But we should try to do
> > it properly, at some point those incorrect things become annoying.
> 
> > But we don't want to break the good old lisp indenting. There is the
> > 'lispwords' options. Perhaps we need a 'lispoptions' option? I rather not add
> > yet another option, but we also have 'cinoptions'.
> 
> > As mentioned, we should be able to do this properly. Not hack something that
> > works now and will come back to us when someone wants it to work better.
> 
> If you're comfortable with some kind of option to control this, which
> I had considered, I can try to make that work (something like
> `lispoptions` or maybe a flag in `cpoptions` since it's
> compatibility-related).

When it is about something that is already done by lisp indenting, but
recognizing different parenthesis, then adding 'lispoptions' avoids
replicating the code in a plugin.  But larger changes could perhaps use
the built-in lisp indenting and add on top of that.  This is to avoid
having to change the C code every every little language difference.


-- 
237. You tattoo your email address on your forehead.

PR #10839, opened 2022-08-03
Use no_plugin_maps in Abaqus ftplugin

dkearns:
@costerwi it seems the email address in the headers of your Abaqus runtime files is no longer valid so I'm submitting this here instead.

I noticed that the ftplugin wasn't allowing the the mappings to be disabled via the standard `no_plugin_maps` mechanism so I've fixed that.  See `:help no_plugin_maps` for details.

`browsefilter` has also been available on GTK for some time and I added support for that as well.

Thanks,
Doug


costerwi:
Thanks for the contributions! You may update my email to costerwi@gmail.com if you like.

brammool:
I'll include it, thanks.

PR #10844, opened 2022-08-04
Fix GetCount() example

cornett:


brammool:
Yes, that's better, I'll include it.

PR #10845, opened 2022-08-04
`vim9cmd execute(cmd)` and `legacy call execute(cmd)` execute cmd in wrong context

kat0h:
## issue
Contents of execute function executed by `:legacy call execute("let g:hoge = 'hugo'")` or :vim9cmd in Vim9 script are processed by unintended context.

## cause
execute_common() does not refer to the CMOD_LEGACY and CMOD_VIM9CMD flag, so the return value of in_vim9script() is referenced and the command is executed as wrong context

## solution
CMOD_LEGACY and CMOD_VIM9CMD are now checked.
same as :execute (https://github.com/vim/vim/blob/master/src/eval.c#L6793)

## test code
These scripts are expected to run but return an error
```vim
vim9script
legacy call execute('let g:str = rand()')
legacy call execute(['let g:list1 = rand()', 'let g:list2 = rand()'])

echo g:str
echo g:list1
echo g:list2
```

```vim
let g:str = 0
let g:list1 = 0
let g:list2 = 0

vim9cmd execute('g:str = rand()')
vim9cmd execute(['g:list1 = rand()', 'g:list2 = rand()'])

echo g:str
echo g:list1
echo g:list2
```

### DIFF
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -3929,6 +3929,7 @@ execute_common(typval_T *argvars, typval_T *rettv, int arg_off)
     int		save_redir_off = redir_off;
     garray_T	save_ga;
     int		save_msg_col = msg_col;
+    int		save_sticky_cmdmod_flags = sticky_cmdmod_flags;
     int		echo_output = FALSE;
 
     rettv->vval.v_string = NULL;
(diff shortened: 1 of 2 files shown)


brammool:
Thanks for looking into this.  Your examples should be turned into a test case.
Probably similar to the test for "legacy exe", in Test_cmdmod_execute() in testdir/test_vim9_cmd.vim

kat0h:
Thanks for the confirmation and advice.
I have created a test case. Please check carefully as I am not used to creating tests for Vim.

brammool:
Thanks.  The test looks fine.  Perhaps it is not needed to also test with two assignments, but it doesn't hurt either.

Issue #10856, opened 2022-08-06
unexpected error when indenting python code while `'debug'` set to `throw`


lacygoill:
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        filetype indent on
        &debug = 'throw'
        &filetype = 'python'
        ['func()']->repeat(100)->setline(1)
        normal! gg=G
    EOF
    )

An error is given:

    E121: Undefined variable: dummy

**Expected behavior**

No error is given, because the indent python script should not rely on an error to work correctly.

If it does, it can be more difficult to debug an issue in a different indent script.  Case in point: [the Google python style guide](https://google.github.io/styleguide/pyguide.html) gives a settings file for Vim which includes a wrapper function around `GetPythonIndent()`.  If we need to tweak it and debug an issue inside, getting unrelated errors from `python#GetIndent()` is confusing.

**Version of Vim**

9.0 Included patches: 1-148

**Environment**

Operating system: Ubuntu 20.04.4 LTS
Terminal: xterm
Value of $TERM: xterm-256color
Shell: GNU bash, version 5.0.17

**Additional context**

The issue comes from [this script](https://github.com/vim/vim/blob/master/runtime/autoload/python.vim).
More specifically, from [these kinds of blocks](https://github.com/vim/vim/blob/5ac4b1a24e3ba52698a0d24b22cdd2e1c7895417/runtime/autoload/python.vim#L59-L65):

    " Trick: use the non-existing "dummy" variable to break out of the loop when
    " going too far back.
    let parlnum = searchpair('(\|{\|\[', '', ')\|}\|\]', 'nbW',
            \ "line('.') < " . (plnum - s:maxoff) . " ? dummy :"
            \ . " synIDattr(synID(line('.'), col('.'), 1), 'name')"
            \ . " =~ '\\(Comment\\|Todo\\|String\\)$'",
            \ searchpair_stopline, searchpair_timeout)

Why using a stopline set to 0?

    let searchpair_stopline = 0

It has no effect.  From [`:help search()`](https://vimhelp.org/builtin.txt.html#search%28%29):

    A zero value is equal to not giving the argument.

It *seems* `dummy` is meant to stop `searchpair()` from searching too far away.  That's exactly the purpose of the `{stopline}` argument.  So, why not setting it correctly?

---

Unrelated, but, these regexes:

    [[({]
    [])}]

seem more readable than these:

    (\|{\|\[
    )\|}\|\]

[...]

PR #10860, opened 2022-08-07
remove dummy variable from python indent expression

lacygoill:
This is an attempt at fixing #10856

Also, it changes the indentation inside parens.

Before:

    open_paren_not_at_EOL(100,
            (200,
                300),
            400)

After:

    open_paren_not_at_EOL(100,
                          (200,
                           300),
                          400)

The 2nd one seems more readable, and some people might prefer it.
If that's not desirable, I could remove the new code responsible for this.

It includes a test, but I'm not sure everything is working as intended.

### DIFF
--- a/runtime/autoload/python.vim
+++ b/runtime/autoload/python.vim
@@ -3,12 +3,27 @@
 let s:keepcpo= &cpo
 set cpo&vim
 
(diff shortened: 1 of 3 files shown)


brammool:
Thanks, I'll include it.  We can further extend the test over time.

PR #10869, opened 2022-08-09
Add setcmdline()

Shougo:
It is useful to create plugins for command line.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -506,6 +506,7 @@ setcellwidths({list})		none	set character cell width overrides
 setcharpos({expr}, {list})	Number	set the {expr} position to {list}
 setcharsearch({dict})		Dict	set character search from {dict}
 setcmdpos({pos})		Number	set cursor position in command-line
+setcmdline({str} [, {pos}])	Number	set command-line
 setcursorcharpos({list})	Number	move cursor to position in {list}
 setenv({name}, {val})		none	set environment variable
 setfperm({fname}, {mode})	Number	set {fname} file permissions to {mode}
@@ -3425,7 +3426,8 @@ getcmdcompltype()					*getcmdcompltype()*
 		Only works when the command line is being edited, thus
 		requires use of |c_CTRL-\_e| or |c_CTRL-R_=|.
 		See |:command-completion| for the return string.
-		Also see |getcmdtype()|, |setcmdpos()| and |getcmdline()|.
+		Also see |getcmdtype()|, |setcmdpos()|, |getcmdline()| and
+		|setcmdline()|.
 		Returns an empty string when completion is not defined.
 
 getcmdline()						*getcmdline()*
@@ -3434,7 +3436,8 @@ getcmdline()						*getcmdline()*
 		|c_CTRL-R_=|.
 		Example: >
 			:cmap <F7> <C-\>eescape(getcmdline(), ' \')<CR>
-<		Also see |getcmdtype()|, |getcmdpos()| and |setcmdpos()|.
+<		Also see |getcmdtype()|, |getcmdpos()|, |setcmdpos()| and
+		|setcmdline()|.
 		Returns an empty string when entering a password or using
 		|inputsecret()|.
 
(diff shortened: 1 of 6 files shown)


brammool:
I wonder what happens when calling setcmdline() while in custom completion.  Should have a test for that.
There might be other situations where changing the command line unexpectedly might cause trouble.

Shougo:
> I wonder what happens when calling setcmdline() while in custom completion. 

What means the `custom completion`?
It is command custom completion?   To add tests, I need to understand it.

brammool:
> > I wonder what happens when calling setcmdline() while in custom completion. 
> 
> What means the `custom completion`?
> It is command custom completion?   To add tests, I need to understand it.

For a user command custom completion can be defined with:
	-complete=custom,{func}

See ":help :command-completion-custom".
In the function used here, if you would call setcmdline(), I wonder if
something breaks (using freed memory, for example).

Perhaps a :cmap using <expr> might also cause trouble.

-- 
GALAHAD: No look, really, this isn't nescess ...
PIGLET:  We must examine you.
GALAHAD: There's nothing wrong with ... that.

Shougo:
I have added tests.

Shougo:
```
Failures: 
	From test_cmdline.vim:
	Found errors in Test_setcmdline():
	command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[468]..function RunTheTest[44]..Test_setcmdline line 33: Expected 'let foo=0' but got '"DoCmd January February Mars'
```

I don't know why it is failed in gvim only.
It works for me...

Shougo:
I have changed the map character.  It is passed.

vimpostor:
Do you have an example usecase for this function? I don't quite get where this is useful.

bfrg:
> Do you have an example usecase for this function? I don't quite get where this is useful.

1. Custom command-line popup window (the popup callback function calls `setcmdline()` to set the command-line after user made selection and hit ENTER)
2. If someone has a user command with a command-completion function that obtains its list of candidates asynchronously the job callback function needs a way to set the command-line once the list of candidates are obtained.

runtime/doc/builtin.txt
```diff
@@ -7896,6 +7901,16 @@ setcharsearch({dict})					*setcharsearch()*
...
 
+setcmdline({str} [, {pos}])					*setcmdline()*
+		Set the command line to {str} and set the cursor position to
+		{pos}.
+		If {pos} is omitted, the last position is used.
+		Returns FALSE when successful, TRUE when not editing the
+		command line.
+
+		Can also be used as a |method|: >
+			SetCmdline()->setcmdline()
```
bfrg: How about a better name for the first function? For example:
```
GetText()->setcmdline()
```

Shougo: OK.

habamax:
> 1. Custom command-line popup window (the popup callback function calls `setcmdline()` to set the command-line after user made selection and hit ENTER)

I for now use `feedkeys(":Command")` to have it, I guess proper function would be better.

brammool:
A few review comments (sorry it took a while):
- please use the same formatting as for other functions (function return type goes on a separate line)
- help says returning FALSE or TRUE, but it looks like it's returning one and zero
- use a better error than e_invalid_argument 
- a negative position is rejected without any error
- All declarations go at the start of the block (ANSI C)

Shougo:
> help says returning FALSE or TRUE, but it looks like it's returning one and zero

yes, but it is the same behavior with `setcmdpos()`.  The documentation is wrong??

Shougo:
> please use the same formatting as for other functions (function return type goes on a separate line)

Fixed.

>  a negative position is rejected without any error

Fixed.

> All declarations go at the start of the block (ANSI C)

Fixed.

brammool:
> > help says returning FALSE or TRUE, but it looks like it's returning
> > one and zero
> 
> yes, but it is the same behavior with `setcmdpos()`.  The
> documentation is wrong?? 

Ah, you took it from there, and there it's confusing as well.

FALSE and TRUE are used for argument values, where zero and v:false are
accepted as FALSE and 1 or v:true are accepted as TRUE.  But when
returning a value we know what we are returning, thus we can just use
zero and one.  Othewise users may expect v:false or v:true, which
matters when using Vim9 script.

-- 
It is illegal to take more than three sips of beer at a time while standing.
		[real standing law in Texas, United States of America]

Shougo:
Thank you for the merge.

Issue #10838, opened 2022-08-03
`[s` skips past `SpellCap` errors at start of line.


sedm0784:
### Steps to reproduce

1. Run vim: `vim --clean`
2. Enter the text:
```
           This is a tesst. so is this.
           this should work.
```
3. Switch on spell-checking: `:set spell`. The three spelling errors (`tesst`, lowercase `so` & `this`) are highlighted correctly with `SpellBad` and `SpellCap`.
4. With the cursor at the very end of the buffer, type `[s`

**Expected Behaviour**: The cursor should jump to the lowercase `this` at the start of line 2.
**Observed Behaviour**: The cursor jumps to the lowercase `so` in the middle of line 1.

Tested on Vim 8.2.2859 (Windows) and Vim 9.0.65 (macOS)

### Expected behaviour

`[s` should jump to all highlighted spelling errors, no matter their position in the line.

### Version of Vim

8.2.2859
9.0.65

matteocoder:
@sedm0784 Highlighting SpellCap errors seems to be buggy as well.
Try joining the first and second lines with `J`, then go to the last "this" and insert a carriage return; at least on my machine, the SpellCap highlighting is gone, even though `]s` can jump to it, but `[s` does not.

chrisbra:
yes, it seems like vim doesn't handle the case when the previous line ends with whitespace

sedm0784:
So yet again I have gone to the trouble of writing up a detailed bug report without first taking the basic step of consulting the manual. <sigh>

I'm going to leave this open, because it still feels like a bug even though it's documented, but I guess the fact that it _is_ suggests it's not going to get fixed any time soon.

PR #10870, opened 2022-08-09
Correctly highlight SpellCaps error with trailing ws

chrisbra:
as noticed in #10838 SpellCap highlighting is not applied, if the previous line ends with trailing white space.

So skip calculating cap_col, in `win_line()` when the rest of the line only contains whitespaces.


chrisbra:
failure seems unrelated

PR #10872, opened 2022-08-09
Fix typos in comments

mityu:
Fixed a file name mismatch in the file description comment at the head of vim9expr.c.
(Note that the file description comment is same to the vim9cmds.c's, so maybe it should also be changed?)


mityu:
I found another typo. I'll add it. Please wait a minute.

mityu:
Added.

brammool:
Thanks, I'll include this. I'll include it in some later patch.

PR #10877, opened 2022-08-09
Runtime(swayconfig syntax): Fix swipe and pinch gesture finger numbers

jamespeapen:


brammool:
I'll include it, thanks.

PR #10886, opened 2022-08-10
(Proposal) Fix is/isnot expr of vim9script

kat0h:
This issue is Reported by @thinca 
@kat0h made this patch

## issue
is/isnot expr of Vim9 script are looks broken. 

This is reproduction code.
```vim
vim9script
echo 'foo' is 'foo'    # => false
echo 'foo' is 'bar'    # => false
echo 'foo' isnot 'foo' # => true
echo 'foo' isnot 'bar' # => true

var a = 'foo'
echo a is a            # => false
echo a isnot a         # => true
```

## cause
Vim script uses `typval_compare_string()` function in `src/typval.c` to compare strings.
in Vim9 script context, [it checks if the pointers of the two values passed as arguments equal](https://github.com/vim/vim/blob/master/src/typval.c#L1608-L1625).
But, LOAD command of assembly pushes value to stack after copy it. so values are placed different address.

I think, It seems that string value is not an object. So, I think vim can't compare that the string instances are identical.
```vim
vim9script
var a = 'foo'
var b = a
a = 'bar'
echo b #=> 'foo'
```

## solution
I deleted code which checks that string instances are same.
This patch includes breaking change. Now, `null_string is ''` returns true, and `null_string isnot ''` returns false.

regards.

### DIFF
--- a/src/testdir/test_vim9_expr.vim
+++ b/src/testdir/test_vim9_expr.vim
@@ -886,11 +886,11 @@ def Test_expr4_compare_null()
       assert_false(null != null_string)
 
       assert_true(null_string is test_null_string())
-      assert_false(null_string is '')
-      assert_false('' is null_string)
+      assert_true(null_string is '')
+      assert_true('' is null_string)
       assert_false(null_string isnot test_null_string())
-      assert_true(null_string isnot '')
-      assert_true('' isnot null_string)
+      assert_false(null_string isnot '')
+      assert_false('' isnot null_string)
 
       var ns = null_string
       assert_true(ns == null_string)
@@ -1271,6 +1271,19 @@ def Test_expr4_is()
                             is 0z1234)
       var otherblob = myblob
       assert_true(myblob is otherblob)
+
+      var va = 'foo'
+      var vb = 'foo'
+      var vc = 'bar'
+      var vd = 123
+
+      assert_true(va is va)
+      assert_true(va is vb)
+      assert_false(va is vc)
+
+      assert_false(va isnot va)
+      assert_false(va isnot vb)
+      assert_true(va isnot vc)
   END
   v9.CheckDefAndScriptSuccess(lines)
 enddef
--- a/src/typval.c
+++ b/src/typval.c
@@ -1607,24 +1607,12 @@ typval_compare_string(
 	i = ic ? MB_STRICMP(s1, s2) : STRCMP(s1, s2);
     switch (type)
     {
-	case EXPR_IS:	    if (in_vim9script())
-			    {
-				// Really check it is the same string, not just
-				// the same value.
-				val = tv1->vval.v_string == tv2->vval.v_string;
-				break;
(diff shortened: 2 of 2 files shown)


brammool:
> This issue is Reported by @thinca 
> @kat0h made this patch
> 
> ## issue
> is/isnot expr of Vim9 script are looks broken. 
> 
> This is reproduction code.
> ```vim
> vim9script
> echo 'foo' is 'foo'    # => false
> echo 'foo' is 'bar'    # => false
> echo 'foo' isnot 'foo' # => true
> echo 'foo' isnot 'bar' # => true
> 
> var a = 'foo'
> echo a is a            # => false
> echo a isnot a         # => true
> ```
> 
> ## cause
> Vim script uses `typval_compare_string()` function in `src/typval.c` to compare strings.
> in Vim9 script context, [it checks if the pointers of the two values passed as arguments equal](https://github.com/vim/vim/blob/master/src/typval.c#L1608-L1625).
> But, LOAD command of assembly pushes value to stack after copy it. so values are placed different address.
> 
> I think, It seems that string value is not an object. So, I think vim can't compare that the string instances are identical.
> ```vim
> vim9script
> var a = 'foo'
> var b = a
> a = 'bar'
> echo b #=> 'foo'
> ```
> 
> ## solution
> I deleted code which checks that string instances are same.
> This patch includes breaking change. Now, `null_string is ''` returns
> true, and `null_string isnot ''` returns false.

This choice was made quite a while ago.  Nearly everywhere a string is
copied before it is passed on, thus "is" would nearly always return
false.  Using that as a reason to fall back to comparing the string value
doesn't make sense either, now two completely unrelated strings may have
"is" return true, just because they happen to have the same value.

One can argue for one or the other.  In a case like this I rather keep
it as it is.  Changing it now is likely to cause trouble in a few
places, and I don't see it helping much.

Another way to look at it: like you mention, strings are not objects,
thus "is" always returns false.

Yet another way would be to give an error when trying to use "is" on
strings.  But that likely causes even more trouble.

A very complicated way to solve this would be to use reference counting
for strings.  Thus intead of making a copy the strings would indeed be
identical.  The reference counting is needed to make sure the string is
deleted if, and only if, the last reference to it is dropped.  It might
be a bit more efficient to avoid making copies, but then keeping the
reference counts correct adds overhead (which should not be
underestimated).

-- 
Facepalm reply #9: "Speed up, you can drive 80 here" "Why, the cars behind us
are also driving 60"

kat0h:
Thank you very much for checking my PR!

brammool:
> @brammool
>  This behavior is different from legacy scripts, but this is not documented.
> I believe this should be explained in the documentation.
> 
> `:help vim9-differences`
> `:help convert_legacy_function_to_vim9`
> `:help expr-is`

I thought it was mentioned somewhere, but I can't find it now.
I'll add a note.

-- 
Why doesn't Tarzan have a beard?

PR #10889, opened 2022-08-10
Matchit support for Elixir

mhanberg:


brammool:
Since this uses line continuation, it should set and restore the 'cpo' option.  See other plugins for examples.

mhanberg:
I saw some files did `set cpo&vim` and some did `set cpo-=C`. I'm not sure what the difference is, but I can change it if this is not correct.

brammool:
Both ways work.  Often "cpo&vim" is a bit better because it also affects how patterns work.
I'll include it now, thanks.

PR #10893, opened 2022-08-11
Fix cmdheight problems

### LINKED ISSUE #10881: "-- INSERT --" from showmode never clears with cmdheight=0
If you set cmdheight to 0, the showmode text never clears:

    % vim --noplugin -u NONE +'cmdheight=0 showmode'

"-- INSERT --" shows, but after leaving insert mode it stays around until you press `:` or `^L`. Same with "-- REPLACE --"; it does seem to clear for "-- VISUAL --" or completions.

If you also disable the statusline:

    % vim --noplugin -u NONE +'cmdheight=0 showmode'

Then you get the same, and the `showmode` text displays "in" the statusline, overwriting your statusline text.

---

According to `:help showmode`, "the option has no effect when 'cmdheight' is zero", so that doesn't seem to work quite right. Personally, I think the behaviour when the statusline is disabled is actually okay (except for the not clearing part), and I can see how it might be okay with a statusline too if people right-align their statusline (although can replicate it easily with a custom statusline too).

Vim 9.0.0181 on Linux

Shougo:
Note: neovim does not exist the problems.

I have fixed below problems.

* ~~'showmode' does work even `cmdheight=0`.~~

* ~~Double echoes `:!{cmd}` when `cmdheight=0`.~~

* ~~Skip `<C-c>` messages when `cmdheight=0`.~~

Fix #10881

### DIFF
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -1010,12 +1010,16 @@ do_bang(
     }
     if (addr_count == 0)		// :!
     {
-	// echo the command
-	msg_start();
-	msg_putchar(':');
-	msg_putchar('!');
-	msg_outtrans(newcmd);
-	msg_clr_eos();
+        if (p_ch > 0)
+	{
+	    // echo the command
+	    msg_start();
+	    msg_putchar(':');
+	    msg_putchar('!');
+	    msg_outtrans(newcmd);
+	    msg_clr_eos();
+	}
+
 	windgoto(msg_row, msg_col);
 
 	do_shell(newcmd, 0);
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -1010,16 +1010,12 @@ do_bang(
     }
     if (addr_count == 0)		// :!
     {
-        if (p_ch > 0)
-	{
-	    // echo the command
-	    msg_start();
-	    msg_putchar(':');
-	    msg_putchar('!');
-	    msg_outtrans(newcmd);
-	    msg_clr_eos();
-	}
-
+	// echo the command
+	msg_start();
+	msg_putchar(':');
+	msg_putchar('!');
+	msg_outtrans(newcmd);
+	msg_clr_eos();
 	windgoto(msg_row, msg_col);
 
 	do_shell(newcmd, 0);
--- a/src/message.c
+++ b/src/message.c
@@ -3677,7 +3677,7 @@ msg_clr_eos_force(void)
 	}
     }
     else
-	redraw_all_later(CLEAR);
+	redraw_all_later(UPD_CLEAR);
 }
 
 /*
--- a/src/message.c
+++ b/src/message.c
@@ -3677,7 +3677,10 @@ msg_clr_eos_force(void)
 	}
     }
     else
-	redraw_all_later(UPD_CLEAR);
+    {
+	redraw_all_later(UPD_NOT_VALID);
+	clear_cmdline = TRUE;
+    }
 }
 
 /*
--- a/src/message.c
+++ b/src/message.c
@@ -3677,7 +3677,7 @@ msg_clr_eos_force(void)
 	}
     }
     else
-	redrawWinline(curwin, msg_row);
+	redraw_all_later(VALID);
 }
 
 /*
--- a/src/message.c
+++ b/src/message.c
@@ -3677,7 +3677,7 @@ msg_clr_eos_force(void)
 	}
     }
     else
-	redraw_later(CLEAR);
+	redrawWinline(curwin, msg_row);
 }
 
 /*
--- a/src/message.c
+++ b/src/message.c
@@ -3676,6 +3676,8 @@ msg_clr_eos_force(void)
 	    screen_fill(msg_row + 1, (int)Rows, 0, (int)Columns, ' ', ' ', 0);
 	}
     }
+    else
+	redraw_later(CLEAR);
 }
 
 /*
--- a/src/message.c
+++ b/src/message.c
@@ -3677,7 +3677,7 @@ msg_clr_eos_force(void)
 	}
     }
     else
-	redraw_all_later(VALID);
+	redraw_all_later(CLEAR);
 }
 
 /*
--- a/src/normal.c
+++ b/src/normal.c
@@ -6777,6 +6777,7 @@ nv_esc(cmdarg_T *cap)
 		&& cmdwin_type == 0
 #endif
 		&& !VIsual_active
+		&& p_ch > 0
 		&& no_reason)
 	{
 	    if (anyBufIsChanged())
--- a/src/normal.c
+++ b/src/normal.c
@@ -6777,7 +6777,6 @@ nv_esc(cmdarg_T *cap)
 		&& cmdwin_type == 0
 #endif
 		&& !VIsual_active
-		&& p_ch > 0
 		&& no_reason)
 	{
 	    if (anyBufIsChanged())
--- a/src/screen.c
+++ b/src/screen.c
@@ -4184,6 +4184,7 @@ skip_showmode()
     if (global_busy
 	    || msg_silent != 0
 	    || !redrawing()
+	    || p_ch <= 0
 	    || (char_avail() && !KeyTyped))
     {
 	redraw_mode = TRUE;		// show mode later
--- a/src/screen.c
+++ b/src/screen.c
@@ -4184,7 +4184,6 @@ skip_showmode()
     if (global_busy
 	    || msg_silent != 0
 	    || !redrawing()
-	    || p_ch <= 0
 	    || (char_avail() && !KeyTyped))
     {
 	redraw_mode = TRUE;		// show mode later
--- a/src/testdir/test_messages.vim
+++ b/src/testdir/test_messages.vim
@@ -480,7 +480,6 @@ func Test_cmdheight_showmode()
 
   set showmode&
   set cmdheight&
-  redraw!
 endfunc
 
 func Test_cmdheight_zero_shell()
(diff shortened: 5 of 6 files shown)


brammool:
> Note: neovim does not exist the problems.
> 
> I have fixed two problems.
> 
> * &#39;showmode&#39; does work even `cmdheight=0`.
> 
> * Double echoes `:!{cmd}` when `cmdheight=0`.
> 
> Fix #10881

Please add regression tests for these fixes.

-- 
"Intelligence has much less practical application than you'd think."
		  -- Scott Adams, Dilbert.

Shougo:
I want to add the tests, but it is for screen tests.
I don't know how to add screen tests.
neovim has helper functions for it.

lacygoill:
Not sure it helps here but there is:

   - [`:help terminal-dumptest`](https://vimhelp.org/terminal.txt.html#terminal-dumptest)
   - [`:help terminal-screendump`](https://vimhelp.org/terminal.txt.html#terminal-screendump)
   - [`:help terminal-diffscreendump`](https://vimhelp.org/terminal.txt.html#terminal-diffscreendump)

lacygoill:
See also the [`README.txt`](https://github.com/vim/vim/blob/master/src/testdir/README.txt) file in `src/testdir/`:

    TO ADD A SCREEN DUMP TEST:

    Mostly the same as writing a new style test.  Additionally, see help on
    "terminal-dumptest".  Put the reference dump in "dumps/Test_func_name.dump".

    TO ADD A NEW STYLE TEST:

    1) Create a test_<subject>.vim file.
    2) Add test_<subject>.res to NEW_TESTS_RES in Make_all.mak in alphabetical
       order.
    3) Also add an entry "test_<subject>" to NEW_TESTS in Make_all.mak.
    4) Use make test_<subject> to run a single test.

    At 2), instead of running the test separately, it can be included in
    "test_alot".  Do this for quick tests without side effects.  The test runs a
    bit faster, because Vim doesn't have to be started, one Vim instance runs many
    tests.

    At 4), to run a test in GUI, add "GUI_FLAG=-g" to the make command.


    What you can use (see test_assert.vim for an example):

    - Call assert_equal(), assert_true(), assert_false(), etc.

    - Use assert_fails() to check for expected errors.

    - Use try/catch to avoid an exception aborts the test.

    - Use test_alloc_fail() to have memory allocation fail.  This makes it possible
      to check memory allocation failures are handled gracefully.  You need to
      change the source code to add an ID to the allocation.  Add a new one to
      alloc_id_T, before aid_last.

    - Use test_override() to make Vim behave differently, e.g.  if char_avail()
      must return FALSE for a while.  E.g. to trigger the CursorMovedI autocommand
      event. See test_cursor_func.vim for an example.

    - If the bug that is being tested isn't fixed yet, you can throw an exception
      with "Skipped" so that it's clear this still needs work.  E.g.: throw
      "Skipped: Bug with <c-e> and popupmenu not fixed yet"

    - The following environment variables are recognized and can be set to
      influence the behavior of the test suite (see runtest.vim for details)

      - $TEST_MAY_FAIL=Test_channel_one    - ignore those failing tests
      - $TEST_FILTER=Test_channel    - only run test that match this pattern
      - $TEST_SKIP_PAT=Test_channel  - skip tests that match this pattern
      - $TEST_NO_RETRY=yes           - do not try to re-run failing tests
      You can also set them in Vim:
        :let $TEST_MAY_FAIL = 'Test_channel_one'
        :let $TEST_FILTER = '_set_mode'
        :let $TEST_SKIP_PAT = 'Test_loop_forever'
        :let $TEST_NO_RETRY = 'yes'
      Use an empty string to revert, e.g.:
        :let $TEST_FILTER = ''

    - See the start of runtest.vim for more help.

Shougo:
I have added tests.

brammool:
In the tests "redraw!" is used before restoring 'cmdheight', shouldn't that be the other way around?

brammool:
Clearing the screen at the end of msg_clr_eos_force() looks like a brute force method.  Isn't there a better way?

Shougo:
> In the tests "redraw!" is used before restoring 'cmdheight', shouldn't that be the other way around?

Fixed.

Shougo:
> Clearing the screen at the end of msg_clr_eos_force() looks like a brute force method. Isn't there a better way?

I want to just redraw `msg_row` lines, but I cannot find the method in `drawscreen.c`...
Do you know the function?

brammool:
> > Clearing the screen at the end of msg_clr_eos_force() looks like a
> > brute force method. Isn't there a better way?
> 
> I want to just redraw `msg_row` lines, but I cannot find the method in
> `drawscreen.c`...  Do you know the function?

So this is when there is a message at the bottom of the screen, and you
want to remove it by redrawing the window at the bottom.

The optimal would be to set b_mod_top and b_mod_bot.  You need to figure
out the buffer lines then (perhaps "w_botline - 1" works if there is only
one screen line affected).

If only the last window is affected, then redraw_win_later(wp,
NOT_VALID) should work.  If the screen scrolled, then
redraw_all_later(VALID) probaby can be used, since "msg_scrolled" is
handled in update_screen().

-- 
Apologies for taking up the bandwidth with the apology.  Anything else I
can apologise for ...... er no can't think of anything, sorry about that.
		Andy Hunt (Member of British Olympic Apology Squad)

Shougo:
> The optimal would be to set b_mod_top and b_mod_bot.  You need to figure
out the buffer lines then (perhaps "w_botline - 1" works if there is only
one screen line affected).

Well, but both `b_mod_top` and `b_mod_top` are `line number`.
It is not `screen position`.  How to convert it?

Shougo:
> If only the last window is affected, then redraw_win_later(wp,
NOT_VALID) should work.  If the screen scrolled, then
redraw_all_later(VALID) probaby can be used, since "msg_scrolled" is
handled in update_screen().

And how to compute what windows are affected?
I have read the code and I have not found the optimization.
It seems simple redraw routine.

Shougo:
Hm...  I have created the routine, but it is too difficult and does not work well.

```diff
diff --git a/src/message.c b/src/message.c
index 48012bb32..dbe41be04 100644
--- a/src/message.c
+++ b/src/message.c
@@ -3477,7 +3477,46 @@ msg_clr_eos_force(void)
    }
     }
     else
-   redraw_all_later(VALID);
+    {
+   int scrolled;
+   int type = VALID;
+   win_T   *wp;
+
+   scrolled = Rows - msg_row;
+   if (scrolled > Rows - 5)        // clearing is faster
+       type = CLEAR;
+   else if (type != CLEAR)
+   {
+       check_for_delay(FALSE);
+       if (screen_ins_lines(0, 0, scrolled, (int)Rows, 0, NULL)
+                                      == FAIL)
+       type = CLEAR;
+       FOR_ALL_WINDOWS(wp)
+       {
+       if (wp->w_winrow < scrolled)
+       {
+           if (W_WINROW(wp) + wp->w_height > scrolled
+               && wp->w_redr_type < REDRAW_TOP
+               && wp->w_lines_valid > 0
+               && wp->w_topline == wp->w_lines[0].wl_lnum)
+           {
+           wp->w_upd_rows = scrolled - W_WINROW(wp);
+           wp->w_redr_type = REDRAW_TOP;
+           }
+           else
+           {
+           wp->w_redr_type = NOT_VALID;
+           if (W_WINROW(wp) + wp->w_height + wp->w_status_height
+                                  <= msg_scrolled)
+               wp->w_redr_status = TRUE;
+           }
+       }
+       }
+       redraw_tabline = TRUE;
+   }
+
+   redraw_all_later(type);
+    }
 }
```

Shougo:
`redraw_all_later(CLEAR);`  is not smart.  But it works well without problems.
I like works version of code.

Shougo:
```
message.c: In function ‘msg_clr_eos_force’:
message.c:3481:26: error: ‘CLEAR’ undeclared (first use in this function); did you mean ‘CAR’?
 3481 |         redraw_all_later(CLEAR);
      |                          ^~~~~
      |                          CAR
```

Why?

`CLEAR` is defined in `vim.h` and it is already included.

brammool:
```
message.c: In function ‘msg_clr_eos_force’:
message.c:3481:26: error: ‘CLEAR’ undeclared (first use in this function); did you mean ‘CAR’?
 3481 |         redraw_all_later(CLEAR);
      |                          ^~~~~
      |                          CAR
```

Why?

`CLEAR` is defined in `vim.h` and it is already included.

It was renamed UPD_CLEAR in 9.0.0206

-- 
A real patriot is the fellow who gets a parking ticket and rejoices
that the system works.


brammool:
The messages test fails in some builds, and there is an ASAN error.  Probably the tests uncovered a problem, I'll have a look.

brammool:
I have included the tests and fixed a few problems.
Are the code changes in this PR still needed?  If yes, then please add a test that fails without them.

Shougo:
Thanks.  I will rebase it later.

Shougo:
I think the patch is not needed anymore.  But I will re-create another PR to revert some change.

brammool:
> I think the patch is not needed anymore.  But I will re-create another
> PR to revert some change.

While trying out cmdheight=0 I noticed that once in a while the last
statusline goes missing, the last window is too tall.  I also had to fix
a crash where an index goes over Rows, the two are probably related.
I haven't been able to reproduce it, it might have something to do with
a shell command.

Anyway, please check my question on the vim-dev maillist, discussing
whether we should keep supporting 'cmdheight' zero or revert it.


-- 
ARTHUR: It is I, Arthur, son of Uther Pendragon, from the castle of Camelot.
        King of all Britons, defeator of the Saxons, sovereign of all England!
   [Pause]
SOLDIER: Get away!

Shougo:
> While trying out cmdheight=0 I noticed that once in a while the last
statusline goes missing, the last window is too tall.  I also had to fix
a crash where an index goes over Rows, the two are probably related.
I haven't been able to reproduce it, it might have something to do with
a shell command.

Well, I don't reproduce the problem.
I will check it if it is reproduced.

Shougo:
> Anyway, please check my question on the vim-dev maillist, discussing
whether we should keep supporting 'cmdheight' zero or revert it.

OK.   I will take look later.

PR #10918, opened 2022-08-15
Fix 'shellslash' behaves differently when sourcing a script twice

zeertzjq:
Problem:    Changing 'shellslash' doesn't affect the result of expand()
            when sourcing a script multiple times.
Solution:   Always store si->sn_name on the execution stack.


zeertzjq:
There seems to be something wrong with this PR. I'll open another one.

Issue #10865, opened 2022-08-08
python indent plugin wrongly assumes that a comment does not contain nested syntax items


lacygoill:
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        filetype plugin indent on
        syntax on
        &filetype = 'python'
        &formatoptions = 'o'
        syntax match pythonCommentTitle /#\s*\u\w*\%(\s\+\u\w*\)*:\_s\@=/hs=s+1 contained containedin=pythonComment
        ['# Title: ']->setline(1)
        normal! Go
    EOF
    )

The buffer contains these lines:

    # Title:
        #

**Expected behavior**

The buffer contains these lines:

    # Title:
    #

**Version of Vim**

9.0 Included patches: 1-168

**Environment**

Operating system: Ubuntu 20.04.4 LTS
Terminal: xterm
Value of $TERM: xterm-256color
Shell: GNU bash, version 5.0.17

**Additional context**

The issue comes from this line:
https://github.com/vim/vim/blob/25463610dfc7a4984f70b030463fb98b09772ad9/runtime/autoload/python.vim#L164

It implicitly assumes that a comment does not contain any nested syntax.
This assumption is broken if the user wants to highlight a title (like the legacy Vim syntax plugin does):

    " Title: text
      ^----^
      syntax stack: vimCommentTitle vimLineComment
                    ^-------------^

    # Title: text
      ^----^
      syntax stack: pythonCommentTitle pythonComment
                    ^----------------^

I can see at least 2 ways to fix this issue.  Check that the previous line is not a comment:
```diff
diff --git a/runtime/autoload/python.vim b/runtime/autoload/python.vim
index 7e7bca6fb..a59027504 100644
--- a/runtime/autoload/python.vim
+++ b/runtime/autoload/python.vim
@@ -161,7 +161,7 @@ function python#GetIndent(lnum, ...)
   endif
 
   " If the previous line ended with a colon, indent this line
-  if pline =~ ':\s*$'
+  if pline =~ ':\s*$' && pline !~ '^\s*#'
     return plindent + shiftwidth()
   endif
 
```
Relax the regex used to check the presence of `Comment` inside the name of the syntax item:
```diff
diff --git a/runtime/autoload/python.vim b/runtime/autoload/python.vim
index 7e7bca6fb..b45ebf82b 100644
--- a/runtime/autoload/python.vim
+++ b/runtime/autoload/python.vim
[...]

PR #10921, opened 2022-08-15
fix python indentation when comment contains nested syntax items

lacygoill:
This is an attempt at fixing the issue discussed [here](https://github.com/vim/vim/issues/10865#issuecomment-1214464023).

MRE:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        filetype plugin indent on
        syntax on
        &filetype = 'python'
        &formatoptions = 'o'
        syntax match pythonFoldMarkers /{{{\d*/ contained containedin=pythonComment
        var lines =<< trim END
        # xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx {{{1

        if True:
        pass
        END
        lines->setline(1)
        normal! gg=G
    EOF
    )

Expected:

    # xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx {{{1

    if True:
        pass

Actual:

    # xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx {{{1

                                                  if True:
                                                  pass

---

The PR includes a test which passes only with the patch.
If someone could confirm or infirm that the patch works and doesn't break anything, that would help.

### DIFF
--- a/runtime/autoload/python.vim
+++ b/runtime/autoload/python.vim
@@ -15,8 +15,9 @@ let s:maxoff = 50       " maximum number of lines to look backwards for ()
 
 function s:SearchBracket(fromlnum, flags)
   return searchpairpos('[[({]', '', '[])}]', a:flags,
-          \ {-> synID('.', col('.'), v:true)->synIDattr('name')
-          \ =~ '\%(Comment\|Todo\|String\)$'},
+          \ {-> synstack('.', col('.'))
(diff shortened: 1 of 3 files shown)


brammool:
It works for me, thanks.  I'll include it.  There are probably further improvements to be made, let me know.
Or if someone wants to take over maintaining this Python indent plugin, you are welcome.
I don't write much python.

PR #10926, opened 2022-08-16
Vim9 script functions defined in Vim script cannot take references to script local functions whose names start with a lower case letter

kat0h:
Issue:
Only in Vim9 script functions defined in Vim script files, defining functions with s:foo style names (beginning with a lower case letter) are allowed, but it is not possible to get references to them by using the function() or by assigning variables.
The current behavior is inconvenient in terms of compatibility with Vim script, and I can't find the specifications in document.
```vim
# in def block
var Ref = s:foo
var Ref2 = function('s:foo')
```

Proposal:
Allow Vim9 script functions defined in Vim script to reference functions in s:foo format.

Advantage:
- It will be convenient to replace some of the plugin processes written in Vim script with Vim9 script without changing the function names.
- Improved compatibility with conventional Vim script.
- The current specification is unnatural because it can be processed normally in other cases

Reproducible code:
This code is added to the test.
```
function! s:func()
  echo "s:func() was called"
endfunction

def! s:C1()
  s:func()
enddef

def! s:C2()
  var Ref2 = function('s:func')
  Ref2()
enddef

def! s:C3()
  var Ref3 = function(s:func)
  Ref3()
enddef

def! s:C4()
  var Ref1 = s:func
  Ref1()
enddef


call s:C1()
" Error ↓
call s:C2()
call s:C3()
call s:C4()
```

### DIFF
--- a/src/testdir/test_vim9_func.vim
+++ b/src/testdir/test_vim9_func.vim
@@ -1957,6 +1957,45 @@ def Test_call_funcref()
     g:listarg->assert_equal([1, 2, 3])
   END
   v9.CheckScriptSuccess(lines)
+
+  lines =<< trim END
+    function s:func(num)
+      return a:num * 2
+    endfunction
+
+    def s:CallFuncref()
+      var Funcref = function('s:func')
+      Funcref(3)->assert_equal(6)
+    enddef
+    call s:CallFuncref()
+  END
+  v9.CheckScriptSuccess(lines)
+
+  lines =<< trim END
+    function s:func(num)
+      return a:num * 2
+    endfunction
+
+    def s:CallFuncref()
+      var Funcref = function(s:func)
+      Funcref(3)->assert_equal(6)
+    enddef
+    call s:CallFuncref()
+  END
+  v9.CheckScriptSuccess(lines)
+
+  lines =<< trim END
+    function s:func(num)
+      return a:num * 2
+    endfunction
+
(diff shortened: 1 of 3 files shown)


brammool:
> Issue:
> Only in Vim9 script functions defined in Vim script files, defining functions with s:foo style names (beginning with a lower case letter) are allowed, but it is not possible to get references to them by using the function() or by assigning variables.
> The current behavior is inconvenient in terms of compatibility with Vim script, and I can&#39;t find the specifications in document.
> ```vim
> # in def block
> var Ref = s:foo
> var Ref2 = function('s:foo')
> ```
> 
> Proposal:
> Allow Vim9 script functions defined in Vim script to reference functions in s:foo format.
> 
> Advantage:
> - It will be convenient to replace some of the plugin processes written in Vim script with Vim9 script without changing the function names.
> - Improved compatibility with conventional Vim script.
> - The current specification is unnatural because it can be processed normally in other cases

Although I would prefer always using an uppercase letter for user
functions, in this case it was already possible to define the function,
so we might as well allow for referencing it.

-- 
`When any government, or any church for that matter, undertakes to say to
 its subjects, "This you may not read, this you must not see, this you are
 forbidden to know," the end result is tyranny and oppression no matter how
 holy the motives' -- Robert A Heinlein, "If this goes on --"

kat0h:
Thank you very much for checking and receiving my PR!

PR #10931, opened 2022-08-17
Fix netrw plugin issue using powershell

MiguelBarro:
`Vim\vim82\autoload\netrw.vim plugin` has issues with powershell because the function `s:NetrwExe` falls back to `cmd` on windows. This is used for example to call `curl`.

`netrw.vim` uses vim's `shellescape()` to decorate urls and is witty enough to use single quotes to escape if powershell is detected (`'http://whatever'` for example). Uses it through `s:ShellEscape()`.

`curl` doesn't admit single quoted arguments from `cmd` and breaks. But it works on powershell because it will remove them for the call.

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -12066,7 +12066,7 @@ endfun
 " s:NetrwExe: executes a string using "!" {{{2
 fun! s:NetrwExe(cmd)
 "  call Dfunc("s:NetrwExe(a:cmd<".a:cmd.">)")
-  if has("win32") && &shell !~? 'cmd' && !g:netrw_cygwin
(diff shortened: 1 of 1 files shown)


MiguelBarro:
Thanks for the prompt answer. Sent ... I guess the pull request doesn't make sense anymore.

MiguelBarro:
Unfortunately, it seems to be something wrong with the [mail](https://www.mail-archive.com/vim_use@googlegroups.com/msg57151.html) (DNS complains the domain doesn't exist anymore).
Somebody can hint the good one?

chrisbra:
did you remove the capital letters? the domain should be IIRC `@drchip.org`

MiguelBarro:
My fault. I misspelled the domain. Now it seems to be delivered. Thanks @chrisbra 🙇‍♂️

PR #10932, opened 2022-08-17
Fix typo in diffmode test

zeertzjq:


zeertzjq:
There is something wrong

PR #10939, opened 2022-08-19
Update colorschemes

habamax:
- All colorschemes: better t_Co detection
- Some colorschemes: fix gui colors that should correspond to xterm256
  palette (e.g. #ffdf00 vs #ffd700)
- Fix blue g:terminal_ansi_colors
- Fix slate wrong mapping of cterm colors
- New colorscheme: lunaperche
- Fix check_colors.vim


brammool:
I'll include it, thanks.  I'll soon send out a runtime files update.

PR #10941, opened 2022-08-19
consolidate python indent config

lacygoill:
See https://github.com/vim/vim/pull/10929 for more context.

### DIFF
--- a/runtime/autoload/python.vim
+++ b/runtime/autoload/python.vim
@@ -6,6 +6,9 @@ set cpo&vim
 " need to inspect some old g:pyindent_* variables to be backward compatible
 let g:python_indent = extend(get(g:, 'python_indent', {}), #{
   \ closed_paren_align_last_line: v:true,
+  \ open_paren: get(g:, 'pyindent_open_paren', 'shiftwidth() * 2'),
+  \ nested_paren: get(g:, 'pyindent_nested_paren', 'shiftwidth()'),
+  \ continue: get(g:, 'pyindent_continue', 'shiftwidth() * 2'),
   "\ searchpair() can be slow, limit the time to 150 msec or what is put in
   "\ g:python_indent.searchpair_timeout
   \ searchpair_timeout: get(g:, 'pyindent_searchpair_timeout', 150),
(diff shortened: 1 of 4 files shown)


brammool:
Maybe my comment on the previous PR didn't get through or was unclear.
Currently, if a user changes "g:pyindent_continue" after the indent is used, it will take effect.
After this change that no longer works, thus it's not backwards compatible, even though the value is used once.
It will be a little bit more work, but it is possible to maintain backwards compatibility with some extra work.
Like there currently is an exists() call.
Same for the other existing g:pyindent_ values, looks like there are three.

lacygoill:
> Maybe my comment on the previous PR didn't get through or was unclear.

It seems the comment didn't get through to GitHub.

   > Currently, if a user changes "g:pyindent_continue" after the indent is used, it will take effect.
   > After this change that no longer works, thus it's not backwards compatible, even though the value is used once.
   > It will be a little bit more work, but it is possible to maintain backwards compatibility with some extra work.
   > Like there currently is an exists() call.
   > Same for the other existing g:pyindent_ values, looks like there are three.

OK, I'll try to improve the PR.

lacygoill:
> Currently, if a user changes "g:pyindent_continue" after the indent is used, it will take effect.
   > After this change that no longer works, thus it's not backwards compatible, even though the value is used once.
   > It will be a little bit more work, but it is possible to maintain backwards compatibility with some extra work.
   > Like there currently is an exists() call.
   > Same for the other existing g:pyindent_ values, looks like there are three.

Actually, I think I already noticed the issue and fixed it in [this commit](https://github.com/vim/vim/pull/10929/commits/525620190ca83d876b49d2f98f244b074ee4d6c8).

`g:python_indent.continue` is not computed when the script is sourced.  It's computed when the indent function is called via an `eval()`:

    return indent(a:lnum - 1) + g:python_indent.continue->eval()
                                                          ^----^

Same thing for `g:python_indent.nested_paren` and `g:python_indent.open_paren`:

    return indent(plnum) + g:python_indent.nested_paren->eval()
                                                         ^----^

    return indent(plnum) + g:python_indent.open_paren->eval()
                                                       ^----^

So, it should be backward compatible.  I'll make more tests later to be sure.

BTW, `exists()` is no longer used; it's replaced by `get()`, which defaults to 0 without the third optional argument.  This means that if the keys are not set, they don't add or remove anything to the computed indentation level, just like before.

brammool:
But the value of ""g:pyindent_continue" is only taken once, if the user changes it later it has no effect.
One can change "g:python_indent.continue" but that is not backwards compatible.

brammool:
Looks OK.  The first block in the test should not have INDENT_EXE, so that the default values are set in g:python_indent.

PR #10945, opened 2022-08-20
Allow `prop_remove` to operate on a list of property types

puremourning:
Removing text properties can be slow as it requires scanning the buffer lines internally. The best way is to use a line range when calling `prop_remove`. However, due to the way text properties move around when text is modified, placing a text prop with `prop_add` on line 10 does not mean that it's on line 10 when you might want to remove it. When working with virtual text and semantic tokens, there can be a lot of text properties that need to be refreshed periodically. The simplest way is to remove existing properties matching some ID (or type) and re-add the new ones.

With "classic" text properties, it was possible to assign the same Id to each property and thus remove the props in one scan of the affected range (e.g. whole buffer). With virtual text properties, it's not possible as the ID is generated and returned, so we're left with deleting "all by type" over the range, or recording the IDs of each property type and querying if it's still present (as deleting or undoing can remove it).

We have `prop_find` that can take a list of types to find, so to avoid having to do multiple scans of the lines and to avoid having to remember and query every ID, provide `prop_remove` with the ability to remove all properties with one of a list of property types.

```
Problem: can't remove text properties for multiple types at once
Solution: Allow types to be a list
```

concrete example:

* create text property types for `Param`, `Type` and `Enum` inlay hints, including a type for "padding"
* (periodically) clear all `Param`, `Type`,  `Enum` and padding properties and re-create them based on the result of some query to the semantic engine.

currently requires at least 4 calls to `prop_remove`; can now be done with `prop_remove( { 'types': [ 'Param', 'Type', 'Enum', 'padding' ], 'all': 1 } )`

### DIFF
--- a/runtime/doc/textprop.txt
+++ b/runtime/doc/textprop.txt
@@ -340,7 +340,7 @@ prop_list({lnum} [, {props}])				*prop_list()*
 		Can also be used as a |method|: >
 			GetLnum()->prop_list()
 <
-						*prop_remove()* *E968* *E860*
+					*prop_remove()* *E968* *E860* *E1295*
 prop_remove({props} [, {lnum} [, {lnum-end}]])
 		Remove a matching text property from line {lnum}.  When
 		{lnum-end} is given, remove matching text properties from line
@@ -352,11 +352,15 @@ prop_remove({props} [, {lnum} [, {lnum-end}]])
 		{props} is a dictionary with these fields:
 		   id		remove text properties with this ID
 		   type		remove text properties with this type name
-		   both		"id" and "type" must both match
+		   types	a list of property types to remove
+		   both		"id" and "type"/"types" must both match
 		   bufnr	use this buffer instead of the current one
 		   all		when TRUE remove all matching text properties,
 				not just the first one
-		A property matches when either "id" or "type" matches.
+		Only one of "type" and "types" may be supplied. 
+
+		A property matches when either "id" or one of the supplied
+		types matches.
 		If buffer "bufnr" does not exist you get an error message.
 		If buffer "bufnr" is not loaded then nothing happens.
 
--- a/src/errors.h
+++ b/src/errors.h
@@ -2209,7 +2209,7 @@ EXTERN char e_failed_to_convert_returned_python_object_to_vim_value[]
 #endif
 #ifdef FEAT_PROP_POPUP
 EXTERN char e_need_id_and_type_with_both[]
-	INIT(= N_("E860: Need 'id' and 'type' with 'both'"));
+	INIT(= N_("E860: Need 'id' and 'type' or 'types' with 'both'"));
 # ifdef FEAT_TERMINAL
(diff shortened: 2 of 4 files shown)


puremourning:
Happy to change it so that `type` can be a number or a list if you prefer that.

brammool:
I think providing both "type" and "types" should be considered an error, otherwise a mistake might go unnoticed.

It might also be useful to pass a ist of id's.  We could add that later.

puremourning:
> I think providing both "type" and "types" should be considered an error, otherwise a mistake might go unnoticed.

Sure thing, will add this.

> It might also be useful to pass a ist of id's. We could add that later.

I thought the exact same.

puremourning:
> > I think providing both "type" and "types" should be considered an error, otherwise a mistake might go unnoticed.
> 
> Sure thing, will add this.


Added

lacygoill:
Maybe it should be documented:
```diff
diff --git a/runtime/doc/textprop.txt b/runtime/doc/textprop.txt
index 9a148455e..470af2743 100644
--- a/runtime/doc/textprop.txt
+++ b/runtime/doc/textprop.txt
@@ -352,6 +352,7 @@ prop_remove({props} [, {lnum} [, {lnum-end}]])
 		{props} is a dictionary with these fields:
 		   id		remove text properties with this ID
 		   type		remove text properties with this type name
+		   types	remove text properties with these type names
 		   both		"id" and "type" must both match
 		   bufnr	use this buffer instead of the current one
 		   all		when TRUE remove all matching text properties,
```

lacygoill:
Actually, it might not be obvious that the value should be a list:
```diff
diff --git a/runtime/doc/textprop.txt b/runtime/doc/textprop.txt
index 9a148455e..0c4a12bfc 100644
--- a/runtime/doc/textprop.txt
+++ b/runtime/doc/textprop.txt
@@ -352,6 +352,8 @@ prop_remove({props} [, {lnum} [, {lnum-end}]])
 		{props} is a dictionary with these fields:
 		   id		remove text properties with this ID
 		   type		remove text properties with this type name
+		   types	remove text properties with these type names;
+				the value should be a list
 		   both		"id" and "type" must both match
 		   bufnr	use this buffer instead of the current one
 		   all		when TRUE remove all matching text properties,
```

puremourning:
Wait I swear I updated the doc in my patch... let me check what I messed up

puremourning:
> Wait I swear I updated the doc in my patch... let me check what I messed up

I did. I guess Bram will just update the runtime files at a later date. https://github.com/vim/vim/pull/10945/files#diff-9cdfd291efd41da3f72937703364d769bdb73b9b6145446717674dd1ce6f122a

lacygoill:
Ah, my bad.  I didn't check the PR, only the pushed commit.  Yes, the help will probably be updated later.

brammool:
> Wait I swear I updated the doc in my patch... let me check what I messed up 

Somehow the merge failed and this part went missing.

It will be in the next runtime files update.

-- 
There can't be a crisis today, my schedule is already full.

PR #10957, opened 2022-08-21
Add support for the getscriptnames() function

yegappan:


runtime/doc/builtin.txt
```diff
@@ -4088,6 +4089,18 @@ getregtype([{regname}])					*getregtype()*
 		Can also be used as a |method|: >
 			GetRegname()->getregtype()
 
+getscriptnames()					*getscriptnames()*
+		Returns a |List| with information about all the sourced Vim
+		scripts. (|:scriptnames|)
```
arp242: The `:scriptnames` command lists scripts in the order they were sourced, which is sometimes useful. Is that also the case here? Would be good to have the same behaviour, and also document it here.

yegappan: Yes. The getscriptnames() function also returns the scripts in the order they were sourced.

PR #10962, opened 2022-08-23
Add script version to getscriptinfo() and support for filtering scripts by name

yegappan:


brammool:
It is unlikely someone wants to get all variables and functions of all
scripts.  This adds a lot of overhead to the return value of
getscriptinfo().

The "version" item seems useful.  The "functions" and "variables" are
better obtained for a function that does this for one script.  I do
wonder what the use of it is.  What would justify adding this
functionality?  Making the script-local items accessible can be used for
hacks, the writer of the script may not want to make it so easy to
access them.

-- 
I just planted an Algebra tree. It has square roots.

vim-ml:
Hi Bram,

On Tue, Aug 23, 2022 at 11:55 AM Bram Moolenaar ***@***.***>
wrote:

>
> It is unlikely someone wants to get all variables and functions of all
> scripts. This adds a lot of overhead to the return value of
> getscriptinfo().
>
> The "version" item seems useful.


I will create a separate PR to add the "version" item.


> The "functions" and "variables" are
> better obtained for a function that does this for one script. I do
> wonder what the use of it is. What would justify adding this
> functionality? Making the script-local items accessible can be used for
> hacks, the writer of the script may not want to make it so easy to
> access them.
>
>
All the script-local functions can be already listed by using the
":function" command
without any arguments.  This change just makes it easier to get it as a
List.

Getting the list of script-local variables and their values is useful for
debugging
plugin issues (without modifying the plugin). It is useful for plugin
developers.

I will create a PR to filter the results by a particular script name.

Regards,
Yegappan

vim-ml:
> > It is unlikely someone wants to get all variables and functions of all
> > scripts. This adds a lot of overhead to the return value of
> > getscriptinfo().
> >
> > The "version" item seems useful.
> 
> I will create a separate PR to add the "version" item.

I also added the "sourced" item now.  Only set if the script was a link
that is resolved later.  It's useful that way.

> > The "functions" and "variables" are
> > better obtained for a function that does this for one script. I do
> > wonder what the use of it is. What would justify adding this
> > functionality? Making the script-local items accessible can be used for
> > hacks, the writer of the script may not want to make it so easy to
> > access them.
>
> All the script-local functions can be already listed by using the
> ":function" command
> without any arguments.  This change just makes it easier to get it as a
> List.

OK, knowing the script ID it was already possible, but a bit complicated
to get these functions.  And it's OK to call them, even from outside the
script, as this already happens for various reasons (mappings, option
values, etc.)

> Getting the list of script-local variables and their values is useful for
> debugging plugin issues (without modifying the plugin). It is useful
> for plugin developers.

One thing I want to avoid is to make it possible for script-local
variables to be changed from outside the script.  This makes it less
predictable.  Currently it is a "cannot happen", which reduces the
effort to find out why something happened.  For debugging it can be
useful to inspect the variables, but it would be good if we can make it
impossible to change them.  E.g. by making a copy.

-- 
BEDEVERE: Look!  It's the old man from scene 24 - what's he Doing here?
ARTHUR:   He is the keeper of the Bridge.  He asks each traveler five
          questions ...
GALAHAD:  Three questions.

PR #10977, opened 2022-08-25
fix(regression): Some highlights are cleared in GVim

habamax:
If a `colorscheme desert` (or any other) is called from .vimrc then some
of the highlights are cleared in GVim due to imperfect &t_Co condition check.

Fixes https://github.com/vim/colorschemes/issues/210


brammool:
I'll include it, thanks.

PR #10978, opened 2022-08-25
CI: use ubuntu-20.04 on Coverity as well

pheiduck:
I missed to update it on Coverity CI.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   scan:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     env:
       CC: gcc


brammool:
I see a lot of "deprecated" warnings for GTK include files.  Any way to avoid those?

chrisbra:
I guess you need either `-Wno-deprecated-declarations` or build GTK3 build

pheiduck:
> I guess you need either `-Wno-deprecated-declarations` or build GTK3 build

Its already set: https://github.com/vim/vim/blob/0166e398d11a09662d783fe5db62b414045880f8/.github/workflows/ci.yml#L24

chrisbra:
Hm, I have always just used `-Wno-deprecated-delcarations` like mentioned here:

https://github.com/vim/vim/blob/master/src/Makefile#L618

brammool:
> > I guess you need either `-Wno-deprecated-declarations` or build GTK3 build
> 
> Its already set: https://github.com/vim/vim/blob/0166e398d11a09662d783fe5db62b414045880f8/.github/workflows/ci.yml#L24

That means the warning is not considered an error, but the warning is
still given.  To suppress it we would need:
	-Wno-deprecated-declarations

But it would be better if the warning wasn't given.  I don't see a way
to make that happen, it's in the GTK include files.

I think the deprecated warnings are only annoying and don't help us make
Vim work better, so perhaps we should change that CFLAG value?

-- 
You can be stopped by the police for biking over 65 miles per hour.
You are not allowed to walk across a street on your hands.
		[real standing laws in Connecticut, United States of America]

chrisbra:
Yes,  perhaps we can just set this for builds, that enable the old legacy GTK2 gui

brammool:
> Yes,  perhaps we can just set this for builds, that enable the old
> legacy GTK2 gui

That's a bit more tricky than changing the existing CFLAGS.
Someone want so make a PR?

-- 
The Law, in its majestic equality, forbids the rich, as well as the
poor, to sleep under the bridges, to beg in the streets, and to steal
bread.                       -- Anatole France

pheiduck:
I found this: https://mail.gnome.org/archives/commits-list/2021-August/msg00550.html

pheiduck:
Oh tought we talk about ci.yml :D anyway I made a PR https://github.com/vim/vim/pull/11002 for Coverity to use CFLAGS: -Wno-error=deprecated-declarations

brammool:
> I found this:
> https://mail.gnome.org/archives/commits-list/2021-August/msg00550.html

That appears to be only doing something for generated documentation.

-- 
   Another bucket of what can only be described as human ordure hits ARTHUR.
ARTHUR: ... Right!  (to the KNIGHTS) That settles it!

PR #10989, opened 2022-08-26
CI: use -Wno-deprecated-declarations

pheiduck:


pheiduck:
not what we want to have

PR #10991, opened 2022-08-27
Modify getscriptinfo() to return script-local functions and variables when querying for a specific script

yegappan:


brammool:
>  * Modify getscriptinfo() to return script-local functions and
>  variables when querying for a specific script

How about adding "sid" to the {opts} argument, so that it's easy to
select one specific script.  Selecting by name would require using the
whole file name.  That might even have forward/backward slash issues.

Then, if we can use "sid", how about only returning functions and
variables when "sid" is used?  Since then we have only one script, while
with "name" it may still be a few (e.g. when finding autoload scripts
you would get functions and variables even though you don't need them).

For dict_copy() get_copyID() should be used.

-- 
"Software is like sex... it's better when it's free."
		-- Linus Torvalds, initiator of the free Linux OS
Makes me wonder what FSF stands for...?

vim-ml:
Hi Bram,

On Sat, Aug 27, 2022 at 2:50 AM Bram Moolenaar ***@***.***>
wrote:

>
> > * Modify getscriptinfo() to return script-local functions and
> > variables when querying for a specific script
>
> How about adding "sid" to the {opts} argument, so that it's easy to
> select one specific script. Selecting by name would require using the
> whole file name. That might even have forward/backward slash issues.
>
> Then, if we can use "sid", how about only returning functions and
> variables when "sid" is used? Since then we have only one script, while
> with "name" it may still be a few (e.g. when finding autoload scripts
> you would get functions and variables even though you don't need them).
>
> For dict_copy() get_copyID() should be used.
>
>
I have updated the PR to incorporate these comments.

Regards,
Yegappan

PR #10996, opened 2022-08-27
Hide the messages popup when entering the command line

puremourning:
Problem: Command line is obscured by notification popup when cmdheight=0
Solution: Hide the popup when entering command line

I'm not sure this patch is perfect, but steps to reproduce are below:

1. `vim --clean --cmd 'set cmdheight=0' `
2. `:echo 'this is a test'`
3. (within 3 seconds) type `:something`

Expect: `:something` is shown (i.e. the command line is visible)
Actual: the `this is a test` popup is drawn over the command line.

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -3187,6 +3187,13 @@ nv_colon(cmdarg_T *cap)
 	    }
 	}
 
+#ifdef HAS_MESSAGE_WINDOW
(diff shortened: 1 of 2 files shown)


brammool:
It was working OK for me...   Looks like I missed a file in the patch.  I'll merge this in.

PR #11002, opened 2022-08-28
coverity: use  CFLAGS: -Wno-error=deprecated-declarations to avoid bu…

pheiduck:
…ild error

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -10,6 +10,7 @@ jobs:
 
     env:
       CC: gcc
+      CFLAGS: -Wno-error=deprecated-declarations
       DEBIAN_FRONTEND: noninteractive
 
     steps:


brammool:
Let's just ignore that warning completely.

pheiduck:
> Let's just ignore that warning completely.
What do you mean by that?

brammool:
> > Let's just ignore that warning completely.
> What do you mean by that?

      CFLAGS: -Wno-error=deprecated-declarations

This means the warning is given, but not considered an error.

      CFLAGS: -Wno-deprecated-declarations

This means the warning is not given.  We should do that for the other
builds as well, because it obscures real problems.


-- 
Every time I lose weight, it finds me again!

pheiduck:
@brammool  https://github.com/vim/vim/pull/11003

PR #11006, opened 2022-08-29
feat: add godot game engine syntax/ftplugin/indent files

habamax:
Currently vim has detection of gdscript and gdresource files but actual
syntax/indent/ftplugin files are missing.

Additionally, gdshader detection is added to filetype.vim

Files are extracted and simplified from the plugin I maintain: https://github.com/habamax/vim-godot

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -698,6 +698,9 @@ au BufNewFile,BufRead *.gd			setf gdscript
 " Godot resource
 au BufRead,BufNewFile *.tscn,*.tres			setf gdresource
 
+" Godot shader
+au BufRead,BufNewFile *.gdshader,*.shader	setf gdshader
+
 " Gedcom
 au BufNewFile,BufRead *.ged,lltxxxxx.txt	setf gedcom
 
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -1,0 +1,54 @@
+vim9script
+
+# Vim filetype plugin file
+# Language: gdscript (Godot game engine scripting language)
+# Maintainer: Maxim Kim <habamax@gmail.com>
+
+if exists("b:did_ftplugin") | finish | endif
+
+b:did_ftplugin = 1
+b:undo_ftplugin = 'setlocal cinkeys<'
+      \ .. '| setlocal indentkeys<'
+      \ .. '| setlocal commentstring<'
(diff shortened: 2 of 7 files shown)


brammool:
I made a patch for the filetype recognition.

The syntax file gdresource.vim and gdscript.vim use line continuation, they should set and restore 'cpo'.
I'll add entries to CODEOWNERS

habamax:
Sure I will add it. Actually, I had it in the legacy version of the scripts, but then I "remembered" I was told for vim9script you don't have to save/restore cpo.

habamax:
Oh, those are in syntax files!

lacygoill:
As soon as a script starts with the `:vim9script` command, the usual "save/restore cpo" becomes useless.  It's only needed in legacy scripts.

---

Minor nitpick, but we might want to use `<C-\><C-N>` instead of `<Esc>` in these mappings:

    xmap <silent><buffer><expr> ]] "\<esc>" .. v:count1 .. ']]m>gv'
    xmap <silent><buffer><expr> [[ "\<esc>" .. v:count1 .. '[[m>gv'

This should work:

    xmap <buffer><expr> ]] $'<C-\><C-N>{v:count1}]]m>gv'
    xmap <buffer><expr> [[ $'<C-\><C-N>{v:count1}[[m>gv'

Rationale: In the RHS of a recursive mapping, `<Esc>` can give unexpected results.  For example:
```vim
vim9script
execute("set <F31>=\<Esc>d")
imap <C-B> <Esc>ddi
:0 put =['a', 'b', 'c']
:2
feedkeys("i\<C-B>", 'x')
```
Expected:  The buffer contains these lines:

    a
    c

Actual:  The buffer contains these lines:

    a
    <F31>dib
    c

`<C-\><C-N>` has the same semantics as `<Esc>` but is more reliable:
```vim
vim9script
execute("set <F31>=\<Esc>d")
imap <C-B> <C-\><C-N>ddi
:0 put =['a', 'b', 'c']
:2
feedkeys("i\<C-B>", 'x')
```
    a
    c

Also, I don't think we need `<silent>`, unless the mappings enter the command-line (they don't seem to do that here, but I might be wrong).  And we don't need double quotes and backslashes to encode `<Esc>` in the mappings; mapping commands like `:xmap` understand key notations without.

Again, all of this is nitpicking, but it might make the mappings a little more reliable and less verbose (thus more readable).

habamax:
thx @lacygoill, looks better, will update it

PR #10310, opened 2022-04-28
Fix termguicolors not to break ctermbg=NONE on vim.exe & Windows Terminal.

mattn:
Current implementation of vim.exe, ctermbg=NONE does not make transparent on Windows Terminal.

Repro steps

1. Start Windows Terminal
2. Set background image of cmd.exe to something.
3. Start vim.exe on the cmd.exe
4. `set termguicolor`

image

image

image

image

This change fixes background color not to set invalid color.

image

image

If the ctermbg is set to NONE of Normal highlight group, this works like below.

image

PR #11029, opened 2022-09-01
fix background colors on cmd

mattn:
Update #10310


brammool:
Included with patch 9.0.0344

PR #10045, opened 2022-03-30
add sytax recognition for lua utf8 funcs

nobodyatandnothing:

PR #11030, opened 2022-09-01
Add Lua string and comment delimiter syntax groups

dkearns:
It's not currently possible to distinguish between comment and string delimiters and the respective content of these syntax groups.

While some users like to highlight these delimiters differently my current motivation for this change is to more accurately match these delimiters in the Lua ftplugin when defining `match_words` and `]/` comment mappings.

Any objections to this being included @marcuscf?

### DIFF
--- a/runtime/syntax/lua.vim
+++ b/runtime/syntax/lua.vim
@@ -1,11 +1,11 @@
 " Vim syntax file
-" Language:	Lua 4.0, Lua 5.0, Lua 5.1 and Lua 5.2
-" Maintainer:	Marcus Aurelius Farias <masserahguard-lua 'at' yahoo com>
-" First Author:	Carlos Augusto Teixeira Mendes <cmendes 'at' inf puc-rio br>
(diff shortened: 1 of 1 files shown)


marcuscf:
> Any objections to this being included @marcuscf?

No objections as far as I can tell. I tested it too briefly to be of any real value, but it seems to work without any side effects.

While testing it I noticed that I had local changes related to strings too. If you intend to add more features to this syntax file you can ask me about them. For example, Lua >= 5.3 has has Unicode escapes in the form of `syn match  luaSpecial contained #\\u{[[:xdigit:]]\+}#`

marcuscf:
> @marcuscf, I'll send you a PR when this is merged here to sync you up. I'll also try and have a look at creating a PR from your repo as discussed in #10045 when I get some time in the next couple of weeks.

Thanks! If there are incompatible changes (e.g. the indent file) I can move them to another branch so that experimental changes don't cause trouble.

brammool:
I'll include it, thanks.

PR #11031, opened 2022-09-01
Allow assigning a Funcref to autoload variable

thinca:
In Vim, defining a User function that starts with a lowercase letter is not allowed because it could overwrite a built-in function.
However, defining an autoload function that starts with a lowercase letter is permitted because it does not cause the above problem.

For the same reason, Vim does not allow Funcref to be assigned to a global variable whose name begins with a lowercase letter.
This also applies to the autoload variable; `let foo#tr = function('tr')` is not allowed and must be `let Foo#tr = function('tr')`.
I see no need for this restriction.  `let foo#tr = function('tr')` should be allowed.

### DIFF
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -4012,10 +4012,12 @@ var_wrong_func_name(
 {
     // Allow for w: b: s: and t:.  In Vim9 script s: is not allowed, because
     // the name can be used without the s: prefix.
+    // Allow autoload variable.
     if (!((vim_strchr((char_u *)"wbt", name[0]) != NULL
 		    || (!in_vim9script() && name[0] == 's')) && name[1] == ':')
 	    && !ASCII_ISUPPER((name[0] != NUL && name[1] == ':')
-						     ? name[2] : name[0]))
+						     ? name[2] : name[0])
+	    && vim_strchr(name, '#') == NULL)
     {
 	semsg(_(e_funcref_variable_name_must_start_with_capital_str), name);
 	return TRUE;
--- a/src/testdir/test_let.vim
+++ b/src/testdir/test_let.vim
@@ -8,6 +8,10 @@ func Test_let()
   let Test104#numvar = function('tr')
   call assert_equal("function('tr')", string(Test104#numvar))
 
+  let foo#tr = function('tr')
+  call assert_equal("function('tr')", string(foo#tr))
+  unlet foo#tr
+
   let a = 1
   let b = 2
 


brammool:
> In Vim, defining a User function that starts with a lowercase letter
> is not allowed because it could overwrite a built-in function.
> However, defining an autoload function that starts with a lowercase
> letter is permitted because it does not cause the above problem.
>
> For the same reason, Vim does not allow Funcref to be assigned to a
> global variable whose name begins with a lowercase letter.
> This also applies to the autoload variable; `let foo#tr =
> function(&#39;tr&#39;)` is not allowed and must be `let Foo#tr =
> function(&#39;tr&#39;)`.
> I see no need for this restriction.  `let foo#tr =
> function(&#39;tr&#39;)` should be allowed.

Well, not when used with the "name#" prefix.  But inside the autoload
script it can be used without the prefix, which does cause shadowing a
builtin function.

Thus when you, like in the test, do:

	let foo#tr = function('tr')

Then in the "autoload/foo.vim" script calling tr() does have the
problem.  At least in Vim9 script, where the "s:" prefix is not used.
Right?

-- 
This computer is so slow, it takes forever to execute and endless loop!

brammool:
> In legacy Vim script, `name#` prefix is always necessary.
> 
> ```vim
> " autoload/name.vim
> 
> echo foo  " This is g:foo, not name#foo
> echo s:foo  " This is s:foo, not name#foo
> 
> function name#func()
>   echo foo  " This is l:foo, not name#foo
> endfunction
> ```

OK, and it actually checks if the name of the script starts with a
capital, not the variable itself, so that is pointless.
 
> In Vim9 script, as you say, exported variable is an autoload variable
> and can be access without prefix.
> But there is no problem.
> 
> ```vim
> vim9script
> # autoload/name.vim
> 
> # This occurs an error "E704: Funcref variable name must start with a capital: tr"
> export var tr = () => 'tr'
> ```
> 
> ```vim
> vim9script
> # autoload/name.vim
> 
> # This variable can rewrite by `let name#tr = { -> 'tr' }` from outside of script with this patch.
> export var tr = ''
> 
> def Foo()
>   # But this `tr()` is always built-in function.  Vim9 script is compiled first.
>   echo tr('foo', 'o', 'a')
> enddef
> ```

Hmm, I cannot think of a way to go around this, assigning a variable
outside of a Vim9 script can be expected to always fail.

-- 
BEDEVERE: Wait.  Wait ... tell me, what also floats on water?
ALL:      Bread?  No, no, no.  Apples .... gravy ... very small rocks ...
ARTHUR:   A duck.

PR #11032, opened 2022-09-01
fix compilation warning

mattn:


brammool:
Was done with patch 9.0.0353

PR #11033, opened 2022-09-01
fix echowindow

mattn:
Current implementation of `:echowindow`,

```
func Foo()
  echo 123
  return 'foo'
endfunc

echowindow Foo()
```
This code shows 123 and foo both on echowindow. I think 123 should be displayed on message area, and foo should be displayed on echowindow.


mattn:
Also 
```
:echowindow undefined_variable
```
This error message should be displayed on message area.

brammool:
Thanks.  Please add a regression test.

PR #11038, opened 2022-09-02
fix breakindent for briopt=list:-1

### LINKED ISSUE #11018: breakindent does not indent non-lists when breakindentopt=list:-1 is set
### Steps to reproduce

1. Start gVim via `gvim -i NONE -u NONE -U NONE -N`.
2. Paste the following content into the current buffer and resize the window to make sure that the long line wraps:
```cpp
int my_function() {
	const int long_name = 0;
	return long_name + long_name + long_name + long_name + long_name + long_name + long_name + long_name;
}
```
Wrapped contents
3. Issue `set breakindent`, the result looks as it should, which is as follows:
SetBreakindent
4. Issue `set breakindentopt=list:-1`, this removes the indent even though it should not in my understanding:
SetBreakindentopt

### Expected behaviour

Using `set breakindentopt=list:-1` should not break indentation for non-lists. (This issue does not seem to be exclusive to C++ code.)

### Version of Vim

9.0.0313

### Environment

openSUSE Leap 15.3

### Logs and stack traces

_No response_

habamax:
Fixes #11018

### DIFF
--- a/src/testdir/test_breakindent.vim
+++ b/src/testdir/test_breakindent.vim
@@ -848,6 +848,26 @@ func Test_breakindent20_list()
   let lines = s:screen_lines2(1, 6, 20)
   call s:compare_lines(expect, lines)
 
+  " check breakindent works if breakindntopt=list:-1
+  " for a non list content
+  %delete _
+  call setline(1, ['  Congress shall make no law',
+        \ '    Congress shall make no law',
+        \ '     Congress shall make no law'])
+  norm! 1gg
+  setl briopt=min:5,list:-1
+  setl showbreak=
+  redraw!
+  let expect = [
+	\ "  Congress shall    ",
+	\ "  make no law       ",
+	\ "    Congress shall  ",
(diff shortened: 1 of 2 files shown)


src/testdir/test_breakindent.vim
```diff
@@ -831,6 +831,44 @@ func Test_breakindent20_list()
...
+	\ ">   make no law     ",
+	\ "*** Congress shall  ",
+	\ ">     make no law   ",
+	\ "**** Congress shall ",
+	\ ">      make no law  ",
+	\ ]
+  let lines = s:screen_lines2(1, 6, 20)
+  call s:compare_lines(expect, lines)
+
+  " check breakindent works if breakindntopt=list:-1
```
acc987: I noticed a typo (`breakindntopt` → `breakindentopt`) in this comment, in case you want to fix it.

habamax: thx

PR #11053, opened 2022-09-04
Update colorschemes

habamax:
- Add MessageWindow
- Add PopupNotification
- Tune WarningMsg to be more readable
- Tune lunaperche PHP and Pmenu colors


brammool:
I'll include it, thanks.

PR #11054, opened 2022-09-04
Improve gdscript files

habamax:
- improve xmaps for next/prev section
- add website info

### DIFF
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -3,6 +3,7 @@ vim9script
 # Vim filetype plugin file
 # Language: gdscript (Godot game engine scripting language)
 # Maintainer: Maxim Kim <habamax@gmail.com>
+# Website: https://github.com/habamax/vim-gdscript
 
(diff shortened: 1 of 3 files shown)


brammool:
I'll include it, thanks.

runtime/ftplugin/gdscript.vim
```diff
@@ -49,5 +50,5 @@ enddef
```
dkearns, line 50: These mappings should be guarded by tests for the `no_plugin_maps` and `no_gdscript_maps` global variables to allow  users to prevent their definition.  This is documented at `:help no_plugin_maps`.

The mappings should also be removed in the `undo_ftplugin` command.

See  and see https://github.com/vim/vim/blob/806a273f3c84ecd475913d901890bb1929be9a0a/runtime/ftplugin/zimbu.vim#L188-L194 for an example.

habamax: Missed the undo part indeed. I will also check `no_plugin_maps` and I would like to not introduce any additional vars.

PR #11058, opened 2022-09-05
Use early return to simplify/refactor code in alloc.c and arglist.c

yegappan:


src/arglist.c
```diff
@@ -108,22 +108,22 @@ alist_expand(int *fnum_list, int fnum_len)
     // Don't use 'suffixes' here.  This should work like the shell did the
     // expansion.  Also, the vimrc file isn't read yet, thus the user
     // can't set the options.
```
zeertzjq: It seems this comment also needs to be moved.

yegappan: Yes. Updated the PR to move the comment.

PR #11072, opened 2022-09-07
CI: Fix that test_gui may hang

k-takata:
test_gui often hangs on CI:
https://github.com/vim/vim/runs/8210640448?check_suite_focus=true
https://github.com/vim/vim/runs/8211225394?check_suite_focus=true

It seems that this hangs in Test_gui_dialog_file() when Xdialfile exists. If Xdialfile exists, the following two messages are shown:

```
Question: Save changes to "Xdialfile"?
Question: Overwrite existing file "Xdialfile"?
```

The default answer for the second message is "No", so gvim won't be closed and the test will stop.

Not sure why Xdialfile can be left behind on the CI, but deleting this seems to solve the issue.

Also use the 'D' flag for writefile().

### DIFF
--- a/src/testdir/test_gui.vim
+++ b/src/testdir/test_gui.vim
@@ -1607,12 +1607,13 @@ func Test_gui_CTRL_SHIFT_V()
 endfunc
 
 func Test_gui_dialog_file()
+  call delete('Xdialfile')
   let lines =<< trim END
     file Xdialfile
     normal axxx
     confirm qa
   END
-  call writefile(lines, 'Xlines')
+  call writefile(lines, 'Xlines', 'D')
   let prefix = '!'
   if has('win32')
     let prefix = '!start '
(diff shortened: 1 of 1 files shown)


brammool:
> test_gui often hangs on CI:
> https://github.com/vim/vim/runs/8210640448?check_suite_focus=true
> https://github.com/vim/vim/runs/8211225394?check_suite_focus=true
> 
> It seems that this hangs in Test_gui_dialog_file() when Xdialfile
> exists. If Xdialfile exists, the following two messages are shown:

This has been annoying me for a while.  How did you find out what the
problem was?  I wish that when a test times out the CI would make a
screenshot, so we can see where it got stuck.


-- 
23. You can't call your mother... she doesn't have VOIP

k-takata:
I found it by running the tests locally. I don't know if we can find out it on the CI...

brammool:
It appears old message suddenly show up....

PR #11078, opened 2022-09-07
unused parameter 'lnum' delete it.

pheiduck:


pheiduck:
breaks more things as it fix :/

PR #11085, opened 2022-09-08
Add gitattributes file type & syntax

ObserverOfTime:
The file names & syntax are defined in https://git-scm.com/docs/gitattributes.


brammool:
The syntax file is using line continuation, thus it should set 'cpo' and restore it.

PR #6091, opened 2020-05-16
win32: Drop support for XP and Vista

k-takata:
Microsoft has already dropped support for Windows 7 and earlier.
Vim 9 will be a good timing for dropping support for old Windows.
This patch drops support for XP and Vista, because Windows 7 may still
have quite a few users,

PR #11089, opened 2022-09-09
win32: Drop support for XP and Vista

k-takata:
I tried this in #6091, but it was rejected.
Two years past and Vim 9 has been released. It should be a good timing to drop support for 20-year old OS.

### DIFF
--- a/src/INSTALLpc.txt
+++ b/src/INSTALLpc.txt
@@ -12,7 +12,7 @@ this, then you will get the default behavior as is documented, which should
 be fine for most people.
 
 This document assumes that you are building Vim for Win32 or later (Windows
-XP/2003/Vista/7/8/10).  There are also instructions for pre-XP systems, but
+7/8/10/11).  There are also instructions for pre-Vista and pre-XP systems, but
 they might no longer work.
 
 The recommended way is to build a 32 bit Vim, also on 64 bit systems.  You can
@@ -42,9 +42,9 @@ Contents:
 
 
 The currently recommended way (that means it has been verified to work) is
-using the "Visual Studio Community 2015" installation.  This includes the SDK
-needed to target Windows XP.  But not older Windows versions (95, 98), see
-"OLDER VERSIONS" below for that.
+using the "Visual Studio Community 2015" installation.  This doesn't include
+the SDK for older Windows versions (95, 98), see "OLDER VERSIONS" below for
+that.
 
 
 1. Microsoft Visual C++
@@ -57,16 +57,17 @@ execute the installer from it.
 
 When installing "Visual Studio Community 2015 with Update 3" or "Visual C++
 Build Tools for Visual Studio 2015 with Update 3" make sure to
-select "custom" and check "Windows XP Support for C++" and all checkboxes
-under "Universal Windows App Development Tools".  Or whatever they are called
-now.
+select "custom" and check all checkboxes under "Universal Windows App
+Development Tools".  Or whatever they are called now.
+(If you still want to target Windows XP, check also "Windows XP Support for
+C++".  Note that this is no longer supported.)
 
 
 Visual Studio
 -------------
 
-Building with Visual Studio (VS2010, VS2012, VS2013, VS2015, VS2017 and VS2019)
-is straightforward.
+Building with Visual Studio (VS2010, VS2012, VS2013, VS2015, VS2017, VS2019
+and VS2022) is straightforward.
 
 To build Vim from the command line with MSVC, use Make_mvc.mak.
 Visual Studio installed a batch file called vcvars32.bat, which you must
@@ -103,6 +104,9 @@ Vim with Make_mvc.mak.
 
 Targeting Windows XP with MSVC 2012 and later      *new-msvc-windows-xp*
 ---------------------------------------------
+
+(The support for pre-Vista was removed in patch 9.0.0xxx.  If you want to
+target Windows XP, use the source code before that.)
 
 Beginning with Visual C++ 2012, Microsoft changed the behavior of LINK.EXE
 so that it targets Windows 6.0 (Vista) by default.  In order to override
@@ -141,9 +145,6 @@ compiler by using the "x64" option:
 The following Visual C++ team blog can serve as a reference page:
     http://blogs.msdn.com/b/vcblog/archive/2012/10/08/windows-xp-targeting-with-c-in-visual-studio-2012.aspx
 
-VC 2019 dropped support for targeting Windows XP.  If you want a binary that
-targeting Windows XP, use VC 2017 or earlier.
-
 
 Cross compile support for Windows on ARM64
 ------------------------------------------
@@ -156,7 +157,7 @@ The ARM64 support was provided by Leendert van Doorn.
 
 OLDER VERSIONS
 
-The minimal supported version is Windows XP. Building with older compilers
+The minimal supported version is Windows 7. Building with older compilers
 might still work, but these instructions might be outdated.
 
 If you need the executable to run on Windows 98 or ME, use the 2005 one
@@ -774,12 +775,12 @@ config.h and Ruby's DLL name. Here are the steps for working around them:
       nmake -f Make_mvc.mak
           RUBY=C:\Ruby24 DYNAMIC_RUBY=yes RUBY_VER=24 RUBY_API_VER_LONG=2.4.0
           RUBY_MSVCRT_NAME=msvcrt
-          WINVER=0x501
+          WINVER=0x601
 
     For 64-bit version, replace RUBY=C:\Ruby24 with RUBY=C:\Ruby24-x64.
 
     If you set WINVER explicitly, it must be set to >=0x500, when building
-    with Ruby 2.1 or later.  (Default is 0x501.)
+    with Ruby 2.1 or later.  (Default is 0x601.)
     When using this trick, you also need to set RUBY_MSVCRT_NAME to msvcrt
     which is used for the Ruby's DLL name.
 
@@ -790,11 +791,11 @@ After you install RubyInstaller, just type this (as one line):
 
     mingw32-make -f Make_ming.mak
         RUBY=C:/Ruby24 DYNAMIC_RUBY=yes RUBY_VER=24 RUBY_API_VER_LONG=2.4.0
-        WINVER=0x600
+        WINVER=0x601
 
 For 64-bit version, replace RUBY=C:/Ruby24 with RUBY=C:/Ruby24-x64.
 If you set WINVER explicitly, it must be set to >=0x500, when building with
-Ruby 2.1 or later.  (Default is 0x600.)
+Ruby 2.1 or later.  (Default is 0x601.)
 
 
 
--- a/README.md
+++ b/README.md
@@ -28,10 +28,10 @@ All commands are given with normal keyboard characters, so those who can type
 with ten fingers can work very fast.  Additionally, function keys can be
 mapped to commands by the user, and the mouse can be used.
 
-Vim runs under MS-Windows (XP, Vista, 7, 8, 10), macOS, Haiku, VMS and almost
-all flavours of UNIX.  Porting to other systems should not be very difficult.
-Older versions of Vim run on MS-DOS, MS-Windows 95/98/Me/NT/2000, Amiga DOS,
-Atari MiNT, BeOS, RISC OS and OS/2.  These are no longer maintained.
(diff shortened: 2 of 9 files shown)


src/INSTALLpc.txt
```diff
@@ -104,6 +105,9 @@ Vim with Make_mvc.mak.
 Targeting Windows XP with MSVC 2012 and later      *new-msvc-windows-xp*
 ---------------------------------------------
 
+(The support for pre-Vista was removed in patch 9.0.0xxx.  If you want to
+target Windows XP, use the source code before that.)
```
k-takata, line 109: This part should be updated accordingly.

brammool:
Ken Takata wrote:

> I tried this in #6091, but it was rejected.
> Two years past and Vim 9 has been released. It should be a good timing
> to drop support for 20-year old OS.

I'm in doubt.  This is not changing much code.  It's not like with older
windows systems where we could drop a substantial part of the library
calls.  That was a nice cleanup.  This is only a small change.

Windows XP usage did go down quite a lot.  Windows Vista doesn't even
show up anymore (nobody liked it anyway).  According to Statcounter it
is 0.38%  (to compare: Windows 10 is 72 %).  However, this is measured
by accessing web pages, computers not used to browse the internet are
not counted.

I wonder, is anyone reading this still using a Windows XP system?
Even if it is to drive some hardware.

-- 
This planet has -- or rather had -- a problem, which was this: most
of the people living on it were unhappy for pretty much of the time.
Many solutions were suggested for this problem, but most of these
were largely concerned with the movements of small green pieces of
paper, which is odd because on the whole it wasn't the small green
pieces of paper that were unhappy.

k-takata:
You are right that this doesn't change much code.
But actually I don't want to care about Windows XP anymore when adding a new feature.
I don't have a working XP machine, so I can't check if the new code works on XP.
(At least, I want to remove XP from the list of the supported OS.)

brammool:
I don't see any thumbs-up or thumbs-down on this PR, nobody cares?

koron:
I stand by k-takata's opinion, support of Windows XP should be dropped.

Lack of reactions (thumb-up/down) might indicate that
there are few people are intereted in Windows XP and its support.
Fewer developer can maintain code with Windows XP in mind, 
and it would be a waste of their precious time for few XP users.

> However, this is measured by accessing web pages, computers not used to browse the internet are not counted.

A system that is still running Windows XP is probably a system that does not need to be connected to the Internet,
and it is hard to imagine using the latest Vim on such a system.
If they did use Vim, they would probably use an older version.

pheiduck:
Since there are still infrastructures where XP is used, I would leave the support in. But I would welcome it if we could drop it soon.

skywind3000:
Is there a strong advantage to drop win XP ?

XP still lives in many low-end hardwares (not only PC), which is still widely used in under developed countries. 
So far as I know, especially in schools, hospitals and factories.

Those silent majority are not vim enthusiasts like us and will not access vim.org everyday. Also they are not working 
in big tech companies and will not update their laptops every two years.

But I still wish they can benefit from the latest vim versions. 

Since 9.0 is a big step since 8.0, many future plugins will depend on this.

So, if XP must be dropped, can we just put off this discussion until vim 9.1 or 9.2 ?

brammool:
> Is there a strong advantage to drop win XP ?
> 
> XP still lives in many low-end hardwares (not only PC), which is still
> widely used in under developed countries. 
> So far as I know, especially in schools, hospitals and factories.
> 
> Those silent majority are not vim enthusiasts like us and will not
> access vim.org everyday. Also they are not working 
> in big tech companies and will not update their laptops every two years.

Yes, that is the reason I'm hesitant to just drop it.

> But I still wish they can benefit from the latest vim versions. 
> 
> Since 9.0 is a big step since 8.0, many future plugins will depend on this.
> 
> So, if XP must be dropped, can we just put off this discussion until
> vim 9.1 or 9.2 ?

Would it be sufficient to say that Windows XP users can get the 9.0
release?  They won't be able to use plugins that use the features being
added now (virtual text, :defer, etc.), but isn't it likely that it
won't matter much?

Waiting for 9.1 is just moving it forward, it's not really much
different.

-- 
40. You tell the cab driver you live at
    http://123.elm.street/house/bluetrim.html
41. You actually try that 123.elm.street address.

skywind3000:
> Would it be sufficient to say that Windows XP users can get the 9.0
> release?  They won't be able to use plugins that use the features being
> added now (virtual text, :defer, etc.), but isn't it likely that it
> won't matter much?

The whole 9.0 lifespan (including all patches before 9.1) would be better.
Many users (me too) just start using vim9script after 9.0 got released.

I believe all the 9.0.XXXX versions can be used for collecting feedback and making vim9script more robust.
So that makes it more important than 9.1.XXXX or 9.2.XXXX.

btw: virtual-text is another important feature, it enables some cool plugins like 
vim-copilot, LSP inlay hints and gitlens.

arp242:
> Is there a strong advantage to drop win XP ?

The biggest advantage I see in dropping XP and Vista support is that Vim can then probably also update to a more recent minimal MSVC version so we can get full/proper C99 support.
 
> XP still lives in many low-end hardwares (not only PC), which is still widely used in under developed countries. So far as I know, especially in schools, hospitals and factories.
> 
> Those silent majority are not vim enthusiasts like us and will not access vim.org everyday. Also they are not working in big tech companies and will not update their laptops every two years.
> 
> But I still wish they can benefit from the latest vim versions.

Python dropped Windows XP somewhere around 2015. The newest versions don't even seem to work on Windows 7, according to https://www.python.org/downloads/windows/

NodeJS dropped support in v6, [in 2015](https://github.com/nodejs/node/issues/3804).

Go dropped Windows XP support in [1.11](https://go.dev/doc/go1.11), in 2018.

Java dropped support for XP [in 2014](https://www.zdnet.com/article/java-support-over-for-windows-xp/).

Java, Python, NodeJS, and Go are probably the most popular platforms, especially for new programmers. I bet the situation is similar for most languages/interpreters/compilers.

Things like web browsers and other editors nad tooling have dropped XP support as well. Getting dev environment is still possible of course, but with old version of a lot of software. You're going to have a hard time. Many platforms have already started to drop Windows 7 support.

In another comment you mentioned "virtual-text is another important feature, it enables some cool plugins like vim-copilot, LSP inlay hints and gitlens", but I bet you'll have a hard time getting things like CoPilot (depends on NodeJS) and many LSP things to run on Windows XP.

I'm sympathetic to your argument, but I think at this point we can *probably* safely drop Windows XP and Vista support and no one will notice, because the barriers to run latest version of $anything are already significant. When I lived in Indonesia up to last year I saw a number XP systems: the point-of-sale system at the family store next door, the admin system at my dentist, etc. But no one is trying to run Vim on those machines. At my job (in the Netherlands) we used a Windows 95 system for some very specific software up to about 2016 when it was finally replaced, but no one tried to run Vim on that either.

ychin:
When talking to Vim users, I somewhat frequently run into people who think Vim is a static piece of software and only use old ancient versions of Vim from at least few years ago, and these are people who otherwise stay up to date with their software. I kind of have a feeling that people still on Windows XP, if they even use Vim to begin with, would be on that camp and not necessarily trying to download the latest and greatest. I can't say that number is 0  (see below about MacVim), but I can't imagine the number of active up-to-date Vim users on Windows XP is high today.

> The biggest advantage I see in dropping XP and Vista support is that Vim can then probably also update to a more recent minimal MSVC version so we can get full/proper C99 support.

I feel like this is a pretty big improvement if it's true that dropping XP/Vista supports allows for C99. The requirement to satisfy old-school C-style "variable declarations on top" can make code contributions to Vim feel unnecessarily annoying especially for newcomers, and adds to the sentiment that Vim is more archaic than it should be. It is also a potential vector for bugs as we have function-scope variables everywhere, instead of being able to use C scoping to limit the visibility of variables. I definitely know from personal experience that this has bit me before when doing Vim changes.

> Things like web browsers and other editors nad tooling have dropped XP support as well. Getting dev environment is still possible of course, but with old version of a lot of software. You're going to have a hard time. Many platforms have already started to drop Windows 7 support.

Just looking around, I have found 2 main programs that still support Windows XP. I'm not sure if it's because of development philosophy, or whether they actually have enough active users to justify so:

1. VLC, which *is* a popular tool, and playing videos probably has a larger audience than Vim.
2. 7-zip. It supports down to Windows 2000!! [Release history](https://www.7-zip.org/history.txt) indicates that Windows requirements was last bump in 2013 and dropped 95/98/ME.

On a tangent, I am facing a similar issue with MacVim as Apple has been aggressively removing support from older OS versions (although they operate on a much shorter timeline than Microsoft/Windows), so I can sympathize. MacVim, outside of VLC and a couple other ones like Unarchiver (another unzipping tool like 7-zip), is the only macOS software that I could find that supports back to 10.9.  Every time I think I can drop it there are always some users who come out and let me know that they are still using it…

cecamp:
I'm uncomfortable with the current process for dropping systems/features/os/etc. The common approach is to mark something as deprecated. That, too, is problematic, as it seems that there are plenty of folks that continue to use deprecated stuff forever. So, may I suggest a "+deprecated" feature for configure to include deprecated stuff -- up to a point. Those who are using the latest vim on deprecated h/w and o/ses will likely pipe up, giving you feedback. Once there's no or little feedback after, say, half a year, go ahead and remove the feature. Just an idea...

brammool:
> I'm uncomfortable with the current process for dropping
> systems/features/os/etc. The common approach is to mark something as
> deprecated. That, too, is problematic, as it seems that there are plenty
> of folks that continue to use deprecated stuff forever. So, may I
> suggest a "+deprecated" feature for configure to include deprecated
> stuff -- up to a point. Those who are using the latest vim on deprecated
> h/w and o/ses will likely pipe up, giving you feedback. Once there's no
> or little feedback after, say, half a year, go ahead and remove the
> feature. Just an idea...

I don't see use for it.  The idea of marking something "deprecated" is
to give people time to rewrite anything that depends on it.  In this
case there is nothing that one can do.  Well, upgrade to Windows 7, but
if that would be an option it would very likely already have been done.
I can't imagine that someone finally upgrades because Vim no longer
supports the OS.

In this case one can just use the last version of Vim that supported XP,
which is the current one.  This is hardly a problem, only some plugins
developed later might not work.

For those things that users actually use, we just keep them.  We only
drop something when it's not used (e.g. when it's broken anyway).
Unlike much other software where old script just won't work anymore
(Python 2 is a disappointing example, I still have a very large Python 2
program that will be painful to migrate to Python 3, if at all
possible; I started on it and failed miserably so far).

-- 
59. Your wife says communication is important in a marriage...so you buy
    another computer and install a second phone line so the two of you can
    chat.

skywind3000:
@arp242 , MSVC compiler version has nothing to do with XP compatibility. It depends on what headers/libraries you are using.

You can use the latest MSVC and select v141_xp as the target toolset in the option dialog to build XP applications. See here:

https://docs.microsoft.com/en-us/cpp/build/configuring-programs-for-windows-xp?view=msvc-170

MinGW can do similar thing:

> To compile for older Windows versions with MinGW you need to define _WIN32_WINNT to an older version for (0x0501 for Windows XP) when compiling your application as well as all of it's dependencies, including the standard library provided by MinGW.

ychin:
> @arp242 , MSVC compiler version has nothing to do with XP compatibility. It depends on what headers/libraries you are using.

I think the issue is with whether you want XP/Vista users to be able to build binaries themselves. From looking it up, it seems like Visual Studio 2013 started to [support C99](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/hh409293(v=vs.120)), but it's also the version that started to [require Windows 7](https://docs.microsoft.com/en-us/visualstudio/releases/2013/vs2013-sysrequirements-vs). It's possible to generate Windows XP builds on later Visual Studio versions but the users won't be able to build Vim themselves.

brammool:
Considering the comments, I think it's fine to drop XP support now.  Those few who want to use Vim on an (old) XP system can get Vim version 9.0.  It's easy to understand and provides a Vim version with plenty of features.  Over time some plugins won't work there, which is a very small price to pay.

brammool:
> Since there are still infrastructures where XP is used, I would leave
> the support in. But I would welcome it if we could drop it soon.

But what would "soon" be...?

There might be one other advantage: I believe MSVC does not support
Windows XP in later versions.  I still use MSVC 2015 because I think
later versions create a binary that doesn't run on older Windows
systems.  Not sure which ones exactly.

So we could switch to a more recent build setup, where Windows 7 is the
oldest Windows version supported.  I suppose newer MSVC verions are
better than older ones.

-- 
Don't read everything you believe.

PR #11090, opened 2022-09-09
Make repeat() accept Blob

Bakudankun:
There is no way to make a large Blob with a simple step, so I made `repeat()` accept a Blob.
Can be very fast when making a zero-filled Blob.


brammool:
Probably making "slen" and "len" a size_t avoids that LGTM warning.  Change "len <= 0" to "len == 0" then.
Better bail out early if "n <= 0".

Bakudankun:
@brammool 
I think the LGTM warning is about already fixed things, isn't it?
It says "1 alert when merging `b5e02f1`", which does not include the latest commit `df37334`.

Bakudankun:
The latest analysis says "No new or fixed alerts".
https://lgtm.com/projects/g/vim/vim/rev/pr-75992f9a88e876ce0720953b3f496dcfb01da6ef

Bakudankun:
Bailing out when n <= 0 sounds reasonable.
I'll fix that.

PR #11095, opened 2022-09-09
CI: Fix that test_autocmd is still flaky

k-takata:
It seems that test_autocmd is still flaky on Windows.
Add some more sleeps.

### DIFF
--- a/src/testdir/test_autocmd.vim
+++ b/src/testdir/test_autocmd.vim
@@ -763,7 +763,7 @@ func Test_autocmd_bufwipe_in_SessLoadPost()
 
   call writefile(content, 'Xvimrc', 'D')
   call system(GetVimCommand('Xvimrc') .. ' --not-a-term --noplugins -S Session.vim -c cq')
-  sleep 50m
+  sleep 100m
   let errors = join(readfile('Xerrors'))
   call assert_match('E814:', errors)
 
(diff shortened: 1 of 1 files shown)


brammool:
I think that one of the main reasons MS-Windows tests are more flaky then on other platforms, is that we run the gvim and vim tests in parallel.  We often see some mysterious results, such as "Can't open file Xerrors".  It might very well be that the other run deleted it.  This is a "should not happen" since the runs use different directories.  But it looks like it does happen.  After renaming a lot of commonly named files it did get less flaky.
For tests that use the clipboard this is certainly a problem, since there is only one clipboard.
Maybe we better split the CI runs for vim and gvim, and not run tests in parallel?  I assume each run uses a separate container and doesn't interfere with the other.  Or does it?

k-takata:
There were two reasons when I made them run in parallel:
1. The number of total concurrent jobs is limited to 20 in GHA: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits
2. When built with VIMDLL=yes, both gvim and vim are created. Testing both of them in serial takes a longer time.

I thought this worked nicely, but now, the flakyness is a headache issue. It might be better to stop running them in parallel even if the test time becomes longer.

k-takata:
I created #11101, so analyzing that is low priority now, I think.
I will look into it if I have time.

PR #11099, opened 2022-09-09
Add gitattributes syntax

ObserverOfTime:
Follow-up to #11085


brammool:
i'll include it, thanks.

PR #11101, opened 2022-09-10
CI: Revise windows tests

k-takata:
* Stop running GUI and console tests parallel in a single job.
* Add TINY features
* Add VIMDLL=no
* Reduce the test combination by using the L8(2^7) orthogonal array.


brammool:
I would appreciate if someone with understanding of the CI config have a look at this.  @pheiduck ?

pheiduck:
LGTM

brammool:
Looks like the slowest run on MS-Windows is 11 minutes, that's not so bad.

PR #11104, opened 2022-09-10
update gdscript files

habamax:
- Add `no_plugin_maps`
- undo ft maps
- Improve/simplify gdscript syntax


brammool:
I'll include this, thanks.

PR #11105, opened 2022-09-10
Add crontab filetype plugin

keith:
This sets the correct 'commentstring' for crontab files.


keith:
Thanks, updated!

brammool:
I'll include the latest.

PR #11106, opened 2022-09-11
CI: Work around FreeBSD 12.3 failure

k-takata:
`Test_closure_in_for_loop_fails` always fails on FreeBSD 12.3.
Skip it on FreeBSD 12.3 for now.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,6 +11,9 @@ freebsd_task:
     - name: FreeBSD 12.3
       freebsd_instance:
         image_family: freebsd-12-3
+      env:
+        # FIXME: The following test is disabled for now.
+        TEST_SKIP_PAT: Test_closure_in_for_loop_fails
   only_if: $CIRRUS_TAG == ''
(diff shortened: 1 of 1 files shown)


brammool:
Ken Takata wrote:

> `Test_closure_in_for_loop_fails` always fails on FreeBSD 12.3.
> Skip it on FreeBSD 12.3 for now.

It's actually strange that it passes on other systems.  The check that
startup is done expects "All" or "Top" to show, but it doesn't, because
there is an error message.  We should wait for the screendump to match.
Perhaps the "All" shows for a brief moment and the test sees that before
the error overwrites it.

-- 
Nothing is fool-proof to a sufficiently talented fool.

pheiduck:
We don't need this anymore, it's fixed with this I think:
https://github.com/vim/vim/commit/c069edeab28bdd7289d7096cd92804cb54841557

k-takata:
@brammool Thank you for fixing this. Closing.

Issue #11110, opened 2022-09-11
Color issues on GTK gui builds when copying text to primary clipboard


valacar:
### Steps to reproduce

1. Build both GTK2 and GTK3 builds of gui vim (GVim)
2. Run `vim --clean -g +"colo desert" +"hi"`
3. Make random selection of text on the `:highlight` or `:version` screen.
4. Notice changes in colors. I'll provide zoomed in screenshots below.
5. On the GTK3 build you will mainly see text colors changing after the selection disappears. For example, white text on a dark gray background (of the desert colorscheme) will change to a muddy rainbow-like color. In the GTK3 screenshot below, this is shown in the top right section.
6. On GTK2 you will have to zoom in close, but if your text is subpixel anti-aliased, you will notice sort of a rainbow outline text when it's being selected.

I will send a pull request fixing this problem shortly after reporting this issue. Here are some screenshots...

![vim-invert-fix-gtk3](https://user-images.githubusercontent.com/8539935/189547983-b143f29b-892e-4b51-ba82-ea1200677a6a.png)

![vim-invert-fix-gtk2](https://user-images.githubusercontent.com/8539935/189547991-d21bdf57-7310-4c5b-8301-0cfdb3f85d99.png)

### Expected behaviour

In the GTK3 build, I expected the pixel colors to be restored properly after making a selection. And in the GTK2 build, I expected the selection colors to not be altered quite as harshly when selected.

### Version of Vim

9.0.0445

### Environment

Operating System: Linux (Manjaro with XFCE)

Using GVim here, so terminal, and shell shouldn't matter.

valacar:
Thank you for accepting my pull request. I will close this now since it's fixed in [9.0.0454](https://github.com/vim/vim/releases/tag/v9.0.0454).

PR #3300, opened 2018-08-08
getftype() was not tested with fifo, symlinks , block and character device files

dpelle:
This PR improves test coverage of getftype(): getftype() was not
tested for fifo, symlinks, block and character device files. See:

https://codecov.io/gh/vim/vim/src/f8f88f89e12df516c1fac5851b504238ebc1d2d4/src/evalfunc.c#L5097

The PR still does not test with socket files as I don't think that there is
a an easy way to create socket files types.

PR #11111, opened 2022-09-11
Fix color issues with GTK2 and GTK3 invert rectangle function

valacar:
Fix for issue #11110

### DIFF
--- a/src/gui_gtk_x11.c
+++ b/src/gui_gtk_x11.c
@@ -6281,13 +6281,9 @@ gui_mch_invert_rectangle(int r, int c, int nr, int nc)
     if (gui.drawarea->window == NULL)
 	return;
 
-    values.foreground.pixel = gui.norm_pixel ^ gui.back_pixel;
-    values.background.pixel = gui.norm_pixel ^ gui.back_pixel;
-    values.function = GDK_XOR;
+    values.function = GDK_INVERT;
     invert_gc = gdk_gc_new_with_values(gui.drawarea->window,
 				       &values,
-				       GDK_GC_FOREGROUND |
-				       GDK_GC_BACKGROUND |
 				       GDK_GC_FUNCTION);
     gdk_gc_set_exposures(invert_gc, gui.visibility !=
 						   GDK_VISIBILITY_UNOBSCURED);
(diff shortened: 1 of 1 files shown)


brammool:
Does this work with all kinds of colorschemes?
I wonder why it was using the norm_pixel and back_pixel combined before.

valacar:
Yes it should work for all colorschemes. I tested all the default colorschemes that come with Vim and a few others, and didn't see any issues in either GTK2 or GTK3 builds. The colors are simply inverted so a color like #ff00ff becomes #00ff00, and #336699 becomes #cc9966. The operation is `abs(0xffffff - color)`. Applying the same operation again gives back the original color.

As for the previous XORing of norm_pixel and back_pixel, I'm not sure, but visually it appears to change the selection's background color to the Normal highlight group's foreground color, and then the text colors inside are inverted. I suppose it might fit better with the colorscheme, but inverting the colors (with my patch) usually gives a similar appearance.

Here's a screenshot comparison of the unpatch and patched version on GTK3 using a few stock colorschemes; it should be the same on GTK2.

vim-patch-compare-colorschemes

The main difference seems to be when using a black background with a non-white foreground like the murphy or elflord colorscheme.

brammool:
So why did the original problem happen?  Because the operation applied could not be 100% reverted?  I assume an inversion always works to get back the original color when applied twice, which is why this fix uses that.  Or was the original problem a round-off error?

valacar:
Yeah, it's just not properly being restored. I think the math is just wrong for GTK3. I found a rather extreme case using the default color scheme, but changing the Normal highlight group to green on gray. I started Vim with:

```
vim -g --clean +"hi Normal guifg=#00ff00 guibg=#999999" +"hi"
```

Here the background is `#999999`, and selecting turns it to `#003300` (dark green). The math is:

```
orig_color = 0x999999;
xor_mix = 0x00ff00 ^ 0x999999;  // = 0x996699, Normal highlight group's foreground xor'd with background
selected_color = abs(orig - xor_mix);  // = 0x003300 (dark green)
```

Then to restore the pixels colors when it's unselected, the operation is performed again over the top of the selected color (`#003300`) in the previous calculation. We're wanting to get back the original background color of `#999999`, but instead we get `#993399` (purple):

```
unselected_color = abs(selected_color - xor_mix); // = 0x993399 (purple)
```

vim-gtk3-bad-colors

The purple blocks in the screenshot should be restored back to the original gray background, but they're not.

I haven't really looked into the reason behind the GTK2 color issues yet.

PR #11112, opened 2022-09-12
Make argument check functions consistently use OK and FAIL

zeertzjq:


src/typval.c
```diff
@@ -746,7 +746,7 @@ check_for_string_or_list_or_blob_arg(typval_T *args, int idx)
 check_for_opt_string_or_list_arg(typval_T *args, int idx)
 {
     return (args[idx].v_type == VAR_UNKNOWN
-	    || check_for_string_or_list_arg(args, idx));
+	    || check_for_string_or_list_arg(args, idx) != FAIL);
```
k-takata: ` ? OK : FAIL` should be added?

zeertzjq: Oh, missed this one. Thanks.

PR #11114, opened 2022-09-12
Implement keytrans()

zeertzjq:


zeertzjq:
Done code and tests. TODO: documentation.

zeertzjq:
> TODO: documentation.

Done.

vds2212:
On Windows it seems that it doesn't work perfectly:
`<C-b>` seems to be represented by `<9b><fc>^DB` where "\<*C-b>" is `<80><fc>^DB`

`keytrans('<9b><fc>^DB')` returns `<9b><fc><C-D>B`
`keytrans('<80><fc>^DB')` returns `<C-B>`

Remark: `"\<*C-b>"` and `"\<*C-B>"` are the same which seems to be odd

PR #11116, opened 2022-09-12
Update Lua syntax file

dkearns:
This updates the syntax file to the latest available from https://github.com/marcuscf/vim-lua

I've also added b:undo_indent to the existing indent file but have not included
the latest from the vim-lua repository as @marcuscf doesn't consider it to be
well tested enough for inclusion at this time.

This addresses issue #10045.


brammool:
I'll include it.

PR #11117, opened 2022-09-12
Fix 'nosplitscroll' poblems

### LINKED ISSUE #11115: With 'nosplitscroll', windows are sometimes unnecessarily scrolled after :colder or :cnewer is executed
### Steps to reproduce

Run `vim --clean -S test.vim`, with
```vim
set nosplitscroll
edit foo
call setline(1, range(100))
normal! G
copen
close
```
Then run `:colder`, and after the error is displayed, hit `Enter`.

**Result:** The window has been scrolled.

### Expected behaviour

No scrolling after `:colder` (or `:cnewer`) when an error occurs.

### Version of Vim

9.0.447

cc: @luukvbaal

luukvbaal:
Problem:    w_prev_height may be wrong when adding laststatus or winbar.
Solution:   Initialize w_prev_height and update it when setting laststatus or winbar. Also delay initializing firstwin size until after gui has started.

Instead of having a separate case in win_fix_scroll() for when w_prev_height was uninitialized, we now initialize it in win_init_size(). This prevents scrolling for the first split when 'nosplitbelow' is set. Slipped through the cracks as this only happens for the first window in a new tab when the first split is horizontal. Might be considered hair splitting again but I feel this is more correct and not too involved.

~I don't think this addresses https://github.com/vim/vim/issues/11115 yet. Feel free to keep this open and I'll add the fix to that if I find it.~
Added a commit that should fix #11115.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -375,7 +375,6 @@ main
      * Set the default values for the options that use Rows and Columns.
      */
     ui_get_shellsize();		// inits Rows and Columns
-    win_init_size();
 #ifdef FEAT_DIFF
     // Set the 'diff' option now, so that it can be checked for in a .vimrc
     // file.  There is no buffer yet though.
@@ -542,6 +541,7 @@ vim_main2(void)
 	    mch_exit(1);
     }
 #endif
+    win_init_size();
 
 #ifdef FEAT_VIMINFO
     /*
--- a/src/menu.c
+++ b/src/menu.c
@@ -436,6 +436,7 @@ ex_menu(
 		--curwin->w_height;
 	    curwin->w_winbar_height = h;
 	}
+	curwin->w_prev_height = curwin->w_height;
     }
 
 theend:
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1646,6 +1646,7 @@ func Test_splitscroll_with_splits()
           for so in [0, 5]
             for ls in range(0, 2)
               for pos in ["H", "M", "L"]
+              tabnew | tabonly!
               let tabline = (gui ? 0 : (tab ? 1 : 0))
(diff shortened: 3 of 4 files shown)


brammool:
One related test is failing, perhaps it is flaky?
From test_window_cmd.vim:
12249
	Found errors in Test_splitscroll_with_splits():
12250
	command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[469]..function RunTheTest[44]..Test_splitscroll_with_splits line 81: Expected 24 but got 256

luukvbaal:
Yeah I noticed, not sure what could cause it. I have no idea how this test could fail on only a single platform.

luukvbaal:
I've never seen it be flaky while developing the patch. I have had iterations of the patch where it would pass on terminal but fail on gui. It's the first time I've seen it fail on just a single platform so it seems unlikely that it would be flaky... 

On second thought I can imagine this change making it flaky. I suspect it has something to do with the reason for my delaying the win_init_size() call. I guess there is another point where the gui size might change(why?). Perhaps we could flag it flaky for just gui?

luukvbaal:
Or perhaps we just need a `redraw!` somewhere. I must confess I do not quite understand where I need them in that test(and why it can fix it). I did remove some at some point because they it seemed to pass without and it did increase the runtime a lot. But perhaps that's what made it flaky now, at least for gui.

luukvbaal:
Was able to reproduce some flakiness by running the test with gvim in a floating window(as opposed to tiled by my window manager, which is what I had done up till now). Test seems to pass now with some redraws added in the latest commit. It seems disabling gui scrollbars with 'guioptions' for gvim stops the window from resizing unexpectedly. Might be worth setting guioptions if this test causes more problems but it works for now...

Issue #11119, opened 2022-09-13
With 'nosplitscroll', 'scroll' may unexpectedly changed


k-takata:
### Steps to reproduce

1. `vim --clean -c "set nosplitscroll"
2. `:set scroll?`
  The shown value will be nearly half of the 'line' option.
3. `:sp`
4. `:set scroll?`
  `0` is shown. (unexpected)
5. `:help`
6. `:set scroll?`
  `1` is shown. (unexpected)

### Expected behaviour

The value of 'scroll' should be nearly half of the current window height.

### Version of Vim

9.0.0453

### Environment

Windows 10 GUI, Cygwin Terminal

brammool:
@luukvbaal

chrisbra:
apparently fixed by https://github.com/vim/vim/commit/470a14140bc06f1653edf26ab0b3c9b801080353

PR #11122, opened 2022-09-13
More 'splitscroll' problems

### LINKED ISSUE #11123: 9.0.457 is broken
### Steps to reproduce

`vim -u NONE`

### Expected behaviour

Vim should start like usual.

Instead, I see this repeated five times at the bottom of the screen, as vim messages: `E341: Internal error: lalloc(0, )`

### Version of Vim

9.0.457

### Environment

Ubuntu 20.04
terminal= kitty 0.26.2
shell: bash 5.0.17

### Logs and stack traces

_No response_

luukvbaal:
Problem:    'scroll' is not updated when resizing windows, alloc error due to unitialized window size in terminal communication?
Solution:   Update 'scroll' in win_fix_scroll(), revert win_init_size() relocation in main.c and update prev_height for gui instead. (closes https://github.com/vim/vim/issues/11123, https://github.com/vim/vim/issues/11119, https://github.com/vim/vim/issues/11125)

This was being skipped for cases where scroll_to_fraction was not used in win_fix_scroll/cursor. Might be able to refactor this such that is only called when we do not scroll_to_fraction in either win_fix_scroll or win_fix cursor but it seems an inexpensive function so I'm not sure we care enough to? Does this need a test? Seems trivial.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -375,6 +375,7 @@ main
      * Set the default values for the options that use Rows and Columns.
      */
     ui_get_shellsize();		// inits Rows and Columns
+    win_init_size();
 #ifdef FEAT_DIFF
     // Set the 'diff' option now, so that it can be checked for in a .vimrc
     // file.  There is no buffer yet though.
@@ -539,9 +540,9 @@ vim_main2(void)
 	// don't have them.
 	if (!gui.in_use && params.evim_mode)
 	    mch_exit(1);
+	firstwin->w_prev_height = firstwin->w_height; // may have changed
     }
 #endif
-    win_init_size();
 
 #ifdef FEAT_VIMINFO
     /*
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1668,6 +1668,7 @@ func Test_splitscroll_with_splits()
               execute 'norm gg' . pos
               split | redraw | wincmd k
               call assert_equal(1, line("w0"))
+              call assert_equal(&scroll, winheight(0) / 2)
               wincmd j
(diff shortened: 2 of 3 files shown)


brammool:
There are easy repro steps thus it should be easy to add a regression test.

luukvbaal:
> There are easy repro steps thus it should be easy to add a regression test.

Sure, easy enough. I just thought it might have been redundant as it is basically just testing win_comp_scroll which I assume is tested elsewhere. Gave up on the suggested refactor I mentioned in https://github.com/vim/vim/issues/11123#issuecomment-1245710418. Didn't work correctly for win_close control flow.

luukvbaal:
Just to be sure; this should fix both outstanding issues and is ready to be merged as far as I'm concerned.

brammool:
> > There are easy repro steps thus it should be easy to add a regression test.
> 
> Sure, easy enough. I just thought it might have been redundant as it
> is basically just testing win_comp_scroll which I assume is tested
> elsewhere. Gave up on the suggested refactor I mentioned in
> https://github.com/vim/vim/issues/11123#issuecomment-1245710418.
> Didn't work correctly for win_close control flow.

There probably is some testing, but since the combination of features
and user actions is infinite there is never 100% coverage.  If someone
ran into a problem and reported it, it's always good to add a test for
that specific situation, since someone was doing that.  Ideally we might
think "if this fails, then perhaps that fails too?" and add a few
related tests.  But it's quite often already more work to write a test
than to fix the problem, therefore we are flexible.

-- 
Q: Should I clean my house or work on Vim?
A: Whatever contains more bugs.

brammool:
Included with patch 9.0.0461

PR #11127, opened 2022-09-14
fix: `make test_cmdline` did not clean up tmp files Xnamedir/*

dpelle:
`make test_cmdline` did not clean up tmp files Xnamedir/*

### DIFF
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -2149,7 +2149,7 @@ func Test_wildmenu_pum()
   call VerifyScreenDump(buf, 'Test_wildmenu_pum_13', {})
 
   " Directory name completion
-  call mkdir('Xnamedir/XdirA/XdirB', 'p')
+  call mkdir('Xnamedir/XdirA/XdirB', 'pR')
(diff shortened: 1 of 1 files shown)


dpelle:
I see that it's a recurrent issue to have tests that leave temporary files behind.
How about checking that there is no `testdir/X*` files left when all tests have run?
Or preferably checking after running each test.

`make test` could then fail and such issue would be detected earlier.

brammool:
> I see that it's a recurrent issue to have tests that leave temporary
> files behind.  How about checking that there is no `testdir/X*` files
> left when all tests have run?
> Or preferably checking after running each test.
> 
> `make test` could then fail and such issue would be detected earlier.

Well, I don't want to make CI more flaky.  On MS-Windows you can't
delete a file that is still in use, and asynchronous operations can keep
a file open longer than expected.  Would have to set a timer and clean
it up after a few seconds.

We could have a "test thoroughly" make target.

-- 
67. Your hard drive crashes. You haven't logged in for two hours.  You start
    to twitch. You pick up the phone and manually dial your ISP's access
    number. You try to hum to communicate with the modem.  You succeed.

PR #11128, opened 2022-09-14
Don't make cursor moves when closing cmdwin with nosplitscroll

mityu:
When `set nosplitscroll`, sometimes the cursor position of the current window is changed after closing cmdwin.
It shouldn't happen since no cursor moves on the current window happens when leaving command-line.

**Steps to reproduce**
- Open vim with `vim -u NONE --cmd "set nosplitscroll | call setline(1, range(1, &lines)) | normal L"`.
- Type `q:<C-w>q` and check the cursor position.

**Images**
- before
nosplitscroll-before

- after
nosplitscroll-after

### DIFF
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -4405,6 +4405,7 @@ open_cmdwin(void)
     int			save_restart_edit = restart_edit;
     int			save_State = State;
     int			save_exmode = exmode_active;
+    int			save_p_spsc;
 #ifdef FEAT_RIGHTLEFT
     int			save_cmdmsg_rl = cmdmsg_rl;
 #endif
@@ -4643,7 +4644,11 @@ open_cmdwin(void)
 	// First go back to the original window.
 	wp = curwin;
 	set_bufref(&bufref, curbuf);
+
+	save_p_spsc = p_spsc;
+	p_spsc = 1;
 	win_goto(old_curwin);
+	p_spsc = save_p_spsc;
 
 	// win_goto() may trigger an autocommand that already closes the
 	// cmdline window.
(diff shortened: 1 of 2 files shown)


mityu:
Oh sorry, information about my environment is missing.

- OS: macOS Monterey
- Terminal: wezterm 20220905-102802-7d4b8249
- Vim: v9.0.461

luukvbaal:
Hmm scrolling happens when the cursor is at the end of a buffer.

mityu:
@luukvbaal How can I reproduce it?

luukvbaal:
Wow that was a hard find @mityu @brammool. I noticed the issue in some dotfile on my system, but figuring out a repro command proved difficult. The issue somehow only happens when the cursor is on a matched character i.e. (){}[]. AFAIK matchparen is responsible for adding this highlighting so to reproduce it you need matchparen enabled. Seems very unrelated I know, but it seems to be what causes the scrolling...

    vim +"set nospsc nowrap | call setline(1, range(100)) | call setline(100, '()') | norm GL"

[asciicast](https://asciinema.org/a/wsfC9muYhrgvj2bTjpdMspaSI)

mityu:
Reproduced with Vim9.0.465.

- Open Vim with `vim -u NONE --cmd "set nosplitscroll | call setline(1, range(&lines)) | autocmd WinEnter <buffer> call line('w0')"`
- Type `Gq:<C-w>q`

nosplitscroll-cmdwin

mityu:
It doesn't happen with Vim9.0.464, so it's must a new bug caused by this PR.  I'll looking into this later.

luukvbaal:
Yeah I had noticed the cursor changing seemingly unnecessarily while developing the option. Didn't investigate further yet as it seemed complicated. I think I found a fix now for the cursor issue that doesn't mess up the scrolling for autocmd case. I'll open the PR, thanks for pointing me in the right direction @mityu.

PR #11134, opened 2022-09-14
cursor moves when cmdwin is closed when 'splitscroll' is off

luukvbaal:
Problem:    Cursor moves when cmdwin is closed when 'splitscroll' is off.
Solution:   Skip win_fix_cursor if called when cmdwin is open or closing.

Reverts change to ex_getln.c from https://github.com/vim/vim/commit/e697d488901b6321ddaad68b553f0a434c97d849, keep the test. Can you confirm this works @mityu.

### DIFF
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -4405,7 +4405,6 @@ open_cmdwin(void)
     int			save_restart_edit = restart_edit;
     int			save_State = State;
     int			save_exmode = exmode_active;
-    int			save_p_spsc;
 #ifdef FEAT_RIGHTLEFT
     int			save_cmdmsg_rl = cmdmsg_rl;
 #endif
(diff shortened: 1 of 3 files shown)


mityu:
I found one problem with this patch at least.  With `win_execute()` function, you can change cursor positions on other windows while in cmdwin.  However, with this patch, cursor position changes for the current window by user-side scripts will be reverted when leaving cmdwin.
Example:
- `vim -u NONE --cmd "set nosplitscroll | call setline(1, range(&lines))"
- Type `Gq::call win_execute(winnr('#')->win_getid(), 'call cursor(1, 1)')<CR><C-w>q`

Without this patch, the cursor position will be `(1, 1)`, but with this patch, it's not.

- With this patch
with-11134

- Without this patch
without-11134

luukvbaal:
Or set a global of course. I think the latest commit should fix the cursor issue without side-effects.

mityu:
It seems that this patch works well. LGTM.

PR #11139, opened 2022-09-15
'splitscroll' test takes too long

luukvbaal:
Problem:    'splitscroll' test takes too long. Cursor position is not always maintained when not in normal mode.
Solution:   Get rid of nested for loops. Center cursor in window when not in normal mode.

I think this should test most of what we actually wanted to test with the nested loops. Tested option values generated by `call add(v:errors, "so=".run.", tab=".(run % 5).", winbar=".(run % 2).", ls=".(run % 3).", ea=".(run % 2).", sb=".(run % 3).", pos=".pos)`:

```vim
so=0, tab=0, winbar=0, ls=0, ea=0, sb=0, pos=H
so=1, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
so=2, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
so=3, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
so=4, tab=1, winbar=0, ls=1, ea=0, sb=1, pos=L
so=5, tab=0, winbar=1, ls=2, ea=1, sb=1, pos=M
so=6, tab=1, winbar=0, ls=0, ea=0, sb=0, pos=H
so=7, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
so=8, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
so=9, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
so=10, tab=0, winbar=0, ls=1, ea=0, sb=1, pos=L
```

EDIT: investigating failing test.

### DIFF
--- a/runtime/doc/quickref.txt
+++ b/runtime/doc/quickref.txt
@@ -919,7 +919,7 @@ Short explanation of each option:		*option-list*
 'spellsuggest'	  'sps'     method(s) used to suggest spelling corrections
 'splitbelow'	  'sb'	    new window from split is below the current one
 'splitright'	  'spr'     new window is put right of the current one
-'splitscroll'	  'spsc'    determines scroll behavior when splitting windows
+'splitscroll'	  'spsc'    determines scroll behavior for split windows
 'startofline'	  'sol'     commands move cursor to first non-blank in line
 'statusline'	  'stl'     custom format for the status line
 'suffixes'	  'su'	    suffixes that are ignored with multiple match
--- a/src/window.c
+++ b/src/window.c
@@ -6403,7 +6403,6 @@ win_fix_scroll(int resize)
     static void
 win_fix_cursor(int normal)
 {
-    int      top = FALSE;
     win_T    *wp = curwin;
     long     so = get_scrolloff_value();
     linenr_T nlnum = 0;
@@ -6418,7 +6417,7 @@ win_fix_cursor(int normal)
     so = MIN(wp->w_height / 2, so);
     // Check if cursor position is above topline or below botline.
     if (wp->w_cursor.lnum < (wp->w_topline + so) && wp->w_topline != 1)
-	top = nlnum = MIN(wp->w_topline + so, wp->w_buffer->b_ml.ml_line_count);
+	nlnum = MIN(wp->w_topline + so, wp->w_buffer->b_ml.ml_line_count);
     else if (wp->w_cursor.lnum > (wp->w_botline - so - 1)
 	    && (wp->w_botline - wp->w_buffer->b_ml.ml_line_count) != 1)
 	nlnum = MAX(wp->w_botline - so - 1, 1);
@@ -6436,7 +6435,11 @@ win_fix_cursor(int normal)
 	}
 	else
 	{   // Ensure cursor stays visible if we are not in normal mode.
-	    wp->w_fraction = top ? 0 : FRACTION_MULT;
+	    wp->w_fraction = 0.5 * FRACTION_MULT;
+	    // Make sure cursor is closer to topline than botline.
+	    if (so == wp->w_height / 2
+			  && nlnum - wp->w_topline > wp->w_botline - 1 - nlnum)
+		wp->w_fraction++;
 	    scroll_to_fraction(wp, wp->w_prev_height);
 	}
     }
--- a/runtime/optwin.vim
+++ b/runtime/optwin.vim
@@ -515,7 +515,7 @@ call <SID>AddOption("splitbelow", gettext("a new window is put below the current
 call <SID>BinOptionG("sb", &sb)
 call <SID>AddOption("splitright", gettext("a new window is put right of the current one"))
 call <SID>BinOptionG("spr", &spr)
-call <SID>AddOption("splitscroll", gettext("determines scroll behavior when spliting windows"))
+call <SID>AddOption("splitscroll", gettext("determines scroll behavior for split windows"))
 call <SID>BinOptionG("spsc", &spsc)
 call <SID>AddOption("scrollbind", gettext("this window scrolls together with other bound windows"))
 call append("$", "\t" .. s:local_to_window)
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1638,126 +1638,113 @@ func Test_splitscroll_with_splits()
   set nowrap
   set nosplitscroll
   let gui = has("gui_running")
-  inoremap c <cmd>:copen<CR>
-  for winbar in [0, 1]
-    for sb in [0, 1]
-      for ea in [0, 1]
-        for tab in [0, 1]
-          for so in [0, 5]
-            for ls in range(0, 2)
-              for pos in ["H", "M", "L"]
-              tabnew | tabonly! | redraw
-              let tabline = (gui ? 0 : (tab ? 1 : 0))
-              let winbar_sb = (sb ? winbar : 0)
-              execute 'set scrolloff=' . so
-              execute 'set laststatus=' . ls
-              execute 'set ' . (ea ? 'equalalways' : 'noequalalways')
-              execute 'set ' . (sb ? 'splitbelow' : 'nosplitbelow')
-              execute tab ? 'tabnew' : ''
-              execute winbar ? 'nnoremenu 1.10 WinBar.Test :echo' : ''
-              call setline(1, range(1, 256))
-              " No scroll for restore_snapshot
-              norm G
-              try
-                copen | close | colder
-              catch /E380/
-              endtry
-              call assert_equal(257 - winheight(0), line("w0"))
+  inoremap <expr> c "<cmd>copen<bar>wincmd k<CR>"
+  for run in range(0, 10)
+    tabnew | tabonly! | redraw    
+    let tabline = (gui ? 0 : ((run % 5) ? 1 : 0))
+    let winbar_sb = (run % 2) && (run % 3)
+    execute 'set scrolloff=' . !(run % 3) ? 0 : run
+    execute 'set laststatus=' . (run % 3)
+    execute 'set ' . ((run % 2) ? 'equalalways' : 'noequalalways')
+    execute 'set ' . ((run % 3) ? 'splitbelow' : 'nosplitbelow')
+    execute (run % 5) ? 'tabnew' : ''
+    execute (run % 2) ? 'nnoremenu 1.10 WinBar.Test :echo' : ''
+    let pos = !(run % 3) ? 'H' : ((run % 2) ? 'M' : 'L')
+    call setline(1, range(1, 256))
+    " No scroll for restore_snapshot
+    norm G
+    try
+      copen | close | colder
+    catch /E380/
+    endtry
+    call assert_equal(257 - winheight(0), line("w0"))
 
-              " No scroll for firstwin horizontal split
-              execute 'norm gg' . pos
-              split | redraw | wincmd k
-              call assert_equal(1, line("w0"))
-              call assert_equal(&scroll, winheight(0) / 2)
-              wincmd j
-              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
+    " No scroll for firstwin horizontal split
+    execute 'norm gg' . pos
+    split | redraw | wincmd k
+    call assert_equal(1, line("w0"))
+    call assert_equal(&scroll, winheight(0) / 2)
+    wincmd j
+    call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
 
-              " No scroll when resizing windows
-              wincmd k | resize +2
-              call assert_equal(1, line("w0"))
-              wincmd j
-              call assert_equal(win_screenpos(0)[0] - tabline - winbar_sb, line("w0"))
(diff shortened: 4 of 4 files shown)


mityu:
I don't know why but it seems that test fails when window height is short.
At least in my environment, test fails when window height <= 30.

Here's the test results in my environment:

|Terminal width x height|Result|
|:--:|:--:|
|80x24|failed|
|80x30|failed|
|80x31|passed|
|80x35|passed|
|90x30|failed|
|90x31|passed|

Environment:
- OS: macOS Monterey
- Terminal: iTerm2 3.4.16

brammool:
> Problem:    'splitscroll' test takes too long.
> Solution:   Get rid of nested for loops.
> 
> I think this should test most of what we actually wanted to test with
> the nested loops. Tested option values generated by `call
> add(v:errors, "so=".run.", tab=".(run % 5).", winbar=".(run % 2).",
> ls=".(run % 3).", ea=".(run % 2).", sb=".(run % 3).", pos=".pos)`:
> 
> ```vim
> so=0, tab=0, winbar=0, ls=0, ea=0, sb=0, pos=H
> so=1, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
> so=2, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
> so=3, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
> so=4, tab=1, winbar=0, ls=1, ea=0, sb=1, pos=L
> so=5, tab=0, winbar=1, ls=2, ea=1, sb=1, pos=M
> so=6, tab=1, winbar=0, ls=0, ea=0, sb=0, pos=H
> so=7, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
> so=8, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
> so=9, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
> so=10, tab=0, winbar=0, ls=1, ea=0, sb=1, pos=L
> ```

This should be much better.  I tried not using "winbar" and "tab", but
it didn't make the resizing go away.

Most of the values are binary, but 'scrolloff' is a value.  Now there is
only one test with 'scrolloff' set to zero.  I would suggest to do about
half the tests with 'scrolloff' zero.  Perhaps something like:

	so = run % 2 == 0 ? 0 : run / 2

-- 
86. E-mail Deficiency Depression (EDD) forces you to e-mail yourself.

luukvbaal:
> I don't know why but it seems that test fails when window height is short. At least in my environment, test fails when window height <= 30.

Yeah it has to do with the 'scrolloff' insert mode case. I think I have found the fix.

> Most of the values are binary, but 'scrolloff' is a value.  Now there is
only one test with 'scrolloff' set to zero.

Yeah you're right. Although 'scrolloff' is the more complicated case that could do with more testing, only doing a single run with 'so==0' might result in missing things.

brammool:
> > I don't know why but it seems that test fails when window height is
> > short. At least in my environment, test fails when window height <=
> > 30.
> 
> Yeah it has to do with the 'scrolloff' insert mode case. I think I
> have found the fix.
> 
> > Most of the values are binary, but 'scrolloff' is a value.  Now there is
> > only one test with 'scrolloff' set to zero.
> 
> Yeah you're right. Although 'scrolloff' is the more complicated case
> that could do with more testing, only doing a single run with 'so==0'
> might result in missing things.

BTW: I can avoid the window resizing by using:

	" disallow window resizing
  	let save_WS = &t_WS
  	set t_WS=

And later restoring the option:

	let &t_WS = save_WS

I suppose the place where you tried before had t_WS empty by default.

-- 
Ed's Radiator Shop: The Best Place in Town to Take a Leak.

luukvbaal:
> Yeah it has to do with the 'scrolloff' insert mode case. I think I have found the fix.

To reiterate: if the windows are resized while we are not in normal mode (e.g. by autocommands/plugins), we want to make sure the cursor position stays valid, rather than changing it. Currently we were doing scroll_to_fraction with w_fraction set to 0 or 1, depending if the cursor went out of view above or below the window, to scroll the minimal amount. This does not work properly when w_height < 'scrolloff'. The fix I propose is just scrolling the cursor to the middle of the window instead. This works for the w_height < 'scrolloff' case and does not seem too offensive otherwise, even assuming 'scrolloff == 0'. Since scrolling will happen anyways we might as well place the cursor in the middle of the screen if that is where you were typing/selecting.

luukvbaal:
Latest commit tests these values:

```
so=0, tab=0, winbar=0, ls=0, ea=0, sb=0, pos=H
so=1, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
so=2, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
so=0, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
so=4, tab=1, winbar=0, ls=1, ea=0, sb=1, pos=L
so=5, tab=0, winbar=1, ls=2, ea=1, sb=1, pos=M
so=0, tab=1, winbar=0, ls=0, ea=0, sb=0, pos=H
so=7, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
so=8, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
so=0, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
so=10, tab=0, winbar=0, ls=1, ea=0, sb=1, pos=L
```

Increasing `run` range as far as 100 still passes the test so that makes me confident that 'scrolloff' is now handled correctly.

> BTW: I can avoid the window resizing by using:

	" disallow window resizing
  	let save_WS = &t_WS
  	set t_WS=

Should I add it?

src/window.c
```diff
@@ -6436,7 +6435,11 @@ win_fix_cursor(int normal)
 	}
 	else
 	{   // Ensure cursor stays visible if we are not in normal mode.
-	    wp->w_fraction = top ? 0 : FRACTION_MULT;
+	    wp->w_fraction = 0.5 * FRACTION_MULT;
+	    // Make sure cursor is closer to topline than botline.
+	    if (so == wp->w_height / 2
+			  && nlnum - wp->w_topline > wp->w_botline - 1 - nlnum)
+		wp->w_fraction += 1L;
```
luukvbaal: I think this will be covered by the test if we increase the loop range. Should we? Not sure how to determine what range will cover it. I tested it on my local system and confirmed this fixes a problem but when this happens depends on the window height the test runs in.

brammool:
> Latest commit tests these values:
> 
> ```
> so=0, tab=0, winbar=0, ls=0, ea=0, sb=0, pos=H
> so=1, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
> so=2, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
> so=0, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
> so=4, tab=1, winbar=0, ls=1, ea=0, sb=1, pos=L
> so=5, tab=0, winbar=1, ls=2, ea=1, sb=1, pos=M
> so=0, tab=1, winbar=0, ls=0, ea=0, sb=0, pos=H
> so=7, tab=1, winbar=1, ls=1, ea=1, sb=1, pos=M
> so=8, tab=1, winbar=0, ls=2, ea=0, sb=1, pos=L
> so=0, tab=1, winbar=1, ls=0, ea=1, sb=0, pos=H
> so=10, tab=0, winbar=0, ls=1, ea=0, sb=1, pos=L
> ```
> 
> Increasing `run` range as far as 100 still passes the test so that
> makes me confident that 'scrolloff' is now handled correctly.

OK.  Some combinations are not tested though, e.g. so=0 with ls=1.

> > BTW: I can avoid the window resizing by using:
> 
> 	" disallow window resizing
>   	let save_WS = &t_WS
>   	set t_WS=
> 
> Should I add it?

Did you manage the reproduce the window resizing?  I suppose it would
still happen with a reduced number of loops.  Might as well do that.

-- 
90. Instead of calling you to dinner, your spouse sends e-mail.

brammool:
> @luukvbaal commented on this pull request.
> 
> 
> 
> > @@ -6436,7 +6435,11 @@ win_fix_cursor(int normal)
>  	}
>  	else
>  	{   // Ensure cursor stays visible if we are not in normal mode.
> -	    wp->w_fraction = top ? 0 : FRACTION_MULT;
> +	    wp->w_fraction = 0.5 * FRACTION_MULT;
> +	    // Make sure cursor is closer to topline than botline.
> +	    if (so == wp->w_height / 2
> +			  && nlnum - wp->w_topline > wp->w_botline - 1 - nlnum)
> +		wp->w_fraction += 1L;
> 
> I think this will be covered by the test if we increase the loop
> range. Should we? Not sure how to determine what range will cover it.
> I tested it on my local system and confirmed this fixes a problem but
> when this happens depends on the window height the test runs in.

We actually still have FEAT_FLOAT around float computations.
It probably isn't needed in most places, it actually means that we have
all the float functions, such as atan() and strod().

I think it's OK to use 0.5.  If nothing breaks on any system then I'll
have a look into changing FEAT_FLOAT into the actually intended
HAVE_FLOAT_FUNCS.  Or perhaps FEAT_EVAL_FLOAT.  Or check each function
separately and only disable a tiny part that is using one.

-- 
91. It's Saturday afternoon in the middle of May and you
    are on computer.

luukvbaal:
> OK.  Some combinations are not tested though, e.g. so=0 with ls=1.

Yes, it is rather arbitrary. Increasing the loop range will increase the coverage. Since the runtime is significantly smaller now, I guess we could afford to. But it's more about testing different values. I don't think each combination of options leads to different control flow in the code. If that were the case, we would need the nested loops.

> Did you manage the reproduce the window resizing?  I suppose it would
still happen with a reduced number of loops.  Might as well do that.

I didn't try any further but no I couldn't reproduce it on my current system. I've seen it happening when I was testing gvim though. Feel free to add it, and/or increase the loop range when you include this patch, IDK if it matters.

> We actually still have FEAT_FLOAT around float computations.
It probably isn't needed in most places, it actually means that we have
all the float functions, such as atan() and strod().

> I think it's OK to use 0.5.  If nothing breaks on any system then I'll
have a look into changing FEAT_FLOAT into the actually intended
HAVE_FLOAT_FUNCS.  Or perhaps FEAT_EVAL_FLOAT.  Or check each function
separately and only disable a tiny part that is using one.

TBH I added this because I was porting it to neovim and their CI was complaining about it. Do as you see fit.

brammool:
> > OK.  Some combinations are not tested though, e.g. so=0 with ls=1.
> 
> Yes, it is rather arbitrary. Increasing the loop range will increase
> the coverage. Since the runtime is significantly smaller now, I guess
> we could afford to. But it's more about testing different values. I
> don't think each combination of options leads to different control
> flow in the code. If that were the case, we would need the nested
> loops.

OK.  If any problems get reported we can add a specific regression test,
that usually works better than random conbinations.  Since it's a setup
that someone actually uses vs. a synthetic one.

> > Did you manage the reproduce the window resizing?  I suppose it would
> > still happen with a reduced number of loops.  Might as well do that.
> 
> I didn't try any further but no I couldn't reproduce it on my current
> system. I've seen it happening when I was testing gvim though. Feel
> free to add it, and/or increase the loop range when you include this
> patch, IDK if it matters.

I do still see it, so let me add that.  Might actually use this in more
tests, it should improve speed and reduce flakiness.

> > We actually still have FEAT_FLOAT around float computations.
> > It probably isn't needed in most places, it actually means that we have
> > all the float functions, such as atan() and strod().
> 
> > I think it's OK to use 0.5.  If nothing breaks on any system then I'll
> > have a look into changing FEAT_FLOAT into the actually intended
> > HAVE_FLOAT_FUNCS.  Or perhaps FEAT_EVAL_FLOAT.  Or check each function
> > separately and only disable a tiny part that is using one.
> 
> TBH I added this because I was porting it to neovim and their CI was
> complaining about it. Do as you see fit.

I asked about it on the vim-dev maillist, no comments so far...

-- 
From "know your smileys":
 ~#:-(	I just washed my hair, and I can't do nuthin' with it.

runtime/doc/quickref.txt
```diff
@@ -919,7 +919,7 @@ Short explanation of each option:		*option-list*
 'spellsuggest'	  'sps'     method(s) used to suggest spelling corrections
 'splitbelow'	  'sb'	    new window from split is below the current one
 'splitright'	  'spr'     new window is put right of the current one
-'splitscroll'	  'spsc'    determines scroll behavior when splitting windows
+'splitscroll'	  'spsc'    determines scroll behavior for split windows
```
luukvbaal, line 922: This change was missed in https://github.com/vim/vim/commit/594f9e09cd68e6277b8aa08094405bc642c5792a.

brammool:
> @luukvbaal commented on this pull request.
> 
> > -'splitscroll'	  'spsc'    determines scroll behavior when splitting windows
> +'splitscroll'	  'spsc'    determines scroll behavior for split windows
> 
> This change was missed in https://github.com/vim/vim/commit/594f9e09cd68e6277b8aa08094405bc642c5792a.

Quite often runtime file updates are bundled up.  Only when introducing
something new or changing syntax it gets included with the patch.

-- 
From "know your smileys":
 :q	vi user saying, "How do I get out of this damn emacs editor?"

PR #11142, opened 2022-09-16
Fix substitute() doesn't pass submatches to varargs function

zeertzjq:


brammool:
The test doesn't fail without the fix.  Adding an argument to the function works then.
This also reveals that the "==" isn't right, it should be "<=".

PR #11150, opened 2022-09-17
'nosplitscroll' problems

### LINKED ISSUE #11145: With 'nosplitscroll' the current window is unexpectedly scrolled when 'syntax' is set in a popup window
### Steps to reproduce

1. Run`vim --clean -S test.vim`, with
```vim
vim9script

set nosplitscroll

setline(1, range(100)->mapnew((_, i: number): string => i .. repeat(' test', 10)))
normal! G
normal! zz
normal! $

def Test()
    const winid: number = popup_atcursor('// test test test', {})
    setbufvar(winbufnr(winid), '&syntax', 'c')
enddef

nnoremap g<cr> <scriptcmd>Test()<cr>
```

2. While the cursor is located on the last line and centered on the screen, press `g<cr>`.
3. Result: The window scrolls unexpectedly. In some situations the cursor jumps to the beginning of a line (which I haven't been able to reproduce with a minimal example yet).

The scrolling is definitively caused when setting the syntax for the buffer displayed in the popup window:
```vim
setbufvar(winbufnr(winid), '&syntax', 'c')
```

No issues arise when `splitscroll` is enabled.

### Expected behaviour

No scrolling and no cursor jumps when the syntax is set for a buffer displayed in a popup window.

### Version of Vim

9.0.478

### LINKED ISSUE #11147: With 'nosplitscroll' and 'splitbelow' a window is sometimes unexpectedly scrolled when Vim's help window is closed
### Steps to reproduce

1. First set the terminal window to approximately 40 lines and 80 columns. Check with `$LINES` and `$COLUMNS`.
2. Then run `vim --clean test.vim -S test.vim`, with

```vim
set nosplitscroll
set splitbelow

" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines
" empty lines

" augroup test
"     autocmd!
" augroup END
```
3. Move the cursor to the very last line and place it on `augroup`. Then press `K` to open the help for `augroup`. When opening the split, the previously focused window is NOT scrolled. Immediately close the help window with `:q`.
4. Result: The previously focused window that displays `test.vim` has been scrolled.

For some reasons this depends on the size of the terminal window as well as on the location of the cursor in the current window before the split occurs. When opening a new split, the last window is never scrolled but when closing the new split the last window is often scrolled.

### Expected behaviour

No scrolling when Vim's help window is closed.

### Version of Vim

9.0.478

### Environment

Linux, alacritty terminal

### Logs and stack traces

_No response_

luukvbaal:
Problem:	Text is scrolled when aucmd_win is opened and when help win is closed when buffer has less lines than window height. 
Solution:	Skip win_fix_scroll when next_curwin is aucmd_win and do not skip when buffer has less lines than window height.

The aucmd_win skip also prevents https://github.com/vim/vim/issues/11115, so we remove the earlier fix. The  w_buffer->b_ml.ml_line_count < w_height skip was just there as a performance consideration. Apparently we cannot always skip in this case to prevent scrolling though, so remove it. Closes https://github.com/vim/vim/issues/11147, closes https://github.com/vim/vim/issues/11145.

### DIFF
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1789,4 +1789,34 @@ function Test_nosplitscroll_cmdwin_cursor_position()
   %bwipeout!
   set splitscroll&
 endfunction
+
+" No scroll when aucmd_win is opened.
+function Test_nosplitscroll_aucmdwin()
+  set nosplitscroll
+
+  call setline(1, range(1, &lines))
+  norm Gzz
+  let top = line('w0')
+  call setbufvar(bufnr("test", 1) , '&buftype', 'nofile')
+  call assert_equal(top, line('w0'))
+
+  %bwipeout!
+  set splitscroll&
+endfunc
+
+" No scroll when help is closed and buffer line count < window height.
+function Test_nosplitscroll_helpwin()
+  set nosplitscroll
+  set splitbelow
+
+  call setline(1, range(&lines - 10))
+  norm G
+  let top = line('w0')
+  help | quit
+  call assert_equal(top, line('w0'))
+
+  set splitbelow&
+  set splitscroll&
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
--- a/src/window.c
+++ b/src/window.c
@@ -1325,7 +1325,7 @@ win_split_ins(
 	win_equal(wp, TRUE,
 		(flags & WSP_VERT) ? (dir == 'v' ? 'b' : 'h')
 		: dir == 'h' ? 'b' : 'v');
-    else if (!p_spsc)
+    else if (!p_spsc && wp != aucmd_win)
 	win_fix_scroll(FALSE);
 
     // Don't change the window height/width to 'winheight' / 'winwidth' if a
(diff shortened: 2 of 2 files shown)


src/testdir/test_window_cmd.vim
```diff
@@ -1789,4 +1789,35 @@ function Test_nosplitscroll_cmdwin_cursor_position()
...
+  call setline(1, range(1, &lines))
+  norm Gzz
+  let top = line('w0')
+  let id = popup_atcursor('test', {})
+  call setbufvar(winbufnr(id), '&syntax', 'c')
+  call assert_equal(top, line('w0'))
+  call popup_close(id)
+
+  set splitscroll&
+endfunc
```
luukvbaal: @brammool if you could wait with this patch I will try to find a test that also works for neovim.

luukvbaal: Ready now.

PR #11153, opened 2022-09-18
update colosrchemes

habamax:
- pablo: fix 8c invisible comments
- quiet: fix EndOfBuffer and Directory background colors (set to NONE)
- lunaperche: add syntax overrides for cs, java, rst; make comments a stronger blue, make light close to dark wrt using colors (strings are yellowish for both now).


brammool:
I'll include this, thanks.

PR #11167, opened 2022-09-19
fix “patter” typo in Cscope code

Smattr:


brammool:
Thanks, I'll include it, will be in a collection patch later.

PR #11169, opened 2022-09-19
Add HEEx ftplugin

mhanberg:
Sets basic stuff like the comment string and the tab settings.

### DIFF
--- a/runtime/ftplugin/heex.vim
+++ b/runtime/ftplugin/heex.vim
@@ -1,0 +1,20 @@
+" Elixir filetype plugin
+" Language: HEEx
+" Maintainer:	Mitchell Hanberg <vimNOSPAM@mitchellhanberg.com>
+" Last Change: 2022 September 19
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+setlocal shiftwidth=2 softtabstop=2 expandtab
+
+setlocal comments=:<%#
+setlocal commentstring=<%#\ %s\ %>
+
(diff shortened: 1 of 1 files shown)


brammool:
i'll include it, thanks.  Setting and restoring 'cpo' is not really needed in this case.

habamax (commented):
`b:undo_ftplugin` is missing

mhanberg:
@brammool really sorry about this, but I made a couple of mistakes here.

`commentstring` should be `set commentstring=<%!--\ %s\ --%>`
`comments` should be `set comments=:<%!--` (I'm not actually sure if that `:` needs to be there for `comments`).

This comment syntax was added in Elixir 1.14, which was released very recently, and should be used moving forward.

Thanks, and sorry about that.

brammool:
> @brammool really sorry about this, but I made a couple of mistakes here.
> 
> `commentstring` should be `set commentstring=<%!--\ %s\ --%>`
> `comments` should be `set comments=:<%!--` (I'm not actually sure if that `:` needs to be there for `comments`).
> 
> This comment syntax was added in Elixir 1.14, which was released very
> recently, and should be used moving forward.
> 
> Thanks, and sorry about that.

Thus the comment style mentioned here is no longer used anywhere:
```
	<%# This is a comment. %>
```

Seems like a drastic change.

Adding the undo variable, I end up with this:

```

setlocal shiftwidth=2 softtabstop=2 expandtab

setlocal comments=:<%!--
setlocal commentstring=<%!--\ %s\ --%>

let b:undo_ftplugin = 'set sw< sts< et< com< cms<'
```

-- 
119. You are reading a book and look for the scroll bar to get to
     the next page.

PR #11170, opened 2022-09-19
Only postpone :redrawstatus if messages have scrolled

zeertzjq:
After patch 9.0.0507 it is hard to redraw statusline in cmdline mode.
I think it is better to keep previous behavior if message have not
scrolled, as it doesn't seems to cause problems in that case.


Shane-XB-Qian:
image

1, vim --clean /etc/apt/source.lst  (or whatever a system readonly fs)
2, then keep pressing 'u'

not sure if same situation, but seems it happened (behavior changed) recently..
// in the past, looks it never req 'enter', but now it happened to have chance like that..

zeertzjq:
If the warning comes from a script, the hit-enter prompt allows you to see which script the warning comes from.

Shane-XB-Qian:
no, it (the w10 warning msg) came from vim itself
you could try old version vim e.g v8.2.xxxx

zeertzjq:
I know what the problem is, and I already have a fix. I will open a PR after I write a test for it.

zeertzjq:
> I know what the problem is, and I already have a fix. I will open a PR after I write a test for it.

#11178

PR #11178, opened 2022-09-20
Do not scroll for a warning if there is no stacktrace

zeertzjq:
- Don't scroll when no stacktrace is available
- Do not scroll for a warning if there is no stacktrace


zeertzjq:
Accidentally create two commits.

PR #11185, opened 2022-09-21
changing curwin in callback results in scrolling for 'nosplitscroll'

luukvbaal:
changing curwin in callback results in scrolling for 'nosplitscroll'

Problem: w_valid flags are not updated accordingly for 'nosplitscroll', missing test for terminal mode callback.
Solution: Let validate_cursor handle w_valid flags except for VALID_TOPLINE, add callback tests.

This fixes scrolling when the current window is changed inside a callback as I alluded to in: https://github.com/vim/vim/pull/11166
> Moreover, if we move the wincmd k to the callback function, the current fix is not enough to catch that and scrolling will still happen. Trying to fix it with a watchpoint on firstwin->w_topline and backtracing led me to trying to add workarounds in redraw_after_callback() or ex_docmd.c but it got quite complicated. Not sure if it's worth to catch this issue.

If we now catch all possible cases that could result in scrolling due to a user script callback I'm fine with this patch. I however can't be 100% certain that that is the case. I did try various other split/window commands inside the callback script and the added test keeps passing. I'm sure some user will report this issue at some point so up to you if you think this is a valid patch @brammool. Also adds missing test for the terminal mode callback patch added previously.


src/testdir/test_window_cmd.vim
```diff
@@ -1817,4 +1817,35 @@ function Test_nosplitscroll_misc()
   set splitscroll&
 endfunc
 
+source screendump.vim
```
zeertzjq: This is usually put at the top at the file.

brammool:
> Problem:	Changing curwin in callback results in scrolling for 'nosplitscroll', missing test for terminal mode callback.
> Solution:	Prevent scrolling in callback, add missing test.
> 
> This fixes scrolling when the current window is changed inside a callback as I alluded to in: https://github.com/vim/vim/pull/11166
> > Moreover, if we move the wincmd k to the callback function, the
> > current fix is not enough to catch that and scrolling will still
> > happen. Trying to fix it with a watchpoint on firstwin->w_topline
> > and backtracing led me to trying to add workarounds in
> > redraw_after_callback() or ex_docmd.c but it got quite complicated.
> > Not sure if it's worth to catch this issue.
> 
> If we now catch all possible cases that could result in scrolling due
> to a user script callback I'm fine with this patch. I however can´t
> been 100% certain that that is the case. I did try various other
> split/window commands inside the callback script and the added test
> keeps passing. I'm sure some user will report this issue at some point
> so up to you if you think this is a valid patch @brammool. Also adds
> missing test for the terminal mode callback patch added previously.

This calls validate_botline_win() after every callback.  That is an
expensive operation, it has to compute the size of every visible line.

Is there a way to add a condition, that something changed in the window
layout, cursor position, etc?  Actually, the current mechanism to only
invalidate botline (w_valid flags VALID_BOTLINE and VALID_BOTLINE_AP)
should already take care of this.

If you leave out the "invalidate_botline_win()" call, then what fails?
Perhaps the solution is to call invalidate_botline() in some situation
and there check for 'splitscroll' to be false.

-- 
"Computers in the future may weigh no more than 1.5 tons."
                                   Popular Mechanics, 1949

luukvbaal:
> Is there a way to add a condition, that something changed in the window layout, cursor position, etc? Actually, the current mechanism to only invalidate botline (w_valid flags VALID_BOTLINE and VALID_BOTLINE_AP) should already take care of this.

> If you leave out the "invalidate_botline_win()" call, then what fails? Perhaps the solution is to call invalidate_botline() in some situation and there check for 'splitscroll' to be false.

It seems botline is not (in)validated before the window is entered, inside the callback. Thus, leaving it out results in win_fix_cursor() using an outdated value to determine whether topline < cursor.lnum < botline. I'm not sure if this outdated botline could be considered a bug or if it's because vim (without 'nosplitscroll') doesn't care about botline being up-to-date during a callback?
I can look into it once more.

src/window.c
```diff
@@ -6383,6 +6383,7 @@ win_fix_scroll(int resize)
 		wp->w_cursor.lnum = lnum;
 	    }
 	    invalidate_botline_win(wp);
+	    wp->w_valid &= ~VALID_CROW;
```
luukvbaal: Simply invalidating VALID_CROW fixes the issue. But since invalidating it is only necessary for this callback case, I wonder if there is a better place to invalidate it without needlessly making comp_botline() more expensive.

luukvbaal: At the very least we could guard it behind get_callback_depth(), but this still indicates to me that we need to invalidate it somewhere else.

luukvbaal:
> Just above the cursor is moved, thus removing VALID_CROW makes a lot of
sense.

Here we are actually just, changing it temporarily(to scroll to bottom) and restoring it. So invalidating it here still doesn't quite make sense I think.

> This is all getting quite complicated, we should really try to use the
"w_valid" flags in a proper way, otherwise we just keep updating values
that don't need updating.

Yes, I'm not sure why the check_cursor_moved() call in comp_botline() does not invalidate it. I guess it has to do with the fact that we delay moving the cursor in a window until it is entered(such that the cursor is not needlessly moved if the window layout changes again before it is entered). I'll try to make sense of the W_VALID flags and see if I can figure out where we need to set which in win_fix_cursor/scroll.

src/window.c
```diff
@@ -6382,6 +6382,8 @@ win_fix_scroll(int resize)
 		scroll_to_fraction(wp, wp->w_prev_height);
 		wp->w_cursor.lnum = lnum;
 	    }
+	    if (wp == curwin)
+		wp->w_valid &= ~VALID_CROW;
 	    invalidate_botline_win(wp);
```
luukvbaal, line 6383: I couldn't figure out a way to re-use validate_cursor()/check_cursor_moved(). This is the call we are skipping in
win_new_height() that is responsible for invalidating w_crow but using that unconditionally in win_fix_scroll/cursor results in scrolling. Invalidating it only seems to be necessary for curwin, and not invalidating it seems to cause an issue only in this callback case. Not sure if it makes sense to also guard behind get_callback_depth().

luukvbaal:
Latest commit lets the validate_cursor() call in win_new_height() handle the w_valid flags, except for VALID_TOPLINE because we update the cursor position in win_fix_cursor instead. I think this patch now makes sense and it prevents the scrolling that happened in the test case.

luukvbaal:
Scratch that, I thought I was doing a bitwise OR. Looking into it once more...

luukvbaal:
This is the best I could come up with. It fixes the test case and I don't see any issues but I can't say I'm confident that are updated properly for 'nosplitscroll'.

brammool:
> This is the best I could come up with. It fixes the test case and I
> don't see any issues but I can't say I'm confident that are updated
> properly for 'nosplitscroll'.

I'm glad to see you found a simple solution.

-- 
138. You develop a liking for cold coffee.

PR #11189, opened 2022-09-21
CI: macOS m1(arm64) build

pheiduck:
build and tests works as expected.


pheiduck:
@brammool @k-takata please review my CI Implementation.
Thank you.

brammool:
The last run failed.  The failure in the channel test is probably just flaky.
But the failure in Test_quit_long_message() is uncommon. any idea?

brammool:
Actually, I can see a problem with Test_quit_long_message(), it waits for the ruler to show up but there isn't one.
I'll make a patch for that, please trigger CI when merging it.

pheiduck:
@brammool Test_terminal_cwd() failed now.

brammool:
I guess the path is too long to fit in a line, so we get only "tes" from "testdir/Xtermdir"
I made a patch for that.
Test_quit_long_message() also is still failing.  It looks like there is an error from the Vim command.

pheiduck:
@brammool with can use it with  `./configure --with-features=tiny`

brammool:
The failing message test shows this in the screendump:
```
Error detected while processing command line..script /private/var/folders/b
f/0cr6_1t525j9r585tdpgnv_m0000gn/T/cirrus-ci-build/src/testdir/Xtest_quit_m
essage:                                                                    
line    1: 
```
No idea what would be wrong with that line.  Can you locally increase the size of the terminal used from "rows: 6" to "rows: 10", so we can see the error message?

pheiduck:
> The failing message test shows this in the screendump:
> 
> ```
> Error detected while processing command line..script /private/var/folders/b
> f/0cr6_1t525j9r585tdpgnv_m0000gn/T/cirrus-ci-build/src/testdir/Xtest_quit_m
> essage:                                                                    
> line    1: 
> ```
> 
> No idea what would be wrong with that line. Can you locally increase the size of the terminal used from "rows: 6" to "rows: 10", so we can see the error message?

I think this is not possible on the CI.

brammool:
For changing the "rows" you need to edit src/testdir/test_messages.vim.  Just to see the error, won't include it.

brammool:
Cirrus CI reports a problem: Error while parsing tasks: INVALID_ARGUMENT: yaml: line 44: mapping values are not allowed in this context!

pheiduck:
I saw that, will close this PR, its not working

brammool:
> > @brammool we can use it with `./configure --with-features=tiny` or
> > with other configure opts and fix the tests
> 
> As I mentioned before we can use tiny conf opt

Most people will use Vim with huge features, compiling with tiny may
very well hide actual problems.

It may take some effort to get all tests to pass, but it's worth it.
If it really doesn't work we could skip individual tests.
Is there a reliable way we can recognize the system?

-- 
"The sun oozed over the horizon, shoved aside darkness, crept along the
greensward, and, with sickly fingers, pushed through the castle window,
revealing the pillaged princess, hand at throat, crown asunder, gaping
in frenzied horror at the sated, sodden amphibian lying beside her,
disbelieving the magnitude of the frog's deception, screaming madly,
"You lied!"
    - Winner of the Bulwer-Lytton contest (San Jose State University),
      wherein one writes only the first line of a bad novel

pheiduck:
> Is there a reliable way we can recognize the system?

not sure, found it here: https://github.com/cirruslabs/tart

.cirrus.yml
```diff
@@ -26,3 +26,21 @@ freebsd_task:
...
+  timeout_in: 20m
+  install_script:
+    - brew update
+    - brew install gettext libtool
+  build_script:
+    - NPROC=$(getconf _NPROCESSORS_ONLN)
+    - ./configure --with-features=${FEATURES}
+    - make -j${NPROC}
+  test_script:
+    - sudo sed -z 's/let buf = RunVimInTerminal('', {'rows': 6})/let buf = RunVimInTerminal('', {'rows': 10})/g' src/testdir/test_messages.vim
```
fkorotkov, line 44: I think our YAML parser is not happy here because of the `:`. You should probably extract it in a separate script with `"`:

```yaml
patch_script: "sudo sed -z 's/let buf = RunVimInTerminal('', {'rows': 6})/let buf = RunVimInTerminal('', {'rows': 10})/g' src/testdir/test_messages.vim"
```

pheiduck: Thanks for hint :) https://github.com/vim/vim/pull/11226

PR #11191, opened 2022-09-22
win32: Update MSVC makefile

k-takata:
Remove old stuffs for older versions of MSVC.


k-takata:
Added another commit.
Recent MSVC doesn't generate *.exe.manifest anymore. So, no need to use mt.exe.

PR #11192, opened 2022-09-22
No need to use `errbuf` for option argument

zeertzjq:
When out-of-memory causes an early return in do_set_string(), `*arg` may
be set to a local pointer to `whichwrap`, which becomes invalid after
returning. Although `arg` will no longer be used in do_set() if
do_set_string() returns `FAIL`, it is still better to avoid an invalid
pointer.

It should also be unnecessary to use `errbuf` for the argument.

Add a test for :set global 'keywordprg' to empty.

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -1323,13 +1323,11 @@ do_set_string(
 	/*
 	 * Set 'keywordprg' to ":help" if an empty
 	 * value was passed to :set by the user.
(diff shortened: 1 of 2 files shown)


zeertzjq:
The first part of this PR has been superseded by patch 9.0.0540. There are still some applicable.

PR #11194, opened 2022-09-22
Give a link to ex-cmd-index from documentation of :g

alkino:
To give an idea to people of how they can use :global


brammool:
I can include the tag, although I would think most readers will not need it.  "non-Ex command" is a bit confusing, lets make that "Normal mode command".

PR #11200, opened 2022-09-22
Add hare.vim to runtime files

rsaihe:
This pull request adds support for the Hare programming language.


brammool:
I'll include it, thanks.

brammool:
Oh, the indent file uses line continuation in one place.  I just joined the lines, an alternative would be to set and restore 'cpo'.

PR #11204, opened 2022-09-23
Add syntax highlighting for HLS playlists

benoitryder:
This PR adds syntax highlighting for HTTP Live Streaming playlists (.m3u8):
https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-11

Feedbacks are welcome, especially regarding good practices for syntax plugins.


brammool:
> This PR adds syntax highlighting for HTTP Live Streaming playlists (.m3u8):

Can we use a nicer name for this filetype?  Perhaps hlsplaylist?

Please also add an entry in the test src/testdir/test_filetype.vim


-- 
141. You'd rather go to http://www.weather.com/ than look out your window.

benoitryder:
> Can we use a nicer name for this filetype?  Perhaps hlsplaylist?

hlsplaylist is good. I changed the name, and added the test as requested.

brammool:
The syntax file is using line continuation.  the 'cpo' option needs to be set and restored.  Se other syntax files for an example.

benoitryder:
Thanks for noticing. I fixed it on my branch but since the the PR was already closed, the diff does not include it.
Updated file available [here](https://raw.githubusercontent.com/benoitryder/vim/master/runtime/syntax/hlsplaylist.vim).

ObserverOfTime:
The `.m3u` extension is also valid. Also, I think `mpegurl` would be a better name.

https://datatracker.ietf.org/doc/html/draft-pantos-http-live-streaming-17.html#section-4

>   Each Playlist file MUST be identifiable by either the path component
>   of its URI or by HTTP Content-Type.  In the first case, the path MUST
>   end with either .m3u8 or .m3u.  In the second, the HTTP Content-type
>   MUST be "application/vnd.apple.mpegurl" or "audio/mpegurl".

benoitryder:
HLS playlists are based on the old M3U format, but they define their own format. It is not a superset of the M3U format with its common extensions, and it has not the same purpose. The HLS playlist mime-type is `mpegurl` most likely for historical reasons. But this name is never used when working with HLS. Using it has filetype can be confusing for users.

I've never seen the `.m3u` extension being used for HLS playlists. IMHO it's fine to not support it (and maybe reserve it for the old M3U format if support is added in the future), but adding it would not hurt.

brammool:
> HLS playlists are based on the old M3U format, but they define their
> own format. It is not a superset of the M3U format with its common
> extensions, and it has not the same purpose. The HLS playlist
> mime-type is `mpegurl` most likely for historical reasons. But this
> name is never used when working with HLS. Using it has filetype can be
> confusing for users.
> 
> I've never seen the `.m3u` extension being used for HLS playlists.
> IMHO it's fine to not support it (and maybe reserve it for the old M3U
> format if support is added in the future), but adding it would not
> hurt.

Are the .m3u8 and .m3u files different enough to justify a different
filetype?  Would they need different settings, indenting and syntax?

-- 
Send $25.00 for handy leaflet on how to make money by selling leaflets

ObserverOfTime:
M3U8 is just M3U with UTF-8 character encoding.

brammool:
> M3U8 is just M3U with UTF-8 character encoding.

Information about the files is a bit confusing and sometimes misleading.

Calling them HLS playlists also doesn't seem right, the files can be
used by many players.

"M3U files that are UTF-8 encoded are instead saved in the M3U8 file
format."

The name comes from "MP3 URL", but now other files are also supported,
thus "mp3" isn't quite right.

Something like "URL playlist" would be more accurate, but would users
recognize that name?

What if we just call it "playlist", would that have a problem?


-- 
165. You have a web page burned into your glasses

benoitryder:
> Are the .m3u8 and .m3u files different enough to justify a different filetype?  Would they need different settings, indenting and syntax?

Syntax is compatible. But HLS playlists define elements specific to them, and don't use elements specific to M3U playlists. So both formats are compatible, but neither strict superset nor subset of each other.
Technically, the extension alone is not enough identify a .m3u8/.m3u as an M3U playlist or an HLS playlist. Hence some confusion in previous comments.

The proposed syntax file does not provide highlighting to tags not supported by HLS playlists but used by M3U playlists (but it could). It does not highlight wrong/erroneous constructs as errors; if it did, errors to report would not be the same for both cases.

> What if we just call it "playlist", would that have a problem?

There are other playlist formats that use a completely different syntax (e.g. `.pls`). `m3uplaylist` would be a better compromise, recognizable by users.

brammool:
> > Are the .m3u8 and .m3u files different enough to justify a different
> > filetype?  Would they need different settings, indenting and syntax?
> 
> Syntax is compatible. But HLS playlists define elements specific to
> them, and don't use elements specific to M3U playlists. So both
> formats are compatible, but neither strict superset nor subset of each
> other.
> Technically, the extension alone is not enough identify a .m3u8/.m3u
> as an M3U playlist or an HLS playlist. Hence some confusion in
> previous comments.

Thus an .m3u or .m3u8 file could also be another type, not a HLS
playlist.  If we would have separate file types, we would have to
inspect the contents to see what it is?  Assuming some HLS elements are
found there.

> The proposed syntax file does not provide highlighting to tags not
> supported by HLS playlists but used by M3U playlists (but it could).
> It does not highlight wrong/erroneous constructs as errors; if it did,
> errors to report would not be the same for both cases.
> 
> > What if we just call it "playlist", would that have a problem?
> 
> There are other playlist formats that use a completely different
> syntax (e.g. `.pls`). `m3uplaylist` would be a better compromise,
> recognizable by users.

OK, so we should not use just "playlist".

-- 
167. You have more than 200 websites bookmarked.

benoitryder:
> If we would have separate file types, we would have to inspect the contents to see what it is?

Yes. We could check for a line starting with `#EXT-X-` among the first lines; and if there is one, consider it's an HLS playlist.
Note that some content providers add comments or empty lines at the beginning, so a good heuristic would be need to be smarter (e.g. defaulting to "HLS playlist" if extension is `.m3u8` and M3U if extension is `.m3u`).

brammool:
> > If we would have separate file types, we would have to inspect the
> > contents to see what it is?
> 
> Yes. We could check for a line starting with `#EXT-X-` among the first
> lines; and if there is one, consider it's an HLS playlist.
> Note that some content providers add comments or empty lines at the
> beginning, so a good heuristic would be need to be smarter (e.g.
> defaulting to "HLS playlist" if extension is `.m3u8` and M3U if
> extension is `.m3u`).

Well, we can add that check when it is needed.  For now calling all
those files hlsplaylist would be OK.

So we should also recognize *.m3u, or are only *.m3u8 files used for
hlsplaylist?

-- 
"Hit any key to continue" does _not_ mean you can hit the on/off button!

benoitryder:
> So we should also recognize *.m3u, or are only *.m3u8 files used for hlsplaylist?

Both are technically valid for HLS playlists, so we can recognize both.

brammool:
> Thanks for noticing. I fixed it on my branch but since the the PR was already closed, the diff does not include it.
> Updated file available [here](https://raw.githubusercontent.com/benoitryder/vim/master/runtime/syntax/hlsplaylist.vim).

I'll merge it in.

-- 
Although the scythe isn't pre-eminent among the weapons of war, anyone who
has been on the wrong end of, say, a peasants' revolt will know that in
skilled hands it is fearsome.
					-- (Terry Pratchett, Mort)

PR #11216, opened 2022-09-24
add syntax highlighting for some sqlplus commands and define variables

lee-lindley:
Thanks to Carsten Czarski for fine work adding common sqlplus syntax highlighting.


brammool:
I'll include it, thanks.

PR #11218, opened 2022-09-24
CI: m1/arm64 rm --disable-channel

pheiduck:
maybe we have again failing tests here...


brammool:
I sent out a fix the the test that ignores /private on only one string.

The failure in Test_quit_long_message we have seen before.  In your PR, can you change the value from "rows" from 6 to 10 so that we can see the error that is hiding now?

brammool:
Hmm, the dump diff correctly says that the number of lines differs, but that's not what we wanted to see.
Let me make a patch, it might be useful to see more lines whenever the test fails.

pheiduck:
@brammool it's ready to merge

brammool:
Merged as Patch 9.0.0596 (reference was missing in the submit)

PR #11223, opened 2022-09-25
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   linux:
     runs-on: ubuntu-20.04
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -21,8 +21,15 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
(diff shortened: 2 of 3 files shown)


brammool:
@pheiduck  @k-takata

pheiduck:
@brammool Let's run the CI to see that it still runs with the modification.

pheiduck:
LGTM

Shane-XB-Qian:
coverity.yml failed in foked repo every day, would you refine/fix it?

Shane-XB-Qian:
failing when wget the `coverity_tool.tgz`

Shane-XB-Qian:
yes, it failed at forked vim repo, and sent out failed mail everyday. is it annoyed?

--
shane.xb.qian

Shane-XB-Qian:
> You can disable the workflow.

perhaps good correction should be checking 
was it a main repo or forked repo 
or was token was there or not
before really wget.

Shane-XB-Qian:
or not sent out alert mail as well if it was not main repo?
if it obviously would be failed, why sent it everyday? but let all forked user manually to disable it................... :rofl:

PR #11226, opened 2022-09-25
CI: add patch_script on .cirrus.yml file

pheiduck:
Problem: I think our YAML parser is not happy here because of the :. You should probably extract it in a separate script with ":
Solution: add patch_script
Reported by: Fedor Korotkov


brammool:
> Problem: I think our YAML parser is not happy here because of the :. You should probably extract it in a separate script with &quot;:
> Solution: add patch_script
> Reported by: Fedor Korotkov

If you sync to head this patch should no longer be needed.

-- 
If Microsoft would build a car...
... Occasionally, executing a maneuver such as a left turn
would cause your car to shut down and refuse to restart, in
which case you would have to reinstall the engine.

PR #11227, opened 2022-09-25
fix(docs): add missing word in `g;`

dmyTRUEk:


brammool:
I think "go to position" is correct, "go to cursor position" is confusing, thus changing the "g,"  text would be better.

PR #11228, opened 2022-09-25
Avoiding highlight result of <f-args> as option in map.txt

zeertzjq:


brammool:
Thanks for the suggestions.  I do not like putting backticks on the
<f-args>, these are strings and backticks are for commands.

The first one is highlighted as an option.  That is better solved in the
syntax file.

I'll include the other changes.

-- 
When I die, I want to go peacefully in my sleep like my grandfather.
Not screaming in terror, like his passengers.

PR #11235, opened 2022-09-26
Recognise .mts & .cts as TypeScript files

ObserverOfTime:
See: https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions


brammool:
Included with 9.0.0602

PR #11239, opened 2022-09-27
feat: solidity.vim of syntax

coti-z:
hi 

I looked it up and found that the **file was missing**. How about supporting solidity syntax?


brammool:
Why did you close this?  Looks like the original file uses the MIT license, that should be OK.

PR #11251, opened 2022-09-30
Use <C-x> instead of ^X in example mappings

arp242:
Using literal ^X control characters is possible, but using \<C-X> is much more standard, so use that instead.

This led to some confusion at https://vi.stackexchange.com/q/38705

### DIFF
--- a/runtime/doc/insert.txt
+++ b/runtime/doc/insert.txt
@@ -667,9 +667,10 @@ When the popup menu is displayed there are a few more special keys, see
 |popupmenu-keys|.
 
 Note: The keys that are valid in CTRL-X mode are not mapped.  This allows for
-":map ^F ^X^F" to work (where ^F is CTRL-F and ^X is CTRL-X).  The key that
-ends CTRL-X mode (any key that is not a valid CTRL-X mode command) is mapped.
-Also, when doing completion with 'complete' mappings apply as usual.
+":map <C-f> <C-x><C-f>" to work (where <C-f> is CTRL-F and <C-x> is CTRL-X).
+The key that ends CTRL-X mode (any key that is not a valid CTRL-X mode
+command) is mapped. Also, when doing completion with 'complete' mappings apply
+as usual.
 
 								*E565*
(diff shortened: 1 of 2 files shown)


brammool:
Good idea.  It does require that the "<" flag is not in 'cpoptions', I'll add a remark about that.

Not sure why you use lower case letters, I am used to upper case letters, like ```<C-X>```.

benknoble:
> Not sure why you use lower case letters, I am used to upper case letters, like `<C-X>`.

With the exception of `<C-L>` (where `L` is better than `l` in terms of possible ambiguity), using lowercase letters can be clearer to newcomers who may think that `<C-X>` means hold Control and Shift while typing x. You and I both know that really only Control and x are needed. We've had confusion about this before on vi.stackexchange.com IIRC.

And, if I understand the history correctly, there used to be no real difference in Control-Shift-x and Control-x due to the way terminals handled input; that is changing rapidly, and it may one day be the case that they are quite different?

arp242:
The convention is to use capitals for the most part:

    % ag -o '<C-[a-z]>' | wc -l
    11
    % ag -o '<C-[A-Z]>' | wc -l
    427

I just use lower-case because it's easier to type and I'm lazy.

I do agree lower-case is less confusing for those that aren't very familiar to people who aren't well versed in how terminal keys work (i.e. most people).

brammool:
> The convention is to use capitals for the most part:
> 
>     % ag -o '<C-[a-z]>' | wc -l
>     11
>     % ag -o '<C-[A-Z]>' | wc -l
>     427
> 
> I just use lower-case because it's easier to type and I'm lazy.
> 
> I do agree lower-case is less confusing for those that aren't very
> familiar to people who aren't well versed in how terminal keys work
> (i.e. most people).

On the other hand, if you do want to include the shift key you need to
use <C-S-X>.   Changing <C-x> to <C-X> doesn't have an effect.
Thus using lower case is confusing in that way.

-- 
260. Co-workers have to E-mail you about the fire alarm to get
     you out of the building.

PR #11257, opened 2022-10-01
Input redirection doesn't work with powershell

yegappan:
Patch contributed by Enan Ajmain.

### DIFF
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -1531,56 +1531,70 @@ make_filter_cmd(
 {
     char_u	*buf;
     long_u	len;
-
-#if defined(UNIX)
-    int		is_fish_shell;
     char_u	*shell_name = get_isolated_shell_name();
+    int		is_powershell;
+#ifdef UNIX
+    int		is_fish_shell;
+#endif
 
     if (shell_name == NULL)
 	return NULL;
 
+#if defined(UNIX)
     // Account for fish's different syntax for subshells
     is_fish_shell = (fnamecmp(shell_name, "fish") == 0);
-    vim_free(shell_name);
     if (is_fish_shell)
 	len = (long_u)STRLEN(cmd) + 13;		// "begin; " + "; end" + NUL
     else
 #endif
+    {
+	is_powershell = (shell_name[0] == 'p')
+			&& (fnamecmp(shell_name, "powershell") == 0
+				|| fnamecmp(shell_name, "powershell.exe") == 0
+				|| fnamecmp(shell_name, "pwsh") == 0
+				|| fnamecmp(shell_name, "pwsh.exe") == 0);
 	len = (long_u)STRLEN(cmd) + 3;			// "()" + NUL
+    }
+
     if (itmp != NULL)
-	len += (long_u)STRLEN(itmp) + 9;		// " { < " + " } "
+    {
+	if (is_powershell)
+	    // "& { Get-Content " + " | & " + " }"
+	    len += (long_u)STRLEN(itmp) + 24;
+	else
+	    len += (long_u)STRLEN(itmp) + 9;		// " { < " + " } "
+    }
     if (otmp != NULL)
 	len += (long_u)STRLEN(otmp) + (long_u)STRLEN(p_srr) + 2; // "  "
+
+    vim_free(shell_name);
+
     buf = alloc(len);
     if (buf == NULL)
 	return NULL;
 
-#if defined(UNIX)
-    /*
-     * Put braces around the command (for concatenated commands) when
-     * redirecting input and/or output.
-     */
-    if (itmp != NULL || otmp != NULL)
+    if (is_powershell)
     {
-	if (is_fish_shell)
-	    vim_snprintf((char *)buf, len, "begin; %s; end", (char *)cmd);
+	if (itmp != NULL)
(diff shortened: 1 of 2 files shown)


src/ex_cmds.c
```diff
@@ -1531,26 +1531,47 @@ make_filter_cmd(
...
+    {
+	is_powershell = (shell_name[0] == 'p')
+			&& (fnamecmp(shell_name, "powershell") == 0
+				|| fnamecmp(shell_name, "powershell.exe") == 0
+				|| fnamecmp(shell_name, "pwsh") == 0
+				|| fnamecmp(shell_name, "pwsh.exe") == 0);
+	if (is_powershell)
+	    len = (long_u)STRLEN(cmd) + 3;
+	else
+	    len = (long_u)STRLEN(cmd) + 3;			// "()" + NUL
```
3N4N: Check for powershell looks redundant.

src/ex_cmds.c
```diff
@@ -1531,26 +1531,47 @@ make_filter_cmd(
...
     // Account for fish's different syntax for subshells
     is_fish_shell = (fnamecmp(shell_name, "fish") == 0);
-    vim_free(shell_name);
     if (is_fish_shell)
 	len = (long_u)STRLEN(cmd) + 13;		// "begin; " + "; end" + NUL
     else
 #endif
-	len = (long_u)STRLEN(cmd) + 3;			// "()" + NUL
+    {
+	is_powershell = (shell_name[0] == 'p')
```
3N4N, line 1551: PowerShell is nowadays available for Linux, too, in the form of PowerShell Core (v7). (Ignore if we don't care about that too much; who would use powershell in Linux anyway?)

src/ex_cmds.c
```diff
@@ -1531,26 +1531,47 @@ make_filter_cmd(
...
+	else
+	    len = (long_u)STRLEN(cmd) + 3;			// "()" + NUL
+    }
+
     if (itmp != NULL)
-	len += (long_u)STRLEN(itmp) + 9;		// " { < " + " } "
+    {
+	if (is_powershell)
+	    // "& { Get-Content" + " | & " + " }"
+	    len += (long_u)STRLEN(itmp) + 23;
```
3N4N: There is a space after `Get-Content` (see L1603). Addition to `len` should be 24, I think.

src/ex_cmds.c
3N4N: Never mind. I misread `#endif` as `#else`.

src/ex_cmds.c
```diff
@@ -1575,45 +1596,61 @@ make_filter_cmd(
 	STRCAT(buf, itmp);
     }
 #else
-    // For shells that don't understand braces around commands, at least allow
-    // the use of commands in a pipe.
-    if (*p_sxe != NUL && *p_sxq == '(')
+    if (is_powershell)
```
3N4N: Okay, here is where I wanted to mention powershell being available for Linux, too (mistakenly did so on a previous line). Taking the `if is_powershell` block _above_ `#ifdef UNIX` should be enough. (Unless we don't care for people using PowerShell on Linux, assuming anyone does.)

src/ex_cmds.c
yegappan: Yes.  In the original patch, the length was set to zero.  But that broke some other functionality.  When I reverted that change, I forgot to remove this.

src/ex_cmds.c
yegappan: Changed the value to 24.

src/ex_cmds.c
yegappan: Moved the powershell handling code to outside of the '#ifdef/#endif'.

3N4N:
The patch now looks like it should be fine. I can't tell why the tests are failing (haven't looked at the test files yet).

brammool:
The current patch checks for powershell on any system.
It is available on Linux, but does it use exactly the same syntax then?
Is it even practical to set 'shell' to powershell on any other system
than MS-Windows?

-- 
256. You are able to write down over 250 symptoms of being an internet
     addict, even though they only asked for 101.

3N4N:
> It is available on Linux, but does it use exactly the same syntax then?

There may be differences in some cases, but as far as I have used it I haven't found any.

> Is it even practical to set 'shell' to powershell on any other system
> than MS-Windows?

I don't think so. First, practically no one will use it. Second, there are several bugs/oddities in pwsh in Linux: e.g., using `sort` with `Start-Process sort -RedirectStandardInput in.txt` works fine, but if we redirect it to a file (`-RedirectStandardOutput out.txt`) then it eats all the empty lines.

But I want to leave it to the maintainer (you, Bram) to decide. We can always take the other choice later, if required; it's not a huge change.

vim-ml:
Hi Bram,

On Mon, Oct 3, 2022 at 5:06 AM Bram Moolenaar ***@***.***>
wrote:

>
> The current patch checks for powershell on any system.
> It is available on Linux, but does it use exactly the same syntax then?
> Is it even practical to set 'shell' to powershell on any other system
> than MS-Windows?
>
>
>
Setting 'shell' to 'pwsh' or 'powershell' (Powershell) was supported on
Linux before
this patch.  But "%!sort" was failing on Linux (just like on MS-Windows)
before
this patch.  This patch makes it work on both Linux and MS-Windows.

- Yegappan

PR #11258, opened 2022-10-01
'splitscroll' option is subjective

luukvbaal:
Problem:	'splitscroll' option is subjective.
Solution:	Replace with 'splitkeep' option that provides an option to keep the relative cursor position, screen lines or topline the same.

@brammool I think some user feedback on this patch is warranted(is keeping the topline useful?) although now that I went through the trouble of making the option non-boolean we might as well include it. I first wanted to keep the option simple but I can see how some users might prefer to keep the topline the same instead, and the change required turned out not to be too involved.

@ychin does 'set splitkeep=topline' behave as you would expect with this patch?


ychin:
Yes! This is exactly what I was hoping for. Thanks a lot for spending the time on this. Playing around with this more, I can definitely see why some people may want `splitkeep=screen` (or even the original `splitkeep=cursor`), but to me the new `splitkeep=topline` is still probably what I prefer the most, since it seems to be the least surprising to me and most "stable" (in its own way).

PR #11265, opened 2022-10-03
Display patchlevel on first line in :version

arp242:
Currently the intro screen displays the version with the patch number:

                 VIM - Vi IMproved

                  version 9.0.648
             by Bram Moolenaar et al.
    Vim is open source and freely distributable

But the :version command and "vim --version" displays it a bit different:

	:version
	VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Oct  3 2022 15:19:47)
	Included patches: 1-648
	Compiled by martin@x270.arp242.net

Many times I've seen people report only the first line from :version, which isn't all that useful; almost everyone is using *some* patchlevel, and "Vim 8.2" was 2.5 years of development, 8.1 a year and a half, etc.

### DIFF
--- a/src/version.c
+++ b/src/version.c
@@ -69,7 +69,7 @@ init_longVersion(void)
 	size_t len = strlen(msg)
 		    + strlen(VIM_VERSION_LONG_ONLY)
 		    + strlen(VIM_VERSION_DATE_ONLY)
-		    + hp
+		    + 8 // For hp; should be more than enough.
 		    + strlen(date_time);
 
 	longVersion = alloc(len);
--- a/src/version.c
+++ b/src/version.c
@@ -64,10 +64,12 @@ init_longVersion(void)
 #else
 	char *date_time = __DATE__ " " __TIME__;
 #endif
-	char *msg = _("%s (%s, compiled %s)");
+	int hp = highest_patch();
+	char *msg = _("%s.%d (%s, compiled %s)");
 	size_t len = strlen(msg)
 		    + strlen(VIM_VERSION_LONG_ONLY)
 		    + strlen(VIM_VERSION_DATE_ONLY)
+		    + hp
 		    + strlen(date_time);
 
 	longVersion = alloc(len);
@@ -75,7 +77,7 @@ init_longVersion(void)
 	    longVersion = VIM_VERSION_LONG;
 	else
 	    vim_snprintf(longVersion, len, msg,
-		      VIM_VERSION_LONG_ONLY, VIM_VERSION_DATE_ONLY, date_time);
+		      VIM_VERSION_LONG_ONLY, hp, VIM_VERSION_DATE_ONLY, date_time);
     }
 }
 # endif


src/version.c
```diff
@@ -64,18 +64,20 @@ init_longVersion(void)
 #else
 	char *date_time = __DATE__ " " __TIME__;
 #endif
-	char *msg = _("%s (%s, compiled %s)");
+	int hp = highest_patch();
+	char *msg = _("%s.%d (%s, compiled %s)");
```
ychin, line 68: I think this will break localized strings (Vim uses `gettext` for localization). E.g. if you look at `src/po/zh_CN.UTF-8.po` or `src/po/es.po` you will see entries for this string. You can test this by running `LANG=es_ES.UTF-8 vim` in terminal.

src/version.c
```diff
@@ -64,18 +64,20 @@ init_longVersion(void)
...
 	size_t len = strlen(msg)
 		    + strlen(VIM_VERSION_LONG_ONLY)
 		    + strlen(VIM_VERSION_DATE_ONLY)
+		    + hp
```
ychin: This seems wrong. `len` is the length of the string, which should be a fixed value rather that a runtime value. E.g. `msg_outnum` in message.c just uses 20.

src/version.c
arp242: It will, but I'm not sure what the best approach for this is? I suppose I could update all the strings in the .po files(?)

src/version.c
arp242: Oops, yes; fixed it.

src/version.c
ychin: @brammool can answer better but I think yes, you have to update all the .po files. Note that not all .po files contain entries for this string as some decided not to translate this. They have individual owners though, but I'm not sure if a change like this which are more programmatic in nature than changing the texts need a review from them (I would guess not).

brammool:
> Currently the intro screen displays the version with the patch number:
> 
>                  VIM - Vi IMproved
> 
>                   version 9.0.648
>              by Bram Moolenaar et al.
>     Vim is open source and freely distributable
> 
> But the :version command and "vim --version" displays it a bit different:
> 
> 	:version
> 	VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Oct  3 2022 15:19:47)
> 	Included patches: 1-648
> 	Compiled by ***@***.***
> 
> Many times I've seen people report only the first line from :version,
> which isn't all that useful; almost everyone is using *some*
> patchlevel, and "Vim 8.2" was 2.5 years of development, 8.1 a year and
> a half, etc.

Hmm, I'm not sure I want to change this.  "9.0" is the minor version.
The patchlevel is separate.  If you look at some distributions, they
cherry pick certain (security) patches, and the highest patch number
does not represent the build.  The one I have on Ubuntu:

Included patches: 1-3995, 4563, 4646, 4774, 4895, 4899, 4901, 4919

We could check for the continuous range, but then we get stuck at some
level if the builder leaves out one.

For when people report their version, they should report the "Included
patches", otherwise the information is not complete.


-- 
Women are probably the main cause of free software starvation.

brammool:
> @ychin commented on this pull request.
> 
> 
> > @@ -64,18 +64,20 @@ init_longVersion(void)
>  #else
>  	char *date_time = __DATE__ " " __TIME__;
>  #endif
> -	char *msg = _("%s (%s, compiled %s)");
> +	int hp = highest_patch();
> +	char *msg = _("%s.%d (%s, compiled %s)");
> 
> @brammool can answer better but I think yes, you have to update all
> the .po files. They have individual owners though, but I'm not sure if
> a change like this which are more programmatic in nature than changing
> the texts need a review (I would guess not).

I think in this case I rather not make the change.

But anyway, since translation updates are always behind, and this change
only adds something that would not change when translating, updating all
the .po files is the best solution.

-- 
There is a fine line between courage and foolishness.
Unfortunately, it's not a fence.

arp242:
It's not 100% clear to me if you want to add this or not @brammool? I can update the .po files, or feel free to close this PR if you just don't want to include it regardless.

brammool:
Let me repeat: "I think in this case I rather not make the change.".
We can tell people to include more lines from the :version output.  Some builds also mention the system.  On Ubuntu
it mentions "Compiled by team+vim@tracker.debian.org", which can also be useful.

If there are places where we should be clearer, we could change those.

arp242:
> Let me repeat: "I think in this case I rather not make the change.".

I guess the second paragraph where you suggested updating the po files threw me off and that you didn't close it. No worries.

PR #11266, opened 2022-10-03
Remove VisVim

arp242:
This removes the "VisVim", which allowed Vim to run inside Visual Studio. It only worked up to VS 6, which is from 1998. The last change other than generic typo fixes was for Vim 7.3.278, in 2011.

It's not built by default, I can find no references to it on the issue tracker or mailing list, you'll have a hard time to use it with anything especially now that XP support is removed, and newer VS versions (since 2002) have it built in. It's probably fairly safe to say no one is using it.


arp242:
Builds failures seem related to b6aab8f

brammool:
Included with Patch 9.0.0698

PR #11268, opened 2022-10-03
Always include FEAT_CMDWIN, alias 'tiny' build to 'small'

arp242:
FEAT_CMDWIN is the only difference between the "tiny" and "small" versions, and the difference in generated code is very small (4 to 8K, depending on compile flags):j

	      | -O2                         | -Os
	tiny  | 1633984 (1.6M)              | 1190576 (1.2M)
	small | 1638240 (1.6M), +4256 bytes | 1198928 (1.2M), +8352 bytes

Even on very small and constrained systems this increase is negligible, and there are quite a few #ifdefs that can be removed.

Since the 'tiny' and 'small' builds are now identical this removes the 'tiny' build altogether, but make --with-features=tiny behave like 'small' was given for compatibility so no one's build should break.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,7 +6,6 @@ environment:
   matrix:
     - FEATURE: HUGE
 # disabled
-#    - FEATURE: TINY
 #    - FEATURE: SMALL
 #    - FEATURE: NORMAL
 #    - FEATURE: BIG
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -1124,7 +1124,6 @@ Also see |`=|.
 In the command-line window the command line can be edited just like editing
 text in any window.  It is a special kind of window, because you cannot leave
 it in a normal way.
-{not available when compiled without the |+cmdwin| feature}
 
 
 OPEN						*c_CTRL-F* *q:* *q/* *q?*
--- a/runtime/doc/various.txt
+++ b/runtime/doc/various.txt
@@ -318,8 +318,7 @@ g8			Print the hex values of the bytes used in the
 			Here is an overview of the features.
 			The first column shows the smallest version in which
 			they are included:
-			   T	tiny (always)
-			   S	small
+			   S	small (always)
 			   N	normal
 			   B	big
 			   H	huge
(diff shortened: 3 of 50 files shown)


brammool:
> FEAT_CMDWIN is the only difference between the "tiny" and "small" versions, and the difference in generated code is very small (4 to 8K, depending on compile flags):j
> 
> 	      | -O2                         | -Os
> 	tiny  | 1633984 (1.6M)              | 1190576 (1.2M)
> 	small | 1638240 (1.6M), +4256 bytes | 1198928 (1.2M), +8352 bytes
> 
> Even on very small and constrained systems this increase is negligible, and there are quite a few #ifdefs that can be removed.
> 
> Since the 'tiny' and 'small' builds are now identical this removes the
> 'tiny' build altogether, but make --with-features=tiny behave like
> 'small' was given for compatibility so no one's build should break.

Thanks for making this PR.

I prefer keeping the "tiny" build as the name for the merged tiny and
small builds.  I think more people used tiny and hardly anyone was using
the small build.  I'll make that adjustment (I see a few related things
that could be improved).

-- 
God made the integers; all else is the work of Man.
		-- Kronecker

arp242:
> I prefer keeping the "tiny" build as the name for the merged tiny and small builds. I think more people used tiny and hardly anyone was using the small build. I'll make that adjustment (I see a few related things that could be improved).

I used "small" because at 1.6M it's not really "tiny", so it seemed like a better name to me, and we probably want to maintain compatibility in `--with-features` forever anyway so I'm not sure it matters if 'tiny' or 'small' is used more frequently.

Anyway, I can change and test if you want.

brammool:
> > I prefer keeping the "tiny" build as the name for the merged tiny and
> > small builds. I think more people used tiny and hardly anyone was
> > using the small build. I'll make that adjustment (I see a few related
> > things that could be improved).
> 
> I used "small" because at 1.6M it's not really "tiny", so it seemed like
> a better name to me, and we probably want to maintain compatibility in
> `--with-features` forever anyway so I'm not sure it matters if 'tiny' or
> 'small' is used more frequently.
> 
> Anyway, I can change and test if you want.

I have done it already.  It actually avoids some renaming, also a reason
it works better.

Since quite a few features were "graduated", no longer in #ifdefs, the
smallest Vim has grown.  Can't really call it "small" or "tiny" if you
look at the executable size.  But excluding +eval and +syntax kind of
justifies that name.


-- 
GOD: That is your purpose Arthur ... the Quest for the Holy Grail ...

PR #11274, opened 2022-10-04
Add sound support for macOS

ychin:
This finally adds +sound support for Vim on macOS. In macOS, there are *quite* a few audio libraries to choose from. In the end, I decided to go with AppKit which is the user-level library which we already use for interfacing with the system clipboard. That's because the NSSound class from it is the easiest to use and essentially plug-and-play (with the caveat of needing a run loop as described below). Other libraries like CoreAudio are much too powerful and overkill for Vim's use case of just needing to play a file and be done with it, and they also involve a bit more code to set up and use.

One caveat of using NSSound is that it's an Objective C class so I had to do the implementation in os_macosx.m, and call that to/from sound.c. In order for the sound callback to work, I also had to set up a run loop that's ticked periodically in order to make sure the delegate will get invoked whenever the sound finishes playing. The run loop is ticked the same time as when we would have called `invoke_sound_callback()` in the Linux / Canberra implementation.

For sound event names, I'm using the builtin NSSound's `soundNamed` function which can locate the set of default installed sound files (e.g. "Tink", "Sosumi", "Submarine", "Bottle", etc). I was looking up how to get the more specific sounds (e.g. the "empty trash can" sound), and I think NSSound doesn't know how to load them (I think Core Audio has functions to load them but I don't think that's worth the trouble of using Core Audio). Looking up from
https://macreports.com/how-to-change-system-sounds-empty-trash-screenshot-sent-message-etc-on-mac/, it seems like those files can be located under
`/System/Library/Components/CoreAudio.component/Contents/SharedSupport/SystemSounds/`, but I didn't feel like adding special code to map those files to event names. Interested plugin authors or users can simply hard-code those path names on macOS.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -8630,6 +8630,9 @@ sound_playevent({name} [, {callback}])
 <		On MS-Windows, {name} can be SystemAsterisk, SystemDefault,
 		SystemExclamation, SystemExit, SystemHand, SystemQuestion,
 		SystemStart, SystemWelcome, etc.
+		On macOS, {name} refers to files located in
+		/System/Library/Sounds (e.g. "Tink").  It will also work for
+		custom installed sounds in folders like ~/Library/Sounds.
 
 		When {callback} is specified it is invoked when the sound is
 		finished.  The first argument is the sound ID, the second
--- a/runtime/doc/todo.txt
+++ b/runtime/doc/todo.txt
@@ -450,8 +450,6 @@ find them. (Max Kukartsev, #6218)
 
 Enable 'termbidi' if $VTE_VERSION >= 5703 ?
 
-Sound: support on Mac?  Or does libcanberra work there?
-
 Python 3.8 doesn't work. (Antonios Hadjigeorgalis, #5509)
 
 "--cleanFOO" does not result in an error. (#5537)
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -4553,7 +4553,7 @@ if test "$MACOS_X" = "yes"; then
   AC_MSG_CHECKING([whether we need macOS frameworks])
   if test "$MACOS_X_DARWIN" = "yes"; then
     if test "$features" = "tiny"; then
-      dnl Since no FEAT_CLIPBOARD, no longer need for os_macosx.m.
+      dnl Since no FEAT_CLIPBOARD or FEAT_SOUND, no longer need for os_macosx.m.
       OS_EXTRA_SRC=`echo "$OS_EXTRA_SRC" | sed -e 's+os_macosx.m++'`
       OS_EXTRA_OBJ=`echo "$OS_EXTRA_OBJ" | sed -e 's+objects/os_macosx.o++'`
       AC_MSG_RESULT([yes, we need CoreServices])
--- a/src/feature.h
+++ b/src/feature.h
@@ -484,7 +484,7 @@
 #endif
 
 /*
- * sound - currently only with libcanberra
+ * sound
  */
 #if !defined(FEAT_SOUND) && defined(HAVE_CANBERRA)
 # define FEAT_SOUND
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -2326,6 +2326,10 @@ parse_queued_messages(void)
 # ifdef FEAT_TERMINAL
 	free_unused_terminals();
 # endif
+
+# ifdef FEAT_SOUND_MACOSX
+	process_cfrunloop();
+# endif
 # ifdef FEAT_SOUND_CANBERRA
 	if (has_sound_callback_in_queue())
 	    invoke_sound_callback();
--- a/src/os_macosx.m
+++ b/src/os_macosx.m
@@ -384,6 +384,139 @@ timer_delete(timer_t timerid)
 
 #endif /* FEAT_RELTIME */
 
+#ifdef FEAT_SOUND
+
+static NSMutableDictionary<NSNumber*, NSSound*> *sounds_list = nil;
+
+/// A delegate for handling when a sound has stopped playing, in
+/// order to clean up the sound and to send a callback.
+@interface SoundDelegate : NSObject<NSSoundDelegate>;
+
+- (id) init:(long) sound_id callback:(soundcb_T*) callback;
+- (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag;
+
+@property (readonly) long sound_id;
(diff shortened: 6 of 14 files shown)


ychin:
I'm not sure why testing is reporting coverage drops. Looking through it, it appears that it doesn't use Mac testing as part of the consideration? The new sound logic should still be tested by the existing sound unit test functionality.

brammool:
Ignore coverage reports on a PR, they are not reliable.

Have you tried playing the killer sheep game?  https://github.com/vim/killersheep

ychin:
Oh that's a good point. Let me try it out. I just booted it and it's not playing sound. Let me take a quick look why.

ychin:
Ok an update. I think KillerSheep incorrectly calculated the `g:killersheep_sound_ext` value to be "ogg" (Apple doesn't support ogg), and after making the fix to "mp3" it worked! I did manage to kill some poor innocent 🐑 after the fix.

The change to your plugin is as follows:

```diff
diff --git a/autoload/killersheep.vim b/autoload/killersheep.vim
index 276c7d7..51f61a2 100644
--- a/autoload/killersheep.vim
+++ b/autoload/killersheep.vim
@@ -158,7 +158,7 @@ func s:Init()
   hi def KillerLevelX ctermbg=yellow guibg=yellow
 
   if !exists('g:killersheep_sound_ext')
-    if has('win32') || len(s:sound_cmd)
+    if has('win32') || has('osx') || len(s:sound_cmd)
       " most systems can play MP3 files
       let g:killersheep_sound_ext = ".mp3"
     else
```

Also interesting how you are using the command-line tool `afplay` and spawning it using `job_start()` to get it working on macOS before. I did wonder about how much difference it makes when implementing this patch as a dedicated user could already use that (albeit a little more clunky than a native API).

dpelle:
This is strange.  On my macOS M1 laptop and without this PR, using vim-9.0.656): 
* `:version` shows `-sound`  (so no sound)
* running `make test_sound` skips sound tests (since there is no sound) 
* yet running `:KillKillKill`  I do get the sound?!

ychin:
That's what I was talking about. The plug-in used the command-line tool `afplay` to play audio before since there was no native sound support on Mac. Ultimately it still works albeit needing to spawn another process for each sound clip.

brammool:
> Ok an update. I think KillerSheep incorrectly calculated the
> `g:killersheep_sound_ext` value to be "ogg" (Apple doesn't support
> ogg), and after making the fix to "mp3" it worked! I did manage to
> kill some poor innocent 𐟐‘ after the fix.

Well, innocent, I'm not sure.

> The change to your plugin is as follows:
> 
> ```diff
> diff --git a/autoload/killersheep.vim b/autoload/killersheep.vim
> index 276c7d7..51f61a2 100644
> --- a/autoload/killersheep.vim
> +++ b/autoload/killersheep.vim
> @@ -158,7 +158,7 @@ func s:Init()
>    hi def KillerLevelX ctermbg=yellow guibg=yellow
>  
>    if !exists('g:killersheep_sound_ext')
> -    if has('win32') || len(s:sound_cmd)
> +    if has('win32') || has('osx') || len(s:sound_cmd)
>        " most systems can play MP3 files
>        let g:killersheep_sound_ext = ".mp3"
>      else
> ```

I'll commit that, thanks.

> Also interesting how you are using the command-line tool `afplay` and
> spawning it using `job_start()` to get it working on macOS before. I
> did wonder about how much difference it makes when implementing this
> patch as a dedicated user could already use that (albeit a little more
> clunky than a native API).

It more or less worked.  Trial and error...  I don't have a Mac anymore
thus I can't try it out now.

-- 
       We're knights of the Round Table
       Our shows are formidable
       But many times
       We're given rhymes
       That are quite unsingable
       We're opera mad in Camelot
       We sing from the diaphragm a lot.

ychin:
If you want another person to try this pull request out first to make sure it works, maybe @dpelle could help? (sorry for volunteering you)

dpelle:
@ychin wrote:
> maybe @dpelle could help?

I can try it, but I have limited time. I'll try in my weekend.

lifepillar:
I'd like to try this. How do you enable the sound feature? It's off in a standard build.

lifepillar:
>It's off in a standard build.

My bad, I fired the wrong Vim. It's building and working fine!

>The plug-in used the command-line tool `afplay` to play audio before

Spawning a process for each sound event is not idea, thoughl: having built-in sound as in this PR is much more desirable. Thanks for proposing this!

ychin:
The implementation doesn't do any caching right now. If you are using `sound_playevent` that would be cached for sure, but `sound_playfile` is currently calling the same API every time and doesn't reuse existing loaded sound objects, so we are relying on NSSound to cache for us, which I don't know if it does or not.

We could probably implement caching if we really want to, but it would complicate the behavior a bit in deciding how much to cache etc. I looked at the other implementations and the Linux/Canberra one seems to do some caching via the "volatile" flag although I don't know how much caching that does, and I'm not sure about the Windows implementation. NSSound also doesn't really expose the internals of loading a sound, so caching logic would only work if you play the same sound clip after it has finished playing. If we want a more generic caching mechanism we may need to use a deeper API.

brammool:
> The implementation doesn't do any caching right now. If you are using
> `sound_playevent` that would be cached for sure, but `sound_playfile`
> is currently calling the same API every time and doesn't reuse
> existing loaded sound objects, so we are relying on NSSound to cache
> for us, which I don't know if it does or not.
> 
> We could probably implement caching if we really want to, but it would
> complicate the behavior a bit in deciding how much to cache etc. I
> looked at the other implementations and the Linux/Canberra one seems
> to do some caching via the "volatile" flag although I don't know how
> much caching that does, and I'm not sure about the Windows
> implementation. NSSound also doesn't really expose the internals of
> loading a sound, so caching logic would only work if you play the same
> sound clip after it has finished playing. If we want a more generic
> caching mechanism we may need to use a deeper API.

Before implementing caching, check the timing.  If you have an SSD, like
most recent sytems, the load time is likely very short anyway, less than
a msec.  On top of that, the OS will do file caching.  File systems have
been optimized over decades, often you can just rely on that.  Your
biggest problem might be how you want to measure this...

-- 
"Intelligence has much less practical application than you'd think."
		  -- Scott Adams, Dilbert.

ychin:
Yeah I'm definitely not suggesting I would work on caching functionality for first pass at all. For all I know the current implementation should be good to merge but not sure if you want more Mac people to concur it works first.

PR #11276, opened 2022-10-04
WinNew to provide window-ID

### LINKED ISSUE #10633: `WinNew` to provide window-ID
It is impossible to get an ID of the new window when `WinNew` is triggered, `<amatch>` expands to an empty string.

Allow expanding `<amatch>` to window-ID during `WinNew`, similar to how `WinClosed` works.

Iterating through all windows and sorting by window ID. The largest ID supposedly belongs the new window.

noscript:
Fixed #10633.

### DIFF
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -1361,9 +1361,11 @@ WinLeave			Before leaving a window.  If the window to be
 				Not used for ":qa" or ":q" when exiting Vim.
 
 							*WinNew*
-WinNew				When a new window was created.  Not done for
-				the first window, when Vim has just started.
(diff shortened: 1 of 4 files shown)


brammool:
Matching the new window ID in the pattern does not appear useful, since the ID is not known yet.  Only once the window exists you can use the window ID, that is why it works with WinClose.
The proposed tests only show a pattern that guesses the next window ID, I cannot think of a practical use for this.

To get the window ID in the WinNew autocommand win_getid() can be used.

noscript:
True, from perspective of ID matching this doesn't make sense. Somehow I forgot about win_getid().

PR #11283, opened 2022-10-04
Alias 'big' version to 'normal'

arp242:
The difference between the 'normal' and 'big' builds is not very large:

	normal    | 3195080 (3.1M)        | 2514680 (2.4M)
	big       | 3261704 (3.2M), +65K  | 2560856 (2.5M), +45K

And I'm not sure many people are using the 'big' featureset.

This removes the 'big' featureset, and aliases that to 'normal', leaving us with three builds:

	          | -O2            | -Os
	tiny      | 1633984 (1.6M) | 1190576 (1.2M)
	normal    | 3195080 (3.1M) | 2514680 (2.4M)
	huge      | 3410232 (3.3M) | 3070104 (3.0M)

Which are all fairly distinct.

Features now included in 'normal':

	FEAT_LANGMAP
	FEAT_KEYMAP
	FEAT_EMACS_TAGS
	FEAT_CSCOPE
	FEAT_CONCEAL
	FEAT_TERMGUICOLORS
	FEAT_VARTABS
	FEAT_MOUSE_NET, FEAT_MOUSE_DEC, FEAT_MOUSE_URXVT
	FEAT_SIGNS
	FEAT_AUTOCHDIR
	FEAT_RIGHTLEFT  (unless --disable-rightleft)
	FEAT_ARABIC     (unless --disable-arabic)
	FEAT_SODIUM     (unless --disable-sodium, or libsodium isn't found)
	FEAT_SOUND      (unless --disable-canberra, or libcanberra not found)
	FEAT_MBYTE_IME  (Haiku)

### DIFF
--- a/runtime/doc/various.txt
+++ b/runtime/doc/various.txt
@@ -332,8 +332,8 @@ g8			Print the hex values of the bytes used in the
 							*+feature-list*
    *+acl*		|ACL| support included
    *+ARP*		Amiga only: ARP support included
-B  *+arabic*		|Arabic| language support
-B  *+autochdir*		support 'autochdir' option
+N  *+arabic*		|Arabic| language support
+N  *+autochdir*		support 'autochdir' option
 T  *+autocmd*		|:autocmd|, automatic commands.  Always enabled since
 			8.0.1564
 H  *+autoservername*	Automatically enable |clientserver|
@@ -358,9 +358,9 @@ T  *+cmdline_hist*	command line history |cmdline-history|
 N  *+cmdline_info*	|'showcmd'| and |'ruler'|
 T  *+cmdwin*		|cmdline-window| support; Always enabled since 9.0.0657
 T  *+comments*		|'comments'| support
-B  *+conceal*		"conceal" support, see |conceal| |:syn-conceal| etc.
+N  *+conceal*		"conceal" support, see |conceal| |:syn-conceal| etc.
 N  *+cryptv*		encryption support |encryption|
-B  *+cscope*		|cscope| support
+N  *+cscope*		|cscope| support
 T  *+cursorbind*	|'cursorbind'| support
 m  *+cursorshape*	|termcap-cursor-shape| support
 m  *+debug*		Compiled for debugging.
@@ -371,7 +371,7 @@ N  *+diff*		|vimdiff| and 'diff'
 N  *+digraphs*		|digraphs| *E196*
    *+directx*		Win32 GUI only: DirectX and |'renderoptions'|
    *+dnd*		Support for DnD into the "~ register |quote_~|.
-B  *+emacs_tags*	|emacs-tags| files
+N  *+emacs_tags*	|emacs-tags| files
 N  *+eval*		expression evaluation |eval.txt|
 T  *+ex_extra*		always on now, used to be for Vim's extra Ex commands
 N  *+extra_search*	|'hlsearch'| and |'incsearch'| options.
@@ -397,9 +397,9 @@ T  *+insert_expand*	|insert_expand| Insert mode completion
 m  *+ipv6*		Support for IPv6 networking |channel|
 m  *+job*		starting and stopping jobs |job|
 T  *+jumplist*		|jumplist|; Always enabled since 8.2.3795
-B  *+keymap*		|'keymap'|
+N  *+keymap*		|'keymap'|
 N  *+lambda*		|lambda| and |closure|
-B  *+langmap*		|'langmap'|
+N  *+langmap*		|'langmap'|
 N  *+libcall*		|libcall()|
 N  *+linebreak*		|'linebreak'|, |'breakat'| and |'showbreak'|
 t  *+lispindent*	|'lisp'|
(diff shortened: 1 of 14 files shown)


arp242:
> This change does not make the code simpler, #ifdefs remain. It does remove a choice, which has always been a bit confusing, there is no clear argumentation why a feature is in normal, big or only in huge.

Yes, my motivation was that I was looking at what exactly the different feature sets actually do and I found that the differences between "small" and "tiny" and "normal" and "big" were actually pretty small.

IMHO it's confusing now what the differences are. It's not really a "code thing" to make the code easier, but to make it easier for people who want to build Vim. That's also why I sent the other patch to merge the "tiny" and "small" builds; that it removed a bunch of #ifdefs was more of a side-effect than anything else.

I think the real question is: how often are people using the 'normal' or 'big' builds in the first place? My suspicion is that most people are using the 'huge' build, and a small subsection the 'tiny' build.

For example Ubuntu and Debian have the "vim" package (huge) and "vim-tiny", but no "vim-big" or "vim-normal". Alpine Linux (focused on being small) just has the "huge" build. Arch Linux just has a package for the "huge" build too. There are countless of distros, but I can't find any that uses the 'normal' or 'big' builds. The Windows installer is 'huge', macvim is 'huge'.

So in that sense I'm not sure if it really matters what we include in 'normal' or 'huge'; I'm fine with any choice on this, it's the removal of a feature set that's the improvement IMHO. I think we could probably just merge the 'normal' and 'huge' builds too (leaving just 'tiny' and 'huge') and no one will notice.

> I can't guess how many users there are for each feature, such as +langmap, but it's easier to say "if you want language-specific features, use huge".

It seems to come up fairly frequently: https://vi.stackexchange.com/search?q=langmap

---

Aside: organising the docs in +feature-list could also help maybe, like so:

	Features in tiny version (always enabled):
		*+autocmd*		|:autocmd|, automatic commands.
		*+cursorbind*	|'cursorbind'| support
		.. etc..
	
	Features in normal version (includes everything in tiny):
		*+arabic*		|Arabic| language support
		*+autochdir*		support 'autochdir' option
		.. etc..

But it still leaves with the thing that the differences are pretty small and that AFAIK few people are using the 'normal' or 'big' versions.

vim-ml:
* Martin Tournoij ***@***.***> [221004 14:16]:
> > This change does not make the code simpler, #ifdefs remain. It does
> > remove a choice, which has always been a bit confusing, there is no
> > clear argumentation why a feature is in normal, big or only in huge.
> 
> Yes, my motivation was that I was looking at what exactly the
> different feature sets actually do and I found that the differences
> between "small" and "tiny" and "normal" and "big" were actually pretty
> small.
> 
> IMHO it's confusing now what the differences are. It's not really a
> "code thing" to make the code easier, but to make it easier for people
> who want to build Vim. That's also why I sent the other patch to merge
> the "tiny" and "small" builds; that it removed a bunch of #ifdefs was
> more of a side-effect than anything else.

I haven't been following this thread, so I may not understand what the
actual changes are.  That being said, I would be very hesitant to remove
tiny in favor of small; rather, I would do it the other way around.

The purpose of tiny (or at least the way it is currently being used in
Debian and Ubuntu; probably others) is to have a usable version of Vim
that is not only as small as possible on disk, but uses as little memory
as possible while running.  It is used when you are on systems or in
situations where resources are scarce.

The differences between small, normal, big, and huge are really
user preferences, while tiny has a very specific functional mandate that
cannot be replaced by a larger version.  I.e. if a user wants normal,
but it is not available, big will usually suffice.  However, when a user
needs tiny, small will not do.

...Marvin

ychin:
> Aside: organising the docs in +feature-list could also help maybe, like so:
> 
> ```
> Features in tiny version (always enabled):
> 	*+autocmd*		|:autocmd|, automatic commands.
> 	*+cursorbind*	|'cursorbind'| support
> 	.. etc..
> 
> Features in normal version (includes everything in tiny):
> 	*+arabic*		|Arabic| language support
> 	*+autochdir*		support 'autochdir' option
> 	.. etc..
> ```

I don't care much about big vs normal vs huge, but please don't split the features like this. Most users really don't care about this huge vs normal distinction, and as you pointed out this is mostly an option for the party building Vim, not for the users. There is no need to expose such grouping to the user, as it would make scanning for the feature you want much more time consuming (since I just search alphabetically).

tonymec:
Don't judge other people's preferences by your own.

I use Big Vim day-to-day, because most differences between Big and Huge are features I don't need, and the few that I do need can be enabled individually at configure time.

FWIW, here are the environments settings I use to compile my day-to-day build of Vim:

# Environment variables to be set by SOURCING this script in bash before compiling a Big build of Vim
export CONF_OPT_GUI='--enable-gui=gtk3'
export CONF_OPT_MULTIBYTE='--enable-multibyte'
export CONF_OPT_AUTOSERVE='--enable-autoservername'
export CONF_OPT_FEAT='--with-features=big'
export CONF_OPT_COMPBY='"--with-compiledby=antoine.mechelynck@gmail.com"'

Best regards,
Tony.

tonymec:
For some reason I can't modify a post after sending it.

```bash
export CONF_OPT_GUI='--enable-gui=gtk3'
export CONF_OPT_MULTIBYTE='--enable-multibyte'
export CONF_OPT_AUTOSERVE='--enable-autoservername'
export CONF_OPT_FEAT='--with-features=big'
#export CONF_ARGS2='--with-vim-name=vim-big'
export CONF_OPT_COMPBY='"--with-compiledby=antoine.mechelynck@gmail.com"'
```

brammool:
Tony wrote:

> Don't judge other people's preferences by your own.
> 
> I use Big Vim day-to-day, because most differences between Big and
> Huge are features I don't need, and the few that I do need can be
> enabled individually at configure time.

That is similar to what I was thinking: keep "normal" for including the
most useful features, and leave the less frequently used ones for
"huge".  Then you can use "normal" and add a couple of features that you
specifically need.

> FWIW, here are the environments settings I use to compile my
> day-to-day build of Vim:
> 
> # Environment variables to be set by SOURCING this script in bash before compiling a Big build of Vim
> export CONF_OPT_GUI='--enable-gui=gtk3'
> export CONF_OPT_MULTIBYTE='--enable-multibyte'
> export CONF_OPT_AUTOSERVE='--enable-autoservername'
> export CONF_OPT_FEAT='--with-features=big'
> export ***@***.***"'

You can drop "--enable-multibyte', it is always included now.

-- 
press CTRL-ALT-DEL for more information

ychin:
Seems like this  broke `+sound` on macOS builds. Please make sure to scrub existing usages of macros before `undef`'ing them. Filed #11497 to fix this.

PR #11284, opened 2022-10-04
Fix synmenu.vim

### LINKED ISSUE #11278: Error detected while compiling function SetSyn: Line 8:
###GVIM Windows V9.0 64-bit GUI with OLE support

#### This is an issue when clicking on the GUI Menu:
When Clicking on Syntax menu and then show
Then Clicking on Syntax menu again and any Syntax type, you receive an error

Error detected while compiling function SetSyn: 
Line 8:
Missing double quote: " do not replace the filetype but add whitespace on top
Missing double quote: " do not replace the filetype but add whitespace on top

arp242:
Fixes #11278

### DIFF
--- a/runtime/synmenu.vim
+++ b/runtime/synmenu.vim
@@ -14,7 +14,7 @@ def SetSyn(name: string)
     filetype = "fvwm"
   endif
   if name == "whitespace"
-    " do not replace the filetype but add whitespace on top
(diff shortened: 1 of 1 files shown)


brammool:
I had already made patch 9.0.0659 before seeing this, it's the same change.

Issue #9099, opened 2021-11-06
netrw on .vim/ directory gets vim filetype


habamax:
### Steps to reproduce

1. `vim --clean`
2. `:runtime plugin/netrwPlugin.vim`
3. `:Ex ~/.vim`

### Expected behaviour

netrw buffer with `filetype=netrw`

![image](https://user-images.githubusercontent.com/234774/140599838-5daa9190-1b1e-4e56-91f3-d6fc9476606a.png)

### Operating system

WSL debian

### Version of Vim

8.2.3582

lacygoill:
Not sure it's the right way, but this patch should fix the issue:
```diff
diff --git a/runtime/autoload/netrw.vim b/runtime/autoload/netrw.vim
index 7d1a35c1a..3a3d06608 100644
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -1789,7 +1789,7 @@ fun! s:NetrwOptionsRestore(vt)
 "  call Dfunc("s:NetrwOptionsRestore(vt<".a:vt.">) win#".winnr()." buf#".bufnr("%")."<".bufname("%")."> winnr($)=".winnr("$"))
 "  call Decho("(s:NetrwOptionsRestore) lines=".&lines)
 "  call Decho("settings buf#".bufnr("%")."<".bufname("%").">: ".((&l:ma == 0)? "no" : "")."ma ".((&l:mod == 0)? "no" : "")."mod ".((&l:bl == 0)? "no" : "")."bl ".((&l:ro == 0)? "no" : "")."ro fo=".&l:fo." a:vt=".a:vt,'~'.expand("<slnum>"))
-  if !exists("{a:vt}netrw_optionsave")
+  if !exists("{a:vt}netrw_optionsave") && &filetype != 'netrw'
 "   call Decho("case ".a:vt."netrw_optionsave : doesn't exist",'~'.expand("<slnum>"))
 "   call Decho("..doing filetype detect anyway")
    filetype detect
```

habamax:
> Not sure it's the right way, but this patch should fix the issue:

Yes, indeed.

brammool:
I'll include #11290 to fix this.

cecamp:
I've had a different solution for this on my [website](http://www.drchip.org/astronaut/vim/index.html#NETRW) since Sep 19, 2022; ie. v172g. Want to try that and tell me if it fixes your issue?

PR #11290, opened 2022-10-05
Do not run filetype detect for directories in netrw

### LINKED ISSUE #9875: Directories named *.com treated as dcl files
### Steps to reproduce

0. Setup vim with netrw.

```
% cat test.vim
set nocompatible
filetype plugin on
```

2. Create a directory `% mkdir -p google.com/example`.
3. Open said directory in vim `% vim -u test.vim google.com/example`
4. Press `-` to open the `google.com` directory
5. `echo &ft` returns `dcl`.

### Expected behaviour

I expect the directory to have a filetype of `netrw` as other directories do, not `dcl`.

I'm fine being able to override this, and have tried to do so with:

```viml
autocmd BufReadPost *.com set filetype=netrw
```

And failed.

### Version of Vim

8.2.4227

### Environment

OS: macOS 12.0.1
Terminal: Kitty
$TERM: xterm-256color
shell: zsh 5.8

### Logs and stack traces

```shell
```

teoljungberg:
Resolves #9875, #9099

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -1792,7 +1792,7 @@ fun! s:NetrwOptionsRestore(vt)
   if !exists("{a:vt}netrw_optionsave")
 "   call Decho("case ".a:vt."netrw_optionsave : doesn't exist",'~'.expand("<slnum>"))
 "   call Decho("..doing filetype detect anyway")
-   filetype detect
+   if !isdirectory(expand("%")) | filetype detect | endif
 "   call Decho("..settings buf#".bufnr("%")."<".bufname("%").">: ".((&l:ma == 0)? "no" : "")."ma ".((&l:mod == 0)? "no" : "")."mod ".((&l:bl == 0)? "no" : "")."bl ".((&l:ro == 0)? "no" : "")."ro fo=".&l:fo." a:vt=".a:vt,'~'.expand("<slnum>"))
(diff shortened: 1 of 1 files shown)


brammool:
I actually think there is never a point in "filetype detect" to try figure out the filetype for a directory.
Or we should just set filetype to "directory".
How about that?  Would that break anything?

brammool:
Actually, it appears "filetype detect" already rejects directories.  In the netrw plugin it uses a buffer to display the directory contents, but that buffer is not a directory itself.

Hmm, I have trouble reproducing the original problem.  How do you get the filetype to be something else than "netrw"?

habamax:
I guess this is related to https://github.com/vim/vim/issues/9099 which is easy to reproduce.

brammool:
OK, I'll include it.  Let me close the other issues and leave this one open for Charles.

brammool:
> I guess this is related to https://github.com/vim/vim/issues/9099
> which is easy to reproduce.

Aha, it requires using ":Explore".  I was using ":edit".

-- 
This is an airconditioned room, do not open Windows.

teoljungberg:
I’ve reproduced this using `:Explore` and entering a directory under a `.vim` or `.com` and then bubbling back up to the parent directory, that corresponds to setting `ft={vim,dcl}` respectively - while it should be `netrw`.

teoljungberg:
@brammool While resolving the merge conflicts I saw that https://github.com/vim/vim/commit/3c053a1a5ad2a3c924929e11f2b9af20a8b901e2 was merged into master yesterday.

Which includes the gist of this patch.

Should we then close this PR?

teoljungberg:
I tested out https://github.com/vim/vim/commit/3c053a1a5ad2a3c924929e11f2b9af20a8b901e2 and what’s on main and it works as expected.

Issue #1368, opened 2017-01-09
Feature Request: Line numbers in hex


Type1J:
Please add an option such as ":set numberbase=16" to allow line number bases.

vim-ml:
2017-01-09 18:39 GMT+03:00 Jay Sistar <vim-dev-github@256bit.org>:
> Please add an option such as ":set numberbase=16" to allow line number
> bases.

Use case?

justinmk:
This is another case in favor of the 'numberexpr' (or 'numberfunc') option I alluded to in:

https://github.com/vim/vim/pull/680

Zapeth:
I would also be in favor of a 'numberformat' option, not because of alignment or number base issues but because of the <space> delimiter offering not enough of a distinction between the line numbers and text sections for me.

I know, the line numbers section can be colored differently but I'd still prefer a more clear distinction, like this -> http://imgur.com/a/IfsGK

chrisbra:
this seems to be a bit niche, so closing as unplanned

Type1J:
That's unforunate. I guess I'll just make an editor that can.

justinmk:
@Type1J Neovim already supports this via https://neovim.io/doc/user/options.html#'statuscolumn'

dkaszews:
> @Type1J Neovim already supports this via https://neovim.io/doc/user/options.html#'statuscolumn'
 
`set stc=%C%s%{printf('0x%X\ ',v:num)}` works great for me, looks the same as default except for hex line numbers. 

My usecase is that I am currently working on firmware with limited debug capabilities, where traces are just CRC16 of filename, line number and a single user value, all in hex. This makes it way faster to see which line the trace came from.

Issue #10952, opened 2022-08-21
Using highlight in statusline causes window to not react to scrolling


ychin:
### Steps to reproduce

1. Install Lightline (https://github.com/itchyny/lightline.vim). I just put in under `~/.vim/pack/...` using the standard Vim package installation method. (I'm using the latest commit, b02ef0d, in case other versions don't exhibit this behavior)
2. Launch Vim.
3. Make sure mouse is enabled by doing `:set mouse=a`.
4. Enter `:put =range(1,100)` to fill some text. (Or just open any file)
5. Enter `:bot sp` to open a new split *at the bottom*. Right now there should be two splits, with the active window at the bottom.
6. Position the mouse cursor on the *top* window (which is not active). Attempt to scroll using mouse wheel.

### Expected behaviour

The window does not react to mouse wheel. However, if you click on it, or move the active window to it somehow, it will immediately be repainted with the correctly scrolled position. That's because the window was indeed scrolling, but didn't properly repaint itself.

### Version of Vim

9.0.0233

### Environment

OS: macOS 12.5.1
Tested on regular Vim on terminal and also in MacVim.

ychin:
I wanted to find a more concise repro but was failing at it. The issue here is related to 65ed136844f (v8.0.1164). Basically, Lightline is calling some `hi link` commands within statusline itself, which causes `redraw_all_later` to be called which tries to update `must_redraw` and the window's `w_redr_type`. The logic introduced in v8.0.1164 tries to set it back, but is a little buggy and not properly restoring the state. 

In particular, the bug here is that if you search for `\<curwin\> = ` within the function (`build_stl_str_hl`), you will find that inside the handling of `case STL_VIM_EXPR: // '{'`, it temporarily sets `curwin` to the current window before executing, and so if that window's `w_redr_type` gets change, the restoration code at the bottom would not be sufficient in properly restoring the state.

ychin:
The patch that seems to fix this issue is as follows, but honestly this feels a little fragile due to how much `build_stl_str_hl` seems to try to prevent the recursive draw from happening but that could really happen from anywhere.

```diff
diff --git a/src/buffer.c b/src/buffer.c
index 314ab64b2a..8c17584560 100644
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -4240,6 +4240,7 @@ build_stl_str_hl(
     stl_hlrec_T *sp;
     int		save_must_redraw = must_redraw;
[...]

Issue #3588, opened 2018-11-02
simultaneous absolute and relative line numbering


jerabaul29:
I would like to have simultaneous relative and absolute line numbering, side by side.

I found some tricks (with splits) and some plugins to do this, but all these have shortcomings. The split thing is a buggy hack, and the plugging slows vim a lot.

Any way to open for a native support of both numbers side by side?

chrisbra:
isn't that a confusing UI? possible duplicate of #3250

tonymec:
#3250 was about relative screen-line numbering. Normally I use neither 'number' nor 'relativenumber' but I played a little with them:
- When they are both off, there are of course no line numbers.
- when 'rnu' is on, all lines other than the current one have relative numbers regardless of the 'nu' setting. The current line in that case is numbered either 0 with 'nonu' or with the absolute line number with 'nu'.
- If 'nu' is on and 'rnu' is off, all lines have absolute numbers.
- In 'wrap' mode, only the first part of a wrapped line has a number, the rest don't.

IMHO it wouldn't be absurd or confusing to have absolute and relative numbering in separate columns if the user so wishes (even though I wouldn't use them), and if some developer is ready to write the corresponding code changes (I wouldn't be able to). In fact it might even be less confusing than what Vim displays now after `:setlocal nu rnu`.
What I use instead is a little function I wrote a week or two ago to move the cursor to a particular line and column (e.g. when some verification utility tells me that there is a syntax error at some line and column), as follows:
```vim
" Go to line and column
function GoTo(line, column)
  exe min([line("$"), a:line]) "| normal" a:column . "|"
endfunction
```
and OTOH, the current line number appears near the right end my status line — it appears at the same position on the default status line if 'ruler' is on but mine is customised, as follows:
```vim
if has('statusline')
  let &statusline = "%<%f %h%m%r%=%k[%{(&fenc==''?&enc:&fenc)}%{(&bomb?',BOM':'')}]"
      \ . "[U+%04B] %-14.(%l,%c%V%) %P"
endif
```
Best regards,
Tony.

brammool:
> I would like to have simultaneous relative and absolute line
> numbering, side by side.
> 
> I found some tricks (with splits) and some plugins to do this, but all
> these have shortcomings. The split thing is a buggy hack, and the
> plugging slows vim a lot.
> 
> Any way to open for a native support of both numbers side by side?

No.  And I don't think there are more than a few people who would find
this useful.

[...]

PR #2204, opened 2017-10-12
Add relativenumberformat option

jeanCarloMachado:
The request for this feature was made in this issue vim/vim/issues/1875
 and this pr vim/vim/pull/680.

PR #11303, opened 2022-10-07
[WIP] unable to customize the number column

### LINKED ISSUE #1875: Alignment of Vim relative numbers
Is there any way to align current line number to right? It looks strange by default, current line is aligning to left. Why there is not option to change it?

I tried to find a solution with `:help cul`

This thing looks ugly

vim

luukvbaal:
Problem:	Unable to customize the number column.
Solution:	Add a 'numberformat' option that follows the 'statusline' syntax to customize the number column.

Not finished with this yet but it's functional so I thought I would post the patch for feedback.
I'm not sure the way I implemented the option makes sense. I made it window-local because '(relative)number' is but this seems to be the first window-local userfunc callback option. I tried to copy what happens when setting buffer-local user callback options such as 'completefunc' instead.
Fixes https://github.com/vim/vim/issues/1875 and https://github.com/vim/vim/issues/1368.


luukvbaal:
Found https://github.com/vim/vim/issues/3588 as well. Should be possible with this as well, either by modifying the function signature to include both the relative and absolute line number, or by doing it in the userfunc itself by calling `line('.')` and adding or subtracting the relative number.

chrisbra:
Hm,
this sounds like it may impact rendering performance a bit. Is there any way to measure performance when scrolling a large file?

luukvbaal:
> Hm, this sounds like it may impact rendering performance a bit. Is there any way to measure performance when scrolling a large file?

Yeah it probably will, haven't benchmarked yet and I'm not sure how to either. Quoting justin from https://github.com/neovim/neovim/pull/4941#issuecomment-1268356232: 
> Function call overhead should not be too bad, especially for a Lua function: we already do much more than that for things like highlights/decorations. So a callback is probably the preferred approach.

brammool:
I very much doubt this is a good idea.  It is opening a can of worms.
It will be very easy to make a function that messes up the display.
It is certainly going to slow down rerdraws.
What if the function returns something else when it's called for the same line?
Next is to ask for highlighting the text.

luukvbaal:
> I very much doubt this is a good idea. It is opening a can of worms.
It will be very easy to make a function that messes up the display.

Messing up in the sense of a string that is very large that will make the number column very wide yes, other that that I'm not so sure what can go wrong? 

> It is certainly going to slow down rerdraws.

Sure, but not noticeable to me so for with the use-cases that were mentioned in the issue requests. It's a tradeoff that the user might find worth it.

> What if the function returns something else when it's called for the same line?

Not sure what you mean.

> Next is to ask for highlighting the text.

Doesn't seem necessary, we already have LineNr, CursorLineNr highlight groups.

brammool:
The returned string should undergo a check, so that any errors are caught early.
Besides being the wrong size, not fitting in the number column, there are such things as invalid character (e.g. Tab),
ending with half a double-width character and many things we haven't thought of yet.

From 'statusline' we learned that we keep tweaking when it needs to be evaluated, which depends on whatever the function is doing to make the string. What about showing "!!!" when an asynchronous callback has detected an error?

I'm sure plugin writers want to specify a different color for different lines (again, 'statusline' shows this).

This is typically something that starts as a simple function, which will very quickly grow and require a lot of maintenance.

If we would do this, then why not also cover 'foldcolumn' and 'signcolumn', so that there is more space?

luukvbaal:
> The returned string should undergo a check, so that any errors are caught early. Besides being the wrong size, not fitting in the number column, there are such things as invalid character (e.g. Tab), ending with half a double-width character and many things we haven't thought of yet.

The current implementation simply adjusts the number column width rather than constraining the returned string. Would be up to the user how wide they want it. Yeah I can confirm rendering gets broken when the string contains a Tab but at that point the user could simply unset the option until they fix their function? Do we really need to parse the userfunc extremities?

> From 'statusline' we learned that we keep tweaking when it needs to be evaluated, which depends on whatever the function is doing to make the string. What about showing "!!!" when an asynchronous callback has detected an error?

Currently I'm simply un-setting the option when an error occurs, falling back to the built-in number column. Same reasoning as above, is that not valid?

> I'm sure plugin writers want to specify a different color for different lines (again, 'statusline' shows this).

Yeah, I guess so.
 
> This is typically something that starts as a simple function, which will very quickly grow and require a lot of maintenance.

Yeah I can see that happening if we want to take the responsibility of catching any and all invalid strings. I thought checking that the return type of the callback is per spec would be sufficient.

> If we would do this, then why not also cover 'foldcolumn' and 'signcolumn', so that there is more space?

Not sure what you mean by this, again I took the approach of making the number column as wide as the largest string the callback returns.

arp242:
I did a bit of rudimentary performance testing, and it seems the overhead isn't a big problem; when using Vim9Script it's even better than not using a function at all.

Of course, people can do slow/dumb stuff in 'numberfunc', but that's already the case for many func and expr options and autocmds like CursorMoved.

Some comments:

- I'm not sure if the function signature is the best; I'd prefer returning a string which is expected to have a length to 'numberwidth' (can automatically pad/trim if it's shorter/longer).
- It's not clear to me how this works exactly when both 'number' and 'relativenumber' is set. Right now it only seems to work for 'relativenumber'
- I've seen people ask about alignment and such on a number of occasions; and it seems clear there's a demand for solving the problem. There was a 'numberformat' patch too: https://github.com/vim/vim/pull/2204 as an alternative (but this had some issues, working only for 'relativenumber' being the main one). A 'numerformat' similar to 'statusline' alleviates some of the "dumb stuff" performance concerns, although you can still use %{ .. } expressions in that, and do silly slow things.

---

Test results:

    % for i in {1..4}; time ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c
    ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c  1.94s user 0.60s system 98% cpu 2.592 total
    ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c  1.98s user 0.60s system 97% cpu 2.647 total
    ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c  2.24s user 0.22s system 99% cpu 2.466 total
    ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c  1.99s user 0.25s system 98% cpu 2.272 total

~/num script, which tests just setting 'relativenumber' (differences are similar when setting both 'number' and 'relativenumber'):

    " for i in {1..4}; time ./vim --clean +"let g:test=$i" +'so ~/num' buffer.c

    if g:test == 1
        " Run without any numbers
    elseif g:test == 2
        set relativenumber
    elseif g:test == 3
        " VimScript
        func! s:Nuf(lnum, cursor)
            if (a:cursor)
                return [string(a:lnum), v:false]
            else
                return [string(a:lnum), v:true]
            endif
        endfunc

        set relativenumber
        set numberfunc=<SID>Nuf
    elseif g:test == 4
        " Vim9Script
        " Not supported yet: tuple<a: string, b: bool>
        def <SID>Nuf2(lnum: number, cursor: bool): list<any>
            if (cursor)
                return [lnum->string(), v:false]
            else
                return [lnum->string(), v:true]
            endif
        enddef
        defcompile

        set relativenumber
        set numberfunc=<SID>Nuf2
    endif

    normal! gg
    while line('.') != line('$')
        exe "normal! \<C-d>"
        redraw!
    endwhile
    quit

luukvbaal:
> I did a bit of rudimentary performance testing, and it seems the overhead isn't a big problem; when using Vim9Script it's even better than not using a function at all.

Thanks for that!

> I'm not sure if the function signature is the best; I'd prefer returning a string which is expected to have a length to 'numberwidth' (can automatically pad/trim if it's shorter/longer).

That's what I went for first, but I didn't see a clear approach for alignment in the user func. I think the user function needs to know the width to pad to before hand but the width is dependent on the user func itself. Relying on printf() format expression `(-)*` padding, after the callback can handle this for us but to do this we need to know where the user wants to align the string. Maybe I took the wrong approach though. Alternatively we could parse the string and assign the first character "L/R" for alignment.

> It's not clear to me how this works exactly when both 'number' and 'relativenumber' is set. Right now it only seems to work for 'relativenumber'.

Should work for 'norelativenumber' as well, just leave out the `if (a:cursorline)` clause then, that's only to be able to distinguish the cursorline for 'relativenumber'. I thought user might want to print "=>" instead of "0" for example:
```vim
function! F(lnum, cursorline)
	if (a:cursorline)
		return [string("=>"), v:true]
	else
		return [string(a:lnum), v:true]
	end
endfunc
```

> A 'numerformat' similar to 'statusline' alleviates some of the "dumb stuff" performance concerns, although you can still use %{ .. } expressions in that, and do silly slow things.

Yeah I implemented it with a status-line expression first actually. But you quickly end up needing to rely on %{} and passing along the line number to be printed in a (window-local) variable. At that point you might as well call a user function. Relevant discussion here: https://github.com/neovim/neovim/pull/4941#issuecomment-1267671735.

brammool:
> > The returned string should undergo a check, so that any errors are
> > caught early. Besides being the wrong size, not fitting in the
> > number column, there are such things as invalid character (e.g.
> > Tab), ending with half a double-width character and many things we
> > haven't thought of yet.
> 
> The current implementation simply adjusts the number column width
> rather than constraining the returned string. Would be up to the user
> how wide they want it. Yeah I can confirm rendering gets broken when
> the string contains a Tab but at that point the user could simply
> unset the option until they fix their function? Do we really need to
> parse the userfunc extremities?

How this usually goes is that the user includes some cool plugin, and
it turns out there is a context that the plugin author didn't think of,
and eventually the Vim code needs to be fixed.  Been there, done that...

> > From 'statusline' we learned that we keep tweaking when it needs to
> > be evaluated, which depends on whatever the function is doing to
> > make the string. What about showing "!!!" when an asynchronous
> > callback has detected an error?
> 
> Currently I'm simply un-setting the option when an error occurs,
> falling back to the built-in number column. Same reasoning as above,
> is that not valid?

That makes sense.  But there should be enough information about why the
option was reset.  There are some examples for other options (although
not very good examples).  It does require a proper check that the
result is OK.

> > I'm sure plugin writers want to specify a different color for
> > different lines (again, 'statusline' shows this).
> 
> Yeah, I guess so.
>  
> > This is typically something that starts as a simple function, which
> > will very quickly grow and require a lot of maintenance.
> 
> Yeah I can see that happening if we want to take the responsibility of
> catching any and all invalid strings. I thought checking that the
> return type of the callback is per spec would be sufficient.
> 
> > If we would do this, then why not also cover 'foldcolumn' and
> > 'signcolumn', so that there is more space?
> 
> Not sure what you mean by this, again I took the approach of making
> the number column as wide as the largest string the callback returns.

I do not think the number column width should change depending on what
the function returns, since jumping through the file can have weird
effects.  Rather 'numberwidth' should be set once and the function
should use it.

But again, I'm not sure we want to go this way at all.

-- 
Just think of all the things we haven't thought of yet.

luukvbaal:
I see. Well let me know if you decide in favor of this feature then I'll look into highlighting...
The way I see it, the fact that previous multiple PRs were closed because they were too generic, attests to the fact that a user function is warranted. But I understand your hesitancy to include the feature on grounds of performance/maintenance.

luukvbaal:
> I do not think the number column width should change depending on what the function returns, since jumping through the file can have weird effects. Rather 'numberwidth' should be set once and the function should use it.

I agree that it shouldn't change dynamically BTW. What I meant, and how I implemented it, is making a best guess at the maximum width the function will return by checking the length of the string that the function returns in number_width() and setting the 'numberwidth' cache accordingly. 

This should work in most cases. I cannot imagine why a user would want to return a longer string for a specific case that is not covered by the check in number_width(). If they do, I would call that abusing the 'numberfunc' function and indeed it could result in not fitting in the number column. Not a problem in my eyes.

ychin:
If we do want to add this, I will add that I also prefer a `statusline` approach more than a user func, for multiple reasons:

1. It's consistent with how `tabline` and `statusline` work. Consistency is very useful when learning a new API.
2. It's much richer and allows for certain necessary features "for free", like left/right alignment, highlight groups (e.g. inevitably you are going to find people wanting to highlight odd/even numbers differently), and other more complicated stuff.
3. The common use case (of using builtin atoms of the form `%X`) should be more efficient than call a function. If the user wants to, you can still call expressions using `%{` / `%{%`. It's functionally a superset of just a function.
4. It's easier to author simple stuff, but possible to author complicated features using `%{my_func()}`. If you use a function-based approach, I find that kind of annoying if you just want to change small things. Not everyone wants to use a lightline/airline-esque plugin just to draw numbers. We can include useful atoms like "localized numbers" (meaning "1.234" in Europe and "1,234" in US) for example, and that would already cover a large use case (in addition to builtin ability to do left/right alignment in `statusline`). You can still call function calls but it doesn't need to be the default way of customization.

Seems like the only reason to go for a function-based approach is wanting to pass arguments in, but `statusline` already set a precedent of setting a global var temporarily (see `g:statusline_winid`/ `g:actual_curbuf` / etc). It's not the absolutely cleanest solution in a vacuum but I do think the consistency is going to win over a new approach, not to mention the other reasons I mentioned.

---

As for the function safeguarding, I can see part of where Bram is coming from. The moment you accept user functions, you have to now guard against all kinds of crazy stuff. It's not responsible from the software to just say "you can shoot yourself in the foot and result in a weird / inconsistent Vim state" especially when it's not clear what parts of functionalities are shooting yourself in the foot and what parts of them are simply clever applications. Give people a tool and they will eventually use all the functionality exposed.

Here's an example bug that I just filed 1 month ago: #10952. In this case, there are issues with statusline authors (in this case, lightline) calling functionality `hi link` which triggers a redraw while already in a redraw. Vim already had some safeguards around this infinite recursion problem, but the safeguards led to another issue of an inconsistent state, so more fixes needed to be piled on top of it to protect that. It's not just about a slightly wrong highlighting outcome, but more that functionality like this could end up crashing / locking / resulting in an inconsistent state that is the issue. It's not always possible for plugin authors to know that such functionality are not allowed, unless Vim has an explicit whitelist / blacklist capability (again, more work, and not included in this PR) to block those things off properly, or we just make sure to check for all those potentially problematic aspects.

Using a `statusline` approach does allow us to start introducing this functionality *without* the `%{` support at first to ease it in, but maybe that's not a great argument because you will quickly see people asking for it that you may as well get it right in the beginning.

ychin:
I do think it's also worth thinking about how this could expand to foldcolumns and signs per mentioned above. We don't have to support it now, but we should at least have an idea of how that could be incorporated if we want to do it say 1 year later and we don't overly constrain ourselves here. One thing I like about the `statusline` approach is that when people use `tabline`, they don't just use it for tabs, but potentially other info as well. Let's say people want to customize how folds are rendered (since the default fold drawing is kind of iffy IMO), would this work for that as well?

luukvbaal:
Yes I mentioned some of these points over in the neovim issue tracker, thanks for summarizing them here. You might have swayed back over to a status line expression for this feature although I must say I don't really see how you would end up with a "lightline/airline-esque plugin". As you say the feature is meant to just to draw numbers, not much more. But I guess you're right that if about giving people a tool, somebody will at some point. In that sense it would make sense to be conformant with similar options and opt for a statusline expression. Like I said, I implemented it with build_stl_string_hl()(albeit less fleshed out) first so if Bram is still open to the feature I could change this PR to that instead.

luukvbaal:
@brammool I fleshed out the 'statusline'-expression approach for this feature, I have updated the branch. Now I'm wondering if there is merit in further discussing the feature here or do you consider the consider the concerns you mentioned blocking? If so, I will go ahead and start porting to neovim as it might be included there.

brammool:
> @brammool I fleshed out the 'statusline'-expression approach for this
> feature, I have updated the branch. Now I'm wondering if there is
> merit in further discussing the feature here or do you consider the
> consider the concerns you mentioned blocking? If so, I will go ahead
> and start porting to neovim as it might be included there.

I'm afraid I find this too complex and most likely requiring quite a bit
of maintenance in the future, without the expected usage justifying it.
I also think that it still might not cover everything (using the sign
and fold columns was mentioned).

In other words: I think there are more useful things to work on.

-- 
I'm sure that I asked CBuilder to do a "full" install.  Looks like I got
a "fool" install, instead.		Charles E Campbell, Jr, PhD

PR #11305, opened 2022-10-07
Add Path of Exile item filter syntax

ObserverOfTime:
**Sources:**
* https://www.pathofexile.com/item-filter/about
* https://pathofexile.fandom.com/wiki/Guide:Item_filter#Syntax

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1391,6 +1391,9 @@ au BufNewFile,BufRead *.dpr,*.lpr			setf pascal
 " Free Pascal makefile definition file
 au BufNewFile,BufRead *.fpc				setf fpcmake
 
+" Path of Exile item filter
+au BufNewFile,BufRead *.filter				setf poefilter
+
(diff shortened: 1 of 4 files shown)


runtime/filetype.vim
```diff
@@ -1391,6 +1391,9 @@ au BufNewFile,BufRead *.dpr,*.lpr			setf pascal
 " Free Pascal makefile definition file
 au BufNewFile,BufRead *.fpc				setf fpcmake
 
+" Path of Exile item filter
+au BufNewFile,BufRead *.filter				setf poefilter
```
arp242, line 1395: .filter is a very generic name, and I think it will probably erroneously apply to other files too. Looking at your links, I think `Show` or `Hide` should appear as the first directive? I think it would be better to also check for that.

ObserverOfTime: I don't see a reason to add a heuristic when no other supported language uses that extension.

PR #11312, opened 2022-10-08
Add :confirm support for terminal buffers

ychin:
Previously terminal windows never really worked well with `:confirm` commands as most places still treated it like a buffer with modification (and the subsequent "Yes" confirmation would subsequently fail as Vim doesn't know how to "save" a terminal). The only place that really worked was `:confirm qa`.

Fix up the code so most places (at least the ones that I could find) that take `:confirm` also handles terminal buffers properly.

The list of commands modified:

- `check_changed()`, used for :q/:only/:enew/:next etc.
- `:bdelete` / `bwipe`
- `:close`
- `:buffer N`

(For `:buffer` and `:close` I had to manually call `free_terminal()`. For `:buffer` it's because the fallthrough logic is to just hide the buffer while the :confirm behavior is to kill the terminal, and for `:close` that's because I find that otherwise it leaves a straggling buffer that doesn't do anything, see below design question (3)).

Also, updated File.Close menu to be able to handle terminals as well.

Also added tests for both the :confirm and [!] behaviors for the above commands.

Design questions / notes:

1. [x] Terminal conf dialog boxes used to say yes/no/cancel, but I changed it to yes/no, because "no" and "cancel" used to do the same thing if you reason through the code. That said, I'm debating whether a yes/cancel makes more sense as usually the decision is to either "kill the terminal", or "cancel whatever operation you are doing". I think yes/cancel makes more sense personally, so if we want to do this I can update it (I just need to update all the localization po files as well).

2. [x] When writing tests, I noticed that `:q!` on a terminal has a quirk where it would leave a straggler buffer that is empty, but not deleted. This seems different from the other ways a terminal gets killed. If you just exit the terminal normally (e.g. type "exit" in the shell), it will remove itself from the buffer list. I think this is a bug. For `:confirm close` I found a similar issue, but I worked around it by calling `free_terminal()`.
    - (Edit: This and other related quirks will be fixed in https://github.com/vim/vim/pull/11323)
4. [x] `term_start` allows you to set a `term_kill` option on a terminal. Right now, if you set it, and then do `:qa`, Vim will just quit without warning you about the terminal. However, just doing regular `:q` *will* warn you about the terminal. I'm not really sure which is the correct behavior so I kept it as it is, but it seems slightly inconsistent. It's not directly related to adding `:confirm` support anyway.
    - (Edit: Filed #11325 for tracking)

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -49,6 +49,7 @@ static int	append_arg_number(win_T *wp, char_u *buf, int buflen, int add_file);
 static void	free_buffer(buf_T *);
 static void	free_buffer_stuff(buf_T *buf, int free_options);
 static int	bt_nofileread(buf_T *buf);
+static void	no_write_message_buf(buf_T *buf);
 
 #ifdef UNIX
 # define dev_T dev_t
@@ -1367,21 +1368,30 @@ do_buffer_ext(
 #if defined(FEAT_GUI_DIALOG) || defined(FEAT_CON_DIALOG)
 	    if ((p_confirm || (cmdmod.cmod_flags & CMOD_CONFIRM)) && p_write)
 	    {
-		dialog_changed(buf, FALSE);
-		if (!bufref_valid(&bufref))
-		    // Autocommand deleted buffer, oops!  It's not changed
-		    // now.
-		    return FAIL;
-		// If it's still changed fail silently, the dialog already
-		// mentioned why it fails.
-		if (bufIsChanged(buf))
-		    return FAIL;
+# ifdef FEAT_TERMINAL
+		if (term_job_running(buf->b_term))
+		{
+		    if (term_confirm_stop(buf) == FAIL)
+			return FAIL;
+		}
+		else
+# endif
+		{
+		    dialog_changed(buf, FALSE);
+		    if (!bufref_valid(&bufref))
+			// Autocommand deleted buffer, oops!  It's not changed
+			// now.
+			return FAIL;
+		    // If it's still changed fail silently, the dialog already
+		    // mentioned why it fails.
+		    if (bufIsChanged(buf))
+			return FAIL;
+		}
 	    }
 	    else
 #endif
 	    {
-		semsg(_(e_no_write_since_last_change_for_buffer_nr_add_bang_to_override),
-								 buf->b_fnum);
+		no_write_message_buf(buf);
 		return FAIL;
 	    }
 	}
@@ -1552,15 +1562,34 @@ do_buffer_ext(
 #if defined(FEAT_GUI_DIALOG) || defined(FEAT_CON_DIALOG)
 	if ((p_confirm || (cmdmod.cmod_flags & CMOD_CONFIRM)) && p_write)
 	{
-	    bufref_T bufref;
+# ifdef FEAT_TERMINAL
+	    if (term_job_running(curbuf->b_term))
+	    {
+		if (term_confirm_stop(curbuf) == FAIL)
+		    return FAIL;
+		// Manually kill the terminal here because this command will
+		// hide it otherwise.
+		free_terminal(curbuf);
+	    }
+	    else
+# endif
+	    {
+		bufref_T bufref;
 
-	    set_bufref(&bufref, buf);
-	    dialog_changed(curbuf, FALSE);
-	    if (!bufref_valid(&bufref))
-		// Autocommand deleted buffer, oops!
-		return FAIL;
+		set_bufref(&bufref, buf);
+		dialog_changed(curbuf, FALSE);
+		if (!bufref_valid(&bufref))
+		    // Autocommand deleted buffer, oops!
+		    return FAIL;
+
+		if (bufIsChanged(curbuf))
+		{
+		    no_write_message();
+		    return FAIL;
+		}
+	    }
 	}
-	if (bufIsChanged(curbuf))
+	else
 #endif
 	{
 	    no_write_message();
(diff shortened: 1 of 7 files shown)


src/buffer.c
```diff
@@ -5751,8 +5792,8 @@ bt_nofile(buf_T *buf)
 #endif
 
 /*
- * Return TRUE if "buf" is a "nowrite", "nofile", "terminal" or "prompt"
- * buffer.
+ * Return TRUE if "buf" is a "nowrite", "nofile", "terminal", "prompt", or
+ * "popup" buffer.
```
ychin, line 5796: Note: This function still works, but I think the fact that it worked for "popup" was completely coincidental, so just updated comment to make it intentional.

src/ex_cmds2.c
```diff
@@ -198,6 +205,7 @@ dialog_changed(
 			|| (cmdmod.cmod_flags & CMOD_BROWSE)
 #endif
 			)
+		    && !bt_dontwrite(buf2)
```
ychin, line 208: This fixes an issue when you do `:confirm qall` with a live terminal, and it complains about not having autocmds associated with it. It still works but this makes sure we don't hit that code path

ychin:
This commit should be good to merge as-is, but I'm curious about the 3 design questions I had. In particular for question 1, I kind of want to change the prompt to yes/cancel, which I think makes much more sense to the user (you either want to kill the terminal, or just cancel out of the whole thing. "no" is a bad choice to use here). If we want to do that I can make the modification, but I just wanted to wait to confirm, before I go and update all the localization files.

brammool:
> Previously terminal windows never really worked well with `:confirm`
> commands as most places still treated it like a buffer with
> modification (and the subsequent "Yes" confirmation would subsequently
> fail as Vim doesn't know how to "save" a terminal). The only place
> that really worked was `:confirm qa`.
> 
> Fix up the code so most places (at least the ones that I could find)
> that take `:confirm` also handles terminal buffers properly.
> 
> The list of commands modified:
> 
> - `check_changed()`, used for :q/:only/:enew/:next etc.
> - `:bdelete`
> - `:close`
> - `:buffer`
> 
> (For `:buffer` and `:close` I had to manually call `free_terminal()`.
> For `:buffer` it's because the fallthrough logic is to just hide the
> buffer while the :confirm behavior is to kill the terminal, and for
> `:close` that's because I find that otherwise it leaves a straggling
> buffer that doesn't do anything, see below design question (3)).
> 
> Also, updated File.Close menu to be able to handle terminals as well.
> 
> Also added tests for both the :confirm and [!] behaviors for the above commands.
> 
> Design questions / notes:
> 
> 1. [ ] Terminal conf dialog boxes used to say yes/no/cancel, but I
> changed it to yes/no, because "no" and "cancel" used to do the same
> thing if you reason through the code. That said, I'm debating whether
> a yes/cancel makes more sense as usually the decision is to either
> "kill the terminal", or "cancel whatever operation you are doing". I
> think yes/cancel makes more sense personally, so if we want to do this
> I can update it (I just need to update all the localization po files
> as well).

It depends on the question.  If it is "close the terminal" then yes/no
is a better choice.  Only for "save changes?" it makes sense to have
yes/no/cancel.  Since you can not save changes and still close a window.
Thus there are really three different choices.

For "kill the terminal" I think yes/no is clearer than "yes/cancel".
It looks like that is what the code does.

> 2. [ ] When writing tests, I noticed that `:q!` on a terminal has a
> quirk where it would leave a straggler buffer that is empty, but not
> deleted. This seems different from the other ways a terminal gets
> killed. If you just exit the terminal normally (e.g. type "exit" in
> the shell), it will remove itself from the buffer list. I think this
> is a bug. For `:confirm close` I found a similar issue, but I worked
> around it by calling `free_terminal()`.

Leaving an empty buffer behind does not appear to serve a purpose.  It
doesn't have the scrollback, right?  It does sound like a bug.

> 3. [ ] `term_start` allows you to set a `term_kill` option on a
> terminal. Right now, if you set it, and then do `:qa`, Vim will just
> quit without warning you about the terminal. However, just doing
> regular `:q` *will* warn you about the terminal. I'm not really sure
> which is the correct behavior so I kept it as it is, but it seems
> slightly inconsistent. It's not directly related to adding `:confirm`
> support anyway.

If term_kill is set then stopping the terminal does not involve a
warning, the job is killed. When it is empty then a warning should be
given.  I'm not sure why using ":q" would be different from ":qa" (both
without exclamation mark).  That might be a bug.

-- 
"Shoot for the moon. Even if you miss, you'll land among the stars."

ychin:
> It depends on the question. If it is "close the terminal" then yes/no is a better choice. Only for "save changes?" it makes sense to have yes/no/cancel. Since you can not save changes and still close a window. Thus there are really three different choices. For "kill the terminal" I think yes/no is clearer than "yes/cancel". It looks like that is what the code does.

Yes, I was specifically talking about terminal closing. For files of course we still want to have three choices (yes/no/cancel). But sure I can keep yes/no for terminals.

> Leaving an empty buffer behind does not appear to serve a purpose. It doesn't have the scrollback, right? It does sound like a bug.

That's correct. The stale buffer is simply completely empty. I think a `free_terminal()` call before closing the window (that's what I'm doing in this PR for `:conf close` and `:conf buffer`) would actually fix this, but I'm not sure if there are other gotchas.

Actually I find a flipside of the bug. If I do `:term ++noclose`, then finish the job by typing `exit` in the shell, then `:hide`, the finished terminal scrollback buffer actually gets removed from the buffer list, which seems like a bug since `:hide` should not be a destructive action. `:hide` on an active terminal does still work though.

> If term_kill is set then stopping the terminal does not involve a warning, the job is killed. When it is empty then a warning should be given. I'm not sure why using ":q" would be different from ":qa" (both without exclamation mark). That might be a bug.

I think the "`term_kill` => no warning" behavior is currently only for `:qa`. All the other commands that I could find (e.g. `:q`, `:close`, `:bdelete`, `:buffer`) do not obey that behavior. (For reference the behavior is enforced in `term_try_stop_job`, which only gets called in `:qa`)

---

I wonder if the above bugs should be fixed separately from this pull request. Otherwise I can try to locate them. Other than that my PR should be in a good-to-merge state.

brammool:
Let's not make this PR more complicated.  If you can tackle those other problems in another PR that would be great.

PR #11321, opened 2022-10-09
Clarify docs on g:python_indent

arp242:
It's not clear that you need to define the dictionary first; just copy/pasting the examples will be an error.

Led to some confusion: https://vi.stackexchange.com/q/38823/51


brammool:
I'll include it and reword it a bit.

PR #11322, opened 2022-10-09
Make autoconf 2.71 work

### LINKED ISSUE #10528: Autoconf 2.69 is required
### Steps to reproduce

run "make autoconf".
This works fine with autoconf 2.69.
With autoconf 2.71 a lot of "obsolete" warnings are produced.
Running the suggested "autoupdate" does not work.

### Expected behaviour

Most recent version of autoconf 2.71 works without problems.

I tried manually updating configure.ac, but got stuck.  I guess quoting works differently.
I would appreciate someone looking into this.

### Version of Vim

8.2.5057

### Environment

Linux, Autoconf 2.71

### Logs and stack traces

_No response_

arp242:
Update configure.ac to work with autoconf 2.71; changes:

- Remove AC_HEADER_STDC; documented as "All hosted environments that are still of interest for portable code provide all of the headers specified in ISO C90".

- Remove AC_HEADER_TIME; documented as "‘time.h’ is universally present, and the systems on which ‘sys/time.h’ conflicted with ‘time.h’ are obsolete."

- Remove AC_ISC_POSIX; documented as "Sun dropped support for the obsolete INTERACTIVE Systems Corporation Unix on 2006-07-23.  New programs need not use this macro."

- AC_PROGRAM_EGREP is now AC_PROG_EGREP.

- AC_PROG_CC_C99 is now AC_PROG_CC; add a compatibility line for older versions of autoconf.

I ran "make autoconf" with 2.71 and 2.69, and both seem to work now.

Also, I see no more references to "configure.lineno" in current versions of autoconf, and "When the value of $MAKE contains a path, configure can't handle this" from todo.txt doesn't seem to be an issue any more.

Fixes #10528


brammool:
Thanks for looking into this.  I started on it but got stalled.

> Update configure.ac to work with autoconf 2.71; changes:
> 
> - Remove AC_HEADER_STDC; documented as "All hosted environments that
> are still of interest for portable code provide all of the headers
> specified in ISO C90".
> 
> - Remove AC_HEADER_TIME; documented as "‘time.h’ is universally
> present, and the systems on which ‘sys/time.h’ conflicted with
> ‘time.h’ are obsolete."

I am a bit worried about this.  I can understand that lots of programs
will drop support for old systems.  However, autoconf is exactly the
tool that is supposed to work *everywhere*.  I do not understand how
autoconf can call something "obsolete".  Does this mean there is not a
single running system left in the world?  The problem of conflicting
time header files was real and we had Vim users complain, at least at
some point in time this check was required.  How do we know it's 100%
gone, no user is in this situation?

> - Remove AC_ISC_POSIX; documented as "Sun dropped support for the
> obsolete INTERACTIVE Systems Corporation Unix on 2006-07-23.  New
> programs need not use this macro."

But old Sun systems are around and cannot be updated to a newer OS.  One
might argue that these systems belong in a museum, but there are people
working hard to keep these old systems up an running.  It's
disappointing to intentionally drop support, without a clear reason why
this is needed.

I suppose old systems would need to use an old Vim version...
However, it's quite unclear what systems are affected, and how many
users this affects.  Installing your favorite editor is the first thing
to do on any system you boot up.

-- 
MORTICIAN:    Bring out your dead!
              [clang]
              Bring out your dead!
              [clang]
              Bring out your dead!
CUSTOMER:     Here's one -- nine pence.
DEAD PERSON:  I'm not dead!

tonymec:
Maybe it's time for us to do what Mozilla did some years ago? They found that their autoconf scripts were incompatible with the newest versions of autoconf and decided to require autoconf 2.13. The result is that, for instance, on my openSUSE system there is a package "autoconf213", required to compile Mozilla (Firefox, Thunderbird, SeaMonkey), which can be installed together with "autoconf" (currently autoconf 2.71). If your distro doesn't provide autoconf 2.13 I think I remember that it can be downloaded, either from the Mozilla site or from somewhere linked therefrom, for all platforms where Firefox can be compiled.

I don't know what the differences are between these two versions of autoconf so it's just an idea I'm throwing against the wall to see if it sticks.

Best regards,
Tony.

tonymec:
P.S. The executable names from the autoconf213 packages have "-2.13" postposed: autoconf-2.13, autoheader-2.13, autoreconf-2.13, autoscan-2.13, autoupdate-2.13 and ifnames-2.13. The runtime files from that package are in a separate directory /usr/share/autoconf-2.13/.

Best regards,
Tony.

arp242:
I'm going by what the autoconf documentation tells me; since the autoconf people tend to take compatibility fairly serious so I'm *assuming* when they say something is "no longer needed" that really no one actually needs it any more.

AC_HEADER_STDC and AC_HEADER_TIME were already declared as "obsolete" in 2013 (https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=86c213d0), and AC_ISC_POSIX in 2007 (https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=6add6e92), so they're not very new changes either.

But it's really hard for me to judge what is or isn't needed any more and what systems people are and aren't using. There isn't a whole lot of context on the "why" of these changes in autoconf; for example for AC_ISC_POSIX, adding AC_SEARCH_LIBS([strerror], [cposix]) should have the same effect, but the way I read the docs I *think* you don't need it for older Solaris systems, but that it's a compatibility thing Solaris had with ISC Unix (sold to Sun in 1991).

Personally I'd take the "see if people complain, modify if needed" approach here.

> They found that their autoconf scripts were incompatible with the newest versions of autoconf and decided to require autoconf 2.13.

There are many errors if I run autoconf 2.13, mostly because AC_DEFINE, AC_MSG_*, AC_*_IFELSE are undefined, as well as a few others. It would require quite a bit of work to use 2.13.

Firefox also vendors autoconf; we could do the same with autoconf 2.69 (or any version) and then use that; no need to choose 2.13 specifically.

brammool:
> I'm going by what the autoconf documentation tells me; since the
> autoconf people tend to take compatibility fairly serious so I'm
> *assuming* when they say something is "no longer needed" that really
> no one actually needs it any more.
> 
> AC_HEADER_STDC and AC_HEADER_TIME were already declared as "obsolete" in 2013 (https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=86c213d0), and AC_ISC_POSIX in 2007 (https://git.savannah.gnu.org/cgit/autoconf.git/commit/?id=6add6e92), so they're not very new changes either.
> 
> But it's really hard for me to judge what is or isn't needed any more
> and what systems people are and aren't using. There isn't a whole lot
> of context on the "why" of these changes in autoconf; for example for
> AC_ISC_POSIX, adding AC_SEARCH_LIBS([strerror], [cposix]) should have
> the same effect, but the way I read the docs I *think* you don't need
> it for older Solaris systems, but that it's a compatibility thing
> Solaris had with ISC Unix (sold to Sun in 1991).
> 
> Personally I'd take the "see if people complain, modify if needed"
> approach here.

The big question is if the people mainaining autoconf intend to support
very old systems.  Perhaps they argue that they don't have access to
these systems, thus can't test it there, and have given up on them.

> > They found that their autoconf scripts were incompatible with the
> > newest versions of autoconf and decided to require autoconf 2.13.
> 
> There are many errors if I run autoconf 2.13, mostly because
> AC_DEFINE, AC_MSG_*, AC_*_IFELSE are undefined, as well as a few
> others. It would require quite a bit of work to use 2.13.
> 
> Firefox also vendors autoconf; we could do the same with autoconf 2.69
> (or any version) and then use that; no need to choose 2.13
> specifically.

2.13 seems very old.  We have been using 2.69 for quite a while and I
am not aware of real problems.  Sticking to it is an option, but the
main problem is that changes made for new systems will be missing.
I doubt they will backport those changes to older autoconf versions.

The changes I made to avoid most warnings from 2.71 are messy.  They
make the script much harder to read, because there is more nesting.  And
it is very easy to misplace a comma or other character, without getting
a useful error.  Or not get any error and end up with a broken shell
script.  The script language used (m4) is very bad in that way, removing
a comma may give the opposite result.

For me this (using more nesting) is a sign that the maintainers don't
have the right mindset.  It reminds me of gcc versions that were
advertised as having a much better optimizer, but turned out to have
several obscure bugs.

Anyway, I agree that it's probably best to switch to autoconf 2.71 and
very carefully make it work.  There are some comments here and there
that help deciding.

-- 
MORTICIAN:    What?
CUSTOMER:     Nothing -- here's your nine pence.
DEAD PERSON:  I'm not dead!
MORTICIAN:    Here -- he says he's not dead!
CUSTOMER:     Yes, he is.
DEAD PERSON:  I'm not!

arp242:
> 2.13 seems very old.  We have been using 2.69 for quite a while and I am not aware of real problems.  Sticking to it is an option, but the main problem is that changes made for new systems will be missing.

Firefox seems to get by with 2.13, but I don't know how much headaches it's causing them. Modern (i.e. from this century) systems are much easier to support anyway, because we have things like POSIX and whatnot and it's not the mess from the "Unix Wars" days. We can *probably* get by with 2.69 for quite a few more years (2.69 it was released in 2012, so it's already ten years old; there was a long gap before the 2.70 release), but I can't really predict the future.

Personally, I'd be in favour of merging this as-is, and if someone complains and the fix is non-trivial just revert it and decide what we want to do then. It would only be a minor inconvenience for whomever can't build the latest master of Vim for a day. There a decent chance no one will complain, and doing a lot of annoying work in autoconf/M4 for no benefit is no one's idea of fun. And *if* someone complains we also have a better idea of what *exactly* breaks, and we have an idea what to fix.

ilya-bobyr:
I've prepared a 3 step branch that does pretty much the same as this PR.
Realized this work was already pending, only after I created the changes.
First part: https://github.com/vim/vim/pull/12888

Doing it in smaller steps might provide a better opportunity to fix any possible breakages.
Though, if @arp242 is still on it and the consensus is that a single change is fine, I'll close my PR.

arp242:
I rebased on master – can go with Illia's PR and close this too, I'd be fine with that. That said, quoting myself from the last comment:

> I'd be in favour of merging this as-is, and if someone complains and the fix is non-trivial just revert it and decide what we want to do then. It would only be a minor inconvenience for whomever can't build the latest master of Vim for a day. There a decent chance no one will complain, and doing a lot of annoying work in autoconf/M4 for no benefit is no one's idea of fun. And if someone complains we also have a better idea of what exactly breaks, and we have an idea what to fix.

My general takeaway from looking in to this last year is that with autoconf you can never be quite sure what will work where and with which version; the entire thing is hilariously under-documented and opaque and never seemed to have moved beyond 1987-esque development practices 🤷

ilya-bobyr:
*Can not leave a comment for a line that was not modified*

`autoupdate` from 2.71 also suggested removal of `AC_DECL_SYS_SIGLIST`.
It is actually already commented out in `configure.ac`:
https://github.com/vim/vim/blob/544b209a2d4b10e0a4874c925036def30d33d708/src/configure.ac#L3607-L3609

`AC_DECL_SYS_SIGLIST` was deprecated in autoconf 2.69: https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html#index-AC_005fDECL_005fSYS_005fSIGLIST-2014

Seems like this line was added as part of work that was never finished:

https://github.com/vim/vim/blob/544b209a2d4b10e0a4874c925036def30d33d708/src/os_unix.c#L253-L265

I can, of course, send a separate PR for this particular part.
Though, maybe you want to include it here, as `autoupdate` 2.71 removed the macro for me.

arp242:
I removed the commented out line, but left the bit in `vim/src/os_unix.c` in place – I didn't look who added that and when, and it could be useful for someone.

chrisbra:
so, how much of this do we need after f39842f148c9a8112432644b7f4a0129315de771 and f39842f148c9a8112432644b7f4a0129315de771 ?

ilya-bobyr:
These 3 PRs should cover all the changes currently present in this PR:
* #12888 (merged as https://github.com/vim/vim/commit/f39842f148c9a8112432644b7f4a0129315de771)
* #12949 (merged as https://github.com/vim/vim/commit/0e82b59df3fc12dc006e45e17222bcbb0462903f)
* #12958

PR #11323, opened 2022-10-09
Fix terminal buffer closure behaviors

ychin:
Fix a couple bugs and inconsistencies in terminal window management and buffer deletion:

1. Using `:q!` on a live terminal would previously leave an empty buffer behind. Now it would completely remove the buffer as expected.
2. Using `:bunload!` on a live terminal would previously not do anything with the terminal job and just hide the window. Now it would do as the name suggests and actually kill the job and unload the buffer.
3. Using `:hide` on a finished terminal window would previously unload the buffer. Now it would actually keep the buffer around so you can load it back using `:buffer` later.

Also updated some docs to make it clearer.

These bugs were identified when working on #11312.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -538,7 +538,7 @@ close_buffer(
 	unload_buf = TRUE;
 
 #ifdef FEAT_TERMINAL
-    if (bt_terminal(buf) && (buf->b_nwindows == 1 || del_buf))
+    if (bt_terminal(buf) && (buf->b_nwindows <= 1 || del_buf))
     {
 	CHECK_CURBUF;
 	if (term_job_running(buf->b_term))
@@ -550,6 +550,11 @@ close_buffer(
 
 		// Wiping out or unloading a terminal buffer kills the job.
 		free_terminal(buf);
+
+		// A terminal buffer is wiped out when job has finished.
+		del_buf = TRUE;
+		unload_buf = TRUE;
+		wipe_buf = TRUE;
 	    }
 	    else
 	    {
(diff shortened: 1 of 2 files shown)


src/buffer.c
```diff
@@ -538,7 +538,7 @@ close_buffer(
 	unload_buf = TRUE;
 
 #ifdef FEAT_TERMINAL
-    if (bt_terminal(buf) && (buf->b_nwindows == 1 || del_buf))
+    if (bt_terminal(buf) && (buf->b_nwindows <= 1 || del_buf))
```
ychin, line 541: This is necessary because `close_buffer` gets called multiple times. When using `bunload`, it will be called without `unload_buf` set first, and then called again when `b_nwindows == 0`. This change will make `bunload` work properly.

src/buffer.c
```diff
@@ -550,6 +550,11 @@ close_buffer(
 
 		// Wiping out or unloading a terminal buffer kills the job.
 		free_terminal(buf);
+
+		// A terminal buffer is wiped out when job has finished.
```
ychin, line 554: This prevents the stale empty buffer left behind when `:q!` is used.

brammool:
> Fix a couple bugs and inconsistencies in terminal window management
> and buffer deletion:
> 
> 1. Using `:q!` on a live terminal would previously leave an empty
> buffer behind. Now it would completely remove the buffer as expected.
> 2. Using `:bunload!` on a live terminal would previously not do
> anything with the terminal job and just hide the window. Now it would
> do as the name suggests and actually kill the job and unload the
> buffer.
> 3. Using `:hide` on a finished terminal window would previously unload
> the buffer. Now it would actually keep the buffer around so you can
> load it back using `:buffer` later.
> 
> Also updated some docs to make it clearer.

Thanks.  But I don't see a change for the docs.  I'll make some, please
check.

-- 
DEAD PERSON:  I'm getting better!
CUSTOMER:     No, you're not -- you'll be stone dead in a moment.
MORTICIAN:    Oh, I can't take him like that -- it's against regulations.

ychin:
> Thanks. But I don't see a change for the docs. I'll make some, please check.

I must have missed the file but it was only a line or two. The new one seems fine to me.

PR #11343, opened 2022-10-11
Typo in os_haiku.txt

m12n:
Simply add ncurses and not ncurses6 to build 9.x from scratch on haiku beta 3. Using ncurses6 will cause configure to fail. Also, the multi-line command examples didn't end properly in my mind but it's out of scope for this fix.


brammool:
I'll include it, thanks.

PR #7193, opened 2020-10-23
make myplugin#myfunction() work in vimrc even when myplugin is a package

bfredl:
Situation:

Create the following file `/home/USER/.vim/pack/foo/start/bar/autoload/testpak.vim` (or elsewhere on `&packpath`):

```
function testpak#testfun()
  return -10
endfunction
```
and somewhere in `vimrc` do

```
call testpak#testfun()
```

problem: autoloading functions from functions before VimEnter does not work
solution: make it work

NB: you can use `packadd! testpak` of course. But I don't think "autoload" means anything if it is not automatic anymore. Then you could just as well make it a single-file plugin already and not use autoload ever again.

If needed, I can make a test.

PR #11349, opened 2022-10-11
Add documentation for autoload (and :color) auto-load packages

ychin:
Autoload functionality now automatically searches 'packpath' to support calling package functions from a vimrc file (#7193). However, the documentation were not updated in that commit, so it's still a little confusing to the user exactly how this works. Document that both in the packages docs and autoload docs.

Also add some mentioning of :colorscheme's behavior in the ~~autoload~~ (Edit: package) docs as well as it also does something similar.


brammool:
Thanks, I'll include it.

PR #11353, opened 2022-10-12
sed.vim: allow end-of-line comments

### LINKED ISSUE #5876: GNU sed syntax coloring for comments
I think only one line should be changed in the file `/usr/share/vim/vim82/syntax/sed.vim`, from this
```
syn match sedComment	"^\s*#.*$"
```
to **maybe** this:
```
syn match sedComment	"\s*#.*$"
```
(I would have expected this to be too much, as by itself it would match even starting from a `#` which is a literal part of a regex or of a substitiution, or whatever, but this seems to be not the case, maybe because some other rule rules this out.)

The systax coloring for sed files should color the comments correctly for GNU sed, which allows leading comments on any line. Since this definition of the comment is less restrictive than then one for some prehistoric sed, this should not affect the syntax coloring of those scripts which are written for the latter.

The alternative is to keep having a feast of colors whenever I want to put a comment on the right of a line.

I can make a pull request myself, but it would a 1 char change commit.

jacktose:
Closes #5876.

Allows (does not highlight as Error) end-of-line comments. Makes this default, but allows old behavior with `sed_syntax_style="bsd"`.

I have tested on Windows gVim 9.0.626 & linuxbrew Vim 9.0.700. I don't have an easy way to test on BSD or other platforms, but nothing should break; the worst case should be highlighting inconsistent with the platform's sed syntax.

Note [from the issue](https://github.com/vim/vim/issues/5876#issuecomment-608265394) that the file is apparently abandoned by the listed maintainer.

### DIFF
--- a/runtime/syntax/sed.vim
+++ b/runtime/syntax/sed.vim
@@ -2,7 +2,7 @@
 " Language:	sed
 " Maintainer:	Haakon Riiser <hakonrk@fys.uio.no>
 " URL:		http://folk.uio.no/hakonrk/vim/syntax/sed.vim
-" Last Change:	2010 May 29
+" Last Change:	2022 Oct 13
 
 " quit when a syntax file was already loaded
 if exists("b:current_syntax")
@@ -17,8 +17,13 @@ syn match sedAddress	"[[:digit:]$]"
 syn match sedAddress	"\d\+\~\d\+"
 syn region sedAddress   matchgroup=Special start="[{,;]\s*/\(\\/\)\="lc=1 skip="[^\\]\(\\\\\)*\\/" end="/I\=" contains=sedTab,sedRegexpMeta
 syn region sedAddress   matchgroup=Special start="^\s*/\(\\/\)\=" skip="[^\\]\(\\\\\)*\\/" end="/I\=" contains=sedTab,sedRegexpMeta
-syn match sedComment	"^\s*#.*$"
 syn match sedFunction	"[dDgGhHlnNpPqQx=]\s*\($\|;\)" contains=sedSemicolon,sedWhitespace
+if exists("g:sed_dialect") && g:sed_dialect ==? "bsd"
+    syn match sedComment	"^\s*#.*$"
+else
+    syn match sedFunction	"[dDgGhHlnNpPqQx=]\s*\ze#" contains=sedSemicolon,sedWhitespace
+    syn match sedComment	"#.*$"
+endif
 syn match sedLabel	":[^;]*"
 syn match sedLineCont	"^\(\\\\\)*\\$" contained
 syn match sedLineCont	"[^\\]\(\\\\\)*\\$"ms=e contained
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -3024,12 +3024,21 @@ To make tabs stand out from regular blanks (accomplished by using Todo
 highlighting on the tabs), define "highlight_sedtabs" by putting >
 
 	:let highlight_sedtabs = 1
-
+<
 in the vimrc file.  (This special highlighting only applies for tabs
 inside search patterns, replacement texts, addresses or text included
 by an Append/Change/Insert command.)  If you enable this option, it is
 also a good idea to set the tab width to one character; by doing that,
 you can easily count the number of tabs in a string.
+
+GNU sed allows comments to follow other text on the same line.
+BSD sed only allows comments when "#" is the first character of the line.
+To enforce BSD-style comments, i.e. mark end-of-line comments as errors, use: >
+
+	:let sed_dialect = "bsd"
+<
+Note that there are other differences between GNU sed and BSD sed which are
+not (yet) affected by this setting.
 
 Bugs:
 


jacktose:
I have tested on Windows gVim 9.0.626 & linuxbrew Vim 9.0.700. I don't have an easy way to test on BSD or other platforms, but nothing should break; the worst case should be highlighting inconsistent with the platform's sed syntax.

jacktose:
Examples:

<details open>
<summary>Default</summary>

sed-default

Note that there are still other highlighting issues, but fixing those is out of scope.
</details>

<details>
<summary>Old style: <code>:let sed_syntax_style = "bsd"</code></summary>

sed-bsd
Note that the error highlighting doesn't work great here. That is the current behavior, and fixing it is out of scope.
</details>

<details>
<summary>Variable value is case-insensitive: <code>:let sed_syntax_style = "BSD"</code></summary>

sed-BSD-caps
</details>

<details>
<summary>Other values behave like default</summary>

sed-blank
</details>

(Please don't judge me by my kludgy sed code.)

brammool:
> Closes #5876
> 
> Allows (does not highlight as Error) end-of-line comments. Makes this
> default, but allows old behavior with `sed_syntax_style="bsd"`.

The help says:

+GNU sed allows comments to follow other text on the same line. BSD sed only
+allows whole-line comments, i.e. `#' must be the first character of the line.
+To mark GNU-style end-of-line comments as errors, use: >
+
+	:let sed_syntax_style = "bsd"

That should probably be "Tom mark BSD-style".

+if !exists("sed_syntax_style")
+    let g:sed_syntax_style = "gnu"
+endif

This makes it hard to see where g:sed_syntax_style was set.  It is
simpler if only the user sets it.  Just use the "exists() where the
value is used, it's only one place.

-- 
FATHER:    You only killed the bride's father - that's all -
LAUNCELOT: Oh dear, I didn't really mean to...
FATHER:    Didn't mean to?  You put your sword right through his head!
LAUNCELOT: Gosh - Is he all right?

jacktose:
Thanks for looking! I was prematurely optimizing for doing more stuff with that value, so I'll simplify it.

Sorry, what is `Tom mark BSD-style`?

jacktose:
@brammool Thanks for looking! I was prematurely optimizing for doing more stuff with that value, so I'll simplify it.

Sorry, what is `Tom mark BSD-style`?

arp242:
> Sorry, what is Tom mark BSD-style?

`s/Tom/To/`

In your example you say "To mark GNU-style end-of-line comments as errors, use: `:let sed_syntax_style = "bsd"`", and that's probably not right.

jacktose:
@arp242 Ah, thank you. That is what I meant, but it's obviously not clear enough, so I'll rewrite it.

jacktose:
@brammool @arp242 is it clearer now? I changed it to:
```
GNU sed allows comments to follow other text on the same line.
BSD sed only allows comments when "#" is the first character of the line.
To enforce BSD-style comments, i.e. mark end-of-line comments as errors, use: >

	:let sed_syntax_style = "bsd"

<
```

dkearns:
@jacktose, thanks for picking this up.  Are you planning on taking over as maintainer?  I think I managed to track the maintainer down to LinkedIn but that's a dead end for me.  It seems I have a quite a few changes in my local copy I could send to you or I could create PRs.

I vaguely recall that I was intending to investigate how other dialects behave but didn't get to it.  My experience is almost exclusively with GNU sed.

I think `sed_dialect` may be a better name for the config variable?  There's no standards but that is used by at least a couple of other runtime files.

runtime/syntax/sed.vim
```diff
@@ -19,6 +19,10 @@ syn region sedAddress   matchgroup=Special start="[{,;]\s*/\(\\/\)\="lc=1 skip="
 syn region sedAddress   matchgroup=Special start="^\s*/\(\\/\)\=" skip="[^\\]\(\\\\\)*\\/" end="/I\=" contains=sedTab,sedRegexpMeta
 syn match sedComment	"^\s*#.*$"
 syn match sedFunction	"[dDgGhHlnNpPqQx=]\s*\($\|;\)" contains=sedSemicolon,sedWhitespace
+if !exists("g:sed_syntax_style") || g:sed_syntax_style !=? "bsd"
+    syn match sedComment	"\s*#.*$"
```
dkearns: This pattern doesn't need the leading whitespace as it only currently follows `sedFunction` which itself includes trailing whitespace and consumes it all.

This doesn't make any difference at the moment but might in the future because this syntax file explicitly matches whitespace with `sedWhitespace`.

This is also taking precedence over the full line comment, on line 20, which now never matches.  Maybe someone has a reason to distinguish between them?  If so, you could name the new one `sedTailComment` and if not you could just move the full line definition to after the conditional block.

runtime/syntax/sed.vim
```diff
@@ -19,6 +19,10 @@ syn region sedAddress   matchgroup=Special start="[{,;]\s*/\(\\/\)\="lc=1 skip="
...
+    syn match sedFunction	"[dDgGhHlnNpPqQx=]\s*" nextgroup=sedComment contains=sedSemicolon,sedWhitespace
```
dkearns: This is relaxing the definition of `sedFunction` so it now matches `d` in `djunk` which would otherwise be highlighted as an error. Maybe use something like the following so it only matches with tail comments.

```vim
syn match sedFunction "[dDgGhHlnNpPqQx=]\s*\ze#" nextgroup=sedComment contains=sedSemicolon,sedWhitespace
```

jacktose: Added `\ze#` per your suggestion. Also I think I misunderstood `nextgroup`, removed. See updated file.

runtime/syntax/sed.vim
jacktose: >This pattern doesn't need the leading whitespace as it only currently follows sedFunction which itself includes trailing whitespace and consumes it all.

You're right, removed the whitespace.

jacktose: I hadn't thought of distinguishing `sedLineComment` and `sedTailComment`, but sure, maybe it helps. But I don't understand the logic enough to make it work. Something with `contains` or `nextgroup` to make sure it comes after one of the other groups?

jacktose:
> jacktose, thanks for picking this up. Are you planning on taking over as maintainer? I think I managed to track the maintainer down to LinkedIn but that's a dead end for me. It seems I have a quite a few changes in my local copy I could send to you or I could create PRs.

I wasn't planning to. I could, if it just meant receiving a couple emails & PRs per decade 😏. But I barely understand the vim syntax language at all, so I don't think I'm a great candidate.

> I think sed_dialect may be a better name for the config variable? There's no standards but that is used by at least a couple of other runtime files.

Yep, that's better. Changed.

brammool:
Looks good now, I'll include it.

dkearns:
> I wasn't planning to. I could, if it just meant receiving a couple emails & PRs per decade smirk. But I barely understand the vim syntax language at all, so I don't think I'm a great candidate.

OK,  I'll take over as maintainer then because I'll probably annoy you with a few more than that now that you've revived this topic. :)  Thanks for the fix.

jacktose:
@dkearns That's great! If I figure out how to fix any of the funny stuff I'm seeing, I'll let you know.

dkearns:
Feel free to send reports.

I'll take care of the single line GNU-specific i, a, c commands like the one highlighted as an error at the top of your example file.

PR #11356, opened 2022-10-12
Fix off-by-one error in column calculation for popup_setpos in mouse (and bugs in visual line and also one-char selection)

ychin:
This fixes an off-by-one error arbitrarily introduced in the popup_setpos calculation (in `get_fpos_of_mouse()`) code in mouse.c. I'm not really sure why it's there as we intentionally decrement the column by 1, but from looking at the code I cannot see why it's there at all, considering the column is already 0-indexed.

Testing this in GTK and MacVim confirms that it fixes the issue, and the function isn't used outside of popup_setpos.

### DIFF
--- a/src/mouse.c
+++ b/src/mouse.c
@@ -596,7 +596,19 @@ do_mouse(
 			jump_flags = MOUSE_MAY_STOP_VIS;
 		    else
 		    {
-			if ((LT_POS(curwin->w_cursor, VIsual)
+			if (VIsual_mode == 'V')
+			{
+			    if ((curwin->w_cursor.lnum <= VIsual.lnum
+				    && (m_pos.lnum < curwin->w_cursor.lnum
+					|| VIsual.lnum < m_pos.lnum))
+				|| (VIsual.lnum < curwin->w_cursor.lnum
+				    && (m_pos.lnum < VIsual.lnum
+					|| curwin->w_cursor.lnum < m_pos.lnum)))
+			    {
+				jump_flags = MOUSE_MAY_STOP_VIS;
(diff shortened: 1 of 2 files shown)


ychin:
Here's a video (taken under Ubuntu GTK, with latest Vim 9.0.735):

https://user-images.githubusercontent.com/1217449/195467169-7c8e0a6f-5fdb-47a7-9995-261ee9278957.mov

Note that right clicking one column to the right still kept the selection, but somehow right clicking at the left-most column of the selection would lose it. This shows the off-by-one error, but I'm not 100% sure why the original code was there. Probably just an oversight.

ychin:
I pushed an additional commit to fix other bugs i found in `popup_setpos` mode, where apparently visual line mode basically doesn't work properly, and if you have normal visual mode with only one character selected it also wouldn't work.

Copy-pasting the commit message below:

Fix popup_setpos bugs in visual line mode, and when selecting only one char

Fix two bugs related to popup_setpos:

1. When selecting only one character in visual mode, right-clicking away
   wouldn't deselect the selection. That's because the code was only
   using less-than operators to check w_cursor against VIsual, instead
   of using less-than-or-equal.
2. Visual line didn't really work properly. For one, it suffered the
   above issue of only one character selected, and second, the logic
   didn't really account for line selection (so if you click at the
   beginning or end of line, you would end up deselecting even though
   you still right clicked on the selection). Fix this by adding a
   special check for visual line mode to make sure it works by line
   number instead of cursor pos.

ychin:
Also added tests. Should be good to review now.

PR #11365, opened 2022-10-14
Reuse the same grow buffer across quickfix entries

yegappan:
The grow buffer is allocated and freed for each quickfix entry. This is inefficient.  Reuse the same grow buffer across all the entries.

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -472,15 +472,7 @@ free_all_mem(void)
 # endif
 
 # ifdef FEAT_QUICKFIX
-    {
-	win_T	    *win;
-	tabpage_T   *tab;
-
-	qf_free_all(NULL);
-	// Free all location lists
-	FOR_ALL_TAB_WINDOWS(tab, win)
-	    qf_free_all(win);
-    }
+    free_quickfix();
 # endif
 
     // Close all script inputs.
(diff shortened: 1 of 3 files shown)


brammool:
> The grow buffer is allocated and freed for each quickfix entry. This
> is inefficient.  Reuse the same grow buffer across all the entries.

Thanks, I'll include it.

Perhaps the buffer should be cleared once an operation is done, e.g.
when it uses more than 300 bytes of memory.

-- 
TIM: To the north there lies a cave,  the cave of Caerbannog, wherein, carved
     in mystic runes, upon the very living rock, the last words of Olfin
     Bedwere of Rheged make plain the last resting place of the most Holy
     Grail.

vim-ml:
Hi Bram,

On Fri, Oct 14, 2022 at 5:08 AM Bram Moolenaar ***@***.***>
wrote:

>
> > The grow buffer is allocated and freed for each quickfix entry. This
> > is inefficient. Reuse the same grow buffer across all the entries.
>
> Thanks, I'll include it.
>
> Perhaps the buffer should be cleared once an operation is done, e.g.
> when it uses more than 300 bytes of memory.
>
>
>
I have created PR #11379 to free the grow array buffer after use if it uses
more
than 2K bytes of memory.

- Yegappan

PR #11366, opened 2022-10-14
Add TextChangedT

Shougo:
It is the port of https://github.com/neovim/neovim/pull/20567.

It is useful to create the auto completion plugin for the terminal.

スクリーンショット_2022-10-14_17-13-42

### DIFF
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -409,6 +409,7 @@ Name			triggered by ~
 			when popup menu is not visible
 |TextChangedP|		after a change was made to the text in Insert mode
 			when popup menu visible
+|TextChangedT|		after a change was made to the text in Terminal mode
 |TextYankPost|		after text has been yanked or deleted
 
 |SafeState|		nothing pending, going to wait for the user to type a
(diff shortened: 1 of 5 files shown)


brammool:
> It is the port of https://github.com/neovim/neovim/pull/20567.
> 
> It is useful to create the auto completion plugin for the terminal.
> 
> スクリーンショット_2022-10-14_17-13-42

Please extend the test with some nasty things that an autocommand can
do:
- Have the terminal update while the current buffer is another window.
	Does it get the right buffer?
- Have the autocmd change to another window.
- Have the autocmd close the terminal (should fail, I suppose).

Probably a few more things.  If you look at wheren EVENT_TEXTCHANGEDI is
triggered it uses aucmd_prepbuf() and aucmd_restbuf(), perhaps something
like that is needed?  We can probably assume the terminal is visible in
a window (oh, make sure to not trigger when it isn't).   Maybe b_locked
needs to be set, like it is in parse_osc().

Sorry that this may seem like a lot of comments for a simple change, but
we know that autocommands have created bugs and crashes in the past, we
need to at least try to prevent that when adding a new event.

-- 
TALL KNIGHT: When you have found the shrubbery, then you must cut down the
             mightiest tree in the forest ... with a herring.

Shougo:
> Sorry that this may seem like a lot of comments for a simple change, but
we know that autocommands have created bugs and crashes in the past, we
need to at least try to prevent that when adding a new event.

OK.  I will add them.

Shougo:
> Does it get the right buffer?

It is the bug.  I have fixed it.

I have added the tests.

PR #11371, opened 2022-10-15
Add missing Ex commands to help index, and add test to catch them in future

ychin:
This is similar to test_function_lists.vim, which makes sure we won't forget to add this again in the future.

I omitted the Vim9script specific stuff as those seems more like they are only useful for scripting and not really for Ex cmds.

### DIFF
--- a/runtime/doc/index.txt
+++ b/runtime/doc/index.txt
@@ -1150,7 +1150,7 @@ tag		command		action ~
 |:!!|		:!!		repeat last ":!" command
 |:#|		:#		same as ":number"
 |:&|		:&		repeat last ":substitute"
-|:star|		:*		execute contents of a register
+|:star|		:*		use the last selected area in Visual mode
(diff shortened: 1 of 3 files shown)


ychin:
When I run `cd src; make test_cmd_lists`, *with* the runtime/doc folder reverted, this is what I see, which shows that the test is working correctly:

```
From test_cmd_lists.vim:
Executed Test_cmd_lists()                          in   0.018973 seconds
Executed 1 test                          in   0.062265 seconds
1 FAILED:
Found errors in Test_cmd_lists():
command line..script /Users/ychin/Dev/external/vim/src/testdir/runtest.vim[515]..function RunTheTest[52]..Test_cmd_lists line 55: Missing commands from `:help ex-cmd-index`: [':defer', ':echowindow', ':horizontal']: Expected 0 but got 3
make[1]: *** [test_cmd_lists] Error 1
make: *** [test_cmd_lists] Error 2
```

PR #11372, opened 2022-10-15
Don't clear previous command on :!

arp242:
If I run a command with :!

        :!echo asd
        asd

I can repeat it quickly with :!!

        :!!
        :!echo asd
        asd

Now I accidentally use :! instead of :!!

        :!

And the previous command is gone:

        :!!

Need to hit ENTER to clear the prompt, press :!, up, Enter to get it back.

Been bugging me for years.

---

Also: I'm not sure if there's a good reason for :! to do anything at all instead of being a no-op? It just runs sh -c '', but I don't know if there is some reason for it to behave as it does?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,7 @@ src/testdir/messages
 src/testdir/viminfo
 src/testdir/opt_test.vim
 src/testdir/failed
+src/testdir/starttime
 runtime/indent/testdir/*.out
 runtime/indent/testdir/*.fail
 src/memfile_test
--- a/runtime/doc/various.txt
+++ b/runtime/doc/various.txt
@@ -249,7 +249,8 @@ g8			Print the hex values of the bytes used in the
 
 							*:!cmd* *:!*
 :!{cmd}			Execute {cmd} with the shell.  See also the 'shell'
-			and 'shelltype' option.
+			and 'shelltype' option. :! without a {cmd} is a no-op
+			that does nothing.
 							*E34*
 			Any '!' in {cmd} is replaced with the previous
 			external command (see also 'cpoptions').  But not when
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -957,6 +957,11 @@ do_bang(
 	}
     } while (trailarg != NULL);
 
+    // Don't do anything if there is no command as there isn't really anything
+    // useful to do except run "sh -c ''".
+    if (STRLEN(newcmd) == 0)
+	return;
+
     vim_free(prevcmd);
     prevcmd = newcmd;
 
--- a/src/testdir/test_shell.vim
+++ b/src/testdir/test_shell.vim
@@ -251,4 +251,33 @@ func Test_set_shell()
   call delete('Xtestout')
 endfunc
 
+func Test_shell_repeat()
+  let save_shell = &shell
+
+  call writefile(['#!/bin/sh', 'echo "Cmd: [$*]" > Xlog'], 'Xtestshell', 'D')
+  call setfperm('Xtestshell', "r-x------")
+  set shell=./Xtestshell
+  defer delete('Xlog')
+
+  call feedkeys(":!echo coconut\<CR>", 'xt')   " Run command
+  call assert_equal(['Cmd: [-c echo coconut]'], readfile('Xlog'))
+
+  call feedkeys(":!!\<CR>", 'xt')              " Re-run previous
+  call assert_equal(['Cmd: [-c echo coconut]'], readfile('Xlog'))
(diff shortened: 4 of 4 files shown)


brammool:
> If I run a command with :!
> 
>         :!echo asd
>         asd
> 
>         Press ENTER or type command to continue
> 
> I can repeat it quickly with :!!
> 
>         :!!
>         :!echo asd
>         asd
> 
>         Press ENTER or type command to continue
> 
> Now I accidentally use :! instead of :!!
> 
>         :!
> 
>         Press ENTER or type command to continue
> 
> And the previous command is gone:
> 
>         :!!
> 
>         Press ENTER or type command to continue
> 
> Need to hit ENTER to clear the prompt, press :!, up, Enter to get it back.
> 
> Been bugging me for years.
> 
> ---
> 
> Also: I'm not sure if there's a good reason for :! to do anything at
> all instead of being a no-op? It just runs sh -c '', but I don't know
> if there is some reason for it to behave as it does?

I don't think there was any thought about this, it looks like the code
does not expect the user to type ":!<CR>".  I cannot think of a reason
why executing the shell with an empty command has any purpose.  For
typing ":<CR>" nothing happens, thus making ":!<CR>" do nothing, just
bail out early, should be OK.

Oh, there is one reason I can think off: what if a user intentionally
wants to clear the previous command?  Is that relevant?

-- 
I just planted an Algebra tree. It has square roots.

arp242:
Okay, I changed it so that `:!` is a no-op, which also doesn't trigger the hit-enter prompt.

> Oh, there is one reason I can think off: what if a user intentionally wants to clear the previous command?  Is that relevant?

Hm, I don't know... Could put this behaviour behind an option I suppose.

Another change I want to make is have :! not display a hit-enter prompt if the output is just one line, and the ability to output :! in the echo window. We we need a "bangopt" for that anyway.

brammool:
> Okay, I changed it so that `:!` is a no-op, which also doesn't trigger
> the hit-enter prompt.
> 
> > Oh, there is one reason I can think off: what if a user
> > intentionally wants to clear the previous command?  Is that
> > relevant?
> 
> Hm, I don't know... Could put this behaviour behind an option I suppose.

Yet another option?  No, this would be something the user types, thus it
would have to be a command rather than an option.  Something like
:!NONE.  But like I said, it might not be needed, just something that
was possible before but not clear if anybody used it.  We can just wait
for someone to ask for it.

> Another change I want to make is have :! not display a hit-enter
> prompt if the output is just one line, and the ability to output :! in
> the echo window. We we need a "bangopt" for that anyway.

Perhaps a command modifier?  ":winout !cmd".  Could be useful for other
commands as well to avoid a hit-enter prompt by putting the output in
the message window.  It would disappear after three seconds, but that's
fine if you just want to check that it worked.

The problem with shell commands is that they might output something to
the terminal anyway and mess up the display.  Well, you could use CTRL-L
if needed.

-- 
ARTHUR:  No, hang on!  Just answer the five questions ...
GALAHAD: Three questions ...
ARTHUR:  Three questions ...  And we shall watch ... and pray.

arp242:
> Something like :!NONE.  But like I said, it might not be needed, just something that was possible before but not clear if anybody used it.  We can just wait for someone to ask for it.

Sounds good!

> Perhaps a command modifier?  ":winout !cmd".

It's a lot more effort to type than ":!cmd" and ":!!" though.

brammool:
> > Perhaps a command modifier?  ":winout !cmd".
> 
> It's a lot more effort to type than ":!cmd" and ":!!" though.

Well, if you actually type this, then hitting "Enter" (after checking
the output) isn't that much effort.  If it happens because of what some
plugin is doing, especially when in many cases there is not output, but
when there is you might want to have a look at it, then the command
modifier is in a script and it won't need to be typed.

-- 
BRIDGEKEEPER: What is your favorite colour?
GAWAIN:       Blue ...  No yelloooooww!

arp242:
>>> Perhaps a command modifier? ":winout !cmd".
>> 
>> It's a lot more effort to type than ":!cmd" and ":!!" though.
>
> Well, if you actually type this, then hitting "Enter" (after checking the output) isn't that much effort. If it happens because of what some plugin is doing, especially when in many cases there is not output, but when there is you might want to have a look at it, then the command modifier is in a script and it won't need to be typed.

I mostly just use :! manually; :!go run %, :!go test, stuff like that.

Either way, bit of a different issue; I might have a look at some point, or maybe not. We'll see.

brammool:
As mentioned in #11409 we probably need to change this to execute :! as before, but then not update the previous command.

PR #11376, opened 2022-10-15
CI: add missing pkgs lua ruby perl (mac m1)

pheiduck:


pheiduck:
does not take any effect.

PR #11378, opened 2022-10-15
CI: appveyor clone_depth: 3

pheiduck:
By default, AppVeyor clones the entire repository with the entire history.
Limit the number of recent commits to clone. As we do on cirrus ci


brammool:
Can you provide information about this change?  Perhaps adding a comment would help future changes to understand what it's for.

chrisbra:
does it actually gain us anything? I think limiting the depth for checkouts is not really worth it, last time I checked it was only some seconds faster than a full check out, so in my opininion not worth it.

PR #11379, opened 2022-10-15
Free the quickfix grow array buffer (if it is large) after use

yegappan:


brammool:
Thanks.  Assuming most lines aren't that long, I think using a limit of 1000 makes sense.  It's rather arbitrary anyway.

Issue #11279, opened 2022-10-04
vim9 indent issue


habamax:
### Steps to reproduce

1. `vim -Nu NONE`
2. `filetype indent on`
3. paste text:
```
vim9script
def Test()
    var winid = popup_create([], {
        filter: (id, key) => {
            if key == "\<esc>"
                popup_settext(id, Printify(filtered_items, []))
                popup_setoptions(id,
                    {title: $"hello world"})
            endif
        }
    })
enddef
```
4. `set ft=vim`
5. `gg=G`

### Expected behaviour

`popup_setoptions` should have same indent level as `popup_settext`:

```
vim9script
def Test()
	var winid = popup_create([], {
		filter: (id, key) => {
			if key == "\<esc>"
				popup_settext(id, Printify(filtered_items, []))
				popup_setoptions(id,
					{title: $"hello world"})
			endif
		}
	})
enddef
```

Instead
```
vim9script
def Test()
	var winid = popup_create([], {
		filter: (id, key) => {
			if key == "\<esc>"
				popup_settext(id, Printify(filtered_items, []))
			popup_setoptions(id,
				{title: $"hello world"})
			endif
		}
	})
enddef
``` 

 https://asciinema.org/a/lchB6YegVCFahA4mzkLJajBsB

### Version of Vim

9.0.0654

### Environment

- Fedora 36
- Any term
- Any shell

lacygoill:
Patch fixing the issue:
```diff
diff --git a/runtime/autoload/dist/vimindent.vim b/runtime/autoload/dist/vimindent.vim
index 00977dc3d..79d8b73fa 100644
--- a/runtime/autoload/dist/vimindent.vim
+++ b/runtime/autoload/dist/vimindent.vim
@@ -2,7 +2,7 @@ vim9script
 
 # Language:     Vim script
 # Maintainer:   github user lacygoill
-# Last Change:  2022 Sep 24
+# Last Change:  2022 Oct 5
 
 # NOTE: Whenever you change the code, make sure the tests are still passing:
 #
@@ -156,7 +156,7 @@ const ASSIGNS_HEREDOC: string = $'^\%({COMMENT}\)\@!.*\%({HEREDOC_OPERATOR}\)\s\
 
 # CD_COMMAND {{{3
 
-const CD_COMMAND: string = $'[lt]\=cd!\=\s\+-{END_OF_COMMAND}'
[...]

Issue #11377, opened 2022-10-15
Incorrect indent for vimscript in recent releases.


skywind3000:
### Steps to reproduce

The recent PR https://github.com/vim/vim/pull/11079 broke the `elseif` indent in vimscript

`elseif` will indent to a wrong position if you press `<c-n>` and a `<space>` after `elseif`

Most well known completion plugins will feed a `<c-n>` to automatically display the popup menu if user has input 2 or more alphabets in one word.

reproduce:

edit `error.vim`:

```VimL
set nocompatible
set backspace=eol,start,indent
set cindent
set autoindent
set number

set shiftwidth=4
set softtabstop=4
set tabstop=4
set noexpandtab

syntax enable
syntax on

filetype plugin indent on

set cpt=.,w,k
set completeopt=menu,menuone,noselect
set shortmess+=c

setlocal ft=vim 
call feedkeys('i')
call feedkeys("if 1\n")
call feedkeys("echo 11\n")
call feedkeys("elseif")

" wrong indent here
call feedkeys("\<c-n> ")

call feedkeys("\<esc>")
```

and start the latest vim (9.0.752) with it:

```bash
vim -u error.vim
```

### Expected behaviour

the buffer content should be:

```
if 1
    echo 11
elseif {cursor}
```

All old releases (before PR https://github.com/vim/vim/pull/11079) works as expected.

After switching to the latest 9.0.752, the buffer becomes:

```
if 1
    echo 11
    elseif {cursor}
```

Currently, every time I type `elseif` I always get a wrong indentation 
when I am using `YCM` or some other completion plugins.

### Version of Vim

9.0.752

### Environment

OS: Debian 11
Terminal: SecureCRT

brammool:
@lacygoill

lacygoill:
```diff
diff --git a/runtime/autoload/dist/vimindent.vim b/runtime/autoload/dist/vimindent.vim
index 00977dc3d..6ba04c2ba 100644
--- a/runtime/autoload/dist/vimindent.vim
+++ b/runtime/autoload/dist/vimindent.vim
@@ -2,7 +2,7 @@ vim9script
 
 # Language:     Vim script
 # Maintainer:   github user lacygoill
-# Last Change:  2022 Sep 24
+# Last Change:  2022 Oct 15
 
 # NOTE: Whenever you change the code, make sure the tests are still passing:
 #
[...]

PR #11381, opened 2022-10-15
fix vim indent issues

lacygoill:
This should fix issues https://github.com/vim/vim/issues/11279 and https://github.com/vim/vim/issues/11377


brammool:
I'll include it, thanks.

PR #11383, opened 2022-10-16
help.vim: update helpHeadline and fix version5.txt

c4rlo:
The effects of these tweaks can be seen [here](https://gist.github.com/c4rlo/b6a3f37dbe6934067c383ad0235376a3/revisions) (format explained [here](https://gist.github.com/c4rlo/b6a3f37dbe6934067c383ad0235376a3#file-readme-md)).

### DIFF
--- a/runtime/doc/version5.txt
+++ b/runtime/doc/version5.txt
@@ -118,7 +118,7 @@ Added					|added-5.8|
 Fixed					|fixed-5.8|
 
 ==============================================================================
-				 INCOMPATIBLE		*incompatible-5*
+INCOMPATIBLE				 		*incompatible-5*
 
 Default value for 'compatible' changed			*cp-default*
 --------------------------------------
@@ -360,7 +360,7 @@ CTRL-_ key for this |i_CTRL-_|. >
    :imap <C-B> <C-O>:set revins!<CR>
 
 ==============================================================================
-				 NEW FEATURES		*new-5*
+NEW FEATURES				 		*new-5*
 
 Syntax highlighting					*new-highlighting*
 -------------------
@@ -635,7 +635,7 @@ Included support for the Farsi language (Shiran).  Only when enabled at
 compile time.  See |farsi|.
 
 ==============================================================================
-				 IMPROVEMENTS		*improvements-5*
+IMPROVEMENTS				 		*improvements-5*
 
 Performance:
 - When 'showcmd' was set, mappings would execute much more slowly because the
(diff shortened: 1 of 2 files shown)


brammool:
> The effects of these tweaks can be seen in a gist which I will publish
> shortly.

Hmm, this actually highlights several files with too many spaces.
Replacing them with Tabs is a better solution.

Some headers that intentionally have some indent are no longer
highlighted, that's not good.

Adding the question mark at the end looks good.  But it's probably
better to only accept one after the text, I don't see a case where a
question mark would go halfway the text.

I also see things in the gist that I don't see in the help files, e.g.
in pi_getscript.xt, lines starting with a star...  Ah, the spaces are
highlighted, but it's invisible.

I think it's better to accept spaces at the start and then at least one
non-white character:   ^ *[-A-Z.]

Then the match becomes:

syn match helpHeadline		"^ *[-A-Z.][-A-Z0-9 .()_]*?\=\ze\(\s\+\*\|$\)"


-- 
Close your shells, or I'll kill -9 you
Tomorrow I'll quota you
Remember the disks'll always be full
And then while I'm away
I'll write ~ everyday
And I'll send-pr all my buggings to you.
    [ CVS log "Beatles style" for FreeBSD ports/INDEX, Satoshi Asami ]

c4rlo:
> Adding the question mark at the end looks good.  But it's probably better to only accept one after the text, I don't see a case where a question mark would go halfway the text.

Ok; I've updated my proposal. It happens to make no difference with the current help files.

> Some headers that intentionally have some indent are no longer highlighted, that's not good.

I don't think that's true. All the lines that are no longer highlighted don't seem intended as headings. Though there are indeed some indented headings that aren't highlighted, but they also weren't before. For example, [this](https://github.com/vim/vim/blob/436e5d395fd629c8d33b5cf7b373aad007f16851/runtime/doc/version5.txt#L121).

> I think it's better to accept spaces at the start and then at least one non-white character:   `^ *[-A-Z.]`
>
> Then the match becomes:
>
> ```vim
> syn match helpHeadline		"^ *[-A-Z.][-A-Z0-9 .()_]*?\=\ze\(\s\+\*\|$\)"
> ```

This still doesn't catch the example I gave above, where the leading whitespace is tabs not spaces. (It also doesn't include the `,` which means it misses out on some headlines.)

Now, we could modify the regex to accept leading tabs as well as spaces. But it would catch a lot of things that aren't headings.

And upon closer inspection, I see that the headings in `version5.txt` (such as the above example) are the only cases where a legitimate headline has leading whitespace.

I've therefore updated this PR to leave my proposed regex mostly as is, apart from the question mark change mentioned above; and to also update the headings in `version5.txt` to have no leading whitespace.

https://github.com/c4rlo/vim-help-headline contains an analysis of the different options for the regex, with the result for each of them.

brammool:
OK, let me include it like this then.  It's really fine tuning now.

PR #11386, opened 2022-10-16
Use early return to simplify code in change.c and channel.c

yegappan:


brammool:
It should be OK now to have a declaration halfway a block.  I wasn't
100% sure at first that all compilers would handle it, but the first
changes went out a few weeks ago and there were no complaints.

I'm not saying we should go all over the place and change it now, but in
cases like this there is no need to move a declaration.  I have just
done this for libvterm, to reduce the difference with the original code.

I'll change it for this PR.

-- 
Sometimes I think the surest sign that intelligent life exists elsewhere
in the universe is that none of it has tried to contact us.     (Calvin)

PR #11387, opened 2022-10-16
'scroll' is computed in a weird place

luukvbaal:
Problem:	'scroll' is computed in a weird place.
Solution:	Compute 'scroll' when window size is set.

Although it doesn't actually solve an issue here, I see no reason to call `win_comp_scroll()` at the end of `scroll_to_fraction()`(also inadvertently calling it twice for `spk=="screen"`) rather than when actually setting the size in `win_new_height()`. This change happens to fix an issue occurring for neovim floating windows: https://github.com/neovim/neovim/pull/20685.

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -6378,7 +6378,6 @@ win_fix_scroll(int resize)
 	    invalidate_botline_win(wp);
 	    validate_botline_win(wp);
 	}
-	win_comp_scroll(wp);
 	wp->w_prev_height = wp->w_height;
 	wp->w_prev_winrow = wp->w_winrow;
     }
@@ -6472,6 +6471,7 @@ win_new_height(win_T *wp, int height)
 
     wp->w_height = height;
     wp->w_skipcol = 0;
+    win_comp_scroll(wp);
 
     // There is no point in adjusting the scroll position when exiting.  Some
     // values might be invalid.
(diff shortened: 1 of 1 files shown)


luukvbaal:
Of course it's possible I'm just missing the reason why it was in `scroll_to_fraction()` in the first place. If so, I would like to know, then I'll look for another solution to https://github.com/neovim/neovim/issues/20684.

brammool:
> Of course it's possible I'm just missing the reason why it was in
> `scroll_to_fraction()` in the first place. If so, I would like to
> know, then I'll look for another solution to
> https://github.com/neovim/neovim/issues/20684.

It must have been there for a long time.  So long as it was the only
function being called after moving around it was probably OK, but after
the split to win_fix_scroll() it was doubled.

The value only depends on w_height, and you moved it to the only place
where w_height is changed, looks like that is OK.  Perhaps it makes the
verbose info about where it was set more accurate.

-- 
So when I saw the post to comp.editors, I rushed over to the FTP site to
grab it.  So I yank apart the tarball, light x candles, where x= the
vim version multiplied by the md5sum of the source divided by the MAC of
my NIC (8A3FA78155A8A1D346C3C4A), put on black robes, dim the lights,
wave a dead chicken over the hard drive, and summon the power of GNU GCC
with the magic words "make config ; make!".
		[Jason Spence, compiling Vim 5.0]

PR #11390, opened 2022-10-17
Add a guard to manpager.vim

### LINKED ISSUE #11388: Command already exists: add ! to replace it: MANPAGER call s:ManPager()
### Steps to reproduce

When I vim a file, no matter new one or existed one, I will encounter an error:
```
Error detected while processing /usr/share/vim/vim90/plugin/manpager.vim:
line    6:
E174: Command already exists: add ! to replace it: MANPAGER call s:ManPager()
```

It seems *Apple M1 Max* brings the problem. But how can I fix it?

### Expected behaviour

No more annoying error alert

### Version of Vim

9.0.217

### Environment

macOS Monterey 12.6
Apple M1 Max

### Logs and stack traces

_No response_

k-takata:
Closes #11388

Add a guard to manpager.vim to avoid double loading.

### DIFF
--- a/runtime/plugin/manpager.vim
+++ b/runtime/plugin/manpager.vim
@@ -2,6 +2,11 @@
 " Maintainer: Enno Nagel <ennonagel+vim@gmail.com>
 " Last Change: 2022 Sep 30
 
+if exists('g:loaded_manpager_plugin')
+  finish
+endif
+let g:loaded_manpager_plugin = 1
+
 " Set up the current buffer (likely read from stdin) as a manpage
 command MANPAGER call s:ManPager()
 


brammool:
Not sure if @casparen is the right github user

chrisbra:
I think it is @konfekt ?

Konfekt:
Looks good to me. 

@k-takata : The other boilerplate guards, such as

```vim
let s:keepcpo          = &cpo
set cpo&vim

...

let &cpo= s:keepcpo
unlet s:keepcpo
```

are not necessary?

k-takata:
The cpo guard is only needed when line continuation is used, and it is not used in this file. So, it is optional.

Issue #7562, opened 2020-12-27
Terminal left in normal mode when execuiting interactive shell command from auto command


g0xA52A2A:
**Describe the bug**

When running an interactive shell command with `!` from an auto command it appears that the terminal is left in normal mode rather than being returned to application mode.

This can be observed as the arrow keys not functioning as expected.

I'm reasonably sure this issue is normal/application or CSI/SS3 issue as the faulty state can be reminded with.

```
!tput smkx
```

**To Reproduce**

The following `test.vim` file serves as reproducer. I would note this does not seem to be specific to the `QuickFixCmdPost` event.

```
autocmd   QuickFixCmdPost * !less test.vim
doautocmd QuickFixCmdPost
```

Run the following.

```
vim --clean -S test.vim
```

Then press `<Left>` to get the error.

```
E349: No identifier under cursor
```

Whilst in this faulty state let us enter `<C-V><Left>` in insert mode to see what Vim is receiving. Rather telling as `[D` would have produced the previously seen error.

```
^[[D
```

By comparison when not on this state we get the following.

```
^[OD
```

In both faulty and good states querying `set t_kl?` shows the same.

```
t_kl <Left>      ^[O*D
```

As a note prefixing the test with the following makes no difference.

```
let &t_TI = ""
let &t_TE = ""
```

**Expected behavior**

Behaviour of keys when returning from an interactive command should be the same, whether that command was typed by the user in the command-line or run form an auto command.

**Environment (please complete the following information):**
 - Vim version
```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jun  3 2020 15:02:02)
macOS version
Included patches: 1-891
Compiled by gbrown@lappy.local
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             +mouse_sgr         +tag_binary
+arabic            +file_in_path      -mouse_sysmouse    -tag_old_static
+autocmd           +find_in_path      +mouse_urxvt       -tag_any_white
+autochdir         +float             +mouse_xterm       -tcl
-autoservername    +folding           +multi_byte        +termguicolors
-balloon_eval      -footer            +multi_lang        +terminal
+balloon_eval_term +fork()            -mzscheme          +terminfo
-browse            +gettext           +netbeans_intg     +termresponse
++builtin_terms    -hangul_input      +num64             +textobjects
+byte_offset       +iconv             +packages          +textprop
+channel           +insert_expand     +path_extra        +timers
[...]

PR #11396, opened 2022-10-18
Fix Home/End keys not recognized at hit-Enter prompt

trygveaa:
As mentioned in issue #7562 and patch 8.2.2246, cursor keys are not recognized at the hit-Enter prompt after executing an external command. While arrow keys were fixed by that, the Home and End keys are also considered cursor keys by xterm and also change control sequences between normal and application mode, so they were still not working.

To be honest, the reason I discovered this is that after enabling the kitty keyboard protocol as discussed in PR #11364 the Home and End keys stopped working. While debugging this, I noticed this inconsistency between the handling of the arrow keys and Home/End keys since the arrow keys still worked. The reason for this is that in this mode, kitty sends the arrow and Home/End keys the same way as in normal terminal mode, and vim supports that for arrow keys, but not for Home/End keys.

I would hesitate to change the builtin termcap for xterm to just fix behavior in kitty, but since this is an inconsistency between the handling of the arrow keys and the Home/End keys that I don't think should be there, and since it actually fixes the #7562 issue for Home/End keys in xterm I think it makes sense to do this change. And it happens to also fix the problem I have with Home/End keys in the kitty keyboard protocol (at least as long as TERM is some xterm name).

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -871,13 +871,13 @@ static struct builtin_term builtin_termcaps[] =
     {K_HELP,		"\033[28;*~"},
     {K_UNDO,		"\033[26;*~"},
     {K_INS,		"\033[2;*~"},
-    {K_HOME,		"\033[1;*H"},
+    {K_HOME,		"\033[@;*H"},
     // {K_S_HOME,		"\033O2H"},
     // {K_C_HOME,		"\033O5H"},
     {K_KHOME,		"\033[1;*~"},
     {K_XHOME,		"\033O*H"},	// other Home
     {K_ZHOME,		"\033[7;*~"},	// other Home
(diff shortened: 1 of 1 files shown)


brammool:
You appear to refer to patch patch 8.2.2246.  Please use patch numbers, those hex hashes are not human readable.

trygveaa:
> You appear to refer to patch patch 8.2.2246. Please use patch numbers, those hex hashes are not human readable.

Okay, fixed.

brammool:
Hmm, for the cursor keys we use this special @ code for the K_X codes, while for the End and Home it's the other way around.  Not sure if that was intentional.  Let's not change it, there is no purpose other than to make it look nicer.

PR #11401, opened 2022-10-19
Fixed win32 terminal mouse scrolling, no longer needs dll

zewpo:
still todo - make mappings work...
But, I think that will take more time, and I guess no reason to wait on mappings before releasing this, at least we can make basic mouse scrolling feature work.


zewpo:
Sorry about the noise generated by this PR.  John's patch to fix my patch conflicted with this new patch - yay for git.  Anyway, this PR looks right to me now.

Issue #11397, opened 2022-10-18
It's not possible to edit "legacy" DOS files ending in ^Z


arp242:
Old systems such as CP/M, MSX, and some versions of DOS used CRLF line endings and a ^Z at the end of the file; for example:

    % cat -v msx.bas
    10 ? "HELLO"^M
    20 GOTO 10^M
    ^Z

    [~]% hexdump -C msx.bas
    00000000  31 30 20 3f 20 22 48 45  4c 4c 4f 22 0d 0a 32 30  |10 ? "HELLO"..20|
    00000010  20 47 4f 54 4f 20 31 30  0d 0a 1a                 | GOTO 10...|

Currently, when fileformat=dos is used it will always drop the ^Z at the end, as documented:

> When "dos" is used, CTRL-Z at the end of a file is ignored. See file-formats and file-read.

This is disabled when 'binary' is set, but then it always uses Unix line endings, for example after editing that file with `vim --clean -b msx.bas` and adding a line I end up with:

    00000000  31 30 20 3f 20 22 48 45  4c 4c 4f 22 0d 0a 31 35  |10 ? "HELLO"..15|
    00000010  20 3f 20 22 57 4f 52 4c  44 22 0a 32 30 20 47 4f  | ? "WORLD".20 GO|
    00000020  54 4f 20 31 30 0d 0a 1a                           |TO 10...|

The second line has a LF, rather than CRLF, but the previous lines and ^Z is preserved.

---

I'm not sure what the best fix for this would be? I know these are really old files for long-since obsolete systems, but there are still enthusiasts playing around with these systems. 

Adding an entire new option for this seems a bit meh, but could be done I suppose. Maybe a way to make 'binary' use CRLF line endings? Although that also has other side-effects, it may be enough for this relatively obscure use case(?)

k-takata:
If you append ^Z at the end of the file then do `:setlocal noendofline nofixendofline`, you can write a file with ^Z.
For example, the following function might be used to append ^Z:

```vim
func AppendCtrlZ()
  let EOF = "\x1a"
  if search(EOF .. '\%$', 'n') == 0
    " Ctrl-Z not found.
    if &endofline
      call append(line('$'), EOF)
    else
      call setline('$', getline('$') .. EOF)
    endif
  endif
  setlocal noendofline nofixendofline
endfunc
```

0cjs:
[...]

PR #11402, opened 2022-10-19
Support offset in readblob()

k-takata:
Support specifying offset and size in `readblob()`.
This should be useful when someone wants to read only a part of a large file.

See: https://github.com/vim/vim/issues/11397#issuecomment-1283487837


brammool:
```
+	if (size == 0)
+	    size = st.st_size - offset;
```
It's better not to change the arguments of a function, it makes debugging a bit harder.
"size" is size_t, which is unsigned.  If offset happens to be larger than st.st_size things go wrong.
It should probably return an empty result if offset is beyond the end of the file.

If offset is negative and -offset is larger than the file size, I wonder what happens.
The fseek() manual doesn't say.

k-takata:
> It's better not to change the arguments of a function, it makes debugging a bit harder.

agree.

> If offset is negative and -offset is larger than the file size, I wonder what happens.
The fseek() manual doesn't say.

It says:

```
ERRORS
       EINVAL The whence argument to fseek() was not  SEEK_SET,  SEEK_END,  or
              SEEK_CUR.  Or: the resulting file offset would be negative.
```

So, that should be an error. However, it doesn't say about when offset goes beyond the end of the file.

k-takata:
I'm thinking whether we should accept "0" as a normal value for "size".
File size can be 0, and `readblob()` returns `0z` for such files.
However, reading 0 bytes from a non-empty file seems not so useful...?

brammool:
> I'm thinking whether we should accept "0" as a normal value for "size".
> File size can be 0, and `readblob()` returns `0z` for such files.
> However, reading 0 bytes from a non-empty file seems not so useful...?

It's always good that an optional argument has a value that means it
works like the argument wasn't there.  Zero seems like a good value for
that.  However, if the size results from some computation, and the
result of the computation ends up being zero, then it's unexpected to
get the whole file.  Perhaps using -1 for "as if the argument wasn't
there" is better?


-- 
   [The rest of the ARMY stand around looking at a loss.]
INSPECTOR END OF FILM: (picks up megaphone) All right!  Clear off!  Go on!

k-takata:
Okay, updated. Let's see if the CI passes.

brammool:
I think it's better to only give an error message when the file can't be read.
When trying to read bytes that aren't there we can just have an empty result.
I'll make it work like that.

0cjs:
How long is this likely to take to appear in currently-used operating systems, such as Debian 11 and Ubuntu 22.04? If this isn't on a branch from which they're taking patches, is there a way to test if this second parameter is available?

chrisbra:
For Debian and Ubuntu not at all, as long as they do not distribute new versions (or unless you add a new repository that provides updated vim builds). You may work around it by using the vim-appimage available e.g. here: https://github.com/vim/vim-appimage/

> is there a way to test if this second parameter is available?

    :echo has("patch-9.0.0795")

PR #11407, opened 2022-10-20
Make readblob() possible to read from character device

k-takata:
With this patch, readblob() can read from a character device if "offset" is zero and "size" is specified (on Unix):

```vim
let b = readblob('/dev/random', 0, 10)
```

(Not sure if this is useful, though.)

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -6859,7 +6859,11 @@ readblob({fname} [, {offset} [, {size}]])			*readblob()*
 			readblob('file.bin', 0, 100)
 <		If {size} is -1 or omitted, the whole data starting from
 		{offset} will be read.
-		When the file can't be opened an error message is given and
+		This can be also used to read the data from a character device
+		on Unix when {size} is explicitly set.  {offset} should be
+		normally zero.  E.g. to read 10 bytes from "/dev/random": >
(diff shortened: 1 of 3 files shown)


brammool:
> With this patch, readblob() can read from a character device if
> "offset" is zero and "size" is specified (on Unix):
> 
> ```vim
> let b = readblob('/dev/random', 0, 10)
> ```
> 
> (Not sure if this is useful, though.)

Hard to say, but it's only a tiny bit of code.

Please also update the help.  And add that example.


-- 
(letter from Mark to Mike, about the film's probable certificate)
      I would like to get back to the Censor and agree to lose the shits, take
      the odd Jesus Christ out and lose Oh fuck off, but to retain 'fart in
      your general direction', 'castanets of your testicles' and 'oral sex'
      and ask him for an 'A' rating on that basis.

k-takata:
Updated.

chrisbra:
> let b = readblob('/dev/random', 0, 10)

Hm, why would we read from `/dev/random`, we have the `rand()` function already. 
Perhaps a better example would be to read using `/dev/ttyS0` to read something from a serial console?

k-takata:
Okay, updated.

k-takata:
Updated about {offset} for character devices.

PR #11408, opened 2022-10-20
Don't remove CTRL-Z

### LINKED ISSUE #11397: It's not possible to edit "legacy" DOS files ending in ^Z

Shougo:
Fix #11397

I have changed the remove of `CTRL-Z` behavior.
It is desired behavior?

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -2272,16 +2272,8 @@ failed:
     /*
      * If we get EOF in the middle of a line, note the fact and
      * complete the line ourselves.
-     * In Dos format ignore a trailing CTRL-Z, unless 'binary' set.
      */
-    if (!error
-	    && !got_int
-	    && linerest != 0
-	    && !(!curbuf->b_p_bin
-		&& curbuf->b_p_fixeol
-		&& fileformat == EOL_DOS
-		&& *line_start == Ctrl_Z
-		&& ptr == line_start + 1))
+    if (!error && !got_int && linerest != 0)
     {
 	// remember for when writing
 	if (set_options)
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -2278,6 +2278,7 @@ failed:
 	    && !got_int
 	    && linerest != 0
 	    && !(!curbuf->b_p_bin
+		&& curbuf->b_p_fixeol
 		&& fileformat == EOL_DOS
 		&& *line_start == Ctrl_Z
 		&& ptr == line_start + 1))


brammool:
> I have changed the remove of `CTRL-Z` behavior.
> It is desired behavior?

The patch as it currently is changes behavior without any way for the
user to keep the old behavior.

Keeping or dropping the trailing CTRL-Z is very similar to keeping a
missing EOL at the end of the file.  How about this:

- Add an option that indicates a trailing CTRL-Z was found. 'endoffile'?
  The CTRL-Z is not added to the buffer (no change from before).
- When 'fixeol' is on (the default) we keep the current behavior, the
  CTRL-Z is not written.  This is the most common case, there are hardly
  apps that need the CTRL-Z, while many can't handle it.
- When 'fixeol' is off, then write the CTRL-Z if 'endoffile' is set.
  Users that care about keeping the end-of-file characters the same will
  have to reset 'fixeol', which is simple enough.

I don't think there is a need for a separate 'fixendoffile' option, just
using 'fixeol' for both EOL and CTRL-Z makes sense.

The user can reset 'endoffile', like 'endofline' can be set.

-- 
"I've been teaching myself to play the piano for about 5 years and now write
most of my songs on it, mainly because I can never find any paper."
		Jeff Lynne, ELO's greatest hits

Shougo:
I think the behavior is better.  But the problem is it adds the complexity.

0cjs:
I understand the idea of wanting to preserve the old behaviour as a default, and have no objection to that in general.

>  Keeping or dropping the trailing CTRL-Z is very similar to keeping a missing EOL at the end of the file.

The thing is, it's _orthogonal._ You can have any of the following, where  `NL` is `LF`, `CR LF` or `CR` depending on `filemode`:
* file without Ctrl-Z EOF and without final newline terminator: `xx`
* file without Ctrl-Z EOF and with final newline terminator: `xx NL`
* file with Ctrl-Z EOF and without final newline terminator: `xx ^Z`
* file with Ctrl-Z EOF and with final newline terminator: `xx NL ^Z`

I am not particularly keen on the idea of overloading `fixeol` to take care of behaviour other than whether or not a file should be allowed to have a "partial" (i.e., not terminated by a newline sequence) last lines; I think this gets a bit unintuitive and even confusing. That's why I proposed something slightly different in [this comment on #11397](https://github.com/vim/vim/issues/11397#issuecomment-1285038931).

The proposal there always keeps the Ctrl-Z it reads; I did that to avoid having another flag. However, if adding another flag is acceptable, how about just adding an `eofstrip` (or whatever name you think is good) flag that defaults to on, maintaining the current behaviour of Vim, and if turned off, does not remove the Ctrl-Z EOF on read? That alone is enough for me, but the extra behaviour I described to automatically fix the pre-EOF EOL for users adding or removing EOF manually would be a bonus.

I'd be happy to look at implementing a patch if this idea is not considered a non-starter.

brammool:
> I understand the idea of wanting to preserve the old behaviour as a default, and have no objection to that in general.
> 
> >  Keeping or dropping the trailing CTRL-Z is very similar to keeping a missing EOL at the end of the file.
> 
> The thing is, it's _orthogonal._ You can have any of the following,
> where  `NL` is `LF`, `CR LF` or `CR` depending on `filemode`:
> * file without Ctrl-Z EOF and without final newline terminator: `xx`
> * file without Ctrl-Z EOF and with final newline terminator: `xx NL`
> * file with Ctrl-Z EOF and without final newline terminator: `xx ^Z`
> * file with Ctrl-Z EOF and with final newline terminator: `xx NL ^Z`

It's been a long time, but I think only a few situations are common, and
more so today.  On Unix and the like we are used to files always being
NL terminated (extra CR optional, could even be some lines only).  A
missing NL is considered a fault by many people, which is why Vim
usually terminals the file with NL, unless you explicitily say you don't
need that.

On MS-DOS the situation was the other way around: Adding a CR-NL to the
last line might change the meaning and caused problems in a few cases.
Preserving the existig situation was preferred.  This can't be combined
with the previous, thus the 'fixeol' option was added, with the 'noeol'
option to remember what to restore.  Allowing the user to change this,
e.g. based on the file type.  Pascal might need something else than the
C compiler.

The CTRL-Z mostly caused a problem, hardly ever was it really needed.
It has come up now after so many years, thus I don't think changing
default behavior is justified.  And I do think that the one 'fixeol'
option is probably sufficient, since you can separately change whether
EOL or EOF is restored. Or neither.  You'll have to write an autocommand
anyway if you want to do something non-default.  I don't think we can do
more automatically, it depends too much on what you want to do with the
file.  Even when the filetype would use 'noeol' 'eof' (CTRL-Z terminated
without CR-LF before it) you don't want that when concatenating two
files or writing part of a file.

If you are working on one specific system I'm sure you have a certain
idea from that point of view, please keep in mind there are many more
different situations.  Let's make the common situation easy, specific
cases (such as working with old systems in an emulator) may require
extra care from the user.  Can't please everybody!


-- 
Bad programs can be written in any language.

Shougo:
It is fixed in 9.0.0831.  Close.

PR #11413, opened 2022-10-20
Only set seenModifyOtherKeys when it's safe

### LINKED ISSUE #9014: Ctrl shortcuts stop working after Ctrl-;
### Steps to reproduce

I reported this potential bug as https://groups.google.com/g/vim_dev/c/Kt1BPSgvx-s a few weeks ago. I'm reposting it here in the hopes this is the right place :)

I've had a weird bug for about a year now, where these four keyboard shortcuts would stop working eventually:

```
nnoremap <C-J> <C-W><C-J>
nnoremap <C-K> <C-W><C-K>
nnoremap <C-L> <C-W><C-L>
nnoremap <C-H> <C-W><C-H>
```

In particular, they would revert to their original function, almost as if they were never mapped in my vimrc. For example, `<C-L>` would revert to "clear", and `<C-H>` to "backspace". The only way to resolve this issue would be to restart Vim.

With the help of @myitcv, who also ran into this bug, we were able to narrow it down: it happens after we press `<C-;>`. That is, control-semicolon.

Note that some terminals do not seem to emit escape codes for `<C-;>`. For example, for xterm, one needs this extra setting in `.Xresources`:

    XTerm*vt100.modifyOtherKeys: 1

I have been able to reproduce this with xterm 369 and vim 8.2.3441. Both with only the options listed above, and no extra plugins.
With a feature like "modifyOtherKeys" disabled, the bug disappears, as it seems like xterm just sends ";" when pressing `<C-;>`.

For clarity, here's what I do to reproduce it with two sample files:

1) vim foo.txt
2) :vsp bar.txt
3) Verify that <C-H> and <C-L> work to switch back and forth
4) Press <C-;>
5) Notice that <C-H> and <C-L> no longer work to switch buffers 

### Expected behaviour

I expected `<C-;>` to not break my configured shortcuts.

I don't really mind what the behavior of that shortcut by default is, otherwise. I assume it should do nothing, given that my vimrc does not map it.

I've spent hours investigating this bug, and this is as far as I've got. I'm hoping someone in this mailing list can shed some light into it. My best guess is this is a bug in how Vim handles ANSI escape codes.

Note that if I "xterm -e dash" and press <C-;>, I get the following escape code: `^[[27;5;59~`

Here are other scenarios I've ruled out:

* Terminal emulator bug; it also reproduces with my emulator of choice, https://codeberg.org/dnkl/foot/, which implements "modifyOtherKeys".

* Issue with my vimrc; reproduced with just those four lines.

* Issue with vim running inside bash; happens with "xterm -e vim" too.

* System shortcuts; Paul and myself can reproduce on different setups.

Any help appreciated. Happy to try patches or capture debug logs. 

### Operating system

Linux 5.14.12, running Arch Linux, with Sway (wayland)

### Version of Vim

8.2.3441

### Logs and stack traces

_No response_

### LINKED ISSUE #11403: mapping stop working "randomly"
### Steps to reproduce

Hi, I use basic mapping like:

```
nmap <C-H> <C-W>h
nmap <C-J> <C-W>j
nmap <C-K> <C-W>k
nmap <C-L> <C-W>l
```
But sometimes after moving in and out of my terminal, the mapping stop working, after this <C-L> behave like there is no mapping at all. For example `<C-L>` will clear the screen for a micro-second instead of switching windows.
`<C-W>h/j/k/l` still work when my mapping stop working.

To move around my terminal windows, I use a mapping also based on ctrl and os key, I don't know if it could lead to a strange behavior regarding vim. I don't think so because the issue appears in other terminal without those terminal mappings.

My configuration is pretty normal, I use termdebug and man plugins.
Loading them in my .vimrc following:
```
runtime ftplugin/man.vim
packadd termdebug
```

:nmap show correct mapping when the bug occurs
```
n  <C-L>         <C-W>l
n  <C-K>         <C-W>k
n  <C-J>         <C-W>j
n  <C-H>         <C-W>h
```
When it appears, if I save the session using :mks and inspect the Session.vim.
The previous mapping appears early in the file, but there is also "weird" mapping 
image
image
I don't know if thoses mappings here are normal.

How can I investigate more precisely this behavior ?

### Expected behaviour

Mapping should work all the time.

### Version of Vim

9.0

### Environment

Operating system: Gentoo:
Terminal: kitty (reproduced on many terminals: foot, alacritty)
$TERM: xterm-kitty
shell: zsh
I also disabled modifyOtherKeys only for kitty terminal since it doesn't support it.
Using:
```
if &term == "xterm-kitty"
    let &t_TI = ''
    let &t_TE = ''
endif
```

### Logs and stack traces

_No response_

trygveaa:
There are certain cases where a key with modifier is seen, but keys with
the Ctrl modifier are still sent as control characters, so if
seenModifyOtherKeys is set, mappings that include the Ctrl modifier will
stop working.

There's three known cases of this:

1. The kitty terminal emulator by default, when the kitty keyboard
   protocol is not enabled by the user in the vim config, because kitty
   sends certain keys not traditionally supported by terminals as keys
   with modifiers by default and kitty doesn't support modifyOtherKeys 2.

2. The foot terminal emulator by default, because foot has
   modifyOtherKeys 1 enabled, and modifyOtherKeys 2 is not enabled
   unless done so by the user in the vim config, because vim only
   enables modifyOtherKeys 2 when TERM looks like an xterm name.

3. In a certain configuration of xterm and vim where modifyOtherKeys 1
   is enabled in xterms config, and vim is configured not to enable
   modifyOtherKeys 2. This is probably not likely to happen, but just to
   note that it is possible to encounter this in xterm too.

This patch makes sure that seenModifyOtherKeys is only set when the
control sequence to enable either kittys keyboard protocol or
modifyOtherKeys has been sent.

Fixes #9014, fixes #11403

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -4807,6 +4807,42 @@ handle_version_response(int first, int *arg, int argc, char_u *tp)
 }
 
 /*
+ * Determine if it's safe to set seenModifyOtherKeys when a sequence with key
+ * and modifier is seen. There's four known cases we see a key with modifier:
+ *  1. Kittys keyboard protocol has been enabled by a control sequence.
+ *  2. Kitty is used, but the keyboard protocol has not been enabled.
+ *  3. modifyOtherKeys 2 has been enabled by a control sequence.
+ *  4. modifyOtherKeys 1 is enabled by default in a terminal or has been enabled by a control sequence.
+ *  Only in cases 1 and 3 should seenModifyOtherKeys be set. If
+ *  seenModifyOtherKeys is set in case 2 or 4, mappings that include the Ctrl
+ *  modifier will stop working because they are still sent as control
+ *  characters.
+ */
+    static int
+should_set_seen_modify_other_keys()
+{
+    char_u  *term = T_NAME;
+    if (term == NULL || *term == NUL)	    // 'term' not defined yet
+	return FALSE;
+
+    // If we have enabled kittys keyboard protocol, we should set it.
+    if (strstr((char *)T_CTI, "\e[>1u") != NULL) {
+	return TRUE;
+    }
+    // Else if term is xterm-kitty, we should not set it because kitty doesn't
+    // support modifyOtherKeys.
+    if (STRNICMP(term, "xterm-kitty", 11) == 0) {
+	return FALSE;
+    }
+    // Else if we have enabled modifyOtherKeys 2 we should set it.
+    if (strstr((char *)T_CTI, "\e[>4;2m") != NULL) {
+	return TRUE;
+    }
+    // If we have not enabled modifyOtherKeys 2, we should not set it.
+    return FALSE;
+}
+
+/*
  * Add "key" to "buf" and return the number of bytes used.
  * Handles special keys and multi-byte characters.
  */
@@ -4848,7 +4884,9 @@ handle_key_with_modifier(
     int	    modifiers;
     char_u  string[MAX_KEY_CODE_LEN + 1];
 
-    seenModifyOtherKeys = TRUE;
+    if (!seenModifyOtherKeys && should_set_seen_modify_other_keys())
+	seenModifyOtherKeys = TRUE;
+
     if (trail == 'u')
 	key = arg[0];
     else


brammool:
> The kitty terminal emulator sends CSI u keys with the meta modifier
> and keys in the Unicode Private Use Area by default (encoded it it's
> own keyboard protocol which is similar, but not identical to
> modifyOtherKeys) and doesn't support modifyOtherKeys, so don't set
> seenModifyOtherKeys for those. Setting seenModifyOtherKeys means that
> vim don't use mappings for control characters, which kitty still use
> for Ctrl-key presses, so those would stop working.
> 
> Since the Meta modifier is not present on standard keyboards anymore,
> I think it's very rarely used, so therefore I don't see a problem with
> not setting seenModifyOtherKeys for it. It's very likely that Shift,
> Control or Alt would be used before Meta, so then seenModifyOtherKeys
> would be set by those.
> 
> Keys in the Unicode Private Use Area is not used by modifyOtherKeys as
> far as I can see, so not setting seenModifyOtherKeys for those should
> not affect terminals supporting modifyOtherKeys, only terminals
> supporting the kitty keyboard protocol.
> 
> Fixes #11403 in kitty. That issue still persists in the foot terminal
> emulator though, but that's caused by a different problem so it needs
> a different solution.

Hmm, this looks a bit arbitrary.  The Meta modifier may be used by some
terminal emulators (e.g. to represent the Cmd key on Mac), just to get
it to change what a key does.  I have no idea who would use the Unicode
Private Use area, this looks a bit too much like guessing.

How about checking $TERM for containinng "kitty"?  Since kitty doesn't
support modifyOtherKeys (and it's very unlikely it will), that is a more
direct way.  It appears the default value for $TERM is "xterm-kitty",
thus that should work.

-- 
Mynd you, m00se bites Kan be pretty nasti ...

trygveaa:
> Hmm, this looks a bit arbitrary. The Meta modifier may be used by some terminal emulators (e.g. to represent the Cmd key on Mac), just to get it to change what a key does. I have no idea who would use the Unicode Private Use area, this looks a bit too much like guessing. How about checking $TERM for containinng "kitty"? Since kitty doesn't support modifyOtherKeys (and it's very unlikely it will), that is a more direct way. It appears the default value for $TERM is "xterm-kitty", thus that should work.

I see your point. I hadn't thought about Mac using Meta to represent Cmd for instance. Usually I don't like checking TERM because it won't help if other terminals does the same, but I agree that in this case it makes sense since the issue is very specific.

I have updated the commit to check TERM against xterm-kitty. I also added a check for if the control sequence to enable modifyOtherKeys or the kitty keyboard protocol has been sent and only set `seenModifyOtherKeys` if it has. By doing this, the issue is also fixed in foot. I have updated the commit message with a description of this new behavior. I will also update the PR description with this message.

kovidgoyal:
If you mean the response to the secondary device attributes query kitty will always send
1; 4000 + primary kitty version number ; secondary kitty version number

so if the kitty version number is 1.23.45 you will get 1;4001;23
For the current kitty release you will get 
1;4000;26 since the kitty version number is 0.26.4

Incidentally the adding of 4000 was put in there precisely because of vim turning on SGR_MOUSE mode only if the number was high enough. 

And just touching briefly on our last discussion, in order to get this response, you have to send the terminal the bytes ```ESC[>c ```. To query for the kitty keyboard protocol you need to send ```ESC[?u```. Since you are not worried about the former "leaking" on terminals that do not handle it, its a bit, umm extreme, to be worrying about the latter leaking. Both are very simple CSI escape codes. If there is indeed a terminal out there that parses the former and not the latter I would be amazed.

Tachi107:
As @dnkl has been involved in issue #9014 and foot development I think it is appropriate to tag him, as GitHub PRs do not notify users subscribed to the linked issues

trygveaa:
> Checking the contents of t_TI is not reliable. The code may be there to enable the protocol, but that doesn't mean it works.

Well, yes, it's not perfect. However, when I submitted this patch seenModifyOtherKeys was set unconditionally, which is even less reliable. I think only setting seenModifyOtherKeys when we have at least tried to enable the protocol is a lot better than unconditionally setting it.

I see that you've changed it to not be set for kitty now, but the problem with foot remains without this patch.

It would of course be better to know for sure if modifyOtherKeys 2 is enabled, but from what I've seen in the issue comments, there isn't any way to do this? If there was, we could just use that and we wouldn't need to set seenModifyOtherKeys when a key with a modifier is seen at all as far as I understand.

Even if this is not the perfect solution, I don't think the idea of a perfect solution should get in the way of a partial solution which improves the current situation. Issue #9014 has been open for a year, and this patch addresses it. There is a possibility that there exist a terminal with this problem and that it isn't fixed by this patch. However, this patch fixes it for the two problems we know about, which is with kitty and foot, and it doesn't make the situation worse for any other terminal.

By the way, it would be better to set seenModifyOtherKeys for kitty when the kitty keyboard protocol is enabled, like this patch does, because it would make it possible to bind Ctrl-i and Tab as separate keys if the user has set t_TI to enable the kitty keyboard protocol.

brammool:
> > Checking the contents of t_TI is not reliable. The code may be there
> > to enable the protocol, but that doesn't mean it works.
> 
> Well, yes, it's not perfect. However, when I submitted this patch
> seenModifyOtherKeys was set unconditionally, which is even less
> reliable. I think only setting seenModifyOtherKeys when we have at
> least tried to enable the protocol is a lot better than
> unconditionally setting it.

Not true, unconditionally makes it predictable.  You never know what
state a terminal is in, the user can send raw sequences, there are
menus, config files, etc.  Believe me, we had quite a few trial and
error solutions before we ended up with the current one.  Maybe not
perfect, but definitely better than other solutions.

> I see that you've changed it to not be set for kitty now, but the
> problem with foot remains without this patch.

Yeah, can't possibly handle every bad behaving terminal.  Let's be clear
that Vim existed for many, many years, new terminals should make sure
they work with Vim, not the other way around.

> It would of course be better to know for sure if modifyOtherKeys 2 is
> enabled, but from what I've seen in the issue comments, there isn't
> any way to do this?

Not that I'm aware of.  I consider any terminal having only the "1"
version enabled as broken, we can't possibly handle those.

> If there was, we could just use that and we
> wouldn't need to set seenModifyOtherKeys when a key with a modifier is
> seen at all as far as I understand.

The terminal configuration stuff is totally inadequate.  I had some
ideas to make it work, but nothing that would be good enough to push
forward.  I'm also not aware of anybody else having a solution (other
than individual terminal solutions that won't ever be implemented
widely).  Unless something is going to happen, we might add an entry in
the list of key codes requested (this works for xterm and has a chance
that others might implement it too).  It's a bit of a hack, but might be
the only way out.  See the req_more_codes_from_term() and
got_code_from_term() functions.

> Even if this is not the perfect solution, I don't think the idea of a
> perfect solution should get in the way of a partial solution which
> improves the current situation.

It does not improve the situation.  It only makes it more complicated.

> Issue #9014 has been open for a year,
> and this patch addresses it. There is a possibility that there exist a
> terminal with this problem and that it isn't fixed by this patch.
> However, this patch fixes it for the two problems we know about, which
> is with kitty and foot, and it doesn't make the situation worse for
> any other terminal.

It only solves it for a very specific situation and I doubt there are no
side effects.  Usually these things seem OK until it's committed and
then people who were not paying attention speak up that it breaks
something for them.

> By the way, it would be better to set seenModifyOtherKeys for kitty
> when the kitty keyboard protocol is enabled, like this patch does,
> because it would make it possible to bind Ctrl-i and Tab as separate
> keys if the user has set t_TI to enable the kitty keyboard protocol.

I would think seenModifyOtherKeys will automatically be set if kitty
sends these keys.  Why wouldn't it?

-- 
Q: Why do Norwegians excel at text editing?
A: Their ancestors are Vi-kings

brammool:
I have just submitted patch 9.0.0930 which adds the 'keyprotocol' option.
It's not fully implemented yet, but goes a long way in the right direction.

I have tried to find other ways to detect that the terminal has modifyOtherKeys level 2 enabled,
but it appears there is no such way.  The trick in this PR to inspect the t_TI termcap entry for
an escape sequence is unreliable, the terminal may have modifyOtherKeys enabled in another
way, and this check would break things, without providing the user with any way to work around it.

For those terminals that support the kitty keyboard protocol and enable it there should not be
a problem, because requesting the state will trigger a response that they kitty keyboard protocol
is enabled, which would then reset seenModifyOtherKeys (not implemented yet though).

Another change that might help is to set seenModifyOtherKeys only for a sequence that the kitty
protocol does not send: {lead}27;{modifier};{key}~.  With xterm it is possible to change to the
other form {lead}{key};{modifier}u, but I doubt it is actually used in practice.  And we can tell
users to simply not do that.

I'll await comments on patch 9.0.0930 and make the other changes mentioned.  After that please
check if you still see a problem.  If so then open a new issue or PR.  I'm closing this one now,
since it is outdated.

PR #11419, opened 2022-10-21
Use `setl lisp` in runtime/ftplugin/dune.vim

ddickstein:
Use `setl lisp` instead of `set lisp` in runtime/ftplugin/dune.vim so other buffers aren't affected.

### DIFF
--- a/runtime/ftplugin/dune.vim
+++ b/runtime/ftplugin/dune.vim
@@ -11,7 +11,7 @@ if exists("b:did_ftplugin")
 endif
 let b:did_ftplugin=1
 
-set lisp
+setl lisp
 
 " Comment string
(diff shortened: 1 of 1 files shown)


dkearns:
This was fixed some time ago in the [vim-ocaml repo](https://github.com/ocaml/vim-ocaml/blame/master/ftplugin/dune.vim#L14).  Maybe you can nudge them to send an update here.

dkearns:
I made a related commit yesterday but didn't notice this was still a problem.

Could you please add a "Last Change:" header entry noting that it was updated by yourself and then we can twist @chrisbra's arm to include it.

These bad option settings from ftplugins are *very* annoying.

PR #11421, opened 2022-10-22
Add g:html_{no_links,no_doc,no_modeline}

arp242:
g:html_no_links disables transforming URL-like text to `<a>` tags.

g:html_no_doc disables generating a full HTML document, and outputs only the actual code inside a `<pre>` or `<div>` tag. This is useful if you use 2html to transform several files.

g:html_no_modeline disables the foldmethod=manual modeline.


arp242:
This patch appears larger than it actually is due to the indentation changes; best to view with `git diff -w` or ignoring whitespace in the GitHub diff viewer (https://github.com/vim/vim/pull/11421/files?diff=unified&w=1)

brammool:
I'll include it.

PR #11422, opened 2022-10-22
Add Vim ftplugin file for Apache .conf files

### LINKED ISSUE #11417: Buffer with Apache configuration file has wrong commentstring on `filetype plugin on`
### Steps to reproduce

`set nocompatible`
`filetype plugin on`

In a buffer: `set ft=apache` and then `set commentstring`.
This shows `commentstring=/*%s*/`

Possible fix: https://github.com/dubgeiser/vim/commit/aa6bddb14a2407d8791c001434f9904401d86538

There are commenting-plugins that rely on the value of `commentstring`, for instance https://github.com/numToStr/Comment.nvim/issues/264 or https://github.com/tpope/vim-commentary/blob/master/README.markdown?plain=1#L35

It might be a good thing to include this?

### Expected behaviour

`commentstring=# %s`

### Version of Vim

9.0.217

### Environment

MacOSx
Kitty (xterm-kitty)
Bash 5.2.2(1)-release

### Logs and stack traces

_No response_

dubgeiser:
Since Apache configuration files only use `#` for comments, the default values for `comments` and `commentstring` are not completely correct.

Closes #11417

### DIFF
--- a/runtime/ftplugin/apache.vim
+++ b/runtime/ftplugin/apache.vim
@@ -1,0 +1,16 @@
+" Vim filetype plugin
+" Language:	apache configuration file
+" Maintainer:	Per Juchtmans <dubgeiser+vimNOSPAM@gmail.com>
+" Last Change:	2022 Oct 22
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+setlocal comments=:#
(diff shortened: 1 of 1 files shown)


brammool:
Thanks, I'll also update CODEOWNERS

PR #11428, opened 2022-10-23
Fix typos

zeertzjq:


dpelle:
Thanks. Since the patch only fixes typos, consider looking at https://fossies.org/linux/test/vim-master.9c50eeb.221022.tar.gz/codespell.html which shows more typos.

zeertzjq:
Closing in favor of #11432

Issue #11373, opened 2022-10-15
 Vim with termguicolors in Windows Terminal outputs wrong color


3N4N:
### Steps to reproduce

```
vim --clean +"set ls=2 tgc"
```

### Expected behaviour

In the old conhost terminal, the colors are output properly. Same thing should happen in Windows Terminal.

[![vim in conhost][1]][1]

[1]: https://i.stack.imgur.com/fUNtg.png

### Actual Behavior

But in Windows Terminal (v1.16.2524.0), the colors with `:h termguicolors` enabled is incorrect.

[![vim in Windows Terminal][2]][2]

  [2]: https://i.stack.imgur.com/G6XAq.png

### Version of Vim

9.0.648

### Environment

Operating system: Windows 11 version 21H2 (OS Build 22000.1098)
Terminal: Windows Terminal v1.16.2642.0 with Solarized Light colorscheme (comes with the terminal by default)
Value of $TERM: nil (also tried with 'xterm-256colors')
Shell: cmd.exe (also tried with powershell)

3N4N:
@chrisbra sorry for pinging you but if you remember: I opened question in vi.SE about this bug and you confirmed that you could reproduce it. I tried today and the colors are displayed properly now. If you could try it once more to see if it got fixed on your end, too, then this issue can be closed (although what fixed it is unknown).

zewpo:
This is really interesting.   

What you are seeing is the default windows console foreground color, non-intense white, which is MS Windows defaults, however only appropriate for a dark background.    

When Vim goes into tgc,  it requests the appropriate rgb background and foreground colors from the host console api - before applying anything set (if any) from the user configured colorscheme or highlight groups.    

In the old Command Prompt console you can only configure fg and bg colors by selecting from one of the values from the colortable, ie, one of the 16 confgured rgb colors.   However, in the new WT,  fg and bg colors can be configured independantly to any rgb value you want - they don't have to be selected from one of the 16 rgb colors of the colortable. 

The old way of querying which index of the colortable to use for bg and fg, seems to be inappropriate for the new WT.   I'm trying to see how to query the new conpty api for the correct colors, it doesnt seem properly documented in windows api docos - but its all opensource on github, so I'm having a trawl through that. 

![image](https://user-images.githubusercontent.com/10286211/197650459-9ff1be84-bab6-4753-936e-573b497057c8.png)

[...]

PR #11431, opened 2022-10-23
Fixed 'restorescreen' on exit for Windows Terminal

zewpo:
Problem: Vim does not restore the screen buffer corectly on exit in the new Windows Terminal
Solution: Make use of the USE_WT flag

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -1601,7 +1601,7 @@ decode_mouse_event(
     static void
 mch_set_cursor_shape(int thickness)
 {
-    if (USE_VTP || is_conpty_stable())
+    if (USE_VTP || CONPTY_STABLE)
     {
 	if (*T_CSI == NUL)
 	{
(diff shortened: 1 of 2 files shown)


zewpo:
I got no idea what CI tests are failing - is it related to my patch?

k-takata:
No. It seems that v9.0.0815 is broken.

zewpo:
Problem: Windows Terminal had flaky color control, see issue #11373 
Solution: CONPTY is now stable, no longer need USE_WT flag.

src/os_win32.c
```diff
@@ -334,7 +330,7 @@ read_console_input(
 
     if (s_dwMax == 0)
     {
-	if (!USE_WT && nLength == -1)
+	if (!CONPTY_STABLE && nLength == -1)
```
zewpo, line 333: not sure why this? workaround for buggy WT in win 10?   I think its no longer an issue in Win11, so I cant test this line on my pc at the moment.

PR #11436, opened 2022-10-24
Error in screen dump for Test_smoothscroll_one_long_line.

errael:
The corrected screen dump puts the cursor on the 1st screen line, the current dump erroneously has the cursor on the 2nd screen line. So the test seems to pass, but in fact it fails. This is related to https://groups.google.com/g/vim_dev/c/YkQbUsC-pek/m/WeyQk4lhAwAJ

BTW, if I do `call term_dumpload("Test_smooth_one_long_2.dump")` the cursor is shown in the wrong spot. I'm guessing this is an issue in `term_dumpload`, but this area is not well known (or even poorly known) to me. The first line on the screen is
```
<<<ts of text with lots of text with lot
```
The `term_dumpload` shows the cursor on column 0; but it should be on column 3, the first visible text character on that line.

### DIFF
--- a/src/testdir/dumps/Test_smooth_one_long_2.dump
+++ b/src/testdir/dumps/Test_smooth_one_long_2.dump
@@ -1,6 +1,6 @@
-|<+0#4040ff13#ffffff0@2|t+0#0000000&|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t
->s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f
+|<+0#4040ff13#ffffff0@2>t+0#0000000&|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t
+|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f| |t|e|x|t| |w|i|t|h| |l|o|t|s| |o|f
 | |t|e|x|t| @34
+@40
 |~+0#4040ff13&| @38
-|~| @38
-| +0#0000000&@21|1|,|8|1| @9|A|l@1| 
+| +0#0000000&@21|1|,|4@1| @9|A|l@1| 


brammool:
> The corrected screen dump puts the cursor on the 1st screen line, the
> current dump erroneously has the cursor on the 2nd screen line. So the
> test seems to pass, but in fact it fails. This is related to
> https://groups.google.com/g/vim_dev/c/YkQbUsC-pek/m/WeyQk4lhAwAJ
> 
> BTW, if I do `call
> term_dumpload(&quot;Test_smooth_one_long_2.dump&quot;)` the cursor is
> shown in the wrong spot. I&#39;m guessing this is an issue in
> `term_dumpload`, but this area is not well known (or even poorly
> known) to me. The first line on the screen is
> ```
> <<<ts of text with lots of text with lot
> ```
> The `term_dumpload` shows the cursor on column 0; but it should be on
> column 3, the first visible text character on that line.

I tried to make it so that when using CTRL-E the cursor moves to a
visible part of the text, but as little as possible.  With 'scrolloff'
zero it might end up on the showbreak or "<<<", then moving it a bit
right seemed the best solution. But one screen line down might be
better.  And when 'scrolloff' is non-zero it gets much more complicated.
Try moving the cursor left-right after some CTRL-E and CTRL-Y, or using
"gj" and "gk".  I do think we can ignore 'scrolljump' here, it's not
important when 'smoothscroll' is set.

I was trying to tackle one part at a time, but there are conflicting
requirements, might need  more wholistic approach.  And good test
coverage (many options interfere, such as 'showbreak').

-- 
OLD WOMAN: King of the WHO?
ARTHUR:    The Britons.
OLD WOMAN: Who are the Britons?

errael:
>> On 22/10/27 11:45 AM, Bram Moolenaar wrote:
>>>> (got my system back from the shop; so good)
>>>>
>>>> Given: "xxx" is a file with a single line covering 4 screen lines
>>>> Do:
>>>>
>>>>      gvim -u NONE -U NONE xxx
>>>>      :set smoothscroll
>>>>      Ctrl-E
>>>>
>>>> Note that cursor starts at beginning of line, column 0.
>>>>
>>>> After entering Ctrl-E, observe cursor is on 2nd screen line column 0.
>>>>
>>>> It should be on the 1st screen line column 3, just after "<<<".
>>> I thought this was nicer
>>>
>> I see. It also happens when "showbreak=+++", in that case skipping a 
>> screen line doesn't look right to me.
>
>You mean putting the cursor below the +++, instead of right of it?
>Since the scrolling is up-down I think it's logical that the cursor is
>moved down, not right.  Isn't that happening?  Please give a repro
>example if needed.

Moved over to github from mailing list, to attach some images. This PR is about the cursor ending up on the 2nd screen line after CTRL-E so it fits here.

Here's an example with one long line

Before CTRL-E:
ctrl-e-1
After CTRL-E:
ctrl-e-2

ychin:
I filed #11504 which will fix this.

PR #11448, opened 2022-10-27
horizontal mouse scroll in terminal

zewpo:
Problem:  Horizontal mouse wheel scroll code was dependent on gui, but does not actually need gui, as it works in terminal already if features compiled with gui.  Vertical scrolling was already independent of gui.

Solution:  Move the horizontal mouse wheel scroll code into same dependencies as vertical scrolling.


zewpo (commented):
restore readme to keep git diff clean

zewpo (commented):
OK, ready for review.  Will have to fix mappings in a separate PR, this one is useful and ready to go as is.

brammool:
Error reported by CI: os_win32.c(1264): error C2086: 'win_T *wp': redefinition

zewpo (commented):
Thanks for alerting me @brammool , I've fixed the merge conflict.

brammool:
Instead of having a global "longest_lnum" used in two files, better make it static in one file and use a function update_longest_lnum() to update the value and use get_longest_lnum() to get the value.

It looks like currently for the GUI longest_lnum is updated quite often, called from gui_write() after displaying something.
When not using the GUI, it is probably good to reset longest_lnum when we're not sure it is valid.
That might be ui_write(), but it basically means it won't be used.  Which may work just fine anyway, since horizontal
scrolling is either not used and there is no reason to update longest_lnum, or a horizontal scroll happens, which already has enough overhead to not be much impacted by recomputing the longest_lnum every time.
Not sure, perhaps the longest_lnum variable can just be dropped completely and the code simplified dramatically.

zewpo:
I like the idea of simplifying things.  I will be away a few days, then I can look at this properly next week.

brammool:
Please let us know when you are done with what you planned to do.  You could mark this PR as "draft" for the time being.

zewpo (commented):
Simplified code around getting longest_lnum.

brammool:
> @zewpo commented on this pull request.
> 
> Simplified code around getting longest_lnum.

Let me do some more cleanup and I'll merge it.  Then please check and we
can make additional changes if needed.
I wish we could add some testing for this...

-- 
75. You start wondering whether you could actually upgrade your brain
    with a Pentium Pro microprocessor 80.  The upgrade works just fine.

zewpo:
> @brammool 
> I wish we could add some testing for this...

w.r.t. testing.  Yes, working on that.    

My end goal of this mouse stuff was to make sure that scroll-wheel mappings also work properly with modifier keys, when used in windows.  Which is not quite working right yet.  At the moment, only basic mappings work.   I expect to write test scripts around all of this - want to make sure its all covered.

PR #11451, opened 2022-10-27
New configuration options from current OpenSSH 

Jakuje:


brammool:
Thanks, I'll include it.  Next time please also update the header.

PR #11465, opened 2022-10-29
Test_fileformats() errors

ant0sha:
(those are probably not catched by CI because no assertion is failed, however they confuse one a lot if test is running interatively)

bwipe causes come back at some point to the buffer with source code of test itself; content of that buffer is reinterpreted with ff=mac, due to this file content looks like junk (one long string with lot of ^J) - leads to modeline confusion which blames about "expandtab trailing characters"

### DIFF
--- a/src/testdir/test_fileformat.vim
+++ b/src/testdir/test_fileformat.vim
@@ -94,6 +94,8 @@ func Test_fileformats()
   call assert_equal("mac\rmac\r\n", s:file2str('Xtest'))
   bwipe XXUnix XXDos XXMac
 
+  new
+
   " try with 'fileformat' set to 'dos'
   set fileformat=dos
   e! XXUnix
@@ -106,6 +108,8 @@ func Test_fileformats()
   w! Xtest
   call assert_equal("mac\rmac\r\r\n", s:file2str('Xtest'))
   bwipe XXUnix XXDos XXMac
+
+  new
 
   " try with 'fileformat' set to 'mac'
   set fileformat=mac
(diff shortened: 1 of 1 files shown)


brammool:
This opens lots of new windows.  How about using ":enew!" instead?  Or have "new" followed by ":only!".

ant0sha:
> This opens lots of new windows.

It only seems so, in reality the new window is always closed (due to individual ":bwipe" calls in every section)

> How about using ":enew!" instead? Or have "new" followed by ":only!".

This will not help. After ":enew!" or "new"+":only!" there will be again comeback to original background buffer, which _may_ contain some modeline expressions, which will force vim to blame. But what will work is ":only! | :bwipe" right in the beginning. I will try to re-make the PR...

brammool:
Looks like some CI failures that should have been fixed already, did you sync to head?

How about doing it this way:
```
  only!
  bwipe!
  enew
```
The :%bwipe! command is quite drastic, I try to avoid it.

zeertzjq:
Isn't this already done before each test?
https://github.com/vim/vim/blob/cf2594fbf34d9a6776bd9d33f845cb8ceb1e1cd0/src/testdir/runtest.vim#L224-L226<>

brammool:
The version I have locally changes the "new" command to:
```
  only!
  bwipe!
  enew
```
And so far that appears to work well.  Any reason to do otherwiswe?

brammool:
The changes were included in a slightly different way, and there was no response on the comment.
Therefore closing.  If you still see confusing errors please create a new issue or PR.

PR #11467, opened 2022-10-29
checking for 'stl' error is done all over the place

luukvbaal:
Problem:	Error checking for 'statusline'-format options is done all over the place.
Solution:	Do this locally in build_stl_str_hl() instead by passing in the option name.

This irked me while working on the 'numberformat' feature. Not sure if this is a welcome change but I thought I'd propose it. Similar refactor is done for the `use_sandbox()` calls in `win_redr_custom()`.

Currently, for some options `did_emsg` is checked while for others `called_emsg` is checked. The proposed change checks for `did_emsg` in all cases, which seemed like a harmless change to me. It doesn't seem to cause any regressions in the test suite at least.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -3959,20 +3959,9 @@ maketitle(void)
 	{
 #ifdef FEAT_STL_OPT
 	    if (stl_syntax & STL_IN_TITLE)
-	    {
-		int	use_sandbox = FALSE;
-		int	called_emsg_before = called_emsg;
-
-# ifdef FEAT_EVAL
-		use_sandbox = was_set_insecurely((char_u *)"titlestring", 0);
-# endif
-		build_stl_str_hl(curwin, title_str, sizeof(buf),
-					      p_titlestring, use_sandbox,
-					      0, maxlen, NULL, NULL);
-		if (called_emsg > called_emsg_before)
-		    set_string_option_direct((char_u *)"titlestring", -1,
-					   (char_u *)"", OPT_FREE, SID_ERROR);
-	    }
+		build_stl_str_hl(curwin, title_str, sizeof(buf), p_titlestring,
+					    (char_u *)"titlestring", 0,
+						0, maxlen, NULL, NULL);
 	    else
 #endif
 		title_str = p_titlestring;
@@ -4090,20 +4079,8 @@ maketitle(void)
 	{
 #ifdef FEAT_STL_OPT
 	    if (stl_syntax & STL_IN_ICON)
-	    {
-		int	use_sandbox = FALSE;
-		int	called_emsg_before = called_emsg;
-
-# ifdef FEAT_EVAL
-		use_sandbox = was_set_insecurely((char_u *)"iconstring", 0);
-# endif
-		build_stl_str_hl(curwin, icon_str, sizeof(buf),
-						    p_iconstring, use_sandbox,
-						    0, 0, NULL, NULL);
-		if (called_emsg > called_emsg_before)
-		    set_string_option_direct((char_u *)"iconstring", -1,
-					   (char_u *)"", OPT_FREE, SID_ERROR);
-	    }
+		build_stl_str_hl(curwin, icon_str, sizeof(buf), p_iconstring,
+				 (char_u *)"iconstring", 0, 0, 0, NULL, NULL);
 	    else
 #endif
 		icon_str = p_iconstring;
@@ -4228,7 +4205,8 @@ build_stl_str_hl(
     char_u	*out,		// buffer to write into != NameBuff
     size_t	outlen,		// length of out[]
     char_u	*fmt,
-    int		use_sandbox UNUSED, // "fmt" was set insecurely, use sandbox
+    char_u	*opt_name,      // option name corresponding to "fmt"
+    int		scope,		// scope corresponding to "opt_name"
     int		fillchar,
     int		maxwidth,
     stl_hlrec_T **hltab,	// return: HL attributes (can be NULL)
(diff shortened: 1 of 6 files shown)


brammool:
> Problem:	Error checking for 'statusline'-format options is done all over the place.
> Solution:	Do this locally in build_stl_str_hl() instead by passing in the option name.
> 
> This irked me while working on the 'numberformat' feature. Not sure if
> this is a welcome change but I thought I'd propose it. Similar
> refactor for the `use_sandbox()` call in `win_redr_custom()`.
> 
> Currently, for some options `did_emsg` is checked while for others
> `called_emsg` is checked. The proposed change checks for `did_emsg` in
> all cases, which seemed like a harmless change to me. It doesn't seem
> to cause any regressions in the test suite anyhow.

Not everything is covered by tests, I would prefer a straight refactor
and avoid making subtle changes that might be OK, but we can't be 100%
sure.  The difference between did_emsg and called_emsg happens when
using exceptions and also when running tests (where exceptions are
caught, while executing the code normally would not do that), it's quite
tricky.

This line:
	use_sandbox = OPT_LOCAL;
Is using the use_sandbox confusingly, better add a new variable for the
2nd argument of was_set_insecurely().

The new opt_name argument deserves a comment.

This line:    if (did_emsg != save_did_emsg) {
Should probably use ">" instead of "!=".  It should actually reset
did_emsg before doing the work, otherwise a previous error would cause
the option to be reset even when it's working OK.  Then you can just
check if did_emsg got set.

The scope should be passed in, guessing it inside build_stl_str_hl() is
brittle.

In win_redr_status() you removed saved_did_emsg, but did_emsg is still
reset, that isn't right.  I think you can remove the line that resets
did_emsg, it should be handled inside win_redr_custom() completely.


-- 
15. Five days in advance, tell your friends you can't attend their
    party because you're not in the mood.

luukvbaal:
Hey, nice to see you're catching up with PRs again. Hope you're doing well.

> This line:
	use_sandbox = OPT_LOCAL;
Is using the use_sandbox confusingly, better add a new variable for the 2nd argument of was_set_insecurely().

> The new opt_name argument deserves a comment.

> This line:    if (did_emsg != save_did_emsg) {
Should probably use ">" instead of "!=". 

> The scope should be passed in, guessing it inside build_stl_str_hl() is brittle.

Done.

> In win_redr_status() you removed saved_did_emsg, but did_emsg is still reset, that isn't right.  I think you can remove the line that resets did_emsg, 

Right, that seems to have been an oversight in my porting this change from the neovim source. Removed.

> it should be handled inside win_redr_custom() completely.

I think you mean `build_stl_str_hl()`, at least that's what would make sense in the context of this refactor (to me).

> I would prefer a straight refactor and avoid making subtle changes that might be OK, but we can't be 100% sure.

I thought all of these checks were actually checking for errors in a potential `%{}` eval block. Simply saving the current did_emsg and checking that it got bigger at the end of `build_stl_str_hl()` would catch that in my mind. If not, I'll leave the details of checking/(resetting) the `emsg` var up to you, if you want to include it. I'm not sure what a straight refactor would entail, so I don't think I'll be taking that on.

luukvbaal:
Moved the `was_set_insecurely()` call to `build_stl_str_hl()` as well, no longer needing the `use_sandbox` argument.

brammool:
> > I would prefer a straight refactor and avoid making subtle changes
> > that might be OK, but we can't be 100% sure.
> 
> I thought all of these checks were actually checking for errors in a
> potential `%{}` eval block. Simply saving the current did_emsg and
> checking that it got bigger at the end of `build_stl_str_hl()` would
> catch that in my mind. If not, I'll leave the details of
> checking/(resetting) the `emsg` var up to you, if you want to include
> it. I'm not sure what a straight refactor would entail, so I don't
> think I'll be taking that on.

The difference between did_emsg and called_emsg is subtle and it's easy
to not see what happens.  Mainly because of the complicated way
exceptions are handled (which means an error isn't reported right away
but later, after returning out of nested calls an commands).  Since
tests usually catch exceptions, so as to detect reported errors,
behavior in tests is often slightly different from normal usage.  With
care this can be made to work, but it requires trying out things and
checking the tests actually check what happens with normal use.  That is
tricky, which is why it is better not do some cleanup when refactoring.

This even got more complicated with Vim9 script, where after an error we
usually abort executing commands.  In legacy script Vim would happily
continue trying to do stuff even though a previous command failed (so
you could end up making changes to the wrong file). Now did_emsg might
not be set (because of exception handling) but we do want to stop
execution, thus need to use another flag.  I've been trying to come up
with a generic way, but keep running into corner cases.


-- 
I am also told that there is a logical proof out there somewhere
that demonstrates that there is no task which duct tape cannot handle.
					-- Paul Brannan

luukvbaal:
Right, I understand the complications. However this was just meant as a cleanup patch. Here, we just want to make sure that the 'statusline'-format options are reset when an error occurs, to avoid redraw issues. I _think_ this patch still handles that correctly, do as you see fit.

brammool:
> Right, I understand the complications. However this was just meant as
> a cleanup patch. Here, we just want to make sure that the
> 'statusline'-format options are reset when an error occurs, to avoid
> redraw issues. I _think_ this patch still handles that correctly, do
> as you see fit.

Well, most existing code was using called_emsg, but when I tried to use
that in the updated code tests are failing.  I went back to using
did_emsg and that works for the tests.  I left a TODO comment, because
I'm not sure it is actually right.  Let's see if anybody complains.
Theoretically using did_emsg should be better, but it's impossible to
predict all situations where it matters.

Let's consider this done for now.

-- 
`The Guide says there is an art to flying,' said Ford, `or at least a
knack. The knack lies in learning how to throw yourself at the ground
and miss.' He smiled weakly.

PR #11473, opened 2022-10-30
Termdebug: allow disabling the window toolbar

### LINKED ISSUE #11442: Termdebug: allow disabling the "window toolbar"
### Please describe the problem.

Currently Termdebug automatically adds the window toolbar if mouse is enabled. However I don't want this behavior since the winbar clutters the screen.

The way I'm currently coping with this is via clearing `mouse`, opening `:Termdebug` and then re-enabling `mouse=a` in a function. (I don't use mouse much, but I like to keep it enabled just in case).

### Describe the solution you'd like

Instead of assuming mouse being enabled means the user wants the window toolbar, add an option (let's say `g:termdebug_no_winbar`) which allows the user to explicitly disable the winbar.

N-R-K:
Problem:    The termdebug plugin assumes mouse being enabled means the user wants the window toolbar, which may not be correct.
Solution:   Add an option to explicitly disable the window toolbar (closes #11442)

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1537,6 +1537,13 @@ If you don't want this then disable it with: >
 or if there is no g:termdebug_config: >
 	let g:termdebug_popup = 0
 
+Window toolbar ~
+							*termdebug_winbar*
+
+By default the Termdebug plugin creates a window toolbar if mouse is enabled.
+If you don't want this then disable it with: >
+	let g:termdebug_config['winbar'] = 0
+
 
 Vim window width ~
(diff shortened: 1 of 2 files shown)


brammool:
I'll include it, thanks.

N-R-K:
Not to be pushy, but I don't think this change (or anything equivalent) made it into master.
Have you perhaps forgotten about it?

brammool:
Will be in the next runtime files update.

N-R-K:
> Will be in the next runtime files update.

I see. Wasn't familiar with the workflow here, sorry for the inconvenience.

PR #8077, opened 2021-04-06
Add SF Symbols to double-wide emoji list for macOS

ychin:
Only do this for macOS because they would not render correctly outside of Apple ecosystem anyway as these are private symbols and not standardized under Unicode.

PR #11474, opened 2022-10-31
Add SF Symbols 4 Unicode values to double-width table (macOS)

ychin:
With the release of macOS 13 Ventura, Apple has also released SF Symbols 4 which contains new symbols to be used by apps. Just as before, these symbols also have corresponding Unicode values by using the Supplmentary Private Use Area-B range. Add the new symbols' Unicode values to the double-width ranges so Vim will show them properly. (In reality some of the symbols like the ones for cars like 'suv.side' / U+00101800 are quite long and go longer than double-width, but this is a decent enough compromise since Vim doesn't have triple-width characters, and I don't think any of them are narrow enough to use single-width)

### DIFF
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -1565,10 +1565,10 @@ utf_char2cells(int c)
 	{0x1f6f3, 0x1f6f3}
 
 #ifdef MACOS_X
-	// Include SF Symbols characters, which should be rendered as
+	// Include SF Symbols 4 characters, which should be rendered as
 	// double-width. All of them are in the Supplementary Private Use
-	// Area-B range. The exact range was determined by downloading the "SF
-	// Symbols" app from Apple, and then selecting all symbols, copying
+	// Area-B range. The exact range was determined by downloading the 'SF
+	// Symbols 4' app from Apple, and then selecting all symbols, copying
 	// them out, and inspecting the unicode values of them.
 	, {0x100000, 0x1018c7}
 #endif
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -1566,10 +1566,26 @@ utf_char2cells(int c)
 
 #ifdef MACOS_X
 	// Include SF Symbols 4 characters, which should be rendered as
-	// double-width. All of them are in the Supplementary Private Use
-	// Area-B range. The exact range was determined by downloading the 'SF
-	// Symbols 4' app from Apple, and then selecting all symbols, copying
-	// them out, and inspecting the unicode values of them.
+	// double-width.  SF Symbols is an Apple-specific set of symbols and
+	// icons for use in Apple operating systems.  They are included as
+	// glyphs as part of the default San Francisco fonts shipped with
+	// macOS.  The symbols are released in a versioned manner, and the
+	// latest version is SF Symbols 4.
+	//
+	// These Apple-specific glyphs are not part of standard Unicode, and
+	// all of them are in the Supplementary Private Use Area-B range. The
+	// exact range was determined by downloading the 'SF Symbols 4' app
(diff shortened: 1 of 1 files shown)


ychin:
Just for reference #8077 was where we added the SF Symbols 1-3 Unicode codes.

brammool:
> With the release of macOS 13 Ventura, Apple has also released SF Symbols
> 4 which contains new symbols to be used by apps. Just as before, these
> symbols also have corresponding Unicode values by using the Supplmentary
> Private Use Area-B range. Add the new symbols' Unicode values to the
> double-width ranges so Vim will show them properly. (In reality some of
> the symbols like the ones for cars like 'suv.side' / U+00101800
> are quite long and go longer than double-width, but this is a decent
> enough compromise since Vim doesn't have triple-width characters,
> and I don't think any of them are narrow enough to use single-width)

Where does that value 0x1018c7 come from?  The comment above it should
be more explicit, since the value changed but the comment didn't.

-- 
14. Put mosquito netting around your work area. Play a tape of jungle
    sounds all day.

ychin:
I updated the comments to explicitly say SF Symbols **4** now.

But I essentially retrieved that value the same way as before. Downloaded SF Symbols 4 app from Apple, selected all symbols, pasted it to Vim and then sorted them. Go to the last one and inspected the hex value using `ga`.

Unfortunately Apple never really published the exact Unicode ranges they intend to reserve for SF Symbols, so I don't know if there's a fixed upper-bound that we can use here.

brammool:
> I updated the comments to explicitly say SF Symbols **4** now.
> 
> But I essentially retrieved that value the same way as before.
> Downloaded SF Symbols 4 app from Apple, selected all symbols, pasted
> it to Vim and then sorted them. Go to the last one and inspected the
> hex value using `ga`.
> 
> Unfortunately Apple never really published the exact Unicode ranges
> they intend to reserve for SF Symbols, so I don't know if there's a
> fixed upper-bound that we can use here.

OK, it's a little better.  However, I don't have any clue what "SF
Symbols 4" means.  Is there a short explanation or a web page we can
link to?  I did find this page: https://developer.apple.com/sf-symbols/
We could add that and something like "symbols defined by Apple to
integrate with their system font"?

We should probably also mention that these are not really double-width
characters, but it's the best way Vim can handle them.

-- 
In many of the more relaxed civilizations on the Outer Eastern Rim of the
Galaxy, "The Hitchhiker's Guide to the Galaxy" has already supplanted the
great "Encyclopedia Galactica" as the standard repository of all knowledge
and wisdom, for though it has many omissions and contains much that is
apocryphal, or at least wildly inaccurate, it scores over the older, more
pedestrian work in two important respects.
First, it is slightly cheaper; and second, it has the words "DON'T PANIC"
inscribed in large friendly letters on its cover.

ychin:
Right. SF Symbols is basically a bunch of symbols Apple released as part of their OSes. "SF Symbols 4" is the version 4 of them, which comes with new customization features and also new symbols (each symbol has a unique name, and a unique Unicode code point). Apple usually releases a new version of SF Symbols with new OS releases (iOS/macOS/etc), which are released annually around September to October.

These symbols can either be used in a system image API where you can load the image by name (e.g. "gear.2"), or you can just use them in text mixed with other Unicode text, similar to emojis (this is where this PR comes in, by making sure we treat them as double-width in text). The link you found is indeed the official page, which sadly is light on details. I found the Wikipedia article at least a little more descriptive (https://en.wikipedia.org/wiki/San_Francisco_(sans-serif_typeface)#SF_Symbols).

Also updated the code comments to be a little verbose. Feel free to let me know if you want more edits on them.

PR #11481, opened 2022-11-01
Put CursorLineFold before CursorLineSign in syntax.txt

zeertzjq:
Because FoldColumn also comes before SignColumn


brammool:
I'll include it.

PR #11485, opened 2022-11-01
win32: Recognize VSCMD_ARG_TGT_ARCH environment variable

k-takata:
When "Developer Command Prompt for VS 2017" is selected on Windows Terminal, the `VSCMD_ARG_TGT_ARCH` environment variable is set, but the `Platform` environment variable is not set.

Recognize `VSCMD_ARG_TGT_ARCH`.

### DIFF
--- a/src/Make_mvc.mak
+++ b/src/Make_mvc.mak
@@ -222,8 +222,12 @@ CPU = i386
 !  endif
 ! else  # !CPU
 CPU = i386
-!  if !defined(PLATFORM) && defined(TARGET_CPU)
+!  ifndef PLATFORM
+!   ifdef TARGET_CPU
 PLATFORM = $(TARGET_CPU)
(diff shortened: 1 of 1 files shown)


zewpo:
This is good.  Its elegent.  I've tested this in my setup, and it works perfectly.  

It makes my powershell script a lot more similar to the bat script.
It saves me from having to include this in all my powershell scripts:  
`$env:PLATFORM='x64'`

An equivalent of this is not needed in bat scripts, but MS changed how the vcvars works in powershell, it no longer sets that everironment variable automatically.

I think that `VSCMD_ARG_TGT_ARCH` is idiomatically more correct anyway than simply `PLATFORM`.     Because PLATFORM is ambiguous really, dont know if it is referring to target or current environment.  I think MS did the right thing by changing this behavior.

PR #4049, opened 2019-02-28
Add str2blob, blob2str

mattn:
This is useful to gt string from blob. Also blob from string.

PR #11486, opened 2022-11-01
Fix 'eof' option

k-takata:
Fix the bugs that Ctrl-Z is inserted on top of the file and 'eof' is not correctly detected.

Add tests for 4 combinations of 'eol' and 'eof'.

### DIFF
--- a/src/bufwrite.c
+++ b/src/bufwrite.c
@@ -2050,10 +2050,6 @@ restore_backup:
 		len = 0;
 		write_info.bw_start_lnum = lnum;
 	    }
-	    if (!buf->b_p_fixeol && buf->b_p_eof)
-		// write trailing CTRL-Z
-		(void)write_eintr(write_info.bw_fd, "\x1a", 1);
-
 	    // write failed or last line has no EOL: stop here
 	    if (end == 0
 		    || (lnum == end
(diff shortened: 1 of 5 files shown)


src/testdir/test_fixeol.vim
```diff
@@ -48,4 +48,71 @@ func Test_fixeol()
   enew!
 endfunc
 
+func Test_eof()
+  let data = 0z68656c6c6f.0d0a.776f726c64   " "hello\r\nworld"
```
k-takata, line 52: It should be nice if we could use `str2blob()` here.

k-takata: cf. #4049

src/bufwrite.c
```diff
@@ -2158,6 +2154,13 @@ buf_write(
 	    nchars += len;
 	}
 
+	if (!buf->b_p_fixeol && buf->b_p_eof)
+	{
+	    // write trailing CTRL-Z
+	    (void)write_eintr(write_info.bw_fd, "\x1a", 1);
+	    nchars++;
+	}
```
k-takata, line 2162: Should we restrict this only for `fileformat == EOL_DOS`?

brammool: We are dealing with corner cases here, not really trying to catch common uses cases.  I think it's best to give the user ways to control what is written as much is possible.  It seems that adding a restriction to writing CTRL-Z depending on 'fileformat' makes it more complicated.
How about writing the help for these options, can be a separate section that each of the option entries link to.  "If you want files in Unix format (every line NL terminated:...", "If you want to preserve the fileformat exactly as-is, including final CTRL-Z: ...", "If you want to use an old MS-DOS file in a modern environment, fixing line endings and dropping CTRL-Z: ...".  In that style it should make clear how the options work together.

k-takata: I've updated the document. Could you review it?

PR #11492, opened 2022-11-03
Fix smoothscroll cursor calculations bugs when 'number' is set

ychin:
The 'smoothscroll' feature currently doesn't work well when 'number' (or signs) is set. If you scroll using `<C-E>` when the cursor is on the first line, the cursor would end up at a wrong position, and there's also a disconnect between the window column calculation (`wincol()`) and the actual buffer position (`col('.')`). Commands like `g0` and `g$` would also start to behave incorrectly.

This was a regression caused by v9.0.0746 (856c5d2bc7). That commit actually fixed this issue for Vi compatibility mode `set cpo+=n`, but broke the normal nocompatible case. Fixed the logic to work in both case by correctly subtracting from `width2` instead of the entire window's width. This makes the calculation consistent with how the "long line wrapping" code below calculates the overflow wcol calculation.

For tests, fixed up an existing screen dump test (the old dump was actually demonstrating the incorrect behavior, but it is quite hard to tell IMO, because `term_dumpload` does *not* show the cursor position in a visual way and therefore it's hard to visually inspect). Also, add new tests to confirm that the cursor position is calculated correctly and we don't have a desync between `col('.')` and `wincol()`. Don't use screen dump tests for these because they are testing specific logic instead of an end-to-end tests like screen dumps, and asserts are clearer and easier to specify this way.

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -1114,13 +1114,15 @@ curs_columns(
 		&& curwin->w_skipcol > 0
 		&& curwin->w_wcol >= curwin->w_skipcol)
 	{
-	    // w_skipcol excludes win_col_off().  Include it here, since w_wcol
-	    // counts actual screen columns.
+	    // Deduct by multiples of width2.  This allows the long line
+	    // wrapping formula below to correctly calculate the w_wcol value
+	    // when we are wrapped.
 	    if (curwin->w_skipcol <= width1)
-		curwin->w_wcol -= curwin->w_width;
+		curwin->w_wcol -= width2;
 	    else
-		curwin->w_wcol -= curwin->w_width
+		curwin->w_wcol -= width2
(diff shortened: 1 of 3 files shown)


ychin:
If you want to reproduce the bug, simply do the following:

1. Make a file `test.txt` that have a long line, enough to wrap 4-5 screen lines.
2. `vim --clean -c "set scrolloff=0" -c "set smoothscroll" -c "set wrap" -c "set number" test.txt`
3. Do `<C-E>`, note that the cursor position is wrong.
4. Do `g0` and `g$` and note that those commands are also wrong.

See this clip for demo:

[asciicast](https://asciinema.org/a/534455)

ychin:
Seems like the Windows gVim tests are failing in the channel tests, but I didn't touch that part at all, so I'm not sure whether this is just some existing flake or not.

PR #11496, opened 2022-11-04
configure.ac: Fix -Wimplicit-int

thesamesam:
Clang 16 makes -Wimplicit-int an error by default. Fixes errors like:
```
error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Werror,-Wimplicit-int]
```

We already use proper declarations with every other test anyway, so let's be consistent.

Signed-off-by: Sam James <sam@gentoo.org>


thesamesam:
Thanks!

PR #11500, opened 2022-11-04
Fix bugs around assigning functions

mityu:
**Description**

This PR fixes two things:

1. Functions are still able to assign variables whose name doesn't start with capital letter.

Reproduce steps:

- Open Vim with `vim --clean`
- Put the following script on a buffer
- Do `:%source`

```vim
let g:cos = 'foo'
let g:.cos = {-> 42}  " Should be error.
echo cos(0)
" => 42
```

2. Inconsistent behavior in trying to assign function to `g:` dictionary with `extend()`.

Reproduce steps:

- Open Vim with `vim --clean`
- Put the first script in below on a buffer
- Do `:%source`
- Do the same thing for the second script.

```vim
let g:foo = 'foo'
call extend(g:, {'foo': {-> 42}}, 'keep')  " No error.
```

```vim
let g:foo = 'foo'
call extend(g:, {'foo': function('function')}, 'keep')
"=> E704: Funcref variable name must start with a capital: foo
```

I believe the latter case shouldn't throw an error like the former case since `g:foo` variable is not modified in the both case.

**Environment**

- OS: macOS Monterey
- Terminal: wezterm 20220905-102802-7d4b8249
- Vim: v9.0.828


zeertzjq:
This seems mostly the same as #11468

mityu:
Oh sorry, I didn't noticed that. As you say, this patch is mostly same to #11468. Closing.

mityu:
@zeertzjq It seems that #11468 doesn't have tests for the first case.  Could you add an test for it in your patch?

```diff
diff --git a/src/testdir/test_let.vim b/src/testdir/test_let.vim
index 3f9bdd4db..4e4a386df 100644
--- a/src/testdir/test_let.vim
+++ b/src/testdir/test_let.vim
@@ -316,6 +316,7 @@ func Test_let_errors()
   call assert_fails('let l += 2', 'E734:')
   call assert_fails('let g:["a;b"] = 10', 'E461:')
   call assert_fails('let g:.min = function("max")', 'E704:')
+  call assert_fails('let g:cos = "" | let g:.cos = {-> 42}', 'E704:')
   if has('channel')
     let ch = test_null_channel()
     call assert_fails('let ch += 1', 'E734:')

```

zeertzjq:
Added.

mityu:
Thank you!

PR #11502, opened 2022-11-04
Fix smoothscroll long line move cursor down out of bounds bug

ychin:
With smoothscroll set, if you have a really long line that spans more than the entire window when wrapped, if you use `j` to move cursor one or more lines, the cursor will end up being out of bounds.

This change fixes it by properly calculating how many out-of-bounds lines we have when scrolling down (in `scroll_cursor_bot()`). We need to scroll that many lines in addition to the existing ones. That's because when `smoothscroll` is on, `scrollup()` starts to only scroll by screen lines and not buffer lines, so we need to account for how many extra screen lines to scroll just to get past the hump of the current line, so to speak.

If the current line is only barely long enough to span a little more than the window size, the new cursor position will be at the bottom. If the current line is really long, Vim will use the "center cursor" logic, but because it doesn't really know how to handle smoothscroll right now, it will just place the cursor at the top of teh screen instead. This is similar to `zz` and `zb` right now as those commands are also not smoothscroll-aware. A problem for another time to fix.

Also, add regression tests by extending the existing `Test_smoothscroll_one_long_line()`.

Edit: Found a bug when using `zz`. Debugging and converted to draft issue.


ychin:
For repro of the bug / demo, see the following. You can clearly se the cursor going out of bounds. I mapped j/k to use `:echowindow` to print the command to make it clearer what is pressed (asciinema doesn't have a "show input" mode right now):

[asciicast](https://asciinema.org/a/2lcbVi41qMZ21YwILvBkD81tj)

With the fix, it looks good now:

[asciicast](https://asciinema.org/a/STLTUIlG7X9OvTXTYmwOg7O1V)

---

Note that I feel like there should probably be some checks to make sure the cursor never goes out of bounds as a failsafe, but maybe all it will do is masking bugs like this making it hard to detect (since a cursor going out of bounds is very glaringly obvious).

ychin:
I think I found a small bug in this change (doing `zz` seems to do weird things). Debugging. Converting issue to draft for now. I think the other smoothscroll PRs should still be good to go.

ychin:
Ok on further testing, what I found was a general bug in `zt` and `zz` regarding smoothscroll, and unrelated to my change here. I'll submit separate PR to fix it.

brammool:
CI reports failing tests - is this PR ready to include or does it require some work?

ychin:
It should be good to go. Let me take a look at the failing tests and get back to you in a bit.

ychin:
I looked through the CI issue and seems like it was mostly failed builds etc. Just rebased and pushed, but to answer the question, this PR should be good to go. The issue I mentioned (`zz` being buggy) is an existing smoothscroll issue that I a working on for a new PR.

PR #11504, opened 2022-11-05
Smoothscroll only move 2 lines down if necessary / fix off-by-one bug

ychin:
**Note**: This depends on #11492 and uses that commit as a base. I can rebase after that PR is merged. Right now, ignore the first commit which is identical to #11492.

**Update**: The dependent PR is now merged in v9.0.0893 (01ee52ba). I have rebased this pull request so this is a single commit now. Also fixed up `src/testdir/dumps/Test_smooth_long_showbreak_2.dump` which was added in v9.0.0898 (75ac25b4) to regression test this.

Off-by-one bug:
-----

Currently smoothscroll has an off-by-one bug in in `update_topline()`, where it does a `curwin->w_skipcol + 3 >= vcol` check. This check is supposed to make sure whenever the character is on the '<<<' marker, Vim will scroll more to show the characters, but the off-by-one bug makes it apply to the fourth column as well. This results in 2 bugs I could see.  1st bug is if you move to the fourth column of a smooth-scrolled wrapped line, even though it's not obscured by <<<, Vim will still try to scroll up to show more. Another bug is if you do <C-E> on the fourth column, sometimes the logic will get stuck and refuses to scroll more because this off-by-one logic is not agreeing with other parts of the code.

Only move 2 lines down when smooth scrolling if necessary:
-----

Currently, if Vim thinks <C-E> will result in the cursor being obscured by the <<< smoothscroll marker, it moves the cursor twice to avoid that.  The problem is it doesn't account for settings like `number` (which adds a padding on the ledge so the <<< wouldn't obscure the cursor) or `showbreak` (where we don't draw <<< to begin with), or when `scrolloff` is greater than 1. This is annoying becaue we are unnecessary moving down by more than we need. Fix the logic to correctly determine how much is obscured by the <<< marker and only moves twice *if* necessary.

Also see #11436 and https://groups.google.com/g/vim_dev/c/YkQbUsC-pek/m/lHZmRVGnAgAJ.

-----

Also, added regression tests to cover both cases described above.


ychin:
For off-by-one bug, I already described how to reproduce this, and the test case is updated to test for regression, but below is an asciinema recording to demonstrate:

This recording show that you literally cannot <C-E> to scroll in certain situations when doing it at the 4th column:

[asciicast](https://asciinema.org/a/W2Bu8PfpGiXUpqpS7zSMJAohA)

This recording shows that Vim unnecessary moves up when going to the 4th column, even though it's not obscured by the "<<<":

[asciicast](https://asciinema.org/a/r9tcS5t9OjM7qd7unATPxg2Kl)

This recording shows the issues being fixed:

[asciicast](https://asciinema.org/a/GzQ8wo8pHscqOvXhatJofQ2to)

ychin:
Only move 2 lines down when smooth scrolling if necessary:

Screencast showing what it looks like with the change (also encoded in the test cases):

[asciicast](https://asciinema.org/a/Pg0vZ5pbccd0p9Dv1nFcBpcva)

ychin:
**Update**: The dependent PR is now merged in v9.0.0893 (01ee52ba). I have rebased this pull request so this is a single commit now. Also fixed up `src/testdir/dumps/Test_smooth_long_showbreak_2.dump` which was added in v9.0.0898 (75ac25b4) to regression test this.

PR #11505, opened 2022-11-05
Fix argdedupe not cleaning different paths to same files

### LINKED ISSUE #9402: argdedupe does not dedupe different paths that point to the same file
### Steps to reproduce

1. Open vim
2. :args <file_a>
3. :argadd <file_a_but_different_path>
4. :argdedupe
5. :args

### Expected behaviour

I expect after running the command argdedupe that all duplicates will be cleaned

But it seems argdedupe does not remove duplicates of same files but having different paths (Example: Using the absolute path one time and the other time using the relative path)

### Version of Vim

9.0.831

### Environment

OS: Windows 10

### Logs and stack traces

_No response_

nir9:
Solution: While checking for duplicate paths, first expand to absolute paths before comparing

closes #9402

### DIFF
--- a/src/arglist.c
+++ b/src/arglist.c
@@ -782,10 +782,22 @@ ex_argdedupe(exarg_T *eap UNUSED)
 {
     int i;
     int j;
+    char_u* firstFullname;
+    char_u* secondFullname;
+    int areNamesDuplicate;
 
     for (i = 0; i < ARGCOUNT; ++i)
 	for (j = i + 1; j < ARGCOUNT; ++j)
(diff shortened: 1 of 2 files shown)


brammool:
Please add a test for what this is fixing.

The firstFullname is created and freed many more times than needed.  Move it a block up, so it only needs to be done once for each argument.  And move the vim_free() call to the end of that block.

The added variables can be declared in the block where they are used.

nir9:
@brammool Added a test and optimized according to your suggestions

Issue #3328, opened 2018-08-14
patch for  broken runtime vim80/compiler/pylint.vim, it has no file to check


te987:
diff --git a/pylint.vim.backup b/pylint.vim
index 93079ce..512d60b 100644
--- a/pylint.vim.backup
+++ b/pylint.vim
@@ -12,5 +12,5 @@ if exists(":CompilerSet") != 2		" older Vim always used :setlocal
   command -nargs=* CompilerSet setlocal <args>
 endif
 
-CompilerSet makeprg=pylint\ --output-format=text\ --msg-template=\"{path}:{line}:{column}:{C}:\ [{symbol}]\ {msg}\"\ --reports=no
+CompilerSet makeprg=pylint\ --output-format=text\ --msg-template=\"{path}:{line}:{column}:{C}:\ [{symbol}]\ {msg}\"\ --reports=no\ %
 CompilerSet errorformat=%A%f:%l:%c:%t:\ %m,%A%f:%l:\ %m,%A%f:(%l):\ %m,%-Z%p^%.%#,%-G%.%#

brammool:
> diff --git a/pylint.vim.backup b/pylint.vim
> index 93079ce..512d60b 100644
> --- a/pylint.vim.backup
> +++ b/pylint.vim
> @@ -12,5 +12,5 @@ if exists(":CompilerSet") != 2		" older Vim always used :setlocal
>    command -nargs=* CompilerSet setlocal <args>
>  endif
>  
> -CompilerSet makeprg=pylint\ --output-format=text\ --msg-template=\"{path}:{line}:{column}:{C}:\ [{symbol}]\ {msg}\"\ --reports=no
> +CompilerSet makeprg=pylint\ --output-format=text\ --msg-template=\"{path}:{line}:{column}:{C}:\ [{symbol}]\ {msg}\"\ --reports=no\ %
>  CompilerSet errorformat=%A%f:%l:%c:%t:\ %m,%A%f:%l:\ %m,%A%f:(%l):\ %m,%-Z%p^%.%#,%-G%.%#

Doesn't this prevent the user from typing ":make other.py"?
If one does want to lint the current file, typing ":make %" is easy.

-- 
LETTERS TO THE EDITOR (The Times of London)

Dear Sir,

I am firmly opposed to the spread of microchips either to the home or
to the office.  We have more than enough of them foisted upon us in
public places.  They are a disgusting Americanism, and can only result
in the farmers being forced to grow smaller potatoes, which in turn
[...]

PR #11509, opened 2022-11-06
Fix shellcheck compiler for shell scripts

balki:
Problem: :compiler shellcheck does not work
Solution: Add ' %' to makeprg to pass current file to shellcheck

### DIFF
--- a/runtime/compiler/shellcheck.vim
+++ b/runtime/compiler/shellcheck.vim
@@ -15,7 +15,7 @@ endif
 let s:cpo_save = &cpo
 set cpo&vim
 
-CompilerSet makeprg=shellcheck\ -f\ gcc
+CompilerSet makeprg=shellcheck\ -f\ gcc\ %
 CompilerSet errorformat=%f:%l:%c:\ %trror:\ %m\ [SC%n],
 		       \%f:%l:%c:\ %tarning:\ %m\ [SC%n],
 		       \%f:%l:%c:\ %tote:\ %m\ [SC%n],


balki:
@dkearns Please approve. Currently it is broken when running `:make`

```
No files specified.

Usage: shellcheck [OPTIONS...] FILES...
  -a                  --check-sourced            Include warnings from sourced files
  -C[WHEN]            --color[=WHEN]             Use color (auto, always, never)
  -i CODE1,CODE2..    --include=CODE1,CODE2..    Consider only given types of warnings
...
```

dkearns:
Thanks for the report, however, leaving the file argument empty is actually intentional.  Doing so offers the flexibility to specify
files other than the current buffer when running `:make`.  This might make more sense if you consider setting the options globally with `:compiler!.`

As you may be aware you can change the behaviour for yourself by adding the line in your patch to `~/.vim/after/compiler/shellcheck.vim`.

There's some relevant discussion in this issue: https://github.com/vim/vim/issues/3328

vim-ml:
On 2022-11-05, Balki wrote:
> Problem: :compiler shellcheck does not work
> Solution: Add ' %' to makeprg to pass current file to shellcheck

The 'makeprg' set by compiler files generally does not include %.
If the user wishes to make the current file, then they supply the
% themselves.

Regards,
Gary

balki:
Thank you for clarifying. Makes sense. Just realized it is very helpful to type `:make **/*sh` to see errors from all scripts in a project.

PR #11510, opened 2022-11-06
Fix flaky Test_no_unknown_error_after_error

ychin:
In MacVim's CI, `Test_no_unknown_error_after_error` in `test_vim9_script.vim` would occasionally fail, with an error message saying the following:

    Expected 'E1012:' but got 'E684: List index out of range: 0':

The reason for that is that this test uses a job which triggers both an exit and an out callback, and relies on the out_cb being called first before the exit_cb. From the docs for exit_cb, sometimes we can have buffered data, which means the exit_cb could have finished before out_cb instead. Fix the test to not rely on either finishing before the other, as it's trying to make sure we don't get an E1099 (unknown error) anyway.

### DIFF
--- a/src/testdir/test_vim9_script.vim
+++ b/src/testdir/test_vim9_script.vim
@@ -4087,7 +4087,7 @@ def Test_no_unknown_error_after_error()
       endfor
   END
   writefile(lines, 'Xdef', 'D')
-  assert_fails('so Xdef', ['E684:', 'E1012:'])
+  assert_fails('so Xdef', ['E684:\|E1012:', 'E1012:\|E684:'])
 enddef
 
 def InvokeNormal()


ychin:
Here's an example of a flaky failure: https://github.com/macvim-dev/macvim/actions/runs/3402920131/jobs/5659104444. The test is failing with:

> command line..script /Users/runner/work/macvim/macvim/src/testdir/runtest.vim[533]..function RunTheTest[52]..Test_no_unknown_error_after_error line 33: Expected 'E1012:' but got 'E684: List index out of range: 0': so Xdef

The fix I made should make this more tolerant of this issue. I did make a local change and confirmed that it's capable of testing regression still:

Diff:

```diff
diff --git a/src/ex_docmd.c b/src/ex_docmd.c
index f5b398046..425fe58fa 100644
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -743,9 +743,6 @@ do_cmdline(
      * cancel the whole command line, and any if/endif or loop.
      * If force_abort is set, we cancel everything.
      */
-#ifdef FEAT_EVAL
-    did_emsg_cumul += did_emsg;
-#endif
     did_emsg = FALSE;
 
     /*
@@ -778,9 +775,6 @@ do_cmdline(
 #endif
 							)
 	{
-#ifdef FEAT_EVAL
-	    did_emsg_cumul += did_emsg;
-#endif
 	    did_emsg = FALSE;
 	}
```

Re-running the test (with my changes in this PR) still allows us to catch errors:

```
1 FAILED:
Found errors in Test_no_unknown_error_after_error():
command line..script /Users/ychin/Dev/external/vim/src/testdir/runtest.vim[515]..function RunTheTest[52]..Test_no_unknown_error_after_error line 33: Expected 'E1012:\\|E684:' but got 'E1099: Unknown error while executing <SNR>271_Exit_cb': so Xdef
SKIPPED Test_vim9_comment_gui(): cannot start the GUI
make[1]: *** [test_vim9_script] Error 1
make: *** [test_vim9_script] Error 2
```

PR #11514, opened 2022-11-07
Fix smoothscroll mouse click position wrong bug

ychin:
When 'smoothscroll' is on, mouse clicking is off when clicking on any lines below the top line. Fix the logic to properly accounting for the clipped rows.

### DIFF
--- a/src/mouse.c
+++ b/src/mouse.c
@@ -2979,11 +2979,11 @@ mouse_comp_pos(
 		    row -= win->w_topfill;
 		else
 		    row -= diff_check_fill(win, lnum);
-		count = plines_win_nofill(win, lnum, TRUE);
+		count = plines_win_nofill(win, lnum, FALSE);
 	    }
 	    else
 #endif
-		count = plines_win(win, lnum, TRUE);
+		count = plines_win(win, lnum, FALSE);
 	    if (plines_cache != NULL && cache_idx < Rows)
 		plines_cache[cache_idx] = count;
 	}
--- a/src/mouse.c
+++ b/src/mouse.c
@@ -2987,6 +2987,21 @@ mouse_comp_pos(
 	    if (plines_cache != NULL && cache_idx < Rows)
 		plines_cache[cache_idx] = count;
 	}
+
+	if (win->w_skipcol > 0 && lnum == win->w_topline)
+	{
+	    // Adjust for 'smoothscroll' clipping the top physical lines
+	    int skip_lines = 0;
+	    int width1 = win->w_width - win_col_off(win);
+	    int width2 = width1 + win_col_off2(win);
+	    // similar formula is used in curs_columns()
+	    if (win->w_skipcol > width1)
+		skip_lines += (win->w_skipcol - width1) / width2 + 1;
+	    else if (win->w_skipcol > 0)
+		skip_lines = 1;
+	    count -= skip_lines;
+	}
+
 	if (count > row)
 	    break;	// Position is in this buffer line.
 #ifdef FEAT_FOLDING
@@ -3009,7 +3024,8 @@ mouse_comp_pos(
 	    col = off;
 	col += row * (win->w_width - off);
 	// add skip column (for long wrapping line)
-	col += win->w_skipcol;
+	if (lnum == win->w_topline)
+	    col += win->w_skipcol;
     }
 
     if (!win->w_p_wrap)
--- a/src/testdir/test_scroll_opt.vim
+++ b/src/testdir/test_scroll_opt.vim
@@ -355,6 +355,15 @@ func Test_smoothscroll_mouse_pos()
   call s:check_mouse_click(3, 1, 3, 1)
   call s:check_mouse_click(3, 6, 3, 6)
 
+  " Make a new first line 11 physical lines tall so it's taller than window
+  " height, to test overflow calculations with really long lines wrapping.
+  normal gg
+  call setline(1, "12345678901234567890"->repeat(11))
+  exe "normal 6\<C-E>"
+  call s:check_mouse_click(5, 1, 1, 201)
+  call s:check_mouse_click(6, 1, 2, 1)
+  call s:check_mouse_click(7, 1, 3, 1)
+
   let &mouse = save_mouse
   let &term = save_term
   let &ttymouse = save_ttymouse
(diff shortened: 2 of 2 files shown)


src/mouse.c
```diff
@@ -2987,6 +2987,21 @@ mouse_comp_pos(
...
 	}
+
+	if (win->w_skipcol > 0 && lnum == win->w_topline)
+	{
+	    // Adjust for 'smoothscroll' clipping the top physical lines
+	    int skip_lines = 0;
+	    int width1 = win->w_width - win_col_off(win);
+	    int width2 = width1 + win_col_off2(win);
+	    // similar formula is used in curs_columns()
+	    if (win->w_skipcol > width1)
```
ychin, line 2998: I'm actually contemplating whether I should refactor this logic here. I have seen it in a lot of places now: `scroll_cursor_bot()` (in my pull request #11502), `curs_columns()`, `validate_cursor_col()`, and now this function `mouse_comp_pos()`. It's only a few lines but we have to repeatedly copy this logic over.

There are two variants though. One is for calculating "physical line from w_skipcols", and the other is for calculating "physical line from w_wcols", so need to think about what the parameters should be.

ychin: To be clear I meant the logic to do this:

```c
	    int width1 = win->w_width - win_col_off(win);
 	    int width2 = width1 + win_col_off2(win);
 	    // similar formula is used in curs_columns()
 	    if (win->w_skipcol > width1)
 		skip_lines += (win->w_skipcol - width1) / width2 + 1;
 	    else if (win->w_skipcol > 0)
 		skip_lines = 1;
```

brammool:
> @ychin commented on this pull request.
> 
> > @@ -2987,6 +2987,21 @@ mouse_comp_pos(
>  	    if (plines_cache != NULL && cache_idx < Rows)
>  		plines_cache[cache_idx] = count;
>  	}
> +
> +	if (win->w_skipcol > 0 && lnum == win->w_topline)
> +	{
> +	    // Adjust for 'smoothscroll' clipping the top physical lines
> +	    int skip_lines = 0;
> +	    int width1 = win->w_width - win_col_off(win);
> +	    int width2 = width1 + win_col_off2(win);
> +	    // similar formula is used in curs_columns()
> +	    if (win->w_skipcol > width1)
> 
> I'm actually contemplating whether I should refactor this logic here.
> I have seen it in a lot of places now: `scroll_cursor_bot()` (in my
> pull request #11502), `curs_columns()`, `validate_cursor_col()`, and
> now this function `mouse_comp_pos()`. It's only a few lines but we
> have to repeatedly copy this logic over.
> 
> There are two variants though. One is for calculating "physical line
> from w_skipcols", and the other is for calculating "physical line from
> w_wcols", so need to think about what the parameters should be.

It's probably good to put logic in a common function, but make sure that
the goal is actually the same.  A good check is always that you can
put a comment above the function that states exactly what it's for.

With these adjustments it often depends on what triggered the
computation: did the user move the cursor or did the user cause a
scroll?  In the first case the scrolling must be adjusted for where the
cursor is, in the second case the cursor must be adjusted to be in a
valid location after the scrolling.  And then 'scrolloff' applies, need
to make sure that when using "h" or "l" next the scroll is recomputed
differently.

-- 
47. You are so familiar with the WWW that you find the search engines useless.

ychin:
Right. I can try to do some of that refactor but with so many outstanding smoothscroll pull requests out I think I prefer to do it later or only do it for the last one to avoid having to change all the previous commits. I have one more to submit (currently `zb`/`zt` are kind of broken regarding `smoothscroll` in multiple ways) but after that I don't think I have more.

brammool:
> Right. I can try to do some of that refactor but with so many
> outstanding smoothscroll pull requests out I think I prefer to do it
> later or only do it for the last one to avoid having to change all the
> previous commits. I have one more to submit (currently `zb`/`zt` are
> kind of broken regarding `smoothscroll` in multiple ways) but after that
> I don't think I have more.

I'm going to work through the smoothscroll issues and PRs soon, one at a
time.  I expect there will be quite a few more corner cases to take care
of, hopefully we find some generic solution.  But there are some user
expectations depending on whether the scroll causes the cursor to move
or moving the cursor causes a scroll.  And 'scrolloff' makes that
complicated.  We'll iterate and make sure heading in the direction of a
solution, not flipping back and forth between problems.

If you have a preferred ordering of including the PRs, let me know.
Otherwise I'll just start with the oldest first.

-- 
83. Batteries in the TV remote now last for months.

ychin:
> If you have a preferred ordering of including the PRs, let me know.

@brammool Yes, going through them chronologically sounds good. #11504 does depend on #11492, which works because it comes later chronologically. You just need to only look at the commit that is new, and not the first one which just came from #11492.

> But there are some user expectations depending on whether the scroll causes the cursor to move or moving the cursor causes a scroll.  And 'scrolloff' makes that complicated.  We'll iterate and make sure heading in the direction of a solution, not flipping back and forth between problems.

For sure. I believe the PRs I'm submitting should mostly be incremental fixes for obvious logic gaps, so I would be surprised if there are too many flip-flopping of "fix this feature, break that one". The issue is mostly with the spread out nature of the logic, which makes hunting them down hard. Some examples that I have found so far:

1. Any time we update `w_topline`, we need to think about whether `w_skipcol` needs to be refreshed / reset (one of the problems of `zt`/etc that I'm fixing now).
2. Usages of `PLINES_NOFILL` / etc tend to be problematic too because by default it caps the number of lines by window size, but with super long lines (more than one window height) the assumption may not work and we need to call it without capping window size instead (e.g. this PR #11514, or #11502)
3. The calculation of "number of physical rows" is spread out, and it's tricky because `width1` is different from `width2` depending on `number`, `cpo+=n`, etc. It also depends on whether you are calculating based on `w_skipcol` (starting from first column of text) or `w_wcol` (starting from first window column). This is what e.g. #11492 is fixing.

> I'm going to work through the smoothscroll issues and PRs soon, one at a time.

Sure, take your time! I definitely understand not wanting to cause regressions with these core move.c changes. My motive was to mostly fix up all the obvious and glaring bugs that I found from using it so I could feel comfortable turning on the feature to use it myself.

ychin:
Also note: Other than the "#11504 dependd on #11492" bit, the other PRs are independent and designed to be individually cherry-pickable so you don't *have* to do them in chronological order.

PR #11516, opened 2022-11-07
MS-Win Restore screen buffer on exit

### LINKED ISSUE #11507: MS-Win Cmd Prompt buffer restore on vim exit misses eol chars.  Resize wraps wrong, change in MS-Win behavior.
### Steps to reproduce

1. open the MS-win console in MS-Win 10 or 11 (WT or legacy cmd, has same issue.)
2. resize the console, notice wrapping of text in buffer remains correct.   buffer has correct eol on each line.
3. open vim (vim saves a copy of the screen buffer)
4. exit vim  (vim restores the screen buffer)
5. resize the console, notice wrapping is messed up on resize.  buffer is missing all eol.

### Expected behaviour

Expect wrapping at eol to remain in tact. 

### Version of Vim

Vim 8 and 9.  not sure about Vim 7.

### Environment

Problem occurs both on the legacy Cmd prompt console and new WT, and on both MS-WIndows 10 and 11, 

Is not a problem in MS-Windows 7 and 8 command prompt consoles.

I suspect related to the newish MS-Win console feature "Wrap text output on resize."  Introduced in windows 10 command prompt.

image

### Logs and stack traces

_No response_

zewpo:
Problem:  Windows 10 introduced new terminal line wrapping on resize features into command prompt.  This makes it behave similar to the new WT, but not quite the same.
Solution:  Restore screen behavior checks for various WT and VTP combinations to act correctly.

### DIFF
--- a/src/testdir/test_crypt.vim
+++ b/src/testdir/test_crypt.vim
@@ -3,11 +3,20 @@
 source check.vim
 CheckFeature cryptv
 
-let s:xxd_cmd = ''
-if empty($XXDPROG) && executable('..\xxd\xxd.exe')
-  let s:xxd_cmd = '..\xxd\xxd.exe'
-elseif !empty($XXDPROG) && executable($XXDPROG)
+" Use the xxd command from:
+" 1: $XXDPROG if set and it is executable
+" 2: the ../xxd directory if the executable is found there
+if !empty($XXDPROG) && executable($XXDPROG)
   let s:xxd_cmd = $XXDPROG
+elseif executable('..\xxd\xxd.exe')
+  " we're on MS-Windows
+  let s:xxd_cmd = '..\xxd\xxd.exe'
+elseif executable('../xxd/xxd')
+  " we're on something like Unix
+  let s:xxd_cmd = '../xxd/xxd'
+else
+  " looks like xxd wasn't build (yet)
+  let s:xxd_cmd = ''
 endif
 
 func Common_head_only(text)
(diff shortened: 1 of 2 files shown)


zewpo (commented):
Ready for review.  Will see if it passes CI testing this time.
Ahh, still  Failures: From test_popup.vim:  Found errors in Test_popup_and_window_resize():

These dont fail on my pc, its hard to replicate why it fails.

zewpo:
So, using an alternate screen buffer definitely breaks one of the popup tests.   Will try and debug it on my pc, however, my difficulty is that it passes all of those tests on my pc - and not only my pc, but all the versions of windows virtual machines that I have tried (7,8,10,11_insider, and Server 2022), they pass these same tests.  It ONLY seems to fail inside this CI test environment, and I don't know how to replicate it exactly.  Any advice on how to get this to pass would be much appreciated.

k-takata:
I tried this PR and `Test_popup_and_window_resize()` failed (in the different way):

```
Found errors in Test_popup_and_window_resize():
Run 1, 18:14:51 - 18:15:02:
command line..script C:/work/vim-git/src/testdir/runtest.vim[515]..function RunTheTest[52]..Test_popup_and_window_resize[21]..WaitForAssert[2]..<SNR>4_WaitForCommon[11]..<lambda>2 line 1: Pattern '^!\\s*$' does not match ''
command line..script C:/work/vim-git/src/testdir/runtest.vim[515]..function RunTheTest[52]..Test_popup_and_window_resize[29]..WaitForAssert[2]..<SNR>4_WaitForCommon[11]..<lambda>4 line 1: Pattern '^!\\s*$' does not match '#\[  occurs 118 times]'
command line..script C:/work/vim-git/src/testdir/runtest.vim[515]..function RunTheTest[52]..Test_popup_and_window_resize line 31: Pattern '^!\\s*$' does not match ''
```

I haven't debugged this.

* Windows 10 1909 (build 18363.2274)
* Both on Windows Terminal and Command Prompt

zewpo:
thanks @k-takata   
OK, so, you have confirmed a version of windows where it fails.
I've got Win 10:  Version 10.0.19045.2193
I will try and spin up and an older version of win 10 to test it on.

zewpo:
@k-takata   I was not using winpty.   that is the difference, it fails when winpty is used.   I can finally replicate the issue... and see if it is real or just a side-effect of difference between winpty and conpty.

PR #11525, opened 2022-11-10
Find debug vim when sourcing single test file

zewpo:
Problem:  Sourcing single test scripts in DEBUG mode failed on MS-Win, because the scripts didn't find "..\vimd.exe"
Solution:   Fix test script fallback, was hardcoded to '..\vim.exe', now the fallback checks for '..\vimd.exe' if '..\vim.exe' not found.


zewpo (commented):
Hopefully helpful to others as well.

PR #11526, opened 2022-11-10
MS-Win Restore screen buffer on exit

### LINKED ISSUE #11507: MS-Win Cmd Prompt buffer restore on vim exit misses eol chars.  Resize wraps wrong, change in MS-Win behavior.

zewpo:
Moved from PR #11516 after merge conflict.

Problem: Windows 10 introduced new terminal line wrapping on resize features into command prompt. This makes it behave similar to the new WT, but not quite the same.

Solution: Restore screen behavior checks for various WT and VTP combinations to act correctly.

Hopefully Closes #11507

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -238,6 +238,7 @@ static int suppress_winsize = 1;	// don't fiddle with console
 static char_u *exe_path = NULL;
 
 static BOOL win8_or_later = FALSE;
+static BOOL win11_or_later = FALSE;
 
 /*
  * Get version number including build number
@@ -892,6 +893,10 @@ PlatformId(void)
 	if ((ovi.dwMajorVersion == 6 && ovi.dwMinorVersion >= 2)
 		|| ovi.dwMajorVersion > 6)
 	    win8_or_later = TRUE;
+
+	if ((ovi.dwMajorVersion == 10 && ovi.dwBuildNumber >= 22000)
+		|| ovi.dwMajorVersion > 10)
+	    win11_or_later = TRUE;
 
 #ifdef HAVE_ACL
 	// Enable privilege for getting or setting SACLs.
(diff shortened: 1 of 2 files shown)


zewpo (commented):
@k-takata   Not sure what defines CONPTY_STABLE_BUILD, but testing it out to see if it passes CI.

zewpo:
@k-takata   and @brammool 

I see you guys have been around this merry-go-round before me for some time.  Perhaps since windows first started offering VTP sequences in their terminals.  5+ years or so.   I can see that I'm digging up old ground.  

What do you guys think about using ConPTY in preference to winpty, at least for newer versions of windows?   Does the new ConPTY still have any problems with multi-width chars, and/or cursor/scroll/ or other issues still?  What do we have to see before can we consider ConPTY stable?

So, I'm trying to deal with a problem for this PR, it is caused when Vim is using the newer windows console alternate screen-buffer mode `"\033[?1049h"` while hosting an internal virtual Winpty terminal, that is itself running another instance of vim.   The problem goes away when using straight ConPTY for hosting the terminal session inside vim.    Winpty behaves unexpectedly in this situation.  For example, the cursor appears in the wrong place after a resize - and that is why the CI tests were failing for this PR.  I dont actually know how big a problem that really is.    But, it occurred to me that a simple solution might be to just prefer ConPTY over Winpty for newer versions of windows - say 19045 and on - IF ConPTY is good enough in newer versions.

brammool:
I have no opinion about these changes.  Let me know if there is consensus to include this or not.

PR #11533, opened 2022-11-12
test case for cmdline bang arg (`!!` caused segfault)

Shane-XB-Qian:


brammool:
Was fixed with patch 9.0.0864

PR #11537, opened 2022-11-13
Simply fix:  syntax/help - helpHeadline highlight element chararcter

tsuyoshicho:
Simply fix for help syntax:

- Add `'` to helpHeadline 

For usr_10.txt - `WHEN IT DOESN'T WORK` small caption.

https://github.com/vim/vim/blob/28c56d501352bd98472d23667bade683877cadcc/runtime/doc/usr_10.txt#L740

It does not highlighted.


brammool:
I'll include it, thanks

PR #11540, opened 2022-11-13
obl not recognized

katawful:


brammool:
I don't know why you use "obl" for the filetype, isn't "oblivion" much clearer?

katawful:
'obl' is already a poor attempt at unifying the names the community has come up with for this language. People refer to it as 'obscript', 'obse', 'oblivion script', 'oblivion'. There's no official name otherwise, and I see a lot of different ways its referred to that can cause annoyances in communication. 'obl' is an abbreviation of 'Oblivion Language', as "OB" is a common abbreviation of Oblivion.  I declared the language as 'obl', and one of the valid file extensions is also 'obl', rather than shortening to 'obl'

As well, there is at least one DSL in progress that compiles to it, which may add confusion with just having 'oblivion' as the filetype name in this PR as end users may not know the filetype specific to this language or the DSL. 'obl' as the filetype name further specifies that this filetype is specific to this language. It is roundabout, but I could not think of a more complete solution overall even if this adds some confusion on the Vim-side of things (which I hope is at least partially alleviated by the extensions I've added)

runtime/filetype.vim
```diff
@@ -1334,6 +1334,9 @@ au BufNewFile,BufRead *.nse			setf lua
 " NSIS
 au BufNewFile,BufRead *.nsi,*.nsh		setf nsis
 
+" Oblivion Language
+au BufRead,BufNewFile *.obl,*.obse,*.oblivion,*.obscript set ft=obl
```
dkearns: This should use `setf obl`.  See `:help :setfiletype` for details.

runtime/ftplugin/obl.vim
```diff
@@ -0,0 +1,58 @@
+" Vim filetype plugin file
+" Language:    Oblivion Language (obl)
+" Original Creator: Kat <katisntgood@gmail.com>
+" Maintainer:  Kat <katisntgood@gmail.com>
+" Created:     August 08, 2021
+" Last Change: 13 November 2022
+
```
dkearns, line 7: Could you please set `b:undo_ftplugin` to undo the configuration changes?

See `:help undo_ftplugin` for details and https://github.com/vim/vim/tree/master/runtime/ftplugin for examples?

runtime/indent/obl.vim
```diff
@@ -0,0 +1,52 @@
+" Vim indent file
+" Language:    Oblivion Language (obl)
+" Original Creator: Kat <katisntgood@gmail.com>
+" Maintainer:  Kat <katisntgood@gmail.com>
+" Created:     01 November 2021
+" Last Change: 13 November 2022
+
```
dkearns, line 7: Likewise, could you please set `b:undo_indent` to undo the configuration changes?

See `:help undo_indent` for details and https://github.com/vim/vim/tree/master/runtime/indent for examples?

runtime/syntax/obl.vim
```diff
@@ -0,0 +1,3354 @@
+" Vim syntax file
+" Language: Oblivion Language (obl)
+" Original Creator: Ulthar Seramis
+" Maintainer: Kat <katisntgood@gmail.com>
+" Latest Revision: 13 November 2022
+
```
dkearns, line 6: Could you please set and restore `'cpo'` appropriately to handle the line continuations in this file?

See `:help line-continuation` for details and https://github.com/search?q=repo%3Avim%2Fvim%20cpo_save&type=code for examples.

dkearns:
A quick look suggests that "obscript" is an abbreviation of "Oblivion Script" and "obse" of "Oblivion Script Extender".  Are these the exact same language or subtly different dialects?

I think the filetype name in Vim needs to be more concerned with clearly identifying it in the global language namespace rather than getting caught up in the minutiae of possible dialect differences within the game's ecosystem.

VS Code uses "tes4script(syntax)" and I didn't find anything for Emacs.

And, of course, life is hard so there is [Oblivion Programming Language](https://jweinst1.github.io/OblivionWebsite/)

dkearns:
Well if OBSE is a superset the common way to implement this is with a base filetype and a configuration variable to include the superset features or to create a second filetype sourcing the base filetype.

That would at least free you up from trying to create a unifying name and you could go with something like TES4(Script) and OBSE?

It seems from the filetype detection that you might be able to distinguish these based on file extension alone?

Just a suggestion from the bleachers.

katawful:
Almost all projects use OBSE to begin with, ones that don't no longer see any development that would find use in Vim. OBSE doesn't just add more language features, its also how developers provide easy user-interaction so it caught on almost immediately. After 16 years, and in my experience, I can't think of any projects that do not wish to use it. Some old projects even optionally use it as this is an interpreted language, so a simple condition can use OBSE if available

As well, you can't use external editors in full (like with Godot for instance), so awareness while in Vim about whether OBSE is in use or not isn't possible since there is no header format. And if you desire to use a community add-on that provides editor logs, that requires OBSE to be in use anyways and does not warn for OBSE usage. For all practical usage, OBSE is the norm for development not the other way around. It seems to me that making a filetype for the base language has no real reason to exist

brammool:
> 'obl' is already a poor attempt at unifying the names the community has
> come up with for this language. People refer to it as 'obscript',
> 'obse', 'oblivion script', 'oblivion'. There's no official name
> otherwise, and I see a lot of different ways its referred to that can
> cause annoyances in communication. 'obl' is an abbreviation of 'Oblivion
> Language', as "OB" is a common abbreviation of Oblivion.  I declared the
> language as 'obl', and one of the valid file extensions is also 'obl',
> rather than shortening to 'obl'

My problem is that when I see "obl" I have no clue whatever that means.
"oblivion" is a lot easier to recognize.  But if that doesn't cover the
contents we should not use it.

I would prefer something longer, I can't say which one would cover this
best.  "obse" doesn't sound better than "obj" to me.

> As well, there is at least one DSL in progress that compiles to it,

What do you mean with "DSL"?

> which may add confusion with just having 'oblivion' as the filetype name
> in this PR as end users may not know the filetype specific to this
> language or the DSL. 'obl' as the filetype name further specifies that
> this filetype is specific to this language. It is roundabout, but I

I still actually haven't got what you mean with "this language".  Where
can we find the documentation?

> could not think of a more complete solution overall even if this adds
> some confusion on the Vim-side of things (which I hope is at least
> partially alleviated by the extensions I've added)

-- 
64. The remote to the T.V. is missing...and you don't even care.

brammool:
> Almost all projects use OBSE to begin with, ones that don't no longer
> see any development that would find use in Vim. OBSE doesn't just add
> more language features, its also how developers provide easy
> user-interaction so it caught on almost immediately. After 16 years, and
> in my experience, I can't think of any projects that do not wish to use
> it. Some old projects even optionally use it as this is an interpreted
> language, so a simple condition can use OBSE if available

This is irrelevant.  Files lay around everywhere, old and new.

What matters is: When opening a file, can the filetype be detected
accurately?  If yes, when there are variants to a language, we should
use different names.  It's very easy to share files between different
filetype names, it is quite difficult to have one filetype name that
actually covers various variants that the user then needs to control
somehow (usually by setting global variables, which requires reading the
documentation, etc.).  It is even more of a problem when taking an old
file (old syntax) and making a new version (new syntax), since then you
can't just set global variables to what you want, every buffer needs its
own setting.

A good example is the shell syntax, which is kind of a mess, since the
files may end in ".sh" or have a line with "#!/bin/sh" but still be one
of many shell types, with different syntax.  For that we have various
global variables, which is confusing.

> As well, you can't use external editors in full (like with Godot for
> instance), so awareness while in Vim about whether OBSE is in use or not
> isn't possible since there is no header format. And if you desire to use
> a community add-on that provides editor logs, that requires OBSE to be
> in use anyways and does not warn for OBSE usage. For all practical
> usage, OBSE is the norm for development not the other way around. It
> seems to me that making a filetype for the base language has no real
> reason to exist

Are you saying that if you have one syntax file, supporting the latest
version of the language with all the extensions added, that this can be
used also for the old files, and not a single user will have a problem
with that?


-- 
63. You start using smileys in your snail mail.

dkearns:
There's https://obse.silverlock.org/ and https://github.com/llde/xOBSE/ so it does appear to be branded as OBSE.

katawful:
I can't reply to everything since I am on my phone, but OBSE only documents it's own additions to the language. The core of the language has no official documentation or reference and is strewn across various websites. Developers looking to get information types, base statements, expressions, and operators will not find them from OBSE as OBSE only provides more to them. It does not change the ones found in the core language

katawful:
> My problem is that when I see "obl" I have no clue whatever that means. "oblivion" is a lot easier to recognize. But if that doesn't cover the contents we should not use it. I would prefer something longer, I can't say which one would cover this best. "obse" doesn't sound better than "obj" to me.

This can be answered below 


> What do you mean with "DSL"?

A language that compiles to this language, like fennel compiles to Lua, but still built with needs of the original language in mind. They have been developed to get past syntax troubles that plague the language


> I still actually haven't got what you mean with "this language". Where can we find the documentation?

This is continued from the first point. There is no documentation for the core language. It is a language used in the development tools for a video game, but as far as anyone knows was never given a name by the video game developer. They gave just the tools to end users without any documentation, and over the past 16 years the community around the game has attempted to document it but almost no one involved was anything more than a fan. So documentation is sparse. I have attempted to make a reference, but that a) assumes 'OBSE' is needed and b) is where I decided that a unified name like 'obl' was needed. If you look at some of the other syntax files for other editors (VS Code and Notepad++ come to mind), they also just simply make up a filetype name that I don't see in use else where. The Vim syntax file in which I based this PR off of both called the language 'obscript' and 'obse' confusingly, but is also the norm when in communication with other developers


> This is irrelevant. Files lay around everywhere, old and new. What matters is: When opening a file, can the filetype be detected accurately? ... For that we have various global variables, which is confusing.

I believe we could come back to this if needed, but you do make good points here. Snipped for brevity

> Are you saying that if you have one syntax file, supporting the latest version of the language with all the extensions added, that this can be used also for the old files, and not a single user will have a problem with that?

'OBSE' does not override any old syntax, there are no breaking changes to the old language as that would break the game entirely. However, due to lack of access to the game's source, additions made by OBSE aren't 100% cleanly added to the old language. This does not mean pure syntactical errors, to me they are mostly semantical. An example would probably clear up a bit of confusion

First, all scripts written in this language ultimately go through the "compiler" in the dev tools. It parses the code for syntax and basic semantical errors then saves it to memory for when you save your project. So if you write in Vim you copy and paste into the editor in the dev tools

OBSE introduces fairly standard container types (all referred to as arrays in OBSE but unimportant outside of context). As this is a new type, it can't *directly* interface with features found in the core language. Particularly, this occurs with variable initialization. To get around this, OBSE instead introduces an entirely new variable init statement. Highly edited example for clarity:

```
set int_var to array[1] ; compiler/runtime error, this statement can't take this type
let int_var := array[1] ; no error
```

*The* issue with this is that you have to be aware of this no matter what with 'OBSE'. The new syntax is purely an addition, and you get errors on bad usage when saving anyways. This goes even if the user is *not* using 'OBSE'. If I were to try to save a project with the second statement above in this case, I would see a compiler error and be unable to continue before the project was saved

So to me, since 'OBSE' does not overwrite old syntax in anyway (as it would break the game) and since the compiler you are forced to use reports on if you used incompatible syntax or semantics, it does not seem useful to directly delineate the 2 within Vim. I hope that I've cleared up confusion, it is a rather confusing topic to begin with for the reasons I've stated in the beginning

brammool:
> > My problem is that when I see "obl" I have no clue whatever that
> > means. "oblivion" is a lot easier to recognize. But if that doesn't
> > cover the contents we should not use it. I would prefer something
> > longer, I can't say which one would cover this best. "obse" doesn't
> > sound better than "obj" to me.
> 
> This can be answered below 

Reading the discussion and following links, I do see the name "obse"
used many times, while "obl" is rarely used in a sentence.  Just based
on that, and because the term "Oblivion Script Extender" appears to
describe it quite well, I currently would prefer "obse".

From your description it appears that there is really only one filetype
that we can detect, which has the "extended" syntax on top of the
existing syntax.  That should be OK.  A few users may like to have some
extensions not highlighted (or even highlighted as an error, if their
setup won't allow for them).  For that the global variables mentioned
ealier can be used to tweak to do with the filetype.


-- 
Q: How many legs does a giraffe have?
A: Eight: two in front, two behind, two on the left and two on the right

katawful:
I will have to defer to the community on this

I will add the global variables as well

brammool:
Well, no comments for a week.  Let me take the decision then to use the "obse" name.

katawful:
The community either didn't seem interested or moves too slow. If I get feedback that changes the filetype name I will make sure to open a PR to reflect that

brammool:
I also included the ftplugin/indent/syntax files.  I made a few changes to have them pass the regular checks.
Please check out the result at the next runtime files update, and consider sending me further improvements.

PR #11545, opened 2022-11-14
Fix mouse shape remains in op-pending mode after failed change operation

zeertzjq:
If finish_op is reset but not restored, ui_cursor_shape() is not called
after do_pending_operator() returns, so mouse shape remains in operator
pending mode. I can somehow not reproduce this with cursor shape in GUI,
only with mouse shape. Steps to reproduce:
1. `:set mouseshape+=o:up-arrow`
2. `:set nomodifiable`
3. press "c" followed by a motion
4. mouse shape remains up-arrow.

Restoring finish_op seems to fix the problem. No idea how to test this.

### DIFF
--- a/src/ops.c
+++ b/src/ops.c
@@ -1734,6 +1734,7 @@ op_change(oparg_T *oap)
     long		pre_textlen = 0;
     long		pre_indent = 0;
     char_u		*firstline;
+    int			save_finish_op = finish_op;
     char_u		*ins_text, *newp, *oldp;
     struct block_def	bd;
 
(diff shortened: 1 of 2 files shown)


brammool:
> If finish_op is reset but not restored, ui_cursor_shape() is not called
> after do_pending_operator() returns, so mouse shape remains in operator
> pending mode. I can somehow not reproduce this with cursor shape in GUI,
> only with mouse shape. Steps to reproduce:
> 	:set mouseshape+=o:up-arrow
> 	:set nomodifiable
> And press "c" followed by a motion, and mouse shape remains up-arrow.
> 
> Restoring finish_op seems to fix the problem. No idea how to test this.

I have just added the getmouseshape() function in the hope you can use
it in a test for this fix.


-- 
67. Your hard drive crashes. You haven't logged in for two hours.  You start
    to twitch. You pick up the phone and manually dial your ISP's access
    number. You try to hum to communicate with the modem.  You succeed.

zeertzjq:
I have added a test by running GUI Vim in a separate process and starting several timers.

zeertzjq:
Actually in this case `finish_op` is not reset at all if saving for undo fails.

PR #11561, opened 2022-11-17
Move insert mode mouse scroll code out of nv_mousescroll, and back into ins_mousescroll

zewpo:
This PR removes some redundant function calls and superfluous condition checking at runtime for vertical mouse scrolling.

Also, I felt it was a bit clumsy to pass around the "mode" variable, where we probably should just use the global variable "State" if needed.

Anyway - one risk - Is there a valid reason why ins mode mousescroll should NOT also use the move onepage(), the same as nv mode mouse scroll uses?  (ie. this is for when ctrl or shift is pressed along with the vertical mouse scroll event.)

If there is a valid reason why it should not use the same, then I will revert a part of this PR.   But, if it's OK, then the code can be simplified more here.  I can collapse this;

```
if (mode == MODE_INSERT)
{
    long step = (long)(curwin->w_botline - curwin->w_topline);
    scroll_redraw(cap->arg, step);
}
else
{
    did_ins_scroll = onepage(cap->arg ? FORWARD : BACKWARD, 1L);
}
```
to simply be this one line here;
```
onepage(cap->arg, 1L);
```

I have to ask, since when I tried it, I had to adjust `test_edit.vim` slightly as onepage() doesn't quite scroll the same distance as insert mode was. 

```
	command line..script /home/runner/work/vim/vim/src/shadow/testdir/runtest.vim[515]..function RunTheTest[52]..Test_edit_MOUSE 
	line 15: Expected 14 but got 12
	command line..script /home/runner/work/vim/vim/src/shadow/testdir/runtest.vim[515]..function RunTheTest[52]..Test_edit_MOUSE 
	line 19: Expected 24 but got 20
	command line..script /home/runner/work/vim/vim/src/shadow/testdir/runtest.vim[515]..function RunTheTest[52]..Test_edit_MOUSE 
	line 21: Expected [0, 24, 2, 0] but got [0, 20, 2, 0]
	command line..script /home/runner/work/vim/vim/src/shadow/testdir/runtest.vim[515]..function RunTheTest[52]..Test_edit_MOUSE 
	line 24: Expected [0, 27, 2, 0] but got [0, 23, 2, 0]
	command line..script /home/runner/work/vim/vim/src/shadow/testdir/runtest.vim[515]..function RunTheTest[52]..Test_edit_MOUSE 
	line 28: Expected [0, 28, 2, 0] but got [0, 24, 2, 0]
```


zewpo (commented):
ready for review

brammool:
> This PR removes some redundant function calls and superfluous
> condition checking at runtime for vertical mouse scrolling.
> 
> Also, I felt it was a bit clumsy to pass around the "mode" variable,
> where we probably should just use the global variable "State" if
> needed.
> 
> Anyway - one risk - Is there a valid reason why ins mode mousescroll
> should NOT also use the move onepage(), the same as nv mode mouse
> scroll uses?  (ie. this is for when ctrl or shift is pressed along
> with the vertical mouse scroll event.)
> 
> If there is a valid reason why it should not use the same, then I will
> revert a part of this PR.   But, if it's OK, then the code can be
> simplified more here.  I can collapse this;

As a rule of thumb: Never mix refactoring with changing the logic.
Refactoring should always result in exactly the same behavior.

When making changes in logic, even when it's subtle or "should be OK" it
is better done in a separate commit.  When later someone reports a
problem we know the refactoring one should not be the cause (it should
be easy to revert and check the problem is still there).  And then we
can revert the commit that did change behavior without having to revert
the refactoring as well.

I do not recall why in Insert mode the scrolling is done differently,
but I suspect it is related to how 'scrolloff' is applied.  There are
also places where it's different between forward and backward scrolling.
These days I always try to add a comment with a hint why it's done this
way, but it looks like these conditions are old, it's going to be hard
to find out why it was done this way.  Perhaps better just leave it
around, the code isn't ugly and I don't think we can drop more than a
few lines, thus there is hardly a win.

> I have to ask, since when I tried it, I had to adjust `test_edit.vim`
> slightly as onepage() doesn't quite scroll the same distance as insert
> mode was. 

Also an indication that it's better to just leave it as-is.

-- 
From "know your smileys":
 O:-)	Saint

zewpo:
Thanks @brammool   I appreciate your conservative approach.   

OK, this PR no longer has any functional change - at least not intentional.   I've removed that part about changing ins mode to use the move onepage() function.  Leaving it in its original logic pathway.   Then, this PR really just achieves some refactoring cleanup, and simplifying some of the superfluous conditional checks - which should help a bit with performance anyway.

brammool:
> Thanks @brammool   I appreciate your conservative approach.   
> 
> OK, this PR no longer has any functional change - at least not
> intentional.   I've removed that part about changing ins mode to use
> the move onepage() function.  Leaving it in its original logic
> pathway.   Then, this PR really just achieves some refactoring
> cleanup, and simplifying some of the superfluous conditional checks -
> which should help a bit with performance anyway.

The code that is currently at the end of do_mousescroll() with the
condition that "mode" is MODE_INSERT was moved to below where
nv_mousescroll() is called.  But you added:

	 if (has_winscrolled())

The code block handles the popup menu, it doesn't appear to relate to
the WinScrolled autocmd event.  So why this change?


The current code checks pum_visible() also when mouse_row or mouse_col
is negative.  That means we get a scroll event without a position, thus
scroll in the current window.  But after the change pum_visible() isn't
checked then.

Please double check that this is actually a refactoring, currently it
looks like there are logic changes that I don't know what effect they
have.


-- 
From "know your smileys":
 :q	vi user saying, "How do I get out of this damn emacs editor?"

zewpo:
Thanks @brammool  for the questions.

> The code that is currently at the end of do_mousescroll() with the
> condition that "mode" is MODE_INSERT was moved to below where
> nv_mousescroll() is called.  ...

Yeah, back to ins_mouse scroll, where it was before, it is irrelevant for normal mode scroll.

>   ... But you added:
> 
> 	 if (has_winscrolled())
> 
> The code block handles the popup menu, it doesn't appear to relate to
> the WinScrolled autocmd event.  So why this change?

Before, it was setting a local variable as well as calling a function that also set "something? autocmd";
```
	did_scroll = TRUE;
	may_trigger_winscrolled();
```

and it was using **did_scroll** in this block;

```
    if (pum_visible() && **did_scroll**)
    {
	redraw_all_later(UPD_NOT_VALID);
	ins_compl_show_pum();
    }

    if (!EQUAL_POS(curwin->w_cursor, tpos))
    {
	start_arrow(&tpos);
	set_can_cindent(TRUE);
    }
```
this block, I changed to;

```
    if (has_winscrolled())
    {
	if (pum_visible())
	{
	    redraw_all_later(UPD_NOT_VALID);
	    ins_compl_show_pum();
	}
	if (!EQUAL_POS(curwin->w_cursor, tpos))
	{
	    start_arrow(&tpos);
	    set_can_cindent(TRUE);
	}
    }
```

About the lower part, why check for `!EQUAL_POS(curwin->w_cursor, tpos))`  if it didn't actually scroll.  So that part went into the the condition of if has_winscrolled().

For the bigger change here.    When I first started this PR, I tried to merge the ins mode scroll to be able to use onepage(), but that left a potential for it to set `did_scroll = TRUE`, even if it didn't actually scroll.   Consequently it would potentially call redraw_all_later when it didn't actually need to.

I noticed that `may_trigger_winscrolled()`   enables `has_winscrolled()` to return TRUE.   It is more robust, because it actually checks if it really did scroll the window, so that seemed to solve the problem of the local variable did_scroll being potentially wrong.

Then, when I reverted the ins mode scroll, to not use onepage() in this PR, I didn't revert that part of the potential issue being solved anyway.   

may_trigger_winscrolled() and has_winscrolled() seems nominally correct.   But, yeah, it makes it rely on code that was written specifically for autocmd functionality.   It would be helpful if those functions are useable aside from autocmd functionality.  I will consider this more carefully - might raise a separate PR if it looks helpful.

> The current code checks pum_visible() also when mouse_row or mouse_col
> is negative.  That means we get a scroll event without a position, thus
> scroll in the current window.  But after the change pum_visible() isn't
> checked then.

Not exactly - because it was also OR'd with (curwin != old_curwin)

The logic before was like this;
```
    // Don't scroll the window in which completion is being done.
    if (!pum_visible() || curwin != old_curwin)
```

So, this will allow scroll in either of the following two conditions; 

1) when pum_visible() is FALSE - regardless of which window is being requested.

2) when pum_visible() is TRUE - it can still scroll as long as it is requested at a different window.

I find it easier to considering the negative, ie. What conditions does pum_visible() actually prevent scrolling?  

If I'm not mistaken, the ONLY time that pum_visible() can prevent scrolling, is when  pum_visible() is TRUE   AND  the window being scrolled is the current window.  
Essentially, reversing the current logic, means don't scroll when;
```
     pum_visible() && curwin == old_curwin
```

This also correctly matches the comment:  // Don't scroll the window in which completion is being done.

Now, the only way to know if we are scrolling the current window already requires mouse_row and mouse_col to be positive.

So, moving it back into ins_mousescroll, I ended up just returning if the condition is not met, it wont go on to call nv_mousescroll;

```
    if (mouse_row >= 0 && mouse_col >= 0)
    {
	int row = mouse_row;
	int col = mouse_col;

	// find the window at the pointer coordinates
	if (curwin == mouse_find_win(&row, &col, FIND_POPUP))
	{
	    // Don't scroll the window in which completion is being done.
	    if (pum_visible())
		return;
	    else
		undisplay_dollar();
	}
    }
```

zewpo:
@brammool   OK, for now, I've just committed a change decoupling the ins_mousescroll code from the autocmd specific has_winscrolled().   It only a few lines for it to check if the target w_topline or w_leftcol actually changed.   Note, I couldn't revert back to using the did_scroll variable, because of how it now calls the common nv_mousescroll code - which contains most of the common code between the two modes.

brammool:
> Thanks @brammool  for the questions.
> 
> > The code that is currently at the end of do_mousescroll() with the
> > condition that "mode" is MODE_INSERT was moved to below where
> > nv_mousescroll() is called.  ...
> 
> Yeah, back to ins_mouse scroll, where it was before, it is irrelevant
> for normal mode scroll.

I understand that the Insert-specific parts can be moved out of
do_mousescroll() to where it is called.  It's easy for what happens at
the very start and end, however, there are a couple of places in the
middle of the function where "mode" is checked, that's more tricky,
since the order might matter.

> >   ... But you added:
> > 
> > 	 if (has_winscrolled())
> > 
> > The code block handles the popup menu, it doesn't appear to relate to
> > the WinScrolled autocmd event.  So why this change?
> 
> Before, it was setting a local variable as well as calling a function
> that also set "something? autocmd";
> ```
> 	did_scroll = TRUE;
> 	may_trigger_winscrolled();
> ```
> 
> and it was using **did_scroll** in this block;
> 
> ```
>     if (pum_visible() && **did_scroll**)
>     {
> 	redraw_all_later(UPD_NOT_VALID);
> 	ins_compl_show_pum();
>     }
> 
>     if (!EQUAL_POS(curwin->w_cursor, tpos))
>     {
> 	start_arrow(&tpos);
> 	set_can_cindent(TRUE);
>     }
> ```
> this block, I changed to;
> 
> ```
>     if (has_winscrolled())
>     {
> 	if (pum_visible())
> 	{
> 	    redraw_all_later(UPD_NOT_VALID);
> 	    ins_compl_show_pum();
> 	}
> 	if (!EQUAL_POS(curwin->w_cursor, tpos))
> 	{
> 	    start_arrow(&tpos);
> 	    set_can_cindent(TRUE);
> 	}
>     }
> ```

I see "did_ins_scroll" there now.  But that variable name doesn't cover
what it's set to, the return value of onepage() when "mode" is NOT
MODE_INSERT.  And onepage() returns OK/FAIL.  This is one of those
places where the code is wrong and refactoring must be done with great
care.  In fact, you should ignore the names of variables and inspect the
behavior.

I guess what was intended is getting a flag that tells us whether any
scrolling happened.  The OK/FAIL doesn't cover that when a count is
passed.  I would think we should instead check if the value of w_topline
changed.  That's not refactoring though, that should be a separate
commit.

Again: redrawing the popup menu has nothing to do with whether there is
a WinScrolled event.  I really don't see how you can have
has_winscrolled() there without changing what happens.  I think the only
thing that matters is that after scrolling the popup menu may need to be
redrawn.

> About the lower part, why check for `!EQUAL_POS(curwin->w_cursor,
> tpos))`  if it didn't actually scroll.  So that part went into the the
> condition of if has_winscrolled().

Same problem, this doesn't depend on an autocommand event existing.
Hmm, are you reading has_winscrolled() as "did the window scroll"?  
That's not what that function does.

> For the bigger change here.    When I first started this PR, I tried
> to merge the ins mode scroll to be able to use onepage(), but that
> left a potential for it to set `did_scroll = TRUE`, even if it didn't
> actually scroll.   Consequently it would potentially call
> redraw_all_later when it didn't actually need to.
> 
> I noticed that `may_trigger_winscrolled()`   enables
> `has_winscrolled()` to return TRUE.

Eh, no, only adding a WinScrolled autocmd event does that.

> It is more robust, because it actually checks if it really did scroll
> the window, so that seemed to solve the problem of the local variable
> did_scroll being potentially wrong.

Yeah, this confirms you are mixing things up.  We can change the code to
check if w_topline changed, that is the real check that a scroll
happened.  This does require saving w_topline before at some point.  But
again, that's not refactoring but fixing.  We can do this either before
or after refactoring.  Or split it up in more steps, which is better
than trying to do it all at once.

> Then, when I reverted the ins mode scroll, to not use onepage() in
> this PR, I didn't revert that part of the potential issue being solved
> anyway.   
> 
> may_trigger_winscrolled() and has_winscrolled() seems nominally
> correct.   But, yeah, it makes it rely on code that was written
> specifically for autocmd functionality.   It would be helpful if those
> functions are useable aside from autocmd functionality.  I will
> consider this more carefully - might raise a separate PR if it looks
> helpful.

This can't be moved to a function, w_topline needs to be remembered
before the scroll code is invoked.

> > The current code checks pum_visible() also when mouse_row or mouse_col
> > is negative.  That means we get a scroll event without a position, thus
> > scroll in the current window.  But after the change pum_visible() isn't
> > checked then.
> 
> Not exactly - because it was also OR'd with (curwin != old_curwin)
> 
> The logic before was like this;
> ```
>     // Don't scroll the window in which completion is being done.
>     if (!pum_visible() || curwin != old_curwin)
> ```
> 
> So, this will allow scroll in either of the following two conditions; 
> 
> 1) when pum_visible() is FALSE - regardless of which window is being
> requested.
> 
> 2) when pum_visible() is TRUE - it can still scroll as long as it is
> requested at a different window.

Or in Normal mode.

> I find it easier to considering the negative, ie. What conditions does
> pum_visible() actually prevent scrolling?  
> 
> If I'm not mistaken, the ONLY time that pum_visible() can prevent
> scrolling, is when  pum_visible() is TRUE   AND  the window being
> scrolled is the current window.  

And in Insert mode.

> Essentially, reversing the current logic, means don't scroll when;
> ```
>      pum_visible() && curwin == old_curwin
> ```
> 
> This also correctly matches the comment:  // Don't scroll the window
> in which completion is being done.
> 
> Now, the only way to know if we are scrolling the current window
> already requires mouse_row and mouse_col to be positive.
> 
> So, moving it back into ins_mousescroll, I ended up just returning if
> the condition is not met, it wont go on to call nv_mousescroll;
> 
> ```
>     if (mouse_row >= 0 && mouse_col >= 0)
>     {
> 	int row = mouse_row;
> 	int col = mouse_col;
> 
> 	// find the window at the pointer coordinates
> 	if (curwin == mouse_find_win(&row, &col, FIND_POPUP))
> 	{
> 	    // Don't scroll the window in which completion is being done.
> 	    if (pum_visible())
> 		return;
> 	    else
> 		undisplay_dollar();
> 	}
>     }
> ```

What appears to be missing is that in Insert mode, when the mouse
coordinates are not set, it would scroll the current window.  The check
for pum_visible() is missing.

-- 
From "know your smileys":
 ...---...   SOS

zewpo:
Thanks @brammool   

Think I have fixed up this PR again.

I shouldn't have set did_[ins]_scroll to the output returned by onepage() - that was a mistake.   Also, using the autocmd logic to check if it actually scrolled was a mistake.     OK, so I've fixed it to now check if w_topline or w_leftcol actually changed.

```
What appears to be missing is that in Insert mode, when the mouse
coordinates are not set, it would scroll the current window.  The check
for pum_visible() is missing.
```

w.r.t pum_visible().   Ahh, yes, sorry, I found what you mean there.   I had missed that scenario.  It's good to have a second pair of eyes review this stuff.  OK, so this PR now checks that as per original code. with this commit https://github.com/vim/vim/pull/11561/commits/ad55dba62f0eb6b5e979120bf68ec1b6ab23f436

Hopefully, it is back in harmony with original functionality.

PR #11562, opened 2022-11-17
fix spelling error

uniontech-lilinjie:
Signed-off-by: lilinjie <102012657+uniontech-lilinjie@users.noreply.github.com>
fix spelling error


brammool:
Have you contacted the maintainer?   I don't know a github user name for "scootergrisen", unfortunately.

brammool:
Note that there is also this PR: https://github.com/vim/vim/pull/10703
I hope someone can find the maintainer, or otherwise someone can take over.

nickspoons:
Is this fix even correct? I know Danish as a second language, not a native speaker, but I would have thought "patchet af" was more correct than "patched af". Tricky because "patch" is not a danish word, so I'm not 100% sure how the past tense is applied here.

nickspoons:
Apologies, I have verified with another dansker that "patched af" is the preferred usage, so this PR is correct.

brammool:
OK, I'll include it then.  If the maintainer objects we can change it again or revert.

PR #11563, opened 2022-11-17
Add new type keyword "any" and "comparable" for the go syntax

kyoh86:
go from 1.18 has new type keyword "any" and "comparable".

from [Go 1.18 Release Notes](https://tip.golang.org/doc/go1.18):
> The new [predeclared identifier](https://tip.golang.org/ref/spec#Predeclared_identifiers) `any` is an alias for the empty interface. It may be used instead of `interface{}`.
> The new [predeclared identifier](https://tip.golang.org/ref/spec#Predeclared_identifiers) `comparable` is an interface that denotes the set of all types which can be compared using `==` or `!=`. It may only be used as (or embedded in) a type constraint.

### DIFF
--- a/runtime/syntax/go.vim
+++ b/runtime/syntax/go.vim
@@ -117,7 +117,7 @@ hi def link     goLabel             Label
 hi def link     goRepeat            Repeat
 
 " Predefined types
-syn keyword     goType              chan map bool string error
+syn keyword     goType              chan map bool string error any comparable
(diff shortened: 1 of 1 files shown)


brammool:
@bhcleek

bhcleek:
Looks good to me.

I also created #11567. It includes this change plus other changes for Go to support generics and other improvements that I've made since the last go syntax file update.

brammool:
Included as part of #11567

PR #11452, opened 2022-10-27
feat: added tape extension

caarlos0:
[VHS](https://github.com/charmbracelet/vhs) is a tool that generates terminal gifs from code, like asciicinema, but with actual gifs as result, and instead of recoding it, you generate them from code.

`.tape` is the extension of those files.

I already PR the treesitter stuff, check https://github.com/nvim-treesitter/nvim-treesitter/pull/3726

This registers `.tape` files to the `vhs` filetype.

refs https://github.com/neovim/neovim/pull/20831

PR #11565, opened 2022-11-17
feat: add CQL filetype to runtime

mgramigna:
Clinical Quality Language (CQL) is a language used in a variety of Healtchare IT/clinical use cases. See https://cql.hl7.org/ and https://github.com/cqframework/clinical_quality_language for more description on CQL, and [here](https://github.com/cqframework/clinical_quality_language/blob/master/Examples/ChlamydiaScreeningCDS.cql) for an example `.cql` file.

This PR simply adds detection of `*.cql` files and sets the filetype accordingly.

I have not yet provided any syntax for the CQL grammar itself within Vim, but let me know if this is necessary in order for this contribution to be accepted. Thank you!

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -420,7 +420,9 @@ au BufNewFile,BufRead configure.in,configure.ac setf config
 au BufNewFile,BufRead *.cook			setf cook
 
 " Clinical Quality Language (CQL)
-au BufNewFile,BufRead *.cql	  		setf cql
+" .cql is also mentioned in 'XDCC Catcher queue list' as a file extension
+" If support for XDCC Catcher is needed in the future, the contents of the file needs to be inspected.
+au BufNewFile,BufRead *.cql	  		setf cqlang
 
 " CSV Files
 au BufNewFile,BufRead *.csv			setf csv
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -419,6 +419,9 @@ au BufNewFile,BufRead configure.in,configure.ac setf config
 " Cooklang
 au BufNewFile,BufRead *.cook			setf cook
 
+" Clinical Quality Language (CQL)
+au BufNewFile,BufRead *.cql	  		setf cql
+
 " CSV Files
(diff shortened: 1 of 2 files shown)


brammool:
> Clinical Quality Language (CQL) is a language used in a variety of
> Healtchare IT/clinical use cases. See https://cql.hl7.org/ and
> https://github.com/cqframework/clinical_quality_language for more
> description on CQL, and
> [here](https://github.com/cqframework/clinical_quality_language/blob/master/Examples/ChlamydiaScreeningCDS.cql)
> for an example `.cql` file.
> 
> This PR simply adds detection of `*.cql` files and sets the filetype
> accordingly.
> 
> I have not yet provided any syntax for the CQL grammar itself within
> Vim, but let me know if this is necessary in order for this
> contribution to be accepted. Thank you!

We always add the filetype detection separately, so this PR is fine by
itself.

However, looking around for the .cql extension, I see mention of "XDCC
Catcher queue list".  Apparently the extension is used for more than one
type of file.  We prefer to give a descriptive name, rather than just
the extension name itself.

Does this "Clinical Quality Language" have another (short) name?
I find this: "ANSI/HL7 CQLANG, R1-2020".  So perhaps we can call it
"cqlang"?

We could just recognize .cql files as "cqlang" for now.  Only when we
care about that "XDCC Catcher queue list" we would have to check the
contents and come up with a name for that filetype.

Doing it this way avoids that we get a "cql" filetype that we don't know
whether it is one or the other.  And renaming later won't be appreciated
by users.

-- 
From "know your smileys":
 |-P	Reaction to unusually ugly C code

mgramigna:
@brammool Thanks for the insight, and I agree with your reasoning. I went ahead and updated to `cqlang`, which I think is an appropriate short name for clinical quality language based on the CQL specification.

I also added a comment clarifying a potential discrepancy if support for XDCC Catcher queue list is needed in the future, which I saw [requested in another filetype PR](https://github.com/vim/vim/pull/11452#issuecomment-1304933859)

brammool:
Included with patch  9.0.0897
Somehow that referenced another PR that was already included.

PR #11567, opened 2022-11-17
Update the go syntax file with some recent changes made to vim-go.

bhcleek:


brammool:
I'll include it, thanks.

PR #11577, opened 2022-11-19
CI: update to clang 16 and llvm 16

pheiduck:


pheiduck:
@brammool this can be merged.

PR #11579, opened 2022-11-19
feat: selectedlines function

doubleloop:
Add new function which allow to get currently selected text without leaving visual mode.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -7862,11 +7862,6 @@ searchpos({pattern} [, {flags} [, {stopline} [, {timeout} [, {skip}]]]])
 		Can also be used as a |method|: >
 			GetPattern()->searchpos()
 
-selectedlines()						*selectedlines()*
-		Returns current visual selection. The result is returned as an
-		array of lines. In modes other then visual, emty array is
-		returned.
-
 server2client({clientid}, {string})			*server2client()*
 		Send a reply string to {clientid}.  The most recent {clientid}
 		that sent a string can be retrieved with expand("<client>").
@@ -10118,6 +10113,11 @@ visualmode([{expr}])						*visualmode()*
 		If {expr} is supplied and it evaluates to a non-zero Number or
 		a non-empty String, then the Visual mode will be cleared and
 		the old value is returned.  See |non-zero-arg|.
+
+visualtext()						*visualtext()*
+		Returns current visual selection. The result is returned as an
+		array of lines. In modes other then visual, emty array is
+		returned.
 
 wildmenumode()					*wildmenumode()*
 		Returns |TRUE| when the wildmenu is active and |FALSE|
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -10115,9 +10115,18 @@ visualmode([{expr}])						*visualmode()*
 		the old value is returned.  See |non-zero-arg|.
 
 visualtext()						*visualtext()*
-		Returns current visual selection. The result is returned as an
-		array of lines. In modes other then visual, emty array is
-		returned.
+		The result is a list of strings where each item is a line
+		from the current visual selection. The result is the same as
+		as what would be yanked to register on yank operation of current
+		sellection. In particular:
+		- in virtualmode enabled, if selection is past the end of line,
+		  result is filled with spaces
+		- if selection starts/ends somewhere in the middle of
+		  multiwidth character (ex. tab), the whole character is not
+		  included, instead its selected part is substituted with
+		  spaces
+
+		In modes other then visual, emty list is returned.
 
 wildmenumode()					*wildmenumode()*
 		Returns |TRUE| when the wildmenu is active and |FALSE|
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -7862,6 +7862,11 @@ searchpos({pattern} [, {flags} [, {stopline} [, {timeout} [, {skip}]]]])
 		Can also be used as a |method|: >
 			GetPattern()->searchpos()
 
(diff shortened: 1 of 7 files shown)


habamax:
If it returns text of an arbitrary visual selection (which I hope it is), then the name should be adjusted:

- `visualtext()`
- `getvisualtext()`
- `getvisual()`
- etc

doubleloop:
>if it returns text of an arbitrary visual selection (which I hope it is)

result is a List where each item is a line from the current selection.

>  the name should be adjusted

Yes, I also think the name should be changed.. The most analogous existing function is `getline()` so `getselection` or `getvisual` (btw, can there be selection which is not visual)?

habamax:
> > if it returns text of an arbitrary visual selection (which I hope it is)
> 
> result is a List where each item is a line from the current selection.

That is not what I expected, it is fairly easy to do with `line("v")` and `line(".")` -- https://asciinema.org/a/n2zSbsrycl3F19wDfS99n1Z7x

What is missing is to get visual text which is charwise. Currently one has to save the register, yank to that register, extract from the register to a variable, restore the register.

> 
> > the name should be adjusted
> 
> Yes, I also think the name should be changed.. The most analogous existing function is `getline()` so `getselection` or `getvisual` ? btw, can there be selection which is not visual?

There is visual mode and select mode.

doubleloop:
> That is not what I expected, it is fairly easy to do with line("v") and line(".")

I know about this, but this PR handles all kinds of selection (charwise/blockwise) with support of virtualedit and various values of seleciton option and also handles multibyte characters or multiple width characters.

doubleloop:
vim6

habamax:
> > That is not what I expected, it is fairly easy to do with line("v") and line(".")
> 
> I know about this, but this PR handles all kinds of selection (charwise/blockwise) with support of virtualedit and various values of seleciton option and also handles multibyte characters or multiple width characters.

Aha, got it. This would be really useful!

brammool:
This seems to be most useful to get the Visually selected text without
having to go through a register (and making sure the area is still
selected).

For the name I would think visualtext() would be clear.  It sorts just
below visualmode().

The help should be very clear about what the returned text is.  The
existing visualmode() function returns the  mode already, thus that
doesn't need to be included.

There are a few choices to be made:

- When the mode is characterwise and the Visual area is only one line,
  does it return a List with one String, or would it just return a String?
  I would think always returning a List of Strings is more consistent.

- When the mode is blockwise and a Tab is partly before and partly
  inside the block, what is returned?  I would think that spaces need to
  be used for the part of the Tab that is inside the block, so that when
  the text is put somewhere it lines up correctly.

- When the mode is blockwise some parts of the block may be shorter and
  not have text.  Does this just get left out or are spaces added so
  that all the lines have the same width?
  Both can be useful, an argument to specify what is desired would be
  useful.

-- 
From "know your smileys":
 :.-(	Crying

runtime/doc/builtin.txt
```diff
@@ -10114,6 +10114,20 @@ visualmode([{expr}])						*visualmode()*
 		a non-empty String, then the Visual mode will be cleared and
 		the old value is returned.  See |non-zero-arg|.
 
+visualtext()						*visualtext()*
+		The result is a list of strings where each item is a line
+		from the current visual selection. The result is the same as
+		as what would be yanked to register on yank operation of current
+		sellection. In particular:
```
habamax, line 10121: typo -- selection

src/list.c
```diff
@@ -690,6 +690,25 @@ list_append_string(list_T *l, char_u *str, int len)
     return OK;
 }
 
+
+/*
+ * Simmilar to list_append_string() but does not make copy of the 'str'.
```
habamax, line 695: Typo -- Similar

habamax:
Today I needed this function again (for a mapping that should pass visual text to an external command).

End up using, as usual, `"0y:<scriptcmd>....getreg("0")<cr>`.

yegappan:
Can you update and rebase this PR to the latest?

runtime/doc/builtin.txt
```diff
@@ -10114,6 +10114,20 @@ visualmode([{expr}])						*visualmode()*
...
+		- in virtualmode enabled, if selection is past the end of line,
+		  result is filled with spaces
+		- if selection starts/ends somewhere in the middle of
+		  multiwidth character (ex. tab), the whole character is not
+		  included, instead its selected part is substituted with
+		  spaces
+
+		In modes other then visual, emty list is returned.
```
ychin, line 10129: Typo: "emty" should be "empty"

runtime/doc/builtin.txt
```diff
@@ -10114,6 +10114,20 @@ visualmode([{expr}])						*visualmode()*
 		a non-empty String, then the Visual mode will be cleared and
 		the old value is returned.  See |non-zero-arg|.
 
+visualtext()						*visualtext()*
```
ychin, line 10117: Minor nitpicking: For adding new help tags, I don't know whose responsibility it is to add to the [tags](https://github.com/vim/vim/blob/master/runtime/doc/tags) file. Normally in the past, I would assume Bram just did it himself when doing runtime file updates but even I had noticed that sometimes the tags were not always updated.

Can this PR just add to the tags file if the PR is going to be a direct merge? You can either run the tags generation using the `helptags` command or what I sometimes do is to just manually add it to the tags file (need to make sure format is correct and sorted).

ychin: Oh actually now I remember, I think we added a "vimtags" check to the [ci.yml ](https://github.com/vim/vim/blob/master/.github/workflows/ci.yml) file so it would check that the tags are correct in CI. So we should definitely add the tags to pass CI.

PR #11580, opened 2022-11-20
fix(sh,zsh): here document

e-kwsm:
Consider the following shell file:
```sh
cat << EOF
EOF extra
# zsh syntax thinks the above line is the end of heredoc
EOF 
# sh syntax thinks the above line is the end of heredoc even if it has trailing space
the below is the actual EOF
EOF
```

(GitHub also misunderstands)

From <https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_07_04>,

> The here-document shall be treated as a single word that begins after the next <newline> and continues until there is a line containing only the delimiter and a &lt;newline&gt;, with no &lt;blank&gt; characters in between.

Therefore a shell must print the heredoc part, i.e. the body except for the first and the last lines.
dash 0.5.10.2, bash 5.0, and zsh 5.8 behaves as such on Ubuntu 20.04.5 LTS.

### DIFF
--- a/runtime/syntax/sh.vim
+++ b/runtime/syntax/sh.vim
@@ -415,21 +415,21 @@ syn match	shBQComment	contained	"#.\{-}\ze`"	contains=@shCommentGroup
 "  (modified by Felipe Contreras)
 " =========================================
 ShFoldHereDoc syn region shHereDoc matchgroup=shHereDoc01 start="<<\s*\z([^ \t|>]\+\)"		matchgroup=shHereDoc01 end="^\z1$"	contains=@shDblQuoteList
-ShFoldHereDoc syn region shHereDoc matchgroup=shHereDoc02 start="<<-\s*\z([^ \t|>]\+\)"		matchgroup=shHereDoc02 end="^\s*\z1$"	contains=@shDblQuoteList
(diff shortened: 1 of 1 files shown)


e-kwsm:
Note that this does not work with:

```shell
echo `cat << EOF
backtick following EOF seems allowed
EOF`

# echo $(cat << EOF
# paren following EOF seems disallowed
# EOF)
```

cecamp:
I'll include the changes for your first test case. The second one, with a backtick expression, is difficult.

e-kwsm:
@cecamp regarding`<<- EOF` syntax, the leading *tabs* are ignored.

chrisbra:
the zsh part is already part of https://github.com/chrisbra/vim-zsh/blob/master/syntax/zsh.vim please confirm. If yes, please remove this part from your PR here. I'll send updated zsh files to Bram in a bit.

e-kwsm:
zsh patch is removed.

chrisbra:
I tried the example from the first message:
```
cat << EOF
EOF extra
# zsh syntax thinks the above line is the end of heredoc
EOF 
# sh syntax thinks the above line is the end of heredoc even if it has trailing space
the below is the actual EOF
EOF
```
and it seems to work fine for shell syntax. So let me close it.

PR #11587, opened 2022-11-21
Fix for modifier keys to work with mouse scroll mappings, in MS-Win.

zewpo:
Problem:  in MS-windows terminal, modifier keys were not working in mappings with mouse scroll events.
Solution  Use K_SPECIAL instead of CSI for the modifier keys, and move code out of vgetc() and into check_termcode().

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -1745,10 +1745,8 @@ vgetc(void)
 
 	    // Get two extra bytes for special keys
 	    if (c == K_SPECIAL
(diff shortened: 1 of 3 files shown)


zewpo (commented):
This PR is ready for review.  This finishes the first stage of my initial goal to enable mouse scrolling in MS-win terminals.

The remaining task is for me to write test scripts to cover all my changes where possible. I know its more ideal to have test scripts first.  Anyway, I want to work on that now.

I don't think the current CI testing fails are related to this PR.

PR #11592, opened 2022-11-23
configure.ac: Remove Ruby LDFLAGS

zdohnal:
Adding Ruby LDFLAGS into our LDFLAGS causes several problems, because Ruby embeds all LDFLAGS it was link with, even the flags from buildsystem.

Ruby LDFLAGS brings unwanted -arch flags on MacOS and -spec on Fedora.

I was able to successfuly build Vim with Ruby support without Ruby LDFLAGS on Fedora, and if other distribution can build as well, I would propose to remove them from configure.

The PR is work in progress until CI tests on PR finishes (it can verify whether the change does not break Ruby on other distros).

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -7683,15 +7683,6 @@ $as_echo "$rubyhdrdir" >&6; }
 	if test "X$librubyarg" != "X"; then
 	  RUBY_LIBS="$librubyarg $RUBY_LIBS"
 	fi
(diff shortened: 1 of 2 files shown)


zdohnal:
Looks like Macos and Linux CI runs were able to build with Ruby without its LDFLAGS and Windows build doesn't configure Vim with Ruby at all, so the result LGTM - switching for the review.

PR #11595, opened 2022-11-23
[css] Update pseudo-classes

adrianheine:
I just copied all non-experimental ones from https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes


brammool:
@JSitter

PR #11604, opened 2022-11-24
Pr correct cov err

Shane-XB-Qian:
to fix annoyed cov err mail sent out every day.
// pls refer #11223

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Download Coverity
         run: |
-          # probably no TOKEN if it was a forked repo, then just cancel the process
+          # probably no TOKEN if it was a forked repo
           [ -n "${TOKEN}" ] && wget -q https://scan.coverity.com/download/cxx/linux64 --post-data "token=$TOKEN&project=vim" -O coverity_tool.tgz || exit 0
           mkdir cov-scan
           tar ax -f coverity_tool.tgz --strip-components=1 -C cov-scan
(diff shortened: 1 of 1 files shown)


k-takata:
How about this patch instead of yours? (Not tested.)

```diff
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -9,6 +9,7 @@ permissions:
 
 jobs:
   scan:
+    if: secrets.COVERITY_SCAN_TOKEN != ''
     runs-on: ubuntu-20.04
 
     env:
```

Shane-XB-Qian:
> How about this patch instead of yours? (Not tested.)

yes, that's another way, req `{{  }}` though.
but seems `unrecognized named-value: 'secrets'` at that timing.

k-takata:
Ah, the [document](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) says: 

> Secrets cannot be directly referenced in `if:` conditionals. Instead, consider setting secrets as job-level environment variables, then referencing the environment variables to conditionally run steps in the job.

I confirmed that this patch (https://github.com/k-takata/vim/commit/e3fa34c7caab0f5d728e6fa0b7951d37615b3e4a) works as expected.
Here is the log: https://github.com/k-takata/vim/actions/runs/3538312492/jobs/5939042106

Shane-XB-Qian:
yes, as that document said it worked in `job-level`, seems that's the reason.
i hear:
1. there is `if: ${{ !github.event.foo.bar.fork }}` something like that can global work through it, but for now i am not sure what var should be in this case, 
2. and looks there is no direct easy way to fully cancel (early return) the actions steps in this ci system, 
so for now just using ugly :rofl: way to workaround it, maybe can refine it later...

PR #11609, opened 2022-11-24
CI: Check COVERITY_SCAN_TOKEN in coverity scan

k-takata:
This fixes the issue that has been fixed by 9.0.0937 in the different way (as I wrote in https://github.com/vim/vim/pull/11604#issuecomment-1326026151).
The patch 9.0.0937 may ignore actual errors also. I don't think it is a good way.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -15,21 +15,22 @@ jobs:
       CC: gcc
       CFLAGS: -Wno-deprecated-declarations
       DEBIAN_FRONTEND: noninteractive
+      TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
 
     steps:
       - name: Checkout repository from github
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
> The patch 9.0.0937 may ignore actual errors also.

it is useless 'actual' error if it is in 'fork' repo, but it should be would not ignore error in main repo.

// but anyway, as the comment https://github.com/vim/vim/pull/11604#issuecomment-1326048294 said the really good way should be finding out `if: ${{ !github.event.foo.bar.fork }}` that var and ignore the cov process globally if that's a `fork` repo.
// but just so far i donot know what's that var, later maybe can refine that pr to that way or someone very knew ci var can tell what's the var is then early return the cov process if it is a forked repo.

brammool:
The sound-dummy problem has been fixed, please sync to head to trigger the tests.

PR #11612, opened 2022-11-25
fix: early return cov ci in fork

Shane-XB-Qian:
#11604 
#11609

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -15,22 +15,22 @@ jobs:
       CC: gcc
       CFLAGS: -Wno-deprecated-declarations
       DEBIAN_FRONTEND: noninteractive
-      TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
 
+    # probably no TOKEN and some others if it was a forked repo, then just cancel the process
+    if: ${{ !github.event.repository.fork }}
     steps:
       - name: Checkout repository from github
-        if: env.TOKEN
         uses: actions/checkout@v3
 
       - name: Download Coverity
-        if: env.TOKEN
         run: |
           wget -q https://scan.coverity.com/download/cxx/linux64 --post-data "token=$TOKEN&project=vim" -O coverity_tool.tgz
           mkdir cov-scan
           tar ax -f coverity_tool.tgz --strip-components=1 -C cov-scan
+        env:
+          TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
 
       - name: Install packages
-        if: env.TOKEN
         run: |
           sudo apt update && sudo apt install -y \
             autoconf \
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
i found the `var` now, supposed this was the really `good way`.

.github/workflows/coverity.yml
```diff
@@ -17,6 +17,8 @@ jobs:
       DEBIAN_FRONTEND: noninteractive
       TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
 
+    # probably no TOKEN if it was a forked repo, then just cancel the process
+    if: ${{ !github.event.repository.fork }}
```
k-takata, line 20: Is there a document about this variable?

Shane-XB-Qian: perhaps, but i am not sure if there was a doc for this.

k-takata:
I haven't tested this, but if this works as expected, then both 9.0.0946 and 9.0.0937 should be reverted?

Shane-XB-Qian:
i've tested a bit in my local forked repo, the cov ci just skipped which would not do anything, vs in main repo here you can see it had generated.
you can test it a bit in your side as well, if ok, i think those `if: env.TOKEN` in each `steps` better be removed which looked duplicated and a bit useless if so.

brammool:
How reliable is this "github.event.repository.fork" flag?   What if someone creates a copy of the repository locally and then puts it on github?  I assume it won't have this flag then.

Shane-XB-Qian:
> What if someone creates a copy of the repository locally and then puts it on github?

well, that would be another official vim repo then? or i assume we do not need to care about that.. :rofl:

Shane-XB-Qian:
> How reliable is this "github.event.repository.fork" flag?

actually i am not 100% sure if that's a fixed flag or not, but per general sense i supposed github would not change it, though that's for all events.
and at your admin actions you perhaps can manually run it to verify if it is ok to run for main repo, or the schedule would run it also at tomorrow.

brammool:
> > What if someone creates a copy of the repository locally and then
> > puts it on github?
> 
> well, that would be another official vim repo then? or i assume we do
> not need to care about that.. :rofl: 

Theoretically someone could add a Coverity key to their own repo.
Or even to a fork.

The point is using the direct connection instead of an indirect
association.  The Coverity part of CI can work if there is the secrect
Coverity key, it doesn't work if that key is missing.  Thus checking for
the key is the direct connection.  Checking if the repo is a fork
doesn't give us the information directly, it's an indirect indication.

-- 
138. You develop a liking for cold coffee.

Shane-XB-Qian:
> The Coverity part of CI can work if there is the secrect Coverity key, it doesn't work if that key is missing.

that's true, but that depended on if you (repo owner) like (willing) to expose the token into repo itself or not.
// i actually saw some repo truly did like that, though i am not sure if that's ok, or how it work.....

or specific to this case / vim repo, i assume most (repo fork) users should be not care the cov result or since the main repo can do it already..

Shane-XB-Qian:
r.i.p bram.

PR #11616, opened 2022-11-25
Update sr.po

eevan78:
Update the Serbian messages translation


brammool:
I'll include it, thanks.

PR #11622, opened 2022-11-26
New method (test_mswin_event) for testing low-level MS-Windows input

zewpo:
There doesn't seem to be any low level testing of Vim's handling of the MS-Windows console input buffer.   

I've been making changes around this in my last few PRs.  I'd like to include proper test scripts, and this PR is a suggestion of how I might go about achieving that.

@brammool   I know you said that you wish there was testing around this stuff.  So, even though this is still just a Work in Progress, I'd like to get some community feedback before I do too much work on it.  I want this testing to fit in with the current testing ecosystem as much as possible.  If I'm going the wrong direction, it would be good to know sooner rather than later.


brammool:
> There doesn't seem to be any low level testing of Vim's handling of
> the MS-Windows console input buffer.   
> 
> I've been making changes around this in my last few PRs.  I'd like to
> include proper test scripts, and this PR is a suggestion of how I
> might go about achieving that.
> 
> @brammool   I know you said that you wish there was testing around
> this stuff.  So, even though this is still just a Work in Progress,
> I'd like to get some community feedback before I do too much work on
> it.  I want this testing to fit in with the current testing ecosystem
> as much as possible.  If I'm going the wrong direction, it would be
> good to know sooner rather than later. 

Hmm, I think this should be split in two parts, one using feedkeys() to
put escape codes in the typeahead buffer, and another part that
synthesizes windows events.

We can't really use the same stuff as in testdir/test_termcodes.vim,
that is made for emulating what escape codes a terminal would send, and
the sequences used for MS-Windows are different.  But we can use
something similar to inject the bytes that mch_inchar() returns.

mch_inchar() in os_win32.c has a typeahead buffer, that is probably
where the bytes should go.  The buffer has to be made bigger and outside
of mch_inchar(), that shouldn't be difficult.  We can then use
feedkeys() to put bytes in this typeahead buffer.

This allows for testing the path where mouse events and other special
codes are generated for MS-Windows, which are then decoded in
check_termcode().  This will require adding tests specifically for
MS-Windows, e.g. in testdir/test_mswincodes.vim which would have similar
tests as in test_termcodes.vim.  Some common parts could be moved to a
shared script.

The other part is more like what we do with test_gui_event().  Obvious
name would be test_mswin_event().  These events would then be queued and
when WaitForChar() is called they would be used.  Not sure how exactly,
but this is probably the lowest level we can do tests at.

I hope this makes sense.


-- 
147. You finally give up smoking...because it made the monitor dirty.

zewpo:
thanks @brammool 

```
test_mswin_event().  queued and ... WaitForChar()
```

Yes, was able to do this via MS Windows  WriteConsoleInput function.  This writes the events directly into the console's own buffer, outside of Vim.   Then WaitChar calls the consoles readinput buffer.  So, this is lowest level possible, and can thoroughly test all of the code I have been working on.

https://learn.microsoft.com/en-us/windows/console/writeconsoleinput

So...
`f_test_mswin_event()` is in a similar logical layer f_test_gui_event() 
and ...
`test_mswin_event()` is in a similar logical layer to test_gui_w32_sendevent()

I've also started to add `feed_mswin_input()`   It can provide the low level implementation for feedkeys() for MS-windows consoles - which did not previously exists.   I think it still should go through writeconsoleinput() windows api function, that way it truly is the lowest level to feed those keys.    

Otherwise, I thought I saw that MS-Windows consoles already allow testing of feedkeys at higher levels, which simply adds to the typeahead buffer.  (If I recall correctly - will double check that.)   My issue was that adding directly to the typeahead buffer is already catered for, and that does bypass some of the handling of the windows INPUT_RECORD events - as you noted anyway.

The use case for me adding `feed_mswin_input()` is not strictly to add functionality.   Because there was already the higher level feedkeys.    But, adding it fits more in-line with a lot of existing test scripts, and makes it easy for the scripts to be similar.

Note:  I also looked at extending the `test_feedinput` function.  That seems to also achieve similar things, and I am not entirely certain of its distinct use case.  seems a bit redundant somehow, but I admit I only looked at it briefly.

zewpo (commented):
Ready for some feedback, because its growing larger.  Especially, the testing of keycodes in windows console.   I still have some work to do on cleaning that up, also need to do the same sort of thing for the windows console mouse input events.

zewpo (commented):
merge conflict resolved

zewpo (commented):
Ready to have someone review my changes.   Especially the new `input_record_buffer` that I added for caching mswin input records. 

A number of tests are passing  now.  Not all yet.  Some things work on my environment, but dont yet know why they are failing in CI.   Also, not all tests are written yet either.  But, at least the low-level testing harness concept itself is starting to work for MS-windows console events.

zewpo:
Why does CI testing give `fatal error U1058: terminated by user`?    Happening for the tests of my PR, on mswin vim (gui=no).    Caught my attention because these seem to all happen only where my changes are in effect, but I don't know what that error message means.

zewpo:
It was NMake testing timeout throws the "fatal error U1058: terminated by user" .   I think I know what I might have broke, suspect related to misunderstanding how "msg_wait_for_multiple_objects" and/or "wait_for_single_object"  is meant to work wrt FEAT_CLIENTSERVER.   Debugging it now to find out.

zewpo:
Hi @brammool 

The primary goal of this test harness, I think, was initially to provide testing for how Vim reads in INPUT_RECORDS.   Starting from read_console_input(), and specifically for testing how VIm decodes those INPUT_RECORD events into bytes and appends those bytes onto Vims internal typeahead & input-buffers.    Unless I'm mistaken, there doesn't seem to be any testing of that translation.    from INPUT_RECORD to input buffer.    And, I think I have already found bugs in there developing this test harness (will come back to those bugs later.)

The new function:  test_mswin_event() is similar to, and overlaps with; test_gui_event(), so I redirect events from test_mswin_event() to test_gui_event() if the gui is running.

test_mswin_event() achieves the primary goal - and is sufficient to test my previous PRs related to handling the MOUSE_EVENT type of INPUT_RECORDs.     

Now, **feedkeys** mode L (lowlevel) is tricky.

According to the builtin.txt feedkeys mode Lowlevel ` Only works for Unix or when using the GUI`

```
'L'	Lowlevel input.  Only works for Unix or when using the
	GUI. Keys are used as if they were coming from the
	terminal.  Other flags are not used.  *E980*
```
So, I assumed I could make feekeys L work for the MS-Windows console by injecting the "keys" into windows console input buffer, as INPUT_RECORD events - using the same injection point (buffer) that I created for test_mswin_event().    Then Vim can process those INPUT_RECORDS.    I got this working also for feedkeys L in windows.

**BUT** - it looks like my change to feedkeys breaks some pre-existing test scripts which already run feedkeys mode L  in MS-windows consoles . ie. feedkeys L already adds to the typeahead buffer, even in windows consoles - despite what the doco claims.  It may be quite flakey how that works as is.  Also I don't yet know why my change broke those tests.  But I am looking into it all that.

feedkeys already sends 'keys' that can be added more-or-less directly onto the typeahead buffer, with little-to-no translation needed.   So, there is the sense that this is not really needed.  What is the value add proposition of translating the keys into INPUT_RECORDS, only for them to be untranslated back again onto the typeahead and input buffers.   

I was wanting to do both parts together, (feedkeys and test_mswin_event), because they both depend on the changes I'm making to be able to inject INPUT_RECORD events at the lowest feasible point.    Just sharing my thoughts, but if I can't work feedkeys issues out soon, then I think I will take all the feedkeys changes out of this PR and put them into a separate PR later.  

Then it struck me, this might be something that you'd prefer me to do anyway!?  Since I'm new to contributing to anything open source at this level, I wondered if I was doing too much in one PR anyway?   Are you OK with me bundling it together? (that is, if I can work out the feedkeys issue soon - then its easier for me to bundle it together at this stage - if its OK with you.)

brammool:
> Hi @brammool 
> 
> The primary goal of this test harness, I think, was initially to
> provide testing for how Vim reads in INPUT_RECORDS.   Starting from
> read_console_input(), and specifically for testing how VIm decodes
> those INPUT_RECORD events into bytes and appends those bytes onto Vims
> internal typeahead & input-buffers.    Unless I'm mistaken, there
> doesn't seem to be any testing of that translation.    from
> INPUT_RECORD to input buffer.    And, I think I have already found
> bugs in there developing this test harness (will come back to those
> bugs later.)
> 
> The new function:  test_mswin_event() is similar to, and overlaps
> with; test_gui_event(), so I redirect events from test_mswin_event()
> to test_gui_event() if the gui is running.
> 
> test_mswin_event() achieves the primary goal - and is sufficient to
> test my previous PRs related to handling the MOUSE_EVENT type of
> INPUT_RECORDs.     
> 
> Now, **feedkeys** mode L (lowlevel) is tricky.
> 
> According to the builtin.txt feedkeys mode Lowlevel ` Only works for
> Unix or when using the GUI`
> 
> ```
> 'L'	Lowlevel input.  Only works for Unix or when using the
> 	GUI. Keys are used as if they were coming from the
> 	terminal.  Other flags are not used.  *E980*
> ```
> So, I assumed I could make feekeys L work for the MS-Windows console
> by injecting the "keys" into windows console input buffer, as
> INPUT_RECORD events - using the same injection point (buffer) that I
> created for test_mswin_event().    Then Vim can process those
> INPUT_RECORDS.    I got this working also for feedkeys L in windows.

The lowlevel feedkeys() is to inject Escape sequences that Vim should be
able to decode.  For many platforms this is what Vim gets from the
terminal, this is the lowest level.

With a Windows console this is different.  Various messages or events
are received, and Vim translates them in to a byte sequence, which is
then handled like the terminal input on other systems.  For the GUI it's
similar.

Thus lowlevel feedkeys() should not create messages or events, but
inject at the level of the byte sequences.  These byte sequences were
made up at some point, they are Vim specific.

The code is confusing and hardly commented, it's not so easy to see how
it works.  It appears that the place where the byte sequences are
created is in mch_inchar().  The typeahead[] buffer there is where I
would expect the feedkeys() low level input to go.

> **BUT** - it looks like my change to feedkeys breaks some pre-existing
> test scripts which already run feedkeys mode L  in MS-windows consoles
> . ie. feedkeys L already adds to the typeahead buffer, even in windows
> consoles - despite what the doco claims.  It may be quite flakey how
> that works as is.  Also I don't yet know why my change broke those
> tests.  But I am looking into it all that.

I don't think that is supposed to work.  Perhaps USE_INPUT_BUF got
defined because FEAT_CLIENTSERVER is defined?  Then the input buffer is
used, code that is in the second half of ui.c.

> feedkeys already sends 'keys' that can be added more-or-less directly
> onto the typeahead buffer, with little-to-no translation needed.   So,
> there is the sense that this is not really needed.  What is the value
> add proposition of translating the keys into INPUT_RECORDS, only for
> them to be untranslated back again onto the typeahead and input
> buffers.   

There currently is no code to convert byte codes to Windows messages and
events, and adding this appears to be a lot of work and not really what
we want to do.

> I was wanting to do both parts together, (feedkeys and
> test_mswin_event), because they both depend on the changes I'm making
> to be able to inject INPUT_RECORD events at the lowest feasible point.
> Just sharing my thoughts, but if I can't work feedkeys issues out
> soon, then I think I will take all the feedkeys changes out of this PR
> and put them into a separate PR later.  
> 
> Then it struck me, this might be something that you'd prefer me to do
> anyway!?  Since I'm new to contributing to anything open source at
> this level, I wondered if I was doing too much in one PR anyway?   Are
> you OK with me bundling it together? (that is, if I can work out the
> feedkeys issue soon - then its easier for me to bundle it together at
> this stage - if its OK with you.)  

In my opinion we should have a way to create MS-Windows events and
messages directly, with any information that we want to test with.  With
that we can test the code that does the decoding.

E.g., if we can fill a MOUSE_EVENT_RECORD then the code in
decode_mouse_event() can be tested.

-- 
Eight Megabytes And Continually Swapping.

zewpo:
Thanks again @brammool 

1. Function "feed_mswin_input".

> There currently is no code to convert byte codes to Windows messages and
> events, and adding this appears to be a lot of work and not really what
> we want to do.

Yes, agreed now, but well, I did create "feed_mswin_input()" to do this, and it does work.  It's not 100%, and yeah, it did take me a few days to get it this far.   Anyway, it does work for all printable ascii characters [32 to 127], and many Vim mouse sequences too, for example I was able to add this to mouse.vim, TerminalEscapeCode()
```
  elseif &term ==# 'win32'
	return printf("\<Esc>[<%d;%d;%d%s", a:code, a:col, a:row, a:m)
  endif
```
And that worked well also, ie; mouse.vim calls this sequence with feedkeys() and my change converts those Vim escape sequences into MOUSE_EVENT_RECORDs. 

My initial motivation here was to be able to make use of existing test scripts that are already written for testing the mouse, for example that are called from test_termcodes.vim.  These tests are now able to test my changes related to decoding MOUSE_EVENT_RECORD.

With my change, quite a few Tests in the existing test_termcodes.vim works this way now in windows console, and I think that might be where the value is.

But, as I'm learning, feedkeys() would still need to be able to handle even more of the specific Vim escape sequences - to be able to work in windows fully with test_termcodes.vim.

2. Function "test_mswin_event"

> In my opinion we should have a way to create MS-Windows events and
> messages directly, with any information that we want to test with.  With
> that we can test the code that does the decoding.
> 
> E.g., if we can fill a MOUSE_EVENT_RECORD then the code in
> decode_mouse_event() can be tested.
> 

Yes, the new function; test_mswin_event() achieves this.   So, I could possibly clean up this PR to make that as the goal, without worrying about feedkeys yet.

But, again, here is the but.   This means that even more new test scripts would need to be written using "test_mswin_event()", if we aren't able to make use of the existing test_termcodes.vim stuff.  The existing test_termcodes.vim would be good at testing my changes related to decoding MOUSE_EVENT_RECORD too,  but doing that needs my proposed changes to feedkeys() to work.

Anyway, I guess I'm trying to justify the work I've already put into developing "feed_mswin_input()" to enable lowlevel feedkeys() work in windows consoles.  I'm hesitant to discard it at the moment, so just double checking with you if you don't think there is much value in me doing it like that.    

In other words, I mean, I made "feedkeys()" able to make use of "feed_mswin_input()" internally, so that we wouldn't have to rewrite a lot of those existing test scripts.  Does that make sense?    

I think none of this would be an issue if feedkeys() wasn't already used by some test scripts on windows - contrary to what the doco claims.    

I'd like to propose the following.   I stash away all my changes to feedkeys() including the new function "feed_mswin_input()",  and submit this PR with them, and just the "test_mswin_event()" changes.    Then this PR will come along with only minimal testing designed to test that particular function.   So, the testing in this PR won't necessarily cover all of my previous PRs in the areas that involve processing the internal buffer sequences, eg. how I made horizontal scrolling work in all terminals that support that - won't be fully tested with this PR.

Does that sound like a reasonable way for me to progress this stuff forward?

zewpo:
typo:    

>  I stash away all my changes to feedkeys() including the new function "feed_mswin_input()",  and submit this PR **without** them

zewpo:
PS.  I would then submit the feedkeys changes in a separate PR.

brammool:
[...]

> But, again, here is the but.   This means that even more new test
> scripts would need to be written using "test_mswin_event()", if we
> aren't able to make use of the existing test_termcodes.vim stuff.  The
> existing test_termcodes.vim would be good at testing my changes
> related to decoding MOUSE_EVENT_RECORD too,  but doing that needs my
> proposed changes to feedkeys() to work.

It depends on what you refer to.  Some parts of test_termcodes.vim use
middle-level functions to create a mouse code event.  E.g.
MouseLeftClick().  The current implementation in mouse.vim uses
MouseLeftClickCode() and passes that to feedkeys().  For MS-Windows
getting that escape sequence is pointless, you would use a Windows event
instead.

This is probably one of the simpler examples, feedkeys() is also used
directly.  We'll have to see if it is useful to do the equivalent test
for MS-Windows.

> Anyway, I guess I'm trying to justify the work I've already put into
> developing "feed_mswin_input()" to enable lowlevel feedkeys() work in
> windows consoles.  I'm hesitant to discard it at the moment, so just
> double checking with you if you don't think there is much value in me
> doing it like that.    

If you refer to using the escape sequences, like they are produced in
mouse.vim, and turning these into Windows messages, I think for the long
term we don't want to do this.  It would depend on both xterm (alike)
specifics and MS-Windows stuff, which is a weird combination.

> In other words, I mean, I made "feedkeys()" able to make use of
> "feed_mswin_input()" internally, so that we wouldn't have to rewrite a
> lot of those existing test scripts.  Does that make sense?    

Those test scripts probably need to be rewritten anyway.  If we can do
this by using the intermediate level functions, such as MouseLeftClick()
and the like, that would be best.

> I think none of this would be an issue if feedkeys() wasn't already
> used by some test scripts on windows - contrary to what the doco
> claims.    
> 
> I'd like to propose the following.   I stash away all my changes to
> feedkeys() including the new function "feed_mswin_input()",  and
> submit this PR with them, and just the "test_mswin_event()" changes.
> Then this PR will come along with only minimal testing designed to
> test that particular function.   So, the testing in this PR won't
> necessarily cover all of my previous PRs in the areas that involve
> processing the internal buffer sequences, eg. how I made horizontal
> scrolling work in all terminals that support that - won't be fully
> tested with this PR.
> 
> Does that sound like a reasonable way for me to progress this stuff
> forward?

I'm afraid I don't know what that means.  I would prefer making a first
step in the right direction.  That it won't test much is less relevant,
that can be added later.  At the end of mouse.vim there are several
MouseWheel functions.  If we can make the equivalent of these using
Windows events instead of terminal-specific escape codes, I think we are
heading in the right direction.


-- 
You were lucky. We lived for three months in a brown paper bag in a 
septic tank. We used to have to get up at six o'clock in the morning, 
clean the bag, eat a crust of stale bread, go to work down mill for 
fourteen hours a day week in-week out. When we got home, our Dad
would thrash us to sleep with his belt!

zewpo:
Thanks @brammool 
Thanks for answering my questions so clearly.
I will shout out again after I've cleaned up this PR.

zewpo:
Hi @brammool 

Do you know if this code intentional?

https://github.com/zewpo/vim/blob/4cd45f1408321905019614b40a2173fe0456a96d/src/os_win32.c#L336-L339

I wonder why it runs PeekConsoleInputW directly after GetNumberOfConsoleInputEvents returns 0 dwEvents.

Won't that always return nothing?  That seems OK if dwEvents is 0 anyway.

But...  

if I'm reading it right, is when GetNumberOfConsoleInputEvents  returns dwEvents > 0, then it will go straight to ReadConsoleInputW(), which will read and POP values from the buffer, and it will do this even though it was called by peek_console_input().   My working assumption was that peek should not pop the buffer.   But, if this is intentional for any reason then I dont want to break it.

I'm rejigging the read_console_input function a bit to work with an internal buffer.  (where I can inject the synthesized INPUT_RECORDs at test time,)  and I don't want my changes to miss any of the pathways through the code.

brammool:
> Hi @brammool 
> 
> Do you know if this code intentional?
> 
> https://github.com/zewpo/vim/blob/4cd45f1408321905019614b40a2173fe0456a96d/src/os_win32.c#L336-L339
> 
> I wonder why it runs PeekConsoleInputW directly after
> GetNumberOfConsoleInputEvents returns 0 dwEvents.
> 
> Won't that always return nothing?  That seems OK if dwEvents is 0 anyway.

No idea.  I haven't spent much time on the MS-Windows code.  Various
people have worked on it over time.  One thing to keep in mind: trying
to clean up this code to make it work "as documented" tends to break
things.  Better rely on how it currently works.

> But...  
> 
> if I'm reading it right, is when GetNumberOfConsoleInputEvents
> returns dwEvents > 0, then it will go straight to ReadConsoleInputW(),
> which will read and POP values from the buffer, and it will do this
> even though it was called by peek_console_input().   My working
> assumption was that peek should not pop the buffer.   But, if this is
> intentional for any reason then I dont want to break it.

Well, it's working, so don't change it.  The code is hardly commented,
thus it's hard to see what happens.  I would assome that when peeking
the events are buffered somewhere and not actually used.  But it may be
needed to do some processing to see what is the next event.

> I'm rejigging the read_console_input function a bit to work with an
> internal buffer.  (where I can inject the synthesized INPUT_RECORDs at
> test time,)  and I don't want my changes to miss any of the pathways
> through the code.

I'm not much help here, since I hardly touched this code.  Perhaps
someone else can give hints.


-- 
Q:  Why do ducks have flat feet?
A:  To stamp out forest fires.

Q:  Why do elephants have flat feet?
A:  To stamp out flaming ducks.

zewpo:
@brammool 

> Well, it's working, so don't change it.

Yes, since we can't be certain, so I'm only submitting minimal changes to get this working.  

I won't submit the refactoring that I wanted to do there for now.  And will leave that for another day.

This PR is finally passing CI tests.  

I did end up making a minor change to feedkeys - and that is, Windows terminals will ignore the L flag, and so will always use the typeahead buffer as if the L flag was not included in the mode arg of feedkeys.

I tracked down the one and only line of test scripts that was using feedkeys mode L (lowlevel) in the windows terminal - and have made sure that continues to work.   So, the one exception to windows terminals ignoring the L flag, and that is Control-C will get added directly to the input_buf.

test_vim9_script.vim has the following;

```
def Test_interrupt_loop()
  var caught = false
  var x = 0
  try
    while 1
      x += 1
      if x == 100
        feedkeys("\<C-C>", 'Lt')
      endif
    endwhile
  catch
    caught = true
    assert_equal(100, x)
  endtry
  assert_true(caught, 'should have caught an exception')
  # consume the CTRL-C
  getchar(0)
enddef
```

This is the only place that does not check if it is in Windows or not.  And, it was lucky that it was working.   Every other instance of feedkeys mode L in the repository (all 141 of them) is under a "CheckUnix" or equivalent. 


For now, I'm just working on increasing the test coverage.

zewpo:
I got no idea why test_channel is randomly failing.   I don't think that is caused by this PR, but that failing is interfering with us being able to see the increases in test coverage that this PR should provide.    Oh well, will try again later.

brammool:
> I got no idea why test_channel is randomly failing.   I don't think
> that is caused by this PR, but that failing is interfering with us
> being able to see the increases in test coverage that this PR should
> provide.    Oh well, will try again later.

It has been flaky for a while.  I have no clue how to reduce flakiness.
Perhaps if someone can reproduce the problem (e.g. by keeping the CPU
busy or enabling a profiler) then we can find a way to improve this.
On my system I often enable valgrind to catch timing problems, since
valgrind makes execution a lot slower.  But that only works in some
situations.

It may also help to try at another time of day, when the CI machines are
less busy.

-- 
Have you heard about the new Barbie doll?  It's called Divorce
Barbie.  It comes with most of Ken's stuff.

zewpo:
@brammool 

It's finally working nicely, codecov/project says it "will increase coverage by **81.61%.**"  :)

Let me clean up some of the verbose comments, and then I will take off [WIP] prefix.

I'd guess you might have some things to say about the code style, naming, organisation etc, I've tried to fit in with Vim's code style, but there are mixtures especially in the MS-windows code naming conventions.

zeertzjq:
The coverage on HEAD is 0 for some reason

PR #11630, opened 2022-11-28
Fix typo in autocommand name

zeertzjq:


src/option.c
```diff
@@ -1618,7 +1618,7 @@ do_set_string(
 
 #if defined(FEAT_EVAL)
     if (*errmsg == NULL)
-	trigger_optionsset_string(opt_idx, opt_flags, saved_origval,
+	trigger_optionset_string(opt_idx, opt_flags, saved_origval,
```
Shane-XB-Qian, line 1621: was it possible meant 'option[s]' ?

PR #9003, opened 2021-10-16
Fix nginx indenting

flosacca:
Currently the indenting for nginx files breaks in certain cases, see https://github.com/chr4/nginx.vim/issues/3. This PR fixes it.

PR #11645, opened 2022-12-01
Fix indention in nginx plugin

chr4:
Currently the indenting for nginx files breaks in certain cases, see chr4/nginx.vim#3. This PR fixes it.

This was handed in earlier, I know found the time to finally review and merge this upstream:
- https://github.com/vim/vim/pull/9003
- https://github.com/chr4/nginx.vim/pull/20

### DIFF
--- a/runtime/indent/nginx.vim
+++ b/runtime/indent/nginx.vim
@@ -1,19 +1,78 @@
 " Vim indent file
 " Language: nginx.conf
 " Maintainer: Chris Aumann <me@chr4.org>
-" Last Change:  2022 Apr 06
+" Last Change:  2022 Dec 01
 
-if exists("b:did_indent")
-    finish
+" Only load this indent file when no other was loaded.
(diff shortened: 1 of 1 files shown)


brammool:
A few suggestions:
Do not use the exclamation mark after :function.  This would not report errors for defining the same function twice.
It is no longer needed when reloading the script.

You can try writing the functions in Vim9 script, using :def will make them much faster.
You can keep the script itself in legacy syntax.

chr4:
Thanks for the suggestion! I use removed the `!` from the function calls.

[The contributor](https://github.com/chr4/nginx.vim/pull/20#issuecomment-1349530960) of the function wants to keep Vim-8 syntax for compatibility reasons. Objections?

Further feedback welcome!

brammool:
It's OK to use legacy Vim script.  But using a Vim9 script compiled function is likely to be faster.

It would be nice to add some tests.  But that can be added later.

PR #11646, opened 2022-12-01
feat: Add support for the Nix programming language.

jsoo1:
The Nix language and package manager is now quite a large project (nixpkgs is of the largest and most active projects on GitHub, see https://github.com/NixOS/nixpkgs). It uses a domain-specific language for configuration. This PR adds syntax for the language.

The syntax is pulled from https://github.com/LnL7/vim-nix which is under MIT license. I have included attribution to the original author and a referred back to the GitHub repository.  I reached out to the maintainer a few months ago to ask about including their code in vim.  I got no response (https://github.com/LnL7/vim-nix/issues/47), so I think it should be ok to proceed including this PR.

Note that I found a maintainer for the syntax file that is not me: @equill (james@electronic-quill.net) volunteered as the contact who is listed as maintainer in runtime/syntax/nix.vim.

@equill would you mind acknowledging that your contact information is correct and that you will be the point of contact for the Nix syntax highlighting?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -394,6 +394,7 @@ runtime/syntax/n1ql.vim			@pr3d4t0r
 runtime/syntax/netrw.vim		@cecamp
 runtime/syntax/nginx.vim		@chr4
 runtime/syntax/ninja.vim		@nico
+runtime/syntax/nix.vim			@equill
 runtime/syntax/nroff.vim		@jmarshall
 runtime/syntax/nsis.vim			@k-takata
 runtime/syntax/openvpn.vim		@ObserverOfTime
--- a/runtime/syntax/nix.vim
+++ b/runtime/syntax/nix.vim
@@ -1,0 +1,204 @@
+" Vim syntax file
+" Language:        Nix
+" Maintainer:	   James Fleming <james@electronic-quill.net>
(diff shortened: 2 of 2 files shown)


brammool:
Can you please also add an entry in .github/CODEOWNERS ?

jsoo1:
Yes I can, though I'll be out till Monday.

jsoo1:
Ok I've added @equill as owner of runtime/syntax/nix.vim. I think there may be other nix support files that could be pulled from https://github.com/LnL7/vim-nix but I just don't know enough about vim to do the work.

brammool:
Thanks, I'll include it.

brammool:
Noticed afterwards: The file uses line continuation, it should set the 'cpo' option and restore it.

jsoo1:
> Noticed afterwards: The file uses line continuation, it should set the 'cpo' option and restore it.

Do you want a separate patch for that? If so, @equill would you mind doing the honors?

brammool:
> > Noticed afterwards: The file uses line continuation, it should set
> > the 'cpo' option and restore it.
> 
> Do you want a separate patch for that? If so, @equill would you mind
> doing the honors?

I fixed it in my copy, it's the standard lines.  I'll make a runtime
files update soon, please sync and make any further changes on top of
that.

-- 
249. You've forgotten what the outside looks like.

PR #11656, opened 2022-12-04
better srand

mattn:
Currrent implementation of init_srand for Windows makes wrong seed value if vim run few times in same seconds.

```
vim -u NONE -V -es -c "echo rand(). ' '" -c quit && vim -u NONE -V -es -c "echo rand(). ' '" -c quit
```

image

### DIFF
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -8156,13 +8156,18 @@ init_srand(UINT32_T *x)
 	}
     }
     if (dev_urandom_state != OK)
-	// Reading /dev/urandom doesn't work, fall back to reltime() or time().
+	// Reading /dev/urandom doesn't work, fall back to:
+	//   randombytes_random()
+	//   reltime()
+	//   time() ^ getpid()
 #endif
     {
-#ifdef FEAT_RELTIME
+#if defined(FEAT_SODIUM)
+	*x = randombytes_random();
+#elif defined(FEAT_RELTIME)
 	proftime_T res;
 	profile_start(&res);
-#  ifdef MSWIN
+#  if defined(MSWIN)
 	*x = (UINT32_T) res.LowPart;
 #  else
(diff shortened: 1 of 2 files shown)


brammool:
vim_time() only has second accuracy.  Much better would be to use reltime(), which should have micro second accuracy.
But it is only available with FEAT_RELTIME.
Best is to use:
1. if FEAT_RELTIME is defined, use call f_reltime() and use the second item of the list (should be micro seconds).
2. if FEAT_RELTIME is not defined, or f_reltime() does not return a list with two numbers, use the value of vim_time()
3. Xor with mch_get_pid()

mattn:
Updated

chrisbra:
Perhaps, you may also want to use `randombytes_random(void);` from libsodium (`ifdef FEAT_SODIUM`) , if build with sodium enabled

src/evalfunc.c
```diff
@@ -8156,9 +8156,26 @@ init_srand(UINT32_T *x)
...
     if (dev_urandom_state != OK)
-	// Reading /dev/urandom doesn't work, fall back to time().
+	// Reading /dev/urandom doesn't work, fall back to:
+	//   randombytes_random()
+	//   reltime()
+	//   time() ^ getpid()
+#endif
+    {
+#if defined(FEAT_SODIUM)
+	*x = randombytes_random();
```
chrisbra: There should be a `sodium_init()` before this, like there: https://github.com/vim/vim/blob/65214053f6a4bdb1099404a4b5dccb85ca3e5c7c/src/crypt.c#L563

dpelle:
`:help srand()` currently says:

```
                - If {expr} is not given, seed values are initialized by
                  reading from /dev/urandom, if possible, or using time(NULL)
                  a.k.a. epoch time otherwise; this only has second accuracy.
```
This would need to be updated according to the implementation.

PR #11673, opened 2022-12-07
Fix mouse shape test timer issue in CI

ychin:
In MacVim CI, the `Test_mouse_shape_after_failed_change` tests would fail due to test logic relying on precise timer sequencing. If any timer gets delayed, it will cause the logic to not work. Fix this by only starting a new timer when a previous step has completed (relying on callbacks), instead of scheduling all timer events right off the bat. This makes sure that we have the minimum of elapsed time between events. Also switched to vim9script because it's easier to do a callback style like this in it.

### DIFF
--- a/src/testdir/test_normal.vim
+++ b/src/testdir/test_normal.vim
@@ -3818,24 +3818,33 @@ func Test_mouse_shape_after_failed_change()
   CheckCanRunGui
 
   let lines =<< trim END
+    vim9script
     set mouseshape+=o:busy
     setlocal nomodifiable
-    let g:mouse_shapes = []
+    var mouse_shapes = []
 
-    func SaveMouseShape(timer)
-      let g:mouse_shapes += [getmouseshape()]
-    endfunc
+    def SaveMouseShape()
+      mouse_shapes += [getmouseshape()]
+    enddef
 
-    func SaveAndQuit(timer)
-      call writefile(g:mouse_shapes, 'Xmouseshapes')
+    def SaveAndQuit()
+      writefile(mouse_shapes, 'Xmouseshapes')
       quit
-    endfunc
+    enddef
 
-    call timer_start(50, {_ -> feedkeys('c')})
-    call timer_start(100, 'SaveMouseShape')
-    call timer_start(150, {_ -> feedkeys('c')})
-    call timer_start(200, 'SaveMouseShape')
-    call timer_start(250, 'SaveAndQuit')
+    feedkeys('c')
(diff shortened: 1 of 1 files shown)


zeertzjq:
BTW, does the `sleep 300m` below need change?

src/testdir/test_normal.vim
```diff
@@ -3818,24 +3818,33 @@ func Test_mouse_shape_after_failed_change()
...
+      mouse_shapes += [getmouseshape()]
+    enddef
 
-    func SaveAndQuit(timer)
-      call writefile(g:mouse_shapes, 'Xmouseshapes')
+    def SaveAndQuit()
+      writefile(mouse_shapes, 'Xmouseshapes')
       quit
-    endfunc
+    enddef
```
zeertzjq: I guess the body of these two functions can just be moved into the lambdas.

ychin: Yeah that's true.

ychin:
> BTW, does the `sleep 300m` below need change?

Hmm, I'm not sure. From experience in GitHub Actions, it didn't seem like I need to do that in order to pass MacVim CI. I'm still not 100% sure what's wrong with the stability of timers when running in CI VMs (I never reproduce them when running locally) but I'm guessing it's some bad interactions with the VM. It appears that the timer-scheduled actions will get bunched up together instead of running sequentially in 50ms segments (and Vim also has a quirk where it ignores the ordering of the timer-scheduled events when this happens which is something I want to look further in).

But either way I think the extra 50ms from the 300ms timer is enough.

ychin:
FWIW this is the failed CI run in question: https://github.com/macvim-dev/macvim/actions/runs/3615779087/jobs/6093125753

Log output:

```
	command line..script /Users/runner/work/macvim/macvim/src/testdir/runtest.vim[533]..function RunTheTest[52]..Test_mouse_shape_after_failed_change line 27: Expected ['busy', 'arrow'] but got ['arrow']
```

Sometimes I would get `['arrow', 'busy']` because of the timer re-ordering issue.

brammool:
AFAIK the machines that run the Mac CI are very busy.  Quite often when a delay is needed, on the Mac machines a longer delay is needed.  I think Vim often has to wait to get CPU cycles, and when it eventually does then several timers (with different delays) get triggered at the same time.
Your solution avoids this problem.  However, it also means the real-life situation of timers triggering in sequence isn't tested.  Would require detecting that timers trigger too late, I suppose.

PR #11679, opened 2022-12-08
Make function name match what it actually does

zeertzjq:


brammool:
Why did you move sb_text_start_cmdline() ?

zeertzjq:
I just noticed that `sb_text_end_cmdline()` is not called after `theend`, so I think it doesn't make sense to call `sb_text_start_cmdline()` before `goto theend`.

brammool:
> I just noticed that `sb_text_end_cmdline()` is not called after
> `theend`, so I think it doesn't make sense to call
> `sb_text_start_cmdline()` before `goto theend`.

Hmm, this depends on there not being a "goto theend" after that line.
It does make the relation between the _end_ and the _start_ function
call clearer.  Not sure if they actually need to be called in pairs.

Better leave this change for another patch, it is not something that is
related to the renaming.

-- 
Edison's greatest achievement came in 1879, when he invented the
electric company.  Edison's design was a brilliant adaptation of the
simple electrical circuit: the electric company sends electricity
through a wire to a customer, then immediately gets the electricity
back through another wire

PR #11684, opened 2022-12-09
Add 'showcmd' statusline item and 'showcmdloc' option

luukvbaal:
This statusline item is mostly intended for use with `cmdheight == 0` in Neovim. Nevertheless, users have expressed interest in displaying `showcmd` in their [statusline](https://vi.stackexchange.com/questions/2663/embed-showcmd-in-airline) and even [tabline](https://vi.stackexchange.com/questions/16857/showcmd-on-first-line-instead-of-last-line).

Just checking if you are open to including this to avoid future naming conflicts of the statusline item. The item name is still up for debate, @justinmk [suggested](https://github.com/neovim/neovim/pull/21202#discussion_r1043888844) dot`.` instead.

I assume potential users of this item would `set noshowcmd`, therefore I decoupled the showcmd routines from `p_sc`, except for the actual drawing in the cmdline.

I tried adding a non-screendump test to `Test_statusline()` but only the visual mode test passes. The pending operator test yields an empty statusline. Does the `call assert_match()` clear the showcmd buffer? That's what it seems like atleast:
```vim
  " %S: 'showcmd' content.
  set statusline=%S
  call assert_match('^\s*$', s:get_statusline())
  call feedkeys("\<C-V>l", "xt")
  call assert_match('^1x2\s*$', s:get_statusline())
  call feedkeys("\<Esc>1234", "xt")
  call assert_match('^1234\s*$', s:get_statusline())
  call feedkeys("\<Esc>", "xt")
  " result:
  " RunTheTest[44]..Test_statusline line 130: Pattern '^\\[1234\\]s*$' does not match '\[  occurs 80 times]'
```

Therefore I grouped them as 2 screendump tests instead. It works as intended in practice.

Redrawing the statusline upon command input caused a test failure in `Test_aa_terminal_focus_events()`, updated the dump files to reflect the statusline update.


zeertzjq:
`feedkeys()` with `x` flag needs to cancel pending operator to finish, so it is not possible to stop at op-pending mode after that. I guess it may be possible to check the screen in op-pending mode using timers, but that may make the test confusing.

zeertzjq:
> updated the dump files to reflect the statusline update.

Did you forget to `git add` the updated dump files?

brammool:
Please first add documentation.  Reverse engineering the code to guess what it's supposed to do isn't the right way to add a new feature.
One problem I can spot is that it's redrawing the status lines even when the showcmd item isn't in there.
Also, what happens when 'showcmd' is off is unclear.

luukvbaal:
> Please first add documentation.
 
Right, completely forgot sorry.

> One problem I can spot is that it's redrawing the status lines even when the showcmd item isn't in there. 

True, should 'statusline' be parsed for this item? Is there precedent for this?

> Also, what happens when 'showcmd' is off is unclear.

When 'showcmd' is off, we still fill the `showcmd_buf`, because the user might have the item in use. If we parse the status/tabline we could set a global and avoid unnecessary work I guess.

luukvbaal:
> If we parse the status/tabline we could set a global and avoid unnecessary work I guess.

I guess there's no reason not to, latest commit sets a global boolean when the 'showcmd' item is encountered. Status line will only be redrawn when TRUE, so reverted the terminal focus testdump change.

brammool:
> > Please first add documentation.
>  
> Right, completely forgot sorry.
> 
> > One problem I can spot is that it's redrawing the status lines even when the showcmd item isn't in there. 
> 
> True, should 'statusline' be parsed for this item? Is there precedent
> for this?

I don't think so.  Currently there is no item that changes on every key
stroke, AFAIK.

Parsing the option only works when when it doesn't start with "!%".  
I think the only way would be to draw the status line, and set a flag if
the "showcmd" value is used.  But then perhaps it isn't always called
(e.g. if there is not enough room), that makes it unreliable.

> > Also, what happens when 'showcmd' is off is unclear.
> 
> When 'showcmd' is off, we still fill the `showcmd_buf`, because the
> user might have the item in use. If we parse the status/tabline we
> could set a global and avoid unnecessary work I guess.

If 'showcmd' would have been a number or string option we could have
done something, but since it's a boolean we can't change it.

We could tackle both problems with a 'showcmdwhere' option (better
name?), to tell Vim where the command will be displayed, thus knowing
when to update status lines.  Since the user has to switch 'showcmd'
off, setting another option isn't really a problem.

-- 
MAN:     You don't frighten us, English pig-dog!  Go and boil your bottoms,
         son of a silly person.  I blow my nose on you, so-called Arthur-king,
         you and your silly English K...kaniggets.
   He puts hands to his ears and blows a raspberry.

luukvbaal:
> Parsing the option only works when when it doesn't start with "!%". I think the only way would be to draw the status line, and set a flag if the "showcmd" value is used.

Yeah I realized, the latest commit does just that. 

> We could tackle both problems with a 'showcmdwhere' option (better
name?), to tell Vim where the command will be displayed, thus knowing when to update status lines.

Right, let me know if that is a route you want to take to include this feature. I can update accordingly. Do we want to facilitate displaying this item in both the status line and tabline? Or even multiple(status/tab/cmdline) at the same time? In which case I suppose it should be a comma separated string option.

> Since the user has to switch 'showcmd' off, setting another option isn't really a problem.

I guess in theory they don't **have** to but it would make the most sense yeah.

brammool:
> > Parsing the option only works when when it doesn't start with "!%".
> > I think the only way would be to draw the status line, and set a
> > flag if the "showcmd" value is used.
> 
> Yeah I realized, the latest commit does just that. 
> 
> > We could tackle both problems with a 'showcmdwhere' option (better
> >name?), to tell Vim where the command will be displayed, thus knowing
> >when to update status lines.
> 
> Right, let me know if that is a route you want to take to include this
> feature. I can update accordingly. Do we want to facilitate displaying
> this item in both the status line and tabline? Or even
> multiple(status/tab/cmdline) at the same time? In which case I suppose
> it should be a comma separated string option.

Would someone really want it in multiple places?  That sounds unusual.
I would not implement that at first, but make the option value in a way
that it's possible in the future.

When you say "status line" it doesn't say where it shows up.  Only the
status line for the most bottom-right window?  Otherwise it's specified
by the 'statusline' option.  But it could also be done when 'statusline'
is empty.

I find it hard to imagine showcmd showing up in the tabline.  I suppose
some space would need to be reserved for it.  Then the choice is on the
far left or the far right.

> > Since the user has to switch 'showcmd' off, setting another option
> > isn't really a problem.
> 
> I guess in theory they don't **have** to but it would make the most
> sense yeah.

With the extra option, probably 'showcmdloc', then 'showcmd' would still
be on.  If it is off then it won't show anywhere.  That's simple.

'showcmdloc' would now be one word, such as "statusline",
"tablineright", etc.  If we ever would want to show it in more than one
place the option can become a comma separated list of these words.

I still have the feeling that very few users are going to use this, thus
consider it low priority.

-- 
ARTHUR: (as the MAN next to him is squashed by a sheep) Knights!  Run away!
   Midst echoing shouts of "run away" the KNIGHTS retreat to cover with the odd
   cow or goose hitting them still.  The KNIGHTS crouch down under cover.

luukvbaal:
> Would someone really want it in multiple places?  That sounds unusual.
I would not implement that at first, but make the option value in a way
that it's possible in the future.

> When you say "status line" it doesn't say where it shows up.  Only the
status line for the most bottom-right window?  Otherwise it's specified
by the 'statusline' option.  But it could also be done when 'statusline'
is empty.

> I find it hard to imagine showcmd showing up in the tabline.  I suppose
some space would need to be reserved for it.  Then the choice is on the
far left or the far right.

Yes showing it in multiple places would be unusual. The way I implemented it currently and what seems to make the most sense to me is to redraw the status line for `curwin`. Whether it shows up is not our concern I don't think. It would indeed be up to the user to reserve space, if they find the item high enough priority.

> With the extra option, probably 'showcmdloc', then 'showcmd' would still
be on.  If it is off then it won't show anywhere.  That's simple.

> 'showcmdloc' would now be one word, such as "statusline",
"tablineright", etc.  If we ever would want to show it in more than one
place the option can become a comma separated list of these words.

> I still have the feeling that very few users are going to use this, thus
consider it low priority.

Very possible, again I only made the PR here to avoid future naming conflicts with Neovim, and get a feel for what item-symbol you would choose. But if we are going to add the feature here we have to add it in a way that makes sense. I think we may not need the 'showcmdloc' option. Since we now pass the option name to `build_stl_str_hl()`, we can simply set a flag for status/tabline and redraw accordingly in `display_showcmd()`.

brammool:
> Yes showing it in multiple places would be unusual. The way I
> implemented it currently and what seems to make the most sense to me
> is to redraw the status line for `curwin`. Whether it shows up is not
> our concern I don't think. It would indeed be up to the user to
> reserve space, if they find the item high enough priority.

It does make sense to show it in the current window.  That's closest to
where the user is looking.

When 'statusline' is empty it should still show up.  Left of the ruler
would be good.

> Very possible, again I only made the PR here to avoid future naming
> conflicts with Neovim, and get a feel for what item-symbol you would
> choose. But if we are going to add the feature here we have to add it
> in a way that makes sense. I think we may not need the 'showcmdloc'
> option. Since we now pass the option name to `build_stl_str_hl()`, we
> can simply set a flag for status/tabline and redraw accordingly in
> `display_showcmd()`.

In my opinion we do need to add the 'showcmdloc' option.  Without it,
the user would have to reset 'showcmd' and add the item to the
'statusline' option.  This would black adding 'showcmdloc' later,
because then 'showcmd' would remain on, thus it can't be backwards
compatible.

-- 

brammool:
I wrote:

> > Yes showing it in multiple places would be unusual. The way I
> > implemented it currently and what seems to make the most sense to me
> > is to redraw the status line for `curwin`. Whether it shows up is not
> > our concern I don't think. It would indeed be up to the user to
> > reserve space, if they find the item high enough priority.
> 
> It does make sense to show it in the current window.  That's closest to
> where the user is looking.
> 
> When 'statusline' is empty it should still show up.  Left of the ruler
> would be good.
> 
> > Very possible, again I only made the PR here to avoid future naming
> > conflicts with Neovim, and get a feel for what item-symbol you would
> > choose. But if we are going to add the feature here we have to add it
> > in a way that makes sense. I think we may not need the 'showcmdloc'
> > option. Since we now pass the option name to `build_stl_str_hl()`, we
> > can simply set a flag for status/tabline and redraw accordingly in
> > `display_showcmd()`.
> 
> In my opinion we do need to add the 'showcmdloc' option.  Without it,
> the user would have to reset 'showcmd' and add the item to the
> 'statusline' option.  This would black adding 'showcmdloc' later,

s/black/block/

> because then 'showcmd' would remain on, thus it can't be backwards
> compatible.

-- 
press CTRL-ALT-DEL for more information

luukvbaal:
>When 'statusline' is empty it should still show up.  Left of the ruler would be good.

Latest commit adds the 'showcmdloc' option. Will need to figure out how to add this to the default statusline if 'showcmdloc' == "statusline".

luukvbaal:
> Will need to figure out how to add this to the default statusline if 'showcmdloc' == "statusline".

Done in the latest commit, as well as for the default tabline. (I don't see it being all that useful in the tabline, I just added it because I came across https://vi.stackexchange.com/questions/16857/showcmd-on-first-line-instead-of-last-line). 

I think the attributes for the showcmd text in the default tabline should be `attr_nosel` without underline. I tried `attr_nosel & ~HL_UNDERLINE` but that also changed the color. Any way I can use `attr_nosel` but remove the underline?

luukvbaal:
> Any way I can use attr_nosel but remove the underline?

I guess it doesn't make sense to forcefully remove the underline from a highlight group (didn't actually figure out how to do that). This is finished then as far as I'm concerned. Tests for the default status/tabline are still missing. If you are open to including this in it's current form I could add additional tests.

brammool:
I thought I had already mentioned this: when 'showcmd' is off then the command should not be displayed anywhere.
It is too surprising that with 'noshowcmd' and then 'showcmdloc' set the command shows anyway.

The default value for 'showcmdloc' should be "last".  Currently you could replace it with "statusline" or "tabline".
If we ever decide to show it in multiple places, it could be "last,statusline", for example.

Coding style: Please put the opening curly on the next line.

src/buffer.c
```diff
@@ -4775,6 +4775,11 @@ build_stl_str_hl(
 	    get_rel_pos(wp, str, TMPLEN);
 	    break;
 
+	case STL_SHOWCMD:
+	    if (p_sc && STRCMP(opt_name, p_sloc) == 0)
```
luukvbaal, line 4779:
```c
case STL_SHOWCMD:
	    if (p_sc && STRCMP(opt_name, p_sloc) == 0)
		str = showcmd_buf;
	    break;
```

Not sure if this check is needed, I think `showcmd_buf` will always be empty here if those conditions are false anyways.

luukvbaal:
> I thought I had already mentioned this: when 'showcmd' is off then the command should not be displayed anywhere.
It is too surprising that with 'noshowcmd' and then 'showcmdloc' set the command shows anyway.

> The default value for 'showcmdloc' should be "last". Currently you could replace it with "statusline" or "tabline".
If we ever decide to show it in multiple places, it could be "last,statusline", for example.

> Coding style: Please put the opening curly on the next line.

Done and additional tests added.

luukvbaal:
Should the tests be in `test_normal.vim` instead? Existing `showcmd` test is in there.

brammool:
> Should the tests be in `test_normal.vim` instead? Existing `showcmd`
> test is in there.

test_normal.vim is already very big.  Perhaps we have enough testing for
this feature to justify a new test file?  In the past we thought the
overhead of starting a new Vim instance would matter, but now it matters
more that if a test fails, e.g. there is a memory leak, it is limited in
what might cause it.

Otherwise, test_statusline.vim would also be related.

-- 
E  M  A  C  S
s  e  l  o  h
c  t  t  n  i
a  a     t  f
p        r  t
e        o
         l

luukvbaal:
> test_normal.vim is already very big. Perhaps we have enough testing for this feature to justify a new test file? In the past we thought the overhead of starting a new Vim instance would matter, but now it matters more that if a test fails, e.g. there is a memory leak, it is limited in what might cause it. Otherwise, test_statusline.vim would also be related.

Yeah depends on whether you want to group these tests under 'showcmd', in which case a `test_showcmd.vim` might be warranted. Currently I have placed the `showcmdloc=statusline` test in `test_statusline.vim` and the `showcmdloc=tabline` in `tabline.vim`, your call.

luukvbaal:
I wonder if adding a builtin `showcmd()` function, analogous to e.g. `mode()` and `searchcount()` will be more useful/idiomatic than adding a statusline item?

PR #11698, opened 2022-12-12
Reset postponed_split for failed <C-W>] commands

bobrippling:
This prevents `postponed_split` affecting later, unrelated commands.

To reproduce, prior to this commit:

In a help file, `Ctrl-W_Ctrl-]` on a tag that doesn't exist. e.g. `:h change.txt | /CTRL-@`
(Find the "CTRL-@" and `Ctrl-W_Ctrl-]` with your cursor over it)

The tag jump will fail (this is ok/expected, since it tries to look for `CTRL-@)`)

Then, run a help command, such as `:h 'mod'`.
The help will open in a new window, not the current one

### DIFF
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1911,4 +1911,16 @@ function Test_splitkeep_status()
   call VerifyScreenDump(buf, 'Test_splitkeep_status_1', {})
 endfunction
 
+function Test_new_help_window_on_error()
+  help change.txt
+  execute "normal! /CTRL-@\<CR>"
+  silent! execute "normal! \<C-W>]"
+
+  let wincount = winnr('$')
+  help 'mod'
+
+  call assert_equal(wincount, winnr('$'))
+  call assert_equal(expand("<cword>"), "'mod'")
+endfunction
+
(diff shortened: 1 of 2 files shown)


brammool:
> This prevents `postponed_split` affecting later, unrelated commands.
> 
> To reproduce, prior to this commit:
> 
> In a help file, `Ctrl-W_Ctrl-]` on a tag that doesn't exist. e.g. `:h change.txt | /CTRL-@`
> (Find the "CTRL-@" and `Ctrl-W_Ctrl-]` with your cursor over it)
> 
> The tag jump will fail (this is ok/expected, since it tries to look for `CTRL-@)`)
> 
> Then, run a help command, such as `:h 'mod'`.
> The help will open in a new window, not the current one

Thanks for looking into this.

It would be good to have a regression test for this.  Since you can
describe how it currently fails, that should not be too difficult.
It should probably go in src/testdir/test_window_cmd.vim.


-- 
FIRST HEAD: All right! All right! We'll kill him first and then have tea and
            biscuits.

bobrippling:
Thanks for the feedback - a regression test's been added. I'm interested in what you think about how reliant it is on the existing help text, I believe I've made it somewhat resilient to changes to the help files.

PR #10142, opened 2022-04-10
Resolve supercollider - scala - scdoc conflicts

ranjithshegde:
Supercollider filtypes `*.sc` and `*.scd` conflict with Scala and scdoc respectively.
At the moment the plugins that implement supercollider (for either vim or neovim) simply override the filetypes which is not an elegant solution.

This PR is an attempt at fixing it with regex. It also adds the other supercollider filetype `*.quark`

EDIT: Apologies but I do not know how to add tests for regex detected filetypes in `testdir/test_filetype.vim`. Are there any guides or examples?

PR #11699, opened 2022-12-13
fix: avoid conflict with supercollider and scala files

ckipp01:
Since the changes introduced in
https://github.com/vim/vim/pull/10142/commits/919ebcb8cbb5710eeccbf9750633e56572ccd471
I've seen an increasing amount of Scala users report that their Scala
scripts are being treated as supercollider files. There is one line in
the check there is problematic since it treats the following (valid
Scala) as supercollider code:

```scala
val example : String = "hello"
```

It's the ` : ` that is problematic here since looking at the notes this
can also be used for `Class : Method` is supercollider, but is also
totally valid Scala. The rest of the checks can't be confused like this
and are safe, but this one is problematic. I think it should be fine to
remove this and instead rely on the other checks to find supercollider
usage where there can't be confusion. This is anecdotal since I'm not a
supercollider user, but searching around on GitHub I'm not really even
finding this patter for supercollider files, which again, I might be
misunderstanding their usage.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -816,7 +816,7 @@ enddef
 #       rely on checks that can't be confused.
 export def FTsc()
   for lnum in range(1, min([line("$"), 25]))
-    if getline(lnum) =~# 'var\s<\|classvar\s<\|\^this.*\||\w*|\|+\s\w*\s{\|\*ar\s'
+    if getline(lnum) =~# 'var\s<\|classvar\s<\|\^this.*\||\w\+|\|+\s\w*\s{\|\*ar\s'
       setf supercollider
       return
     endif
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -810,10 +810,13 @@ export def SQL()
 enddef
 
 # This function checks the first 25 lines of file extension "sc" to resolve
-# detection between scala and SuperCollider
+# detection between scala and SuperCollider.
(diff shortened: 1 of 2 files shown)


brammool:
> Since the changes introduced in
> https://github.com/vim/vim/pull/10142/commits/919ebcb8cbb5710eeccbf9750633e56572ccd471
> I&#39;ve seen an increasing amount of Scala users report that their Scala
> scripts are being treated as supercollider files. There is one line in
> the check there is problematic since it treats the following (valid
> Scala) as supercollider code:
> 
> ```scala
> val example : String = "hello"
> ```
> 
> It's the ` : ` that is problematic here since looking at the notes this
> can also be used for `Class : Method` is supercollider, but is also
> totally valid Scala. The rest of the checks can't be confused like this
> and are safe, but this one is problematic. I think it should be fine to
> remove this and instead rely on the other checks to find supercollider
> usage where there can't be confusion. This is anecdotal since I'm not a
> supercollider user, but searching around on GitHub I'm not really even
> finding this patter for supercollider files, which again, I might be
> misunderstanding their usage.

I hope some users, especially supercollider users, will try this out.

Please add a comment in ft.vim why this check can't be used.  Giving the
Scala and Supercollider example might be sufficient.


-- 
GALAHAD: No look, really, this isn't nescess ...
PIGLET:  We must examine you.
GALAHAD: There's nothing wrong with ... that.

ckipp01:
> Please add a comment in ft.vim why this check can't be used.  Giving the
Scala and Supercollider example might be sufficient.

Sure, I went ahead and added a note.

PR #11700, opened 2022-12-13
Remove 'r' and 'o' from formatoptions.

dundargoc:
I don't think these particular options should be set from a ftplugin as
I don't believe them to be relevant to the C or Vim languages. I know
it's possible to override this with the "after" directory, but I feel
like this is a more "correct" behavior by not setting these in the first
place.

### DIFF
--- a/runtime/ftplugin/c.vim
+++ b/runtime/ftplugin/c.vim
@@ -17,9 +17,8 @@ set cpo-=C
 
 let b:undo_ftplugin = "setl fo< com< ofu< cms< def< inc< | if has('vms') | setl isk< | endif"
 
-" Set 'formatoptions' to break comment lines but not other lines,
-" and insert the comment leader when hitting <CR> or using "o".
-setlocal fo-=t fo+=croql
+" Set 'formatoptions' to break comment lines but not other lines.
+setlocal fo-=t fo+=cql
 
 " These options have the right value as default, but the user may have
 " overruled that.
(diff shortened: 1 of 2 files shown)


brammool:
The options have the correct value.  For these languages a comment continues only
when the comment leader is inserted in the next line.  Without that you cannot edit
multi-line comments without much effort.
For C a "/*" kind of comment could theoretically do without a comment leader, but nearly
all files put a "*" in the next line to indicated the comment continues.

chrisbra:
you can use either use Optionset autocommand or use `after/` directory to override values that you do not like personally.

vim-ml:
On 2022-12-20, madjxatw wrote:
> This behavior is really annoying! Whether to insert a comment leader should be
> decided by user but not assumed by the editor. Most of the time when placing a
> short trailing comment, I don't want a comment continuation. At least o should
> be removed.

From my vimrc:

    autocmd FileType * setlocal formatoptions-=o
                        " Override any filetype plugin's attempt
                        " to set the 'fo' 'o' option, which inserts
                        " a comment leader when opening a new line.

HTH,
Gary

madjxatw:
> you can either use Optionset autocommand or use `after/` directory to override values that you do not like personally.

`OptionSet formatoptions` doesn't work for me while `FileType` works. I used to place a `after/ftplugin/<filetype>.vim` to tweak `formatoptions` for each filetype I was using, but I finally gave up when I found that I had to place a considerable number of them, and I often have to work on different machines where vim is not configured same way. From my point of view, the auto insertion should only apply when the comment being typed in Insert mode exceeds the text width of current line. Some people perhaps need it on some specific filetypes,  but why not leave it to users to decide.

PR #11702, opened 2022-12-13
Fix grammar mistakes in README.md and README.txt

merrittlj:
Minor grammar update, not very exciting. Added some commas and changed some words, see commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ All commands are given with normal keyboard characters, so those who can type
 with ten fingers can work very fast.  Additionally, function keys can be
 mapped to commands by the user, and the mouse can be used.
 
-Vim runs under MS-Windows (7, 8, 10, 11), macOS, Haiku, VMS and almost all
+Vim runs under MS-Windows (7, 8, 10, 11), macOS, Haiku, VMS, and almost all
 flavours of UNIX.  Porting to other systems should not be very difficult.
 Older versions of Vim run on MS-DOS, MS-Windows 95/98/Me/NT/2000/XP/Vista,
 Amiga DOS, Atari MiNT, BeOS, RISC OS and OS/2.  These are no longer maintained.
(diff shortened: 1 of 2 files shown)


dynamicdotvoid:
First pull request, just trying to help

README.md
```diff
@@ -28,7 +28,7 @@ All commands are given with normal keyboard characters, so those who can type
 with ten fingers can work very fast.  Additionally, function keys can be
 mapped to commands by the user, and the mouse can be used.
 
-Vim runs under MS-Windows (7, 8, 10, 11), macOS, Haiku, VMS and almost all
+Vim runs under MS-Windows (7, 8, 10, 11), macOS, Haiku, VMS, and almost all
```
DominiquePelle-TomTom, line 31: The comma before "and" in a list is optional. So either is fine.

zewpo: that comma before the and, it actually looks wrong according to the rules I was taught at school.  
Edit: Huh!  Apparently it's optional but acceptable in some cases in a list like that, as long as there are more than two items.  Called the  Oxford comma or the serial comma.  It feels odd to me anyway.

DominiquePelle-TomTom: Quoting https://justpublishingadvice.com/comma-before-and-in-a-list/

> Do you have to use a comma before and in a list?
> No. If your sentence makes sense and is clear, there is no rule that says you must put a comma.

Quoting https://www.grammarly.com/blog/comma-before-and/

> [...] you have to use a comma after each quality you list except the one that comes immediately before and. That comma is optional.

IMO, it looks nicer without the comma before "and".

zewpo: And, it's always wrong if there are only two elements in the list.

brammool:
My approach is that when two forms are OK, take the shortest one.  Especially when it means having fewer punctuation characters.  That leaves only a few changes, which I'll include now.

PR #11703, opened 2022-12-14
Names of CompleteDone test functions do not match body

zeertzjq:


zeertzjq:
Actually it seems that there is currently no test for 'completefunc' returning a Dictionary without `user_data` key. Does a test for that need to be re-added?

zeertzjq:
Maybe I'll change the function bodies instead.

brammool:
> Actually it seems that there is currently no test for 'completefunc'
> returning a Dictionary without `user_data` key. Does a test for that
> need to be re-added?

It appears you already changed the PR to not have user data in the test
that says that in the function name.  Should be OK, the test should
just check things still work when no user data was passed.

-- 
"My particular problem is with registry entries, which seem to just
accumulate like plastic coffee cups..."           -- Paul Moore

PR #11704, opened 2022-12-14
syntax/fstab.vim: Add syntax highlighting for other strings

zdohnal:
Added 'rpc_pipefs' type keyword and 'managed' and 'pamconsole' general options.

This originates from old Fedora patch.

Would you mind merging it?


rid9:
@chrisbra, @zdohnal, got it, thanks. Added to https://github.com/rid9/vim-fstab/releases/tag/1.6.2

zdohnal:
The fix was committed during the last update of runtime files, thank you for adding it!

PR #11707, opened 2022-12-15
CONPTY_STABLE_BUILD Win 10 from build 19045

### LINKED ISSUE #11706: Bad formatting in Windows Terminal after exiting vim and opening new pane
### Steps to reproduce

1. open Windows Terminal session (default PowerShell is fine)
2. run and exit vim
3. open vertical pane with Ctrl-Alt-=

Pre-existing content in the left-most pane will display as if, apparently, the new terminal width remains that of the old (wide) terminal (see attached).  Navigating back to the left-most pane with Alt-Left and hitting Enter shows that new content seems to display correctly (see attached).

Opening a new vertical pane *without* running Vim at Step 2 leaves the left page correctly displayed.
Running and exiting Nano instead at Step 2 leaves the left page correctly displayed.

This may seem like a minor issue, but I am seeing a plethora of display-garbage problems on multi-pane Windows Terminal when using Vim which makes the combination nearly unusable, and since this *could* be the root of all of them and is easily reproducible (I hope), it might be worth giving some priority.

Windows Terminal after Exiting Vim 9 0 1054 and Creating New Pane, 1 of 2

Windows Terminal after Exiting Vim 9 0 1054 and Creating New Pane, 2 of 2

### Expected behaviour

Pre-existing content in left-most pane should display correctly with the line length matching the pane width.

Windows Terminal after Exiting Nano and Creating New Pane

### Version of Vim

9.0.1054

### Environment

Windows 10.0.19045
Windows Terminal 1.15.2874.0
Windows PowerShell 5.1.19041.2364

### Logs and stack traces

_No response_

zewpo:
MS ConPTY has evolved a lot since 2020, it might be stable now.
The goal of this, if this works, it Closes #11706

### DIFF
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -236,7 +236,7 @@ func s:CompleteDone_CompleteFuncDict( findstart, base )
 	      \ 'menu': 'extra text',
 	      \ 'info': 'words are cool',
 	      \ 'kind': 'W',
-	      \ 'user_data': 'test'
+	      \ 'user_data': ['one', 'two']
 	    \ }
 	  \ ]
 	\ }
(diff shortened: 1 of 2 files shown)


zewpo:
Ahh, its breaking in test_terminal scripts.
I can look at this tomorrow.  gotta run now.

PR #11708, opened 2022-12-15
Add showcmd() builtin function

luukvbaal:
Problem: 	function returning the 'showcmd' string can be more useful than
		statusline item.
Solution: 	Add builtin showcmd() function.

>  I wonder if adding a builtin `showcmd()` function, analogous to e.g.
 `mode()` and `searchcount()` will be more useful/idiomatic than adding
 a statusline item?
> > It looks like it is hardly any code.  We can just do both.

Added it here.


luukvbaal:
Also the quickref description for 'showcmd' seems incorrect:
```diff
-'showcmd'	  'sc'	    show (partial) command in status line
+'showcmd'	  'sc'	    show (partial) command in <command/last> line
```

brammool:
> Forgot the quicref description in the original patch for 'showcmdloc'.
> Latest commit adds those, unless you had already included these for a
> future runtime update? If so, could you share the short description? I
> would like to know it for the Neovim port.

I had just added the entry (since the script I use that checks runtime
files complaineda bout it):

'showcmd'	  'sc'	    show (partial) command somewhere
'showcmdloc'	  'sloc'    where to show (partial) command

I'm not quite happy with "somewhere"...  I tried some other things but
they don't fit in one line.

-- 
I have to exercise early in the morning before my brain
figures out what I'm doing.

luukvbaal:
> I'm not quite happy with "somewhere"... I tried some other things but they don't fit in one line.

What about simply:

'showcmd'	  'sc'	    show the (partial) command
'showcmdloc'	  'sloc'    where to show (partial) command

or "display" rather than "show".

src/testdir/test_tabline.vim
```diff
@@ -178,6 +179,9 @@ func Test_tabline_showcmd()
 
   call feedkeys("\<Esc>1234", "xt")
   call VerifyScreenDump(buf, 'Test_tabline_showcmd_2', {})
+
+  call feedkeys("\<Esc>:set tabline=\<CR>:\<CR>1234", "xt")
+  call VerifyScreenDump(buf, 'Test_tabline_showcmd_3', {})
 endfunc
```
zeertzjq, line 186: It seems that a `StopVimInTerminal(buf)` is missing in the end of this function, and it is also missing in the end of the last test function in `test_statusline.vim`

luukvbaal: I wasn't aware that helper function is required. I have added quite a few screendump tests at this point, it might be missing in all of them then. I don't think the test I copied it from the first time had this helper function.

zeertzjq:
This has conflicts now.

luukvbaal:
Rebased

PR #11709, opened 2022-12-15
Save and Restore shortmess option when loading initial error on listdo

### LINKED ISSUE #11471: When `shortmess` contains `F` and we run `:cdo`, messages from autocommands are not always printed to the message-history
### Steps to reproduce

Run `vim --clean -S test.vim`, with

```vim
vim9script

set shortmess+=F
mkdir('/tmp/vim/', 'p')
writefile(['test'], '/tmp/vim/1')
writefile(['test'], '/tmp/vim/2')
writefile(['test'], '/tmp/vim/3')

augroup test
    autocmd!
    autocmd BufWinEnter * echomsg $'BufWinEnter: {expand('<amatch>')}'
    autocmd BufReadPost * echomsg $'BufReadPost: {expand('<amatch>')}'
augroup END

grep! test /tmp/vim/1 /tmp/vim/2 /tmp/vim/3
cdo! substitute/test/TEST
```

Output:
```
BufWinEnter: /tmp/vim/1
(1 of 3): test
"/tmp/vim/2" 1 line, 5 bytes
BufReadPost: /tmp/vim/2
BufWinEnter: /tmp/vim/2
(2 of 3): test
"/tmp/vim/3" 1 line, 5 bytes
BufReadPost: /tmp/vim/3
BufWinEnter: /tmp/vim/3
(3 of 3): test
```

The current output is inconsistent. Why isn't the `echomsg` from `BufReadPost` for `/tmp/vim/1` written to messages, but `BufWinEnter` is? In my opinion, there are two lines missing in the output:
```
"/tmp/vim/1" 1 line, 5 bytes
BufReadPost: /tmp/vim/1
```

If we remove `set shortmess+=F` and rerun the script, we get exactly that:
```
"/tmp/vim/1" 1L, 5B
BufReadPost: /tmp/vim/1
BufWinEnter: /tmp/vim/1
(1 of 3): test
"/tmp/vim/2" 1 line, 5 bytes
BufReadPost: /tmp/vim/2
BufWinEnter: /tmp/vim/2
(2 of 3): test
"/tmp/vim/3" 1 line, 5 bytes
BufReadPost: /tmp/vim/3
BufWinEnter: /tmp/vim/3
(3 of 3): test
```

Note that if we replace `echomsg` with `popup_notification()` in the script above, in total six autocommands are triggered as one would expect (`BufReadPost` and `BufWinEnter` for all three files that are edited).

### Expected behaviour

When `shortmess` contains the `F` flag, either Vim shouldn't give any file infos at all when running `:cdo`, or it should print them all consistently for every file edited in `:cdo`.

### Version of Vim

9.0.818

### Environment

Linux

### Logs and stack traces

_No response_

chrisbra:
This fixes vim/vim#11471 by making sure, that the shortmess option is also restored not only when going through the quickfix/location list, but also initially when opening the first match in the list.

while at it, also refactor the way to save and restore the shortmess option into two separate function first fix version and instead of allocating and clearing the shortmess option time several times, make use of a static buffer to temporarily store the buffer

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3389,3 +3389,5 @@ EXTERN char e_object_member_not_found_str[]
 EXTERN char e_object_member_is_not_writable_str[]
 	INIT(= N_("E1335: Object member is not writable: %s"));
 #endif
+EXTERN char e_internal_shm_length[]
+	INIT(= N_("EXXXX: Internal error: shortmess too long"));
--- a/src/ex_cmds2.c
+++ b/src/ex_cmds2.c
@@ -460,7 +460,6 @@ ex_listdo(exarg_T *eap)
 #if defined(FEAT_SYN_HL)
     char_u	*save_ei = NULL;
 #endif
-    char_u	*p_shm_save;
 #ifdef FEAT_QUICKFIX
     int		qf_size = 0;
     int		qf_idx;
@@ -541,7 +540,9 @@ ex_listdo(exarg_T *eap)
 		buf = NULL;
 	    else
 	    {
+		save_shm_value();
 		ex_cc(eap);
+		restore_shm_value();
 
 		buf = curbuf;
 		i = eap->line1 - 1;
@@ -566,15 +567,9 @@ ex_listdo(exarg_T *eap)
 		// reloading the file.
 		if (curwin->w_arg_idx != i || !editing_arg_idx(curwin))
 		{
-		    // Clear 'shm' to avoid that the file message overwrites
-		    // any output from the command.
-		    p_shm_save = vim_strsave(p_shm);
-		    set_option_value_give_err((char_u *)"shm",
-							  0L, (char_u *)"", 0);
+		    save_shm_value();
 		    do_argfile(eap, i);
-		    set_option_value_give_err((char_u *)"shm",
-							    0L, p_shm_save, 0);
-		    vim_free(p_shm_save);
+		    restore_shm_value();
 		}
 		if (curwin->w_arg_idx != i)
 		    break;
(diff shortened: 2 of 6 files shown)


brammool:
Is it ever possible that this happens recursively?  Perhaps not now, but in the future.
Better add a counter that is used in save_shm_value() and only save the value when it's zero.
And decrement the counter in restore_shm_value() and restore only when it goes to zero.

```mch_memmove(shm_buf, p_shm, STRLEN(p_shm));```
This doesn't copy the NUL at the end of the string.  If the option gets shorter old characters will be included.
Better use STRCPY().

shm_buf[] is a fixed size and the option value is copied to it without checking the length.  That should be OK, since duplicate flags are removed.  But I still wonder if the option string could get longer somehow.

The test should not use /tmp/vim, it may already exist.  Why not use a directory in "src/testdir"?

The test has lines commented out, thus it doesn't look ready to included.

**Comment by bfrg** [comment]
I know I did it too in my example, but wouldn't it be better to avoid calling `:grep` and set the quickfix list directly with `setqflist()` in the test script? Something like:

```vim
call mkdir('/tmp/testdir', 'p')
call writefile(['test'], '/tmp/testdir/1', 'D')
call writefile(['test'], '/tmp/testdir/2', 'D')
call writefile(['test'], '/tmp/testdir/3', 'D')

call setqflist([
    {'filename': '/tmp/testdir/1', 'lnum': 1, 'col': 1, 'text': 'test', 'vcol': 0},
    {'filename': '/tmp/testdir/2', 'lnum': 1, 'col': 1, 'text': 'test', 'vcol': 0},
    {'filename': '/tmp/testdir/3', 'lnum': 1, 'col': 1, 'text': 'test', 'vcol': 0}
])

cdo! substitute/test/TEST
```

chrisbra:
Let me answer all the feedback in one message :)

> Is it ever possible that this happens recursively? Perhaps not now, but in the future.
> Better add a counter that is used in save_shm_value() and only save the value when it's zero.
> And decrement the counter in restore_shm_value() and restore only when it goes to zero.

Yeah, I wondered whether I should add a recursive guard there. So added it now.

> mch_memmove(shm_buf, p_shm, STRLEN(p_shm));
> This doesn't copy the NUL at the end of the string. If the option gets shorter old characters will be included.
> Better use STRCPY().

That should have been fine, because the use of `vim_memset` should have taken care of that. Nevertheless, I changed it to use `STRCPY` but left the `vim_memset` call there. Feel free to remove it though :)

> shm_buf[] is a fixed size and the option value is copied to it without checking the length. That should be OK, since duplicate flags are removed. But I still wonder if the option string could get longer somehow.

Okay, I have added a safe guard, that should error out if this ever happens now.

> 
> The test should not use /tmp/vim, it may already exist. Why not use a directory in "src/testdir"?

Sure, done.

> The test has lines commented out, thus it doesn't look ready to included.

:) I left it there to help debugging if ever needed.

> I know I did it too in my example, but wouldn't it be better to avoid calling :grep and set the quickfix list directly with setqflist() in the test script?

Thanks, good suggestion. I changed the test to make use of `setqflist()`

PR #11711, opened 2022-12-16
fix(json): allow comments in supported files

kylo252:
Problem:    Some json configuration files should allow comments.
Solution:   Update filetype.vim rules to treat them as 'jsonc'.

Refs:
- https://github.com/github/linguist/blob/c1c34e5260797b4d598f5ec76f19723bfc5a1894/lib/linguist/languages.yml#L3068-L3080
- https://github.com/microsoft/vscode/blob/9e8d78012d76469ca55c2eda2d83d882b30a5479/extensions/typescript-basics/package.json#L43-L55
- https://github.com/microsoft/vscode/blob/main/extensions/json/package.json#L46-L68


brammool:
Please also update src/testdir/test_filetype.vim

brammool:
Maybe "[jt]sconfig.*.json" should be "[jt]sconfig*.json" ?  that is: drop the dot before the star?

kylo252:
> Maybe "[jt]sconfig._.json" should be "[jt]sconfig_.json" ? that is: drop the dot before the star?

@brammool, I was treating `[jt]sconfig.*.json` as a regex, but I guess `aupat` doesn't quite work like that?

image

The idea is to support these naming conventions as well:
- https://github.com/vercel/next.js/blob/canary/tsconfig-tsec.json
- https://github.com/babel/babel/blob/main/tsconfig.base.json

k-takata:
See `:help {aupat}`. It is not a regex.

PR #11713, opened 2022-12-17
apply screen restore fix to win10_22H2_or_later

### LINKED ISSUE #11706: Bad formatting in Windows Terminal after exiting vim and opening new pane

zewpo:
Potentially Closes #11706


zewpo:
This currently fails Test_popup_and_window_resize tests.
Moving to draft for now, will try and fix it later.

zewpo:
Hi @brammool 

Just backporting the fix I did for screen restore on Windows 11, to now also work on WIndows 10 22H2.   This is a simple patch, easy to revert if it causes problems.

It does actually succeed Test_popup_and_window_resize on all Windows systems I have tried.  But this test fails in CI testing.

I can replicate the failure if I use WinPTY instead of ConPTY.   I can then resolve the issue simply by setting termwintype=conpty to bypass WinPTY even if if WinPTY is installed.

But that doesn't fix it in CI testing - it only changes the error message to a timeout error.   I wish I could work out why using the alternate screen buffer combined with WinPTY causes this particular test to fail in CI test environment, or at least why the setting termwintype=conpty doesnt solve there but it does in my environments.  So, I'm not sure how to move forward with that at the moment, and I don't know if this really is an issue, or just a quirk of the CI environment.  I don't know how to replicate that exactly.

I believe it's worth releasing anyway, because it solves real open issues.

zewpo:
@brammool   
Problem solved by not using alternate screen buffer if vim is a guest running in a terminal buffer that is hosted by a parent vim instance.

PR #11727, opened 2022-12-19
Fix foldmethod=indent uses wrong 'tabstop' after :diffoff!

zeertzjq:


brammool:
PR included with Patch 9.0.1078

PR #11730, opened 2022-12-20
Fix a mistake in the help for sign_placelist()

zeertzjq:


brammool:
Using {expr} was wrong.  I find using {buffer} a bit confusing, might as well just let out {expr}.

PR #11731, opened 2022-12-20
Fix memory leak with :sign argument repeated

zeertzjq:


brammool:
Silently dropping a duplicate argument doesn't look good.
It makes it hard to spot mistakes, especially since the names are so similar.
It would be better to give an error for a duplicate argument.

zeertzjq:
But will this break backward-compatibility, in case one constructs a command using string concatenation?

brammool:
> But will this break backward-compatibility, in case one constructs a
> command using string concatenation?

Isn't that exactly what the error should point out?

I don't think anyone would have a duplicate argument intentionally.
Not giving an error because there might be an old mistake would mean we
could never fix problems like this.  I also think "backwards compatible"
isn't the right name for this.  "bug-for-bug compatible"?

-- 
ARTHUR:  Then who is your lord?
WOMAN:   We don't have a lord.
ARTHUR:  What?
DENNIS:  I told you.  We're an anarcho-syndicalist commune.  We take it in
         turns to act as a sort of executive officer for the week.

PR #11732, opened 2022-12-20
add postfix proto main.cf.proto

KodeToad:
Postfix updates include .proto files


brammool:
Please also update the test: src/testdir/test_filetype.vim

PR #11741, opened 2022-12-23
src/hashtab.c: hash_may_resize was far too convoluted, refactor

manuelschiller:
This is in effect a partial rewrite of the hashtable code, with these changes:

- The code of hash_may_resize was far too convoluted (and as a result, contained code which defeated the automatic size adaptation as an optimisation, causing the size to flip-flop and rehashing the table, with the result of much wasted CPU). I have also removed the calls to hash_may_resize at the start of hash_add_item, since it's kind of dangerous. Depending on who changes hash_may_resize, it may invalidate iterators, which is a bug waiting to happen. The new code keeps the load factor below 2/3, shrinking the table when the load factor drops below 1/5. It also rehashes when more than 1/5 elements in the table that have been removed.
- The refactoring uses tail calls where possible to save on code size.
- be bit more consistent with unsigned quantities, and their sizes (int/long/long_u)
- hash_(un)lock* can be represented by a flag (recursive locking does not happen with the current code, and if it did, the code does not do what you want anyway because it will call hash_may_resize for every call to hash_unlock); this reduces the amount of memory needed for a hash table by a tiny bit
- change order of members of hashitem_T to reflect access pattern
- computed update of ht_filled in hash_add_item
- do minimum number of comparisons in hash_do_lookup

I have tried hard to remain true to the original spirit of the code, and refactored the code to be much simpler. Much of the code has been written with reuse in mind. With this patch (and my vim config), vim starts up around 4% faster, and the hash table code becomes around 13% smaller. More specifically, according to valgrind's cachegrind tool, I get these speedups/slowdowns in the hash table code (clang 15, -O2 -ftree-vectorize -march=native, on a Intel i7-2640M CPU):

- hash_find: 1.2% slower per call (or around 2 CPU cycles), consistent with the additional tail call to hash_do_lookup
- hash_add: 89% faster per call (because of better hash_may_resize)
- hash_remove: 2.7% faster (because of better hash_may_resize)

The rest of the hash table routines stay more or less the same in terms of speed. I realise that these numbers will be different of different machines, but, overall, this translates to a big net improvement.

### DIFF
--- a/src/hashtab.c
+++ b/src/hashtab.c
@@ -32,14 +32,21 @@
 #if 0
 # define HT_DEBUG	// extra checks for table consistency  and statistics
 
-static long hash_count_lookup = 0;	// count number of hashtab lookups
-static long hash_count_perturb = 0;	// count number of "misses"
+static long_u hash_count_lookup = 0;	// count number of hashtab lookups
+static long_u hash_count_perturb = 0;	// count number of "misses"
 #endif
 
 // Magic value for algorithm that walks through the array.
 #define PERTURB_SHIFT 5
 
-static int hash_may_resize(hashtab_T *ht, int minitems);
+static hashitem_T *hash_do_lookup(hashitem_T *tbl, char_u *key, hash_T hash,
+                                  long_u mask);
+static hashitem_T *hash_do_rehash(hashitem_T *oldtbl, long_u oldused,
+                                  hashitem_T *newtbl, long_u newmask);
+static hashitem_T *hash_rehash_small(hashtab_T *ht);
+static hashitem_T *hash_rehash_big(hashtab_T *ht, long_u newsize);
+static long_u hash_calc_sz(long_u cap);
+static int hash_may_resize(hashtab_T *ht, long_u minitems);
 
 #if 0 // currently not used
 /*
@@ -104,11 +111,8 @@ hash_clear(hashtab_T *ht)
     void
 hash_clear_all(hashtab_T *ht, int off)
 {
-    long	todo;
-    hashitem_T	*hi;
-
-    todo = (long)ht->ht_used;
-    for (hi = ht->ht_array; todo > 0; ++hi)
+    long_u	todo = ht->ht_used;
+    for (hashitem_T *hi = ht->ht_array; todo > 0; ++hi)
     {
 	if (!HASHITEM_EMPTY(hi))
 	{
@@ -134,35 +138,30 @@ hash_find(hashtab_T *ht, char_u *key)
     return hash_lookup(ht, key, hash_hash(key));
 }
 
+
 /*
- * Like hash_find(), but caller computes "hash".
+ * Hash lookup function, for internal use
  */
-    hashitem_T *
-hash_lookup(hashtab_T *ht, char_u *key, hash_T hash)
+    static hashitem_T *
+hash_do_lookup(hashitem_T* tbl, char_u *key, hash_T hash, long_u mask)
 {
-    hash_T	perturb;
-    hashitem_T	*freeitem;
-    hashitem_T	*hi;
-    unsigned	idx;
-
-#ifdef HT_DEBUG
-    ++hash_count_lookup;
-#endif
-
     /*
      * Quickly handle the most common situations:
      * - return if there is no item at all
      * - skip over a removed item
      * - return if the item matches
      */
-    idx = (unsigned)(hash & ht->ht_mask);
-    hi = &ht->ht_array[idx];
+    long_u idx = (hash & mask);
+    hashitem_T *hi = &tbl[idx], *freeitem;
 
-    if (hi->hi_key == NULL)
+#ifdef HT_DEBUG
+    ++hash_count_lookup;
+#endif
+    if (NULL == hi->hi_key)
 	return hi;
-    if (hi->hi_key == HI_KEY_REMOVED)
+    else if (HI_KEY_REMOVED == hi->hi_key)
 	freeitem = hi;
-    else if (hi->hi_hash == hash && STRCMP(hi->hi_key, key) == 0)
+    else if (hi->hi_hash == hash && 0 == STRCMP(hi->hi_key, key))
 	return hi;
     else
 	freeitem = NULL;
@@ -176,22 +175,32 @@ hash_lookup(hashtab_T *ht, char_u *key, hash_T hash)
      * Return the first available slot found (can be a slot of a removed
      * item).
      */
-    for (perturb = hash; ; perturb >>= PERTURB_SHIFT)
+    for (hash_T perturb = hash; ; perturb >>= PERTURB_SHIFT)
     {
 #ifdef HT_DEBUG
 	++hash_count_perturb;	    // count a "miss" for hashtab lookup
 #endif
-	idx = (unsigned)((idx << 2U) + idx + perturb + 1U);
-	hi = &ht->ht_array[idx & ht->ht_mask];
-	if (hi->hi_key == NULL)
-	    return freeitem == NULL ? hi : freeitem;
-	if (hi->hi_hash == hash
-		&& hi->hi_key != HI_KEY_REMOVED
-		&& STRCMP(hi->hi_key, key) == 0)
+	idx = ((idx << 2U) + idx + perturb + 1U);
+	hi = &tbl[idx & mask];
+	if (NULL == hi->hi_key) {
+	    return NULL == freeitem ? hi : freeitem;
+        } else if (HI_KEY_REMOVED == hi->hi_key) {
+            if (NULL == freeitem)
+                freeitem = hi;
+        } else if (hi->hi_hash == hash
+		&& 0 == STRCMP(hi->hi_key, key)) {
 	    return hi;
-	if (hi->hi_key == HI_KEY_REMOVED && freeitem == NULL)
-	    freeitem = hi;
+        }
     }
+}
+
+/*
+ * Like hash_find(), but caller computes "hash".
+ */
+    hashitem_T *
+hash_lookup(hashtab_T *ht, char_u *key, hash_T hash)
+{
+    return hash_do_lookup(ht->ht_array, key, hash, ht->ht_mask);
 }
 
 #if defined(FEAT_EVAL) || defined(FEAT_SYN_HL) || defined(PROTO)
@@ -205,10 +214,10 @@ hash_debug_results(void)
 {
 # ifdef HT_DEBUG
     fprintf(stderr, "\r\n\r\n\r\n\r\n");
-    fprintf(stderr, "Number of hashtable lookups: %ld\r\n", hash_count_lookup);
-    fprintf(stderr, "Number of perturb loops: %ld\r\n", hash_count_perturb);
-    fprintf(stderr, "Percentage of perturb loops: %ld%%\r\n",
-				hash_count_perturb * 100 / hash_count_lookup);
+    fprintf(stderr, "Number of hashtable lookups: %lu\r\n", hash_count_lookup);
+    fprintf(stderr, "Number of perturb loops: %lu\r\n", hash_count_perturb);
+    fprintf(stderr, "Percentage of perturb loops: %lu%%\r\n",
+				hash_count_perturb * 100lu / hash_count_lookup);
 # endif
 }
 #endif
@@ -221,18 +230,15 @@ hash_debug_results(void)
     int
 hash_add(hashtab_T *ht, char_u *key, char *command)
 {
-    hash_T	hash = hash_hash(key);
-    hashitem_T	*hi;
-
-    if (check_hashtab_frozen(ht, command))
-	return FAIL;
-    hi = hash_lookup(ht, key, hash);
-    if (!HASHITEM_EMPTY(hi))
-    {
-	internal_error("hash_add()");
-	return FAIL;
+    if (!check_hashtab_frozen(ht, command)) {
+        hash_T hash = hash_hash(key);
+        hashitem_T *hi = hash_lookup(ht, key, hash);
+        if (HASHITEM_EMPTY(hi)) {
+            return hash_add_item(ht, hi, key, hash);
+        }
+        internal_error("hash_add()");
     }
-    return hash_add_item(ht, hi, key, hash);
+    return FAIL;
 }
 
 /*
@@ -248,14 +254,9 @@ hash_add_item(
     char_u	*key,
     hash_T	hash)
 {
-    // If resizing failed before and it fails again we can't add an item.
-    if ((ht->ht_flags & HTFLAGS_ERROR) && hash_may_resize(ht, 0) == FAIL)
-	return FAIL;
-
     ++ht->ht_used;
     ++ht->ht_changed;
-    if (hi->hi_key == NULL)
-	++ht->ht_filled;
+    ht->ht_filled += (NULL == hi->hi_key);
     hi->hi_key = key;
     hi->hi_hash = hash;
 
@@ -306,7 +307,7 @@ hash_remove(hashtab_T *ht, hashitem_T *hi, char *command)
     void
 hash_lock(hashtab_T *ht)
 {
-    ++ht->ht_locked;
+    ht->ht_flags |= HTFLAGS_LOCKED;
 }
 
 #if defined(FEAT_PROP_POPUP) || defined(PROTO)
@@ -319,7 +320,7 @@ hash_lock(hashtab_T *ht)
 hash_lock_size(hashtab_T *ht, int size)
 {
     (void)hash_may_resize(ht, size);
-    ++ht->ht_locked;
+    hash_lock(ht);
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


brammool:
> This is in effect a partial rewrite of the hashtable code, with these changes:

The hashtable code has been tuned and carefully crafted.  Refactoring
like this seems dangerous and the results are unpredictable.  I would
rather have it split up in small changes that each can be carefully
checked.

> - The code of hash_may_resize was far too convoluted (and as a result,
>   contained code which defeated the automatic size adaptation as an
>   optimisation, causing the size to flip-flop and rehashing the table,
>   with the result of much wasted CPU).

I thought this would not happen.  Can you give a sequence when this does
happen?

>   I have also removed the calls to
>   hash_may_resize at the start of hash_add_item, since it's kind of
>   dangerous. Depending on who changes hash_may_resize, it may invalidate
>   iterators, which is a bug waiting to happen.

The one call to hash_may_resize() at the start of hash_add_item() may
indeed make the "hi" pointer invalid.  This only happens after an
earlier error, in which case retrying might not make much sense anyway.

The hash_may_resize() at the end must stay, I don't see how this is
dangerous.  Leaving it out will be a problem for multiple "add" calls in
a row, the table will run out of space.  Unless it was pre-allocated, in
which case nothing will happen.

>   The new code keeps the load factor below 2/3, shrinking the table
>   when the load factor drops below 1/5. It also rehashes when more
>   than 1/5 elements in the table that have been removed.

Where do these numbers come from?  I would think they need to stay
exactly as they are, unless there is a very clear reason to change them.

> - The refactoring uses tail calls where possible to save on code size.
> - be bit more consistent with unsigned quantities, and their sizes
>   (int/long/long_u)
> - hash_(un)lock* can be represented by a flag (recursive locking does
>   not happen with the current code, and if it did, the code does not do
>   what you want anyway because it will call hash_may_resize for every
>   call to hash_unlock); this reduces the amount of memory needed for a
>   hash table by a tiny bit

This is wrong, this needs to be a counter.  Maybe it just happens to
never count above one right now, but that is incidental.
hash_may_resize() has a check for the table being locked, the code
defenitely does what it needs to do.

> - change order of members of hashitem_T to reflect access pattern

I don't know what "access pattern" means here.  I would not make changes
just because it looks nicer, you never know what effect a change has.

> - computed update of ht_filled in hash_add_item

That actually looks less efficient.  Probably the optimizer takes care
of it anyway.

> - do minimum number of comparisons in hash_do_lookup

I don't see this.  I do see some pointless changes...

Besides this, some generic remarks:
- Please stick to the Vim coding style.  The opening curly brace goes on
  a separate line, for example.  Avoid declaring two variables in one
  line, unless they are very similar.
- Instead of nesting, prefer early returns.

> I have tried hard to remain true to the original spirit of the code, and
> refactored the code to be much simpler. Much of the code has been
> written with reuse in mind. With this patch (and my vim config), vim
> starts up around 4% faster, and the hash table code becomes around 13%
> smaller. More specifically, according to valgrind's cachegrind tool,
> I get these speedups/slowdowns in the hash table code (clang 15, -O2
> -ftree-vectorize -march=native, on a Intel i7-2640M CPU):
> - hash_find: 1.2% slower per call (or around 2 CPU cycles), consistent with the additional tail call to hash_do_lookup
> - hash_add: 89% faster per call (because of better hash_may_resize)
> - hash_remove: 2.7% faster (because of better hash_may_resize)

You can't judge the hashtable code performance with just a couple of
examples.  It very, very much depends on the usage pattern.  Especially
the mix between the number of changes (usually adding items) and
lookups.  That can vary over a very big range.  You can optimize for
efficient adding, making lookups slower, and the other way around.

> The rest of the hash table routines stay more or less the same in terms
> of speed. I realise that these numbers will be different of different
> machines, but, overall, this translates to a big net improvement.

I would call them insignicant.  What matters much more is that specific
usage patterns don't get much slower.  Have you tried adding a million
items to a dictionary?  Or creating a small dictionary and doing a
million lookups?  Without checks like this I would not change anything
in the tradeoffs that are implemented in the code.

-- 
The early bird gets the worm. The second mouse gets the cheese.

manuelschiller:
> This is in effect a partial rewrite of the hashtable code, with these changes:
> The hashtable code has been tuned and carefully crafted. Refactoring like this seems dangerous and the results are unpredictable. I would rather have it split up in small changes that each can be carefully checked.

Can do - let's put this on the shelf until I'm behind a decent internet connection (likely early next year). If it's any consolation: I've had this in my main editor for about two month now if that puts your mind at ease, but I'm happy to postpone until about a week after new year, and then revisit, if that's okay.

> - The code of hash_may_resize was far too convoluted (and as a result, contained code which defeated the automatic size adaptation as an optimisation, causing the size to flip-flop and rehashing the table, with the result of much wasted CPU).
> I thought this would not happen. Can you give a sequence when this does happen?

Well, this patch has developed out of an experiment with something far more radical (replace all of the implementation with a FNV1a-based one). The complete rewrite was quite a bit faster, and in places I did not expect. So I looked to make the changes a bit more adiabatically (i.e. do not change the hash function quite yet). It turns out that growing by the same factor independent of table size in lines 387 to 389 of hashtab.c gives a large fraction of the speed gain.

> I have also removed the calls to hash_may_resize at the start of hash_add_item, since it's kind of dangerous. Depending on who changes hash_may_resize, it may invalidate iterators, which is a bug waiting to happen.
> The one call to hash_may_resize() at the start of hash_add_item() may indeed make the "hi" pointer invalid. This only happens after an earlier error, in which case retrying might not make much sense anyway. The hash_may_resize() at the end must stay, I don't see how this is dangerous. Leaving it out will be a problem for multiple "add" calls in a row, the table will run out of space. Unless it was pre-allocated, in which case nothing will happen.

Agreed - the one at the end must stay, and is not dangerous. I'm only talking about the one at the start of hash_add_item. (Sorry if I was unclear.)

> The new code keeps the load factor below 2/3, shrinking the table when the load factor drops below 1/5. It also rehashes when more than 1/5 elements in the table that have been removed.
> Where do these numbers come from? I would think they need to stay exactly as they are, unless there is a very clear reason to change them.

Looking (or trying to look through) the code of hash_may_resize, I thought the aim was to keep the load factor below 2/3, and shrink if appropriate. When you fall below 1/3 in load factor, and shrink the table by a factor of 2, and then start adding elements again, you end up rehashing. I figured putting the low water mark a little bit lower, say at a load factor of 1/5, should make the behaviour a bit more stable. Rehashing when there's more than 1/5 of removed elements is done to keep the number of perturb loops low.

> - The refactoring uses tail calls where possible to save on code size. - be bit more consistent with unsigned quantities, and their sizes (int/long/long_u) - hash_(un)lock* can be represented by a flag (recursive locking does not happen with the current code, and if it did, the code does not do what you want anyway because it will call hash_may_resize for every call to hash_unlock); this reduces the amount of memory needed for a hash table by a tiny bit
> This is wrong, this needs to be a counter. Maybe it just happens to never count above one right now, but that is incidental. hash_may_resize() has a check for the table being locked, the code defenitely does what it needs to do.

Okay, if you say so - I'm happy to revert this bit. (I had put an assert there to see if it triggers, but it did not.) I will refactor accordingly over the next couple of days.

> - change order of members of hashitem_T to reflect access pattern
> I don't know what "access pattern" means here. I would not make changes just because it looks nicer, you never know what effect a change has.

Well, we look at hi_key long before we look at hi_hash for the lookups/perturbs. That's why it should come first in the structure, so that when we have to fill a cache-line from slow RAM, the data item that's needed first becomes available first.

> - computed update of ht_filled in hash_add_item
> That actually looks less efficient. Probably the optimizer takes care of it anyway.

Depends a bit on what the target CPU likes doing, I'm happy to admit. I'm trying to avoid a branch here. I'm happy to re-check.

> - do minimum number of comparisons in hash_do_lookup
> I don't see this. I do see some pointless changes... Besides this, some generic remarks: - Please stick to the Vim coding style. The opening curly brace goes on a separate line, for example. Avoid declaring two variables in one line, unless they are very similar. - Instead of nesting, prefer early returns.

On my machine, it was a bit faster, but I'm happy to revert/split up. :)

> I have tried hard to remain true to the original spirit of the code, and refactored the code to be much simpler. Much of the code has been written with reuse in mind. With this patch (and my vim config), vim starts up around 4% faster, and the hash table code becomes around 13% smaller. More specifically, according to valgrind's cachegrind tool, I get these speedups/slowdowns in the hash table code (clang 15, -O2 -ftree-vectorize -march=native, on a Intel i7-2640M CPU): - hash_find: 1.2% slower per call (or around 2 CPU cycles), consistent with the additional tail call to hash_do_lookup - hash_add: 89% faster per call (because of better hash_may_resize) - hash_remove: 2.7% faster (because of better hash_may_resize)
> You can't judge the hashtable code performance with just a couple of examples. It very, very much depends on the usage pattern. Especially the mix between the number of changes (usually adding items) and lookups. That can vary over a very big range. You can optimize for efficient adding, making lookups slower, and the other way around.

True. Do you happen to have a preferred set of tests I can use or code up? (I guess we should have this, and I'd be happy to do some work in that direction - even if that's the only thing that comes out of this MR.)

> The rest of the hash table routines stay more or less the same in terms of speed. I realise that these numbers will be different of different machines, but, overall, this translates to a big net improvement.
> I would call them insignicant. What matters much more is that specific usage patterns don't get much slower. Have you tried adding a million items to a dictionary? Or creating a small dictionary and doing a million lookups? Without checks like this I would not change anything in the tradeoffs that are implemented in the code.

I'm happy to do the benchmarks you suggest - give me a couple of days after the new year. Would you prefer some microbenchmark in C, or something in vim script?

Thanks for the feedback, and sorry for the noise. Would you prefer to close this MR, and open a new one (when I've done some more work), or evolve this MR along the lines you suggest?

> […](#)
> -- The early bird gets the worm. The second mouse gets the cheese. /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net \\\ /// \\\ \\\ sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ /// \\\ help me help AIDS victims -- http://ICCF-Holland.org ///

brammool:
Besides the improvements, I'm especially interested in first fixing any
real problems.

> > - The code of hash_may_resize was far too convoluted (and as a
> > result, contained code which defeated the automatic size adaptation
> > as an optimisation, causing the size to flip-flop and rehashing the
> > table, with the result of much wasted CPU).
> > I thought this would not happen. Can you give a sequence when this
> > does happen?
> 
> Well, this patch has developed out of an experiment with something far
> more radical (replace all of the implementation with a FNV1a-based
> one). The complete rewrite was quite a bit faster, and in places I did
> not expect. So I looked to make the changes a bit more adiabatically
> (i.e. do not change the hash function quite yet). It turns out that
> growing by the same factor independent of table size in lines 387 to
> 389 of hashtab.c gives a large fraction of the speed gain.

You don't give any reason why "the size to flip-flop" would happen.
Of course the size increase can be tuned, but if the size would ever get
smaller while adding items that would be a bug.  So did you see this
happen or not?


-- 
LAUNCELOT: I am, sir. I am a Knight of King Arthur.
FATHER:    'Mm ... very nice castle, Camelot ... very good pig country....

manuelschiller:
Hi Bram,

On Sat, Dec 24, 2022 at 03:10:53AM -0800, Bram Moolenaar wrote:
>
>Besides the improvements, I'm especially interested in first fixing any
>real problems.

Understood, and I apologise for the spotty reachability. I'm at my 
parent's place, and their DSL is down to modem speeds if it works at 
all. That's why I suggested shelving this until the new year.

>> > - The code of hash_may_resize was far too convoluted (and as a
>> > result, contained code which defeated the automatic size adaptation
>> > as an optimisation, causing the size to flip-flop and rehashing the
>> > table, with the result of much wasted CPU).
>> > I thought this would not happen. Can you give a sequence when this
>> > does happen?
>>
>> Well, this patch has developed out of an experiment with something far
>> more radical (replace all of the implementation with a FNV1a-based
>> one). The complete rewrite was quite a bit faster, and in places I did
>> not expect. So I looked to make the changes a bit more adiabatically
>> (i.e. do not change the hash function quite yet). It turns out that
>> growing by the same factor independent of table size in lines 387 to
>> 389 of hashtab.c gives a large fraction of the speed gain.
>
>You don't give any reason why "the size to flip-flop" would happen.
>Of course the size increase can be tuned, but if the size would ever get
>smaller while adding items that would be a bug.  So did you see this
>happen or not?

Well, what happened was that I went ahead and replaced the entire 
hashtab.c with a different implementation (quite disruptive) for my own 
enjoyment. It ended up saving around 4% of CPU cycles on a normal "vim 
-c quit" in a valgrind/callgrind/kcachegrind run. (If you haven't used 
it, it can simulate an entire CPU in software and count the clock 
cycles, (simulated) cache misses, branch mispredictions and so on. It's 
not a perfect copy of existing CPU's microarchitecture, but you do get 
stable measurements between runs, and if it's quite a bit faster on that 
simulated CPU, in my experience, it's also faster on real machines, 
although the precise amount of speedup may vary a bit from CPU to CPU.)

I noticed that my version was quite a bit faster - around 40 Mcycles out 
of around 900 Mcycles saved. This was a surprise for me because usually 
my code is not that brilliant, especially when compared to some well 
tuned piece of code that has been around for a long time. So I tried to 
understand (and backport a bit) where this comes from. If I take the 
lines from hash_maybe_resize in hashtab.c that read

if (ht->ht_used > 1000)
     minsize = ht->ht_used * 2;  // it's big, don't make too much room
else
     minsize = ht->ht_used * 4;  // make plenty of room

and replace that with just

minsize = ht->ht_used * 2;

I get a big speed improvement in hash_may_resize. I did not dig deeper, 
but reasoned I might be hitting this sequence of events (for small hash 
tables):

- minsize gets quadrupled while an element is added
- another element gets added, and the code notices that the load factor 
   is far too small, triggering a rehash of the table to shrink it
- another element is added, the size is small, so the table size gets 
   quadrupled

And so on, until the table gets big enough. I did not put in print 
statements to actually check that that is what is happening, but it 
seemed a plausible enough sequence of events.

To me that fact that this went unnoticed meant that hash_may_resize is 
likely too long a function, because it's hard to reason about while 
reading code (unless you're familiar with the code already). Thus, I 
tried to make it smaller, with short functions that don't do much and 
easily fit on an old-style terminal, or at most two. This is where this 
MR came from...

Cheers, and have a Happy Easter ;),

Manuel

>-- 
>LAUNCELOT: I am, sir. I am a Knight of King Arthur.
>FATHER:    'Mm ... very nice castle, Camelot ... very good pig country....
>                 "Monty Python and the Holy Grail" PYTHON (MONTY) PICTURES LTD
>
> /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net   \\\
>///                                                                      \\\
>\\\        sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ ///
> \\\            help me help AIDS victims -- http://ICCF-Holland.org    ///
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/vim/vim/pull/11741#issuecomment-1364511612
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>

-- 
Homepage: http://www.hinterbergen.de/mala
OpenPGP: 0xA330353E (DSA) or 0xD87D188C (RSA)

brammool:
> On Sat, Dec 24, 2022 at 03:10:53AM -0800, Bram Moolenaar wrote:
> >
> >Besides the improvements, I'm especially interested in first fixing any
> >real problems.
> 
> Understood, and I apologise for the spotty reachability. I'm at my 
> parent's place, and their DSL is down to modem speeds if it works at 
> all. That's why I suggested shelving this until the new year.
> 
> >> > - The code of hash_may_resize was far too convoluted (and as a
> >> > result, contained code which defeated the automatic size adaptation
> >> > as an optimisation, causing the size to flip-flop and rehashing the
> >> > table, with the result of much wasted CPU).
> >> > I thought this would not happen. Can you give a sequence when this
> >> > does happen?
> >>
> >> Well, this patch has developed out of an experiment with something far
> >> more radical (replace all of the implementation with a FNV1a-based
> >> one). The complete rewrite was quite a bit faster, and in places I did
> >> not expect. So I looked to make the changes a bit more adiabatically
> >> (i.e. do not change the hash function quite yet). It turns out that
> >> growing by the same factor independent of table size in lines 387 to
> >> 389 of hashtab.c gives a large fraction of the speed gain.
> >
> >You don't give any reason why "the size to flip-flop" would happen.
> >Of course the size increase can be tuned, but if the size would ever get
> >smaller while adding items that would be a bug.  So did you see this
> >happen or not?
> 
> Well, what happened was that I went ahead and replaced the entire 
> hashtab.c with a different implementation (quite disruptive) for my own 
> enjoyment. It ended up saving around 4% of CPU cycles on a normal "vim 
> -c quit" in a valgrind/callgrind/kcachegrind run. (If you haven't used 
> it, it can simulate an entire CPU in software and count the clock 
> cycles, (simulated) cache misses, branch mispredictions and so on. It's 
> not a perfect copy of existing CPU's microarchitecture, but you do get 
> stable measurements between runs, and if it's quite a bit faster on that 
> simulated CPU, in my experience, it's also faster on real machines, 
> although the precise amount of speedup may vary a bit from CPU to CPU.)
> 
> I noticed that my version was quite a bit faster - around 40 Mcycles out 
> of around 900 Mcycles saved. This was a surprise for me because usually 
> my code is not that brilliant, especially when compared to some well 
> tuned piece of code that has been around for a long time. So I tried to 
> understand (and backport a bit) where this comes from. If I take the 
> lines from hash_maybe_resize in hashtab.c that read
> 
> if (ht->ht_used > 1000)
>      minsize = ht->ht_used * 2;  // it's big, don't make too much room
> else
>      minsize = ht->ht_used * 4;  // make plenty of room
> 
> and replace that with just
> 
> minsize = ht->ht_used * 2;
> 
> I get a big speed improvement in hash_may_resize. I did not dig deeper, 
> but reasoned I might be hitting this sequence of events (for small hash 
> tables):
> 
> - minsize gets quadrupled while an element is added
> - another element gets added, and the code notices that the load factor 
>    is far too small, triggering a rehash of the table to shrink it
> - another element is added, the size is small, so the table size gets 
>    quadrupled
> 
> And so on, until the table gets big enough. I did not put in print 
> statements to actually check that that is what is happening, but it 
> seemed a plausible enough sequence of events.

I put in log statements, and found that between having 42 and 50 items
the hashtab resizes to 256 spaces several times.  Only the first time
the size actually changes, the other times it's very busy doing nothing.

> To me that fact that this went unnoticed meant that hash_may_resize is 
> likely too long a function, because it's hard to reason about while 
> reading code (unless you're familiar with the code already). Thus, I 
> tried to make it smaller, with short functions that don't do much and 
> easily fit on an old-style terminal, or at most two. This is where this 
> MR came from...

I'll make a patch to bail out if the size didn't actually change.
Please see what your changes can do on top of that.

-- 
Q: What do you call a fish without an eye?
A: fsh!
Q: What do you call a deer with no eyes?
A: no eye deer.
Q: What do you call a deer with no eyes and no legs?
A: still no eye deer.

manuelschiller:
Hi Bram,

On Sun, Dec 25, 2022 at 12:45:04PM -0800, Bram Moolenaar wrote:
>
>> On Sat, Dec 24, 2022 at 03:10:53AM -0800, Bram Moolenaar wrote:
>> [...]
>
>I put in log statements, and found that between having 42 and 50 items
>the hashtab resizes to 256 spaces several times.  Only the first time
>the size actually changes, the other times it's very busy doing nothing.

Okay, that basically confirms my hunch - it's working much harder than 
was needed under these circumstances.

>> To me that fact that this went unnoticed meant that hash_may_resize is
>> likely too long a function, because it's hard to reason about while
>> reading code (unless you're familiar with the code already). Thus, I
>> tried to make it smaller, with short functions that don't do much and
>> easily fit on an old-style terminal, or at most two. This is where this
>> MR came from...
>
>I'll make a patch to bail out if the size didn't actually change.
>Please see what your changes can do on top of that.

Okay, I'll have a look in the new year, when I'm back home behind a 
decent internet connection (likely after Jan 9th, when I should be back 
in the Geneva region).

Thanks!

Manuel

>-- 
>Q: What do you call a fish without an eye?
>A: fsh!
>Q: What do you call a deer with no eyes?
>A: no eye deer.
>Q: What do you call a deer with no eyes and no legs?
>A: still no eye deer.
>
> /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net   \\\
>///                                                                      \\\
>\\\        sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ ///
> \\\            help me help AIDS victims -- http://ICCF-Holland.org    ///
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/vim/vim/pull/11741#issuecomment-1364736409
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>

-- 
Homepage: http://www.hinterbergen.de/mala
OpenPGP: 0xA330353E (DSA) or 0xD87D188C (RSA)

brammool:
> On Sun, Dec 25, 2022 at 12:45:04PM -0800, Bram Moolenaar wrote:
> >
> >> On Sat, Dec 24, 2022 at 03:10:53AM -0800, Bram Moolenaar wrote:
> >> [...]
> >
> >I put in log statements, and found that between having 42 and 50 items
> >the hashtab resizes to 256 spaces several times.  Only the first time
> >the size actually changes, the other times it's very busy doing nothing.
> 
> Okay, that basically confirms my hunch - it's working much harder than 
> was needed under these circumstances.
> 
> >> To me that fact that this went unnoticed meant that hash_may_resize is
> >> likely too long a function, because it's hard to reason about while
> >> reading code (unless you're familiar with the code already). Thus, I
> >> tried to make it smaller, with short functions that don't do much and
> >> easily fit on an old-style terminal, or at most two. This is where this
> >> MR came from...
> >
> >I'll make a patch to bail out if the size didn't actually change.
> >Please see what your changes can do on top of that.
> 
> Okay, I'll have a look in the new year, when I'm back home behind a 
> decent internet connection (likely after Jan 9th, when I should be
> back in the Geneva region).

It indeed needs a good look at the context, I managed to break it in two
ways, fixed by patches 9.0.1097 and 9.0.1100.  If we make more changes
we'll have to be careful.

-- 
Why isn't there mouse-flavored cat food?

chrisbra:
this seems to have stalled (and partially included) so I am closing it.

PR #11744, opened 2022-12-24
Stop setting 'b:did_indent = 1' in indent/vue.vim

jparise:
This guard variable was being set before indent/html.vim was loaded, preventing it from doing any actual work.

Instead, don't set `b:did_indent = 1` at all with the expectation that indent/html.vim will do so, just like we do for similar file types (such as in indent/xhtml.vim).


brammool:
I'll include it.

PR #11759, opened 2022-12-29
fix lua indentation of non-lowercase "keywords"

beardedsakimonkey:
The current lua indentation interprets some occurrences of "keywords" containing upper-case characters as actual keywords. (e.g. "iF").

To reproduce the issue, create a file called x.lua containing the following lines:
```lua
local t = {
iF = 1,
a = 1,
}
```

Now, if you open vim with `vim --clean --cmd 'set ignorecase' x.lua`, indenting the file (via "=G") results in the following:
```lua
local t = {
        iF = 1,
                a = 1
        }
```

After this commit, the result is the following:
```lua
local t = {
        iF = 1,
        a = 1
}
```

Note that I tried to reach out to the maintainer listed in the file, but the email bounced back.

Thanks

### DIFF
--- a/runtime/indent/lua.vim
+++ b/runtime/indent/lua.vim
@@ -27,6 +27,16 @@ if exists("*GetLuaIndent")
 endif
 
 function! GetLuaIndent()
+    let ignorecase_save = &ignorecase
+  try
+    let &ignorecase = 0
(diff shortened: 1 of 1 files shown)


brammool:
@marcuscf

PR #10823, opened 2022-07-31
W32 GUI: ctrl_minus as ^_

ant0sha:
NOTE: consider it as WIP state - due to unclear status of imap

(an attempt to fix #10817)

In principle the functionality for ^_ is recovered and works, but what i don't understand, is a subtle difference:

both old w32 gvim 8.1. and new gtk3 gvim allow this mapping and it works:

  ```imap <C-_> BINGO```

with this patch in w32 gvim9 it _do not work_, but only:

  ```exec "imap \x1f BINGO"```

Any ideas, why? Can it be that way mappings are working was changed recently?

PR #11760, opened 2022-12-30
Add 'execute' feature to test_mswin_event

### LINKED ISSUE #10817: W32 GUI: Ctrl+- produces ^- and not ^_ any more
### Steps to reproduce

(Reported on vim_use by  Salman Halim <salmanhalim@gmail.com>)

8.2 w32 gvim used to produce ^_ when Ctrl+- was press.
9.0 w32 gvim produces ^- instead.

E.g. mapping:

```imap <C-_> TEST```

stopped to work and need to be worked around like this:

```imap <C--> TEST```

### Add info/related

GTK3 GUI produces ^_ so perhaps to be consistent it is better to fix it?

### Expected behaviour

^_ is produced

### Version of Vim

9.0.100

### Environment

OS: Windows

### Logs and stack traces

_No response_

zewpo:
Problem:  test_mswin_event can cause mch_inchar to get into an endless loop waiting for typeahead buffer to be filled.
Solution:  implement an 'execute' feature for test_mswin_event, similar to feedkeys mode X

Also fixed a few modifier key combination encoding issues.

This PR also adds more lowlevel event testing, and incorporates PR #10823 which potentially Closes #10817


zewpo:
Hi @brammool 
I think this PR is ready for review now.

PR #11762, opened 2022-12-30
keep artifacts

chrisbra:
Keep Artifacts to give users a way to test builds


chrisbra:
So what this means is, whenever a PR is successfull on appveyor, you should find the compiled binaries at the Appveyor Artifacts tab:

E.g. for this run: 

https://ci.appveyor.com/project/chrisbra/vim/builds/45803435/artifacts

PR #11764, opened 2022-12-31
Fix smoothscroll to work with zz/zt/zb

ychin:
This commit does two things:

---

First, existing zt/zb/zz implementation is buggy when 'smoothscroll' are 'wrap' are set. They can lead to inconsistency bugs with regarding to repainting and cursor states.

For zz, if the user uses `zz` whenever the top line is smooth-scrolled, the resulting draw will be wrong. For zt, if the line above the current line is super long (meaning it is so long that the wrapped line covers more than the entire screen), it will lead to the cursor being in a bad state. For zb, the exact conditions are harder to reproduce, but basically it triggers when the top line is smooth scrolled, and second line is more than one physical line tall.

Most of the reasons why they are broken is because `reset_skipcol()` or `redraw_later()` aren't called in the right place, so this change will make sure to call them appropriately.

---

Second, zb/zz currently will not utilize 'smoothscroll' in a satisfactory fashion. If 'smoothscroll'/'wrap' are set, we should be able to accurate put the current line at the bottom/middle of the screen by smooth-scrolling a long wrapped line at the top of the screen. This change will make those two commands do that.

---

Also, add tests to regression test these features.


ychin:
For the first point (inconsistency bugs), see the following asciicast which demos how to trigger them.

[asciicast](https://asciinema.org/a/uX4mTEfqs6uG12lzxNNPkgDSA)

ychin:
Also, I think `scrolloff` is currently not working perfectly when `smoothscroll` is set. The logic is not taking smoothscroll into account and it also doesn't handle super long lines (lines that cannot be shown in its entirely on the screen). I decided not to tackle them for now as I don't think they lead to any serious bugs, just that it would scroll a little more than necessary.

Ideally, when we have `smoothscroll`, `scrolloff` can be more precise in always just showing that many physical lines for zt and zb (actually, for zb, we don't even need smoothscroll since we can show partial lines depending on the `display` setting).

brammool:
> This commit does two things:
> 
> ---
> 
> First, existing zt/zb/zz implementation is buggy when 'smoothscroll' are
> 'wrap' are set. They can lead to inconsistency bugs with regarding to
> repainting and cursor states.
> 
> For zz, if the user uses `zz` whenever the top line is smooth-scrolled,
> the resulting draw will be wrong. For zt, if the line above the current
> line is super long (meaning it is so long that the wrapped line covers
> more than the entire screen), it will lead to the cursor being in a bad
> state. For zb, the exact conditions are harder to reproduce, but
> basically it triggers when the top line is smooth scrolled, and second
> line is more than one physical line tall.
> 
> Most of the reasons why they are broken is because `reset_skipcol()` or
> `redraw_later()` aren't called in the right place, so this change will
> make sure to call them appropriately.
> 
> ---
> 
> Second, zb/zz currently will not utilize 'smoothscroll' in a
> satisfactory fashion. If 'smoothscroll'/'wrap' are set, we should be
> able to accurate put the current line at the bottom/middle of the screen
> by smooth-scrolling a long wrapped line at the top of the screen. This
> change will make those two commands do that.
> 
> ---
> 
> Also, add tests to regression test these features.

Thanks.  The 'smoothscroll' implementation is tricky, there is quite a
bit of interference with other options and features.  We'll just have to
find and fix the problems over time.

-- 
Violators can be fined, arrested or jailed for making ugly faces at a dog.
		[real standing law in Oklahoma, United States of America]

PR #11765, opened 2022-12-31
Add more filetype detection tests for .tex files

smjonas:
This adds test cases that check for several things in a .tex file:
- a `format` expression in the first line of the document
- a context-specific command (`\usemodule`)
- a LaTeX-specific command (`\newcommand`
- when `tex_flavor` is set or unset

### DIFF
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -1654,16 +1654,44 @@ endfunc
 func Test_tex_file()
   filetype on
 
-  " only tests one case, should do more
-  let lines =<< trim END
-      % This is a sentence.
+  call writefile(['%& pdflatex'], 'Xfile.tex')
+  split Xfile.tex
+  call assert_equal('tex', &filetype)
+  bwipe
 
-      This is a sentence.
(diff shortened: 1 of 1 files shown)


smjonas:
I will add another test case that uses the global `tex_flavor` variable

zeertzjq:
This doesn't cover the case when `g:tex_flavor` is set to `'plain'` instead of being unset.

src/testdir/test_filetype.vim
```diff
@@ -1654,16 +1654,44 @@ endfunc
...
+  call assert_equal('tex', &filetype)
+  bwipe
+
+  " tex_flavor is unset
+  call writefile(['%& plain'], 'Xfile.tex')
+  split Xfile.tex
+  call assert_equal('plaintex', &filetype)
+  bwipe
+
+  let g:tex_flavor = 'plaintex'
```
```suggestion
  let g:tex_flavor = 'plain'
```

smjonas:
> This doesn't cover the case when `g:tex_flavor` is set to `'plain'` instead of being unset.

Updated

PR #11768, opened 2023-01-01
MS-Win special key modifiers

zewpo:
The new test_mswin_event method has revealed that various combinations of modifier keys and special keys (Fn keys, PageUp, Down etc...) arent mappable in the windows console, but they are mappable in the gui as well as xterm and other linux terminals.  

Not sure if its possible to make MS-Win consoles work more like the others, but this PR is a WIP towards that goal.


zewpo:
going to close and reopen, to bump appveyor

zewpo:
Github taking a long time to decide if pull request is mergeable.
appveyor gives up at about 10 seconds:  AppVeyor was unable to build non-mergeable pull request.

zewpo:
Closing and reopening again just to bump appveyor.
this is still very much WIP

zewpo:
see if 3rd time is charmed - closing and reopening one more time.

brammool:
Strange AppVeyor error: "AppVeyor was unable to build non-mergeable pull reque"

zewpo:
@brammool   yeah, its annoying, I tried a few different suggestions to work around it.    

By the way, this PR is still very much WIP.  One of the suggestions I came across was perhaps appveyor would not merge PR with WIP prefix in the name.   But that didn't seem to make a difference!

Anyway, so I've just created myself an appveyor account and working out the problem over there, hopefully creating less noise in the Vim mailing list for the moment. ;p

chrisbra:
if you rebase and squash everything into a single commit and then force-push to your remote, this should trigger a new build on appveyor. I typically use interactive rebase, this also allows to delete unwanted commits.

Also instead of opening and closing this PR, you can either create an empty commit  and then push again. But I would rather use the interactive-rebase + force-push approach :)

zewpo:
Thanks @chrisbra   

Thanks for the info, yeah git has so many features to learn about :).

I've also noticed things happen much faster at quieter times of the day.  Like, just now, git took less than a couple seconds to recognise that it is mergeable.   I read somewhere that appveyor simply gives up after 10 seconds, waiting for github to respond if it is mergeable or not.  Don't know if that is really true, but it seems to match what I saw here as well.

zewpo:
HI @brammool 

Finally, I got the whole set of tests working for the combinations of modifier keys with special keys - working in the MS-Win console.  ie, the same tests that already exist and work for the GUI.

It was tricky, and a bit of a grind, to debug why so many tests failed in CI, because they did not fail on my PC.   I was assuming that the CI environment was able to use VTP - but that seems to be what the hurdle was - as things started working when I relaxed the VTP requirement.

This PR is functional now, and it makes things work better, if you want to merge it now.   

I'm also still keen on doing a clean-up refactoring especially rationalising some of the conditional clauses that I implemented to make it work, I'm sure there is overlap and redundancy going on in there at the moment.  I can do them as separate PRs.

brammool:
> HI @brammool 
> 
> Finally, I got the whole set of tests working for the combinations of
> modifier keys with special keys - working in the MS-Win console.  ie,
> the same tests that already exist and work for the GUI.
> 
> It was tricky, and a bit of a grind, to debug why so many tests failed
> in CI, because they did not fail on my PC.   I was assuming that the
> CI environment was able to use VTP - but that seems to be what the
> hurdle was - as things started working when I relaxed the VTP
> requirement.
> 
> This PR is functional now, and it makes things work better, if you
> want to merge it now.   

Thanks for doing this work.  I can't comment on the details, I'll just
assume you did the right thing.

One thing I prefer is that when checking flags it's better to explicitly
check for zero or non-zero:

 	    else if (nModifs & ALT)

In C this happens to be "true" when some bit is set, but it's better and
avoids mistakes to check that explicitly:

	    else if ((nModifs & ALT) != 0)

I'll change the ones I spot.

> I'm also still keen on doing a clean-up refactoring especially
> rationalising some of the conditional clauses that I implemented to
> make it work, I'm sure there is overlap and redundancy going on in
> there at the moment.  I can do them as separate PRs.

OK, but please be careful not to change any logic.  We've had problems
before that a refactoring actually introduced bugs.  Some of the code is
more tricky than one would expect.


-- 
SOLDIER: What?  Ridden on a horse?
ARTHUR:  Yes!
SOLDIER: You're using coconuts!

zewpo:
thanks @brammool 

OK. preference noted - truthiness vs explicit !=0 and ==0.   yeah, I get that readability is better to avoid mistakes, I've also noted that not all languages implement the C non-zero for truthiness.

Oh, with what I was thinking about the refactor, I was only thinking about the block of code I added to the middle of decode_key_event.   I will make a set of test cases first to be sure.

PR #11777, opened 2023-01-03
Reset vcol_off after drawing a screen line of a wrapping line

lyokha:
This fixes broken `cursorline` and `colorcolumn` behavior when wrapping lines contain concealed text.

Below the problem is shown visually.

Peek-vcol_off_bad

Both `cursorline` and `colorcolumn` are broken when drawn at secondary (wrapped) screen lines because their positions include the values of `vcol_off` calculated from the primary (wrapped) lines.

The solution is to reset `vcol_off` after drawing each screen line.

Below is how `cursorline` and `colorcolumn` are drawn after the patch has been applied.

Peek vcol_off_good

In the middle of the screencast I typed

```vim
:set number list
```

to make sure that they were not affected, and then advanced the first line to make a double wrap.


brammool:
> This fixes broken `cursorline` and `colorcolumn` behavior when
> wrapping lines contain concealed text.
> 
> Below the problem is shown visually.
> 
> Peek-vcol_off_bad
> 
> Both `cursorline` and `colorcolumn` are broken when drawn at secondary
> (wrapped) screen lines because their positions include the values of
> `vcol_off` calculated from the primary (wrapped) lines.
> 
> The solution is to reset `vcol_off` after drawing each screen line.
> 
> Below is how `cursorline` and `colorcolumn` are drawn after the patch
> has been applied.
> 
> Peek vcol_off_good
> 
> In the middle of the screencast I typed
> 
> ```vim
> :set number list
> ```
> 
> to make sure that they were not affected, and then advanced the first
> line to make a double wrap.

Could you add a regression test?  Or at least have a script that
reproduces the prolem.  The animated gifs do not show what is causing
the problem.  A screendump test probably works best for this.

-- 
Mushrooms always grow in damp places and so they look like umbrellas.

lyokha:
I added tests in [src/testdir/test_conceal.vim](src/testdir/test_conceal.vim) as the issue is related to `conceal` feature. The test  launches vim with *20x40* size, makes 3 long lines that must wrap, setup `wrap`, `linebreak`, `cursorcolumn` and `colorcolumn`, and then put the cursor at the end of a wrapping line. The test successfully runs on the patched vim and fails in the master branch.

lyokha:
I see, that there are 2 issues with this PR:

1. I mistakenly replaced word `cursorcolumn` by `cursorline` in its description.
2. In the test, the `colorcolumn` should rather be placed after column 40 (now its 25 as soon as I played with *10x20* grid before the final *20x40*) because it's a better check for this issue.

So,

1. Not sure, if it's possible to replace the desription,
2. but I can change the test to make a bigger `colorcolumn`, say, 50.

PR #11783, opened 2023-01-05
IPv6 test fixes

jamessan:
- Use getaddrinfo() to find localhost address/family for tests
- Use try/finally in Test_nb_quit_with_conn to avoid hit-enter prompt on a failure


brammool:
BSD doesn't seem to like this change.

jamessan:
Yeah, I'll look into that tonight, hopefully.

PR #11788, opened 2023-01-07
fix default link of javaScriptNumber

meqativ:
Makes so that the `javaScriptNumber` highlight group - links to the proper main group (instead of `javaScriptValue` which isn't defined anywhere, use `Number`)


brammool:
@fleiner

PR #11793, opened 2023-01-09
Upgrade Vim's appveyor testing to VS 2022

### LINKED ISSUE #294: Strange characters in console - only from this installer builds
### Steps to reproduce

1. Run the installer, full options
2. open up cmd.exe,
3. start; vim --clean -U NONE
4. open help;  h:
5. use mouse visual select some text, and drag the visual selection around a bit.
6. see weird characters;
image

### Expected behaviour

Expect no weird characters.

I couldn't get it to occur when I compile the same patch levels myself.  I only see the problem with these appveyor builds.

I copied the appveyor.bat file, and tried to reproduce the build as close as possible to what appveyor is doing.  But, I did have to make a couple of changes;

Python version to 3.11.1
set PYTHON3_VER=311
set PYTHON3_RELEASE=3.11.1

Ruby version from 3.0 to 3.2
set RUBY_VER=32
set RUBY_API_VER_LONG=3.2.0
set RUBY_BRANCH=ruby_3_2
set RUBY_RELEASE=3.2.0-1

and some manual steps to get ruby to compile and install - the configure step was failing due to unfound header files.

The only other difference I could think of, is that appveyor.yml file mentions Visual Studio 2015, but I'm using Visual Studio 2022.

I'll install increasingly older versions until I find which one it starts at.  Will update what I find.

### Version of Vim

gvim_9.0.1097_x64.exe, gvim_9.0.1117_x64.exe,  and gvim_9.0.1122_x64.exe

### Environment

Windows 11 21H2

zewpo:
Problem:  Vim's appveyor testing is on VS 2015.
Solution:  Upgrade Vim's appveyor testing to VS 2022

Attempts to Close https://github.com/vim/vim-win32-installer/issues/294

Note: This PR depends on PR #11798 being merged first - otherwise some test cases will start failing.


zewpo (commented):
I've reviewed, and tested.   I think it's good to go now - functionally anyway.

I'd think that @k-takata  and/or @chrisbra  might want to review.

brammool:
What is the relation between the appveyor changes and the rest?
If they are unrelated, it's better to split them off into a separate PR.
Also, going from Python 3.5 to 3.11 seems unrelated.  Or is there a relation with the Visual Studio version?

zewpo:
Thanks for the question @brammool 

Most of the changes in this PR were to enable the testing to pass on appveyor runner image VS 2022.  I did try to reduce the change to only that which is needed to get it to pass.   The only unnecessary change is that there is now additional testing proving more coverage.

OK, to explain in detail...

Yes, that's right Python 3.11 is pre-installed in the VS 2022 build runner image - but 3.5 is not.    I saw that 3.11 is already used in github actions CI testing.   I guess we could keep 3.5 by adding a download and install step for it for the VS 2022 image - if that is preferred?  An intermediate option could be 3.8 - that is the oldest version that is pre-installed on VS 2022 image.  I can understand the value in testing on a range of Python versions - old as well as new.  Oh, and 2.7 is still included also, but do let me know what the preference is here, and I can make it happen.

Another change I made to the build, was to drop the Windows XP target.  Actually, this is the origin saga of this PRs.

It took me a long time to work out why the builds from the MS-Win32-Installer project were causing garbled characters to appear in new terminals on Windows 10, Windows 11, and server versions 2019 and 2022.   The garbled characters appear in both the new windows terminal and command prompt consoles.     But my own local builds were working fine.   I was eventually able to reproduce the problem by building a virtual image with VS 2015 - following the same bat script as the appveyor VS 2015 config in the MS-Win32-Installer project uses.    I found that the XP target was causing it.
  
Before I found that, I tried a number of different ways to get rid of the problem.    I rolled back all changes to Vim source code, to prior to where the problem started occurring in appveyor - BUT this did not solve that problem.   The only thing I found to resolve the problem was to drop the Windows XP target.   I don't know exactly why targeting windows XP started to cause the problem in the first place, I'm not sure how to figure out why it started happening - but I confirmed that keeping the XP target, even in old source code, still caused the problem - so I don't believe it is caused by anything that changed in the Vim source code.   And dropping the XP target resolves the problem. 

Targeting XP limits us to the older VS 2015 image - because we cant target XP in newer versions of VS.    After dropping windows XP target, - to resolve that real issue - we now have the opportunity upgrade to appveyor build image to VS 2022.  

@k-takata  and @chrisbra  both voiced their preference for upgrading to VS 2022.  And, so that is where this PR comes in.

When I first tried to upgrade the appveyor build runner image to VS 2022,  I found that some tests were failing for the GUI.  In particular, the test for the low-level keyboard events was failing - the one using the combinations of Ctrl and Shift along with  Movement characters.     

Another strange thing is (still) that I can't reproduce this test failure anywhere else - except for appveyor's build runner images - all of them fail from VS 2017, 2019 and 2022 - the only appveyor image that was passing was the VS 2015 image.   

It gets stranger, when I downloaded the artifacts that were built and failed the testing in those images, and run the same exe file with those same tests on my own PC environment - the testing passes - the same exe file that failed in appveyor!    I tried to replicate this issue on my own VM images of Windows 7, 8.1, 10, 1, 2019 and 2022.   I could not reproduce the error anywhere else.   Also, the github actions CI testing was passing fine with VS 2022.  So - whatever is going on with that failure - is related to the appveyor test runtime environment - not the  build itself.

Trying to understand what was going on - I updated the test_mswin_event.vim to use nested for loops, to make it test all possible combinations of all modifier keys against each of the PC keyboard movement keys (PgUp, PgDn, Home, End, Up, Down, Left and Right.)       

This revealed the pattern of the failure that only occurs in the appveyor images test environment - and that is - they drop the Shift modifier that is sent with the movement keys - without changing the character.    Everywhere else, the character is changed if the Shift modifier is dropped.   So, I added a section in the test to allow this weirdness - along with an explanation in the comments.     

I recognise this is a bit of a temporary solution, rather than working out exactly the root cause as to why the problem occurs in those appveyor images.   But, given that as of yet, I can't replicate the problem - as it doesn't occur anywhere else - is low risk - and given that this is blocking the upgrade to vs 2022, I thought it would be pragmatic to provide this work-around for the moment.

Anyway, along with this more thorough testing, I found a couple of missing edge cases in the console keyboard handling that I had recently introduced - and that is why there are a few changes in os_win32.c.    In particular, various combinations with the Alt keys with Shift & Ctrl was not being handled - but now it is - and passing the tests.

Part of adding the extra testing, trying to work out the failure pattern, was that I also enabled the Fn + modifier key testing to work in the GUI.   Now, instead of the Fn keys combo tests skipping the GUI altogether, most of the Fn key combos can be tested with the GUI as well.   The changes I made for Fn key testing, is only additions to the test file, providing more test coverage - I did not need to change any C source code.  This did not reveal any failures as such - so technically, this additional testing is not needed in this PR - but it was added along the path of trying to work out the problems associated with this PR - and I thought it was useful to keep the extra test coverage anyway.

zewpo:
Long story short.  OK, yes, I can separate it this way...    I will go ahead and create a PR with just the testing character fixes - and that PR will not depend on this PR.    But, this PR will depend on that PR happening first.

zewpo:
OK @brammool 
This PR is now split into two; this one #11793, and new one; #11798
Both ready for review.

brammool:
> Thanks for the question @brammool 
> 
> Most of the changes in this PR were to enable the testing to pass on
> appveyor runner image VS 2022.  I did try to reduce the change to only
> that which is needed to get it to pass.   The only unnecessary change
> is that there is now additional testing proving more coverage.

OK, thanks for the explanations.  The thing with a big patch is that
when something fails it's often hard to decide whether to roll back the
whole change, or try to fix the problem.  Best is to try and split it
into sensible chuncks.  Then rolling back is often easier and less
dramatic.

> OK, to explain in detail...
> 
> Yes, that's right Python 3.11 is pre-installed in the VS 2022 build
> runner image - but 3.5 is not.    I saw that 3.11 is already used in
> github actions CI testing.   I guess we could keep 3.5 by adding a
> download and install step for it for the VS 2022 image - if that is
> preferred?  An intermediate option could be 3.8 - that is the oldest
> version that is pre-installed on VS 2022 image.  I can understand the
> value in testing on a range of Python versions - old as well as new.

Python 3.5 seems quite old.  Every Python version change may have some
small problems (the tests are very picky about wording of error
messages).  Jumping ahead to a recent version should be OK, if it works.

> Oh, and 2.7 is still included also, but do let me know what the
> preference is here, and I can make it happen.

It will be good to keep testing with a Python 2 version, for as long as
that works.  Python 2 has been deprecated, but I know there are scripts
that still depend on it (converting to Python 3 is often more difficult
than one would think).  AFAIK Python 2.7 is the last in the Python 2
series (there are some minor releases, I see 2.7.18 mentioned).

> Another change I made to the build, was to drop the Windows XP target.
> Actually, this is the origin saga of this PRs.

I suppose later tools no longer support it properly.  Targeting Windows
XP probably changes the libraries used, with unpredictable side effects.

[...]

> Another strange thing is (still) that I can't reproduce this test
> failure anywhere else - except for appveyor's build runner images -
> all of them fail from VS 2017, 2019 and 2022 - the only appveyor image
> that was passing was the VS 2015 image.   

I'm not sure if it applies here, but I have had problems with multiple
binaries with the same name, which get picked up depending on the
ordering of directories in %PATH%.  In an image you probably only get
the binary that should be used, no matter the order of directories.  On
a local system you may have several alternatives installed at the same
time, and using %PATH% to pick up the right one, which is tricky.
Then play around with various CygWin and MinGW versions and the mess has
been made...

-- 
Managers are like cats in a litter box.  They instinctively shuffle things
around to conceal what they've done.

brammool:
I have noticed that after including this PR the time that AppVeyor reports has increased from around 9 minutes to around 15 minutes.  This can be seen in the history, since patch 9.0.1177 the time has clearly increased.
It is hard to imagine that VS 2022 is so much slower.

zewpo:
Yeah, VS 2022 shouldn't really be much different. I'll investigate.

zewpo:
Comparing the log files (summarized below) - it looks like the VS 2022 runner image is 2x slower, compared with VS 2015.

There doesn't seem to be any individual step in the process that stands out as a particular cause of the problem.  Every step takes about 2x longer than on the 2015 runner image.  

I wondered if it was something to do with the appveyor project settings, I can't see those settings; it's on @chrisbra  account;   https://ci.appveyor.com/project/chrisbra/vim

Oh, but, just checked, I can see the same slowdown is also happening in my appveyor account, where I replicated the process, ie VS 2022 is consistently 2x slower there too.   So, later on I will be able to play around with the settings on my account and see if I can work out the difference.

Comparing log from 1177:  https://ci.appveyor.com/project/chrisbra/vim/builds/45904275
with log from 1169:  https://ci.appveyor.com/project/chrisbra/vim/builds/45892650

```
Step: git clone
================
     | 0 
[00:00:00]	[00:00:00] Build started
...
[00:00:00]	[00:00:00] git clone -q --branch=master https://github.com/vim/vim.git C:\projects\vim
[00:00:31]	[00:01:02] git checkout -qf .....
   31 sec      62 sec   
=>  step diff 31 seconds

Step: 200% slower

Step: Build Vim
================
[00:00:32]	[00:01:06] "Building MSVC 64bit console Version"
[00:01:39] 	[00:03:17] if exist gvimext.dll.manifest ...
   99 sec     197 sec   
=>  step diff 64 seconds (+1:04)

Step: 205% slower


Step: Build gVim
================
[00:01:39]	[00:03:17] "Building MSVC 64bit GUI Version"
[00:02:45]  [00:05:33] Finished generating code
   66 sec     136 sec   
=>  step diff 70 seconds (+1:10)

Step: 194% slower


Step: Test gVim
================
[00:02:46]  [00:05:35] nmake -f Make_mvc.mak VIMPROG=..\gvim
[00:06:24]  [00:11:47]   Failed:     0 Tests
  218 sec     372 sec
=>  step diff 154 seconds (+2:34)

Step: 242% slower

* Another noteworthy difference, occurred 5x:  
  "Could not find platform dependent libraries <exec_prefix>"

Step: Test Vim
================
[00:06:25]  [00:11:48] nmake -f Make_mvc.mak VIMPROG=..\vim
[00:09:28]  [00:16:12]   Failed:     0 Tests
  183 sec     264 sec
=>  step diff 81 seconds  (+1:21)

Step: 326% slower


Total:
================
[00:00:00]  [00:00:00] Build started
[00:09:30]  [00:16:14] Build success
  570 sec     974 sec
=> Total diff 404 seconds  (+6:44)

Total:  241% slower
```

zewpo:
Edit last post: percentages were using wrong denominator... fixed now.

zewpo:
Posting the fixed edit into a new github comment - because github doesn't seem to send edited comments up to vim_dev mailing group;

```
Step: git clone
================
     | 0 
[00:00:00]	[00:00:00] Build started
...
[00:00:00]	[00:00:00] git clone -q --branch=master https://github.com/vim/vim.git C:\projects\vim
[00:00:31]	[00:01:02] git checkout -qf .....
   31 sec      62 sec   
=>  step diff 31 seconds

Step: 200% slower

Step: Build Vim
================
[00:00:32]	[00:01:06] "Building MSVC 64bit console Version"
[00:01:39] 	[00:03:17] if exist gvimext.dll.manifest ...
   99 sec     197 sec   
=>  step diff 64 seconds (+1:04)

Step: 196% slower


Step: Build gVim
================
[00:01:39]	[00:03:17] "Building MSVC 64bit GUI Version"
[00:02:45]  [00:05:33] Finished generating code
   66 sec     136 sec   
=>  step diff 70 seconds (+1:10)

Step: 206% slower


Step: Test gVim
================
[00:02:46]  [00:05:35] nmake -f Make_mvc.mak VIMPROG=..\gvim
[00:06:24]  [00:11:47]   Failed:     0 Tests
  218 sec     372 sec
=>  step diff 154 seconds (+2:34)

Step: 171% slower

* Another noteworthy difference, occurred 5x:  
  "Could not find platform dependent libraries <exec_prefix>"

Step: Test Vim
================
[00:06:25]  [00:11:48] nmake -f Make_mvc.mak VIMPROG=..\vim
[00:09:28]  [00:16:12]   Failed:     0 Tests
  183 sec     264 sec
=>  step diff 81 seconds  (+1:21)

Step: 144% slower


Total:
================
[00:00:00]  [00:00:00] Build started
[00:09:30]  [00:16:14] Build success
  570 sec     974 sec
=> Total diff 404 seconds  (+6:44)

Total:  171% slower
```

zewpo:
update. I tried a matrix of VS Versions.  It just seems to get progressively slower for each new VS version...  Details below.

So, am now trying the same(ish) process on github actions, just to compare how long it takes there.

https://ci.appveyor.com/project/zewpo/vim/builds/45925642

```
Job name  TestsStatus
VS-2015   Succeeded in  8 min 59 sec
VS-2017   Succeeded in 10 min 16 sec
VS-2019   Succeeded in 14 min 27 sec
VS-2022   Succeeded in 15 min 12 sec
```

k-takata:
We used to use AppVeyor to test with VS2015 because we are supporting VS2015 to VS2022 (See src/INSTALLpc.txt).
Using the same version of VS on both GHA and AppVeyor seems redundant.
Perhaps, we can stop using AppVeyor?

zewpo:
update - gha takes about the same time ~15 minutes.

yeah, redundant to have 2x VS 2022 builds for no added benefit.   I also noticed gha already using 2022, but not 2015.   

Perhaps an option is to rollback this PR, to keep the main vim project repo using 2015 testing on appveyor - to still give confidence that people on older VS 2015 can still compile with it.  This will speed things up - important here because vim gets a number of patches every day.   

I think it's still safe to say that vim can still be compiled with VS  2015.   I'd suggest we could just provide a note that there is a known issue of running vim in newer windows consoles 10, 11 server 2022 - if vim was compiled with older VS 2015.    The garbled characters problem doesn't appear on older windows versions (Win 7, 8) consoles - and I think that would be the typical user who still compiles vim with VS 2015 anyway.

Also, since this is only a problem with newer consoles - it won't occur for the users who compile vim for newer consoles if they also use a newer version of VS to compile with, which is more likely I think anyway

I think the vim-win32-installer project repo can still upgrade to 2022, if you want, see https://github.com/vim/vim-win32-installer/pull/296  but it does take about 6 minutes longer to run each day.  Upgrading is the only way I've found to resolve the garbled characters problem that is still being distributed at the moment.    There may be other ways to resolve it, but I struggle to determine the dll hell that causes the issue.    Now, since the vim-win32-installer project is only running once per day, then the extra 6 minutes doesn't seem too problematic, as it does in the main vim repo.    Also, the VS 2022 image still targets Windows 7 (winver 0x0601) - it still works fine with older consoles. I've tested the installer built from the VS 2022 image Win 7 vm image - had no problems.

zewpo:
Let me check one more thing.  The VS 2017 only seems a little bit slower.  It slowness seems to jump more between 2017 and 2019.   Let me see if 2017 resolves the garbled issue.

zewpo:
So, one more lot investigation results.

The problem still occurs on the VS 2017 build, just like the VS 2015 build!  Thats worth noting as a known issue.

The problem is solved on VS 2019 and VS 2022 builds.

brammool:
> update - gha takes about the same time ~15 minutes.

Perhaps the later versions do more static analysis?  I can't imagine the
basic compilation got slower.  Perhaps the optimizer flags can be
tweaked to make it faster.

> yeah, redundant to have 2x VS 2022 builds for no added benefit.   I
> also noticed gha already using 2022, but not 2015.   

The Appveyor results did come in faster and it's easier for me to check
if something basic is broken.  GHA nearly always has failures, thus I
have to check the individual results to see it's a flaky failure.

> Perhaps an option is to rollback this PR, to keep the main vim project
> repo using 2015 testing on appveyor - to still give confidence that
> people on older VS 2015 can still compile with it.  This will speed
> things up - important here because vim gets a number of patches every
> day.   

I think it's fine to keep Appveyor on VS 2015.  If it works.

> I think it's still safe to say that vim can still be compiled with VS
> 2015.   I'd suggest we could just provide a note that there is a known
> issue of running vim in newer windows consoles 10, 11 server 2022 - if
> vim was compiled with older VS 2015.    The garbled characters problem
> doesn't appear on older windows versions (Win 7, 8) consoles - and I
> think that would be the typical user who still compiles vim with VS
> 2015 anyway.

This is unrelated to CI, right?  A "known problem" note in the help
would be appropriate.  And in src/INSTALLpc.txt.


-- 
Never enter the boss's office unless it's absolutely necessary.  Every boss
saves one corner of the desk for useless assignments that are doled out like
Halloween candy to each visitor.

zewpo:
> This is unrelated to CI, right?

Yes, that's right, VS 2015 is fine in the current CI pipeline.   I don't imagine anyone would use its build output artifacts on Windows >= 10 consoles.

So, I think it's prudent to at least roll back this PR.  #11793  aka Patch 9.0.1177.   

Thankfully, you got me to split it up.   Because, the other part that was split out, was moved into Patch 9.0.1169 and that is really what was blocking the upgrade path to VS 2022 for the vim-win32-installer project.  So, Patch 9.0.1169 was done and making all the tests pass now in VS 2022 (and still passing in VS 2015).   Wouldn't want to roll that part back.

The only problem I see now, is not really for this main vim project - its only because VS 2015 is still being used by the vim-win32-installer project - which is downloaded and used by people on Windows >= 10 where the garbled characters appear.  I have created a separate PR in that project for this issue.

zewpo:
@brammool   So, is it easier for you to just revert the patch, or is it easier for you if I create a separate PR that does the reverse of what the patch did?  I guess with a separate PR, I can suggest some edits to the src/INSTALLpc.txt.  And, if I find an appropriate place in the help files, happy to suggest something there too.

zewpo:
yeah, I don't know the root cause yet, and am currently stuck trying to work it out.

have created pr 
https://github.com/vim/vim/pull/11810

sorry - I've ran out of time this week, have to disappear for the weekend with family.   Sorry I couldn't wrap it all up neatly yet.

PR #11810, opened 2023-01-13
Rollback appveyor to VS2015 and warn of known issue with VS2015 & VS2017 for Win 10

zewpo:
reference discussion and rollback for: https://github.com/vim/vim/pull/11793


zewpo:
OK, with this PR, appveyor gets back to 9 minutes.   I gotta run now, wont be able to follow up until next week.

brammool:
I can include it like this, but we should try to use Python 3.11 instead of 3.8.  Does that work with MSVC 2015?

zewpo:
> ... Python 3.11 instead of 3.8.  Does that work with MSVC 2015?

Maybe 3.11 could be made to work in 2015, if we add an installation step to the workflow.   That might slow it down a little bit, maybe by about a minute or two.  I can test it out and get back to you, i think there may be a way to use caching to speed it up.

Py 3.8 is the newest version of py that comes already pre-installed on the 2015 image.  Also, 3.8 was a convenient version because it comes pre-installed in all of appveyors msvc images, 2015, 2017, 2019 and 2022.

PR #11814, opened 2023-01-14
Upgrade to python 3.11 in AppVeyor MSVC 2015 build

zewpo:
Prob:  The appveyor msvc 2015 image comes with older python 3.8 preinstalled.
Sln:  Add python 3.11 installation step

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -36,7 +36,6 @@ matrix:
 
 before_build:
   # - systeminfo
-  # - wmic /NameSpace:\\root\Microsoft\Windows\Storage Path MSFT_PhysicalDisk get FriendlyName,Model,AllocatedSize,UniqueId
   - setlocal ENABLEDELAYEDEXPANSION
   - call ver
   - set "VSWHERE=%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe"
@@ -58,7 +57,7 @@ build_script:
 test_script:
   - cd src/testdir
     # Testing with MSVC gvim
-  - path C:\Python38-x64;%PATH%
+  - path C:\Python11-x64;%PATH%
   - nmake -f Make_mvc.mak VIMPROG=..\gvim
   - nmake -f Make_mvc.mak clean
     # Testing with MSVC console version
--- a/ci/appveyor.bat
+++ b/ci/appveyor.bat
@@ -1,15 +1,14 @@
 @echo off
 :: Batch file for building/testing Vim on AppVeyor
 
+setlocal ENABLEDELAYEDEXPANSION
+cd %APPVEYOR_BUILD_FOLDER%
+
 :: Python3
 set PYTHON3_VER=311
 set PYTHON3_RELEASE=3.11.1
 set PYTHON3_URL=https://www.python.org/ftp/python/%PYTHON3_RELEASE%/python-%PYTHON3_RELEASE%-amd64.exe
 set PYTHON3_DIR=C:\python%PYTHON3_VER%-x64
-
-
-setlocal ENABLEDELAYEDEXPANSION
-cd %APPVEYOR_BUILD_FOLDER%
 
 if not exist downloads mkdir downloads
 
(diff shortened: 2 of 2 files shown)


brammool:
> Prob:  The appveyor msvc 2015 build can be faster.
> Sln:  Use the VIMDLL option to compile both gVim and Vim in a single pass.
> Result:  This saves ~62 seconds off the build process.

Although the faster build is nice, we do want to test building the GUI
and console versions.  Ideally we build all three.  This is to check
that the conditions are OK.  I guess the separate console and GUI
versions are what is used most often.

> Prob:  The appveyor msvc 2015 image comes with older python 3.8 preinstalled.
> Sln:  Add python 3.11 installation step
> Result:  This adds ~6 seconds to the build process.

I don't care too much, but I assume building with the latest Python is
the best for AppVeyor.  Theoretically we could make a change that breaks
older Python versions, but it's unlikely.

-- 
7. You finally do take that vacation, but only after buying a USB modem
   and a laptop.

zewpo:
Thanks for the feedback @brammool 

OK, I've removed the DLL console change out this PR, so it's now focussed only on upgrading AppVeyor msvc 2015 to use python 3.11.

W.R.T. to the DLL console build.   Is it not sufficient that the GHA matrix already systematically test the scenarios with/without DLL, gVm/vim, msvc/mingw etc?   I thought the only reasons for keeping AppVeyor in the mix, was for speed and testing old VS 2015.

brammool:
> Thanks for the feedback @brammool 
> 
> OK, I've removed the DLL console change out this PR, so it's now
> focussed only on upgrading AppVeyor msvc 2015 to use python 3.11.
> 
> W.R.T. to the DLL console build.   Is it not sufficient that the GHA
> matrix already systematically test the scenarios with/without DLL,
> gVm/vim, msvc/mingw etc?   I thought the only reasons for keeping
> AppVeyor in the mix, was for speed and testing old VS 2015.

It's good that all three configurations are tested somewhere.

Either system may be down or failing for some time, it's good to have a
backup.

-- 
13. You refer to going to the bathroom as downloading.

k-takata:
It looks that python is not working: https://ci.appveyor.com/project/chrisbra/vim/builds/45942497#L17265
```
	test_python2.vim: python feature missing
	test_python3.vim: python3 feature missing
	test_pyx2.vim: python feature missing
	test_pyx3.vim: python3 feature missing
```

zewpo:
ouch, I will investigate.

zewpo:
@k-takata  good news (I think) is that it's not new - not caused by patch 9.0.1200.     

Looking back further in the appveyor logs, I can still see the same earlier this month.  

Looking at the .appveryor.bat script file, it doesn't look like it ever built the console with python.  Only the gui has python.   I checked a sample of the git log history of the appveyor.bat file - going back to about 2016 - and it never did python for console.

So, is it something we want to start making it do?

k-takata:
Oh, I pointed the wrong lines.
The GUI version is compiled with +python3/dyn but it is not working:
https://ci.appveyor.com/project/chrisbra/vim/builds/45942497?fullLog=true#L4477
```
From test_python3.vim:
    Skipped
NO tests executed
SKIPPED test_python3.vim: python3 feature missing
```

I don't think we need to enable python3 on the console version.

k-takata:
Perhaps, the PATH setting is missing?

zewpo:
Ahh, sorry, OK, maybe I missed new py3 PATH for the GUI test.  I will look again :)

zewpo:
yeah, I made a silly mistake.  typo in the path.   new PR coming real soon !

zewpo:
Thanks @k-takata  for spotting the issue.   

I rushed a bit after @brammool earlier feedback about the DLL. I quickly split it out but left a typo in the yaml file.   

PR https://github.com/vim/vim/pull/11828 is now up to fix this.  It refactors it by moving the cmd script segments out of the yml file, into the bat file, and re-using variables properly instead of hardcoded path names.   hopefully it will help avoid the mistake from happening again.

brammool:
> @k-takata  good news (I think) is that it's not new - not caused by patch 9.0.1200.     
> 
> Looking back further in the appveyor logs, I can still see the same earlier this month.  
> 
> Looking at the .appveryor.bat script file, it doesn't look like it ever built the console with python.  Only the gui has python.   I checked a sample of the git log history of the appveyor.bat file - going back to about 2016 - and it never did python for console.
> 
> So, is it something we want to start making it do?

I don't think there is a connection between GUI/console and Python.  If
Python is tested in one of the two builds that should be sufficient.

The only relevant #ifdef I see is in if_py_both.h, for the GUI
need_mouse_correct is set.

-- 
24. You realize there is not a sound in the house and you have no idea where
    your children are.

PR #11818, opened 2023-01-14
Allow attaching custom data to setqflist() via 'user_data' field

tom-anders:
For example, LSP plugins could use this to attach the original LSP `Location` item for things like 'textDocument/reference'

### DIFF
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -43,6 +43,7 @@ struct qfline_S
     char_u	qf_cleared;	// set to TRUE if line has been deleted
     char_u	qf_type;	// type of the error (mostly 'E'); 1 for
 				// :helpgrep
+    typval_T	qf_user_data;	// custom user data associated with this item
     char_u	qf_valid;	// valid error message detected
 };
 
@@ -82,6 +83,7 @@ typedef struct qf_list_S
     int		qf_count;	// number of errors (0 means empty list)
     int		qf_index;	// current index in the error list
     int		qf_nonevalid;	// TRUE if not a single valid entry found
+    int		qf_has_user_data; // TRUE if at least one item has user_data attached
     char_u	*qf_title;	// title derived from the command that created
 				// the error list or set by setqflist
     typval_T	*qf_ctx;	// context set by setqflist/setloclist
@@ -168,7 +170,7 @@ static efm_T	*fmt_start = NULL; // cached across qf_parse_line() calls
 static callback_T qftf_cb;
 
 static void	qf_new_list(qf_info_T *qi, char_u *qf_title);
-static int	qf_add_entry(qf_list_T *qfl, char_u *dir, char_u *fname, char_u *module, int bufnum, char_u *mesg, long lnum, long end_lnum, int col, int end_col, int vis_col, char_u *pattern, int nr, int type, int valid);
+static int	qf_add_entry(qf_list_T *qfl, char_u *dir, char_u *fname, char_u *module, int bufnum, char_u *mesg, long lnum, long end_lnum, int col, int end_col, int vis_col, char_u *pattern, int nr, int type, typval_T *user_data, int valid);
 static void	qf_free(qf_list_T *qfl);
 static char_u	*qf_types(int, int);
 static int	qf_get_fnum(qf_list_T *qfl, char_u *, char_u *);
@@ -951,6 +953,7 @@ typedef struct {
     char_u	*pattern;
     int		enr;
     int		type;
+    typval_T	*user_data;
     int		valid;
 } qffields_T;
 
@@ -1720,6 +1723,7 @@ qf_init_process_nextline(
 		fields->pattern,
 		fields->enr,
 		fields->type,
+		fields->user_data,
 		fields->valid);
 }
 
@@ -1965,6 +1969,7 @@ qf_new_list(qf_info_T *qi, char_u *qf_title)
     qf_store_title(qfl, qf_title);
     qfl->qfl_type = qi->qfl_type;
     qfl->qf_id = ++last_qf_id;
+    qfl->qf_has_user_data = FALSE;
 }
 
 /*
@@ -2145,6 +2150,7 @@ qf_add_entry(
     char_u	*pattern,	// search pattern
     int		nr,		// error number
     int		type,		// type character
+    typval_T	*user_data,     // custom user data or NULL
     int		valid)		// valid entry
 {
     qfline_T	*qfp;
@@ -2173,6 +2179,13 @@ qf_add_entry(
     qfp->qf_col = col;
     qfp->qf_end_col = end_col;
     qfp->qf_viscol = vis_col;
+    if (user_data == NULL || user_data->v_type == VAR_UNKNOWN)
+        qfp->qf_user_data.v_type = VAR_UNKNOWN;
+    else
+    {
+	copy_tv(user_data, &qfp->qf_user_data);
+	qfl->qf_has_user_data = TRUE;
+    }
     if (pattern == NULL || *pattern == NUL)
 	qfp->qf_pattern = NULL;
     else if ((qfp->qf_pattern = vim_strsave(pattern)) == NULL)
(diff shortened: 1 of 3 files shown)


src/quickfix.c
```diff
@@ -951,6 +952,7 @@ typedef struct {
     char_u	*pattern;
     int		enr;
     int		type;
+    typval_T	*user_data;
```
yegappan, line 956: The support for garbage collection needs to be added for "user_data".  You can try setting a Dict as "user_data", invoke test_garbagecollect_now() function to do the garbage collection and then retrieve the user_data using getqflist() and try to access the Dict.

tom-anders: Yeah that indeed seems to clear the dict. How do I add it to garbage collection though? `dict_alloc()`...?

brammool:
> @tom-anders commented on this pull request.
> 
> > @@ -951,6 +952,7 @@ typedef struct {
>      char_u	*pattern;
>      int		enr;
>      int		type;
> +    typval_T	*user_data;
> 
> Yeah that indeed seems to clear the dict. How do I add it to garbage
> collection though? `dict_alloc()`...?

The entry point is set_ref_in_quickfix().  You can see it calls
mark_quickfix_ctx() to handle "qf_ctx", something similar has to happen
for all user_data values.

-- 
Me?  A skeptic?  I trust you have proof.

tom-anders:
> @tom-anders commented on this pull request. > @@ -951,6 +952,7 @@ typedef struct { char_u *pattern; int enr; int type; + typval_T *user_data; Yeah that indeed seems to clear the dict. How do I add it to garbage collection though? `dict_alloc()`...?
> The entry point is set_ref_in_quickfix(). You can see it calls mark_quickfix_ctx() to handle "qf_ctx", something similar has to happen for all user_data values.
> […](#)
> -- Me? A skeptic? I trust you have proof. /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net \\\ /// \\\ \\\ sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ /// \\\ help me help AIDS victims -- http://ICCF-Holland.org ///

Thanks for the pointers, think I got it now

tom-anders:
@brammool @yegappan Does this look good or is there anything else you want me to add? Should I address the code coverage issue?

src/quickfix.c
```diff
@@ -7801,6 +7819,23 @@ set_errorlist(
     return retval;
 }
 
+static int mark_quickfix_user_data(qf_info_T *qi, int copyID) 
+{
+    int abort = FALSE;
+    for (int i = 0; i < LISTCOUNT && !abort; ++i) {
+	qfline_T *qfp;
+	int j;
+        FOR_ALL_QFL_ITEMS(&qi->qf_lists[i], qfp, j)
```
yegappan: This is functionally correct but inefficient.  The garbage collection runs often.  If a user has multiple quickfix/location lists with a large number of entries (without any user data), then this loop will slow down Vim.  Is it possible to set a flag in the list if any of the item has a user data attached to it?  This way only lists with items that have a user data attached need to be inspected.

src/testdir/test_quickfix.vim
```diff
@@ -1806,13 +1806,14 @@ func SetXlistTests(cchar, bnum)
...
   call g:Xsetlist([{'bufnr': a:bnum, 'lnum': 1},
-	      \  {'bufnr': a:bnum, 'lnum': 2, 'end_lnum': 3, 'col': 4, 'end_col': 5}])
+        \  {'bufnr': a:bnum, 'lnum': 2, 'end_lnum': 3, 'col': 4, 'end_col': 5, 'user_data': {'6': [7, 8]}}])
   let l = g:Xgetlist()
   call assert_equal(2, len(l))
   call assert_equal(2, l[1].lnum)
   call assert_equal(3, l[1].end_lnum)
   call assert_equal(4, l[1].col)
   call assert_equal(5, l[1].end_col)
+  call assert_equal({'6': [7, 8]}, l[1].user_data)
```
yegappan, line 1817: Can you add one more test for garbage collection?  You can add a complex type as user data, then call the test_garbagecollect_now() function and then retrieve the user data and check to make sure that it is still present.

brammool:
I would also expect a change in qf_free_items(().  clear_tv() should be called for qf_user_data

src/testdir/test_quickfix.vim
tom-anders: Done, sorry for missing that

src/quickfix.c
tom-anders: This is certainly possible, but tbh I am not yet familiar enough with the code to find all the right places to set this flag - Obviously we should update the flag inside `qf_add_entry()`, but I'm not sure in which other places I'd need to initialize/update/copy this flag..?

tom-anders:
> I would also expect a change in qf_free_items((). clear_tv() should be called for qf_user_data

Added `clear_tv` as suggested, thanks!

src/quickfix.c
dundargoc: @yegappan would you say this is a blocker? If so, mind giving a few pointers so this fellow can get this to the finish line?

yegappan: Yes.  This will slow down the garbage collection with a large number of quickfix/location lists/entries. 

To address this, a flag should be added to qf_list_T that indicates whether any of the entries in that list have user data.  By default, this will be set to FALSE.  In the qf_add_entry() function, when copying the user data, set this flag to TRUE.  When running the garbage collection for the quickfix lists, check this flag.  If the flag is set, then do the garbage collection update for the user data in the quickfix items for that list.  If the flag it not set, the list can be skipped.

tom-anders: Thanks, I was still not sure where to add the default value, but `qf_new_list` looked reasonable, hope this is correct?

src/quickfix.c
```diff
@@ -2173,6 +2179,12 @@ qf_add_entry(
     qfp->qf_col = col;
     qfp->qf_end_col = end_col;
     qfp->qf_viscol = vis_col;
+    if (user_data == NULL || user_data->v_type == VAR_UNKNOWN)
+        qfp->qf_user_data.v_type = VAR_UNKNOWN;
+    else {
```
yegappan: Can you move the curly brace to the next line (to follow the coding style)?

src/quickfix.c
```diff
@@ -7801,6 +7824,27 @@ set_errorlist(
     return retval;
 }
 
+static int mark_quickfix_user_data(qf_info_T *qi, int copyID) 
+{
+    int abort = FALSE;
+    for (int i = 0; i < LISTCOUNT && !abort; ++i) {
+	qf_list_T *qfl = &qi->qf_lists[i];
+	if (qfl->qf_has_user_data)
```
yegappan: To reduce the indentation below, can you change the above to the following?
```
     if (!qfl->qf_has_user_data)
        continue;

File: src/quickfix.c [hunk 6]
```diff
@@ -2335,6 +2347,7 @@ copy_loclist_entries(qf_list_T *from_qfl, qf_list_T *to_qfl)
 		    from_qfp->qf_pattern,
```
yegappan, line 2348: The new flag has to be copied to the new location list in copy_loclist().  Also, do you need to copy the per quickfix item user_data in copy_loclist_entries()?

tom-anders: Done :+1:

yegappan: I think you also need to copy the per-item user_data in the copy_loclist_entries() function?

yegappan (approved):

yegappan (commented):
The changes look good to me.

tom-anders:
@brammool Friendly ping - is there anything stopping this from being included?

zeertzjq:
There are some code style test failures caused by trailing white spacez.

brammool:
It's in the todo list.  I want to include bug fixes first.  And please make the tests pass.

tom-anders:
Ahh, sorry about that - I think the last time I took a look at the CI, I was really confused because the line numbers wouldn't match with my local and branch (and also test_codestyle.vim didn't even exist on my branch).

A rebase helped :)

yegappan:

tom-anders:
Sorry, I was on vacation, thanks for merging!

PR #11828, opened 2023-01-16
Typo in path for python3 in AppVeyor testing.

zewpo:
Prb:  Typo in path for python3 in AppVeyor testing.
Sln:  Use the PYTHON3_DIR variable, instead of hardcoding the same thing in two different files.


zewpo:
I've reorganised the code a bit, moved the cmd script into the bat file, so its not split in half between the yml file and bat file.
Hopefully it will help prevent future mistakes like my last typo.

PR #5546, opened 2020-01-28
Supports Windows Terminal

ghost:
In term=win32, VTP code "reduction" is required.
First, do not redraw the right edge of the screen currently.
Next, remove the trick of the background color.
And, to avoid a terminal bug, the screen output buffer is flushed each time.
This is the content of the previous PR. Same as the first commit in this PR.

In the second commit, enter ambiguous width characters.
Without this change, NUL (0x00) would be entered instead.

Always keep track of Windows Terminal changes.
I just make it usable.

This patch has not been fully tested.

PR #11829, opened 2023-01-16
Fixed vim.exe's broken screen return on exit

ghost:
After the new year, I reinstalled Vim and played around with it. It's pretty broken.
The cause is v9.0.0850.

image

Before and after running vi.exe (my alternate executable in vim/vim:master).
Color management has been demolished from its basic structure. I've only modified the appearance for now.
All the fast drawing code has been disabled, making screen drawing unusually slow. I won't fix that. Why did this happen...

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -8461,26 +8461,23 @@ vtp_flag_init(void)
 vtp_init(void)
 {
 # ifdef FEAT_TERMGUICOLORS
-    if (!vtp_working)
-    {
-	CONSOLE_SCREEN_BUFFER_INFOEX csbi;
-	csbi.cbSize = sizeof(csbi);
-	GetConsoleScreenBufferInfoEx(g_hConOut, &csbi);
-	save_console_bg_rgb = (guicolor_T)csbi.ColorTable[g_color_index_bg];
-	save_console_fg_rgb = (guicolor_T)csbi.ColorTable[g_color_index_fg];
-	store_console_bg_rgb = save_console_bg_rgb;
-	store_console_fg_rgb = save_console_fg_rgb;
+    CONSOLE_SCREEN_BUFFER_INFOEX csbi;
+    csbi.cbSize = sizeof(csbi);
+    GetConsoleScreenBufferInfoEx(g_hConOut, &csbi);
+    save_console_bg_rgb = (guicolor_T)csbi.ColorTable[g_color_index_bg];
+    save_console_fg_rgb = (guicolor_T)csbi.ColorTable[g_color_index_fg];
+    store_console_bg_rgb = save_console_bg_rgb;
+    store_console_fg_rgb = save_console_fg_rgb;
 
-	COLORREF bg;
-	bg = (COLORREF)csbi.ColorTable[g_color_index_bg];
-	bg = (GetRValue(bg) << 16) | (GetGValue(bg) << 8) | GetBValue(bg);
-	default_console_color_bg = bg;
+    COLORREF bg;
+    bg = (COLORREF)csbi.ColorTable[g_color_index_bg];
+    bg = (GetRValue(bg) << 16) | (GetGValue(bg) << 8) | GetBValue(bg);
+    default_console_color_bg = bg;
 
-	COLORREF fg;
-	fg = (COLORREF)csbi.ColorTable[g_color_index_fg];
-	fg = (GetRValue(fg) << 16) | (GetGValue(fg) << 8) | GetBValue(fg);
-	default_console_color_fg = fg;
-    }
+    COLORREF fg;
+    fg = (COLORREF)csbi.ColorTable[g_color_index_fg];
+    fg = (GetRValue(fg) << 16) | (GetGValue(fg) << 8) | GetBValue(fg);
+    default_console_color_fg = fg;
 # endif
     use_alternate_screen_buffer = win10_22H2_or_later && p_rs && vtp_working
 						&& !mch_getenv("VIM_TERMINAL");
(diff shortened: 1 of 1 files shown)


ntak:
I don't think it will start on Windows Vista/7/8.
Windows PE reading process, Windows Update ignoring missing dll thunks entered?
There is no trace of the strangely difficult part of avoiding the old kernel32.dll that does not have ConPTY-specific functions.

zewpo:
Hi @ntak   

I made v9.0.0850 to fix some problems.    And it works for those problems.   But, perhaps there is a conflict of assumptions going on here.  I don't want to break the fixes that I made in v9.0.0850, but if that has caused something else to break, then I want to understand it properly.
 
We obviously need to sort it out.  I'm really sorry if I broke something that you previously worked on.   I can walk you through each change I made, and explain it step by step - and we can work out a good solution.

To start with, could you explain the difference are between what you expect to see vs what you are seeing?   I see the image you posted.   The top has white text.  The bottom has green text.    Is the color the issue?   The background at the bottom seems a bit messed up.     How did you produce each image, which one do you consider wrong, and why is it wrong?

ntak:
Why did you change the contents of USE_VTP? You've removed the inline branch to exclude 256-color and 16-color modes with p_tgc by setting variables on the fly. As a result, the 16-color mode died completely.

Why did you change USE_WT to vtp_working? USE_WT and vtp_working have really really no relevance. as a result.
- Screen colors are not restored (see screenshot).
- UTF-8 ambiguous wide characters cannot be entered from the keyboard.
- At least doubled the time it takes to scroll up and down.
- The drawing color of non-highlighted text is always indeterminate both foreground and background.
- The editor screen color saved when starting an external program is not restored.
- Failed to restore the screen color at startup when the process ended.
- All character output buffering is disabled.

There is more. There is more. I'm sick of it.

zewpo:
>  Why did you change USE_WT to vtp_working? USE_WT and vtp_working have really really no relevance. 

I did that because most of the things that were wrong with WT were also wrong with the console.  The problems were solved in both WT and the console by checking if it has vtp - to make it not rely on the incorrect 16 color table that windows was reporting.    

I saw no real reason to distinguish between WT and the command prompt.   Both had the same problems that were fixed by enabling VTP and using the escape sequences for alternate screen buffer.   Fixing the bugs needed to be applied to both the old console and to WT.

I'm sorry if I missed that there may have been other reasons to distinguish between WT and the older console.  I'm really surprised at your list of thing to do with buffering, speed and input of characters.  This should not have made any difference.  I'd like to understand why you think it made a difference there.

>  * Screen colors are not restored (see screenshot).

It should.  It does on my computer - and others have reported it working better than it used to.   If it has VTP, it now uses the alternate screen-buffer, so when it exits it just tells the console to switch back to the main screen buffer.    Perhaps it's a bug in your version of windows that I was not aware of when I made the changes?

As I mentioned in another thread about colors - windows reports the wrong colours - it reports the wrong 16 color table - which Vim then saves, and Vim then uses the wrong colors to restore -  so, if the user has customised their color-scheme - Vim would end up overwriting the users custom colors.

zewpo:
> 16-color mode died completely.

I might be mis-undstanding something here.  Do all those issues you listed only happen in a low color console?

ntak:
It's not that the 16-color color table is wrong, it's that you dead code saving the color table on startup and dead code restoring the color table on exit...

Why did you replace USE_VTP, which does 256/16 color discrimination, with vtp_working, which has nothing to do with color discrimination and only indicates that vtp is implemented?

Why did you replace USE_WT, which determines additional processing required only by WindowsTerminal, with vtp_working, which has nothing to do with WindowsTerminal and only indicates that vtp is implemented?

USE_VTP is a macro that directs to 16 color mode specific code. Why did you remove the 16-color discrimination logic from it and replace it with vtp_working which really has nothing to do with 16-color mode?

zewpo:
> It's not that the 16-color color table is wrong, it's that you dead code saving the color table on startup and dead code restoring the color table on exit...

I'm not sure where there is dead code - at least, no more dead than it was before!?   I tried to be careful that the changes I made only applied to versions of windows terminals/consoles that use VTP.   If they don't use VTP - then the code should still run the same as it was running before.   - Win 7, 8 should still run the same code.

The gap where I felt may still be some risk - is for older versions of windows 10 that might not have the newer windows console fixes.   Windows 10 should be able to be upgraded for free - so I want to know - does  your problems only occur on older versions of windows 10?

> Why did you replace USE_VTP, which does 256/16 color discrimination, with vtp_working, which has nothing to do with color discrimination and only indicates that vtp is implemented?

I wanted  USE_VTP to actually mean it is using VTP.   I needed VTP to be usable regardless of color.   Let me ask you, why should color and VTP be related?  Maybe the true meaning of the variable was obscure to me.

> Why did you replace USE_WT, which determines additional processing required only by WindowsTerminal, with vtp_working, which has nothing to do with WindowsTerminal and only indicates that vtp is implemented?

On my versions of windows, there was no difference between the processing that WT does, and what the cmd console does.  The same problems were occuring in both, and the same fix I made worked for both!   I feel from your comments that I may have missed something though.  So,  can you please be more specific and say which additional processing is only applicable to WT that is not applicable to the cmd console?  And, if there are differences, are they still relevant in new versions of windows 10?

> USE_VTP is a macro that directs to 16 color mode specific code. Why did you remove the 16-color discrimination logic from it and replace it with vtp_working which really has nothing to do with 16-color mode?

So, again, let me ask you, why does USE_VTP have anything to do with color?  It was confusing the problem I was trying to fix - which actually relies on using VTP regardless of the color setting.   It did seem redundant to have two variable names "USE_VTP" as well as "vtp_working" - so I tried to rationalise that.   

You are correct that I fail to see why USE_VTP needs to be related to color - while vtp_working is not related to color.  So, I want to understand what I missed by making that change.

Today, I will try and collate the list of issues that I fixed - and lets see if we can restore your code while still maintaining the fixes that I tried to make.   It's fine if we find that some of my fixes need to be sacrificed where they cause more problems than they solve.  It's just that I want to understand properly - I want to learn from my mistakes.

zewpo:
And, the respective patches;

patch 9.0.1086: display wrong in Windows terminal after exiting Vim
Problem:    Display wrong in Windows terminal after exiting Vim.
Solution:   Apply screen restore fix for Windows 11 also to Windows 10 builds.
https://github.com/vim/vim/commit/c8b126d70d567269464fd70c54e22fdd78ec50ad


patch 9.0.0868: MS-Windows: after Vim exits console resizing problem
Problem:    MS-Windows: after Vim exits console resizing does not work
            properly.
Solution:   Restore screen behavior checks for various WT and VTP
            combinations. 
https://github.com/vim/vim/commit/1140b51e8369fed2321ce16bd4530a0952362126


patch 9.0.0850: MS-Windows Terminal has unstable color control
Problem:    MS-Windows Terminal has unstable color control.
Solution:   Do not try to read the old command prompt colortable, use modern
            VT sequences. 
https://github.com/vim/vim/commit/38804d64576c07caf06b381fae4578b88f535aa9

zewpo:
And, here are a couple other patches that will be partially undone by this PR;

patch 9.0.0344: MS-Windows: background color wrong in Console 
Yasuhiro Matsumoto on 9/1/2022, 8:31:45 PM
Problem:    MS-Windows: background color wrong in Console.
Solution:   Figure out the default console background color. (Yasuhiro
            Matsumoto, issue #10310)
https://github.com/vim/vim/commit/e42c8dae32f3575557b83a371c4be66c09afbf7f


patch 8.2.0848: MS-Windows: the Windows terminal code has some flaws 
Bram Moolenaar on 5/31/2020, 1:49:25 AM
Problem:    MS-Windows: the Windows terminal code has some flaws.
Solution:   Do not redraw the right edge of the screen.  Remove the background
            color trick.  Flush the screen output buffer often.  (Nobuhiro
            Takasaki, #5546)
https://github.com/vim/vim/commit/06b7b58455f9c09be3d1c36d174ffbfdf4efcd79

zewpo:
@ntak 
When I apply this PR to my own code; vim reverts to getting the wrong default colors;

with tgc:
image

Where does that pale gray come from?   See how it's hard to read!  That color is NOT the PINK color that I specified in here;

image

If I recall correctly, this seems to be the value from the 7th color in the ColorTable[16].

Now, not only is it the wrong color, Vim was writing the value of that color from position 7, into position 15 on exit.

Look at what this looks like before I start Vim;
image

Then, after I leave vim, and open up the properties dialog again;
image

See that Vim has indeed changed the users settings.  These end up being saved in the registry.

zewpo:
Another example;

I start with setting the text color to index position 2 (dark green)

image

vim -u NONE --clean +"set tgc ls=2"

image


VIm does not use the dark green color that I specified - ie. index 2.  
Instead, Vim has fallen back to using the color from color index 7 - the pale green color - which is wrong.

And then, worse;  After I exit Vim;
image

See, Vim has overwritten my custom value - in position index 2 - that was dark green before, and Vim has made it light green - ie. it has taken the value from index 7 and written that value into index 2.

So, I started trying to fix this stuff - and then realised, I could not see any good reason for vim to touch the users color table at all.

zewpo:
@ntak 
So, I think that will settle the color issues.


You also mentioned issues about buffering, speed and no longer able to enter ambi-width key events.

Is this where the problem is?

image

It used to check if !USE_WT and the change I made, was to make it say !vtp_working

Or, are you referring to other issues?

If you revert that back to !USE_WT, does it fix the ambi-width key events?  Sorry that I don't know how to test that.  If I could, I would test it and then I'd know what the problem is.

Also, buffering and scrolling speed are real issues that I want to fix, especially if I broke them as you blamed me for them.   But, please let me know where you think I broke those things, because I cant see where or how.  If you just give me some pointers, then I can know where to dig and investigate further.   I'd really appreciate that, thanks.

brammool:
This PR is included in the changes of PR #11836, therefore closing.
Please let us know if something of this PR went missing.

Issue #11532, opened 2022-11-12
[Regression][Windows] visualbell causes persistent color inversion in 9.0.856


kevinoid:
### Steps to reproduce

1. Install [Vim 9.0.0856 x64](https://github.com/vim/vim-win32-installer/releases/tag/v9.0.0856)
2. Run `vim --clean --cmd "set visualbell"` in PowerShell or `cmd.exe`
3. Press <kbd>k</kbd>

Observe that the terminal colors are now inverted:

![vim-9 0 856-visualbell-inverted](https://user-images.githubusercontent.com/336433/201454884-42e1a5b5-5c04-4577-b80f-e47e588cb593.png)

They are incrementally un-inverted as text is inserted:

![vim-9 0 856-visualbell-inverted-insert](https://user-images.githubusercontent.com/336433/201454925-5dbece1a-e53b-4492-b42e-af64af8f7955.png)

And fully un-inverted by pressing <kbd>Ctrl</kbd>+<kbd>L</kbd>:

![vim-9 0 856-visualbell-uninverted](https://user-images.githubusercontent.com/336433/201454957-6708981a-3f3b-4424-b04d-faaeabaa765c.png)

### Expected behaviour

Terminal colors would invert very briefly on alerts, as they did in 9.0.848.

### Version of Vim

9.0.856 and 9.0.861

### Environment

Windows 10.0.19043.2251
ConHost with PowerShell or cmd.exe

ntak:
The problem is in v9.0.0850.
This patch kills all ConPTY support I wrote earlier.
Why?

The 16-color mode has been completely crushed. I'm going to fix it steadily, but my heart is a little heavy.

zewpo:
For example;

![image](https://user-images.githubusercontent.com/10286211/212675790-32b918c2-762c-4f23-b461-f52325d51156.png)

set the text and background colours to something "customised", see the attached picture for example.

Then, see what old vim use to do to those colours.  Windows would report the colors based on a 16 colortable that has nothing to do with the actual current console colors.   Vim would save these values, assuming that they are the correct ones to restore after Vim exits.  BUT, they are not the correct colors to restore after the session.

zewpo:
more from that example - setting a bunch of odd "custom" colors;

![image](https://user-images.githubusercontent.com/10286211/212676582-b80fb252-5253-48fd-910e-3a0e1725fe21.png)

And, the new vim is correctly using those colors;

[...]

PR #11830, opened 2023-01-16
Revert visualbell in vim.exe 16 color mode

ghost:
#11532

16-color visualbell, screen recovery from background color inversion.
I can do it right away, so I did it right away, meow meow.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -6790,7 +6790,9 @@ visual_bell(void)
 			       coordOrigin, &dwDummy);
 
     Sleep(15);	    // wait for 15 msec
+# ifdef FEAT_TERMGUICOLORS
     if (!(p_tgc || (!p_tgc && t_colors >= 256)))
+# endif
 	WriteConsoleOutputAttribute(g_hConOut, oldattrs, Rows * Columns,
 				coordOrigin, &dwDummy);
     vim_free(oldattrs);
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -6790,7 +6790,7 @@ visual_bell(void)
 			       coordOrigin, &dwDummy);
 
     Sleep(15);	    // wait for 15 msec
-    if (!vtp_working)
+    if (!(p_tgc || (!p_tgc && t_colors >= 256)))
 	WriteConsoleOutputAttribute(g_hConOut, oldattrs, Rows * Columns,
 				coordOrigin, &dwDummy);
     vim_free(oldattrs);


zewpo:
Thanks @ntak   
I think this PR is good.   It fixes another thing that I missed.

brammool:
While we are at it: If the condition is false then WriteConsoleOutputAttribute() is not called.  That means "oldattrs" is not used.  Can we then skip ReadConsoleOutputAttribute() as well?  And also not allocate "oldattrs".

Although I wonder, if only FillConsoleOutputAttribute() is invoked to invert the attributes, when are the attributes restored?  And if that works, do we need the code to save and restore the attributes at all?

zewpo:
@brammool  

> when are the attributes restored?

good q, I will investigate.

zewpo:
So, current build from master - in a 16 color mode WCH (Windows Console Host).  It does NOT restore - it just stays reversed color!

unable to test with higher color mode on WCH at the moment.

It does restore for WT.  Not sure why yet - still investigating.

zewpo:
Alright, I've verified and confirm that this PR fixes the color restore problem for the visual bell.   It fixes it everywhere that is currently broken where I can replicate it.  Even in the WCH tgc issue, which I can test now too.

I'm still not sure how it restores in the case it is using tgc - still investigating that.  But, color does restore there too - not sure if related to this PR, or maybe it works via a fork in the control earlier in the logic before it gets to here.  If so, then that might explain the apparent dead code.!?  Still investigating.

PS. I've also worked out the cause of the WCH tgc regression issue - will issue a separate PR for that shortly.  Just mentioning it here because I was actually able to verify and confirm that this PR for visual bell is also fixed in this case too.

zewpo:
> While we are at it: If the condition is false then WriteConsoleOutputAttribute() is not called. That means "oldattrs" is not used. Can we then skip ReadConsoleOutputAttribute() as well? And also not allocate "oldattrs".

I tested this, and yes, it all still works, if we use the same if condition on the Read.    oldattrs are only needed if it is going to write them.   Let me know if you want another PR for this too?

> Although I wonder, if only FillConsoleOutputAttribute() is invoked to invert the attributes, when are the attributes restored? And if that works, do we need the code to save and restore the attributes at all?

yeah, restoring of the oldattr attributes here IS still needed for 16 color mode.   But not needed when color is >= 256,  then the colors are restored by code in term.c.  Actually - which my last PR broke!  I'm about to submit a PR that fixes what I broke in term.c

brammool:
Is this now fixes with Patch 9.0.1235 ?  I adjusted the condition for restoring to only do the work when saving succeeded.
Hopefully that works best.

zewpo:
I confirm that this visual bell fix merged into Patch 9.0.1235 is working well for MS-Console and Win-Terminal apps, ie. both working in 16 color mode as well as termguicolor modes.

And, so, I think the color regressions are mostly OK now.

Still lurking a regression somewhere for screen restore on exit - for the console buffer size and cursor shape don't get restored properly.   I can see an easy way to revert that regression, but that easy-fix then breaks the EOL fix that I did.    I know there is a way to fix both sets of issues, because I see that some projects using libuv manage to do both - just need some time to research how they do it and figure it out properly.

zewpo:
Just to wrap up this thread. PR https://github.com/vim/vim/pull/11882 completes fixing all of the regressions that I mentioned above.

brammool:
@ntak - please close this if the included patches cover what is in this PR.
Otherwise, please sync and let us know what's missing.

PR #11834, opened 2023-01-17
Accept 0x80-0xFF for setcellwidths()

k-takata:
Currently, setcellwidths() accepts only characters with value 0x100 or higher. However, there are many ambiguous width characters between 0xA1 to 0xFE.

See: https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt

This makes setcellwidths() to accept characters between 0x80 to 0xFF.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -7958,7 +7958,7 @@ setcellwidths({list})					*setcellwidths()*
 		{low} and {high} can be the same, in which case this refers to
 		one character.  Otherwise it is the range of characters from
 		{low} to {high} (inclusive).		*E1111* *E1114*
-		Only characters with value 0x100 and higher can be used.
+		Only characters with value 0x80 and higher can be used.
 
 		{width} must be either 1 or 2, indicating the character width
 		in screen cells.			*E1112*
--- a/src/errors.h
+++ b/src/errors.h
@@ -2841,8 +2841,8 @@ EXTERN char e_list_item_nr_cell_width_invalid[]
 	INIT(= N_("E1112: List item %d cell width invalid"));
 EXTERN char e_overlapping_ranges_for_nr[]
 	INIT(= N_("E1113: Overlapping ranges for 0x%lx"));
-EXTERN char e_only_values_of_0x100_and_higher_supported[]
-	INIT(= N_("E1114: Only values of 0x100 and higher supported"));
+EXTERN char e_only_values_of_0x80_and_higher_supported[]
+	INIT(= N_("E1114: Only values of 0x80 and higher supported"));
 EXTERN char e_assert_fails_fourth_argument[]
 	INIT(= N_("E1115: \"assert_fails()\" fourth argument must be a number"));
 EXTERN char e_assert_fails_fifth_argument[]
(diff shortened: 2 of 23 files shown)


brammool:
> Currently, setcellwidth() accepts only characters with value 0x100 or
> higher. However, there are many ambiguous width characters between
> 0xA1 to 0xFE.
> 
> See: https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt
> 
> This makes setcellwidth() to accept characters between 0x80 to 0xFF.


How about unprintable characters between 0x80 and 0x100?  I think four
should be returned then.  Something like:

	if (c >= 0x80)
	{
	  if (!vim_isprintc(c))
		return 4;		// unprintable, displays <xx>
	  int n = cw_value(c);
 	  if (n != 0)
 	    return n;



-- 
3. Every time someone asks you to do something, ask if they want fries
   with that.

k-takata:
I think they are already covered by this condition:
https://github.com/vim/vim/blob/35a12d2a17119dabeda7d234b7103157dea874c5/src/mbyte.c#L1626-L1628

Or, should we check if a character is unprintable before calling cw_value()?
For c >= 0x100, we call cw_value() before checking if it is unprintable.

k-takata:
I've updated the tests, and I've confirmed that `strwidth("\u0081")` returns 4 and `strwidth("\uFEFF")` returns 6 as expected.

> Might need a screendump test to verify each character shows up as expected.

It might be better to use a screendump, but I use `strwidth()` for now.

k-takata:
I had to change 'isprint' on Windows to get the expected result.

brammool:
> I've updated the tests, and I've confirmed that `strwidth("\u0081")`
> returns 4 and `strwidth("\uFEFF")` returns 6 as expected.
> 
> > Might need a screendump test to verify each character shows up as expected.
> 
> It might be better to use a screendump, but I use `strwidth()` for now.

I just checked, and it appears it's already wrong without your patch.
This works OK:

    call assert_equal(6, strwidth("\uFEFF"))

But then:

    call setcellwidths([[0xFEFF, 0xFEFF, 1]])
    call assert_equal(1, strwidth("\uFEFF"))

However, entering 0xFEFF in the text shows it's displayed as <feff>,
thus it does take 6 characters.  Nothing in the terminal can change
that.

The width that strwidth() returns should match what happens on the
display.  With your change, please also test with a character between
0x80 and 0x100.  E.g.:

    call setcellwidths([[0x81, 0x81, 1]])
    call assert_equal(4, strwidth("\u0081"))


-- 
`The Guide says there is an art to flying,' said Ford, `or at least a
knack. The knack lies in learning how to throw yourself at the ground
and miss.' He smiled weakly.

k-takata:
Updated. Could you check it?

brammool:
> Updated. Could you check it?

Thanks.  Looks like vim_isprintc() also needs to be used when
USE_WCHAR_FUNCTIONS is defined, it should return 4 then.

I'll have the changes to the .po files for the next runtime files
update.

-- 
The only backup you need is the one that you didn't have time for.
(Murphy)

PR #11836, opened 2023-01-17
restore ColorTable[16] usage

zewpo:
This PR includes all of the changes from PR raised by @ntak 
 https://github.com/vim/vim/pull/11829

That PR #11829 restores usage of the windows console 16-ColorTable - as that is still needed for some versions of windows consoles.  Unfortunately, I inadvertently lost that in my earlier PRs - I missed that ColorTable was still needed for some scenarios of windows consoles even if they use vtp.

But, as PR #11829 is now, it will revert other fixes that were made for Windows Terminal. 

So, this PR merges in @ntak 's PR #11829, but it does so while keeping the other fixes that were done for windows terminal.


brammool:
@ntak - please let us know what you think of this PR.

brammool:
Let me include this now.  If @ntak has comments we can handle that in another PR.

zewpo:
argghh..  ouch.  

I've just tested it as merged in the master branch...

So, while this; 
1. does fix Windows Console Hosts (WCH) 16 color mode (t_Co=16), ie. vim's default on windows consoles, and 
2. it keeps the Windows Terminal (WT) tgc color fixes (t_Co >=256).

BUT - unfortunately, it looks like something regressed (quite badly) when the vim setting is tgc  (or t_Co >=256) on the newer WCHs that are capable of using 24 bit colours.

I think we'll need to restore some the vtp < 256 color distinction (for non WT), so also need to restore some WT vs WCH distinction.   

I noticed this on Windows 11 22H2 (released Sept 2022).   So, I'm also going through and testing on a range of MS-windows versions (VMs) to see which ones are affected.

zewpo:
OK, https://github.com/vim/vim/pull/11871 fixes the termguicolor issue for windows conshole hosts.

PR #11837, opened 2023-01-17
Add getcellwidths() function

kat0h:
# [Proposal] Add `getcellwidths()`

Add getcellwidths() function.
This allows the user to retrieve the cell widths of character ranges set by setcellwidths().
The function sort and return the values.

### Symmetry of the function
Previously, it was not possible to retrieve the value set by `setcellwidths()`. The current situation is a bit unnatural since other functions have functions to retrieve settings.

### Add values to the table
Since `setcellwidths()` initializes the value of `cw_table`, it is not possible to set cell widths in multiple places.
`getcellwidths()` allows you to add entries in the following way...

```vim
call setcellwidths([[0x2729, 0x272f, 2], [0x2730, 0x273f, 2], [0x2740, 0x274d, 2]])
echo getcellwidths() " [[10025, 10031, 2], [10032, 10047, 2], [10048, 10061, 2]]

call setcellwidths(getcellwidths() + [[0x274f, 0x274f, 2]])
echo getcellwidths() " [[10025, 10031, 2], [10032, 10047, 2], [10048, 10061, 2], [10063, 10063, 2]]
```

### Easier debugging
Because `setcellwidths()` initializes the value of `cw_table`, if the function is called in multiple places by mistake, only the last setting will be reflected. Knowing the current settings is helpful for debugging.

## Supplement
- The help file may need to be rewritten (I'm not familiar with writing help file)
- It will need to be added the function name to `runtime/syntax/vim.vim`

Thank you.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -3265,9 +3265,9 @@ getbufvar({buf}, {varname} [, {def}])				*getbufvar()*
 			GetBufnr()->getbufvar(varname)
 <
 getcellwidths()						*getcellwidths()*
-		Returns list for cell widths of character ranges which
-		overrides by |setcellwidths()|. If cell widths of character
-		ranges doesn't exist, an empty list is returned.
+		Returns a |List| of cell widths of character ranges overridden
+		by |setcellwidths()|.  If no character ranges have their cell
+		widths overridden, an empty List is returned.
 
 getchangelist([{buf}])					*getchangelist()*
 		Returns the |changelist| for the buffer {buf}. For the use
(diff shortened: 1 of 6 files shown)


runtime/doc/builtin.txt
```diff
@@ -3262,6 +3264,11 @@ getbufvar({buf}, {varname} [, {def}])				*getbufvar()*
 <		Can also be used as a |method|: >
 			GetBufnr()->getbufvar(varname)
 <
+getcellwidths()						*getcellwidths()*
+		Returns list for cell widths of character ranges which
+		overrides by |setcellwidths()|. If cell widths of character
+		ranges doesn't exist, an empty list is returned.
```
```suggestion
		Returns a |List| of cell widths of character ranges overridden
		by |setcellwidths()|.  If no character ranges have their cell
		widths overridden, an empty List is returned.
```

kat0h: Thank you!

PR #11838, opened 2023-01-17
doc: Fix typos, adjust styles, etc.

k-takata:


brammool:
I'll include it, thanks.

PR #11845, opened 2023-01-18
 don't show function in completion candidates, if f_func was set to NULL,

kat0h:
Modified to not display functions that may not exist depending on the function, such as `if_ruby`'s `rubyeval()` function, in the completion candidates.

Example
```vim
:echo getcompletion("rubye", "function") " ['rubyeval(']
:call rubyeval() " E117: Unknown function
```


kat0h:
Thank you!

PR #11849, opened 2023-01-18
Remove hjson filetype detection

h3xx:
This isn't worth having in the base install for a couple of reasons:

1. This filetype doesn't have an associated syntax file; the only way to get syntax rules is via external plugin.
2. The official plugin, https://github.com/hjson/vim-hjson, comes with [its own `ftdetect/hjson.vim`](https://github.com/hjson/vim-hjson/blob/master/ftdetect/hjson.vim), so having this in vanilla Vim isn't even necessary to support the plugin.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -861,9 +861,6 @@ au BufRead,BufNewFile *.heex			setf heex
 " HEX (Intel)
 au BufNewFile,BufRead *.hex,*.h32		setf hex
 
-" Hjson
-au BufNewFile,BufRead *.hjson			setf hjson
-
 " HLS Playlist (or another form of playlist)
 au BufNewFile,BufRead *.m3u,*.m3u8		setf hlsplaylist
 
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -257,7 +257,6 @@ let s:filename_checks = {
     \ 'hercules': ['file.vc', 'file.ev', 'file.sum', 'file.errsum'],
     \ 'hex': ['file.hex', 'file.h32'],
     \ 'hgcommit': ['hg-editor-file.txt'],
-    \ 'hjson': ['file.hjson'],
     \ 'hlsplaylist': ['file.m3u', 'file.m3u8'],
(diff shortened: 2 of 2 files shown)


amaanq:
I don't think this undoes any change I've made, might've mentioned the wrong person 😅

clason:
What's the harm in leaving this in? There are a number of such filetypes that are there so users can rely on them for their own syntax or ftplugin scripts, and having them here makes sure that there are no conflicts with other detected filetypes (possibly added later). This way, everybody wins.

dundargoc:
I'm the one who made this change. This is ultimately up to the Vim project if they want it or not, but I personally don't think this is an improvement for the reasons @clason mention.

For your second point: shouldn't the plugin delete their own ftdetect instead since this functionality is already upstreamed to Vim? In fact, if hjson doesn't have a syntax file in Vim I think vim-hjson should upstream their syntax file to Vim and be a maintainer for the file so that hjson users get the benefit automatically.

brammool:
> I'm the one who made this change. This is ultimately up to the Vim
> project if they want it or not, but I personally don't think this is
> an improvement for the reasons @clason mention.
> 
> For your second point: shouldn't the plugin delete their own ftdetect
> instead since this functionality is already upstreamed to Vim? In
> fact, if hjson doesn't have a syntax file in Vim I think vim-hjson
> should upstream their syntax file to Vim and be a maintainer for the
> file so that hjson users get the benefit automatically.

If we would drop the .hjson detection, I don't see how it would be
better.  It doesn't fall back to another filetype.

With the existing detection it's easier to add a syntax or indent file.

-- 
The startling truth finally became apparent, and it was this: Numbers
written on restaurant checks within the confines of restaurants do not follow
the same mathematical laws as numbers written on any other pieces of paper in
any other parts of the Universe.  This single statement took the scientific
world by storm.  So many mathematical conferences got held in such good
restaurants that many of the finest minds of a generation died of obesity and
heart failure, and the science of mathematics was put back by years.

chrisbra:
Let's keep it, I don't see how it does any harm currently

Issue #11447, opened 2022-10-26
command-completion of `:runtime` should list files and directories in `runtimepath` and/or `packpath` and not current directory


bfrg:
### Steps to reproduce

In Vim type `:runtime` and then press `<C-D>` (or press `<Tab>`).

Result: Vim will list directories and files in the _current working directory_ and not the ones in `runtimepath` and/or `packpath`.

### Expected behaviour

In `:runtime` Vim should tab-complete files and directories in `runtimepath` and/or `packpath` instead of the current working directory.

### Version of Vim

9.0.815

### Environment

Linux

PR #11853, opened 2023-01-20
resolved bug #11447 command-completion of :runtime should list files …

cero1988:
resolved bug #11447 command-completion of :runtime should list files in runtime path

### DIFF
--- a/src/cmdexpand.c
+++ b/src/cmdexpand.c
@@ -56,6 +56,7 @@ cmdline_fuzzy_completion_supported(expand_T *xp)
 	    && xp->xp_context != EXPAND_OLD_SETTING
 	    && xp->xp_context != EXPAND_OWNSYNTAX
 	    && xp->xp_context != EXPAND_PACKADD
+	    && xp->xp_context != EXPAND_RUNTIME
 	    && xp->xp_context != EXPAND_SHELLCMD
 	    && xp->xp_context != EXPAND_TAGS
 	    && xp->xp_context != EXPAND_TAGS_LISTFILES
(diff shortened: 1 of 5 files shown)


brammool:
```
	char *directories[] = {"", NULL};
```

The empty string is used to create a pattern for the matched path:

```
	sprintf((char *)s, "%s/%s*.vim", dirnames[i], pat);
```

Have a try using "**" instead of the empty string, doesn't that work
better?

However, if the pattern already includes a path separator, e.g.
"syntax/f", then it's probably better not to use a directory and only
append "*.vim".

-- 
33. You name your children Eudora, Mozilla and Dotcom.

cero1988:
you are right, it's working better with **

yegappan (commented):
Can you add a test for this new completion type in test_cmdline.vim?

Also, you need to add "runtime" to command_complete[] in usercmd.c.

cero1988:
> Can you add a test for this new completion type in test_cmdline.vim?
> 
> Also, you need to add "runtime" to command_complete[] in usercmd.c.

I have added the test in test_cmdline.vim and the runtime to command_complete

src/cmdexpand.c
```diff
@@ -3019,6 +3026,12 @@ ExpandFromContext(
 	return ExpandRTDir(pat, DIP_START + DIP_OPT, numMatches, matches,
 								directories);
     }
+    if (xp->xp_context == EXPAND_RUNTIME)
+    {
+	char *directories[] = {"", NULL};
+	return ExpandRTDir(pat, DIP_START + DIP_OPT + DIP_PRNEXT, numMatches, matches,
+								directories);
```
yegappan, line 3033: From the above codecov output, it looks like most of the key changes are not covered by tests. Can you modify the tests to cover these lines?

src/cmdexpand.c
```diff
@@ -2311,6 +2313,11 @@ set_context_by_cmdname(
 	    xp->xp_context = EXPAND_COLORS;
 	    xp->xp_pattern = arg;
 	    break;
+	
+        case CMD_runtime:
+	    xp->xp_context = EXPAND_RUNTIME;
+            xp->xp_pattern = arg;
+	    break;
```
yegappan, line 2321: Can you update the tests to cover the above lines?  You can look at the Test_cmdline_complete_various() function in the test_cmdline.vim file for some examples for this test.

PR #11860, opened 2023-01-22
Add additional ability to (optionally) decorate the Colorcolumn with a character and/or custom highlighting groups

jrahm:
Hi all,

Several months ago I created [this](https://github.com/neovim/neovim/pull/19869) pull request for Neovim to add a new fillchar, colorcol. They recommended I create a pull request upstream to see if there was any interest. So after a few months, I finally got up the energy to give it a shot.

I use Vim's colorcolumn often, however, I prefer the look and feel of more recent editors where the "color column" is just a thin line. Highlighting a full block just feels ... inelegant. My solution to add this to vim, is to add a new fillchar, colorcol. This char is displayed in the colorcolumn using the colorcol highlighting. So ,for me, where I prefer the thin line, I set colorcol to `│`, but others may decide to set it to `┆`, `~`, or `.`, or `☃`, though I imagine most would probably set it to some flavor of the vertical box-drawing characters.

Here is an exmaple of how it works with basic configuration in what I imagine will be the typical usage:
colorcolchar_1

And if one is so possessed to use multiple colorcolumns with this option:
colorcolchar_2

Let me know what else might me needed. I tried to add a test and follow existing conventions, but may have overlooked something. Thanks!


brammool:
> Several months ago I created
> [this](https://github.com/neovim/neovim/pull/19869) pull request for
> Neovim to add a new fillchar, colorcol. They recommended I create a
> pull request upstream to see if there was any interest. So after a few
> months, I finally got up the energy to give it a shot.
> 
> I use Vim's colorcolumn often, however, I prefer the look and feel of
> more recent editors where the "color column" is just a thin line.
> Highlighting a full block just feels ... inelegant. My solution to add
> this to vim, is to add a new fillchar, colorcol. This char is
> displayed in the colorcolumn using the colorcol highlighting. So ,for
> me, where I prefer the thin line, I set colorcol to `│`, but others
> may decide to set it to `┆`, `~`, or `.`, or `☃`, though I imagine
> most would probably set it to some flavor of the vertical box-drawing
> characters.

[...]

Using 'fillchars' for this character has one important limitation: you
can specify only one character.  I wonder if that is OK or too limited.

I imagine 'colorcolumn' entries can be used for at least two purposes:
marking columns in the text (e.g. at column 10 and 40) to align things,
and for marking the right margin.  Suppose someone only wants to use the
character for the right margin, since it's where text normally doesn't
show, while using a character in the middle of the text can be annoying.

We only have one highlight group: ColorColumn.  That's limited as well.
Thus perhaps limiting the character is OK?  On the other hand, if we
would be able to define the character for each column, then being able
to define the highlighting for each column is probably needed as well.
If no character is defined the highlight would use the background, while
with a character it would use the foreground color.

The character and the highlighting could be included in the
'colorcolumn' option value:

   set colorcolumn=10//LeftCol,40//MiddleCol,75/|/RightCol

-- 
A day without sunshine is like, well, night.

jrahm:
I like this idea. I thought it might be cool to have multiple colorcolumns with different highlights. One for maybe a "soft" textwidth recommendation and one for a "hard" textwidth (for example). I suspect it wont be too much trouble to implement. (Famous last words, amirite) So I'm willing to give it a shot.

I assume the syntax will be something like
```
colorcol     := <colorcolitem>,...
colorcolitem := <number>        // just use the number with ColorColumn formatting and a <space>' as the character
              | <number>/<character?>/<highlight?>  // Use number, character and highlight. If highlight is omitted, the ColorColumn is used, if character is omitted, then <space> is used.
```

ofc if someone is so possessed as to use a forward-slash for the character, it would need to be escaped.

I like this idea.

habamax:
Imagine bigger text, or scroll 2 lines down and you've lost visual hint for the colorcolchar:

image

Other gui editors draw it in a different way, for example:
image

jrahm:
I understand the concern, and I appreciate you bringing it up, a couple of things:

1. This PR doesn't change the existing semantics of the ColorColumn (at least it's not supposed to), just extends them to give the user more ability to decorate them how they may like. Especially after incorporating brammool's ideas (though this is still a WIP). Perhaps the name of my PR needs to be changed. I apologize if this was unclear.
1. There would still be a visual cue. The letters in the colorcolumn are hl-combine'd with ColorColumn, so characters will inherit ColorColumn. But it would be an issue if all the characters were highlighted. But, sure, if the ColorColumn hlgroup were cleared, then there would be no visual cue, but that's already the case. (Though I personally, for my use case, wish the colorcolumn would just not display if a character was written over it, but that's my use case.)
1. I think it should be on the user to use the proper ColorColumn configuration for their use-case, no? Afterall it's on the user to determine the proper ColorColumn column(s). Maybe I'm a bit myopic because I'm primarily a coder and so the colorcolumn indicating the rightmost margin is really my only real use-case for it, in which case, lines crossing the colorcolumn are sparse so the box drawing vertical line example is perfect. In the example above where the text crossing the colorcolumn is dense, then background highlighting is probably the way to go. I also never use GVim, only terminal vim, so maybe there's some sharp edges I'm not aware of.
1. GUIs are always going to have an advantage when it comes to drawing arbitrary stuff on the window, but, at least in the terminal, I think this is a good compromise.

I'm happy to answer any more questions

jrahm:
PTAL

Here's a couple of example of how I might use this new feature:

colorcol_ex1

here this is set with `set colorcolumn=80/┆/ColorCol2/b,100/│/ColorCol3,60/\ /ColorColumn,4`

colorcol_ex2

This is set with a script that add a colorcolumn at column 4, one at 80, and then ones at every 4 columns from 4 to 80

brammool:
I have my doubts about trying to put all of this into one option value.  It can work for maybe up to three items, but above that it quickly gets messy.  And if we get more requests later it's going to be hard to extend.  We could keep 'colorcolumn' simple and add a function to set more complicated configurations.  Then a dictionary can be used that's easy to extend with more entries later.  When using a script to set this a function will be easier to use as well.
Considering that not much comments are given here it doesn't sound like an important feature to add (compared to all the other features that have been requested).

chrisbra:
While this: 
```
set colorcolumn=80/┆/ColorCol2/b,100/│/ColorCol3,60/\ /ColorColumn,4
```
is quite flexible, I think it is also quite obscure. I think we would also loose the ability to use `:set cc+=1` to highlight the current textwidth. 

I think it would make more sense to keep colorcolumn simple, and have an additional (secondary) option `'colorcolumnopt'` for additional options. Using an Vim dictionary with certain keys `{'char1': '|', 'highlight1': 'ColorColumn1'}` etc should be fine, if no options given, Vim would use the defaults settings.

chrisbra:
closing for now. If you ever finish this up, please create a new PR. Thanks.

PR #11863, opened 2023-01-22
Fix completion for :runtime doesn't match what can be used

zeertzjq:


zeertzjq:
Hmm, this includes superfluous results on Windows.

zeertzjq:
Changed to use `EXPAND_DIRECTORIES` for directories instead. This can also fix `:packadd` expansion including unrelated files. Hopefully this will fix the test failures on Windows.

brammool:
This handles the {where} argument if it's there, but it doesn't do completion of {where}.
I suppose we can include this PR and hope someone adds {where} completion later.

PR #11867, opened 2023-01-22
Remove condition that is always true

zeertzjq:


brammool:
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/vim/vim/pull/11867
> 
> -- Commit Summary --
> 
>   * Remove condition that is always true

The condition is not true the second time a COM_START character is
encountered.

The code could be made clearer by not keeping "lnum_line", initializing
line_len to -1 and then computing line_len only when it's negative.
This uses strlen() which is not that fast. 

I first thought that lnum_line could be used, but it turned out that
wasn't worth it.

-- 
Change is inevitable, except from a vending machine.

zeertzjq:
There is a return statement at the end of the branch, so it will not be reached a second time.

brammool:
> There is a return statement at the end of the branch, so it will not
> be reached a second time.

Only when the condition is true.  Otherwise it will check again.

-- 
Portable Computer:  A device invented to force businessmen
to work at home, on vacation, and on business trips.

zeertzjq:
Then the `*p == COM_START` branch is only ever executed once in this function, and `lnum_line` is also only assigned in this branch. Isn't `lnum_line` only assigned once anyway?

brammool:
> Then the `*p == COM_START` branch is only ever executed once in this
> function, and `lnum_line` is also only assigned in this branch. Isn't
> `lnum_line` only assigned once anyway?

I think what you are trying to say is that once the condition in

	 if (*p == COM_START)

is true, then we are certain to encounter a "return" statement.
Thus the "for" loop will be abandoned.

OK, then we can simplify it to:

	int line_len = (int)STRLEN(ml_get(lnum));

-- 
Just remember...if the world didn't suck, we'd all fall off.

PR #11871, opened 2023-01-23
ms-win flush termguicolors color for visual bell

zewpo:
prb:  ms-win console not flushing termguicolors
sln: flush termguicolors

This fixes the flushing of the termguicolors in the new versions of the windows console host - that are capable of using 24 bit colors.  (ie. the flushing is not just for windows terminal.)

Flushing the termguicolors was also needed for the visual_bell feature to work properly.   So, this PR also merges in all of @ntak PR:  Revert visualbell in vim.exe 16 color mode #11830


zewpo (commented):
ready for review

zewpo:
nice improvement with :) LPWORD oldattrs = NULL;

PR #11872, opened 2023-01-23
Some gpg keywords are not supported

ObserverOfTime:
Problem:    Some gpg keywords are not supported.
Solution:   Add "default-new-key-algo", "weak-digest" keywords.


chrisbra:
Hm,
I believe Nikolai is no longer maintaining any Vim runtime files. This begs the question: Would you like to volunteer to maintain this file?

ObserverOfTime:
> Would you like to volunteer to maintain this file?

Not at this time, but I might change my mind in the future.

brammool:
I'll include it.  Instead of modifying the "latest Revision" line I'll add a line for this update.

PR #11873, opened 2023-01-23
Some xf86conf keywords are not supported

ObserverOfTime:
Problem:    Some xf86conf keywords are not supported.
Solution:   Add keywords for InputClass match entries.

---

I also removed the reference to the dead URL.


ObserverOfTime:
(I won't assume the maintenance of this file either)

brammool:
Have you tried the email in the header?  It was working last year.

ObserverOfTime:
No, because he's listed as the *former* maintainer.

brammool:
OK, I'll include the changes then.  But please notify David, I don't know if he has a copy somewhere else.

PR #11874, opened 2023-01-23
Include values for [where] argument in :runtime completion

zeertzjq:
~I was initially trying to make :runtime completion included values for
[where] argument, during which I found this bug that :runtime completion
doesn't where if pattern is a prefix of a valid [where] argument.~

~I didn't manage to include values for [where] argument in completion,
but I did some refactorings to make this easier to add.~


zeertzjq:
Oh, never mind. I just managed to included values for [where] argument.

brammool:
> I was initially trying to make :runtime completion included values for
> [where] argument, during which I found this bug that :runtime completion
> doesn't where if pattern is a prefix of a valid [where] argument.
> 
> I didn't manage to include values for [where] argument in completion,
> but I did some refactorings to make this easier to add.

Thanks for further improving this.

The logic in expand_runtime_cmd() can be improved a bit.  Suppose there
is a script "Awesome.vim" and the user types ":runtime A<Tab>", then I
expect the suggestions to not only contain "Awesome.vim" but also "ALL",
since both match.  Looking at the code this currently doesn't happen.

I haven't tried the patch, but I guess that get_runtime_cmd_flags()
needs to be adjusted to avoid partial matches of the {where} values.
When typing "A" then "where_len" is probably one, and this condition:

	    if (STRNCMP(arg, "ALL", where_len) == 0)

would be true.  Perhaps add "where_len >= 3" ?

-- 
Latest survey shows that 3 out of 4 people make up 75% of the
world's population.

zeertzjq:
> Suppose there is a script "Awesome.vim" and the user types ":runtime A<Tab>", then I expect the suggestions to not only contain "Awesome.vim" but also "ALL", since both match.

This is already the case after the second commit I pushed. I'll squash the commits then.

PR #11876, opened 2023-01-24
Prefer lead whitespace character over trail whitespace character

timgianitsos:
See description in commit message.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -2892,11 +2892,11 @@ win_line(
 			    multispace_pos = 0;
 		    }
 
-		    else if (ptr > line + trailcol && wp->w_lcs_chars.trail)
-			c = wp->w_lcs_chars.trail;
-
 		    else if (ptr < line + leadcol && wp->w_lcs_chars.lead)
 			c = wp->w_lcs_chars.lead;
+
+		    else if (ptr > line + trailcol && wp->w_lcs_chars.trail)
+			c = wp->w_lcs_chars.trail;
 
 		    else if (leadcol != 0 && wp->w_lcs_chars.space)
 			c = wp->w_lcs_chars.space;
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -2881,42 +2881,43 @@ win_line(
 		}
 
 		if (c == ' ' && ((trailcol != MAXCOL && ptr > line + trailcol)
-				|| (leadcol != 0 && ptr < line + leadcol)))
+				    || (leadcol != 0 && ptr < line + leadcol)))
 		{
-			if (leadcol != 0 && in_multispace && ptr < line + leadcol
-					&& wp->w_lcs_chars.leadmultispace != NULL)
-			{
-				c = wp->w_lcs_chars.leadmultispace[multispace_pos++];
-				if (wp->w_lcs_chars.leadmultispace[multispace_pos]
+		    if (leadcol != 0 && in_multispace && ptr < line + leadcol
+			    && wp->w_lcs_chars.leadmultispace != NULL)
+		    {
+			c = wp->w_lcs_chars.leadmultispace[multispace_pos++];
+			if (wp->w_lcs_chars.leadmultispace[multispace_pos]
 									== NUL)
-				multispace_pos = 0;
-			}
+			    multispace_pos = 0;
+		    }
 
-			else if (ptr < line + leadcol && wp->w_lcs_chars.lead)
-				c = wp->w_lcs_chars.lead;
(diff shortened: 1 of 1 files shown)


timgianitsos:
Fix for:
https://vi.stackexchange.com/questions/39755/how-to-make-lead-override-trail-in-listchars

brammool:
The diff mainly shows lines that have been indented differently.  Please use a 'shiftwidth' value of four.
Once that is fixed we can better see what you are actually changing.
This also deserves a change in the help for 'listchars' and a test case.

zeertzjq:
Why is it preferred to present them as leading?

chrisbra:
> https://vi.stackexchange.com/questions/39755/how-to-make-lead-override-trail-in-listchars

Ah, I guess my hint was useful :) 

This diff here makes it more clear: https://github.com/vim/vim/pull/11876/files?diff=unified&w=1

It basically just switches the `elseif` conditions.

However, if we change this now, I am sure other people will complain :(
So this should be made configurable. But just adding yet another option for this minor change is also not nice...

So perhaps it makes sense to enhance the 'listchars' option to somehow allow to configure priorities for 'lead'/'trail'?

I vaguely remember this being discussed when the 'lead' or 'trail' option was added to 'listchars', but cannot find references.

timgianitsos:
> Ah, I guess my hint was useful :)

Yup, thanks @chrisbra  :)

> This diff here makes it more clear: https://github.com/vim/vim/pull/11876/files?diff=unified&w=1

I also delineated into two separate commits to make it easier to view. The more important commit is [here](https://github.com/vim/vim/pull/11876/commits/dd432addd2d78c07c16b00c600e3734ba372e41b).

That's a cool trick with the `w=1` query param @chrisbra. For future observers' reference, I found a source explaining this feature of Github to ignore whitespace discrepancies in diffs [here](https://github.blog/2018-05-01-ignore-white-space-in-code-review/).

> Why is it preferred to present them as leading?

Whenever one starts typing at the end of an indented line, there is a sudden conversion of all whitespace characters from trailing form to leading form. IMO it would be preferable as default behavior to have them be leading form the whole time instead of this sudden conversion from trailing form to leading form.

What is lost by doing this? Well if we follow the convention of my pull request, then any scenario where you start typing at the _beginning_ of a line that already has indentation will convert leading form into trailing form. But it is almost _never_ the case that you start typing at the beginning of a line before the whitespace.

Having precedence between whitespace characters is not without... precedent :)
In the docs from `:help listchars`, multiple of the options "override" each other. There is no configurability, but the precedence/overriding behavior makes intuitive sense.

zeertzjq:
1. I don't think this PR works at all. You didn't change line 1522.
2. One may set "trail" in 'listchars' to a character that is easy to spot, so they can remove them. Hardly anyone wants to remove leading spaces.

timgianitsos:
1. Since this precedence is recognized in more than one place (line 1522 as you said) was there originally a deeper reason why `trail` is preferred to `lead`?
1. 
> Hardly anyone wants to remove leading spaces.

I agree! They are just not being recognized as leading spaces until you type another character after. While you think about what to type on an indented line, you have trailing characters staring out you that you are tempted to delete. Typing a dummy character after the whitespace is the only remedy.

brammool:
> > This diff here makes it more clear: https://github.com/vim/vim/pull/11876/files?diff=unified&w=1
> 
> I also delineated into two separate commits to make it easier to view. The more important commit is [here](https://github.com/vim/vim/pull/11876/commits/dd432addd2d78c07c16b00c600e3734ba372e41b).

It's not just about viewing the diff, the indent is not right.
The "if" block has 4 spaces too much.

-- 
Light travels faster than sound.  This is why some people
appear bright until you hear them speak

k-takata:
The commit https://github.com/vim/vim/pull/11876/commits/e85b23b1467103b0d0677f85e3ba026801233848 should be removed. The commit message says:

> Fix lines with mixed tab and space to only use tabs

but Vim's source code intentionally uses mixed tab and space.

timgianitsos:
> Vim's source code intentionally uses mixed tab and space

I reverted to using mixed tabs and spaces.

I am not sure how to address comments from @zeertzjq about the potential for the convention (i.e. trail overrides lead) to be assumed elsewhere.

brammool:
I have doubts whether this is an improvement.  For me both trailing spaces after text and a line with just spaces are characters that should not be there.  Leading spaces are different.  Therefore using "trail" for spaces in a line without text makes sense to me.

If some users don't like this, how about adding another character for spaces in a line without any text?  
It could be called "empty" perhaps.  That also makes it backwards compatible.

chrisbra:
let's not do this.

PR #11880, opened 2023-01-25
Some nginx keywords are not supported

ObserverOfTime:
This adds support for several keywords from

* the built-in [HTTP/2 module](https://nginx.org/en/docs/http/ngx_http_v2_module.html),
* the built-in [SSL module](https://nginx.org/en/docs/http/ngx_http_ssl_module.html),
* the built-in [uWSGI module](https://nginx.org/en/docs/http/ngx_http_uwsgi_module.html),
* the experimental [QUIC branch](https://quic.nginx.org/readme.html),
* the third-party [SSL CT module](https://github.com/grahamedgecombe/nginx-ct),
* the third-party [dynamic TLS records patch](https://github.com/cloudflare/sslconfig/blob/master/patches/nginx__dynamic_tls_records.patch).

/cc @chr4

### DIFF
--- a/runtime/syntax/nginx.vim
+++ b/runtime/syntax/nginx.vim
@@ -1,7 +1,7 @@
 " Vim syntax file
 " Language: nginx.conf
 " Maintainer: Chris Aumann <me@chr4.org>
-" Last Change: Apr 15, 2017
+" Last Change: Jan 25, 2023
 
 if exists("b:current_syntax")
   finish
@@ -84,6 +84,8 @@ syn keyword ngxListenOptions default_server contained
 syn keyword ngxListenOptions ssl            contained
 syn keyword ngxListenOptions http2          contained
 syn keyword ngxListenOptions spdy           contained
+syn keyword ngxListenOptions http3          contained
+syn keyword ngxListenOptions quic           contained
 syn keyword ngxListenOptions proxy_protocol contained
 syn keyword ngxListenOptions setfib         contained
 syn keyword ngxListenOptions fastopen       contained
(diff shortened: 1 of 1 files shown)


chr4:
From a [previous pull request](https://github.com/vim/vim/pull/9003), policy was to file it upstream and then let the maintainer create the pull request, if I recall correctly. If that's still the case, feel free to file this to https://github.com/chr4/nginx.vim instead - I can then file the pull request if required. 

My 2ct: Not sure about including third party directives. They seem to be good projects on first sight, just never came up with a policy here. Don't mind those in particular, but it might be a bit overwhelming if there's too many third party plugins included. Any thoughts here from someone with more experience?

ObserverOfTime:
> policy was to file it upstream and then let the maintainer create the pull request

Sure, I can do that.

> Not sure about including third party directives.

I only included them because I saw that a few others were already present.

chr4:
> I only included them because I saw that a few others were already present.

🙈 didn't remember. I think I wouldn't mind a few extra ones, we can still purge once they become an overwhelming amount.

chr4:
I've merged this upstream. Shall I create a new pull request as a maintainer? Merging this would be fine for me, too.

ObserverOfTime:
I can reopen if needed.

chrisbra:
> I've merged this upstream. Shall I create a new pull request as a maintainer? Merging this would be fine for me, too.
Just let me know.

Sure, please go ahead

brammool:
I prefer the author/maintainer to send me a PR or patch or just the new version of the file.  That avoids multiple versions floating around, I might include the wrong one or run into merge conflicts.

chr4:
Follow-Up: https://github.com/vim/vim/pull/11919

PR #11882, opened 2023-01-25
MS-Windows: Console buffer scrollback being cropped off when buffer resized by SetConsoleScreenBufferInfoEx

zewpo:
prb:  Console buffer scrollback being cropped off on Vim exit, after SetConsoleScreenBufferInfoEx is called when using alternate screen buffer.

sln:  Don't call SetConsoleScreenBufferInfoEx when using alternate screen buffer.

risk:  This potentially regresses to an issue that @ntak raised - about the 16-colortable not being restored correctly after exiting Vim under some circumstances.  

But,  I'm unable to reproduce that issue - I've tried a lot of different ways.   I'd really like @ntak 's help to check this PR out.

Previously, my solution for fixing the problems with restoring the screen-buffer on Vim exit, was to bypass SetConsoleScreenBufferInfoEx on the condition "vtp_working".    But, I suspect that bypass is what caused some of the problems that @ntak found.     I've not been able to reproduce that particular issue, so giving @ntak the benefit of the doubt - I removed the condition - and the code was reverted back to always calling SetConsoleScreenBufferInfoEx.    

Unfortunately, calling SetConsoleScreenBufferInfoEx  causes the original buffer scrollback to be cropped on Vim exit.   And I can't see why SetConsoleScreenBufferInfoEx is needed - especially when "use_alterate_screen_buffer" is true.    So, in this PR, I've reinstated the bypass of the SetConsoleScreenBufferInfoEx - but this time based on the condition of "use_alterate_screen_buffer". 

"use_alterate_screen_buffer" is narrower than "vtp_woring", and is more pertinent to the problem - hopefully it solves the restore issue, without regressing the 16-color-table issue again.


zewpo:
@brammool   From my point of view - this completes fixing all of the remaining regressions following on from 9.0.0850.

As you can see, I'm still a little bit uncertain, because I was never able to replicate some of the issues that @ntak raised.   If there are any more problems yet remaining, I will need help to see how to replicate.  I don't mind putting in the time and effort to sort it out, I just need to see what the problems are, if they still exist.

brammool:
I waited a bit to see if any comments popped up, but that didn't happen, so let me include this now.

PR #11891, opened 2023-01-26
Fix a small typo in vimtutor

MisterrrX:
I know it's not much, but I just noticed this while doing the vimtutor and thought why not fix it :)


brammool:
Not sure why you closed this.  I suppose the change is OK.

k-takata:
In that case, you could add the new commit to your branch. No need to open a new PR.

PR #11892, opened 2023-01-26
Fix small typo in vimtutor

MisterrrX:
I know it's not much, but I just noticed this while doing the vimtutor and thought why not fix it :)


brammool:
I'll include it.

PR #11895, opened 2023-01-26
doc: Add {{{ and }}} to the 'fold-marker' help tag

druckdev:
Seeing `{{{` in code without exactly knowing what it does makes one curious. Help find information about the fold markers without knowing how they are called by adding the default 'foldmarker' value to the help tag.


brammool:
I'll add the help tags, a bit lower than your suggestion.

PR #11902, opened 2023-01-28
Fix for messages behavior depends on 'fileformat' of current buffer 

### LINKED ISSUE #11735: messages behavior depends on 'fileformat' of current buffer
### Steps to reproduce

1. Run `vim --clean`
2. Run `:echomsg "\r"`
3. `^M` is printed
4. Run `:messages`
5. `^M` is printed
6. Run `:setlocal fileformat=mac`
7. Run `:messages`
8. `^J` is printed

### Expected behaviour

Changing 'fileformat' of current buffer should not affect messages behavior

### Version of Vim

9.0.1046

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

cero1988:
This fixes #11735

### DIFF
--- a/runtime/doc/map.txt
+++ b/runtime/doc/map.txt
@@ -20,9 +20,10 @@ manual.
    1.8 Examples					|map-examples|
    1.9 Using mappings				|map-typing|
    1.10 Mapping alt-keys			|:map-alt-keys|
-   1.11 Mapping in modifyOtherKeys mode		|modifyOtherKeys|
-   1.12 Mapping with Kitty keyboard protocol	|kitty-keyboard-protocol|
-   1.13 Mapping an operator			|:map-operator|
+   1.11 Mapping meta-keys			|:map-meta-keys|
+   1.12 Mapping in modifyOtherKeys mode		|modifyOtherKeys|
+   1.13 Mapping with Kitty keyboard protocol	|kitty-keyboard-protocol|
+   1.14 Mapping an operator			|:map-operator|
 2. Abbreviations		|abbreviations|
 3. Local mappings and functions	|script-local|
 4. User-defined commands	|user-commands|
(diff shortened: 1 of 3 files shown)


chrisbra:
I slightly changed your description

brammool:
What message depends on the 'fileformat' value of the buffer?  Can't we use all messages with EOL_UNIX?

cero1988:
before to print each message, the eol characters are printed as It were printed originally
before committing, the eol characters were being all printed in regard of the current buffer filetype

cero1988:
now I have understood
I agree, we can use EOL_UNIX for all messages
I will update the pull requests to do so

cero1988:
I have updated the pull request, what do you think ?

brammool:
> now I have understood
> I agree, we can use EOL_UNIX for all messages
> I will update the pull requests to do so

Seeing the adjusted patch, the CAR to NL translation, when 'fileformat'
is "mac" for the buffer, is used for the :echomsg command and for the
:messages command.  I don't understand why.  The translation is useful
when taking text from the buffer that is using Mac format.  These
commands do not really do that.  Especially :messages, which lists
various messages with text from any buffer.

How about this approach instead:
- In transchar_nonprint(), when "buf" is NULL then do not check for
  'fileformat' to be EOL_MAC.  Thus non CAR to NL translation
- In most places in src/message.c where a "transchar" function is
  called, pass a NULL buffer. E.g. "transchar(name)" would become
  "transchar_buf(NULL, name)".


-- 
91. It's Saturday afternoon in the middle of May and you
    are on computer.

cero1988:
thanks for the tips, I will update the code and reopen the pull request

PR #11909, opened 2023-01-29
Add Elsa syntax

glapa-grossklag:
Elsa files are recognized as of patch 9.0.1261. This patch adds a syntax file for the Elsa language.


brammool:
I'll include it, thanks.

PR #11914, opened 2023-01-30
Fix missing help tags for 'keywordprg'

a5ob7r:
Help tags for 'keywordprg' are recognized as help example contents because of a missing close pattern of help example. So Vim can't find these tags.


brammool:
I'll include it, thanks.  Will be in the next runtime files update.

PR #11916, opened 2023-01-30
Add `.stylelintrc` and `.clangd` file type detection

mskelton:
This adds filetype detection for `.stylelintrc` and `.clangd` files.  Below are links to their docs if you are interested.

- https://stylelint.io/user-guide/configure/
- https://clangd.llvm.org/config#files


brammool:
Please add entries in the filetype test: src/testdir/test_filetype.vim

mskelton:
Tests added

PR #11919, opened 2023-01-30
Add support for missing keywords to the nginx syntax plugin

chr4:
This adds support for several keywords from
- the built-in HTTP/2 module,
- the built-in SSL module,
- the built-in uWSGI module,
- the experimental QUIC branch,
- the third-party SSL CT module,
- the third-party dynamic TLS records patch.

Co-Author: @ObserverOfTime <chronobserver@disroot.org>


brammool:
I'll include it, thanks.

PR #11933, opened 2023-02-02
Update Serbian messages translation

eevan78:


brammool:
I'll include it, thanks.

PR #11939, opened 2023-02-03
Fix setbufvar() move cursor position in Visual mode

zeertzjq:


zeertzjq:
The last change is possibly not needed, as it is matters on failure to find the original window, but it's better to be more consistent anyway.

PR #11946, opened 2023-02-04
Motif: Fix scrollbar width after splitting

qsmodo:
The bug is observable if splitright is on and a non-default -scrollbarwidth is passed:

    ./vim --clean -g -sw 5 +'se splitright' +vs

Without the fix, the right window scrollbar is much more than 5px wide, although it would change into the correct 5px width once you interact with the window.

An immediate q: will make the left scrollbar too big as well.

### DIFF
--- a/src/gui_motif.c
+++ b/src/gui_motif.c
@@ -1801,12 +1801,14 @@ gui_mch_create_scrollbar(
 	    XtSetArg(args[n], XmNtopAttachment, XmATTACH_FORM); n++;
 	    XtSetArg(args[n], XmNbottomAttachment, XmATTACH_OPPOSITE_FORM); n++;
 	    XtSetArg(args[n], XmNleftAttachment, XmATTACH_FORM); n++;
+	    XtSetArg(args[n], XmNwidth, gui.scrollbar_width); n++;
 	    break;
 
 	case SBAR_RIGHT:
 	    XtSetArg(args[n], XmNtopAttachment, XmATTACH_FORM); n++;
 	    XtSetArg(args[n], XmNbottomAttachment, XmATTACH_OPPOSITE_FORM); n++;
 	    XtSetArg(args[n], XmNrightAttachment, XmATTACH_FORM); n++;
+	    XtSetArg(args[n], XmNwidth, gui.scrollbar_width); n++;
 	    break;
 
 	case SBAR_BOTTOM:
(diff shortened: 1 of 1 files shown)


brammool:
> The bug is observable if splitright is on and a non-default
> -scrollbarwidth is passed:
> 
>     ./vim --clean -g -sw 5 +'se splitright' +vs
> 
> Without the fix, the right window scrollbar is much more than 5px
> wide, although it would change into the correct 5px width once you
> interact with the window.

I cannot reproduce the problem.  The width of 5 does not result in a
usable scrollbar, the thumb is not visible.  I also tried a value of 10.

Why would it require to add an argument for the width when using a
vertical scrollbar, but not the height when using a horizontal
scrollbar?  Did you not try this, or is the Motif library asymmetric?

-- 
Common sense is what tells you that the world is flat.

qsmodo:
> I cannot reproduce the problem.  The width of 5 does not result in a
usable scrollbar, the thumb is not visible.  I also tried a value of 10.

I now realize it's only reproducible when the Gvim window is full screen. Can you reproduce it under that circumstance? 

Yes a value of 10 is fine to reproduce it, even a big value like 50.

> Why would it require to add an argument for the width when using a
vertical scrollbar, but not the height when using a horizontal
scrollbar?  Did you not try this, or is the Motif library asymmetric?

I hadn't tested it. Yes the horizontal scrollbar needs its fix too, I just added that. The Motif library is asymetric so that will be XmNheight for horizontal scrollbar.

brammool:
> > I cannot reproduce the problem.  The width of 5 does not result in a
> > usable scrollbar, the thumb is not visible.  I also tried a value of 10.
> 
> I now realize it's only reproducible when the Gvim window is full
> screen. Can you reproduce it under that circumstance? 

Yes, when the window is maximized I can reproduce.

> Yes a value of 10 is fine to reproduce it, even a big value like 50.
> 
> > Why would it require to add an argument for the width when using a
> > vertical scrollbar, but not the height when using a horizontal
> > scrollbar?  Did you not try this, or is the Motif library asymmetric?
> 
> I hadn't tested it. Yes the horizontal scrollbar needs its fix too, I
> just added that. The Motif library is asymetric so that will be
> XmNheight for horizontal scrollbar.

OK, but then also use gui.scrollbar_height for the value.

This would only be needed when the window is maximized, but I suppose it
doesn't hurt to always do this.

-- 
Amnesia is one of my favorite words, but I forgot what it means.

PR #11950, opened 2023-02-04
X11: Fix horizontal border correction for screen dimensions

qsmodo:
This prevents column(s) from being taken off upon a :vsplit (which could then be claimed by resizing the window).

I'm unsure about what this function should do relating to vertical calculations. The comment in the function implies that it is only about window decoration, but :h gdr disagrees.

I'm leaving a scaffold for vertical calculations in case it's obvious to maintainers. It can of course be removed if I'm getting it wrong or if it won't be used.

### DIFF
--- a/src/gui_x11.c
+++ b/src/gui_x11.c
@@ -1678,7 +1678,7 @@ gui_mch_set_shellsize(
 }
 
 /*
- * Allow 10 pixels for horizontal borders, 'guiheadroom' for vertical borders.
+ * Allow 'guiheadroom' for vertical borders.
  * Is there no way in X to find out how wide the borders really are?
  */
     void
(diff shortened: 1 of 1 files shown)


brammool:
> This prevents column(s) from being taken off upon a :vsplit (which
> could then be claimed by resizing the window).
> 
> I'm unsure about what this function should do relating to vertical
> calculations. The comment in the function implies that it is only
> about window decoration, but :h gdr disagrees.
> 
> I'm leaving a scaffold for vertical calculations in case it's obvious
> to maintainers. It can of course be removed if I'm getting it wrong or
> if it won't be used.

I don't think it is a good idea to change how 'guiheadroom' is used.
It's hard to predict what happens with different window managers.
The commented-out code should probably just be left out.

The main line in your patch is:

        xborder = x - attr.x > 0 ? x - attr.x : xborder;

Where "x" is the position of the root, the window that fills the screen.
I don't see how it's position can ever be smaller than that of the Vim
window.  Or how you expect to get the border width from this
computation.  Does this only work when the Vim window is maximized?  A
comment to explain this would be very helpful.


-- 
From "know your smileys":
 2B|^2B	  Message from Shakespeare

qsmodo:
> I don't see how it's position can ever be smaller than that of the Vim
window.

Me neither. It was just a sanity check, but if it's deemed unnecessary it can be removed.

> Or how you expect to get the border width from this
computation.  Does this only work when the Vim window is maximized?  A
comment to explain this would be very helpful.

Yes. only when maximized. Before I do so...:

> I don't think it is a good idea to change how 'guiheadroom' is used.
It's hard to predict what happens with different window managers.

I reckoned this could be the case, just wasn't sure about this asymmetry. Most popular desktops nowadays support vertical docks and some even default to them (Gnome).

But thinking this over, there is a better solution that I wonder how I didn't realize before: Instead of doing all these clunky calculations, just use attr.width and attr.height to retrieve the exact values? Wouldn't that be the silver bullet?

chrisbra:
closing. It sounds like we don't really need this.

PR #11960, opened 2023-02-08
Trigger incsearch on cmdline paste

k-takata:
Inserting register literally (e.g. `Ctrl-R +` or `Ctrl-R Ctrl-R "`) in commandline didn't trigger incsearch or hlsearch.

Steps to reproduce:

1. `vim --clean --cmd 'set incsearch hlsearch | let @"="text" | put'`
2. Type `/` to start searching then type `Ctrl-R Ctrl-R "`.
   Unexpectedly, `text` is not highlighted.

~~I also found that `cmdline_insert_reg()` returns only `CMDLINE_NOT_CHANGED` or `GOTO_NORMAL_MODE`. So, the checks of the returned value was redundant and the `goto cmdline_changed;` line was not executed.~~

This PR fixes them.
(I'm not sure how to add a test for this...)

### DIFF
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -2092,7 +2092,7 @@ getcmdline_int(
 		}
 		else if (res == GOTO_NORMAL_MODE)
 		    goto returncmd;
-		goto cmdline_changed;
+		goto cmdline_not_changed;
 
 	case Ctrl_D:
 		if (showmatches(&xpc, FALSE) == EXPAND_NOTHING)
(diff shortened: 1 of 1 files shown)


src/ex_getln.c
```diff
@@ -1252,7 +1254,7 @@ cmdline_insert_reg(int *gotesc UNUSED)
     redrawcmd();
 
     // The text has been stuffed, the command line didn't change yet.
```
k-takata, line 1256: This comment should be updated, but I'm not sure what is the best explanation.

brammool:
Your reproduction steps can be used for a test.  It needs to be a screendump test to see the highlighting.

PR #7022, opened 2020-09-27
Fix macOS default locale to have UTF-8 encoding set

ychin:
If LANG is unset in macOS (either manually or when launching GUI version of Vim from the Dock where this environmental variable is not set by the OS), Vim tries to query the system locale and set that. However, the system API doesn't set the encoding part, and there are tools like new versions of Python that seems to want to have an explicit encoding bit set. Also, macOS defaults to UTF-8 anyway, so just make sure to append ".UTF-8" to the locale to explicitly set it.

Also, add regression test for both this change and for 509f8031 (#7003) which added logic to make sure we set LC_NUMERIC to "C" in the default locale detection.

This is a port of https://github.com/macvim-dev/macvim/pull/1036

PR #11963, opened 2023-02-09
Fix test_environ for mac

WuerfelDev:
The environ test failed on my mac, because the regex couldn't match my :lang string `"en_US.UTF-8/UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8"`.

The second value is only `UTF-8` which fails the regex since it currently checks for `[a-zA-Z_]+\.UTF-8`

I wrapped the checks in optional groups like this: `([a-zA-Z_]+\.)?UTF-8`

<hr>

There is a different fix that seemed to work for me too:
I simply added `unset LC_CTYPE` in line 76:
https://github.com/vim/vim/blob/0caaf1e46511f7a92e036f05e6aa9d5992540117/src/testdir/test_environ.vim#L76

If that is preferred, I prepared a patch file:
[test_environ_unset.patch](https://github.com/vim/vim/files/10698199/test_environ_unset.patch)

### DIFF
--- a/src/testdir/test_environ.vim
+++ b/src/testdir/test_environ.vim
@@ -81,7 +81,7 @@ func Test_mac_locale()
   " 1. The locale is the form of <locale>.UTF-8.
   " 2. Check that fourth item (LC_NUMERIC) is properly set to "C".
   " Example match: "en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8"
-  call assert_match('"\([a-zA-Z_]\+\.UTF-8/\)\{3}C\(/[a-zA-Z_]\+\.UTF-8\)\{2}"',
+  call assert_match('"\(\([a-zA-Z_]\+\.\)\?UTF-8/\)\{3}C\(/\([a-zA-Z_]\+\.\)\?UTF-8\)\{2}"',
         \ lang_results,
         \ "Default locale should have UTF-8 encoding set, and LC_NUMERIC set to 'C'")
 endfunc


ychin:
Is there a reason why your `LC_CTYPE` locale is set to UTF-8? I looked and I don't think it's a "standard" locale, even though it seems like some tools used to set it. For example, if you do `locale -a` on a Mac, you won't see `UTF-8` in the list, even though `C`, `POSIX`, and other names of the form `<lang>.<encoding>` are.

I do see that there is a folder `/usr/share/locale/UTF-8`, but that seems more like a convenient way to symlink all the other definitions to it but it doesn't seem like a public locale.

Feel free to correct me though. I'm genuinely curious why it would have been set that way, as UTF-8 is an encoding, not a locale.

WuerfelDev:
I'm curious as well and I don't have an answer for that. My findings are:
In the Terminal settings the box at `[profile] > Advanced > Set locale environment variables on startup` is checked by default resulting in these differences

I tested and confirmed the faulty behavior with a coworker and a web search showed that the problem is affecting other people as well.


**For `locale`:**
| language | box checked | box unchecked |
| --- | --- | --- |
| English | <pre>LANG=""<br>LC_COLLATE="C"<br>LC_CTYPE="UTF-8"<br>LC_MESSAGES="C"<br>LC_MONETARY="C"<br>LC_NUMERIC="C"<br>LC_TIME="C"<br>LC_ALL=</pre> | <pre>LANG=""<br>LC_COLLATE="C"<br>LC_CTYPE="en_US.UTF-8"<br>LC_MESSAGES="C"<br>LC_MONETARY="C"<br>LC_NUMERIC="C"<br>LC_TIME="C"<br>LC_ALL=</pre> |
| German | <pre>LANG="de_DE.UTF-8"<br>LC_COLLATE="de_DE.UTF-8"<br>LC_CTYPE="de_DE.UTF-8"<br>LC_MESSAGES="de_DE.UTF-8"<br>LC_MONETARY="de_DE.UTF-8"<br>LC_NUMERIC="de_DE.UTF-8"<br>LC_TIME="de_DE.UTF-8"<br>LC_ALL=</pre> | <pre>LANG=""<br>LC_COLLATE="C"<br>LC_CTYPE="en_US.UTF-8"<br>LC_MESSAGES="C"<br>LC_MONETARY="C"<br>LC_NUMERIC="C"<br>LC_TIME="C"<br>LC_ALL=</pre> |



**For `env | grep -e ^LC_ -e ^LANG`:**
| language | box checked | box unchecked |
| --- | --- | --- |
| English | <pre>LC_CTYPE=UTF-8</pre> | <pre>LC_CTYPE=en_US.UTF-8</pre> |
| German | <pre>LANG=de_DE.UTF-8</pre> | <pre>LC_CTYPE=en_US.UTF-8</pre> |

ychin:
That is quite odd . For me (English US) ,when the box is checked, everything is `en_US.UTF-8`, and unchecked would result in everything being `C`. I wonder if there's some odd environmental settings in your configuration.

Either way, I think the proper fix is this:

```diff
diff --git a/src/testdir/test_environ.vim b/src/testdir/test_environ.vim
index d8344817f..037dad21d 100644
--- a/src/testdir/test_environ.vim
+++ b/src/testdir/test_environ.vim
@@ -73,7 +73,7 @@ func Test_mac_locale()
   " If $LANG is not set then the system locale will be used.
   " Run Vim after unsetting all the locale environmental vars, and capture the
   " output of :lang.
-  let lang_results = system("unset LANG; unset LC_MESSAGES; " ..
+  let lang_results = system("unset LANG; unset LC_MESSAGES; unset LC_COLLATE; unset LC_CTYPE; unset LC_TIME; " ..
             \ shellescape(v:progpath) ..
             \ " --clean -esX -c 'redir @a' -c 'lang' -c 'put a' -c 'print' -c 'qa!' ")
 
```

The test isn't really interested in what the tester's configuration is (in fact that's a bad thing because we want tests to be reproducible). This test was added in #7022 in order to test that the locale detection code works properly (it only triggers when locale isn't set) so probably just makes sense to unset everything in the set up to be consistent.

ychin:
Just noticed that what I suggested is your second suggestion in your PR. I personally think that's a better choice as I mentioned.

brammool:
Since the value is wrong, as several other problem reports suggest, unsetting $LC_CTYPE looks like the best solution.

PR #11968, opened 2023-02-10
Update PoE item filter syntax file

ObserverOfTime:
Added a new keyword.


brammool:
I'll include it, thanks.

PR #11974, opened 2023-02-11
The ':let &formatoptions="?"' command doesn't result in error

### LINKED ISSUE #11972: Setting 'formatoptions' to invalid value using :let fails silently
### Steps to reproduce

1. Run `vim --clean`
2. Run `:let &formatoptions = '?'`
3. No error message

### Expected behaviour

```
E539: Illegal character <?>: formatoptions=?
```

### Version of Vim

9.0.1182

### Environment

Operating system: Arch Linux
Terminal: kitty 0.27.1
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

yegappan:
Fixes #11972.

### DIFF
--- a/src/testdir/test_options.vim
+++ b/src/testdir/test_options.vim
@@ -483,6 +483,8 @@ func Test_set_errors()
   call assert_fails('set fileencoding=latin1', 'E21:')
   set modifiable&
   call assert_fails('set t_#-&', 'E522:')
+  call assert_fails('let &formatoptions = "?"', 'E539:')
+  call assert_fails('call setbufvar("", "&formatoptions", "?")', 'E539:')
 endfunc
 
 func Test_set_encoding()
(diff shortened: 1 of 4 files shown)


src/testdir/test_options.vim
```diff
@@ -1465,5 +1467,41 @@ func Test_endoffile_default()
...
+    call add(msg, v:errmsg)
+    silent! call setbufvar('', '&lines', 0)
+    call add(msg, v:errmsg)
+    call writefile(msg, 'XResultsetminlines')
+    let &lines = save_lines
+    let &columns = save_columns
+  END
+  call writefile(lines, 'XTestMinLines', 'D')
+  let cmd = $"{GetVimCommand()} --clean -S XTestMinLines -c qa!"
+  call system(cmd)
```
zeertzjq: It seems that `RunVim()` can be used here (it is also used in the test above):
```suggestion
    qa!
  END
  if !RunVim([], lines, '--clean')
    return
  endif
```

yegappan: Yes. Thanks for the suggestion. I have updated the PR with this change.

PR #11976, opened 2023-02-12
make quote text objects more intuitive

ad-chaos:
# Problem

If you have the following text with `|` indicating the cursor

    'string' |stuff 'string'

now in normal mode if you press `vi'` the region demarcated with `*` is selected

    'string'* stuff *'string'

But we can agree that `stuff` is not part of any string, but the help says that it will always select a "quoted string"

Visual mode gets some more nice (special cased) behaviours as well, taking the same example as above, if we do `vli'` or
`vhi'` we properly select the left or right string. but `di'` or `ci'` don't get just niceties.
There also exists no way to select the nth string by doing `vni'` (somewhat akin to `vnib`).

I propose the following behaviour for the `i{quote}` and `a{quote}` textobjects.

- Select the nearest "balanced" quote object (always selecting forwards but backwards if there is nothing forwards)
- Choose to select the string forwards or backwards (using `vhi'` or `vli'` for example).
- The repeat behaviour is not very obvious, I am open to suggestions but I thought the following made sense:
    - `vi'i'` includes the quotes
    - after doing `vi'i'` the next `i'` selects the next string obeying the rules of `i'`
    - `v2i'` selects the 2nd string (so selecting the nth quote object using `vni'`)
    - repeating `a'` always selects the next string

One caveat is that some languages leverage unbalanced quotes in their grammar (vimscript quotes, rust lifetimes).
Its tricky to handle those cases. My proposal is:

- An even number of quotes are always balanced.
- In case of an odd number of quotes if the user is already inside a _percieved_ balanced string, then select it.
  more precisely if the cursor is after the end of a balanced quote and there is a partial quote present after it, then
  region between the end of the balanced quote and partial quote is selected.
  (`*` marks selection `|` is cursor)
```
              v balanced quote end
    " comment "*string|*"
    ^ start           ^ start but no end
```

This also somewhat simplifies the code by not having too many special cases.

This is certainly not perfect, ideally we would want syntax based text objects, but as a default I think this is much better and an improvement over the current behaviour.

### DIFF
--- a/runtime/doc/motion.txt
+++ b/runtime/doc/motion.txt
@@ -658,26 +658,24 @@ iB			"inner Block", select [count] Blocks, from "[count] [{"
 a"							*v_aquote* *aquote*
 a'							*v_a'* *a'*
 a`							*v_a`* *a`*
-			"a quoted string".  Selects the text from the previous
-			quote until the next quote.  The 'quoteescape' option
-			is used to skip escaped quotes.
+			"a quoted string".  Selects [count]th quoted string
+			The 'quoteescape' option is used to skip escaped quotes.
 			Only works within one line.
-			When the cursor starts on a quote, Vim will figure out
-			which quote pairs form a string by searching from the
-			start of the line.
+			Vim will figure out which quote pairs form a string by
+			searching from the start of the line.
 			Any trailing white space is included, unless there is
 			none, then leading white space is included.
 			When used in Visual mode it is made characterwise.
 			Repeating this object in Visual mode another string is
-			included.  A count is currently not used.
+			included.
 
 i"							*v_iquote* *iquote*
 i'							*v_i'* *i'*
 i`							*v_i`* *i`*
-			Like a", a' and a`, but exclude the quotes and
-			repeating won't extend the Visual selection.
-			Special case: With a count of 2 the quotes are
-			included, but no extra white space as with a"/a'/a`.
+			Like a", a' and a`, but exclude the quotes.
+			If repeated the quotes are included, but no extra
+			white space as with a"/a'/a`, repeating it again will
+			select the next string.
 
 When used after an operator:
 For non-block objects:
--- a/src/testdir/test_textobjects.vim
+++ b/src/testdir/test_textobjects.vim
@@ -118,17 +118,17 @@ func Test_string_html_objects()
     let t = "'foo' 'bar' 'piep'"
     put =t
     normal! 0va'a'rx
-    call assert_equal("xxxxxxxxxxxx'piep'", getline('.'), e)
+    call assert_equal("'foo' xxxxxx'piep'", getline('.'), e)
 
     let t = "bla bla `quote` blah"
     put =t
     normal! 02f`da`
     call assert_equal("bla bla blah", getline('.'), e)
 
-    let t = 'out " in "noXno"'
+    let t = 'out '' in "noXno"'
     put =t
     normal! 0fXdi"
-    call assert_equal('out " in ""', getline('.'), e)
+    call assert_equal('out '' in ""', getline('.'), e)
 
     let t = "\"'\" 'blah' rep 'buh'"
     put =t
@@ -144,7 +144,7 @@ func Test_string_html_objects()
     let t = 'voo "nah" sdf " asdf" sdf " sdf" sd'
     put =t
     normal! $F"va"oha"i"rz
-    call assert_equal('voo "zzzzzzzzzzzzzzzzzzzzzzzzzzzzsd', getline('.'), e)
+    call assert_equal('voo "zzz" sdf " asdf" sdf " sdf" sd', getline('.'), e)
 
     let t = "-<b>asdf<i>Xasdf</i>asdf</b>-"
     put =t
@@ -541,28 +541,34 @@ func Test_textobj_quote()
 
   " Test for visually selecting an inner quote
   %d
-  " extend visual selection from one quote to the next
+  " Select the next quote object
   call setline(1, 'color "red" color "blue"')
   call cursor(1, 7)
   normal v4li"y
-  call assert_equal('"red" color "blue', @")
+  call assert_equal('blue', @")
 
-  " try to extend visual selection from one quote to a non-existing quote
+  " always select a valid quoted string
   call setline(1, 'color "red" color blue')
   call cursor(1, 7)
   call feedkeys('v4li"y', 'xt')
-  call assert_equal('"red"', @")
+  call assert_equal('red', @")
 
-  " try to extend visual selection from one quote to a next partial quote
   call setline(1, 'color "red" color "blue')
   call cursor(1, 7)
+  normal vi"y
+  call assert_equal('red', @")
+
+  " If the textobject is already selected move to the next one
+  " we allowe selecting color because in vimscript a starting quote is a
+  " comment. So there can be unbalanced quotes
+  call cursor(1, 7)
   normal v4li"y
-  call assert_equal('"red" color ', @")
+  call assert_equal(' color ', @")
 
   " select a quote backwards in visual mode
   call cursor(1, 12)
   normal vhi"y
-  call assert_equal('red" ', @")
+  call assert_equal('red', @")
   call assert_equal(8, col('.'))
 
   " select a quote backwards in visual mode from outside the quote
@@ -605,6 +611,56 @@ func Test_textobj_quote()
   call setline(1, 'char *s = "foo\\\\"bar"')
   normal $hhyi"
   call assert_equal('bar', @")
+
+  call setline(1, '" This is vimscript comment "this a string though"')
+  normal $di"
+  call assert_equal("this a string though", @")
+
+  call setline(1, "let a: &'life char = 'a';")
+  normal $di'
+  call assert_equal("a", @")
+
+  call setline(1, '"string1" stuff "string2" stuff "string3" stuff "string4" stuff')
+
+  " select on the nth string forwards using {v,d,c}[n]{i,a}{quote}
+  for i in range(1,4)
+    exe 'normal 0y'.i.'i"'
+    call assert_equal("string".i, @")
+    exe 'normal 0y'.i.'a"'
+    call assert_equal('"string'.i.'" ', @")
+  endfor
+
+  " If if the number given is more than the quoted strings present select
+  " the last one
+  normal 0y10i"
+  call assert_equal("string4", @")
+
+  " allow selecting backwards
+  normal $yi"
+  call assert_equal("string4", @")
+
+  " move through quoted string selections using a{quote}
+  normal 0v
+  for i in range(1,4)
+    normal a"y
+    call assert_equal('"string'.i.'" ', @")
+    normal gv
+  endfor
+  normal v
+
+  " doing i{quote} twice selects the qutoes
+  " but the next time it selects the next quoted string
+  normal 0v
+  for i in range(1,4)
+    normal i"y
+    call assert_equal('string'.i, @")
+    normal gv
+
+    normal i"y
+    call assert_equal('"string'.i.'"', @")
+    normal gv
+  endfor
+  normal v
 
   close!
 endfunc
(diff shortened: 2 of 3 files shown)


brammool:
> # Problem
> 
> If you have the following text with `|` indicating the cursor
> 
>     'string' |stuff 'string'
> 
> now in normal mode if you press `vi'` the region demarcated with `*` is selected
> 
>     'string'* stuff *'string'
> 
> But we can agree that `stuff` is not part of any string, but the help
> says that it will always select a "quoted string"

You cannot decide what is a string without looking in surrounding lines.
You appear to assume that strings can't cross a line break.  That is not
always the case.

I'm not sure your changes are an improvement or just different.  Looking
at the first changed test case, I would say it's actually wrong:

-    call assert_equal("xxxxxxxxxxxx'piep'", getline('.'), e)
+    call assert_equal("'foo' xxxxxx'piep'", getline('.'), e)

This changed result also looks wrong:
-    call assert_equal('voo "zzzzzzzzzzzzzzzzzzzzzzzzzzzzsd', getline('.'), e)
+    call assert_equal('voo "zzz" sdf " asdf" sdf " sdf" sd', getline('.'), e)

The existing behavior extends the Visual area, your changed behavior
appears to keep only the last selected object.  I do not see that as an
improvement.

You have also changed the input for some test cases.  That makes it
difficult to see what the changes are from previous behavior.  Instead
of changing the existing test case, please add a new one.


> - Select the nearest "balanced" quote object (always selecting
> forwards but backwards if there is nothing forwards)

I don't think you really means "nearest", that would be counting the
number of characters forward/backward.  That would be too unpredictable
and hardly ever what the user expects.

I'm not sure if selecting an object backwards is a good idea.  This also
makes it a bit unpredictable.

> - Choose to select the string forwards or backwards (using `vhi'` or
> `vli'` for example).

This makes it complicated.  The user could move the cursor before
selecting the object.  It's often better to have more than one simple
command than one complicated one.

> - The repeat behaviour is not very obvious, I am open to suggestions
> but I thought the following made sense:
>     - `vi'i'` includes the quotes
>     - after doing `vi'i'` the next `i'` selects the next string
>     obeying the rules of `i'`

Isn't this how it is already working?

>     - `v2i'` selects the 2nd string (so selecting the nth quote object
>     using `vni'`)

Generally a count means repeating the command, thus "2i'" would do the
same as "i'i'".  Doing otherwise is unexpected.  Is there a really good
reason to do something else in this case?

>     - repeating `a'` always selects the next string
> 
> One caveat is that some languages leverage unbalanced quotes in their
> grammar (vimscript quotes, rust lifetimes).
> Its tricky to handle those cases. My proposal is:
> 
> - An even number of quotes are always balanced.
> - In case of an odd number of quotes if the user is already inside a
>   _percieved_ balanced string, then select it.
>   more precisely if the cursor is after the end of a balanced quote
>   and there is a partial quote present after it, then
>   region between the end of the balanced quote and partial quote is selected.
>   (`*` marks selection `|` is cursor)
> ```
>               v balanced quote end
>     " comment "*string|*"
>     ^ start           ^ start but no end
> ```

This makes sense.  But what if the cursor is in the "comment" text, it
would select something else?  Then it isn't different from the existing
behavior, is it?

> This also somewhat simplifies the code by not having too many special cases.
> 
> This is certainly not perfect, ideally we would want syntax based text
> objects, but as a default I think this is much better and an
> improvement over the current behaviour.

I think we need to try this out on many corner cases to be able to judge
that.  Also, it's not just about selecting the right thing, but also
making sure it is predictable.  The user can move the cursor to another
location to make the selection cover the right text.

-- 
"Computers in the future may weigh no more than 1.5 tons."
                                   Popular Mechanics, 1949

ad-chaos:
> You cannot decide what is a string without looking in surrounding lines.
You appear to assume that strings can't cross a line break.  That is not
always the case.

But we say that the `i{quote}` and `a{quote}` text object only operates on a line, so that is a reasonable assumption
```help
a"							*v_aquote* *aquote*
a'							*v_a'* *a'*
a`							*v_a`* *a`*
			"a quoted string".  Selects the text from the previous
			quote until the next quote.  The 'quoteescape' option
			is used to skip escaped quotes.
			Only works within one line.
```

> The existing behavior extends the Visual area, your changed behavior
appears to keep only the last selected object.  I do not see that as an
improvement.

Personally I find selecting quoted strings to be much more helpful than extending the area to the next quoted string. I cannot think of where it would be useful of the top of my head and that's why I changed it like that.

> I'm not sure if selecting an object backwards is a good idea.  This also
makes it a bit unpredictable.

Say the case is like the following
```
"string" stuff|
```
With this patch only in such cases when unambigously when there is no quoted string forwards and there is a quoted string backwards do I chose to select it backwards, otherwise it always chooses forwards like how it does now.

> This makes it complicated.  The user could move the cursor before
selecting the object.  It's often better to have more than one simple
command than one complicated one

That's the current behaviour though. I was just listing out what the quote object should do (including existing and new additions)

> Isn't this how it is already working?

No, `vi'i'` selects the quotes but `vi'i'i'` selects the quotes and then extends the visual selection (which I think is unhelpful in most cases). With this patch it selects the next string.

> Generally a count means repeating the command, thus "2i'" would do the
same as "i'i'".  Doing otherwise is unexpected.  Is there a really good
reason to do something else in this case?

the ability for `i'` to respect the count given to it enables doing things like `c3i'` or `d3i'` (changing and deleting the 3rd string respectively). repeating `i'` is only possible in visual mode anyway like `vi'i'`. Doing `ci'i'` and `di'i'`are not supported. Currently `v2i'` selects the quotes around an object, with this patch `v2i'`, `c2i'` and `d2i'` operate on the second string. We can mention that only `i'i'` is special cased for visual mode (like it already was).

> This makes sense.  But what if the cursor is in the "comment" text, it
would select something else?  Then it isn't different from the existing
behavior, is it?

Yes, the behaviour only differs when there are an even number of quotes in a line.

> The user can move the cursor to another
location to make the selection cover the right text

And this patch enables you do to that? I am not sure I understand this comment

brammool:
> > You cannot decide what is a string without looking in surrounding lines.
> > You appear to assume that strings can't cross a line break.  That is not
> > always the case.
> 
> But we say that the `i{quote}` and `a{quote}` text object only
> operates on a line, so that is a reasonable assumption

The text object only works within one line, but that doesn't mean
the quoted strings are restricted to one line.  It does mean that it
won't work well if strings a line boundary.

> > The existing behavior extends the Visual area, your changed behavior
> > appears to keep only the last selected object.  I do not see that as an
> > improvement.
> 
> Personally I find selecting quoted strings to be much more helpful
> than extending the area to the next quoted string. I cannot think of
> where it would be useful of the top of my head and that's why I
> changed it like that.

This is how all text objects work.  And it is useful in a list of
strings, for example.

> > I'm not sure if selecting an object backwards is a good idea.  This also
> > makes it a bit unpredictable.
> 
> Say the case is like the following
> ```
> "string" stuff|
> ```
> With this patch only in such cases when unambigously when there is no
> quoted string forwards and there is a quoted string backwards do I
> chose to select it backwards, otherwise it always chooses forwards
> like how it does now.

This may seem like a good idea for the example, but if a long text
follows, the user would have to look through it for any quoted strings
before the user can predict what will happen.  I would rather see it
fail, then it's clear there is no quoted string after the cursor.  That
the user used the command it must have looked like there was one.

Anyway, you will have understood by now that I have my doubts whether
the change is an improvement.  I wonder what others are thinking.  You
will have to spend time on convincing us that it works better and
doesn't break what users are used to.

-- 
~
~
~
".signature" 4 lines, 50 characters written

ad-chaos:
> And it is useful in a list of
strings

But, here's where this new behaviour of operating on the next string is useful,
and why the trade-off of it not working nicely with list of strings for example is worth it.

Say you're on a long line:

`"string" and then a lot of text with lots of words and possibly some punctuation, but here is another "string"`

now say you're inside the first string and want to select the second one, you
know that most vim motions accept a count do you `v2i"` ah but that selects the
quotes, no problem you press `i"` again and now you have selected the entire
line (not what you wanted) , so you escape and then the cursor is inside the
correct string, you try to do `vi"` again and it has finally selected the string
you wanted.

So, naturally while editing you go `v2i"i"<esc>i"` to select the next string. If
someone is acquainted with this behaviour then they might do something along the
lines of `f";vi"` or just `f"vi"` depending on where they are. With this patch you
just do `v2i"` to select the next string and it reads like `second double quoted string` as well.

This is helpful while not only editing prose but while editing code as well, take the
following contrived example:

```py
parsed = map(lambda line: map(partial(parser, prefix="0b") , line.split()),  with_iter(open("./file.txt")).readlines())
```

and now say that you're on `line.split()` you want to change the `file.txt` to
`binary_data.txt` well you know that vim will search forwards for a "quoted string"
from past experience and proceed to do `ci"` and what happens is the following.

```py
parsed = map(lambda line: map(lambda word: "text"|"./file.txt")).readlines())
```

I hope you will agree that that is not very helpful.
With this patch the following happens when you press `ci"`.

```py
parsed = map(lambda line: map(partial(parser, prefix="0b") , line.split()), with_iter(open("|")).readlines())
```

It operated the `c` on the next balanced `i"` object. 
Which is hopefully inline with what most people expect when they tell vim
 to operate on the next quoted string using the `i"` textobject.

Say a person in some case does want to select that region, then they can do
`F"ct"` which is similar to what you would do in the above example to select the
balanced quote object (`f"ci"`) and its the same number of keystrokes as well.
In that case why not make the more useful case (of selecting balanced quotes)
more friction free by making it part of the semantics of `i"` and others.

Infact it is documented as `"a quoted string"` but as I said originally (and shown here),
it doesn't always select a quoted string.

I found the following entry in the `todo.txt`
https://github.com/vim/vim/blob/1a6476428f63e9fa0c2cbea296e475e60363af11/runtime/doc/todo.txt#L1864

I am not sure if they meant this behaviour, but perhaps this new behaviour I am proposing is more logical?
To me it is certainly more intuitive.

jottkaerr:
The issue that led to the aforementioned todo entry was probably the following mail by Ben Fritz

> I noticed two behaviors of i" and a" that surprised me.
> 
> Consider this text:
> 
>   "Here's a patch," Daniel said.
> 
> First, set selection=exclusive.
> 
> Now place the cursor on the beginning of the line and type va". This selects the whole string "Here's a patch," with the quotes and one trailing whitespace, as expected. Typing va" again should have no effect, because there are no strings to select. Instead Vim beeps and the trailing whitespace is no longer selected.
> 
> Repeating the same process with vi" instead of va" unselects the comma at the end of the string. Repeating further times toggles the comma between selected and unselected.
> 
> That was issue 1.
> 
> Now use text:
> 
>   "Thanks," Bram said. "I'll put it on the TODO list!"
> 
> Again place the cursor at the beginning of the line. va" selects the first string, "Thanks," and trailing whitespace as expected. Type va" again and both strings are now selected, which is documented. So far, so good.
> 
> Now try again with vi". This is not supposed to do anything, as the :help says:
> 
> 			Like a", a' and a`, but exclude the quotes and
> 			repeating won't extend the Visual selection.
> 
> Instead, the entire first string gets unselected, and the entire second string "I'll put it on the TODO list!" gets selected instead (with the quotes this time). vi" again unselects the trailing quote but not the leading quote. Once the trailing quote is unselected, issue 1 applies again.
> 
> That was issue 2.
> 

This seems to be different from what you try to solve/change.

Apart from that I think you are restricting your view to much on source code, where quoted text spanning multiple lines is quite seldom. But if you edit natural language and write quoted speech, you will soon have quotes spanning multiple lines where it is hard to find the opening quote without understanding the text. That's an inherent disadvantage of having identical opening and closing symbols.

In my opinion Vim's rule to act on the first pair of quotes on the current line is a good compromise. Everything else would probably result in unpredictable behavior.

ad-chaos:
> In my opinion Vim's rule to act on the first pair of quotes surrounding the cursor on the current line is a good compromise. Everything else would probably result in unpredictable behavior.

Would you mind giving this PR a go and seeing where it causes unpredictable behaviour?

chrisbra:
It seems this is not a true improvement here, So let me close it.

PR #11980, opened 2023-02-12
Prevent use of :: for dosbatch comments by default

mrdubya:
This patch is to disable (in fact highlight as an error) comments using the :: idiom in batch files as it can use problems on modern Windows systems. For anyone still working with good old MS-DOS (or equivalents) where the idiom still works fine you can set a variable in the vimrc file to support them as before.

Rolled in some miscellaneous fix-ups to the doc and syntax file.


brammool:
I'll include it, thanks.

PR #11982, opened 2023-02-13
Replace {path} with {flags} in mkdir() documentation

gdupras:
This function expects flags as its second argument, not a path.


k-takata:
This line should be also updated: https://github.com/vim/vim/blob/39b789954d78ac4ccb99b3aa4aa88ca5120b68e1/runtime/doc/builtin.txt#L385

gdupras:
Done.

brammool:
I'll include it, thanks.  I'll also adjust the remark about using an empty string.

Issue #8600, opened 2021-07-21
filename-modifer :8 has been changed between 8.0 and 8.2


Grueslayer:
Somewhere between 8.0 and 8.2 the *filename-modifier* `:8` has been changed. The docs saying it is only working on parts of the path that already exist, BUT

in 8.0 it *appends* the rest

```
echo fnamemodify('C:\windows\newfile.txt',':p:8')
```

gives `C:\windows\newfile.txt`.

in 8.2 it *removes* the rest

```
echo fnamemodify('C:\windows\newfile.txt',':p:8')
```

gives `C:\windows`.

That breaks some existing code.

Grueslayer:
@brammool Any news on this?

k-takata:
Could you find which patch caused this?

dpelle:
To @Grueslayer : does the file `C:\windows\newfile.txt` exist on your system ?
I cannot try to reproduce this bug myself as I only have Linux.

brammool:
I hope someone who regularly uses MS-Windows can look into this. The relevant code is probably near the start of src/filepath.c.
Have you actually tried different Vim versions on the same system?  It's also possible that MS-Windows changed the behavior.
Or the removal of support for old MS-Windows systems might have affected it (using the wide API instead of the old one).

chrisbra:
I can reproduce. I bisected to v8.1.2234 (3f39697b73f661d6900c7cf5430d967a129660d7). The logic looks right, not sure why this patch causes it :)

PR #11985, opened 2023-02-13
doc: Remove some items from todo.txt

k-takata:
> MS-Windows: did path modifier :p:8 stop working?  #8600

Fixed in 8.2.4535.

> Add "unicode true" to NSIS installer.  Doesn't work with Windows 95, which we
> no longer support.

Added in 8.1.0616.

> ":cd C:\Windows\System32\drivers\etc*" does not work, even though the
> directory exists. (Sergio Gallelli, 2013 Dec 29)

I think this can happen when 32-bit Vim is used on 64-bit Windows. `C:\Windows\System32` will be redirected to `C:\Windows\SysWOW64`, and it doesn't have the `drivers` directory.
We could make this work by calling `Wow64DisableWow64FsRedirection()`, but I don't think it is needed.
https://learn.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-wow64disablewow64fsredirection

32-bit applications can read the 64-bit system directory by accessing `C:\Windows\Sysnative`, but this is read-only.

> 8   Win32 GUI: With maximized window, ":set go-=r" doesn't use the space that
>     comes available. (Poucet)  It works OK on Win 98 but doesn't work on Win
>     NT 4.0.  Leaves a grey area where the scrollbar was.  ":set go+=r" also
>     doesn't work properly.

Doesn't reproduce on Windows 10.

> -   Win32 GUI: mouse wheel always scrolls rightmost window.  The events arrive
>     in Vim as if the rightmost scrollbar was used.

'scrollfocus' was added in 8.1.2257.

> 7   GvimExt: when there are several existing Vims, move the list to a submenu.
>     (Mike McCollister)

Fixed in 8.1.0492.

### DIFF
--- a/runtime/doc/todo.txt
+++ b/runtime/doc/todo.txt
@@ -385,8 +385,6 @@ IDEA: when drawing the text, store the text byte index in ScreenLinesIdx[].
 When converting screen column to text position use this.
 The line number can be obtained from win->w_lines[].
 
-MS-Windows: did path modifier :p:8 stop working?  #8600
-
 Version of getchar() that does not move the cursor - #10603 Use a separate
 argument for the new flag.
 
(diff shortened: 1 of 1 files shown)


ad-chaos:
You could perhaps remove this item as well
https://github.com/vim/vim/blob/5ceb8157bc4608eeff4b7453da31ebe65f08a862/runtime/doc/todo.txt#L1176-L1177

I believe the relevant patch would be 8.2.3255

brammool:
I'll include it, thanks.

Issue #11937, opened 2023-02-03
sh/bash highlight broken since upgrade of sh.vim version 205 


illia-bobyr:
### Steps to reproduce

Latest [upgrade of the runtime files](https://github.com/vim/vim/commit/be4e01637e71c8d5095c33b9861fd70b41476732) brought in sh.vim version 205.
Unfortunately, it seems to be breaking sh/bash syntax highlight pretty badly for me.

On the left is sh.vim version 204, on the right is sh.vim version 205:
![image](https://user-images.githubusercontent.com/694419/216522630-0ae851eb-8ea5-41ea-9e86-386398efc910.png)

I was writing a few scripts, and every time, after a second or third command, the highlight was switching into `shCmdSubRegion`, covering everything.
On the picture that happened in line 3, and line 4 and below is all highlighted green.

Here is the test case from the picture:
```bash
#!/bin/bash

for name in {1..3}; do
  echo "$name"
done
```

### Expected behaviour

Bash highlight works as before.

### Version of Vim

9.0.1275, build from the lastest source

### Environment

OS: Ubuntu 22.10
Terminal: Alacritty, build from the latest source
$TERM: alacritty
shell: fish, version 3.6.0

derekschrock:
The issue appears to be https://github.com/vim/vim/blob/master/runtime/syntax/sh.vim#L338 

The start value `\$(\ze[^(]\|$`

Mainly `\|$`

An sh command sub. starts with an EOL or `$(` followed by possible multi-lines of characters that are not `)`.  EOL doesn't seem to make sense here?  Don't know if grouping the [^)] and \|$ was expected here or if it's just a mistake.  Removing and grouping seems to fix it.

derekschrock:
grouping the non-) and EOL appears to be what is needed here?

`\$(\ze\([^(]\|$\)`

chdiza:
@cecamp

chdiza:
Even the below shows the brokenness:
```sh
export PATH
###
export PATH
#everything afer the hashes is colored as shCmdSubRegtion  for some reason
```

cecamp:
I've submitted the fixed syntax/sh.vim to Bram already. You can also get it from my [website](http://www.drchip.org/astronaut/vim/index.html#SYNTAX_VIM).

k-takata:
@brammool 
Could you include the latest version of sh.vim soon? Many people have reported this error.

brammool:
[...]

PR #11991, opened 2023-02-14
sh.vim: Fix broken highlighting for shell scopes

zdohnal:
The latest update of runtime files broke syntax highlighting of scopes in shell files.

This PR reverts the specific change in sh.vim which caused the behavior.

See especially colors in IF scopes:
[Broken highlighting](https://bugzilla-attachments.redhat.com/attachment.cgi?id=1944012)
[Working highlighting](https://bugzilla-attachments.redhat.com/attachment.cgi?id=1944014)

I'll write an email to the file maintainer as well.


k-takata:
Have you checked this? https://github.com/vim/vim/issues/11937#issuecomment-1421116531

zdohnal:
@k-takata aha, sorry for the duplicate - I'll check the fix from there.

zdohnal:
Yes, it is the same thing, closing the PR.

PR #11992, opened 2023-02-14
win32: Change the default encoding of the options to utf-8

k-takata:
Default encoding on MS-Windows has been changed to utf-8 in v8.2.2912.
However, some settings still use the current codepage.

Change the default encoding of the options (especially for 'runtimepath', 'packpath', 'helpfile' and 'viewdir') to utf-8.
This allows users to install Vim into a directory that has non-ASCII characters in its name. E.g.: `C:\ソフトウェア\Vim`

Related:
https://github.com/vim/vim-win32-installer/issues/205
https://github.com/vim-jp/issues/issues/517

### DIFF
--- a/src/misc1.c
+++ b/src/misc1.c
@@ -1311,6 +1311,32 @@ free_users(void)
 }
 #endif
 
+#if defined(MSWIN) || defined(PROTO)
+/*
+ * Initilize $VIM and $VIMRUNTIME when 'enc' is updated.
+ */
+    void
+init_vimdir(void)
+{
+    int	    mustfree;
+    char_u  *p;
+
+    mch_get_exe_name();
+
+    mustfree = FALSE;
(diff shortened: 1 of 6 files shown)


tonymec:
Hm, if the user has changed the current codepage away from the Windows default, isn't it better to use the current user-set setting as Vim always did, rather than enforce UTF-8 even if the Vim developers believe that it is the best?

Me too I believe that UTF-8 is objectively the best, but I also believe that downward compatibility and respect of users' preferences should override this preference of mine when other people are running Vim on their own system (or even in e.g. an Internet café).

Best regards,
Tony.

k-takata:
The problem is that Vim doesn't work as expected out of the box if it is installed in a directory with non-ASCII characters, because Vim now uses utf-8 as the default encoding, but some options are initialized with the current codepage.

Issue #11596, opened 2022-11-23
`gm` and `g^` do not remember column for vertical motion in folds


citizenmatt:
### Steps to reproduce

Given some text:

```
    Alice was beginning to get very tired of sitting by her sister on the bank, and
    of having nothing to do: once or twice she had peeped into the book her sister
    was reading, but it had no pictures or conversations in it, “and what is the use
    of a book,” thought Alice “without pictures or conversations?”
```

1. Create a fold of the middle two lines (`jVjzf`)
2. Position cursor on the fold
3. Type `gm`, note that `:set ruler` shows `59-1`
4. Type `k`, the cursor navigates to line 1, column 1

### Expected behaviour

When typing `gm<Up>` I expect Vim to use the current column (59) on the new line

Similarly, if using `g^`, the ruler shows `5-1` but hitting `<Up>` moves to column 1. However, `g$` shows `83-1` in the ruler and hitting `<Up>` will move to column 83 on line 1, as expected.

The column is remembered and used if typing `gm<Right>`, `g^<Right>` or `g$<Right>`. The fold is opened and the caret is correctly moved.

### Version of Vim

9.0 (2022 Jun 28, compiled Oct 19 2022 18:24:48)

### Environment

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Oct 19 2022 18:24:48)
macOS version - x86_64
Included patches: 1-800
Compiled by Homebrew

simmsbra:
I looked into this a bit but will leave the rest to someone else, since it will take me too much time to understand everything and create a thorough patch.
Here are my findings, though:
- I tested old versions and have concluded that this is not a regression.
- The functions that are bugged, like nv_g_home_m_cmd() in normal.c, [set w_set_curswant to TRUE](https://github.com/vim/vim/blob/523f22d57794a73495aa49f667d7f90cc9c8b5c6/src/normal.c#L5747), while a function that is not bugged (nv_g_dollar_cmd()) instead [calls update_curswant_force()](https://github.com/vim/vim/blob/523f22d57794a73495aa49f667d7f90cc9c8b5c6/src/normal.c#L5855). If you replace the statement that sets w_set_curswant to TRUE with a call to update_curswant_force(), the bug seems to be fixed.

brammool:
Updating w_curswant every time adds over head (quite a lot for a long line).
So, the solution could be to call update_curswant_force() if "curwin->w_cline"_folded is set.

PR #11994, opened 2023-02-14
fix for bug 11596 - gm and g^ do not remember column for vertical motion in folds

cero1988:
I changed nv_g_home_m_cmd for doing nothing in case of gm executed on a folded line
Do you think that this could be a good solution ?

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -5719,8 +5719,6 @@ nv_g_home_m_cmd(cmdarg_T *cap)
     int		i;
     int		flag = FALSE;
 
-    if (curwin->w_cline_folded) return; // in case of folded lines, exit
-
     if (cap->nchar == '^')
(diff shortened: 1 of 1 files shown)


brammool:
Although technically moving the cursor horizontally in a closed fold has no meaning, one has to wonder why the user (or a script) would use the command.  Doing nothing (ignoring the command) is not really better than executing the command, right?  Or is there a situation where it would cause trouble?

cero1988:
The problem when executing gm on a folded line is that the ruler show the new column, but when you move cursor on another row, It return to the first column, 
So I thought that doing nothing were better

brammool:
See my comment on #11596 - fixing the cursor columns should be possible.

chrisbra:
closing in favor of #15398

PR #11995, opened 2023-02-14
Fix for messages behavior depends on 'fileformat' of current buffer

cero1988:


brammool:
Please sync to head to avoid the tests failing.

src/charset.c
```diff
@@ -527,26 +527,31 @@ transchar_buf(buf_T *buf, int c)
  */
     char_u *
 transchar_byte(int c)
+{
+    return transchar_byte_buf(curbuf, c);
+}
+
+    char_u *
+transchar_byte_buf(buf_T *buf,int c)
```
k-takata:
```suggestion
transchar_byte_buf(buf_T *buf, int c)
```

src/charset.c
```diff
@@ -527,26 +527,31 @@ transchar_buf(buf_T *buf, int c)
...
 {
     if (enc_utf8 && c >= 0x80)
     {
-	transchar_nonprint(curbuf, transchar_charbuf, c);
-	return transchar_charbuf;
+        transchar_nonprint(buf, transchar_charbuf, c);
+        return transchar_charbuf;
```
k-takata:
```suggestion
	transchar_nonprint(buf, transchar_charbuf, c);
	return transchar_charbuf;
```
Tab characters should be used here.

src/charset.c
```diff
@@ -527,26 +527,31 @@ transchar_buf(buf_T *buf, int c)
...
     }
-    return transchar(c);
+    return transchar_buf(buf, c);
 }
-
 /*
  * Convert non-printable character to two or more printable characters in
- * "charbuf[]".  "charbuf" needs to be able to hold five bytes.
+ * "buf[]".  "charbuf" needs to be able to hold five bytes.
```
k-takata: I don't think this change is needed.

The following might be better?
```
 * Convert non-printable character "c" to two or more printable characters
 * in "charbuf[]".  "charbuf" needs to be able to hold five bytes.
```

src/charset.c
```diff
@@ -527,26 +527,31 @@ transchar_buf(buf_T *buf, int c)
...
  * Does NOT work for multi-byte characters, c must be <= 255.
  */
     void
 transchar_nonprint(buf_T *buf, char_u *charbuf, int c)
 {
     if (c == NL)
 	c = NUL;		// we use newline in place of a NUL
-    else if (c == CAR && get_fileformat(buf) == EOL_MAC)
+    else if (buf != NULL && c == CAR && get_fileformat(buf) == EOL_MAC)
 	c = NL;			// we use CR in place of  NL in this case
```
k-takata, line 555:
```diff
-	c = NL;			// we use CR in place of  NL in this case
+	c = NL;			// we use CR in place of NL in this case
```
This is not your change, but there is an extra space in the comment.

cero1988: thank you for the suggestions, I have commited them all

src/charset.c
```diff
@@ -527,15 +527,20 @@ transchar_buf(buf_T *buf, int c)
...
     if (enc_utf8 && c >= 0x80)
     {
-	transchar_nonprint(curbuf, transchar_charbuf, c);
+	transchar_nonprint(buf, transchar_charbuf, c);
 	return transchar_charbuf;
     }
-    return transchar(c);
+    return transchar_buf(buf, c);
 }
-
```
k-takata, line 538: This empty line should be kept.

PR #11999, opened 2023-02-15
Fix insert mode cursor position for 'splitkeep'

luukvbaal:
Problem:    Incorrect assumption that when `win_fix_cursor()` is called
            from `win_enter_ext()` we are in normal mode.
Solution:   Determine whether we are in normal mode.

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -5328,7 +5328,8 @@ win_enter_ext(win_T *wp, int flags)
     if (*p_spk == 'c')		// assume cursor position needs updating
 	changed_line_abv_curs();
     else
-	win_fix_cursor(TRUE);
+	win_fix_cursor(get_real_state()
+				    & (MODE_NORMAL|MODE_CMDLINE|MODE_TERMINAL));
 
     // Now it is OK to parse messages again, which may be needed in
     // autocommands.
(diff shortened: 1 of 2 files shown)


luukvbaal:
I looked into it. This should fix the issue described in https://github.com/vim/vim/commit/fdbd14e89207dca95549f298ee468793da852086#r100644222.

src/window.c
```diff
@@ -5328,7 +5328,7 @@ win_enter_ext(win_T *wp, int flags)
     if (*p_spk == 'c')		// assume cursor position needs updating
 	changed_line_abv_curs();
     else
-	win_fix_cursor(TRUE);
+	win_fix_cursor(get_real_state() & MODE_NORMAL);
```
zeertzjq: Does the check need to be the same as the one in `win_fix_cursor()`?
```c
    // Ensure cursor is valid when not in normal mode or when resized.
    if (!(get_real_state() & (MODE_NORMAL|MODE_CMDLINE|MODE_TERMINAL)))
	win_fix_cursor(FALSE);
    else if (resize)
	win_fix_cursor(TRUE);
```

luukvbaal: I think the edge cases covered by checking for `MODE_CMDLINE/TERMINAL` only apply when the caller is `win_fix_cursor()`. But I'm not sure, I guess it wouldn't hurt.

luukvbaal: Judging by the failing tests I guess I was wrong ;)

PR #12007, opened 2023-02-16
Screenrow motion doesn't bring cursor to start of closed fold

luukvbaal:
Problem:    Screenrow motion implements it's own code for moving up a
	    non-wrapped line, which also exhibits a problem with closed
	    folds.
Solution:   Re-use inner cursor up and down functions, fixing the
	    problem.

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -2344,20 +2344,11 @@ nv_screengo(oparg_T *oap, int dir, long dist)
 	    else
 	    {
 		// to previous line
-#ifdef FEAT_FOLDING
-		// Move to the start of a closed fold.  Don't do that when
-		// 'foldopen' contains "all": it will open in a moment.
-		if (!(fdo_flags & FDO_ALL))
-		    (void)hasFolding(curwin->w_cursor.lnum,
-						&curwin->w_cursor.lnum, NULL);
-#endif
-		if (curwin->w_cursor.lnum == 1)
+		if (!cursor_up_inner(curwin, 1))
 		{
 		    retval = FAIL;
 		    break;
 		}
-		--curwin->w_cursor.lnum;
-
 		linelen = linetabsize_str(ml_get_curline());
 		if (linelen > width1)
 		    curwin->w_curswant += (((linelen - width1 - 1) / width2)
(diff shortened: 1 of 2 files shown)


brammool:
> Problem:    Screenrow motion doesn't bring cursor to start of a closed
>             fold.
> Solution:   Adjust cursor position when screenrow motion lands on a
>             closed fold.
> 
> I think the current behavior is unexpected.

On a line that doesn't wrap it should work the same as using "k".
For that cursor_up_inner() does:

    // If we entered a fold, move to the beginning, unless in
    // Insert mode or when 'foldopen' contains "all": it will open
    // in a moment.
    if (n > 0 || !((State & MODE_INSERT) || (fdo_flags & FDO_ALL)))
	(void)hasFoldingWin(wp, lnum, &lnum, NULL, TRUE, NULL);



-- 
All true wisdom is found on T-shirts.

luukvbaal:
Yes I agree that it should. Are you saying that it should already work like that currently(it doesn't)? Or are you saying we should use the same insert mode and 'foldopen' checks before running `adjustFoldCursor()` and/or replace `adjustFoldCursor()` with `hasFoldingWin()`?

brammool:
> Yes I agree that it should. Are you saying that it should already work
> like that currently(it doesn't)? Or are you saying we should use the
> same insert mode and 'foldopen' checks before running
> `adjustFoldCursor()` and/or replace `adjustFoldCursor()` with
> `hasFoldingWin()`?

The condition for when to adjust the line number matters most (with the
count down to zero).  I thought the functions do the same thing, but you
might want to check that.  Folding behavior has some unexpected sides.

The main thing is that users expect that "gk" does the same as "k" if
the line doesn't wrap.  Same for "gj" and "j".

-- 
132. You come back and check this list every half-hour.

luukvbaal:
> The main thing is that users expect that "gk" does the same as "k" if
the line doesn't wrap.  Same for "gj" and "j".

That's the thing, I think the added test-case verifies that it doesn't. Can you confirm that? I'm not getting your intention behind your replies, sorry.

Latest commit re-uses the `cursor_up_inner()`/cursor_down_inner()` instead. This also fixes the new test-case and doesn't seem to break anything.

PR #12009, opened 2023-02-17
Add proper test for escaping of "<" in mapping completion

zeertzjq:
I've also noticed that the current behavior of mapping completion is
strange when the pattern contains characters like asterisk or backslash,
because addstar() modifies these characters. Does this need change?


brammool:
> I've also noticed that the current behavior of mapping completion is
> strange when the pattern contains characters like asterisk or backslash,
> because addstar() modifies these characters. Does this need change?

Can you give an example?

-- 
Don't drink and drive.  You might hit a bump and spill your beer.

PR #12015, opened 2023-02-18
Add a option set callback function variable to the vimoption table

yegappan:
Move the callback function for options that take no arguments to the vimoption table.  In a following
PR, move option callback functions will be moved.


brammool:
Yegappan wrote:

> Move the callback function for options that take no arguments to the
> vimoption table.  In a following
> PR, move option callback functions will be moved.

Thanks for doing this.  Getting rid of the long chain of "else if" is a
nice improvement.

Several functions that are not yet in the array take arguments.  Having
"opt_did_set_cb_T" take an argument that is a struct would be best.  Not
only makes this possible to pass arguments such as "opt_flags",
"opt_idx", "oldval" that some functions will need, it will also allow
for similar options to be handled by one function.  Using a struct
avoids having to change all the called functions when we add another
argument.


-- 
I wish there was a knob on the TV to turn up the intelligence.
There's a knob called "brightness", but it doesn't seem to work. 

PR #12020, opened 2023-02-18
Fixed html syntax for javaScript region start pattern

mmaxs:
This fixes some weird syntax highlighting effects for markdown files.


mmaxs:
Sorry. But it seems the correct pattern should be defined more precisely.

Now it does not match a simple case of  `<script>`

The idea is that if there is some text after the beginning of the tag -- `<script`, then it must necessarily begin with a space.

PR #12021, opened 2023-02-18
Fixed html syntax for javaScript region start pattern

mmaxs:
Also, this fixes some weird syntax highlighting effects for markdown files.
The idea behind the change is that `script` must be a separate keyword.

### DIFF
--- a/runtime/syntax/html.vim
+++ b/runtime/syntax/html.vim
@@ -221,7 +221,7 @@ if main_syntax != 'java' || exists("java_javascript")
   " JAVA SCRIPT
   syn include @htmlJavaScript syntax/javascript.vim
   unlet b:current_syntax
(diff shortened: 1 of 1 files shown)


dkearns:
Thanks for the report @mmaxs.  Can you provide an example of the Markdown related highlighting errors?

The fix looks OK to me.  We could tighten that up but it seems to be the general approach taken elsewhere in the file so lets go with it for the moment.

@brammool  can you please include this?

mmaxs:
> Can you provide an example of the Markdown related highlighting errors?

Here is an [exmple.md](https://github.com/vim/vim/files/10777789/exmple.md) with quoted code blocks.

brammool:
I'll include the change to html.vim and update the "Last Change" date.

brammool:
This was included, so closing it now.

PR #12026, opened 2023-02-19
syntax: link goPredefinedIdentifiers to Constant

### LINKED ISSUE #11996: runtime/syntax/go: correct goPredefinedIdentifiers
currently `goPredefinedIdentifiers` are linked to `goBoolean`, which is in turn linked to `Boolean`:

https://github.com/vim/vim/blob/fdbd14e89207dca95549f298ee468793da852086/runtime/syntax/go.vim#L136-L140

I think this is a mistake. Identifiers are not booleans. Even in the context of highlighting, I think its a mistake as well. Vim already has two builtins for Identifiers, one for dark and light:

https://github.com/vim/vim/blob/fdbd14e89207dca95549f298ee468793da852086/runtime/syntax/syncolor.vim#L37

https://github.com/vim/vim/blob/fdbd14e89207dca95549f298ee468793da852086/runtime/syntax/syncolor.vim#L48

so perhaps improved code would look like this:

~~~vim
hi def link goPredefinedIdentifiers Identifier
~~~

bhcleek:
Closes #11996

### DIFF
--- a/runtime/syntax/go.vim
+++ b/runtime/syntax/go.vim
@@ -5,7 +5,7 @@
 " go.vim: Vim syntax file for Go.
 " Language:             Go
 " Maintainer:           Billie Cleek <bhcleek@gmail.com>
-" Latest Revision:      2022-11-17
+" Latest Revision:      2023-02-19
 " License:              BSD-style. See LICENSE file in source repository.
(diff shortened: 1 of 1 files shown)


brammool:
I'll include it, thanks.

PR #12032, opened 2023-02-20
Support Python 3 stable ABI to allow mixed version interoperatbility

ychin:
Vim currently supports embedding Python for use with plugins, and the "dynamic" linking option allows the user to specify a locally installed version of Python by setting `pythonthreedll`. However, one caveat is that the Python 3 libs are not binary compatible across minor versions, and mixing versions can potentially be dangerous (e.g. let's say Vim was linked against the Python 3.10 SDK, but the user sets `pythonthreedll` to a 3.11 lib). Usually, nothing bad happens, but in theory this could lead to crashes, memory corruption, and other unpredictable behaviors. It's also difficult for the user to tell something is wrong because Vim has no way of reporting what Python 3 version Vim was linked with.

For Vim installed via a package manager, this usually isn't an issue because all the dependencies would already be figured out. For prebuilt Vim binaries like MacVim (my motivation for working on this), AppImage (Edit: seems like AppImage uses [static](https://github.com/vim/vim-appimage/blob/master/scripts/build_vim.sh) linkage so maybe not an issue but it does mean it can't use different Python versions) (Edit 2: Nvm I was wrong, AppImage uses Python 2+3 so it has to be dynamic, see issues like https://github.com/vim/vim-appimage/issues/35), and Win32 installer this could potentially be an issue as usually a single binary is distributed. This is more tricky when a new Python version is released, as there's a chicken-and-egg issue with deciding what Python version to build against and hard to keep in sync when a new Python version just drops and we have a mix of users of different Python versions, and a user just blindly upgrading to a new Python could lead to bad interactions with Vim.

Python 3 does have a solution for this problem: stable ABI / limited API (see https://docs.python.org/3/c-api/stable.html). The C SDK limits the API to a set of functions that are promised to be stable across versions. This pull request adds an ifdef config that allows us to turn it on when building Vim. Vim binaries built with this option should be safe to freely link with any Python 3 libraies without having the constraint of having to use the same minor version.

Note: Python 2 has no such concept and this doesn't change how Python 2 integration works (not that there is going to be a new version of Python 2 that would cause compatibility issues in the future anyway).

---

Technical details:
======

The stable ABI can be accessed when we compile with the Python 3 limited API (by defining `Py_LIMITED_API`). The Python 3 code (in `if_python3.c` and `if_py_both.h`) would now handle this and switch to limited API mode. Without it set, Vim will still use the full API as before so this is an opt-in change.

The main difference is that `PyType_Object` is now an opaque struct that we can't directly create "static types" out of, and we have to create type objects as "heap types" instead. This is because the struct is not stable and changes from version to version (e.g. 3.8 added a `tp_vectorcall` field to it). I had to change all the types to be allocated on the heap instead with just a pointer to them.

Other functions are also simply missing in limited API, or they are introduced too late (e.g. `PyUnicode_AsUTF8AndSize` in 3.10) to it that we need some other ways to do the same thing, so I had to abstract a few things into macros, and sometimes re-implement functions like `PyObject_NEW`.

One caveat is that in limited API, `OutputType` (used for replacing `sys.stdout`) no longer inherits from `PyStdPrinter_Type` which I don't think has any real issue other than minor differences in how they convert to a string and missing a couple functions like `mode()` and `fileno()`.

Also fixed an existing bug where `tp_basicsize` was set incorrectly for `BufferObject`, `TabListObject`, `WinListObject`.

Technically, there could be a small performance drop, as there is a little more indirection with accessing type objects, and some APIs like `PyUnicode_AsUTF8AndSize` are missing, but in practice I didn't see any difference, and any well-written Python plugin should try to avoid excessing callbacks to the `vim` module in Python anyway.

I only tested limited API mode down to Python 3.7, which seemes to compile and work fine. I haven't tried earlier Python versions.

Note: On Windows, according to the docs linked above, I think you need to set `pythonthreedll` to `python3.dll` rather than a version-specific one like `python3.9.dll`.

~~WIP~~ Done:
======

I marked this as WIP because there are a few things that I want to finish up, but wanted to gather some feedbacks on this PR first.

- **(Done)** Add way to set `Py_LIMITED_API`. Probably a configure.ac option.
- **(Done)** Add `has()` / `:version` indicator that a Python build has been built with stable ABI. I'm not sure if `:version` really need to be changed, but I'm imagining `+python3/dyn-stable`.
- **(Done)** Add documentation to explain this, and also how to use `has()` to query.
- **(Done)** Add CI coverage to exercise this. (Also need to think about Windows)
- Test out popular Vim plugins written in Python (see below, I would welcome some suggestions) to make sure they still work.
- **(Done)** *Maybe*: Add a `v:python3_version`? This can help the user tell what version of Python Vim was built against. Useful esp for non-stable-ABI mode. Can also use this to write Vim script to automatically set `pythonthreedll`.
- *Maybe*: In the old non-stable-ABI mode, throw a warning when loading a Python 3 DLL that's a different version from the one Vim expects? I may punt on this.
  - Update: Probably not. The reason for that is the [`Py_Version`](https://docs.python.org/3/c-api/apiabiversion.html#c.Py_Version) variable is only exposed in 3.11 (which was released recently). Before that it was somewhat hard to tell what version a Python library is from the C SDK unless you manually call `sys.version` in Python. Seems like not worth the trouble.

---

I only tested on macOS, down to Python 3.7, with both static and dynamic linking. If other people want to try this on Windows and Linux that would be appreciated. I ran a couple Python plugins and they seemed to work fine (e.g. Ultisnips), but I personally don't use any plugins that use Python so I may not be running into much edge cases.

Note: If you want to test this, just pull the PR and build. You can comment out the `#define Py_LIMITED_API 0x03080000` line on the top of if_python3.c to get back the full API behavior. It should pass all tests and run Python plugins just like before. One way to know you are running in limited mode is to run `:py3 print(sys.stdout)` which looks a little different.


ychin:
Just for reference, I found this thread where some developers were discussing potentially deprecating certain aspects of the stable ABI (https://discuss.python.org/t/lets-get-rid-of-the-stable-abi-but-keep-the-limited-api/18458), but seems like it would be quite controversial so it didn't go anywhere or at least won't be done any time soon. The main contention is that right now, even under limited API, the PyObject struct is not opaque and exposes inner workings of CPython (since otherwise it could make the limited API quite inefficient).

ychin:
I added a `--with-python3-stable-abi` flag to the configure script (I didn't regenerate auto/configure though) so it won't use limited API by default now, and you have to specify that during configure. Also added `:version` / `has('python_stable')` for runtime query.

I may stop for now to see if I hear any feedbacks first. The other stuff are more easily ironed out afterwards.

For `:version` output I'm currently using `+python3/dyn-stable` but now I'm wondering if there are people relying on checking the version output of Vim and explicitly searching for `+python3/dyn`. We could remove it if we want to.

For CI, seems like right now Linux builds only test non-dynamic builds, whereas Windows test dynamic Python linkage. Not sure where to slot this in. Also, for Windows, the configuration needs to be modified so that it knows to use `python3.dll` instead of something like `python311.dll` (this is just how the Python 3 stable ABI works on Windows).

brammool:
> I may stop for now to see if I hear any feedbacks first. The other
> stuff are more easily ironed out afterwards.

I hope a few users who use the Python interface can try this out.

> For `:version` output I'm currently using `+python3/dyn-stable` but
> now I'm wondering if there are people relying on checking the version
> output of Vim and explicitly searching for `+python3/dyn`. We could
> remove it if we want to.

Can we keep this short?  E.g. python3/stable or python3/abi ?
It will be good to be able to see how Vim was compiled, especially if
someone has a problem with the Python interface.

> For CI, seems like right now Linux builds only test non-dynamic
> builds, whereas Windows test dynamic Python linkage. Not sure where to
> slot this in. Also, for Windows, the configuration needs to be
> modified so that it knows to use `python3.dll` instead of something
> like `python311.dll` (this is just how the Python 3 stable ABI works
> on Windows).

It would be good to have a test in the CI.  Any change in the Python
interface should be tried with the stable ABI to make sure it doesn't
break.  This doesn't have to cover every combination of choices, perhaps
just one configuration with this feature.


-- 
Wizards had always known that the act of observation changed the thing that
was observed, and sometimes forgot that it also changed the observer too.
			Terry Pratchett  -  Interesting times

ychin:
> I hope a few users who use the Python interface can try this out.

FWIW, I tried out [ultisnips](https://github.com/SirVer/ultisnips) and [black](https://github.com/psf/black) (Python code formatter library / Vim plugin which is also implemented in Python). They both seem to work fine. Using Black to format a really big Python file also resulted in similar performance (which makes sense since most of the work is done in Python, but just sanity checking).

I would have wanted to do some testing with [Powerline](https://github.com/powerline/powerline) and [YouCompleteMe](https://github.com/ycm-core/YouCompleteMe) but the setup for those plugins are complicated enough that I didn't really want to go through with it. I personally don't foresee any issues though.

> Can we keep this short? E.g. python3/stable or python3/abi ? It will be good to be able to see how Vim was compiled, especially if someone has a problem with the Python interface.

What about something like "python3/dyn/stb" or "python3/dyn+stb"? I feel like the "/dyn" have very specific meaning. You can do `:help /dyn` and it's important for the user to know that it's dynamically linked in (since that means you need to set up the correct `pythonthreedll` setting and so on). Maybe we could use a short form for "stable" like "stb"?

> It would be good to have a test in the CI. Any change in the Python interface should be tried with the stable ABI to make sure it doesn't break. This doesn't have to cover every combination of choices, perhaps just one configuration with this feature.

I can play around with the CI matrix and make sure we have at least one that use this.

brammool:
> > Can we keep this short? E.g. python3/stable or python3/abi ? It will
> > be good to be able to see how Vim was compiled, especially if someone
> > has a problem with the Python interface.
> 
> What about something like "python3/dyn/stb" or "python3/dyn+stb"? I
> feel like the "/dyn" have very specific meaning. You can do `:help
> /dyn` and it's important for the user to know that it's dynamically
> linked in (since that means you need to set up the correct
> `pythonthreedll` setting and so on). Maybe we could use a short form
> for "stable" like "stb"?

If someone searches for +python3/dyn in the output of :version it would
match "python3/dyn/stb" as well.  Not sure if that matters.

-- 
Some of the well known MS-Windows errors:
	EMULTI		Multitasking attempted, system confused
	EKEYBOARD	Keyboard locked, try getting out of this one!
	EXPLAIN		Unexplained error, please tell us what happened
	EFUTURE		Reserved for our future mistakes

ychin:
> If someone searches for +python3/dyn in the output of :version it would match "python3/dyn/stb" as well. Not sure if that matters.

Right, that was my intention / proposal. This way someone who only wants to know if Python 3 linkage is dynamic to begin with still has a baseline, and the "stable ABI" part is more an additional detail. I personally don't care too much about this, but I do think it's nice to have the /dyn part in it.

Otherwise I just pushed a new commit with CI coverage for both Linux and Windows (it will now test both python 3 full API and stable ABI in the matrix, and we can confirm that in the "Check Version" part of the CI since it prints the `version` output). Also added a drive-by change to test other scripting languages in both dynamic and static as well (previously all matrices only exercised static linkage).

Other than that I still have `v:python3_version` to implement, if we want that feature to be able to quickly query the compiled version of Python 3 for the user to confirm (in stable ABI I plan to have it be the limited API version, aka the mininum version the user needs to have).

ychin:
Ok, I just committed support for `v:python3_version` which provides a way for the user to query what Python 3 version Vim was built against (which will be useful with or without stable ABI).

I think I have mostly implemented all the things I wanted to. I can rename `python3/dyn-stable` to whatever name we decide on as well.

If there are people watching this thread and use a lot of Python plugins please give this a try! It will make your life easier if you are using dynamic Python loading :)

eirnym:
I'd love this patch to be merged. I work with different versions of python and this would quite help to work on my set of versions.

ychin:
I have been using a few Python plugins and it seems pretty stable so far. I also installed YouCompleteMe, one of the most popular Python-based Vim plugin and it seems to work with no issue.

I did discover some issues with Python 3.7, so I fixed those up in my recent push.

Fundamentally, Python 3.7 or below has a fundamental issue that [`PyIter_Check`](https://docs.python.org/3/c-api/iter.html#c.PyIter_Check) is not supported. That means we don't have an easy way to pass an iterator from Python code to Vim. My previous commit tried to work around it by making a custom implementation but it doesn't really work well, so I just replaced it to just return `FALSE` for now. It does mean building Vim with stable ABI < 3.8 is not advised as you lose this functionality. Note that we could fix it by allowing `PyIter_Check` loading to optionally fail when loading it in `py3_runtime_link_init`, so we can use that to optionally call `PyIter_Check` when running against Python 3.8 runtime libs, but don't call it when running against 3.7 runtime. I decided against implementing that for now as it adds some complexity and also seems to defeat the "stable ABI" promise.

Summary: It's best to tell people to compile stable ABI >= 3.8, but if we want to I can add support for dynamically deciding whether to call `PyIter_Check` if there's enough demand for compatibility with 3.7.

---

Other than that, is there anything else you would like to see @brammool ? I think this is the kind of thing where a lot of Python plugin users may not think too much because usually mixing versions kind of work, even though it's definitely not advised or supported by CPython explicitly and has danger of memory corruption issues.

brammool:
> I have been using a few Python plugins and it seems pretty stable so
> far. I also installed YouCompleteMe, one of the most popular
> Python-based Vim plugin and it seems to work with no issue.
> 
> I did discover some issues with Python 3.7, so I fixed those up in my
> recent push.
> 
> Fundamentally, Python 3.7 or below has a fundamental issue that
> [`PyIter_Check`](https://docs.python.org/3/c-api/iter.html#c.PyIter_Check)
> is not supported. That means we don't have an easy way to pass an
> iterator from Python code to Vim. My previous commit tried to work
> around it by making a custom implementation but it doesn't really work
> well, so I just replaced it to just return `FALSE` for now. It does mean
> building Vim with stable ABI < 3.8 is not advised as you lose this
> functionality. Note that we could fix it by allowing `PyIter_Check`
> loading to optionally fail when loading it in `py3_runtime_link_init`,
> so we can use that to optionally call `PyIter_Check` when running
> against Python 3.8 runtime libs, but don't call it when running against
> 3.7 runtime. I decided against implementing that for now as it adds some
> complexity and also seems to defeat the "stable ABI" promise.
> 
> Summary: It's best to tell people to compile stable ABI >= 3.8, but if
> we want to I can add support for dynamically deciding whether to call
> `PyIter_Check` if there's enough demand for compatibility with 3.7.

Python 3.7 is getting old.  The "End of support" date is in three
months.  There should be no need to support it for something new, but it
would be good that an error is given when someone tries to build
something that won't work.

> Other than that, is there anything else you would like to see @brammool
> ? I think this is the kind of thing where a lot of Python plugin users
> may not think too much because usually mixing versions kind of work,
> even though it's definitely not advised or supported by CPython
> explicitly and has danger of memory corruption issues.

I have no specific opinion, I would like to hear from those actually
using the Python interface, especially plugin authors. 

Python is a popuplar language, the main reason it has not gained more
popularity  for Vim plugins appears to be that using it as an
embedded language isn't well supported.  I hope that this PR improves on
this.

-- 
Some say the world will end in fire; some say in segfaults.
I say it will end in a curly bracket.

eirnym:
As a Python developer I use vim/MacVim with Python all the time. I use YCM and few other plugins to support my editing capabilities in Vim. 

I use a lot of native libraries such as asyncpg, OpenCV, pandas and other scientific libraries. I use virtual environments to enclose my libraries for a project/set of projects and often edit for different version of python for different projects. Why I can't just install binary packages? Because of few binary dependencies shared in company's PyPi server we have no access to source code for. 

@brammool I generally agree with you for new projects, but in some parts of my company's software we use Python 3.7 and I have to comply to it as well. 

And it's a "Russian roulette" if it comes to having no errors on other versions of Python than Vim is compiled with. 

My current solution which works with MacVim is to have a few versions installed simultaneously compiled for different Python versions if I want to have a guaranteed full compliance. This includes several .vim folders for them as well.

ychin:
> My current solution which works with MacVim is to have a few versions installed simultaneously compiled for different Python versions if I want to have a guaranteed full compliance. This includes several .vim folders for them as well.

Why couldn't you just use an updated Python (e.g. 3.10) just for Vim usage? Even if you have internal software that relies on 3.7 and whatnot, Vim doesn't need to really interface with them if you are using YCM and whatnot. Vim doesn't need to use the exact installation of Python as your other usages.

As for 3.7, it just doesn't support the iterator APIs that Vim is using right now. I think having subtle behavior changes is a little annoying so I'm leaning towards just not supporting it for stable ABI.

yegappan:
Can you update and rebase this PR to the latest?

ychin:
Ok done. Pushed and resolved conflicts.

ychin:
Thanks for merging! I'm willing to answer any questions if people have questions on how to configure this for their Vim builds.

ychin:
One thing I noticed is that this results in a pretty giant commit message. I personally don't mind that as I like descriptive commits, but @chrisbra just wondering if you would have preferred if contributor polish up and squash their commits first? Or did you grab the description from the pull request description (which is where I kept my updated notes)? I know you are still probably working on the process but I think if there is a clear expectation it would be easy to follow.

chrisbra:
I would have hated to see those nice commit messages get lost. However at the same time, I did want to squash into a single commit, so that I can increase the patch number. So I thought squash it and merge the commit message (removing things which got resolved). 

I am not sure this is the best way to do it, but at the same time, I wanted to keep your reasoning.

ychin:
Yeah that's totally fine. I was just wondering where you got the final commit message from so I would know where to perform my edits in the future (if I have a typo, and whatnot, since Git commits are kind of forever 😅).

PR #12047, opened 2023-02-22
Fix comparison

j-mattsson:


brammool:
Can you think of a way to test this?  Since you found this mistake, you
have some Vim script that depends on the fix?

-- 
Did you ever see a "Hit any key to continue" message in a music piece?

Issue #12003, opened 2023-02-16
`v:last_register` with name of latest actually used register


echasnovski:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
At the moment there seems to be no way for plugin authors to know which register user used for most recent operation.

*Example*. Let's say that plugin needs to know a region resulted from user's latest put operation. Start and end positions can be taken from `` `[ `` and `` `] `` marks (although, they are not only for put operations). However, there seems to be no way to reliably get region mode (charwise, linewise, blockwise) which is stored in operation register. The closest approach to get this information is to use `getregtype` and `v:register` and hope that put operation was done without manually supplying register.

To put this into actual steps:
- Create lines `aaa` and `bbb`.
- Put cursor on first `a`.
- Type `<C-v>j"ay` (yank 2x1 block in register `a`) and `yy`.
- Now user can put either with `p` (linewise) or `"ap` (blockwise). There seems to be not way for plugin to know (*after the put*) which region mode to use in order to get the pasted text.

**Describe the solution you'd like**
The `v:last_register` (or `v:register_last`) variable with name of latest actually used register.

**Describe alternatives you've considered**
Remapping all put keys with expression mapping which saves `v:register` when it is yet relevant. This works, but it is far from ideal solution, at least because there are many put keys to remap.

brammool:
I can see that this is useful.  Perhaps the command used would be needed as well?

echasnovski:
> I can see that this is useful. Perhaps the command used would be needed as well?

Ideally, yes. This, for example, might help to filter out only pasted text (and not yanked or changed) by checking if command is one of putting once (`p`, `P`, etc.).

PR #12063, opened 2023-02-25
v:last_reg global variable as feature request #12003

cero1988:
added global variable v:last_reg with the latested putted register as feature request #12003


brammool:
We should have a more generic mechanism for commands previously executed.  The operator used is also relevant.  Making a list of them will also be useful.

I sometimes type something different from what I intended, it would then be good to be able to see what happened, the last few commands executed.  At least for changes, possibly also for other commands.  E.g. if you wonder why a window was closed.

cero1988:
you mean like :history, but with the post operator commands used ?

brammool:
> you mean like :history, but with the post operator commands used ?

The :history command only keeps command lines, as text.
That's probably the best we can do for Ex commands, since they are
largely unstructured.

For Normal mode commands we do have some structure, such as the count,
the register, movement command, operator, etc.  I'm not sure what would
be useful to provide.  Alternatively a list of commands as text could be
provided.  For operators that's often what is in the redo buffer.  But
it may get complicated, e.g. when using the Visual area.

-- 
What is the difference between a professional and an amateur?
The ark was built by an amateur; professionals gave us the Titanic.

cero1988:
I could create a new history like :history nvcmd with all the last normal mode command used, for the non printable keys i could translate the more commonly used like the Arrow keys in text

chrisbra:
I am closing this one for inactivity. If you ever create a new :history command for normal mode commands, please create a new PR then. thanks.

PR #8033, opened 2021-03-29
Underscores in numeric literals.

zvezdan:
Since Python 3.6 underscores are allowed in numeric literals.
The underscore cannot be used at the beginning or end, nor can it be
duplicated. The change follows exactly the specification in Python
documentation referenced in the comment above it.

PR #12066, opened 2023-02-26
Add Python 2 syntax file.

zvezdan:
This is a copy of runtime/syntax/python.vim last commit that still supported Python 2. It also supports Python 3, up to 3.5, because this version is well tested in previous Vim releases. This ensures better support for straddling code (Python 2 and 3 compatible).

The use of this older Python 2/3.5 syntax can be enforced by setting "let python_use_python2_syntax = 1".

The work on runtime/syntax/python.vim after this commit was in support of Python 3 only. This allows us to move forward with other changes of that syntax file.

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -2791,17 +2791,25 @@ For highlighted doctests and code inside: >
 	:let python_no_doctest_highlight = 1
 or >
 	:let python_no_doctest_code_highlight = 1
-(first option implies second one).
+The first option implies the second one.
 
 For highlighted trailing whitespace and mix of spaces and tabs: >
 	:let python_space_error_highlight = 1
 
-If you want all possible Python highlighting (the same as setting the
-preceding last option and unsetting all other ones): >
+If you want all possible Python highlighting:
 	:let python_highlight_all = 1
+This has the same effect as setting python_space_error_highlight and
+unsetting all the other ones.
 
-Note: Only existence of these options matter, not their value. You can replace
-      1 above with anything.
+If you use Python 2 or straddling code (Python 2 and 3 compatible),
+you can enforce the use of an older syntax file with support for
+Python 2 and up to Python 3.5.
+	: let python_use_python2_syntax = 1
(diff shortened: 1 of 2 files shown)


brammool:
Zvezdan Petkovic wrote:

> This is a copy of runtime/syntax/python.vim last commit that still
> supported Python 2. It also supports Python 3, up to 3.5, because this
> version is well tested in previous Vim releases. It was not worth
> splitting Python 2 only syntax and risk bugs.
> 
> The work on runtime/syntax/python.vim after this commit was in support
> of Python 3 only. This allows us to move forward with other changes of
> that syntax file.

OK, we can have a Python 2 specific syntax file.  The question is: How
do users use this?  Setting 'filetype' to "python2" is not a good idea,
because the indent and ftplugin files would not be loaded.

It would be possible to set 'syntax' to "python2", while 'filetype' is
"python".  But currently the user would have to do that manually every
time.

There currently already are a few variables to tweak the Python syntax,
see ":help python-syntax".  It probably is the simplest if we add a
variable to use the Python 2 syntax, e.g.

	:let python_use_python2_syntax = 1

Then the python.vim syntax file can load the python2.vim syntax file.
How about that?

-- 

zvezdan:
@brammool Thank you for the suggestion!
I added an option to load Python 2 syntax in python.vim file and tested that it works when set.
The help file is also updated.
Once this is committed, I'll rebase the other change (PR #8033) and then we can proceed with that and other Python 3 updates.

brammool:
> @brammool Thank you for the suggestion!
> I added an option to load Python 2 syntax in python.vim file and
> tested that it works when set.
> The help file is also updated.
> Once this is committed, I'll rebase the other change (PR #8033) and
> then we can proceed with that and other Python 3 updates.

Thanks.  I'll include this now, we can further tune it over time.

I think it's fine to use the filetype "python" for Python 3, since that
is now the main version.  And allow the user to add tweaks to get Python
2 support if they want it.

-- 
188. You purchase a laptop so you can surf while sitting on the can.

brammool:
This has been included.

PR #12069, opened 2023-02-26
use indexof() to optimize and simplify runtime files

lacygoill:
In the runtime files, we sometimes use hard to read and costly combinations of `synstack()` + `map()` + other function calls.  `indexof()` lets us simplify the code (by removing 1 or 2 function calls) and make it a bit faster.

I've run the indent tests, and they still pass.


brammool:
I'll include it, thanks.

PR #12079, opened 2023-02-27
Add comment support to netrc file syntax

keith:
The netrc format definition doesn't seem to officially support comments, but both curl and python's netrc parser implementations respect a leading octothorp as a comment character. This adds support for highlighting those as such.


keith:
this is kind of a weird one since some parsers will support this and some will not, feel free to close if you think that means it shouldn't be supported!

brammool:
These files do not currently have a maintainer.  Let me just include this.  If someone doesn't like it we'll find out.

keith:
🙏🏻 thanks!

Issue #12089, opened 2023-03-02
vim9class: Calling a base class method through an extending class fails


errael:
### Steps to reproduce

Included two scripts, might be the same problem, but can open separate issue if needed.

Source the following script
```
vim9script

class Base
    def One()
        echo 'Base-One'
    enddef
    def Two()
        echo 'Base-Two'
    enddef
endclass

class Child extends Base
    def Two()
        echo 'Two'
    enddef
endclass

var obj: Base

obj = Child.new()
# this works
obj.One()
obj.Two()

#this fails
def F()
    obj.One()
    obj.Two()
enddef
F()
```
Observe in output that wrong method is called in the function. Experimentation seems to indicate that a method on the base class is **never** called. This work OK at script level.
```
Base-One
Two
Two
Two
```

Source the following script
NOTE: this case fixed by patch 9.0.1703: Vim9 Calling a method in an extended class fails
```
vim9script

class Base
    def One()
        echo 'Base-One'
    enddef
    def Two()
        echo 'Base-Two'
    enddef
endclass

class Child extends Base
    def Two()
        echo 'Two'
    enddef
endclass

var obj: Base

obj = Base.new()
obj.One()
obj.Two()

def F()
    obj.One()
    obj.Two()
enddef
F()
```
Observe the following output, confusing; . Note that script level works as expected
```
Base-One
Base-Two
Error detected while processing command line..script /home/err/bugs/vim/wrong-method-called/wrong_me
thod-2.vim[28]..function <SNR>31_F:
line    1:
class Base not found on interface Base
Base-One
line    2:
class Base not found on interface Base
Base-One
```

### Expected behaviour

The correct method is called.

Notice that in the second script, if the Child class is commented out, it works as expected; no processing errors and correct output.

### Version of Vim

9.0.1362

### Environment

ubuntu

errael:
This issue has two related examples.

The second example is fixed by https://github.com/vim/vim/commit/74cc13cc402fa4df9033fbbc0643b1b403525950: patch 9.0.1703: Vim9 Calling a method in an extended class fails

The first, Calling a base class method through an extending class fails, still doesn't work. I find it fascinating that it works from script level, but fails in a :def. Changing the title to pinpoint the failure.

chrisbra:
thanks @errael , can you please verify again (I have pushed a couple of vim9 class related patches) anf it the issue still exists, please create a new ticket? Thanks.

errael:
[...]

PR #12096, opened 2023-03-02
Fix: SEGV using null_object for method call

errael:
Solves the SEGV. Gives "internal error"; placeholder depending of actual fix.
May be related to #12089.


brammool:
> Solves the SEGV. Gives "internal error"; placeholder depending of actual fix.

The actual fix is a bit complicated, requires handling a variable of
unknown type at runtime, since the variable type could be "any".  Might
be a dict or an object.  I'll add a specific error message for now.

-- 
From the classified section of a city newspaper:
Dog for sale: eats anything and is fond of children.

PR #12102, opened 2023-03-04
Improve Simplified Chinese translation.

ghost:
Yesterday, I sent an email to vim_dev mailist,I said:

> Hello,VIM developers.
> In Chinese, "Children" can be translated not only as “儿童(ertong)", but also as “孩子(haizi)”.
> The word "孩子" is a colloquial usage in Chinese. I think that compared with the original translation(“儿童”), it gives people a more intimate feeling and is more appealing.
> I think the corrected translation is good for charity.
> I hope you can adopt my suggestion, thank you.
> The new translation my suggested is “帮助可怜的乌干达孩子！”.
> 
> 
> Luke Na 
> 03/03/2023

After that, on the mailing list, I received a reply saying that I could create a PR to fix the problem, so I created this PR.

### DIFF
--- a/src/po/zh_CN.UTF-8.po
+++ b/src/po/zh_CN.UTF-8.po
@@ -8,6 +8,7 @@
 # TRANSLATORS
 #   lilydjwg <lilydjwg@gmail.com>
 #   Ada (Haowen) Yu <me@yuhaowen.com>
+#   Luke Luo <njlyf2011@hotmail.com>
 #
 # Original translations.
 #   Edyfox <edyfox@gmail.com>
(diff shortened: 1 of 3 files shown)


chrisbra:
ping @lilydjwg

lilydjwg:
But in such fundraising projects "儿童" is commonly used, e.g. on [this page](https://www.ccafc.org.cn/index.php), there are 346 ocrrurences of "儿童" but only 15 of "孩子". In my opinion using "儿童" makes the intension clearer.

njlyf2011:
But the phrase in the software is a slogan, not an explanatory document, and slogans should be affectionate, and I think a colloquial expression would be more in keeping with the original meaning.

lilydjwg:
It's commonly used as title for materials presented to potential donators (ordinary people) too. Does that count?

njlyf2011:
If so, I think this PR should be left to the developers to decide whether it gets merged or not.

brammool:
> If so, I think this PR should be left to the developers to decide
> whether it gets merged or not.

Since it appears either way it is a valid translation, and which one to
pick is a matter of preference, we'll leave it to the maintainer to
decide.  The maintainer is expected to oversee the whole of the
translations and make sure it's consistent.  Only picking one word and
arguing over how it's translated takes it out of context.  Of course
it's fine to argue with the maintainer why you think a certain
translation works better.

-- 
234. You started college as a chemistry major, and walk out four years
     later as an Internet provider.

PR #12112, opened 2023-03-05
fixed netrw copying files in the same directory (trailing slash issue)

tsheltonprojects:
Addressing these issues:

- https://groups.google.com/g/vim_use/c/6yqU3RX2CWA
- https://vi.stackexchange.com/questions/17450/copy-a-file-in-netrw

TLDR when you press `mt`, `mf`, then `mc` it tries to allow you to copy to a newly named file in the same dir it but it actually doesn't work.

The problem is `s:netrwmftgt` has no trailing slash, and `b:netrw_curdir` does,  for example `/home/me/dir` and `/home/me/dir/` so when it compares the two it thinks they are different directories. So no new filename prompt. Additionally the error message doesn't show.  `g:netrw_keepdir` defaults to `1`, meaning it should check if it IS indeed `1`, the original code checked if it was not `1`.

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -7124,7 +7124,7 @@ fun! s:NetrwMarkFileCopy(islocal,...)
    endif
 
    " copy marked files while within the same directory (ie. allow renaming)
-   if simplify(s:netrwmftgt) == simplify(b:netrw_curdir)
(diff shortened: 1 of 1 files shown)


chrisbra:
ping @cecamp

yegappan:
@cecamp Any updates on this?

chrisbra:
thanks, I included it

PR #12114, opened 2023-03-06
feat: more highlights for the popup menu

mg979:
Schermata 2023-03-06 alle 11 02 17

This is how it looks with the `habamax` color scheme and these highlight groups:
```vim
hi Pmenu          ctermbg=235
hi PmenuSel       ctermfg=234 ctermbg=144
hi PmenuKind      ctermfg=1 ctermbg=235
hi PmenuKindSel   ctermfg=1 ctermbg=144
hi PmenuExtra     ctermfg=243 ctermbg=235 cterm=NONE
hi PmenuExtraSel  ctermfg=0 ctermbg=144 cterm=NONE
```

The file edited is `src/eval.c`, and `omnifunc` is set to `ccomplete#Complete`.

The problem is that at the moment it doesn't have a proper default, so unless you define those groups, they will use the groups for the default color scheme (and it's bad).

Those groups should be linked by default to `Pmenu` and `PmenuSel`, so that there is no difference, unless they are defined in a color scheme.


brammool:
I suppose it's fine to support the different highlighting.

To link the groups by default, change the highlight_init_both array, instead of highlight_init_light and highlight_init_dark.
Similarly to "default link CursorLineSign SignColumn".

Please add a test.  A screendump test will be the simplest to check that the right attributes are used.

habamax:
Should new hl groups also be added to syntax.txt?

image

mg979:
@habamax yes

Btw do you have any idea on how to test the highlight used inside the popup menu? I was writing a test like
```vim
" Test extra highlight groups for popup menu
func Test_pum_highlights()
  CheckScreendump
  let lines =<< trim END
    setlocal omnifunc=ccomplete#Complete
    call feedkeys("istat\<C-X>\<C-o>", "x")
  END
  call writefile(lines, 'Xscript', 'D')
  let buf = RunVimInTerminal('-S Xscript', {})

  " wait for the script to finish
  call TermWait(buf, 100)

  " Test that extra popup highlight groups are used
  call VerifyScreenDump(buf, 'Test_popup_highlights_01', {})
  call WaitForAssert({-> assert_true('*** WHAT HERE? ***', Screenline(4))})

  call StopVimInTerminal(buf)
endfunc
```

But I don't know how to proceed.

habamax:
Is there a simple way to simulate Pmenu you have on screenshot @mg979 ?
We would need it in vim/colorschemes repo (I guess if test would be added we can grab it from there)?

mg979:
@habamax

Set cwd to `src`.
Source this:
```vim
color habamax
hi PmenuKind      ctermfg=1 ctermbg=235
hi PmenuKindSel   ctermfg=1 ctermbg=144
hi PmenuExtra     ctermfg=243 ctermbg=235 cterm=NONE
hi PmenuExtraSel  ctermfg=0 ctermbg=144 cterm=NONE
```

Then do
```vim
new
setlocal omnifunc=ccomplete#Complete
```

Write something and test omnifunc.

mg979:
@habamax did you manage? Any ideas for the test?

habamax:
@mg979 yes, thank you, I was able to get that Pmenu.

As for test, I've never done screendump tests but the process described in `:h terminal-screendump` and there are examples in test dir, like https://github.com/vim/vim/blob/master/src/testdir/test_popupwin.vim#L3735-L3735

mg979:
Thanks I'll check that.

mg979:
The tests succeed locally but not in CI. There it says `pattern not found` when completing, while locally it finds the matches. In the test I'm using the same `ccomplete#Complete` that should find matches in the `src` directory, but the CI doesn't know about that?

habamax:
I think using `ccomplete#Complete` on a vim src wouldn't be great for a unit test as it is not deterministic (I might be wrong though).

The best thing would be to "construct" simplified Pmenu with all Kind* and Extra* using vimscript and test it. If it is possible.

mg979:
I'll try with that, thanks.

src/testdir/test_popup.vim
```diff
@@ -1250,4 +1250,49 @@ func Test_pum_scrollbar()
...
+  call TermWait(buf)
+  call term_sendkeys(buf, "istat\<C-X>\<C-o>")
+  call TermWait(buf, 50)
+  call VerifyScreenDump(buf, 'Test_pum_highlights_01', {})
+  call term_sendkeys(buf, "\<C-E>\<Esc>u")
+  call TermWait(buf)
+  call StopVimInTerminal(buf)
+endfunc
+
+
```
zeertzjq: Extra blank line
```suggestion

```

mg979:
@zeertzjq thanks

src/testdir/test_popup.vim
```diff
@@ -1250,4 +1250,70 @@ func Test_pum_scrollbar()
...
+      endif
+      return {
+            \ 'words': [
+            \ { 'word': 'aword1', 'menu': 'extra text 1', 'kind': 'W', },
+            \ { 'word': 'aword2', 'menu': 'extra text 2', 'kind': 'W', },
+            \ { 'word': 'aword3', 'menu': 'extra text 3', 'kind': 'W', },
+            \]}
+    endfunc
+    lcd ../..
+    source runtime/autoload/ccomplete.vim
```
zeertzjq: No longer needed
```suggestion
```

src/testdir/test_popup.vim
```diff
@@ -1250,4 +1250,70 @@ func Test_pum_scrollbar()
...
```
zeertzjq: No longer needed
```suggestion
```

mg979:
fixed

habamax:
Have just tried, looks good so far:

image

mg979:
For me it should be ok.

PR #12120, opened 2023-03-08
Add a macro to iterate over all the grow array items

yegappan:


brammool:
Yegappan wrote:

>  * Add a macro to iterate over all the grow array items

Hmm, I'm not sure this is really an improvement.  There isn't any
complexity that the macro takes care of, and it shortens the code only a
little bit.  A disadvantage is that it obscures what the for loop really
does.  And inlining the declaration of the loop variable would require
another macro.

What do others think?

-- 
250. You've given up the search for the "perfect woman" and instead,
     sit in front of the PC until you're just too tired to care.

vim-ml:
Hi Bram,

On Wed, Mar 8, 2023 at 1:59 PM Bram Moolenaar ***@***.***>
wrote:

>
> Yegappan wrote:
>
> > * Add a macro to iterate over all the grow array items
>
> Hmm, I'm not sure this is really an improvement. There isn't any
> complexity that the macro takes care of, and it shortens the code only a
> little bit. A disadvantage is that it obscures what the for loop really
>

The macro makes it explicit that we are looping over all the items in a
grow array.
So it makes the intent of the loop more apparent (and it is used in many
places).


> does. And inlining the declaration of the loop variable would require
> another macro.
>

Yes.  With this macro, we cannot inline the loop variable declaration.

Regards,
Yegappan


>
> What do others think?
>
>
>

chrisbra:
@yegappan do you still think this is good to include?

chrisbra:
closing because seems to have been staled

PR #12123, opened 2023-03-08
Make the "Diff_with_Vim" use vertical splits

chrisbra:
I find it more naturally to view diffs in vertical splits, so please consider this change.


tonymec:
Aren't the default? I see the following under the help for *start-vimdiff*:
```
What happens is that Vim opens a window for each of the files.  This is like
using the |-O| argument.  This uses vertical splits.
```
However, horizontal splits will become the default if your `'diffopt'` option includes "horizontal".

chrisbra:
oh you are right, I must have configured that in my vimrc. And I was so annoyed 🤦

PR #12129, opened 2023-03-10
For profiling, use CLOCK_MONOTONIC if available.

errael:
NOTE: profiling with `CLOCK_MONOTONIC` is enabled based on
`HAVE_TIMER_CREATE`; that is probably sufficient but a more specific test
could be added to `configure.ac`.

This PR modifies profiling to use `CLOCK_MONOTONIC`, nsec resolution, when
available. Current profiling uses `clockgettime()`, usec resolution, which
isn't suitable for micro benchmarking, where many lines of vimscript are
around a micro-second or less. The initial results show high accuracy as
indicated by the low standard deviation. The results suggest that, when
using CLOCK_MONOTONIC, meaningful data could be obtained from vim9script VM
instruction level profiling (something to consider in the future).

Below are two sets of profiling results for the given vim9 function. The
first uses `CLOCK_MONOTONIC`, the second `gettimeofday()`. A column reports
the function execution `nKeys * 10` times; and there are seven runs. Each
script line has three rows of output

- the usec per/script line, the fastest run as recommended
- average value
- standard deviation

There were occasional outliers and the tabulator was modified to throw out
the slowest result before computing the standard deviation. This got rid of
almost all high SD values in the NEW results, but the OLD results are still
messy. I'm curious about what the occasional interrupt/pause/whatever might
be.

```
def A_play2(_d: dict<list<number>>, _k: string, _v: list<number>): number
    var entry = 1               ###-0
    NullFunction(4)             ###-1
    var k = _k                  ###-2
    var d = _d                  ###-3
    d = {x: []}                 ###-4
    var e = {x: []}             ###-5
    var vdict = {[_k]: _v}      ###-6
    d->extend(vdict, 'keep')    ###-8
    var ret = 2                 ###-9
    return 1
enddef
```

=== NEW - CLOCK_MONOTONIC - toss out largest value for standard deviation

```
$ ./summarize | ./keys_algo_table 
     10     30     65    100    200    300 : nKeys (micro-sec/algorithm)
  0.023  0.024  0.023  0.023  0.023  0.023 : var entry = 1   ###-0
  0.024  0.024  0.023  0.023  0.023  0.025 :     average
  2.94%  0.30%  1.66%  0.92%  1.51%  9.40% :     std dev as % of avg
  0.420  1.675  3.172  4.158  5.556  6.989 : NullFunction(4)   ###-1
  1.002  2.145  3.530  4.758  5.965  7.415 :     average
 41.68% 16.66% 11.22%  7.31%  5.90%  5.27% :     std dev as % of avg
  0.063  0.064  0.103  0.107  0.107  0.098 : var k = _k   ###-2
  0.064  0.083  0.107  0.110  0.110  0.110 :     average
  2.06% 25.02%  3.10%  4.94%  1.60%  5.81% :     std dev as % of avg
  0.104  0.163  0.298  0.370  0.908  1.173 : var d = _d   ###-3
  0.106  0.163  0.323  0.388  0.922  1.213 :     average
  1.32%  0.23%  7.57%  3.45%  1.70%  1.94% :     std dev as % of avg
  0.166  0.166  0.158  0.161  0.179  0.231 : d = {x: []}   ###-4
  0.169  0.167  0.166  0.167  0.186  0.244 :     average
  1.78%  0.49%  5.84%  4.94%  2.53%  3.22% :     std dev as % of avg
  0.128  0.128  0.127  0.126  0.131  0.139 : var e = {x: []}   ###-5
  0.133  0.131  0.129  0.136  0.136  0.145 :     average
  2.18%  1.22%  1.76%  7.95%  3.17%  2.29% :     std dev as % of avg
  0.162  0.166  0.155  0.163  0.169  0.172 : var vdict = {[_k]: _v}   ###-6
  0.164  0.170  0.172  0.169  0.172  0.177 :     average
  1.50%  2.08% 13.54%  3.84%  1.54%  2.45% :     std dev as % of avg
  0.289  0.290  0.280  0.280  0.279  0.283 : d->extend(vdict, 'keep')   ###-8
  0.293  0.294  0.294  0.290  0.283  0.294 :     average
  0.94%  1.02%  4.08%  2.89%  0.82%  2.20% :     std dev as % of avg
  0.033  0.033  0.032  0.032  0.035  0.035 : var ret = 2   ###-9
  0.033  0.033  0.032  0.032  0.035  0.037 :     average
  0.58%  0.10%  3.33%  1.59%  0.87%  8.16% :     std dev as % of avg
```

=== OLD - gettimeofday - toss out largest value for standard deviation

```
$ ./summarize | ./keys_algo_table 
     10     30     65    100    200    300 : nKeys (micro-sec/algorithm)
  0.000  0.010  0.014  0.025  0.018  0.011 : var entry = 1   ###-0
  0.022  0.023  0.026  0.028  0.023  0.024 :     average
 67.94% 45.18% 56.29%  8.32% 21.57% 28.72% :     std dev as % of avg
  0.450  1.757  3.303  4.449  5.601  7.018 : NullFunction(4)   ###-1
  0.943  2.263  3.667  4.898  6.091  7.445 :     average
 37.47% 16.53%  8.44%  6.67%  6.07%  4.97% :     std dev as % of avg
  0.020  0.040  0.072  0.080  0.052  0.025 : var k = _k   ###-2
  0.053  0.085  0.088  0.095  0.092  0.094 :     average
 45.41% 39.66% 12.60% 15.94% 28.79% 36.34% :     std dev as % of avg
  0.060  0.130  0.278  0.290  0.832  1.049 : var d = _d   ###-3
  0.092  0.150  0.305  0.353  0.855  1.064 :     average
 25.27%  8.20%  9.66% 10.12%  1.56%  1.00% :     std dev as % of avg
  0.080  0.150  0.148  0.146  0.090  0.165 : d = {x: []}   ###-4
  0.142  0.162  0.167  0.155  0.149  0.177 :     average
 28.73%  8.22% 17.35%  4.50% 23.36%  7.73% :     std dev as % of avg
  0.080  0.107  0.120  0.104  0.115  0.128 : var e = {x: []}   ###-5
  0.110  0.119  0.129  0.145  0.162  0.141 :     average
 19.07%  9.33%  6.26% 17.33% 18.00%  9.35% :     std dev as % of avg
  0.140  0.120  0.149  0.135  0.129  0.157 : var vdict: dict<list<number>> = {[_k]: _v}   ###-6
  0.162  0.141  0.172  0.151  0.156  0.169 :     average
 10.65% 13.86%  9.32% 11.10% 13.20%  4.31% :     std dev as % of avg
  0.200  0.267  0.255  0.226  0.196  0.189 : d->extend(vdict, 'keep')   ###-8
  0.295  0.301  0.292  0.278  0.271  0.281 :     average
 19.27%  7.27% 11.09%  9.76% 13.53% 16.07% :     std dev as % of avg
  0.010  0.017  0.025  0.030  0.026  0.008 : var ret = 2   ###-9
  0.032  0.037  0.036  0.035  0.035  0.034 :     average
 67.48% 36.36% 26.67%  9.85% 14.54% 38.46% :     std dev as % of avg
```

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -7150,7 +7150,8 @@ reltimestr({time})				*reltimestr()*
 			call MyFunction()
 			echo reltimestr(reltime(start))
 <		Note that overhead for the commands will be added to the time.
-		The accuracy depends on the system.
+		The accuracy depends on the system. Use reltimefloat() for the
+		greatest accuracy which is nanoseconds on some systems.
 		Leading spaces are used to make the string align nicely.  You
 		can use split() to remove it. >
 			echo split(reltimestr(reltime(start)))[0]
@@ -10753,6 +10754,7 @@ persistent_undo		Compiled with support for persistent undo history.
 postscript		Compiled with PostScript file printing.
 printer			Compiled with |:hardcopy| support.
 profile			Compiled with |:profile| support.
+prof_nsec		Profile results are in nano seconds.
 python			Python 2.x interface available. |has-python|
 python_compiled		Compiled with Python 2.x interface. |has-python|
 python_dynamic		Python 2.x interface is dynamically loaded. |has-python|
--- a/runtime/doc/repeat.txt
+++ b/runtime/doc/repeat.txt
@@ -1148,9 +1148,10 @@ long you take to respond to the input() prompt is irrelevant.
 Profiling should give a good indication of where time is spent, but keep in
 mind there are various things that may clobber the results:
 
-- The accuracy of the time measured depends on the gettimeofday() system
-  function.  It may only be as accurate as 1/100 second, even though the times
-  are displayed in micro seconds.
+- The accuracy of the time measured depends on the gettimeofday(), or
+  clock_gettime if available, system function. The accuracy ranges from 1/100
+  second to nano seconds. With clock_gettime the times are displayed in nano
+  seconds, otherwise micro seconds.
 
 - Real elapsed time is measured, if other processes are busy they may cause
   delays at unpredictable moments.  You may want to run the profiling several
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -6145,6 +6145,13 @@ f_has(typval_T *argvars, typval_T *rettv)
 		0
 #endif
 		},
+	{"prof_nsec",
+#ifdef HAVE_TIMER_CREATE
+		1
+#else
+		0
+#endif
+		},
 	{"reltime",
 #ifdef FEAT_RELTIME
 		1
@@ -8226,7 +8233,7 @@ init_srand(UINT32_T *x)
 #  if defined(MSWIN)
 	    *x = (UINT32_T)res.LowPart;
 #  else
-	    *x = (UINT32_T)res.tv_usec;
+	    *x = (UINT32_T)res.tv_fsec;
 #  endif
 #else
 	    *x = vim_time();
--- a/src/profiler.c
+++ b/src/profiler.c
@@ -24,7 +24,7 @@ profile_start(proftime_T *tm)
 # ifdef MSWIN
     QueryPerformanceCounter(tm);
 # else
-    gettimeofday(tm, NULL);
+    PROF_GET_TIME(tm);
 # endif
 }
 
@@ -40,12 +40,12 @@ profile_end(proftime_T *tm)
     QueryPerformanceCounter(&now);
     tm->QuadPart = now.QuadPart - tm->QuadPart;
 # else
-    gettimeofday(&now, NULL);
-    tm->tv_usec = now.tv_usec - tm->tv_usec;
+    PROF_GET_TIME(&now);
+    tm->tv_fsec = now.tv_fsec - tm->tv_fsec;
     tm->tv_sec = now.tv_sec - tm->tv_sec;
-    if (tm->tv_usec < 0)
+    if (tm->tv_fsec < 0)
     {
-	tm->tv_usec += 1000000;
+	tm->tv_fsec += TV_FSEC;
 	--tm->tv_sec;
     }
 # endif
@@ -60,11 +60,11 @@ profile_sub(proftime_T *tm, proftime_T *tm2)
 # ifdef MSWIN
     tm->QuadPart -= tm2->QuadPart;
 # else
-    tm->tv_usec -= tm2->tv_usec;
+    tm->tv_fsec -= tm2->tv_fsec;
     tm->tv_sec -= tm2->tv_sec;
-    if (tm->tv_usec < 0)
+    if (tm->tv_fsec < 0)
     {
-	tm->tv_usec += 1000000;
+	tm->tv_fsec += TV_FSEC;
 	--tm->tv_sec;
     }
 # endif
@@ -85,7 +85,7 @@ profile_msg(proftime_T *tm)
     QueryPerformanceFrequency(&fr);
     sprintf(buf, "%10.6lf", (double)tm->QuadPart / (double)fr.QuadPart);
 # else
-    sprintf(buf, "%3ld.%06ld", (long)tm->tv_sec, (long)tm->tv_usec);
+    sprintf(buf, PROF_TIME_FORMAT, (long)tm->tv_sec, (long)tm->tv_fsec);
 # endif
     return buf;
 }
@@ -102,7 +102,7 @@ profile_float(proftime_T *tm)
     QueryPerformanceFrequency(&fr);
     return (float_T)tm->QuadPart / (float_T)fr.QuadPart;
 # else
-    return (float_T)tm->tv_sec + (float_T)tm->tv_usec / 1000000.0;
+    return (float_T)tm->tv_sec + (float_T)tm->tv_fsec / (float_T)TV_FSEC;
 # endif
 }
 
@@ -123,12 +123,12 @@ profile_setlimit(long msec, proftime_T *tm)
 	QueryPerformanceFrequency(&fr);
 	tm->QuadPart += (LONGLONG)((double)msec / 1000.0 * (double)fr.QuadPart);
 # else
-	long	    usec;
+	long	    fsec;
 
-	gettimeofday(tm, NULL);
-	usec = (long)tm->tv_usec + (long)msec * 1000;
-	tm->tv_usec = usec % 1000000L;
-	tm->tv_sec += usec / 1000000L;
+	PROF_GET_TIME(tm);
+	fsec = (long)tm->tv_fsec + (long)msec * (TV_FSEC / 1000);
+	tm->tv_fsec = fsec % (long)TV_FSEC;
+	tm->tv_sec += fsec / (long)TV_FSEC;
 # endif
     }
 }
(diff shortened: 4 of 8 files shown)


brammool:
I see the profile test fails because the header changed.  Let me move that text into one place so it's easier to change.
Please sync after patch 9.0.1398

errael:
> I see the profile test fails because the header changed. Let me move that text into one place so it's easier to change. Please sync after patch 9.0.1398

DONE. 

I left test_profile.vim as seen below. I don't know if there's a feature or how to add one so that this works as expected.
```
" TODO: "if has(xxx)" for nsec vs usec output
if 1
  let s:header = 'count     total (s)      self (s)'
  let s:header_func = 'count     total (s)      self (s)  function'
else
  let s:header = 'count  total (s)   self (s)'
  let s:header_func = 'count  total (s)   self (s)  function'
endif
```

brammool:
> > I see the profile test fails because the header changed. Let me move
> > that text into one place so it's easier to change. Please sync after
> > patch 9.0.1398
> 
> DONE. 
> 
> I left test_profile.vim as seen below. I don't know if there's a
> feature or how to add one so that this works as expected.
> ```
> " TODO: "if has(xxx)" for nsec vs usec output
> if 1
>   let s:header = 'count     total (s)      self (s)'
>   let s:header_func = 'count     total (s)      self (s)  function'
> else
>   let s:header = 'count  total (s)   self (s)'
>   let s:header_func = 'count  total (s)   self (s)  function'
> endif
> ```

Hmm, so this actually depends on how Vim was build?  The macro appears
to be HAVE_TIMER_CREATE, for which I don't see a way to check for in a
Vim script.  It would require adding an item for the has() function.
Using the name "timer_create" doesn't make sense, how about something
like "profiling with nano seconds"?  "prof_nsec" could be used.
This only requires some extra lines in the f_has() function.
And a line in the help.

-- 
257. Your "hundred-and-one" lists include well over 101 items, since you
     automatically interpret all numbers in hexadecimal notation.
     (hex 101 = decimal 257)

errael:
> so this actually depends on how Vim was built?

Yes. The `clock_gettime(CLOCK_MONOTONIC)` system call
is not available on all systems.

> "profiling with nano seconds"?  "prof_nsec" could be used

The PR now has a feature `prof_nsec` usable with `has()`.

#### There is still an issue with this PR.

This PR currently has a workaround in `test_sleep.vim`
for an intermittent failure introduced by this PR.

the `reltimestr()` help says
```
String ... the number of seconds, a dot and the number of microseconds
```

But this PR currently returns the number of nano seconds.

In test_sleep, if I change `func! s:get_time_ms()` to simply return
```
return float2nr(reltimefloat(reltime()) * 1000)
```
instead of parsing the result from `reltimestr()` it fixes the problem.

Another way to fix the test is to only keep the micro sec digits
```
let msec = str2nr(timestr[dotidx + 1 :][0 : 5])
```
and then the string parsing works as expected and the test passes.

#### What to do about reltimestr()?

Is the right thing to continue to return a string that's micro seconds?

If so, then can kludge a fix in `time.c::f_reltimestr()`, and add a comment
to `reltimestr()` that `reltimefloat()` should be used for greatest accuracy.

brammool:
> > so this actually depends on how Vim was built?
> 
> Yes. The `clock_gettime(CLOCK_MONOTONIC)` system call
> is not available on all systems.
> 
> > "profiling with nano seconds"?  "prof_nsec" could be used
> 
> The PR now has a feature `prof_nsec` usable with `has()`.

OK, so using that in the test works.

> #### There is still an issue with this PR.
> 
> This PR currently has a workaround in `test_sleep.vim`
> for an intermittent failure introduced by this PR.
> 
> the `reltimestr()` help says
> ```
> String ... the number of seconds, a dot and the number of microseconds
> ```
> 
> But this PR currently returns the number of nano seconds.
> 
> In test_sleep, if I change `func! s:get_time_ms()` to simply return
> ```
> return float2nr(reltimefloat(reltime()) * 1000)
> ```
> instead of parsing the result from `reltimestr()` it fixes the problem.

It sounds good to not parse the result from reltimestr(), remove the
dependency on how that works.  Multiplying by 1000 means that you only
count milli seconds, right?  Have you checked this doesn't end up being
zero?

> Another way to fix the test is to only keep the micro sec digits
> ```
> let msec = str2nr(timestr[dotidx + 1 :][0 : 5])
> ```
> and then the string parsing works as expected and the test passes.

That seems a bit clumsy.

> #### What to do about reltimestr()?
> 
> Is the right thing to continue to return a string that's micro seconds?
> 
> If so, then can kludge a fix in `time.c::f_reltimestr()`, and add a comment
> to `reltimestr()` that `reltimefloat()` should be used for greatest accuracy.

I would think all computations are done with reltime() and the floats
returned by reltimefloat().  Then reltimestr() is only for showing the
result somewhere.  Isn't micro seconds sufficient for that?

-- 
261. You find diskettes in your pockets when doing laundry.

errael:
> > The PR now has a feature `prof_nsec` usable with `has()`.
> 
> OK, so using that in the test works.

Yes. It fixed the failures in Windows that resulted from always using
nanoseconds header. I don't think the tests use a non-windows system
that doesn't have clock_gettime.

> > In test_sleep, if I change `func! s:get_time_ms()` to simply return
> > ```
> > return float2nr(reltimefloat(reltime()) * 1000)
> > ```
> > instead of parsing the result from `reltimestr()` it fixes the problem.
> 
> It sounds good to not parse the result from reltimestr(), remove the
> dependency on how that works.  Multiplying by 1000 means that you only
> count milli seconds, right?

That's the idea.

> Have you checked this doesn't end up being
> zero?

Before the last version of thePR, I extracted the test's
`get_time_ms()` function, modified it, adding
```
timeval = float2nr(reltimefloat(relt) * 1000)
if parse_result != timeval
    echo "FAIL:" parse_result timeval
endif
```
and ran it 10,000 times with no problem.

I don't understand your quesiton about a zero result.
Is the question about `reltimefloat(reltime())`?

> > Is the right thing to continue to return a string that's micro seconds?
> 
> > If so, then can kludge a fix in `time.c::f_reltimestr()`, and add a comment
> > to `reltimestr()` that `reltimefloat()` should be used for greatest accuracy.
> 
> I would think all computations are done with reltime() and the floats
> returned by reltimefloat().  Then reltimestr() is only for showing the
> result somewhere.  Isn't micro seconds sufficient for that?

Can't say, don't know the use cases. I'd hope reltimefloat() is used, but
the test itself was a counter example.

I'll modify the PR, so that reltimestr() does it's own thing
and returns microseconds.

brammool:
> > Have you checked this doesn't end up being zero?
> 
> Before the last version of thePR, I extracted the test's
> `get_time_ms()` function, modified it, adding
> ```
> timeval = float2nr(reltimefloat(relt) * 1000)
> if parse_result != timeval
>     echo "FAIL:" parse_result timeval
> endif
> ```
> and ran it 10,000 times with no problem.
> 
> I don't understand your quesiton about a zero result.
> Is the question about `reltimefloat(reltime())`?

If you only use milliseconds and it's rounded down, then the result
might be zero.  While if you would use microseconds the value would be
non-zero.  Just making sure the test is actualy testing something
(reminds me of a bug found in code that slipped through because the test
for a timing-sensitive operation just accepted any time).

> > I would think all computations are done with reltime() and the floats
> > returned by reltimefloat().  Then reltimestr() is only for showing the
> > result somewhere.  Isn't micro seconds sufficient for that?
> 
> Can't say, don't know the use cases. I'd hope reltimefloat() is used, but
> the test itself was a counter example.
> 
> I'll modify the PR, so that reltimestr() does it's own thing
> and returns microseconds.

I think it's better to keep it doing the same thing, unless there is a
reason to change it.

-- 
264. You turn to the teletext page "surfing report" and are surprised that it
     is about sizes of waves and a weather forecast for seaside resorts.

errael:
**All checks have passed**

Wow, not even some spurious network timeout.

> > I'll modify the PR, so that reltimestr() does it's own thing
> > and returns microseconds.

> I think it's better to keep it doing the same thing, unless there is a
> reason to change it.

Done. My first struggle with `test_codestyle`, `if/else/{}`

Issue #12097, opened 2023-03-02
[xxd] Little-endian (-e) with 'weird' cols causes missing characters


aapo:
### Steps to reproduce

    echo -n "ABCDEF" | xxd  -c6 -e
    00000000: 44434241     4645CDEF

And

    echo -n "ABCDEFGHI" | xxd  -c9 -e
    00000000: 44434241 48474645     AB49EFGHI

### Expected behaviour

I expected that using -e for little-endian will not cause missing characters.

Expected output:

    00000000: 44434241     4645  ABCDEF

Or

    00000000: 44434241 4645  ABCDEF

### Version of Vim

From v7.4.652 to patch 9.0.1373

### Environment

Ubuntu, not relating terminal/shell.

aapo:
-c3 still misses space

    echo -n "ABCDEFGHI" | ./xxd  -c3 -e
    00000000: 43  41 42ABC
    00000003: 46  44 45DEF
    00000006: 49  47 48GHI

brammool:
> -c3 still misses space
> 
>     echo -n "ABCDEFGHI" | ./xxd  -c3 -e
>     00000000: 43  41 42ABC
>     00000003: 46  44 45DEF
>     00000006: 49  47 48GHI

Little-endian only works properly with a multiple of four octets.
With three octets I would not know what to display.

-- 
If you're sending someone Styrofoam, what do you pack it in?

 /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net   \\\
///                                                                      \\\
\\\        sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ ///
 \\\            help me help AIDS victims -- http://ICCF-Holland.org    ///

PR #12131, opened 2023-03-10
XXD: color-support feature

aapo:
Adds colors in xxd. Coloring the output might help some hexdumping tasks (visualizing).

xxd

The hex-value and value are both colored with the same color depending on the hex-value, e.g.:

- 0x00 = white
- 0xff = blue
- printable = green
- non-printable = red
- tabs and linebreaks = yellow
- (should be more?)

Each character needs 11 more bytes to contain color. (Same color in a row could contain only one overhead but the logic how xxd creates colums must be then changed.) Size of colored output is increased by factor of ~6. Also grepping the output will break when colors is used. 

Flag for color is "-R", because _less_ uses "-R".

Color uses parameters _auto,always,never_ same as _less_ and _grep_ (among others).

E.g. 
```
xxd -R always $FILE | less -R
```

Issues:
-  _ _ MVS _ _ not tested
-  not all col/group -combinations in little-endian (-e) works (check #12097)

### Basic
R

### ebcdic
E

### Binary (bits are not colored)
b

### Little-endian (Decimal offsets, Uppercase)
edu

### Test-script
Running 7632 combinations of flags run two times: without and with color (total of 15264 rounds).  Color is then dropped with _ansifilter_ (http://andre-simon.de/doku/ansifilter/en/ansifilter.php) and compared against known-good version of xxd.

When -c and -g are limited to the power-of-two then also all little-endian tests passed.

```bash
#!/bin/bash

for i in `printf '%.2x\n' {0..255}` ; do echo -ne "\x$i" ; done > asciiA.txt
#add one extra character because length of the data is not always 2^x
echo -n A >> asciiA.txt

DATA=asciiA.txt
XXD=./xxd_ff226d49fed2d8fc668084324c7b0f00117c5e74

function test_color_no() {
    echo -n ./xxd "$1" -R never "  "
    str1=$( $XXD $1 $DATA | md5sum)

    color_no=$(./xxd -R never $1 $DATA | md5sum)        
    if [ "$str1" == "$color_no" ]; then
        echo -n $(echo $str1 | cut -d" " -f1) " OK"
    else
        echo -n FAILED $str1 and $color_no
    fi
    echo
}

function test_color_yes() {
    echo -n ./xxd "$1" -R always " "
    str1=$( $XXD $1 $DATA | md5sum)

    color_yes=$(./xxd -R always $1 $DATA | ansifilter | md5sum)
    if [ "$str1" == "$color_yes" ]; then
        echo -n $(echo $str1 | cut -d" " -f1) " OK"
    else
        echo -n FAILED $str1 and $color_yes
    fi
    echo
}

function not_power_of_two () {
    declare -i n=$1
    (( n > 0 && (n & (n - 1)) != 0 ))
}

for a in "" "-a"
do
 for b in "" "-e" "-b" 
 do
  for d in "" "-d"
  do
   for E in "" "-E"
   do
    for u in "" "-u"
    do
     for c in {1..21}
     do
     for (( g=1; g<=c; g++ ))
      do
      if  [ "$b" == "-e" ] && not_power_of_two "$g" 
       then
        :
      elif [ "$b" == "-e" ] && not_power_of_two "$c" 
       then
        :
       else
        test_color_no  "$a $b $d $E $u -c $c -g $g"
        test_color_yes "$a $b $d $E $u -c $c -g $g"
       fi
      done
     done
    done
   done
  done
 done
done
exit 0
```

### DIFF
--- a/src/testdir/test_xxd.vim
+++ b/src/testdir/test_xxd.vim
@@ -415,4 +415,132 @@ func Test_xxd_little_endian_with_cols()
   bwipe!
 endfunc
 
+func Test_xxd_color()
+  "Test: color=never
+  let s:test = 1
+
+  "Note Quotation mark escaped
+  "Note Aposhpere vaihdettu apostrophe replaced with 0x00
+  "Note Backslash replaced with 0x00
+  let data = [
+      \ "00000000: 0001 0203 0405 0607 0809 0a0b 0c0d 0e0f  ................",
+      \ "00000010: 1011 1213 1415 1617 1819 1a1b 1c1d 1e1f  ................",
+      \ "00000020: 2021 2223 2425 2600 2829 2a2b 2c2d 2e2f   !\"#$%&.()*+,-./",
+      \ "00000030: 3031 3233 3435 3637 3839 3a3b 3c3d 3e3f  0123456789:;<=>?",
+      \ "00000040: 4041 4243 4445 4647 4849 4a4b 4c4d 4e4f  @ABCDEFGHIJKLMNO",
+      \ "00000050: 5051 5253 5455 5657 5859 5a5b 005d 5e5f  PQRSTUVWXYZ[.]^_",
+      \ "00000060: 6061 6263 6465 6667 6869 6a6b 6c6d 6e6f  `abcdefghijklmno",
+      \ "00000070: 7071 7273 7475 7677 7879 7a7b 7c7d 7e7f  pqrstuvwxyz{|}~.",
+      \ "00000080: 8081 8283 8485 8687 8889 8a8b 8c8d 8e8f  ................",
+      \ "00000090: 9091 9293 9495 9697 9899 9a9b 9c9d 9e9f  ................",
+      \ "000000a0: a0a1 a2a3 a4a5 a6a7 a8a9 aaab acad aeaf  ................",
+      \ "000000b0: b0b1 b2b3 b4b5 b6b7 b8b9 babb bcbd bebf  ................",
+      \ "000000c0: c0c1 c2c3 c4c5 c6c7 c8c9 cacb cccd cecf  ................",
+      \ "000000d0: d0d1 d2d3 d4d5 d6d7 d8d9 dadb dcdd dedf  ................",
+      \ "000000e0: e0e1 e2e3 e4e5 e6e7 e8e9 eaeb eced eeef  ................",
+      \ "000000f0: f0f1 f2f3 f4f5 f6f7 f8f9 fafb fcfd feff  ................"]
+  call writefile(data,'Xinput')
+
+  silent exe '!' . s:xxd_cmd . ' -r < Xinput > XXDfile'
+
+  %d
+  exe '0r! ' . s:xxd_cmd . ' -R never ' . ' XXDfile'
+  $d
+  let expected = [
+      \ "00000000: 0001 0203 0405 0607 0809 0a0b 0c0d 0e0f  ................",
+      \ "00000010: 1011 1213 1415 1617 1819 1a1b 1c1d 1e1f  ................",
+      \ "00000020: 2021 2223 2425 2600 2829 2a2b 2c2d 2e2f   !\"#$%&.()*+,-./",
+      \ "00000030: 3031 3233 3435 3637 3839 3a3b 3c3d 3e3f  0123456789:;<=>?",
+      \ "00000040: 4041 4243 4445 4647 4849 4a4b 4c4d 4e4f  @ABCDEFGHIJKLMNO",
+      \ "00000050: 5051 5253 5455 5657 5859 5a5b 005d 5e5f  PQRSTUVWXYZ[.]^_",
+      \ "00000060: 6061 6263 6465 6667 6869 6a6b 6c6d 6e6f  `abcdefghijklmno",
+      \ "00000070: 7071 7273 7475 7677 7879 7a7b 7c7d 7e7f  pqrstuvwxyz{|}~.",
+      \ "00000080: 8081 8283 8485 8687 8889 8a8b 8c8d 8e8f  ................",
+      \ "00000090: 9091 9293 9495 9697 9899 9a9b 9c9d 9e9f  ................",
+      \ "000000a0: a0a1 a2a3 a4a5 a6a7 a8a9 aaab acad aeaf  ................",
+      \ "000000b0: b0b1 b2b3 b4b5 b6b7 b8b9 babb bcbd bebf  ................",
+      \ "000000c0: c0c1 c2c3 c4c5 c6c7 c8c9 cacb cccd cecf  ................",
+      \ "000000d0: d0d1 d2d3 d4d5 d6d7 d8d9 dadb dcdd dedf  ................",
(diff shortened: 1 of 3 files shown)


chrisbra:
Hi, 
thanks for your contribution. I really like to include it. However, I have two questions:
- can you please add a test in [test_xxd.vim](https://github.com/vim/vim/blob/master/src/testdir/test_xxd.vim)
- can you please adjust the [manpage](https://github.com/vim/vim/blob/master/runtime/doc/xxd.1)

Thanks!

chrisbra:
some build error on Windows:
```
xxd.c(520): error C2065: 'STDOUT_FILENO': undeclared identifier
NMAKE : fatal error U1077: 'cl /nologo -DWIN32 xxd.c -link -subsystem:console' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.35.32215\bin\HostX64\x64\nmake.exe" /NOLOGO -f Make_mvc.mak ' : return code '0x2'
Stop.
Error: Process completed with exit code 2.
```

src/testdir/test_xxd.vim
```diff
@@ -415,4 +415,132 @@ func Test_xxd_little_endian_with_cols()
   bwipe!
 endfunc
 
+func Test_xxd_color()
```
LemonBoy, line 418: The test should use vim's screendump features.

src/xxd/xxd.c
```diff
@@ -498,6 +517,11 @@ main(int argc, char *argv[])
   char *pp;
   char *varname = NULL;
   int addrlen = 9;
+  int color=0;
```
LemonBoy: Missing space around `=`, ditto for several other places in the newly-added code.

src/xxd/xxd.c
```diff
@@ -498,6 +517,11 @@ main(int argc, char *argv[])
...
+
+#ifndef  __MVS__
```
LemonBoy: `__MVS__` is checking if the host system is z/OS, which I believe is not what you meant.

src/xxd/xxd.c
```diff
@@ -844,8 +888,61 @@ main(int argc, char *argv[])
       c = addrlen + 1 + (grplen * x) / octspergrp;
       if (hextype == HEX_NORMAL || hextype == HEX_LITTLEENDIAN)
 	{
+          if (color)
+            {
+            COLOR_PROLOGUE
+
+            if (ebcdic)  /* EBCDIC */
```
LemonBoy: This code is repeated a few lines below, consider introducing a lookup table for converting the EBCDIC/ASCII codepoint to color.

aapo: I reimplemented it with function.

src/testdir/test_xxd.vim
aapo: I don't know anything about vim's screendump features.

LemonBoy: `:h terminal-screendump`

src/xxd/xxd.c
```diff
@@ -482,6 +501,51 @@ static unsigned char etoa64[] =
     0070,0071,0372,0373,0374,0375,0376,0377
 };
 
+   static char
+begin_coloring_char(char *l,int *c,int e,int ebcdic) {
```
LemonBoy: The spacing is still wonky (e.g. no space after commas).
The function's return type is `char` but it doesn't return anything. Perhaps make it return `int` so that the index `c` can be taken by value.

src/xxd/xxd.c
```diff
@@ -482,6 +501,51 @@ static unsigned char etoa64[] =
...
+  if (ebcdic)  /* EBCDIC */
```
LemonBoy: This comment made me laugh out loud, can be safely removed :D

src/xxd/xxd.c
```diff
@@ -855,36 +958,122 @@ main(int argc, char *argv[])
...
+            c += 1;
+
+            COLOR_PROLOGUE
+            begin_coloring_char(l,&c,e,ebcdic);
+
+            #ifdef __MVS__
+            if (e >= 64) l[c++] = e;
+            else  l[c++] = '.';
+            #else
+
```
LemonBoy, line 976: Extra line.

src/xxd/xxd.c
```diff
@@ -498,6 +562,12 @@ main(int argc, char *argv[])
   char *pp;
   char *varname = NULL;
   int addrlen = 9;
+  int color = 0;
+
+
```
LemonBoy: Extra line.

src/xxd/xxd.c
```diff
@@ -855,36 +958,122 @@ main(int argc, char *argv[])
...
+        int x = p;
+        if (hextype == HEX_LITTLEENDIAN)
+          {
+            int fill = octspergrp - (p % octspergrp);
+            if (fill == octspergrp) fill = 0;
+
+            c = addrlen + 1 + (grplen * (x - (octspergrp-fill))) / octspergrp;
+
+            int i;
+            for (i = 0;i < fill;i++)
```
LemonBoy: Missing spaces after `;`, you can declare `i` in the for loop.

src/xxd/xxd.c
```diff
@@ -855,36 +958,122 @@ main(int argc, char *argv[])
...
+        if (hextype != HEX_BITS)
+          {
+            c = addrlen + 1 + (grplen * x) / octspergrp;
+            c += cols - p;
+            c += (cols - p) / octspergrp;
+
+            int i;
+            for (i = cols - p;i > 0;i--)
+              {
+                COLOR_PROLOGUE
```
LemonBoy, line 1060: You can emit the prologue and epilogue only once and generate the `n` whitespaces.
Also, why is the space printed in red?

src/testdir/test_xxd.vim
aapo: I have never used vim.

src/xxd/xxd.c
aapo: Color is not important but this is the easiest way to calculate/print the place when using `-e`

Without invisible (red) white spaces:
```
echo -n A | ./xxd -e -R never
00000000:       41                              A
echo -n A | ./xxd -e -R always
00000000:                                        41                              A
```

src/xxd/xxd.c
aapo: Declaring variables inside a for loop wasn't valid C until C99.

Currently `xxd` can be compiled:
```
cc --ansi -DUNIX -o xxd xxd.c
```

I think this is not the reason to break that.

And after writing that I just figured I just broke `--pedantic`.

src/xxd/xxd.c
aapo: I planned to use return value for something but then figured that passing pointer would be cleaner. (And then forgot type of return value.)

chrisbra:
Thanks, I have made a few more changes (mainly to fix style and add tests) and I am about to merge the following:

https://github.com/vim/vim/compare/master...chrisbra:vim:gh-12131?expand=1

Any comments before I merge?

PR #12142, opened 2023-03-12
docs(fold/`za`): use keywords for easy search

UnkwUsr:
I constantly forget this map and start searching for "toggle" in `:help fold.txt`, but no matches, so I have to open google. This will solve it.
(From https://github.com/neovim/neovim/pull/22642)


brammool:
I can add something like this.  It's easier to read when it comes first.

PR #12146, opened 2023-03-12
shell syntax file: add shDblParen for sh{Case,Loop}List

grondilu:
This should allow arithmetic compound commands inside case and loop blocks.

Minimal code below.  If I'm not mistaken original syntax file fails to parse this correctly:
```bash
for ((;;))
do
  ((1 << 2))
  break
done

case $RANDOM in
  1) ((1 << 2));;
  *)
esac
```


grondilu:
I've just realized that this affects ksh too, but I just checked and there are arithmetic compounds commands in ksh too, with the same syntax.  So it should be ok.

brammool:
ping @cecamp

yegappan:
@cecamp Any updates on this?

PR #12148, opened 2023-03-13
Fix aucmd_restbuf not stopping insert mode when leaving prompt buffers.

### LINKED ISSUE #12147: Appending to prompt buffer enters insert mode
### Steps to reproduce

Run the following script:

```vim
new prompt
set buftype=prompt
set bufhidden=hide
startinsert
close
call appendbufline("prompt", "$", "Test")
```
Now hit any key and we are in insert mode.

### Expected behaviour

I would expect to still be in normal mode.

### Version of Vim

9.0.1403

### Environment

OS: Manjaro 22.0.4
Terminal: kitty 0.27.1
$TERM: xterm-kitty
shell: fish 3.6.0

### Logs and stack traces

_No response_

t-troebst:
This closes #12147.

### DIFF
--- a/src/autocmd.c
+++ b/src/autocmd.c
@@ -1657,6 +1657,10 @@ aucmd_restbuf(
 	    }
 	}
 win_found:
+#ifdef FEAT_JOB_CHANNEL
+	// May need to stop insert mode if we were in a prompt buffer.
+	leaving_window(curwin);
+#endif
 
 	// Remove the window and frame from the tree of frames.
 	(void)winframe_remove(curwin, &dummy, NULL);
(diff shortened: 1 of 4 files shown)


t-troebst:
Could I get any input on this? Is the macOS m1 failure a flake? This change is a nop for everything that doesn't use prompt buffers so I don't see how Test_spelldump_bang could be affected (not to mention that all other test suites pass, including the github CI macOS suites).

brammool:
Also see my comment on #12147 - The ":startinsert" in the script takes effect after the script is finished.
The newly added test is wrong, it should end up in Insert mode.  This actually indicates that the source
code change is breaking the behavior of ":startinsert".

I suspect you need another way to reproduce the actual problem you are trying to fix.  Probably without using ":startinsert", Insert mode is started automatically in a prompt buffer.

PR #12152, opened 2023-03-14
Add Free Form ILE RPG filetype files

andlrc:
It seems like vim doesn't ship with syntax / filetype files for the ILE RPG programming language:

https://en.wikipedia.org/wiki/IBM_RPG and  
https://www.ibm.com/docs/en/i/7.5?topic=rpg-ile-reference

This PR contains ftdetect, syntax, indent and ftplugin files for Free Form ILE RPG files.

The files `*.rpgle` and `*.rpgleinc` will be detected as being the `filetype=rpgle`.

This PR contains adopted code from https://github.com/andlrc/rpgle.vim

### DIFF
--- a/runtime/ftplugin/rpgle.vim
+++ b/runtime/ftplugin/rpgle.vim
@@ -1,0 +1,132 @@
+" Vim ftplugin file
+" Language:             Free-Form ILE RPG
+" Maintainer:           Andreas Louv <andreas@louv.dk>
+" Last Change:          Mar 14, 2023
+" Version:              1
+
+if exists('b:did_ftplugin')
+  finish
+endif
+
+let b:did_ftplugin = 1
+
+setlocal iskeyword+=-,%
+
+setlocal suffixesadd=.rpgle,.rpgleinc
+setlocal include=^\\s*/\\%(include\\\|copy\\)
+
+setlocal comments=s1:/*,mb:*,ex:*/,://,:*
+
+" ILE RPG is in case-sensitive
+setlocal tagcase=ignore nosmartcase ignorecase
+
+let b:match_words = '\<select\>:\<when\>:\<other\>:\<endsl\>'
+                \ . ',\<if\>:\<elseif\>:\<else\>:\<endif\>'
+                \ . ',\<do[uw]\>:\<iter\>:\<leave\>:\<enddo\>'
+                \ . ',\<for\>:\<iter\>:\<leave\>:\<endfor\>'
+                \ . ',\<begsr\>:\<endsr\>'
+                \ . ',\<dcl-proc\>:\<return\>:\<end-proc\>'
+                \ . ',\<dcl-pi\>:\<end-pi\>'
+                \ . ',\<dcl-pr\>:\<end-pr\>'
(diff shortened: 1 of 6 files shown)


amaanq:
Forgive me if this is a stupid question stemming from my curiosity, but is ILE RPG the same as IBM's? RPG was on my todo list of creating treesitter grammars for to then add (really good) syntax highliighting support in [neovim treesitter](https://github.com/nvim-treesitter/nvim-treesitter)

brammool:
> It seems like vim doesn't ship with syntax / filetype files for the
> ILE RPG programming language:
> 
> https://en.wikipedia.org/wiki/IBM_RPG and  
> https://www.ibm.com/docs/en/i/7.5?topic=rpg-ile-programmers-guide
> 
> This PR contains ftdetect, syntax, indent and ftplugin files for Free
> Form ILE RPG files.
> 
> The files `*.rpgle` and `*.rpgleinc` will be detected as being the
> `filetype=rpgle`.

I can include the filetype detection part.  If there is something in
common with other RPG filetypes that can be handled in the plugins.

You should not use "function!" in plugins.  If the script is loaded
again, a function can be redefined once, also without the exclamation
mark.  The exclamation mark allows for redefining multiple times, which
can hide making the mistake of using the same function name twice.

In the syntax file please use "hi def link" instead of "highlight link".

There is no need to include an update for doc/tags, it's generated.

-- 
A cow comes flying over the battlements,  lowing aggressively.  The cow
lands on GALAHAD'S PAGE, squashing him completely.

runtime/ftplugin/rpgle.vim
```diff
@@ -0,0 +1,101 @@
...
+nnoremap <script> <buffer> <silent> <Plug>RpgleNextBlock
+       \ :call rpgle#NextNest('')<CR>
+
+" Operator Pending brackets
+noremap <script> <buffer> <silent> <Plug>RpgleAroundBlock
+       \ :<C-U>call rpgle#Operator('a')<CR>
+noremap <script> <buffer> <silent> <Plug>RpgleInnerBlock
+       \ :<C-U>call rpgle#Operator('i')<CR>
+
+if !get(g:, 'rpgle_skipMapping', 0)
```
dkearns: It would be better to use the standard mechanism to disable plugin mappings.  See `:help no_plugin_maps`.

So this line would be something like:

```vim
if !exists("g:no_plugin_maps") && !exists("g:no_rpgle_maps")
```

This file should also set b:undo_ftplugin.

andlrc:
> I can include the filetype detection part.  If there is something in
common with other RPG filetypes that can be handled in the plugins.

The way I read this is that the ftplugin, syntax, indent etc, should be handled by an external plugin, I guess that makes sense.

I can see that you already included the filetype detection part, so I'll close up this PR then.

> There is no need to include an update for doc/tags, it's generated.

Good to know.

> You should not use "function!" in plugins. [...]
> 
> In the syntax file please use "hi def link" instead of "highlight link".

and

> It would be better to use the standard mechanism to disable plugin mappings. See :help no_plugin_maps.
> [...]
> This file should also set b:undo_ftplugin.

Thanks for the valuable feedback, I'll bring it upstream in the previous linked plugin.

PR #3476, opened 2018-09-23
WIP: l10n(zh_TW): update translations

pan93412:
## Progress
- [x] Start project
- [ ] Review all the translation
- [ ] Re-add the header
- [ ] Generate zh_TW.po (BIG5)

If you are interested in this project, please open a
issue in this translation repository (tw-translation/vim:master)
that I can allocate the part you have to translate.

Thanks for you guys help.

PR #12153, opened 2023-03-15
Make converted locale files build in macOS, and fix zh_TW.po

ychin:
Vim's build system currently uses iconv to pre-convert .po files to different encodings to accommodate Vim versions that don't have iconv. It uses sed to add a generated file / do-not-edit marker to the converted file (to make sure contributors don't erroneously try to edit the generated version), but that step fails in macOS because macOS's version of sed is locale aware and gets confused (see https://stackoverflow.com/questions/19242275/re-error-illegal-byte-sequence-on-mac-os-x). To fix this, make sure we prepend `LANG=C` before calling sed which fixes the issue.

Also fix a couple issues with locale files:

- zh_TW.po is currently generated from zh_TW.UTF-8.po but it's not configured as such in the Makefile. In fact, it doesn't contain a couple updates from the UTF-8 version. Just add it to the Makefile and re-generate it as well.
- Minor: ko.po is generated from ko.UTF-8.po, so it should be under `$(MOCONVERTED)` instead of `$(MOFILES)`.

### DIFF
--- a/runtime/lang/Makefile
+++ b/runtime/lang/Makefile
@@ -25,24 +25,25 @@ all: \
 		menu_hu_hu.iso_8859-2.vim \
 		menu_sk_sk.iso_8859-2.vim \
 
+SED = LANG=C sed
 
 # Convert menu_zh_cn.utf-8.vim to create menu_chinese_gb.936.vim.
 menu_chinese_gb.936.vim: menu_zh_cn.utf-8.vim
 	rm -f menu_chinese_gb.936.vim
 	iconv -f utf-8 -t cp936 menu_zh_cn.utf-8.vim | \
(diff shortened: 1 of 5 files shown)


ychin:
Also tested this on Ubuntu Linux and made sure it still works.

k-takata:
To convert zh_TW.po from zh.UTF-8.po, how about using this?
https://github.com/vim/vim/pull/3476#issuecomment-431583112
(not tested, though)

ychin:
> To convert zh_TW.po from zh.UTF-8.po, how about using this? [#3476 (comment)](https://github.com/vim/vim/pull/3476#issuecomment-431583112) (not tested, though)

I did see that PR when digging through zh_TW translations but sad to see it never merged :(.

I think maybe we should tackle the backslash issue in a dedicated PR? This was mostly trying to fix basic infrastructure issues and I feel like I would need to look more closely with your backslash correction tool which would complicate this more. Currently there are already rules around backlisted characters and I don't think the zh_TW.UTF-8.po file violates that.

PR #12157, opened 2023-03-15
Fix mouseshape to be set correctly when using 'r' or 'gr'

ychin:
Currently, when entering the "pretend" or single character replace modes using `r` or `gr` (in GUI), the mouse cursor doesn't immediately update until you have re-focused the window or moved the mouse. This is because it's not calling `update_mouseshape(-1)` immediately, so the cursor will only be updated when it's called by other functions like `gui_mouse_focus` (which happens just if you move the mouse or re-focus the application).

To fix this, just make sure we call this `update_mouseshape(-1)`. It's what we do when entering Insert or Replace modes for example.

I noticed this when trying to figure out why MacVim CI is failing in `Test_mouse_shape_after_cancelling_gr` (introduced in #12110), but I think that test is only passing in Vim GTK CI by accident, since this issue happens there too. I think the window captured focus after the mouse `gr` call which triggers a mouse shape change but it probably would have failed under other circumstances.

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -418,6 +418,9 @@ normal_cmd_get_more_chars(
 #ifdef CURSOR_SHAPE
 	    ui_cursor_shape();	// show different cursor shape
 #endif
+#ifdef FEAT_MOUSESHAPE
+	    update_mouseshape(-1);
+#endif
 	}
 	if (lang && curbuf->b_p_iminsert == B_IMODE_LMAP)
 	{


brammool:
> Currently, when entering the "pretend" or single character replace
> modes using `r` or `gr`, the mouse cursor doesn't immediately update
> until you have re-focused the window or moved the mouse. This is
> because it's not calling `update_mouseshape(-1)` immediately, so the
> cursor will only be updated when it's called by other functions like
> `gui_mouse_focus`.
> 
> To fix this, just make sure we call this `update_mouseshape(-1)`. It's
> what we do when entering Insert or Replace modes for example.
> 
> I noticed this when trying to figure out why MacVim CI is failing in
> `Test_mouse_shape_after_cancelling_gr` (introduced in #12110), but I
> think that test is only passing in Vim GTK CI by accident, since this
> issue happens there too. I think the window captured focus after the
> mouse `gr` call which triggers a mouse shape change but it probably
> would have failed under other circumstances.

Thanks for looking into this.  Could you add a test for this?  There are
some examples at the end of src/testdir/test_normal.vim.  However, the
ugly "sleep 300m" should be removed, use WaitForAssert() instead.

-- 
Every person is responsible for the choices he makes.

ychin:
Sure. Let me take a look.

yegappan:
Any updates on this?

ychin:
I remember the tests were a little tricky to actually write to make it a proper regression test, and then other things came up and I neglected to follow up on this. Let me take a look again.

chrisbra:
any updates here please ?

ychin:
Hi, sorry i never wrote a proper regression test for this, and totally forgot to respond. Thanks for merging this.

Just for context, I think I saw something similar to #14660 intermittently so in a way just passing the tests show that it works,but then the tests didn't seem to always fail before so I wasn't sure what the behavior actually was.

ychin:
For #16042 to at least use Bram's suggestion (https://github.com/vim/vim/pull/12157#issuecomment-1510482845) to use `WaitForAssert` for faster tests (also what #14710 did). This doesn't change the regression test but make them run better.

PR #12159, opened 2023-03-16
Fix wrong syntax of ldSections in ld.vim

Freed-Wu:
This is an example:

```ld
SECTIONS
{
  .text: {*(.text)}
}
```

The `)` after `.text` shouldn't be recognized as `ldSections`.

Same as <https://github.com/neovim/neovim/blob/master/runtime/syntax/ld.vim> and <https://github.com/sheerun/vim-polyglot/blob/master/syntax/ld.vim>.

### DIFF
--- a/runtime/syntax/ld.vim
+++ b/runtime/syntax/ld.vim
@@ -43,7 +43,7 @@ syn match   ldSpecial       '/DISCARD/'
 syn keyword ldIdentifier    ORIGIN LENGTH
 
 syn match   ldSpecSections  '\.'
-syn match   ldSections      '\.\S\+'
+syn match   ldSections      '\.[^ \t)]\+'
 syn match   ldSpecSections  '\.\%(text\|data\|bss\|symver\)\>'
 
 syn match   ldNumber        display '\<0[xX]\x\+\>'


brammool:
The Neovim file still has ```\S```.  The other link you gave too.
I don't know this syntax, but perhaps this should match keyword characters instead of non-white characters?

Freed-Wu:
Currently:

screen-2023-03-27-16-44-49

After modification:

screen-2023-03-27-16-45-52

brammool:
I understand that excluding the right parenthesis fixes the highlighting for this text.
I just wonder if there is a more generic change.  Is there another character that can follow "ldSections"?
Thus should be excluded from the matched characters?

chrisbra:
> Same as https://github.com/neovim/neovim/blob/master/runtime/syntax/ld.vim and https://github.com/sheerun/vim-polyglot/blob/master/syntax/ld.vim.

Both of those seem to be the old Vim version. So let me include this now, seems like it is a clear improvement

chrisbra:
ah, this was already included as part of https://github.com/vim/vim/commit/71badf9547e8f89571b9a09518#diff-7d1f3e200f03eb3ec75445eab6aac321fb7862510a849454b0de5b716801425c 

so closing

PR #12165, opened 2023-03-16
Separate wat and wast filetypes

amaanq:
While very similar, wast and wat files are not the same (wast is a superset of wat)

This distinction makes it easier to add tree-sitter specific highlighting for wat and wast, and allows other plugins to distinguish the two


brammool:
filetype test is failing.

amaanq:
Should be fixed now..sorry

amaanq:
Ping @brammool

PR #12166, opened 2023-03-16
Implement foreach({expr1}, {expr2})

errael:
Not sure which lock errors might be returned, will adjust the help as needed. Assuming can put the list/dict into `lockvar 1` mode and pass on whatever errors show up.

If setting/clearing the lock is cheap, is it worth considering retrofitting `map()` and `filter()` to have `expr1` under lock until a change needs to be made?

#### Test
`foreach` should work if called with an `expr1`  that has `lockvar 1` set.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -198,6 +198,8 @@ foldclosedend({lnum})		Number	last line of fold at {lnum} if closed
 foldlevel({lnum})		Number	fold level at {lnum}
 foldtext()			String	line displayed for closed fold
 foldtextresult({lnum})		String	text for closed fold at {lnum}
+foreach({expr1}, {expr2})	List/Dict/Blob/String
+					for each item in {expr1} call {expr2}
 foreground()			Number	bring the Vim window to the foreground
 fullcommand({name} [, {vim9}])	String	get full command from {name}
 funcref({name} [, {arglist}] [, {dict}])
@@ -2995,6 +2997,45 @@ foldtextresult({lnum})					*foldtextresult()*
 
 		Can also be used as a |method|: >
 			GetLnum()->foldtextresult()
+
+foreach({expr1}, {expr2})					*foreach()*
+		{expr1} must be a |List|, |String|, |Blob| or |Dictionary|.
+		For each item in {expr1} execute {expr2}. {expr1} is not
+		modified; its values may be, as with |:lockvar| 1. *E741*
+		See |map()| and |filter()| to modify {expr1}.
+
+		{expr2} must be a |string| or |Funcref|.
+
+		If {expr2} is a |string|, inside {expr2} |v:val| has the value
+		of the current item.  For a |Dictionary| |v:key| has the key
+		of the current item and for a |List| |v:key| has the index of
+		the current item.  For a |Blob| |v:key| has the index of the
+		current byte. For a |String| |v:key| has the index of the
+		current character.
+		Examples: >
+			call foreach(mylist, 'used[v:val] = true')
+<		This records the items that are in the {expr1} list.
+
+		Note that {expr2} is the result of expression and is then used
+		as a command.  Often it is good to use a |literal-string| to
+		avoid having to double backslashes.
+
+		If {expr2} is a |Funcref| it must take two arguments:
+			1. the key or the index of the current item.
+			2. the value of the current item.
+		With a legacy script lambda you don't get an error if it only
+		accepts one argument, but with a Vim9 lambda you get "E1106:
+		One argument too many", the number of arguments must match.
+		If the function returns a value, it is ignored.
+
+		Returns {expr1} in all cases.
+		When an error is encountered while executing {expr2} no
+		further items in {expr1} are processed.
+		When {expr2} is a Funcref errors inside a function are ignored,
+		unless it was defined with the "abort" flag.
+
+		Can also be used as a |method|: >
+			mylist->foreach(expr2)
 <
 							*foreground()*
 foreground()	Move the Vim window to the foreground.  Useful when sent from
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -5167,6 +5167,7 @@ E738	eval.txt	/*E738*
 E739	builtin.txt	/*E739*
 E74	message.txt	/*E74*
 E740	userfunc.txt	/*E740*
+E741	builtin.txt	/*E741*
 E741	eval.txt	/*E741*
 E742	userfunc.txt	/*E742*
 E743	eval.txt	/*E743*
@@ -7148,6 +7149,7 @@ foldtextresult()	builtin.txt	/*foldtextresult()*
 font-sizes	gui_x11.txt	/*font-sizes*
 fontset	mbyte.txt	/*fontset*
 forced-motion	motion.txt	/*forced-motion*
+foreach()	builtin.txt	/*foreach()*
 foreground()	builtin.txt	/*foreground()*
 fork	os_unix.txt	/*fork*
 form.vim	syntax.txt	/*form.vim*
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -798,6 +798,7 @@ List manipulation:					*list-functions*
 	filter()		remove selected items from a List
 	map()			change each List item
 	mapnew()		make a new List with changed items
+	foreach()		apply function to List items
 	reduce()		reduce a List to a value
 	slice()			take a slice of a List
 	sort()			sort a List
@@ -829,6 +830,7 @@ Dictionary manipulation:				*dict-functions*
 	filter()		remove selected entries from a Dictionary
 	map()			change each Dictionary entry
 	mapnew()		make a new Dictionary with changed items
+	foreach()		apply function to Dictionary items
 	keys()			get List of Dictionary keys
 	values()		get List of Dictionary values
 	items()			get List of Dictionary key-value pairs
--- a/src/blob.c
+++ b/src/blob.c
@@ -641,25 +641,28 @@ blob_filter_map(
 	if (filter_map_one(&tv, expr, filtermap, fc, &newtv, &rem) == FAIL
 		|| did_emsg)
 	    break;
-	if (newtv.v_type != VAR_NUMBER && newtv.v_type != VAR_BOOL)
+	if (filtermap != FILTERMAP_FOREACH)
 	{
-	    clear_tv(&newtv);
-	    emsg(_(e_invalid_operation_for_blob));
-	    break;
-	}
-	if (filtermap != FILTERMAP_FILTER)
-	{
-	    if (newtv.vval.v_number != val)
-		blob_set(b_ret, i, newtv.vval.v_number);
-	}
-	else if (rem)
-	{
-	    char_u *p = (char_u *)blob_arg->bv_ga.ga_data;
+	    if (newtv.v_type != VAR_NUMBER && newtv.v_type != VAR_BOOL)
+	    {
+		clear_tv(&newtv);
+		emsg(_(e_invalid_operation_for_blob));
+		break;
+	    }
+	    if (filtermap != FILTERMAP_FILTER)
+	    {
+		if (newtv.vval.v_number != val)
+		    blob_set(b_ret, i, newtv.vval.v_number);
+	    }
+	    else if (rem)
+	    {
+		char_u *p = (char_u *)blob_arg->bv_ga.ga_data;
 
-	    mch_memmove(p + i, p + i + 1,
-		    (size_t)b->bv_ga.ga_len - i - 1);
-	    --b->bv_ga.ga_len;
-	    --i;
+		mch_memmove(p + i, p + i + 1,
+			    (size_t)b->bv_ga.ga_len - i - 1);
+		--b->bv_ga.ga_len;
+		--i;
+	    }
 	}
 	++idx;
     }
--- a/src/dict.c
+++ b/src/dict.c
@@ -1329,8 +1329,8 @@ dict_extend_func(
 }
 
 /*
- * Implementation of map() and filter() for a Dict.  Apply "expr" to every
- * item in Dict "d" and return the result in "rettv".
+ * Implementation of map(), filter(), foreach() for a Dict.  Apply "expr" to
+ * every item in Dict "d" and return the result in "rettv".
  */
     void
 dict_filter_map(
@@ -1392,7 +1392,6 @@ dict_filter_map(
 			    arg_errmsg, TRUE)))
 		break;
 	    set_vim_var_string(VV_KEY, di->di_key, -1);
-	    newtv.v_type = VAR_UNKNOWN;
 	    r = filter_map_one(&di->di_tv, expr, filtermap, fc, &newtv, &rem);
 	    clear_tv(get_vim_var_tv(VV_KEY));
 	    if (r == FAIL || did_emsg)
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -607,10 +607,11 @@ arg_list_or_dict_or_blob_or_string_mod(
 }
 
 /*
- * Check second argument of map() or filter().
+ * Check second argument of map(), filter(), foreach().
  */
     static int
-check_map_filter_arg2(type_T *type, argcontext_T *context, int is_map)
+check_map_filter_arg2(type_T *type, argcontext_T *context,
+							filtermap_T filtermap)
 {
     type_T *expected_member = NULL;
     type_T *(args[2]);
@@ -663,12 +664,14 @@ check_map_filter_arg2(type_T *type, argcontext_T *context, int is_map)
     {
 	where_T where = WHERE_INIT;
 
-	if (is_map)
+	if (filtermap == FILTERMAP_MAP)
 	    t_func_exp.tt_member = expected_member == NULL
 					|| type_any_or_unknown(type->tt_member)
 				? &t_any : expected_member;
-	else
+	else if (filtermap == FILTERMAP_FILTER)
 	    t_func_exp.tt_member = &t_bool;
+	else // filtermap == FILTERMAP_FOREACH
+	    t_func_exp.tt_member = &t_unknown;
 	if (args[0] == NULL)
 	    args[0] = &t_unknown;
 	if (type->tt_argcount == -1)
@@ -693,7 +696,7 @@ arg_filter_func(type_T *type, type_T *decl_type UNUSED, argcontext_T *context)
 	return OK;
 
     if (type->tt_type == VAR_FUNC)
-	return check_map_filter_arg2(type, context, FALSE);
+	return check_map_filter_arg2(type, context, FILTERMAP_FILTER);
     semsg(_(e_string_or_function_required_for_argument_nr), 2);
     return FAIL;
 }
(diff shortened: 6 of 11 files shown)


errael:
`foreach()` now works for a list (limited testing). A review of the
approach is appreciated before looking at dict/blob/string.

#### One line lambda with a command

I was hoping to be able to do something like
```
l->foreach((_, v) => dict1[v] += 1)
```
which fails because `dict1[v] += 1` is not an expression.

The following, `void` return and/or `:`, also don't work, but might be a
reasonable way to signal the intent for a semantic extension.
Using `: void` seems the most clear.
```
l->foreach((_, v): void => dict1[v] += 1)
l->foreach((_, v) => :dict1[v] += 1)
```

Would something like this be acceptable?

#### Variable locking

Turns out that in `list_filter_map()` there is some locking; keep that
behavior for `foreach()`. Since top level lock is an integer assignment, it
should be cheap to keep a FILTERed list locked uless/until an element is
removed. Is that worth adding to prevent list modification while traversing
the list?

#### Executing a string

In `filter_map_one()`, at the top, there's now
```
    if (filtermap == FILTERMAP_FOREACH && expr->v_type == VAR_STRING)
    {
	// foreach() is not limited to an expression
	do_cmdline_cmd(expr->vval.v_string);
        ...
```
Is this the best way? Or at least good enough?

#### Help text

In the help for `map()` and `filter()` there's
```
When {expr2} is a Funcref errors inside a function are ignored,
unless it was defined with the "abort" flag.
```
It's included in the `foreach()` help, but I don't know what/why. What's
a better way to word this?

errael:
This PR is needs to be udpated with https://github.com/vim/vim/commit/e6d40dcdc7227594935d2db01eca29f0e575dcee, and add some more tests.

errael:
Last push is a rebase to master, should be no other changes. Now for tests and fixes.

errael:
Patch is ready for review/integration.

One thing, the help text for `foreach()`, `map()`, `filter()` has
```
When {expr2} is a Funcref errors inside a function are ignored,
unless it was defined with the "abort" flag.
```
Which I found confusing. Having done tests for `foreach()`, I have a better
sense of what this is about.

But it seems like the text should be updated for the modern era, but
uncertain about the best wordsmithing.

errael:
Link to mailing list discussion that drove and references this PR
https://groups.google.com/g/vim_dev/c/Q3S6rjBWqGY/m/6dmfhWapDAAJ

errael:
Rebase PR to todays master.

errael:
Rebase PR to today's master.

brammool:
> `foreach()` now works for a list (limited testing). A review of the
> approach is appreciated before looking at dict/blob/string.
> 
> #### One line lambda with a command
> 
> I was hoping to be able to do something like
> ```
> l->foreach((_, v) => dict1[v] += 1)
> ```
> which fails because `dict1[v] += 1` is not an expression.

You can use an inline function in Vim9 script, see
":help inline-function".  The overhead will be more than evaluating an
expression string, but that has limited functionality.

If you only pass the index or key to the function, the item in the
Dict/List needs to be looked up again.  We don't have pointers to use
for this.  If you do want to change the value of each item, then map()
would be used, thus it should not really matter much.

> #### Variable locking
> 
> Turns out that in `list_filter_map()` there is some locking; keep that
> behavior for `foreach()`. Since top level lock is an integer assignment, it
> should be cheap to keep a FILTERed list locked uless/until an element is
> removed. Is that worth adding to prevent list modification while traversing
> the list?

I would find it unexpected that foreach() changes the structure it's
iterating over.  At least adding/removing items interferes with looping
over all of them.  Changing the item values should not be a problem.
Goal is that it works predictably: Either it just works or an error is
given that the user will easily understand.

> #### Executing a string
> 
> In `filter_map_one()`, at the top, there's now
> ```
>     if (filtermap == FILTERMAP_FOREACH && expr->v_type == VAR_STRING)
>     {
> 	// foreach() is not limited to an expression
> 	do_cmdline_cmd(expr->vval.v_string);
>         ...
> ```
> Is this the best way? Or at least good enough?

How would the index or item value be used here?  It only makes sense
when the command includes an expression that uses local variables for
this.  The use seems rather limited.  Restricting this to a function
reference should keep it simple.  Unless you can think of a useful need
that foreach() would do that won't work with that.  Considering that a
function is a sequence of commands, it should always cover the case of
executing one command.

> #### Help text
> 
> In the help for `map()` and `filter()` there's
> ```
> When {expr2} is a Funcref errors inside a function are ignored,
> unless it was defined with the "abort" flag.
> ```
> It's included in the `foreach()` help, but I don't know what/why. What's
> a better way to word this?

Ignoring errors was something we did longer ago.  New functionality
added the last few years usually bails out at the first error.  I would
think that foreach() should abort when an error is found, instead of
continuing with the following items (of which there might be many,
resulting in screens full of error messages).  In case some error should
be ignored try/catch can be used, but most errors would be caused by a
mistake in the invoked commands and just need to be fixed

-- 
"Making it up?  Why should I want to make anything up?  Life's bad enough
as it is without wanting to invent any more of it."
		-- Marvin, the Paranoid Android in Douglas Adams'
		   "The Hitchhiker's Guide to the Galaxy"

errael:
_I originally posted this to the wrong PR, sorry for the dup_

Since the PR was posted, it is now complete; what I mean to say is now
**ready for review**. It's consistent with the spec for, and shares code
with: map(), mapnew() and filter(). There is help and tests for the new
functionality, and some new tests for the older functions particularly
around locking, for example exceptions if trying to modify stucture while
iterating. List/dict/blob are all handled.

Of course, if similarity with map/filter isn't important then things can
probably be simplified.

There's some curious performance data at the end

> > `foreach()` now works for a list (limited testing). A review of the
> > approach is appreciated before looking at dict/blob/string.
> >
> > #### One line lambda with a command
> >
> > I was hoping to be able to do something like
> > ```
> > l->foreach((_, v) => dict1[v] += 1)
> > ```
> > which fails because `dict1[v] += 1` is not an expression.
> 
> You can use an inline function in Vim9 script, see
> ":help inline-function". The overhead will be more than evaluating an
> expression string, but that has limited functionality.

Right. I was hoping to have a one-liner, but
```
l->foreach((_, v) => {
    found[v] = true
})
```

works fine.

> If you only pass the index or key to the function, the item in the
> Dict/List needs to be looked up again. We don't have pointers to use
> for this. If you do want to change the value of each item, then map()
> would be used, thus it should not really matter much.
> 
> > #### Variable locking
> >
> > Turns out that in `list_filter_map()` there is some locking; keep that
> > behavior for `foreach()`. Since top level lock is an integer assignment, it
> > should be cheap to keep a FILTERed list locked uless/until an element is
> > removed. Is that worth adding to prevent list modification while traversing
> > the list?
> 
> I would find it unexpected that foreach() changes the structure it's
> iterating over. At least adding/removing items interferes with looping
> over all of them. Changing the item values should not be a problem.
> Goal is that it works predictably: Either it just works or an error is
> given that the user will easily understand.

Since originally writing this, PR#12183 was merged, and resolved most
issues in my mind. An exception is thrown if foreach tries to modify
the structure of the input.

> > #### Executing a string
> >
> > In `filter_map_one()`, at the top, there's now
> > ```
> > if (filtermap == FILTERMAP_FOREACH && expr->v_type == VAR_STRING)
> > {
> > // foreach() is not limited to an expression
> > do_cmdline_cmd(expr->vval.v_string);
> > ...
> > ```
> > Is this the best way? Or at least good enough?
> 
> How would the index or item value be used here?

The example I put into help is
```
call foreach(mylist, 'found[v:val] = true')
```

> It only makes sense
> when the command includes an expression that uses local variables for
> this. The use seems rather limited. Restricting this to a function
> reference should keep it simple.

It should be easy to take out this functionality if you want.
I kept it consistent with map/mapnew/filter.

> Unless you can think of a useful need
> that foreach() would do that won't work with that. Considering that a
> function is a sequence of commands, it should always cover the case of
> executing one command.
> 
> > #### Help text
> >
> > In the help for `map()` and `filter()` there's
> > ```
> > When {expr2} is a Funcref errors inside a function are ignored,
> > unless it was defined with the "abort" flag.
> > ```
> > It's included in the `foreach()` help, but I don't know what/why. What's
> > a better way to word this?
> 
> Ignoring errors was something we did longer ago. New functionality
> added the last few years usually bails out at the first error. I would
> think that foreach() should abort when an error is found, instead of
> continuing with the following items (of which there might be many,
> resulting in screens full of error messages). In case some error should
> be ignored try/catch can be used, but most errors would be caused by a
> mistake in the invoked commands and just need to be fixed

Errors abort. At the time of writing, I was confused by the "abort" flag
which I think is legacy.

Below are profiling results of different ways to use foreach. The value
given for a cell is the result of the line executed 99 times; the length
of the list being operated on is 1, 2, 3, 6, 12, 24. There were 20 runs,
and the smallest (fastest) value selected. Each line, except `###-0`, does
some form of `add(result, val)`.

```
=== foreach builtin micro-prof, 20 run/params ===
   1x99   2x99   3x99   6x99  12x99  24x99 : nListXnLoop (usec/op)
  8.802  8.637  8.526  8.719  8.712  8.622 : var TFEC = FEC('(_, val: number): void =>: add( ###-0
  1.503  2.758  3.994  7.697 15.106 29.666 : foreach(input, 'add(result, v:val)')            ###-4
  0.175  0.291  0.409  0.749  1.401  2.709 : for l in input|add(result, l)|endfor            ###-9
  0.982  1.307  1.547  2.379  4.227  7.556 : foreach(input, VarFEC)                          ###-E
  0.572  0.898  1.115  1.901  3.674  7.100 : foreach(input, # ()=>                           ###-L
  0.669  1.187  1.610  2.944  5.758 11.616 : foreach(input, Foreach1)                        ###-R
  0.679  1.197  1.583  2.989  5.806 11.145 : foreach(input, RefForeach1)                     ###-j
  0.552  0.877  1.115  1.935  3.645  7.000 : foreach(input, ScriptFEC)                       ###-p
```
Using/evaluating a string, `###-4`, is very slow, I guess it's running
through the whole interpreter each time; maybe using a string argument
should give an error as you suggested. An inline for loop, `###-9`, is the
big winner as expected at around 2.7 usec/op.

After inline, `###-L` and `###-p` weigh in at about 7 usec/op where the
function argument is a lambda.

`###-R` and `###-j` are over 11 usec/op where the function argument is a
top level def from the script.

Looks like a regular function is around 40% slower than a lambda;
this was surprising to me.

`###-L` is
```
foreach(input, # ()=>                 ###-L
    (_, val: number): void => {
        add(result, val)
})
```

`###-p` uses ScriptFEC, which is a script level variable generated by
```
var ScriptFEC = FEC('(_, val: number): void =>: add(result, val)')
```
where `FEC` parses and compiles it's input and returns a function, it's
signature:
```
def FEC(code: string): func(any, any): void
```

`###-R` and `###-j` are defined as
```
def Foreach1(_, val: number): void # : list<number>
    add(result, val)
enddef
var RefForeach1 = Foreach1
```

brammool:
> Of course, if similarity with map/filter isn't important then things can
> probably be simplified.

Well, map() and filter() may have some ballast from old days that we
don't need to replicate with foreach().  But for users it may not be
clear what the reason is something works the way it does.  We can try to
make things work similarly, unless the difference is obvious or just
better (and can be explained).

> > > #### Executing a string
> > >
> > > In `filter_map_one()`, at the top, there's now
> > > ```
> > > if (filtermap == FILTERMAP_FOREACH && expr->v_type == VAR_STRING)
> > > {
> > > // foreach() is not limited to an expression
> > > do_cmdline_cmd(expr->vval.v_string);
> > > ...
> > > ```
> > > Is this the best way? Or at least good enough?
> > 
> > How would the index or item value be used here?
> 
> The example I put into help is
> ```
> call foreach(mylist, 'found[v:val] = true')
> ```
> 
> > It only makes sense
> > when the command includes an expression that uses local variables for
> > this. The use seems rather limited. Restricting this to a function
> > reference should keep it simple.
> 
> It should be easy to take out this functionality if you want.
> I kept it consistent with map/mapnew/filter.

Well, it's not really consistent.  It is a string, but instead of used
as an expression it is used as an Ex command.  That does make sense,
since for an expression we don't have an obvious place to store the
result.  But a user may need to read the help carefully to understand
this.

Someone using foreach() may wonder whether to use a function with the
command or use a string with the command directly.  Is there a clear
explanation for which one should be used?  Using the string is simpler,
any disadvantage?

> Errors abort. At the time of writing, I was confused by the "abort" flag
> which I think is legacy.

Correct, compiled functions always bail out at the first error, as if
the "abort" flag is always present.

> Below are profiling results of different ways to use foreach. The value
> given for a cell is the result of the line executed 99 times; the length
> of the list being operated on is 1, 2, 3, 6, 12, 24. There were 20 runs,
> and the smallest (fastest) value selected. Each line, except `###-0`, does
> some form of `add(result, val)`.
> 
> ```
> === foreach builtin micro-prof, 20 run/params ===
>    1x99   2x99   3x99   6x99  12x99  24x99 : nListXnLoop (usec/op)
>   8.802  8.637  8.526  8.719  8.712  8.622 : var TFEC = FEC('(_, val: number): void =>: add( ###-0
>   1.503  2.758  3.994  7.697 15.106 29.666 : foreach(input, 'add(result, v:val)')            ###-4
>   0.175  0.291  0.409  0.749  1.401  2.709 : for l in input|add(result, l)|endfor            ###-9
>   0.982  1.307  1.547  2.379  4.227  7.556 : foreach(input, VarFEC)                          ###-E
>   0.572  0.898  1.115  1.901  3.674  7.100 : foreach(input, # ()=>                           ###-L
>   0.669  1.187  1.610  2.944  5.758 11.616 : foreach(input, Foreach1)                        ###-R
>   0.679  1.197  1.583  2.989  5.806 11.145 : foreach(input, RefForeach1)                     ###-j
>   0.552  0.877  1.115  1.935  3.645  7.000 : foreach(input, ScriptFEC)                       ###-p
> ```
> Using/evaluating a string, `###-4`, is very slow, I guess it's running
> through the whole interpreter each time; maybe using a string argument
> should give an error as you suggested.

Yes, the command will have to be parsed every time, and a function call
used from an Ex command has quite a bit of overhead.

> An inline for loop, `###-9`, is the big winner as expected at around
> 2.7 usec/op.

The difference is bigger than I would expect.  You did execute this in a
compiled function, where the commands have been turned into byte code?
Then the function call overhead is probably the lowest.  Oh wait, the
add() call is actually inlined if the argument is known to be a list or
blob.  That must help a lot.  Using another function may give very
different results.

> After inline, `###-L` and `###-p` weigh in at about 7 usec/op where the
> function argument is a lambda.
> 
> `###-R` and `###-j` are over 11 usec/op where the function argument is a
> top level def from the script.
> 
> Looks like a regular function is around 40% slower than a lambda;
> this was surprising to me.

For using a function call, I still want to have another look at whether
the overhead can be reduced.  But there are so many other things that
compete for my time...


-- 
This planet has -- or rather had -- a problem, which was this: most
of the people living on it were unhappy for pretty much of the time.
Many solutions were suggested for this problem, but most of these
were largely concerned with the movements of small green pieces of
paper, which is odd because on the whole it wasn't the small green
pieces of paper that were unhappy.

errael:
There seems to be general agreement on `foreach()` spec.
The only remaining issue is whether or not a `string` needs
to be supported in addition to `func`/`funcref`.

>
> > ...
> > The example I put into help is
> > ```
> > call foreach(mylist, 'found[v:val] = true')
> > ```
> >
> > ...
>
> Well, it's not really consistent. It is a string, but instead of used
> as an expression it is used as an Ex command. That does make sense,
> since for an expression we don't have an obvious place to store the
> result. But a user may need to read the help carefully to understand
> this.
>
> Someone using foreach() may wonder whether to use a function with the
> command or use a string with the command directly. Is there a clear
> explanation for which one should be used? Using the string is simpler,
> any disadvantage?

Performance seems the primary disadvantage as seen in the previously
provided performance results.

I suppose optimization is possible in the future. Could the string be
compiled and saved in a map of `cache[string] = compiled_code`. Unless
such a mechanism is useful in other situations, probably not worth it
since using a function/funcref works.

But the string is more compact since with a lambda, one needs to do
```
l->foreach((_, v) => {
    found[v] = true
})
```

Which is why I put together a vim9 function,
```
def LC(code: string): func(any, any): void
```
compiles `code` returning a function which can be used like

```
foreach(mylist, LC('(_, val) =>: found[val] = true'))
```
Where LC is LambdaCompile. I used `=>:` to indicate that
- the RHS is a statement, not an expression
- only one line, can not continue to next line, `{}` block not allowed
- minimize confusion with the standard `vim9 lambda expression`

`LC()` is a hack and has the major disadvantage that  there's
no context capture. But it was fun putting together the function
that parses the input and compiles a function.

brammool:
> > Someone using foreach() may wonder whether to use a function with the
> > command or use a string with the command directly. Is there a clear
> > explanation for which one should be used? Using the string is simpler,
> > any disadvantage?
> 
> Performance seems the primary disadvantage as seen in the previously
> provided performance results.
> 
> I suppose optimization is possible in the future. Could the string be
> compiled and saved in a map of `cache[string] = compiled_code`. Unless
> such a mechanism is useful in other situations, probably not worth it
> since using a function/funcref works.

Compiling has most effect for commands related to execution flow, such
as if/while/try/etc. and expressions.  Otherwise most Ex commands are
not compiled currently.  One that might be useful is assignment.  But
only doing that for this specific use case is quite a lot of work with
little gain.  And when the string is not a constant compilation isn't
feasible anyway.

> But the string is more compact since with a lambda, one needs to do
> ```
> l->foreach((_, v) => {
>     found[v] = true
> })
> ```
> 
> Which is why I put together a vim9 function,
> ```
> def LC(code: string): func(any, any): void
> ```
> compiles `code` returning a function which can be used like
> 
> ```
> foreach(mylist, LC('(_, val) =>: found[val] = true'))
> ```
> Where LC is LambdaCompile. I used `=>:` to indicate that
> - the RHS is a statement, not an expression
> - only one line, can not continue to next line, `{}` block not allowed
> - minimize confusion with the standard `vim9 lambda expression`
> 
> `LC()` is a hack and has the major disadvantage that  there's
> no context capture. But it was fun putting together the function
> that parses the input and compiles a function.

It's a nice trick, but not without disadvantages.  The example you
started with isn't really so bad, it's all in one place.  It may have
two line breaks but otherwise no extra characters other than the {}.

-- 
Due knot trussed yore spell chequer two fined awl miss steaks.

errael:
Rebase PR to todays master.

errael:
Fix for `test_codestyle`.

errael:
Rebase PR to today's master.

This PR adds a new builtin. Up to this this point, review has mostly been conceptual. Some in this PR and notice there's a link to discussion in vimgroup. 

This is not urgent.

errael:
Rebase to current master

errael:
Rebased to current master.

**History of this PR**

Adds `foreach()` which is similar to `filter()`, `map()` and `mapnew()`  (and shares their implementation). Each of the 4 has the same argument signature: `(list|string|blob|dict, funcref|string)`

The difference between `foreach()` and the rest is that 
- the original list is never modified
- `{expr2}` is executed, not evaluated


The forum discussion that led to this PR is 
https://groups.google.com/d/msgid/vim_dev/b8e2b8fa-c6d8-cc41-ac63-6b1d21a75f41%40raelity.com

This PR has some additional tests for the existing `filter()` like functions, particularly for when `{expr1}` is locked.

While testing/debugging for this PR ran into a crash in released code fixed by https://github.com/vim/vim/commit/e6d40dcdc7227594935d2db01eca29f0e575dcee

The discussion with Bram in this PR's comments took place in May of 2023.

The PR has not been modified, other than rebase, since March of 2023. (well, there was a minor change in June,
https://github.com/vim/vim/compare/7ee5f12638ef84844eb6d560f71277cfb846aa08..6937bd3ff5cda356f82dda6e19f72286de8603d0, 
because `test_codestyle` did more checking)

errael:
For help do `make tags`.

chrisbra:
thanks. Let me include this.

PR #12178, opened 2023-03-18
Cython files are not recognized

amaanq:
https://cython.org/

Problem: Cython files are not recognized.
Solution: Add a pattern for Cython files. (Amaan Qureshi)

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -448,6 +448,9 @@ au BufNewFile,BufRead *.cu,*.cuh		setf cuda
 " Cue
 au BufNewFile,BufRead *.cue			setf cue
 
+" Cython
+au BufNewFile,BufRead *.pyx,*.pxd,*.pxi		setf cython
+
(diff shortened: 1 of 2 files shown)


brammool:
Conflicts with Pyrex patterns.  Perhaps the difference can be seen in the file contents?

amaanq:
Both Pyrex and Cython are extensions of Python to interop with C, syntax will be very similar to Python and I don't know of an easier way to distinguish them. Cython seems to be used far more in my opinion, last stable release of Pyrex was 12 years ago and seems to be hardly used (I hadn't heard of it until you pointed it out).

chrisbra:
You could add a `filetype_pyx` variable and simply document that the user should set those variables to have `cython` files instead of `pyrex`, see https://github.com/vim/vim/blob/4b3fab14dbde971f15d8783e9ef125b19fdbc829/runtime/doc/filetype.txt#L135-L142

chrisbra:
closing for inactivity

PR #12179, opened 2023-03-18
Extend the reverse() function to support strings

yegappan:
The reverse() function supports reversing a List or a Blob.  Extend it to support reversing strings.  The function returns a copy of the reversed string.


runtime/doc/builtin.txt
```diff
@@ -7375,9 +7376,9 @@ resolve({filename})					*resolve()* *E655*
 
 reverse({object})					*reverse()*
 		Reverse the order of items in {object} in-place.
```
user202729: this doesn't look in-place for string anymore...?

jdjohnston:
@yegappan Nice addition, but the error message (E899) still says the argument of reverse() must be a List or a Blob

yegappan:
@jdjohnston I have created the PR #12377 to use the proper error message.

PR #12183, opened 2023-03-19
Fix SEGV when modifying collection being filtered.

errael:
Lock the collection during `filter()`.
Also lock blob when `map()`; seems to have been missing.
Add some tests.

---

Same fix for list/dict/blob; they all got SEGV.
Needed some additional work for blob since it didn't protect
for modifications during map().

The fix changes
```
    if (filtermap != FILTERMAP_FILTER && l->lv_lock == 0)
	l->lv_lock = VAR_LOCKED;
```
to
```
    if (l->lv_lock == 0)
	l->lv_lock = VAR_LOCKED;
```
during the start of `list_filter_map()`. Similarly for `dict`/`blob`.

The fix doesn't run into problems because with
```
    else if (filtermap == FILTERMAP_FILTER && rem)
        listitem_remove(l, li);
```
since `listitem_remove()` doesn't check locks.

I tried a variety of things, and things seem to behave
with this PR.
Not certain why FILTERMAP_FILTER was excluding from
locking the collection.
There may be other cases I didn't consider that could
run into problems, but I couldn't find them.

The following got a SEGV for me. Similar scripts for `dict`/`blob`.
```
vim9script
def FilterRemoveTest2()
    var l = [1]
    l->add(777)

    def FilterRemove(_, _): bool
        while l->len() > 0
            l->remove(0)
        endwhile
        return false
    enddef
    l->filter(FilterRemove)
enddef
FilterRemoveTest2()
```

With this PR, a lock error is reported:

```
Error detected while processing command line
    ..script /home/err/play/xxx.vim[14]
    ..function <SNR>34_FilterRemoveTest2[10]..<lambda>1:
line    2:
E741: Value is locked: remove() argument
```
---

With the fix, using range() as in
```
vim9script
def FilterRemoveTest2()
    var l = range(4)
    ...
```

doesn't get the lock error or the SEGV. I guess this is an
implementation detail; doesn't seem right.

### DIFF
--- a/src/blob.c
+++ b/src/blob.c
@@ -592,8 +592,10 @@ blob_filter_map(
 	blob_T		*blob_arg,
 	filtermap_T	filtermap,
 	typval_T	*expr,
+	char_u		*arg_errmsg,
 	typval_T	*rettv)
 {
+    int		prev_lock;
     blob_T	*b;
     int		i;
     typval_T	tv;
@@ -609,8 +611,11 @@ blob_filter_map(
 	rettv->v_type = VAR_BLOB;
 	rettv->vval.v_blob = NULL;
     }
-    if ((b = blob_arg) == NULL)
+    if ((b = blob_arg) == NULL || (filtermap == FILTERMAP_FILTER
+			    && value_check_lock(b->bv_lock, arg_errmsg, TRUE)))
 	return;
+
+    prev_lock = b->bv_lock;
 
     b_ret = b;
     if (filtermap == FILTERMAP_MAPNEW)
@@ -622,6 +627,9 @@ blob_filter_map(
 
     // set_vim_var_nr() doesn't set the type
     set_vim_var_type(VV_KEY, VAR_NUMBER);
+
+    if (b->bv_lock == 0)
+	b->bv_lock = VAR_LOCKED;
 
     // Create one funccal_T for all eval_expr_typval() calls.
     fc = eval_expr_get_funccal(expr, &newtv);
@@ -658,6 +666,7 @@ blob_filter_map(
 	++idx;
     }
 
+    b->bv_lock = prev_lock;
     if (fc != NULL)
 	remove_funccal();
 }
--- a/src/dict.c
+++ b/src/dict.c
@@ -1365,7 +1365,7 @@ dict_filter_map(
     // Create one funccal_T for all eval_expr_typval() calls.
     fc = eval_expr_get_funccal(expr, &newtv);
 
-    if (filtermap != FILTERMAP_FILTER && d->dv_lock == 0)
+    if (d->dv_lock == 0)
 	d->dv_lock = VAR_LOCKED;
     ht = &d->dv_hashtab;
     hash_lock(ht);
--- a/src/list.c
+++ b/src/list.c
@@ -2398,7 +2398,7 @@ list_filter_map(
     // set_vim_var_nr() doesn't set the type
     set_vim_var_type(VV_KEY, VAR_NUMBER);
 
-    if (filtermap != FILTERMAP_FILTER && l->lv_lock == 0)
+    if (l->lv_lock == 0)
 	l->lv_lock = VAR_LOCKED;
 
     // Create one funccal_T for all eval_expr_typval() calls.
@@ -2578,7 +2578,8 @@ filter_map(typval_T *argvars, typval_T *rettv, filtermap_T filtermap)
 	dict_filter_map(argvars[0].vval.v_dict, filtermap, type, func_name,
 		arg_errmsg, expr, rettv);
     else if (argvars[0].v_type == VAR_BLOB)
-	blob_filter_map(argvars[0].vval.v_blob, filtermap, expr, rettv);
+	blob_filter_map(argvars[0].vval.v_blob, filtermap, expr,
+			arg_errmsg, rettv);
     else if (argvars[0].v_type == VAR_STRING)
 	string_filter_map(tv_get_string(&argvars[0]), filtermap, expr,
 		rettv);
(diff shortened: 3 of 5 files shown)


errael:
Since the PR was posted, it is now complete; what I mean to say is now
**ready for review**. It's consistent with the spec for, and shares code
with: map(), mapnew() and filter(). There is help and tests for the new
functionality, and some new tests for the older functions particularly
around locking, for example exceptions if trying to modify stucture while
iterating. List/dict/blob are all handled.

Of course, if similarity with map/filter isn't important then things can
probably be simplified.

There's some curious performance data at the end

> > `foreach()` now works for a list (limited testing). A review of the
> > approach is appreciated before looking at dict/blob/string.
> >
> > #### One line lambda with a command
> >
> > I was hoping to be able to do something like
> > ```
> > l->foreach((_, v) => dict1[v] += 1)
> > ```
> > which fails because `dict1[v] += 1` is not an expression.
> 
> You can use an inline function in Vim9 script, see
> ":help inline-function". The overhead will be more than evaluating an
> expression string, but that has limited functionality.

Right. I was hoping to have a one-liner, but
```
l->foreach((_, v) => {
    found[v] = true
})
```

works fine.

> If you only pass the index or key to the function, the item in the
> Dict/List needs to be looked up again. We don't have pointers to use
> for this. If you do want to change the value of each item, then map()
> would be used, thus it should not really matter much.
> 
> > #### Variable locking
> >
> > Turns out that in `list_filter_map()` there is some locking; keep that
> > behavior for `foreach()`. Since top level lock is an integer assignment, it
> > should be cheap to keep a FILTERed list locked uless/until an element is
> > removed. Is that worth adding to prevent list modification while traversing
> > the list?
> 
> I would find it unexpected that foreach() changes the structure it's
> iterating over. At least adding/removing items interferes with looping
> over all of them. Changing the item values should not be a problem.
> Goal is that it works predictably: Either it just works or an error is
> given that the user will easily understand.

Since originally writing this, PR#12183 was merged, and resolved most
issues in my mind. An exception is thrown if foreach tries to modify
the structure of the input.

> > #### Executing a string
> >
> > In `filter_map_one()`, at the top, there's now
> > ```
> > if (filtermap == FILTERMAP_FOREACH && expr->v_type == VAR_STRING)
> > {
> > // foreach() is not limited to an expression
> > do_cmdline_cmd(expr->vval.v_string);
> > ...
> > ```
> > Is this the best way? Or at least good enough?
> 
> How would the index or item value be used here?

The example I put into help is
```
call foreach(mylist, 'found[v:val] = true')
```

> It only makes sense
> when the command includes an expression that uses local variables for
> this. The use seems rather limited. Restricting this to a function
> reference should keep it simple.

It should be easy to take out this functionality if you want.
I kept it consistent with map/mapnew/filter.

> Unless you can think of a useful need
> that foreach() would do that won't work with that. Considering that a
> function is a sequence of commands, it should always cover the case of
> executing one command.
> 
> > #### Help text
> >
> > In the help for `map()` and `filter()` there's
> > ```
> > When {expr2} is a Funcref errors inside a function are ignored,
> > unless it was defined with the "abort" flag.
> > ```
> > It's included in the `foreach()` help, but I don't know what/why. What's
> > a better way to word this?
> 
> Ignoring errors was something we did longer ago. New functionality
> added the last few years usually bails out at the first error. I would
> think that foreach() should abort when an error is found, instead of
> continuing with the following items (of which there might be many,
> resulting in screens full of error messages). In case some error should
> be ignored try/catch can be used, but most errors would be caused by a
> mistake in the invoked commands and just need to be fixed

Errors abort. At the time of writing, I was confused by the "abort" flag
which I think is legacy.

Below are profiling results of different ways to use foreach. The value
given for a cell is the result of the line executed 99 times; the length
of the list being operated on is 1, 2, 3, 6, 12, 24. There were 20 runs,
and the smallest (fastest) value selected. Each line, except `###-0`, does
some form of `add(result, val)`.

```
=== foreach builtin micro-prof, 20 run/params ===
   1x99   2x99   3x99   6x99  12x99  24x99 : nListXnLoop (usec/op)
  8.802  8.637  8.526  8.719  8.712  8.622 : var TFEC = FEC('(_, val: number): void =>: add( ###-0
  1.503  2.758  3.994  7.697 15.106 29.666 : foreach(input, 'add(result, v:val)')            ###-4
  0.175  0.291  0.409  0.749  1.401  2.709 : for l in input|add(result, l)|endfor            ###-9
  0.982  1.307  1.547  2.379  4.227  7.556 : foreach(input, VarFEC)                          ###-E
  0.572  0.898  1.115  1.901  3.674  7.100 : foreach(input, # ()=>                           ###-L
  0.669  1.187  1.610  2.944  5.758 11.616 : foreach(input, Foreach1)                        ###-R
  0.679  1.197  1.583  2.989  5.806 11.145 : foreach(input, RefForeach1)                     ###-j
  0.552  0.877  1.115  1.935  3.645  7.000 : foreach(input, ScriptFEC)                       ###-p
```
Using/evaluating a string, `###-4`, is very slow, I guess it's running
through the whole interpreter each time; maybe using a string argument
should give an error as you suggested. An inline for loop, `###-9`, is the
big winner as expected at around 2.7 usec/op.

After inline, `###-L` and `###-p` weigh in at about 7 usec/op where the
function argument is a lambda.

`###-R` and `###-j` are over 11 usec/op where the function argument is a
top level def from the script.

Looks like a regular function is around 40% slower than a lambda;
this was surprising to me.

`###-L` is
```
foreach(input, # ()=>                 ###-L
    (_, val: number): void => {
        add(result, val)
})
```

`###-p` uses ScriptFEC, which is a script level variable generated by
```
var ScriptFEC = FEC('(_, val: number): void =>: add(result, val)')
```
where `FEC` parses and compiles it's input and returns a function, it's
signature:
```
def FEC(code: string): func(any, any): void
```

`###-R` and `###-j` are defined as
```
def Foreach1(_, val: number): void # : list<number>
    add(result, val)
enddef
var RefForeach1 = Foreach1
```

PR #12187, opened 2023-03-21
runtime/syntax/gpg.vim: sync with gnupg-2.4.0

tmzullinger:
A number of options have been added since the last update.  Take the output from `gpg --dump-option-table` and manually filter the commands which take args from those which do not.

For reference, this is what was run:

    gpg --dump-option-table | awk -F: '{printf "%-30s %s\n", $1, $3}' |
        sort >/tmp/gpg-options
    grep -Evw '(0|64|128|25[68]|512)$' /tmp/gpg-options |
        awk '{print $1}' >/tmp/gpg-options-arg
    awk '/\<0$/ {print $1}' /tmp/gpg-options >/tmp/gpg-options-noarg
    grep -Ew '(128|25[68])$' /tmp/gpg-options |
        awk '{print $1}' >/tmp/gpg-commands

Unlike the gpg options, the gpg commands were manually split for the `gpgArg` and `gpgArgError` groups, based on man gpg(1) and the `opts` array in `g10/gpg.c`.

I noticed that the last update to the syntax was done via a pull request, so I'm hoping this isn't too far afield from the proper method described in CONTRIBUTING.  As with the previous submission, I'm not confident I have the time to properly maintain the syntax file, unfortunately.

Hopefully this update which brings support for the options and commands in the current gnupg-2.4.0 will serve other Vim users well for some time.

I tried to do as much of it programatically as I could.  Without changes in gnupg, I don't know if there's a lot more that can be automated with respect to parsing the `gpg --dump-option-table` output.  This does include some keywords which are unlikely to be used in a config file -- but unlikely for me may be a common use case for another person.  Including all of the keywords from `--dump-options-table` might make it easier to sync and update the syntax in the future.

Using `git show --color-words` provides a decent method of reviewing the keywords added, changed, and removed.

Thanks!

### DIFF
--- a/runtime/syntax/gpg.vim
+++ b/runtime/syntax/gpg.vim
@@ -2,7 +2,7 @@
 " Language:		gpg(1) configuration file
 " Previous Maintainer:	Nikolai Weibull <now@bitwi.se>
 " Latest Revision:	2010-10-14
-" Updated:		2023-01-23 @ObserverOfTime: added a couple of keywords
+" Updated:		2023-03-21 Todd Zullinger <tmz@pobox.com>: sync with gnupg-2.4.0
 
 if exists("b:current_syntax")
   finish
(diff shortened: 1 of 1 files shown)


brammool:
Thanks, I'll include it.

PR #12188, opened 2023-03-21
Source `.exrc`/`.vimrc` in current work dir with -u flag

veselyn:
Vim does not respect the `set exrc` option when included in a custom vimrc that is provided with the `-u` flag.

The proposed change is to load the files from the current directory if the value of the `-u` flag is a custom vimrc with `set exrc` inside and not "DEFAULTS", "NONE" or "NORC".

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -3288,7 +3288,7 @@ source_startup_scripts(mparm_T *parmp)
      * Only do this if VIMRC_FILE is not the same as USR_VIMRC_FILE or
      * SYS_VIMRC_FILE.
      */
-    if (p_exrc && use_exrc)
+    if (!silent_mode && p_exrc && use_exrc)
     {
 #if defined(UNIX) || defined(VMS)
 	// If ".vimrc" file is not owned by user, set 'secure' mode.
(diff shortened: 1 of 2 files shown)


brammool:
The use of the 'exrc' option is tricky.  Many users will not realize how dangerous this is, or have it set to try out something and forget to remove it again.  So I rather not make it used more.  The only reason the option exists is for Vi compatibility, I would not expect it to be used for any other reason.

I'm missing the reason you want to make this change.  Have you considered alternatives?

v3s1e:
> The use of the 'exrc' option is tricky. Many users will not realize how dangerous this is, or have it set to try out something and forget to remove it again. So I rather not make it used more. The only reason the option exists is for Vi compatibility, I would not expect it to be used for any other reason.
> 
> I'm missing the reason you want to make this change. Have you considered alternatives?

I must admit that the reason for my pull request was that I believed this behavior to be a bug or oversight in Vim's implementation. However, I understand now that the 'exrc' option should be used with caution and only for Vi compatibility purposes.

I will take your advice to heart and explore alternatives to this proposed change. I appreciate your guidance on this matter.

Thank you for taking the time to share your thoughts and for your work on Vim. Your contributions to the community are greatly appreciated!

PR #12189, opened 2023-03-22
Find Python 3 dynamic library in macOS in configure script

ychin:
Currently, using `./configure --enable-python3interp=dynamic` does not really work properly on macOS for most Python 3 installations because most Python installations use a "framework" mode of installations that's unique to macOS. This is enabled when configuring CPython with `--enable-framework` flag (see
https://docs.python.org/3/using/configure.html).

When using framework mode, the dynamic library's path works a little differently. The full path for Homebrew installation is `/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/Python` and Xcode version is
`/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.9/Python3`. When Vim's configure script tries to use just `INSTSONAME` for setting the DLL path (will be used as the default for `pythonthreedll`), it will just grab the part that says `Python.framework/Versions/3.11/Python`. The rest of the path is encoded in a Makefile var called `PYTHONFRAMEWORKPREFIX` which contains the
`/opt/homebrew/opt/python@3.11/Frameworks` part.

The CPython logic for configuring these is in
https://github.com/python/cpython/blob/3.11/configure.ac, where we can see how the paths are constructed:

```
LDLIBRARY='$(PYTHONFRAMEWORKDIR)/Versions/$(VERSION)/$(PYTHONFRAMEWORK)'
LDSHARED="$LDSHARED "'$(PYTHONFRAMEWORKPREFIX)/$(PYTHONFRAMEWORKDIR)/Versions/$(VERSION)/$(PYTHONFRAMEWORK)'
```

This change makes sure to use `PYTHONFRAMEWORKPREFIX` to help construct a full path for the lib. With this change, I can now just configure as normal, and it seems to be able to find Python 3's dynamic library correctly.

I don't think this affects any non-macOS builds because `--enable-framework` should be a macOS-only configure option. Note that installations like miniconda on macOS do not use the framework mode, but they also won't be affected because `PYTHONFRAMEWORKPREFIX` will be empty just like other platforms.

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -6860,6 +6860,7 @@ __:
 	@echo "python3_SYSLIBS='$(SYSLIBS)'"
 	@echo "python3_DLLLIBRARY='$(DLLLIBRARY)'"
 	@echo "python3_INSTSONAME='$(INSTSONAME)'"
+	@echo "python3_PYTHONFRAMEWORKPREFIX='$(PYTHONFRAMEWORKPREFIX)'"
 eof
 	    	    eval "`cd ${PYTHON3_CONFDIR} && make -f "${tmp_mkf}" __ | sed '/ directory /d'`"
 	    rm -f -- "${tmp_mkf}"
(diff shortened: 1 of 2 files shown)


ychin:
Just for reference, this is why MacVim has always had to manually specify the Python library location in CI using `vi_cv_dll_name_python3 `: https://github.com/macvim-dev/macvim/blob/86af12a7ee802331357bebe7afa6431d91253ad8/.github/workflows/ci-macvim.yaml#L31-L32

PR #12190, opened 2023-03-22
Problem: Hollywood syntax highlighting is outdated.

sodero:
Solution: Update syntax highlighting to Hollywood version 10.


brammool:
I'll include it, thanks.

PR #12192, opened 2023-03-24
Improve profiling results.

errael:
Reduce overhead of checking if a function should be profiled, by caching results of checking (which are done with regexp). Cache cleared when regexps are changed.
Break at first match for has_profiling lookup.

### DIFF
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -528,6 +528,35 @@ static int has_expr_breakpoint = FALSE;
 #ifdef FEAT_PROFILE
 // Profiling uses file and func names similar to breakpoints.
 static garray_T prof_ga = {0, 0, sizeof(struct debuggy), 4, NULL};
+
+// Profiling caches results of regexp lookups for function/script name.
+#define N_PROF_HTAB 2
+static hashtab_T prof_cache[N_PROF_HTAB];
+#define PROF_HTAB_FUNCS 0
+#define PROF_HTAB_FILES 1
+static int prof_cache_initialized;
+typedef struct profentry_S
+{
+    char    pen_flags;	// cache data booleans: profiling, forceit
+    char_u  pen_name[1]; // actually longer
+} profentry_T;
+#define PEN_FLAG_PROFILING 1
+#define PEN_FLAG_FORCEIT 2
+#define PEN_SET_PROFILING(pe) ((pe)->pen_flags |= PEN_FLAG_PROFILING)
+#define PEN_SET_FORCEIT(pe) ((pe)->pen_flags |= PEN_FLAG_FORCEIT)
+#define PEN_IS_PROFILING(pe) (((pe)->pen_flags & PEN_FLAG_PROFILING) != 0)
+#define PEN_IS_FORCEIT(pe) (((pe)->pen_flags & PEN_FLAG_FORCEIT) != 0)
+
+#define PE2HIKEY(pe)	((pe)->pen_name)
+#define HIKEY2PE(p)	((profentry_T *)((p) - (offsetof(profentry_T, pen_name))))
+#define HI2PE(hi)	HIKEY2PE((hi)->hi_key)
+
+static void prof_clear_cache(void);
+#define PROF_CLEAR_CACHE(gap) do {if ((gap) == &prof_ga) prof_clear_cache();} while (0)
+// Can enable to get some info about profile caching
+// #define PROF_CACHE_LOG
+#else
+#define PROF_CLEAR_CACHE(gap) do {} while (0)
 #endif
 #define DBG_FUNC	1
 #define DBG_FILE	2
@@ -708,6 +737,7 @@ ex_breakadd(exarg_T *eap)
 		++debug_tick;
 	    }
 	    ++gap->ga_len;
+	    PROF_CLEAR_CACHE(gap);
 	}
     }
     else
@@ -845,6 +875,7 @@ ex_breakdel(exarg_T *eap)
 	if (!del_all)
 	    break;
     }
+    PROF_CLEAR_CACHE(gap);
 
     // If all breakpoints were removed clear the array.
     if (gap->ga_len == 0)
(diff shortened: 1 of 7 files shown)


errael:
Add `uf_hash` to `ufunc_S`.
Did not add hash to `scriptitem_T`. (easy to do)
Didn't do anything in `copy_lambda_to_global` for `uf_hash`.
Left `TODO` note in vim9class about possible `uf_hash` stuff.
Close to 10% improvement by caching `uf_hash`.

errael:
Link to mailing list discussion that drove and references this PR
https://groups.google.com/g/vim_dev/c/jnrN7ufFrBY/m/Emz8dY3JAQAJ?pli=1

errael:
Rebase PR to todays master.

errael:
Rebase PR to today's master. Here's some profiling results while testing the rebase. Notice larger measured empty-function-call before this change, in usec/op. It's interesting to see how the number of args affects the results.

| before  | after  | function call             |
| ------- | ------ | ------------------------- |
| 0.946   | 0.142  | NullFunction0()           |
| 0.990   | 0.179  | NullFunction2(1, 2)       |
| 1.029   | 0.215  | NullFunction4(1, 2, 3, 4) |

errael:
> Here are a few remarks.

Thanks, should be clean now.

> + hash = *hashp;
> + if (!hash)
> 
> Although I'm not sure why checking for zero works here, a hash value can
> be any number, including zero (e.g. an empty string).

Is it worth having a field in ufunc_T for "name hash computed"? Doesn't seem like it. Especially in this case when it's ulikely (impossible?) for name to be the empty string. Worst case compute the hash each time, and the empty string's hash is fast.

brammool:
> > Here are a few remarks.
> 
> Thanks, should be clean now.
> 
> > + hash = *hashp;
> > + if (!hash)
> > 
> > Although I'm not sure why checking for zero works here, a hash value can
> > be any number, including zero (e.g. an empty string).
> 
> Is it worth having a field in ufunc_T for "name hash computed"?
> Doesn't seem like it. Especially in this case when it's ulikely
> (impossible?) for name to be the empty string. Worst case compute the
> hash each time, and the empty string's hash is fast.

Well, quite often when you think "not worth taking into account" it
turns out it happens quite often.  Unfortunately, since the hash value
is unsigned, using -1 for "not computed" won't work.  We could use
ULONG_MAX, but it's not really better than using zero.  As you mention,
an empty name won't happen, thus I suppose using zero is still te best.

-- 
If they don't keep on exercising their lips, he thought, their brains
start working.

errael:
Rebase PR to todays master.

errael:
Fix for `test_codestyle`.

errael:
Rebase PR to today's master.

Lucky to have gotten a review: https://github.com/vim/vim/pull/12192#issuecomment-1540185641

errael:
Requesting consideration to move this PR to milestone 9.1

When I first saw this in this PR's comments:
```
modified the milestones: [vim-9.1](https://github.com/vim/vim/milestone/1), [backlog](https://github.com/vim/vim/milestone/2)
```
I thought it was targeting 9.1. But it doesn't seem so.

I've been using this PR since it was first written. Without it, the profiling results of calling a user defined function can be 1 to 2 orders of magnitude greater than expected due to overhead in checking if a function should be profiled. This problem became obvious after profiling started using the nsec clock and "micro-profiling" became possible. Without this patch there can be a bunch of regex run on the name function name each time the function is called.

errael:
@yegappan , @chrisbra I posted https://github.com/vim/vim/pull/12192#issuecomment-1678464559 about the possibility of moving this to 9.1. Didn't see it in mailing list, so pinging. BTW, the message https://groups.google.com/g/vim_dev/c/jnrN7ufFrBY/m/-qvZcd_xAQAJ shows early results of this change.

chrisbra:
sorry, this doesn't seem to risky to include now.

errael:
> sorry, this doesn't seem to risky to include now.

@chrisbra Does the statement have a typo? (Unless it's directed to @yegappan)

errael:
Rebased to current master

@chrisbra , @yegappan I got the impression this was now OK for 9.1. Is there anything I need to do to facilitate its inclusion?

errael:
Rebase to current master

chrisbra:
sorry, yeah this was a typo. Unless other members approve, I'd rather leave this to after a release, thus keep the `backlog` label for now.

errael:
> Unless other members approve, I'd rather leave this to after a release, thus keep the `backlog` label for now.

@mattn @chrisbra @k-takata @dpelle @yegappan @zeertzjq 

Looking for vim members interested in accurate profile results. This PR was guided to this form by Bram (see discussion); it's a companion/followup to #12129 and https://github.com/vim/vim/commit/076de79ad832558267b3ff903c048df2f4c1a5d6 . This PR was opened a week after that first PR. I use it regularly, the latest was this morning to produce these results.
```
=== assign micro-prof ===
     50    100    200 : nLoops (usec/op)
  0.044  0.044  0.043 : var local = n_value            ###-A
  0.036  0.035  0.034 : var local = C.static_var       ###-D
  0.050  0.047  0.046 : var local = a_class.this_var   ###-G
  0.173  0.170  0.164 : var local = a_class.Getter()   ###-J
```

Without this PR, it is not possible to get accurate results for calling a user function; I've seen results up to almost 2 orders of magnitude slower than what seems to be actual. This PR is "risky" and I have no idea why. It can't be because of the possibility of giving inaccurate results.

Hoping some member(s) have the interest and bandwidth to review this. I'd like to provide the infrastructure, .vim file and some awk/shell scripts, that produce tables like the result above (it took a few minutes to put the test together). I'm going to introduce it with "runs on vim version #.#", and not with "first apply this patch".

chrisbra:
```
=== assign micro-prof ===
     50    100    200 : nLoops (usec/op)
  0.044  0.044  0.043 : var local = n_value            ###-A
  0.036  0.035  0.034 : var local = C.static_var       ###-D
  0.050  0.047  0.046 : var local = a_class.this_var   ###-G
  0.173  0.170  0.164 : var local = a_class.Getter()   ###-J
```
How did you create those? I just checked the PR and I did not notice any issues, but my profiling results looked more like the traditional profiling results

errael:
I have a vim9script that run the profiler many times, and then some shell/gawk that process and summarize the results into a table.

errael:
Oh, and when I said
> I'd like to provide the infrastructure, .vim file and some awk/shell scripts,
> that produce tables like the result above (it took a few minutes to put the test
> together).

The infrastructure I'm talking about is these scripts which take the results from the builtin vim profiler.

errael:
Rebased to current master.

chrisbra:
thanks, let me merge this. Seems okay.

PR #12203, opened 2023-03-27
Add filetype detection for Livebook files

majjoha:
There is not yet any filetype detection in Vim for [Livebook](https://livebook.dev/) files, so I thought it would be worthwhile adding.

Please, let me know if anything is missing.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1162,6 +1162,9 @@ au BufNewFile,BufRead *.lite,*.lt		setf lite
 " LiteStep RC files
 au BufNewFile,BufRead */LiteStep/*/*.rc		setf litestep
 
+" Livebook
+au BufNewFile,BufRead *.livemd			setf markdown
+
 " Login access
 au BufNewFile,BufRead */etc/login.access	setf loginaccess
 
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -359,7 +359,7 @@ let s:filename_checks = {
     \ 'manconf': ['/etc/man.conf', 'man.config', 'any/etc/man.conf'],
     \ 'map': ['file.map'],
     \ 'maple': ['file.mv', 'file.mpl', 'file.mws'],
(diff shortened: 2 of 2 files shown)


majjoha:
@brammool: Good point. When referencing how other projects have added Livebook support, it seems that they are generally doing it by associating them with Markdown files as they are simply a subset of Markdown. This is, for instance, what GitHub have done with [linguist](https://github.com/github/linguist) in https://github.com/github/linguist/pull/5672, and [Gitea](https://github.com/go-gitea/gitea) goes about it in a similar way in https://github.com/go-gitea/gitea/pull/22730/. When searching GitHub for how people are handling Livebook files in their Vim configurations, it seems that most people are approaching it in this manner too, as you can see from this search query [here](https://github.com/search?q=bufread+livemd&type=code).

I've updated the pull request, so `*.livemd` files are now listed under Markdown in the tests.

brammool:
> @brammool: Good point. When referencing how other projects have added
> Livebook support, it seems that they are generally doing it by
> associating them with Markdown files as they are simply a subset of
> Markdown. This is, for instance, what GitHub have done with
> [linguist](https://github.com/github/linguist) in
> https://github.com/github/linguist/pull/5672, and
> [Gitea](https://github.com/go-gitea/gitea) goes about it in a similar
> way in https://github.com/go-gitea/gitea/pull/22730/. When searching
> GitHub for how people are handling Livebook files in their Vim
> configurations, it seems that most people are approaching it in this
> manner too, as you can see from this search query
> [here](https://github.com/search?q=bufread+livemd&type=code).
> 
> I've updated the pull request, so `*.livemd` files are now listed
> under Markdown in the tests.

I prefer using a specific filetype, because there may be a reason to do
something else than markdown at some point.  The livebook
indent/syntax/etc plugins can load the markdown plugins to make it work
at first.  If we would call *.livemd files markdown, then it will be
very difficult to make a difference later.

Let me include it like that now. Feel free to send the livebook plugins
that redirect to markdown.

-- 
A computer programmer is a device for turning requirements into
undocumented features.  It runs on cola, pizza and Dilbert cartoons.
					Bram Moolenaar

PR #12209, opened 2023-03-29
runtime: recognize .busted files as lua filetype

craigmac:


src/testdir/test_filetype.vim
```diff
@@ -90,6 +90,7 @@ let s:filename_checks = {
     \ 'blueprint': ['file.blp'],
     \ 'bsdl': ['file.bsd', 'file.bsdl'],
     \ 'bst': ['file.bst'],
+    \ 'busted': ['.busted'],
```
zeertzjq: Shouldn't this be added to the `'lua': ` below?

brammool:
Are these file literally called ".busted"?  So you can only have one per directory, and they are hidden files?

craigmac:
> Are these file literally called ".busted"? So you can only have one per directory, and they are hidden files?

From what I understand in the [usage](https://lunarmodules.github.io/busted/) section, the config files are literally `.busted`, I'm presuming they are hidden (with leading '.'), and it seems you create one in the root (doesn't say whether it supports more than one per directory).

brammool:
> > Are these file literally called ".busted"? So you can only have one
> > per directory, and they are hidden files?
> 
> From what I understand in the
> [usage](https://lunarmodules.github.io/busted/) section, the config
> files are literally `.busted`, I'm presuming they are hidden (with
> leading '.'), and it seems you create one in the root (doesn't say
> whether it supports more than one per directory).

OK, the section "Predefined Busted Tasks" explains it.

The test fails, since ".busted" gets filetype "lua", but the test
expects the filetype "busted".  So should the filetype detection be
adjusted or the test?

A quick look at the docs suggest that this .busted file has Lua syntax,
but follows test specs with a certain format.  What we often do in such
cases is use a specific filetype, and have the indent/syntax/filetype
plugin files load the more generic file type plugins.  That allows for
tuning to this specific filetype when needed.

-- 
My Go, this amn keyboar oesn't have a .

craigmac:
@brammool I've changed it now to just put `.busted` in the correct spot in the test file, sorry about that, it's completely lua, no extra need for distinct indent/syntax/ftplugin rules.

PR #12220, opened 2023-04-02
perf(runtime): fix strace regex

fedemengo:
I think this regex can be simplified with no undesired effects. 

When opening text file with long lines of numbers (bitstrings in my case) the additional and optional `[0-9:.]* *` put unnecessary work on the regex engine. 

Honestly I don't see any reason why the regex should have the additional pattern. The two tests works fine with the simplified version

### DIFF
--- a/runtime/autoload/dist/script.vim
+++ b/runtime/autoload/dist/script.vim
@@ -338,7 +338,7 @@ def DetectFromText(line1: string)
     set ft=virata
 
     # Strace
-  elseif line1 =~ '[0-9:.]* *execve(' || line1 =~ '^__libc_start_main'
+  elseif line1 =~ 'execve(' || line1 =~ '^__libc_start_main'
     set ft=strace
 
     # VSE JCL
--- a/runtime/autoload/dist/script.vim
+++ b/runtime/autoload/dist/script.vim
@@ -338,7 +338,8 @@ def DetectFromText(line1: string)
     set ft=virata
 
     # Strace
-  elseif line1 =~ 'execve(' || line1 =~ '^__libc_start_main'
+  elseif (line1 =~ 'execve(' && line1 =~ '^[0-9:.]* *execve(')
+     || line1 =~ '^__libc_start_main'
     set ft=strace
 
     # VSE JCL
--- a/runtime/autoload/dist/script.vim
+++ b/runtime/autoload/dist/script.vim
@@ -338,6 +338,7 @@ def DetectFromText(line1: string)
     set ft=virata
 
     # Strace
+    # fast match first and complete regex if necessary
   elseif (line1 =~ 'execve(' && line1 =~ '^[0-9:.]* *execve(')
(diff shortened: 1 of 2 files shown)


fedemengo:
@brammool do you see any value in this?

brammool:
Federico Mengozzi wrote:

> I think this regex can be simplified with no side effects apart from
> an improvement in performance. 
> 
> When opening text file with long lines of numbers (bitstrings in my
> case) the additional and optional `[0-9:.]* *` put unnecessary work on
> the regex engine.
> 
> This is even more true in neovim, whose lua regex engine is not as
> good as vim. Honestly I don&#39;t see any reason why the regex should
> have the additional pattern. The two tests works fine with the
> simplified version

Hmm, I'm not 100% sure this won't cause problems.  execve() is also used
in C programs at least.  To avoid changing the functionality and still
have nearly all of the performance improvement, we can first check for
a match with "execve(" and only then check for the current pattern:

   elseif (line1 =~ 'execve(' && line1 =~ '[0-9:.]* *execve(') || line1 =~ '^__libc_start_main'

How about that?

-- 
The Law of VIM:
For each member b of the possible behaviour space B of program P, there exists
a finite time t before which at least one user u in the total user space U of
program P will request b becomes a member of the allowed behaviour space B'
(B' <= B).
In other words: Sooner or later everyone wants everything as an option.
                                        -- Vince Negri

fedemengo:
@brammool I think that would definitely be better, however I'm not sure it's really required since the additional `[0-9:.]* *` is optional. It's not really changing the result of the simpler regex.

For example, with this config `.vimrc`

```
filetype on
```

a file with this line

```
execve()
``` 

it's already being picked with as `strace` type. 

Screenshot 2023-04-13 at 18 40 05


That's why I was thinking we could either drop the `[0-9:.]* *` or we need to fix it up a bit.

brammool:
> @brammool I think that would definitely be better, however I'm not
> sure it's really required since the additional `[0-9:.]* *` is
> optional. It's not really changing the result of the simpler regex.
> 
> For example, with this config `.vimrc`
> 
> ```
> filetype on
> ```
> 
> a file with this line
> 
> ```
> execve()
> ``` 
> 
> it's already being picked with as `strace` type. 

Right, it appears this pattern was already faulty.  I guess a "^" should
have been at the start.  That way, if "execve" would be preceded by
something else than a timestamp it would not match.  So it would need to
be:
```
elseif (line1 =~ 'execve(' && line1 =~ '^[0-9:.]* *execve(') || line1 =~ '^__libc_start_main'
```

To avoid false matches.  This would still be fast.

-- 
VOICE OVER: As the horrendous Black Beast lunged forward, escape for Arthur
            and his knights seemed hopeless,  when, suddenly ... the animator
            suffered a fatal heart attack.
ANIMATOR:   Aaaaagh!
VOICE OVER: The cartoon peril was no more ... The Quest for Holy Grail could
            continue.

PR #12224, opened 2023-04-04
Fix typo

zeertzjq:


runtime/doc/pi_netrw.txt
DominiquePelle-TomTom, line 1: Can we use the opportunity to fix other typos in `pi_netrw.txt`?

* (line 525) these sort of problems -> these sorts of problems
* (line 1219 and 1444) its stored -> it's stored
* (line 1896) prpvided -> provided
* (line 2364) return the to -> return to the
* (line 3392) if its a local-directory -> if it's a local-directory
* (line 3717) its unable -> it's unable
* (line 3771) Its likely -> It's likely
* (line 4159) in a open subdirectory -> in an open subdirectory
* (line 4226) if its executable -> if it's executable
* (line 4292) "Tony Mechylynck" -> "Tony Mechelynck"
* (line 4389) so its in the process  -> so it's in the process

PR #12239, opened 2023-04-07
termdebug: fix winbar config

N-R-K:
Problem:  Despite having `winbar` disabled in `g:termdebug_config`, the
          termdebug plugin will still end up creating the winbar in certain
          situations. Such as when called from GotoSourcewinOrCreateIt().
Solution: Check for winbar inside InstallWinbar(), additionally use a
          `force` parameter to ensure explicit :Winbar commands are respected.


brammool:
I'll include it, thanks.

PR #12249, opened 2023-04-11
Add msghistadd()

Shougo:
It adds a message to the history without print.
It is useful for debug or to provide additional information.

### DIFF
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -3444,4 +3444,18 @@ func Test_setcmdline()
   cunmap a
 endfunc
 
+func Test_cmdhistadd()
+  messages clear
+
+  call assert_equal(execute('messages'), '')
+
+  call cmdhistadd('foo')
+
+  call assert_equal(execute('messages'), "\nfoo")
+
+  call cmdhistadd('bar', 'WarningMsg')
+
+  call assert_equal(execute('messages'), "\nfoo\nbar")
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -3451,11 +3451,11 @@ func Test_cmdhistadd()
 
   call cmdhistadd('foo')
 
-  call assert_equal(execute('messages'), "\nfoo")
+  call assert_equal(['foo'], GetMessages())
 
   call cmdhistadd('bar', 'WarningMsg')
 
-  call assert_equal(execute('messages'), "\nfoo\nbar")
+  call assert_equal(['foo', 'bar'], GetMessages())
 endfunc
 
 " vim: shiftwidth=2 sts=2 expandtab
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -3447,8 +3447,6 @@ endfunc
 func Test_cmdhistadd()
   messages clear
 
-  call assert_equal(execute('messages'), '')
-
   call cmdhistadd('foo')
 
   call assert_equal(['foo'], GetMessages())
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -122,6 +122,8 @@ charidx({string}, {idx} [, {countcc}])
 chdir({dir})			String	change current working directory
 cindent({lnum})			Number	C indent for line {lnum}
 clearmatches([{win}])		none	clear all matches
+cmdhistadd({msg}[, {highlight}])
+				Number	add an {msg} to command line history
 col({expr} [, {winid}])		Number	column byte index of cursor or mark
 complete({startcol}, {matches}) none	set Insert mode completion
 complete_add({expr})		Number	add completion match
@@ -1562,6 +1564,15 @@ clearmatches([{win}])					*clearmatches()*
 
 		Can also be used as a |method|: >
 			GetWin()->clearmatches()
+<
+cmdhistadd({msg}[, {highlight}])			*cmdhistadd()*
+		Add the String {msg} to the message history.
+		If {highlight} is set, the highlight group is used.
+		The result is a Number: TRUE if the operation was successful,
+		otherwise FALSE is returned.
+
+		Can also be used as a |method|: >
+			GetMsg()->cmdhistadd()
 <
 col({expr} [, {winid})					*col()*
 		The result is a Number, which is the byte index of the column
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -1049,6 +1049,7 @@ Command line:					*command-line-functions*
 	getcmdwintype()		return the current command-line window type
 	getcompletion()		list of command-line completion matches
 	fullcommand()		get full command name
+	cmdhistadd()		add an msg to command line history
 
 Quickfix and location lists:			*quickfix-functions*
 	getqflist()		list of quickfix errors
--- a/src/cmdhist.c
+++ b/src/cmdhist.c
@@ -538,6 +538,40 @@ del_history_idx(int histype, int idx)
 }
 
 /*
+ * "cmdhistadd()" function
+ */
+    void
+f_cmdhistadd(typval_T *argvars UNUSED, typval_T *rettv)
+{
+    char_u	*msg;
+    int		attr = 0;
+
+    rettv->vval.v_number = FALSE;
+    if (check_secure())
+	return;
+
+    if (in_vim9script()
+	    && (check_for_string_arg(argvars, 0) == FAIL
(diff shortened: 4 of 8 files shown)


src/testdir/test_cmdline.vim
```diff
@@ -3444,4 +3444,18 @@ func Test_setcmdline()
   cunmap a
 endfunc
 
+func Test_cmdhistadd()
+  messages clear
+
+  call assert_equal(execute('messages'), '')
```
ichizok: The order of arguments are `assert_equal({expected}, {actual})`.

ichizok: and I think should use [`GetMessages()`](https://github.com/vim/vim/blob/master/src/testdir/shared.vim#L385-L394) than `execute('message')` directly since the former will drop the maintainer entry.

brammool:
> It adds a message to the history without print.
> It is useful for debug or to provide additional information.

Is this really a good idea to add this?  It changes the command history,
thus one can no longer be sure if the history is really the list of
recent messages or some plugin added an entry somewhere.

I don't think this is a good way for debugging.  The ch_log() function
exists for that.

What additional information would need to be hidden in the message
history?  Either it's worth showing, or it should not appear at all.

If we can really be convinced that messages can be added to the history,
they should at least be recognizable, e.g. with a prefix to the text.

-- 
Michael: There is no such thing as a dump question.
Bernard: Sure there is.  For example "what is a core dump?"

src/testdir/test_cmdline.vim
Shougo: Fixed.

Shougo:
> What additional information would need to be hidden in the message
history?  Either it's worth showing, or it should not appear at all.

For example:

print(`echomsg`) the error message and print stacktrace in the message history.

If I use `echomsg` both long information printed and the error messages are hidden.
I think it is not intended.

brammool:
> > What additional information would need to be hidden in the message
> > history?  Either it's worth showing, or it should not appear at all.
> 
> For example:
> 
> print(`echomsg`) the error message and print stacktrace in the message
> history.
> 
> If I use `echomsg` both long information printed and the error
> messages are hidden.  I think it is not intended.

I don't understand this remark.  Vim does not have a print() function.

-- 
The war between Emacs and Vi is over.  Vi has won with 3 to 1.
http://m.linuxjournal.com/files/linuxjournal.com/linuxjournal/articles/030/3044/3044s1.html

ddickstein:
@brammool The idea here is something like using `echomsg` to display a brief message explaining the error so you know that an error happened, but not to emit long error output with a stack trace, etc. The error message could say "(see :messages for details)" to clarify that there's more to see there. Then when the user goes to check `:messages`, they will see full context of the error - the message along with the stack trace or other additional context. So if the user cares about the error information, all of it is available, they don't get hit with a large wall of text from the outset.

@Bakudankun you're right that this function could be abused, but I'm not sure that worry should prevent its availability, because there are ways to use it responsibly. Plugins that might log a lot of information to `:messages` should support some sort of `verbose` flag to make sure that's the user's intent. Another way to address that concern would be to namespace the message history so plugins could have their own spaces for their messages. Also I think your example doesn't quite work for the hit-enter prompt.

Bakudankun:
Okay, I just thought there was something left to discuss. It's true that my worry is not what prevents a feature.

By the way, the subject says `msghistadd()`, but actually the function is named `cmdhistadd()`. I feel `msghistadd()` is more accurate.
Also, I might prefer that `:silent echomsg` works in that way (if that does not have bad effects, of course).

Shougo:
```c
    void
add_msg_hist(
    char_u	*s,
    int		len,		// -1 for undetermined length
    int		attr)
{
    struct msg_hist *p;

    if (msg_hist_off || msg_silent != 0)
	return;

```

> Also, I might prefer that :silent echomsg works in that way (if that does not have bad effects, of course).

In current implementation, `add_msg_hist()` is skipped if silent.
I can remove the check instead of  to add new function.

But It changes current behavior.  It is OK?

zeertzjq:
> The idea here is something like using `echomsg` to display a brief message explaining the error so you know that an error happened, but not to emit long error output with a stack trace, etc. The error message could say "(see :messages for details)" to clarify that there's more to see there. Then when the user goes to check `:messages`, they will see full context of the error - the message along with the stack trace or other additional context. So if the user cares about the error information, all of it is available, they don't get hit with a large wall of text from the outset.

I don't think adding to message history is the best way to solve the original problem mentioned. There is a hard coded limit to the number of messages in message history. Why can't the extra information be stored into a variable and shown by a custom command?

Shougo:
@ddickstein What do you think?

brammool:
> > The idea here is something like using `echomsg` to display a brief
> > message explaining the error so you know that an error happened, but
> > not to emit long error output with a stack trace, etc. The error
> > message could say "(see :messages for details)" to clarify that
> > there's more to see there. Then when the user goes to check
> > `:messages`, they will see full context of the error - the message
> > along with the stack trace or other additional context. So if the
> > user cares about the error information, all of it is available, they
> > don't get hit with a large wall of text from the outset.
> 
> I don't think adding to message history is the best way to solve the
> original problem mentioned. There is a hard coded limit to the number
> of messages in message history. Why can't the extra information be
> stored into a variable and shown by a custom command?

It would be possible, but we'll end up with each plugin doing its own
thing, requiring the user to find and read the help to discover how to
show the extra information.  Having one way to do this makes it a lot
simpler.


-- 
An indication you must be a manager:
You give constructive feedback to your dog.

brammool:
> I expect `:mes` to show what I've seen just now. I may be confused if
> vast amounts of ghost messages appear when I call it.
> If this patch is adopted, the message history would turn into the
> common logstack of every plugins. Is that what you want?

Right, this change may cause users to get lost in the extra information.
Perhaps one should need to use ":verbose messages" to see it.
Thus the extra messages can be added, but will not be visible unless the
user explicitly wants to see them.  The messages would then also not
count to the maximum number of messages being kept.

-- 
An indication you must be a manager:
You feel sorry for Dilbert's boss.

brammool:
> ```c
>     void
> add_msg_hist(
>     char_u	*s,
>     int		len,		// -1 for undetermined length
>     int		attr)
> {
>     struct msg_hist *p;
> 
>     if (msg_hist_off || msg_silent != 0)
> 	return;
> 
> ```
> 
> > Also, I might prefer that :silent echomsg works in that way (if that does not have bad effects, of course).
> 
> In current implementation, `add_msg_hist()` is skipped if silent.
> I can remove the check instead of  to add new function.
> 
> But It changes current behavior.  It is OK?

Changing behavior is not OK.  For ":silent echo" itself the intention
may be clear, but for ":silent call SomeFunc()" it is not.

-- 
ROBIN:  The what?
ARTHUR: The Holy Hand Grenade of Antioch.  'Tis one of the sacred relics
        Brother Maynard always carries with him.
ALL:    Yes. Of course.
ARTHUR: (shouting) Bring up the Holy Hand Grenade!

brammool:
> @brammool The idea here is something like using `echomsg` to display a
> brief message explaining the error so you know that an error happened,
> but not to emit long error output with a stack trace, etc. The error
> message could say "(see :messages for details)" to clarify that there's
> more to see there. Then when the user goes to check `:messages`, they
> will see full context of the error - the message along with the stack
> trace or other additional context. So if the user cares about the error
> information, all of it is available, they don't get hit with a large
> wall of text from the outset.

In principle this idea sounds fine.  If the user sees a warning/error
message and knows why it appears then the extra info isn't needed.
However,  I would only show the extra info when asked for with ":verbose
messages".  Otherwise the displayed messages an the output of
":messages" doesn't match, which is confusing.


> @Bakudankun you're right that this function could be abused, but I'm not
> sure that worry should prevent its availability, because there are ways
> to use it responsibly. Plugins that might log a lot of information to
> `:messages` should support some sort of `verbose` flag to make sure
> that's the user's intent. Another way to address that concern would be
> to namespace the message history so plugins could have their own spaces
> for their messages. Also I think your example doesn't quite work for the
> hit-enter prompt.

Keep in mind: if something can be abosed, it will be abused.

-- 
An indication you must be a manager:
You can explain to somebody the difference between "re-engineering",
"down-sizing", "right-sizing", and "firing people's asses".

mg979:
Even limiting the displaying of extra context to `:verbose messages` wouldn't give a clear output, because the error context would be mixed with all other messages, and possibly with previous error contexts. Quite confusing.

If the goal is to have some built-in logging facility, why mixing logs with regular messages? I think a `:logs <context>` command would be better. For example:
```vim
call logadd(context, msg) " `context` may be a plugin name, it will be created if not existing
" this command shows all messages registered for `context`
:logs <context>
```

Shougo:
Hm...  It should be addressed by another feature.

PR #12251, opened 2023-04-12
Distinguish between Forth and Fortran filetypes for *.f files

dkearns:
Also adds *.4th as a Forth filetype

@jkotlinski, could you please run your eyes over this?

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -287,6 +287,37 @@ export def FTe()
   endif
 enddef
 
+def IsForth(): bool
+  var first_line = nextnonblank(1)
+
+  # SwiftForth block comment (line is usually filled with '-' or '=') or
+  # OPTIONAL (sometimes precedes the header comment)
+  if getline(first_line) =~? '^\%({\%(\s\|$\)\|OPTIONAL\s\)'
+    return true
+  endif
+
+  var n = first_line
+  while n < 100 && n <= line("$")
+    # Forth comments and colon definitions
+    if getline(n) =~ '^[:(\\] '
+      return true
+    endif
+    n += 1
+  endwhile
+  return false
+enddef
+
+# Distinguish between Forth and Fortran
+export def FTf()
+  if exists("g:filetype_f")
+    exe "setf " .. g:filetype_f
+  elseif IsForth()
+    setf forth
+  else
+    setf fortran
+  endif
+enddef
+
 export def FTfrm()
   if exists("g:filetype_frm")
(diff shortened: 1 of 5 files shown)


jkotlinski:
Hey, sorry, I missed the notification. I will have a look.

runtime/autoload/dist/ft.vim
```diff
@@ -286,6 +286,39 @@ export def FTe()
...
+    exe "setf " .. g:filetype_f
+    return
+  endif
+
+  var first_line = getline(nextnonblank(1))
+
+  # SwiftForth block comment, line is usually filled with '-' or '='
+  var header_comment   = first_line =~ '^{\%(\s\|$\)'
+  # SwiftForth OPTIONAL, sometimes appears before the header comment
+  var optional_include = first_line =~ '^OPTIONAL\s'
```
jkotlinski: Could the SwiftForth checks give false positives for F#? If not, I think they would be better moved to the IsForth() function.

dkearns:
I don't believe so.  I'll push the extra tests into IsForth().

Thanks.

runtime/filetype.vim
```diff
@@ -710,16 +710,19 @@ au BufNewFile,BufRead auto.master		setf conf
...
-au BufNewFile,BufRead *.ft,*.fth		setf forth
+au BufNewFile,BufRead *.ft,*.fth,*.4th		setf forth
 
 " Reva Forth
 au BufNewFile,BufRead *.frt			setf reva
 
 " Fortran
 if has("fname_case")
-  au BufNewFile,BufRead *.F,*.FOR,*.FPP,*.FTN,*.F77,*.F90,*.F95,*.F03,*.F08	 setf fortran
+  au BufNewFile,BufRead *.F,*.FOR,*.FPP,*.FTN,*.F77,*.F90,*.F95,*.F03,*.F08	setf fortran
```
jkotlinski, line 734: Is it intentional that *.F is still mapped to Fortran only?

dkearns: It is on the grounds that:

- Fortran has some conventions for uppercase extensions that makes them meaningful.
- We don't support uppercase for any other Forth extensions or other filetypes in general.
- SwiftForth (a notable user of `*.f`) doesn't implicitly recognise `*.F` as it does `*f`.

Is there some history of intentionally capitalised extensions?

jkotlinski: Not that I am aware of. I was simply unaware of Fortran conventions.

yegappan:

PR #12258, opened 2023-04-13
CI: Workaround for test_diff_screen() on Cirrus CI (macOS 13)

ichizok:
Since macOS 13 changed diff command from GNU diff to Apple diff, but this seems to have a bug so causes `test_diff_screen()` fails on Cirrus CI (which uses macOS 13).
As a workaround, install GNU diffutils before testing.

```
$ seq 10 > file0
$ seq 0 9 > file1
$ diff -U0 file0 file1
```

GNU diff:
```
--- file0
+++ file1
@@ -0,0 +1 @@
+0
@@ -10 +10,0 @@
-10
```

Apple diff (macOS 13.2):
```
--- file0
+++ file1
@@ -1,0 +1 @@
+0
@@ -10 +10,0 @@
-10
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -36,7 +36,7 @@ macos_task:
   timeout_in: 20m
   install_script:
     - brew update
-    - brew install gettext libtool
+    - brew install gettext libtool diffutils
   build_script:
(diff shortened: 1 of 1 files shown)


ichizok:
This and #12247 should make macOS CI fix.

PR #12260, opened 2023-04-13
Redraw statusline and tabline later for empty 'showcmd'

luukvbaal:
Problem:    Statusline and tabline may be redrawn unnecessarily for
	    'showcmdloc' if already marked for redraw.
Solution:   Mark for redraw later when showcmd buffer is empty.

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -1803,9 +1803,19 @@ display_showcmd(void)
     cursor_off();
 
     if (*p_sloc == 's')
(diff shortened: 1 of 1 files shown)


luukvbaal:
The same pattern could be applied with `clear_cmdline` for 'showcmdloc' == "last" but that will be less impactful as opposed to statusline/tabline which may run user scripts. But it could avoid redundantly clearing the showcmd area if `clear_cmdline` was already set.

PR #12261, opened 2023-04-14
runtime: Recognise ".cppm", etc. files as C++ 20 Modules

puremourning:
clang's c++ 20 modules uses the "cppm" and similiar extensions: https://clang.llvm.org/docs/StandardCPlusPlusModules.html#file-name-requirement

Recognise .cppm, .ccm, .cxxm and .c++m as a C++ file (C++ 20 Module)

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -338,6 +338,10 @@ if has("fname_case")
   au BufNewFile,BufRead *.C,*.H setf cpp
 endif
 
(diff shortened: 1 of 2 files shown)


brammool:
Please add entries in src/testdir/test_filetype.vim

puremourning:
Yes sure thing will do!

puremourning:
> Please add entries in src/testdir/test_filetype.vim

That's done.

PR #12269, opened 2023-04-17
Fix: `execute({list})` is not equivalent from to `redir` contrary to help description

ichizok:
`redir` can handle line-continuation but `execute({list})` cannot.

`redir`: no error
```vim
redir => var
echo 1 +
    \ 2
redir END
```

`execute({list})`: `E15: Invalid expression: "1 +"`
```vim
echo execute(['echo 1 +',  '\ 2'])
```

This PR modifies `get_list_line()` to handle line-continuation as `getsourceline()` and in addtion define `get_list_nextline()` to handle vim9script's line-continuation.

### DIFF
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -4028,18 +4028,95 @@ get_list_line(
     int	    c UNUSED,
     void    *cookie,
     int	    indent UNUSED,
-    getline_opt_T options UNUSED)
+    getline_opt_T options)
 {
     listitem_T **p = (listitem_T **)cookie;
     listitem_T *item = *p;
     char_u	buf[NUMBUFLEN];
-    char_u	*s;
+    char_u	*line;
+    int		do_vim9_all = in_vim9script() && options == GETLINE_CONCAT_ALL;
+    int		do_bar_cont = do_vim9_all || options == GETLINE_CONCAT_CONTBAR;
 
     if (item == NULL)
 	return NULL;
-    s = tv_get_string_buf_chk(&item->li_tv, buf);
-    *p = item->li_next;
-    return s == NULL ? NULL : vim_strsave(s);
+    line = tv_get_string_buf_chk(&item->li_tv, buf);
+    line = line == NULL ? NULL : vim_strsave(line);
+    item = item->li_next;
+
+    // Only concatenate lines starting with a \ when 'cpoptions' doesn't
+    // contain the 'C' flag.
+    if (line != NULL && item != NULL && options != GETLINE_NONE
+				      && vim_strchr(p_cpo, CPO_CONCAT) == NULL)
+    {
+	int comment_char = in_vim9script() ? '#' : '"';
+	char_u *nextline;
+	char_u *s;
+
+	// Get the next line and concatenate it when it starts with a
+	// backslash. We always need to read the next line, keep it in
+	// sp->nextline.
+	/* Also check for a comment in between continuation lines: "\ */
+	// Also check for a Vim9 comment, empty line, line starting with '|',
+	// but not "||".
+	nextline = tv_get_string_buf_chk(&item->li_tv, buf);
+	if (nextline != NULL
+		&& (*(s = skipwhite(nextline)) == '\\'
+			      || (s[0] == comment_char
+						&& s[1] == '\\' && s[2] == ' ')
+			      || (do_vim9_all && (*s == NUL
+						     || vim9_comment_start(s)))
+			      || (do_bar_cont && s[0] == '|' && s[1] != '|')))
+	{
+	    garray_T    ga;
+
+	    ga_init2(&ga, sizeof(char_u), 400);
+	    ga_concat(&ga, line);
+	    vim_free(line);
+	    if (*s == '\\')
+		ga_concat(&ga, s + 1);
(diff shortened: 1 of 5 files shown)


ichizok:
For a plugin or a plugin manager, can use `execute({list})` to load config from a part of a script file without `:source`, or use `execute()` without consideration of line-cotinuation.

In addition, for the plugin which set its config (written in vim script) embedded into other format e.g. TOML, when loading the config by `execute({list})` should be useful to able to handle line-continuation.

e.g. plugin manager

user's config; describe each plugin's config by heredoc

```vim
let g:plmgr.config = {}
let g:plmgr.config['plugin-A'] =<< trim EOS
  let g:plugin_a_config = {
    \   'abc': '123',
    \ }
]
EOS
let g:plmgr.config['plugin-B'] =<< trim EOS
  let g:plugin_b_config = {
    \   'def': '456',
    \ }
]
EOS
```

plugin manager's code; load plugin's config which is enabled

```vim
" user want to enable plugin depending on an environment
if s:is_enabled(plugin_name)
  call execute(g:plmgr.config[plugin_name])
endif
```

e.g. config in TOML

```toml
[feature]
# load the following by using execute({list})
config = '''
let g:plugin_a_feature_a = {
  \   'foo': '123',
  \ }
'''
```

Shougo:
It also works for Vim9 script?

Shougo:
Current `execute()` behavior:

```vim
let g:cmds = [
      \ "var foo = {",
      \ "  'runner': 'job',",
      \ "  'outputter': 'buffer',",
      \ "  'outputter/buffer/close_on_empty': 1,",
      \ "}",
      \ ]

vim9 call execute(g:cmds, "")
```

```
Error detected while processing /home/shougo/work/test_vim9.vim:
line    9:
E723: Missing end of Dictionary '}':
```

Bakudankun:
I think an extra flag should be added to `execute()` if this PR changes existing valid behavior (especially in vim9).

Shougo:
I have tested your branch and it works on Vim9.  Yeah.

hokorobi:
Oh, I would appreciate it if this is available.

ichizok:
As @brammool and @Bakudankun are concerned, at not vim9 this PR makes an error statement be no error at most, but at vim9 could change the interpretation of valid statements.
If add a option which switches to enable to handle line-continuation as the 3rd argument of `execute()`, since currently the default value of the 2nd argument of `execute()` is `'silent'` it makes some annoyance that when use the 3rd argument need specify the 2nd argument explicitly as "silent".
e.g. `execute({command}, 'silent', 1)` (`1` means handling line-continuation)

I think, if need a new option, should make `execute()` handle also dictionary as the 2nd argument:
```vim
execute({command} [, {silent}])  " {silent} is '' or 'silent' or 'silent!'
" or
execute({command} [, {option}])  " {option} is as #{silent: 'silent', multiline: v:true}
```

brammool:
> As @brammool and @Bakudankun are concerned, at not vim9 this PR makes an
> error statement be no error at most, but at vim9 could change the
> interpretation of valid statements.
> If add a option which switches to enable to handle line-continuation as
> the 3rd argument of `execute()`, since currently the default value of
> the 2nd argument of `execute()` is `'silent'` it makes some annoyance
> that when use the 3rd argument need specify the 2nd argument explicitly
> as "silent".
> e.g. `execute({command}, 'silent', 1)` (`1` means handling line-continuation)
> 
> I think, if need a new option, should make `execute()` handle also dictionary as the 2nd argument:
> ```vim
> execute({command} [, {silent}])  " {silent} is '' or 'silent' or 'silent!'
> " or
> execute({command} [, {option}])  " {option} is as #{silent: 'silent', multiline: v:true}
> ```

Changing the arguments of execute() is not going to result in a nice
solution.  Adding a variant of execute(), such as execute_concat(),
doesn't seem a good choice either.

I wonder if it would be better to have a separate function that does the
concatenation.  Thus you would use something like:

	execute(script_concatenate(list))

Not sure if this is a good solution.

For the implementation, a lot of code lines were copied from
getsourceline(), leading to duplication.  That should be avoided.  It
shouldn't be too difficult to make a common function for this.

-- 
I learned the customs and mannerisms of engineers by observing them, much the
way Jane Goodall learned about the great apes, but without the hassle of
grooming.

chrisbra:
I don't think this is too useful, because:

1) the help expclitly states

> {command} can be a string or a List.  In case of a List the lines are executed one by one.

That wouldn't be true any more (*the lines are executed one by one*) and joining all the list items together may have some performance/memory implications and I am worried this would cause slight incompatibilities in that the result is then different.

And the help also states

> 		This is more or less equivalent to: >
>			redir => var
>			{command}
>			redir END

So it doesn't really say that `execute()` and `:redir` is always exactly the same. 

2) You can already use:
```
let a= execute('1 + '.
 \ ' + 2') 
```
So string concatenation works.

3) and I am missing an convincing argument how come that you have in your list string concatenation. So why do you have a list at all and not a string? It seems `execute(list + list2)` should already work. And you are using a list, when you actually would want to use a string instead?


So I am rather negative about this

chrisbra:
I am closing this, as I am not convinced ot its usefulnes

PR #12277, opened 2023-04-18
Fix warnings on macOS

mojotx:
Clang 14.0.3 is complaining that a function declaration without a prototype is deprecated in all versions of C. 

These just clean up the warnings. 

Example:

```
crypt.c:43:25: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
    int (* self_test_fn)();
                        ^
                         void
```

Also, in one place (`src/os_unix.c:673`) I fixed where a comment didn't match the opening comment.

### DIFF
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -198,7 +198,7 @@ static void deathtrap SIGPROTOARG;
 
 static void catch_int_signal(void);
 static void set_signals(void);
-static void catch_signals(void (*func_deadly)(), void (*func_other)());
+static void catch_signals(void (*func_deadly)(int), void (*func_other)(int));
 #ifdef HAVE_SIGPROCMASK
 # define SIGSET_DECL(set)	sigset_t set;
 # define BLOCK_SIGNALS(set)	block_signals(set)
@@ -670,7 +670,7 @@ mch_delay(long msec, int flags)
 	}
 #  endif // HAVE_SELECT
 # endif // HAVE_NANOSLEEP
-#endif // HAVE_USLEEP
+#endif // HAVE_NANOSLEEP
 #ifdef FEAT_MZSCHEME
 	}
 	while (total > 0);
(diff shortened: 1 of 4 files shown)


src/os_unix.c
```diff
@@ -670,7 +670,7 @@ mch_delay(long msec, int flags)
 	}
 #  endif // HAVE_SELECT
 # endif // HAVE_NANOSLEEP
-#endif // HAVE_USLEEP
+#endif // HAVE_NANOSLEEP
```
dpelle: It should be:
```
   672 # endif // HAVE_USLEEP
   673 #endif // HAVE_NANOSLEEP
```

brammool:
> @dpelle commented on this pull request.
> 
> > @@ -670,7 +670,7 @@ mch_delay(long msec, int flags)
>  	}
>  #  endif // HAVE_SELECT
>  # endif // HAVE_NANOSLEEP
> -#endif // HAVE_USLEEP
> +#endif // HAVE_NANOSLEEP
> 
> It should be:
> ```
>    672 # endif // HAVE_USLEEP
>    673 #endif // HAVE_NANOSLEEP
> ```

I'll correct it (not making a separate patch for it though).

I wonder how useful this really is.  The matching #ifdef is only 30
lines away.  These extra comments can be helpful if the start is far
away, but even then a "%" command should find it quickly.

In this case I think it's better to just leave it out, also because
there is an #else in between.  How about that?

-- 
A parent can be arrested if his child cannot hold back a burp during a church
service.
		[real standing law in Nebraska, United States of America]

DominiquePelle-TomTom:
@brammool asked:
> In this case I think it's better to just leave it out, also because
> there is an #else in between.  How about that?

I don't mind either way.
No comment or correct comment is better than incorrect comment.

mojotx:
I just pushed a change backing out the comment change. Sorry about that!

src/os_unix.c
mojotx: Whoops. *blush* Just pushed 76eca116a to fix this.

brammool:
> I just pushed a change backing out the comment change. Sorry about that! 

Did the patch get the right version of this PR?
I decided to drop the #endif comments, they are not very useful here.


-- 
A special cleaning ordinance bans housewives from hiding dirt and dust under a
rug in a dwelling.
		[real standing law in Pennsylvania, United States of America]

Issue #12204, opened 2023-03-28
sodium encryption misuses crypto_pwhash API


dgl:
### Steps to reproduce

1. Review: https://github.com/vim/vim/blob/master/src/crypt.c#L883-L884
2. See https://github.com/jedisct1/libsodium/blob/master/src/libsodium/include/sodium/crypto_pwhash.h#L97-L102

Nothing stores the parameters used for the password hash function. Libsodium is [currently using argon2id](https://github.com/jedisct1/libsodium/blob/089f850608737f9d969157092988cb274fe7f8d4/src/libsodium/include/sodium/crypto_pwhash.h#L25), but those constants could change per the libsodium API docs.

### Expected behaviour

`cryptmethod=xchacha20` should store the parameters to the pwhash function.

Given the docs say this format is currently experimental it should be possible to add the parameters somewhere to provide algorithm agility.

[cc: @chrisbra]

### Version of Vim

9.0.1429

### Environment

Anything with libsodium enabled.

dgl:
BTW, I've not fully reviewed the xchacha20 method, it looks sane, but given an omission like this I'd far prefer to see something like [age's scrypt](https://github.com/C2SP/C2SP/blob/main/age.md#the-scrypt-recipient-type) format added to Vim, so it is interoperable with other tools, but that's getting more into a feature request than this issue.

chrisbra:
Thanks for the review. Do you mean more like using `crypto_pwhash_ALG_ARGON2ID13`, `crypto_pwhash_argon2id_OPSLIMIT_INTERACTIVE` and `crypto_pwhash_argon2id_MEMLIMIT_INTERACTIVE`  explicitly or are you more thinking of making this configurable for the user? (e.g. like a hyptothetical  `:set cryptoptions=pwhash:alg_argon2id3` etc)?

Using those defines explicitly would mean, we stay compatible and be sure to decrypt old encrypted files (at least if the implementation does not change) at the same time, we possibly wouldn't notice when a more secure default is selected in libsodium.

chrisbra:
Ah, I might have misunderstood. I think you want us to store those constants together with the hash?

brammool:
I suppose the values of crypto_pwhash_OPSLIMIT_INTERACTIVE, crypto_pwhash_MEMLIMIT_INTERACTIVE and crypto_pwhash_ALG_DEFAULT need to be stored in the header, like salt and seed.  Since the header size changes, this requires a new ctrypt method "magic name".
[...]

PR #12279, opened 2023-04-19
Improve Libsodium API usage and use better xchacha20 version

chrisbra:
This is the PR for fixing #12204 

I have tried to split up the changes to make reviewing easier. It starts by refactoring some of the crypt code to make integrating the new libsodium method easier. 

There are 2 gotchas:
- I am not sure if the way the hashing parameters are stored is portable. It works on my system, but I suppose on different architectures this could be problematic. 
- Reading the file was a bit tricky. Since the XChaCha20 encryption uses a fixed block size, but the file header changes, it's tricky to make decryption work between both versions. The way it works currently is, to use the slightly larger block size, but when reading the first block and we notice we are using the previous xchacha20 encryption, to first decrypting the internal block size (when decrypting the first block), but we also need to rewind the file descriptor by those bytes, else on the next block it would start reading those 20 bytes too later and we have to adjust for that. 

I have added some tests to verify that the current and old encryption works when more than 1 block is read. Also added one test to make sure decrypting works with custom hashing parameters used, to verify that reading those parameters from the file work.

### DIFF
--- a/runtime/doc/editing.txt
+++ b/runtime/doc/editing.txt
@@ -1533,9 +1533,10 @@ To disable the encryption, reset the 'key' option to an empty value: >
 
 You can use the 'cryptmethod' option to select the type of encryption, use one
 of these: >
-	:setlocal cm=zip        " weak method, backwards compatible
-	:setlocal cm=blowfish   " method with flaws
-	:setlocal cm=blowfish2  " medium strong method
+	:setlocal cm=zip          " weak method, backwards compatible
+	:setlocal cm=blowfish     " method with flaws
+	:setlocal cm=blowfish2    " medium strong method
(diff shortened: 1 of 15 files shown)


brammool:
I have the habit of using a prefix for struct members, so that it's easy to change them and avoid mistakes with using the wrong name.  The "cat_" prefix should work well for "crypt_arg_T".

The "#ifdef FEAT_EVAL" before using p_verbose doesn't make sense.  Since this appears to be for debugging, putting the #ifdef around the use of crypt_sodium_report_hash_params() should work better (and make the tiny build a bit smaller).

chrisbra:
I pushed 2 commits to address those two points.

PR #12281, opened 2023-04-20
Detect filetypes for *.v files

### LINKED ISSUE #12274: Add support for the V programming language. Detect filetypes for *.v files
Fieltype detection for the V programming language is currently not supported.
.v files are always set to verilog by default.

Since verilog syntax has semicolon line endings, I assume it could serve as a good indicator. E.g., the following would set the ft to verilog when a line is found that is is not a comment and it ends with a semicolon.

```vim
# Determine if a *.v file is V or Verilog
export def FTv()
  var numberOfLines = line('$')
  for i in range(1, numberOfLines)
    var currentLine = trim(getline(i))
    var firstCharacter = currentLine[0]
    let lastCharacter = strchar(current_line, -1)
    if firstCharacter !=? "/" && lastCharacter ==? ";"
      setf verilog
      return
    endif
  endfor
  setf v
enddef
```

Is adding this support feasible?

ttytm:
closes #12274

Edit: Uff, what did I do to get this cov report? 🙈
Also feel free to point out or change formatting or naming preferences. I haven't found a consistent style in the codebase that I could model myself after. Since ci takes a bit, I'll hold off on making further changes on my end.

Edit 2: Okay codecov was re-subbed I see.🍀

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1106,7 +1106,7 @@ export def FTlsl()
   endif
 enddef
 
-# Set the filetype of a *.v file to Verilog, V or Cog based on the first 100 non-comment lines.
+# Set the filetype of a *.v file to Verilog, V or Cog based on the first 200 lines.
 export def FTv()
   if did_filetype()
     return
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1106,5 +1106,26 @@ export def FTlsl()
   endif
 enddef
 
+# Set the filetype of a *.v file to V or Verilog based on the first 100 non-comment lines.
+# While V doesn't use semicolons for line endings, statements in Verilog end with a semicolon.
+# It's taken into account that a semicolon can be followed by a comment.
+export def FTv()
+  var numberOfLines = line('$')
+  var nonCommentLine = 0
+  for ln in getline(1, numberOfLines)
+    if nonCommentLine > 99
+      break
+    endif
+    if ln[0] ==? "/"
+      continue
+    elseif ln =~ ';\s\?\(/.*\)\?$'
+      setf verilog
+      return
+    endif
+    nonCommentLine += 1
+  endfor
+  setf v
+enddef
+
 # Uncomment this line to check for compilation errors early
 # defcompile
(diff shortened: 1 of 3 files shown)


chrisbra:
> for ln in getline(1, numberOfLines)

Could `.v` files be million lines long? Wouldn't that take too much memory? 
Perhaps that should be changed to ` for ln in getline(1, min(200, numberOfLines))` to be on the safe side?

chrisbra:
yes, I think that should also work. Thanks!

benknoble:
Please note that `.v` is also an extension used for Coq files. The most commonly-used Coq-related Vim plugin https://github.com/whonore/Coqtail unconditionally detects `*.v` as the filetype `coq`.

If need be, words like `Require`, `Import`, `Definition`, `Inductive`, and `(* *)` comments could be used for disambiguation.

tobealive:
Thanks for the hint @benknoble.
As far as I can see, the most distinct patter next to comment style are dot character code line endings `.`. Since it looked to me that there are a few more keywords in Coq that would match the dot ending is probably on the safer side. What do you think about using this for disambiguation?

benknoble:
> Thanks for the hint @benknoble. As far as I can see, the most distinct patter next to comment style are dot character code line endings `.`. Since it looked to me that there are a few more keywords in Coq that would match the dot ending is probably on the safer side. What do you think about using this for disambiguation?

Seems reasonable. Most Coq commands can be thought of as "sentences" ending with `.`, as you said.

tobealive:
This one might be ready. When a moment frees up you can take a look @brammool 🙂:+1:

brammool:
Thanks for making this PR.  A few remarks.

The use of "non_comment_line_num" doesn't look very useful.  Dropping it
makes it simpler, and I don't see much of a disadvantage.

I suppose this condition is to detect the start of a comment:

    if ln[0] ==? "/"

This doesn't allow for any white space before the slash.  To allow for
that this should work better:

    if ln[0] =~ '^\s*/'

If 'filetype' was already set then the ":setf" commands won't do
anything.  Still we go through the lines to do the detection, which is a
waste of time.  To avoid it put this at the top of FTv():

	if did_filetype()
	   return
	endif

The test is minimal, but it's a good start.

-- 
A law to reduce crime states: "It is mandatory for a motorist with criminal
intentions to stop at the city limits and telephone the chief of police as he
is entering the town.
		[real standing law in Washington, United States of America]

tobealive:
Thanks a lot @brammool. I applied your suggestions.

Issue #11899, opened 2023-01-27
Proposal with proof of concept patch - allow customization of tab background color in Windows GUI version


roel-v:
Microsoft in its infinite wisdom has decided that on modern Windows versions (11, at least), the background colors of selected and not selected tabs (when using the standard common controls w32 tab control, which vim does) should be only a tiny bit different. This makes it quite difficult to see which tab is selected and even more difficult what changes wrt tab selection when switching between tabs with the keyboard. There is a way to change background colors for console tabs (TabLine and TabLineSel in the color scheme), but the Windows GUI version doesn't use that. It would be nice to have a way to make the current tab stand out more.

I have attached a small proof of concept patch that will highlight the current tab in (a quite unusable for now) red. Maybe a real patch would use the colors specified in TabLine and TabLineSel? Or maybe there should be a separate setting somewhere? Or maybe it should be possible to enable/disable highlighting background colors in GUI tabs all together? There are design questions to be answered before making a real patch, hence me opening this issue.

As you can see, it's a rather minor change to draw tab backgrounds in a custom color. From what I can tell, there don't seem to be many risks wrt impact on other code or other side effects, but I haven't done much testing.

So I'm requesting feedback on if and how I should proceed with implementing this.

[w32_tab_background_color.patch](https://github.com/vim/vim/files/10519183/w32_tab_background_color.patch)

zewpo:
Hi Roel, 

I agree with your motivation there - it's not easy to see which tab is selected.     I guess the original default win32 styling theme was designed to work well on the old CRT type of displays.   But, like my laptop, with a recent type of LCD display, has a much higher dpi and much more detailed color pallette - than the old CRTs could ever hope to provide.    Also, I like to use a dark theme these days which makes the default win32 apps look a bit odd.   I think in this environment, the old default win32 themes don't really achieve what they were originally designed for.  

So, yeah, lots of potential for improving Vim here :)

I'd like to see how to change the color of the text as well as the background.

The idea of using the existing colors from TabLine and TabLineSel make sense to me.  If they have been configured - then use them - otherwise, we can use sensible defaults. 

Something else that I've been looking at, slightly related, is being able to darken the light background of the gui - for example, for the Windows 11 Dark application mode.   Like, I noticed how Notepad++ now has a dark mode - and I quite like it.

Even the good old Notepad looks a LOT better now in dark mode on the latest version of Windows 11.

[...]

PR #12282, opened 2023-04-20
win32: Enable dark mode when guioptions contains 'd'

### LINKED ISSUE #3922: windows 10 dark mode border style
I have a question / feature wish.

would it be possible for gvim on windows 10 to take dark mode into account?, the window borders stay lightgrey even in dark mode...

vim_windows10_windowborders

LemonBoy:
Enabling the dark mode makes the windows title bar dark, the option is already being honored by the GTK UI so let's follow the lead and enable it on Win32 too.

Closes #3922

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -3912,6 +3912,18 @@ A jump table for the options with a short description can be found at |Q_op|.
 	In its simplest form the value is just one font name.
 	See |gui-font| for the details.
 
+					*'guidarkmode'*
+'guidarkmode'		string	(default "automatic")
+			global
+			{only available when compiled with GUI enabled}
+	This option specifies whether Vim GUI elements should be drawn using a
+	dark color scheme if available.
+	The allowed values are:
+	  automatic	Follow the current system's preference.
+	  prefer_dark	Apply the dark mode to the GUI elements.
+	  prefer_light	Apply the light mode to the GUI elements.
+	  use_bg	Follow the value of Vim's |background| option.
+
 					*'guifontset'* *'gfs'*
 					*E250* *E252* *E234* *E597* *E598*
 'guifontset' 'gfs'	string	(default "")
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -4011,7 +4011,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 		choices.
 								*'go-d'*
 	  'd'	Use dark theme variant if available. Currently only works for
-		GTK+ GUI.
+		Win32 and GTK+ GUI.
 								*'go-e'*
 	  'e'	Add tab pages when indicated with 'showtabline'.
 		'guitablabel' can be used to change the text in the labels.
--- a/src/feature.h
+++ b/src/feature.h
@@ -513,7 +513,7 @@
 /*
  * GUI dark theme variant
  */
-#if defined(FEAT_GUI_GTK) && defined(USE_GTK3)
+#if (defined(FEAT_GUI_GTK) && defined(USE_GTK3)) || defined(FEAT_GUI_MSWIN)
 # define FEAT_GUI_DARKTHEME
 #endif
 
--- a/src/gui.c
+++ b/src/gui.c
@@ -741,6 +741,7 @@ gui_init(void)
 	resettitle();
 
 	init_gui_options();
+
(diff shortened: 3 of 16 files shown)


brammool:
Did anybody try this out?  Does it work as advertised, no bad side effects?

Bakudankun:
I tried. It works well, but I feel it's not perfect.

* The title bar is darkened. On a Win10 machine I have to minimize and restore the window for the change to take effect, but it is small problem if you write the setting in vimrc
* Context menu is also darkened.
* The menubar, toolbar, tabbar, and scrollbars are not darkened. Can you make them dark?

image

Bakudankun:
Anyway only the dark title bar would help many people. Actually I've hidden all other GUI elements.

LemonBoy:
> The menubar, toolbar, tabbar, and scrollbars are not darkened. Can you make them dark?

I guess so, but the Win32 API offers no straightforward way to do so (using `SetWindowTheme` to apply an Explorer-specific theme) and requires a decent amount of refactoring of the win32 frontend code (as we'd have to expilcitly keep track of every element HWND to apply/unapply the theming).

> Anyway only the dark title bar would help many people. Actually I've hidden all other GUI elements.

Me too, further improvements are welcome but not a blocker for this PR IMO.

> The title bar is darkened. :set guioptions+=d takes effect immediately on Win11, but on a Win10 machine I have to minimize and restore the window for the change to take effect.

Uh, weird. What's the build number on that Windows 10 version?

Bakudankun:
> > The title bar is darkened. :set guioptions+=d takes effect immediately on Win11, but on a Win10 machine I have to minimize and restore the window for the change to take effect.
> 
> Uh, weird. What's the build number on that Windows 10 version?

Win10 22H2 (10.0.19045.2673).
Also, my description was not accurate. Actually I have to make the OS redraw the title bar for the change to take effect.
That means, unfocusing is enough because that redraws the title bar to change its color.

ychin:
I know this is just porting the GTK version over by allowing `set guioptions+=d`, but is this really the best API for doing so? I would imagine the default behavior should be to use the system light/dark mode (I *think* this is the case now?), and allow for the user to customize light or dark mode? It seems like with this option you are always forcing a light or dark mode without an ability to just say "use the system settings" which I imagine is what most people want.

MacVim implemented dark mode separately (it was before the GTK one was implemented) and the 4 settings we have are:

1. "Automatic" (default)
2. Light mode (always force light mode even if system settings is dark)
3. Dark mode (always force dark mode)
4. Use 'background' option (use dark mode or not depending on the `'background'` setting in Vmi)

brammool:
> I know this is just porting the GTK version over by allowing `set
> guioptions+=d`, but is this really the best API for doing so? I would
> imagine the default behavior should be to use the system light/dark
> mode (I *think* this is the case now?), and allow for the user to
> customize light or dark mode? It seems like with this option you are
> always forcing a light or dark mode without an ability to just say
> "use the system settings" which I imagine is what most people want.
> 
> MacVim implemented dark mode separately (it was before the GTK one was
> implemented) and the 4 settings we have are:
> 
> 1. "Automatic" (default)
> 2. Light mode (always force light mode even if system settings is dark)
> 3. Dark mode (always force dark mode)
> 4. Use 'background' option (use dark mode or not depending on the `'background'` setting in Vmi)

This sounds like a good way to do this, especially for new users.

Where are these settings made?  Adding more flags to 'guioptions' isn't
great, it's getting harder to see what all the flags are doing.

-- 
Looking at Perl through Lisp glasses, Perl looks atrocious.

LemonBoy:
> This sounds like a good way to do this, especially for new users.

I've implemented this under a new option named `guidarkmode`, I'll push the code as soon as possible.
What I'm not sure about is how this option should interact with the `d` specifier in `guioptions`, should the latter take over the former when specified?

zewpo:
Oh, we were working on something related to this before
#11899 

But, I've been moving house and I dropped the ball on this.  Anyway, its good to see some movement in this area.  When I get my new office set up in my new place, then I'll check in again to see if I can contribute anything that I learned about this stuff.

chrisbra:
can you please rebase?

chrisbra:
any update here @LemonBoy ?

chrisbra:
I am going to close this for the lack of activity.

PR #12284, opened 2023-04-20
syntax documentation: typo fixup: 'indivisual' to 'individual'

jayaddison:
A small typo fixup, spotted while reading about syntax matching and highlighting rules.


jayaddison:
Thank you!

PR #12286, opened 2023-04-21
Command-line is not cleared when popupmenu in it is closed

luukvbaal:
Problem:    Command-line is not cleared when popupmenu in it is closed.
Solution:   Clear command-line when popup menu in it is closed.


luukvbaal:
The problem:
image

After this patch the popupmenu may still be drawn over the statusline (but atleast this is redrawn when another mouse click is received so it does not show ghost popupmenus as opposed to the command line). Should the popupmenu be allowed to be drawn over the statusline?

zeertzjq:
When the window heights aren't large enough a popup menu will have to be drawn over statusline anyway.

luukvbaal:
> When the window heights aren't large enough a popup menu will have to be drawn over statusline anyway.

Yeah I guess, same could be said for the cmdline though. Would the preferred solution be to clear the cmdline instead?

brammool:
Please give reproduction steps for that screenshot.

For the patch, please add a regression test (using a screendump should work best).

luukvbaal:
> Please give reproduction steps for that screenshot.

The required commands to reproduce are in the screenshot, after which clicking near the command line multiple times will result in the depicted problem.

> For the patch, please add a regression test (using a screendump should work best).

Yeah I can do that but I think we should first determine what the desired behavior is:

* clear the cmdline area (when the popupmenu will be drawn there) so that the previous popopmenu is cleared
* avoid drawing the popupmenu in the cmdline area (the current patch)
* avoid drawing the popopmenu over the statusline (and cmdline)
*

brammool:
> > Please give reproduction steps for that screenshot.
> 
> The required commands to reproduce are in the screenshot, after which
> clicking near the command line multiple times will result in the
> depicted problem.

You mean right-clicking.  Keep in mind that showing a screenshot means I
have to type the commands (hopefully without a typo) to be able to try
it out.  Not really encouraging to take the time for this.

> > For the patch, please add a regression test (using a screendump
> > should work best).
> 
> Yeah I can do that but I think we should first determine what the
> desired behavior is:
> 
> * clear the cmdline area (when the popupmenu will be drawn there) so that the previous popopmenu is cleared
> * avoid drawing the popupmenu in the cmdline area (the current patch)
> * avoid drawing the popopmenu over the statusline (and cmdline)

The popup menu can have many entries, thus using all the space available
is best.  Besides that having the popup menu appear in a predictable
position is also relevant.  If it suddenly is different from where the
mouse click was this may be annoying for the user.

We can't avoid drawing it over a statusline when there are several split
windows.  Thus the main problem is that the command line isn't
redrawn/cleared when the popup menu is undrawn.

Actually redrawing the command line is only possible in command-line
mode.  Otherwise clearing it is really the only way.  Unfortunately this
will mean any messages will be gone.  You may be able to use some of the
logic in the redraw_after_callback() function (don't copy the code,
refactor it if needed, so that any further improvements are shared).

I would prefer trying this first.  If it's not giving us an acceptable
solution, then the alternative is to avoid drawing over the command
line.

-- 
ARTHUR: CHARGE!
   [The mighty ARMY charges.  Thundering noise of feet.  Clatter of coconuts.
   Shouts etc.   Suddenly there is a wail of a siren and a couple of police
   cars roar round in front of the charging ARMY and the POLICE leap out and
   stop them.  TWO POLICEMAN and the HISTORIAN'S WIFE.  Black Marias skid up
   behind them.]
HISTORIAN'S WIFE: They're the ones, I'm sure.

luukvbaal:
> Keep in mind that showing a screenshot means I have to type the commands (hopefully without a typo) to be able to try it out. Not really encouraging to take the time for this.

Of course (I didn't aks you to spend time on trying to reproduce it from the picture, just posted it to show te problem). Like I said, I refrained from adding a testcase/repro example because I'm willing to add it myself but the desired behavior wasn't clear yet...

luukvbaal:
> The popup menu can have many entries, thus using all the space available
is best.  Besides that having the popup menu appear in a predictable
position is also relevant.

We are just changing the grow direction in the current patch, the position is still predictable: at the mouse cursor. I think it is also a fair assumption that there is usually more space available above the command line than below. So I think the current patch might actually the best solution? It avoids unnecessarily clearing the command line and I think it is more likely to increase the chances of being able to draw the full popupmenu than decrease when the window is small.

luukvbaal:
Hmm but I guess even with the current patch the menu may spill over into the cmdline area if the window is small enough when clicking further away from the cmdline. I suppose clearing is necessary then.

luukvbaal:
I'm trying to write a test that shows the problem on master but it doesn't seem to work:
```vim
" Test for right-clicking near command-line
func Test_mousemodel_menu_cmdline()
  CheckRunVimInTerminal

  let lines =<< trim END
    set mousemodel=popup noruler
    menu PopUp.Test1 :<CR>
    menu PopUp.Test2 :<CR>
    menu PopUp.Test3 :<CR>
    call setline(1, repeat(['abcde'], 5))
    noremap! <expr> ] "<cmd>call test_setmouse(1, 5)<CR>"
  END
  call writefile(lines, 'Xpopupcmdline', 'D')
  let buf = RunVimInTerminal('-S Xpopupcmdline', {'rows': 4})

  call term_sendkeys(buf, "\<RightMouse>\<RightRelease>")
  call term_sendkeys(buf, "]")
  call term_sendkeys(buf, "\<RightMouse>\<RightRelease>")
  call VerifyScreenDump(buf, 'Test_mousemodel_menu_cmdline', {})

  call StopVimInTerminal(buf)
endfunc
```

The mapping to move the mouse to open a second menu at a different position doesn't seem to go through. What mapping mode is active while a popupmenu is open? Or how else can I move the mouse for this test?

brammool:
> Hmm but I guess even with the current patch the menu may spill over
> into the cmdline area if the window is small enough when clicking
> further away from the cmdline. I suppose clearing is necessary then.

Well, I don't want to ask you to make it too complicated, but finding a
balance between showing sufficient entries and not having to clear the
command line would be nice.  E.g. do use the space of the command line
if the menu would only show three or less entries otherwise.

-- 
NEIL INNES PLAYED: THE FIRST SELF-DESTRUCTIVE MONK, ROBIN'S LEAST FAVORITE
                   MINSTREL, THE PAGE CRUSHED BY A RABBIT, THE OWNER OF A DUCK

brammool:
> I'm trying to write a test that shows the problem on master but it
> doesn't seem to work:
> ```vim
> " Test for right-clicking near command-line
> func Test_mousemodel_menu_cmdline()
>   CheckRunVimInTerminal
> 
>   let lines =<< trim END
>     set mousemodel=popup noruler
>     menu PopUp.Test1 :<CR>
>     menu PopUp.Test2 :<CR>
>     menu PopUp.Test3 :<CR>
>     call setline(1, repeat(['abcde'], 5))
>     noremap! <expr> ] "<cmd>call test_setmouse(1, 5)<CR>"
>   END
>   call writefile(lines, 'Xpopupcmdline', 'D')
>   let buf = RunVimInTerminal('-S Xpopupcmdline', {'rows': 4})
> 
>   call term_sendkeys(buf, "\<RightMouse>\<RightRelease>")
>   call term_sendkeys(buf, "]")
>   call term_sendkeys(buf, "\<RightMouse>\<RightRelease>")
>   call VerifyScreenDump(buf, 'Test_mousemodel_menu_cmdline', {})
> 
>   call StopVimInTerminal(buf)
> endfunc
> ```
> 
> The mapping to move the mouse to open a second menu at a different
> position doesn't seem to go through. What mapping mode is active while
> a popupmenu is open? Or how else can I move the mouse for this test?

I haven't tried this, but these mouse clicks can indeed be a bit tricky.
First of all, make sure they are not recognized as double-click.  You
can try adding some ":sleep" commands to see what happens, then add a
proper WaitFor() or WaitForAssert() call to avoid the arbitrary wait
time.

Your mapping for "]" should not be an expression mapping, the argument
is not an expression but a "command mapping".

Your first <RightMouse> click is not preceded with a test_setmouse()
call.  Did you want to use some other position?

Use "set noruler" in the script causes trouble for RunVimInTerminal(),
it won't be able to see when startup is done.  Disable the ruler with
term_sendkeys().

-- 
ERIC IDLE PLAYED: THE DEAD COLLECTOR, MR BINT (A VILLAGE NE'ER-DO -WELL VERY
                  KEEN ON BURNING WITCHES), SIR ROBIN, THE GUARD WHO DOESN'T
                  HICOUGH BUT TRIES TO GET THINGS STRAIGHT, CONCORDE (SIR
                  LAUNCELOT'S TRUSTY STEED), ROGER THE SHRUBBER (A SHRUBBER),
                  BROTHER MAYNARD

luukvbaal:
Opening multiple menus is not necessary to show the problem, this works, I'll look into fixing it now...

```vim
" Test for right-clicking near command-line
func Test_mousemodel_menu_cmdline()
  CheckRunVimInTerminal

  let lines =<< trim END
    set mousemodel=popup
    menu PopUp.Test1 :<CR>
    menu PopUp.Test2 :<CR>
    menu PopUp.Test3 :<CR>
    call setline(1, repeat(['abcde'], 5))
  END
  call writefile(lines, 'Xpopupcmdline', 'D')
  let buf = RunVimInTerminal('-S Xpopupcmdline', {'rows': 4})

  " cmdline area should be cleared when popupmenu that was in it is closed
  call term_sendkeys(buf, "\<RightMouse>\<RightRelease>\<Esc>")
  call VerifyScreenDump(buf, 'Test_mousemodel_menu_cmdline', {})

  call StopVimInTerminal(buf)
endfunc
```

luukvbaal:
> Well, I don't want to ask you to make it too complicated, but finding a
balance between showing sufficient entries and not having to clear the
command line would be nice.

I also don't want to make it too complicated. The latest commit fixes the problem by clearing the command line only when a menu is closed that was (partially) in the command line area. The problem is not exclusive to the right mouse button `PopUp` menu so I updated the commit message.

luukvbaal:
> I also don't want to make it too complicated.

Besides, the logic already in place seems sufficient to handle the grow direction/size constraints. Now we just clear the command line when necessary which even if it contained a message would already have been "corrupted" by the popup menu.

luukvbaal:
Writing up [this](https://github.com/neovim/neovim/issues/23275) neovim issue (where the problem is slightly different) reaffirmed my believe that the initially proposed patch might be a better solution.

The chance that drawing the popup-menu downward, and thus potentially into the cmdline area, will result in more entries being shown than when drawing upward seems very small (only for very small window size or large cmdheight). Thus, is it not better to flip the grow direction as soon as the popup menu would draw into the cmdline area so that we can avoid being forced to clear the command line area (potentially showing a message)?

brammool:
> > Well, I don't want to ask you to make it too complicated, but finding a
> > balance between showing sufficient entries and not having to clear the
> > command line would be nice.
> 
> I also don't want to make it too complicated. The latest commit fixes
> the problem by clearing the command line only when a menu is closed
> that was (partially) in the command line area. The problem is not
> exclusive to the right mouse button `PopUp` menu so I updated the
> commit message.

OK, it's nice to keep it simple.  If users complain we can further tune
it.

-- 
Every time I lose weight, it finds me again!

PR #12292, opened 2023-04-24
allow undotree() to return tree for any buffer

### LINKED ISSUE #4001: undotree() in context of another window
I know `undotree()` returns a RO undo tree, but is it possible to be able to pass a `bufnr()` (or a `winnr()`) to it? It is impossible to get the undotree for another buffer or window. It would be nice if it could be a read-only variable inside `&undotree`, but I am not sure if that is possible (so that one could use `getbufvar()` and `getwinvar()`)

djpohly:
Problem:    undotree() does not support getting the undo tree for any
            buffer other than the current one.
Solution:   Add an optional "buffer number" parameter to undotree().  If
            omitted, use the current buffer for backwards compatibility.

Fixes #4001.

### DIFF
--- a/src/testdir/test_undo.vim
+++ b/src/testdir/test_undo.vim
@@ -93,6 +93,48 @@ func FillBuffer()
   endfor
 endfunc
 
+func Test_undotree_bufnr()
+  new
+  let buf1 = bufnr()
+
+  normal! Aabc
+  set ul=100
+
+  " Save undo tree without bufnr as ground truth for buffer 1
+  let d1 = undotree()
+
+  new
+  let buf2 = bufnr()
+
+  normal! Adef
+  set ul=100
+
+  normal! Aghi
+  set ul=100
+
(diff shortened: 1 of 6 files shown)


src/testdir/test_undo.vim
```diff
@@ -93,6 +93,48 @@ func FillBuffer()
...
+  " Switch buffers and check again
+  wincmd p
+
+  let d = undotree(buf1)
+  call assert_equal(d1, d)
+
+  let d = undotree(buf2)
+  call assert_notequal(d1, d)
+  call assert_equal(d2, d)
+endfunc
```
zeertzjq, line 136:
```suggestion

  bwipe!
  bwipe!
  set ul&
endfunc
```

zeertzjq, line 141:
```suggestion

  " Drop created windows
  set ul&
  new
  only!
endfunc
```

yegappan:

PR #12295, opened 2023-04-25
Change the LSP HTTP header Content-Type field

yegappan:
Based on the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart), change the HTTP header Content-Type field from vim-jsonrpc to vscode-jsonrpc.

Some LSP servers (e.g. https://github.com/hashicorp/terraform-ls), compare the value of the Content-Type field against
vscode-jsonrpc and reject the LSP message if it is set to some other value.

### DIFF
--- a/src/json.c
+++ b/src/json.c
@@ -105,7 +105,7 @@ json_encode_lsp_msg(typval_T *val)
     ga_init2(&lspga, 1, 4000);
     vim_snprintf((char *)IObuff, IOSIZE,
 	    "Content-Length: %u\r\n"
-	    "Content-Type: application/vim-jsonrpc; charset=utf-8\r\n\r\n",
+	    "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n\r\n",
 	    ga.ga_len - 1);
     ga_concat(&lspga, IObuff);
     ga_concat_len(&lspga, ga.ga_data, ga.ga_len);
--- a/src/testdir/test_channel.vim
+++ b/src/testdir/test_channel.vim
@@ -2670,7 +2670,7 @@ func LspTests(port)
   " " Test for sending a raw message
   " let g:lspNotif = []
   " let s = "Content-Length: 62\r\n"
-  " let s ..= "Content-Type: application/vim-jsonrpc; charset=utf-8\r\n"
+  " let s ..= "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n"
   " let s ..= "\r\n"
   " let s ..= '{"method":"echo","jsonrpc":"2.0","params":{"m":"raw-message"}}'
   " call ch_sendraw(ch, s)
--- a/src/testdir/test_channel_lsp.py
+++ b/src/testdir/test_channel_lsp.py
@@ -35,7 +35,7 @@ class ThreadedTCPRequestHandler(socketserver.BaseRequestHandler):
             v['id'] = msgid
         s = json.dumps(v)
         resp = "Content-Length: " + str(len(s)) + "\r\n"
-        resp += "Content-Type: application/vim-jsonrpc; charset=utf-8\r\n"
+        resp += "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n"
         resp += "\r\n"
         resp += s
         if self.debug:
@@ -46,7 +46,7 @@ class ThreadedTCPRequestHandler(socketserver.BaseRequestHandler):
         v = 'wrong-payload'
         s = json.dumps(v)
         resp = "Content-Length: " + str(len(s)) + "\r\n"
-        resp += "Content-Type: application/vim-jsonrpc; charset=utf-8\r\n"
+        resp += "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n"
         resp += "\r\n"
         resp += s
         self.request.sendall(resp.encode('utf-8'))
@@ -60,7 +60,7 @@ class ThreadedTCPRequestHandler(socketserver.BaseRequestHandler):
 
     def send_empty_payload(self):
         resp = "Content-Length: 0\r\n"
-        resp += "Content-Type: application/vim-jsonrpc; charset=utf-8\r\n"
+        resp += "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n"
         resp += "\r\n"
         self.request.sendall(resp.encode('utf-8'))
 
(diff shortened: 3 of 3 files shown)


Shane-XB-Qian:
seems it just says `Defaults to application/vscode-jsonrpc; charset=utf-8`, but did not say `Must be to` ?

Shane-XB-Qian:
the specification just default to vscode-jsonrpc, so did not mean vim-jsonrpc was incorrect, IMO.
// so this modification perhaps is unnecessary, which server side perhaps should to compliance it better.

yegappan:
We are not modifying the language server specification and adding anything Vim specific to the protocol or the headers.
So it is better to follow the specification and use the default value for the Content-Type field (to gain maximum compatibility
with the existing language servers).

Shane-XB-Qian:
IMO: the specification just tried to identify it is jsonrpc,that should be the maximum compatibility, (vs 'vscode-' or 'vim-' should be just a prefix), if the lsp server hard code to vscode-jsonrpc only, looks it was its flaw :-)

skywind3000:
I belief that the responsible course of action is to raise issues with LSP servers in order to ensure that they are capable of accommodating non-VSCode clients, rather than modifying the `Content-type` ourselves. 

LSP servers should be cognizant of the fact that they are not exclusively serving VSCode, and should strive to maintain maximum compatibility with other clients. 

The `Content-type` serves as a means of identifying client type, and therefore should not be altered to `vscode-jsonapi`, as this is merely a default value in the specification and not a mandatory requirement.

skywind3000:
at least, make it default to `vim-jsonapi` and provide a way to change it like:

```vimL
set lsptype=vscode-jsonapi
```

or something.

brammool:
Yegappan wrote:

> We are not modifying the language server specification and adding
> anything Vim specific to the protocol or the headers.
> So it is better to follow the specification and use the default value
> for the Content-Type field (to gain maximum compatibility
> with the existing language servers).

In other words: we handle the vscode implementation of the
specification.  I don't know what VScode-specific things there are, but
it's probably best to do it that way, we can expect it's widely
accepted.

-- 
Kiss me twice.  I'm schizophrenic.

benknoble:
Don’t we now run the risk of all editors pretending to be VSCode, much like how [all the browsers pretend to be each other](https://humanwhocodes.com/blog/2010/01/12/history-of-the-user-agent-string/)?

vimpostor:
@yegappan Is there any reason why the `Content-Type` header is set in the first place? It causes problems with some LSP servers such as [haskell-language-server](https://github.com/haskell/haskell-language-server), which doesn't support that header and crashes with the following error when it receives that header (when using [your vim9 LSP client](https://github.com/yegappan/lsp)):

```
Error | Failed to parse message header:
:  string
```

Obviously we could fix that particular server, but I think there is incentive to not send that header at all, because:
- it is an optional header
- the value that we set (`Content-Type: application/vscode-jsonrpc; charset=utf-8`) [is already the default value](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart)
- other popular LSP clients also omit this header

I have tested the following LSP clients and not a single one of them sets the `Content-Type` header, only the `Content-Length` header is set for all of them:
- coc.nvim
- [Neovim LSP](https://github.com/neovim/neovim/blob/5d8ab32f3871b0232972cac1116ac7cba98389e5/runtime/lua/vim/lsp/rpc.lua#L22)
- VSCode
- [Ale](https://github.com/dense-analysis/ale/blob/115ad17ace047cab20ccc67f79c943aaf3f0f291/autoload/ale/lsp.vim#L149)
- YouCompleteMe

Therefore in order to improve compatibility with slightly misbehaving LSP clients, I propose to drop this header:

```diff
diff --git a/src/json.c b/src/json.c
index acf7ac57c..e2a011309 100644
--- a/src/json.c
+++ b/src/json.c
@@ -105,8 +105,7 @@ json_encode_lsp_msg(typval_T *val)
     ga_init2(&lspga, 1, 4000);
     // Header according to LSP specification.
     vim_snprintf((char *)IObuff, IOSIZE,
-	    "Content-Length: %u\r\n"
-	    "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n\r\n",
+	    "Content-Length: %u\r\n\r\n",
 	    ga.ga_len - 1);
     ga_concat(&lspga, IObuff);
     ga_concat_len(&lspga, ga.ga_data, ga.ga_len);
```

vim-ml:
Hi,

On Fri, Aug 25, 2023 at 4:26 PM Magnus Groß ***@***.***>
wrote:

> @yegappan <https://github.com/yegappan> Is there any reason why the
> Content-Type header is set in the first place? It causes problems with
> some LSP servers such as haskell-language-server
> <https://github.com/haskell/haskell-language-server>, which doesn't
> support that header and crashes with the following error when it receives
> that header (when using your vim9 LSP client
> <https://github.com/yegappan/lsp>):
>

I originally set it to follow the LSP specification.  But as we learned
over a period of time, some of
the LSP servers are not able to properly handle/ignore this header.


> Error | Failed to parse message header:
> :  string
>
> Obviously we could fix that particular server, but I think there is
> incentive to not send that header at all, because:
>
>    - it is an optional header
>    - the value that we set (Content-Type: application/vscode-jsonrpc;
>    charset=utf-8) is already the default value
>    <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart>
>    - other popular LSP clients also omit this header
>
> I have tested the following LSP clients and not a single one of them sets
> the Content-Type header, only the Content-Length header is set for all of
> them:
>
>    - coc.nvim
>    - Neovim LSP
>    <https://github.com/neovim/neovim/blob/5d8ab32f3871b0232972cac1116ac7cba98389e5/runtime/lua/vim/lsp/rpc.lua#L22>
>    - VSCode
>    - Ale
>    <https://github.com/dense-analysis/ale/blob/115ad17ace047cab20ccc67f79c943aaf3f0f291/autoload/ale/lsp.vim#L149>
>    - YouCompleteMe
>
> Therefore in order to improve compatibility with slightly misbehaving LSP
> clients, I propose to drop this header:
>
> diff --git a/src/json.c b/src/json.c
> index acf7ac57c..e2a011309 100644--- a/src/json.c+++ b/src/json.c@@ -105,8 +105,7 @@ json_encode_lsp_msg(typval_T *val)
>      ga_init2(&lspga, 1, 4000);
>      // Header according to LSP specification.
>      vim_snprintf((char *)IObuff, IOSIZE,-	    "Content-Length: %u\r\n"-	    "Content-Type: application/vscode-jsonrpc; charset=utf-8\r\n\r\n",+	    "Content-Length: %u\r\n\r\n",
>  	    ga.ga_len - 1);
>      ga_concat(&lspga, IObuff);
>      ga_concat_len(&lspga, ga.ga_data, ga.ga_len);
>
>
I don't see a problem in removing the Content-Type header.  Can you create
a PR with this change?

Thanks,
Yegappan

PR #12296, opened 2023-04-25
Fix: Crypt xchacha20v2 header size differs between 32-bit and 64-bit

ichizok:
## Problem:

`Test_uncrypt_xchacha20v2()` and `Test_uncrypt_xchacha20v2_custom()` in test_crypt fails on 32-bit environment.

## Cause:

Xchacha20v2 header includes the fields of `unsigned long long`, `size_t`, and `int`;
https://github.com/vim/vim/blob/cfc788c38848adcd54e4130f16f2813cd1e215dd/src/crypt.c#L918-L920

but the size of `size_t` is 4 on 32-bit (ILP32) and is 8 on 64-bit (LP64), thus, on between each environment, different header size and different encryption/decryption result.

### DIFF
--- a/src/crypt.c
+++ b/src/crypt.c
@@ -30,23 +30,22 @@
  */
 
 typedef struct {
-    char    *name;	// encryption name as used in 'cryptmethod'
-    char    *magic;	// magic bytes stored in file header
-    int	    salt_len;	// length of salt, or 0 when not using salt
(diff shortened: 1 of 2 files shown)


k-takata:
I've confirmed that the tests passes after applying this change.

chrisbra:
thanks for fixing this!

brammool:
Included with Patch 9.0.1488.
s/closses/closes/

PR #12298, opened 2023-04-25
Fix ModeChanged triggered by <MouseMove>

zeertzjq:


zeertzjq:
This also fixes another problem that changing the motion force of an operator triggers ModeChanged twice.

PR #12300, opened 2023-04-25
improve Device Tree file type support

rohieb:
* Add indent definitions for .dts and .dtsi files, because the default autoindent behaviour wrongly unindents properties starting with `#`, e.g. `#address-cells` and `#size-cells`, or labels ending in `:`.
* cc @zonque: Improve syntax highlighting for .dts and .dtsi files to include support for label-relative path references, which were [introduced in dtc 1.7](https://git.kernel.org/pub/scm/utils/dtc/dtc.git/commit/?id=ec7986e682cf9a5078bdbad341feb807a56b21e0). Currently this syntax is matched by the `dtsKeyword` element, so make that more specific to only match directives like `/dtc-v1/`, `/delete-node/` and the like. Screenshots below show before and after states, with an additional `hi link dtsReference Identifier` to distinguish them from other similarly-colored syntax elements.

Before:
667e2cc6-b795-42e2-8e5b-3d82f94d12e1

After:
9e7eff49-df0d-4414-a6e3-8c3d89134ab0

### DIFF
--- a/runtime/indent/dts.vim
+++ b/runtime/indent/dts.vim
@@ -1,0 +1,63 @@
+" Vim indent file
+" Language:		Device Tree
+" Maintainer:		Roland Hieber, Pengutronix <rhi@pengutronix.de>
+"
+if exists("b:did_indent")
+    finish
+endif
+let b:did_indent = 1
+
+setlocal autoindent
+setlocal nosmartindent
+setlocal indentkeys=o,O,0},0<>>,!<Ctrl-F>
+setlocal indentexpr=GetDTSIndent()
+setlocal nolisp
+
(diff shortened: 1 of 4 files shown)


rohieb:
Where would I add a test for indent files?

rohieb:
Ahh, I found runtime/indent/testdir/README.txt, that explains it!

rohieb:
Ahh good thing you told me to write tests, I actually found an error… 😅 Also added a `@` in `dtsReference` to match addresses in node names, like in `indent/testdir/dts.in`.

brammool:
I'll leave the syntax file changes to  @zonque
I'll include the indent file and test.

zonque (approved):
LGTM

brammool:
Thanks, I'll include it then.  I'll update the date in the header.

PR #12311, opened 2023-04-28
Add i18n support for percent value

bitigchi:
The percentage value on the initial Vim buffer did not support i18n before. This particular value is not simple to internationalize, since the positioning is absolute, and this causes problems for languages like Turkish that uses the `%100`, `%10`, `%1` percentage format, leading to outputs like `% 10` and/or `%  1`. This commit adds a helper to make i18n process accurately.

I don't think this is the most elegant solution however. Any comments to further improve the logic is most welcome.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -5231,8 +5231,8 @@ build_stl_str_hl(
 #endif // FEAT_STL_OPT
 
 /*
- * Get relative cursor position in window into "buf[buflen]", in the form 99%,
- * using "Top", "Bot" or "All" when appropriate.
+ * Get relative cursor position in window into "buf[buflen]", in the localized
+ * percentage form like %99, 99%; using "Top", "Bot" or "All" when appropriate.
  */
     void
 get_rel_pos(
@@ -5240,6 +5240,7 @@ get_rel_pos(
     char_u	*buf,
     int		buflen)
 {
+    int		per; // localized percentage value
     long	above; // number of lines above window
     long	below; // number of lines below window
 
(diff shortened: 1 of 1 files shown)


brammool:
There are several problems with this solution.  One of which is that memory leaks.

I would think only the printf string "%d%%" needs to be translated.  The result can be right-aligned
in the available space, e.g. by first calling vim_snprintf(), getting the length and then when it's too short
prepend a space and repeat.  Or compute the percentage, and when it smaller than 10 insert a space.
Translation should never change the decimal number value.

bitigchi:
> I would think only the printf string "%d%%" needs to be translated. The result can be right-aligned in the available space, e.g. by first calling vim_snprintf(), getting the length and then when it's too short prepend a space and repeat. Or compute the percentage, and when it smaller than 10 insert a space. Translation should never change the decimal number value.

Thanks for the tips! Indeed the updated solution turned out to be much more concise. I am still not sure about memory management though, since I haven't used C for anything other than snippets like these. Please let me know if it's not okay!

brammool:
> > I would think only the printf string "%d%%" needs to be translated.
> > The result can be right-aligned in the available space, e.g. by
> > first calling vim_snprintf(), getting the length and then when it's
> > too short prepend a space and repeat. Or compute the percentage, and
> > when it smaller than 10 insert a space. Translation should never
> > change the decimal number value.
> 
> Thanks for the tips! Indeed the updated solution turned out to be much
> more concise. I am still not sure about memory management though,
> since I haven't used C for anything other than snippets like these.
> Please let me know if it's not okay!

The new version doesn't allocate memory, thus that's taken care of.

We can put the space in in the code, so that there is no need for two
translated text entries.  I'll make it like that, please check.

-- 
Hanson's Treatment of Time:
	There are never enough hours in a day, but always too
	many days before Saturday.

PR #12313, opened 2023-04-29
Fix compiler warning about sigset()

h-east:
The compiler outputs 21 warnings about `sigset()` in os_unix.c.
Below is one of them.

```bash
os_unix.c: In function 'sig_winch':
os_unix.c:872:5: warning: 'sigset' is deprecated: Use the signal and sigprocmask functions instead [-Wdeprecated-declarations]
  872 |     signal(SIGWINCH, (void (*)(int))sig_winch);
      |     ^~~~~~
In file included from /usr/include/x86_64-linux-gnu/sys/param.h:28,
                 from os_unix.h:60,
                 from vim.h:244,
                 from os_unix.c:21:
/usr/include/signal.h:367:23: note: declared here
  367 | extern __sighandler_t sigset (int __sig, __sighandler_t __disp) __THROW
      |                       ^~~~~~
```

## My env.
```
$ cat /etc/os-release | grep "VERSION="
VERSION="22.04.2 LTS (Jammy Jellyfish)"

$ gcc --version
gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0

$ ldd --version
ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35
```

## Cause code

[os_unixx.h: L16]
```c
#if defined(HAVE_SIGSET) && !defined(signal)
# define signal sigset
#endif
```

`signal` is replaced with `sigset` here.
1. `!defined(signal)`  is not right. This does not determine whether there is a `signal()` function or not.
2.  `sigset()` is obsolete than `signal()`.
Both are obsoleted, but my gcc outputs warnings only for `sigset()`.

Ubuntu Manpage: 
[sigset(3)](https://manpages.ubuntu.com/manpages/kinetic/man2/sighold.2freebsd.html)

> This interface is made obsolete by sigsuspend(2) and sigaction(2).

[signal(2)](https://manpages.ubuntu.com/manpages/kinetic/en/man2/signal.2.html)
  
> WARNING:  the  behavior  of  signal()  varies  across  UNIX  versions, and has also varied historically across different versions of Linux.  Avoid its use: use sigaction(2) instead.  See Portability below.

Archlinux Manpage: 
[sigset(3)](https://man.archlinux.org/man/sigset.3.en)

> These functions are provided in glibc as a compatibility interface for programs that make use of the historical System V signal API. This API is obsolete: new applications should use the POSIX signal API (sigaction(2), sigprocmask(2), etc.)

[signal(2)](https://man.archlinux.org/man/signal.2)

> WARNING: the behavior of signal() varies across UNIX versions, and has also varied historically across different versions of Linux. Avoid its use: use sigaction(2) instead. See Portability below.

## Solution

In this PR, I just stopped using `sigset()`.
The next step is to replace `signal()` with `sigaction()`, I think. (Wrapping it with `mch_signal()`?)

--
Best regards,
Hirohito Higashi (h_east)

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -12525,7 +12525,7 @@ fi
 for ac_func in fchdir fchown fchmod fsync getcwd getpseudotty \
 	getpwent getpwnam getpwuid getrlimit gettimeofday localtime_r lstat \
 	memset mkdtemp nanosleep opendir putenv qsort readlink select setenv \
-	getpgid setpgid setsid sigaltstack sigstack sigset sigsetjmp sigaction \
+	getpgid setpgid setsid sigaltstack sigstack sigsetjmp sigaction \
 	sigprocmask sigvec strcasecmp strcoll strerror strftime stricmp strncasecmp \
 	strnicmp strpbrk strptime strtol tgetent towlower towupper iswupper \
 	tzset usleep utime utimes mblen ftruncate unsetenv posix_openpt \
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -12525,7 +12525,7 @@ fi
 for ac_func in fchdir fchown fchmod fsync getcwd getpseudotty \
 	getpwent getpwnam getpwuid getrlimit gettimeofday localtime_r lstat \
 	memset mkdtemp nanosleep opendir putenv qsort readlink select setenv \
-	getpgid setpgid setsid sigaltstack sigstack sigsetjmp sigaction \
+	getpgid setpgid setsid sigaltstack sigstack sigset sigsetjmp sigaction \
 	sigprocmask sigvec strcasecmp strcoll strerror strftime stricmp strncasecmp \
 	strnicmp strpbrk strptime strtol tgetent towlower towupper iswupper \
 	tzset usleep utime utimes mblen ftruncate unsetenv posix_openpt \
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -3706,7 +3706,7 @@ dnl Can only be used for functions that do not require any include.
 AC_CHECK_FUNCS(fchdir fchown fchmod fsync getcwd getpseudotty \
 	getpwent getpwnam getpwuid getrlimit gettimeofday localtime_r lstat \
 	memset mkdtemp nanosleep opendir putenv qsort readlink select setenv \
-	getpgid setpgid setsid sigaltstack sigstack sigsetjmp sigaction \
+	getpgid setpgid setsid sigaltstack sigstack sigset sigsetjmp sigaction \
 	sigprocmask sigvec strcasecmp strcoll strerror strftime stricmp strncasecmp \
 	strnicmp strpbrk strptime strtol tgetent towlower towupper iswupper \
 	tzset usleep utime utimes mblen ftruncate unsetenv posix_openpt \
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -3706,7 +3706,7 @@ dnl Can only be used for functions that do not require any include.
 AC_CHECK_FUNCS(fchdir fchown fchmod fsync getcwd getpseudotty \
 	getpwent getpwnam getpwuid getrlimit gettimeofday localtime_r lstat \
 	memset mkdtemp nanosleep opendir putenv qsort readlink select setenv \
-	getpgid setpgid setsid sigaltstack sigstack sigset sigsetjmp sigaction \
+	getpgid setpgid setsid sigaltstack sigstack sigsetjmp sigaction \
 	sigprocmask sigvec strcasecmp strcoll strerror strftime stricmp strncasecmp \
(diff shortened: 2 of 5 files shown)


brammool:
> The compiler outputs 21 warnings about `sigset()` in os_unix.c.
> Below is one of them.
> 
> ```bash
> os_unix.c: In function 'sig_winch':
> os_unix.c:872:5: warning: 'sigset' is deprecated: Use the signal and sigprocmask functions instead [-Wdeprecated-declarations]
>   872 |     signal(SIGWINCH, (void (*)(int))sig_winch);
>       |     ^~~~~~
> In file included from /usr/include/x86_64-linux-gnu/sys/param.h:28,
>                  from os_unix.h:60,
>                  from vim.h:244,
>                  from os_unix.c:21:
> /usr/include/signal.h:367:23: note: declared here
>   367 | extern __sighandler_t sigset (int __sig, __sighandler_t __disp) __THROW
>       |                       ^~~~~~
> ```
> 
> ## My env.
> ```
> $ cat /etc/os-release | grep "VERSION="
> VERSION="22.04.2 LTS (Jammy Jellyfish)"
> 
> $ gcc --version
> gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
> 
> $ ldd --version
> ldd (Ubuntu GLIBC 2.35-0ubuntu3.1) 2.35
> ```
> 
> ## Cause code
> 
> [os_unixx.h: L16]
> ```c
> #if defined(HAVE_SIGSET) && !defined(signal)
> # define signal sigset
> #endif
> ```
> 
> `signal` is replaced with `sigset` here.
> 1. `!defined(signal)`  is not right. This does not determine whether
> there is a `signal()` function or not.

I think it was right.  This is not to check whether a signal() function
exists (it nearly always does exist) but whether the macro already
exists.  Whether signal() or sigset() works better heavily depends on
the system used.

> 2.  `sigset()` is obsolete than `signal()`.
> Both are obsoleted, but my gcc outputs warnings only for `sigset()`.

This is only on Ubuntu, it probably also applies to other recent Linux
systems.

> Ubuntu Manpage: 

Keep in mind that Linux manpages can only be trusted for Linux.  Whether
they accurately describe other systems isn't guaranteed anywhere.  There
simply are too many different Unix-like systems, each with their own
problems (aka bugs).  Linux man page maintainers may or may not have
taken the time to dive into this.

> [sigset(3)](https://manpages.ubuntu.com/manpages/kinetic/man2/sighold.2freebsd.html)
> 
> > This interface is made obsolete by sigsuspend(2) and sigaction(2).
> 
> [signal(2)](https://manpages.ubuntu.com/manpages/kinetic/en/man2/signal.2.html)
>   
> > WARNING:  the  behavior  of  signal()  varies  across  UNIX  versions, and has also varied historically across different versions of Linux.  Avoid its use: use sigaction(2) instead.  See Portability below.
> 
> Archlinux Manpage: 
> [sigset(3)](https://man.archlinux.org/man/sigset.3.en)
> 
> > These functions are provided in glibc as a compatibility interface for programs that make use of the historical System V signal API. This API is obsolete: new applications should use the POSIX signal API (sigaction(2), sigprocmask(2), etc.)
> 
> [signal(2)](https://man.archlinux.org/man/signal.2)
> 
> > WARNING: the behavior of signal() varies across UNIX versions, and has also varied historically across different versions of Linux. Avoid its use: use sigaction(2) instead. See Portability below.

Although these warnings may be useful when writing new code, the Vim
implementation has existed for a very long time.  If there would be an
actual problem with how signal() and related functions are used, I would
be more eager to make changes.  But since I'm not aware of actual
problems, making changes has a risk of introducing problems instead of
avoiding them.

> ## Solution
> 
> In this PR, I just stopped using `sigset()`.

I do not like the way this is done.  Since the warnings are for Linux it
would be good to only make the changes for Linux.

> The next step is to replace `signal()` with `sigaction()`, I think.
> (Wrapping it with `mch_signal()`?)

Only when we know the sigaction() implementation on all relevant system
is actually working the way we use it.  Considering the many mistakes
made in the past, it's not unlikely that new solutions also have
mistakes.  I would not be surprised if the problems mentioned in those
warnings are taken care of, while things that were working fine before
are now broken.  It most likely requires using a configure check that
the system has proper sigaction() support.  I have no idea how to test
that though.  Does this depend on a kernel version?  glibc version?

-- 
Any resemblance between the above views and those of my employer, my terminal,
or the view out my window are purely coincidental.  Any resemblance between
the above and my own views is non-deterministic.  The question of the
existence of views in the absence of anyone to hold them is left as an
exercise for the reader.  The question of the existence of the reader is left
as an exercise for the second god coefficient.  (A discussion of
non-orthogonal, non-integral polytheism is beyond the scope of this article.)
						(Ralph Jennings)

h-east:
@brammool 
Thanks for the reply. I understand.
It seems that test_suspend is failing in some cases on Linux as well.
it's difficult... Closing.

PR #12321, opened 2023-04-30
Add support for the diff() function using the built-in diff support

yegappan:
Add support for computing diff between two Lists of strings using a function.

When using a LSP plugin, whenever a buffer is modified, the modifications are sent to the language server.
To do this, the LSP plugins have implemented computing the diff in Vimscript or in Lua:

https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/utils/diff.vim
https://github.com/natebosch/vim-lsc/blob/master/autoload/lsc/diff.vim
https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/sync.lua

These computations are expensive as they compare every character and line in a buffer.
This new function will help in efficiently computing the diff.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -147,6 +147,8 @@ delete({fname} [, {flags}])	Number	delete the file or directory {fname}
 deletebufline({buf}, {first} [, {last}])
 				Number	delete lines from buffer {buf}
 did_filetype()			Number	|TRUE| if FileType autocmd event used
+diff({fromlist}, {tolist} [, {options}])
+				List	diff two Lists of strings
 diff_filler({lnum})		Number	diff filler lines about {lnum}
 diff_hlID({lnum}, {col})	Number	diff highlighting at {lnum}/{col}
 digraph_get({chars})		String	get the |digraph| of {chars}
@@ -2046,6 +2048,67 @@ did_filetype()	Returns |TRUE| when autocommands are being executed and the
 		editing another buffer to set 'filetype' and load a syntax
 		file.
 
+diff({fromlist}, {tolist} [, {options}])		*diff()*
+		Returns a String or a List containing the diff between the
+		strings in {fromlist} and {tolist}.  Uses the Vim internal
+		diff library to compute the diff.
+
+							*E106*
+		The optional "output" item in {options} specifies the returned
+		diff format.  The following values are supported:
+		    indices	Return a List of the starting and ending
+				indices and a count of the strings in each
+				diff hunk.
+		    unified	Return the unified diff output as a String.
+				This is the default.
+
+		If the "output" item in {options} is "indices", then a List is
+		returned.  Each List item contains a Dict with the following
+		items for each diff hunk:
+		    from_idx	start index in {fromlist} for this diff hunk.
+		    from_count	number of strings in {fromlist} that are
+				added/removed/modified in this diff hunk.
+		    to_idx	start index in {tolist} for this diff hunk.
+		    to_count	number of strings in {tolist} that are
+				added/removed/modified in this diff hunk.
+
+		The {options} Dict argument also specifies diff options
+		(similar to 'diffopt') and supports the following items:
+		    iblank		ignore changes where lines are all
+					blank.
+		    icase		ignore changes in case of text.
+		    iwhite		ignore changes in amount of white
+					space.
+		    iwhiteall		ignore all white space changes.
+		    iwhiteeol		ignore white space changes at end of
+					line.
+		    indent-heuristic	use the indent heuristic for the
+					internal diff library.
+		    algorithm		Dict specifying the diff algorithm to
+					use.  Supported boolean items are
+					"myers", "minimal", "patience" and
+					"histogram".
+		For more information about these options, refer to 'diffopt'.
+
+		Returns an empty List or String if {fromlist} and {tolist} are
+		identical.
+
+		Examples:
+		    :echo diff(['abc'], ['xxx'])
+		     @@ -1 +1 @@
+		     -abc
+		     +xxx
+
+		    :echo diff(['abc'], ['xxx'], {'output': 'indices'})
+		     [{'from_idx': 0, 'from_count': 1, 'to_idx': 0, 'to_count': 1}]
+		    :echo diff(readfile('oldfile'), readfile('newfile'))
+		    :echo diff(getbufline(5, 1, '$'), getbufline(6, 1, '$'))
+
+		For more examples, refer to |diff-func-examples|
+
+		Can also be used as a |method|: >
+			GetFromList->diff(to_list)
+<
 diff_filler({lnum})					*diff_filler()*
 		Returns the number of filler lines above line {lnum}.
 		These are the lines that were inserted at this point in
--- a/runtime/doc/diff.txt
+++ b/runtime/doc/diff.txt
@@ -476,4 +476,43 @@ Otherwise, the expression is evaluated in the context of the script where the
 option was set, thus script-local items are available.
 
 
+DIFF FUNCTION EXAMPLES				*diff-func-examples*
+
+Some examples for using the |diff()| function to compute the diff indices
+between two Lists of strings are below.
+>
+  " some lines are changed
+  :echo diff(['abc', 'def', 'ghi'], ['abx', 'rrr', 'xhi'], {'output': 'indices'})
+   [{'from_idx': 0, 'from_count': 3, 'to_idx': 0, 'to_count': 3}]
+
+  " few lines added at the beginning
+  :echo diff(['ghi'], ['abc', 'def', 'ghi'], {'output': 'indices'})
+   [{'from_idx': 0, 'from_count': 0, 'to_idx': 0, 'to_count': 2}]
+
+  " few lines removed from the beginning
+  :echo diff(['abc', 'def', 'ghi'], ['ghi'], {'output': 'indices'})
+   [{'from_idx': 0, 'from_count': 2, 'to_idx': 0, 'to_count': 0}]
+
+  " few lines added in the middle
+  :echo diff(['abc', 'jkl'], ['abc', 'def', 'ghi', 'jkl'], {'output': 'indices'})
+   [{'from_idx': 1, 'from_count': 0, 'to_idx': 1, 'to_count': 2}]
+
+  " few lines removed in the middle
+  :echo diff(['abc', 'def', 'ghi', 'jkl'], ['abc', 'jkl'], {'output': 'indices'})
+   [{'from_idx': 1, 'from_count': 2, 'to_idx': 1, 'to_count': 0}]
+
+  " few lines added at the end
+  :echo diff(['abc'], ['abc', 'def', 'ghi'], {'output': 'indices'})
+   [{'from_idx': 1, 'from_count': 0, 'to_idx': 1, 'to_count': 2}]
+
+  " few lines removed from the end
+  :echo diff(['abc', 'def', 'ghi'], ['abc'], {'output': 'indices'})
+   [{'from_idx': 1, 'from_count': 2, 'to_idx': 1, 'to_count': 0}]
+
+  " disjointed changes
+  :echo diff(['ab', 'def', 'ghi', 'jkl'], ['abc', 'def', 'ghi', 'jk'], {'output': 'indices'})
+   [{'from_idx': 0, 'from_count': 1, 'to_idx': 0, 'to_count': 1},
+    {'from_idx': 3, 'from_count': 1, 'to_idx': 3, 'to_count': 1}]
+<
+
  vim:tw=78:ts=8:noet:ft=help:norl:
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -4137,6 +4137,7 @@ E1056	vim9.txt	/*E1056*
 E1057	vim9.txt	/*E1057*
 E1058	vim9.txt	/*E1058*
 E1059	vim9.txt	/*E1059*
+E106	builtin.txt	/*E106*
 E1060	vim9.txt	/*E1060*
 E1061	vim9.txt	/*E1061*
 E1062	eval.txt	/*E1062*
@@ -6759,7 +6760,9 @@ dict-identity	eval.txt	/*dict-identity*
 dict-modification	eval.txt	/*dict-modification*
 did_filetype()	builtin.txt	/*did_filetype()*
 diff	diff.txt	/*diff*
+diff()	builtin.txt	/*diff()*
 diff-diffexpr	diff.txt	/*diff-diffexpr*
+diff-func-examples	diff.txt	/*diff-func-examples*
 diff-mode	diff.txt	/*diff-mode*
 diff-options	diff.txt	/*diff-options*
 diff-original-file	diff.txt	/*diff-original-file*
--- a/runtime/doc/todo.txt
+++ b/runtime/doc/todo.txt
@@ -956,9 +956,6 @@ When 'sidescrolloff' is set, using "zl" to go to the end of the line, suddenly
 scrolls back.  Should allow for this scrolling, like 'scrolloff' does when
 using CTRL-E. (Yee Cheng Chin, #3721)
 
-Add function to make use of internal diff, working on two lists and returning
-unified diff (list of lines).
-
 When splitting a window with few text lines, the relative cursor position is
 kept, which means part of the text isn't displayed.  Better show all the text
 when possible. (Dylan Lloyd, #3973)
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -1368,6 +1368,7 @@ Various:					*various-functions*
 	changenr()		return number of most recent change
 	cscope_connection()	check if a cscope connection exists
 	did_filetype()		check if a FileType autocommand was used
+	diff()			diff two Lists of strings
 	eventhandler()		check if invoked by an event handler
 	getpid()		get process ID of Vim
 	getscriptinfo()		get list of sourced vim scripts
--- a/src/diff.c
+++ b/src/diff.c
@@ -42,6 +42,10 @@ static int	diff_flags = DIFF_INTERNAL | DIFF_FILLER | DIFF_CLOSE_OFF;
 
 static long diff_algorithm = 0;
 
+#define DIFF_INTERNAL_OUTPUT_UNIFIED	1
+#define DIFF_INTERNAL_OUTPUT_INDICES	2
+static int diff_internal_output_fmt = DIFF_INTERNAL_OUTPUT_INDICES;
+
 #define LBUFLEN 50		// length of line in diff file
 
 static int diff_a_works = MAYBE; // TRUE when "diff -a" works, FALSE when it
@@ -97,7 +101,8 @@ static void diff_copy_entry(diff_T *dprev, diff_T *dp, int idx_orig, int idx_new
 static diff_T *diff_alloc_new(tabpage_T *tp, diff_T *dprev, diff_T *dp);
 static int parse_diff_ed(char_u *line, diffhunk_T *hunk);
 static int parse_diff_unified(char_u *line, diffhunk_T *hunk);
-static int xdiff_out(long start_a, long count_a, long start_b, long count_b, void *priv);
+static int xdiff_out_indices(long start_a, long count_a, long start_b, long count_b, void *priv);
+static int xdiff_out_unified(void *priv, mmbuffer_t *mb, int nbuf);
 
 #define FOR_ALL_DIFFBLOCKS_IN_TAB(tp, dp) \
     for ((dp) = (tp)->tp_first_diff; (dp) != NULL; (dp) = (dp)->df_next)
@@ -1142,7 +1147,10 @@ diff_file_internal(diffio_T *diffio)
 
     emit_cfg.ctxlen = 0; // don't need any diff_context here
     emit_cb.priv = &diffio->dio_diff;
-    emit_cfg.hunk_func = xdiff_out;
+    if (diff_internal_output_fmt == DIFF_INTERNAL_OUTPUT_INDICES)
+	emit_cfg.hunk_func = xdiff_out_indices;
+    else
+	emit_cb.out_line = xdiff_out_unified;
     if (xdl_diff(&diffio->dio_orig.din_mmfile,
 		&diffio->dio_new.din_mmfile,
 		&param, &emit_cfg, &emit_cb) < 0)
@@ -3327,10 +3335,10 @@ parse_diff_unified(
 
 /*
  * Callback function for the xdl_diff() function.
- * Stores the diff output in a grow array.
+ * Stores the diff output (indices) in a grow array.
  */
     static int
-xdiff_out(
+xdiff_out_indices(
 	long start_a,
 	long count_a,
 	long start_b,
@@ -3354,6 +3362,25 @@ xdiff_out(
     p->lnum_new   = start_b + 1;
     p->count_new  = count_b;
     ((diffhunk_T **)dout->dout_ga.ga_data)[dout->dout_ga.ga_len++] = p;
+    return 0;
+}
+
+/*
+ * Callback function for the xdl_diff() function.
+ * Stores the unified diff output in a grow array.
+ */
+    static int
+xdiff_out_unified(
+	void *priv,
+	mmbuffer_t *mb,
+	int nbuf)
+{
+    diffout_T	*dout = (diffout_T *)priv;
+    int		i;
+
+    for (i = 0; i < nbuf; i++)
+	ga_concat_len(&dout->dout_ga, (char_u *)mb[i].ptr, mb[i].size);
+
     return 0;
 }
 
@@ -3436,6 +3463,207 @@ f_diff_hlID(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
 	    hlID = HLF_CHD;			// changed line
     }
     rettv->vval.v_number = hlID == (hlf_T)0 ? 0 : (int)hlID;
+#endif
+}
+
+/*
+ * Parse the diff options passed in "optarg" to the diff() function and return
+ * the options in "diffopts" and the diff algorithm in "diffalgo".
+ */
+    static int
+parse_diff_optarg(
+    typval_T	*opts,
+    int		*diffopts,
+    long	*diffalgo,
+    int		*diff_output_fmt)
+{
+    dict_T *d = opts->vval.v_dict;
+
+    char_u  *algo = dict_get_string(d, "algorithm", FALSE);
+    if (algo != NULL)
+    {
+	if (STRNCMP(algo, "myers", 5) == 0)
+	    *diffalgo = 0;
+	else if (STRNCMP(algo, "minimal", 7) == 0)
+	    *diffalgo = XDF_NEED_MINIMAL;
+	else if (STRNCMP(algo, "patience", 8) == 0)
+	    *diffalgo = XDF_PATIENCE_DIFF;
+	else if (STRNCMP(algo, "histogram", 9) == 0)
+	    *diffalgo = XDF_HISTOGRAM_DIFF;
+    }
+
+    char_u  *output_fmt = dict_get_string(d, "output", FALSE);
+    if (output_fmt != NULL)
+    {
+	if (STRNCMP(output_fmt, "unified", 7) == 0)
+	    *diff_output_fmt = DIFF_INTERNAL_OUTPUT_UNIFIED;
+	else if (STRNCMP(output_fmt, "indices", 7) == 0)
+	    *diff_output_fmt = DIFF_INTERNAL_OUTPUT_INDICES;
+	else
+	{
+	    semsg(_(e_unsupported_diff_output_format_str), output_fmt);
+	    return FAIL;
+	}
+    }
+
+    if (dict_get_bool(d, "iblank", FALSE))
+	*diffopts |= DIFF_IBLANK;
+    if (dict_get_bool(d, "icase", FALSE))
+	*diffopts |= DIFF_ICASE;
+    if (dict_get_bool(d, "iwhite", FALSE))
+	*diffopts |= DIFF_IWHITE;
+    if (dict_get_bool(d, "iwhiteall", FALSE))
+	*diffopts |= DIFF_IWHITEALL;
+    if (dict_get_bool(d, "iwhiteeol", FALSE))
+	*diffopts |= DIFF_IWHITEEOL;
+    if (dict_get_bool(d, "indent-heuristic", FALSE))
+	*diffalgo |= XDF_INDENT_HEURISTIC;
+
+    return OK;
+}
+
+/*
+ * Concatenate the List of strings in "l" and store the result in
+ * "din->din_mmfile.ptr" and the length in "din->din_mmfile.size".
+ */
+    static void
+list_to_diffin(list_T *l, diffin_T *din, int icase)
+{
+    garray_T	ga;
+    listitem_T	*li;
+    char_u	*str;
+
+    ga_init2(&ga, 512, 4);
+
+    FOR_ALL_LIST_ITEMS(l, li)
+    {
+	str = tv_get_string(&li->li_tv);
+	if (icase)
+	{
+	    str = strlow_save(str);
+	    if (str == NULL)
+		continue;
+	}
+	ga_concat(&ga, str);
+	ga_concat(&ga, (char_u *)NL_STR);
+	if (icase)
+	    vim_free(str);
+    }
+    if (ga.ga_len > 0)
+	((char *)ga.ga_data)[ga.ga_len] = NUL;
+
+    din->din_mmfile.ptr = (char *)ga.ga_data;
+    din->din_mmfile.size = ga.ga_len;
+}
+
+/*
+ * Get the start and end indices from the diff "hunk".
+ */
+    static dict_T *
+get_diff_hunk_indices(diffhunk_T *hunk)
+{
+    dict_T	*hunk_dict;
+
+    hunk_dict = dict_alloc();
+    if (hunk_dict == NULL)
+	return NULL;
+
+    dict_add_number(hunk_dict, "from_idx", hunk->lnum_orig - 1);
+    dict_add_number(hunk_dict, "from_count", hunk->count_orig);
+    dict_add_number(hunk_dict, "to_idx", hunk->lnum_new  - 1);
+    dict_add_number(hunk_dict, "to_count", hunk->count_new);
+
+    return hunk_dict;
+}
+
+/*
+ * "diff()" function
+ */
+    void
+f_diff(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
+{
+#ifdef FEAT_DIFF
+    diffio_T dio;
+
+    if (check_for_nonnull_list_arg(argvars, 0) == FAIL
+	    || check_for_nonnull_list_arg(argvars, 1) == FAIL
+	    || check_for_opt_nonnull_dict_arg(argvars, 2) == FAIL)
+	return;
+
+    CLEAR_FIELD(dio);
+    dio.dio_internal = TRUE;
+    ga_init2(&dio.dio_diff.dout_ga, sizeof(char *), 1000);
+
+    list_T *orig_list = argvars[0].vval.v_list;
+    list_T *new_list = argvars[1].vval.v_list;
+
+    // Save the 'diffopt' option value and restore it after getting the diff.
+    int		save_diff_flags = diff_flags;
+    long	save_diff_algorithm = diff_algorithm;
+    long	save_diff_output_fmt = diff_internal_output_fmt;
+    diff_flags = DIFF_INTERNAL;
+    diff_algorithm = 0;
+    diff_internal_output_fmt = DIFF_INTERNAL_OUTPUT_UNIFIED;
+    if (argvars[2].v_type != VAR_UNKNOWN)
+	if (parse_diff_optarg(&argvars[2], &diff_flags, &diff_algorithm,
+					&diff_internal_output_fmt) == FAIL)
+	{
+	    diff_internal_output_fmt = save_diff_output_fmt;
+	    return;
+	}
+
+    // Concatenate the List of strings into a single string using newline
+    // separator.  Internal diff library expects a single string.
+    list_to_diffin(orig_list, &dio.dio_orig, diff_flags & DIFF_ICASE);
+    list_to_diffin(new_list, &dio.dio_new, diff_flags & DIFF_ICASE);
+
+    // Compute the diff
+    int diff_status = diff_file(&dio);
+
+    if (diff_status == FAIL)
+	goto done;
+
+    int		hunk_idx = 0;
+    dict_T	*hunk_dict;
+
+    if (diff_internal_output_fmt == DIFF_INTERNAL_OUTPUT_INDICES)
+    {
+	if (rettv_list_alloc(rettv) != OK)
+	    goto done;
+	list_T	*l = rettv->vval.v_list;
+
+	// Process each diff hunk
+	diffhunk_T	*hunk = NULL;
+	while (hunk_idx < dio.dio_diff.dout_ga.ga_len)
+	{
+	    hunk = ((diffhunk_T **)dio.dio_diff.dout_ga.ga_data)[hunk_idx++];
+
+	    hunk_dict = get_diff_hunk_indices(hunk);
+	    if (hunk_dict == NULL)
+		goto done;
+
+	    list_append_dict(l, hunk_dict);
+	}
+    }
+    else
+    {
+	ga_append(&dio.dio_diff.dout_ga, NUL);
+	rettv->v_type = VAR_STRING;
+	rettv->vval.v_string =
+			vim_strsave((char_u *)dio.dio_diff.dout_ga.ga_data);
+    }
+
+done:
+    clear_diffin(&dio.dio_new);
+    if (diff_internal_output_fmt == DIFF_INTERNAL_OUTPUT_INDICES)
+	clear_diffout(&dio.dio_diff);
+    else
+	ga_clear(&dio.dio_diff.dout_ga);
+    clear_diffin(&dio.dio_orig);
+    // Restore the 'diffopt' option value.
+    diff_flags = save_diff_flags;
+    diff_algorithm = save_diff_algorithm;
+    diff_internal_output_fmt = save_diff_output_fmt;
 #endif
 }
 
--- a/src/errors.h
+++ b/src/errors.h
@@ -258,8 +258,9 @@ EXTERN char e_escape_not_allowed_in_digraph[]
 EXTERN char e_using_loadkeymap_not_in_sourced_file[]
 	INIT(= N_("E105: Using :loadkeymap not in a sourced file"));
 #endif
-// E106 unused
 #ifdef FEAT_EVAL
+EXTERN char e_unsupported_diff_output_format_str[]
+	INIT(= N_("E106: Unsupported diff output format: %s"));
 EXTERN char e_missing_parenthesis_str[]
 	INIT(= N_("E107: Missing parentheses: %s"));
 EXTERN char e_no_such_variable_str[]
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -1148,6 +1148,7 @@ static argcheck_T arg3_buffer_number_number[] = {arg_buffer, arg_number, arg_num
 static argcheck_T arg3_buffer_string_any[] = {arg_buffer, arg_string, arg_any};
 static argcheck_T arg3_buffer_string_dict[] = {arg_buffer, arg_string, arg_dict_any};
 static argcheck_T arg3_dict_number_number[] = {arg_dict_any, arg_number, arg_number};
+static argcheck_T arg3_diff[] = {arg_list_string, arg_list_string, arg_dict_any};
 static argcheck_T arg3_list_string_dict[] = {arg_list_any, arg_string, arg_dict_any};
 static argcheck_T arg3_lnum_number_bool[] = {arg_lnum, arg_number, arg_bool};
 static argcheck_T arg3_number[] = {arg_number, arg_number, arg_number};
@@ -1950,6 +1951,8 @@ static funcentry_T global_functions[] =
 			ret_number_bool,    f_deletebufline},
     {"did_filetype",	0, 0, 0,	    NULL,
 			ret_number_bool,    f_did_filetype},
+    {"diff",		2, 3, FEARG_1,	    arg3_diff,
+			ret_list_dict_any,  f_diff},
     {"diff_filler",	1, 1, FEARG_1,	    arg1_lnum,
 			ret_number,	    f_diff_filler},
     {"diff_hlID",	2, 2, FEARG_1,	    arg2_lnum_number,
--- a/src/proto/diff.pro
+++ b/src/proto/diff.pro
@@ -30,4 +30,5 @@ linenr_T diff_get_corresponding_line(buf_T *buf1, linenr_T lnum1);
 linenr_T diff_lnum_win(linenr_T lnum, win_T *wp);
 void f_diff_filler(typval_T *argvars, typval_T *rettv);
 void f_diff_hlID(typval_T *argvars, typval_T *rettv);
+void f_diff(typval_T *argvars, typval_T *rettv);
 /* vim: set ft=c : */
--- a/src/proto/typval.pro
+++ b/src/proto/typval.pro
@@ -26,6 +26,7 @@ int check_for_opt_list_arg(typval_T *args, int idx);
 int check_for_dict_arg(typval_T *args, int idx);
 int check_for_nonnull_dict_arg(typval_T *args, int idx);
 int check_for_opt_dict_arg(typval_T *args, int idx);
+int check_for_opt_nonnull_dict_arg(typval_T *args, int idx);
 int check_for_chan_or_job_arg(typval_T *args, int idx);
 int check_for_opt_chan_or_job_arg(typval_T *args, int idx);
 int check_for_job_arg(typval_T *args, int idx);
--- a/src/testdir/test_diffmode.vim
+++ b/src/testdir/test_diffmode.vim
@@ -1712,5 +1712,182 @@ func Test_diff_put_and_undo()
   set nodiff
 endfunc
 
+" Test for the diff() function
+def Test_diff_func()
+  # string is added/removed/modified at the beginning
+  assert_equal("@@ -0,0 +1 @@\n+abc\n",
+               diff(['def'], ['abc', 'def'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 0, to_idx: 0, to_count: 1}],
+               diff(['def'], ['abc', 'def'], {output: 'indices'}))
+  assert_equal("@@ -1 +0,0 @@\n-abc\n",
+               diff(['abc', 'def'], ['def'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 0}],
+               diff(['abc', 'def'], ['def'], {output: 'indices'}))
+  assert_equal("@@ -1 +1 @@\n-abc\n+abx\n",
+               diff(['abc', 'def'], ['abx', 'def'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abc', 'def'], ['abx', 'def'], {output: 'indices'}))
+
+  # string is added/removed/modified at the end
+  assert_equal("@@ -1,0 +2 @@\n+def\n",
+               diff(['abc'], ['abc', 'def'], {output: 'unified'}))
+  assert_equal([{from_idx: 1, from_count: 0, to_idx: 1, to_count: 1}],
+               diff(['abc'], ['abc', 'def'], {output: 'indices'}))
+  assert_equal("@@ -2 +1,0 @@\n-def\n",
+               diff(['abc', 'def'], ['abc'], {output: 'unified'}))
+  assert_equal([{from_idx: 1, from_count: 1, to_idx: 1, to_count: 0}],
+               diff(['abc', 'def'], ['abc'], {output: 'indices'}))
+  assert_equal("@@ -2 +2 @@\n-def\n+xef\n",
+               diff(['abc', 'def'], ['abc', 'xef'], {output: 'unified'}))
+  assert_equal([{from_idx: 1, from_count: 1, to_idx: 1, to_count: 1}],
+               diff(['abc', 'def'], ['abc', 'xef'], {output: 'indices'}))
+
+  # string is added/removed/modified in the middle
+  assert_equal("@@ -2,0 +3 @@\n+xxx\n",
+               diff(['111', '222', '333'], ['111', '222', 'xxx', '333'], {output: 'unified'}))
+  assert_equal([{from_idx: 2, from_count: 0, to_idx: 2, to_count: 1}],
+               diff(['111', '222', '333'], ['111', '222', 'xxx', '333'], {output: 'indices'}))
+  assert_equal("@@ -3 +2,0 @@\n-333\n",
+               diff(['111', '222', '333', '444'], ['111', '222', '444'], {output: 'unified'}))
+  assert_equal([{from_idx: 2, from_count: 1, to_idx: 2, to_count: 0}],
+               diff(['111', '222', '333', '444'], ['111', '222', '444'], {output: 'indices'}))
+  assert_equal("@@ -3 +3 @@\n-333\n+xxx\n",
+               diff(['111', '222', '333', '444'], ['111', '222', 'xxx', '444'], {output: 'unified'}))
+  assert_equal([{from_idx: 2, from_count: 1, to_idx: 2, to_count: 1}],
+               diff(['111', '222', '333', '444'], ['111', '222', 'xxx', '444'], {output: 'indices'}))
+
+  # new strings are added to an empty List
+  assert_equal("@@ -0,0 +1,2 @@\n+abc\n+def\n",
+               diff([], ['abc', 'def'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 0, to_idx: 0, to_count: 2}],
+               diff([], ['abc', 'def'], {output: 'indices'}))
+
+  # all the strings are removed from a List
+  assert_equal("@@ -1,2 +0,0 @@\n-abc\n-def\n",
+               diff(['abc', 'def'], [], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 2, to_idx: 0, to_count: 0}],
+               diff(['abc', 'def'], [], {output: 'indices'}))
+
+  # First character is added/removed/different
+  assert_equal("@@ -1 +1 @@\n-abc\n+bc\n",
+               diff(['abc'], ['bc'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abc'], ['bc'], {output: 'indices'}))
+  assert_equal("@@ -1 +1 @@\n-bc\n+abc\n",
+               diff(['bc'], ['abc'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['bc'], ['abc'], {output: 'indices'}))
+  assert_equal("@@ -1 +1 @@\n-abc\n+xbc\n",
+               diff(['abc'], ['xbc'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abc'], ['xbc'], {output: 'indices'}))
+
+  # Last character is added/removed/different
+  assert_equal("@@ -1 +1 @@\n-abc\n+abcd\n",
+               diff(['abc'], ['abcd'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abc'], ['abcd'], {output: 'indices'}))
+  assert_equal("@@ -1 +1 @@\n-abcd\n+abc\n",
+               diff(['abcd'], ['abc'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abcd'], ['abc'], {output: 'indices'}))
+  assert_equal("@@ -1 +1 @@\n-abc\n+abx\n",
+               diff(['abc'], ['abx'], {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1}],
+               diff(['abc'], ['abx'], {output: 'indices'}))
+
+  # partial string modification at the start and at the end.
+  var fromlist =<< trim END
+    one two
+    three four
+    five six
+  END
+  var tolist =<< trim END
+    one
+    six
+  END
+  assert_equal("@@ -1,3 +1,2 @@\n-one two\n-three four\n-five six\n+one\n+six\n", diff(fromlist, tolist, {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 3, to_idx: 0, to_count: 2}],
+               diff(fromlist, tolist, {output: 'indices'}))
+
+  # non-contiguous modifications
+  fromlist =<< trim END
+    one two
+    three four
+    five abc six
+  END
+  tolist =<< trim END
+    one abc two
+    three four
+    five six
+  END
+  assert_equal("@@ -1 +1 @@\n-one two\n+one abc two\n@@ -3 +3 @@\n-five abc six\n+five six\n",
+               diff(fromlist, tolist, {output: 'unified'}))
+  assert_equal([{from_idx: 0, from_count: 1, to_idx: 0, to_count: 1},
+                {from_idx: 2, from_count: 1, to_idx: 2, to_count: 1}],
+               diff(fromlist, tolist, {output: 'indices'}))
+
+  # add/remove blank lines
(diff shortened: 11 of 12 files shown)


brammool:
> Add support for computing diff between two Lists of strings using a
> function.
> 
> When using a LSP plugin, whenever a buffer is modified, the
> modifications are sent to the language server.
> To do this, the LSP plugins have implemented computing the diff in
> Vimscript or in Lua:
> 
> https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/utils/diff.vim
> https://github.com/natebosch/vim-lsc/blob/master/autoload/lsc/diff.vim
> https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/sync.lua
> 
> These computations are expensive as they compare every character and
> line in a buffer.
> This new function will help in efficiently computing the diff.

I suppose you made the structure of the return value work for your
purpose.  It's a bit strange though: List of Dict of Dicts.  And using a
special "end" item to indicate added or removed text (the help is not
clear, but I guess this is only used when a whole item was
added/removed, not when a word was added/removed).

To find inserted or removed list items requires checking whether the
"byte" item indicates the end of the string.

Wouldn't it be easier to use when returning something similar to unified
diff:
- Indicate range of deleted items (like a "-" line)
- Indicate range of inserted items (like a "+" line)
- Indicate a modified item (like a "-" line followed by a "+" line)

Could be done with a "count" item, which is negative for deleted items,
positive for inserted items and zero if items were modified.

Does this still fit in with what you need for LSP support?


Minor remarks:

+diff({list}, {list} [, {options}])
+				List	compute the diff of two List of strings

"List" -> "Lists".
"of" sounds a bit wrong, using "between" makes the text wrap.  How about
"diff two Lists of strings"?

-- 
It is too bad that the speed of light hasn't kept pace with the
changes in CPU speed and network bandwidth. -- ***@***.***>

vim-ml:
Hi Bram,

On Sun, Apr 30, 2023 at 1:26 PM Bram Moolenaar ***@***.***>
wrote:

>
> > Add support for computing diff between two Lists of strings using a
> > function.
> >
> > When using a LSP plugin, whenever a buffer is modified, the
> > modifications are sent to the language server.
> > To do this, the LSP plugins have implemented computing the diff in
> > Vimscript or in Lua:
> >
> >
> https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/utils/diff.vim
> > https://github.com/natebosch/vim-lsc/blob/master/autoload/lsc/diff.vim
> >
> https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/sync.lua
> >
> > These computations are expensive as they compare every character and
> > line in a buffer.
> > This new function will help in efficiently computing the diff.
>
> I suppose you made the structure of the return value work for your
> purpose. It's a bit strange though: List of Dict of Dicts. And using a
> special "end" item to indicate added or removed text (the help is not
> clear, but I guess this is only used when a whole item was
> added/removed, not when a word was added/removed).
>

The "end" item contains the end index and the byte offset of the diff hunk.
If the string was not present earlier, then the byte offset cannot be
calculated.  So it will be -1.   When a word is removed, it contains the
position of the end of the word.


>
> To find inserted or removed list items requires checking whether the
> "byte" item indicates the end of the string.
>
> Wouldn't it be easier to use when returning something similar to unified
> diff:
>
- Indicate range of deleted items (like a "-" line)
> - Indicate range of inserted items (like a "+" line)
> - Indicate a modified item (like a "-" line followed by a "+" line)
>

The start and end index values in the returned Dict does contain the
above information (range of deleted or inserted or modified items).


>
> Could be done with a "count" item, which is negative for deleted items,
> positive for inserted items and zero if items were modified.
>
> Does this still fit in with what you need for LSP support?
>
>
> Minor remarks:
>
> +diff({list}, {list} [, {options}])
> + List compute the diff of two List of strings
>
> "List" -> "Lists".
> "of" sounds a bit wrong, using "between" makes the text wrap. How about
> "diff two Lists of strings"?
>
>
I have updated the help text.

Regards,
Yegappan

vim-ml:
Hi Bram,

On Sun, Apr 30, 2023 at 2:01 PM vim-dev ML ***@***.***>
wrote:

> On Sun, Apr 30, 2023 at 1:26 PM Bram Moolenaar ***@***.***>
> wrote:
>
> >
> > > Add support for computing diff between two Lists of strings using a
> > > function.
> > >
> > > When using a LSP plugin, whenever a buffer is modified, the
> > > modifications are sent to the language server.
> > > To do this, the LSP plugins have implemented computing the diff in
> > > Vimscript or in Lua:
> > >
> > >
> >
> https://github.com/prabirshrestha/vim-lsp/blob/master/autoload/lsp/utils/diff.vim
> > > https://github.com/natebosch/vim-lsc/blob/master/autoload/lsc/diff.vim
> > >
> >
> https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/sync.lua
> > >
> > > These computations are expensive as they compare every character and
> > > line in a buffer.
> > > This new function will help in efficiently computing the diff.
> >
> > I suppose you made the structure of the return value work for your
> > purpose. It's a bit strange though: List of Dict of Dicts. And using a
> > special "end" item to indicate added or removed text (the help is not
> > clear, but I guess this is only used when a whole item was
> > added/removed, not when a word was added/removed).
> >
>
> The "end" item contains the end index and the byte offset of the diff hunk.
> If the string was not present earlier, then the byte offset cannot be
> calculated. So it will be -1. When a word is removed, it contains the
> position of the end of the word.
>
> >
> > To find inserted or removed list items requires checking whether the
> > "byte" item indicates the end of the string.
> >
> > Wouldn't it be easier to use when returning something similar to unified
> > diff:
> >
> - Indicate range of deleted items (like a "-" line)
> > - Indicate range of inserted items (like a "+" line)
> > - Indicate a modified item (like a "-" line followed by a "+" line)
> >
>
> The start and end index values in the returned Dict does contain the
> above information (range of deleted or inserted or modified items).
>
>
I am attaching some examples to describe the return value of this function.

Example 1:
---------- old text -----------
aaaaaa
aaaaaa
aaaaaa
--------------------------------

---------- new text -----------
aaaxx
xxxxxx
xxaaa
----------------------------------

The modification starts at index 0 and byte 3 and ends at index 2 byte 2 in
the
old text.  In the new text, the changes start at index 0 byte 3 and end at
index
2 byte 1.  In this case, the diff() function returns the following:

[{'from': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 2}},
  'to': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 1}}}]

Example 2:
---------- old text -----------
aaaaaa
bbbbbb
cccccc
--------------------------------

---------- new text -----------
aaaaaa
----------------------------------

In the old text, the modification starts at index 1 byte 0 and ends at
index 2 byte 5.
In the new text, the modification starts at index 1.  But the byte cannot
be calculated
as the line is removed.  Similarly the end of the modification also cannot
be calculated.
In this case, the diff() function returns the following:

[{'from': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}},
  'to': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}}}]

Example 3:
---------- old text -----------
aaaaaa
--------------------------------

---------- new text -----------
aaaaaa
bbbbbb
cccccc
----------------------------------

In the old text, the modification starts at index 1.  But the starting byte
and ending index/byte
cannot be calculated.  In the new text, the change starts at index 1 byte 0
and ends at
index 2 byte 5.  In this case, the diff() function returns the following:

[{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}},
  'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}}}]

Hopefully the above examples clarify your questions.

Regards,
Yegappan

vim-ml:
> I am attaching some examples to describe the return value of this function.
> 
> Example 1:
> ---------- old text -----------
> aaaaaa
> aaaaaa
> aaaaaa
> --------------------------------
> 
> ---------- new text -----------
> aaaxx
> xxxxxx
> xxaaa
> ----------------------------------
> 
> The modification starts at index 0 and byte 3 and ends at index 2 byte 2 in
> the
> old text.  In the new text, the changes start at index 0 byte 3 and end at
> index
> 2 byte 1.  In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 2}},
>   'to': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 1}}}]

So, how does one decide whether items were inserted or removed?
I think that's by computing the size in items of "from" and doing the
same for "to" and then they turn out to be equal.  If someone cares
about items being inserted/deleted this is a bit of an indirect way to
find out.


> Example 2:
> ---------- old text -----------
> aaaaaa
> bbbbbb
> cccccc
> --------------------------------
> 
> ---------- new text -----------
> aaaaaa
> ----------------------------------
> 
> In the old text, the modification starts at index 1 byte 0 and ends at
> index 2 byte 5.
> In the new text, the modification starts at index 1.  But the byte cannot
> be calculated
> as the line is removed.  Similarly the end of the modification also cannot
> be calculated.
> In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}},
>   'to': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}}}]

Now how does one compute the number of items inserted/removed?  Is it
true that when "to" has an "end" with index -1, that there are no items
at all?  Not sure that is always true.


> Example 3:
> ---------- old text -----------
> aaaaaa
> --------------------------------
> 
> ---------- new text -----------
> aaaaaa
> bbbbbb
> cccccc
> ----------------------------------
> 
> In the old text, the modification starts at index 1.  But the starting byte
> and ending index/byte
> cannot be calculated.  In the new text, the change starts at index 1 byte 0
> and ends at
> index 2 byte 5.  In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}},
>   'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}}}]

So the "from" position indicates "end of the list".  And the "to" info
has two items, thus two items have been added.

I hope you see that, depending on what the user of the function wants to
know, the returned value can be hard to understand.

-- 
BEDEVERE: And that, my lord, is how we know the Earth to be banana-shaped.

vim-ml:
> > I suppose you made the structure of the return value work for your
> > purpose. It's a bit strange though: List of Dict of Dicts. And using a
> > special "end" item to indicate added or removed text (the help is not
> > clear, but I guess this is only used when a whole item was
> > added/removed, not when a word was added/removed).
> 
> The "end" item contains the end index and the byte offset of the diff hunk.
> If the string was not present earlier, then the byte offset cannot be
> calculated.  So it will be -1.   When a word is removed, it contains the
> position of the end of the word.

Before or after removing the word?  That's probably the main confusion
I'm seeing, that one has to translate the provided information into what
it really means.

> > Wouldn't it be easier to use when returning something similar to unified
> > diff:
> >
> > - Indicate range of deleted items (like a "-" line)
> > - Indicate range of inserted items (like a "+" line)
> > - Indicate a modified item (like a "-" line followed by a "+" line)
> 
> The start and end index values in the returned Dict does contain the
> above information (range of deleted or inserted or modified items).

Well, in some form.  My point is that it would be easier to interpret
the returned value if you put it in the form I suggested.

Note that some information goes missing when changing N lines/items to M
lines/items: some lines/items might only have a tiny change (e.g.
indent) while other lines/items may have been completely removed or
inserted.  There was someone who looked into this (mostly for how to
display a diff hunk) but I haven't heard about it for a long time.
In case this work gets picked up, it would be nice if it can be included
in the returned value.

-- 
If you had to identify, in one word, the reason why the
human race has not achieved, and never will achieve, its
full potential, that word would be "meetings."

vim-ml:
Hi Bram,

On Mon, May 1, 2023 at 10:22 AM vim-dev ML ***@***.***>
wrote:

>
> > I am attaching some examples to describe the return value of this
> function.
> >
> > Example 1:
> > ---------- old text -----------
> > aaaaaa
> > aaaaaa
> > aaaaaa
> > --------------------------------
> >
> > ---------- new text -----------
> > aaaxx
> > xxxxxx
> > xxaaa
> > ----------------------------------
> >
> > The modification starts at index 0 and byte 3 and ends at index 2 byte 2
> in
> > the
> > old text. In the new text, the changes start at index 0 byte 3 and end at
> > index
> > 2 byte 1. In this case, the diff() function returns the following:
> >
> > [{'from': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 2}},
> > 'to': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 1}}}]
>
> So, how does one decide whether items were inserted or removed?
> I think that's by computing the size in items of "from" and doing the
> same for "to" and then they turn out to be equal. If someone cares
> about items being inserted/deleted this is a bit of an indirect way to
> find out.
>

I have updated the PR to include the "count" item to "from" and "to".
It indicates the number of items added or modified.

The way to interpret the "from" and "to" items is:
The text beginning at the "start" item and ending at the "end" item in
"from" is replaced by the text beginning at the "start" item and ending
at the "end" item in "to" in a diff hunk.


>
>
> > Example 2:
> > ---------- old text -----------
> > aaaaaa
> > bbbbbb
> > cccccc
> > --------------------------------
> >
> > ---------- new text -----------
> > aaaaaa
> > ----------------------------------
> >
> > In the old text, the modification starts at index 1 byte 0 and ends at
> > index 2 byte 5.
> > In the new text, the modification starts at index 1. But the byte cannot
> > be calculated
> > as the line is removed. Similarly the end of the modification also cannot
> > be calculated.
> > In this case, the diff() function returns the following:
> >
> > [{'from': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}},
> > 'to': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}}}]
>
> Now how does one compute the number of items inserted/removed? Is it
> true that when "to" has an "end" with index -1, that there are no items
> at all? Not sure that is always true.
>
>
> > Example 3:
> > ---------- old text -----------
> > aaaaaa
> > --------------------------------
> >
> > ---------- new text -----------
> > aaaaaa
> > bbbbbb
> > cccccc
> > ----------------------------------
> >
> > In the old text, the modification starts at index 1. But the starting
> byte
> > and ending index/byte
> > cannot be calculated. In the new text, the change starts at index 1 byte
> 0
> > and ends at
> > index 2 byte 5. In this case, the diff() function returns the following:
> >
> > [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte':
> -1}},
> > 'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}}}]
>
> So the "from" position indicates "end of the list". And the "to" info
> has two items, thus two items have been added.
>
> I hope you see that, depending on what the user of the function wants to
> know, the returned value can be hard to understand.
>
>
>
Hopefully the new "count" item will help with this.

Regards,
Yegappan

chrisbra:
there is also a related issue #4241 
I am not sure the current design helps with that however. I find the result quite hard to understand.

yegappan:
@chrisbra: The result contains the starting and ending position for the range of added/removed/modified lines in the original list and the new list.  This will allow a plugin to get the new range of text that replaces the original range of text.  Note that I have included both the line offset and the starting and ending byte offset for the change.

chrisbra:
I am still confused by the output, even after reading the help section several times. So `index` is the index into the `list` input and byte is the start byte from that `list[index]`? And `count` is the number of changes performed? That is a bit confusing, because if an item was added to the list, like here:
```
" few lines added at the end
  :echo diff(['abc'], ['abc', 'def', 'ghi'])
   [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1},
	      'count': 0},
     'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 2},
	    'count': 2}}]
```
the `count` in the `to` dict means two additional items have been added (appended) to the list. But then if you take this example:
``` 
  " word is removed in the middle of a string
  :echo diff(['abc def ghi'], ['abc ghi'])
   [{'from': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': 7},
	      'count': 1},
     'to': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': -1},
	    'count': 1}}]
```
Then count does not mean to add or remove a single item to the list, but apparently that only one single change has been done, e.g. removing from byte 4 to byte 7.

Also, I guess the help could be a bit more precise, I suppose the `-1` applies only for added items in the `from` Dict and `removed` items in the `to` Dict
```
    Each item in the returned List is a Dict containing
    information about a diff hunk.  Each Dict contains the
    following items:
        from  Dict with {list1} diff hunk information
        to    Dict with {list2} diff hunk information
    The "from" and "to" Dicts contain the following items:
        start  Dict containing the starting index and byte of the
         diff hunk.  The "byte" is -1 if text is added 
         (in the "from" Dict) or removed (in the "to" Dict).
        end    Dict containing the ending index and byte of the
         diff hunk.  The "idx" and "byte" are -1 if text is
         added (in the "from" Dict) or removed (in the "to" Dict).
        count  number of items added/removed/modified in this diff
         hunk.
```

And how would this look like if we have several hunks? Can we have this as example as well?

Also I would have thought you would need something like `oldstart,oldcount newstart,newcount` which basically is the hunk header from a unified diff. So perhaps we can just also add the raw `diff output per hunk` as well?

brammool:
> I have updated the PR to include the "count" item to "from" and "to".
> It indicates the number of items added or modified.

The help for this is:

		    count  number of items added/removed/modified in this diff
			   hunk.

I'm afraid this doesn't really help.  First of al, I would expect the
info not inside the "from" and "to" Dicts, but besides them.  And I
would expect a separate number for items added/deleted and items
modified.  Something like:

		Each item in the returned List is a Dict containing
		information about a diff hunk.  Each Dict contains the
		following items:
		    from      Dict with {list1} diff hunk information
		    to        Dict with {list2} diff hunk information
		    extra     Number of items added (positive) or
			      removed (negative)
		    modified  Number of items that were modified

> The way to interpret the "from" and "to" items is:
> The text beginning at the "start" item and ending at the "end" item in
> "from" is replaced by the text beginning at the "start" item and ending
> at the "end" item in "to" in a diff hunk.

This needs the addition that the "to" item refers to the index and byte
in {list2}.  This is important, since previous hunks may have added and
removed items.

This part of the help is confusing and probably not right:

			   The "byte" is -1 if text is added or removed.

In the "from" dict this doesn't make sense.  If text is added (and
nothing modified) then "start" and "end should be equal, the position
where the text is inserted. If text is removed then "start" gives the
start of the removed text and "end" the end.  The text exists, thus
there is no reason to use -1.

It's important to get this right.  We had similar issues to solve with
the listener callback function.  See the help for listner_add().  It
might actually be helpful if the information is in the same form.

-- 
The acknowledged parents of reengineering are Michael Hammer and James Champy.
When I say they're the "parents" I don't mean they had sex - and I apologize
for making you think about it.  I mean they wrote the best-selling business
book _Reengineering the Corporation_, which was published in 1993.
   Businesses flocked to reengineering like frat boys to a drunken
cheerleader.  (This analogy wasn't necessary, but I'm trying to get my mind
off that Hammer and Champy thing.)

vim-ml:
[resend, picky postmaster refused the message]

> I am attaching some examples to describe the return value of this function.
> 
> Example 1:
> ---------- old text -----------
> aaaaaa
> aaaaaa
> aaaaaa
> --------------------------------
> 
> ---------- new text -----------
> aaaxx
> xxxxxx
> xxaaa
> ----------------------------------
> 
> The modification starts at index 0 and byte 3 and ends at index 2 byte 2 in
> the
> old text.  In the new text, the changes start at index 0 byte 3 and end at
> index
> 2 byte 1.  In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 2}},
>   'to': {'start': {'idx': 0, 'byte': 3}, 'end': {'idx': 2, 'byte': 1}}}]

So, how does one decide whether items were inserted or removed?
I think that's by computing the size in items of "from" and doing the
same for "to" and then they turn out to be equal.  If someone cares
about items being inserted/deleted this is a bit of an indirect way to
find out.


> Example 2:
> ---------- old text -----------
> aaaaaa
> bbbbbb
> cccccc
> --------------------------------
> 
> ---------- new text -----------
> aaaaaa
> ----------------------------------
> 
> In the old text, the modification starts at index 1 byte 0 and ends at
> index 2 byte 5.
> In the new text, the modification starts at index 1.  But the byte cannot
> be calculated
> as the line is removed.  Similarly the end of the modification also cannot
> be calculated.
> In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}},
>   'to': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}}}]

Now how does one compute the number of items inserted/removed?  Is it
true that when "to" has an "end" with index -1, that there are no items
at all?  Not sure that is always true.


> Example 3:
> ---------- old text -----------
> aaaaaa
> --------------------------------
> 
> ---------- new text -----------
> aaaaaa
> bbbbbb
> cccccc
> ----------------------------------
> 
> In the old text, the modification starts at index 1.  But the starting byte
> and ending index/byte
> cannot be calculated.  In the new text, the change starts at index 1 byte 0
> and ends at
> index 2 byte 5.  In this case, the diff() function returns the following:
> 
> [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1}},
>   'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 5}}}]

So the "from" position indicates "end of the list".  And the "to" info
has two items, thus two items have been added.

I hope you see that, depending on what the user of the function wants to
know, the returned value can be hard to understand.

-- 
BEDEVERE: And that, my lord, is how we know the Earth to be banana-shaped.

vim-ml:
[resend, picky postmaster refused the message]

> > I suppose you made the structure of the return value work for your
> > purpose. It's a bit strange though: List of Dict of Dicts. And using a
> > special "end" item to indicate added or removed text (the help is not
> > clear, but I guess this is only used when a whole item was
> > added/removed, not when a word was added/removed).
> 
> The "end" item contains the end index and the byte offset of the diff hunk.
> If the string was not present earlier, then the byte offset cannot be
> calculated.  So it will be -1.   When a word is removed, it contains the
> position of the end of the word.

Before or after removing the word?  That's probably the main confusion
I'm seeing, that one has to translate the provided information into what
it really means.

> > Wouldn't it be easier to use when returning something similar to unified
> > diff:
> >
> > - Indicate range of deleted items (like a "-" line)
> > - Indicate range of inserted items (like a "+" line)
> > - Indicate a modified item (like a "-" line followed by a "+" line)
> 
> The start and end index values in the returned Dict does contain the
> above information (range of deleted or inserted or modified items).

Well, in some form.  My point is that it would be easier to interpret
the returned value if you put it in the form I suggested.

Note that some information goes missing when changing N lines/items to M
lines/items: some lines/items might only have a tiny change (e.g.
indent) while other lines/items may have been completely removed or
inserted.  There was someone who looked into this (mostly for how to
display a diff hunk) but I haven't heard about it for a long time.
In case this work gets picked up, it would be nice if it can be included
in the returned value.

-- 
If you had to identify, in one word, the reason why the
human race has not achieved, and never will achieve, its
full potential, that word would be "meetings."

vim-ml:
Hi Christian,

On Tue, May 2, 2023 at 9:43 AM Christian Brabandt ***@***.***>
wrote:

> I am still confused by the output, even after reading the help section
> several times. So index is the index into the list input and byte is the
> start byte from that list[index]?
>

Yes.


> And count is the number of changes performed?
>

Yes.  "count" is the number of items in the list added or removed or
modified in the diff hunk.


> That is a bit confusing, because if an item was added to the list, like
> here:
>
> " few lines added at the end
>   :echo diff(['abc'], ['abc', 'def', 'ghi'])
>    [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1},
> 	      'count': 0},
>      'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 2},
> 	    'count': 2}}]
>
> the count in the to dict means two additional items have been added
> (appended) to the list. But then if you take this example:
>
>   " word is removed in the middle of a string
>   :echo diff(['abc def ghi'], ['abc ghi'])
>    [{'from': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': 7},
> 	      'count': 1},
>      'to': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': -1},
> 	    'count': 1}}]
>
> Then count does not mean to add or remove a single item to the list, but
> apparently that only one single change has been done, e.g. removing from
> byte 4 to byte 7.
>

In this example, there is only one item in the "from" and "to" lists.  As
this item is modified, the "count"
is 1 in both the "from" and "to" dicts.


> Also, I guess the help could be a bit more precise, I suppose the -1
> applies only for added items in the from Dict and removed items in the to
> Dict
>

I will update the help text to clarify this.


>     Each item in the returned List is a Dict containing
>     information about a diff hunk.  Each Dict contains the
>     following items:
>         from  Dict with {list1} diff hunk information
>         to    Dict with {list2} diff hunk information
>     The "from" and "to" Dicts contain the following items:
>         start  Dict containing the starting index and byte of the
>          diff hunk.  The "byte" is -1 if text is added
>          (in the "from" Dict) or removed (in the "to" Dict).
>         end    Dict containing the ending index and byte of the
>          diff hunk.  The "idx" and "byte" are -1 if text is
>          added (in the "from" Dict) or removed (in the "to" Dict).
>         count  number of items added/removed/modified in this diff
>          hunk.
>
> And how would this look like if we have several hunks? Can we have this as
> example as well?
>

Yes.  I will add additional examples for multiple diff hunks.


> Also I would have thought you would need something like oldstart,oldcount
> newstart,newcount which basically is the hunk header from a unified diff.
> So perhaps we can just also add the raw diff output per hunk as well?
>
>
>
The raw diff output from xdiff contains the line number and count for both
the original
and new lines.  Are you referring to including these in the output?

Regards,
Yegappan

chrisbra:
> The raw diff output from xdiff contains the line number and count for both
the original
and new lines.  Are you referring to including these in the output?

Yes the raw diff output. I think it is okay to include it, since you get it anyway from the `xdl` function.

vim-ml:
Hi Bram,

On Tue, May 2, 2023 at 12:54 PM Bram Moolenaar ***@***.***>
wrote:

>
> > I have updated the PR to include the "count" item to "from" and "to".
> > It indicates the number of items added or modified.
>
> The help for this is:
>
> count number of items added/removed/modified in this diff
> hunk.
>
> I'm afraid this doesn't really help. First of al, I would expect the
> info not inside the "from" and "to" Dicts, but besides them. And I
> would expect a separate number for items added/deleted and items
> modified. Something like:
>
> Each item in the returned List is a Dict containing
> information about a diff hunk. Each Dict contains the
> following items:
> from Dict with {list1} diff hunk information
> to Dict with {list2} diff hunk information
> extra Number of items added (positive) or
> removed (negative)
> modified Number of items that were modified
>

I have updated the PR to add the "added" and "modified" items for every
diff hunk.
The "added" item will be positive if new strings are added to the original
List
and will be negative if strings are removed from the original List.  The
"modified"
item indicates the number of strings modified from the original List.
I have also removed the "count" item.

Regards,
Yegappan


> > The way to interpret the "from" and "to" items is:
> > The text beginning at the "start" item and ending at the "end" item in
> > "from" is replaced by the text beginning at the "start" item and ending
> > at the "end" item in "to" in a diff hunk.
>
> This needs the addition that the "to" item refers to the index and byte
> in {list2}. This is important, since previous hunks may have added and
> removed items.
>
> This part of the help is confusing and probably not right:
>
> The "byte" is -1 if text is added or removed.
>
> In the "from" dict this doesn't make sense. If text is added (and
> nothing modified) then "start" and "end should be equal, the position
> where the text is inserted. If text is removed then "start" gives the
> start of the removed text and "end" the end. The text exists, thus
> there is no reason to use -1.
>
> It's important to get this right. We had similar issues to solve with
> the listener callback function. See the help for listner_add(). It
> might actually be helpful if the information is in the same form.
>
>
>

vim-ml:
Hi Christian,

On Tue, May 2, 2023 at 9:43 AM Christian Brabandt
***@***.***> wrote:
>
> I am still confused by the output, even after reading the help section several times. So index is the index into the list input and byte is the start byte from that list[index]? And count is the number of changes performed? That is a bit confusing, because if an item was added to the list, like here:
>
> " few lines added at the end
>   :echo diff(['abc'], ['abc', 'def', 'ghi'])
>    [{'from': {'start': {'idx': 1, 'byte': -1}, 'end': {'idx': -1, 'byte': -1},
>      'count': 0},
>      'to': {'start': {'idx': 1, 'byte': 0}, 'end': {'idx': 2, 'byte': 2},
>    'count': 2}}]
>
> the count in the to dict means two additional items have been added (appended) to the list. But then if you take this example:
>
>   " word is removed in the middle of a string
>   :echo diff(['abc def ghi'], ['abc ghi'])
>    [{'from': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': 7},
>      'count': 1},
>      'to': {'start': {'idx': 0, 'byte': 4}, 'end': {'idx': 0, 'byte': -1},
>    'count': 1}}]
>
> Then count does not mean to add or remove a single item to the list, but apparently that only one single change has been done, e.g. removing from byte 4 to byte 7.
>
> Also, I guess the help could be a bit more precise, I suppose the -1 applies only for added items in the from Dict and removed items in the to Dict
>
>     Each item in the returned List is a Dict containing
>     information about a diff hunk.  Each Dict contains the
>     following items:
>         from  Dict with {list1} diff hunk information
>         to    Dict with {list2} diff hunk information
>     The "from" and "to" Dicts contain the following items:
>         start  Dict containing the starting index and byte of the
>          diff hunk.  The "byte" is -1 if text is added
>          (in the "from" Dict) or removed (in the "to" Dict).
>         end    Dict containing the ending index and byte of the
>          diff hunk.  The "idx" and "byte" are -1 if text is
>          added (in the "from" Dict) or removed (in the "to" Dict).
>         count  number of items added/removed/modified in this diff
>          hunk.
>
> And how would this look like if we have several hunks? Can we have this as example as well?

I have updated the help with an example for multiple diff hunks.

Regards,
Yegappan

>
> Also I would have thought you would need something like oldstart,oldcount newstart,newcount which basically is the hunk header from a unified diff. So perhaps we can just also add the raw diff output per hunk as well?
>

vim-ml:
Hi Bram,

On Tue, May 2, 2023 at 12:54 PM Bram Moolenaar ***@***.***>
wrote:

>
> This part of the help is confusing and probably not right:
>
> The "byte" is -1 if text is added or removed.
>
> In the "from" dict this doesn't make sense. If text is added (and
> nothing modified) then "start" and "end should be equal, the position
> where the text is inserted. If text is removed then "start" gives the
> start of the removed text and "end" the end. The text exists, thus
> there is no reason to use -1.
>
>
>
If the "endbyte" is not -1, then removing a character cannot be
distinguished
from modifying a character as shown below:

Modifying the first character of a string:
  :echo diff(['abc'], ['xbc'])
   {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
    'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
    'added': 0, 'modified': 1}

Adding a first character to a string:
  :echo diff(['bc'], ['abc'])
   {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': -1},
    'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
    'added': 0, 'modified': 1}

In this case, if the "from.endbyte" is 0 instead of -1, then this case
cannot
be differentiated from the previous example.

Removing the first character from a string:
  :echo diff(['abc'], ['bc'])
   {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
    'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': -1},
    'added': 0, 'modified': 1}

In this case, if the "to.endbyte" is 0 instead of -1, then this case cannot
be differentiated from the first example.

There are some more examples where the "endbyte" or the "startbyte" is -1:

Removing a character from the middle of a string:
  :echo diff(['abc'], ['ac'])
   {'from': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': 1},
    'to': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': -1},
    'added': 0, 'modified': 1}

Adding a character to the middle of a string:
  :echo diff(['ac'], ['abc'])
   {'from': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': -1},
    'to': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': 1},
    'added': 0, 'modified': 1}

Removing the last character from a string:
  :echo diff(['abc'], ['ab'])
   {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
    'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': -1},
    'added': 0, 'modified': 1}

Adding a character to the end of a string:
  :echo diff(['ab'], ['abc'])
   {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': -1},
    'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
    'added': 0, 'modified': 1}

Modifying the last character in a string:
  :echo diff(['abc'], ['abx'])
   {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
    'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
    'added': 0, 'modified': 1}

Adding a new item:
  :echo diff(['a'], ['a', 'b'])
   {'from': {'startidx': 1, 'startbyte': -1, 'endidx': -1, 'endbyte': -1},
    'to': {'startidx': 1, 'startbyte': 0, 'endidx': 1, 'endbyte': 0},
    'added': 1, 'modified': 0}

Removing an item from the end:
  :echo diff(['a', 'b'], ['a'])
   {'from': {'startidx': 1, 'startbyte': 0, 'endidx': 1, 'endbyte': 0},
    'to': {'startidx': 1, 'startbyte': -1, 'endidx': -1, 'endbyte': -1},
    'added': -1, 'modified': 0}

Regards,
Yegappan

vim-ml:
> > This part of the help is confusing and probably not right:
> >
> > The "byte" is -1 if text is added or removed.
> >
> > In the "from" dict this doesn't make sense. If text is added (and
> > nothing modified) then "start" and "end should be equal, the position
> > where the text is inserted. If text is removed then "start" gives the
> > start of the removed text and "end" the end. The text exists, thus
> > there is no reason to use -1.
> >
> >
> >
> If the "endbyte" is not -1, then removing a character cannot be
> distinguished
> from modifying a character as shown below:
> 
> Modifying the first character of a string:
>   :echo diff(['abc'], ['xbc'])
>    {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
>     'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
>     'added': 0, 'modified': 1}
> 
> Adding a first character to a string:
>   :echo diff(['bc'], ['abc'])
>    {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': -1},
>     'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
>     'added': 0, 'modified': 1}
> 
> In this case, if the "from.endbyte" is 0 instead of -1, then this case
> cannot be differentiated from the previous example.

When adding a character then to "to" "endbyte" should be a positive
value, since the text is longer than it was before.  Thus the
from-endbyte should be zero, indicating that the "from" text was empty,
while the to-endbyte would be one (or as many bytes as the new character
occupies), indicating the text that was inserted.  This matches what
would be highlighted with DiffText.

> Removing the first character from a string:
>   :echo diff(['abc'], ['bc'])
>    {'from': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': 0},
>     'to': {'startidx': 0, 'startbyte': 0, 'endidx': 0, 'endbyte': -1},
>     'added': 0, 'modified': 1}
> 
> In this case, if the "to.endbyte" is 0 instead of -1, then this case cannot
> be differentiated from the first example.

The from-endbyte here should be positive, indicating the length of the
text that was removed.  The to-endbyte should be zero, indicating that
after the change the text is no longer there.

If no items were added or deleted, then "endbyte - startbyte" should
indicate the text affected.  If this is zero, then it means that text is
not present.  The other entry (from or to) then will be non-zero and
indicates the text removed or added.  If both are non-zero it indicates
text was changed.


> There are some more examples where the "endbyte" or the "startbyte" is -1:
> 
> Removing a character from the middle of a string:
>   :echo diff(['abc'], ['ac'])
>    {'from': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': 1},
>     'to': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': -1},
>     'added': 0, 'modified': 1}

Here from-endbyte should be 2 (the "b" was deleted) and to-endbyte
should be 1 (same as to-startbyte, empty text, indicating the "b" that
was removed).

> Adding a character to the middle of a string:
>   :echo diff(['ac'], ['abc'])
>    {'from': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': -1},
>     'to': {'startidx': 0, 'startbyte': 1, 'endidx': 0, 'endbyte': 1},
>     'added': 0, 'modified': 1}

Here from-endbyte should be 1 and to-endbyte should be 2.

> Removing the last character from a string:
>   :echo diff(['abc'], ['ab'])
>    {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
>     'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': -1},
>     'added': 0, 'modified': 1}

Here from-endbyte should be 3 and to-endbyte should be 2.

> Adding a character to the end of a string:
>   :echo diff(['ab'], ['abc'])
>    {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': -1},
>     'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
>     'added': 0, 'modified': 1}

Here from-endbyte should be 2 and to-endbyte should be 3.

> Modifying the last character in a string:
>   :echo diff(['abc'], ['abx'])
>    {'from': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
>     'to': {'startidx': 0, 'startbyte': 2, 'endidx': 0, 'endbyte': 2},
>     'added': 0, 'modified': 1}

Here both from-endbyte and to-endbyte should be 3.

> Adding a new item:
>   :echo diff(['a'], ['a', 'b'])
>    {'from': {'startidx': 1, 'startbyte': -1, 'endidx': -1, 'endbyte': -1},
>     'to': {'startidx': 1, 'startbyte': 0, 'endidx': 1, 'endbyte': 0},
>     'added': 1, 'modified': 0}

When adding an item the byte index doesn't have a meaning, we only need
to know the item index.  Here from-endidx should be 1, indicating the
index of where the new item was added.

Having the to-start and to-end be equal suggests that the text is empty.
Only adjusting "endbyte" to the length of the text makes it difficult to
distinguish between adding a whole item and inserting text in an
existing item.  Making the "end" exclusive would be simpler, then
to-endidx can be 2 and to-endbyte can remain zero.


> Removing an item from the end:
>   :echo diff(['a', 'b'], ['a'])
>    {'from': {'startidx': 1, 'startbyte': 0, 'endidx': 1, 'endbyte': 0},
>     'to': {'startidx': 1, 'startbyte': -1, 'endidx': -1, 'endbyte': -1},
>     'added': -1, 'modified': 0}

Also applying the exclusive index here means from-endidx should be 2.
That avoids having from-start and from-end having the same values, thus
it looks like nothing is deleted.  to-endidx should be the same as
to-startidx, indicating there is nothing left.  The byte indexes in "to"
should also be zero, so it is clear that no text is left.

Hopefully this works out and I actually expect the computations to be
simpler and consistent.


-- 
Q:  What's a light-year?
A:  One-third less calories than a regular year.

rickhowe:
I am interested in this builtin function as I have developed some plugins to support word/character level diff in lines ([this](https://github.com/rickhowe/diffchar.vim)) and range/area selectable diff in buffers ([this](https://github.com/rickhowe/spotdiff.vim)). I implemented the O(NP) comparison algorithm in vim script, but I have been looking for a builtin function like yours. Several month ago, I noticed that neovim has vim.diff() function so I switched to use it for neovim in my script.

Honestly, I am confused about a return value of your function. vim.diff() actually returns diff in a unified format as default, from which is very easy to generate a shortest edit script (SES).

For example, how easy can I generate this SES from your return value?

```
list1 = ['a', 'b', 'c', 'a', 'b', 'b', 'a']
list2 = ['c', 'b', 'a', 'b', 'a', 'c']
SES = '--=-=+==+'
```

vim-ml:
Hi,

On Sat, May 6, 2023 at 4:44 AM rickhowe ***@***.***> wrote:

> I am interested in this builtin function as I have developed some plugins
> to support word/character level diff in lines (this
> <https://github.com/rickhowe/diffchar.vim>) and range/area selectable
> diff in buffers (this <https://github.com/rickhowe/spotdiff.vim>). I
> implemented the O(NP) comparison algorithm in vim script, but I have been
> looking for a builtin function like yours. Several month ago, I noticed
> that neovim has vim.diff() function so I switched to use it for neovim in
> my script.
>
> Honestly, I am confused about a return value of your function. vim.diff()
> actually returns diff in a unified format as default, from which is very
> easy to generate a shortest edit script (SES).
>
> For example, how easy can I generate this SES from your return value?
>

Does your plugin use the starting line number and count in the original and
new files
for a diff hunk (just like shown in a unified diff output)?

If it does, then the diff() function return value contains these values.
To get these values,
you can use [from.startidx, from.endidx - from.startidx + 1, to.startidx,
to.endidx - to.startidx + 1]
for a diff hunk.  I see that the Neovim vim.diff() function returns these
values when
using the {result_type = 'indices'} option.

Regards,
Yegappan


> list1 = ['a', 'b', 'c', 'a', 'b', 'b', 'a']
> list2 = ['c', 'b', 'a', 'b', 'a', 'c']
> SES = '--=-=+==+'
>
>
>

vim-ml:
On Sat, May 6, 2023 at 4:44 AM rickhowe ***@***.***> wrote:

> I am interested in this builtin function as I have developed some plugins
> to support word/character level diff in lines (this
> <https://github.com/rickhowe/diffchar.vim>) and range/area selectable
> diff in buffers (this <https://github.com/rickhowe/spotdiff.vim>). I
> implemented the O(NP) comparison algorithm in vim script, but I have been
> looking for a builtin function like yours. Several month ago, I noticed
> that neovim has vim.diff() function so I switched to use it for neovim in
> my script.
>
> Honestly, I am confused about a return value of your function. vim.diff()
> actually returns diff in a unified format as default, from which is very
> easy to generate a shortest edit script (SES).
>
>
> The Neovim vim.diff() function (
https://neovim.io/doc/user/lua.html#lua-diff) returns the
diff between two strings.  It returns either a String or a List depending
on the "result_type"
option value.  If "result_type" is set to "unified" (which is the default),
then this function returns
a String which is the unified diff output between the two string
arguments.  If the "result_type"
option is set to "indices", then vim.diff() returns a List with 4 numbers
(starting line number in
the first string, line count, starting line number in the second string,
line count).

The return value of vim.diff() is not fully suitable for getting the change
information needed for
the LSP DidChangeTextDocument notification (
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentContentChangeEvent
).
For this notification, we need the range of characters in the old text
(starting and ending
text position) and the new text that replaces the old text.

I can modify the new diff() function to return a value similar to that
returned by the
Neovim vim.diff() function.  i.e. either a string with the unified diff or
a List with the indices.
We can then add an additional option ("range" or "extended" or "position")
to return
the starting and ending position of the change in the old and the new text.

Bram: What do you think about this approach?

Regards,
Yegappan

rickhowe:
It is much better to provide unified diff format. For example, we can simply use a builtin func instead of external diff command in diffexpr on neovim now:

```
set diffexpr=MyDiff()
function! MyDiff()
  let f1 = join(readfile(v:fname_in), "\n") . "\n"
  let f2 = join(readfile(v:fname_new), "\n") . "\n"
  call writefile(split(v:lua.vim.diff(f1, f2), "\n"), v:fname_out)
endfunction
```
No need to modify the result. Also to generate SES:

```
function! SES(l1, l2)
  let l1 = join(a:l1, "\n") . "\n"
  let l2 = join(a:l2, "\n") . "\n"
  let vd = v:lua.vim.diff(l1, l2, {'result_type': 'indices'})
  let ses = ''
  let p1 = 1
  for [s1, c1, s2, c2] in vd + [[len(a:l1), 0, 0, 0]]
    if c1 == 0 | let s1 += 1 | endif
    let ses .= repeat('=', s1 - p1) . repeat('-', c1) . repeat('+', c2)
    let p1 = s1 + c1
  endfor
  return ses
endfunction
```
It is simple enough to make use of it.

yegappan:
@rickhowe For these two use cases that you have cited, the unified diff output format and the diff hunk indices output format are sufficient.  But as I described in my earlier reply, to get the range of changed text for the language server, these two output information is not sufficient.  We need the start and end position (line and column number) of the change in the original text and the corresponding new text.

vim-ml:
Yegappan wrote:

> On Sat, May 6, 2023 at 4:44 AM rickhowe ***@***.***> wrote:
> 
> > I am interested in this builtin function as I have developed some plugins
> > to support word/character level diff in lines (this
> > <https://github.com/rickhowe/diffchar.vim>) and range/area selectable
> > diff in buffers (this <https://github.com/rickhowe/spotdiff.vim>). I
> > implemented the O(NP) comparison algorithm in vim script, but I have been
> > looking for a builtin function like yours. Several month ago, I noticed
> > that neovim has vim.diff() function so I switched to use it for neovim in
> > my script.
> >
> > Honestly, I am confused about a return value of your function. vim.diff()
> > actually returns diff in a unified format as default, from which is very
> > easy to generate a shortest edit script (SES).
> >
> >
> > The Neovim vim.diff() function (
> https://neovim.io/doc/user/lua.html#lua-diff) returns the
> diff between two strings.  It returns either a String or a List depending
> on the "result_type"
> option value.  If "result_type" is set to "unified" (which is the default),
> then this function returns
> a String which is the unified diff output between the two string
> arguments.  If the "result_type"
> option is set to "indices", then vim.diff() returns a List with 4 numbers
> (starting line number in
> the first string, line count, starting line number in the second string,
> line count).
> 
> The return value of vim.diff() is not fully suitable for getting the change
> information needed for
> the LSP DidChangeTextDocument notification (
> https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentContentChangeEvent
> ).
> For this notification, we need the range of characters in the old text
> (starting and ending
> text position) and the new text that replaces the old text.
> 
> I can modify the new diff() function to return a value similar to that
> returned by the
> Neovim vim.diff() function.  i.e. either a string with the unified diff or
> a List with the indices.
> We can then add an additional option ("range" or "extended" or "position")
> to return
> the starting and ending position of the change in the old and the new text.
> 
> Bram: What do you think about this approach?

The unified diff will contain less information, it only shows
added/removed/modified lines, not column information.  This then needs
to be parsed, thus a sequence of digits converted to a number.  For just
getting the information this adds overhead, thus it would only be useful
if a unified diff is actually needed.

Also being able to make a diff between two string with NL characters
seems hardly useful.  I would not know where such a string comes from.
When using buffer lines we have a list of strings.  If needed split()
can be used before passing the text to diff().

-- 
13. Go to a poetry recital and ask why the poems don't rhyme.

brammool:
> It is much better to provide unified diff format. For example, we can
> simply use a builtin func instead of external diff command in diffexpr
> on neovim now:
> 
> ```
> set diffexpr=MyDiff()
> function! MyDiff()
>   let f1 = join(readfile(v:fname_in), "\n") . "\n"
>   let f2 = join(readfile(v:fname_new), "\n") . "\n"
>   call writefile(split(v:lua.vim.diff(f1, f2), "\n"), v:fname_out)
> endfunction
> ```

Using join() is very inefficient, lots of memory allocation and moving
text around.  The 'diffexpr' was really intended for using an external
program, that's why it uses files.  When using some internal diff
implementation going through files is a weird detour.  Thus this example
is not relevant.

> No need to modify the result. Also to generate SES:
> 
> ```
> function! SES(l1, l2)
>   let l1 = join(a:l1, "\n") . "\n"
>   let l2 = join(a:l2, "\n") . "\n"
>   let vd = v:lua.vim.diff(l1, l2, {'result_type': 'indices'})
>   let ses = ''
>   let p1 = 1
>   for [s1, c1, s2, c2] in vd + [[len(a:l1), 0, 0, 0]]
>     if c1 == 0 | let s1 += 1 | endif
>     let ses .= repeat('=', s1 - p1) . repeat('-', c1) . repeat('+', c2)
>     let p1 = s1 + c1
>   endfor
>   return ses
> endfunction
> ```
> It is simple enough to make use of it.

What is SES?

-- 
Giving a commit hash to refer to a patch is like giving longitude and
lattitude to refer to a city.

vimpostor:
> What is SES?

It stands for shortest edit script: https://arxiv.org/abs/2208.08823

> I can modify the new diff() function to return a value similar to that
> returned by the
> Neovim vim.diff() function.  i.e. either a string with the unified diff or
> a List with the indices.
> We can then add an additional option ("range" or "extended" or "position")
> to return
> the starting and ending position of the change in the old and the new text.
> 
> Bram: What do you think about this approach?

I think adding an option to select the behaviour would be a good idea, as it allows for all usecases.
Even though Bram is right about the `join()` being inefficient as a `diffexpr`, the unified diff may still be useful for some other usecases.
But of course it's @brammool call in which direction to go here.

chrisbra:
@yegappan I think this is a good idea:

> I can modify the new diff() function to return a value similar to that
returned by the Neovim vim.diff() function.  i.e. either a string with the unified diff or
a List with the indices.
We can then add an additional option ("range" or "extended" or "position")
to return the starting and ending position of the change in the old and the new text.

Can you make this change please?

yegappan:
> @yegappan I think this is a good idea:
> 
> > I can modify the new diff() function to return a value similar to that
> > returned by the Neovim vim.diff() function.  i.e. either a string with the unified diff or
> > a List with the indices.
> > We can then add an additional option ("range" or "extended" or "position")
> > to return the starting and ending position of the change in the old and the new text.
> 
> Can you make this change please?

@chrisbra I will work on this in a few days.

yegappan:
> @yegappan I think this is a good idea:
> 
> > I can modify the new diff() function to return a value similar to that
> > returned by the Neovim vim.diff() function.  i.e. either a string with the unified diff or
> > a List with the indices.
> > We can then add an additional option ("range" or "extended" or "position")
> > to return the starting and ending position of the change in the old and the new text.
> 
> Can you make this change please?

I have updated the PR to return either the unified diff or the indices.  In a later PR, I will
add the support for returning the range information which is needed for LSP.

chrisbra:
Thanks @yegappan

errael:
@yegappan Somewhat off topic... 

I work with a port of a 3 way merge tool. I'm wondering if having the kind of information provided by `diff()` might be useful for adding features. The tool has up to 4 buffers involved in a diff. This new `diff()` is the closest I've seen for getting info about diffs (but I could easily have missed something). I'm wondering how hard it would be to get this info for active diff buffers.

This is not a feature request; I'm curious if `diff()`s implementation would be a good guide for getting the data on active diff buffers if it turns out to be useful for the merge tool.

yegappan:
> @yegappan Somewhat off topic...
> 
> I work with a port of a 3 way merge tool. I'm wondering if having the kind of information provided by `diff()` might be useful for adding features. The tool has up to 4 buffers involved in a diff. This new `diff()` is the closest I've seen for getting info about diffs (but I could easily have missed something). I'm wondering how hard it would be to get this info for active diff buffers.
> 
> This is not a feature request; I'm curious if `diff()`s implementation would be a good guide for getting the data on active diff buffers if it turns out to be useful for the merge tool.

@errael What sort of information are you looking for in an active diff buffer?  Which lines are added, removed and modified in a diff buffer?

rickhowe:
Hi,

Thank you for providing diff(). I checked how it works while comparing with nvim's vim.diff() as follows.

```
diff(['a'], ['x', 'a'])
unified: @@ -0,0 +1 @@
indices: {0, 0, 0, 1} *
nvim:    [0, 0, 1, 1]

diff(['a'], ['a', 'x'])
unified: @@ -1,0 +2 @@
indices: {1, 0, 1, 1} *
nvim:    [1, 0, 2, 1]

diff(['x', 'a'], ['a'])
unified: @@ -1 +0,0 @@
indices: {0, 1, 0, 0} *
nvim:    [1, 1, 0, 0]

diff(['a', 'x'], ['a'])
unified: @@ -2 +1,0 @@
indices: {1, 1, 1, 0} *
nvim:    [2, 1, 1, 0]

*: {from_idx, from_count, to_idx, to_count}
```

I am a bit confused about idx, In unified, idx 0 always means '^'. But in indices, idx 0 means '^' if count = 0 else the first index of hunk. In nvim, vim.diff() returns the equivalent value of the unified as indices.

Is it difficult to follow vim.diff() as indices to make it simple and consistent between vim and nvim?

errael:
> > @yegappan Somewhat off topic...
> > I work with a port of a 3 way merge tool. I'm wondering if having the kind of information provided by `diff()` might be useful for adding features. The tool has up to 4 buffers involved in a diff. This new `diff()` is the closest I've seen for getting info about diffs (but I could easily have missed something). I'm wondering how hard it would be to get this info for active diff buffers.
> > This is not a feature request; I'm curious if `diff()`s implementation would be a good guide for getting the data on active diff buffers if it turns out to be useful for the merge tool.
> 
> @errael What sort of information are you looking for in an active diff buffer? Which lines are added, removed and modified in a diff buffer?

I'm only beginning to explore outstanding issues on the original merge tool. So I don't really know what information I might need. When I saw this PR/issue it seemed worth exploring. The merge tool mostly uses vim builtin commands, like `:diffget` and `:diffput`. 

There's a request for a command that merges **both** sides into the result; there have been times I wanted that myself. Seems it might be tricky. Maybe a popup that show a some options and lets the user pick the one that's closet to what they want; the user will probably still need to do some editing to get the right result. I'm wondering if having the specific diff information would help. I'm hoping I can determine the text from both sides (I mean the text that would be used by vim's `:diffput`/`:diffget`) and then I need to come up with some algorithms to provide options on how to combine both sides.

rickhowe:
> When the from_count is zero, the strings from to_idx to (to_idx + to_count) in {list2} are inserted at from_idx in {list1}. When the to_count is zero, the strings from from_idx to (from_idx + from_count) are removed from {list2} at to_idx.

Let me make sure the relation between unified and indices.
```
unified: @@ -from_line,from_count +to_line,to_count @@
indices: {from_idx, from_count, to_idx, to_count}

for both from_ and to_:

if 0 < conut
  line = idx + 1
else
  line = idx
endif
```
Is that correct?

yegappan:
> > When the from_count is zero, the strings from to_idx to (to_idx + to_count) in {list2} are inserted at from_idx in {list1}. When the to_count is zero, the strings from from_idx to (from_idx + from_count) are removed from {list2} at to_idx.
> 
> Let me make sure the relation between unified and indices.
> 
> ```
> unified: @@ -from_line,from_count +to_line,to_count @@
> indices: {from_idx, from_count, to_idx, to_count}
> 

The `unified` diff uses line numbers which start at 1 whereas `indices` returns
List indexes which start at 0.

> for both from_ and to_:
> 
> if 0 < conut
>   line = idx + 1
> else
>   line = idx
> endif
> ```
> 
> Is that correct?

No.  In both the cases, the line number is `idx + 1`.

Some examples below:
```
  :echo diff(['1', '2'], ['0', '1', '2'], {'output': 'indices'})
   [{'from_count': 0, 'to_idx': 0, 'to_count': 1, 'from_idx': 0}]

  :echo diff(['1', '2'], ['1', '2', '3'], {'output': 'indices'})
   [{'from_count': 0, 'to_idx': 2, 'to_count': 1, 'from_idx': 2}]
```

In both the examples, the `from_count` is 0 and the `to_count` is 1.  In the first example, the string at index 0 (`to_idx`) in the second List is inserted at index 0 (`from_idx`) in List1.   In the second example, the string at index 2 (`to_idx`) in the second List  is inserted at index 2 (`from_idx) in List1.

rickhowe:
I checked again on your examples about how unified relates to indices.

> ```
>   :echo diff(['1', '2'], ['0', '1', '2'], {'output': 'indices'})
>    [{'from_count': 0, 'to_idx': 0, 'to_count': 1, 'from_idx': 0}]
> ```

unified: @@ -0,0 +1 @@

from_line (0) = from_idx (0) (where from_conut = 0)
to_line (1) = to_idx (0) + 1 (where 0 < to_count)

> ```
>   :echo diff(['1', '2'], ['1', '2', '3'], {'output': 'indices'})
>    [{'from_count': 0, 'to_idx': 2, 'to_count': 1, 'from_idx': 2}]
> ```

unified: @@ -2,0 +3 @@

from_line (2) = from_idx (2) (where from_conut = 0)
to_line (3) = to_idx (2) + 1 (where 0 < to_count)

> No. In both the cases, the line number is `idx + 1`.

 The line number is not always `idx + 1` but is `idx` if `count = 0`, right?

rickhowe:
This is a different question. It looks like that diff()` calls a function which is set in &diffexpr. &diffopt says that `internal` is ignored when &diffexpr is set. It is not necessary for 'diff()` to work along with those options, right?

yegappan:
> This is a different question. It looks like that `diff()` calls a function which is set in &diffexpr. &diffopt says that `internal` is ignored when &diffexpr is set. It is not necessary for `diff()` to work along with those options, right?

This is a bug.  Can you open a separate issue for this?  I will open a PR to fix this.

errael:
Excuse my ignorance of the details of this, but a few days ago, in https://github.com/vim/vim/pull/12321#issuecomment-1926128065

There's
> I checked again on your examples about how unified relates to indices.
> [snip]
> The line number is not always `idx + 1` but is `idx` if `count = 0`, right?

Does this need to be addressed/documented/???

yegappan:
> I checked again on your examples about how unified relates to indices.
> 
> > ```
> >   :echo diff(['1', '2'], ['0', '1', '2'], {'output': 'indices'})
> >    [{'from_count': 0, 'to_idx': 0, 'to_count': 1, 'from_idx': 0}]
> > ```
> 
> diff() returns as unified: @@ -0,0 +1 @@ then, from_line (0) = from_idx (0) (where from_conut = 0) to_line (1) = to_idx (0) + 1 (where 0 < to_count)

The values returned for the 'indices' option always refer to the item indices in
`List1` and `List2`.  So the corresponding line number is idx + 1.

The unified diff output returns line numbers.  But there is a special case when
a change is made to the very first line.  In this case it returns 0.

What is the use case for comparing the unified diff output and the indices?

> 
> > ```
> >   :echo diff(['1', '2'], ['1', '2', '3'], {'output': 'indices'})
> >    [{'from_count': 0, 'to_idx': 2, 'to_count': 1, 'from_idx': 2}]
> > ```
> 
> diff() returns as unified: @@ -2,0 +3 @@ then, from_line (2) = from_idx (2) (where from_conut = 0) to_line (3) = to_idx (2) + 1 (where 0 < to_count)
> 
> > No. In both the cases, the line number is `idx + 1`.
> 
> The line number is not always `idx + 1` but is `idx` if `count = 0`, right?

yegappan:
> Excuse my ignorance of the details of this, but a few days ago, in [#12321 (comment)](https://github.com/vim/vim/pull/12321#issuecomment-1926128065)
> 
> There's
> 
> > I checked again on your examples about how unified relates to indices.
> > [snip]
> > The line number is not always `idx + 1` but is `idx` if `count = 0`, right?
> 
> Does this need to be addressed/documented/???

This needs an documentation update.

rickhowe:
> The unified diff output returns line numbers. But there is a special case when a change is made to the very first line. In this case it returns 0.

Yes, that is why I am asking. Let me make sure that the line number is not always idx + 1 but is idx if count = 0.

> What is the use case for comparing the unified diff output and the indices?

[My plugin](https://github.com/rickhowe/diffchar.vim) will use a builtin diff function in vim and nvim and handle both outputs. I need to understand the exact difference between them. And, for example in &diffexpr, it may be possible that a function reads `v:fname_in` and `v:fname_new` files, handles and refines those diff as indices, and writes to `v:fname_out` file as unified.

yegappan:
> > The unified diff output returns line numbers. But there is a special case when a change is made to the very first line. In this case it returns 0.
> 
> Yes, that is why I am asking. Let me make sure that the line number is not always idx + 1 but is idx if count = 0.
> 

I have created PR https://github.com/vim/vim/pull/14010 to add support for specifying the optional
unified diff context length.  If you specify the 'context' length as 1 or above, then you will get 1 as the
line number for these case.

rickhowe:
> > Yes, that is why I am asking. Let me make sure that the line number is not always idx + 1 but is idx if count = 0.
> 
> I have created PR #14010 to add support for specifying the optional unified diff context length. If you specify the 'context' length as 1 or above, then you will get 1 as the line number for these case.

I am confused again. I do not know how useful the context length is in vim script. I just want to make sure the relation between unified and indices.

Is that same as `ctxlen` in `vim.diff()` on nvim?

yegappan:
> > > Yes, that is why I am asking. Let me make sure that the line number is not always idx + 1 but is idx if count = 0.
> > 
> > 
> > I have created PR #14010 to add support for specifying the optional unified diff context length. If you specify the 'context' length as 1 or above, then you will get 1 as the line number for these case.
> 
> I am confused again. I do not know how useful the context length is in vim script. I just want to make sure the relation between unified and indices.
> 
> Is that same as `ctxlen` in `vim.diff()` on nvim? If so, I will not use it (use 0 as default).

Yes. It is same as the `ctxlen` item in nvim.

rickhowe:
> > Is that same as `ctxlen` in `vim.diff()` on nvim? If so, I will not use it (use 0 as default).
> 
> Yes. It is same as the `ctxlen` item in nvim.

OK but why its default  is 1, which is different from `vim.diff()`?

rickhowe:
Another topic:
If you would update diff,txt help file, it is time to exclude external "diff" command. For example, in `diff-diffexpr` section, how would you like to introduce your diff(), such as:

```
Example (this does almost the same as 'diffexpr' being empty): >

set diffexpr=MyDiff()
function MyDiff()
  let in = readfile(v:fname_in)
  let new = readfile(v:fname_new)
  let out = diff(in, new, {'icase': &diffopt =~ "icase", 'iwhite': &diffopt =~ "iwhite"})
  call writefile(split(out, "\n"), v:fname_out)
endfunction
```

yegappan:
> > > Is that same as `ctxlen` in `vim.diff()` on nvim? If so, I will not use it (use 0 as default).
> > 
> > 
> > Yes. It is same as the `ctxlen` item in nvim.
> 
> OK but why its default is 1, which is different from `vim.diff()`?

When the context length is 0, diff optimizes the output and the line number in a diff hunk can
be 0 in a few cases.  When the context length is 1, then the number of cases where the line
number is 0 is reduced and this will help in avoiding confusion with the line number.

Note that the Linux "diff" command uses a default value of 3 for the context length.

yegappan:
> Another topic: If you would update diff,txt help file, it is time to exclude external "diff" command. For example, in `diff-diffexpr` section, how would you like to introduce your diff(), such as:
> 
> ```
> Example (this does almost the same as 'diffexpr' being empty): >
> 
> set diffexpr=MyDiff()
> function MyDiff()
>   let in = readfile(v:fname_in)
>   let new = readfile(v:fname_new)
>   let out = diff(in, new, {'icase': &diffopt =~ "icase", 'iwhite': &diffopt =~ "iwhite"})
>   call writefile(split(out, "\n"), v:fname_out)
> endfunction
> ```

Can you create a PR with this example added to diff.txt (in addition to the existing example
for using an external command)?

rickhowe:
> > OK but why its default is 1, which is different from `vim.diff()`?
> 
> When the context length is 0, diff optimizes the output and the line number in a diff hunk can be 0 in a few cases. When the context length is 1, then the number of cases where the line number is 0 is reduced and this will help in avoiding confusion with the line number.

No.  line number 0 is not a special case and there is no confusion in existing diff. And I need diff() to optimize the result as a default. What I have confused is that a relation between unified and 0-based indices.

> Note that the Linux "diff" command uses a default value of 3 for the context length.

Yes, because it is a command, A user can directly see the result. We are talking about diff() function.

I found that `MyDiff()` using diff() does not work, because vim does not accept context:1 unified format. `diff-diffexpr` section says:
```
For a unified diff no context lines can be used.
Using "diff -u" will NOT work, use "diff -U0".
```
Do we need to always specify context:0 to diff()?

yegappan:
> > > OK but why its default is 1, which is different from `vim.diff()`?
> > 
> > 
> > When the context length is 0, diff optimizes the output and the line number in a diff hunk can be 0 in a few cases. When the context length is 1, then the number of cases where the line number is 0 is reduced and this will help in avoiding confusion with the line number.
> 
> No. line number 0 is not a special case and there is no confusion in existing diff. And I need diff() to optimize the result as a default. What I have confused is that a relation between unified and 0-based indices.
> 
> > Note that the Linux "diff" command uses a default value of 3 for the context length.
> 
> Yes, because it is a command, A user can directly see the result. We are talking about diff() function.
> 
> I found that `MyDiff()` using diff() does not work, because vim does not accept context:1 unified format. `diff-diffexpr` section says:
> 
> ```
> For a unified diff no context lines can be used.
> Using "diff -u" will NOT work, use "diff -U0".
> ```
> 
> Do we need to always specify context:0 to diff()?

Good point.  I missed this note in the documentation about the unified diff expecting context length of 0.
We can use 0 as the default context length for the diff() function.

PR #12325, opened 2023-05-01
Fix strange error when content of heredoc looks like heredoc

zeertzjq:


brammool:
Why did you change the argument names of functions such as skipwhite()?
The reason a local variable is used is that this way the original string
can be inspected in a debugger.  In general, function arguments should
not be changed inside a function.

I do not like using "<<" directly after ":lua", ":ruby", etc., without
whitespace.  I do not think this was ever possible, and I see no good
reason to start supporting it now.  The functionality is the same, the
command is less readable.  I rather encourage users to write commands in
a readable way, by separating arguments with whitespace.

-- 
Back off man, I'm a scientist.
              -- Peter, Ghostbusters

zeertzjq:
Oh wait, some of the is still applicable.

zeertzjq:
I've update this PR to only keep changes related to `:let`.

zeertzjq:
AppVeyor seems to have some Internet connection problems.

PR #12331, opened 2023-05-02
Remove incorrect use of W_WINROW

zeertzjq:
In structs.h it is mentioned that w_wrow is relative to w_winrow, so
using W_WINROW doesn't make sense when comparing with window height.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -525,8 +525,8 @@ edit(
 #else
 		(int)curwin->w_wcol < mincol - curbuf->b_p_ts
 #endif
-		    && curwin->w_wrow == W_WINROW(curwin)
-				 + curwin->w_height - 1 - get_scrolloff_value()
+		    && curwin->w_wrow ==
+				   curwin->w_height - 1 - get_scrolloff_value()
 		    && (curwin->w_cursor.lnum != curwin->w_topline
 #ifdef FEAT_DIFF
 			|| curwin->w_topfill > 0


brammool:
> In structs.h it is mentioned that w_wrow is relative to w_winrow, so
> using W_WINROW doesn&#39;t make sense here.

Hmm, there are other hints suggesting that comment is wrong and the
code is right.  Various other places where w_wrow is used it is added to
W_WINROW().  I think that comment was written before w_winbar_height was
added.

If the code does need correction, it should be possible to see the
effect with a test.

Unfortunately these computations have become quite hard to understand
because of the features that have been added over time.  Refactoring
this will not be easy.  There might also be some more places where
w_winbar_height is not taken into account while it should.

-- 
I think that you'll agree that engineers are very effective in their social
interactions.  It's the "normal" people who are nuts.

zeertzjq:
Looking at this again, I still thin this change doesn't lead to any observable behavior change.

The condition intends to check if there are 'scrolloff' lines between the current cursor when the bottom of the window. When `W_WINROW(curwin)` is added, the condition is instead satisfied when the cursor is on some screen below that position. However,
- If 'scrolloff' is smaller than half the window height, this condition can only be satisfied when `W_WINROW(curwin) == 0`. And if it is not satisfied, `update_topline()` does the actual scrolling.
- If 'scrolloff' is larger than half the window height, `update_topline()` will put the cursor at the center of the window soon afterwards anyway, because `set_topline()` no longer sets `VALID_TOPLINE` starting from <https://github.com/vim/vim-history/commit/7db7bb45b0f919ff0615d463ebd4fde881c69d1f>.

Or, to put it in another way, <https://github.com/vim/vim-history/commit/7db7bb45b0f919ff0615d463ebd4fde881c69d1f> make `update_topline()` correct the mistakes made in this block, so this incorrect use of `W_WINROW()` no longer affects observable behavior.

chrisbra:
Thanks, let me merge this now. Hopefully this won't break anything.

PR #12337, opened 2023-05-03
Modernize Objective-C

LumioseSil:
I know it sounds silly, but at the end of the day, this is going to be more efficient and easier to read.


brammool:
This appears to depend on the " @autoreleasepool" annotation.  What versions of objective-C support this, when was it introduced and working properly?

The changes look very obscure to me.  Please add more explanation about what this change is doing.  I hope some developers familiar with Objective-C can comment on whether this is safe to include and an actual improvement.

ychin:
I'm curious what the lowest macOS that we currently support now. I was surprised to hear that right now, Vim actually still builds on macOS 10.4 (https://github.com/vim/vim/pull/13115). I think the `@autoreleasepool` syntax is a little cutting it a little tight, and I think ARC (Automatic Release Counting) was introduced even later in WWDC 2011? I'm not sure if that is considered old enough by Vim standards (which I admit is not very standard).

ychin:
Note: The sound code was written a little later and hence supported newer syntax, and when I wrote it I didn't really realize the impact it had on older macOS versions (per the issue I linked to above), but at least we can just turn it off as it's a new feature. Other stuff are existing features that already work so I'm a little wary of this breaking older macOS builds. I do agree the current Objective C is quite old-school though.

src/os_macosx.m
```diff
@@ -421,18 +418,14 @@ - (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag
...
-    // holds weak reference to this object.
-    [self release];
 }
 @end
 
     void
 process_cfrunloop(void)
 {
-    if (sounds_list != nil && [sounds_list count] > 0)
+    if (sounds_list != nil && [sounds_list count])
```
ychin, line 428: FWIW I don't particularly like using `if (<int>) {}`. It loses the semantic meaning of what you are trying to do and relying on an implicit behavior. As an optimization it also wouldn't gain much as the compiler understands how to optimize `!= 0` etc away. It's a little easier to write accidental bugs when not spelling out the condition this way.

chrisbra:
@ychin what's your opinion on this patch please? @AreaZR can you please rebase and fix the above mentioned comment?

AreaZR:
Alright.

src/os_macosx.m
```diff
@@ -501,9 +494,9 @@ - (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag
 	{
 	    for (NSSound *sound in [sounds_list allValues])
 	    {
-		[sound stop];
+		    [sound stop];
```
ychin, line 497: Unnecessary indentation change?

src/os_macosx.m
```diff
@@ -421,18 +418,14 @@ - (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag
 	    delete_sound_callback(_callback);
 	    _callback = NULL;
 	}
-	[sounds_list removeObjectForKey:[NSNumber numberWithLong:_sound_id]];
```
ychin, line 424: Isn't this going to leak with this change? You moved this to `sound_mch_stop()` but that function is only called if you manually stop a playing sound using the `sound_stop()` API. This delegate is going to get called when you do that, but it's **also** going to be called when the sound finishes playing naturally. This function was placed here for that reason, so that in both cases the item gets removed from the `sounds_list` dictionary properly.

src/os_macosx.m
```diff
@@ -482,12 +475,12 @@ - (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag
 {
     @autoreleasepool
     {
-	NSSound *sound = sounds_list[[NSNumber numberWithLong:sound_id]];
+    NSNumber *_sound_id = [NSNumber numberWithLong:sound_id];
+	NSSound *sound = sounds_list[_sound_id];
 	if (sound != nil)
 	{
-	    // Stop the sound. No need to release it because the delegate will do
```
ychin, line 488: This comment should not be removed, since I think we should remove the sound from sound_list at the delegate.

src/os_macosx.m
```diff
@@ -482,12 +475,12 @@ - (void) sound:(NSSound *)sound didFinishPlaying:(BOOL)flag
...
-	    // it for us.
 	    [sound stop];
+        [sounds_list removeObjectForKey:_sound_id];
```
ychin, line 483: I think this should be done at the delegate per the other [comment](https://github.com/vim/vim/pull/12337/files#r1835587863) mentioned.

ychin:
> @ychin what's your opinion on this patch please? @AreaZR can you please rebase and fix the above mentioned comment?

@chrisbra I think my comment (https://github.com/vim/vim/pull/12337#issuecomment-1758719789) is still unresolved:

> I'm curious what the lowest macOS that we currently support now. I was recently surprised to hear that right now, Vim actually still builds on macOS 10.4 (#13115). I think the `@autoreleasepool` syntax is already cutting it a little tight, and I think ARC (Automatic Release Counting) (which was turned on in this PR) was introduced even later in WWDC 2011? I'm not sure if that is considered old enough by Vim standards (which I admit is not very standard). 2011 was macOS 10.6 Snow Leopard I think.

At least from #13115, *someone* (as in, at least one person), still cares about building Vim on older Macs / Xcode versions. The tooling used by the person who filed that issue was really old, as he's targeting macOS 10.4 - 10.6 support, where Vim apparently still works for. The main change this PR is doing is moving from manual reference counting to Automatic Reference Counting (ARC) which was introduced in Xcode 4.2 which was released in 2011, and supports backwards compatibility down to deploying to macOS 10.6 (released 2009). I'm not sure why @AreaZR (https://github.com/vim/vim/pull/12337#issuecomment-1536323265) said anything newer than 2006 as after some research I'm pretty sure it requires LLVM 3 / Xcode 4.2 to work.

In general, Apple heavily encourages switching to using ARC because of reduced chances of programming error and potentially better performance, but I just wanted to point out that this change would be an intentional increase in system requirements for Vim for old Mac compatibility. In terms of gains it's mostly a code cleanup and marginal performance and reliability improvements.

---

Either way I think this PR has introduced some bugs in the sound feature so I'm looking through that still and have some other comments that I will post in a bit.

chdiza:
> 2011 was macOS 10.6 Snow Leopard I think.

No, it was Lion (10.7), in the middle of summer of 2011.

ychin:
Right. As I mentioned in the later comment it Snow Leopard was 2009. ARC was designed for Lion (10.7) but was back ported to Snow Leopard (10.6) by deploying some aspects of the runtime along with the code. I got confused when I first typed it.

chrisbra:
We dropped Windows XP and Vista support some time ago ([v9.0.0496](https://github.com/vim/vim/releases/tag/v9.0.0496)), which was supported until 2014 (XP)/2017 (Vista). I would say anything significant older than 10 years (2014) could be considered un-supported. So if we raise minimum requirement on Mac to Lion 10.7 (released 2011), that is okay and we do not have to neccessarily support Snow Leopard.

So if this patch causes us to drop Snow Leopard support, I think that is acceptable.

chrisbra:
I tried to document current operating system status here: `:h os-support` (as of 18defabae)

chrisbra:
alright, let me put it into DRAFT. @AreaZR are you ready to improve this? Otherwise I am going to drop it here completely.

chrisbra:
I am closing this, as there is no activity here.

PR #12341, opened 2023-05-04
Behavior of assert_false('False') is confusing

zeertzjq:
Using `assert_false('False')` leaving the following in `v:errors`:

	Expected 'False' but got 'False'

which looks a bit confusing.

Solution:   Only add quotes to expected value when using assert_fails().


zeertzjq:
I've also created another PR #12342 that is a superset of this PR and attempts to fix another I found.

PR #12342, opened 2023-05-04
{msg} argument of assert_inrange() behaves differently from others

zeertzjq:
In assert_inrange(), the {msg} argument replaces the default message,
whereas in other assert functions {msg} is prepended to the default
message.

This PR changes the behavior of assert_inrange() to also prepend {msg}
to the default message by using fill_assert_error().

However, the help for all assert function seems to suggest that {msg}
replaces the default message. Maybe this needs change?

This PR is a superset of #12341.

### DIFF
--- a/src/testdir/test_assert.vim
+++ b/src/testdir/test_assert.vim
@@ -9,11 +9,11 @@ func Test_assert_false()
   call assert_equal(0, v:false->assert_false())
 
   call assert_equal(1, assert_false(123))
-  call assert_match("Expected 'False' but got 123", v:errors[0])
+  call assert_match("Expected False but got 123", v:errors[0])
   call remove(v:errors, 0)
 
   call assert_equal(1, 123->assert_false())
-  call assert_match("Expected 'False' but got 123", v:errors[0])
+  call assert_match("Expected False but got 123", v:errors[0])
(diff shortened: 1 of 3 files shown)


brammool:
> In assert_inrange(), the {msg} argument replaces the default message,
> whereas in other assert functions {msg} is prepended to the default
> message.
> 
> This PR changes the behavior of assert_inrange() to also prepend {msg}
> to the default message by using fill_assert_error().
> 
> However, the help for all assert function seems to suggest that {msg}
> replaces the default message. Maybe this needs change?

Yes, the intention is that the message is added.  The idea behind is
that normally the location of the error is sufficient to see the context
of the assert and figure out what is wrong.  However, when the assert is
in a function called from multiple places, or inside a for loop that
tries different values, the context may be unclear.  The extra message
can then provide the information needed to see in what case the assert
is failing.

I'll adjust the help text, it was a bit confusing anyway.  Most of them
should say "When {msg} is present it is prefixed to the message".
I haven't checked if it is actually prefixed, some may append {msg}.  We
should probably fix that eventually to make it consistent.

-- 
Engineers are widely recognized as superior marriage material: intelligent,
dependable, employed, honest, and handy around the house.

PR #12346, opened 2023-05-05
Add filecopy()

Shougo:
I have added `filecopy()` function in `todo.txt`.
It is useful to create file manager.

It supports both file and directory.

~~Todo: symbolic link support.~~

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -2761,6 +2761,7 @@ filecopy({from}, {to})					*filecopy()*
 		result is a Number, which is 0 if the file was copied
 		successfully, and non-zero when the copy failed.
 		If {to} exists, it will fail.
+		Note It can not copy directories.
 
 		This function is not available in the |sandbox|.
 
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -2760,7 +2760,7 @@ filecopy({from}, {to})					*filecopy()*
 		Copy the file by the name {from} to the name {to}.  The
 		result is a Number, which is 0 if the file was copied
 		successfully, and non-zero when the copy failed.
-		If {to} exists, it is failed.
+		If {to} exists, it will fail.
 
 		This function is not available in the |sandbox|.
 
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -178,6 +178,7 @@ extendnew({expr1}, {expr2} [, {expr3}])
 				List/Dict like |extend()| but creates a new
 					List or Dictionary
 feedkeys({string} [, {mode}])	Number	add key sequence to typeahead buffer
+filecopy({from}, {to})		any	copy file from {from} to {to}
 filereadable({file})		Number	|TRUE| if {file} is a readable file
 filewritable({file})		Number	|TRUE| if {file} is a writable file
 filter({expr1}, {expr2})	List/Dict/Blob/String
@@ -2754,6 +2755,16 @@ feedkeys({string} [, {mode}])				*feedkeys()*
 
 		Can also be used as a |method|: >
 			GetInput()->feedkeys()
+
+filecopy({from}, {to})					*filecopy()*
+		Copy the file by the name {from} to the name {to}.  The
+		result is a Number, which is 0 if the file was copied
+		successfully, and non-zero when the copy failed.
+		NOTE: If {to} exists it is overwritten without warning.
+		This function is not available in the |sandbox|.
+
+		Can also be used as a |method|: >
+			GetOldName()->filecopy(newname)
 
 filereadable({file})					*filereadable()*
 		The result is a Number, which is |TRUE| when a file with the
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -178,7 +178,7 @@ extendnew({expr1}, {expr2} [, {expr3}])
 				List/Dict like |extend()| but creates a new
 					List or Dictionary
 feedkeys({string} [, {mode}])	Number	add key sequence to typeahead buffer
-filecopy({from}, {to} [, {flags}])	Number	copy file from {from} to {to}
+filecopy({from}, {to})		Number	copy file from {from} to {to}
 filereadable({file})		Number	|TRUE| if {file} is a readable file
 filewritable({file})		Number	|TRUE| if {file} is a writable file
 filter({expr1}, {expr2})	List/Dict/Blob/String
@@ -2756,17 +2756,11 @@ feedkeys({string} [, {mode}])				*feedkeys()*
 		Can also be used as a |method|: >
 			GetInput()->feedkeys()
 
-filecopy({from}, {to} [, {flags}])				*filecopy()*
+filecopy({from}, {to})					*filecopy()*
 		Copy the file by the name {from} to the name {to}.  The
 		result is a Number, which is 0 if the file was copied
 		successfully, and non-zero when the copy failed.
-
-		When {flags} is "f": If {to} exists it is overwritten without
-		warning.
-
-		The result is a Number, which is 0/false if the copy
-		operation was successful, which is -2 if {to} exists and
-		-1/true when the copy failed.
+		If {to} exists, it is failed.
 
 		This function is not available in the |sandbox|.
 
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -178,7 +178,7 @@ extendnew({expr1}, {expr2} [, {expr3}])
 				List/Dict like |extend()| but creates a new
 					List or Dictionary
 feedkeys({string} [, {mode}])	Number	add key sequence to typeahead buffer
-filecopy({from}, {to})		any	copy file from {from} to {to}
+filecopy({from}, {to} [, {flags}])	Number	copy file from {from} to {to}
 filereadable({file})		Number	|TRUE| if {file} is a readable file
 filewritable({file})		Number	|TRUE| if {file} is a writable file
 filter({expr1}, {expr2})	List/Dict/Blob/String
@@ -2756,11 +2756,18 @@ feedkeys({string} [, {mode}])				*feedkeys()*
 		Can also be used as a |method|: >
 			GetInput()->feedkeys()
 
-filecopy({from}, {to})					*filecopy()*
+filecopy({from}, {to} [, {flags}])				*filecopy()*
 		Copy the file by the name {from} to the name {to}.  The
 		result is a Number, which is 0 if the file was copied
 		successfully, and non-zero when the copy failed.
-		NOTE: If {to} exists it is overwritten without warning.
+
+		When {flags} is "f": If {to} exists it is overwritten without
+		warning.
+
+		The result is a Number, which is 0/false if the copy
+		operation was successful, which is -2 if {to} exists and
+		-1/true when the copy failed.
+
 		This function is not available in the |sandbox|.
 
 		Can also be used as a |method|: >
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -3936,7 +3936,7 @@ vim_copyfile(char_u *from, char_u *to)
     ret = mch_lstat((char *)from, &st);
     if (ret >= 0 && S_ISLNK(st.st_mode))
     {
-        ret = -1;
+        ret = FAIL;
 
 	len = readlink((char *)from, linkbuf, MAXPATHL);
 	if (len > 0)
@@ -3962,7 +3962,7 @@ vim_copyfile(char_u *from, char_u *to)
 #ifdef HAVE_ACL
 	mch_free_acl(acl);
 #endif
-	return -1;
+	return FAIL;
     }
 
     // Create the new file with same permissions as the original.
@@ -3974,7 +3974,7 @@ vim_copyfile(char_u *from, char_u *to)
 #ifdef HAVE_ACL
 	mch_free_acl(acl);
 #endif
-	return -1;
+	return FAIL;
     }
 
     buffer = alloc(WRITEBUFSIZE);
@@ -3985,7 +3985,7 @@ vim_copyfile(char_u *from, char_u *to)
 #ifdef HAVE_ACL
 	mch_free_acl(acl);
 #endif
-	return -1;
+	return FAIL;
     }
 
     while ((n = read_eintr(fd_in, buffer, WRITEBUFSIZE)) > 0)
@@ -4017,9 +4017,9 @@ vim_copyfile(char_u *from, char_u *to)
     if (errmsg != NULL)
     {
 	semsg(errmsg, to);
-	return -1;
+	return FAIL;
     }
-    return 0;
+    return OK;
 }
 
 static int already_warned = FALSE;
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -3898,7 +3898,7 @@ vim_rename(char_u *from, char_u *to)
      */
     ret = vim_copyfile(from, to);
     if (ret != OK)
-	return ret;
+	return -1;
 
     /*
      * Remove copied original file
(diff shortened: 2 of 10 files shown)


tani:
I think this is a very useful function. First, it would allow files to be copied without going through a temporary buffer, which would prevent other plug-ins from malfunctioning with respect to the temporary buffer. It would also be faster and use less memory. A similar approach would be to launch an external process to copy the files, but this requires consideration of differences between operating systems. Furthermore, string command operations require constant attention. I support this change as it is a promising approach that is both safe and fast.

On the other hand, I point out that the function name is inappropriate: the name filecopy, with file being singular and prefixed with file, suggests that only one file is to be copied. If you follow vim's traditional naming convention, the verb should come first (e.g. isdirectory). It is also possible to copy directories recursively by specifying not only files but also directories. In particular, if the directory is empty, the function will not copy any files, only the directory. This is not consistent with the name and behavior.

I am not familiar with vim conventions, so I will defer the specific naming to others, but I encourage you to reconsider for the above reasons.

Shougo:
> On the other hand, I point out that the function name is inappropriate: the name filecopy, with file being singular and prefixed with file, suggests that only one file is to be copied. If you follow vim's traditional naming convention, the verb should come first (e.g. isdirectory). It is also possible to copy directories recursively by specifying not only files but also directories. In particular, if the directory is empty, the function will not copy any files, only the directory. This is not consistent with the name and behavior.

The name is not good.  But it is already todo listed and I don't know what is the best name.

tani:
I understand the situation. So it may not be appropriate to discuss it here. I would like to propose this matter not as a proposal to the author of this pull request, but as a proposal to those who have merge authority.

src/filepath.c
```diff
@@ -2648,6 +2648,27 @@ f_browsedir(typval_T *argvars UNUSED, typval_T *rettv)
...
+    void
+f_filecopy(typval_T *argvars, typval_T *rettv)
+{
+    char_u	buf[NUMBUFLEN];
+
+    rettv->vval.v_number = -1;
+    if (check_restricted() || check_secure())
+	return;
+
+    if (in_vim9script()
```
yegappan: As this is a new function, there is no need for the in_vim9script() check here.

src/fileio.c
```diff
@@ -3878,10 +3872,36 @@
     /*
      * Rename() failed, try copying the file.
      */
-    perm = mch_getperm(from);
+    ret = copyfile(from, to);
+    if (ret != 0)
+	return ret;
+
+    mch_remove(from);
```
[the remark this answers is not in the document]
Shougo: Well.  How to fix the failure?

Shougo:
Almost tests are passed.  But I don't know how to fix the CodeQL error.
The original routine is already exists...

tani:
I have read todo.txt.

> 	filecopy(from, to)	Copy a file

As TODO author intended to copy single file ("a file"), not directory.
Though it is convenient to deal with multiple file with single file,
to follow the todo.txt, would you mind to implement `dircopy` ?

Shougo:
I don't want to create both `filecopy()` and `dircopy()`.

tani:
Could I ask you the reason? The naming consistency is important in my sense.

Shougo:
The functions like `getftype()` works for both file and directory.

tani:
Thank you for the explanation. When we use getftype, we have no way to know that the string represents the directory or file in advance. Thus, the function is need to work with them. However, this function filecopy is not. We know their category (file/directory) in advance.

Shougo:
I will follow Bram's decision.

src/fileio.c
```diff
@@ -3878,6 +3872,61 @@ vim_rename(char_u *from, char_u *to)
...
+    ret = mch_lstat((char *)from, &st);
+    if (ret >= 0 && S_ISLNK(st.st_mode))
+    {
+	len = readlink((char *)from, linkbuf, MAXPATHL);
+	if (len > 0)
+	{
+	    linkbuf[len] = NUL;
+
+	    // Create link
+	    ret = symlink(linkbuf, (char *)to);
```
yegappan, line 3947: Can you add a test for copying a symbolic link in Unix-like systems?  I don't know whether there is a equivalent test on MS-Windows.

src/fileio.c
```diff
@@ -5082,6 +5130,86 @@ delete_recursive(char_u *name)
...
+	    return -1;
+	}
+
+	for (i = 0; i < ga.ga_len; ++i)
+	{
+	    // Copy {from}/{file} to {to}/{file} directory
+	    vim_snprintf((char *)sbuf, MAXPATHL, "%s/%s", exp,
+		    ((char_u **)ga.ga_data)[i]);
+	    vim_snprintf((char *)dbuf, MAXPATHL, "%s/%s", to,
+		    ((char_u **)ga.ga_data)[i]);
```
yegappan: It looks like many lines this function are not covered by test.  Can you add tests to cover all the missed lines?

yegappan (changes requested):

brammool:
Keep in mind that the entry in the todo list was just something that was requested quite a while ago.  There was no discussion about the details.  So don't base the name or functionality on that short note in the todo list.

There are many ways to copy a file/directory.  Look at the man page for "cp".  Especially handling directories, links, existing files, etc.  Many choices to be made.  I don't think we will want to add more than the basic functionality, just copying a file.  Other functions, such as getftype() can be used to decide what to do before calling filecopy().  Handling a directory, whether it should be copied recursively or not, can be done by the plugin.  This should not add much overhead.  Do watch out for circular references, that can be quite tricky...

To stay on the safe side it might be better to not overwrite an existing file, unless a flag is passed to do this anyway.  That might be the only option we need for now.

src/fileio.c
Shougo: It is already added in `Test_copy_symbolic_link()`.

Shougo:
>  I don't think we will want to add more than the basic functionality, just copying a file. Other functions, such as getftype() can be used to decide what to do before calling filecopy(). Handling a directory, whether it should be copied recursively or not, can be done by the plugin. This should not add much overhead. Do watch out for circular references, that can be quite tricky...

Hm..  OK.  I can just add the file copy function.

> To stay on the safe side it might be better to not overwrite an existing file, unless a flag is passed to do this anyway. That might be the only option we need for now.

Well, I can add the flag.  But it is needed?  `rename()` already overwrites the existing file.
And plugin can check the file exists.

brammool:
> >  I don't think we will want to add more than the basic
> >  functionality, just copying a file. Other functions, such as
> >  getftype() can be used to decide what to do before calling
> >  filecopy(). Handling a directory, whether it should be copied
> >  recursively or not, can be done by the plugin. This should not add
> >  much overhead. Do watch out for circular references, that can be
> >  quite tricky...
> 
> Hm..  OK.  I can just add the file copy function.
> 
> > To stay on the safe side it might be better to not overwrite an
> > existing file, unless a flag is passed to do this anyway. That might
> > be the only option we need for now.
> 
> Well, I can add the flag.  But it is needed?  `rename()` already
> overwrites the existing file.
> And plugin can check the file exists.

That involves a race condition.  In hindsight rename() should also have
this behavior, but now it's too late to change.

It may seem inconsistent to have a flag for overwriting on filecopy()
and not on rename(), but it can be useful.  If you are careful you can
use filecopy() without the flag, and when it succeeds delete the
original.  Then you basically have rename() with overwrite protection.
We can mention this in the help.

A detail is that filecopy() may fail for several reasons besides the
file already existing.  If it's not too complicated, the return value
can indicate the reason for why it failed.

-- 
You have heard the saying that if you put a thousand monkeys in a room with a
thousand typewriters and waited long enough, eventually you would have a room
full of dead monkeys.

src/fileio.c
```diff
@@ -3878,6 +3871,61 @@ vim_rename(char_u *from, char_u *to)
...
+	len = readlink((char *)from, linkbuf, MAXPATHL);
+	if (len > 0)
+	{
+	    linkbuf[len] = NUL;
+
+	    // Create link
+	    ret = symlink(linkbuf, (char *)to);
+	}
+
+	return ret == 0 ? ret : -1;
```
yegappan: The function currently returns either 0 or -1.  In this case, is it better to change this function to return a Boolean value (TRUE or FALSE)?

runtime/doc/builtin.txt
```diff
@@ -2658,6 +2659,16 @@ feedkeys({string} [, {mode}])				*feedkeys()*
 		Can also be used as a |method|: >
 			GetInput()->feedkeys()
 
+filecopy({from}, {to})					*filecopy()*
+		Copy the file by the name {from} to the name {to}.  The
+		result is a Number, which is 0 if the file was copied
+		successfully, and non-zero when the copy failed.
```
yegappan, line 2762: Should this function return a Boolean value (v:true or v:false)?

src/testdir/test_filecopy.vim
```diff
@@ -0,0 +1,68 @@
...
+
+  " Can't copy into a non-existing directory.
+  call assert_notequal(0, filecopy('Xfilecopy', 'Xdoesnotexist/Xfilecopy'))
+
+  " Can't copy a non-existing file.
+  call assert_notequal(0, filecopy('Xdoesnotexist', 'Xfilecopy2'))
+  call assert_equal('', glob('Xfilecopy2'))
+
+  " Can't copy to en empty file name.
+  call assert_notequal(0, filecopy('Xfilecopy', ''))
```
yegappan, line 66: If the filecopy() function is changed to return a Boolean value, then all the assert_equal() and assert_notequal() calls in these tests can be changed to assert_true() and assert_false().

src/fileio.c
Shougo: No.  Because,

> A detail is that filecopy() may fail for several reasons besides the
file already existing.  If it's not too complicated, the return value
can indicate the reason for why it failed.

Shougo:
> That involves a race condition.  In hindsight rename() should also have
this behavior, but now it's too late to change.

I can add the check in `filecopy()`.  But it cannot fix the race condition completely.

Shougo:
I have added `{flags}`.

src/fileio.c
yegappan: The current implementation of the function returns only -1 or 0.  This should be modified to return different values based on the error condition.

brammool:
> > That involves a race condition.  In hindsight rename() should also have
> > this behavior, but now it's too late to change.
> 
> I can add the check in `filecopy()`.  But it cannot fix the race
> condition completely.

On Unix you can use an open() call with an exclusive flag.  Then it
either fails if the file already existed, or it creates the file and you
can write to it.  That avoids the race condition.

-- 
7. You finally do take that vacation, but only after buying a USB modem
   and a laptop.

Shougo:
Hm...  I should remove the overwrite flag and it should fail if the dest exists.
It is smarter.

chrisbra:
what is the status of this one please? Can you please rebase?

chrisbra:
@Shougo are you still working on this?

Shougo:
I will rebase it if you needed.

chrisbra:
yeah, if you still think this is useful, let's update the PR and I'll have a closer look. I am also wondering whether we should call it `copy()` or `cp()` instead?

src/fileio.c
```diff
@@ -3878,6 +3872,61 @@
...
      */
+    ret = copyfile(from, to);
+    if (ret != 0)
+	return ret;
+
+    /*
+     * Remove copied original file
+     */
+    if (mch_stat((char *)from, &st) >= 0)
+	mch_remove(from);
```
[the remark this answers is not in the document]
Shougo: Well, the warning can be solved?

Shougo:
> yeah, if you still think this is useful, let's update the PR and I'll have a closer look. I am also wondering whether we should call it copy() or cp() instead?

`copy()` is already added in builtin function.  I think `cp()` will be confusing.
`filecopy()` is the name on `todo.txt`.

src/fileio.c
chrisbra: I think it should return OK or FAIL and then please have the caller in `fileio.c` also check for `FAIL/OK`

runtime/doc/builtin.txt
```diff
@@ -2755,6 +2756,17 @@ feedkeys({string} [, {mode}])				*feedkeys()*
...
+		If {to} exists, it is failed.
```
chrisbra:
```suggestion
		If {to} exists, it will fail.
```

runtime/doc/builtin.txt
```diff
@@ -2755,6 +2756,17 @@ feedkeys({string} [, {mode}])				*feedkeys()*
 		Can also be used as a |method|: >
 			GetInput()->feedkeys()
 
+filecopy({from}, {to})					*filecopy()*
+		Copy the file by the name {from} to the name {to}.  The
```
chrisbra, line 2760: just files? or also directories? Ah, I get it from the test, that directory copies are not supported. This should be documented I think

src/filepath.c
```diff
@@ -2649,6 +2649,30 @@ f_browsedir(typval_T *argvars UNUSED, typval_T *rettv)
...
+f_filecopy(typval_T *argvars, typval_T *rettv)
+{
+    char_u	*from;
+    stat_T	st;
+
+    rettv->vval.v_number = -1;
+
+    if (check_restricted() || check_secure()
+	    ||check_for_string_arg(argvars, 0) == FAIL
+	    || check_for_string_arg(argvars, 1) == FAIL)
```
chrisbra: alignment

src/testdir/test_filecopy.vim
```diff
@@ -0,0 +1,70 @@
...
+  call delete('Xtestfile')
+endfunc
+
+func Test_copy_dir_to_dir()
+  call mkdir('Xcopydir1')
+  call writefile(['foo'], 'Xcopydir1/Xfilecopy')
+  call mkdir('Xcopydir2')
+
+  " Directory copy is not supported
+  call assert_notequal(0, filecopy('Xcopydir1', 'Xcopydir2'))
```
chrisbra, line 47: similar as in the previous functions:
```suggestion
  call assert_equal(-1, filecopy('Xcopydir1', 'Xcopydir2'))
```

src/testdir/test_filecopy.vim
```diff
@@ -0,0 +1,70 @@
...
+
+  call delete('Xcopydir2', 'rf')
+  call delete('Xcopydir1', 'rf')
+endfunc
+
+func Test_copy_fails()
+  call writefile(['foo'], 'Xfilecopy', 'D')
```
chrisbra, line 56:
```suggestion
    CheckUnix
  
    call writefile(['foo'], 'Xfilecopy', 'D')
```

chrisbra (changes requested):

runtime/doc/builtin.txt
Shougo: Yes.  It copies just files.  I will update the documentation.

src/filepath.c
Shougo: Fixed.

src/testdir/test_filecopy.vim
```diff
@@ -0,0 +1,72 @@
...
+  call assert_equal(-1, filecopy('Xcopydir1', 'Xcopydir2'))
+
+  call delete('Xcopydir2', 'rf')
+  call delete('Xcopydir1', 'rf')
+endfunc
+
+func Test_copy_fails()
+    CheckUnix
+  
+    call writefile(['foo'], 'Xfilecopy', 'D')
```
zeertzjq:
```suggestion
  CheckUnix

  call writefile(['foo'], 'Xfilecopy', 'D')
```

src/filepath.c
```diff
@@ -2649,6 +2649,30 @@ f_browsedir(typval_T *argvars UNUSED, typval_T *rettv)
...
+    void
+f_filecopy(typval_T *argvars, typval_T *rettv)
+{
+    char_u	*from;
+    stat_T	st;
+
+    rettv->vval.v_number = -1;
+
+    if (check_restricted() || check_secure()
+	||check_for_string_arg(argvars, 0) == FAIL
```
zeertzjq:
```suggestion
	|| check_for_string_arg(argvars, 0) == FAIL
```

Shougo:
The tests are passed.

PR #12354, opened 2023-05-07
Fix wrong redo of typed command containing control characters

zeertzjq:
Also add missing check for block_redo, although maybe it is always off
in do_pending_operator().

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -43,8 +43,8 @@ static buffheader_T recordbuff = {{NULL, {NUL}}, NULL, 0, 0};
 static int typeahead_char = 0;		// typeahead char that's not flushed
 
 /*
- * when block_redo is TRUE redo buffer will not be changed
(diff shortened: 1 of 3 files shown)


zeertzjq:
BTW, it is mentioned in source code that:
```c
	if (cap->oap->op_type != OP_NOP)
	{
	    /* Using ":" as a movement is characterwise exclusive. */
	    cap->oap->motion_type = MCHAR;
	    cap->oap->inclusive = FALSE;
	}
```
But this doesn't seem documented?

zeertzjq:
Oh, it's documented here:
```
You can use a ":" command for a motion.  For example "d:call FindEnd()".
But this can't be repeated with "." if the command is more than one line.
This can be repeated: >
	d:call search("f")<CR>
This cannot be repeated: >
	d:if 1<CR>
	   call search("f")<CR>
	endif<CR>
Note that when using ":" any motion becomes charwise exclusive.
```

PR #12356, opened 2023-05-07
Some error messages are not translated

zeertzjq:
Not sure about the two changes for internal error messages, as there are
other internal error messages without an error number that are not
translated.

### DIFF
--- a/src/change.c
+++ b/src/change.c
@@ -1269,7 +1269,7 @@ del_bytes(
     // If "count" is negative the caller must be doing something wrong.
     if (count < 1)
     {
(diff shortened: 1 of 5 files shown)


brammool:
> Not sure about the two changes for internal error messages, as there are
> other internal error messages without an error number that are not
> translated.

I don't think translating the internal errors is useful.  In the rare
case such a message appears, someone has to dig into the code to find
out what happened.  End users do not need to understand the message
itself.  Well, other than that there is something wrong with Vim.


-- 
11. You find yourself typing "com" after every period when using a word
    processor.com

PR #12363, opened 2023-05-09
Add typst filetype

GaetanLepage:
This adds the `typst` filetype for [typst](https://typst.app) documents (`.typ` extension)

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -470,7 +470,7 @@ enddef
 
 # Returns true if file content looks like LambdaProlog module
 def IsLProlog(): bool
-  # skip apparent comments and blank lines, what looks like 
+  # skip apparent comments and blank lines, what looks like
   # LambdaProlog comment may be RAPID header
   var l: number = nextnonblank(1)
   while l > 0 && l < line('$') && getline(l) =~ '^\s*%' # LambdaProlog comment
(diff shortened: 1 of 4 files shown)


zeertzjq:
This PR doesn't actually work, because `.typ` files are recognized as SQL in a previous autocommand.

GaetanLepage:
> This PR doesn't actually work, because `.typ` files are recognized as SQL in a previous autocommand.

Indeed ! I tried to copy what is being done for "conflicting filetypes".

brammool:
> This adds the `typst` filetype for [typst](https://typst.app)
> documents (`.typ` extension)

Please add a test for this in src/testdir/test_filetype.vim.
Also fix the expectance for "file.typ".


-- 
Q: What is the difference between open-source and commercial software?
A: If you have a problem with commercial software you can call a phone
   number and they will tell you it might be solved in a future version.
   For open-source software there isn't a phone number to call, but you
   get the solution within a day.

GaetanLepage:
I am not able to find any reference to the fact that the `.typ` extension can be used for SQL files.
But I guess that we cannot take the liberty to remove completely the `.typ` - `sql` association, can we ?

If so, I would modify the PR to directly associate `.typ` files to the `typst` filetype.

brammool:
> I am not able to find any reference to the fact that the `.typ`
> extension can be used for SQL files.
> But I guess that we cannot take the liberty to remove completely the
> `.typ` - `sql` association, can we ?
> 
> If so, I would modify the PR to directly associate `.typ` files to the
> `typst` filetype.

The comment mentions "SQL for Oracle Designer".  That's the only hint I
can find.  Perhaps someone knows if ".typ" files are actually in use and
deserved to be recognized as "sql"?

Otherwise, recognizing the filetype by the contents is the usual
approach.  Sql and typst are hopefully different enough to tell the
difference from the first few lines.


-- 
6. In the memo field of all your checks, write "for sexual favors".

GaetanLepage:
I was able to find documentation about those sql `.typ` files: https://docs.oracle.com/cd/A58617_01/server.804/a58234/ott.htm

I rewrote the detection code and the test.
Now it tries to find the typical content of an sql type definition file and, if it is not able to do so, it affects the `typst` filetype.

Issue #12218, opened 2023-04-02
Cursor is moved down to a wrong line when &foldmethod is 'diff'


rickhowe:
### Steps to reproduce

This is related to [this problem](https://github.com/vim/vim/issues/12211) .
When the step #4 is:

4. `:set foldmethod=diff`

The cursor is jumped from line 3 to line 6.
This problem happens if &fooldmethod is other than 'manual'.

### Expected behaviour

Cursor should be moved down to the next line.

### Version of Vim

9.0.0807 or later (no problem on 9.0.0806)

### Environment

Windows 10

PR #12364, opened 2023-05-09
Fix problem with moving cursor below botline

luukvbaal:
Problem:   Lines are skipped when cursor is moved below botline when
           'smoothscroll' is not set. Unused code.
Solution:  Only use "scrolled" amount to scroll up with if 'smoothscroll'
           is set. (issue #12218) Remove unused code.


src/move.c
```diff
@@ -1752,22 +1752,6 @@ scrolldown(
     }
 }
 
-/*
- * Return TRUE if scrollup() will scroll by screen line rather than text line.
- */
-    static int
-scrolling_screenlines(int byfold UNUSED)
```
luukvbaal, line 1759: It seems this condition is actually only valid in `scrollup()`, not in `scroll_cursor_bot()` where it is also used. Hence removed the helper function.

src/move.c
```diff
@@ -2725,22 +2709,12 @@ scroll_cursor_bot(int min_scroll, int set_topbot)
...
-    else
+    else if (line_count > 0)
     {
-	// With 'smoothscroll' scroll at least the height of the cursor line,
-	// unless it would move the cursor.
-	if (curwin->w_p_wrap && curwin->w_p_sms && line_count < min_scrolled
-		&& (curwin->w_cursor.lnum < curwin->w_topline
-		    || (curwin->w_virtcol - curwin->w_skipcol >=
-					  curwin->w_width - curwin_col_off())))
-	    line_count = min_scrolled;
```
luukvbaal, line 2736: Unused since if 'smoothscroll' and 'wrap' are set, we are scrolling by "scrolled" lines, not "line_count". The problem this was solving also seems to be prevented elsewhere, as attested by the still passing tests that were added in https://github.com/vim/vim/commit/9bab7a024393200bb2c03b3abddfda86436990a7.

src/move.c
```diff
@@ -2725,22 +2709,12 @@ scroll_cursor_bot(int min_scroll, int set_topbot)
...
-	if (line_count > 0)
-	{
-	    if (scrolling_screenlines(TRUE))
-		scrollup(scrolled, TRUE);  // TODO
-	    else
-		scrollup(line_count, TRUE);
-	}
+	if (do_sms)
```
luukvbaal, line 2714: Checking for 'wrap' and 'smoothscroll' here instead of `scrolling_screenlines()` is the only functional change in this patch that prevents #12218. If you had different plans in mind (perhaps pertaining to the TODO note here?), feel free to ignore this patch or only use the test when you get around to fixing this.

PR #12367, opened 2023-05-10
Display and cursor corrupted when replacing a smooth-scrolled line

luukvbaal:
Problem:    When the topline is replaced while it is smooth-scrolled,
            "w_skipcol" is not adjusted and the cursor becomes invalid.
Solution:   Reset "w_skipcol" when the topline is replaced.


zeertzjq:
This issue is also reproducible when using if_pyth.

With Vim script:
```vim
40vsplit
rightbelow 4new
call setline(1, ['aaa', 'bbb', 'c'->repeat(548), 'ddd'])
set smoothscroll scrolloff=0
normal! G
redraw
call setline(3, 'c'->repeat(188))
```

With if_pyth:
```vim
40vsplit
rightbelow 4new
call setline(1, ['aaa', 'bbb', 'c'->repeat(548), 'ddd'])
set smoothscroll scrolloff=0
normal! G
redraw
python3 vim.current.buffer[2] = 'c' * 188
```

So maybe a more general solution is needed?

luukvbaal:
Maybe we should save the line length of topline and reset skipcol in `update_topline()` if it has changed. Or maybe it is necessary in `check_cursor_moved()` although that might be too expensive.

luukvbaal:
Or will adding this to each module that replaces buffer lines be preferable? Are there others?

luukvbaal:
I guess `changed_lines()` is common to any modules that change buffer lines. I'll try moving it there.

luukvbaal:
Should we use `win_linetabsize()` instead of `STRLEN()`?

src/change.c
```diff
@@ -560,6 +560,12 @@ changed_common(
 	    if (!redraw_not_allowed && wp->w_redr_type < UPD_VALID)
 		wp->w_redr_type = UPD_VALID;
 
+	    // Reset "w_skipcol" if the topline length is now smaller.
+	    if (wp->w_topline >= lnum && wp->w_topline < lnume
```
zeertzjq: I think this save some calls
```suggestion
	    if (wp->w_skipcol > 0
	    	    && wp->w_topline >= lnum && wp->w_topline < lnume
```

luukvbaal:
IDK if this the right approach to fixing this problem but it does seem to work properly.
Not sure what's going on with the test failures either, it passed here: https://github.com/luukvbaal/vim/actions/runs/4939090814.

luukvbaal:
>> Instead of resetting w_skipcol to zero a "smoother" solution would be to use the largest value that works. But it probably only makes a difference in very few situations.

> Hmm right. I guess that means setting w_skipcol to the largest integer multiple of the window width (accounted for col offset) that is smaller than the linetabsize.

Though I think even that would not necessarily be the "smoothest". That would probably be setting w_skipcol to the value that ensures the same amount of lines of the topline are visible as before the change. That is not necessarily the same as setting it to the largest value that works. Not sure that is even feasible without storing the number of visible skipcol lines.

PR #12373, opened 2023-05-10
Replace `signal()` with `mch_signal()`

ichizok:
Add `mch_signal()`; which is the wrapper of `sigaction()` or `signal()` (`sigset()`).

### Motivation:

Current glibc marks `sigset()` as a deprecated symbol, so should switch to `sigaction()` if available.
```
os_unix.c: In function ‘sig_winch’:
os_unix.c:872:5: warning: ‘sigset’ is deprecated: Use the signal and sigprocmask functions instead [-Wdeprecated-declarations]
  872 |     signal(SIGWINCH, (void (*)(int))sig_winch);
      |     ^~~~~~
In file included from /usr/include/x86_64-linux-gnu/sys/param.h:28,
                 from os_unix.h:60,
                 from vim.h:244,
                 from os_unix.c:21:
/usr/include/signal.h:367:23: note: declared here
  367 | extern __sighandler_t sigset (int __sig, __sighandler_t __disp) __THROW
      |                       ^~~~~~
```
`mch_signal()` imitates the behavior of `sigset()` by using `sigaction()`.

### DIFF
--- a/src/gui_haiku.cc
+++ b/src/gui_haiku.cc
@@ -785,8 +785,8 @@ VimApp::ReadyToRun()
      * Apparently signals are inherited by the created thread -
      * disable the most annoying ones.
      */
-    signal(SIGINT, SIG_IGN);
-    signal(SIGQUIT, SIG_IGN);
+    mch_signal(SIGINT, SIG_IGN);
+    mch_signal(SIGQUIT, SIG_IGN);
 }
 
     void
@@ -1067,8 +1067,8 @@ VimFormView::AllAttached()
      * Apparently signals are inherited by the created thread -
      * disable the most annoying ones.
      */
-    signal(SIGINT, SIG_IGN);
-    signal(SIGQUIT, SIG_IGN);
+    mch_signal(SIGINT, SIG_IGN);
+    mch_signal(SIGQUIT, SIG_IGN);
 
     if (menuBar && textArea) {
 	/*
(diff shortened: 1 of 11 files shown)


brammool:
> Add `mch_signal()`; which is the wrapper of `sigaction()` or
> `signal()` (`sigset()`).
> 
> ### Motivation:
> 
> Current glibc marks `sigset()` as a deprecated symbol, so should
> switch to `sigaction()` if available.
> `mch_signal()` imitates the behavior of `sigset()` by using `sigaction()`.

Can we move the #define from src/os_win32.h to src/vim.h?  Then it works
for all non-Unix systems.  Probably like this:

#if (defined(UNIX) || defined(VMS)) \
	&& (!defined(MACOS_X) || defined(HAVE_CONFIG_H))
# include "os_unix.h"	    // bring lots of system header files
#else
# define mch_signal(signum, sighandler) signal(signum, sighandler)
#endif

The #define for signal() in os_unixx.h becomes unused and should be
removed, right?


-- 
8. Don't use any punctuation marks.

brammool:
I do not see a change to os_unixx.h, is there a reason you want to keep the #define for "signal" there?

ichizok:
> I do not see a change to os_unixx.h, is there a reason you want to keep the #define for "signal" there?

I hesitated how to handle `sigset()`, but the last commit makes the selection of `sigaction()` `sigset()` `signal()` in `mch_signal()`.

PR #12382, opened 2023-05-12
Fix example of prompt_setcallback() to work without errors

zeertzjq:


brammool:
I'll include it and update the comments a bit.  It's still basic, perhaps there is a more useful (but still short) example we can put here.

PR #12390, opened 2023-05-12
Correctly check if the linux-modules-extra package exists

chrisbra:
commit e741f039cf46ccfbdaaa274618810787129d152b re-enabled the sound-dummy modules in the CI.

Unfortunately, the commit does not correctly check whether the linux-modules-extra- package exist, and so the CI throws an error
```
/home/runner/work/_temp/8c235aad-f1bd-4bdc-adcc-308dd3c8d320.sh: line 2: ]: command not found
```
e.g. here: https://github.com/vim/vim/actions/runs/4961614845/jobs/8878622463#step:8:30

So instead, properly check, whether apt-cache shows is able to show the package description. We do not need the shell test builtin

  if [ ... ]; then ... fi

because `if` can directly check the return value of the given command, which makes the test slightly easier to understand.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,8 +180,8 @@ jobs:
           DEST_DIR: ${{ env.TMPDIR }}/linux-modules-extra-${{ env.LINUX_VERSION }}
         run: |
           cd /lib/modules/${{ env.LINUX_VERSION }}
-          if [ sudo apt-get install -d -y linux-modules-extra-${{ env.LINUX_VERSION }} 2>&1 | $(grep "Unable to locate package ") ]; then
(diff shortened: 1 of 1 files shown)


pheiduck:
@chrisbra the if condition seams to be always true. https://packages.ubuntu.com/focal/linux-modules-extra-5.15.0-1037-azure

My intention was to check if the package is able to download or not, if not it should not exit with 1 and let the CI run without that package.

chrisbra:
Yeah, I think the condition is inverted currently. I noticed it yesterday and wanted to fix it when creating the PR and then forgot about it (it was too late and I was busy with the translations) 🙈 
Let me check your PR and see if we can further improve it

PR #12391, opened 2023-05-12
CI: use more recent gcc 12 version

pheiduck:
- gcc 12.3.0 (current latest 12 Version) is been installed via brew
- linked to gcc


brammool:
I thought brew was intended for MacOS, but here it is used for all systems.  Is this the right thing to do?

pheiduck:
I think we can use brew for the Linux CI as well, since it is used there as well and thus we can use more up-to-date packages.

Shane-XB-Qian:
https://github.com/vim/vim/pull/12405

PR #12392, opened 2023-05-12
updated German translations

chrisbra:


brammool:
I'll include it, thanks.

PR #12393, opened 2023-05-13
'smoothscroll' marker handling is incomplete

luukvbaal:
Problem:    'smoothscroll' marker overlap calculation is not always used
            and does not account for 'list' and 'listchars' "precedes".
Solution:   Add 'list' and 'listchars' "precedes" to the 'smoothscroll'
            marker calculation and use it in more places.

### DIFF
--- a/src/change.c
+++ b/src/change.c
@@ -568,8 +568,8 @@ changed_common(
 			    && wp->w_topline < lnume
 			    && win_linetabsize(wp, wp->w_topline,
 					ml_get(wp->w_topline), (colnr_T)MAXCOL)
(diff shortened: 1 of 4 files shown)


brammool:
I'll include it.  However, the test passes without the code changes, thus it doesn't cover the situation that was previously don wrong.

luukvbaal:
> However, the test passes without the code changes, thus it doesn't cover the situation that was previously don wrong.

Oh I should have checked that, it was intended to cover some of what was wrong atleast. Perhaps I can look at it again in a future patch.

PR #12395, opened 2023-05-13
Problems with going above topline in a line that doesn't fit the window

luukvbaal:
The changed test would have _looked_ wrong pre 9.0.1551 (cursor hidden behind the precedes character) but did not actually affect the cursor position. Change test to one that does, and would have failed pre 9.0.1551.


luukvbaal:
I think there are already screendump tests that verify that the cursor is no longer placed on top of the smoothscroll marker. I think coverage is fine now that we are using the common function where we should.

luukvbaal:
Changed this PR to fix and outline some problems with moving the cursor in a long line above the top of the window (also including the small previous test change). This didn't work properly for "b" and "ge" with 'smoothscroll', adding `adjust_skipcol()` for those motions fixes that.

However, there are similar problems when 'smoothscroll' is disabled, 'cpo'+=n is set and 'scrolloff' is small. I'm not sure what should be done about this. Either fix the `extra == 1` algorithm in `curs_columns()` (w_skipcol is set there but incorrectly under the mentioned conditions), or perhaps use `adjust_skipcol()` even when 'smoothscroll' is not enabled? I added a FIXME test that accepts the wrong result.

Also; moved these calls outside of the loop as they didn't seem to belong there:
```diff
+++ b/src/move.c
@@ -1960,12 +1960,14 @@ adjust_skipcol(void)
            curwin->w_skipcol -= width2;
        else
            curwin->w_skipcol -= width1;
-       redraw_later(UPD_NOT_VALID);
        scrolled = TRUE;
-       validate_virtcol();
     }
     if (scrolled)
+    {
+       validate_virtcol();
+       redraw_later(UPD_NOT_VALID);
        return;  // don't scroll in the other direction now
+    }

     int col = curwin->w_virtcol - curwin->w_skipcol + scrolloff_cols;
     int row = 0;
```

brammool:
I'll include it.  However, the tests pass even without the code changes.  I suppose you could reproduce the problem you fixed, can you turn that into a test?

luukvbaal:
> However, the tests pass even without the code changes. I suppose you could reproduce the problem you fixed, can you turn that into a test?

Commenting out `adjust_skipcol()` in `bck_word()/bckend_word()` results in these failures on my end:
```
Found errors in Test_display_cursor_long_line():
Run 1, 19:26:20 - 19:26:22:
command line..script /home/luuk/dev/vim/src/testdir/runtest.vim[569]..function RunTheTest[52]..Test_display_cursor_long_line[23]..VerifyScreenDump line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_display_cursor_long_line_3.dump", "testdir/dumps/Test_display_cursor_long_line_3.dump"); difference in line 1: "|<+0#4040ff13#ffffff0@2|b+0#0000000&@2>b@68"
command line..script /home/luuk/dev/vim/src/testdir/runtest.vim[569]..function RunTheTest[52]..Test_display_cursor_long_line[26]..VerifyScreenDump line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_display_cursor_long_line_4.dump", "testdir/dumps/Test_display_cursor_long_line_4.dump"); difference in line 1: "|<+0#4040ff13#ffffff0@2|b+0#0000000&>b@70"
```

PR #12396, opened 2023-05-14
Remove outdated note in if_perl.txt

zeertzjq:
This works since patch 6.2.475.


brammool:
I'll include it, thanks.

PR #12400, opened 2023-05-15
win32: Fix that $PATH may contain duplicated executable path

### LINKED ISSUE #12372: 'set enc=utf-8(latin1)' adds one more vim's installed path to $path
### Steps to reproduce

1. > gvim --clean
2. > :set clipboard=unnamed
3. > :let @+ = $path
4. (paste)
5. > :set enc=utf-8
6. > :let @+ = $path
7. (paste)
8. > :set enc=latin1
9. > :let @+ = $path
10. (paste)

### Expected behaviour

no duplicates in $path

### Version of Vim

9.0.1536

### Environment

Windows 11 x64 22H2

### Logs and stack traces

_No response_

k-takata:
Closes #12372

Keep the old exe_path and remove it before adding the new exe_path.

Also check if $PATH ends with ';' before adding it.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -467,7 +467,8 @@ mch_get_exe_name(void)
 {
     // Maximum length of $PATH is more than MAXPATHL.  8191 is often mentioned
     // as the maximum length that works (plus a NUL byte).
-#define MAX_ENV_PATH_LEN 8192
+    // Added 5 for "PATH=".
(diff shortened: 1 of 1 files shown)


k-takata:
This can change the order of the $PATH.
I'll check if the order can be kept.

k-takata:
Updated.

* Use Wide functions to handle $PATH.
* Add exe_path to $PATH only when it is not included.

PR #12401, opened 2023-05-15
Update sr.po

eevan78:
This one adds new command-window message string translation.


brammool:
I'll include it, thanks.

Issue #12402, opened 2023-05-15
Termdebug add support for local variables and arguments


laburnumT:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
It is cumbersome to use termdebug if you want to keep an eye on the local variables.

(If it is related to runtime files, please check their header for where to discuss enhancements.)
This is not mentioned in runtime/pack/dist/opt/termdebug/plugin/termdebug.vim

**Describe the solution you'd like**
I'd like to have more of gdb's information exposed to the user akin to the experience in emacs when using gdb-many-windows.

**Describe alternatives you've considered**
While there is nothing that this would enable that was not possible before, so the alternative would be to not implement the functionality.

**Additional context**
I would be willing to implement some or all of these features, and have taken a preliminary stab at a buffer for the variables exposed using `-stack-list-variables`.

PR: https://github.com/vim/vim/pull/12403

laburnumT:
Thanks. I'm waiting for some more reactions to the pull request and the way
I've done it. If it looks fine, I'll put in a merge request for a registers and
a memory window

PR #10393, opened 2022-05-09
Termdebug: add Frame window

GitMensch:
fixing #9235

open

* (likely minor): fix the check of `:Frame` arg type
* for me, huge, but @brammool said "that's easy possible", so I hope it will be added with or shortly after the integration: Possibility to select in the Frames Window, by pressing ENTER on the line on wants to go to ("real frames" are enough for the first cut, handling constructed frames, which are seldom and can be inserted from frame-filters, can be added later)

PR #12403, opened 2023-05-15
Add support to view local and argument variables in termdebug

laburnumT:
This is a pull request tied to a feature request https://github.com/vim/vim/issues/12402.

This would add the `Var` command to spawn a window similar to `Asm` in which local variables and function arguments can viewed.


brammool:
Thanks for looking into this.  The help for the :Var command is still missing.  It would be good to explain what the window will show.  Does it update when returning from a function or when calling a function?

No need to update the tags file, this is done automatically.

laburnumT:
I have added the help text.

Would you like me to discard any changes to the runtime/doc/tags file?

brammool:
This is still marked as draft.  Do you plan to add more or does it need some more testing?

laburnumT:
I have reverted the tags file.

I will unmark this as a draft. I'm looking at adding some of the other functionality that could be helpful, like a memory explorer or a register window. But I will create separate commits for those.

brammool:
> I have reverted the tags file.
> 
> I will unmark this as a draft. I'm looking at adding some of the other
> functionality that could be helpful, like a memory explorer or a
> register window. But I will create separate commits for those.

I'll leave this open for a little while, waiting for users to try it out
and give feedback.


-- 
George:  "I just got a new set of golf clubs for my wife!"
  John:  "Great trade!"

chrisbra:
has someone used this? Is this working as expected?

Shane-XB-Qian:
no, I don't think so.
some value of local var didn't print exactly.

and should `slient` or ignore the echo when reading/opening that var buf.

--
shane.xb.qian

laburnumT:
> no, I don't think so. some value of local var didn't print exactly. and should `slient` or ignore the echo when reading/opening that var buf.
> […](#)
> -- shane.xb.qian

Do you remember the code and situation?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1371,59 @@ func s:GotoAsmwinOrCreateIt()
...
+    setlocal number
+    setlocal noswapfile
+    setlocal buftype=nofile
+    setlocal modifiable
+
+    let varbuf = bufnr('Termdebug-variables-listing')
+    if varbuf > 0
+      exe 'buffer' . varbuf
+    else
+      exe 'file Termdebug-variables-listing'
```
Shane-XB-Qian: looks here, perhaps better prefix with `silent`, this should be not exist when auto init to open it.
silent file Termdebug-variables-listing

Shane-XB-Qian:
and `-stack-list-variables` seems this would not print exactly some value of local var vs/but identifying it was a `complex struct` something, though by `p xxx` can print it.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
laburnumT: Is This something that needs to be changed for all windows then? Because I just followed the existing practice for the existing assembly window.

https://github.com/vim/vim/blob/6c313bbb043745a60a7ee30a78edb907260b764a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim#L1285-L1289

laburnumT:
> and `-stack-list-variables` seems this would not print exactly some value of local var vs/but identifying it was a `complex struct` something, though by `p xxx` can print it.

There was indeed a problem with this around strings, but for structs this is an issue with gdb.`p` and `-stack-list-variables --all-values` give you more magic than `-stack-list-variables --simple-values`. But they leave out the type. 

My opinion was that the types and then doing the rest with `p` was more relevant. Especially because most structs are passed around as pointers, which both versions of `-stack-list-variables` show.

Shane-XB-Qian:
I think no reason to use `exe` here, and yes, I think should modify (silent) this for `Asm` too.

and actually if this Var and Asm (etc?) both set to init auto open, I felt what was that `file` would be, just a feeling, you may can test it.

--
shane.xb.qian

laburnumT:
> I think no reason to use `exe` here, and yes, I think should modify (silent) this for `Asm` too. and actually if this Var and Asm (etc?) both set to init auto open, I felt what was that `file` would be, just a feeling, you may can test it.
> […](#)
> -- shane.xb.qian

I'm not sure I'm following what you're suggesting the alternative would be.

Shane-XB-Qian:
> I'm not sure I'm following what you're suggesting the alternative would be.

I meant if Var and Asm (etc) both set to init auto open (there is a config to toggle them to open auto or manually) that maybe a race of `file` cmd for them.

anyway, I suggest those all `exe file xxx` modify to `silent file xxx`, but anyway please try if there any others side-effects.

--
shane.xb.qian

laburnumT:
> > I'm not sure I'm following what you're suggesting the alternative would be.
> > I meant if Var and Asm (etc) both set to init auto open (there is a config to toggle them to open auto or manually) that maybe a race of `file` cmd for them. anyway, I suggest those all `exe file xxx` modify to `silent file xxx`, but anyway please try if there any others side-effects.
> > […](#)
> > -- shane.xb.qian

The auto open just ensures that the `Goto(Asm|Variables)winOrCreateIt` function is called from `StartDebug_internal`. Calling `:Var` or `:Asm` at that point is exactly the same as just calling it twice. It will transport you to that buffer.

Shane-XB-Qian:
that bufnr>0 confirm that, vs `file` is to modify current.

anyway, I guess a race, (and I am not nearing my PC for now,) but/though that maybe not impact the result. however, a `silent` should be more nice operation.

--
shane.xb.qian

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1371,59 @@ func s:GotoAsmwinOrCreateIt()
...
+    setlocal number
+    setlocal noswapfile
+    setlocal buftype=nofile
+    setlocal modifiable
+
+    let varbuf = bufnr('Termdebug-variables-listing')
+    if varbuf > 0
+      exe 'buffer' . varbuf
+    else
+      silent exe 'file Termdebug-variables-listing'
```
Shane-XB-Qian: silent file Termdebug-variables-listing

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1285,7 +1350,7 @@ func s:GotoAsmwinOrCreateIt()
     if asmbuf > 0
       exe 'buffer' . asmbuf
     else
-      exe 'file Termdebug-asm-listing'
+      silent exe 'file Termdebug-asm-listing'
```
Shane-XB-Qian: silent file Termdebug-asm-listing

Shane-XB-Qian:
1. should be `silent file Termdebug-variables-listing` and `silent file Termdebug-asm-listing` without (no req) `exe` here.
2. seems useless only type but no value exactly was printed.
image
3. `Error detected while processing function <SNR>80_CommOutput[25]..<SNR>80_HandleError: line    8:` err was judged by `msg =~ '^\^error,msg='` looks was not accurate, seems some log would be printed prefix with `error` but it was just a log, specially if print via `echoerr substitute(msgVal, '\\"', '"', 'g')`, it looks like vim had some err.
4. once quit from gdb window (by `q`), it would auto close `program` window (default auto open) with `gdb` window, howerver this `Var` window would not be closed with it, is it good to keep the same?

laburnumT:
1. Done
2. Again that's a thing with gdb. It works for simple values only.
   Theoretically a double pass could be done. `--simple-values` to get the
   types and `--all-values` for the "complex" values. But I thought that wasn't
   a great idea
3. Fixed
4. I was on the fence about that. The main reason I chose not to do that
   initially was that the assembly window did not implement it either. But now
   that we're changing that too, I have implemented that both windows get
   closed on `Termdebug` exit.

Shane-XB-Qian:
> 2\. Theoretically a double pass could be done. `--simple-values` to get the
>         types and `--all-values` for the "complex" values. But I thought that wasn't
>         a great idea

but Var window supposed to check local var's value, that's its value, if only type, seems useless, specially perhaps there was not much simple-values cases, but most were complex var.

Shane-XB-Qian:
actually i verified with golang, it's mostly all were complex var.
to type, since there is a source code window, normally i knew it's type, so value perhaps was much more valuable.

---------
> 3\. Fixed

this item seems not fixed yet.

Shane-XB-Qian:
2. or let it be, to c/c++, perhaps you're right, or if you knew how to make it be workable for golang as well.
3. this item, an easy workaround i think is changing `echoerr` to 'echohl error | something... | echohl none'.

chrisbra:
so, is this ready to be included? Has anybody tried it out?

yegappan:
I have tried this out for a C program and it seems to work for me.

Shane-XB-Qian:
> 2. or let it be, to c/c++, perhaps you're right, or if you knew how to make it be workable for golang as well.

better be making golang working as well (maybe `--all-values` better, or not sure if it worked for golang),
or let it be, your call.

> 3. this item, an easy workaround i think is changing `echoerr` to 'echohl error | something... | echohl none'.

i opened PR #12856 for this item.

------
5. some broken indent
PR  #12857 for this item.

laburnumT:
I've been meaning to look at some none c/c++ languages, but haven't gotten around to it yet.

I'm going to take a look tomorrow, but I don't think I'd change it from `simple-values`

I'll also look at the hybrid approach, but it gives me bad vibes.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -822,6 +846,59 @@ func s:HandleDisasmMsg(msg)
...
+  let valueIdx = matchstrpos(a:varinfo, ',value="\(.*\)"}')
+  if valueIdx[1] == -1
+    let dict['value'] = 'Complex value'
+  else
+    let dict['value'] = a:varinfo[valueIdx[1] + 8 : valueIdx[2] - 3]
+  endif
+  return dict
+endfunc
+
+func s:SpacesString(num)
```
yegappan: This function can be replaced with `repeat(' ', num)`

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -206,6 +210,12 @@ func s:StartDebug_internal(dict)
     call win_gotoid(curwinid)
   endif
 
+  if s:GetVariablesWindow()
+    let curwinid = win_getid(winnr())
```
yegappan: To get the id of the current window, win_getid() can be used without any arguments.  No need to call winnr() here.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -822,6 +846,59 @@ func s:HandleDisasmMsg(msg)
...
+  let ret = ''
+  while i < a:num
+    let ret = ret . ' '
+    let i += 1
+  endwhile
+  return ret
+endfunc
+
+func s:HandleVariablesMsg(msg)
+  let curwinid = win_getid(winnr())
```
yegappan: The `winnr()` argument can be removed.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -822,6 +846,59 @@ func s:HandleDisasmMsg(msg)
...
+  if win_gotoid(s:varwin)
+
+    silent normal! gg0"_dG
```
yegappan: This can be replaced with the `silent! %delete _` command.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -822,6 +846,59 @@ func s:HandleDisasmMsg(msg)
...
+    let spaceBuffer = 20
+    call setline(1, 'Type' .
+	  \ s:SpacesString(16) .
+	  \ 'Name' .
+	  \ s:SpacesString(16) .
+	  \ 'Value')
```
yegappan, line 874: Use double dots (..) instead of a single dot (.) for string concatenation.  It will be more readable, if you use string interpolation:  `$'Type{repeat(' ', 16)}Name{repeat(' ', 16)}Value`

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1281,11 +1362,11 @@ func s:GotoAsmwinOrCreateIt()
     setlocal buftype=nofile
     setlocal modifiable
 
-    let asmbuf = bufnr('Termdebug-asm-listing')
-    if asmbuf > 0
-      exe 'buffer' . asmbuf
+    if s:asmbuf > 0
+      exe 'buffer' . s:asmbuf
```
yegappan, line 1358: Use double dots for string concatenation and use a space after "buffer".

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1387,59 @@ func s:GotoAsmwinOrCreateIt()
...
+
+func s:GotoVariableswinOrCreateIt()
+  if !win_gotoid(s:varwin)
+    if win_gotoid(s:sourcewin)
+      exe 'rightbelow new'
+    else
+      exe 'new'
+    endif
+
+    let s:varwin = win_getid(winnr())
```
yegappan: winnr() can be removed here.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1387,59 @@ func s:GotoAsmwinOrCreateIt()
...
+
+    setlocal nowrap
+    setlocal number
```
yegappan: Why is the 'number' option set in this window?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1387,59 @@ func s:GotoAsmwinOrCreateIt()
...
+    setlocal noswapfile
+    setlocal buftype=nofile
+    setlocal modifiable
```
yegappan, line 1417: Other options typically cleared for a scratch window are: relativenumber, signcolumn and foldcolumn.  Also you can set the 'buhidden' option to 'wipe' to remove this buffer.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,6 +1387,59 @@ func s:GotoAsmwinOrCreateIt()
...
+
+    if s:varbuf > 0
+      exe 'buffer' . s:varbuf
```
yegappan, line 1420: Use double dots and add a space after "buffer".

yegappan (changes requested):

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
laburnumT: Single dots are used throughout most of the file for string concatenation. Should I change it in all the locations?

Shane-XB-Qian: it is a code style and vim9 may require double dots, vs for legacy vim script, as long as clear enough should be ok.
// IMO: unless refactor/rewrite to vim9 one day, perhaps better if kept its current style.
// or if you have time then change its all, but actually indent was not all same style there.

yegappan: We can address the single dot to double dots conversion in a separate PR.  You can leave that out of this PR.

yegappan (approved):

Shane-XB-Qian:
> I've been meaning to look at some none c/c++ languages, but haven't gotten around to it yet.
> I'm going to take a look tomorrow, but I don't think I'd change it from `simple-values`
> I'll also look at the hybrid approach, but it gives me bad vibes.

thanks, this is the key point, wish you make this work through,
or if really not possible or difficult, then let it be for now, thanks~

chrisbra:
Included as of 9f29621415146abc046471440515e9e34f3e57a1 

I suppose we can make further adjustments as needed.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
GitMensch: Oops, I've thought those to be a typo and dropped them in #10393 - I guess it would be good to revert that and instead use it in every place that was actually changed, right?

PR #12412, opened 2023-05-18
Use int64_t for nanosecond calculation

svardew:
In src/profiler.c, profie_setlimit() assign multiplied nanosecond to (long)fsec.
For 32-bit cpu (ac_cv_sizeof_long=4), this cause many testdir/test_FEATNAME.vim fails:
 runtest.vim -> RunTheTest() -> timer_start(30000,'TestTimeout') -> profile_setlimit(30000,tm)
 -> (long)fsec = 30*10^9  ( > 2^35). Overflowed fsec is negative, assume time has passed.
Same as MSWIN (LONGLONG, double), use int64_t for nanosecond.

### DIFF
--- a/src/profiler.c
+++ b/src/profiler.c
@@ -123,10 +123,10 @@ profile_setlimit(long msec, proftime_T *tm)
 	QueryPerformanceFrequency(&fr);
 	tm->QuadPart += (LONGLONG)((double)msec / 1000.0 * (double)fr.QuadPart);
 # else
-	int64_t	    fsec;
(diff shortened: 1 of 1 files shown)


svardew:
@brammool 
> Using int64_t is not portable, not all compilers support it.
> How about using varnumber_T instead?  It already has the #ifdefs to make
> it a 64 bit type whenever possible.

Yes I agree.
Changing use varnumber_T, my local pc build & test result fine.

Issue #12411, opened 2023-05-18
Cursor ends up in not (fully) visible lastline


luukvbaal:
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
winsize 40 8
call setline(1, ['Line' .. (' with some text'->repeat(7))]->repeat(7))
set smoothscroll scrolloff=0
norm G
```
3. Press `4<C-Y>G`
4. Note that the cursor is in the last line but it is not displayed as 'display' does not contain "lastline".
5. Type `:set display=lastline`(not necessary) and press `G4<C-Y>G` again.
6. Note that the cursor is in the last line again, which is displayed, but moving the cursor to the wrapped part of the line (press `$` or `l` repeatedly) will still not adjust the topline/skipcol and bring the cursor into the command line.

### Expected behaviour

Topline (and skipcol) should be adjusted in this case.
(Is it an intended goal of 'smoothscroll' to allow the cursor to be in a not fully displayed lastline (with "lastline" in 'display')?)

### Version of Vim

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled May 18 2023 12:30:36) Included patches: 1-1563

### Environment

Operating system: Arch Linux
Terminal: st
Value of $TERM: st-256color
Shell: zsh

brammool:
> ### Steps to reproduce
> 
> 1. Run `vim --clean`
> 2. Source the following file:
> ```vim
> winsize 40 8
> call setline(1, ['Line' .. (' with some text'->repeat(7))]->repeat(7))
> set smoothscroll scrolloff=0
> norm G
> ```
> 3. Press `4<C-E>G`
> 4. Note that the cursor is in the last line but it is not displayed as 'display' does not contain "lastline".

It would be logical to add "lastline" to 'display' when 'smoothscroll' is
set.  I'll add a note in the help.

> 5. Type `:set display=lastline` and press `4<C-E>G` again.
> 6. Note that the cursor is in the last line again, which is displayed, but moving the cursor to the wrapped part of the line (press `$` or `l` repeatedly) will still not adjust the topline/skipcol and bring the cursor into the command line.
> 
> ### Expected behaviour
> 
> Topline (and skipcol) should be adjusted in this case.
> (Is it an intended goal of 'smoothscroll' to allow the cursor to be in a not fully displayed lastline (with "lastline" in 'display')?)

'smoothscroll' and 'display' are independent.  It makes a lot of sense
[...]

Issue #12409, opened 2023-05-18
Display moves up and down with 'incsearch' and 'smoothscroll'


zeertzjq:
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
set incsearch
7split
50vnew
setlocal number scrolloff=0 smoothscroll
call setline(1, repeat([''], 20))
call setline(11, repeat('a', 100))
call setline(14, 'bbbb')
```
3. Type `/bbbb`, and display moves up and down on every `b` keystroke

### Expected behaviour

Display doesn't move up and down when the highlighted match's position is the same

### Version of Vim

9.0.1563

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

zeertzjq:
Bisected to [patch 9.0.1525](https://github.com/vim/vim/commit/3ce8c389155fc1257082cdb0cef7801b49f6aaf9)

luukvbaal:
So before patch 9.0.1525 the cursor was scrolled half-way but w_skipcol was immediately reset (thus not placing the cursor halfway). After the patch it is placed in the middle each time but the reset is alternated due to the new `curwin->w_skipcol == old_skipcol` condition on `reset_skipcol()` which is still incorrect apparently.

zeertzjq:
Yes this `curwin->w_skipcol == old_skipcol` looks unreliable.

luukvbaal:
TBH I think we don't need to reset skipcol at all anymore there. This results in failing tests which is why I added this condition instead last time. I think we can adjusts the tests instead.

brammool:
> TBH I think we don't need to reset skipcol at all anymore there. This
> results in failing tests which is why I added this condition instead
> last time. I think we can adjusts the tests instead.

Are you going to make a PR for this?

-- 
The 50-50-90 rule: Anytime you have a 50-50 chance of getting
something right, there's a 90% probability you'll get it wrong.

 /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net   \\\
///                                                                      \\\
\\\        sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ ///
 \\\            help me help AIDS victims -- http://ICCF-Holland.org    ///

PR #12415, opened 2023-05-18
Cursor row is not adjusted for skipcol with 'smoothscroll'

luukvbaal:
Problem:    Cursor row calculation does not account for skipcol when
            'diff' is not set.
Solution:   Account for skipcol when it is non-zero.

I noticed the `<C-E>` step in the affected test behaved unexpectedly since https://github.com/vim/vim/commit/0222c2d103ad9298bec4dc8864cd80b4e7559db1. Besides the skipcol reset, we also lost a `redraw_later()` with that patch. It seems this is still necessary.


luukvbaal:
I added a regression test for https://github.com/vim/vim/issues/12409.

src/move.c
```diff
@@ -893,7 +895,7 @@ curs_rows(win_T *wp)
 	}
 	if (valid
 #ifdef FEAT_DIFF
-		&& (lnum != wp->w_topline || !wp->w_p_diff)
+		&& (lnum != wp->w_topline || wp->w_skipcol == 0)
```
luukvbaal: I think we need to execute the `else` branch when skipcol is non-zero, not just when 'diff' is set. This affects two screendumps which currently contained invalid cursor positions, which I updated. This in combination with the added redraw_later() also fixes #12411.

zeertzjq: This `#ifdef FEAT_DIFF` should be removed then.

zeertzjq: And the `else` branch is executed when 'diff' is not set.

zeertzjq: Oops, I read it wrong.

luukvbaal:
Screendumps smooth_wrap_{2,4} are now actually back to how they were before https://github.com/vim/vim/commit/0222c2d103ad9298bec4dc8864cd80b4e7559db1.
Thus I added a testcase from #12411 that failed even before https://github.com/vim/vim/commit/0222c2d103ad9298bec4dc8864cd80b4e7559db1, but passes now.

src/move.c
luukvbaal: Hmm no I think we need to check for skipcol besides checking for 'diff' filler lines, not instead of. Updated.

PR #12423, opened 2023-05-22
Added new structurizr keywords introduced sinced 2022-02.

venthur:
Added three keywords to the structurizr syntax file, that have been introduced since the last update in 2022.

https://github.com/structurizr/dsl/blob/master/docs/changelog.md


brammool:
Thanks, I'll include it.

PR #12428, opened 2023-05-23
SpellCap highlight is not always shown

### LINKED ISSUE #12420: SpellCap highlight not shown if blank line above is folded
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
call setline(1, ['hello', 'world', '', 'hello', 'world'])
set spell
```
3. The `hello` on lines 1 and 4 are highlighted
4. Run `1,3fold`
5. The `hello` on line 4 is no longer highlighted

### Expected behaviour

The `hello` on line 4 is still highlighted

### Version of Vim

9.0.1506

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

luukvbaal:
Problem:    The SpellCap highlight is not shown at the start of a line when redrawing the first row in a window, the top line with changed text in a window, or if the previous line above is folded.
Solution:   Check for sentence end in the line above.

Fix #12420


brammool:
> Fix #12420

This only checks the last line of the fold to be empty.  There are many
more possible situations, such as a fold with more than one empty line
(might not matter though), or the fold not ending a sentence.

I don't see a simple fix for this.  I would think that when a line is
displayed just below a fold, then 'spellcapcheck' needs to be used on
the last non-white text in the fold to find out if a sentence ended in
the fold.  Without a fold this is done while displaying a line, where
b_cap_prog is used in spell_check().  The code commented with "Check for
end of sentence" can be re-used.  There is also check_need_cap() but I'm
not sure if it can be used for this.

-- 
Time is money.  Especially if you make clocks.

luukvbaal:
Right I had not considered anything other than the case in the linked issue.
I guess I can try to make it work more extensively like you suggested, no promises though.

luukvbaal:
> There is also check_need_cap() but I'm
not sure if it can be used for this.

Introducing a `win_check_need_cap()` for this seems to work. This function allocates it's own buffer to concatenate two lines, though whereas in `win_line()` we already have the `nextline` buffer. Would it be required/worth it to make `win_check_need_cap()` re-use that buffer when called from `win_line()`?

src/testdir/test_spell.vim
```diff
@@ -999,13 +999,21 @@ func Test_spell_screendump_spellcap()
   call VerifyScreenDump(buf, 'Test_spell_3', {})
 
   " Deleting a full stop removes missing Cap in next line
-  call term_sendkeys(buf, "5Gddk$x")
+  call term_sendkeys(buf, "5Gdd\<C-L>k$x")
```
luukvbaal, line 1002:
```diff
@@ -999,13 +999,21 @@ func Test_spell_screendump_spellcap()
   call VerifyScreenDump(buf, 'Test_spell_3', {})

   " Deleting a full stop removes missing Cap in next line
-  call term_sendkeys(buf, "5Gddk$x")
+  call term_sendkeys(buf, "5Gdd\<C-L>k$x")
   call VerifyScreenDump(buf, 'Test_spell_4', {})
```
This test was ineffective because without forcing a redraw the SpellCap highlight is not there in the first place after "dd". This could/should be resolved as per discussion in https://github.com/vim/vim/issues/12420#issuecomment-1560184585. It may be worth noting that the SpellCap highlight already _is_ updated when 'conceallevel' > 0. I think we should also force redrawing the old cursorline when 'spell' is set to make sure that the spell highlighting is updated after these kinds of edits.

brammool:
> > There is also check_need_cap() but I'm not sure if it can be used for this.
> 
> Introducing a `win_check_need_cap()` for this seems to work. This
> function allocates it's own buffer to concatenate two lines, though
> whereas in `win_line()` we already have the `nextline` buffer. Would
> it be required/worth it to make `win_check_need_cap()` re-use that
> buffer when called from `win_line()`?

It looks like extra allocation doesn't always happen.  And the pattern
matching already adds overhead, avoiding the allocation won't help much.
I don't think we want to make it more complicated for a tiny performance
advantage.

BTW: there is no check if contact_str() returns NULL.  That happens when
out of memory.  Might as well return then.

-- 
66. You create a homepage with the impression to cure the afflicted...but
    your hidden agenda is to receive more e-mail.

luukvbaal:
Should the `win_T *` argument just be added to `check_need_cap()`? Keeping the `curwin` version might be unnecessary, there are only a few callers.

brammool:
> Should the `win_T *` argument just be added to `check_need_cap()`?
> Keeping the `curwin` version might be unnecessary, there are only a
> few callers.

I would just always pass the window argument.  Then it's only one
function, and it's clear that "curwin" is being used where it's calle.

-- 
70. ISDN lines are added to your house on a hourly basis

luukvbaal:
> I would just always pass the window argument. Then it's only one function, and it's clear that "curwin" is being used where it's calle.

Yes, that is what I meant. Updated.

luukvbaal:
I fixed and added a test for another case where SpellCap is not applied for the first row in a window.
Also added a few `<C-L>` to the tests to make sure that we are testing that it is least working properly when doing a full redraw. Removing the `<C-L>`'s also shows that it is not working properly yet when making edits.

luukvbaal:
> Removing the `<C-L>`'s also shows that it is not working properly yet when making edits.

The problem is not that the line is not being redrawn when doing these kinds of edits, but that we are not checking for sentence end. We need to also check for sentence end if `win_line()` is drawing the first row in a particular `win_update()`. Is there anyway to check that from `win_line()`? Checking that `nochange == false` fixes the issue but I guess this parameter can be true for multiple lines per `win_update()`, in which case it wouldn't be strictly correct. `win_redr_top/bot` are already reset at that point so those can't be checked either.

Alternatively we could add a global variable...

src/drawline.c
```diff
@@ -1314,11 +1314,18 @@ win_line(
...
 	    // word starting with capital in this line.  In line 1 always check
-	    // the first word.
-	    if (lnum != capcol_lnum)
-		cap_col = -1;
-	    if (lnum == 1)
+	    // the first word.  Also check for sentence end in the line above
+	    // when updating changed text, the first row in a window, or if the
+	    // previous line is folded.
+	    if (lnum == 1
+		    || ((!nochange || startrow == 0
```
luukvbaal: So yeah just to be clear these two conditions could/should be combined into one that checks that we are in the first call to `win_line()` in a particular `win_update()`.
That would be the correct condition for when to check for sentence end in the previous line, as opposed to for all lines with edited text and the first row in a window.

luukvbaal: This works I guess:
```diff
diff --git a/src/drawline.c b/src/drawline.c
index 72d1bfa5b..a8889840f 100644
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -1318,7 +1318,8 @@ win_line(
            // when updating changed text, the first row in a window, or if the
            // previous line is folded.
            if (lnum == 1
-                   || ((!nochange || startrow == 0
+                   || ((startrow == 0
+                       || (!nochange && lnum == wp->w_buffer->b_mod_top)
 #ifdef FEAT_FOLDING
                        || hasFoldingWin(wp, lnum - 1, NULL, NULL, TRUE, NULL)
 #endif
```

luukvbaal: Changing the `nochange` argument to just be `mod_top` probably makes more sense:
```diff
diff --git a/src/drawline.c b/src/drawline.c
index c08aaa40d..3791ece43 100644
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -1062,7 +1062,7 @@ win_line(
     linenr_T	lnum,
     int		startrow,
     int		endrow,
-    int		nochange UNUSED,	// not updating for changed text
+    int		mod_top UNUSED,	        // top line updated for changed text
     int		number_only)		// only update the number column
 {
     winlinevars_T	wlv;		// variables passed between functions
@@ -1316,10 +1316,9 @@ win_line(
 	    // word starting with capital in this line.  In line 1 always check
 	    // the first word.  Also check for sentence end in the line above
 	    // when updating the first row in a window, the top line with
-	    // changed text window, or if the previous line is folded.
+	    // changed text in a window, or if the previous line is folded.
 	    if (lnum == 1
-		    || ((startrow == 0
-			|| (!nochange && lnum == wp->w_buffer->b_mod_top)
+		    || ((startrow == 0 || mod_top == lnum
 #ifdef FEAT_FOLDING
 			|| hasFoldingWin(wp, lnum - 1, NULL, NULL, TRUE, NULL)
 #endif
@@ -2795,7 +2794,7 @@ win_line(
 			    p = prev_ptr;
 			cap_col -= (int)(prev_ptr - line);
 			len = spell_check(wp, p, &spell_hlf, &cap_col,
-								    nochange);
+								mod_top == 0);
 			word_end = v + len;
 
 			// In Insert mode only highlight a word that
diff --git a/src/drawscreen.c b/src/drawscreen.c
index a07c176d0..0912d7a15 100644
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -2487,8 +2487,7 @@ win_update(win_T *wp)
 #endif
 
 		// Display one line.
-		row = win_line(wp, lnum, srow, wp->w_height,
-							  mod_top == 0, FALSE);
+		row = win_line(wp, lnum, srow, wp->w_height, mod_top, FALSE);
 
 #ifdef FEAT_FOLDING
 		wp->w_lines[idx].wl_folded = FALSE;
@@ -2535,7 +2534,7 @@ win_update(win_T *wp)
 		    fold_line(wp, fold_count, &win_foldinfo, lnum, row);
 		else
 #endif
-		    (void)win_line(wp, lnum, srow, wp->w_height, TRUE, TRUE);
+		    (void)win_line(wp, lnum, srow, wp->w_height, mod_top, TRUE);
 	    }
 
 	    // This line does not need to be drawn, advance to the next one.
```
I'll push one more time, feel free to include the previous version if you disagree.

brammool:
> > Removing the `<C-L>`'s also shows that it is not working properly
> > yet when making edits.
> 
> The problem is not that the line is not being redrawn when doing these
> kinds of edits, but that we are not checking for sentence end. We need
> to also check for sentence end if `win_line()` is drawing the first row
> in a particular `win_update()`. Is there anyway to check that from
> `win_line()`? Checking that `nochange == false` fixes the issue but I
> guess this parameter can be true for multiple lines per `win_update()`,
> in which case it wouldn't be strictly correct. `win_redr_top/bot` are
> already reset at that point so those can't be checked either.
> 
> Alternatively we could add a global variable...

There are a few local variables in win_line() that are static.  These
are weird, because the values are not really global, but passed from one
call of the function to the next call.

It's probably better when these values are passed as an argument.  Then
the caller can clear them on the first call of a sequence.

In win_update() this is fairly simple.  And I don't see win_line()
called anywhere else.  Perhaps you can make a PR for this after I
include this one.  I suppose it might be possible to revert some of this
PR, it will be good to have CI check with the tests that are added now.

-- 
77. The phone company asks you to test drive their new PBX system

PR #12442, opened 2023-05-25
Require quickfix filename to always be valid

sugrob9000:
The cases where the file name is not valid are usually misparses of compiler output - e.g., thinking that "make: *** [Makefile" is a filename. Vim also already requires the file to exist with %O, %P, and %Q; this patch simply makes the check apply in all cases.

The behavior of opening a garbage buffer with a nonsensical name is annoying. Vim presumes the user will look for and fix errors in that file, but Vim already knows the buffer will be empty.

### DIFF
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -972,13 +972,7 @@ qf_parse_fmt_f(regmatch_T *rmp, int midx, qffields_T *fields, int prefix)
     expand_env(rmp->startp[midx], fields->namebuf, CMDBUFFSIZE);
     *rmp->endp[midx] = c;
 
-    // For separate filename patterns (%O, %P and %Q), the specified file
-    // should exist.
-    if (vim_strchr((char_u *)"OPQ", prefix) != NULL
-	    && mch_getperm(fields->namebuf) == -1)
-	return QF_FAIL;
-
-    return QF_OK;
+    return mch_getperm(fields->namebuf) == -1 ? QF_FAIL : QF_OK;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


brammool:
> The cases where the file name is not valid are usually misparses of
> compiler output - e.g., thinking that "make: *** [Makefile" is a
> filename. Vim also already requires the file to exist with %O, %P, and
> %Q; this patch simply makes the check apply in all cases.
> 
> The behavior of opening a garbage buffer with a nonsensical name is
> annoying. Vim presumes the user will look for and fix errors in that
> file, but Vim already knows the buffer will be empty.

I can think of several situations where the name is useful, even though
it does not appear to exist: wrong directory, generated file that was
deleted, etc.  Even though the file can't be opened directly, it may
very well be a hint for the user about what the problem is.  Especially
when working on the 'errorformat' option.

-- 
From "know your smileys":
 :----}  You lie like Pinocchio

Crowbarous:
In that case, perhaps an option would be useful. I initially made that exact patch for Neovim (boolean option, old behavior by default), but I was told:
a) that an option to toggle between two behaviors is rarely a good design choice, and
b) to check this in with you first.
Heeding the former point, I decided to start simple. (Also, with Vim, adding new options is more involved, since I'm unfamiliar with your codebase yet. Sorry!)
P.S. It's not like when quickfix is unable to parse a line, any information is lost - it's still visible in the output and in the quickfix window. But this may be an issue to people whose quickfix workflow relies on `:cnext` and `:cprev` only, less the window.

PR #12448, opened 2023-05-26
Code for spell checking is spread out and uses static variables

luukvbaal:
Problem: Code for spell checking is spread across multiple blocks in
win_line() and uses static variables that should be scoped
per window in win_update().
Solution: Group spell checking blocks in win_line() and introduce
variable struct used for spell checking that is passed around.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -1062,8 +1062,8 @@ win_line(
     linenr_T	lnum,
     int		startrow,
     int		endrow,
-    int		mod_top UNUSED,	        // top line updated for changed text
-    int		number_only)		// only update the number column
+    int		number_only,	// only update the number column
+    spellvars_T	*spv UNUSED)
 {
     winlinevars_T	wlv;		// variables passed between functions
 
@@ -1139,7 +1139,6 @@ win_line(
     int		reset_extra_attr = FALSE;
 #endif
 #ifdef FEAT_SPELL
-    int		has_spell = FALSE;	// this buffer has spell checking
     int		can_spell = FALSE;
 # define SPWORDLEN 150
     char_u	nextline[SPWORDLEN * 2];// text with start of the next line
@@ -1148,11 +1147,6 @@ win_line(
 					// starts
     int		spell_attr = 0;		// attributes desired by spelling
     int		word_end = 0;		// last byte with same spell_attr
-    static linenr_T  checked_lnum = 0;	// line number for "checked_col"
-    static int	checked_col = 0;	// column in "checked_lnum" up to which
-					// there are no spell errors
-    static int	cap_col = -1;		// column to check for Cap word
-    static linenr_T capcol_lnum = 0;	// line number where "cap_col" used
     int		cur_checked_col = 0;	// checked column for current line
 #endif
     int		extra_check = 0;	// has extra highlighting
@@ -1286,47 +1280,6 @@ win_line(
 	    extra_check = TRUE;
 	    get_term_attr = TRUE;
 	    wlv.win_attr = term_get_attr(wp, lnum, -1);
-	}
-#endif
-
-#ifdef FEAT_SPELL
-	if (spell_check_window(wp))
-	{
-	    // Prepare for spell checking.
-	    has_spell = TRUE;
-	    extra_check = TRUE;
-
-	    // Get the start of the next line, so that words that wrap to the
-	    // next line are found too: "et<line-break>al.".
-	    // Trick: skip a few chars for C/shell/Vim comments
-	    nextline[SPWORDLEN] = NUL;
-	    if (lnum < wp->w_buffer->b_ml.ml_line_count)
-	    {
-		line = ml_get_buf(wp->w_buffer, lnum + 1, FALSE);
-		spell_cat_line(nextline + SPWORDLEN, line, SPWORDLEN);
-	    }
-
-	    // When a word wrapped from the previous line the start of the
-	    // current line is valid.
-	    if (lnum == checked_lnum)
-		cur_checked_col = checked_col;
-	    checked_lnum = 0;
-
-	    // When there was a sentence end in the previous line may require a
-	    // word starting with capital in this line.  In line 1 always check
-	    // the first word.  Also check for sentence end in the line above
-	    // when updating the first row in a window, the top line with
-	    // changed text in a window, or if the previous line is folded.
-	    if (lnum == 1
-		    || ((startrow == 0 || mod_top == lnum
-#ifdef FEAT_FOLDING
-			|| hasFoldingWin(wp, lnum - 1, NULL, NULL, TRUE, NULL)
-#endif
-			) && check_need_cap(wp, lnum, 0)))
-		cap_col = 0;
-	    else if (lnum != capcol_lnum)
-		cap_col = -1;
-	    capcol_lnum = 0;
 	}
 #endif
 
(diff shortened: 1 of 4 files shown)


luukvbaal:
> There are a few local variables in win_line() that are static.  These
are weird, because the values are not really global, but passed from one
call of the function to the next call.

> It's probably better when these values are passed as an argument.  Then
the caller can clear them on the first call of a sequence.

> In win_update() this is fairly simple.  And I don't see win_line()
called anywhere else.  Perhaps you can make a PR for this after I
include this one.  I suppose it might be possible to revert some of this
PR, it will be good to have CI check with the tests that are added now.

I made this patch to address this. However, I'm not sure that it is a clear improvement, what do you think? Perhaps I can refine it somewhat or you can close this if it's not desired.

luukvbaal:
It should reduce some calls to `spell_check_window()` and `hasFoldingWin()` so there's that.

luukvbaal:
> and introduce variable struct used for spell checking that is passed around

I guess we don't use FEAT guarded structs in function definitions. Changed to separate variable pointers instead...

brammool:
> > and introduce variable struct used for spell checking that is passed around
> 
> I guess we don't use FEAT guarded structs in function definitions. Changed to
> separate variable pointers instead...

That is unfortunate, I would prefer passing a struct instead of several
arguments, which are then not used without the spell feature.  The structure
can be defined without that feature, so that the function prototype is correct.
That doesn't take any space after compiling without the feature defined.  The
struct would just not be used, pass a NULL for the argument.

There seem to be a few unrelated changes that might actually be wrong, such as
moving these two lines up:

-	    wp->w_lines[idx].wl_lnum = lnum;
-	    wp->w_lines[idx].wl_valid = TRUE;

Better avoid making unrelated changes in this PR, it can only make it difficult
to figure out why something changed unrelated to what this PR was made for.

-- 
86. E-mail Deficiency Depression (EDD) forces you to e-mail yourself.

luukvbaal:
> That is unfortunate, I would prefer passing a struct instead of several arguments, which are then not used without the spell feature. The structure can be defined without that feature, so that the function prototype is correct. That doesn't take any space after compiling without the feature defined. The struct would just not be used, pass a NULL for the argument.

Right, makes sense.

> There seem to be a few unrelated changes that might actually be wrong, such as moving these two lines up: - wp->w_lines[idx].wl_lnum = lnum; - wp->w_lines[idx].wl_valid = TRUE; Better avoid making unrelated changes in this PR, it can only make it difficult to figure out why something changed unrelated to what this PR was made for.

I did this because I needed to use `lnum + foldcount` 3 times in this block so I incremented `lnum` but that meant having to move up those two lines up while cleaning up an `#ifdef FEAT_FOLD` block below.
```diff
@@ -2450,10 +2469,19 @@ win_update(win_T *wp)
                fold_line(wp, fold_count, &win_foldinfo, lnum, row);
                ++row;
                --fold_count;
+               lnum += fold_count;
                wp->w_lines[idx].wl_folded = TRUE;
-               wp->w_lines[idx].wl_lastlnum = lnum + fold_count;
+               wp->w_lines[idx].wl_lastlnum = lnum;
 # ifdef FEAT_SYN_HL
                did_update = DID_FOLD;
+# endif
+# ifdef FEAT_SPELL
+               // Check if the line after this fold requires capital.
+               if (has_spell && check_need_cap(wp, lnum + 1, 0))
+               {
+                   cap_col = 0;
+                   capcol_lnum = lnum + 1;
+               }
 # endif
            }
            else
```

I think it is correct but it is indeed not necessary for this PR.

luukvbaal:
MSVC doesn't allow empty structs apparently: https://github.com/vim/vim/actions/runs/5095746609/jobs/9160966932#step:8:64.

zeertzjq:
Search for "dummy typedefs" in structs.h

luukvbaal:
I see, I just moved one (non-dummy) struct member outside the `#ifdef FEAT_SPELL`. I'll leave it up to Bram.

brammool:
> MSVC doesn't allow empty structs apparently:
> https://github.com/vim/vim/actions/runs/5095746609/jobs/9160966932#step:8:64.

Hah, well, it makes sense for an actual structure.

About initializing it:

	    spellvars_T sv = { 0 };

I often use CLEAR_FIELD() as a simple way to avoid forgetting to
initialize an entry.  It does require an extra line, but manually
setting fields to zero can then be dropped.

The has_spell field looks like a boolean, using FALSE instead of zero
makes that clearer (if not using CLEAR_FIELD()).

I like using a prefix to make it easier to find struct fields.  In this
case "spv_" would work.  In this case a local "has_spell" variable can
then easily be distinguised from "spv_has_spell".

-- 
88. Every single time you press the 'Get mail' button...it does get new mail.

brammool:
> > There seem to be a few unrelated changes that might actually be
> > wrong, such as moving these two lines up: - wp->w_lines[idx].wl_lnum
> > = lnum; - wp->w_lines[idx].wl_valid = TRUE; Better avoid making
> > unrelated changes in this PR, it can only make it difficult to
> > figure out why something changed unrelated to what this PR was made
> > for.
> 
> I did this because I needed to use `lnum + foldcount` 3 times in this
> block so I incremented `lnum` but that meant having to move up those
> two lines up while cleaning up an `#ifdef FEAT_FOLD` block below.
> ```diff
> @@ -2450,10 +2469,19 @@ win_update(win_T *wp)
>                 fold_line(wp, fold_count, &win_foldinfo, lnum, row);
>                 ++row;
>                 --fold_count;
> +               lnum += fold_count;
>                 wp->w_lines[idx].wl_folded = TRUE;
> -               wp->w_lines[idx].wl_lastlnum = lnum + fold_count;
> +               wp->w_lines[idx].wl_lastlnum = lnum;
>  # ifdef FEAT_SYN_HL
>                 did_update = DID_FOLD;
> +# endif
> +# ifdef FEAT_SPELL
> +               // Check if the line after this fold requires capital.
> +               if (has_spell && check_need_cap(wp, lnum + 1, 0))
> +               {
> +                   cap_col = 0;
> +                   capcol_lnum = lnum + 1;
> +               }
>  # endif
>             }
>             else
> ```
> 
> I think it is correct but it is indeed not necessary for this PR.

I thought I spotted using wl_lnum before setting it to "lnum":

	&& wp->w_lines[idx].wl_lnum == lnum

Assigning "lnum" to wl_lnum before that would break something.

-- 
87. Everyone you know asks why your phone line is always busy ...and
    you tell them to send an e-mail.

luukvbaal:
The has_spell field looks like a boolean, using FALSE instead of zero
makes that clearer (if not using CLEAR_FIELD()).
> I like using a prefix to make it easier to find struct fields.  In this
case "spv_" would work.  In this case a local "has_spell" variable can
then easily be distinguised from "spv_has_spell".

Done.

> I thought I spotted using wl_lnum before setting it to "lnum":

>	&& wp->w_lines[idx].wl_lnum == lnum

> Assigning "lnum" to wl_lnum before that would break something.

Indeed.

>	    spellvars_T sv = { 0 };

> I often use CLEAR_FIELD() as a simple way to avoid forgetting to
initialize an entry.  It does require an extra line, but manually
setting fields to zero can then be dropped.

The other fields are not boolean though, and this line also initializes those to 0. So I don't think using `FALSE` there is better.
Feel free to use `CLEAR_FIELD()` if you prefer that. I think it's better if you make the final stylistic changes.

brammool:
> > The has_spell field looks like a boolean, using FALSE instead of zero
> > makes that clearer (if not using CLEAR_FIELD()).
> > I like using a prefix to make it easier to find struct fields.  In this
> > case "spv_" would work.  In this case a local "has_spell" variable can
> > then easily be distinguised from "spv_has_spell".
> 
> Done.
> 
> > I thought I spotted using wl_lnum before setting it to "lnum":
> 
> >	&& wp->w_lines[idx].wl_lnum == lnum
> 
> > Assigning "lnum" to wl_lnum before that would break something.
> 
> Indeed.
> 
> >	    spellvars_T sv = { 0 };
> 
> > I often use CLEAR_FIELD() as a simple way to avoid forgetting to
> > initialize an entry.  It does require an extra line, but manually
> > setting fields to zero can then be dropped.
> 
> The other fields are not boolean though, and this line also
> initializes those to 0. So I don't think using `FALSE` there is
> better.
> Feel free to use `CLEAR_FIELD()` if you prefer that. I think it's
> better if you make the final stylistic changes.

I did change it to use CLEAR_FIELD().  The single zero only
initializes the first field, some compilers will complain about missing
initializers.

I'm still not happy with the win_line() function, but let's leave that
for another time.

-- 
From "know your smileys":
 :~)	A man with a tape recorder up his nose

luukvbaal:
>  The single zero only
initializes the first field, some compilers will complain about missing
initializers.

As far as I know, omitted fields are "empty-initialized" to 0/NULL. Since C23 according to this page: https://en.cppreference.com/w/c/language/initialization#Empty_initialization.

brammool:
> > The single zero only initializes the first field, some compilers
> > will complain about missing initializers.
> 
> As far as I know, omitted fields are "empty-initialized" to 0/NULL.
> Since C23 according to this page:
> https://en.cppreference.com/w/c/language/initialization#Empty_initialization.

We use ANSI C (C89) as the oldest standard we support.  In practice we
have to deal with compilers that do not implement all of the standard or
just lag behind (MSVC was very slow adapting newer C standards).

I'm not sure if it is a gap in the standard or a compiler bug, but I
have seen incomplete structure initializers result in further fields
filled with random bytes.  Especially on the stack.

On top of that compilers try to be helpful by giving warnings for common
mistakes.  Thus even when something is valid by the standard, it may
still trigger a warning.  And we prefer not to get any warnings or
errors while building.

-- 
In order for something to become clean, something else must become dirty;
but you can get everything dirty without getting anything clean.

Issue #12443, opened 2023-05-25
Error messages E1106 and E1190 are not marked as plural


RestorerZ:
### Steps to reproduce

1. Run Vim -u NONE --noplugin
2. Type the following lines:
```
vim9script
echo map([1, 2, 3], () => 1)
```
3. Enter the command `:%so`
An error message will be displayed `E1106: One argument too many`
And it should be `E1106: 2 arguments too many` according to the logic of the map() function, which returns two arguments (index and value), and the test [task](https://github.com/vim/vim/blob/2ac6497f0ef186f0e3ba67d7f0a485bfb612bb08/src/testdir/test_vim9_func.vim#L4267) in the test_vim9_func.vim file.

The situation is similar for the test [task](https://github.com/vim/vim/blob/2ac6497f0ef186f0e3ba67d7f0a485bfb612bb08/src/testdir/test_vim9_import.vim#L1619) in the test_vim9_import.vim file
The same discrepancy is present for error E1190 and the corresponding test tasks in the [test_vim9_builtin.vim](https://github.com/vim/vim/blob/2ac6497f0ef186f0e3ba67d7f0a485bfb612bb08/src/testdir/test_vim9_builtin.vim#L2603) file and the [test_vim9_import.vim](https://github.com/vim/vim/blob/2ac6497f0ef186f0e3ba67d7f0a485bfb612bb08/src/testdir/test_vim9_import.vim#L1667) file

The bug most likely occurred in patch 9.0.0004 (#10606 ). Since build [9.0.0001](https://github.com/vim/vim-win32-installer/releases/tag/v9.0.0001) works correctly.

Also, in my opinion, the [description](https://vim-jp.org/vimdoc-en/vim9.html#E1106) of the E1106 error in the documentation is not entirely correct.

### Expected behaviour

Correct display of the message in accordance with the error conditions.

### Version of Vim

9.0.1572

### Environment

Windows NT6.1

brammool:
I cannot reproduce the problem, the example script does give "2 arguments too many".

I tried on Linux.  Could this be platform dependent? NGETTEXT() has three different implementations, but
I would think they all work the same way.  What does your ":version" output show for "gettext"?

RestorerZ:
Bram, and you will not reproduce it, since the problem turned out to be in a completely different place.
The fact is that in the errors.h file, this error message is split into two separate messages:
```
EXTERN char e_one_argument_too_many[]
	INIT(= N_("E1106: One argument too many"));
[...]

PR #12455, opened 2023-05-27
Added format specifier for messages E1106 and E1190 with a singular

RestorerZ:
Continuation #12443 
https://github.com/vim/vim/issues/12443#issuecomment-1565704369


RestorerZ:
Bram found the best solution to this.
https://github.com/vim/vim/issues/12443#issuecomment-1566205085

PR #12456, opened 2023-05-28
Invalid line pointer used after patch 9.0.1585

luukvbaal:
Grouping the spell checking blocks in the previous patch resulted in the "line" pointer in win_line() possibly being cleared by getting the pointer to the next line, apologies for that.

I guess this is why the block for getting the start of the next line was earlier in win_line() in the first place. This patch also works and keeps the spell checking code grouped.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -1449,9 +1449,6 @@ win_line(
 	area_highlighting = TRUE;
 #endif
 
-    line = ml_get_buf(wp->w_buffer, lnum, FALSE);
-    ptr = line;
-
 #ifdef FEAT_SPELL
     if (spv->spv_has_spell && !number_only)
     {
@@ -1466,6 +1463,17 @@ win_line(
 	    spv->spv_cap_col = -1;
 	spv->spv_checked_lnum = 0;
 
+	// Get the start of the next line, so that words that wrap to the
+	// next line are found too: "et<line-break>al.".
+	// Trick: skip a few chars for C/shell/Vim comments
+	nextline[SPWORDLEN] = NUL;
+	if (lnum < wp->w_buffer->b_ml.ml_line_count)
+	{
+	    line = ml_get_buf(wp->w_buffer, lnum + 1, FALSE);
+	    spell_cat_line(nextline + SPWORDLEN, line, SPWORDLEN);
+	}
+	line = ml_get_buf(wp->w_buffer, lnum, FALSE);
+
 	// For checking first word with a capital skip white space.
 	if (spv->spv_cap_col == 0)
 	    spv->spv_cap_col = getwhitecols(line);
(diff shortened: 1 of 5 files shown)


zeertzjq:
This make the problem catchable by ASAN:
```diff
diff --git a/src/testdir/test_spell.vim b/src/testdir/test_spell.vim
index 350077ad2..d8251d7cd 100644
--- a/src/testdir/test_spell.vim
+++ b/src/testdir/test_spell.vim
@@ -958,6 +958,7 @@ func Test_spell_screendump()
   CheckScreendump
 
   let lines =<< trim END
+       call test_override('alloc_lines', 1)
        call setline(1, [
              \ "This is some text without any spell errors.  Everything",
              \ "should just be black, nothing wrong here.",
@@ -980,6 +981,7 @@ func Test_spell_screendump_spellcap()
   CheckScreendump
 
   let lines =<< trim END
+       call test_override('alloc_lines', 1)
        call setline(1, [
              \ "   This line has a sepll error. and missing caps and trailing spaces.   ",
              \ "another missing cap here.",
@@ -1027,6 +1029,7 @@ func Test_spell_compatible()
   CheckScreendump
 
   let lines =<< trim END
+       call test_override('alloc_lines', 1)
        call setline(1, [
              \ "test "->repeat(20),
              \ "",
```

luukvbaal:
I also found another problem where SpellCap is not applied at the start of a line when the "mod_mid" or "mid_start" area is not adjacent to the area above it.
I added a test and applied a fix for this to `check_need_cap()` and set "spv_cap_col" to zero when that is the case.

luukvbaal:
> I also found another problem where SpellCap is not applied at the start of a line when the "mod_mid" or "mid_start" area is not adjacent to the area above it.
I added a test and applied a fix for this to check_need_cap() and set "spv_cap_col" to zero when that is the case.

(This problem was also present before the refactor done in 9.0.1585.)

luukvbaal:
> I also found another problem where SpellCap is not applied at the start of a line when the "mod_mid" or "mid_start" area is not adjacent to the area above it.
    I added a test and applied a fix for this to check_need_cap() and set "spv_cap_col" to zero when that is the case.

> (This problem was also present before the refactor done in 9.0.1585.)

Hmm I see further problems with SpellCap highlight not being drawn in this case. I reverted this PR to just the cleared line pointer fix. I'll look into the SpellCap issue further and open a differen PR for that later.

luukvbaal:
> Hmm I see further problems with SpellCap highlight not being drawn in this case. I reverted this PR to just the cleared line pointer fix. I'll look into the SpellCap issue further and open a differen PR for that later.

Since this PR is still open, and the patch depends on this one, I added it in a seperate commit here:

    SpellCap highlight not drawn when adding empty lines

    Problem:    SpellCap highlight is not drawn if the previous line is not
                drawn in a particular win_update(). SpellCap highlight is not
                drawn with multiple empty lines.
    Solution:   Reset "spv_capcol_lnum" to indicate that the next drawn
                line should be checked for capital. Always check for empty line.

PR #12427, opened 2023-05-22
Not all error messages are ready for translation.

RestorerZ:
Not all error messages are ready for translation. Added N_(). Continued patch 9.0.1572

PR #12459, opened 2023-05-28
Internal error messages are not for translation

RestorerZ:
Removed  \_() calls in the iemsg() and siemsg() functions.
Removed N_() in errors.h for internal error messages and debug messages.
Removed \_() calls in messages in \*DEBUG* definitions
The messages E864, E868, E874, E875, E876, E878 remained in question. They are output to the user via emsg() and are not marked as debugging in the code, but the documentation says that "This is only useful when debugging Vim".
As well as messages displayed via the PERROR() macro. Some of these messages have translations in errors.h, and some do not. And many of these messages are clearly debugging in kind.

### DIFF
--- a/runtime/doc/message.txt
+++ b/runtime/doc/message.txt
@@ -76,9 +76,9 @@ See `:messages` above.
 
 
 LIST OF MESSAGES
-			*E222* *E228* *E232* *E293* *E298* *E304* *E317*
-			*E318* *E356* *E438* *E439* *E440* *E316* *E320* *E322*
-			*E323* *E341* *E473* *E570* *E292*
+			*E222* *E228* *E232* *E292* *E293* *E298* *E304* *E316*
+			*E317* *E318* *E320* *E322* *E323* *E341* *E356* *E438*
+			*E439* *E440* *E473* *E570*
   Add to read buffer ~
   makemap: Illegal mode ~
   Cannot create BalloonEval with both message and callback ~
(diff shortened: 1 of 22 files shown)


RestorerZ:
@brammool


> I noticed a few calls in regexp.c that were still using _() and removed those.

I used /iemsg\\|siemsg so I overlooked it apparently.



> I don't see a reason why the translation for mch_errmsg() in main.c was removed. I restored it.

I proceeded from the fact that this message is contained in the definition NBDEBUG and acted by analogy with [this](https://github.com/vim/vim/blob/eb43b7f0531bd13d15580b5c262a25d6a52a0823/src/netbeans.c#L514), which you once [pointed out](https://github.com/vim/vim/pull/12427#issuecomment-1560184587). It is possible that this is not applicable in this case.

brammool:
> > I don't see a reason why the translation for mch_errmsg() in main.c
> > was removed. I restored it.
> 
> I proceeded from the fact that this message is contained in the
> definition NBDEBUG and acted by analogy with
> [this](https://github.com/vim/vim/blob/eb43b7f0531bd13d15580b5c262a25d6a52a0823/src/netbeans.c#L514),
> which you once [pointed
> out](https://github.com/vim/vim/pull/12427#issuecomment-1560184587).
> It is possible that this is not applicable in this case.

I had not spotted the NBDEBUG dependency.  I just thought it was a
useful message.  But yes, to be consistent this message is not to be
translated.

-- 
# echo reboot >universe
# chmod +x universe
# ./universe

PR #12463, opened 2023-05-29
Recognise new cabal config location

coot:


coot:
I tried to set `XDG_CONFIG_HOME` in `Test_filetype_detection` with:
```
  if !empty(windowsversion())
    let $XDG_CONFIG_HOME=expand("$HOME/.config")
  endif
```
and `vim` complains with:
```
Found errors in Test_filetype_detection():
Caught exception in Test_filetype_detection(): Vim(call):E1126: Cannot use :let in Vim9 script @ command line..script /home/coot/src/vim/src/testdir/runtest.vim[569]..function Ru
nTheTest[52]..Test_filetype_detection, line 2
```

coot:
Oh, I see `var` should be used, but it doesn't support setting environment variables. Maybe there's a function for that now...

coot:
@brammool I added a test which caught a mistake :tada:.

brammool:
> I tried to set `XDG_CONFIG_HOME` in `Test_filetype_detection` with:
> ```
>   if !empty(windowsversion())
>     let $XDG_CONFIG_HOME=expand("$HOME/.config")
>   endif
> ```
> and `vim` complains with:
> ```
> Found errors in Test_filetype_detection():
> Caught exception in Test_filetype_detection(): Vim(call):E1126: Cannot use :let in Vim9 script @ command line..script /home/coot/src/vim/src/testdir/runtest.vim[569]..function RunTheTest[52]..Test_filetype_detection, line 2
> ```

Right, you can just drop the "let":
     $XDG_CONFIG_HOME = expand("$HOME/.config")

For checking if an environment variable exists using exists('$NAME')
is simpler and more explicit than has_key() and environ().  I'll make it
like that.

It's better not the create test files under $HOME, in case the test
aborts files may be left behind unnoticed.  That requires a few more
changes, I'll do that.  Please check with what I ended up with.

-- 
An error has occurred.  Hit any user to continue.

PR #12466, opened 2023-05-30
doc: Write about --cmd in startup part

k-takata:
`:h startup` doesn't describe where the --cmd arguments are executed. Write about it.

Note:
The command line is parsed in command_line_scan() in main.c, then mch_init(), set_init_2() and some initializations are executed, then exe_pre_commands() executes the --cmd arguments.

### DIFF
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -795,6 +795,7 @@ accordingly.  Vim proceeds in this order:
 	inspected.  Buffers are created for all files (but not loaded yet).
 	The |-V| argument can be used to display or log what happens next,
 	useful for debugging the initializations.
+	The |--cmd| arguments are also executed here.
 
 3. Execute Ex commands, from environment variables and/or files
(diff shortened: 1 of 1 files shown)


brammool:
> `:h startup` doesn't describe where the --cmd arguments are executed.
> Write about it.
> 
> Note:
> The command line is parsed in command_line_scan() in main.c, then
> mch_init(), set_init_2() and some initializations are executed, then
> exe_pre_commands() executes the --cmd arguments.

I think it should be above the last line in that section:

2. Process the arguments
	The options and file names from the command that start Vim are
	inspected.
	The |-V| argument can be used to display or log what happens next,
	useful for debugging the initializations.
	The |--cmd| arguments are executed.
	Buffers are created for all files (but not loaded yet).

A few more things are happening, but we don't want too many details in
this overview.

-- 
From "know your smileys":
 (X0||)   Double hamburger with lettuce and tomato

PR #12467, opened 2023-05-30
Fix that setvbuf causes assertion error

k-takata:
When Vim is compiled with debug mode on MSVC, the following command causes an assertion error:

```
> vimd -e -s -c q
```

```
Debug Assertion Failed!

Program: C:\work\vim-msvc\src\vim32d.dll
File: minkernel\crts\ucrt\src\appcrt\stdio\setvbuf.cpp
Line: 55

Expression: 2 <= buffer_size_in_bytes && buffer_size_in_bytes <= INT_MAX
```

This is because MSVC's setvbuf() doesn't accept 0 as the 4th parameter. https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setvbuf?view=msvc-170

Additionally, it seems that passing NULL to the 2nd parameter is no-op in some C libraries.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -76,10 +76,11 @@ static char *(main_errors[]) =
 // Various parameters passed between main() and other functions.
 static mparm_T	params;
 
+static void *s_vbuf = NULL;		// buffer for setvbuf()
+
 #ifndef NO_VIM_MAIN	// skip this for unittests
 
 static char_u *start_dir = NULL;	// current working dir on startup
-static void *s_vbuf = NULL;		// buffer for setvbuf()
 
 static int has_dash_c_arg = FALSE;
 
--- a/src/main.c
+++ b/src/main.c
@@ -356,7 +356,10 @@ main
     // Ensure output works usefully without a tty: buffer lines instead of
     // fully buffered.
     if (silent_mode)
-	setvbuf(stdout, NULL, _IOLBF, 0);
+    {
+	void *buf = malloc(BUFSIZ);
+	setvbuf(stdout, buf, _IOLBF, BUFSIZ);
+    }
 #endif
 
     // This message comes before term inits, but after setting "silent_mode"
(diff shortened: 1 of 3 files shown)


vim-ml:
Hi,

Op di 30 mei 2023 08:00 schreef K.Takata ***@***.***>:

> When Vim is compiled with debug mode on MSVC, the following command causes
> an assertion error:
>

This leaks memory, though, because of the malloc() call...

Christ van Willegen

>

brammool:
Making the buffer a global variable should avoid a warning for leaked memory.

k-takata:
Updated to free the buffer when EXITFREE is defined.

src/main.c
```diff
@@ -436,6 +442,19 @@ main
 #endif // NO_VIM_MAIN
 #endif // PROTO
 
+#if defined(EXITFREE) || defined(PROTO)
+    void
+free_vbuf(void)
+{
+    if (s_vbuf != NULL)
+    {
+	setvbuf(stdout, NULL, _IONBF, 0);
```
chrisbra, line 451: > setvbuf(stdout, NULL, _IONBF, 0);

Is this call required? Because now you are introducing the same problem that this PR is fixing? Shouldn't `size` be larger than 1?

k-takata: Before freeing `s_vbuf` we should stop using it in `setvbuf()`.
So I use `_IONBF` (not `_IOLBF`) here to stop buffering.

chrisbra: okay, but what about the size then? I think 0 is not valid according to your link.

k-takata: The assertion error relates only to `_IOFBF` and `_IOLBF`, not to `_IONBF`.

> `_IONBF`  No buffer is used, regardless of `buffer` or `size`.

chrisbra: ah, thanks. Makes sense

k-takata: If you have installed MSVC, you should be able to find the source code of `setvbuf()` here:
`C:\Program Files (x86)\Windows Kits\10\Source\<BuildNumber>\ucrt\stdio\setvbuf.cpp`

PR #12469, opened 2023-05-30
Fix wrong character with 'smoothscroll' marker and double-width char

zeertzjq:


src/drawline.c
```diff
@@ -838,7 +838,13 @@ wlv_screen_line(win_T *wp, winlinevars_T *wlv, int negative_width)
 	{
 	    ScreenLines[off] = '<';
 	    if (enc_utf8)
+	    {
 		ScreenLinesUC[off] = 0;
+		if (i + 1 + skip < wp->w_width && ScreenLines[off + 1] == 0)
+		    // When the first half of a double-width character is
+		    // overwritten, change the second half to a space.
+		    ScreenLines[off + 1] = ' ';
```
zeertzjq, line 846: For 'listchars' "precedes", `MB_FILLER_CHAR` is used, but using it here leads to `>>><`, which looks a bit strange.

src/drawline.c
```diff
@@ -836,6 +838,10 @@ wlv_screen_line(win_T *wp, winlinevars_T *wlv, int negative_width)
 
 	for (int i = 0; i < 3 && i + skip < wp->w_width; ++i)
 	{
+	    if ((*mb_off2cells)(off, max_off) > 1)
+		// When the first half of a double-width character is
+		// overwritten, change the second half to a space.
+		ScreenLines[off + 1] = ' ';
```
zeertzjq, line 844: For 'listchars' "precedes", `MB_FILLER_CHAR` is used, but using it here leads to `>>><`, which looks a bit strange.

zeertzjq:
This now also fixes the whole first line becoming completely wrong when using DBCS, but I cannot find a way to test that.

zeertzjq:
> This now also fixes the whole first line becoming completely wrong when using DBCS, but I cannot find a way to test that.

Functions like `screenstring()` don't seem to support DBCS, and it is worthy to add DBCS support to these functions just to test this?

k-takata:
> Functions like `screenstring()` don't seem to support DBCS,

I think it is supported. See Test_screenchar_utf8() in test_utf8.vim.

brammool:
> Ah, I wasn't aware that this supports UTF-8 but doesn't support DBCS.

DBCS is stored differently, this was implemented before Unicode support
and makes use of the property that with this encoding a double-wide
character uses two bytes.  These two bytes are then stored in two
positions in ScreenLines[].  Except for DBCS_JPNU which uses
ScreenLines2[] for single-width double-byte characters.

It should be fairly easy to implement.  Search for ScreenLines2[] in
src/drawline.c to see how bytes are stored.  f_screenstring() then needs
to do the reverse.

-- 
% cat /usr/include/real_life.h
void life(void);

src/drawline.c
brammool: Although it looks a bit strange, the computation of w_skipcol gets more complicated if we want to avoid it.  Also, using CTRL-E should not cause the text to shift left or right, thus we have this round-off for double width characters.
A nicer way would be to write the character on top of the last ">", thus we would only use two ">" characters, giving ">>##" (where "##" is the double width character).  But again this would be more code for covering a corner case.  We can leave it for later.

brammool:
Do you plan to add DBCS support to screenstring()?  Or are you waiting for someone to implement this?

zeertzjq:
I'll try to do that soon.

zeertzjq:
I changed functions like `screenstring()` to use `screen_getbytes()`.

brammool:
> I changed functions like `screenstring()` to use `screen_getbytes()`.

Let me make a patch for this first, the code and test for it as separate
from fixing the original problem.

-- 
From "know your smileys":
 :.-(	Crying

PR #12473, opened 2023-05-31
Add missing flag in :registers

julio-b:
Considering that `:display` and `:registers` are essentially the same command, it makes sense for them to have identical flags.

Sandbox flag was added specifically for `:dis` in vim-6-0h and has remained unchanged since then.

https://github.com/vim/vim-history/commit/523952c293cc00bd18c4f4db4a70c6873c122999#diff-e4f9580853ad69360251409927f914ae4d57bce9a3c8b1c345992c9cbaf55548L231-L233

Is this patch correct @brammool or both commands need Sandbox?


julio-b:
> The commands are the same, thus the flags should indeed be the same. But instead of removing EX_SBOXOK from "display", it should be added to "registers". These commands only displays information, there should be no problem using them in the sandbox.
> 
> -- From "know your smileys": 2B|^2B Message from Shakespeare /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net \\\ /// \\\ \\\ sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ /// \\\ help me help AIDS victims -- http://ICCF-Holland.org ///

Thank you for the clarification, i fixed the patch.

PR #12475, opened 2023-05-31
swayconfig syntax file updates

jamespeapen:
- add `seat` as a keyword 
- fix: add `container` to `mouse_warping` keyword

Sway syntax file specific commits from jamespeapen/swayconfig.vim#13.


brammool:
I'll include it, thanks.

PR #12477, opened 2023-06-01
Fix wrong display when scrolling multiple lines with 'smoothscroll'

### LINKED ISSUE #12468: Wrong display when scrolling multiple lines with `smoothscroll`
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
9split
40vnew
setlocal cursorline scrolloff=0 smoothscroll
call setline(1, repeat([''], 9))
call setline(3, repeat('a', 50))
call setline(8, 'bbb')
call setline(9, 'ccc')
redraw
```
3. Press `3<C-E>`
4. The `bbb` on line 8 disappears and the wrapped `aaaaaaaaaa` at the top is drawn twice

### Expected behaviour

The `bbb` on line 8 doesn't disappear and the wrapped `aaaaaaaaaa` at the top is drawn once

### Version of Vim

9.0.1591

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

zeertzjq:
Fix #12468


brammool:
I think the redraw may also be needed when w_skipcol becomes zero.  So I added prev_skipcol and compare with that.
The check for prev_topline may not be needed then: If w_topline does not change then w_skipcol must change, right?
I haven't tried removing prev_topline.

zeertzjq:
> I think the redraw may also be needed when w_skipcol becomes zero. So I added prev_skipcol and compare with that. The check for prev_topline may not be needed then: If w_topline does not change then w_skipcol must change, right? I haven't tried removing prev_topline.

I don't think this is correct. If w_topline changed but w_skipcol is the same non-zero value then a redraw is still needed. I've created a new PR #12490 for this case.

PR #12480, opened 2023-06-02
Cursor not adjusted at buffer boundaries when 'splitkeep' is not "cursor"

luukvbaal:
Problem:    Cursor not adjusted at buffer boundaries when 'splitkeep' is
            not "cursor"
Solution:   Move error handling to outer cursor move functions, inner
            functions should only return valid cursor positions.

I could just as well have accounted for this in `win_fix_cursor()` but I think it is unexpected that `cursor_up/down_inner()` could return 0. Thus, I moved the error handling to the outer functions so that the inner functions always return a valid cursor position.
Let me know if you would rather keep the error handling in the inner functions. Then I could check the return value in `win_fix_cursor()` instead.

Regarding the updated TODO comment for `win_fix_scroll()`, this relates back to the discussion in https://github.com/vim/vim/issues/12411#issuecomment-1552965532. If you don't think this should be allowed, perhaps remove the TODO note instead.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -2762,10 +2762,6 @@ cursor_up_inner(win_T *wp, long n)
 {
     linenr_T	lnum = wp->w_cursor.lnum;
 
-    // This fails if the cursor is already in the first line or the count is
-    // larger than the line number and '-' is in 'cpoptions'
-    if (lnum <= 1 || (n >= lnum && vim_strchr(p_cpo, CPO_MINUS) != NULL))
-	return 0;
     if (n >= lnum)
 	lnum = 1;
     else
@@ -2806,9 +2802,13 @@ cursor_up(
     long	n,
     int		upd_topline)	    // When TRUE: update topline
 {
-    if (n > 0 && cursor_up_inner(curwin, n) == 0)
+    // This fails if the cursor is already in the first line or the count is
+    // larger than the line number and '-' is in 'cpoptions'
+    linenr_T lnum = curwin->w_cursor.lnum;
+    if (n > 0 && (lnum <= 1 || (n >= lnum && vim_strchr(p_cpo, CPO_MINUS))))
 	return FAIL;
 
+    cursor_up_inner(curwin, n);
     // try to advance to the column we want to be at
     coladvance(curwin->w_curswant);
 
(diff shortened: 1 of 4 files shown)


brammool:
> Problem:    Cursor not adjusted at buffer boundaries when 'splitkeep' is
>             not "cursor"
> Solution:   Move error handling to outer cursor move functions, inner
>             functions should only return valid cursor positions.
> 
> I could just as well have accounted for this in `win_fix_cursor()` but
> I think it is unexpected that `cursor_up/down_inner()` could return 0.
> Thus, I moved the error handling to the outer functions so that the
> inner functions always return a valid cursor position.
> Let me know if you would rather keep the error handling in the inner
> functions. Then I could check the return value in `win_fix_cursor()`
> instead.

It's OK to move the "error handling" outside of the inner functions.
It's really a boundary check, and what to do when running into the
boundary depends on what it is used for, thus it makes sense to check
that in the outer functions.

The inner functions now return the new cursor line number, that is not
very useful, and all the calls but one don't use the return value.  So
we might as well drop that.

I also prefer to explicitly check the return value of vim_strchr() for
NULL, since it doesn't return a boolean. It's clearer that way.

I'll make these minor changes.

> Regarding the updated TODO comment for `win_fix_scroll()`, this
> relates back to the discussion in
> https://github.com/vim/vim/issues/12411#issuecomment-1552965532. If
> you don't think this should be allowed, perhaps remove the TODO note
> instead.

Well, now that 'smoothscroll' is working, it would be very nice if it
also makes 'splitkeep' set to "screen" work better.  It's going to be
tricky to make this work, but for a user who intentionally is using this
combination of option values it would be annoying if text moves up/down
even though it would be possible to avoid it (it is not possible when
running into the top of the buffer, we don't have a good way to blank
out the text above the first line.  Well, we do have filler lines in
diff mode, but it's highly unexpected to see them when not in diff mode.
Probably would trigger several other bugs, since filler lines are
currently only taken into account in diff mode.  Let's not go that way.

-- 
92. It takes you two hours to check all 14 of your mailboxes.

luukvbaal:
> Well, now that 'smoothscroll' is working, it would be very nice if it
also makes 'splitkeep' set to "screen" work better.  It's going to be
tricky to make this work

Indeed, I have attempted it before but there are a few things in the way of making it work properly. Basically I think we need to perform a "$" -> "L" -> "<diff>gj/k" -> "zb" instead of the current "L" -> "j/k" -> "scroll_to_fraction()" in the `win_fix_scroll()` loop. That way we can make use of the already existing `'smoothscroll'` logic.

That will only work properly if:
* An insignificant amount of "win_" function variants are added(`nv_screengo()`, `coladvance()` `scroll_cursor_bot()`, it goes deep...), or "curwin" is saved and temporarily set inside the `win_fix_scroll()` loop.
* `scroll_cursor_bot()` brings the cursor to the actual last screenline of a window, even when that line is wrapped.

PR #12481, opened 2023-06-02
Ignore some Make output in errorformat

gpanders:
Problem: GNU Make issues messages such as

    make[1]: *** [foo/bar/baz/Makefile:210: src/foo.o] Error 1

The default 'errorformat' value matches these messages because of the
%f:%l:%m pattern, and creates quickfix list entries that appear valid
but create a new buffer named `make[1]: *** [foo/bar/baz/Makefile`,
which is obviously incorrect.

Solution: Add entries to 'errorformat' to ignore these messages.

The entries added are:

1. `%-Gg%\?make[%*\d]: *** [%f:%l:%m`
2. `%-Gg%\?make: *** [%f:%l:%m`

This covers all of the following cases:

    make[1]: *** [foo/bar/baz/Makefile:200: src/foo.o] Error 1
    gmake[1]: *** [foo/bar/baz/Makefile:200: src/foo.o] Error 1
    make: *** [Makefile:114: all] Error 2
    gmake: *** [Makefile:114: all] Error 2

### DIFF
--- a/src/option.h
+++ b/src/option.h
@@ -89,7 +89,7 @@ typedef enum {
 #   ifdef VMS
 #    define DFLT_EFM	"%A%p^,%C%%CC-%t-%m,%Cat line number %l in file %f,%f|%l| %m"
 #   else // Unix, probably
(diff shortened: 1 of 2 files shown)


brammool:
> Problem: GNU Make issues messages such as
> 
>     make[1]: *** [foo/bar/baz/Makefile:210] Error 1
> 
> The default 'errorformat' value matches these messages because of the
> %f:%l:%m pattern, and creates quickfix list entries that appear valid
> but create a new buffer named `make[1]: *** [foo/bar/baz/Makefile`,
> which is obviously incorrect.
> 
> Solution: Add entries to 'errorformat' to ignore these messages.
> 
> The entries added are:
> 
> 1. `%-Gg%\?make[%*\d]: *** [%f:%l:%m`
> 2. `%-Gg%\?make: *** [%f:%l:%m`
> 
> This covers all of the following cases:
> 
>     make[1]: *** [foo/bar/baz/Makefile:200] Error 1
>     gmake[1]: *** [foo/bar/baz/Makefile:200] Error 1
>     make: *** [Makefile:114: all] Error 2
>     gmake: *** [Makefile:114: all] Error 2

Thanks, looks useful.

Can you add a test for this?  The option value is rather complicated,
further additions may break this again.  For an example look in
src/testdir/test_quickfix.vim, Test_efm1().  You can add a new function
to avoid cluttering the existing one.  Essential is to keep the
"CheckUnix" line, since your change only is in the Unix default value.

-- 
In a world without walls and borders, who needs windows and gates?

gpanders:
There is actually already a line in the `Test_efm1` error file like this:

```
    make: *** [vim] Error 1
```

I changed this to

```
    make: *** [src/vim/testdir/Makefile:100: test_quickfix] Error 1
```

which causes the test to fail on `master`, but passes with this patch applied. Hopefully that is sufficient.

CervEdin:
I'm very excited by the prospect of not having `make: *** [` buffers` created all the time it's been driving me insane.

But does this ignore the messages? Wouldn't the sane default by to not ignore the error, but match it and properly locate the error to the appropriate location (in the makefile)?

PR #12486, opened 2023-06-03
Fix screenpos() not taking skipcol into account

### LINKED ISSUE #12476: screenpos() returns wrong result for wrapped line
### Steps to reproduce

```
vim --clean
```
```
call setline(1, 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet')
exe "normal! z1\<CR>$"
```
The result is all zeros:
image

### Expected behaviour

Should return something non-zero, because the position is visible.

Related: [8.2.3193](https://github.com/vim/vim/commit/189663bdac1156237c49925f77bd197c1bdea12c)

### Version of Vim

9.0.1558

### Environment

Ubuntu 22.04

### Logs and stack traces

_No response_

zeertzjq:
Fix #12476

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -1479,6 +1479,7 @@ textpos2screenpos(
 	    col += off;
 	    width = wp->w_width - off + win_col_off2(wp);
 
+	    col -= wp->w_skipcol;
 	    // long line wrapping, adjust row
 	    if (wp->w_p_wrap
 		    && col >= (colnr_T)wp->w_width
--- a/src/testdir/test_cursor_func.vim
+++ b/src/testdir/test_cursor_func.vim
@@ -118,7 +118,7 @@ func Test_screenpos()
 
   let wininfo = getwininfo(winid)[0]
   call setline(3, ['x']->repeat(wininfo.height))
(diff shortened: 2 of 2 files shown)


brammool:
> Fix #12476

Thanks.  I noticed that w_skipcol is subtracted unconditionally, but it
should only happen for w_topline.  Adding a test for this I found
another problem in plines_win_nofold() where w_skipcol also needs to be
subtracted.  The tests pass now, I hope this doesn't break another
corner case.

-- 
To be rich is not the end, but only a change of worries.

brammool:
Why did you close this PR?  Fixed by another PR perhaps?

zeertzjq:
Oops, I thought this will soon be included in a larger patch.

brammool:
I had already included it and made some changes on top.   Let me finish this and then please check it doesn't interfere with the other PRs.

PR #12488, opened 2023-06-03
'splitkeep' handles undrawn cursor position

luukvbaal:
Problem:    Cursor is adjusted to keep "topline" or "screen" for
            a cursor position that was not drawn in the first place.
Solution:   Avoid adjusting cursor position when the current cursor
            position was never drawn.

### DIFF
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -2589,6 +2589,7 @@ win_update(win_T *wp)
     wp->w_last_cursorline = wp->w_p_cul ? wp->w_cursor.lnum : 0;
 #endif
     wp->w_last_cursor_lnum_rnu = wp->w_p_rnu ? wp->w_cursor.lnum : 0;
+    wp->w_last_cursor_lnum_drawn = wp->w_cursor.lnum;
 
 #ifdef FEAT_VTP
     // Rewrite the character at the end of the screen line.
--- a/src/structs.h
+++ b/src/structs.h
@@ -3709,6 +3709,8 @@ struct window_S
 
     linenr_T    w_last_cursor_lnum_rnu;  // cursor lnum when 'rnu' was last
 					 // redrawn
+    linenr_T    w_last_cursor_lnum_drawn;// cursor lnum when window was last
+					 // redrawn
 
     lcs_chars_T	w_lcs_chars;	    // 'listchars' characters
     fill_chars_T w_fill_chars;	    // 'fillchars' characters
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1703,7 +1703,7 @@ func Test_splitkeep_options()
     " No scroll when resizing windows
     wincmd k | resize +2 | redraw
     call assert_equal(1, line("w0"))
-    wincmd j
+    wincmd j | redraw!
     call assert_equal(&spk == 'topline' ? 1 : win_screenpos(0)[0] - tl - wsb, line("w0"))
 
     " No scroll when dragging statusline
@@ -1715,7 +1715,7 @@ func Test_splitkeep_options()
     " No scroll when changing shellsize
     set lines+=2
     call assert_equal(1, line("w0"))
-    wincmd j
+    wincmd j | redraw
     call assert_equal(&spk == 'topline' ? 1 : win_screenpos(0)[0] - tl - wsb, line("w0"))
     set lines-=2
     call assert_equal(&spk == 'topline' ? 1 : win_screenpos(0)[0] - tl - wsb, line("w0"))
@@ -1824,10 +1824,12 @@ func Test_splitkeep_misc()
   " Cursor is adjusted to start and end of buffer
   norm M
   wincmd s
+  redraw
   resize 1
   call assert_equal(1, line('.'))
   wincmd j
   norm GM
+  redraw
   resize 1
   call assert_equal(&lines, line('.'))
   only!
@@ -1858,6 +1860,30 @@ func Test_splitkeep_misc()
 
   %bwipeout!
   set splitbelow&
+  set splitkeep&
+endfunc
+
+func Test_splitkeep_cursor_resize()
+  set splitkeep=screen
+
+  func CursorResize()
+    call cursor(100, 1)
+    wincmd -
+  endfunc
+  func CursorEqualize()
+    call cursor(100, 1)
+    wincmd =
+  endfunc
+
+  call setline(1, range(1, 200))
+  call CursorResize()
+  call assert_equal(100, line('.'))
+
+  norm gg
+  call CursorEqualize()
+  call assert_equal(100, line('.'))
+
+  bwipeout!
   set splitkeep&
 endfunc
 
--- a/src/window.c
+++ b/src/window.c
@@ -6821,7 +6821,11 @@ win_fix_scroll(int resize)
 	wp->w_prev_winrow = wp->w_winrow;
     }
     skip_update_topline = FALSE;
-    // Ensure cursor is valid when not in normal mode or when resized.
+    // If the cursor position hasn't changed since it was last drawn, ensure
+    // cursor is valid when not in normal mode or when resized.  If it has
+    // changed, let update_topline() move the cursor onto the screen.
+    if (curwin->w_cursor.lnum != curwin->w_last_cursor_lnum_drawn)
+	return;
     if (!(get_real_state() & (MODE_NORMAL|MODE_CMDLINE|MODE_TERMINAL)))
 	win_fix_cursor(FALSE);
     else if (resize)
@@ -6874,7 +6878,7 @@ win_fix_cursor(int normal)
 	{
 	    wp->w_fraction = (nlnum == bot) ? FRACTION_MULT : 0;
 	    scroll_to_fraction(wp, wp->w_prev_height);
-	    validate_botline_win(curwin);
+	    validate_botline();
 	}
     }
 }


brammool:
> Problem:    Outdated topline and botline may be used to update the cursor
>             position when 'splitkeep' is not "cursor".
> Solution:   Update topline when cursor is set and validate botline when
>             adjusting cursor position for 'splitkeep'.
> 
> Are we okay with always calling `update_topline()` in
> `set_cursorpos()`? I don't think it's harmful but I guess it is not
> necessary when 'splitkeep' is "cursor". I don't think it can be
> avoided for when 'splitkeep' is not cursor.

Generally update_topline() is called only before using w_topline.
Both because it is an "expensive" function and because there are
situations where we want to move the cursor, do something there, then
restore the cursor and expect w_topline not to change.

In win_fix_scroll() updating w_topline to scroll to be able to see the
cursor position is disabled by setting "skip_update_topline".  It's a
clear conflict between moving the cursor and expecting the window to
scroll to be able to see where it ended up, versus resizing/scrolling
and moving the cursor to make sure it is visible in the displayed lines.

In your test you do both: Move the cursor and resize the window.  For
the first w_topline should be updated to show the cursor, for the resize
the cursor may need to be moved to show the text according to
'splitkeep'.

What we could try is to check if w_topline is valid and update it when
not.  Perhaps these lines at the start of win_fix_scroll(), before
setting skip_update_topline will work:

    check_cursor_moved(curwin);
    if ((curwin->w_valid & VALID_TOPLINE) == 0)
	update_topline();

It's hard to predict if this breaks anything, hopefully we have enough
test coverage to notice any problems this causes.

-- 

luukvbaal:
> Generally update_topline() is called only before using w_topline.
Both because it is an "expensive" function and because there are
situations where we want to move the cursor, do something there, then
restore the cursor and expect w_topline not to change.

I understand. Perhaps we can set a global in `set_cursorpos()` to indicate that "w_topline" should be updated in `win_fix_cursor()`? That way it doesn't affect 'splitkeep' == "cursor" and will only affect 'splitkeep' != "cursor" when a `cursor()` call is also paired with a resize (before the screen is redrawn). I really don't think we can add an `update_topline()` to `win_fix_scroll()`, elaborated below. And I do think that when a user explicitly calls `cursor()` the cursor should actually be moved there, even when this is paired with changing window sizes and 'splitkeep' != "cursor", like in the added test.

> What we could try is to check if w_topline is valid and update it when
not.  Perhaps these lines at the start of win_fix_scroll(), before
setting skip_update_topline will work:

>    check_cursor_moved(curwin);
    if ((curwin->w_valid & VALID_TOPLINE) == 0)
	update_topline();

> It's hard to predict if this breaks anything, hopefully we have enough
test coverage to notice any problems this causes.

I think we cannot call `update_topline()` until _after_ `win_fix_scroll()` has moved the cursor into the visible region. The topline might be marked invalid despite the cursor not having moved explicitly, in which case we do not want to update the topline but move the cursor instead. Adding the suggested lines to the start of `win_fix_scroll()` does therefor indeed result in numerous test failures for 'splitkeep', which I do think is tested quite rigorously.

If the cursor position is set explicitly however, we need to update the topline instead. If we cannot do this unconditionally, I'll try updating this to use a global variable to see if you like that better.

luukvbaal:
I think we are fine to re-use `skip_win_fix_scroll` and avoid "fixing" the cursor position altogether after the cursor position was explicitly set. Updated again:

    Cursor is adjusted after it was explicitly set when 'splitkeep' is not "cursor"

    Problem:    Cursor is adjusted after it was explicitly set when
                'splitkeep' is not "cursor".
    Solution:   Avoid adjusting cursor position for a change in window
                height after it was explicitly set.

brammool:
> > Generally update_topline() is called only before using w_topline.
> > Both because it is an "expensive" function and because there are
> > situations where we want to move the cursor, do something there, then
> > restore the cursor and expect w_topline not to change.
> 
> I understand. Perhaps we can set a global in `set_cursorpos()` to
> indicate that "w_topline" should be updated in `win_fix_cursor()`?

There are many more ways to move the cursor.  Setting a flag every time
the cursor is moved (by the user or by a script) is not realistic.
w_valid_cursor is used to know whether the cursor moved since the last
time w_topline was updated.  The only thing we don't know is whether
w_cursor was set by the user or as a side effect of scrolling the text
and moving the cursor into the visible text.  More about that below.

> That way it doesn't affect 'splitkeep' == "cursor" and will only
> affect 'splitkeep' != "cursor" when a `cursor()` call is also paired
> with a resize (before the screen is redrawn). I really don't think we
> can add an `update_topline()` to `win_fix_scroll()`, elaborated below.
> And I do think that when a user explicitly calls `cursor()` the cursor
> should actually be moved there, even when this is paired with changing
> window sizes and 'splitkeep' != "cursor", like in the added test.

It's hard to say what the expected behavior is when moving the cursor
and then splitting the window, especially when 'splitkeep' is not
"cursor".

In the test the cursor is moved to line 100.  Then the window sizes are
equalized while 'slitkeep' is set to "screen".  Now does the user expect
the cursor to be still in line 100, or the text not moved up/down and
the cursor moved if needed?  I would think that the last action
overrides a previous one.  Especially if some time passes or other
commands are executed in between the two actions.

> > What we could try is to check if w_topline is valid and update it when
> > not.  Perhaps these lines at the start of win_fix_scroll(), before
> > setting skip_update_topline will work:
> 
> >    check_cursor_moved(curwin);
> >    if ((curwin->w_valid & VALID_TOPLINE) == 0)
> >	update_topline();
> 
> > It's hard to predict if this breaks anything, hopefully we have enough
> > test coverage to notice any problems this causes.
> 
> I think we cannot call `update_topline()` until _after_
> `win_fix_scroll()` has moved the cursor into the visible region. The
> topline might be marked invalid despite the cursor not having moved
> explicitly, in which case we do not want to update the topline but move
> the cursor instead. Adding the suggested lines to the start of
> `win_fix_scroll()` does therefor indeed result in numerous test failures
> for 'splitkeep', which I do think is tested quite rigorously.
> 
> If the cursor position is set explicitly however, we need to update the
> topline instead. If we cannot do this unconditionally, I'll try updating
> this to use a global variable to see if you like that better.

The existing flags in w_valid, especially VALID_TOPLINE, are intended
for this.  Calling check_cursor_moved() to update w_valid should have
the same effect as a global flag to indicate the cursor was moved.

We do have to make sure the flag (and w_valid_cursor) are updated when
moving the cursor for another reason (e.g. in win_fix_cursor()).  That
is the tricky part.  But not more tricky than using a global flag.

-- 
Amnesia is one of my favorite words, but I forgot what it means.

luukvbaal:
> There are many more ways to move the cursor. Setting a flag every time the cursor is moved (by the user or by a script) is not realistic. w_valid_cursor is used to know whether the cursor moved since the last time w_topline was updated. The only thing we don't know is whether w_cursor was set by the user or as a side effect of scrolling the text and moving the cursor into the visible text. More about that below.

Yeah good point I guess.

> It's hard to say what the expected behavior is when moving the cursor
and then splitting the window, especially when 'splitkeep' is not
"cursor".

> In the test the cursor is moved to line 100.  Then the window sizes are
equalized while 'slitkeep' is set to "screen".  Now does the user expect
the cursor to be still in line 100, or the text not moved up/down and
the cursor moved if needed?  I would think that the last action
overrides a previous one.  Especially if some time passes or other
commands are executed in between the two actions.

Note BTW that it is not a purely synthetic test, it is a simplified version of an issue a Neovim user reported for a "jump to definition" function that does something similar. I.e. move the cursor and resize some windows. The point of 'splitkeep' is to keep screenlines stable upon window resize. If the cursor position is set explicitly and paired with a window resize in a user function like in the test, there is no point in "stabilizing" that intermediate state as it was never seen by the user. `win_fix_scroll()` "fixing" the cursor position is then only counter productive I think, and the fact that the cursor is not in the position it was moved to is unexpected.

> The existing flags in w_valid, especially VALID_TOPLINE, are intended
for this.  Calling check_cursor_moved() to update w_valid should have
the same effect as a global flag to indicate the cursor was moved.

> We do have to make sure the flag (and w_valid_cursor) are updated when
moving the cursor for another reason (e.g. in win_fix_cursor()).  That
is the tricky part.  But not more tricky than using a global flag.

I can investigate if the w_valid flags can somehow be better used to achieve the desired effect. But I suspect it may look more like abusing the flags as I don't think they are designed with the current 'splitkeep' != "cursor" behavior in mind.

brammool:
> > There are many more ways to move the cursor. Setting a flag every
> > time the cursor is moved (by the user or by a script) is not
> > realistic.  w_valid_cursor is used to know whether the cursor moved
> > since the last time w_topline was updated. The only thing we don't
> > know is whether w_cursor was set by the user or as a side effect of
> > scrolling the text and moving the cursor into the visible text. More
> > about that below.
> 
> Yeah good point I guess.
> 
> > It's hard to say what the expected behavior is when moving the
> > cursor and then splitting the window, especially when 'splitkeep' is
> > not "cursor".
> 
> > In the test the cursor is moved to line 100.  Then the window sizes
> > are equalized while 'slitkeep' is set to "screen".  Now does the user
> > expect the cursor to be still in line 100, or the text not moved
> > up/down and the cursor moved if needed?  I would think that the last
> > action overrides a previous one.  Especially if some time passes or
> > other commands are executed in between the two actions.
> 
> Note BTW that it is not a purely synthetic test, it is a simplified
> version of an issue a Neovim user reported for a "jump to definition"
> function that does something similar. I.e. move the cursor and resize
> some windows. The point of 'splitkeep' is to keep screenlines stable
> upon window resize. If the cursor position is set explicitly and paired
> with a window resize in a user function like in the test, there is no
> point in "stabilizing" that intermediate state as it was never seen by
> the user. `win_fix_scroll()` "fixing" the cursor position is then only
> counter productive I think, and the fact that the cursor is not in the
> position it was moved to is unexpected.

The problem here is: How does Vim know that the cursor position is more
important?  Basing this on something like the two commands being short
after each other is too unpredictable.   The key point is that "resize
some windows" here is not intended to use 'splitkeep'.  Can Vim somehow
know, or does the user have to tell Vim to not use 'splitkeep'?

If 'splitkeep' would only be used for actually splitting or closing
windows, then it would work, right?  But only for this example, if a
window would be split (like when using :stag, jumping to a definition
and showing it in a new window) then the problem is there again.

I can't think of a way for Vim to know what the user's intention is.  It
looks like the only solution is that the user temporarily resets
'splitkeep'.  The luxury version would be to add a command modifier for
it.

> > The existing flags in w_valid, especially VALID_TOPLINE, are intended
> > for this.  Calling check_cursor_moved() to update w_valid should have
> > the same effect as a global flag to indicate the cursor was moved.
> 
> > We do have to make sure the flag (and w_valid_cursor) are updated when
> > moving the cursor for another reason (e.g. in win_fix_cursor()).  That
> > is the tricky part.  But not more tricky than using a global flag.
> 
> I can investigate if the w_valid flags can somehow be better used to
> achieve the desired effect. But I suspect it may look more like
> abusing the flags as I don't think they are designed with the current
> 'splitkeep' != "cursor" behavior in mind.

Right, the use of w_valid is already quite tricky, let's not make it more
complicated.

-- 
120. You ask a friend, "What's that big shiny thing?" He says, "It's the sun."

luukvbaal:
> The problem here is: How does Vim know that the cursor position is more important? Basing this on something like the two commands being short after each other is too unpredictable. The key point is that "resize some windows" here is not intended to use 'splitkeep'. Can Vim somehow know, or does the user have to tell Vim to not use 'splitkeep'? 

> If 'splitkeep' would only be used for actually splitting or closing windows, then it would work, right? But only for this example, if a window would be split (like when using :stag, jumping to a definition and showing it in a new window) then the problem is there again. 

> I can't think of a way for Vim to know what the user's intention is. It looks like the only solution is that the user temporarily resets 'splitkeep'. The luxury version would be to add a command modifier for it.

I think basing it on whether the current cursor position was actually drawn may be a valid solution that would make the test case behave like expected. I have re-opened and updated this PR with that change. I repurposed `w_last_cursorline` for this, made sure that it is always updated and thus moved the `w_p_cul` check to the appropriate places.

luukvbaal:
Hmm, the change results in some failing 'splitkeep' tests depending on the window height. I tested locally in a larger terminal and the tests passed, in a smaller terminal it fails. I think it may be resolved by adding a `:redraw` somehwere.

src/change.c
```diff
@@ -670,7 +670,7 @@ changed_common(
 	    {
 		if (xtra == 0)
 		    redraw_win_later(wp, UPD_VALID);
-		else if (lnum <= wp->w_last_cursorline)
+		else if (wp->w_p_cul && lnum <= wp->w_last_cursorline)
```
zeertzjq: Why is this condition checked again when it is already checked above?

brammool:
> > The problem here is: How does Vim know that the cursor position is
> > more important? Basing this on something like the two commands being
> > short after each other is too unpredictable. The key point is that
> > "resize some windows" here is not intended to use 'splitkeep'. Can
> > Vim somehow know, or does the user have to tell Vim to not use
> > 'splitkeep'? 
> 
> > If 'splitkeep' would only be used for actually splitting or closing
> > windows, then it would work, right? But only for this example, if a
> > window would be split (like when using :stag, jumping to a
> > definition and showing it in a new window) then the problem is there
> > again. 
> 
> > I can't think of a way for Vim to know what the user's intention is.
> > It looks like the only solution is that the user temporarily resets
> > 'splitkeep'. The luxury version would be to add a command modifier
> > for it.
> 
> I think basing it on whether the current cursor position was actually
> drawn may be a valid solution that would make the test case behave
> like expected. I have re-opened and updated this PR with that change.
> I repurposed `w_last_cursorline` for this, made sure that it is always
> updated and thus moved the `w_p_cul` check to the appropriate places.

Hmm, this changes the meaning of w_last_cursorline in a way it no longer
reflects the last line where 'cursorline' was drawn.  It may happen to
work in most cases, but it looks like it now redraws the previous line
of the cursor for no good reason.  And when toggling 'cursorline' I
suspect the display may be wrong, because w_p_cul is checked at the next
redraw, not when the 'cursorline' was drawn.  Thus it may not be
redrawn when it should.  There were bugs around this in the past.

We also have w_last_cursor_lnum_rnu for a similar reason.  Adding
another field, e.g. w_last_cursor_lnum_drawn, makes sure we don't make
it more complicated and confusing.

Anyway, I'm not convinced this is the right solution.  The user doesn't
have much control over when a redraw happens, especially for avoiding
it.  Thus adding a ":redraw" after moving the cursor can change the
scrolling behavior, but there is no way to get the opposite effect.
We don't specify what commands will cause a redraw, and we may add a
redraw somewhere for any reason.  There is the 'lazyredraw' option, but
using that to change how the display scrolls is very unexpected.

So, back to why w_topline wasn't updated before scrolling?  We have the
flags to tell us w_topline needs to be updated, that is nothing new.  I
lost track of the reason why w_topline can't be updated at the start of
win_fix_scroll().  If the window was resized or scrolled before,
w_topline should already be valid.  If the cursor was moved somehow,
then updating w_topline before applying 'splitkeep' should just work,
right?

-- 
The difference between theory and practice, is that in theory, there
is no difference between theory and practice.

luukvbaal:
> Anyway, I'm not convinced this is the right solution.  The user doesn't
have much control over when a redraw happens, especially for avoiding
it.  Thus adding a ":redraw" after moving the cursor can change the
scrolling behavior, but there is no way to get the opposite effect.

Indeed, but I don't think there is any reason to get the opposite effect. Don't you agree with the statements made in the commit message/added comment in win_fix_scroll()? 'splitkeep' "topline" and "screen" are intended to make observable screen states, before and after windows changed in size, more in line with what the user expects. As such, there is no reason for 'splitkeep' to try to "stabilize" a screen state that was never seen by the user, and doing so will only result in unexpected behavior like in the added test.

I don't think this change will have any effect on singular window commands entered by the user, but it will make chained cursor movements and window commands in user scripts behave as expected. The added `:redraw`'s in the tests only serve to simulate an observable screen state, I don't think the fact that this behavior is not reversible if a redraw already would happen matters. But I'm not totally sure either.

> So, back to why w_topline wasn't updated before scrolling?  We have the
flags to tell us w_topline needs to be updated, that is nothing new.  I
lost track of the reason why w_topline can't be updated at the start of
win_fix_scroll().  If the window was resized or scrolled before,
w_topline should already be valid.  If the cursor was moved somehow,
then updating w_topline before applying 'splitkeep' should just work,
right?

Well for one, "applying 'splitkeep'" uses the old w_botline to determine the new scroll position, which `update_topline()` may update. Other than that I think updating w_topline should be indeed be fine in principle, but adding `update_topline()` to the start of `win_fix_scroll()` (while saving the old w_botline) still results in numerous 'splitkeep' test failures. IDK if I can be bothered debugging those tests again and I think the proposed workaround should have the same result.

brammool:
Luuk van Baal wrote:

> > Anyway, I'm not convinced this is the right solution.  The user
> > doesn't have much control over when a redraw happens, especially for
> > avoiding it.  Thus adding a ":redraw" after moving the cursor can
> > change the scrolling behavior, but there is no way to get the
> > opposite effect.
> 
> Indeed, but I don't think there is any reason to get the opposite
> effect. Don't you agree with the statements made in the commit
> message/added comment in win_fix_scroll()? 'splitkeep' "topline" and
> "screen" are intended to make observable screen states, before and
> after windows changed in size, more in line with what the user
> expects. As such, there is no reason for 'splitkeep' to try to
> "stabilize" a screen state that was never seen by the user, and doing
> so will only result in unexpected behavior like in the added test.

"never seen by the user" should not change what happens.  Vim optimizes
performance by skipping redraws that will soon be overwritten by
something else.  So long as there is typeahead the redraw is often
postponed.  We don't want something else to happen depending on whether
the user types fast or slow.

> I don't think this change will have any effect on singular window
> commands entered by the user, but it will make chained cursor
> movements and window commands in user scripts behave as expected. The
> added `:redraw`'s in the tests only serve to simulate an observable
> screen state, I don't think the fact that this behavior is not
> reversible if a redraw already would happen matters. But I'm not
> totally sure either.
> 
> > So, back to why w_topline wasn't updated before scrolling?  We have
> > the flags to tell us w_topline needs to be updated, that is nothing
> > new.  I lost track of the reason why w_topline can't be updated at
> > the start of win_fix_scroll().  If the window was resized or
> > scrolled before, w_topline should already be valid.  If the cursor
> > was moved somehow, then updating w_topline before applying
> > 'splitkeep' should just work, right?
> 
> Well for one, "applying 'splitkeep'" uses the old w_botline to
> determine the new scroll position, which `update_topline()` may
> update. Other than that I think updating w_topline should be indeed be
> fine in principle, but adding `update_topline()` to the start of
> `win_fix_scroll()` (while saving the old w_botline) still results in
> numerous 'splitkeep' test failures. IDK if I can be bothered debugging
> those tests again and I think the proposed workaround should have the
> same result.

It appears that the reason we have such a hard time making this work
properly, is that we don't have a good definition of what should happen.
When 'splitkeep' is set to "screen" then the text of a window is
supposed to stay in the same place on the screen when the window is
moved or resized, and the cursor location is adjusted so that it is in a
valid position.  The help for 'splitkeep' explicitly says:

	For the "screen" and "topline" values, the cursor position will be
	changed when necessary.

But the change here is doing the opposite: it make sure the cursor stays
in the same position.  The big question is: When exactly would we use
this different behavior?

The original example was about moving the cursor and then resizing the
window, with the intention of making the text around the cursor visible.
Thus it should work like 'splitkeep' is "cursor".  But how do we know
that?  And why do something else than what the option value tells us?
Some would even call that a bug.

So, we might as well call the original example a case of wrong
expectations.  The cursor can move, because when 'splitkeep' is set to
"screen" that is what you asked for.  Expecting it to not move is
against what the option value specifies.

If the cursor must end up in a specific location, then changing the
order of commands, or temporarily setting 'splitkeep' to "cursor", can
make it work.  I do not think Vim should try to guess what the user
wants and do something that goes against the option values.

-- 
Statistics say that you can have a baby per month on average:
Just get nine women pregnant.

luukvbaal:
Perhaps you're right, having to work around 'splitkeep' might be annoying from the perspective from plugin authors though. I still suspect there is no harm in making sure that 'splitkeep' doesn't handle screen states that weren't drawn in the first place.

PR #12490, opened 2023-06-04
Fix wrong display when scrolling multiple lines with 'smoothscroll'

zeertzjq:
Why only a w_skipcol > 0 check is enough:
- If w_topline didn't change, w_skipcol must have increased.
- If w_topline did change and the new w_skipcol is 0, wl_size of the new
  w_topline is valid, so a redraw is not needed.
- If w_topline did change and the new w_skipcol is non-zero, wl_size of
  the new w_topline is invalid, so a redraw is needed anyway.


brammool:
> Why only a w_skipcol > 0 check is enough:
> - If w_topline didn't change, w_skipcol must have increased.
> - If w_topline did change and the new w_skipcol is 0, wl_size of the new
>   w_topline is valid, so a redraw is not needed.

How do you know wl_size is valid?  Doesn't that depend on what the new
w_topline is?  This would also depend on wl_valid to be FALSE when going
further down.

> - If w_topline did change and the new w_skipcol is non-zero, wl_size of
>   the new w_topline is invalid, so a redraw is needed anyway.

Isn't it that when w_topline changes a redraw will happen anyway?
But it may only scroll then, thus the redraw method must be set to
UPD_NOT_VALID to recompute more,

It's all quite tricky, let's try to add relevant comments and stay on
the safe side.

-- 
117. You are more comfortable typing in html.

zeertzjq:
Oh, maybe "valid" is the wrong word here. What I wanted to mean is that then w_topline increases (it cannot decrease when scrolling up) and the new w_skipcol is zero, the old w_topline is hidden whether the previous w_skipcol is zero or not, so this is a situation that code in drawscreen.c can already handle.

zeertzjq:
I've change the PR to redraw when one of the previous skipcol and current skipcol is non-zero.

PR #12503, opened 2023-06-08
Make charidx() and utf16idx() return value consistent with byteidx() return value

yegappan:
The byteidx() and byteidxcomp() functions return the length of a string in bytes when the specified character
index is equal to the number of characters in the string.  But the charidx() function returns -1 when the specified
byte index is equal to the number of bytes in the string.  Same behavior applies to the utf16idx() function also.
Make the return value of these two functions consistent with the byteidx() function.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1528,11 +1528,12 @@ charidx({string}, {idx} [, {countcc} [, {utf16}]])
 		When {utf16} is present and TRUE, {idx} is used as the UTF-16
 		index in the String {expr} instead of as the byte index.
 
-		Returns -1 if the arguments are invalid or if {idx} is greater
-		than the index of the last byte in {string}.  An error is
-		given if the first argument is not a string, the second
-		argument is not a number or when the third argument is present
-		and is not zero or one.
+		Returns -1 if the arguments are invalid or if there are less
+		than {idx} bytes. If there are exactly {idx} bytes the length
+		of the string in characters is returned.  An error is given if
+		the first argument is not a string, the second argument is not
+		a number or when the third argument is present and is not zero
+		or one.
 
 		See |byteidx()| and |byteidxcomp()| for getting the byte index
 		from the character index and |utf16idx()| for getting the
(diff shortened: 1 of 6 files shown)


ychin:
May I ask what the motivation is for such a change? This seems counter to any programming language libraries that I'm aware of. You will end up returning an "index" that's clearly out of bounds and seems to actively encourage unsafe programming where you now have to check for length *or* -1 to tell if the result is invalid. Maybe I'm missing something by it seems more like byteidx is the undesired behavior here.

ychin:
I guess this is to make it so that you can do something like `byteidx(charidx(…))` and vice versa and have it return itself? And you are essentially using `charidx` double duty as a fictional `strcharlen()` function (since `strlen()` returns bytes and we don't have a function to return number of Unicode characters in a string)?

vim-ml:
Hi,

On Tue, Jun 20, 2023 at 2:04 PM Yee Cheng Chin ***@***.***>
wrote:

> May I ask what the motivation is for such a change? This seems counter to
> any programming language libraries that I'm aware of. You will end up
> returning an "index" that's clearly out of bounds and seems to actively
> encourage unsafe programming where you now have to check for length *or*
> -1 to tell if the result is invalid. Maybe I'm missing something by it
> seems more like byteidx is the undesired behavior here.
>
>
> A LSP server can return a character position which is at the end of a line
(equal to the number
of characters in the line).  This happens, for example, when doing
insert-mode completion at the
end of a line.  In this case, if charidx() is used to translate the
returned LSP position to a character
position, it returns -1.  So the plugin needs to check for this case and
then use strchars() and use
the return value as the character position:

https://github.com/yegappan/lsp/blob/main/autoload/lsp/offset.vim#L55

With this PR, the charidx() function returns the number of characters in
the line if the supplied index
is equal to the string length and the plugin can directly use the return
value avoid an additional
call to strchars().

Regards,
Yegappan

vim-ml:
Hi,

On Tue, Jun 20, 2023 at 2:08 PM Yee Cheng Chin ***@***.***>
wrote:

> I guess this is to make it so that you can do something like
> byteidx(charidx(…)) and vice versa and have it return itself? And you are
> essentially using charidx double duty as a fictional strcharlen()
> function (since strlen() returns bytes and we don't have a function to
> return number of Unicode characters in a string)?
>
>
>
A LSP server can encode a position offset using UTF-8 or UTF-16 or UTF-32
encoding as described here:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments
.

A Vim LSP plugin needs to decode and encode the character position in a
file to the corresponding
encoding supported by the LSP server.  For example, see the functions in
the following file:

https://github.com/yegappan/lsp/blob/main/autoload/lsp/offset.vim

This PR helps in simplifying some of the code in the above functions.

Regards,
Yegappan

PR #12504, opened 2023-06-08
Fix multiple errors when passing wrong variable types to an option

zeertzjq:
Problem:    Passing a wrong variable type to an option gives multiple
	    errors.
Solution:   Bail out early when getting the new value fails.


zeertzjq:
Problem:    Passing a wrong variable type to an option gives multiple
	    errors.
Solution:   Bail out early when getting the new value fails.

PR #12505, opened 2023-06-08
Add filetype t32

xasc:
TRACE32 is a debug tool for embedded systems. It includes a [scripting language](https://repo.lauterbach.com/pdf/practice_user.pdf) for automation purposes. Typically, scripts have the file extension _CMM_. The suffix _T32_ is used for auto-generated files with configuration settings.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2234,6 +2234,9 @@ au BufNewFile,BufRead *.toml			setf toml
 " TPP - Text Presentation Program
 au BufNewFile,BufRead *.tpp			setf tpp
 
+" TRACE32 Script Language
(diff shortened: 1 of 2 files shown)


dkearns:
Is there a reason to not name the file type trace32?  It would seem a lot clearer though there may be some subtlety I'm missing.

xasc:
> Is there a reason to not name the file type trace32? It would seem a lot clearer though there may be some subtlety I'm missing.

This is a valid point. "T32" is the official short name and is frequently used as prefix for related utilities. With that in mind, I would say that users are likely to make that association. However, I fully agree that it might confuse people that are not familiar.

Issue #12506, opened 2023-06-08
Cursor movement prevented when 'splitkeep' is not "cursor" and 'scrolloff'


luukvbaal:
### Steps to reproduce

1. Write test.vim with the following lines:
```vim
set splitkeep=screen
let &scrolloff = &lines / 4
call setline(1, range(1, 200))
wincmd s
au CursorMoved * windo echo
```
2. Run `vim --clean -S test.vim`
3. Try to move the cursor down with "j"

The cursor does not move because 'splitkeep' "fixes" the cursor position.
Any autocmd on CursorMoved that calls `win_enter_ext()` which in turn calls `win_fix_cursor()` would trigger this issue if 'scrolloff' is larger than half the window height.

(Original Neovim issue: https://github.com/neovim/neovim/issues/23956)

### Expected behaviour

The cursor is moved down.

Technically this is simply a 'scrolloff' roundoff error which might be handled directly in `win_fix_cursor()`, not sure though. Making sure that 'splitkeep' does not try to fix cursor positions that were not drawn in the first place like proposed in https://github.com/vim/vim/pull/12488 would also prevent this though (PR would need to be adjusted slightly).

### Version of Vim

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jun  8 2023 18:35:01) Included patches: 1-1601

### Environment

Operating System: Arch Linux
Terminal: st
TERM: st-256color
SHELL: zsh

luukvbaal:
Changing the repro steps to:
1. Write test.vim with the following lines:
```vim
set splitkeep=screen scrolloff=0
call setline(1, range(1, 200))
wincmd s
au CursorMoved * windo echo
```
2. Run `vim --clean -S test.vim`
3. Try to move the cursor to the end of the buffer with "G"

results in a related problem where "fixing" the cursor does not actually bring the cursor to the end of the buffer even without 'scrolloff' (which the proposed fix would also prevent).

luukvbaal:
Fixed in patch 9.0.1772.

PR #12509, opened 2023-06-09
Cursor is adjusted in window that did not change in size by 'splitkeep'

luukvbaal:
Problem:    Cursor is adjusted in window that did not change in size by
            'splitkeep'.
Solution:   Only check that cursor position is valid in a window that
            has changed in size.

I believe this would have been a valid patch even if it didn't fix any problems. It also happens to fix some of the problems outlined in #12506 and #12488 where windows didn't actually change in size. `win_fix_cursor()` shouldn't have to execute in that case, and the fact that it did resulted in problems.

Cursor movement paired with actual window resizes in user scripts could still be seen as behaving unexpectedly but perhaps there is no way around that since it is working according to the option description for 'splitkeep'. It may be avoided by changing the order of the cursor/resize commands.


brammool:
> Problem:    'splitkeep' may adjust cursor in window that has not changed
>             in size.
> Solution:   Only check that cursor position is valid in a window that
>             has changed in size.
> 
> I believe this change could have been made even if it didn't fix any
> problems but it happens to fix some of the problems outlined in #12506
> and #12488. Cursor movement paired with actual window resizes still
> could be seen as behaving unexpectedly but perhaps there is no way
> around that since it is working according to the option description
> for 'splitkeep'.

I don't quite understand what the w_splitkeep_cursor field means:

	int	w_splitkeep_cursor; // if TRUE cursor may be invalid

It appears to only cause win_fix_cursor() to bail out when it is not
set.  Perhaps w_do_win_fix_cursor would be a better name?


-- 
Fingers not found - Pound head on keyboard to continue.

luukvbaal:
> I don't quite understand what the w_splitkeep_cursor field means:

>	int	w_splitkeep_cursor; // if TRUE cursor may be invalid

> It appears to only cause win_fix_cursor() to bail out when it is not
set.  Perhaps w_do_win_fix_cursor would be a better name?

Correct, it is just a boolean that is set for resized windows in `win_fix_scroll()`, and `win_fix_cursor()` returns early when it is not set.

PR #12510, opened 2023-06-09
Update fennel keywords

gpanders:
New special forms added:

* case
* case-try
* faccumulate
* fcollect


brammool:
I'll include it, thanks.

PR #12511, opened 2023-06-09
termdebug: frame related commands

GitMensch:
partially fixing #10393

... its been a while that this was setting around. I do hope someone may take this as an option to get the commands already into termdebug before I try to add a window - as that is complicated and we have no finished design for that.

I do know that this PR is missing something: the help text - if possible please add it (as the PRs are not merged "directly" I think that's fine). Also I did not have another look at the FIXME note in the Frames command.

One thing I'd do suggest is to add a default mapping for Down as "D" and Up as "d". But as - again - only K is default mapped I was not sure if that's "too intrusive".

Thanks for considering this contribution!

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1376,6 +1376,21 @@ This is similar to using "print" in the gdb window.
 You can usually shorten `:Evaluate` to `:Ev`.
 
 
+Navigation in the Stack ~
+					*termdebug-variables* *:Frame*
+ `:Frame`	Select the given frame, using  either the frame's
+                stack number, address, or function name.
+ `:Up`	        Select the frame that called the current one with an
+                optional argument to say how many frames to go up.
+ `+`		same (see |termdebug_map_plus| to disable)
+ `:Down`	        Select the frame called by the current one with
+                an optional argument to say how many frames to go down.
+ `-`		same (see |termdebug_map_minus| to disable)
+
+This is similar to using "print" in the gdb window.
+You can usually shorten `:Evaluate` to `:Ev`.
+
+
 Other commands ~
 							*termdebug-commands*
  *:Gdb*	     jump to the gdb window
@@ -1453,6 +1468,14 @@ The K key is normally mapped to :Evaluate. If you do not want this use: >
 	let g:termdebug_config['map_K'] = 0
 If there is no g:termdebug_config you can use: >
 	let g:termdebug_map_K = 0
+<
+						*termdebug_map_minus*
+The - key is normally mapped to :Down. If you do not want this use: >
+	let g:termdebug_config['map_minus'] = 0
+<
+						*termdebug_map_plus*
+The + key is normally mapped to :Up. If you do not want this use: >
+	let g:termdebug_config['map_plus'] = 0
 <
 						*termdebug_disasm_window*
 If you want the Asm window shown by default, set the "disasm_window" flag to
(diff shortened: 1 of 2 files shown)


brammool:
> partially fixing #10393
> 
> ... its been a while that this was setting around. I do hope someone
> may take this as an option to get the commands already into termdebug
> before I try to add a window - as that is complicated and we have no
> finished design for that.

Assuming that a few people will try out the patches and improvements are
made, it's probably easier to just have one patch.  Otherwise merge
conflicts will make it harder to include both.

Just to be clear: with "window" you mean a Vim window that displays the
current stack, like what "where" would output.  With an indication of
the current location, if not at the top.  The window would be updated
whenever the program is stopped.  Commands in the window and clicking
with the mouse could be used to select a certain frame, like what "frame
123" would do in the gdb window.

> I do know that this PR is missing something: the help text - if
> possible please add it (as the PRs are not merged "directly" I think
> that's fine). Also I did not have another look at the FIXME note in
> the Frames command.

Please do add help text early.  Not only to make it easy for users to
try it out, it also helps checking that it works in a way it's easy to
explain to someone.  If writing the help text is hard, then perhaps the
code needs to be adjusted to make the usage simpler.

> One thing I'd do suggest is to add a default mapping for Down as "D"
> and Up as "d". But as - again - only K is default mapped I was not
> sure if that's "too intrusive".

The frame "up" and "down" commands are probably used most often.  Using
uppercase would be "U" and "D", but both are regularly used commands,
thus that gets confusing.  Perhaps using "+" and "-" is better?  They
are regular commands, but personally I don't use them, thus I would be
OK with mapping them.

I have several other things to work on, thus I'll await what you and
others can come up with, possibly until it's ready to include.

-- 
139. You down your lunch in five minutes, at your desk, so you can
     spend the rest of the hour surfing the Net.

GitMensch:
> Just to be clear: with "window" you mean a Vim window that displays the
current stack, like what "where" would output.  With an indication of
the current location, if not at the top.  The window would be updated
whenever the program is stopped.

Yes

> Commands in the window and clicking
with the mouse could be used to select a certain frame, like what "frame
123" would do in the gdb window.

Sounds cool - that would be the case if you (or someone else) add it - I only can do the "view" part and the code for up/down/number - as that is what the `Frame` command (currently in this PR) does already.

> The frame "up" and "down" commands ... using "+" and "-" is better ...  I would be OK with mapping them.

OK, I'll do that.

> Assuming that a few people will try out the patches and improvements are made, it's probably easier to just have one patch. 

I'll add the help text and the mapping for this PR, because the window is really "a separate" thing from the commands and "the way the window is created/positioned" is the part that's not clear yet - so we can add it with a later PR.

chrisbra:
any update here please @GitMensch ?

GitMensch:
@chrisbra Implemented with help and default mapping as requested. Leaving the window to a separate PR to be re-implemented like `:Var`.

yegappan (approved):

GitMensch:
Is there an actual option to test-coverage those changes?

GitMensch:
> Currently there is no test coverage for the termdebug plugin. But it will be good to add some test coverage though.

Sounds good, moved that to a new issue.

As I'm new to this changed scenario - how will merges be done (as before "in a bunch", then committed with a new patch number)?

yegappan:
Christian is merging the runtime file changes directly (without a patch number).  This way the authors of the runtime file changes
get credit for their changes while at the same time there is no proliferation of patches.  The patch number is used only for the
core source changes (src and testdir).  So for any runtime file changes open a PR.

PR #12513, opened 2023-06-10
Add functions to get a character index with and without counting combining characters

yegappan:
Many of the Vim functions like setcursorcharpos() accept a character index which doesn't count the combining characters as
separate characters.  The LSP servers count combining characters as separate characters.  So to use the character position values
returned by the LSP servers, the Vim plugins need to use the byteidxcomp() and the charidx() functions to translate the character
index.

Add two new functions charidx_addcc() and charidx_dropcc() to translate a character index to either include or exclude
combining characters.


ychin:
I'm a little confused. There's another function called `charidx` which does take a `countcc` parameter. How is this different? And if they are different I was having trouble understanding it from the docs.

FWIW it seems like we are adding a lot of new definitions of "length" and "index" in the API but in a little bit of a patchwork fashion, and sometimes the terminology is a little confusing to understand. Just wanted to throw this out there for how some other languages handle this and ways to query the 3-4 definitions of "length" ( 1: number of bytes/UTF-8 code points, 2: number of UTF-16 units, 3: number of unique Unicode code points / "count composing charter").

ychin:
Sorry forgot to include link: https://hsivonen.fi/string-length/

chrisbra:
@yegappan do you think this is still useful?

chrisbra:
closing for now.

PR #12516, opened 2023-06-10
Fix typo in documents

h-east:


brammool:
I'll include it, thanks.  I'll make some more changes in the 2html text, having the variable names as tags right-aligned does not work well as a header.

PR #12519, opened 2023-06-11
Fix expanding pattern interfering with cmdline completion

zeertzjq:


src/cmdexpand.c
```diff
@@ -985,7 +985,8 @@ ExpandOne(
 	if (compl_match_array != NULL)
 	    cmdline_pum_remove();
     }
-    findex = 0;
+    if (mode != WILD_EXPAND_FREE && mode != WILD_ALL && mode != WILD_ALL_KEEP)
```
zeertzjq, line 988: This condition may not make 100% sense. It's hard to understand what this `findex = 0` is intended for in the first space.

brammool: "findex" is used with WILD_NEXT, WILD_PAGEUP, etc. to store the current position.
Having these as static variables is weird.  I think "findex" should be part of expand_T.
This fix might not be needed then.
But we can first include this fix and refactor it later.

PR #12522, opened 2023-06-11
Termdebug remote mode switch

EdwarDu:
To add a switch to allow Termdebug to send `continue` instead of `run` when doing `:Termdebug!`.

Use case:  when need to connect to a remote gdb server automatically though command line to launch Termdebug, it is possible to start Termdebug and manually do `target remote ...`, but it is annoying to do that especially if it needs to be repeated for several times.
So, this can be put to a vim file, e.g. remote_termdebug.vim, then can launch vim as `vim -c 'source remote_termdebug.vim' -c 'exe :Termdebug!'`
```
let Termdebug_remote={cmd, pty -> [join(cmd), "-q", 
\ "-tty", pty,
\ "-l", 120,
\ '-iex', "set pagination off", 
\ '-iex', "set mi-async on",
\ '-iex', "set sysroot /",
\ '-iex', "set tcp auto-retry on",
\ '-iex', "set tcp connect-timeout 120",
\ '-ex', "echo Waiting for gdbserver connection ".g:gdb_port." (2mins) ...\n",
\ '-ex', "target remote localhost:".g:gdb_port,
\ g:gdb_x_opt, g:gdb_x_opt_value,
\ '-ex', "echo startupdone\n"]}

let g:termdebug_config={
\ "command": 'gdb',
\ "command_add_args": Termdebug_remote,
\ "remotemode": 1,
\ "wide": 320}
```

Use `command_add_args` to "hijack" Termdebug gdb launch commands. This does not work with `use_prompt: 1` though.
`g:gdb_x_opt, g:gdb_x_opt_value` can be used to load extra file containing gdb command such as set up breakpoints etc. 

It would be much nicer if all can be done with `:Termdebug <host>:<port>`, but that probbaly requries much more changes and it is over my current capability.


EdwarDu:
I may created the pull requests including a merge commit by mistake, don't know how to remove that, sorry.

brammool:
> I may created the pull requests including a merge commit by mistake,
> don't know how to remove that, sorry.

Don't worry, individual commits rarely matter, I usually just look at
the final result.

-- 
192. Your boss asks you to "go fer" coffee and you come up with 235 FTP sites.

brammool:
> To add a switch to allow Termdebug to send `continue` instead of `run`
> when doing `:Termdebug!`.

This sounds like a useful addition.  I do have a few remarks.

Please also update the help.  Otherwise it will be hard to understand
how this works.  Or that the support even exists.  The examples you have
given here are a good start.

There is no need for "g:termdebug_remotemode".  The other
"g:termdebug_..." variables exist for backwards compatibility before
g:termdebug_config was added.  All new config options should go into
g:termdebug_config to avoid polluting the global namespace.

For using "=" in an assignment, please surround it with spaces, so that
it is easier to spot the separation between LHS and RHS.

Can "continue" be send with s:SendResumingCommand() ?  Then s:stopped
will be updated properly.  And it will work with a prompt buffer.

> Use `command_add_args` to "hijack" Termdebug gdb launch commands. This
> does not work with `use_prompt: 1` though.

It also makes it difficult to add more arguments.

> `g:gdb_x_opt, g:gdb_x_opt_value` can be used to load extra file
> containing gdb command such as set up breakpoints etc. 

This should be explained in the help.  When used in an example, there
should be a check if the variables exist.

> It would be much nicer if all can be done with `:Termdebug
> <host>:<port>`, but that probbaly requries much more changes and it is
> over my current capability.

It also conflicts with the current use of the argument, such as "core"
or "processID".  Adding a new command avoids confusion:

	TermdebugRemote {host}:{port}

Would that work?

-- 
191. You rate eating establishments not by the quality of the food,
     but by the availability of electrical outlets for your PowerBook.

Issue #12528, opened 2023-06-12
Divide-by-zero vulnerability in function `scroll_cursor_bot`


fullwaywang:
* I reported this vulnerability on huntr.dev but had not got a response for a week, so I think maybe disclosing it here is easier to keep in touch. 

# Description
Recently I have been reviewing CVEs, checking whether there are cognate/recurring bugs as known ones. While reviewing VIM, I found one quite similar to CVE-2023-0512.

The CVE-2023-0512 is an issue where VIM under ex mode falsely caculated the width of current window. In `scroll_cursor_bot` caculates the `topline` and `bottomline` likely, only in normal visual mode. In the [caculation](https://github.com/vim/vim/blob/master/src/move.c#L2593), `width1` is assigned the width of current window minus widths of linenumber and foldcolumn, which can be negative. If `foldcolumen` is on, then `width2` can be assigned `0`, resulting in a Divide-by-zero fault in the following division.

# Version of Vim

>9.0.0908

# Environment

CentOS 8 stream

# Logs and stack traces

```shell
/data/project/exp/build/cloned/vim
➜ cat /tmp/vim_test.data
silent normal yy
silent normal 19p
wi0 19
vsplit
vertical resize 0
set foldcolumn=1
set number
set smoothscroll
silent normal G

/data/project/exp/build/cloned/vim
➜ src/vim -u NONE -i NONE -n -m -X -Z -S /tmp/vim_test.data -c :qa!
Vim: Caught deadly signal FPE
Vim: Finished.
[1]    3037351 floating point exception (core dumped)  src/vim -u NONE -i NONE -n -m -X -Z -S /tmp/vim_test.data -c
                                                                                                                                   %
```

GDB:
```shell
/data/project/exp/build/cloned/vim
➜ gdb src/vim
Reading symbols from src/vim...
(gdb) set args -u NONE -i NONE -n -m -X -Z -S /tmp/vim_test.data
(gdb) r
           i
d argument:
wfw=0  38:
E474: Program received signal SIGFPE, Arithmetic exception.
                                                           0x0000000000526911 in scroll_cursor_bot (min_scroll=1, set_topbot=0) at move.c:2593
2593			skip_lines += (curwin->w_skipcol - width1) / width2 + 1;
(gdb) list
2588		    int skip_lines = 0;
2589		    int width1 = curwin->w_width - curwin_col_off();
2590		    int width2 = width1 + curwin_col_off2();
2591		    // similar formula is used in curs_columns()
[...]

PR #12529, opened 2023-06-12
Fix divide-by-zero vulnerability in `scroll_cursor_bot`

fullwaywang:
Fix issue #12528


brammool:
> Fix issue #12528 

Thanks for making a PR.  It doesn't work though:

    move.c:2594:10: error: use of undeclared identifier 'width'; did you mean 'width1'?
            if (width > 0)

You could reproduce the problem in #12528.  Can you please use that to
add a regression test?  A few hints:
	- The text file can easily be generated.
	- The commands starting with "sil0norm0" can most likely be
	  written as "silent norm ".  That is a lot easier to read.
	- ":set co=2" makes the Vim window 2 columns wide, that won't
	  work.  The minimum value is 12.

The test can probably be simplified, try to drop some commands.

I am lagging behind on problems reported on huntr.dev.  Mainly because
the reproduction steps are often random (result of running a fuzzer),
making it both difficult to find out what is happening and making it
difficult to turn this into a useful regression test.

-- 
161. You get up before the sun rises to check your e-mail, and you
     find yourself in the very same chair long after the sun has set.

fullwaywang:
> Fix issue #12528
> Thanks for making a PR. It doesn't work though: move.c:2594:10: error: use of undeclared identifier 'width'; did you mean 'width1'? if (width > 0) You could reproduce the problem in #12528. Can you please use that to add a regression test? A few hints: - The text file can easily be generated. - The commands starting with "sil0norm0" can most likely be written as "silent norm ". That is a lot easier to read. - ":set co=2" makes the Vim window 2 columns wide, that won't work. The minimum value is 12. The test can probably be simplified, try to drop some commands. I am lagging behind on problems reported on huntr.dev. Mainly because the reproduction steps are often random (result of running a fuzzer), making it both difficult to find out what is happening and making it difficult to turn this into a useful regression test.
> […](#)
> -- hundred-and-one symptoms of being an internet addict: 161. You get up before the sun rises to check your e-mail, and you find yourself in the very same chair long after the sun has set. /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net \\\ /// \\\ \\\ sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ /// \\\ help me help AIDS victims -- http://ICCF-Holland.org ///

Sorry my bad : ( It's a typo, should be `width1` instead.

The issue in #12528 has been edited. A simplified commands sequence is given (the former one contained junk data while trying to trigger the bug; the new one contains only 7 necessary commands; `set co=2` is unnecessary). And the text file is rather simple, just 20 empty lines so the moving of curson can trigger the bug.

Please try the new PoC and let me know if it can be reproduced there.

As for a regression testcase, I'm sorry that I did not get very familiar with the testing configuration in this project, since VIM is just one among many open source projects I found recurring vulnerabilities which I have been reporting issues and simple descriptions. So maybe you could please add those `testdir` tests for this case given the simplified PoC.

fullwaywang:
Made a new PR with clean commits and accompanied test files. Close this.

PR #12534, opened 2023-06-13
lang: Update Japanese translation

k-takata:
ref: https://github.com/vim-jp/lang-ja/releases/tag/20230613


brammool:
I'll include it, thanks.

PR #12535, opened 2023-06-13
Fix syntaxtest

k-takata:
Especially on FreeBSD.

It failed like this: https://github.com/vim/vim/pull/12534/checks?check_run_id=14218242478

```
�[?25hcd runtime/syntax &&  make clean &&  make test VIM="../../src/vim"
rm -f testdir/failed/* testdir/done/* testdir/vimcmd
VIMRUNTIME=../.. vim --clean --not-a-term -u testdir/runtest.vim
/bin/sh: vim: not found
*** Error code 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ VIM_FOR_SYNTAXTEST = ../../src/vim
 syntaxtest:
 	cd runtime/syntax && \
 		$(MAKE) clean && \
-		$(MAKE) test VIM="$(VIM_FOR_SYNTAXTEST)"
+		$(MAKE) test VIMPROG="$(VIM_FOR_SYNTAXTEST)"
 
 
 #########################################################################


k-takata:
Hmm, it seems that this change is not enough.

In L32 in runtime/syntax/testdir/runtest.vim, `cd testdir` is executed.
VIMPROG must be relative to runtime/syntax/testdir/ after that line. It causes a mismatch.

brammool:
> Hmm, it seems that this change is not enough.
> 
> In L32 in runtime/syntax/testdir/runtest.vim, `cd testdir` is executed.
> VIMPROG must be relative to runtime/syntax/testdir/ after that line.
> It causes a mismatch.

Right, $VIMPROG is used in two directories, thus after "cd testdir"
another "../" needs to be prefixed.  I have added this and it works for
me.  But when $VIMROG is just "vim" using "../vim" won't work.  Well,
let's leave that for later.

-- 
164. You got out to buy software, instead of going out for a beer.

PR #12536, opened 2023-06-13
Fix make clean/make distclean/make testclean from top level

puremourning:
Problem: make distclean, clean, testclean fail from top level
Solution: restore the cwd after cleaning inside runtime directories.

Previously, the embedded script would:

* cd to runtime/indent
* run make clean
* attempt to cd to runtime/syntax, failing (as runtime/indent/runtime/syntax doesn't exit)

Solution is to run the cd & make in a sub shell.

Also considered this:

```
	@if test "$@" = "clean" -o "$@" = "distclean" -o "$@" = "testclean"; then \
		$(MAKE) -C runtime/indent clean; \
		$(MAKE) -C runtime/syntax clean; \
	fi
```

But wasn't sure if `make -C` is universally supported, and avoided it given Vim's makefiles don't appear to rely on it in other places.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,10 +46,8 @@ all install uninstall tools config configure reconfig proto depend lint tags typ
 	fi
 	@# When the target is "clean" also clean for the indent and syntax tests.
 	@if test "$@" = "clean" -o "$@" = "distclean" -o "$@" = "testclean"; then \
-		cd runtime/indent && \
-			$(MAKE) clean; \
-		cd runtime/syntax && \
-			$(MAKE) clean; \
+		(cd runtime/indent && $(MAKE) clean); \
+		(cd runtime/syntax && $(MAKE) clean); \
 	fi
 
 # Executable used for running the indent tests.


puremourning:
Hmm. The tests failed on FreeBSD, with suspicious-looking problem:

```
�[?25hcd runtime/syntax &&  make clean &&  make test VIM="../../src/vim"
rm -f testdir/failed/* testdir/done/* testdir/vimcmd
VIMRUNTIME=../.. vim --clean --not-a-term -u testdir/runtest.vim
/bin/sh: vim: not found
*** Error code 127
```

however, this looks to be `make indenttest` which should be unaffected by this change (ditto `make syntaxtest`) unless there's another issue with CI. I'll see if I can work it out.

puremourning:
That failure seems to be because runtime/syntax/Makefile expects the "vim" path to be specified in `VIMPROG`, but I'd actually specified in `VIM`:

```
syntaxtest:
	cd runtime/syntax && \
		$(MAKE) clean && \
		$(MAKE) test VIM="$(VIM_FOR_SYNTAXTEST)"

```

Should probably be more like:

```
syntaxtest:
	cd runtime/syntax && \
		$(MAKE) clean && \
		$(MAKE) test VIM="$(VIM_FOR_SYNTAXTEST)" VIMRUNTIME=..

```

brammool:
> Hmm. The tests failed on FreeBSD, with suspicious-looking problem:
> 
> ```
> �[?25hcd runtime/syntax &&  make clean &&  make test VIM="../../src/vim"
> rm -f testdir/failed/* testdir/done/* testdir/vimcmd
> VIMRUNTIME=../.. vim --clean --not-a-term -u testdir/runtest.vim
> /bin/sh: vim: not found
> *** Error code 127
> ```
> 
> however, this looks to be `make indenttest` which should be unaffected
> by this change (ditto `make syntaxtest`) unless there's another issue
> with CI. I'll see if I can work it out.

The mixup of $VIM and $VIMPROG was fixed in patch 9.0.1628.
I'm not totally happy with how it ended up, it will now fail when using
"vim" (without a path).

-- 
"Hit any key to continue" is very confusing when you have two keyboards.

PR #12542, opened 2023-06-15
fix rust indent

glepnir:
1. new line indent after `struct *** <T>` should same as prev
2. new line after `where` level should add one.

### DIFF
--- a/runtime/indent/rust.vim
+++ b/runtime/indent/rust.vim
@@ -132,6 +132,22 @@ function GetRustIndent(lnum)
 		return indent(prevlinenum) + 6
 	endif
 
+	"match newline after struct with generic bound like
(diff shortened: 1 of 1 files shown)


brammool:
The file header mentions:  For bugs, patches and license go to https://github.com/rust-lang/rust.vim

glepnir:
okay if rust.vim merge then patch to vim？

chrisbra:
Usually the maintainer decides when to submit a new version for inclusion with Vim. It usually depends on how he assesses the current version of runtime file. If he thinks the version is good enough, he will likely submit it to Bram then. 

However, the current version seems to be 6 years old, so let me ping @chris-morgan to send an update soon ;)

glepnir:
thanks :)

PR #12544, opened 2023-06-15
feat: added zserio syntax file

dpelle:


dkearns:
Should the syntax group name prefix be "zserio" rather than just "zs"?  The prefix is usually the filetype name but there's probably exceptions and I'm not sure if it's required.

Just something I noticed...

DominiquePelle-TomTom:
@dkearns wrote:

> Should the syntax group name prefix be "zserio" rather than just "zs"? The prefix is usually the filetype name but there's probably exceptions and I'm not sure if it's required.
> 
> Just something I noticed...

Thanks.  I just changed the prefix of syntax groups to "zserio".

brammool:
> @dkearns wrote:
> 
> > Should the syntax group name prefix be "zserio" rather than just
> > "zs"? The prefix is usually the filetype name but there's probably
> > exceptions and I'm not sure if it's required.
> > 
> > Just something I noticed...
> 
> Thanks.  I just changed the prefix of syntax groups to "zserio".

Good, this naming we can do consistently.  It avoids any name clashes in
case the user wants to define specific highlighting for a syntax file.

-- 
207. You're given one phone call in prison and you ask them for a laptop.

PR #12548, opened 2023-06-16
Move "findex" static variable to expand_T

zeertzjq:
Name it "xp_selected", which is similar to "compl_selected".


zeertzjq:
Note: this is a followup to this comment: https://github.com/vim/vim/pull/12519#discussion_r1231564658
> "findex" is used with WILD_NEXT, WILD_PAGEUP, etc. to store the current position.
Having these as static variables is weird. I think "findex" should be part of expand_T.
This fix might not be needed then.
But we can first include this fix and refactor it later.

PR #12557, opened 2023-06-17
Remove backslashes in man pages using Vim script

### LINKED ISSUE #12301: :Man for any manpage opens an empty window
### Steps to reproduce

```
1. export MANPAGER=vim +MANPAGER --not-a-term -
2. man man
3. Vim opens with the man page for 'man'
4. :Man ls
```

https://user-images.githubusercontent.com/2244241/234430818-33932dae-747d-41c5-8ab0-bfd0bc7cb7a4.mov

### Expected behaviour

It should open a new empty buffer in the same window and read in the manpage for `ls`

I've printed out the command that is being executed and did some debugging:
`env -u MANPAGER GROFF_NO_SGR=1 man 'ls' | col -b`
1. removing silent shows "shell returned 1"
2. removing `| col -b` and reading the command in displays the contents of the man page for `ls`

### Version of Vim

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 24 2023 18:31:22) Included patches: 1-1486

### Environment

```
Linux 5.15.102-1-pve #1 SMP PVE 5.15.102-1 (2023-03-14T13:48Z) x86_64 GNU/Linux
$SHELL = zsh 5.8 (x86_64-debian-linux-gnu)
Terminal = iTerm2
$TERM = screen-256color
```

> VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 24 2023 18:31:22)
Included patches: 1-1486
Compiled by tart@lemontart
Huge version without GUI.  Features included (+) or not (-):
+acl               +file_in_path      +mouse_urxvt       -tag_any_white
+arabic            +find_in_path      +mouse_xterm       -tcl
+autocmd           +float             +multi_byte        +termguicolors
+autochdir         +folding           +multi_lang        +terminal
-autoservername    -footer            -mzscheme          +terminfo
-balloon_eval      +fork()            +netbeans_intg     +termresponse
+balloon_eval_term -gettext           +num64             +textobjects
-browse            -hangul_input      +packages          +textprop
++builtin_terms    +iconv             +path_extra        +timers
+byte_offset       +insert_expand     -perl              +title
+channel           +ipv6              +persistent_undo   -toolbar
+cindent           +job               +popupwin          +user_commands
-clientserver      +jumplist          +postscript        +vartabs
-clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +vim9script
+cmdline_hist      +langmap           -python            +viminfo
+cmdline_info      +libcall           -python3           +virtualedit
+comments          +linebreak         +quickfix          +visual
+conceal           +lispindent        +reltime           +visualextra
+cryptv            +listcmds          +rightleft         +vreplace
+cscope            +localmap          -ruby              +wildignore
+cursorbind        -lua               +scrollbind        +wildmenu
+cursorshape       +menu              +signs             +windows
+dialog_con        +mksession         +smartindent       +writebackup
+diff              +modify_fname      -sodium            -X11
+digraphs          +mouse             -sound             -xfontset
-dnd               -mouseshape        +spell             -xim
-ebcdic            +mouse_dec         +startuptime       -xpm
+emacs_tags        -mouse_gpm         +statusline        -xsmp
+eval              -mouse_jsbterm     -sun_workshop      -xterm_clipboard
+ex_extra          +mouse_netterm     +syntax            -xterm_save
+extra_search      +mouse_sgr         +tag_binary
-farsi             -mouse_sysmouse    -tag_old_static
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
      user exrc file: "$HOME/.exrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
  fall-back for $VIM: "/usr/local/share/vim"
Compilation: gcc -c -I. -Iproto -DHAVE_CONFIG_H -O2 -fno-strength-reduce -Wall -Wno-deprecated-declarations -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1
Linking: gcc -L/usr/local/lib -Wl,--as-needed -o vim -lm -ltinfo -lrt -ldl

### Logs and stack traces

_No response_

lifecrisis:
Closes: vim/vim#12301

### DIFF
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -1,6 +1,6 @@
 " Vim filetype plugin autoload file
 " Language:	man
-" Maintainer:	Jason Franklin <vim@justemail.net>
+" Maintainer:	Jason Franklin <jason@oneway.dev>
 " Maintainer:	SungHyun Nam <goweol@gmail.com>
 " Autoload Split: Bram Moolenaar
 " Last Change: 	2022 Jun 18
@@ -154,8 +154,13 @@ func dist#man#GetPage(cmdmods, ...)
   endif
   let env_cmd = s:env_has_u ? 'env -u MANPAGER' : 'env MANPAGER=cat'
   let env_cmd .= ' GROFF_NO_SGR=1'
-  let man_cmd = env_cmd . ' man ' . s:GetCmdArg(sect, page) . ' | col -b'
+  let man_cmd = env_cmd . ' man ' . s:GetCmdArg(sect, page)
+
   silent exec "r !" . man_cmd
+
+  " Emulate piping the buffer through the "col -b" command.
+  " Ref: https://github.com/vim/vim/issues/12301
+  silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//ge
 
   if unsetwidth
     let $MANWIDTH = ''


zzzyxwvut:
Depending on Col's provenance, _tabs_ may be substituted for
runs of _spaces_, [with](https://salsa.debian.org/debian/util-linux/-/blob/8c3c94f9bc309af0665959185fada1a4e3611e45/Documentation/col.txt#L5) or [without](https://salsa.debian.org/debian/util-linux/-/blob/8c3c94f9bc309af0665959185fada1a4e3611e45/text-utils/col.1#L37) its `-h` or `-x` flags.
The emulation comment should admit that there is no such
translation before or after `:substitute`.

Compare output:
```shell
{ man col | col -bx; } > /tmp/col.1.x
{ man col | col -bh; } > /tmp/col.1.h
{ man col | col -b; } > /tmp/col.1
diff -q /tmp/col.1.h /tmp/col.1
vimdiff -o /tmp/col.1.x /tmp/col.1
```

lifecrisis:
>  The emulation comment should admit that there is no such translation before or after `:substitute`.

I think referring to the issue is sufficient. There are other details left out of the comment that are equally relevant. Brevity requires that a number of details be omitted, which is why the link to the issue was added.

Personally, I would much rather remove any attempt to reformat the output of `man` at all since `man` makes use of `col` to reformat its text when output is not a terminal. All of this seems redundant now.

If `man` does the same thing on MacOS, I think we can get rid of it.

chrisbra:
should be okay to include. One question however:
@lifecrisis @goweol are you both co-maintaining the plugin?

chrisbra:
thanks!

ychin:
Hi, this change seems to have broken this plugin on Mac. This is the before:

Before

Now it looks like this (only when you invoke `:Man ls` from Vim, not when you are using it as a MANPAGER):

After


I haven't taken a closer look yet but can we revert the change while someone works on a fix?

Additionally, can we add some unit tests to prevent regressions in the future in the supported platforms? I think right now the bundled plugins are mostly untested which have from time to time result in regressions like this and basic regression tests on the main platforms would help catch the issues before they get merged.

ychin:
Actually, I found the issue. It's because I have `set gdefault` in my vimrc, and the substitute command uses the `g` flag, which gets inverted under this. Which to be fair, even if we have regression tests, this may not have been tested as it's an edge case.

I think we can fix it like so (I will open PR):

```diff
diff --git a/runtime/autoload/dist/man.vim b/runtime/autoload/dist/man.vim
index 315636a2ef..e517714f58 100644
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -196,7 +196,11 @@ func dist#man#GetPage(cmdmods, ...)
 
   " Emulate piping the buffer through the "col -b" command.
   " Ref: https://github.com/vim/vim/issues/12301
-  silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//ge
+  if &gdefault
+    silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//e
+  else
+    silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//ge
+  endif
 
   if unsetwidth
     let $MANWIDTH = ''
```

The docs for `gdefault` does say that it's deprecated because of this reason, but I find it a little unsatisfactory:

```
				   *'gdefault'* *'gd'* *'nogdefault'* *'nogd'*
'gdefault' 'gd'		boolean	(default off)
			global
...
	DEPRECATED: Setting this option may break plugins that are not aware
	of this option.  Also, many users get confused that adding the /g flag
	has the opposite effect of that it normally does.
	This option is not used in |Vim9| script.
```

Feels to me we should just expose a way to do substitution that can ignore `gdefault` instead (like how `ignorecase` is ignored if `\c` or `\C` is used).

PR #12558, opened 2023-06-17
Fix lines being removed when modifying characters before a text prop

### LINKED ISSUE #12244: Modifying characters before a wrapping virtual text, causes the following lines to disappear
### Steps to reproduce

1. `vim -u NONE`
2. `:set all& number lines=40 columns=70`
3. `:call setline(1, 'aaaaaaaaaa')`
4. `:call prop_type_add('test', #{highlight: 'DiffAdd'})`
5. `:call prop_add(1, 11, #{type: 'test', text: repeat('a', 65)})`
6. `:call setline(2, 'bbbbbbbbbbb')`
7. `:call setline(3, 'cccccccccccc')`
8. `:call setline(4, 'ddddddddddd')`

Pressing `x` repeatedly on the `a` causes the lines below to disappear. Inserting characters before the virtual text also causes the lines below to disappear.

Screen Shot 2023-04-10 at 10 45 41 pm
The lines will reappear if they are updated (ie: inserting/deleting character, entering visual mode, etc)

Interestingly enough, if there are characters after the virtual text this bug does not happen.

There also seems to be an issue where the drawn cursor does not correlate with the actual cursor.

### Expected behaviour

For the lines below to still appear.
Screen Shot 2023-04-10 at 10 51 04 pm

### Version of Vim

9.0.1400

### Environment

Operating System: MacOS Mojave 12.6.2
Terminal: Kitty 0.26.3
$TERM: xterm-kitty
Shell: Fish 3.5.1

### Logs and stack traces

_No response_

SleepySwords:
Fixes #12244
This seems to be caused by not taking the text prop length into account if it is at the end of the line. I'm unsure if `s[charlen] == NUL` is required for something. The tests seem to run fine without it though.


chrisbra:
> s[charlen] == NUL is required

Yeah, I wondered the same actually...

Issue #12550, opened 2023-06-16
Termdebug display wrong file name in the GDB prompt buffer


skywind3000:
### Steps to reproduce

```viml
:packadd termdebug
:Termdebug xxx
```

and type "run" in the gdb prompt window

### Expected behaviour

display the correct file name, not this:

![图片](https://github.com/vim/vim/assets/3035071/8c756421-58d6-436d-b436-a74690e70bbe)

The full path of the executable is `E:\Lab\workshop\scratch\scratch\test_gdb.exe`

gdb displays it as an escaped C string: `E:\\Lab\\workshop\\scratch\\scratch\\test_gdb.exe`

But in the last segment of the path, the `\\test` has been replaced as a backslash and a tab character 
(the tab character is visible as the bar character according to my listchar option).

This only happens on Windows. 

Definitely a minor display error, but it would be nice to get fixed, since termdebug is the 
only gdb front-end that can work on MS Windows (excluding vscode).

I want to give it a go, and wish it could be polished on Windows.

### Version of Vim

9.0.1627

### Environment

OS: Windows 10
Terminal: None (GVim)

skywind3000:
the display error happens in the command line as well:

![图片](https://github.com/vim/vim/assets/3035071/56989a3b-114f-4eb3-aeac-38a9ab1ffdce)

see the `^I` in the right side of the command line.

chrisbra:
Can you try if `:set shellslash` fixes the issue?

skywind3000:
same, after `:set shellslash` :

![图片](https://github.com/vim/vim/assets/3035071/f67a2d89-b867-48ee-9bbc-04efefd59ab3)

and

![图片](https://github.com/vim/vim/assets/3035071/bde4b92d-7b12-4712-bd28-de836cb5255e)

skywind3000:
and I found that the breakpointer in the signcolumn is not displayed on Windows (I have disabled vim-signify and vim-signature for potential conflict).

The current line highlighting is invisible too.

I'm not sure is it related to the same issue (file name can't be correctly read) ?

chrisbra:
this seems to be a duplicate of #12560

PR #12565, opened 2023-06-20
termdebug: Parse filenames literally

### LINKED ISSUE #12560: [termdebug] breakpoint & current position marks are not displayed
### Steps to reproduce

1. edit `test_gdb.c`:

```c
#include <stdio.h>

int main()
{
        printf("Start:\n");
        for (int i = 0; i < 10; i++) {
                int y = 1 + i * 2 + i * 10;
                printf("y = %d\n", y);
        }
        return 0;
}
```

2. compile it with mingw:

```
gcc -g test_gdb.c -o test_gdb.exe
```

3. edit `test_gdb.vim`:

```VimL
set nocompatible
set ttimeout ttimeoutlen=50
set cmdheight=1 ruler display=lastline
set signcolumn=yes
set number showtabline=2 laststatus=2

syntax enable
syntax on

autocmd VimEnter * call EnterVim()

function! EnterVim() abort
        edit test_gdb.c
        set ft=c

        packadd termdebug
        Termdebug test_gdb.exe

        stopinsert
        wincmd w
        exec ':8'

        Break
        Run
        redraw!
endfunc
```

4. start vim:

```
vim -u start_gdb.vim
```

### Expected behaviour

the program should stop at line 8 (printf ...)
breakpoint (debugBreakpoint) and current position (debugPC) should be displayed
but they are not:

图片

Other features of termdebug are working properly.

This happens in both vim and GVim
With this bug fixed, termdebug will work smoothly on Windows.

### Version of Vim

9.0.1627

### Environment

OS: Windows 10
Terminal: cmd.exe

### Logs and stack traces

_No response_

chrisbra:
On Windows with a default path-separator of \ a filename like test.c may occur, which would be parsed by the termdebug plugin as \t and therefore replaced by a literal tab.

The same may also happen on Unix, but there filenames with a backslash in its name are much more uncommon, while on Windows this may happen just because of the default directory separator.

So stop replacing \n and \t by their typical values when parsing filenames.

Fixes: #12560 and #12550

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -588,14 +588,14 @@ func s:GdbOutCallback(channel, text)
     return
   endif
   if a:text =~ '^\^error,msg='
-    let text = s:DecodeMessage(a:text[11:])
+    let text = s:DecodeMessage(a:text[11:], v:false)
     if exists('s:evalexpr') && text =~ 'A syntax error in expression, near\|No symbol .* in current context'
       " Silently drop evaluation errors.
       unlet s:evalexpr
       return
     endif
   elseif a:text[0] == '~'
-    let text = s:DecodeMessage(a:text[1:])
+    let text = s:DecodeMessage(a:text[1:], v:false)
   else
     call s:CommOutput(a:channel, a:text)
     return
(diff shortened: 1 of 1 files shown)


skywind3000:
@brammool, any chance to get this patch merged ?

skywind3000:
thanks

PR #12573, opened 2023-06-21
Files for haskell persistent library are not recognized

lykahb:
Add filetype for the [persistent library](https://hackage.haskell.org/package/persistent). The extension `.persistentmodels` is recommended [here](https://hackage.haskell.org/package/persistent-2.14.5.0/docs/Database-Persist-TH.html#v:persistFileWith).

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -891,6 +891,7 @@ if exists('$XDG_CONFIG_HOME')
 endif
 au BufNewFile,BufRead $HOME/.config/cabal/config setf cabalconfig
 au BufNewFile,BufRead cabal.config		setf cabalconfig
+au BufNewFile,BufRead *.persistentmodels	setf haskell.persistent
 
 " Haste
 au BufNewFile,BufRead *.ht			setf haste
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -891,7 +891,7 @@ if exists('$XDG_CONFIG_HOME')
 endif
 au BufNewFile,BufRead $HOME/.config/cabal/config setf cabalconfig
 au BufNewFile,BufRead cabal.config		setf cabalconfig
-au BufNewFile,BufRead *.persistentmodels	setf haskell.persistent
+au BufNewFile,BufRead *.persistentmodels	setf haskellpersistent
 
 " Haste
 au BufNewFile,BufRead *.ht			setf haste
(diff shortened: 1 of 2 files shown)


brammool:
> Add filetype for the [persistent library](https://hackage.haskell.org/package/persistent). The extension `.persistentmodels` is recommended [here](https://hackage.haskell.org/package/persistent-2.14.5.0/docs/Database-Persist-TH.html#v:persistFileWith).

This uses the 'filetype' value "haskell.persistent".  This implies that
"persistent" by itself is a filetype.  Would it be used with anything
besides haskell?

I could not find an example of such a file.  What does it look like?

-- 
244. You use more than 20 passwords.

lykahb:
I do not think that this format would ever be used by anything else other than the haskell persistent library. It is a rather narrow use case, so I prefixed the filetype with haskell to avoid the potential name clashes.

Here are a few examples of the persistent files:
- https://github.com/naruto-unison/naruto-unison/blob/959786fe6f98fd8f9387c42c89ea04db5e34d544/config/models.persistentmodels
- https://github.com/Happy0/wordify-webapp/blob/7247dd751c2c20a55b6a8cc75ef7acdc5e9807be/config/models.persistentmodels
- https://github.com/chiroptical/cataract/blob/main/config/models.persistentmodels

brammool:
> I do not think that this format would ever be used by anything else
> other than the haskell persistent library. It is a rather narrow use
> case, so I prefixed the filetype with haskell to avoid the potential
> name clashes.

When using a dot in the filetype: {base}.{extend}  this means that the
file has the filetype {base} with extra syntax items defined by
{extend}.  So you can have "c.doxygen" and "cpp.doxygen", where the
doxygen items (annotations found in comments) are used in "c" and "cpp"
filetypes.

If "persistent" is not used on top of any other filetype, then using a
specific name, such as "haskellpersist", works better.

It also looks like the examples do not contain any haskell syntax items.
Calling it "haskell plus some extra" is misleading then.

> Here are a few examples of the persistent files:
> - https://github.com/naruto-unison/naruto-unison/blob/959786fe6f98fd8f9387c42c89ea04db5e34d544/config/models.persistentmodels
> - https://github.com/Happy0/wordify-webapp/blob/7247dd751c2c20a55b6a8cc75ef7acdc5e9807be/config/models.persistentmodels
> - https://github.com/chiroptical/cataract/blob/main/config/models.persistentmodels

Thanks.  It looks like a specification of database entries.

-- 
246. You use up your free 1 Gbyte in two days.

lykahb:
Thank you for explaining the intended meaning of dot. The file may have haskell types occurring inside of it. Also, it may be included into a haskell file as an interpreted multiline string. It's not haskell plus extra, though.

Renamed to haskellpersistent.

PR #12574, opened 2023-06-22
Not able to use the matchfuzzy() function returning a list of Dicts in a Vim9 script

yegappan:


src/testdir/test_vim9_builtin.vim
```diff
@@ -2821,13 +2821,31 @@ def Test_matchfuzzy()
   v9.CheckDefAndScriptFailure(['matchfuzzy([], 1)'], ['E1013: Argument 2: type mismatch, expected string but got number', 'E1174: String required for argument 2'])
   v9.CheckDefAndScriptFailure(['matchfuzzy([], "a", [])'], ['E1013: Argument 3: type mismatch, expected dict<any> but got list<unknown>', 'E1206: Dictionary required for argument 3'])
   matchfuzzy(['abc', 'xyz'], '')->assert_equal([])
+  var lines =<< trim END
+    var items = [{name: 'xyz', id: 1}, {name: 'def', id: 2},
+                 {name: 'abc', id: 3}]
+    var l: list<dict<any>> = matchfuzzy(items, 'abc', {key: 'name'})
+    assert_equal([{name: 'abc', id: 3}], l)
+    var k: list<string> = matchfuzzy(['one', 'two', 'who'], 'o')
```
Shane-XB-Qian, line 2829: perhaps better to add test for `w` and `wo`, not only the `o`.

src/testdir/test_vim9_builtin.vim
```diff
@@ -2821,13 +2821,31 @@ def Test_matchfuzzy()
   v9.CheckDefAndScriptFailure(['matchfuzzy([], 1)'], ['E1013: Argument 2: type mismatch, expected string but got number', 'E1174: String required for argument 2'])
   v9.CheckDefAndScriptFailure(['matchfuzzy([], "a", [])'], ['E1013: Argument 3: type mismatch, expected dict<any> but got list<unknown>', 'E1206: Dictionary required for argument 3'])
   matchfuzzy(['abc', 'xyz'], '')->assert_equal([])
+  var lines =<< trim END
+    var items = [{name: 'xyz', id: 1}, {name: 'def', id: 2},
+                 {name: 'abc', id: 3}]
+    var l: list<dict<any>> = matchfuzzy(items, 'abc', {key: 'name'})
```
Shane-XB-Qian, line 2827: perhaps better to add test for `key` with `seq` and/or `limit`.

src/testdir/test_vim9_builtin.vim
yegappan: The goal of these new tests is to verify the different types of return values in a Vim 9 script and not to test all the functionalities of these functions.  The matchfuzzy() and matchfuzzypos() functions are tested extensively in the test_matchfuzzy.vim file.

src/testdir/test_vim9_builtin.vim
yegappan: The goal of these new tests is to verify the different types of return values in a Vim 9 script and not to test all the functionalities of these functions.  The matchfuzzy() and matchfuzzypos() functions are tested extensively in the test_matchfuzzy.vim file.

PR #12581, opened 2023-06-23
No error out on empty CmdlineEnter autocommand

### LINKED ISSUE #12578: Empty CmdlineEnter autocommand causes errors in Ex mode
### Steps to reproduce

1. Run `vim --clean`
2. Run `:autocmd CmdlineEnter * :`
3. Press `gQ`
4. Run any command (e.g. `:echo 1`)
5. `E749: Empty buffer`

### Expected behaviour

No error when running command that doesn't move cursor in Ex mode

### Version of Vim

9.0.1628

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

chrisbra:
Save and Restore ex_pressedreturn when calling autocommands.

fixes #12578

### DIFF
--- a/src/autocmd.c
+++ b/src/autocmd.c
@@ -2019,6 +2019,9 @@ apply_autocmds_group(
     save_redo_T	save_redo;
     int		save_KeyTyped = KeyTyped;
     int		save_did_emsg;
+#if defined(FEAT_TIMERS) || defined(PROTO)
+    int		save_ex_pressedreturn;
+#endif
     ESTACK_CHECK_DECLARATION;
 
     /*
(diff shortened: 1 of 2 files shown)


brammool:
Some #ifdefs need to be adjusted.

chrisbra:
ah, I didn't notice. Will fix it tomorrow or monday :)

chrisbra:
should be fixed now

brammool:
The fix I had in mind was removing the #ifdefs where the functions are defined.  I'll change that.

brammool:
Included with Patch 9.0.1665

chrisbra:
> The fix I had in mind was removing the #ifdefs where the functions are defined. I'll change that.

🤦‍♂️ Oh that makes sense thank you!

Shane-XB-Qian:
image
this is the only patch last night, supposed this is the cause, so any possibility? it just happened when to emit some cmd.

chrisbra:
hard to say without much more information :/

PR #12582, opened 2023-06-23
PEM files are not recognized

ObserverOfTime:
The format is defined in [RFC7468](https://www.rfc-editor.org/rfc/rfc7468).

Things to consider:
- Should the filetype be "pem" or something else?
- Should we support more extensions (e.g. `.key`)?

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1532,6 +1532,9 @@ au BufNewFile,BufRead *.pm
 " Perl POD
 au BufNewFile,BufRead *.pod			setf pod
 
+" PEM
+au BufNewFile,BufRead *.pem,*.cer,*.crt,*.csr	setf pem
+
 " Php, php3, php4, etc.
 " Also Phtml (was used for PHP 2 in the past).
 " Also .ctp for Cake template file.
--- a/runtime/syntax/pem.vim
+++ b/runtime/syntax/pem.vim
@@ -1,0 +1,33 @@
+
+" Vim syntax file
+" Language:	PEM
+" Maintainer:	ObserverOfTime <chronobserver@disroot.org>
+" Filenames:	*.pem,*.cer,*.crt,*.csr
+" Last Change:	2023 Jun 24
+
+if exists('b:current_syntax')
+    finish
+endif
+
+let s:cpo_save = &cpoptions
+set cpoptions&vim
+
+syn region pemHeader matchgroup=pemDashes
+            \ start=/^-----\zeBEGIN/ end=/-----$/ contains=pemBegin
+syn region pemFooter matchgroup=pemDashes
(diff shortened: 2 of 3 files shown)


chrisbra:
the highlighting looks strange: 

image

I think pemLabel should be the same as pemBegin?

brammool:
> the highlighting looks strange: 
> 
> image
> 
> I think pemLabel should be the same as pemBegin?

The RFC points out the label after BEGIN and END, thus recognizing it
separately in the syntax makes sense.  Section 4 also lists known
labels, these could be highlighted specifically.

-- 
238. You think faxes are old-fashioned.

brammool:
> The format is defined in [RFC7468](https://www.rfc-editor.org/rfc/rfc7468).
> 
> Things to consider:
> - Should the filename be "pem" or something else?

Using "pem" for the 'filetype' name should be OK.  It could be more
specific, but I doubt that is useful.  The RFC mentions that PEM stands
for Privacy Enhanced Mail:

   The tradition within the RFC series can be traced back to Privacy-
   Enhanced Mail (PEM) [RFC1421], based on a proposal by Marshall Rose
   in Message Encapsulation [RFC934].  Originally called "PEM
   encapsulation mechanism", "encapsulated PEM message", or (arguably)
   "PEM printable encoding", today the format is sometimes referred to
   as "PEM encoding".  Variations include OpenPGP ASCII armor [RFC4880]
   and OpenSSH key file format [RFC4716].

"PEM encoding" should be more accurate, but is that generally used?
And I would not know a short name that is not confusing.  "pemencoding"
doesn't ring a bell for me.

> - Should we support more extensions (e.g. `.key`)?

The essential part of this filetype is encoded text in between lines:

	"-----BEGIN {label} -----"

and

	"-----END {label} -----"

If a file contains more than that we should give it a different filetype
name.

Section 5.3 of the RFC mentions ".crt" and ".cer".  Otherwise we can use
extensions that seem useful and don't conflict with other filetypes.
Generally it's better to first add the filetype and later add more
extensions when we find some that make sense.

-- 
237. You tattoo your email address on your forehead.

ObserverOfTime:
> Section 4 also lists known labels, these could be highlighted specifically.

I don't know if it's possible to highlight the known labels while keeping the catch-all match.

> The essential part of this filetype is encoded text in between lines:
>
>         "-----BEGIN {label} -----"

I suppose we could check if the first line matches `^-----BEGIN `.
Or the first few lines to account for possible comments?

brammool:
Well, it can be several nested regions:
toplevel: from BEGIN to END line
marker: the line with BEGIN or END
label: the label part of "marker"

The toplevel can appear any number of times.  Just avoid matching it inside another one.

PR #12583, opened 2023-06-24
test: extend test a bit to make sure 'matchfuzzy' worked well

Shane-XB-Qian:
note: would be failed unless #12574 fixed it.


Shane-XB-Qian:
r.i.p bram.

PR #12587, opened 2023-06-25
Make test for #12528 fail properly without the fix

zeertzjq:
The test for #12528 doesn't fail without the fix because the original
steps to reproduce run Vim with `-u NONE -i NONE` which puts Vim in
Vi-compatible mode where 'cpoptions' contains the "n" flag, whereas
RunVimInTerminal() runs Vim with `--clean` which doesn't add the "n"
flag to 'cpoptions'.

Adding the "n" flag to 'cpoptions' is enough to make the test work
properly. The `:winsize` command is also unnecessary as the zero-width
window created by `:vsplit` and `:vertical resize 0` is enough.


brammool:
> It seems that even without this PR, running the test fails locally for
> me in kitty, though it passes in XTerm. The means that the `:winsize`
> makes the screendump of the test depend on terminal capabilities, and
> therefore should be removed.

Right, the :winsize command is not portable.  Not only does it not work
in all places, a zero width won't work anyway (minimal number of columns
is 12).  And the code to check if the Vim running in a terminal window
has finished startup won't work, the ruler won't show up in the expected
location.

You can remove that TermWait() below RunVimInTerminal() now, it should
not be needed when we do wait for the ruler.


-- 
"The amigos also appear to be guilty of not citing the work of others who had
gone before them.  Even worse, they have a chapter about modeling time and
space without making a single reference to Star Trek!"
			(Scott Ambler, reviewing the UML User Guide)

PR #12590, opened 2023-06-25
Fix wrong display of control char scrolled to the left with 'nowrap'

zeertzjq:


zeertzjq:
This probably isn't the right solution. I guess the right solution is to use the value returned in the second argument of `win_lbr_chartabsize()`. I may create another PR later.

PR #12592, opened 2023-06-25
Fix text property highlighting with list mode and sign highlights

### LINKED ISSUE #12584: Text property highlight doesn't override a sign highlight over a tab character
### Steps to reproduce

Source the following script:

```
func Test()
    split Xbuffer
    call setline(1, ['one', "\ttab", '        space', 'three'])
    call prop_type_add('Prop1', #{highlight: 'Search', override: v:true})
    sign define sign1 text=>> linehl=DiffAdd
    sign place 10 line=2 name=sign1
    sign place 20 line=3 name=sign1
    call prop_add(1, 1, #{end_lnum: 4, end_col: 5, type: 'Prop1'})
endfunc
call Test()
```

### Expected behaviour

The text property highlight group should override the sign line highlight group.  In the second line, before "tab", the sign highlight group is used.  But the first column uses the text property highlight group and the columns after the tab character use the text property highlight group.

### Version of Vim

9.0.1660

### Environment

Operating System: Ubuntu 22.04
Terminal: GNOME Terminal

### Logs and stack traces

_No response_

chrisbra:
This fixes a few problems of text properties:
- text property highlighting when `override=true` does not overwrite `TAB` highlighting
- text property highlighting when `override=true` does not overwrite `TAB` highlighting with `:set list`
- text property highlighting is used instead of sign highlight after the actual text ends when `signs` are present with `linehl` is set 

includes a test 

fixes: #12584


chrisbra:
included as of dbeadf05b6a152e7d9c5cc23d9202057f8e99884

PR #12597, opened 2023-06-26
Fix wrong display with wrapping virtual text and 'smoothscroll'

zeertzjq:
Problem: Wrong display with wrapping virtual text or unprintable chars,
'showbreak' and 'smoothscroll'.
Solution: Don't skip cells taken by 'showbreak' in screen lines before
"w_skipcol". Combined "n_skip" and "skip_cells".


zeertzjq:
I've found another bug #12769 when testing this PR. More work may be needed to make sure that `mb_added` is the correct value to subtract from `skip_cells` hee.

zeertzjq:
Hmm, neither `mb_added` nor `head` are the correct values to subtract:
- `mb_added` doesn't include the size of any 'showbreak', which should be subtracted.
- `head` includes the size of 'showbreak' in all screen lines occupied by the character and the virtual text before it, not just the size of those before `w_skipcol`.

zeertzjq:
This probably needs a `cts_max_vcol` argument to `win_lbr_chartabsize()`.

zeertzjq:
This should mostly work for single width chars now. When double-width chars are involved there will be some problems with truncation, maybe those can be solved in another PR.

chrisbra:
agree thanks.

src/drawline.c
```diff
@@ -1698,20 +1695,9 @@ win_line(
 	{
 	    wlv.vcol -= charsize;
 	    ptr = prev_ptr;
-	    // If the character fits on the screen, don't need to skip it.
-	    // Except for a TAB.
-	    if (((*mb_ptr2cells)(ptr) >= charsize || *ptr == TAB)
-							       && wlv.col == 0)
-	       n_skip = v - wlv.vcol;
```
chrisbra, line 1705: Is that correct to remove? What happens with TAB at the end of a screen-line?

chrisbra:
Also, it seems to fail for me in `test_visual`
```
        From test_visual.vim:
        Found errors in Test_visual_hl_with_showbreak():
        Run 1, 19:43:25 - 19:43:26:
        command line..script /home/chrisbra/code/vim-upstream/src/testdir/runtest.vim[601]..function RunTheTest[54]..Test_visual_hl_with_showbreak[11]..VerifyScreenDump line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_visual_hl_with_showbreak.dump", "testdir/dumps/Test_visual_hl_with_showbreak.dump"); difference in line 2: "|++0#4040ff13&|a+0#0000000&@9| @38"
        Run 2, 19:43:28 - 19:43:30:
        command line..script /home/chrisbra/code/vim-upstream/src/testdir/runtest.vim[636]..function RunTheTest[54]..Test_visual_hl_with_showbreak[11]..VerifyScreenDump line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_visual_hl_with_showbreak.dump", "testdir/dumps/Test_visual_hl_with_showbreak.dump"); difference in line 2: "|++0#4040ff13&|a+0#0000000&@9| @38"
        Run 3, 19:43:32 - 19:43:33:
        command line..script /home/chrisbra/code/vim-upstream/src/testdir/runtest.vim[636]..function RunTheTest[54]..Test_visual_hl_with_showbreak[11]..VerifyScreenDump line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_visual_hl_with_showbreak.dump", "testdir/dumps/Test_visual_hl_with_showbreak.dump"); difference in line 2: "|++0#4040ff13&|a+0#0000000&@9| @38"
        Flaky test failed too often, giving up
```
grafik

src/drawline.c
zeertzjq: I don't think I "removed" anything here. I just renamed `n_skip` to `skip_cells`.

zeertzjq:
Rebased, so test should no longer fail now.

src/drawline.c
zeertzjq: Also tests added in patch 8.0.0394 should cover this.

zeertzjq: Also added another test that only works when `skip_cells` and `n_skip` are combined.

PR #12606, opened 2023-06-28
Fix blockwise Visual highlighting not working properly with virtual text

### LINKED ISSUE #12605: Blockwise Visual highlight disappears when double-width virtual text is on its left end
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
call setline(1, repeat(['1234567'], 3))
call prop_type_add('theprop', #{highlight: 'Special'})
call prop_add(2, 2, {'type': 'theprop', 'text': '口'})
exe "norm! gg2l\<C-V>$2j"
```
3. There is no Visual highlight on the second line

### Expected behaviour

There is Visual highlight on the second line

### Version of Vim

9.0.1628

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

zeertzjq:
Problem:    Wrong blockwise Visual highlight with virtual text.
Solution:   Reset the correct variable at the end of virtual selection.
            Check for double-width char inside virtual text.

Fix #12605


zeertzjq:
I've also noticed some other bugs with blockwise Visual highlight and virtual text. I'll update the PR to fix them later.

zeertzjq:
Should be fixed now.

PR #12608, opened 2023-06-28
man.vim: Uniformly place cursor at the same column with `Ctrl-t`

goweol:
Functions col and cursor count each tab (0x9) as a byte, and are complementary. On the other hand, the | command motion takes into consideration how many screen columns a tab does occupy and may move cursor to a column closer to the start of line than col would report at that position.

The provided changes prefer the cursor function to the | command.

### DIFF
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -3,7 +3,7 @@
 " Maintainer:	Jason Franklin <vim@justemail.net>
 " Maintainer:	SungHyun Nam <goweol@gmail.com>
 " Autoload Split: Bram Moolenaar
-" Last Change: 	2022 Jun 18
+" Last Change: 	2023 Jun 28
 
 let s:cpo_save = &cpo
 set cpo-=C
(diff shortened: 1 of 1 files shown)


chrisbra:
there is a slight difference between using `|` and `cursor()`. It matters e.g. when using a `tab`. Does that make a difference here?

zzzyxwvut:
There is a difference when there are `tab`s (0x9).  Try, for  
example:

```shell
# Launch Vim as follows:
MANWIDTH=80 vim --not-a-term --clean '+runtime ftplugin/man.vim' '+Man man' '+setl nofen' '+/man(7)'

# Press Ctrl-] with cursor on _m_: "\t    man(7), groff(7), man-pages(7)".
# _______________________________________[^]
#
# Then press Ctrl-t and behold the cursor is NOT at _m_.
#
# Manually confirm it by issuing: w, :echo col('.'), 5|.
#
# (Man v2.11.2)


# Launch Vim as follows:
MANWIDTH=80 vim --not-a-term --clean '+runtime ftplugin/man.vim' '+Man git' '+setl nofen' '+/git-add(1)'

# Press Ctrl-] with cursor on _g_: "       git-add(1)".
# ________________________________________[^]
#
# Then press Ctrl-t and behold the cursor is at _g_.
#
# (Git v2.39.2)
```

chrisbra:
okay, if you are fine with it, so am I :)

chrisbra (approved):

PR #12610, opened 2023-06-28
Python 3.11 depreciations seen as well on gcc compiler

pheiduck:
`__GNUC__` avoid this warnings on both compilers.

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -82,8 +82,9 @@
 #endif
 
 // Suppress Python 3.11 depreciations to see useful warnings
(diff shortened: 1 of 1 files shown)


chrisbra:
so this fixes deprecation warnings for GCC. But why were those pragma statements removed for clang?

k-takata:
`__GNUC__` is also defined in Clang.

pheiduck:
@chrisbra this what @k-takata say is correct.

yegappan:

pheiduck:
> Can you please update and rebase this PR to the latest?

@yegappan done already.

PR #12612, opened 2023-06-28
fix(indent): do not conflate dictionary key with end of block

lacygoill:
This script:
```vim
vim9script
if true
var d = {
end: 0}
endif
```
is indented like this:
```vim
vim9script
if true
    var d = {
        end: 0}
    endif
```
It should be indented like this:
```vim
vim9script
if true
    var d = {
        end: 0}
endif
```


chrisbra:
I am fine with that. Do we need a minor-patch for that?

chrisbra:
no no, it's more a question if indent test scrips in general (should) require a minor patch number or not.

PR #12617, opened 2023-06-30
Fix passing multiple patterns to :runtime not working for 'packpath'

zeertzjq:


chrisbra:
is this actually fixing a bug?

zeertzjq:
Yes. Previously when passing multiple patterns to e.g. `:runtime OPT`, only the first pattern works properly, because the 'packpath' prefix was prepended to the whole argument string instead of each argument.

PR #12618, opened 2023-06-30
Fix alignment in filetype.txt

zeertzjq:
There are three spaces because the "<" is concealed.


chrisbra:
thanks, obviously correct.

PR #12619, opened 2023-06-30
debsources files using deb822 format are not recognized

lacygoill:
If we open `/etc/apt/sources.list.d/debian.sources`, the filetype is not set.  It should be set to `debsources`.  The format `deb822` allows APT sources to be specified in a more structured way.  But this format can only be used in a `.sources` file;  not in a `.list` file.  See:

   > Files in this format have the extension .sources.

Source: https://manpages.debian.org/stretch/apt/sources.list.5.en.html#DEB822-STYLE_FORMAT

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -538,8 +538,8 @@ au BufNewFile,BufRead copyright
 	\| endif
 
 " Debian Sources.list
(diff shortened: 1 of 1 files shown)


jamessan:
@jayaddison has provided an initial implementation upstream, which I'll try to review soon and send to Bram.

PR #12625, opened 2023-07-03
Fix screenpos() not working properly with w_skipcol and cpoptions+=n

### LINKED ISSUE #12547: screenpos() wrong with smoothscroll
### Steps to reproduce

Assuming width 80 terminal,
```
vim --clean -S <(tee << 'EOF'
    set smoothscroll
    call setline(1, 'Lorem ipsum dolor sit amet, qui minim labore adipisicing minim sint cillum sint consectetur cupidatat.')
    call setline(2, '')
    exe "normal! G\<C-e>"
EOF
)
```

Result
```
{'col': 1, 'row': 3, 'endcol': 1, 'curscol': 1}
```
image

### Expected behaviour

'row' should be 2

### Version of Vim

9.0.1636

### Environment

ubuntu 22.04

### Logs and stack traces

_No response_

zeertzjq:
Problem:    screenpos() wrong result with w_skipcol and cpoptions+=n
Solution:   Use adjust_plines_for_skipcol() instead of subtracting
            w_skipcol.

Also fix #12547

### DIFF
--- a/src/misc1.c
+++ b/src/misc1.c
@@ -342,12 +342,13 @@ plines(linenr_T lnum)
 plines_win(
     win_T	*wp,
     linenr_T	lnum,
-    int		winheight)	// when TRUE limit to window height
+    int		limit_winheight)	// when TRUE limit to window height
 {
 #if defined(FEAT_DIFF) || defined(PROTO)
     // Check for filler lines above this buffer line.  When folded the result
     // is one line anyway.
-    return plines_win_nofill(wp, lnum, winheight) + diff_check_fill(wp, lnum);
+    return plines_win_nofill(wp, lnum, limit_winheight)
+						   + diff_check_fill(wp, lnum);
 }
 
 /*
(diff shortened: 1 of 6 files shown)


src/misc1.c
```diff
@@ -497,7 +497,7 @@ plines_win_col(win_T *wp, linenr_T lnum, long column)
 }
 
     int
-plines_m_win(win_T *wp, linenr_T first, linenr_T last)
+plines_m_win(win_T *wp, linenr_T first, linenr_T last, int winheight)
```
yegappan: The 'winheight' argument is not specifying the window height but instead is a Boolean indicating whether to limit
to the value of the 'winheight' option.  For readability reasons, can you rename this argument to 'use_winheight'
or 'limit_winheight'? I noticed that this name is also used in the existing plines_win() and plines_win_nofill()
functions.

zeertzjq: Renamed to `limit_winheight`.

yegappan (approved):

PR #12627, opened 2023-07-03
Add a NULL check of cl in object_clear()

XidianGeneral:
In the commit d13dd30240e3, `rettv->vval.v_object->obj_class` or `cl` can be NULL, and thus a NULL check is required. However, in the following call stack, `obj->obj_class` or `cl` is not checked:
`handle_subscript()` -> `eval_lambda()` -> `clear_tv()` -> `object_unref()` -> `object_clear()`
Thus, a NULL check of `cl` should be added in `object_clear()`.

### DIFF
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -1502,6 +1502,9 @@ object_clear(object_T *obj)
 
     class_T *cl = obj->obj_class;
 
+    if (!cl)
+        return;
+
     // the member values are just after the object structure
     typval_T *tv = (typval_T *)(obj + 1);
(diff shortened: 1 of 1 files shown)


chrisbra:
Since you have a call-stack, what commands do reproduce this problem please?

XidianGeneral:
This possible bug is reported by a static analysis implemented by myself. This tool checks the call trace of the known bug, to extend and find extra bug similar to the known bugs. Thus, the reported (extra) bug is not found via actual execution.
However, the tool provides the detailed code path from the buggy variable assignment to its use. The call stack in the description is simpified from the provided code path.

PR #12628, opened 2023-07-03
Remove dead code

zeertzjq:
linetabsize_col() calls init_chartabsize_arg() with 0 as "lnum", so
cts.cts_has_prop_with_text is always FALSE.


yegappan (approved):

chrisbra:
fixed by d3515a1e88cf25c1d5eae8faa965b587a124e687

PR #12633, opened 2023-07-05
Add a {winid} argument to virtcol()

zeertzjq:
Other functions col(), charcol() and virtcol2col() support a {winid}
argument, so it makes sense for virtcol() to also support that.

Also add test for virtcol2col() with 'showbreak' and {winid}.

### DIFF
--- a/src/testdir/test_cursor_func.vim
+++ b/src/testdir/test_cursor_func.vim
@@ -540,9 +540,28 @@ func Test_virtcol2col()
   call assert_equal(8, virtcol2col(0, 1, 7))
   call assert_equal(8, virtcol2col(0, 1, 8))
 
+  let w = winwidth(0)
+  call setline(2, repeat('a', w + 2))
+  let win_nosbr = win_getid()
+  split
+  setlocal showbreak=!!
+  let win_sbr = win_getid()
+  call assert_equal(w, virtcol2col(win_nosbr, 2, w))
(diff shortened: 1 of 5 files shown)


chrisbra (approved):

src/testdir/test_cursor_func.vim
```diff
@@ -540,9 +540,28 @@ func Test_virtcol2col()
...
+  split
+  setlocal showbreak=!!
+  let win_sbr = win_getid()
+  call assert_equal(w, virtcol2col(win_nosbr, 2, w))
+  call assert_equal(w + 1, virtcol2col(win_nosbr, 2, w + 1))
+  call assert_equal(w + 2, virtcol2col(win_nosbr, 2, w + 2))
+  call assert_equal(w + 2, virtcol2col(win_nosbr, 2, w + 3))
+  call assert_equal(w, virtcol2col(win_sbr, 2, w))
+  call assert_equal(w + 1, virtcol2col(win_sbr, 2, w + 1))
+  call assert_equal(w + 1, virtcol2col(win_sbr, 2, w + 2))
```
zeertzjq, line 555: This may be subject to change. Should the result be `w + 1` or `w` here on 'showbreak'?

zeertzjq: Oh, `w + 1` seems to match the behavior of mouse clicking.

PR #12643, opened 2023-07-06
Make scrolloff and sidescrolloff  global-local

### LINKED ISSUE #12596: Setting local value of global-local window option sets local value for other buffers in window
### Steps to reproduce

```viml
setlocal scrolloff=2
setglobal scrolloff=3
setl modified " Force fresh buffer with enew
hide enew
echo &l:scrolloff &g:scrolloff
```
The script above prints "2 3". I don't understand why the local value of the new buffer is set to 2. If this is expected, is there documentation that explains this behavior?

### Expected behaviour

The script should print -1 3

### Version of Vim

9.0.1664

### Environment

Linux

### Logs and stack traces

_No response_

chrisbra:
It was noticed, that sidescrolloff and scrolloff options behave differently in comparison to other global-local window options like 'listchars'

So make those two behave like other global-local options.

Add a few tests to make sure all global-local window options behave similar

fixes #12596

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -252,6 +252,8 @@ opt+=val" the expansion is done before the adding or removing.
 
 Handling of local options			*local-options*
 
+Note: The following also applies to |global-local| options.
+
 Some of the options only apply to a window or buffer.  Each window or buffer
 has its own copy of this option, thus each can have its own value.  This
 allows you to set 'list' in one window but not in another.  And set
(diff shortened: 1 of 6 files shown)


runtime/doc/options.txt
```diff
@@ -374,6 +374,12 @@ used.  Thus it does the same as: >
 Note: In the future more global options can be made |global-local|.  Using
 ":setlocal" on a global option might work differently then.
 
+							*window-global-local*
+One special feature are |global-local| window options. Those options are not
+only local to the window, but in addition to the buffer inside the window.
+Each buffer remembers uses its own local version of that option value, so when
+you edit the same buffer in a different window, it carries along and re-uses
+the window-local options.
```
zeertzjq: I think this also applies to window-local options, and is already documented at `:h local-options`.

chrisbra: ah indeed, I have removed it again.

PR #12647, opened 2023-07-08
Improve the teapot

seandewar:
Update the builtin-function-list entry. (Implicitly returns 0, but such functions usually use "none")

~~Correct the RFC number.~~ Actually it's OK.

tv_get_string_strict() can not return NULL.


runtime/doc/builtin.txt
```diff
@@ -156,7 +156,7 @@ digraph_setlist({digraphlist})	Boolean	register multiple |digraph|s
 echoraw({expr})			none	output {expr} as-is
 empty({expr})			Number	|TRUE| if {expr} is empty
 environ()			Dict	return environment variables
-err_teapot()			Number	produce error 418
+err_teapot([{expr}])		none	give E418 or E503 if {expr} is |TRUE|
```
```suggestion
err_teapot([{expr}])		none	give E418, or E503 if {expr} is |TRUE|
```

Issue #12586, opened 2023-06-24
Test_uncrypt_xchacha20v2 and _custom fail on big-endian systems


jamessan:
### Steps to reproduce

1. Run test suite on big endian system

### Expected behaviour

Tests are either skipped if the environment isn't supported or pass

### Version of Vim

9.0.1658

### Environment

Debian sid big-endian systems

### Logs and stack traces

```shell
Found errors in Test_uncrypt_xchacha20v2():
command line..script /home/jamessan/vim_sid_s390x/vim-9.0.1658/src/testdir/runtest.vim[573]..function RunTheTest[52]..Test_uncrypt_xchacha20v2[16]..Uncrypt_stable_xxd line 9: Expected ['abcdefghijklmnopqrstuvwxyzäöü', 'ZZZ_äüöÄÜÖ_!@#$%^&*()_+=-`~"'] but got ['©|/\n\x0b\x19ë\x1d\x1e¥?"ÁyK>', 'ë\x19ccø"-µÑ<s{åÕGêDÌl%x?ÕØ6Ç\n\x01"»0zY±*èçÛ\x11:÷2z4\x133Ü\x14Q g¹\x0f:µl\x1fS°$\x16 ZL_ÞM¬&\x11H$ðº\nÁ`']
Found errors in Test_uncrypt_xchacha20v2_custom():
command line..script /home/jamessan/vim_sid_s390x/vim-9.0.1658/src/testdir/runtest.vim[573]..function RunTheTest[52]..Test_uncrypt_xchacha20v2_custom[13]..Uncrypt_stable_xxd line 9: Expected ['', 'foo', 'bar', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10'] but got ['¹s_3éTü\x13º>\x04k15·w^¬\x7fã\fÒ\x14ßíuKec\x05\x03\\ì¤Ï3Òu\x07ì8ºb£''hØ­&\x07?¦ù]~¨ùH \f']
```

jamessan:
I tried changing `Uncrypt_stable_xxd` to use `xxd -e -r`, but that still fails:

```
Found errors in Test_uncrypt_xchacha20v2():
[...]

PR #12653, opened 2023-07-11
run CI on additional architectures using qemu

chrisbra:
in #12586 it was noted, that the s390x CI architecture was gone when migrating away from Travis CI,
So let's use qemu to build and run the Vim test suite on the following architectures: `s390x`, `riscv64` and `aarch64`

While creating this, I noticed a couple of other failing tests besides the sodium encryption test that was mentioned in #12586 (and which is still on my todo-list). However I did not figure out, why those are failing, so I have disabled them in a few tests.

@jamessan did you notice some of the tests failures in Debian as well? 

Note: this CI creates a docker image, which will be available in this repository and the link would look like this (once the CI has successfully run), search for `ghcr.io` In the resulting logfiles to find out:
- https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-riscv64-ubuntu22-04
- https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-aarch64-ubuntu22-04
- https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-s390x-ubuntu22-04

I suppose it could be possible to `docker pull` those images and inspect the resulting builds using qemu. But I haven't explored it in much details, it was way too slow on my system to try to run it. Relevant Docker Documentation: https://github.com/multiarch/qemu-user-static

### DIFF
--- a/src/testdir/test_excmd.vim
+++ b/src/testdir/test_excmd.vim
@@ -364,6 +364,9 @@ endfunc
 func Test_confirm_write_ro()
   CheckNotGui
   CheckRunVimInTerminal
+  for arch in ['s390x', 'aarch64', 'riscv64']
+    exe "CheckArch "  .. arch
+  endfor
 
   call writefile(['foo'], 'Xconfirm_write_ro', 'D')
   let lines =<< trim END
--- a/src/testdir/test_excmd.vim
+++ b/src/testdir/test_excmd.vim
@@ -364,9 +364,6 @@ endfunc
 func Test_confirm_write_ro()
   CheckNotGui
   CheckRunVimInTerminal
-  for arch in ['s390x', 'aarch64', 'riscv64']
-    exe "CheckArch "  .. arch
-  endfor
 
   call writefile(['foo'], 'Xconfirm_write_ro', 'D')
   let lines =<< trim END
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -1,0 +1,36 @@
+name: Qemu
+
+on: [push, pull_request]
+
+jobs:
+  qemu_job:
+    timeout-minutes: 60
+    runs-on: ubuntu-22.04
+    name: Vim on Qemu
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ "s390x", "aarch64", "riscv64" ]
+    steps:
+      - uses: actions/checkout@v3
+        name: Checkout Source
+      - uses: uraimo/run-on-arch-action@v2
+        name: Build on Qemu
+        id: runcmd
+        with:
+          arch: ${{ matrix.arch }}
+          distro: ubuntu22.04
+          githubToken: ${{ github.token }}
+          shell: /bin/bash
+
+          run: |
+            lscpu
+            apt-get update  -q -y
+            apt-get install -q -y --no-install-recommends build-essential libncurses-dev libsodium-dev
+            pushd src
+            ./configure --with-features=huge --enable-gui=no
+            make -j$(getconf _NPROCESSORS_ONLN)
+            ./vim --version
+            cd testdir
+            make
+            popd
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -32,5 +32,6 @@ jobs:
             make -j$(getconf _NPROCESSORS_ONLN)
             ./vim --version
             cd testdir
+            export CI_QEMU=1
             make
             popd
--- a/src/testdir/check.vim
+++ b/src/testdir/check.vim
@@ -127,6 +127,19 @@ func CheckNotMacM1()
   endif
 endfunc
 
+" Command to check for Architecture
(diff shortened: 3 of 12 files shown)


brammool:
Christian wrote:

> in #12586 it was noted, that the s390x CI architecture was gone when
> migrating away from Travis CI, So let&#39;s use qemu to build and run
> the Vim test suite on the following architectures: `s390x`, `riscv64`
> and `aarch64`
>
> While creating this, I noticed a couple of other failing tests besides
> the sodium encryption test that was mentioned in #12586 (and which is
> still on my todo-list). However I did not figure out, why those are
> failing, so I have disabled them in a few tests.
> 
> @jamessan did you notice some of the tests failures in Debian as well? 

The failures I noticed were timeouts.

> Note: this CI creates a docker image, which will be available in this
> repository and the link would look like this (once the CI has
> successfully run), search for `ghcr.io` In the resulting logfiles to
> find out:
> - https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-riscv64-ubuntu22-04
> - https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-aarch64-ubuntu22-04
> - https://github.com/vim/vim/pkgs/container/vim%2Frun-on-arch-vim-vim-qemu-s390x-ubuntu22-04
> 
> I suppose it could be possible to `docker pull` those images and
> inspect the resulting builds using qemu. But I haven't explored it
> in much details, it was way too slow on my system to try to run it.
> Relevant Docker Documentation:
> https://github.com/multiarch/qemu-user-static

The test log should also contain the different text.  You can edit the
screendump, change the text as indicated and see what the difference is
using term_dumpdiff().

-- 

brammool:
Christian wrote:

> hm, now there are few more failing tests, that did not occur in my
> local fork. Wondering if I should increase the global timeout for some
> of those.... 

I noticed two failures.  After waiting a few minutes it was OK again.
Not sure what happened.

-- 
       Bravely bold Sir Robin, rode forth from Camelot,
       He was not afraid to die, Oh Brave Sir Robin,
       He was not at all afraid to be killed in nasty ways
       Brave, brave, brave, brave Sir Robin.

jamessan:
> So let's use qemu to build and run the Vim test suite on the following architectures: s390x, riscv64 and aarch64
> ...
> @jamessan did you notice some of the tests failures in Debian as well?

They've all passed in my last upload from v9.0.1672 (which skips the crypt tests on big-endian systems).

* arm64 - https://buildd.debian.org/status/fetch.php?pkg=vim&arch=arm64&ver=2%3A9.0.1672-1&stamp=1688550690&raw=0
* riscv64 - https://buildd.debian.org/status/fetch.php?pkg=vim&arch=riscv64&ver=2%3A9.0.1672-1&stamp=1688520221&raw=0
* s390x - https://buildd.debian.org/status/fetch.php?pkg=vim&arch=s390x&ver=2%3A9.0.1672-1&stamp=1688563502&raw=0

Sometimes there's flakiness in the tests, but a couple reruns works.  Alpha is a bit worse off.  It hasn't successfully passed tests in a while - https://buildd.debian.org/status/logs.php?pkg=vim&arch=alpha

src/testdir/test_excmd.vim
```diff
@@ -364,6 +364,9 @@ endfunc
 func Test_confirm_write_ro()
   CheckNotGui
   CheckRunVimInTerminal
+  for arch in ['s390x', 'aarch64', 'riscv64']
+    exe "CheckArch "  .. arch
```
jamessan: The arch isn't actually relevant, right?  You're using this as a proxy for slowness, which doesn't seem correct. As I showed in the Debian build looks, these do pass on those architectures.

jamessan (commented):
I don't think CheckArch is actually needed.

src/testdir/test_excmd.vim
chrisbra: Yeah, I'll revisit, once the sodium fix is included.

chrisbra:
updated to current master, enabled all tests again.

chrisbra:
so even more tests are flaky when run from qemu. I don't think this is worth it...

jamessan:
> so even more tests are flaky when run from qemu

Yeah, qemu is slower.  It's going to expose non-determinism more.  The non-determinism is _already_ a problem, just one that is hidden on more common/faster architectures.

Debian routinely has ~4-5 architectures that fail to build due to test issues.  They happen on the unofficial architectures, so I don't look into them as often since I'm already stretched thin on Debian stuff.

chrisbra:
>  It's going to expose non-determinism more. The non-determinism is already a problem, just one that is hidden on more common/faster architectures.

Yes exactly. That means the current tests are already a problem, currently the CI rarely ever shows this project to be "green" which is already quite a big problem. Therefore I don't think adding more CI jobs that will cause more such issues are not particular useful 🤷 

The real problem will be to fix the non-determistic tests and that is already quite a hard job, especially when not being easily reproducible.

PR #9681, opened 2022-02-01
force restricted-mode for nologin shell

matveyt:
Problem: Vim only enters restricted-mode if -Z switch supplied on command line
Solution: For Unix force restricted-mode also when $SHELL is set to "nologin" or "false" (closes #8126).

This partially closes a well-known hole when "nologin" user runs Vim in order to get access to normal shell.

Also, it helps for better error handling by plugins when run by "nologin" user. I.e. always `catch /E145:/` for all shell commands instead of different and unexpected `/E484:/` or such. Notably this applies to perl standard plugin.

PR #12657, opened 2023-07-11
forced restricted mode based on $SHELL astonishes

### LINKED ISSUE #12656: forced restricted mode based on $SHELL astonishes
Please consider reverting the change applied to `vim` in patch 8.2.4282, in which the value of the `SHELL` environment variable is inspected and used to cause `vim` to enter "restricted" mode (a la `rvim` or the `-Z` command line option).

In the presence of a `SHELL` environment variable that has been placed in the environment (e.g., via `sudo`, rather than explicitly set by the user in a shell startup script, or similar) and happens to have an unfortunate value ending in `false` or `nologin`, the behavior of `vim` acting as `rvim` is unexpected, confusing, and breaks basic job control usage in which `vim` can be suspended via `CTRL-z`.

**EDITED:** I have a tested patch in PR #12657 which basically reverses the changes from patch 8.2.4282. ~~I'll create a PR for it shortly.~~

### Steps to reproduce the issue:

(since [patch 8.2.4282](https://github.com/vim/vim/commit/adbb1bf21dad5697cd82d46d9dd9e8e8d0f647e6)):
```sh
    $ SHELL=/usr/bin/nologin /path/to/vim
```
OR:
```sh
    $ SHELL=/la/la/la/false /path/to/vim
```
OR similar...

Once inside `vim`, press `CTRL-z` to attempt to suspend the `vim` process (standard *nix job control). Rather than `vim` being suspended, an error is shown:
```
    E145: Shell commands and some functionality not allowed in rvim
```

Having `$SHELL` set (intentionally or otherwise) to any value that ends with either `nologin` or `false` prevents `vim` from being suspended.

**WORKAROUND:** Prior to invoking `vim`, either `unset SHELL` or ensure `$SHELL` is set to some value that will not trigger this behavior:
```sh
    $ unset SHELL
    $ /path/to/vim
```
OR
```sh
    $ SHELL=/bin/sh /path/to/vim
```

See also:
   * [patch 8.2.4282](https://github.com/vim/vim/commit/adbb1bf21dad5697cd82d46d9dd9e8e8d0f647e6)
   * #9681 

### Expected behaviour

I was expecting `vim` to be suspended when I pressed `CTRL-z`, per normal Unix job control behavior, and per long-standing `vim` behavior.

### Version of Vim

```tty
    $ /usr/bin/vim --version | head -n 4
    VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 15 2023 04:26:05)
    macOS version - x86_64
    Included patches: 1-1424
    Compiled by root@apple.com
```

### Environment

Operating system: macOS 13.4.1 ("Ventura")
```tty
    $ /usr/bin/sw_vers
    ProductName:            macOS
    ProductVersion:         13.4.1
    BuildVersion:           22F82
```

Terminal: XTerm (running on GNU/Linux; access to the macOS machine is over ssh)
Value of $TERM: `xterm-256color`
Shell: Bash 5.2.15(1)-release (from MacPorts)

### Logs and stack traces

Here's what the session looked like that led me to this issue:
```shell
    $ /usr/bin/sudo /opt/local/bin/bash
    Password:

    # whoami
    root

    # /usr/bin/vim /etc/whatever

    <Edit a bunch of changes (not yet written)>
    <Attempt to suspend vim via CTRL-z to check on something>
    <trip over "vim behaving as rvim" behavior>
```

The above is a common flow for any administrator. I happened to trip over the restricted behavior because our company MDM push forced an OS update on me (macOS 11.x ("Big Sur") to macOS 13.x ("Ventura")) and changed the `root` user's login shell to `/usr/bin/false` in the macOS Directory Services. That caused `sudo` to use that value to set the `SHELL` environment variable in the newly created Bash shell for the `root` user.

While the issue can be easily worked around once you know the root cause, the root cause is hardly apparent to the user. And the user should not need to do that. The behavior is [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) because it breaks an important decades-old, legit usage pattern. Also, the purpose of the `SHELL` environment variable is to allow a user to indicate a preference for a particular shell to applications (such as `vim`) that allow the user to "shell out" and similar; imposing 'restricted' semantics of the user's `vim` session based on the value of `$SHELL` is overloading the semantics and is confusing to the user.

Here's [what POSIX says about environment variables](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) in general:

> If the variables in the following two sections are present in the
> environment during the execution of an application or utility, they
> shall be given the meaning described below. Some are placed into the
> environment by the implementation at the time the user logs in; all
> can be added or changed by the user or any ancestor of the current
> process. The implementation adds or changes environment variables
> named in POSIX.1-2017 only as specified in POSIX.1-2017. If they are
> defined in the application's environment, the utilities in the Shell
> and Utilities volume of POSIX.1-2017 and the functions in the System
> Interfaces volume of POSIX.1-2017 assume they have the specified
> meaning. Conforming applications shall not set these environment
> variables to have meanings other than as described.

and the `SHELL` environment variable in particular:

> SHELL
>     This variable shall represent a pathname of the user's preferred
>     command language interpreter. If this interpreter does not conform
>     to the Shell Command Language in XCU Shell Command Language,
>     utilities may behave differently from those described in
>     POSIX.1-2017.

salewski:
Problem:    Overloading semantics of $SHELL interferes with legit job control
Solution:   Avoid triggering restricted mode based on $SHELL value (closes #12656)

This PR basically reverses the changes from #9681, which were incorporated into [patch 8.2.4282](https://github.com/vim/vim/commit/adbb1bf21dad5697cd82d46d9dd9e8e8d0f647e6).

More detailed rationale for this PR is in issue #12656.

### DIFF
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -256,8 +256,6 @@ a slash.  Thus "-R" means recovery and "-/R" readonly.
 		Interfaces, such as Python, Ruby and Lua, are also disabled,
 		since they could be used to execute shell commands.  Perl uses
 		the Safe module.
-		For Unix restricted mode is used when the last part of $SHELL
-		is "nologin" or "false".
 		Note that the user may still find a loophole to execute a
 		shell command, it has only been made difficult.
 
--- a/src/option.c
+++ b/src/option.c
@@ -307,24 +307,6 @@ set_init_default_printexpr(void)
 }
 #endif
 
-#ifdef UNIX
-/*
- * Force restricted-mode on for "nologin" or "false" $SHELL
- */
-    static void
-set_init_restricted_mode(void)
-{
-    char_u	*p;
-
-    p = get_isolated_shell_name();
-    if (p == NULL)
-	return;
-    if (fnamecmp(p, "nologin") == 0 || fnamecmp(p, "false") == 0)
-	restricted = TRUE;
-    vim_free(p);
-}
-#endif
-
 #ifdef CLEAN_RUNTIMEPATH
 /*
  * When Vim is started with the "--clean" argument, set the default value
@@ -570,10 +552,6 @@ set_init_1(int clean_arg)
      * value.  Also set the global value for local options.
      */
     set_options_default(0);
-
-#ifdef UNIX
-    set_init_restricted_mode();
-#endif
 
 #ifdef CLEAN_RUNTIMEPATH
     if (clean_arg)
--- a/src/testdir/test_restricted.vim
+++ b/src/testdir/test_restricted.vim
@@ -105,14 +105,6 @@ func Test_restricted_mode()
   if RunVim([], [], '-Z --clean -S Xrestricted')
     call assert_equal([], readfile('Xresult'))
   endif
-  call delete('Xresult')
-  if has('unix') && RunVimPiped([], [], '--clean -S Xrestricted', 'SHELL=/bin/false ')
-    call assert_equal([], readfile('Xresult'))
-  endif
-  call delete('Xresult')
-  if has('unix') && RunVimPiped([], [], '--clean -S Xrestricted', 'SHELL=/sbin/nologin')
-    call assert_equal([], readfile('Xresult'))
-  endif
 
   call delete('Xresult')
 endfunc


salewski:
I see the failed test for:
```tty
    GitHub CI / linux (huge, gcc, testgui, true, true) (pull request) Failing after 23m
```

(details at: https://github.com/vim/vim/actions/runs/5525706560/jobs/10107033370?pr=12657 )

That failure seems unrelated to the changes from this PR, and indeed looking at recent test runs for other PRs that predate this one shows what is apparently the same failure:
```
...
06:44 Executing Test_lvimgrep_crash()

Error detected while processing command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[585]..function
RunTheTest[54]..Test_lvimgrep_crash[16]..TestTimeout[12]..VimLeavePre Autocommands for "*"..function EarlyExit[7]..FinishTesting:
line   70:
E445: Other window contains changes
E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txt
E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txt
E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txt
Error detected while processing command line..script <f0><9b>ֺ<8f>U[585]..function
RunTheTe<c0>^A[54]..Test_lvimgrep_crash[16]..TestTime<c0>^A[12]..VimLeavePre Autocommands for "*"..function
EarlyExiP^B[7]..FinishTesting:malloc(): unsorted double linked list corrupted

Error: The action has timed out.
```

chrisbra:
closing for the reasons mentioned in #12656

PR #12659, opened 2023-07-12
Add filetype detection for eyaml files

VannTen:
https://github.com/voxpupuli/hiera-eyaml/ uses and produces the eyaml
format, which is simply yaml with some encrypted values.

It's convenient to edit the file without decrypting when not touching
encrypted values (or when you don't have access to the decryption key),
which is why vim should treat those files as yaml files.


chrisbra:
Thanks!

PR #12660, opened 2023-07-12
if_python3.c: Fix building dynamic Python3 interpreter with Python 3.12

zdohnal:
There are new extern global variables defined in Python 3.12 development files, which types are within python3 library, so they break dynamic python3 interpret Vim plugin.

Since the variables are used in macro `Py_SIZE` which is used in other python3 headers, the dummy variables have to defined before including Python.h.

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -68,6 +68,8 @@
 #endif
 
 #define PY_SSIZE_T_CLEAN
+#define PyLong_Type (*py3_PyLong_Type)
+#define PyBool_Type (*py3_PyBool_Type)
 
 #include <Python.h>
 
@@ -270,7 +272,6 @@ static HINSTANCE hinstPy3 = 0; // Instance of python.dll
 # define PyFloat_Type (*py3_PyFloat_Type)
 # define PyNumber_Check (*py3_PyNumber_Check)
 # define PyNumber_Long (*py3_PyNumber_Long)
-# define PyBool_Type (*py3_PyBool_Type)
 # define PyErr_NewException py3_PyErr_NewException
 # ifdef Py_DEBUG
 #  define _Py_NegativeRefcount py3__Py_NegativeRefcount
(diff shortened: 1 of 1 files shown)


chrisbra (approved):

zdohnal:
Hi @chrisbra ,

would you mind telling me what needs to changed? I can update the PR.

zdohnal:
@chrisbra I have answered via email and sent patch - I don't have a Win dev machine with python 3.12, so I'm waiting on Nicolas' verification. If it works for him, I'll create a PR.

The patch [vim-py312win.patch.txt](https://github.com/vim/vim/files/12334668/vim-py312win.patch.txt)

zdohnal:
The new PR is here https://github.com/vim/vim/pull/12800 , Nicolas verified the functionality on Win and I on Fedora 38 and 40.

chrisbra:
thanks both

ychin:
Hi, this change (along with #12800) may have fixed dynamic linkage, but I think it broke static linking. It's easy to see why: this relies on `py3_PyLong_Type` and `py3_PyBool_Type`, which are only defined when `DYNAMIC_PYTHON3` is used. Please make sure to test both and avoid regressions when trying to fix one side (dynamic).

Furthermore, my opinion is that this is a bug on Python's side. Their headers should be resilient and allows for dynamic linkage. This seems like a new issue in 3.12 (which is not officially released yet) and the first time it caused an issue like this on our integration. My understanding is that Python allows dynamic linkage even if stable ABI is not used (as long as both sides use the same minor version), so Vim's usage isn't out of the norm.

I'm also wondering if the addition here is correct. Python's new code is doing this ([source](https://github.com/python/cpython/blob/9c31d9405027cea9c2d039ade672d604663ed5b0/Include/object.h#L231)):

```c
    assert(ob->ob_type != &PyLong_Type);
    assert(ob->ob_type != &PyBool_Type);
```

We are relying on `ob_type` matching the `PyLong_Type` that we loaded from the DLL which seems like it should work but I'm not 100% sure.

My vote is to revert the change and file an issue in CPython clarifying the correct behavior (I can do that). I'm suggesting this because what we are doing here seems quite fragile and reactive to individual version's minor implementation details; and reminds me of the mess what `if_perl.xs` is doing by essentially copy-pasting inline functions due to how the Perl interface works.

ychin:
Another concern I have about this type of fix is that you have to do it before including `Python.h`. This means you won't be able to use `PY_VERSION_HEX` to selectively control the behavior of what we want to do as that is only defined in `Python.h` and we have a chicken-and-egg issue.

PR #12663, opened 2023-07-13
Fix wrong cursor pos with mouse click in two cases

zeertzjq:
1. Clicking on a TAB or an unprintable character
2. Clicking after concealed text when 'virtualedit' is set

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -1830,9 +1830,9 @@ win_line(
     win_line_start(wp, &wlv, FALSE);
 
     // Repeat for the whole displayed line.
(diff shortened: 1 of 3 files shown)


zeertzjq:
This seems to be wrong when a TAB wraps and 'number' or 'showbreak' is set, so there needs to be a better way to calculate coladd.

zeertzjq:
Note to self: since a TAB can span infinitely many times, the best way to fix this is probably to store virtual columns instead of text columns in `ScreenCols[]`. Then `coladvance()` can be used just like the case without `ScreenCols[]`.

PR #12666, opened 2023-07-13
Cleanup Tests

chrisbra:
Clean up temporary left-over files after tests run.

There were a few failures in 'linux (huge, gcc, testgui, true, true)' e.g. here: https://github.com/vim/vim/actions/runs/5497376153/jobs/10018060156

,----
| Error detected while processing command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[585]..function RunTheTest[54]..Test_lvimgrep_crash[16]..TestTimeout[12]..VimLeavePre Autocommands for "*"..function EarlyExit[7]..FinishTesting:
| line   70:
| E445: Other window contains changes
| E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txt
| E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txt
| E937: Attempt to delete a buffer that is in use: Xtest_stable_xxd.txtmalloc(): unsorted double linked list corrupted
`----

Which is puzzling, because the Xtest_stable_xxd file should have been long gone after test_crypt.vim is run (and definitely no longer be staying around in test_quickfix.vim).

So try to clean up properly after a test script is run, just in case any X<file> is still around. During testing, a found a few leftover files, which I also fixed in the relevant test-file.

Unfortunately, the test workflow 'linux (huge, gcc, testgui, true, true)' now seems to fail with 'E1230: Encryption: sodium_mlock()' in test_crypt.vim. Hopefully this is only temporary hickup :/

### DIFF
--- a/src/testdir/runtest.vim
+++ b/src/testdir/runtest.vim
@@ -411,6 +411,24 @@ func RunTheTest(test)
   endif
 endfunc
 
+function Delete_Xtest_Files()
+  for file in glob('X*', v:false, v:true)
+    if file ==? 'XfakeHOME'
+      " Clean up files created by setup.vim
+      call delete('XfakeHOME', 'rf')
+      continue
+    endif
+    " call add(v:errors, file .. " exists when it shouldn't, trying to delete it!")
+    call delete(file)
+    if !empty(glob(file, v:false, v:true))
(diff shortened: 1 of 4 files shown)


chrisbra:
was merged, closing

PR #12668, opened 2023-07-14
Use Lua 5.1 compatible numeric escapes in Lua tests

jamessan:
Only Lua 5.2+ and LuaJIT understand hexadecimal escapes in strings.  Lua
5.1 uses decimal escapes:

> A character in a string can also be specified by its numerical value
> using the escape sequence \ddd, where ddd is a sequence of up to three
> decimal digits. (Note that if a numerical escape is to be followed by a
> digit, it must be expressed using exactly three digits.) Strings in Lua
> can contain any 8-bit value, including embedded zeros, which can be
> specified as '\0'.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
         features: [tiny, normal, huge]
         compiler: [clang, gcc]
         extra: [none]
+        luaver: [lua5.4]
         include:
           - features: tiny
             compiler: clang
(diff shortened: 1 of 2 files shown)


chrisbra:
I think you need to add a 2 more:
```
luaver: lua5.4
````
in the matrix where `extra` is defined and `features=huge` at least that's how I explain why it currently wants to install a package `lib-dev` in some of the `huge` jobs

PR #12673, opened 2023-07-14
Remove unnecessary 'exe'

balki:


jamessan:
Quoting `:help :bar`

> These commands see the '|' as their argument, and can therefore not be
followed by another Vim command:
>     :argdo
>     :autocmd
>     :bufdo

That's why the `:exe` _is_ necessary.

balki:
> Quoting `:help :bar`
> 
> > These commands see the '|' as their argument, and can therefore not be
> > followed by another Vim command:
> > :argdo
> > :autocmd
> > :bufdo
> 
> That's why the `:exe` _is_ necessary.

Thank you for the note! It seemed to work for me without `:exe`. Maybe it didn't and following autocommands got added to the group. I have updated to use below syntax from `:help autocmd-remove`

```
:au[tocmd]! [group]     Remove ALL autocommands.                              
                        Note: a quote will be seen as argument to the :autocmd
                        and won't start a comment.                            
                        Warning: You should normally not do this without a    
                        group, it breaks plugins, syntax highlighting, etc.   

```

tonymec:
> FWIW, `:autocmd` can be followed by a bar since [8.2.3626](https://github.com/vim/vim/releases/tag/v8.2.3626).

Well, it depends what you mean. If I write
```vim
autocmd VimEnter * Command1 | Command2 | Command3
```
will that be equivalent to
(case 1)
```vim
exe "autocmd VimEnter * Command1"
Command2 | Command3
```
or (case 2)
```vim
autocmd VimEnter * exe "Command1 | Command2 | Command3"
```
i.e. will the scope of the autocommand stop before the first bar (case 1), or only at the end of the line after including any bar-separated commands (case 2)? I would expect the latter.

Best regards,
Tony.

chrisbra:
hm, undecided. Opinions?

chrisbra:
thanks all, included

PR #12679, opened 2023-07-16
Manpager: apply s commands with nogdefault

toxeus:
Problem: The `s` command with `g` flag only substitutes
         one occurrence when `gdefault` is set.<br>
Solution: Unset `gdefault`.

### DIFF
--- a/runtime/plugin/manpager.vim
+++ b/runtime/plugin/manpager.vim
@@ -30,10 +30,10 @@ function s:ManPager()
   setlocal modifiable
 
   " Emulate 'col -b'
-  silent! keepj keepp %s/\v(.)\b\ze\1?//ge
+  exe 'silent! keepj keepp %s/\v(.)\b\ze\1?//e' .. (&gdefault ? '' : 'g')
 
   " Remove ansi sequences
-  silent! keepj keepp %s/\v\e\[%(%(\d;)?\d{1,2})?[mK]//ge
+  exe 'silent! keepj keepp %s/\v\e\[%(%(\d;)?\d{1,2})?[mK]//e' .. (&gdefault ? '' : 'g')
 
   " Remove empty lines above the header
   call cursor(1, 1)


chrisbra:
'gdefault' is a global option, so `setlocal` is not neccessary. Also, instead of disabling it (which may be surprising for users), I think it is better to simply adjust the `:s` call like this:

```
exe 'silent! keepj keepp %s/\v(.)\b\ze\1?//' .. (&gdefault ? '' : 'g') .. 'e'
```

Or you should restore the option again after the `:s` call

ping @Konfekt

toxeus:
I didn't know that `setlocal` doesn't work for `gdefault`. Of course we don't want to change a user's settings globally. Thanks for the input. Fixed.

Konfekt:
Thank you, looks good to me, and this makes sense:

```
	DEPRECATED: Setting this option may break plugins that are not aware
	of this option.  Also, many users get confused that adding the /g flag
	has the opposite effect of that it normally does.
	This option is not used in |Vim9| script.
```

runtime/plugin/manpager.vim
```diff
@@ -30,10 +30,10 @@ function s:ManPager()
   setlocal modifiable
 
   " Emulate 'col -b'
-  silent! keepj keepp %s/\v(.)\b\ze\1?//ge
+  silent! 'keepj keepp %s/\v(.)\b\ze\1?//e' .. (&gdefault ? '' : 'g')
 
   " Remove ansi sequences
-  silent! keepj keepp %s/\v\e\[%(%(\d;)?\d{1,2})?[mK]//ge
+  silent! 'keepj keepp %s/\v\e\[%(%(\d;)?\d{1,2})?[mK]//e' .. (&gdefault ? '' : 'g')
```
chrisbra: same here: wrap into an exe call

toxeus:
AppVeyor failed for reasons unrelated to this change.

Konfekt:
@chrisbra Isn't `. (&gdefault ? '' : 'g')` instead of `.. (&gdefault ? '' : 'g')` preferable for backwards compatibility? Other than that, please go ahead and merge.

chrisbra:
Hm, '..' comes from 0f248b006c2574abc00c9aa7886d8f33620eb8 which was commited more than 4 years ago. I think we should be good with using it in the runtime files (and it is being used elsewhere already).

PR #12681, opened 2023-07-17
Fix getcompletion for user commands

### LINKED ISSUE #12680: getcompletion() cmdline is broken when user defined command
### Steps to reproduce

```vim
command! -nargs=* -complete=command Test echo <q-args>
echo getcompletion('Test ', 'cmdline')
```

### Expected behaviour

It should be same with `echo getcompletion('', 'command')`.

```
['!', '#', ...]
```

### Actual behaviour

It is empty list.

```
[]
```

### Version of Vim

9.0.1677

### Environment

Manjaro Linux
foot
zsh

### Logs and stack traces

_No response_

chrisbra:
fixes #12680

### DIFF
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -3498,4 +3498,14 @@ func Test_rulerformat_position()
   call StopVimInTerminal(buf)
 endfunc
 
+func Test_usercmd_completion()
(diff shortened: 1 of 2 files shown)


Shougo:
I have tested it and it works for me.

PR #12693, opened 2023-07-20
Cannot wrap around in popup_filter_menu()

### LINKED ISSUE #12689: Can't cycle popup entries, can we?
Let's say that we have a popup defined, something like this:

    :call popup_menu(['aaa', 'bbb', 'ccc', 'ddd'], #{filter: 'Filter_function', callback: 'Callback_function'})

And if the cursor is on `aaa`, then pressing `k`, `<C-p>`, or `<Up>` doesn't move the cursor to `ddd`, right? Would be an UI enhancement, though, what do you think? :)

chrisbra:
Currently, it is not possible, to wrap around at the end of the list using e.g. down (and go back to the top) or up at the beginning of the list and go directly to the last item. So let's just allow this.

fixes vim/vim#12689

Adjut test


chrisbra:
waiting to hear some feedback. I think it is more consistent when compared to e.g. handling the pum menu, but not sure if this is a clear improvement.

Shane-XB-Qian:
> waiting to hear some feedback. I think it is more consistent when compared to e.g. handling the pum menu, but not sure if this is a clear improvement.

we've discuss in the mail, somehow didnot show up in the page, so paste it again (IMO):

not exactly same thing.
complete menu worked like special popup menu, and it really can filter matches items (etc) vs self-implement filter (this).
sounds like he just wish someone told him how to implement it, so mostly like a question.

-- 
shane.xb.qian

chrisbra:
I know, but I still think this is more consistent behaviour.

Shane-XB-Qian:
> I know, but I still think this is more consistent behaviour.

unfortunately, if so.
// perhaps we do really need some kinds of vote, hate to say, if so.

--
shane.xb.qian

chrisbra:
> // perhaps we do really need some kinds of vote, hate to say, if so

That's fine. You can put emojis for voting on PRs. There wasn't much, so I merged it because I am still convinced is't slightly better UI. Thanks for feedback anyhow.

Shane-XB-Qian:
though tiny, but this is not a bug fix, but a default behavior change, bram is missed, now it is time to and i supposed we need to keep the open source is open, to such decision I suggest vim members core team can setup a vote meeting to make such decision, vs/but not still bram's style, because bram is unique.

--
shane.xb.qian

PR #12696, opened 2023-07-22
Add commentstring for nix file format

keith:


dkearns:
You might want to set 'comments' as well.

keith:
Added!

dkearns:
LGTM, thanks.

chrisbra:
thanks both.

keith:
Thanks!

PR #12698, opened 2023-07-23
enable <D-*> mappings in GTK

dctucker:
# Description

As a developer who works in both MacVim and gvim on Linux using the same keyboard, it can be frustrating having to remember different key combinations or having to rely on system utilities to remap keys.

This change allows `<D-z>` `<D-x>` `<D-c>` `<D-v>` etc. to be recognized by the `map` commands, along with the `<D-S-...>` shifted variants. These mappings represent using the Super modifier on Linux, which typically corresponds to the Windows key on PC keyboards.

## Example .vimrc

```vimrc
if has('gui_gtk') && has("gui_running")
	nnoremap  <D-z>    u
	nnoremap  <D-S-Z>  <C-r>
	vnoremap  <D-x>    "+d
	vnoremap  <D-c>    "+y
	cnoremap  <D-v>    <C-R>+
	inoremap  <D-v>    <C-o>"+gP
	nnoremap  <D-v>    "+P
	vnoremap  <D-v>    "-d"+P
	nnoremap  <D-s>    :w<CR>
	inoremap  <D-s>    <C-o>:w<CR>
	nnoremap  <D-w>    :q<CR>
	nnoremap  <D-q>    :qa<CR>
	nnoremap  <D-t>    :tabe<CR>
	nnoremap  <D-S-T>  :vs#<CR><C-w>T
	nnoremap  <D-a>    ggVG
	vnoremap  <D-a>    <ESC>ggVG
	inoremap  <D-a>    <ESC>ggVG
	nnoremap  <D-f>    /
	nnoremap  <D-g>    n
	nnoremap  <D-S-G>  N
	vnoremap  <D-x>    "+x
endif
```

### DIFF
--- a/runtime/doc/map.txt
+++ b/runtime/doc/map.txt
@@ -21,7 +21,7 @@ manual.
    1.9 Using mappings				|map-typing|
    1.10 Mapping alt-keys			|:map-alt-keys|
    1.11 Mapping meta-keys			|:map-meta-keys|
-   1.12 Mapping super-keys			:map-super-keys
+   1.12 Mapping super-keys			|:map-super-keys|
    1.13 Mapping in modifyOtherKeys mode		|modifyOtherKeys|
    1.14 Mapping with Kitty keyboard protocol	|kitty-keyboard-protocol|
    1.15 Mapping an operator			|:map-operator|
@@ -984,13 +984,13 @@ For the Meta modifier the "T" character is used.  For example, to map Meta-b
 in Insert mode: >
 	:imap <T-b> terrible
 
-1.12 MAPPING SUPER-KEYS					:map-super-keys
+1.12 MAPPING SUPER-KEYS					*:map-super-keys*
 
-The Super modifier is available in GUI mode (when :gui_running is 1) for
+The Super modifier is available in GUI mode (when |gui_running| is 1) for
 GVim on Linux and MacVim on Mac OS. If you're on a Mac, this represents the
 Command key. The character "D" is used for the Super modifier.
 
-For example, to map Command-b in Insert mode:
+For example, to map Command-b in Insert mode: >
 	:imap <D-b> barritone
 
 1.13 MAPPING IN modifyOtherKeys mode			*modifyOtherKeys*
--- a/runtime/doc/map.txt
+++ b/runtime/doc/map.txt
@@ -21,9 +21,10 @@ manual.
    1.9 Using mappings				|map-typing|
    1.10 Mapping alt-keys			|:map-alt-keys|
    1.11 Mapping meta-keys			|:map-meta-keys|
-   1.12 Mapping in modifyOtherKeys mode		|modifyOtherKeys|
-   1.13 Mapping with Kitty keyboard protocol	|kitty-keyboard-protocol|
-   1.14 Mapping an operator			|:map-operator|
+   1.12 Mapping super-keys			:map-super-keys
+   1.13 Mapping in modifyOtherKeys mode		|modifyOtherKeys|
+   1.14 Mapping with Kitty keyboard protocol	|kitty-keyboard-protocol|
+   1.15 Mapping an operator			|:map-operator|
 2. Abbreviations		|abbreviations|
 3. Local mappings and functions	|script-local|
 4. User-defined commands	|user-commands|
@@ -983,8 +984,16 @@ For the Meta modifier the "T" character is used.  For example, to map Meta-b
 in Insert mode: >
 	:imap <T-b> terrible
 
+1.12 MAPPING SUPER-KEYS					:map-super-keys
 
-1.12 MAPPING IN modifyOtherKeys mode			*modifyOtherKeys*
+The Super modifier is available in GUI mode (when :gui_running is 1) for
+GVim on Linux and MacVim on Mac OS. If you're on a Mac, this represents the
+Command key. The character "D" is used for the Super modifier.
+
+For example, to map Command-b in Insert mode:
+	:imap <D-b> barritone
+
+1.13 MAPPING IN modifyOtherKeys mode			*modifyOtherKeys*
 
 Xterm and a few other terminals can be put in a mode where keys with modifiers
 are sent with a special escape code.  Vim recognizes these codes and can then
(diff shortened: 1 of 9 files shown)


tonymec:
How woud you produce a <D- (Command) modifier on GTK (i.e. on Unix/Linux with X11)? There is no Command key other than on Mac.
If you want to use the same mappings everywhere, don't use Cmd- mappings but only Shift- Ctrl- and/or Alt- (i.e. Meta- ).
In my experience, using F1 to F12 with or without Shift, except unshifted F1 and F10, as `{lhs}` of a mapping, is a good idea because these keys have no default function in Vim.

Best regards,
Tony.

chrisbra:
hm, can we have tests please? Can we make it a bit more explicit, that `<D-` is done using the `<super>` key on non-Macs?

dctucker:
Replying first to @tonymec:

> How woud you produce a <D- (Command) modifier on GTK (i.e. on Unix/Linux with X11)?
> There is no Command key other than on Mac.

My keyboard has two Command keys on it, and on my installed distro this key is mapped to Super. I'm pretty sure standard keyboards have Windows keys which behave similarly. I mentioned that the Super key is used for this in GTK in the [documentation](https://github.com/vim/vim/pull/12698/files#diff-ef8946d89919ad628b9ac65c49d3067e4e017b0980cac70eb25d832a59ef677dR3424) though I may have failed to make this clear in the commit message / PR description.

> If you want to use the same mappings everywhere, don't use Cmd- mappings but only Shift- Ctrl- and/or Alt- (i.e. Meta- ).

I've been using vim for more than 20 years, MacVim for 15 years, and the GTK port for about 5 years. Shouldn't muscle memory be considered a valid reason to configure a particular binding?

I already have mappings on these modifiers. When I go to press Cmd-anything on the GTK port presently, it ignores that the Cmd key is pressed entirely. It's an awful experience when you expect Cmd+C to copy a visual selection to the global clipboard only for gvim to instead erase everything because it thought you meant 'C' for change, and equally frustrating when Cmd+V doesn't paste as expected. Already attempted remedies include using xmodmap, xbindkey, KDE's keyboard shortcut settings and those in GNOME (pre GTK).

> In my experience, using F1 to F12 with or without Shift, except unshifted F1 and F10, as {lhs} of a mapping, is a good idea because these keys have no default function in Vim.

Others may have a different experience that is equally valid. My keyboard does not have dedicated F1-F12 keys; I access those through a mod layer. Not everyone uses the same hardware and people develop their own preferences, especially in an application such as vim where everything can be mapped. Accessibility is important and this pull request makes the GTK port more accessible by enabling another modifier.

dctucker:
@chrisbra

> hm, can we have tests please? Can we make it a bit more explicit, that `<D-` is done using the `<super>` key on non-Macs?

I may need a bit of advice on how to write a test for this. I imagine the main test case would be checking that `<D-` is recognized on Mac and GTK as a shortcut by the parser and is accepted as a binding by the map command. Would you be able to point me in the right direction with a link to documentation for test writing, or to an existing test that I could adapt towards this case? Are there additional test cases I should be considering?

I'd also appreciate suggestions on how to make this more explicit. I wasn't able to identify any other parts of the builtin documentation that mention this previously Mac-only functionality, so I just amended those sections to indicate that it's available on GTK. Please let me know what more is needed. I'm comfortable doing the technical writing if it's needed, I would just need to know what to write and where.

chrisbra:
There are a few things that could be tested. What happens, when pressing `i_CTRL-V` followed by e.g. `Super-M`? Does Vim insert `<D-M>`? Then if you map a super-key, does it show up in the mapmode listing? And finally of course, does it do what it is supposed to be doing when mapping `<D-..>` and feeding this character as input to `feedkeys()` ? What about mapping to a multi-byte character? Does Vim get the right key under xterm (with modifyOtherKeys set to 2 (enabled) and `0` (disabled)) What about kitty protocol for the kitty terminal?

See [test_mapping](https://github.com/vim/vim/blob/master/src/testdir/test_mapping.vim) See e.g. `Test_map_meta_quotes()` and `Test_map_meta_multibyte()` 

You may need to protect the test by testing for `if has("gui_gtk") || has("gui_gtk3") || has("macos")`. 

Regarding the Documentation, I guess this [line](https://github.com/vim/vim/blob/master/runtime/doc/intro.txt#L473)  should be adjusted as well. And perhaps add a section for the Super/Cmd-Key like [this one here](https://github.com/vim/vim/blob/4c0089d696b8d1d5dc40568f25ea5738fa5bbffb/runtime/doc/map.txt#L973)

vim-ml:
Hi,

On Sun, Jul 23, 2023 at 10:33 AM Casey Tucker ***@***.***>
wrote:

> @chrisbra <https://github.com/chrisbra>
>
> hm, can we have tests please? Can we make it a bit more explicit, that <D-
> is done using the <super> key on non-Macs?
>
> I may need a bit of advice on how to write a test for this. I imagine the
> main test case would be checking that <D- is recognized on Mac and GTK as
> a shortcut by the parser and is accepted as a binding by the map command.
> Would you be able to point me in the right direction with a link to
> documentation for test writing, or to an existing test that I could adapt
> towards this case? Are there additional test cases I should be considering?
>

There are quite a few test cases in src/testdir/test_termcodes.vim for
testing various terminal key codes.
For example, refer to the Test_special_term_keycodes() function.  For
MS-Windows, the tests are
in the test_mswin_event.vim file.  Refer to the LoopTestKeyArray() function.

Regards,
Yegappan


> I'd also appreciate suggestions on how to make this more explicit. I
> wasn't able to identify any other parts of the builtin documentation that
> mention this previously Mac-only functionality, so I just amended those
> sections to indicate that it's available on GTK. Please let me know what
> more is needed. I'm comfortable doing the technical writing if it's needed,
> I would just need to know what to write and where.
>
>
>

zeertzjq:
Super key is also supported by kitty keyboard protocol: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers

runtime/doc/map.txt
```diff
@@ -983,8 +984,16 @@ For the Meta modifier the "T" character is used.  For example, to map Meta-b
 in Insert mode: >
 	:imap <T-b> terrible
 
+1.12 MAPPING SUPER-KEYS					:map-super-keys
```
chrisbra: `*:map-super-keys*`

chdiza:
I would hope that this isn't restricted to GTK* versions, but also would work in the console---assuming the kitty keyboard protocol.

dctucker:
@chdiza I've focused my effort entirely on implemented this for the GTK version, so I can't guarantee that the scope of this PR will grow to accommodate other terminals, but I'm happy to ride in the passenger seat on a larger effort.

PR #12699, opened 2023-07-23
Fix mapset() not working for <SID>

### LINKED ISSUE #12697: maparg() does not restore for script mappings
### Steps to reproduce

```vim
" test_map.vim
inoremap <silent> <SID>Return <C-R>=<CR>
inoremap <script><buffer> <CR> <CR><SID>Return
```

```vim
" vimrc
source ~/work/test/test_map.vim

inoremap <C-j> <Cmd>call RestoreMap()<CR>

function RestoreMap()
  const map = maparg('<CR>', 'i', v:false, v:true)
  call mapset('i', v:false, map)
endfunction
```

1. `vim -Nu vimrc`
2. `i<C-j>`
3. `<CR>`

### Expected behaviour

It should insert `<CR>` like before restore mappings.

But actual behavior is `<SID>Return` is inserted in the buffer.
It seems `map-<script>` does not work.

### Version of Vim

9.0.1677

### Environment

Manjaro Linux
Alacritty
zsh

### Logs and stack traces

_No response_

zeertzjq:
Fix #12697

### DIFF
--- a/src/clientserver.c
+++ b/src/clientserver.c
@@ -34,7 +34,7 @@ server_to_input_buf(char_u *str)
     //  The last but one parameter of replace_termcodes() is TRUE so that the
     //  <lt> sequence is recognised - needed for a real backslash.
     p_cpo = (char_u *)"Bk";
-    str = replace_termcodes(str, &ptr, REPTERM_DO_LT, NULL);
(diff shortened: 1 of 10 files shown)


yegappan:

PR #12700, opened 2023-07-23
allow inputlist to return after entering a specified number of digits

EliasA5:
This would allow us to implement more interactive menus for similar commands and possibly free up some key-bindings.

For example we can add the following to .vimrc to pick between tab or space indentation style, and the user will only need to hit 1 number **(without pressing enter)** to choose:

```vimscript
function! UseTabs()
  set tabstop=4     " Size of a hard tabstop (ts).
  set shiftwidth=4  " Size of an indentation (sw).
  set noexpandtab   " Always uses tabs instead of space characters (noet).
  set autoindent    " Copy indent from current line when starting a new line (ai).
endfunction

function! UseSpaces()
  set tabstop=2     " Size of a hard tabstop (ts).
  set shiftwidth=2  " Size of an indentation (sw).
  set expandtab     " Always uses spaces instead of tab characters (et).
  set softtabstop=0 " Number of spaces a <Tab> counts for. When 0, featuer is off (sts).
  set autoindent    " Copy indent from current line when starting a new line.
  set smarttab      " Inserts blanks on a <Tab> key (as per sw, ts and sts).
endfunction

function PickTabStyle()
  let l:opts = ['Pick indentation style', '1. Tabs', '2. Spaces']
  let l:idx = inputlist(opts,1)
  if idx == 1
    call UseTabs()
  elseif idx == 2
    call UseSpaces()
  else
    echo "\n"
    echoerr "invalid option"
  endif
endfunction

nnoremap gt :call PickTabStyle()<cr>

```


chrisbra:
I am sorry to be the one, but is that really that much of an improvement? I am not conviced this is needed, but please convince me :)

src/testdir/test_functions.vim
```diff
@@ -2037,6 +2037,12 @@ func Test_inputlist()
   call assert_equal(2, c)
   call feedkeys(":let c = inputlist(['Select color:', '1. red', '2. green', '3. blue'])\<cr>3\<cr>", 'tx')
   call assert_equal(3, c)
+  call feedkeys(":let c = inputlist(['Select color:', '1. red', '2. green', '3. blue'], 1)\<cr>1", 'tx')
+  call assert_equal(1, c)
+  call feedkeys(":let c = inputlist(['Select color:', '1. red', '2. green', '3. blue'], 2)\<cr>10", 'tx')
+  call assert_equal(10, c)
+  call feedkeys(":let c = ['Select color:', '1. red', '2. green', '3. blue']->inputlist(1)\<cr>2", 'tx')
+  call assert_equal(2, c)
```
yegappan, line 2045: In case this PR is accepted and merged, can you add a test for passing a non-number type (e.g. List) as the num_digits argument and passing a negative number as the argument?

Also, the Test_inputlist() function in src/testdir/test_vim9_builtin.vim should be enhanced to add tests for this.

PR #12702, opened 2023-07-24
Pip requirements files are not recognized

ObserverOfTime:
This only includes the `.pip` extension and the most common filename ('requirements.txt').

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1558,6 +1558,10 @@ au BufNewFile,BufRead *.rcp			setf pilrc
 " Pine config
 au BufNewFile,BufRead .pinerc,pinerc,.pinercex,pinercex		setf pine
 
+" Pip requirements
+au BufNewFile,BufRead *.pip			setf requirements
+au BufNewFile,BufRead requirements.txt		setf requirements
+
 " Pipenv Pipfiles
 au BufNewFile,BufRead Pipfile			setf toml
 au BufNewFile,BufRead Pipfile.lock		setf json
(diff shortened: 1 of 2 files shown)


dkearns:
I don't suppose there's some creative way to disambiguate requirements.txt from a generic text file is there?  I actually have some plain text files with that name.

ObserverOfTime:
The contents just be a list of words so no.

ObserverOfTime:
> According to [here](https://pip.pypa.io/en/stable/reference/requirements-file-format/), can't we check, that the first X lines only consist of Comments for single words optionally followed by a version?

The comments are not mandatory. The following is a perfectly valid requirements file:

```
foobar
```

chrisbra:
> The comments are not mandatory. The following is a perfectly valid requirements file:
> 
> foobar

Yes. What I am trying to say is, does it make sense to only set the `pymanifest` filetype, if the file either contains single words per line (optionally allowing Comments and version identifier)?

ObserverOfTime:
No since it can have paths, urls, extras, etc..
Should I just drop the filename?

benknoble:
Here's two other plugins that support these files; perhaps they can be used as inspiration: https://github.com/benknoble/requirements.txt.vim, https://github.com/raimon49/requirements.txt.vim

chrisbra:
yeah, I guess it is fine to include then. It may be a minor annoyance when using  `requirements.txt` file for actual text files, but that may be acceptable I guess?

chrisbra (approved):

ObserverOfTime:
> It may be a minor annoyance when using `requirements.txt` file for actual text files, but that may be acceptable I guess?

Well, there won't be any highlighting without a plugin so it should be fine.

PR #12704, opened 2023-07-24
Avoid restoring incorrect heights in `win_size_restore()`

seandewar:
Problem: changing `'showtabline'` or `'cmdheight'` in the cmdwin restores incorrect window heights after closing the cmdwin.

This may produce a gap between the cmdline and the window above.

Solution: restore window sizes only if the number of lines available for windows changed; subtract the rows of the tabline, cmdline and last window's statusline from `'lines'` (other statuslines don't matter).

### DIFF
--- a/src/window.c
+++ b/src/window.c
@@ -5943,7 +5943,6 @@ shell_new_columns(void)
  */
     void
 win_size_save(garray_T *gap)
-
 {
     win_T	*wp;
 
@@ -5951,8 +5950,8 @@ win_size_save(garray_T *gap)
     if (ga_grow(gap, win_count() * 2 + 1) == FAIL)
 	return;
 
-    // first entry is value of 'lines'
-    ((int *)gap->ga_data)[gap->ga_len++] = Rows;
+    // first entry is the total lines available for windows
+    ((int *)gap->ga_data)[gap->ga_len++] = ROWS_AVAIL - last_stl_height(FALSE);
 
     FOR_ALL_WINDOWS(wp)
     {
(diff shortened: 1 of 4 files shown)


src/window.c
```diff
@@ -5943,16 +5943,15 @@ shell_new_columns(void)
...
     win_T	*wp;
 
     ga_init2(gap, sizeof(int), 1);
     if (ga_grow(gap, win_count() * 2 + 1) == FAIL)
 	return;
 
-    // first entry is value of 'lines'
-    ((int *)gap->ga_data)[gap->ga_len++] = Rows;
+    // first entry is the total lines available for windows
+    ((int *)gap->ga_data)[gap->ga_len++] = Rows - tabline_height() - p_ch;
```
zeertzjq:
```suggestion
    ((int *)gap->ga_data)[gap->ga_len++] = ROWS_AVAIL
```

src/window.c
```diff
@@ -5974,7 +5973,7 @@ win_size_restore(garray_T *gap)
     int		i, j;
 
     if (win_count() * 2 + 1 == gap->ga_len
-	    && ((int *)gap->ga_data)[0] == Rows)
+	    && ((int *)gap->ga_data)[0] == Rows - tabline_height() - p_ch)
```
zeertzjq:
```suggestion
	    && ((int *)gap->ga_data)[0] == ROWS_AVAIL)
```

src/window.c
seandewar: Cheers!

chrisbra:
I think this looks good, but I have to verify it again later.

PR #12707, opened 2023-07-25
PyPA manifest files are not recognized

ObserverOfTime:
The filetype is subject to change since there's no official name for it.

I will also add a basic syntax file later.

Docs: https://packaging.python.org/en/latest/guides/using-manifest-in/

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1676,6 +1676,9 @@ au BufNewFile,BufRead *.pk			setf poke
 " Protocols
 au BufNewFile,BufRead */etc/protocols		setf protocols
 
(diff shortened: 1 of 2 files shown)


chrisbra:
thank you. Let me include this now.

ObserverOfTime:
What about the patch number?
Will you be adding those in separate commits?

chrisbra:
oh, you mean because of `src/testdir/test_filetype.vim` ? I forgot... 🙈 There were so many small runtime file updates, I wanted to get them merged.

chrisbra:
I'll try to do better, sorry for inconveniences.

PR #12711, opened 2023-07-27
Abort creation of popup windows, if there are errors

chrisbra:
In case of an error when creating a new popup, e.g. because of an expected option, let's abort creation of the popup, instead of displaying it.


cvwillegen:
Code compiles and looks sane. Tests fail without the modification. Screen layout looks good after the modification.

LGTM

The build failure(s) look(s) unrelated.

PR #12713, opened 2023-07-28
Fix 'incsearch' restoring wrong viewport with 'smoothscroll'

zeertzjq:


cvwillegen:
Checked that test fails without the patch and succeeds with the patch. Noticed that scrolled columns are correctly restored after the ``<Esc>`` is pressed. Also noticed that without the patch, after ``/xx`` the view port jumps up, and after typing the third ``x`` the view port correctly shows the text searched for. Not sure if that was the intended behavior, but it looks like this fixes a 'related' bug.

LGTM.

chrisbra (approved):

PR #12714, opened 2023-07-29
Fix nixInherit matcher in nix.vim

equill:
Perform the lookahead in `nixInheritAttributeScope`, then hand over to a new region called `nixInheritAttributeSubExpr`, which sets the match start to one char after the opening bracket to avoid a double-match.

Finally, only do a lookahead to `)` in `nixInheritAttributeSubExpr` (and thus make sure the region is closed to not get a match of `nixParen` here) and let `nixInheritAttributeScope` close the bracket.

### DIFF
--- a/runtime/syntax/nix.vim
+++ b/runtime/syntax/nix.vim
@@ -5,7 +5,7 @@
 " Acknowledgement: Based on vim-nix maintained by Daiderd Jordan <daiderd@gmail.com>
 "                  https://github.com/LnL7/vim-nix
 " License:         MIT
-" Last Change:     2022 Dec 06
+" Last Change:     2023 Jul 28
 
 if exists("b:current_syntax")
   finish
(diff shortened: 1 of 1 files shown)


equill:
> did you check with the maintainer of the file?

I am the maintainer :)

I've accepted responsibility for merging changes to this file from https://github.com/LnL7/vim-nix. This was negotiated in https://github.com/vim/vim/pull/11646

This is my first time actually exercising it, so if I need to do this differently, please let me know and I'll make sure to do it that way.

chrisbra:
oh okay, you did everything. It's just that from your github username, it was not entirely obvious, that you are the maintainer. So you may please ammend the syntax file and mention your github username as well please (or use your name from the Maintainer line in your github account).

equill:
That makes perfect sense. Thanks also for the affirmation that I did it right.
I've updated the maintainer info in the file, and added that commit.

chrisbra:
included as of 72904d5fda0a1bc78f4bc9080c158226e6e90d5c

PR #12729, opened 2023-08-05
Specify global/local 'undolevels' to avoid unintentional changes.

drmikehenry:
When saving and restoring 'undolevels', the constructs `&undolevels` and `:set undolevels` are problematic.

The construct `&undolevels` reads an unpredictable value; it will be the local option value (if one has been set), or the global option value (otherwise), making it unsuitable for saving a value for later restoration.

Similarly, if a local option value has been set for 'undolevels', temporarily modifying the option via `:set undolevels` changes the local value as well as the global value, requiring extra work to restore both values.

Saving and restoring the option value in one step via the construct `:let &undolevels = &undolevels` appears to make no changes to the 'undolevels' option, but if a local option has been set to a different value than the global option, it has the unintended effect of changing the global 'undolevels' value to the local value.

Update the documentation to explain these issues and recommend explicit use of global and local option values when saving and restoring.  Update some unit tests to use `g:undolevels`.

### DIFF
--- a/runtime/doc/undo.txt
+++ b/runtime/doc/undo.txt
@@ -111,9 +111,19 @@ use CTRL-G u.  This is useful if you want an insert command to be undoable in
 parts.  E.g., for each sentence.  |i_CTRL-G_u|
 
 Setting the value of 'undolevels' also closes the undo block.  Even when the
-new value is equal to the old value.  In |Vim9| script: >
-	&undolevels = &undolevels
+new value is equal to the old value.  Use `g:undolevels` to explicitly read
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks, looks good. Let me merge this.

chrisbra:
after CI finishes I meant

chrisbra:
so here we have it: should this update the minor patch number or not?

yegappan:
>
> so here we have it: should this update the minor patch number or not?
>

As this PR changes a src/testdir file, the minor patch number should be updated.

chrisbra (approved):

PR #12740, opened 2023-08-06
Bring Bram's last patch (9.0.1678) to GitHub

k-takata:
Bram's last patch was 9.0.1678: http://ftp.vim.org/pub/vim/patches/9.0/9.0.1678

Unfortunately, it was not committed to this repository.
Bring his last commit to here.

```
patch 9.0.1678: blade files are not recognized

Problem:    Blade files are not recognized.
Solution:   Add a pattern for Blade files. (closes #12650)
```

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -254,6 +254,9 @@ au BufNewFile,BufRead named*.conf,rndc*.conf,rndc*.key	setf named
 au BufNewFile,BufRead named.root		setf bindzone
 au BufNewFile,BufRead *.db			call dist#ft#BindzoneCheck('')
 
+" Blade
+au BufNewFile,BufRead *.blade.php		setf blade
+
 " Blank
 au BufNewFile,BufRead *.bl			setf blank
 
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -122,6 +122,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     bicep: ['file.bicep'],
     bindzone: ['named.root', '/bind/db.file', '/named/db.file', 'any/bind/db.file', 'any/named/db.file'],
     bitbake: ['file.bb', 'file.bbappend', 'file.bbclass', 'build/conf/local.conf', 'meta/conf/layer.conf', 'build/conf/bbappend.conf', 'meta-layer/conf/distro/foo.conf'],
+    blade: ['file.blade.php'],
     blank: ['file.bl'],
     blueprint: ['file.blp'],
     bsdl: ['file.bsd', 'file.bsdl'],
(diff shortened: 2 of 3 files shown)


k-takata:
Done.

arp242:
Maybe this should be released/tagged as Vim 9.1? I know that in #12736 @chrisbra said that he wants to "merge any un-contraditionary changes" first, but unless there's some critical stuff in there that really ought to be fixed, I think it would be kind of nice to make a last "pure" Bram-Vim release, before we continue with post-Bram-Vim. Maybe it's silly, but I think it would be nice, mostly for emotional reasons rather than technical.

tonymec:
I confirm that this patch has reached the Mercurial mirror, but AFAICT it hasn't been tagged v9.0.1678 there. I was a little surprised to see a mention of "Bram@vim.org" (I would have expected "[Bb]ram@moolenaar.net") but then I though that if it's good enough for Christian it's good enough for me : I ought not to expect a third-party takeover under Christian's nose.

tonymec:
P.S. "Bram@vim.org" was the author neme used in previous changesets so I ought not to have been surprised after all.

wsdjeg:
@chrisbra it is better to use rebase merged feature of github and ignore merge commit.

zeertzjq:
This may be only PR where the merge button can be used, because in most PRs it's not possible to determine the patch and the change to version.c in advance.

chrisbra:
oh yes, will tag it in a second

ObserverOfTime:
> Maybe this should be released/tagged as Vim 9.1? I know that in https://github.com/vim/vim/discussions/12736 @chrisbra said that he wants to "merge any un-contraditionary changes" first, but unless there's some critical stuff in there that really ought to be fixed, I think it would be kind of nice to make a last "pure" Bram-Vim release, before we continue with post-Bram-Vim. Maybe it's silly, but I think it would be nice, mostly for emotional reasons rather than technical.

I'm of the same opinion, but since this was a patch that Bram released (just not on GitHub) it should be in 9.0.

chrisbra:
Guys, I don't want that now, because Vim has always included contributions from various other people. so it's not like this is changing much now.  Let's first clean up the PRs and additional runtime file updates and then go ahead with a new release. It's adding a lot more work if I need to find out how to do it properly now (including adding patch descriptions), when I already have a lot of other things to take care of right now (including my family). So this has to wait a bit.

chrisbra:
> This may be only PR where the merge button can be used, because in most PRs it's not possible to determine the patch number and the change to version.c in advance.

Yes, I'll most likely do this on the command line and apply the patches there, adding minor patch number and tags and push out (I assume similar to how Bram did it) 🤷‍♂️

PR #12741, opened 2023-08-07
Update `.wast` files syntax highlighting

rhysd:
This is a patch to upstream from https://github.com/rhysd/vim-wasm.

Recently Wasm introduced many changes. This patch does not fully follow them, but it includes several improvements.


chrisbra:
thanks, let me include this.

PR #12746, opened 2023-08-08
Include libvterm changes from revision 818 to 839

zeertzjq:
Notable fix: libvterm now handles DECSM/DECRM with multiple arguents,
so several ncurses programs (e.g. nnn) can enable mouse properly when
run in Vim's terminal in XTerm.


chrisbra:
can you please squash? Otherwise looks good I think

chrisbra (approved):

chrisbra:
thanks, updated it as of ec8deb092cd645fcc5c3bd3117f61c9bd47f5b03

PR #5458, opened 2020-01-08
Add LICENSE file

othree:
As I talked at vim_dev last year[[1][1]]. I am working on making GitHub able to recognize Vim License. After some communication and contribution. The Licensee(used by GitHub) can detect Vim License with supporting project name field now.

The next step is to add the license to choosealicense[[2][2]]. To do that I need provide three notable project[[3][3]] using Vim License which can be detect by Licensee. I believe Vim itself should be the first item of the list. So this PR is to submit the LICENSE file to Vim code repository. It is required for the Licensee to detect the license in use.

[1]:https://groups.google.com/forum/#!searchin/vim_dev/vim$20license%7Csort:date/vim_dev/DlTVMew1ZSo/b9QCTlXgEAAJ
[2]:https://github.com/github/choosealicense.com/
[3]:https://github.com/github/choosealicense.com/blob/gh-pages/CONTRIBUTING.md#adding-a-license

PR #12749, opened 2023-08-08
Farewell to Bram and dedicate upcoming Vim 9.1 to him

chrisbra:


runtime/doc/version9.txt
```diff
@@ -43,6 +43,17 @@ when the release was being prepared.  Sven was a long time supporter of Vim.
 He registered the vim.org domain and created the first Vim website.  We will
 remember him!
 
+							*Bram-Moolenaar* *Bram*
+Vim version 9.1 is dedicated to Bram Moolenaar, who passed away August 3rd 2023
```
yegappan, line 47: "who passed away on"?

cvwillegen: I would suggest "who passed away August 3rd, 2023". The "on" is optional to my (non-native) eyes.

chrisbra:
I have slightly updated it according to the suggestions, and replaced Brams email address in a few more places (e.g. the license and for reporting bugs). Hope you all find that okay.

runtime/doc/version9.txt
k-takata: You mean adding "," between the date and the year?

Imusing: @k-takata he means remove the word "on"

runtime/autoload/ccomplete.vim
```diff
@@ -2,9 +2,10 @@ vim9script noclear
 
 # Vim completion script
 # Language:     C
-# Maintainer:   Bram Moolenaar <Bram@vim.org>
+# Maintainer: The Vim Project <https://github.com/vim/vim>
+# Former Maintainer:   Bram Moolenaar <Bram@vim.org>
```
k-takata:
```suggestion
# Maintainer:   The Vim Project <https://github.com/vim/vim>
# Former Maintainer: Bram Moolenaar <Bram@vim.org>
```
How about aligning "Language:", "Maintainer:" and "Last Change:"?
Not sure if "Former Maintainer:" should be also aligned.

runtime/doc/version9.txt
mcnaveen: English is not my native language. I suggest.

> Vim version 9.1 is dedicated to Bram Moolenaar, who passed away `on` August 3rd 2023

runtime/autoload/ccomplete.vim
chrisbra: okay, will adjust later.

chrisbra:
are we okay including this?

Shane-XB-Qian:
say something in `readme` as well, and something may need modification too.

// bram used to announce something via vim_announce, (not much people in vim_dev for some reason seems e.g dup mails from itself and github), we may should keep that tradition as well.

--
shane.xb.qian

Shane-XB-Qian:
@Christian you would modify readme.md for the contact of bram, and say something there, right?

--
shane.xb.qian

chrisbra:
Yeah, should mention it there as well..ThanksAm 13.08.2023 um 11:33 schrieb Shane-XB-Qian ***@***.***>:﻿
@Christian you would modify readme.md for the contact of bram, and say something there, right?

--
shane.xb.qian

src/message.c
```diff
@@ -1114,7 +1114,7 @@ ex_messages(exarg_T *eap)
 	    msg_attr(
 		    // Translator: Please replace the name and email address
 		    // with the appropriate text for your translation.
-		    _("Messages maintainer: Bram Moolenaar <Bram@vim.org>"),
+		    _("Messages maintainer: The Vim Project"),
```
RestorerZ, line 1117: Maybe you should add an email address <vim-dev@vim.org>

ychin:
This change modified the Vim license file. I'm just assuming that's an ok thing to do (was previously going to ask about this on a separate discussion as license issues can be tricky), but wondering if this affects other things like the GitHub license detection (it still seems to work so far), and the spdx entry: https://spdx.org/licenses/Vim.html, as well as https://vim-license.dev/.

Pinging @othree who filed https://github.com/vim/vim/pull/5458 and also did a lot of work on getting the Vim license recognized.

othree:
Thanks for pinging me, I can help to communicate and send the PR. There are two major repos we need to send PR:

- SPDX
- choosealicense, which is the source for license recognization.

But I think it will cause some issue if we change the text in choosealicense directly. Which will cause the existing repo uses the Vim license to become unrecognized. The ideal case is we have a new version of the license, such as Vim License 1.1. Then we didn't need to worry the existing repo affected. But it also will cause another issue. The choosealicense only accepts widely used license. Vim License 1.1 will be a new license and is not widely used at this moment. So there will be a period of time that GitHub won't recognize it.

ychin:
When you say "existing repo", do you mean this one (https://github.com/vim/vim/)? It seems like GitHub is still recognizing it despite this PR being merged already, but I wonder if there is a time delay in rescanning done by GitHub. If that's the case that means GitHub will soon stop recognizing the license?

othree:
The "existing repo" means every repo uses Vim License such as "tpope/vim-pathogen". "vim/vim" is one of them. I will try to verify my assumption later(maybe next week). GitHub has a special cache for the result of license detection. So it is possible that the license result is not changed now. I am not sure when it will be invalid.

chrisbra:
I was a bit hesitant to change the license file directly, but I did not see any alternatives. Surely, just changing the maintainer line should be allowed and okay, since we did not touch the actual license clauses. But if you think of a different solution, please let me know.

othree:
A quick update:

The Licensee can recognize the new version of the Vim License text

vim-license-licensee

The similarity is 98.45%. The threshold is 98%, so there is not much space for other differences. I will spend some time checking other projects using the Vim License and have the software name correctly modified. Currently, I feel positive about the results. It seems Licensee has a better algorithm to compare the license text with substitutions than before.

If the result is good. We don't need to worry about the existing software in the wild. We just need to:
1. Discuss how to deal with this case with the SPDX
2. Update the license text or create a new version of the license based on the answer of 1
3. Update the license text on choosealicense.com if necessary

chrisbra:
that are good news, thanks!

PR #12751, opened 2023-08-09
Catch sodium_mlock() not possible in the CI

chrisbra:
sodium_mlock() seems to fail consistently on the Github CI. Perhaps [sodium_mlock()](https://libsodium.gitbook.io/doc/memory_management#text-locking-memory) is called too often or with too much memory by the runners so that this starts failing.

Let's just try to catch this and skip the test, when this starts happening.


chrisbra:
I hope this will fix the constantly failing CIs, so that after merging this, the CI reports get actually useful again.

Issue #12595, opened 2023-06-25
Ctrl-[ no longer behaves as Esc on Czech keyboard on Windows (it worked in g/vim 8)


risa2000:
### Steps to reproduce

Using gVim on Windows (version 9.0.1664).
Pressing `Ctrl-[` (or FWIW `Ctrl-]`) no longer works as expected (as `Esc`, or `goto tag`). It writes the Czech letters instead:
`Ctrl-[` writes `ú`,
`Ctrl-]` writes `)`.
which are the actual letters on the Czech layout at the corresponding keys.
I did not check other `Ctrl-<??>` combinations.

### Expected behaviour

I have found two similar issues already closed here: #11831, #10454
but since I am not familiar with Belgian layout, or behavior the Belgian users expect, I cannot really say if it concerns my case. But it definitely does not resolve it.

From what I read in those and some other threads about the `Ctrl` handling I assume there is a confusion among some people how `Ctrl-[` works on non English layouts. I will try to explain on the Czech layout on Windows.

Note: For someone not familiar with Czech layouts, there are several "Czech" layouts with different degree of localization, but the one I will be talking about is the one named just `Czech` (the others are usually certain mix of English and Czech layout).

On the Czech layout the key right to `P` (which on the English layout is the one with `[`) has a Czech letter `ú`. The symbol `[` is however written by `AltGr+f` on the Czech layout. For English key `]` (which is right to `[`) the situation on the Czech keyboard is similar again - there is `)` on Czech layout and `]` is written by `AltGr+g`.

When Microsoft was localizing the keyboard they actually did a smart thing and did not map `Ctrl-[` to `Ctrl-AltGr-f` (as happened on some linux distros), but realized that `Ctrl-[` is indeed a control code (and not an actual letter) and that the usability should be more important than formal correctness. We may say they did not localize `Ctrl-[` by the character, but by the key (i.e. kept the same key in the layout regardless its localized character).

I have been used to use `Ctrl-[` for `Esc` and it worked in linux terminal, or Windows, regardless the keyboard layout for the past 30 years. The only dire cases were GUIs in some linux distros where they indeed expected users to write `Ctrl-AltGr-f`.

I am not sure what happened in g/Vim from ver 8 to ver 9, but it looks to me, like someone made an effort to actually break this behavior for some reason. I wonder what was the motivation and what is the proposed way forward?

### Version of Vim

9.0.1664

### Environment

Windows 10 Pro 64 bit,
[...]

Issue #10615, opened 2022-06-25
<C-_> bug still not fixed in v8.2.5157: different behavier between vim and gvim, about <C-_>


dantezhu:
### Steps to reproduce

1. open vim.exe, type `Ctrl` and `-`, wait sometimes, it will show like this:
    ![image](https://user-images.githubusercontent.com/1105212/174634949-2cc4fe11-12c2-4423-bb37-b59953265afc.png)

2. open gvim.exe, type as before, it will show like this:
    ![image](https://user-images.githubusercontent.com/1105212/174635040-86c33601-001f-44f8-b5aa-f76b2736feb8.png)

3. The different behavier will make some bugs, such as tcomment.vim use `<c-_><c-_>` to comment line, but now it counld not works well.

4. open gvim.exe, type `Ctrl` + `Shift` + `-`, it will show the same value as vim.exe:
    ![image](https://user-images.githubusercontent.com/1105212/175753880-2aa0c9e0-7395-4c9c-af70-3262eb0b61ea.png)

    but this is not same behavier with vim, and not same with old gvim.

### Expected behaviour

type `Ctrl` and `-`, gvim should output as vim.
![image](https://user-images.githubusercontent.com/1105212/174634949-2cc4fe11-12c2-4423-bb37-b59953265afc.png)

### Version of Vim

v8.2.5157

### Environment

windows 10 x64
china

dantezhu:
this commit: [patch 8.2.5157] has not fix the bug(https://github.com/vim/vim/commit/4e0fc8956649d3208aeaa1642c5efc44e385d77a)

dantezhu:
pls help to check it again, tks.
@LemonBoy

dantezhu:
@brammool

ladayaroslav:
But https://github.com/vim/vim/issues/10579 is, indeed, fixed (WFM, at least).
Thank you, @LemonBoy!

Though now I noticed some strange behavior related to langmap (with any value of langremap).
Using correct langmap for Russian:
```
set langmap=ёйцукенгшщзхъфывапролджэячсмитьбюЁЙЦУКЕНГШЩЗХЪФЫВАПРОЛДЖЭЯЧСМИТЬБЮ;`qwertyuiop[]asdfghjkl\;'zxcvbnm\\,.~QWERTYUIOP{}ASDFGHJKL:\"ZXCVBNM<>
```

[...]

PR #12752, opened 2023-08-09
W32 ime classic-n-experimental

ant0sha:
To untangle complicated circumstances introduced in v8.2.4807 old way of interpreting keyboard input via TranslateMessage() is reverted to be default again. At the same time, new/experimental "input method" which is using ToUnicode() Win API call is still available, in case some users are already dependent on it. To change between both "input methods" in run-time testing events infrastructure is (mis?)used:

```
call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'experimental'})
call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'classic'})
```

Documentation is updated accordingly.

### DIFF
--- a/runtime/doc/gui_w32.txt
+++ b/runtime/doc/gui_w32.txt
@@ -480,5 +480,13 @@ Similarly, in case you need to turn back "classic" keycode translation method
 >
 	:call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'classic'})
 <
+Alternatively (this method is especially useful for the TINY GVIM build, where
+test_mswin_event() cannot be called), environment variable
+VIM_KEYCODE_TRANS_STRATEGY can be set to the desired value (`experimental' or
+`classic'), to override default. E.g., type in dos prompt:
+>
+        set VIM_KEYCODE_TRANS_STRATEGY=experimental
+        gvim.exe
+<
 
  vim:tw=78:sw=4:ts=8:noet:ft=help:norl:
--- a/runtime/doc/gui_w32.txt
+++ b/runtime/doc/gui_w32.txt
@@ -453,28 +453,32 @@ To try out if XPM support works do this: >
 You may need to get the vim16x16.xpm file from github:
 https://github.com/vim/vim/blob/master/runtime/vim16x16.xpm
 
-Experimental w32 keyboard "input method"       *w32-experimental-kbd-ime*
+                            
+Keycode translation strategy    *w32-experimental-keycode-trans-strategy*
 
-Since v8.2.4807 W32 GVIM was changed over to experimental "input method" with
-the aim to be able to use more keyboard shortcuts and especially supporting
-non-standard keyboard layouts. In order to implement this support Win API
-TranslateMessage() call was dropped, and instead of that recognition of
-keycode was changed over to ToUnicode() Win API call. This approach uncovered
-numerous corner cases, which are apparently covered by TranslateMessage()
-implementation, each of it is necessary to be dealt on individual basis.
-Therefore decision was taken to make this functionality experimental and
-"classic" "input method" was made default again.
+Since v8.2.4807 W32 GVIM was changed over to experimental keycode translation
+method with the aim to be able to use more keyboard shortcuts and especially
+supporting non-standard keyboard layouts. In order to implement this support
+Win API TranslateMessage() call was dropped, and instead of that recognition
+of keycode was changed over to ToUnicode() Win API call. This approach
+uncovered numerous corner cases, which are apparently covered by
+TranslateMessage() implementation, each of it is necessary to be dealt with on
+individual basis. Therefore decision was taken to declare this functionality
+experimental for the time being and to recover "classic" keycode translation
+method as default again.
 
-Discussion about use of "experimental" input method will probably last some
-time yet. In the meantime, if you are impacted by change over back to "classic"
-input method in W32 GVIM, you can enable "experimental" input method again in
-your vimrc using following snippet:
+Discussion about use of "experimental" keycode translation method will
+probably last some time yet. In the meantime, if you are impacted by this
+change over back to "classic" keycode translation method in W32 GVIM, you can
+enable "experimental" translation method again in your vimrc using following
+snippet:
 >
-	:call test_mswin_event('set_w32_ime', {'ime': 'experimental'})
+	:call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'experimental'})
 <
-Similarly, in case you wish to turn back "classic" input method, please use:
+Similarly, in case you need to turn back "classic" keycode translation method
+(for example for testing purposes), please use:
 >
-	:call test_mswin_event('set_w32_ime', {'ime': 'classic'})
+	:call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'classic'})
 <
 
  vim:tw=78:sw=4:ts=8:noet:ft=help:norl:
--- a/runtime/doc/gui_w32.txt
+++ b/runtime/doc/gui_w32.txt
@@ -453,7 +453,7 @@ To try out if XPM support works do this: >
 You may need to get the vim16x16.xpm file from github:
 https://github.com/vim/vim/blob/master/runtime/vim16x16.xpm
 
-                            
+
 Keycode translation strategy    *w32-experimental-keycode-trans-strategy*
 
 Since v8.2.4807 W32 GVIM was changed over to experimental keycode translation
--- a/runtime/doc/gui_w32.txt
+++ b/runtime/doc/gui_w32.txt
@@ -453,5 +453,28 @@ To try out if XPM support works do this: >
 You may need to get the vim16x16.xpm file from github:
 https://github.com/vim/vim/blob/master/runtime/vim16x16.xpm
 
+Experimental w32 keyboard "input method"       *w32-experimental-kbd-ime*
+
+Since v8.2.4807 W32 GVIM was changed over to experimental "input method" with
+the aim to be able to use more keyboard shortcuts and especially supporting
+non-standard keyboard layouts. In order to implement this support Win API
+TranslateMessage() call was dropped, and instead of that recognition of
+keycode was changed over to ToUnicode() Win API call. This approach uncovered
+numerous corner cases, which are apparently covered by TranslateMessage()
+implementation, each of it is necessary to be dealt on individual basis.
+Therefore decision was taken to make this functionality experimental and
+"classic" "input method" was made default again.
+
+Discussion about use of "experimental" input method will probably last some
+time yet. In the meantime, if you are impacted by change over back to "classic"
+input method in W32 GVIM, you can enable "experimental" input method again in
+your vimrc using following snippet:
+>
+	:call test_mswin_event('set_w32_ime', {'ime': 'experimental'})
+<
+Similarly, in case you wish to turn back "classic" input method, please use:
+>
+	:call test_mswin_event('set_w32_ime', {'ime': 'classic'})
+<
 
  vim:tw=78:sw=4:ts=8:noet:ft=help:norl:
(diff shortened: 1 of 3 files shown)


k-takata:
The word "input method" has already different meaning, I think.
https://en.wikipedia.org/wiki/Input_method

It might be better to use another term. (Not sure what is the best term, though. "key event handling"?)

ant0sha:
> The word "input method" has already different meaning, I think. https://en.wikipedia.org/wiki/Input_method
> 
> It might be better to use another term. (Not sure what is the best term, though. "key event handling"?)

Yes, "key event handling" would fit of course, with perhaps a little disadvantage of being too unspecific...

How about - my personal favorite: "keyboard layout translation strategy" (strategy in a sense of design pattern "strategy") // shortened "kbdlayout_trans_strategy"

Or perhaps: "keycode translation strategy" // shortened "keycode_trans_strategy"

Yet another variant: "keyboard layout handling" // shortened "kbdlayout_handling"

Let me know your preference to update PR accordingly.

k-takata:
I checked the document of `TranslateMessage`:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-translatemessage
> Translates virtual-key messages into character messages.

I think "keycode translation strategy" is good.

ant0sha:
> I think "keycode translation strategy" is good.

updated accordingly, please check

k-takata:
I'm thinking if using `test_mswin_event()` is okay.
This cannot be used on the tiny builds.

ant0sha:
> I'm thinking if using test_mswin_event() is okay.
This cannot be used on the tiny builds.

Hm... Do we have such biest "gvim tiny version" at all ? Or it is perhaps issue with makefile for Mingw. I am using cross compiler with MINGW from linux - Make_cyg_ming.mak (with CROSS=yes), set "FEATURES=TINY" there, instead of "FEATURES=HUGE" and get bunch of errors when I try to build tiny gvim for windows, linking step fails [1].

Are there perhaps some good instructions how to build it otherwise? I quickly checked offline version of Tony's page but it is only talking about how to build unix flavors, nothing about cross-compile.

(Well I can try later to build on another machine with cygwin mingw perhaps with more luck but... Will appreciate short advise to save time)

[1] failure from linker on opensuse with cross-mingw for TINY version
```
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/alloc.o:alloc.c:(.rdata$.refptr.alloc_fail_id[.refptr.alloc_fail_id]+0x0): undefine
d reference to `alloc_fail_id'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/alloc.o:alloc.c:(.rdata$.refptr.alloc_fail_repeat[.refptr.alloc_fail_repeat]+0x0): 
undefined reference to `alloc_fail_repeat'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/alloc.o:alloc.c:(.rdata$.refptr.alloc_fail_countdown[.refptr.alloc_fail_countdown]+
0x0): undefined reference to `alloc_fail_countdown'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arabic.o:arabic.c:(.rdata$.refptr.p_tbidi[.refptr.p_tbidi]+0x0): undefined referenc
e to `p_tbidi'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arabic.o:arabic.c:(.rdata$.refptr.p_arshape[.refptr.p_arshape]+0x0): undefined refe
rence to `p_arshape'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x1086): undefined reference to `error_if_popup_window'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x1333): undefined reference to `error_if_popup_window'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x1619): undefined reference to `error_if_popup_window'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x1909): undefined reference to `error_if_popup_window'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x3454): undefined reference to `find_win_by_nr_or_id'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x34a6): undefined reference to `check_for_opt_number_ar
g'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x34bc): undefined reference to `tv_get_number'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x352b): undefined reference to `check_for_opt_number_arg'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x354c): undefined reference to `find_tabwin'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x3579): undefined reference to `check_for_opt_number_arg'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x35cd): undefined reference to `find_win_by_nr_or_id'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x35fb): undefined reference to `tv_get_number_chk'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x3665): undefined reference to `tv_get_number_chk'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x3674): undefined reference to `rettv_list_alloc'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x36ab): undefined reference to `list_append_string'
/usr/lib64/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld: gobjx86-64/arglist.o:arglist.c:(.text+0x36d6): undefined reference to `check_for_opt_number_ar
...
```

chrisbra:
```
call test_mswin_event('set_w32_ime', {'ime': 'experimental'})
call test_mswin_event('set_w32_ime', {'ime': 'classic'})
```

Is this just for now or are we going with some new option value?

ant0sha:
> ```
> call test_mswin_event('set_w32_ime', {'ime': 'experimental'})
> call test_mswin_event('set_w32_ime', {'ime': 'classic'})
> ```
> 
> Is this just for now or are we going with some new option value?

Hi @chrisbra , right now the plan is like this:
```
	:call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'experimental'})
	:call test_mswin_event('set_keycode_trans_strategy', {'strategy': 'classic'})
```

ant0sha:
Please ignore my comment about problems with TINY GVIM build (removing all *.o files manually solved my issue with it)

Now the question is really, how to change the keycode_trans_strategy in TINY w32 GVIM. To introduce new normal option for that sounds somehow silly... In the beginning I was thinking we will check some environment variable once and do lazy initialize the strategy if it is not yet initialized - but that way we can only have one state per running instance, which cannot be changed on the fly, which quite restrict the possibilities for experimenting. Are there some similar "hidden developer back doors" existing, beside to let user to call some special function?

ant0sha:
In the meantime I will implement fallback check first time, if environment variable, let us say VIM_KEYCODE_TRANS_STRATEGY is defined, and if yes, it value (classic|experimental) will be used to initialize the strategy first time. So no dynamic change on the fly will be possible in TINY gvim.

chrisbra:
I have been wondering: 
- could it happen that users need to use mappings from one keycode translation and different mappings from other keycode translations?

ant0sha:
> I have been wondering:
> 
>     * could it happen that users need to use mappings from one keycode translation and different mappings from other keycode translations?

Kind of, especially on non-English keyboard layouts - with 'experimental' strategy one can start to map things, which are usually "overriden" by ```TranlsateMessage()``` in 'classic' strategy.

Examples

* consider German keyboard layout and mapping ```imap <C-ü> something```
  * With 'classic' strategy it  is void, because ```<C-ü>``` never reaches ```_OnChar()```, but Windows API ```TranslateMessage()``` silently converts it to <ESC> (\027) instead.
  * With 'experimental' strategy it works - but on the other hand, users seems to prefer to have <ESC> here emitted always
 
Also what puzzles me yet: it do not works to ```map <C-ü> <ESC>``` on German keyboard layout to mimic 'classic' behavior, it only works with ```imap ...``` - not investigated why, since the point now is not to "fix" the 'experimental' strategy, but come back to the "lost" 'classic' strategy with least possible effort, at the same time keeping the door open if there are some users out there who get hooked on new 'experimental' strategy - and made available through it new mappings.

Another example:

* consider Cyrillic keyboard layout and mapping ```imap <C-ы> something```
  * With 'classic' strategy it  is void, because ```<C-ы>``` never reaches ```_OnChar()```, but Windows API ```TranslateMessage()``` silently converts it to <C-S>.
  * With 'experimental' strategy it works - but on the other hand, users seems to prefer to have <C-S> here emitted always

* consider Cyrillic keyboard layout and mapping ```imap <C-с> something``` (russian "es" letter)
  * With 'classic': <C-C> (normal english "C")
  * With 'experimental' - "something" is inserted but users seems to prefer <C-C> (english "C") here

* consider Cyrillic keyboard layout and mapping ```imap <C-м> something``` (russian "em" letter)
  * With 'classic': <C-V> (english "V")
  * With 'experimental' - "something" is inserted but users seems to prefer <C-V> (english "V") here

Basically for Cyrillic layout seems that most russian letters with CTRL producing with 'classic' their underlying latin-key control codes (but strangely not <ESC> like on German! - well this is a question to Microsoft and their TranslateMessage() implementation) and that is what users prefer to have restored back. However after more then one year of 'experimental' strategy being in the wild, we are never sure we break someone configuration by changing back to 'classic' - without supporting some migration possibility. Also I am not sure how 'classic' works with completely and completely non standard keyboard layouts like 'bepo' (https://en.wikipedia.org/wiki/B%C3%89PO) and if there more advantages on such layouts we have from 'experimental'.

ant0sha:
This is assigned to "backlog" milestone right now.

Should fix:
#12595
#10615

chrisbra:
can we use the 'keyprotocol' option here?

PR #12754, opened 2023-08-10
runtime-files: Update Headers

chrisbra:
After merging in commits 81b8bf5b4a33 and 6a500661a9cb, updating the header was forgotten.  So do it now.  While at it, also change the maintainer for the 'defaults.vim' runtime file from Bram to the Vim project.

### DIFF
--- a/runtime/doc/tips.txt
+++ b/runtime/doc/tips.txt
@@ -1,4 +1,4 @@
-*tips.txt*      For Vim version 9.0.  Last change: 2021 Nov 06
+*tips.txt*      For Vim version 9.0.  Last change: 2023 Aug 10
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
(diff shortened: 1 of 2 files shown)


runtime/defaults.vim
```diff
@@ -1,7 +1,8 @@
 " The default vimrc file.
 "
-" Maintainer:	Bram Moolenaar <Bram@vim.org>
-" Last change:	2023 May 10
+" Maintainer:	The Vim Project https://github.com/vim/vim
```
k-takata:
```suggestion
" Maintainer:	The Vim Project <https://github.com/vim/vim>
```
Is it better to add `< >`?

chrisbra: yes better. But let me remove defaults.vim here. It is handled as part of https://github.com/vim/vim/pull/12749

k-takata (approved):

chrisbra:
closed in favor of #12790

PR #12575, opened 2023-06-22
if_perl.xs: fix compatibility for Perl 5.36 and up

pheiduck:
Test Env:
Fedora 39 + Perl 5.37.11 and Perl 5.38.0-RC1
`./configure --enable-gui=no --with-features=Huge --enable-perlinterp`

Fixes: https://github.com/vim/vim/issues/12543
Closes: https://github.com/Perl/perl5/issues/21164

PR #12755, opened 2023-08-10
Fix compilation error with Perl 5.32

k-takata:
https://github.com/vim/vim/pull/12575 caused compilation error with Perl 5.32.

Strawberry Perl is still 5.32. Windows users may stuck with this version.

### DIFF
--- a/src/if_perl.xs
+++ b/src/if_perl.xs
@@ -40,7 +40,7 @@
 /* Work around for perl-5.18.
  * Don't include "perl\lib\CORE\inline.h" for now,
  * include it after Perl_sv_free2 is defined. */
-#if (PERL_REVISION == 5) && (PERL_VERSION >= 18)
+#ifdef DYNAMIC_PERL
 # define PERL_NO_INLINE_FUNCTIONS
 #endif
 
@@ -709,8 +709,142 @@ S_POPMARK(pTHX)
 #  define Perl_POPMARK S_POPMARK
 # endif
 
+# if (PERL_REVISION == 5) && (PERL_VERSION >= 32)
+PERL_STATIC_INLINE U8
+Perl_gimme_V(pTHX)
+{
+    I32 cxix;
+    U8  gimme = (PL_op->op_flags & OPf_WANT);
+
+    if (gimme)
+        return gimme;
+    cxix = PL_curstackinfo->si_cxsubix;
+    if (cxix < 0)
+	return
+#  if (PERL_REVISION == 5) && (PERL_VERSION >= 34)
+	    PL_curstackinfo->si_type == PERLSI_SORT ? G_SCALAR:
+#  endif
+	    G_VOID;
+    assert(cxstack[cxix].blk_gimme & G_WANT);
+    return (cxstack[cxix].blk_gimme & G_WANT);
+}
+# endif
+
+# if (PERL_REVISION == 5) && (PERL_VERSION >= 38)
+#  define PERL_ARGS_ASSERT_SVPVXTRUE             \
+        assert(sv)
+PERL_STATIC_INLINE bool
+Perl_SvPVXtrue(pTHX_ SV *sv)
+{
+    PERL_ARGS_ASSERT_SVPVXTRUE;
+
+    if (! (XPV *) SvANY(sv)) {
+        return false;
+    }
+
+    if ( ((XPV *) SvANY(sv))->xpv_cur > 1) { /* length > 1 */
+        return true;
+    }
+
+    if (( (XPV *) SvANY(sv))->xpv_cur == 0) {
+        return false;
+    }
+
+    return *sv->sv_u.svu_pv != '0';
+}
+
+#  define PERL_ARGS_ASSERT_SVGETMAGIC            \
+        assert(sv)
+PERL_STATIC_INLINE void
+Perl_SvGETMAGIC(pTHX_ SV *sv)
+{
+    PERL_ARGS_ASSERT_SVGETMAGIC;
+
+    if (UNLIKELY(SvGMAGICAL(sv))) {
+        mg_get(sv);
+    }
+}
+
+PERL_STATIC_INLINE char *
+Perl_SvPV_helper(pTHX_
+                 SV * const sv,
+                 STRLEN * const lp,
+                 const U32 flags,
+                 const PL_SvPVtype type,
+                 char * (*non_trivial)(pTHX_ SV *, STRLEN * const, const U32),
+                 const bool or_null,
+                 const U32 return_flags
+                )
+{
+    /* 'type' should be known at compile time, so this is reduced to a single
+     * conditional at runtime */
+    if (   (type == SvPVbyte_type_      && SvPOK_byte_nog(sv))
+        || (type == SvPVforce_type_     && SvPOK_pure_nogthink(sv))
+        || (type == SvPVutf8_type_      && SvPOK_utf8_nog(sv))
+        || (type == SvPVnormal_type_    && SvPOK_nog(sv))
+        || (type == SvPVutf8_pure_type_ && SvPOK_utf8_pure_nogthink(sv))
+        || (type == SvPVbyte_pure_type_ && SvPOK_byte_pure_nogthink(sv))
+   ) {
+        if (lp) {
+            *lp = SvCUR(sv);
+        }
+
+        /* Similarly 'return_flags is known at compile time, so this becomes
+         * branchless */
+        if (return_flags & SV_MUTABLE_RETURN) {
+            return SvPVX_mutable(sv);
+        }
+        else if(return_flags & SV_CONST_RETURN) {
+            return (char *) SvPVX_const(sv);
+        }
+        else {
+            return SvPVX(sv);
+        }
+    }
+
+    if (or_null) {  /* This is also known at compile time */
+        if (flags & SV_GMAGIC) {    /* As is this */
+            SvGETMAGIC(sv);
+        }
+
+        if (! SvOK(sv)) {
+            if (lp) {   /* As is this */
+                *lp = 0;
+            }
+
+            return NULL;
+        }
+    }
+
+    /* Can't trivially handle this, call the function */
+    return non_trivial(aTHX_ sv, lp, (flags|return_flags));
+}
+
+#  define PERL_ARGS_ASSERT_SVNV                  \
+        assert(sv)
+PERL_STATIC_INLINE NV
+Perl_SvNV(pTHX_ SV *sv) {
+    PERL_ARGS_ASSERT_SVNV;
+
+    if (SvNOK_nog(sv))
+        return SvNVX(sv);
+    return sv_2nv(sv);
+}
+
+#  define PERL_ARGS_ASSERT_SVIV                  \
+        assert(sv)
+PERL_STATIC_INLINE IV
+Perl_SvIV(pTHX_ SV *sv) {
+    PERL_ARGS_ASSERT_SVIV;
+
+    if (SvIOK_nog(sv))
+        return SvIVX(sv);
+    return sv_2iv(sv);
+}
+# endif
+
 /* perl-5.34 needs Perl_SvTRUE_common; used in SvTRUE_nomg_NN */
-# if (PERL_REVISION == 5) && (PERL_VERSION == 34)
+# if (PERL_REVISION == 5) && (PERL_VERSION >= 34)
 PERL_STATIC_INLINE bool
 Perl_SvTRUE_common(pTHX_ SV * sv, const bool sv_2bool_is_fallback)
 {
@@ -737,7 +871,7 @@ Perl_SvTRUE_common(pTHX_ SV * sv, const bool sv_2bool_is_fallback)
 # endif
 
 /* perl-5.32 needs Perl_SvTRUE */
-# if (PERL_REVISION == 5) && (PERL_VERSION == 32)
+# if (PERL_REVISION == 5) && (PERL_VERSION >= 32)
 PERL_STATIC_INLINE bool
 Perl_SvTRUE(pTHX_ SV *sv) {
     if (!LIKELY(sv))


k-takata:
@pheiduck Could you check if this works in your environment?

src/if_perl.xs
```diff
@@ -40,7 +40,7 @@
 /* Work around for perl-5.18.
  * Don't include "perl\lib\CORE\inline.h" for now,
  * include it after Perl_sv_free2 is defined. */
-#if (PERL_REVISION == 5) && (PERL_VERSION >= 18)
+#ifdef DYNAMIC_PERL
```
k-takata, line 43: `PERL_REVISION` and `PERL_VERSION` cannot be used here, because they are defined later.

src/if_perl.xs
```diff
@@ -334,6 +334,7 @@ typedef int perl_key;
 # if (PERL_REVISION == 5) && (PERL_VERSION >= 24)
 #  define Perl_savetmps dll_Perl_savetmps
 # endif
+# define Perl_gimme_V dll_Perl_gimme_V
```
k-takata: I'm not sure if we need a version check here.

pheiduck:
@k-takata It's working anyway no issues
Test Env:
Fedora 39 + Perl 5.39.2
`./configure --enable-gui=no --with-features=Huge --enable-perlinterp`

k-takata:
It seems that my change was not enough. I will check later.

k-takata:
I've updated the PR.
I've tested it with Perl 5.32, 5.36, and 5.38 on Windows with /dyn.
Could anyone try this on other environments?

pheiduck:
Works on my env:
Fedora 39 + Perl 5.39.2
./configure --enable-gui=no --with-features=Huge --enable-perlinterp

chrisbra:
verified on Ubuntu using Perl 5.32 (static and dynamic). 

But with Perl 5.38 dynamic, I see this:

```
gcc -c -I. -Iproto -DHAVE_CONFIG_H     -g -O2  -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1        version.c -o objects/version.o
link.sh: $LINK_AS_NEEDED set to 'yes': invoking linker directly.
  gcc   -Wl,-E   -L/usr/local/lib -Wl,--as-needed       -o vim objects/alloc.o objects/arabic.o objects/arglist.o objects/autocmd.o objects/beval.o objects/buffer.o objects/change.o objects/blob.o objects/blowfish.o objects/cindent.o objects/clientserver.o objects/clipboard.o objects/cmdexpand.o objects/cmdhist.o objects/crypt.o objects/crypt_zip.o objects/debugger.o objects/dict.o objects/diff.o objects/digraph.o objects/drawline.o objects/drawscreen.o objects/edit.o objects/eval.o objects/evalbuffer.o objects/evalfunc.o objects/evalvars.o objects/evalwindow.o objects/ex_cmds.o objects/ex_cmds2.o objects/ex_docmd.o objects/ex_eval.o objects/ex_getln.o objects/fileio.o objects/filepath.o objects/findfile.o objects/float.o objects/fold.o objects/getchar.o objects/gui_xim.o objects/hardcopy.o objects/hashtab.o objects/help.o objects/highlight.o objects/if_cscope.o objects/if_xcmdsrv.o objects/indent.o objects/insexpand.o objects/list.o objects/locale.o objects/logfile.o objects/map.o objects/mark.o objects/match.o objects/mbyte.o objects/memline.o objects/menu.o objects/misc1.o objects/misc2.o objects/mouse.o objects/move.o objects/normal.o objects/ops.o objects/option.o objects/optionstr.o objects/os_unix.o objects/pathdef.o objects/popupmenu.o objects/popupwin.o objects/profiler.o objects/pty.o objects/quickfix.o objects/regexp.o objects/register.o objects/screen.o objects/scriptfile.o objects/search.o objects/session.o objects/sha256.o objects/sign.o objects/sound.o objects/spell.o objects/spellfile.o objects/spellsuggest.o objects/strings.o objects/syntax.o objects/tag.o objects/term.o objects/terminal.o objects/testing.o objects/textformat.o objects/textobject.o objects/textprop.o objects/time.o objects/typval.o objects/ui.o objects/undo.o objects/usercmd.o objects/userfunc.o objects/version.o objects/vim9class.o objects/vim9cmds.o objects/vim9compile.o objects/vim9execute.o objects/vim9expr.o objects/vim9instr.o objects/vim9script.o objects/vim9type.o objects/viminfo.o objects/window.o objects/bufwrite.o  objects/vterm_encoding.o objects/vterm_keyboard.o objects/vterm_mouse.o objects/vterm_parser.o objects/vterm_pen.o objects/vterm_screen.o objects/vterm_state.o objects/vterm_unicode.o objects/vterm_vterm.o   objects/if_perl.o objects/if_perlsfio.o      objects/netbeans.o objects/job.o objects/channel.o objects/xdiffi.o objects/xemit.o objects/xprepare.o objects/xutils.o objects/xhistogram.o objects/xpatience.o  objects/charset.o objects/json.o objects/main.o objects/memfile.o objects/message.o    -lSM -lICE -lXpm -lXt -lX11 -lXdmcp -lSM -lICE  -lm -ltinfo  -lselinux -lcanberra -lsodium -lrt -lacl -lattr -lgpm -ldl   -Wl,-E  -fstack-protector-strong -L/usr/local/lib  -L/home/chrisbra/local/stow/perl53711/lib/perl5/5.36.1/x86_64-linux/CORE -lperl -lpthread -ldl -lm -lcrypt -lutil -lc
/usr/bin/ld: objects/if_perl.o: in function `PerlIOVim_pushed':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1624: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Count':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2363: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Number':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2332: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Name':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2297: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_DESTROY':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2268: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Append':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2589: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o:/home/chrisbra/code/vim-upstream/src/if_perl.xs:2097: more undefined references to `Perl_SvIV' follow
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Append':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:2100: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Delete':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2506: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Delete':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:2041: undefined reference to `Perl_SvIV'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:2042: undefined reference to `Perl_SvIV'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:2036: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Set':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2444: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o:/home/chrisbra/code/vim-upstream/src/if_perl.xs:1998: more undefined references to `Perl_SvIV' follow
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Set':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:2001: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Get':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2399: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIBUF_Get':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1975: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIWIN_Cursor':
/home/chrisbra/code/vim-upstream/src/if_perl.c:2220: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIWIN_Cursor':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1912: undefined reference to `Perl_SvIV'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1913: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o:/home/chrisbra/code/vim-upstream/src/if_perl.c:2148: more undefined references to `Perl_SvIV' follow
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_Blob':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1765: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_Eval':
/home/chrisbra/code/vim-upstream/src/if_perl.c:1944: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_DoCommand':
/home/chrisbra/code/vim-upstream/src/if_perl.c:1923: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_SetOption':
/home/chrisbra/code/vim-upstream/src/if_perl.c:1901: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_Buffers':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1806: undefined reference to `Perl_SvIV'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1812: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `perl_to_vim':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1344: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1457: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1333: undefined reference to `Perl_SvNV'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1337: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_Windows':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1849: undefined reference to `Perl_SvIV'
/usr/bin/ld: objects/if_perl.o: in function `XS_VIM_Msg':
/home/chrisbra/code/vim-upstream/src/if_perl.c:1863: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.c:1874: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:879: undefined reference to `Perl_SvGETMAGIC'
/usr/bin/ld: objects/if_perl.o: in function `ex_perl':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1242: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE_common':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:858: undefined reference to `Perl_SvPVXtrue'
/usr/bin/ld: objects/if_perl.o: in function `do_perleval':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1523: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:879: undefined reference to `Perl_SvGETMAGIC'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE_common':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:858: undefined reference to `Perl_SvPVXtrue'
/usr/bin/ld: objects/if_perl.o: in function `ex_perldo':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1567: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: /home/chrisbra/code/vim-upstream/src/if_perl.xs:1584: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:879: undefined reference to `Perl_SvGETMAGIC'
/usr/bin/ld: objects/if_perl.o: in function `replace_line':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:1262: undefined reference to `Perl_SvPV_helper'
/usr/bin/ld: objects/if_perl.o: in function `Perl_SvTRUE_common':
/home/chrisbra/code/vim-upstream/src/if_perl.xs:858: undefined reference to `Perl_SvPVXtrue'
collect2: error: ld returned 1 exit status
link.sh: Linking failed
make: *** [Makefile:2062: vim] Error 1
```

compiling without `dynamic` works however.

pheiduck:
@chrisbra I can not reproduce your issue on **Ubuntu + Perl 5.38.0**

ant0sha:
@k-takata tried this PR on:

- Linux/OpenSuse Tumbleweed:
  - perl 5.36, both are fine - dynamic, static build
  - perl 5.32, both are fine - dynamic, static build
- Win10_Cygwin:
  - perl 5.32, both are fine - dynamic, static build

chrisbra:
okay, looks good now. I guess I had used not the released 5.38 version. I tried again with a self-compiled Perl 5.38 and can verify it works now.

ychin:
@k-takata / @pheiduck this seems to be causing some issues with Perl 5.30 when using dynamic libs. MacVim is using this to build right now (Using Perl 5.30 from macOS itself, and using dynamic linking), and I'm noticing that this change caused Vim to now statically depends on the library instead of purely dynamic.

This is the job that failed (https://github.com/ychin/macvim/actions/runs/5933905278/job/16090004063). It essentially calls `otool -L ${VIM}` and I can see that it's now depending on the file "/System/Library/Perl/5.30/darwin-thread-multi-2level/CORE/libperl.dylib" which it didn't use to before.

This doesn't cause a compilation error because even when we are using dynamic compilation, `auto/config.mk` still uses `-lperl` in the `PERL_LIBS` flag, which I think is wrong. When configuring using `--enable-perlinterp=dynamic`, that flag should be removed to prevent accidental linkage like this.

When I removed that flag, this is the error:

```
Undefined symbols for architecture arm64:
  "_Perl_block_gimme", referenced from:
      _XS_VIM_Buffers in if_perl.o
      _XS_VIM_Windows in if_perl.o
```

I'm not 100% sure what the last two commits (this PR #12755, and #12575), but I'm guessing the new `GIMME_V` is pulling in new dependencies or something? Should we add this function to `perl_funcname_table` and load it in dynamically? I'm not familiar enough with the Perl interface to understand this.

ychin:
Ok, looking at it more now I understand what's going on.

Seems like the main culprit is that we moved from using the old `GIMME` macro which is technically deprecated ([docs](https://perldoc.perl.org/perlapi#GIMME)) and we used the newer `GIMME_V` change instead. Was there a reason why we needed to use the new macro other than it's the newer one @pheiduck? I guess my concern is that `GIMME` is entirely implemented in the [Perl header file](https://github.com/Perl/perl5/blob/blead/op.h), but `GIMME_V` depends on other functionality defined in [`inline.h`](https://github.com/Perl/perl5/blob/blead/inline.h) which we are not including since we set `PERL_NO_INLINE_FUNCTIONS`, and that means we have to re-implement the functionality (which was done in this PR), which is quite fragile and version dependent. Can we just revert back to using `GIMME` if all we wanted to was to fix the Perl interface to compile on newer versions? It is more robust to future version changes, and seems like we should get the smallest change in possible before we make more radical changes (Perl interface is rarely used anyway, so I think anything we do to minimize maintenance cost is best).

Looking at this part of the code in general, I'm concerned at how we essentially need to re-implement the code from Perl's internal `inline.h` and duplicate it here. If a new Perl version updates its implementation we would need to update this and any bugs would be subtle as we wouldn't know it deviated. I guess we have been doing it for a while but is there a way we could avoid re-implementing these functions like `Perl_gimme_V` (e.g. just use the old `GIMME` macro instead of `GIMME_V`)? Alternatively, we could just not set `PERL_NO_INLINE_FUNCTIONS` and let Perl includes the inline header. I think it was previously used because `inline.h` contains other functions that may cause us to link against the library statically, and the exact behavior depends on the compiler and how it handles unused functions in the header, so it may not work on some compilers. It just seems like a somewhat fragile system to me but maybe that's just how Perl is.

Either way my vote is to just use GIMME instead of GIMME_V for now unless there was a reason for this change.

pheiduck:
@ychin The reason was this:
It is planned to remove GIMME from a future release of Perl. Do not use it for new code; remove it from existing code.
But maybe you are right and we have to postpone the implementation of it. Since the code has changed quite a bit I am unsure myself if I can reverse engineer this.

ychin:
Ah ok. Is there documentation for this? Just curious to read more. I wonder if we should have an ifdef to just use GIMME in older Perl versions to avoid having to re-implement the inline functions from the header due to the issues I mentioned earlier.

ychin:
Ok, I have filed PR #12914 to try to fix this. See that PR for more info.

PR #12759, opened 2023-08-10
Fix "All Files" browsefilter patterns to include files with no extension

### LINKED ISSUE #12685: File open can't show Makefiles
### Steps to reproduce

in gvim 9.0.1189, in a directory with a Makefile (or any file with no extension, select menu item File -> Open to get the system File Open dialog box.  Note that All Files is shown as (*).  Navigate to a folder where there are, say, C files.  With All Files still selected, note that the C files do not appear.  Change the file type selector to, say, C Source (*.c,*.h).  Now C source files are visible, and Makefiles are not.  This is correct behavior.

Note that in the list of file types to show, All Files is given as (*.*).  This means that files with no extension, such as Makefile, are not shown.  

All Files should always be specified as (*, *.*)

### Expected behaviour

Selecting All Files should make all files visible, including those with no extension

### Version of Vim

gvim 9.0.1189

### Environment

Ubuntu 23.04

### Logs and stack traces

_No response_

dkearns:
Fix "All Files" browsefilter patterns to include files with no extension
    
Also normalises the browsefilter conditional test to check for both Win32 and GTK and an unset b:browsefilter.

Fixes #12685.

This applies to all ftplugins not only those I maintain.

### DIFF
--- a/runtime/doc/editing.txt
+++ b/runtime/doc/editing.txt
@@ -1329,8 +1329,9 @@ b:browsefilter variable.  You would most likely set b:browsefilter in a
 filetype plugin, so that the browse dialog would contain entries related to
 the type of file you are currently editing.  Disadvantage: This makes it
 difficult to start editing a file of a different type.  To overcome this, you
-may want to add "All Files\t*.*\n" as the final filter, so that the user can
-still access any desired file.
+may want to add "All Files (*.*)\t*\n" as the final filter on Windows or "All
+Files (*)\t*\n" on other platforms, so that the user can still access any
+desired file.
 
 To avoid setting browsefilter when Vim does not actually support it, you can
 use has("browsefilter"): >
--- a/runtime/ftplugin/aap.vim
+++ b/runtime/ftplugin/aap.vim
@@ -1,7 +1,7 @@
 " Vim filetype plugin file
 " Language:	Aap recipe
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
-" Last Change:	2023 Aug 10
+" Last Change:	2024 Jan 14
 " Former Maintainer:	Bram Moolenaar <Bram@vim.org>
 
 " Only do this when not done yet for this buffer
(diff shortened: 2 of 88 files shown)


chrisbra:
why was this put in draft mode? I think we may need as well to adjust the Last Change header.

dkearns:
I was bit short on time and hadn't finished it.  I was going to update the help with an example as well as update the headers.

I also wanted to discuss it first.

One of the things I've been doing over the last couple of years is attempting to normalise the user-maintained runtime files.  Bram, due to his workflow, was extremely reluctant to perform bulk updates as reversions easily slipped through when he received updates and he requested that I contact the maintainers individually with patches.

Chasing down 50 - 100 maintainers for each of these changes has had mixed results. Maybe as many as 20% of maintainers are non-responsive and some, particularly those with external repositories, accept the patches but fail to send updates for Vim to distribute.

I take it from your comment that you're open to these sort of bulk updates?

I'll finish this off tomorrow.

chrisbra:
> I take it from your comment that you're open to these sort of bulk updates?

bulk updates used to be extremely annoying due to the way how runtime files are maintained. So perhaps it's time to think about handling those here centrally and asking the runtime file maintainers to re-sync the files with what Vim distributes before sending updates. I think that should work.

dkearns:
I added a new entry to the "Last Change" headers, "YYYY MMM DD by Vim Project".  This is how it was generally done in the past only it used to list the individual contributor.  I think this will be easier for maintainers to understand and hopefully notice.

dkearns:
Marking as Draft for the moment as there appears to be some intentional difference between the default "All Files" filter pattern on Windows (possibly just DOS) and Unix but this was not propagated to the ftplugins.

https://github.com/vim/vim/blob/20b33b56ad5d92c1c11e0859dc9333166625e0ad/src/vim.h#L2391C1-L2414C7

Perhaps we should also use "All Files (*)" for the label like the default value for consistency.

chrisbra:
Hm,
the `*.*` seems to be fine for Windows and even shows  files without extension. And I think the  `*` pattern also  worked there. So not sure, we need to make a difference between Windows and Unix.

dkearns:
So as I suspected, or perhaps vaguely recalled, the `*.*` filter pattern is a legacy of the wildcard matching algorithm for 8.3 filenames from the days of DOS and CP/M where it actually matched all files.

We don't support DOS anymore so I think the only reason to keep it for Windows would be for aesthetic reasons as it's still the 'standard' filter label there.  We could provide a `SetBrowseFilter()` function for the ftplugins to use that would set the appropriate "All Files (...)" label for the current platform but I'm not sure it's worth the trouble.

I'd probably vote for using the current not-Windows default label of "All Files (*)" rather than "All Files".  This is also used on Windows in the wild.

k-takata:
The default open dialog box on Windows 11 still uses "All Files (*.*)", so this looks natural for Windows users.

dkearns:
This takes the simple boilerplate approach to continue using the quirky `*.*` filter label on Windows which seems to be most in keeping with the current style.

I can't think of any reasonable way to push this into 'core' that wouldn't cause backward compatibility issues.

chrisbra:
can you please re-base?

chrisbra:
I can imagine that this was quite some work. Thanks for that!

PR #12761, opened 2023-08-10
Add the `WinNewPre` event

### LINKED ISSUE #10635: Introduce `WinNewPre`
It is impossible to do any preparation work before a new window is created.

When a new window is created, it steals its width or height from other windows and it may also scroll other windows. I would like to preserve old windows as much as possible, so e.g. to preserve the sizes I would instead resize the GUI window. To achieve that I need to record the necessary windows info (width, height, topline etc), and to have the info up-to-date this has to be done just before `WinNew`.

Introduce new `WinNewPre` autocommand, that is triggered before creating a new window.

Using `SafeState` and `SafeStateAgain` to record the windows info, but that is very expensive and affects VIM performance.

noscript:
Problem: No event is triggered before creating a new window.
Solution: Add the WinNewPre event. (closes #10635)


src/window.c
```diff
@@ -955,6 +955,8 @@ win_split_ins(
     // Do not redraw here, curwin->w_buffer may be invalid.
     ++RedrawingDisabled;
 
+    apply_autocmds(EVENT_WINNEWPRE, NULL, NULL, FALSE, NULL);
```
yegappan: What will happen if the WINNEWPRE autocmd handler creates a new window or modifies the window layout?
Is the window layout locked here?

chrisbra: or when it closes other windows?

noscript: >Is the window layout locked here?

I think locking the layout may limit the use of `WinNewPre`. I pushed the change, so the event is non-recursive now.

> What will happen if the WINNEWPRE autocmd handler creates a new window or modifies the window layout?

> or when it closes other windows?

Modifying layout should be possible and allowed. I added two new tests.

chrisbra: > Modifying layout should be possible and allowed. I added two new tests.

We had in the past a lot of crashes, because the curwin or curbuf pointer became invalid, after the autocommands did something unexpected.

noscript: I made a change to lock the layout during the event.

noscript:
Fixed merge conflicts.

noscript:
I published a plugin that demonstrates the use case https://github.com/noscript/ultrawide.vim

The plugin relies on `xwininfo` tool and thus only supports X11. Adding other platforms should be trivial.

chrisbra:
so you basically just remember the current state and compare it after the new window has been created to decide what to do.

noscript:
That's correct.

Issue #12758, opened 2023-08-10
Building from source with commit 1d7caa58e fails on macOS 13.3.1 (a) and Perl 5.30


al3xandru:
### Steps to reproduce

1. checkout commit 1d7caa58e (tag: v9.0.1681) patch 9.0.1681: Build Failure with Perl 5.38 
2. `make clean`
3. `make install`

### Expected behaviour

Builds successfully

### Version of Vim

1d7caa58e

### Environment

macOS 13.3.1 (a)

### Logs and stack traces

```shell
gcc -c -I. -Ilibvterm/include -Iproto -DHAVE_CONFIG_H   -DMACOS_X -DMACOS_X_DARWIN  -g -O2 -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1        -DINLINE="" -DVSNPRINTF=vim_vsnprintf -DSNPRINTF=vim_snprintf -DIS_COMBINING_FUNCTION=utf_iscomposing_uint -DWCWIDTH_FUNCTION=utf_uint2cells -o objects/vterm_vterm.o libvterm/src/vterm.c
/usr/bin/perl -e 'unless ( $] >= 5.005 ) { for (qw(na defgv errgv)) { print "#define PL_$_ $_\n" }}' > auto/if_perl.c
/usr/bin/perl /System/Library/Perl/5.30/ExtUtils/xsubpp -prototypes -typemap \
            /System/Library/Perl/5.30/ExtUtils/typemap if_perl.xs >> auto/if_perl.c
gcc -c -I. -DDYNAMIC_PERL_DLL=\"libperl.dylib\"   -g -DPERL_USE_SAFE_PUTENV  -iwithsysroot /System/Library/Perl/5.30/darwin-thread-multi-2level/CORE  -Iproto -DHAVE_CONFIG_H   -DMACOS_X -DMACOS_X_DARWIN  -g -O2 -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1         -o objects/if_perl.o auto/if_perl.c
if_perl.xs:666:1: error: redefinition of 'S_SvREFCNT_dec'
S_SvREFCNT_dec(pTHX_ SV *sv)
^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Perl/5.30/darwin-thread-multi-2level/CORE/inline.h:209:1: note: previous definition is here
[...]

PR #12766, opened 2023-08-11
Fix perl compilation errors

chrisbra:
Commit 1d7caa58e3c87f75a4becbceabbd1af181ace11e slightly changed how to integrate with Perl 5.37. Unfortunately, it seems it unexpectedly broke integration with older Perls.

So parly revert part of the patch:

1) GIMME was changed to GIMME_V for Perl5.37. Let's instead keep the GIMME
   macro and define it to GIMME_V for Perl.37 instead.

2) Change the line
   #if (PERL_REVISION == 5) && (PERL_VERSION >= 18)
back to
  #ifdef DYNAMIC_PERL

I did verify that this comiles using Perl 5.32 and Perl 5.37 in dynamic mode, but for some reason loading the perl.a library fails, so not hundert percent sure this is correct.

related #12758

### DIFF
--- a/src/if_perl.xs
+++ b/src/if_perl.xs
@@ -40,7 +40,7 @@
 /* Work around for perl-5.18.
  * Don't include "perl\lib\CORE\inline.h" for now,
  * include it after Perl_sv_free2 is defined. */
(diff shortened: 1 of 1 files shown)


chrisbra:
closing in favor of #12755

PR #12770, opened 2023-08-12
Fix mapset() test failure when using GUI

zeertzjq:
Problem:    Test for mapset() fails when using GUI.
Solution:   Feed an unsimplified Ctrl-B.


zeertzjq:
This seems to be related to [patch 8.2.0855](https://github.com/vim/vim/commit/ebe9d34aa07037cff2188a8dd424ee1f59cbb0bf)

PR #12771, opened 2023-08-12
Fix undo test flakiness

zeertzjq:
Problem:    Undo test is flaky.
Solution:   Change time to "1 second ago" in both dumps.


zeertzjq:
Hmm, it can actually be "2 seconds ago" in https://github.com/vim/vim/actions/runs/5838947710/job/15836564531?pr=12625

chrisbra:
I am confused. Where are those `.vim` files used? They are not dumps, and you did not reference it in any tests?

zeertzjq:
They are passed to the `ReadAndFilter()` function in `screendump.vim`.

chrisbra:
ah I see.

PR #12779, opened 2023-08-13
cache linux-modules-extra to reduce failed downloads

pheiduck:
This GitHub action install Linux APT packages and cache them for later use. When executed next time with same package list, and any other environment are not changed, installed files are extracted from the cached archive.

When valid cached archive is not found, all packages are installed by apt-get command. If you want partial install/cache behavior, use this action multiple times.
ref: https://github.com/marketplace/actions/install-and-cache-apt-tools#actions-use-apt-tools

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,23 +180,20 @@ jobs:
           sudo usermod -a -G audio "${USER}"
           sudo bash ci/setup-xvfb.sh
 
-      - name: Check if linux-modules-extra is available
-        id: check-snd-dummy
-        if: (!(contains(matrix.extra, 'unittests') || contains(matrix.extra, 'vimtags')))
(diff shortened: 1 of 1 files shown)


pheiduck:
Package seams not available at all.
Rebase later again when it is.

chrisbra:
Yeah, I think we need to wait until github updates the runner images

pheiduck:
todo:
re-enable:
```
sudo depmod --verbose
sudo modprobe --verbose snd-dummy
```

chrisbra:
is this ready?

pheiduck:
> is this ready?

Yes, but I would apply it when the image runner update is done.https://github.com/actions/runner-images/issues/8080#issuecomment-1680088869

pheiduck:
@chrisbra I guess we can merge it.

PR #12780, opened 2023-08-13
runtime: a few more changes to the runtime files

chrisbra:
as mentioned here: https://github.com/vim/vim/commit/1688938dd5ac78ab67e54299b9d5b93499dba762#r124379828

### DIFF
--- a/runtime/syntax/shared/typescriptcommon.vim
+++ b/runtime/syntax/shared/typescriptcommon.vim
@@ -1,9 +1,9 @@
 " Vim syntax file
 " Language:     TypeScript and TypeScriptReact
-" Maintainer:   Bram Moolenaar, Herrington Darkholme
-" Last Change:	2021 Sep 22
+" Maintainer:   Herrington Darkholme
+" Last Change:	2023 Aug 13
 " Based On:     Herrington Darkholme's yats.vim
-" Changes:      See https:github.com/HerringtonDarkholme/yats.vim
+" Changes:      See https://github.com/HerringtonDarkholme/yats.vim
 " Credits:      See yats.vim on github
 
 if &cpo =~ 'C'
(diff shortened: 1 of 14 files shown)


runtime/syntax/shared/typescriptcommon.vim
```diff
@@ -1,9 +1,9 @@
 " Vim syntax file
 " Language:     TypeScript and TypeScriptReact
-" Maintainer:   Bram Moolenaar, Herrington Darkholme
-" Last Change:	2021 Sep 22
+" Maintainer:   Moolenaar, Herrington Darkholme
```
ZuBB: Only the first name was deleted

chrisbra: yes you are right, thanks.

runtime/tutor/tutor.vim
```diff
@@ -1,7 +1,7 @@
 " Vim tutor support file
-" Author: Eduardo F. Amatria <eferna1@platea.pntic.mec.es>
-" Maintainer: Bram Moolenaar
-" Last Change:	2019 Nov 11
+" Author:	Eduardo F. Amatria <eferna1@platea.pntic.mec.es>
+" Maintainer:	The·Vim·Project·<https://github.com/vim/vim>
```
seandewar, line 3:
```suggestion
" Maintainer:	The Vim Project <https://github.com/vim/vim>
```

runtime/syntax/zimbu.vim
```diff
@@ -1,7 +1,8 @@
 " Vim syntax file
 " Language:	Zimbu
-" Maintainer:	Bram Moolenaar
-" Last Change:	2014 Nov 23
+" Maintainer:	The·Vim·Project·<https://github.com/vim/vim>
```
seandewar, line 3:
```suggestion
" Maintainer:	The Vim Project <https://github.com/vim/vim>
```

seandewar (commented):
I think these are supposed to be spaces

PR #12783, opened 2023-08-13
Create stale.yml

pheiduck:


chrisbra:
I am not a big fan of closing issues, just because they became old.

ii14:
In neovim the stale bot only closes the inactive issues if they were manually labelled with `needs:response` and `needs:repro`

PR #12785, opened 2023-08-14
if_lua: Prevent SEGV/ABRT for Lua functions invoked via Vim callbacks (fixed issues with PR #12782)

jessepav:
`if_lua` employs a cache system using upvalues that fails when a Lua function is invoked as a Vim callback with certain types of parameters (lists, dicts, blobs, buffers, windows).

A MRE that causes a SEGV with PUC Lua 5.1 and an ABRT with Luajit:

```vim
let Lua_funcref = luaeval('function(d) local s = "in Lua callback" end')
call Lua_funcref({'a' : 'b'})
```

This patch uses the Lua registry rather than upvalues for the userdata cache.

### DIFF
--- a/src/if_lua.c
+++ b/src/if_lua.c
@@ -16,6 +16,12 @@
 #include <lualib.h>
 #include <lauxlib.h>
 
+#if __STDC_VERSION__ >= 199901L
+#  define LUAV_INLINE inline
+#else
+#  define LUAV_INLINE
+#endif
+
 // Only do the following when the feature is enabled.  Needed for "make
 // depend".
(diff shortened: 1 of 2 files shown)


chrisbra:
can you please create a corresponding test at [test_lua.vim](https://github.com/vim/vim/blob/master/src/testdir/test_lua.vim) ?

jessepav:
I've added `Test_lua_funcref_with_params()`  in `test_lua.vim`. It turns out the crash was more straightforward to cause than I'd previously thought.

PR #12790, opened 2023-08-14
runtime(helpfiles): collection of updates to the helpfiles

chrisbra:
as mentioned at:
https://github.com/vim/vim/commit/4a8eb6e7a9df10f79bf95301ced012f0d6a13088#r124414618

### DIFF
--- a/runtime/doc/channel.txt
+++ b/runtime/doc/channel.txt
@@ -1,4 +1,4 @@
-*channel.txt*      For Vim version 9.0.  Last change: 2022 Dec 01
+*channel.txt*      For Vim version 9.0.  Last change: 2023 Aug 15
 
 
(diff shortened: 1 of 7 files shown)


Houl:
The new \*local-noglobal\* was meant to replace both \*special-local-window-option\* and \*special-local-buffer-option\*.
So my suggestion included
`:s/local to buffer\s*\zs|special-local-buffer-option|/|local-noglobal|/`
for the buffer options `'buftype'`, `'bufhidden'`, etc.

chrisbra:
Thanks, I'll update it soon. There seems to be further changes required to some helpfiles, so I changed the title a bit and leave it open until enough doc fixes have accumulated to merge it in.

PR #12797, opened 2023-08-14
Update syntax/muttrc.vim to latest mutt

lunasophia:
I'm not sure what the procedure should be for this. Before he passed, Bram said to send the updates directly to him. Unfortunately, I wasn't able to finish them in time, so instead I'm filing a PR. If there's some other, better way to handle this, please let me know! Additionally, I understand and appreciate that project management is quite hard right now, so please take all the time you need to get to this!

The number of actual changes in this PR is smaller than the diff would suggest: I have tried to keep the line wrappings roughly as they were when I started working on the file. This involved rewrapping many lines, which has significantly--yet artificially--inflated the size of the diff. I should also mention here that this is my first time working on a vim syntax file, and indeed working on any kind of file related to vim besides my own personal vimrc, so I would appreciate any feedback people have on this. In particular, if anyone has any suggestions for improving how the file is laid out to improve its git-friendliness, I would love to hear them!

Finally, thank you for even taking a look at this PR.

### DIFF
--- a/runtime/syntax/muttrc.vim
+++ b/runtime/syntax/muttrc.vim
@@ -1,10 +1,10 @@
 " Vim syntax file
 " Language:	Mutt setup files
 " Original:	Preben 'Peppe' Guldberg <peppe-vim@wielders.org>
-" Maintainer:	Kyle Wheeler <kyle-muttrc.vim@memoryhole.net>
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. You did everything right in creating a PR for this. Thanks!

PR #12801, opened 2023-08-15
[CI] Make the linux-modules-extras conditional

zdohnal:
Several runs fail on sound dummy part, check if this works...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,21 +180,23 @@ jobs:
           sudo usermod -a -G audio "${USER}"
           sudo bash ci/setup-xvfb.sh
 
-      - name: Set up snd-dummy
+      - name: Check if linux-modules-extra is available
+        id: check-snd-dummy
         if: (!(contains(matrix.extra, 'unittests') || contains(matrix.extra, 'vimtags')))
+        continue-on-error: true
+        run: |
+          sudo apt-get install -d -y linux-modules-extra-${{ env.LINUX_VERSION }}
+
+      - name: Set up snd-dummy
+        if: steps.check-snd-dummy.outcome == 'success' && (!(contains(matrix.extra, 'unittests') || contains(matrix.extra, 'vimtags')))
         env:
           DEST_DIR: ${{ env.TMPDIR }}/linux-modules-extra-${{ env.LINUX_VERSION }}
         run: |
           cd /lib/modules/${{ env.LINUX_VERSION }}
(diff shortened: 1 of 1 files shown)


chrisbra:
Hm, why did this not run any github CI Actions?

zdohnal:
@chrisbra I'm confused myself :D

zdohnal:
@chrisbra ok, my bad https://github.com/vim/vim/actions/runs/5864529562 - looks like {{ }} is only for yaml substitutions.

chrisbra:
relevant github issue: https://github.com/actions/runner-images/issues/8080

zdohnal:
But still the CI should go the report it is missing and skip... unless apt returns 0 if it doesn't find package :)

.github/workflows/ci.yml
```diff
@@ -180,24 +180,26 @@ jobs:
           sudo usermod -a -G audio "${USER}"
           sudo bash ci/setup-xvfb.sh
 
-      - name: Set up snd-dummy
+      - name: Check if linux-modules-extra is available
+        id: check-snd-dummy
         if: (!(contains(matrix.extra, 'unittests') || contains(matrix.extra, 'vimtags')))
```
jamacku, line 185: You can try this condition:

```suggestion
        if: (!(contains(matrix.extra, 'unittests') || contains(matrix.extra, 'vimtags'))) && runner.name != 'ubuntu-22.04'
```

[runner context](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context)

pheiduck:
Nice, but I would prefer caching of the package like I try to get ready here:
https://github.com/vim/vim/pull/12779

zdohnal:
> Nice, but I would prefer caching of the package like I try to get ready here: #12779

@pheiduck  That's a good idea - however the current problem is that package does not install at all in the container,  which makes CI fail (and people have to check the errors...), so unless you want to store the package on your own mirror and maintain it, caching does nothing at the moment for CI errors.

Caching of the downloaded packages can be applied on this PR once this PR is merged. (basically you merge the steps again and use only continue-on-error - maybe some additional conds in next steps will be needed too).

pheiduck:
@zdohnal yep, I know, I will continue working on this when the Package is available again.
Rebased to this PR when merged.

zdohnal:
@chrisbra macos test passed before, I don't see a reaso why it should fail now https://github.com/vim/vim/actions/runs/5866102553/job/15904274375?pr=12801 .

PR #12802, opened 2023-08-15
Github CI fails, because the SND Module is not available

chrisbra:
Disable it for now, so that tests can run


pheiduck:
I am a bit confused why my if condition is no working.
Because it should let the CI run also when the Package is not available. 🤔

chrisbra:
> Because it should let the CI run also when the Package is not available. 

Yeah, I think the `apt-cache` commands still succeeds (maybe because the package is still referenced), but the actual install commands fails instead.

chrisbra:
I think this commit properly checks the result: d47431954aaa5b2d9c8c2fc9504ad918fa47bfc6

chrisbra:
let me squash and re-submit

chrisbra:
closing in favor of #12801

PR #12804, opened 2023-08-15
Fix getcompletion() "cmdline" not working for command after :autocmd

zeertzjq:
Problem:    getcompletion() "cmdline" fails for command after :autocmd.
Solution:   Use set_cmd_context() instead of set_one_cmd_context().

### DIFF
--- a/src/usercmd.c
+++ b/src/usercmd.c
@@ -329,11 +329,7 @@ set_context_in_user_cmdarg(
 	return set_context_in_menu_cmd(xp, cmd, arg, forceit);
 #endif
     if (context == EXPAND_COMMANDS)
-    {
-	if (xp->xp_context == EXPAND_NOTHING)
-	    xp->xp_context = context;
 	return arg;
-    }
     if (context == EXPAND_MAPPINGS)
 	return set_context_in_map_cmd(xp, (char_u *)"map", arg, forceit, FALSE,
 							FALSE, CMD_map);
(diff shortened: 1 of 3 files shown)


src/usercmd.c
```diff
@@ -329,11 +329,7 @@ set_context_in_user_cmdarg(
 	return set_context_in_menu_cmd(xp, cmd, arg, forceit);
 #endif
     if (context == EXPAND_COMMANDS)
-    {
-	if (xp->xp_context == EXPAND_NOTHING)
-	    xp->xp_context = context;
```
yegappan, line 334: Is this change needed?  Will this break something else?

zeertzjq: It was just added two days ago to fix a problem with `getcompletion()`.

zeertzjq: After this PR, `set_one_cmd_context()` is only called by `set_cmd_context()`.

When this branch is reached, `set_context_in_user_cmdarg()` returns non-`NULL` to `set_context_by_cmdname()` -> `set_one_cmd_context()` -> `set_cmd_context()`, and `set_cmd_context()` calls `set_one_cmd_context()` again, which sets `xp->xp_context = EXPAND_COMMAND` immediately. Therefore this assignment is no longer needed.

yegappan (approved):

src/usercmd.c
chrisbra: okay thanks

PR #12805, opened 2023-08-15
fix: take account of skipcol in screenpos

### LINKED ISSUE #12547: screenpos() wrong with smoothscroll

tomtomjhj:
Fixes #12547

### DIFF
--- a/src/misc1.c
+++ b/src/misc1.c
@@ -517,12 +517,7 @@ plines_m_win(win_T *wp, linenr_T first, linenr_T last)
 	else
 #endif
 	{
-#ifdef FEAT_DIFF
-	    if (first == wp->w_topline)
-		count += plines_win_nofill(wp, first, TRUE) + wp->w_topfill;
-	    else
-#endif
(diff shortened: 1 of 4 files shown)


zeertzjq:
I think this will break other places where `plines_m_win()` is used. #12625 passes other tests.

zeertzjq:
Hmm, this also passes tests.

tomtomjhj:
Oh I wasn't aware of that PR.

PR #12810, opened 2023-08-15
Add QML support to vim

### LINKED ISSUE #12806: Upstreaming syntax files
I have a [repository](https://github.com/peterhoeg/vim-qml/) with syntax files for QML and was wondering if you are interested in getting it merged into vim upstream?

I no longer use vim, but it looks like quite a few people are using the plugin, so for the benefit of the project, I think it would make sense to get it included properly.

I must admit, I'm not sure how to do this while minimizing the impact for people using the plugin, so if anyone could send a link to somewhere this is described, I'll gladly do up the PR with the required changes to get this going.

ChaseKnowlden:
New syntax support for QML

Fixes #12806

### DIFF
--- a/runtime/ftplugin/qml.vim
+++ b/runtime/ftplugin/qml.vim
@@ -1,0 +1,37 @@
+" Vim filetype plugin file
+" Language: qml
+" Last change: 2014 Feb 8
+
+if exists( 'b:did_ftplugin' )
+   finish
+endif
+let b:did_ftplugin = 1
+
+let s:cpoptions_save = &cpoptions
+set cpoptions&vim
+
+" command for undo
+let b:undo_ftplugin =
+   \ 'let b:browsefilter = "" | ' .
+   \ 'setlocal ' .
+   \    'comments< '.
+   \    'commentstring< ' .
+   \    'formatoptions< ' .
+   \    'indentexpr<'
+
+if has( 'gui_win32' )
+\ && !exists( 'b:browsefilter' )
+   let b:browsefilter =
+      \ 'qml files (*.qml)\t*.qml\n' .
+      \ 'All files (*.*)\t*.*\n'
+endif
+
+" Set 'comments' to format dashed lists in comments.
+setlocal comments=sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,://
+setlocal commentstring=//%s
+
+setlocal formatoptions-=t
+setlocal formatoptions+=croql
+
+let &cpoptions = s:cpoptions_save
+unlet s:cpoptions_save
--- a/runtime/ftplugin/qml.vim
+++ b/runtime/ftplugin/qml.vim
@@ -1,6 +1,7 @@
 " Vim filetype plugin file
-" Language: qml
-" Last change: 2014 Feb 8
+" Language: QML
+" Maintainer: Chase Knowlden <haroldknowlden@gmail.com>
+" Last Change: 2023 Aug 16
 
 if exists( 'b:did_ftplugin' )
    finish
@@ -11,19 +12,12 @@ let s:cpoptions_save = &cpoptions
 set cpoptions&vim
 
 " command for undo
-let b:undo_ftplugin =
-   \ 'let b:browsefilter = "" | ' .
-   \ 'setlocal ' .
-   \    'comments< '.
-   \    'commentstring< ' .
-   \    'formatoptions< ' .
-   \    'indentexpr<'
+let b:undo_ftplugin = "setlocal formatoptions< comments< commentstring< indentexpr<"
 
-if has( 'gui_win32' )
-\ && !exists( 'b:browsefilter' )
+if (has("gui_win32") || has("gui_gtk")) && !exists( 'b:browsefilter' )
    let b:browsefilter =
-      \ 'qml files (*.qml)\t*.qml\n' .
-      \ 'All files (*.*)\t*.*\n'
+      \ 'QML files (*.qml,*.qbs)\t*.qml;*.qbs\n' .
+      \ 'All files\t*\n'
 endif
 
 " Set 'comments' to format dashed lists in comments.
--- a/runtime/ftplugin/qml.vim
+++ b/runtime/ftplugin/qml.vim
@@ -12,12 +12,12 @@ let s:cpoptions_save = &cpoptions
 set cpoptions&vim
 
 " command for undo
-let b:undo_ftplugin = "setlocal formatoptions< comments< commentstring< indentexpr<"
+let b:undo_ftplugin = "setlocal formatoptions< comments< commentstring<"
 
 if (has("gui_win32") || has("gui_gtk")) && !exists( 'b:browsefilter' )
    let b:browsefilter =
-      \ 'QML files (*.qml,*.qbs)\t*.qml;*.qbs\n' .
-      \ 'All files\t*\n'
+      \ 'QML Files (*.qml,*.qbs)\t*.qml;*.qbs\n' .
+      \ 'All Files\t*\n'
 endif
 
 " Set 'comments' to format dashed lists in comments.
--- a/runtime/indent/qml.vim
+++ b/runtime/indent/qml.vim
@@ -1,0 +1,59 @@
+" Vim indent file
+" Language:     QML
+" Author:       Robert Kieffer
+" URL:
+" Last Change:  2017-10-27
+"
+" Improved JavaScript indent script.
+
+" Indent script in place for this already?
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+
+setlocal indentexpr=GetJsIndent()
+setlocal indentkeys=0{,0},0),0],:,!^F,o,O,e,*<Return>,=*/
+
+" Only define functions once per session
+if exists("*GetJsIndent")
+  finish
+endif
+
+" Clean up a line of code by removing trailing '//' and '/* */' comments, and trimming
+" whitespace
+function! Trim(line)
+  return substitute(substitute(substitute(a:line, '// .*', '', ''), '/\* .* \*/', '', ''), '^\s*\|\s*$', '', 'g')
+endfunction
+
+function! GetJsIndent()
+  let num = v:lnum
+  let line = Trim(getline(num))
+
+  let pnum = prevnonblank(num - 1)
+  if pnum == 0
+    return 0
+  endif
+  let pline = Trim(getline(pnum))
+
+  let ind = indent(pnum)
+
+  " bracket/brace/paren blocks
+  if pline =~ '[{[(]$'
+    let ind += &sw
+  endif
+  if line =~ '^[}\])]'
+    let ind -= &sw
+  endif
+
+  " '/*' comments
+  if pline =~ '^/\*.*\*/'
+    " no indent for single-line form
+  elseif pline =~ '^/\*'
+    let ind += 1
+  elseif pline =~ '^\*/'
+    let ind -= 1
+  endif
+
+  return ind
+endfunction
--- a/runtime/indent/qml.vim
+++ b/runtime/indent/qml.vim
@@ -10,7 +10,7 @@ if exists("b:did_indent")
   finish
 endif
 let b:did_indent = 1
-let b:undo_indent = "setlocal indentexpr< indkeys<"
+let b:undo_indent = "setlocal indentexpr< indentkeys<"
 
 setlocal indentexpr=s:GetQmlIndent()
 setlocal indentkeys=0{,0},0),0],:,!^F,o,O,e,*<Return>,=*/
--- a/runtime/indent/qml.vim
+++ b/runtime/indent/qml.vim
@@ -9,7 +9,8 @@
 if exists("b:did_indent")
   finish
 endif
-let b:did_indent = "setlocal indentexpr< indkeys<"
+let b:did_indent = 1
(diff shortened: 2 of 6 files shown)


yegappan:
The tests are failing.   It looks like the change to the runtime/filetype.vim file to detect QML file type is missing.

Are the last lines in all the new files ending with a newline character?

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
+" Vim filetype plugin file
+" Language: qml
```
dkearns: Could you please capitalise QML?

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
...
+" Last change: 2014 Feb 8
```
dkearns: Could you please capitalise "Change"?

You'll note that the syntax highlighters only match "Last Change:" and not "Last change:".

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
...
+   finish
+endif
+let b:did_ftplugin = 1
+
+let s:cpoptions_save = &cpoptions
+set cpoptions&vim
+
+" command for undo
+let b:undo_ftplugin =
+   \ 'let b:browsefilter = "" | ' .
```
dkearns: This should delete `b:browsefilter` with `unlet! b:browsefilter`, rather than set it to an empty string.

Ideally this should only be added to `b:undo_ftplugin` if it was set by the ftplugin.  See https://github.com/vim/vim/blob/master/runtime/ftplugin/aap.vim for an example.

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
...
+   \ 'setlocal ' .
+   \    'comments< '.
+   \    'commentstring< ' .
+   \    'formatoptions< ' .
+   \    'indentexpr<'
+
+if has( 'gui_win32' )
```
dkearns: We also need to test for "gui_gtk" here.  See https://github.com/vim/vim/blob/master/runtime/ftplugin/aap.vim for an example.

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
...
+\ && !exists( 'b:browsefilter' )
+   let b:browsefilter =
+      \ 'qml files (*.qml)\t*.qml\n' .
```
dkearns: The filter text should be capitlaised - "QML Files".

Should `*.qbs` files also be included?  https://github.com/github-linguist/linguist/blob/916bd8f3df802fa98b0ea85539e67bd0b88ef158/lib/linguist/languages.yml#L5519C1-L5519C4

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
+" Vim indent file
+" Language:     QML
+" Author:       Robert Kieffer
+" URL:
+" Last Change:  2017-10-27
```
dkearns: Can you please normalise the date formats?  It's not required but it makes grepping  a little easier on occasion.

The default format is 2023 Aug 16 as used in the ftplugin.

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+  finish
+endif
+
+" Clean up a line of code by removing trailing '//' and '/* */' comments, and trimming
+" whitespace
+function! Trim(line)
+  return substitute(substitute(substitute(a:line, '// .*', '', ''), '/\* .* \*/', '', ''), '^\s*\|\s*$', '', 'g')
+endfunction
+
+function! GetJsIndent()
```
dkearns: This should be renamed to GetQmlIndent and other references changed.

You can make it script-local if desired.  See https://github.com/vim/vim/blob/master/runtime/indent/vb.vim for an example.

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,37 @@
...
+      \ 'All files (*.*)\t*.*\n'
```
dkearns: This should now be `"All Files\t*\n" so that files with no extension are also matched by this filter.  There's a [pending PR](https://github.com/vim/vim/pull/12759) to update this for all ftplugins.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1145 @@
+" Vim syntax file
+" Language:     QML
+" Maintainer:   Peter Hoeg <peter@hoeg.com>
```
dkearns: This should be changed to "Previous Maintainer:" assuming Peter won't be maintaining the Vim distributed file.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1145 @@
...
+" Define the default highlighting.
+" For version 5.7 and earlier: only when not done already
+" For version 5.8 and later: only when an item doesn't have highlighting yet
+if version >= 508 || !exists("did_qml_syn_inits")
+  if version < 508
+    let did_qml_syn_inits = 1
+    command -nargs=+ HiLink hi link <args>
+  else
+    command -nargs=+ HiLink hi def link <args>
+  endif
```
dkearns: The HiLink command has been removed from the distributed syntax files as Vim 6 was released over twenty years ago.

These lines can be reduced to, for example, `hi def link qmlComment Comment`

See https://github.com/vim/vim/blob/master/runtime/syntax/aap.vim for an example.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1145 @@
...
+" Based on javascript syntax (as is QML)
+
+if !exists("main_syntax")
+  if version < 600
+    syntax clear
+  elseif exists("b:current_syntax")
+    finish
+  endif
+  let main_syntax = 'qml'
+endif
```
dkearns, line 19: The version check has been removed from the distributed syntax files as Vim 6 was released over twenty years ago.

This can be simplified to:

```vim
  if exists("b:current_syntax")
    finish
  endif

  if !exists("main_syntax")
    let main_syntax = 'qml'
  endif
```

src/testdir/test_filetype.vim
```diff
@@ -551,6 +551,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     pyrex: ['file.pyx', 'file.pxd'],
     python: ['file.py', 'file.pyw', '.pythonstartup', '.pythonrc', 'file.ptl', 'file.pyi', 'SConstruct'],
     ql: ['file.ql', 'file.qll'],
+    qml: ['file.qml'],
```
dkearns: This will need a filetype detection entry in https://github.com/vim/vim/blob/master/runtime/filetype.vim

If *.qbs files are also going to be supported could you please add an entry and tests for that as well?

dkearns (changes requested):
Thanks for the PR.

I assume you're willing to maintain the files so can you please list yourself as "Maintainer: " in the header of each and also add yourself to https://github.com/vim/vim/blob/master/.github/CODEOWNERS?

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+" Improved JavaScript indent script.
+
+" Indent script in place for this already?
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+
+setlocal indentexpr=GetJsIndent()
+setlocal indentkeys=0{,0},0),0],:,!^F,o,O,e,*<Return>,=*/
```
dkearns, line 16: `b:undo_indent` needs to be set like `b:undo_ftplugin`.

```vim
let b:undo_indent = "setlocal indentexpr< indkeys<"
```

dkearns (changes requested):

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1141 @@
...
+" Define the default highlighting.
+" For version 5.7 and earlier: only when not done already
+" For version 5.8 and later: only when an item doesn't have highlighting yet
+if version >= 508 || !exists("did_qml_syn_inits")
+  if version < 508
+    let did_qml_syn_inits = 1
+    command -nargs=+ hi link <args>
+  else
+    command -nargs=+ hi def link <args>
+  endif
```
dkearns: All of this can be remove, so lines 1090-1099 and 1136.

That won't actually run now, without error, as there's no command name specified.

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,31 @@
...
+if exists( 'b:did_ftplugin' )
+   finish
+endif
+let b:did_ftplugin = 1
+
+let s:cpoptions_save = &cpoptions
+set cpoptions&vim
+
+" command for undo
+let b:undo_ftplugin = "setlocal formatoptions< comments< commentstring< indentexpr<"
```
dkearns: indentexpr isn't set here but in the indent plugin so it needs to be removed.

Sorry, I missed that first time around.

runtime/ftplugin/qml.vim
```diff
@@ -0,0 +1,31 @@
...
+
+if (has("gui_win32") || has("gui_gtk")) && !exists( 'b:browsefilter' )
+   let b:browsefilter =
+      \ 'QML files (*.qml,*.qbs)\t*.qml;*.qbs\n' .
+      \ 'All files\t*\n'
```
dkearns: "files" should be capitalised as "Files"

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+" Improved JavaScript indent script.
+
+" Indent script in place for this already?
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+let b:undo_indent = "setlocal indentexpr< indkeys<"
+
+setlocal indentexpr=GetQmlIndent()
```
dkearns: The GetQmlIndent function needs to be referenced everywhere with the `s:` prefix.

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+setlocal indentkeys=0{,0},0),0],:,!^F,o,O,e,*<Return>,=*/
+
+" Only define functions once per session
+if exists("*GetQmlIndent")
```
dkearns: This needs the `s:` prefix as well

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+  finish
+endif
+
+" Clean up a line of code by removing trailing '//' and '/* */' comments, and trimming
+" whitespace
+function! Trim(line)
```
dkearns: This should be script-local as well.  So the definition and all references need the `s:` prefix.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1141 @@
...
+if exists("b:current_syntax")
+  finish
+endif
+
+if !exists("main_syntax")
+  let main_syntax = 'qml'
+endif
+
+" Drop fold if it set but vim doesn't support it.
+if version < 600 && exists("qml_fold")
```
dkearns: If you want to do this you should use the has("folding") feature test rather than the version check and maybe test it down at line 1067 where you're defining the fold regions.

dkearns (changes requested):
Thanks for the updates.  There's just a few more changes needed.

chrisbra:
can you please resolve those mentioned things? Then we are ready to merge. Thanks

ChaseKnowlden:
Fixed all of them.

runtime/indent/qml.vim
```diff
@@ -0,0 +1,59 @@
...
+" Last Change:  2023 Aug 16
+"
+" Improved JavaScript indent script.
+
+" Indent script in place for this already?
+if exists("b:did_indent")
+  finish
+endif
+let b:did_indent = 1
+let b:undo_indent = "setlocal indentexpr< indkeys<"
```
dkearns: Should be `indentkeys<` not `indkeys<`.  Sorry for the typo in my earlier comment.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1130 @@
...
+endif
+
+if !exists("main_syntax")
+  let main_syntax = 'qml'
+endif
+
+" Drop fold if it set but vim doesn't support it.
+if has("folding")
+  unlet qml_fold
+endif
```
dkearns, line 24: This should be

```vim
if !has("folding")
  unlet! qmlfold
endif
```

You want to remove the flag if folding is not supported and it may not have been set so `unlet!` won't give an error if it doesn't exist.

runtime/syntax/qml.vim
```diff
@@ -0,0 +1,1130 @@
...
+
+syntax keyword qmlObjectLiteralType YAnimator
+
+syntax keyword qmlObjectLiteralType ZoomBlur
+
+" }}}
+
+if get(g:, 'qml_fold', 0)
+  syn match   qmlFunction      "\<function\>"
+  syn region  qmlFunctionFold  start="\<function\>.*[^};]$" end="^\z1}.*$" transparent fold keepend
```
dkearns: This doesn't look right.  `\z1` would need a corresponding `\z(\)` group in the start pattern.

It looks like the whitespace (indent) is supposed to match.  Try this

```vim
syn region  qmlFunctionFold  start="^\z(\s*\)\<function\>.*[^};]$" end="^\z1}.*$" transparent fold keepend
```

dkearns (changes requested):

dkearns:
LGTM, thanks.

Issue #12701, opened 2023-07-23
[vim9script] Type mismatch in new() not caught when exception is checked inside a function


lifepillar:
### Steps to reproduce

Source this script:

```vim
vim9script

class C
  def new(x: number) # NOTE: this bug depends on new() having a typed parameter
  enddef
endclass

def Check()
  try
    C.new('abc')
  catch
    echo $'OK! This exception was caught: {v:exception}'
  endtry
enddef

try
  C.new('abc')
catch
  echo $'OK! This exception was caught: {v:exception}'
endtry

Check() # Fails with E1013
```

The exception is caught correctly by the try-catch block at the script level, but E1013 is raised when executing `Check()`. Note that this bug shows up only when `new()`'s arguments are typed, as in the snippet above: if `new()`'s arguments are members, then the code works as one would expect. That is, all exceptions are caught if the definition of the class is changed as follows:

```vim
class C
  this.x: number
  def new(this.x)
  enddef
endclass
```

### Expected behaviour

An type mismatch exception should be caught also when checking for the exception happens inside a function.

### Version of Vim

9.0.1677

### Environment

macOS 13.4.1
Apple Terminal
xterm-256color
ZSH 5.9

LemonBoy:
Not sure this is a bug, `Check` is a function and is thus being bytecode-compiled on first use, the error you're seeing is not caught because it's not happening at runtime. Script-level statements, on the other hand, are not compiled making the type error/exception a runtime event.

h-east:
About the matter of a title, it is a specification.  Not a bug.
Please see `:help :def`.
```
                        The function will be compiled into instructions when
                        called, or when :disassemble or :defcompile is
                        used.  Syntax and type errors will be produced at that
                        time.
```


---

And actually, it is a bug that an exception occurs in the following code.
(Since it is a type error, it should be an error at compile time.)

> if new()'s arguments are members, then the code works as one would expect. That is, all exceptions are caught if the definition of the class is changed as follows:
```vim
class C
  this.x: number
  def new(this.x)
  enddef
endclass
```

When I investigated the Vim source code, I found that the type check was not performed when the formal argument was the object member ("this.~").

A patch has been created to fix this. I'll make a PR after adding the test.
Please wait.

errael:
IIUC, the behavior has changed from the initial description. 9.0.1922

PR #12816, opened 2023-08-16
Fix: vim9class constructor argument type checking bug

h-east:
Type checking is not performed when the formal argument of a vim9class constructor is an object member.

Related: #12701 ([my comment](https://github.com/vim/vim/issues/12701#issuecomment-1680387283))

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -528,6 +528,8 @@ def Test_class_new_with_object_member()
         this.num: number
         def new(this.str, this.num)
         enddef
+        def newVals(this.str, this.num)
+        enddef
       endclass
 
       def Check()
@@ -535,32 +537,12 @@ def Test_class_new_with_object_member()
           var c = C.new('cats', 2)
           assert_equal('cats', c.str)
           assert_equal(2, c.num)
-        catch
-          echo $'Exception was caught: {v:exception}'
-        endtry
-      enddef
 
-      Check()
-  END
-  v9.CheckScriptSuccess(lines)
-
-  lines =<< trim END
-      vim9script
-
-      class C
-        this.str: string
-        this.num: number
-        def newVals(this.str, this.num)
-        enddef
-      endclass
-
-      def Check()
-        try
-          var c = C.newVals('dogs', 4)
+          c = C.newVals('dogs', 4)
           assert_equal('dogs', c.str)
           assert_equal(4, c.num)
         catch
-          echo $'Exception was caught: {v:exception}'
+          assert_report($'Unexpected exception was caught: {v:exception}')
         endtry
       enddef
 
@@ -582,7 +564,7 @@ def Test_class_new_with_object_member()
         try
           var c = C.new(1, 2)
         catch
-          echo $'Exception was caught: {v:exception}'
+          assert_report($'Unexpected exception was caught: {v:exception}')
         endtry
       enddef
 
@@ -604,7 +586,7 @@ def Test_class_new_with_object_member()
         try
           var c = C.newVals('dogs', 'apes')
         catch
-          echo $'Exception was caught: {v:exception}'
+          assert_report($'Unexpected exception was caught: {v:exception}')
         endtry
       enddef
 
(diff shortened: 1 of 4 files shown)


src/testdir/test_vim9_class.vim
```diff
@@ -518,6 +518,101 @@ def Test_class_default_new()
...
+        enddef
+      endclass
+
+      def Check()
+        try
+          var c = C.new('cats', 2)
+          assert_equal('cats', c.str)
+          assert_equal(2, c.num)
+        catch
+          echo $'Exception was caught: {v:exception}'
```
yegappan: Can you use assert_report() here?

src/testdir/test_vim9_class.vim
```diff
@@ -518,6 +518,101 @@ def Test_class_default_new()
...
+        enddef
+      endclass
+
+      def Check()
+        try
+          var c = C.newVals('dogs', 4)
+          assert_equal('dogs', c.str)
+          assert_equal(4, c.num)
+        catch
+          echo $'Exception was caught: {v:exception}'
```
yegappan: Can you use assert_report() here?

src/testdir/test_vim9_class.vim
```diff
@@ -518,6 +518,101 @@ def Test_class_default_new()
...
+        this.num: number
+        def new(this.str, this.num)
+        enddef
+      endclass
+
+      def Check()
+        try
+          var c = C.new(1, 2)
+        catch
+          echo $'Exception was caught: {v:exception}'
```
yegappan: Can you use assert_report() here?

src/testdir/test_vim9_class.vim
```diff
@@ -518,6 +518,101 @@ def Test_class_default_new()
...
+        this.num: number
+        def newVals(this.str, this.num)
+        enddef
+      endclass
+
+      def Check()
+        try
+          var c = C.newVals('dogs', 'apes')
+        catch
+          echo $'Exception was caught: {v:exception}'
```
yegappan: Can you use assert_report() here?

src/testdir/test_vim9_class.vim
```diff
@@ -518,6 +518,101 @@ def Test_class_default_new()
...
+  v9.CheckScriptSuccess(lines)
+
+  lines =<< trim END
+      vim9script
+
+      class C
+        this.str: string
+        this.num: number
+        def newVals(this.str, this.num)
+        enddef
```
yegappan: The difference between this test and the previous test is that instead of using the new() constructor function, this test uses the  newVals() member function?  Is that correct?  Can these tests be combined into a single test?

src/vim9instr.c
```diff
@@ -1777,6 +1777,7 @@ generate_BLOBAPPEND(cctx_T *cctx)
     int
 generate_CALL(
 	cctx_T	    *cctx,
+	type_T	    *mtype,	// method type
```
yegappan: As the main argument for generate_CALL() is "ufunc", can you move the "mtype" argument to after "mi"?

src/vim9instr.c
```diff
@@ -1823,6 +1826,23 @@ generate_CALL(
 		if (ufunc->uf_arg_types == NULL)
 		    continue;
 		expected = ufunc->uf_arg_types[i];
+
+		// class constructor with object member
+		if (class_constructor && expected->tt_type == VAR_ANY)
+		{
+		    class_T *clp = mtype->tt_class;
```
yegappan, line 1835: Can you add a comment here describing what this block of code is doing?

yegappan (commented):
Thanks for working on these changes.

chrisbra:
thanks. looks good. awaiting some changes after @yegappan review.

src/testdir/test_vim9_class.vim
h-east: combined.

src/vim9instr.c
h-east: Done.

src/vim9instr.c
h-east: I have added a comment, but if there is an error or a better explanation, please change it:smile:

yegappan (approved):

PR #12828, opened 2023-08-17
configure: use AC_CHECK_SIZEOF to find sizeof(wchar_t)

floppym:
This fixes an error when cross compiling.

Bug: https://bugs.gentoo.org/889430

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -1980,114 +1980,6 @@ $as_echo "$ac_res" >&6; }
 
 } # ac_fn_c_check_func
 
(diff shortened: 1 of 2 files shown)


floppym:
I don't see how the test failure could possibly be related to this change.

chrisbra:
wondering, is this safe to include for 9.1?

floppym:
This change should have no effect on runtime behavior; it just fixes a build issue with cross compiling.

I think it should be safe to include for any release.

Issue #12652, opened 2023-07-10
Possible SEGV (null pointer dereference) in handle_viminfo_register()


GJDuck:
### Steps to reproduce

1. Copy the attached file to `~/.viminfo`
2. Run `vim`

Attachment: [viminfo.txt](https://github.com/vim/vim/files/11997840/viminfo.txt)

### Expected behaviour

The attached file is corrupt.  However, `vim` should handle it gracefully rather than crashing.

### Version of Vim

9.0 (tested both github HEAD and ubuntu distro version)

### Environment

Operating system: `Ubuntu 23.04`
Term: `xterm-256color`
Shell: `bash-5.2.15`

Pierre-Colin:
I did some debugging. The crash seems to come from the loop at [src/viminfo.c:1800](https://github.com/vim/vim/blob/4112acae0816e3f64c082b1e773dd5f4a100685e/src/viminfo.c#L1800). It assumes `vp[i + 6].bv_type == BVAL_STRING`. The crash occurs at the line `|3,1,1,1,1,0,71489,,"125"`. The empty entry breaks the assertion.

I don’t know how to best handle that, for `handle_viminfo_register` can’t return errors.

I tried inserting the following at the beginning of the loop body. Vim does diagnose an error in `.viminfo` around where that line is. However, it still segfaults when quitting.
```c
        if (vp[i + 6].bv_type != BVAL_STRING)
            continue;
```

chrisbra:
Hm, I guess this patch is needed then:
```
diff --git a/src/viminfo.c b/src/viminfo.c
index b772fc8f3..5e40974a4 100644
--- a/src/viminfo.c
+++ b/src/viminfo.c
@@ -1804,6 +1804,8 @@ handle_viminfo_register(garray_T *values, int force)
            y_ptr->y_array[i] = vp[i + 6].bv_string;
            vp[i + 6].bv_string = NULL;
        }
+       else if (vp[i + 6].bv_type == BVAL_EMPTY)
+           y_ptr->y_array = NULL;
        else
            y_ptr->y_array[i] = vim_strsave(vp[i + 6].bv_string);
     }
```

Pierre-Colin:
I don’t think so. The array `vp` can have more than 7 elements. @GJDuck’s example yields one of size 8, and the code parsing that line supports arbitrary lengths.

@chrisbra’s patch still segfaults for `BVAL_NR`, and it also leaks memory. I have addressed both issues and am putting together a PR.

[...]

PR #12845, opened 2023-08-18
Fix a segfault in `handle_viminfo_register`

Pierre-Colin:
This commit addresses #12652 by expanding upon @chrisbra’s proposed patch.

I replaced the test to prevent the segfault from happening for integer elements as well. I also added a `free` call to prevent memory leaks.

I tested the resulting build with no segfault. Valgrind found no definitely-lost memory block.

### DIFF
--- a/src/testdir/test_viminfo.vim
+++ b/src/testdir/test_viminfo.vim
@@ -614,6 +614,26 @@ func Test_viminfo_bad_syntax2()
   rviminfo Xviminfo
 endfunc
 
+" This used to crash Vim (GitHub issue #12652)
+func Test_viminfo_bad_syntax3()
+  let lines =<< trim END
+    call writefile([], 'Xvbs3.result')
+    qall!
+  END
+  call writefile(lines, 'Xvbs3script', 'D')
+
+  let lines = []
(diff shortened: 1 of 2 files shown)


yegappan:
Can you add the following test to testdir/test_viminfo.vim (after the existing Test_viminfo_bad_syntax2() function)?

```
" This used to crash Vim (Github issue #12652)
func Test_viminfo_bad_syntax3()
  let lines =<< trim END
    call writefile([], 'Xvbs3.result')
    qall!
  END
  call writefile(lines, 'Xvbs3script', 'D')

  let lines = []
  call add(lines, '|1,4')
  " bad viminfo syntax for register barline
  call add(lines, '|3,1,1,1,1,0,71489,,"125"') " empty line1
  call writefile(lines, 'Xviminfo', 'D')

  call RunVim([], [], '--clean -i Xviminfo -S Xvbs3script')
  call assert_true(filereadable('Xvbs3.result'))

  call delete('Xvbs3.result')
endfunc
```

Pierre-Colin:
I added the test you specified. Not sure how it works exactly, but it works. :smile:

I changed the string for a number, since numbers caused the crash too.

PR #12767, opened 2023-08-11
.cirrus.yml: skip pkg update

pheiduck:
PKG Install does not continue on 13.1

```
pkg update -f
Updating FreeBSD repository catalogue...
Fetching meta.conf: . done
Fetching packagesite.pkg: .......... done
Processing entries: 
Newer FreeBSD version for package zziplib:
To ignore this error set IGNORE_OSVERSION=yes
- package: 1302001
- running kernel: 1301000
Ignore the mismatch and continue? [y/N]: pkg: repository FreeBSD contains packages for wrong OS version: FreeBSD:13:amd64
Processing entries... done
Unable to update repository FreeBSD
Error updating repositories!
```

PR #12850, opened 2023-08-19
CI: set IGNORE_OSVERSION=yes

pheiduck:
Newer FreeBSD version for package zxing-cpp:
To ignore this error set IGNORE_OSVERSION=yes
Follow up: https://github.com/vim/vim/pull/12767


pheiduck:
Oh, I looked at a old run not needed

PR #12853, opened 2023-08-19
Fix type check tests are broken due to error message change

rhysd:
Some type check tests fail due to changed error message. The change seemed to be caused at c5d2744c045f9ad058cbc799f2434d90a6d83516. These failures occur on current master branch CI. This PR fixes the tests by adjusting the expected error messages.


chrisbra:
ah thanks!

PR #12860, opened 2023-08-19
Not able to get virtual text with the prop_list() function

yegappan:
Currently it is not possible to get the virtual text placed by the prop_add() function using the prop_list() function.
Return the text related properties for virtual text.  As this is the new feature introduced after Vim 9.0, it will be good
to get this included for the upcoming Vim 9.1 release.


chrisbra:
this makes sense and I think we should include it for vim 9.1

PR #12862, opened 2023-08-20
Fix crash when passing invalid buffer to undotree()

### LINKED ISSUE #12830: passing -1 to undotree() causes SEGV
### Steps to reproduce

Since 9.0 patch 1686, it is possible to specify a buf id for undotree(), to select undo histories other than the currently used buffer, which is great.

However if one, (quite likely by inputting a non-matching string into bufnr()) passes -1 to undotree(), vim segfaults.

1. :echom undotree(-1)

### Expected behaviour

An empty undo dictionary like for an unmodified buffer: `{'seq_last': 0, 'entries': [], 'time_cur': 0, 
 save_last': 0, 'synced': 1, 'save_cur': 0, 'seq_cur': 0}` was expected.

I'd prefer having an error thrown at the user.

### Version of Vim

9.0.1721

### Environment

likely irrelevant

### Logs and stack traces

```shell
will provide strace in a few hours
```

zeertzjq:
Problem: Crash when passing invalid buffer to undotree().
Solution: Use get_buf_arg() instead of tv_get_buf_from_arg().

Fix #12830

### DIFF
--- a/src/testdir/test_undo.vim
+++ b/src/testdir/test_undo.vim
@@ -134,6 +134,10 @@ func Test_undotree_bufnr()
   call assert_notequal(d1, d)
   call assert_equal(d2, d)
 
+  " error cases
+  call assert_fails('call undotree(-1)', 'E158:')
+  call assert_fails('call undotree("nosuchbuf")', 'E158:')
+
   " Drop created windows
   set ul&
   new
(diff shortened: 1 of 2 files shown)


yegappan (approved):

chrisbra:
Thanks. I have adjusted the test slightly.

PR #12864, opened 2023-08-20
Russian PO file updated to Vim version 9.0.1757

RestorerZ:
The "ru.po" file has been updated to Vim version 9.0.1757


chrisbra:
thank you!

PR #12867, opened 2023-08-20
Implement instanceof() builtin

LemonBoy:
Implemented in the same form as Python's isinstance because it allows for checking multiple class types at the same time.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -5052,6 +5052,14 @@ insert({object}, {item} [, {idx}])			*insert()*
 		Can also be used as a |method|: >
 			mylist->insert(item)
 
+instanceof({object}, {class})				*instanceof()*
+		The result is a Number, which is |TRUE| when the {object} argument is a
+		direct or indirect instance of a |Class| specified by {class}.
+		When {class} is a |List| the function returns |TRUE| when {object} is an
+		instance of any of the specified classes.
+		Example: >
+			instanceof(animal, [Dog, Cat])
+
 interrupt()						*interrupt()*
 		Interrupt script execution.  It works more or less like the
 		user typing CTRL-C, most commands won't execute and control
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -310,6 +310,7 @@ inputrestore()			Number	restore typeahead
 inputsave()			Number	save and clear typeahead
 inputsecret({prompt} [, {text}]) String	like input() but hiding the text
 insert({object}, {item} [, {idx}]) List	insert {item} in {object} [before {idx}]
+instanceof({object}, {class})	Number	|TRUE| if {object} is an instance of {class}
 interrupt()			none	interrupt script execution
 invert({expr})			Number	bitwise invert
 isabsolutepath({path})		Number	|TRUE| if {path} is an absolute path
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -5061,6 +5061,9 @@ instanceof({object}, {class})				*instanceof()*
 		Example: >
 			instanceof(animal, [Dog, Cat])
 
+<		Can also be used as a |method|: >
+			myobj->instanceof(mytype)
+
 interrupt()						*interrupt()*
 		Interrupt script execution.  It works more or less like the
 		user typing CTRL-C, most commands won't execute and control
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -277,6 +277,15 @@ arg_number(type_T *type, type_T *decl_type UNUSED, argcontext_T *context)
 }
 
 /*
+ * Check "type" is an object.
+ */
+    static int
+arg_object(type_T *type, type_T *decl_type UNUSED, argcontext_T *context)
+{
+    return check_arg_type(&t_object, type, context);
+}
+
+/*
  * Check "type" is a dict of 'any'.
  */
     static int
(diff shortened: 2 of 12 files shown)


src/evalfunc.c
```diff
@@ -2124,6 +2124,8 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_inputsecret},
     {"insert",		2, 3, FEARG_1,	    arg23_insert,
 			ret_first_arg,	    f_insert},
+    {"instanceof",	2, 2, FEARG_1,	    NULL,
```
yegappan: The argument types need to be checked here, so that the types are checked during compilation.  Instead of using "NULL",
can you implement a arg2_object_class function?

src/testdir/test_vim9_class.vim
```diff
@@ -2331,6 +2331,55 @@ def Test_call_method_in_extended_class()
   v9.CheckScriptSuccess(lines)
 enddef
 
+def Test_instanceof()
```
yegappan, line 2344: Can you also add a test to the test_vim9_builtin.vim file which passes different types of arguments to the new instanceof() function?

yegappan (changes requested):

yegappan (commented):
Can you also update the runtime/doc/usr_41.txt file to mention the new intanceof() function?

errael:
FYI, there's an old thread with comment from Bram about instanceof

https://groups.google.com/g/vim_dev/c/xtr3mpm-uWA/m/8V2UBLvIAAAJ

errael:
> comment from Bram about instanceof

The biggest issue Bram brought up at the time was same class name across multiple files or even same name in the same file. 

@LemonBoy Does this PR handle these issues OK?

He also mentioned an extension of typename that would handle the multiple classes with same name issue.

LemonBoy:
> Does this PR handle these issues OK?

The solution is simple, instead of taking strings we take a class object (or a list of them), this way there's no ambiguity.

Do you also need `IsAssignableFrom`?

errael:
> Do you also need `IsAssignableFrom`?

No.

Not that I can think of anyway. (it's just reversed argument isn't it?)

runtime/doc/builtin.txt
```diff
@@ -5050,6 +5051,14 @@ insert({object}, {item} [, {idx}])			*insert()*
...
 			mylist->insert(item)
 
+instanceof({object}, {class})				*instanceof()*
+		The result is a Number, which is |TRUE| when the {object} argument is a
+		direct or indirect instance of a |Class| specified by {class}.
+		When {class} is a |List| the function returns |TRUE| when {object} is an
+		instance of any of the specified classes.
+		Example: >
+			instanceof(animal, [Dog, Cat])
+
```
errael, line 5063: There's usually a "Can also be used as a method:" example.

src/vim9class.c
```diff
@@ -1904,5 +1904,75 @@ object_free_nonref(int copyID)
...
+    void
+f_instanceof(typval_T *argvars, typval_T *rettv)
+{
+    typval_T	*object_tv = &argvars[0];
+    typval_T	*classinfo_tv = &argvars[1];
+    listitem_T	*li;
+
+    rettv->vval.v_number = FALSE;
+
+    if (object_tv->v_type != VAR_OBJECT || object_tv->vval.v_object == NULL)
```
yegappan: Instead of this check, code similar to the following is needed here:

```
    if (check_for_object_arg(argvars, 0) == FAIL
                || check_for_class_or_list_arg(argvars, 1) == FAIL)
        return;
```

The functions check_for_object_arg() and check_for_class_or_list_arg() should be added to typval.c.
With this check in place, the last else statement below is not needed.

src/vim9class.c
```diff
@@ -1923,5 +1923,73 @@ object_free_nonref(int copyID)
...
+	}
+    }
+    else if (classinfo_tv->v_type == VAR_CLASS)
+    {
+	rettv->vval.v_number = class_instance_of(object_tv->vval.v_object->obj_class,
+		classinfo_tv->vval.v_class);
+    }
+    else
+    {
+	emsg(_(e_class_required));
```
yegappan: With the newly added call to the `check_for_class_or_list_arg()` function above, this error message is no longer needed.

chrisbra:
let me know, once you think this is ready

src/evalfunc.c
```diff
@@ -744,6 +753,20 @@ arg_string_or_func(type_T *type, type_T *decl_type UNUSED, argcontext_T *context
...
+ * Check "type" is a list of 'any' or a class.
+ */
+    static int
+arg_class_or_list(type_T *type, type_T *decl_type UNUSED, argcontext_T *context)
+{
+    if (type->tt_type == VAR_CLASS
+	    || type->tt_type == VAR_LIST
+	    || type_any_or_unknown(type))
+	return OK;
+    arg_type_mismatch(&t_class, type, context->arg_idx + 1);
```
yegappan, line 766: Can you find out why some of these lines are not covered by the tests?

LemonBoy: I guess it's because there's no test with a `def` function, I couldn't get the testing harness to collaborate (I guess the `class` must be declared outside of the function itself) and I got tired.

yegappan (approved):

yegappan:
Looks good to me.

mattn:
Sorry about off topic. Some functions return v:true/v:false or 1/0. Which would we should go? For example, digraph_set return v:true/v:false.

chrisbra:
> Good point. The instanceof() function is applicable only in a Vim9 script, so it makes sense to
change the return value to true/false instead of 1/0.

I'll squash it in

PR #12868, opened 2023-08-20
If type is not specified for a member, set it during object creation instead of during class definition

yegappan:


chrisbra:
is this ready? Does anything speak against including #12869 on top of it?

yegappan:
Yes. This PR is ready and can be merged.

PR #12875, opened 2023-08-21
Fix various Termdebug issues

seandewar:
Problem: Cleanup for :Var and :Asm buffers did not apply to prompt mode, and
E86 was possible if they were hidden.
Solution: Move cleanup to s:EndDebugCommon. Check that the buffers exist before
switching.

E86 could be triggered by simply `:quit`ting the :Asm or :Var windows due to the `bufhidden=wipe` changes.

---

Problem: :Asm does not work in prompt mode.
Solution: Make it work by handling disassembly-related messages properly.

The previous implementation depended on the typed or sent (via s:SendCommand())
"disassemble ..." message being visible to s:CommOutput(), but this was only
true for the terminal-based job.

A more robust solution would be to use GDB MI's -data-disassemble command. I may
implement this in a future PR.

---

Problem: Termdebug does not escape gdb messages properly.
Solution: Improve the logic. Do not mangle messages if they have inner escaped
quotes. Use line continuation comments properly.

Interestingly, due to the missing line continuation comments (`"\`), most of
these substitutions were ignored.

Presumably, this logic still isn't exact. For example, if a message ends in
`\\"`, the quote may be preserved, even though it's the `\` being escaped
(similar issues may exist for the other escapes). This may not be a problem in
practice, though.

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -229,10 +229,10 @@ endfunc
 func s:CloseBuffers()
   exe 'bwipe! ' . s:ptybuf
   exe 'bwipe! ' . s:commbuf
-  if s:asmbuf > 0
+  if s:asmbuf > 0 && bufexists(s:asmbuf)
     exe 'bwipe! ' . s:asmbuf
   endif
-  if s:varbuf > 0
+  if s:varbuf > 0 && bufexists(s:varbuf)
     exe 'bwipe! ' . s:varbuf
   endif
   s:running = 0
@@ -709,15 +709,7 @@ func s:EndTermDebug(job, status)
   endif
 
   exe 'bwipe! ' . s:commbuf
-  if s:asmbuf > 0
-    exe 'bwipe! ' . s:asmbuf
-  endif
-  if s:varbuf > 0
-    exe 'bwipe! ' . s:varbuf
-  endif
-  let s:running = 0
   unlet s:gdbwin
-
   call s:EndDebugCommon()
 endfunc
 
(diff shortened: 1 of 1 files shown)


laburnumT:
Thanks for taking care of that.

I'm not seeing an existing file that tests things to do with termdebug. It might be a good thing to create something like that.

yegappan (approved):

GitMensch:
> I may implement this in a future PR.

@seandewar: please do!
... and You may also have a good idea to improve #10393 :-)

skywind3000:
Can we have a `BreakToggle` command ? like in most IDE ?

Currently, the primitive `Clear` and `Break` commands are hard to use and will occupy two keymap.

If we have a `BreakToggle`, I can bind it to `<space>` and just use `<space>` to set or clear break points.

That will be much handy.

GitMensch:
> Can we have a BreakToggle command ? like in most IDE ?

That's a bit "not GDB like", because GDB has support for multiple breakpoints at the same place.

Especially when you add a condition to a breakpoint and then add commands for the breakpoint it is also quite useful to have multiple breakpoints in the same place.
(`:Break [pos] if something` is currently not implemented, so that part is missing, but you may do this in GDB or gdb scripts, also after setting the break in vim; for adding commands it would be nice if we could do  that in a popup, but one can switch to the GDB console for that, or source gdb scripts, too)

The current implementation supports that nicely, adding a break at the position and the Clear command only removes the newest one on that place.

skywind3000:
Adding a toggle command will not remove the existing ones

GitMensch:
So you suggest to drop (all) breakpoints on the current line if there are some (seems like there's no mi command available but `call s:SendCommand('-interpreter-exec mi "console ' . at . '"')` should do the trick) otherwise enable one, right?

laburnumT:
> A more robust solution would be to use GDB MI's -data-disassemble command. I may
> implement this in a future PR.

Please let me know if you start working on this as I've also been eyeing to rewrite that section.

GitMensch:
Let's move that out to a separate issue - and whoever writes first in this issue "I'm working on it" will do it.

PR #12876, opened 2023-08-21
Update Go syntax file with 1.21 builtins

corps-fini:
[Go 1.21](https://tip.golang.org/doc/go1.21) has new builtins:

- [`clear`](https://pkg.go.dev/builtin#clear)
- [`min`](https://pkg.go.dev/builtin#min)
- [`max`](https://pkg.go.dev/builtin#max)

### DIFF
--- a/runtime/syntax/go.vim
+++ b/runtime/syntax/go.vim
@@ -5,7 +5,7 @@
 " go.vim: Vim syntax file for Go.
 " Language:             Go
 " Maintainer:           Billie Cleek <bhcleek@gmail.com>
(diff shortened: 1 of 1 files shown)


berggeist:
> [Go 1.21](https://tip.golang.org/doc/go1.21) has new builtins [`min`](https://pkg.go.dev/builtin#min) and [`max`](https://pkg.go.dev/builtin#max).

Please consider also the new [`clear`](https://pkg.go.dev/builtin#clear) built-in function. Thanks.

PR #12888, opened 2023-08-22
src/configure.ac: Remove obsolete macros

illia-bobyr:
These macros are declared obsolete in autoconfg 2.69, which is almost 10 years old by now:

  https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html

They generate warnings when in a subsequent upgrade to autoconf 2.71.

`autoupdate` from autoupdate 2.71 suggests most of these changes, except that it also adds obsolete warnings, that where individually checked and removed.

Regenerated `src/auto/configure` by running:

  cd src
  autoconf2.69 --output=auto/configure configure.ac
  sed --in-place --expression='s@>config.log@>auto/config.log@g' auto/configure

---

This is one step to upgrade to autoconf 2.71.
Essentially, 1/3 of the changes in here: https://github.com/vim/vim/pull/11322

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -619,6 +619,7 @@ ac_includes_default="\
 # include <unistd.h>
 #endif"
 
+ac_header_list=
 ac_subst_vars='LTLIBOBJS
 LIBOBJS
 LINK_AS_NEEDED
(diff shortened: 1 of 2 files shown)


ilya-bobyr:
Second step: https://github.com/vim/vim/pull/12949

PR #12889, opened 2023-08-22
configure: Python3: Use sysconfig for -I

illia-bobyr:
It seems better to use tools provided by Python for determining the include directories, rather than construct them "manually".

Current system is broken when using virtual environments for python 3.11.4.  It used to work before, but now it detects a incorrect value for `-I`.

It would probably make sense to switch to a similar logic for lib folders, that is for the `-l` switch.  There are also `sysconfig.get_config_h_filename()` and
`sysconfig.get_makefile_filename()`, that could replace more Python specific logic in the current `configure{.ac,}`.

sysconfig provides the necessary tools since Python 2.7.


chrisbra:
thanks, I have included your 2 PRs, will you provide the last one to switch to autoconf 2.71 as well?

Issue #12043, opened 2023-02-22
vim9class: null_object - constant type 17 not supported


errael:
### Steps to reproduce

(It may be that null_object is only partially implemented, and so this is not unexpected)

When there's clarification, I can try expected behavior

```
vim9script

# OK
var nullo = null_object
def F(): any
    return nullo
enddef

# Maybe OK: output: type(F()) 13
echo "type(F())" type(F())

# Maybe OK: output: typename(F()) object<Unknown>
echo "typename(F())" typename(F())

# OK
var o1: any
o1 = nullo

# OK
echo o1 == null_object ? "o1 == null_object" : "o1 != null_object"

# FAIL: expect o1 == null
echo o1 == null ? "o1 == null" : "o1 != null"

# FAIL compilation error
### This works as of patch 9.0.1357
def G()
    var x = null_object
enddef

class C
endclass

# FAIL
var o2: C
o2 = null_object
```

### Expected behaviour

The expected results are indicated in the steps to reproduce.

### Version of Vim

9.0.1333

### Environment

ubuntu

errael:
This issue has several examples of using null_object. 3 of them failed when the issue was filed. One of the failures now works. In particular
```
        def G()
            var x = null_object
        enddef
```
Used to fail compilation. It was fixed by https://github.com/vim/vim/commit/c4e1b86cb0d88fa5ec1141d3c600e026dcc1bc21

errael:
Some additional output and comments for the last failure
```
class C
endclass

# FAIL
var o2: C
echo "o2:" o2
echo "o2:" typename(o2)
o2 = null_object
```
there's the output
```
    o2: object of [unknown]
    o2: object<Unknown>
    Error detected while processing command line..script /.../12043-b.vim:
    line    9:
    E1012: Type mismatch; expected object<C> but got object<Unknown>
```
It seems like a bug, otherwise there would be no way to null out an object, like for garbage collection. As a workaround, I guess one could craft an empty object of the correct type and use it as a null analog.

errael:
And there's
```
$ srcgrep null_class
./src/eval.c:	case 10: if (STRNCMP(s, "null_class", 10) == 0)
./src/testdir/test_vim9_class.vim:def Test_using_null_class()
./src/testdir/test_vim9_class.vim:      @_ = null_class.member
```

yegappan:
[...]

Issue #12006, opened 2023-02-16
[vim9script] "E1177: For loop on class not supported" when accessing static member via alias


lifepillar:
### Steps to reproduce

Source this script:

```vim
vim9script

class Meta
  static x = 1
endclass

def Foo()
  const MetaAlias = Meta

  for _ in [1, 2]
    const y = MetaAlias.x
    echo y
  endfor
enddef

Foo()
```

Vim should raise `E1177: For loop on class not supported`. Replacing the first line in the body of the for loop with `const y = Meta.x` fixes the error.

### Expected behaviour

The code shoudl run just fine and echo `1` twice.

### Version of Vim

9.0.1311

### Environment

macOS 13.2
Apple Terminal
xterm-256color
ZSH 5.8.1

brammool:
There doesn't appear to be anything wrong with the for loop, I don't see how you can expect E1177 for this.

The real problem is ```const MetaAlias = Meta```.  What is the type of MetaAlias here?  It's not an object and we don't have a type that is a class.  I am not sure this is ever expected to work.  You could perhaps use a typedef (once that gets implemented).

I wonder why you came up with this code.  Just using "Meta.x" works, as you mention.  What do you need MetaAlias for?

lifepillar:
>What is the type of MetaAlias here? It's not an object and we don't have a type that is a class. I am not sure this is ever expected to work.

Ok, I see. Feel free to close this issue.

brammool:
Once something like a typedef has been implemented please check this again.  Closing for now.

PR #12890, opened 2023-08-23
Improve null_object usage: assignment/compare

errael:
Address #12043
Fix "xvar == null", where xvar might have been assigned null_object.
Fix compilation failure: "var o2: C = null_object".

### DIFF
--- a/src/typval.c
+++ b/src/typval.c
@@ -1578,12 +1578,15 @@ typval_compare_null(typval_T *tv1, typval_T *tv2)
 #ifdef FEAT_JOB_CHANNEL
 	    case VAR_CHANNEL: return tv->vval.v_channel == NULL;
 #endif
+	    // TODO: null_class handling
+	    // case VAR_CLASS: return tv->vval.v_class == NULL;
 	    case VAR_DICT: return tv->vval.v_dict == NULL;
 	    case VAR_FUNC: return tv->vval.v_string == NULL;
 #ifdef FEAT_JOB_CHANNEL
 	    case VAR_JOB: return tv->vval.v_job == NULL;
 #endif
 	    case VAR_LIST: return tv->vval.v_list == NULL;
+	    case VAR_OBJECT: return tv->vval.v_object == NULL;
 	    case VAR_PARTIAL: return tv->vval.v_partial == NULL;
 	    case VAR_STRING: return tv->vval.v_string == NULL;
 
--- a/src/vim9type.c
+++ b/src/vim9type.c
@@ -907,6 +907,8 @@ check_type_maybe(
 		return MAYBE;	// use runtime type check
 	    if (actual->tt_type != VAR_OBJECT)
 		return FAIL;	// don't use tt_class
+	    if (actual->tt_class == NULL)
+		return OK;	// A null object matches
 
 	    if (class_instance_of(actual->tt_class, expected->tt_class) == FALSE)
 		ret = FAIL;
--- a/src/proto/vim9instr.pro
+++ b/src/proto/vim9instr.pro
@@ -28,6 +28,7 @@ int generate_PUSHCHANNEL(cctx_T *cctx);
 int generate_PUSHJOB(cctx_T *cctx);
 int generate_PUSHBLOB(cctx_T *cctx, blob_T *blob);
 int generate_PUSHFUNC(cctx_T *cctx, char_u *name, type_T *type, int may_prefix);
+int generate_PUSHOBJ(cctx_T *cctx);
 int generate_AUTOLOAD(cctx_T *cctx, char_u *name, type_T *type);
 int generate_GETITEM(cctx_T *cctx, int index, int with_op);
 int generate_SLICE(cctx_T *cctx, int count);
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -337,7 +337,7 @@ def Test_object_not_set()
       var bg: Background           # UNINITIALIZED
       echo Colorscheme.new(bg).GetBackground()
   END
-  v9.CheckScriptFailure(lines, 'E1012: Type mismatch; expected object<Background> but got object<Unknown>')
+  v9.CheckScriptFailure(lines, 'E1360:')
 
   # TODO: this should not give an error but be handled at runtime
   lines =<< trim END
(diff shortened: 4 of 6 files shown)


errael:
@yegappan This fixes some cases, I haven't played with it thoroughly yet and it's draft. I copied in t_object stuff from the instanceof PR; I'll update when that gets merged. It's a relatively small PR and no new tests yet. Simple fixes to COMPAREANY/COMPARENULL. And a compile fix.

But I have questions...

- `generate_PUSHOBJ` now invoked from another file. How to create proto? Manually?
- Would `t_object_null` be better than `t_object`?
- I'm not sure about the compile fix; I left a few TODO in there, comments appreciated.

I ran
```
make $(ls test_vim9*.vim | sed 's/\.vim//g'|grep -v test_vim9_builtin)
```
without errors. test_vim9_builtin got 3 errors with or without this PR.

src/vim9type.c
```diff
@@ -907,6 +907,8 @@ check_type_maybe(
 		return MAYBE;	// use runtime type check
 	    if (actual->tt_type != VAR_OBJECT)
 		return FAIL;	// don't use tt_class
+	    if (actual == &t_object)
```
yegappan: Instead of this check, can you use the following?
```
+           // tt_class is NULL for an "unknown" object (either a null object
+           // or an uninitialized object)
+           if (actual->tt_class == NULL)
+               return OK;
+
```

src/vim9type.c
```diff
@@ -907,6 +907,8 @@ check_type_maybe(
...
+		return OK;	// NOTE: t_object *is* a null_object
```
yegappan: I am working on adding the test_null_class() and test_null_object() functions.  I am using the following test.  You can leverage this test for this PR and remove the references to the test_null_class() and test_null_object() functions:

```
" Test for the test_null_object() and test_null_class() functions
def Test_null_class_object()
  var lines =<< trim END
    vim9script

    var c = test_null_class()
    var obj = test_null_object()
    assert_true(c == null_class)
    assert_true(obj == null_object)
    assert_fails('echo c != obj', 'E1072:')
    assert_equal('class<Unknown>', typename(c))
    assert_equal('object<Unknown>', typename(obj))
    assert_true(c == null)
    assert_true(obj == null)

    var nullo = null_object
    def F(): any
      return nullo
    enddef
    assert_equal('object<Unknown>', typename(F()))

    var o1: any = nullo
    assert_true(o1 == null_object)
    assert_true(o1 == null)

    def G()
      var x = null_object
    enddef

    class C
    endclass
    var o2: C
    o2 = null_object
  END
  v9.CheckScriptSuccess(lines)
enddef
```

src/vim9type.c
```diff
@@ -907,6 +907,8 @@ check_type_maybe(
...
```
yegappan, line 912: With the support for the null_object, the existing Test_object_not_set() test in test_vim9_class.vim is failing with the `E1360: Using a null object` error instead of the `E1012: Type mismatch; expected object<Background> but got object<Unknown>` error.  You need to update that test with the new error.

src/globals.h
```diff
@@ -534,7 +534,15 @@ EXTERN int	garbage_collect_at_exit INIT(= FALSE);
 #define t_super			(static_types[80])
 #define t_const_super		(static_types[81])
 
-EXTERN type_T static_types[82]
+// ================= taken from the instanceof patch
```
yegappan: I assume you will remove this comment once the instanceof() PR is merged?

src/globals.h
```diff
@@ -534,7 +534,15 @@ EXTERN int	garbage_collect_at_exit INIT(= FALSE);
...
+// TODO: should this be t_object_null
```
yegappan: We can keep the below definitions instead of renaming this to t_object_null.

yegappan (changes requested):

src/vim9type.c
errael: I had it that way initially; got test failures. I didn't track it down, I'll put it back and check out the failures.

src/globals.h
errael: Right

src/vim9type.c
yegappan: With this change, the tests are passing for me (except for the one change to Test_object_not_set() mentioned in the other comment).

src/globals.h
errael: OK. I don't really grok what these static types are all about and how they're used. So any a VAR_OBJECT with type where `tt_class == NULL` is a null object.

src/vim9type.c
errael: Is this getting merged soon? These new tests go into test_vim9_class.vim?

yegappan: The new test function is in the test_vim9_class.vim file.  You can leverage this test and modify it as needed and feel free to include this in your PR.  I will wait for your PR to be merged.

errael: OK. And this PR needs #12867.

src/typval.c
```diff
@@ -1578,12 +1578,15 @@ typval_compare_null(typval_T *tv1, typval_T *tv2)
 #ifdef FEAT_JOB_CHANNEL
 	    case VAR_CHANNEL: return tv->vval.v_channel == NULL;
 #endif
+	    // TODO: null_class handling
```
yegappan, line 1581: Is this comment still applicable?

yegappan (changes requested):

src/typval.c
errael: Yes. I didn't do anything for or about a variable that aliases a class.  I accidentally found out that you can do
```
var vclass = C
var o = vclass.new()
```
and mentioned it in https://github.com/vim/vim/issues/12043#issuecomment-1688584460 where I questioned whether or not it was supposed to work. Other than that, I didn't try anything. I've since found in todo.txt
> When "Meta" is a class, is "const MetaAlias = Meta" allowed?  It should
> either work or given an error. possibly give an error now and implement it
> later (using a typedef).  #12006

So this is a pretty open question. Might be good idea to make the assignment fail until there's a plan on how to handle this.

I'm pretty sure it won't work in general. I don't think there is a `class` type. Certainly `var foo: class` doesn't work.

But, if you want `vclass == null` then the commented part is needed, along with plenty more.

yegappan (approved):

errael:
Rebase to current master

PR #12891, opened 2023-08-23
Fix wrong cursor position with 'showbreak' and lcs-eol

zeertzjq:
Problem: Wrong cursor position with 'showbreak' and 'listchars' "eol".
Solution: Add size of 'showbreak' before when 'listchars' "eol" is used.
Also fix wrong cursor position with wrapping virtual text on empty line
and 'showbreak'.


zeertzjq:
I dedented a block, the diff is smaller if viewed with [`?w=1`](https://github.com/vim/vim/pull/12891/files?w=1).

PR #12894, opened 2023-08-23
CI: use apt-fast to download pkgs in parallel

pheiduck:


pheiduck:
can not be used when pkgs are saved into a variable.

PR #12895, opened 2023-08-23
runtime(menu): define shortcut for File->Open Tab

chrisbra:
Seems missing as noted by Antonio Giovanni Colombo. So add it and use the 'T' as shortcut, which does not seem to be used in the File dialog.

Verified on Windows.


ychin:
@chrisbra I think this needs to update all the localization files. I filed #12993 to reference this.

PR #12900, opened 2023-08-23
Add 'syntax' completion

dkearns:
Problem:  The 'syntax' option has no completion.
Solution: Add syntax option completion.

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -7349,6 +7349,14 @@ set_context_in_set_cmd(
 
     xp->xp_pattern = p + 1;
 
+#ifdef FEAT_SYN_HL
+    if (options[opt_idx].var == (char_u *)&p_syn)
+    {
+	xp->xp_context = EXPAND_OWNSYNTAX;
+	return;
+    }
+#endif
+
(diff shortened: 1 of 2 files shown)


src/optiondefs.h
```diff
@@ -2434,7 +2434,7 @@ static struct vimoption options[] =
 			    {(char_u *)0L, (char_u *)0L}
 #endif
 			    SCTX_INIT},
-    {"syntax",	    "syn",  P_STRING|P_ALLOCED|P_VI_DEF|P_NOGLOB|P_NFNAME,
+    {"syntax",	    "syn",  P_STRING|P_EXPAND|P_ALLOCED|P_VI_DEF|P_NOGLOB|P_NFNAME,
```
chrisbra: why is the `P_EXPAND` flag required?

dkearns: It isn't, sorry.

The patch is years old at my end but it looks like it was just copying the incorrect implementation of `:set filetype=` completion.

If the updated patch looks right to you I'll also update it for 'filetype'.

PR #12902, opened 2023-08-23
runtime: Distribute the editorconfig with vim

### LINKED ISSUE #2286: Proposal: ship with EditorConfig plugin by default
[EditorConfig][1] is an editor-independent file format to describe
common editor configuration settings for a project, such as tab size,
character set, etc.

Many editors and IDEs support EditorConfig files (the web page has a
list); for Vim there's the [editorconfig-vim][2] plugin.

I think it would be a good idea to ship this with Vim by default. I
understand that there are *many* useful and popular plugins that Vim
could ship with by default, but most of those are "personal" plugins
that affect only the user's work-flow, whereas this plugin is one which
helps Vim users work well with other people (regardless of which editor
they use).

Modelines kind of solve the same problem, but are not perfect:

1. It's specific to Vim.
2. Needs to be added to every single file (potentially hundreds or
   thousands of them), which is kind of tedious and makes changing
   settings or adding new ones hard.
3. Modelines won't work when creating a new file.

Is this something you're willing to consider? The existing Vim plugin
works very well, but if changes need to be made then I can do that.

[1]: http://editorconfig.org/
[2]: https://github.com/editorconfig/editorconfig-vim

chrisbra:
This is the editorconfig-vim plugin Commit e014708e917b457e8f6c57f357d55dd3826880d4 from https://github.com/editorconfig/editorconfig-vim

closes: #2286
related: https://github.com/editorconfig/editorconfig-vim/issues/223

Signed-off-by: Christian Brabandt <cb@256bit.org>

### DIFF
--- a/runtime/pack/dist/opt/editorconfig/.appveyor.yml
+++ b/runtime/pack/dist/opt/editorconfig/.appveyor.yml
@@ -1,0 +1,105 @@
+# appveyor.yml for editorconfig-vim.  Currently only tests the core.
+# Modified from https://github.com/ppalaga/ec4j/commit/1c849658fb189cd95bc41af95acd43b4f0d75a48
+#
+# Copyright (c) 2017--2019 Angelo Zerr and other contributors as
+# indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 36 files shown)


chdiza:
I don't think it's appropriate for this commit to include a plugin-specific test-suite and a bunch of plugin-specific git detritus?

There is no reason why a bundled plugin should have to be identical to the project's upstream git repo, is there?  People who want to continually update it can do it in their ~/.vim using the pack infrastructure that was made specifically to make this easier.

chdiza:
Looks good, but I'd also remove mkzip.sh.  Thanks!

PR #12916, opened 2023-08-25
Update Ukrainian translation

sakhnik:
Here's the Ukrainian translation update. Rest in peace, Bram.

### DIFF
--- a/src/po/uk.cp1251.po
+++ b/src/po/uk.cp1251.po
@@ -12,18 +12,18 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: vim 8.1\n"
+"Project-Id-Version: Vim Ukrainian\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-05-02 17:40+0300\n"
-"PO-Revision-Date: 2022-05-20 17:12+0300\n"
+"POT-Creation-Date: 2023-08-24 11:51+0300\n"
+"PO-Revision-Date: 2023-08-25 17:16+0300\n"
 "Last-Translator: Анатолій Сахнік <sakhnik@gmail.com>\n"
 "Language-Team: Ukrainian\n"
 "Language: uk\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=cp1251\n"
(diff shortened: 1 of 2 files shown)


src/po/uk.cp1251.po
```diff
@@ -12,18 +12,18 @@
...
-"POT-Creation-Date: 2022-05-02 17:40+0300\n"
-"PO-Revision-Date: 2022-05-20 17:12+0300\n"
+"POT-Creation-Date: 2023-08-24 11:13+0300\n"
+"PO-Revision-Date: 2023-08-25 13:12+0300\n"
 "Last-Translator: ������� ������ <sakhnik@gmail.com>\n"
 "Language-Team: Ukrainian\n"
 "Language: uk\n"
 "MIME-Version: 1.0\n"
-"Content-Type: text/plain; charset=cp1251\n"
+"Content-Type: text/plain; charset=UTF-8\n"
```
k-takata: This looks wrong.
`uk.cp1251.po` should be automatically updated by running `make` in the `src/po/` directory.

sakhnik: It didn't, by the way. But I've updated it manually. Thanks!

src/po/uk.cp1251.po
```diff
@@ -1,7 +1,7 @@
 #
 # Ukrainian translation for Vim
 #
-# Generated from uk.po, DO NOT EDIT
+# Original translations
```
k-takata: This looks also wrong.
This should not happen if you run `make`.

sakhnik: Sorry

src/po/uk.cp1251.po
k-takata: > It didn't, by the way.

Hmm. If `uk.cp1251.po` is newer than `uk.po`, `make` will not update it.
I'm not sure if this is the case, though.

sakhnik: All right, that was my manual intervention that messed things up. When I just edit uk.po and leave uk.cp1251.po alone, everything works fine. I just need to remember this.

sakhnik:
Thanks, @chrisbra. I'm still refining the messages while updating the translation of Neovim. So I'm going to follow up with another pull request soon.

PR #12917, opened 2023-08-25
Suppress C4090 warnings

k-takata:
MSVC shows the following warnings:
```
strings.c(2436): warning C4090: 'function': different 'const' qualifiers
strings.c(2774): warning C4090: 'function': different 'const' qualifiers
strings.c(3865): warning C4090: 'function': different 'const' qualifiers
```

Suppress them.

### DIFF
--- a/src/strings.c
+++ b/src/strings.c
@@ -2433,7 +2433,8 @@ adjust_types(
 	if (*ap_types == NULL)
 	    new_types = ALLOC_CLEAR_MULT(const char *, arg);
 	else
-	    new_types = vim_realloc(*ap_types, arg * sizeof(const char *));
+	    new_types = vim_realloc((char **)*ap_types,
+						arg * sizeof(const char *));
 
 	if (new_types == NULL)
 	    return FAIL;
@@ -2771,7 +2772,7 @@ parse_fmt_types(
     return OK;
 
 error:
-    vim_free(*ap_types);
+    vim_free((char**)*ap_types);
     *ap_types = NULL;
     *num_posarg = 0;
     return FAIL;
@@ -3862,7 +3863,7 @@ vim_vsnprintf_typval(
     if (tvs != NULL && tvs[num_posarg != 0 ? num_posarg : arg_idx - 1].v_type != VAR_UNKNOWN)
 	emsg(_(e_too_many_arguments_to_printf));
 
-    vim_free(ap_types);
+    vim_free((char*)ap_types);
     va_end(ap);
 
     // Return the number of characters formatted (excluding trailing nul


vimpostor:
@chrisbra Is there any reason, why we still don't merge pull requests and instead push a different commit that closes them?
This is a very outdated git workflow and it's a bit annoying, because you can no longer easily see in the web UI which PRs were merged and which ones were dropped.

If I understand correctly, the reason for the separate commit is that you need to change the commit message such that we have a tagged patch `9.0.xxxx` and such that the commit message has lines for `Problem:` and `Solution:`.

But this can all easily done by using the "Squash and merge" button, where Github always gives you the chance to edit the commit message before merging.

Is there any reason why we don't merge like that, it would certainly be a lot less weird for contributors having closed pull requests all the time.

k-takata:
Not only updating the commit message but also the version number must be updated.

chrisbra:
Yes, exactly for the minor version number

nickspoons:
There are other projects that do this, for example curl - there is a recent brief discussion of their workflow [here](https://github.com/curl/curl/pull/11729). It lets them keep a clean and consistent history, and to incorporate contributions from non-github sources, e.g. `git send-email`. Github not showing the PRs as merged is a github UI issue.

vimpostor:
> Github not showing the PRs as merged is a github UI issue.

It's easy to blame Github here, but there is absolutely no way to detect this. After all both the SHA of the pushed commit and the commit message are different. Even if you were to compare the diffs, it wouldn't work because the pushed commit contains an additional diff in `src/version.c`.
Sourcehut (which is mailing-list based) has a special email header dedicated to marking a patch as applied: https://man.sr.ht/lists.sr.ht/#email-controls. There is no equivalent thing for Github and it would also be less elegant because such an extra header would need to end up in the actual commit message.

I know there are several other famous projects that just close the PRs instead of merging, e.g. this also annoys me in tmux: https://github.com/tmux/tmux/issues/3218#issuecomment-1316699416
But in tmux it's a whole different beast, because main development happens in OpenBSD cvs, while in our case the only problem seems to be the `included_patches` array in `version.c`. Personally I doubt that the `included_patches` feature is even that useful, given that 99% of downstream consumers probably don't cherry-pick individual patches, but use all patches up to a given point.

chrisbra:
> But in tmux it's a whole different beast, because main development happens in OpenBSD cvs, while in our case the only problem seems to be the included_patches array in version.c. Personally I doubt that the included_patches feature is even that useful, given that 99% of downstream consumers probably don't cherry-pick individual patches, but use all patches up to a given point.

FWIW: One thing why I like the minor patch number: When reporting problems it is easily possible to see whether a certain fix has been included in a version or not and whether or not there exists a newer version (and how many). That's not so easily possible with hash strings. 

But as I said before, the current process doesn't scale well, so we may change it to something different, if this is to cumbersome and all agree (after the next release).

benknoble:
PS Might I suggest we move some unrelated conversation to a GitHub Discussion or separate Issue?

PR #12920, opened 2023-08-26
Add support for private object/class methods

yegappan:
If the name of a class or object method starts with an underscore, then make it a
private method.  It can be used by class/object methods in the current class
and by classes derived from the current class.

### DIFF
--- a/runtime/doc/todo.txt
+++ b/runtime/doc/todo.txt
@@ -129,6 +129,7 @@ Further Vim9 improvements, possibly after launch:
 	    or: def _Func()
     Perhaps use "private" keyword instead of "_" prefix?
   - "final" object members - can only be set in the constructor.
+  - Support export/import of classes and interfaces.
   - Cannot use class type of itself in the method (Issue #12369)
   - Cannot use an object method in a lambda  #12417
 	Define all methods before compiling them?
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -178,6 +178,26 @@ number to the total number of lines: >
 	enddef
 
 
+Private methods ~
+If you want object methods to be accessible only from other methods of the
+same class and not used from outside the class, then you can make them
+private.  This is done by prefixing the method name with an underscore: >
+
+    class SomeClass
+	def _Foo(): number
+	  return 10
+	enddef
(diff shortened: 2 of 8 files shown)


errael:
A coding pattern I use more and more frequently  in another language is private constructor and static method that returns an object. I have no idea how useful that would be in vim9script. Of course it's a moot point at this time since
```
vim9script
class C
    static def Create(): C
    enddef
endclass
```
gives `E1010: Type not recognized: C`

And a private constructor could be added later without breaking source compatibility.

errael:
Doh! But the following does work
```
vim9script
class C
    static def Create(): any
        return C.new()
    enddef
endclass

var x: C = C.Create()
```

yegappan:
> 
> To recap, when a class variable is declared "public" it is readable and
> writable from outside the class. If "public" is omitted, the variable is
> readable but not writable. If the variable name starts with an underscore,
> it's private and neither readable nor writable from outside the class.
> 

Yes.  Just to clarify, this PR introduces the support for private class/object
methods and doesn't change the accessibility of the class/object members
described above.

>
> This is already quite a lot to remember, imo.
> 
> In my opinion the current syntax for classes is too complex. I would remove
> the "public" keyword entirely and just use underscores for both variables and
> methods. There's no need for read-only class variables. If someone needs to
> access a variable from outside the class just write a setter like in any
> other language. Vimscript should stay as simple as possible.
> 
> Like others already mentioned on `vim_dev`, an alternative to underscores is
> to use "public" and "private" access modifiers for both variables and
> methods.
>

I do have the changes ready to make the object members "public" by default.
I will send a separate email on this and open a separate PR.

yegappan:
I have created the PR https://github.com/vim/vim/pull/12932 to make the member variable access
public by default.

Shane-XB-Qian:
> Like others already mentioned 

that maybe was me? 
aha, i've pointed out several times, but not sure the final version was going to be now -_-#
anyway, maybe it was designed to be completed and secured, but truly a bit complex,
if you insist leading underscore was better for private var/method, it's ok,
but meanwhile perhaps we may no need public/private keyword anymore,
as for interface/abstract (and 'static'?), vim9 maybe basic class was good enough, specially for scripting.

PR #12923, opened 2023-08-26
runtime(vim): Improve :let-heredoc syntax highlighting

dkearns:
"trim" and "eval" are allowed in any order.

@cecamp, I sent this to you about a year ago but didn't receive a reply.  This is unusual so I'm submitting the change here as well in case my email has dissipated in the ether.


chrisbra:
@cecamp please

dkearns:
I think this can be included @chrisbra, it's an isolated change.

chrisbra:
thanks, makse sense to include

PR #12929, opened 2023-08-27
After a object/class method call, remove the object/class from the stack

### LINKED ISSUE #12081: vim9class: problem compiling object method call as function call argument
### Steps to reproduce

Source 
```
vim9script

class Elem
endclass

class ElemContainer
    def Current(): Elem
        return Elem.new()
    enddef
endclass

var elems = ElemContainer.new()

def F()
    []->index(elems.Current())
enddef
F()
```

Observe
```
Error detected while compiling command line..script /home/err/bugs/vim/parse-object-method-arg/a.vim
[17]..function <SNR>31_F:
line    1:
E1013: Argument 1: type mismatch, expected list<any> but got object<ElemContainer>
```

### Expected behaviour

No error
Notice that the following works OK.
```
def G()
    var cur = elems.Current()
    echo []->index(cur)
enddef
G()
```

### Version of Vim

9.0.1362

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #12081.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -2747,4 +2747,51 @@ def Test_object_lockvar()
   v9.CheckScriptSuccess(lines)
 enddef
 
+" Test for using the return value of a class/object method as a function
+" argument.
+def Test_objmethod_funcarg()
+  var lines =<< trim END
+    vim9script
+
+    class C
+      def Foo(): string
+        return 'foo'
+      enddef
+    endclass
+
+    def Bar(a: number, s: string): string
+      return s
+    enddef
+
+    def Baz(c: C)
+      assert_equal('foo', Bar(10, c.Foo()))
+    enddef
+
+    var t = C.new()
+    Baz(t)
+  END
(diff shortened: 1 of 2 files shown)


errael:
Those one liners (well, two) can be the worst. The hardest part of this for me will be finding the workarounds; I'll run some tests tomorrow.

mg979:
Maybe it's not needed there but it's similar after https://github.com/vim/vim/blob/78378c0ea0853a8f2ea274cf2b9e6aa1e2c2dd40/src/vim9instr.c#L1927-L1930

yegappan:
>
> Maybe it's not needed there but it's similar after
> 
> https://github.com/vim/vim/blob/78378c0ea0853a8f2ea274cf2b9e6aa1e2c2dd40/src/vim9instr.c#L1927-L1930
>

The `generate_UCALL()` function is used to generate instructions for calling
a function that is not yet defined.  As the function is not yet defined, the number of
arguments and their types are not yet known and the argument type checks cannot 
be performed here.  So the change is not needed here.

PR #12931, opened 2023-08-27
When foldopen contains "search", expand folds the cursor is in

bobrippling:
For example:

```python
if cond:
    f()    #
    g()    # these three lines are folded
    h()    #
```

Displaying something like:

```python
  if cond:
+--  3 lines: f()    #
```

If the cursor is within the fold and the user types `/g`, vim reports no match (the same for subsequent `n` / `N` presses).

With this change, the fold is expanded and the line containing `g()` is visible. This is only done if `'foldopen'` contains `"search"`

---

I'm interested in opinions on this - normally if the cursor is on the only match in a file, `n`/`N` will report no matches, but the user can see that there's a match under the cursor. The same behaviour with folds is a little less useful, as the user can't see that they're on a match within the fold - a few times I've thought a search didn't exist in a file when in fact it was within a fold I was in.

(If the feature sounds good, I'll sort out testing)

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -1220,6 +1220,9 @@ do_search(
     searchit_arg_T  *sia)	// optional arguments or NULL
 {
     pos_T	    pos;	// position of the last match
+#ifdef FEAT_FOLDING
+    linenr_T	    start, end;
+#endif
     char_u	    *searchstr;
     soffset_T	    old_off;
     int		    retval;	// Return value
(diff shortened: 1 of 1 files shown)


chrisbra:
> If the cursor is within the fold and the user types /g, vim reports no match (the same for subsequent n / N presses).

That behaviour is documented [here](https://github.com/vim/vim/blob/d809c0a90387a23aed21ba37d0b65332fb5dafe7/runtime/doc/fold.txt#L602-L605)

bobrippling:
> That behaviour is documented [here](https://github.com/vim/vim/blob/d809c0a90387a23aed21ba37d0b65332fb5dafe7/runtime/doc/fold.txt#L602-L605)

Thanks for the link - I was wondering if this is something open to altering, going off the use-case where the user might assume the search isn't found, when it's actually hidden in a fold?

chrisbra:
I haven't heard any complaints about the current behaviour, so I am tending to keep the current behaviour as is.

bobrippling:
Ok, I appreciate you taking the time to discuss even so

Issue #11827, opened 2023-01-15
Can we allow member variables to be public by default?


zewpo:
https://github.com/vim/vim/blob/ae3205aa55741152b0143dbd7d637b35f576793e/runtime/doc/vim9class.txt#L115-L139

In my opinion - the developer ought to know if setting a member variable has side-effects, so it's their responsibility to make it private.  I think that most of the time, you just want the convenience of setting the variable directly - without writing a setter function for it - or needing to remember to make it public.   So, if the default visibility is public, (explicitly writing "public" becomes optional) then that makes it easier most of the time - and I think that any potential risk that brings is something that the developer ought to be aware of.

vim-ml:
On 23/01/15 2:56 PM, Christopher Plewright (Vim Github Repository) wrote:
> the developer ought to know if setting a member variable has 
> side-effects, so it's their responsibility to make it private. I think 
> that most of the time, you just want the convenience of setting the 
> variable directly - without writing a setter function for it - or 
> needing to remember to make it public.

Unless there's a clear reason, I'll usually be conservative and limit 
the scope; most of the time I don't want to set member variables from 
someplace else. If the need arises for public access it's trivial to 
make public. This is particularly good/convenient for code that gets 
looked at after some time; if there's an issue, it forces me to revisit 
what access rules I want.

-ernie

zewpo:
@vim-ml    Yes, there are some good reasons to default private, I'm aware its preferred by a lot of traditional OO programming languages - and being conservative/safe is good.

Actually, my last suggestion on the vim_dev mailing list would make this a mute point.  There would be no ambiguity about what is the default, because it would enforce all member variables to be explicitly public or private, simply by the prefix "." or "_", which has other advantages too - like removing the requirement of the "this." prefix.

ref:  https://groups.google.com/d/msgid/vim_dev/3f966524-130c-4a81-8a68-fc531883cca5n%40googlegroups.com

Shane-XB-Qian:
the default perhaps should be `protect`, neither `public` nor `private`, just available in class and its sub. IMO.
and so called `this` perhaps actually a bit confused..

// anyway, i felt vim script became a bit complex more than a `script`, do we really need class/abstract/extend/interface...?!

brammool:
Also see the "Class implementation" thread in the vim-dev maillist.
Hopefully this link works: https://groups.google.com/g/vim_dev/c/yYpFNUHdRho

Shane-XB-Qian:
[...]

PR #12932, opened 2023-08-27
Make object/class member variables public by default

yegappan:
To be consistent and simple, make the default access for class/object member variables public.
If a member variable name starts with an underscore, then it is private.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -114,31 +114,10 @@ naming convention makes the object members easy to spot.  Also, when a
 variable does not have the "this." prefix you know it is not an object member.
 
 
-Member write access ~
+Member accessibility ~
 
-Now try to change an object member directly: >
+By default object members are public and can be read and modified directly.
 
-	pos.lnum = 9
-<							*E1335*
-This will give you an error!  That is because by default object members can be
-read but not set.  That's why the TextPosition class provides a method for it: >
-
-	pos.SetLnum(9)
-
-Allowing to read but not set an object member is the most common and safest
-way.  Most often there is no problem using a value, while setting a value may
-have side effects that need to be taken care of.  In this case, the SetLnum()
-method could check if the line number is valid and either give an error or use
-the closest valid value.
-							*:public* *E1331*
-If you don't care about side effects and want to allow the object member to be
-changed at any time, you can make it public: >
-
-	public this.lnum: number
-	public this.col: number
-
-Now you don't need the SetLnum(), SetCol() and SetPosition() methods, setting
-"pos.lnum" directly above will no longer give an error.
 							*E1334*
 If you try to set an object member that doesn't exist you get an error: >
 	pos.other = 9
@@ -204,12 +183,12 @@ including "this.", as the argument to new() means the value provided in the
 new() call is assigned to that object member.  This mechanism comes from the
 Dart language.
 
-Putting together this way of using new() and making the members public results
-in a much shorter class definition than what we started with: >
+Putting together this way of using new() results in a much shorter class
+definition than what we started with: >
 
 	class TextPosition
-	   public this.lnum: number
-	   public this.col: number
+	   this.lnum: number
+	   this.col: number
 
 	   def new(this.lnum, this.col)
 	   enddef
@@ -255,14 +234,13 @@ prefix: >
 Since the name is used as-is, shadowing the name by a function argument name
 or local variable name is not allowed.
 
-Just like object members the access can be made private by using an underscore
-as the first character in the name, and it can be made public by prefixing
-"public": >
+Just like object members the class members are public by default and the
+access can be made private by using an underscore as the first character in
+the name: >
 
     class OtherThing
-	static total: number	      # anybody can read, only class can write
+	static total: number	      # anybody can read and write
 	static _sum: number	      # only class can read and write
-	public static result: number  # anybody can read and write
     endclass
 <
 							*class-function*
@@ -482,7 +460,6 @@ Inside a class, in between `:class` and `:endclass`, these items can appear:
 - An object member declaration: >
 	this._memberName: memberType
 	this.memberName: memberType
-	public this.memberName: memberType
 - A constructor method: >
 	def new(arguments)
 	def newName(arguments)
@@ -864,9 +841,8 @@ you will have to track down and change.  You may have to make it a "set"
 method call.  This reflects the real world problem that taking away access
 requires work to be done for all places where that access exists.
 
-An alternative would have been using the "private" keyword, just like "public"
-changes the access in the other direction.  Well, that's just to reduce the
-number of keywords.
+An alternative would have been using the "private" keyword.  Well, that's just
+to reduce the number of keywords.
 
 
 No protected object members ~
--- a/src/errors.h
+++ b/src/errors.h
@@ -3405,8 +3405,6 @@ EXTERN char e_cannot_get_object_member_type_from_initializer_str[]
 	INIT(= N_("E1329: Cannot get object member type from initializer: %s"));
 EXTERN char e_invalid_type_for_object_member_str[]
 	INIT(= N_("E1330: Invalid type for object member: %s"));
-EXTERN char e_public_must_be_followed_by_this_or_static[]
-	INIT(= N_("E1331: Public must be followed by \"this\" or \"static\""));
 EXTERN char e_public_member_name_cannot_start_with_underscore_str[]
 	INIT(= N_("E1332: Public member name cannot start with underscore: %s"));
 EXTERN char e_cannot_access_private_member_str[]
--- a/src/eval.c
+++ b/src/eval.c
@@ -1573,19 +1573,19 @@ get_lval(
 			    switch (om->ocm_access)
 			    {
 				case VIM_ACCESS_PRIVATE:
-					semsg(_(e_cannot_access_private_member_str),
+				    semsg(_(e_cannot_access_private_member_str),
 								 om->ocm_name);
-					return NULL;
+				    return NULL;
 				case VIM_ACCESS_READ:
-					if ((flags & GLV_READ_ONLY) == 0)
-					{
-					    semsg(_(e_member_is_not_writable_str),
-								 om->ocm_name);
-					    return NULL;
-					}
-					break;
+				    if ((flags & GLV_READ_ONLY) == 0)
+				    {
+					semsg(_(e_member_is_not_writable_str),
+						om->ocm_name);
+					return NULL;
+				    }
+				    break;
 				case VIM_ACCESS_ALL:
-					break;
+				    break;
 			    }
 
 			    lp->ll_valtype = om->ocm_type;
--- a/src/structs.h
+++ b/src/structs.h
@@ -1489,8 +1489,8 @@ typedef struct {
 #define TTFLAG_SUPER	    0x40    // object from "super".
 
 typedef enum {
-    VIM_ACCESS_PRIVATE,	// read/write only inside th class
-    VIM_ACCESS_READ,	// read everywhere, write only inside th class
+    VIM_ACCESS_PRIVATE,	// read/write only inside the class
+    VIM_ACCESS_READ,	// read everywhere, write only inside the class
     VIM_ACCESS_ALL	// read/write everywhere
 } omacc_T;
 
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -719,7 +719,6 @@ def Test_class_object_member_access()
       class Triple
          this._one = 1
          this.two = 2
-         public this.three = 3
 
          def GetOne(): number
            return this._one
@@ -729,26 +728,15 @@ def Test_class_object_member_access()
       var trip = Triple.new()
       assert_equal(1, trip.GetOne())
       assert_equal(2, trip.two)
-      assert_equal(3, trip.three)
       assert_fails('echo trip._one', 'E1333')
 
       assert_fails('trip._one = 11', 'E1333')
-      assert_fails('trip.two = 22', 'E1335')
-      trip.three = 33
-      assert_equal(33, trip.three)
+      trip.two = 22
+      assert_equal(22, trip.two)
 
       assert_fails('trip.four = 4', 'E1334')
   END
   v9.CheckScriptSuccess(lines)
-
-  # Test for a public member variable name beginning with an underscore
-  lines =<< trim END
-    vim9script
-    class A
-      public this._val = 10
-    endclass
-  END
-  v9.CheckScriptFailure(lines, 'E1332:')
 
   lines =<< trim END
       vim9script
@@ -828,23 +816,14 @@ def Test_class_object_member_access()
   END
   v9.CheckScriptSuccess(lines)
 
-  # Test for "public" cannot be abbreviated
+  # Test for using the "public" keyword
   lines =<< trim END
     vim9script
     class Something
-      pub this.val = 1
+      public this.val = 1
     endclass
   END
-  v9.CheckScriptFailure(lines, 'E1065:')
-
-  # Test for "public" keyword must be followed by "this" or "static".
-  lines =<< trim END
-    vim9script
-    class Something
-      public val = 1
-    endclass
-  END
-  v9.CheckScriptFailure(lines, 'E1331:')
+  v9.CheckScriptFailure(lines, 'E1318:')
 
   # Test for "static" cannot be abbreviated
   lines =<< trim END
@@ -983,7 +962,7 @@ def Test_class_member()
          this.col = 1
          static counter = 0
          static _secret = 7
-         public static  anybody = 42
+         static anybody = 42
 
          static def AddToCounter(nr: number)
            counter += nr
@@ -1001,8 +980,8 @@ def Test_class_member()
       assert_equal(3, GetCounter())
 
       assert_fails('TextPos.noSuchMember = 2', 'E1337:')
-      assert_fails('TextPos.counter = 5', 'E1335:')
-      assert_fails('TextPos.counter += 5', 'E1335:')
+      TextPos.counter = 5
+      assert_equal(5, TextPos.counter)
 
       assert_fails('echo TextPos._secret', 'E1333:')
       assert_fails('TextPos._secret = 8', 'E1333:')
@@ -1522,14 +1501,14 @@ def Test_class_implements_interface()
       vim9script
 
       interface Result
-        public this.label: string
+        this.label: string
         this.errpos: number
       endinterface
 
       # order of members is opposite of interface
       class Failure implements Result
         this.errpos: number = 42
-        public this.label: string = 'label'
+        this.label: string = 'label'
       endclass
 
       def Test()
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -31,20 +31,13 @@
     static int
 parse_member(
 	exarg_T	*eap,
-	char_u	*line,
 	char_u	*varname,
-	int	has_public,	    // TRUE if "public" seen before "varname"
 	char_u	**varname_end,
 	garray_T *type_list,
 	type_T	**type_ret,
 	char_u	**init_expr)
 {
     *varname_end = to_name_end(varname, FALSE);
-    if (*varname == '_' && has_public)
-    {
-	semsg(_(e_public_member_name_cannot_start_with_underscore_str), line);
-	return FAIL;
-    }
 
     char_u *colon = skipwhite(*varname_end);
     char_u *type_arg = colon;
@@ -122,7 +115,6 @@ add_member(
 	garray_T    *gap,
 	char_u	    *varname,
 	char_u	    *varname_end,
-	int	    has_public,
 	type_T	    *type,
 	char_u	    *init_expr)
 {
@@ -130,8 +122,7 @@ add_member(
 	return FAIL;
     ocmember_T *m = ((ocmember_T *)gap->ga_data) + gap->ga_len;
     m->ocm_name = vim_strnsave(varname, varname_end - varname);
-    m->ocm_access = has_public ? VIM_ACCESS_ALL
-		      : *varname == '_' ? VIM_ACCESS_PRIVATE : VIM_ACCESS_READ;
+    m->ocm_access = *varname == '_' ? VIM_ACCESS_PRIVATE : VIM_ACCESS_ALL;
     m->ocm_type = type;
     if (init_expr != NULL)
 	m->ocm_init = init_expr;
@@ -1031,24 +1022,6 @@ early_ret:
 	    break;
 	}
 
-	int has_public = FALSE;
-	if (checkforcmd(&p, "public", 3))
-	{
-	    if (STRNCMP(line, "public", 6) != 0)
-	    {
-		semsg(_(e_command_cannot_be_shortened_str), line);
-		break;
-	    }
-	    has_public = TRUE;
-	    p = skipwhite(line + 6);
-
-	    if (STRNCMP(p, "this", 4) != 0 && STRNCMP(p, "static", 6) != 0)
-	    {
-		emsg(_(e_public_must_be_followed_by_this_or_static));
-		break;
-	    }
-	}
-
 	int has_static = FALSE;
 	char_u *ps = p;
 	if (checkforcmd(&p, "static", 4))
@@ -1065,7 +1038,6 @@ early_ret:
 	// object members (public, read access, private):
 	//	"this._varname"
 	//	"this.varname"
-	//	"public this.varname"
 	if (STRNCMP(p, "this", 4) == 0)
 	{
 	    if (p[4] != '.' || !eval_isnamec1(p[5]))
@@ -1077,12 +1049,11 @@ early_ret:
 	    char_u *varname_end = NULL;
 	    type_T *type = NULL;
 	    char_u *init_expr = NULL;
-	    if (parse_member(eap, line, varname, has_public,
-			  &varname_end, &type_list, &type,
-			  is_class ? &init_expr: NULL) == FAIL)
+	    if (parse_member(eap, varname, &varname_end, &type_list, &type,
+					is_class ? &init_expr: NULL) == FAIL)
 		break;
 	    if (add_member(&objmembers, varname, varname_end,
-					  has_public, type, init_expr) == FAIL)
+						type, init_expr) == FAIL)
 	    {
 		vim_free(init_expr);
 		break;
@@ -1174,17 +1145,15 @@ early_ret:
 	    // class members (public, read access, private):
 	    //	"static _varname"
 	    //	"static varname"
-	    //	"public static varname"
 	    char_u *varname = p;
 	    char_u *varname_end = NULL;
 	    type_T *type = NULL;
 	    char_u *init_expr = NULL;
-	    if (parse_member(eap, line, varname, has_public,
-		      &varname_end, &type_list, &type,
-		      is_class ? &init_expr : NULL) == FAIL)
+	    if (parse_member(eap, varname, &varname_end, &type_list, &type,
+					is_class ? &init_expr : NULL) == FAIL)
 		break;
 	    if (add_member(&classmembers, varname, varname_end,
-				      has_public, type, init_expr) == FAIL)
+						type, init_expr) == FAIL)
 	    {
 		vim_free(init_expr);
 		break;


errael:
No way to declare a member read only? It's a shame to sacrifice safety for marginal simplicity. And with a performance hit thrown it; it's almost 4x slower using a Getter(). The main goal of `vim9script` is performance.

```
=== assign micro-prof ===
     50    100    200 : nLoops (usec/op)
  0.044  0.044  0.043 : var local = n_value            ###-A
  0.036  0.035  0.034 : var local = C.static_var       ###-D
  0.050  0.047  0.046 : var local = a_class.this_var   ###-G
  0.173  0.170  0.164 : var local = a_class.Getter()   ###-J
```

yegappan:
>
> No way to declare a member read only? It's a shame to sacrifice safety for
> marginal simplicity. And with a performance hit thrown it; it's almost 4x
> slower using a Getter(). The main goal of `vim9script` is performance.
> 

If we want to keep the object method and member accessibility consistent,
then we need this PR.   Otherwise we can keep the current behavior.

Shane-XB-Qian:
> If we want to keep the object method and member accessibility consistent

maybe adding another keyword 'readonly',
but make the default (if did not specify) then it is 'prevent' (vs not the 'public') means in pkg.
// if so, if we are working all with 'keyword', then i again recommend you use 'private' keyword too (vs not leading underscore)

-- 
shane.xb.qian

Shane-XB-Qian:
simplest way was just basic class which none (get rid) of those keyword, i'm ok with that, it's good enough for scripting.
or the most complex one: not only change control but also access control, adding 'readonly', and all control with keyword.

errael:
> If we want to keep the object method and member accessibility consistent,
> then we need this PR.

I don't understand that. Currently there is `_` for private methods/members; `public` for for public. Nicely consistent. The issue revolves around what does it mean if there is no access modifier?

I thought read only member by default was elegant; it's simple to explain and offers both safety and performance. Here's some ways to look at it

1. Require an access modifier, currently `_` and `public`
    There is no default, and other access schemes, like introducing a `readonly` modifier for member, can be implemented in a source compatible way. 
2. Default is public, get rid of the `public` access modifier, it's not needed. If simplicity is the goal, it's clutter.
    Members/methods are public by default. Can introduce a `readonly` access modifier for current capability.
3. Keep the current.
    Is it too complex/confusing that a method is not read only but by default a member is? members and methods are different; most languages don't have a concept of writing to a method.

jessepav:
I would definitely weigh in in favor of changing the current system, where an unadorned member is read-only, given that it's different from most (all?) other mainstream languages.

`errael`'s second option (default public, `_` private, `readonly` keyword) feels like the cleanest way to retain all current capabilities, though I personally wouldn't mourn the loss of `readonly` either, given that the semantics of vim9script's read-only members is unique as well.

yegappan:
>
> > If we want to keep the object method and member accessibility consistent,
> > then we need this PR.
> 
> I don't understand that. Currently there is `_` for private methods/members;
> `public` for for public. Nicely consistent.
>

The `public` keyword is used only for object/class members and not for methods.
The methods are public by default and private if the name starts with an underscore.

>
> The issue revolves around what does it mean if there is no access modifier?
>

Yes.

> 
> I thought read only member by default was elegant; it's simple to explain and
> offers both safety and performance.
>

With the latest code, you will get this behavior (without this PR).  If this is the
desired behavior, then we don't need to make any further changes.

>
> Here's some ways to look at it
> 
> 1. Require an access modifier, currently `_` and `public`
>    There is no default, and other access schemes, like introducing a
>    `readonly` modifier for member, can be implemented in a source compatible
>    way.
> 2. Default is public, get rid of the `public` access modifier, it's not
>    needed. If simplicity is the goal, it's clutter.  Members/methods are
>    public by default. Can introduce a `readonly` access modifier for current
>    capability.
>

This PR introduces the behavior in 2 except for the `readonly` keyword.

>
> 3. Keep the current.
>    Is it too complex/confusing that a method is not read only but by default
>    a member is? members and methods are different; most languages don't have
>    a concept of writing to a method.
>

For 3, as described above, we don't need to make any further changes.

If we all agree approach 3 is acceptable (even though it is not consistent between
methods and members), then we don't need any additional changes.

Regards,
Yegappan

errael:
> > 2. Default is public, get rid of the `public` access modifier, it's not
> >    needed. If simplicity is the goal, it's clutter.  Members/methods are
> >    public by default. Can introduce a `readonly` access modifier for current
> >    capability.
> 
> This PR introduces the behavior in 2 except for the `readonly` keyword.

The PR also takes out the internal support for readonly, rather than just at the parsing/specification level.

I was looking at what it would take to introduce a `readonly` keyword, before asking how adding that would be received.  If this PR left the infrastructure for readonly, it would be a lot simpler to provide that capability.

mg979:
+1 for `readonly` and making members public by default (removing `public` altogether).

I don't think it's needed to make it an exclusive keyword, it can just be checked inside classes before member declaration, as other 'keywords' currently work (`extends`, `implements`, etc.).

errael:
>  Retain some parts of the read only member access code

Thanks. Assuming you're not planning to do read only, I'll take a look tomorrow. (What other plans?).

Idea is basically put back `public` using `readonly`; has_readonly instead of has_public. Tweak as needed and flavor with tests.

errael:
Oh, and `readonly` is the only one I've heard floated; haven't thought of anything else that's _perfect_. I could live with `read` or `ronly`. Once everything else is in place, it's simple to change the word.

dkearns:
`readable`...paint it green!

errael:
When I build objmember, with no modifications, and run this I see two problems

1. Able to access private static
2. Compile error when trying to write public static

Here's a test
```
vim9script
class C
    this._mpfoo: number
    static _spfoo: number
    this.mfoo: number
    static sfoo: number

    def new()
        this._mpfoo = 1
        _spfoo = 2
        this.mfoo = 201
        sfoo = 202
    enddef
endclass

def F()
    var o = C.new()
    #echo "_mpfoo" o._mpfoo
    echo "_spfoo" C._spfoo  ### <<<<<<<<< able to access private static
    echo "mfoo" o.mfoo
    echo "sfoo" C.sfoo

    o.mfoo = 301
    #C.sfoo = 302       ### <<<<<<<<< compilation fails
    echo "mfoo" o.mfoo
    echo "sfoo" C.sfoo
enddef
F()
finish
```

errael:
Oh, and this is what I'm running
```
 hg glog -r ::objmember -l 2
@  changeset:   18285:bdb16ee51e4c
|  bookmark:    objmember
|  tag:         default/objmember
|  tag:         tip
|  parent:      18253:025bc462c74c
|  user:        Yegappan Lakshmanan <yegappan@yahoo.com>
|  date:        Sun Aug 27 07:59:56 2023 -0700
|  summary:     Make object/class member variables public by default
|
o  changeset:   18253:025bc462c74c
|  user:        zeertzjq <zeertzjq@outlook.com>
~  date:        Sun Aug 27 11:17:39 2023 +0200
   summary:     patch 9.0.1792: problem with gj/gk/gM and virtual text
```

errael:
Has readonly ever worked? I just ran some tests with vim I built yesterday, standard vim9 stuff. Happily writes to read only variable.

errael:
I looked around to get a feel... The following fixes the "access to private static" var in the example, if you want to include it. Not tested much.

```
diff --git a/src/vim9expr.c b/src/vim9expr.c
--- a/src/vim9expr.c
+++ b/src/vim9expr.c
@@ -434,14 +434,25 @@
     {
        // load class member
        int idx;
+       ocmember_T *m;
        for (idx = 0; idx < cl->class_class_member_count; ++idx)
        {
-           ocmember_T *m = &cl->class_class_members[idx];
+           m = &cl->class_class_members[idx];
            if (STRNCMP(name, m->ocm_name, len) == 0 && m->ocm_name[len] == NUL)
                break;
        }
        if (idx < cl->class_class_member_count)
        {
+           // TODO: Would kindof prefer (@yegappan what do you think)
+           //       even though it's more likely to get a cache miss.
+           // if (m->ocm_access == VIM_ACCESS_PRIVATE
+           //                                      && !inside_class(cctx, cl))
+           // Maybe not, *name == '_' is simple to look at but...
+           if (*name == '_' && !inside_class(cctx, cl))
+           {
+               semsg(_(e_cannot_access_private_member_str), m->ocm_name);
+               return FAIL;
+           }
            *arg = name_end;
            return generate_CLASSMEMBER(cctx, TRUE, cl, idx);
        }
```

vim-ml:
Hi,

On Mon, Aug 28, 2023 at 12:14 PM errael ***@***.***> wrote:

> When I build objmember, with no modifications, and run this I see two
> problems
>
>    1. Able to access private static
>    2. Compile error when trying to write public static
>
>
I am able to reproduce the second issue even using a two months old image.
It looks like writing to a class member from a def function never worked.

Regards,
Yegappan


>
>
> Here's a test
>
> vim9script
> class C
>     this._mpfoo: number
>     static _spfoo: number
>     this.mfoo: number
>     static sfoo: number
>
>     def new()
>         this._mpfoo = 1
>         _spfoo = 2
>         this.mfoo = 201
>         sfoo = 202
>     enddef
> endclass
>
> def F()
>     var o = C.new()
>     #echo "_mpfoo" o._mpfoo
>     echo "_spfoo" C._spfoo  ### <<<<<<<<< able to access private static
>     echo "mfoo" o.mfoo
>     echo "sfoo" C.sfoo
>
>     o.mfoo = 301
>     #C.sfoo = 302       ### <<<<<<<<< compilation fails
>     echo "mfoo" o.mfoo
>     echo "sfoo" C.sfoo
> enddef
> F()
> finish
>
>
>

vim-ml:
Hi,

On Mon, Aug 28, 2023 at 6:47 PM errael ***@***.***> wrote:

> I looked around to get a feel... The following fixes the "access to
> private static" var in the example, if you want to include it. Not tested
> much.
>
> diff --git a/src/vim9expr.c b/src/vim9expr.c
> --- a/src/vim9expr.c
> +++ b/src/vim9expr.c
> @@ -434,14 +434,25 @@
>      {
>         // load class member
>         int idx;
> +       ocmember_T *m;
>         for (idx = 0; idx < cl->class_class_member_count; ++idx)
>         {
> -           ocmember_T *m = &cl->class_class_members[idx];
> +           m = &cl->class_class_members[idx];
>             if (STRNCMP(name, m->ocm_name, len) == 0 && m->ocm_name[len] == NUL)
>                 break;
>         }
>         if (idx < cl->class_class_member_count)
>         {
> +           // TODO: Would kindof prefer ***@***.*** what do you think)
> +           //       even though it's more likely to get a cache miss.
> +           // if (m->ocm_access == VIM_ACCESS_PRIVATE
> +           //                                      && !inside_class(cctx, cl))
> +           // Maybe not, *name == '_' is simple to look at but...
> +           if (*name == '_' && !inside_class(cctx, cl))
> +           {
> +               semsg(_(e_cannot_access_private_member_str), m->ocm_name);
> +               return FAIL;
> +           }
>             *arg = name_end;
>             return generate_CLASSMEMBER(cctx, TRUE, cl, idx);
>         }
>
>
>
This is the proper fix for this issue.  A similar check is used before this
block of code for
readonly object members.  Can you open a PR for this with a test for this
condition?

- Yegappan

errael:
> Can you open a PR for this with a test for this condition?

OK, but not tonight, tomorrow. I haven't looked at the "write public static compile error".

For my current work area, I took your objmember PR, transplant to work area, rebased to current master, added the readonly parse, I'll need to unwind and stash. I just saw `compile_assign_unlet` I believe that's the spot for the readonly check. The lhs_T looks like is has all the info needed for the check VIM_ACCESS_READ before any instructions are generated (3 for storing into the object when object is in local var).

> A similar check is used before this block of code for
> readonly object members

Assuming you meant private.

Shane-XB-Qian:
upd & link & refer : https://github.com/vim/vim/issues/11827

errael:
`readonly` is working AFAICT. I still need to put together some tests. I can't try writing to a readonly static because of the compilation failure bug that prevents writing to class statics. One nice thing about `readonly` is that it doesn't affect anything unless you use it.

 I'm going to put this on the shelf while I put together the PR for preventing writing to private members.

The `readonly` PR is based on this PR. So this PR, #12932, has to go in before I can open a PR for `readonly`. Is this PR ready to go?

dkearns:
Is there some strong objection to _requiring_ a keyword modifier?  Why is a default needed at all?

We could require `private` and keep the required underscore as well.  This is is already done for functions/methods which require an uppercase character.

The language uses strictly unnecessary keywords everywhere.  In declarations, in particular, they're cheap and clear.

This would also obviate the problem with initialised declarations looking the same as plain assignments.

mg979:
@errael before asking for this PR to be merged, can you show your code with the `readonly` implementation? Because I was also working on an implementation but it's incompatible with this PR. So I'd like to discuss how we should implement `readonly` first.

My take is [here](https://github.com/mg979/vim/commit/f821ee4eb732d760b732838f68caad69146906fa) (WIP, not working).

Basically I wanted `readonly` to work as in C#:

- both private and public members can be readonly
- readonly means the member can still be changed (set) by the constructor, but not later
- readonly is only for object members (`this.`), not for class members (`static`) because the constructor can change it

So it's different from `const`, that could be actually implemented to have readonly class (static) members.

errael:
The implementation I'm working is the current read only semantics that you get by default and is described in the current documentation; see ":he vim9-class-member", it says "anybody can read, only class can write", no other rules. Note that even this has been considered by some as too complex. What you describe seems to go beyond the goal of simplicity.

The change I'm working on is about access control,  and does not affect object constructin. It affects both members/statics.

There's a comment from Bram somewhere that a problem with using the `readonly` keyword is that it meant something else in another language (I'm guessing C#). Maybe `readable` would be a better keyword to avoid possible confusion. Fortunately changing the keyword is straightforward.

It seems like what you are working on is orthogonal to what I'm doing. It is a different mechanism entirely, which could be added at some point.

mg979:
The problem with the semantics Bram came up with, is that they're different from any other language (the use of `public` for example). The semantics for `readonly` you are proposing follow this line of thought. `readonly` should mean the member can't be written by anybody, otherwise it's not readonly. Just like public should mean publicly accessible, not publicly writable.

So I'm against your proposal. I think it doesn't matter what the current specification says, because mostly it doesn't work (do abstract classes and interfaces actually work?). The class specification should be considered a WIP all around so.

Rather than confusing simplicity, familiarity and consistence of semantics (on the base of what they already mean in other languages) is more important here. Then you say it's more complex but I don't even think it is.

Let's hear what @yegappan says.

mg979:
The same concept of `anybody can read, only class can write`, I don't think it exists in any other language. So it's not _simple_.

errael:
> I think it doesn't matter what the current specification says, because mostly it doesn't work

I have written a project using the current readonly semantics and I like them quite a bit. In particular I have a presentation class that presents a readonly view of some internal structures. This is easy to implement allowing concise code and doesn't require the much lower performance getters.

> readonly should mean ...

OK, I guess the access modifier `readable` is the way to go.

mg979:
Also `readable` doesn't exist anywhere else. If simplicity is the ultimate goal, I'd say this PR is ok in just removing `public`. We can live without `readonly` but even more so without stuff invented for vimscript that doesn't exist anywhere else.

chrisbra:
so let me summarize: the current proposal is to remove the `public` keyword completely (?) and make all attributes public by default (unless they start with `_`)? So by removing this keyword, no-one complains that `private` keyword is also missing?

I like that approach, it makes it easier to spot private members directly in the code. But I am probably not the one going to use this a lot, so I am not having any saying here :)

vim-ml:
Hi Christian,

On Tue, Aug 29, 2023 at 12:06 PM Christian Brabandt <
***@***.***> wrote:

> so let me summarize: the current proposal is to remove the public keyword
> completely (?) and make all attributes public by default (unless they start
> with _)? So by removing this keyword, no-one complains that private
> keyword is also missing?
>

Yes.  The only supported keyword for members/methods is "static" for
declaring class members and methods.
If a member or method name starts with an underscore, then it is private.
Otherwise it is public.

Regards,
Yegappan

I like that approach, it makes it easier to spot private members directly
> in the code. But I am probably not the one going to use this a lot, so I am
> not having any saying here :)
>
>
>

vim-ml:
Hi,

On Tue, Aug 29, 2023 at 11:03 AM Gianmaria Bajo ***@***.***>
wrote:

> Also readable doesn't exist anywhere else. If simplicity is the ultimate
> goal, I'd say this PR is ok in just removing public. We can live without
> readonly but even more so without stuff invented for vimscript that
> doesn't exist anywhere else.
>
>
> I agree.  I think this will be simple and less confusing.

Regards,
Yegappan

mg979:
Just to mention another alternative:
```
this.member     # full access from everywhere
this._member    # semi-private: read-only from outside the class
this.__member   # fully private
```

errael:
> * Able to access private static
> 
> * Compile error when trying to write public static

There's another one (a real shocker).

It's OK to write to a private member.
```
vim9script

class C
    this._m: number
    #static _s: number
    def Dump()
        echo this._m
    enddef
endclass

def F()
    var o = C.new()
    o.Dump()
    o._m = 5
    o.Dump()
    #var s = C._s
enddef
F()
```
I put in the `Dump()` because I had to see it to believe it.

chrisbra:
Should the example here https://github.com/vim/vim/blob/d01a0df61d2a322645fb393ed704fcec38e0d0fe/runtime/doc/vim9class.txt#L186-L200

also be changed to use underscores? Otherwise we wouldn't need the and Set() function, right?

mg979:
> There's another one (a real shocker).
> 
> It's OK to write to a private member.

Confirmed, but only inside a `def`, it works at script level.

errael:
> Can you open a PR for this with a test for this condition?

I got everything wrapped up as a follow up to this PR, and it doesn't look like that will work out. So no.

vim-ml:
Hi,

On Tue, Aug 29, 2023 at 1:42 PM Christian Brabandt <
***@***.***> wrote:

> Should the example here
> https://github.com/vim/vim/blob/d01a0df61d2a322645fb393ed704fcec38e0d0fe/runtime/doc/vim9class.txt#L186-L200
>
> also be changed to use underscores? Otherwise we wouldn't need the and
> Set() function, right?
>
>
>
Yes. This example needs to be updated.

Regards,
Yegappan

vim-ml:
Hi,

On Tue, Aug 29, 2023 at 1:38 PM errael ***@***.***> wrote:

>
>    -
>
>    Able to access private static
>    -
>
>    Compile error when trying to write public static
>
> There's another one (a real shocker).
>
> It's OK to write to a private member.
>
> vim9script
>
> class C
>     this._m: number
>     #static _s: number
>     def Dump()
>         echo this._m
>     enddef
> endclass
>
> def F()
>     var o = C.new()
>     o.Dump()
>     o._m = 5
>     o.Dump()
>     #var s = C._s
> enddef
> F()
>
> I put in the Dump() because I had to see it to believe it.
>
>
>
I have created the PR https://github.com/vim/vim/pull/12963 to address this.

Regards,
Yegappan

Shane-XB-Qian:
> Just to mention another alternative:
> 
> ```
> this.member     # full access from everywhere
> this._member    # semi-private: read-only from outside the class
> this.__member   # fully private
> ```

vim9 is for scripting mostly, i think better be making it simple.
// @yegappan double (or more) leading `__` perhaps should be a `err`.

> Yes. The only supported keyword for members/methods is "static" for declaring class members and methods. If a member or method name starts with an underscore, then it is private. Otherwise it is public.

ok, i think we can live with it (at least me), and actually this is the better one, leading underscore is acceptable if so.
// though i think should be not much people really would use vim9 to write OOP script.

dkearns:
> > Is there some strong objection to _requiring_ a keyword modifier? Why is a default needed at all?
> 
> Are you suggesting `private`/`public` for functions, members, static.

Yes, all class features where the modifier is applicable.  I was, however, assuming there'd be more than the two; there will be one day. :)

dkearns:
> so let me summarize: the current proposal is to remove the `public` keyword completely (?) and make all attributes public by default (unless they start with `_`)? So by removing this keyword, no-one complains that `private` keyword is also missing?
> 
> I like that approach, it makes it easier to spot private members directly in the code. But I am probably not the one going to use this a lot, so I am not having any saying here :)

For the record, I have been complaining for a year about some of the quirky class definition syntax.

Limiting this to private underscore for the moment and since you have no dog in the fight, maybe you could take the hit and elaborate on why "spot private members in the code" is of such obvious value that we need a new identifier naming rule?  My blind spot is huge.

It seems to me to be a completely arbitrary partial Hungarian notation with semantics.  There are several attributes I'm interested in at the call site, accessibility is only one of them and not the most interesting.

In my experience the underscore idiom was developed in languages that didn't enforce accessibility and doesn't make much sense in those that do.  I'm assuming this feels natural to people who already use the idiom in other languages.  I find it extremely ugly in a language that I otherwise find consistent and quite pretty in its Vim 9 release state.

When you lump this in with required 'this' qualification one has to prefix every private field with six characters of extremely dubious value.  A few variables on the same line and I'm already cobbling together a syn-conceal group to ease the pain.

This is largely bike-shedding, of course, but isn't that what surface level language design is all about? :)

errael:
> why "spot private members in the code" is of such obvious value

I agree with most everything you've said (and although I've used (and like quite a bit) python, I've never used it in a larger project where `_` was used.

But now that my beloved `readable` is in the trash, I will rarely have an unadorned field name; everything is private. I'd find it more useful to visually flag something public so the warning bells go off in my head.

chrisbra:
For the record, I am not insisting on any of this and the reason I have not yet merged is, I'd like to see what the dicussion results in :) 

But since you seem to be asking me directly 😎

Code tends to be read more often then written and for me, being able to see while reading existing code, `ah this must be a private member` outweights the disadvantage that the `_` brings as part of the changes in semantics of the language. (I think Bram brought this argument as well at some time). Yes you need to know that `_foobar` has a special meaning, but if you are writing code, you should know where to lookup the specifics of your language anyhow. 

there is another disadvantage, that one cannot easily refactor code, but I suppose this is not done so often. So for me this is okay. 

and finally whether one uses `this._member` or `this.member` doesn't make such a huge difference for me. 

But then I am not a developer in my profession and I don't know what other languages tend to do, so my opinion is as good or bad as anybody elses here :)

errael:
> I'd like to see what the dicussion results in :)

In that case

I'm disappointed that there was no response to the observations/concerns with the performance hit of not having direct read access. Of the possible responses, e.g. "not that important", no response is the worst.

In the discussion on simplicity, there wasn't much discussion of the visual clutter introduced by requiring getters. Recall that the vim9class doc carefully shows how simple a class can look for common cases; vim9script _is_ a scripting language. Safety was traded off in favor of not having `public` and adding visual clutter. And one of the "key" discussion points was that anything new or innovative could not be simple, I don't accept that.

chrisbra:
> I'm disappointed that there was no response to the observations/concerns with the performance hit of not having direct read access. Of the possible responses, e.g. "not that important", no response is the worst.

I don't follow. By default we have public access, so that shouldn't have any negative performance impact.

> In the discussion on simplicity, there wasn't much discussion of the visual clutter introduced by requiring getters. Recall that the vim9class doc carefully shows how simple a class can look for common cases; vim9script is a scripting language. Safety was traded off in favor of not having public and adding visual clutter. And one of the "key" discussion points was that anything new or innovative could not be simple, I don't accept that.

You need Getters only for private members, right? For common cases using public members you wouldn't need it, right?

errael:
> > I'm disappointed that there was no response to the observations/concerns with the performance hit of not having direct read access. Of the possible responses, e.g. "not that important", no response is the worst.
> 
> I don't follow. By default we have public access, so that shouldn't have any negative performance impact.

I don't use public, it is unsafe. The performance hit is that Getters are required.
https://github.com/vim/vim/pull/12932#issuecomment-1694703159

> > In the discussion on simplicity, there wasn't much discussion of the visual clutter introduced by requiring getters. Recall that the vim9class doc carefully shows how simple a class can look for common cases; vim9script is a scripting language. Safety was traded off in favor of not having public and adding visual clutter. And one of the "key" discussion points was that anything new or innovative could not be simple, I don't accept that.
> 
> You need Getters only for private members, right? For common cases using public members you wouldn't need it, right?

Public members are OK for tiny things, but public members are not safe. I think you are mistaken; for a stongly typed OOP, public members are **not the common case**. As a project grows, allowing arbitrary actors to modify internal state is not a common design pattern. Make everything private and add Getter() is what one needs to do with the proposed semantics.

mg979:
What @errael wants, is members that can be publicly read but not written, so that you don't need to set up Getters for them (with the associated performance hit). I think it's a valid case after all.

Other OOP languages have ways to define this behaviour (for example with properties) but we only want simple stuff here, certainly not properties, but it would be convenient to have here. I still think `readable` is a bad name, as is `readonly` for this. Maybe `restrict` or similar.

Shane-XB-Qian:
even 'public' it actually was default in pkg only as my understanding, so I was said actually it was 'prevent' was the correct default.
I supposed this can reduce your safety concern, though I still think not much people really would use vim9 to wrote OOP script.

--
shane.xb.qian

errael:
Found it, here's some comments from Bram about this topic

In the email thread
https://groups.google.com/g/vim_dev/c/yYpFNUHdRho/m/xjgrKqMoBQAJ?pli=1

The only comment I saw from Bram in the thread specifically about default public versus read-only has to do with simple versus safe. From the thread, in response to a suggestion that public should be the default:

    About public/private/read-only: Making the default public is certainly
    possible. I think most languages have it that way, even though it's not
    a safe choice.

    I do like the read-only access. Quite often you want to disallow any
    code outside of the class to change a member, so that you can make sure
    the value is valid. But making it private means you have add a
    "getter" and call it. [...]

errael:
> can reduce your safety concern, though I still think not much people really would use vim9 to wrote OOP

But we are specifically talking about OOP. Someone using it should be assumed when considering a comment.

errael:
> Other OOP languages have ways to define this behaviour...

> I still think readable is a bad name, as is readonly for this. Maybe restrict or similar.

The discussion should be about whether or not the semantics should be available in vim9script.

I think it best to not discuss specific names, that derailed previous discussions.

Shane-XB-Qian:
> > can reduce your safety concern, though I still think not much people really would use vim9 to wrote OOP

> But we are specifically talking about OOP. Someone using it should be assumed when considering a comment.

I mean you can have a bit confident to use that `public`, that default actually was pkg only, not much safety problem in vim9 scripting, even really someone to write huge project in vim9 OOP.

if so, not much getter() required.

--
shane.xb.qian

errael:
@yegappan  I appreciate that in your position you want to observe and let the discussion flow. But I would appreciate if you can find a way to express your opinions on the various topics. Not just state the facts.

How useful are read-only semantics? How big are projects going to get (I've never written vimscript or looked into the ecosystem, but I've seen projects grow and grow no matter the language). How important is safety (it is strongly typed...) and how do the available syntax/semantics affect that. What do you think about the performance hit. Anything else?

mg979:
Currently `public` isn't even needed by the way. I learned of it with this PR.

```
vim9script

class C
    this.str: string
endclass

def F()
    var tvar = C.new()
    echo tvar.str
    tvar.str = "a"
    echo tvar.str
enddef

F()
```

yegappan:
> 
> How useful are read-only semantics? How big are projects going to get (I've
> never written vimscript or looked into the ecosystem, but I've seen projects
> grow and grow no matter the language). How important is safety (it is
> strongly typed...) and how do the available syntax/semantics affect that.
> What do you think about the performance hit. Anything else?
>

I developed and maintain the following large and widely used Vim plugins:

https://github.com/yegappan/lsp
https://github.com/yegappan/taglist
https://github.com/yegappan/mru

Based on this experience, I can say that type safety and access control
are important.  I have spent quite a bit of time optimizing these plugins
and found that reducing functions calls definitely helps.  So having a
readonly access to the members will help.

yegappan:
>
> Currently `public` isn't even needed by the way. I learned of it with this PR.
>

This is a bug.  I will open a PR to address this.

Shane-XB-Qian:
> have spent quite a bit of time optimizing these plugins and found that reducing functions calls definitely helps. So having a readonly access to the members will help.

I am there with you too specially that lsp plugin :smile:

as my limit experience, readonly normally was not part of general OOP, but you may make `const` and `final` work on class var.
with the here `public` default and `const` or `final` then concern solved.
// anyway I supposed we were choosing the simplest way (vs that most complex) which I indicated two ways in above comment.

--
shane.xb.qian

Shane-XB-Qian:
> so I guess we should then leave it as it?

1. merger this pr
2. make `const` and `final` work on class var.
// then concern solved.

// P.S: double or more leading underscore may  should be an err.

--
shane.xb.qian

errael:
> > Currently `public` isn't even needed by the way. I learned of it with this PR.
> 
> This is a bug. I will open a PR to address this.

@yegappan If we get this PR, #12932, settled along with what semantics for access will be supported, and a list of outstanding access control features that are problems and need fixing. Then I can assist; I've got some work done, but stopped when it became unclear what was needed.

Not sure the best way to enumerate the issues and at indicate who is working on what.

errael:
> so I guess we should then leave it as it?
> 1. merger this pr 2. make `const` and `final` work on class var. // then concern solved. // P.S: double or more leading underscore may should be an err.
> […](#)
> -- shane.xb.qian

@Shane-XB-Qian You might have a misunderstanding of the read-only semantics under discussion. In the vim9class docs it says
```
anybody can read, only class can write
```
The traditional `const`/`final` do not meet this criteria.

errael:
> okay, so I guess we should then leave it as it? Seems like that is prefered from the community point of view instead...

@chrisbra Can you give a timeline on when this will be finalized. I'm unable/unwilling to work on issues until it's clear what needs to be done.

errael:
> > Currently `public` isn't even needed by the way. I learned of it with this PR.
> 
> This is a bug. I will open a PR to address this.

And specifically, I've got a fix that handles ACCESS_PRIVATE/ACCESS_READ in `compile_assign_unlet`. Which I think covers this particular issue.

chrisbra:
it's not clear to me what you want. 

But to be clear, there is no rush, if you want something fixed, then just open an issue/PR and mention you are working on this. I am not in a hurry to release anything. We can tag this with the vim 9.1 Milestone and only when this is done get the release done. 

For the Vim9 oop semantics, I trust you and Yegappan to let me know once you thing this is in a state that is ready to be released :) 

And of course anybody else

Shane-XB-Qian:
________________________________

errael:
> it's not clear to me what you want.

The first thing is a decision on what access semantics are going to be supported in vim9script for the 9.1 release. That is still an open issue AFAICT.

The other problem is that access control was mostly not implemented at all as of last week. There's not been an issue opened. @yegappan and others have been addressing them as they come up; they have mostly been mentioned in unrelated PRs. There are/have-been to many specific cases to open an issue for each one (at least that's how it has been); but then again, I'm only saying what I've seen, accuracy might be off.

Maybe a discussion, with the access matrix, what's been tested, what works, ..., If someone wants to address a specific case, they could say so.

Shane-XB-Qian:
On 23/08/30 10:34AM, errael wrote:
> > so I guess we should then leave it as it?
> > 1. merger this pr 2. make `const` and `final` work on class var. // then concern solved. // P.S: double or more leading underscore may should be an err.
> > […](#)
> > -- shane.xb.qian
> 
> @Shane-XB-Qian You might have a misunderstanding of the read-only semantics under discussion. In the vim9class docs it says
> ```
> anybody can read, only class can write
> ```
> The traditional `const`/`final` do not meet this criteria.
> 

maybe let's focus on your concern : 'safety' and 'getter()'
i donot understand that doc (your pasted) about 'read-only' what's that going to be (any OOP like that?)
or anyway, this part bram should be actually not impl yet, meant doc was just a rough design (in vim9 doc somewhere had said that too),
and `const` even different in legacy vim script and vim9, so supposed ok if a bit different in vim9 class var too (if really required).

-- 
shane.xb.qian

yegappan:
> > > Currently `public` isn't even needed by the way. I learned of it with this PR.
> > 
> > This is a bug. I will open a PR to address this.
> 
> And specifically, I've got a fix that handles ACCESS_PRIVATE/ACCESS_READ 
> in `compile_assign_unlet`. Which I think covers this particular issue.
>

Please feel free to open a PR with this fix and a test.

errael:
> > > > Currently `public` isn't even needed by the way. I learned of it with this PR.
> > > 
> > > 
> > > This is a bug. I will open a PR to address this.
> > 
> > 
> > And specifically, I've got a fix that handles ACCESS_PRIVATE/ACCESS_READ
> > in `compile_assign_unlet`. Which I think covers this particular issue.
> 
> Please feel free to open a PR with this fix and a test.

Until there is a decision on access control syntax/semantics that's not going to happen (at least by me). So as long as you're not in a hurry... I'll do it when there's some decisions, otherwise it's potentially throw away code.

How should co-ordination, specifically about access control, be handled? It's currently a bit of a hodge-podge.

Shane-XB-Qian:
i am ok with current had made decision which is:
 only `static` keyword, default is `public` and remove that keyword, and one leading underscore is private.

only left thing is your concern about 'readonly', if above explaination or suggestion cannot solve it, then i have no idea... 
// then same like chrisbra, i left this to you guys to find a better or workable one.. :smile:

-- 
shane.xb.qian

chrisbra:
> Maybe a discussion, with the access matrix, what's been tested, what works, ..., If someone wants to address a specific case, they could say so.

Feel free to open as discussion. I am not in the position to test this currently.

errael:
> > Maybe a discussion, with the access matrix, what's been tested, what works, ..., If someone wants to address a specific case, they could say so.
> 
> Feel free to open as discussion. I am not in the position to test this currently.

@yegappan You're the vim9script lead in this _benign dictatorship_. Whatever, if anything, we do must have your blessing or it's not worth doing.

@chrisbra Is there a timeline for when a decision will be made? Nothing to do or work on if there's no target.

errael:
Just in case there is a doubt. I would favor closing this PR _without merging_ as described by @yegappan in the comment below extracted from https://github.com/vim/vim/pull/12932#issuecomment-1699493034

To anyone who wishes to make clear their preference, keeping in mind that this is not a vote. Please be brief. There has been plenty of discussion.

> to continue to support the existing readonly default access,
> [...]
> We then don't need the changes in this PR and can keep
> the existing behavior.

Shane-XB-Qian:
> keeping in mind that this is not a vote

this is a some kinds of 'vote' and discussion, otherwise where was this PR came from?
anyway, seems @yeggapan changed mind again which this was discarded vs #12978 opened.
// 'public' & 'readonly' & '_' looks odd, but if that's final version, then let it be~

-- 
shane.xb.qian

yegappan:
> > > Maybe a discussion, with the access matrix, what's been tested, what works, ..., If someone wants to address a specific case, they could say so.
> > 
> > 
> > Feel free to open as discussion. I am not in the position to test this currently.
> 
> @yegappan You're the vim9script lead in this _benign dictatorship_. Whatever, if anything, we do must have your blessing or it's not worth doing.
> 

I have started a discussion (https://github.com/vim/vim/discussions/12979) to capture the
access matrix.  I have tested the various access controls for an object method and the
different ways it can be accessed.  All of these works as expected.  Let me know if I missed
any cases.

Accessing a class member variable from a def function currently doesn't work.

vim-ml:
On Thu, Aug 31, 2023 at 6:50 AM errael ***@***.***> wrote:

> I have started a discussion (#12979
> <https://github.com/vim/vim/discussions/12979>) to capture the
> access matrix.
>
> Thank you. I'll take a look. I would have done something if you'd said "go
> ahead" .
>
>
> I have marked the failing cases with "BUG".  Most of these are related to
writing to
a class member variable in a def function.

PR #12945, opened 2023-08-28
vim9class: no static constructors

mg979:
Cannot declare constructors as "static".

Constructors are always static, using "static" keyword is an error.

### DIFF
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -1188,15 +1188,16 @@ early_ret:
 	    {
 		char_u *name = uf->uf_name;
 		int is_new = STRNCMP(name, "new", 3) == 0;
-		if (is_new && is_abstract)
-		{
-		    emsg(_(e_cannot_define_new_function_in_abstract_class));
-		    success = FALSE;
-		    func_clear_free(uf, FALSE);
-		    break;
-		}
 		if (is_new)
 		{
+		    // Constructors are not allowed in abstract classes.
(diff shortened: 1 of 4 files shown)


src/vim9class.c
```diff
@@ -1188,15 +1188,16 @@ ex_class(exarg_T *eap)
...
-		    break;
-		}
 		if (is_new)
 		{
+		    // Constructors are not allowed in abstract classes.
+		    if (is_abstract)
+		    {
+			emsg(_(e_cannot_define_new_function_in_abstract_class));
+			success = FALSE;
+			func_clear_free(uf, FALSE);
```
mg979: @yegappan 

`func_clear_free()` isn't called in other places where errors occur and flow is broken. Are you sure it must be this way?

Edit: ok I see now, but `success = FALSE` shouldn't be needed there, it can't be `TRUE` at that point because the only point where it's set to TRUE is at https://github.com/vim/vim/blob/8952dc2f3d8eb066a09f33c8fa620a9d02ec36ff/src/vim9class.c#L1070-L1079 and there it exits from the loop.

mg979:
> but in my experience a constructor is not declared static

That's what this PR is about.

errael:
> > but in my experience a constructor is not declared static
> 
> That's what this PR is about.

Oops, I had it backwards.

yegappan (approved):

PR #12949, opened 2023-08-29
src/configure.ac: Remove AC_HEADER_STDC

illia-bobyr:
Autoconf 2.71 says AC_HEADER_STDC is obsolete and that

> All hosted environments that are still of interest for portable code
> provide all of the headers specified in ISO C90

autoupdate from autoconf 2.71 suggests removal of this macro altogether.

---

This is the second step of an upgrade to autoconf 2.71.
Second of 3 parts that combined are mostly the same as changes in here: https://github.com/vim/vim/pull/11322

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -1723,48 +1723,6 @@ fi
 
 } # ac_fn_c_try_cpp
 
-# ac_fn_c_try_run LINENO
-# ----------------------
-# Try to link conftest.$ac_ext, and return whether this succeeded. Assumes
-# that executables *can* be run.
-ac_fn_c_try_run ()
-{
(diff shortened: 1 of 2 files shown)


chrisbra:
so after this is included, we also need to include #11322 ?

ilya-bobyr:
> so after this is included, we also need to include #11322 ?

I have another change pending that will switch to 2.71.
All 3 combined (https://github.com/vim/vim/pull/12888, this, and the upcoming) do pretty much the same as #11322.

#11322 has a few nice fixes for the todo list and the Makefile.
I was planning to include those in the third change, making the combination identical to  #11322.

chrisbra:
Hm, so in addition to removing the `AC_HEADER_STDC` check, you also moved some of the test around. Was that done intentionally?

PR #12954, opened 2023-08-29
Fix sed command for luajit version detection in configure script

ichizok:
* Make compatible with BSD sed; it doesn't accept `\?`


chrisbra:
oh, I always thought `\?` is BRE, but seems indeed be a [GNU extension](https://www.gnu.org/software/sed/manual/html_node/BRE-syntax.html). Thanks looks good

PR #12956, opened 2023-08-29
Fix wrong cursor position with virtual text before space and 'linebreak'

zeertzjq:
Problem: Wrong cursor position with virtual text before a whitespace
character and 'linebreak'.
Solution: Always set "col_adj" to "size - 1" and apply 'linebreak' after
adding the size of 'breakindent' and 'showbreak'.


chrisbra:
will you amend this commit to fix the other cases as well, or shall I first merge this one?

zeertzjq:
I think I've fixed the case with 'breakindent'/'showbreak' as well. It turns out that I need to change `col_adj` instead of starting at a different virtual columns.

PR #12958, opened 2023-08-29
autoconf: 2.69 -> 2.71

illia-bobyr:
Autoconf 2.69 is almost 10 years old.  And 2.71 is also a few years old as well.  Should be pretty well tested by now.  It brings a lot of improvements and there seems to be an ongoing work on autoconf 2.72 already.

This change just addresses two minor changes `autoupdate` suggested, and then `src/auto/configure` is regenerated by running
```
cd src
make AUTOCONF=autoconf2.71 autoconf
```


chrisbra:
Thanks, verified it works with autoconf2.69 and autoconf 2.71

ilya-bobyr:
> Thanks, verified it works with autoconf2.69 and autoconf 2.71

Awesome :)
Hope it would not break any of those very old systems that might still be in use.

ilya-bobyr:
> > > Thanks, verified it works with autoconf2.69 and autoconf 2.71
> > 
> > Awesome :) Hope it would not break any of those very old systems that might still be in use.
> 
> Bram mentioned about this some time ago (old version of autoconf used in some old HP/Solaris systems)

At least part of that discussion is here: https://github.com/vim/vim/pull/11322

PR #12967, opened 2023-08-30
Remove :Sman command via the undo_ftplugin mechanism

Konfekt:
Regards to @dkearns as noticed in
https://github.com/vim/vim/commit/2ac708b548660b232a32c52d89bde3d8596646c0

### DIFF
--- a/runtime/ftplugin/gpg.vim
+++ b/runtime/ftplugin/gpg.vim
@@ -27,7 +27,7 @@ if has('unix') && executable('less')
   if exists(':Sman') == 2
     setlocal iskeyword+=-
     setlocal keywordprg=:Sman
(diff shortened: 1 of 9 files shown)


chrisbra:
hm, those are buffer-local commands, they should be automatically deleted once the buffer is unloaded. I guess it doesn't hurt to explicitly delete them however. So okay

chrisbra (approved):

dkearns:
It's for when the current buffer's filetype is changed.

Thanks @Konfekt, I was going to follow up on the other ones but didn't get around to it.

chrisbra:
> It's for when the current buffer's filetype is changed.

yeah, that's what I thought

PR #12971, opened 2023-08-30
Make order of 'info' in completion_info consistent

### LINKED ISSUE #12230: `complete_info()` returns incorrect index with behavior that depends on `completeopt`
### Steps to reproduce

`complete_info()` returns incorrect index programmatically, causing problems for completion plugins relying on `complete_info()` for native `pumenu` interoperability. There is some subtle change in behavior depending at least on if `completopt` includes `noselect`.

See hrsh7th/nvim-cmp#1326 for more context on how this bug was discovered and neovim/neovim#22892 for the initial bug report there.

### Repro path
- Save this file as `minimal_vimrc`
  ```vim
  set completeopt=menu,preview
  set cpoptions-=<
  nmap \\ iaaa<CR>bbb<CR>ccc<CR>ddd<CR>eee<CR>fff<CR>
  imap , <Cmd>put =complete_info(['selected'])['selected']<CR>
  ```
- `vim --clean -u minimal_vimrc`
- normal mode `\\`, which is nmapped to give minimal example file
  ```
  aaa
  bbb
  ccc
  ddd
  eee
  fff
  ```
  and will put you in insert mode on a newline
- insert mode `<C-X><C-L>` to open lines `pumenu` -- with default `completeopt=menu,preview` will select "fff"
- insert mode `<C-P>`, to navigate up the list once -- selecting "eee"
- insert mode `,,,,`, imapped to `put` into the buffer the value of `complete_info` index
      - The `,` was hit multiple times so that the `complete_info` index isn't hidden behind the `pumenu` and you can see the value of the index

**BUG: The inserted index value is `1` when it should be `4`**

The bug behavior is that index counts from the bottom instead of from the top. However, whether it counts from top or bottom depends on `completeopt` value. 
- `completeopt=menu,preview`
    - behavior described above and seems to always count from bottom
- if `completeopt=menu,preview,noselect`, then the indexing shows 2 divergent behaviors
    - navigate `pumenu` with `<C-P>` first --> you get the reversed index (counting from bottom instead of top), acting like the `menu,preview` case 
    - navigate `pumenu` with `<C-N>` first --> you get the expected index, counting from top

### Expected behaviour

Hitting `,` should insert `4` from it's call to `vim.fn.complete_info({'selected'}).selected`

### Version of Vim

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Mar 21 2023 20:51:42) Included patches: 1-1420 Compiled by Arch Linux

### Environment

Operating System: Arch LInux, on WSL2 on Win11
Terminal: windows terminal
$Term: xterm-246color
Shell: bash

### Logs and stack traces

_No response_

LemonBoy:
Start the iteration from the same point and follow the same direction as done when assigning the completion numbers. This way we remove the dependance on the completion direction and make the order of 'info' consistent.

Fixes #12230 

Some more testing and/or a thorough review would be greatly appreciated ;)

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3040,6 +3040,43 @@ ins_compl_update_sequence_numbers(void)
     }
 }
 
+    static int
+info_add_completion_info(list_T *li)
+{
+    compl_T	*match;
+
+    if (compl_first_match == NULL)
+	return OK;
+
+    // Skip the element with the CP_ORIGINAL_TEXT flag at the beginning, in case of
+    // forward completion, or at the end, in case of backward completion.
+    match = compl_dir_forward()
+	    ? compl_first_match->cp_next : compl_first_match->cp_prev->cp_prev;
+    while (match != NULL && !match_at_original_text(match))
+    {
+	dict_T *di = dict_alloc();
+
+	if (di == NULL)
+	    return FAIL;
+	if (list_append_dict(li, di) == FAIL)
+	    return FAIL;
+	dict_add_string(di, "word", match->cp_str);
+	dict_add_string(di, "abbr", match->cp_text[CPT_ABBR]);
+	dict_add_string(di, "menu", match->cp_text[CPT_MENU]);
+	dict_add_string(di, "kind", match->cp_text[CPT_KIND]);
+	dict_add_string(di, "info", match->cp_text[CPT_INFO]);
+	if (match->cp_user_data.v_type == VAR_UNKNOWN)
+	    // Add an empty string for backwards compatibility
+	    dict_add_string(di, "user_data", (char_u *)"");
+	else
+	    dict_add_tv(di, "user_data", &match->cp_user_data);
+
+	match = compl_dir_forward() ? match->cp_next : match->cp_prev;
+    }
+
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks. I think it makes sense to refactor into a separate function. Let me include this.

Shane-XB-Qian:
is this possible did not handle or update index of 'selected' not well when textchangdp or cursormovedi event?
`info.items[info.selected]` happened out of range err.
// cannot repro at second time, just accidentally happened, please just review and think if it had outage there..

chrisbra:
possibly 🤷 
Please try to reproduce and open a new issue for it

Shane-XB-Qian:
// https://github.com/vim/vim/issues/12230#issuecomment-1511583354
so not sure `why it was done this way` and i actually not very understand what was the issue problem there, 
but looks it made new problem now, i cannot repro it or next time faced it, or wish someone else to jump into..

Shane-XB-Qian:
@chrisbra i donot know how to fix, but i prepared a test case: https://github.com/vim/vim/pull/13402

PR #12973, opened 2023-08-30
fix: crash when accessing a NULL object

mg979:
An object is NULL when the variable is declared, but the constructor isn't called. Accessing/setting a member on the object crashes Vim.

Note: this happens inside def functions, at script level things work differently. Accessing a NULL object member results in E1360 (correctly), while setting a value on it results in E1012 (type mismatch) so there's still something to fix.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -1180,6 +1180,36 @@ def Test_class_member()
   END
   v9.CheckScriptFailure(lines, 'E1010:')
 
+  # Test for setting a member on a null object
+  lines =<< trim END
+    vim9script
+    class A
+        this.val: string
+    endclass
+
+    def F()
+        var obj: A
+        obj.val = ""
+    enddef
+    F()
+  END
+  v9.CheckScriptFailure(lines, 'E1360: Using a null object')
+
+  # Test for accessing a member on a null object
+  lines =<< trim END
+    vim9script
+    class A
+        this.val: string
+    endclass
+
+    def F()
+        var obj: A
+        echo obj.val
+    enddef
+    F()
+  END
+  v9.CheckScriptFailure(lines, 'E1360: Using a null object')
+
   # Test for no space before or after the '=' when initializing a member
   # variable
   lines =<< trim END
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -1210,6 +1210,32 @@ def Test_class_member()
   END
   v9.CheckScriptFailure(lines, 'E1360: Using a null object')
 
+  # Test for setting a member on a null object, at script level
+  lines =<< trim END
+    vim9script
+    class A
+        this.val: string
+    endclass
+
+    var obj: A
+    obj.val = ""
+  END
+  # FIXME(in source): this should give E1360 as well!
+  v9.CheckScriptFailure(lines, 'E1012: Type mismatch; expected object<A> but got string')
+
+  # Test for accessing a member on a null object, at script level
+  lines =<< trim END
+    vim9script
+    class A
+        this.val: string
+    endclass
+
+    var obj: A
+    echo obj.val
+    F()
+  END
+  v9.CheckScriptFailure(lines, 'E1360: Using a null object')
+
   # Test for no space before or after the '=' when initializing a member
   # variable
   lines =<< trim END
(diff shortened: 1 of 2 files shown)


mg979:
To reproduce:
```vim
vim9script

class C
    this.member: string
endclass

def F()
    var obj: C
    obj.member = ""
enddef
F()
```

or:
```vim
vim9script

class C
    this.member: string
endclass

def F()
    var obj: C
    echo obj.member
enddef
F()
```

src/vim9execute.c
```diff
@@ -2147,7 +2147,14 @@ execute_storeindex(isn_T *iptr, ectx_T *ectx)
     // -1 dict, list, blob or object
     tv = STACK_TV_BOT(-3);
     SOURCING_LNUM = iptr->isn_lnum;
-    if (dest_type == VAR_ANY)
+
+    // Make sure an object has been initialized
+    if (dest_type == VAR_OBJECT && tv_dest->vval.v_object == NULL)
```
yegappan, line 2152: I think this check should be made a compile time check and not a runtime check (for performance reasons).

mg979: I'm not sure it's possible, for example if the NULL object is the argument of a function, how do you check that at compile time? Like in
```vim
vim9script

class C
    static val: string
endclass

def B(o: obj)
    echo o.val
enddef

def F()
    var obj: C
    B(obj)
enddef
F()
```

mg979: An alternative would be to disallow the initialization of an object without a constructor (that is, to always require `new()`) but I don't know how (if) it can be done, for example when the object is again the parameter of a function.

mg979: A NULL check doesn't seem too taxing anyway at runtime. Or not?

yegappan: We do need this runtime check when the object is used at the script level (where it is not compiled) and in other cases that you are referring to.  The change looks good to me.

yegappan (approved):

src/testdir/test_vim9_class.vim
```diff
@@ -1180,6 +1180,36 @@ def Test_class_member()
...
+        this.val: string
+    endclass
+
+    def F()
+        var obj: A
+        echo obj.val
+    enddef
+    F()
+  END
+  v9.CheckScriptFailure(lines, 'E1360: Using a null object')
```
yegappan, line 1211: Can you also add a test for accessing (read/write) the uninitialized object from the script level?

mg979: Ok

mg979: Added, but note that at script level it wasn't crashing. It gives a wrong error message in the case the variable is set and not accessed, but it's not a big deal.

errael: > Added, but note that at script level it wasn't crashing. It gives a wrong error message in the case the variable is set and not accessed, but it's not a big deal.

Actually, maybe it is.
@mg979 @yegappan  I just updated. What's the correct error? I can incorporate a fix into a PR I'm opening soon.

```
make test_vim9_class
...
Executed 53 tests                        in   0.824146 seconds
1 FAILED:
Found errors in Test_class_member():
command line..script /src/tools/vim/src/testdir/runtest.vim[601]..function RunTheTest[54]
    ..Test_class_member[170]..<SNR>9_CheckScriptFailure line 6: ['vim9script', 'class A', '
    this.val: string', 'endclass', '', 'def F()', '    var obj: A', '    obj.val = ""', 'enddef',
    'F()']: Expected 'E1360: Using a null object' but got 'E46: Cannot change read-only variable
    "val"': ['vim9script', 'class A', '    this.val: string', 'endclass', '', 'def F()', '    var
    obj: A', '    obj.val = ""', 'enddef', 'F()']

make: *** [Makefile:70: test_vim9_class] Error 1
```

errael: > > Added, but note that at script level it wasn't crashing. It gives a wrong error message in the case the variable is set and not accessed, but it's not a big deal.
> 
> Actually, maybe it is. @mg979 @yegappan I just updated. What's the correct error? I can incorporate a fix into a PR I'm opening soon.

@yegappan Oops, I just read the message more carefully. It might be a bug in 
```
patch 9.0.1829: Vim9 missing access-checks for private vars
```

mg979: @errael tests must be fixed because they were based on master where `public` didn't work.

PR #12974, opened 2023-08-30
runtime(optwin): Fix for 'splitkeep' option

ghost:
'spk' was used as a boolean, rather than a string option.


chrisbra:
thanks , makes sense

PR #12975, opened 2023-08-30
Add CI testing for i386

jamessan:
message_test currently fails on i386, which exposed a gap in the CI
testing.  Convert the shadowdir job to one that runs on i386 so we get
32-bit test coverage.

Since the GHA runners are x86_64, we can enable the i386 architecture
in dpkg and install i386 packages for the i386 CI jobs.  However, this
can't currently be done with features=huge since that would require
installing python3-dev:i386, which breaks the CI environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,10 @@ jobs:
         features: [tiny, normal, huge]
         compiler: [clang, gcc]
         extra: [[]]
+        # Only use non-native architecture when features != huge.
+        # features=huge tries to install python3-dev, which fails to install
+        # for the non-native architecture.
+        architecture: [native]
         include:
           - features: tiny
             compiler: clang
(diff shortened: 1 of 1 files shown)


jamessan:
Merging this would temporarily break the shadowdir job, but the shadowdir infrastructure is less affected by day-to-day changes, so maybe it's worth merging this to help debug the message_test failure.

chrisbra:
so do we still need this one? And if yes, since this is in conflict with the shadowdir job, can we make those two work together?

jamessan:
> so do we still need this one?

I would say yes, so we catch problems like this in the future _before_ the changes get merged.

> since this is in conflict with the shadowdir job, can we make those two work together?

It's not in conflict with it.  I just figured that we could make the shadowdir job run with i386 to avoid adding a new item to the matrix.  The shadow functionality is orthogonal to the architecture, so we can get two tests in one.

chrisbra:
okay thanks!

PR #12980, opened 2023-08-31
Fix some documentation example formatting

dkearns:
Problem:  Help documentation examples are badly formatted.
Solution: Add the appropriate markup.

@chrisbra, maybe you can add this to some batch update if you don't want to add it now. Thanks.


chrisbra:
included here: #12790

Thanks!

PR #12981, opened 2023-08-31
runtime(ruby): Update syntax, indent and ftplugin files

dkearns:


chrisbra:
thanks, included as of da16a1b471aa717f58909cc6531cb6dbbff14d22 (I also squashed in @tpope s  #12994 )

PR #12982, opened 2023-08-31
fix: vim tests for null object

mg979:
`public` is now required because it actually works.


mg979:
See https://github.com/vim/vim/pull/12973#discussion_r1311962738

chrisbra:
thanks all for the quick fix!

PR #12983, opened 2023-08-31
runtime(gitrebase): Add update-ref to syntax

nkouevda:
This was introduced in git version 2.38.0


dkearns:
@tpope

tpope:
I have https://github.com/tpope/vim-git/pull/90 on my end I've been meaning to deal with. I'll make sure it makes it into the next update.

PR #10265, opened 2022-04-24
highlight more C keywords, including some from C23

albinahlback:
Added some keywords, some that are from C99 and some from C23.

As C23 is not completed yet, I chose to enforce users to define `c_c23` if they want such features.

It is very okay if this PR sits for a year until C23 has been set in stone.

PR #12984, opened 2023-08-31
runtime(c): Support new C23 features

### LINKED ISSUE #13667: Add support for #elifdef and #elifndef to c.vim
The new C2x and C++23 standards add new preprocessing conditionals, `#elifdef` and `#elifndef`

For C the proposal was:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2645.pdf

For C++ the proposal was:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2334r1.pdf

They've been supported by GCC for a couple of years:
https://gcc.gnu.org/gcc-12/changes.html#c

Vim does not recognize these and so doesn't syntax highlight them.

```c
#ifdef foo
#elifdef bar
#elifndef baz
#endif
```

dkearns:
Conditionally highlight C23 features:
    
- #embed, #elifdef and #elifndef preprocessor directives
- predefined macros
- UTF-8 character constants
- binary integer constants, _BitInt literals, and digit separators
- nullptr_t type and associated constant
- decimal real floating-point, bit precise and char types
- typeof operators
    
Matchit:
- update for new preprocessor directives
    
Fixes #13667 and fixes #13679.


albinahlback:
Status of this? I'll close #10265 in favor of this

runtime/syntax/c.vim
```diff
@@ -277,6 +306,13 @@ if !exists("c_no_c99") " ISO C99
   syn keyword	cType		intptr_t uintptr_t
   syn keyword	cType		intmax_t uintmax_t
 endif
+if !exists("c_no_c23")
+  syn keyword	cOperator	typeof typeof_unqual
+  syn keyword	cType           _BitInt _Decimal32 _Decimal64 _Decimal128
```
jwakely: Maybe I'm misreading the conditions, but does this make these enabled in C++ too? These keywords are defined in the C23 standard, but are not in any C++ standard (although `typeof` and the `_DecimalNN` types are supported by G++ as extensions). It won't hurt to treat `_BitInt` and the decimal types as keywords in C++ code, since they're reserved names and so can't be used for users' identifiers. But in theory a C++ program could use `typeof` and `typeof_unqual` as variable names, or functions, or macros, but this would highlight them as operator keywords (... unless I'm just reading it wrong).

dkearns: Yes, these will be propagated to the C++ filetype.  I'll fix it, thanks.

jwakely (approved):

chrisbra:
thanks

Soupertonic:
I could be wrong but `constexpr` seems to be not included which is as of C23 standard is a storage-class specifier and should be highlighted.

Issue #12946, opened 2023-08-29
message_test fails on x86 after #12140


jamessan:
### Steps to reproduce

Build and test in an x86 32-bit environment.  This appeared in the CI for the Debian packaging.  I haven't had a chance to look at it further, but wanted to post the issue so it was known. I'll look into it over the next few days if @cvwillegen isn't able to.

### Expected behaviour

Tests all pass.

### Version of Vim

9.0.1811

### Environment

Debian sid

### Logs and stack traces

```shell
message_test: message_test.c:446: test_vim_snprintf_positional: Assertion `n == 6' failed.
Aborted (core dumped)
```

k-takata:
https://github.com/vim/vim/blob/00d07a97ad750f2d2849a649f85ab1610a0b0866/src/message_test.c#L445-L446
It looks that passing a 64-bit number causes the issue. (`(uvarnumber_T)12`)
`fmt_06pb` is defined as `"%1$0.*2$b"`. Should it be `"%1$0.*2$Lb"` to specify that the 1st parameter is 64-bit?
Line 455 should be also the same.

jamessan:
> `fmt_06pb` is defined as `"%1$0.*2$b"`. Should it be `"%1$0.*2$Lb"` to specify that the 1st parameter is 64-bit?

That results in

```
E1400: Cannot mix positional and non-positional arguments: %1$0.*2$Lbmessage_test: message_test.c:446: test_vim_snprintf_positional: Assertion `n == 6' failed.
Aborted
```

jamessan:
Also, the `fmt_06b` tests work fine on 32-bit system, but fail the `n == 6` assert when changed to use `%06Lb`.

cvwillegen:
@RestorerZ also reported an error using positional arguments. Luckily, that was only a documentation error :-)

This one looks like a code error, I'm looking into it. Changing to ``Lb`` gives an assertion failure on my system, as well...

chrisbra:
I think this was fixed by #12985

PR #12140, opened 2023-03-11
Format positional arguments

cvwillegen:
This PR adds the possibility to use positional arguments in the printf() function, and in translation files, opening the possibility for translators to specify word order in translations. 

Ex: echo printf("My name is %2$s, %1$s", "Bram", "Moolenaar")

echo printf("My date of birth is: %1$02d/%2$02d/%3$04d", 13, 7, 2002)

echo printf("My date of birth is: %3$04d/%2$02d/%1$02d", 13, 7, 2002)

echo printf("My date of birth is: %3$0*1$d/%4$0*1$d/%5$0*2$d", 2, 4, 13, 7, 2002)

PR #12985, opened 2023-08-31
Fix positional arguments

cvwillegen:
This PR fixes problems that were detected in #12140 and #12946 

It adds tests for the examples given in the builtin help, (probably) fixes problems with the width of the argument supplied to printf() and adds source code comments to make the functions clearer.


cvwillegen:
> When I cherry-pick this on top of #12975, the 32-bit build fails with
> 
> ```
> message_test: message_test.c:377: test_vim_snprintf_positional: Assertion `n == 15' failed.
> ```

I think that's positive! The test errors later now... I'll go and update the other ``UL`` constants to use ``ll`` specifiers.

And I'll add an internal error for ``ap_types`` being ``NULL`` while I'm at it.

k-takata:
> I'll go and update the other `UL` constants to use `ll` specifiers.

Is it correct?
I think `UL` constants are 32bit and `ll` specifiers are 64bit on 32bit platforms.

cvwillegen:
> > I'll go and update the other `UL` constants to use `ll` specifiers.
> 
> Is it correct? I think `UL` constants are 32bit and `ll` specifiers are 64bit on 32bit platforms.

No, it isn't :-(

But, then I'm not sure why the tests fail in 32 bits if adding an `ll` does fix it _sometimes_.

I'll get back to this, if you don't have any insights now. I'm stumped!

k-takata:
If the format specifier is 'b' or 'B', `length_modifier` is not checked:
https://github.com/vim/vim/blob/816fbcc262687b81fc46f82f7bbeb1453addfe0c/src/strings.c#L2332-L2349

k-takata:
So, how about removing these two lines?

```diff
--- a/src/strings.c
+++ b/src/strings.c
@@ -2329,8 +2329,6 @@ format_typeof(
 
 	    if (fmt_spec == 'p')
 		return TYPE_POINTER;
-	    else if (fmt_spec == 'b' || fmt_spec == 'B')
-		return TYPE_UNSIGNEDINT;
 	    else if (fmt_spec == 'd')
 	    {
 		// signed
```

cvwillegen:
> If the format specifier is 'b' or 'B', `length_modifier` is not checked:
> 
> https://github.com/vim/vim/blob/816fbcc262687b81fc46f82f7bbeb1453addfe0c/src/strings.c#L2332-L2349

I've installed a (virtual) 32-bits system to find out what went wrong, and I fixed the error in getting the correct argument.

src/strings.c
```diff
@@ -2330,7 +2333,7 @@ format_typeof(
 	    if (fmt_spec == 'p')
 		return TYPE_POINTER;
 	    else if (fmt_spec == 'b' || fmt_spec == 'B')
-		return TYPE_UNSIGNEDINT;
+		return TYPE_UNSIGNEDLONGLONGINT;
```
k-takata, line 2323: Do the following tests work after this change?

```
	n = vim_snprintf(buf, bsize, "%1$0.*2$llb", (uvarnumber_T)12, 6);
	assert(n == 6);
	assert(bsize == 0 || STRNCMP(buf, "001100", bsize_int) == 0);
	assert(bsize == 0 || buf[MIN(n, bsize_int)] == '\0');

	n = vim_snprintf(buf, bsize, "%1$0.*2$b", 12, 6);
	assert(n == 6);
	assert(bsize == 0 || STRNCMP(buf, "001100", bsize_int) == 0);
	assert(bsize == 0 || buf[MIN(n, bsize_int)] == '\0');
```

k-takata: Rethinking. 'b' is a Vim-specific specifier, so it might be not mandatory to support the latter case.

src/strings.c
```diff
@@ -2927,12 +2928,17 @@ vim_vsnprintf_typval(
...
+    int		usetvs = 0;
     va_list	ap;
     const char	**ap_types = NULL;
 
     if (parse_fmt_types(&ap_types, &num_posarg, fmt, tvs) == FAIL)
 	return 0;
 
+# if defined(FEAT_EVAL)
+    usetvs = (tvs != NULL);
+# endif
```
k-takata: Is this needed?

src/strings.c
```diff
@@ -3067,7 +3073,8 @@ vim_vsnprintf_typval(
 # if defined(FEAT_EVAL)
 		    tvs != NULL ? tv_nr(tvs, &arg_idx) :
 # endif
-			(skip_to_arg(ap_types, ap_start, &ap, &arg_idx, &arg_cur, fmt),
+			(skip_to_arg(ap_types, ap_start, &ap, &arg_idx,
+				     &arg_cur, usetvs, fmt),
```
k-takata: `usetvs` is always false here.

src/strings.c
```diff
@@ -2848,13 +2849,13 @@ skip_to_arg(
...
 	{
-	    semsg(e_aptypes_is_null_str, fmt, *arg_cur);
+	    semsg(e_aptypes_is_null_str_nr, fmt, *arg_cur);
 	    return;
 	}
 
 	p = ap_types[*arg_cur];
 
-	int fmt_type = format_typeof(p, TRUE);
+	int fmt_type = format_typeof(p, usetvs);
```
k-takata: So, this should be as follows, I think.
```
	int fmt_type = format_typeof(p, FALSE);
```

Additionally, the following changes might be needed?
```diff
--- a/src/strings.c
+++ b/src/strings.c
@@ -2293,7 +2293,7 @@ format_typeof(
     {
 	switch (fmt_spec)
 	{
-	    case 'd': case 'u': case 'o': case 'x': case 'X':
+	    case 'd': case 'u': case 'o': case 'x': case 'X': case 'b': case 'B':
 		if (length_modifier == '\0')
 		    length_modifier = 'L';
 	}
```

src/strings.c
cvwillegen: > Rethinking. 'b' is a Vim-specific specifier, so it might be not mandatory to support the latter case.

Yes, and 'b' has always been 64 bits as well. If you look at the way the format specifier is put into the source code, as a global ``char *`` and not in the sprintf() function proper, you can see that this has been done to work around a compiler warning (if you put the format string in inside the call, the compiler will warn you about the format specifier not being ll, and the argument is. Oh, no, that's not it: It complains about the 'b' specifier being unknown).

In any case, I think that the 'b' specifier is always 64 bit, because it has always been that way in the past.

I'll re-work the the 'usetvs' you mentioned, like  (what was I thinking...)

chrisbra:
is this ready?

cvwillegen:
> is this ready?

The test failures look unrelated, and the tests that should pass, do pass. I've tested on 32 and 64 bits myself. 

Perhaps @k-takata can take one last look at it, because I've resolved their feedback?

chrisbra:
okay thanks. I'll check it tomorrow then :)

k-takata:
Looks OK.

And, how about this document update?
```diff
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -6706,8 +6706,10 @@ printf({fmt}, {expr1} ...)				*printf()*
 			a conversion is wider than the field width, the field
 			is expanded to contain the conversion result.
 			The 'h' modifier indicates the argument is 16 bits.
-			The 'l' modifier indicates the argument is 32 bits.
-			The 'L' modifier indicates the argument is 64 bits.
+			The 'l' modifier indicates the argument is a long
+			integer.  The size will be 32 bits or 64 bits
+			depending on your platform.
+			The "ll" modifier indicates the argument is 64 bits.
 			Generally, these modifiers are not useful. They are
 			ignored when type is known from the argument.
 
```

jamessan:
New checks in #12975 pass with these changes, too. 👍🏻

chrisbra:
thanks included it. There were some conflicts in message_test.c, I hope I resolved it correctly 🙈

PR #12986, opened 2023-08-31
xxd: Fix -R argument parsing

illia-bobyr:
`-R` argument is specified as optional, yet, as written, it always consumes an argument.
Or was it intended to be a required value and the documentation need to be changed from `[WHEN]` to `WHEN`?

"auto" value has no effect, meaning if `-R` is repeated and the second case uses "auto" it does not overwrite the previous setting.

---

I wonder if `-R` is the best name for this argument.

If we use `less` as a reference, in `less` `-R` is for `--RAW-CONTROL-CHARS`.
`less` has a separate color argument called `-D` or `--color`.
`ls` uses `--color[=WHEN]` with no short form.

Existing arguments in `xxd` have long names, even though they are not documented.
It is probably a bit annoying to write `--color` all the time in the command line.
But the argument default is actually "auto".
`-c` is already used.
Next unused would be either `-D` or `-j`.

### DIFF
--- a/src/xxd/xxd.c
+++ b/src/xxd/xxd.c
@@ -719,14 +719,38 @@ main(int argc, char *argv[])
         }
       else if (!STRNCMP(pp, "-R", 2))
         {
-	  if (!argv[2])
-	    exit_with_usage();
-	  if (!STRNCMP(argv[2], "always", 2))
-	    color = 1;
-	  else if (!STRNCMP(argv[2], "never", 1))
-	    color = 0;
-	  argv++;
-	  argc--;
+          if (argv[2] &&
+                  (!STRNCMP(argv[2], "always", 6) ||
+                   !STRNCMP(argv[2], "never", 5) ||
+                   !STRNCMP(argv[2], "auto", 4) ) )
+            {
+              if (argv[2][1] == 'l')
+                /* "a[l]ways" */
+                color = 1;
+              else if (argv[2][0] == 'n')
+                /* "[n]ever" */
+                color = 0;
+              else
+                {
+                  /* "auto" */
+#ifdef UNIX
+                  color = isatty(STDOUT_FILENO);
+#else
+                  color = 0;
+#endif
+                }
+              argv++;
+              argc--;
+            }
+          else
+            {
+              /* "auto" is default */
+#ifdef UNIX
+              color = isatty(STDOUT_FILENO);
+#else
+              color = 0;
+#endif
+            }
         }
       else if (!strcmp(pp, "--"))	/* end of options */
 	{


chrisbra:
Thanks, I merged Kens patch #12997 but I think it fixes the same.
can you please check, if https://github.com/vim/vim/commit/f6fc255e8d9c46a0e51e03f16a508833d309dee6 works as expected?

aapo:
```
xxd -R always -R auto xxd | head
```
Should not contain colors. Fixed on https://github.com/vim/vim/pull/12986/commits/0f073c317b58a1758d22d2499d8b50a7fb808508. Not fixed on https://github.com/vim/vim/commit/f6fc255e8d9c46a0e51e03f16a508833d309dee6

Manpage was misleading. It should not contain brackets on `when`. Fixed on https://github.com/vim/vim/commit/f6fc255e8d9c46a0e51e03f16a508833d309dee6.

Respecting "NO_COLOR" environment variable is good. Fixed on https://github.com/vim/vim/commit/f6fc255e8d9c46a0e51e03f16a508833d309dee6.

`-R` might not be the best choice. I don't know what letter to use but how about `-j = color` and `-J =not-color`, omitting them is `color=auto`.

chrisbra:
Can you please send a PR?Am 03.09.2023 um 16:57 schrieb Aapo Rantalainen ***@***.***>:﻿
xxd -R always -R auto xxd | head

Should not contain colors. Fixed on 0f073c3. Not fixed on f6fc255
Manpage was misleading. It should not contain brackets on when. Fixed on f6fc255.
Respecting "NO_COLOR" environment variable is good. Fixed on f6fc255.
-R might not be the best choice. I don't know what letter to use but how about -j = color and -J =not-color, omitting them is color=auto.

aapo:
```
    -------------------------------- src/xxd/xxd.c --------------------------------
    index f8601eb09..7c23ba463 100644
    @@ -759,7 +759,7 @@ main(int argc, char *argv[])
     	  else if (!STRNCMP(pw, "never", 5))
     	    color = 0;
     	  else if (!STRNCMP(pw, "auto", 4))
    -	    ;	/* Do nothing. */
    +	    color = enable_color();
     	  else
     	    exit_with_usage();
             }
```

chrisbra:
??? pleas send a PR!

k-takata:
#13021

ilya-bobyr:
> #13021

While it seems to do the job, I am a bit confused by the `enable_color()` method.
It seems that the name and the functionality are not completely aligned.
It checks if the stdout is a console, but then it also enables VT100 processing on Windows.
And it returns `isatty()` value, except if we failed to enable VT100.

I think, an attempt to enable VT100 processing should happen if `color` ends up being `1` at the end of argument processing.
Which makes a minor difference, if we assume that `-R` can be specified multiple times.

If `SetConsoleMode()` is moved out of `enable_color()`, then it just becomes `isatty()`.
In any case, it is a bit confusing that a method called `enable_color()` returns a value.
It is unclear what the return value means, until you read the method implementation.

If the above makes sense, I'll send a PR later, to split the VT100 specific Win32 logic from the argument processing code.

---

I think `-j`/`-J` for `--color=always`/`--color=never` seems reasonable.
I'll send a separate PR for this, if nobody beats me to it.

k-takata:
> If `SetConsoleMode()` is moved out of `enable_color()`, then it just becomes `isatty()`.
> In any case, it is a bit confusing that a method called `enable_color()` returns a value.

agree.

If the following code is added, then we should be able to use `isatty(STDOUT_FILENO)` both on Windows and Linux.
```
#ifndef STDOUT_FILENO
# define STDOUT_FILENO 1
#endif
```

> I think `-j`/`-J` for `--color=always`/`--color=never` seems reasonable.

I'd like to know where the "j" comes from.

ilya-bobyr:
> > I think `-j`/`-J` for `--color=always`/`--color=never` seems reasonable.
> 
> I'd like to know where the "j" comes from.

Just the next unused letter.
Look at the first post in this PR: https://github.com/vim/vim/pull/12986#issue-1876138156

PR #12978, opened 2023-08-31
Use the proper check for private/read-only member variable

yegappan:
Access level for a member cannot be changed in a class implementing an interface.  Update the code indentation in a few places.

PR #12989, opened 2023-08-31
Modify class_member_type() to provide more information

errael:
This is a follow on to #12978. In that PR a pointer argument was added to `class_member_type()` through which a value was returned. This PR modifies `class_member_type()` to instead indirectly and optionally return a pointer to an existing structure which contains the desired value plus a bit more. This allows the  getting rid of an alloc/free; may be useful in the future.


yegappan (approved):

errael:
@yegappan milestone 9.1? It feels like part of #12978 and while it's fresh in mind...

yegappan (approved):

PR #12993, opened 2023-08-31
runtime(menu): Fix menu localization not updated for "Open Tab..." item

ychin:
THe menu item "Open Tab..." was changed to "Open &Tab..." in #12895 but localization files were not updated. Update it here.


chrisbra:
oh my, if I knew what I did there 🙈 

Thanks!

PR #12994, opened 2023-08-31
Check for cwd ruby does not account for :cd

tpope:
Problem:    Check for cwd ruby does not account for :cd (after 9.0.1833)
Solution:   Move exepath() conditional after the :cd invocation.


chrisbra:
thanks, included as of https://github.com/vim/vim/commit/da16a1b471aa717f58909cc6531cb6dbbff14d22 (I squashed it into #12981

PR #12914, opened 2023-08-25
Fix Perl 5.30 build issues, and remove GIMME_V usage

ychin:
This makes Perl 5.30 dynamic builds work again, by reverting GIMME_V back to GIMME. In Perl headers, GIMME_V requires usage of functions from inline.h which requires manual copy-paste into if_perl.xs which is fragile. Just use GIMME for now (which was the case before) as its implementation in Perl headers does not require using inline functions.

Also, fix the configure script so that if we are using dynamic linkage, we don't pass `-lperl` to the build flags, to avoid accidental external linkage. This is similar to how Python integration works.

Also fix misleading comments in if_perl.xs saying that we include perl\lib\CORE\inline.h manually. That was the case in the past but changed to manually copying inline functions over to fix 5.22 build issues with PERL_NO_INLINE_FUNCTIONS.

PR #12996, opened 2023-09-01
Fix Perl interface to load PL_current_context from library

ychin:
In #12914, in order to fix an issue with Perl 5.36 dynamic builds, (that version introduced a thread-local `PL_current_context`), the file added the variable manually so we can satisfy the linker. However, the variable is a different one from the one in the library, so there could be unpredictable behavior. Instead, just use `dlsym` to load the context from the library. The fact that it's thread-local doesn't matter too much to us because Vim's interface is single-threaded so it will work properly.

### DIFF
--- a/src/if_perl.xs
+++ b/src/if_perl.xs
@@ -315,6 +315,9 @@ typedef int perl_key;
 #   define PL_thr_key *dll_PL_thr_key
 #  endif
 # endif
+# ifdef PERL_USE_THREAD_LOCAL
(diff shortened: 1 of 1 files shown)


ychin:
This is to address this comment: https://github.com/vim/vim/commit/55460da26c2756ec057c03c7d8641eda861bfcd2#r125864409

From testing they both build, but this should be the more correct way to do it.

PR #12997, opened 2023-09-01
xxd: Improve colored output

k-takata:
* Support colored output on Windows.
  SetConsoleMode() is required to enable ANSI color sequences.
* Support "NO_COLOR" environment variable.
  If "NO_COLOR" is defined and not empty, colored output should be disabled.
  See https://no-color.org/
* "-R" should only accept "always", "never" or "auto" as the parameter.
* Allow omitting space after "-R".
* Adjust help and documentation.
  "-R" cannot omit the parameter. Remove surrounding brackets.

Related #12131

### DIFF
--- a/runtime/doc/xxd.1
+++ b/runtime/doc/xxd.1
@@ -135,9 +135,9 @@ to read plain hexadecimal dumps without line number information and without a
 particular column layout. Additional whitespace and line breaks are allowed
 anywhere.
 .TP
-.IR \-R " "[WHEN]
+.IR \-R " " when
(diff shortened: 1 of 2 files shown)


k-takata:
Man page of xxd says:

       Note  that  a  "lazy" parser is used which does not check for more than
       the first option letter, unless the option is followed by a  parameter.
       Spaces  between  a single option letter and its parameter are optional.
       Parameters to options can be specified in decimal, hexadecimal or octal
       notation.  Thus -c8, -c 8, -c 010 and -cols 8 are all equivalent.

> which does not check for more than the first option letter,

This doesn't say about the parameters.
Personally, I think that "-R" should only accept "always", "never" or "auto" as the parameter.

PR #13005, opened 2023-09-01
feat: add runtime files of zserio syntax

dpelle:


dpelle:
I had previously made a PR for the zserio syntax at https://github.com/vim/vim/pull/12544/files
Bram had integrated some files in that old PR in patch 9.0.1645, but not all of them, delaying the remaining files for the runtime update, which sadly never happened as Bram sadly passed away.

PR #13007, opened 2023-09-01
Prevent direct access to interface statics. Handle read static from :def.

errael:
Prevent direct access to interface statics.
For `:def F1(i: Iface)` and `:def F2(o: Obj)` handle reading statics.
static access tests.


errael:
@yegappan This, #13007, seems to take care of preventing access to the interface statics.

There 's the problem that
```
def F(i: I)
    var x = i.s_var
enddef
```
gets the compilation error as mentioned. The error's coming from vim9expr.c (I took a look (now that my  panic over statics is milder (still don't like the semantics but it is what it is)). Need to generate the code with the index from the object not the interface, the object was just pushed, not sure how to get hold of it and then generate the code.

src/vim9expr.c
```diff
@@ -439,6 +459,12 @@ compile_class_object_index(cctx_T *cctx, char_u **arg, type_T *type)
 	    ocmember_T *m = &cl->class_class_members[idx];
 	    if (STRNCMP(name, m->ocm_name, len) == 0 && m->ocm_name[len] == NUL)
 	    {
+		if ((cl->class_flags & CLASS_INTERFACE) != 0)
+					//&& m->ocm_type->tt_type == VAR_CLASS)
+		{
+		    semsg(_(e_interface_static_direct_access_str), m->ocm_name);
+		    return FAIL;
```
yegappan, line 466: Can you add tests for the two new cases?

errael: > Can you add tests for the two new cases?

Of course, when it's ready to come out of draft. 

I've left this in draft because of the following problem, mentioned above in https://github.com/vim/vim/pull/13007#issuecomment-1703677466. In vim9expr.c I left a chunk of _not working_ code commented out. I'm still investigating; do you have any suggestions , or can point me at an example that's similar?

What about the following which fails with
```
Error detected while compiling :source buffer=1[14]..function <SNR>36_F:
line    1:
E1326: Member not found on object "I": s_var
```
produce by
```
vim9script

interface I
    public static s_var: number
endinterface
class C implements I
    public static s_var = 3
endclass

def F(i: I)
    var x = i.s_var
enddef
var c = C.new()
F(c)
```

src/vim9class.c
```diff
@@ -183,9 +183,12 @@ add_members_to_class(
...
     int
-object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl)
+object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl,
+								int is_static)
 {
+    // TODO: should the following be ">="
     if (idx > (is_method ? itf->class_obj_method_count
+				   : is_static ? itf->class_class_member_count
 						: itf->class_obj_member_count))
     {
```
errael: Looks like a bug; should be `>=` instead of `>`. If reviewer(s) agree, I'll make the change.

errael:
This PR got a lot bigger, it now handles `:def F1(i: Iface)` and `:def F2(o: Obj)` reading statics through the argument. I haven't tried to write a static, or tried much at the script level; haven't tried access control; so there will need to be follow on PR(s).

Most important is to take a look at the overall approach, adding a `is_static` argument and `isn_arg.classmember.cm_static`. I did this first for the `GET_ITF_MEMBER`, and was going to ask for review on that. But then I took a look at `GET_OBJ_MEMBER` and saw that it was almost done as well; after all, the two are handled together in the code.

I think there's also a bug around vim9class.c::189. I'll fold in the fix if you agree.

Will look at tests now. They'll be needed whether or not this approach is sound.

src/vim9expr.c
errael: > Can you add tests for the two new cases?

See `Test_interface_static_member_access()`

src/vim9class.c
errael: Made the change, didn't want to lose it. This PR should be ready to review/go without changes

errael:
Added small commit to fix scriptlevel crash: `var x = I.s_var1` wrong messge: `I.s_var1 = 7`

errael:
Crap. There's a conflict with the last item in error.h. Wonder how that got there. I'll check it out.

mg979:
Just my opinion, but a big PR that touches a lot of files and even adds a new struct member (probably avoidable) only to provide a temporary workaround to a crash that should be solved in some other way, also concerning the current bad implementation of interfaces that needs to be redesigned, is not worth it.

errael:
Re-assigned Enumber  for new error message, fixes conflict. vim9 tests pass.

errael:
Added some static access tests

Some additional tests to cover areas exposed by Codecov.
One part of static access protection (not generally covered yet)
```
    vim9expr.c::compile_class_object:423
        semsg(_(e_cannot_access_private_member_str), m->ocm_name);
```
Note: the additional siemsg not tested
```
    vim9class.c::object_index_from_itf_index:227
```

Not part of this change.
```
    vim9compile::compile_load_lhs with_index()
        return generate_GET_ITF_MEMBER(cctx, cl, lhs->lhs_member_idx, type,
```

errael:
@yegappan I think this is complete.

I was looking through my notes and list_instructions probably needs a tweak; it can be done as a follow on PR or an additional commit to this PR.

 This PR essentially adds a `cm_static` option to `ISN_GET_OBJ_MEMBER` and `ISN_GET_ITF_MEMBER`.

Currently when disassembling we see
```
1 ITF_MEMBER 0 on I
1 OBJ_MEMBER 0
```
One thought is to append `[ static]` as needed. Current listing is unchanged.

src/errors.h
```diff
@@ -3513,5 +3513,7 @@ EXTERN char e_method_str_type_mismatch_expected_str_but_got_str[]
 	INIT(= N_("E1407: Member \"%s\": type mismatch, expected %s but got %s"));
 EXTERN char e_aptypes_is_null_str_nr[]
 	INIT(= "E1408: Internal error: ap_types or ap_types[idx] is NULL: %d: %s");
+EXTERN char e_interface_static_direct_access_str[]
+	INIT(= N_("E1409: Interface static \"%s\": cannot directly access"));
```
yegappan: Should this read: "Cannot directly access interface \"%s\" static member"?

yegappan (approved):

src/errors.h
errael: I didn't like the wording either. So for an interface, `Iface` static variable `s_var1` your suggestion would be
```
E1409: Cannot directly access interface "s_var1" static member
```
That reads a little funny too. How about
```
E1409: Cannot directly access interface static member "s_var1"
```
I thought only naming the variable would be sufficient since there's the line number.
But, assuming the name of the interface is available, it could be changed to
```
E1409: Cannot directly access interface "Iface" static member "s_var1"
```

I won't make a change until hearing back, and there may be more changes to come ;-)

errael: There's the question of list_instructions mentioned at https://github.com/vim/vim/pull/13007#issuecomment-1705341446

Something should be done (next PR is fine); suggestions/comments?

yegappan: The second message with both the interface name and the member name looks good to me.

yegappan: As this PR is getting larger, can you create a separate PR for the list_instructions() change?  We also need to add a lot more tests to the test_vim9_disassemble.vim file for Vim9 classes, objects and interfaces.

errael:
Rebase/push to incorporate current master and fix conflict.

Next push, enhanced error message, will be easier to see only those changes.

errael:
@yegappan pushed error message update.

src/vim9class.c
```diff
@@ -245,8 +247,28 @@ object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl)
...
-    int *table = (int *)(i2c + 1);
-    return table[idx];
+    if (is_static)
+    {
+	// TODO: Need a table for fast lookup?
+	char_u *name = itf->class_class_members[idx].ocm_name;
+	for (int i = 0; i < i2c->i2c_class->class_class_member_count; ++i)
+	{
+	    ocmember_T *m = &i2c->i2c_class->class_class_members[i];
+	    if (STRCMP(name, m->ocm_name) == 0)
```
yegappan, line 257: Looking at this code now, comparing the name of every class member variable is expensive to do at run time for every access.  Is it possible to move this to compile time and use a lookup table?

errael: > comparing the name of every class member variable is expensive to do at run time

I agree, that's why I left the comment
```
    if (is_static)
    {
	// TODO: Need a table for fast lookup?
```

>  Is it possible to move this to compile time and use a lookup table?

Probably, but I don't know how the table used for non-static is built. First thing was to get something correct, this may be good enough for 9.1. And with the accessing statics through objects up in the air, it may never be needed. 

(and looking at how method_offset is used in #13053 I'm feeling even more confused, need to draw some diagrams and check out the code that builds the tables, if and when the time comes)

I can open a performance improvement issue about it if you like.

PR #13009, opened 2023-09-02
Rework interfaces

mg979:
The PR works as is, marked as draft because it needs discussion.

It's a rework of the `interface` implementation. Actually only rather
superficial changes were needed, but the way interfaces are used is very
different.

Consider that the current implementation of `interface`:
- is only slightly different from an abstract class
- notwithstanding that, you can implement many interfaces, all with their
  members
- I think you can even extend an interface with a class, since they are classes

In short, they aren't interfaces, they are another sort of abstract classes, but
while you're limited to one parent class, you can have a class implement any
number of interfaces... That are full classes.

This PR aims at making them behave as interfaces, as they work in other OOP
languages.

Main changes are:

Interfaces aren't classes: they cannot define `new()` constructors since they
cannot be instantiated as objects.

Interfaces can only define methods, not members (not even static).

Interfaces can only be implemented by classes with the implements keyword:

- they cannot extend a class with `extends`
- they cannot be extended with `extends`, neither by interfaces or classes
- they cannot implement other interfaces with `implements`

As for methods, currently all types are supported, including private methods
(with underscore), something that maybe could be disallowed.

A possibility would be to allow interfaces to extend other interfaces,
currently not possible.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -400,7 +400,7 @@ a number.  This example extends the one above: >
 	      return this.base * this.height / 2
 	   enddef
 	endclass
-< 							*E1349*
+<							*E1349*
 If a class declares to implement an interface, all the methods specified in
 the interface must appear in the class, with the same signatures.
 
@@ -596,7 +596,7 @@ return type, but without the body and without `:enddef`.  Example: >
 	interface HasSurface
 	   def Surface(): number
 	endinterface
-< 							*E1343*
+<							*E1343*
 An interface name must start with an uppercase letter.
 The "Has" prefix can be used to make it easier to guess this is an interface
 name, with a hint about what it provides.
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -47,9 +47,10 @@ An object can only be created by a class.  A class provides:
 - State shared by all objects of the class: class variables (class members).
 - A hierarchy of classes, with super-classes and sub-classes, inheritance.
 
-An interface is used to specify properties of an object:
+An interface is used to specify methods of an object:
 - An object can declare several interfaces that it implements.
-- Different objects implementing the same interface can be used the same way.
+- Objects belonging to different classes, implementing the same interface, can
+  be used in the same way.
 
 The class hierarchy allows for single inheritance.  Otherwise interfaces are
 to be used where needed.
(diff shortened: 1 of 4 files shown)


yegappan:
I don't think we should make this change to remove members from interfaces.  I am not sure what is the benefit of doing this. Note that Bram had planned for a "specifies" mechanism to directly use a class as an interface (as described in vim9class.txt):

```
A class defining an interface 
							*specifies*
A class can declare its interface, the object members and methods, with a
named interface.  This avoids the need for separately specifying the
interface, which is often done in many languages, especially Java.
```

The support for "specifies" is not yet implemented.

The following text in vim9class.txt indirectly describes why the members are part of the interface:

```
Explicitly declaring that a class supports an interface makes it easy to see
what a class is intended for.  It also makes it possible to do proper type
checking.  When an interface is changed any class that declares to implement
it will be checked if that change was also changed.  The mechanism to assume a
class implements an interface just because the methods happen to match is
brittle and leads to obscure problems, let's not do that.
```

mg979:
What are called `interfaces` in vim9 aren't interfaces, they're classes, but you can have classes implement lots of them, as if they were interfaces. This leads to potentially infinite inheritance.

Right now there isn't a problem that interfaces can solve, that cannot be solved with abstract classes and regular inheritance.
This kind of 'interfaces' are only making the picture infinitely more complex.

If you check golang interfaces, they work like this. If you check c# or java interfaces, they also work like this, only with additional support for static members, that makes already little sense to me because they are like globals that could stay anywhere else.

mg979:
```
Explicitly declaring that a class supports an interface makes it easy to see
what a class is intended for.  It also makes it possible to do proper type
checking.  When an interface is changed any class that declares to implement
it will be checked if that change was also changed.  The mechanism to assume a
class implements an interface just because the methods happen to match is
brittle and leads to obscure problems, let's not do that.
```

This can be done with regular inheritance as well, especially with abstract classes.

mg979:
You can also take a look at how abstract classes are implemented in [c#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/abstract). In vim9script they __prevent__ redefining the same member/method, while an abstract class should only __declare__ a member/method and then __require__ its definition in the inheriting class.
Basically it's how interfaces are working right now, except that interfaces should be different, what interfaces currently are should be abstract classes instead.

This would solve the refactor/error check problem mentioned in that document. So what I think should be done is:
- make interfaces behave like interfaces (this PR)
- turn the abstract classes into what they should be (like interfaces are now)

The big difference is that a class can extend a single class, while it can implement multiple interfaces.

I hope you'll agree that the class specification, as it was designed, has flaws that need to be addressed. For me these are the simplest ways, and require very little effort at this point to get them right:
- interfaces like in go (this PR)
- abstract classes like in c# (to be done but should be easy)

dkearns:
The aim, as I understand it, was to essentially have something close to Dart's implementation.

I agree having both interfaces and "specifying" abstract classes almost represents a distinction without a difference.  So, why not get rid of interfaces, specifying concrete classes are useful.

Isn't your desired interface behaviour provided for by creating a completely deferred specifying abstract class without fields.  If you don't want fields don't include them.

mg979:
@dkearns 
The main concept is that interfaces aren't a form of inheritance, because they hold no data that is tied to the object, so you can safely have a class implement many of them without leading to complex inheritance situations, but you can still use the interface as a type, to act as a glue between objects of different types. Instead a class can inherit only from one class, without interfaces it's very hard to make objects of unrelated classes work together.

dkearns:
We might be talking past each other but I simply meant that once implemented the following would be functionally equivalent with regards to interfaces.

```vim
interface IA
  def Meth1(): void
endinterface
```
and

```vim
abstract class A specifies IA
  abstract def Meth1(): void
endclass
```

The only overarching difference being that A could also be extended but that could also be prevented by sealing it if desired.

I'm finding this difficult to discuss because it's entirely unclear to me what's the result of incomplete design, partial or buggy implementation or novel ideas.  I have no idea where the lines are.

E.g., is it intentional that abstract method declarations are neither implemented nor documented?  That was the first thing I tried and it doesn't work but seems an unlikely oversight.

mg979:
> I'm finding this difficult to discuss because it's entirely unclear to me what's the result of incomplete design, partial or buggy implementation or novel ideas

The vim9 class design is all upside down. You mentioned the Dart language, but there (checking their documentation) the abstract class works like in any other language: you _declare_ members, that must be _defined_ in the inheriting class. Sort of how interfaces are currently working. The way abstract classes work in vim9script isn't seen anywhere. So now in vim9script you have _on paper_ single inheritance, but _in practice_ you have infinite inheritance through misconceived interfaces (that work like real classes).

Now it seems @yegappan can't see how Bram's class design was wrong and wants to merge that #13007 monstrosity, I guess there's no hope for vim9script, I might as well waste my time in some other way.

dkearns:
> - You mentioned the Dart language, but there (checking their documentation) the abstract class works like in any other language: you declare members, that must be defined in the inheriting class.

I didn't mean to suggest otherwise.  I was just referring to their [use in defining interfaces](https://dart.dev/language/class-modifiers#abstract-interface).

> I guess there's no hope for vim9script, I might as well waste my time in some other way.

You have to give some thought to the context of the current development effort and assume that there's an attempt to implement Bram's initial 'documented' design as a matter of respect.

Spare yourself the context shift and continue wasting your time here. :)

PR #13012, opened 2023-09-02
Add CI check to make sure documentation tags are up to date

ychin:
Also, re-generate the tags file with updated list so it will pass CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,6 +242,12 @@ jobs:
           do_test() { sg audio "sg $(id -gn) '$*'"; }
           do_test make ${SHADOWOPT} ${TEST}
 
+      - name: Vim tags
+        if: contains(matrix.extra, 'vimtags')
+        run: |
+          # This will exit with an error code if the generated vim tags differs from source.
+          git diff --exit-code -- runtime/doc/tags
+
       - name: Generate gcov files
         if: matrix.coverage
         run: |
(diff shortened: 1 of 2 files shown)


ychin:
If you only commit the CI changes, you can see that the CI check will correctly fail the build (https://github.com/ychin/vim/actions/runs/6057601828/job/16438921630):

```

Run git diff --exit-code -- runtime/doc/tags
diff --git a/runtime/doc/tags b/runtime/doc/tags
index 8ee1702..b390ef4 100644
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -4442,6 +4442,7 @@ E1362	vim9class.txt	/*E1362*
 E1363	vim9class.txt	/*E1363*
 E1364	recover.txt	/*E1364*
 E137	starting.txt	/*E137*
+E1370	vim9class.txt	/*E1370*
 E138	starting.txt	/*E138*
 E139	message.txt	/*E139*
 E140	message.txt	/*E140*
@@ -8031,6 +8032,7 @@ i_CTRL-O	insert.txt	/*i_CTRL-O*
 i_CTRL-P	insert.txt	/*i_CTRL-P*
 i_CTRL-Q	insert.txt	/*i_CTRL-Q*
 i_CTRL-R	insert.txt	/*i_CTRL-R*
+i_CTRL-R_-	insert.txt	/*i_CTRL-R_-*
 i_CTRL-R_=	insert.txt	/*i_CTRL-R_=*
 i_CTRL-R_CTRL-O	insert.txt	/*i_CTRL-R_CTRL-O*
 i_CTRL-R_CTRL-P	insert.txt	/*i_CTRL-R_CTRL-P*
Error: Process completed with exit code 1.
```

With the correct tags, it will pass the check: https://github.com/ychin/vim/actions/runs/6057612129/job/16438942288

This will make sure we don't forget to update help tags, which I find is very easy to miss in a PR or any commit otherwise. Another solution is to simply build the help tags during the build step but I don't think Vim is currently set up to do that as runtime is a separate component from the binary.

PR #13014, opened 2023-09-02
fixup: comparison of integers of different signs

pheiduck:


chrisbra:
I think I just fixed that with commit 15cbaae31382f17c96a4519e334dd7ba4a90ecba :)

PR #13020, opened 2023-09-03
Fix wrong format specifiers in e_aptypes_is_null_str_nr

zeertzjq:


cvwillegen:
Whoops! But: I actually _wanted_ the %d as the first specifier in the format string (it's easier to see it, because the string might be very long), but the arguments in the siemg() should have been the other way around. 

Thanks for spotting this!

PR #13023, opened 2023-09-04
Update Serbian messages translation

eevan78:


cvwillegen:
Hi, 

That's weird:


+#, c-format
+msgid "E1406: Member \"%s\": type mismatch, expected %s but got %s"
+msgstr "E1406: Члан „%d”: неодговарајући тип, очекује се %s али је наведено %s"

That first %d should have been a %s, but doesn't the script that checks messages complain about this? Or does it only check for the number of arguments??

Christ van Willegen

eevan78:
Hi,
You're right, I spotted this one, and I thought that I made it right,
changed to %s. But it ended up wrong.

I think that check script is not veryfing these arguments...

I'll change this, this is for reporting.

пон, 4. сеп 2023. 13:46 cvwillegen ***@***.***> је написао/ла:

chrisbra:
thanks!

chrisbra:
included as of e4ba5df7c7359742999263115b2faeff94f5a35c

PR #13024, opened 2023-09-04
undo is synced after character find

### LINKED ISSUE #13022: Undoing :%norm is an undo entry per line instead of one undo entry in total
### Steps to reproduce

1. Open vim to a buffer containing two lines `asa` and `bsb`: `vim -u NONE -U NONE +'norm iasa' +'norm obsb'`
2. `:%norm dfs` - removes "as" and "bs" from the two lines
3. Press u to undo the `:%norm`

### Expected behaviour

Buffer is reverted back to the initial state, containing two lines `asa` `bsb`

Actual behavior: Only the change to the last line is undone, so the buffer contains two lines `a` `bsb`

I've bisected the bug. It's introduced in commit bacc83009bc38c9ba0247aaa22b76d1993d57993 (@zeertzjq authored and @chrisbra committed) patch 9.0.1694: wrong mapping applied when replaying a char search

### Version of Vim

9.0.1736

### Environment

OS: Arch Linux
Terminal: GNOME Terminal
TERM: xterm-256color
shell: GNU bash 5.1.16

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Undo is synced after character find.
Solution: Set `no_u_sync` when calling `gotchars_nop()`.

Fix #13022

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -570,7 +570,9 @@ normal_cmd_get_more_chars(
 	    // but when replaying a recording the next key is already in the
 	    // typeahead buffer, so record a <Nop> before that to prevent the
 	    // vpeekc() above from applying wrong mappings when replaying.
+	    ++no_u_sync;
 	    gotchars_nop();
+	    --no_u_sync;
 	}
     }
     --no_mapping;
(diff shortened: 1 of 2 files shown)


chrisbra:
wow that was quick!

PR #13026, opened 2023-09-04
ftplugin ruby: wrong cwd checked

ant0sha:
after ```exe cd .. a:root ..``` we don't need to compare with ```getcwd()```, which is ```a:root``` obviously. Compare with ```cwd``` instead.

Test: 

OK (cwd outside of /bin  folder):

```
cd /tmp; vim tmp.rb +'echo g:ruby_default_path' +q

['/usr/lib64/ruby/site_ruby/3.1.0', '/usr/lib64/ruby/site_ruby/3.1.0/x86_64-linux-gnu', '/usr/lib64/ruby/site_ruby', '/usr/lib64/ruby/vendor_ruby/3.1.0', '/usr/lib64/ruby/ven
dor_ruby/3.1.0/x86_64-linux-gnu', '/usr/lib64/ruby/vendor_ruby', '/usr/lib64/ruby/3.1.0', '/usr/lib64/ruby/3.1.0/x86_64-linux-gnu']
```

NOK (cwd inside /bin folder, should be empty dictionary):

```
cd /bin; vim tmp.rb +'echo g:ruby_default_path' +q

['/usr/lib64/ruby/site_ruby/3.1.0', '/usr/lib64/ruby/site_ruby/3.1.0/x86_64-linux-gnu', '/usr/lib64/ruby/site_ruby', '/usr/lib64/ruby/vendor_ruby/3.1.0', '/usr/lib64/ruby/ven
dor_ruby/3.1.0/x86_64-linux-gnu', '/usr/lib64/ruby/vendor_ruby', '/usr/lib64/ruby/3.1.0', '/usr/lib64/ruby/3.1.0/x86_64-linux-gnu']
```

PR #13027, opened 2023-09-04
Exepath allow curdir if in path

ant0sha:
perl,zip,ruby,... - exec from curdir sometimes ok

perl,gzip,zip,ruby,zig - exec from curdir is sometimes ok

In case the current directory is present as valid $PATH entry, it is OK
to call the program from it, even if vim curdir is in that same
directory.

(Without that patch, for instance, you will not be able to open .zip
files while your current directory is /bin)

(While on that, small bug in ruby.vim was found and submitted as separate
PR #13026)

### DIFF
--- a/runtime/autoload/gzip.vim
+++ b/runtime/autoload/gzip.vim
@@ -11,7 +11,10 @@ fun s:check(cmd)
   let name = substitute(a:cmd, '\(\S*\).*', '\1', '')
   if !exists("s:have_" . name)
     " safety check, don't execute anything from the current directory
-    let f = fnamemodify(exepath(name), ":p:h") !=# getcwd()
+    let s:tmp_cwd = getcwd()
+    let f = (fnamemodify(exepath(name), ":p:h") !=# s:tmp_cwd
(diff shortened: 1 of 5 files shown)


chrisbra:
thanks. I guess this is okay. 

BTW: I am thinking to disable calling the perl/ruby/zig executables by default, only if a user really wants this, to have him set a separate variable. Because I think it may be unexpected for users that those executables are just called whenever the filetype plugin is sourced.

What would you all think here?

ant0sha:
> BTW: I am thinking to disable calling the perl/ruby/zig executables by default, only if a user really wants this, to have him set a separate variable. Because I think it may be unexpected for users that those executables are just called whenever the filetype plugin is sourced.
> 
> What would you all think here?

For the zip/gzip I would somehow naturally expect that external program is called, so they are fine I guess.

Another deal is for perl/ruby/zig - here I would agree - behavior is surprising a bit, would be good to control somehow if that should happen at all. Just need to warn users who expects their ```&path``` for example to contain their ```@INC``` value, during editing perl files, and it will stop to - unless they proactively enable it back.

(we should start dedicated discussion though to attract more attention - such things are easy skipped when asked in random PR comments...)

PR #13031, opened 2023-09-05
Fix "Open &Tab..." CJK translations to have shortcut key

ychin:
Previous PR (#12993) fixed localization files to point to "Open &Tab..." but they didn't add the shortcut key to the translated names. This adds the shortcut keys to the CJK translations in the form of "(&T)".

Note that this doesn't add the shortcut to latin script languages like Czech. These types of translated names tend to also localize the shortcut keys for them to make sense to the user and it's up to each translator to decide how to do so. CJK translations tend to just take the English key directly since it doesn't make sense to have a localized shortcut key in general.


k-takata:
For the Japanese translations, I'm now preparing them in the translation repository:
https://github.com/vim-jp/lang-ja/pull/153

chrisbra:
thanks!

PR #13032, opened 2023-09-05
Minor fixes to inside_class() and object_index_from_itf_index(). More tests.

errael:
Based on tests from Vim9: Class/Object member variable access control #12979

- "Access from child class extending a class:" copied from #12979 converted to test
- `inside_class`: incorporate `class_instance_of` check (fix from yegappan)
- `object_index_from_itf_index`: check super classes while looking for implementing interface

### DIFF
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -237,10 +237,16 @@ object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl,
     if (cl == itf)
 	return idx;
 
-    itf2class_T *i2c;
-    for (i2c = itf->class_itf2class; i2c != NULL; i2c = i2c->i2c_next)
-	if (i2c->i2c_class == cl && i2c->i2c_is_method == is_method)
-	    break;
+    itf2class_T *i2c = NULL;
+    int searching = TRUE;
+    for (class_T *super = cl; super != NULL && searching;
+						super = super->class_extends)
+	for (i2c = itf->class_itf2class; i2c != NULL; i2c = i2c->i2c_next)
+	    if (i2c->i2c_class == super && i2c->i2c_is_method == is_method)
+	    {
+		searching = FALSE;
+		break;
+	    }
     if (i2c == NULL)
     {
 	siemsg("class %s not found on interface %s",
@@ -1940,7 +1946,8 @@ class_member_index(char_u *name, size_t len, class_T **cl_ret, cctx_T *cctx)
 inside_class(cctx_T *cctx_arg, class_T *cl)
 {
     for (cctx_T *cctx = cctx_arg; cctx != NULL; cctx = cctx->ctx_outer)
-	if (cctx->ctx_ufunc != NULL && cctx->ctx_ufunc->uf_class == cl)
+	if (cctx->ctx_ufunc != NULL
+			&& class_instance_of(cctx->ctx_ufunc->uf_class, cl))
 	    return TRUE;
     return FALSE;
 }
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -933,6 +933,50 @@ def Test_class_object_member_access()
     endclass
   END
   v9.CheckScriptFailure(lines, 'E1368: Static cannot be followed by "this" in a member name')
+
+  # Access from child class extending a class:
+  lines =<< trim END
+      vim9script
+      class A
+        this.ro_obj_var = 10
+        public this.rw_obj_var = 20
+        this._priv_obj_var = 30
+
+        static ro_class_var = 40
+        public static rw_class_var = 50
+        static _priv_class_var = 60
+      endclass
+
+      class B extends A
+        def Foo()
+          var x: number
+          x = this.ro_obj_var
+          this.ro_obj_var = 0
+          x = this.rw_obj_var
+          this.rw_obj_var = 0
+          x = this._priv_obj_var
+          this._priv_obj_var = 0
+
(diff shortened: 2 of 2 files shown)


errael:
@yegappan I'm going through the discussion #12979 and modifying the tests where the `:def` functions take an argument; in particular for _accessing a static through an object/interface variable rather than through the class name_. Along the way I ran into a problem with one of the cases, see the commit for this draft PR. Seeing
```
E46: Cannot change read-only variable "ro_class_var"
```
Coming from https://github.com/vim/vim/blob/59adcb4c203371258c72e76011da93eac8540317/src/vim9compile.c#L1886-L1893

I guess the check needs to be like "current_inside_class instanceof cl" or somesuch. Does this make sense? I found class_instance_of; there may be better way.

errael:
> the check needs to be like "current_inside_class instanceof cl" or somesuch. Does this make sense?

@yegappan 
And I guess I can find the current class using code from `inside_class`; although `inside_class` has a loop, wonder why it didn't find it?

Oh well, something to do tomorrow.

yegappan:
Yes. The following diff fixes this issue:

```
diff --git a/src/vim9class.c b/src/vim9class.c
index 30a3b6647..3d07d50b8 100644
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -1940,7 +1940,8 @@ class_member_index(char_u *name, size_t len, class_T **cl_ret, cctx_T *cctx)
 inside_class(cctx_T *cctx_arg, class_T *cl)
 {
     for (cctx_T *cctx = cctx_arg; cctx != NULL; cctx = cctx->ctx_outer)
-       if (cctx->ctx_ufunc != NULL && cctx->ctx_ufunc->uf_class == cl)
+       if (cctx->ctx_ufunc != NULL
+               && class_instance_of(cctx->ctx_ufunc->uf_class, cl))
            return TRUE;
     return FALSE;
 }
```

This needs additional tests (e.g for the interface loop in the class_instance_of() function)

errael:
> The following diff fixes this issue

Cool, a fix in `inside_class` is most convenient. So you want this incorporated into this PR? 

I'm not sure exactly what you have in mind
> This needs additional tests (e.g for the interface loop in the class_instance_of() function)

I haven't looked at the code yet. I'm guessing you're talking about `C extends B extends A` and each of A,B,C might implement a few interfaces; then check that C can access members and static from inside the class, outside the class, when an object of C is passed to a function that takes an interface as an argument.

Some of this is probably tested. Any suggestions to help me focus on what needs testing. I'll take a look at class_instance_of implementation and see what I can discern. Anything else about this?

yegappan:
> > The following diff fixes this issue
> 
> Cool, a fix in `inside_class` is most convenient. So you want this
> incorporated into this PR?
>

Yes.

> 
> I'm not sure exactly what you have in mind
> 
> > This needs additional tests (e.g for the interface loop in the
> > class_instance_of() function)
> 
> I haven't looked at the code yet. I'm guessing you're talking about `C
> extends B extends A` and each of A,B,C might implement a few interfaces; then
> check that C can access members and static from inside the class, outside the
> class, when an object of C is passed to a function that takes an interface as
> an argument.
>

Yes.  It will be good to have tests for these cases.

> 
> Some of this is probably tested. Any suggestions to help me focus on what
> needs testing. I'll take a look at class_instance_of implementation and see
> what I can discern. Anything else about this?
>

I usually look at the code coverage results at
https://app.codecov.io/gh/vim/vim.  I look at an individual source file (e.g.
vim9class.c) to see the lines that are currently tested and the lines that are
not yet tested and then add tests for those lines.

errael:
@yegappan Added a 2nd fix to vim9class: problem when class extending class implementing interface.

At first, I didn't think it would work for statics; but couldn't break it. Finally deciding that yes it would (don't think I was just fooling myself).

errael:
Note: code coverage "failure" seems due to bug in codecov.

errael:
Change state to "Ready for Review". Further fixes in new PR.

src/vim9class.c
```diff
@@ -237,10 +237,14 @@ object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl,
...
-    itf2class_T *i2c;
-    for (i2c = itf->class_itf2class; i2c != NULL; i2c = i2c->i2c_next)
-	if (i2c->i2c_class == cl && i2c->i2c_is_method == is_method)
-	    break;
+    itf2class_T *i2c = NULL;
+    for (class_T *super = cl; super != NULL; super = super->class_extends)
+	for (i2c = itf->class_itf2class; i2c != NULL; i2c = i2c->i2c_next)
+	    if (i2c->i2c_class == super && i2c->i2c_is_method == is_method)
+	    {
+		goto found_class_with_interface;
```
yegappan: Instead of a "goto" here, can you use the "break" statement and remove the curly braces?

yegappan (changes requested):

src/vim9class.c
errael: It's a nested for loop. Has "C" learned how to do a break to label?

The reason for the curly braces is that there's a warning without them

yegappan: I overlooked the nested loop.  The other approach is to use a "done" or "success" flag and break out of the nested loops.
This is used in a few places.  The "goto" is used mostly for doing cleanups on error.

errael: OK, I guess the complexity and extra instructions to exit the loop wont hurt much.

yegappan (approved):

PR #13033, opened 2023-09-05
Vim9: skip newline chars

### LINKED ISSUE #13015: cannot split list items on multiple lines in command block
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        command Cmd {
            echo [1,
            2]
        }
        Cmd
    EOF
    )

This error is given:

    E1069: White space required after ',': ,^@        2]^@    }

**Expected behavior**

No error is given.  `[1, 2]` is echo'ed.

**Version of Vim**

9.0 Included patches: 1-1853

**Environment**

Terminal: XTerm(353)

**Additional context**

No issue if the block is executed without `:command`:
```vim
vim9script
{
    echo [1,
    2]
}
```
    [1, 2]

---

Same issue with a function call:
```vim
vim9script
command Cmd {
    and(0x80,
    0x80)
}
Cmd
```
    E1069: White space required after ',': ,^@    0x80)^@}
    E116: Invalid arguments for function and(0x80,^@    0x80)^@}

---

I looked at the [issues](https://github.com/vim/vim/issues?page=1&q=is%3Aissue+is%3Aopen+created%3A%3E2021-07-26+command+block) opened after [8.2.3228](https://github.com/vim/vim/releases/tag/v8.2.3228), but couldn't find any relevant one.  The closest one is https://github.com/vim/vim/issues/9352. But the latter is about the closing bracket at the end of a dictionary; not about items in the middle of a list.

---

It *might* be impossible to fix:

   > Parsing the lines inside the block would require rewriting the whole
   > command parsing code, that is infeasible.

[source](https://github.com/vim/vim/issues/9352#issuecomment-994997867)

---

If this is working as intended, or impossible to fix, giving a more meaningful error would help.  From the user's perspective, and contrary to what `E1069` suggests, they did not forget a whitespace after the comma.  The issue is that Vim can't or doesn't want to skip the newline before checking the presence of the mandatory whitespace.

If giving a more meaningful error can't be done either, maybe this limitation should be documented.  I couldn't find anything relevant at [`:help :command-repl`](https://vimhelp.org/map.txt.html#%3Acommand-repl) nor at [`:help command-block`](https://vimhelp.org/vim9.txt.html#command-block).  The former tag mentions these limitations:

   1. there must be white space before the "{"
   2. no nesting is supported
   3. inline functions cannot be used
   4. commands where a "|" may appear in the argument, such as commands with an expression argument, cannot be followed by a "|" and another command

`1.` doesn't apply here: there is whitespace before `{`:

        command Cmd {
                   ^

`2.` doesn't apply either; it's about nesting blocks; there is only one block here.
Nor does `3.`; there is no inline function as documented at [`:help inline-function`](https://vimhelp.org/vim9.txt.html#inline-function).
And nor does `4.`; there is no bar in the block.

The latter help tag only mentions that if the block contains a dictionary, its closing bracket must not be written at the start of a line.  But there is no dictionary here; only a list.

Anyway, as a suggestion, here is a patch:
```diff
diff --git a/runtime/doc/map.txt b/runtime/doc/map.txt
index 1dcd99bf9..a09472045 100644
--- a/runtime/doc/map.txt
+++ b/runtime/doc/map.txt
@@ -1777,10 +1777,11 @@ Example: >
 		g:calledMyCommand = true
 	    }
 <							*E1231*
-There must be white space before the "{".  No nesting is supported, inline
-functions cannot be used.  Commands where a "|" may appear in the argument,
-such as commands with an expression argument, cannot be followed by a "|" and
-another command.
+There must be white space before the "{".  No nesting is supported; inline
+functions cannot be used; arguments in function calls, and items in lists or
+dictionaries cannot be written on multiple lines.  Commands where a "|" may
+appear in the argument, such as commands with an expression argument, cannot be
+followed by a "|" and another command.
 
 If the command is defined in Vim9 script (a script that starts with
 `:vim9script` and in a `:def` function) then {repl} will be executed as in Vim9
```
---

BTW, even though that's out of scope for this issue, I would also suggest trying to merge the previously mentioned help tags, because currently none of them seem to give an exhaustive list of the limitations.

chrisbra:
fixes: #13015

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -3954,7 +3954,7 @@ find_ex_command(
 #ifdef FEAT_EVAL
     if (eap->cmdidx < CMD_SIZE
 	    && vim9
-	    && !IS_WHITE_OR_NUL(*p) && *p != '\n' && *p != '!' && *p != '|'
+	    && !IS_WHITE_NL_OR_NUL(*p) && *p != '!' && *p != '|'
 	    && (eap->cmdidx < 0 ||
 		(cmdnames[eap->cmdidx].cmd_argt & EX_NONWHITE_OK) == 0))
     {
--- a/src/list.c
+++ b/src/list.c
@@ -1592,7 +1592,7 @@ eval_list(char_u **arg, typval_T *rettv, evalarg_T *evalarg, int do_error)
 	had_comma = **arg == ',';
 	if (had_comma)
 	{
-	    if (vim9script && !IS_WHITE_OR_NUL((*arg)[1]) && (*arg)[1] != ']')
+	    if (vim9script && !IS_WHITE_NL_OR_NUL((*arg)[1]) && (*arg)[1] != ']')
 	    {
 		semsg(_(e_white_space_required_after_str_str), ",", *arg);
 		goto failret;
--- a/src/macros.h
+++ b/src/macros.h
@@ -43,6 +43,7 @@
  */
 #define VIM_ISWHITE(x)		((x) == ' ' || (x) == '\t')
 #define IS_WHITE_OR_NUL(x)	((x) == ' ' || (x) == '\t' || (x) == NUL)
+#define IS_WHITE_NL_OR_NUL(x)	((x) == ' ' || (x) == '\t' || (x) == '\n' || (x) == NUL)
 
 /*
  * LINEEMPTY() - return TRUE if the line is empty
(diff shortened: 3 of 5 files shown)


chrisbra:
fixed by 00cb247465856eaa546c520b65cf0ccc753ee1cd

PR #13036, opened 2023-09-05
Filelist: Add missing directory `crash`

zdohnal:
Test_crash_1 test fails in Fedora, because the directory ```crash``` is not in Filelist (we use ```make unixall``` to generate source tarball).


chrisbra:
thanks for noticing.

PR #13037, opened 2023-09-05
Add a Vim9 script disassemble test for an interface with static members

yegappan:


errael:
I never learned vim's regex flavor in any detail. When I saw the `\_s*` yesterday, I had no idea what was going on. I dug a little deeper after seeing thisPR , finally found it, and had to read it a few times before I caught that `\_` means include new line.

Is there a toolkit around to assist in tasks like converting `disassemble` output to a test? I wouldn't be surprised if there's a slew of things like that; maybe hiding on Bram's system. Speaking of which, there's the convention where follow on lines of function args are shifted to the right until they almost bump into the end. Is there a format/indent setting/option for that?

PR #13040, opened 2023-09-05
test_xxd.vim: flaky on some distros

pheiduck:
Compiling VIM for ArchLinux and other see: https://github.com/vim/vim/pull/13040#issuecomment-1707772164
leads to flaky test on `Test_xxd_color2()`
see: https://gitlab.archlinux.org/archlinux/packaging/packages/vim/-/commit/30422a45a0d674e9883921b271cba0b367ca790d
Workaround with this Patch the Test will be skipped, but `make test` will pass then.

### DIFF
--- a/src/testdir/check.vim
+++ b/src/testdir/check.vim
@@ -103,11 +103,11 @@ func CheckLinux()
   endif
 endfunc
 
-" Command to check for not running on Linux
-command CheckNotLinux call CheckNotLinux()
-func CheckNotLinux()
-  if has('linux')
-    throw 'Skipped: does not work on Linux'
+" Command to check for dash is not executable
+command CheckDash call CheckDash()
+func CheckDash()
+  if !executable('dash')
+    throw 'Skipped: dash not executable!'
   endif
 endfunc
 
--- a/src/testdir/check.vim
+++ b/src/testdir/check.vim
@@ -103,6 +103,14 @@ func CheckLinux()
   endif
 endfunc
 
+" Command to check for not running on Linux
+command CheckNotLinux call CheckNotLinux()
+func CheckNotLinux()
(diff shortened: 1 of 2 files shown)


pheiduck:
Is it possible to use `if has('ArchLinux')` so these test is not skipped on all Linux Systems?¶

zdohnal:
This test fails on Fedora/Centos Stream/RHEL as well :( . What is interesting, this test does not fail in Github CI (where Ubuntu is used) - see the [link](https://pipelines.actions.githubusercontent.com/serviceHosts/92f62718-4ec0-49ac-808e-b8a9321a7f5d/_apis/pipelines/1/runs/25658/signedlogcontent/8?urlExpires=2023-09-06T06%3A35%3A39.2926260Z&urlSigningMethod=HMACV1&urlSignature=94H8tR8uJLxePOxeWksPvSmYWvx%2F6aWrJBXT8qR88tc%3D) and search for Test_xxd_color2.

But if you look further, you don't see any call for ```xxd -R always``` where we found one difference in our failing [tests](https://www.diffchecker.com/uRcwTXr2/) (the other difference is change of number on the first line)...

chrisbra:
yes, it seems an issue with the shell and I haven't found a way to make sure the prompt and command is always hidden.

chrisbra:
if we apply this patch, I am afraid it won't run anywhere :(

chrisbra:
perhaps this works instead?
```
  if !executable('dash')
    throw 'Skipped: dash not executable!'
  endif

```

pheiduck:
@chrisbra I will try this later.

zdohnal:
@pheiduck @chrisbra I've verified the fix, the tests pass now. Thank you!

PR #13042, opened 2023-09-06
Incorrect duplicate class member detection when variable names have the same prefix

yegappan:
Using the variable name start pointer and end pointer in STRNCMP() to compare the variable names leads to incorrect duplicate member name detection.  Copy the variable name to temporary buffer and do the string comparison.

Also contains the fix to support accessing a class member variable using an object and coding style fixes.

### DIFF
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -467,8 +467,8 @@ validate_interface_methods(
 						: objmethods_gap->ga_len;
 	for (int if_i = 0; if_i < if_count; ++if_i)
 	{
-	    char_u *if_name = if_fp[if_i]->uf_name;
-	    int cl_i;
+	    char_u	*if_name = if_fp[if_i]->uf_name;
+	    int		cl_i;
 	    for (cl_i = 0; cl_i < cl_count; ++cl_i)
 	    {
 		char_u *cl_name = cl_fp[cl_i]->uf_name;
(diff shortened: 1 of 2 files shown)


src/vim9class.c
```diff
@@ -610,22 +610,24 @@ check_func_arg_names(
...
-    char_u *pstr = (*varname == '_') ? varname + 1 : varname;
+    char_u	*name = vim_strnsave(varname, varname_end - varname);
+    char_u	*pstr = (*name == '_') ? name + 1 : name;
+    int		dup = FALSE;
 
     for (int i = 0; i < mgap->ga_len; ++i)
     {
 	ocmember_T *m = ((ocmember_T *)mgap->ga_data) + i;
 	char_u *qstr = *m->ocm_name == '_' ? m->ocm_name + 1 : m->ocm_name;
-	if (STRNCMP(pstr, qstr, varname_end - pstr) == 0)
```
errael: The pattern I usally see is something like
```
if (STRNCMP(pstr, qstr, varname_end - pstr) == 0
        && *(pstr + varname_end - pstr) == NULL)
```
much cheaper.

errael: I didn't remember the pattern quite correctly, a quick look in vim9compile showed several ways to check if the match is good, depending on what is known about the args. The check I showed is definitely not correct.

PR #13045, opened 2023-09-06
Share statics with super class.

errael:
DRAFT. For comment/discussion only.

NOTE: much easier to see with white space hidden

See #13004 for discussion


errael:
No longer relevant.

Issue #13044, opened 2023-09-06
Cannot declare abstract methods


dkearns:
### Steps to reproduce

There's currently no way to declare abstract methods in an abstract class.

This topic doesn't appear to be mentioned in the help, TODO, source or Vim-9.1 milestone issues.

Abstract classes without the ability to declare abstract methods don't seem very useful.

### Expected behaviour

There should be a mechanism to mark a method declaration as abstract.  This should presumably use the same syntax as declarations in interfaces but also be prefixed with the `abstract` modifier.

### Version of Vim

9.0.1876

### Environment

Ubuntu 20.04

yegappan:
Should we support abstract class methods in Vim9?  It looks like some languages don't support it:

Java:
https://stackoverflow.com/questions/370962/why-cant-static-methods-be-abstract-in-java
https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.4.3
https://docs.oracle.com/javase/tutorial/java/IandI/abstract.html

C#:
https://stackoverflow.com/questions/3284/why-cant-i-have-abstract-static-methods-in-c

Typescript:
https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members

yegappan:
With the changes in PR #13046, abstract class methods are supported.  If this is not needed, then changes are needed to
detect this at compile time and fail.

PR #13046, opened 2023-09-07
Add support for abstract methods.

yegappan:
Add support for defining abstract methods in an abstract class (#13044)

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -358,6 +358,16 @@ class, for which objects can be created.  Example: >
 An abstract class is defined the same way as a normal class, except that it
 does not have any new() method. *E1359*
 
+						*abstract-method*
+An abstract method can be defined in an abstract class by using the "abstract"
+prefix when defining the function: >
+
+	abstract class Shape
+	   abstract def Draw()
+	endclass
+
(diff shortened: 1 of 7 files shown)


dkearns:
Thanks @yegappan 

I'll give it a run tomorrow when I'm back at a dev box.

dkearns:
Thanks again, it looks good to me at first blush.

I did, however, notice that an abstract class is not allowed to define constructors which seems odd but maybe it makes sense given the way super-class constructors are, IIRC, entirely inaccessible from the base class.

I'll open a separate issue for that, if required, when I've made more sense of it.

yegappan:
> 
> I did, however, notice that an abstract class is not allowed to define constructors 
> which seems odd but maybe it makes sense given the way super-class constructors
> are, IIRC, entirely inaccessible from the base class.
> 

As you cannot instantiate an abstract class, constructors are not supported in an abstract class.

dkearns:
Right but that is abnormal in terms of the design-inspiring languages.  It would be a natural consequence of this:

> Unlike other languages, the constructor of the base class does not need to be
> invoked.  In fact, it cannot be invoked.  If some initialization from the base
> class also needs to be done in a child class, put it in an object method and
> call that method from every constructor().

PR #13049, opened 2023-09-07
Improve loading a static through an object.

errael:
Generate `LOAD CLASSMEMBER` if possible.
Instead of two instruction `LOAD; OBJ_MEMBER [STATIC]`.

### DIFF
--- a/src/testdir/test_vim9_disassemble.vim
+++ b/src/testdir/test_vim9_disassemble.vim
@@ -3105,26 +3105,26 @@ def Test_disassemble_interface_static_member()
     '11 STORE $0\_s*' ..
     '12 RETURN void\_s*',
     g:instr1)
-  assert_match('<SNR>\d*_F2\_s*' ..
-    'var x: number\_s*' ..
-    'x = o.s_var\_s*' ..
-    '0 LOAD arg\[-1\]\_s*' ..
-    '1 OBJ_MEMBER 0 \[STATIC\]\_s*' ..
-    '2 STORE $0\_s*' ..
-    'x = o.o_var\_s*' ..
-    '3 LOAD arg\[-1\]\_s*' ..
-    '4 OBJ_MEMBER 0\_s*' ..
-    '5 STORE $0\_s*' ..
-    'x = o.s_var2\_s*' ..
-    '6 LOAD arg\[-1\]\_s*' ..
-    '7 OBJ_MEMBER 1 \[STATIC\]\_s*' ..
-    ' 8 STORE $0\_s*' ..
-    'x = o.o_var2\_s*' ..
-    '9 LOAD arg\[-1\]\_s*' ..
-    '10 OBJ_MEMBER 1\_s*' ..
-    '11 STORE $0\_s*' ..
-    '12 RETURN void',
-    g:instr2)
+  #assert_match('<SNR>\d*_F2\_s*' ..
+  #  'var x: number\_s*' ..
+  #  'x = o.s_var\_s*' ..
+  #  '0 LOAD arg\[-1\]\_s*' ..
+  #  '1 OBJ_MEMBER 0 \[STATIC\]\_s*' ..
+  #  '2 STORE $0\_s*' ..
+  #  'x = o.o_var\_s*' ..
+  #  '3 LOAD arg\[-1\]\_s*' ..
+  #  '4 OBJ_MEMBER 0\_s*' ..
+  #  '5 STORE $0\_s*' ..
+  #  'x = o.s_var2\_s*' ..
+  #  '6 LOAD arg\[-1\]\_s*' ..
+  #  '7 OBJ_MEMBER 1 \[STATIC\]\_s*' ..
+  #  ' 8 STORE $0\_s*' ..
+  #  'x = o.o_var2\_s*' ..
+  #  '9 LOAD arg\[-1\]\_s*' ..
+  #  '10 OBJ_MEMBER 1\_s*' ..
+  #  '11 STORE $0\_s*' ..
+  #  '12 RETURN void',
+  #  g:instr2)
 
   unlet g:instr1
   unlet g:instr2
--- a/src/vim9expr.c
+++ b/src/vim9expr.c
@@ -640,6 +640,68 @@ generate_funcref(cctx_T *cctx, char_u *name, int has_g_prefix)
 }
 
 /*
+ * Look for ".member", check if a member of the specified class.
+ *
+ * If found, fill in the information about the member and the parsing;
+ * and return TRUE. Some information may be saved even when return FALSE.
+ *
+ * TODO: Some cases may be missed, for example if continues on next line.
+ *
+ * TODO: useful to have char_u **name_start?
+ */
+
+    static int
+find_class_object_member(
+    class_T	*cl,
+    char_u	*check_start,
+    char_u	**name_end,
+    int		*idx,
+    int		*is_class,
+    ocmember_T	**p_m)
+{
+    char_u *name = skipwhite(check_start);
+    if (*name++ != '.')
+	return FALSE;
+    char_u *end = find_name_end(name, NULL, NULL, FNE_CHECK_START);
+    size_t len = end - name;
+    if (len == 0)
+	return FALSE;
+
+    int searching = TRUE;
+    ocmember_T *m = NULL;
+    for (int i = 0; i < cl->class_obj_member_count; ++i)
+    {
+	m = &cl->class_obj_members[i];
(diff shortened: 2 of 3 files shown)


errael:
@yegappan 
As described in https://github.com/vim/vim/discussions/13041#discussioncomment-6932157 and
as a follow up to #13007, the code generated for reading a static through an object can be improved. 

Almost all of the code in #13007 was for handling reading a static through an object declared as an interface, generating instructions like `ITF_MEMBER 1 on I [STATIC]` and that code is still needed.

Now taking a look at writing to statics, which fails with a compilation error.

For
```
def RS(o: C): number
    return o.sval1
enddef
```
Instead of
```
   0 LOAD arg[-1]
   1 OBJ_MEMBER 0 [STATIC]
```
This PR will typically generate
```
   0 LOAD CLASSMEMBER C.sval1
```

errael:
Test to cover an early return.

src/testdir/test_vim9_disassemble.vim
```diff
@@ -3105,26 +3105,26 @@ def Test_disassemble_interface_static_member()
...
+  #  'x = o.s_var2\_s*' ..
+  #  '6 LOAD arg\[-1\]\_s*' ..
+  #  '7 OBJ_MEMBER 1 \[STATIC\]\_s*' ..
+  #  ' 8 STORE $0\_s*' ..
+  #  'x = o.o_var2\_s*' ..
+  #  '9 LOAD arg\[-1\]\_s*' ..
+  #  '10 OBJ_MEMBER 1\_s*' ..
+  #  '11 STORE $0\_s*' ..
+  #  '12 RETURN void',
+  #  g:instr2)
```
yegappan, line 3127: Can this be replaced with another test?  If this is not needed, then this can be removed?

src/vim9expr.c
```diff
@@ -811,7 +873,34 @@ compile_load(
...
+	{
+	    // Optimization: check if can generate LOAD CLASSMEMBER.
+	    int did_classmember = FALSE;
+	    if (type->tt_type == VAR_OBJECT
+		&& (type->tt_class->class_flags & CLASS_INTERFACE) == 0
+		&& !gen_load_outer)
+	    {
+		// Got an object that is not an interface.
+		int is_class;
+		char_u *name_end;
```
yegappan, line 885: The coding convention is to use tabs between the types and the variable names and align the variable names.

src/vim9expr.c
```diff
@@ -811,7 +873,34 @@ compile_load(
...
+		ocmember_T *m;
+		int idx01;
```
yegappan, line 887: Usually numbers are not used in variable names in the Vim source code.

src/vim9expr.c
```diff
@@ -639,6 +639,68 @@ generate_funcref(cctx_T *cctx, char_u *name, int has_g_prefix)
...
+	if (STRNCMP(name, m->ocm_name, len) == 0 && m->ocm_name[len] == NUL)
+	{
+	    *idx = i;
+	    *is_class = FALSE;
+	    searching = FALSE;
+	    break;
+	}
+    }
+    if (searching)
+	for (int i = 0; i < cl->class_class_member_count; ++i)
```
yegappan, line 684: I tend to prefer using curly brace in these cases (for readability).

src/vim9expr.c
```diff
@@ -811,7 +873,34 @@ compile_load(
...
-	    res = generate_LOAD(cctx, ISN_LOAD, idx, NULL, type);
+	{
+	    // Optimization: check if can generate LOAD CLASSMEMBER.
+	    int did_classmember = FALSE;
+	    if (type->tt_type == VAR_OBJECT
+		&& (type->tt_class->class_flags & CLASS_INTERFACE) == 0
+		&& !gen_load_outer)
+	    {
+		// Got an object that is not an interface.
+		int is_class;
```
yegappan, line 884: Can you initialize "is_class" to FALSE here?

src/vim9expr.c
```diff
@@ -811,7 +873,34 @@ compile_load(
...
+		char_u *name_end;
+		ocmember_T *m;
+		int idx01;
+		if (find_class_object_member(type->tt_class, end_arg,
+					    &name_end, &idx01, &is_class, &m)
+		    && is_class
+		    && m->ocm_access != VIM_ACCESS_PRIVATE)
```
yegappan, line 891: For readability reasons, it will be good to separate out the call to the "find_class_object_member()" function to a separate if statement followed by the check for "is_class" and "m->ocm_access".

src/testdir/test_vim9_disassemble.vim
errael: Part of why this is draft, just to get overall feedback. If statics can not be accessed through an instance, then' there's a bunch of cleanup to do.

src/vim9expr.c
```diff
@@ -811,7 +873,34 @@ compile_load(
...
+		{
+		    // Have index of a readable "static" of the class.
+		    if ((res = generate_CLASSMEMBER(cctx, TRUE, type->tt_class,
```
yegappan, line 894: Can you add a disassemble test for this case?

yegappan (changes requested):

src/vim9expr.c
errael: If there's an `index`, and you want a temporary index, how is that typically named? `tindex`?

src/vim9expr.c
errael: Right, I've just been getting my head wrapped around that. I've actually started doing that, some some cruft left around.

errael:
@yegappan At this point, I'm guessing this draft will be closed pending resolution on accessing static through an object. When there's a decision on that, this draft may be closed without merge, otherwise there will be review cleanup.

src/vim9expr.c
yegappan: In this case, can you use 'member_index' or 'cm_idx'?

src/vim9expr.c
```diff
@@ -639,6 +639,68 @@ generate_funcref(cctx_T *cctx, char_u *name, int has_g_prefix)
     return generate_PUSHFUNC(cctx, ufunc->uf_name, ufunc->uf_func_type, TRUE);
 }
 
+/*
+ * Look for ".member", check if a member of the specified class.
+ *
+ * If found, fill in the information about the member and the parsing;
+ * and return TRUE. Some information may be saved even when return FALSE.
+ *
```
errael, line 647: @yegappan This commit is to optimize something that probably doesn't belong in the spec, and will probably be closed without merge.

But the `find_class_object_member()` might be useful. It's like a "peek at next token, return true if a member of the given class". It provides enough input stream info so the caller can consume the token if wanted.

Could get rid of everything in this PR, except for that function; change the title. Then it's a pending PR with something for the toolkit when/if needed. Probably cleaner to just open another PR.

src/vim9expr.c
errael: > The coding convention is to use tabs between the types and the variable names and align the variable names.

Do you mean "hard tabs"? The modeline has `sts=4`. I don't think I know how to enter a hard tab when `sts` is set.

yegappan: I meant aligning the variable names using a soft-tab.

src/testdir/test_vim9_disassemble.vim
yegappan: After patch 9.0.1887, the static member variables and methods can be accessed only using the class name and cannot be
accessed using an instance.

errael: OK. I'll close this. I built this morning and see that `Found errors in Test_disassemble_interface_static_member():` is failing. I'll open a PR. (which may end up being redundant if it's in process elsewhere.

src/vim9expr.c
errael: > In this case, can you use 'member_index' or 'cm_idx'?

Nice thing about idx01, is that it's obviously short term local var. In a large function, and vim has some monsters, that can be useful. Anyway, I got it.

yegappan: Vim source does have quite a few large functions.  Over a period of several years, I have refactored many large functions and files.  But still we have a long way to go in terms of improving the code.

src/testdir/test_vim9_disassemble.vim
yegappan: I didn't notice this test failure.  I was running only the tests in the test_vim9_class.vim file.  Opened PR https://github.com/vim/vim/pull/13064 to fix this issue.

errael:
With #13057, this is not needed.

Issue #8547, opened 2021-07-11
`:TOhtml` syntax not support `hi link`


dantezhu:
**Describe the bug**
`:TOhtml` syntax not support `hi link`

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Run Vim/gVim/MacVim
2. exe `hi! link LineNr Function`
3. exe `:TOhtml`
4.  LineNr 's syntax is not same with Function

**Expected behavior**
as describe above, LineNr's syntax should be same with Function

**Screenshots**

**Environment (please complete the following information):**
 - Vim version 8.2.3072
 - OS:  win/mac
 - Terminal: gui/term

**Additional context**

lacygoill:
Can you test this patch?
```diff
diff --git a/runtime/syntax/2html.vim b/runtime/syntax/2html.vim
index 13504ed71..bb82780c7 100644
--- a/runtime/syntax/2html.vim
+++ b/runtime/syntax/2html.vim
@@ -51,14 +51,14 @@ else
 endif
 
 if !exists('s:FOLDED_ID')
-  let s:FOLDED_ID  = hlID("Folded")     | lockvar s:FOLDED_ID
-  let s:FOLD_C_ID  = hlID("FoldColumn") | lockvar s:FOLD_C_ID
-  let s:LINENR_ID  = hlID('LineNr')     | lockvar s:LINENR_ID
-  let s:DIFF_D_ID  = hlID("DiffDelete") | lockvar s:DIFF_D_ID
-  let s:DIFF_A_ID  = hlID("DiffAdd")    | lockvar s:DIFF_A_ID
-  let s:DIFF_C_ID  = hlID("DiffChange") | lockvar s:DIFF_C_ID
-  let s:DIFF_T_ID  = hlID("DiffText")   | lockvar s:DIFF_T_ID
-  let s:CONCEAL_ID = hlID('Conceal')    | lockvar s:CONCEAL_ID
+  let s:FOLDED_ID  = hlID("Folded")->synIDtrans()     | lockvar s:FOLDED_ID
+  let s:FOLD_C_ID  = hlID("FoldColumn")->synIDtrans() | lockvar s:FOLD_C_ID
+  let s:LINENR_ID  = hlID('LineNr')->synIDtrans()     | lockvar s:LINENR_ID
+  let s:DIFF_D_ID  = hlID("DiffDelete")->synIDtrans() | lockvar s:DIFF_D_ID
[...]

PR #13050, opened 2023-09-08
Update TOhtml to version 9.0v2

fritzophrenic:
Modified behavior:
  - Change default value of g:html_use_input_for_pc from "fallback" to "none". This means with default settings, only the standards-based method to make special text unselectable is used. The old method relying on unspecified browser behavior for <input> tags is now only used if a user specifically enables it.
  - Officially deprecate g:use_xhtml option (in favor of g:html_use_xhtml) by issuing a warning message when used.

Bugfixes:
  - Fix issue #8547: LineNr and other special highlight groups did not get proper style rules defined when using "hi link".
  - Fix that diff filler was not properly added for deleted lines at the end of a buffer.

Other:
  - Refactored function definitions from long lists of strings to use :let-heredoc variable assignment instead.
  - Corrected deprecated "." string concatenation operator to ".." operator in more places.

### DIFF
--- a/runtime/autoload/tohtml.vim
+++ b/runtime/autoload/tohtml.vim
@@ -1,6 +1,6 @@
 " Vim autoload file for the tohtml plugin.
 " Maintainer: Ben Fritz <fritzophrenic@gmail.com>
-" Last Change: 2023 Jan 01
+" Last Change: 2023 Sep 03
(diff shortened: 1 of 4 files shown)


chrisbra:
thank you!

PR #13052, opened 2023-09-08
runtime(perl): Update ftplugin and indent files

### LINKED ISSUE #336: Update vim-perl to be included in Vim 9.1
There's a possibly imminent 9.1 release so we should update the mother ship.

I can do that, if you like, @petdance.

dkearns:
Closes https://github.com/vim-perl/vim-perl/issues/336


chrisbra:
thanks!

PR #13054, opened 2023-09-08
syntax(i3config): improved highlighting

litoj:
This is, in a sense, a followup attempt of #7969 at getting the improved highlighting of `i3`/`sway` config files originally created by @mboughaba merged. I rewrote the vast majority of the code to have better matches and more highlighting, and also full support for sway's block variants of the basic i3 commands.

Though there is some basic highlighting already, it doesn't really help the user verify the correctness of the code. My goal was to make it clear to the user whether the code is correctly written just by highlighting.

### DIFF
--- a/runtime/syntax/i3config.vim
+++ b/runtime/syntax/i3config.vim
@@ -2,8 +2,9 @@
 " Language: i3 config file
 " Original Author: Mohamed Boughaba <mohamed dot bgb at gmail dot com>
 " Maintainer: Quentin Hibon (github user hiqua)
-" Version: 0.4
-" Last Change: 2022 Jun 05
+" Version: 0.4.22
+" Reference version (JosefLitos/i3config.vim): 4.22
+" Last Change: 2023-09-09
 
 " References:
 " http://i3wm.org/docs/userguide.html#configuring
@@ -18,7 +19,7 @@ endif
 scriptencoding utf-8
 
 " Error
-syn match i3ConfigError /.*/
+syn match i3ConfigError /.\+/
 
 " Todo
 syn keyword i3ConfigTodo TODO FIXME XXX contained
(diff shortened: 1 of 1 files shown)


jdholtz:
This works very well for me. One issue I had was that mode names with spaces did not work correctly. I looked into it and improved the regex a bit for mode matching:
```diff
-syn region i3ConfigBlock start=/^\(mode \S\+\|bar\|\s\+colors\)\+\s\+{$/ end=/^\s*}$/ 
+syn region i3ConfigBlock start=/^\(mode \(\--.* \)\?".\+"\|bar\|\s\+colors\)\+\s\+{$/ end=/^\s*}$/ 
```

This regex now requires quotation marks (and at least one character inside) which lines up with a valid i3 config. It also optionally allows the `--pango_markup` flag to be there (I matched with any flag, but it could be changed to just pango_markup as that is the only valid one).

julio-b:
with `vim --clean /etc/i3/config` I'm getting this error:
```
Error detected while processing BufRead Autocommands for "*/i3/config"..FileType Autocommands for "*"..Syntax Autocommands for "*"..function <SNR>6_SynSet[25]..script /home/j/vim/runtime/syntax/i3config.vim:
line  220:
W18: Invalid character in group name
```
also the `set $var` syntax is not recognized:
2023-09-09-011113_903x112_scrot

jdholtz:
Related to julio-b's comment, there are many characters that can be in a variable name that are not being colored correctly, including `:`, `[`, `|`, and `-`.

JosefLitos:
> Related to julio-b's comment, there are many characters that can be in a variable name that are not being colored correctly, including `:`, `[`, `|`, and `-`.

Oh, I didn't know that, will make sure to include this in the syntax. How did you find this out? It wasn't in [i3 variables](https://i3wm.org/docs/userguide.html#variables).

@julio-b I know where this error comes from, I was using neovim treesitter highlight groups for highlighting a parameter, since default vim has no distinction for that kind of highlight. I will probably comment out those lines and leave the highlights to the user.

> This works very well for me. One issue I had was that mode names with spaces did not work correctly. I looked into it and improved the regex a bit for mode matching

I will also fix this, thank you for your suggestion.

jdholtz:
> How did you find this out?

I just tried it out myself (I was also surprised how there were no seemingly no restrictions to the names as well).

I agree that I don’t think it is absolutely necessary to make sure every single possible variable name is highlighted correctly.

JosefLitos:
@jdholtz I resorted to just adding those chars to the possible parts of the name, no pairing verification.

I also went through all the bar options of i3 user guide and improved some misses in `bar` block highlighting.

@hiqua @jamespeapen could you please review the PR? I apologize for creating such a large diff, but even with reordering to match the original, the diff wouldn't have changed much. Also the highlights make more sense this way - order of definition, but I would welcome any suggestions.

hiqua:
@JosefLitos can you split this PR into two, one for sway and one for i3? Given that we have two different syntax files anyway, we might as well review changes independently. And we don't block each other with our reviews.

JosefLitos:
will do

JosefLitos:
I've separated (hopefully) all sway-specific modifications to #13060 and left this PR for `i3config` syntax only.

jdholtz:
This PR also fixes #10508

JosefLitos:
Oh no, I'm so sorry. I had to make some essential changes because of easier compatibility with sway. I still have a tiny fix to do, so I haven't pushed it yet.
Would it be possible to create another another PR with that commit or is there a better way to go around that?

chrisbra:
@JosefLitos no problem, just open a new PR with the additional changes you want to include.

JosefLitos:
ok, thanks

PR #13060, opened 2023-09-09
syntax(swayconfig): improved highlighting

litoj:
This PR relies on matches set by #13054 (i3config highlighting improvement).

The purpose of this PR is to make highlighting of the config file thorough enough to inform the users about the correctness of their config files just by seeing if the text is correctly highlighted. It makes use of more groups to better distinguish different actions, their options and so on.

I would love to share this improvement with everyone using vim and sway, so I welcome any suggestions for changes and improvements.

@jamespeapen could you please take a look at this PR?

### DIFF
--- a/runtime/syntax/swayconfig.vim
+++ b/runtime/syntax/swayconfig.vim
@@ -4,7 +4,7 @@
 " Maintainer: James Eapen <james.eapen@vai.org>
 " Version: 0.2.2
 " Reference version (JosefLitos/i3config.vim): 1.8.1
-" Last Change: 2023-09-09
+" Last Change: 2023-09-12
 
 " References:
 " http://i3wm.org/docs/userguide.html#configuring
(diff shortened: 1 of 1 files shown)


jamespeapen:
I'll test and review this over the next two weeks.
How do you see maintenance going forward since this is a significant rewrite? I was making PRs in the vim repo based on my work and other users PRs in my [swayconfig.vim repo](https://github.com/jamespeapen/swayconfig.vim) (which is sway specific and not necessarily backwards compatible with i3). If all changes here are good, I can deprecate my repo and send users to yours so development can continue and changes sent upstream. I can help with maintaining your repo or just review any PRs you make here.

JosefLitos:
Sure, I would love to. I think it shouldn't be an issue.

jamespeapen:
Looks like you've used the vim highlight groups better than I have. 
An initial check using https://github.com/JosefLitos/i3config.vim/ as a plugin shows some missing (sway config variables is one I noticed) and incorrect highlights. Some of these can be fixed by checking against my test config which has been made sway specific: https://github.com/jamespeapen/swayconfig.vim/blob/master/test.swayconfig. I'll make a list of changes needed.

chrisbra:
why did you change the `OriginalAuthor?` Other than that, it looks okay.

JosefLitos:
> why did you change the `OriginalAuthor?` Other than that, it looks okay.

Because the syntax is based on the code originally writen by the `i3config.vim` plugin author and although this file is completely different, it relies on highlights set by the other PR, which still contains the plugin's roots (like font highlighting).

But regarding this file alone the author probably should be me.

JosefLitos:
> Some of these can be fixed by checking against my test config which has been made sway specific

I've taken a quick look and will check and fix the issues I can find (hopefully) later today. I will also go through all the `man 5 sway` and check all commands listed.

From what I've noticed so far, despite the manpages mentioning your test for `floating enable|disable|toggle` when I actually put this in my real config it is considered an invalid command by `sway` itself.

JosefLitos:
I fixed all the issues. Tomorrow I will try to improve the code structure so it's more coherent and uniform.

JosefLitos:
I've updated the syntax file, the changes now should be final, together with an accompanying `i3config` update PR: #13080.

I removed support for `new_window`/`new_float` since it's marked as deprecated and sway doesn't mention it in the documentation at all.

PR #13061, opened 2023-09-09
Update masm.vim, mainly for AVX-2 and AVX-512

adah1972:


adah1972:
It seems the CI failure is not related to my changes. Let me know if there are any issues in masm.vim.

chrisbra:
thanks!

PR #13063, opened 2023-09-09
Add syntax highlighting for Mojo

### LINKED ISSUE #13062: Add Mojo highlighting
### Problem

Mojo is a new language, a superset of Python. Vim currently doesn't support it.

### Expected behavior

Support syntax highlighting Mojo files.

mahmoudajawad:
Mojo is a new language, a superset of Python. Vim currently doesn't support it. This PR adds necessary basic support.

Closes #13062

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1368,6 +1368,10 @@ au BufNewFile,BufRead *.rc,*.rch
 	\   setf rc |
 	\ endif
 
+" Mojo
+" Mojo files use either .mojo or .🔥 as extension
+au BufNewFile,BufRead *.mojo,*.🔥		setf python
+
 " MuPAD source
 au BufRead,BufNewFile *.mu			setf mupad
 
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1370,7 +1370,7 @@ au BufNewFile,BufRead *.rc,*.rch
 
 " Mojo
 " Mojo files use either .mojo or .🔥 as extension
-au BufNewFile,BufRead *.mojo,*.🔥		setf python
+au BufNewFile,BufRead *.mojo,*.🔥		setf mojo
 
 " MuPAD source
 au BufRead,BufNewFile *.mu			setf mupad
--- a/runtime/syntax/mojo.vim
+++ b/runtime/syntax/mojo.vim
@@ -185,17 +185,20 @@ endif
 " https://docs.modular.com/mojo/lib.html
 "
 if !exists("mojo_no_builtin_highlight")
-  syn keyword mojoBuiltin	Byte ListLiteral slice constrained CoroutineContext
-  syn keyword mojoBuiltin	Coroutine debug_assert DType dtype type invalid bool
-  syn keyword mojoBuiltin	int8 si8 unit8 ui8 int16 si16 unit16 ui16 int32 si32
-  syn keyword mojoBuiltin	uint32 ui32 int64 si64 uint64 ui64 bfloat16 bf16
-  syn keyword mojoBuiltin	float16 f16 float32 f32 float64 f64 index address
-  syn keyword mojoBuiltin	Error FloatLiteral Int put_new_line print 
-  syn keyword mojoBuiltin	print_no_newline len Attr range rebind SIMD Int8 UInt8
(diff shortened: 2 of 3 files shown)


dkearns:
Thanks for the PR.

Could you add filetype detection and related tests to:
https://github.com/vim/vim/blob/master/runtime/filetype.vim
https://github.com/vim/vim/blob/master/src/testdir/test_filetype.vim

Since you can largely just copy the existing Python files it would be nice to add filetype and indent plugins as well at some stage.

runtime/syntax/mojo.vim
```diff
@@ -0,0 +1,313 @@
...
+        \ "\<\d\%(_\=\d\)*\.\%([eE][+-]\=\d\%(_\=\d\)*\)\=[jJ]\=\%(\W\|$\)\@="
+  syn match   mojoNumber
+        \ "\%(^\|\W\)\zs\%(\d\%(_\=\d\)*\)\=\.\d\%(_\=\d\)*\%([eE][+-]\=\d\%(_\=\d\)*\)\=[jJ]\=\>"
+endif
+
+" The built-ins are added in the same order of appearance in Mojo stdlib docs
+" https://docs.modular.com/mojo/lib.html
+"
+if !exists("mojo_no_builtin_highlight")
+  syn keyword mojoBuiltin	Byte ListLiteral slice constrained CoroutineContext
```
dkearns: The types here might be better highlighted using the `Type` highlight group.  See `:help group-name`

runtime/syntax/mojo.vim
```diff
@@ -0,0 +1,313 @@
...
+
+" These keywords are based on Python syntax highlight, and adds to it struct,
+" fn, alias, var, let
+"
+syn keyword mojoStatement	False None True
+syn keyword mojoStatement	as assert break continue del global
+syn keyword mojoStatement	lambda nonlocal pass return with yield
+syn keyword mojoStatement	class def nextgroup=mojoFunction skipwhite
+syn keyword mojoStatement	struct fn nextgroup=mojoFunction skipwhite
+syn keyword mojoIdentifier	alias var let
```
dkearns: This is linked to the Identifier highlight group but that is usually used for the variable name, if at all.  Using Statement for these would be the standard approach.

mahmoudajawad: Are you referring to `alias, var, let`? I found that `typescript` is using `Identifier` and followed that. I got confused myself on this.

runtime/syntax/mojo.vim
mahmoudajawad: If `Type` refers to buil-in type then this is definitely the case. I'm updating this.

mahmoudajawad: More on this.
Some of what I defined here are built-in functions such as `debug_assert`. Should I keep these in a `Builtin` group, while everything else that represents a type is changed to `Type`?

dkearns: Yes, I'd create a mojoType syntax group and leave the rest of the builtins in the existing group.

runtime/syntax/mojo.vim
dkearns: Yes.  Some of the syntax definitions try and classify the keywords in this manner but it's definitely not the norm.

runtime/filetype.vim
```diff
@@ -1368,6 +1368,10 @@ au BufNewFile,BufRead *.rc,*.rch
 	\   setf rc |
 	\ endif
 
+" Mojo
+" Mojo files use either .mojo or .🔥 as extension
+au BufNewFile,BufRead *.mojo,*.🔥		setf python
```
dkearns: This should be `setf mojo`.

dkearns:
I never thought I'd see a fire emoji in filetype.vim but here we are...

dkearns:
LGTM, thanks @mahmoudajawad

chrisbra:
thanks I include it. BTW is this:
```
.🔥
```
a valid filetype extension?

mahmoudajawad:
> thanks I include it. BTW is this:
> 
> ```
> 
> .🔥
> 
> ```
> 
> a valid filetype extension?

It is.

Thanks for merging.

Adamkaram:
lgtm

PR #13066, opened 2023-09-09
confusing error when using .= in compiled functions

### LINKED ISSUE #12972: confusing error when using legacy compound assignment operator `.=` in compiled code
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        def Func()
            var d: dict<string>
            d.k .= ''
        enddef
        defcompile
    EOF
    )

This error is given:

    E1017: Variable already declared: d

**Expected behavior**

A less confusing error is given.  The current one implies that we're trying to re-declare `d`; we're not.  The issue is that `.=` is only valid in legacy Vim script.

**Version of Vim**

9.0 Included patches: 1-1826

**Environment**

Terminal: XTerm(353)

**Additional context**

At the script level, a better error is given:
```vim
vim9script
var d: dict<string>
d.k .= ''
```
    E985: .= is not supported with script version >= 2

---

Same issue if we replace the dot with some punctuation character which doesn't form a valid compound assignment operator with `=`:
```vim
vim9script
def Func()
    var d: dict<string>
    d.k ,= ''
enddef
defcompile
```
    E1017: Variable already declared: d
```vim
vim9script
var d: dict<string>
d.k ,= ''
```
    E488: Trailing characters:  ,= ''

Again, notice how the errors are different between the script level and the `def` function, and how the error is less confusing at the script level.

chrisbra:
closes: #12972

### DIFF
--- a/src/vim9compile.c
+++ b/src/vim9compile.c
@@ -1669,7 +1669,14 @@ compile_lhs(
 	    {
 		if (is_decl)
 		{
-		    semsg(_(e_variable_already_declared_str), lhs->lhs_name);
+		    // if we come here with what looks like an assignment like .=
+		    // but which has been reject by assignment_len() from may_compile_assignment
+		    // give a better error message
+		    char_u *p = skipwhite(lhs->lhs_end);
+		    if (p[0] == '.' && p[1] == '=')
+			emsg(_(e_dot_equal_not_supported_with_script_version_two));
+		    else
+			semsg(_(e_variable_already_declared_str), lhs->lhs_name);
 		    return FAIL;
 		}
 	    }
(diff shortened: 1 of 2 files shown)


src/vim9compile.c
```diff
@@ -1613,7 +1613,16 @@ compile_lhs(
...
 		if (is_decl)
 		{
-		    semsg(_(e_variable_already_declared_str), lhs->lhs_name);
+		    // if we come here with what looks like an assignment like .=
+		    // but which has been reject by assignment_len() from may_compile_assignment
+		    // give a better error message
+		    char_u *p = skipwhite(lhs->lhs_end);
+		    if (p[0] == '.' && p[1] == '=')
+			emsg(_(e_dot_equal_not_supported_with_script_version_two));
+		    else if (!ends_excmd2(lhs->lhs_end, p))
```
chrisbra: I think this may be wrong. I think it will cause to no longer cause the `e_variable_already_declared_str` message again. So I'll remove it again.

PR #13067, opened 2023-09-09
Code for looking up a class/object variable/method is repeated multiple times

yegappan:
Code for looking up a class/object variable/method is repeated multiple times.  Refactor the code to use lookup functions.
This will help in the future to optimize the lookup functions to use a hash table.

This also includes the PR with the fix for the test failure (https://github.com/vim/vim/pull/13064)


errael:
There's a naming inconsistency (though I admit to not like seeing class_class_*). For example, there's
```
int class_member_idx(class_T *cl, char_u *name, size_t namelen);
int class_obj_member_idx(class_T *cl, char_u *name, size_t namelen);
```
Couldn't it be simplified and the 2nd just be
```
int object_member_idx(class_T *cl, char_u *name, size_t namelen);
```
-----
Frequently see something like
```
    if (v_type == VAR_OBJECT)
        om = class_obj_member_lookup(cl, key, p - key, &m_idx);
    else
        om = class_member_lookup(cl, key, p - key, &m_idx);
```
wonder about having single lookup, something like
```
om = member_lookup(cl, name, len, VAR_CLASS, &m_idx)
om = member_lookup(cl, name, len, VAR_OBJECT, &m_idx)
```
Could even have a
```
type struct {
    int         idx;
    ocmember_T  *m;
}
```
which gets filled in, and retrurn OK/FAIL so there's
```
if (member_lookup(cl, name, len, VAR_CLASS, &result)) {...}
```
or really rad, have member_lookup return the struct, it's tiny. then could do
```
int idx = member_lookup(cl, name, len, VAR_CLASS, &result).idx;
```
and if not found, the struct has idx == -1 and m == NULL.  Or assign the returned structure if want both items.

-----

The search I needed for getting the right error message `static` with `object` needed to go up the hierarchy and return the class where the `static` was defined. Oh, well. Whatever it ends up looking like, there's a place to put it now; it can go with these other search functions.

-----

In the future, when passing a name where `len != 0` probably need to build a null terminated string; but maybe not could have a version of hash_hash that takes a length along with the char*. Guess I don't need the comment about using using a stack buffer if len < magic_number, then can avoid malloc/free. I'm really asking: how good/efficient the allocater is these days?

errael:
Not asking for a change, just wanted to throw some stuff out there and see if anything strikes a chord.

src/vim9class.c
```diff
@@ -2041,82 +2030,151 @@ find_class_func(char_u **arg)
...
+	if (namelen)
 	{
-	    if (cl_ret != NULL)
-		*cl_ret = cl;
+	    if (STRNCMP(name, m->ocm_name, namelen) == 0
+		    && m->ocm_name[namelen] == NUL)
 	    return i;
 	}
+	else if (STRCMP(name, m->ocm_name) == 0)
+	    return i;
```
errael, line 2117: Half as many lines
```
if (namelen ? STRNCMP(name, m->ocm_name, namelen) == 0
                                        && m->ocm_name[namelen] == NUL
            : STRCMP(name, m->ocm_name) == 0)
    return i;
```

And why doesn't
```
if (STRCMP(name, m->ocm_name) == 0)
    return i;
```
work? Can the last char in name's buffer be non zero?

src/vim9class.c
```diff
@@ -269,14 +269,10 @@ object_index_from_itf_index(class_T *itf, int is_method, int idx, class_T *cl,
...
-	{
-	    ocmember_T *m = &i2c->i2c_class->class_class_members[i];
-	    if (STRCMP(name, m->ocm_name) == 0)
-	    {
-		return i;
-	    }
-	}
+	int	m_idx = class_member_idx(i2c->i2c_class, name, 0);
+	if (m_idx >= 0)
+	    return m_idx;
```
errael, line 274: In object_index_from_itf_index, the `is_static` param can be removed I think. It was added to support accessing a static with an object. So this branch can go away.

src/vim9class.c
```diff
@@ -2041,82 +2030,151 @@ find_class_func(char_u **arg)
...
-class_method_index(char_u *name, size_t len, class_T **cl_ret, cctx_T *cctx)
+    ocmember_T *
+class_member_lookup(class_T *cl, char_u *name, size_t namelen, int *idx)
 {
-    if (cctx == NULL || cctx->ctx_ufunc == NULL
-					  || cctx->ctx_ufunc->uf_class == NULL)
-	return -1;
-    class_T *cl = cctx->ctx_ufunc->uf_class;
+    *idx = class_member_idx(cl, name, namelen);
+    return *idx >= 0 ? &cl->class_class_members[*idx] : NULL;
```
errael, line 2066: Considering some of the _blue sky_ comments in my original msg, this could be something like
```
member_lookup(cl, name, nlen, VAR_CLASS, &result);
*idx = result.idx;
return result.m;
```
but the the `?:` and array lookup are prety cheap.

src/vim9class.c
errael, line 2128: This one doesn't have `cl_ret` to deal with.

errael: Crap, I put this comment in the wrong place. In this case, `cl_ret`, has to be handled.

src/vim9class.c
errael: Took a quick look. To actually take `is_static` out, would have to start where it's first used and go forward. That can be done as a cleanup, in a post release time frame (assuming accessing static with an object doesn't come back).

src/vim9class.c
yegappan: When comparing the characters, we want to be conservative and compare only the characters that we already know are part of a valid "name".

src/vim9class.c
yegappan: The member_lookup() function will use the class_member_lookup() function, so this cannot call member_lookup().

yegappan:
Thanks for the review.

> There's a naming inconsistency (though I admit to not like seeing class_class_*). For example, there's
> 
> ```
> int class_member_idx(class_T *cl, char_u *name, size_t namelen);
> int class_obj_member_idx(class_T *cl, char_u *name, size_t namelen);
> ```
> 
> Couldn't it be simplified and the 2nd just be
> 
> ```
> int object_member_idx(class_T *cl, char_u *name, size_t namelen);
> ```
> 

Yes.  I made this change.

>
> Frequently see something like
> 
> ```
>     if (v_type == VAR_OBJECT)
>         om = class_obj_member_lookup(cl, key, p - key, &m_idx);
>     else
>         om = class_member_lookup(cl, key, p - key, &m_idx);
> ```
> 
> wonder about having single lookup, something like
> 
> ```
> om = member_lookup(cl, name, len, VAR_CLASS, &m_idx)
> om = member_lookup(cl, name, len, VAR_OBJECT, &m_idx)
> ```
>
Yes.  This will simplify the code.  I added the member_lookup() and method_lookup() functions.
 
> Could even have a
> 
> ```
> type struct {
>     int         idx;
>     ocmember_T  *m;
> }
> ```

All the callers then need to declare a variable of this structure type and use it.  There are callers (e.g. class_member_type()) who need to store the index in some other variable. As there are only two variables, I would like to keep this simple and directly use the variable names.

> 
> which gets filled in, and retrurn OK/FAIL so there's
> 
> ```
> if (member_lookup(cl, name, len, VAR_CLASS, &result)) {...}
> ```
> 
> or really rad, have member_lookup return the struct, it's tiny. then could do
> 
> ```
> int idx = member_lookup(cl, name, len, VAR_CLASS, &result).idx;
> ```
> 
> and if not found, the struct has idx == -1 and m == NULL. Or assign the returned structure if want both items.
> 
>

src/vim9class.c
errael: I wasn't proposing a specific thing or name. Only noticing that there could be a routine at the lowest level that provided both `idx` and `m`, and then the caller does not need to have a conditional and then calculate something that was already produced by the callee . The statement
```
*idx >= 0 ? &cl->class_class_members[*idx] : NULL
```
is redundant.

yegappan: I have updated the PR incorporating some of your review comments.  Please take a look at the latest diff.

src/vim9class.c
errael: I understand your words, but I still don't get it. AFAICT, in this situation, when `len` is non zero, `name[len]` is never a valid identifier character and so will never .... **Oh, wait**, I see now.

If `name` points to the characters "ident # followed by comment", and `ocm_name` is "ident" null terminated, then `STRCMP` fails. `name[len]` is not null because we're parsing on the fly and not tokenizing. Sorry for the noise.

chrisbra:
Thanks!  I enjoy how you are driving the Vim9 class implementation. That is very much appreciated 👍

src/vim9class.c
errael: > take a look at the latest diff.

Guess I'll just pull and take a look there ;-)

PR #13068, opened 2023-09-09
Better error message for accessing static with object.

errael:


errael:
For this script
```
vim9script

class A
    public static svar1 = "A"
endclass

class B extends A
endclass

def FB(o: B): string
    return o.svar1
enddef

echo "FB ob:" FB(B.new())

```

with this PR this message is produced

```
E1410: Cannot access static "B.svar1" with an object
```
The same error is given from script level.


But when you go to B.svar1, you don't see the variable; it's confusing as has been discussed.

Could do something like

```
Cannot access static "B.sval1" with an object; inherited from "class A"
```

And a final decision on static inheritance can affect the message.

If the spec is changed such that the static is shared, then the message should be

```
E1410: Cannot access static "A.svar1" with an object
```

There's a flag in the newly added function `get_member()` to specified whether or not to search the hierarchy and pick up the class where it's originally defined; for shared statics, flip that flag.

There are a variety of test failures from the wrong message; after some feedback, they'll be fixed in this PR.

During review, take a look at all the `TODO:` that are sprinkled in the new code; comment as the spirit moves you.

I just noticed the lookup class/obj refactoring, I'll take a look and may want to change this PR.

errael:
@yegappan I'm wondering about adding a function, like 
```
int check_object_static_access(class_T* cl, char_u* name, size_t len)
```
which returns `OK`/`FAIL`. Does that fit with `vim` style? It's used for both `:def` and `script` level. Where's a good spot for it?

yegappan:
> @yegappan I'm wondering about adding a function, like
> 
> ```
> int check_object_static_access(class_T* cl, char_u* name, size_t len)
> ```
> 
> which might output an error, returns `OK`/`FAIL`. Does that fit with `vim` style? It's used for both `:def` and `script`
> level. Where's a good spot for it?
>

Yes.  I am trying to move most of the Vim9 class/object related code to vim9class.c from other files.  Can you add
this function to the vim9class.c file?

errael:
@yegappan , @dkearns 

The current error message when accessing a static member with an object is like
```
E1410: Cannot access static "B.svar1" with an object
```
I wondering if the following is better
```
E1410: Cannot access a static with an object: "B.svar1"
```

But it's likely neither will be the final message, check out https://github.com/vim/vim/pull/13068#issuecomment-1712658044 for other options and considerations.

This PR is pretty clean now, will probably need some adjustments after review.

yegappan:
> @yegappan , @dkearns
> 
> The current error message when accessing a static member with an object is like
> 
> ```
> E1410: Cannot access static "B.svar1" with an object
> ```
> 
> I wondering if the following is better
> 
> ```
> E1410: Cannot access a static with an object: "B.svar1"
> ```
> 

I like the format of the first error message as the name of the static member "B.svar1" is clear.

>
> But it's likely neither will be the final message, check out [#13068 (comment)](https://github.com/vim/vim/pull/13068#issuecomment-1712658044) for other options and considerations.
> 
> This PR is pretty clean now, will probably need some adjustments after review.

src/vim9class.c
```diff
@@ -1784,6 +1784,76 @@ ex_type(exarg_T *eap UNUSED)
...
+    {
+	*ret_m = found_m;
+	*ret_cl = found_cl;
+	return OK;
+    }
+
+    return FAIL;
+}
+
+    int
```
yegappan, line 1847: Can you add a comment here with the description of this function?

src/vim9class.c
```diff
@@ -1784,6 +1784,76 @@ ex_type(exarg_T *eap UNUSED)
...
+check_object_static_member_access(class_T* cl, char_u* name, size_t len)
+{
+    ocmember_T  *m;
+    class_T	    *cl_dfn;
+
+    if (find_class_member_definition(cl, name, len, &m, &cl_dfn))
```
yegappan, line 1853: As `find_class_member_definition()` function returns OK/FAIL instead of the Boolean TRUE/FALSE value, can you explicitly check for `== OK` here?

src/vim9class.c
```diff
@@ -1784,6 +1784,76 @@ ex_type(exarg_T *eap UNUSED)
...
+//	 be another use for it.
+    static int
+find_class_member_definition(
+    class_T	*cl,
+    char_u	*name,
+    size_t	namelen,
+    ocmember_T	**ret_m,
+    class_T	**ret_cl)
+{
+    int		check_super = FALSE; // If statics are shared, not inherited,
```
yegappan, line 1806: Once we decide static variables are shared or inherited, then we don't need the `check_super` variable?

dkearns:
> @yegappan , @dkearns
> 
> The current error message when accessing a static member with an object is like
> 
> ```
> E1410: Cannot access static "B.svar1" with an object
> ```
> 
> I wondering if the following is better
> 
> ```
> E1410: Cannot access a static with an object: "B.svar1"
> ```

`E1410: Cannot access static member with object: "obj.svar1"`

- referring to static members collectively as "statics" sounds a bit odd to my 'ear'
- shouldn't the subject string be the erroneous code?
- leaving out the "a" is generally in keeping with the prevailing style

We could be here for weeks.

src/vim9class.c
errael: Oh yeah, good idea.

src/vim9class.c
errael: I'll change it to TRUE/FALSE, but I'm currently thinking once the spec/requirements are clear, probably get rid of this function and there will only be `check_object_static_member_access()`.

src/vim9class.c
errael: I think that's true.

yegappan:
Is this PR needed after the recent changes?  Can this be closed?

yegappan:
> 
> What do you think about reviewing and cleaning up messages?

Yes. We should review and remove redundant error codes.  I did some cleanup of the error
codes last week.  We should also try to include the context information (e.g. class name,
member name, etc.) in the error message.

errael:
Some things to consider

- Only vim9script, particularly classes, related.
- Use a different error code where appropriate, like use `E1335` instead `E46`.
- Spec change from `member` to `variable`; for a class `function` to `method`.
  Should messages be changed from `member` to `variable`? Other nomenclature changes?
- Style issues, for example should class/method/variable names be quoted? Or `this one: item`
  Compare `E1335` versus `E46` above. 

I guess consistency was guaranteed by the style guide in Bram's head. 
@chrisbra @yegappan If there's a written guide, I don't know where to find it; it not, someone should write one. Where would it go?

chrisbra:
There is not much unfortunately. I would think it should go into https://github.com/vim/vim/blob/master/runtime/doc/develop.txt

errael:
Closed by recent commits. See #13135 and  #13136 for following up about error messages.

PR #13071, opened 2023-09-11
Fix "below" virtual text not working with 'rightleft'

zeertzjq:
Problem:  "below" virtual text doesn't work with 'rightleft'.
Solution: Use column from right border with 'rightleft'.


chrisbra:
wow, kudos for adding a right-left test!

PR #13073, opened 2023-09-11
(object|class)_(member|method)_idx, use *_lookup.

errael:
Implement (object|class)_(member|method)_lookup as the base code.
Implement `*_idx` using `*_lookup`.
Take advantage of NULL idx ptr where possible.

### DIFF
--- a/src/vim9expr.c
+++ b/src/vim9expr.c
@@ -394,10 +394,10 @@ compile_class_object_index(cctx_T *cctx, char_u **arg, type_T *type)
 
     if (type->tt_type == VAR_OBJECT)
     {
-	int m_idx = object_member_idx(cl, name, len);
+        int m_idx;
+        ocmember_T *m = object_member_lookup(cl, name, len, &m_idx);
 	if (m_idx >= 0)
 	{
-	    ocmember_T *m = &cl->class_obj_members[m_idx];
 	    if (*name == '_' && !inside_class(cctx, cl))
 	    {
 		semsg(_(e_cannot_access_private_member_str), m->ocm_name);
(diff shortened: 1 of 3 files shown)


errael:
@yegappan Didn't get a chance to further comment on #13067 (and I'm glad it was quickly checked in). I took a look at the PR, was curious about what it would be like to put `*_lookup` at the bottom, and `*_idx` on top of it; it was pretty straightforward so I went ahead to take a look at it after going all the way. This PR goes 95% of the way to remove 4 methods.

src/vim9expr.c
```diff
@@ -412,7 +412,7 @@ compile_class_object_index(cctx_T *cctx, char_u **arg, type_T *type)
 	}
 
 	// Could be a function reference: "obj.Func".
-	m_idx = object_method_idx(cl, name, len);
+	object_method_lookup(cl, name, len, &m_idx);
```
yegappan: I think as the return value is ignored, you need to use the `(void)` cast here.  Otherwise static analysis may complain.  This comment also applies to other place where the `object_method_lookup` and `class_method_lookup` are used.

errael: > I think as the return value is ignored, you need to use the `(void)` cast here. Otherwise static analysis may complain. This comment also applies to other place where the `object_method_lookup` and `class_method_lookup` are used.

In the first experiment, any usage of the `*_idx` functions was removed, with the idea that they would be removed; but that doesn't make sense.  The `*_idx` version are convenient and easily remembered with the naming convention.

With this, there is never a need to declare an integer and then ignore its value, both
- `int idx = xxx_idx(...)`
- `m = xxx_lookup(..., NULL)`

can be used. And don't need any cumbersome `(void)`.

yegappan (approved):

chrisbra:
thanks!

PR #13077, opened 2023-09-12
fix(matchparen): do not highlight inactive windows

### LINKED ISSUE #25099: MatchParen highlight is applied to floating windows
### Problem

`MatchParen` highlights are applied in non-active floating windows, such as the one for LSP hover.

### Steps to reproduce

1. Use the tippy-top of nightly Neovim.
2. In a `lua` file write:
```lua
local foo = {
  bar = 0,
}
```
3. Hover over `bar`.
4. Notice how the parentheses around `field` are underlined (the `MatchParen` highlight is being applied) despite my cursor being in an entirely different window:
image

### Expected behavior

For the `MatchParen` highlight to only be applied in active windows and based on the cursor position.

### Neovim version (nvim -v)

v0.10.0-dev-1074+g4258f4d53

No

### Operating system/version

Mac Ventura 13.5.2

### Terminal name/version

kitty 0.28.1

xterm-kitty

### Installation

From source

MariaSolOs:
Fixes https://github.com/neovim/neovim/issues/25099

The issue here is that when `BufWinEnter` is fired, the cursor's position won't correctly indicate whether the cursor is inside the window or not. Instead the plugin should try to match parentheses once the cursor has entered the window with `WinEnter`.

### DIFF
--- a/runtime/plugin/matchparen.vim
+++ b/runtime/plugin/matchparen.vim
@@ -20,7 +20,7 @@ endif
 
 augroup matchparen
   " Replace all matchparen autocommands
-  autocmd! CursorMoved,CursorMovedI,WinEnter,BufWinEnter,WinScrolled * call s:Highlight_Matching_Pair()
+  autocmd! CursorMoved,CursorMovedI,WinEnter,WinScrolled * call s:Highlight_Matching_Pair()
   autocmd! WinLeave,BufLeave * call s:Remove_Matches()
   if exists('##TextChanged')
     autocmd! TextChanged,TextChangedI * call s:Highlight_Matching_Pair()


zeertzjq:
I think this issue can also be reproduced using `win_execute()`
```vim
source $VIMRUNTIME/plugin/matchparen.vim
set hidden
call setline(1, '{}')
enew
new
call win_execute(1000, 'buffer 1')
```

zeertzjq:
And as shown by the test failure, this isn't the correct solution.

MariaSolOs:
@zeertzjq oh well, I tried...

chrisbra:
the `win_execute()` issue can be thought of to be expected however and there is a big warning in the help text, that such things can happen:

https://github.com/vim/vim/blob/596ad66d1ddb742ef349e98eb06b8e4052f68f51/runtime/doc/builtin.txt#L10417-L10423

I don't know about the neovim issue or how this happens. There may be different solutions to that (e.g. use `NoMatchParen()` or do not run autocommands, or whatever other things are possible to block autocommands in Neovim 🤷

PR #13080, opened 2023-09-12
syntax(i3config): structure cleanup

litoj:
This is a cleanup of #13054. The highlights should stay mostly the same, only the structure of the file has been made more uniform and uncluttered. Better comments and order according to the *i3 Guide*.


chrisbra:
Thanks!

PR #13081, opened 2023-09-12
CI: linux 6.2 does not modeprobe snd-dummy

pheiduck:
- downgrade to ubuntu 20.04
- toolchain can be used anyway


pheiduck:
hm... we should stay on 22.04 because of 32bit builds. Should figure out another patch

PR #13086, opened 2023-09-14
Class members are accesible only from the class where they are defined.

yegappan:
Based on the https://github.com/vim/vim/discussions/13004 discussion, the following changes are made:

1. Static variables and methods are accessible only using the class name and inside the class where they are defined.
2. Static variables and methods can be used without the class name in the class where they are defined.
3. Static variables of a super class are not copied to the sub class.
4. A sub class can declare a class variable with the same name as the super class.
5. When a method or member is found during compilation, use more specific error messages.

This aligns the Vim9 class variable/method implementation with the Dart implementation.

Static variables in interfaces are not changed in this PR.

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3411,16 +3411,14 @@ EXTERN char e_public_member_name_cannot_start_with_underscore_str[]
 	INIT(= N_("E1332: Public member name cannot start with underscore: %s"));
 EXTERN char e_cannot_access_private_member_str[]
 	INIT(= N_("E1333: Cannot access private member: %s"));
-EXTERN char e_object_member_not_found_str[]
-	INIT(= N_("E1334: Object member not found: %s"));
 EXTERN char e_member_is_not_writable_str[]
 	INIT(= N_("E1335: Member is not writable: %s"));
 #endif
 EXTERN char e_internal_error_shortmess_too_long[]
 	INIT(= "E1336: Internal error: shortmess too long");
 #ifdef FEAT_EVAL
-EXTERN char e_class_member_not_found_str[]
-	INIT(= N_("E1337: Class member not found: %s"));
+EXTERN char e_class_member_str_not_found_in_class_str[]
+	INIT(= N_("E1337: Class member \"%s\" not found in class \"%s\""));
 EXTERN char e_member_not_found_on_class_str_str[]
 	INIT(= N_("E1338: Member not found on class \"%s\": %s"));
 #endif
@@ -3488,7 +3486,6 @@ EXTERN char e_cannot_access_private_method_str[]
 	INIT(= N_("E1366: Cannot access private method: %s"));
 EXTERN char e_member_str_of_interface_str_has_different_access[]
 	INIT(= N_("E1367: Access level of member \"%s\" of interface \"%s\" is different"));
-
 EXTERN char e_static_cannot_be_followed_by_this[]
 	INIT(= N_("E1368: Static cannot be followed by \"this\" in a member name"));
 EXTERN char e_duplicate_member_str[]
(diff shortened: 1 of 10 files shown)


errael:
Just ran a few quick tests.

C extends B extends A. In class C, using A.svar, can read/write private members of A, tried that because I remembered you could do that before. Some might consider that a bug. Not sure what I think.

Looking good. I'll dig in more tomorrow.

Messages, I'm fine with the way they are, but some things that occurred to me.

1. In C, svar get message ` Class member "svar" accessible only inside class "A"`, 
  wondering about working `use A.svar` and/or `from class "A"` in there somewhere.

2. With `var oa = A.new()` and doing `oa.sval` get `... accessible only using class "A"` 
  wondering about `... only using A.svar"`

3. With `var oc = C.new()` and `oc.sval` get the old favorite `Member not found on object`
wondering, since sval is in C's kind of parentage, if a message like in 1. is useful.

yegappan:
> Just ran a few quick tests.
>

Thanks for testing the PR.

>
> C extends B extends A. In class C, using A.svar, can read/write private members of A, tried that because
> I remembered you could do that before. Some might consider that a bug. Not sure what I think.
>

I have updated the PR with a fix for this.

> 
> Looking good. I'll dig in more tomorrow.
>

errael:
I played around with vim including this PR. While trying stuff, ran into https://github.com/vim/vim/discussions/13041#discussioncomment-7004644; doesn't have anything to do with this PR.

Unless there's something particular you'd like explored, I'll wait for the merge.

vim-ml:
On Thu, Sep 14, 2023 at 10:32 AM errael ***@***.***> wrote:

> I played around with vim including this PR. While trying stuff, ran into #13041
> (comment)
> <https://github.com/vim/vim/discussions/13041#discussioncomment-7004644>;
> doesn't have anything to do with this PR.
>
> Unless there's something particular you'd like explored, I'll wait for the
> merge.
>

Thanks.  Did you see any issues with using class variables and methods in
script context, def function context, derived class, another class, using
an instance, instance method, class method and different access controls?
I know this is a long list. But there are too many scenarios where a class
variable and method can be used.

Thanks,
Yegappan

>

errael:
Duplicate method name for both object/class method not flagged as an error.

This output should be a dup name error, but it runs OK. I started putting the test code here, then checked, it's not caused by this PR.

https://github.com/vim/vim/discussions/13041#discussioncomment-7008328

yegappan:
> Duplicate method name for both object/class method not flagged as an error.
> 
> This output should be a dup name error, but it runs OK. I started putting the test code here, then checked, it's not caused by this PR.
> 

I have updated the PR with a fix for this issue also.

errael:
Did you take a peek at https://github.com/vim/vim/discussions/13041#discussioncomment-7004644? `_F()` in one class and  `F()`  in a child or base class gets no error.

yegappan:
> Did you take a peek at [#13041 (comment)](https://github.com/vim/vim/discussions/13041#discussioncomment-7004644)? `_F()` in one class and `F()` in a child or base class gets no error.

This is now addressed in the updated PR.

yegappan:
> Haven't found any more bugs with this PR (original comment went to wrong spot)

Thanks for testing the PR.

chrisbra:
Coverity complains about this part here:
```
*** CID 1544700:  Control flow issues  (DEADCODE)
/src/vim9class.c: 1114 in add_classfuncs_objmethods()
1108                if (loop == 1 && STRNCMP(
1109                                    extends_cl->class_class_functions[i]->uf_name,
1110                                    "new", 3) == 0)
1111                    ++skipped;
1112                else
1113                {
>>>     CID 1544700:  Control flow issues  (DEADCODE)
>>>     Execution cannot reach the expression "extends_cl->class_class_functions" inside this statement: "pf = (loop == 1) ? extends_...".
1114                    ufunc_T *pf = (loop == 1
1115                                            ? extends_cl->class_class_functions
1116                                            : extends_cl->class_obj_methods)[i];
1117                    (*fup)[gap->ga_len + i - skipped] = copy_function(pf);
1118
1119                    // If the child class overrides a function from the parent
```

vim-ml:
Hi Christian,

On Sat, Sep 16, 2023 at 4:20 AM Christian Brabandt <
***@***.***> wrote:

> Coverity complains about this part here:
>

I have opened the PR https://github.com/vim/vim/pull/13103 to fix this.

Regards,
Yegappan


> *** CID 1544700:  Control flow issues  (DEADCODE)
> /src/vim9class.c: 1114 in add_classfuncs_objmethods()
> 1108                if (loop == 1 && STRNCMP(
> 1109                                    extends_cl->class_class_functions[i]->uf_name,
> 1110                                    "new", 3) == 0)
> 1111                    ++skipped;
> 1112                else
> 1113                {
> >>>     CID 1544700:  Control flow issues  (DEADCODE)
> >>>     Execution cannot reach the expression "extends_cl->class_class_functions" inside this statement: "pf = (loop == 1) ? extends_...".
> 1114                    ufunc_T *pf = (loop == 1
> 1115                                            ? extends_cl->class_class_functions
> 1116                                            : extends_cl->class_obj_methods)[i];
> 1117                    (*fup)[gap->ga_len + i - skipped] = copy_function(pf);
> 1118
> 1119                    // If the child class overrides a function from the parent
>
>
>

PR #13092, opened 2023-09-15
runtime(vim): Highlight all :loadkeymap abbreviations

dkearns:
I forwarded this on to @cecamp a few years ago but seems to have been missed.

It only catches my eye because at least the Czech keymap uses an abbreviation of :loadkeymap that is not highlighted.


chrisbra:
thanks. Let me include this.

PR #13093, opened 2023-09-15
CI: disable snd-dummy until kernel modules updated

pheiduck:
issue for tracking: https://github.com/actions/runner-images/issues/8295


chrisbra:
I suppose this should clear itsself, once ubuntu pushes a new kernel which includes the sound module again. So not sure this is worth it ...

PR #13097, opened 2023-09-16
Fix :Man plugin not working if gdefault is set

ychin:
Fix the issue introduced by #12557. `:substitue` commands in plugins need to take into account whether `gdefault` is set or not because that depends on the user.

### DIFF
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -196,11 +196,7 @@ func dist#man#GetPage(cmdmods, ...)
 
   " Emulate piping the buffer through the "col -b" command.
   " Ref: https://github.com/vim/vim/issues/12301
-  if &gdefault
-    silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//e
-  else
-    silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//ge
-  endif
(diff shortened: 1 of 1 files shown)


ychin:
I guess I don't like mixing vim9 and non-vim9 code. If I'm a casual reader reading this I would actually be a little confused about the sprinkling of vim9 there and not sure what the intention is as this is relying on a somewhat obscure thing that people may not know. I think if the entire thing is in vim9, it would "just work", but I also don't particularly think it's worth it (or should) to change the whole file over to Vim9.

chrisbra:
how about something similar to patch 64dea84bb05a29117:

```patch
diff --git a/runtime/autoload/dist/man.vim b/runtime/autoload/dist/man.vim
index 315636a2e..7f7d13711 100644
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -196,7 +196,7 @@ func dist#man#GetPage(cmdmods, ...)

   " Emulate piping the buffer through the "col -b" command.
   " Ref: https://github.com/vim/vim/issues/12301
-  silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//ge
+  exe 'silent! keepjumps keeppatterns %s/\v(.)\b\ze\1?//e' .. (&gdefault ? '' : 'g')

   if unsetwidth
     let $MANWIDTH = ''
```
(untested)

ychin:
Sure. Just tested and it works as well.

Issue #11307, opened 2022-10-07
Syntax highlighting for Vim script comments is broken


chdiza:
### Steps to reproduce

Make a buffer with &ft=vim and syntax highlighting on.  Make a comment line that also has an additional double-quote in it later, e.g.,:

```vim
" This is what we call " blah
```

### Expected behaviour

That entire line should be colored as a comment.  But only the part from the second quote onward is colored that way.  It looks like `"This is what we call ` has the highlight group "vimIsCommand".

This is a bad regression; Vim didn't formerly do this.  It's common to have lines that have a trailing comment, and then to comment out those very lines later.  In my colorschemes, this makes the `"This is what we call ` part colored the same as the "Normal" group; it defeats the purpose of syntax highlighting if one can't tell comments from non-comments by color.

Bisection reveals that the first bad commit is the following recent runtime file update:
```
f269eabc6c4f5bdcef989cd5b4b95ba8ccaa4d8a is the first bad commit
commit f269eabc6c4f5bdcef989cd5b4b95ba8ccaa4d8a
Author: Bram Moolenaar <Bram@vim.org>
Date:   Mon Oct 3 18:04:35 2022 +0100

    Update runtime files

```

### Version of Vim

9.0.688

### Environment

Ubuntu 20.04, though it also happens on macOS.
Terminal is kitty, or uxterm, doesn't matter
Also happens in gVim.

brammool:
> ### Steps to reproduce
> 
> Make a buffer with &ft=vim and syntax highlighting on.  Make a comment
> line that also has an additional double-quote in it later, e.g.,:
> 
> ```vim
> " This is what we call " blah
> ```
> 
> ### Expected behaviour
> 
> That entire line should be colored as a comment.  But only the part from the second quote onward is colored that way.  It looks like `"This is what we call ` has the highlight group "vimIsCommand".
> 
> This is a bad regression; Vim didn't formerly do this.  It's common to
> have lines that have a trailing comment, and then to comment out those
> very lines later.  In my colorschemes, this makes the `"This is what
> we call ` part colored the same as the "Normal" group; it defeats the
> purpose of syntax highlighting if one can't tell comments from
> non-comments by color.

The problem is that in Vim9 script a string in double quotes is looking
just like a comment in legacy script.  It's very hard to tell them
apart.  For the time being, an unmatched quote in a legacy comment is
considered uncommon, therefore the highlighting as it is now.

A proper solution would know whether the line is in legacy script or
[...]

Issue #9587, opened 2022-01-21
vim9: vim9CommentTitle syntax match is not defined


craigmac:
vim9CommentTitle should be defined right here, but we are missing that one:

https://github.com/vim/vim/blob/e70dd11ef41f69bd5e94f630194e6b3c4f3f2102/runtime/syntax/vim.vim#L652

I don't know much about the complexitites of this so didn't dare touch it.

chrisbra:
ping @cecamp

chrisbra:
Hi @craigmac is this still relevant?

PR #13104, opened 2023-09-16
runtime(vim): Improve Vim9 and legacy-script comment highlighting

### LINKED ISSUE #13047: Alternate file in vimscript occasionally breaks syntax highlighting
### Steps to reproduce

1. Open $VIMRUNTIME/**/defaults.vim and navigate to where the DiffOrig command is defined.
2. In `r ++edit #`, the alternate file `#` and the trailing text is highlighted as a comment, see attached screenshot below.

AlternateBufferSyntaxHighlighting

### Expected behaviour

I expect the use of `#` not to break syntax highlighting in this case.

I'm assuming that highlighting breaks because `comments=sO:# -,mO:#  ,eO:##,:#,sO:" -,mO:"  ,eO:"",:"` is set by the vim.vim ftplugin, most likely due to Vim9Script, but that does not necessarily have to be the root cause as syntax highlighting still works correctly in, for example, `nnoremap <silent> ,b :buffer! #<CR>`. So I guess this is more of a syntax parsing issue than an issue with the way `'comments'` is set. I have not been able to make any progress beyond that and would appreciate it, if someone familiar with the code base could look into this.

### Version of Vim

9.0.1632

### Environment

Operating system: openSUSE Leap

### Logs and stack traces

_No response_

dkearns:
This is a first-pass attempt to limit matching of Vim9 and legacy-script comments to the appropriate syntactic contexts.

Vim9-script comments are highlighted at top level in a Vim9-script file, in all :def functions, and in all :autocmd and :commmand command blocks. Legacy-script comments are highlighted at top level in a legacy script file, in all :func functions and in the Vim9-script preamble before the :vim9script marker command.

Fixes https://github.com/vim/vim/issues/13047, https://github.com/vim/vim/issues/11307 and https://github.com/vim/vim/issues/9587.


chrisbra:
Thanks. Can you please rebase and resolve the existing conflicts?

chrisbra:
thanks.

chrisbra (approved):

habamax:
Hi @zeertzjq, before this improvement I was able to "lower" vimscript syntax noise by substituting `vimCommand` syntax group with minimum set of keywords:

```
syn clear vimCommand
syn keyword vimCommand contained vim9cmd vim9script import autoload export def enddef call function endfunction defer defcompile delfunction return
syn keyword vimCommand contained if else elseif endif
syn keyword vimCommand contained for endfor while endwhile continue break
syn keyword vimCommand contained class endclass interface endinterface enum endenum
syn keyword vimCommand contained throw try endtry catch finally
syn keyword vimCommand contained silent unsilent
syn keyword vimCommand contained public static final const var let unlet
syn keyword vimCommand contained highlight runtime source eval finish verbose command delcommand
```

However now it somehow influences comment highlighting:

image

Before the patch:

image

Could you suggest how can I set my own keywords for `vimCommand`?

habamax:
oops, sorry @zeertzjq, I meant to ping @dkearns

habamax:
If I change `vimPreVim9script` to:
```
syn region vimPreVim9script start="\%^" end="^\s*vim9s\%[cript]\>" contains=@vimLegacyTop,vimComment,vimLineComment keepend
```

adding `keepend` and removing `\zs` -- my changes to `vimCommand` work how it was before:

[asciicast](https://asciinema.org/a/Ljc4wuegLpj8wc9vBCNbsQxXC)

habamax:
> do you do that because of all the false positive matches for commands?

Yes, short abbreviated commands especially.

> I have a fix for that if I can catch enough edge cases not to annoy everyone in the short term

Would be glad to check it out once it is available.

> Could you please try this patch with your additions and check folding works for a legacy header section before vim9script?

Looks like it works:

image

PR #13106, opened 2023-09-16
Remove FEAT_FLOAT as that should always be true

dundargoc:


tonymec:
Always true ? Even on TINY builds, which are compiled without +eval ? Or would FEAT_FLOAT maybe make no difference if FEAT_EVAL is not defined ?

dundargoc:
"Always true" as in it has been removed in 73e28dcc6125f616cf1f2d56443d22428a79e434

PR #13108, opened 2023-09-16
ci: remove freebsd 13.1 and mac m1

dundargoc:
Cirrus ci has started introducing monthly limits. Vim has exceeded the
monthly limit. CI no worky.


pheiduck:
we have to drop mac m1 as well

dundargoc:
done

PR #13111, opened 2023-09-17
Add liquidsoap to filetype.vim

toots:
This PR adds proper filetype detection for liquidsoap script files.

Liquidsoap is scripting language for creating media streams and more. It's got about [1.5k](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.liq) script files on github.

Thanks for your consideration!


chrisbra:
thanks!

PR #13112, opened 2023-09-17
Make r_CTRL-C behave consistent in visual mode

### LINKED ISSUE #13091: Possibly undocumented behavior with replace after visual
### Steps to reproduce

I cannot explain the following state machine which was extracted from interacting with vim.
The machine has been trimmed to focus on the issue reported.
 
filtered

Consider the following sequence starting from `s0` (Normal mode):
1. `<C-v>`
2. `r` (or `gr`)
3. `<C-c>` (ends up in `s29`)
4. `g`
5. `r` (ends up in `s57`)

Once in `s57` every keystroke I tried ended up going back to Normal mode `s0`.

Note: The issue seems to be more than a decade old. Indeed, the diagram at https://rawgit.com/darcyparker/1886716/raw/eab57dfe784f016085251771d65a75a471ca22d4/vimModeStateDiagram.svg described a similar undocumented behavior already in 2012. 

### Expected behaviour

I don't know what is supposed to happen after emitting `<C-c>` from `s29` but state `s62` seems to visit unknown/undocumented parts of the editor. I am not sure if that was ever the intention of the devs to have these "states". The labelling of the nodes using `mode()` is also confusing. Explanations are welcome.

### Version of Vim

9.0.1544

### Environment

Mac OS
Terminal.app

### Logs and stack traces

_No response_

chrisbra:
closes: #13091

in visual mode, r CTRL-C behaves strange in Unix like environments. It seems to end visual mode, but still is waiting for few more chars, however it never seems to replace it by any characters and eventually just returns back into normal mode.

In contrast in Windows GUI mode, r_CTRL-C replaces in the selected area all characters by a literal CTRL-C.

Not sure why it behaves like this. It seems in the Windows GUI, got_int is not set and therefore behaves as if any other normal character has been pressed.

So remove the special casing of what happens when got_int is set and make it always behave like in Windows GUI mode. Add a test to verify it always behaves like replacing in the selected area each selected character by a literal CTRL-C.

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -4799,7 +4799,7 @@ nv_replace(cmdarg_T *cap)
     if (VIsual_active)
     {
 	if (got_int)
-	    reset_VIsual();
+	    got_int = FALSE;
 	if (had_ctrl_v)
 	{
 	    // Use a special (negative) number to make a difference between a
--- a/src/testdir/test_visual.vim
+++ b/src/testdir/test_visual.vim
@@ -1573,4 +1573,18 @@ func Test_visual_hl_with_showbreak()
   call StopVimInTerminal(buf)
 endfunc
 
+func Test_Visual_r_CTRL_C()
+  new
+  " visual r_cmd
+  call setline(1, ['   '])
+  call feedkeys("\<c-v>$r\<c-c>", 'tx')
+  call assert_equal([''], getline(1, 1))
+
+  " visual gr_cmd
+  call setline(1, ['   '])
+  call feedkeys("\<c-v>$gr\<c-c>", 'tx')
+  call assert_equal([''], getline(1, 1))
+  bw!
+endfu
+
 " vim: shiftwidth=2 sts=2 expandtab


justinmk:
> r_CTRL-C replaces in the selected area all characters by a literal CTRL-C.

How is this useful? `<c-v><c-c>` can be used for that (or if not, that should be fixed). 

`<c-c>` should always cancel the mode (except when `<c-v>` is pending ? :)

chrisbra:
it's documenting existing behaviour 🤷

k-takata:
> it's documenting existing behaviour 🤷

Where can I find it? I think it's inconsistent.

* In normal mode, `r<c-c>` cancels the replacement.
* In normal mode, `r<c-v><c-c>` replaces the character to `^C`.
* In visual mode, `<c-c>` ends visual mode.
* In visual mode, `r<c-c>` is not well defined? Shouldn't it cancel the replacement?
* In visual mode, `r<c-v><c-c>` replaces the characters to `^C`.

chrisbra:
> In visual mode, r<c-c> is not well defined? Shouldn't it cancel the replacement?

Yes, perhaps we should change this behaviour.

But until that patch, `r<c-c>` in Windows GVim replaces in the selection with `^C`. Compile it with `VIMDLL` defined or test Windows UI `vim.exe` it ends visual selection. 

I believe that's because in Windows GUI, `<C-C>` does not set `got_int`, but only `Ctrl-Break` does it. 

But don't know

PR #13113, opened 2023-09-17
Fix class constructor regression

### LINKED ISSUE #13102: [vim9script] Regression after fixing constructor argument type checking bug
### Steps to reproduce

Patch 9.0.1724 (“vim9class constructor argument type checking bug”) has introduced a regression in one of my Vim 9 libraries. I have [a test suite that caught that](https://github.com/lifepillar/vim-devel/blob/main/start/librelalg/test/test_librelalg.vim), but I was able to nail down the issue to the following self-contained script:

```vim
vim9script

def ListifyKeys(keys: any): list<list<string>>
  if type(keys) == v:t_string
    return [[keys]]
  endif
  return keys
enddef

class Rel
  this.keys: list<list<string>>

  def new(keys: any)
    const keys_: list<list<string>> = ListifyKeys(keys)
  enddef
endclass

def Test()
  var R = Rel.new('A')
enddef

Test()
```

This fails with `E1013: Argument 1: type mismatch, expected list<list<string>> but got string`.

Interestingly, if the `this.keys` attribute is removed, no error occurs.

**Edit:** no error occurs if `R` is created outside the function, either.

### Expected behaviour

I would expect the script above to run without errors. That was the case before patch 9.0.1724.

### Version of Vim

9.0.1899

### Environment

macOS 13.5.1
Apple Terminal
xterm-256color
zsh 5.9

### Logs and stack traces

_No response_

h-east:
Close #13102

### DIFF
--- a/src/userfunc.c
+++ b/src/userfunc.c
@@ -65,6 +65,7 @@ one_function_arg(
 	garray_T    *newargs,
 	garray_T    *argtypes,
 	int	    types_optional,
+	garray_T    *is_argobjm,
 	evalarg_T   *evalarg,
 	exarg_T	    *eap,
 	int	    is_vararg,
@@ -136,7 +137,8 @@ one_function_arg(
     }
 
     // get any type from "arg: type"
-    if (argtypes != NULL && (skip || ga_grow(argtypes, 1) == OK))
+    if (argtypes != NULL && (skip || ga_grow(argtypes, 1) == OK)
+	    && is_argobjm != NULL && (skip || ga_grow(is_argobjm, 1) == OK))
     {
 	char_u *type = NULL;
 
@@ -172,6 +174,7 @@ one_function_arg(
 		type = vim_strsave((char_u *)
 					    (is_vararg ? "list<any>" : "any"));
 	    ((char_u **)argtypes->ga_data)[argtypes->ga_len++] = type;
+	    ((int8_T *)is_argobjm->ga_data)[is_argobjm->ga_len++] = 0;
 	}
     }
 
@@ -221,6 +224,7 @@ get_function_args(
     garray_T	*newargs,
     garray_T	*argtypes,	// NULL unless using :def
     int		types_optional,	// types optional if "argtypes" is not NULL
+    garray_T	*is_argobjm,	// NULL unless using :def
     evalarg_T	*evalarg,	// context or NULL
     int		*varargs,
     garray_T	*default_args,
@@ -241,6 +245,8 @@ get_function_args(
 	ga_init2(newargs, sizeof(char_u *), 3);
     if (argtypes != NULL)
 	ga_init2(argtypes, sizeof(char_u *), 3);
+    if (is_argobjm != NULL)
+	ga_init2(is_argobjm, sizeof(int8_T), 3);
     if (!skip && default_args != NULL)
 	ga_init2(default_args, sizeof(char_u *), 3);
 
@@ -295,7 +301,7 @@ get_function_args(
 
 		arg = p;
 		p = one_function_arg(p, newargs, argtypes, types_optional,
-						     evalarg, eap, TRUE, skip);
+					is_argobjm, evalarg, eap, TRUE, skip);
 		if (p == arg)
 		    break;
 		if (*skipwhite(p) == '=')
@@ -350,11 +356,13 @@ get_function_args(
 					       vim_strnsave(arg, argend - arg);
 		newargs->ga_len++;
 
-		if (argtypes != NULL && ga_grow(argtypes, 1) == OK)
+		if (argtypes != NULL && ga_grow(argtypes, 1) == OK
+			&& is_argobjm != NULL && ga_grow(is_argobjm, 1) == OK)
 		{
 		    // TODO: use the actual type
 		    ((char_u **)argtypes->ga_data)[argtypes->ga_len++] =
 						  vim_strsave((char_u *)"any");
+		    ((int8_T *)is_argobjm->ga_data)[is_argobjm->ga_len++] = 1;
 
 		    // Add a line to the function body for the assignment.
 		    if (ga_grow(newlines, 1) == OK)
@@ -391,7 +399,7 @@ get_function_args(
 
 	    arg = p;
 	    p = one_function_arg(p, newargs, argtypes, types_optional,
-						    evalarg, eap, FALSE, skip);
+					is_argobjm, evalarg, eap, FALSE, skip);
 	    if (p == arg)
 		break;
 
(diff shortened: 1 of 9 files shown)


h-east:
change point
- Stopped referencing the function body in compile time
  (When check the function body, there is no end to it.)
- Added TTFLAG_ARG_OBJM for tt_flags. Set when the constructor's formal argument is an object member.
  (I thought about including "this." in ufunc->uf_args, but I decided to go with this method because it seemed like it would have a wide range of effects.)
  (However, I think it would be good to challenge someone.)

h-east:
Hmm, This patch seems to have a problem.
`fp->uf_arg_types[i]` is a pointer.  copy an address of `type`.
Therefore, changing the value of a member will affect others...

```
parse_argument_types(ufunc_T *fp, garray_T *argtypes, int varargs, garray_T *argflags)
{
    ....
    fp->uf_arg_types = ALLOC_CLEAR_MULT(type_T *, len);
    ....
    type = parse_type(&p, &fp->uf_type_list, TRUE);
    ....
    fp->uf_arg_types[i] = type;
    type = parse_type(&p, &fp->uf_type_list, TRUE);
    ....
    fp->uf_arg_types[i]->tt_flags &= ~TTFLAG_ARG_OBJM;              // BAD
    ....
}
```

h-east:
Fix it!

change point
- I stopped checking in generate_CALL() and set the type of object member in fp->uf_arg_types[i] in parse_argument_types().

yegappan:
> Fix it!
> 
> change point
> 
> * I stopped checking in `generate_CALL()` and set the type of object member to `fp->uf_arg_types[i]` 
> in `parse_argument_types()`.
>

It looks like the latest change impacts the compilation of arguments to every object method,
The type checking for object methods is working correctly.  I thought only the type checking
for arguments to the new() constructor function with the same name as a member variable
is not working.

h-east:
@yegappan, @chrisbra 
Yes, I am going to start talking about the language specification of Vim9 script.
I think it is not good that object members can be used as formal arguments for methods other than class constructors.
Reading vim9class.txt, my guess is that Bram wanted to apply this mechanism only to class constructors.

Proposal:
Let's make it an error when an object member is specified as formal arguments to a class method that is not a constructor?

h-east:
Currently, the following code is working. Is this good?

```vim
vim9script

class C
  this.str: string
  this.num: number
  def new(this.str, this.num)
  enddef
  def FuncA(this.str, this.num)
  enddef
endclass

def Check()
  try
    var c = C.new('a', 1)
	c.FuncA('xyz', 789)
	echo c.str
	echo c.num
  catch
    echo $'Exception was caught: {v:exception}'
  endtry
enddef

Check()
```

yegappan:
> @yegappan, @chrisbra Yes, I am going to start talking about the language specification of Vim9 script. I think it is not good that object members can be used as formal arguments for methods other than class constructors. Reading vim9class.txt, my guess is that Bram wanted to apply this mechanism only to class constructors.
> 
> Proposal: Let's make it an error when an object member is specified as formal arguments to a class method that is not a constructor?

Yes.  Sounds good.  The compilation of an class or object method should fail under the following conditions:

1. The "this." prefix is used for an argument name,
2. The argument name overlaps with a class variable name.

Note that it is acceptable to use an argument name to an object method with the same
name as the object variable name without the "this." prefix.

h-east:
First of all, could you please accept this patch?
After that, would you like to discuss the following suggestions on issues/#13152 ?

> Proposal:
Let's make it an error when an object member is specified as formal arguments to a class method that is not a constructor?

yegappan:
> > But if the arguments to the FuncA()
> > method are changed to "(str, num)", without the "this." prefix, then the
> > compilation should succeed.
> 
> No. The following code will result in an error regardless of whether with my patch or not. Type specification is required if the formal argument is not an object member. This is correct. This behavior will remain the same in the future.
> 

Yes.  What I meant is that it is allowed to use a parameter name (with the type specification)
with the same name as an object variable name.

src/userfunc.c
```diff
@@ -172,6 +174,7 @@ one_function_arg(
 		type = vim_strsave((char_u *)
 					    (is_vararg ? "list<any>" : "any"));
 	    ((char_u **)argtypes->ga_data)[argtypes->ga_len++] = type;
+	    ((int8_T *)is_argobjm->ga_data)[is_argobjm->ga_len++] = 0;
```
yegappan: Can you rename "is_argobjm" to "is_arg_objm" or "arg_objm" (it is more clear)?

yegappan: As "is_argobjm" is a list of Booleans, can you use "FALSE" instead of "0"?

src/userfunc.c
```diff
@@ -350,11 +356,13 @@ get_function_args(
...
 		newargs->ga_len++;
 
-		if (argtypes != NULL && ga_grow(argtypes, 1) == OK)
+		if (argtypes != NULL && ga_grow(argtypes, 1) == OK
+			&& is_argobjm != NULL && ga_grow(is_argobjm, 1) == OK)
 		{
 		    // TODO: use the actual type
 		    ((char_u **)argtypes->ga_data)[argtypes->ga_len++] =
 						  vim_strsave((char_u *)"any");
+		    ((int8_T *)is_argobjm->ga_data)[is_argobjm->ga_len++] = 1;
```
yegappan: As "is_argobjm" is a list of Booleans, can you use "TRUE" instead of "1"?

yegappan (changes requested):

h-east:
> Yes. What I meant is that it is allowed to use a parameter name (with the type specification)
with the same name as an object variable name.

Yes, I understand that.

This PR is a fix for the code below which unexpectedly caused an error.
(In addition, type checking is also performed when an object method (this.~) is specified as formal arguments in a method other than the constructor.)   <--- I would like to discuss this part later.

```vim
vim9script

class C
  this.str: string
  this.num: number
  def newFuncA(str: any, num: any)
  enddef
endclass

def Check()
  try
    var c = C.newFuncA(1, 2)
  catch
    echo $'Exception was caught: {v:exception}'
  endtry
enddef

Check()
```

src/userfunc.c
h-east: I'll rename to `arg_objm`.

yegappan (approved):

chrisbra:
thanks!

vim-ml:
Hi,

On Sun, Sep 24, 2023 at 6:44 AM Christian Brabandt <
***@***.***> wrote:

> thanks!
>
>
>
The CI ASAN test is failing after this patch.  There is a memory leak with
this PR.

Regards,
Yegappan

h-east:
Hmm, memory leak not fixed.

PR #13116, opened 2023-09-18
Calling a class method from another method without the class name prefix doesn't work properly

yegappan:
Update the help text with additional description for static members and interface members.
Update error numbers in the help.  Add additional comments to tests.


runtime/doc/vim9class.txt
```diff
@@ -312,14 +321,38 @@ as the first character in the method name: >
...
+	    _Foo()
 	enddef
     endclass
 <
 							*E1370*
 Note that constructors cannot be declared as "static", because they always
 are.
 
+To access the class methods and class variables of a super class in an
+extended class, the class name prefix should be used: >
```
errael: Is it worth mentioning that this is how they are access from anywhere other than the defining class. Maybe something like
```
... class name prefix should be used, _just as from anywhere outside of the defining class_: >

File: src/testdir/test_vim9_class.vim [hunk 2]
```diff
@@ -588,7 +588,37 @@ def Test_member_any_used_as_object()
...
+      this.value: number = 0
+    endclass
+
+    class Outer
+      this.inner: Inner
+    endclass
+
+    def F(outer: Outer)
+      outer.inner.value = 1
+    enddef
```
errael, line 606: Shouldn't this fail because writing  to a read-only variable outside of the class that defined it?

runtime/doc/vim9class.txt
yegappan: Yes. I will update the text.  Let me know if the vim9class.txt file needs to be updated with any additional information to reflect the recent changes in the specification.

src/testdir/test_vim9_class.vim
yegappan: Yes.  This should fail.  This is a bug.

yegappan: I have updated the PR with a fix for this issue.  Let me know if you find any other issue.

errael:
```
*vim9class.txt*	For Vim version 9.0.  Last change: 2023 Mar 22
```
The date needs updating

errael:
The message from script level isn't accurate (it's good inside a def)
```
line   28:
E1335: Member is not writable: inner2
```

```
vim9script

class Inner
    this.value: number = 0
endclass

class Outer
    this.inner: Inner
endclass

class Outer2
    this.inner2: Outer
endclass

def F(outer: Outer)
    outer.inner.value = 1
enddef

def G(outer: Outer2)
    outer.inner2.inner.value = 1
enddef

var innerScript = Inner.new(3)
var outerScript = Outer.new(innerScript)
var outer2Script = Outer2.new(outerScript)
#outerScript.inner.value = 1     # <<<<< error message confusing
echo outer2Script.inner2.inner.value
outer2Script.inner2.inner.value = 1     # <<<<< error message confusing

#F(outerScript)
G(outer2Script)
```

errael:
(not related to this PR)
Not an error to call methods on a null object; seems like it should be.
If the method called on a null object references a variable, then there is a runtime error.
Behavior not the same between `:def` and script level
```
vim9script

class C
    def F()
        echo "C.F()"
    enddef

    static def S(o_any: any)
        var o_typed: C = o_any
        o_typed.F()
    enddef
endclass

#var o = null_object        # this declaration get compilation error
var o: C = null_object      # but with a type, slips by

def T()
    C.S(o)              # <<<<< NO ERROR
    o.F()               # <<<<< NO ERROR
enddef
T()

C.S(o)              # <<<<< NO ERROR
o.F()               # <<<<< run time error
```

chrisbra:
thanks

errael:
Report about incorrect error message moved to #13118.
Report about method call on null object sometimes succeeding moved to #13119.

Deleted from this PR's comments.

PR #12745, opened 2023-08-07
[in remember of Bram]: CI: clang-17

pheiduck:

PR #13117, opened 2023-09-18
Fix UBSAN error in regexp and simplify upper/lowercase modifier code

ychin:
The implementation of \u / \U / \l / \L modifiers (see `:h s/\u`) in the substitute command relies on remembering the state by setting function pointers on func_all/func_one in the code. The code signature of `fptr_T` is supposed to return void* (due to C function signatures not being able to return itself due to type recursion), and the definition of the functions (e.g. to_Upper) didn't follow this rule, and so the code tries to cast functions of different signatures, resulting in undefined behavior error under UBSAN in Clang 17. See #12745.

We could just fix `do_Upper`/etc to just return void*, which would fix the problem. However, these functions actually do not need to return anything at all. It used to be the case that there was only one pointer "func" to store the pointer, which is why the function needs to either return itself or NULL to indicate whether it's a one time or ongoing modification. However, c2c355df6f094cdb9e599fd395a78c14486ec697 (7.3.873) already made that obsolete by introducing `func_one` and `func_all` to store one-time and ongoing operations separately, so these functions don't actually need to return anything anymore because it's implicit whether it's a one-time or ongoing operation. Simplify the code to reflect that.

### DIFF
--- a/src/regexp.c
+++ b/src/regexp.c
@@ -1709,46 +1709,20 @@ cstrchr(char_u *s, int c)
 //		      regsub stuff			      //
 ////////////////////////////////////////////////////////////////
 
-/*
- * We should define ftpr as a pointer to a function returning a pointer to
- * a function returning a pointer to a function ...
- * This is impossible, so we declare a pointer to a function returning a
- * void pointer. This should work for all compilers.
- */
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

PR #13121, opened 2023-09-18
Fix SEGV when `instanceof()` null object.

errael:


errael:
@yegappan @chrisbra Could this trivial fix for vim crash get reviewed

chrisbra:
looks good thanks. I'll merge later today

yegappan (approved):

PR #13122, opened 2023-09-18
Fix if_python undefined behavior for function pointer casts

ychin:
Identified by clang 17 UBSAN (see #12745). Make sure to cast function pointers with the same signature only.


chrisbra:
I thought we could also get rid of the range_initializer type def. But apparently not.

Issue #13123, opened 2023-09-19
[vim9script] regression `ctx.state["meta"] = 1` fails


errael:
### Steps to reproduce

The following test fails with
```
E18: Unexpected characters in :let
```
with `patch 9.0.1890: Vim9: lookup code for class/object repaeated` or later.
NOTE that it works in a `:def`
```
vim9script

class Context
  public this.state: dict<any> = {}
endclass

var ctx = Context.new()

# Following gets key not found as expectd
# var x = ctx.state["meta"]

def F()
    ctx.state["meta"] = 1
enddef
F()

ctx.state["meta"] = 1
```

### Expected behaviour

no error

### Version of Vim

9.0.1890 or later

### Environment

ubuntu

errael:
Discovered this regression looking at #12024 for another issue.

PR #13125, opened 2023-09-19
When using nested objects, check for the member variable access correctly

### LINKED ISSUE #13118: [vim9script] inaccurate error message
### Steps to reproduce

The following produces the error
```
line   28:
E1335: Member is not writable: inner2
```
In a `:def`, the correct error is produced
```
E46: Cannot change read-only variable "value"
```

```
vim9script

class Inner
    this.value: number = 0
endclass

class Outer
    this.inner: Inner
endclass

class Outer2
    this.inner2: Outer
endclass

def F(outer: Outer)
    outer.inner.value = 1
enddef

def G(outer: Outer2)
    outer.inner2.inner.value = 1
enddef

var innerScript = Inner.new(3)
var outerScript = Outer.new(innerScript)
var outer2Script = Outer2.new(outerScript)
#outerScript.inner.value = 1     # <<<<< error message confusing
echo outer2Script.inner2.inner.value
outer2Script.inner2.inner.value = 1     # <<<<< error message confusing

#F(outerScript)
G(outer2Script)```

### Expected behaviour

Same/similar error as given in a `:def`

### Version of Vim

9.0.1909

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #13118 and #13123

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -1582,7 +1582,6 @@ get_lval(
 					lp->ll_tv->vval.v_object + 1)) + m_idx;
 			else
 			    lp->ll_tv = &cl->class_members_tv[m_idx];
-			break;
 		    }
 		}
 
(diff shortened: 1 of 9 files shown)


yegappan:
The changes in this PR check for the member access only under some conditions.  For example, if a object
member of type Dict or List is assigned to a local variable and then the Dict is modified, then the access is
not checked.  I will reopen this PR after looking into this issue.

PR #13128, opened 2023-09-19
Fix more undefined behaviors in if_python

ychin:
Fix remaining UBSAN errors from Clang 17 in if_python in casting function pointers.

Also fix a mistake where `PyMem_Free()` should be returning void, by the dynamic build is mistakenly casting it as a function that returns an int.

### DIFF
--- a/src/if_py_both.h
+++ b/src/if_py_both.h
@@ -2079,9 +2079,10 @@ typedef struct
 } dictiterinfo_T;
 
     static PyObject *
-DictionaryIterNext(dictiterinfo_T **dii)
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks. I did started to work on it yesterday, but some of the function signatures were way over my head :/

ychin:
Ah. Sorry for the scheduling conflict.

I think the key is just to not cast any function pointers (since any casting is undefined behavior) and just use the same functions as defined and then fix it to work when it fails to compile.

chrisbra:
No no, I appreciate taking an active role here :) Thanks!

PR #13131, opened 2023-09-20
Write variable in nested read-only object reference.

### LINKED ISSUE #13130: Can't write to public object variables
### Steps to reproduce

In https://github.com/vim/vim/issues/13118#issuecomment-1726020513 it says 
```
In the current implementation the read-only object variable behaves like a "final" variable.
```
The following script gives the error
```
E1335: Member is not writable: inner
```
While it's true that `inner` is not writable, the test does not attempt to write to the variable `inner`. It is writing to the `class Inner` public variable `val`.
```
vim9script

class Inner
  public this.val = 13
endclass

class Outer
    this.inner = Inner.new()
endclass

var o1 = Outer.new()

o1.inner.val = 11    # <<<<< can't write
```

### Expected behaviour

No error. The object's variable `val` is written to.

### Version of Vim

9.0.1916

### Environment

ubuntu

### Logs and stack traces

_No response_

errael:
Fix #13130

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -1564,6 +1564,9 @@ get_lval(
 					om->ocm_name);
 				return NULL;
 			    case VIM_ACCESS_READ:
+				// If [idx] or .key following, read only OK.
+				if (*p == '[' || *p == '.')
+				    break;
 				if ((flags & GLV_READ_ONLY) == 0)
 				{
 				    semsg(_(e_member_is_not_writable_str),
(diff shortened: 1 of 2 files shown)


yegappan (approved):

errael:
@yegappan I added a couple tests. Changed the target variable to read only to check the write fail. No code change.

yegappan:
> @yegappan I added a couple tests. Changed the target variable to read only to check the write fail. No code change.

Thanks for adding the tests.  Looks good to me.

chrisbra:
thanks!

PR #13133, opened 2023-09-20
LSP server request message is misinterpreted as a response message

yegappan:
When waiting for a response message from the LSP server, if a new request message is received from the server with the same sequence number, then the request message is incorrectly interpreted as a response message.

This is reported in https://github.com/yegappan/lsp/issues/395

The LSP request and notifications messages will have the "method" field in the header.  The LSP response messages will not
have this field.  When looking for a response message with a matching sequence number, use this field to ignore request
and notification messages.

### DIFF
--- a/src/channel.c
+++ b/src/channel.c
@@ -2927,9 +2927,16 @@ may_invoke_callback(channel_T *channel, ch_part_T part)
 	    seq_nr = 0;
 	    if (d != NULL)
 	    {
-		di = dict_find(d, (char_u *)"id", -1);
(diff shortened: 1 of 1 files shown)


chrisbra:
so what is happening to the messages with the `method` field then? It looks like we discard them?

yegappan:
> so what is happening to the messages with the `method` field then? It looks like we discard them?

When adding tests for this change, I realized the change is not correct.  I have opened PR https://github.com/vim/vim/pull/13141 to fix this properly.

PR #7738, opened 2021-01-23
Add an option for using jump list like tag stack

yegappan:
Port the "jumpoptions" support from NeoVim:

https://neovim.io/doc/user/motion.html#jumplist-stack
https://github.com/neovim/neovim/commit/39094b3faedde9601160806901941e4808925410
https://github.com/neovim/neovim/pull/11530
https://vi.stackexchange.com/questions/18344/how-to-change-jumplist-behavior

PR #13134, opened 2023-09-20
Add an option for using the jump list like tag stack

yegappan:
I am not able to reopen the previous PR (https://github.com/vim/vim/pull/7738).  So opening a new one.

Port the "jumpoptions" support from NeoVim:

https://neovim.io/doc/user/motion.html#jumplist-stack
https://github.com/neovim/neovim/commit/39094b3faedde9601160806901941e4808925410
https://github.com/neovim/neovim/pull/11530
https://vi.stackexchange.com/questions/18344/how-to-change-jumplist-behavior

Based on the feedback in the previous PR, it looks like many people like this option.

Bram wanted a much more elaborate infrastructure (similar to the undo tree) that is based on
timestamps to go back in time to different jump locations.  This will take quite a bit of effort
to implement.  So instead we can with this option for now.

### DIFF
--- a/runtime/doc/motion.txt
+++ b/runtime/doc/motion.txt
@@ -1135,7 +1135,46 @@ When you split a window, the jumplist will be copied to the new window.
 If you have included the ' item in the 'viminfo' option the jumplist will be
 stored in the viminfo file and restored when starting Vim.
 
+							*jumplist-stack*
+When 'jumpoptions' option includes "stack", the jumplist behaves like the tag
+stack.  When jumping to a new location from the middle of the jumplist, the
+locations after the current position will be discarded. With this option set
+you can move through a tree of jump locations. When going back up a branch and
+then down another branch, CTRL-O still takes you further up the tree.
 
+Given a jumplist like the following in which CTRL-O has been used to move back
+three times to location X: >
+
+     jump line  col file/text
+       2  1260    8 mark.c		<-- location X-2
+       1   685    0 eval.c		<-- location X-1
+    >  0   462   36 eval.c		<-- location X
+       1   479   39 eval.c
+       2   213    2 mark.c
+       3   181    0 mark.c
+<
+jumping to (new) location Y results in the locations after the current
+locations being removed: >
+
+     jump line  col file/text
+       3  1260    8 mark.c		<-- location X-2
+       2   685    0 eval.c		<-- location X-1
+       1   462   36 eval.c		<-- location X
+    >
+<
+
+Then, when yet another location Z is jumped to, the new location Y appears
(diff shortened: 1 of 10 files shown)


chrisbra:
thanks. sounds like there is some demand for this, so let me include this.

Shane-XB-Qian:
> sounds like there is some demand for this,

not sure if i had readed or tried correctly, this seems made ctrl-i useless, was that really useful demand or it is a bug of this.

Shane-XB-Qian:
> sounds like there is some demand for this,

not sure if i read or tried correctly, this seems made ctrl-i useless, was that really a useful demand or it is a bug of this.

yegappan:
> > sounds like there is some demand for this,
> 
> not sure if i read or tried correctly, this seems made ctrl-i useless, was that really a useful demand or it is a bug of this.

The CTRL-O/CTRL-I commands to browse the jump list still work.  Note that without this option, if you go back
in the jump list using CTRL-O and then jump to a new location, the new jump location will be added to the end
of the jump list.  Now if you press CTRL-O, it will go to a different location (confusing the users).  With this option,
the jump list will behave like the tag stack.  Please refer to the previous PR for the comments and also you can
try this option out in the latest Vim.

yegappan:
There is a thread in Vi stack exchange about this: https://vi.stackexchange.com/questions/18344/how-to-change-jumplist-behavior

Shane-XB-Qian:
as my understanding, the traditional one maybe a bit messed order after ctrl-o and new jump,
but this added 'stack' one may clear the record, i felt did not it make case be worse?
a.k.a bram said based on timestamp maybe was the right one IF really want to refine.

chrisbra:
Think of it as using the browser based forward/backward navigations. 

This may actually make me use the jumplist, because I always found it to be too confusing where to jump next.

Yes a timestamp based navigation (similar to the undo-stack) may be even more desirable, but we then have to think about additional commands to navigate into the different branches (and it may still be confusing then)

Shane-XB-Qian:
> because I always found it to be too confusing where to jump next

yes, that traditional one sometime made me feel confused too, 
but i may like keeping the jump list history there too, this added 'stack' one made me feel it worse.

chrisbra:
well, if you don't have to use this option.

Shane-XB-Qian:
yes, i did not and would not set it.

but, hi, my point was let's keeping the style which bram or classic vim's style which preferred to add confirmed refined one, vs not a mid-stage of something, a.k.a not just someone said it is `neovim or so-called IDE` style then hurry to add it.
// and i doubt was this really a IDE style?

PR #13136, opened 2023-09-20
Adjust message `e_member_is_not_writable_str` to include more info.

errael:
Change `member` to `variable`; include class in msg; quote program names. Message based on comments from #13068 and #13135.

Update `E1335` to
```
E1335: Variable "value" in "class C" is not writable
```
from
```
E1335: Member is not writable: value
```
Change one use of `E46` to `E1335`
```
E46: Cannot change read-only variable "value"
```

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3411,7 +3411,7 @@ EXTERN char e_cannot_access_private_member_str[]
 	INIT(= N_("E1333: Cannot access private member: %s"));
 // E1334 unused
 EXTERN char e_member_is_not_writable_str[]
-	INIT(= N_("E1335: Member is not writable: %s"));
+	INIT(= N_("E1335: Variable \"%s\" in class \"%s\" is not writable"));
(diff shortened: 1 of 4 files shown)


errael:
@yegappan Draft, after decision tests need update. Along with your thoughts, there's this PR and #13135 to consider.

errael:
Don't include the word class in quotes. The message now looks like
```
E1335: Variable "value" in class "C" is not writable
```
as suggested in https://github.com/vim/vim/discussions/13135#discussioncomment-7062755

yegappan:
The change looks good to me.

yegappan (approved):

errael:
> The change looks good to me.

OK. I'll fix the tests to account for the new message, then take it out of draft.

Feedback on initial message of #13135 and examples in https://github.com/vim/vim/discussions/13135#discussioncomment-7065452

PR #13140, opened 2023-09-20
Fix more if_python undefined behaviors in casting function pointers

ychin:
Fix remaining issues in the if_python code in casting incompatible function pointers leading to Clang 17 giving runtime errors during UBSAN.


chrisbra:
thanks. I actually did run the various `python` tests yesterday with a clang-17 compiled with sanitizer flags and it did not complain about any thing :/

pheiduck:
I hope this is the last patch we need for clang-17 to run on CI Environment.

PR #13141, opened 2023-09-21
LSP server request message is misinterpreted as a response message (not properly fixed by 9.0.1922)

yegappan:


src/optionstr.c
```diff
@@ -1858,7 +1858,7 @@ did_set_isopt(optset_T *args)
  * The 'jumpoptions' option is changed.
  */
     char *
-did_set_jumpoptions(optset_T *args)
+did_set_jumpoptions(optset_T *args UNUSED)
```
Shane-XB-Qian, line 1861: what's this? related?

src/channel.c
```diff
@@ -2466,6 +2466,12 @@ channel_get_json(
 		d = item->jq_value->vval.v_dict;
 		if (d == NULL)
 		    goto nextitem;
+		// When looking for a response message from the LSP server,
+		// ignore new LSP request and notification messages.  LSP
+		// request and notification messages have the "method" field in
+		// the header and the response messages do not have this field.
+		if (dict_has_key(d, "method"))
+		    goto nextitem;
```
Shane-XB-Qian, line 2474: is that really ok? 
in that reported case, seems if ignore or not handle that workspace/configure msg from lsp server, that server looks would be hanging there.

src/optionstr.c
yegappan: I don't want to open another PR just to fix this compiler warning.

src/channel.c
yegappan: Yes.  The request messages are not dropped, they will be processed after processing the response message.

Shane-XB-Qian: i felt it's not same behavior or worked one for that reported case.

Shane-XB-Qian:
v9.0.1922 was helped that reported case, however looks this not.
@lithammer maybe you can try again at your side too.

lithammer:
Yeah I still have the issue after this patch (tested on 1927).

```
Error: request textDocument/inlayHint failed (no result):
{'method': 'workspace/configuration', 'jsonrpc': '2.0', 'id': 2, 'params': {'items': [{'section': 'gopls'}]}}
```

PR #13144, opened 2023-09-21
Better globbing support when using bash

### LINKED ISSUE #13002: Starstar and {} expansion do not work together
### Steps to reproduce

Starstar (**) expansion does not work when combined with {} enumeration.
Suppose we want to run vimgrep on *.h and *.cpp files in the directory my_project and its subdirectories:
```
:vimgrep iostream my_project/**/*.{h,cpp}
```
This does not work as expected, no files are found.
Here is the files tree:
```
$ find
.
./my_project
./my_project/my_lib
./my_project/my_lib/backend
./my_project/my_lib/backend/a.h
./my_project/my_lib/backend/a.cpp
```

### Expected behaviour

I expect that the search in my_project/**/*.{h,cpp} will work as in shell e.g.
```
$ ls my_project/**/*.{h,cpp}
my_project/my_lib/backend/a.cpp  my_project/my_lib/backend/a.h
```

### Version of Vim

8.1 (Included patches: 1-2269, 3612, 3625, 3669, 3741)

### Environment

OS: Ubuntu 20.04
Terminal: Gnome Terminal 3.36.2
Shell: GNU bash, version 5.0.17(1)-release

### Logs and stack traces

_No response_

chrisbra:
Starting with bash 4.0 supports extended globbing using the globstar shell option. This makes matching recursively below a certain directory using the ** pattern work as expected nowadays.  However, we need to explicitly enable this using the 'shopt -s globstar' bash command.

So let's assume nowadays all bashs support this option (it was releases 2009, so 14 years ago) and make use of this option, when the shell option ends with 'bash'.

fixes: #13002

### DIFF
--- a/runtime/doc/editing.txt
+++ b/runtime/doc/editing.txt
@@ -1,4 +1,4 @@
-*editing.txt*   For Vim version 9.0.  Last change: 2023 Apr 23
+*editing.txt*   For Vim version 9.0.  Last change: 2023 Sep 22
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -385,7 +385,9 @@ as a wildcard when "[" is in the 'isfname' option.  A simple way to avoid this
 is to use "path\[[]abc]", this matches the file "path\[abc]".
 
 					*starstar-wildcard*
-Expanding "**" is possible on Unix, Win32, macOS and a few other systems.
+Expanding "**" is possible on Unix, Win32, macOS and a few other systems (but
+it may depend on your 'shell' setting. It's known to work correctly for zsh; for
+bash this requires at least bash version >= 4.X).
 This allows searching a directory tree.  This goes up to 100 directories deep.
 Note there are some commands where this works slightly differently, see
 |file-searching|.
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -6701,14 +6701,17 @@ mch_expand_wildcards(
 #define STYLE_GLOB	1	// use "glob", for csh
 #define STYLE_VIMGLOB	2	// use "vimglob", for Posix sh
 #define STYLE_PRINT	3	// use "print -N", for zsh
-#define STYLE_BT	4	// `cmd` expansion, execute the pattern
-				// directly
+#define STYLE_BT	4	// `cmd` expansion, execute the pattern directly
+#define STYLE_GLOBSTAR	5	// use extended shell glob for bash (this uses extended
+				// globbing functionality using globstar)
     int		shell_style = STYLE_ECHO;
     int		check_spaces;
     static int	did_find_nul = FALSE;
     int		ampersand = FALSE;
 		// vimglob() function to define for Posix shell
     static char *sh_vimglob_func = "vimglob() { while [ $# -ge 1 ]; do echo \"$1\"; shift; done }; vimglob >";
+		// globstar() function with globstar setting enabled
+    static char *sh_globstar_opt = "shopt -s globstar; ";
 
     *num_file = 0;	// default: no files found
     *file = NULL;
@@ -6755,6 +6758,8 @@ mch_expand_wildcards(
      *	    If we use *zsh, "print -N" will work better than "glob".
      * STYLE_VIMGLOB:	NL separated
      *	    If we use *sh*, we define "vimglob()".
+     * STYLE_GLOBSTAR:	NL separated
+     *	    If we use *bash*, we define "vimglobstar()".
      * STYLE_ECHO:	space separated.
      *	    A shell we don't know, stay safe and use "echo".
      */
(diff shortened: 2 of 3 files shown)


chrisbra:
Am I old, if I remember the bash 2.1 times? 

In any case, I am wondering, if we shouldn't rather document the quirks or bash instead?

something like this:
```patch
diff --git a/runtime/doc/editing.txt b/runtime/doc/editing.txt
index da9aeb7fe..d696e52f0 100644
--- a/runtime/doc/editing.txt
+++ b/runtime/doc/editing.txt
@@ -385,7 +385,9 @@ as a wildcard when "[" is in the 'isfname' option.  A simple way to avoid this
 is to use "path\[[]abc]", this matches the file "path\[abc]".

                                        *starstar-wildcard*
-Expanding "**" is possible on Unix, Win32, macOS and a few other systems.
+Expanding "**" is possible on Unix, Win32, macOS and a few other systems (but it
+may depend on your shell setting. It's known to work correctly for zsh, for
+bash you need to configure it specifically |bash-starstar-wildcard|)
 This allows searching a directory tree.  This goes up to 100 directories deep.
 Note there are some commands where this works slightly differently, see
 |file-searching|.
@@ -407,6 +409,18 @@ Note that the path with "/sys" is included because it does not need to match
 "/inc".  Thus it's like matching "/usr/inc*/*/*...", not
 "/usr/inc*/inc*/inc*".

+                                       *bash-starstar-wildcard*
+Bash supports expansion of the "**" pattern only, when the globstar setting
+is enabled in your bash configuration file (typically .bashrc).  But this
+requires at least bash version 4. To make filename expansion in Vim using the
+"**" pattern work when you are using the bash shell, add the following setting
+to your .bashrc file: >
+       shopt -s globstar
+Now that bash is prepared to expand the ** pattern, we need to tell Vim to use
+and interactive shell, whenever it needs to expand the ** pattern, so that it
+will source your bashrc file. To do set, set the following: >
+       set shellcmdflag=-ic
+<
                                        *backtick-expansion* *`-expansion*
 On Unix and a few other systems you can also use backticks for the file name
 argument, for example: >
```

benknoble:
On both macOS 13.5.1 and 11.7, `/bin/bash --version` is `3.2.57(1)-release (x86_64-apple-darwin20)` [or `darwin22` on the newer OS].

---

I think documenting the quirk suffices to cover older shells; you can also set `BASH_ENV` to a file to execute in non-interactive mode to avoid messing with `shellflags` and if you want a "minimal" bash-for-Vim config.

Shane-XB-Qian:
it is not a good idea to set shellcmdflag to -ic, to a complex vimrc it maybe leading to unexpected result.
bash_env looks not a good idea too, since it's globe setting which may impact other script running.
how about just your PR gave a err if 'shopt -s foo' returned a failure, otherwise then good to go.
// there are some others differences between shells and versions, perhaps just pick up some to commonly support.
// which is e.g by default bash redir is '>' but after vim startup it is default to '2>&1', it is history maybe not ok to change.
// or compatible.

benknoble:
For `BASH_ENV`, you could set it to a file from within Vim that is Vim-specific. The idea about testing `shopt` is clever, but might hurt startup times.

Shane-XB-Qian:
yes, i am saying 'minimal', which if to and just to make starstar work.

chrisbra:
okay, how about this then?

```
[ ${BASH_VERSINFO[0]} -ge 4 ] && shopt -s globstar;
```

benknoble:
I'm not sure what I meant about startup times, since I assume this code only runs when starstar is found in things like vimgrep. I'll try to find some time to test this out.

benknoble:
I built this branch with `./configure && make -j $(nproc)`. Then I did
```sh
mkdir -p /tmp/proj/back
touch !$/a.{sh,cpp,h}
```

When I run `PATH=/bin:$PATH SHELL=/bin/bash src/vim --clean` and `:vimgrep /^/ /tmp/**/*.{cpp,h}`, I (correctly) get an error, because `shell=/bin/bash` which is too old.

When I run `SHELL=/usr/local/bin/bash src/vim --clean` and the same `:vimgrep`, I find both expect files in `:clist` (`shell=/usr/local/bin/bash` is new enough).

Seems like it's working to me.

zeertzjq:
This is failing Windows CI:
```
Failures: 
	From test_functions.vim:
	Found errors in Test_glob_extended_bash():
	command line..script D:/a/vim/vim/src/testdir/runtest.vim[601]..function RunTheTest[54]..Test_glob_extended_bash line 13: Expected ['Xtestglob/foo/bar/src/foo.cpp', 'Xtestglob/foo/bar/src/foo.h'] but got []
```

k-takata:
I've just created #13205.

Shane-XB-Qian:
// https://github.com/vim/vim/pull/13208

chrisbra:
Hm, I am slightly confused. According to [this](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md) Windows should have installed bash 5, so wondering why this actually fails 🤔 
For MacOS, it comes with bash 3 according to [here](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md)  so I would think the failure is expected here.

k-takata:
This PR only changed `mch_expand_wildcards()` in `os_unix.c`, so it won't affect Windows.

chrisbra:
ah yes, that is expected. I was confused because the app-runner installation talked about git-for-windows, so I automatically assumed we had a unix-like vim as well. Okay makes sense yes

benknoble:
> This PR only changed `mch_expand_wildcards()` in `os_unix.c`, so it won't affect Windows.

Instead of disabling the test on Windows, is it a reasonable effort to port the changes to the windows code?

chrisbra:
Well, how common is it to use a Windows Vim build with a Linux Shell?

PR #13149, opened 2023-09-22
Consistently use class/object variable and class/object method in the help text

yegappan:


mattn:
Sorry OT. I'm not native English speaker but I wonder "member method" can be just "member" ?

dkearns:
> Sorry OT. I'm not native English speaker but I wonder "member method" can be just "member" ?

I am, allegedly, a native English speaker and find this particular topic a bit confusing. :)

It's not uncommon in English to refer to a specific element of a set with the same name used, more generally, for the set itself.

So here we have "members" consisting of "member variables" and "member methods" but we might choose, by convention, to refer to the more commonly used of the two as just a "member".  

In this case there is not really any clear choice which makes it confusing, to me at least.

Bram often used "member" for "member variable" and "method" for "member method" but he also used other variations.  This might be the common usage in some circles but I haven't come across it.

Thanks for clarifying it @yegappan

mattn (approved):

PR #13155, opened 2023-09-22
Matching definitions in errors.h variable/method for class/object

RestorerZ:
Matching definitions in errors.h (and where these messages exist) variable/method for class/object


yegappan (commented):
You need to make the corresponding changes to the src/testdir/test_vim9_class.vim file.

errael:
#13135 is all about changing the text error messages for consistency within vim9class. Several of the messages modified by this PR are probably targeted; and the probable style that's emerging doesn't match some of these messages. 

Until the style issues are settled, I'm not sure it makes sense to modify messages that may soon be modified to meet the style guide. I'd probably solicit comments from member of vim, before undertaking a large task. Either way, it's not my call. After looking at the PR changing the program names not be that much of a merge problem

yegappan (approved):

PR #13159, opened 2023-09-23
lockvar on object or class

errael:
There are no defined semantics for using `lockvar` on an object or class.
Make it illegal for now, then it can possibly be allowed in the future.
(vim9 TODO item: how about lock/unlock?)

Currently it is OK to `lockvar` an `object` or `class`;  that doesn't prevent modification of an `object`'s variables.
 
```
final o = C.new()   # OK
const o = C.new()   # "cannot lock object"

final c = C         # OK
const c = C         # "cannot lock class"
```

Don't allow lockvar, even in situations where it might be OK, if `class`/`object` involved.

Doing `lockvar` on a variable holding a class is OK, but give an error since
can't easily distinguish between the following two `lockvar` statements.

```
var c = C
lockvar 0 c     # technically OK, but give error. Use final c = C
lockvar 0 C     # can't lock a class, error at any depth
```

To be consistent and simpler, don't allow lockvar with any object situation.

```
var o = C.new()
lockvar 0 o     # technically OK, but give error. Use final o = C.new()
lockvar! 0 o    # not defined to lock an object.
```

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3517,6 +3517,10 @@ EXTERN char e_member_str_type_mismatch_expected_str_but_got_str[]
 	INIT(= N_("E1382: Member \"%s\": type mismatch, expected %s but got %s"));
 EXTERN char e_method_str_type_mismatch_expected_str_but_got_str[]
 	INIT(= N_("E1383: Method \"%s\": type mismatch, expected %s but got %s"));
+EXTERN char e_cannot_lock_object[]
+	INIT(= N_("E1384: Cannot lock or unlock an object"));
+EXTERN char e_cannot_lock_class[]
+	INIT(= N_("E1384: Cannot lock or unlock a class"));
 #endif
 EXTERN char e_cannot_mix_positional_and_non_positional_str[]
 	INIT(= N_("E1400: Cannot mix positional and non-positional arguments: %s"));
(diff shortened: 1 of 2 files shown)


errael:
@yegappan Putting out this draft to get opinions.
Still need to address `final`/`const` as shown above, if that seems the right course.

Since most lockvar error messages say "lock or unlock", use that in this PR.
Not sure that's correct.


The following crashes, can include a fix in this PR. I can stop the
crash, but looking at it, it's one of those dict/object messes. Just output
a message where the crash occurs? What message?

```
vim9script

class C
    public this.v2: number = 2
endclass

var o1 = C.new()
lockvar! o1.v2      # <<<<< CRASH
```

errael:
Haven't run tests yet, but if there's a `list` or `dict` that has an object, then with this PR it probably can't be locked and will get an error, so might want a lockvar option that says don't lock `object`/`class`.

One option is to just document that `lockvar` ignores `object`/`class`. And if locking related to those two is wanted, introduce a new command or options to lockvar. This has the problem about what `const` should do.

errael:
Assumption is that default behavior of `lockvar` on an `object` or `class` is to _do nothing_, which is what it does now. There's a vim9class todo item
```
object empty(), len() - can class define a method to be used for them?
```
When/if some form of object locking is implemented, it can follow whatever pattern is set up for how a class specifies that it want to handle `empty()`/`len()`. Maybe there is a set of magically named interfaces that are implemented.

May open another PR with additional tests for `lockvar` and fixes as needed.

errael:
Opened #13174 as alternative

PR #13161, opened 2023-09-23
Don't allow changing the type of a v: variable using if_lua

zeertzjq:
Problem:  Can change the type of a v: variable using if_lua.
Solution: Add additional handling of v: variables like :let.

### DIFF
--- a/src/if_lua.c
+++ b/src/if_lua.c
@@ -1900,6 +1900,16 @@ luaV_setvar(lua_State *L)
 	}
 	else
 	{
+	    int type_error = FALSE;
+	    if (dict == get_vimvar_dict()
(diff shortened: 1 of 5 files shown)


src/if_lua.c
```diff
@@ -1900,6 +1900,16 @@ luaV_setvar(lua_State *L)
...
 	else
 	{
+	    int type_error = FALSE;
+	    if (dict == get_vimvar_dict()
+	       && !before_set_vvar((char_u *)name, di, &tv, TRUE, &type_error))
+	    {
+		clear_tv(&tv);
+		if (type_error)
+		    return luaL_error(L,
+				"Setting v:%s to value with wrong type", name);
```
zeertzjq, line 1910: Other error messages here aren't translated and don't have an error number, so treat type error message similarly.

Issue #13154, opened 2023-09-22
[vim9script] E1360: Using a null object (regression after patch 9.0.1914)


lifepillar:
### Steps to reproduce

Patch 9.0.1914 has introduced a regression in my parser library. I have reduced the [failing test](https://github.com/lifepillar/vim-devel/blob/23a117788c899aeb91b0edce983f4039639468a4/start/libparser/test/test_libparser.vim#L142) to the following script:

```vim
vim9script

class Context
endclass

class Result
endclass

def Failure(): Result
  return Result.new()
enddef

def Bol(ctx: Context): Result
  return Failure()
enddef

def Test_Bol()
  var ctx = Context.new()
  var result = Bol(ctx)
enddef

Test_Bol()
```

Sourcing the script above results in `E1360: Using a null object`.

### Expected behaviour

The script should not produce any error.  That is the case with v9.0.1913 or earlier.

### Version of Vim

9.0.1927

### Environment

macOS 13
Apple Terminal
xterm-256color
ZSH 5.9

PR #13163, opened 2023-09-23
Check for a null object only when calling an object method

yegappan:
Fixes the issue reported in #13154.


chrisbra:
This seems to cause a stack error: https://github.com/vim/vim/actions/runs/6297960978/job/17095936823#step:16:13778

looks like we are popping the stack 1 times to often. I haven't yet found out where exactly this happens.

PR #13166, opened 2023-09-23
fix: clang compilation warnings with -Wstrict-prototypes

dpelle:
Change fixes this kind of compilation warnings with clang-17:
```
proto/if_python3.pro:13:20: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
   13 | int python3_version();
      |                    ^
      |                     void
```

### DIFF
--- a/src/debugger.c
+++ b/src/debugger.c
@@ -1000,7 +1000,7 @@ has_profiling(
 }
 
     static void
-prof_clear_cache()
+prof_clear_cache(void)
 {
     if (!prof_cache_initialized)
     {
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -2070,7 +2070,7 @@ set_ref_in_python3(int copyID)
 }
 
     int
-python3_version()
+python3_version(void)
 {
 #ifdef USE_LIMITED_API
     return Py_LIMITED_API;
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -1858,7 +1858,7 @@ ex_xrestore(exarg_T *eap)
     static int
 test_x11_window(Display *dpy)
 {
-    int			(*old_handler)();
+    int			(*old_handler)(Display*, XErrorEvent*);
     XTextProperty	text_prop;
 
     old_handler = XSetErrorHandler(x_error_check);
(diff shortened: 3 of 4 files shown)


pheiduck:
Interesting how do you compile vim?
I have made a PR to get clang-17 on the CI but it runs fine on latest patch.:
https://github.com/vim/vim/pull/12745

dpelle:
@pheiduck  asked:

> Interesting how do you compile vim?
> I have made a PR to get clang-17 on the CI but it runs fine on latest patch.:
> https://github.com/vim/vim/pull/12745

To reproduce the warning with clang-17, I had uncommented out this line to enable `-Wall -Wmissing-prototypes`:

```
$ git diff
diff --git a/src/Makefile b/src/Makefile
index 4d0a5ebad..a5c0d9dea 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -602,7 +602,7 @@ CClink = $(CC)
 #CFLAGS = -g -O2 -fno-strength-reduce -Wall -Wmissing-prototypes
 #CFLAGS = -g -Wall -Wmissing-prototypes
 #CFLAGS = -O6 -fno-strength-reduce -Wall -Wshadow -Wmissing-prototypes
-#CFLAGS = -g -DDEBUG -Wall -Wshadow -Wmissing-prototypes
+CFLAGS = -g -DDEBUG -Wall -Wshadow -Wmissing-prototypes
 #CFLAGS = -g -O2 '-DSTARTUPTIME="vimstartup"' -fno-strength-reduce -Wall -Wmissing-prototypes

 # Use this with GCC to check for mistakes, unused arguments, etc.
```

I now see that the warning is not specific to clang-17 actually, it also happens with clang-14, clang-15, gcc-9, gcc-11 for example, so I'll reword the commit message.

ychin:
I think if certain warnings are important, such as `-Wmissing-prototypes` and `-Wstrict-prototypes`, we should turn them on in ci/config.mk.sed or ci/config.mk.gcc.sed or ci/config.mk.clang.sed. This way it will actually checked in CI.

Otherwise it's just whack-a-mole every time. We should either decide if such warnings are important (in which case we check them in CI), or unimportant (in which case we don't bother fixing them).

ychin:
Hmm, I tried adding this to CI and seems like GTK and Ruby headers don't really like them, so it would be a little annoying to just add them in unless we also add warning suppression when including those headers…

PR #13167, opened 2023-09-24
runtime(csh): Fix `:unlet` error in ftplugin

roccomao:
It can be said that csh is a subset of tcsh. In *ftplugin/tcsh.vim*:

https://github.com/vim/vim/blob/ceee7a808ce82b0c6bd84e0b6fc1dfb0475c99aa/runtime/ftplugin/tcsh.vim#L17

But if the current file type is tcsh, and we manually switch to csh through `:setlocal filetype=csh` (or when switching to any file type, even tcsh to tcsh), an error will be reported:

```
Error detected while processing FileType Autocommands for "*"..function <SNR>13_LoadFTPlugin:
E108: No such variable: "b:browsefilter"
```

or

```
Error detected while processing FileType Autocommands for "*"..function <SNR>13_LoadFTPlugin:
E108: No such variable: "b:match_words"
```

### DIFF
--- a/runtime/ftplugin/csh.vim
+++ b/runtime/ftplugin/csh.vim
@@ -38,13 +38,13 @@ if exists("loaded_matchit") && !exists("b:match_words")
 	\   s:line_start .. 'case\s\+:' .. s:line_start .. 'default\>:\<breaksw\>:' ..
 	\   s:line_start .. 'endsw\>'
   unlet s:line_start
-  let b:undo_ftplugin ..= " | unlet b:match_words"
+  let b:undo_ftplugin ..= " | unlet! b:match_words"
 endif
 
 if (has("gui_win32") || has("gui_gtk")) && !exists("b:browsefilter")
   let  b:browsefilter="csh Scripts (*.csh)\t*.csh\n" ..
 	\	      "All Files (*.*)\t*.*\n"
-  let b:undo_ftplugin ..= " | unlet b:browsefilter"
+  let b:undo_ftplugin ..= " | unlet! b:browsefilter"
 endif
 
 let &cpo = s:save_cpo


dkearns:
Thanks.

The variable handling in the tcsh ftplugin doesn't look right. E.g., `b:browsefilter` is being unconditionally unset and this was already conditionally handled by the sourced csh ftplugin.

I'll come up with a more general fix in the next couple of days when I'm back at a dev box.

I haven't done anything with those tcsh files since I took over maintainership so if you have any long standing annoyances beyond this let me know.  It's good to know they're still being used. :)

dkearns:
You can set the `g:filetype_csh` variable to override the filetype detection if that suits.  See `:help g:filetype_csh`

roccomao:
Thank you for your reply.

I'm not very clear about the entire logic of built-in filetype detection, but when I searched the entire *runtime/ftplugin* folder, similar `:unlet` code appeared only in *csh.vim* and *cmake.vim*, and an error may be reported, while cmake did not have any problems. In addition, `:unlet!` is used in all variables about `b:undo_ftplugin` in other files, so here I simply added an `!` to suppress the output of the error message. And this indeed solved the problem I described above, at least on the surface it looks OK. So I look forward to your better solution, thanks.

Regarding the immediate solution, this is not what bothers me, I can use `silent! setlocal filetype=xxx` too. I can share the case where I discovered the problem: I had a scratch buffer in which I might temporarily write various languages, and for correct highlighting, I would set the filetype. This all worked fine until I set up tcsh filetype, no matter what filetype I manually switch to at this time, an error will be reported.

dkearns:
I haven't forgotten.  I'll fix it today...or tomorrow.  Please ping me if it slips.

PR #13168, opened 2023-09-24
Add a missing LT to the help of strutf16len()

a5ob7r:


chrisbra:
thanks, seems to be right!

PR #13171, opened 2023-09-24
fix(ftplugin/sh): do not invoke bash for kp unless `b:is_bash` is true

e-kwsm:
Problem: `keywordprg` is set to `Help` which invokes bash, even though it is not a bash file
Solution: use `Help` only when it is a bash file

You should not assume that sh is bash in terms of portability.

### DIFF
--- a/runtime/ftplugin/sh.vim
+++ b/runtime/ftplugin/sh.vim
@@ -39,8 +39,7 @@ if (has("gui_win32") || has("gui_gtk")) && !exists("b:browsefilter")
   let b:undo_ftplugin ..= " | unlet! b:browsefilter"
 endif
 
-if (exists("b:is_bash") && (b:is_bash == 1)) ||
-      \ (exists("b:is_sh") && (b:is_sh == 1))
+if (exists("b:is_bash") && (b:is_bash == 1))
   if !has("gui_running") && executable("less")
     command! -buffer -nargs=1 Help silent exe '!bash -c "{ help "<args>" 2>/dev/null || man "<args>"; } | LESS= less"' | redraw!
   elseif has('terminal')


Shane-XB-Qian:
sh is just some distr like to link to dash, but does not mean it is dash always.
but used to treat xxx.sh is sh or bash in most cases, so why/what this PR for?

runtime/ftplugin/sh.vim
```diff
@@ -39,8 +39,7 @@ if (has("gui_win32") || has("gui_gtk")) && !exists("b:browsefilter")
   let b:undo_ftplugin ..= " | unlet! b:browsefilter"
 endif
 
-if (exists("b:is_bash") && (b:is_bash == 1)) ||
-      \ (exists("b:is_sh") && (b:is_sh == 1))
```
Shane-XB-Qian, line 43: this check was correct per tradition or history as my knowledge.

Shane-XB-Qian:
only one thing:  @cecamp if user did not set g:is_dash and the shebang was not dash, i think we should respect the used-way which is the foo.sh (there no shebang) should be sh which should be NOT judged/reply on its bin link?

chrisbra:
> You should not assume that sh is bash in terms of portability.

May be we shouldn't, but perhaps executing `bash`s help command isn't wrong after all?. Perhaps we just need instead a 

```
&& executable('bash')
```
instead?

e-kwsm:
> May be we shouldn't, but perhaps executing `bash`s help command isn't wrong after all?

It can show bash-extensions. For example, the POSIX options of `cd` is `-L` and `-P`, but bash adds `-e` and `-@`.

chrisbra:
okay, fine

dkearns:
I didn't get a chance to look at this earlier but this seems a bit baby and bathwater to me.

I understand the issue here but dropping support for everything other than bash is less than ideal.  Am I missing something or does it need to be added?

PR #13172, opened 2023-09-24
fix: correct sh detection

Shane-XB-Qian:
// https://github.com/vim/vim/pull/13171#issuecomment-1732624524

### DIFF
--- a/runtime/syntax/sh.vim
+++ b/runtime/syntax/sh.vim
@@ -20,24 +20,6 @@ elseif getline(1) =~ '\<bash\>'
  let b:is_bash      = 1
 elseif getline(1) =~ '\<dash\>'
  let b:is_dash      = 1
-elseif !exists("g:is_kornshell") && !exists("g:is_bash") && !exists("g:is_posix") && !exists("g:is_sh") && !exists("g:is_dash")
- " user did not specify which shell to use, and 
- " the script itself does not specify which shell to use. FYI: /bin/sh is ambiguous.
- " Assuming /bin/sh is executable, and if its a link, find out what it links to.
- let s:shell = ""
- if executable("/bin/sh")
-  let s:shell = resolve("/bin/sh")
- elseif executable("/usr/bin/sh")
-  let s:shell = resolve("/usr/bin/sh")
- endif
- if     s:shell =~ '\<ksh\>'
-  let b:is_kornshell= 1
- elseif s:shell =~ '\<bash\>'
-  let b:is_bash = 1
- elseif s:shell =~ '\<dash\>'
-  let b:is_dash = 1
- endif
- unlet s:shell
 endif
 
 " handling /bin/sh with is_kornshell/is_sh {{{1
@@ -49,31 +31,31 @@ endif
 " respectively.
 if !exists("b:is_kornshell") && !exists("b:is_bash") && !exists("b:is_dash")
   if exists("g:is_posix") && !exists("g:is_kornshell")
-   let g:is_kornshell= g:is_posix
+    let g:is_kornshell = g:is_posix
   endif
   if exists("g:is_kornshell")
-    let b:is_kornshell= 1
+    let b:is_kornshell = 1
     if exists("b:is_sh")
       unlet b:is_sh
     endif
   elseif exists("g:is_bash")
-    let b:is_bash= 1
+    let b:is_bash = 1
     if exists("b:is_sh")
       unlet b:is_sh
     endif
   elseif exists("g:is_dash")
-    let b:is_dash= 1
+    let b:is_dash = 1
     if exists("b:is_sh")
       unlet b:is_sh
     endif
   else
-    let b:is_sh= 1
+    let b:is_sh = 1
   endif
 endif
 
 " if b:is_dash, set b:is_posix too
 if exists("b:is_dash")
- let b:is_posix= 1
+ let b:is_posix = 1
 endif
 
 " set up default g:sh_fold_enabled {{{1


cecamp:
Well, this "fix" would definitely be problematic. You've taken out the section where syntax/sh.vim is trying to figure out which shell it should use when the user hasn't specified one.

Shane-XB-Qian:
> You've taken out the section where syntax/sh.vim is trying to figure out which shell it should use when the user hasn't specified one.

i can not get you, what you mean? 
or anyway this is Try To make no shenbang 'foo.sh' have and keep a stable default b: var, which not rely on bin link,
e.g ubuntu is default link sh to dash, but user like to keep it as sh at that, or he moved to another machine which no dash,
unless g: var override it.

chrisbra:
I don't understand what problem you are trying to solve here

Shane-XB-Qian:
please refer that reference comment in ticket description.

is it ok to rely on bin link to detect sh type?
specially how to make your one no shenbang `sh` file behavior and syntax same in different machines?
I supposed it should default to the one it used-to, unless user g: var setting overrode it by need.

--
shane.xb.qian

Shane-XB-Qian:
anyway let @cecamp make decision, I am ok to pull back this PR, to this one I am just trying to help clarify but it's not my goal to sell it :smile:

--
shane-xb-qian

fritzophrenic:
> please refer that reference comment in ticket description. is it ok to rely on bin link to detect sh type? specially how to make your one no shenbang `sh` file behavior and syntax same in different machines? I supposed it should default to the one it used-to, unless user g: var setting overrode it by need.

Sorry this is still very unclear. I am sure we are dealing with a language barrier here. Can you maybe illustrate the problem with an example? What *specifically* is Vim doing that you don't think it should do, and why?

By "that reference comment in ticket description" I assume you are talking about this? https://github.com/vim/vim/pull/13171#issuecomment-1732624524

Reading that comment I still have no idea what you are trying to accomplish by removing shell detection.

Shane-XB-Qian:
> still have no idea what you are trying to accomplish by removing shell detection.

it did Not remove shell detection, please note it just to make b:is_xxsh not rely on symblink, and did not change ft.
please image you are writing a foo.sh in your ubuntn and macos and debian/suse, you may find it highlight different.

dkearns:
@Shane-XB-Qian is, I think, arguing that the installed binary isn't relevant for the purposes of filetype detection.  Which is certainly a reasonable perspective.

There's not necessarily any relationship between the file that's being edited and the system it's being edited on.  If the dialect can't be determined from the file content or a specific user configuration override then it should just default to plain old sh.

Shane-XB-Qian:
@dkearns cool, you are my man, oh no, my bro.

fritzophrenic:
> @Shane-XB-Qian is, I think, arguing that the installed binary isn't relevant for the purposes of filetype detection. Which is certainly a reasonable perspective. There's not necessarily any relationship between the file that's being edited and the system it's being edited on.

From the comments in the code, and from what the code is actually doing, it *looks* like the removed code is intended to account for when the script file is specifying /bin/sh or /usr/bin/sh. It seems a little strange to just assume "sh" in the case that the script doesn't specify *any* shell but maybe there is a good reason for that. Regardless, per the comments in the code if the script had a "#!/bin/sh" then figuring out which shell that actually points to on the installed system actually makes a lot of sense. Presumably this script is intended to actually run on this system at some point in most cases.

chrisbra:
> please image you are writing a foo.sh in your ubuntn and macos and debian/suse, you may find it highlight different.

But this makes sense to highlight depending on what your actual shell does support, no?

Shane-XB-Qian:
> > please image you are writing a foo.sh in your ubuntn and macos and debian/suse, you may find it highlight different.
> 
> But this makes sense to highlight depending on what your actual shell does support, no?

i think no:  `there's not necessarily any relationship between the file that's being edited and the system it's being edited on`, e.g you are editting a java code but didnot meant you had to install jdk first (-_- perhaps not a good example :lol:)
// you can control it via g:is_xxsh (or shenbang) anyway.

chrisbra:
I would think differently. We are changing the current default behaviour when it's not possible to know what shell is being used for `/bin/sh`. I assume this would mean, the syntax highlighting will use default POSIX highlighting only and not consider bash constructs, even if `/bin/sh` is a symlink to `/usr/bin/bash` and so is totally fine with such special constructs. 

Especially for shell scripts, I think it would be okay to detect the current shell, because unless you are writing a larger or more complex shell script, I would think you do it on the system where you will run it (so you can also test it etc). At least that's how I do it usually. And for complex or larger shell scripts, I would think you carefully decide what shell to use in your shebang line anyhow (or you at least set one of those configuration variables).

Shane-XB-Qian:
> I would think differently. We are changing the current default behaviour when it's not possible to know what shell is being used for `/bin/sh`. I assume this would mean, the syntax highlighting will use default POSIX highlighting only and not consider bash constructs, even if `/bin/sh` is a symlink to `/usr/bin/bash` and so is totally fine with such special constructs.

i think you are thinking it just opposite to it, foo.sh used to be treated it as sh (if no shenbang), meant `sh` is `sh`, 
game changed just recently i guess it is ubuntu changed sh to link it to dash, that's the one changed default behavior,
i am just trying to recover the tradition or history, that's it.

chrisbra:
yeah exactly, a lot of people use bashism even if using `/bin/sh` in the shebang line (which is okay, if your `/bin/sh` is a symlink to bash, because it will happily accept it, even though this is not considered good style).

Shane-XB-Qian:
problem is e.g ubuntu changed it link to dash recently (note: it is a 'd'), and please check the case which reported in #13171 , now it is beginning to break the game. so I or we better to rescue the game. :smile:


--
shane.xb.qian

chrisbra:
> problem is e.g ubuntu changed it link to dash recently (note: it is a 'd'), and please check the case which reported in [#13171](https://github.com/vim/vim/pull/13171)

I remember that, but that was like 10 years ago, so it is not recently. And I checked #13171 and I think it's wrong over there and one should rather use `exectuable("bash")` instead as mentioned over there.

Shane-XB-Qian:
I would say you probably remember not correctly and besides it normally would be a bit delayed from dev version to LTS version to end users, and I recalled that #13171 was for or tried to fix another ticket which there is a difference between dash and sh/bash, so your mentioned exectuable('bash') was Not the problem or solution.
anyway, I wish sh back to sh (and indeed generally it was a subset of bash).
// BTW: in Linux world, my opinion, vim & bash & firefox is his/her faith. :smile:

--
shane.xb.qian

Shane-XB-Qian:
seems clarify done.

PR #13181, opened 2023-09-25
Fix memory leak

yegappan:


chrisbra:
thanks, I started looking yesterday about it, but then got side-tracked by some other things.

PR #13182, opened 2023-09-25
Add cmdline tab-completion for setting string options

ychin:
Add tab-completion for setting string options on the cmdline using `:set=` (along with `:set+=` and `:set-=`).

The existing tab completion for setting options currently only works when nothing is typed yet, and it only fills in with the existing value, e.g. when the user does `:set diffopt=<Tab>` it will be completed to `set diffopt=internal,filler,closeoff` and nothing else. This isn't too useful as a user usually wants auto-complete to suggest all the possible values, such as 'iblank', or 'algorithm:patience'.

For set= and set+=, this adds a new optional callback function for each option that can be invoked when doing completion. This allows for each option to have control over how completion works. For example, in 'diffopt', it will suggest the default enumeration, but if `algorithm:` is selected, it will further suggest different algorithm types like 'meyers' and 'patience'. When using set=, the existing option value will be filled in as the first choice to preserve the existing behavior (this helps doing small edits to the option value). When using set+= this won't happen as it doesn't make sense to append an option by its own value in general.

For flag list options (e.g. 'mouse' and 'guioptions'), completion will take into account existing typed values (and in the case of set+=, the existing option value) to make sure it doesn't suggest duplicates.

For set-=, there is a new `ExpandSettingSubtract` function which will handle flag list and comma-separated options smartly, by only suggesting values that currently exist in the option so you won't try to subtract the wrong value.

Note that Vim has some existing code that adds special handling for 'filetype', 'syntax', and misc dir options like 'backupdir'. This change preserves them as they already work, instead of converting to the new callback API for each option.


ychin:
Marking this as draft to gather some feedback first. I still need to do a couple things but would like to the basic feedback before proceeding. The couple things are:

- [ ] Implement more options. Currently I only implemented a few as a proof of concept, and I tried to cover the ones that have complicated auto-complete rules like `diffopt` and `spellsuggest` to make sure they work.
- [ ] Add tests
- [ ] Documentation

For the API design, I decided to add a generic function to each option (similar to what we did with `opt_did_set_cb_T`) that can do any expansion logic they like. I originally thought about an API that just returns a list of completable strings instead of doing the regex matching itself (this is similar to how user functions work when using `:command -complete`) but I decided it may not provide enough flexibility, as the callback function may want to call other existing expansion functions.

I also decided not to touch 'filetype', 'syntax', 'backupdir' etc for now. Those have special handling and I left them be.

There are also some interesting behaviors I found with commas. E.g. if you have a file with a comma in it (e.g. `file,with,comma`, auto-complete currently fills the name as-is, which is wrong as it should be `file\\,with\\,comma` instead), but I don't want to fix that for now. There are other minor existing comma parsing issues. May file PR after this to fix it.

ychin:
Some example screenshots:

`set diffopt=<Tab>`:

image

(special handling for individual algorithms):

image

`set encoding=u<Tab>`:

image


`set-=`:

(note that these are grabbing the existing values that is set by the user and spit by comma)

image

(for flags, it also only presents the existing option values only. Example below has `:set mouse=vn`)

image

chrisbra:
Oh, I really like this

benknoble:
Awesome! Now I wonder if it’s possible to do something like `completeopt`’s `popup` for info: that is, when suggesting myers or a shortmess flag, display (a snip of) the relevant help text?

Probably would only be enabled for `wop` containing `pum` (bah, nonsense syllables).

ychin:
> Awesome! Now I wonder if it’s possible to do something like `completeopt`’s `popup` for info: that is, when suggesting myers or a shortmess flag, display (a snip of) the relevant help text?
> 
> Probably would only be enabled for `wop` containing `pum` (bah, nonsense syllables).

That's a good idea! It's definitely possible within this framework to be added __later__. I think the callback can return some info on the `opt_expand_cb_T` struct which tells Vim what to show in the popup menu (e.g. a help tag).

The more annoying problem is actually figuring out what to show. The 'shortmess' option has individiual help tags for each flag (e.g. `shm-i`), but not so for most other options (flags or comma-separated). I don't think we want each expansion callback to be overly long or intricate and having to do complicated logic, so I think a generic way to find the correct snippet could be implemented to show the relevant bits given some basic data (e.g. "diffopt" -> "meyers" will find the relevant docs and determine how many lines it is).

I do think if we want to do this (in a later commit), we may as well go all the way. Make it so that we could add popup docs for completing option names (e.g. `:set dif<Tab>`), or command names (`:nno<Tab>`), and user-provided completions when they define their own commands. Also, would be nice to expose that to `getcompletion()`, but I think that is more secondary.

benknoble:
Yep, agreed with all you said, esp. the bits about _later_ :)

src/option.c
```diff
@@ -7348,8 +7350,16 @@ set_context_in_set_cmd(
 	}
     }
     // handle "-=" and "+="
+    expand_option_append = FALSE;
+    int expand_option_subtract = FALSE;
     if ((nextchar == '-' || nextchar == '+' || nextchar == '^') && p[1] == '=')
     {
+	if (nextchar == '-')
+	{
```
yegappan: To be consistent the style followed for the next if statement, can you remove the braces for this if statement?

src/option.c
```diff
@@ -7418,34 +7453,57 @@ set_context_in_set_cmd(
...
+		    break;
+		}
 	    }
 	}
+    }
 
+    // An option that is a list of single-character flags should always start
+    // at the end as we don't complete words.
+    if (flags & P_FLAGLIST)
+    {
```
yegappan: Even though, I personally prefer always using curly braces for an if statement, the coding style used in the Vim source base is to not use the curly braces if there is only one statement in the if condition.

src/option.c
```diff
@@ -7739,33 +7826,204 @@ ExpandOldSetting(int *numMatches, char_u ***matches)
...
-	    STRMOVE(var, var + 1);
-#endif
-
-    *matches[0] = buf;
+    (*matches)[0] = buf;
     *numMatches = 1;
     return OK;
 }
 
+// Expansion handler for :set=/:set+= when the option has a custom expansion handler.
```
yegappan: For function headers, the style is to use the three part "/* */" comments.

yegappan (commented):
This is very nice.  Thanks for working on this.  This will require quite a few new tests to make sure all the expansion cases are covered.

src/option.c
ychin: I can fix up the misc code styles issues. I do think it will be nice if at some point Vim can add a proper code formatter/linter that we can just run instead of having to manually conform. It's always a fair bit of work for the initial setup and code cleanups though so not suggesting right now.

ychin:
Just pushed an update. I have now implemented all the options that I intended to. Among them, most of them are simple enumeration type options, or flags, and just consist of a few lines of additions. Some other options are a little more involved, and I tried to split the commits up to make them easier to review. If it's too much, I could split them into separate PRs to do later.

Among them, `highlight` is probably the most complicated, but `diffopt` / `previewpopup` / `completepopup` / `keyprotocol` are non-trivial as well.

Among the remaining options, I think some of them could have auto-complete added to them but I don't feel like adding them for now:

* `mouseshape` / `guicursor` (possible, just need custom parsing for the dashes)
* Misc font options like `guifont` (platform-dependent)
* `renderoptions` (too niche)
* `viminfo` (complicated to parse)
* `statusline` (just… no)

src/testdir/test_options.vim
```diff
@@ -360,13 +360,17 @@ func Test_set_completion()
...
+  call feedkeys(":set spellsuggest=best,file:test_options.v\<Tab>\<C-B>\"\<CR>", 'xt')
+  call assert_equal("\"set spellsuggest=best,file:test_options.vim", @:)
 
   " Expand value for 'key'
   set key=abcd
   call feedkeys(":set key=\<Tab>\<C-B>\"\<CR>", 'xt')
   call assert_equal('"set key=*****', @:)
+  call feedkeys(":set key-=\<Tab>\<C-B>\"\<CR>", 'xt')
+  call assert_equal('"set key-=*****', @:)
   set key=
```
zeertzjq, line 374: This seems to also need `if has('+key')`. 'key' option doesn't work if `FEAT_CRYPT` isn't enabled.

chrisbra: I would remove the whole completion for `set key` (I mean what sense does it make to autocomplete `*****`) and I also don't want to accidentally leak the crypt-key. ah this already seems to happen with the current vim. So probably fine. Thanks!

ychin: I was actually going to suggest that in a separate PR! Since I was looking into cmdline completion, I think `key` needs some hardening.

I was going to propose removing completion for `set key`, and also remove `set-=` / `set+=` / `set^=`. Those have some security ramifications since they allow you to guess a substring of the key. We can just keep `set=`.

chrisbra: thanks, yes makes sense!

src/testdir/test_options.vim
```diff
@@ -382,6 +386,278 @@ func Test_set_completion()
   call assert_equal('"set syntax=' .. getcompletion('a*', 'syntax')->join(), @:)
 endfunc
 
+" Test handling of expanding individual string option values
+func Test_set_completion_string_values()
```
zeertzjq, line 390: Most of the `assert_equal()` calls in this function have wrong order of arguments. This will make error message confusing if test fails.

ychin: Oh right… the first item is the expected value. I internalized the weirdness when I was debugging this but didn't correct it. I could fix it in a follow-up commit.

src/screen.c
```diff
@@ -4956,6 +4955,32 @@ set_listchars_option(win_T *wp, char_u *val, int apply)
...
+ * 'listchars' option.
+ */
+    char_u *
+get_listchars_name(expand_T *xp UNUSED, int idx)
+{
+    if (idx >= (int)(sizeof(lcstab) / sizeof(lcstab[0])))
+	return NULL;
+
+    return (char_u*)lcstab[idx].name;
+}
```
zeertzjq, line 4982: These don't include `multispace` and `leadmultispace`

zeertzjq: Created PR #13225 to fix this.

src/option.c
```diff
@@ -1837,7 +1837,7 @@ stropt_get_newval(
 					     &(options[opt_idx]), OPT_GLOBAL));
     else
     {
-	++arg;	// jump to after the '=' or ':'
+	++arg;	// joption_value2stringump to after the '=' or ':'
```
zeertzjq, line 1840: Looks like an accidental mistake

src/option.c
```diff
@@ -7418,34 +7451,55 @@ set_context_in_set_cmd(
...
+	    if (*p == ' ' || *p == ',' ||
+		    (*p == ':' && (flags & P_COLON)))
 	    {
-		xp->xp_pattern = p + 1;
-		break;
+		s = p;
+		while (s > xp->xp_pattern && *(s - 1) == '\\')
+		    --s;
+		if ((*p == ' ' && (xp->xp_backslash == XP_BS_THREE && (p - s) < 3))
+			|| (*p == ',' && (flags & P_COMMA) && ((p - s) % 1) == 0)
```
zeertzjq, line 7471: This `(p - s) % 1` looks strange. Isn't any number module 1 always 0?

src/option.c
ychin: Yeah it was pointed out in https://github.com/vim/vim/commit/900894b09a95398dfc75599e9f0aa2ea25723384#r128808467 as well. I have a follow-up PR that will fix this.

src/option.c
ychin: You are absolutely right. I only moved this code around and added the `P_COLON` check below it but I didn't change the logic for handling commas here. As you said, the logic here is wrong. I didn't want the scope of PR to grow any more which is why I didn't fix it.

I think the correct logic here is actually `(p - s) >= 2`. Currently, if you have a file/folder of the name `foo,bar`, the correct way to specify it in an option is do so something like `:set backupdir=dir1,foo\\,bar,dir2` (note that this is equivalent to `:set backupdir=dir1,foo\\\,bar,dir2`). Note that you will see that in the test cases that I added, they actually check that `set-=` will correctly escape the commas to insert those "\\" when doing `set path-=<Tab>`.

Note: there's another cmd auto-complete bug (that's why I wanted to fix this in another PR instead of growing in scope). In the above example, if you do auto-complete as follows: `set backupdir=fo<Tab>`, if will complete to `set backupdir=foo,bar`, forgetting to insert the `\\` there. In general the option setting syntax is a little weird regarding commas, as I believe the syntax is incomplete and doesn't allow certain types of names to be entered (you cannot enter a folder/file that ends with "\", because that slash will end up escaping the comma instead).

tldr: Yes it's broken, but it's broken before my change and I could submit another PR to fix it.

yegappan:
The following warning messages are displayed in the Appveyor CI build (introduced by this PR):

```
option.c(7992): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(866): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(1768): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(2453): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(2939): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(2950): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
optionstr.c(2961): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
```

ychin:
> The following warning messages are displayed in the Appveyor CI build (introduced by this PR):
> 
> ```
> option.c(7992): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(866): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(1768): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(2453): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(2939): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(2950): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> optionstr.c(2961): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
> ```

Ok, I will fix them up in #13237. Found a couple other places that trigger warnings as well if I turn on `-Wsign-conversion`. I wish warnings like these are turned into errors though (like in the main GitHub CI build). Otherwise it's quite hard to find out.

src/option.c
ychin: The commas will be fixed in #13303

PR #13186, opened 2023-09-25
Check for a null object before adjusting the execution stack

yegappan:


chrisbra:
hm, that seems to be reverting PR #13163 (patch 9.0.1932). Won't that cause the same issue, that this particular patch was trying to fix?

yegappan:
No.  The relevant fix for #13154 is the check for `IS_OBJECT_METHOD(ufunc)`.  That is retained in this PR.  There is a new test to check for the issue reported in #13154, which will fail if this fix is not correct.

Issue #13185, opened 2023-09-25
List of implemented interfaces does not require whitespace like other lists


dkearns:
### Steps to reproduce

```vim
interface I
endinterface
interface J
endinterface

class A implements I,J
endclass
```

### Expected behaviour

An error message: `E1069: White space required after ',': ,J`

Vim9 requires strict whitespace separation of list items so this should also be required for a list of implemented interfaces.

### Version of Vim

9.0.1941

### Environment

Ubuntu 20.04

chrisbra:
fixed by 2dede3dbfa3cb52f464f942d46d3ec0f66e3e354

Issue #13184, opened 2023-09-25
public modifier allowed in static but not object method declarations


dkearns:
### Steps to reproduce

```vim
class A
  def Foo()
  enddef
  public def Bar()
  enddef
endclass
```

Executing this script results in an error: `E1331: Public must be followed by "this" or "static"`

```vim
class A
  static def Foo()
  enddef
  public static def Bar()
  enddef
endclass
```
No error is output.

### Expected behaviour

Class and object method declarations should exhibit the same behaviour with regard to the `public` modifier's presence or absence.

### Version of Vim

9.0.1941

### Environment

Ubuntu 20.04

yegappan:
This issue should be addressed in the latest Vim.  Can you try the latest version?

yegappan:
This issue should be fixed now.  Can you try the latest version and close this issue?

dkearns:
Sorry, I missed your comment last week. 

In 9.0.2018 I see `E1331: Public must be followed by "this" or "static"` for object methods and `E1388: Public keyword not supported for a method` for class methods.

Shouldn't they be the same error message, or at least very similar?  I think `Public` should also be lower case.

PR #13188, opened 2023-09-26
Add additional checks when parsing a class definition

### LINKED ISSUE #13183: :def (list all functions) is executed in a class declaration
### Steps to reproduce

```vim
vim9script
class A
  def
endclass
```

When this script is executed :def is executed without error, listing all functions.  At least `static def` and `def Foo` also execute without error.

### Expected behaviour

An error message: `E1318: Not a valid command in a class: def`

### Version of Vim

9.0.1941

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Also, support only read-only object variables in an interface.

Fixes #13183, #13184 and #13185.

### DIFF
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -4479,6 +4479,9 @@ E1380	vim9class.txt	/*E1380*
 E1381	vim9class.txt	/*E1381*
 E1382	vim9class.txt	/*E1382*
 E1383	vim9class.txt	/*E1383*
+E1387	vim9class.txt	/*E1387*
+E1388	vim9class.txt	/*E1388*
+E1389	vim9class.txt	/*E1389*
 E139	message.txt	/*E139*
 E140	message.txt	/*E140*
 E1400	builtin.txt	/*E1400*
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -464,10 +464,10 @@ The interface name can be used as a type: >
 	   echo $'the surface is {shape.Surface()}'
 	endfor
 <
-						*E1378* *E1379* *E1380*
-An interface can have only instance variables (read-only and read-write
-access) and methods.  An interface cannot contain private variables, private
-methods, class variables and class methods.
+					*E1378* *E1379* *E1380* *E1387*
+An interface can contain only object methods and read-only object variables.
+An interface cannot contain read-write and private object variables, private
+object methods, class variables and class methods.
 
 An interface can extend another interface using "extends".  The sub-interface
 inherits all the instance variables and methods from the super interface.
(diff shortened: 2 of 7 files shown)


errael:
About getting rid of the read/write variables in an interface.

Assuming this is about properties, there are 4 cases
```
getProperty, setProperty, read-only variable, read-write variable.
```
If property is part of your syntax, you could probably specify `read-propery` and `write-property` as separate features of the property, might even be able to specify `direct access` or `function access`.

In early discussions, I only saw a need for read-only; at least for how I would use it. The use cases I considered didn't need read-write; when changing a property, I figured you'd always want to do something.

Thinking about it again, I thought of what might be a common case: a simple bulletin board. Something with no code; state info only; write anything you want; no code for conflict resolution; last writer wins. A `read-write` property fits the bill.

I don't understand why take away the 1 of 4 property access methods? Why not leave it up to the programmer? Once there are variables in the interface...

I don't think it's a big deal either way. If someone really needs it, probably easy to add.

yegappan:
> About getting rid of the read/write variables in an interface.
> 
> Assuming this is about properties, there are 4 cases
> 
> ```
> getProperty, setProperty, read-only variable, read-write variable.
> ```
> 
> If property is part of your syntax, you could probably specify `read-propery` and `write-property` as separate features of the property, might even be able to specify `direct access` or `function access`.
> 
> In early discussions, I only saw a need for read-only; at least for how I would use it. The use cases I considered didn't need read-write; when changing a property, I figured you'd always want to do something.
> 
> Thinking about it again, I thought of what might be a common case: a simple bulletin board. Something with no code; state info only; write anything you want; no code for conflict resolution; last writer wins. A `read-write` property fits the bill.
> 
> I don't understand why take away the 1 of 4 property access methods? Why not leave it up to the programmer? Once there are variables in the interface...
> 
> I don't think it's a big deal either way. If someone really needs it, probably easy to add.

We have a few choices for the interface (as discussed in the other thread):

1. Interface supports only object methods and doesn't support object variables.  This is similar to other OOP languages.
2. Interface supports object methods and read-only object variables.
3. Interface supports object methods, read-only and read-write object variables.

I decided to go with option 2 as a compromise between option 1 and option 3.  I consider the read-only object
variable as an optimized way to get the object state (instead of using many getter functions).

If not supporting read-only object variables in the interface becomes an issue in the future, we can add it back.

errael:
> If not supporting read-only object variables in the interface becomes an issue in the future, we can add it back.

I sincerely hope you meant **read-write**.

yegappan:
> > If not supporting read-only object variables in the interface becomes an issue in the future, we can add it back.
> 
> I sincerely hope you meant **read-write**.

Yes.  That was a typo.  I meant read-write variables. Edited the response.

PR #13189, opened 2023-09-26
Fix 'scrollbind' not working when scrolling inactive window

zeertzjq:
Problem:  'scrollbind' doesn't work when scrolling inactive window.
Solution: Call do_check_scrollbind() after scrolling inactive window.

### DIFF
--- a/src/mouse.c
+++ b/src/mouse.c
@@ -1171,6 +1171,9 @@ do_mousescroll(cmdarg_T *cap)
 	    leftcol = 0;
 	do_mousescroll_horiz((long_u)leftcol);
     }
+
+    if (curwin->w_p_scb)
+	do_check_scrollbind(TRUE);
     may_trigger_win_scrolled_resized();
 }
 
--- a/src/move.c
+++ b/src/move.c
@@ -3587,7 +3587,7 @@ do_check_cursorbind(void)
     FOR_ALL_WINDOWS(curwin)
     {
 	curbuf = curwin->w_buffer;
-	// skip original window  and windows with 'noscrollbind'
+	// skip original window and windows with 'nocursorbind'
 	if (curwin != old_curwin && curwin->w_p_crb)
 	{
 # ifdef FEAT_DIFF
--- a/src/testdir/test_scrollbind.vim
+++ b/src/testdir/test_scrollbind.vim
@@ -30,6 +30,9 @@ func Test_scrollbind()
   wincmd p
   setl noscrollbind
   call assert_equal(0, topLineLeft - topLineRight)
+
+  bwipe!
+  bwipe!
 endfunc
 
 " Test for 'scrollbind'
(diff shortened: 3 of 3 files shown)


vim-ml:
On 2023-09-25, zeertzjq wrote:
> Problem: 'scrollbind' doesn't work when scrolling inactive window.
> Solution: Call do_check_scrollbind() after scrolling inactive window.

I thought it was supposed to work that way.  It _is_ handy to be
able to scroll the two windows independently by just clicking and
moving the mouse.

Regards,
Gary

errael:
> I thought it was supposed to work that way. It _is_ handy to be able to scroll the two windows independently by just clicking and moving the mouse.

I actually was using that feature today. I'd never know explicitly about it, and was surprised, but it was _very handy_.

zeertzjq:
Oh right, I just noticed this:
```
							*scrollbind-quickadj*
The 'scrollbind' flag is meaningful when using keyboard commands to vertically
scroll a window, and also meaningful when using the vertical scrollbar of the
window which has the cursor focus.  However, when using the vertical scrollbar
of a window which doesn't have the cursor focus, 'scrollbind' is ignored.
This allows quick adjustment of the relative offset of 'scrollbind' windows.
```

vim-ml:
On 2023-09-25, errael wrote:
>     I thought it was supposed to work that way. It is handy to be able to
>     scroll the two windows independently by just clicking and moving the mouse.
> 
> I actually was using that feature today. I'd never know explicitly about it,
> and was surprised, but it was very handy.

I did a little looking.  The current behavior is described in ":help
scroll-binding".

    When using the scrollbars, the binding only happens when
    scrolling the window with focus (where the cursor is).  You can
    use this to avoid scroll-binding for a moment without resetting
    options.

Therefore, this proposed "fix" should not be accepted.

Regards,
Gary

Shane-XB-Qian:
looks generally it worked in vimdiff mode which natively enabled this option for it.
// BTW: somehow sometime i faced that maybe one window off (not scrolling with another), 
// i cannot and didnot dig to repro solidly, but if someone faced same situation, may look into further.

zeertzjq:
Actually this seems achievable using `WinScrolled` autocommand and `:windo`. An autocommand for diff mode:
```vim
func ScrollBind()
  let oldwin = win_getid()
  for [win, delta] in items(v:event)
    let id = str2nr(win)
    if id == 0
      continue
    endif
    if delta.topline == 0 && delta.topfill == 0
      continue
    endif
    let nr = win_id2win(id)
    if getwinvar(nr, '&scrollbind') && getwinvar(nr, '&diff')
      exe nr .. 'windo :'
      break
    endif
  endfor
  call win_gotoid(oldwin)
endfunc

autocmd WinScrolled * call ScrollBind()
```

It's mainly useful in diff mode, where `diff_set_topline()` is used for 'scrollbind'.

Shane-XB-Qian:
i was saying this 'scrollbind' was mainly used and default enabled in vimdiff,
but sometime it didnot sync scrolling in between (those windows), 
i cannot solidly repro what happened, so deleted comment.

but the problem seems was the status of fold, if it was open/close, and there one there or not, and related lines range....
so it should be not just about `delta`.

tonymec:
IIUC, this is intentional (and documented):
```
			*'scrollbind'* *'scb'* *'noscrollbind'* *'noscb'*
'scrollbind' 'scb'	boolean  (default off)
			local to window
	See also |scroll-binding|.  When this option is set, scrolling the
	current window also scrolls other scrollbind windows (windows that
	also have this option set).  This option is useful for viewing the
	differences between two versions of a file, see 'diff'.
```
* Scrolling the active window scrolls all scrollbound windows to keep them synchronized.
* Scrolling an inactive window scrolls only that window, to resynchronize it in case it had become unsynchronized for some reason.

Best regards,
Tony.

PR #13193, opened 2023-09-26
Update rmd runtime files (ftplugin and syntax)

jalvesaq:
ftplugin/rmd.vim:

  - Set 'commentstring' dynamically according to code region.

syntax/rmd.vim:

  - Include syntax highlighting of fenced languages dynamically.
  - Add conceal char for line break.


jalvesaq:
The proposed changes are in https://github.com/jalvesaq/R-Vim-runtime since May 2023.

chrisbra:
thank you!

PR #13205, opened 2023-09-28
win32,macOS: Disable Test_glob_extended_bash

k-takata:
This test doesn't work on Windows even if bash can be executed.


k-takata:
It seems that it also fails on macOS:
https://github.com/vim/vim/actions/runs/6333372535/job/17201386137?pr=13205
I'm not sure if this is expected.

benknoble:
For macOS I expect you need a newer bash. The default is old. See my comments on the original PR.

PR #13207, opened 2023-09-28
Reserve 100 error codes for future enhancements to the Vim9 class support

yegappan:


chrisbra:
can you add a comment to `error.h` that error `E15XX` is reserved for vim9 class support?

yegappan:
> can you add a comment to `error.h` that error `E15XX` is reserved for vim9 class support?

Yes. The error codes E1390 to E1499 are reserved for the Vim9 class support.

chrisbra:
ah I misread. Yeah that is fine.

chrisbra:
this had some conflicts. I hope I fixed them all correctly. Thanks!

PR #13208, opened 2023-09-28
fix: bash ver check

Shane-XB-Qian:
// i have no macos and windows machines, but try this one, perhaps help the test work through.

### DIFF
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -6711,7 +6711,7 @@ mch_expand_wildcards(
 		// vimglob() function to define for Posix shell
     static char *sh_vimglob_func = "vimglob() { while [ $# -ge 1 ]; do echo \"$1\"; shift; done }; vimglob >";
 		// vimglob() function with globstar setting enabled, only for bash >= 4.X
-    static char *sh_globstar_opt = "[[ ${BASH_VERSINFO[0]} -ge 4 ]] && shopt -s globstar; ";
+    static char *sh_globstar_opt = "[ ${BASH_VERSINFO:-0} -ge 4 ] && shopt -s globstar; ";
 
     *num_file = 0;	// default: no files found
     *file = NULL;


chrisbra:
Well, 
 - it's an array, so you need to use `[0]` to check the major version number
 - I would keep the `[[` since it's clear this is using bash here anyhow.

but happy to keep the *default value*  substitution

Shane-XB-Qian:
> * the major version number

seems it default was major, 
and to be more compatible and it is scripting actually.
// e.g against to dash, and no req strict type checking here.

chrisbra:
> and to be more compatible and it is scripting actually.

But it is not. It only goes into that part, if the shell ends with `bash`. So you don't need to worry about bashism

Shane-XB-Qian:
> goes into that part, if the shell ends with `bash`. So you don't need to worry about bashism

ok, but this made it work even it goes wrong, or i cannot image other various cases, maybe there.

-- 
shane.xb.qian

chrisbra:
don't understand. What exactly do you think can go wrong?

Shane-XB-Qian:
```c
    if (shell_style == STYLE_ECHO)
    {
       if (strstr((char *)gettail(p_sh), "bash") != NULL)
	    shell_style = STYLE_GLOBSTAR;
```

currently it is relying on p_sh (which is 'shell') set correctly and shell_style check correctly, or someone did not link /bin/sh to something else, this pr make it at least not err at checking bash ver.
// maybe a bit extreme, but tiny change seems better.

chrisbra:
yes exactly, we only run that part of the code, if the users shell option is set to `bash`. So I am still not sure what problem you are trying to fix. 

I would be okay with this patch however:

```patch
diff --git a/src/os_unix.c b/src/os_unix.c
index c5a54e419..cebcdc1dc 100644
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -6711,7 +6711,7 @@ mch_expand_wildcards(
                // vimglob() function to define for Posix shell
     static char *sh_vimglob_func = "vimglob() { while [ $# -ge 1 ]; do echo \"$1\"; shift; done }; vimglob >";
                // vimglob() function with globstar setting enabled, only for bash >= 4.X
-    static char *sh_globstar_opt = "[[ ${BASH_VERSINFO[0]} -ge 4 ]] && shopt -s globstar; ";
+    static char *sh_globstar_opt = "[[ ${BASH_VERSINFO[0]:-0} -ge 4 ]] && shopt -s globstar; ";

     *num_file = 0;     // default: no files found
     *file = NULL;
```

I'd like to keep the `[[`  and the `[0]`, because this part will only be run by a bash shell anyhow and it is clear from the code we are trying to access an Bash array variable

Shane-XB-Qian:
> not sure what problem you are trying to fix.

trying to kill any chance (wish) it was not bash or not right bash.

chrisbra:
yes, but that shouldn't happen

Shane-XB-Qian:
um, i wish so, but just to make sure even it happened would not impact this bash ver check.
// i know it is a bit extreme. so if you felt not like or not worthy to merge, i am ok to pull back.

chrisbra:
>  i know it is a bit extreme. so if you felt not like or not worthy to merge, i am ok to pull back.

please make the adjustments I have mentioned above a few times

Shane-XB-Qian:
> > i know it is a bit extreme. so if you felt not like or not worthy to merge, i am ok to pull back.
> 
> please make the adjustments I have mentioned above a few times

that adjustment i think should be not required if your assumption it really was and right ver was bash.

PR #13209, opened 2023-09-28
Add check for duplicate object and class variable names and for using reserved names

### LINKED ISSUE #13204: [vim9script] duplicate name not always detected
### Steps to reproduce

The following script has `static sval` and `this.sval`. 
- The duplicate name is **not** detected when **processing the class**.
- The duplicate name is **not** detected when **running the static method**.
- The duplicate is finally detected when `C.new()` is executed.
- The duplicate is **not** detected when `C.new()` is manually defined and executed.

Source this script and notice that `C.F2()` works ok. The duplicate name is not detected.
```
vim9script

class C
    public static sval: list<number>
    public this.sval: list<number>
    def F1()
        echo this.sval
    enddef
    static def F2()
        echo sval
    enddef
endclass

C.F2()              # no error
var o = C.new()     # error
```

This gets an error
```
vim9script

class C
    public static sval: list<number>
    public this.sval: list<number>
endclass

var o = C.new()   # error
```
This gets no error
```
vim9script

class C
    public static sval: list<number>
    public this.sval: list<number>
    def new()
    enddef
endclass

var o = C.new()     # no error
```

Notice that in this example the duplicate method name is detected during processing the class
```
vim9script

class C
    def F1()
    enddef
    def F1()
    enddef
endclass
```

### Expected behaviour

Consistent detection of duplicate names in a class.

Consistent error messages would be nice. 

1. `Duplicate`
2. `Argument already declared`

The 2nd is confusing as well. `Argument`? (I get the automatically defined constructor)

### Version of Vim

9.0.1946

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #13204

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -4139,7 +4139,7 @@ def Test_dup_member_variable()
     assert_equal(10, C.val)
     assert_equal(20, c.val)
   END
-  v9.CheckSourceSuccess(lines)
+  v9.CheckSourceFailure(lines, 'E1369: Duplicate variable: val', 4)
 
   # Duplicate object member variable in a derived class
   lines =<< trim END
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks! Should we document the reserved names?

PR #13211, opened 2023-09-28
Apply termcap entries when 'keyprotocol' changes

gpanders:
When the 'keyprotocol' option was set after startup (including in a
user's .vimrc) the termcap entries associated with the matching protocol
were not applied. Thus, setting the option has no effect.

When 'keyprotocol' is changed it should also update the termcap entries.

### DIFF
--- a/src/optionstr.c
+++ b/src/optionstr.c
@@ -1945,8 +1945,12 @@ did_set_keymodel(optset_T *args UNUSED)
     char *
 did_set_keyprotocol(optset_T *args UNUSED)
 {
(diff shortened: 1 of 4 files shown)


gpanders:
Test failures look unrelated (looks like they started in 9eb1ce531527a7177d16373b0f8689bbcd3d5f73).

chrisbra:
can we have a test for this please?

gpanders:
@chrisbra Added in 9593e4efc2ca85f8c5a23418a1b2e06487e480c1.

PR #13212, opened 2023-09-28
Fix: Object variable names can be used as formal arguments of methods other than constructors

### LINKED ISSUE #13152: Initializing formal parameters are available to all methods
### Steps to reproduce

Is it intended that this works?

```vim
vim9script

class Foo
  this.foo: number
  def SetFoo(this.foo)
  enddef
  def GetFoo(): number
    return this.foo
  enddef
endclass

var f = Foo.new()
f.SetFoo(42)
echomsg f.GetFoo()
```

### Expected behaviour

I don't feel strongly about it one way or the other but allowing this feature is normally (from a brief survey of other languages) only available in constructors.

### Version of Vim

9.0.1927

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

h-east:
Close: #13152

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3529,6 +3529,8 @@ EXTERN char e_public_keyword_not_supported_for_method[]
 	INIT(= N_("E1388: Public keyword not supported for a method"));
 EXTERN char e_missing_name_after_implements[]
 	INIT(= N_("E1389: Missing name after implements"));
+EXTERN char e_cannot_use_a_object_variable_except_with_the_new_method_str[]
+	INIT(= N_("E1390: Cannot use a object variable \"this.%s\" except with the \"new\" method"));
 #endif
 EXTERN char e_cannot_mix_positional_and_non_positional_str[]
 	INIT(= N_("E1400: Cannot mix positional and non-positional arguments: %s"));
@@ -3544,4 +3546,4 @@ EXTERN char e_invalid_format_specifier_str[]
 	INIT(= N_("E1405: Invalid format specifier: %s"));
 EXTERN char e_aptypes_is_null_nr_str[]
 	INIT(= "E1408: Internal error: ap_types or ap_types[idx] is NULL: %d: %s");
-// E1390 - E1399 unused
+// E1391 - E1399 unused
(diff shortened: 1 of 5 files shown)


src/userfunc.c
```diff
@@ -320,6 +320,16 @@ get_function_args(
 		++p;
 	    char_u *argend = p;
 
+	    // object variable can use constructor
```
yegappan: Can you update this comment to "object variable this.<name> can be used only in a constructor"?

src/testdir/test_vim9_class.vim
```diff
@@ -955,6 +955,28 @@ def Test_class_new_with_object_member()
   END
   v9.CheckSourceFailure(lines, 'E1013:')
 
+  lines =<< trim END
+      vim9script
```
yegappan: The PR https://github.com/vim/vim/pull/13206 will adjust the indentation to use two spaces in a heredoc.  Can you use two spaces for indentation in the heredoc?

src/testdir/test_vim9_class.vim
```diff
@@ -955,6 +955,28 @@ def Test_class_new_with_object_member()
...
+      def Check()
+        try
+          var c = C.new()
+          c.MethodA('abc')
+        catch
+          assert_report($'Unexpected exception was caught: {v:exception}')
+        endtry
+      enddef
+
+      Check()
```
yegappan: As the compilation of the class will fail, will it compile the Check() function and invoke it?  Is this function needed in this test?

src/errors.h
```diff
@@ -3529,6 +3529,8 @@ EXTERN char e_public_keyword_not_supported_for_method[]
 	INIT(= N_("E1388: Public keyword not supported for a method"));
 EXTERN char e_missing_name_after_implements[]
 	INIT(= N_("E1389: Missing name after implements"));
+EXTERN char e_cannot_use_a_object_variable_except_with_the_new_method_str[]
+	INIT(= N_("E1390: Cannot use a object variable \"this.%s\" except with the \"new\" method"));
```
yegappan: Can you change "use a object" to "use an object"?

src/testdir/test_vim9_class.vim
```diff
@@ -955,6 +955,28 @@ def Test_class_new_with_object_member()
...
+  END
+  v9.CheckSourceFailure(lines, 'E1390:')
```
yegappan: I started using complete error messages for verification as this makes the test easier to read and understand.  Can you include the complete the error message here?

yegappan (changes requested):

h-east:
@yegappan Thanks for the review!
I have accepted all suggestions and made the corrections.

yegappan (approved):

PR #13213, opened 2023-09-28
runtime(sh): Rename :Help user command

dkearns:
Support :Help and 'keywordprg' for shells other than Bash.

I'm not sure parameterising this makes it any more readable so it's copy-paste for now.

Follow up to #13171.

@Konfekt does this still meet your original intentions?

### DIFF
--- a/runtime/ftplugin/sh.vim
+++ b/runtime/ftplugin/sh.vim
@@ -3,7 +3,8 @@
 " Maintainer:		Doug Kearns <dougkearns@gmail.com>
 " Previous Maintainer:	Dan Sharp
 " Contributor:		Enno Nagel <ennonagel+vim@gmail.com>
-" Last Change:		2023 Aug 29
+"			Eisuke Kawashima
+" Last Change:		2023 Sep 28
 
 if exists("b:did_ftplugin")
   finish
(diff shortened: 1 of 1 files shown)


Konfekt:
I am a bit lost turning 

```vim
if (exists("b:is_bash") && (b:is_bash == 1))
```

into

```vim
if get(b:, "is_bash", 0)
```

as they look opposite to each other, as if the new else block does what the former if block did.

As `man` is the standard `keywordprg` on Unix, there is no need to set it in this case.
So I am not sure what is achieved other than inverting the (formerly correct ?!) logic

dkearns:
It's essentially the same conditional test. It means use the value of `b:is_bash` if it exists, otherwise use 0 (false).

The old approach to boolean config variables was to _only_ test for existence.  The value was irrelevant, even a false value.  There's been a half-hearted attempt to move to proper boolean values, it's debatable if we should test for more than existence here but as we already were I carried on. See `:help ft-c-syntax` for some discussion.

> As man is the standard keywordprg on Unix, there is no need to set it in this case.

I don't particularly want different UI experiences for the different shells.  The default `K` behaviour in the GUI uses the pseudo TTY, for example.  It looks like you made some effort to improve on that?

Are you suggesting the `:Help` command exists only to support a user-defined `keywordprg`?  Only supporting that for Bash seems odd.

It's suprising that `keywordprg` doesn't accept a function.

Konfekt:
> Are you suggesting the :Help command exists only to support a user-defined keywordprg? 

Yes, it exists mainly to call `'bash -c "help <args>" 2>/dev/null || man "<args>"'` as it seemed difficult to pass it to `keywordprg` directly. In case `less` exists, then I found it a convenient pager to look through the output.

chrisbra:
> It's suprising that keywordprg doesn't accept a function.

You mean a vim function? That would require a new option 'keywordexpr' or not?

dkearns:
> > It's suprising that keywordprg doesn't accept a function.
> 
> You mean a vim function? That would require a new option 'keywordexpr' or not?

I guess it would require a new option for consistency with similar options.  I have certainly seen ftplugins remapping `K` in the wild.

It might be useful if user commands could be declared script-local for this sort of scenario.

chrisbra:
thanks!

Issue #13082, opened 2023-09-12
Netrw can cause a stacktrace when executing bd! during shutdown


bstaletic:
### Steps to reproduce

1. `vim -Nu NONE --cmd 'source /usr/share/vim/vim90/plugin/netrwPlugin.vim' -c 'set hidden' $HOME $HOME/.profile`
2. Close the netrw buffer, so that only the file remains.
3. Close the file buffer.
4. Observe that, briefly, a stack trace is printed during shut down.

### Expected behaviour

Vim should exit cleanly.

Note that this only ever happened to me during shutdown, so the stack trace does not actually impede editing of files.

### Version of Vim

9.0.1777

### Environment

OS: Gentoo linux
Terminal: st with tmux
`$TERM`: `xterm-256color`
Shell: zsh

### Logs and stack traces

Stack trace:

```shell
Error detected while processing VimLeave Autocommands for "*"..function <SNR>2_NetrwBookHistSave[12]..<SNR>2_NetrwEnew[7]..<SNR>2_NetrwBufRemover:
line    9:
E516: No buffers were deleted: bd! 3
```

State of buffers before closing the last one:
```
:ls
  1 #h   "~"                            line 1
  2 %a   ".profile"                     line 1
```
```

**Comment by bstaletic** [comment]
Easier way to repro:

1. `vim -Nu NONE --cmd 'source /usr/share/vim/vim90/plugin/netrwPlugin.vim' -c 'set hidden' $HOME $HOME/.profile`
2. `:q` - quit the netrw buffer.
3. `:doautoall VimLeave` - simulate the VimLeave event to trigger the traceback.

**Comment by chrisbra** [comment]
Hm, this seems to come from here:

https://github.com/vim/vim/blob/6ffcc58be32aa1b337bc839cfe173b68cfde7085/runtime/autoload/netrw.vim#L12040-L12050

I guess this should be a:

```
exe "sil! bd! ".a:bufid
```

to ignore the error in case a buffer cannot be unloaded? ping @cecamp

bstaletic:
Just wanted to confirm that the patch actually avoids the stacktrace. Thanks!

PR #13215, opened 2023-09-29
NetRw: Fix not to `bd` unloaded buf

KSR-Yasuda:
NetRW sometimes comes here and gets an error (Perhaps in `scp://` dir access?).


chrisbra:
does this also fix #13082 ?

chrisbra:
ping?

KSR-Yasuda:
> does this also fix #13082 ?

Oops, should I check it?

I suppose so.
As my check, `bd!` call is avoided by `bufloaded(a:bufid)` being false.

To make sure, adding `silent!` to `bd!` may be better, though.

PR #13218, opened 2023-09-29
Fix misplacement of unused comment

k-takata:


chrisbra:
I actually placed it intentionally at the bottom, because I thought that's were it is most easily seen. But okay, thanks!

PR #13220, opened 2023-09-29
Remove an unused static field and parameter

yegappan:


errael:
Since statics can no longer be accessed through an object, I was going to offer to clean this up after 9.1 is released. Nice to have things magically disappear from my todo list.

chrisbra:
you are welcome :)

PR #13221, opened 2023-09-29
Support contra-variant type check for object method arguments (similar to Dart)

### LINKED ISSUE #12965: Methods are covariant in their parameter and return types
### Steps to reproduce

```vim
vim9script
    
class A
endclass

class B extends A
endclass

class C extends B
endclass
  
class Foo
  def Doit(p: B): B
    return B.new()
  enddef
endclass

class Bar extends Foo
  def Doit(p: C): C
    return C.new()
  enddef
endclass
```

### Expected behaviour

This should not compile.  There should be type errors for both the parameter and return type as the help says:

> Object methods of the base class can be overruled.  The signature (arguments,
> argument types and return type) must be exactly the same. 

However, ideally the current implementation should just be changed to make the methods contravariant in their parameter types.  The existing covariance of return types is correct.

As there's no complications with regard to overloaded methods,which are not supported, is there a reason not to relax the restriction as stated in the help?

So, the following would compile - contravariant parameter types and covariant return types.

```vim
class Baz extends Foo
  def Doit(p: A): C
    return C.new()
  enddef
endclass
```

Irrespective, covariant parameter types should obviously not be allowed.

I don't have time to experiment further at the moment but will attempt to in the near future.

### Version of Vim

9.0.182

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Fixes #12965

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -530,6 +530,10 @@ If the type of a variable is not explicitly specified in a class, then it is
 set to "any" during class definition.  When an object is instantiated from the
 class, then the type of the variable is set.
 
+The following reserved keyword names cannot be used as an object or class
+variable name: "super", "this", "true", "false", "null", "null_blob",
+"null_dict", "null_function", "null_list", "null_partial", "null_string",
+"null_channel" and "null_job".
 
 Extending a class ~
 							*extends*
@@ -543,9 +547,11 @@ Object variables from the base class are all taken over by the child class.  It
 is not possible to override them (unlike some other languages).
 
 						*E1356* *E1357* *E1358*
-Object methods of the base class can be overruled.  The signature (arguments,
-argument types and return type) must be exactly the same.  The method of the
-base class can be called by prefixing "super.".
+Object methods of the base class can be overruled.  The number of arguments
+must be exactly the same.  The method argument type can be a contra-variant
+type of the base class method argument type.  The method return value type can
+be a covariant type of the base class method return value type.  The method of
+the base class can be called by prefixing "super.".
 
 						*E1377*
 The access level of a method (public or private) in a child class should be
--- a/src/proto/vim9class.pro
+++ b/src/proto/vim9class.pro
@@ -29,5 +29,5 @@ int object_free_nonref(int copyID);
 void method_not_found_msg(class_T *cl, vartype_T v_type, char_u *name, size_t len);
 void member_not_found_msg(class_T *cl, vartype_T v_type, char_u *name, size_t len);
 void f_instanceof(typval_T *argvars, typval_T *rettv);
-int class_instance_of(class_T *cl, class_T *other_cl);
+int class_instance_of(class_T *cl, class_T *other_cl, int covariance_check);
 /* vim: set ft=c : */
--- a/src/structs.h
+++ b/src/structs.h
@@ -4798,14 +4798,19 @@ typedef enum {
     WT_ARGUMENT,
     WT_VARIABLE,
     WT_MEMBER,
-    WT_METHOD,
+    WT_METHOD,		// object method
+    WT_METHOD_ARG,	// object method argument type
+    WT_METHOD_RETURN	// object method return type
 } wherekind_T;
 
-// Struct used to pass to error messages about where the error happened.
+// Struct used to pass the location of a type check.  Used in error messages to
+// indicate where the error happened.  Also used for doing covariance type
+// check for object method return type and contra-variance type check for
+// object method arguments.
 typedef struct {
     char	*wt_func_name;  // function name or NULL
     char	wt_index;	// argument or variable index, 0 means unknown
-    wherekind_T	wt_kind;	// "variable" when TRUE, "argument" otherwise
+    wherekind_T	wt_kind;	// type check location
 } where_T;
 
 #define WHERE_INIT {NULL, 0, WT_UNKNOWN}
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -6318,4 +6318,80 @@ def Test_reserved_varname()
   endfor
 enddef
 
+" Test for checking the type of the arguments and the return value of a object
+" method in an extended class.
+def Test_extended_obj_method_type_check()
+  var lines =<< trim END
+    vim9script
+
+    class A
+    endclass
+    class B extends A
+    endclass
+    class C extends B
+    endclass
+
+    class Foo
+      def Doit(p: B): B
+        return B.new()
+      enddef
+    endclass
+
+    class Bar extends Foo
+      def Doit(p: A): C
+        return C.new()
+      enddef
+    endclass
+  END
+  v9.CheckSourceSuccess(lines)
+
+  lines =<< trim END
+    vim9script
+
+    class A
+    endclass
+    class B extends A
+    endclass
+    class C extends B
+    endclass
+
+    class Foo
+      def Doit(p: B): B
+        return B.new()
+      enddef
+    endclass
+
+    class Bar extends Foo
+      def Doit(p: C): B
+        return B.new()
+      enddef
+    endclass
+  END
+  v9.CheckSourceFailure(lines, 'E1383: Method "Doit": type mismatch, expected func(object<B>): object<B> but got func(object<C>): object<B>', 20)
+
+  lines =<< trim END
+    vim9script
+
+    class A
+    endclass
+    class B extends A
+    endclass
+    class C extends B
+    endclass
+
+    class Foo
+      def Doit(p: B): B
+        return B.new()
+      enddef
+    endclass
+
+    class Bar extends Foo
+      def Doit(p: B): A
+        return A.new()
+      enddef
+    endclass
+  END
+  v9.CheckSourceFailure(lines, 'E1383: Method "Doit": type mismatch, expected func(object<B>): object<B> but got func(object<B>): object<A>', 20)
+enddef
+
 " vim: ts=8 sw=2 sts=2 expandtab tw=80 fdm=marker
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -2561,7 +2561,7 @@ inside_class(cctx_T *cctx_arg, class_T *cl)
 {
     for (cctx_T *cctx = cctx_arg; cctx != NULL; cctx = cctx->ctx_outer)
 	if (cctx->ctx_ufunc != NULL
-			&& class_instance_of(cctx->ctx_ufunc->uf_class, cl))
+			&& class_instance_of(cctx->ctx_ufunc->uf_class, cl, TRUE))
 	    return TRUE;
     return FALSE;
 }
@@ -2871,28 +2871,38 @@ member_not_found_msg(class_T *cl, vartype_T v_type, char_u *name, size_t len)
  * interfaces matches the class "other_cl".
  */
     int
-class_instance_of(class_T *cl, class_T *other_cl)
+class_instance_of(class_T *cl, class_T *other_cl, int covariance_check)
 {
     if (cl == other_cl)
 	return TRUE;
 
-    // Recursively check the base classes.
-    for (; cl != NULL; cl = cl->class_extends)
+    if (covariance_check)
     {
-	if (cl == other_cl)
-	    return TRUE;
-	// Check the implemented interfaces and the super interfaces
-	for (int i = cl->class_interface_count - 1; i >= 0; --i)
+	// Recursively check the base classes.
+	for (; cl != NULL; cl = cl->class_extends)
 	{
-	    class_T	*intf = cl->class_interfaces_cl[i];
-	    while (intf != NULL)
+	    if (cl == other_cl)
+		return TRUE;
+	    // Check the implemented interfaces and the super interfaces
+	    for (int i = cl->class_interface_count - 1; i >= 0; --i)
 	    {
-		if (intf == other_cl)
-		    return TRUE;
-		// check the super interfaces
-		intf = intf->class_extends;
+		class_T	*intf = cl->class_interfaces_cl[i];
+		while (intf != NULL)
+		{
+		    if (intf == other_cl)
+			return TRUE;
+		    // check the super interfaces
+		    intf = intf->class_extends;
+		}
 	    }
 	}
+    }
+    else
+    {
+	// contra-variance
+	for (; other_cl != NULL; other_cl = other_cl->class_extends)
+	    if (cl == other_cl)
+		return TRUE;
     }
 
     return FALSE;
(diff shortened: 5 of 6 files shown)


chrisbra:
what is contr-variant?

RestorerZ:
I was curious about that too, as I didn't understand what it was about at all.
Here's what wikipedia says:
[Covariance and contravariance (computer science)](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science))
And this is what I was able to find on Dart
[Dart and covariant](https://richardjharris.github.io/dart-and-covariant/)
But if you,  @yegappan , could explain it in your own words, it would be much better.

vim-ml:
Hi Christian,

On Fri, Sep 29, 2023 at 10:30 AM Christian Brabandt <
***@***.***> wrote:

> what is contr-variant?
>
>
> Consider the following example (from #12965):

class A
endclass

class B extends A
endclass

class C extends B
endclass

In this example, type "C" is a covariant of type "B" and "A" is
contra-variant of type "B".

This matters when overruling methods in an extended class.  The method
return value
type in an extended class can use a covariant type of return value in the
parent class.
The type of method arguments in an extended class can use a contra-variant
type.

For example (continuing the above example):

class Foo
  def Doit(p: B): B
    return B.new()
  enddef
endclass

class Baz extends Foo
  def Doit(p: A): C
    return C.new()
  enddef
endclass

The Doit() method in class Foo uses "B" as the argument type and "B" as the
return type.
The Doit() method in class Baz uses "A" as the argument type and "C" as the
return type.
"A" is a contra-variant of "B" and "C" is a co-variant of "B".

A detailed explanation about this is in
https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
.

I wanted to follow the model used in the Dart language (
https://medium.com/dartlang/dart-declaration-site-variance-5c0e9c5f18a5).

Regards,
Yegappan

vim-ml:
On Fri, Sep 29, 2023 at 12:50 PM vim-dev ML ***@***.***>
wrote:

> Hi Christian,
>
> On Fri, Sep 29, 2023 at 10:30 AM Christian Brabandt <
> ***@***.***> wrote:
>
> > what is contr-variant?
> >
> >
> > Consider the following example (from #12965):
>
> class A
> endclass
>
> class B extends A
> endclass
>
> class C extends B
> endclass
>
> In this example, type "C" is a covariant of type "B" and "A" is
> contra-variant of type "B".
>
> This matters when overruling methods in an extended class. The method
> return value
> type in an extended class can use a covariant type of return value in the
> parent class.
> The type of method arguments in an extended class can use a contra-variant
> type.
>
> For example (continuing the above example):
>
> class Foo
> def Doit(p: B): B
> return B.new()
> enddef
> endclass
>
> class Baz extends Foo
> def Doit(p: A): C
> return C.new()
> enddef
> endclass
>
> The Doit() method in class Foo uses "B" as the argument type and "B" as
> the
> return type.
> The Doit() method in class Baz uses "A" as the argument type and "C" as
> the
> return type.
> "A" is a contra-variant of "B" and "C" is a co-variant of "B".
>
> A detailed explanation about this is in
>
> https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
> ..
>
> I wanted to follow the model used in the Dart language (
> https://medium.com/dartlang/dart-declaration-site-variance-5c0e9c5f18a5).
>
>
>
Currently the type of the method arguments and return value can be a
covariant type
of the ones in the parent class method.

- Yegappan

errael:
Shades of `Java`'s `PECS`.

There were two ways to deal with #12965.

1. Adjust the parser to meet the spec. 
2. Change the spec and various implementation things.

Doing 1. is backwards compatible with future implementation of the requested features.
Doing 2. seems rushed. 

My first look, not running any code, this change seems to add little, at the expense of _conceptual complexity_. I was in the process of manually working through some things related to this, so I could make a comment, when it got checked in (seemed pretty quick for a change that wasn't understood). 

What's to rush to add new features that have not been debated?

vim-ml:
Hi,

On Fri, Sep 29, 2023 at 2:51 PM errael ***@***.***> wrote:

> Shades of Java's PECS.
>
> There were two ways to deal with #12965
> <https://github.com/vim/vim/issues/12965>.
>
>    1. Adjust the parser to meet the spec.
>
>
The spec currently states that:

Object methods of the base class can be overruled. The signature (arguments,
argument types and return type) must be exactly the same.

The current implementation doesn't match with this spec.   If we follow
this spec,
then the type in an overruled method cannot be a covariant or
contra-variant.
I think that will be limiting, as the method in an extended class may want
to return
a covariant type.


>
>    1. Change the spec and various implementation things.
>
> Doing 1. is backwards compatible with future implementation of the
> requested features.
> Doing 2. seems rushed.
>
> My first look, not running any code, this change seems to add little, at
> the expense of *conceptual complexity*. I was in the process of manually
> working through some things related to this, so I could make a comment,
> when it got checked in (seemed pretty quick for a change that wasn't
> understood).
>

Let me know if you see any cases that will break with the latest changs.

> What's to rush to add new features that have not been debated?
>
>
>
No rush.  I didn't see any objection to the proposal in
https://github.com/vim/vim/issues/12965
and this is closer to the Dart spec.

As the changes are minimal to support this, if needed, we can make
adjustments.

Regards,
Yegappan

errael:
> >  Shades of Java's PECS.
> > 
> >  There were two ways to deal with #12965
> > 
> >     1. Adjust the parser to meet the spec.

> The spec currently states that:
> 
> Object methods of the base class can be overruled. The signature (arguments,
> argument types and return type) must be exactly the same.
> 
> The current implementation doesn't match with this spec.

Right, I mentioned that and one option was to meet the spec. With this change,
even with the rewording of the spec, it can be argued (but not definitively)
that the **changed** spec is violated. Note the original spec says

```
Object methods of the base class can be overruled.  The signature (arguments,
argument types and return type) must be exactly the same.  The method of the
base class can be called by prefixing "super.".
```
The new spec still has the `prefixing "super."` phrase.

When I read `base class can be called by prefixing "super."`, in the context
of `The signature ... must be exactly the same`, it`s easy to interpret this
as meaning, using the current example, that

```
class Bar extends Foo
  def Doit(p: B): B
    return super.Doit(p)
  enddef
endclass
```
will compile and run OK.  But if covariant and/or contravariant types are used
you'll easily get compile or runtime error, if `super.` is used in this fashion.

> If we follow this spec,
> then the type in an overruled method cannot be a covariant or
> contra-variant.
> I think that will be limiting, as the method in an extended class may want
> to return
> a covariant type.

I strongly agree that a method may want to return a covariant type.
Nothing in the previous spec (before this change) prevents returning a
covariant type. In fact, in the following example notice the output

```
object<C>
object<C>
```
Both `Foo.Doit` and `Bar.Doit` return the covariant type `C`.
```
vim9script
    
class A
endclass

class B extends A
endclass

class C extends B
endclass
  
class Foo
  def Doit(p: B): B
    return C.new()
  enddef
endclass

class Bar extends Foo
  def Doit(p: B): B
    return C.new()
  enddef
endclass

echo typename(Foo.new().Doit(C.new()))
echo typename(Bar.new().Doit(C.new()))
```

> >     2. Change the spec and various implementation things.
> > 
> >  Doing 1. is backwards compatible with future implementation of the
> >  requested features.
> >  Doing 2. seems rushed.
> > 
> >  My first look, not running any code, this change seems to add little, at
> >  the expense of *conceptual complexity*. I was in the process of manually
> >  working through some things related to this, so I could make a comment,
> >  when it got checked in (seemed pretty quick for a change that wasn't
> >  understood).

> Let me know if you see any cases that will break with the latest changs.

> >  What's to rush to add new features that have not been debated?

> No rush.  I didn't see any objection to the proposal in #12965

And there was no support either; it was not debated. My excuse, not a great one,
is that I was focused on getting 9.1 release to meet spec and uncover any
things that would violate backwards compatibility.

> and this is closer to the Dart spec.

So what? The original intent was for simplicity, AFAICT. There was a recent 
case, accessing statics, where `Dart` is simple and was chosen.

In this case, `Dart` is more complex and doesn't give an advantage, IMHO.
You can already return covariant types with out this change. And adding
contravariant parameter types is confusing and I don't see what it gets you.
I don't see any advantage to making this spec change now, and some disadvantage.
If there's a need, it can be added post 9.1, but not the other way around.

> As the changes are minimal to support this, if needed, we can make
> adjustments.

dkearns:
Is that really three variations of variance in about as many days?  I can't keep up.

This _is_ a foundational principle of OOP.  If we can't start from a common understanding of reality these threads aren't very useful.

Referencing other languages is becoming counter-productive.  I've also done this in past discussions because Bram had done so in the documentation and as a short-hand but it's becoming distracting.  Dart, in isolation, is not relevant here; it's not a pioneer of contravariance.

However, for general reference, TypeScript supports bivariance with a compiler flag for strict contravariance and Dart supports contravariance with a keyword for covariance.  The following languages, OTTOMH, all support contravariance - Python, PHP, Kotlin, Scala, Swift, Flow and others.  Java doesn't support it, presumably because it doesn't play nicely with their overloading implementation.

I'm not sure what parallel there is with the static members implementation and discussion.  The initial implementation was "weird" and it was made sane.  The functionality wasn't hobbled in some nebulous pursuit of simplicity.  This latest change makes the type system less expressive.  Given the state of play has existed for forty years I'm not sure what is going to be illuminated by deferring a decision.

Anyway, my initial report was primarily interested in having the unsound covariance resolved which it has been, thanks.

errael:
> Is that really three variations of variance in about as many days? I can't keep up.
> 
> This _is_ a foundational principle of OOP. If we can't start from a common understanding of reality these threads aren't very useful.

Yes and Yes. Hopefully we're just talking about implementation issues.

> Referencing other languages is becoming counter-productive. I've also done this in past discussions because Bram had done so in the documentation and as a short-hand but it's becoming distracting. Dart, in isolation, is not relevant here; it's not a pioneer of contravariance.
> 
> However, for general reference, TypeScript supports bivariance with a compiler flag for strict contravariance and Dart supports contravariance with a keyword for covariance. The following languages, OTTOMH, all support contravariance - Python, PHP, Kotlin, Scala, Swift, Flow and others. Java doesn't support it, presumably because it doesn't play nicely with their overloading implementation.
> 
> I'm not sure what parallel there is with the static members implementation and discussion. The initial implementation was "weird" and it was made sane. 

Exactly, the spec was never finished regarding statics, and a sane implementation was required. The point of mentioning that, is that the spec was unfinished and most agreed the implementation was broken.

This discussion, Xvariant behavior, is very different; it's about a spec change. I don't think Bram accidentally wrote, back in February
```
The signature (arguments, argument types and return type) must be exactly the same.
```
That sentence now appears twice in the spec. Implementing that in no way constrains future development, AFAICT. In fact, I think it provides maximum flexibility going forward. Less is more.

> The functionality wasn't hobbled in some nebulous pursuit of simplicity. 

I'm not sure to what this is referring. In general I agree with the implied goal.

> This latest change makes the type system less expressive. 

True. But, from what I've read, the type system isn't finished. I hope the type system is extended in a well thought out and coordinated fashion, rather than piecemeal (like Java?).

> Given the state of play has existed for forty years I'm not sure what is going to be illuminated by deferring a decision.

**Syntax**. You've pointed out that there are several ways the concept can be expressed in a language. And, at least in Java, it's not supported in method signatures at all; but the concept is intimate with Java generics (PECS). The type system needs time and effort, and I would think Xvariance needs to be part of that.

> Anyway, my initial report was primarily interested in having the unsound covariance resolved which it has been, thanks.

Agree. When I saw the initial report, I figured another spec violation to fix (and with suggestions for the future); not a new feature at this late date. Personally, I think every spec change requires discussion and burn in time; there are sometimes unintended consequences or implications.

Every time the spec is changed, the release date must be pushed out to insure reliability.

I'd like to see vim9.1 release soon, maybe that's a point of disagreement.

dkearns:
Sorry, I don't think I was notified of this reply but noticed it when checking the history of this general issue.  A day late and a dollar short...my usual style.

> This discussion, Xvariant behavior, is very different; it's about a spec change. I don't think Bram accidentally wrote, back in February
> 
> ```
> The signature (arguments, argument types and return type) must be exactly the same.
> ```
> 
> That sentence now appears twice in the spec. Implementing that in no way constrains future development, AFAICT. In fact, I think it provides maximum flexibility going forward. Less is more.

In my opinion it's a pre-alpha design/help document covering a first pass partial implementation.  We can't divine how this would otherwise have turned out and I think we should just attempt to develop the best language possible.  Bram was actually pretty flexible and, for example, reworked closures after the 9.0 release.

The first line of the "spec" states: "NOTE - This is not finished yet, anything can still change! - NOTE"

It's leveraged and ignored at will, and as has come to pass again, it's been changed with the resolution to this issue.

> > Given the state of play has existed for forty years I'm not sure what is going to be illuminated by deferring a decision.
> 
> **Syntax**. You've pointed out that there are several ways the concept can be expressed in a language. And, at least in Java, it's not supported in method signatures at all; but the concept is intimate with Java generics (PECS). The type system needs time and effort, and I would think Xvariance needs to be part of that.

I wasn't proposing _any_ syntax changes.  This is a simple type-checking.

errael:
> In my opinion [the spec is] a pre-alpha

I'm not sure I'd go that far, I could see alpha, if you think of the eventual vim9.1 release as a beta, then anything implemented in the release should be beta. Haven't heard much about release targets lately (maybe some implicit hints). But I've also hear that vim9.1 class code is supposed to be compatible with the releases after that; I don't see how that's reconciled with an alpha or beta spec (for what's implemented).

> we should just attempt to develop the best language possible

I agree. I guess the basic question/issue is whether or not we can release something that is compatible with future releases.

> > > Given the state of play has existed for forty years I'm not sure what is going to be illuminated by deferring a decision.
> > 
> > 
> > **Syntax**. You've pointed out that there are several ways the concept can be expressed in a language. And, at least in Java, it's not supported in method signatures at all; but the concept is intimate with Java generics (PECS). The type system needs time and effort, and I would think Xvariance needs to be part of that.
> 
> I wasn't proposing _any_ syntax changes. This is a simple type-checking.

What I meant by that, assuming there's evolution in the type system, is that for specifying contra-variant parameters it may be desirable to have some syntax. I think some language uses `in:`/`out:` for example (reminds me of Java's `extends`/`super`). 

Saying that parameters can be contra-variant feels like a big change that could cause problems. And it would be good to know how the type system works with future stuff, there's generics in the todo list. Unless someone can say with confidence that "there's no problems and it all work" with suggestions on how it fits in with future development... If a vim9.1 release can be delayed, I'm not sure for how long, then these questions could probably be resolved before it's released.

I'm not saying contra-variant parameters shouldn't be handled. I'm asking if they need to be handled right now? What's the downside if they are not?

PR #13174, opened 2023-09-24
[vim9script] fully implement lockvar; avoid `SEGV`; accurate error messages.

errael:
- `get_lval()` detects/returns object/class/lval_root in `lval_T`
- `compile_lock_unlock()` generate code for bare_static or obj_arg access
- `do_lock_var()` check lval for `ll_object`/`ll_class` and fail if so.

Details:
- Add `ll_object`/`ll_class`/`ll_oi` to `lval_T`.
- Add `lockunlock_T` to `isn_T` for `is_arg` to specify handling of `lval_root` in `get_lval()`.
- In `get_lval()`, fill in `ll_object`/`ll_class`/`ll_oi` as needed; when no `[idx] or .key`, check lval_root on the way out.
- In `do_lock_var()` check for `ll_object`/`ll_class`; also bullet proof ll_dict case and give `Dictionay required` if problem. (not needed to avoid lockvar crash anymore)
- In `compile_lock_unlock()` compile for the class variable and func arg cases.

This PR gives these vim9class semantics for `lockvar` in vim9.1
- `lockvar` of a `class` or `object` does nothing and no error.
  A possible future extension would allow a class to hook into `lockvar`, ass suggested for `len()`, `empty()` in todo
- handle lockvar of :def argument and other expected scenarios.
- `lockvar` of a class/object variable is an error.

PR #13222, opened 2023-09-30
Clean up TODOs from lockvar #13174.

errael:
I'll probably do this PR as a sequence of commits, addressing items individually, then squash before taking it out of draft.

- 1st commit:  - `TODO: could put lval_root info in struct`
  Did this first since it's a style issue, the change looks cleaner to my eyes, but ...
- 2nd commit: - Use `cl_exec`, the executing class, to check permissions in `get_lval()`
  turns out 1st commit is required to make this commit clean.
- 3rd commit: - Add class name to private_variable_str message.

Squashed

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3411,7 +3411,7 @@ EXTERN char e_public_must_be_followed_by_this_or_static[]
 EXTERN char e_public_variable_name_cannot_start_with_underscore_str[]
 	INIT(= N_("E1332: Public variable name cannot start with underscore: %s"));
 EXTERN char e_cannot_access_private_variable_str[]
-	INIT(= N_("E1333: Cannot access private variable: %s"));
+	INIT(= N_("E1333: Cannot access private variable \"%s\" in class \"%s\""));
 // E1334 unused
 EXTERN char e_variable_is_not_writable_str[]
 	INIT(= N_("E1335: Variable \"%s\" in class \"%s\" is not writable"));
--- a/src/eval.c
+++ b/src/eval.c
@@ -985,6 +985,52 @@ eval_foldexpr(win_T *wp, int *cp)
 }
 #endif
 
+#ifdef LOG_LOCKVAR
+typedef struct flag_string_S
+{
+    int	    flag;
+    char    *str;
+} flag_string_T;
+
+    static char *
+flags_tostring(int flags, flag_string_T *_fstring, char *buf, size_t n)
+{
+    char *p = buf;
+    *p = NUL;
+    for (flag_string_T *fstring = _fstring; fstring->flag; ++fstring)
+    {
+	if ((fstring->flag & flags) != 0)
+	{
+	    size_t len = STRLEN(fstring->str);
+	    if (n > p - buf + len + 7)
+	    {
+		STRCAT(p, fstring->str);
+		p += len;
+		STRCAT(p, " ");
+		++p;
+	    }
+	    else
+	    {
+		STRCAT(buf, "...");
+		break;
+	    }
+	}
+    }
+    return buf;
+}
+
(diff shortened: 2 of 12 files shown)


errael:
@yegappan This should be it, unless there are some unforeseen problems/issues.

get_lval correctly check access

Almost all the leftover TODO's were about getting accurate access check.
Following is my reasoning, and questions, around this change.

Careful consideration appreciated.

To do access checking in `get_lval()` during run-time, must have the
class which is executing, cl_exec. This is taken from `lval_root`.

When compiling is `cctx->ctx_ctx_ufunc->uf_class`, the class that is executing?
It seems that cl_exec is all that is needed to check permissions, agree?
This PR saves the executing class in `ISN_LOCKUNLOCK`. I'm not very familiar
with lifecycle issues in `vim9class`, so the handling is a best guess.

What other uses of `get_lval()` need to do access checking? Looking at
the code, if `cl_exec` is null then the access check stuff should only run
if it's known to be script level; otherwise the results are unpredictable.
Is there some way to tell if it's at script level? It looks like during
compilation, GLV_COMPILING, that `get_lhs()` bails early (at least in most cases).
If an access check is done and it is not a :def or script level, and
cl_exec is NULL, then it is an (internal) error; don't know how to check that.

Are there other situations in which `get_lval()` needs `cl_exec`.

TODO
- Still need to add some testing of long lval: root.obj1var.obj2var.ob31var...
- Add class name to private_variable_str message.
- When investigating usage of `get_lval()`, ran into `islocked()`. Quickly got a SEGV. Another PR.

An aside: In `get_lval_check_access()`, originally passed in is_index/is_write
but, that required calculations that may not be needed, sigh!

errael:
FYI, handling `lockvar` cleanup stuff, I discovered I really didn't understand how the access check in `get_lhs()`. Here's a trace that demonstrates what I learned.

Code is `lockvar this._val1` where `Lock()` is in `class C`
```
    def Lock()
        lockvar this._val1
    enddef
```
**The trace**

Each section of the trace is followed by a one line comment on that section.

```
: LKVAR: get_lval(): name this._val1, lval_root (nil)
: LKVAR:    ...: GLV flags NO_AUTOLOAD NO_DECL COMPILING 
```
compiling, so early exit

---

```
: LKVAR: compile_lock_unlock(): cookie 0x7ffcfa7dab80, name this._val1
: LKVAR:    ... lookup_local: name this._val1
: LKVAR:    ... INS_LOCKUNLOCK this._val1
: LKVAR:    ... buf lockvar 2 this._val1
```
compilation complete

---

```
: LKVAR: execute INS_LOCKUNLOCK isn_arg lockvar 2 this._val1
: LKVAR: get_lval(): name this._val1, lr_tv 0x556ad9f728a0 lr_is_arg 0
: LKVAR:    ...: GLV flags NO_DECL 
: LKVAR: get_lval_check_access(), cl_exec 0x556ad9c96000, cl 0x556ad9c96000, 
```
access check during execution, no problem

---

```
: LKVAR: do_lock_var(): name this._val1, is_root 0
: ERROR: Error detected while processing :source buffer=8[24]..function C.Lock:
: raw GUI output: "s2105hError detected while processing :source buffer=8[24]..function C.Lock:"
: ERROR: E1391: Cannot (un)lock variable "this._val1" in class "C"
```
but can not lock an object variable

errael:
@yegappan Squashed and ready to go.

When investigating usage of `get_lval()`, ran into previously unknown `islocked()` function. Quickly got a SEGV. Yet another follow on  PR.

yegappan (approved):

yegappan:
> @yegappan Squashed and ready to go.
> 

Looks good to me.

PR #13223, opened 2023-09-30
Add -Wstrict-prototypes -Wmissing-prototypes warnings check to CI

ychin:
Add two new warnings to CI:

- `strict-prototypes` helps prevent declaring a function with an empty argument list, e.g. `int func()`. In C++, that's equivalent to `int func(void)`, but in C, that means a function that can take any number of arguments which is rarely what we want.

- `missing-prototypes` makes sure we use `static` for file-only internal functions. Non-static functions should have been declared on a prototype file.

Also, fix up misc code warnings that this found.

GTK header needs to have #pragma warning suppressiong because GTK2 headers will warn on `-Wstrict-prototypes`, and it's included by gui.h and so we can't just turn off the warning in a couple files.


ychin:
I mentioned in https://github.com/vim/vim/pull/13166, but I think we should just add these warnings to be checked in CI.

chrisbra:
thanks, let me squash in #13226 as well into this

Issue #11157, opened 2022-09-18
Allow more C99 features


senioria:
It's said in the document that some C99 features are not used in Vim:

https://github.com/vim/vim/blob/b2f0ca820eae50994745106d824e215d87bd7926/runtime/doc/develop.txt#L218-L227

However, it may be possible to loose some of these restrictions:

- After dropping support for pre-Vista Windows systems, all Windows Vim users can use a newer version of MSVC (MSVC 2013) to compile code with declaration after statements. Besides, not allowing declarations after statements had caused some issues in Linux kernel; so it may be reasonable to also drop support to old MSVC, and allow this feature;
- VLAs are widely disallowed; maybe a reason other than compiler support can be used;
- [The thread reporting that flexible array members are not supported by HP-UX C compiler](https://groups.google.com/g/vim_dev/c/c6p9CXp6Zjw/m/dDs-QihHCQAJ) is 5 years old, and even at that time, it's a legacy problem; so this restriction may safely be removed now.

brammool:
> It's said in the document that some C99 features are not used in Vim:
> 
> https://github.com/vim/vim/blob/b2f0ca820eae50994745106d824e215d87bd7926/runtime/doc/develop.txt#L218-L227
> 
> However, it may be possible to loose some of these restrictions:
> 
> - After dropping support for pre-Vista Windows systems, all Windows
> Vim users can use a newer version of MSVC (MSVC 2013) to compile code
> with declaration after statements. Besides, not allowing declarations
> after statements had caused some issues in Linux kernel; so it may be
> reasonable to also drop support to old MSVC, and allow this feature;

We can see if this will work now.  I do know that in the past one or two
of these new style declarations slipped through and some people
complained.  I'm not sure if that is related to dropping XP support,
probably not.  If a newer MSVC version is used then we'll find the next
compiler where it fails...

I actually like declaring all variables at the start of the block, that
way they are easy to find.  The main advantage of C99 is being able to
declare the loop variable inside "for".  We could use that in a couple
of places and see who has a problem with it.

> - VLAs are widely disallowed; maybe a reason other than compiler
> support can be used;

Variable length arrays by themselves are not all that useful.  We have
the garray_T which works well, because it stores the size as well.  And
allows for dynamic growing as well.

> - [The thread reporting that flexible array members are not supported
> by HP-UX C
> compiler](https://groups.google.com/g/vim_dev/c/c6p9CXp6Zjw/m/dDs-QihHCQAJ)
[...]

PR #13224, opened 2023-09-30
Harden crypto 'key' option: turn off cmdline completion, disable set-=

ychin:
"set-=" can be used maliciously with a crypto key, as it allows an attacker (who either has access to the computer or a plugin author) to guess a substring by observing the modified state. Simply turn off set+=/-=/^= for this option as there is no good reason for them to be used.

Update docs to make that clear as well.

Also, don't allow cmdline completion for 'key' as it just shows ***** which is not useful and confusing to the user what it means (if the user accidentally hits enter they will have replaced their key with "*****" instead).

### DIFF
--- a/src/testdir/test_crypt.vim
+++ b/src/testdir/test_crypt.vim
@@ -438,4 +438,27 @@ func Test_crypt_set_key_segfault()
   bwipe!
 endfunc
 
+func Test_crypt_set_key_disallow_append_subtract()
+  new Xtest4
+
+  set key=foobar
+  call assert_true(&modified)
+  setl nomodified
+
+  call assert_fails('set key-=foo', 'E474:')
+  call assert_fails('set key-=bar', 'E474:')
+  call assert_fails('set key-=foobar', 'E474:')
+  call assert_fails('set key-=test1', 'E474:')
+
+  call assert_false(&modified)
+  call assert_equal('*****', &key)
+
+  call assert_fails('set key+=test2', 'E474:')
+  call assert_fails('set key^=test3', 'E474:')
+
+  call assert_false(&modified)
+  set key=
+  bwipe!
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -4928,6 +4928,8 @@ A jump table for the options with a short description can be found at |Q_op|.
 	"echo &key".  This is to avoid showing it to someone who shouldn't
 	know.  It also means you cannot see it yourself once you have set it,
 	be careful not to make a typing error!
+	You also cannot use |:set-=|, |:set+=|, |:set^=| on this option to
+	prevent an attacker from guessing substrings in your key.
 	You can use "&key" in an expression to detect whether encryption is
 	enabled.  When 'key' is set it returns "*****" (five stars).
 
--- a/src/option.c
+++ b/src/option.c
@@ -1315,13 +1315,6 @@ ex_set(exarg_T *eap)
  * :set operator types
  */
 typedef enum {
-    OP_NONE = 0,
-    OP_ADDING,		// "opt+=arg"
-    OP_PREPENDING,	// "opt^=arg"
-    OP_REMOVING,	// "opt-=arg"
-} set_op_T;
-
-typedef enum {
     PREFIX_NO = 0,	// "no" prefix
     PREFIX_NONE,	// no prefix
     PREFIX_INV,		// "inv" prefix
@@ -1935,7 +1928,7 @@ do_set_option_string(
 	char_u	    **argp,
 	int	    nextchar,
 	set_op_T    op_arg,
-	int	    flags,
+	long_u	    flags,
 	int	    cp_val,
 	char_u	    *varp_arg,
 	char	    *errbuf,
@@ -2037,7 +2030,7 @@ do_set_option_string(
 	// be triggered that can cause havoc.
 	*errmsg = did_set_string_option(
 			opt_idx, (char_u **)varp, oldval, newval, errbuf,
-			opt_flags, value_checked);
+			opt_flags, op, value_checked);
 
 	secure = secure_saved;
     }
@@ -7802,6 +7795,10 @@ ExpandOldSetting(int *numMatches, char_u ***matches)
     char_u  *var = NULL;	// init for GCC
     char_u  *buf;
 
+    if (expand_option_idx >= 0 &&
+	    (options[expand_option_idx].flags & P_NO_CMD_EXPAND))
+	return FAIL;
+
     *numMatches = 0;
     *matches = ALLOC_MULT(char_u *, 1);
     if (*matches == NULL)
@@ -7897,6 +7894,10 @@ ExpandSettingSubtract(
     if (expand_option_idx < 0)
 	// term option
 	return ExpandOldSetting(numMatches, matches);
+
+    if (expand_option_idx >= 0 &&
+	    (options[expand_option_idx].flags & P_NO_CMD_EXPAND))
+	return FAIL;
 
     char_u *option_val = *(char_u**)get_option_varp_scope(
 	expand_option_idx, expand_option_flags);
--- a/src/option.c
+++ b/src/option.c
@@ -7369,6 +7369,15 @@ set_context_in_set_cmd(
     else
 	expand_option_idx = opt_idx;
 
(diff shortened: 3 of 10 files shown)


ychin:
Noticed this issue when working on https://github.com/vim/vim/pull/13182

chrisbra:
Hm, this throws this error in CI:
```
In file included from vim.h:1941,
                 from option.c:39:
option.h:21:25: error: unsigned conversion from ‘long long int’ to ‘long unsigned int’ changes value from ‘4294968844’ to ‘1548’ [-Werror=overflow]
   21 | #define P_STRING        0x04    // the option is a string
      |                         ^~~~
optiondefs.h:1470:29: note: in expansion of macro ‘P_STRING’
 1470 |     {"key",         NULL,   P_STRING|P_ALLOCED|P_VI_DEF|P_NO_MKRC|P_NO_CMD_EXPAND,
      |                             ^~~~~~~~
gcc -c -I.       -Iproto -DHAVE_CONFIG_H     -O2 -fno-strength-reduce -Wall -Wno-deprecated-declarations -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -Wall -Wextra -Wshadow -Werror -Wno-deprecated-declarations -Wno-error=maybe-uninitialized   -DEXITFREE      -Wno-error=unused-function -Wno-shadow   -Wno-error=unused-parameter  -o objects/optionstr.o optionstr.c
gcc -c -I. -Iproto -DHAVE_CONFIG_H     -O2 -fno-strength-reduce -Wall -Wno-deprecated-declarations -D_REENTRANT -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -Wall -Wextra -Wshadow -Werror -Wno-deprecated-declarations -Wno-error=maybe-uninitialized   -DEXITFREE     -o objects/os_unix.o os_unix.c
```

zeertzjq:
Maybe `P_NO_CMD_EXPAND` can be set to `0x20`, which doesn't have a flag?

chrisbra:
I tried the following on top of your patch and it seems to work:

```patch
diff --git a/src/option.c b/src/option.c
index e25ef5a9a..fe7f52149 100644
--- a/src/option.c
+++ b/src/option.c
@@ -2968,7 +2968,7 @@ insecure_flag(int opt_idx, int opt_flags)
        }

     // Nothing special, return global flags field.
-    return &options[opt_idx].flags;
+    return (unsigned long int *) &options[opt_idx].flags;
 }
 #endif

diff --git a/src/optiondefs.h b/src/optiondefs.h
index f850d19b1..12370817e 100644
--- a/src/optiondefs.h
+++ b/src/optiondefs.h
@@ -263,7 +263,7 @@ struct vimoption
 {
     char       *fullname;      // full option name
     char       *shortname;     // permissible abbreviation
-    long_u     flags;          // see in option.h
+    long_u8    flags;          // see in option.h
     char_u     *var;           // global option: pointer to variable;
                                // window-local option: VAR_WIN;
                                // buffer-local option: global value
diff --git a/src/vim.h b/src/vim.h
index c78b12b16..342c4c001 100644
--- a/src/vim.h
+++ b/src/vim.h
@@ -364,12 +364,21 @@ typedef unsigned int      int_u;
 // On Win64, longs are 32 bits and pointers are 64 bits.
 // For printf() and scanf(), we need to take care of long_u specifically.
 #ifdef _WIN64
+typedef unsigned __int64       long_u8;
 typedef unsigned __int64       long_u;
 typedef                 __int64        long_i;
 # define SCANF_HEX_LONG_U       "%llx"
 # define SCANF_DECIMAL_LONG_U   "%llu"
 # define PRINTF_HEX_LONG_U      "0x%llx"
+#elif defined(i386) || defined(__i386__) || defined(__i386) || defined(_M_IX86)
+typedef unsigned long          long_u;
+typedef long long int          long_u8;
+typedef                 long           long_i;
+# define SCANF_HEX_LONG_U       "%lx"
+# define SCANF_DECIMAL_LONG_U   "%lu"
+# define PRINTF_HEX_LONG_U      "0x%lx"
 #else
+typedef unsigned long          long_u8;
 typedef unsigned long          long_u;
 typedef                 long           long_i;
 # define SCANF_HEX_LONG_U       "%lx"
```

not sure if this is the best way to do and keep compatible with i386

ychin:
> Maybe `P_NO_CMD_EXPAND` can be set to `0x20`, which doesn't have a flag?

I could do that. I looked at code history and seems like 0x20 has been missing since Vim 7, so maybe it's some really early flag that got removed and remains unused (didn't feel like digging through older code history since they aren't in the main git repo). We still have to address this problem when we want to add new flags though. I can at least leave a comment there indicating that we need to fix this.

We can either fix this by using a 64-bit int, or two separate flags (e.g. a `flag` and `flag2` variable).

---

> I tried the following on top of your patch and it seems to work:

I'm not a huge fan of declaring a new typedef personally because we may end up running into more compatibility issues on platforms that are not tested in CI, and I find the name confusing (most people think in how many bits a number has, not bytes, so it should be "64", not "8").

Now, since we technically moved to C99 (see #11157 for more discussions), we should technically have access to `<stdint.h>` which defines the convenient `uint64_t` for us which seems like what we should use instead (incidentally `long_u` probably could have been `uintptr_t` but Vim obviously predates the popularity of C99.

One issue is that I'm not 100% sure if C99 support is fully rolled out. The only real problematic place I see is OpenVMS support, where `os_vms_conf.h` seems to suggest that the max value of `long long` under VAX is only 32-bit. Also, the platform undefines `HAVE_STDINT_H` so we may not have access to the header. Looking around (https://github.com/vim/vim/issues/11157#issuecomment-1327813832 and https://groups.google.com/g/comp.os.vms/c/Bnh3tIOc7bo) I guess only the latest VMS has support for stdint.h and C99. Either way the VMS version of Vim linked to from vim.org (http://www.polarhome.com/vim/) is still on Vim 8.2 so I'm not sure if this is even still technically "supported" or not as Vim only started migrating to C99 in Vim 9 and I didn't see too much discussion about OpenVMS in that discussion.

On this topic, I feel like I don't fully know what the full list of "supported" platforms for Vim are. Maybe that's because it's been a little opportunistic in a "fix it as much as possible" and Vim is used in so many places across different versions that it's hard to have a complete count compared to newer / more controlled editors.

---

So my suggestion is I just use 0x20 and punt this down the line for the tomorrow version of us 😅.

To properly fix this I would rather just use uint64_t and polyfill it with `long long` if that's not defined. I would rather use existing C-standard names that is well understood (for flags, it's good to have explicit bitness guarantees).

The docs does say we should have guaranteed `long long`, but I don't know if that sentence was written with OpenVMS or other obscure platforms in mind:

https://github.com/vim/vim/blob/e379e21ddc2dc12abea9dcfd0708e27788bb568c/runtime/doc/develop.txt#L213-L216

zeertzjq:
I can find the removal of `0x20` here: https://github.com/vim/vim-history/commit/c3cba9278b34e6ff131e99d5ef39e346176957de

src/testdir/test_crypt.vim
```diff
@@ -438,4 +438,27 @@ func Test_crypt_set_key_segfault()
   bwipe!
 endfunc
 
+func Test_crypt_set_key_disallow_append_subtract()
```
zeertzjq, line 441: This should also need `CheckFeature cryptv`

ychin: That already exists at the top of the file. Every test in this file is to test the crypto feature so I don't think we need to add that in each test?

The only other feature checking the tests do are `CheckFeature sodium` since that's an additional requirement.

zeertzjq: Oops, didn't notice the name of the file.

ychin:
Ok I pushed a new commit to just use the unused 0x20 since I don't want to fix this now :)

Speaking of which, I find some of the flags kind of… suspect. `P_DEF_ALLOCED`, `P_WAS_SET`, and `P_INSECURE` are actually runtime properties on an option, rather than a specification of an option. `P_INSECURE` also needs to work on window-specific options which is why each window has members like `w_p_stl_flags` whose existence is just to hold those flags. Personally, if we need more flags in the future, I would refactor those away to a "runtime_flags" first, as it makes a lot more sense to me to separate specification flags and runtime values. It only gives a grand total of 3 additional flags for us to use but I think that's a little clearer anyway.

chrisbra:
thanks, looks good now

PR #13229, opened 2023-09-30
Check for xattr may fail

zeertzjq:
Problem:  Check for xattr may fail.
Solution: Check for "sys/xattr.h" instead of "attr/xattr.h".

Fix #13227

PR #13230, opened 2023-09-30
macos: --disable-xattr

pheiduck:
As in (Linux only)  [patch 9.0.1962](https://github.com/vim/vim/commit/e085dfda5d8dde064b0332464040959479696d1c) we can disable the check on macOS
Follow-up: https://github.com/vim/vim/pull/13229

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -326,7 +326,7 @@ jobs:
 
       - name: Configure
         run: |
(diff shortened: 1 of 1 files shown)


chrisbra:
I don't this is right. This just makes sure, that the CI is compiled with non-xattr support, but we should rather disable xattr completly when compiling on Mac OS X. (because otherwise all Mac users will have to add the new configure flag `--disable-xattr`). At least until we have working xattr support on Mac OS X

That's what #13232 is trying to perform.

PR #13234, opened 2023-09-30
configure: Do not print "6" when checking for libruby

illia-bobyr:
`expr` will print the matched string length to the standard output. Current `configure` output looks like this:

```
checking Ruby header files...  /usr/include/ruby-3.1.0
6
```

The script really only cares about `expr` exit code.


ilya-bobyr:
`linux (huge, gcc, uchar, testgui, true, dynamic)` has timed out after 25 minutes in the Test step:

```
07:22 Executing Test_hidden_terminal()
07:22 Executing Test_issue_5607()
07:22 Executing Test_term_and_startinsert()
07:22 Executing Test_term_func_invalid_arg()
07:22 Executing Test_term_keycode_translation()
Error: The action has timed out.
```

PR #13237, opened 2023-10-01
Clean up cmdline option tab-completion code

ychin:
- Fix manual array size calculations to just use `ARRAY_LENGTH()`.
- Fix unintentional typo in comments due to copy-paste error.
- Fix assert_equal() usages to pass the expected value to first parameter instead of 2nd one to avoid confusion.
- Fix misc signed vs unsigned warnings.

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -7989,7 +7989,7 @@ ExpandSettingSubtract(
 	    return FAIL;
 	}
 
-	int num_flags = STRLEN(option_val);
+	size_t num_flags = STRLEN(option_val);
 	if (num_flags == 0)
 	    return FAIL;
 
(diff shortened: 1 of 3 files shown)


ychin:
Per this comment (https://github.com/vim/vim/pull/13182#issuecomment-1741975961), the last commit also introduced some signed vs unsigned warnings. Added a commit to address that here. Apparently Appveyor CI build warnings are not considered errors.

chrisbra:
I guess you now added some style errors.

ychin:
> I guess you now added some style errors.

Should be fixed

ychin:
Feel free to combine with #13249 as it's related.

chrisbra:
thanks, will do!

PR #13247, opened 2023-10-02
Fix typo in document (Related: #12516)

h-east:
#12516 is closed by Bram.  However, it is not reflected.
This PR is #12516 with some typo corrections.

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -788,7 +788,7 @@ An example for subtracting (which isn't very useful): >
 On this text:
 	1 one ~
 	2 two ~
-	3 three FOLDED~
+	3 three FOLDED ~
 	4 four FOLDED ~
 	5 five FOLDED ~
 	6 six FOLDED ~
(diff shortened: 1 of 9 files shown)


h-east:
Hmm, Tag E1363 duplicate.
```
runtime/doc/vim9.txt|1560 col 34-39| If a type is incomplete you get *E1363* , e.g. when you have an object for
runtime/doc/vim9class.txt|661 col 58-63| even when the variable name is invalid. *E1360* *E1362* *E1363*
```

Originally there was a tag label for E1363 in vim9.txt, but it was not recognized as a tag label because of the ',' immediately following it.
I have corrected that in this PR, which caused the tag duplication.
https://github.com/vim/vim/pull/13247/commits/bc64426139947ce0f472b3256f0ec2bf1912494d#diff-276d2fe6d42680e6055997d47473f23e9e2811387be6377ccc488071bb9d324fL1560-R1560

I think it would be better to remove the vim9class.txt's one.

h-east:
My modification to syntax.txt was incorrect. It has been reverted.
I thought the indentation was a subtle white space. :sweat:

chrisbra:
thanks, let me include this

PR #13250, opened 2023-10-02
win32: Add "**" test

k-takata:
Vim supports "**" on MS-Windows. However, it is not tested by `Test_glob_extended_bash`.

Unlike Unix, it doesn't use 'shell' and doesn't support {,} expansion. So, I added as a separate test.

Related: #13205


chrisbra:
thanks!

PR #13251, opened 2023-10-02
Fix macOS FEAT_SOUND guards to be less restrictive

ychin:
This allows +sound to work on older macOS platforms again. The +sound implementation uses APIs available in 10.6, but the code itself uses generics with type parameters which was only added in Xcode 7 / clang 7, which was released for macOS 10.11. This means as long as Vim is compiled under 10.11+, and using a deployment target >= 10.6, the feature will work.

### DIFF
--- a/src/vim.h
+++ b/src/vim.h
@@ -173,8 +173,8 @@
 #  define FEAT_CLIPBOARD
 # endif
 # if defined(FEAT_HUGE) && !defined(FEAT_SOUND) && \
-   defined(MAC_OS_X_VERSION_MIN_REQUIRED) && \
-    MAC_OS_X_VERSION_MIN_REQUIRED >= 101100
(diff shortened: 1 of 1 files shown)


ychin:
See https://github.com/vim/vim/pull/13115#issuecomment-1738012879

ychin:
From testing in VMs, it seems to work fine. I can compile this on macOS 10.11 (using Xcode / clang 7) and can deploy to older versions fine. And I confirmed using an older macOS 10.9 (using Xcode / clang 5) that it would correctly guard against this and turn off FEAT_SOUND as desired (otherwise it would fail compilation as it doesn't understand the syntax for typed parameters).

sevan:
Built fine on 10.4/5 auto-disabling sound support.

PR #13252, opened 2023-10-02
win32: Better support of Per-Monitor V2 High DPI

### LINKED ISSUE #11812: Dragging GVIM between monitors of different DPIs jumps away from the cursor
### Steps to reproduce

1. Requires multiple monitors
2. Settings, system, display, click on each monitor and verify that the Scale (DPI) value for each monitor is different.
3. Drag GVIM window between monitors

It may be necessary to make the window a bit wider and drag from both the left/right side of the title bar to experience the broken behavior.

This is caused by the app handling WM_DPICHANGED incorrectly. This window message's lParam is a rect that represents the new window size and position (at the new DPI). The app MUST call SetWindowPos with this exact rect.

The app instead is picking its own new size (using rows/columns) but this causes the window to jump, The rect provided by the system when dragging between monitors is adjusted to keep the window over the same place on the window and picking another arbitrary position or size will cause the window to jump away from the cursor. It even may cause the window to jump back to another monitor and change its DPI again (potentially bouncing between the two monitors).

To control the size when dragged between monitors of different DPIs the app should use WM_GETDPISCALEDSIZE. This is sent prior to WM_DPICHANGED and allows the app to pick an explicit size at the new DPI (which then becomes the size of the rect provided by WM_DPICHANGED).

https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-getdpiscaledsize

WM_GETDPISCALEDSIZE's lParam is a SIZE. This size is the proposed size, scaled to the window's current DPI. To pick a 'non-linear' size (to override the size at the new DPI) the app must scale this size from the current DPI to the DPI provided as wParam. Note that windows with non-client area that want to scale only their client rect (for example to size it based on a font) must use AdjustWindowRectForDpi.

Calling this API with an empty rect and the current DPI shows the size of the non-client area that needs to be removed from the proposed size from the system and calling again with the scaled client rect and the proposed DPI adds the non-client area at the new DPI.

### Expected behaviour

The window should resize (because it is changing its DPI) but the cursor should remain in the same logical position over the window (left/middle/right of title bar).

### Version of Vim

0.0.118

### Environment

Windows 11

### Logs and stack traces

_No response_

k-takata:
Closes #11812

Implement WM_GETDPISCALEDSIZE.
It is not so easy to calculate the new size without actually changing the size. So, this returns an approximate size.

This doesn't work well when a bitmap font (e.g. FixedSys) is selected, but I think this is acceptable.

### DIFF
--- a/src/gui_w32.c
+++ b/src/gui_w32.c
@@ -226,6 +226,10 @@ gui_mch_set_rendering_options(char_u *s)
 # define WM_DPICHANGED			0x02E0
 #endif
 
+#ifndef WM_GETDPISCALEDSIZE
+# define WM_GETDPISCALEDSIZE		0x02E4
+#endif
+
 #ifndef WM_MOUSEHWHEEL
 # define WM_MOUSEHWHEEL			0x020E
 #endif
@@ -379,6 +383,7 @@ typedef enum DPI_AWARENESS {
 static int		s_dpi = DEFAULT_DPI;
 static BOOL		s_in_dpichanged = FALSE;
 static DPI_AWARENESS	s_process_dpi_aware = DPI_AWARENESS_INVALID;
+static RECT		s_suggested_rect;
 
 static UINT (WINAPI *pGetDpiForSystem)(void) = NULL;
 static UINT (WINAPI *pGetDpiForWindow)(HWND hwnd) = NULL;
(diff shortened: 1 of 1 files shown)


chrisbra:
thank you!

PR #13256, opened 2023-10-03
Cannot scroll up in diff mode with many filler lines and scrolloff=0

zeertzjq:
Problem:  Cannot scroll up in diff mode with many filler lines and zero
          'scrolloff'.
Solution: Invalidate w_cline_row before calling comp_botline().

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -3262,6 +3262,7 @@ onepage(int dir, long count)
 #ifdef FEAT_DIFF
 			curwin->w_topfill = 0;
 #endif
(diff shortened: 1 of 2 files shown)


zeertzjq:
Note: this fixes a regression from patch 9.0.0807. The added test fails in v9.0.0807 but passes in v9.0.0806.

chrisbra:
thanks. I just noticed a similar issue with very long wrapped lines, that take up the whole window size and then  `<c-b>` was not able to scroll properly. 
Took me a while before I noticed I did not scroll up properly when reading long logfiles. I hope this also fixes this issue.

Issue #3236, opened 2018-07-20
Bug of ag or bug of vim?


Yggdroot:
My plugin [LeaderF](https://github.com/Yggdroot/LeaderF) encountered an [issue](https://github.com/Yggdroot/LeaderF/issues/190), [ag](https://github.com/ggreer/the_silver_searcher) can not list files on Windows.

**Steps to reproduce:**
1. enter vim
2. run the below commands:
   ```python
   :py import subprocess
   :py print subprocess.Popen('ag -g "" .', shell=True, stdout=subprocess.PIPE).stdout.read()
   ```
no output.(suppose there are some files in current directory)

**Expected result:**

all the files can be listed just like running `ag -g "" .` in git-bash.

If purely run the python code, everything works fine.
For example, execute the command ` py -2 test.py`, `test.py` is as below:
   ```python
   import subprocess
   print ( subprocess.Popen('ag -g "" .', shell=True, stdout=subprocess.PIPE).stdout.read() )
   ```
So, I think this probably is a bug of vim.

chrisbra:
Have you tried with a different binary than ag?

Yggdroot:
I have tried [rg](https://github.com/BurntSushi/ripgrep), it works well.

JRuiChen:
I have encountered the same problem, but when i change 'ag' to 'Ag', the LeaderFile works well.  For my installation of ag, the system don't accept 'ag' but 'Ag'. Can you try it ?

Yggdroot:
@JRuiChen  'Ag' does not work for me either.

JRuiChen:
@Yggdroot

JRuiChen:
@Yggdroot  
enter vim
:py import subprocess
:py print subprocess.Popen('ag -g "" .', shell=True, stdout=subprocess.PIPE).stdout.read()
I have no output either.

1. The problem i faced is,  in  .vimrc:
      **let g:Lf_IndexTimeLimit=20
      let g:Lf_DefaultExternalTool="ag"**
      let g:Lf_RootMarkers=['version.txt', 'server', 'client']
Enter vim , the LeaderFile command has no output.   In .LFCache Dir, there is no cache files greater than 0 created.
2. The i changed .vimrc to:
      **let g:Lf_IndexTimeLimit=120
      let g:Lf_DefaultExternalTool="Ag"**
      let g:Lf_RootMarkers=['version.txt', 'server', 'client']
[...]

PR #13260, opened 2023-10-04
if_python3,win32: Fix Python3 stable ABI

k-takata:
There were some issues in current stable ABI implementation on Windows:
* Python DLL name should be `python3.dll` instead of `python311.dll` and so on. (See: https://docs.python.org/3/c-api/stable.html)
* Some non-stable API functions were used:
  - `_PyObject_NextNotImplemented`
  - `PyStdPrinter_Type`
* `reset_stdin()` and `hook_py_exit()` didn't work with `python3.dll`. `python3.dll` is a special type of DLL called forwarder DLL. It just forwards the functions to other DLL (e.g. `python311.dll`). There were two issues regarding these functions:
  - `python3.dll` doesn't have import tables. This caused a crash in `get_imported_func_info()`. Add a check whether the specified DLL has an import table.
  - `reset_stdin()` and `hook_py_exit()` should be applied to the forwarded DLL (e.g. `python311.dll`), not to `python3.dll`. Check the export directory of `python3.dll` to find the forwarded DLL and apply the functions to it.


ychin:
Thanks for looking into this. I kind of did the bare minimum for Windows when implementing it.


>       * `python3.dll` doesn't have import tables. This caused a crash in `get_imported_func_info()`. Add a check whether the specified DLL has an import table.
>       * `reset_stdin()` and `hook_py_exit()` should be applied to the forwarded DLL (e.g. `python311.dll`), not to `python3.dll`. Check the export directory of `python3.dll` to find the forwarded DLL and apply the functions to it.

Looking at the code, you calling `get_imported_func_info` on the forwarded DLL already right? Wouldn't that work? Are you adding a check to `get_imported_func_info` just to be safe (which isn't a bad thing)?

FWIW I kind of dislike how the Python subsystem share the same stdin/stdout with the main Vim process. It leads to odd issues like #3236. But then it's kind of hard to fix the issue because we are just loading a DLL in to the main process instead of interacting with Python as a separate process.

k-takata:
> Looking at the code, you calling `get_imported_func_info` on the forwarded DLL already right? Wouldn't that work? Are you adding a check to `get_imported_func_info` just to be safe (which isn't a bad thing)?

No. Only setting `pythonthreedll=python3.dll` without my patch, `get_imported_func_info` was called on `python3.dll` and it caused a crash.
If pythonthreedll points a forwarder DLL (python3.dll), then we have to find the actual DLL (e.g. python311.dll) and have to call `get_imported_func_info` on the actual DLL.

chrisbra:
thanks!

ychin:
Speaking of which would it make sense to switch the Windows builds (https://github.com/vim/vim-win32-installer) to use stable Python ABI? I would imagine it should be strictly better but I don't know if there are edge cases in the Windows environment that I'm thinking of. I think the python3.dll file is usually located in the same location right (C:\Windows\system32)? I don't know what happens when the user has multiple versions of Python installed though (assuming they are using the standard installer, which a lot of people don't, as there are at least a million ways to install Python).

This way users don't have to worry when a new Python version comes out and it "just works".

k-takata:
> I think the python3.dll file is usually located in the same location right (C:\Windows\system32)?

No, it is installed in the python installation directory (normally `C:\Program Files\Python3XX` (64-bit), `C:\Program Files (x86)\Python3XX-32` (32-bit), or each user's AppData directory). It will be added to PATH only when the user selected it.

Currently, if_python3 can find python3XX.dll from the registry even the python installation directory is not in PATH, but it doesn't work when python3.dll is used. This is because python3.dll just forwards to python3XX.dll but the forwarded DLL can be loaded only when it is in PATH.
I'm now thinking if there is a workaround for this.

k-takata:
See #13315

PR #13261, opened 2023-10-04
CI: re-enable Cirrus CI

zeertzjq:
Ref patch 9.0.1912:

> Perhaps at the beginning of the next month we can revisit and enable
> just a build without testing it.  Hopefully this is won't take too
> many credits and we can at least verify that building works.

Actually enabling testing should be fine. In the last month there were
three Cirrus CI jobs and credits ran out on Sep 15, but now there is
only one Cirrus CI job, so credits shouldn't run out.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,20 +1,20 @@
-#env:
-#  CIRRUS_CLONE_DEPTH: 3
-#  FEATURES: huge
-#
-#freebsd_task:
-#  name: FreeBSD
-#  matrix:
-#    - name: FreeBSD 14.0
-#      freebsd_instance:
-#        image_family: freebsd-14-0
-#  timeout_in: 20m
-#  install_script:
-#    - pkg install -y gettext
-#  build_script:
-#    - NPROC=$(getconf _NPROCESSORS_ONLN)
-#    - ./configure --with-features=${FEATURES}
-#    - make -j${NPROC}
+env:
+  CIRRUS_CLONE_DEPTH: 3
+  FEATURES: huge
+
+freebsd_task:
+  name: FreeBSD
+  matrix:
+    - name: FreeBSD 14.0
+      freebsd_instance:
+        image_family: freebsd-14-0
+  timeout_in: 20m
+  install_script:
+    - pkg install -y gettext
+  build_script:
+    - NPROC=$(getconf _NPROCESSORS_ONLN)
+    - ./configure --with-features=${FEATURES}
+    - make -j${NPROC}
 #  test_script:
 #    - src/vim --version
 #      # run tests as user "cirrus" instead of root


zeertzjq:
Actually enabling testing should be fine. In the last month there were
three Cirrus CI jobs and credits ran out on Sep 15, but now there are
only one Cirrus CI jobs, so credits shouldn't run out.

chrisbra:
I am not sure this will work. We can try out until we run out of credits for the month, but I believe testing will take too much time. 

According to https://cirrus-ci.com/settings/github/vim we have used 22 credits for BSD until mid september and we are [allowed to use 50](https://cirrus-ci.org/faq/#are-there-any-limits). So perhaps it will work out, depending on how many commits we will run I guess.

chrisbra:
thanks, I have now enabled the test again. Let's see how far we come with this :)

ychin:
I think we can actually get the tests to run much faster if we want. Right now the tests have to run serially, and when I run them locally, the CPU is mostly sitting idle, meaning that the tests are either sleeping/waiting on some event or just generally not really optimizing its use of the CPU. However, you can't just naively run `make -j10` to run 10 tests concurrently because the tests share the same `testdir` folder (some tests generate temporary files), and more importantly they use the same tty.

If I just split the `newtestssilent: $(NEW_TESTS_RES)` target in `src/testdir/Makefile` into multiple smaller targets (by splitting `NEW_TESTS_RES` up), and then run each test in a separate folder (e.g. "testdir2/3/4"), and a separate terminal, I can reduce the time for the full tests to finish on my own computer significantly. It takes 6+ minutes for just running the Vimscript tests, but if I split them up and run in 4 terminals separately I can easily get them completed in 1-2 minutes.

I guess it depends on how the CPUs are provisioned on Cirrus CI. I don't know how many CPUs you get by default for the FreeBSD image. But I would imagine even in 1-2 CPU VMs, splitting the tests can really speed up running them as they aren't CPU-bound. We would probably need to run each chunk of tests in a separate tty though, so may be use tmux or screen to facilitate that and just dump the logs when each chunk is done?

I was always annoyed with the long test times but since GitHub Actions is free didn't really have enough incentive to work on this.

chrisbra:
yeah, we have been slowly migrating from command test names to unique test names, but not sure if the testsuite is already ready to be run in parallel. Splitting up into separate testdirs sounds like a good idea however.

ychin:
Hmm, what did you mean by "command test name" and "unique" ones? Aren't the "new" tests all just independent scripts that you can run? They all require spawning a new Vim instance so they can't really share state anyway, and you have always been able to run each test independently. I think as long as we can separate the tty / folder they should work (they seem to work from local testing as well).

chrisbra:
Oh I meant `common` temp filenames. We used to just use always the same name, like test 1 edited a file `Xtestfile` and perhaps test4 also tried to test something totally different but used the same name. Which works when run sequentually but may fail when run in parallel.

ychin:
Oh ok right. So we were talking about the same thing. I think the tests also all reuse the test logging files like "messages" and `starttime` as well, so there are more than just the temp file names that would conflict.

PR #13263, opened 2023-10-04
Fix Windows clean build target in Make_mvc.mak

ychin:
Make sure to delete the object files in the correct location.

### DIFF
--- a/src/Make_mvc.mak
+++ b/src/Make_mvc.mak
@@ -1312,7 +1312,7 @@ notags:
 
 clean: testclean
 	- if exist $(OUTDIR)/nul $(DEL_TREE) $(OUTDIR)
-	- if exist *.obj del *.obj
+	- if exist $(OUTDIR)\*.obj del $(OUTDIR)\*.obj
 	- if exist $(VIM).exe del $(VIM).exe
 	- if exist $(VIM).ilk del $(VIM).ilk
 	- if exist $(VIM).pdb del $(VIM).pdb


src/Make_mvc.mak
```diff
@@ -1312,7 +1312,7 @@ notags:
 
 clean: testclean
 	- if exist $(OUTDIR)/nul $(DEL_TREE) $(OUTDIR)
-	- if exist *.obj del *.obj
+	- if exist $(OUTDIR)\*.obj del $(OUTDIR)\*.obj
```
k-takata, line 1315: I think this is correct.

`$(OUTDIR)\*.obj` will be deleted by `$(DEL_TREE)` command in the previous line.
Some .obj files will be created in current directory, so `del *.obj` is needed.

k-takata: Currently, the following files are created in the `src` directory, not in `$(OUTDIR)`:
dosinst.obj, uninstall.obj, vimrun.obj

It might be better to move them to `$(OUTDIR)`, though.

ychin: Hmm, ok. I think the issue is that `$(OUTDIR)/nul` is not created. At least when I'm building locally it doesn't seem to be created and as such the `$(DEL_TREE)` never gets called. Do you think it's some relic that got removed?

ychin: But yes you are right there are some odd .obj's being created. I guess I just wanted to make sure that the clean target is properly cleaning, and right now it doesn't seem to be doing so.

> It might be better to move them to `$(OUTDIR)` though.

I think it depends if they should be in the specific folders, because each configuration (e.g. GUI, DIRECTX, DEBUG) will result in a separate folder.

k-takata: ~~Oh, `$(OUTDIR)/nul` should be `$(OUTDIR)\nul`.~~ (Both `/nul` and `\nul` can be used.)
`dirname\nul` can be used to check the existence of `dirname` directory.

ychin: Hmm, something must be wrong either with the make file or my environment then (using VS 2022's developer tools). I could take a quick look.

ychin: Actually I think you are right. I must have made a mistake when I was looking into this. Let me close this. Thanks for sanity checking this.

PR #13264, opened 2023-10-04
Add guifont cmdline completion (for Windows and GTK)

ychin:
For Windows, auto-complete will only suggest monospace fonts as that's the only types allowed. Will also suggest font options after the colon, including suggesting the current font size for convenience, and misc charset and quality options like `cANSI` and `qCLEARTYPE`.

For GTK, auto-complete will suggest only monospace fonts for `guifont` but will include all fonts for `guifontwide`. The completion code doesn't currently suggest the current font size, as the GTK guifont format does not have a clear delimiter (':' for other platforms).

### DIFF
--- a/src/gui_gtk_x11.c
+++ b/src/gui_gtk_x11.c
@@ -5316,6 +5316,62 @@ gui_mch_free_font(GuiFont font)
 }
 
 /*
+ * Cmdline expansion for setting 'guifont' / 'guifontwide'. Will enumerate
+ * through all fonts for completion. When setting 'guifont' it will only show
+ * monospace fonts as it's unlikely other fonts would be useful.
+ */
+    void
+gui_mch_expand_font(optexpand_T *args, void *param, int (*add_match)(char_u *val))
+{
+    PangoFontFamily	**font_families = NULL;
+    int			n_families = 0;
+    int			wide = *(int *)param;
+
+    if (args->oe_include_orig_val && *args->oe_opt_value == NUL && !wide)
+    {
+	// If guifont is empty, and we want to fill in the orig value, suggest
+	// the default so the user can modify it.
+	if (add_match((char_u *)DEFAULT_FONT) != OK)
+	    return;
+    }
+
+    pango_context_list_families(
+	    gui.text_context,
+	    &font_families,
+	    &n_families);
+
+    for (int i = 0; i < n_families; i++)
+    {
+	if (!wide && !pango_font_family_is_monospace(font_families[i]))
+	    continue;
+
+	const char* fam_name = pango_font_family_get_name(font_families[i]);
+	if (input_conv.vc_type != CONV_NONE)
+	{
+	    char_u *buf = string_convert(&input_conv, (char_u*)fam_name, NULL);
+	    if (buf != NULL)
+	    {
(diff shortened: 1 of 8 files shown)


ychin:
Windows:
guifont_autocomplete_1
guifont_autocomplete_2

ychin:
GTK:

image

image

ychin:
Code coverage is complaining a lot of lines aren't testing. Most of them are due to guards around not being able to allocate memory. I'm following the pattern that I see in other areas in Vim's codebase and making sure to check against allocation results, but I don't know if there are good ways to test such situations.

yegappan:
> Code coverage is complaining a lot of lines aren't testing. Most of them are due to guards around not being able to allocate memory. I'm following the pattern that I see in other areas in Vim's codebase and making sure to check against allocation results, but I don't know if there are good ways to test such situations.

You can use the test_alloc_fail() function to test for allocation failures.  To use this function,
you need to add an ID in alloc_id_T and use the alloc_id() function for memory allocation.

ychin:
> You can use the test_alloc_fail() function to test for allocation failures. To use this function,
> you need to add an ID in alloc_id_T and use the alloc_id() function for memory allocation.

Hmm that seems like a lot of work though. As in, the pattern that I see (and therefore follow) usually call `vim_strsave` / `alloc` or some other generic ways to get an allocated buffer and then NULL-check to protect against memory allocation issues. Only a few are tagged with an alloc ID.

yegappan:
> > You can use the test_alloc_fail() function to test for allocation failures. To use this function,
> > you need to add an ID in alloc_id_T and use the alloc_id() function for memory allocation.
> 
> Hmm that seems like a lot of work though. As in, the pattern that I see (and therefore follow) usually call `vim_strsave` / `alloc` or some other generic ways to get an allocated buffer and then NULL-check to protect against memory allocation issues. Only a few are tagged with an alloc ID.

Bram initially added the basic infrastructure (in 7.4.1058 and 7.4.1073) for testing memory
allocation failures.  I used this infra in the features that I developed (quickfix, signs, 
and tag stack) and used this in a few other places.  As you said there are still a large
number of other places where this is not used yet.  But this is the only mechanism
available currently to test memory allocation failures.

ychin:
Right. That's useful context to know. It feels to me this is quite manual and difficult to roll out across the board as it requires exhaustively tagging all your allocation, and testing each feature against it.

FWIW I think sometimes test coverage % could be a fickle thing. I'm working on another PR to fix cmdline expansion code related to escaping commas and spaces, and there are some tests that while technically "cover" the lines in question by getting the code to visit that code path (therefore increasing the code coverage %), do not actually test that the underlying behavior (it will be fixed in my PR) works.

ychin:
As an aside, if you look at the change, in GTK Linux, if you don't have a `guifont` set, if you do `:set guifont=<Tab>` it will automatically suggest the default font "Monospace 10" for the user so they can adjust it. In Windows I decided not to implement this feature because I really hate the default Fixedsys font (maybe it's time to come up with a better default?) and not sure if there's any value in auto-filling the default. It can be changed of course.

chrisbra:
thanks!

PR #13266, opened 2023-10-04
win32: Improve default font

### LINKED ISSUE #12919: Default font choosen for Windows gvim 9.0 too small
### Steps to reproduce

1. Get Windows 10
2. Get latest (as of 2023-08-25) vim/gvim (9.0)
3. Either ensure no vim config set or launch `gvim --clean`

gvim 9.0 is defaulting to Fixedsys, Regular, 4

Font is unreadable on a 4k display (I don't have another display to check, but even a non-4k display the choosen font is too small).

## Screen shot of main window on desktop

image

image

## Screen shot of font settings (from Edit->Select Font...)

image

This appears to be https://github.com/vim/vim/issues/1059 and https://github.com/vim/vim/issues/4375.
NOTE #1059 was closed with comment https://github.com/vim/vim/issues/1059#issuecomment-306402092  "It looks like Windows 10 1703 solved this problem.", later comments appear to indicate later Windows 10 releases have similar issues. One suggested workaround in that thread is to make use of a Windows option to change the way Windows handles DPI https://github.com/vim/vim/issues/1059#issuecomment-294358464 this has **no** impact for me.

Selecting a reasonable font size (Consolas, Regular, 12) works for me (icons are not great but readable, I'm about to turn them off in my config ;-))

image

image

### Expected behaviour

To be able to read text when gvim opens.

### Version of Vim

9.0

### Environment

	C:\>cmd.exe /c ver

	Microsoft Windows [Version 10.0.19042.2965]

	C:\>vim --version
	VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jun 28 2022 13:10:49)
	MS-Windows 32-bit console version
	Compiled by appveyor@APPVYR-WIN
	Huge version without GUI.  Features included (+) or not (-):
	+acl                +ex_extra           +multi_lang         -tag_any_white
	+arabic             +extra_search       +mzscheme/dyn       -tcl
	+autocmd            -farsi              -netbeans_intg      +termguicolors
	+autochdir          +file_in_path       +num64              +terminal
	+autoservername     +find_in_path       +packages           -termresponse
	-balloon_eval       +float              +path_extra         +textobjects
	+balloon_eval_term  +folding            +perl/dyn           +textprop
	-browse             -footer             +persistent_undo    -tgetent
	++builtin_terms     +gettext/dyn        +popupwin           +timers
	+byte_offset        -hangul_input       -postscript         +title
	+channel            +iconv/dyn          +printer            -toolbar
	+cindent            +insert_expand      +profile            +user_commands
	+clientserver       +ipv6               +python/dyn         +vartabs
	+clipboard          +job                +python3/dyn        +vertsplit
	+cmdline_compl      +jumplist           +quickfix           +vim9script
	+cmdline_hist       +keymap             +reltime            +viminfo
	+cmdline_info       +lambda             +rightleft          +virtualedit
	+comments           +langmap            +ruby/dyn           +visual
	+conceal            +libcall            +scrollbind         +visualextra
	+cryptv             +linebreak          +signs              +vreplace
	+cscope             +lispindent         +smartindent        +vtp
	+cursorbind         +listcmds           +sodium/dyn         +wildignore
	+cursorshape        +localmap           +sound              +wildmenu
	+dialog_con         +lua/dyn            +spell              +windows
	+diff               +menu               +startuptime        +writebackup
	+digraphs           +mksession          +statusline         -xfontset
	-dnd                +modify_fname       -sun_workshop       -xim
	-ebcdic             +mouse              +syntax             -xpm_w32
	+emacs_tags         -mouseshape         +tag_binary         -xterm_save
	+eval               +multi_byte_ime/dyn -tag_old_static
	   system vimrc file: "$VIM\vimrc"
	     user vimrc file: "$HOME\_vimrc"
	 2nd user vimrc file: "$HOME\vimfiles\vimrc"
	 3rd user vimrc file: "$VIM\_vimrc"
	      user exrc file: "$HOME\_exrc"
	  2nd user exrc file: "$VIM\_exrc"
	       defaults file: "$VIMRUNTIME\defaults.vim"
	Compilation: cl -c /W3 /GF /nologo -I. -Iproto -DHAVE_PATHDEF -DWIN32  -DFEAT_CSCOPE -DFEAT_TERMINAL -DFEAT_SOUND  -DFEAT_JOB_CHANNEL -DFEAT_IPV6    -DHAVE_SODIUM -DDYNAMIC_SODIUM -DDYNAMIC_SODIUM_DLL=\"libsodium.dll\" /I "C:\libsodium\include" -DWINVER=0x0501 -D_WIN32_WINNT=0x0501 /source-charset:utf-8 /MP -DHAVE_STDINT_H /Ox /GL -DNDEBUG /arch:IA32 /Zl /MT /D_CRT_SECURE_NO_DEPRECATE /D_CRT_NONSTDC_NO_DEPRECATE -DFEAT_MBYTE_IME -DDYNAMIC_IME -DDYNAMIC_ICONV -DDYNAMIC_GETTEXT -DFEAT_LUA -DDYNAMIC_LUA -DDYNAMIC_LUA_DLL=\"lua54.dll\" -DFEAT_PYTHON -DDYNAMIC_PYTHON -DDYNAMIC_PYTHON_DLL=\"python27.dll\" -DFEAT_PYTHON3 -DDYNAMIC_PYTHON3 -DDYNAMIC_PYTHON3_DLL=\"python310.dll\" -DFEAT_MZSCHEME -I "C:\racket\include" -DMZ_PRECISE_GC -DDYNAMIC_MZSCHEME -DDYNAMIC_MZSCH_DLL=\"libracket3m_da32rk.dll\" -DDYNAMIC_MZGC_DLL=\"libracket3m_da32rk.dll\" -DFEAT_PERL -DPERL_IMPLICIT_CONTEXT -DPERL_IMPLICIT_SYS -DDYNAMIC_PERL -DDYNAMIC_PERL_DLL=\"perl532.dll\" -DFEAT_RUBY -DDYNAMIC_RUBY -DDYNAMIC_RUBY_DLL=\"msvcrt-ruby300.dll\" -DRUBY_VERSION=30 -DFEAT_HUGE /Fd.\ObjCULYHRZi386/ /Zi
	Linking: link /nologo /opt:ref /LTCG oldnames.lib kernel32.lib advapi32.lib shell32.lib gdi32.lib  comdlg32.lib ole32.lib netapi32.lib uuid.lib user32.lib  /machine:i386   libcmt.lib   /nodefaultlib:lua54.lib  /STACK:8388608  /nodefaultlib:python27.lib /nodefaultlib:python310.lib    winmm.lib WSock32.lib Ws2_32.lib   /PDB:vim.pdb -debug

### Logs and stack traces

```shell
N/A
```

k-takata:
Currently, Fixedsys is the default font on Windows.
It is not suitable for recent High DPI environments.

* Change the default font to Consolas.
* Allow to change the default font by the translation message. E.g.:
  ```
  msgid "DefaultFontNameForWindows"
  msgstr "Courier New"
  ```

Close #12919

### DIFF
--- a/src/testdir/test_mswin_event.vim
+++ b/src/testdir/test_mswin_event.vim
@@ -727,14 +727,14 @@ func Test_mswin_event_mouse()
   if has('gui_running')
     let args = { }
     let args.row = 9
-    let args.col = 7
+    let args.col = 5
     let args.move = 1
     let args.cell = 1
     call test_mswin_event("mouse", args)
     call feedkeys("\<Esc>", 'Lx!')
     let pos = getmousepos()
     call assert_equal(9, pos.screenrow)
-    call assert_equal(7, pos.screencol)
+    call assert_equal(5, pos.screencol)
 
     let args.cell = 0
     call test_mswin_event("mouse", args)
--- a/src/gui_w32.c
+++ b/src/gui_w32.c
@@ -392,16 +392,6 @@ static int (WINAPI *pGetSystemMetricsForDpi)(int, UINT) = NULL;
 static DPI_AWARENESS_CONTEXT (WINAPI *pSetThreadDpiAwarenessContext)(DPI_AWARENESS_CONTEXT dpiContext) = NULL;
 static DPI_AWARENESS (WINAPI *pGetAwarenessFromDpiAwarenessContext)(DPI_AWARENESS_CONTEXT) = NULL;
 
-    static UINT WINAPI
-stubGetDpiForSystem(void)
-{
-    HWND hwnd = GetDesktopWindow();
-    HDC hdc = GetWindowDC(hwnd);
-    UINT dpi = GetDeviceCaps(hdc, LOGPIXELSY);
-    ReleaseDC(hwnd, hdc);
-    return dpi;
-}
-
     static int WINAPI
 stubGetSystemMetricsForDpi(int nIndex, UINT dpi UNUSED)
 {
@@ -5286,7 +5276,7 @@ load_dpi_func(void)
 
 fail:
     // Disable PerMonitorV2 APIs.
-    pGetDpiForSystem = stubGetDpiForSystem;
+    pGetDpiForSystem = vimGetDpiForSystem;
     pGetDpiForWindow = NULL;
     pGetSystemMetricsForDpi = stubGetSystemMetricsForDpi;
     pSetThreadDpiAwarenessContext = NULL;
--- a/src/os_mswin.c
+++ b/src/os_mswin.c
@@ -2747,19 +2747,22 @@ quality_id2name(DWORD id)
     return qp->name;
 }
 
+// The default font height in 100% scaling (96dpi).
+// (-12 in 96dpi equates to roughly 9pt)
+#define DEFAULT_FONT_HEIGHT	(-12)
+
 static const LOGFONTW s_lfDefault =
 {
-    -12, 0, 0, 0, FW_NORMAL, FALSE, FALSE, FALSE, DEFAULT_CHARSET,
+    DEFAULT_FONT_HEIGHT,
+    0, 0, 0, FW_NORMAL, FALSE, FALSE, FALSE, DEFAULT_CHARSET,
     OUT_DEFAULT_PRECIS, CLIP_DEFAULT_PRECIS,
     PROOF_QUALITY, FIXED_PITCH | FF_DONTCARE,
-    L"Fixedsys"	// see _ReadVimIni
+    L""	// Default font name will be set later based on current language.
 };
 
-// Initialise the "current height" to -12 (same as s_lfDefault) just
-// in case the user specifies a font in "guifont" with no size before a font
-// with an explicit size has been set. This defaults the size to this value
-// (-12 equates to roughly 9pt).
-int current_font_height = -12;		// also used in gui_w32.c
+// This will be initialized when set_default_logfont() is called first time.
+// The value will be based on the system DPI.
+int current_font_height = 0;		// also used in gui_w32.c
 
 /*
  * Convert a string representing a point size into pixels. The string should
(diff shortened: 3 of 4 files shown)


src/testdir/test_mswin_event.vim
```diff
@@ -727,14 +727,14 @@ func Test_mswin_event_mouse()
...
-    let args.col = 7
+    let args.col = 5
     let args.move = 1
     let args.cell = 1
     call test_mswin_event("mouse", args)
     call feedkeys("\<Esc>", 'Lx!')
     let pos = getmousepos()
     call assert_equal(9, pos.screenrow)
-    call assert_equal(7, pos.screencol)
+    call assert_equal(5, pos.screencol)
```
k-takata, line 737: Without this change, the following error occurs:
https://github.com/k-takata/vim/actions/runs/6406241128/job/17390448472#step:11:9232
> ```
> 	From test_mswin_event.vim:
> 	Found errors in Test_mswin_event_mouse():
> 	command line..script D:/a/vim/vim/src/testdir/runtest.vim[601]..function RunTheTest[54]..Test_mswin_event_mouse line 56: Expected 1 but got 2
> ```

Not sure why.

k-takata:
> Is there a reason why we need to provide a localized default font?

I don't think about using Fixedsys at all.
I am considering using MS Gothic as the Japanese default font. It is the most widely used Japanese fixed-pitch vector font.
I think there are similar kinds of fonts for other CJK languages.
That's the reason why I'm going to provide a way to set a localized default font.

The problem might be that the ANSI part is not so beautiful compared to Consolas.

`set gfn=MS_Gothic`
image


> I guess one issue is that the substituted font is usually not a monospace font

Yes. The Japanese part is not shown well aligned if Consolas is used.

`set gfn=Consolas`
image

Using 'guifontwide' shows the ANSI part and the Japanese part very nicely. However, using 'guifontwide' has some pitfalls.
So, I don't want to make this the default.

`set gfn=Consolas gfw=MS_Gothic`
image

ychin:
Right. FWIW, part of the reason why I'm asking is that I'm changing how MacVim is rendering fonts and was looking at the GTK/Win32 code for reference.

I'm planning to change it so that in MacVim, if you do `set guifont=Consolas,MSGothic`, MSGothic will be chosen as a fallback font and automatically used when the first font doesn't support the Unicode value of the character you need (e.g. Japanese characters). Right now, the Vim fallback logic is just picking the first font in the list that loads, but it doesn't do per-character fallback. Doing this also kind of means `guifontwide` is unnecessary as you can just specify the entire fallback list you want in `guifont`. Only issue is you have to query the font to for every character you want to render to see if it's supported by the font you want, so I need to profile if there's a performance cost at all or whether caching is necessary.

Also, the existing behavior is also why people have to rely on patched fonts (e.g. Nerd Font) for Powerline etc because Vim can only load one font at a time. Terminals like Kitty actually already provide fallback font options so you can do what I just mentioned above and have Kitty automatically pick a font that supports the character you are typing. I think some terminals also support different Italic/Bold font as well (I think I may support that in MacVim as well using a special syntax in `guifont`, e.g. something like `set guifont=Menlo:h12,Monaco:sItalic`).

Sorry to get a little off topic, just wanted to share this before I work on it in case the Windows/GTK implementations of gVim may want something like that as well and want to share the syntax.

> Yes. The Japanese part is not shown well aligned if Consolas is used.

Regardless of default font choice, I kind of feel like this could be fixed up? I think the Win32 code mandates fixed-width font right now but I never knew if there's a particular technical reason or whether it's just because it looks bad otherwise. The issue with the mandate is that the OS is going to internally substitute with a non-fixed-width font anyway when they try to render CJK characters (as was the case when you were using Consolas). Is it something that could be fixed up in the font rendering code to make them align better? I haven't touched the Win32 font rendering code so I don't have a lot of contexts.

---

Also, just curious how are you generating that list of CJK / Hiragana / Katakana etc symbols? Is that a custom script / file you use?

k-takata:
> Also, just curious how are you generating that list of CJK / Hiragana / Katakana etc symbols?

`:dig!`
This is useful when checking the font rendering. (This doesn't include Kanji / Hanzi, though.)

src/testdir/test_mswin_event.vim
k-takata: The font size of Consolas is smaller than Fixedsys when the DPI is 100%.
So, this can happen.

PR #13270, opened 2023-10-04
Shift up error message, fixing numbering.

cvwillegen:
Because the internal error message for the positional arguments was added after a few class related error messages, and it was shifted up by 100, there was a 'hole' at the end of the errors.h file. This shifts the error message inside the 'hole'.


chrisbra:
can you please add a tag for the help file?

zeertzjq:
Shouldn't E1520 tag also be removed now?

cvwillegen:
Op do 5 okt. 2023 08:30 schreef zeertzjq ***@***.***>:

> Shouldn't E1520 tag also be removed now?
>

Yes, it's in a weird place, thanks for noticing! I didn't expect that it
was already there...

I'll take it out.

Christ van Willegen

>

PR #13271, opened 2023-10-04
Add just filetype

dundargoc:


chrisbra:
thanks!

runtime/filetype.vim
```diff
@@ -1093,6 +1093,9 @@ au BufNewFile,BufRead *.jsonnet,*.libsonnet	setf jsonnet
 " Julia
 au BufNewFile,BufRead *.jl			setf julia
 
+" Just
+autocmd BufNewFile,BufRead *[jJ]ustfile,.justfile,*.just setfiletype just
```
zeertzjq, line 1097: The autocommand in https://github.com/NoahTheDuke/vim-just/blob/main/ftdetect/just.vim doesn't have a `*` before the `\cjustfile`, so having a `*` before `[jJ]ustfile` looks strange. Also it should be `au` and `setf` here for consistency, not `autocmd` and `setfiletype`.

PR #13272, opened 2023-10-04
Use defining class in variable access error messages.

errael:
In vim9class.c:
```
void emsg_var_cl_define(char *msg, char_u *name, size_t len, class_T *cl) 
static class_T * class_defining_member(class_T *cl, char_u *name, size_t len, 
                                       vartype_T *p_typ, ocmember_T **p_m)
```
used as in
```
emsg_var_cl_define(e_cannot_access_private_variable_str, m->ocm_name, 0, cl);
```
Related uses, for example to get the class_T* of the defining class, things can be made public

### DIFF
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -2172,8 +2172,8 @@ get_member_tv(
 
     if (*name == '_')
     {
-	semsg(_(e_cannot_access_private_variable_str), m->ocm_name,
-							    cl->class_name);
+	emsg_var_cl_define(e_cannot_access_private_variable_str,
+							m->ocm_name, 0, cl);
 	return FAIL;
     }
 
@@ -2332,8 +2332,8 @@ class_object_index(
 
 	if (*name == '_')
 	{
-	    semsg(_(e_cannot_access_private_variable_str), m->ocm_name,
-							    cl->class_name);
+	    emsg_var_cl_define(e_cannot_access_private_variable_str,
+							m->ocm_name, 0, cl);
 	    return FAIL;
 	}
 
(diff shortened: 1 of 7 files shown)


errael:
@yegappan As suggested, use the defining class; see `def Test_defining_class_message()`.
It surprised me that after the change, there were no errors, thus the new tests.

src/vim9class.c
```diff
@@ -2583,6 +2583,63 @@ member_lookup(
...
+ * Return the class that defines the member "name", else NULL.
+ * Fill in "p_typ" if specified, for found class.
+ * "p_typ" is either VAR_OBJECT or VAR_CLASS; VAR_UNKNOWN if not found
+ */
+
+// TODO: if requested could indirectly return ocmember_T and idx.
+
+    static class_T *
+class_defining_member(class_T *cl, char_u *name, size_t len,
+					    vartype_T *p_typ, ocmember_T **p_m)
```
yegappan: It looks like `p_typ` is always NULL? Is this parameter needed?

errael: Not really; but it's the only way for a caller to determine if the name is object or class. The function is static, so no reason to provide more features than needed.

Shall I take it out and leave a not about how it can be extended?

BTW, I was just looking and I think I can cover 2632. It's a path that should be tested. Shouldn't need to change any code, just the test. (famous last words)

errael:
Remove unused arguments. Couple more tests, increased coverage.

src/vim9class.c
```diff
@@ -2856,6 +2907,19 @@ object_free_nonref(int copyID)
...
+/*
+ * Output message which takes a variable name and the class that defines it.
+ * "cl" is that class where the name was found. Search "cl"'s hierarchy to
+ * find the defining class.
+ */
+    void
+emsg_var_cl_define(char *msg, char_u *name, size_t len, class_T *cl)
+{
+    ocmember_T	*m;
+    class_T	*cl_def = class_defining_member(cl, name, len, &m);
```
yegappan, line 2919: The class_defining_member() function may return a NULL pointer? Do we need a check here cl_def == NULL?

errael: It is only null if the class/object does not have a member by the  name. Considering the first sentence of the comment

> Output message which takes a variable name and the class that defines it.

I don't see how it could be null.

Could have it return OK/FAIL (instead of void). Do nothing if NULL and return FAIL (if you're in a defensive mood). Which today would be ignored by all callers. LMK.

errael: One option would be to report E340, it is a better choice than SEGV.

yegappan: Using the E340 message sounds good to me.

errael: > Using the E340 message sounds good to me.

Done.

chrisbra:
thanks!

Issue #13259, opened 2023-10-03
Inconsistency between version names in documentation and current status


RestorerZ:
### Steps to reproduce

If I understand correctly, based on this PR (#11268 ) and this PR (#11283 ) and the [definition](https://github.com/vim/vim/blob/b74ebfc6bf343ed825b111157261a8d99da6d670/src/version.c#L4979) in `version.c`, there is a clear discrepancy with the specified version name in the documentation. Section [*version*](https://github.com/vim/vim/blob/6ee7b521fa7531ef356ececc8be7575c3800f872/runtime/doc/various.txt#L309) and [*+feature-list*](https://github.com/vim/vim/blob/6ee7b521fa7531ef356ececc8be7575c3800f872/runtime/doc/various.txt#L333)

>The first column shows the smallest version in which they are included:
                           T    tiny (always)
                           S    small
                           N    normal
                           B    big
                           H    huge

### Expected behaviour

The features correspond to the current state of the versions.

### Version of Vim

Vim 9.0.1976

### Environment

All

chrisbra:
I don't understand. Where exactly do you see a discrepancy? Or is this about the *feature-list*? (there used to be small and big builds, but those are gone nowadays I believe).

RestorerZ:
@chrisbra 
I mean the description in the `version` section.
As I understand it, it should be something like this:
```
The first column shows the smallest version in which they are included:
			   T	tiny (always)
			   N	normal
			   H	huge
			   m	manually enabled or depends on other features
			   -    never, feature was removed
			 (none) system dependent
Thus if a feature is marked with "N", it is included in the normal and huge versions of Vim.
```
The `feature-list` section is fine now, sorry.

RestorerZ:
I'll do a PR later tonight.
There are some more documentation files that also mention `small` or `big` versions (starting.txt, testing.txt, usr_90.txt).

PR #13273, opened 2023-10-04
consistency between version names and current status

RestorerZ:
#13259 
consistency between version names and current status

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,7 +29,6 @@ environment:
 # disabled
 #    - FEATURE: TINY
 #    - FEATURE: NORMAL
-#    - FEATURE: BIG
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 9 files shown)


chrisbra:
Hm, not so sure about Amiga. It seems the latest Amiga binary version is for [Vim64](https://ftp.nluug.nl/pub/vim/amiga/vim64bin.tgz) for which a *big* built was still available back then.

RestorerZ:
@chrisbra 
Regarding the Amiga, I was going by what is stated [here](https://github.com/vim/vim/blob/290b887e8cc2c0d3dfc7f315b2052472c7c589cc/src/Make_ami.mak#L9).
The default build is set to 'huge'.
```
#
# Makefile for AROS, AmigaOS4 and MorphOS.
#
BIN = vim
CC ?= gcc
LD = $(CC)
UNM ?= $(shell uname)
DEBUG ?= no
BUILD ?= huge
CFLAGS = -c -O3
```

RestorerZ:
@chrisbra 
I see. Thanks for the clarification. Will do.

PR #13277, opened 2023-10-05
translation(it): revert permission changes for xxd manpages again

k-takata:
The permission of xxd manpages was changed to 755 unintentionally in https://github.com/vim/vim/commit/b9a974df9e22c7b56d8faac3246fb5f80a9351d4.


k-takata:
Cc: @azc100, @chrisbra

chrisbra:
ah, sorry, I usually just extract the sent over translations and commit them. I'll need to make sure the permissions are sane. Thanks for catching it!

PR #13278, opened 2023-10-05
Avoid two messages for same error.

errael:
It's my understanding that only the first message should be output. I've seen this pattern used for this PR elsewhere.
```
vim9script

class A
    this._priv: string
endclass

var o = A.new()
echo o._priv
```
```
E1333: Cannot access private variable "_priv" in class "A"
E1326: Variable not found on object "A": _priv
```


yegappan (approved):

chrisbra:
thanks!

PR #13279, opened 2023-10-05
Access to freed memory when trying to use class member type

yegappan:
Use the type list grow array associated with the class than using a temporary type list grow array
to allocate the class member type.

For simple types, a predefined type is used.  For complex types, the type is dynamically
allocated from a grow array.   For class variables, the type grow array in the class should
be used.  So that the lifetime of the type is same as the lifetime of the class.


chrisbra:
thanks!

PR #13280, opened 2023-10-05
win32: Improve Per-Monitor V2

k-takata:
Related: #11812, #13252

Support calculating the new size even if a bitmap font is used.
Calculate the new font size before actually change the Window size.


chrisbra:
thanks!

PR #13286, opened 2023-10-05
Enable xxd -r -b to revert a bits dump

tristhaus:
Add code, testcases, partial documentation to revert a bits dump by calling `xxd -r -b`.


chrisbra:
thank you! I also incremented the version output.

PR #13288, opened 2023-10-05
remove unnecessary code

dundargoc:


dundargoc:
Someone explain to me how code coverage *decreases* after I delete code?

chrisbra:
thanks, I include it. Next time, please check the style

dundargoc:
Could you give me some pointers on what specifically? I skimmed through `develop.txt` but I couldn't find anything that related to the changes in this PR.

chrisbra:
In this specific case it's just that we don't need braces around single statements after the if condition. Not sure if this is documented anywhere.

ychin:
FWIW I think the removed `col = col % width2;` was actually an useful line. It helps to make sure `col` is calculated correctly so if it's used afterwards (e.g. someone adds some lines after this condition in the future) they wouldn't accidentally use `col` as-is which would cause a buffer overflow. It's about defending against future mistakes even if it's unused right now. The actual calculation would be optimized out right now anymore since the compiler can detect that `col` is not used after this.

dundargoc:
@ychin I'd prefer to not get involved at the moment. Please make a PR reverting the necessary lines if you'd like.

chrisbra:
thanks, I have partially reverted it and left a comment

PR #13290, opened 2023-10-06
if_python: Fix python312

### LINKED ISSUE #13281: if_python3.c(499) warning C4273 py3_PyBool_Type
Hello

When Vim 9.0.1984 is compiled with msvc C/C++ version 19.29.30140, there is a warning: 
```
if_python3.c(499): warning C4273 py3_PyBool_Type : inconsistent dll linkage
Python310\Include\boolobject.h(10): see previous definition of 'py3_PyBool_Type'

```

k-takata:
Close #13281

Py_SIZE() uses PyLong_Type and PyBool_Type starting from Python 3.12. We need to define our own Py_SIZE() to replace Py{Bool,Long}_Type with py3_Py{Bool,Long}_Type.
We also need to redefine PyTuple_GET_SIZE() and PyList_GET_SIZE(), because they use Py_SIZE().

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -68,8 +68,6 @@
 #endif
 
 #define PY_SSIZE_T_CLEAN
-#define PyLong_Type (*py3_PyLong_Type)
-#define PyBool_Type (*py3_PyBool_Type)
 
 #ifdef Py_LIMITED_API
 # define USE_LIMITED_API // Using Python 3 limited ABI
@@ -297,6 +295,10 @@ static HINSTANCE hinstPy3 = 0; // Instance of python.dll
 # define PyFloat_Type (*py3_PyFloat_Type)
 # define PyNumber_Check (*py3_PyNumber_Check)
 # define PyNumber_Long (*py3_PyNumber_Long)
+# define PyBool_Type (*py3_PyBool_Type)
+# if PY_VERSION_HEX >= 0x030c00b0
+#  define PyLong_Type (*py3_PyLong_Type)
+# endif
 # define PyErr_NewException py3_PyErr_NewException
 # ifdef Py_DEBUG
 #  define _Py_NegativeRefcount py3__Py_NegativeRefcount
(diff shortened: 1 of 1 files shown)


k-takata:
I saw a report that using Python 3.12 on macOS failed with the following error:

```
ld: Undefined symbols:
  _py3_PyBool_Type, referenced from:
      _DictionaryUpdate in if_python3.o
      _PyList_GET_SIZE in if_python3.o
      _PyTuple_GET_SIZE in if_python3.o
  _py3_PyLong_Type, referenced from:
      _DictionaryUpdate in if_python3.o
      _PyList_GET_SIZE in if_python3.o
      _PyTuple_GET_SIZE in if_python3.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [vim] Error 1
make: *** [first] Error 2
```
This PR also fix this.

ychin:
Just for reference this is where the stable ABI path remaps the macros to just call the functions themselves:

https://github.com/vim/vim/blob/ac9c6d5c786a016ca1dce297d167a5a963881f7e/src/if_py_both.h#L153-L156

k-takata:
Actually, I'm wondering if we need to support non-stable-ABI implementation for Python 3.8 or later.

ychin:
Yeah this is a good question. When I did the PR I kept backwards compatibility because it made sense to have a way to revert back to the old behavior, and it's quite useful for testing. It probably also makes sense to not use if you aren't using a dynamic build since you already know what lib you are linking against but even then it's not that big of a difference.

Technically, the non-stable ABI is a little faster (e.g. the Py_SIZE macro in question here), and the object types are structs defined in the header instead of opaque types that you have to allocate in heap through pre-defined APIs. I'm not sure if Vim's Python bindings are doing so much that it actually matters though. If you set the minimum compatible Python version (Py_LIMITED_API) to be higher (3.11) it will take advantage of more recent APIs as well.

ychin:
Ok. I still stand by what I suggested in the above comment though. I think it's just much simpler to bind the `PyTuple_GET_SIZE` macro to a `PyTuple_Size` function call and avoid having to copy-and-paste from some Python headers that may change in the future. I may do another PR and discuss it there.

PR #13293, opened 2023-10-06
Invalid memory access when 'foldexpr' returns empty string

zeertzjq:
Problem:  Invalid memory access when 'foldexpr' returns empty string.
Solution: Check for NUL.


chrisbra:
thanks!

PR #13294, opened 2023-10-06
Elaborate guideline

dundargoc:


runtime/doc/develop.txt
```diff
@@ -386,6 +385,25 @@ OK:	    if (cond)
                cmd;
 	    }
 
+Wrong:	    if (cond) {
```
k-takata:
```suggestion
Wrong:	    if (cond)
	    {
```

runtime/doc/develop.txt
```diff
@@ -386,6 +385,25 @@ OK:	    if (cond)
...
+
+Wrong:	    if (cond)
+               cmd;
+	    else
+	    {
+               cmd;
+               cmd;
+	    }
+
+Wrong:	    if (cond) {
```
k-takata:
```suggestion
Wrong:	    if (cond)
	    {
```

errael:
I recently modified a long chain that was like
```
if (cond)
    stmt;
else if (cond)
    stmt;
...
else
    stmt;
```
by adding
```
else if (cond)
{
    stmt;
    stmt;
}
```
in the middle. Is the expectation that "all blocks must have braces" after the change?

What's wrong with "usually looks better" as a guideline?

IIUC, this is the only change in this PR (other than wanting to pull up `{` :)

zeertzjq:
The last two "Wrong:" examples in this PR appear in a lot of places.

yegappan:
The convention used is if a block has only one statement then braces are not used.

Shane-XB-Qian:
perhaps you can just fill up this collection to https://github.com/vim/vim/discussions/13087
// vs a PR
// and looks there was a general style check already

-- 
shane.xb.qian

dundargoc:
> Is the expectation that "all blocks must have braces" after the change?

I guess? That is the implication, I've just made it explicit.

> What's wrong with "usually looks better" as a guideline?

I prefer my style guides to just tell me what it wants rather than gaslight me
into thinking what it wants looks good.

dundargoc:
> The last two "Wrong:" examples in this PR appear in a lot of places.

If the vim project chooses not to follow its own style guide then that is up to
them. I've just made the style guide explicit for new contributors.

Shane-XB-Qian:
> What's wrong with "usually looks better" as a guideline?

i donot understand what's wrong here too :-)
it was just saying 'better', why this PR this modification required?

> >    The last two "Wrong:" examples in this PR appear in a lot of places.
> If the vim project chooses not to follow its own style guide then that is up to them. I've just made the style guide explicit or new contributors.

there were many others committed code into vim codebase, and seems it was no good style check/review,
so/but perhaps let's just follow that develop style now, and rephrase old existed later when found :smile:

-- 
shane.xb.qian

Shane-XB-Qian:
> The convention used is if a block has only one statement then braces are not used.

i guess this is what @dundargoc was arguing, as per guideline, your/this statement was not always true:

```txt
When a block has one line the braces can be left out.  When an if/else has
braces on one block, it usually looks better when the other block also has
braces:
OK:	    if (cond)
	       cmd;
	    else
               cmd;

OK:	    if (cond)
	    {
	       cmd;
	    }
	    else
	    {
               cmd;
               cmd;
	    }
```

but this just a 'usually looks better', seems just a recommendation, not a must. as per my terrible non-native english. :rofl:

ychin:
FWIW, I think it's useful to have code formatters and linters that will tell us explicitly that a particular code style is wrong. "Usually looks better" has proven historically to not work out well as a project scales because everyone's definition of that is different, and coding style never has a clearcut right-or-wrong answer. As a contributor myself I find myself struggling with exactly how to indent my multi-line if statements, etc, and I wish there's a clearly established formatter / linter that would just tell me.

I think one issue with this PR is that we can see a lot of single-line if statements not wrapped in braces when used in multi-condition if blocks, so clearly saying it's a "NO" may not be accurate to the established style. The coding style seems to be more a "both are fine" camp.

I would personally advocate just adding a clang formatter/linter rule to the repo and ask new contributors to use said formatter to format their code. The rule should work on most of the existing codebase but obviously not all of them, and we don't need to fix it right now, but new code can just use it. If a full formatter is too strict (since it completely removes flexibility which is sometimes needed) we could at least have tools that can detect clear style violations that are not allowed (e.g. lines that are too long, single-line if statement that use braces). It would also help catch indentation issues since Vim has a weird mixed-tab/spaces method of indentation that is easy to mess up if someones uses say VS Code (blasphemy) to touch the code.

I know we added test_codestyles written in Vim script but it's better for this to be done by a compiler like clang which is designed to compile code.

Shane-XB-Qian:
> I think one issue with this PR is that we can see a lot of single-line if statements not wrapped in braces when used in multi-condition if blocks, so clearly saying it's a "NO" may not be accurate to the established style.

'usually looks better' just a recommendation, 
and in practice above answers from @yegappan and others said `if a block has only one statement then braces are not used`, so to `braces when used in multi-condition` supposed that recommendation was just a *second* choice.

perhaps that's clear enough, some linter seems worked that way too (was not so 0/1 strict).

chrisbra:
yeah, I agree. A code formatter/linter would be nice. I don't know how easily it would be to configure it for the existing style (and I'd rather avoid lengthy discussions about style changes in case it cannot be configured for the current style), but if anybody has some experience on this and wants to prepare a PR for this, this would be greatly appreciated. 

Also not sure, if we can make use of editorconfig (or if this even makes sense for the Vim project)

PR #13295, opened 2023-10-06
Fix builtin `islocked()` SEGV when arg is class/object variable.

errael:
Now f_islocked() handles lval_T for object/class variable.

SEGV example
```
vim9script

class C1
  this.o0: list<list<number>> = [ [0],  [1],  [2]]
endclass
var obj = C1.new()
islocked("obj.o0")
```

Probably follow on PR for `islocked()`, but wanted to get the SEGV out of the way.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -1375,6 +1375,7 @@ get_lval(
 		      && v_type != VAR_OBJECT
 		      && v_type != VAR_CLASS)
 	{
+	    // TODO: have a message with obj/class, not just dict,
 	    if (!quiet)
 		semsg(_(e_dot_can_only_be_used_on_dictionary_str), name);
 	    return NULL;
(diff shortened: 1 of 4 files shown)


errael:
Note the two TODO in eval.c. Two old error messages are not accurate since vim9class.

Since they are more syntax error and unlikely to be caught by msg in a user program, could
1. Fix the text and keep the number.
2. Introduce a new corrected error and start using that.
   Remove the old error if it's not used elsewhere.
3. No change and give an inaccurate error message.

Preferences?

chrisbra:
If the error codes are used elsewhere, I think number 2 should be done. If not,we can go with number 1.

errael:
> If the error codes are used elsewhere, I think number 2 should be done. If not,we can go with number 1.

Noted. I'll check it out and open a new PR.

BTW, that has nothing to do with this PR (which only modifies code in `f_islocked()` to avoid a SEGV).

chrisbra:
thanks

PR #13301, opened 2023-10-07
remove redundant code

dundargoc:
x % 1 = 0

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -7470,7 +7470,7 @@ set_context_in_set_cmd(
 		while (s > xp->xp_pattern && *(s - 1) == '\\')
 		    --s;
 		if ((*p == ' ' && (xp->xp_backslash == XP_BS_THREE && (p - s) < 3))
-			|| (*p == ',' && (flags & P_COMMA) && ((p - s) % 1) == 0)
+			|| (*p == ',' && (flags & P_COMMA))
 			|| (*p == ':' && (flags & P_COLON)))
 		{
 		    xp->xp_pattern = p + 1;


src/option.c
```diff
@@ -7470,7 +7470,7 @@ set_context_in_set_cmd(
 		while (s > xp->xp_pattern && *(s - 1) == '\\')
 		    --s;
 		if ((*p == ' ' && (xp->xp_backslash == XP_BS_THREE && (p - s) < 3))
-			|| (*p == ',' && (flags & P_COMMA) && ((p - s) % 1) == 0)
+			|| (*p == ',' && (flags & P_COMMA))
```
errael, line 7473: what if `(p -s )` is negative?

dundargoc: 👀 is that allowed?

dundargoc: Is it still not zero?

errael: I'm not sure. (but probably)

chrisbra:
I suppose this should have been `((p - s) & 1) == 0` like several lines above https://github.com/vim/vim/blob/9771b2a67f825bdc6e5c615141d22c665952dc86/src/option.c#L7271 to account for backslash escaped commas? @ychin ?

ychin:
This line exists for a reason, albeit it's buggy and doesn't do anything… It was pointed out in https://github.com/vim/vim/pull/13182#discussion_r1342059019 too.

I was working on a PR to fix it and it's basically working but haven't sent it out yet, so I just made it right now: #13303. The correct logic is `(p-s) >= 1` in Windows and `(p-s) >= 2` in other platforms.

The PR has more explanation, but basically when you are doing cmdline expansion for a comma-separated list, let's say you have `:set backupdir=foo,b<Tab>`, then "b" is considered a new file name to be completed. If you have `:set backupdir=foo\\,b<Tab>` then what you have is `foo,b` as a whole is being considered a file name to be completed.

ychin:
Sorry I really meant the flipped logic. It should be `(p-s) < 1` and `(p-s) < 2`.

src/option.c
ychin: I pointed out what the logic is doing elsewhere already but either way `p-s` cannot be negative here. `s` is counted down from `p` in the above line.

ychin:
So for example, `set backupdir=foo\\\,bar<Tab>` would expand file to the following pattern: `foo\,bar`.

`foo\\\\,bar<Tab>` => `foo\\,bar`
`foo\\\\\,bar<Tab>` => `foo\\,bar`
`foo\\\\\\,bar<Tab>` => `foo\\\,bar`
`foo\\\\\\\,bar<Tab>` => `foo\\\,bar`

That's why it's an inequality check just like the `(p-3) <3` above. As long as you have enough backslashes you know the comma is not delimiting a new file.

You need one fewer backslash in Windows because in Windows, patterns like `\,` does not escape to `,` for convenient since there are a lot of backslashes in Windows file names and I guess Vim doesn't want them to have to type `\\` everywhere.

dundargoc:
OK, looks like this will be correctly handled in https://github.com/vim/vim/pull/13303 then.

ychin:
Just a minor question: What's the rationale for this PR? Was it triggered by some static analyzers? I think we should be a little careful deleting "unused logic" like this because they usually contain some intention behind them, even if buggy. I think if it's blatantly useless like this, we should try to make sure to understand the intention behind them first as blanket deletion makes debugging harder later as you lose the context. I do admit understanding the full intention could be a little time consuming as it requires essentially unwinding the stack, but I think otherwise it's better to leave them be.

dundargoc:
I understand. I'll try to not to send these types of fixes anymore then. @zeertzjq for reference ☝️

errael:
> not to send these types of fixes

When something like this is encountered, as @ychin states
> we should try to make sure to understand the intention behind them first 

A PR with a comment like
```
// TODO: This looks like a bug
```
might be good.

ychin:
Sorry, I wasn't trying to discourage such effort. I'm just pointing out that it could be worth tagging these as potential bugs (as was in this case) since they tend to reflect a mismatch of expectations of a programmer and what the code is actually doing. So it's definitely useful to detect these. I'm just advising against deleting them unless we have a fair idea why it was written to begin with.

Shane-XB-Qian:
> So it's definitely useful to detect these. I'm just advising against deleting them unless we have a fair idea why it was written to begin with.

// the author of that piece code put a comment at there about why there maybe better? :smile:

Shane-XB-Qian:
> Just a minor question: What's the rationale for this PR? Was it triggered by some static analyzers?

let me guess: was by gpt or copilot or something like that? :smile:
// per your answer, i think human can fight and win them... ahaha.....

-- 
shane.xb.qian

PR #13302, opened 2023-10-08
Remove the unused class variable and class method related code for interfaces

yegappan:
Remove the unused class variable and class method related code for interfaces.  Refactor the code.  Optimize the object/class member double lookup in the compile_lhs() function.  Change unused global functions to static functions.


chrisbra:
thanks

PR #13303, opened 2023-10-08
Fix command-line expansions for options with filenames with commas

ychin:
Cmdline expansion for option values that take a comma-separated list of file names is currently not handling file names with commas as the commas are not escaped. For such options, the commas in file names need to be escaped (to differentiate from a comma that delimit the list items). The escaped comma is unescaped in `copy_option_part()` during option parsing.

Fix as follows:
- Cmdline completion for option values with comma-separated file/folder names will not start a new match when seeing `\\,` and will instead consider it as one value.
- File/folder regex matching will strip the `\\` when seeing `\\,` to make sure it can match the correct files/folders.
- The expanded value will escape `,` with `\\,`, similar to how spaces are escaped to make sure the option value is correct on the cmdline.

This fix also takes into account the fact that Win32 Vim handles file name escaping differently. Typing `\,` for a file name results in it being handled literally but in other platforms `\,` is interpreted as a simple ',' and commas need to be escaped using `\\,` instead.

Also, make sure this new logic only applies to comma-separated options like 'path'. Non-list options like 'set makeprg=<Tab>' and regular ex commands like `:edit <Tab>` do not require escaping and will continue to work.

Also fix up documentation to be clearer. The original docs are slightly misleading in how it discusses triple slashes for 'tags'.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -192,10 +192,27 @@ To include white space in a string option value it has to be preceded with a
 backslash.  To include a backslash you have to use two.  Effectively this
 means that the number of backslashes in an option value is halved (rounded
 down).
+In options 'path', 'cdpath', and 'tags', spaces have to be preceded with three
+backslashes instead for compatibility with version 3.0 where the options can
+be separated by either commas or spaces.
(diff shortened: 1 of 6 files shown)


chrisbra:
thanks!

PR #13307, opened 2023-10-09
Enforce invariant type check for arguments and return value when assigning a funcref

### LINKED ISSUE #13299: Covariant parameter types allowed when assigning functions
### Steps to reproduce

```vim
vim9script

class A
endclass
class B extends A
endclass

def Foo(p: B): B
  return B.new()
enddef

var Bar: func(A): A = Foo
```

### Expected behaviour

An error message should be output:
`"E1012: Type mismatch; expected func(object<A>): object<A> but got func(object<B>): object<B>"`

I'm a bit surprised these checks are handled separately.

### Version of Vim

9.0.1999

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Fixes #13299.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -7154,4 +7154,41 @@ def Test_recursive_class_method_call()
   v9.CheckSourceSuccess(lines)
 enddef
 
+" Test for checking the argument types and the return type when assigning a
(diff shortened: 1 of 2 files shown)


chrisbra:
included as of e4671890220ef3f2bca43fde6ffe5d3ef3ed0e42

PR #13310, opened 2023-10-10
if_x: Avoid using #ifndef - #else - #endif

k-takata:
Using #ifndef - #else - #endif is sometimes confusing. Use #ifdef - #else - #endif instead.


chrisbra:
thanks

PR #13312, opened 2023-10-10
Fix message for some single use error messages.

errael:
As discussed around https://github.com/vim/vim/pull/13295#issuecomment-1751769772


yegappan (approved):

chrisbra:
thanks!

PR #13315, opened 2023-10-11
win32,python3: Load DLL from registry path

k-takata:
Support loading `python3.dll` and/or `python3xx.dll` from the path written in the registry.  To support Stable ABI's forwarder DLL (`python3.dll`), use the `LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR` flag for `LoadLibraryEx()`.

If Stable ABI is used, search the latest version from the registry (both from `HKEY_CURRENT_USER` and `HKEY_LOCAL_MACHINE`).  If Stable ABI is not used, search only the matching version.


chrisbra:
thanks!

PR #13316, opened 2023-10-11
Recognize .INI files as dosini

arp242:
It previo\~1 only worked for lower-case .ini files, but upperc\~1 .INI is also somewhat common on account of DOS' old 8.3 upperc\~2 only filena\~1.


chrisbra:
Thank you, it will be includ~1 in patch v9.0.2011

PR #13318, opened 2023-10-11
Add formatter

dundargoc:
Uncrustify is an extremly configurable formatter which makes it ideal
for introducing it for established projects with peculiar formatting
quirks.

### DIFF
--- a/src/sound.c
+++ b/src/sound.c
@@ -1,6 +1,6 @@
 /* vi:set ts=8 sts=4 sw=4 noet:
  *
- * VIM - Vi IMproved	by Bram Moolenaar
+ * VIM - Vi IMproved    by Bram Moolenaar
  *
  * Do ":help uganda"  in Vim to read copying and usage conditions.
  * Do ":help credits" in Vim to see a list of people who contributed.
@@ -15,32 +15,32 @@
 
 #if defined(FEAT_SOUND) || defined(PROTO)
 
-static long	    sound_id = 0;
+static long sound_id = 0;
 
 // soundcb_T is typedef'ed in proto/sound.pro
 
 struct soundcb_S {
     callback_T	snd_callback;
-#ifdef MSWIN
+# ifdef MSWIN
     MCIDEVICEID	snd_device_id;
     long	snd_id;
-#endif
+# endif
     soundcb_T	*snd_next;
 };
 
-static soundcb_T    *first_callback = NULL;
+static soundcb_T *first_callback = NULL;
 
 /*
  * Return TRUE when a sound callback has been created, it may be invoked when
  * the sound finishes playing.  Also see has_sound_callback_in_queue().
  */
-    int
+int
 has_any_sound_callback(void)
 {
     return first_callback != NULL;
 }
 
-    static soundcb_T *
+static soundcb_T *
 get_sound_callback(typval_T *arg)
 {
     callback_T	callback;
@@ -70,8 +70,11 @@ get_sound_callback(typval_T *arg)
 /*
  * Call "soundcb" with proper parameters.
  */
-    void
-call_sound_callback(soundcb_T *soundcb, long snd_id, int result)
+void
+call_sound_callback(soundcb_T	*soundcb,
+		    long	snd_id,
+		    int		result
+		    )
 {
     typval_T	argv[3];
     typval_T	rettv;
@@ -89,7 +92,7 @@ call_sound_callback(soundcb_T *soundcb, long snd_id, int result)
 /*
  * Delete "soundcb" from the list of pending callbacks.
  */
-    void
+void
 delete_sound_callback(soundcb_T *soundcb)
 {
     soundcb_T	*p;
@@ -108,23 +111,23 @@ delete_sound_callback(soundcb_T *soundcb)
 	}
 }
 
-#if defined(HAVE_CANBERRA) || defined(PROTO)
+# if defined(HAVE_CANBERRA) || defined(PROTO)
 
 /*
  * Sound implementation for Linux/Unix using libcanberra.
  */
-# include <canberra.h>
+#  include <canberra.h>
 
-static ca_context   *context = NULL;
+static ca_context *context = NULL;
 
 // Structure to store info about a sound callback to be invoked soon.
 typedef struct soundcb_queue_S soundcb_queue_T;
 
 struct soundcb_queue_S {
     soundcb_queue_T	*scb_next;
-    uint32_t		scb_id;		// ID of the sound
-    int			scb_result;	// CA_ value
-    soundcb_T		*scb_callback;	// function to call
+    uint32_t		scb_id;         // ID of the sound
+    int			scb_result;     // CA_ value
+    soundcb_T		*scb_callback;  // function to call
 };
 
 // Queue of callbacks to invoke from the main loop.
@@ -135,15 +138,15 @@ static soundcb_queue_T *callback_queue = NULL;
  * That is because the callback may be called from another thread and invoking
  * another sound function may cause trouble.
  */
-    static void
-sound_callback(
-	ca_context  *c UNUSED,
-	uint32_t    id,
-	int	    error_code,
-	void	    *userdata)
+static void
+sound_callback(ca_context *c	UNUSED,
+	       uint32_t		id,
+	       int		error_code,
+	       void		*userdata
+	       )
 {
-    soundcb_T	    *soundcb = (soundcb_T *)userdata;
-    soundcb_queue_T *scb;
+    soundcb_T		*soundcb = (soundcb_T *)userdata;
+    soundcb_queue_T	*scb;
 
     scb = ALLOC_ONE(soundcb_queue_T);
     if (scb == NULL)
@@ -153,7 +156,7 @@ sound_callback(
     scb->scb_id = id;
     scb->scb_result = error_code == CA_SUCCESS ? 0
 			  : error_code == CA_ERROR_CANCELED
-					    || error_code == CA_ERROR_DESTROYED
+		      || error_code == CA_ERROR_DESTROYED
 			  ? 1 : 2;
     scb->scb_callback = soundcb;
 }
@@ -161,7 +164,7 @@ sound_callback(
 /*
  * Return TRUE if there is a sound callback to be called.
  */
-    int
+int
 has_sound_callback_in_queue(void)
 {
     return callback_queue != NULL;
@@ -170,7 +173,7 @@ has_sound_callback_in_queue(void)
 /*
  * Invoke queued sound callbacks.
  */
-    void
+void
 invoke_sound_callback(void)
 {
     soundcb_queue_T *scb;
@@ -188,8 +191,11 @@ invoke_sound_callback(void)
     redraw_after_callback(TRUE, FALSE);
 }
 
-    static void
-sound_play_common(typval_T *argvars, typval_T *rettv, int playfile)
+static void
+sound_play_common(typval_T	*argvars,
+		  typval_T	*rettv,
+		  int		playfile
+		  )
 {
     if (in_vim9script() && check_for_string_arg(argvars, 0) == FAIL)
 	return;
@@ -206,10 +212,11 @@ sound_play_common(typval_T *argvars, typval_T *rettv, int playfile)
     if (soundcb == NULL)
     {
 	res = ca_context_play(context, sound_id,
-		playfile ? CA_PROP_MEDIA_FILENAME : CA_PROP_EVENT_ID,
-		tv_get_string(&argvars[0]),
-		CA_PROP_CANBERRA_CACHE_CONTROL, "volatile",
-		NULL);
+			      playfile ? CA_PROP_MEDIA_FILENAME :
+			      CA_PROP_EVENT_ID,
+			      tv_get_string(&argvars[0]),
+			      CA_PROP_CANBERRA_CACHE_CONTROL, "volatile",
+			      NULL);
     }
     else
     {
@@ -220,14 +227,14 @@ sound_play_common(typval_T *argvars, typval_T *rettv, int playfile)
 	{
 	    if (playfile)
 		ca_proplist_sets(proplist, CA_PROP_MEDIA_FILENAME,
-			(char *)tv_get_string(&argvars[0]));
+				 (char *)tv_get_string(&argvars[0]));
 	    else
 		ca_proplist_sets(proplist, CA_PROP_EVENT_ID,
-			(char *)tv_get_string(&argvars[0]));
+				 (char *)tv_get_string(&argvars[0]));
 	    ca_proplist_sets(proplist, CA_PROP_CANBERRA_CACHE_CONTROL,
-		    "volatile");
+			     "volatile");
 	    res = ca_context_play_full(context, sound_id, proplist,
-		    sound_callback, soundcb);
+				       sound_callback, soundcb);
 	    if (res != CA_SUCCESS)
 		delete_sound_callback(soundcb);
 
@@ -237,8 +244,10 @@ sound_play_common(typval_T *argvars, typval_T *rettv, int playfile)
     rettv->vval.v_number = res == CA_SUCCESS ? sound_id : 0;
 }
 
-    void
-f_sound_playevent(typval_T *argvars, typval_T *rettv)
+void
+f_sound_playevent(typval_T	*argvars,
+		  typval_T	*rettv
+		  )
 {
     sound_play_common(argvars, rettv, FALSE);
 }
@@ -246,8 +255,10 @@ f_sound_playevent(typval_T *argvars, typval_T *rettv)
 /*
  * implementation of sound_playfile({path} [, {callback}])
  */
-    void
-f_sound_playfile(typval_T *argvars, typval_T *rettv)
+void
+f_sound_playfile(typval_T	*argvars,
+		 typval_T	*rettv
+		 )
 {
     sound_play_common(argvars, rettv, TRUE);
 }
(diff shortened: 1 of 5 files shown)


dundargoc:
The style guide was short, so I just winged most of it. Lemme know if something needs to change.

chrisbra:
you are impatient ;) 

I haven't had a chance to look at this in detail and the diff is very messy. But in general, yes I think we have interest as I mentioned before somewhere. 

Can we test it against some small file first? perhaps sound.c?

src/alloc.c
```diff
@@ -28,57 +28,55 @@ static long_u mem_peak;
...
-    static void
-mem_post_alloc(
-    void **pp,
-    size_t size)
+static void
+mem_post_alloc(void **pp, size_t size)
 {
     if (*pp == NULL)
-	return;
+        return;
```
k-takata: At least, I would not support this if this kind of indentation difference existed.

dundargoc: Do you mean using spaces over tabs?

src/arabic.c
```diff
@@ -105,113 +105,111 @@ static struct achar {
...
+        else
+            l = m + 1;
     }
     return NULL;
 }
 
 /*
  * Change shape - from Combination (2 char) to an Isolated
  */
-    static int
```
yegappan: The convention is to indent the return value of the function.  The function name starts in column 1.  This makes it easy to search for the location where a function is defined.

src/arabic.c
```diff
@@ -220,27 +218,26 @@ chg_c_laa2i(int hid_c)
...
-	    break;
-	case a_ALEF_HAMZA_BELOW:
-	    tempc = a_f_LAM_ALEF_HAMZA_BELOW;
-	    break;
-	case a_ALEF:
-	    tempc = a_f_LAM_ALEF;
-	    break;
-	default:
-	    tempc = 0;
+    switch (hid_c) {
```
yegappan: The convention is to place the opening curly braces for the switch statement in a separate line.

src/arabic.c
```diff
@@ -262,36 +259,35 @@ can_join(int c1, int c2)
...
 /*
  * Check whether we are dealing with Arabic combining characters.
  * Note: these are NOT really composing characters!
  */
-    int
-arabic_combine(
-    int		one,	    // first character
-    int		two)	    // character just after "one"
+int
+arabic_combine(int one,        // first character
```
yegappan: The convention is to place the arguments in separate lines (one per line) if placing all of them in a single line will exceed 80 columns.  Another convention is to align the argument names by using a softtab.

src/alloc.c
dundargoc: I don't understand what you mean lol

src/arglist.c
```diff
@@ -95,35 +95,35 @@ alist_new(void)
...
-    char_u	**new_arg_files;
-    int		new_arg_file_count;
-    char_u	*save_p_su = p_su;
-    int		i;
+    char_u **old_arg_files;
+    int old_arg_count;
+    char_u **new_arg_files;
+    int new_arg_file_count;
+    char_u *save_p_su = p_su;
+    int i;
```
yegappan: As mentioned in the other comment, the existing formatting for the variables should be kept.  The variable names should be aligned using a softtab.

yegappan (changes requested):

src/arglist.c
dundargoc: When should tabs be used and when should spaced be used?

yegappan: The following indentation settings are used in the Vim source code:

tabstop=8
shiftwidth=4
softtabstop=4

Spaces are used for indentation.  If the number of spaces is equal to a tab width (8),
then a tab character is used instead of 8 spaces.  This behavior is described in the
help for 'softtabstop'.  The idea is to reduce the number of characters used for
indentation.

dundargoc: Wait, you indent with spaces but align with tabs? Cursed.

src/arabic.c
dundargoc: How is the return type be indented related to the function name starting on column 1?

dundargoc: Oh, you mean the return type is indented *and* the function name starts at column 1? Alright.

dundargoc:
> you are impatient ;)

I prefer the term productive 💅 

> Can we test it against some small file first? perhaps sound.c?

Done

src/sound.c
```diff
@@ -15,36 +15,36 @@
...
-    callback_T	snd_callback;
-#ifdef MSWIN
-    MCIDEVICEID	snd_device_id;
-    long	snd_id;
-#endif
-    soundcb_T	*snd_next;
+    callback_T snd_callback;
+# ifdef MSWIN
+    MCIDEVICEID snd_device_id;
+    long snd_id;
```
yegappan: As commented on the previous changes, the variable names should be aligned on a softtab boundary and not with a single space.

src/sound.c
```diff
@@ -15,36 +15,36 @@
...
 
-static soundcb_T    *first_callback = NULL;
+static soundcb_T *first_callback = NULL;
 
 /*
  * Return TRUE when a sound callback has been created, it may be invoked when
  * the sound finishes playing.  Also see has_sound_callback_in_queue().
  */
-    int
+int
```
yegappan, line 37: The return type should be indented by one 'shiftwidth'.

src/sound.c
```diff
@@ -297,67 +293,67 @@ sound_free(void)
...
+    soundcb_T *p;
 
     switch (message)
     {
-	case MM_MCINOTIFY:
-	    for (p = first_callback; p != NULL; p = p->snd_next)
-		if (p->snd_device_id == (MCIDEVICEID) lParam)
-		{
-		    char	buf[32];
+    case MM_MCINOTIFY:
```
yegappan: The "case" statement should be indented one 'shiftwidth' from the "switch" statement.

yegappan (changes requested):

src/sound.c
dundargoc: This might be the one rule that could cause a problem. I'll see what I can do.

src/arabic.c
dundargoc: Is 80 max width in general?

src/arglist.c
dundargoc: > The idea is to reduce the number of characters used for
indentation.

Why is this desirable?

yegappan: This is done to reduce the size of the source files.  Note that this convention was followed from the time when storage space was expensive.

chrisbra:
it seems, this needs more discussions, let me move it to backlog and we can re-visit after vim9.1 release

dundargoc:
> it seems, this needs more discussions, let me move it to backlog and we can re-visit after vim9.1 release

@chrisbra revisit away

src/sound.c
chrisbra: we touched this also in https://github.com/vim/vim/pull/15939#discussion_r1818170193 it seems this is not easily possible. I would be fine with changing the return type to start at the beginning of the line.

src/sound.c
```diff
@@ -70,8 +70,12 @@ get_sound_callback(typval_T *arg)
 /*
  * Call "soundcb" with proper parameters.
  */
-    void
-call_sound_callback(soundcb_T *soundcb, long snd_id, int result)
+void
+call_sound_callback(
+		    soundcb_T	*soundcb,
```
chrisbra: is it possible to have the first argument directly at the function definition without the first line break?

chrisbra:
I am fine with the suggested changes. Any other opinions?

k-takata:
Does this support the style I commented on in https://github.com/vim/vim/pull/15789#discussion_r1787152356?

```c
    if (cond1
            && cond2)   // <- additional indentation here
    {
        ...
    }
```

src/sound.c
```diff
@@ -108,23 +112,23 @@ delete_sound_callback(soundcb_T *soundcb)
...
 typedef struct soundcb_queue_S soundcb_queue_T;
 
 struct soundcb_queue_S {
     soundcb_queue_T	*scb_next;
-    uint32_t		scb_id;		// ID of the sound
-    int			scb_result;	// CA_ value
-    soundcb_T		*scb_callback;	// function to call
+    uint32_t		scb_id;         // ID of the sound
+    int			scb_result;     // CA_ value
+    soundcb_T		*scb_callback;  // function to call
```
k-takata, line 130: The tabs before the comments are changed to spaces.
Can we keep using tabs here?

saccarosium: I think we should get rid of tabs all together and just use spaces

k-takata: This conflicts with the current modeline settings.

As I said somewhere else before, changing tabs to spaces should be a separate patch.

saccarosium: Ah yes I remember. You've mentioned in here https://github.com/vim/vim/pull/15789#discussion_r1812794284.

> This conflicts with the current modeline settings.

The modeline should be changed gradually as we format more files

> As I said somewhere else before, changing tabs to spaces should be a separate patch.

Do you mean do a `:retab`? I can do a separate PR and then @dundargoc can rebase on that.

saccarosium:
> Can we incorporate this automated formatting immediately after the next major release (9.2) for all the source files?

@yegappan that would be great! I'm working on some large changes myself so I would gladly help with the migration.

The toughest files for the formatters are those with high usage of the preprocessor. So I think after this PR we should focus on dial the formatter for those files (`src/structs.h` or `src/evalfunc.h` comes to my mind).

dundargoc:
> Can we incorporate this automated formatting immediately after the next major release (9.2) for all the source files?

Uh, we already postponed the formatting for 9.1. Why is this arbitrary requirement even useful?

dundargoc:
>×Another problem is that applying autoformatting or especially changing tabs breaks all the pending PRs and patches.

Yes? Most of the work is avoided by formatting the PR and rebasing. This is further mitigated if the formatting is done in chunks. Other than that, yes, some modicum of effort from contributors will be needed in some instances. If people think that is unacceptable then introducing a formatter will essentially be impossible.

> Moreover, quite a few people own their patches to customize Vim.

I don't understand what this means.

src/sound.c
dundargoc: The formatter (both clang-fornat and uncrustify) can change tabs to spaces. This should be done via the formatter and not retab.

k-takata:
> Uh, we already postponed the formatting for 9.1.

There was no agreement at that time.

> Most of the work is avoided by formatting the PR and rebasing.

Rebasing cannot be (easily) automated.

> > Moreover, quite a few people own their patches to customize Vim.
> 
> I don't understand what this means.

I have about 10~15 patches that I haven't created PR yet. Some of them are not mature yet and some others are too niche and need to be refined.
I also see people who have such patches.

dundargoc:
> There was no agreement at that time.

I am so confused on what point you're trying to make. Before we postponed because there was no agreement. Now we should/shouldn't postpone because we have/don't have agreement?

> Rebasing cannot be (easily) automated.

So what are you suggesting? If you don't want a formatter then it would be more productive for you to just say that, rather than trying to convince me that introducing a formatter is somehow an exotic or unsolved problem. It's not, it's been done countless times before, and rebasing and conflict fixing is just a fact of life.

k-takata:
Introducing auto-formatting should be okay.
However, expanding tabs should be avoided. It conflicts with the current modeline settings.
(Expanding tabs should be a separate step if it is needed and should be after 9.2.)

dundargoc:
> Does this support the style I commented on in [#15789 (comment)](https://github.com/vim/vim/pull/15789#discussion_r1787152356)?
> 
> ```c
>     if (cond1
>             && cond2)   // <- additional indentation here
>     {
>         ...
>     }
> ```

Kinda, but not really. I think the following is the closest and should achieve the same thing?

```
    return a1
	   && a2
	   && (  a1->initial
	      || a1->medial)
	   && (  a2->final
	      || a2->medial);
```

dundargoc:
> Expanding tabs should be a separate step if it is needed and should be after 9.2.

OK, but again, why? What is so special about 9.2 that we should wait for it? What conditions are different if we do it before 9.2 and after 9.2? No one has still actually explained this.

I could not find a way to only keep tabs for comments. How would you like to proceed?

saccarosium:
> However, expanding tabs should be avoided. It conflicts with the current modeline settings.

Yes ok it will change gradually. When we format a new file we will change the modeline. It is not something that will be done immediately for the all codebase.

> (Expanding tabs should be a separate step if it is needed and should be after 9.2.)

It will be needed since we *don't* want tabs we want all spaces... I don't really get this part either. I know your did mentioned that it will mess some patches, and that's why I would like ideally to format gradually the codebase before having it run on every commit.

> I have about 10~15 patches that I haven't created PR yet. Some of them are not mature yet and some others are too niche and need to be refined.

Ok but if they are immature and they need refinement why not simply either format them or simply makes two PRs one addressing the formatting and one addressing the patch itself? Also I want to reiterate that we are moving all spaces. *It is the future of the codebase*. It is better if new code is at least is using all spaces so it is less work when we will inevitably change that.

k-takata:
It seems that `FOR_ALL_*` macros are not correctly indented. (E.g. `FOR_ALL_SIGNS()` in sign.c)

dundargoc:
> It seems that `FOR_ALL_*` macros are not correctly indented. (E.g. `FOR_ALL_SIGNS()` in sign.c)

Uh, I haven't touched sign.c? I've only formatted sound.c as recommended to start small.

k-takata:
I tried your uncrustify.cfg and checked if it can be used for other files.

dundargoc:
I guess we should format sign.c in this PR as well if we want to discuss it. I can update the PR later when I can.

dundargoc:
> It seems that `FOR_ALL_*` macros are not correctly indented. (E.g. `FOR_ALL_SIGNS()` in sign.c)

I don't know what the "correct" indentation would be for defines, which is why reviewing files not in the PR is not that useful. I've formatted sign.c.

dundargoc:
As for clang-format vs uncrustify: I don't personally think it matters much. Uncrustify is more configurable, which means more annoying to use, but might be easier to get buy-in from projects with pre-existing style standard. It's possible to change later down the line if wanted.

saccarosium:
I just want to add that I'm working on adding support to uncrustify to indent return types. It should be done in about a week.

puremourning:
The de facto formatter for c/c++ these days is clang-format, and it's supported by things like clangd LSP.

What was the motivation for uncrustify vs that vs the other options that could be used?

clason:
> As for clang-format vs uncrustify: I don't personally think it matters much. Uncrustify is more configurable, which means more annoying to use, but might be easier to get buy-in from projects with pre-existing style standard. It's possible to change later down the line if wanted.

☝️

dundargoc:
> What was the motivation for uncrustify vs that vs the other options that could be used?

What was the motivation to not read the PR description or the comments 🤪?

But yeah, I think uncrustify is easier to introduce to codebases with already established style format as it's very configurable. Especially considering the extreme resistance and weird nitpicking I figured I might need all the edge I could use. I don't mind using clang-format over uncrustify (in fact, I think I might prefer clang-format).

@chrisbra are you fine with using clang-format? It might mean some parts of the formatting will be more rigid.

chrisbra:
> @chrisbra are you fine with using clang-format? It might mean some parts of the formatting will be more rigid.

I haven't used neither and I don't have an opinion on whether to prefer one over the other. So I don't really mind

saccarosium:
> What was the motivation for uncrustify vs that vs the other options that could be used?

The main motivation was that clang-format doesn't play nice with the preprocessor + the quirks of the current codebase. The main problem was alignment of variable names, that it is broken in most files without any obvious way to fix it.

I did a PR where I was trying to introduce a clang format (#15789) and I couldn't compile anymore if I let clang-format format the entire codebase. So the best movement, at least for now is to use uncrustify, which is working really well with the vim codebase.

dundargoc:
> I did a PR where I was trying to introduce a clang format (https://github.com/vim/vim/pull/15789) and I couldn't compile anymore if I let clang-format format the entire codebase.

Tbh that is not that unusual, and I bet if I used uncrustify over the whole codebase the same problem would occur. It's common to just disable formatting the parts that break compilation, even with uncrustify. I can test your PR to see if it's large sections that we need to disable with clang-format. If not then I think clang-format is preferable since it's easier to use.

saccarosium:
Ok, I've revisited my clang-format PR. So I can distill some pros and cons.

## Clang-format

### Pros

- Widely used (most C devs probably know it already)
- Simple to get installed (most C devs probably they have it already installed)
- Healthy and maintained (since it is part of the clang toolchain it is maintained and I don't have much fear of need it to fork it)
- Simpler to work with

### Cons

- We loose reliable alignment of variables (but I remembered that any of the core maintainers didn't complain about it)
- We loose indented return types (but it was accepted)
- We need to be careful to format gradually since vim won't compile if we format everything without checking and reviewing (we need to do it anyway)

## Uncrustify

### Pros

- It does parsing of the file and gives much better result for niche formatting tasks such as aligning comments and align variables names
- It gives you much more control over the formatting

### Cons

- More niche (most devs won't know it) 
- There aren't very up to date packages so maintainers need to compile it from source (I'm basing this assumption on the state of the arch package. I'm probably wrong about this)
- There might be files that it's not capable to format in a timely fashion. I've tried to format `channel.c` it didn't finish after 1+ minutes
- It is more complicated to use and manage since it does give you a lot of option to play with (but we will probably configure it once and than forget about it. Also it is probably more of a plus in our use case)
- Not sure if it healthy. The code is very messy and the maintainer says they don't have much time to work on it. I don't know if we will, down the line, need to fork it. 

This is my summary after some more research. Let me know what you guys think is the best path foward.

dundargoc:
Ye, let's go ahead with clang-format.

dundargoc:
Superseded by https://github.com/vim/vim/pull/16019

Issue #11934, opened 2023-02-02
Syntax filetypes: add support for deb822-formatted Debian source-list files 


jayaddison:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
`vim` currently supports syntax-highlighting matching for [two Debian source-list file patterns](https://github.com/vim/vim/blob/be4e01637e71c8d5095c33b9861fd70b41476732/runtime/filetype.vim#L515-L517) (files named `/etc/apt/source.list`, and files named `*.list` within `/etc/apt/sources.list.d/`).

Debian is gradually introducing [deb822-format source lists](https://manpages.debian.org/bullseye/apt/sources.list.5.en.html#DEB822-STYLE_FORMAT), and it would be nice to match those filetypes within `vim`.

**Describe the solution you'd like**
Extend the filetype matching logic in `runtime/filetype.vim` to support deb822-format sources files.

**Describe alternatives you've considered**
N/A

**Additional context**
The existing [`debsources.vim`](https://github.com/vim/vim/blob/be4e01637e71c8d5095c33b9861fd70b41476732/runtime/syntax/debsources.vim) may provide suitable initial syntax highlighting, although it doesn't highlight field names (`Components`, for example).  I'll file a separate issue to see whether they could/should be supported there.

jayaddison:
> The existing [`debsources.vim`](https://github.com/vim/vim/blob/be4e01637e71c8d5095c33b9861fd70b41476732/runtime/syntax/debsources.vim) may provide suitable initial syntax highlighting, although it doesn't highlight field names (`Components`, for example). I'll file a separate issue to see whether they could/should be supported there.

On second thoughts: it might be distracting/spammy for me to create another GitHub issue primarily to ask that question.  Asking here instead:

If-and-when adding syntax highlighting for [deb822-format `apt` source lists](https://manpages.debian.org/bullseye/apt/sources.list.5.en.html#DEB822-STYLE_FORMAT), is there a recommendation/preference about whether to do that by modifying the existing `debsources.vim` compared to adding a new syntax file?

brammool:
Our Debian specialist is @jamessan 

If there are two filetypes with a similar syntax, the best way to implement is to have two separate syntax files, and have them include the common parts from a shared file, possibly auto-loaded.  How exactly depends on what the maintainer(s) prefer to do.

PR #13320, opened 2023-10-12
Add new deb822sources filetype

jamessan:
I'll follow up with a Debian runtime update, including the new syntax
file for deb822sources (to address #11934), after this.


chrisbra:
Thanks. I include it and await a follow up then ;)

PR #13321, opened 2023-10-12
Fix getmousepos() returning wrong "column" when clicking on a TAB

zeertzjq:
Problem:  When clicking in the middle of a TAB, getmousepos() returns
          the column of the next char instead of the TAB.
Solution: Break out of the loop when the vcol to find is inside current
          char. Fix invalid memory access when calling virtcol2col() on
          an empty line.


chrisbra:
thanks!

PR #13323, opened 2023-10-12
make :0cq return zero in exmode

chrisbra:
Currently, when an error occurs in ex-mode (e.g. -e or -E) and an error occurs, '-0cq' will still abort with an error.

Let's make it possible to return zero even in that case, so that Vim explicitly exits with zero exit code, even if an error occurred.

Add a few tests to verify this.

This also allows to use '&&' in the test_crash test suite.

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -5954,6 +5954,9 @@ ex_quit(exarg_T *eap)
     static void
 ex_cquit(exarg_T *eap UNUSED)
 {
+    if (eap->addr_count > 0 && (int)eap->line2 == 0 && exmode_active)
+	ex_exitval = 0;
+
     // this does not always pass on the exit code to the Manx compiler. why?
     getout(eap->addr_count > 0 ? (int)eap->line2 : EXIT_FAILURE);
 }
--- a/src/testdir/test_crash.vim
+++ b/src/testdir/test_crash.vim
@@ -17,7 +17,7 @@ func Test_crash1()
   let buf = RunVimInTerminal('sh', opts)
 
   let file = 'crash/poc_huaf1'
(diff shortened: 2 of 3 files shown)


chrisbra:
BTW: I am not sure, this is actually a good idea :/

chrisbra:
I will postpone this, perhaps only include the tests

errael:
> BTW: I am not sure, this is actually a good idea :/

When I first saw this PR, I strongly agreed with that statement. Why not just use `:qall!` and make the intent clear? (I only learned about `:cq` because it's an important exit option from `Splice9`.)

But, the docs, `:help cq` do say
```
{N} can also be zero, in which case Vim exits normally.
```
And that is how it works today, v9.0.2043. Ah, but I tested typing `./vim`. After further examination, I should say that's how it sometimes works today.
1. `vim`, `:0cq` exits with `0`
2. `vim -e`, `0cq` exits with `0`
3. `vim -E`, `0cq` exits with `0`
4. `ex`, `0cq` exits with `1`
5.  `vim`, `gQ`, `0cq`, exits with `0`
6. `vim`, `:unmap Q`, `Q`, `0cq` exits with `0`.

(2.) is a surprise, since `vim -e` is supposed to be `like "ex"`. And (5.) is also somewhat of a surprise.

Any idea why pure ex mode behaves that way (usually)? Some kind of compatibility mode? (Now, all I have to say is: 
my head hurts, please make it stop)

chrisbra:
I am surprised, your `ex -c '0cq'` does not exit with zero value. I think it only exits with non-zero, if an error occured.

errael:
> I am surprised, your `ex -c '0cq'` does not exit with zero value. I think it only exits with non-zero, if an error occured.

Some of the tests I ran were on a local repo that was fresh build, it got errors trying to read initialization scripts. That affected the results. I'll take another look later.

PR #9707, opened 2022-02-07
Port the Neovim popup menu support for command-line completion

yegappan:
The following Neovim patches are part of this PR:

Author: Björn Linse <bjorn.linse@gmail.com>
Commit: https://github.com/neovim/neovim/commit/be8ebba325451b387c0aedacfcda6c53e6c51188

Author: Björn Linse <bjorn.linse@gmail.com>
Commit: https://github.com/neovim/neovim/commit/0636b25f28e408c8b16026354db7edfef079614a

Author: Billy Su <g4691821@gmail.com>
Commit: https://github.com/neovim/neovim/commit/961c528afc7f3a21a8c4b118a2c585dbe670c5c5

Author: akovaski <akaskik@gmail.com>
Commit: https://github.com/neovim/neovim/commit/97dcc48c998ccecaa37a3cbea568d85c2f1407f9

Author: glacambre <code@lacamb.re>
Commit: https://github.com/neovim/neovim/commit/958ca938b44ea7d6f126887f5bec98ee6d2f6f15

The following Neovim patch is not applied as the Vim screen redraw code is different and the popup
menu is already correctly positioned:

Author: Björn Linse <bjorn.linse@gmail.com>
Commit: https://github.com/neovim/neovim/commit/b98d1cf68bd1d78d8170a8d994ba7dae5540962a

If we run into any problems related to positioning the popup menu, then we need to revisit this patch.

PR #13331, opened 2023-10-12
Improve command-line completion docs

ychin:
Add more details about 'ignorecase' and its effect on cmdline completion.

Make sure keys used in wildmenu are properly documented and linked in the keys' documentation entries, and in `:h index` for proper cross-referencing, as wildmenu popup is slightly different from insert-mode popup menu.

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -90,7 +90,7 @@ CTRL-SHIFT-Q	Works just like CTRL-V, unless |modifyOtherKeys| is active,
 		In the GUI the |key-notation| is inserted without simplifying.
 
 							*c_<Left>* *c_Left*
-<Left>		cursor left. See 'wildmenu' for behavior during wildmenu
+<Left>		cursor left.  See 'wildmenu' for behavior during wildmenu
 		completion mode.
 							*c_<Right>* *c_Right*
 <Right>		cursor right.  See 'wildmenu' for behavior during wildmenu
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -90,9 +90,11 @@ CTRL-SHIFT-Q	Works just like CTRL-V, unless |modifyOtherKeys| is active,
 		In the GUI the |key-notation| is inserted without simplifying.
 
 							*c_<Left>* *c_Left*
-<Left>		cursor left
+<Left>		cursor left. See 'wildmenu' for behavior during wildmenu
+		completion mode.
 							*c_<Right>* *c_Right*
-<Right>		cursor right
+<Right>		cursor right.  See 'wildmenu' for behavior during wildmenu
+		completion mode.
 							*c_<S-Left>*
 <S-Left> or <C-Left>					*c_<C-Left>*
 		cursor one WORD left
@@ -102,7 +104,8 @@ CTRL-SHIFT-Q	Works just like CTRL-V, unless |modifyOtherKeys| is active,
 CTRL-B or <Home>				*c_CTRL-B* *c_<Home>* *c_Home*
 		cursor to beginning of command-line
 CTRL-E or <End>					*c_CTRL-E* *c_<End>* *c_End*
-		cursor to end of command-line
+		cursor to end of command-line.  See 'wildmenu' for behavior
+		during wildmenu completion mode.
 
 							*c_<LeftMouse>*
 <LeftMouse>	Move the cursor to the position of the mouse click.
@@ -237,6 +240,7 @@ CTRL-\ e {expr}						*c_CTRL-\_e*
 CTRL-Y		When there is a modeless selection, copy the selection into
 		the clipboard. |modeless-selection|
 		If there is no selection CTRL-Y is inserted as a character.
+		See 'wildmenu' for behavior during wildmenu completion mode.
 
 CTRL-M or CTRL-J		*c_CTRL-M* *c_CTRL-J* *c_<NL>* *c_<CR>* *c_CR*
 <CR> or <NL>	start entered command
@@ -252,12 +256,14 @@ CTRL-C		quit command-line without executing
 
 							*c_<Up>* *c_Up*
 <Up>		recall older command-line from history, whose beginning
-		matches the current command-line (see below).
+		matches the current command-line (see below).  See 'wildmenu'
+		for behavior during wildmenu completion mode.
 		{not available when compiled without the |+cmdline_hist|
 		feature}
 							*c_<Down>* *c_Down*
 <Down>		recall more recent command-line from history, whose beginning
-		matches the current command-line (see below).
+		matches the current command-line (see below).  See 'wildmenu'
+		for behavior during wildmenu completion mode.
 		{not available when compiled without the |+cmdline_hist|
 		feature}
 
@@ -463,10 +469,15 @@ When repeating 'wildchar' or CTRL-N you cycle through the matches, eventually
 ending up back to what was typed.  If the first match is not what you wanted,
 you can use <S-Tab> or CTRL-P to go straight back to what you typed.
 
-The 'wildignorecase' option can be set to ignore case in filenames.
-
 The 'wildmenu' option can be set to show the matches just above the command
 line.
+
+The 'wildoptions' option provides additional configuration to use a popup menu
+for 'wildmenu', and to use fuzzy matching.
+
+The 'wildignorecase' option can be set to ignore case in filenames.  For
+completing other texts (e.g. command names), the 'ignorecase' option is used
+instead (fuzzy matching always ignores case, however).
 
 If you like tcsh's autolist completion, you can use this mapping:
 	:cnoremap X <C-L><C-D>
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -9336,9 +9336,8 @@ A jump table for the options with a short description can be found at |Q_op|.
 			  parent directory or parent menu.
 	<Right>		- in filename/menu name completion: move into a
 			  subdirectory or submenu.
-	<Up>		- select previous match (like CTRL-P)
-	<PageUp>	  Select a match several entries back, but don't insert it.
-	<PageDown>	  Select a match several entries further, but don't insert it.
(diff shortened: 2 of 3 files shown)


ychin:
Inspired from https://github.com/neovim/neovim/issues/25253.

For Neovim downstream, I was playing around with Neovim, and there are some minor differences that surprised me.
1. In Vim, Up/Down vs Left/Right is flipped in `wildmenu` depending on whether you are using `wildoptions+=pum` or not, which makes intuitive sense to me. In Neovim, that is not the case.
2. Also, in Neovim, Ctrl-E and Ctrl-Y to dismiss/select a choice works both with/without `pum`. In Vim, it only works with `pum`. I actually think it may not be a bad idea to make Ctrl-E/Ctrl-Y works across the board in Vim as well.

Otherwise, probably done with my cmdline completion PRs. Contemplating if I want to work on `has(` and `exists(` completion as well as I have had situations where I wanted that to complete but it didn't, but it seems a little too niche of a feature.

runtime/doc/cmdline.txt
```diff
@@ -90,9 +90,11 @@ CTRL-SHIFT-Q	Works just like CTRL-V, unless |modifyOtherKeys| is active,
 		In the GUI the |key-notation| is inserted without simplifying.
 
 							*c_<Left>* *c_Left*
-<Left>		cursor left
+<Left>		cursor left. See 'wildmenu' for behavior during wildmenu
```
zeertzjq:
```suggestion
<Left>		cursor left.  See 'wildmenu' for behavior during wildmenu
```

runtime/doc/options.txt
```diff
@@ -9314,29 +9314,33 @@ A jump table for the options with a short description can be found at |Q_op|.
...
 
 	When using the popup menu for command line completion, the following
 	keys have special meanings:
-	<Down>		- select next match (like CTRL-N)
+	<Up> <Down>	- select previous/next match (like CTRL-P/CTRL-N)
 	<Left>		- in filename/menu name completion: move up into
 			  parent directory or parent menu.
 	<Right>		- in filename/menu name completion: move into a
 			  subdirectory or submenu.
 	<Up>		- select previous match (like CTRL-P)
```
zeertzjq: This line should be removed now.

runtime/doc/options.txt
```diff
@@ -9314,29 +9314,33 @@ A jump table for the options with a short description can be found at |Q_op|.
...
+	<PageUp>	  Select a match several entries back, but don't insert it.
+	<PageDown>	  Select a match several entries further, but don't insert it.
```
```suggestion
	<PageUp>	- Select a match several entries back, but don't insert it.
	<PageDown>	- Select a match several entries further, but don't insert it.
```

yegappan:
> Inspired from [neovim/neovim#25253](https://github.com/neovim/neovim/issues/25253).
> 
> For Neovim downstream, I was playing around with Neovim, and there are some minor differences that surprised me.
> 
> 1. In Vim, Up/Down vs Left/Right is flipped in `wildmenu` depending on whether you are using `wildoptions+=pum` or not, which makes intuitive sense to me. In Neovim, that is not the case and Left/Right always select previous/next, and Up/Down always go up/down a menu/folder level.

Yes.  When I implemented this (https://github.com/vim/vim/pull/9707), Bram gave a comment to change the cursor keys to make it more intuitive.

> 2. Also, in Neovim, Ctrl-E and Ctrl-Y to dismiss/select a choice works both with/without `pum`. In Vim, it only works with `pum`. I actually think it may not be a bad idea to make Ctrl-E/Ctrl-Y works across the board in Vim as well.

In the command-line mode, CTRL-E moves the cursor to the end of the command-line.  As long as
this functionality is kept, we can use CTRL-E to cancel the wildmenu.

> 
> Otherwise, probably done with my cmdline completion PRs. Contemplating if I want to work on `has(` and `exists(` completion as well as I have had situations where I wanted that to complete, but it seems a little too niche of a feature.

ychin:
> > 2. Also, in Neovim, Ctrl-E and Ctrl-Y to dismiss/select a choice works both with/without `pum`. In Vim, it only works with `pum`. I actually think it may not be a bad idea to make Ctrl-E/Ctrl-Y works across the board in Vim as well.
> 
> In the command-line mode, CTRL-E moves the cursor to the end of the command-line. As long as this functionality is kept, we can use CTRL-E to cancel the wildmenu.

Playing around with it, actually both CTRL-E and CTRL-Y (with "pum") seem to have an odd (buggy) behavior. They would cancel / select the selection, but also put you at the end of the line. So let's say you have the cursor in the middle of the line like so: `:set di<cursor> wrap`. You press Tab to bring up auto-complete and then use CTRL-E/CTRL-Y to cancel/select, the cursor will be pushed all the way to the end after "wrap". Seems like a minor bug to me and would only matter if you start auto-complete in the middle of the line instead of at the end. If you just shift-tab back to an empty choice your cursor will stay in place, just as expected.

chrisbra:
> In the command-line mode, CTRL-E moves the cursor to the end of the command-line. As long as
this functionality is kept, we can use CTRL-E to cancel the wildmenu.

Agree, I am also all for consistency here.

Shane-XB-Qian:
> 2\. Ctrl-E and Ctrl-Y to dismiss/select a choice works both with/without `pum`

`ctrl-e` (or something else?) to dismiss choice (even not when 'pum') maybe useful IF wildmode is `list,full`, 
otherwise the output list items perhaps took up too much screen but seems no good way to clean/reset it.

ychin:
> `ctrl-e` (or something else?) to dismiss choice (even not when 'pum') maybe useful IF wildmode is `list,full`,
> otherwise the output list items perhaps took up too much screen but seems no good way to clean/reset it.

Just to clarify, did you mean otherwise (if not using `list,full`) it's not useful to have Ctrl-E dismiss choice? I think either way the proposal is to have Ctrl-E/Y dismiss/select the current choice only when `wildmenu` is up (we have two flavors of wildmenu: popup or non-popup). The only potential drawback of this is that if you have wildmenu up, and you want to go to the end, you have to press Ctrl-E to dismiss wildmenu, then Ctrl-E to go to the end, which I think is not a big deal.

For the `list` part of `wildmode`, Ctrl-E won't activate since there is no wildmenu used here. Yes, it may take up a lot of screen real estate due to the listing, but I'm not talking about using Ctrl-E to clear the listed items.

Shane-XB-Qian:
> > `ctrl-e` (or something else?) to dismiss choice (even not when 'pum') maybe useful IF wildmode is `list,full`,
> > otherwise the output list items perhaps took up too much screen but seems no good way to clean/reset it.
> 
> Just to clarify, did you mean otherwise (if not using `list,full`) it's not useful to have Ctrl-E dismiss choice?

there are 3 cases as my knowledge:
1, 'pum' like compl 'popup'
2, 'wildmenu' like 'stl'
3, 'list' like 'cli'

for now:
seems 'c-e' already be used to dismiss 'pum' and 'wildmenu' if they are presented,
// or/but to 'wildmenu' seems use 'c-b' (to line head) etc also can dismiss it.
so now the problem is 'list', seems no way to dismiss the choice output,
i didnot mean it have to use 'c-e' to dismiss it, but just say to 'list' it is no way to dismiss.

-- 
shane.xb.qian

ychin:
> seems 'c-e' already be used to dismiss 'pum' and 'wildmenu' if they are presented,

No, Ctrl-E only works for "pum", not old-style "wildmenu". I'm suggesting Ctrl-E works with old-style "wildmenu" as well.

But yes, with a long list-style it's annoying when it completely takes over the screen. I think it's a somewhat different issue with redrawing a screen.

Shane-XB-Qian:
> > seems 'c-e' already be used to dismiss 'pum' and 'wildmenu' if they are presented,
> 
> No, Ctrl-E only works for "pum", not old-style "wildmenu". I'm suggesting Ctrl-E works with old-style "wildmenu" as well.
> 
> But yes, with a long list-style it's annoying when it completely takes over the screen. I think it's a somewhat different issue from this one.

um.. even `<up>` and `<down>` (not only `<c-e>`) can dismiss 'wildmenu' in 'stl-like' bar, at my v9.0.2018
or whatever, i just say 'list' may need some key ('c-e' or something else) to dismiss its choice output list.
// but yes, seems it is another/different issue.

ychin:
<Up> / <Down> keeps the selection, which is not what Ctrl-E does. Ctrl-E cancels the selection and go back to the original typed text. It's the same behavior as insert-mode popup completion. Maybe the word "dismiss" was poorly chosen. "Cancel" is more correct.

What <Up> / <Down> does is more similar to Ctrl-Y, which does keep the selection and dismisses the wildmenu. One caveat is that <Up>/<Down> has special behaviors when selecting files as they let you navigate through the folder structure, whereas Ctrl-Y would just keep the selection and exits wildmenu. You should read the documentation that I updated in this PR :).

Shane-XB-Qian:
> <Up> / <Down> keeps the selection, which is not what Ctrl-E does. Ctrl-E cancels the selection and go back to the original typed text. It's the same behavior as insert-mode popup completion. Maybe the word "dismiss" was poorly chosen. "Cancel" is more correct.
> 
> What <Up> / <Down> does is more similar to Ctrl-Y, which does keep the selection and dismisses the wildmenu. One caveat is that <Up>/<Down> has special behaviors when selecting files as they let you navigate through the folder structure, whereas Ctrl-Y would just keep the selection and exits wildmenu. You should read the documentation that I updated in this PR :).

i meant 'dismiss'
// or whatever, 'dismiss' or 'cancel' both no way to 'list'
// and i knew 'cancel' what it is, actually you were talking about 'cancel'
// which is use 'c-e' to dismiss 'wildmenu' that 'stl-like' bar and cancel compl

-- 
shane.xb.qian

chrisbra:
thanks!

ychin:
> > In the command-line mode, CTRL-E moves the cursor to the end of the command-line. As long as
> > this functionality is kept, we can use CTRL-E to cancel the wildmenu.
> 
> Agree, I am also all for consistency here.

#13362 will try to make Ctrl-E/Ctrl-Y work more consistently.

Issue #12960, opened 2023-08-29
vim9script: "var F: func = obj.Method" or similar is missing


errael:
### Steps to reproduce

There's considerable discussion, with comments about it from Bram, start around https://github.com/vim/vim/issues/12081#issuecomment-1555978215 but that issue was not about this topic and has been closed.

### Expected behaviour

Should be a way to do that.

Often times
```
var F: func = () => obj.Method()
```
is a workaround. But doesn't play with varargs and/or `call()` according to comments.

### Version of Vim

9.1

### Environment

9.1

chrisbra:
closed by 29bb67f1beefc7fd393dbfd9ee77d92f1db3a3c0

Issue #12417, opened 2023-05-20
No way to register a class method as callback in the same class?


lilydjwg:
### Steps to reproduce

Define a class within whose constructor method (`new`), try to register a method as callback:

1.
```vim
vim9script

class MyClass
  this.job: job

  def new()
    this.job = job_start('cat', {
      out_cb: this.OnRequest, # <--- here
    })
  enddef

  def OnRequest(ch: channel, msg: string)
    echo msg
  enddef

endclass
```

=> `E722: Missing comma in Dictionary: OnRequest,`

2. If changed to `out_cb: function('this.OnRequest'),` => `E475: Invalid argument: this.OnRequest`

3.
```vim
vim9script

def OnRequest(c: MyClass, ch: channel, msg: string)
  echo msg
  c.Do()
enddef

class MyClass
  this.job: job

  def new()
    this.job = job_start('cat', {
      out_cb: function(OnRequest, [this]),
    })
  enddef

  def Do()
  enddef

endclass

var t = MyClass.new()
```

=> `E1010: Type not recognized: MyClass`

4.
```vim
vim9script

def OnRequest(c: any, ch: channel, msg: string)
  echo msg
  c.Do()
enddef

class MyClass
  this.job: job

  def new()
    this.job = job_start('echo a', {
      out_cb: function(OnRequest, [this]),
    })
  enddef

  def Do()
  enddef

endclass

var t = MyClass.new()
```
=>
```
Error detected while processing function <SNR>185_OnRequest:
line    2:
E715: Dictionary required 
```

5.
```vim
vim9script

def OnRequest(c: any, ch: channel, msg: string)
  echo msg
  c.Do()
enddef

class MyClass
  this.job: job

  def new()
    this.job = job_start('echo a', {
      out_cb: function(OnRequest, [], this),
    })
  enddef

  def Do()
  enddef

endclass

var t = MyClass.new()
```
=>
```
Error detected while compiling /home/lilydjwg/tmpfs/a.vim[22]..function new:
line    2:
E1013: Argument 3: type mismatch, expected dict<any> but got object<MyClass> 
```

6. 

### Expected behaviour

There should be a way to register class method as callback inside the class' method.

### Version of Vim

9.0.1566

### Environment

Arch Linux

brammool:
Please try 9.0.1569 or later.  At least the first example appears to be fixed now.

lilydjwg:
With 9.0.1570, the same error is only delayed to the time when the class is initialized, e.g.

[...]

Issue #12324, opened 2023-04-30
[vim9script] cannot call class member of funcref type directly, E1325: Method not found on class


qeatzy:
### Steps to reproduce

1. create a file with any name. eg, `repro.vim` with below contents.

```
vim9script

if !exists('g:fns')
    g:fns = [['main.go', 'main_windows.go', 'version.go', 'conn/bind_std.go', 'conn/bind_windows.go', 'conn/boundif_android.go', 'conn/conn.go', 'conn/controlfns.go', 'conn/controlfns_linux.go', 'conn/controlfns_unix.go', 'conn/controlfns_windows.go', 'conn/default.go'], 0]
endif

def Open(filename: string)
    :e `=filename`
enddef

export class FloatWindow
    public this.fns: list<any>
    this.idx: number
    this.cb: func(any) = Open

    # idx = step, return fns[new idx]
    def To(step: number, do: bool = 0): any
        var x = step % len(this.fns)
        x = x >= 0 ? x : x + len(this.fns) 
        var ret = this.fns[this.idx]
        this.idx = x
        if do
            # this.cb(ret)
            # error E1325: Method not found on class "FloatWindow": cb(ret)
            # below is alternative
            var F = this.cb
            F(ret)
        endif
        return ret
    enddef

    # idx += step, return fns[old idx]
    def Inc(step: number): any
        var x = (this.idx + step) % len(this.fns)
        x = x >= 0 ? x : x + len(this.fns) 
        var ret = this.fns[this.idx]
        this.idx = x
        return ret
    enddef

endclass

var fw = FloatWindow.new(g:fns[0])
# echo fw
fw.To(3, 1)
echo mapnew(fw.fns, 'fw.Inc(1)')
g:fw = fw

```

2 `vim --clean -c "so repro.vim"`

### Expected behaviour

Inside method `def To(step: number, do: bool = 0): any`

below is the original version throws error, also try change this.cb to this.Cb, same result
```
        if do
            # E1325: Method not found on class "FloatWindow": cb(ret)
            this.cb(ret)
        endif
```
below is workaround to suppress the `E1325: Method not found on class` message
```
        if do
            # workaround
            var F = this.cb
            F(ret)
        endif
```

Here is the question, is this behavior -- not able to call class member of Funcref type directly -- final and cannot be fixed, or it's just due to some imperfection/oblivion/difficulty in current implementation?

### Version of Vim

[...]

PR #13333, opened 2023-10-13
Support using object and class methods as funcrefs

### LINKED ISSUE #11981: [vim9script] Cannot pass (instance or static) method to function
### Steps to reproduce

Source this script:

```vim
vim9script

def Map(F: func(number): number): func(number): number
  return (n: number) => F(n)
enddef

def Double(n: number): number
  return 2 * n
enddef

class Math
  def Double(n: number): number
    return 2 * n
  enddef

  static def StaticDouble(n: number): number
    return 2 * n
  enddef
endclass

echo Map(Double)(24)                # OK
echo Map((n: number) => 2 * n)(24)  # OK

const math = Math.new()

echo Map(math.Double)(24)           # Error: E1326
echo Map(Math.StaticDouble)(24)     # Error: E1338
```

### Expected behaviour

The script should run without errors.

### Version of Vim

9.0.1306

### Environment

macOS 13.2
Apple Terminal
xterm256-color
ZSH 5.8.1

### Logs and stack traces

_No response_

yegappan:
Fixes #11981, #12417 #12960 and #12324.


errael:
I've been trying a variety of things, haven't been able to break it (yet?). 

While trying stuff with `call` and varargs, I ran into something that _may_ be an issue at the end of the following script.

```
vim9script

class C
  def VA(...args: list<number>): list<number>
	return args
  enddef
  static def SVA(...args: list<any>): list<any>
	return args
  enddef
endclass
var obj = C.new()

echo call(obj.VA, [4, 2])

var FF: func
FF = obj.VA
echo call(FF, [ 1, 2, 3])
echo call(C.SVA, [ 7, 9 ])
echo call("C.SVA", [ 7, 9 ])	# <<< Unknown function: C.SVA
```

yegappan:
> I've been trying a variety of things, haven't been able to break it (yet?).
> 

Thanks for testing the PR and it is good to hear that you didn't encounter any issues.

>
> While trying stuff with `call` and varargs, I ran into something that _may_ be an issue at the end of the following script.
> 
> ```
> vim9script
> 
> class C
>   def VA(...args: list<number>): list<number>
> 	return args
>   enddef
>   static def SVA(...args: list<any>): list<any>
> 	return args
>   enddef
> endclass
> var obj = C.new()
> 
> echo call(obj.VA, [4, 2])
> 
> var FF: func
> FF = obj.VA
> echo call(FF, [ 1, 2, 3])
> echo call(C.SVA, [ 7, 9 ])
> echo call("C.SVA", [ 7, 9 ])	# <<< Unknown function: C.SVA
> ```

Using the object/class method name as a string is currently not supported. You can use it as a funcref (as you have already found out). I will update the documentation.

errael:
> Using the object/class method name as a string is currently not supported

FWICT, using a variable in a string for a function name doesn't work, so I don't think **object method** could ever be used as a string. But maybe static.

errael:
Found something, looks weird.
(fails both in def and script level)

```
T1 111
Error detected while processing :source buffer=2[16]..function <SNR>11_FDef[2]..C.F2:
line    2:
E1327: Object required, found number
```

```
vim9script

class C
  this.val: number
  def F2(): number
	#return 3 		# <<< this works in all cases
	return this.val		# <<< fails when invoked with `call`
  enddef
endclass

def FDef(obj: C)
    echo "T1" obj.F2()
    echo "T2" call(obj.F2, [])
enddef

FDef(C.new(111))
```

errael:
Don't even need to use `call`

```
vim9script

class C
  this.val: number
  def F2(): number
	return this.val
  enddef
endclass

var obj = C.new(333)
echo obj.F2()
var F = obj.F2
echo F()
```

chrisbra:
thanks! let me merge it now, those problems can then be fixed in a followup PR.

chrisbra:
included as of 29bb67f1beefc7fd393dbfd9ee77d92f1db3a3c0

yegappan:
> Don't even need to use `call`

I have created the PR https://github.com/vim/vim/pull/13341 to address this.

> 
> ```
> vim9script
> 
> class C
>   this.val: number
>   def F2(): number
> 	return this.val
>   enddef
> endclass
> 
> var obj = C.new(333)
> echo obj.F2()
> var F = obj.F2
> echo F()
> ```

PR #13334, opened 2023-10-13
Don't free non-heap b:changedtick value

ychin:
The b:changedtick value is managed as part of the `file_buffer` buffer structure and not allocated with alloc(). As such, calling vim_free on it is invalid.

### DIFF
--- a/src/dict.c
+++ b/src/dict.c
@@ -275,6 +275,23 @@ dictitem_remove(dict_T *dict, dictitem_T *item, char *command)
 }
 
 /*
+ * Remove item "item" from Dictionary "dict" but do *not* free it. Only call
+ * this is "item" is managed by the caller to avoid leaks.
+ * "command" is used for the error message when the hashtab if frozen.
+ */
+    void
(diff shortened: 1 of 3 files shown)


ychin:
Discovered this compiling with `-O3 -flto` when playing around with compiler settings.

src/dict.c
```diff
@@ -274,6 +274,23 @@ dictitem_remove(dict_T *dict, dictitem_T *item, char *command)
     dictitem_free(item);
 }
 
+/*
+ * Remove item "item" from Dictionary "dict" but do *not* free it. Only call
+ * this is "item" is managed by the caller to avoid leaks.
```
zeertzjq, line 279:
```suggestion
 * this if "item" is managed by the caller to avoid leaks.
```

ychin:
Nevermind I think I might have made a mistake here.

ychin:
There's a clear check for `DI_FLAGS_ALLOC`, so the code path wouldn't have been triggered. I was kind of blindly following the compiler warning here. Either way this should have caused issues in the past if it was really freeing invalid memory.

PR #13335, opened 2023-10-14
Add a "coladd" field to getmousepos() result

zeertzjq:
Problem:  Cannot accurately get mouse clicking position when clicking on
          a TAB or with virtual text.
Solution: Add a "coladd" field to getmousepos() result.


chrisbra:
thanks

PR #13336, opened 2023-10-14
runtime(doc): correct <PageUp>/<PageDown> behavior in 'wildmenu'

zeertzjq:
"but don't insert it" is wrong.
Also move them just below <Up> <Down> as they are more similar.


chrisbra:
thanks!

PR #13338, opened 2023-10-14
Update `TextChangedI` to only trigger when on text change in Insert mode

### LINKED ISSUE #13265: `TextChangedI` is triggered upon entering Insert mode after triggered `TextChanged`
### Steps to reproduce

1. Create an init file with the following contents:
```vim
put ='a'

let g:n = 0
au TextChangedI * let g:n = g:n + 1
```
2. Open `vim -u <path-to-init-file>`
3. Execute `:echo g:n`. It shows 0.
4. Press `i` to enter Insert mode followed by `<Esc>` to exit it.
5. Execute `:echo g:n` again. It shows 1 while there were no changes in Insert mode done.
6. Press `yyp` three times. This would trigger `TextChanged` again.
7. Repeat steps 3-5. Before entering Insert mode it would show 1, after enter-exit - 2. Although again no changes are made in Insert mode.

### Expected behaviour

`TextChangedI` is triggered *only* when change is *actually* made in Insert mode, as per documentation: "After a change was made to the text in the current buffer in Insert mode."

This was originally discovered in neovim/neovim#25490 and seems to be the result of Neovim porting [8.2.3517 patch](https://github.com/vim/vim/commit/db3b44640d69ab27270691a3cab8d83cc93a0861).

### Version of Vim

9.0.1882

### Environment

Operating system: EndeavourOS Linux x86_64 (6.5.5-arch1-1)
Terminal: st-0.9
Value of `$TERM`: st-256color
Shell: zsh 5.9

### Logs and stack traces

_No response_

echasnovski:
Problem:  `TextChangedI` can trigger on entering Insert mode if there was previously a change not in Insert mode.
Solution: Make it trigger only when text is actually changed in Insert mode.

Resolves #13265. This is basically a @chrisbra patch with updated tests.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -174,6 +174,9 @@ edit(
 	return FALSE;
     }
     ins_compl_clear();	    // clear stuff for CTRL-X mode
+    // Reset Changedtick_i, so that TextChangedI will only be triggered for stuff
+    // from insert mode
+    curbuf->b_last_changedtick_i = CHANGEDTICK(curbuf);
 
     /*
      * Trigger InsertEnter autocommands.  Do not do this for "r<CR>" or "grx".
@@ -840,6 +843,7 @@ doESCkey:
 		if (cmdchar != 'r' && cmdchar != 'v' && c != Ctrl_C)
 		    ins_apply_autocmds(EVENT_INSERTLEAVE);
 		did_cursorhold = FALSE;
+		curbuf->b_last_changedtick = CHANGEDTICK(curbuf);
 		return (c == Ctrl_O);
 	    }
 	    continue;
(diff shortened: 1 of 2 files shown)


chrisbra:
ah thanks. I must have forgotten about this, sorry.

chrisbra:
can we also add a similar test: A change in insert mode, does no longer trigger TextChanged mode? I suppose that also currently happens.

echasnovski:
> ah thanks. I must have forgotten about this, sorry.

No worries. You have enough on your plate right now :)

> can we also add a similar test: A change in insert mode, does no longer trigger TextChanged mode? I suppose that also currently happens.

Done.

echasnovski:
Thanks, Christian!

PR #13344, opened 2023-10-14
Separate vim binary and unittest dependencies, make them parallelizable

ychin:
Clean up make dependencies so Vim and unit test binaries only depend on the object files they need. This fixes an existing issue where after running unit tests, the Vim binary would be invalidated, which results in it having to be linked again when running script tests, even though Vim was already previously built.

Make link.sh (script we use to link those binaries) generate namespaced temporary files for each app to avoid them colliding with each other. This allows `unittesttargets` to be built in parallel.

These fixes are useful when using link-time-optimization as the link phase could now take minutes rather than a few seconds.


chrisbra:
thanks!

Issue #13332, opened 2023-10-13
<ctype.h> is*() and to*() functions are being used unsafely


Keith-S-Thompson:
### Steps to reproduce

There is no easily reproducible bug on most platforms.

The `is*()` and `to*()` standard functions declared in `<ctype.h>` require an argument that either is equal to `EOF` (typically -1) or is in the range of `unsigned char`.

If plain `char` is signed, as it is on many systems, a character extracted from a command line argument, from user interactive input, or from a file can have a negative value, resulting in undefined behavior. Typically this applies only to non-ASCII characters.

In most implementations, these functions behave reasonably sanely for negative values, so in spite of the undefined behavior, the observed behavior is likely to be benign.

One exception: Microsoft's C implementation (I used Visual Studio 2022 on Windows 11) aborts on an invalid argument to these functions, but only when an application is built in Debug mode. I've built `xxd.c` as a standalone application under Visual Studio. The resulting executable crashes when invoked as follows (I named the source file `c.c` for this quick-and-dirty test):
```
Release\c.exe -i héllö.txt
```

I've created a fork of this repo and a branch with a proposed fix for this issue:
https://github.com/Keith-S-Thompson/vim/tree/ctype-fix

I have not yet created a pull request. I'll be glad to do so.

The fix adds a number of macros, such as:
```
#define SAFE_isalnum(c)  (isalnum ((unsigned char)(c)))
```
and replaces most calls to the standard functions with invocations of these macros.

Note that there is already *some* code that's designed to avoid problems calling these functions with negative arguments. This existing code does not cover all cases; my update is intended to do so. For now, I've left most such code in place, though it's probably unnecessary.

(An aside: vim has code that works around implementations where `toupper()` and `tolower()` operate incorrectly on arguments that are not respectively lowercase or uppercase letters. I'm skeptical that there are any existing implementations that (a) have this bug and (b) are able to compile vim, since `toupper()` and `tolower()` have been required to operate correctly since the 1989 ANSI C standard. This is not directly related to this issue, but it's a potential opportunity for some cleanup.)

### Expected behaviour

Calls to `is*()` and `to*()` functions declared in `<ctype.h>` should behave correctly -- which is in fact what currently happens 99% of the time. (Behaving correctly is just one possible consequence of undefined behavior.)

### Version of Vim

9.0.2018

### Environment

OS: Ubuntu 22.04.3 ; Windows 11
Terminal: xterm, tmux ; cmd.exe
[...]

PR #13347, opened 2023-10-15
Use SAFE_() macros for is*() and to*() functions

Keith-S-Thompson:
The standard is*() and to*() functions declared in <ctype.h> have undefined behavior for negative arguments other than EOF.  If plain char is signed, passing an unchecked value from argv for from user input to one of these functions has undefined behavior.

Solution: Add SAFE_*() macros that cast the argument to unsigned char.

Most implementations behave sanely for negative arguments, and most character values in practice are non-negative, but it's still best to avoid undefined behavior.

This addresses https://github.com/vim/vim/issues/13332.

### DIFF
--- a/runtime/tools/ccfilter.c
+++ b/runtime/tools/ccfilter.c
@@ -249,7 +249,7 @@ int main( int argc, char *argv[] )
 
 	    stay = (echogets(Line2, echo) != NULL);
 	    while ( stay && (Line2[0] == '|') )
-	      { for (p=&Line2[2]; (*p) && (SAFE_isspace(*p)); p++);
+	      { for (p=&Line2[2]; (*p) && (isspace((unsigned char)*p)); p++);
 		strcat( Reason, ": " );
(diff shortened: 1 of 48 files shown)


chrisbra:
Thanks, note however that appveyor build fails:

```
iscygpty.obj : error LNK2001: unresolved external symbol SAFE_isdigit
iscygpty.obj : error LNK2001: unresolved external symbol SAFE_isxdigit
vim.exe : fatal error LNK1120: 2 unresolved externals
```

chrisbra:
as I mentioned, I'll not immediately merge it, but wait with this after the vim9.1 release.

chrisbra:
thanks, this looks good so far.

chrisbra:
I am including this now, except for the change in `iscygpty.c`, which should have been fixed since commit https://github.com/vim/vim/commit/ac709e2fc0db6d31abb7da96f743c40956b60c3a

Please verify.

Thanks!

PR #13349, opened 2023-10-15
fix: remove the word being completed after completion restart

ghost:
Problem:
  The word being completed is shown in a completion list after
  completion restart, because it isn't removed from the current buffer
  before searching for matches:
  ```vim
    set complete=. completeopt=menuone
    call setline(1, 'workhorse workhorse')
    call feedkeys("$a\<C-n>\<BS>\<BS>", "t")
  ```

Solution:
  1. Remove a word being completed from the current buffer after restart
  2. Adjust `Test_complete_add_onechar` for the new behavior

Related https://github.com/neovim/neovim/issues/8480

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -4157,7 +4157,7 @@ ins_compl_next(
 	ins_compl_update_shown_match();
 
     if (allow_get_expansion && insert_match
-	    && (!(compl_get_longest || compl_restarting) || compl_used_match))
+	    && (!compl_get_longest || compl_used_match))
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks, I think this makes sense.

chrisbra:
you can have a look at the following test: https://github.com/vim/vim/blob/master/src/testdir/test_ins_complete.vim#L598-L613 it should be fairly similar.

chrisbra:
thanks

Issue #13343, opened 2023-10-14
[vim9script/vim9class] incorrect argument type checking involving varargs


errael:
### Steps to reproduce

This problem shows up in
- vim9script funcref assignment, this first example
- vim9class function overriding, the example in the first comment

The statement `Func = F2` in the following should produce an error since the arguments types are not the same.
Some of the problems that arise from this are demonstrated in the examples that follow.
```
vim9script

def F(...stuff: list<string>)
    echo stuff
enddef

var Func: func(list<string>)

Func = F            # <<< expect an error

def Test1()
    F('one')
    # Func('two')   # <<< compilation error
enddef

def Test2()
    Func(['two'])
    #F(['one'])      # <<< compilation error
enddef

Test1()
Test2()             # <<< runtime error
```

### Expected behaviour

See comments above

### Version of Vim

9.0.2026

### Environment

ubuntu

errael:
A similar problem shows up in `vim9class` overriding methods.
```
vim9script

class B
    this._val: string
    def F(...xxx: list<string>)
    enddef
endclass
class C extends B
    def F(xxx: list<string>)        # <<< expect compile error
    enddef
endclass

def F(o: B)
    o.F('a')
enddef

F(B.new())
F(C.new())          # <<< runtime error
```

errael:
@yegappan This seems to fix the problem (at least the ones in this report). But I'm not sure what other exposure there might be.

```
$ hg diff
diff --git a/src/vim9type.c b/src/vim9type.c
--- a/src/vim9type.c
+++ b/src/vim9type.c
@@ -884,6 +884,9 @@
                else
                    ret = MAYBE;
            }
+           if (ret != FAIL && (expected->tt_flags & TTFLAG_VARARGS)
+                                   != (actual->tt_flags & TTFLAG_VARARGS))
+               ret = FAIL;
            if (ret != FAIL && expected->tt_argcount != -1
                    && actual->tt_min_argcount != -1
                    && (actual->tt_argcount == -1
```

errael:
Oh, and the message is like:
```
E1383: Method "F": type mismatch, expected func(...list<string>) but got func(list<string>)
```

errael:
This change causes
```
Found errors in Test_varargs_mismatch():
```
It's not clear to me what that test is testing. The check might want to be less restrictive is `any` is involved, but matching varargs with non-varargs signature feels wrong.

The "right thing" is ??? 

[...]

PR #13351, opened 2023-10-16
Fix funcref assignment type checking involving `varargs` and `unknown`.

errael:
Note that this PR changes a test that passed into one that fails; it is `def Test_varargs_mismatch()`.
See https://github.com/vim/vim/issues/13343#issuecomment-1763486853, and surrounding comments, for discussion of this change in behavior. @yegappan , @dkearns comments?

There's an additional question; it has to do with funcref assignment and function argument type checking.

The following test passes, and I wonder if it should?
Since it involves `any`, I guess that's why it should pass

```
def Test_assign_funcref_arg_any()
  var lines =<< trim END
    vim9script
    var FuncAnyVA: func(any): number
    FuncAnyVA = (v): number => v
  END
  # TODO: Verify this should succeed.
  v9.CheckScriptSuccess(lines)
enddef
```

### DIFF
--- a/src/testdir/test_vim9_assign.vim
+++ b/src/testdir/test_vim9_assign.vim
@@ -1863,6 +1863,62 @@ def Test_assign_lambda()
   v9.CheckDefAndScriptFailure(lines, 'E1051:')
 enddef
 
+def Test_assign_funcref_args()
+  # unspecified arguments match everything, including varargs
+  var lines =<< trim END
+    vim9script
+
+    var FuncUnknown: func: number
+
+    FuncUnknown = (v): number => v
+    assert_equal(5, FuncUnknown(5))
+
+    FuncUnknown = (v1, v2): number => v1 + v2
+    assert_equal(7, FuncUnknown(3, 4))
+
+    FuncUnknown = (...v1): number => v1[0] + v1[1] + len(v1) * 1000
+    assert_equal(4007, FuncUnknown(3, 4, 5, 6))
+
+    FuncUnknown = (v: list<any>): number => v[0] + v[1] + len(v) * 1000
+    assert_equal(5009, FuncUnknown([4, 5, 6, 7, 8]))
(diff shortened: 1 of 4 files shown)


errael:
Add a few tests.

chrisbra:
@yegappan are you fine with including this?

yegappan:
> Note that this PR changes a test that passed into one that fails; it is `def Test_varargs_mismatch()`. See [#13343 (comment)](https://github.com/vim/vim/issues/13343#issuecomment-1763486853), and surrounding comments, for discussion of this change in behavior. @yegappan , @dkearns comments?
> 

Looks good to me.  The documentation should be updated to mention this.

> There's an additional question; it has to do with funcref assignment and function argument type checking.
> 
> The following test passes, and I wonder if it should? Since it involves `any`, I guess that's why it should pass
> 

Yes.  This is a valid use case and this should pass.

> ```
> def Test_assign_funcref_arg_any()
>   var lines =<< trim END
>     vim9script
>     var FuncAnyVA: func(any): number
>     FuncAnyVA = (v): number => v
>   END
>   # TODO: Verify this should succeed.
>   v9.CheckScriptSuccess(lines)
> enddef
> ```

chrisbra:
Thanks, I included it. @errael can you please add something to the documentation?

errael:
@chrisbra @yegappan There's an outstanding request for documentation. I'm not sure where it would go. Please advise.

Note that this change is basically a bug fix.

You can examine https://github.com/vim/vim/pull/13351#issuecomment-1766886578 to see some possibilities.

yegappan:
The updated help text should go into vim9.txt file.  Can you add the new text under "type-checking"?
We should mention that in a funcref assignment, the variable number of arguments is checked and
should match.  In addition to the above examples, it might be useful to include example funcref 
assignments that will fail.

PR #13353, opened 2023-10-16
Suppress GCC warning

k-takata:
Suppress the following warning:
```
filepath.c: In function ‘f_resolve’:
filepath.c:2162:27: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
 2162 |                     q[-1] = NUL;
```

Close #13352

### DIFF
--- a/src/filepath.c
+++ b/src/filepath.c
@@ -2159,7 +2159,7 @@ f_resolve(typval_T *argvars, typval_T *rettv)
 		if (q > p && *q == NUL)
 		{
 		    // Ignore trailing path separator.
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. I think it is not really a *suppress* but rather a fix :)

k-takata:
`q[-1]` and `p[q - p - 1]` point exactly the same address:
`p[q - p - 1]` == `*(p + (q - p - 1))` == `*(q - 1)` == `q[-1]`
But they are semantically different? not sure. 😄

PR #13358, opened 2023-10-16
trim(): Use default 'mask' when it is empty

illia-bobyr:
The default 'mask' value is pretty complex, as it includes many characters.  Yet, if one needs to specify the trimming direction, the third argument, 'trim()' currently requires the 'mask' value to be provided explicitly.

Currently, an empty 'mask' will make 'trim()' call return 'text' value that is passed in unmodified.  It is unlikely that someone is using it, so the chances of scripts being broken by this change are low.

---

An alternative approach is to use `v:none` instead of an empty string: https://github.com/vim/vim/pull/13363.

### DIFF
--- a/src/strings.c
+++ b/src/strings.c
@@ -1977,6 +1977,8 @@ f_trim(typval_T *argvars, typval_T *rettv)
     if (argvars[1].v_type == VAR_STRING)
     {
 	mask = tv_get_string_buf_chk(&argvars[1], buf2);
+	if (*mask == NUL)
+	    mask = NULL;
 
 	if (argvars[2].v_type != VAR_UNKNOWN)
 	{
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -10119,9 +10119,9 @@ trim({text} [, {mask} [, {dir}]])				*trim()*
 		Return {text} as a String where any character in {mask} is
 		removed from the beginning and/or end of {text}.
 
-		If {mask} is not given, {mask} is all characters up to 0x20,
-		which includes Tab, space, NL and CR, plus the non-breaking
-		space character 0xa0.
+		If {mask} is not given, or is an empty string, {mask} is all
+		characters up to 0x20, which includes Tab, space, NL and CR,
+		plus the non-breaking space character 0xa0.
 
 		The optional {dir} argument specifies where to remove the
 		characters:
--- a/src/testdir/test_functions.vim
+++ b/src/testdir/test_functions.vim
@@ -2236,6 +2236,10 @@ func Test_trim()
   let chars = join(map(range(1, 0x20) + [0xa0], {n -> n->nr2char()}), '')
   call assert_equal("x", trim(chars . "x" . chars))
 
+  call assert_equal("x", trim(chars . "x" . chars, '', 0))
+  call assert_equal("x" . chars, trim(chars . "x" . chars, '', 1))
+  call assert_equal(chars . "x", trim(chars . "x" . chars, '', 2))
+
   call assert_fails('let c=trim([])', 'E730:')
 endfunc
 


src/strings.c
```diff
@@ -1977,6 +1977,8 @@ f_trim(typval_T *argvars, typval_T *rettv)
     if (argvars[1].v_type == VAR_STRING)
     {
 	mask = tv_get_string_buf_chk(&argvars[1], buf2);
+	if (strlen(mask) == 0)
```
zeertzjq: Need to check for NULL (e.g. when passing a Dictionary which causes a type error)

zeertzjq:
```suggestion
	if (mask != NULL && STRLEN(mask) == 0)
```

illia-bobyr: The type of the argument is checked in the `if` statement two lines above.
It can only be `VAR_STRING`.

illia-bobyr: When the argument passed in has a string type, `tv_get_string_buf_chk` never returns a `NULL`.
Does it make sense to check, if it should never happen?

zeertzjq: Oops, didn't notice that.

zeertzjq (commented):
Ignore the previous comments

src/strings.c
illia-bobyr: No worries.
Thank you for looking into it!

src/strings.c
```diff
@@ -1977,6 +1977,8 @@ f_trim(typval_T *argvars, typval_T *rettv)
     if (argvars[1].v_type == VAR_STRING)
     {
 	mask = tv_get_string_buf_chk(&argvars[1], buf2);
+	if (STRLEN(mask) == 0)
```
yegappan: Instead of using STRLEN() function, you can also use `*mask == NUL` here.

illia-bobyr: Applied.

illia-bobyr: Seems like both GCC and clang generate identical code for both `strlen(str) == 0` and `*str == NUL` cases:

https://stackoverflow.com/a/67125598/1989046

I do not write C regularly, but to me, it seems that the `strlen(str) == 0` is a bit easier to read.
Do you still think `*mask == NUL` is preferred?

zeertzjq: `*mask == NUL` appears more commonly in the codebase.

errael:
Would `v:none` be more appropriate than an empty mask?

```
v:none		[...]
		This can also be used as a function argument to use the
		default value
```

ilya-bobyr:
> Would `v:none` be more appropriate than an empty mask?
> 
> ```
> v:none		[...]
> 		This can also be used as a function argument to use the
> 		default value
> ```

Here: https://github.com/vim/vim/pull/13363

It seems to make sense from the user standpoint.
But it would be the first time a built-in function uses `v:none` to mean "use default value".

errael:
> it would be the first time a built-in function uses `v:none` to mean "use default value"

I didn't know that.

chrisbra:
I have included #13363 so we can close this one then. Thanks!

Shane-XB-Qian:
> > Would `v:none` be more appropriate than an empty mask?
> > ```
> > v:none		[...]
> > 		This can also be used as a function argument to use the
> > 		default value
> > ```
> 
> Here: #13363
> 
> It seems to make sense from the user standpoint. But it would be the first time a built-in function uses `v:none` to mean "use default value".

looks vim's tradition or used to use `''` empty string to stand for 'default', cannot remember somewhere else had 'v:none'
if this is true, i think this is a bad start, which going to diverge or confuse, is it still possible to revert back to `''` ?

Shane-XB-Qian:
@ilya-bobyr please consider to revert to use `''` instead by reopen or another  PR.

errael:
For personal use, I don't care one way or the other. And as I mentioned, I was not aware `v:none` was not used with builtin functions; even for one of `v:none`s documented purposes in builtin functions.

But using `v:none` is much cleaner, safer, and consistent than the traditional method (magic value of given type); and an improvement to the language. I hope moving forward it can, and is, used to request the default value for builtin functions as well as user functions.

In this case the assumption is made that `mask` is hardcoded and not a computed value.

ilya-bobyr:
I also think that the fact that existing functions use a magic value to mean "default" is not the best argument.
It is a know antipattern, and if Vim9 has already introduced a better alternative, why not use it?

It has to start somewhere - one of those functions will be the first :)
So, maybe the question should not be "should an empty string be used", maybe it should be "is there a good reason not to use `v:none`"?

Shane-XB-Qian:
to me I am not interested in what's the difference between v:none v:null '' 0 zero or something else, and the difference between `is` == or something else.
to me as long as let me happy & ran performance to script is good enough.
to me as that's the tradition and if no good reason to change, (i donot think it is a magic value, here it is a string), then I think as vimer we should insist.
// but let me introduce another popular language's principle: 'less is more'.
// and I guess you are both from c++ or Java background :smile:

--
shane.xb.qian

errael:
> to me I am not interested in what's the difference between v:none v:null '' 0 zero or something else, and the difference between `is` == or something else. to me as long as let me happy & ran performance to script is good enough. 

I agree. The most important thing is being able to get the job done.

> to me as that's the tradition and if no good reason to change, (i donot think it is a magic value, here it is a string), then I think as vimer we should insist.

I see a distinction between `magic value` and `special value` as a parameter
- magic value
  if a parameter is normally a string, and one particular value of string means something entirely different, `''` in this case means `default value`, then I call that a magic value.
- special value
  a parameter that is not of the expected type that means something, in this case `v:none` means `default value`.

I don't like "magic value", and would rather have a "special value". A "magic value" can run into problems; for example if you want the parameter to mean its ordinary meaning and not its magic meaning.

> // but let me introduce another popular language's principle: 'less is more'. 
> // and I guess you are both from c++ or Java background 😄

// There's too many issues  there; a different discussion for another day ;-)

Shane-XB-Qian:
> > to me I am not interested in what's the difference between v:none v:null '' 0 zero or something else, and the difference between `is` == or something else. to me as long as let me happy & ran performance to script is good enough. 
> 
> I agree. The most important thing is being able to get the job done.

@errael um... donot get me wrong, there no job bother me.. ahaha..
i just say `''` is a tradition rule and simple&clear, so i recommend let's follow.
and have fun! anyway~ :smile:

-- 
shane.xb.qian

PR #13359, opened 2023-10-17
Use a simpler fix for Python 3.12 build issues

ychin:
Python 3.12 introduced link dependencies to their `Py_SIZE()` inline function, which #13290 fixed by copying the inline function to Vim's Python binding code. This works but it's fragile, as a future update may change the implementation of `Py_SIZE` and there is no way for us to know.

The reason we need `Py_SIZE` to begin with is that we use `PyTuple_GET_SIZE()` which calls that. Just fix it by mapping that to (confusingly similarly named) `PyTuple_Size()` (which is guaranteed to be stable), which we already do in the stable ABI implementation. There's a minor performance cost in that it's not inlined and it does error checking but that's fine as we only call this function rarely (in an error handler).


chrisbra:
thanks, appreciate the rework!

PR #13360, opened 2023-10-17
Not able to use object method as a callback function for popups and timers

yegappan:
After using an object method funcref, the object type is not removed from the type stack.
Remove the object type from the stack after getting the funcref.

While addressing this issue, I realized that the ISN_DEFEROBJ instruction introduced by
patch 9.0.1250 is no longer needed as the partial now has a pointer to the object.
The ISN_DEFER instruction can also handle deferring object methods.
So removed ISN_DEFEROBJ instruction.


chrisbra:
thanks!

Issue #11677, opened 2022-12-08
Using freed memory with wildoptions=pum and wildchar=<C-E>


zeertzjq:
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
set wildoptions=pum
set wildchar=<C-E>
call feedkeys(":\<C-E>\<C-E>", 'tx')
```
3. Crash

### Expected behaviour

No crash

### Version of Vim

9.0.1027

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

```
=================================================================
==134300==ERROR: AddressSanitizer: heap-use-after-free on address 0x602000082950 at pc 0x56377c4d069e bp 0x7ffcc7c3be40 sp 0x7ffcc7c3be30
READ of size 1 at 0x602000082950 thread T0
    #0 0x56377c4d069d in vim_strnsize **/src/charset.c:687
    #1 0x56377c4d0534 in vim_strsize **/src/charset.c:675
    #2 0x56377bd61abe in pum_compute_size **/src/popupmenu.c:61
    #3 0x56377bd63b7c in pum_display **/src/popupmenu.c:233
    #4 0x56377b737d50 in cmdline_pum_display **/src/cmdexpand.c:374
    #5 0x56377b73b116 in get_next_or_prev_match **/src/cmdexpand.c:771
    #6 0x56377b73ca19 in ExpandOne **/src/cmdexpand.c:964
    #7 0x56377b7362d8 in nextwild **/src/cmdexpand.c:244
    #8 0x56377b9aa27e in cmdline_wildchar_complete **/src/ex_getln.c:976
    #9 0x56377b9b08bd in getcmdline_int **/src/ex_getln.c:1971
    #10 0x56377b9adcd6 in getcmdline **/src/ex_getln.c:1552
    #11 0x56377b9b5bd4 in getexline **/src/ex_getln.c:2836
    #12 0x56377b92c241 in do_cmdline **/src/ex_docmd.c:877
    #13 0x56377bc6929d in nv_colon **/src/normal.c:3190
    #14 0x56377bc52e8e in normal_cmd **/src/normal.c:939
    #15 0x56377b97dab2 in exec_normal **/src/ex_docmd.c:8867
    #16 0x56377b896962 in f_feedkeys **/src/evalfunc.c:4446
    #17 0x56377b88a1b4 in call_internal_func **/src/evalfunc.c:3058
[...]

PR #13361, opened 2023-10-17
Fix wildchar/wildmenu/pum memory corruption with special wildchar's

ychin:
Currently, using `wildchar=<Esc>` or `wildchar=<C-\>` can lead to a memory corruption if using wildmenu+pum, or wrong states if only using wildmenu. This is due to the code only using one single place inside the cmdline process loop to perform wild menu clean up (by checking `end_wildmenu`) but there are other odd situations where the loop could have exited and we need a post-loop clean up just to be sure. If the clean up was not done you would have a stale popup menu referring to invalid memory, or if not using popup menu, incorrect status line (if `laststatus=0`).

For example, if you hit `<Esc>` two times when it's wildchar, there's a hard-coded behavior to exit command-line as a failsafe for user, and if you hit `<C-\><C-\><C-N>` it will also exit command-line, but the clean up code would not have hit because of specialized `<C-\>` handling.

Fix Ctrl-E / Ctrl-Y to not cancel/accept wildmenu if they are also used for 'wildchar'/'wildcharm'. Currently they don't behave properly, and also have potentially memory unsafe behavior as the logic is currently not accounting for this situation and try to do both. (Previous patch that addressed this: #11677)

Also, correctly document Escape key behavior (double-hit it to escape) in wildchar docs as it's previously undocumented.

In addition, block known invalid chars to be set in `wildchar` option, such as Ctrl-C and `<CR>`. This is just to make it clear to the user they shouldn't be set, and is not required for this bug fix.


ychin:
I found this when working on #13362. Since this fixes a memory corruption issue I decided to split into two PRs, but #13362 is the continuation of this. You could merge this first, or merge both together.

Shane-XB-Qian:
> hit `<Esc>` two times

have not looked into your PR deeply, but just a drop-by msg : 
what we discussed at #13331 that 'dismiss' i was wish it work like 'ctrl-l' in bash cli, which just to clean list choices,
but this double `<esc>` in vim cli is different, it totally cancel everything, the had input prompt chars was cleared too.

ychin:
> > hit `<Esc>` two times
> 
> have not looked into your PR deeply, but just a drop-by msg : what we discussed at #13331 that 'dismiss' i was wish it work like 'ctrl-l' in bash cli, which just to clean list choices, but this double `<esc>` in vim cli is different, it totally cancel everything, the had input prompt chars was cleared too.

I think that's what Ctrl-E/Ctrl-Y is for? This double <Esc> behavior is a very special case where it's used as wildchar (which I honestly don't understand why anyone would set it this way to be honest). <Esc> is usually for escaping the command-line mode anyway and I don't think it's a useful key to overload with more meaning.

ychin:
Or are you talking about not using wildmenu, and something like the list choices displayed in Ctrl-D or `wildmode=list` like you mentioned in the other PR? I think that's a different discussion from what I'm trying to fix here. Either way I don't think `<Esc>` should be used for that purpose as the key has very special meaning in Vim.

Shane-XB-Qian:
i was trying to help clarify the different, or if you can make 'c-e' work on wildmode=list which make its list choice be dismissed by 'c-e' when it presented like wildmune or pum would be great, but didnot expect it have to change.
and yes, `<esc>` is a special key, and you're a TUI guy you may feel it's a similar behavior like others tui tools, sometime it worked like 'alt', that's another reason i jumpped in, wish this not changed, it is history :smile:

ychin:
I think what I don't understand is what you mean by "dismiss"  wildmode=list? Recall that when you have wildmode=list, the completion does *not* fill in the choice. Or did you mean wildmost=full (in which case it's true that the current change would not make that work)?

Shane-XB-Qian:
> I think what I don't understand is what you mean by "dismissing"  wildmode=list? Recall that when you have wildmode=list, the completion does *not* fill in the choice. Or did you mean wildmost=full (in which case it's true that the current change would not make that work)?

yes, list and/with full which had said that option in https://github.com/vim/vim/pull/13331#issuecomment-1761223994

i think these 3 list/wildmenu/pum are similar cases, now this PR you're trying to make it work for 'wildmenu' (and 'c-e' worked for 'pum' already)
// agagin, i didnot expect you have to make it work for 'list', but it would be nice (enhancement) one :simile: it would work like bash cli if so.

-- 
shane.xb.qian

ychin:
Yeah it could be a future enhancement but I have to point out `list` is very different from `full`. You were saying that it takes up too much screen real estate when using list, and the issue there is more one of redraw, and something you will also see when using Ex-mode and `Ctrl-D` as I mentioned. I think `full` could be implemented as an enhancement though.

If you want to add a clear for `wildmode=list`, you could just map the following (Ctrl-O in this example):

```vim
cnoremap <C-O> <C-\>e<SID>CmdlineRedraw()<CR>
function! s:CmdlineRedraw()
    let c = getcmdline()
    redraw
    return c
endfunction
```

Shane-XB-Qian:
ok, i think you got my point, and yea, some kind of vim script could make it work also.
so, natively support would be great, but 'not' also is ok, that's why i say didnot expect it have to change.

-- 
shane.xb.qian

chrisbra:
thanks!

PR #13362, opened 2023-10-17
Fix cmdline completion fuzzy/Ctrl-E/Ctrl-Y/options when not used at the end

ychin:
A few places in the cmdline completion code only works properly when the user hits Tab (or 'wildchar') at the end of the cmdline, even though it's supposed to work even in the middle of the line.

For fuzzy search, `:e ++ff`, and `:set hl=`, fix completion code to make sure to use `xp_pattern_len` instead of assuming the entire `xp_pattern` is the search pattern (since it contains texts after the cursor).

Fix Ctrl-E / Ctrl-Y to not jump to the end when canceling/accepting a wildmenu completion. Also, make them work even when not using `set wildoptions+=pum` as there is no drawback to doing so.  (Related issue where this was brought up: #13331)


ychin:
This PR is based on #13361. I decided to split into 2 separate PRs because that PR fixes a more important memory corruption issue and fixes a different issue than this one which is more a simple enhancement. Just sending both PRs out to make it see what the total change is. You could combine the two together if you want.

PR #13363, opened 2023-10-17
trim(): Use default 'mask' when it is v:none

illia-bobyr:
The default 'mask' value is pretty complex, as it includes many characters.  Yet, if one needs to specify the trimming direction, the third argument, 'trim()' currently requires the 'mask' value to be provided explicitly.

'v:none' is already used to mean "use the default argument value" in user defined functions.  See |none-function_argument| in help.

---

An alternative would be to use an empty string, instead of `v:none`: https://github.com/vim/vim/pull/13358.

### DIFF
--- a/src/typval.c
+++ b/src/typval.c
@@ -451,6 +451,37 @@ check_for_opt_string_arg(typval_T *args, int idx)
 }
 
 /*
+ * Check for an optional string argument at 'idx', that can also be 'v:none' to
+ * use the default value.
+ *
+ * If 'is_none' is non-NULL it is set to 0 and updated to 1 when "args[idx]" is
+ * 'v:none'.
+ */
+    int
+check_for_opt_string_or_none_arg(typval_T *args, int idx, int *is_none)
+{
+    if (is_none != NULL)
+	*is_none = 0;
+
+    if (args[idx].v_type == VAR_UNKNOWN)
+	return OK;
+
+    if (args[idx].v_type == VAR_SPECIAL
+	    && args[idx].vval.v_number == VVAL_NONE)
+    {
+	if (is_none != NULL)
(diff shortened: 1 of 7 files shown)


src/typval.c
```diff
@@ -450,6 +450,37 @@ check_for_opt_string_arg(typval_T *args, int idx)
...
 
+/*
+ * Check for an optional string argument at 'idx', that can also be 'v:none' to
+ * use the default value.
+ *
+ * If 'is_none' is non-NULL it is set to 0 and updated to 1 when "args[idx]" is
+ * 'v:none'.
+ */
+    int
+check_for_opt_string_or_none_arg(typval_T *args, int idx, int *is_none)
```
chrisbra, line 461: this `int *is_none` seems to be unused.

illia-bobyr: I used it at first, but then, it turned out to be unnecessary.

It seems that distinguishing between the "value given" and "use default" cases could be important.
So, I figured, the next use of this helper function could actually benefit from the `is_none` flag, and kept it.
Sorry, I should have mentioned it.

I can remove it, if you think that it can be added when an actual need will arise.

chrisbra: yeah, that's what I thought. Thanks

Shane-XB-Qian:
> I think `v:none` is fine, even so I personally find the name `none` to be slightly misleading. What do others think?

@chrisbra https://github.com/vim/vim/pull/13358#issuecomment-1766148524

yegappan:
> I think `v:none` is fine, even so I personally find the name `none` to be slightly misleading. What do others think?

Other built-in functions use an empty string to refer to a default value.  So it makes sense to support that here.
If the `mask` is an empty string, then the default value should be used.

PR #13369, opened 2023-10-17
win32,perl: Suppress warnings

k-takata:
Suppress warning C4273 (inconsistent DLL linkage).

Also adjust inconsistent use of const.


chrisbra:
thanks!

PR #13370, opened 2023-10-18
win32: Suppress MS_WIN64 redefined warning

k-takata:
Suppress the following warning:
```
In file included from C:/Python312-64/include/Python.h:12,
                 from if_python3.c:79:
C:/Python312-64/include/pyconfig.h:240: warning: "MS_WIN64" redefined
  240 | #define MS_WIN64
      |
<command-line>: note: this is the location of the previous definition
```

`MS_WIN64` is not defined in `Make_mvc.mak` either. This should be able to be safely removed.

### DIFF
--- a/src/Make_cyg_ming.mak
+++ b/src/Make_cyg_ming.mak
@@ -513,9 +513,6 @@ endif # RUBY
 DEF_GUI=-DFEAT_GUI_MSWIN -DFEAT_CLIPBOARD
 DEFINES=-DWIN32 -DWINVER=$(WINVER) -D_WIN32_WINNT=$(WINVER) \
 	-DHAVE_PATHDEF -DFEAT_$(FEATURES) -DHAVE_STDINT_H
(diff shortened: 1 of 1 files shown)


chrisbra:
Hm, this seems to come from your patch [v7.3.1313](https://github.com/vim/vim/commit/254e028e907d8f261f8911242eb0c)

https://github.com/vim/vim/blob/a0e1f06f04da3444e278ddf47e2ea3d5857a7dec/runtime/doc/version7.txt#L17912-L17916

Let me include it anyhow, thanks!

k-takata:
> Hm, this seems to come from your patch [v7.3.1313](https://github.com/vim/vim/commit/254e028e907d8f261f8911242eb0c)

Aha, I totally forgot that! 😄

k-takata:
It seems that this was needed 10 years ago: https://groups.google.com/g/vim_dev/c/xmtbXiE25fA/m/MJ7tRmsZCdcJ

PR #13371, opened 2023-10-18
Don't check for swap files if test is skipped

zeertzjq:
Check for swap files takes a considerable about of time, so don't do
that for skipped tests to avoid wasting time.


zeertzjq:
Other clean up steps may be skipped as well, but checking for swap files is the one that takes the most time.

chrisbra:
thanks!

PR #13372, opened 2023-10-18
Deferred function not fully executed after an exception

### LINKED ISSUE #13364: Unexpected `:defer`, `throw`, `catch` interactions.
### Steps to reproduce

In legacy script, if an exception that triggers a defer is caught then only the first line of the deferred function is executed. If the exception is not caught, then all the lines are executed.

Consider this script.
```
func Defer()
  echo "one"
  echo "two"
  echo "three"
endfunc

func Foo()
  defer Defer()
  throw "OOPS"
endfunc

" comment out these two lines to see the problem
call Foo()
finish

" execute this and only the first line of "Defer()" is executed
try
  call Foo()
catch
  echo v:exception
endtry
```
Run the script and you'll get this output.
```
Error detected while processing :source buffer=1[13]..function Foo[2]..:source buffer=1[13]..functio
n Foo:
line    2:
E605: Exception not caught: OOPS
one
two
three
```
Coment out the first `call Foo()` and `finish` and you'll see this output
```
one
OOPS
```

### Expected behaviour

The defer should fully execute before the exception is processed.

### Version of Vim

9.0.2032

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #13364

### DIFF
--- a/src/testdir/test_user_func.vim
+++ b/src/testdir/test_user_func.vim
@@ -870,5 +870,31 @@ func Test_defer_wrong_arguments()
   call v9.CheckScriptFailure(lines, 'E1013: Argument 1: type mismatch, expected string but got number')
 endfunc
 
+" Test for calling a deferred function after an exception
+func Test_defer_after_exception()
+  let g:callTrace = []
+  func Defer()
+    let g:callTrace += ['a']
+    let g:callTrace += ['b']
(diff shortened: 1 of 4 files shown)


chrisbra:
thanks!

PR #13375, opened 2023-10-18
draft: Trigger TextChangedI for 'o/O' in normal mode

### LINKED ISSUE #13367: None of `TextChanged` or `TextChangedI` is triggered after either `o` or `O`
### Steps to reproduce

1. Create a config file with the following content:
```vim
let g:n = 0
au TextChanged,TextChangedI * let g:n = g:n + 1
```
2. Start `vim` with that config.
3. Press `o<Esc>` several times. Then press `O<Esc>` several times.
4. Execute `:echo g:n`. It shows 0.

### Expected behaviour

Either `TextChanged` or `TextChangedI` should be triggered after both `o` and `O`.

Judging by their help description ("Begin a new line below/above the cursor and insert text"), I'd say that this is equivalent to making change in Normal mode and then starting Insert mode on target line. So `TextChanged` here seems to be more suitable.

On the other hand, the final state is Insert mode, so maybe `TextChangedI` is more suitable.

**Note**: Executing `A<CR>` instead of `o` works as expected.

------

This is most likely a regression after #13338.

### Version of Vim

9.0.2040

### Environment

Operating system: EndeavourOS Linux x86_64 (6.5.7-arch1-1)
Terminal: st-0.9
Value of $TERM: st-256color
Shell: zsh 5.9

### Logs and stack traces

_No response_

chrisbra:
draft only for now

fixes: #13367

todo: needs to verified and add some more tests

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -174,9 +174,6 @@ edit(
 	return FALSE;
     }
     ins_compl_clear();	    // clear stuff for CTRL-X mode
-    // Reset Changedtick_i, so that TextChangedI will only be triggered for stuff
-    // from insert mode
-    curbuf->b_last_changedtick_i = CHANGEDTICK(curbuf);
 
     /*
      * Trigger InsertEnter autocommands.  Do not do this for "r<CR>" or "grx".
--- a/src/normal.c
+++ b/src/normal.c
@@ -6292,6 +6292,8 @@ n_opencmd(cmdarg_T *cap)
 	(void)hasFolding(curwin->w_cursor.lnum,
 		NULL, &curwin->w_cursor.lnum);
 #endif
+    // trigger TextChangedI for the 'o/O' command
+    curbuf->b_last_changedtick_i = CHANGEDTICK(curbuf);
     if (u_save((linenr_T)(curwin->w_cursor.lnum -
 		    (cap->cmdchar == 'O' ? 1 : 0)),
 		(linenr_T)(curwin->w_cursor.lnum +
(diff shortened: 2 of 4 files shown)


echasnovski:
@chrisbra, I've tested it and everything seems to work.

I've tried writing tests for all this "Normal and Insert modes mix" cases. Here is the patch:
```diff
diff --git a/src/testdir/test_autocmd.vim b/src/testdir/test_autocmd.vim
index cdac84273..80b6cef3d 100644
--- a/src/testdir/test_autocmd.vim
+++ b/src/testdir/test_autocmd.vim
@@ -3620,6 +3620,24 @@ func Test_Changed_ChangedI()
   call feedkeys("ibar\<esc>", 'tnix')
   call assert_equal('', g:autocmd_n)

+  " If change is a mix of Normal and Insert modes, TextChangedI should trigger
+  func s:validate_mixed_textchangedi(keys)
+    call feedkeys("ifoo\<esc>", 'tnix')
+    let g:autocmd_i = ''
+    let g:autocmd_n = ''
+    call feedkeys(a:keys, 'tnix')
+    call assert_notequal('', g:autocmd_i)
+    call assert_equal('', g:autocmd_n)
+  endfunc
+
+  call s:validate_mixed_textchangedi("o\<esc>")
+  call s:validate_mixed_textchangedi("O\<esc>")
+  call s:validate_mixed_textchangedi("ciw\<esc>")
+  call s:validate_mixed_textchangedi("cc\<esc>")
+  call s:validate_mixed_textchangedi("C\<esc>")
+  call s:validate_mixed_textchangedi("s\<esc>")
+  call s:validate_mixed_textchangedi("S\<esc>")
+
   " CleanUp
   call test_override("char_avail", 0)
   au! TextChanged  <buffer>
```

It seems to pass both on 56b2aeec0086e68de86c9717edbeda43527b974b and 63afbbef7511355f128498be615d8f48d829c9a1, i.e. I couldn't make them break by reverting changes at 56b2aeec. So these might not cover all scenarios.

zeertzjq:
`feedkeys()` doesn't invoke `main_loop()`, so it never triggers `TextChanged`.

echasnovski:
> `feedkeys()` doesn't invoke `main_loop()`, so it never triggers `TextChanged`.

Yeah, it seems to be an issue [even prior to this update](https://github.com/vim/vim/blob/5f5131d775bf9966976e39aa38b070036cbfe969/src/testdir/test_autocmd.vim#L3604-L3606). Completely unaware what to do here.

But tests in patch seem to still be good as they concern Insert mode and are breakable.

PR #13376, opened 2023-10-18
Crash when a deferred function is called after an exception and another exception is thrown

### LINKED ISSUE #13377: SEGV in :defer handling after exception in deferred function
### Steps to reproduce

The following gives a SEGV
```
func InnerExcept()
  echo "InnerExcept"
  throw "INNER-OOPS"
endfunc

func Defer()
  echo "one"
  echo "two"
  call InnerExcept()
  echo "three"
  echo "four"
endfunc

func Foo()
  defer Defer()
  throw "OOPS"
endfunc

" comment out these two lines to see the problem
"call Foo()
"finish

" execute this and only the first line of "Defer()" is executed
try
  call Foo()
catch
  echo v:exception
endtry
```

### Expected behaviour

This is probably the most mysterious (at least to me) part of vim[9]script programs; and since this issue encompasses any scripting old/new. I still have no clue, but I saw an unused field.

In the docs
```
Errors are reported but do not cause aborting execution of deferred functions.
```

Here's a few things of interest that I saw.

**execution stack**
- a grow array, exestack.ga_*, in scriptfile
- `scriptfile.c` has `estack_*` functions used by `vim9execute` and more
- In do_cmd.c there's
  `estack_push(ETYPE_EXCEPT, current_exception->throw_name, current_exception->throw_lnum)`
- estack_T has a `union{..., except_T *except}` but I **can't find a reference to except.**

**ex_eval.c**

There's a variety of exception handling/posting/state functions in here.

### Version of Vim

9.0.2048

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Save and restore the current_exception state.  Fixes #13377.

### DIFF
--- a/src/ex_eval.c
+++ b/src/ex_eval.c
@@ -748,6 +748,42 @@ finish_exception(except_T *excp)
 }
 
 /*
+ * Save the current exception state in "estate"
+ */
+    void
+exception_state_save(exception_state_T *estate)
+{
+    estate->estate_current_exception = current_exception;
+    estate->estate_did_throw = did_throw;
+    estate->estate_need_rethrow = need_rethrow;
+    estate->estate_trylevel = trylevel;
+}
+
+/*
+ * Restore the current exception state from "estate"
+ */
+    void
+exception_state_restore(exception_state_T *estate)
+{
+    current_exception = estate->estate_current_exception;
+    did_throw = estate->estate_did_throw;
+    need_rethrow = estate->estate_need_rethrow;
+    trylevel = estate->estate_trylevel;
+}
+
+/*
+ * Clear the current exception state
+ */
+    void
+exception_state_clear(void)
+{
+    current_exception = NULL;
+    did_throw = FALSE;
+    need_rethrow = FALSE;
+    trylevel = 0;
+}
+
+/*
  * Flags specifying the message displayed by report_pending.
  */
 #define RP_MAKE		0
--- a/src/ex_eval.c
+++ b/src/ex_eval.c
@@ -765,10 +765,11 @@ exception_state_save(exception_state_T *estate)
     void
 exception_state_restore(exception_state_T *estate)
 {
-    current_exception = estate->estate_current_exception;
-    did_throw = estate->estate_did_throw;
-    need_rethrow = estate->estate_need_rethrow;
-    trylevel = estate->estate_trylevel;
+    if (current_exception == NULL)
+	current_exception = estate->estate_current_exception;
+    did_throw |= estate->estate_did_throw;
+    need_rethrow |= estate->estate_need_rethrow;
+    trylevel |= estate->estate_trylevel;
 }
 
 /*
--- a/src/proto/ex_eval.pro
+++ b/src/proto/ex_eval.pro
@@ -11,6 +11,9 @@ char *get_exception_string(void *value, except_type_T type, char_u *cmdname, int
 int throw_exception(void *value, except_type_T type, char_u *cmdname);
 void discard_current_exception(void);
 void catch_exception(except_T *excp);
+void exception_state_save(exception_state_T *estate);
+void exception_state_restore(exception_state_T *estate);
+void exception_state_clear(void);
 void report_make_pending(int pending, void *value);
 int cmd_is_name_only(char_u *arg);
 void ex_eval(exarg_T *eap);
--- a/src/structs.h
+++ b/src/structs.h
@@ -1088,6 +1088,19 @@ struct cleanup_stuff
     except_T *exception;	// exception value
 };
 
+/*
+ * Exception state that is saved and restored when calling timer callback
+ * functions and deferred functions.
+ */
+typedef struct exception_state_S exception_state_T;
+struct exception_state_S
+{
+    except_T	*estate_current_exception;
+    int		estate_did_throw;
+    int		estate_need_rethrow;
+    int		estate_trylevel;
+};
+
 #ifdef FEAT_SYN_HL
 // struct passed to in_id_list()
 struct sp_syn
--- a/src/testdir/test_user_func.vim
+++ b/src/testdir/test_user_func.vim
@@ -873,11 +873,21 @@ endfunc
 " Test for calling a deferred function after an exception
 func Test_defer_after_exception()
   let g:callTrace = []
+  func Bar()
+    let g:callTrace += [1]
+    throw 'InnerException'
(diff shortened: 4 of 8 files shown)


errael:
Key statement from the doc
```
A much simpler solution [instead of `try`/`finally`] is using `defer`.
Errors are reported but do not cause aborting execution of deferred functions.
```

The following script aborts execution of the deferred function. 
It calls `Bar()` without it being in a `try`.

```
let g:callTrace = []
func Bar()
  let g:callTrace += [1]
  throw 'InnerException'
endfunc

func Defer()
  let g:callTrace += [2]
  let g:callTrace += [3]

  call Bar()

  let g:callTrace += [5]
  let g:callTrace += [6]
endfunc

func Foo()
  defer Defer()
  throw "TestException"
endfunc

try
  call Foo()
catch /TestException/
  let g:callTrace += [7]
endtry

"call assert_equal([2, 3, 1, 5, 6, 7], g:callTrace)
echo g:callTrace

delfunc Defer
delfunc Foo
unlet g:callTrace
```

errael:
BTW, a similar vim9 test fails in the same way.

yegappan:
> Key statement from the doc
> 
> ```
> A much simpler solution [instead of `try`/`finally`] is using `defer`.
> Errors are reported but do not cause aborting execution of deferred functions.
> ```
> 

I don't think this statement means that all the exceptions are ignored when executing the deferred functions.
If an exception is encountered when running a deferred function and is not caught, then the execution
of that function will be aborted.

For example, consider the following two examples (one using try/catch and the other using defer):

```
func Foo()
    try
        echo 1
        throw "foo"
        echo 2
    catch
        echo 3
        throw "bar"
        echo 4
    finally
        echo 5
        throw "baz"
        echo 6
    endtry
endfunc

try
    echo 7
    call Foo()
    echo 8
catch
    echo 9
finally
    echo 10
endtry
```

The above example, will echo "7 1 3 5 9 10".  The statements following an exception in a function are not executed.

```
func Fn1()
    echo 1
    throw "foo"
    echo 2
endfunc

func Fn2()
    echo 3
    throw "foo"
    echo 4
endfunc

func Fn3()
    echo 5
    throw "foo"
    echo 6
endfunc

func Foo()
    echo 7
    defer Fn1()
    defer Fn2()
    defer Fn3()
    echo 8
    throw "foo"
    echo 9
endfunc

try
    echo 10
    call Foo()
    echo 11
catch
    echo 12
finally
    echo 13
endtry

echo 14
```

The above example, will echo "10 7 8 5 3 1 12 13 14" (ignoring the exceptions).  As you can
see when an exception is encountered, the following statements will not be executed.

errael:
```
Errors ... do not cause aborting execution of deferred functions.
```

I don't see how that could be interpreted to mean the execution of a deferred
function should be aborted.

**Ohhhhh**, unless that final `s` is the key to understanding.

Maybe some clarification would be good.
```
A deferred function that gets an error is aborted 
and processing continues with the next deferred function.
```
Which does work.

errael:
There is still a minor problem, the following script outputs 
```
[30, 31, 20, 9990, 10, 11, 7]
```
as expected (as I've come to understand). But it does not report the "InnerException"
```
vim9script

var callTrace: list<number> = []
def Bar()
  callTrace += [9990]
  throw 'InnerException'
enddef

def Defer(tag: number)
  callTrace += [tag]
  callTrace += [tag + 1]
enddef

def DeferWithExcept(tag: number)
  callTrace += [tag]
  Bar()
  callTrace += [tag + 1]
enddef


def Foo()
  defer Defer(10)
  defer DeferWithExcept(20)
  defer Defer(30)
  throw "TestException"
enddef

try
  Foo()
catch /TestException/
  callTrace += [7]
endtry

echo callTrace
```

errael:
I wonder if the design of `:defer`/`excpeption` interactions was never complete. I did see that unused field `except_T except` in `estack_T`. And I've always been curious as to why, in 
`Errors are reported but do not cause aborting execution of deferred functions.`
the word "reported" is used.

I ask because two exceptions are thrown, but only one is reported. The first excepetion is lost. Simplify output by getting rid of the `try/catch/finally` and just running Foo().

I wonder if an exception in a deferred function should only terminate the function and produce output but not propagate as an exception outside of the deferred function; kind of like the deferred functions are run as separate threads, an exception in one thread does not affect the execution of others. That seems to fit with the idea that the current deferred function is aborted, and then the other deferred functions are executed.

The following using the "lost exception" commit.

Note: `&verbose = 14`
```
Throwing TestException
Exception thrown: TestException
Exception thrown: InnerException
Exception made pending: InnerException
finally [30, 31, 20, 9990, 10, 11]
v:exception:
Exception resumed: InnerException
Exception discarded: InnerException
Error detected while processing :source buffer=1[38]..:source buffer=1[32]..function <SNR>11_Foo[5].
.<SNR>11_DeferWithExcept[2]..<SNR>11_Bar:
line    2:
E605: Exception not caught: InnerException
```

```
vim9script

&verbose = 14

var callTrace: list<number> = []
def Bar()
  callTrace += [9990]
  throw 'InnerException'
enddef

def Defer(tag: number)
  callTrace += [tag]
  callTrace += [tag + 1]
enddef

def DeferWithExcept(tag: number)
  callTrace += [tag]
  Bar()
  callTrace += [tag + 1]
enddef


def Foo()
  defer Defer(10)
  defer DeferWithExcept(20)
  defer Defer(30)
  echo "Throwing TestException"
  throw "TestException"
enddef

try
  Foo()
catch /TestException/
  callTrace += [7]
finally
  echo "finally" callTrace
  echo "v:exception:" v:exception
endtry

echo "finished" callTrace
```

chrisbra:
thanks!

src/ex_eval.c
```diff
@@ -747,6 +747,43 @@ finish_exception(except_T *excp)
...
+ * Restore the current exception state from "estate"
+ */
+    void
+exception_state_restore(exception_state_T *estate)
+{
+    if (current_exception == NULL)
+	current_exception = estate->estate_current_exception;
+    did_throw |= estate->estate_did_throw;
+    need_rethrow |= estate->estate_need_rethrow;
+    trylevel |= estate->estate_trylevel;
```
zeertzjq, line 772: Performing a bitwise operation on `trylevel` looks strange.

yegappan: That is a typo.  I will fix that in the next PR (#13386).

PR #13380, opened 2023-10-19
Add digraph for quadruple prime

quag:
Unicode 1.1.0 (1993) added single primes (′), double primes (″), and triple primes (‴). It wasn't until Unicode 3.2.0 (2002) that quadruple primes (⁗) were added. Vim has the digraphs '1, '2, '3 for the earlier primes, but does not have '4 for the quadruple primes.

This patch is my first attempted submission to vim. Sorry if I have blundered, should use a different approach, or if this sort of change is unwelcome. Thank you for your patience.


chrisbra:
Thanks. It looks like the quadruple prime is not part of RFC1345 (probably that was defined before the Unicode added it), but I guess that should be acceptable (we have added the Rouble digraph before, which also isn't part of the RFC).
So can you please add a test into  `src/testdir/test_digraphs.vim`, check the [Test_digraphs()](https://github.com/vim/vim/blob/master/src/testdir/test_digraph.vim#L15) function, it should be self explanatory.

quag:
> So can you please add a test into `src/testdir/test_digraphs.vim`, check the [Test_digraphs()](https://github.com/vim/vim/blob/master/src/testdir/test_digraph.vim#L15) function, it should be self explanatory.

Done. I added a test for quadruple prime to src/testdir/test_digraph.vim. Are there other things I should also change?

chrisbra:
thanks, looks good now.

PR #13381, opened 2023-10-19
MatchParen: Suppress match ID in-use error

KSR-Yasuda:
Sometimes it raises E801, ID in-use error.

I don't have any stable procedure to repro,
but am likely to see this error in using [surround.vim](https://github.com/tpope/vim-surround) plugin.

### DIFF
--- a/runtime/plugin/matchparen.vim
+++ b/runtime/plugin/matchparen.vim
@@ -186,7 +186,7 @@ func s:Highlight_Matching_Pair()
   " If a match is found setup match highlighting.
   if m_lnum > 0 && m_lnum >= stoplinetop && m_lnum <= stoplinebottom 
     if exists('*matchaddpos')
-      call matchaddpos('MatchParen', [[c_lnum, c_col - before], [m_lnum, m_col]], 10, 3)
+      silent! call matchaddpos('MatchParen', [[c_lnum, c_col - before], [m_lnum, m_col]], 10, 3)
     else
       exe '3match MatchParen /\(\%' . c_lnum . 'l\%' . (c_col - before) .
 	    \ 'c\)\|\(\%' . m_lnum . 'l\%' . m_col . 'c\)/'


bfrg:
I've run into the same error in the past. The issue is that the plugin always uses a matchid 3 even when using `matchaddpos()`. IDs 1, 2 and 3 are actually reserved for `:match`, `:2match` and `:3match`, respectively. This means if you (or some plugin) manually called `:3match`, it cannot be used in `matchaddpos()` anymore, or  you'll get the above error.

The help for the plugin under `:h mathparen` doesn't say anything that, when using the plugin, `:3match` cannot be used anymore.

I think, since the plugin already uses `matchaddpos()`, it should just pass -1 as the `{id}` parameter to `matchaddpos()` and rely on the auto-generated id instead of hard-coding 3. The id returned by `matchaddpos()` needs to be saved in a window-local variable, and can later be used to remove the highlighting with `matchdelete()`.

KSR-Yasuda:
I'm not explicitly using match ID 3, and I think surround.vim is not either.

I doubt there is some rare case that matchparen plugin reuses ID 3 before it clears the previous paren match
(I'm not sure what is the trigger...).

Anyway, I'll check ID 3 match in the error case with putting some log there.

KSR-Yasuda:
Otherwise, it may be better to force overwriting match ID 3,
as [the else-case](https://github.com/vim/vim/blob/master/runtime/plugin/matchparen.vim#L191) that is calling `:3match` command instead of `matchaddpos()`.

```diff
--- a/usr/share/vim/vim90/plugin/matchparen.vim
+++ b/usr/share/vim/vim90/plugin/matchparen.vim
@@ -186,7 +186,8 @@ func s:Highlight_Matching_Pair()
   " If a match is found setup match highlighting.
   if m_lnum > 0 && m_lnum >= stoplinetop && m_lnum <= stoplinebottom 
     if exists('*matchaddpos')
-      call matchaddpos('MatchParen', [[c_lnum, c_col - before], [m_lnum, m_col]], 10, 3)
+      silent! call matchdelete(3)
+      call matchaddpos('MatchParen', [[c_lnum, c_col - before], [m_lnum, m_col]], 10, 3)
     else
       exe '3match MatchParen /\(\%' . c_lnum . 'l\%' . (c_col - before) .
 	    \ 'c\)\|\(\%' . m_lnum . 'l\%' . m_col . 'c\)/'
```

chrisbra:
> I think, since the plugin already uses matchaddpos(), it should just pass -1 as the {id} parameter to matchaddpos() and rely on the auto-generated id instead of hard-coding 3. The id returned by matchaddpos() needs to be saved in a window-local variable, and can later be used to remove the highlighting with matchdelete().

Yeah, there is something to be improved. I have created #13393 for this.

Shane-XB-Qian:
> > I think, since the plugin already uses matchaddpos(), it should just pass -1 as the {id} parameter to matchaddpos() and rely on the auto-generated id instead of hard-coding 3. The id returned by matchaddpos() needs to be saved in a window-local variable, and can later be used to remove the highlighting with matchdelete().
> 
> Yeah, there is something to be improved. I have created #13393 for this.

um.. my first feeling is 3 is the reserved id specifically for matchparen.
that err should be one user's specific case, then why not he/she debug find out which one accidentally take 3?
// vs we should/can change the common used way/id 3?

-- 
shane.xb.qian

chrisbra:
> um.. my first feeling is 3 is the reserved id specifically for matchparen.
that err should be one user's specific case, then why not he/she debug find out which one accidentally take 3?
// vs we should/can change the common used way/id 3?

Well, no. 3 was used specifically for the matchparen plugin (using the `:3match` command). But that was mainly because back then, we did not have the `matchadd()` family of functions. I see no reason, why we should use old ex-commands, when we can use vimscript functions that provide more flexibility.

Shane-XB-Qian:
> > um.. my first feeling is 3 is the reserved id specifically for matchparen.
> that err should be one user's specific case, then why not he/she debug find out which one accidentally take 3?
> // vs we should/can change the common used way/id 3?
> 
> Well, no. 3 was used specifically for the matchparen plugin (using the `:3match` command). But that was mainly because back then, we did not have the `matchadd()` family of functions. I see no reason, why we should use old ex-commands, when we can use vimscript functions that provide more flexibility.

ok, but maybe more interesting is why someone or some plugin would use '3match' other than matchparen,
i would think that's his/her mistake by accident,
so compared to change at vim repo side (your PR) perhaps better to find that specific case, and fix that.

-- 
shane.xb.qian

chrisbra:
yes maybe more insteresting, but I rather make sure the problem does no longer occur. Why not make use of existing vimscript functions?

bfrg:
> ok, but maybe more interesting is why someone or some plugin would use '3match' other than matchparen,
i would think that's his/her mistake by accident,

Well, sometimes I need to highlight something very quickly. Often I use `:match`, `:2match` and `:3match` for that. If matchparen is enabled eventually I will get the same error that was reported above. There's nothing wrong with using `:match` interactively. So I think improving the matchparen plugin to use any available ID makes sense.

Shane-XB-Qian:
to me if the doc said that was reserved for mathparen then it is.
and it had reserved match and 2match for temp usage or quickly to hl something already.
one may argue two is not enough, um.. why three is enough?

--
shane.xb.qian

chrisbra:
it's not, especialy when using plugins. That's why Vim eventually introduced the matchadd functionality. And I think it makes sense, to use the matchadd() functions for plugins like , ehm the matchparen plugin and leave the `:match` ex commands more for interactive use without disturbing the user too much. (and they are slightly more easier to use). 

I am not sure what problems you are actualy facing with improving the plugin?

Shane-XB-Qian:
um.. I do not have problem, I may don't care that matchparan either; my first feeling was perhaps we should not break that rule which is well-known 3 is reserved.
if :match :2match is not enough, then find way to add more, that maybe real problem they faced.


--
shane.xb.qian

yegappan:
A similar problem was reported against the LSP plugin (https://github.com/yegappan/lsp/issues/339).  I debugged this problem and found that if a buffer is reloaded, the matchparen plugin may try to reuse the ID again.  But I couldn't come up with the steps to reproduce this problem.  It makes sense to modify the matchparen plugin to use the matchadd() API.

chrisbra:
> we should not break that rule which is well-known 3 is reserved.

The current [warning is this](https://vimhelp.org/pattern.txt.html#%3A3match):

```
The ":3match" command is used by the |matchparen| plugin.  You
are suggested to use ":match" for manual matching and
":2match" for another plugin.
```

in #13393 we change it to this instead (note, that we are still not recommending to use it):
```
The ":3match" command is used by (older Vims) |matchparen|
plugin. You are suggested to use ":match" for manual matching
and ":2match" for another plugin or even better make use of
the more flexible |matchadd()| (and similar) functions instead.
```

Shane-XB-Qian:
ok.
- "older Vims" maybe should more specific, how about v9.1?
- added a comma at ', or even better ...'

--
shane.xb.qian

PR #13385, opened 2023-10-19
runtime(json5): Add new ftplugin

dkearns:


chrisbra:
are we using `*.*` still for all files in browsefilter?

dkearns:
I was probably just planning on updating them all at once where "planning" was more a subconscious desire for consistency than anything actually planned. ;)

I can change it if you like.

chrisbra:
okay, thanks. Let me merge it then.

PR #13386, opened 2023-10-19
When restoring the exception state, process any outstanding exceptions

### LINKED ISSUE #13382: `:defer` problems with exception reporting and propogation
### Steps to reproduce

(This is copied from #13376)

I've been curious as to why, in 
`Errors are reported but do not cause aborting execution of deferred functions.`
the word "reported" is used.

I ask because **two exceptions are thrown**, but only **one is reported**. The first exception is lost. Run the script below; to get simpler output, get rid of the `try/catch/finally` and just run Foo().

Note: `&verbose = 14`
```
Throwing TestException
Exception thrown: TestException
Exception thrown: InnerException
Exception made pending: InnerException
finally [30, 31, 20, 9990, 10, 11]
v:exception:
Exception resumed: InnerException
Exception discarded: InnerException
Error detected while processing :source buffer=1[38]..:source buffer=1[32]..function <SNR>11_Foo[5].
.<SNR>11_DeferWithExcept[2]..<SNR>11_Bar:
line    2:
E605: Exception not caught: InnerException
```

```
vim9script

&verbose = 14

var callTrace: list<number> = []
def Bar()
  callTrace += [9990]
  throw 'InnerException'
enddef

def Defer(tag: number)
  callTrace += [tag]
  callTrace += [tag + 1]
enddef

def DeferWithExcept(tag: number)
  callTrace += [tag]
  Bar()
  callTrace += [tag + 1]
enddef


def Foo()
  defer Defer(10)
  defer DeferWithExcept(20)
  defer Defer(30)
  echo "Throwing TestException"
  throw "TestException"
enddef

try
  Foo()
catch /TestException/
  callTrace += [7]
finally
  echo "finally" callTrace
  echo "v:exception:" v:exception
endtry

echo "finished" callTrace
```

### Expected behaviour

(more like suggested behavior, I just don't know)

Exceptions in a deferred function do not propagate beyond the function.
They do not interfere with the regular flow of exception handling.

It as though they run like
```
try
    DeferredFunction()
catch
    report_exception()
endtry
```

It's kind of like the deferred functions are run as separate threads, an exception in one thread does not affect the execution of other threads. That seems to fit with the idea that if the current deferred function is aborted, then the other deferred functions are executed.

I just read (some of) the `:help exception-handling`, wow.

### Version of Vim

9.0.2050 (but `:version` shows 2049)

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #13382.

### DIFF
--- a/runtime/doc/userfunc.txt
+++ b/runtime/doc/userfunc.txt
@@ -443,7 +443,8 @@ Any return value of the deferred function is discarded.  The function cannot
 be followed by anything, such as "->func" or ".member".  Currently `:defer
 GetArg()->TheFunc()` does not work, it may work in a later version.
 
-Errors are reported but do not cause aborting execution of deferred functions.
+Errors are reported but do not cause aborting execution of deferred functions
+or altering execution outside of deferred functions.
 
 No range is accepted.  The function can be a partial with extra arguments, but
 not with a dictionary. *E1300*
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -28,7 +28,8 @@ Vim9 classes, objects, interfaces, types and enums.
 The fancy term is "object-oriented programming".  You can find lots of study
 material on this subject.  Here we document what |Vim9| script provides,
 assuming you know the basics already.  Added are helpful hints about how to
-use this functionality effectively.
+use this functionality effectively.  Vim9 classes and objects cannot be used
+in legacy Vim scripts and legacy functions.
 
 The basic item is an object:
 - An object stores state.  It contains one or more variables that can each
--- a/src/ex_eval.c
+++ b/src/ex_eval.c
@@ -757,6 +757,7 @@ exception_state_save(exception_state_T *estate)
     estate->estate_did_throw = did_throw;
     estate->estate_need_rethrow = need_rethrow;
     estate->estate_trylevel = trylevel;
+    estate->estate_did_emsg = did_emsg;
 }
 
 /*
@@ -765,11 +766,14 @@ exception_state_save(exception_state_T *estate)
     void
 exception_state_restore(exception_state_T *estate)
 {
-    if (current_exception == NULL)
-	current_exception = estate->estate_current_exception;
-    did_throw |= estate->estate_did_throw;
-    need_rethrow |= estate->estate_need_rethrow;
-    trylevel |= estate->estate_trylevel;
+    // Handle any outstanding exceptions before restoring the state
+    if (did_throw)
+	handle_did_throw();
+    current_exception = estate->estate_current_exception;
+    did_throw = estate->estate_did_throw;
+    need_rethrow = estate->estate_need_rethrow;
+    trylevel = estate->estate_trylevel;
+    did_emsg = estate->estate_did_emsg;
 }
 
 /*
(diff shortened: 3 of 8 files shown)


errael:
Looks pretty good. To avoid muddying up the PR, several examples of output/behavior with current PR will be added to #13386. (originally posted this message to wrong place, apologies for dup)

One thing that might be useful is an indication of which exceptions are
coming from deferred functions. The name of the function is already in
the exception trace. If it's worthwhile, here are two possibilities as examples

1. Adjust the existing message.
    ```
    Error detected while processing deferred function ...
    ```
2. Precede the message with something on it's own line
    ```
    Deferred function error.
    Error detected while processing ...
    ```

The trace for deferred functions doesn't have the line where the defer was posted; it seems like if it's triggered by an exception you see the exception line, when triggered by function exit you see the first line of the function. Don't know if it's worth adding this information, but it's something to notice. Hmm, might argue for using (2.) and then "deferred from" information could be more easily displayed.

errael:
Posted https://github.com/vim/vim/issues/13382#issuecomment-1771557004

There's one potential problem I can identify, depends on the spec.

Working from a spec like

---

Exceptions in a deferred function do not propagate beyond the function.
They do not interfere with the regular flow of exception handling.

It as though they run like
```
try
    DeferredFunction()
catch
    report_exception()
endtry
```

The line in the `:defer` docs could be modified by adding a line
something like
```
Errors are reported but do not cause aborting execution of deferred functions
or altering execution outside of deferred functions.
```

---

If the above is the spec then there's one problem.
If that's not the spec, it would be useful to know what the spec is.

I especially like the above spec because
- It is in line with the defer doc
- totally decouples execution of deferred functions from regular exception handling. Regular exception handling is complex and documented in minute detail, see `:help exception-handling`. I would prefer to leave all that alone.

The problem can be seen in the examples mentioned at the start of this note,
look for "POSSIBLE PROBLEM". In that case the exception throwin in a
deferred function leaks into the mainline execution flow.

yegappan:
> Posted examples [#13382 (comment)](https://github.com/vim/vim/issues/13382#issuecomment-1771557004)
> 
> There's one potential problem I can identify, depends on the spec.
> 

Updated the PR with a fix for this.  Can you try the latest diff from the PR?

errael:
> > Posted examples [#13382 (comment)](https://github.com/vim/vim/issues/13382#issuecomment-1771557004)
> > There's one potential problem I can identify, depends on the spec.
> 
> Updated the PR with a fix for this. Can you try the latest diff from the PR?

Looks good, fixed the exception in deferred function leaking in to mainline.

When this finds it way into main repo, I'll try more stuff.

errael:
I can open a feature request for the stuff mentioned in https://github.com/vim/vim/pull/13386#issuecomment-1771530302.

Just now, going through the tests again, confirms it is very difficult to pinpoint the :defer statement that queued up the function that got the exception.

runtime/doc/userfunc.txt
```diff
@@ -444,6 +444,7 @@ be followed by anything, such as "->func" or ".member".  Currently `:defer
 GetArg()->TheFunc()` does not work, it may work in a later version.
 
 Errors are reported but do not cause aborting execution of deferred functions.
```
zeertzjq:
```suggestion
Errors are reported but do not cause aborting execution of deferred functions
```

chrisbra:
thanks!

PR #13391, opened 2023-10-19
Adjust vim9class messages to follow common pattern.

errael:
The pattern is like:
[Variable|Method] "var-or-meth-name" ... class "class-name"


yegappan (approved):

chrisbra:
makes sense, thanks!

PR #13392, opened 2023-10-20
win32: Update iscygpty to the latest version

k-takata:
Import the latest version from https://github.com/k-takata/ptycheck.
This addresses #13332.


chrisbra:
thanks!

PR #13395, opened 2023-10-21
tests: remove incorrect :unlet

zeertzjq:
This :unlet actually causes an error, and slows testing by 1 second.


zeertzjq:
I may have checked something incorrectly. I'll open another PR later.

PR #13396, opened 2023-10-21
tests: avoid error when there are no swap files

zeertzjq:
When there are no swap files, g:name is never created.


zeertzjq:
Currently this doesn't seem to actually happen, but when it happens it doesn't cause test failure, just a 1-second delay, which isn't easy to notice.

chrisbra:
thanks!

PR #13397, opened 2023-10-21
Detect syntax error in funcref vararg declaration

errael:
and add help with funcref assignment examples.

**Before PR**

Normal funcref with varargs declaration
```
var Func6: func(...list<string>): number
Func6 = (...v: list<string>): number => 3  # OK
```

Error in function varargs declaration
```
def F(...v: string)
enddef          # E1180: Variable arguments type must be a list: string
```

Should be an error in varargs declaration as in previous
```
var Func6: func(...string): number      # OK, well at least no error given
Func6 = (...v: list<string>): number => 3
#E1012: Type mismatch; expected func(...string): number but got func(...list<string>): number
```

**After PR**

As expected.
```
var Func6: func(...string): number
E1180: Variable arguments type must be a list: string
```

The only change in the tests was different output messages;
and the only tests affected were added/changed as part of https://github.com/vim/vim/commit/96952b2e98cb5f05bc3b3adfb0029056f39cffef

Since no other tests failed with this change; I'm assuming that in a `funcref`
`...string` is not magic shorthand for something.

### DIFF
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -11010,6 +11010,7 @@ vim9-differences	vim9.txt	/*vim9-differences*
 vim9-export	vim9.txt	/*vim9-export*
 vim9-false-true	vim9.txt	/*vim9-false-true*
 vim9-final	vim9.txt	/*vim9-final*
+vim9-func-declaration	vim9.txt	/*vim9-func-declaration*
 vim9-function-defined-later	vim9.txt	/*vim9-function-defined-later*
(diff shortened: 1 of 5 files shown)


errael:
There's an error in `vimtags`, I added a tag, not sure what to do about that. Manual procedure?

yegappan (approved):

yegappan:
> There's an error in `vimtags`, I added a tag, not sure what to do about that. Manual procedure?

You can run the `make tags` command in the runtime/doc directory to regenerate the tags file.

PR #13399, opened 2023-10-21
filetype: identify Pacman hooks as confini

guidocella:
Because confini has much better syntax highlighting than conf.

For reference, I identified pacman.conf and pacman hooks as dosini in https://github.com/vim/vim/pull/6335, then
https://github.com/vim/vim/pull/10213 changed them to conf, then https://github.com/vim/vim/pull/10518 changed pacman.conf to confini but forgot to change hooks.


chrisbra:
thanks!

PR #13400, opened 2023-10-21
filetype(janet): Janet files are not recognised

dkearns:
Problem:  Janet files are not recognised
Solution: Add filename and shebang detection

http://www.janet-lang.org


chrisbra:
thanks!

PR #13402, opened 2023-10-21
test: to reproduce regression or broken by #12971

Shane-XB-Qian:
test for https://github.com/vim/vim/pull/12971#issuecomment-1759505911

### DIFF
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -2279,8 +2279,16 @@ func Test_complete_info_index()
   call feedkeys("Go\<C-X>\<C-P>\<C-P>\<F5>\<Esc>_dd", 'tx')
   call assert_equal("eee", g:compl_info['items'][g:compl_info['selected']]['word'])
 
+  " Check if index out of range
+  " https://github.com/vim/vim/pull/12971
+  call feedkeys("Go\<C-X>\<C-N>\<C-P>\<C-P>\<F5>\<Esc>_dd", 'tx')
+  call assert_equal("fff", g:compl_info['items'][g:compl_info['selected']]['word'])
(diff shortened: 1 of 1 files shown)


chrisbra:
can you please verify using #13408? CC @LemonBoy can you also have a look please?

Shane-XB-Qian:
seems still failed when 'noselect', i added more test.

Shane-XB-Qian:
i verified this test with v8.2, the test passed,
so that change is a regression or broken, and your patch seems still not fix that.

chrisbra:
not sure about the noselect, have only concentrated on the other issue so far. Can you please move this to the PR?

chrisbra:
oh wait, your test doesn't make any sense. if 'noselect' is set, you cannot possibly try to access the selected items, because `completion_info['selected']` will report `-1` and that is expected here.

Shane-XB-Qian:
I do not understand what you are saying..
'noselect' just impact the init start, but you should see the had been moved by these `<c-n> or <c-p>`

actually I roughly check the problem was: the items missed one which is that `fff` lost, though the err is index out of range.
there are only 5 items after round, so of course it is out of range.

--
shane.xb.qian

chrisbra:
hm, I see, it's a different problem, caused by 'noselect'. Please report issues not here, but in the corresponding PR.

Shane-XB-Qian:
i think whatever 'noselect' or not which just about the start index was different.
and anyway, this ticket is just for #12971 which i said for that index out of range.
as for your patch #13408 no idea if the same purpose, but for now did not pass this test.

this test can be passed at v8.2, so i preferred whichever later patch can fix this issue, but once re-ran this test passed, then ok to go, as for you'd like to merge this PR or not, it is not a matter.

chrisbra:
tests included as part of #13408 so closing this one here.

Issue #12961, opened 2023-08-30
[vim9script] variable used as a class incomplete


errael:
### Steps to reproduce

Consider the following which works OK
```
vim9script

class C
    def CF()
        echo "Boo"
    enddef
endclass

def F()
    var c = C
    var o = c.new()
    o.CF()
enddef
F()
```
If you add `o = null_class` as the last line of `F` you get an error.

This was (is?) mentioned in the todo list, but didn't make it to the list of known vim9script issues that was recently posted to the dev forum. Opening this so it's not lost.

### Expected behaviour

- `o = null_class` should work
 - There is no way to type a variable to alias a class. 
 - comparison against null/null_class should work for an object typed as null.

### Version of Vim

9.1

### Environment

9.1

mg979:
https://github.com/vim/vim/blob/1bed993c01c4b5e4c6ab55e80e2410ba87de487c/runtime/doc/vim9.txt#L1030-L1043

Tbh I think it's nonsense, `null` should be good enough for all of them.

dkearns:
`o` is an object. 

`o = null_object` works as does `C = null_class`.

`c = null_class` doesn't work.

We seem to have another fan of the family of null thingamebobs. :)

errael:
> It seems as though more of the class type is implemented than is not.

Now that `:type` is implemented, there's a possibility that `vim9script` will not allow a type to be used as a variable or value. This seems related to `:help todo`
> When "Meta" is a class, is "const MetaAlias = Meta" allowed?  It should
> either work or given an error. Possibly give an error now and implement it
> later (using a typedef).  #12006

And #13450 is also related.

errael:
The recent "type can not be a value" PRs resolve this.

Issue #12369, opened 2023-05-10
Cannot use class type of itself in the method


thinca:
### Steps to reproduce

```vim
vim9script

class Foo
  this.num: number
  def Equals(other: Foo): bool
    return this.num == other.num
  enddef
endclass
```

The following error occurs when this file is `:source`

```
E1010: Type not recognized: Foo
```

### Expected behaviour

The class can be defined without errors.

### Version of Vim

9.0.1536

### Environment

OS: Arch Linux
terminal: mintty (ssh from Windows)
value of $TERM: screen-256color

lifepillar:
A workaround is to set the type to `any` and be satisfied with runtime checks. To avoid the generic `E715: Dictionary required` error if the type at runtime is incorrect, you may assign to a `Foo`-typed variable inside the method. Like this:

```vim
vim9script

class Foo
  this.num: number
  def Equals(other: any): bool
    const other_: Foo = other
    return this.num == other_.num
  enddef
endclass

var f1 = Foo.new(1)
var f2 = Foo.new(2)

def F()
  echo f1.Equals(f2)  # OK
  echo f1.Equals([])  # E1012: type mismatch
enddef

F()
```

You get at least an informative message when the script is run.

The issue is also for methods that *return* the object itself. For instance:

```vim
class Foo
  this.num: number
  def Clone(): any  # <== Cannot be Foo
    return Foo.new(this.num)
  enddef
endclass

var f1 = Foo.new(1)

def F()
  var f2: Foo = f1.Clone()  # Without the explicit type, raises E715
  echo f2 is f1             # OK, false
  echo f2.num == f1.num     # OK, true
enddef

F()
```

PR #13404, opened 2023-10-21
Prevent assigning class name to variable.

errael:
@lifepillar @yegappan @dkearns please review.
Code reviewers, please consider where the errors are generated. Is there a better place to check? Are all the cases caught?

Error is `E1393: Cannot use a class as a variable or value`

**notes on changes**

Prevent assignment
- from non-value items, e.g. `class` or `typealias`.
- of non-value items as function argument or return.
- to `dict`/`list`: `var l = [1, C]`, `var d = {x: C}`.

NOTE: currently non-value items can be added to a list (while #13421 is being sorted out)
NOTE: a builtin must not return a non-value:, bare, in `list`, in `dict`

Here's an overview of the code changes made (so far).
```
simple assignments
    scriptlevel assignment  ex_let_one in evalvars.c
    :def assignment         compile_assignment in vim9compile.c

Function related checks
    scriptlevel argument    get_func_arguments in userfunc.c
    argument, call in def   check_ufunc_arg_types in vim9execute.c
    return value            ISN_RETURN (but gives wrong line number)
    defer/funcref           check_func_args_from_type in vim9instr.c

dict/list checks
    script list         eval_list in list.c
    script dict         eval_dict in dict.c
    compile list/dict   get_member_type_from_stack in vim9instr.c
```
There is a separate commit for `instanceof()` experimentation.
There is a separate commit for uniform builtin argument checking.

**TODO**
- More tests.
  Fixed tests as needed, but have around a dozen or so tests to write based on debug scripts. Plan to add tests after initial review (particularly of the error message).

**What now?**

This PR is very conservative. Some of the checks could be backed out.
The advantage of aggressive checks is uniformity, which should be less
confusing.
- No extra builtin checks, builtins have their own checks.
- Allow non-values in list/dict.
  This would ease some problems with builtins like `instanceof()`
  This still wouldn't allow the use of the non-value.
  Need to prevent non-value from being the result of an expression;
  protects against usage like aList[0].new().

---

**Notes on why the change**

With this PR, `Test_call_constructor_from_legacy()` failed, the test has been
replaced with `Test_construct_object_from_legacy()` which shows an alternate
way to do the same thing.

There's the following TODO item:
```
When "Meta" is a class, is "const MetaAlias = Meta" allowed?  It should
either work or given an error. Possibly give an error now and implement it
later (using a typedef).  #12006
```

This PR "give[s] an error now".
The important question: does preventing this take away an important capability?

When I first saw this TODO, I wondered why not just let it be, and take care
of it later. I played with it, #12961, and noted that it was very limited with
common stuff not working. I think I didn't read the original issue, #12006,
carefully; just reading it now, Bram's argument is
- it should be handled by typedef whenever something like that becomes available
- there is no class type.

If it's left in, **it leaves a lot of technical debt** to get it fully working.

Some uses for this, can be achieved with something like
```
class C
    static def CreateC(): any ... enddef
endclass
var CreateC: func = C.Create
```

### DIFF
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -630,6 +630,29 @@ is_scoped_variable(char_u *name)
 }
 
 /*
+ * Check if the typval_T is a value; report an error if it is not.
+ * Only a value can be assigned to a variable; if it's not a value
+ * it's probably a type.
+ *
+ * Return OK if it's a value, else FAIL
+ */
+    int
+check_is_value(vartype_T typ)
+{
+    if (typ == VAR_CLASS)
+    {
+	emsg(_(e_cannot_assign_class_to_variable));
+	return FAIL;
+    }
+    //else if (typ == VAR_TYPEALIAS)
+    //{
+    //    emsg(_(e_cannot_assign_typealias_to_variable));
+    //    return FAIL;
+    //}
+    return OK;
+}
+
+/*
  * Evaluate one Vim expression {expr} in string "p" and append the
  * resulting string to "gap".  "p" points to the opening "{".
  * When "evaluate" is FALSE only skip over the expression.
@@ -1876,6 +1899,8 @@ ex_let_one(
 	p = get_lval(arg, tv, &lv, FALSE, FALSE, lval_flags, FNE_CHECK_START);
 	if (p != NULL && lv.ll_name != NULL)
 	{
+	    if (check_is_value(tv->v_type) == FAIL)
+		return NULL;
 	    if (endchars != NULL && vim_strchr(endchars,
 					   *skipwhite(lv.ll_name_end)) == NULL)
 	    {
--- a/src/dict.c
+++ b/src/dict.c
@@ -1015,6 +1015,15 @@ eval_dict(char_u **arg, typval_T *rettv, evalarg_T *evalarg, int literal)
 		clear_tv(&tvkey);
 	    goto failret;
 	}
+	if (check_is_value(tv.v_type) == FAIL)
+	{
+	    if (evaluate)
+	    {
+		clear_tv(&tvkey);
+		clear_tv(&tv);
+	    }
+	    goto failret;
+	}
 	if (evaluate)
 	{
 	    item = dict_find(d, key, -1);
--- a/src/errors.h
+++ b/src/errors.h
@@ -3552,8 +3552,10 @@ EXTERN char e_missing_typealias_type[]
 	INIT(= N_("E1398: Missing type alias type"));
 EXTERN char e_type_can_only_be_used_in_script[]
 	INIT(= N_("E1399: Type can only be used in a script"));
+EXTERN char e_cannot_assign_class_to_variable[]
+	INIT(= N_("E1400: Cannot use a class as a variable or value"));
 #endif
-// E1400 - E1499 unused (reserved for Vim9 class support)
+// E1401 - E1499 unused (reserved for Vim9 class support)
 EXTERN char e_cannot_mix_positional_and_non_positional_str[]
 	INIT(= N_("E1500: Cannot mix positional and non-positional arguments: %s"));
 EXTERN char e_fmt_arg_nr_unused_str[]
--- a/src/errors.h
+++ b/src/errors.h
@@ -1779,8 +1779,8 @@ EXTERN char e_can_only_compare_list_with_list[]
 	INIT(= N_("E691: Can only compare List with List"));
 EXTERN char e_invalid_operation_for_list[]
 	INIT(= N_("E692: Invalid operation for List"));
-EXTERN char e_list_or_class_required_for_argument_nr[]
-	INIT(= N_("E693: List or Class required for argument %d"));
+EXTERN char e_class_or_typealias_required_for_argument_nr[]
+	INIT(= N_("E693: Class or Typealias required for argument %d"));
 EXTERN char e_invalid_operation_for_funcrefs[]
 	INIT(= N_("E694: Invalid operation for Funcrefs"));
 EXTERN char e_cannot_index_a_funcref[]
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -1627,7 +1627,7 @@ typedef struct
     char	*f_name;	// function name
     char	f_min_argc;	// minimal number of arguments
     char	f_max_argc;	// maximal number of arguments
-    char	f_argtype;	// for method: FEARG_ values
+    char	f_argtype;	// for method: FEARG_ values; bits FEFLG_
     argcheck_T	*f_argcheck;	// list of functions to check argument types
     type_T	*(*f_retfunc)(int argcount, type2_T *argtypes,
 							   type_T **decl_type);
@@ -1637,10 +1637,12 @@ typedef struct
 } funcentry_T;
 
 // values for f_argtype; zero means it cannot be used as a method
-#define FEARG_1    1	    // base is the first argument
-#define FEARG_2    2	    // base is the second argument
-#define FEARG_3    3	    // base is the third argument
-#define FEARG_4    4	    // base is the fourth argument
+#define FEARG_1    0x01	    // base is the first argument
+#define FEARG_2    0x02	    // base is the second argument
+#define FEARG_3    0x03	    // base is the third argument
+#define FEARG_4    0x04	    // base is the fourth argument
+#define FEARG_MASK 0x0F	    // bits in f_argtype used as argument index
+#define FEFLG_XVAL 0x10	    // builtin accepts a non-value (class, typealias)
 
 #if defined(HAVE_MATH_H)
 # define MATH_FUNC(name) name
@@ -3115,7 +3117,7 @@ check_internal_func(int idx, int argcount)
     else if (argcount > global_functions[idx].f_max_argc)
 	res = FCERR_TOOMANY;
     else
-	return global_functions[idx].f_argtype;
+	return global_functions[idx].f_argtype & FEARG_MASK;
 
     name = internal_func_name(idx);
     if (res == FCERR_TOOMANY)
@@ -3172,14 +3174,14 @@ call_internal_method(
     fi = find_internal_func(name);
     if (fi < 0)
 	return FCERR_UNKNOWN;
-    if (global_functions[fi].f_argtype == 0)
+    if ((global_functions[fi].f_argtype & FEARG_MASK) == 0)
 	return FCERR_NOTMETHOD;
     if (argcount + 1 < global_functions[fi].f_min_argc)
 	return FCERR_TOOFEW;
     if (argcount + 1 > global_functions[fi].f_max_argc)
 	return FCERR_TOOMANY;
 
-    if (global_functions[fi].f_argtype == FEARG_2)
+    if ((global_functions[fi].f_argtype & FEARG_MASK) == FEARG_2)
     {
 	if (argcount < 1)
 	    return FCERR_TOOFEW;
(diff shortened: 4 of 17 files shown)


errael:
Add additional test, it does: `call(A.new, args)`

errael:
Transitioning this to `Ready for review`.  Not `Ready for commit`, although it is and I'd like to see it committed at some point.

dkearns:
This is just a quick general comment for the moment, not addressing the specifics of the PR.

It seems to me, from the existing implementation, that classes were intended to to be first-class objects.  This is in keeping with the language in general. 

> there is no class type.

`:help type()` lists `v:t_class` and it is implemented.

errael:
> a quick general comment

Appreciated. Main question is whether or not `var c = SomeClass`, which I think is accidentally implemented,  should be allowed in v9.1; assuming vim9.1 is relatively soon and that there's no work to make class a full blown variable type or that there's not going to be an alternate way (like typedef) to do this. Is there a downside to disallowing the assignment for now, until a decision is made?

> > there is no class type.
> 
> `:help type()` lists `v:t_class` and it is implemented.

Interesting point, that statement is based on Bram's comment "It's not an object and we don't have a type that is a class" in https://github.com/vim/vim/issues/12006#issuecomment-1546748966; Bram closed that issue with "Once something like a typedef has been implemented please check this again. Closing for now."

Maybe the distinction is that it's not a type because you can't do
```
var someClass: class
```
although internally there is a vim object of type VAR_CLASS. And you can do
```
vim9script
class C
endclass
echo type(C)
echo typename(C)
```
```
12
class<C>
```

But there are several breadcrumbs about a class type.

lifepillar:
I am actually assigning classes to a const variables in my code. This is my use case:

    import `mylibrary.vim`
    const SomeClass = mylibrary.SomeClass

As you mention, that might be:

    typedef SomeClass mylibrary.SomeClass

but that is not implemented yet, afaik. If the class assignment is removed, one is forced to use imported classes with a prefix, I think. Or is there another way?

I personally find it very convenient to be able to assign non-prefixed aliases to imported names; it makes my code much easier to read and write.

lifepillar:
@errael  Can you elaborate on your proposed workaround? I have tried this:

```vim

vim9script
# lib.vim

export class Context
  this.text: string
  public this.index: number = 0

  static def MakeContext(text: string, index = 0): any
    return Context.new(text, index)
  enddef

  def new(this.text, this.index = v:none)
  enddef
endclass
```

```vim
vim9script

import './lib.vim'
const MakeContext = lib.Context.MakeContext

def Test_MakeContext()
  var ctx = MakeContext("X", 42)
  assert_equal(v:t_object, type(ctx))
  assert_equal("X", ctx.text)  # <== E715
  assert_equal(42, ctx.index)  # <== E715
enddef

Test_MakeContext()
```

Sourcing the latter script results in `E715: Dictionary required` any tyme a member is accessed.

lifepillar:
Ok, I can make it work like this:

```vim
vim9script
# lib.vim

export class Context
  this.text: string
  public this.index: number = 0

  def new(this.text, this.index = v:none)
  enddef
endclass

export def NewContext(text: string, index = 0): Context
  return Context.new(text, index)
enddef
```

```
vim9script

import './lib.vim'
const NewContext = lib.NewContext

def Test_MakeContext()
  var ctx = NewContext("X", 42)
  assert_equal(v:t_object, type(ctx))
  assert_equal("X", ctx.text)
  assert_equal(42, ctx.index)
enddef

Test_MakeContext()
```

It is still the case, though, that if I want to define a function that takes a `Context` object as an argument, I need to qualify it:

```
def F(context: lib.Context)
...
```

A bit cumbersome, but maybe not too bad, after all. Please consider the ergonomics of removing the assignment without providing a replacement. It feels a bit unnatural to me that I can assign anything but classes.

errael:
> ```
> import `mylibrary.vim`
> const SomeClass = mylibrary.SomeClass
> ```
> 
> 
> I personally find it very convenient to be able to assign non-prefixed aliases to imported names; it makes my code much easier to read and write.

Oh crap, how could I forget this important case, which I use all the type with functions. Without `:type` this assignment is needed.

errael:
> 
> ```viml
> vim9script
> # lib.vim
> 
> export class Context
>   static def MakeContext(text: string, index = 0): any
> endclass
> ```

Right, `MakeContext()` can't be properly declared because of #12369.

> ```viml
> vim9script
> def Test_MakeContext()
>   var ctx = MakeContext("X", 42)
>   assert_equal("X", ctx.text)  # <== E715
> enddef
> ```
> 
> Sourcing the latter script results in `E715: Dictionary required` any time a member is accessed.

If `object` was a type, #13404, then could do `static def MakeContext(): object`, that should provided sufficient information to the compiler to avoid the all to familiar `E715`.  But that's a workaround in this case, (and it's not implemented anyway). In other cases it can be useful.


> It is still the case, though, that if I want to define a function that takes a `Context` object as an argument, I need to qualify it:
> 
> ```
> def F(context: lib.Context)
> ...
> ```
> 
> A bit cumbersome, but maybe not too bad, after all.

You're right, I don't think that's acceptable. But maybe `:type`.

> Please consider the ergonomics of removing the assignment without providing a replacement. It feels a bit unnatural to me that I can assign anything but classes.

`class C` defines a type, you can do `var o: C`.
If `var x = C` works, would you expect `var x = number` to work?

But it's beginning to look like `:type` might be an option, #13407.
Would `type Context lib.Context` be OK?

lifepillar:
I see. Clearly, something like `var x = C` where `C` is a class (that is, a type) makes sense if classes are themselves values (as in languages like Ruby, where each class is an instance of `Class`, iirc). But the Ruby way doesn't fit Vim 9 script.

>Would `type Context lib.Context` be OK?

It's the proper way to do it, imo. If you are going to implement `:type` then I have no objections against merging this PR. `:type` is the only feature I miss in Vim 9 script. I'm looking forward to replacing things like `func(func(list<dict<any>>)): dict<any>` with a readable name :)

errael:
> If you are going to implement :type then I have no objections against merging this PR.

Cool. I'll put it back to "Draft PR", and reopen when `:type` is available. That should allow the smoothest transition after breaking your code.

dkearns:
> Main question is whether or not `var c = SomeClass`, which I think is accidentally implemented, should be allowed in v9.1; assuming vim9.1 is relatively soon and that there's no work to make class a full blown variable type or that there's not going to be an alternate way (like typedef) to do this. Is there a downside to disallowing the assignment for now, until a decision is made?

Probably not now that `:type` is available to take care of the aliasing cases.

I think you've missed assignments to parameters.

```vim
vim9script

class A
endclass

def Foo(c: any)
  echomsg typename(c)
enddef

Foo(A)
```

> > > there is no class type.
> > 
> > 
> > `:help type()` lists `v:t_class` and it is implemented.
> 
> Interesting point, that statement is based on Bram's comment "It's not an object and we don't have a type that is a class" in [#12006 (comment)](https://github.com/vim/vim/issues/12006#issuecomment-1546748966); Bram closed that issue with "Once something like a typedef has been implemented please check this again. Closing for now."

I can't understand the comment there, given that the help entry was added only a month earlier.

> But there are several breadcrumbs about a class type.

Indeed.  I wouldn't like to see it sent out into the world in its current state even with assignment disabled without some sort of reasonable narrative.

It looks a bit like class/singleton-metaclass pairs.  I suppose something like that could fall out of the implementation but `class<Foo>` looks quite specific.

errael:
> I think you've missed assignments to parameters.
> 
> ```viml
> vim9script
> 
> class A
> endclass
> 
> def Foo(c: any)
>   echomsg typename(c)
> enddef
> 
> Foo(A)
> ```

Ah, of course. Thanks. Need a function like `check_is_value(typval_T*)`

> > > > there is no class type.
> > > 
> > > 
> > > `:help type()` lists `v:t_class` and it is implemented.
> > 
> > 
> > Interesting point, that statement is based on Bram's comment "It's not an object and we don't have a type that is a class" in [#12006 (comment)](https://github.com/vim/vim/issues/12006#issuecomment-1546748966); Bram closed that issue with "Once something like a typedef has been implemented please check this again. Closing for now."
> 
> I can't understand the comment there, given that the help entry was added only a month earlier.
> 
> > But there are several breadcrumbs about a class type.
> 
> Indeed. I wouldn't like to see it sent out into the world in its current state even with assignment disabled without some sort of reasonable narrative.

I found @lifepillar's distinction between `type` and `value` clarifying. https://github.com/vim/vim/pull/13404#issuecomment-1774181117

> 
> It looks a bit like class/singleton-metaclass pairs. I suppose something like that could fall out of the implementation but `class<Foo>` looks quite specific.

`class<Foo>` might be an artifact of how `typename()` displays non-value items. And now there's
```
vim9script
type LN list<number>
echo typename(LN)
```
which outputs `typealias<list<number>>`

errael:
Implementation note: From what I've seen there's a dictionary of accessible names.
```
dictitem_T *v = find_var(name, ...)
typval_T *tv = &v->di_tv;
```
Maybe pre vim9class, tv was always a value. Now it can also be a type.

src/evalvars.c
```diff
@@ -1158,7 +1158,9 @@ ex_let(exarg_T *eap)
     // declaration, not the expression.
     SOURCING_LNUM = cur_lnum;
 
-    if (!eap->skip && eval_res != FAIL)
+    if (!eap->skip && eval_res != FAIL && rettv.v_type == VAR_CLASS)
```
yegappan: You can use list assignment to assign multiple variables (B is a class in the statement below):

```
    [a, b] = [10, B]
```

To handle this case, this check should be moved to the ex_let_one() function.

errael: > To handle this case, this check should be moved to the ex_let_one() function

Thanks, it was still on my list.

errael:
Pushed a tweak to allow non-value items in `list` while `instanceof()` issues are sorted out.

errael:
> Can you elaborate on your proposed workaround? I have tried this:
[snip]
> 
> ```viml
> vim9script
> 
> import './lib.vim'
> const MakeContext = lib.Context.MakeContext
> 
> def Test_MakeContext()
>   var ctx = MakeContext("X", 42)
>   assert_equal(v:t_object, type(ctx))
>   assert_equal("X", ctx.text)  # <== E715
>   assert_equal(42, ctx.index)  # <== E715
> enddef
> 
> Test_MakeContext()
> ```
> 
> Sourcing the latter script results in `E715: Dictionary required` any time a member is accessed.

@lifepillar With `:type` (now merged) you can do the following to easily specify the type and avoid `E715`.

```
vim9script

import '/tmp/xlib.vim'

type Context = xlib.Context
const MakeContext = xlib.Context.MakeContext

def Test_MakeContext()
  var ctx: Context = MakeContext("X", 42)
  echo ctx.text
  echo ctx.index
enddef

Test_MakeContext()
```

There are still things that are a hassle or unexpected. Which make this a workaround.
- If  #12369 is fixed, the type could be automatically assigned.
- The `typealias` can not be used as a class type in all situations
  `Context.new('X', 42)` does not work, necessitating the `MakeContext` class function.
  `const MakeContext = Context.MakeContext` does not work

errael:
Rebase with `:type`. Without design review/feedback (see description https://github.com/vim/vim/pull/13404#issue-1955679155) this PR is on hold.

yegappan:
Why was this PR closed?

errael:
> Why was this PR closed? Is this not needed anymore?

I closed it in favor of https://github.com/yegappan/vim/pull/2 which is identical to this PR.

yegappan:
> * The `typealias` can not be used as a class type in all situations

I have updated PR https://github.com/vim/vim/pull/13441 to address this issue.

chrisbra:
@yegappan can you please check , if  Commit feaccd239573a6265d39d3a917862ee40742eab4  includes your latest change?

yegappan:
> @yegappan can you please check , if Commit [feaccd2](https://github.com/vim/vim/commit/feaccd239573a6265d39d3a917862ee40742eab4) includes your latest change?

Yes.  Your commit includes the latest change.  Thanks.

chrisbra:
Thanks. Glad I didn't merge 5 minutes earlier :)

PR #13405, opened 2023-10-22
gn files are not recognized

amaanq:
GN (generate ninja files) are not recognized

https://gn.googlesource.com/gn/


chrisbra:
thanks. I suppose it is okay to not have a ftplugin for now?

PR #13407, opened 2023-10-22
Add support for type aliases (:type)

yegappan:
Implement the :type command (in the todo list) to create type aliases.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -743,12 +743,13 @@ constructor methods.
 
 7.  Type definition					*Vim9-type* *:type*
 
-{not implemented yet}
+A type definition is giving a name to a type specification.  This also known
+type alias.  For Example: >
 
-A type definition is giving a name to a type specification.  For Example: >
+	:type ListOfStrings = list<string>
 
-	:type ListOfStrings list<string>
-
+The type alias can be used wherever a built-in type can be used.  The type
+alias name must start with an upper case character.
 
 ==============================================================================
 
--- a/src/errors.h
+++ b/src/errors.h
@@ -3384,17 +3384,17 @@ EXTERN char e_invalid_object_variable_declaration_str[]
 EXTERN char e_not_valid_command_in_class_str[]
 	INIT(= N_("E1318: Not a valid command in a class: %s"));
 EXTERN char e_using_class_as_number[]
-	INIT(= N_("E1319: Using a class as a Number"));
+	INIT(= N_("E1319: Using a Class as a Number"));
 EXTERN char e_using_object_as_number[]
-	INIT(= N_("E1320: Using an object as a Number"));
+	INIT(= N_("E1320: Using an Object as a Number"));
 EXTERN char e_using_class_as_float[]
-	INIT(= N_("E1321: Using a class as a Float"));
+	INIT(= N_("E1321: Using a Class as a Float"));
 EXTERN char e_using_object_as_float[]
-	INIT(= N_("E1322: Using an object as a Float"));
+	INIT(= N_("E1322: Using an Object as a Float"));
 EXTERN char e_using_class_as_string[]
-	INIT(= N_("E1323: Using a class as a String"));
+	INIT(= N_("E1323: Using a Class as a String"));
 EXTERN char e_using_object_as_string[]
-	INIT(= N_("E1324: Using an object as a String"));
+	INIT(= N_("E1324: Using an Object as a String"));
 EXTERN char e_method_not_found_on_class_str_str[]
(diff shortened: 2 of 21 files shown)


bfrg:
Personally, I would prefer a syntax with an equal sign in-between, like:

```
type foo_bar_baz = list<dict<any>>
```

It's a lot easier to spot the LHS and RHS of the type alias.

errael:
> 
> ```viml
> vim9script
> # lib.vim
> 
> export class Context
>   static def MakeContext(text: string, index = 0): any
> endclass
> ```

Right, `MakeContext()` can't be properly declared because of #12369.

> ```viml
> vim9script
> def Test_MakeContext()
>   var ctx = MakeContext("X", 42)
>   assert_equal("X", ctx.text)  # <== E715
> enddef
> ```
> 
> Sourcing the latter script results in `E715: Dictionary required` any time a member is accessed.

If `object` was a type, #13404, then could do `static def MakeContext(): object`, that should provided sufficient information to the compiler to avoid the all to familiar `E715`.  But that's a workaround in this case, (and it's not implemented anyway). In other cases it can be useful.


> It is still the case, though, that if I want to define a function that takes a `Context` object as an argument, I need to qualify it:
> 
> ```
> def F(context: lib.Context)
> ...
> ```
> 
> A bit cumbersome, but maybe not too bad, after all.

You're right, I don't think that's acceptable. But maybe `:type`.

Please consider the ergonomics of removing the assignment without providing a replacement. It feels a bit unnatural to me that I can assign anything but classes.

`class C` defines a type, you can do `var o: C`.
If `var x = C` works, would you expect `var x = number` to work?

But it's beginning to look like `:type` might be an option, #13407.
Would `type Context lib.Context` be OK?

dkearns:
Thanks for implementing this @yegappan.

> Personally, I would have preferred a syntax with an equal sign in-between the two types

It appears that might be the norm: https://en.wikipedia.org/wiki/Type_aliasing

errael:
Doh! I hadn't noticed/realized that this was **alias** and not **typedef**. Too bad, but oh well.

I wonder if a different command name, for example `:alias`, would be `better` than `:type`? Leave the door open to a typedef type of command when/if the type system is extended.

Shane-XB-Qian:
> 18912bc5c2aea35661c62c65d7e83d7c74b72048  Add support for type aliases (:type)
> 90874c8ad41030bce070e0ec3258184f50f53f7f  Support using = when defining a type.  Handle more error conditions

seems generally type alias is with '=' vs type define then no '='
just per your commit msg, here looks right opposite it, is it ok?

yegappan:
> > [18912bc](https://github.com/vim/vim/commit/18912bc5c2aea35661c62c65d7e83d7c74b72048)  Add support for type aliases (:type)
> > [90874c8](https://github.com/vim/vim/commit/90874c8ad41030bce070e0ec3258184f50f53f7f)  Support using = when defining a type.  Handle more error conditions
> 
> seems generally type alias is with '=' vs type define then no '=' just per your commit msg, here looks right opposite it, is it ok?

The syntax for creating a type alias is:

```
  type MyType = list<number>
```

yegappan:
> Thanks for implementing this @yegappan.
> 
> > Personally, I would have preferred a syntax with an equal sign in-between the two types
> 
> It appears that might be the norm: https://en.wikipedia.org/wiki/Type_aliasing

I have updated the PR to require "=" after the type alias name.

chrisbra:
is this ready now?

yegappan:
> is this ready now?

Yes.  This is ready.

chrisbra:
thanks!

dkearns:
> Doh! I hadn't noticed/realized that this was **alias** and not **typedef**. Too bad, but oh well.
> 
> I wonder if a different command name, for example `:alias`, would be `better` than `:type`? Leave the door open to a typedef type of command when/if the type system is extended.

https://web.archive.org/web/20231026083713/https://moolenaar.net/z/zimbu/spec/zimbu.html#ALIAS%32and%32TYPE

errael:
> > Doh! I hadn't noticed/realized that this was **alias** and not **typedef**. Too bad, but oh well.
> > I wonder if a different command name, for example `:alias`, would be `better` than `:type`? Leave the door open to a typedef type of command when/if the type system is extended.
> 
> https://web.archive.org/web/20231026083713/https://moolenaar.net/z/zimbu/spec/zimbu.html#ALIAS%32and%32TYPE

Thanks, (very cool), I didn't realize (forgot?) about `zimbu` being Bram's.

This argues for using `:alias` for what the current `:type` does. Then there's room for expansion to a more sophisticated type system without having a naming conflict/problem.

PR #13408, opened 2023-10-22
completion: move in right direction when  filling completion_info()

chrisbra:
When moving through the insert completion menu and switching directions, we need to make sure we start at the correct position in the list and move correctly forward/backwards through it, so that we do not skip entries and the selected item points to the correct entry in the list of completion entries generated by the completion_info() function.

The general case is this:

1) CTRL-X CTRL-N, we will traverse the list starting from
   compl_first_match and then go forwards (using the cp->next pointer)
   through the list (skipping the very first entry, which has the
   CP_ORIGINAL_TEXT flag set (since that is the empty/non-selected entry

2) CTRL-X CTRL-P, we will traverse the list starting from
   compl_first_match (which now points to the last entry). The previous
   entry will have the CP_ORIGINAL_TEXT flag set, so we need to start
   traversing the list from the second prev pointer.

There are in fact 2 special cases after starting the completion menu with CTRL-X:

3)  CTRL-N and then going backwards by pressing CTRL-P again.
    compl_first_match will point to the same entry as in step 1 above,
    but since compl_dir_foward() has been switched by pressing CTRL-P
    to backwards we need to pretend to be in still in case 1 and still
    traverse the list in forward direction using the cp_next pointer

4)  CTRL-P and then going forwards by pressing CTRL-N again.
    compl_first_match will point to the same entry as in step 2 above,
    but since compl_dir_foward() has been switched by pressing CTRL-N
    to forwards we need to pretend to be in still in case 2 and still
    traverse the list in backward direction using the cp_prev pointer

related: #13402
related: #12971

### DIFF
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -2267,24 +2267,72 @@ func Test_complete_info_index()
   " Ensure 'index' in complete_info() is coherent with the 'items' array.
 
   set completeopt=menu,preview
-  " Search forward.
+  " Search forward
   call feedkeys("Go\<C-X>\<C-N>\<F5>\<Esc>_dd", 'tx')
   call assert_equal("aaa", g:compl_info['items'][g:compl_info['selected']]['word'])
+  call assert_equal(6 , len(g:compl_info['items']))
   call feedkeys("Go\<C-X>\<C-N>\<C-N>\<F5>\<Esc>_dd", 'tx')
   call assert_equal("bbb", g:compl_info['items'][g:compl_info['selected']]['word'])
+  call assert_equal(6 , len(g:compl_info['items']))
+  call feedkeys("Go\<C-X>\<C-N>\<C-N>\<C-N>\<F5>\<Esc>_dd", 'tx')
+  call assert_equal("ccc", g:compl_info['items'][g:compl_info['selected']]['word'])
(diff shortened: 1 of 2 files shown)


src/testdir/test_ins_complete.vim
```diff
@@ -2289,6 +2344,16 @@ func Test_complete_info_index()
   call feedkeys("Go\<C-X>\<C-P>\<F5>\<Esc>_dd", 'tx')
   call assert_equal(-1, g:compl_info['selected'])
 
+  " Check if index out of range
+  " https://github.com/vim/vim/pull/12971
+  call feedkeys("Go\<C-X>\<C-N>\<C-P>\<F5>\<Esc>_dd", 'tx')
+  call assert_equal(0, g:compl_info['selected'])
```
Shane-XB-Qian, line 2350: though test passed, but i donot understand why here the selected is 0?
looks this made 'noselect' much special.

chrisbra: what do you mean? It's index `0` from the bottom. that should be correct, no?

Shane-XB-Qian: i think it is 5, which same like non-noselect, and same like if it's not around.
// but if it is the used way (as someone said `i donot know why was done that way`), then to be compatible, let it be.

PR #13412, opened 2023-10-23
runtime(vim): Update ftplugin - comment motions

dkearns:
Fix the pattern used by comment-motion mappings to match the start of a
block comment.

Block-comment start lines were being ignored if the previous line
contained a double-quote character anywhere in the line.  Line comments
should only be ignored if the previous line is a full-line comment and,
therefore, part of the current block comment.


chrisbra:
thanks!

PR #13413, opened 2023-10-23
centralize safe executable check

benknoble:
Follow up to 816fbcc26 (patch 9.0.1833: [security] runtime file fixes, 2023-08-31) and f7ac0ef50 (runtime: don't execute external commands when loading ftplugins, 2023-09-06).

This puts the logic for safe executable checks in a single place so all filetypes benefit from consistency.

Notable changes:
- The gzip and zip plugins need to be opted into by enabling execution of those programs (or the global plugin_exec). This needs documentation or discussion.
- The ruby check is performed twice because of how the code is organized. Perhaps this could be refactored? But the working directory is changed as part of the function, so for now I think it should stay as-is.
- This fixes a bug in the zig plugin: code setting s:tmp_cwd was removed in f7ac0ef50 (runtime: don't execute external commands when loading ftplugins, 2023-09-06), but the variable was still referenced. Since the new function takes care of that automatically, the variable is no longer needed.

---

I haven't updated the last changed headers yet; I'll let maintainers weigh in on that.

CC maintainers: @cecamp, @dkearns, Martin Florian (?), Perl maintainers @petdance, @tpope, Zig upstream (@andrewrk ?).

### DIFF
--- a/runtime/autoload/dist/vim.vim
+++ b/runtime/autoload/dist/vim.vim
@@ -1,0 +1,17 @@
+vim9script
+
+# Vim runtime support library
+#
+# Maintainer:	The Vim Project <https://github.com/vim/vim>
+# Last Change:	2023 Oct 25
+
+export def IsSafeExecutable(filetype: string, executable: string): bool
+  var cwd = getcwd()
+  return get(g:, filetype .. '_exec', get(g:, 'plugin_exec', 0))
+    && (fnamemodify(exepath(executable), ':p:h') !=# cwd
+        || (split($PATH, has('win32') ? ';' : ':')->index(cwd) != -1
+            && cwd != '.'))
+enddef
+
+# Uncomment this line to check for compilation errors early
+# defcompile
--- a/runtime/autoload/gzip.vim
+++ b/runtime/autoload/gzip.vim
@@ -11,10 +11,7 @@ fun s:check(cmd)
   let name = substitute(a:cmd, '\(\S*\).*', '\1', '')
   if !exists("s:have_" . name)
     " safety check, don't execute anything from the current directory
-    let s:tmp_cwd = getcwd()
-    let f = (fnamemodify(exepath(name), ":p:h") !=# s:tmp_cwd
-          \ || (index(split($PATH,has("win32")? ';' : ':'), s:tmp_cwd) != -1 && s:tmp_cwd != '.'))
-    unlet s:tmp_cwd
+    let f = dist#vim#IsSafeExecutable('gzip', name)
     if !f
       echoerr "Warning: NOT executing " .. name .. " from current directory!"
     endif
--- a/runtime/autoload/zip.vim
+++ b/runtime/autoload/zip.vim
@@ -57,14 +57,10 @@ if !exists("g:zip_extractcmd")
  let g:zip_extractcmd= g:zip_unzipcmd
 endif
 
-let s:tmp_cwd = getcwd()
-if (fnamemodify(exepath(g:zip_unzipcmd), ":p:h") ==# getcwd()
-          \ && (index(split($PATH,has("win32")? ';' : ':'), s:tmp_cwd) == -1 || s:tmp_cwd == '.'))
- unlet s:tmp_cwd
+if !dist#vim#IsSafeExecutable('zip', g:zip_unzipcmd)
  echoerr "Warning: NOT executing " .. g:zip_unzipcmd .. " from current directory!"
  finish
 endif
-unlet s:tmp_cwd
 
 " ----------------
 "  Functions: {{{1
--- a/runtime/ftplugin/awk.vim
+++ b/runtime/ftplugin/awk.vim
@@ -37,8 +37,8 @@ if exists("g:awk_is_gawk")
     let b:undo_ftplugin .= " | setl fp<"
   endif
 
-  " Disabled by default for security reasons.  
-  if get(g:, 'awk_exec', get(g:, 'plugin_exec', 0))
+  " Disabled by default for security reasons.
+  if dist#vim#IsSafeExecutable('awk', 'gawk')
     let path = system("gawk 'BEGIN { printf ENVIRON[\"AWKPATH\"] }'")
     let path = substitute(path, '^\.\=:\|:\.\=$\|:\.\=:', ',,', 'g') " POSIX cwd
     let path = substitute(path, ':', ',', 'g')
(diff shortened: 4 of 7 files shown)


benknoble:
Oh, and this probably needs tests 😅

andrewrk:
Hello, nice work on fixing this stuff. I actually don't maintain the zig vim plugin although I am indeed a vim user. Do what you gotta do; we're happy to deal with whatever changes are needed in the plugin.

benknoble:
@andrewrk ah, thanks. I saw your name in some commits to the Zig upstream—do you know who a good contact is there?

runtime/autoload/dist/ft.vim
```diff
@@ -1218,5 +1218,13 @@ export def FTvba()
   endif
 enddef
 
+export def IsSafeExecutable(filetype: string, executable: string): bool
+  var cwd = getcwd()
+  return get(g:, filetype..'_exec', get(g:, 'plugin_exec', 0))
```
dkearns: Does this line actually work without whitespace around the `..` string concat operator?

dkearns:
Thanks for this @benknoble.

The `ft` module is really only intended for internal use.  As `IsSafeExecutable()` isn't related to filetype detection should it hitch a ride with `ft` or is it the start of a runtime support library?

runtime/autoload/dist/ft.vim
benknoble: 👀 probably not, but I haven't actually tested it. I'll add the whitespace.

benknoble:
> The `ft` module is really only intended for internal use. As `IsSafeExecutable()` isn't related to filetype detection should it hitch a ride with `ft` or is it the start of a runtime support library?

I envisioned more the latter (it could also be useful for setting `b:undo_ftplugin` and `b:undo_indent`: see [this kind of idea](https://github.com/benknoble/Dotfiles/blob/master/links/vim/autoload/bk/ftplugin.vim) and [it's usage](https://github.com/search?q=repo%3Abenknoble%2FDotfiles%20bk%23ftplugin%23undo&type=code)). Perhaps it would be better as `vim#IsSafeExecutable()`? We could keep the `dist#` prefix, too, if needed.

chrisbra:
yes a runtime library seems the way to go. I think adding dist to keep it unique is also probably worth it

benknoble:
Will make changes and push new version tonight. Tests will take longer.

tpope:
As the maintainer of `ftplugin/ruby.vim`, I'd prefer _not_ to couple the runtime file to a file distributed with Vim, as it would interfere with development and and user testing by requiring a brand new version of Vim to run it, and it doesn't really do anything I can't do myself. Not long after the incident first went down, I did a rewrite I'm much happier with, and just pushed it to the [upstream repository](https://github.com/vim-ruby/vim-ruby/commit/f06f069ce67bdda6f2cd408f8859cdf031e5b6b4).

chrisbra:
@tpope I suppose you can opt-out, but I suppose other maintainers that are not so proficient with vimscript may benefit from such a library eventually

tpope:
> @tpope I suppose you can opt-out, but I suppose other maintainers that are not so proficient with vimscript may benefit from such a library eventually

Sure, no objections on that front.

benknoble:
I've reverted changes to `runtime/ftplugin/ruby.vim`. I'll probably start drafting some documentation changes soon.

Interested parties I haven't heard from: @cecamp, @petdance

chrisbra:
I think this is fine. Thanks!

Please make some documentation changes in a followup PR

dkearns:
@chrisbra, IIUC this is going to cause Neovim to EOL these plugins because of the Vim9 script.

chrisbra:
oh, damn, I didn't think of it. I also notice, it breaks openeing of zip and gzip files, because by default the `zip_exec` variable is not set.

chrisbra:
I added 4f174f0de90b52937ddaf1e6db98e9731930ff7c on top of this. This should work for Vim and Neovim, right?

clason:
Thank you! We would probably have manually translated this back to vimscript, but this saves us (me) some effort now and down the line. I'll test asap and will let you know.

dkearns:
> I added [4f174f0](https://github.com/vim/vim/commit/4f174f0de90b52937ddaf1e6db98e9731930ff7c) on top of this. This should work for Vim and Neovim, right?

I think so.  You can test `has("vim9script")` rather than `exists(":def")` but that's not important at the moment.  See `:help vim9-mix` for some discussion on mixed version files.

@classon are you happy to have useless Vim9 script in your tree?  There's probably a reasonable way to organise it into multiple files so you don't have to if this library grows out into something significant.

clason:
> @clason are you happy to have useless Vim9 script in your tree? There's probably a reasonable way to organise it into multiple files so you don't have to if this library grows out into something significant.

As long as we don't have to maintain it, I don't see the harm. But it's probably cleaner to just omit the file since it'll never be called in Neovim. The main thing for us is to avoid needless changes that make it harder to merge in fixes from Vim (and ensuring consistent behavior so plugins can target Vim and Neovim at the same time).

Shane-XB-Qian:
tho #13487 claimed that's for older version vim, but this dist/vim.vim really would be used in older vim?
supposed this should would be released only with vim v9.0 in future.

anyway, the question is: 
would you really plan to maintain two copy of dist/vim.vim (and dist/vim9.vim) or maybe something similar things both?
// vim9script is hard developing by bram and yegappan (and others) now, if that's not be using mainly, looks a bit wasted?
// meant: i supposed either just write new and maintain existed in legacy vim script for now, or rewrite to vim9script later.

chrisbra:
It's a tricky question, but I need to consider Neovim devs needs as well. I know this may lead to subtle bugs between those two implementations, but I hope Neovim devs will be able to contribute back in case issues happen in the legacy vim script version compared to the vim9 script version.

benknoble:
Sorry for all the scrambling and disruption. And thanks to all for the documentation and other hotfix patches.

I thought from the commit message it was clear I didn't quite consider this ready for merge, but next time I will communicate that more clearly. In particular:

> Notable changes:
> - dist#vim because [sic] the (autoload) namespace for a new runtime support
>   library. Supporting functions should get documentation. It might make
>   life easier for NeoVim devs to make the documentation a new file
>   rather than cram it into existing files, though we may want
>   cross-references to it somewhere…

I think this is mostly addressed, though NeoVim folks might want to adjust the new docs a bit to avoid vim9?

Also, I just noticed that "because" should have been "becomes."

> - The gzip and zip plugins need to be opted into by enabling execution
>   of those programs (or the global plugin_exec). This needs
>   documentation or discussion.

I think this is addressed by Christian's patch that defaults them to enabled.

PR #13414, opened 2023-10-24
xxd: Fix that color was disabled on Cygwin

k-takata:
"windows.h" was unintentionally included on Cygwin since 9.0.1834. This accidentally disabled coloring on Cygwin.
Stop including "windows.h" on Cygwin.


chrisbra:
thanks!

PR #13416, opened 2023-10-24
Remove P_EXPAND flag from the 'filetype' option

dkearns:
Problem:  P_EXPAND flag is erroneously set for 'filetype' option
Solution: Remove the P_EXPAND flag

This flag is used by string options that accept file path values.  See
:help set_env.

This appears to have been included in d5e8c92 and resulted from an
incorrect implementation of 'filetype' completion.

See #12900 for a small discussion.


chrisbra:
thanks!

PR #13419, opened 2023-10-25
Add support for parsing a buffer number using '%b' in 'errorformat'

yegappan:


chrisbra:
I am not understanding the last test. Whats happening when buffer numbers are used in the quickfix list, that do not exist yet? Can this please be added to the documentation?

yegappan:
> I am not understanding the last test. Whats happening when buffer numbers are used in the quickfix list, that do not exist yet? Can this please be added to the documentation?

If a buffer with the specified number is not found, then those lines are treated as non-error lines.  I will update the help text.

chrisbra:
thanks, makes sense!

PR #13423, opened 2023-10-25
Debian runtime files update

### LINKED ISSUE #11934: Syntax filetypes: add support for deb822-formatted Debian source-list files

jamessan:
- Update Debian runtime files
- Add myself as codeowner for Debian-related runtime files

Closes #11934

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -124,6 +124,9 @@ runtime/ftplugin/cs.vim			@nickspoons
 runtime/ftplugin/csh.vim		@dkearns
 runtime/ftplugin/css.vim		@dkearns
 runtime/ftplugin/cucumber.vim		@tpope
+runtime/ftplugin/debchangelog.vim	@jamessan
+runtime/ftplugin/debcontrol.vim		@jamessan
+runtime/ftplugin/debsources.vim		@jamessan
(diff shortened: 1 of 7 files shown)


chrisbra:
thanks!

jayaddison:
Thank you @jamessan!

PR #13425, opened 2023-10-26
Added the objdump file/text format

ColinKennedy:
This PR adds filetype recognition for [Objdump](https://man7.org/linux/man-pages/man1/objdump.1.html) output files.

About objdump's file extensions, `.objdump` [seems to be the most common](https://github.com/search?q=path%3A**%2F*.objdump&type=code) among GitHub users. GitHub linguist claims that there's [several other recognized extensions](https://github.com/github-linguist/linguist/blob/7ca3799b8b5f1acde1dd7a8dfb7ae849d3dfb4cd/lib/linguist/languages.yml#L1278-L1285) such as `.c++-objdump`, `.cpp-objdump`, `.cppobjdump`, and the like but they seemed to be pretty rare. I can include those as well if you'd like.

### About runtime/ftplugin/objdump.vim

Though a bit uncommon, objdump files do have auto-generated comments so I've added that to `runtime/ftplugin/objdump.vim`. You can see a couple examples here: https://github.com/ColinKennedy/tree-sitter-objdump/blob/master/test/corpus/comment.txt

Please let me know if you'd like me to include anything for this PR

### DIFF
--- a/runtime/ftplugin/objdump.vim
+++ b/runtime/ftplugin/objdump.vim
@@ -1,0 +1,18 @@
+" Vim filetype plugin file
+" Language:     Objdump
+" Maintainer:   Colin Kennedy <colinvfx@gmail.com>
+" Last Change:  2023 October 25
+
+if exists("b:did_ftplugin")
+  finish
+endif
+
+let s:cpo_save = &cpo
+set cpo&vim
+
+let b:did_ftplugin = 1
+
+setlocal commentstring=#\ %s
+
+let &cpo = s:cpo_save
+unlet s:cpo_save
--- a/runtime/ftplugin/objdump.vim
+++ b/runtime/ftplugin/objdump.vim
@@ -7,12 +7,8 @@ if exists("b:did_ftplugin")
   finish
 endif
 
-let s:cpo_save = &cpo
-set cpo&vim
-
 let b:did_ftplugin = 1
 
-setlocal commentstring=#\ %s
+let b:undo_ftplugin = "setlocal cms<"
 
-let &cpo = s:cpo_save
-unlet s:cpo_save
+setlocal commentstring=#\ %s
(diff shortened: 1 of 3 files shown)


runtime/ftplugin/objdump.vim
```diff
@@ -0,0 +1,18 @@
...
+" Language:     Objdump
+" Maintainer:   Colin Kennedy <colinvfx@gmail.com>
+" Last Change:  2023 October 25
+
+if exists("b:did_ftplugin")
+  finish
+endif
+
+let s:cpo_save = &cpo
+set cpo&vim
```
dkearns: This `cpo` save/restore dance is only needed if using line continuations.  So lines 10,11,17,18 can be deleted.

runtime/ftplugin/objdump.vim
```diff
@@ -0,0 +1,18 @@
...
+
+let b:did_ftplugin = 1
+
+setlocal commentstring=#\ %s
```
dkearns: This option needs to be added to `b:undo_ftplugin`.

```vim
let b:undo_ftplugin = "setlocal cms<"
```

runtime/ftplugin/objdump.vim
ColinKennedy: Ah okay, yes there are no line continuations in objdump's grammar. It's a line-by-line format. Done!

runtime/ftplugin/objdump.vim
ColinKennedy: Done!

runtime/ftplugin/objdump.vim
chrisbra: It's not about objdump's grammar, it's more about whether you use line-continuation in this Vim script :)

chrisbra:
thanks. I have slightly amended your testfile however :)

PR #13432, opened 2023-10-27
fix: cmdheight increase overwrites messages that need hit-enter

ghost:
```vim
set laststatus=2 cmdheight=1
echo 'foo'
echo 'bar'
set cmdheight=2
```
Shows only hit-enter prompt.


chrisbra:
what exact problem does this solve? I cannot seem to see a difference with or without that patch.

nwounkn:
If you type `:echo "one\ntwo\nthree\nfour" | set ch=2` in an already running vim with `cmdheight=1`, without the patch the lines 'three' and 'four' aren't shown, with the patch the message is shown completely. This is to fix neovim's issue https://github.com/neovim/neovim/issues/22646 , but since vim is somewhat affected too, I decided to make the PR here.  
Sorry for bothering, if it's too insignificant to fix.

chrisbra:
hm, I not sure this is correct, I saw some rendering artifacts when using Vim with this patch and echoing something. Let me use this for a while and see if I can reproduce it or if this works correct.

ychin:
I have found some other issues with this type of usages (with or without this patch). For example, try running (with a clean config or make sure `cmdheight=1`) this: `:version | set ch=2`.

What you see is a normal version output, and after the version output it goes back to looking like `cmdheight` is 1, but it's actually set to 2. There are all kinds of weird things you can do after that, such as being able to set `cmdheight` to 0, which is disallowed in Vim, and in general there seems to be an off-by-one-error that is stuck in the `cmdheight` calculation.

Seems like in general the `command_height` function doesn't really work well when called when a message is shown. I think this patch will fix some of the issues but it may leave others unfixed. I wonder if it's better to just delay calling the entire `command_height` until hit-enter to prevent any logic inconsistencies because otherwise I find it pretty hard to reason about each state change piecemeal.

FWIW, what kind of situation does this come up @nwounkn ? I'm curious why we would see an echo message *and* a cmdheight set on the same line. It's still a bug, of course, but I just don't see this pattern often.

chrisbra:
yes, I assume this is some kind of a very specific problem. But it would be good if we could also fix the remaining problems if possible

PR #13439, opened 2023-10-28
Reset w_skipcol when unsetting 'wrap'

### LINKED ISSUE #12970: wrong cursor position when resetting 'wrap' on long wrapped line while 'smoothscroll' is set
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        ['', 'x'->repeat(2'000)]->setline(1)
        normal! G$
        &columns = 80
        &lines = 24
        &scrolloff = 1
        &showtabline = 2
        &smoothscroll = true
        timer_start(0, (_) => {
            &l:wrap = false
        })
    EOF
    )

The cursor is inside the tabline.

**Expected behavior**

The cursor is below the tabline.

**Version of Vim**

9.0 Included patches: 1-1826

**Environment**

Terminal: XTerm(353)

**Video**

https://github.com/vim/vim/assets/8505073/4debd99b-6fcd-40d3-94ad-54902db5722d

**Additional context**

It's not a regression; I can reproduce on 9.0.0641.

luukvbaal:
Problem:  With 'smoothscroll' set, "w_skipcol" is not reset when unsetting 'wrap'.
          Resulting in incorrect calculation of the cursor position.
Solution: Reset "w_skipcol" when unsetting 'wrap'.

fixes: #12970

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -4081,11 +4081,9 @@ did_set_showtabline(optset_T *args UNUSED)
     char *
 did_set_smoothscroll(optset_T *args UNUSED)
 {
-    if (curwin->w_p_sms)
-	return NULL;
+    if (!curwin->w_p_sms)
+	curwin->w_skipcol = 0;
 
-    curwin->w_skipcol = 0;
-    changed_line_abv_curs();
     return NULL;
 }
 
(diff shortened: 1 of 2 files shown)


chrisbra:
can we have a test for this?

src/option.c
```diff
@@ -4081,11 +4081,9 @@ did_set_showtabline(optset_T *args UNUSED)
     char *
 did_set_smoothscroll(optset_T *args UNUSED)
 {
-    if (curwin->w_p_sms)
-	return NULL;
+    if (!curwin->w_p_sms)
+	curwin->w_skipcol = 0;
 
-    curwin->w_skipcol = 0;
-    changed_line_abv_curs();
```
luukvbaal, line 4088: > can we have a test for this?

Added one. Also removed this `changed_line_abv_curs()` call for `did_set_smoothscroll()` because it is a `P_RWIN` option and it is already called in `changed_window_setting()`.

chrisbra:
thanks!

Issue #13438, opened 2023-10-27
`typename()` of type alias doesn't work inside `:def`


errael:
### Steps to reproduce

The error
```
typealias<object<C>>
Error detected while compiling :source buffer=1[9]..function <SNR>37_F:
line    1:
E1395: Type alias "Ctype" cannot be used as a variable
```
is given for
```
vim9script
class C
endclass
type Ctype = C

echo typename(Ctype)

def F()
    echo typename(Ctype)
enddef
F()
```

### Expected behaviour

No error.

### Version of Vim

9.0.2077

### Environment

ubuntu

Issue #13437, opened 2023-10-27
Innacurate `typealias` error message


errael:
### Steps to reproduce

The message
```
Error detected while compiling :source buffer=1[8]..function <SNR>37_F:
E1395: Type alias "Ctype" cannot be used as a variable
```
is inaccurate for the script, `Ctype` is not used as a variable
```
vim9script
class C
endclass
type Ctype = C
def F()
    var x = Ctype
enddef
F()
```

### Expected behaviour

A message that reflects that `Ctype` is being used as a value.

### Version of Vim

9.0.2077

### Environment

ubuntu

PR #13441, opened 2023-10-28
Check for more error conditions and add additional tests

### LINKED ISSUE #13434: `typename()` for typealias of class looks wrong
### Steps to reproduce

This output
```
C: class<C>
OldStyleMyC: class<C>
MyC: typealias<object<C>>
```
generated by following script
```
vim9script

class C
endclass

echo 'C:' typename(C)
const OldStyleMyC = C
echo 'OldStyleMyC:' typename(OldStyleMyC)
type MyC = C
echo 'MyC:' typename(MyC)
```

### Expected behaviour

Note the 3rd line of
```
C: class<C>
OldStyleMyC: class<C>
MyC: typealias<class<C>>
```

### Version of Vim

9.0.2077

### Environment

ubuntu

### Logs and stack traces

_No response_

yegappan:
Fixes #13434, #13437 and #13438 and supports using class type aliases to create new objects.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -749,7 +749,9 @@ type alias.  For Example: >
 	:type ListOfStrings = list<string>
 
 The type alias can be used wherever a built-in type can be used.  The type
-alias name must start with an upper case character.
+alias name must start with an upper case character.  A type alias can be
+created only at the script level and not inside a function.  A type alias can
+be exported and used across scripts.
 
 ==============================================================================
 
--- a/src/errors.h
+++ b/src/errors.h
@@ -3542,8 +3542,8 @@ EXTERN char e_type_can_only_be_defined_in_vim9_script[]
 	INIT(= N_("E1393: Type can only be defined in Vim9 script"));
 EXTERN char e_type_name_must_start_with_uppercase_letter_str[]
 	INIT(= N_("E1394: Type name must start with an uppercase letter: %s"));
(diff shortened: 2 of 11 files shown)


runtime/doc/vim9class.txt
```diff
@@ -749,7 +749,9 @@ type alias.  For Example: >
 	:type ListOfStrings = list<string>
 
 The type alias can be used wherever a built-in type can be used.  The type
-alias name must start with an upper case character.
+alias name must start with an upper case character.  A type alias can be
+created only at the script level and not inside a function.  A type alias can
+be exported and used across scripts.
```
errael, line 754: While it's true that "...used wherever a built-in type can be used." To avoid confusion, might want sentence about the limitations of how a type alias works with user defined types. In particular, a type alias of a class cannot be used wherever a **type** can be used.

For example
```
vim9script
class C
endclass
type Ctype = C
echo Ctype.new()
```

yegappan: What about changing that sentence to "..used wherever a built-in type is used to specify a variable type"?

errael: The existing wording is accurate.

My confusion came from not being able to use the type alias as the type; the behavior didn't meet my expectations, so I missed the implications when reading the help.

errael:
Oh, and it strikes me that that  wording isn't actually accurate. A type alias can be used with certain builtin functions.

chrisbra:
thanks!

PR #13445, opened 2023-10-28
[runtime, bash.vim] add shDblParen to shLoopList

grondilu:
This should allow code such as:

```bash
declare -i i j
for i in foo bar
do ((j = 1 << j))
done
```

to parse correctly


chrisbra:
Thanks,
does that need to be in an condition  `exists('b:is_bash')`? Or in other words, is this plain posix shell syntax or only valid for kornshell/bash?

grondilu:
> Thanks, does that need to be in an condition `exists('b:is_bash')`? Or in other words, is this plain posix shell syntax or only valid for kornshell/bash?

Not sure.

The following script:

```bash
for i in 1 2 3
do
   ((j = 1 << i))
   echo $j
done
```

seems to work fine when bash is given the `--posix` option, so my guess is it's fine.  I suppose to be sure we'd have to consurt IEEE 1003.1

chrisbra:
Well, it seems `dash` doesn't know `(( expression ))` but only `$(( expression ))`

grondilu:
> Well, it seems `dash` doesn't know `(( expression ))` but only `$(( expression ))`

Better safe than sorry, then.  I'll add an other commit.

chrisbra:
thanks

PR #13448, opened 2023-10-28
All secure options should note this restriction in the documentation

dkearns:
Problem:  Not all secure options document their status
Solution: Describe secure context :set restrictions in each help entry


chrisbra:
thanks!

PR #13449, opened 2023-10-28
Add Dart shebang filetype detection

dkearns:
Problem:  Not all Dart files detected
Solution: Add shebang filetype detection for Dart


chrisbra:
Thanks!

Issue #13451, opened 2023-10-29
`complete_info()` returns empty items list for some completion modes with active `noselect`


echasnovski:
### Steps to reproduce

1. Create init file with the following content:
```vim
set nocompatible
set completeopt=menu,noselect
inoremap <C-b> <Cmd>let g:complete_info=complete_info()<CR>
```
2. Open `vim` with `vim -u <path-to-config>`.
3. Start Insert mode with `i` and type `aa\nbb\n` creating two lines and ending on third.
4. Press `<C-x><C-l>` (or `<C-p>`). This should open completion menu.
5. Press `<C-b>` to save output of `complete_info()`.
6. Escape to Normal mode and type `:echo g:complete_info`. This shows `{'pum_visible': 1, 'mode': 'whole_line', 'selected': -1, 'items': []}`.

### Expected behaviour

Output of `complete_info()` always shows up to date information, i.e. `{'pum_visible': 1, 'mode': 'whole_line', 'selected': -1, 'items': [{'word': 'aa', 'menu': '', 'user_data': '', 'info': '', 'kind': '', 'abbr': ''}, {'word': 'bb', 'menu': '', 'user_data': '', 'i
nfo': '', 'kind': '', 'abbr': ''}]}`

This happens in the following cases:
- When completion is activated with `<C-n>` or `<C-x><C-n>`.
- When `noselect` is not part of `completeopt`.
- When built on 650dcfc8d12e68aa05a358301ec15f9e6dbd03ba indicating that daef8c74375141974d61b85199b383017644978c introduced the issue.

### Version of Vim

9.0.2081

### Environment

Operating system: EndeavourOS Linux x86_64 (6.5.9-arch2-1)
Terminal: st-0.9
Value of $TERM: st-256color
Shell: zsh 5.9

chrisbra:
hm, I see the problem

echasnovski:
Thank you, Christian!

PR #13452, opened 2023-10-29
Completion_info not filled

chrisbra:
Unfortunately, Commit daef8c74375141974d61b85199b383017644978c introduced a regression, that when ':set completeopt+=noselect' is set and no completion item has been selected yet, it did not fill the complete_info['items'] list.

This happened, because the current match item did not have the CP_ORIGINAL_TEXT flag set and then the cp->prev pointer did point to the original flag item, which caused the following while loop to not being run but being skipped instead.

So when the 'noselect' is set, only start with to the previous selection item, if the initial completion item has the CP_ORIGINAL_TEXT flag set, else use the 2nd previous item instead.

fixes: #vim/vim#13451

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3080,7 +3080,7 @@ info_add_completion_info(list_T *li)
 
     // Skip the element with the CP_ORIGINAL_TEXT flag at the beginning, in case of
     // forward completion, or at the end, in case of backward completion.
-    match = forward ? match->cp_next : (compl_no_select ? match->cp_prev : match->cp_prev->cp_prev);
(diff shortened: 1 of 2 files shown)


echasnovski:
This indeed now prints non-trivial `complete_info()` output.

There is one observation and I am not sure if it is intended. The output of #13451 repro now is `{'pum_visible': 1, 'mode': 'whole_line', 'selected': -1, 'items': [{'word': 'bb', 'menu': '', 'user_data': '', 'info': '', 'kind': '', 'abbr': ''}, {'word': 'aa', 'menu': '', 'user_data': '',  'info': '', 'kind': '', 'abbr': ''}]}`, i.e. first `bb`, then `aa`. While completion menu really is showing `aa` and then `bb`.

PR #13464, opened 2023-10-31
Disallow the use of abstract in an interface

### LINKED ISSUE #13456: abstract keyword allowed when declaring interface features
### Steps to reproduce

```vim
interface A
  abstract foo: number
  abstract def Bar()
endinterface
```
This compiles without error.

### Expected behaviour

An error indicating that `abstract` is not allowed when defining interface features.

### Version of Vim

9.0.2078

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Instead of silently ignoring the `abstract` keyword in an interface, give an error message.

Fixes #13456.


chrisbra:
thanks. Please verify, not sure I made an error, since there was a conflict

PR #13466, opened 2023-10-31
Disallow abstract static methods

yegappan:
Abstract static methods are not allowed in other OOP languages:

https://docs.oracle.com/javase/specs/jls/se9/html/jls-8.html#jls-8.4.3
https://www.reddit.com/r/java/comments/76d874/can_abstract_methods_be_declared_static/
https://stackoverflow.com/questions/20741737/can-one-declare-a-static-method-within-an-abstract-class-in-dart
https://github.com/microsoft/TypeScript/issues/14600

Disallow abstract static methods in Vim9 classes also. Fixes #13462.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -411,6 +411,8 @@ prefix when defining the method: >
 	   abstract static def SetColor()
 	endclass
 <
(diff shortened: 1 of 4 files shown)


chrisbra:
thanks, makes sense

PR #13467, opened 2023-10-31
Makefile for the Vim documentation on Windows

RestorerZ:
Preparation of documentation (not all) on Vim in Windows.

Sometimes I'm too lazy to run a UNIX VM, so I did it for myself, maybe it will be useful to someone else.


runtime/doc/vim2html.pl
```diff
@@ -1,16 +1,23 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
```
chrisbra: that looks wrong

runtime/doc/vim2html.pl
```diff
@@ -1,16 +1,23 @@
...
-$date = `date`;
-chop $date;
+# 30.11.23, Restorer:
+# This command does not work in OS Windows.
+# The "date" command in Windows is different from its counterpart in UNIX-like systems.
+# The closest analog is the "date /t" command, but how it would work in UNIX,
+# I don't know. I've corrected it as best I can. I don't know Perl.
+#$date = `date`;
+#chop $date;
+$date = localtime();
```
chrisbra: Can we format the localtime() command more like a normal `date` output?

chrisbra:
I think it is okay, except for the minor things in the perl file. 

Can you please translate the russian comments into English?

runtime/doc/vim2html.pl
RestorerZ: As I understand it, this is not really required for Windows OS. On the contrary, it is required for UNIX-like systems. But judging by [this comment](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my), the attitude to "env" is not quite unambiguous.
I did it as it was originally.

runtime/doc/vim2html.pl
RestorerZ: `$date = localtime(); # outputs like this Fri Nov 3 00:56:59 2023`
Which, as I recall, is about the same as the "date" command with no arguments.
I made that now the date is output in the format DD.MM.YYYY.

RestorerZ:
> Can you please translate the russian comments into English?
Done.

chrisbra:
okay thanks!

PR #13469, opened 2023-11-01
Update sh.vim

grondilu:
Add shDblParen to shCaseList


k-takata:
Cc @cecamp

chrisbra:
Thanks, I include it

PR #13470, opened 2023-11-01
configure.ac: Add --with-xsubpp configure option

zdohnal:
Some environments (such as flatpaks) cannot count on xsubpp being in the common Perl directory, so a configure option should be used for clean solution.

Would you mind adding it to the project?

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -836,6 +836,7 @@ with_luajit
 enable_mzschemeinterp
 with_plthome
 enable_perlinterp
+with_xsubpp
 enable_pythoninterp
 with_python_command
 with_python_config_dir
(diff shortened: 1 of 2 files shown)


chrisbra:
and what is that used for?

zdohnal:
@chrisbra AFAIK xsubpp is needed for building Perl plugin (Perl is now available as module, so flatpaks can use it during build, and GVim flatpak maintainer would like to have Perl plugin in GVim flatpak) and xsubpp is not part of runtime (kind of buildroot for flatpaks), so it will be provided by other module during build, where it is not guaranteed that xsubpp will be in common Perl dirs, which are hardcoded right now, and will be somewhere else - f.e. in /usr/bin.

Flatpak people would like to pass ```$(which xsubpp)``` to ```vi_cv_perl_xsubpp``` variable in configure, which does the trick - Vim gets the current xsubpp location - but I thought having this as proper configure option would be clearer. I've added fallback to checking common directories if the value provided via option does not exist as a file as well.

chrisbra:
okay, thanks. If you think this is useful, then so be it :)

PR #13471, opened 2023-11-01
`sound_playfile()` fails on windows if 'shell' is powershell

MiguelBarro:
I detected testing [a plugin](https://github.com/AndrewRadev/typewriter.vim) that `sound_playfile()` was not working.
After some testing I found out that the windows [API calls](https://learn.microsoft.com/en-us/windows/win32/multimedia/mci) used to test the code were failing.
The windows MCI API receives a [command string](https://learn.microsoft.com/en-us/windows/win32/multimedia/open) where the sound file to open is specified. The faulty command was:
```
    open 'C:\Users\xxxxx\vimfiles\plugin\typewriter.vim\autoload/../sounds/carriage1.wav' alias sound000001
    Error: Specify a device or driver name that is less than 79 characters.
```
but it works if I do either:
```
    open C:\Users\xxxxx\vimfiles\plugin\typewriter.vim\autoload/../sounds/carriage1.wav alias sound000001
```
or:
```
    open "C:\Users\xxxxx\vimfiles\plugin\typewriter.vim\autoload/../sounds/carriage1.wav" alias sound000001
```
The second option is preferable to allow whitespaces in the file name.
After some reviewing I found that an update in the `shellescape()` was to blame. It was updated for powershell where no escaping is necessary if single quotes are used, thus, cmd and powershell differ in escaping.
Nevertheless escaping in this case is unnecessary because the MCI interpreter doesn't rely in the shell.

### DIFF
--- a/src/sound.c
+++ b/src/sound.c
@@ -376,7 +376,7 @@ f_sound_playfile(typval_T *argvars, typval_T *rettv)
 {
     long	newid = sound_id + 1;
     size_t	len;
-    char_u	*p, *esc;
+    char_u	*p, *filename;
     WCHAR	*wp;
     soundcb_T	*soundcb;
     char	buf[32];
@@ -385,17 +385,15 @@ f_sound_playfile(typval_T *argvars, typval_T *rettv)
     if (in_vim9script() && check_for_string_arg(argvars, 0) == FAIL)
 	return;
 
-    esc = vim_strsave_shellescape(tv_get_string(&argvars[0]), FALSE, FALSE);
+    filename = tv_get_string(&argvars[0]);
 
-    len = STRLEN(esc) + 5 + 18 + 1;
+    len = STRLEN(filename) + 5 + 18 + 2 + 1;
     p = alloc(len);
     if (p == NULL)
     {
-	free(esc);
 	return;
     }
-    vim_snprintf((char *)p, len, "open %s alias sound%06ld", esc, newid);
-    free(esc);
+    vim_snprintf((char *)p, len, "open \"%s\" alias sound%06ld", filename, newid);
 
     wp = enc_to_utf16((char_u *)p, NULL);
     free(p);
--- a/src/sound.c
+++ b/src/sound.c
@@ -322,7 +322,7 @@ sound_wndproc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 
 		    vim_snprintf(buf, sizeof(buf), "close sound%06ld",
 								p->snd_id);
-		    mciSendString(buf, NULL, 0, 0);
+		    mciSendStringA(buf, NULL, 0, 0);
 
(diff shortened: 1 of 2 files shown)


src/sound.c
```diff
@@ -385,30 +385,23 @@ f_sound_playfile(typval_T *argvars, typval_T *rettv)
...
+    vim_snprintf((char *)p, len, "open \"%s\" alias sound%06ld", filepath, newid);
 
-    wp = enc_to_utf16((char_u *)p, NULL);
+    err = mciSendStringA(p, NULL, 0, sound_window());
     free(p);
-    if (wp == NULL)
-	return;
-
-    err = mciSendStringW(wp, NULL, 0, sound_window());
-    free(wp);
```
k-takata: This part should be kept. Otherwise, Unicode filenames cannot be handled.

chrisbra:
is the change from `mciSendString` to `mciSendStringA` intentionally? It seems this is for ASCII only? Or did I google wrong?

src/sound.c
MiguelBarro: Right. I thought that now that windows and linux use utf-8 as multibyte character set that was no longer required. But after some testing is clear that I was wrong.

k-takata:
> is the change from `mciSendString` to `mciSendStringA` intentionally? It seems this is for ASCII only? Or did I google wrong?

`UNICODE` is not defined in the Vim source code. So, `mciSendString` and `mciSendStringA` are exactly the same.

MiguelBarro:
@chrisbra Windows has a funny way of managing UNICODE. In windows NT they decided to migrate all API calls from ANSI to UNICODE so they added a `W` at the end. For example `CreateWindowEx` became `CreateWindowExW` and all the `char` arguments were promoted to `wchar_t`.
In order to keep backward compatibility the old functions were renamed appending `A` at the end. Thus the old `CreateWindowEx` was renamed as `CreateWindowExA`. The old `A` functions translate all arguments to UNICODE (utf-16) and call the new `W` ones.
Finally they decided to keep the old function names as macros:
```c++
#ifdef UNICODE
#    define TCHAR char
#    define CreateWindowEx CreateWindowExW
#else
#    define TCHAR wchar_t
#    define CreateWindowEx CreateWindowExA
#endif
```
I guess the idea was to be able to get an ANSI or UNICODE program just by introducing the UNICODE define and simplify portability.

src/sound.c
```diff
@@ -376,7 +376,7 @@ f_sound_playfile(typval_T *argvars, typval_T *rettv)
 {
     long	newid = sound_id + 1;
     size_t	len;
-    char_u	*p, *filepath;
+    char_u	*p, *esc;
```
k-takata: It looks like you reverted some necessary lines.

MiguelBarro: You are right. I've squashed the unicode fix and retest the whole thing. Now it should be right.

chrisbra:
okay, thanks

PR #13473, opened 2023-11-02
Mark tests as skipped when ch_open encounters E901

jamessan:
On some of the Debian build systems, the IPv6 channel tests fail because `ch_open('[::1]:<port>', ...)` raises the error "E901: getaddrinfo() in channel_open(): Address family for hostname not supported".

This appears to happen because `getaddrinfo()` can't perform the reverse lookup for the ::1, which is a config issue on that system.  Therefore, instead of reporting a test failure, mark the test as skipped due to the bad network config.

### DIFF
--- a/src/testdir/shared.vim
+++ b/src/testdir/shared.vim
@@ -113,6 +113,8 @@ func RunServer(cmd, testfunc, args)
     endif
 
     call call(function(a:testfunc), [port])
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks, I was first wondering, why `CheckIPv6Loopback()` didn't catch it, but since this seems to be a system configuration issue I agree it is fine to skip in your case.

PR #13474, opened 2023-11-02
feat: beauty termdebug var/asm window display and shorten var type name

Shane-XB-Qian:


chrisbra:
can you show a screenshot of how this currently behaves and how it will change?

Shane-XB-Qian:
1. e.g type from golang it maybe like foo.com/bar/babababababa/what.type, let's shortten to what.type
2. if your display screen is wide enough e.g 78 * 2, then this would auto adjust the asm/var window (when opening it) to 'side' window (vs 'below') since generally it might be too tight to those windows if such layout.

chrisbra:
so can you please show a screenshot?

chrisbra:
and please add a test for this

src/testdir/test_termdebug.vim
```diff
@@ -80,6 +80,24 @@ func Test_termdebug_basic()
...
+    Var
+    call assert_equal(winnr(), winnr('$'))
+    bw!
+    Asm
+    call assert_equal(winnr(), winnr('$'))
+    bw!
+  endif
+  set columns=160
+  Var
+  call assert_equal(winnr(), winnr('$') - 1)
```
yegappan, line 100: You should also check the window is split vertically using winlayout().

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1420,8 +1421,15 @@ endfunc
 
 func s:GotoAsmwinOrCreateIt()
   if !win_gotoid(s:asmwin)
+    let mdf = ''
     if win_gotoid(s:sourcewin)
-      exe 'rightbelow new'
+      " 60 is approx spaceBuffer * 3
+      if winwidth(0) > (78 + 60)
+        let mdf = 'vert'
+        exe mdf .. ' ' .. 60 .. 'new'
```
yegappan, line 1429: Instead of hard-coding the window width as 60, can you add a new configuration item to g:termdebug_config and use that here?  You can use g:termdebug_config['asm_winwidth'] and g:termdebug_config['var_winwidth'].  The default for these can be 60.

yegappan (changes requested):

src/testdir/test_termdebug.vim
Shane-XB-Qian: ok, done.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
Shane-XB-Qian: " 60 is approx spaceBuffer * 3
it is fixed **minimal** value according to 'spaceBuffer' (of 'Var').

PR #13475, opened 2023-11-02
fix: termdebug mappings maybe not functional and spreading and un-recoverable

Shane-XB-Qian:
// e.g those mappings maybe remained there even had quit termdebug.

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1462,17 +1462,20 @@ If there is no g:termdebug_config you can use: >
 	let g:termdebug_use_prompt = 1
 <
 						*termdebug_map_K*
-The K key is normally mapped to |:Evaluate|. If you do not want this use: >
+The K key maybe mapped to |:Evaluate| if it was no e.g buffer local mapping
+stop it.  If you do not want this use: >
 	let g:termdebug_config['map_K'] = 0
 If there is no g:termdebug_config you can use: >
 	let g:termdebug_map_K = 0
 <
 						*termdebug_map_minus*
-The - key is normally mapped to |:Down|. If you do not want this use: >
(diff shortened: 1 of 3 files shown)


Shane-XB-Qian:
this PR is not going to change the behavior, but just make sure it was really working, 
e.g there was buf local 'K' already, that old code logic of that mapping actually was never really worked,
and worse: oppositely it would remained there together with your buf local 'K', this PR is going to clear it really after quit termdebug.

chrisbra:
I don't follow. Take this example here:
```
    if !empty(s:k_map_saved) && !s:k_map_saved.buffer || empty(s:k_map_saved)
      nnoremap K :Evaluate<CR>
    endif
```
This means, `K` won't be mapped when there is a buffer local mapping. I am not sure, if this can happen but who knows...

So what problem is this solving? Can you give an example, when the plugin currently does not restore the mapping correctly then?

Shane-XB-Qian:
> This means, `K` won't be mapped when there is a buffer local mapping. I am not sure, if this can happen but who knows...

yes, it can happen, but if user had buf local 'K' mapping already, then better not map another global 'K', it is meaningless;
and actually current global 'K' would be remained there if really in such situation, the global 'K' was not unmap when quit termdebug;
or IF really overrode that buf local 'K', i think it's difficult to revert it from all your visited buffers, or again meaningless.


> So what problem is this solving? Can you give an example, when the plugin currently does not restore the mapping correctly then?

it is a global 'K', but your debug probably just for one specific ft, i supposed it's easy to understand, it would impact sth.
// please check above `e.g`, those are the examples.

chrisbra:
hm okay. Can you please add a test for what you changed?

chrisbra:
thanks for adding the test. I'll check it out soon.

PR #13476, opened 2023-11-02
fixup vim9class.c:  patch 9.0.2085

pheiduck:
https://github.com/vim/vim/commit/2b358adde079d7cc8a7f42fcd5627866641c9a59#diff-8e2956e6a4ccbdea6b0afe8c3941b810c32aaa94e11b404d5e7aa1a01d953a4c


tonymec:
I don't know if this is the right way to fix the build failure, but I confirm that something at patchlevel 9.0.2086 or shortly before it causes a build error at vim9class.c:1563 (called by vim9class.c:15) in builds with `+eval`.

Best regards,
Tony.

PR #13479, opened 2023-11-03
Disallow abstract static methods

yegappan:
Add back the changes that were lost because of the merge conflict between the PR #13466 and PR #13456.


chrisbra:
thanks!

PR #13486, opened 2023-11-04
Fix the regression caused by 9.0.2072

### LINKED ISSUE #13485: [vim9script] Regression: Cannot convert list to string when using += et al operators with dictionary inside function
### Steps to reproduce

Source this script:

```vim
vim9script

def F()
  var d: dict<dict<any>> = {}

  for group in keys(d)
    d['a']['b'] += 1
  endfor
enddef

F()
```

This results in `E1105: Cannot convert list to string`. The error does not happen if the affected code is not wrapped in a function, or if `+=` is replaced by normal assignment.

It's a regression introduced in c229a6ac0775e07dff456ca8832c516e57a74e74.

### Expected behaviour

The code should compile and run without errors.

### Version of Vim

9.0.2090

### Environment

macOS 14.1
Apple Terminal
xterm-256color
ZSH 5.9

### Logs and stack traces

_No response_

yegappan:
The instruction to convert the dict index to a string is generated later in the compile_member() function.  So no need to generate that in compile_load_lhs().

Fixes #13485.

### DIFF
--- a/src/testdir/test_vim9_assign.vim
+++ b/src/testdir/test_vim9_assign.vim
@@ -2992,13 +2992,67 @@ def Test_list_item_assign()
     vim9script
 
     def Foo()
-        var l: list<list<string>> = [['x', 'x', 'x'], ['y', 'y', 'y']]
-        var z: number = 1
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #13487, opened 2023-11-05
Make `autoload/dist/vim.vim` work properly when lacking vim9script support

seandewar:
`:return` cannot be used outside of `:function` in older Vims lacking Vim9script support or in Neovim, even when evaluation is being skipped in the dead `:else` branch.

Instead, use the pattern described in `:h vim9-mix`, which uses `:finish` to end script processing before it reaches the Vim9script stuff.


chrisbra:
yeah, thanks. Makes sense.

PR #13488, opened 2023-11-05
statusline fillchar may be different than expected

### LINKED ISSUE #13366: 'stl' fillchar not respected when linking StatusLineNC to StatusLine
### Steps to reproduce

minimal.vim:

```
filetype plugin indent on
syntax on
set fillchars=stl:\ ,stlnc:\  
split
" when this is commented out there is no problem
hi! link StatusLineNC StatusLine
```

`vim --clean -u minimal.vim`

Notice the '^' character is used for stl, even though we've set both stl and stlnc to be a blank character.

### Expected behaviour

Vim should always use the user set fillchars for 'stl' and 'stlnc' if they were set by the user.

### Version of Vim

9.0.2040

### Environment

macOS
Wezterm nightly
'wezterm'
zsh

### Logs and stack traces

_No response_

chrisbra:
If the highlighting group for the statusline for the current window |hl-StatusLine| or the non-current window |hl-StatusLineNC| are cleared (or do not differ from each other), than Vim will use the hard-coded fallback values '^' (for the non-current windows) or '=' (for the current window).  I believe this was done, to make sure the statusline will always be visible and be distinguishable from the rest of the window.

However, this may be unexpected, if a user explicitly defined those fillchar characters just to notice that those values are then not used by Vim.

So, let's assume users know what they are doing and just always return the configured stl and stlnc values.  And if they want the statusline to be non-distinguishable from the rest of the window space, so be it.  It is their responsibility and Vim shall not know better what to use.

fixes vim/vim#13366


RestorerZ:
@chrisbra 

> So, let's assume users know what they are doing and just always return the configured stl and stlnc values. And if they want the statusline to be non-distinguishable from the rest of the window space, so be it. It is their responsibility and Vim shall not know better what to use.

Are you sure users know what they're doing?
Unfortunately, that's not always the case (knowing from what I've done myself).
I agree that it should be possible to choose, not depend on what the developer has programmed. But it seems to me that at the very least there should be some kind of warning (notification) that things are not as they should be.

chrisbra:
Not sure we should add a warning. First it's not clear where to set it and when. And I am sure people would complain about seeing a warning. Perhaps we can only show it using `:verbose` but then it may again go un-noticed. Also from the linked ticket, it seems that nobody complained about this behaviour over at Neovim, so I think it should be okay. 

We can still add a warning, if people find this behaviour confusing later I think

RestorerZ:
@chrisbra 

> We can still add a warning, if people find this behaviour confusing later I think

Then it's not clear that

> First it's not clear where to set it and when.

I too have been stupefied by this behavior at times. I hope that there will be no complaints and everything will work as it should.

craigmac:
> Are you sure users know what they're doing?

Well, if user has set stl or stlnc fillchars, then we have to assume they do, right? Otherwise, you could say this about any option that was set by the user.

PR #13499, opened 2023-11-07
prevent if_interp option skips Test_modeline_fails_always()

svardew:
Test_modeline_fails_always() skipped at some options.
Seeing github CI log, skipped at 4 options: luadll, mzschemedll, fsync, cscopeprg.
Language interpreter (if_interp) is default disabled, user explicity selective feature.
It is not good that Test skipped due to absence of intrep-related option for default build.
And this Test expect interpreter dll, but CheckOption() not consider whether interpreter is dynamic.

this PR move interp-related options into new Test and check interpreter is dynamic or not.
- list +interp/dyn feature from :version output.
- if no +interp/dyn feature exist, skip Test and log 'no interpreter enabled.' to messages
- if any +interp/dyn feature exist, test all options. each test proceed check if dynamic interpreter
- pyxversion option is without check if dyanmic interp. this option works both static and dynamic

Unfortunately, some other option still skip Test_modeline_fails_always(), t_XX option not tested.
And looking optiondefs.h, some P_SECURE options missing from this test.


zeertzjq:
I think what should be done here is to revert the change from patch 8.2.1432.

chrisbra:
thanks, this makes sense

PR #13500, opened 2023-11-07
Unsupported option causes rest of modeline test to be skipped

zeertzjq:
Problem:  Unsupported option causes rest of modeline test to be skipped.
Solution: Revert the change from patch 8.2.1432.

Close #13499


chrisbra:
thanks, that makes sense!

PR #13502, opened 2023-11-08
Recursive callback limit causes problems on some architectures

### LINKED ISSUE #13495: Test_popup_close_callback_recursive occasionally crashes Vim
### Steps to reproduce

Since the Debian package got v9.0.2030, I've seen some Debian builds fail due to Vim crashing while running this test.  It's been particularly problematic on one of the armel (32-bit ARM) systems.

I was able to reproduce it on one of the armel systems and this seems to be the problematic part of the stack:

```
#28 0x0061d98c in apply_autocmds (event=event@entry=EVENT_VIMLEAVEPRE, fname=fname@entry=0x0, fname_io=fname_io@entry=0x0, force=force@entry=0, buf=<optimized out>) at autocmd.c:1779
#29 0x008d98a0 in getout (exitval=1) at main.c:1691
#30 <signal handler called>
#31 0x006458d4 in dict_unref (d=0xffffffff) at dict.c:178
#32 0x00845f40 in clear_tv (varp=varp@entry=0xffe68438) at typval.c:142
#33 0x007791c4 in invoke_popup_callback (result=0xffe68798, wp=<optimized out>) at popupwin.c:2401
#34 popup_close_and_callback (wp=<optimized out>, arg=0xffe68798) at popupwin.c:2466
#35 0x00683754 in call_internal_func (name=name@entry=0x0, argcount=2, argcount@entry=0, argvars=0xffe68788, argvars@entry=0x2, rettv=0xffe68778, rettv@entry=0xffe68788) at evalfunc.c:3145
#36 0x008602d4 in call_func (funcname=0x0, funcname@entry=0x1d21cf0 "popup_close", len=len@entry=-1, rettv=0xffe68788, rettv@entry=0xffe68778, argcount_in=argcount_in@entry=2,
    argvars_in=<optimized out>, funcexe=<optimized out>) at userfunc.c:3925
#37 0x00860b08 in call_callback (callback=<optimized out>, len=len@entry=-1, rettv=rettv@entry=0xffe68778, argcount=argcount@entry=2, argvars=<optimized out>, argvars@entry=0xffe68780)
    at userfunc.c:3603
```

Note that frame 31 shows the `dict_T *` we're trying to unref has the value `0xffffffff`, which causes a segfault when we try to dereference it.

Looking at frame 37, this is the 100th invocation of the callback, so it seems like we are on the edge of tripping some condition (low memory?) and when the conditions are wrong, we do and crash.

Not sure if there should be a separate `'mfd'` value depending on the platform (32-bit vs 64-bit), lowered in general, or if there's some other issue here.

### Expected behaviour

No crash

### Version of Vim

9.0.2087

### Environment

Debian

### Logs and stack traces

_No response_

chrisbra:
Since commit 47510f3d6598a1218958c03ed11337a43b73f48d we have a test that causes a recursive popup callback function to be executed. However it seems the current limit of 'maxfuncdepth' option value is still too recursive for some 32bit architectures.

So instead of allowing a default limit of 100 (default value for 'maxfuncdepth'), let's reduce this limit to 20. I don't think there is a use case where one would need such a high recursive callback limit and a limit of 20 seems reasonable (although it is currently hard-coded).

closes: #13495

### DIFF
--- a/runtime/doc/message.txt
+++ b/runtime/doc/message.txt
@@ -1,4 +1,4 @@
-*message.txt*   For Vim version 9.0.  Last change: 2023 May 24
+*message.txt*   For Vim version 9.0.  Last change: 2023 Nov 08
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -135,6 +135,8 @@ This happens when an Ex command executes an Ex command that executes an Ex
 command, etc.  The limit is 200 or the value of 'maxfuncdepth', whatever is
 larger.  When it's more there probably is an endless loop.  Probably a
 |:execute| or |:source| command is involved.
+Can also happen with a recursive callback function (|job-callback|).
+A limit of 20 is used here.
 
 							*E254*
   Cannot allocate color {name} ~
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1,4 +1,4 @@
-*options.txt*	For Vim version 9.0.  Last change: 2023 Oct 23
+*options.txt*	For Vim version 9.0.  Last change: 2023 Nov 08
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
@@ -5489,7 +5489,6 @@ A jump table for the options with a short description can be found at |Q_op|.
 	Increasing this limit above 200 also changes the maximum for Ex
 	command recursion, see |E169|.
 	See also |:function|.
-	Also used for maximum depth of callback functions.
 
 						*'maxmapdepth'* *'mmd'* *E223*
 'maxmapdepth' 'mmd'	number	(default 1000)
--- a/src/userfunc.c
+++ b/src/userfunc.c
@@ -14,6 +14,9 @@
 #include "vim.h"
 
 #if defined(FEAT_EVAL) || defined(PROTO)
+
+#define MAX_CALLBACK_DEPTH 20
+
 /*
  * All user-defined functions are found in this hashtable.
  */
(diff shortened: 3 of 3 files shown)


tonymec:
Would it be very difficult to keep the "high" default when compiled for 64 bit architectures ?

chrisbra:
no, would be possible

runtime/doc/message.txt
```diff
@@ -135,6 +135,8 @@ This happens when an Ex command executes an Ex command that executes an Ex
 command, etc.  The limit is 200 or the value of 'maxfuncdepth', whatever is
 larger.  When it's more there probably is an endless loop.  Probably a
 |:execute| or |:source| command is involved.
+Can also happen with a recursive callback function (|job-callback).
```
zeertzjq:
```suggestion
Can also happen with a recursive callback function (|job-callback|).
```

chrisbra:
I checked briefly the standard [pre-defined compiler macros](https://github.com/cpredef/predef/blob/master/Architectures.md)  and there doesn't seem to be one single macro which we can use to detect 32bit vs. 64bit. So I would rather go with a default callback limit of 20 for everywhere. It's not clear to me, why one would need a higher limit anyhow. The test itself is a pathological test-case and we just want to make sure we do not crash.

jamessan:
> I checked briefly the standard [pre-defined compiler macros](https://github.com/cpredef/predef/blob/master/Architectures.md) and there doesn't seem to be one single macro which we can use to detect 32bit vs. 64bit.

There's a separate [endian](https://github.com/cpredef/predef/blob/master/Endianness.md#detect-endianness-at-compile-time) page, but it's probably simpler to just have consistent behavior.

PR #13505, opened 2023-11-08
runtime(doc): fix grammar in termdebug doc, remove trailing spaces

zeertzjq:


chrisbra:
damn, I was trying to carefully write proper English :(

chrisbra:
thanks, I include it

Shane-XB-Qian:
> damn, I was trying to carefully write proper English :( 

TBH i didnot feel yours vs his that modification had much difference, or even mine,
all meant if a buf local mapping was existed already,
damn, so horrible to write 'proper English', ahaha~

-- 
shane.xb.qian

wilhelmtell:
The original was fine English, there were no grammar errors. I agree though that all things being equal in writing like in programming less is more. One word shorter? Sure I'll take that.

zeertzjq:
Ah yes, actually it was fine, but the `map-local` tag should be `:map-local`.

PR #13506, opened 2023-11-09
labeler.yml: add documentation

pheiduck:


chrisbra:
thanks!

.github/labeler.yml
```diff
@@ -8,6 +8,9 @@ CI:
 - any: ['.appveyor.yml']
 - any: ['.codecov.yml']
 
+documentation:
+- any: ['*/doc/*']
```
zeertzjq, line 12: I think this should be `all:`, otherwise any PR that updates documentation will have this label, which doesn't look right.

pheiduck: Done: https://github.com/vim/vim/pull/13507

PR #13507, opened 2023-11-09
labeler.yml: doc: use all instead of any

pheiduck:
fixup: https://github.com/vim/vim/pull/13506#pullrequestreview-1722013761


chrisbra:
thanks!

PR #13510, opened 2023-11-10
runtime(lynx): Update for Lynx 2.8.9

dkearns:


chrisbra:
Thanks, but why did you remove the `:set cpo` settings?

dkearns:
Because my local versions are in Vim9 script and I forgot to do the 'cpo' dance when converting them back to legacy script.

Sorry about that, I'll update it later today.

chrisbra:
Thank you!

PR #13516, opened 2023-11-10
Add cypher filetype support

meistermeier:
Cypher query language support to work with graph databases.

Already existing lsp support in Neovim's nvim-lspconfig: https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#cypher_ls


chrisbra:
thanks

PR #13518, opened 2023-11-10
Modified and extended makefiles and related files for .po files

RestorerZ:
1. Modified and extended "make_mvc.mak" file to be able to perform almost all operations with PO files (excluding *.desktop).
2. To the "Makefile" file added a rule for converting zh_TW.po with backslash characters (the rule is commented out) and a refinement to the "update-po" rule.
3. Added file big5corr.c
4. Clarifications in readme*.txt files regarding "make_all.mak".
5. Other minor fixes.


RestorerZ:
@chrisbra ,
Do you think it might be worth adding to the readme*.txt explanations about $(MOCONVERTED) and when and what to add there for new (and old) languages?

I didn't do it, firstly, because I write in English even worse than I read in English and, secondly, I'm not sure if it should be done.

src/po/big5corr.c
```diff
@@ -0,0 +1,61 @@
+
```
k-takata: Please remove this empty line.

src/po/Make_mvc.mak
```diff
@@ -53,43 +517,73 @@ PO_INPUTLIST = \
...
 	$(RM) *.js
 
+# 06.11.23, added by Restorer
+# When updating ja.sjis.po there are a bunch of errors and a crash.
+# The files that are converted to a different encoding clearly state "DO NOT EDIT".
+update-po : $(MOFILES:.mo=)
+
 # Don't add a dependency here, we only want to update the .po files manually
-$(LANGUAGES):
+$(LANGUAGES) :
```
k-takata: Please revert this line (and similar other changes).
Space is not inserted between the filename and the colon in the most of the other makefiles,

src/po/Makefile
```diff
@@ -155,12 +155,42 @@ zh_TW.po: zh_TW.UTF-8.po
...
+#	$(CC) -o big5corr big5corr.c
+
+
 # Convert ko.UTF-8.po to create ko.po.
 ko.po: ko.UTF-8.po
 	rm -f ko.po
 	iconv -f UTF-8 -t euc-kr ko.UTF-8.po | \
 		$(SED) -e 's/charset=UTF-8/charset=euc-kr/' \
-		    -e 's/# Korean translation for Vim/# Generated from ko.UTF-8.po, DO NOT EDIT/' \
+		    -e 's/# Original translations/# Generated from ko.UTF-8.po, DO NOT EDIT/' \
```
k-takata: `ko.po` should be also updated.

src/po/big5corr.c
```diff
@@ -0,0 +1,61 @@
...
+/*
+ * Simplistic program to correct Big5 inside strings.  When a trail byte is a
+ * backslash it needs to be doubled.
+ * Public domain.
+ * Time 2018-10-20 22:37:05
+ * Author K.Takata <kentkt@csc.jp>
```
k-takata: These two lines are not needed. (sjiscorr.c doesn't have such lines.)

src/po/Make_mvc.mak
```diff
@@ -34,13 +68,443 @@ LSFLAGS = /b /on /l /s
...
+
+
+# 06.11.23, added by Restorer
+# Convert zh_TW.UTF-8.po to create zh_TW.po with backslash characters
+# Requires doubling backslashes in the second byte.  Don't depend on big5corr,
+# it should only be compiled when zh_TW.po is outdated.
+#
+#  For more details, see:
+#  https://github.com/vim/vim/pull/3476
+#  https://github.com/vim/vim/pull/3476
```
k-takata: Duplicated.

src/po/Makefile
```diff
@@ -155,12 +155,42 @@ zh_TW.po: zh_TW.UTF-8.po
...
+
+# 06.11.23, added by Restorer
+# Convert zh_TW.UTF-8.po to create zh_TW.po with backslash characters
+# Requires doubling backslashes in the second byte.  Don't depend on big5corr,
+# it should only be compiled when zh_TW.po is outdated.
+
+#
+#  For more details, see:
+#  https://github.com/vim/vim/pull/3476
+#  https://github.com/vim/vim/pull/3476
```
k-takata: Duplicated.

src/po/big5corr.c
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+/*
+ * 06.11.23, added by Restorer:
+ * For more details, see:
+ * https://github.com/vim/vim/pull/3476
+ * https://github.com/vim/vim/pull/3476
```
k-takata: Duplicated.

src/po/big5corr.c
```diff
@@ -0,0 +1,61 @@
...
+	char *p;
+
+	while (fgets(buffer, BUFSIZ, stdin) != NULL)
+	{
+		for (p = buffer; *p != 0; p++)
+		{
+			if (strncmp(p, "charset=utf-8", 13) == 0
+				|| strncmp(p, "charset=UTF-8", 13) == 0)
+			{
+				fputs("charset=big5", stdout);
```
k-takata: You changed cp932 to CP932 in sjiscorr.c. How about this line, then?

src/po/Make_mvc.mak
```diff
@@ -6,26 +6,60 @@
...
-VIMRUNTIME = ..\..\runtime
-!endif
+# 27.10.23, added by Restorer
+!IF [powershell -nologo -noprofile "exit $$psversiontable.psversion.major"] == 2
+!ERROR The program "PowerShell" version 3.0 or higher is required to work
+!ENDIF
+
+# 27.10.23, added by Restorer
+!IFNDEF LANGUAGE
+!ERROR Set the environment variable %LANGUAGE%. See readme_mvc.txt
```
k-takata:
```suggestion
!ERROR Set the environment variable %LANGUAGE%. See README_mvc.txt
```

src/po/Make_mvc.mak
```diff
@@ -34,13 +68,443 @@ LSFLAGS = /b /on /l /s
 INSTALLDIR = $(VIMRUNTIME)\lang\$(LANGUAGE)\LC_MESSAGES
 
 .SUFFIXES:
-.SUFFIXES: .po .mo .pot
+.SUFFIXES: .po .mo .pot .ck
+
+all : $(MOFILES) $(MOCONVERTED)
+
+# 27.10.23, added by Restorer
```
k-takata: I don't think you need this kind of comment. (Other than you describe some additional information.)
You can see your name from the git log.

src/po/Make_mvc.mak
```diff
@@ -34,13 +68,443 @@ LSFLAGS = /b /on /l /s
...
+!IF EXIST ("$(GETTEXT_PATH)\msgconv.exe")
+	"$(GETTEXT_PATH)\msgconv.exe" -t CP932 $? | .\sjiscorr.exe > $@
+!ELSEIF DEFINED (ICONV)
+	$(ICONV) -f UTF-8 -t CP932 $? | .\sjiscorr.exe > $@
+!ELSE
+	powershell -nologo -noprofile -Command\
+	[System.IO.File]::WriteAllText(\"$@\",\
+	[System.IO.File]::ReadAllText(\"$?\",\
+	[System.Text.Encoding]::GetEncoding(65001)),\
+	[System.Text.Encoding]::GetEncoding(932))
```
k-takata: How about adding indents and spaces for readability?
```suggestion
	powershell -nologo -noprofile -Command \
		[System.IO.File]::WriteAllText(\"$@\", \
		[System.IO.File]::ReadAllText(\"$?\", \
		[System.Text.Encoding]::GetEncoding(65001)), \
		[System.Text.Encoding]::GetEncoding(932))
```
(Same for other PowerShell commands.)

src/po/Makefile
```diff
@@ -231,7 +262,10 @@ gvim.desktop: gvim.desktop.in $(POFILES) vim.desktop
 	if command -v desktop-file-validate; then desktop-file-validate tmp_gvim.desktop; fi
 	mv tmp_gvim.desktop gvim.desktop
 
-update-po: $(LANGUAGES)
+# 06.11.23, changed by Restorer
+# When updating ja.sjis.po there are a bunch of errors and a crash.
+# The files that are converted to a different encoding clearly state "DO NOT EDIT".
+update-po : $(MOFILES:.mo=)
```
k-takata: Changing `$(LANGUAGES)` to `$(MOFILES:.mo=)` doesn't look correct.
Could you explain this?

src/po/Make_mvc.mak
RestorerZ: Corrected. Thank you.

src/po/Make_mvc.mak
RestorerZ: Uh, excuse me. Habit. I was once taught that if you fix someone else's code, you must specify when and who fixed it. There were no version control systems back then... ;-)
Fixed it.

src/po/Make_mvc.mak
RestorerZ: Did it. Thank you.

src/po/Make_mvc.mak
RestorerZ: Corrected.

src/po/Make_mvc.mak
RestorerZ: > A target must be at the start of the line. It can't be indented with any spaces or tabs. Use a colon (:) to separate targets from dependents. Spaces or tabs are allowed between targets, the colon separator (:), and dependents. To split the dependency line, use a backslash (\) after a target or dependent.

[https://learn.microsoft.com/en-us/cpp/build/reference/description-blocks?view=msvc-170](https://learn.microsoft.com/en-us/cpp/build/reference/description-blocks?view=msvc-170)

And

> If the name of the last target before the colon (:) is a single character, you must
put a space between the name and the colon; otherwise, NMAKE interprets the
letter-colon combination as a drive specifier.

From the documentation LMAETC16.DOC Chapter 16 Managing Projects with NMAKE

Also, if you pay attention, all the examples in the documentation, both on Microsoft's site and in Chapter 16 Managing Projects with NMAKE, are written exactly like this. That is, with a space after the target and before the colon.

src/po/Makefile
RestorerZ: Fix it.

src/po/Makefile
RestorerZ: I'm sorry, but I don't understand what needs to be fixed.
Now in the file ko.UTF-8.po it says:

```
# Korean translation for Vim
#
# FIRST AUTHOR SungHyun Nam <goweol@gmail.com>, 2000-2018
#
# Original translations.
```

And in the ko.po file, after conversion:

```
# Generated from ko.UTF-8.po, DO NOT EDIT
#
# FIRST AUTHOR SungHyun Nam <goweol@gmail.com>, 2000-2018
#
# Original translations.
```

Which is clearly not correct.

src/po/Makefile
RestorerZ: The files that are converted to a different encoding clearly state "DO NOT EDIT".
As I understand it, the PO files to be created from the original translation PO files may not exist in the "po" directory at all and are created dynamically from the original translation files. We have to update the original translation files, edit and then convert the original translation file to another necessary encoding. Am I understanding this correctly?
Now the $(LANGUAGES) macro lists all PO files.
According to this rule, we take the files listed in the \\$(MOFILES) macro (i.e. the original translation PO files), trim the ".mo" extension and send them to the $(LANGUAGES): rule, where the PO files are updated.

Plus, when updating `ja.sjis.po` there are a bunch of errors and a crash.

Tested on "msgmerge" version 0.21 from [here](https://github.com/mlocati/gettext-iconv-windows) and version 0.22 from [here](https://github.com/vslavik/gettext-tools-windows). Why this happens, I haven't figured out yet.

src/po/big5corr.c
RestorerZ: Fixed.

src/po/big5corr.c
RestorerZ: Fixed.

src/po/big5corr.c
RestorerZ: Fixed.

src/po/big5corr.c
RestorerZ: Fixed it. Thank you.

src/po/Make_mvc.mak
chrisbra: yes agree

src/po/Make_mvc.mak
```diff
@@ -53,43 +502,68 @@ PO_INPUTLIST = \
...
-clean:
+clean : checkclean
 	$(RM) *.mo
 	$(RM) *.pot
 	$(RM) files
+	$(RM) sjiscorr.obj sjiscorr.exe
+#	$(RM) big5corr.obj big5corr.exe
+
+
+
```
chrisbra: can you please remove those additional empty lines?

src/po/Makefile
```diff
@@ -102,78 +102,128 @@ nb.po: no.po
 ja.sjis.po: ja.po
 	@$(MAKE) sjiscorr
 	rm -f ja.sjis.po
-	iconv -f utf-8 -t cp932 ja.po | ./sjiscorr > ja.sjis.po
+	iconv -f UTF-8 -t CP932 ja.po | ./sjiscorr > ja.sjis.po
```
chrisbra, line 105: is there a reason for upper-casing the names of the encodings?

src/po/Makefile
```diff
@@ -231,7 +282,10 @@ gvim.desktop: gvim.desktop.in $(POFILES) vim.desktop
 	if command -v desktop-file-validate; then desktop-file-validate tmp_gvim.desktop; fi
 	mv tmp_gvim.desktop gvim.desktop
 
-update-po: $(LANGUAGES)
+# 06.11.23, changed by Restorer
+# When updating ja.sjis.po there are a bunch of errors and a crash.
+# The files that are converted to a different encoding clearly state "DO NOT EDIT".
+update-po: $(MOFILES:.mo=)
```
chrisbra, line 287: Is that correct?

src/po/Makefile
```diff
@@ -186,6 +236,7 @@ prefixcheck:
 clean: checkclean
 	rm -f core core.* *.old.po *.mo *.pot sjiscorr
 	rm -f LINGUAS vim.desktop gvim.desktop tmp_*desktop
+#	rm -f big5corr
```
chrisbra, line 239: doesn't that leave big5corr file around?

src/po/Make_mvc.mak
RestorerZ: Removed.

src/po/Makefile
RestorerZ: Well, it will work both ways.
It's just more visual. And in the "Gettext" manual it is [recommended](https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html) to write in uppercase.

> The character encoding name can be written in either upper or lower case. Usually upper case is preferred.

src/po/Makefile
RestorerZ: The rule itself is correct. UNIX make as well as MSFT nmake support character substitution in macros.
@k-takata will suggest to explain why it is done this way, the reason, as I understand it, I have [indicated](https://github.com/vim/vim/pull/13518#discussion_r1390050900).

src/po/Makefile
RestorerZ: I left this line commented out, since the rule itself, where big5corr is created, is also commented out.
This rule is added in case there are backslash characters in the "zh_TW.po" file.

src/po/Makefile
k-takata: Ah, you are correct. I misunderstood that `ko.UTF-8.po` had a line `# Korean translation for Vim.`

src/po/Make_mvc.mak
k-takata: > If the name of the last target before the colon (:) is a single character, you must
put a space between the name and the colon; otherwise, NMAKE interprets the
letter-colon combination as a drive specifier.

Ah, I didn't know this pitfall. (I think this is a very rare case, though.)

src/po/Makefile
k-takata: Okay, understood.
`update-po` is used to update the original translation files based on the latest source code.
Converted po files are updated by other rules.

Then, the following part
```
# Don't add a dependency here, we only want to update the .po files manually
$(LANGUAGES):
```
should be also updated like this?
```
# Don't add a dependency here, we only want to update the .po files manually
$(MOFILES:.mo=):
```

src/po/Makefile
k-takata: (You mentioned the wrong person. I edited your comment.)

src/po/Make_mvc.mak
RestorerZ: Yes, I agree with you that the case is quite rare. And we are unlikely to encounter this situation.
But the Russians have a saying «иногда лучше перебдеть, чем недобдеть».
Which means that it is better to try to foresee all possible situations now than to deal with unpleasant consequences later.
Maybe someone wants to make a translation for C.

src/po/Makefile
RestorerZ: Yes, this is how I originally wanted to do it, but unfortunately it doesn't work quite right. For example, when the command to update a single language from the command line is executed.
`nmake -f Make_mvc.mak <LNG>`
And secondly, we still have the option to manually update PO files that are created during conversion. If we need it for some reason.

src/po/Make_mvc.mak
chrisbra: I'd prefer if we stay consistent here with other Makefiles and also avoid changing too many lines for no benefit

RestorerZ: Our benefit, in this case, is that we follow the official recommendations and try to minimize possible unpleasantness. nmake syntax is, in any case, slightly different from make syntax.

RestorerZ: unimportant for now. withdrawn whitespace

src/po/Make_mvc.mak
```diff
@@ -6,26 +6,55 @@
...
+!INCLUDE Make_all.mak
+
+!IFNDEF VIMRUNTIME
+VIMRUNTIME = ..\..\runtime
+!ENDIF
 
 PACKAGE = vim
-VIM = ..\vim
+# Correct the following line for the where executeable file vim is installed
+VIMEXE = D:\Programs\Vim\vim90\vim.exe
```
k-takata: Are there any reasons you changed the variable name and the value?

k-takata:
What are the files `big5_chr-bslh.txt` and `big5_corr-chr-bslh.txt` used for?

src/po/Make_mvc.mak
RestorerZ: Yes. The variable value is used here to run Vim and process scripts. The previous value of the variable gives an error:

```
F:\RepoGit\vimOrigin\src\po>E:\Project\vim\tools\nmake.exe -f Make_mvc.mak ru

Microsoft (R) Program Maintenance Utility Version 14.27.29111.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        ..\vim -u NONE --not-a-term -S tojavascript.vim vim.pot ../../runtime/optwin.vim  ../../runtime/defaults.vim
'..\vim' is not recognized as an internal or external command,
operable program or batch file.
NMAKE : fatal error U1077: '..\vim' : return code '0x1'
Stop.
NMAKE : fatal error U1077: 'E:\Project\vim\tools\nmake.exe' : return code '0x2'
Stop.
```

The name of the variable is chosen so as to convey more precisely what it is used for (i.e. it should contain the Vim executable). And to be uniformly similar to [this](https://github.com/vim/vim/blob/ce3b0136c6d9d09af41969d3dc9634f115505a32/runtime/doc/Makefile).

RestorerZ:
> What are the files `big5_chr-bslh.txt` and `big5_corr-chr-bslh.txt` used for?

Ah. Must have been left over from when I tested "big5corr" work. Yes, it is better to delete them, they are unnecessary. Thanks, I will.

chrisbra:
thanks, it's looking better now.

chrisbra:
can you please also add  src/po/big5corr.c  to the [Filelist](https://github.com/vim/vim/blob/master/Filelist)?

src/po/Make_mvc.mak
k-takata: If you build Vim from the source code, `vim.exe` should exist in the `src` directory. `..\vim` refers to it.

The variable `VIM` is used in other makefiles in the same directory. So, using `VIM` is consistent.

RestorerZ: Ah, yes, building from source. I more often work with this file directly without building the source and overlooked it. Of course that should be corrected. Thanks, I'll fix it now.

So let's rename this variable. And then we will have makefiles in this directory where the variable name represents its contents. We're assuming that Vim has been built but not installed. So we use the executable in the build directory. Similar to [this](https://github.com/vim/vim/blob/3770574e4a70e810add9929973c51f9070c8c851/src/Makefile#L2352).
What do you think about that, @k-takata ?

RestorerZ:
> thanks, it's looking better now.

It's probably just more familiar that way.

> can you please also add src/po/big5corr.c to the [Filelist](https://github.com/vim/vim/blob/master/Filelist)?

Added. Also added `runtime/lang/Make_all.mak` and `runtime/lang/Make_mvc.mak`.  Thanks for the prompt!

src/po/Make_mvc.mak
```diff
@@ -81,15 +532,27 @@ $(LANGUAGES):
...
 	if not exist $(INSTALLDIR) $(MKD) $(INSTALLDIR)
 	$(CP) $(LANGUAGE).mo $(INSTALLDIR)\$(PACKAGE).mo
 
 install-all: all
-	FOR %%l IN ($(LANGUAGES)) DO @IF NOT EXIST $(VIMRUNTIME)\lang\%%l\LC_MESSAGES $(MKD) $(VIMRUNTIME)\lang\%%l\LC_MESSAGES
-	FOR %%l IN ($(LANGUAGES)) DO @$(CP) %%l.mo $(VIMRUNTIME)\lang\%%l\LC_MESSAGES\$(PACKAGE).mo
+	for %%l in ($(LANGUAGES)) do @if not exist $(VIMRUNTIME)\lang\%%l\LC_MESSAGES \
+	$(MKD) $(VIMRUNTIME)\lang\%%l\LC_MESSAGES
+	for %%l in ($(LANGUAGES)) do @$(CP) %%l.mo \
+	$(VIMRUNTIME)\lang\%%l\LC_MESSAGES\$(PACKAGE).mo
```
k-takata:
```suggestion
	for %%l in ($(LANGUAGES)) do @if not exist $(VIMRUNTIME)\lang\%%l\LC_MESSAGES \
		$(MKD) $(VIMRUNTIME)\lang\%%l\LC_MESSAGES
	for %%l in ($(LANGUAGES)) do @$(CP) %%l.mo \
		$(VIMRUNTIME)\lang\%%l\LC_MESSAGES\$(PACKAGE).mo
```

RestorerZ: formatting

k-takata (approved):

src/po/Make_mvc.mak
```diff
@@ -6,26 +6,55 @@
...
-!ifndef VIMRUNTIME
-VIMRUNTIME = ..\..\runtime
-!endif
+!IF [powershell -nologo -noprofile "exit $$psversiontable.psversion.major"] == 2
+!ERROR The program "PowerShell" version 3.0 or higher is required to work
+!ENDIF
+
+!IFNDEF LANGUAGE
+!ERROR Set the environment variable %LANGUAGE%. See README_mvc.txt
+!ENDIF
```
k-takata, line 15: I'm rethinking whether this check is needed or not.
This check breaks my build script.

`$(LANGUAGE)` is used only in three make targets: `first_time`, `install`, `cleanup-po`.
Other make targets don't need this at all.

k-takata: See #13567

k-takata: vim-win32-installer is also broken: https://ci.appveyor.com/project/chrisbra/vim-win32-installer/build/job/a5rw0d0y1a4e9498

PR #13520, opened 2023-11-10
[vim9class] document `_` as protected instead of private

### LINKED ISSUE #13504: [vim9class] `_` description does not conform to common meaning
### Steps to reproduce

This issue comes from #13457, it's the only thing in that discussion that has general agreement.

Note: this issue is only about documentation; no code changes are implied/suggested. Read #13457 for the broader discussion.

The following table is summary, taken from https://en.wikipedia.org/wiki/Access_modifiers,

| Keyword | maximal access |
| -------- | ------ |
| private   | class |
| protected | derived classes |

shows two widely used access methods and how they are named.

Vim uses `_` for what most OOP languages call protected, OK; the problem is that it describes `_` as private. This causes confusion and is contrary to accepted nomenclature.

Additionally, if there is ever work to extend/update the access model, for example
- hiding variables from derived classes (commonly called private)
- accessible from within the file defining the class

then using "private" in the docs, when "protected" is the common meaning, likely complicates future changes.

@yegappan, @chrisbra If there's agreement to change the docs from private to protected, and no one wants to address the issue, then I can take a first cut at it and it can be wordsmithed and cleaned up in PR review.
(AFAIK, no response is not agreement)

### Expected behaviour

`_` is documented as "protected"

### Version of Vim

9.0.2092

### Environment

ubuntu

### Logs and stack traces

_No response_

errael:
Fixes #13504.

The implementation of this PR takes a minimalist approach. Mostly just change "private" to "protected". Keep an eye out for things that don't read naturally, even if not directly part of this change. There is some additional text and some |links| have been added. Any issues with them?

Some error messages are changed along with the tests that reference them.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -117,7 +117,7 @@ variable.
 
 
 Object variable write access ~
-
+						    *read-only-variable*
 Now try to change an object variable directly: >
 
 	pos.lnum = 9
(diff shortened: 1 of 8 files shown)


chrisbra:
thanks

Issue #12430, opened 2023-05-23
Incorrect display of the command shell prompt after exiting the Vim program


RestorerZ:
### Steps to reproduce

1. Open the standard Windows Command shell --- cmd.exe
2. For illustrative purposes, fill the screen with some text, for example, enter the command `help` in the command shell prompt.
3. Run the `Vim ‐u NONE --noplugin`
4. After that, immediately exit the Vim program with the command `:q`
5. The command shell prompt will be displayed at the top of the command shell window over the text already there.

![not_correct_promt-cmd12c02b9ec6d29c5d6.png](https://imgbb.su/images/2023/05/23/not_correct_promt-cmd12c02b9ec6d29c5d6.png)

This is clearly visible in the attached screenshot.
The starting position of the invitation before launching Vim is marked with a blue stripe.
The position of the prompt after exiting Vim is marked with a red stripe.

This behavior of Vim has been observed since version [9.0.1065](https://github.com/vim/vim-win32-installer/releases/tag/v9.0.1065).
Versions [9.0.1060](https://github.com/vim/vim-win32-installer/releases/tag/v9.0.1060) and earlier worked correctly.

I think something is broken either in the patch [9.0.1063](https://github.com/vim/vim/releases/tag/v9.0.1063) or in the patch [9.0.1065](https://github.com/vim/vim/releases/tag/v9.0.1065).

### Expected behaviour

The command prompt is displayed at the bottom of the screen and does not overlap the existing text.
As in this screenshot.

![correct_promt-cmd1fd287fa150c79e25.png](https://imgbb.su/images/2023/05/23/correct_promt-cmd1fd287fa150c79e25.png)

### Version of Vim

9.0.1572

### Environment

WindowsNT 6.2

brammool:
> ### Steps to reproduce
> 
> 1. Open the standard Windows Command shell --- cmd.exe
> 2. For illustrative purposes, fill the screen with some text, for example, enter the command `help` in the command shell prompt.
> 3. Run the `Vim ‐u NONE --noplugin`
> 4. After that, immediately exit the Vim program with the command `:q`
> 5. The command shell prompt will be displayed at the top of the command shell window over the text already there.
> 
> ![not_correct_promt-cmd12c02b9ec6d29c5d6.png](https://imgbb.su/images/2023/05/23/not_correct_promt-cmd12c02b9ec6d29c5d6.png)
> 
> This is clearly visible in the attached screenshot.
[...]

PR #13521, opened 2023-11-11
Fix bug of silent mode clearing terminal screen on Windows

### LINKED ISSUE #12822: "vim -e -s" doesn't exit properly
### Steps to reproduce

1. run Vim with the `-e -s` arguments, and exit it. (for example: `vim.exe -e -s -c q`)
2. The screen won't show the previous history, and you will not be able to scroll upwards, as if Vim was still running.

Notice the greyed out scrollbar on this screenshot.
image

### Expected behaviour

Vim exits, and your previous command history is shown again and you can scroll up too. This is what happens if you run Vim with just the `-e` argument, without `-s` (for example: `vim.exe -e -c q`)

image

### Version of Vim

9.0.1712, also 9.0.1636.

### Environment

OS: Windows 11 22h2
Terminal and shell: doesn't make a difference

### Logs and stack traces

_No response_

nir9:
Problem: Terminal control codes (https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) are sent even when silent mode is on, causing the terminal to clear up

Solution: Block any terminal codes when silent mode is on

closes #12822

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -8578,6 +8578,9 @@ vtp_printf(
     DWORD   result;
     int	    len;
 
+    if (silent_mode)
+	return 0;
+
     va_start(list, format);
     len = vim_vsnprintf((char *)buf, 100, (char *)format, list);
     va_end(list);


chrisbra:
thanks for fixing this!

RestorerZ:
Thank you so much!
With this fix, this problem was also resolved.
#12430

nir9:
@chrisbra notice that you referenced this pull request from another unrelated commit 4ae16d7 (probably by mistake) - so there might be a pull request that is still open that can be closed

chrisbra:
yeah, I noticed after pushing, that it referenced the wrong PR to close. So I manually closed it #13530

PR #13523, opened 2023-11-11
textobjects: Make selecting inner empty blocks an error

### LINKED ISSUE #13514: Inconsistent behavior for visual selection of inner 'block' text-objects when block is empty i( i[ i{
### Steps to reproduce

1. run vim and paste the following examples into a buffer
2. position the cursor as indicated by the highlight (e.g. the cursor is here ->`.`)
3. type `vib` to extend the selection to the inner part of the parenthesis
4. check the hightlighted range and compare it to what it's expected to be.

### Expected behaviour

**Problem**
In the context of multiple parentheses groups chain (e.g. '`(text)()`'), when visually selecting the inner part of _empty_ parentheses using `vib` with the cursor positionned on the opening parenthesis, the block selected is actually the previous one.

Moreover, if the preceding group is also empty, the pair of parentheses of the _previous_ group is selected despite it being an 'inner' selection (see Example3)

In these contexts, I would expect a fail of the text-object, hence leading to only the character at the current position being visually selected.

In the examples below, `highlighted range` represent 
- the cursor position in the starting position (nothing selected)
- visually selected range in the resulting/expected position

______________________________________________________________________
**Example1**
start:
> (text)`(`)

command:
>vib

result:
> (`text`)()

expectation:
>(text)`(`)
________________________________________________________________________
**Example2**
start:
> (text)`(`
> )

command:
>vib

result:
> (`text`)(
> )

expectation:
> (text)`(`
> )
________________________________________________________________________
**Example3**
start:
> ()`(`)

command:
>vib

result:
> `()`()

expectation:
> ()`(`)
________________________________________________________________________
**Final note**
Same behavior applies to other 'block' text-objects such as `i[` and `i{`

### Version of Vim

:version VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Mar  8 2023 17:49:58) Included patches: 1-1393 

### Environment

Operating system : Linux 3.10.0-1160.53.1.el7.x86_64 (scientific linux 7)
Shell: tcsh 6.18.01
Term: xterm version XTerm(327)
$TERM: xterm-256color

### Logs and stack traces

_No response_

chrisbra:
fixes: #13514

### DIFF
--- a/src/testdir/test_textobjects.vim
+++ b/src/testdir/test_textobjects.vim
@@ -400,7 +400,7 @@ func Test_paragraph()
   call assert_beeps("normal Vipip")
   exe "normal \<C-C>"
 
-  close!
+  bw!
 endfunc
 
 " Tests for text object aw
@@ -606,7 +606,7 @@ func Test_textobj_quote()
   normal $hhyi"
   call assert_equal('bar', @")
 
-  close!
+  bw!
 endfunc
 
 " Test for i(, i<, etc. when cursor is in front of a block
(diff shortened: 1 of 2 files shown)


chrisbra:
this might potentially be controversial, so will mark as backlog however

fritzophrenic:
Do I understand correctly this is only if the cursor would not move when evaluating the text object? And only for "inner" objects where there would be nothing to act on if doing "di" on the object, for example? Meaning that it mostly has an impact for paired items like parentheses or quotes with no text between them? Mostly that is probably OK, but I want to be sure we can still do things like "cit" inside an XML open tag, to jump to insert mode inside the tag. Useful for example if you have <tag attr1='val' attr2='val' attr3='val'></tag> with the cursor on "attr1" and you want to quickly insert some text inside the tag.

chrisbra:
Yeah, it should happen, for *inner* objects, when the initial found start and end positions are the same (which basically means there is nothing to select and therefore returning an error makes sense I think.

habamax:
> this might potentially be controversial, so will mark as backlog however

It is, indeed.

While selecting empty inner blocks might be an error, we need to keep previous behavior for `change` commands:

```
# pre-patch, cursor is on column 0, cib -- place cursor inside () and you in insert mode
# post-patch, cursor is on column 0, cib -- place cursor on (
def SomeFunc()
```

vim-ml:
***@***.***

Christian Brabandt ***@***.***> 于2024年1月5日周五 05:13写道：

PR #13524, opened 2023-11-11
Clear matchparen highlighting in completion mode

### LINKED ISSUE #13493: Matching parenthesis highlights are not updated when completion popup is visible
### Steps to reproduce

1. Create an 'init.vim' file with the following content:
```vim
set nocompatible
```
2. Open Vim with that init file and 'matchparen' plugin loaded. For example, for local build I used `VIMRUNTIME=./runtime src/vim --clean -u init-matchparen.vim`.
3. Enter Insert mode and add lines `aa`, `aaa`, and `(a)`. Exit Insert mode.
4. Put cursor on the closing bracket in last line. This should highlight matching brackets.
5. Enter Insert mode with `i` and press `<C-n>`. This should open completion popup and select first `aa` entry. The highlight of the right bracket did not change and now over the second `a`. Pressing `<C-n>` again changes inserted text to `aaa` but highlight is still at the same place.

### Expected behaviour

Matched parenthesis are always highlighted even if completion popup is visible.

------

I've bisected the history, and it was the #13393 that introduced the issue. My guess would be because `s:Highlight_MatchingPair()` is called on `CursorMovedI` and `TextChangedI`, both not triggered if popup menu is visible. Adding `TextChangedP` [here](https://github.com/vim/vim/blob/2a94e9879283c55b162cf4e6d9ac7e0b0c35bc97/runtime/plugin/matchparen.vim#L28) fixes the issue, but I am not sure if this is a good solution here. Also not sure why it had worked before, but started to show only after #13393.

### Version of Vim

9.0.2092

### Environment

Operating system: EndeavourOS Linux x86_64 (6.5.9-arch2-1)
Terminal: st-0.9
Value of $TERM: st-256color
Shell: zsh 5.9

### Logs and stack traces

_No response_

chrisbra:
closes: #13493


echasnovski:
A nitpick note: there is a typo in comment message ("Cleap" -> "Clear").

PR #12589, opened 2023-06-25
termdebug: fix sign text is longer than 2 characters (E239)

skywind3000:
In termdebug, when breakpoint id is greater than 99 the `CreateBreakpoint` function will fail because
the `text` element in `sign_defined()` function only accept strings no more than 2 characters.

solution:

1) always display the last two characters of the breakpoint id, in other words, 133 will display as `33`.
2) if `g:termdebug_config['sign']` is defined by user, display it in the signcolumn instead of the the breakpoint id.

closes https://github.com/vim/vim/issues/12588

PR #13525, opened 2023-11-11
fix: a bit nicer termdebug breakpoint sign label

Shane-XB-Qian:
1. this make the breakpoint sign label can hold 255 bp id (>100).
2. this will not re-use existed bp id if over 255 (less confusion).

// related #12589
// that should be the last chat (I) with Bram, r.i.p

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1588,9 +1588,9 @@ If there is no g:termdebug_config you can use: >
 
 Change default signs ~
 							*termdebug_signs*
-Termdebug uses the last two characters of the breakpoint ID in the
-signcolumn to represent breakpoints. For example, breakpoint ID 133
-will be displayed as `33`.
+Termdebug uses the hex number of the breakpoint ID in the signcolumn to
+represent breakpoints. if it is greater than "0xFF", then it will be displayed
+as "F+", due to we really only have two screen cells for the sign.
 
 If you want to customize the breakpoint signs: >
 	let g:termdebug_config['sign'] = '>>'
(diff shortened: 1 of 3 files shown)


runtime/doc/terminal.txt
```diff
@@ -1588,9 +1588,9 @@ If there is no g:termdebug_config you can use: >
 
 Change default signs ~
 							*termdebug_signs*
-Termdebug uses the last two characters of the breakpoint ID in the
-signcolumn to represent breakpoints. For example, breakpoint ID 133
-will be displayed as `33`.
+Termdebug uses the hex number of the breakpoint ID in the signcolumn to
+represent breakpoints. if it is greater than "0xFF", then it will be displayed
+as "^F", due to we really only have two screen cells for the sign.
```
chrisbra: Why using "^F"? that seems a bit arbitrary. How about using `!!` instead? And perhaps even make this configurable?

Shane-XB-Qian:
"^F" means up/over than "FF"
i don't want to make termdebug too complex but: though simple but clear and correct.

--
shane.xb.qian

chrisbra:
I see, thanks

zeertzjq:
Maybe it can be `F+` (like how over 99 search matches shows `99+` in search stats) instead? `^F` can already mean a `\x06` byte.

Shane-XB-Qian:
> `^F` can already mean a `\x06` byte.

here '^' is a pure char '^', but ok, let me change it to 'F^' (means 'FF' is up/over), maybe better.

chrisbra:
let's make it `F+` please

Shane-XB-Qian:
> let's make it `F+` please

ok, fine, as long as it would not re-use existed bp id.

PR #13534, opened 2023-11-14
runtime(tutor): Use more general terms

pohaosu:
Since files may not be stored on disks, we should use more general terms to describe the devices used to store files.

### DIFF
--- a/runtime/tutor/tutor
+++ b/runtime/tutor/tutor
@@ -700,13 +700,14 @@ NOTE:  You can also read the output of an external command.  For example,
 	  :!dir		   :!ls		   -  shows a directory listing.
 	  :!del FILENAME   :!rm FILENAME   -  removes file FILENAME.
 
-  2.  :w FILENAME  writes the current Vim file to disk with name FILENAME.
+  2.  :w FILENAME  writes the current Vim file to a storage device with name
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, but isn't disk nowadays a sloppy term for all kind of persistent storage?

pohaosu:
Yes @chrisbra, but I'm thinking that it is possible that people may use vimtutor inside memory-backed mounts, e.g. tmpfs. In other words, if users' current directory is mounted by tmpfs, the file they created while using vimtutor will be a non-persistent file. Therefore, instead of calling it disk files, I would suggest just calling it storage devices.

chrisbra:
okay, but then the file won't be persisted anyhow as soon as the memory-mounts are gone

pohaosu:
Right, but while users are using vimtutor, the temporary file is still there and vimtutor still calls the file a disk file. This issue probably is more against this kind of scenario. If we don't need to bother with this case, I will close it.

nickspoons:
Personally I find the original text simple and clear. Introducing a "storage device" concept does not make the message clearer.

chrisbra:
let's just close it

pohaosu:
thanks, guys.

PR #13535, opened 2023-11-14
Add a couple of minor and security relevant fixes

chrisbra:
There were several low security relevant issues found using fuzzying. Mostly some variables were overflowing. So fix them here altogether. I don't think this is overly critical, that's why I am creating a public PR here so that CI can hopefully run successfully.

Details are in each commit message.


src/normal.c
```diff
@@ -2562,7 +2562,14 @@ nv_z_get_count(cmdarg_T *cap, int *nchar_arg)
...
 	    n /= 10;
 	else if (VIM_ISDIGIT(nchar))
+	{
+	    if (n > LONG_MAX/10)
+	    {
+		clearopbeep(cap->oap);
+		break;
+	    }
 	    n = n * 10 + (nchar - '0');
+	}
```
errael: It's probably been over 30 years since I was checking for "C" code overflow but this looks suspicious. I don't know what the above code is doing or what values might actually be encountered, but the precondition check 
```
n > LONG_MAX/10
``` 
looks wrong. Notice that there is overflow after the first check in the following example.

```
#include <stdio.h>
#include <limits.h>

#define SAFE_SPACE 100

int main(int ac, char *av[]) {
    int nchar = '9' - '0';  // Not sure what legit values nchar can take.

    long n = LONG_MAX / 10;
    if (n > LONG_MAX / 10)
        printf("BREAK\n");
    else
        printf("%ld\n", n * 10 + ('9' - '0'));

    n = LONG_MAX / 10;
    if (n > LONG_MAX / 10 - SAFE_SPACE)
        printf("BREAK\n");
    else
        printf("%ld\n", n * 10 + ('9' - '0'));
}
```
```
-9223372036854775807
BREAK
```

src/ops.c
```diff
@@ -260,7 +260,12 @@ shift_line(
 		count = 0;
 	}
 	else
-	    count += sw_val * amount;
+	{
+	    if ((long)sw_val * (long)amount > INT_MAX - count)
```
k-takata: `(long)sw_val * (long)amount` may overflow when `long` is 32 bits.

src/misc1.c
```diff
@@ -975,6 +975,8 @@ get_number(
 	c = safe_vgetc();
 	if (VIM_ISDIGIT(c))
 	{
+	    if ( n > INT_MAX / 10)
```
k-takata:
```suggestion
	    if (n > INT_MAX / 10)
```

src/ops.c
chrisbra: I guess that's the reason why the test fails on Windows and on i386?

src/normal.c
chrisbra: true. But I think this will break already in the previous iteration of the loop

chrisbra: okay, should be fixed now

src/ops.c
chrisbra: should be fixed now

src/misc1.c
chrisbra: should be fixed now

src/normal.c
```diff
@@ -2562,7 +2562,14 @@ nv_z_get_count(cmdarg_T *cap, int *nchar_arg)
...
 	    n /= 10;
 	else if (VIM_ISDIGIT(nchar))
+	{
+	    if (n > LONG_MAX / 10 - (nchar - '0'))
+	    {
+		clearopbeep(cap->oap);
+		break;
+	    }
 	    n = n * 10 + (nchar - '0');
+	}
```
errael: I think you should go back to the original: `if (n > LONG_MAX/10)`,
otherwise "9...9" is considered overflow one loop too early.

```
Precondition:   n * 10 + 9  <= max
    ==>         n <= (max - 9) / 10 
Invert!:        n > (max - 9) / 10
```
With some handwaving, I'd say you can drop the "- 9" because of the "/ 10".
So you end up with the original
```
    ==>         n > max / 10
```
The test I'm running gives the same result with both
`n > (max - 9) / 10` and `n > max / 10`.

BTW, don't want a variable in the comparison, only constants,
so it's simply a compare against a constant.

With "n > max / 10 - 9" it's wrong when the value is like '999...'

(I wrote a simple program, nested loops to try all kinds of values,
to discover and bolster an intuitive sense to reveal why/how I screwed up.
(And I'm still working on the intuitive sense))

In the following the number in parens is the last "valid" number befor aborting
due to overflow. Notice 
```
=== n > max / 10 - 9
 ...
 6 - OVERFLOW 1 - 688 (688) 8
 7 - OVERFLOW 1 - 788 (788) 8
 8 - OVERFLOW 1 - 888 (888) 8
 9 - OVERFLOW 1 - 98 (98) 8
 ...
 6 - OVERFLOW 1 - 699 (699) 9
 7 - OVERFLOW 1 - 799 (799) 9
 8 - OVERFLOW 1 - 899 (899) 9
 9 - OVERFLOW 1 - 99 (99) 9

=== n > max / 10 
 ...
 6 - OVERFLOW 1 - 688 (688) 8
 7 - OVERFLOW 1 - 788 (788) 8
 8 - OVERFLOW 1 - 888 (888) 8
 9 - OVERFLOW 1 - 988 (988) 8
 ...
 6 - OVERFLOW 1 - 699 (699) 9
 7 - OVERFLOW 1 - 799 (799) 9
 8 - OVERFLOW 1 - 899 (899) 9
 9 - OVERFLOW 1 - 999 (999) 9
```

chrisbra: okay thanks done

src/ops.c
```diff
@@ -229,11 +229,11 @@ shift_line(
     int	amount,
     int call_changed_bytes)	// call changed_bytes()
 {
-    int		count;
+    long long	count;
     int		i, j;
     int		sw_val = (int)get_sw_value_indent(curbuf);
 
-    count = get_indent();	// get current indent
+    count = (long)get_indent();	// get current indent
```
k-takata:
```suggestion
    count = (long long)get_indent();	// get current indent
```
Converting to a larger type will be done automatically, so the cast may not be needed. However, if you want to cast it explicitly, I think this should be `(long long)`.

chrisbra: ah yes, I missed it when switching to (long long). Will fix

PR #13539, opened 2023-11-17
Better overflow detection

chrisbra:


src/misc1.c
```diff
@@ -2817,3 +2816,24 @@ may_trigger_modechanged(void)
     restore_v_event(v_event, &save_v_event);
 #endif
 }
+
+// For overflow detection, add an digit safely to an int value
```
errael: :s/an digit/a digit/

errael: :s/$/./

chrisbra: thanks, adjusted

Issue #1247, opened 2016-11-11
macros/mappings and 'keymap' or :lmap when iminsert=1 and/or imsearch=1


hardenedapple:
It seems that macros ignore the `'keymap'` setting or `:lmap` mappings, I think it would be preferable to do the same as for `:imap` mappings.

Test case:
Start vim with `vim -N -u NONE`
```
:set iminsert=1
:lmap a l<CR>
qaiaaa<CR><Esc>q@a
```

You see two lines in the file, the first one is `lll` and the second one is `aaa`.
Similar happens for search patterns when `'imsearch'` is either -1 or 1, and also for f/F t/T.

If you do the same with an `:imap` you get two lines of `lll`, I believe it would be less surprising if `:lmap` and `'keymap'` follow this behaviour.

The same happens for example with `:nmap` commands that insert some text.

Bricktech2000:
Bumping this. I use mappings like:

```vimscript
lnoremap <c-space> <c-k>NS
lnoremap <c-.> <c-k>,.
lnoremap <c-`> <c-k>'6
lnoremap <c-'> <c-k>'9
```

`lmap` is nice because the mappings will work in all the right places: insert mode, search, `f`, `t`, `r`... So it's a shame they break in macros.

PR #13540, opened 2023-11-17
Syntax file for SWIG (Simplified Wrapper Interface Generator) description files

mcepl:
The default syntax for .i files highlights comments in a reverse color scheme which doesn't look well.  This syntax builds on vim's c++ syntax by adding highlighting for common swig directives and user defined directives.  For an alternative syntax, see [vimscript #1247](https://www.vim.org/scripts/script.php?script_id=1247) (which I found after writing this).

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2183,6 +2183,9 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 " Swift Intermediate Language or SILE
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
+" Swig
+autocmd BufNewFile,BufRead *.i,*.swg,*.swig set filetype=swig
+
 " Sysctl
 au BufNewFile,BufRead */etc/sysctl.conf,*/etc/sysctl.d/*.conf	setf sysctl
 
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1688,7 +1688,7 @@ au BufNewFile,BufRead .procmail,.procmailrc	setf procmail
 " Progress or CWEB
 au BufNewFile,BufRead *.w			call dist#ft#FTprogress_cweb()
 
-" Progress or assembly
+" Progress or assembly or Swig
 au BufNewFile,BufRead *.i			call dist#ft#FTprogress_asm()
 
 " Progress or Pascal
@@ -2184,7 +2184,7 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
 " Swig
-autocmd BufNewFile,BufRead *.i,*.swg,*.swig set filetype=swig
+au BufNewFile,BufRead *.swg,*.swig set filetype=swig
 
 " Sysctl
 au BufNewFile,BufRead */etc/sysctl.conf,*/etc/sysctl.d/*.conf	setf sysctl
--- a/runtime/syntax/swig.vim
+++ b/runtime/syntax/swig.vim
@@ -1,0 +1,29 @@
+" Vim syntax file
+" Language:	SWIG
+" Maintainer:	Roman Stanchak (rstanchak@yahoo.com)
+" Last Change:	2023 April 27
+
+if exists("b:current_syntax")
+	finish
+endif
+
+" Read the C++ syntax to start with
+runtime! syntax/cpp.vim
+unlet b:current_syntax
+
+" SWIG extentions
+syn keyword swigDirective %typemap %define %apply %fragment %include %enddef %extend %newobject %name
+syn keyword swigDirective %rename %ignore %keyword %typemap %define %apply %fragment %include
+syn keyword swigDirective %enddef %extend %newobject %name %rename %ignore %template %module %constant
+syn match swigDirective "%\({\|}\)"
+syn match swigUserDef "%[-_a-zA-Z0-9]\+"
+
+" Default highlighting
+command -nargs=+ HiLink hi def link <args>
+HiLink swigDirective	Exception
+HiLink swigUserDef 		PreProc
+delcommand HiLink
+
+let b:current_syntax = "swig"
+
+" vim: ts=8
--- a/runtime/syntax/swig.vim
+++ b/runtime/syntax/swig.vim
@@ -1,10 +1,10 @@
 " Vim syntax file
 " Language:	SWIG
 " Maintainer:	Roman Stanchak (rstanchak@yahoo.com)
-" Last Change:	2023 April 27
+" Last Change:	2023 November 23
 
 if exists("b:current_syntax")
-	finish
+  finish
 endif
 
 " Read the C++ syntax to start with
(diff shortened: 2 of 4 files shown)


runtime/filetype.vim
```diff
@@ -2183,6 +2183,9 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 " Swift Intermediate Language or SILE
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
+" Swig
+autocmd BufNewFile,BufRead *.i,*.swg,*.swig set filetype=swig
```
dkearns: The `*.i` pattern is already in use by other filetypes so you'll need to modify this function to heuristically detect Swig files by content. https://github.com/vim/vim/blob/cc6be3e7557cba790c78812d8789f30596e09ecc/runtime/autoload/dist/ft.vim#L593

Tests also need to be added to https://github.com/vim/vim/blob/master/src/testdir/test_filetype.vim

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,29 @@
+" Vim syntax file
+" Language:	SWIG
+" Maintainer:	Roman Stanchak (rstanchak@yahoo.com)
```
dkearns, line 3: If you're not the author are you willing to maintain this file?

If so, can you add yourself to https://github.com/vim/vim/blob/master/.github/CODEOWNERS and also change the "Maintainer:" header line.   Some sort of attribution to the original author should also be included, this is usually done with a separate "Original Author:"  header field.

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,29 @@
...
+if exists("b:current_syntax")
+	finish
+endif
+
+" Read the C++ syntax to start with
+runtime! syntax/cpp.vim
+unlet b:current_syntax
+
+" SWIG extentions
+syn keyword swigDirective %typemap %define %apply %fragment %include %enddef %extend %newobject %name
```
dkearns: Unless you have `%` included in the `iskeyword` or `syn iskeyword` option values these `syn keyword` definitions won't match.

You may not have noticed as they're being matched by the later `swigUserDef` syntax group.

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,29 @@
...
+syn keyword swigDirective %rename %ignore %keyword %typemap %define %apply %fragment %include
+syn keyword swigDirective %enddef %extend %newobject %name %rename %ignore %template %module %constant
+syn match swigDirective "%\({\|}\)"
+syn match swigUserDef "%[-_a-zA-Z0-9]\+"
+
+" Default highlighting
+command -nargs=+ HiLink hi def link <args>
```
dkearns: This block can be reduced to

```vim
hi def link swigDirective Exception
hi def link swigUserDef PreProc
```

The HiLink command was used in the past for compatibility with Vim 5.x but we no longer do that in distributed syntax files.

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,29 @@
...
+HiLink swigDirective	Exception
```
dkearns: The highlight groups should, ideally, semantically match the syntax group they're highlighting.  `Exception` doesn't seem like the best choice here.

dkearns (changes requested):
Thanks for the PR.

Maybe you could consider adding a basic filetype plugin as well?

chrisbra:
@mcepl are you planning to address those comments?

mcepl:
> @mcepl are you planning to address those comments?

yes, I am. I am sorry, my reaction time is slow.

chrisbra:
thanks, no hurries, take the time you need and just ping us once those are addressed.

runtime/filetype.vim
jmarrec: I suggest using

```
# These include the leading '%' sign
var ft_swig_keywords = '^\s*%\%(endoffile\|fragment\|ignore\|typemap\|enddef\|warn\|parms\|pragma\|constant\|define\|module\|newobject\|include\|feature\|namewarn\|keyword\|extend\|native\|importfile\|varargs\|import\|echo\|includefile\|clear\|template\|types\|rename\|apply\|name\|beginfile\|typedef\|addmethods\|insert\|inline\)'
# This is the start/end of a block that is copied literally to the processor file (C/C++)
var ft_swig_verbatim_block_start = '^\s*%{'
var ft_swig_verbatim_block_end = '^\s*}%'
```

jmarrec: @mcepl I've tried to address some of these comments after seeing your message on the SWIG-devel mailing list. Made a PR pointing to your branch at https://github.com/openSUSE-Python/vim/pull/1

jmarrec:
I managed to run the test I wrote locally, clearly I botched something, seems like it's a `"` that should be a `'`

src/testdir/test_filetype.vim
```diff
@@ -2340,4 +2341,30 @@ func Test_vba_file()
...
+  " Swig
+
+  call writefile(['%module mymodule'], 'Xfile.i', 'D')
+  split Xfile.i
+  call assert_equal('swig', &filetype)
+  bwipe!
+
+  " Swig
+
+  call writefile(['%{', '#include <header.hpp>", '%}'], 'Xfile.i', 'D')
```
jmarrec, line 2356:
```suggestion
  call writefile(['%{', '#include <header.hpp>', '%}'], 'Xfile.i', 'D')
```

jmarrec: That still doesn't suffice though

```
1 FAILED:
Found errors in Test_swig_file():
command line..script /home/julien/Software/Others/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_swig_file line 7: Expected 'swig' but got 'progress'
command line..script /home/julien/Software/Others/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_swig_file line 20: Expected 'asm' but got 'progress'
```

mcepl:
I am adding commits kindly provided by @jmarrec (THANK YOU!). Could anybody look at this again and review the result, please? I think all reviewer's comments were addressed, but I am not 100% certain.

runtime/syntax/swig.vim
jmarrec: @mcepl This isn't addressed, but I don't know where to change this one

src/testdir/test_filetype.vim
jmarrec: Ok, I fixed it locally. @mcepl Are you planning on wrapping this PR? Otherwise I should open another PR, because I don't have rights to write to your branch and this is going to get annoying.

runtime/syntax/swig.vim
mcepl: I would be willing to maintain it just for the good ol’ times sake, but I am not using vim (or neovim) that much anymore (where I can I mostly switched to [vis](https://git.sr.ht/~martanne/vis)). @jmarrec would be much better choice, if he is willing.

Originally the code came from https://www.vim.org/scripts/script.php?script_id=1615 by (I guess) @rstanchak with some modifications from https://www.vim.org/scripts/script.php?script_id=1247 by [Beau Kissinger](mailto:mrkissinger@gmail.cn).

src/testdir/test_filetype.vim
mcepl: See [my other comment](https://github.com/vim/vim/pull/13540#discussion_r1403268413) … if you are willing to take over, then you are more than welcome. Just please leave there at least one commit from me. If you make new PR, let me know, and I will close this one.

mcepl:
Closing as superseded by #13562.

PR #13547, opened 2023-11-19
Coverity fixes

chrisbra:
For some reasons, Coverity detected 2 more issues in quickfix.c and gui_gtk_x11.c even so there were no code changes there recently. Still trying to fix those problems.

@yegappan can you please check if the change to qf_free_items() looks good? Thanks

### DIFF
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -4000,8 +4000,9 @@ qf_free_items(qf_list_T *qfl)
 		// to avoid crashing when it's wrong.
 		// TODO: Avoid qf_count being incorrect.
 		qfl->qf_count = 1;
+	    else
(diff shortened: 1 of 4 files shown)


src/quickfix.c
```diff
@@ -4001,7 +4001,9 @@ qf_free_items(qf_list_T *qfl)
 		// TODO: Avoid qf_count being incorrect.
 		qfl->qf_count = 1;
 	}
-	qfl->qf_start = qfpnext;
+	// avoid using feed memory
+	if (!stop)
```
yegappan: As the above `if` statement already uses the `!stop` condition, can this be moved into the above if statement?

chrisbra: sure, changed it

PR #13550, opened 2023-11-20
runtime(vim): Improve keymap file highlighting

dkearns:
Match :loadkeymap to EOF as a region and contain only allowed items.
    
Add highlighting for <Char- notation.


chrisbra:
thanks

chrisbra:
is it too much, if I ask for a screen-dump test?

PR #13551, opened 2023-11-21
Filelist: Add new sub-Makefiles

zdohnal:
Tags for help files disappeared with the latest Vim update in Fedora, which is caused by silent error (it didn't stop the build) about missing file. I use 'make unixall' in Fedora to get the latest patchlevels and the new files were missing from Filelist file which is used for generating the tarball.

The error from the build:
```
cd ../runtime/doc; if test -z "" -a -f tags; then \
	mv -f tags tags.dist; fi
generating help tags
make[1]: Entering directory '/builddir/build/BUILD/vim90/runtime/doc'
Makefile:17: Make_all.mak: No such file or directory
make[1]: *** No rule to make target 'Make_all.mak'.  Stop.
make[1]: Leaving directory '/builddir/build/BUILD/vim90/runtime/doc'
make: [Makefile:2350: installrtbase] Error 2 (ignored)
cd ../runtime/doc; \
	files=`ls *.txt tags`; \
	files="$files `ls *.??x tags-?? 2>/dev/null || true`"; \
	cp $files  /builddir/build/BUILDROOT/vim-9.0.2105-1.fc38.ppc64le/usr/share/vim/vim90/doc; \
	cd /builddir/build/BUILDROOT/vim-9.0.2105-1.fc38.ppc64le/usr/share/vim/vim90/doc; \
	chmod 644 $files
ls: cannot access 'tags': No such file or directory
```

[Fedora bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2250722)

### DIFF
--- a/Filelist
+++ b/Filelist
@@ -698,6 +698,7 @@ RT_ALL =	\
 		runtime/doc/*.pl \
 		runtime/doc/*.txt \
 		runtime/doc/Makefile \
+		runtime/doc/Make_all.mak \
 		runtime/doc/doctags.c \
 		runtime/doc/doctags.vim \
 		runtime/doc/test_urls.vim \
@@ -846,6 +847,7 @@ RT_AMI_DOS =	\
 # DOS runtime (also in the extra archive)
 RT_DOS =	\
 		README_dos.txt \
+		runtime/doc/Make_mvc.mak \
 		vimtutor.bat \
 
 # DOS runtime without CR-LF translation (also in the extra archive)


chrisbra:
how do you generate the filelist? Manually?

zdohnal:
@chrisbra the update script clones the Vim Github repository, runs 'make unixall' and rename the tarball to f.e. vim-9.0.2105.tar.gz (major.minor.patchlevel) - Filelist is taken from the project Github repository, I don't generate the Filelist in any way.

zdohnal:
In case there is an error, I manually check if there is a new file in Github repository, which is needed for compilation/tests and if it is present in Filelist.

psimonyi:
Filelist categorizes files, setting variables for the Makefile, so I don't see how it could be updated automatically.  But perhaps there should be a check during build or CI to make sure all files are listed?  It would require adding an explicit ignore-list, as there are a few files like SECURITY.md that are tracked in git but not used in the build.

Here is a quick script to list git-tracked files that aren't referenced in Filelist.  If you're interested I could clean it up a bit and make a separate patch:
<details><summary>unlisted.sh</summary>

```bash
#!/usr/bin/bash
# List all files that are tracked in git but not listed in Filelist.
# Exit code is 5 when any such files exist.

readarray -t patterns < <(sed -re 's/^\t+(.*) \\/\1/; t; d' Filelist)
readarray -t -d '' files < <(git ls-files -z)
any_matches=0

for f in "${files[@]}"; do
    matched=0
    for p in "${patterns[@]}"; do
        # We want to match glob patterns from the Makefile:
        # shellcheck disable=SC2053
        if [[ "$f" == $p ]]; then
            matched=1
            any_matches=1
            break
        fi
    done
    if [[ $matched == 0 ]]; then
        echo "$f"
    fi
done

if [[ $any_matches == 1 ]]; then exit 5; fi
```
</details>

(Incidentally, this script points out that runtime/pack/dist/opt/editorconfig is also not in Filelist, and maybe it should be.)

chrisbra:
Yes thanks, if we could add this to CI, similar to what is done to check for [doc/tags](https://github.com/vim/vim/blob/master/.github/workflows/ci.yml#L258-L262), that would be appreciated

And I think editorconfig should be part of the Filelist as well.

PR #13553, opened 2023-11-21
OpenVMS X86_64 platform port

arpadffy:
I have been working with the OpenVMS port since 1998, trying to keep up with the Vim development and keeping the OpenVMS port updated.
OpenVMS runs on four architectures: VAX, Alpha (AXP), Itanium (IA64), and very recently on Intel x86 (X86_64).

With this pull request the code has been aligned to work on all four supported architectures - the very first time on x86 that was in demand for very long time.

Thanks.
Z

### DIFF
--- a/src/testdir/test_textprop.vim
+++ b/src/testdir/test_textprop.vim
@@ -4410,4 +4410,19 @@ func Test_virtual_text_get()
   bwipe!
 endfunc
 
+" This used to throw: E967
+func Test_textprop_notype_join()
+  new Xtextprop_no_type_join
+  call setline(1, range(1, 3))
+  call cursor(1, 1)
+  let name = 'a'
+  call prop_type_add(name, {})
+  call prop_add(line('.'), col('.'), { 'type': name })
+  call prop_type_delete(name, {})
+  join
+  call assert_equal(["1 2", "3"], getline(1, '$'))
+
+  bwipe!
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
--- a/src/textprop.c
+++ b/src/textprop.c
@@ -630,7 +630,7 @@ get_text_props(buf_T *buf, linenr_T lnum, char_u **props, int will_change)
     size_t textlen;
     size_t proplen;
 
-    // Be quick when no text property types have been defined or the buffer,
+    // Be quick when no text property types have been defined for the buffer,
     // unless we are adding one.
     if ((!buf->b_has_textprop && !will_change) || buf->b_ml.ml_mfp == NULL)
 	return 0;
@@ -714,8 +714,7 @@ count_props(linenr_T lnum, int only_starting, int last_line)
 	// previous line, or when not in the last line and it is virtual text
 	// after the line.
 	if ((only_starting && (prop.tp_flags & TP_FLAG_CONT_PREV))
-		|| (!last_line && prop.tp_col == MAXCOL)
-		|| !text_prop_type_valid(curbuf, &prop))
+		|| (!last_line && prop.tp_col == MAXCOL))
 	    --result;
     }
     return result;
--- a/src/version.c
+++ b/src/version.c
@@ -705,6 +705,8 @@ static char *(features[]) =
 static int included_patches[] =
 {   /* Add new patch number below this line */
 /**/
+    2146,
+/**/
     2145,
 /**/
     2144,


chrisbra:
thank you. Except for the bits about the overflowing, it looks mostly good I think

errael:
(maybe more of a vim issue than directly about this PR)
Out of curiosity, would something like `typedef long long ll_t` simplify some things? But I guess the VMS compiler is not C99 compliant which violates a basic assumption.

arpadffy:
Hi,

>  I guess the VMS compiler is not C99 compliant which violates a basic
assumption

VMS currently runs on four very different architectures:
- VAX a 32 bit system that is commercially not supported any more, but many
hobbyists still use it. Has two C compilers (VAXC and an older DECC) that
are not C99 compliant.
- Alpha and Itanium are 64 bit architectures, have modern DECC compilers
that are C99 compliant.
- and the x86_64 architecture... that is very modern - in some cases not
fully backwards compatible with the older VMSes, that makes
different troubles with porting..

I have been doing this acrobatics all the time aligning the VMS code as
much as possible.
I haven't touched the code since 8.2... and I noticed that we lost old
compiler support in some places.

If there is any chance, I would suggest keeping us to the C90 standard...
otherwise we'll lose all older systems from where Vi/Vim has originated
from and users still love using their favorite editor with modern, new
features.

Thanks,
Z

chrisbra:
oh, I wasn't aware that long long is not compatible with C90. 
I think in general C90 is still the supported standard for the Vim codebase. But I don't know the solution for the overflow issue for C90 then.

arpadffy:
>  long long is not compatible with C90

no, it has nothing to do with the C90. It is a VAX related problem.
Old 32 bit VAX architecture C compilers do not support long long.

src/ops.c
```diff
@@ -249,23 +256,32 @@ shift_line(
 	}
 	else
 	    i += amount;
-	count = (long long)i * (long long)sw_val;
+	count = i * sw_val;
```
errael: Maybe something like (I don't think "`long = int*int` needs overflow check?)
```
#ifdef HAVE_NO_LONG_LONG
count = (long)i * (long)sw_val;
#else
count = (long long)i * (long long)sw_val;
#endif
```

jamessan: > I don't think "`long = int*int` needs overflow check?

On some platforms (e.g., Windows and 32-bit Linux) long and int are the same width.  A lot of Vim's code would benefit from using fixed-width integer types to avoid this inconsistency across platforms.

src/ops.c
```diff
@@ -249,23 +256,32 @@
...
     }
     else		// original vi indent
     {
 	if (left)
 	{
-	    count -= (long long)sw_val * (long long)amount;
+	    count -= sw_val * amount;
```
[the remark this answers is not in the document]
errael: If no `long long`, then cast to long. But should change what's already there (I think).

src/ops.c
errael: > On some platforms (e.g., Windows and 32-bit Linux) long and int are the same width. 

Doh! Of course.

> A lot of Vim's code would benefit from using fixed-width integer types to avoid this inconsistency across platforms.

I don't disagree. But then some stuff, like uint64_t, doesn't map 32bit VAX?

src/ops.c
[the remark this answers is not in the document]
errael: Oops. That should be "... should not change ..."

errael:
> oh, I wasn't aware that long long is not compatible with C90. I think in general C90 is still the supported standard for the Vim codebase. But I don't know the solution for the overflow issue for C90 then.

I misremembered what `:help develop` said. `C99` is not the minimal version. But there are some `C99` features listed that can be used.

Under `:help style-compiler`, one of the allowed `C99` features:
```
"long long" is allowed and can be expected to be 64 bits.
```

There's a simple test for overflow[1]
```
x = a * b;
if (a != 0 && x / a != b) {
    // overflow handling
}
```
[1] https://stackoverflow.com/questions/1815367/catch-and-compute-overflow-during-multiplication-of-two-large-integers

(BTW, looking at this as part of my penance for screwing up some overflow analysis last week)

errael:
@chrisbra Related, at `shift_line()` entry `int sw_val = (int)get_sw_value_indent(curbuf);`, if the function actually returns something bigger than `INT_MAX` then `sw_val` might go negative. Something might be detected later; in any event, looks weird.

chrisbra:
yeah, looks right

src/ops.c
arpadffy: > I don't disagree. But then some stuff, like uint64_t, doesn't map 32bit VAX?

Yes, you are right @errael . On 32bit architectures - including VAX - this needs to be handled. This is part of the porting acrobatics between different architectures. 
This is exactly what I did.

I wonder, why this "long long overflow" related conversation is started here. This piece of code has been already there. How did it pass the CI tests when it was first committed - and all other consecutive runs since?

errael: > I wonder, why this "long long overflow" related conversation is started here.

This change takes away the `long long` casting. Is that intentional? (that's how the conversation got started here).

errael: And  I wanted to point out how multiply overflow might be detected, if that's needed for VAX32.

k-takata:
We have `varnumber_T`, which was defined for the Number type of Vim script.
Using `varnumber_T` may not suitable for this case, but we can define a similar type.

src/ops.c
arpadffy: Thank you for turning my attention to this.
I did not intentionally remove the casting there. It was a consequence of working on four different architectures without functional git. The master has changed since my checkout... and I worked against 9..0.2112 and copied over the changed files - that caused this issue.

I have corrected the mistake... please review.
Thanks.

src/ops.c
```diff
@@ -249,18 +256,34 @@ shift_line(
 	}
 	else
 	    i += amount;
-	count = (long long)i * (long long)sw_val;
+	count = i * sw_val;
```
errael: `(long long)` cast removed.

errael: What is the `sizeof(long)` on VAX32?

src/ex_cmds.h
```diff
@@ -1950,6 +1950,9 @@ struct exarg
     int		bad_char;	// BAD_KEEP, BAD_DROP or replacement byte
     int		useridx;	// user command index
     char	*errmsg;	// returned error message
+#if defined (VMS) || defined (X86_64)
+    char_u  *(*getdelim)(int, void *, int, getline_opt_T );
```
k-takata:
```suggestion
    char_u	*(*getdelim)(int, void *, int, getline_opt_T );
```

src/ex_docmd.c
```diff
@@ -401,6 +401,9 @@ struct loop_cookie
     int		repeating;		// TRUE when looping a second time
     // When "repeating" is FALSE use "getline" and "cookie" to get lines
     char_u	*(*getline)(int, void *, int, getline_opt_T);
+#if defined (VMS) || defined (X86_64)
+    char_u  *(*getdelim)(int, void *, int, getline_opt_T);
```
k-takata:
```suggestion
    char_u	*(*getdelim)(int, void *, int, getline_opt_T);
```

src/ops.c
```diff
@@ -229,12 +229,19 @@ shift_line(
     int	amount,
     int call_changed_bytes)	// call changed_bytes()
 {
+#ifdef HAVE_NO_LONG_LONG // long long does not exits on all architectures
+    long   count;
```
k-takata:
```suggestion
    long	count;
```

src/ops.c
```diff
@@ -249,18 +256,34 @@ shift_line(
...
     }
     else		// original vi indent
     {
 	if (left)
 	{
-	    count -= (long long)sw_val * (long long)amount;
+#ifdef HAVE_NO_LONG_LONG
+	    count -= (long)sw_val * (long)amount;
+#else
+        count -= (long long)sw_val * (long long)amount;
```
k-takata:
```suggestion
	    count -= (long long)sw_val * (long long)amount;
```

src/strings.c
```diff
@@ -2789,7 +2789,11 @@ skip_to_arg(
     {
 	// Reset ap to ap_start and skip arg_idx - 1 types
 	va_end(*ap);
-	va_copy(*ap, ap_start);
+#ifdef  va_copy
+	va_copy (*ap, ap_start);
+#else
+    *ap = ap_start;
```
k-takata:
```suggestion
	*ap = ap_start;
```

src/proto/crypt.pro
```diff
@@ -30,5 +30,9 @@ void crypt_sodium_lock_key(char_u *key);
 int crypt_sodium_munlock(void *const addr, const size_t len);
 void crypt_sodium_randombytes_buf(void *const buf, const size_t size);
 int crypt_sodium_init(void);
+#ifdef VMS
+UINT32_T crypt_sodium_randombytes_random(void);
+#else
 uint32_t crypt_sodium_randombytes_random(void);
+#endif
```
k-takata: `crypt.pro` is automatically generated by using cproto. This change will be discarded next time.

arpadffy: Thank you for the information.
What would be the correct way to implement this change?

k-takata: I think we can simply use UINT32_T here (and in crypt.c).

src/ops.c
arpadffy: On VAX I got the following:
The size of int  is: 4 bytes
The size of long is: 4 bytes

	printf("The size of long long is: %d bytes\n", sizeof(long long));
	......................................................^
%CC-E-NOLONGLONG, In this statement, 64-bit integral types are not supported on
 this platform.

The size of float is: 4 bytes
The size of double is: 8 bytes

vim-ml:
On Tue 21 Nov 2023 at 14:27:52 -0800, errael wrote:
> There's a simple test for overflow[1]
> ```
> x = a * b;
> if (a != 0 && x / a != b) {
>     // overflow handling
> }
> ```
> [1] https://stackoverflow.com/questions/1815367/catch-and-compute-overflow-during-multiplication-of-two-large-integers

Unfortunately checking for overflow after the fact cannot work. If
Undefined Behavior happens, the compiler can do anything. For example,
when optimizing, gcc seems to assume that overflows do not happen, and
therefore that the if condition is never true. So you cannot count on
the overflow handling to be effective.

src/proto/crypt.pro
arpadffy: Thank you. 
I have just committed the change in this pull request.

src/ops.c
```diff
@@ -249,18 +256,34 @@ shift_line(
...
+	    if ((long)sw_val * (long)amount > INT_MAX - count)
+		count = INT_MAX;
+	    else
+		count += (long)sw_val * (long)amount;
+#else
+	    if ((long long)sw_val * (long long)amount > INT_MAX - count)
+		count = INT_MAX;
+	    else
+		count += (long long)sw_val * (long long)amount;
+#endif
```
Shane-XB-Qian: i'm curious to ask: 
the port here the trick is like this to make it pick up different length 'count'?
vs/but not a way to mock it work like a fake 64bit long?
not sure if it's possible, or con/pro or outage.

src/ops.c
```diff
@@ -249,18 +256,34 @@
 	}
 	else
 	    i += amount;
-	count = (long long)i * (long long)sw_val;
+	count = i * sw_val;
```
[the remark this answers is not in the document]
errael: @arpadffy Here's another case of the merge problem.

errael (changes requested):
In `shift_line()` fix merge problem around `long long`

chrisbra:
please rebase on v9.0.2122, I had to touch the ops.c just now so make sure it doesn't interfer your work here.

src/ops.c
```diff
@@ -249,20 +256,39 @@ shift_line(
...
+		count = INT_MAX;
+	    else
+		count += (long long)sw_val * (long long)amount;
+#endif
+	}
     }
 
+    if (count > INT_MAX)
+	count = INT_MAX;
+
```
errael: These were removed in favor on using `trim_to_int(count)` in the following statements.

src/ops.c
```diff
@@ -229,12 +229,19 @@ shift_line(
...
+#ifdef HAVE_NO_LONG_LONG // long long does not exits on all architectures
+    long	count;
+#else
     long long	count;
-    int		i, j;
-    int		sw_val = trim_to_int(get_sw_value_indent(curbuf));
+#endif
 
+    int		i, j;
+    int		sw_val = (int)get_sw_value_indent(curbuf);
```
errael: This should keep the original: `trim_to_int(get_sw_value_indent(curbuf))`

src/ops.c
```diff
@@ -229,12 +229,19 @@ shift_line(
...
+#ifdef HAVE_NO_LONG_LONG
+    count = (long)get_indent();         // get current indent
+#else
     count = (long long)get_indent();	// get current indent
-
+#endif
```
errael: @chrisbra Are you OK with 
```
count = get_indent()
```
the same for regular and vax32. 

No matter what `get_indent()` returns, count is the biggest int, so never a narrowing assignment, no overflow problem.

src/ops.c
```diff
@@ -249,20 +256,39 @@ shift_line(
 	}
 	else
 	    i += amount;
-	count = (long long)i * (long long)sw_val;
+	count = i * sw_val;
```
errael: This change in incorrect. The `(long long)` cast is required for the regular case.

src/ops.c
```diff
@@ -249,20 +256,39 @@ shift_line(
...
+	{
+#ifdef HAVE_NO_LONG_LONG
+	    if ((long)sw_val * (long)amount > INT_MAX - count)
+		count = INT_MAX;
+	    else
+		count += (long)sw_val * (long)amount;
+#else
+	    if ((long long)sw_val * (long long)amount > INT_MAX - count)
+		count = INT_MAX;
+	    else
```
errael: These are not needed when sizeof(int) != sizeof(long long).

errael (changes requested):

errael:
@arpadffy Doesn't the gnu toolchain work on vax?

arpadffy:
> @arpadffy Doesn't the gnu toolchain work on vax?

@errael , unfortunately it does not. The VAX technology is about 10 years older than the GNU project and it has never got supported on VAX. There has been initiatives with the GNV project to port the GNU chain to other VMS architectures, but I am not aware that it has got a wide spread use. Software porting projects do not count on GNU tools existence (especially that Vim is part of the GNV)
The VMS community is very resistant - as the VMS operating system was brilliant and outstanding with many advanced features from the beginning.

chrisbra:
I am still slightly worried about possibly overflowing in `ops.c` otherwise, looks mostly good I think

arpadffy:
Thank you Christian.

I have tested on four architectures both terminal and GUI version... and it
works well now.
I plan to publish these executables (9.0-2130).
The X86_64 port is especially important... as there were no Vim on that
platform yet.
You can imagine what a nightmare was porting and programming in an
environment without Vim (if you got used to it on other platforms.)

Thank you for all your help and support.

Regards,
Z.

chrisbra:
> I reviewed the changes 2 days ago and identified some things that revert (and not in a good way) what's currently committed.

> A response would be appreciated.

You mean me? I think I am with you here.

errael:
> I have tested on four architectures both terminal and GUI version... and it works well now. I plan to publish these executables (9.0-2130). The X86_64 port is especially important... as there were no Vim on that platform yet.

Since 3 of the 4 OpenVMS architectures have `long long`, and since @arpadffy is publishing the vim/gvim binaries, I wonder if it makes sense to only include the changes for 
- Alpha (AXP)
- Itanium (IA64)
- Intel x86 (X86_64)

into the vim source code repository. And forgo the `VAX` changes since its compiler does not meet the vim requirements as stated in `:help develop`.

This approach might serve the most people without removing the `long long` requirement.

errael:
> > I reviewed the changes 2 days ago and identified some things that revert (and not in a good way) what's currently committed.
> 
> > A response would be appreciated.
> 
> You mean me? I think I am with you here.

I wasn't sure anyone had seen the review.

chrisbra:
I know what you mean, I saw your comments and since I didn't see any reactions, I added my comments below yours.

yegappan:
The test_indent.vim test is failing with this change: (https://github.com/vim/vim/actions/runs/7007399609)

```
From test_indent.vim:
Found errors in Test_indent_overflow_count2():
command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_indent_overflow_count2 line 10: Expected 0 but got 2147483647
```

arpadffy:
>Since 3 of the 4 OpenVMS architectures have long long, [...] I wonder if
it makes sense to only include the >changes for Alpha (AXP), Itanium
(IA64)Intel x86 (X86_64)
>  And forgot the VAX changes since its compiler does not meet the vim
requirements as stated in :help develop.

I appreciate your suggestion @errael and understand the reasoning behind -
that clearly makes sense.
However, I haven't seen anything in the latest develop.txt that would
suggest discontinuing support for the OpenVMS VAX architecture.

4. Assumptions						**design-assumptions**

Size of variables:
char	    8 bit signed
char_u	    8 bit unsigned
int	    32 or 64 bit signed (16 might be possible with limited features)
unsigned    32 or 64 bit unsigned (16 as
<http://polarhome.com/vim/manual/v90/motion.html#as> with ints)
long	    32 or 64 bit signed, can hold a pointer

Note that some compilers cannot handle long lines or strings.  The C89
standard specifies a limit of 509 characters.


I would read the assumptions above, that assuming the existence of 64
bit variables and irresponsible use of the long long type is against
Vim's multiplatform design.

Also that C89 standard should be considered.

I know that a poll on Vim OpenVMS download site
(http://polarhome.com/vim/) is not relevant, but about third of the
users voted that use Vim on VAX

*What platform and OpenVMS version do you use?*
VAX - 7.1 or older *26.42 *%
VAX - 7.2 or newer *8.94 *%
Alpha - 7.2 or older *4.88 *%
Alpha - 7.3 or newer *29.13 *%
Itanium - 8.0 or newer *25.75 *%
I do not use OpenVMS *3.25 *%
I do not know *1.63 *%

*In what mode do you use Vim on OpenVMS?*
Console mode  *81.41 *%
GUI/Motif  *6.69 *%
GUI/GTK  *3.35 *%
Mixing  *6.69 *%
I don't use Vim  *1.86 *%

Why is the VAX - a 30 years old - architecture still popular?

OpenVMS has alway run on very special hardware - expensive, cutting
edge, one of the most powerful architectures of their time - that
ordinary home users could not appreciate.

This led to the decline of the Operating system that introduced
virtual memory and many other advanced features to the world that has
been appreciated by many professionals, still.

Thanks to the SIMH project (http://simh.trailing-edge.com/) the VAX
architecture became available to a wider audience, and many OpenVMS
hobbyists use it.

Why the Itanium and Alpha hardware are very rarely available,
expensive to run and server and most available simulators are costly,
the VAX was hanging around.

The very recent VSI OpenVMS port to X86_64 architecture made a leap
forward towards the 21st century's commonly used  Intel architecture,
first time in the history providing VMS to the masses.


Therefore, I kindly ask to reconsider keeping VAX among the Vim
supported platforms.

Thanks,

Z

errael:
> I haven't seen anything in the latest develop.txt that would suggest discontinuing support for the OpenVMS VAX architecture.

On that page, `:help develop`, search for `long long`. Under the section `C COMPILER`, there is
```
"long long" is allowed and can be expected to be 64 bits.  Use %lld in printf
formats.  Also "long long unsigned" with %llu.
```

In the discussion on VMS, it looks pretty clear that the growth is in non-VAX hardware; all the hardware, other than VAX, supports `long long`. Not only is there currently a working vim on VAX, you are planning to provide current vim binaries for VAX based on your work.

And then the [OpenVMS – A guide to the strategy and roadmap](https://vmssoftware.com/about/roadmap/)
says for "VAX Support Roadmap"
```
PVS: Prior Version Support without Sustaining Engineering.
```
It looks like **OpenVMS is abandoning the VAX**.

arpadffy:
> > I am still slightly worried about possibly overflowing in `ops.c` otherwise, looks mostly good I think
> I reviewed the changes 2 days ago and identified some things that revert (and not in a good way) what's currently committed.
> A response would be appreciated.

I am sorry, I haven't noticed this change after the rebase.
I have corrected the issue, tested on both architectures with and without long long and commited the new ops.c

arpadffy:
> It looks like *OpenVMS is abandoning the VAX*.

Indeed, this is the sad truth that VSI, who took over the OpenVMS operating
system from HP, supports just Alpha, Itanium and very recently X86.
...mostly because the VAX architecture is not suitable for commercial use
any more, but the OpenVMS community still widely uses VAX.
I hope it will hang around for a few more years.

k-takata:
Adding `#ifdef HAVE_NO_LONG_LONG` everywhere makes the source code noisy.
C99's stdint.h has `intmax_t`. How about using it when available?
E.g.:
```c
#ifdef HAVE_STDINT_H
// Use intmax_t from stdint.h.
#else
// Define intmax_t by ourselves.
# ifdef HAVE_NO_LONG_LONG
typedef long intmax_t;
# else
typedef long long intmax_t;
# endif
#endif
```
Or, define our own integer type like `intmax_T`.
Or, as I said above, we might be able to use `varnumber_T`.

arpadffy:
> Adding `#ifdef HAVE_NO_LONG_LONG` everywhere makes the source code noisy. C99's stdint.h has `intmax_t`. How about using it when available? 

I fully agree with you.

src/strings.c
```diff
@@ -2789,7 +2789,11 @@ skip_to_arg(
     {
 	// Reset ap to ap_start and skip arg_idx - 1 types
 	va_end(*ap);
-	va_copy(*ap, ap_start);
+#ifdef  va_copy
+	va_copy (*ap, ap_start);
+#else
+	*ap = ap_start;
+#endif
```
k-takata: How about defining `va_copy` like this?
```c
#ifndef va_copy
# define va_copy(dst, src)	((dst) = (src))
#endif
```

src/xdiff/xmacros.h
```diff
@@ -24,7 +24,7 @@
 #define XMACROS_H
 
 
-#ifdef __hpux
+#if defined( __hpux ) || defined ( VMS )
```
k-takata:
```suggestion
#if defined(__hpux) || defined(VMS)
```

src/proto/misc1.pro
```diff
@@ -55,5 +55,9 @@ void restore_v_event(dict_T *v_event, save_v_event_T *sve);
 void may_trigger_modechanged(void);
 int vim_append_digit_int(int *value, int digit);
 int vim_append_digit_long(long *value, int digit);
+#ifdef HAVE_NO_LONG_LONG
+int trim_to_int(long x);
+#else
 int trim_to_int(long long x);
+#endif
```
k-takata: As I said above, *.pro files are generated files. This change will be overwritten next time.

src/os_vms_conf.h
```diff
@@ -148,27 +155,51 @@
...
 #define VIM_SIZEOF_LONG 4
 #define LONG_LONG_MIN  (-2147483647-1)
 #define LONG_LONG_MAX  (2147483647)
 #define ULONG_LONG_MAX (4294967295U)
-#else // AXP and IA64
+
+#else // ALPHA, IA64, X86_64
+#if defined(__alpha) || defined(__ALPHA)
+#if !defined(ALPHA)
+#define ALPHA
```
k-takata: Aren't the definition name `ALPHA`, `IA64` and `X86_64` too generic?
Are they defined by the OpenVMS compiler?

arpadffy: because of the explicit restriction by __CRTL_VER_OVERRIDE=80400000 the X86_64 is lost and needed to define for current and future use.
But you are right... defining others like ALPHA is an overkill.
I'll remove them.

errael:
> Adding `#ifdef HAVE_NO_LONG_LONG` everywhere makes the source code noisy. C99's stdint.h has `intmax_t`. How about using it when available?
> ...
> 
> Or, define our own integer type like `intmax_T`. Or, as I said above, we might be able to use `varnumber_T`.

I'd like to see the source code noise gone. But note that the overflow detection code in `ops.c` depends on `sizeof(long long) >= 2 * sizeof(int)`; using `intmax_t` might be confusing. I don't know of any plans, but I wouldn't be surprised if more overflow detection code is introduced as time goes on.

`int_overflow_T` seems verbose, but the intent is clear; maybe just `overflow_T` since floating point is a whole different thing. I don't know what, if any, impact there might be on overflow detection code if the assumed size constraint is violated.

arpadffy:
I have updated the VMS related documentation after the first binary release of the 9.0 version of the X86_64 OpenVMS build that is exposed on http://polarhome.com/vim/ 
The OpenVMS community finally got a functional Vim on the X86_64 platform too.

I wonder what else needs to be done  to get this pull request approved and merged?
Thanks.

chrisbra:
Thanks. Can you please rebase on top of https://github.com/vim/vim/commit/fda700cb04612fe2f9301a9ba820309175decabf ? This should make your life easier with the overflow detection. Just need to define a proper vimlong_T datatype.

errael:
> I wonder what else needs to be done to get this pull request approved and merged?

In addition to rebase, there's also https://github.com/vim/vim/pull/13553#discussion_r1407314149 to define `va_copy` for vax-vms.

src/vim.h
```diff
@@ -435,6 +435,15 @@ typedef unsigned short sattr_T;
...
 
+/*
+ * The vimlong_T has sizeof(vimlong_T) >= 2 * sizeof(int).
+ * One use is simple handling of overflow in int calculations.
+ */
+#ifdef HAVE_NO_LONG_LONG
+typedef long vimlong_T;
+#else
+typedef long long vimlong_T;
+#endif
```
k-takata:
```suggestion
#endif

```

src/vim.h
```diff
@@ -435,6 +435,15 @@ typedef unsigned short sattr_T;
  */
 typedef unsigned int u8char_T;	// int is 32 bits or more
 
+/*
+ * The vimlong_T has sizeof(vimlong_T) >= 2 * sizeof(int).
+ * One use is simple handling of overflow in int calculations.
+ */
+#ifdef HAVE_NO_LONG_LONG
```
errael:
```suggestion
#if defined(VMS) && defined(VAX)
```

Using system definitions seems more consistent with what's already in `vim.h`

Then can get rid of definitions and references to `HAVE_NO_LONG_LONG`

errael:
@arpadffy The suggestion from https://github.com/vim/vim/pull/13553#discussion_r1407314149 could be done after
```
#include <stdarg.h>
```
in `vim.h`

arpadffy:
I am sorry, I have terrible merging problems while I try to rebase. I
already spent 2 hours circling merges
I'll sleep on it... and tomorrow I'll just discard all 77 commits and start
from scratch.
It will be easier, as the master contains now most of the changes that VMS
needs.

arpadffy:
> @arpadffy The suggestion from [#13553 (comment)](https://github.com/vim/vim/pull/13553#discussion_r1407314149) could be done after
> 
> ```
> #include <stdarg.h>
> ```
> 
> in `vim.h`

Thank you for the reminder. This is a much better solution. I'll add it to the pull request.

src/vim.h
arpadffy: Introducing HAVE_NO_LONG_LONG was Bram's idea back in the time as there are other different platforms and architectures that do not have long long time at all.
Now searching the code, seems you are right. Just VAX is the only one left using it.

I would prefer leaving HAVE_NO_LONG_LONG as this clearly explains why we are doing this acrobatics with the types (also it is defined in the os_*.h files that clearly narrows the scope)

PR #13555, opened 2023-11-22
shiftwidth value may overflow

### LINKED ISSUE #13554: A couple expression overflow
### Steps to reproduce

```
echo foobar > xxx
gvim xxx
:set sw=0x180000000
<<
```
This results in
```
E342: Out of memory!  (allocating 18446744071562067974 bytes)
```
Not sure there is a bug here. I found two undetected overflows on the way to the out of memory error. I recently saw some error handing that was added to prevent overflow; in those cases the overflow didn't necessarily result in a bad memory access but it was detected by "Code scanning / CodeQL".

### Expected behaviour

It's unclear to me what error handling is **required** to do (other than prevent accessing memory that's not yours). Are overflows OK if they don't cause memory access errors? I guess they're not OK if a reliability tool flags them.

This report might be more about getting clarification; __closed as not planned__ seems OK. _Maybe this should be a discussion._

With the above setting of `sw`, upon entry to `shift_line()` there's
```
sw_val = (int)get_sw_value_indent(curbuf)
```
with values as seen by the debugger
```
p get_sw_value_indent(curbuf)           : $6 = 6442450944
p (int)get_sw_value_indent(curbuf)      : $7 = -2147483648
```
that demonstrates a negative `sw_val` from overflow.

At the end of `shift_line()` there's a call to `set_indent()` with count `INT_MAX`.
And then in `set_indent()` around line 681 there's
```
newline = alloc(ind_len + line_len)
```
the addition overflows
```
(gdb) p/x ind_len               : $12 = 0x7fffffff
(gdb) p line_len                : $13 = 13
(gdb) p/x ind_len + line_len    : $14 = 0x8000000c     ***OVERFLOW***
```
And at entry to lalloc, the negative number is sign extended into a size_t
```
(gdb) p/x size                  : $15 = 0xffffffff8000000c
```
Which is the value reported by the `Out of memory`

### Version of Vim

9.0.2116

### Environment

ubuntu

### Logs and stack traces

_No response_

chrisbra:
The shiftwidth option values are defines as being long. However, when calculating the actual amount of indent, we cast down to (int), which may cause the shiftwidth value to become negative and later it may even cause Vim to try to allocate a huge amount of memory.

We already use long and long long variable types to calculate the indent (and detect possible overflows), so the cast to (int) seems superfluous and can be safely removed. So let's just remove the (int) cast and calculate the indent using longs.

fixes: #13554

### DIFF
--- a/src/ops.c
+++ b/src/ops.c
@@ -230,8 +230,8 @@ shift_line(
     int call_changed_bytes)	// call changed_bytes()
 {
     long long	count;
-    int		i, j;
-    int		sw_val = (int)get_sw_value_indent(curbuf);
+    long	i, j;
+    long	sw_val = get_sw_value_indent(curbuf);
 
     count = (long long)get_indent();	// get current indent
 
--- a/src/cindent.c
+++ b/src/cindent.c
@@ -1730,10 +1730,17 @@ parse_cino(buf_T *buf)
     char_u	*p;
     char_u	*l;
     char_u	*digits;
-    int		n;
+    long long	n;
     int		divider;
     int		fraction = 0;
-    int		sw = (int)get_sw_value(buf);
+    int		sw;
+    long long	t = get_sw_value(buf);
+
+    // needed for cino-(, it will be multiplied by 2 again
+    if (t > INT_MAX / 2)
+	sw = INT_MAX / 2;
+    else
+	sw = (int)t;
 
     // Set the default values.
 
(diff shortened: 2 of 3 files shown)


src/ops.c
```diff
@@ -230,8 +230,8 @@ shift_line(
     int call_changed_bytes)	// call changed_bytes()
 {
     long long	count;
-    int		i, j;
-    int		sw_val = (int)get_sw_value_indent(curbuf);
+    long	i, j;
+    long	sw_val = get_sw_value_indent(curbuf);
```
errael, line 234: Changing `sw_val`, and the rest to `long` invalidates at least some of the overflow checking/calculation is this function (I think, probably depending on platform) for the case where `sizeof(long) == sizeof(long long)`

I wonder if something like the following would be useful (or exists?)
```
#define INT_TRIM(x) ((x) > INT_MAX ? INT_MAX : (x) < INT_MIN ? INT_MIN : (x))
```
Then could do
```
long sw_tmp = get_sw_value_indent(curbuf);
int  sw_val = INT_TRIM(sw_tmp);
```
In the code this PR adds to cindent.c could do
```
n = INT_TRIM(n);
```
right before the `switch`.

Using an inline function, rather than a macro, would be nice to avoid needing to manually create the tmp before using the macro.

chrisbra: Let me merge it, seems to be okay now. Please go ahead and create a PR for this followup change then.

errael: > Let me merge it, seems to be okay now. Please go ahead and create a PR for this followup change then.

OK.

(I guess that "seems to be okay" means that options are now < INT_MAX)

chrisbra: Yes, practically. I have not documented this explicitly, as I am not sure if we may need to support longs for other options specifically.

PR #13556, opened 2023-11-22
Introduce and use INT_TRIM/trim_to_int.

errael:
The macro typically used to constrain a long to fit into an int.

Initially used to trim int options sizes, and defensively in `shift_line()`.

@chrisbra If `sw` remains constrained to an int, then the use of INT_TRIM at the entry in `shift_line()` is not required. LMK if you want that change.

### DIFF
--- a/src/cindent.c
+++ b/src/cindent.c
@@ -1734,7 +1734,7 @@ parse_cino(buf_T *buf)
     int		divider;
     int		fraction = 0;
     int		sw;
-    long long	t = get_sw_value(buf);
+    long	t = get_sw_value(buf);
 
     // needed for cino-(, it will be multiplied by 2 again
     if (t > INT_MAX / 2)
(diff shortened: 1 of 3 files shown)


errael:
I was looking at _gcc_ docs. I guess `INT_TRIM` could be made a function, and with sufficient optimization, then function inlining automagically happens. So instead of the macro, could put `int trimToInt(long long)` somewhere. (In ops.c?). Then the entry to `shift_line()` could avoid the temp variable and simply have
```
    int		sw_val = trimToInt(get_sw_value_indent(curbuf));
```

Thinking about it, if `get_sw_*` functions are contrained to `int`, then why are they declared to return `long`; probably best to be defensive and assume `long` until/unless their declaration changes. So still want trim.

chrisbra:
> Thinking about it, if get_sw_* functions are contrained to int, then why are they declared to return long; probably best to be defensive and assume long until/unless their declaration changes. So still want trim.

Yes agree. 

Coverity also detected another issue in `parse_cino`:
```
*** CID 1554344:  Integer handling issues  (OVERFLOW_BEFORE_WIDEN)
/src/cindent.c: 1905 in parse_cino()
1899                if (p == digits)
1900                    n = sw; // just "s" is one 'shiftwidth'
1901                else
1902                {
1903                    n *= sw;
1904                    if (divider)
>>>     CID 1554344:  Integer handling issues  (OVERFLOW_BEFORE_WIDEN)
>>>     Potentially overflowing expression "sw * fraction" with type "int" (32 bits, signed) is evaluated using 32-bit arithmetic, and then used in a context that expects an expression of type "long long" (64 bits, signed).
1905                        n += (sw * fraction + divider / 2) / divider;
1906                }
1907                ++p;
1908            }
1909            if (l[1] == '-')
1910                n = -n;
```

So a bit more to fix :(

errael:
> So a bit more to fix :(

Handled (I think).

This PR needs some "final" cleanup. (waiting for comment on macro vs function vs inline) Preference?

chrisbra:
I don't really have a preference. In general I am with Bram on the inlining but I don't really mind

errael:
Ready for "final" review.

chrisbra:
thanks looks good. Appreciate it 🙏

PR #13559, opened 2023-11-23
Avoid unused assignments when checking the value of 'listchars'

zeertzjq:
Problem:  Unused assignments when checking the value of 'listchars'.
Solution: Loop only once when just checking the value.  Add a test to
          check that this change doesn't cause double-free.


chrisbra:
thanks!

PR #13561, opened 2023-11-23
File info disappears immediately when 'cmdheight' has just decreased

### LINKED ISSUE #13560: File message disappears immediately after switching from tabpage with different 'cmdheight' when 'shortmess' doesn't contain 'o' or 'O'
### Steps to reproduce

1. Run `vim --clean`.
2. Source the following file:
```vim
set shortmess-=o
set shortmess-=O
tabnew
set cmdheight=2
tabnext
edit newfile.txt
```
3. `"newfile.txt" [New]` appears on the last but one screen row for a very short time, and then disappears immediately.

### Expected behaviour

`"newfile.txt" [New]` appears on the last screen row and doesn't disappear.

### Version of Vim

9.0.2112

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

zeertzjq:
Problem:  File info disappears immediately when 'cmdheight' has just
          decreased due to switching tabpage and 'shortmess' doesn't
          contain 'o' or 'O'.
Solution: Make sure msg_row isn't smaller than cmdline_row.

Fix #13560

### DIFF
--- a/src/message.c
+++ b/src/message.c
@@ -1446,6 +1446,9 @@ msg_start(void)
 {
     int		did_return = FALSE;
 
+    if (msg_row < cmdline_row)
+	msg_row = cmdline_row;
+
     if (!msg_silent)
     {
 	VIM_CLEAR(keep_msg);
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #13562, opened 2023-11-23
Syntax file for SWIG (Simplified Wrapper Interface Generator) description files

jmarrec:
* Supersedes https://github.com/vim/vim/pull/13540 by @mcepl
* Redid the syntax file from scratch, wrote some tests, added a minimalistic ftplugin file to set the `iskeyword+=%`

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -590,18 +590,27 @@ export def FTprogress_cweb()
   endif
 enddef
 
+# These include the leading '%' sign
+var ft_swig_keywords = '^\s*%\%(addmethods\|apply\|beginfile\|clear\|constant\|define\|echo\|enddef\|endoffile\|extend\|feature\|fragment\|ignore\|import\|importfile\|include\|includefile\|inline\|insert\|keyword\|module\|name\|namewarn\|native\|newobject\|parms\|pragma\|rename\|template\|typedef\|typemap\|types\|varargs\|warn\)'
+# This is the start/end of a block that is copied literally to the processor file (C/C++)
+var ft_swig_verbatim_block_start = '^\s*%{'
+var ft_swig_verbatim_block_end = '^\s*}%'
+
 export def FTprogress_asm()
   if exists("g:filetype_i")
     exe "setf " .. g:filetype_i
     return
   endif
-  # This function checks for an assembly comment the first ten lines.
+  # This function checks for an assembly comment or a SWIG keyword or verbatim block in the first 50 lines.
   # If not found, assume Progress.
   var lnum = 1
-  while lnum <= 10 && lnum < line('$')
+  while lnum <= 50 && lnum < line('$')
     var line = getline(lnum)
     if line =~ '^\s*;' || line =~ '^\*'
       FTasm()
+      return
+    elseif line =~ ft_swig_keywords || line =~ ft_swig_verbatim_block_start || line =~ ft_swig_verbatim_block_end
+      setf swig
       return
     elseif line !~ '^\s*$' || line =~ '^/\*'
       # Not an empty line: Doesn't look like valid assembly code.
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -594,9 +594,8 @@ enddef
 var ft_swig_keywords = '^\s*%\%(addmethods\|apply\|beginfile\|clear\|constant\|define\|echo\|enddef\|endoffile\|extend\|feature\|fragment\|ignore\|import\|importfile\|include\|includefile\|inline\|insert\|keyword\|module\|name\|namewarn\|native\|newobject\|parms\|pragma\|rename\|template\|typedef\|typemap\|types\|varargs\|warn\)'
 # This is the start/end of a block that is copied literally to the processor file (C/C++)
 var ft_swig_verbatim_block_start = '^\s*%{'
-var ft_swig_verbatim_block_end = '^\s*}%'
 
-export def FTprogress_asm()
+export def FTi()
   if exists("g:filetype_i")
     exe "setf " .. g:filetype_i
     return
@@ -609,7 +608,7 @@ export def FTprogress_asm()
     if line =~ '^\s*;' || line =~ '^\*'
       FTasm()
       return
-    elseif line =~ ft_swig_keywords || line =~ ft_swig_verbatim_block_start || line =~ ft_swig_verbatim_block_end
+    elseif line =~ ft_swig_keywords || line =~ ft_swig_verbatim_block_start
       setf swig
       return
     endif
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -612,10 +612,6 @@ export def FTprogress_asm()
     elseif line =~ ft_swig_keywords || line =~ ft_swig_verbatim_block_start || line =~ ft_swig_verbatim_block_end
       setf swig
       return
-    elseif line !~ '^\s*$' || line =~ '^/\*'
-      # Not an empty line: Doesn't look like valid assembly code.
-      # Or it looks like a Progress /* comment
-      break
     endif
     lnum += 1
   endwhile
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2184,7 +2184,7 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
 " Swig
-au BufNewFile,BufRead *.swg,*.swig set filetype=swig
+au BufNewFile,BufRead *.swg,*.swig setf swig
 
 " Sysctl
 au BufNewFile,BufRead */etc/sysctl.conf,*/etc/sysctl.d/*.conf	setf sysctl
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1689,7 +1689,7 @@ au BufNewFile,BufRead .procmail,.procmailrc	setf procmail
 au BufNewFile,BufRead *.w			call dist#ft#FTprogress_cweb()
 
 " Progress or assembly or Swig
-au BufNewFile,BufRead *.i			call dist#ft#FTprogress_asm()
+au BufNewFile,BufRead *.i			call dist#ft#FTi()
 
 " Progress or Pascal
 au BufNewFile,BufRead *.p			call dist#ft#FTprogress_pascal()
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2183,6 +2183,9 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 " Swift Intermediate Language or SILE
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
+" Swig
+autocmd BufNewFile,BufRead *.i,*.swg,*.swig set filetype=swig
+
 " Sysctl
 au BufNewFile,BufRead */etc/sysctl.conf,*/etc/sysctl.d/*.conf	setf sysctl
 
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1688,7 +1688,7 @@ au BufNewFile,BufRead .procmail,.procmailrc	setf procmail
 " Progress or CWEB
 au BufNewFile,BufRead *.w			call dist#ft#FTprogress_cweb()
 
-" Progress or assembly
+" Progress or assembly or Swig
 au BufNewFile,BufRead *.i			call dist#ft#FTprogress_asm()
 
 " Progress or Pascal
@@ -2184,7 +2184,7 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
 " Swig
-autocmd BufNewFile,BufRead *.i,*.swg,*.swig set filetype=swig
+au BufNewFile,BufRead *.swg,*.swig set filetype=swig
 
 " Sysctl
 au BufNewFile,BufRead */etc/sysctl.conf,*/etc/sysctl.d/*.conf	setf sysctl
--- a/runtime/ftplugin/swig.vim
+++ b/runtime/ftplugin/swig.vim
@@ -1,0 +1,11 @@
+" Vim filetype plugin file
+" Language:	SWIG
+" Maintainer:	Julien Marrec <julien.marrec 'at' gmail com>
+" Last Change:	2023 November 23
+
+" Only do this when not done yet for this buffer
+if exists("b:did_ftplugin")
+  finish
+endif
+
+setlocal iskeyword+=%
--- a/runtime/ftplugin/swig.vim
+++ b/runtime/ftplugin/swig.vim
@@ -7,5 +7,7 @@
 if exists("b:did_ftplugin")
   finish
 endif
+let b:did_ftplugin = 1
 
+let b:undo_ftplugin = "setlocal iskeyword<"
 setlocal iskeyword+=%
--- a/runtime/syntax/swig.vim
+++ b/runtime/syntax/swig.vim
@@ -1,0 +1,29 @@
+" Vim syntax file
+" Language:	SWIG
+" Maintainer:	Roman Stanchak (rstanchak@yahoo.com)
+" Last Change:	2023 April 27
+
+if exists("b:current_syntax")
+	finish
+endif
+
+" Read the C++ syntax to start with
+runtime! syntax/cpp.vim
+unlet b:current_syntax
+
+" SWIG extentions
+syn keyword swigDirective %typemap %define %apply %fragment %include %enddef %extend %newobject %name
+syn keyword swigDirective %rename %ignore %keyword %typemap %define %apply %fragment %include
+syn keyword swigDirective %enddef %extend %newobject %name %rename %ignore %template %module %constant
+syn match swigDirective "%\({\|}\)"
+syn match swigUserDef "%[-_a-zA-Z0-9]\+"
+
+" Default highlighting
+command -nargs=+ HiLink hi def link <args>
+HiLink swigDirective	Exception
+HiLink swigUserDef 		PreProc
+delcommand HiLink
+
+let b:current_syntax = "swig"
+
+" vim: ts=8
--- a/runtime/syntax/swig.vim
+++ b/runtime/syntax/swig.vim
@@ -62,6 +62,7 @@ syn keyword swigDirective %std_nodefconst_type %typecheck %typemaps_string %uniq
 syn match swigVerbatimStartEnd "%[{}]"
 
 syn match swigUserDef "%[-_a-zA-Z0-9]\+"
+syn match swigVerbatimMacro "^\s*%#[-_a-zA-Z0-9]\+\%( .*\)\?$"
 
 " Default highlighting
 hi def link swigInclude Include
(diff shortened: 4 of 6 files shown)


runtime/autoload/dist/ft.vim
```diff
@@ -590,23 +590,28 @@ export def FTprogress_cweb()
...
+var ft_swig_verbatim_block_start = '^\s*%{'
+var ft_swig_verbatim_block_end = '^\s*}%'
+
 export def FTprogress_asm()
   if exists("g:filetype_i")
     exe "setf " .. g:filetype_i
     return
   endif
-  # This function checks for an assembly comment the first ten lines.
+  # This function checks for an assembly comment or a SWIG keyword or verbatim block in the first 50 lines.
```
jmarrec, line 603: The idea here is that since a SWIG interface file is similar to a C/C++ file, it's not uncommon to have them start with a huge copyright notice or something, so parsing 10 lines isn't enough.

runtime/autoload/dist/ft.vim
```diff
@@ -590,23 +590,28 @@ export def FTprogress_cweb()
...
-  while lnum <= 10 && lnum < line('$')
+  while lnum <= 50 && lnum < line('$')
     var line = getline(lnum)
     if line =~ '^\s*;' || line =~ '^\*'
       FTasm()
       return
-    elseif line !~ '^\s*$' || line =~ '^/\*'
-      # Not an empty line: Doesn't look like valid assembly code.
-      # Or it looks like a Progress /* comment
-      break
```
jmarrec, line 609: Cannot assume this break here anymore.

runtime/ftplugin/swig.vim
```diff
@@ -0,0 +1,11 @@
...
+" Language:	SWIG
+" Maintainer:	Julien Marrec <julien.marrec 'at' gmail com>
+" Last Change:	2023 November 23
+
+" Only do this when not done yet for this buffer
+if exists("b:did_ftplugin")
+  finish
+endif
+
+setlocal iskeyword+=%
```
jmarrec, line 13: set iskeyword to include %

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,94 @@
...
+syn keyword swigDirective %implicit %implicitconv %interface %interface_custom %interface_impl %intrusive_ptr %intrusive_ptr_no_wrap
+syn keyword swigDirective %mutable %naturalvar %nocallback %nocopyctor %nodefaultctor %nodefaultdtor %nonaturalvar %nonspace
+syn keyword swigDirective %nspace %pointer_cast %pointer_class %pointer_functions %predicate %proxycode
+syn keyword swigDirective %refobject %set_output %shared_ptr %std_comp_methods
+syn keyword swigDirective %std_nodefconst_type %typecheck %typemaps_string %unique_ptr %unrefobject %valuewrapper
+
+syn match swigVerbatimStartEnd "%\({\|}\)"
+syn region swigVerbatim  contained start="%{" end="%}"
+
+syn match swigUserDef "%[-_a-zA-Z0-9]\+"
```
jmarrec: Classify the keywords

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,94 @@
...
+hi def link swigPHP swigOtherLanguageSpecific
+hi def link swigPython swigOtherLanguageSpecific
+hi def link swigRuby swigOtherLanguageSpecific
+hi def link swigScilab swigOtherLanguageSpecific
+hi def link swigOtherLanguageSpecific Special
+
+hi def link swigInsertSection PreProc
+
+hi def link swigVerbatim Statement
+hi def link swigVerbatimStartEnd Statement
```
jmarrec: Try to assign semantically correct highligh groups per @dkearns 's comment https://github.com/vim/vim/pull/13540#discussion_r1398101676

This is mostly a shot in the dark (I'm a noob with respect to writing vim syntax files...), I'm open to suggestions!

src/testdir/test_filetype.vim
```diff
@@ -673,6 +673,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     swayconfig: ['/home/user/.sway/config', '/home/user/.config/sway/config', '/etc/sway/config', '/etc/xdg/sway/config'],
     swift: ['file.swift'],
     swiftgyb: ['file.swift.gyb'],
+    swig: ['file.swg', 'file.swig'],
```
jmarrec, line 676: Test for files recognized on extension only

src/testdir/test_filetype.vim
```diff
@@ -2340,4 +2341,30 @@ func Test_vba_file()
...
+  bwipe!
+
+  " ASM
+  call writefile(['; comment', ';'], 'Xfile.i', 'D')
+  split Xfile.i
+  call assert_equal('asm', &filetype)
+  bwipe!
+
+  filetype off
+endfunc
```
jmarrec: Test autodetect when "*.i" ext

runtime/syntax/swig.vim
dkearns: These look pretty good, thanks.

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,94 @@
...
+syn keyword swigDirective %attribute_ref %attributeref %attributestring %attributeval %auto_ptr %callback
+syn keyword swigDirective %delete_array %delobject %extend_smart_pointer %factory %fastdispatch %freefunc %immutable
+syn keyword swigDirective %implicit %implicitconv %interface %interface_custom %interface_impl %intrusive_ptr %intrusive_ptr_no_wrap
+syn keyword swigDirective %mutable %naturalvar %nocallback %nocopyctor %nodefaultctor %nodefaultdtor %nonaturalvar %nonspace
+syn keyword swigDirective %nspace %pointer_cast %pointer_class %pointer_functions %predicate %proxycode
+syn keyword swigDirective %refobject %set_output %shared_ptr %std_comp_methods
+syn keyword swigDirective %std_nodefconst_type %typecheck %typemaps_string %unique_ptr %unrefobject %valuewrapper
+
+syn match swigVerbatimStartEnd "%\({\|}\)"
+syn region swigVerbatim  contained start="%{" end="%}"
```
dkearns: This region doesn't do anything.  It's marked as `contained` but no other syntax group actually `contains` it.

You can probably just delete it and make do with the delimiters matched with swigVerbatimStartEnd?

That pattern could also be simplified to `"%[{}]".

runtime/ftplugin/swig.vim
dkearns, line 13: Needs the following line here:

```vim
let b:undo_ftplugin = "setlocal iskeyword<"
```

See `:help undo_ftplugin`

runtime/ftplugin/swig.vim
```diff
@@ -0,0 +1,11 @@
+" Vim filetype plugin file
+" Language:	SWIG
+" Maintainer:	Julien Marrec <julien.marrec 'at' gmail com>
+" Last Change:	2023 November 23
+
+" Only do this when not done yet for this buffer
+if exists("b:did_ftplugin")
+  finish
+endif
+
```
dkearns, line 11: Needs the following line here:

```vim
let b:did_ftplugin = 1
```

See `:help write-filetype-plugin`

runtime/filetype.vim
```diff
@@ -1688,7 +1688,7 @@ au BufNewFile,BufRead .procmail,.procmailrc	setf procmail
 " Progress or CWEB
 au BufNewFile,BufRead *.w			call dist#ft#FTprogress_cweb()
 
-" Progress or assembly
+" Progress or assembly or Swig
 au BufNewFile,BufRead *.i			call dist#ft#FTprogress_asm()
```
dkearns: Could you please rename this function to `FTi` while you're there.  It's the emerging scheme.  No, it's not great but they may as well be consistent and the current name is overly specific.

runtime/autoload/dist/ft.vim
```diff
@@ -590,23 +590,28 @@ export def FTprogress_cweb()
...
+    elseif line =~ ft_swig_keywords || line =~ ft_swig_verbatim_block_start || line =~ ft_swig_verbatim_block_end
```
dkearns: Can you ever match `ft_swig_verbatim_block_end` without first having seen `ft_swig_verbatim_block_start`?

src/testdir/test_filetype.vim
```diff
@@ -2340,4 +2341,30 @@ func Test_vba_file()
   filetype off
 endfunc
 
+func Test_swig_file()
+  filetype on
+
+  " Swig
+
```
dkearns: We need a test for the default filetype of `progress` here. Something like,

```vim
 " *.i defaults to progress
 call writefile(['looks like progress'], 'Xfile.i', 'D')
 split Xfile.i
 call assert_equal('progress', &filetype)
 bwipe!
```

dkearns (changes requested):
Thanks for the update.

jmarrec:
⁣Get BlueMail for Android ​

runtime/autoload/dist/ft.vim
jmarrec: I just removed the bit that looks for the closing one. Let me know if you want me to match both?

runtime/syntax/swig.vim
jmarrec: I initially wanted to do something with it, but this was too ambitious. Removed and simplified the regex

(What I would have wished for was for eg to apply python formating inside a `%pythoncode %{ ... %}` for eg, but this is too complicated for me).

runtime/filetype.vim
jmarrec: I also renamed the test.

src/testdir/test_filetype.vim
jmarrec: Done, and thanks for the snippet. Note that this should have been a pre-existing test :)

jmarrec:
> I noticed that `%#ifdef` is not highlighted. The `%` is not matched and its presence breaks the highlighting of `#ifdef`.
> This is just something I noticed that you might want to fix at some stage in the future.

Thanks for the catch! (are you a SWIG user or just very thorough?). 

I added that. And that made me also realize I needed to style the `$varname` special variables while I was at it.

Generally speaking, thanks a lot for the detailed review, and being pedagogic while at it, I really appreciate it!

I hope the changes I made addressed all the above review comments.

runtime/ftplugin/swig.vim
```diff
@@ -0,0 +1,13 @@
...
+" Last Change:	2023 November 23
+
+" Only do this when not done yet for this buffer
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+let b:undo_ftplugin = "setlocal iskeyword<"
+setlocal iskeyword+=%
```
jmarrec, line 13: Added the requested did/undo ftpplugin

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,99 @@
...
+syn keyword swigDirective %delete_array %delobject %extend_smart_pointer %factory %fastdispatch %freefunc %immutable
+syn keyword swigDirective %implicit %implicitconv %interface %interface_custom %interface_impl %intrusive_ptr %intrusive_ptr_no_wrap
+syn keyword swigDirective %mutable %naturalvar %nocallback %nocopyctor %nodefaultctor %nodefaultdtor %nonaturalvar %nonspace
+syn keyword swigDirective %nspace %pointer_cast %pointer_class %pointer_functions %predicate %proxycode
+syn keyword swigDirective %refobject %set_output %shared_ptr %std_comp_methods
+syn keyword swigDirective %std_nodefconst_type %typecheck %typemaps_string %unique_ptr %unrefobject %valuewrapper
+
+syn match swigVerbatimStartEnd "%[{}]"
+
+syn match swigUserDef "%\w\+"
```
jmarrec, line 64: Symplify regex

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,99 @@
...
+syn match swigVerbatimMacro "^\s*%#\w\+\%( .*\)\?$"
+
+" SWIG: typemap var and typemap macros (eg: $1, $*1_type, $&n_ltype, $self)
+syn match swigTypeMapVars "\$[*&_a-zA-Z0-9]\+"
```
jmarrec, line 68: Catch the "$specialvar" and the verbatim macros

image

runtime/filetype.vim
```diff
@@ -1688,8 +1688,8 @@ au BufNewFile,BufRead .procmail,.procmailrc	setf procmail
 " Progress or CWEB
 au BufNewFile,BufRead *.w			call dist#ft#FTprogress_cweb()
 
-" Progress or assembly
-au BufNewFile,BufRead *.i			call dist#ft#FTprogress_asm()
+" Progress or assembly or Swig
+au BufNewFile,BufRead *.i			call dist#ft#FTi()
```
jmarrec, line 1692: Renamed to FTi

src/testdir/test_filetype.vim
```diff
@@ -2340,4 +2341,34 @@ func Test_vba_file()
   filetype off
 endfunc
 
+func Test_i_file()
+  filetype on
```
jmarrec, line 2345: Renamed test too

src/testdir/test_filetype.vim
```diff
@@ -2340,4 +2341,34 @@ func Test_vba_file()
...
+  call writefile(['; comment', ';'], 'Xfile.i', 'D')
+  split Xfile.i
+  call assert_equal('asm', &filetype)
+  bwipe!
+
+  " *.i defaults to progress
+  call writefile(['looks like progress'], 'Xfile.i', 'D')
+  split Xfile.i
+  call assert_equal('progress', &filetype)
+  bwipe!
```
jmarrec, line 2369: Add the progress test

runtime/filetype.vim
```diff
@@ -2183,6 +2183,9 @@ au BufNewFile,BufRead *.swift.gyb		setf swiftgyb
 " Swift Intermediate Language or SILE
 au BufNewFile,BufRead *.sil			call dist#ft#FTsil()
 
+" Swig
+au BufNewFile,BufRead *.swg,*.swig set filetype=swig
```
dkearns: This has to use `setf swig` rather than `set filetype=swig`.  See `:help setfiletype`.

Sorry for missing that on the first pass, I assume.

dkearns (changes requested):

runtime/syntax/swig.vim
dkearns: I can take a look at adding that soon how about we just get this in for now?

runtime/filetype.vim
jmarrec:
```suggestion
au BufNewFile,BufRead *.swg,*.swig setf swig
```

dkearns (approved):
LGTM, thanks.

dkearns (approved):

chrisbra:
thanks all

jmarrec:
@dkearns thanks a lot for your help and patience getting this ready, much appreciated!

runtime/syntax/swig.vim
```diff
@@ -0,0 +1,99 @@
...
+syn keyword swigCstring %cstring_bounded_mutable %cstring_bounded_output %cstring_chunk_output %cstring_input_binary %cstring_mutable
+syn keyword swigCstring %cstring_output_allocate %cstring_output_allocate_size %cstring_output_maxsize %cstring_output_withsize
+syn keyword swigCWstring %cwstring_bounded_mutable %cwstring_bounded_output %cwstring_chunk_output %cwstring_input_binary %cwstring_mutable
+syn keyword swigCWstring %cwstring_output_allocate %cwstring_output_allocate_size %cwstring_output_maxsize %cwstring_output_withsize
+syn keyword swigCMalloc %malloc %calloc %realloc %free %sizeof %allocators
+
+syn keyword swigExceptionHandling %catches %raise %allowexception %exceptionclass %warn %warnfilter %exception
+syn keyword swigContract %contract %aggregate_check
+
+syn keyword swigDirective %addmethods %array_class %array_functions %attribute %attribute2 %attribute2ref
```
ojwb, line 53: Sorry for not looking at this sooner, but I notice there are a few long obsolete and deprecated macros included which will be gone completely in the next SWIG release (4.2.0): `%addmethods`, `%attribute_ref`, `%name`.  Not a big problem but it seems more helpful not to include these.

There also seem to be some omissions too - e.g. I don't see `%copyctor`, `%defaultctor`, `%defaultdtor` (but the "no"-prefixed inverted versions **are** listed).  These three aren't well documented, but neither is e.g. `%set_output` which is in the list so it's not that only things listed in the SWIG docs are included.

And thanks for working on this - it'll be great to have decent vim highlighting for SWIG interface files (currently I just set them to highlight as C++ which is better than nothing but far from perfect).

mcepl: > Sorry for not looking at this sooner, but I notice there are a few long obsolete and deprecated macros included which will be gone completely in the next SWIG release (4.2.0): 

vim should be able to work with old SWIG files as well, shouldn't it?

Omissions are real issue.

ojwb: > vim should be able to work with old SWIG files as well, shouldn't it?

I think in these three cases it is more unhelpful than helpful to treat them as valid syntax.

`%addmethods` was renamed to `%extend` (because it can add member variables too) in SWIG 1.3.12 (June 2, 2002) and has emitted a deprecation warning since then.

`%name` has emitted a deprecation warning since SWIG 1.3.24 (December 14, 2004).  It has to appear right before the thing you want to rename so doesn't work if you want to get SWIG to parse a library's API headers, which is the typical way to use SWIG.  The replacement is `%rename` which can rename things declared in an API header.

You might stick with a deprecated directive while you want to stay compatible with releases before the replacement was added, but I can't see that anyone would have been ignoring a deprecation warning for 19+ years, and I'd argue it's more helpful if they don't highlight as valid directives to give a visual signal that this needs updating if you do have an ancient interface file.

`%attribute_ref` is perhaps less clear cut - it was deprecated in SWIG 1.3.32 (November 15, 2007) but without a deprecation warning.

The replacement is `%attributeref` (with no underscore, which is more consistent with other macro names; also the parameter order now matches the other attribute macros).  I'd still argue for not highlighting it as valid because the old deprecated name is an easy typo to make for the new one.

jmarrec: @ojwb I realized some of these were deprecated, but I haven't tried to look at all of these.

I couldn't find a definitive, easy to consume, source for the accepted keywords. So basically what I did was using python regexes to parse through the entire swig/swig repo and trying to derive the keywords, a little bit of manual cleaning after that, and trying to quickly see if some of them were language specific to classify them a bit more. 

If you want to adjust the vim syntax file with your intricate knowledge of swig internals that'd be awesome, and I'm happy to review.

jmarrec: I agree that if something has been deprecated for so many years, it's probably best to not highlight them.

ojwb: Sadly SWIG's docs are patchy in places.  We now require docs and tests for new features, but in the past that wasn't always the case and it's a bit of a daunting task to try to fill in missing docs.  I've been involved in SWIG dev since ~2005 and I've little idea about some of these features and sometimes all there is to go on is a brief mention in CHANGES, or maybe nothing at all.

The next release will pretty much remove everything marked as deprecated more than about 10 years ago which at least reduces the number of such things.

I can't really think of a good way to generate such a list.  It doesn't help that some of these are built in features handle by SWIG's parser whereas others are defined as macros in the library (the user mostly doesn't need to know or care about the difference but it makes generating a list of them more challenging).

I'll certainly feedback improvements I spot.

PR #13563, opened 2023-11-23
Problem with initializing the length of range() lists

### LINKED ISSUE #13557: Empty list returned by range() with stride segfaults upon access
### Steps to reproduce

With `vim --clean`:

```vim
:echo get(range(1, 0, 2), 0)
```

### Expected behaviour

Expecting to see the default `get()` return value of 0.

The requirements to reproduce this issue seem to be that the range is empty, and the stride is greater than 1.

Workaround is to pass the range to `copy()`.

Bonus: here's a less fatal variant that is exclusive to Vim9 Script:

```vim
def RangeTest()
  const range = range(1, 0, 2)
  for i in range
    echo i
  endfor
enddef

call RangeTest()
```

This script should produce no output, but instead, it echoes `1`.

### Version of Vim

[8.2.0159](https://github.com/vim/vim/commit/50985eb1f0bd3c73ce727f9bbd66c839c92ef0da) and newer (bisected)

### Environment

Reproduced on Ubuntu 22.04

### Logs and stack traces

_No response_

chrisbra:
range() may cause a wrong calculation of list length, which may later then cause a segfault in list_find().  This is usually not a problem, because range_list_materialize() resets the length, when it materializes the list.

In addition, in list_find() when the length of the range was wrongly initialized, it may seem to be valid, so the check for list index out-of-bounds will not be true, because it is called before the list is actually materialized. And so we may eventually try to access a null pointer, causing a segfault.

So this patch does 2 things:

- In f_range(), when we know that the list should be empty, explicitly set the list->lv_len value to zero. This should fix the underlying issue properly. However,

- as a safety measure, let's check that the requested index is not out of range, after the list has been materialized one more time and return NULL in case it suddenly is.

fixes: #13557


chrisbra:
hm, IIRC I tried a similar fix and still had some test failures, so I made it a bit more strict

tpope:
Ah, I forgot to take a negative stride into account. This should be the correct fix:

```patch
diff --git i/src/evalfunc.c w/src/evalfunc.c
index 7f7914eca..50c13c924 100644
--- i/src/evalfunc.c
+++ w/src/evalfunc.c
@@ -8646,7 +8646,11 @@ f_range(typval_T *argvars, typval_T *rettv)
     list->lv_u.nonmat.lv_start = start;
     list->lv_u.nonmat.lv_end = end;
     list->lv_u.nonmat.lv_stride = stride;
-    list->lv_len = (end - start) / stride + 1;
+    if (stride > 0 ? end < start : end > start)
+	list->lv_len = 0;
+    } else {
+	list->lv_len = (end - start) / stride + 1;
+    }
 }
 
 /*
```

This `stride > 0` ternary matches the one for `e_start_past_end` a few lines up, but without the `1` offsets.

chrisbra:
thanks, looks good. I'll amend :)

PR #13567, opened 2023-11-24
Update translation makefiles after 9.0.2127

k-takata:
* Change how to check `%LANGUAGE%`.
  Check it only when needed.
* Add double quotes to where `GETTEXT_PATH` is used.
  Before 9.0.2127, this worked: `nmake -f Make_mvc.mak GETTEXT_PATH="\"C:\Program Files\Git\usr\bin\""` (which was a bit tricky.)
  9.0.2127 broke this and a syntax error occurred.
  This doesn't work either in 9.0.2127: `nmake -f Make_mvc.mak GETTEXT_PATH="C:\Program Files\Git\usr\bin"`
  With this PR, this works: `nmake -f Make_mvc.mak GETTEXT_PATH="C:\Program Files\Git\usr\bin"`
* Better error report for the `check` target.
  Show the line number of the error. (Imported from vim-jp/lang-ja.)

### DIFF
--- a/src/po/Make_mvc.mak
+++ b/src/po/Make_mvc.mak
@@ -526,7 +526,7 @@ update-po: $(MOFILES:.mo=)
 
 # Don't add a dependency here, we only want to update the .po files manually
 $(LANGUAGES):
-	@$(MAKE) -nologo -f Make_mvc.mak $(PACKAGE).pot GETTEXT_PATH=$(GETTEXT_PATH)
+	@$(MAKE) -nologo -f Make_mvc.mak $(PACKAGE).pot GETTEXT_PATH="$(GETTEXT_PATH)"
 	$(CP) $@.po $@.po.orig
 	$(MV) $@.po $@.po.old
 	$(MSGMERGE) $@.po.old $(PACKAGE).pot -o $@.po
--- a/src/po/Make_mvc.mak
+++ b/src/po/Make_mvc.mak
@@ -10,10 +10,6 @@
 !ERROR The program "PowerShell" version 3.0 or higher is required to work
 !ENDIF
 
-!IFNDEF LANGUAGE
-!ERROR Set the environment variable %LANGUAGE%. See README_mvc.txt
-!ENDIF
-
 # get LANGUAGES, MOFILES, MOCONVERTED and others
 !INCLUDE Make_all.mak
 
@@ -28,9 +24,9 @@ VIM = ..\vim
 # Correct the following line for the directory where gettext et al is installed
 GETTEXT_PATH = D:\Programs\GetText\bin
 
-MSGFMT = $(GETTEXT_PATH)\msgfmt -v
-XGETTEXT = $(GETTEXT_PATH)\xgettext
-MSGMERGE = $(GETTEXT_PATH)\msgmerge
+MSGFMT = "$(GETTEXT_PATH)\msgfmt" -v
+XGETTEXT = "$(GETTEXT_PATH)\xgettext"
+MSGMERGE = "$(GETTEXT_PATH)\msgmerge"
 
 # In case some package like GnuWin32, UnixUtils, gettext
 # or something similar is installed on the system.
@@ -68,7 +64,8 @@ INSTALLDIR = $(VIMRUNTIME)\lang\$(LANGUAGE)\LC_MESSAGES
 all: $(MOFILES) $(MOCONVERTED)
 
 .po.ck:
-	$(VIM) -u NONE -e -X -S check.vim -c "if error == 0 | q | endif" -c cq $<
+	$(VIM) -u NONE --noplugins -e -s -X --cmd "set enc=utf-8" -S check.vim \
+		-c "if error == 0 | q | else | num 2 | cq | endif" $<
 	$(TOUCH)
 
 check: $(CHECKFILES)
@@ -77,6 +74,9 @@ checkclean:
 	$(RM) *.ck
 
 converted: $(MOCONVERTED)
+
+checklanguage:
+	@if "$(LANGUAGE)"=="" (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
 
 nl.po:
 	@( echo \# > nl.po )
@@ -497,7 +497,7 @@ PO_INPUTLIST = \
 files: $(PO_INPUTLIST)
 	$(LS) $(LSFLAGS) $(PO_INPUTLIST) > .\files
 
-first_time: files
+first_time: checklanguage files
 	$(VIM) -u NONE --not-a-term -S tojavascript.vim $(LANGUAGE).po \
 		$(PO_VIM_INPUTLIST)
 	set OLD_PO_FILE_INPUT=yes
@@ -532,7 +532,7 @@ $(LANGUAGES):
 	$(MSGMERGE) $@.po.old $(PACKAGE).pot -o $@.po
 	$(RM) $@.po.old
 
-install: $(LANGUAGE).mo
+install: checklanguage $(LANGUAGE).mo
 	if not exist $(INSTALLDIR) $(MKD) $(INSTALLDIR)
 	$(CP) $(LANGUAGE).mo $(INSTALLDIR)\$(PACKAGE).mo
 
@@ -542,7 +542,7 @@ install-all: all
 	for %%l in ($(LANGUAGES)) do @$(CP) %%l.mo \
 		$(VIMRUNTIME)\lang\%%l\LC_MESSAGES\$(PACKAGE).mo
 
-cleanup-po: $(LANGUAGE).po
+cleanup-po: checklanguage $(LANGUAGE).po
 	$(VIM) -u NONE -e -X -S cleanup.vim -c wq $**
 
 cleanup-po-all: $(POFILES)
--- a/src/po/Make_mvc.mak
+++ b/src/po/Make_mvc.mak
@@ -43,9 +43,9 @@ ICONV="$(GETTEXT_PATH)\iconv.exe"
 # If the "touch" program is installed on the system, but it is not registered
 # in the %PATH% environment variable, then specify the full path to this file.
 !IF EXIST ("touch.exe")
-TOUCH = touch.exe $@
+TOUCH_TARGET = touch.exe $@
 !ELSE
-TOUCH = @if exist $@ ( copy /b $@+,, ) else ( type nul >$@ )
+TOUCH_TARGET = @if exist $@ ( copy /b $@+,, ) else ( type nul >$@ )
 !ENDIF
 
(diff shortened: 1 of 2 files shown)


k-takata:
Also renamed the variable `TOUCH` to `TOUCH_TARGET`, which was confusing.
I expected `$(TOUCH)` to work the same as the touch command. However, its definition included `$@`, therefore it worked differently.

src/po/Make_mvc.mak
```diff
@@ -47,9 +43,9 @@ ICONV="$(GETTEXT_PATH)\iconv.exe"
 # If the "touch" program is installed on the system, but it is not registered
 # in the %PATH% environment variable, then specify the full path to this file.
 !IF EXIST ("touch.exe")
-TOUCH = touch.exe $@
+TOUCH_TARGET = touch.exe $@
 !ELSE
-TOUCH = @if exist $@ ( copy /b $@+,, ) else ( type nul >$@ )
+TOUCH_TARGET = @if exist $@ ( copy /b $@+,, ) else ( type nul >$@ )
 !ENDIF
```
RestorerZ, line 50: To make it work for both target files and dependencies let's change it this way:
```
!IF EXIST ("D:\Programs\Git\usr\bin\touch.exe")
TOUCH = "D:\Programs\Git\usr\bin\touch.exe" %1
!ELSE
TOUCH = if exist %1 ( copy /b %1+,, ) else ( type nul >%1 )
!ENDIF
```

Then the call to it needs to be changed as well. In this case, it will be called using the "inline file" mechanism.
That is
```
target : dependence
       <<touch.bat {$@ | $?}
         $(TOUCH)
<<
```

src/po/Make_mvc.mak
```diff
@@ -68,8 +64,9 @@ INSTALLDIR = $(VIMRUNTIME)\lang\$(LANGUAGE)\LC_MESSAGES
 all: $(MOFILES) $(MOCONVERTED)
 
 .po.ck:
-	$(VIM) -u NONE -e -X -S check.vim -c "if error == 0 | q | endif" -c cq $<
-	$(TOUCH)
+	$(VIM) -u NONE --noplugins -e -s -X --cmd "set enc=utf-8" -S check.vim \
+		-c "if error == 0 | q | else | num 2 | cq | endif" $<
+	$(TOUCH_TARGET)
```
RestorerZ, line 71: Call $(TOUCH) which will work for both target files and dependencies.
```
.po.ck:
	$(VIM) -u NONE --noplugins -e -s -X --cmd "set enc=utf-8" -S check.vim \
		-c "if error == 0 | q | else | num 2 | cq | endif" $<
	<<touch.bat $@
          $(TOUCH)
<<
```

src/po/Make_mvc.mak
```diff
@@ -28,9 +24,9 @@ VIM = ..\vim
 # Correct the following line for the directory where gettext et al is installed
 GETTEXT_PATH = D:\Programs\GetText\bin
 
-MSGFMT = $(GETTEXT_PATH)\msgfmt -v
-XGETTEXT = $(GETTEXT_PATH)\xgettext
-MSGMERGE = $(GETTEXT_PATH)\msgmerge
+MSGFMT = "$(GETTEXT_PATH)\msgfmt" -v
+XGETTEXT = "$(GETTEXT_PATH)\xgettext"
+MSGMERGE = "$(GETTEXT_PATH)\msgmerge"
```
RestorerZ, line 31: That's weird. I checked now, no changes have been made to the "gettext" path definition.
Maybe the formatting of the lines in the command call was affected, but it shouldn't have been. That's weird. I'll check.

src/po/Make_mvc.mak
```diff
@@ -78,6 +75,9 @@ checkclean:
 
 converted: $(MOCONVERTED)
 
+checklanguage:
+	@if "$(LANGUAGE)"=="" (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
+
```
RestorerZ, line 81: it's a lot better than it was.

src/po/Make_mvc.mak
RestorerZ: As you can see, the first error is thrown at the "iconv.exe" file availability check line. And you use escape spaces in the "gettext" path when calling nmake.
```
E:\Project\vim\en-ru\po\tst\tst2>E:\Project\vim\tools\nmake.exe -f Make_mvc.mak GETTEXT_PATH="""D:\Program Files\GetText\bin"""

Microsoft (R) Program Maintenance Utility Version 14.27.29111.0
Copyright (C) Microsoft Corporation.  All rights reserved.

Make_mvc.mak(39) : fatal error U1023: syntax error in expression
Stop.
```
Well, this is a well-known and old DOS and Windows error from the time when FAT32 support was introduced (and maybe even earlier).

Because of my old habit and following the documentation recommendations, I enclose paths with spaces and other special characters in quotes.

> Long Filenames
> You can use long filenames if they are supported by your file system. However,
> you must enclose the name in double quotation marks, as in the following
> dependency line:
> all : "VeryLongFileName.exe"

> EXIST (path)
> Evaluates to true if path exists. EXIST can be used in complex expressions
> using binary logical operators. If path contains spaces (allowed in some file
> systems), enclose it in double quotation marks.

And you use quotation marks. That is, we get double escaping. That's why the error occurs.
Yes, I should have foreseen this and commented on it and changed the code in the makefile.
Well, it should work fine now. Thank you for fixing it!    

       
 
  By the way, instead of using GETTEXT_PATH="\\"C:\Program Files\Git\usr\bin\\"", you can use GETTEXT_PATH="""C:\Program Files\Git\usr\bin\""". It probably looks weird too, but it works the same way.

src/po/Make_mvc.mak
```diff
@@ -28,9 +24,9 @@ VIM = ..\vim
 # Correct the following line for the directory where gettext et al is installed
 GETTEXT_PATH = D:\Programs\GetText\bin
```
RestorerZ, line 26: @k-takata  , would you be able to add the following comment above this line?
"`# Please do not put the path in quotes.`"

src/po/Make_mvc.mak
RestorerZ: Since we are checking the system environment variable, it seems to me that this code should be like this:
```
checklanguage:
	@if NOT DEFINED LANGUAGE (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
```

src/po/Make_mvc.mak
```diff
@@ -542,7 +542,7 @@ install-all: all
 	for %%l in ($(LANGUAGES)) do @$(CP) %%l.mo \
 		$(VIMRUNTIME)\lang\%%l\LC_MESSAGES\$(PACKAGE).mo
 
-cleanup-po: $(LANGUAGE).po
+cleanup-po: checklanguage $(LANGUAGE).po
 	$(VIM) -u NONE -e -X -S cleanup.vim -c wq $**
```
RestorerZ: When processing this rule, an error occurs:
```
"checklanguage" 0 lines, 0 bytes
"checklanguage" 0 lines, 0 bytes written
Error detected while processing command line:
E173: 1 more file to edit
Entering Ex mode.  Type "visual" to go to Normal mode.
:args
[checklanguage] ru.po
:
```
As you can see, the "checklanguage" rule is passed as a dependency file.
Two variants have been tested so far:
First variant, remove "checklanguage" from dependencies and write the environment variable check in the command block.
```
cleanup-po: $(LANGUAGE).po
	if NOT DEFINED LANGUAGE (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
	$(VIM) -u NONE -e -X -S cleanup.vim -c wq $**
```
The second variant is to remove the "$**" macro from the Vim run argument, and directly specify the dependency as a run argument.
```
cleanup-po: checklanguage $(LANGUAGE).po
	$(VIM) -u NONE -e -X -S cleanup.vim -c wq $(LANGUAGE).po
```

src/po/Make_mvc.mak
RestorerZ: @k-takata , Please enclose the $(GETTEXT_PATH) macro in quotes to avoid errors when updating PO files.
```
# Don't add a dependency here, we only want to update the .po files manually
$(LANGUAGES):
	@$(MAKE) -nologo -f Make_mvc.mak $(PACKAGE).pot GETTEXT_PATH="$(GETTEXT_PATH)"
	$(CP) $@.po $@.po.orig
	$(MV) $@.po $@.po.old
	$(MSGMERGE) $@.po.old $(PACKAGE).pot -o $@.po
	$(RM) $@.po.old
```

src/po/Make_mvc.mak
k-takata: It doesn't work when `LANGUAGE` is set from the command line.
E.g. `nmake -f Make_mvc LANGUAGE=xy`

RestorerZ: When %LANGUAGE% is cleared:
```
E:\Project\vim\en-ru\po\tst\tst2>echo %LANGUAGE%
%LANGUAGE%

E:\Project\vim\en-ru\po\tst\tst2>E:\Project\vim\tools\nmake.exe -f Make_mvc.mak install

Microsoft (R) Program Maintenance Utility Version 14.27.29111.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        if NOT DEFINED LANGUAGE (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
Set the environment variable %LANGUAGE%. See README_mvc.txt.
NMAKE : fatal error U1077: 'if' : return code '0x1'
Stop.
```
When setting the %LANGUAGE% variable from the command line:
```
E:\Project\vim\en-ru\po\tst\tst2>E:\Project\vim\tools\nmake.exe -f Make_mvc.mak LANGUAGE=ru install

Microsoft (R) Program Maintenance Utility Version 14.27.29111.0
Copyright (C) Microsoft Corporation.  All rights reserved.

        if NOT DEFINED LANGUAGE (echo Set the environment variable ^%LANGUAGE^%. See README_mvc.txt. && exit 1)
        set OLD_PO_FILE_INPUT=yes
        "D:\Program Files\GetText\bin\msgfmt" -v -o ru.mo ru.po
2943 переведённых сообщения, 4 неточных перевода.
        if not exist ..\..\runtime\lang\ru\LC_MESSAGES mkdir ..\..\runtime\lang\ru\LC_MESSAGES
        copy /y ru.mo ..\..\runtime\lang\ru\LC_MESSAGES\vim.mo
Скопировано файлов:         1.
```
At least on my Windows NT 6.1 it works. Although I admit that it may not work on other systems or configurations. I'll have to check it out.

chrisbra:
thanks!

PR #13569, opened 2023-11-24
feat: commad-line overstrike detection with mode()

Sam-programs:
detecting overstrike by changing the command line causes flickering,
and i think adding a way to detect overstrike will allow command line mappings to be more consistent.


runtime/doc/builtin.txt
```diff
@@ -6413,8 +6413,11 @@ mode([expr])	Return a string that indicates the current mode.
 		   Rvx	    Virtual Replace mode |i_CTRL-X| completion
 		   c	    Command-line editing
 		   ct	    Command-line editing via Terminal-Job mode
+		   cr	    Command-line while in overstrike mode |c_<Insert>|
 		   cv	    Vim Ex mode |gQ|
+		   cvr	    Vim Ex while in overstrike mode |c_<Insert>|
 		   ce	    Normal Ex mode |Q|
+		   cer	    Normal Ex while in overstrike mode |c_<Insert>|
```
zeertzjq: This is likely not possible, as normal Ex mode doesn't invoke cmdline mode.

Sam-programs: oh thanks

Sam-programs:
the ci is failing because i forgot to move the defined checks mb.

src/misc1.c
```diff
@@ -720,6 +720,8 @@ get_mode(char_u *buf)
 	    buf[i++] = 'v';
 	else if (exmode_active == EXMODE_NORMAL)
 	    buf[i++] = 'e';
+	if (cmdline_overstrike() == TRUE)
```
zeertzjq: When executing commands in Ex mode (inside `do_cmdline()` but outside `getexline()`), Cmdline mode is not active and the result of `cmdline_overstrike()` is not valid, so a check for Cmdline mode is needed here.
```suggestion
	if ((State & MODE_CMDLINE) && cmdline_overstrike())
```

src/misc1.c
```diff
@@ -720,8 +720,11 @@ get_mode(char_u *buf)
 	    buf[i++] = 'v';
 	else if (exmode_active == EXMODE_NORMAL)
 	    buf[i++] = 'e';
+# if defined(MCH_CURSOR_SHAPE) || defined(FEAT_GUI) \
+	|| defined(FEAT_MOUSESHAPE) || defined(PROTO)
```
zeertzjq: I think this change is better:
```diff
diff --git a/src/ex_getln.c b/src/ex_getln.c
index 8f0be5208..057c832ea 100644
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -3169,8 +3169,6 @@ redraw:
     return (char_u *)line_ga.ga_data;
 }
 
-# if defined(MCH_CURSOR_SHAPE) || defined(FEAT_GUI) \
-	|| defined(FEAT_MOUSESHAPE) || defined(PROTO)
 /*
  * Return TRUE if ccline.overstrike is on.
  */
@@ -3180,6 +3178,8 @@ cmdline_overstrike(void)
     return ccline.overstrike;
 }
 
+# if defined(MCH_CURSOR_SHAPE) || defined(FEAT_GUI) \
+	|| defined(FEAT_MOUSESHAPE) || defined(PROTO)
 /*
  * Return TRUE if the cursor is at the end of the cmdline.
  */
```

src/misc1.c
Sam-programs: tbh i assumed ccline is zero initalized, i thought it wouldn't matter since the check would always fail if that was the case.
sorry am not used to going into new c code bases.

chrisbra:
can you add a test please? I think a screen dump test using `mode()` in a statusline should be able to test it properly

zeertzjq:
For cmdline mode there are some tests using `feedkeys()` in `Test_mode()` in `test_functions.vim` which should be easy to extend.

Sam-programs:
thanks zeertzjq

src/testdir/test_functions.vim
```diff
@@ -951,6 +951,11 @@ func Test_mode()
     call assert_equal('nt', mode(1))
     call feedkeys(":echo \<C-R>=Save_mode()\<C-U>\<CR>", 'xt')
     call assert_equal("c-c", g:current_modes)
+    call feedkeys(":\<insert>\<C-r>=Save_mode()\<CR>",'xt')
```
zeertzjq: These shouldn't be inside this `if has('terminal')` block. I think they should be moved just below line 918.

Sam-programs: i am sorry for making so many mistakes.

src/testdir/test_functions.vim
```diff
@@ -951,8 +951,13 @@ func Test_mode()
     call assert_equal('nt', mode(1))
     call feedkeys(":echo \<C-R>=Save_mode()\<C-U>\<CR>", 'xt')
     call assert_equal("c-c", g:current_modes)
+
     call feedkeys("aexit\<CR>", 'xt')
   endif
+  call feedkeys(":\<insert>\<C-r>=Save_mode()\<CR>",'xt')
```
zeertzjq: Putting these just after exiting terminal won't work, because exiting terminal needs a delay. Like I previously said, I'd suggest putting these lines just below line 918, grouped with other cmdline mode test cases.

src/testdir/test_functions.vim
```diff
@@ -951,8 +951,13 @@ func Test_mode()
     call assert_equal('nt', mode(1))
     call feedkeys(":echo \<C-R>=Save_mode()\<C-U>\<CR>", 'xt')
     call assert_equal("c-c", g:current_modes)
+
```
zeertzjq: Unnecessary blank line

zeertzjq (approved):

Sam-programs:
> I think a screen dump test using mode() in a statusline should be able to test it properly

@chrisbra may i ask, why u wanted it to be tested that way?

chrisbra:
> @chrisbra may i ask, why u wanted it to be tested that way?

It's an easy way to make sure that the mode() is actually correctly shown.

Sam-programs:
it won't be without an explicit `redrawstatus` tho.

chrisbra:
It doesn't seem to work as expected:
grafik

That's the reason I wanted to have a screendump test :) 

I guess Vim isn't redrawing the statusline.

chrisbra:
> it won't be without an explicit redrawstatus tho.

So shouldn't that be made implicitly then?

Sam-programs:
I am not sure how to make the command line update the statusline.
The CmdlineChanged autocmd isn't working.

chrisbra:
I think this one would do it:
````patch
diff --git a/src/ex_getln.c b/src/ex_getln.c
index 0aae45fc4..ba87f8d3c 100644
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -2050,6 +2050,8 @@ getcmdline_int(
 #ifdef CURSOR_SHAPE
                ui_cursor_shape();      // may show different cursor shape
 #endif
+               status_redraw_curbuf();
+               redraw_statuslines();
                goto cmdline_not_changed;

        case Ctrl_HAT:
```

Sam-programs:
yea thanks!

Sam-programs:
why can i never get it right on the first time.

zeertzjq:
Oh, this will also need to trigger `ModeChanged`.

Sam-programs:
Is there anything else that happens when the mode changes in vim?

chrisbra:
wait a second, enabling overstrike does not change the mode, right? So it does not need to trigger ModeChanged.

zeertzjq:
There is a sentence `This will be triggered on every minor mode change.`

chrisbra:
ah okay, yes then it is true. Can we get a test for that one as well please?

Sam-programs:
yes sir!

Sam-programs:
how do i record a dump with a specific size?

Sam-programs:
nvm.

Sam-programs:
why are they both the same dump ?
```vim
call delete('mode_c.dump')
call delete('mode_cr.dump')
let g:buf = term_start('vim', {'term_rows' : 12,'term_cols' : 25})
call term_sendkeys(g:buf,":set laststatus=2 statusline=%!mode(1)\<CR>")
call term_sendkeys(g:buf,":echo\<insert>")
call term_dumpwrite(g:buf,'mode_c.dump')
call term_sendkeys(g:buf,"\<insert>")
call term_dumpwrite(g:buf,'mode_cr.dump')
```
```
> +0&#ffffff0@24
@25
@25
@25
@25
@25
@25
@25
@25
@25
@25
@25
```

Sam-programs:
should i put `CheckRunVimInTerminal` in a try catch block?

src/testdir/test_functions.vim
```diff
@@ -940,6 +940,18 @@ func Test_mode()
   execute "normal! gR\<C-o>g@l\<Esc>"
   call assert_equal('n-niV', g:current_modes)
 
+  " Test statusline updates for overstike mode
+  CheckRunVimInTerminal
```
zeertzjq: It's better to either use an `if CanRunVimInTerminal()` block here or move this block to a separate function, so that the cleanup in this function doesn't get skipped.

PR #13585, opened 2023-11-26
translation(de): Updated German translation

chrisbra:


src/po/de.po
```diff
@@ -21,9 +21,11 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
+#: ../alloc.c:94 ../alloc.c:109
```
k-takata: It looks like you haven't run `:so cleanup.vim`.

chrisbra:
opps, you are right, let me fix this.

PR #13588, opened 2023-11-27
Add test for mode() result when executing commands in Ex mode

zeertzjq:
Problem:  No test for mode() result when executing commands in Ex mode.
Solution: Add some test cases and simplify several other test cases.
          Also add a few more test cases for ModeChanged.


chrisbra:
thanks!

PR #13589, opened 2023-11-28
Revise runtime/lang/Makefile

k-takata:
* Use predefined variables (`$@`, `$<`) instead of the actual file names.
* Adjust spacing.
* Regenerate converted menu files.


chrisbra:
thanks!

PR #13591, opened 2023-11-28
runtime(html): Update syntax file

dkearns:
Add missing search element and update ARIA attribute list.

Add a very basic test file to check all elements are matched.


chrisbra:
thanks!

chrisbra:
oh, I think the html syntax test fails in CI.

chrisbra:
I think it failes for the `htmlItalic` highlighting on a Mac, it seems both `<em>` and `<i>` highlighting produce slight differences in the expected dump file

chrisbra:
I cannot reproduce the failure, but this patch "filters" those out, I think:

```patch
diff --git a/runtime/syntax/testdir/dumps/html_02.vim b/runtime/syntax/testdir/dumps/html_02.vim
new file mode 100644
index 000000000..5615395c8
--- /dev/null
+++ b/runtime/syntax/testdir/dumps/html_02.vim
@@ -0,0 +1 @@
+8s_\.+\d#0000000&@2_.+4#0000000\&@2_
diff --git a/runtime/syntax/testdir/dumps/html_03.vim b/runtime/syntax/testdir/dumps/html_03.vim
new file mode 100644
index 000000000..5615395c8
--- /dev/null
+++ b/runtime/syntax/testdir/dumps/html_03.vim
@@ -0,0 +1 @@
+8s_\.+\d#0000000&@2_.+4#0000000\&@2_
```

PR #13593, opened 2023-11-28
[security]: fix use-after-free in win-enter

chrisbra:
win_goto() may stop visual mode, if it is active. However, this may in turn trigger the ModeChanged autocommand, which could free the wp pointer which just was valid until now and now has become stale pointing to now freed memory.

So before calling win_enter(), let's verify one more time, that the wp pointer still points to a valid window structure.


chrisbra:
this is now part of #13596 so closing it

PR #13594, opened 2023-11-29
Detect angular & mustache templates

ObserverOfTime:
I couldn't find any angular syntax plugin so I set the filetype to `html.angular` in order to keep the html syntax.

---

- https://angular.io/guide/architecture-components#template-syntax
- https://github.com/mustache/mustache#templates


chrisbra:
thanks!

nickspoons:
What is the value of setting the filetype of *.component.html files to "html.angular"? There are several problems with this.

1. The "component" part of an angular file name is only a convention, angular HTML files don't need to be called that so this won't always catch them.
2. "component" is a very generic term and so I'm sure there are non-angular files in the wild called something.component.html
3. `FileType html` autocmds are no longer triggered, making this a breaking change. LSP plugins which use the "html" filetype to detect angular files will now need to detect "html" for older versions of vim, and "html.angular" for newer versions.

chrisbra:
Oh, I wasn't aware of those points. It seems 3) is quite serious, so let me rollback that particular change. Thanks for letting me know

ObserverOfTime:
> What is the value of setting the filetype of *.component.html files to "html.angular"? There are several problems with this.

It's precisely to *help* LSP and other plugins detect angular files.

PR #13595, opened 2023-11-29
runtime(html): Fix syntax test failures

dkearns:
Rendered element content results in test failures on macOS and FreeBSD.

The included content for these elements was incidental to the primary
intent of the test to check element tag highlighting so set
g:html_no_rendering to disable content rendering.

FreeBSD fails for even an empty `<strike>` element.

See https://github.com/vim/vim/pull/13591 and https://github.com/vim/vim/pull/14215.


chrisbra:
I don't know, but perhaps the FreeBSD CI does not support the `strikethrough` attribute? I don't think we have tested that one before in a screen-dump test. Perhaps we should leave that one out (or skip the test on BSD and Mac) in this case?

zzzyxwvut:
[A similar problem](https://github.com/vim/vim/pull/14215#issuecomment-2002615268) was encountered for a Markdown syntax test  
file during the merge effort for #14215.  At the time, [a fix](https://github.com/vim/vim/pull/14215#issuecomment-2002625279)  
to do `set t_ZH= t_ZR=` before a screen dump is taken was  
suggested by Christopher.  The test passed with that setting  
and it was included in `testdir/runtime.vim`:

https://github.com/vim/vim/blob/9f26e5a9bcedb3caef26e9d77849ea37a3626bbf/runtime/syntax/testdir/runtest.vim#L163

and the test passes ever since.


Now, I have followed [my own advice](https://github.com/vim/vim/issues/14245#issuecomment-2159121628) and, currently, there are  
[no related failures](https://github.com/zzzyxwvut/vimvim/actions/runs/9488523351/attempts/2) for HTML tests.

Perhaps, the masking of `input/html_html` can be revisited.

chrisbra:
yes, we could try out, if it still breaks (hopefully not 🤞 )

chrisbra:
thanks!

PR #13597, opened 2023-11-29
runtime(cmake): sync runtime files with upstream

e-kwsm:


runtime/indent/cmake.vim
```diff
@@ -17,8 +17,6 @@ let b:did_indent = 1
 setlocal indentexpr=CMakeGetIndent(v:lnum)
 setlocal indentkeys+==ENDIF(,ENDFOREACH(,ENDMACRO(,ELSE(,ELSEIF(,ENDWHILE(
 
-let b:undo_indent = "setl inde< indk<"
-
```
dkearns: This should not be removed.

dkearns (changes requested):
Thanks for the update.

It looks like this is maintained and distributed as part of the CMake project.  If you're managing updates to the Vim distribution, rather than Dimitri, it might be a good idea to also list yourself as a maintainer in the header and add yourself to https://github.com/vim/vim/blob/master/.github/CODEOWNERS

@chrisbra might have some thoughts on how to manage this scenario.

chrisbra:
Thanks, I don't have anything more to add, you covered it all :)

e-kwsm:
They were
* https://gitlab.kitware.com/cmake/cmake/-/blob/44a2f3f3324a608062eb7b88072c7640f80f4a5c/Auxiliary/vim/indent/cmake.vim
* https://gitlab.kitware.com/cmake/cmake/-/blob/c1d787e4734ca107e8ae084b3832e441651a6266/Auxiliary/vim/syntax/cmake.vim

MR [cmake!9034](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9034).

e-kwsm:
The upstream is https://github.com/pboettch/vim-cmake-syntax.
Consequently this goes to pboettch/vim-cmake-syntax#33.

chrisbra:
so is this ready? Or is there anything missing?

runtime/indent/cmake.vim
e-kwsm: Thank you for your review.

e-kwsm:
Ready but pboettch/vim-cmake-syntax#33 is not merged.

runtime/indent/cmake.vim
```diff
@@ -3,9 +3,9 @@
 " Author:       Andy Cedilnik <andy.cedilnik@kitware.com>
 " Maintainer:   Dimitri Merejkowsky <d.merej@gmail.com>
 " Former Maintainer: Karthik Krishnan <karthik.krishnan@kitware.com>
-" Last Change:  2022 Apr 06
+" Last Change:  2022 Mar 22
```
chrisbra: Why is the Last Change header now before the old Last Change Header?

e-kwsm: This was a conflict with this and CMake repositories.

PR #13598, opened 2023-11-29
Add vimlong_T; used to detect int overflow.

errael:
Assumption is `sizeof(overflow_T) >= 2 * sizeof(int)`. This makes `overflow_T` suitable for detecting `int` overflow with simple comparisons and makes it easy to find locations with this kind of overflow checking.

@k-takata @chrisbra  I'm not sure of the best name or where in the code to put it; please advise. (or if it is even wanted).

A side effect is that it is easier for unsupported and/or older build systems to create `vim` binaries without changing `.c` files.

### DIFF
--- a/src/cindent.c
+++ b/src/cindent.c
@@ -1730,7 +1730,7 @@ parse_cino(buf_T *buf)
     char_u	*p;
     char_u	*l;
     char_u	*digits;
-    long long	n;
+    vimlong_T	n;
     int		divider;
     int		fraction = 0;
     int		sw;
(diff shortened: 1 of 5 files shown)


errael:
> the typname `overflow_T` is sounds wrong. how about using `vimlong_T` or something similar?

Done.

chrisbra:
thanks!

wilhelmtell:
Is the idea to typedef `long long` and use that for ints, with a clamp function? That would be more of a clamping and silencing of an overflow than an overflow detection, no?

PR #13600, opened 2023-11-30
Include several missing files in Filelist for packaging/distribution

psimonyi:
Files need to be listed in `Filelist` (which is included into the Makefile) in order to be included in the tar archive.  Fedora (at least) uses the archive from `make unixall` so anything not listed there will be excluded from their package.  See PR #13551 for a recent case; this is part of the follow-up.

This is part of a two-patch series: see PR #13601 which adds a CI job to check for such overlooked files.

These files were discovered to be missing from Filelist, and thus distribution tarballs:
- editorconfig plugin
- extra files for Rust support
- readme for Haiku OS builds

It seems like all of them ought to be included, but I don't know whether they might have been excluded deliberately.  Ping @gpanders for editorconfig, @lilyball for Rust/Cargo, and @bitigchi for Haiku.

Incidentally, the SRC_HAIKU variable actually appears to be unused in makefiles... Is it actually used for building on Haiku?

### DIFF
--- a/Filelist
+++ b/Filelist
@@ -757,6 +757,17 @@ RT_ALL =	\
 		runtime/pack/dist/opt/dvorak/dvorak/enable.vim \
 		runtime/pack/dist/opt/dvorak/dvorak/disable.vim \
 		runtime/pack/dist/opt/editexisting/plugin/editexisting.vim \
+		runtime/pack/dist/opt/editorconfig/.editorconfig \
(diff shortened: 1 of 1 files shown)


gpanders:
I was not involved in adding the EditorConfig plugin to Vim, so I cannot give you any guidance there. 

I did recently update the Rust runtime files, however. None of those files were intentionally omitted, so please include them.

chrisbra:
Thanks, makes sense.

PR #13611, opened 2023-12-03
Text properties causes wrong line wrapping to be drawn

zeertzjq:
Problem:  Text properties causes wrong line wrapping to be drawn.
Solution: Find the index of the last text property that inserts text.


chrisbra:
thanks!

PR #13615, opened 2023-12-03
Skip test terminal_resize2() in GH CI in gui mode

chrisbra:


chrisbra:
yes, this seems to be the same issue

PR #13618, opened 2023-12-03
runtime: add pacman.log syntax

rpigott:
pacman.log is a filetype common to Arch Linux and related distributions.

pacman.log records every pacman operation performed on a machine. It's a single line format that is simply 

> `[<TIMESTAMP>] [<CALLER>] <MSG>`

My goal with this syntax is to highlight the log messages that installed or removed packages differently, and to delineate each pacman transaction more clearly in the log, to improve legibility. I tried to choose sensible defaults for the highlight groups, otherwise whatever looked good to me.

I haven't created any vim syntax definitions before this one, so there may be oddities or errors. PTAL.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -450,6 +450,7 @@ runtime/syntax/nix.vim			@equill
 runtime/syntax/nroff.vim		@jmarshall
 runtime/syntax/nsis.vim			@k-takata
 runtime/syntax/openvpn.vim		@ObserverOfTime
+runtime/syntax/pacmanlog.vim		@rpigott
 runtime/syntax/pascal.vim		@dkearns
 runtime/syntax/pbtxt.vim		@lakshayg
 runtime/syntax/pdf.vim			@tpope
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1521,6 +1521,9 @@ au BufNewFile,BufRead *.hook
 	\   setf confini |
 	\ endif
 
+" Pacman log
+au BufNewFile,BufRead pacman.log			setf pacmanlog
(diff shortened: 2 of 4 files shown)


k-takata:
Could you also update `.github/CODEOWNERS`?

runtime/syntax/paclog.vim
```diff
@@ -0,0 +1,40 @@
+" Vim syntax file
+" Language: pacman.log
+" Maintainer: Ronan Pigott <ronan@rjp.ie>
```
dkearns: Could you please also add a Last Change header field here?

`Last Change: 2023 Dec 04`

dkearns (changes requested):
Is pac a common abbreviation for pacman?  If not it would be better to just use "pacmanlog" for the filetype.

The filetype names aren't always that consistent, and I realise you were probably just following the "deb*" filetypes as an example, but we've been trying to use more explicit names for new additions.

Could you also add a test for the new filetype [here](https://github.com/vim/vim/blob/0d0b3b19517c321b089d637919e88e49a07a3d85/src/testdir/test_filetype.vim#L513)?

rpigott:
Adding color was actually a little inspired by the `paclog` tool from pacutils, which mostly serves to add color to pacman.log.  I was looking at my log recently and decided it would be more convenient to just have the "paclog" colors in vim.

I think I agree that pacmanlog is a better syntax name though, I will change it.

ObserverOfTime:
[pacman-contrib](https://gitlab.archlinux.org/pacman/pacman-contrib) might also be interested in this.

chrisbra:
thanks, looks good.

PR #13623, opened 2023-12-04
OpenVMS X86_64 platform port with rebase

arpadffy:
This pull request contains the summary of the previous attempts after rebase.


errael (changes requested):
@arpadffy Wow, cool.

There is still
- https://github.com/vim/vim/pull/13553#issuecomment-1837209300
- https://github.com/vim/vim/pull/13553#discussion_r1412842241

from the previous PR. Could you either incorporate those suggested changes or explain why making no change is better or do something better? BTW, the `stdarg` change should fix the code coverage warning.

(I didn't check the PR (just glanced at it), and noted those earlier issues)

src/strings.c
```diff
@@ -2789,7 +2789,7 @@ skip_to_arg(
     {
 	// Reset ap to ap_start and skip arg_idx - 1 types
 	va_end(*ap);
-	va_copy(*ap, ap_start);
+	va_copy (*ap, ap_start);
```
k-takata: Please revert this change.

src/strings.c
```diff
@@ -2888,7 +2888,11 @@ vim_vsnprintf_typval(
     if (parse_fmt_types(&ap_types, &num_posarg, fmt, tvs) == FAIL)
 	return 0;
 
-    va_copy(ap, ap_start);
+#ifdef va_copy
+    va_copy (ap, ap_start);
+#else
+    ap = ap_start;
+#endif
```
k-takata: Please also revert this change.

src/vim.h
```diff
@@ -439,8 +439,11 @@ typedef unsigned int u8char_T;	// int is 32 bits or more
  * The vimlong_T has sizeof(vimlong_T) >= 2 * sizeof(int).
  * One use is simple handling of overflow in int calculations.
  */
+#ifdef HAVE_NO_LONG_LONG
+typedef long      vimlong_T;
+#else
 typedef long long vimlong_T;
-
+#endif
```
k-takata, line 447:
```suggestion
#endif

```
Please insert an empty line after `#endif` here.

src/vim.h
```diff
@@ -439,8 +439,11 @@ typedef unsigned int u8char_T;	// int is 32 bits or more
  * The vimlong_T has sizeof(vimlong_T) >= 2 * sizeof(int).
  * One use is simple handling of overflow in int calculations.
  */
+#ifdef HAVE_NO_LONG_LONG
```
errael:
```suggestion
#if defined(VMS) && defined(VAX)
// unsupported compiler
```

I scanned `vim.h` again. Using VMS/VAX looks consistent with the rest of the file. As you mentioned "VAX is the only one".

I agree with your comment (but not your conclusion)

> this clearly explains why we are doing this acrobatics with the types

The answer, particularly in this architectural file, is VAX/VMS. A further comment could be included here. Needing to go to some other file obscures the situation.

errael (changes requested):

src/strings.c
arpadffy: I am sorry. Now I am confused. I have removed this part because of @k-takata  and @errael 's suggestion in https://github.com/vim/vim/pull/13553#discussion_r1407314149 .
Why should we revert this change? 
I fully agree that the proposed solution using #ifndef va_copy... is a superior solution.

k-takata: There were two `#ifdef va_copy` blocks in strings.c.
You reverted only one block (L2792), but L2891-L2895 still remains.

src/strings.c
k-takata: You don't need to insert a space between `va_copy` and `(` here (L2792).

src/strings.c
arpadffy: Yes. Now I see.
I am sorry... I have missed the second block.
I have commited the change, please review.
Thank you

src/strings.c
k-takata:
```suggestion
	va_copy(*ap, ap_start);
```

src/strings.c
```diff
@@ -2888,7 +2888,7 @@ vim_vsnprintf_typval(
     if (parse_fmt_types(&ap_types, &num_posarg, fmt, tvs) == FAIL)
 	return 0;
 
-    va_copy(ap, ap_start);
+    va_copy (ap, ap_start);
```
k-takata:
```suggestion
    va_copy(ap, ap_start);
```

k-takata (approved):
Looks good.

chrisbra:
Thanks, I think this looks good now

PR #13624, opened 2023-12-05
'breakindent' is not drawn after diff filler lines

zeertzjq:
Problem:  'breakindent' is not drawn after diff filler lines.
Solution: Correct check for whether 'breakindent' should be drawn.


chrisbra:
wow, it has been like this for a very long time and nobody noticed...

zeertzjq:
Yes, and the cursor position still takes 'breakindent' into account, so it doesn't match drawn text.

PR #13625, opened 2023-12-05
Use 'any' instead of 'unknown' for List and Dict containers

### LINKED ISSUE #13582: Incorrect type reported for list<any> variable assigned an empty list
### Steps to reproduce

```vim
vim9script

var la: list<any> = []
echomsg typename(la)

var ln: list<number> = []
echomsg typename(ln)
```

Output: 

```
list<unknown>
list<number>
```

### Expected behaviour

Output: 

```
list<any>
list<number>
```

### Version of Vim

9.0.2130

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
When a container (List or Dict) item type cannot be determined, use "any" instead of "unknown".
Fixes #13582.

### DIFF
--- a/src/testdir/test_listdict.vim
+++ b/src/testdir/test_listdict.vim
@@ -516,7 +516,7 @@ func Test_dict_func_remove()
       var d = {1: 'a', 3: 'c'}
       call remove(d, [])
   END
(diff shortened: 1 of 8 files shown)


chrisbra:
thanks!

PR #13626, opened 2023-12-05
Docs: non-existent parameter in command

Roy-Orbison:
The variant with the `{count}` parameter is explained in the next item.


chrisbra:
thanks

PR #13628, opened 2023-12-05
Use size_t for errbuflen in string option functions

mrdubya:
This started as size_t to int warning on Windows in do_set_option_string().  Other option setting functions are using size_t for errbuflen so I changed the type in the argument list and then chased further compiler warnings into the string option functions in optionstr.c changing ints to size_t as I went.
I also changed the type of os_errbuflen in optset_T to size_t and updated related argument lists in optionstr.c, as well as updating set_string_option() to also use size_t for errbuflen for consistency with other string option setting functions.


chrisbra:
thanks, makes sense

PR #13630, opened 2023-12-05
runtime(doc): sort option-list alphabetically

tetzng:
Most of the options seem to be in alphabetical order, but a few were not, so I fixed them.


chrisbra:
yes, I think it should be sorted, thanks!

PR #13633, opened 2023-12-05
runtime(doc): sort options alphabetically

tetzng:
related to https://github.com/vim/vim/pull/13630#issuecomment-1841106280


chrisbra:
that was already included as of #13634 , sorry.

PR #13634, opened 2023-12-05
runtime(doc): Sort options.txt alphabetically

laburnumT:
Related to: vim/vim#13630

1. Sort section 3 (option-summary) of runtime/doc/options.txt in alphabetical order.
2. Remove double newlines in section 3 (option-summary).

I did not alphabetically sort ttimeout and ttimeoutlen because they seem to belong with timeout and timeoutlen respectively, but I would like some input on that decision.


chrisbra:
thanks. I include it. One issue I noticed:
before the section of 'cinwords', one needs to quit the previous example from `:h cinscopedecls`. I fixed it alreay.

PR #13637, opened 2023-12-06
Generate errors when class/typealias involved in assignment.

errael:
This PR prevents the use of a type, both user defined (class) and typealias,
on the RHS of an assignment.

There's several more cases to do, as discussed below, which can be done in
a follow on PR.

In addition there should be a PR/message for using a type on the LHS.

---

The following is copied from #13404.
This is the overall goal. NOTE: **only partially implemented with this PR**

Prevent use of type, both user defined (class) and typealias, as a value.
- from non-value items, e.g. `class` or `typealias`.
- of non-value items as function argument or return.
- to `dict`/`list`: `var l = [1, C]`, `var d = {x: C}`.

Here's an overview of the code change targets (not all done yet)
```
simple assignments
    scriptlevel assignment  ex_let_one in evalvars.c
    :def assignment         compile_assignment in vim9compile.c

Function related checks
    scriptlevel argument    get_func_arguments in userfunc.c
    argument, call in def   check_ufunc_arg_types in vim9execute.c
    return value            ISN_RETURN (but gives wrong line number)
    defer/funcref           check_func_args_from_type in vim9instr.c

dict/list checks
    script list         eval_list in list.c
    script dict         eval_dict in dict.c
    compile list/dict   get_member_type_from_stack in vim9instr.c
```

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -3557,10 +3557,16 @@ EXTERN char e_using_typealias_as_float[]
 	INIT(= N_("E1401: Using type alias \"%s\" as a Float"));
 EXTERN char e_using_typealias_as_string[]
 	INIT(= N_("E1402: Using type alias \"%s\" as a String"));
-EXTERN char e_using_typealias_as_value[]
+EXTERN char e_using_typealias_as_value_str[]
 	INIT(= N_("E1403: Type alias \"%s\" cannot be used as a value"));
 EXTERN char e_abstract_cannot_be_used_in_interface[]
 	INIT(= N_("E1404: Abstract cannot be used in an interface"));
+EXTERN char e_using_class_as_value_str[]
+	INIT(= N_("E1403: Class \"%s\" cannot be used as a value"));
+EXTERN char e_using_class_as_var_val[]
+	INIT(= N_("E1405: Cannot use a Class as a variable or value"));
+EXTERN char e_using_typealias_as_var_val[]
+	INIT(= N_("E1406: Cannot use a Typealias as a variable or value"));
 #endif
 // E1405 - E1499 unused (reserved for Vim9 class support)
 EXTERN char e_cannot_mix_positional_and_non_positional_str[]
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -3093,25 +3093,77 @@ def Test_closure_in_class()
   v9.CheckSourceSuccess(lines)
 enddef
 
-def Test_call_constructor_from_legacy()
+def Test_construct_object_from_legacy()
+  # Cannot directly invoke constructor from legacy
   var lines =<< trim END
     vim9script
 
-    var newCalled = 'false'
+    var newCalled = false
 
     class A
-      def new()
-        newCalled = 'true'
+      def new(arg: string)
+        newCalled = true
       enddef
     endclass
 
-    export def F(options = {}): any
-      return A
+    export def CreateA(...args: list<any>): A
+      return call(A.new, args)
     enddef
 
-    g:p = F()
-    legacy call p.new()
-    assert_equal('true', newCalled)
+    g:P = CreateA
+    legacy call g:P('some_arg')
+    assert_equal(true, newCalled)
+    unlet g:P
+  END
+  v9.CheckSourceSuccess(lines)
+
+  lines =<< trim END
+    vim9script
+
+    var newCalled = false
+
+    class A
+      static def CreateA(options = {}): any
+        return A.new()
+      enddef
+      def new()
+        newCalled = true
+      enddef
+    endclass
+
+    g:P = A.CreateA
+    legacy call g:P()
+    assert_equal(true, newCalled)
+    unlet g:P
+  END
+  v9.CheckSourceSuccess(lines)
+
+  # This also tests invoking "new()" with "call"
+  lines =<< trim END
+    vim9script
+
+    var createdObject: any
+
+    class A
+      this.val1: number
+      this.val2: number
+      static def CreateA(...args: list<any>): any
+        createdObject = call(A.new, args)
+        return createdObject
+      enddef
+    endclass
+
+    g:P = A.CreateA
+    legacy call g:P(3, 5)
+    assert_equal(3, createdObject.val1)
+    assert_equal(5, createdObject.val2)
+    legacy call g:P()
+    assert_equal(0, createdObject.val1)
+    assert_equal(0, createdObject.val2)
+    legacy call g:P(7)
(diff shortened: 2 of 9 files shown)


src/vim9type.c
```diff
@@ -1842,4 +1842,60 @@ f_typename(typval_T *argvars, typval_T *rettv)
...
+    else if (typ == VAR_TYPEALIAS)
+    {
+        emsg(_(e_using_typealias_as_var_val));
+        return FAIL;
+    }
+    return OK;
+}
+
+    int
+check_typval_is_value(typval_T *tv)
```
yegappan, line 1852: Can you add a comment before this and the next function?

src/vim9execute.c
```diff
@@ -3813,13 +3813,21 @@ exec_instructions(ectx_T *ectx)
 	    case ISN_STORE:
 		--ectx->ec_stack.ga_len;
 		tv = STACK_TV_VAR(iptr->isn_arg.number);
-		if (STACK_TV_BOT(0)->v_type == VAR_TYPEALIAS)
+		// DON'T THINK THIS EVER TRIGGERS, GETS CAUGHT IN COMPILATION?
```
yegappan: From the Codecov results, I do see the below error message is covered.  You can try commenting out this code and run the tests and see which test fails.

src/errors.h
```diff
@@ -3557,10 +3557,16 @@ EXTERN char e_using_typealias_as_float[]
...
+EXTERN char e_using_typealias_as_value_str[]
 	INIT(= N_("E1403: Type alias \"%s\" cannot be used as a value"));
 EXTERN char e_abstract_cannot_be_used_in_interface[]
 	INIT(= N_("E1404: Abstract cannot be used in an interface"));
+EXTERN char e_using_class_as_value_str[]
+	INIT(= N_("E1403: Class \"%s\" cannot be used as a value"));
+EXTERN char e_using_class_as_var_val[]
+	INIT(= N_("E1405: Cannot use a Class as a variable or value"));
+EXTERN char e_using_typealias_as_var_val[]
+	INIT(= N_("E1406: Cannot use a Typealias as a variable or value"));
```
yegappan, line 3569: Without the tests, it is hard to see which assignment cases will fail now.  What are the list of assignments that will after these changes?

yegappan (changes requested):

src/errors.h
errael: I'll fix the tests for what's done so far.

errael:
Note there are _two sides, the LHS and RHS, of an assignment. The use of a **type**, either user defined or alias, should be an error on either side. The earlier PR #13404, now closed, focused on catching the use of a type on the RHS. I'm planning to add the rest of the RHS and the catch any LHS usages.

The key point of this PR is to come up with a common error message about using a type as a value.  I'd like to throw out the `cannot use a \<type> as a string`, `cannot use a \<type> as a number`, ... and only have one message. **If this approach is not OK, I'd like to know soonest**.

yegappan:
> 
> The key point of this PR is to come up with a common error message about using a type as a value. I'd like to throw out the `cannot use a \<type> as a string`, `cannot use a \<type> as a number`, ... and only have one message. **If this approach is not OK, I'd like to know soonest**.

This approach sounds good to me.  It will be good to see some examples where the assignment or use will fail.

src/vim9execute.c
errael: Turns out this is catching
```
var x = eval("A")
```
in a function.

errael:
> It will be good to see some examples where the assignment or use will fail.

The PR currently covers RHS of an assignment. Function related checks and dict/list checks (as described in https://github.com/vim/vim/pull/13637#issue-2027660537) can be added in a follow on PR.

errael:
In the following test `var c = X` doesn't get the most descriptive error.
This is traced to
```
if (check_type_is_value(rhs_type) == FAIL)
    goto theend;
```
Which is called in `vim9compile::compile_assignment()`
The problem is that when rhs_type is an alias, there's
no way to get the name. Suggestions?


```
vim9script

class C
endclass
type X = number

var a = X       # E1403: Type alias "X" cannot be used as a value

def F()
    var b = C       # E1403: Class "C" cannot be used as a value
    var c = X       # E1406: Cannot use a Typealias as a variable or value
enddef
F()
```

src/testdir/test_vim9_class.vim
```diff
@@ -3093,25 +3093,78 @@ def Test_closure_in_class()
...
+    legacy call g:P('some_arg')
     assert_equal('true', newCalled)
+    unlet g:P
+  END
+  v9.CheckSourceSuccess(lines)
+
+  lines =<< trim END
+    vim9script
+
+    var newCalled = 'false'
```
yegappan: Can you use a Boolean variable instead of a string true/false?

src/vim9type.c
```diff
@@ -1842,4 +1842,66 @@ f_typename(typval_T *argvars, typval_T *rettv)
...
+	emsg(_(e_using_class_as_var_val));
+	return FAIL;
+    }
+    else if (typ == VAR_TYPEALIAS)
+    {
+        emsg(_(e_using_typealias_as_var_val));
+        return FAIL;
+    }
+    return OK;
+}
```
yegappan, line 1905: Are the above lines not covered by tests?  Codecov coverage for these new functions is low.

yegappan (changes requested):

src/testdir/test_vim9_class.vim
errael: OK; makes sense; DONE. Also changed the original test that I cloned from.

src/vim9type.c
errael: Right. Since there will be follow on PRs that use (might use) these methods. Fully implemented the methods so they handled any legal args.

In general, what's the right thing to do with something like
```
check_vartype_is_value(vartype_T typ)
```
Shouldn't all the cases be considered, even if the function is never is never called with some particular type?

yegappan (approved):

src/vim9type.c
yegappan: > Shouldn't all the cases be considered, even if the function is never is never called with some particular type?

We can use a default return value for most of the types where this check is not applicable.

errael: @yegappan If you have the opportunity, could you review #13644, after that's committed I can do the follow on PR with function related checks and dict/list checks for a type used as a value.

chrisbra:
Thanks, makes sense!

errael:
I just noticed that this PR has some E#### overlap around E1404. I'll correct it in #13644.

PR #13642, opened 2023-12-06
fix: c/cpp ft detection maybe wrong if fic set

### LINKED ISSUE #13641: why some vim src recogized as cpp?
### Steps to reproduce

i am curious to ask why e.g 'src/vim.h' (its ft) is 'cpp'?
as i knew vim is running on C90 or something, not C++, right?
// not sure since when it was recognized like that. :smile:
// what's the benefit to recoginze '.h' as 'cpp' vs not 'c'?

### Expected behaviour

should be 'c'

### Version of Vim

v9.0.2153

### Environment

linux

### Logs and stack traces

_No response_

Shane-XB-Qian:
closes #13641

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1,7 +1,7 @@
 " Vim support file to detect file types
 "
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
-" Last Change:	2023 Aug 10
+" Last Change:	2023 Dec 06
 " Former Maintainer:	Bram Moolenaar <Bram@vim.org>
 
 " Listen very carefully, I will say this only once
@@ -348,7 +348,7 @@ au BufNewFile,BufRead *.cypher			setf cypher
 " C++
 au BufNewFile,BufRead *.cxx,*.c++,*.hh,*.hxx,*.hpp,*.ipp,*.moc,*.tcc,*.inl setf cpp
 if has("fname_case")
-  au BufNewFile,BufRead *.C,*.H setf cpp
+	au BufNewFile,BufRead *.C,*.H if !&fileignorecase | setf cpp | endif
 endif
 
 " C++ 20 modules (clang)


Shane-XB-Qian:
please check https://github.com/vim/vim/issues/13641#issuecomment-1844339486 here,
even set g:c_syntax_for_h, '.h' maybe recognized as 'cpp', since that 'au' on '.H' (upppercase),
because user may have set 'fic' too.

chrisbra:
and if the user doesn't have fileignorecase set, we won't set the filetype? So we are left without filetype and highlighting. That cannot be a solution

zeertzjq:
The filetype will be `c` when 'fileignorecase' is set.

Shane-XB-Qian:
> and if the user doesn't have fileignorecase set, we won't set the filetype?

if user did Not set 'fileignorecase', then everything is same like all it was working like;

if user did --- set 'fileignorecase', *Now, after this PR, '.H' is ~~cpp~~ c, '.h' is c. 
// of course, when g:c_syntax_for_h also was set.

what to fix is: '.h' can be 'c'.

chrisbra:
So if I understand correctly, when `fileignorecase` is set, it won't use this condition, but rather the normal `*.c` and `*.h` patterns?

However, shouldn't that be rather:
```
	au BufNewFile,BufRead *.H if !&fileignorecase | setf cpp | endif
	au BufNewFile,BufRead *.C if !&fileignorecase | call dist#ft#FTlpc() | endif
```
?

Shane-XB-Qian:
> So if I understand correctly, when `fileignorecase` is set, it won't use this condition, but rather the normal `*.c` and `*.h` patterns?

not sure what you meant 'this condition', if 'fileignorecase' is set, '.h' was overrode as 'cpp', that's why and what this modification/PR for.

> However, shouldn't that be rather:
> ```
> 	au BufNewFile,BufRead *.H if !&fileignorecase | setf cpp | endif
> 	au BufNewFile,BufRead *.C if !&fileignorecase | call dist#ft#FTlpc() | endif
> ```
> ? 

i think it's not a good idea, since had such au 'dist#ft#FTlpc()' on '*.c' already, (actually similar on '.h' also), here should be just about 'upper/lower case'.

---
BTW: there some others 'fname_case' (if you meant this 'condition') au, there should be similar broken on others ft, you may need to fix it later as well.
// but here i am only talking about 'c/cpp'

-- 
shane.xb.qian

Shane-XB-Qian:
BTW: there some others 'fname_case' (if you meant this 'condition') au, there should be similar broken on others ft, you may need to fix it later as well.
// but here i am only talking about 'c/cpp'

chrisbra:
Sorry, I don't follow. Please give answers to concrete questions. I find it hard to understand what you mean. 

You said: 
> if user did Not set 'fileignorecase', then everything is same like all it was working like;

What exactly do you mean by that? Apparently not how I understood you:

> So if I understand correctly, when `fileignorecase` is set, it won't use this condition, but rather the normal `*.c` and `*.h` patterns?

Then this one:

> i think it's not a good idea, since had such au 'dist#ft#FTlpc()' on '*.c' already, (actually similar on '.h' also), here should be just about 'upper/lower case'.

So why should `*.C` with uppercase `C` then be set to filetype `cpp` and not use the condition to detect `c` files?

Shane-XB-Qian:
> Sorry, I don't follow. Please give answers to concrete questions. I find it hard to understand what you mean. 
> 
> You said: 
> > if user did Not set 'fileignorecase', then everything is same like all it was working like;
> 
> What exactly do you mean by that? Apparently not how I understood you:
> 
> > So if I understand correctly, when `fileignorecase` is set, it won't use this condition, but rather the normal `*.c` and `*.h` patterns?
> 
> Then this one:
> 
> > i think it's not a good idea, since had such au 'dist#ft#FTlpc()' on '*.c' already, (actually similar on '.h' also), here should be just about 'upper/lower case'.
> 
> So why should `*.C` with uppercase `C` then be set to filetype `cpp` and not use the condition to detect `c` files?

please check:

line#284 and line#361 of filetype.vim, these are the au for '*.c' and '*.h';
line#351 this is the modification here, which to avoid 'fname_case' conflict;

only you can complain is these were spread out (but not was me) which seems messed,
otherwise i think this modification was simple enough which just modified one line.

-- 
shane.xb.qian

chrisbra:
you did not answer my question... :( 

So let me ask you again, please give clear answers.

Shouldn't this:
```
	au BufNewFile,BufRead *.C,*.H if !&fileignorecase | setf cpp | endif
```
Rather be 
```
	au BufNewFile,BufRead *.C if !&fileignorecase | call dist#ft#FTlpc() | endif
	au BufNewFile,BufRead *.H if !&fileignorecase | call dist#ft#FTheader() | endif
```
instead? 

Otherwise, since I haven't heard of any actual problems for those lines, I tend to just leave them as is, they have been like that at least since https://github.com/vim/vim/commit/5c73622a9066#diff-c07c0771a9f5af08e375703487a8fe7bbddd71f98a7bbe06acd78365525a243e (which is more than 10 years already)

Shane-XB-Qian:
> you did not answer my question... :( 
> 
> So let me ask you again, please give clear answers.
> 
> Shouldn't this:
> ```
> 	au BufNewFile,BufRead *.C,*.H if !&fileignorecase | setf cpp | endif
> ```
> Rather be 
> ```
> 	au BufNewFile,BufRead *.C if !&fileignorecase | call dist#ft#FTlpc() | endif
> 	au BufNewFile,BufRead *.H if !&fileignorecase | call dist#ft#FTheader() | endif
> ```
> instead? 

ok, my answer is 'No'
// since it wrapped with 'if has("fname_case")', some system may no such feat,
// and it is a dup code with the au at line#284 and line#361

> Otherwise, since I haven't heard of any actual problems for those lines, I tend to just leave them as is, they have been like that at least since https://github.com/vim/vim/commit/5c73622a9066#diff-c07c0771a9f5af08e375703487a8fe7bbddd71f98a7bbe06acd78365525a243e (which is more than 10 years already)

you can simple find the issue via such:
`vi --clean --cmd 'let g:c_syntax_for_h = 1' --cmd 'set fic' x.h` and check ft.

-- 
shane.xb.qian

Shane-XB-Qian:
> that at least since [5c73622a9066#diff-c07c0771a9f5af08e375703487a8fe7bbddd71f98a7bbe06acd78365525a243e](https://github.com/vim/vim/commit/5c73622a9066#diff-c07c0771a9f5af08e375703487a8fe7bbddd71f98a7bbe06acd78365525a243e) (which is more than 10 years already)

it was v7.2.328, fic was added v7.3.883, so such code was added when it was no 'fic' yet.

chrisbra:
I think it is better to make it behave similar to the `*.[ch]` filetypes, so I'll make it consistent

Shane-XB-Qian:
> I think it is better to make it behave similar to the `*.[ch]` filetypes, so I'll make it consistent

ok, as long as '.h' can be 'c' ft, you can modify mine to fit the result, currently it do spread out a bit messed.
// and as mentined above comment, others 'fname_case' those au for others ft maybe need to be re-check as well.

-- 
shane.xb.qian

chrisbra:
included.

PR #13643, opened 2023-12-06
none: ternary opt doc typo

Shane-XB-Qian:
// 'expr1 or expr1' seems odd
// 'expr2 or expr1' should be

### DIFF
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -939,7 +939,7 @@ to avoid running out of stack and crashing. *E1169*
 expr1				*expr1* *ternary* *falsy-operator* *??* *E109*
 -----
 
-The ternary operator: expr2 ? expr1 : expr1
+The ternary operator: expr2 ? expr2 : expr1
 The falsy operator:   expr2 ?? expr1
 
 Ternary operator ~


Shane-XB-Qian:
> No, this is correct. exprN represents the priority of the expression as explained in `:h expression-syntax`.

yes, it was correct, but it is confused. 
it should be better to say similiar exper between ternary and falsy.
i supposed you noticed the different between origin and modification.

// as for your said ':h expression-syntax', supposed you meant that 'if-then-else', i modified that too now, 
```diff
-       expr2 ? expr1 : expr1   if-then-else
+       expr2 ? expr1 : expr3   if-then-else
```

do you mind or could you re-open this ticket?

k-takata:
I agree that the current document is a bit confusing. But,

> expr2 ? expr1 : expr3

This is technically incorrect.
The 2nd and 3rd expressions are both `expr1` (expression priority 1).
It could be written `expr2 ? expr1-1 : expr1-2`? But then, other parts should also be changed in the same way.

k-takata:
I meant to change like this:

```diff
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -852,12 +852,12 @@ 2. Expression syntax					*expression-syn
 Expression syntax summary, from least to most significant:
 
 |expr1|	expr2
-	expr2 ? expr1 : expr1	if-then-else
+	expr2 ? expr1-1 : expr1-2	if-then-else
 
 |expr2|	expr3
-	expr3 || expr3 ...	logical OR
+	expr3-1 || expr3-2 ...	logical OR
 
 |expr3|	expr4
-	expr4 && expr4 ...	logical AND
+	expr4-1 && expr4-2 ...	logical AND
 
 |expr4|	expr5
```
Not sure if this is the best way...

Shane-XB-Qian:
ok, whatever, to me, i felt `The ternary operator: expr2 ? expr1 : expr1` here 'expr1 or expr1' seems odd.
// i saw you opened #13645, please try to get it a bit clear at there. thanks.

Issue #13501, opened 2023-11-07
[vim9class] `class`/`typealias` names used as variables and/or values


errael:
### Steps to reproduce

This issue has backwards compatibility implications.

Using a class/typealias name as a variable or value is not specified.
Either the behavior should be specified or generate an error.
(Opening this issue so it's not overlooked)

May supersede #12961, #13450. May affect #13498

The following generates no error. I believe just about every line should be an error.

```
vim9script

class C
endclass

class D
endclass

type Ctyp = C

def F(arg: any)
enddef

# Shouldn't all the following be errors?

D = C
D = Ctyp
var x = C
var y = Ctyp

F(C)
F(Ctyp)

var Func: func = F
Func(C)
Func(Ctyp)

var d: dict<any> = {a: C}
var l: list<any> = [C]

def G()
    D = C D = Ctyp
    var x0 = C
    var y0 = Ctyp
    F(C)
    F(Ctyp)
    Func(C)
    Func(Ctyp)

    var d0: dict<any> = {a: C}
    var l0: list<any> = [C]
enddef
G()

# returning a type thing
def F1(arg: bool): any
    if arg
        return C
    else
        return Ctyp
    endif
enddef
F1(true)
```

### Expected behaviour

Unspecified behavior should generate an error. This avoids backward compatibility issues.

### Version of Vim

9.0.2092

### Environment

ubuntu

errael:
Taken care of somewhere in #13637, #13675, #13683, #13688

PR #13644, opened 2023-12-07
Modify `instanceof()` to use varargs instead of list

### LINKED ISSUE #13421: `instanceof()` takes a list of classes
### Steps to reproduce

@LemonBoy @lifepillar @dkearns @yegappan, and any others, please review

Issue #13404 prevents non-value items from being used as values. _It does allow non-value items to be used as arguments to builtin functions_.

This interferes with the way that `instanceof` currently works. Note that in many languages instanceof is built into the language/compiler; `vim9script` uses a function.

Without a messy carve out for handling `instanceof` here are some options

Assuming "Base1", "Base2", "Intf" are classes or interfaces
1. This is the current function interface, something like this needs to work
    `instanceof(b3, [Base1, Base2, Intf1])`
2. Use varargs with the builtin
    `instanceof(b3, Base1, Base2, Intf1)`
3. Pass in classes by name (as strings)
    `instanceof(b3, ["Base1", "Base2", "Intf1"])`

Please weigh in on (2.) or (3.) or propose other suggestions.

And certainly bring up any related issues or considerations.

[1] #12867 introduced `instanceof()`

### Expected behaviour

(2.) above seem minimally disruptive.

### Version of Vim

9.0.2059

### Environment

ubuntu

### Logs and stack traces

_No response_

errael:
Fix #13421
Valid `instanceof()` arguments are `type`s. A `type` is not a value; it cannot be added to a list.

This change is non-compatible with the current usage of `instanceof()`; but `instanceof()` is relatively new and it's a trivial change. `instanceof(o, [C1, C2])` is changed to `instanceof(o, C1, C2)`. See #13421 for discussion.

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -1778,8 +1778,8 @@ EXTERN char e_can_only_compare_list_with_list[]
 	INIT(= N_("E691: Can only compare List with List"));
 EXTERN char e_invalid_operation_for_list[]
 	INIT(= N_("E692: Invalid operation for List"));
-EXTERN char e_list_or_class_required_for_argument_nr[]
-	INIT(= N_("E693: List or Class required for argument %d"));
+EXTERN char e_class_or_typealias_required_for_argument_nr[]
+	INIT(= N_("E693: Class or class typealias required for argument %d"));
 EXTERN char e_invalid_operation_for_funcrefs[]
 	INIT(= N_("E694: Invalid operation for Funcrefs"));
 EXTERN char e_cannot_index_a_funcref[]
@@ -3562,13 +3562,13 @@ EXTERN char e_using_typealias_as_value_str[]
 EXTERN char e_abstract_cannot_be_used_in_interface[]
 	INIT(= N_("E1404: Abstract cannot be used in an interface"));
 EXTERN char e_using_class_as_value_str[]
-	INIT(= N_("E1403: Class \"%s\" cannot be used as a value"));
+	INIT(= N_("E1405: Class \"%s\" cannot be used as a value"));
 EXTERN char e_using_class_as_var_val[]
-	INIT(= N_("E1405: Cannot use a Class as a variable or value"));
+	INIT(= N_("E1406: Cannot use a Class as a variable or value"));
 EXTERN char e_using_typealias_as_var_val[]
-	INIT(= N_("E1406: Cannot use a Typealias as a variable or value"));
+	INIT(= N_("E1407: Cannot use a Typealias as a variable or value"));
 #endif
-// E1405 - E1499 unused (reserved for Vim9 class support)
+// E1408 - E1499 unused (reserved for Vim9 class support)
 EXTERN char e_cannot_mix_positional_and_non_positional_str[]
 	INIT(= N_("E1500: Cannot mix positional and non-positional arguments: %s"));
 EXTERN char e_fmt_arg_nr_unused_str[]
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -758,17 +758,23 @@ arg_string_or_func(type_T *type, type_T *decl_type UNUSED, argcontext_T *context
 }
 
 /*
- * Check "type" is a list of 'any' or a class.
+ * Check varargs' "type" are class.
  */
     static int
-arg_class_or_list(type_T *type, type_T *decl_type UNUSED, argcontext_T *context)
+varargs_class(type_T *type UNUSED,
+	      type_T *decl_type UNUSED,
+	      argcontext_T *context)
 {
-    if (type->tt_type == VAR_CLASS
-	    || type->tt_type == VAR_LIST
-	    || type_any_or_unknown(type))
-	return OK;
-    arg_type_mismatch(&t_class, type, context->arg_idx + 1);
-    return FAIL;
+    for (int i = context->arg_idx; i < context->arg_count; ++i)
+    {
+	type2_T *types = &context->arg_types[i];
+	if (types->type_curr->tt_type != VAR_CLASS)
+	{
+	    semsg(_(e_class_or_typealias_required_for_argument_nr), i + 1);
+	    return FAIL;
+	}
+    }
+    return OK;
 }
 
 /*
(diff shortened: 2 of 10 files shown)


src/errors.h
```diff
@@ -1778,8 +1778,8 @@ EXTERN char e_can_only_compare_list_with_list[]
 	INIT(= N_("E691: Can only compare List with List"));
 EXTERN char e_invalid_operation_for_list[]
 	INIT(= N_("E692: Invalid operation for List"));
-EXTERN char e_list_or_class_required_for_argument_nr[]
-	INIT(= N_("E693: List or Class required for argument %d"));
+EXTERN char e_class_or_typealias_required_for_argument_nr[]
+	INIT(= N_("E693: Class required for argument %d"));
```
yegappan: Class or typealias?

errael:
@yegappan @chrisbra After this PR and #13637 are committed, I can open a couple PRs to finish up #13501 (which considers some backwards compatibility issues).

If there's some considerations, interfering issue, or further discussion required, I'd appreciate knowing about it.

src/evalfunc.c
```diff
@@ -1152,7 +1150,7 @@ static argcheck_T arg1_len[] = {arg_len1};
 static argcheck_T arg3_libcall[] = {arg_string, arg_string, arg_string_or_nr};
 static argcheck_T arg14_maparg[] = {arg_string, arg_string, arg_bool, arg_bool};
 static argcheck_T arg2_filter[] = {arg_list_or_dict_or_blob_or_string_mod, arg_filter_func};
-static argcheck_T arg2_instanceof[] = {arg_object, arg_class_or_list};
+static argcheck_T arg219_instanceof[] = {arg_object, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class, arg_class};
```
yegappan: This way of specifying the variable length argument types looks messy.  Is there any other way to simplify this?

Many of the internal functions evolved over time by adding additional optional arguments and in most cases Dicts were used for these arguments.  If we specify the maximum number of argument types like this, then we cannot add a different argument type in the future to instanceof().

src/evalfunc.c
```diff
@@ -2152,7 +2150,7 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_inputsecret},
     {"insert",		2, 3, FEARG_1,	    arg23_insert,
 			ret_first_arg,	    f_insert},
-    {"instanceof",	2, 2, FEARG_1,	    arg2_instanceof,
+    {"instanceof",	2, 19, FEARG_1,	    arg219_instanceof,
```
yegappan: Instead of using the maximum number of arguments (19) to specify the variable length arguments, can you use -1?  You can then check for -1 in the internal_func_check_arg_types() function and skip the compile time checks.  You need to use runtime checks in the instanceof() function.

yegappan (changes requested):

yegappan:
> @yegappan @chrisbra After this PR and #13637 are committed, I can open a couple PRs to finish up #13501 (which considers some backwards compatibility issues).
> 
> If there's some considerations, interfering issue, or further discussion required, I'd appreciate knowing about it.

I had some concerns about the way variable length argument types to instanceof() are added.  Added some comments to the PR.

src/evalfunc.c
errael: > Instead of using the maximum number of arguments (19) to specify the variable length arguments, can you use -1? You can then check for -1 in the internal_func_check_arg_types() function

OK, sounds better; I'll take a look at that.

> and skip the compile time checks.

I'll try to keep a compile time check if it's not too much trouble. `instanceof()` has simple, specific and limited argument requirements. `instanceof()` is a strange case. Strictly, it should be part of the language; for example Java does
```
if (<var> instanceof <type>)
```
Deferring the check to runtime seems wrong.

> You need to use runtime checks in the instanceof() function.

There is a runtime check in f_instanceof; or am I missing something?

yegappan: Yes.  The runtime check is already present.  You can check the first two argument types at compile time (minimum number of arguments).  The rest is anyway checked at runtime.

src/errors.h
errael: Oops; changing.

errael: Now I remember. I guess it's an implementation detail. But in the code, a class typealias is never seen (AFAICT). I guess it gets converted early so that all the normal class stuff works. Changed it to
```
E693: Class or class typealias required for argument %d
```
So it's clear when the code is
```
vim9script
class C
endclass
type T2 = list<string>
var v = instanceof(C.new(), T2)
```
```

**Comment by errael** [comment]
For a varargs builtin, this PR now handles

1. sets `funcentry_T.f_max_argc` to `VARGS`
    where `#define VARGS CHAR_MAX`
2. A simple mechanism so that the compile time check can verify all of the varargs.

By using a positive number for f_max_argc the too many args check remains unchanged.

File: src/evalfunc.c [hunk 4]
```diff
@@ -2152,7 +2167,7 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_inputsecret},
     {"insert",		2, 3, FEARG_1,	    arg23_insert,
 			ret_first_arg,	    f_insert},
-    {"instanceof",	2, 2, FEARG_1,	    arg2_instanceof,
+    {"instanceof",	2, VARGS, FEARG_1,  arg2_instanceof,
```
yegappan, line 2170: Can you also update the description of the instanceof() arguments in builtin.txt help file?

errael:
To point out for reference, there's the comment
```
 * The function may be varargs. In that case
 *	- f_max_argc == VARGS
 *	- f_argcheck must be NULL terminated. Last non-null argument
 *	  should check all the remaining args.
 *	  NOTE: if varargs, there can only be one NULL in f_argcheck array.
```
But over time it could be turned into a single meaning for NULL in f_argcheck array. Could/should define
```
    static int
arg_any(type_T *type UNUSED,
	      type_T *decl_type UNUSED,
	      argcontext_T *context UNUSED)
{
    return OK;
}
```
Which can be used in an f_argcheck array instead of null. For example, instead of
```
static argcheck_T arg3_number_any_dict[] = {arg_number, NULL, arg_dict_any};
```
it could be
```
static argcheck_T arg3_number_any_dict[] = {arg_number, arg_any, arg_dict_any};
```
Then eventually all the "NULL" could be replaced and its use wouldn't be ambiguous.

yegappan (approved):

yegappan:
Thanks for making the changes.  Looks good to me.

PR #13645, opened 2023-12-07
runtime(doc): Update eval.txt

k-takata:
expr5 is also expr6 itself.


Shane-XB-Qian:
this seems had not updated the conflict contents in https://github.com/vim/vim/pull/13643 yet before close.
// should, or let it be? since mine had been closed.

k-takata:
@Shane-XB-Qian 
Could you create a new PR? We may still need discussion about this.

Shane-XB-Qian:
> @Shane-XB-Qian 
> Could you create a new PR? We may still need discussion about this.

no, i have no interesting on that doc now, let it be or you can update it on your taste later.

-- 
shane.xb.qian

Issue #13639, opened 2023-12-06
the left odd vim9 script type


Shane-XB-Qian:
### Steps to reproduce

the left odd things in vim9 script type, just a mark to @yegappan
// refer: https://github.com/vim/vim/issues/13582

```vim
vim9script

var a: any
echom typename(a)  # number # tho instr in E1134 but why not is 'any'?

var as: any = "abc"
echom typename(as)  # string

var al: any = [1, 2, 3]
echom typename(al)  # any # why here is a 'any' not is 'list<number>'?
```

### Expected behaviour

n/a

### Version of Vim

v9.0.2153

### Environment

linux

yegappan:
When creating a script local variable, the variable needs to have a specific type and value.  So it is initialized to a number.
See: https://github.com/vim/vim/blob/master/src/vim9script.c#L850

Shane-XB-Qian:
> When creating a script local variable, the variable needs to have a specific type and value.  So it is initialized to a number.
> See: https://github.com/vim/vim/blob/master/src/vim9script.c#L850

how about think that case 'var a: any' (without assignment) is non-specific type yet?
a.k.a it didnot need to be inited, if so, it is a 'any'
// which should be better than as an odd 'number' type?

-- 
shane.xb.qian

Shane-XB-Qian:
:h e1134 said that too, but just i felt that's a bit odd to 'any', vs/tho make sense to others direct types.
// this comment in this ticket just is a mark anyway, i may suggest or wish to improve,
// but actually not a big deal, so up to you.

Nickwiz:
> how about think that case 'var a: any' (without assignment) is non-specific type yet?

There is no such thing as a *non specific type*. Closest, I guess,, is `null`, but even if one had `undefined`, `void` or what not those would also be `any`.  The `any` type only makes sense during declaration given the tools for identifying, namely `type` and `typename`. In legacy script one are forced to explicitly give variables a value, in v9 this is done automatically.

Both type and typename reflect the *value*. One would need a *new* function aka `is_any` for `any` alone that reflected the definition instead of value..

It is hard to see a good use-case where `type` or `typename` returning `any` would be of any (good) use.

```
IF x IS ANY
	Do Anything
END IF

vs

Do Anything
```

Shane-XB-Qian:
> > how about think that case 'var a: any' (without assignment) is non-specific type yet?
> 
> There is no such thing as a _non specific type_. Closest, I guess,, is `null`

[...]

PR #13646, opened 2023-12-07
When assigning a List or a Dict value to a variable of type 'any', keep the type

yegappan:
When a number, float, boolean, string or blob value is assigned to a variable of type "any", the variable type is set to the type of the value.  But when a List or Dict value is assigned to the variable, the type of the variable is set to "any".  Make it consistent and retain the type of the value.

Fixes one of the issues mentioned in #13639.


chrisbra:
thanks!

PR #13648, opened 2023-12-07
The options[] array is not sorted alphabetically

zeertzjq:
Problem:  The options[] array is not sorted alphabetically.
Solution: Sort it alphabetically.  Add a test.  Avoid unnecessary loop
          iterations in findoption().

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -4933,11 +4933,13 @@ findoption(char_u *arg)
 	opt_idx = quick_tab[26];
     else
 	opt_idx = quick_tab[CharOrdLow(arg[0])];
-    for ( ; (s = options[opt_idx].fullname) != NULL; opt_idx++)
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

chrisbra:
@zeertzjq it seems the options array is not completely sorted, `mzscheme` comes before `mzquantum`: https://github.com/vim/vim/blob/3afc9f2556faf2574d7950d879f2bf61612f3367/src/optiondefs.h#L1765-L1789

This causes a test breakage on vim-win32-installer repo: https://ci.appveyor.com/project/chrisbra/vim-win32-installer/build/job/ppx3bg6fy7d57awu

Can you please check it?

zeertzjq:
Ah yes. I should have sorted the result of filtering `optiondefs.h` instead of only options that are enabled. I've create PR #13690 to fix this.

PR #13650, opened 2023-12-07
runtime(syntax): unlet b:filetype_in_cpp_family

laburnumT:
Update runtime/syntax/cpp.vim to unlet b:filetype_in_cpp_family as it remains set even after updating the ft of a file manually or through a modeline, not allowing c specific keywords to be highlighted.

### DIFF
--- a/runtime/syntax/cpp.vim
+++ b/runtime/syntax/cpp.vim
@@ -125,5 +125,6 @@ hi def link cppFloat		Number
 hi def link cppModule		Include
 
 let b:current_syntax = "cpp"
+unlet b:filetype_in_cpp_family
 
 " vim: ts=8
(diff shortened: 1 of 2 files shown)


dkearns (commented):
It might be better to unlet the variable just after the c file is sourced (e.g., line 16 in cpp.vim) rather than at the end of the file as it's no longer of any use after that.

It makes its purpose a little clearer.

laburnumT:
Hmm. That makes sense, but the end of the file feels like a more logical
place to me personally. Not that it matters for these files, but what if
there was a file that wanted to pull in more than one syntax file and
not all at the beginning?

Though maybe that is too far-fetched of an idea and we shouldn't worry
about that.

chrisbra:
Thanks, I have made that change on top of your change and updated the last-change header in addition.

PR #13653, opened 2023-12-09

dkearns:
Improve variable highlighting in :let, :unlet, :const and :for commands.

Match registers and local, global and terminal option variables.


chrisbra:
thanks!

PR #13655, opened 2023-12-10
Support final and const object and class variables

yegappan:
Implement the following item from the todo list:

- "final" object members - can only be set in the constructor.

### DIFF
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -4505,7 +4505,10 @@ E1401	vim9class.txt	/*E1401*
 E1402	vim9class.txt	/*E1402*
 E1403	vim9class.txt	/*E1403*
 E1407	vim9class.txt	/*E1407*
+E1408	vim9class.txt	/*E1408*
+E1409	vim9class.txt	/*E1409*
 E141	message.txt	/*E141*
+E1410	vim9class.txt	/*E1410*
 E142	message.txt	/*E142*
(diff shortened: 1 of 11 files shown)


Shane-XB-Qian:
was there something changed since that decision https://github.com/vim/vim/pull/12932#issuecomment-1699624713
which said 'final/const' is not but 'readonly' was preferred?

-- 
shane.xb.qian

Shane-XB-Qian:
anyway, i think 'final' should be there, but the 'readonly' it was a confused concept.

yegappan:
> was there something changed since that decision [#12932 (comment)](https://github.com/vim/vim/pull/12932#issuecomment-1699624713) which said 'final/const' is not but 'readonly' was preferred?

Any variable can be made a constant by using the `final` keyword.  This PR adds support for making an object variable constant (and can be assigned only in the constructor).  These variables cannot be reassigned even by other methods in the object.
This to make the use of the `final` keyword consistent between object variables and variables defined outside of an object.

Shane-XB-Qian:
yea, i agreed and preferred 'final' (or 'const') should be there.

vim-ml:
Hi,

I was waiting for PR https://github.com/vim/vim/pull/13670 to be merged.
As that PR is now merged, this PR (support for final and const object/class
variables) is ready.

Regards,
Yegappan

chrisbra:
thanks!

PR #13656, opened 2023-12-10
Termdebug: add Tbreak command

iam28th:
Add `:Tbreak` command to insert a temporary breakpoint at the cursor location. 

Currently it can be achieved by using `tbreak` in the gdb-window, but typing the address in there looks much less convenient compared to vim navigation.

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -953,7 +953,7 @@ func s:InstallCommands()
   set cpo&vim
 
   command -nargs=? Break call s:SetBreakpoint(<q-args>)
-  command -nargs=? Tbreak call s:SetBreakpoint(<q-args>, 1)
+  command -nargs=? Tbreak call s:SetBreakpoint(<q-args>, v:true)
   command Clear call s:ClearBreakpoint()
   command Step call s:SendResumingCommand('-exec-step')
   command Over call s:SendResumingCommand('-exec-next')
@@ -1169,7 +1169,7 @@ func s:Until(at)
 endfunc
 
 " :Break - Set a breakpoint at the cursor position.
-func s:SetBreakpoint(at, t=0)
+func s:SetBreakpoint(at, tbreak=v:false)
   " Setting a breakpoint may not work while the program is running.
   " Interrupt to make it work.
   let do_continue = 0
@@ -1182,7 +1182,7 @@ func s:SetBreakpoint(at, t=0)
   " Use the fname:lnum format, older gdb can't handle --source.
   let at = empty(a:at) ?
 	\ fnameescape(expand('%:p')) . ':' . line('.') : a:at
-  if a:t
+  if a:tbreak
     let cmd = '-break-insert -t ' . at
   else
     let cmd = '-break-insert ' . at
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -953,6 +953,7 @@ func s:InstallCommands()
   set cpo&vim
 
   command -nargs=? Break call s:SetBreakpoint(<q-args>)
+  command -nargs=? Tbreak call s:SetBreakpoint(<q-args>, 1)
   command Clear call s:ClearBreakpoint()
   command Step call s:SendResumingCommand('-exec-step')
   command Over call s:SendResumingCommand('-exec-next')
@@ -1067,6 +1068,7 @@ endfunc
 " Delete installed debugger commands in the current window.
 func s:DeleteCommands()
   delcommand Break
+  delcommand Tbreak
   delcommand Clear
   delcommand Step
   delcommand Over
@@ -1167,7 +1169,7 @@ func s:Until(at)
 endfunc
 
 " :Break - Set a breakpoint at the cursor position.
-func s:SetBreakpoint(at)
+func s:SetBreakpoint(at, t=0)
   " Setting a breakpoint may not work while the program is running.
   " Interrupt to make it work.
   let do_continue = 0
@@ -1180,7 +1182,12 @@ func s:SetBreakpoint(at)
   " Use the fname:lnum format, older gdb can't handle --source.
   let at = empty(a:at) ?
 	\ fnameescape(expand('%:p')) . ':' . line('.') : a:at
-  call s:SendCommand('-break-insert ' . at)
+  if a:t
+    let cmd = '-break-insert -t ' . at
+  else
+    let cmd = '-break-insert ' . at
+  endif
+  call s:SendCommand(cmd)
   if do_continue
     Continue
   endif
--- a/src/testdir/test_termdebug.vim
+++ b/src/testdir/test_termdebug.vim
@@ -18,8 +18,8 @@ if g:GCC->empty()
   throw 'Skipped: gcc is not found in $PATH'
 endif
 
-function s:generate_files(bin_name) 
-  let src_name = a:bin_name . '.c'
+function s:generate_files(bin_name)
+  let src_name = a:bin_name .. '.c'
   let lines =<< trim END
     #include <stdio.h>
     #include <stdlib.h>
@@ -47,18 +47,18 @@ function s:generate_files(bin_name)
   END
   call writefile(lines, src_name)
   call system($'{g:GCC} -g -o {a:bin_name} {src_name}')
-endfunction 
+endfunction
 
 function s:cleanup_files(bin_name)
   call delete(a:bin_name)
-  call delete(a:bin_name . '.c')
+  call delete(a:bin_name .. '.c')
 endfunction
 
 packadd termdebug
 
 func Test_termdebug_basic()
   let bin_name = 'XTD_basic'
-  let src_name = bin_name . '.c'
+  let src_name = bin_name .. '.c'
   call s:generate_files(bin_name)
 
   edit XTD_basic.c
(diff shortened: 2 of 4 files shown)


chrisbra:
Thanks. Can you please add a test for this?

iam28th:
Add test for `Tbreak` functionality. It checks signs in the buffer, similar to what the first part of [Test_termdebug_basic()](https://github.com/vim/vim/blob/master/src/testdir/test_termdebug.vim#L23) does.

iam28th:
Is there anything else I should do? 
The failed test seems to be about autogenerated index files, so maybe failing this test is not a big deal...

Shane-XB-Qian:
bram said he donot want to make too many gdb cmd to be termdebug cmd,
so i am thinking if it was really want to add this cmd, how about to be a sub cmd of Break?
// which default is 'Break' which no args, vs if 'Break t[mp]' then same like 'Tbreak' cmd or 'tbreak' gdb cmd?

iam28th:
I would argue against this, because currently most Termdebug commands that are wrappers around corresponding `gdb` commands use the particular naming scheme (run -> Run, continue -> Continue, etc, the only exception is where this naming clashes with the `:Next` vim command). 
Implementing `tbreak` as `:Break t[mp]` would unexpectedly break this pattern.

I agree that not all `gdb` commands need to have corresponding command in Termdebug, but from my user experience - `tbreak` certainly needs one.

Shane-XB-Qian:
> I would argue against this, because currently most Termdebug commands that are wrappers around corresponding `gdb` commands use the particular naming scheme (run -> Run, continue -> Continue, etc, the only exception is where this naming clashes with the `:Next` vim command). 
> Implementing `tbreak` as `:Break t[mp]` would unexpectedly break this pattern.

the different is '-break-insert' and '-break-insert -t', so i think it was a special case of 'break'

-- 
shane.xb.qian

iam28th:
> the different is '-break-insert' and '-break-insert -t', so i think it was a special case of 'break'
> […](#)
> -- shane.xb.qian

From the internal perspective - yes, but not when using `gdb` from the terminal. Here you have separate `break` and `tbreak` commands.

src/testdir/test_termdebug.vim
```diff
@@ -152,6 +152,71 @@ func Test_termdebug_basic()
   %bw!
 endfunc
 
+func Test_termdebug_tbreak()
+  " the same as the first half part of Test_termdebug_basic(),
+  " but only 'debugPC' sign should be present after the breakpoint
+  " was hit once
```
yegappan: Instead of cloning the code from Test_termdebug_basic(), can you try to reuse that?  You can either move the new test to the Test_termdebug_basic() function or move the code for generating a test executable to a common function.

src/testdir/test_termdebug.vim
```diff
@@ -152,6 +152,71 @@ func Test_termdebug_basic()
...
+
+      printf("%d is %s prime\n", n, isprime(n) ? "a" : "not a");
+
+      return 0;
+    }
+  END
+  call writefile(lines, 'XTD_basic.c', 'D')
+  call system($'{g:GCC} -g -o XTD_basic XTD_basic.c')
+
+  edit XTD_basic.c
```
yegappan: A different file name should be used.  The "XTD_basic.c" is used in Test_termdebug_basic().

src/testdir/test_termdebug.vim
```diff
@@ -152,6 +152,71 @@ func Test_termdebug_basic()
...
+        \  'group': 'TermDebug'}],
+        \ sign_getplaced('', #{group: 'TermDebug'})[0].signs)})
+  Finish
+  call term_wait(gdb_buf)
+  redraw!
+  call WaitForAssert({-> assert_equal([
+        \ {'lnum': 20, 'id': 12, 'name': 'debugPC',
+        \  'priority': 110, 'group': 'TermDebug'}],
+        \ sign_getplaced('', #{group: 'TermDebug'})[0].signs)})
+  Continue
```
yegappan: This test doesn't check that the temporary breakpoint is hit only once.  Can you set a temporary breakpoint inside the for loop and a breakpoint in the main function after the call to printf(), then run "Continue" after the first break point is hit to make sure that the breakpoint is not hit again?

src/testdir/test_termdebug.vim
iam28th: Moved the code to generate and cleanup test files to functions.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -953,6 +953,7 @@ func s:InstallCommands()
   set cpo&vim
 
   command -nargs=? Break call s:SetBreakpoint(<q-args>)
+  command -nargs=? Tbreak call s:SetBreakpoint(<q-args>, 1)
```
yegappan: Can you use "v:true" instead of 1 here?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1167,7 +1169,7 @@ func s:Until(at)
 endfunc
 
 " :Break - Set a breakpoint at the cursor position.
-func s:SetBreakpoint(at)
+func s:SetBreakpoint(at, t=0)
```
yegappan: Can you use "tbreak = v:false" instead of "t=0"?

src/testdir/test_termdebug.vim
```diff
@@ -18,9 +18,8 @@ if g:GCC->empty()
   throw 'Skipped: gcc is not found in $PATH'
 endif
 
-packadd termdebug
-
-func Test_termdebug_basic()
+function s:generate_files(bin_name)
+  let src_name = a:bin_name . '.c'
   let lines =<< trim END
```
yegappan, line 23: The single dot for string concatenation is deprecated now.  Can you use ".." here?

src/testdir/test_termdebug.vim
```diff
@@ -46,8 +45,21 @@ func Test_termdebug_basic()
...
   END
-  call writefile(lines, 'XTD_basic.c', 'D')
-  call system($'{g:GCC} -g -o XTD_basic XTD_basic.c')
+  call writefile(lines, src_name)
+  call system($'{g:GCC} -g -o {a:bin_name} {src_name}')
+endfunction
+
+function s:cleanup_files(bin_name)
+  call delete(a:bin_name)
+  call delete(a:bin_name . '.c')
```
yegappan: Use double dots here.

src/testdir/test_termdebug.vim
```diff
@@ -46,8 +45,21 @@ func Test_termdebug_basic()
...
+endfunction
+
+packadd termdebug
+
+func Test_termdebug_basic()
+  let bin_name = 'XTD_basic'
+  let src_name = bin_name . '.c'
```
yegappan: Use double dots here.

src/testdir/test_termdebug.vim
```diff
@@ -148,7 +160,70 @@ func Test_termdebug_basic()
...
+endfunc
+
+func Test_termdebug_tbreak()
+  let bin_name = 'XTD_tbreak'
+  let src_name = bin_name . '.c'
+
+  eval s:generate_files(bin_name)
+
+  execute 'edit ' . src_name
+  execute 'Termdebug ./' . bin_name
```
yegappan: Use double dots here.

src/testdir/test_termdebug.vim
```diff
@@ -148,7 +160,70 @@ func Test_termdebug_basic()
...
   call assert_equal([], sign_getplaced('', #{group: 'TermDebug'})[0].signs)
 
-  call delete('XTD_basic')
+  call s:cleanup_files(bin_name)
+  %bw!
+endfunc
+
+func Test_termdebug_tbreak()
+  let bin_name = 'XTD_tbreak'
+  let src_name = bin_name . '.c'
```
yegappan: Use double dots here.

src/testdir/test_termdebug.vim
```diff
@@ -148,7 +160,70 @@ func Test_termdebug_basic()
...
+  execute 'Termdebug ./' . bin_name
+
+  call WaitForAssert({-> assert_equal(3, winnr('$'))})
+  let gdb_buf = winbufnr(1)
+  wincmd b
+
+  let bp_line = 22        " 'return' statement in main
+  let temp_bp_line = 10   " 'if' statement in 'for' loop body
+  execute "Tbreak " . temp_bp_line
+  execute "Break " . bp_line
```
yegappan: Use double dots here.

yegappan (changes requested):
Thanks for making the changes to the test.

yegappan (approved):

iam28th:
> The help tags check CI test is failing (https://github.com/vim/vim/actions/runs/7183328851/job/19562047377). You need to run the "make tags" command in the runtime/doc directory and commit the updated tags file.

Out of curiosity - what is the reasoning behind here? This file can always be auto-generated from source during build, so why is there a need to manually update it?

chrisbra:
Is the test flaky? 
```
command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_termdebug_tbreak line 21: Expected [{'lnum': 10, 'id': 1014, 'name': 'debugBreakpoint1.0', 'priority': 110, 'group': 'TermDebug'}, {'lnum': 22, 'id': 2014, 'name': 'debugBreakpoint2.0', 'priority': 110, 'group': 'TermDebug'}] but got [{'lnum': 22, 'id': 2014, 'name': 'debugBreakpoint2.0', 'priority': 110, 'group': 'TermDebug'}]
```

chrisbra:
> Out of curiosity - what is the reasoning here? This file can always be auto-generated from source during build, so why is there a need to manually update it?

Basically if people are downloading the runtime files manually, they don't have to generate the help tags file to make use of it with Vim.

iam28th:
> Is the test flaky?
> 
> ```
> command line..script /home/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_termdebug_tbreak line 21: Expected [{'lnum': 10, 'id': 1014, 'name': 'debugBreakpoint1.0', 'priority': 110, 'group': 'TermDebug'}, {'lnum': 22, 'id': 2014, 'name': 'debugBreakpoint2.0', 'priority': 110, 'group': 'TermDebug'}] but got [{'lnum': 22, 'id': 2014, 'name': 'debugBreakpoint2.0', 'priority': 110, 'group': 'TermDebug'}]
> ```

[This is the line](https://github.com/iam28th/vim/blob/98436b3d5cec338c7926efc2cb7f7eb2a7a5bfab/src/testdir/test_termdebug.vim#L188) where the check fails.

Could be the wait is too short? `call term_wait(gdb_buf)` defaults to 10 ms.
In the master branch, `call term_wait(gdb_buf, 400)` is [used once](https://github.com/vim/vim/blob/2025af165ec68d831f0f0f668a3ceac3f39142ef/src/testdir/test_termdebug.vim#L65) (only after `Run`, though).

Though the result is odd: the second breakpoint has a sign, and the first does not.

<details>
  <summary>Locally I can run this test 100 times without failures:</summary>
  
image
  
</details>

chrisbra:
okay thanks. Let me mark it as flaky for now.

PR #13657, opened 2023-12-10
Fix: ftplugin/sbt.vim should not set b:did_ftplugin

lervag:
The `b:did_ftplugin` guard was set and prevented us from actually sourcing `ftplugin/scala.vim`. Since the latter script also sets the guard properly, we can simply remove the guard here.

PS! I can explain this in more detail if necessary, but I believe this is a very simple and clear fix.

### DIFF
--- a/runtime/ftplugin/sbt.vim
+++ b/runtime/ftplugin/sbt.vim
@@ -9,7 +9,6 @@ if exists('b:did_ftplugin') || &cp
   finish
 endif
 
-let b:did_ftplugin = 1
-
 runtime! ftplugin/scala.vim
 
+let b:did_ftplugin = 1


chrisbra:
hm thanks. Yes it makes sense. But just for consistency, please either add a comment, why `b:did_ftplugin` should not be set, or just set it after the `runtime!` line please. 

I also checked a few other files, it seems it's seems this is either always set after the `runtime!` command or not at all (which seems inconsistent). The other ones that looks similary wrong is `ftplugin/rnowweb.vim` and `ftplugin/jsonc.vim` (jus ta note to myself).

lervag:
I'll add the `b:did_ftplugin = 1` at the end, then, for consistency.

chrisbra:
thanks!

chrisbra:
thanks!

PR #13664, opened 2023-12-11
labeler.yml: update for v5

pheiduck:
merge before:  https://github.com/vim/vim/pull/13658

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -62,9 +62,12 @@ if_python:
               - 'src/proto/if_python3.pro'
 
 if_ruby:
-- 'src/if_ruby.c'
-- 'src/proto/if_ruby.pro'
-- 'runtime/doc/if_ruby.txt'
+  - all:
+      - changed-files:
+          - any-glob-to-all-files:
+              - 'src/if_ruby.c'
+              - 'src/proto/if_ruby.pro'
+              - 'runtime/doc/if_ruby.txt'
 
 if_tcl:
   - all:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,106 +1,152 @@
 # list of labels and minimatch globs to match to apply the label.
 
 CI:
-- any: ['.cirrus.yml']
-- any: ['.github/dependabot.yml']
-- any: ['.github/labeler.yml']
-- any: ['.github/workflows/*']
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

.github/labeler.yml
```diff
@@ -1,106 +1,155 @@
...
 CI:
-- any: ['.cirrus.yml']
-- any: ['.github/dependabot.yml']
-- any: ['.github/labeler.yml']
-- any: ['.github/workflows/*']
-- any: ['.appveyor.yml']
-- any: ['.codecov.yml']
+  - all:
+      - changed-files:
+          - any-glob-to-all-files:
```
k-takata, line 6: I think `any-glob-to-any-file` should be used instead of `any-glob-to-all-files`.
In my understanding, if any files that don't match this glob are changed, this label won't be applied when `any-glob-to-all-files` is used.

.github/labeler.yml
```diff
@@ -1,106 +1,155 @@
...
+              - '.github/labeler.yml'
+              - '.github/workflows/*'
+              - '.appveyor.yml'
+              - '.codecov.yml'
 
 documentation:
-- all: ['*/doc/*']
+  - all:
+      - changed-files:
+          - any-glob-to-all-files:
```
k-takata, line 17: For the `documentation` label, I think using `any-glob-to-all-files` is correct because `all:` was used before this MR.

.github/labeler.yml
```diff
@@ -1,106 +1,155 @@
...
+              - '*/doc/*'
 
 translation:
-- any: ['src/po/*.po']
+  - all:
+      - changed-files:
+          - any-glob-to-all-files:
+              - 'src/po/*.po'
```
k-takata, line 24: With the default settings of Vim, the indents become different:
```diff
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -19,9 +19,9 @@ documentation:
 
 translation:
   - all:
-      - changed-files:
-          - any-glob-to-all-files:
-              - 'src/po/*.po'
+    - changed-files:
+      - any-glob-to-all-files:
+        - 'src/po/*.po'
 
 if_lua:
   - all:
```

Maybe it is better to follow the default setting of Vim. (Or add modelines?)

PR #13668, opened 2023-12-12

dkearns:
Fix highlighting of :for command.  Link the vimFor syntax group to the
vimCommand highlight group.

Error introduced in commit f686921.


chrisbra:
thanks!

PR #13669, opened 2023-12-12
Fix crash when calling screenpos() with a negative column

zeertzjq:


chrisbra:
can the same happen for a negative row?

zeertzjq:
That's already covered by the `pos->lnum >= wp->w_topline` check.

chrisbra:
thanks!

PR #13670, opened 2023-12-12
Use :var for object and class variable declarations

dkearns:
This PR modifies the object and class variable declarations to require use of the :var keyword.

The current object/class variable declaration syntax is inconsistent with that used in other language contexts where `:var`, `:const`, or `:final` introduce variables.  It is also inconsistent with object/class method declarations which reuse the syntax employed for all other function declarations, i.e., `:def`.

Classes and interfaces are not special enough, as syntactic structures, to warrant these deviations.  Similar programming languages usually reuse their general variable declaration syntax for field declarations, and the small number that don't use a dedicated and internally consistent class syntax.

Aside, from being a needless deviation from the general syntax with no obvious benefit, it also introduces various syntactic oddities, as highlighted by the [proposed syntax for final object/class variables](13655).

I initially raised this topic in and around the first call for comments on Vim9 classes, as did others.  The last I heard was that it was still "under consideration" but, with the recent unfortunate developments, some of these decisions appear to have been implicitly crystallised without further discussion.  As I dislike the syntax as much now, having actually used it anger, as I did when I first saw it twelve months ago it seems worth raising again while there is still an opportunity to change it.

For reference this topic is covered in vim9class.txt under "Rationale" -> "Declaring object and class variables" and quoted at the end here for convenience.

### Current syntax

Object variables are currently declared with a specific syntax unused elsewhere in the language and are not introduced with the `var` keyword.

```vim
this.x: number
this.y: number = 42
this.z = 87
```

On the other hand methods are defined with exactly the same syntax used for functions in other contexts using `:def` and an unqualified name. The `this.` prefix is neither required nor allowed.

```vim
def Foo(): number
  return 42
enddef
```

This special object variable syntax is distinct even from that used for class variables.  These use a specific modifier keyword `static` and the declaration has the same shape as other declarations using `var`, `final` and `const`.

```vim
static x: number
static y: number = 42
static z = 87
```

This could, of course, be viewed as a modifier keyword for an implied `var` declaration. 

```vim
static {var} x: number
static def Y()
enddef
```

One might have expected class variable declarations to have used something more consistent with their object variable counterparts like the class name or another class-wide shared object as a prefix, but this was not chosen on the grounds that it was too cumbersome.

```vim
class Foo
  Foo.x : number = 42
endclass
```

### Other languages

Similar languages, most commonly, reuse their general variable syntax when declaring class fields. 

E.g., Swift (Kotlin, Scala and others): 

```swift
import Foundation

var foo: Int = 42

func bar() -> String {
    var baz: String = "Forty Two"
    return baz
}

class Qux {
    var quux: Int = 42
    func corge() -> Int {
        var grault: Int = 2
        return quux * grault
    }
}

let o = Qux()
print(o.corge())
```

E.g., Dart (Java, C# and others using a C-style syntax):

```dart
int foo = 42;

String bar() {
  String baz = "Forty Two";
  return baz;
}

class Qux {
  int quux = 42;
  int corge() {
    int grault = 2;
    return quux * grault;
  }
}

void main() {
  var o = Qux();
  print(o.corge());
}

```

A notable exception to this is TypeScript which Bram referenced when introducing this Vim9-script class syntax.

```typescript
var foo: number = 42

function bar(): string {
  var baz: string = "Forty Two"
  return baz
}

class Qux {
  quux: number = 42
  corge(): number {
    var grault: number = 2
    return this.quux * grault
  }
}

var o = new Qux()
console.log(o.corge())
```

However, here a different and consistent method declaration is used which would look like the following for Vim9 script, if based on the existing object variable syntax.

```vim
this.foo: number = 42
this.Bar(arg: number): number {
  return this.foo + arg
}
```

### Proposed change

Following common language trends and, arguably, user expectations Vim9 script should simply reuse the exiting variable declaration syntax.

```vim
class Demo
  static var quux: string = "Forty Two"
  static def Grault(): string
    return quux;
  enddef
  var foo: number = 42
  def Bar(arg: number): number
    return this.x + arg
  enddef
endclass
```

### Future issues

There is a TODO item (currently available as a draft PR) to allow for final object/class variables.  The proposed syntax, again, does not mirror the use of the `var` and `final` sibling keywords as used in other declaration contexts.

```vim
static final a: number
static b: number
final this.x: number
# final x: number
this.y: number
```

Even if the final object member declarations are not `this` qualified either choice will be inconsistent with one of the current method or variable declaration styles.  `final` also tends to read as a modifier keyword here in this context rather than a variable declaration keyword, as these were previously not needed for object/class variable declarations. E.g.,

```vim
final {var} this.x: number
```

Using `var` to declare object variables allows `final` to be included without any extra syntactic quirks.

```vim
class Demo
  static final a: number
  static var b: number
  final x: number
  var y: string
endclass
```

### Previous discussions

https://groups.google.com/g/vim_dev/c/MKbzhgqvSj8/m/CTpMTe5YDQAJ
https://groups.google.com/g/vim_dev/c/yYpFNUHdRho/m/xjgrKqMoBQAJ?pli=1

Bram proposed the current syntax, aware that some would see it as unnatural, and with what I consider an unconvincing argument.

> Notice that there is no "var" keyword. It's not needed to recognize the declaration. I can't think of a good reason to add "var" here, other than that a declaration would be expected to always have "var". Well, I don't have that expectation.

`var` is not needed to recognise a declaration in other contexts either and yet it _is_ required everywhere else.  I find it hard to see that a different syntax for object members has any utility at all.  It seems needlessly quirky.

He later softened his opinion when adding the following to the help below:

> There is no clear preference whether to use "var" or not.

This seems like reason enough not to introduce different syntax.  Unform syntax is a nice language feature and one of the strong aspects of the initial Vim9-script release.

> The two main reasons to leave it out are:
> 1. TypeScript, Java and other popular languages do not use it.

Actually Java _does_ use a the same syntax for fields and local variables.  It just uses a C-style syntax, without a keyword.
As already mentioned, TypeScript uses a dedicated and consistent class syntax for _both_ variable and method declarations.

> 2. Less clutter.

The redundant `this.` is at least as cluttering and doesn't clearly announce a declaration.  A type inferred object variable declaration is indistinguishable from a plain assignment.

The proposed change is also one character shorter.

Others have previously made these points and suggested this same change in the referenced discussions.

***

### Rationale (from doc/vim9class.txt)

> Declaring object and class variables
>
> The main choice is whether to use "var" as with variable declarations.
> TypeScript does not use it:
>         class Point {
>           x: number;
>           y = 0;
>         }
>
> Following that Vim object variables could be declared like this:
>         class Point
>           this.x: number
>           this.y = 0
>         endclass
>
> Some users pointed out that this looks more like an assignment than a
> declaration.  Adding "var" changes that:
>         class Point
>           var this.x: number
>           var this.y = 0
>         endclass
>
> We also need to be able to declare class variables using the "static" keyword.
> There we can also choose to leave out "var":
>         class Point
>           var this.x: number
>           static count = 0
>         endclass
>
> Or do use it, before "static":
>         class Point
>           var this.x: number
>           var static count = 0
>         endclass
>
> Or after "static":
>         class Point
>           var this.x: number
>           static var count = 0
>         endclass
>
> This is more in line with "static def Func()".
>
> There is no clear preference whether to use "var" or not.  The two main
> reasons to leave it out are:
> 1. TypeScript, Java and other popular languages do not use it.
> 2. Less clutter.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -78,8 +78,8 @@ Let's start with a simple example: a class that stores a text position (see
 below for how to do this more efficiently): >
 
 	class TextPosition
-	   this.lnum: number
-	   this.col: number
+	   var lnum: number
+	   var col: number
 
 	   def new(lnum: number, col: number)
 	      this.lnum = lnum
@@ -156,8 +156,8 @@ On the other hand, if you do not want the object variables to be read directly
 from outside the class or its sub-classes, you can make them protected.  This
 is done by prefixing an underscore to the name: >
 
-	this._lnum: number
-	this._col number
+	var _lnum: number
+	var _col number
 
 Now you need to provide methods to get the value of the protected variables.
 These are commonly called getters.  We recommend using a name that starts with
@@ -209,8 +209,8 @@ Many constructors take values for the object variables.  Thus you very often
 see this pattern: >
 
 	 class SomeClass
-	   this.lnum: number
-	   this.col: number
+	   var lnum: number
+	   var col: number
 
 	   def new(lnum: number, col: number)
 	      this.lnum = lnum
@@ -235,8 +235,8 @@ Putting together this way of using new() and making the variables public
 results in a much shorter class definition than what we started with: >
 
 	class TextPosition
-	   public this.lnum: number
-	   public this.col: number
+	   public var lnum: number
+	   public var col: number
 
 	   def new(this.lnum, this.col)
 	   enddef
@@ -277,8 +277,8 @@ Class members are declared with "static".  They are used by the name without a
 prefix in the class where they are defined: >
 
 	class OtherThing
-	   this.size: number
-	   static totalSize: number
+	   var size: number
+	   static var totalSize: number
 
 	   def new(this.size)
 	      totalSize += this.size
@@ -297,9 +297,9 @@ underscore as the first character in the name, and it can be made public by
 prefixing "public": >
 
     class OtherThing
-	static total: number	      # anybody can read, only class can write
-	static _sum: number	      # only class can read and write
-	public static result: number  # anybody can read and write
+	static var total: number	  # anybody can read, only class can write
+	static var _sum: number	          # only class can read and write
+	public static var result: number  # anybody can read and write
     endclass
 <
 							*class-method*
@@ -308,8 +308,8 @@ variables but they have no access to the object variables, they cannot use the
 "this" keyword:
 >
 	class OtherThing
-	   this.size: number
-	   static totalSize: number
+	   var size: number
+	   static var totalSize: number
 
 	   # Clear the total size and return the value it had before.
 	   static def ClearTotalSize(): number
@@ -345,14 +345,14 @@ outside of the defining class: >
 
     vim9script
     class Vehicle
-	static nextID: number = 1000
+	static var nextID: number = 1000
 	static def GetID(): number
 	    nextID += 1
 	    return nextID
 	enddef
     endclass
     class Car extends Vehicle
-	this.myID: number
+	var myID: number
 	def new()
 	    this.myID = Vehicle.GetID()
 	enddef
@@ -380,20 +380,20 @@ it is.  The Shape class functions as the base for a Square and a Triangle
 class, for which objects can be created.  Example: >
 
 	abstract class Shape
-	   this.color = Color.Black
-	   this.thickness = 10
+	   var color = Color.Black
+	   var thickness = 10
 	endclass
 
 	class Square extends Shape
-	   this.size: number
+	   var size: number
 
 	   def new(this.size)
 	   enddef
 	endclass
 
 	class Triangle extends Shape
-	   this.base: number
-	   this.height: number
+	   var base: number
+	   var height: number
 
 	   def new(this.base, this.height)
 	   enddef
@@ -430,8 +430,8 @@ interface called HasSurface, which specifies one method Surface() that returns
 a number.  This example extends the one above: >
 
 	abstract class Shape
-	   this.color = Color.Black
-	   this.thickness = 10
+	   var color = Color.Black
+	   var thickness = 10
 	endclass
 
 	interface HasSurface
@@ -439,7 +439,7 @@ a number.  This example extends the one above: >
 	endinterface
 
 	class Square extends Shape implements HasSurface
-	   this.size: number
+	   var size: number
 
 	   def new(this.size)
 	   enddef
@@ -450,8 +450,8 @@ a number.  This example extends the one above: >
 	endclass
 
 	class Triangle extends Shape implements HasSurface
-	   this.base: number
-	   this.height: number
+	   var base: number
+	   var height: number
 
 	   def new(this.base, this.height)
 	   enddef
@@ -598,13 +598,13 @@ Items in a class ~
 						*E1318* *E1325* *E1388*
 Inside a class, in between `:class` and `:endclass`, these items can appear:
 - An object variable declaration: >
-	this._protectedVariableName: memberType
-	this.readonlyVariableName: memberType
-	public this.readwriteVariableName: memberType
+	var _protectedVariableName: memberType
+	var readonlyVariableName: memberType
+	public var readwriteVariableName: memberType
 - A class variable declaration: >
-	static _protectedClassVariableName: memberType
-	static readonlyClassVariableName: memberType
-	static public readwriteClassVariableName: memberType
+	static var _protectedClassVariableName: memberType
+	static var readonlyClassVariableName: memberType
+	static var public readwriteClassVariableName: memberType
 - A constructor method: >
 	def new(arguments)
 	def newName(arguments)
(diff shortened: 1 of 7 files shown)


runtime/doc/vim9class.txt
```diff
@@ -156,8 +156,8 @@ On the other hand, if you do not want the object variables to be read directly
 from outside the class or its sub-classes, you can make them protected.  This
 is done by prefixing an underscore to the name: >
 
-	this._lnum: number
-	this._col number
+	var_lnum: number
+	var_col number
```
yegappan: A space is needed between var and "_lnum" and "_col".

src/errors.h
```diff
@@ -3567,6 +3565,8 @@ EXTERN char e_using_class_as_var_val[]
 	INIT(= N_("E1405: Cannot use a Class as a variable or value"));
 EXTERN char e_using_typealias_as_var_val[]
 	INIT(= N_("E1406: Cannot use a Typealias as a variable or value"));
+EXTERN char e_static_must_be_followed_by_var_or_def[]
+	INIT(= N_("E1407: Static must be followed by \"var\" or \"def\""));
```
yegappan: Can you reuse E1368 instead of introducing a new error number?

src/testdir/test_vim9_class.vim
```diff
@@ -74,7 +74,7 @@ def Test_class_basic()
       this
     endclass
   END
-  v9.CheckSourceFailure(lines, 'E1317: Invalid object variable declaration: this', 3)
+  v9.CheckSourceFailure(lines, 'E1318: Not a valid command in a class: this', 3)
```
yegappan, line 104: Can you add a test for using the prefix "variable" instead of "var" for declaring an object variable?

src/testdir/test_vim9_class.vim
```diff
@@ -101,13 +101,13 @@ def Test_class_basic()
...
   END
-  v9.CheckSourceFailure(lines, 'E1317: Invalid object variable declaration: this. count', 3)
+  v9.CheckSourceFailure(lines, 'E1318: Not a valid command in a class: this. count', 3)
 
   # Use "that" instead of "this"
   lines =<< trim END
     vim9script
     class Something
-      this.count: number
+      var count: number
```
yegappan, line 137: Can you add a test for using the previous syntax using the "this" prefix with object and class variables and check whether they all fail?

runtime/doc/vim9class.txt
```diff
@@ -690,8 +690,8 @@ value for the object variables will be used.  This is a more useful example,
 with default values: >
 
 	class TextPosition
-	   this.lnum: number = 1
-	   this.col: number = 1
+	   var lnum: number = 1
+	   var col: number = 1
 	endclass
 
 If you want the constructor to have mandatory arguments, you need to write it
```
yegappan, line 697: The "Declaring object and class variables" section needs to be updated to reflect the rationale for using the "var" prefix now.

yegappan (changes requested):

yegappan:
I like the rationale for making the variable declaration consistent inside and outside a class.

chrisbra:
I agree. I think consistently requiring the use of `:var` makes sense

runtime/doc/vim9class.txt
dkearns: Done

src/errors.h
dkearns: Done

runtime/doc/vim9class.txt
dkearns: Done

src/testdir/test_vim9_class.vim
dkearns: Done

Shane-XB-Qian:
great! thank you very muuuuuuuuuuch:
seems you are fixing another it was an odd case: `this.`  :smile:
// (https://github.com/vim/vim/issues/11827#issuecomment-1383448606)

yegappan (approved):

Shane-XB-Qian:
vim9cmd var x = 123
in cmdline seems still cannot use 'var', is it designed, or forgot?

Shane-XB-Qian:
yea, so do you know what's the reason why it was disallowed? 
i felt it much useful if adhoc to run something..

Shane-XB-Qian:
> @Shane-XB-Qian, it looks like you've been here before. :smile:

:shrug: um... probably i may felt if scope was a matter, then why 'vim9cmd var t:x = 123' was not allowed?
i meant if `vim9cmd x = 123` err, but `vim9cmd t:x = 123` ok, then `vim9cmd var t:x = 123` should be ok too?
a.k.a here the latter one the 'var' keyword should be optional, if err, then err should be err about scope.
// BTW: var t:x = 123 also is disallowed in script (vs cmdline), i may recommend that 'var' keyword should be optional too.

-- 
shane.xb.qian

Shane-XB-Qian:
oh, one reason i can image perhaps because : to vim9, if 'var t:x' then x would a be type name.
if such, then let it be, how do you think?

Shane-XB-Qian:
but here it is 't:x' not 't: x', so i have no idea why..

chrisbra:
let's just leave it as it currently is.

Shane-XB-Qian:
> I thought it was a bit weird when I first saw it as well.  If you think of them more as explicit global dictionaries, as they're presented via `get()`, it's easier to understand.
> We could disallow type inference for these but then you have optional `var` (as the current mechanism has to be supported) _and_ disallowed type inference.
> Even if it was going to be supported it could be done in the future without any compatibility concerns.

ok, for this one, let's keep it as currently is for now,
// i may felt the 'var' should be optional, since the assignment is clear standed there,
// the err should be not about 'var'

-- 
shane.xb.qian

Shane-XB-Qian:
> Yes, the error is confusing.  Is "E1079: Cannot declare a script-local variable on the command line" good enough?

1, yes, i think so, "E1079: Cannot declare a vim9 script-local variable on the command line", named it with 'vim9' then better.

2, but 'var t:x = 123' is not allowed in script also, so another point perhaps should be making 'var' is an optional in this case.
// this one perhaps can be planed in the future.

-- 
shane.xb.qian

PR #13671, opened 2023-12-12
runtime(vim): Highlight multiline :syntax commands

dkearns:
Support multiline :syntax commands.

Match :syn-cchar option in :syn-{keyword,region}.


chrisbra:
thanks!

PR #13672, opened 2023-12-12
configure.ac: Free the allocated memory

zdohnal:
If the memory is not freed, GTK GUI VIM cannot be build with address sanitizer for debugging purposes - configure script will report missing GTK, because the testing file compilation fails due reported memory leak.


chrisbra:
thanks, makes sense.

PR #13673, opened 2023-12-12
label.yml: fixup for v5

pheiduck:


pheiduck:
@chrisbra when merged `sync-labels` should be `false`

k-takata:
As I wrote in #13664, I think there are some issue in labeler.yml, but let me merge this first to fix the CI error.

PR #13675, opened 2023-12-13
Prevent using `type` as function arg/return

errael:


errael:
@yegappan The first cut of this functionality, caught func arg/return at runtime. This PR keeps that in place, and also catches these (or at least some of these) at compile time.

chrisbra:
thanks, makes sense.

PR #13679, opened 2023-12-13
Support #elifdef and #elifndef in c.vim

### LINKED ISSUE #13667: Add support for #elifdef and #elifndef to c.vim

jwakely:
C2x and C++23 add these new preprocessor conditionals, as shorter forms of `#elif defined` and `#elif !defined`.

Fixes #13667

### DIFF
--- a/runtime/syntax/c.vim
+++ b/runtime/syntax/c.vim
@@ -380,7 +380,7 @@ if !exists("c_no_c99") " ISO C99
 endif
 
 " Accept %: for # (C99)
-syn region	cPreCondit	start="^\s*\zs\%(%:\|#\)\s*\%(if\|ifdef\|ifndef\|elif\)\>" skip="\\$" end="$" keepend contains=cComment,cCommentL,cCppString,cCharacter,cCppParen,cParenError,cNumbers,cCommentError,cSpaceError
+syn region	cPreCondit	start="^\s*\zs\%(%:\|#\)\s*\%(if\|ifdef\|ifndef\|elif\|elifdef\|elifndef\)\>" skip="\\$" end="$" keepend contains=cComment,cCommentL,cCppString,cCharacter,cCppParen,cParenError,cNumbers,cCommentError,cSpaceError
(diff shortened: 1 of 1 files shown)


jwakely:
This is the minimal change to make vim recognize them, so that syntax highlighting works, and you can jump between them using `%`. I have no idea if I should have also changed `cCppOutIf2`, `cCppInElse`, etc. because I don't know what they're used for.

jwakely:
Another way to define that regex would be:

    start="^\s*\zs\%(%:\|#\)\s*\%(\(el\)\?\(if\|ifdef\|ifndef\)\)\>"

Since each of `if`, `ifdef`, and `ifndef` now has an "elif" form, we can just give them all an optional `el` prefix.

jwakely:
> `cCpp{In,Out}*` syntax groups are used to highlight 'disabled' code as comments. So these changes should, ideally, be applied there as well. See `:help c_no_if0`

OK, I'll try to do that too ...

chrisbra:
please ping me, once this is ready.

chrisbra:
this seems to have stalled, so let me close it for now. Please create a new PR once you have a new version ready.

jwakely:
I didn't figure out how to make the requested changes, sorry.

N.B. both C23 and C++23 have now been published and the previous standards withdrawn. So `#elifdef` and #elifndef` are in the current ISO standards.

PR #13680, opened 2023-12-13
CI: Fix i386 build

k-takata:
Fix the following error:
```
The following packages have unmet dependencies:
 binutils : Conflicts: binutils:i386 but 2.38-4ubuntu2.4 is to be installed
 binutils:i386 : Conflicts: binutils but 2.38-4ubuntu2.3 is to be installed
 binutils-common : Breaks: binutils-common:i386 (!= 2.38-4ubuntu2.3) but 2.38-4ubuntu2.4 is to be installed
 binutils-common:i386 : Breaks: binutils-common (!= 2.38-4ubuntu2.4) but 2.38-4ubuntu2.3 is to be installed
 libbinutils : Breaks: libbinutils:i386 (!= 2.38-4ubuntu2.3) but 2.38-4ubuntu2.4 is to be installed
 libbinutils:i386 : Breaks: libbinutils (!= 2.38-4ubuntu2.4) but 2.38-4ubuntu2.3 is to be installed
 libctf0 : Breaks: libctf0:i386 (!= 2.38-4ubuntu2.3) but 2.38-4ubuntu2.4 is to be installed
 libctf0:i386 : Breaks: libctf0 (!= 2.38-4ubuntu2.4) but 2.38-4ubuntu2.3 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
```

Run `apt-get upgrade` before installing additional packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
               libattr1-dev
             )
           fi
-          sudo apt-get update && sudo apt-get install -y "${PKGS[@]}"
+          sudo apt-get update && sudo apt-get upgrade && sudo apt-get install -y "${PKGS[@]}"
 
       - name: Install gcc-${{ env.GCC_VER }}
         if:  matrix.compiler == 'gcc'


k-takata:
I'm thinking whether we should always run `apt-get upgrade` or only when `matrix.architecture` is i386...

chrisbra:
Ideally, this should be fixed by the Github runners, I created the following bug report: https://github.com/actions/runner-images/issues/9016

We had similar breakages, when Ubuntu updated some kernel modules, so perhaps running `apt-get upgrade` upgrade will fix this as well? It does seem to take 1.5 minutes longer however:
see here,
[This PR](https://github.com/vim/vim/actions/runs/7194270124/job/19594460334), `install packages`: `2m3s`, previously successful run, like [this one ](https://github.com/vim/vim/actions/runs/7158558062/job/19490792741): `install packages`: `30s`

k-takata:
> I'm thinking whether we should always run `apt-get upgrade` or only when `matrix.architecture` is i386...

I tried this in the new commit: https://github.com/vim/vim/pull/13680/commits/8c7933c1714889832bb77fc266579f3887c4eda8
Total duration is 16m 2s: https://github.com/vim/vim/actions/runs/7203798038

The previous commit took 17m 15s: https://github.com/vim/vim/actions/runs/7194270124

> so perhaps running `apt-get upgrade` upgrade will fix this as well?

Yes, I think so.

> like [this one ](https://github.com/vim/vim/actions/runs/7158558062/job/19490792741): `install packages`: `30s`

This example took 20m 54s in total: https://github.com/vim/vim/actions/runs/7194270124

Running `apt-get upgrade` makes the CI time about 1m longer, but other factors can affect more than that.
Maybe always running `apt-get upgrade` is better for avoiding package confliction?

k-takata:
I removed this commit: https://github.com/vim/vim/commit/8c7933c1714889832bb77fc266579f3887c4eda8

chrisbra:
perhaps we can also get rid of that work-around around next thursday, when new Github runner images are supposed to be released. Or we can just leave it as is and decide to not care :)

PR #13685, opened 2023-12-14
Not able to use imported interfaces and classes

### LINKED ISSUE #13661: Cannot use imported interface
### Steps to reproduce

https://github.com/vim/vim/assets/390059/4322e21e-9458-4987-b6e6-cb64d5438869

I may be doing something wrong here, but it seems when an interface is imported from another file, it cannot be implemented. 

### Expected behaviour

Should be able to use the imported after the interface has been exported properly from some library file.

### Version of Vim

9.0

### Environment

Mac OS, iTerm, xterm-256color-italic, ZSH

### Logs and stack traces

_No response_

yegappan:
Fixes #13661.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -3030,6 +3030,62 @@ def Test_class_import()
   v9.CheckScriptSuccess(lines)
 enddef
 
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #13688, opened 2023-12-15
Restrict builtin functions that accept a `type`

errael:
@chrisbra @yegappan This PR finishes off detection and prevention of using a
`type` as a value. It takes care of builtin functions. These are a wierd case
because some of these functions are part of the language and can handle a
`type`. In particular: `instanceof()`, `type()`, `typename()`, `string()`.

A "bit", `FE_X`, is added to funcentry_T; when set, the builtin function can
handle a `type`.

**notes/questions**
Had to set the `FE_X` bit for `test_refcount()` since it is used with a `type`.

Should other `test_*()` builtins have the `FE_X` bit set?

Discovered that in `compile_call()` the builtin `add()` is compiled inline
instead of calling the builtin. Had to add a check there.

### DIFF
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -1651,7 +1651,7 @@ typedef struct
     char	*f_name;	// function name
     char	f_min_argc;	// minimal number of arguments
     char	f_max_argc;	// maximal number of arguments
-    char	f_argtype;	// for method: FEARG_ values
+    char	f_argtype;	// for method: FEARG_ values; bits FE_
     argcheck_T	*f_argcheck;	// list of functions to check argument types;
 				// use "arg_any" (not NULL) to accept an
 				// argument of any type
@@ -1666,10 +1666,12 @@ typedef struct
 #define VARGS    CHAR_MAX
 
 // values for f_argtype; zero means it cannot be used as a method
-#define FEARG_1    1	    // base is the first argument
-#define FEARG_2    2	    // base is the second argument
-#define FEARG_3    3	    // base is the third argument
-#define FEARG_4    4	    // base is the fourth argument
+#define FEARG_1	    0x01    // base is the first argument
+#define FEARG_2     0x02    // base is the second argument
+#define FEARG_3     0x03    // base is the third argument
+#define FEARG_4     0x04    // base is the fourth argument
+#define FEARG_MASK  0x0F    // bits in f_argtype used as argument index
+#define FE_X	    0x10    // builtin accepts a non-value (class, typealias)
 
 #if defined(HAVE_MATH_H)
 # define MATH_FUNC(name) name
@@ -2181,7 +2183,7 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_inputsecret},
     {"insert",		2, 3, FEARG_1,	    arg23_insert,
 			ret_first_arg,	    f_insert},
-    {"instanceof",	2, VARGS, FEARG_1,  arg2_instanceof,
+    {"instanceof",	2, VARGS, FEARG_1|FE_X,	arg2_instanceof,
 			ret_bool,	    f_instanceof},
     {"interrupt",	0, 0, 0,	    NULL,
 			ret_void,	    f_interrupt},
@@ -2659,7 +2661,7 @@ static funcentry_T global_functions[] =
 			ret_number,	    f_strgetchar},
     {"stridx",		2, 3, FEARG_1,	    arg3_string_string_number,
 			ret_number,	    f_stridx},
-    {"string",		1, 1, FEARG_1,	    NULL,
+    {"string",		1, 1, FEARG_1|FE_X, NULL,
 			ret_string,	    f_string},
     {"strlen",		1, 1, FEARG_1,	    arg1_string_or_nr,
 			ret_number,	    f_strlen},
@@ -2821,7 +2823,7 @@ static funcentry_T global_functions[] =
 			ret_void,	    f_test_option_not_set},
     {"test_override",	2, 2, FEARG_2,	    arg2_string_number,
 			ret_void,	    f_test_override},
-    {"test_refcount",	1, 1, FEARG_1,	    NULL,
+    {"test_refcount",	1, 1, FEARG_1|FE_X, NULL,
 			ret_number,	    f_test_refcount},
     {"test_setmouse",	2, 2, 0,	    arg2_number,
 			ret_void,	    f_test_setmouse},
@@ -2853,9 +2855,9 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_trim},
     {"trunc",		1, 1, FEARG_1,	    arg1_float_or_nr,
 			ret_float,	    f_trunc},
-    {"type",		1, 1, FEARG_1,	    NULL,
+    {"type",		1, 1, FEARG_1|FE_X, NULL,
 			ret_number,	    f_type},
-    {"typename",	1, 1, FEARG_1,	    NULL,
+    {"typename",	1, 1, FEARG_1|FE_X, NULL,
 			ret_string,	    f_typename},
     {"undofile",	1, 1, FEARG_1,	    arg1_string,
 			ret_string,	    f_undofile},
@@ -2926,6 +2928,15 @@ static funcentry_T global_functions[] =
     {"xor",		2, 2, FEARG_1,	    arg2_number,
 			ret_number,	    f_xor},
 };
+
+/*
+ * Return true if specified function allows a type as an argument.
+ */
+    static int
+func_allows_type(int idx)
+{
+    return (global_functions[idx].f_argtype & FE_X) != 0;
+}
 
 /*
  * Function given to ExpandGeneric() to obtain the list of internal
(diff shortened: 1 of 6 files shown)


errael:
> Is there some way to feed a dynamically constructed 'list' of types to `instanceof`? `eval`ing doesn't seem to help as you can't store the intermediate results.

I've wondered about this kind of thing. I recall the suggestion to use string names (but I don't see where it came up). I keep coming back to `instanceof()` is part of the language. I also couldn't think of a use-case where you need/want to dynamically test a list of `type` for `instanceof()`. I only know how Java does it; what about other languages? The way things currently are, if you had to dynamically construct a list to check, you could have a list of lambdas, and run the ones you care about.

Note that with https://github.com/vim/vim/commit/2025af165ec68d831f0f0f668a3ceac3f39142ef `instanceof()` no longer takes a list. And there's the original discussion of `instanceof()` in #13421.

errael:
Actually, it looks like the eval trick works. Or maybe I'm missing something.

The following seems to do the right thing
```
vim9script

class C1
endclass
class C2
endclass
class C3
endclass
class C4
endclass

var o = C4.new()

echo eval("o->instanceof(C1, C2, C3)")
echo eval("o->instanceof(C1, C2, C3, C4)")
```

Shane-XB-Qian:
> Note that with https://github.com/vim/vim/commit/2025af165ec68d831f0f0f668a3ceac3f39142ef `instanceof()` no longer takes a list. And there's the original discussion of `instanceof()` in #13421.

the reason you claimed in #13421 about 'list' to 'varargs' maybe not accurate,
since in vim9 perhaps you noticed `[]` did not always mean `list`,
but/vs was a mark of `variable`, behaved like it was a var.
// but since #13421 had been merged, so..

-- 
shane.xb.qian

errael:
> since in vim9 perhaps you noticed `[]` did not always mean `list`, but/vs was a mark of `variable`, behaved like it was a var.

I don't understand when `[]` is not a list. Could you give an example.

src/evalfunc.c
```diff
@@ -1651,7 +1651,8 @@ typedef struct
     char	*f_name;	// function name
     char	f_min_argc;	// minimal number of arguments
     char	f_max_argc;	// maximal number of arguments
-    char	f_argtype;	// for method: FEARG_ values
+    char	f_argtype;	// for method: FEARG_ values; bits FE_
+				// Note: usused char.
```
yegappan: Typo: unused char?

src/evalfunc.c
```diff
@@ -3059,6 +3071,13 @@ internal_func_check_arg_types(
 	int	argcount,
 	cctx_T	*cctx)
 {
+    if (!(func_allows_type(idx)))
+    {
+        for (int i = 0; i < argcount; ++i)
```
yegappan, line 3077: Can you add a comment here?  Some internal functions accept types like Class as arguments.  For other functions, check the arguments are not types.

src/evalfunc.c
```diff
@@ -3059,6 +3071,13 @@ internal_func_check_arg_types(
...
+            if (check_type_is_value(types[i].type_curr) == FAIL)
+		return FAIL;
+    }
```
yegappan, line 3080: The above block of code is a duplicate of the one in check_args_for_type() except for the return value.  Can these two be merged?

src/evalfunc.c
```diff
@@ -3153,6 +3172,21 @@ check_internal_func(int idx, int argcount)
...
+    static int
+check_args_for_type(int idx, int argcount, typval_T *argvars)
+{
+    if (!func_allows_type(idx))
+    {
+	for (int i = 0; i < argcount; ++i)
+	    if (check_typval_is_value(&argvars[i]) == FAIL)
+		return FCERR_OTHER;
+    }
+    return FCERR_NONE;
```
yegappan: The FCERR_xxx values are used for error codes when calling a function.  Can you change this function to return OK or FAIL?

src/testdir/test_vim9_builtin.vim
```diff
@@ -5048,4 +5048,156 @@ def Test_writefile()
...
+  lines =<< trim END
+    vim9script
+    class C
+    endclass
+    def F()
+      empty(C)
+    enddef
+    F()
+  END
+  # TODO: SHOULD FAIL
```
yegappan: Are these TODOs still applicable?

yegappan (changes requested):

errael:
> Actually, it looks like the eval trick works. Or maybe I'm missing something.
> 
> The following seems to do the right thing
> 
> ```
> vim9script
> 
> class C1
> endclass
> class C2
> endclass
> class C3
> endclass
> class C4
> endclass
> 
> var o = C4.new()
> 
> echo eval("o->instanceof(C1, C2, C3)")
> echo eval("o->instanceof(C1, C2, C3, C4)")
> ```

Although the above works, it occurs to me it occurs to me that 
```viml
def G()
    var o1 = C4.new()
    echo eval("o1->instanceof(C1, C2, C3)")
enddef
G()
```
doesn't work because eval doesn't use a context (an `eval_in_context` would be nice). But
```viml
var Ftmpobj: any
def F()
    Ftmpobj = C4.new()
    echo eval("Ftmpobj->instanceof(C1, C2, C3)")
    echo eval("Ftmpobj->instanceof(C1, C2, C3, C4)")
enddef
F()
```
is a workaround.

@yegappan @dkearns  `instanceof()` could be changed to accept either a class/type or string. If needed, this could be added now or post vim9.1. (I still wonder if there's much of a use case)

src/evalfunc.c
errael: I'll remove it. Just noting an unused char in the struct on most any architecture.

src/evalfunc.c
errael: OK.

src/evalfunc.c
errael: Not easily. This is working with a `type2_T*`, but`check_args_for_type()` takes a `typeval_T*`.

src/evalfunc.c
errael: Changedl; I prefer OK/FAIL. (I thought checking args was part of calling a function).

src/testdir/test_vim9_builtin.vim
errael: Oops. Cleaned up.

src/testdir/test_vim9_typealias.vim
```diff
@@ -780,4 +780,137 @@ def Test_class_as_func_argument_or_return_value()
...
+  # check argument to empty in :def
+  lines =<< trim END
+    vim9script
+    type T = number
+    def F()
+      empty(T)
+    enddef
+    F()
+  END
+  # TODO: SHOULD FAIL
```
yegappan: These TODOs also need to be removed?

errael: Another one! Sigh. I finally tried grep. Think I caught them all.

errael:
@yegappan I left a TODO in vim9type.c, in the last function in the file. Any ideas on that?

src/vim9type.c
```diff
@@ -1886,38 +1888,21 @@ check_typval_is_value(typval_T *tv)
...
     {
         semsg(_(e_using_class_as_value_str), type->tt_class->class_name);
 	return FAIL;
     }
     else if (type->tt_type == VAR_TYPEALIAS)
     {
-	// Not sure what could be done here to get a name
-	// TODO: MAYBE AN OPTIONAL ARGUMENT
+	// TODO: Not sure what could be done here to get a name.
+	//       Maybe an optional argument?
```
yegappan, line 1901: The typealias_S structure has the typealias name.  A pointer to this is currently not stored in type_S. For a class, it is stored tt_class.  We need to add a variable (tt_typealias) to type_S to get to the name.

errael: Since you're requesting a change, I guess you want this change made.

I can do it, but I need some guidance. Where does this change go?

errael: And should `tt_class`/`tt_typealias` be in a union?

errael: There's also
```
vim9script
class C
endclass
type T = C
def F()
    abs(T)
enddef
F()
```
which outputs
```
E1405: Class "C" cannot be used as a value
```
Would/could adding a field "tt_typealias" help give an accurate message when using a typealias for a class?

yegappan: > Since you're requesting a change, I guess you want this change made.
> 
> I can do it, but I need some guidance. Where does this change go?

This change should be done in a separate follow-on PR.

errael: > > Since you're requesting a change, I guess you want this change made.
> > I can do it, but I need some guidance. Where does this change go?
> 
> This change should be done in a separate follow-on PR.

OK. I'll do "re-request review" to clear the "change requested" flag. And we can go from there.

yegappan (approved):

chrisbra:
Thanks. I agree it makes sense.

src/vim9type.c
yegappan: > The typealias_S structure has the typealias name. A pointer to this is currently not stored in type_S. For a class, it is stored tt_class. We need to add a variable (tt_typealias) to type_S to get to the name.

After looking into this some more, I realized that `type->tt_type` will never be set to VAR_TYPEALIAS.  In parse_type(), when parsing a type alias, the associated aliased type is returned (line 1402 in vim9type.c).

errael: I opened #13696 yesterday about having better error messages.

PR #13689, opened 2023-12-15
runtime: add filetype detection for execline

mazunki:
as a prior to adding syntax support for execline (see https://github.com/djpohly/vim-execline/issues/2), i went ahead and made the filetype detection for execline scripts.

i'm not entirely sure if calling the `SetExeclineFiletype` function on filenames matching all files (`*`) would be a better idea here, or if that's avoided for other filetypes due to performance reasons. it would make the code a bit more maintainable if that was done instead.


runtime/autoload/dist/ft.vim
```diff
@@ -263,6 +263,19 @@ export def EuphoriaCheck()
   endif
 enddef
 
+export def SetExeclineFiletype()
+  if getline(1) =~# '^#!.*\<execlineb\>'
```
dkearns: Shebang lines are tested in [script.vim](https://github.com/vim/vim/blob/master/runtime/autoload/dist/script.vim).

runtime/autoload/dist/ft.vim
```diff
@@ -263,6 +263,19 @@ export def EuphoriaCheck()
...
+    # shebang includes execlineb
+    setf execline
+  elseif expand('%:p') =~# 's6' && expand('%:t') =~# '^\(up\|down\|run\|finish\)$'
+    # path includes s6 somewhere, and filename is exactly one of the expected service runners
+    setf execline
+  elseif expand('%:t') =~# '^s6-'
+    # filename starts with s6-
+    setf execline
+  endif
```
dkearns: `autcmd` filename patterns are tested against the absolute file path if there's a `/` in the pattern.  So, you should be able to achieve these tests with patterns alone in `filetype.vim`.

dkearns (changes requested):

runtime/autoload/dist/ft.vim
mazunki: is it preferred to keep it in `filetype.vim`, instead of moving it into a standalone script? i moved it out mainly to keep `filetype.vim` as tidy as possible.

runtime/autoload/dist/ft.vim
mazunki: thanks, going to update it

runtime/autoload/dist/ft.vim
dkearns: I meant that you should be able to do it with `:autocmd` file patterns alone.  Something like:

```vim
au BufNewFile,BufRead *s6*/\(up\|down\|run\|finish\)    setf execline
au BufNewFile,BufRead s6-*                              setf execline
```

chrisbra:
please make those suggested changes.

runtime/autoload/dist/ft.vim
mazunki: oh, that's actually quite clean. thanks, i'll change it :)

mazunki:
Yeah, it kind of depends on user setup/preference. By default, there's a few paths with the `s6-` prefix. I went with `*s6*` as I can imagine a few people have paths such as /etc/s6 and/or `/var/lib/s6`, `/var/log/s6`, and/or `/run/s6`.

For what it's worth, there's `/run/service` and `/etc/init.d` as common paths too, but it's hard to capture all the potential paths.

runtime/autoload/dist/ft.vim
mazunki: Quick question: Does `*s6*/up` include a path such as `/etc/s6/sub/path/up` too? I imagine the `*` includes `/` too, but better to be safe than sorry.

dkearns: Yes, it does.

PR #13690, opened 2023-12-15
The options[] array is still not sorted alphabetically

zeertzjq:


chrisbra:
thanks, hopefully , the vim-win32-installer will now build again.

PR #13692, opened 2023-12-15
Add keymap completion

dkearns:
Problem:  Keymap completion is not available.
Solution: Add keymap completion.

Add keymap completion to the 'keymap' option, user commands and builtin
completion functions.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -3566,6 +3566,7 @@ getcompletion({pat}, {type} [, {filtered}])		*getcompletion()*
 		help		help subjects
 		highlight	highlight groups
 		history		|:history| suboptions
+		keymap		keyboard mappings
 		locale		locale names (as output of locale -a)
 		mapclear	buffer argument
 		mapping		mapping name
--- a/runtime/doc/map.txt
+++ b/runtime/doc/map.txt
@@ -1629,6 +1629,7 @@ completion can be enabled:
 	-complete=help		help subjects
 	-complete=highlight	highlight groups
 	-complete=history	:history suboptions
+	-complete=keymap	keyboard mappings
 	-complete=locale	locale names (as output of locale -a)
 	-complete=mapclear	buffer argument
 	-complete=mapping	mapping name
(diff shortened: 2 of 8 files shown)


yegappan (approved):

src/Makefile
```diff
@@ -955,6 +955,7 @@ TOOLS = xxd/xxd$(EXEEXT)
 #
 # Uncomment the next line to install Vim in your home directory.
 #prefix = $(HOME)
+prefix = $(HOME)/opt
```
benknoble: This probably isn't supposed to make it into the final version.

runtime/doc/builtin.txt
```diff
@@ -3566,6 +3566,7 @@ getcompletion({pat}, {type} [, {filtered}])		*getcompletion()*
 		help		help subjects
 		highlight	highlight groups
 		history		|:history| suboptions
+		keymap		keyboard mappings
```
benknoble, line 3569: A note that this is only available with `|+keymap|` or similar might be good.

runtime/doc/map.txt
```diff
@@ -1629,6 +1629,7 @@ completion can be enabled:
 	-complete=help		help subjects
 	-complete=highlight	highlight groups
 	-complete=history	:history suboptions
+	-complete=keymap	keyboard mappings
```
benknoble, line 1632: Ditto: a note that this is only available with `|+keymap|` or similar might be good.

src/Makefile
benknoble: Hm. Something is fishy: I don't seem able to mark this comment resolved. But it is. :)

runtime/doc/builtin.txt
dkearns: Thanks.  As it's not done for the other completion types in these lists I took the opportunity to deliberately not do so for this addition.

I'll look at updating them in a followup PR if I can discern a pattern in the help.

chrisbra:
okay thanks. However I'll move this to after the 9.1 release

PR #13694, opened 2023-12-15
Adjust deprecation notice wording on gdefault.

dundargoc:
Deprecated can be misunderstood as being slated for removal; slightly
change wording to be clearer.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -3838,9 +3838,9 @@ A jump table for the options with a short description can be found at |Q_op|.
 		:s///gg		  subst. all	  subst. one
 
 	NOTE: This option is reset when 'compatible' is set.
-	DEPRECATED: Setting this option may break plugins that are not aware
-	of this option.  Also, many users get confused that adding the /g flag
-	has the opposite effect of that it normally does.
+	Setting this option may break plugins that rely on the default
+	behavior of the `/g` flag. This will also make the /g flag have the
+	opposite effect of that documented in |:s_g|.
 	This option is not used in |Vim9| script.
 
 						*'grepformat'* *'gfm'*


errael:
> Deprecated can be misunderstood as being slated for removal

Why is that a problem?

chrisbra:
> Why is that a problem?

It's one of the obscure options, that many users (and plugin writers) are not aware of and tends to cause subtle problems.

Can we keep this one please?

> Also, many users get confused that adding the /g flag has the opposite effect of that it normally does.

I think this is important to keep stated.

dundargoc:
Done.

clason:
> I think this is important to keep stated.

Maybe we can reword this to be slightly less patronizing? (I'm assuming the audience here is plugin authors, in contrast to the previous sentence that warns _users_ about the effect on _plugins_?)

E.g.,

> This will also make the /g flag have the opposite effect of that documented in |:s_g|.

chrisbra:
that suggestion is fine as well 👍

dundargoc:
Fixed

chrisbra:
thanks

chrisbra:
I noticed, I forgot to ask you to update the Last Change Header, so I slightly fixed the wording and adjusted the last change header myself. Thanks!

errael:
`DEPRECATED` does not mean "it will be removed". It means **do not use**.

IMHO, weakening the warning by removing `DEPRECATED`, just reduces reliability.

errael:
> > Deprecated can be misunderstood as being slated for removal
> 
> Why is that a problem?

My response here was misunderstood. I was asking "why is it a problem if someone thinks the feature is slated for removal?".

Shane-XB-Qian:
that's on he (a people) stands. in *official vim*, that should be (ok) to be replaced by the term: 'Not recommend';
as for why? please check vim develop guide, that had said what vim is and what vim not, as long as a feat published, we may need to try best or make sure basically would Not break existed user usage.

--
shane.xb.qian

dundargoc:
> My response here was misunderstood. I was asking "why is it a problem if someone thinks the feature is slated for removal?".

Well, is it slated for removal? If not, then the problem is that it may not set correct expectations to users.

errael:
I'll address the most important issue first: reliability.

> > IMHO, weakening the warning by removing `DEPRECATED`, just reduces reliability.
> 
> I don't understand what you mean by text change reducing "reliability".
> The reliability...of what?

This phrase is in the before and after version of the help text
```
this option may break plugins
```
From this I discern that if "option" is used it reduces the overall
reliabity of **Vim** as experienced by users.

I admit that I do not read documentation as carefully as I might. I could
easily miss that phrase. But compare the before and after when viewing the
text in `:help`.

BEFORE
WithDeprecated

AFTER
WithoutDeprecated

Assuming someone looks at the docs, the before text makes it much more
likely that the user realizes there's a problem if the option is used.


> > `DEPRECATED` does not mean "it will be removed". It means **do not use**.
> 
> Not really, you just made that up. People use the term "deprecated" in a
> wide variety of ways, only one of them being the way you described.

In this line you say both
- "you just made that up"
- "People use the term ... one of them being the way you described"

These two statements are contradictory. An indication, at least to me, that
your purpose is to be argumentative and not really about making vim a
better tool.

And here's some things that showed up in a search I just made.
```
If I understand right, deprecated means it shouldn't be used anymore,
because it has been replaced by a better alternative
```

```
In information technology (IT), deprecation means that although something
is available or allowed, it is not recommended
```

```
Deprecated refers to a software or programming language feature that is
tolerated or supported but not recommended
```

It is true that some languages/systems use deprecated to indicate that
something will be removed in the future; but that is not what the word
means.

> Well, is it slated for removal? If not, then the problem is that it may
> not set correct expectations to users.

To what problem are you referring? To what expectation are you referring?
No matter how careful a documenter is, it is impossible to write anything
that can not be misunderstood.

The user should expect that if they use this feature they are making a
mistake and may well run into trouble.

dundargoc:
> These two statements are contradictory. 

They are not. That "deprecated" **definitively** means only what as you described is what I don't agree with.

> An indication, at least to me, that
your purpose is to be argumentative and not really about making vim a
better tool.

Yet I am the one who made a concrete proposal in the form of a PR, and you are the one arguing here. Curious.

If you disagree with the change that much, then please make followup/counter PR with the changes you'd like to see (if that's just a straight revert, then go for it). The change is already included, arguing here won't change that.

errael:
I did not say "**definitively** means".

Do you disagree that deprecated means do not use?

dundargoc:
> I did not say "**definitively** means".

Not in those exact words, but you sure seem adamant that your usage of the word is correct.

> Do you disagree that deprecated means do not use?

Yes, in the sense that it means *only* that with no implications of removal. But since I know others may not agree with this definition, I opted to just *explain* what the documentation actually wanted to say instead of demanding my usage of the word is The One True definition. If the all caps, eye grabbing warning is what you liked before then I'd suggest to just add a "DO NOT USE:" before. That would accomplish the exact same thing as DEPRECATED except for us not needing to have this discussion.

errael:
> add a "DO NOT USE:" before. That would accomplish the exact same thing as DEPRECATED

That's not entirely correct. It wouldn't have the highlighting; and I believe that's useful for pointing out something important. And I believe DEPRECATED, with its several connotations and common usage in technical documentation, is most appropriate.

dundargoc:
> That's not entirely correct. It wouldn't have the highlighting; and I believe that's useful for pointing out something important.

Sounds more to me there's a serious limitation in the documentation system if the content needs to be changed to achieve desried formatting.

> And I believe DEPRECATED, with its several connotations and common usage in technical documentation, is most appropriate.

Then please make a PR then with those changes. I don't have push rights to vim, so you're barking up the wrong tree.

Issue #10060, opened 2022-04-01
`:highlight clear` sets color scheme to `default`


Ace-Who:
### Steps to reproduce

    :highlight clear

### Expected behaviour

Don't change color scheme.

### Version of Vim

8.2.4602

### Environment

Windows 7, cmd.exe and gvim

brammool:
":hi clear" intentionally removes all non-default highlighting, that is what it's for.
If you want to re-apply the colorscheme (after you messed with the colors) use ":exe 'colors ' .. colors_name".

Ace-Who:
Thanks. Seems misguided by its behavior in an old version.

PR #13700, opened 2023-12-16
Only reload colorscheme when bg is changed

gpanders:
Currently the highlight groups are re-initialized and the colorscheme
(if any) is reloaded anytime 'background' is set, even if it is not
changed. This is unnecessary, because if the value was not changed then
there is no need to change highlight groups or do anything with the
colorscheme. Instead, only reload the colorscheme if the value of
'background' was actually changed.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -997,11 +997,11 @@ A jump table for the options with a short description can be found at |Q_op|.
 	what the background color looks like.  For changing the background
 	color, see |:hi-normal|.
 
-	When 'background' is set Vim will adjust the default color groups for
-	the new value.  But the colors used for syntax highlighting will not
-	change.					*g:colors_name*
+	When 'background' is changed Vim will adjust the default color groups
+	for the new value.  But the colors used for syntax highlighting will
+	not change.					*g:colors_name*
 	When a color scheme is loaded (the "g:colors_name" variable is set)
-	setting 'background' will cause the color scheme to be reloaded.  If
+	changing 'background' will cause the color scheme to be reloaded.  If
 	the color scheme adjusts to the value of 'background' this will work.
 	However, if the color scheme sets 'background' itself the effect may
 	be undone.  First delete the "g:colors_name" variable when needed.
--- a/src/optionstr.c
+++ b/src/optionstr.c
@@ -1050,10 +1050,14 @@ expand_set_ambiwidth(optexpand_T *args, int *numMatches, char_u ***matches)
  * The 'background' option is changed.
  */
     char *
-did_set_background(optset_T *args UNUSED)
+did_set_background(optset_T *args)
 {
     if (check_opt_strings(p_bg, p_bg_values, FALSE) == FAIL)
 	return e_invalid_argument;
+
+    if (args->os_oldval.string != NULL && args->os_oldval.string[0] == *p_bg)
+	// Value was not changed
+	return NULL;
 
 #ifdef FEAT_EVAL
     int dark = (*p_bg == 'd');


src/optionstr.c
```diff
@@ -1050,8 +1050,13 @@ expand_set_ambiwidth(optexpand_T *args, int *numMatches, char_u ***matches)
  * The 'background' option is changed.
  */
     char *
-did_set_background(optset_T *args UNUSED)
+did_set_background(optset_T *args)
 {
+    if (args->os_oldval.string[0] == *p_bg) {
```
zeertzjq, line 1055:
```suggestion
    if (args->os_oldval.string[0] == *p_bg)
    {
```

src/optionstr.c
```diff
@@ -1050,8 +1050,13 @@ expand_set_ambiwidth(optexpand_T *args, int *numMatches, char_u ***matches)
...
+	// Value was not changed
+	return NULL;
+    }
```
zeertzjq:
```suggestion
    if (args->os_oldval.string[0] == *p_bg)
	// Value was not changed
	return NULL;
```

src/optionstr.c
```diff
@@ -1050,8 +1050,14 @@ expand_set_ambiwidth(optexpand_T *args, int *numMatches, char_u ***matches)
...
  */
     char *
-did_set_background(optset_T *args UNUSED)
+did_set_background(optset_T *args)
 {
+    if (args->os_oldval.string[0] == *p_bg)
+    {
+	// Value was not changed
+	return NULL;
+    }
```
zeertzjq:
```suggestion
	// Value was not changed
	return NULL;
```

runtime/doc/options.txt
```diff
@@ -997,11 +997,11 @@ A jump table for the options with a short description can be found at |Q_op|.
...
 
-	When 'background' is set Vim will adjust the default color groups for
-	the new value.  But the colors used for syntax highlighting will not
-	change.					*g:colors_name*
+	When 'background' is changed Vim will adjust the default color groups
+	for the new value.  But the colors used for syntax highlighting will
+	not change.					*g:colors_name*
 	When a color scheme is loaded (the "g:colors_name" variable is set)
-	setting 'background' will cause the color scheme to be reloaded.  If
+	changing 'background' will cause the color scheme to be reloaded.  If
```
Shane-XB-Qian, line 1004: probably this is Not a good improvement,
which you may find a few places in vim was intended to do similar things like 'set bg=&bg'.

Shane-XB-Qian:
i think this maybe a break change, since classic vimer maybe used to like such tricky things.

src/optionstr.c
```diff
@@ -1050,8 +1050,12 @@ expand_set_ambiwidth(optexpand_T *args, int *numMatches, char_u ***matches)
  * The 'background' option is changed.
  */
     char *
-did_set_background(optset_T *args UNUSED)
+did_set_background(optset_T *args)
 {
+    if (args->os_oldval.string != NULL && args->os_oldval.string[0] == *p_bg)
+	// Value was not changed
+	return NULL;
+
```
chrisbra, line 1061: Can you please move this after the `check_opt_strings()` check below? I guess if you now use `:set bg=light | :set bg=lime`  it won't error out otherwise

runtime/doc/options.txt
chrisbra: You mean `:let &bg=&bg` ? and users rely on the fact, that this reloads the color scheme?  that sounds too obscure honestly.

src/optionstr.c
gpanders: Good catch, done.

runtime/doc/options.txt
Shane-XB-Qian: yes, 'let &bg=&bg', 
sometime you may not know or regardless current colorscheme was, but just want to refresh it.
and this is a simple tricky way to make it, and is a similar tricky in some vim classic usage, i think.

Shane-XB-Qian: yes, 'let &bg=&bg', 
sometime you may not know or regardless current colorscheme was, but just want to refresh it.
and this is a simple tricky way to make it, and is a classic similar tricky in some vim usage, i think.

clason: If "reload current colorscheme" is something that is useful enough, there should be an explicit way for it (e.g., `:colorscheme!`) without having to rely on a hack abusing unintended behavior. (And if it's not, no need to support it; you can use `g:colors_name`.)

gpanders: > yes, 'let &bg=&bg', sometime you may not know or regardless current colorscheme was, but just want to refresh it. and this is a simple tricky way to make it, and is a classic similar tricky in some vim usage, i think.

`exec 'colors' g:colors_name`

clason:
Most of these actually change the value of bg, though ("toggle background"), so wouldn't be affected by this change?

runtime/doc/options.txt
chrisbra: I would use `exe 'colors' get(g:, 'colors_name', 'default')` but yes, in general I agree, we shouldn't rely on such hacks.

chrisbra:
it seems to be slightly controversial. Even so I agree in general, I'll move it to after the 9.1 release.

runtime/doc/options.txt
Shane-XB-Qian: i knew above those  your concat ways, 
but come on, when you type cmd in cli, 
please be honest, would you really like to type so long and complex chars?

seandewar: > sometime you may not know or regardless current colorscheme was, but just want to refresh it.

Note that `:let &bg=&bg` does _not_ reload the default colorscheme (even if any `:highlight`s are modified) if `g:colors_name` was not set (like at startup), which may be undesirable if you want to reload "regardless [of what the] current colorscheme was".

> but come on, when you type cmd in cli,
> please be honest, would you really like to type so long and complex chars?

Optimizing for CLI typeability sounds very niche. Replicating the behaviour of `:exe 'colors' get(g:, 'colors_name', 'default')` with `:let &bg=&bg` would involve testing `g:colors_name` and executing `:colors default` if it's unset anyway. If the behaviour of skipping reloading the default colors is ever wanted, it can still be achieved with `:colorscheme`.

Compatibility seems like the only concrete concern to me (despite this, as an indicator, GitHub code searches produce very few examples of legitimate `:let &bg=&bg` usage).

Shane-XB-Qian: > > sometime you may not know or regardless current colorscheme was, but just want to refresh it.
> 
> Note that `:let &bg=&bg` does _not_ reload the default colorscheme (even if any `:highlight`s are modified) if `g:colors_name` was not set (like at startup), which may be undesirable if you want to reload "regardless [of what the] current colorscheme was".

colo 'default' is Not a colorscheme, it is a 'init', or a clean to reset!
and 'refresh' did not only mean refresh highlight, also if wish to re-kick 'au of colorscheme' etc.

> > but come on, when you type cmd in cli,
> > please be honest, would you really like to type so long and complex chars?
> 
> Optimizing for CLI typeability sounds very niche. Replicating the behaviour of `:exe 'colors' get(g:, 'colors_name', 'default')` with `:let &bg=&bg` would involve testing `g:colors_name` and executing `:colors default` if it's unset anyway. If the behaviour of skipping reloading the default colors is ever wanted, it can still be achieved with `:colorscheme`.
> 
> Compatibility seems like the only concrete concern to me (despite this, as an indicator, GitHub code searches produce very few examples of legitimate `:let &bg=&bg` usage).

you would write all your typed commands all wrote in github? if yes, show me.

Shane-XB-Qian: i believe the original author (which mainly was bram) had been thought that very carefully, 
'bg' was added since v7.0001, that tricky was there almost 20 years, i donot think any reason to break it.
// if there was a flaw, we can think to refine it, or another way, but break it i think it is not a good try/style.
at least official vim should not (though now it is on @chrisbra ), you can play your chaotic in neovim whatever.

seandewar: > colo 'default' is Not a colorscheme, it is a 'init', or a clean to reset!
> and 'refresh' did not only mean refresh highlight, also if wish to re-kick 'au of colorscheme' etc.

I interpreted your comment as referring to a color scheme in the general sense, but as you're talking about when a color scheme _file_ is not loaded (e.g: `g:colors_name` is unset), that can still be addressed using `:colorscheme` as referenced in my previous comment.

However, note that "colo 'default'" can also refer to the color scheme file at `colors/default.vim`, which uses the colors from `:hi clear`, and even [describes itself as a color scheme](https://github.com/vim/vim/blob/master/runtime/colors/default.vim#L6).

> you would write all your typed commands all wrote in github? if yes, show me.

[dkearns already responded to you earlier with two such links](https://github.com/vim/vim/pull/13700#issuecomment-1858817712), but it also includes results where a ternary is used to toggle the value of `&bg`, which is not affected by the proposed change; there are only 3 relevant hits.

[This link](https://github.com/search?q=path%3A*.vim+%2F%26%28bg%7Cbackground%29%5Cs*%3D%5Cs*%26%28bg%7Cbackground%29%5Cs*%24%2F&type=code) combines both queries while filtering out the results that toggle `&bg`.

> i believe the original author (which mainly was bram) had been thought that very carefully,
> 'bg' was added since v7.0001, that tricky was there almost 20 years, i donot think any reason to break it.

It's a niche behaviour that, with GitHub code search as an indicator, entertains little use (even anecdotally, I could only find [Bram suggesting the `:colorscheme` alternative](https://github.com/vim/vim/issues/10060#issuecomment-1086604996) in the past).

An example of how this change has been useful: Neovim, after startup, is now able to [always set `&bg` after finishing background detection](https://github.com/neovim/neovim/pull/26595#issue-2044421977), provided `&bg` was not set by the user, allowing `OptionSet background` autocommands to also be used by scripts to determine what value was detected.

In comparison, Vim can not currently do this without potentially reloading the color scheme, and [guards against the condition](https://github.com/vim/vim/blob/1c8e233cb8357b4f218bfa777f6d81456d0b43a2/src/term.c#L5758). The alternative is to manually parse `v:termrbgresp`, and copy [Vim's logic here](https://github.com/vim/vim/blob/1c8e233cb8357b4f218bfa777f6d81456d0b43a2/src/term.c#L5733-L5746) to determine if it would have resulted in an changed value of `&bg` or not.

> you can play your chaotic in neovim whatever.

Very classy.

Shane-XB-Qian: > > colo 'default' is Not a colorscheme, it is a 'init', or a clean to reset!
> > and 'refresh' did not only mean refresh highlight, also if wish to re-kick 'au of colorscheme' etc.
> 
> I interpreted your comment as referring to a color scheme in the general sense, but as you're talking about when a color scheme _file_ is not loaded (e.g: `g:colors_name` is unset), that can still be addressed using `:colorscheme` as referenced in my previous comment.
> 
> However, note that "colo 'default'" can also refer to the color scheme file at `colors/default.vim`, which uses the colors from `:hi clear`, and even [describes itself as a color scheme](https://github.com/vim/vim/blob/master/runtime/colors/default.vim#L6).

would you mind open that colors/default.vim file, see what content in there? 
or neovim had different content like official vim?

> > you would write all your typed commands all wrote in github? if yes, show me.
> 
> [dkearns already responded to you earlier with two such links](https://github.com/vim/vim/pull/13700#issuecomment-1858817712), but it also includes results where a ternary is used to toggle the value of `&bg`, which is not affected by the proposed change; there are only 3 relevant hits.
> 
> [This link](https://github.com/search?q=path%3A*.vim+%2F%26%28bg%7Cbackground%29%5Cs*%3D%5Cs*%26%28bg%7Cbackground%29%5Cs*%24%2F&type=code) combines both queries while filtering out the results that toggle `&bg`.

i guess @dkearns want to describe how many 'bg' usage in script?
anyway i am saying 'type', such quick cmd was easy to type in cmdline, would you write all your typed cmd into github? i donot think so.

---
again, `fix something while break something` was not offical vim's style, to get your demand of your change not must to break existed useful set, if it was a flaw had to be fixed, then please try just fix or another way, it had no reason to hurt other people which felt it useful, was it fair? fix Not equal to break it, unless it was not there at all or no user any more.

seandewar: > would you mind open that colors/default.vim file, see what content in there?
> or neovim had different content like official vim?

I linked the relevant line from Vim's `colors/default.vim` in my previous comment.

> anyway i am saying 'type', such quick cmd was easy to type in cmdline, would you write all your typed cmd into
> github? i donot think so.

In that case, I misinterpreted your response.
Regardless, I have already expressed my opinion that "optimizing for CLI typeability sounds very niche".

I understand Vim has a preference for compatibility, but again, not improving a behaviour because _it's easier to type on a CLI_ sounds unconvincing.

If you really have to type `if exists('g:colors_name')|exe 'colo' g:colors_name|en` to mimic the current behaviour of `:let &bg=&bg` so commonly that you wish to do less, there are many existing ways to achieve that (e.g: user-defined Ex-command, mapping, script you can easily `:so`, .etc).

> again, fix something while break something was not offical vim's style, to get your demand of your change not > must to break existed useful set, if it was a flaw had to be fixed, then please try just fix or another way, it had no > reason to hurt other people which felt it useful, was it fair? fix Not equal to break it, unless it was not there at all > or no user any more.

Yes, I am aware of Vim's preference for compatibility. I made that clear in my first comment as the only "concrete concern" against this change. Despite this, Vim has not been immune to incompatible changes (`:h incompatible-<Tab>`) in the past if seen fit.

There doesn't feel like there's any new points being brought up here. We clearly agree to disagree, which is OK.
Regardless, chrisbra has already stated his preference for this change; we'll see if that remains after 9.1.

dkearns: Sorry, I should have posted a better comment when I wasn't busy.  I was just attempting to indicate that there were a few examples out in the wild and, perhaps, it warranted further investigation.

Was this a promoted in the past as a "Vim Tip", or similar?

chrisbra:
Even so it seems controversial, let's include it. I think it is a good change and one shouldn't rely on the fact that `:let &bg=&bg` reloads the colorscheme (which would only happen, if one was loaded anyhow). We can re-consider, if anybody complains.

PR #13702, opened 2023-12-16
compiling defer fails in an if statement with false condition

### LINKED ISSUE #13698: [vim9script] unexpected behavior of the `defer` command
### Steps to reproduce

`vim --clean` and source this script
```
vim9script

def Hello()
  # const x = false
  if false
    defer execute('echow "hello"', "")
  endif
enddef

Hello()
```

Error `E129: Function name required` thrown. If literal `false` replaced with variable x and line above is uncommented, then function executes as expected.

Reported this in case some more serious bug involved here, because definitely no one should use such code.

### Expected behavior

No error should be thrown. Legacy vimscript works fine in such situation.

### Version of Vim

VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Dec 11 2023 15:41:16) Included patches: 1-2155

### Environment

os: Archlinux
terminal: st

### Logs and stack traces

_No response_

yegappan:
When skipping statements, don't check for types.  Fixes #13698.

### DIFF
--- a/src/testdir/test_vim9_script.vim
+++ b/src/testdir/test_vim9_script.vim
@@ -4783,6 +4783,19 @@ def Test_multidefer_with_exception()
   v9.CheckSourceSuccess(lines)
 enddef
 
+" Test for using ":defer" inside an if statement with a false condition
+def Test_defer_skipped()
+  var lines =<< trim END
+    def Foo()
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #13707, opened 2023-12-16
reg_executing() returns wrong result in :normal with range

zeertzjq:
Problem:  reg_executing() returns wrong result in :normal with range
          when 'showcmd' is set (after 8.2.4705).
Solution: Reset "pending_end_reg_executing" when executing a register.


chrisbra:
thanks!

PR #13712, opened 2023-12-17
runtime (Fortran) syntax update 

Ajit-Thakkar:
a) Update fortran syntax file to the latest ISO Fortran standard (Fortran 2023). b) Take advantage of the dropping of dialect support in the previous commit to streamline code. c) Minor improvements. d) Update  the fortran section of syntax.txt to reflect this change.


chrisbra:
The very first sentence in syntax.vim looks a bit strange, I have simplified it a bit.

Ajit-Thakkar:
Thanks.

runtime/syntax/fortran.vim
```diff
@@ -322,14 +301,13 @@ syn cluster fortranCommentGroup contains=fortranTodo
...
-    " but some vendor extensions allow longer lines
     if exists("fortran_extended_line_length")
+    " Vendor extensions allow lines with a text width of 132
       syn match fortranSerialNumber	excludenl "^.\{133,}$"lc=132
-    elseif exists("fortran_cardimage_line_length")
-      syn match fortranSerialNumber	excludenl "^.\{81,}$"lc=80
     else
-      syn match fortranSerialNumber	excludenl "^.\{73,}$"lc=72
+    " Standard requires fixed format to have a text width of 72,
+    " but all current compilers use 80 instead
```
uranix, line 309: > all current compilers use 80 instead

This is not the case for Intel Fortran Compiler. Both classic and LLVM compiler are using 72 for fixed-form by default with an option to extend it to 80 or 132.

PR #13713, opened 2023-12-17
runtime(erlang): Add support for matchit plugin

hcs42:
This commit updates the Erlang runtime files to be in sync with the `vim-erlang-runtime` repository. In particular, it adds the following commit: https://github.com/vim-erlang/vim-erlang-runtime/commit/6ea8b85bc9c93b94c68ec53489a74f5687d898b0 (a contribution from @subc2)

### DIFF
--- a/runtime/ftplugin/erlang.vim
+++ b/runtime/ftplugin/erlang.vim
@@ -5,7 +5,8 @@
 " Contributors: Ricardo Catalinas Jiménez <jimenezrick@gmail.com>
 "               Eduardo Lopez (http://github.com/tapichu)
 "               Arvid Bjurklint (http://github.com/slarwise)
-" Last Update:  2021-Nov-22
+"               Paweł Zacharek (http://github.com/subc2)
+" Last Update:  2023-Dec-20
 " License:      Vim license
 " URL:          https://github.com/vim-erlang/vim-erlang-runtime
 
@@ -57,7 +58,7 @@ setlocal suffixesadd=.erl,.hrl
 let &l:include = '^\s*-\%(include\|include_lib\)\s*("\zs\f*\ze")'
 let &l:define  = '^\s*-\%(define\|record\|type\|opaque\)'
 
-let s:erlang_fun_begin = '^\a\w*(.*$'
+let s:erlang_fun_begin = '^\l[A-Za-z0-9_@]*(.*$'
 let s:erlang_fun_end   = '^[^%]*\.\s*\(%.*\)\?$'
 
 if !exists('*GetErlangFold')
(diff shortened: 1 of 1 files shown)


runtime/ftplugin/erlang.vim
```diff
@@ -99,6 +100,23 @@ let b:undo_ftplugin = "setlocal keywordprg< foldmethod< foldexpr< foldtext<"
...
+    let s:sw = shiftwidth()
+  endif
+
+  let b:match_words =
+    \ '\<\%(begin\|case\|fun\|if\|maybe\|receive\|try\)\>' .
+    \ ':\<\%(after\|catch\|else\|of\)\>' .
+    \ ':\<end\>,' .
+    \ '^\l[A-Za-z0-9_@]*' .
+    \ ':^\%(\%(\t\| \{' . s:sw . '}\)\%([^\t\ %][^%]*\)\?\)\?;\s*\%(%.*\)\?$\|\.[\t\ %]\|\.$'
+  let b:match_skip = 's:comment\|string\|erlangmodifier\|erlangquotedatom'
```
dkearns: You should set `b:match_ignorecase` rather than rely on the `'ignorecase'` option.  See `:help b:match_ignorecase`.

These `b:match_*` variables should also be `:unlet` in the `b:undo_ftplugin` command.

runtime/ftplugin/erlang.vim
```diff
@@ -99,6 +100,23 @@ let b:undo_ftplugin = "setlocal keywordprg< foldmethod< foldexpr< foldtext<"
       \ . " comments< commentstring< formatoptions< suffixesadd< include<"
       \ . " define<"
 
+" The following lines enable the macros/matchit.vim plugin for
+" extended matching with the % key.
+if exists("loaded_matchit")
+  let s:sw = &sw
```
dkearns: This `s:sw` should be `:unlet` at some point, see `:help :unlet`.

> This is especially useful to clean up used global variables and script-local variables (these are not deleted when the script ends).

dkearns (changes requested):
Thanks for the PR.

runtime/ftplugin/erlang.vim
hcs42: Thank you for your advice. I added `b:match_ignorecase` and updated `b:undo_ftplugin`.

runtime/ftplugin/erlang.vim
hcs42: I added an `unlet` for `s:sw`.

chrisbra:
sorry, I did not notice earlier. Can you please make those two mentioned changes?

hcs42:
@chrisbra: I made those changes, and pushed them to `hcs42:master` just 20 minutes before your comment. I didn't set the two review comments to "Resolved" because I thought that the reviewer might want to do that when they are satisfied with the changes.

chrisbra:
yeah no problem. Can you please make the following changes in addition?



- the shiftwidht() function is available since v7.3.694. I think by now we can assume it is always available :) So can you please remove the exists('*shiftwidth') ?

- Also, the exists("loaded_matchit") depends on the fact, that the matchit plugin wasn't loaded by the time the erlang filetype plugin has been loaded. I think we can just always define those b:match_.... (matchit specific buffer variables).
Then it should also work, when a user loads matchit later manually.

hcs42:
I impemented @chrisbra 's suggestion. I saw @dkearns 's comment only afterwards. If you decide that it would be better to put back the `exists` guard in some form, just let me know.

chrisbra:
@dkearns thanks for letting me know. I'll await a new batch of updates then in the coming year :) 

In the meantime, I'll merge this now. Thanks everybody.

PR #13723, opened 2023-12-19
translation(ja): Update Japanese translations

k-takata:
ref: https://github.com/vim-jp/lang-ja/releases/tag/20231219


chrisbra:
thank you!

PR #13724, opened 2023-12-19
CI: change dependabot prefix to "CI"

zeertzjq:
The default prefix of dependabot is "build(deps)".  Change it to "CI".

Ref https://github.com/vim/vim/commit/c1c3b83816c8e22c9a710fb0c0cebc180ee6ce1e which used "CI" prefix.


chrisbra:
thank you. Makes sense obviously :)

PR #13725, opened 2023-12-19
runtime(doc): Update change.txt

k-takata:
Fix e06f2b498ccca921f34a1bec4464f042a5a2cabd.


k-takata:
The updated paragraph should be accurate than before. (I'm not sure whether this paragraph is useful, though.)

chrisbra:
Thanks! I think it is better than before. Please also update the last Change Header.

k-takata:
done

chrisbra:
thanks!

PR #13729, opened 2023-12-19
Add missing error messages.

errael:
The two lines in error don't output an error message
```
vim9script
class A
endclass

var o = A.new()
def F()
    A += 3  # produces generic message
    o += 4  # produces generic message
enddef
F()
```
Give this generic message
```
E1028: Compiling :def function failed
```

With this PR, get the messages
```
E1405: Class "A" cannot be used as a value
E1411: Missing dot after object "o"
```


chrisbra:
thanks, can you please add a test?

PR #13730, opened 2023-12-19
Added support to a modern C feature

alaanvv:
Removed the syntax line responsible by marking that as an error:

```c
someFunc((SomeStruct) { values });
// Specifically, marking the "{" and "}"
```

I think this should be default now

### DIFF
--- a/runtime/syntax/c.vim
+++ b/runtime/syntax/c.vim
@@ -174,7 +174,6 @@ else
     " cCppParen: same as cParen but ends at end-of-line; used in cDefine
     syn region	cCppParen	transparent start='(' skip='\\$' excludenl end=')' end='$' contained contains=ALLBUT,@cParenGroup,cErrInBracket,cParen,cBracket,cString,@Spell
     syn match	cParenError	display "[\])]"
-    syn match	cErrInParen	display contained "[\]{}]\|<%\|%>"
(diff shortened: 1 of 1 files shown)


chrisbra:
I guess you want: 
```
    let c_no_curly_error=1
```

> I think this should be default now

In what version is this no longer an error?

alaanvv:
Yes, I think it would be the same. I'm at the latest version.

Our solutions remove a problem but add a lot others, because we just want to remove the error highlight from that situation.
I don't really know how to write syntax files (just removed some lines until that was vanished), but it would be really good if just that specific error highlighting were removed, but keeping the mismatched curly braces.

chrisbra:
I meant specifically, for what Version of C is that allowed? We had in the past various discussions around this. I think it is hard to make c.vim act according to any C standard and some projects (e.g. Vim) would not use such constructs therefore it was made to display an error (and probably is wasn't allowed in older C Standards). If you want to allow this specifically for your projects, I think the 
```
:let g:c_no_curly_error=1
```
is the way to go. See `:h ft-c-syntax`

alaanvv:
Ah, the C version, this is called compound literal and was added in C99 (1999) according to [this website](https://en.cppreference.com/w/c/language/history). I don't think standards should affect syntax highlighting that way

PR #13734, opened 2023-12-20
runtime(masm): add variants of opcodes

adah1972:
that can actually be generated by compilers


adah1972:
When I added the opcodes, I mainly used Wikipedia as reference. It seems some variants were missing.

chrisbra:
thanks!

PR #13737, opened 2023-12-20
translation(sr): Update Serbian messages translation

eevan78:
Updated with newly added E1411


chrisbra:
Thanks. I noticed after merging, that the new error code will cause an issue for all translators. Sorry for late including this!

PR #13739, opened 2023-12-20

dkearns:
Match all ex-commands after ":" and the "|" command separator.

Exceptions are not handled yet and :insert/:change/:append are still not
matched after the command separator bar.


chrisbra:
thanks!

PR #13741, opened 2023-12-20
Ruby 3.3 change NoMethodError message format, and remove transient heap.

svardew:
Ruby 3.3 remove transient heap.
due to this change, +ruby/dyn vim cannot do :ruby command.
when vim load libruby.so, transient heap related function not found, cause E448 error.

Ruby 3.3 change NoMethodError message format.
due to this change, test_ruby.vim fails at Test_ruby_Vim_buffer_get().
pattern in assert_fails() missmatch v:errors, so test fail.

this PR fix:
if_ruby.c, disable rb_ary_detransient etc for ruby 3.3+
test_ruby.vim, make the pattern for NoMethodError looser

ruby/ruby Transient heap remove:
[issue](https://bugs.ruby-lang.org/issues/19730)
[commit](https://github.com/ruby/ruby/commit/1e7b67f73370056c7521306b9f037b6cbccf640d)
ruby/ruby NoMethodError message fomat:
[issue](https://bugs.ruby-lang.org/issues/18285)
[commit](https://github.com/ruby/ruby/commit/e7b8d32e166815f2e7edebf32aa178915d191b8c)
[commit](https://github.com/ruby/ruby/commit/1fd181b453a6a2dd16897473afa2f402c7dba8aa)


chrisbra:
thanks. It seems Ruby 3.3 is not yet released, so I'll put this into backlog for after the Vim 9.1 release

k-takata:
Ruby 3.3 is available now.

PR #13742, opened 2023-12-20
Reserve the double underscore prefix for class/object method names for future use

yegappan:
The PR https://github.com/vim/vim/pull/13238 for supporting builtin functions like len() for Vim9 objects is still under discussion.
For now, reserve the double underscore prefix for class/object method names for future use.  After the Vim9.1 release, we can
either decide to use the double underscore prefix or unreserve it.

While at this, also fix a code indentation issue.


chrisbra:
yes, thanks. Make sense

Issue #11770, opened 2023-01-01
[vim9] Equality is not transitive (when nulls are involved)


lifepillar:
### Steps to reproduce

Source this script:

```vim
vim9script 
echo null == null_string  # true
echo null_string == ''    # true
echo '' == null           # false

echo null == null_dict    # true
echo null_dict == {}      # true
echo {} == null           # false

# etc…
```

### Expected behaviour

Fundamental mathematical properties such as transitivity of equality should not be violated.

Mandatory [wat!?](https://www.destroyallsoftware.com/talks/wat) Jokes apart, couldn't be `null` comparisons be always made `false`? Or `null_string` be different from an empty string, `null_list` be different from an empty list, etc.?

More generally, are all those different types of nulls necessary? One of the design principles of Vim 9 script is to use conventions that are typical of other modern languages. It seems to me that that principle is violated in this case.

### Version of Vim

9.0.1117

### Environment

macOS 13.1
Apple Terminal
xterm-256color
ZSH 5.8.1

brammool:
> ### Steps to reproduce
> 
> Source this script:
> 
> ```vim
> vim9script 
> echo null == null_string  # true
> echo null_string == ''    # true
> echo '' == null           # false
> 
> echo null == null_dict    # true
> echo null_dict == {}      # true
> echo {} == null           # false
> 
> # etc…
> ```
> 
> ### Expected behaviour
> 
> Fundamental mathematical properties such as transitivity of equality
> should not be violated.

Well, would be nice, but this is a script language, mathematics is just
one of the things we might want to take into account.

> Mandatory [wat!?](https://www.destroyallsoftware.com/talks/wat) Jokes
> apart, couldn't be `null` comparisons be always made `false`? Or
> `null_string` be different from an empty string, `null_list` be
> different from an empty list, etc.?

An empty string and a null string are considered to have the same value.
Same for list and dict.  This is mainly so that uninitialized variables
are like "empty" variables.

If you care about the difference use the "is" operator.

> More generally, are all those different types of nulls necessary? One
> of the design principles of Vim 9 script is to use conventions that
> are typical of other modern languages. It seems to me that that
> principle is violated in this case.

Yes, they were added for a reason.

Changing this now most likely causes more trouble than it would solve.
If we could agree on how it should work anyway.  Before the Vim 9
release we might have considered changes, but now that users rely on the
current behavior any changes are going to create obscure problems.

-- 
"I love deadlines.  I especially like the whooshing sound they
[...]

Issue #13433, opened 2023-10-27
[vim9script] `typename()` gives incorrect/inconsistent output


errael:
### Steps to reproduce

This output
```
l: list<number>
l: list<unknown>
l: list<number>
o: object<Unknown>
o: object<C>
```
is generated by the following script
```
vim9script

class C
endclass

var l: list<number>
echo 'l:' typename(l)

l = null_list
# <<< expecte "list<number>" outputs "list<unknown>" # ERROR
echo 'l:' typename(l)

l = []
echo 'l:' typename(l)   # <<< outputs "list<number>"

#l = null_list
#l = ['a']       # <<< this gets "expected list<number>", the type is known

# Similar problem with objects
var o: C
echo 'o:' typename(o)   # output "object<Unknown>" expect "object<C>" # ERROR

o = C.new()
echo 'o:' typename(o)   # as expected: "object<C>"
```

### Expected behaviour

Correct/consistent type information.
```
l: list<number>
l: list<number>
l: list<number>
o: object<C>
o: object<C>
```

### Version of Vim

9.0.2077

### Environment

ubuntu

yegappan:
When you assign null_list to the variable `l`, the type of `l` changes to the type of null_list (which is list<unknown>).
To retain the variable type, but remove the list, you can assign it to an empty list.

```
var l: list<number> = [1, 2, 3]
l = []
echo typename(l)
```

errael:
> When you assign null_list to the variable `l`, the type of `l` changes to the type of null_list (which is list). To retain the variable type, but remove the list, you can assign it to an empty list.
> 
> ```
> var l: list<number> = [1, 2, 3]
> l = []
> echo typename(l)
> ```

Note that you get this errror
```
l: list<number>
l: list<unknown>
Error detected while processing :source buffer=1:
line   12:
E1012: Type mismatch; expected list<number> but got list<string>
```
from 
```
vim9script

class C
endclass

var l: list<number>
echo 'l:' typename(l)

l = null_list
echo 'l:' typename(l)

l = ['a']
```
So `l` does retain the type even as a `null_list`. The behavior we see, not outputting the correct type information, seems more an implementation artifact that has leaked into programmer visible space.

Assigning `[]` is not viable; I have code where an empty list is different from a null list.

errael:
[...]

PR #13750, opened 2023-12-22
Add `:help` for variable categories and null related.

errael:
Here's some documentation as planned in and around https://github.com/vim/vim/discussions/13458#discussioncomment-7568523.
Much of the info is taken from #11770.

@dkearns @lifepillar @yegappan @rhialto @Shane-XB-Qian @chrisbra and any others. Review appreciated for
1. accuracy of information
2. organization
3. wordsmithing
4. visuals
5. is there too much information?

Note: requests for changes in the currently implemented spec will be ignored (quietly or otherwise); this about documenting what exists in vim9.1.

### DIFF
--- a/runtime/doc/vim9.txt
+++ b/runtime/doc/vim9.txt
@@ -1,4 +1,4 @@
-*vim9.txt*	For Vim version 9.0.  Last change: 2023 Dec 09
+*vim9.txt*	For Vim version 9.0.  Last change: 2023 Dec 24
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
@@ -1055,8 +1055,11 @@ variable, since they cannot be deleted with `:unlet`.  E.g.: >
 
 The values can also be useful as the default value for an argument: >
 	def MyFunc(b: blob = null_blob)
-	   if b == null_blob
-	      # b argument was not given
+	    # Note: compare against null, not null_blob,
+	    #       to distinguish the default value from an empty blob.
+	    if b == null
+	        # b argument was not given
+See |null-compare| for more information about testing against null.
 
 It is possible to compare `null`  with any value, this will not give a type
 error.  However, comparing `null` with a number, float or bool will always
@@ -1697,6 +1700,155 @@ argument type checking: >
 			 *E1297* *E1298* *E1301*
 Types are checked for most builtin functions to make it easier to spot
 mistakes.
+
+Categories of variables, defaults and null handling ~
+				*variable-categories* *null-variables*
+There are categories of variables:
+	primitive	number, float, boolean
+	container	string, blob, list, dict
+	specialized	function, job, channel, user-defined-object
+
+When declaring a variable without an initializer, an explicit type must be
+provided. Each category has different default initialization semantics. Here's
+an example for each category: >
+	var num: number		# primitives default to a 0 equivalent
+	var cont: list<string>	# containers default to an empty container
+	var spec: job		# specialized variables default to null
+<
+Vim does not have a familiar null value; it has various null_<type> predefined
+values, for example |null_string|, |null_list|, |null_job|. Primitives do not
+have a null_<type>. The typical use cases for null_<type> are:
+- to `clear a variable` and release its resources;
+- as a `default for a parameter` in a function definition, see |null-compare|.
+
+For a specialized variable, like `job`, null_<type> is used to clear the
+resources. For a container variable, resources can also be cleared by
+assigning an empty container to the variable. For example: >
+	var j: job = job_start(...)
+	# ... job does its work
+	j = null_job	# clear the variable and release the job's resources
+
+	var l: list<any>
+	# ... add lots of stuff to list
+	l = []  # clear the variable and release container resources
+Using the empty container, rather than null_<type>, to clear a container
+variable may avoid null complications as described in |null-anomalies|.
+
+The initialization semantics of container variables and specialized variables
+differ. An uninitialized container defaults to an empty container: >
+	var l1: list<string>		    # empty container
+	var l2: list<string> = []	    # empty container
+	var l3: list<string> = null_list    # null container
+"l1" and "l2" are equivalent and indistinguishable initializations; but "l3"
+is a null container. A null container is similar to, but different from, an
+empty container, see |null-anomalies|.
+
+Specialized variables default to null. These job initializations are
+equivalent and indistinguishable: >
+	var j1: job
+	var j2: job = null_job
+	var j3 = null_job
+
+When a list or dict is declared, if the item type is not specified and can not
+be inferred, then the type is "any": >
+	var d1 = {}		# type is "dict<any>"
+	var d2 = null_dict	# type is "dict<any>"
+
+Declaring a function, see |vim9-func-declaration|, is particularly unique.
+
+						*null-compare*
+For familiar null compare semantics, where a null container is not equal to
+an empty container, do not use null_<type> in a comparison: >
+	vim9script
+	def F(arg: list<string> = null_list)
+	    if arg == null
+	       echo "null"
+	    else
+		echo printf("not null, %sempty", empty(arg) ? '' : 'not ')
+	    endif
+	enddef
+	F()		# output: "null"
+	F(null_list)	# output: "null"
+	F([])		# output: "not null, empty"
+	F([''])		# output: "not null, not empty"
+The above function takes a `list of strings` and reports on it.
+Change the above function signature to accept different types of arguments: >
+	def F(arg: list<any> = null_list)   # any type of list
+	def F(arg: any = null)		    # any type
+<
+In the above example, where the goal is to distinguish a null list from an
+empty list, comparing against `null` instead of `null_list` is the correct
+choice. The basic reason is because "null_list == null" and "[] != null".
+Comparing to `null_list` fails since "[] == null_list". In the following section
+there are details about comparison results.
+
+					*null-details* *null-anomalies*
+This section describes issues about using null and null_<type>; included below
+are the enumerated results of null comparisons. In some cases, if familiar
+with vim9 null semantics, the programmer may chose to use null_<type> in
+comparisons and/or other situations.
+
+Elsewhere in the documentation it says:
+	Quite often a null value is handled the same as an
+	empty value, but not always
+Here's an example: >
+	vim9script
+	var s1: list<string>
+	var s2: list<string> = null_list
+	echo s1		    # output: "[]"
+	echo s2		    # output: "[]"
+	
+	echo s1 + ['a']     # output: "['a']"
+	echo s2 + ['a']     # output: "['a']"
+	
+	echo s1->add('a')   # output: "['a']"
+	echo s2->add('a')   # E1130: Can not add to null list
+<
+Two values equal to a null_<type> are not necessarily equal to each other: >
+	vim9script
+	echo {} == null_dict      # true
+	echo null_dict == null    # true
+	echo {} == null           # false
+<
+Unlike the other containers, an uninitialized string is equal to null. The
+'is' operator can be used to determine if it is a null_string: >
+	vim9script
+	var s1: string
+	var s2 = null_string
+	echo s1 == null		# true - this is unexpected
+	echo s2 == null		# true
+	echo s2 is null_string	# true
+
+	var b1: blob
+	var b2 = null_blob
+	echo b1 == null		# false
+	echo b2 == null		# true
+<
+Any variable initialized to the null_<type> is equal to the null_<type> and is
+also equal to null. For example: >
+	vim9script
+	var x = null_blob
+	echo x == null_blob	# true
+	echo x == null		# true
+<
+An uninitialized variable is usually equal to null; it depends on its type:
+	var s: string		s == null
+	var b: blob		b != null   ***
+	var l: list<any>	l != null   ***
+	var d: dict<any>	d != null   ***
+	var f: func		f == null
+	var j: job		j == null
+	var c: channel		c == null
+	var o: Class		o == null
+
+A variable initialized to empty equals null_<type>; but not null:
+	var s2: string = ""	  == null_string	!= null
+	var b2: blob = 0z	  == null_blob		!= null
+	var l2: list<any> = []	  == null_list		!= null
+	var d2: dict<any> = {}	  == null_dict		!= null
+
+NOTE: the specialized variables, like job, default to null value and have no
+corresponding empty value.
 
 ==============================================================================
 
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -9081,7 +9081,11 @@ notepad	gui_w32.txt	/*notepad*
 nr2char()	builtin.txt	/*nr2char()*
 nroff.vim	syntax.txt	/*nroff.vim*
 null	vim9.txt	/*null*
+null-anomalies	vim9.txt	/*null-anomalies*
+null-compare	vim9.txt	/*null-compare*
+null-details	vim9.txt	/*null-details*
 null-variable	eval.txt	/*null-variable*
+null-variables	vim9.txt	/*null-variables*
 null_blob	vim9.txt	/*null_blob*
 null_channel	vim9.txt	/*null_channel*
 null_class	vim9.txt	/*null_class*
@@ -10944,6 +10948,7 @@ val-variable	eval.txt	/*val-variable*
 valgrind	debug.txt	/*valgrind*
 values()	builtin.txt	/*values()*
 var-functions	usr_41.txt	/*var-functions*
+variable-categories	vim9.txt	/*variable-categories*
 variable-scope	eval.txt	/*variable-scope*
 variable-types	vim9.txt	/*variable-types*
 variables	eval.txt	/*variables*


runtime/doc/vim9.txt
```diff
@@ -1698,6 +1698,140 @@ argument type checking: >
...
+	    if arg == null
+	       echo "null"
+	    else
+		echo printf("not null, %sempty", empty(arg) ? '' : 'not ')
+	    endif
+	enddef
+	F()		# output: "null"
+	F(null_list)	# output: "null"
+	F([])		# output: "not null, empty"
+	F([0])		# output: "not null, not empty"
```
Shane-XB-Qian: 1, though this work, and `is` is the right way to compare, but i think it is quite easy to write wrong or problem code.

Shane-XB-Qian:
2. as i concluded here https://github.com/vim/vim/issues/13639 , i may suggest 'var a: any' without assignment should default to 'null' (or 'any' maybe better), this better be modified before v9.1 released/published. @yegappan 
// for an instance what we faced, when decode a json to vim9 var, the unknown type string would be convert to 'null', so i think this should be similar situation, a 'var a: any' it's better be default to 'null' (compare to 'number' 0), it was 'unknown' situation too.

Shane-XB-Qian:
> 2\. this better be modified before v9.1 released/published. @yegappan

but this had been like that more than 3 years, 
i am a little hesitated to continue to push this,
besides, happy new year to your guys!
// perhaps just let it be~ :smile:

runtime/doc/vim9.txt
```diff
@@ -1055,7 +1055,7 @@ variable, since they cannot be deleted with `:unlet`.  E.g.: >
 
 The values can also be useful as the default value for an argument: >
 	def MyFunc(b: blob = null_blob)
-	   if b == null_blob
+	   if b == null
```
chrisbra: is that correct? According to the descriptions in `null-variables` , it should be an empty container, so `null_blob`? I would also think for clarity, the value should be compared against `null_blob`, since that's what we use as default argument explicitly

errael: > is that correct? According to the descriptions in `null-variables` , it should be an empty container, so `null_blob`? I would also think for clarity, the value should be compared against `null_blob`, since that's what we use as default argument explicitly

Take a look at the example at |null-compare|. The description of it says
```
For familiar null compare semantics, do not use null_<type> in a comparison,
then it easy to distinguish an empty container from a null container:
```
You may be assuming that equality is transitive.

More specifically to the exmple you bring up run, if I run
```
vim9script

def F(b: blob = null_blob)
    if b == null_blob
        echo "NO ARG"
    else
        echo "detected arg:" b
    endif
    echo "using blob:" b
enddef
var b0 = 0z
F()
echo ' '
F(b0)
```
The output is
```
NO ARG
using blob: 0z

NO ARG
using blob: 0z
```
Is that what you would expect? It looks very wrong to me.

Changing the compare to `b == null`, produces the output
```
NO ARG
using blob: 0z

detected arg: 0z
using blob: 0z
```

Maybe I should change the description at the example you're discussing.
But change it how? Maybe add "see |null-compare|" at a minimum?

errael:
@chrisbra I'm looking at making some tweaks. Could you clarify a few things about your comment.

> According to the descriptions in null-variables , **it** should be an empty container, so null_blob?

Wha/wheret exactly in `nulll-variables`? And what does **it** refer to?

> I would also think for clarity, the value should be compared against null_blob, since that's what we use as default argument explicitly

The declaration has `b: blob = null_blob`, it is a syntax error to do `b: blob = null`. Is this comment related to equality not being transitive? Or maybe it's about a null container usually acting like an empty container. And as noted in my previous response, comparing to `null_blob` gives unexpected results.

chrisbra:
Let's step back for a moment. I am slightly confused about this here:
```vim
vim9script

def F(b: blob = null_blob)
    if b == null
        echo "NO ARG"
    else
        echo "detected arg:" b
    endif
    echo "using blob:" b
enddef
F()
F(0z)
F(null_blob)
```

I would have expected, that the function compares against the default argument (e.g. `null_blob` instead of `null`) inside the function. Ah, it's because because `null == null_blob` implicitly. But perhaps for clarity, we should add a comment into the example, that we need to compare against `null` to be able to distinguish between `no argument` and an empty container. 

```vim
vim9script

def F(b: blob = null_blob)
# Note: we are not comparing against null_blob, but null
# in order to distinguish an empty blob from the default argument
    if b == null
        echo "NO ARG"
    else
        echo "detected arg:" b
    endif
enddef
F() # echo "NO ARG"
F(0z) # echo "detected arg"
# Also Note: null_blob will be handled like no argument
F(null_blob)  # echo "NO ARG"
```

does that make sense? If not please ignore my comment :)

Shane-XB-Qian:
> Is that what you would expect? It looks very wrong to me.

- @errael i think you were getting confusion to yourself,
you passed '0z' which is a "same" one like that arg claimed, so the result was not like your expected.

- and i think the instr `null-compare` you are now trying to add would make it more confused, in that case, `null_<type>` should be preferred way to compare, tho `null` is workable.
// so maybe just let it be, none maybe better than confusion.

- if you'd like to add instr, i may wish you can emphasize `var a: any` the type of `a` is 'number' (which i said in above comment regarding to #13639), that looked was different like most usual coding-langs we can see.

errael:
> > Is that what you would expect? It looks very wrong to me.
> 
>     * @errael i think you were getting confusion to yourself,
>       you passed '0z' which is a "same" one like that arg claimed, so the result was not like your expected.

@Shane-XB-Qian Perhaps you don't know what is being discussed.
Try `:help null`, go down about 30 lines and there is
```
	def MyFunc(b: blob = null_blob)
	   if b == null_blob
	      # b argument was not given
```
Note the comment `argument was not given`. If you do
```
MyFunc()            # detects no arg
MyFunc(0z)          # detects no arg
var some_blob: blob     # this creates an empty container, you can add to it
MyFunc(some_blob)   # detects no arg
```
All three calls to `MyFunc` detect `argument was not given`. If you change the
condition to `if b == null` then only `MyFunc()`, and not the other two calls,
detect `argument was not given`. This is the correct and expected behavior. The
original example is broken.

> 
>     * and i think the instr `null-compare` you are now trying to add would make it more confused, in that case, `null_<type>` should be preferred way to compare, tho `null` is workable.
>       // so maybe just let it be, none maybe better than confusion.

Wow, yesterday you commented https://github.com/vim/vim/pull/13750#discussion_r1434789002
> though this work, and is is the right way to compare, but i think it is quite
> easy to write wrong or problem code.

I think **vim9 nulls are confusing**, see #13458 where **most everyone confirms that**.
Leaving out the example, which you agree is the "right way to compare", would be
IMHO, a disservice to the vim9 programming community.

If a programmer chooses to use `null`s then they should have more information
about this difficult topic, not less.

errael:
Updated text. Essentially the same with additional comments and clarifications.

Shane-XB-Qian:
yes. it's confusing. and what I meant yesterday is: that your adding 'null-compare' i think it is quite easy to write wrong or problem code, though that programming semantic was correct, but at that situation, it should/better be use `null_<type>` to compare.
1. i do not have better wording to rephrase it,
2. it's truly there some odd difference (vs other langs), I wish to correct it, but it had been designed like that too long.
3. but consider legacy vim script usage and experience, it may be acceptable somehow though, for example '' and null sometime it was "same" thing.
4. maybe it was some advantage, and truly was, but just now legacy vs vim9 maybe a little difference made it a bit confusing.

--
shane.xb.qian

errael:
> yes. it's confusing. and what I meant yesterday is: that your adding 'null-compare' i think it is quite easy to write wrong or problem code, though that programming semantic was correct, but at that situation, it should/better be use `null_<type>` to compare.

Can we agree to disagree?

Using `null_<type>` gives broken code in the original vim9 example for detect "argument was not given". I do agree that it is "easy to write wrong code" when you want code that is correct in any situation.

Shane-XB-Qian:
I disagree. your modification made it worse.
I think I had explained you passed '0z' is not the right case to understand its purpose.

--
shane.xb.qian

errael:
These are the last changes I'll make.
Unless there's a request/suggestion for changes that I can't ignore ;-)

This "last" change adds the following at the end of the `null-compare` section.
LMK if you think adding this paragraph is not a good change.
```
In the above example, where the goal is to distinguish a null list from an
empty list, comparing against `null` instead of `null_list` is the correct
choice. The basic reason is because "null_list == null" and "[] != null".
Comparing to null_list fails since "[] == null_list". In the following section
there are details about comparison results.
```

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+	j = null_job	# clear the variable and release the job's resources
+
+	var l: list<any>
+	# ... add lots of stuff to list
+	l = []  # clear the variable and release container resources
+Using the empty container, rather than null_<type>, to clear a container
+variable may avoid null complications as described in |null-anomalies|.
+
+The initialization semantics of container variables and specialized variables
+differ. An uninitialized container default to an empty container: >
```
chrisbra: default**s** instead of default

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+When a list or dict is declared, if the item type is not specified and can not
+be inferred, then the type is "any": >
+	var d1 = {}		# type is "dict<any>"
+	var d2 = null_dict	# type is "dict<any>"
+
+Declaring a function, see |vim9-func-declaration|, is particularly unique.
+
+						*null-compare*
+For familiar null compare semantics, do not use null_<type> in a comparison,
+then it is easy to distinguish an empty container from a null container: >
```
chrisbra: > For familiar null semantics ... then it is easy

That sounds unusal/strange, perhaps better is to leave out the `then`?

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+
+A variable initialized to empty equals null_<type>; but not null:
+	var s2: string = ""	  == null_string	!= null
+	var b2: blob = 0z	  == null_blob		!= null
+	var l2: list<any> = []	  == null_list		!= null
+	var d2: dict<any> = {}	  == null_dict		!= null
+
+NOTE: the specialized variables, like job, default to null value and have no
+corresponding empty value.
+
```
chrisbra, line 1852: remove one empty line.

chrisbra:
Nice, I left a few minor commments, mainly style. Loos good, thanks!

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+<
+Any variable initialized to the null_<type> is equal to the null_<type> and is
+also equal to null. For example: >
+	vim9script
+	var x = null_blob
+	echo x == null_blob	# true
+	echo x == null		# true
+<
+An uninitialized variable is usually equal to null; it depends on its type:
+	var s: string		s == null
```
Shane-XB-Qian, line 1835: this maybe was a bug.

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+	echo {} == null_dict      # true
+	echo null_dict == null    # true
+	echo {} == null           # false
+<
+Unlike the other containers, an uninitialized string is equal to null. The
+'is' operator can be used to determine if it is a null_string: >
+	vim9script
+	var s1: string
+	var s2 = null_string
+	echo s1 == null		# true - this is unexpected
```
Shane-XB-Qian, line 1818: this maybe was a bug.
// should be 'false'

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+	var s1: list<string>
+	var s2: list<string> = null_list
+	echo s1		    # output: "[]"
+	echo s2		    # output: "[]"
+	
+	echo s1 + ['a']     # output: "['a']"
+	echo s2 + ['a']     # output: "['a']"
+	
+	echo s1->add('a')   # output: "['a']"
+	echo s2->add('a')   # E1130: Can not add to null list
```
Shane-XB-Qian, line 1805: this maybe was a bug.
should not change s2 elements' type from string to any.

runtime/doc/vim9.txt
```diff
@@ -1698,6 +1701,156 @@ argument type checking: >
...
+	    if arg == null
+	       echo "null"
+	    else
+		echo printf("not null, %sempty", empty(arg) ? '' : 'not ')
+	    endif
+	enddef
+	F()		# output: "null"
+	F(null_list)	# output: "null"
+	F([])		# output: "not null, empty"
+	F([0])		# output: "not null, not empty"
```
Shane-XB-Qian: wrong or typo.
// should be F([''])

runtime/doc/vim9.txt
Shane-XB-Qian: again, i may suggest rephrasing this wording, specially the `do not use`, i would prefer `null_<type>` somehow sometime, tho generally it was `null`.

Shane-XB-Qian:
i left 5 review comments:
one is wording, another one is typo or wrong code, and 3 i supposed were bugs.
// and such 3 "bugs" is not same like 'var a: any', which difficult to convince myself to get used to it. :-(

runtime/doc/vim9.txt
errael: Check

runtime/doc/vim9.txt
errael: The firs sentence of `null-compare` is reworded.

runtime/doc/vim9.txt
errael: Check

runtime/doc/vim9.txt
errael: Maybe, open an issue if you want. This is how it works in vim9.0.

runtime/doc/vim9.txt
errael: `s2` is of type `list<string>`. And see #13433

runtime/doc/vim9.txt
errael: Thanks

runtime/doc/vim9.txt
errael: Note that the first paragraph of `null-details` is
> This section describes issues about using null and null_<type>; included below
> are the enumerated results of null comparisons. **In some cases, if familiar
> with vim9 null semantics, the programmer may chose to use null_\<type> in
> comparisons and/or other situations**.

The idea is that most of this PR's help text is consistent on how to do compare.

errael:
These are the last changes I'll make. 
Unless there's a request/suggestion for changes that I can't ignore ;-)
Thanks to all for review.

The first sentence of `null-compare` is reworded:
```
For familiar null compare semantics, where a null container is not equal to
an empty container, do not use null_<type> in a comparison: >
```

runtime/doc/vim9.txt
Shane-XB-Qian: there no v9.0, v9.0 never published, only v8.2 and the coming v9.1
---
anyway, @yegappan the instr from bram just say non-init string should be 'empty', but emtpy did not mean 'null', so '' (emtpy string) should be more make sense?

runtime/doc/vim9.txt
Shane-XB-Qian: @yegappan if so, it will be make the comparsion of string more consistent like others.

runtime/doc/vim9.txt
Shane-XB-Qian: no idea what's that ticket you referred to.
but till now the latest dev ver of vim, the s2 is `list<any>` which manipulated its original type,
@yegappan i think it is a bug.

runtime/doc/vim9.txt
Shane-XB-Qian: i donot know if the users really understood you described terms and conditions here, at least to me, for this specific sentences, i would not accept your `do not use` suggestion, vs `null_<type>` still will be my choice.

Shane-XB-Qian:
> These are the last changes I'll make.
> Unless there's a request/suggestion for changes that I can't ignore ;-)
> Thanks to all for review.

what we joining (or just I) is for making vim9 script easy to use, donot get it wrong, it never was because whose ticket.
@chrisbra @yegappan the poetical bugs is described in above review comments, please think if worthy to fix, or after v9.1 published then perhaps too late to change.

chrisbra:
thanks. Let me merge it now. Null behaviour is indeed a bit confusing, so it is good to have the current behaviour documented.

Shane-XB-Qian:
if kept current behavior, i may suggest just keep current doc, if so, later had chance to rephrase or fix it.
// tho of course, all now it's your call.

Shane-XB-Qian:
ok, seems you would not change mind to revert merging, let me re-open mine ticket #13639 and refer here to continue to follow up.
// though not sure if someone would still be interested to continue to refine them in future.

runtime/doc/vim9.txt
yegappan: When a string variable is not initialized, memory is not allocated for the string.  Internally in the code, the string pointer in typval is NULL (https://github.com/vim/vim/blob/master/src/structs.h#L1629).  So the string is `null` and is not `empty`.

yegappan: > ## there no v9.0, v9.0 never published, only v8.2 and the coming v9.1

What do you mean by v9.0 was never published? See https://www.vim.org/vim90.php

runtime/doc/vim9.txt
yegappan: As described in the previous comment, when a string variable is not initialized, the memory for the string is not allocated.  So it is equal to null.

runtime/doc/vim9.txt
yegappan: The `s2` variable is explicitly initialized to a null list.  So you cannot add a new item to a null list.  If you initialize `s2` to an empty list instead of a null list, then you can add items to it.  The expression `s2 + ['a']` is not extending the null list, it is using it as a value.  A null list is considered to be an empty list when used as a value.

runtime/doc/vim9.txt
Shane-XB-Qian: > So the string is `null` and is not `empty`.

the 'empty' is not my word, it is bram's instr in the ':h e1134'.
i may feel you and errael thought it too c/c++ style, it is a scripting language, in script world, there is no 'pointer' and deconstructer, and if no init, usually to string it is a "".
// anyway, now i am not against it since i felt it is already difficult to change it, so i have opened PR #13769 to refine the doc a bit, that's all.

Shane-XB-Qian: > > ## there no v9.0, v9.0 never published, only v8.2 and the coming v9.1
> 
> What do you mean by v9.0 was never published? See https://www.vim.org/vim90.php

most recent linux distribution did not use v9.0 yet, it is still v8.2, vs v9.0 mostly was our dev ver.

runtime/doc/vim9.txt
Shane-XB-Qian: see my previous comment https://github.com/vim/vim/pull/13750#discussion_r1436115577

runtime/doc/vim9.txt
Shane-XB-Qian: IF it was not a bug, then i may think such `var s2: list<string> = null_list` should give it a err or warn, since it implicitly changed its element type from string to any, such statement became sick.
// or anyway, see my previous comment https://github.com/vim/vim/pull/13750#discussion_r1436115577

runtime/doc/vim9.txt
yegappan: Recent Linux distributions do use Vim 9.0.  Refer to the following page for the Ubuntu Vim package:

https://packages.ubuntu.com/search?keywords=vim

As you can see, starting with the lunar release, Ubuntu started using Vim 9.0.

Similarly, for the Debian Vim package, refer to https://packages.debian.org/search?keywords=vim.
Starting with the bookworm release, Vim 9.0 is used.

Fedora Linux also started using Vim9.0 from the Fedora 37 release.
See https://rpmfind.net/linux/rpm2html/search.php?query=vim-common.

MacOS ships with Vim 9.0.

Vim 9.0 is not a dev-only version.  It ships with all the major distributions.

Shane-XB-Qian: ok, fine, or whatever, i had said many times also worried it was too late to change also.
so anyway, see my previous comment https://github.com/vim/vim/pull/13750#discussion_r1436115577

PR #13751, opened 2023-12-22
Vim9: Coverity complains about not checking return value

chrisbra:
Problem:  Coverity complains about not checking return value
                in compare_isn_not_values (after 9.0.2184)
Solution: cast return value to "(void)" to make intention clear

### DIFF
--- a/src/vim9instr.c
+++ b/src/vim9instr.c
@@ -418,9 +418,9 @@ generate_two_op(cctx_T *cctx, char_u *op)
 compare_isn_not_values(typval_T *tv, type_T *type)
 {
     if (tv != NULL)
-	check_typval_is_value(tv);
+	(void)check_typval_is_value(tv);
     else
-	check_type_is_value(type);
+	(void)check_type_is_value(type);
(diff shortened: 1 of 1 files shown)


errael:
Oops. I use gcc at home. Is there an option to get a warning for this. There's might be more of these in stuff I recently added.

chrisbra:
Not sure if this is easily detectable by gcc. It comes through the Coverity static code scanners, that run for each build:
https://scan.coverity.com/projects/vim

If you have an account there, you could see it here: https://scan5.scan.coverity.com/reports.htm#v41242/p10101/fileInstanceId=256137292&defectInstanceId=53709748&mergedDefectId=1559861

If you need an account, I can grant you the membership and then you should get a notification for each warning it produces.

errael:
> If you need an account, I can grant you the membership and then you should get a notification for each warning it produces.

I'm not a member (AFAICT). Could you grant the membership. Don't want more noise, but I know how to use delete.

chrisbra:
I sent out an invite to your github email

PR #13756, opened 2023-12-23
Clarify that new() is not static.

### LINKED ISSUE #13146: [vim9script] Spec
### Steps to reproduce

In the spec
```
A class member declaration
    static this._privateMemberName: memberType
    static this.readonlyMemberName: memberType
    static public this.readwriteMemberName: memberType
```
Shouldn't have the word `this`.

There's
```
Note that constructors cannot be declared as "static",
because they always are.
```

That's not really correct; a constructor executes as a object method in the context of the newly created object. In vim9class, and other languages, they just have special invocation rules that use the class name. If they were static, they could not use "this".

### Expected behaviour

Accurate text.

### Version of Vim

9.0.1927

### Environment

ubuntu

### Logs and stack traces

_No response_

errael:
Closes #13146

This is the last piece of the issue.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -346,8 +346,8 @@ underscore as the first character in the method name: >
     endclass
 <
 							*E1370*
-Note that constructors cannot be declared as "static", because they always
-are.
+Note that constructors cannot be declared as "static". They are called like a
+static but execute as an object method; they have access to "this".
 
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. Can you pleas update the last change header?

chrisbra:
nevermind, I'll fix it locally

Issue #4537, opened 2019-06-14
":wincmd p" is ignored after closing current or previous window, when other windows are visible


fldef:
Closing the current or previous window may set `prevwin` to `curwin` or `NULL`, respectively.
There may be other windows that could be used as `prevwin` instead. (See tests below.)

Should `prevwin` point to a valid, different window in these cases?

Vim versions: 8.1.1486, 7.4
OS: Windows 10
Terminals: GUI, Windows console

```vim
func Test_window_prevwin_after_close_current()
  " start with one window, open two more, close the current one.
  split
  split
  close
  call assert_notequal(winnr('#'), winnr()) "fails
  call assert_notequal(winnr('#'), 0)
  let l:prevwin = winnr()
  wincmd p
  call assert_notequal(l:prevwin, winnr())  "fails: jumped to itself
  only!
  bwipe!
endfunc

func Test_window_prevwin_after_close_previous()
  " start with one window, open two more, close the previous one.
  split
  split
  exec winnr('#').'wincmd c'
  call assert_notequal(winnr('#'), winnr())
  call assert_notequal(winnr('#'), 0)      "fails
  let l:prevwin = winnr()
  wincmd p
  call assert_notequal(l:prevwin, winnr()) "fails: can't jump to window 0
  only!
  bwipe!
endfunc
```

chrisbra:
I am not sure. If you close the last accessed window, I would expect that `wincmd p` does not work anymore. Using another window would mean we need to keep track all previous windows and adjust accordingly. And even than it might be confusing to the user. I would leave it as it is currently.

PR #13762, opened 2023-12-24
Avoid `prevwin == curwin` when closing `curwin`

seandewar:
Problem: When closing the current window (or when moving it to a tabpage), the
previous window may refer to the new current window (`winnr() == winnr('#')`) if
that window is selected as the new current window (due to the layout).

Solution: Set `prevwin = NULL` when switching away from an invalid `curwin` and
the target window was the `prevwin`.

_Or maybe the behaviour should just be documented instead?_

---

#4537 briefly touched on the issue, but the answer didn't address the `prevwin == curwin` oddity.

Note that it may be more consistent to *always* clear `prevwin` when closing `curwin`, as closing
`curwin` must have meant we switched away from it to have a valid new `curwin`, making it the real
previous window.

However, that may be too aggressive; for example, one may want to jump to a different window, do stuff,
close it, then use `<C-W>p` to return to the original window (as the new curwin Vim chose after closing may
not be the original, depending on which window fills the space).

### DIFF
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -131,6 +131,64 @@ func Test_window_quit()
   bw Xa Xb
 endfunc
 
+func Test_window_curwin_not_prevwin()
+  botright split
+  call assert_equal(2, winnr())
+  call assert_equal(1, winnr('#'))
+  quit
+  call assert_equal(1, winnr())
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks. Let me put this after the vim 9.1 release

chrisbra:

PR #13766, opened 2023-12-25
Add the matchstrlist() and matchbufline() functions

yegappan:
Add a function to get all the matches in a string (similar to the Python re.findall() function).
Support matching a pattern in a List of strings.  This can be used to search for all the matches
in a file/buffer using readfile()/getbufline().

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -370,6 +370,8 @@ matchadd({group}, {pattern} [, {priority} [, {id} [, {dict}]]])
 matchaddpos({group}, {pos} [, {priority} [, {id} [, {dict}]]])
 				Number	highlight positions with {group}
 matcharg({nr})			List	arguments of |:match|
+matchbufline({buf}, {pat}, {lnum}, {end}, [, {dict})
+				List	all the {pat} matches in buffer {buf}
 matchdelete({id} [, {win}])	Number	delete match identified by {id}
 matchend({expr}, {pat} [, {start} [, {count}]])
 				Number	position where {pat} ends in {expr}
@@ -381,6 +383,8 @@ matchlist({expr}, {pat} [, {start} [, {count}]])
 				List	match and submatches of {pat} in {expr}
 matchstr({expr}, {pat} [, {start} [, {count}]])
 				String	{count}'th match of {pat} in {expr}
+matchstrlist({list}, {pat} [, {dict})
+				List	all the {pat} matches in {list}
 matchstrpos({expr}, {pat} [, {start} [, {count}]])
 				List	{count}'th match of {pat} in {expr}
 max({expr})			Number	maximum value of items in {expr}
@@ -6054,6 +6058,51 @@ matcharg({nr})							*matcharg()*
 
 		Can also be used as a |method|: >
 			GetMatch()->matcharg()
+<
+							*matchbufline()*
+matchbufline({buf}, {pat}, {lnum}, {end}, [, {dict}])
+		Returns the |List| of matches in lines from {lnum} to {end} in
+		buffer {buf} where {pat} matches.
+
+		{lnum} and {end} can either be a line number or the string "$"
+		to refer to the last line in {buf}.
+
+		The {dict} argument supports following items:
+		    submatches	include submatch information (|/\(|)
+
+		For each match, a |Dict| with the following items is returned:
+		    byteidx	starting byte index of the match
+		    lnum	line number where there is a match
+		    text	matched string
+		Note that there can be multiple matches in a single line.
+
+		This function works only for loaded buffers. First call
+		|bufload()| if needed.
+
+		When {buf} is not a valid buffer, the buffer is not loaded or
+		{lnum} or {end} is not valid then an error is given and an
+		empty |List| is returned.
+
+		Examples: >
+		    " Assuming line 3 in buffer 5 contains "a"
+		    :echo matchbufline(5, '\<\k\+\>', 3, 3)
+		    [{'lnum': 3, 'byteidx': 0, 'text': 'a'}]
+		    " Assuming line 4 in buffer 10 contains "tik tok"
+		    :echo matchbufline(10, '\<\k\+\>', 1, 4)
+		    [{'lnum': 4, 'byteidx': 0, 'text': 'tik'}, {'lnum': 4, 'byteidx': 4, 'text': 'tok'}]
+<
+		If {submatch} is present and is v:true, then submatches like
+		"\1", "\2", etc. are also returned.  Example: >
+		    " Assuming line 2 in buffer 2 contains "acd"
+		    :echo matchbufline(2, '\(a\)\?\(b\)\?\(c\)\?\(.*\)', 2, 2
+						\ {'submatches': v:true})
+		    [{'lnum': 2, 'byteidx': 0, 'text': 'acd', 'submatches': ['a', '', 'c', 'd', '', '', '', '', '']}]
+<		The "submatches" List always contains 9 items.  If a submatch
+		is not found, then an empty string is returned for that
+		submatch.
+
+		Can also be used as a |method|: >
+			GetBuffer()->matchbufline('mypat', 1, '$')
 
 matchdelete({id} [, {win})		       *matchdelete()* *E802* *E803*
 		Deletes a match with ID {id} previously defined by |matchadd()|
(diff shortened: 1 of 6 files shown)


runtime/doc/builtin.txt
```diff
@@ -6040,6 +6042,30 @@ matchaddpos({group}, {pos} [, {priority} [, {id} [, {dict}]]])
...
+		{list} is a |List| of strings.  {pat} is matched against each
+		string in {list}.
+
+		Returns a |List| where each item is a |List|.  Each item in
+		the inner list has two elements:
+			The starting index of the match.
+			The matched string.
+		The inner list item is empty if there is no match.  Example: >
+			:echo matchall(['abs', 'are'], 'a.*')
+			[[[0, 'abs']], [[0, 'are']]]
```
Shane-XB-Qian: return a 3 level deeper list, was it difficult to use?

Shane-XB-Qian:
was not it 'matchstr'? but just req a loop;
or anyway, could you make a 'string' can be as its first parameter too?

chrisbra:
Thanks, I think this is fine. Can you please also add a test using a zero-width regex atom?  Like `matchall(['foobar'], '\zs')`?

Shane-XB-Qian:
for instance, 'match()' also accept string or list as its first parameter.

yegappan:
> was not it 'matchstr'? but just req a loop;

If you call matchstr() to loop over a string multiple times to search for multiple matches and
loop over multiple strings, then the performance overhead will be high (there are two loops here).

> or anyway, could you make a 'string' can be as its first parameter too?

The argument to a built-in function should be a single type so that proper type
checking can be performed.  Before Vim9, many built-in functions accepted
different types for an argument and returned different types.  This made it
difficult to do proper type checking.  So the new built-in functions should start
using a single type for an argument and return a single type.

yegappan:
> Thanks, I think this is fine. Can you please also add a test using a zero-width regex atom? Like `matchall(['foobar'], '\zs')`?

Yes.  I have added this test.

Shane-XB-Qian:
> > was not it 'matchstr'? but just req a loop;
> 
> If you call matchstr() to loop over a string multiple times to search for multiple matches and
> loop over multiple strings, then the performance overhead will be high (there are two loops here).

i donot check the src code, but i guess it was using 'quick out', e.g 10th counter would quickly give-up previous 9 had done ones;
comparing to now this 3 level deepness list, user may still need to loop it.
but anyway, i am not against to add this func, but just say it had existed a similar one.

> > or anyway, could you make a 'string' can be as its first parameter too?
> 
> The argument to a built-in function should be a single type so that proper type
> checking can be performed.  Before Vim9, many built-in functions accepted
> different types for an argument and returned different types.  This made it
> difficult to do proper type checking.  So the new built-in functions should start
> using a single type for an argument and return a single type.

no offense, but i donot think so, you can think its first parameter was a 'any'.
// it should be make this func be more useful, if it can be a string also, then the returned can just a 2 level deepness list.

-- 
shane.xb.qian

yegappan:
> for instance, 'match()' also accept string or list as its first parameter.

The `match()` function was added a long time ago when it was acceptable for an argument to have multiple types.

runtime/doc/builtin.txt
bfrg: Shouldn't the returned list look like:
```
[[0, 'abs'], [0, 'are']]
```

bfrg:
>  This can be used to search for all the matches in a file/buffer using readfile()/getbufline().

But wouldn't a function like `searchbufline({pattern}, {buf}, {lnum}[, {end}])` be more efficient? Since it wouldn't require a call of `getbufline()` when the buffer has already been loaded.

runtime/doc/builtin.txt
yegappan: No.  The inner list is used because you may have multiple matches within a string.  Example:
```
   :echo matchall(['a1a2', 'xa3a4'], 'a.')
   [[[0, 'a1'], [2, 'a2']], [[1, 'a3'], [3, 'a4']]]
```
We can convert the two-item list into a Dict to avoid this confusion.

yegappan:
> > This can be used to search for all the matches in a file/buffer using readfile()/getbufline().
> 
> But wouldn't a function like `searchbufline({pattern}, {buf}, {lnum}[, {end}])` be more efficient? Since it wouldn't require a call of `getbufline()` when the buffer has already been loaded.
>

The searchbufline() function cannot be used with a list of strings generated
at runtime or received through a channel.  We then need to introduce two
separate functions.  One for searching buffers and one for searching in a list of strings.

> 
> Of course in that case the returned list would have to include the line number as well, like:
> 
> ```
> [[{lnum}, {col}, {match}], ...]
> ```

runtime/doc/builtin.txt
bfrg: > We can convert the two-item list into a Dict to avoid this confusion.

I think, mentioning it in the help is enough (with the above example).

yegappan:
> > This can be used to search for all the matches in a file/buffer using readfile()/getbufline().
> 
> But wouldn't a function like `searchbufline({pattern}, {buf}, {lnum}[, {end}])` be more efficient? Since it wouldn't require a call of `getbufline()` when the buffer has already been loaded.
> 
> Of course in that case the returned list would have to include the line number as well, like:
> 
> ```
> [[{lnum}, {col}, {match}], ...]
> ```

I have updated the PR to add the matchbufline() function which searches for a pattern
in the specified buffer and returns all the matches.

runtime/doc/builtin.txt
```diff
@@ -6055,6 +6059,43 @@ matcharg({nr})							*matcharg()*
 		Can also be used as a |method|: >
 			GetMatch()->matcharg()
 
+matchbufline({buf}, {pat}, {lnum} [, {dict}])		*matchbufline()*
+		Returns the |List| of matches in line {lnum} in buffer {buf}
+		where {pat} matches.
```
bfrg: Currently, it is not documented what happens when `matchbufline()` is invoked on a buffer that hasn't been loaded yet (or an invalid buffer number is passed). Shouldn't the help be similar to `:h setbufline()`?

This is from `:h setbufline()`:
```
This function works only for loaded buffers. First call
|bufload()| if needed.

...

When {buf} is not a valid buffer, the buffer is not loaded or
{lnum} is not valid then 1 is returned.  In |Vim9| script an
error is given.
```

yegappan: Yes.  This should be similar to the setbufline() function.  I have updated the code, test and the help text for an unloaded buffer.

chrisbra:
Thanks. I have one comment however. I just tried it out in a random help buffer:

    :echo matchbufline('%', 'Vim', 1, #{end_lnum: '$', submatches: v:true})

This returned a lot of empty lists (for non-matching lines), which I do not think is useful.

I think we may need something like this in addition, to only add matches to the resutling list:
```diff
diff --git a/src/evalfunc.c b/src/evalfunc.c
index eb0d5a6d7..6421228ba 100644
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -8226,13 +8226,14 @@ f_matchbufline(typval_T *argvars, typval_T *rettv)
        if (mlist == NULL)
            goto cleanup;

-       if (list_append_list(retlist, mlist) == FAIL)
-           goto cleanup;
-
        str = ml_get_buf(buf, slnum, FALSE);
        if (get_matches_in_str(str, &regmatch, mlist, slnum, submatches, TRUE)
                == FAIL)
            goto cleanup;
+
+       if (mlist->lv_len && list_append_list(retlist, mlist) == FAIL)
+           goto cleanup;
+
        slnum++;
     }
```

A similar thing applies to `matchstrlist()`, but there we do not know the index of the match, so not sure if we can somehow add the index for the matching list item or it is fine to keep empty lists for non-matching items.

yegappan:
> Thanks. I have one comment however. I just tried it out in a random help buffer:
> 
> ```
> :echo matchbufline('%', 'Vim', 1, #{end_lnum: '$', submatches: v:true})
> ```
> 
> This returned a lot of empty lists (for non-matching lines), which I do not think is useful.
> 
> 
> A similar thing applies to `matchstrlist()`, but there we do not know the index of the match, so not sure if we can somehow add the index for the matching list item or it is fine to keep empty lists for non-matching items.

Yes.  If a match is not found in a line or a string item, then the corresponding List item in
the returned List is empty.

I can modify this to return a flat list with one item for each match.  The returned match dict
item will contain either the line number (for matchbufline()) or the List index (for matchstrlist()).

yegappan:
> Thanks. I have one comment however. I just tried it out in a random help buffer:
> 
> ```
> :echo matchbufline('%', 'Vim', 1, #{end_lnum: '$', submatches: v:true})
> ```
> 
> This returned a lot of empty lists (for non-matching lines), which I do not think is useful.
> 

Note that `submatches` is needed only if sub-expressions are used in the pattern.  In the above
example, the pattern `Vim` doesn't contain any sub-expressions.  So `submatches` is not needed.

chrisbra:
> Note that submatches is needed only if sub-expressions are used in the pattern. In the above
example, the pattern Vim doesn't contain any sub-expressions. So submatches is not needed.

Yeah, I know. I was just playing around what would be happening :)

Thanks. Looks good now. I include it.

runtime/doc/builtin.txt
```diff
@@ -6054,6 +6058,51 @@ matcharg({nr})							*matcharg()*
...
+		{lnum} and {end} can either be a line number or the string "$"
+		to refer to the last line in {buf}.
+
+		The {dict} argument supports following items:
+		    submatches	include submatch information (|/\(|)
+
+		For each match, a |Dict| with the following items is returned:
+		    byteidx	starting byte index of the match
+		    lnum	line number where there is a match
+		    text	matched string
```
zeertzjq, line 6076: Is it intentional that spaces and non-breaking spaces are used alternate here and at several places below?

zeertzjq, line 6076: Is it intentional that spaces and non-breaking spaces are used alternately here and at several places below?

yegappan: No.  This is not intentional.  I cut and pasted the changes from one system to another and that resulted in this.  This needs to be fixed.

PR #13769, opened 2023-12-25
doc: rewording vim9 null description a bit

Shane-XB-Qian:
refine: https://github.com/vim/vim/pull/13750 a bit

### DIFF
--- a/runtime/doc/vim9.txt
+++ b/runtime/doc/vim9.txt
@@ -1757,8 +1757,7 @@ be inferred, then the type is "any": >
 Declaring a function, see |vim9-func-declaration|, is particularly unique.
 
 						*null-compare*
-For familiar null compare semantics, where a null container is not equal to
-an empty container, do not use null_<type> in a comparison: >
+It maybe a bit confusing to use null or null_<type> in a comparison: >
 	vim9script
 	def F(arg: list<string> = null_list)
 	    if arg == null
@@ -1771,16 +1770,17 @@ an empty container, do not use null_<type> in a comparison: >
 	F(null_list)	# output: "null"
 	F([])		# output: "not null, empty"
 	F([''])		# output: "not null, not empty"
+<
 The above function takes a `list of strings` and reports on it.
 Change the above function signature to accept different types of arguments: >
 	def F(arg: list<any> = null_list)   # any type of list
 	def F(arg: any = null)		    # any type
 <
-In the above example, where the goal is to distinguish a null list from an
-empty list, comparing against `null` instead of `null_list` is the correct
-choice. The basic reason is because "null_list == null" and "[] != null".
-Comparing to `null_list` fails since "[] == null_list". In the following section
-there are details about comparison results.
+NOTE: In the above example, the arg was assigned to a null container,
+comparing against `null` instead of `null_list` is ok. The basic reason is
+because "null_list == null" and "[] != null". Comparing to `null_list` fails
+since "[] == null_list". In the following section there are details about
+comparison results.
 
 					*null-details* *null-anomalies*
 This section describes issues about using null and null_<type>; included below
@@ -1804,6 +1804,9 @@ Here's an example: >
 	echo s1->add('a')   # output: "['a']"
 	echo s2->add('a')   # E1130: Can not add to null list
 <
+NOTE: the un-initialized var their type or value maybe unusual like other
+coding languages, specially e.g "var a: any" and "var s: string", see `E1134`.
+
 Two values equal to a null_<type> are not necessarily equal to each other: >
 	vim9script
 	echo {} == null_dict      # true


runtime/doc/vim9.txt
```diff
@@ -1757,8 +1757,7 @@ be inferred, then the type is "any": >
 Declaring a function, see |vim9-func-declaration|, is particularly unique.
 
 						*null-compare*
-For familiar null compare semantics, where a null container is not equal to
-an empty container, do not use null_<type> in a comparison: >
+It maybe a bit confusing to use null or null_<type> in a comparison: >
```
errael, line 1760: The programmer must decide which one to pick. Why do you take out the reasoning for which one to pick: `null` versus `null_<type>`? The question is: does the programmer want to distinguish null container from empty container. The difficulty in doing that is one of the main points of discussion in #13458.

runtime/doc/vim9.txt
```diff
@@ -1771,16 +1770,17 @@ an empty container, do not use null_<type> in a comparison: >
...
 	def F(arg: list<any> = null_list)   # any type of list
 	def F(arg: any = null)		    # any type
 <
-In the above example, where the goal is to distinguish a null list from an
-empty list, comparing against `null` instead of `null_list` is the correct
-choice. The basic reason is because "null_list == null" and "[] != null".
-Comparing to `null_list` fails since "[] == null_list". In the following section
-there are details about comparison results.
+NOTE: In the above example, the arg was assigned to a null container,
+comparing against `null` instead of `null_list` is ok. The basic reason is
```
errael, line 1780: This proposed statement is misleading. Using `arg == null`, and not `arg == null_list`, is required for the example; changing the compare changes the output. **This change breaks the example**.

The whole idea of the example is to show the programmer that there are issues with "null container" versus "empty container"; why do you take out the most important part of the description? That is what this is all about. If the programmer wants them to be different, **which is the way most (all?) strongly typed programming languages work**, the help text should describe how to do it.

runtime/doc/vim9.txt
```diff
@@ -1804,6 +1804,9 @@ Here's an example: >
 	echo s1->add('a')   # output: "['a']"
 	echo s2->add('a')   # E1130: Can not add to null list
 <
+NOTE: the un-initialized var their type or value maybe unusual like other
+coding languages, specially e.g "var a: any" and "var s: string", see `E1134`.
```
errael, line 1808: The example has `E1130` which is the same paragraph that you mention here `E1134`. This seems redundant.

This new `NOTE` doesn't belong right here. There is no `any` in the example.

As mentioned in #13433 typename can give confusing information. There may be a place in help to discuss this, but it is not at this specific location.

errael (changes requested):
Maybe a short sentence at the beginning of `null-compare` would help. Something like:
```
This section is primarily about detecting the difference between null
container and empty container; the reader can skip it if that is
unimportant to them.
```

runtime/doc/vim9.txt
Shane-XB-Qian: yes, it is their choice, no `do not use` such thing here.

runtime/doc/vim9.txt
Shane-XB-Qian: no, it is just to clear the leading and why that example came true.

runtime/doc/vim9.txt
Shane-XB-Qian: no, it is just explained or giving a note about why that Err happened.

Shane-XB-Qian:
> Maybe a short sentence at the beginning of `null-compare` would help. Something like:
> 
> ```
> This section is primarily about detecting the difference between null
> container and empty container; the reader can skip it if that is
> unimportant to them.
> ```

i donot mind if revert the whole https://github.com/vim/vim/pull/13750 merged, but it merged, 
then i think should explain the vim9 odd type situation a bit, see my previous comment https://github.com/vim/vim/pull/13750#discussion_r1436115577

Shane-XB-Qian:
which part was misleading?
I felt I was trying to correct the misleading.

this PR was just saying the situation itself (and the example came true based condition, and the reason of error), did not sell some preference, even it was, just was a 'recommend', vs/but not a 'do not use'.

--
shane.xb.qian

errael:
> which part was misleading? I felt I was trying to correct the misleading. 

In all strongly typed languages a null container is different from an empty container (AFAICT). In vim9script it is difficult to understand how to tell the difference between "null container" and "empty container". The purpose of the `null-compare` section and its example is to show how to tell the difference. 

This PR changes the `null-compare` section and takes out the words "null container" and "empty container" in the section's first sentence and takes out "null list"  and "empty list" after the example. These are the **most important words** to describe the problem and why the example is there.

Why does the PR take out those words? Taking out those words hides the reason for the example.

The example does not say "never use" it says "if you want null vs empty then do not use".

Unfortunately, I have no experience or training in writing documentation for readers who have English as a 2nd language.; my apologies  for not being able to communicate effectively with you.

> this PR was just saying the situation itself (and the example came true based condition, and the reason of error), did not sell some preference, even it was, just was a 'recommend', vs/but not a 'do not use'.
> […](#)
> -- shane.xb.qian

Shane-XB-Qian:
the modification of this PR was in the 'files' section of this page, I think most people know what changed, and the difference:
1. removed the 'do not use' and 'choose xxx is correct' such things.
2. explained why that 'null' (vs `null_<type>`) was used to compare in that example was that came true.
3. mentioned the vim9 odd type situation noted the user to be careful.
that's all.
by the way, do not be too confident on these terms what you made up, `null_<type>` and `list<string>` etc are Not some funny design.

and please try not barking English all the time in many tickets, you should have seen some people's response, where is your 'feeling good' (on English) came from? or again if you not mind, we can switch to use another language.

--
shane.xb.qian

errael:
> the modification of this PR was in the 'files' section...

The one thing I kept trying to find out: 
**Why were the "null container" versus "empty container" words taken out?** (since this is what the `null-compare` section is all about)

I use bold to call attention to something I believe is important to consider. (If that is what is meant by "barking")

> we can switch to use another language

Sadly, I'm only conversant in English. (and, for better or worse, English seems to be Vim's default language)

I'll no longer try to reconcile our differences. This is my last comment. Take the last word.

Shane-XB-Qian:
let it be, or take a relax, these are all about 'stands', but from tech perspective.
what was arguing, I was trying to reduce the misleading in the doc, though initially was tried to reduce odd buggy things in vim9 script.

happy new year!
tho to me, more important new year is one month later. :-)

the left things let chrisbra and yegappan (and ..?) to make. :smile:

--
shane.xb.qian

chrisbra:
I am sorry, I don't think this helps. Let me close this.

Shane-XB-Qian:
ok, i am not interested too.
// glad you had released v9.1, the left things supposed @yegappan would try to refine or rephrase in some kind ways later
// msg had sent, people will be looking this still (by search or something) anyway, wish it still be some kind useful unofficially! :smile:

PR #13771, opened 2023-12-25
runtime(elixir): fix indentation

georgeguimaraes:
Hi!

As of right now, Elixir is being indented (in insert mode) from:

```elixir
defmodule Foo do
  def bar do
    {# Cursor here
  end
end
```

to:

```elixir
defmodule Foo do
  def bar do
{# Cursor here
  end
end
```

That's because we need to remove `0{`and others from `indentkeys`. I also added some other useful indentation keys.

@mhanberg, what do you think? Should we move these into a new `runtime/indent/elixir.vim` file?

### DIFF
--- a/runtime/ftplugin/elixir.vim
+++ b/runtime/ftplugin/elixir.vim
@@ -1,7 +1,7 @@
 " Elixir filetype plugin
 " Language: Elixir
 " Maintainer:	Mitchell Hanberg <vimNOSPAM@mitchellhanberg.com>
-" Last Change: 2022 Sep 20
+" Last Change: 2023 Dec 26
 
 if exists("b:did_ftplugin")
   finish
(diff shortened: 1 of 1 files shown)


chrisbra:
Please update the last change header.

chrisbra:
Also @mhanberg do you approve?

mhanberg (approved):

chrisbra:
> This should, ideally, be moved to a dedicated indent plugin so that it can be managed via the :filetype command.

True. But let's leave that for later and merge it for now :) 

Thanks all.

chrisbra:
fixed it

georgeguimaraes:
> This should, ideally, be moved to a dedicated indent plugin so that it can be managed via the `:filetype` command.

Do you mean moving to `runtime/indent/elixir.vim`(and potentially a new `runtime/syntax/elixir.vim`)? Or do you mean an external plugin?

chrisbra:
no, basically moving the indent options to `runtime/indent/elixir.vim`

PR #13772, opened 2023-12-25
Add help tag multiple-constructors.

errael:
I have trouble finding this help text; probably because I start `:help new(`. Anyway...


errael:
Doh! Since I'm inclined to do `:help new(` first, putting links from there seems right. I'm assuming `new(` is more common than `construct`.

Making it the first line seems a little weird, but anywhere else interferes with the text's train of thought. The last line would work, but that is less useful.

chrisbra:
thanks!

Issue #13453, opened 2023-10-29
Vim crashes on Windows 11 arm64


wbpayne22902:
### Steps to reproduce

1. Compile vim on Windows 11 arm64 with Visual Studio and MSVC
2. Run it.
3. The terminal ends up with corruption and vim abnormally exits with a log of a crash placed in the Windows event log.

A crash log is as follows:
```Faulting application name: vim.exe, version: 9.0.0.0, time stamp: 0x653ea118
Faulting module name: vim.exe, version: 9.0.0.0, time stamp: 0x653ea118
Exception code: 0xc0000005
Fault offset: 0x000000000021d41c
Faulting process id: 0x0x1DC
Faulting application start time: 0x0x1DA0A95BE82E49C
Faulting application path: C:\tools\c\viminst\vim\src\vim.exe
Faulting module path: C:\tools\c\viminst\vim\src\vim.exe
Report Id: 45ce53a8-375e-4ff8-ae02-5f6142d77c9d
Faulting package full name: 
Faulting package-relative application ID: 
```

```Fault bucket 1744598794270781105, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: vim.exe
P2: 9.0.0.0
P3: 653ea118
P4: vim.exe
P5: 9.0.0.0
P6: 653ea118
P7: c0000005
P8: 000000000021d41c
P9: 
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER.d45d6800-e776-486c-ba68-da72de767334.tmp.WERInternalMetadata.xml

These files may be available here:
\\?\C:\ProgramData\Microsoft\Windows\WER\ReportArchive\AppCrash_vim.exe_aea879d8a26ec452e7c8965ef49ab4c8711edb_32c84bb0_78c8fc2c-acf2-4c6c-b23e-db9f204cb8e1

Analysis symbol: 
Rechecking for solution: 0
Report Id: 45ce53a8-375e-4ff8-ae02-5f6142d77c9d
Report Status: 2147487744
Hashed bucket: 09305fb3a3c1d5c648360f5c3015aeb1
Cab Guid: 0
```

![Screenshot 2023-10-29 142922](https://github.com/vim/vim/assets/107161609/43d9703d-2592-4e01-9afe-8de68e51ce6b)

### Expected behaviour

It shouldn't run and crash.

### Version of Vim

9.0.2081

### Environment

[...]

PR #13775, opened 2023-12-26
win32: add a workaround for a LTCG issue on Windows ARM64

compnerd:
Problem: LTCG would miscompile the `write_chars`
Solution: Explicitly initialise the storage to use data rather than BSS

It appears that the implicit initialisation which would push `g_coords` into BSS causes an aliasing issue with LTCG on ARM64.  By explicitly initialising the value, we use usual data storage but prevent the aliasing.  This allows the console version of VIM to run on Windows ARM64 again.

Issue: #13453


chrisbra:
oh wow. good analysis.

PR #13776, opened 2023-12-26
runtime(diff): Update default links

### LINKED ISSUE #13759: Update default links for `diffAdded` and `diffRemoved`

Right now highlight groups for `diff` filetype (used in patches) `diffAdded` and `diffRemoved` highlight groups are by default linked to [`Identifier`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/syntax/diff.vim#L383) and [`Special`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/syntax/diff.vim#L381). With time these choices seem to outgrow themselves as `diffAdded` is usually associated with green color while `diffRemoved` - with red, while current choices are rarely associated with these colors.

Another point for the change is that some of bundled color schemes override current default links with something a bit more green and red:
- [`blue`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/blue.vim#L79-L80), [`darkblue`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/darkblue.vim#L57-L58), and [`evening`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/evening.vim#L31-L32) use `String` and `WarningMsg` as targets. While the first one has more chance of being "green" in any given color scheme, the second one is usually orange or yellow (`ErrorMsg` is usually red, though).
- [`habamax`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/habamax.vim#L109-L110), [`lunaperche`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/lunaperche.vim#L167-L168), and [`wildcharm`](https://github.com/vim/vim/blob/6b9492e299c0b4ea76062160c4daf3985897f0b8/runtime/colors/wildcharm.vim#L96-L97) directly override groups to be directly green and red-ish.

Update default links:
- For `diffAdded` to link to a group with higher chance of being "green". Possible candidates: `String`, `DiffAdd`.
- For `diffRemoved` to link to a group with higher chance of being "red". Possible candidates: `ErrorMsg`, `DiffDelete`.

echasnovski:
Problem: Current default links for `diffAdded`, `diffChanged`, and
  `diffRemoved` do not address the diff nature of the filetype.

Solution: Use `DiffAdd`, `DiffChange`, and `DiffDelete`.

------

Resolves #13759. I also updated default link for `diffChanged`, as there is also a good `DiffChange` alternative.

### DIFF
--- a/runtime/syntax/diff.vim
+++ b/runtime/syntax/diff.vim
@@ -378,9 +378,9 @@ hi def link diffBDiffer		Constant
 hi def link diffIsA		Constant
 hi def link diffNoEOL		Constant
 hi def link diffCommon		Constant
-hi def link diffRemoved		Special
-hi def link diffChanged		PreProc
-hi def link diffAdded		Identifier
+hi def link diffRemoved		DiffDelete
+hi def link diffChanged		DiffChange
+hi def link diffAdded		DiffAdd
 hi def link diffLine		Statement
 hi def link diffSubname		PreProc
 hi def link diffComment		Comment


chrisbra:
thanks!

lilydjwg:
`Diff*` groups are documented as diff mode things. Perhaps the documentation should be updated?

echasnovski:
`Diff*` groups are still used as diff mode things. Now they are also used as default links for `diff*` groups. There is no reason to document it; same as with other default links. For example, it would be strange to also document that [`Constant` is used to display `diffCommon` by default](https://github.com/vim/vim/blob/b4ddc6c11e95cef4b372e239871fae1c8d4f72b6/runtime/syntax/diff.vim#L380).

lilydjwg:
```
DiffAdd		Diff mode: Added line. |diff.txt|
```

This sounds like it's only used in diff mode (and it was). I have a colorscheme which made this assumption and broke diff filetype highlighting.

On the other hand, `Constant` isn't tagged. It's below "group-name" rather than "hl-XXX". If you compare the two parts of the documentation, it's clear that "hl-XXX" part intends for Vim's interface, while the "group-name" one intends for filetype syntax definition.

chrisbra:
> DiffAdd		Diff mode: Added line. |diff.txt|

So `DiffAdd` is now used for Diff mode as well as for diff filetypes to mark lines that will be added. Similar for `DiffDelete` and `Diffchanged`. It seems similar enough, that I did not think documentation should be necessary. So how about the following then:

```patch
index 040780b35..b96978a3c 100644
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -5383,11 +5383,11 @@ CursorLine      Screen line that the cursor is in when 'cursorline' is set.
                                                        *hl-Directory*
 Directory      Directory names (and other special names in listings).
                                                        *hl-DiffAdd*
-DiffAdd                Diff mode: Added line. |diff.txt|
+DiffAdd                Diff mode and diff filetype: Added line. |diff.txt|
                                                        *hl-DiffChange*
-DiffChange     Diff mode: Changed line. |diff.txt|
+DiffChange     Diff mode and diff filetype: Changed line. |diff.txt|
                                                        *hl-DiffDelete*
-DiffDelete     Diff mode: Deleted line. |diff.txt|
+DiffDelete     Diff mode and diff filetype: Deleted line. |diff.txt|
                                                        *hl-DiffText*
 DiffText       Diff mode: Changed text within a changed line. |diff.txt|
                                                        *hl-EndOfBuffer*
```

echasnovski:
@chrisbra, adding information about default links is a bit misleading. The groups in 'diff' filetype are `diffAdded`, `diffRemoved`, and `diffChanged`; *these* are actually used and not `DiffAdd`, `DiffDelete`, `DiffChange`.

It might be better to document `diff*` groups directly. For example, by adding `*ft-diff-syntax*` in 'doc/syntax.txt', similar to [`*ft-tex-syntax*`](https://github.com/vim/vim/blob/80beeef0c6a4c44b190631725bce3dcc5635e3ee/runtime/doc/syntax.txt#L3317C19-L3317C32).

chrisbra:
I see. I'll add something later

ychin:
> This sounds like it's only used in diff mode (and it was). I have a colorscheme which made this assumption and broke diff filetype highlighting.

Just out of curiosity why would this change break diff filetype highlighting? Just the colors chosen doesn't match?

lilydjwg:
> Just out of curiosity why would this change break diff filetype highlighting? Just the colors chosen don't match?

In diff mode, DiffRemoved never has useful text (they are all `----`) so I made a dimmed color that was very similar to the background.

benknoble:
Possible problems were mentioned: https://github.com/vim/vim/issues/13759#issuecomment-1869503122

clason:
This is just about the default link; colorschemes can (and should) still style these however they wish.

echasnovski:
> Those are two different usages that require different stylings.

Nobody argues against that. The question is about default links for `diff*` highlight groups which would with higher probability result into what is usually considered "diff colors": green for `diffAdded`, red for `diffRemoved`, not green and not red for `DiffChanged`.

chrisbra:
It's true, the new default looks noisy. Not sure if we should make all color scheme authors additional work to fix this issue on their side.
Can we find a better default? that also fixes #13759 ? I am open to suggestions.

clason:
Note that _not_ having these links added more work for other colorscheme authors. No matter what you pick, it will be bad for some and better for others. If you want to be objective, you would need to count how many colorschemes (that do not set these directly -- which professional colorschemes should) profit from each default.

chrisbra:
Yeah also true. Whatever you do, you will choice wrong 🙈

Shane-XB-Qian:
really? if you are willing to listen, the solution had emphasis many times: Not `fixing something while break something`, bram chose compatible more than feature, and was Not so easy to break the used `default`, but you chose the work style like neovim.

--
shane.xb.qian

zeertzjq:
Hmm, maybe it's better to just set these to green and red directly instead of link them to some other groups? That may achieve the intended purpose for this PR better.

habamax:
> Hmm, maybe it's better to just set these to green and red directly instead of linking them to some other groups? That may achieve the intended purpose for this PR better than relying on assumptions about colorschemes.

FWIW, we can update bundled colorschemes and set greenish/reddish colors to diffAdded/diffRemoved.

romainl:
OK, let's rewind a little to the initial issue.

This is how the `diff` syntax looks in the following environment, which has been more or less the baseline for decades: 16 colors $TERM, default xterm palette, `default` colorscheme.

Capture d’écran 2024-01-04 à 19 22 53

That is rather obviously bad, if only because `diffRemoved` and `diffChanged` have the same color. Can we do better?

Yes we can, and it is easy to do it without breaking anything: specify `diffAdded`, `diffChanged`, and `diffRemoved` *explicitly* in `$VIMRUNTIME/syntax/syncolor.vim`:

Capture d’écran 2024-01-05 à 08 24 35

- The highlight groups in `diff` files have a nice *semantically correct* default in `default`,
- colorscheme authors and users can override them how they want,
- no one's setup is broken.

Now that we have seen the screenshot above, does the crap below *still* looks like a solution to anyone?

Capture d’écran 2024-01-04 à 19 25 14

chrisbra:
> really? if you are willing to listen, the solution had emphasis many times: Not `fixing something while break something`, bram chose compatible more than feature, and was Not so easy to break the used `default`, but you chose the work style like neovim.


Wow, thank you for this constructive feedback

chrisbra:
> Yes we can, and it is easy to do it without breaking anything: specify diffAdded, diffChanged, and diffRemoved explicitly in $VIMRUNTIME/syntax/syncolor.vim


So can you prepare a PR for this then?

romainl:
Yes, I can do it tonight (Europe/Paris time).

`red` for `diffRemoved` and `darkgreen` for `diffAdded` are a no-brainer, but feel free to suggest alternatives to `blue` for `diffChanged` in the mean time.

Shane-XB-Qian:
> > really? if you are willing to listen, the solution had emphasis many times: Not `fixing something while break something`, bram chose compatible more than feature, and was Not so easy to break the used `default`, but you chose the work style like neovim.
> 
> Wow, thank you for this constructive feedback

thanks, since you are saying you are `open to suggestions`, and I think most people do not like his/her/their (a people) `default` or used setup was becoming sick at next day wake up, so, sorry for my/this crap, and again. no offensive.

--
shane.xb.qian

vim-ml:
On 2024-01-04, Christian Brabandt wrote:
> Yeah also true. Whatever you do, you will choice wrong 🙈

I hadn't followed this discussion until just now when I did a git
commit and looked at the diff section.  It is butt ugly and
unreadable.

I use Vim to look at diffs daily, mostly in diff mode but often to
look at patches.  I have never run into any problems with the
previous defaults.

If some new application needs some new diff color thingy then add
something, but retain the existing defaults and don't make drastic
changes to defaults that had been working fine.

Regards,
Gary

chrisbra:
Yeah, I'll try to be more careful next time. Sorry for this. Hopefully we can improve this again.

echasnovski:
> I have never run into any problems with the
previous defaults.
> 
> ... but retain the existing defaults and don't make drastic
changes to defaults that had been working fine.

Having working fine for you doesn't necessarily mean that it is a proper choice of default links. Having default highlighting with quite low probability of being green-ish and red-ish is a bit odd based on establied UI colors.

> If some new application needs some new diff color thingy then add
something...

Same goes for users which care about their diff colors.

Shane-XB-Qian:
> > I have never run into any problems with the
> > previous defaults.
> > ... but retain the existing defaults and don't make drastic
> > changes to defaults that had been working fine.
> 
> Having working fine for you doesn't necessarily mean that it is a proper choice of default links.

~~crap?~~ it had been proper 20 years (https://github.com/vim/vim/pull/13700#discussion_r1429014637), now you are saying it's not proper? where were those tens of thousands of Vimer in that 20 years hiding, no one complain? // since IF the work looked like was basing on 'complain'.

clason:
You are welcome to disagree, but please stay civil.

> where were those tens of thousands of Vimer in that 20 years hiding

Probably just setting these groups explicitly in their (personal and plugin) colorschemes.

Shane-XB-Qian:
what is 'misleading'? that is the misleading.
if you thought it's not proper, should complain it at the beginning (like me "complain" vim9 script https://github.com/vim/vim/pull/13769) or when it still had chance, though i know it is still difficult, but later 20 years and after 3 major (v7/v8/v9) release, you want to break whole/most Vimer used way?

clason:
<plonk>

Shane-XB-Qian:
please help official vim keep consistent, thanks.
// as for this ticket itself, https://github.com/vim/vim/pull/13776#issuecomment-1877390894 had answer it.
// please revert this change, or @romainl https://github.com/vim/vim/pull/13776#issuecomment-1878264244

chrisbra:
@Shane-XB-Qian can you please calm down a bit? 

Please let's all stay civil, it doesn't help anybody to heat up this dicussion further. Let's see if we can fix this problem. I see @romainl already working on a fix. So let's be nice, test the suggested changes (once they are ready) give feedback and if it's fine, I'll merge it. 

Hope this is plan.

romainl:
@chrisbra I just submitted my idea of a fix for a) the original issue, and b) its original fix.

Shane-XB-Qian:
@chrisbra i am cool (tempeture 7℃) , just worried you were mislead-ed.
@romainl thanks.

PR #13778, opened 2023-12-26
runtime(context): update ConTeXt keywords and other minor fixes

lifepillar:
Update to the ConTeXt runtime files. Changes:

1. shared syntax files updated with `mtxrun --script interface --vim` using the latest ConTeXt LMTX.

2. fixed reference to `make` tag in the help file.

3. added `keepend` to mitigate issues with embedded Lua syntax (see below).

4. the latest revision date of each ConTeXt runtime file has been updated to the date of this commit.

The issue about embedded Lua was reported by a user:

>Take the following valid ConTeXt file:

>    `\starttext`
>    `\ctxlua{context("Text generated from Lua.")}`
>    `\ctxlua{context("Another text generated from Lua.")}`
>    `\stoptext`

>On my Vim installation (including when I start Vim with `--clean`), the
>closing bracket and curly braces on line 2 are highlighted red and the
>syntax highlighting after that is off.

>I was trying to dig a little bit into what was going on, using the
>`synID()` and `synIDattr()` functions. It appears that the closing
>bracket on line 2 is matched as a `luaParentError` instead of the end
>of the `luaParen` region. Therefore, the `luaParen` region continues
>all the way to the end of the file. The closing curly brace on line
>2 is matched as a `luaError`, the 2nd `\ctxlua` on line 3 as
>`luaParen`, etc.

>This issue doesn't occur in a plain Lua file, where the closing bracket
>is correctly matched as the end of the `luaParen` region. So it seems
>that something goes wrong when the Lua syntax file is included in the
>ConTeXt one.

By adding `keepend`, the right parenthesis for some reason is still highlighted as a `luaParenError`, but at least the right curly brace should correctly end the Lua block.

From what I've seen, I think it is very difficult to embed Lua syntax properly without help from the Lua syntax file (that is, without patching it). It has global rules such as:

    syn match  luaParenError ")"
    syn match  luaError "}"

which make it difficult, if not impossible, to contain Lua syntax without `keepend` (and its limitations).


chrisbra:
Thank you!

PR #13785, opened 2023-12-28
Wrong display when 'briopt' contains "sbr" and 'nobreakindent'

zeertzjq:
Problem:  Wrong display when 'breakindentopt' contains "sbr" and
          'showbreak' and 'nobreakindent' are set.
Solution: Always reset wlv->need_showbreak regardless of the values of
          'breakindent' and 'showbreak', as they aren't checked when
          setting wlv->need_showbreak.


zeertzjq:
Another solution is to check the value of `w_p_bri` before checking the value of `w_briopt_sbr`, but that won't be necessary if `wlv->need_showbreak` is update consistently, and I think it's a bit strange for `wlv->need_showbreak` to always remain true when some options aren't set.

chrisbra:
thanks

PR #13790, opened 2023-12-28
Filelist fixes

RestorerZ:
File `coveralls.yml` removed from `$(SRC_ALL)`, see patch 9.0.1752 (#12851)
`runtime/lang/Make_mvc.mak` moved from `$(LANG_DOS)` to `$(RT_DOS)`, inconsistency with destination


chrisbra:
thanks!

PR #9477, opened 2022-01-05
Update Git runtime files

tpope:
These are the latest versions of the `git*.vim` runtime files that I maintain. Note that I am requesting **removal** of `ftplugin/git.vim`, as I believe everything in there was of extremely dubious value.

I am also including a few tweaks to filetype detection:

* Detect the files created by `git notes edit` and `git branch --edit-description` as `gitcommit`.
* Loosen the 40 character SHA-1 matches to account for upcoming SHA-256 support.
* Remove an extremely old rule for detecting `gitsendemail`. (Git switched to `.gitsendemail.msg.*` in 2008.)

PR #13794, opened 2023-12-28
Updates to tpope-maintained runtime files

tpope:
This is everything except my Ruby files, which are coordinated separately.

Previously in #9477 I removed `ftplugin/git.vim` because it wasn't doing anything useful. This brings it back to do one useful thing (`setlocal nomodeline`).


chrisbra:
Thanks!

PR #13795, opened 2023-12-28
Bring up to date the :public example

zzzyxwvut:
The current example is no longer supported (`E1331`), as of
patch `9.0.2167`.


chrisbra:

PR #13799, opened 2023-12-29
Check for printf argument overflow

cvwillegen:
Check arguments for the printf function if the width or positional argument will overflow.


chrisbra:
thanks. I'll move this to after the vim 9.1 release however

chrisbra:
what is the magic `6400` here used? Can we make this a define and add a short description why it is 6400?

chrisbra:
is this ready now?

vim-ml:
On Sat, Mar 9, 2024 at 6:06 PM Christian Brabandt ***@***.***>
wrote:

> is this ready now?
>

Yes. Or at least, ready for review :-)

Christ van Willegen

src/strings.c
```diff
@@ -2468,6 +2468,55 @@ adjust_types(
...
+    while (VIM_ISDIGIT((int)(**p)) && *uj < MAX_ALLOWED_STRING_WIDTH)
+    {
+	*uj = 10 * *uj + (unsigned int)(**p - '0');
+	++*p;
+    }
+
+    if (*uj > MAX_ALLOWED_STRING_WIDTH)
+    {
+	format_overflow_error(pstart);
+	return -1;
```
chrisbra: return FAIL

src/strings.c
```diff
@@ -2468,6 +2468,55 @@ adjust_types(
...
+    }
+
+    return 0;
```
chrisbra: return OK;

src/strings.c
```diff
@@ -2531,10 +2581,11 @@ parse_fmt_types(
 		}
 
 		// Positional argument
-		unsigned int uj = *p++ - '0';
+		unsigned int uj;
+
+		if (get_unsigned_int(pstart, &p, &uj) == -1)
```
chrisbra: == `FAIL`

src/strings.c
```diff
@@ -2571,10 +2622,10 @@ parse_fmt_types(
 		if (VIM_ISDIGIT((int)(*p)))
 		{
 		    // Positional argument field width
-		    unsigned int uj = *p++ - '0';
+		    unsigned int uj;
 
-		    while (VIM_ISDIGIT((int)(*p)))
-			uj = 10 * uj + (unsigned int)(*p++ - '0');
+		    if (get_unsigned_int(arg + 1, &p, &uj) == -1)
```
chrisbra: == `FAIL`

src/strings.c
```diff
@@ -2601,10 +2652,11 @@ parse_fmt_types(
 	    {
 		// size_t could be wider than unsigned int; make sure we treat
 		// argument like common implementations do
-		unsigned int uj = *p++ - '0';
+		const char *digstart = p;
+		unsigned int uj;
 
-		while (VIM_ISDIGIT((int)(*p)))
-		    uj = 10 * uj + (unsigned int)(*p++ - '0');
+		if (get_unsigned_int(digstart, &p, &uj) == -1)
```
chrisbra: same here

src/strings.c
```diff
@@ -2625,10 +2677,10 @@ parse_fmt_types(
 		    if (VIM_ISDIGIT((int)(*p)))
 		    {
 			// Parse precision
-			unsigned int uj = *p++ - '0';
+			unsigned int uj;
 
-			while (VIM_ISDIGIT((int)(*p)))
-			    uj = 10 * uj + (unsigned int)(*p++ - '0');
+			if (get_unsigned_int(arg + 1, &p, &uj) == -1)
```
chrisbra: and here

chrisbra:
a left a few comments, please change the return value from `0/-1` to `OK/FAIL` instead.

chrisbra:
thanks. Let me include it.

PR #13802, opened 2023-12-30
fortran syntax improvements

e-kwsm:
- check word boundary
- use `\%(\)` instead of `\(\)`

### DIFF
--- a/runtime/syntax/fortran.vim
+++ b/runtime/syntax/fortran.vim
@@ -68,18 +68,18 @@ if b:fortran_fixed_source == 1
 else
   syn match fortranConstructName	"^\s*\zs\a\w*\ze\s*:"
 endif
-syn match fortranConstructName          "\(\<end\s*do\s\+\)\@11<=\a\w*"
-syn match fortranConstructName          "\(\<end\s*if\s\+\)\@11<=\a\w*"
-syn match fortranConstructName          "\(\<end\s*select\s\+\)\@15<=\a\w*"
-syn match fortranConstructName          "\(\<end\s*where\s\+\)\@14<=\a\w*"
-syn match fortranConstructName          "\(\<end\s*block\s\+\)\@14<=\a\w*"
-syn match fortranConstructName          "\(\<\%(exit\|cycle\)\s\+\)\@11<=\a\w*"
-syn match fortranConstructName          "\(\<end\s*forall\s\+\)\@15<=\a\w*\>"
-syn match fortranConstructName          "\(\<end\s*critical\s\+\)\@17<=\a\w*\>"
-syn match fortranConstructName          "\(\<end\s*associate\s\+\)\@18<=\a\w*\>"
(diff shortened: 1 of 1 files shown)


e-kwsm:
At the moment, `IMPLICIT NONEtheless`, `reCALL`, `STOP1`, etc are wrongly highlighted.

runtime/syntax/fortran.vim
```diff
@@ -189,30 +189,30 @@ syn match fortranOctal	display	"o["'][0-7]\+["']"
...
-syn match fortranLabelNumber	display	"^\d\{1,5}\s"me=e-1
-syn match fortranLabelNumber	display	"^ \d\{1,4}\s"ms=s+1,me=e-1
-syn match fortranLabelNumber	display	"^  \d\{1,3}\s"ms=s+2,me=e-1
-syn match fortranLabelNumber	display	"^   \d\d\=\s"ms=s+3,me=e-1
-syn match fortranLabelNumber	display	"^    \d\s"ms=s+4,me=e-1
+syn match fortranLabelNumber	display	"^\zs\d\{1,5}\ze\s"me=e-1
+syn match fortranLabelNumber	display	"^ \zs\d\{1,4}\ze\s"ms=s+1,me=e-1
+syn match fortranLabelNumber	display	"^  \zs\d\{1,3}\ze\s"ms=s+2,me=e-1
+syn match fortranLabelNumber	display	"^   \zs\d\d\=\ze\s"ms=s+3,me=e-1
+syn match fortranLabelNumber	display	"^    \zs\d\ze\s"ms=s+4,me=e-1
```
dkearns, line 202: The `ms` and `me` offsets can presumably be removed?

dkearns (changes requested):

Ajit-Thakkar (changes requested):
Thank you for all this effort.

I have four small suggestions

Item 1 
On lines 198-202 beginning with 'syn match fortranLabelNumber	display'
the 'ms' and 'me' offsets are equivalent to the \zs and \ze markers.
Only one pair should be used. I do not know which is more efficient. 
If you like \zs and \ze better, then please change the lines to 
syn match fortranLabelNumber	display	"^\zs\d\{1,5}\ze\s"
syn match fortranLabelNumber	display	"^ \zs\d\{1,4}\ze\s"
syn match fortranLabelNumber	display	"^  \zs\d\{1,3}\ze\s"
syn match fortranLabelNumber	display	"^   \zs\d\d\=\ze\s"
syn match fortranLabelNumber	display	"^    \zs\d\ze\s"

Item 2
Please delete line 215
syn match fortranKeyword	"^\s*\zs\d\+\s\+continue\>"
because it is unnecessary; I left it behind accidentally in a previous change of mine.

Item 3
Both lines 357 and 369 beginning 'syn region fortranModule' contain an incorrect 
\z\%(
which should be changed to just
\z\(
because \%(\) cannot be used after \z

Item 4
Line 3, please change version from v109 to v110 and update the date appropriately.

PR #13803, opened 2023-12-30
Refactored runtime/ftplugin/zig.vim, removed upstream comment, and removed aucmd and auto formatting support.

Tiseno:
https://github.com/ziglang/zig.vim added auto formatting and linting (and population of the qflist) on save as a aucmd with an **opt-out** configuration, and was then merged into vim runtime. This seems to me like something that should be left up to that plugin and not be a part of vim.

It should probably be:
**opt-in**, as it is very unexpected behavior and uncharacteristic of vim to do something like this automatically
**or removed entirely**, as it should be left up to the user to register that kind of behavior if they want it through a plugin.
This PR does the latter.

I could not find any concrete guidelines of what built in file type plugins should and should not do, I am personally fine with whatever, opt-in, or remove the format/astcheck functionality. But in any case I think it should not register any aucmds by default without opting in to it (i.e. the check for opt in should be for registering the aucmd, not for running the formatter or not). I have noted that the rust filetype support also has a lot of options available as opt-in, and it also **always** registers an aucmd.
Is this the direction vim is moving in? i.e. upstreaming oppinionated plugins for integrating languages tightly?
My expectation before going down this rabbit hole was that built-in language supports only provided: filetype, filetype detection, syntax, indentation, and a makeprg.

The `zig.vim` plugin wants to be an ide-like plugin, and this PR removes the upstream dependency for the ftplugin implementation. I also made some consistency changes to the code in how undo_ftplugin is handled and added the `zig_recommended_style` as consistent with other ftplugins. The only significant thing left is the g:zig_std_dir/path and execution of zig env, which might also be up for discussion if it should be removed.


habamax:
I believe autocmd that does autoformat should absolutely not happen by default.

habamax:
I would also avoid all `g:zig_std_dir` set up:

```
" Safety check: don't execute zig from current directory
if !exists('g:zig_std_dir') && exists('*json_decode') &&
    \  executable('zig') && dist#vim#IsSafeExecutable('zig', 'zig')
    silent let s:env = system('zig env')
    if v:shell_error == 0
        let g:zig_std_dir = json_decode(s:env)['std_dir']
    endif
    unlet! s:env
endif
```
On windows gvim `system()` call looks ugly and you probably don't want to see cmd.exe window popup upon opening zig file.

Should be either opt-in or just a ftplugin help entry on how to set up `g:zig_std_dir`

habamax:
> Is there a reason not to send your more general changes through the maintainer's repo?

They might disagree with the proposed changes, but yeah it needs to be addressed there too(first?).

PS,
Do we have vim runtime files guide? What to avoid in ftplugin/syntax files?

Tiseno:
The general changes can be added to the zig.vim plugin after the fact, if they want them.
This is a proposal to split from that plugin where vim maintains its own ftplugin implementation (as zig.vim wants to do things that should probably not be in the vim runtime).

And please note that I am only referring to `runtime/ftplugin/zig.vim` not the syntax/indentation/compiler implementations, I see no problem with those, so they shold still be upstreamed.

Tiseno:
Just as an example, the go ft plugin was split into two, one bare bones and one more ide-like.
https://github.com/google/vim-ft-go
https://github.com/fatih/vim-go
The former one was merged into vim and is now obsolete. `zig.vim` is comparable to the vim-go plugin.

runtime/ftplugin/zig.vim
```diff
@@ -1,44 +1,44 @@
...
-setlocal shiftwidth=4
-
 setlocal formatoptions-=t formatoptions+=croql
-
-setlocal suffixesadd=.zig,.zir
+setlocal suffixesadd=.zig,.zir,.zon
+let &l:define='\v(<fn>|<const>|<var>|^\s*\#\s*define)'
+let b:undo_ftplugin = 'setl isk< fo< sua< mp< def<'
+
+if get(g:, 'zig_recommended_style', 1)
```
chrisbra, line 25: > if get(g:, 'zig_recommended_style', 1)

That variable should be documented. Please place a short section into runtime/doc/syntax.txt (e.g. see how it is done for other synax files, like `:h ft-python-syntax` . Don't forget to add a tag and update the last change header please.

flexagoon:
What is currently blocking this PR? Is it just the undocumented variable?

chrisbra:
yes

flexagoon:
@Tiseno could you please add the missing doc section? This is a blocking PR for https://github.com/neovim/neovim/issues/28878

gpanders:
Here is a patch for the docs:

```diff
diff --git a/runtime/doc/filetype.txt b/runtime/doc/filetype.txt
index deb947cf1..feb946252 100644
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -829,6 +829,12 @@ The mappings can be disabled with: >
 	let g:no_vim_maps = 1
 
 
+ZIG							*ft-zig-plugin*
+
+						*g:zig_recommended_style*
+The Zig |ftplugin| sets options according to Zig's recommended style. This can
+be disabled by setting |g:zig_recommended_style| to 0.
+
 ZIMBU							*ft-zimbu-plugin*
 
 The Zimbu filetype plugin defines mappings to move to the start and end of
```

I put it in `doc/filetype.txt` rather than `doc/syntax.txt` since the variable is relevant for the ftplugin, not the syntax file.

chrisbra:
Thanks, makes sense. I'll have a look later today

Tiseno:
The doc comment!
But there is also the matter of the zig_std_dir, which I would like to remove, but I do not even know fully what its purpose is.

gpanders:
> But there is also the matter of the zig_std_dir, which I would like to remove because it also seems intrusive. But I do not even know fully what its purpose is.

It is used to set the `'path'` variable, which is useful for e.g. `gf`, `:find`, `[I`, etc. If the ftplugin is setting `'include'`, `'includexpr'`, and `'define'` already, then setting `'path'` is probably a useful thing to do. And gating it behind the existence of `g:zig_std_dir` (so that users can just set it themselves) seems harmless enough.

Tiseno:
Ok, seems good to me, but is the implementation good enough?
@habamax pointed out
> On windows gvim system() call looks ugly and you probably don't want to see cmd.exe window popup upon opening zig file.
> Should be either opt-in or just a ftplugin help entry on how to set up g:zig_std_dir

runtime/ftplugin/zig.vim
```diff
@@ -1,44 +1,44 @@
 " Vim filetype plugin file
 " Language: Zig
-" Upstream: https://github.com/ziglang/zig.vim
```
chrisbra, line 3: So no more upstream?

runtime/doc/filetype.txt
```diff
@@ -829,6 +829,13 @@ The mappings can be disabled with: >
 	let g:no_vim_maps = 1
 
 
+ZIG							*ft-zig-plugin*
+
+						*g:zig_recommended_style*
+The Zig |ftplugin| sets options according to Zig's recommended style. This can
+be disabled by setting |g:zig_recommended_style| to 0.
```
chrisbra: Do we have a reference, what the Zig recommended style is?

runtime/ftplugin/zig.vim
```diff
@@ -50,19 +50,10 @@ if !exists('g:zig_std_dir') && exists('*json_decode') &&
 endif
 
 if exists('g:zig_std_dir')
```
chrisbra, line 45: `g:zig_std_dir` should also be documented

runtime/ftplugin/zig.vim
Tiseno: What is meant by tag and last change header?

gpanders:
> Ok, seems good to me, but is the implementation good enough? @habamax pointed out
> 
> > On windows gvim system() call looks ugly and you probably don't want to see cmd.exe window popup upon opening zig file.
> > Should be either opt-in or just a ftplugin help entry on how to set up g:zig_std_dir

We don't need to shell out or call `system()`. If the user sets `g:zig_std_dir` themselves, the ftplugin can use it to set `'path'`. Otherwise leave `'path'` alone. But `g:zig_std_dir` probably needs to be included in the docs too.

runtime/ftplugin/zig.vim
gpanders: The changes here differ significantly enough from upstream that it is not really a copy of the upstream plugin at all anymore. Perhaps we can leave a comment that the ftplugin is _based_ on the upstream plugin though.

runtime/doc/filetype.txt
gpanders: Can link to https://ziglang.org/documentation/0.12.0/#Style-Guide maybe? But that might link rot.

gpanders:
@chrisbra The original implementation (upstream) has a `system()` call (`system('zig env')`). The question was (I believe) whether this should be preserved or removed.

runtime/ftplugin/zig.vim
chrisbra: Then please also add: `Maintainer: This runtime file is looking for a new maintainer.` (similar to other files):

runtime/doc/filetype.txt
Tiseno: Yes https://ziglang.org/documentation/master/#Whitespace
Should that be referenced in the section?

runtime/ftplugin/zig.vim
Tiseno: Yes, that was the idea. This is a split from the ziglang/zig.vim implementation of the ftplugin/zig.vim

chrisbra:
Ah, okay, let's leave it out for now. I am not a big fan of running system commands just by loading a filetype. Just document that `g:zig_std_dir` can be set by `system('zig env')` in a custom filetype auto command

runtime/ftplugin/zig.vim
gpanders: I am happy to act as maintainer (unless @Tiseno would like to). I upstreamed the original runtime files.

gpanders:
>Just document that `g:zig_std_dir` can be set by `system('zig env')` in a custom filetype auto command

Specifically, the `std_dir` field from the output of `zig env`. If we want a one-liner, it's `:echo json_decode(system('zig env'))['std_dir']`.

runtime/ftplugin/zig.vim
Tiseno: I can become maintainer, that would be cool!

Something like
```
" Maintainer:   Mathias Lindgren <math.lindgren@gmail.com>
" Based on:     https://github.com/ziglang/zig.vim
```
then?

chrisbra: yes exactly like this.

runtime/doc/filetype.txt
chrisbra: Either like this, or have a look how it is done for python, explicitly mentioned what properties are set `:h ft-python-plugin`

runtime/ftplugin/zig.vim
chrisbra: Basically add a help jump target, like it is done now: `*g:zig_recommended_style*` in the help file. And then we usually have a separate line, mentioning when the last change was done, like here: https://github.com/vim/vim/blob/master/runtime/ftplugin/c.vim#L4

Tiseno:
> Ah, okay, let's leave it out for now. I am not a big fan of running system commands just by loading a filetype. Just document that g:zig_std_dir can be set by system('zig env') in a custom filetype auto command

Now I just declared that one can put it in a aucmd, but do we want a full example for that instead?
How would one do that in the most idiomatic way?

runtime/doc/filetype.txt
```diff
@@ -829,6 +829,30 @@ The mappings can be disabled with: >
...
 
 
+ZIG							*ft-zig-plugin*
+
+						*g:zig_recommended_style*
+By default the following indentation options are set, in accordance with Zig's
+recommended style (https://ziglang.org/documentation/master/):
+
+	`setlocal expandtab shiftwidth=4 softtabstop=4 tabstop=8`
+
```
gpanders: For code, use `>`, not backticks (which is Markdown syntax).

```suggestion
recommended style (https://ziglang.org/documentation/master/): >

	setlocal expandtab shiftwidth=4 softtabstop=4 tabstop=8
<
```

gpanders: Same for the line below.

runtime/doc/filetype.txt
```diff
@@ -829,6 +829,30 @@ The mappings can be disabled with: >
...
+To disable this behavior, set g:zig_recommended_style to 0:
+
+	`let g:zig_recommended_style = 0`
+
+						*g:zig_std_dir*
+The path to the zig standard library. The Zig |ftplugin| reads g:zig_std_dir
+and appends it to the |path| for zig files. Where the zig standard library
+is located is system and installation method dependant.
```
gpanders:
```suggestion
The path to the Zig standard library. The Zig |ftplugin| reads g:zig_std_dir
and appends it to 'path' for Zig files. Where the Zig standard library
is located is system and installation method dependent.
```

runtime/doc/filetype.txt
```diff
@@ -829,6 +829,30 @@ The mappings can be disabled with: >
...
+
+One can automatically read the std_dir using `zig env`:
+	`let g:zig_std_dir = json_decode(system('zig env'))['std_dir']`
+
+This can, for example, be put in a FileType |autocmd| to only load when a zig
+file is opened.
```
gpanders:
```suggestion
One can automatically set |g:zig_std_dir| using `zig env`: >
	let g:zig_std_dir = json_decode(system('zig env'))['std_dir']
<
This can, for example, be put in a FileType |autocmd| or user |ftplugin| to only load when a Zig
file is opened.
```

runtime/ftplugin/zig.vim
Tiseno: Fixed

runtime/doc/filetype.txt
Tiseno: Fixed, I added both the link and the set fields

chrisbra:
thanks all!

andrewrk:
I do not maintain the vim plugin that this patches, and I did not write the aucmd and autoformatting functionality that this disables. However, I am a vim user and a user of this plugin.

I agree that auto format on save is not appropriate for being the default. However, it is appropriate for `zig ast-check foo.zig` to be enabled by default. This *only* checks for errors, and it does so instantly, making it extremely useful. Zig programmers absolutely should see errors instantly when they save by default. Unfortunately, the logic that this PR disables did both - formatting and ast-check. The appropriate model would be a tristate enum:
* fully disabled (what this PR does)
* ast-check only (correct default in my opinion)
* zig fmt plus ast-check (previous default)

Tiseno:
I am now the maintainer of the zig ftplugin in the vim runtime!
There is nothing preventing you from installing the ziglang/zig.vim to get your desired automatic behavior. We deemed that implementation to be too intrusive for inclusion in the runtime, and hence removed the upstream dependency on it. 

You can merge this implementation back into ziglang/zig.vim if you want and then change it to have better defaults, and propose to upstream it again. But looking at the runtime plugins for other languages this sort of integration is quite uncommon, interfacing with a specific external formatter/checker tool (there are only a handful of exceptions, rust being one of them).

Personally I think this sort of functionality should not be included in the runtime for any language (favoritism for a particular tool), it should be installed by a user as a plugin (ziglang/zig.vim or a more general one). Note that this is my personal opinions and as stated in the description I could not find any guidelines regarding what fits in the runtime philosophically.

This was not (for me at least) a discussion about how useful automatic formatting is (i use it myself) but how it was implemented and respecting the users choice. There are a vast number of different workflows in vim for having automatic formatting/linting (lsp-integration, coq, ale, neomake, etc). Removing this automatic behavior from the runtime until we have a solid implementation that respects ALL users was the only reasonable way forward (as I see it).

andrewrk:
Congrats, that's great. The ziglang/zig.vim repository is also unmaintained, so let me know if you would like to also maintain that.

Just to be clear, are you saying that it's the standard vim practice for plugins to not do any kind of (non-mutating) verification or checking of files by default? I want to make sure you didn't interpret my comment above as advocating for reverting this PR - rather I am suggesting a new, third thing become the default which the ziglang/zig.vim plugin does not even support today, and which I have no plans to implement myself.

Tiseno:
> Congrats, that's great. The ziglang/zig.vim repository is also unmaintained, so let me know if you would like to also maintain that.
Thanks, but that plugin is obsolete for me personally so I would have no stake in it, I use zls with neovim lsp which works great. The only reason I took an interest in the zig vim runtime files was because of disruptive builtin defaults.

> Just to be clear, are you saying that it's the standard vim practice for plugins to not do any kind of (non-mutating) verification or checking of files by default?
The builtin plugins yes, **I think so**, they should have no default behavior added to the basic vim experience.
Adding opt-in configurations for enabling automatic formatting and such things is not uncommon. Having anything like this on by default, however, goes against how most (all?) other **builtin** plugins are currently implemented.
If ziglang/zig.vim stays as an external plugin, it can have any default behavior it wants.

andrewrk:
I see, thank you for the clarifications.

PR #13805, opened 2023-12-31
runtime(ruby): Update ftplugin and omni-complete

dkearns:


chrisbra:
Thanks. I suppose it is intentional, that you removed yourself from being a Release coordinator?

PR #13806, opened 2023-12-31
Update sh.vim

grondilu:
add `local` in shStatement (for bash)

It should help matching local array declarations:

```bash
foo() {
    local -a bar=(a b c)
}
```


chrisbra:
thanks!

PR #13807, opened 2024-01-02
doc(INSTALLpc): Updated Ruby config.h instructions

dertuxmalwieder:
RubyInstaller will want different directories. Updated.

### DIFF
--- a/src/INSTALLpc.txt
+++ b/src/INSTALLpc.txt
@@ -746,11 +746,11 @@ config.h and Ruby's DLL name. Here are the steps for working around them:
 
     For 32-bit version:
 
-      xcopy /s .ext\include C:\Ruby24\include\ruby-2.4.0
+      xcopy /s .ext\include\i386-mswin32_140\config.h C:\Ruby24\include\ruby-2.4.0\ruby
 
     For 64-bit version:
 
-      xcopy /s .ext\include C:\Ruby24-x64\include\ruby-2.4.0
+      xcopy /s .ext\include\x64-mswin64_140 C:\Ruby24-x64\include\ruby-2.4.0\ruby
 
     Note that 2.4.0 is Ruby API version of Ruby 2.4.X.
     You may need to close the console and reopen it to pick up the new $PATH.


src/INSTALLpc.txt
```diff
@@ -746,11 +746,11 @@ config.h and Ruby's DLL name. Here are the steps for working around them:
 
     For 32-bit version:
 
-      xcopy /s .ext\include C:\Ruby24\include\ruby-2.4.0
+      xcopy /s .ext\include\i386-mswin32_140\config.h C:\Ruby24\include\ruby-2.4.0\ruby
 
     For 64-bit version:
 
-      xcopy /s .ext\include C:\Ruby24-x64\include\ruby-2.4.0
+      xcopy /s .ext\include\x64-mswin64_140 C:\Ruby24-x64\include\ruby-2.4.0\ruby
```
k-takata, line 753: `config.h` should be placed at `C:\RubyXY\include\ruby-X.Y.0\i386-mswin32_140\config.h` (32-bit) or `C:\RubyXY-x64\include\ruby-X.Y.0\x64-mswin64_140\config.h` (64-bit), not at `C:\RubyXY{,-x64}\include\ruby-X.Y.0\ruby\config.h`.
The original commands correctly copy them.

dertuxmalwieder: That leads - at least on my machine ... - to an error that config.h could not be found. The updated commands fixes that for me.

k-takata: Some of the build options might not be correctly set? (E.g. `RUBY_API_VER_LONG`, `RUBY_PLATFORM`)

dertuxmalwieder: The problem is that Ruby's own `internal/config.h` (used by most Ruby header files) explicitly requires a `ruby/config.h`...

--> https://github.com/ruby/ruby/blob/5124f9ac7513eb590c37717337c430cb93caa151/include/ruby/internal/config.h#L22

k-takata: > `config.h` should be placed at `C:\RubyXY\include\ruby-X.Y.0\i386-mswin32_140\config.h` (32-bit) or `C:\RubyXY-x64\include\ruby-X.Y.0\x64-mswin64_140\config.h` (64-bit),

Sorry, the correct paths are:
32-bit: `C:\RubyXY\include\ruby-X.Y.0\i386-mswin32_140\ruby\config.h`
64-bit: `C:\RubyXY-x64\include\ruby-X.Y.0\x64-mswin64_140\ruby\config.h`

`C:\RubyXY\include\ruby-X.Y.0\i386-mswin32_140` or `C:\RubyXY-x64\include\ruby-X.Y.0\x64-mswin64_140` is specified as an include directory by the `-I` option in the makefile.
So, the compiler should be able to find `ruby/config.h`.

dertuxmalwieder: Indeed, it does. Turns out it was a configuration mistake on my side. 🤦‍♂️
Sorry.

PR #13808, opened 2024-01-03
runtime: add jumpoptions to vim syntax file

hugovandevliert:
Hi there 👋🏼 

While testing the new `jumpoptions` option, I noticed it's not highlighted correctly in my `.vimrc` file. Looks like the option was missing in `runtime/syntax/vim.vim`.


chrisbra:
thanks!

PR #13810, opened 2024-01-03
runtime(vim): Update syntax file - match class declarations

dkearns:
Improve class, enum and interface highlighting

- Enable folding of class, enum and interface declarations.
- Highlight constructor names with the Function highlight group, like other special methods.
- Mark function definitions using special method names as errors.
- Highlight :type arguments.

Fixes issue [#14393#issuecomment-2042796198](https://github.com/vim/vim/issues/14393#issuecomment-2042796198).


zzzyxwvut:
I'm aware that this is still a draft stage for this PR, but  
there is a regression with the changeset `f8d0ab` that is  
manifested in the broken recognition of `abstract` methods of  
a `class` (and can also be seen without folding).

Below is a screenshot made for the Vim command (`v9.1.0645`)  
loading the proposed `vim.vim` version and a test file taken    
and adapted from [here](https://github.com/vim/vim/issues/14393#issuecomment-2042796198):

```shell
vim --clean --cmd 'setl rtp^=~/.vim/ fdc=2 fdl=99 fdm=syntax' \
  --cmd "let g:vimsyn_folding = 'cif'" \
  /tmp/test.vim
```
type_folding


Below is a screenshot made for the Vim command (`v9.1.0645`)  
loading the local `master` version of `vim.vim` and a copy of  
the test file with the comments expunged:

```shell
vim --clean /tmp/test_.vim
```
type_unfolding

zzzyxwvut:
Also, `:help builtin-object-methods` can be valid `interface`  
members.

```
interface Textual
    def string(): string
endinterface
```

zzzyxwvut:
Cool!  Now, to round up type related definitions, what  
remains is to mark as invalid (`E1267`) all such script  
function definitions in legacy and Vim9 contexts (possibly  
with `g:` variants and in nested declarations):

```
def id(o: any): any
    return o
enddef

def new()
enddef
def newOther()
enddef
def newyetanother()
enddef

def empty(): bool
    return true
enddef
def len(): number
    return 0
enddef
def string(): string
    return ""
enddef
```

chrisbra:
thanks!

Issue #13809, opened 2024-01-03
Use ssh to connecting Linux, mapped shortcut keys in "mswin.vim" to operate the clipboard is not working.


znsoooo:
### Steps to reproduce

1. Use Moba software to connect to Linux development board by ssh.
2. Open vim and enter ":source $VIMRUNTIME/mswin.vim"
3. I can't use the Ctrl-C / Ctrl-X / Ctrl-V hotkeys for clipboard operation. But the Ctrl-Z and Ctrl-Y hotkeys is working.

### Expected behaviour

Use the Ctrl-C / Ctrl-X / Ctrl-V shortcut keys for clipboard operation while I'am using "mswin.vim" script.

### Version of Vim

9.1.0

### Environment

Ubuntu 18.04.6 LTS

znsoooo:
And I an already fixed this bug: https://github.com/vim/vim/commit/1d9f46bfb718c0e67a77a06138df412ad462f820

chrisbra:
You can create a pull request, but using the unnamed register is something very different then using the (system) wide clipboard.

znsoooo:
> You can create a pull request, but using the unnamed register is something very different then using the (system) wide clipboard.

Yes, so it only takes effect when the system clipboard is not available.

ciupicri:
The patch breaks gvim (vim-X11-9.1.158-1.fc39.x86_64) when running under X11 (XFCE) on Fedora Linux. The keyboard shortcuts don't work anymore.

chrisbra:
what exactly does not work?

ciupicri:
For example if I select some text in gvim and press <kbd>Control</kbd>+<kbd>C</kbd> , then <kbd>Control</kbd>+<kbd>V</kbd> in say Firefox the old content of the clipboard gets pasted. This means that the copy operation didn't put something new into the clipboard, i.e. it doesn't work.

On the other hand, select, right click →  **Copy** works fine.

---

```console
$ cat ~/.vim/vimrc
source $VIMRUNTIME/mswin.vim
```

chrisbra:
Please give more information. 
So you are visual selecting anything and press `<c-c>`? What does `:vnoremap <c-c>` give you? What about `:echo has("clipboard_working")`?

If you change line 27 from:
```
if has("clipboard_working")
```
to 
```
if has("clipboard")
```

Does it start working?

ciupicri:
> So you are visual selecting anything and press `<c-c>`?

[...]

PR #13813, opened 2024-01-03
Use unnamed register while clipboard not exist

znsoooo:
It only takes effect when the system clipboard is not available.

From issue: https://github.com/vim/vim/issues/13809

### DIFF
--- a/runtime/mswin.vim
+++ b/runtime/mswin.vim
@@ -1,8 +1,8 @@
 " Set options and add mapping such that Vim behaves a lot like MS-Windows
 "
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
-" Last Change:	2023 Aug 10
-" Former Maintainer:	Bram Moolenaar <Bram@vim.org>
+" Last Change:	2024 Jan 3
+" Former Maintainer:	Shixian Li <lsx7@sina.com>
 
 " Bail out if this isn't wanted.
 if exists("g:skip_loading_mswin") && g:skip_loading_mswin
@@ -39,6 +39,23 @@ if has("clipboard")
 
     cmap <C-V>		<C-R>+
     cmap <S-Insert>		<C-R>+
+else
+    " Use unnamed register while clipboard not exist
+
+    " CTRL-X and SHIFT-Del are Cut
+    vnoremap <C-X>   ""x
+    vnoremap <S-Del> ""x
+
+    " CTRL-C and CTRL-Insert are Copy
+    vnoremap <C-C>      ""y
+    vnoremap <C-Insert> ""y
+
+    " CTRL-V and SHIFT-Insert are Paste
+    noremap <C-V>      ""gP
+    noremap <S-Insert> ""gP
+
+    inoremap <C-V>      <C-R>"
+    inoremap <S-Insert> <C-R>"
 endif
 
 " Pasting blockwise and linewise selections is not possible in Insert and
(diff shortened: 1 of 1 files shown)


runtime/mswin.vim
```diff
@@ -1,8 +1,8 @@
 " Set options and add mapping such that Vim behaves a lot like MS-Windows
 "
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
-" Last Change:	2023 Aug 10
-" Former Maintainer:	Bram Moolenaar <Bram@vim.org>
+" Last Change:	2024 Jan 3
+" Former Maintainer:	Shixian Li <lsx7@sina.com>
```
zeertzjq: Why are you changing former maintainer?

arpadffy: Would not it be better keeping the "former maintainer" and adding "current maintainer" ?

chrisbra: Don't change this line

runtime/mswin.vim
```diff
@@ -39,6 +39,23 @@ if has("clipboard")
...
+    " CTRL-C and CTRL-Insert are Copy
+    vnoremap <C-C>      ""y
+    vnoremap <C-Insert> ""y
+
+    " CTRL-V and SHIFT-Insert are Paste
+    noremap <C-V>      ""gP
+    noremap <S-Insert> ""gP
+
+    inoremap <C-V>      <C-R>"
+    inoremap <S-Insert> <C-R>"
```
chrisbra, line 58: Do we need to specify the unnamed register explicitly? That should be the default anyhow

runtime/mswin.vim
```diff
@@ -47,7 +64,7 @@ endif
 " Uses the paste.vim autoload script.
 " Use CTRL-G u to have CTRL-Z only undo the paste.
 
-if 1
+if has("clipboard")
```
chrisbra: `has("clipboard_working")` ?

runtime/mswin.vim
znsoooo: > Why are you changing former maintainer?

@zeertzjq 
Sorry, I misunderstood the meaning of this statement. I've reverted it.

znsoooo: > Don't change this line

@chrisbra 
Excuse me, do you mean don't modify all lines, including dates, or don't modify the author, or instead of deleting but adding a line of author?

chrisbra: Don't change the `Former Maintainer`, unless  you are intending to make a statement about who was the former maintainer.

chrisbra:
can you please make the following changes? 

- Change it to `has("clipboard_working")`
- don't map using the default register.

znsoooo:
@chrisbra Sorry, I forgot to check the messages recently.

----

> Change it to `has("clipboard_working")`

But why is "clipboard_working"?

I looked at "feature-list", but only saw "clicpboard", and there is no "clipboard_working":

image

And as I mentioned before, "some system clipboard is available", on these computers there did has the "clicpboard" register but no "clipboard_working" register. If I do change, it will cause the computers that "mswin.vim" can be effacted do not work anymore.

----

> don't map using the default register.

What register do you recommend I use?

And my consideration for using the default registers is that if you copy text from another command (e.g. yy), you can also paste the copied text with ctrl-v.

chrisbra:
> But why is "clipboard_working"?

It is here: https://github.com/vim/vim/blob/1af35631f85d2fcdc83c5d457af8273697f5146a/runtime/doc/builtin.txt#L11102

you want to make sure the clipboard actually works (this matters when using vim over a remote connection).


> What register do you recommend I use?

> And my consideration for using the default registers is that if you copy text from another command (e.g. yy), you can also paste the copied text with ctrl-v.

Why use a register at all? If you don't specify a register it goes to the default one, right? And if you want to use a different register, you can simply use `"b<ctrl-x>` and it would end up in register b.

znsoooo:
@chrisbra 
I got it, I will make the change. I'm on vacation now and my computer is not with me. I'll change it later.

PR #13819, opened 2024-01-05
cib, ciB and ci[ should work with empty inner textobjects

habamax:
Should fix regression introduced by https://github.com/vim/vim/pull/13523

Relates to #13514


chrisbra:
thanks, let me merge it to fix this recent annoyance

PR #13821, opened 2024-01-05
zos: fix xxd and set UTF-8 as the default encoding

IgorTodorovskiIBM:
* This removes the z/OS EBCDIC 1047 assumptions in xxd. We now build all new applications in ASCII/UTF8 mode for z/OS 
* z/OS support auto-conversion of tagged files, but for xxd we just want to compare the binary, therefore I've also added code to disable auto-conversion.
* I've also changed the Makefile to allow passing in of extra libraries (`$LIBS`). On z/OS, we typically link in another open source library, known as zoslib which provides compat with Linux.

For Vim:
* This also sets the default encoding to UTF-8 for z/OS (similar to win32)

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -450,9 +450,10 @@ set_init_default_encoding(void)
     char_u	*p;
     int		opt_idx;
 
-# ifdef MSWIN
+# if defined(MSWIN) || defined(__MVS__)
     // MS-Windows has builtin support for conversion to and from Unicode, using
     // "utf-8" for 'encoding' should work best for most users.
+    // z/OS built should default to UTF-8 mode as setlocale does not respect utf-8 environment variable locales
     p = vim_strsave((char_u *)ENC_DFLT);
 # else
     // enc_locale() will try to find the encoding of the current locale.
--- a/src/option.h
+++ b/src/option.h
@@ -130,7 +130,7 @@ typedef enum {
 #define ENC_UCSBOM	"ucs-bom"	// check for BOM at start of file
 
 // default value for 'encoding'
-#ifdef MSWIN
+#if defined(MSWIN) || defined(__MVS__)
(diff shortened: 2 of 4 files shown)


chrisbra:
Thanks. Is this dependent on a certain version for zOS?

IgorTodorovskiIBM:
> Thanks. Is this dependent on a certain version for zOS?

ASCII support is available on all supported IBM z/OS releases and also in the compilers (xlc, xlclang, and clang)

chrisbra:
What I mean with this, are there still versions of z/OS /MVS that need the old way of converting the encodings? Or is this not a problem in praxis anymore?

IgorTodorovskiIBM:
> What I mean with this, are there still versions of z/OS /MVS that need the old way of converting the encodings? Or is this not a problem in praxis anymore?

We're trying to move away from EBCDIC, but there is still the option to build in EBCDIC mode so I suppose someone could still build vim in ebcdic mode. The compilers all set [__CHARSET_LIB](https://www.ibm.com/docs/en/ent-metalc-zos/3.1?topic=features-macros-related-compiler-option-settings) depending on which mode is set. I've updated the xxd source code to reflect that.

chrisbra:
thanks. Looks good I think

chrisbra:
BTW: can you please split the commits into:
- one commit for the xxd related changes
- one commit for changing the encoding on Vim

Is this then all that is required to build Vim on z/OS?

IgorTodorovskiIBM:
> BTW: can you please split the commits into:
> 
> * one commit for the xxd related changes
> * one commit for changing the encoding on Vim
> 
> Is this then all that is required to build Vim on z/OS?

Thanks, I've split it into two commits.

Yes, just these changes will enable builds of Vim on z/OS with the newest compilers.

PR #13823, opened 2024-01-05
Netrw plugin fix for windows

MiguelBarro:
The move command doesn't work on windows due:
+ The use of an uninitialized variable.
+ A filepath is prefixed twice with its dir.

~~This draft is for accountability purposes the corresponding patch has been mailed to Mr. Campbell.~~

Mr. Campbell has kindly answered me he is no longer supporting netrw actively. Let's fix this via PR then.


chrisbra:
thanks

PR #13825, opened 2024-01-06
Remove diff* links added in #13776 and doc added in commit b1392be

romainl:
The links added in #13776 are way too noisy for the contexts in which the `diff` syntax is applied (git commits, patches, etc.).

This commit…

- removes those links
- adds hopefully semantically correct styling for `diffAdded`, `diffChanged`, and `diffRemoved`, valid when using the `default` colorscheme and other colorschemes that don't explicitly override those highlight groups

For reference:

- `diff` before the change introduced in #13776

  Capture d’écran 2024-01-04 à 19 22 53
  
  Capture d’écran 2024-01-06 à 13 57 24

- `diff` *after* the change introduced in #13776 

  Capture d’écran 2024-01-04 à 19 25 14

  Capture d’écran 2024-01-06 à 14 00 19

- `diff` *after* this commit

   Capture d’écran 2024-01-06 à 14 02 31

  Capture d’écran 2024-01-06 à 14 02 13

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -1,4 +1,4 @@
-*syntax.txt*	For Vim version 9.1.  Last change: 2024 Jan 03
+*syntax.txt*	For Vim version 9.1.  Last change: 2024 Jan 06
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
@@ -1261,7 +1261,7 @@ To highlight KDE-reserved features, set >
 g:desktop_enable_kde follows g:desktop_enable_nonstd if not supplied
 
 
-DIFF						*diff.vim* *ft-diff-syntax*
+DIFF							*diff.vim*
 
 The diff highlighting normally finds translated headers.  This can be slow if
 there are very long lines in the file.  To disable translations: >
@@ -1270,15 +1270,6 @@ there are very long lines in the file.  To disable translations: >
 
 Also see |diff-slow|.
 
-Since the Vim 9.1 release |version-9.1| the diff filetype links the diffAdded,
-diffRemoved and diffChanged highlighting groups to |hl-DiffAdd|,
-|hl-DiffDelete| and |hl-DiffChange| by default.  If you do not want this, you
-can change it to the previous groups like this in your |.vimrc| >
-
-	hi link diffRemoved Special
-	hi link diffChanged PreProc
-	hi link diffAdded Identifier
-<
 DIRCOLORS			       *dircolors.vim* *ft-dircolors-syntax*
 
 The dircolors utility highlighting definition has one option.  It exists to
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -271,6 +271,10 @@ you can see the actual color, except for "Ignore"):
 	*Todo		anything that needs extra attention; mostly the
 			keywords TODO FIXME and XXX
 
+	*diffAdded	added line in a diff
+	*diffChanged	changed line in a diff
+	*diffRemoved	removed line in a diff
+
 The names marked with * are the preferred groups; the others are minor groups.
 For the preferred groups, the "syntax.vim" file contains default highlighting.
 The minor groups are linked to the preferred groups, so they get the same
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -7207,7 +7207,6 @@ ft-cynlib-syntax	syntax.txt	/*ft-cynlib-syntax*
 ft-dart-syntax	syntax.txt	/*ft-dart-syntax*
 ft-dash-syntax	syntax.txt	/*ft-dash-syntax*
 ft-desktop-syntax	syntax.txt	/*ft-desktop-syntax*
-ft-diff-syntax	syntax.txt	/*ft-diff-syntax*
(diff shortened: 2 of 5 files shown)


clason:
This essentially adds new filetype-specific "default group names" (as in `:h group-name`); if this is an option, this is clearly a better alternative than relying on default links.

I think this is a good idea, since the current, programming language-oriented, default groups are a poor fit for many markup or data languages (Markdown, Vimdoc, LaTeX, JSON, ...)

Are you generally considering adding more of such groups (as need arises, and obviously not in this PR)?

chrisbra:
Thanks, definitely an improvement. Can you please also document those new groups at `:h group-name`? 

> Are you generally considering adding more of such groups (as need arises, and obviously not in this PR)?

If need arises

clason:
(And you probably want to update the bundled colorschemes to account for this change.)

habamax:
In default gvim (white bg), `diffAdded` is kind of hard to read:

image

habamax:
What about `DarkGreen` for light bg?

image

habamax:
limegreen is still too bright to my taste/eyes

image

romainl:
I've…

- added the three names to `:help group-name`,
- added them to `syntax/help.vim`,
- changed `green` to `limegreen` for `diffAdded` in GUI,
- changed `blue` to `dodgerblue` for `diffChanged` in GUI.

Light background:

Capture d’écran 2024-01-07 à 10 00 29

Dark background:

Capture d’écran 2024-01-07 à 10 01 14

`limegreen` is darker and less saturated than `green` without being as dark as `darkgreen`, so it is more legible on white and less jarring on black.

`dodgerblue` is lighter than `blue` but not too bright so it works equally well on white and black.

romainl:
> limegreen is still too bright to my taste/eyes
> 
> image

Could you suggest a few alternatives from https://en.wikipedia.org/wiki/X11_color_names#Color_name_chart ?

habamax:
Interesting, in your screenshots limegreen is actually readable compared to mine. I guess font-size matters a lot here.

habamax:
> > limegreen is still too bright to my taste/eyes
> > image
> 
> Could you suggest a few alternatives from https://en.wikipedia.org/wiki/X11_color_names#Color_name_chart ?

- SeaGreen
- ForestGreen
- WebGreen

chrisbra:
The header `diffOldFile` and `diffNewFile` link to `DiffFile` which resolves to Type which uses  `SeaGreen`, so let's use this one for `DiffAdded` as well.

chrisbra:
I have made a few minor changes, which turned out to be even more involved:
- Renamed those new groups to `Added`, `Changed` and `Removed` (I initially went with capitalizing the `diffAdded` to `DiffAdded`, but since hightlighting groups are not case sensitive, it causes problems for linking the groups in the color schemes, so I had to make the name different and removed the `diff` prefix)
- Changed the links in the default colorschemes
- linked those new groups back in the `diff` syntax file.
- capitalize the color names consistently

That is what I have temporarily queued, not sending out yet, would like to get an okay for that.:

https://github.com/chrisbra/vim/commit/0fea0abf5a2e4a9b91df9b1c0084d42e9a44f82d

vim-ml:
On 2024-01-07, Christian Brabandt wrote:
> I have made a few minor changes, which turned out to be even more involved:
> 
>   • Renamed those new groups to Added, Changed and Removed (I initially went
>     with capitalizing the diffAdded to DiffAdded, but since hightlighting
>     groups are not case sensitive, it causes problems for linking the groups in
>     the color schemes, so I had to make the name different and removed the diff
>     prefix)
>   • Changed the links in the default colorschemes
>   • linked those new groups back in the diff syntax file.
>   • capitalize the color names consistently
> 
> That is what I have temporarily queued, not sending out yet, would like to get
> an okay for that.:
>
> https://github.com/chrisbra/vim/commit/0fea0abf5a2e4a9b91df9b1c0084d42e9a44f82d

I tried cloning your repo without pulling down everything,

    $ clone --depth 3 --no-single-branch https://github.com/chrisbra/vim.git

but it didn't include that commit.  The latest relevant commit
appeared to be cdd869fd9, so I checked that out and built it.  The
result looks pretty good on an xterm, tested with the following.

    $ git diff cdd869fd9^ cdd869fd9 | VIMRUNTIME=runtime src/vim -N -u NONE -i NONE --cmd 'set rtp=runtime' -c 'syntax on' -

Regards,
Gary

chrisbra:
> but it didn't include that commit. 

You would have to checkout branch `gh-13825`, e.g.: https://github.com/chrisbra/vim/tree/gh-13825

Let me include it now, we can always further re-adjust

habamax:
Default colors for Added/Changed/Removed are good enough for most of the bundled colorschemes.

The only outlier was `murphy` where green is the normal fg color and it gets in the way with `Added`, so I changed it to white:

image

neutaaaaan:
Do we have a reliable, unequivocal way of pointing out to distro maintainers that some patches are fixes to existing features that don't work properly in the current major or minor version, so they don't rebase upon that without including those fixes ?

As we've seen with the colourscheme work that was pushed late into the vim 8.2 cycle, and all the fixes pushed early into the vim 9.0 cycle, or with this specific pull request, this is sorely needed.

PR #13828, opened 2024-01-06
Improve the documentation for Vim9 classes

zzzyxwvut:
The proposed changes are as follows:
- Emend textual typos
- Emend syntactic errors in examples
- Acknowledge no support for abstract static methods
- Acknowledge the non-ubiquity of instance qualification
- Recognise abstract child classes
- Reword an ambiguous turn of phrase

TODO:

The indentation for examples could also be made uniform; it  
seems that a single tab (0x09) is common (`:h eval.txt`).

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -1,4 +1,4 @@
-*vim9class.txt*	For Vim version 9.1.  Last change: 2024 Jan 01
+*vim9class.txt*	For Vim version 9.1.  Last change: 2024 Jan 06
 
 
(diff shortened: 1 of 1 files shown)


errael:
This looks pretty good to me, correcting various problems. 

The following comment has nothing to do directly with this PR: reviewing, I read a section `Declaring object and class variables` that I don't remember having seen before. I was confused until I realized it was under `Rationale`. The last paragraph does summarize why `var` is required in a class definition. But I had to run a test to _know_ that `static var` is correct and `var static` is an error.

yegappan (approved):

chrisbra:
Thanks!

PR #13830, opened 2024-01-07
Add initial support for Added/Removed/Changed

habamax:
For some of the colorschemes where `diffAdded` and `diffRemoved` were explicitly set up.

Default colors looks ok for most of the bundled colorschemes, however they might be added in the future.


chrisbra:
Thanks!

PR #12637, opened 2023-07-06
fix(tar): fix variable assignment and typo

e-kwsm:

PR #8109, opened 2021-04-15
Rename .tzs to the correct .tzst

C0rn3j:
Fixes #8105

PR #13831, opened 2024-01-07
tar.vim: Fix .tar.zst handling

### LINKED ISSUE #12639: unable to update `.tar.zst`
### Steps to reproduce

```console
$ echo foo > a.dat
$ tar -acf a.tar.zst a.dat
```

`vim -u NORC -N a.tar.zst`, select a.dat, `:s/foo/bar/`, `:wq` (`tarfile::a.dat`), and `:q` (`a.tar.zst`).
However, `a.tar.zst` is not updated, and `a.tar` is created with the updated content:

```console
$ tar -xOf a.tar.zst
foo
$ tar -xOf a.tar
bar
```

### Expected behaviour

`a.tar.zst` is updated and `a.tar` does not exist.

### Version of Vim

9.0.1672 (#12637 is applied)

### Environment

- OS: Debian 12.0
- terminal: GNOME Terminal 3.46.8 using VTE 0.70.3 +BIDI +GNUTLS +ICU +SYSTEMD
- $TERM: xterm-256color

### Logs and stack traces

_No response_

c4rlo:
Updating `.tar.zst` files was broken. Fixes #12639.

Extracting files from `.tar.zst` / `.tzs` files was also broken and works now.

@cecamp: It seems you maintain an "upstream" for this plugin at http://www.drchip.org/astronaut/vim/index.html#TAR, but I didn't see a way of proposing updates there, hence doing it here. Hope this works.

### DIFF
--- a/runtime/autoload/tar.vim
+++ b/runtime/autoload/tar.vim
@@ -468,9 +468,9 @@ fun! tar#Write(fname)
    let compress= "xz -- ".shellescape(tarfile,0)
 "   call Decho("compress<".compress.">")
   elseif tarfile =~# '\.zst'
-   call system("zstd --decompress -- ".shellescape(tarfile,0))
+   call system("zstd --decompress --rm -- ".shellescape(tarfile,0))
    let tarfile = substitute(tarfile,'\.zst','','e')
-   let compress= "zstd -- ".shellescape(tarfile,0)
+   let compress= "zstd --rm -- ".shellescape(tarfile,0)
   elseif tarfile =~# '\.lzma'
    call system("lzma -d -- ".shellescape(tarfile,0))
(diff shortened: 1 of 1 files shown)


c4rlo:
I notice that  #12637 has some overlap with this PR (not in a conflicting way though).

#8109 is also related, which points out the correct short-form file extension for `.tar.zst` is actually `.tzst` not `.tzs`…

chrisbra:
Thanks, I included this as well as the others now with Commit 3a5b3df7764daa058a3e779183e8f38a8418b164. Even so it was non-conflicting overlapping, I had to apply the patches manually, since those changes prevented patch/git apply from applying it 🤷 . Hope it is all fine now.

PR #13832, opened 2024-01-08
fix: command complete vim9 autoload func highlight

habamax:
if you have autoloaded command complete function in vim9script, it shouldn't be highlighted as Error:

```
import autoload "share.vim"
command! -range=% -nargs=? -complete=custom,share.Complete Share share.Paste(<q-args>, <line1>, <line2>)
```

`share.Complete` is a valid complete function.

Before patch:

image

After patch:

image


chrisbra:
Thanks!

PR #13842, opened 2024-01-10
runtime(netrw): Decode multibyte percent-encoding filename correctly

### LINKED ISSUE #13787: netrw doesn't decode multibyte percent-encoding filename correctly
### Steps to reproduce

1. `touch 中文`
2. `vim file:///...../%e4%b8%ad%e6%96%87` (the `file://` url to the file just created)
3. Vim edits a file called "ä¸­æ<96><87>"

### Expected behaviour

Vim should edit the file named `中文`.

### Version of Vim

9.0.2173

### Environment

Arch Linux, UTF-8 locale.

### Logs and stack traces

_No response_

k-takata:
Use `printf("%c")` instead of `nr2char()` to handle '%xx' as a byte.

Close #13787


chrisbra:
Thanks!

Issue #13156, opened 2023-09-22
[vim9script] refactor support


errael:
- compile all files of a project
- populate `quickfix`

In the vim9class spec it mentions difficulties in refactoring. I'm not enough
of a vim user to know what IDE-like tools are available for vim9script. But
one thing that would help is being able to compile every function/method in
the project all at once. Otherwise **changing a member name in a class could
easily be a time bomb**, lying undetected until it after it get's into public
hands.  (of course if you have tests that cover every line of code, it's less
of a problem, but...)

In the vim9class TODO there's an item

```
Make ":defcompile ClassName" compile all functions and methods in the class.
```
which is certainly relevant and could be part of a solution.

Another thought is putting together an antlr4 vim9script parser that could at
least find undefined items and maybe other problems. Lacygoill's vim9-syntax
(been happily using it since the beginning) has some useful lists of things
that would be helpful in putting together a good parser.

errael:
> compile every function/method in the project all at once

Should read "every file and every function/method".

PR #13844, opened 2024-01-11
Support compiling all the methods in a class using :defcompile

yegappan:


src/proto/vim9class.pro
```diff
@@ -31,6 +31,9 @@ void object_free_items(int copyID);
 void emsg_var_cl_define(char *msg, char_u *name, size_t len, class_T *cl);
 void method_not_found_msg(class_T *cl, vartype_T v_type, char_u *name, size_t len);
 void member_not_found_msg(class_T *cl, vartype_T v_type, char_u *name, size_t len);
+void defcompile_class(class_T *cl);
+void defcompile_classes_in_script();
```
errael:
```suggestion
void defcompile_classes_in_script(void);
```

errael:
There's #13156. With this PR, it might be possible to compile everything in a project. Any hints on how to set that up? And I'm wondering about checking imports, does that just happen if they're referenced from a class that's compiled? 

I guess there's also the issue of checking script level references, but that has nothing to do with this PR.

runtime/doc/vim9.txt
```diff
@@ -1260,10 +1260,9 @@ Script-local variables in a |Vim9| script must be declared at the script
 level.  They cannot be created in a function, also not in a legacy function.
 
 						*:defc* *:defcompile*
-:defc[ompile]		Compile functions defined in the current script that
-			were not compiled yet.
+:defc[ompile]		Compile functions and classes defined in the current
+			script that were not compiled yet.
 			This will report any errors found during compilation.
-			This excludes functions defined inside a class.
```
chrisbra, line 1266: Can we need to add a reference to `:defc[ompile] MyClass`.  What happens, for `:defcompile FooBar` when FooBar is a class and a function?

src/proto/vim9class.pro
yegappan: Thanks.  Incorporated in the latest update to the PR.

runtime/doc/vim9.txt
yegappan: > Can we need to add a reference to `:defc[ompile] MyClass`.

I have added a help tag for compiling a class.

> What happens, for `:defcompile FooBar` when FooBar is a class and a function?

You cannot have a function and a class with the same name in a script.  Internally for every class, a script local variable with the same name is created.  If you use the same name for a function and a class, then you will
get the `E1041: Redefining script item: "FooBar"` error.

yegappan:
> There's #13156. With this PR, it might be possible to compile everything in a project. Any hints on how to set that up?
>

The `:defcompile` command currently compiles only the `def` functions and classes defined in a script.  It doesn't compile
functions and classes defined in imported scripts.  So currently using the `:defcompile` command you cannot compile the
entire project.  But this can be extended (if needed).

> And I'm wondering about checking imports, does that just happen if they're referenced from a class that's compiled?

Are you referring to compiling imported functions and classes?

> 
> I guess there's also the issue of checking script level references, but that has nothing to do with this PR.

errael:
> > There's #13156. With this PR, it might be possible to compile everything in a project. Any hints on how to set that up?
> 
> The `:defcompile` command currently compiles only the `def` functions and classes defined in a script. It doesn't compile functions and classes defined in imported scripts. So currently using the `:defcompile` command you cannot compile the entire project. But this can be extended (if needed).

Sorry, I wasn't clear. Given that `:defcompile` now compiles everything in a file. I'm wondering how/if I can best set up a script, like `bash` or `vim9` or ???, that would compile all the files in a project and record any problems.

I've got a vim9script project which has some very old parts. I'd like to clean some things up without breaking anything.

> > And I'm wondering about checking imports, does that just happen if they're referenced from a class that's compiled?
> 
> Are you referring to compiling imported functions and classes?

Not necessarily, but maybe that is the implication. If I'm refactoring and want to make sure that I didn't break something in a different file, I'm wondering how I would do that.

I guess the first step would be creating a cross-reference; a cross reference would help a lot. Given that there's `:export`/`:import` maybe I can put together a script that scans a list of files. Handling `:export` might be straightforward; but _finding all the references to imported items seems tricky_. With the import/export info for all the files, combine into a **where used** indexed by var/func/class/class-var/class-method. Hmm, finding out about class variables, and where they're referenced in other files, might be a problem. Anyway that's what I'm wondering about. If I was going to hack parts of this I'd probably start with `bash`/`gawk`. Some vim support to get the list of _exported items_ and _import references_, maybe using options to `:defcompile`, would simplify things. With the **where used**, manual refactoring would be less hit and miss. Hmm, an accurate **where used** could form the basis of an automated/semi-automated refactoring tool.

> > I guess there's also the issue of checking script level references, but that has nothing to do with this PR.

PR #13846, opened 2024-01-12
i_CTRL-R- doesn't work for multibyte chars in Replace mode

zeertzjq:
Problem:  i_CTRL-R- doesn't work for multibyte chars in Replace mode.
Solution: Use mb_charlen() and del_chars() instead.


chrisbra:
thanks, makes sense!

PR #13849, opened 2024-01-12
Fixing Windows gVim feature of interactive pipelining into commands

MiguelBarro:
On Windows gvim uses pipelines instead of files for redirection. Originally (4b9669f1dc circa 2011) vim allows interaction with commands through the input pipeline.
For example, calling `xxd` without arguments would trigger `xxd` to read from stdin and the user could type the input. If the user typed "abcd(CTRL-D)" it would get:

```vim
:read !xxd
00000000: 6162 6364 650d 0a                        abcde..
```
Unfortunately since commit 77fc0b02e5 (2022) it no longer works because CTRL-D (which in windows is map to EOF) is not captured anymore. The only way of leaving interaction is CTRL-C which kills `xxd` and provides no output.

The actual issue is that `_OnChar` function has modified the encoding of CTRL-D in the input buffer. Thus, `mch_system_piped` can no longer identified CTRL-D. The pull request fixes the CTRL-D detection and restores the former behavior.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -5194,7 +5194,7 @@ mch_system_piped(char *cmd, int options)
 		     */
 		    if ((len == 1 || len == 4 ) && cmd != NULL)
 		    {
-			if (ta_buf[0] == Ctrl_C
+			if (ta_buf[0] == Ctrl_D
 			    || (ta_buf[0] == CSI
 				&& ta_buf[1] == KS_MODIFIER
 				&& ta_buf[3] == Ctrl_D))
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -5176,7 +5176,7 @@ mch_system_piped(char *cmd, int options)
 		{
 		    /*
 		     * For pipes: Check for CTRL-C: send interrupt signal to
-		     * child.  Check for CTRL-D: EOF, close pipe to child.
+		     * child.
 		     */
 		    if (len == 1 && cmd != NULL)
 		    {
@@ -5186,10 +5186,21 @@ mch_system_piped(char *cmd, int options)
 			// now put 9 as SIGKILL
 			    TerminateProcess(pi.hProcess, 9);
 			}
-			if (ta_buf[ta_len] == Ctrl_D)
+		    }
+
+		    /*
+		     * Check for CTRL-D: EOF, close pipe to child.
+		     * Ctrl_D receives no special treatment in _OnChar() as Ctrl_C does
+		     */
+		    if (len == 4 && cmd != NULL)
+		    {
+			if (ta_buf[0] == CSI &&
+			    ta_buf[1] == KS_MODIFIER &&
+			    ta_buf[3] == Ctrl_D)
 			{
 			    CloseHandle(g_hChildStd_IN_Wr);
 			    g_hChildStd_IN_Wr = NULL;
+			    len = 0;
 			}
 		    }
 
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -5190,13 +5190,14 @@ mch_system_piped(char *cmd, int options)
 
 		    /*
 		     * Check for CTRL-D: EOF, close pipe to child.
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. Is it possible to get a test for this?

k-takata:
PR #12752 is going to restore the old behavior.
Is it better to keep the two ways of Ctrl-D checking?

MiguelBarro:
@k-takata  I have restore the old behavior to make the fix independent of the chosen option introduced in https://github.com/vim/vim/pull/12752.
@chrisbra I was not able to replicate the user interaction in a test. I tried:
+ Using `:normal`, as bellow, executes the `:read!` command but ignores the user keystrokes *vim\<C-d>*. 
```vim
    :exe "normal :read !xxd\<CR>vim\<C-d>"
```
+ Using `feedkeys()` I got the same issue:
```vim
    :call feedkeys(":read !xxd\<CR>vim\<C-d>")
```
+ Using the test functions as in *vim/src/testdir/test_mswin_event.vim* the same happens:
```vim
    :call SendKeyWithModifiers(VK.OEM_PERIOD, MOD_MASK_SHIFT) " Spanish keyboard
    :call SendKeyStr([VK.KEY_R, VK.KEY_E, VK.KEY_A, VK.KEY_D, VK.SPACE])
    :call SendKeyWithModifiers(VK.KEY_1, MOD_MASK_SHIFT) " Spanish keyboard
    :call SendKeyStr([VK.KEY_X, VK.KEY_X, VK.KEY_D, VK.ENTER, VK.KEY_V, VK.KEY_I, VK.KEY_M])
    :call SendKeyWithModifiers(VK.KEY_D, MOD_MASK_CTRL)
```
My educated guess is that `mch_system_piped()` own message loop implementation effectively disables any test chance 💩.

src/os_win32.c
```diff
@@ -5186,10 +5186,22 @@ mch_system_piped(char *cmd, int options)
...
-			if (ta_buf[ta_len] == Ctrl_D)
+		    }
+
+		    /*
+		     * Check for CTRL-D: EOF, close pipe to child.
+		     * Ctrl_D may be decorated by _OnChar()
+		     */
+		    if ((len == 1 || len == 4 ) && cmd != NULL)
+		    {
+			if (ta_buf[0] == Ctrl_C
```
k-takata: Is this correct?  Not Ctrl_D?
Additionally, the original code used `ta_len` for the index.
```suggestion
			if (ta_buf[ta_len] == Ctrl_D
```

src/os_win32.c
```diff
@@ -5186,10 +5186,22 @@ mch_system_piped(char *cmd, int options)
...
+			    || (ta_buf[0] == CSI
+				&& ta_buf[1] == KS_MODIFIER
+				&& ta_buf[3] == Ctrl_D))
```
k-takata, line 5200: `ta_len` is not used. Is it correct?

src/os_win32.c
MiguelBarro: My mistake! It must be `Ctrl_D`.

src/os_win32.c
MiguelBarro: According to the code the only way `len != 0` is having a `ta_en==0`. Thus, using either option is possible, though, I believe using 0 is consistent with the other comparison and less confusing.

MiguelBarro:
@chrisbra I have discovered the reason why testing is not possible.

 It turns out that in `
vim/src/gui_w32.c` all the calls to the WIN32 API [SendInput](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendinput) only involve a single keystroke.
When the `:read!` command is executed the `mch_system_piped()` function stalls the script execution for the user interaction. Only after the user interaction is finished the `SendInput()` calls are performed for the remaining virtual keystrokes.
In order to enable a test, another `test_gui_w32_sendevent_keyboard()` function which accepts multiple events is required.

chrisbra:
okay thanks. No worries. Then let's just fix it for now. Thanks for that.

PR #13853, opened 2024-01-13
fix: make i_ctrl-] not activate InsertCharPre

altermo:
Currently pressing `ctrl-]` in insert-mode activates InsertCharPre. This is probably unintended as `ctrl-]` doesn't insert a character but does an action.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -5371,6 +5371,9 @@ do_insert_char_pre(int c)
     if (!has_insertcharpre())
 	return NULL;
 
+    if (c == Ctrl_RSB)
+	return NULL;
+
     if (has_mbyte)
 	buf[(*mb_char2bytes)(c, buf)] = NUL;
     else
--- a/src/testdir/test_edit.vim
+++ b/src/testdir/test_edit.vim
@@ -2123,4 +2123,39 @@ func Test_edit_shift_bs()
   call StopVimInTerminal(buf)
 endfunc
 
+func Test_edit_Ctrl_RSB()
+  new
+  let g:triggered = []
+  autocmd InsertCharPre <buffer> let g:triggered += [v:char]
+
+  " i_CTRL-] should not trigger InsertCharPre
+  exe "normal! A\<C-]>"
+  call assert_equal([], g:triggered)
+
+  " i_CTRL-] should expand abbreviations but not trigger InsertCharPre
+  inoreabbr <buffer> f foo
+  exe "normal! Af\<C-]>a"
+  call assert_equal(['f', 'f', 'o', 'o', 'a'], g:triggered)
+  call assert_equal('fooa', getline(1))
+
+  " CTRL-] followed by i_CTRL-V should not expand abbreviations
+  " i_CTRL-V doesn't trigger InsertCharPre
+  call setline(1, '')
+  exe "normal! Af\<C-V>\<C-]>"
+  call assert_equal("f\<C-]>", getline(1))
+
+  let g:triggered = []
+  call setline(1, '')
+
+  " Also test assigning to v:char
+  autocmd InsertCharPre <buffer> let v:char = 'f'
+  exe "normal! Ag\<C-]>h"
+  call assert_equal(['g', 'f', 'o', 'o', 'h'], g:triggered)
+  call assert_equal('ffff', getline(1))
+
+  autocmd! InsertCharPre
+  unlet g:triggered
+  bwipe!
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab


chrisbra:
Thanks!
While you are at it, can you please add a test, that `<c-]>` correctly triggers expansion of a macro? Also, an additional test for `<c-v><c-]>` in insert mode please (I suppose this should trigger `InsertCharPre` autocommand again)?

altermo:
Okay, I added the test.

Also, other characters after `<C-v>` don't activate `InsertCharPre`, so `<C-v><C-]>` should not activate it.

src/testdir/test_i_ctrl_rsb.vim
```diff
@@ -0,0 +1,24 @@
+" Test for InsertCharPre autocmd
+
```
zeertzjq: This comment doesn't match the file name

zeertzjq: Or, maybe, these functions should just be in `test_edit.vim` rather than a new file, and should follow the convention (e.g. 2-space indent) of existing test functions.

src/testdir/test_i_ctrl_rsb.vim
```diff
@@ -0,0 +1,24 @@
...
+def Test_i_ctrl_rsb_not_trigger_insertcharpre()
+    new
+    autocmd InsertCharPre <buffer> v:char = "a"
+    exe "normal! i\<C-]>"
+    call assert_equal('', getline(1))
+    bwipeout!
```
zeertzjq: Also need to delete the autocommand

src/testdir/test_i_ctrl_rsb.vim
```diff
@@ -0,0 +1,24 @@
...
+enddef
+
+def Test_i_ctrl_rsb_expand_abbr()
+    new
+    abbr <buffer> f foo
+    exe "normal! if\<C-]>a"
+    call assert_equal('fooa', getline(1))
+    bwipeout!
```
zeertzjq, line 16: Also need to delete the abbreviation

zeertzjq:
Hmm, also i_CTRL-V currently doesn't trigger `InsertCharPre`, regardless of the inserted character.

zeertzjq:
I think it's better to put these in a single function, so further adjustments will be easier:
```diff
diff --git a/src/testdir/test_edit.vim b/src/testdir/test_edit.vim
index d662f4a0f..36e052556 100644
--- a/src/testdir/test_edit.vim
+++ b/src/testdir/test_edit.vim
@@ -2123,4 +2123,39 @@ func Test_edit_shift_bs()
   call StopVimInTerminal(buf)
 endfunc
 
+func Test_edit_Ctrl_RSB()
+  new
+  let g:triggered = []
+  autocmd InsertCharPre <buffer> let g:triggered += [v:char]
+
+  " i_CTRL-] should not trigger InsertCharPre
+  exe "normal! A\<C-]>"
+  call assert_equal([], g:triggered)
+
+  " i_CTRL-] should expand abbreviations but not trigger InsertCharPre
+  inoreabbr <buffer> f foo
+  exe "normal! Af\<C-]>a"
+  call assert_equal(['f', 'f', 'o', 'o', 'a'], g:triggered)
+  call assert_equal('fooa', getline(1))
+
+  " CTRL-] followed by i_CTRL-V should not expand abbreviations
+  " i_CTRL-V doesn't trigger InsertCharPre
+  call setline(1, '')
+  exe "normal! Af\<C-V>\<C-]>"
+  call assert_equal("f\<C-]>", getline(1))
+
+  let g:triggered = []
+  call setline(1, '')
+
+  " Also test assigning to v:char
+  autocmd InsertCharPre <buffer> let v:char = 'f'
+  exe "normal! Ag\<C-]>h"
+  call assert_equal(['g', 'f', 'o', 'o', 'h'], g:triggered)
+  call assert_equal('ffff', getline(1))
+
+  autocmd! InsertCharPre
+  unlet g:triggered
+  bwipe!
+endfunc
+
 " vim: shiftwidth=2 sts=2 expandtab
```

zeertzjq:
Hmm, now I doubt whether this change is really desired, as one may indeed want to check for `"\<C-]>"` in `InsertCharPre` autocommand to detect if user is expanding an abbreviation.

src/testdir/test_i_ctrl_rsb.vim
altermo: Technically not needed, as Vim automatically deletes a buffer-local autocmd when the buffer is deleted(wiped).

altermo:
Well, that may be a use case, but I feel like it's too much of a special case, and there are other ways of doing it.
The documentation (and the [c comment](https://github.com/vim/vim/blob/master/src/vim.h#L1391)) states that it'll be activated by a character before it gets inserted (we are not inserting a character but doing an action).
This will leave the user thinking that `<C-]>` will not activate `InsertCharPre`, so if they want to detect `<C-]>`, they'll probably do it another way unless they know this undocumented exception as this is the only special key in insert mode that activates `InsertCharPre` (I checked the source code); Why should this be the exception?

zeertzjq:
Yes, I also think so, but it doesn't seem easy to find another way to do such a thing.

altermo:
Yeah...
Why I made this pull request:
Currently, the only way to expand the abbreviation easily is to use `<C-]>`, so some auto-pair (like [jiangmiao/auto-pairs](https://github.com/jiangmiao/auto-pairs/blob/39f06b873a8449af8ff6a3eee716d3da14d63a76/plugin/auto-pairs.vim#L556)) plugins when they add a space expr-mapping the return value has a `<C-]>` before the space to activate abbreviation. The problem is that multi-cursor plugins typically use InsertCharPre to get the inserted char, and this makes them insert a `<C-]>`.

altermo:
I have thought about it and... I'll just create a pull request for Vim that adds a small snippet of documentation about this, and I'll reopen a similar pull request on Neovim, as there you can easily detect `<C-]>` (using `vim.on_keys`), so there would be no downsides of changing this there.

chrisbra:
Wait a second, not so fast please. 
So we have `<c-]>` triggering InsertCharPre, but `i_CTRL-T` and `i_CTRL-D` not triggereing it. That is inconsistent. 
I would also think, a user wanting to check if `<c-]>` was pressed, could also map `<c-]>` and act on it if required, so if we change it, there is at least another way to check if this was pressed.

PR #13855, opened 2024-01-13
runtime(doc): revision makefiles

RestorerZ:
`Make_mvc.mak` and `Makefile`.                                                                                                                                                                                                              
Added --clean key to Vim call from "test_urls" rule.                                                                                                                                                                        
Continued #13589 :
Use predefined variables (`$@`, `$<`) instead of the actual file names.


chrisbra:
thanks!

PR #13856, opened 2024-01-13
runtime(lang): revision makefiles

RestorerZ:
`Make_mvc.mak` and `Makefile`
The "clean" rule has been added.
`Make_mvc.mak`
Continued #13589 :
Use predefined variables (`$@`, `$<`) instead of the actual file names.
Regenerate converted "menu_cs_cz.iso_8859-2.vim" menu file


chrisbra:
thanks!

PR #13857, opened 2024-01-13
src/po: revision makefiles and document files

RestorerZ:
`Make_mvc.mak`
Added check of setting of LANGUAGE environment variable and, if necessary,
automatic assignment of the value based on the one used in the OS. With
corresponding message output to the user. This allows all available rules to
work with the intended user preferences.
Removed the phony rule "checklanguage" and related dependencies, which should
make future maintenance easier.
`Make_mvc.mak` and `Makefile`
Added "originals" rule to match "languages" rule from src/Makefile.
Continued #13589 :
Use predefined variables (`$@`, `$<` for `Makefile` (UNIX) and `$@`, `$?` for `Make_mvc.mak` (MSFT)) instead of the actual file names.
Other corrections.
`README.txt`, `README_mvc.txt` and `README_mingv.txt`
Updated documentation.

### DIFF
--- a/src/po/Make_mvc.mak
+++ b/src/po/Make_mvc.mak
@@ -1,13 +1,33 @@
 # Makefile for the Vim message translations for MSVC
 # (based on make_ming.mak)
 #
-# Mike Williams <mrw@eandem.co.uk>
+# Mike Williams, <mrw@eandem.co.uk>
+# 06.01.24, Restorer, <restorer@mail2k.ru>
 #
 # Please read README_mvc.txt before using this file.
 #
 
-!IF [powershell -nologo -noprofile "exit $$psversiontable.psversion.major"] == 2
+!IF [powershell.exe -nologo -noprofile "exit $$psversiontable.psversion.major"] == 2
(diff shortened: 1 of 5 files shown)


chrisbra:
Same comment applies here: `$<` and `$@` are not portable as far as I know.

jamessan:
> `$<` and `$@` are not portable as far as I know.

They're defined by [POSIX](https://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html#tag_04_84_13_07), and already in use by the makefiles.

chrisbra:
oh are they? Okay I take everything back then

RestorerZ:
> > `$<` and `$@` are not portable as far as I know.
> 
> They're defined by [POSIX](https://pubs.opengroup.org/onlinepubs/009695399/utilities/make.html#tag_04_84_13_07), and already in use by the makefiles.

Variable (macro) `$<` for "nmake.exe" (MSFT) with some restrictions.

> $<
> The dependent file that has a later time stamp than the current target. You
> can use this macro only in commands in inference rules.

PR #13859, opened 2024-01-13
runtime(tutor) revision makefiles

RestorerZ:
`Make_mvc.mak` and `Makefile`.                                                                                                                                                                                                              
Continued #13589 :
Use predefined variables (`$@`, `$<` for `Makefile` (UNIX) and `$@`, `$?` for `Make_mvc.mak` (MSFT)) instead of the actual file names.
The "clean" rule has been added.

### DIFF
--- a/runtime/tutor/Make_mvc.mak
+++ b/runtime/tutor/Make_mvc.mak
@@ -27,6 +27,9 @@ ICONV = "$(ICONV_PATH)\iconv.exe"
 
 RM = del /q
 CP = copy /y
+PS = PowerShell.exe
+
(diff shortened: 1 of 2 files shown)


chrisbra:
I think those variables: `$<` and `$@` are not portable, so we should not use them.

RestorerZ:
> I think those variables: `$<` and `$@` are not portable, so we should not use them.

Yes, not exactly portable. For UNIX they used `$<` and `$@`, and for MSFT `$?` and `$@`. In the description of PR I made a corresponding explanation.

PR #13867, opened 2024-01-15
Odin lang filetype/syntax/indent plugin

habamax:
Vim has Odin filetype detection but misses actual Odin filetype, syntax and indent.

Repo: https://github.com/habamax/vim-odin

image

image

### DIFF
--- a/runtime/ftplugin/odin.vim
+++ b/runtime/ftplugin/odin.vim
@@ -1,7 +1,7 @@
 vim9script
 
 # Vim filetype plugin file
-# Language: odin
+# Language: Odin
 # Maintainer: Maxim Kim <habamax@gmail.com>
 # Website: https://github.com/habamax/vim-odin
 # Last Change: 2024-01-15
@@ -12,7 +12,9 @@ endif
 b:did_ftplugin = 1
 
 b:undo_ftplugin = 'setlocal commentstring<'
+      \ .. '| setlocal comments<'
       \ .. '| setlocal suffixesadd<'
 
 setlocal suffixesadd=.odin
 setlocal commentstring=//%s
+setlocal comments=s1:/*,mb:*,ex:*/,://
--- a/runtime/ftplugin/odin.vim
+++ b/runtime/ftplugin/odin.vim
@@ -1,0 +1,17 @@
+vim9script
+
+# Vim filetype plugin file
+# Language: odin
+# Maintainer: Maxim Kim <habamax@gmail.com>
+# Website: https://github.com/habamax/vim-odin
+
+if exists("b:did_ftplugin")
+    finish
+endif
+b:did_ftplugin = 1
+
+b:undo_ftplugin = 'setlocal commentstring<'
+      \ .. '| setlocal suffixesadd<'
+
+setlocal suffixesadd=.odin
+setlocal commentstring=//%s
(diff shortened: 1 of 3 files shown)


runtime/ftplugin/odin.vim
```diff
@@ -0,0 +1,18 @@
+vim9script
+
+# Vim filetype plugin file
+# Language: odin
```
dkearns: s/odin/Odin/

runtime/ftplugin/odin.vim
```diff
@@ -0,0 +1,18 @@
...
+if exists("b:did_ftplugin")
+    finish
+endif
+b:did_ftplugin = 1
+
+b:undo_ftplugin = 'setlocal commentstring<'
+      \ .. '| setlocal suffixesadd<'
+
+setlocal suffixesadd=.odin
+setlocal commentstring=//%s
```
dkearns, line 19: Could we also set `comments` for formatting commands?

runtime/syntax/odin.vim
```diff
@@ -0,0 +1,103 @@
...
+
+syntax match odinAddressOf "&" display
+syntax match odinDeref "\^" display
+
+syntax match odinMacro "#\<\w\+\>" display
+
+syntax match odinTemplate "$\<\w\+\>"
+
+syntax region odinLineComment start=/\/\// end=/$/  contains=@Spell,odinTodo
+syntax region odinBlockComment start=/\v\/\*/ end=/\v\*\// contains=@Spell,odinTodo,odinBlockComment
```
dkearns: The `\v` modifier isn't doing any work here.

dkearns (changes requested):
Thanks, I was just watching an interview with the ginger one.

You're probably aware but Vim9-script runtime files won't be picked up by Neovim.  They're not rewriting them in legacy script.

habamax:
> Thanks, I was just watching an interview with the ginger one.

Indeed language looks promising.

> You're probably aware but Vim9-script runtime files won't be picked up by Neovim. They're not rewriting them in legacy script.

I think they can update it to legacy vimscript if needed? It shouldn't be too difficult, it is `GetOdinIndent()` function and a bit of script variable assignments here and there.

habamax:
found an issue with the last commit (wrt indentation), investigating.

habamax:
while cindent does almost what I think Odin needs, on edge cases it fails miserably. Let me convert this to draft for now, I will need to re-implement indentation algorithm from scratch.

habamax:
Indentation looks ok so far -- tested it on `demo.odin` bundled with the language.

clason:
> I think they can update it to legacy vimscript if needed? It shouldn't be too difficult, it is GetOdinIndent() function and a bit of script variable assignments here and there.

I won't, for the record; I'll just include the syntax file (and possibly the ftplugin; that indeed looks fairly trivial to translate back). Non-trivial vim9script files will be skipped until we get automated transpiling working (which is looking unlikely soon). If someone were to contribute a manual translation that is close enough to the original so changes can be backported (again, manually) easily, that could also be accepted. (Nobody has, so far, so there's no official policy around that yet.)

(Not complaining, to be clear; just wanted to let you know what is what from our side since you do take it into consideration -- which I appreciate. No hard feelings whatever you do!)

chrisbra:
Thanks!

PR #13870, opened 2024-01-15
Runtime: Fortran syntax

Ajit-Thakkar:
- Support most remaining features of Fortran 2018/2023 
- Small improvements to folding etc.
- Code cleanup: use \? instead of mix of \= and \?


chrisbra:
thanks!

PR #13871, opened 2024-01-16
installer: Add runtime/keymap

### LINKED ISSUE #331: windows version of vim lacks `runtime/keymap/*`
Is there a reason why vim version lacks all of the keymaps from `runtime/keymap` directory?

image

I remember they were never part of a windows vim as long as I used it(since many years ago) and I specifically carried selected keymaps from vim's runtime in my `~/vimfiles`.

Can we add them?

k-takata:
Fixes https://github.com/vim/vim-win32-installer/issues/331


chrisbra:

PR #13872, opened 2024-01-16
Insert mode not stopped if closing prompt buffer modifies hidden buffer

zeertzjq:
Problem:  Insert mode not stopped if an autocommand modifies a hidden
          buffer while closing a prompt buffer.
Solution: Don't set b_prompt_insert if stop_insert_mode is already set.


chrisbra:
thanks!

PR #13875, opened 2024-01-17
Modifying hidden buffer still interferes with prompt buffer mode changes

zeertzjq:
Problem:  Modifying a hidden buffer still interferes with prompt buffer
          mode changes.
Solution: Save and restore b_prompt_insert.


chrisbra:
Seems fine. I was worried about removing the `save_curbuf`, but since it is not used in `aucmd_restbuf()` that is safe.

Thanks!

PR #13877, opened 2024-01-17
Errors for setting 'listchars' or 'fillchars' are too vague

zeertzjq:
Problem:  Errors for setting 'listchars' or 'fillchars' are too vague.
Solution: Include the field name in error message.


chrisbra:
Thank you. I was slightly confused, why you removed the doc note about `{not available when compiled without the |+folding| feature}`. But this change seems correct, even so I couldn't find out, when the 'fillchars'  option has been graduated.

PR #13878, opened 2024-01-17
Unnecessary loop in getvcol()

zeertzjq:
Problem:  Unnecessary loop in getvcol().
Solution: Compare next char position with pos->col directly.

The loop below already handles end of line before checking for posptr,
and the next char is after pos->col whether pos->col is at the start or
in the middle of the char in question, so neither the NUL check nor the
mb_head_off() are needed when comparing the position of the next char
with pos->col directly.

### DIFF
--- a/src/charset.c
+++ b/src/charset.c
@@ -1482,7 +1482,6 @@ getvcol(
 {
     colnr_T	vcol;
     char_u	*ptr;		// points to current char
-    char_u	*posptr;	// points to char at pos->col
     char_u	*line;		// start of the line
     int		incr;
     int		head;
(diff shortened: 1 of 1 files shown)


zeertzjq:
The NUL check was initially added in [patch 8.0.0307](https://github.com/vim/vim/commit/955f198fc546cc30a34361932d3f454a61df0efa) to prevent an memory error in the `mb_head_off()`, so the NUL check isn't needed if `md_head_off()` isn't needed, as reaching NUL is already checked in the loop below.

chrisbra:
thanks!

PR #13879, opened 2024-01-17
Calling get_breakindent_win() repeatedly when computing virtcol

zeertzjq:
Problem:  Calling get_breakindent_win() repeatedly when computing
          virtual column, and get_breakindent_win() does a STRCMP() on
          the whole line since patch 9.0.0016.
Solution: Cache the result, since the line doesn't change.


chrisbra:
thank you!

PR #13880, opened 2024-01-17
Make size type returned by -i option compatible with C standard

### LINKED ISSUE #13876: xxd Code Generation is not compatible with C standard for large arrays
### Steps to reproduce

The option -i is supposed to generate an array length variable as an int. The C standard only guarantees 16 bits for an int. An array greater than 2^16 entries should cause issues.

### Expected behaviour

Generate a type that can hold the maximum array size for a platform, e.g. size_t

### Version of Vim

I read the source code on the repo

### Environment

Irrelevant 

### Logs and stack traces

_No response_

Kuratius:
resolves #13876

### DIFF
--- a/src/testdir/test_xxd.vim
+++ b/src/testdir/test_xxd.vim
@@ -103,7 +103,7 @@ func Test_xxd()
     unsigned char XXDfile[] = {
       0x54, 0x45, 0x53, 0x54, 0x61, 0x62, 0x63, 0x64, 0x30, 0x39, 0x0a
     };
-    unsigned int XXDfile_len = 11;
+    size_t XXDfile_len = 11;
   [CODE]
 
   call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
@@ -119,7 +119,7 @@ func Test_xxd()
       unsigned char XXDFILE[] = {
         0x54, 0x45, 0x53, 0x54, 0x61, 0x62, 0x63, 0x64, 0x30, 0x39, 0x0a
       };
-      unsigned int XXDFILE_LEN = 11;
+      size_t XXDFILE_LEN = 11;
     [CODE]
     call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
   endfor
@@ -233,7 +233,7 @@ func Test_xxd()
       unsigned char varName[] = {
         0x54, 0x45, 0x53, 0x54, 0x61, 0x62, 0x63, 0x64, 0x30, 0x39, 0x0a
       };
-      unsigned int varName_len = 11;
+      size_t varName_len = 11;
     [CODE]
 
     call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
@@ -247,7 +247,7 @@ func Test_xxd()
     unsigned char StdIn[] = {
       0x54, 0x45, 0x53, 0x54, 0x61, 0x62, 0x63, 0x64, 0x30, 0x39, 0x0a
     };
-    unsigned int StdIn_len = 11;
+    size_t StdIn_len = 11;
   [CODE]
   call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
 
@@ -263,7 +263,7 @@ func Test_xxd()
       unsigned char VARNAME[] = {
         0x54, 0x45, 0x53, 0x54, 0x61, 0x62, 0x63, 0x64, 0x30, 0x39, 0x0a
       };
-      unsigned int VARNAME_LEN = 11;
+      size_t VARNAME_LEN = 11;
     [CODE]
     call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
   endfor
--- a/src/xxd/xxd.c
+++ b/src/xxd/xxd.c
@@ -977,7 +977,7 @@ main(int argc, char *argv[])
       if (varname != NULL)
 	{
 	  fputs_or_die("};\n", fpo);
-	  FPRINTF_OR_DIE((fpo, "unsigned int %s", isdigit((unsigned char)varname[0]) ? "__" : ""));
+	  FPRINTF_OR_DIE((fpo, "size_t %s", isdigit((unsigned char)varname[0]) ? "__" : ""));
 	  for (e = 0; (c = varname[e]) != 0; e++)
 	    putc_or_die(isalnum((unsigned char)c) ? CONDITIONAL_CAPITALIZE(c) : '_', fpo);
 	  FPRINTF_OR_DIE((fpo, "_%s = %d;\n", capitalize ? "LEN" : "len", p));


chrisbra:
Thank you. Can you please create a test for it? See [test_xxd.vim](https://github.com/vim/vim/blob/master/src/testdir/test_xxd.vim)

Kuratius:
> Thank you. Can you please create a test for it? See [test_xxd.vim](https://github.com/vim/vim/blob/master/src/testdir/test_xxd.vim)

Do you mean as in checking if sizeof(type returned by xxd) >= sizeof(size_t) ? Because otherwise there is nothing to test, only to change it so existing tests, which have int hardcoded as the type dont fail.

chrisbra:
No I'd like to have a test, that verifies, that `xxd -i` now generates a `size_t` array

Kuratius:
> No I'd like to have a test, that verifies, that `xxd -i` now generates a `size_t` array

Do you mean changing the existing tests that check the type of the size variable, to use size_t ?

There's no "size_t" array being generated, only a size variable that has the type size_t.

The only platforms a test for this could fail on would be ones that don't follow the C standard, in which case I'd say having a test is kind of redundant.

Do you mean creating a test where the version of xxd without this change fails? You can maybe force it even without a 16 bit platform if you generate an array that is more than 4-5 GB ?

Otherwise you can probably force the compiler optimizations to mess up your code and use that as the test, as e.g. there are cases where using a type for a variable that is too small causes compiler optimizations to fail.

See e.g. the first example here: 
https://mohitmv.github.io/blog/Shocking-Undefined-Behaviour-In-Action/

Kuratius:
Also, separate from this, it may be desirable to make sure that arrays larger than 2^32 (a few GB) don't fail to announce their size correctly, but that would require actual tests, or trusting the compiler to determine it with sizeof(array)/sizeof(array[0]) instead of letting vim handle printing the size from a variable.

night199uk:
Heads up that this probably breaks a bunch of software that uses -i for source code generation, as they will now potentially need to include stddef.h IF xxd > latest. I just had to troubleshoot at least one project that no longer builds with build tools updated to xxd latest. 

Might be useful to add a flag to revert to the native unsigned int type to make migration easier than having to update source code templates everywhere.

Kuratius:
Maybe this helps?

https://github.com/Szybet/InkWatchy/blob/master/resources/global_functions.sh

https://github.com/Szybet/InkWatchy/blob/2c085c3a1db3a38fb6a814c44cd1b4ea0934e33a/resources/xxd_legacy.sh (this one needs xxd input with -i from a pipe, so xxd -i |xxdwrapper )
You will need to edit build scripts either way.

Otherwise I'd suggest coming up with a test environment have that doesn't build, as apparently all the existing test. environments have the type available by default or don't include a test.
What project were you building and how?

glebfm:
Before this change, a file generated using `xxd -i` compiled successfully without requiring any additional edits.  I suppose that `xxd` should simply use the `long unsigned int` type for a very long array instead of breaking backward compatibility.

Kuratius:
Are you by any chance compiling  C projects which use headers in the wrong order? Say if you include  a generated file from xxd -i before any of the common standard headers that nearly every project uses, then you might get this error. Or if you're trying to compile this as C++ code even though it's only supposed to generate C.
I suspect all the tests have size_t available for this reason.


It's more complicated to support fully dynamic type generation and doesn't have the same guarantees as size_t with regard to choosing an appropriate type on a platform. You might end up with a type larger than what is natively supported on a 16 bit system for example.

It's also better to fail to compile than to compile with a bug, as using an array size that is too large for int isn't considered a compiler error and the warning might be overlooked when compiling a large project.

glebfm:
1. My point is that before this change, ` xxd -i` didn't require a program to include any additional headers. Therefore, this change introduces a serious regression.
2. According to the C standard, `size_t` is at least 16 bits, `unsigned int` is at least 16 bits, `unsigned long` is at least 32 bits, and `unsigned long long` is at least 64 bits.  So, changing `unsigned int` to `size_t` did not solve any problem anyway.

xxd should use a type that is appropriate (based on the standard, not the platform it is compiled for), according to the size of an array.

Kuratius:
> 2\. So, changing `unsigned int` to `size_t` did not solve any problem anyway.

I think maybe you didn't read the issue. A platform can have 32 bit addressing but can have an int type that is 16 bits. We already have similar issues with 64 bit platforms using 64 bit addressing but using 32 bit int types.

glebfm:
> > 2. So, changing `unsigned int` to `size_t` did not solve any problem anyway.
> I think maybe you didn't read the issue.
Ok, I agree that sizeof(size_t) is usually greater than sizeof(unsigned int) nowdays, but you can't rely on that based on the standard.

>  Int is not guaranteed to hold the size of an array for a platform.
Yes, if an array size doesn't fit in 16 bits you should use the type that is guaranteed to be at least 32 bit (same for 32 -> 64 bits).  size_t is not guaranteed to be more then 16 bits.

Kuratius:
You can't store an array in memory that is bigger than size_t on any platform.

Kuratius:
Technically you can have a platform that has 32 bit memory addressing but can only do objects in memory that are 16 bits large, but since size_t is the type returned by sizeof() it's not a problem.

glebfm:
Ok, I agree, but it doesn't change that the header requirement is a regression.

Kuratius:
The only change that'd help with that while not violating the standard might be using int if the array is smaller than 16 bits and size_t for anything larger, but that means you need to generate the type dynamically.

I'm also not 100% sure that you can't run into the undefined behaviour situation I linked earlier if you do math to compute the size of one array using another.

glebfm:
In terms of generating a header file, adding an include makes a lot of sense, but still may break some users.

glebfm:
However, this feature was previously functional for C++ projects without any issues.  So, I still believe that use 'unsigned long' and 'unsigned long long' remains the best solution we have.

Kuratius:
The documentation specifies that xxd -i is meant to generate C-style code, not C++.

glebfm:
I agree that `size_t` is better, but you broke something that worked for almost 3 decades, it can't be a new feature to unbreak it.
Again, xxd should use a type appropriate for array size, `unsigned` if it fits in 16 bits, `unsigned long` if it fits in 32 bits and `unsigned long long` if it fits in 64 bits.

glebfm:
JFYI, gcc built to compile for AVR triggers a error for 1Mb array (error: size of array is too large) and also issues a warning for `size_t foo_len = 1048576;` (warning: unsigned conversion from ‘long int’ to ‘size_t’ {aka ‘unsigned int’} changes value from ‘1048576’ to ‘0’ [-Woverflow]).

Kuratius:
size_t is a type that is appropriate for every possible array size.

You can make a pull request for it if you think you can come up with a good solution that covers every use case without performance costs while supporting C++.

chrisbra:
alright, let me roll back that change. It's seems to cause too much issues. Perhaps in some future we can make `-i` use correct type depending on size then.

Kuratius:
@chrisbra possibly reopen the corresponding issue

PR #13881, opened 2024-01-17
Fix: current empty buffer won't be reused by `:drop` command

### LINKED ISSUE #13851: There is an empty buffer when `--remote` first starts Vim
**TLDR;** The root cause may be caused by the `:drop` command.

When we open Vim without a filename parameter, there is an unnamed empty `[No Name]` buffer. This is normal. When we pass the filename parameters, this empty buffer no longer exists. This is the normal working mode we expect.

But when `--remote` related parameters are included, there will always be a `[No Name]` buffer. To reproduce: Command line launch `gvim --remote-silent test.txt` when NO Vim exists; Or when starting Vim for the first time with a server name that does not currently exist like `gvim --servername NotExist --remote-silent test.txt`. Anyway, make sure you **start Vim** with the `--remote` parameter **for the first time**.

Know from the documentation:

>The `--remote` argument will cause a `:drop` command to be constructed from the rest of the command line and sent as described above.

So I started Vim directly, just entered `gvim` without any parameters, and then called the `:drop` command. The original `[No Name]` buffer will also not be reused.

--------

**Solutions I considered**

This may be solved directly in the code related to *clientserver*, but it may be better to directly modify the judgment logic of the `:drop` command. For the `:drop file` command, if the `file` has not been opened, at the same time, the current buffer is a `[No Name]` buffer and has not been modified yet, then reuse the current buffer just like `:edit file`. Maybe you have a better solution. Looking forward to your review and implementation, much appreciated.

--------

**Other Informations**

- OS: Microsoft Windows 10.0.19045
- Version of Vim: latest gVim 64-bit [v9.1.0026](https://github.com/vim/vim-win32-installer/releases/tag/v9.1.0026) from [vim-win32-installer](https://github.com/vim/vim-win32-installer)

roccomao:
Fix #13851

If the current buffer is an empty `[No Name]` buffer and has not been modified yet, the `:drop` command does not reuse it like the `:next` command does. This is caused by the default parameters of the `do_arglist` function when calling the `set_arglist` function in `ex_drop` as follows:

https://github.com/vim/vim/blob/211211052d0426394cbd5f42f3f3f78a64822e2a/src/ex_cmds.c#L5450

I can't modify it directly here, for example `do_arglist(eap->arg, AL_SET, 0, TRUE);`, because that need to use macro `AL_SET` defined in the *arglist.c* file. By globally searching for calls to the function `set_arglist`, only called here in `ex_drop`, so I modified the initialization parameters of the `do_arglist` function directly.

```c
/*
 * Redefine the argument list.
 */
    void
set_arglist(char_u *str)
{
    do_arglist(str, AL_SET, 0, TRUE);
}
```

Also, I made a slight change to the comment for the `:drop` command because the original comment was ambiguous. Like the `:next` command, the `:drop` command will reset the argument list, even if there is only one parameter, as described in the documentation [`:h :drop`](https://github.com/vim/vim/blob/211211052d0426394cbd5f42f3f3f78a64822e2a/runtime/doc/windows.txt#L801).

### DIFF
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -5425,8 +5425,7 @@ ex_smile(exarg_T *eap UNUSED)
 
 /*
  * ":drop"
- * Opens the first argument in a window.  When there are two or more arguments
- * the argument list is redefined.
+ * Opens the first argument in a window, and the argument list is redefined.
  */
     void
 ex_drop(exarg_T *eap)
@@ -5463,6 +5462,8 @@ ex_drop(exarg_T *eap)
 	// edited in a window yet.  It's like ":tab all" but without closing
 	// windows or tabs.
 	ex_all(eap);
+	cmdmod.cmod_tab = 0;
+	ex_rewind(eap);
 	return;
     }
 
@@ -5486,7 +5487,6 @@ ex_drop(exarg_T *eap)
 		buf_check_timestamp(curbuf, FALSE);
 		curbuf->b_p_ar = save_ar;
 	    }
-	    return;
 	}
     }
 
--- a/src/arglist.c
+++ b/src/arglist.c
@@ -502,7 +502,7 @@ do_arglist(
     void
 set_arglist(char_u *str)
 {
-    do_arglist(str, AL_SET, 0, FALSE);
+    do_arglist(str, AL_SET, 0, TRUE);
 }
 
 /*
--- a/src/testdir/test_excmd.vim
+++ b/src/testdir/test_excmd.vim
@@ -84,18 +84,27 @@ endfunc
 " Test for the :drop command
 func Test_drop_cmd()
   call writefile(['L1', 'L2'], 'Xdropfile', 'D')
+  " Test for reusing the current buffer
   enew | only
+  let expected_nr = bufnr()
   drop Xdropfile
+  call assert_equal(expected_nr, bufnr())
   call assert_equal('L2', getline(2))
   " Test for switching to an existing window
   below new
   drop Xdropfile
   call assert_equal(1, winnr())
-  " Test for splitting the current window
+  " Test for splitting the current window (set nohidden)
   enew | only
   set modified
   drop Xdropfile
   call assert_equal(2, winnr('$'))
+  " Not splitting the current window even if modified (set hidden)
+  set hidden
+  enew | only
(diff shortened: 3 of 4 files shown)


chrisbra:
Thank you. Can you please create a test for this?

roccomao:
Okay, but I'm sorry that I didn't conduct a complete test before. I just saw the CI error and am reproducing the unexpected error. I am new to C language projects and unit testing, please give me some time.

roccomao:
> Yes, when I `:drop` a file in the empty buffer, it seems to be successful, but the buffer does not display any content of the file. If I open a buffer at will, and then use any buffer command to jump back, e.g. `:buffer #`, the file content can be displayed correctly now.

In fact, referring to the implementation of the `:next` command, `do_argfile(eap, i);` must be called manually. The `:drop` command edits the first file, which is `do_argfile(eap, 0);`, which is equivalent to `ex_rewind(eap);`. It is called on the last line in `ex_drop` function, but when the current buffer can be reused, the `ex_drop` function will always return early, so solving it to ensure that the `:first` / `:rewind` command can be executed at the end.

In addition, some tests are added on the original basis. Many of the tests for `:tab drop` are written in the `Test_tabpage` function. In order to clarify the role, I placed the newly added tests in the `Test_tabpage_drop` function. Please review it, thank you.

roccomao:
**Review Request**

@chrisbra Sorry to bother you, I have implemented the requested changes and tests, is there anything else I need to do? Thanks for your response. 😄

src/ex_cmds.c
```diff
@@ -5486,7 +5487,6 @@ ex_drop(exarg_T *eap)
 		buf_check_timestamp(curbuf, FALSE);
 		curbuf->b_p_ar = save_ar;
 	    }
-	    return;
```
chrisbra, line 5489: why is this removed? Shoudln't we just add a `ex_rewind` before the return?

roccomao: Sorry, I was not familiar with the code at the beginning and was not careful enough. I previously confused the `FOR_ALL_TAB_WINDOWS` traversal with the buffer traversal.

roccomao:
Thank you so much for the review, I’ve learned a lot through this journey.

PR #13884, opened 2024-01-18
sh.vim: Add handling for ksh93 shared-state comsubs and mksh valsubs

### LINKED ISSUE #9514: ksh93 shared-state command substitution syntax isn't handled correctly
### Steps to reproduce

The following ksh93 script uses shared-state command substitutions to print a few lines. The syntax is valid but vim doesn't seem to handle the script correctly:
```ksh
#!/bin/ksh

print $(echo 'string one')
print ${ echo 'string two'; }
print ${ echo 'string three' ;}
print ${ echo 'string four' }
exit 0
```

### Expected behaviour

KornShell shared-state command substitutions should be rendered the same way normal command substitutions are.

### Version of Vim

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec 25 2021 11:17:40) Included patches: 1-3890  (using sh.vim from current git commit)

### Environment

Operating system: Arch Linux x86_64
Terminal: kitty 0.24.1
$TERM: xterm-kitty
Shell: ksh93u+m

### Logs and stack traces

_No response_

JohnoKing:
This commit adds support for ksh93's shared-state command substitutions (syntax: `${ command; }`) and mksh's value substitutions (syntax: `${|command;}`).

Below is a test script you can use for bug testing these changes:
```sh
#!/bin/ksh
# The following syntax applies to both ksh93 and mksh
print ${ echo one }
print ${	echo two
}
print ${
echo three	}
print ${ echo 'four'; }
print ${ echo 'five' ;}
print ${ echo 'six'
}
print ${	echo 'seven'	}
echo ${ print 'eight'	}
typeset nine=${ pwd; }

# Value substitutions of the form ${|command} are only
# supported by mksh, not ksh93.
if ! command eval '((.sh.version >= 20070703))' 2>/dev/null; then
	valsubfunc() {
		REPLY=$1
	}
	echo ${|valsubfunc ten}
	print "${|valsubfunc eleven;}"
	printf '%s' "${|valsubfunc twelve	}"
	unlucky=${|valsubfunc thirteen
}
	typeset notafloat=${|valsubfunc notanumber	}
	print $unlucky $notanumber
	${|echo foo}
	${|echo bar
}
fi

# Shared-state command substitutions using the syntax ${<file;}
# are only supported by ksh93, not mksh.
echo ${
	printf %s str
} > /tmp/strfile
echo ${</tmp/strfile;}

exit 0
```

Fixes https://github.com/vim/vim/issues/9514


chrisbra:
thanks. Can you make a syntax test out of this please? Check https://github.com/vim/vim/blob/master/runtime/syntax/testdir/README.txt and make a screendump test from your snippet please?

JohnoKing:
I have added the syntax test and screendumps to the pull request.

chrisbra:
thanks!

Issue #13886, opened 2024-01-18
add official filetype support for ChucK (.ck)


andreacfromtheapp:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**

Nothing impossible, just nice to have included.

**Describe the solution you'd like**

Adding something like this? Not sure how vim defines filetypes:

`au BufRead,BufNewFile *.ck setfiletype chuck`

**Describe alternatives you've considered**

At the moment I'm using `ftdetect/chuck.vim`

`au BufRead,BufNewFile *.ck setfiletype chuck`

**Additional context**

[ChucK](https://chuck.cs.princeton.edu) as a language, filetype, and extension has existed since 2002. It would be nice to have native support.

Also, I opened a feature request with NerdFonts for the logo to be included: https://github.com/ryanoasis/nerd-fonts/discussions/1479

chrisbra:
Thanks. I have created #13888 to detect the filetype. However there won't be any filetype, syntax or indent scripts. I'll leave that for later for someone who actually knows the language and can contribute those things

gacallea:
> Thanks. I have created #13888 to detect the filetype. However there won't be any filetype, syntax or indent scripts. I'll leave that for later for someone who actually knows the language and can contribute those things

thanks! I do have a syntax (that I didn't write) but it's from 8 years ago and I'm unsure it's complete. Should I share anyway?

gacallea:
Just in case, I'm sharing the syntax I have and the link to the API reference where all terms are present. Since the syntax file is 8y old, this API reference may help complete it: https://chuck.cs.princeton.edu/doc/reference/

```vim
if exists("b:current_syntax")
    finish
endif

syntax keyword chuckBoolean true false maybe
syntax keyword chuckBoolean null NULL
highlight link chuckBoolean Boolean

syntax match chuckComment "\v//.*$"
syntax region  chuckComment start="/\*"  end="\*/"
highlight link chuckComment Comment

syntax keyword chuckConditional if else
highlight link chuckConditional Conditional

syntax keyword chuckKeyword fun function
syntax keyword chuckKeyword return
syntax keyword chuckKeyword const
syntax keyword chuckKeyword new
syntax keyword chuckKeyword now
syntax keyword chuckKeyword pi
syntax keyword chuckKeyword me
syntax keyword chuckKeyword samp ms second minute hour day week
syntax keyword chuckKeyword repeat break continue
syntax keyword chuckKeyword class extends public private static pure this
[...]

PR #13895, opened 2024-01-21
runtime(chuck): adding ChucK syntax

andreacfromtheapp:
adding syntax for Chuck. https://github.com/vim/vim/issues/13886#issuecomment-1902682791

### DIFF
--- a/runtime/syntax/chuck.vim
+++ b/runtime/syntax/chuck.vim
@@ -1,0 +1,187 @@
+" Vim syntax file
+" Language: ChucK
+" Maintainer: Andrea Callea
+" URL: https://github.com/gacallea/chuck.vim
+" Last Change: 2024 Jan 21 by Andrea Callea
+
+" Sources used for this syntax
+" https://chuck.cs.princeton.edu/doc/language/
+" https://chuck.cs.princeton.edu/doc/reference/
+
+" HISTORY:
+" 2024 Jan 21 - Initial revision
+
+" quit when a syntax file was already loaded
+if exists("b:current_syntax")
+  finish
+endif
+
+" https://chuck.cs.princeton.edu/doc/language/type.html
+syn keyword chuckPrimitiveType int float time dur void vec3 vec4
+syn keyword chuckComplexType complex polar
+syn keyword chuckReferenceType Object Event UGen array string
+syn keyword chuckBoolean true false maybe
+syn keyword chuckBoolean null NULL
+
+" https://chuck.cs.princeton.edu/doc/language/oper.html
+syn match chuckOperator "=>"
+syn match chuckOperator "@=>"
+
+syn match chuckOperator "=^"
+
+syn match chuckOperator "+"
+syn match chuckOperator "-"
+syn match chuckOperator "*"
+syn match chuckOperator "/"
+syn match chuckOperator "%"
+syn match chuckOperator "!"
+syn match chuckOperator "&"
+syn match chuckOperator "|"
+syn match chuckOperator "\^"
+
+syn match chuckOperator "+=>"
+syn match chuckOperator "-=>"
+syn match chuckOperator "*=>"
+syn match chuckOperator "/=>"
+syn match chuckOperator "%=>"
+syn match chuckOperator "!=>"
+syn match chuckOperator "&=>"
+syn match chuckOperator "|=>"
+syn match chuckOperator "\^=>"
+
+syn match chuckOperator "&&"
+syn match chuckOperator "||"
+syn match chuckOperator "=="
+syn match chuckOperator "!="
+syn match chuckOperator ">"
+syn match chuckOperator ">="
+syn match chuckOperator "<"
+syn match chuckOperator "<="
+
+syn match chuckOperator ">>"
+syn match chuckOperator "<<"
+
+syn match chuckOperator "++"
+syn match chuckOperator "--"
+syn match chuckOperator "<<<"
+syn match chuckOperator ">>>"
+
+syn keyword chuckOperator new
+
+" https://chuck.cs.princeton.edu/doc/language/ctrl.html
+syn keyword chuckConditional if else
+syn keyword chuckRepeat while do until for each
+
+" https://chuck.cs.princeton.edu/doc/language/time.html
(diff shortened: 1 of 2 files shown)


runtime/syntax/chuck.vim
```diff
@@ -0,0 +1,154 @@
...
+syn match chuckHex /\<0[xX]\x\+[lL]\=\>/ display
+syn match chuckFloat /\%(\i\|\$\)\@<![-]\?\%(\d*\.\d\+\|\d\+\.\)/ display
+
+" this may need fixing/improvements
+syn match chuckComment "//.*$"
+syn region chuckComment start="/\*" end="\*/"
+syn match chuckSpecialChar contained "\\n"
+syn match chuckSpecialChar contained "\\t"
+syn match chuckSpecialChar contained "\\a"
+syn match chuckSpecialChar contained "\\""
```
dkearns: The quote needs to be escaped here but it's easier just to use a different delimiter.

```vim
syn match chuckSpecialChar contained /\\"/
```

runtime/syntax/chuck.vim
```diff
@@ -0,0 +1,154 @@
...
+hi def link chuckInputOutput Special
+hi def link chuckUtilities Special
+hi def link chuckNumber Number
+hi def link chuckHex Number
+hi def link chuckFloat Float
+hi def link chuckComment Comment
+hi def link chuckSpecialChar SpecialChar
+hi def link chuckString String
+
+let b:current_syn = "chuck"
```
dkearns: `b:current_syntax` - it's an identifier and can't be abbreviated.

dkearns (changes requested):
Other than these two final changes, it looks good to me.

Thanks for playing.

gacallea:
> Other than these two final changes, it looks good to me.
> 
> Thanks for playing.

thank you for your help and support :)

runtime/syntax/chuck.vim
```diff
@@ -0,0 +1,154 @@
...
+
+" https://chuck.cs.princeton.edu/doc/language/type.html
+syn keyword chuckPrimitiveType int float time dur void vec3 vec4
+syn keyword chuckComplexType complex polar
+syn keyword chuckReferenceType Object Event UGen array string
+syn keyword chuckBoolean true false maybe
+syn keyword chuckBoolean null NULL
+
+" https://chuck.cs.princeton.edu/doc/language/oper.html
+syn match chuckOperator "=>" "@=>" "=^"
```
dkearns: These `:syn-match` commands only take one pattern.  So you can use pattern alternation or split them up like,

```vim
syn match chuckOperator "=>"
syn match chuckOperator "@=>"
syn match chuckOperator "=^"
```

dkearns (changes requested):
Sorry, missed this...

runtime/syntax/chuck.vim
andreacfromtheapp: just those? or the whole category? (just checking)

dkearns: Yes, all of the chuckOperators.

andreacfromtheapp: I did them all to make sure :)

gacallea:
latest push to fix the date.

runtime/syntax/chuck.vim
```diff
@@ -0,0 +1,173 @@
...
+syn match chuckOperator "^"
+syn match chuckOperator "++"
+syn match chuckOperator "--"
+syn match chuckOperator "!"
+syn match chuckOperator "+"
+syn match chuckOperator "-"
+syn keyword chuckOperator new
+syn match chuckOperator "<<<"
+syn match chuckOperator ">>>"
+
```
dkearns, line 68: This reorders the chuckOperator section as a lot of these weren't being matched.    In short, later matches override earlier ones.

I also added some operators that appeared to be missing from my quick reading of the docs, like `+=>`.  I may be _very_ wrong as I'm not familiar with ChucK.

Whitespace separating lines can be removed, up to you.

There's still an issue with `@=>` being matched as keyword `@` followed by operator `=>`.  This can be fixed most simply by moving the operator definitions to somewhere after the keyword definition like just before the chuckNumber definitions.

```vim
" https://chuck.cs.princeton.edu/doc/language/oper.html
syn match chuckOperator "=>"
syn match chuckOperator "@=>"

syn match chuckOperator "=^"

syn match chuckOperator "+"
syn match chuckOperator "-"
syn match chuckOperator "*"
syn match chuckOperator "/"
syn match chuckOperator "%"
syn match chuckOperator "!"
syn match chuckOperator "&"
syn match chuckOperator "|"
syn match chuckOperator "\^"

syn match chuckOperator "+=>"
syn match chuckOperator "-=>"
syn match chuckOperator "*=>"
syn match chuckOperator "/=>"
syn match chuckOperator "%=>"
syn match chuckOperator "!=>"
syn match chuckOperator "&=>"
syn match chuckOperator "|=>"
syn match chuckOperator "\^=>"

syn match chuckOperator "&&"
syn match chuckOperator "||"
syn match chuckOperator "=="
syn match chuckOperator "!="
syn match chuckOperator ">"
syn match chuckOperator ">="
syn match chuckOperator "<"
syn match chuckOperator "<="

syn match chuckOperator ">>"
syn match chuckOperator "<<"

syn match chuckOperator "++"
syn match chuckOperator "--"
syn match chuckOperator "<<<"
syn match chuckOperator ">>>"

syn keyword chuckOperator new
```

dkearns (changes requested):
I just gave this a quick run and noticed some operators were not being matched.

runtime/syntax/chuck.vim
andreacfromtheapp: those make perfect sense. I wasn't aware of precedence order. I do was aware of the missing chuck operators. I wasn't sure about them tho, since they are compound (e.g: => and + together). nice spotting and amazing care on your side :)

ps: I'm only about to learn ChucK, thus I'm not an expert either. With time I hope to make this syntax better as well. For example I'm not too convinced about a couple of categorizations and their hightlights. Time will tell.

chrisbra:
thanks. Let me include this now

PR #13901, opened 2024-01-23
runtime(doc): change "VIsual mode" to "Visual mode" in :h SafeState

zeertzjq:
"Visual mode" is used everywhere else in the help when not referring to
something in the source code.


chrisbra:
thanks!

PR #13906, opened 2024-01-23

dkearns:
Highlight :2match and :3match and add these to :help ex-cmd-index.


chrisbra:
thanks!

PR #13909, opened 2024-01-24
Fix blockwise paste with multibyte characters

vanaigr:
Problem: incorrect number of trailing spaces is inserted for multibyte characters when pasting a blockwise register in blockwise visual mode.

When pasting in blockwise visual mode, and the register type is \<CTRL-V\>, vim will align the text after the replaced area by inserting spaces after pasted lines that are shorter than the longest line.
When a shorter line contains multibyte characters, each trailing UTF-8 byte's width is counted in addition to the width of the character itself. Each trailing byte counts as being 4 cells wide (since it would be displayed as `<xx>`).
image image

Solution:  skip over trailing UTF-8 bytes when computing the number of trailing spaces.

### DIFF
--- a/src/register.c
+++ b/src/register.c
@@ -1903,10 +1903,10 @@ do_put(
 		spaces = y_width + 1;
 		init_chartabsize_arg(&cts, curwin, 0, 0,
 						      y_array[i], y_array[i]);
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks, looks correct.

PR #13910, opened 2024-01-24
runtime(netrw): Don't change global options

k-takata:
Originally reported at: https://github.com/vim-jp/issues/issues/1428
'isk' was unintentionally changed by netrw.

`let &l:` should be used for non-global options.


k-takata:
netrw v172 caused the regression about 'isk': https://github.com/vim/vim/commit/71badf9547e8f89571b9a095183671cbb333d528

chrisbra:
thanks!

PR #13915, opened 2024-01-25
'linebreak' may still apply to leading whitespace

zeertzjq:
Problem:  'linebreak' may still apply to leading whitespace.
Solution: Compare pointers instead of virtual columns.

related: neovim/neovim#27180

Co-authored-by: VanaIgr <vanaigranov@gmail.com>


chrisbra:
thanks!

PR #13918, opened 2024-01-26
Fix document vim9.txt

h-east:
Note: I don't think it's a good idea to use backticks as markup.

### DIFF
--- a/runtime/doc/vim9.txt
+++ b/runtime/doc/vim9.txt
@@ -1265,7 +1265,7 @@ level.  They cannot be created in a function, also not in a legacy function.
 			yet.  This will report any errors found during
 			compilation.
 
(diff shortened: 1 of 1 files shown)


tonymec:
Backticks currently change the highlight: in gvim 9.1.59 I see the help text enclosed in backticks displayed as blue on white instead of the usual black on white. Invoking
```vim
:echo synIDattr(synID(line("."), col("."), 1), "name")
```
on that blue-on-white text answers `helpCommand`.

Best regards,
Tony.

h-east:
When the file type is Vim, `CTRL-]` on a word surrounded by backticks will jump to the help item for that word.
Therefore, it feels strange to use backtick for other purposes.

chrisbra:
thanks

PR #13919, opened 2024-01-26

dkearns:
Add new foreach() function.


zeertzjq:
Hmm, it cannot be generated?

RestorerZ:
> Hmm, it cannot be generated?

There was something on github.
[Here](https://github.com/vim-jp/syntax-vim-ex/tree/generator), if I'm not mistaken. The "generator" branch.

chrisbra:
Yes, unfortunately I believe Charles is mostly inactive now and I don't know how the vim.vim syntax file has been generated.

PR #13921, opened 2024-01-26
feat: add hurl commentstring

mawkler:
Here's the documentation for hurl comments: https://hurl.dev/docs/hurl-file.html#comments


chrisbra:
thanks!

PR #13923, opened 2024-01-27
Highlight string interpolation in Vim script

thinca:
Before:
image

After:
image

- Highlight `{ ... }` in interpolated string (`$'...'` and `$"..."`)
- Highlight `{{` and `}}` as escaped character (they are `{` and `}` in interpolated string)


chrisbra:
thanks. Can you please add a syntax test for it?

thinca:
@chrisbra I added a test.  This is my first time adding a syntax test.  Please point out any problems.

chrisbra:
thanks, I think this should be fine.

PR #13924, opened 2024-01-27
runtime(vim): Update syntax and ftplugin files - line continuations and comments

dkearns:
Improve matching of line-continuations and interspersed comments.

These are now matched in multiline syntax command patterns, dictionary
literals, and parenthesised expressions and argument lists.


chrisbra:
Thanks!

Issue #13916, opened 2024-01-25
Macros KS_EXTRA keycodes change file encoding


iddqd505:
### Steps to reproduce

I noticed when I record a simple macro with some combinations of modifier keys "Control" or "Alt" or make a motion like f) and then paste that macro register content in a buffer Vim prints some strange characters I think are called "KS_EXTRA keycodes"

1. I record a macro for this test line 
`TESTLINE_2001 (testing_test)`

2. Macro w (qw)
`0<C-A>f)vi)<C-C><M-j><M-j>`

3. (PASTE 1) This is the content of that macro pasted with "wp
`0f)ýavi)üCüjüj`

4. (PASTE 2) This is the same macro w pasted with execute command :echo keytrans (@w)
`0<C-A>f)<t_<fd>a>vi)<C-C><M-j><M-j>`

5. (PASTE 3) And this is the same macro w pasted in insert mode with Ctrl-R+w
`0f)<t_ý>avi)üCjj`

6. Aside getting 3 different pastes the real problem is when I leave written the macro register then save and close the file this happens:

    6.1. The file encoding changes from utf-8 to none

    6.2. Sometimes the buffer opens like read only file

    6.3. The pasted macro register changes from original form (PASTE 1) to (PASTE 3)

7. As a side note I have just noticed if I copy-paste (Ctrl+C and Ctrl+V in Windows) the macro register content (PASTE 1) the result also changes to (PASTE 3)

### Expected behaviour

Im not a programmer but I use gVim on Windows 10 as my primary text editor for a long time and the macros contents never pasted these extra keycodes until now

Also all my text files are in utf-8 encoding and unix fileformat since it helps with compatibility with Linux systems and gVim never had a problem to edit them that way until now

Im not an advanced Vim user and when I try to look for references or help about this problem I always get entangled in the rabbit hole that is file encoding configurations

I also have not done more extensive testing because like I say I actually have no idea what am I doing in this particular field

Please help

### Version of Vim

gVim version 9.0.2189

### Environment

Windows 10 Version 21H2 (OS Build 19044.1288)

Here is the output of the :version command:
https://github.com/iddqd505/dotfiles/blob/main/vim%20version

Here is my :scriptnames list of plugins. However I still can replicate the problem exactly the same even if I remove all my installed plugins and reload gVim
https://github.com/iddqd505/dotfiles/blob/main/vim%20scriptnames

And here is a sanitized and debloated version of my _vimrc file:
[...]

PR #13925, opened 2024-01-27
Recorded register cannot be translated using keytrans()

zeertzjq:
Problem:  Recorded register cannot be translated using keytrans() when
          it involves character search.
Solution: Record a K_IGNORE instead of a K_NOP.

related: #13916


chrisbra:
thank you !

PR #13926, opened 2024-01-27
runtime(ant): Update syntax file

dkearns:
Remove invalid display option from syn-keyword commands.

Take over maintenance of this file as Johannes hasn't been active for many
years.


chrisbra:
thanks!

PR #13931, opened 2024-01-29
Internal error when :luado deletes lines

zeertzjq:
Problem:  Internal error when :luado deletes lines.
Solution: Add another check for the line number.


chrisbra:
thanks!

PR #13934, opened 2024-01-30
fixed a logical error in help doc

chenkui164:
There should be a mistake in the original doc

### DIFF
--- a/runtime/doc/pattern.txt
+++ b/runtime/doc/pattern.txt
@@ -730,7 +730,7 @@ overview.
 	\(an\_s\+\)\@<=file	"file" after "an" and white space or an
 				end-of-line
 	For speed it's often much better to avoid this multi.  Try using "\zs"
-	instead |/\zs|.  To match the same as the above example:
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. Please change it like this however: `|/\@<=|`

runtime/doc/pattern.txt
```diff
@@ -730,7 +730,7 @@ overview.
 	\(an\_s\+\)\@<=file	"file" after "an" and white space or an
 				end-of-line
 	For speed it's often much better to avoid this multi.  Try using "\zs"
-	instead |/\zs|.  To match the same as the above example:
+	instead "\@<=".  To match the same as the above example:
```
k-takata, line 733: I think that `|/\zs|` is just a link. What the original sentence wants to say is:
Try using "\zs" instead. Refer to |/\zs|.

chenkui164: Ah, I see now. I misunderstood.

PR #13935, opened 2024-01-30
work around neovim bug for lispoptions<

benknoble:
This copies commit 64edf95 (indent: only reset some options when has vim9, 2024-01-30) from https://github.com/benknoble/vim-racket.


chrisbra:
thanks!

benknoble:
Ah, there's a slight bug here. I'll push a fix ASAP tomorrow.

PR #13936, opened 2024-01-30
runtime(vim): Update base-syntax, match multiline continued comments

dkearns:
Match multiline (continued) line comments.

Continued tail comments are not supported yet.


chrisbra:
Thanks!

PR #13938, opened 2024-01-30
ci.yml: get rid of snap

pheiduck:
seems to be a bottleneck for CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Uninstall snap
         run: |
           pushd /etc/apt/preferences.d/
-          cat > nosnap.pref <<EOF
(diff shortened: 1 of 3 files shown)


chrisbra:
How much faster is this?

pheiduck:
My intention was to prevent snap upgrades due it was hanging yesterday on a CI job.

pheiduck:
> How much faster is this?

 17m 3s compared to  [21m 54s](https://github.com/vim/vim/actions/runs/7718380848)

chrisbra:
oh cool. I guess technically, we don't need the `systemctl disable` command? and the `rm -rf` to clean up the snap directories?

pheiduck:
Okay CI Time is now: 16m 25s very nice.

PR #13941, opened 2024-01-30
fix degenerate splines in vimlogo.svg

ShayHill:
Fixed degenerate splines in vimlogo.svg from https://www.vim.org/logos.php. Also cleaned up the design just a touch. I tried to separate what were interesting idiosyncrasies of the original design and what were errors. This update will allow vinyl cutting, simple editing, large prints, and better conversion to raster formats.

FWIW, there are five file formats of the Vim logo in the repo and all five are different. This best matches the Vim logo on vim.org.


chrisbra:
thanks. 

> as of 2023 Jan 30,

I suppose this is a typo? I'll fix it up when I include it.

ShayHill:
> thanks.
> 
> > as of 2023 Jan 30,
> 
> I suppose this is a typo? I'll fix it up when I include it.

Not a typo. It looks like over time each file format for the Vim logo was put together by a different person either from scratch or from a different source. That would explain why they are all different. I'm happy to create a pr to make them all match if that is desired, or they could all be left as they are to give a little bit of choice.

Thank you.

PR #13942, opened 2024-01-30
Fixing remote file removal via ssh

MiguelBarro:
When trying to delete a file with a path like `scp://user@host//opt/program/file.ext` it fails.
After some tampering I found the actual ssh command was trying to delete `program/file.ext` instead of `/opt/program/file.ext`.
The issue is the regular expression used to extract the path. The original one (`^.*//[^/]\+/\(.*\)$`) is greedy and the '.*' expression matches `scp://user@host`.
In order to get it right we must use a non-greedy one (`.\{-}`) which will only match `scp:`.

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -10668,7 +10668,7 @@ fun! s:NetrwRemoteRmFile(path,rmfile,all)
       NetrwKeepj call netrw#ErrorMsg(s:ERROR,"for some reason b:netrw_curdir doesn't exist!",53)
       let ok="q"
      else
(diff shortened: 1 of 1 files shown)


chrisbra:
Hm, did you really use: `scp://user@host//opt/program/file.ext` , e.g. `//opt`? 
I think the patch makes sense, just wondering if this was a typo.

MiguelBarro:
Is not a typo. Is actually explained in `:help netrw-path`. The `\\` notation means *root-relative* and plain `\` means *user-directory* relative.
For windows the drive unit notation (`C:\`) makes the distinction. I believe the regexp pattern was cunningly crafted to manage this case too.

chrisbra:
oh, I wasn't aware of that distinction. Thanks for pointing out

PR #13945, opened 2024-01-31
fixup! runtime(racket): undo some indent options only when vim9script…

benknoble:
… is available (#13935)

This copies commit 2a4862a (fixup! indent: only reset some options when has vim9, 2024-01-31) from https://github.com/benknoble/vim-racket and fixes 26b0176a9 (runtime(racket): undo some indent options only when vim9script is available (#13935), 2024-01-30).


chrisbra:
thanks!

PR #13948, opened 2024-02-01
runtime(vim): Update syntax file - string escape sequences and special key notation

dkearns:
Improve string escape sequence and special key matching.


chrisbra:
thanks, looks good

PR #13952, opened 2024-02-01
runtime: Miranda files are not recognised

dkearns:
Problem:  Miranda files are not recognised
Solution: Add shebang and *.m extension filetype detection

Add new syntax and ftplugin files


zzzyxwvut:
The non-sequential `*_99.dump` files are no longer generated.
:)

PR #13957, opened 2024-02-02
fix compl_curr_match not correct when new leader add

glepnir:
Problem: when delete leader then add new leader the `compl_curr_match` value not sync to shown match value . it will cause the `completed_item` in `CompelteChanged` event not correct.

Solution: update the `compl_curr_match` to `compl_shown_match` when new leader add after delete leader when pum display.

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1363,6 +1363,10 @@ ins_compl_show_pum(void)
     pum_display(compl_match_array, compl_match_arraysize, cur);
     curwin->w_cursor.col = col;
 
+    // After adding leader, set the current match to shown match.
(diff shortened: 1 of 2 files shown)


src/insexpand.c
```diff
@@ -1363,6 +1363,10 @@ ins_compl_show_pum(void)
     pum_display(compl_match_array, compl_match_arraysize, cur);
     curwin->w_cursor.col = col;
 
+    // after add leader the curr value need sync to shown value.
```
zeertzjq:
```suggestion
    // After adding leader, set the current match to the shown match.
```

chrisbra:
thanks!

PR #13958, opened 2024-02-02
did_set_breakat() should be in optionstr.c

zeertzjq:
Problem:  did_set_breakat() should be in optionstr.c as 'breakat' is a
          string option.
Solution: Move did_set_breakat() to optionstr.c.


chrisbra:
thanks!

PR #13961, opened 2024-02-02
syntax highlighter: gpg.conf: Drop dangerous use-embedded-filename

dkg:
The syntax highlighter is likely to encourage people to use the listed commands.

but `use-embedded-filename` is a dangerous option that can cause GnuPG to write arbitrary data to arbitrary files whenever GnuPG encounters malicious data.

GnuPG upstream explicitly warns against using this option:

https://dev.gnupg.org/T4500

I recommend that vim not encourage users to place this option in their gpg.conf by highlighting it as a valid option.

I've also asked GnuPG upstream to explicitly deprecate the option due to its hazardous nature:

https://dev.gnupg.org/T6972

### DIFF
--- a/runtime/syntax/gpg.vim
+++ b/runtime/syntax/gpg.vim
@@ -145,7 +145,7 @@ syn keyword gpgOption	contained skipwhite nextgroup=gpgArgError
 			\ no-sk-comments no-skip-hidden-recipients
 			\ no-symkey-cache not-dash-escaped no-textmode
 			\ no-throw-keyids no-tty no-use-agent
-			\ no-use-embedded-filename no-utf8-strings no-verbose
+			\ no-utf8-strings no-verbose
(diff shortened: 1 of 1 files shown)


chrisbra:
But error is clearly wrong, as it is still supported. Not sure if we use Warning highlighting in syntax scripts.

Ajit-Thakkar:
Fortran.vim uses Todo highlighting for obsolescent features.

chrisbra:
I marked it with WarningMsg for now, since it looks like this option isn't going away from gpg according to the referenced tickets.

dkg:
> @dkg, this syntax file needs a maintainer if you're willing and able.

I'm not a vim user, let alone comfortable enough with vim to be a maintainer, but thanks for the offer.  I'm just trying to clean up dangerous corners of the ecosystem.

PR #13962, opened 2024-02-03
Allow modifying a final variable with a List value in-place using the += compound operator

### LINKED ISSUE #13745: [vim9script] Unexpected error using `+=` to modify a `final` list
### Steps to reproduce

The error
```
E46: Cannot change read-only variable "ll2"
```
is produced by the script
```
vim9script

# Note both ll2 and ll2x are the same list; the list can be modified.

final ll2: list<string> = []
var ll2x = ll2

ll2->add('b')
echo 'same:' ll2 is ll2x

# Note the following modifies the list, is does not modify either variable.
ll2x += ['cx']
echo 'same:' ll2 is ll2x

ll2 += ['d']
```

### Expected behaviour

No error. When `+=` is used on a `list`, the variable is not modifed. Only the referenced list is modifed.

### Version of Vim

9.0.2181

### Environment

ubuntu/gtk3/gnome

### Logs and stack traces

_No response_

### LINKED ISSUE #13959: odd err of 'final' when to concat list
### Steps to reproduce

```vim
vim9cmd final x = ['a'] | vim9cmd g:x += ['b']
# e1122 but opt actually applied   --------
echo g:x
# ['a', 'b']
```

### Expected behaviour

`['a']` or should no err?

### Version of Vim

v9.1.71

### Environment

linux

### Logs and stack traces

_No response_

yegappan:
Fixes #13745 and #13959.

As discussed in #13745, the "+=" operator modifies a List variable in-place.  So allow the List value of a final variable to be modified using "+=".

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -2093,7 +2093,7 @@ diff({fromlist}, {tolist} [, {options}])		*diff()*
 		Returns an empty List or String if {fromlist} and {tolist} are
 		identical.
 
-		Examples:
+		Examples: >
 		    :echo diff(['abc'], ['xxx'])
 		     @@ -1 +1 @@
 		     -abc
@@ -2103,7 +2103,7 @@ diff({fromlist}, {tolist} [, {options}])		*diff()*
 		     [{'from_idx': 0, 'from_count': 1, 'to_idx': 0, 'to_count': 1}]
 		    :echo diff(readfile('oldfile'), readfile('newfile'))
 		    :echo diff(getbufline(5, 1, '$'), getbufline(6, 1, '$'))
-
+<
 		For more examples, refer to |diff-func-examples|
 
 		Can also be used as a |method|: >
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -425,6 +425,18 @@ To change part of a list you can specify the first and last item to be
 modified.  The value must at least have the number of items in the range: >
 	:let list[3:5] = [3, 4, 5]
 
+To add items to a List in-place, you can use the |+=| operator: >
+	:let listA = [1, 2]
+	:let listA += [3, 4]
+<
+When two variables refer to the same List, changing one List in-place will
+cause the referenced List to be changed in-place: >
+	:let listA = [1, 2]
+	:let listB = listA
+	:let listB += [3, 4]
+	:echo listA
+	[1, 2, 3, 4]
+<
 Adding and removing items from a list is done with functions.  Here are a few
 examples: >
 	:call insert(list, 'a')		" prepend item 'a'
--- a/src/eval.c
+++ b/src/eval.c
@@ -1901,7 +1901,7 @@ set_var_lval(
 			   && !tv_check_lock(&di->di_tv, lp->ll_name, FALSE)))
 			&& tv_op(&tv, rettv, op) == OK)
 		    set_var_const(lp->ll_name, lp->ll_sid, NULL, &tv, FALSE,
-							    ASSIGN_NO_DECL, 0);
+				ASSIGN_NO_DECL | ASSIGN_COMPOUND_OP, 0);
 		clear_tv(&tv);
 	    }
 	}
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -3977,8 +3977,14 @@ set_var_const(
 	    if (check_typval_is_value(&di->di_tv) == FAIL)
 		goto failed;
 
+	    // List and Blob types can be modified in-place using the "+="
+	    // compound operator.  For other types, this is not allowed.
+	    int type_inplace_modifiable =
+		(di->di_tv.v_type == VAR_LIST || di->di_tv.v_type == VAR_BLOB);
+
 	    if (var_in_vim9script && (flags & ASSIGN_FOR_LOOP) == 0
-		    && (flags & ASSIGN_COMPOUND_OP) == 0)
+		    && ((flags & ASSIGN_COMPOUND_OP) == 0
+			|| !type_inplace_modifiable))
 	    {
 		where_T where = WHERE_INIT;
 		svar_T  *sv = find_typval_in_script(&di->di_tv, sid, TRUE);
@@ -3999,8 +4005,11 @@ set_var_const(
 		}
 	    }
 
+	    // Modifying a final variable with a List value using the "+="
+	    // operator is allowed.  For other types, it is not allowed.
 	    if (((flags & ASSIGN_FOR_LOOP) == 0
-			&& (flags & ASSIGN_COMPOUND_OP) == 0)
+			&& ((flags & ASSIGN_COMPOUND_OP) == 0
+			    || !type_inplace_modifiable))
 				 ? var_check_permission(di, name) == FAIL
 				 : var_check_ro(di->di_flags, name, FALSE))
 		goto failed;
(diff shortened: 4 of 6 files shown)


errael:
The PR is not restricted to lists. The following should generate an error, but it does not
```
vim9script
final i: number = 1
i += 2
echo i
i *= 3
echo i
```

chrisbra:
Thanks. Shall we add the following doc-patch on top of it:
```patch
diff --git a/runtime/doc/eval.txt b/runtime/doc/eval.txt
index 9c12fb850..f786a0a45 100644
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -307,6 +307,13 @@ Two lists can be concatenated with the "+" operator: >

 To prepend or append an item, turn the item into a list by putting [] around
 it.  To change a list in-place, refer to |list-modification| below.
+Note:  Because of |list-identity| below, this may cause referenced lists to be
+changed in-place: >
+       let l1 = [1, 2, 3]
+       let l2 = l1
+       let l2 += [4, 5]
+       echo l1
+       [1, 2, 3, 4, 5]


 Sublist ~
```

yegappan:
> The PR is not restricted to lists. The following should generate an error, but it does not
> 
> ```
> vim9script
> final i: number = 1
> i += 2
> echo i
> i *= 3
> echo i
> ```

Yes.  This allows modification of other types also.  Should we allow this only for Lists
and block for other types?  or should we fail this for all the types?

errael:
I was also playing with the doc when the latest change came in. Here it is for reference. 

In the 2nd hunk below, adding the word `usually` and the additional example may be useful.

```diff
diff --git a/runtime/doc/eval.txt b/runtime/doc/eval.txt
--- a/runtime/doc/eval.txt
+++ b/runtime/doc/eval.txt
@@ -307,6 +307,13 @@
 
 To prepend or append an item, turn the item into a list by putting [] around
 it.  To change a list in-place, refer to |list-modification| below.
+Note:  List |+=| is an in-place modification, because of |list-identity|
+below, this may cause referenced lists to be changed in-place: >
+	let l1 = [1, 2, 3]
+	let l2 = l1
+	let l2 += [4, 5]
+	echo l1
+	[1, 2, 3, 4, 5]
 
 
 Sublist ~
@@ -425,13 +432,14 @@
 modified.  The value must at least have the number of items in the range: >
 	:let list[3:5] = [3, 4, 5]
 
-Adding and removing items from a list is done with functions.  Here are a few
-examples: >
+Adding and removing items from a list is usually done with functions.  Here
+are a few examples: >
 	:call insert(list, 'a')		" prepend item 'a'
 	:call insert(list, 'a', 3)	" insert item 'a' before list[3]
 	:call add(list, "new")		" append String item
 	:call add(list, [1, 2])		" append a List as one new item
 	:call extend(list, [1, 2])	" extend the list with two more items
+	:let list += [1, 2]		" extend the list with two more items
 	:let i = remove(list, 3)	" remove item 3
 	:unlet list[3]			" idem
 	:let l = remove(list, 3, -1)	" remove items 3 to last item
```

yegappan:
> > The PR is not restricted to lists. The following should generate an error, but it does not
> > ```
> > vim9script
> > final i: number = 1
> > i += 2
> > echo i
> > i *= 3
> > echo i
> > ```
> 
> Yes. This allows modification of other types also. Should we allow this only for Lists and block for other types? or should we fail this for all the types?

I have updated the PR to allow in place modification of List and Blob type final variables.

errael:
> This allows modification of other types also. Should we allow this only for Lists
> and block for other types?

That's the only way it could work.

For primitive types, doing this is a violation of the `:final` contract to modify the variable.

For lists it is OK, since the variable is not modified (only the value is modified).

> or should we fail this for all the types?

Answering that is above my pay grade. Allowing it for lists and failing it for primitives is expected (IMHO), others may disagree. @dkearns You're a resident language lawyer.

qeatzy:
I think we should **not** allow it. Below is my opinion at [issues/13745](https://github.com/vim/vim/issues/13745#issuecomment-1925685444).


> 7. lst = lst is disallowed if lst is declared final.
> 8. += is assignment too. The final constraint applies too. Though an exception could be made in the language spec level.
> 
> At last, as for point 8 (and the error these issues encounter)
> I'm against to make a hole to += and final, which create a quirk in the language for no benefit. Where the alternative is well defined. (just use add()/extend() in this case. Perhaps mention at doc describing final.
> 
> 
>

qeatzy:
I created a pull request for more accurate description of `+=` for list-concatenation documentation at [here](https://github.com/qeatzy/vim/pull/1).
Since I don't how the workflow of vim pull request yet, the above pull request may not work at all. But the idea is clearly describe and the change is minimal and there is no conflict.

chrisbra:
Thanks. Let me include this. @qeatzy can you please create a follow-up PR against this repo here?

qeatzy:
> Thanks. Let me include this. @qeatzy can you please create a follow-up PR against this repo here?

I'm ok to learn the workflow of vim pull request workflow and create PR here.

But as I said previously, I hope
1. documentation being concise & clear.
2. Not allow `+=` for `final` and `list` as exception, since `+=` is disable final variable of other type.

Since my previous comment, it looks like this pull request being accepted.
So I checked [commit 1af3563](https://github.com/vim/vim/commit/1af35631f85d2fcdc83c5d457af8273697f5146a)
If the that's settled down, I'm ok with point 2, there's no big deal.

As for point 1 and my previous PR attempt before *commit 1af3563*
My PR attempt prefer conciseness. Adding only 2 lines to doc `eval.txt`.
Then I checked [commit 1af3563](https://github.com/vim/vim/commit/1af35631f85d2fcdc83c5d457af8273697f5146a)
Which add a plethora to doc `eval.txt`.

As for modification to doc, these two approach is fundamentally different.
Should I
1. Just forget it and accept commit 1af3563 as final.
2. Create a PR and discuss there (which will revert change to eval.txt in commit 1af3563).
3. Discuss further here.
Which one should I choose? @chrisbra 

P.S., as for reference.
My appoach, [2 line addition to eval.txt](https://github.com/qeatzy/vim/pull/1/commits/f911dc007011e37eab02859b811f4f9e129c0616)
commit 1af3563, [More than 10 lines to eval.txt, but IMO still not clear.](https://github.com/vim/vim/commit/1af35631f85d2fcdc83c5d457af8273697f5146a#diff-b3aeef96275167ddcc1e5fee6d51b8f2018f9359f921d86ad6e7035ca80f20e9)

errael:
> Dart will fail when applying `+=` to anything declared `final`. It only allows the `+` operator to be defined and the compound operator expands to `l = l + r`. However, others like Scala and Kotlin allow the compound operator to be defined explicitly, (falling back to `+` if it's not) and these are generally [implemented](https://github.com/JetBrains/kotlin/blob/0938b46726b9c6938df309098316ce741815bb55/libraries/stdlib/src/kotlin/collections/MutableCollections.kt#L49) to allow mutations.

Thanks for that link. I couldn't find explicit specification for `+=` for the builtin collections, but when I saw that you could define `plusAssign`...

So, it looks like there are two typical approaches
1. Rewrite `l += r` as `l = l + r`; `l` can not be final
2. `+=` is in place/extends; `l` can be final (for default containers; behavior can be overridden).

Before the new/merged behavior `vimscript` was doing neither. Requiring backward compatibility meant (1) could not be done.

PR #13963, opened 2024-02-03
Looping over modifier_keys_table[] unnecessarily

zeertzjq:
Problem:  Looping over modifier_keys_table[] unnecessarily with only
          MOD_MASK_ALT or MOD_MASK_CMD, as modifier_keys_table[] only
          contains MOD_MASK_SHIFT and MOD_MASK_CTRL, and the loop won't
          do anything.
Solution: Remove MOD_MASK_ALT and MOD_MASK_CMD from the condition.

### DIFF
--- a/src/misc2.c
+++ b/src/misc2.c
@@ -1130,11 +1130,7 @@ simplify_key(int key, int *modifiers)
     int	    key0;
     int	    key1;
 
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

Issue #13666, opened 2023-12-11
syntax/dosbatch.vim should not flag colon comments as errors by default


mike-clark-8192:
Repro

1) Do a clean install of latest nightly build of vim (tested with 9.0.2103)
2) Use above vim to open `$VIMRUNTIME/vimtutor.bat` (on Windows), or open the file from the vim git repo (*nix).
3) Notice the `::` comments are flagged as errors (must have :syntax on, etc.)

Blame:
https://github.com/vim/vim/blame/a40e1687e757ea4aa41e84006c739f575db75703/runtime/syntax/dosbatch.vim#L95

My 2 cents:
`syntax/dosbatch.vim` default behavior should be to NOT show `::` comments as errors. This type of comment is [very common](https://github.com/search?q=%2F%5E%3A%3A%2F%20path%3A*.bat&type=code).

~~`syntax/dosbatch.vim` should consider using `g:` variables for its configurable behavior.~~
~~`exists("dosbatch_cmdextversion")` ==> `exists("g:dosbatch_cmdextversion")`~~
~~`exists("dosbatch_colons_comment")` ==> `exists("g:dosbatch_colons_comment")`~~

chrisbra:
1) This was discussed in this PR: https://github.com/vim/vim/pull/11980 

2) using non-scoped variable names is fine, since they will be `g:`-scoped by default.

mike-clark-8192:
Thanks. I am sorry that I wasn't able to find that issue. I did try to search.  :)

That issue is closed.  Should I try to discuss there, or here? I don't know the etiquette.

Even Microsoft themselves use colon comments in their most modern software:
VSCode's CLI launcher: https://github.com/microsoft/vscode/blob/main/scripts/code.bat

Vim itself uses colon comments in bat files.

They're just everywhere. I don't know any IDE or syntax-highlighting text editor that flags them as an error.  Yes, they can break scripts [when used inside a FOR loops](https://www.robvanderwoude.com/comments.php), but these problems are obvious if you test the batch file. 

I think that's about all I can say on this.

`:: text here` is not a syntax error in dosbatch, and I think it's an overreach to flag it as such. If you would like to colorize it in something that makes people raise their eyebrows, great, but I really don't fancy reading .bat files from bug-free shipping software with red squiggles everywhere. Yes, I know I can customize vim, but the out-of-the-box experience is important to discuss too, right?

chrisbra:
let me ping the maintainer here: @mrdubya

[...]

PR #13965, opened 2024-02-03
runtime(dosbatch): improve :: comment highlighting

mrdubya:
Added a syntax region for command blocks so that the highlighting of `::` comments in them can be controlled.  The `dosbatch_colons_comment` variable now controls if all `::` comments in a code block are highlighted as comments or errors.  A `::` comment at the end of a command block is always highlighted as an error.

This re-enables the highlighting of `::` comments in `.bat` files as requested in #13666, while allowing control of highlighting them in command blocks requested in #11980.

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -1331,11 +1331,14 @@ Stack Overflow -
 
 https://stackoverflow.com/questions/12407800/which-comment-style-should-i-use-in-batch-files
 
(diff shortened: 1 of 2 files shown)


chrisbra:
@mike-clark-8192 is it okay?

chrisbra:
thanks all. I include it.

chrisbra:
included as of f7f33e371

PR #13966, opened 2024-02-03
Fix indent after line with literal dict

AndrewRadev:
There seems to be an issue with indentation when the previous line contains a literal dict with `#{ }` delimiters:

```vim
call prop_type_add('indent_after_literal_dict', #{ foo: 'bar' })
    call prop_type_delete('indent_after_literal_dict')
```

The attached patch fixes the issue. The "line continuation" regex ignores a tailing `#.*`, but if the next character after the `#` is a `{`, that shouldn't be considered a comment. This actually means that `#` alone won't be matched there, but oddly enough, indentation seems to work fine in that case for me.


chrisbra:
Thanks, let me include it now

PR #13969, opened 2024-02-03

dkearns:
Improve string interpolation highlighting.

Use the vimSep group to highlight interpolation braces as vimOperParen
has no highlighting of its own and employs vimSep via matchgroup.

Add vimNumber to the interpolation group's contained list.

@thinca, let me know if this doesn't make sense to you.


thinca:
I didn't know much about these.  Thank you for your improvement.

chrisbra:
thanks all, I include it then

PR #13971, opened 2024-02-03
Make references to [range] consistent

JJCUBER:
Some parts of the vimdocs refer to ranges with `{range}` instead of `[range]`.  Since there are more instances of `[range]` and only information on the `[range]` variant exists (`:h [range]` and `:h :[range]` exist, `:h {range}` does not), I made **all** of the references consistent.

---

The EX commands index is the only instance where it is left as `{range}` but consequently clarified as normally being denoted with `[range]`, since that section of the helpdocs uses `[...]` to mean optional.  (`[range]` is not optional for `:[range]`, since it would otherwise be `:`, which is specified as a separate command.)

I also noticed that this line used the wrong tag for the command (it linked to information about ranges, `[range]`, instead of information about the command `:[range]`).

Below is context for what I am referring to (with my proposed changes included):
```
6. EX commands					*ex-cmd-index* *:index*

This is a brief but complete listing of all the ":" commands, without
mentioning any arguments.  The optional part of the command name is inside [].
The commands are sorted on the non-optional part of their name.

tag		command		action ~
------------------------------------------------------------------------------
```
```diff
|:|		:		nothing
-|:range|	:{range}	go to last line in {range}
+|:[range]|	:{range}	go to last line in {range} (normally denoted |[range]|)

```

### DIFF
--- a/runtime/doc/fold.txt
+++ b/runtime/doc/fold.txt
@@ -297,8 +297,8 @@ zf{motion}  or
 							*zF*
 zF		Create a fold for [count] lines.  Works like "zf".
 
-:{range}fo[ld]						*:fold* *:fo*
-		Create a fold for the lines in {range}.  Works like "zf".
+:[range]fo[ld]						*:fold* *:fo*
+		Create a fold for the lines in [range].  Works like "zf".
 
 							*zd* *E351*
 zd		Delete one fold at the cursor.  When the cursor is on a folded
@@ -399,15 +399,15 @@ zr		Reduce folding: Add |v:count1| to 'foldlevel'.
 zR		Open all folds.  This sets 'foldlevel' to highest fold level.
 
 							*:foldo* *:foldopen*
-:{range}foldo[pen][!]
-		Open folds in {range}.  When [!] is added all folds are
-		opened.  Useful to see all the text in {range}.  Without [!]
+:[range]foldo[pen][!]
+		Open folds in [range].  When [!] is added all folds are
+		opened.  Useful to see all the text in [range].  Without [!]
 		one level of folds is opened.
 
 							*:foldc* *:foldclose*
-:{range}foldc[lose][!]
-		Close folds in {range}.  When [!] is added all folds are
-		closed.  Useful to hide all the text in {range}.  Without [!]
+:[range]foldc[lose][!]
+		Close folds in [range].  When [!] is added all folds are
+		closed.  Useful to hide all the text in [range].  Without [!]
 		one level of folds is closed.
 
 							*zn*
--- a/runtime/doc/index.txt
+++ b/runtime/doc/index.txt
@@ -1161,7 +1161,7 @@ The commands are sorted on the non-optional part of their name.
 tag		command		action ~
 ------------------------------------------------------------------------------
 |:|		:		nothing
-|:range|	:{range}	go to last line in {range} (normally denoted |[range]|)
+|:[range]|	:{range}	go to last line in {range} (normally denoted |[range]|)
 |:!|		:!		filter lines or execute an external command
 |:!!|		:!!		repeat last ":!" command
 |:#|		:#		same as ":number"
--- a/runtime/doc/index.txt
+++ b/runtime/doc/index.txt
@@ -1161,7 +1161,7 @@ The commands are sorted on the non-optional part of their name.
 tag		command		action ~
 ------------------------------------------------------------------------------
 |:|		:		nothing
-|:range|	:{range}	go to last line in {range}
+|:range|	:{range}	go to last line in {range} (normally denoted |[range]|)
 |:!|		:!		filter lines or execute an external command
 |:!!|		:!!		repeat last ":!" command
 |:#|		:#		same as ":number"
--- a/runtime/doc/insert.txt
+++ b/runtime/doc/insert.txt
@@ -1972,8 +1972,8 @@ too long when appending characters a line break is automatically inserted.
 9. Ex insert commands					*inserting-ex*
 
 							*:a* *:append*
-:{range}a[ppend][!]	Insert several lines of text below the specified
-			line.  If the {range} is missing, the text will be
+:[range]a[ppend][!]	Insert several lines of text below the specified
+			line.  If the [range] is missing, the text will be
 			inserted after the current line.
 			Adding [!] toggles 'autoindent' for the time this
 			command is executed.
@@ -1982,8 +1982,8 @@ too long when appending characters a line break is automatically inserted.
 			name.
 
 							*:i* *:in* *:insert*
-:{range}i[nsert][!]	Insert several lines of text above the specified
-			line.  If the {range} is missing, the text will be
+:[range]i[nsert][!]	Insert several lines of text above the specified
+			line.  If the [range] is missing, the text will be
 			inserted before the current line.
 			Adding [!] toggles 'autoindent' for the time this
 			command is executed.
@@ -2043,7 +2043,7 @@ NOTE: These commands cannot be used with |:global| or |:vglobal|.
 			the cursor.
 			See |++opt| for the possible values of [++opt].
 
-:{range}r[ead] [++opt] [name]
+:[range]r[ead] [++opt] [name]
 			Insert the file [name] (default: current file) below
 			the specified line.
 			See |++opt| for the possible values of [++opt].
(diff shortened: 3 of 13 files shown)


zeertzjq:
I don't think this makes sense. `{range}` means it's mandatory and `[range]` means it's optional.

JJCUBER:
> I don't think this makes sense. `{range}` means it's mandatory and `[range]` means it's optional.

The issue, in my opinion, is that
1. `[ranges]` are inherently optional (they default to a range over the current line)
2. Information on ranges themselves are explicitly put under `[ranges]` in the helpdocs, as opposed to `{ranges}` or `ranges`

Additionally, the very helpdocs for `[range]` say this:

> Some Ex commands accept a line range in front of them.  This is noted as
> [range].  It consists of one or more line specifiers, separated with ',' or
> ';'.
This makes it sound like it should always be denoted as `[range]`.

---

I feel like this unification I proposed makes the most sense overall, though other methods of unification might work.

zeertzjq:
IMO, if omitting the range leads to completely different behavior, the range isn't really optional.  `:normal` with and without range is an example.

JJCUBER:
It seems to me that the behavior of both with and without the `[range]` is identical for `:normal`.  The only difference documentation-wise is an extra clarification for how it behaves with ranges; this could easily be combined with the description of `:normal` (instead, just having `:[range]normal`).  The extra clarification would still be accurate for when the range is over a single line.

This would be consistent with how some commands work which only have a single helpdocs entry for the range variant, while still having the "optional" notation (i.e. `:[range]g` `:[range]>` which both don't have a non-range variant, yet use the "optional" notation).

Of course, I did not help create the helpdocs over the years, so maybe this is just some artifact of changes in notation and I am misunderstanding things; however, none of the commands with separate entries say anything about the (explicit) range which would be inconsistent for a single line (implicit range).

runtime/doc/fold.txt
```diff
@@ -297,8 +297,8 @@ zf{motion}  or
 							*zF*
 zF		Create a fold for [count] lines.  Works like "zf".
 
-:{range}fo[ld]						*:fold* *:fo*
-		Create a fold for the lines in {range}.  Works like "zf".
+:[range]fo[ld]						*:fold* *:fo*
```
chrisbra, line 300: I think here `[range]` is not really optional. Please leave it as before

runtime/doc/fold.txt
```diff
@@ -399,15 +399,15 @@ zr		Reduce folding: Add |v:count1| to 'foldlevel'.
 zR		Open all folds.  This sets 'foldlevel' to highest fold level.
 
 							*:foldo* *:foldopen*
-:{range}foldo[pen][!]
-		Open folds in {range}.  When [!] is added all folds are
-		opened.  Useful to see all the text in {range}.  Without [!]
+:[range]foldo[pen][!]
```
chrisbra, line 402: same here

runtime/doc/fold.txt
```diff
@@ -399,15 +399,15 @@ zr		Reduce folding: Add |v:count1| to 'foldlevel'.
...
+		Open folds in [range].  When [!] is added all folds are
+		opened.  Useful to see all the text in [range].  Without [!]
 		one level of folds is opened.
 
 							*:foldc* *:foldclose*
-:{range}foldc[lose][!]
-		Close folds in {range}.  When [!] is added all folds are
-		closed.  Useful to hide all the text in {range}.  Without [!]
+:[range]foldc[lose][!]
```
chrisbra, line 408: same here

runtime/doc/index.txt
```diff
@@ -1161,7 +1161,7 @@ The commands are sorted on the non-optional part of their name.
 tag		command		action ~
 ------------------------------------------------------------------------------
 |:|		:		nothing
-|:range|	:{range}	go to last line in {range}
+|:[range]|	:{range}	go to last line in {range} (normally denoted |[range]|)
```
chrisbra, line 1164: same here, [range] is not really optionally

runtime/doc/insert.txt
```diff
@@ -2043,7 +2043,7 @@ NOTE: These commands cannot be used with |:global| or |:vglobal|.
 			the cursor.
 			See |++opt| for the possible values of [++opt].
 
-:{range}r[ead] [++opt] [name]
+:[range]r[ead] [++opt] [name]
 			Insert the file [name] (default: current file) below
 			the specified line.
```
chrisbra, line 2048: Below the specified {range} ?

runtime/doc/quickref.txt
```diff
@@ -1455,7 +1455,7 @@ Context-sensitive completion on the command-line:
 		set foldmethod=marker	folding by 'foldmarker'
 
 |zf|		zf{motion}		operator: Define a fold manually
-|:fold|		:{range}fold		define a fold for {range} lines
+|:fold|		:[range]fold		define a fold for [range] lines
```
chrisbra, line 1458: not really optionally I think.

runtime/doc/userfunc.txt
```diff
@@ -116,7 +116,7 @@ See |:verbose-cmd| for more information.
 			When the [range] argument is added, the function is
 			expected to take care of a range itself.  The range is
 			passed as "a:firstline" and "a:lastline".  If [range]
-			is excluded, ":{range}call" will call the function for
+			is excluded, ":[range]call" will call the function for
```
chrisbra, line 119: not really optional, since the manual describes how `:call` is handled when a `{range}` is given.

runtime/doc/usr_25.txt
```diff
@@ -134,9 +134,9 @@ if your text can't be properly formatted with Vim's builtin command.  See the
 
 To center a range of lines, use the following command: >
 
-	:{range}center [width]
+	:[range]center [width]
```
chrisbra, line 137: not really optional from the context above.

runtime/doc/various.txt
```diff
@@ -185,7 +185,7 @@ g8			Print the hex values of the bytes used in the
 := [flags]		Print the last line number.
 			See |ex-flags| for [flags].
 
-:{range}= [flags]	Prints the last line number in {range}.  For example,
+:[range]= [flags]	Prints the last line number in [range].  For example,
```
chrisbra, line 188: this is specifically how the `=` command handles the `[range]`. So not really optional from the context (else it would be the same as line 185 above).

runtime/doc/various.txt
```diff
@@ -230,9 +230,9 @@ g8			Print the hex values of the bytes used in the
 			Example: >
 				:exe "normal \<c-w>\<c-w>"
 
-:{range}norm[al][!] {commands}				*:normal-range*
+:[range]norm[al][!] {commands}				*:normal-range*
 			Execute Normal mode commands {commands} for each line
-			in the {range}.  Before executing the {commands}, the
+			in the [range].  Before executing the {commands}, the
```
chrisbra, line 235: not really optional as well.

src/ex_eval.c
```diff
@@ -76,7 +76,7 @@ static int cause_abort = FALSE;
 
 /*
  * Return TRUE when immediately aborting on error, or when an interrupt
- * occurred or an exception was thrown but not caught.  Use for ":{range}call"
+ * occurred or an exception was thrown but not caught.  Use for ":[range]call"
```
chrisbra, line 79: not really optional from context.

src/match.c
```diff
@@ -437,7 +437,7 @@ next_search_hl(
     int		called_emsg_before = called_emsg;
     int		timed_out = FALSE;
 
-    // for :{range}s/pat only highlight inside the range
+    // for :[range]s/pat only highlight inside the range
```
chrisbra, line 440: not really optional as well from context.

chrisbra:
not really sure it is worth it. Mostly it seems the `{range}` is used to denote when a range is not really optional. There are a few cases however, where `[range]` is okay I think.

JJCUBER:
Would it make sense, as an alternative, to have "{range}" also jump to the same helpdocs location as "[range]"?  Currently, it doesn't have any entry.

chrisbra:
yes, makes sense.

runtime/doc/index.txt
JJCUBER: The issue here is `:range` links to something other than what they were expecting.  However, once making {range} point to the same place as [range], this will be easy to resolve.

chrisbra:
sure thanks.

PR #13973, opened 2024-02-04
Redrawing diff filler lines unnecessarily with 'relativenumber'

zeertzjq:
Problem:  Redrawing diff filler lines unnecessarily with 'relativenumber'.
Solution: Stop at WL_NR drawing state.


zeertzjq:
I just noticed that `LineNrAbove` and `LineNrBelow` background behave incorrectly with wrapped lines. I'll open another PR with different commit message.

PR #13974, opened 2024-02-04
LineNrAbove and LineNrBelow background wrong on wrapped lines

zeertzjq:
Problem:  LineNrAbove and LineNrBelow background wrong on wrapped lines.
Solution: Update number column also for wrapped part of a line.


chrisbra:
thanks. Can you show before/after screenshots?

luukvbaal:
In Neovim we have renamed "number_only" to "col_rows" in a similar patch. Might I suggest doing the same here instead of obfuscating an integer value in a boolean variable name?

zeertzjq:
I don't really like the `col_rows` variable name and I don't think `number_only` sounds boolean-only, but I'll add the `== 0` / `> 0` checks.

luukvbaal:
Yeah I suppose "col_rows" also obfuscates the fact that only the columns are being redrawn l, which is arguably more important. TBH my main reason for replacing "number_only" was the fact that not only the number, but also the fold and sign columns are redrawn.

zeertzjq:
Before:
Before

After:
After

PR #13975, opened 2024-02-04
ci.yml: skip upgrade

pheiduck:
Revert: https://github.com/vim/vim/pull/13680
Fixed in: https://github.com/actions/runner-images/issues/9016


chrisbra:
thanks!

PR #13977, opened 2024-02-05
[BREAKING] terminal background detection defaults to dark instead of light

habamax:
- Green and blue color parts are mixed in `handle_osc`, probably a typo?
- Make `dark` background default with `COLORFGBG` to be checked if it is `light`

Tested with setting terminal `bg=#d0d0d0` and `fg=#000000`, light background is correctly identified and set for:
- urxvt
- xterm


habamax:
@chrisbra I will fix the tests if this would be considered for inclusion (apparently screendumps are different now due to 'background' change)

habamax:
@neutaaaaan could you check it with `urxvt`? It uses `COLORFGBG` to report whether background is light or dark

habamax:
Hi @romainl, would you be able to try it out on OSX?

habamax:
Have tried `xterm` and `urxvt` on debian -- light background is correctly detected.

habamax:
Removed my "fixed" tests -- looks like vim tests expect "light" background and even set it in `RunVimInTerminal` function, but for some reason I still have proper dark background in my screendumps.

habamax:
There are 3 tests failing I have no idea why:

## `Test_conceal_cuc_01`
image

## `Test_conceal_cuc_02`
image

## `Test_functions_echoraw`
image

habamax:
Reset to before test updates.

I need to park it for a bit -- tests are failing even though they should be independent of `background`.

neutaaaaan:
This is a bad idea, you're substituting broken behaviour for broken behaviour that will affect fewer users.

There are only 2 ways to fix this : either implement a detection system that can actually be relied upon over ssh and inside multiplexers, or fix the various terminal versions of the default light colorscheme to be background agnostic, **which it already mostly is**, thanks to its glaring implementation issues.

The latter simply makes more sense.

habamax:
> This is a bad idea, you're substituting broken behaviour for broken behaviour that will affect fewer users.

I think it is less broken, condition was inverted to prefer dark if "bg detection fails".
Background detection checks for the `COLORFGBG` env var and OSC, thus if COLORFGBG is not set or terminal doesn't speak OSC, dark is preferred.

> There are only 2 ways to fix this : either implement a detection system that can actually be relied upon over ssh and inside multiplexers, or fix the various terminal versions of the default light colorscheme to be background agnostic, which it already mostly is, thanks to its glaring implementation issues.

I can't see a reason why not have both.

habamax:
I can't find tests related to failing ones in MacOS and FreeBSD: `FAILED: 6: ['c', 'modula2_iso', 'modula2_pim', 'modula2_r10', 'sh_01', 'sh_02']`.

habamax:
Let me separate this PR into 2:

1. fix color component assignments in `handle_osc`, which is part of background detection but shouldn't be breaking
2. and the rest.

habamax:
Closing as it might be too breaking. Probably people got used to "light" background colors on dark background terminal.

Patch is easy enough (if you don't count tests) to replicate, but future bravehearts might get an idea of how many screendumps one need to verify with bare eyes to fix tests. :)

PR #13979, opened 2024-02-05
Unnecessary call to redraw_for_cursorline() in nv_mousescroll()

zeertzjq:
Problem:  The call to redraw_for_cursorline() in nv_mousescroll() is
          unnecessary because redraw_for_cursorline() only sets redraw
          type to UPD_VALID, and all code paths in do_mousescroll()
          already set redraw type to at least UPD_VALID.
Solution: Remove call to redraw_for_cursorline() in nv_mousescroll().


chrisbra:
thanks!

PR #13981, opened 2024-02-06
fix green and blue color components assignment in handle_osc

habamax:


chrisbra:
hm, what problem does this fix?

habamax:
> hm, what problem does this fix?

For xterm or any other term that report background via OSC with 

```
      {lead}{code};rgb:{rrrr}/{gggg}/{bbbb}{tail}
or    {lead}{code};rgb:{rr}/{gg}/{bb}{tail}
```

color components (red, green and blue) are used to detect whether to use light or dark background.

So we have

```
		    rval = hexhex2nr(tp_r);
		    gval = hexhex2nr(tp_g);
		    bval = hexhex2nr(tp_b);
#endif
		    if (is_bg)
		    {
			char *new_bg_val = (3 * '6' < *tp_r + *tp_g +
					     *tp_b) ? "light" : "dark";
```

where green and blue components are mixed, most probably by mistake.

It shouldn't matter that much as, I guess, most of xterm backgrounds users set the same red, green and blue for the chosen color (`#000000`, `#d0d0d0`, `#1c1c1c` etc) but there might be ppl using less plain colors, e.g. `#3faf5f` (imaginary example).

chrisbra:
I see thanks!

PR #13982, opened 2024-02-06
trigger CompleteChanged even no match item when completion active

glepnir:
Problem:
after add new leader and when is no matched items (`compl_match_array` is NULL) , vim doesn't fire any events.

Solution:
when completion is actived and no item matched.  still trigger CompletChanged event


chrisbra:
Hm, the test seems to succeed even without your code changes?

glepnir:
yup sry for confused. should be fine now.

chrisbra:
thanks!

PR #13985, opened 2024-02-07
Too much redraw when inserting/deleting lines with 'number'

zeertzjq:
Problem:  Too much redraw when inserting/deleting lines with 'number'.
Solution: Only redraw the number column of lines below changed lines.
          Add a test as this wasn't previously tested.


zeertzjq:
It seems that two tests in `test_number.vim` are disabled, but they do pass apart from a 'smoothscroll' marker, so I enabled them.

chrisbra:
thanks

Issue #4862, opened 2019-08-25
CursorLine highlighting applied to previous line after pasting a new line above


lacygoill:
**Describe the bug**

After pasting a line above the current one, while `'cursorline'` is set, the previous line is sometimes wrongly highlighted by CursorLine.

**To Reproduce**

1. Run this shell command:

       $ echo test | vim -Nu NONE +'set cursorline' -

2. Press `yy` to yank the current line, then `P` to paste it above.

Both lines are highlighted with CursorLine.

**Expected behavior**

Only the first line is highlighted with CursorLine.

**Screenshots**

![gif](https://user-images.githubusercontent.com/8505073/63648685-947d2600-c733-11e9-8528-e74329a005ef.gif)

**Environment**

 - Vim version: 8.1  Included patches: 1-1890
 - OS: Ubuntu 16.04.6 LTS
 - Terminal: st

**Additional context**

The same issue applies when `:put!` is run, or `[p` is pressed.

The issue disappears:

   - after the cursor is moved on another line
   - if `'conceallevel'` is set to a non-zero value
   - if some syntax item is installed (i.e. the output of `:syn list` is not `No Syntax items defined for this buffer`)

brammool:
> **Describe the bug**
> 
> After pasting a line above the current one, while `'cursorline'` is
> set, the previous line is sometimes wrongly highlighted by CursorLine.
> 
> **To Reproduce**
> 
> 1. Run this shell command:
> 
>        $ echo test | vim -Nu NONE +'set cursorline' -
> 
> 2. Press `yy` to yank the current line, then `P` to paste it above.
> 
> Both lines are highlighted with CursorLine.
> 
> **Expected behavior**
> 
> Only the first line is highlighted with CursorLine.
> 
> **Screenshots**
> 
> ![gif](https://user-images.githubusercontent.com/8505073/63648685-947d2600-c733-11e9-8528-e74329a005ef.gif)
> 
> **Environment**
> 
>  - Vim version: 8.1  Included patches: 1-1890
>  - OS: Ubuntu 16.04.6 LTS
>  - Terminal: st
> 
> **Additional context**
> 
> The same issue applies when `:put!` is run, or `[p` is pressed.
> The issue disappears if `'conceallevel'` is set to a non-zero value.
> The issue disappears after the cursor is moved on another line.

I can easily reproduce it. And can fix it.  Strangely I could not make
it fail in a screenshot test.  Would be good to have a test for this.

-- 
[...]

PR #13986, opened 2024-02-07
Too much redraw when deleting lines with 'cursorline'

zeertzjq:
Problem:  Too much redraw when deleting lines with 'cursorline'.
Solution: Use smarter invalidation and adjustment.  Remove unnecessary
          UPD_VALID as it is already set at the top of the loop.  Make
          the test for #4862 fail without the fix.


chrisbra:
thanks!

PR #13992, opened 2024-02-08
fix restore lastused_tabpage too early

glepnir:
Problem:
Restore `lastused_tabpage` too eary  in `do_arg_all` function it still changed in under logic of function.

Solution:
Restore at the end of tabpage operation done.


chrisbra:
thanks!

PR #13996, opened 2024-02-09
Visual hl wrong when it ends before multibyte 'showbreak'

### LINKED ISSUE #11272: end of visual selection wrongly extended on wrapped line when multibyte character in 'showbreak'
**Steps to reproduce**

Run this shell command:

    vim -Nu NONE -S <(tee <<'EOF'
        vim9script
        &wrap = true
        &linebreak = true
        &showbreak = '↪ '
        ['xxxxx ' .. 'y'->repeat(&columns - 6) .. ' zzzz']->setline(1)
        normal! wvel
    EOF
    )

The last visually selected character is the second `z` on the third screen line.

**Expected behavior**

The last visually selected character is the space after the last `y` on the second screen line.

**Version of Vim**

9.0 Included patches: 1-651

**Environment**

Operating system: Ubuntu 20.04.5 LTS
Terminal: XTerm(353)

**Gif**

gif

**Additional context**

No issue if we replace `↪` with `+`:
```vim
vim9script
&wrap = true
&linebreak = true
&showbreak = '+ '
['xxxxx ' .. 'y'->repeat(&columns - 6) .. ' zzzz']->setline(1)
normal! wvel
```
I guess the issue is related to the fact that the 'showbreak' character is encoded with multiple bytes (`e2 86 aa` if we press `g8` over it), contrary to `+` which is encoded with a single byte (`2b` if we press `g8` over it).

---

The cursor is drawn in the correct position; it's only the end of the visual selection which seems to be extended 2 characters to the right.  And if we press `l`, the end of the visual selection moves 1 character back to the left.  The new position is correct, but the way the selection changes is confusing (if we press `l`, it should extend; instead, it shrinks).

zeertzjq:
Problem:  Visual hl wrong when it ends before multibyte 'showbreak'.
Solution: Use vcol_sbr instead of adding n_extra.

Fix #11272


chrisbra:
thanks!

PR #13998, opened 2024-02-09
feat: Getregion function

Shougo:
Continues of #11579

It is not only for visual mode.


runtime/doc/builtin.txt
```diff
@@ -4261,6 +4261,19 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2}				*getregion()*
+		The result is a list of strings where each item is a line
+		from {pos1} to {pos2}.
```
zeertzjq: The docs doesn't mention that `pos1` and `pos2` are swapped if `pos1` is after `pos2`.

Shougo: OK.  I have added it.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4261,20 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2}				*getregion()*
```
chrisbra: missing closing parenthesis

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4261,20 @@ getreginfo([{regname}])					*getreginfo()*
...
+		The result is a list of strings where each item is a line
+		from {pos1} to {pos2}.  {pos1} and {pos2} are swapped if
```
chrisbra: > The result is a list of strings where each item is a line from {pos1} to {pos2}

I don't understand that. What does this mean.

> {pos1} and {pos2} are swapped if {pos1} is after {pos2}.

Same here, I think you mean, it will always return from the upper left position to the lower right position, right?

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	{
+	    // handle 'selection' == "exclusive"
+	    if (*p_sel == 'e' && !EQUAL_POS(p1, p2)) {
+		if (p2.coladd > 0)
+		    p2.coladd--;
+		else if (p2.col > 0)
+		{
+		    p2.col--;
+		    mb_adjustpos(curbuf, &p2);
+		} else if (p2.lnum > 1)
```
chrisbra: separate line please

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+			p2.col--;
+			mb_adjustpos(curbuf, &p2);
+		    }
+		}
+	    }
+	    // if fp2 is on NUL (empty line) inclusive becomes false
+	    if (*ml_get_pos(&p2) == NUL && !virtual_op) {
+		inclusive = FALSE;
+	    }
+	} else if (VIsual_mode == Ctrl_V) {
```
chrisbra: same here

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	    oap.start_vcol = MIN(sc1, sc2);
+	    if (*p_sel == 'e' && ec1 < sc2 && 0 < sc2 && ec2 > ec1)
+		oap.end_vcol = sc2 - 1;
+	    else
+		oap.end_vcol = MAX(ec1, ec2);
+	}
+    }
+
+    // Include the trailing byte of a multi-byte char.
+    const int l = utfc_ptr2len((char *)ml_get_pos(&p2));
```
chrisbra: please move declaration to the top of the function.

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	{
+	    list_free(rettv->vval.v_list);
+	    break;
+	}
+    }
+    virtual_op = MAYBE;
+}
+
+/*
+ * "wildmenumode()" function
```
chrisbra: why are you moving the wilemenumode() and windowsversion() functions around? Please don't change lines that do not belong to the functionality.

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    if (ret != NULL) {
+	p = ret;
+	memset(p, ' ', bd->startspaces);
+	p += bd->startspaces;
+	memmove(p, bd->textstart, bd->textlen);
+	p += bd->textlen;
+	memset(p, ' ', bd->endspaces);
+	*(p + bd->endspaces) = NUL;
+    }
+    return ret;
```
chrisbra, line 5478: I am wondering what happens if ret == NULL?

src/list.c
```diff
@@ -697,6 +697,25 @@ list_append_string(list_T *l, char_u *str, int len)
     return OK;
 }
 
+
+/*
+ * Similar to list_append_string() but does not make copy of the 'str'.
+ * Returns FAIL when out of memory.
+ */
+    int
+list_append_string_move(list_T *l, char_u *str)
```
chrisbra: This function seems to similar to `list_append_string()` that  Ithink we should add a new argument to `list_append_string()` on whether we should allocate the string items or not. So that we do not duplicate code.

chrisbra:
I think this is a nice addition. I think we should call it `getline_region()` so that it comes close to the description of the `getline()` function.

However, we need tests here. Including testing what happens with invalid positions, multibyte character support, tab handling, and testing failure modes if possible.

src/evalfunc.c
```diff
@@ -5452,6 +5455,173 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    }
+
+    // Include the trailing byte of a multi-byte char.
+    const int l = utfc_ptr2len((char *)ml_get_pos(&p2));
+    if (l > 1)
+	p2.col += l - 1;
+
+    for (lnum = p1.lnum; lnum <= p2.lnum; lnum++)
+    {
+	switch (VIsual_mode)
```
zeertzjq: It's strange for the type of the region to depend on current Visual mode type when the function is not only for Visual mode.

Shougo:
> I think this is a nice addition. I think we should call it getline_region() so that it comes close to the description of the getline() function.

Well, the rename is needed?

> However, we need tests here. Including testing what happens with invalid positions, multibyte character support, tab handling, and testing failure modes if possible.

Yes.  The tests are needed...

src/evalfunc.c
zeertzjq: How is this fixed? You are still using the value of `VIsual_mode`.

Shougo: I get it.  I have added the check.

Shougo:
I have fixed the bug and I have added basic visual mode tests.

> However, we need tests here. Including testing what happens with invalid positions, multibyte character support, tab handling, and testing failure modes if possible.

Can you add them?

chrisbra:
Thinking about this some more, does it handle block mode correctly? I mean for a text like this:

```
one
foobar
a
```
With `:call cursor(1,1)`, how do we get the proper region for: `:norm! \<c-v>jj`  vs:  `norm! \<c-v>jj$` ?  Don't we need to add an argument for the type of selection?

I am also wondering, if instead of values for the `line()` function, the positions should be what e.g. `getpos()` returns. That would be more flexible I think.

Hm, actually, reading the documentation again:
```
		The result is a list of strings where each item is a line
		from {pos1} to {pos2}. 
                For possible values of {pos1} and {pos2} see |line()|.
```

This seems strange.  `line()` returns only the line number, how is this different from `getline({pos1}, {pos2})` ? That does exactly the same? Or not?

habamax:
I am hoping that this PR inherits what was in original `visualtext()` that works with visual block selections as well. https://github.com/vim/vim/pull/11579

Shougo:
> This seems strange. line() returns only the line number, how is this different from getline({pos1}, {pos2}) ? That does exactly the same? Or not?

The arguments are same with `getpos()`.
But in `getpos()` documentation,

```
							*getpos()*
getpos({expr})	Get the position for String {expr}.  For possible values of
		{expr} see |line()|.  For getting the cursor position see
		|getcurpos()|.
```

So I have inherited it.

Shougo:
> With `:call cursor(1,1)`, how do we get the proper region for: `:norm! \<c-v>jj` vs: `norm! \<c-v>jj$` ? Don't we need to add an argument for the type of selection?

The selection type is known from Vim internal state.

Shougo:
I have added the visual block mode tests.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,20 @@ getreginfo([{regname}])					*getreginfo()*
...
```
habamax: Returns the list of strings from {pos1} to {pos2}.

For possible values of {pos1} and {pos2} see |line()|.

Useful to get text starting and ending in different columns, such as |characterwise-visual| selection.

Note that 
    - if |virtualmode| is enabled and selection is past the end of line, result lines are filled with spaces,
    - if selection starts or ends in the middle of multiwidth character (e.g. <tab>), it is not included but it's selected part is substituted with spaces.

Example of getting text of visual selection:
....
Example of getting text of visual-block selection:
....
Example of getting text from arbitrary pos1 and pos2
...
Example of getting text from mark1 and mark2
...

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,20 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2}				*getregion()*
```
mityu: Seems a typo.  Closing parenthesis is missing after `{pos2}`.

runtime/doc/builtin.txt
Shougo: I have updated.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,23 @@ getreginfo([{regname}])					*getreginfo()*
...
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2})				*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
+		It is useful to get text starting and ending in different
+		columns, such as |characterwise-visual| selection.
+
+		Note that
+		- {pos1} and {pos2} are swapped if {pos1} is after {pos2}.
```
habamax: thx! minor thing about typography (I don't know what vim help prefers), but either way list items should be separated by `,` or `;` and/or `.` for the last list item.

Shougo: I have checked the other documentation.  It seems terminated by ".".

chrisbra: I have mentioned it before, I think this is confusing and I still don't know what is meant with swapping. I think it is more clear to say:

```
Order of {pos1} and {pos2} doesn't matter, it will always return content from the upper left position to the lower right position.
```

mityu:
How about to add functionality to specify the way to get text such as charwise, linewise, and blockwise?
For example, add a optional third argument in dictionary:
```vim
call getregion("'a", "'b", #{mode: 'v'})  " Get text in charwise mode.
call getregion("'a", "'b", #{mode: 'V'})  " Get text in linewise mode.
call getregion("'a", "'b", #{mode: "\<C-v>"})  " Get text in blockwise mode.
```
Different from the base `visualtext` function, `getregion` function can get not only the selected text in visual mode but also any text between two marks.  Therefore it's not reasonable that the type of region is just set from the current mode and users cannot change it.

I'm not sure for this but maybe it's better to have a functionality to specify selection (old, inclusive, or exclusive) like this: `getregion("'a", "'b", #{selection: 'exclusive'})`.

src/evalfunc.c
```diff
@@ -2131,6 +2132,8 @@ static funcentry_T global_functions[] =
 			ret_getreg,	    f_getreg},
     {"getreginfo",	0, 1, FEARG_1,	    arg1_string,
 			ret_dict_any,	    f_getreginfo},
+    {"getregion",	2, 2, FEARG_2,	    NULL,
```
yegappan: Can you use a proper Vim9 argument type checking function here instead of NULL?

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,23 @@ getreginfo([{regname}])					*getreginfo()*
...
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
+		It is useful to get text starting and ending in different
+		columns, such as |characterwise-visual| selection.
+
+		Note that:
+		- {pos1} and {pos2} are swapped if {pos1} is after {pos2}.
+		- If |virtualmode| is enabled and selection is past the end of
+		  line, result lines are filled with spaces.
+		- If selection starts or ends in the middle of multiwidth
```
yegappan: Typo: "in the middle of a multibyte"

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,23 @@ getreginfo([{regname}])					*getreginfo()*
...
+getregion({pos1}, {pos2})				*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
+		It is useful to get text starting and ending in different
+		columns, such as |characterwise-visual| selection.
+
+		Note that:
+		- {pos1} and {pos2} are swapped if {pos1} is after {pos2}.
+		- If |virtualmode| is enabled and selection is past the end of
+		  line, result lines are filled with spaces.
```
yegappan: Suggestion: "resulting lines are filled"

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,40 @@ func Test_visual_substitute_visual()
   bwipe!
 endfunc
 
+func Test_visual_getregion()
+  new
+  call setline(1, ['one', 'two', 'three'])
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>vl"
```
yegappan: Is it possible to use feedkeys() here Instead of "normal!"?  Does it end the visual selection?

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,40 @@ func Test_visual_substitute_visual()
...
+  call assert_equal(['on'], getregion('.', 'v'))
+  call assert_equal(['o'], getregion('v', 'v'))
+  call assert_equal(['n'], getregion('.', '.'))
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>Vll"
+  call assert_equal(['one'], getregion('v', '.'))
+  call assert_equal(['one'], getregion('.', 'v'))
+  call assert_equal(['one'], getregion('v', 'v'))
+  call assert_equal(['e'], getregion('.', '.'))
```
yegappan: Can you add a test for line-wise selection across multiple lines?

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,40 @@ func Test_visual_substitute_visual()
...
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>\<C-v>jj"
+  call assert_equal(['o', 't', 't'], getregion('v', '.'))
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>\<C-v>jj$"
+  call assert_equal(['one', 'two', 'three'], getregion('v', '.'))
+
+  bwipe!
```
yegappan, line 1758: Can you add a Vim9 test to the test_vim9_builtin.vim script?

src/ops.c
```diff
@@ -2413,6 +2413,81 @@ block_prep(
 #endif
 }
 
+    void
```
yegappan, line 2419: Can you add a comment here describing the purpose of this function?

src/ops.c
```diff
@@ -2413,6 +2413,81 @@ block_prep(
...
+charwise_block_prep(
+    pos_T start,
+    pos_T end,
+    struct block_def *bdp,
+    linenr_T lnum,
+    int inclusive)
```
yegappan: The Vim coding style is to align the function arguments.

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    pos_T *fp = NULL;
+    char_u *str1, *str2;
+    int is_visual = FALSE;
+
+    if (rettv_list_alloc(rettv) == FAIL)
+	return;
+
+    virtual_op = virtual_active();
+
+    if (argvars[0].v_type != VAR_STRING || argvars[1].v_type != VAR_STRING)
```
yegappan: Can you use the check_for_string_arg() function here for both the arguments?

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,40 @@ func Test_visual_substitute_visual()
...
+  call assert_equal(['e'], getregion('.', '.'))
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>\<C-v>jj"
+  call assert_equal(['o', 't', 't'], getregion('v', '.'))
+
+  call cursor(1, 1)
+  execute "normal!" "\<ESC>\<C-v>jj$"
+  call assert_equal(['one', 'two', 'three'], getregion('v', '.'))
+
```
yegappan, line 1757: Can you add a test for passing invalid type of arguments to getregion()?

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	    // if fp2 is on NUL (empty line) inclusive becomes false
+	    if (*ml_get_pos(&p2) == NUL && !virtual_op)
+	    {
+		inclusive = FALSE;
+	    }
+	}
+	else if (VIsual_mode == Ctrl_V)
+	{
+	    colnr_T sc1, ec1, sc2, ec2;
+	    getvvcol(curwin, &p1, &sc1, NULL, &ec1);
```
yegappan: Can you add an empty line here?

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
     getpos_both(argvars, rettv, FALSE, FALSE);
 }
 
+   static char_u *
```
yegappan, line 5461: Can you add a description of this function here?

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+block_def2str(struct block_def *bd)
+{
+    char_u *p, *ret;
+    size_t size = bd->startspaces + bd->endspaces + bd->textlen;
+
+    ret = alloc(size + 1);
+    if (ret != NULL)
+    {
+	p = ret;
+	memset(p, ' ', bd->startspaces);
```
yegappan: Can you use vim_memset() here?

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	p += bd->startspaces;
+	memmove(p, bd->textstart, bd->textlen);
```
yegappan: Can you use mch_memmove() here?

src/evalfunc.c
```diff
@@ -5452,6 +5455,174 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	p += bd->textlen;
+	memset(p, ' ', bd->endspaces);
```
yegappan: Can you use vim_memset() here?

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,23 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2})				*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
```
yegappan, line 4269: This function uses the line and column number from the specified position.  The line() function returns only the line number.  I think that should be mentioned here.   The supplied position argument specifies the starting and ending line number and column number.

src/testdir/test_visual.vim
yegappan, line 1758: Can you add a test for using this function in virtual mode and check for the additional space characters?

yegappan, line 1758: Can you add a test for using this function with multibyte characters and also if the selection ends in the middle of a multibyte character?

yegappan (changes requested):

Shougo:
Oh, too many request here...  Please wait.  I need time to fix them.

runtime/doc/builtin.txt
Shougo: Updated.

Shougo:
> How about to add functionality to specify the way to get text such as charwise, linewise, and blockwise?
For example, add a optional third argument in dictionary:

Well, the suggestion may be good.
But it will increase complexity for interface and implementation.

I think the simple interface should be implemented in the first instead of perfect interface.
You can extend the implementation if you really needed.

src/testdir/test_visual.vim
Shougo: What is the additional space characters?

Please provide the examples.

Shougo: Well, it is possible?
Please provide the example test.

habamax: The ones you add with visual block that extends past endofline

Shougo:
The original API `virtualtext()` cannot be extended.
But this API can be added the following opts.

`winid`, `mode`, `inclusive`

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,25 @@ getreginfo([{regname}])					*getreginfo()*
...
+		This function uses the line and column number from the
+		specified position.
+		It is useful to get text starting and ending in different
+		columns, such as |characterwise-visual| selection.
+
+		Note that:
+		- {pos1} and {pos2} are swapped if {pos1} is after {pos2}.
+		- If |virtualmode| is enabled and selection is past the end of
```
yegappan: I think you meant "virtualedit" here instead of "virtualmode".

src/testdir/test_visual.vim
yegappan: I am referring to the following text in the help for getregion():

```
If |virtualmode| is enabled and selection is past the end of line, result lines are filled.
```

You can add the following test:

```
  " Virtual edit
  set virtualedit=all
  call cursor(1, 3)
  call feedkeys("\<Esc>\<C-v>lljj", 'xt')
  call assert_equal(['e  ', 'o  ', 'ree'], getregion('v', '.'))
  set virtualedit&
```

yegappan: I am referring to the following text in the help for getregion():

```
If selection starts or ends in the middle of multibyte character, it is not included but it's selected part is substituted with spaces.
```

You can use the following test:
```
  " Selection in starts or ends in the middle of a multibyte character
  %d _
  call setline(1, ['abcdefghijk', "\U0001f1e6\u00ab\U0001f1e7\u00ab\U0001f1e8\u00ab\U0001f1e9", '1234567890'])
  call cursor(1, 3)
  call feedkeys("\<Esc>\<C-v>ljj", 'xt')
  call assert_equal(['cd', "\u00ab ", '34'], getregion('v', '.'))
  call cursor(1, 4)
  call feedkeys("\<Esc>\<C-v>ljj", 'xt')
  call assert_equal(['de', "\U0001f1e7", '45'], getregion('v', '.'))
  call cursor(1, 5)
  call feedkeys("\<Esc>\<C-v>jj", 'xt')
  call assert_equal(['e', ' ', '5'], getregion('v', '.'))
```

yegappan (changes requested):

src/evalfunc.c
```diff
@@ -2131,6 +2132,8 @@ static funcentry_T global_functions[] =
 			ret_getreg,	    f_getreg},
     {"getreginfo",	0, 1, FEARG_1,	    arg1_string,
 			ret_dict_any,	    f_getreginfo},
+    {"getregion",	2, 2, FEARG_2,	    arg2_string,
```
yegappan: I think this should be FEARG_1 instead of FEARG_2.  This determines which argument is used when getregion() is called using the method syntax.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4262,25 @@ getreginfo([{regname}])					*getreginfo()*
...
+		  line, result lines are filled.
+		- If selection starts or ends in the middle of multibyte
+		  character, it is not included but it's selected part
+		  is substituted with spaces.
+
+		Examples: >
+			:xnoremap <CR>  <Cmd>echo getregion('v', '.')<CR>
+<
```
yegappan, line 4302: Can you also add the following text?

```
    Can also be used as a method:
                  StartPos->getregion(endpos)
```

src/evalfunc.c
```diff
@@ -5452,6 +5455,180 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	vim_memset(p, ' ', bd->endspaces);
+	*(p + bd->endspaces) = NUL;
+    }
+    return ret;
+}
+
+/*
+ * "getregion()" function
+ */
+static void f_getregion(typval_T *argvars, typval_T *rettv)
```
yegappan: The Vim coding convention is to place the return value in a separate line.  Can you change the above to:

```
   static void
f_getregion(...)
```

runtime/doc/builtin.txt
Shougo: Thanks.  I will fix it.

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,56 @@ func Test_visual_substitute_visual()
...
+  call assert_equal(['one'], getregion('v', 'v'))
+  call assert_equal(['e'], getregion('.', '.'))
+
+  " Block visual mode
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>ll", 'tx')
+  call assert_equal(['one'], getregion('v', '.'))
+  call assert_equal(['one'], getregion('.', 'v'))
+  call assert_equal(['o'], getregion('v', 'v'))
+  call assert_equal(['e'], getregion('.', '.'))
```
yegappan: Can you change one of the getregion() calls to use the method calling syntax?

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,56 @@ func Test_visual_substitute_visual()
...
+
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>jj$", 'tx')
+  call assert_equal(['one', 'two', 'three'], getregion('v', '.'))
+
+  " Invalid pattern
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>jj$", 'tx')
+  call assert_fails("call getregion(1, 2)")
+  call assert_fails("call getregion('.', {})")
```
yegappan: In the assert_fails() calls, can you also include the actual error number and the message?

yegappan (changes requested):

src/evalfunc.c
```diff
@@ -5452,6 +5455,180 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    linenr_T lnum;
+    oparg_T oap;
+    struct block_def bd;
+    char_u *akt;
+    int inclusive = TRUE;
+    int fnum = -1;
+    pos_T p1, p2;
+    pos_T *fp = NULL;
+    char_u *str1, *str2;
+    int is_visual = FALSE;
```
yegappan: The Vim coding convention is to align all the variable names after the type.

mityu:
> Well, the suggestion may be good.
> But it will increase complexity for interface and implementation.
>
> I think the simple interface should be implemented in the first instead of perfect interface.
> You can extend the implementation if you really needed.

I see. OK, thank you.

src/evalfunc.c
```diff
@@ -5452,6 +5455,181 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	    akt = block_def2str(&bd);
+	}
+
+	if (akt == NULL || list_append_string_move(rettv->vval.v_list, akt) == FAIL)
+	{
+	    list_free(rettv->vval.v_list);
+	    break;
+	}
+    }
+    virtual_op = MAYBE;
```
yegappan: Instead of overwriting `virtual_op`, should the old value be saved at the beginning of this function and restored here?

src/evalfunc.c
```diff
@@ -5452,6 +5455,181 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+			p2.col--;
+			mb_adjustpos(curbuf, &p2);
+		    }
+		}
+	    }
+	    // if fp2 is on NUL (empty line) inclusive becomes false
+	    if (*ml_get_pos(&p2) == NUL && !virtual_op)
+	    {
+		inclusive = FALSE;
+	    }
```
yegappan, line 5575: Vim coding convention is not to use curly brackets for a single-statement block.

src/evalfunc.c
```diff
@@ -5452,6 +5455,181 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+
+    // Include the trailing byte of a multi-byte char.
+    const int l = utfc_ptr2len((char_u *)ml_get_pos(&p2));
+    if (l > 1)
+	p2.col += l - 1;
+
+    for (lnum = p1.lnum; lnum <= p2.lnum; lnum++)
+    {
+	if (is_visual && VIsual_mode == 'V')
+		akt = vim_strsave(ml_get(lnum));
```
yegappan: I think the indentation here is not correct.

src/evalfunc.c
```diff
@@ -5452,6 +5455,181 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	else if (is_visual && VIsual_mode == Ctrl_V)
+	{
+		block_prep(&oap, &bd, lnum, FALSE);
+		akt = block_def2str(&bd);
```
yegappan: I think the indentation for this block is not correct.

src/evalfunc.c
```diff
@@ -5452,6 +5455,181 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    }
+
+    is_visual = (str1[0] == 'v' && str1[1] == NUL)
+	|| (str2[0] == 'v' && str2[1] == NUL);
+    if (is_visual)
+    {
+	if (VIsual_mode == 'v')
+	{
+	    // handle 'selection' == "exclusive"
+	    if (*p_sel == 'e' && !EQUAL_POS(p1, p2))
```
yegappan: Can you add some tests for using exclusive `selection`?  There is quite a bit of logic here for handling exclusive selections which are not yet tested.

yegappan (changes requested):

runtime/doc/builtin.txt
```diff
@@ -263,6 +263,7 @@ getqflist({what})		Dict	get specific quickfix list properties
 getreg([{regname} [, 1 [, {list}]]])
 				String or List   contents of a register
 getreginfo([{regname}])		Dict	information about a register
+getregion({pos1}, {pos2})	String	get the text from {pos1} to {pos2}
```
yegappan: This function returns a List of strings.  So change "String" -> "List".

chrisbra:
> > How about to add functionality to specify the way to get text such as charwise, linewise, and blockwise?
> > For example, add a optional third argument in dictionary:
> 
> Well, the suggestion may be good. But it will increase complexity for interface and implementation.
> 
> I think the simple interface should be implemented in the first instead of perfect interface. You can extend the implementation if you really needed.

I think the suggestion is good however. Otherwise, what the `getregion()` returns depends on whether the user used `v`, `V` or `Ctrl-V` before calling the `getregion()` function. I think we should explicitly allow for a `blockmode` option therefore

src/evalfunc.c
Shougo: Well, can you add the tests?

runtime/doc/builtin.txt
Shougo: It is already fixed.

runtime/doc/builtin.txt
Shougo: > Same here, I think you mean, it will always return from the upper left position to the lower right position, right?

Yes.

src/evalfunc.c
Shougo: I am very busy to fix for other comments.

src/evalfunc.c
Shougo:
```c
	if (akt == NULL || list_append_string_move(rettv->vval.v_list, akt) == FAIL)
	{
	    list_free(rettv->vval.v_list);
	    break;
	}
```

It is checked and break the loop.

src/list.c
Shougo: I have removed the function.

Shougo:
> I think we should explicitly allow for a blockmode option therefore

Well, I should add the argument?

src/evalfunc.c
yegappan: You can add the following test:

```
  " exclusive selection
  new
  set selection=exclusive
  call setline(1, ["a\tc", "x\tz", '', ''])
  call cursor(1, 1)
  call feedkeys("\<Esc>v2l", 'xt')
  call assert_equal(["a\t"], getregion('v', '.'))
  call cursor(1, 1)
  call feedkeys("\<Esc>v$G", 'xt')
  call assert_equal(["a\tc", "x\tz", ''], getregion('v', '.'))
  call cursor(1, 1)
  call feedkeys("\<Esc>v$j", 'xt')
  call assert_equal(["a\tc", "x\tz"], getregion('v', '.'))
  call cursor(1, 1)
  call feedkeys("\<Esc>\<C-v>$j", 'xt')
  call assert_equal(["a\tc", "x\tz"], getregion('v', '.'))
  call cursor(1, 1)
  call feedkeys("\<Esc>\<C-v>$G", 'xt')
  call assert_equal(["a", "x", '', ''], getregion('v', '.'))
  set virtualedit=all
  call cursor(1, 1)
  call feedkeys("\<Esc>2lv2lj", 'xt')
  call assert_equal(['      c', 'x   '], getregion('v', '.'))
  call cursor(1, 1)
  call feedkeys("\<Esc>2l\<C-v>2l2j", 'xt')
  call assert_equal(['  ', '  ', '  '], getregion('v', '.'))
  set virtualedit&
  set selection&
  bw!
```

chrisbra:
> > I think we should explicitly allow for a blockmode option therefore
> 
> Well, I should add the argument?

Yes please. Something like this:
```
getregion({pos1}, {pos2}, {dict})			*getregion()*
		Returns the list of strings from {pos1} to {pos2}.
		For possible values of {pos1} and {pos2} see |line()|.
		This function uses the line and column number from the
		specified position.
		It is useful to get text starting and ending in different
		columns, such as |characterwise-visual| selection.

		The {dict} argument allows to specify additional options.
		Currently it supports the following members:
		
			type	Type of the selection. One of
				"v" for |characterwise| text,
				"V" for |linewise| text or
				"<CTRL-V>" for |blockwise| text
				selection.
```

That gives us the opportunity to specify more behaviours (like the selection setting mentioned elsewhere) when the need arises.

src/evalfunc.c
Shougo: Thank you!

Shougo:
I think `opt` argument should be added in the next PR.

But I will add explicit selection mode.

Shougo:
I have added the argument.

zeertzjq:
Memory leak
```
==16754==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2 byte(s) in 2 object(s) allocated from:
    #0 0x564cfabe91cf in __interceptor_malloc ??:?
    #1 0x564cfac269e7 in lalloc /home/runner/work/vim/vim/src/alloc.c:246:11
    #2 0x564cfae6368d in block_def2str /home/runner/work/vim/vim/src/evalfunc.c:5467:11
    #3 0x564cfae3d379 in f_getregion /home/runner/work/vim/vim/src/evalfunc.c
    #4 0x564cfb7af51d in call_bfunc /home/runner/work/vim/vim/src/vim9execute.c:1332:5
    #5 0x564cfb7561f6 in exec_instructions /home/runner/work/vim/vim/src/vim9execute.c:4379:7
    #6 0x564cfb793c00 in call_def_function /home/runner/work/vim/vim/src/vim9execute.c:6325:11
    #7 0x564cfb6b89cb in call_user_func /home/runner/work/vim/vim/src/userfunc.c:2866:6
    #8 0x564cfb6b7b00 in call_user_func_check /home/runner/work/vim/vim/src/userfunc.c:3299:10
    #9 0x564cfb6b22df in call_func /home/runner/work/vim/vim/src/userfunc.c:3923:15
    #10 0x564cfb6b0772 in get_func_tv /home/runner/work/vim/vim/src/userfunc.c:2013:8
    #11 0x564cfb6db13b in ex_call_inner /home/runner/work/vim/vim/src/userfunc.c:6103:6
    #12 0x564cfb6db13b in ex_call /home/runner/work/vim/vim/src/userfunc.c:6451:11
    #13 0x564cfaed5097 in do_one_cmd /home/runner/work/vim/vim/src/ex_docmd.c:2582:2
    #14 0x564cfaeccac3 in do_cmdline /home/runner/work/vim/vim/src/ex_docmd.c:994:17
    #15 0x564cfae15126 in ex_execute /home/runner/work/vim/vim/src/eval.c:7556:6
    #16 0x564cfaed5097 in do_one_cmd /home/runner/work/vim/vim/src/ex_docmd.c:2582:2
    #17 0x564cfaeccac3 in do_cmdline /home/runner/work/vim/vim/src/ex_docmd.c:994:17
    #18 0x564cfb6ba85c in call_user_func /home/runner/work/vim/vim/src/userfunc.c:3126:2
    #19 0x564cfb6b7b00 in call_user_func_check /home/runner/work/vim/vim/src/userfunc.c:3299:10
    #20 0x564cfb6b22df in call_func /home/runner/work/vim/vim/src/userfunc.c:3923:15
    #21 0x564cfb6b0772 in get_func_tv /home/runner/work/vim/vim/src/userfunc.c:2013:8
    #22 0x564cfb6db13b in ex_call_inner /home/runner/work/vim/vim/src/userfunc.c:6103:6
    #23 0x564cfb6db13b in ex_call /home/runner/work/vim/vim/src/userfunc.c:6451:11
    #24 0x564cfaed5097 in do_one_cmd /home/runner/work/vim/vim/src/ex_docmd.c:2582:2
    #25 0x564cfaeccac3 in do_cmdline /home/runner/work/vim/vim/src/ex_docmd.c:994:17
    #26 0x564cfb425507 in do_source_ext /home/runner/work/vim/vim/src/scriptfile.c:1761:5
    #27 0x564cfb42336e in do_source /home/runner/work/vim/vim/src/scriptfile.c:1907:12
    #28 0x564cfb42336e in cmd_source /home/runner/work/vim/vim/src/scriptfile.c:1253:14
    #29 0x564cfaed5097 in do_one_cmd /home/runner/work/vim/vim/src/ex_docmd.c:2582:2
    #30 0x564cfaeccac3 in do_cmdline /home/runner/work/vim/vim/src/ex_docmd.c:994:17
    #31 0x564cfbaa01b4 in exe_commands /home/runner/work/vim/vim/src/main.c:3173:2
    #32 0x564cfbaa01b4 in vim_main2 /home/runner/work/vim/vim/src/main.c:790:2
    #33 0x564cfba98c0c in main /home/runner/work/vim/vim/src/main.c:441:12
```

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2}, {type})			*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
+		The {type} is region type:
```
yegappan, line 4270: The third argument should be an optional Dict.  This will allow us to add additional optional arguments as Dict items in the future.  Otherwise, you need to specify all the mandatory arguments.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
...
+			"v" for |charwise| mode
+			"V" for |linewise| mode
+			"<CTRL-V>" for |blockwise-visual| mode
```
yegappan, line 4273: It is not clear to me how the specified visual mode type is used.  I can see that irrespective of the current visual mode and the mode type argument, getregion() always returns the visually selected region of text.  I expected that if a region of text is selected using character-wise visual mode and getregion() is called with "V", then it will return an empty List.

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,118 @@ func Test_visual_substitute_visual()
...
+  call assert_equal(['one', 'two', 'three'], getregion('v', '.', 'V'))
+
+  " Multiline with block visual mode
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>jj", 'tx')
+  call assert_equal(['o', 't', 't'], getregion('v', '.', "\<C-v>"))
+
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>jj$", 'tx')
+  call assert_equal(['one', 'two', 'three'], getregion('v', '.', "\<C-v>"))
```
yegappan, line 1680: Can you add a test for using getregion() with "V" but the text is selected using block-wise visual mode.  And additional tests for other combinations?

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,118 @@ func Test_visual_substitute_visual()
...
+
+  " 'virtualedit'
+  set virtualedit=all
+  call cursor(1, 1)
+  call feedkeys("\<ESC>\<C-v>10ljj$", 'tx')
+  call assert_equal(['one   ', 'two   ', 'three '],
+        \ getregion('v', '.', "\<C-v>"))
+  set virtualedit&
+
+  " Invalid pattern
```
yegappan, line 1690: I see that even if the text is not visually selected, getregion() always returns some text.  I expected that getregion() will return an empty List, if text is not visually selected.

yegappan (changes requested):

runtime/doc/builtin.txt
Shougo: I don't think `type` should be optional.

The optional arguments should be added later.

Shougo:
> Memory leak

Fixed.

src/testdir/test_visual.vim
Shougo: Fixed.

src/testdir/test_visual.vim
Shougo: I have added them.

runtime/doc/builtin.txt
Shougo: `{pos1}`, `{pos2}` is just used as position.
position `v` means `use visual mode start position`.

yegappan: I am referring to the "v" or "V" value in the new {type} argument.  For example, you select some characters using character-wise visual mode and then call getregion() with the "V" or "<CTR-V>" for the {type} argument, what value will it return?

Shougo: It depends on your cursor position.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
+getregion({pos1}, {pos2}, {type})			*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
+		For possible values of {pos1} and {pos2} see |line()|.
+		The {type} is region type:
+			"v" for |charwise| mode
```
```suggestion
			"v" for |characterwise| mode
```

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
...
+
+		Note that:
+		- Order of {pos1} and {pos2} doesn't matter, it will always
+		  return content from the upper left position to the lower
+		  right position.
+		- If 'virtualedit' is enabled and selection is past the end of
+		  line, result lines are filled.
+		- If selection starts or ends in the middle of multibyte
+		  character, it is not included but it's selected part
+		  is substituted with spaces.
```
```suggestion
		- If selection starts or ends in the middle of a multibyte
		  character, it is not included but its selected part is
		  substituted with spaces.
```

src/evalfunc.c
```diff
@@ -5452,6 +5455,186 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+    pos1 = tv_get_string(&argvars[0]);
+    pos2 = tv_get_string(&argvars[1]);
+    type = tv_get_string(&argvars[2]);
+
+    is_visual = (pos1[0] == 'v' && pos1[1] == NUL)
+	|| (pos2[0] == 'v' && pos2[1] == NUL);
+
+    if (is_visual && !VIsual_active)
+	return;
+
```
zeertzjq, line 5529: Is this check still needed? It looks rather special-cased, and may access invalid memory if one position is "v" while the other position is an empty string.

Shougo: The check is needed to fix this.

https://github.com/vim/vim/pull/13998#discussion_r1493061228

Shougo: > Is this check still needed? It looks rather special-cased, and may access invalid memory if one position is "v" while the other position is an empty string.

I don't think it accesses to invalid memory.

zeertzjq: Hmm, I think it will be less special-cased if you only keep the check for "v", i.e. using "v" in either position when not in Visual mode returns an empty list.

Shougo: I don't understand your changes.  Can you patch the code?

src/evalfunc.c
```diff
@@ -5452,6 +5455,186 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	return;
+    p1 = *fp;
+
+    fp = var2fpos(&argvars[1], TRUE, &fnum, FALSE);
+    if (fp == NULL)
+	return;
+    p2 = *fp;
+
+    pos1 = tv_get_string(&argvars[0]);
+    pos2 = tv_get_string(&argvars[1]);
```
zeertzjq, line 5521: Using `tv_get_string()` here will lead to errors if the arguments are lists.

src/evalfunc.c
zeertzjq: Oops, sorry, your code already does that. I messed up.

src/evalfunc.c
Shougo: Yes.  But it is already checked.

```c
    if (check_for_string_arg(argvars, 0) == FAIL
	    || check_for_string_arg(argvars, 1) == FAIL
	    || check_for_string_arg(argvars, 2) == FAIL)
	return;
```

zeertzjq: Oh, OK.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
...
+		  right position.
+		- If 'virtualedit' is enabled and selection is past the end of
+		  line, result lines are filled.
+		- If selection starts or ends in the middle of a multibyte
+		  character, it is not included but its selected part is
+		  substituted with spaces.
+
+		Examples: >
+			:xnoremap <CR>
+			\ <Cmd>echo getregion('v', '.', visualmode())<CR>
```
zeertzjq: `visualmode()` returns the type of the last Visual mode, while a `<Cmd>` mapping remains in the current Visual mode, which should be obtained using `mode()`, so this mapping should either using `mode()` instead of `visualmode()` or use `:` instead of `<Cmd>`.

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,36 @@ getreginfo([{regname}])					*getreginfo()*
...
+<
+		Can also be used as a |method|: >
+			StartPos->getregion(endpos)
```
zeertzjq: Variables in method examples seem to be usually in lower-case (as opposed to user functions), and using lower-case `startpos` here is more consistent with `endpos`.
```suggestion
			startpos->getregion(endpos)
```

runtime/doc/builtin.txt
mityu: In my previous suggestion (https://github.com/vim/vim/pull/13998#issuecomment-1941481650), the {type} argument is for specifying which style is used to get text from {pos1} to {pos2}.  In other words, if you pass "V" for the {type} argument getregion() will return text from {pos1} to {pos2} as if it's selected in line-wise visual mode regardless of the current visual mode.  (So `getregion({pos1}, {pos2}, "V")` is also equal to `getline(getpos({pos1})[1], getpos({pos2})[1])`.)

At least I tried, the current implementation of getregion() follows this, so I think getregion() call with "V" for the {type} argument should return text of lines between the line of {pos1} and the line of {pos2}.

About getregion() call with "" for the {type} argument, I don't have an absolute one answer but I think the behavior should be one of these:
- Give an error like "{type} argument cannot be empty" or "Invalid value for the {type} argument"
- Use the current visual mode in visual mode and chararacter-wise mode in the other mode
- Use the current visual mode in visual mode and the last visual mode in the other mode

yegappan: > In my previous suggestion ([#13998 (comment)](https://github.com/vim/vim/pull/13998#issuecomment-1941481650)), the {type} argument is for specifying which style is used to get text from {pos1} to {pos2}. In other words, if you pass "V" for the {type} argument getregion() will return text from {pos1} to {pos2} as if it's selected in line-wise visual mode regardless of the current visual mode. (So `getregion({pos1}, {pos2}, "V")` is also equal to `getline(getpos({pos1})[1], getpos({pos2})[1])`.)
> 
> At least I tried, the current implementation of getregion() follows this, so I think getregion() call with "V" for the {type} argument should return text of lines between the line of {pos1} and the line of {pos2}.
> 

A few tests should be added for this.  Make a characterwise selection, call getregion() with the "v", "V" and "<C-V>" types and check the results.  Do the same for the linewise and blockwise selections.

yegappan: The help text also needs to be updated to clarify the use of the "type" argument.  Irrespective of the current visual mode, the supplied "type" will be used to determine the region of text returned.

Shougo: > A few tests should be added for this. Make a characterwise selection, call getregion() with the "v", "V" and "" types and check the results. Do the same for the linewise and blockwise selections.

I have added this.

Shougo: > The help text also needs to be updated to clarify the use of the "type" argument. Irrespective of the current visual mode, the supplied "type" will be used to determine the region of text returned.

I have updated the documentation.  But the suggestion is wellcome.

runtime/doc/builtin.txt
chrisbra: Agree, the selection type should always be given, no?

src/list.c
```diff
@@ -697,6 +697,7 @@ list_append_string(list_T *l, char_u *str, int len)
     return OK;
 }
 
+
```
chrisbra: unneccessarly line break

chrisbra:
I think it's almost good :) Just a few more tests for the different `type()` values as Yegappan suggested.

Shougo:
I have added tests.

runtime/doc/usr_41.txt
```diff
@@ -944,6 +944,7 @@ Working with text in the current buffer:		*text-functions*
 	searchdecl()		search for the declaration of a name
 	getcharsearch()		return character search information
 	setcharsearch()		set character search information
+	getregion()		get a region text from the buffer
```
zeertzjq: I think it's better to place this line below the `getline()` line.

yegappan: Can you change "get a region text" to "get a region of text"?

runtime/doc/builtin.txt
```diff
@@ -4261,6 +4263,39 @@ getreginfo([{regname}])					*getreginfo()*
...
+		Note that:
+		- Order of {pos1} and {pos2} doesn't matter, it will always
+		  return content from the upper left position to the lower
+		  right position.
+		- If 'virtualedit' is enabled and selection is past the end of
+		  line, result lines are filled.
+		- If selection starts or ends in the middle of a multibyte
+		  character, it is not included but its selected part is
+		  substituted with spaces.
+
```
yegappan, line 4295: Typically the error cases for a builtin function are described at the end of the description.  Can you add the conditions under which this function will return an empty List here?  For example, if an unsupported visual mode type is specified, an empty List is returned.  If "v" is specified as the start or end position and the visual mode is not active, then an empty List is returned.  There are other error cases also.

src/evalfunc.c
```diff
@@ -5452,6 +5455,186 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	    else if (p2.lnum > 1)
+	    {
+		p2.lnum--;
+		p2.col = (colnr_T)STRLEN(ml_get(p2.lnum));
+		if (p2.col > 0)
+		{
+		    p2.col--;
+
+		    mb_adjustpos(curbuf, &p2);
+		}
```
yegappan, line 5574: Can you add a test to cover the above two lines?

src/evalfunc.c
```diff
@@ -5452,6 +5455,186 @@ f_getpos(typval_T *argvars, typval_T *rettv)
...
+	if (*p_sel == 'e' && ec1 < sc2 && 0 < sc2 && ec2 > ec1)
+	    oap.end_vcol = sc2 - 1;
+	else
+	    oap.end_vcol = MAX(ec1, ec2);
+    }
+
+    // Include the trailing byte of a multi-byte char.
+    l = utfc_ptr2len((char_u *)ml_get_pos(&p2));
+    if (l > 1)
+	p2.col += l - 1;
```
yegappan, line 5601: Can you add test for this line?

yegappan (changes requested):

src/testdir/test_visual.vim
```diff
@@ -1630,4 +1630,126 @@ func Test_visual_substitute_visual()
   bwipe!
 endfunc
 
+func Test_visual_getregion()
```
yegappan, line 1633: Can you apply the following patch?

```
diff --git a/src/testdir/test_visual.vim b/src/testdir/test_visual.vim
index 6aad45575..b02c8c3b7 100644
--- a/src/testdir/test_visual.vim
+++ b/src/testdir/test_visual.vim
@@ -1717,6 +1717,9 @@ func Test_visual_getregion()
   call cursor(1, 5)
   call feedkeys("\<Esc>\<C-v>jj", 'xt')
   call assert_equal(['e', ' ', '5'], getregion('v', '.', "\<C-v>"))
+  call cursor(1, 1)
+  call feedkeys("\<Esc>vj", 'xt')
+  call assert_equal(['abcdefghijk', "\U0001f1e6"], getregion('v', '.', "v"))
 
   bwipe!
 
@@ -1739,6 +1742,9 @@ func Test_visual_getregion()
   call cursor(1, 1)
   call feedkeys("\<Esc>\<C-v>$G", 'xt')
   call assert_equal(["a", "x", '', ''], getregion('v', '.', "\<C-v>"))
+  call cursor(1, 1)
+  call feedkeys("\<Esc>wv2j", 'xt')
+  call assert_equal(["c", "x\tz"], getregion('v', '.', 'v'))
   set virtualedit=all
   call cursor(1, 1)
   call feedkeys("\<Esc>2lv2lj", 'xt')
```

src/evalfunc.c
Shougo: Well, I don't know how to add the tests.
Can you add it?

src/evalfunc.c
Shougo: Well, I don't know how to add the tests.
Can you add it?

runtime/doc/builtin.txt
Shougo: Fixed.

src/testdir/test_visual.vim
Shougo: I have applied.

src/evalfunc.c
chrisbra: I'll add it

habamax:
Thank you @Shougo and the rest who helped!

justinmk:
Why doesn't this return a list of *positions*? Return text is the least useful part about a region. How can plugin authors get the *positions* that correspond to a visual block region (for example)? That is still an unsolved problem.

chrisbra:
> Why doesn't this return a list of positions? Return text is the least useful part about a region. 

The functionality of this function was to be to get the buffer content of whatever those positions would define. Au contraire, getting the buffer content was an unsolved problem before, positions would need to be known before hand so I fail to see how this would be "more useful".

How can plugin authors get the positions that correspond to a visual block region (for example)? That is still an unsolved problem.

You can use getpos() of those positions already. Don't see why this is a problem.

zeertzjq:
For blockwise Visual selection it'll probably need a combination of `getpos()`, `virtcol()` and `virtcol2col()`.

chrisbra:
correct and `visualmode()`

justinmk:
> You can use getpos() of those positions already. Don't see why this is a problem.

You cannot easily get the dimensions of the buffer text that is described by a region. This is the code that is currently required to do that: https://github.com/neovim/neovim/blob/d35a8a9bbe385daf808984ba0fc7f07f51a96da0/runtime/lua/vim/_editor.lua#L509-L578

clason:
(Note that this is not as trivial as it seems once you add multibyte and ambiwidth unicode characters to the mix.)

Shougo:
Well, I should add the option to return the range feature?

kuuote:
I think range feature is good. but shouldn't add to `getregion()`.
Should split entry point to another function because determine return type by arguments is hard to understand.

chrisbra:
correct, there was a single purpose for the `getregion()` function, which was to retrieve the selected text between 2 different (start and end) positions. Now why the discussion now moves away to get different usecase. That should be a new vimscript function. However, I still don't understand the need for it:

We are giving already positions to the getrange() function. You are talking about dimensions of positions, which I don't understand. What exactly do you mean by that? Can you please give concrete examples? Why exactly are start and end positions not enough? Can you spell out the help documentation for what this function should do?

chrisbra:
Yes, but you already have those positions. So where do those positions come then from? In other words if you already have the positions, why do you need to use `getregionpos()`? 

BTW: I have for years used a similar approach to this in one of my plugins. Set the marks `'<`  and `'>` to those positions and then set the selection type. Use `norm! gv` to reselect those regions. Has worked quite well.

clason:
> Yes, but you already have those positions. So where do those positions come then from? 

No we don't; that's exactly what we're asking from here. (We have a hack that works in _some_ situations, but not enough.)

chrisbra:
Then it's even more curious why you mention it here in the PR for `getregion()`, since that function already receives the position. Can you please create a new issue with whathever you like to have? Please also mention what the help description should explain.

clason:
(To be clear, I'm not the one complaining here; just trying to provide some additional context.)

chrisbra:
sure understood and thanks. (And to be clear, I am just trying to understand the missing pieces)

clason:
Basically, given a **pair** of positions (which could be `'<` and `'>` -- the main use case, in fact) return an **array** of positions `{ line1 = {col_start, col_end}, line2 = {col_start, col_end}}` that respects 
* `setreg`-style selection type (char/line/block)
* whether ending column is inclusive/exclusive
* `'virtualedit'`
* works with double-width (flag emoji) and amiwidth (thai) unicode
* ...

Basically, convert a visual-block selection to an array of line-wise selections to then, say, apply highlights to. (Main use case here: highlight-on-yank as in https://github.com/machakann/vim-highlightedyank.) Especially the last points basically rely on Vim internals to be guaranteed to be correct

justinmk:
> Basically, convert a visual-block selection to an array of line-wise selections to then, say, apply highlights to. 

"to char-wise", right?

> (Main use case here: highlight-on-yank as in https://github.com/machakann/vim-highlightedyank.)

To @chrisbra 's point, if `nvim_buf_set_extmark` accepted `{pos1}, {pos2}, {regtype}, {inclusive}` parameters, then we wouldn't need `getregionpos()` for that particular case.

This raises the more general question about how a buffer region is described. Currently in Vim, a buffer region is described by those 4 parameters. So we have 2 choices:

- every API that accepts a "region" has those 4 parameters
- introduce a `getregionpos()` which returns a list of start/end positions, and then APIs only need to accept start/end parameters, because the caller can use `getregionpos()` and feed its result into any such API.

PR #13999, opened 2024-02-09
CI: Add Codecov token

k-takata:
codecov-action@4 requires a token.
Add it to the repository secrets.

See: https://github.com/vim/vim/pull/13978#issuecomment-1935336624


zeertzjq:
codecov-action is used twice in ci.yml

k-takata:
Ah, for linux and for windows. Fixed.

PR #14003, opened 2024-02-09
xxd: crash with higer number of column

kreijack:
xxd writes the data into a buffer before printing. Unfortunately the buffer doesn't consider the space consumed by the escape sequences used to change the color of the character.

BEFORE:
$ xxd -Ralways -c 256 /etc/passwd
Segmentation fault (core dumped)

AFTER:
$ ./xxd -Ralways -c 256 /etc/passwd
00000000: 726f 6f74 3a78 3a30 3a30 3a72 6f6f 743a 2f72 [...]

To solve this issue I had to increase the size of the buffer considering for each byte of data 11 further characters for the color escape sequence.


chrisbra:
thanks, makes sense.

PR #14004, opened 2024-02-10
Still a qsort() comparison function that returns result of subtraction

zeertzjq:
Problem:  Still a qsort() comparison function fuzzy_match_item_compare()
          that returns result of subtraction (after 9.1.0089).
Solution: Use an explicit comparison instead of subtraction.


chrisbra:
thanks!

PR #14006, opened 2024-02-10
feat: able to change the ordering of popup menu

kiddos:
hi.

I was wondering if we can have a feature to change the ordering of the popup menu. The current popup menu is ordered as "word", "kind", "extra". I'm using an option to control this for now, not sure if there's a better way to accomplish this.

https://github.com/vim/vim/assets/9444583/e4c7fbe4-be0a-4fa2-9757-2466d568e301


chrisbra:
thanks, that sounds interesting. However I think the option value should not be an opaque integer value `012` but rather something like:  `:set pumordering=word,kind,extra` 

Hopefully that is flexible enough and others don't want printf() formatting strings (or add arbitrary items to the pum menu).

kiddos:
I see that some of the computation can be done at the option set callback, do I need to move the computation to the callback?

chrisbra:
I don't think so. The option callback is for completing option values. Does that work?

chrisbra:
can you please fix the failing CI?

kiddos:
I might need to add some tests too.

chrisbra:
yes, please add tests

runtime/doc/options.txt
```diff
@@ -6273,6 +6273,23 @@ A jump table for the options with a short description can be found at |Q_op|.
 	Insert mode completion.  When zero as much space as available is used.
 	|ins-completion-menu|.
 
+						*'pumordering'* *'po'*
+'pumordering' 'po'	string (default "012")
```
chrisbra: this should be "word,kind,extra" ?

runtime/doc/options.txt
```diff
@@ -6273,6 +6273,23 @@ A jump table for the options with a short description can be found at |Q_op|.
...
+	This option specify how the popup menu should look like
+
+	The default value:
+	0: word
+	1: kind
+	2: extra
```
chrisbra: same here, default value is no longer 0,1,2

src/popupmenu.c
```diff
@@ -450,6 +510,27 @@ pum_redraw(void)
...
+    prefix_pum_widths[0] = pum_base_width;
+    prefix_pum_widths[1] = pum_kind_width;
+    prefix_pum_widths[2] = pum_extra_width;
+
+    pum_set_ordering(rounds);
+    // precompute widths
+    for (r = 0; r < 3; ++r)
+    {
+	if (rounds[r] == 0 && prefix_pum_widths[rounds[r]] > 0)
+	{
```
chrisbra: please remove braces around single-line statements

chrisbra:
can you add a few tests for:
- `set pumordering=` (empty)
- `set pumordering=word,kind`
- `set pumordering=word`
- `set pumordering=extra`
- `set pumordering=does-not-exists`

kiddos:
I would say the current behavior for this case `set pumordering=extra` is kinda weird.

I could make it so that `word`, and `kind` would also appear?

runtime/doc/options.txt
kiddos: yeah

runtime/doc/options.txt
kiddos: yea

src/popupmenu.c
kiddos: sure thing

src/testdir/test_popup.vim
```diff
@@ -1333,4 +1333,119 @@ func Test_pum_highlights_custom()
...
+    endfunc
+    set completeopt=menu
+    set completefunc=CompleteFunc
+    set pumordering=word,kind,extra
+  END
+  call writefile(lines, 'Xscript', 'D')
+  let buf = RunVimInTerminal('-S Xscript', {})
+  call TermWait(buf)
+  call term_sendkeys(buf, "iaw\<C-X>\<C-u>")
+  call TermWait(buf, 50)
```
zeertzjq, line 1358: `TermWait()` is not needed before `VerifyScreenDump()`

runtime/doc/options.txt
```diff
@@ -6273,6 +6273,20 @@ A jump table for the options with a short description can be found at |Q_op|.
...
+						*'pumordering'* *'po'*
+'pumordering' 'po'	string (default "word,kind,extra")
+	This option specify how the popup menu should look like
+
+	The default value: word,kind,extra
+
+	So the popup menu should look like the following
+	"word"  "kind"  "extra"
+
+	If this value is set to 102, the popup menu would look like this:
```
chrisbra: still reference to `102`

chrisbra:
> I would say the current behavior for this case `set pumordering=extra` is kinda weird.

What do you mean with it?

> I could make it so that `word`, and `kind` would also appear?

So that would basically mean, we would be able to re-order items, but cannot leave out items, right?

kiddos:
If user do this `set pumordering=extra`, only the `extra` will show up.
Yeah, I kind of want to make it so all items cannot be left out.

runtime/doc/options.txt
kiddos: ooops

chrisbra:
> If user do this set pumordering=extra, only the extra will show up.

Yes, it might be weird, and it sounds like a "don't complain if you use this setting and Vim does what you tell it. Don't do it then".

Not sure if this should be allowed to have  only 'extra' in 'pumordering'. Opinions?

PR #14007, opened 2024-02-10
Fix syntax test fails depending on LANG settings

h-east:
When `LANG=ja_JP.UTF-8`, the syntax test result will be NG unexpectedly.
So, I changed the following.

Fixed the following environment variables to "C" when executing syntax test.

```
LC_ALL
LANG
LANGUAGE
```

--
Best regards,
Hirohito Higashi (h_east)


h-east:
Wow! All tests are skipped in the local environment...
Wait a moment.

h-east:
Ah, After the second time, "make clean" was required.
No problem

chrisbra:
thanks!

chrisbra:
I found a few issues with the current setup, so I made some more adjustements on top of your patch. Please verify it works correctly for you.

h-east:
Works fine! Thanks.

PR #14008, opened 2024-02-10
'breakindent' behaves inconsistently with 'list' and splits

zeertzjq:
Problem:  'breakindent' behaves inconsistently with 'list' and splits.
Solution: Use 'listchars' from the correct window and handle caching
          properly. Move cheaper comparisons to the top.


chrisbra:
thanks!

PR #14009, opened 2024-02-10

dkearns:
- allow comments after :highight commands
- match the bang in a :highlight[!] command
- highlight the bang in :map[!], :menu[!] and :unlet[!] with vimOper
  like all other commands


chrisbra:
thanks!

PR #14010, opened 2024-02-10
diff() should not use 'diffexpr'

### LINKED ISSUE #13989: diff() causes an error if a function is set in &diffexpr
> This is a different question. It looks like that `diff()` calls a function which is set in &diffexpr. &diffopt says that `internal` is ignored when &diffexpr is set. It is not necessary for `diff()` to work along with those options, right?

This is a bug.  Can you open a separate issue for this?  I will open a PR to fix this.

_Originally posted by @yegappan in https://github.com/vim/vim/issues/12321#issuecomment-1932394837_

yegappan:
diff() should not use the 'diffexpr' option setting.
Add support for the unified diff context length.  Default is 0.
Sort the diff() option items in the help.

Fixes #13989.


chrisbra:
thanks!

PR #14013, opened 2024-02-11
Not able to use diff() with 'diffexpr'

yegappan:
Not able to use the diff() function with 'diffexpr'.  Set the default value for the context length to 0.
Update the diff() help text.  Add a test.


chrisbra:
Thanks!

```
- *diff_ctxlen = 1;
```

Ah I stumbled upon this yesterday and wondered, but thought it would be okay... I guess I was wrong

PR #14014, opened 2024-02-12
'breakindentopt' "min" works incorrectly with 'signcolumn'

zeertzjq:
Problem:  'breakindentopt' "min" works incorrectly with 'signcolumn'.
Solution: Use win_col_off() and win_col_off2().


chrisbra:
thanks!

PR #14015, opened 2024-02-12
settabvar() and settabwinvar() change last accessed tabpage

zeertzjq:
Problem:  settabvar() and settabwinvar() change last accessed tabpage.


zeertzjq:
The test and commit message are wrong. I'll open another PR later.

PR #14016, opened 2024-02-12
settabvar() may change last accessed tabpage

zeertzjq:
Problem:  The last accessed tabpage may change when using settabvar() or
          changing options using Python.


zeertzjq:
Hmm, still wrong.

PR #14017, opened 2024-02-12
settabvar() may change the last accessed tabpage

zeertzjq:
Problem:  settabvar() may change the last accessed tabpage.


chrisbra:
thanks. I fixed the style to match what we use normally in Vim however.

PR #14019, opened 2024-02-12
Redrawing can be improved with undo and 'spell'

zeertzjq:
Problem:  When undoing with 'spell', redrawWinline() is called after
          changed_lines(), while later win_update() sets redraw type to
          UPD_NOT_VALID, even though w_redraw_top and w_redraw_bot are
          still valid.
Solution: Only set redraw type to UPD_NOT_VALID when inserting/deleting
          lines after parts of window has pending redraw, i.e., when
          changed_lines() is called after redrawWinline().


chrisbra:
thanks!

Issue #14024, opened 2024-02-13
In a terminal Visual fg is invisible if `set termguicolors` set for default colorscheme


habamax:
### Steps to reproduce

1. `vim -Nu NONE`
2. `:set termguicolors`
3. add some text and select it

![image](https://github.com/vim/vim/assets/234774/5b4e87ee-0443-4581-882a-0c47b193e703)

https://www.reddit.com/r/vim/comments/1apo5xb/is_that_normal_selection_in_vim/

### Expected behaviour

Text in Visual should be visible.

### Version of Vim

9.1.104

### Environment

debian, xterm/st

habamax:
I still think GUI Visual of default should stay the same, but for termguicolors + terminal we would need to set fg to black, @chrisbra what do you think?

habamax:
Or to simplify everything, just set it to black no matter GUI or not.

PR #14025, opened 2024-02-13
problem: Visual is unreadable if termguicolors is set

habamax:
solution: Set Visual guifg to black(light) and lightgrey(dark)

Relates https://github.com/vim/vim/issues/14024

### DIFF
--- a/src/highlight.c
+++ b/src/highlight.c
@@ -222,7 +222,7 @@ static char *(highlight_init_light[]) = {
 	 "SignColumn term=standout ctermbg=Grey ctermfg=DarkBlue guibg=Grey guifg=DarkBlue"),
 #endif
     CENT("Visual ctermbg=Grey ctermfg=Black",
-	 "Visual ctermbg=Grey ctermfg=Black guibg=LightGrey"),
+	 "Visual ctermbg=Grey ctermfg=Black guibg=LightGrey guifg=Black"),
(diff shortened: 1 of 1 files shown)


habamax:
`set tgc bg=light`

image

`set tgc bg=dark`

image

gui

image

gui and `hi normal guibg=black guifg=white`

image

habamax:
I don't know how to distinguish terminal + termguicolors vs pure gui so here `guifg` is set for both, which would might annoy ppl using default gvim.

chrisbra:
so what exactly is the problem here? I tried in an ordinary xterm `vim --clean -c ':set termguicolors'` and it looks okay: 
grafik

habamax:
Try with another terminal that will fail detecting dark background (urxvt, windows terminal, simple terminal, konsole, gnome-terminal, etc).

You either in putty or xterm (putty is forcefully set to use dark background, xterm properly reports bg color with OSC and vim can detect that bg is dark).

habamax:
Those are set to dark:

 * "linux"	    Linux console
 * "screen.linux"   Linux console with screen
 * "cygwin.*"	    Cygwin shell
 * "putty.*"	    Putty program

The rest would depend on `handle_osc` and `COLORFGBG` env variable.



https://github.com/vim/vim/blob/master/src/term.c#L3212-L3244

PR #14026, opened 2024-02-13
filetype: Add dtso files

scosu:
dtso files are devicetree overlay files which have the same syntax as dts or dtsi files.


chrisbra:
thanks. Can you please add a test?

scosu:
Done, I hope this is the test you meant.

chrisbra:
yes, thank you!

PR #14031, opened 2024-02-13
Tweak syntax/generator

h-east:
- Add missing "Last Change:" line.
  The date on this line in `vim.vim` is updated by `update_date.vim` at Make time.  (I made a mistake in the file path)
- Remove unnecessary "b:loaded_syntax_vim_ex".

Suggest:
- Is the "Base File Date:" line in `vim.vim.base` necessary?


runtime/syntax/vim.vim
```diff
@@ -34,7 +34,7 @@ syn keyword vimCommand contained foldd[oopen] folddoc[losed] foldo[pen] for fu[n
...
 syn keyword vimCommand contained stat[ic] stopi[nsert] stj[ump] sts[elect] sun[hide] sus[pend] sv[iew] sw[apname] synti[me] sync[bind] smi[le] t tN[ext] ta[g] tags tab tabc[lose] tabd[o] tabe[dit] tabf[ind] tabfir[st] tabm[ove] tabl[ast] tabn[ext] tabnew tabo[nly] tabp[revious] tabN[ext] tabr[ewind] tabs tc[d] tch[dir] tcl tcld[o] tclf[ile] te[aroff] ter[minal] tf[irst] th[row] thi[s] tj[ump] tl[ast] tm[enu] tn[ext] to[pleft] tp[revious] tr[ewind] try ts[elect] tu[nmenu] ty[pe] u[ndo] undoj[oin] undol[ist] una[bbreviate] unh[ide] unl[et] unlo[ckvar] uns[ilent] up[date] v[global] ve[rsion] verb[ose] vert[ical] vi[sual] vie[w] vim[grep] vimgrepa[dd] vim9[cmd] vim9s[cript] viu[sage] vne[w] vs[plit] w[rite] wN[ext] wa[ll] wh[ile] wi[nsize] winc[md] wind[o] winp[os]
-syn keyword vimCommand contained wn[ext] wp[revious] wq wqa[ll] wu[ndo] wv[iminfo] x[it] xa[ll] xr[estore] y[ank] z dl dell delel deletl deletel dp dep delp delep deletp deletep a a
+syn keyword vimCommand contained wn[ext] wp[revious] wq wqa[ll] wu[ndo] wv[iminfo] x[it] xa[ll] xr[estore] y[ank] z dl dell delel deletl deletel dp dep delp delep deletp deletep a i
```
dkearns, line 37: IUUC `a` and `i` should actually be excluded as they're correctly listed in the generator's exclusion list.  I've added `:change` to that list in  #14020 as well.

h-east: Yes. I don't think it's necessary, but if `a` is missing, the `vim_line_continuation` test will fail, so I have no choice but to add it (following the original).

dkearns: I'd just defer all that matching to the section `" Insertions And Appends: insert append`.  There's no point in supporting only the abbreviated form of the command (`a` but not `append`).  The original script is wrong for having included it and I noted it as a bug years ago.

The failure in the line continuation test isn't a problem as the :append command shouldn't be matching there anyway.

Alternatively, I can remove the special-casing when I finish https://github.com/vim/vim/pull/14020 if you prefer.

h-east: > Alternatively, I can remove the special-casing when I finish https://github.com/vim/vim/pull/14020 if you prefer.

Yes. Please :bow:

chrisbra:
thanks I include it. I made a few more minor adjustments and I also add Doug as co-maintainer since he is doing a lot of vim syntax script adjustments. Hope that is fine.

h-east:
No problem!  Thanks.

PR #13943, opened 2024-01-31
Run tests on M1 Mac on GitHub Actions

rhysd:
## Problem

M1 Mac is not tested on CI

## Solution

Add CI jobs to run tests on M1 Mac on GitHub Actions

## Details

Apple is migrating their CPU architecture from x86_64 to arm64 (Apple Silicon, M1 chip). So arm64 will replace x86_64 and it is important to test Vim on M1 Mac for the stability of macOS support.

Recently GitHub Actions [introduced M1 Mac runner](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/). This PR adds the first jobs to run tests with tiny/normal/huge Vim on M1 Mac.

Some test cases are skipped because I could not fix them. They are mysteriously failing on M1 Mac only.

- `Test_term_gettitle` : Title got via `term_gettitle` was corrupted.
  ```
  From test_terminal2.vim:
  Found errors in Test_term_gettitle():
  Run 1, 06:50:44 - 06:50:49:
  command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_term_gettitle[14]..WaitForAssert[2]..<SNR>4_WaitForCommon[11]..<lambda>14 line 1: Pattern '^\\[No Name\\] - VIM\\d*$' does not match 'e] - VIM'
  Run 2, 06:50:51 - 06:50:57:
  command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[642]..function RunTheTest[57]..Test_term_gettitle[14]..WaitForAssert[2]..<SNR>4_WaitForCommon[11]..<lambda>17 line 1: Pattern '^\\[No Name\\] - VIM\\d*$' does not match 'e] - VIM'
  Run 3, 06:50:59 - 06:51:04:
  command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[642]..function RunTheTest[57]..Test_term_gettitle[14]..WaitForAssert[2]..<SNR>4_WaitForCommon[11]..<lambda>20 line 1: Pattern '^\\[No Name\\] - VIM\\d*$' does not match 'e] - VIM'
  Run 4, 06:51:06 - 06:51:11:
[...]

PR #14032, opened 2024-02-13
CI: Fix macos-14 tests

ychin:
macos-14 runner was turned on in #13943, but it had to turn off a few tests in order for CI to run. Re-enable them and fix the underlying issues.

* `Test_diff_screen`: The test failure is due to a bug in Apple's diff utility. Apple introduced a new diff tool based on FreeBSD in macOS 13 and it has buggy behaviors when using unified diff (`-U0`) and the diff is on the first line of the file. Simply disable this test for now if we detect Apple diff (instead of the old GNU diff). Can re-enable this in the future if Apple fixes the issue.
* `Test_play_event` / `Test_play_silent`: GitHub Actions currently has an issue with playing sound in CI in macos-14 runners. It for some reason triggers a microphone permission dialog popup which blocks the CI action (see https://github.com/actions/runner-images/issues/9330). To fix this, add a temporary step in macos-14 to manually allow microphone permissions in the runner.
* `Test_term_gettitle`: I could not reproduce the failure, so I just turned it on and it seems to run just fine. Maybe it's a timing issue and whatnot but either way that should be fixed when we can reproduce the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,17 +319,17 @@ jobs:
       - name: Checkout repository from github
         uses: actions/checkout@v4
 
-      - name: Install packages on Intel Mac
-        if: matrix.features == 'huge' && matrix.runner == 'macos-latest'
-        run: |
-          brew install lua
-          echo "LUA_PREFIX=/usr/local" >> $GITHUB_ENV
-
-      - name: Install packages on M1 Mac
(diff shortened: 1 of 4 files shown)


pheiduck:
Nice catch but ya, Apple is slow due updating there toolchain.

chrisbra:
thanks!

PR #14035, opened 2024-02-14
runtime(vim): Update syntax file - remove unused vimString regions

dkearns:
Remove unused vimString regions.

These were included with the initial release of the syntax file for Vim
5 and were probably intended to allow for syn-region start/skip/end
patterns with a '!' or '+' delimiter.  However, these cases are
currently handled by the vimSynRegPat group.

The removed patterns never match anywhere in the distributed runtime
files and it is believed that this is generally true.


chrisbra:
thanks!

PR #14039, opened 2024-02-14
filetype: no support for bats files

blmaier:
The '*.bats' file type is for Bash Automated Testing System (BATS) scripts. BATS scripts are Bash with a special '@test' extension but they otherwise work with Vim's bash filetype.

See https://github.com/bats-core/bats-core


chrisbra:
thanks!

PR #14042, opened 2024-02-15
Check for room when split-moving windows, fix some other related stuff

seandewar:
(Details are in the commit messages.)

On a somewhat related note, I noticed `WinNewPre` is fired even before `win_split_ins` checks for room or attempts to create any window (that is, `WinNewPre` may be fired, but not `WinNew` if E36 happens).

Is that intended? I guess it might be as it seems `win_split_ins` ~~may resize windows before giving E36~~ _actually it's only when failing to alloc a new window_, and it's documented as being "before commands that modify window layout".


chrisbra:
yeah, we may need to amend the documentation, that because of E36 WinNew may not be triggered.

seandewar:
Seems good :+1:

Glancing at the code, I can see a few other potential issues with how WinNewPre is currently handled (e.g: it's fired while curwin might be invalid and has no entry in the window list or frame, plus switching windows and tabpages in the autocmd could be problematic). I'll probably try to fix those things and the documentation in a different PR when I have the time.

chrisbra:
thanks. Let me know when it's ready to review and I'll have a closer look.

src/window.c
```diff
@@ -1900,42 +1918,59 @@ win_rotate(int upwards, int count)
...
 
     if (ONE_WINDOW)
-    {
-	beep_flush();
-	return;
-    }
-    if (check_split_disallowed() == FAIL)
-	return;
+	return OK;	// nothing to do
+    if (check_split_disallowed(wp) == FAIL)
```
seandewar, line 1935: This is consistent with how `win_totop` worked, but maybe it's not necessary to check the `wp->w_buffer->b_locked_split` part before split-moving, as it won't produce more views into the same buffer (which I think was the motivation for the flag in [v8.2.2476](https://github.com/vim/vim/commit/983d83ff1cd796ff321074335fa53fbe7ac45a46)). I haven't checked this, though.

chrisbra:
Thanks , looks good

zeertzjq:
> There are also other issues I found with the current implementation of WinNewPre
that need addressing:
>
> - it allows switching windows and tabpages, which can cause incorrect windows to
>   be split/moved, and big problems when switching tabpages.
>
> - it fires before win_split_ins checks for room, before it makes any changes to
>   window sizes or before it considers allocating a new window. This should be
>   changed or documented.

Vim 9.2 release is getting near. Do you still plan to fix these problems?

seandewar:
Ah, sorry! Likely not any time soon, as I'm lacking the motivation.
Feel free to take that over, if you're interested. :hugs:

zeertzjq:
I'm not sure about this. While these indeed are potential problems, they do not cause a crash, as `win_split_ins()` just splits the current window. Even if it switches to a window with a closing buffer, it won't lead to a crash either (as long as `b_nwindows` is correct) because of #19088.

seandewar:
Hmm, my memory is hazy, but I recall `trigger_winnewpre` also being called in `win_new_tabpage` (or somewhere close) at a pretty unsafe spot, which may be what 1. is referring to, but I can't find it now.

That said, the `trigger_winnewpre` in `win_split_ins` looks pretty innocent to me now. :thinking: (especially as it's not called when moving a window any more; maybe I was confused when I wrote that)

PR #14044, opened 2024-02-16
Setting some options changes curswant unnecessarily

zeertzjq:
Problem:  Setting some options changes curswant unnecessarily.
Solution: Add a P_HLONLY flag that prevents changing curswant.


chrisbra:
thanks!

PR #14045, opened 2024-02-16
when match original text disable trigger events

### LINKED ISSUE #679: TextChangedI gets triggered for i_CTRL-P without any changes
`b:changedtick` gets changed and triggers `TextChangedI` although there might be no changes:
1. vim -u NONE -N
2. `ifoo<Esc>`
3. `:echom b:changedtick`
   => 5
4. `a<Esc>`
5. `:echom b:changedtick`
   => 5  (OK)
6. `a<C-p>`
   => "-- Keyword completion (^N^P) Pattern not found"
7. `<Esc>:echom b:changedtick`
   => 11

Note that this is `len("foo") * 2`.

This is likely caused by Vim replacing the word itself with itself internally?!

This is a bit problematic in the context of completions, because it will not display the popup menu with completions, but might re-trigger e.g. deoplete (which hooks into TextChangedI).

I could not reproduce it using the following btw:

```
vim -u NONE --cmd 'norm ifoo' --cmd 'echom b:changedtick' --cmd 'call feedkeys("a\<C-p>\<Esc>")' --cmd 'echom b:changedtick'
```

If this cannot be fixed internally to not change `b:changedtick`, it should at least not trigger `TextChangedI` / `TextChanged` probably.

/cc @Shougo (because I've noticed this in the context of deoplete; Neovim is affected as well)

glepnir:
Problem: 
when reinsert typed part shouldn't trigger some events like CursormovedI and TextchangedI

Solution:
restore before changedtick_i and last_cursormoved avoid trigger events.

Fix https://github.com/vim/vim/issues/679

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3991,15 +3991,23 @@ ins_compl_delete(void)
 ins_compl_insert(int in_compl_func)
 {
     int compl_len = get_compl_len();
+    varnumber_T curr_changedtick_i = curbuf->b_last_changedtick_i;
+    int matched_original = match_at_original_text(compl_shown_match);
 
     // Make sure we don't go over the end of the string, this can happen with
     // illegal bytes.
     if (compl_len < (int)STRLEN(compl_shown_match->cp_str))
 	ins_bytes(compl_shown_match->cp_str + compl_len);
-    if (match_at_original_text(compl_shown_match))
-	compl_used_match = FALSE;
-    else
-	compl_used_match = TRUE;
+    compl_used_match = matched_original ? FALSE : TRUE;
+    // if matched original text shouldn't trigger events like TextchangedI and CursorMovedI
+    if (matched_original) {
+      // restore before changedticki avoid trigger TextChangedI
+      curbuf->b_last_changedtick_i = curr_changedtick_i;
+      // avoid trigger CursorMovedI
+      if (!EQUAL_POS(last_cursormoved, curwin->w_cursor)) {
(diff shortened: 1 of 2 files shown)


glepnir:
emm test case failed and what i confused is in https://github.com/vim/vim/blob/79230f027a25ff12eb7c7b64e1c063297876aae2/src/insexpand.c#L3963-L3969

the comment said `In insert mode: delete the typed part` . if nothing we want insert from completion. why there  need delete the typed part ?   this part commit in `071d4279d6ab81b7187b48f3a0fc61e587b6db6c` 2004 ...just try refactor here avoid insert or delete typed part text when there is no matches and nothing selected `pum_selected_item -1 ` .. some test case failed. not sure this is correct way .

zeertzjq:
I don't think this is correct. It's possible to select and insert a completion item using `<C-N>`, then selection the original text with changing the buffer text using `<Down>`, and using `<C-Y>` to change the buffer text back to the original text, in which case `TextChangedI` needs to be triggered.

chrisbra:
I would also think TextChangedI should be triggered regardless, because the text changed (even so it was back to the original text)

glepnir:
yup there is a point. if nothing selected the `ins_compl_delete` still delte the typed part and then `ins_compl_insert` reinsert typed part , `changedtick` changed in these functions . 

what i think is when we don't insert any completion item we should not delete the typed part in insert mode to avoid any changes. Does this case still require the current row to be marked a flag ?

PR #14046, opened 2024-02-16
runtime(spec): Recognize SourceLicense tag name in RPM specs

ppisar:
rpm-4.19.0 added a new SourceLicense tag. It is used at the same place as License tag.

This patch adds the new tag name into a Vim syntax file to be highligted the same way as the License tag. Note that it has to be defined in the syntax file before Source\d* regexp. Otherwise it's not recognized by Vim.


chrisbra:
this looks fairly safe and is mentioned here: https://rpm-software-management.github.io/rpm/manual/spec.html 

So let me include it.

PR #14048, opened 2024-02-17
Add python f-string and string-prefix highlight

### LINKED ISSUE #14033: python syntax file, support f-string
Vim syntax file for Python, with support for f-string added
- f-string support
- Fixed shortstring with multiple lines still highlighted
- Update string prefix to Python3.10+ syntax

have contacted the maintainer zpetkovic@acm.org But there seems to be no response

**python.vim.patch**
```diff
--- /usr/share/vim/vim91/syntax/python.vim	2024-01-03 12:24:01.000000000 +0800
+++ python.vim	2024-02-14 12:03:07.910000000 +0800
@@ -143,17 +143,35 @@
 
 " Triple-quoted strings can contain doctests.
 syn region  pythonString matchgroup=pythonQuotes
-      \ start=+[uU]\=\z(['"]\)+ end="\z1" skip="\\\\\|\\\z1"
+      \ start=+[uUbB]\=\z(['"]\)+ end="$\|\z1" skip="\\\\\|\\\z1"
       \ contains=pythonEscape,@Spell
 syn region  pythonString matchgroup=pythonTripleQuotes
-      \ start=+[uU]\=\z('''\|"""\)+ end="\z1" keepend
+      \ start=+[uUbB]\=\z('''\|"""\)+ end="\z1" keepend
       \ contains=pythonEscape,pythonSpaceError,pythonDoctest,@Spell
 syn region  pythonRawString matchgroup=pythonQuotes
-      \ start=+[uU]\=[rR]\z(['"]\)+ end="\z1" skip="\\\\\|\\\z1"
+      \ start=+\%([bB][rR]\=\|[rR][bB]\=\)\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\r\=$\|\\\z1"
       \ contains=@Spell
 syn region  pythonRawString matchgroup=pythonTripleQuotes
-      \ start=+[uU]\=[rR]\z('''\|"""\)+ end="\z1" keepend
+      \ start=+\%([bB][rR]\=\|[rR][bB]\=\)\z('''\|"""\)+
+      \ end="\z1" keepend
       \ contains=pythonSpaceError,pythonDoctest,@Spell
+syn region  pythonFString matchgroup=pythonQuotes
+      \ start=+[fF]\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\z1"
+      \ contains=pythonEscape,pythonFStringEscapedBrace,pythonFStringValue,@Spell
+syn region  pythonFString matchgroup=pythonTripleQuotes
+      \ start=+[fF]\z('''\|"""\)+
+      \ end="\z1" keepend
+      \ contains=pythonEscape,pythonSpaceError,pythonFStringEscapedBrace,pythonFStringValue,pythonDoctest,@Spell
+syn region  pythonRawFString matchgroup=pythonQuotes
+      \ start=+\%([fF][rR]\|[rR][fF]\)\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\r\=$\|\\\z1"
+      \ contains=pythonFStringEscapedBrace,pythonFStringValue,@Spell
+syn region  pythonRawFString matchgroup=pythonTripleQuotes
+      \ start=+\%([fF][rR]\|[rR][fF]\)\z('''\|"""\)+
+      \ end="\z1" keepend
+      \ contains=pythonSpaceError,pythonFStringEscapedBrace,pythonFStringValue,pythonDoctest,@Spell
 
 syn match   pythonEscape	+\\[abfnrtv'"\\]+ contained
 syn match   pythonEscape	"\\\o\{1,3}" contained
@@ -161,7 +179,11 @@
 syn match   pythonEscape	"\%(\\u\x\{4}\|\\U\x\{8}\)" contained
 " Python allows case-insensitive Unicode IDs: http://www.unicode.org/charts/
 syn match   pythonEscape	"\\N{\a\+\%(\s\a\+\)*}" contained
-syn match   pythonEscape	"\\$"
+syn match   pythonEscape	"\\\r\=$"
+
+syn region  pythonFStringValue matchgroup=pythonFStringBrace
+      \ start=+{+ end=+=\=\%(![rsa]\)\=\%(:[^{}]*\%({[^{}]*\%({[^{}]*}[^{}]*\)*}[^{}]*\)*\)\=}+ contains=TOP contained excludenl
+syn match   pythonFStringEscapedBrace	"{{\|}}" contained
 
 " It is very important to understand all details before changing the
 " regular expressions below or their order.
@@ -312,9 +334,14 @@
 hi def link pythonTodo			Todo
 hi def link pythonString		String
 hi def link pythonRawString		String
+hi def link pythonFString		String
+hi def link pythonRawFString		String
 hi def link pythonQuotes		String
 hi def link pythonTripleQuotes		pythonQuotes
 hi def link pythonEscape		Special
+hi def link pythonFStringEscapedBrace	Special
+hi def link pythonFStringValue		Normal
+hi def link pythonFStringBrace		Include
 if !exists("python_no_number_highlight")
   hi def link pythonNumber		Number
 endif
```

A4-Tacks:
close #14033

### DIFF
--- a/runtime/syntax/python.vim
+++ b/runtime/syntax/python.vim
@@ -143,17 +143,35 @@ syn keyword pythonTodo		FIXME NOTE NOTES TODO XXX contained
 
 " Triple-quoted strings can contain doctests.
 syn region  pythonString matchgroup=pythonQuotes
-      \ start=+[uU]\=\z(['"]\)+ end="\z1" skip="\\\\\|\\\z1"
+      \ start=+[uUbB]\=\z(['"]\)+ end="$\|\z1" skip="\\\\\|\\\z1"
       \ contains=pythonEscape,@Spell
 syn region  pythonString matchgroup=pythonTripleQuotes
-      \ start=+[uU]\=\z('''\|"""\)+ end="\z1" keepend
+      \ start=+[uUbB]\=\z('''\|"""\)+ end="\z1" keepend
       \ contains=pythonEscape,pythonSpaceError,pythonDoctest,@Spell
 syn region  pythonRawString matchgroup=pythonQuotes
-      \ start=+[uU]\=[rR]\z(['"]\)+ end="\z1" skip="\\\\\|\\\z1"
+      \ start=+\%([bB][rR]\=\|[rR][bB]\=\)\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\r\=$\|\\\z1"
       \ contains=@Spell
 syn region  pythonRawString matchgroup=pythonTripleQuotes
-      \ start=+[uU]\=[rR]\z('''\|"""\)+ end="\z1" keepend
+      \ start=+\%([bB][rR]\=\|[rR][bB]\=\)\z('''\|"""\)+
+      \ end="\z1" keepend
       \ contains=pythonSpaceError,pythonDoctest,@Spell
+syn region  pythonFString matchgroup=pythonQuotes
+      \ start=+[fF]\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\z1"
+      \ contains=pythonEscape,pythonFStringEscapedBrace,pythonFStringValue,@Spell
+syn region  pythonFString matchgroup=pythonTripleQuotes
+      \ start=+[fF]\z('''\|"""\)+
+      \ end="\z1" keepend
+      \ contains=pythonEscape,pythonSpaceError,pythonFStringEscapedBrace,pythonFStringValue,pythonDoctest,@Spell
+syn region  pythonRawFString matchgroup=pythonQuotes
+      \ start=+\%([fF][rR]\|[rR][fF]\)\z(['"]\)+
+      \ end="$\|\z1" skip="\\\\\|\\\r\=$\|\\\z1"
+      \ contains=pythonFStringEscapedBrace,pythonFStringValue,@Spell
+syn region  pythonRawFString matchgroup=pythonTripleQuotes
+      \ start=+\%([fF][rR]\|[rR][fF]\)\z('''\|"""\)+
+      \ end="\z1" keepend
+      \ contains=pythonSpaceError,pythonFStringEscapedBrace,pythonFStringValue,pythonDoctest,@Spell
 
 syn match   pythonEscape	+\\[abfnrtv'"\\]+ contained
 syn match   pythonEscape	"\\\o\{1,3}" contained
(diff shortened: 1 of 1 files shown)


A4-Tacks:
> Following up on comments in #14033
> 
> > > You can add the `pythonFStringValue` group to its own `contains` list to simplify matching these nested replacement fields. Then the end pattern can probably be reduced to a closing brace.
> > 
> > 
> > I cannot add it to its own contains list, it should only take effect on the right side of the colon, and the expression needs to be highlighted on the left side
> 
> I don't know, but I suspect that the likelihood of collection literal matches in the expression is much lower than computed format specs. I found some of the latter and none of the former in a quick search. Simply containing it recursively would probably only result in highlighted braces in the expression, in the worst case, which might be worth living with.
> 
> I just skimmed the PEP (not a Python user, take with a grain of salt) and it seems that this could be matched properly with a bit of effort.
> 
> You could create a series of regions contained in `pythonFStringValue` for the expression, conversion and format specification and have the format specification contain `pythonFStringValue`. It would probably be simplest to anchor the expression region to the start of the `pythonFStringValue` region with a look-behind and match them sequentially with `:syn-nextgroup`.
> 
> Unfortunately, the syntax file doesn't match other brace delimited syntax but those could be added and mixed with `:syn-skip` patterns to ignore parenthesised `!` and `:`.
> 
> I'm not suggesting that this _should_ be done, just that it's possible.

I don't want incorrect highlighting like `f'{exec("if 1:{}")}'`, and if I want to contains it myself, I won't be able to use TOP because it can only be excluded

runtime/syntax/python.vim
```diff
@@ -143,25 +143,47 @@ syn keyword pythonTodo		FIXME NOTE NOTES TODO XXX contained
...
 syn match   pythonEscape	"\\\o\{1,3}" contained
 syn match   pythonEscape	"\\x\x\{2}" contained
 syn match   pythonEscape	"\%(\\u\x\{4}\|\\U\x\{8}\)" contained
 " Python allows case-insensitive Unicode IDs: http://www.unicode.org/charts/
 syn match   pythonEscape	"\\N{\a\+\%(\s\a\+\)*}" contained
-syn match   pythonEscape	"\\$"
+syn match   pythonEscape	"\\\r\=$"
+
+syn region  pythonFStringValue matchgroup=pythonFStringBrace
+      \ start=+{+ end=+=\=\%(![rsa]\)\=\%(:[^{}]*\%({[^{}]*\%({[^{}]*}[^{}]*\)*}[^{}]*\)*\)\=}+ contains=TOP contained excludenl
```
dkearns, line 185: `excludenl` shouldn't be needed here as the pattern doesn't match EOL.

runtime/syntax/python.vim
```diff
@@ -312,9 +334,14 @@ hi def link pythonComment		Comment
 hi def link pythonTodo			Todo
 hi def link pythonString		String
 hi def link pythonRawString		String
+hi def link pythonFString		String
+hi def link pythonRawFString		String
 hi def link pythonQuotes		String
 hi def link pythonTripleQuotes		pythonQuotes
 hi def link pythonEscape		Special
+hi def link pythonFStringEscapedBrace	Special
+hi def link pythonFStringValue		Normal
```
dkearns, line 343: This isn't usually specified as it is the default if no highlight group is linked but does no harm.

dkearns (changes requested):
> I don't want incorrect highlighting like `f'{exec("if 1:{}")}'`, and if I want to contains it myself, I won't be able to use TOP because it can only be excluded

None of my suggestions would highlight that example incorrectly.  The `exec` argument would just be matched as a plain string.

`TOP` is just a convenience.  It would be nice if the `contains` syntax was a bit more expressive to allow groups to be added as well but you can define a new cluster including only the required expression groups.  Technically, of course, `TOP` matches too much anyway.

chrisbra:
closing this one and awaiting a new patch from the maintainer.

PR #14050, opened 2024-02-17
vim.vim: Distinguish Vim9 constructor definitions from the _new_ command

zzzyxwvut:
With the arrival of Vim9 classes, the syntax must allow for  
_new_ constructors; multiple constructor definitions are  
supported for a class, provided distinct suffix-names are  
used.  Currently, the defined constructors match either  
`vimCommand` or `vimFunctionError` (for any `newBar`).

For example:
```vim
vim9script

class Foo
    def new()
    enddef

    def newBar()
    enddef
endclass
```

Since every constructor is required to bear a lower-cased  
_new_ prefix name, it should suffice to distinguish them  
from functions, and so there are no new highlight or syntax  
groups introduced.

(Similar proposed changes to `syntax/vim.vim` have also been
emailed to Charles E. Campbell on October 24th.)


chrisbra:
@dkearns , @h-east ? I suppose this is fine?

runtime/syntax/generator/gen_syntax_vim.vim
```diff
@@ -125,6 +125,19 @@ function! s:parse_vim_command(cmd)
...
+		" instances can be told apart by resorting to rules of lower
+		" `:syn-priority` (manually addressed elsewhere).
+		if has_key(lcmd, 'n')
+			let list = lcmd['n']
+			let idx = index(list, 'new')
+			if idx >= 0
+				call remove(list, idx)
+			endif
+		endif
+
```
dkearns: There's an exclusion list that achieves this and you can just add `new` to that.

https://github.com/vim/vim/blob/fcaed6a70faf73bff3e5405ada556d726024f866/runtime/syntax/generator/gen_syntax_vim.vim#L275-L284

runtime/syntax/generator/vim.vim.base
```diff
@@ -31,6 +31,8 @@ syn cluster vimCommentGroup	contains=vimTodo,@Spell
 
 syn keyword vimCommand contained	2mat[ch] 3mat[ch]
 
+" Lower priority for _new_ to distinguish constructors from the command.
+syn match   vimCommand contained	"\<new\%(\i\{-}(\)\@!"
```
dkearns: I'd suggest the following patttern. The word boundary means you don't need to bother testing for `\i*`.

`\<new\>(\@!`

runtime/syntax/generator/vim.vim.base
```diff
@@ -510,7 +513,7 @@ syn match vimUsrCmd	'^\s*\zs\u\%(\w*\)\@>\%([(#[]\|\s\+\%([-+*/%]\=\|\.\.\)=\)\@
 " Errors And Warnings: {{{2
 " ====================
 if !exists("g:vimsyn_noerror") && !exists("g:vimsyn_novimfunctionerror")
- syn match	vimFunctionError	"\s\zs[a-z0-9]\i\{-}\ze\s*("			contained contains=vimFuncKey,vimFuncBlank
+ syn match	vimFunctionError	"\s\zs\%(new\i\{-}(\)\@![a-z0-9]\i\{-}\ze\s*("			contained contains=vimFuncKey,vimFuncBlank
```
dkearns: I think this can be simplified to `\s\zs\%(new\)\@![a-z0-9]\i\{-}\ze\s*(`

It would be a good idea to add a comment noting that the new-prefix exception should eventually only apply to Vim9 constructor definitions.  I'll try and finish #13810 this week, then it can be restricted to constructors.

runtime/syntax/testdir/input/vim_new.vim
```diff
@@ -0,0 +1,14 @@
+vim9script
+
+# Vim :new command and class constructors.
+class Test
+	def new()
+	enddef
+	def newOther()
+	enddef
```
dkearns, line 8: Could we add a test for an all lowercase named constructor?

dkearns (changes requested):
Thanks for the comprehensive PR.

runtime/syntax/generator/gen_syntax_vim.vim
h-east: There is an exclusion list for vim commands in the `exclude_list` inside the `s:get_vim_command_type()` functions.
It is better to register 'new' there.

```diff
diff --git a/runtime/syntax/generator/gen_syntax_vim.vim b/runtime/syntax/generator/gen_syntax_vim.vim
index 16b26b172..b4bfcd32a 100644
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -126,18 +126,6 @@ function! s:parse_vim_command(cmd)
 			endif
 		endfor
 
-		" Rationale: Preclude generating `:syn-keyword` for matching
-		" the same borne prefix `:new` and `:def new(` so that their
-		" instances can be told apart by resorting to rules of lower
-		" `:syn-priority` (manually addressed elsewhere).
-		if has_key(lcmd, 'n')
-			let list = lcmd['n']
-			let idx = index(list, 'new')
-			if idx >= 0
-				call remove(list, idx)
-			endif
-		endif
-
 		quit!
 
 		for key in sort(keys(lcmd))
@@ -294,6 +282,7 @@ function! s:get_vim_command_type(cmd_name)
 	\	'behave', 'augroup', 'normal', 'syntax',
 	\	'append', 'insert',
 	\	'Next', 'Print', 'X',
+	\	'new',
 	\ ]
 	" Required for original behavior
 	" \	'global', 'vglobal'
```

h-east: Ah, I duplicated what @dkearns pointed out.

zzzyxwvut:
The suggested corrections have been applied, please review,  
@chrisbra, @dkearns, @h-east .

runtime/syntax/generator/gen_syntax_vim.vim
zzzyxwvut: Oops, I've looked for a blacklist in the wrong place.

chrisbra:
thanks!

runtime/syntax/generator/gen_syntax_vim.vim
zzzyxwvut: In hindsight, it was a wrong approach to propose to remove  
a command name (`new`) from a command-name prefix dictionary;  
command names _later_ play a crucial role in looking up their  
own prefix abbreviations, if there are any.  Hypothetically,  
if `:newtonsmethod` were added, it would abbreviate to `new`!

Although, I wonder, if the definition of `exclude_list` would 
be better off when moved into its only caller, and evolved   
from a string list into a prefix dictionary that closely  
follows the structure of the command-name prefix dictionary  
`lcmd`.

I've profiled several times the current implementation and  
the following prototype and it shows a solid 10%+ reduction  
in execution time for `s:get_vim_command_type()`:

```diff
diff --git a/runtime/syntax/generator/Makefile b/runtime/syntax/generator/Makefile
index 3cf4b953d..657827319 100644
--- a/runtime/syntax/generator/Makefile
+++ b/runtime/syntax/generator/Makefile
@@ -21,7 +21,7 @@ vim.vim: vim.vim.rc update_date.vim
 vim.vim.rc: gen_syntax_vim.vim vim.vim.base $(SRC)
 	@echo "Generating vim.vim.rc ..."
 	@rm -f sanity_check.err generator.err
-	@$(RUN_VIM) -S gen_syntax_vim.vim
+	@$(RUN_VIM) '+profile start /tmp/bl.vim' '+profile func *get_vim_command_type' -S gen_syntax_vim.vim
 	@if test -f sanity_check.err ; then \
 		echo ; \
 		echo "Sanity errors:" ; \
diff --git a/runtime/syntax/generator/gen_syntax_vim.vim b/runtime/syntax/generator/gen_syntax_vim.vim
index 44f701613..c8881814f 100644
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -127,7 +127,26 @@ function! s:parse_vim_command(cmd)
 		endfor
 		quit!
 
+		" Keys map as 1:1 to the matching keys of l:lcmd.
+		let exclude_dict = {
+		\	'a': ['append', 'augroup', 'autocmd'],
+		\	'b': ['behave'],
+		\	'd': ['doautoall', 'doautocmd'],
+		\	'e': ['echo', 'echohl', 'execute'],
+		\	'i': ['insert'],
+		\	'm': ['map'],
+		\	'n': ['new', 'normal'],
+		\	's': ['set', 'setglobal', 'setlocal', 'smagic',
+		\		'snomagic', 'substitute', 'syntax'],
+		\	'v': ['var'],
+		\	'~': ['Next', 'Print', 'X'],
+		\ }
+		" Required for original behavior
+"		\	'g': ['global'],
+"		\	'v': ['var', 'vglobal'],
+
 		for key in sort(keys(lcmd))
+			let blacklist = get(exclude_dict, key, [])
 			for my in range(len(lcmd[key]))
 				let omit_idx = 0
 				if my > 0
@@ -174,7 +193,7 @@ function! s:parse_vim_command(cmd)
 				endif
 
 				let item.name = lcmd[key][my]
-				let item.type = s:get_vim_command_type(item.name)
+				let item.type = s:get_vim_command_type(item.name, blacklist)
 				if omit_idx + 1 < strlen(item.name)
 					let item.omit_idx = omit_idx
 					let item.syn_str = item.name[:omit_idx] . '[' . 
@@ -231,7 +250,7 @@ function! s:parse_vim_command(cmd)
 		for i in ['l', 'p']
 			let str = 'delete'
 			let item.name = str . i
-			let item.type = s:get_vim_command_type(item.name)
+			let item.type = s:get_vim_command_type(item.name, [])
 			let item.omit_idx = -1
 			for x in range(strlen(str))
 				let item.syn_str = str[:x] . i
@@ -260,7 +279,7 @@ function! s:parse_vim_command(cmd)
 	endtry
 endfunc
 
-function! s:get_vim_command_type(cmd_name)
+function! s:get_vim_command_type(cmd_name, exclude_list)
 	" Return value:
 	"   0: normal
 	"   1: (Reserved)
@@ -272,21 +291,8 @@ function! s:get_vim_command_type(cmd_name)
 	"   99: (Exclude registration of "syn keyword")
 	let menu_prefix = '^\%([acinosvx]\?\|tl\)'
 	let map_prefix  = '^[acilnostvx]\?'
-	let exclude_list = [
-	\	'map',
-	\	'substitute', 'smagic', 'snomagic',
-	\	'setlocal', 'setglobal', 'set', 'var',
-	\	'autocmd', 'doautocmd', 'doautoall',
-	\	'echo', 'echohl', 'execute',
-	\	'behave', 'augroup', 'normal', 'syntax',
-	\	'append', 'insert',
-	\	'Next', 'Print', 'X',
-	\	'new',
-	\ ]
-	" Required for original behavior
-	" \	'global', 'vglobal'
-
-	if index(exclude_list, a:cmd_name) != -1
+
+	if index(a:exclude_list, a:cmd_name) != -1
 		let ret = 99
 	elseif a:cmd_name =~# '^\%(abbreviate\|noreabbrev\|\l\%(nore\)\?abbrev\)$'
 		let ret = 2

```

Maybe, it's worth looking into.

h-east: Thank you for your feedback.

> In hindsight, it was a wrong approach to propose to remove a command name (`new`) from a command-name prefix dictionary; command names _later_ play a crucial role in looking up their own prefix abbreviations, 

Hmm, I don't think so.

> if there are any. Hypothetically, if `:newtonsmethod` were added, it would abbreviate to `new`!

No. It's "newt".

> 
> Although, I wonder, if the definition of `exclude_list` would be better off when moved into its only caller, and evolved from a string list into a prefix dictionary that closely follows the structure of the command-name prefix dictionary `lcmd`.
> 
> I've profiled several times the current implementation and the following prototype and it shows a solid 10%+ reduction in execution time for `s:get_vim_command_type()`:

`gen_syntax_vim.vim` was written 8 years ago. No improvements have been made in terms of execution speed.
However, the execution time is only a few seconds.
I think that's enough.

For now, I'd like to limit the changes to those that are truly necessary.

zzzyxwvut: Suppose, a command `:newtonsmethod` will have been introduced,  
and in compliance with:

https://github.com/vim/vim/blob/2a65e739447949a7aee966ce8a3b75521b2a79ea/src/ex_cmds.h#L93-L101

it will have been placed somewhere _after_ `:new`, for we want  
for `:new` to continue to abbreviate to itself.

(For demonstration, I don't patch `src/ex_cmds.h`, but rather  
put `newtonsmethod` into the command-name prefix dictionary,  
which is assembled in order of parsed lines read.)

With the corrected and accepted patch, we would have:

```vim
diff --git a/runtime/syntax/generator/gen_syntax_vim.vim b/runtime/syntax/generator/gen_syntax_vim.vim
index bc561dd6c..35a0b9fd7 100644
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -127,6 +127,8 @@ function! s:parse_vim_command(cmd)
                endfor
                quit!

+call insert(lcmd['n'], 'newtonsmethod', (index(lcmd['n'], 'new') + 1))
+
                for key in sort(keys(lcmd))
                        for my in range(len(lcmd[key]))
                                let omit_idx = 0
```

and the following `n*` commands generated:

```vim
n[ext] nb[key] nbc[lose] nbs[tart] newt[onsmethod] noa[utocmd] noh[lsearch] nos[wapfile] nu[mber]
```

With the challenged and abandoned patch, we would have:

```vim
diff --git a/runtime/syntax/generator/gen_syntax_vim.vim b/runtime/syntax/generator/gen_syntax_vim.vim
index bc561dd6c..df2ee5b27 100644
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -127,6 +127,9 @@ function! s:parse_vim_command(cmd)
                endfor
                quit!

+call insert(lcmd['n'], 'newtonsmethod', (index(lcmd['n'], 'new') + 1))
+call remove(lcmd['n'], index(lcmd['n'], 'new'))
+
                for key in sort(keys(lcmd))
                        for my in range(len(lcmd[key]))
                                let omit_idx = 0
@@ -281,7 +284,6 @@ function! s:get_vim_command_type(cmd_name)
        \       'behave', 'augroup', 'normal', 'syntax',
        \       'append', 'insert',
        \       'Next', 'Print', 'X',
-       \       'new',
        \ ]
        " Required for original behavior
        " \     'global', 'vglobal'
```

and the following `n*` commands generated:

```vim
n[ext] nb[key] nbc[lose] nbs[tart] new[tonsmethod] noa[utocmd] noh[lsearch] nos[wapfile] nu[mber]
```

So the problem with removals, as opposed to blacklisting, is  
that current and new command names _may_ claim abbreviations  
belonging to removed command names.

h-east: Your perception is wrong.
When add "newtonsmethod" after "new" in src/ex_cmds.h, the correct behavior is that `newt[onsmethod]` is added to "syn keyword vimCommand contained" in `vim.vim`.

The `exclude_list[]` in `gen_syntax_vim.vim` is **not** a list to remove Vim commands.
It is to avoid adding Vim commands to "syn keyword vimCommand contained" in `vim.vim`. The main reason is to avoid duplicating the specified Vim commands since they are defined in their own "syn keyword vimCommand contained" in `vim.vim.base`.


"runtime/syntax/*" is for controlling syntax highlighting. So it cannot actually kill Vim commands.

zzzyxwvut: Well, I don't dispute the correctness of your reviewed  
change:

```diff
diff --git a/runtime/syntax/generator/gen_syntax_vim.vim b/runtime/syntax/generator/gen_syntax_vim.vim
index 16b26b172..b4bfcd32a 100644
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -126,18 +126,6 @@ function! s:parse_vim_command(cmd)
 			endif
 		endfor
 
-		" Rationale: Preclude generating `:syn-keyword` for matching
-		" the same borne prefix `:new` and `:def new(` so that their
-		" instances can be told apart by resorting to rules of lower
-		" `:syn-priority` (manually addressed elsewhere).
-		if has_key(lcmd, 'n')
-			let list = lcmd['n']
-			let idx = index(list, 'new')
-			if idx >= 0
-				call remove(list, idx)
-			endif
-		endif
-
 		quit!
 
 		for key in sort(keys(lcmd))
@@ -294,6 +282,7 @@ function! s:get_vim_command_type(cmd_name)
 	\	'behave', 'augroup', 'normal', 'syntax',
 	\	'append', 'insert',
 	\	'Next', 'Print', 'X',
+	\	'new',
 	\ ]
 	" Required for original behavior
 	" \	'global', 'vglobal'

```

and with that change applied all what you have said above   
about `:newtonsmethod` is true.  Yes, `exclude_list[]` contains  
blacklisted command names for which no generation should be  
carried out.

All I'm trying to communicate, after the fact, is _what_ was  
algorithmically unsound with _my initial version_ of the patch  
to `gen_syntax_vim.vim`, after taking a closer look at the way  
abbreviations are formed for the parsed command names.  And  
the context of our discussion is well within two functions:  
`s:get_vim_command_type()` and `s:parse_vim_command()`.

PR #14052, opened 2024-02-18
Changes for STRLEN(), part 1

basilisk0315:
My first PR, redux.


chrisbra:
BTW: you can simply force push changes to your branch and don't need to close and re-open a new PR for that.

basilisk0315:
Hi Christian,

I am now down to 2 failures: one for macos-14 and the other is codecov/project.

How should I proceed from here?

Cheers
John

chrisbra:
the failing test in mac-os 14 is a known flaky one.

src/memline.c
```diff
@@ -1754,7 +1754,7 @@ ml_recover(int checkext)
 	for (idx = 1; idx <= lnum; ++idx)
 	{
 	    // Need to copy one line, fetching the other one may flush it.
-	    p = vim_strsave(ml_get(idx));
+	    p = vim_strnsave(ml_get(idx), ml_get_len(idx));
```
chrisbra, line 1757: is that an improvement? This will now do an `ml_get()` twice instead of once...

src/memline.c
```diff
@@ -2675,6 +2675,36 @@ ml_get_cursor(void)
...
+
+    return curbuf->b_ml.ml_line_len - 1;
+}
+
+// return length (excluding the NUL) of the cursor line
+    colnr_T
+ml_get_curline_len(void)
+{
+    if (*ml_get_buf(curbuf, curwin->w_cursor.lnum, FALSE) == NUL)
+        return 0;
```
chrisbra: I wonder if this can be tested

src/normal.c
```diff
@@ -4917,6 +4914,8 @@ nv_replace(cmdarg_T *cap)
 	}
 	else
 	{
+	    char_u *ptr = ml_get_cursor();
```
chrisbra: is it required to get the ptr here ?

chrisbra:
I think looks good already. However there are 2 changes I am slightly surprised. The one in `ml_recover()` I am not sure this is an improvement and the other in `nv_replace()` looks unrelated?

src/memline.c
basilisk0315: I think it is. The first call will find the lnum line and save it's info in curbuf->b_ml, and the second will use it. This will avoid calling STRLEN() on a potentially long string in vim_strsave().

src/normal.c
basilisk0315: No it's not. I brought it forward from the previous code.
Having said that, we need to make sure that ptr after the for loop because it is used unconditionally after the loop if netbeans_active() returns TRUE.

Here is a modification that might be better:
```
`	    char_u *ptr;

	    // Replace the characters within one line.
	    for (n = cap->count1; n > 0; --n)
	    {
		// Get ptr again, because ins_copychar() and showmatch()
		// will have released the line.
		// At the same time we let know that the line will be changed.
		if (cap->nchar == Ctrl_E || cap->nchar == Ctrl_Y)
		{
		  int c = ins_copychar(curwin->w_cursor.lnum
					   + (cap->nchar == Ctrl_Y ? -1 : 1));

		  ptr = ml_get_buf(curbuf, curwin->w_cursor.lnum, TRUE);
		  if (c != NUL)
		    ptr[curwin->w_cursor.col] = c;
		}
		else
		{
		    ptr = ml_get_buf(curbuf, curwin->w_cursor.lnum, TRUE);
		    ptr[curwin->w_cursor.col] = cap->nchar;
		}
		if (p_sm && msg_silent == 0)
		    showmatch(cap->nchar);
		++curwin->w_cursor.col;
	    }
#ifdef FEAT_NETBEANS_INTG
	    if (netbeans_active())
	    {
		colnr_T  start = (colnr_T)(curwin->w_cursor.col - cap->count1);
		ptr = ml_get_curline();

		netbeans_removed(curbuf, curwin->w_cursor.lnum, start,
							   cap->count1);
		netbeans_inserted(curbuf, curwin->w_cursor.lnum, start,
					       &ptr[start], (int)cap->count1);
	    }
#endif
```

src/memline.c
basilisk0315: I have not written a test before but I can give it a go. Any tips?

src/normal.c
chrisbra: thanks, I think this is better and closer to the current version anyhow.  However, unless I am missing something, you don't need the single declaration `char_u *ptr;`  at the start of the `else` condition. It would be shadowing the existing `ptr` variable.

chrisbra: ah, now I see it, you removed that `ptr` declaration at the beginning of the function, in which case this is fine. Thanks.

src/memline.c
```diff
@@ -2675,6 +2675,36 @@ ml_get_cursor(void)
...
+{
+    if (*ml_get_buf(curbuf, lnum, FALSE) == NUL)
+        return 0;
+
+    return curbuf->b_ml.ml_line_len - 1;
+}
+
+// return length (excluding the NUL) of the cursor line
+    colnr_T
+ml_get_curline_len(void)
```
chrisbra, line 2687: this could be instead:
```
   colnr_T
ml_get_curline_len(void)
{
    return ml_get_len(curwin->w_cursor.lnum);
}

basilisk0315: Good thought. See newest push.

chrisbra:
thanks, looks good now.

PR #14053, opened 2024-02-18
Using freed memory with full tag stack and user data

zeertzjq:
Problem:  Using freed memory with full tag stack and user data.
Solution: Clear the user data pointer of the newest entry.


zeertzjq:
Related: neovim/neovim#27498

chrisbra:
thanks, makes sense

PR #14055, opened 2024-02-18
runtime(filetype): Modula-2 files with priority not detected

dkearns:
Problem:  Modula-2 files with a specified priority are not detected.
Solution: Match the priority syntax in module header lines when
          performing heuristic content detection.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -3,7 +3,7 @@ vim9script
 # Vim functions for file type detection
 #
 # Maintainer:		The Vim Project <https://github.com/vim/vim>
-# Last Change:		2024 Jan 05
+# Last Change:		2024 Feb 18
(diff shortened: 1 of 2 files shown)


runtime/autoload/dist/ft.vim
```diff
@@ -1293,4 +1293,4 @@ export def FTvba()
 enddef
 
 # Uncomment this line to check for compilation errors early
-defcompile
+# defcompile
```
k-takata, line 1296: Is this intentional?

dkearns: Yes, it was accidentally left enabled in 68a8947.  I've added a comment to the commit message.

Thanks for paying attention.

chrisbra:
thanks!

KidiIT (commented):
All reviewed and approved🫶❤️❤️

PR #14060, opened 2024-02-20
Add a place holder section for version 9.2

yegappan:
Add a version 9.2 section to version9.txt.  As new features are added, this section should be updated.
This will make it easier to release the 9.2 version later.  Otherwise, we need to go through thousands
of commits to get this information before the 9.2 release.

### DIFF
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -5972,6 +5972,7 @@ added-7.4	version7.txt	/*added-7.4*
 added-8.1	version8.txt	/*added-8.1*
 added-8.2	version8.txt	/*added-8.2*
 added-9.1	version9.txt	/*added-9.1*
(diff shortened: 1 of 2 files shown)


habamax:
Should we have `Breaking changes` section as well?

one that comes to my mind is default colorscheme visual highlight now has defined foreground which breaks colorschemes that were relying on Visual `guifg=NONE`.

vim-ml:
Hi,

On Mon, Feb 19, 2024 at 6:52 PM Maxim Kim ***@***.***> wrote:

> Should we have Breaking changes section as well?
>
> one that comes to my mind is default colorscheme visual highlight now has
> defined foreground which breaks colorschemes that were relying on Visual
> guifg=NONE.
>
>
>
Yes.  I have updated the PR to include the "incompatible changes" section.

Regards,
Yegappan

PR #14061, opened 2024-02-20
runtime: Update syntax for tmux config

huangyxi:
Add the syntax of tmuxEnums: `keep-last` and `keep-group` for the tmux config file.

https://github.com/tmux/tmux/commit/40b97b1715182b0ab0274b0c402b34e2cba98f04

### DIFF
--- a/runtime/syntax/tmux.vim
+++ b/runtime/syntax/tmux.vim
@@ -164,7 +164,8 @@ syn keyword tmuxCommands
 syn keyword tmuxEnums
 \ absolute-centre all always any arrows bar blinking-bar blinking-block
 \ blinking-underline block both bottom centre color colour current default
-\ double emacs external failed heavy largest latest left manual next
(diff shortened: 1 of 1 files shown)


ericpruitt:
I opened https://github.com/vim/vim/pull/14065 which will incorporate the latest configuration options.

ericpruitt:
Also, @huangyxi , if you consistently recompile tmux or otherwise use the latest version, you can use a plugin manager to manage the tmux.vim syntax file as documented at <https://github.com/ericpruitt/tmux.vim/?tab=readme-ov-file#plugin-manager>. When new keywords get added to the tmux source code, the tmux.vim syntax file will generally be updated within 24 hours by a cron job running on my desktop.

huangyxi:
Thank @ericpruitt for the suggestion and for providing continuous maintenance! I'll definitely check out the plugin you mentioned.

PR #14063, opened 2024-02-20
Fix display of below/right virtual text with non-virtual text overlap

dylanahsmith:
## Problem

The following vim script demonstrates the problem

```vim
vim9script

setline(1, ['one', 'two', 'three', 'four'])

prop_type_add("demo_highlight_error", { highlight: 'Error' })

prop_add(2, 4, {
  type: 'demo_highlight_error',
  end_col: 4,
})
prop_add(2, 0, {
  type: 'demo_highlight_error',
  text: 'syntax error',
  text_align: 'below',
})
```

which after using with `vim --clean -S demo.vim` wouldn't show the virtual text property below line 2, it would instead show

```
one
two
three
four
```

where removing the zero-width non-virtual text property would show the expected result

```
one
two
Arbitrary error
three
four
```

with `Arbitrary error` highlighted with a red background.

The problem was that both those text properties would be active at the end of drawing the line (in function `win_line`), but the non-virtual text property would end up being chosen as `used_tpi` because it was given a higher priority in `sort_text_props`.

## Solution

I changed the comparison function for `sort_text_props` to always give virtual text properties a higher priority, to fix the above issue.  I also turned the above demonstration into a regression test.

When looking at the other uses of `sort_text_props` to see if there would be an unintended consequence of the change, and found the earlier use of `sort_text_props` in `win_line` was conditionally initializing the `text_prop_idxs` array with values that would be overwritten before they are read again.  That's because the rest of the function treats that array as if it has length `text_props_active`, which is `0` at the time of that early conditional initialization.  For example, loops like `for (pi = 0; pi < text_props_active; ++pi)` wouldn't read those initialized values and `text_prop_idxs[text_props_active++]` would later replace those values.

The other use of `sort_text_props` is in charset.c to calculate tab sizes, which skips over non-virtual text properties, so it doesn't look like this change should affect its behaviour.


chrisbra:

PR #14064, opened 2024-02-20
fix getregion() examples

habamax:
```
:xnoremap <CR>
\ <Cmd>echow getregion('v', '.', mode())<CR>
```
`echo` while in visual mode has no visible effect, thus people trying example might be frustrated as it looks like nothing happens.

So the option is to change it to `echow` or `echom`. With `echom` it is again has no visible effect but one can at least inspect `:messages`. On the other hand `echow` showes selected text in a popup window.

```
'.'->getregion("'a', 'v')
```

Here is the typo, which makes example invalid, should be `("'a", ...`


chrisbra:
Thanks. I noticed a few minor issues, can you please squash in the following patch on top?

```patch
diff --git a/src/evalfunc.c b/src/evalfunc.c
index 81f747f84..071bf7d0f 100644
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -5584,8 +5584,9 @@ f_getregion(typval_T *argvars, typval_T *rettv)

        getvvcol(curwin, &p1, &sc1, NULL, &ec1);
        getvvcol(curwin, &p2, &sc2, NULL, &ec2);
-       oap.motion_type = OP_NOP;
+       oap.motion_type = MBLOCK;
        oap.inclusive = TRUE;
+       oap.op_type = OP_NOP;
        oap.start = p1;
        oap.end = p2;
        oap.start_vcol = MIN(sc1, sc2);
```

And while your at it :) can you please re-generate `vim.vim`? 
```
cd runtime/syntax/generator && make && git add ../vim.vim
```

PR #14068, opened 2024-02-21
Fix infinite loop & signed overflow with 'smoothscroll' set

seandewar:
A fix for a random thing I noticed yesterday.

Also sneak in some missing changes I forgot to add to my previous PR (sorry!), and relocate Test_smoothscroll_in_zero_width_window to test_scroll_opt.vim, which feels more appropriate.

(Maybe it's possible for Vim to still draw some buffer text within the gaps of the textoff area when `cpo+=n` if the window isn't 0-width, but it seems Vim currently doesn't do that)


chrisbra:
thanks!

PR #14071, opened 2024-02-21
Don't force minlines in syntax/html.vim

Cheaterman:


Cheaterman:
Because it doesn't sync? I have to `:syn sync fromstart` all the time when editing HTML...

chrisbra:
> Because it doesn't sync? I have to :syn sync fromstart all the time when editing HTML...

Sounds more like you want a configuration variable that you can set for your specific use case. 

> How is 10 "not particularly bad" when the average HTML file is hundreds of lines?

I thought  you wanted to disable it for performance reasons. Guess what: there is a reason, why a PR contains a big field where you can and should enter comments :) 

Also, I fail to see, how `minlines` has any thing to do with hundreds of lines. It may still work to sync within 10 lines, right?

> How is it OK to not respect g:vimsyn_minlines

Because that is for `vim` syntax files?

As mentioned, we probably want to make this variable configurable rather than remove it

Cheaterman:
So make it configurable using g:vimsyn_minlines? Since that's a vim syntax file?

Cheaterman:
Before:

before

After:

after

The only thing I did between the two is remove the offending line, then `:e` to re-edit my file.

Cheaterman:
If anything what I want to do locally is increase it, not decrease it. But the static 10 in the syntax file prevents that... so clearly it has to go, I don't even know why it's forced that way instead of relying on user settings?

chrisbra:
No, add a new variable, let's say: `g:htmlsyn_minlines` , mention it in the documentation at `:h ft-html-syntax`  and refernce it then in the html syntax file, similar to how it is done in the  [vim syntax script](https://github.com/vim/vim/blob/3f905ab3c4f66562f4a224bf00f49d98a0b0da91/runtime/syntax/vim.vim#L942-L944)

Cheaterman:
OK but then a follow-up question arises - why isn't there a way to globally set syntax sync minlines for all syntaxes? Related question, what makes you think 10 is a reasonable default (when it's clearly not, see my screenshots above)?

chrisbra:
> OK but then a follow-up question arises - why isn't there a way to globally set syntax sync minlines for all syntaxes? 

Is there such a global thing? Where?

> Related question, what makes you think 10 is a reasonable default (when it's clearly not, see my screenshots above)?

Just because it failed for your html files, doesn't mean it fails for anybody else. I don't work a lot with html files but whenever I did, it worked for me. And I guess nobody has complained until now ?

Cheaterman:
It looks to me like if I do _exactly_ what `vim.syntax` is doing, it will be functionally equivalent to removing the line for most users (having `g:htmlsyn_minlines` unset). Is this what we want?

chrisbra:
> I'm literally asking why it doesn't exist.

Why should it? It clearly depends on the syntax you are using.


> That's the part that genuinely surprises me - clearly when writing SFCs you're gonna have a relatively large script block and a relatively large style block, both of which are gonna be incorrectly synced 90% of the time, which is why (after frustratingly and frantically using :syn sync fromstart for years) I made this PR.

And thanks for that.

> I'm working on the g:htmlsyn_minlines right now, but still I feel like it would be a good idea for users to have a global way to deal with syntax sync minlines, don't you think? Since the main issue with sync minlines is performance, scaling it up depends largely on how fast one's machine is, not really per-language

Ideally it should really be the responsibility of each syntax script maintainer to find reasonable good defaults, so that not each user has to fiddle with a global variable that does not always work for all kinds of languages. But it seems it doesn't always work well enough.


> It looks to me like if I do exactly what vim.syntax is doing, it will be functionally equivalent to removing the line for most users (having g:htmlsyn_minlines unset). Is this what we want?

@dkearns what's your preference?

Cheaterman:
There's an extra issue for the latter point (documenting it in `ft-html-syntax`) - it looks to me like this documentation isn't representative of what we observe when editing HTML files (eg the part about <> and </> being colored differently), so perhaps this entire documentation page is out of date?

PR #14074, opened 2024-02-22
Skip checks in win_gotoid when id == curwin

### LINKED ISSUE #14073: c-x c-o error
### Steps to reproduce

c-x c-o error

### Expected behaviour

image

### Version of Vim

9.1.125

### Environment

linux

### Logs and stack traces

_No response_

seandewar:
Problem: win_gotoid checks for textlock and other things when switching to a window that is already current.

Solution: return early with success when attempting to switch to curwin.

Closes #14073; other potential causes of E565 from win_gotoid after v9.1.0119 should be correct. Plugins can consider using win_execute instead if they wish to temporarily switch windows during textlock (though maybe that's also a bit dubious, especially if win_execute ends up closing the old curwin... maybe closing it is not possible though, as there's some checks for textlock when doing that).

### DIFF
--- a/src/evalwindow.c
+++ b/src/evalwindow.c
@@ -824,6 +824,13 @@ f_win_gotoid(typval_T *argvars, typval_T *rettv)
 	return;
 
     id = tv_get_number(&argvars[0]);
+    if (curwin->w_id == id)
+    {
+	// Nothing to do.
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks. Makes sense. Appreciate the quick turn-around.

PR #14075, opened 2024-02-22
Naming a non-pointer variable "oap" is strange

zeertzjq:
Problem:  Naming a non-pointer variable "oap" is strange.
Solution: Rename it to "oa". Also prevent using freed memory in case of
          memory allocation failure.


chrisbra:
thanks!

PR #14076, opened 2024-02-22
Internal error when passing mark in another buffer to getregion()

zeertzjq:
Problem:  Internal error when passing mark in another buffer to
          getregion().
Solution: Don't allow marks in another buffer.


chrisbra:
yes makes sense, thanks!

PR #14078, opened 2024-02-22
win_splitmove UAF if win_split_ins autocommands delete "wp"

seandewar:
Problem: heap-use-after-free in win_splitmove if Enter/Leave autocommands from
win_split_ins immediately closes "wp".

Solution: check that "wp" is valid after win_split_ins.

(This one's not a regression; just missed this possibility previously)


chrisbra:
thanks for that. Makes sense of course.

PR #14079, opened 2024-02-22
Fix truncation of text_wrap 'wrap' virtual text after EOL list char

### LINKED ISSUE #12725: Virtual text using text_wrap 'wrap' while "list" is enabled will not show wrapped lines correctly
### Steps to reproduce

1. Save the following as test.vim:
 ```vim
 vim9script
:80vsplit
# A
# B 40 * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
# C
# D
set list

prop_type_add('test', {highlight: 'Error'})
prop_add(4, 0, {
  'text': ' > 40 * bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb',
  'type': 'test',
  'text_wrap': 'wrap',
})
 ```
2. Execute `vim --clean test.vim -S test.vim`
3. Move cursor to line 4 and move down (`4Gj`)

### Expected behaviour

The text+virtual text in line 4 is longer than 80 characters and should wrap to the next (visual) line. The cursor should be on line 5 (C).

I.e. it shoud look something like this (I've taken this screenshot with `nolist`):
expected

The actual behavior looks like this:
actual

1. The virtual text is obviously not wrapped.
2. If you step over this line with the cursor, the cursor seems to skip the following line (C) and seems to be on (D) instead. This is only visually, though. If you enter Insert mode, it will be on line 5, as expected.

### Version of Vim

9.0.1677

### Environment

OS: Linux Mint 21.2
Terminal: gnome-terminal 3.44
Shell: bash 5.1.16

### Logs and stack traces

_No response_

dylanahsmith:
Fixes: #12725

As mentioned in https://github.com/vim/vim/issues/12725#issuecomment-1957206271, the problem was with that `|| lcs_eol_one == -1` `break` condition.

It looks like that condition was compensating for an incorrect outer `wlv.p_extra != at_end_str` that intended to check if the EOL character hadn't been added to the screen line, but that would evaluate to true because when `wlv.p_extra` is updated to the end of the string without being set to `at_end_str`, such as when the virtual text has been truncated by `text_prop_position` and the truncated text as finished being added to the screen line.  The inner `lcs_eol_one == -1` was compensating for this for the truncation case, but then caused problems for the cause where truncation of the virtual text shouldn't happen.  So I moved that problematic inner condition and used it to replace the incorrect outer condition.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -4116,7 +4116,7 @@ win_line(
 		    || text_prop_next <= last_textprop_text_idx
 #endif
 		    || (wp->w_p_list && wp->w_lcs_chars.eol != NUL
-						&& wlv.p_extra != at_end_str)
+						&& lcs_eol_one != -1)
 		    || (wlv.n_extra != 0 && (wlv.c_extra != NUL
 						      || *wlv.p_extra != NUL)))
 		)
@@ -4135,16 +4135,13 @@ win_line(
 
 	    // When not wrapping and finished diff lines, or when displayed
 	    // '$' and highlighting until last column, break here.
-	    if (((!wp->w_p_wrap
+	    if (!wp->w_p_wrap
 #ifdef FEAT_DIFF
 			&& wlv.filler_todo <= 0
 #endif
 #ifdef FEAT_PROP_POPUP
 			&& !text_prop_above
-#endif
-		 ) || lcs_eol_one == -1)
-#ifdef FEAT_PROP_POPUP
-		    && !text_prop_follows
+			&& !text_prop_follows
 #endif
 		       )
 		break;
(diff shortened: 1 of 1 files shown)


src/drawline.c
```diff
@@ -4116,7 +4116,7 @@ win_line(
 		    || text_prop_next <= last_textprop_text_idx
 #endif
 		    || (wp->w_p_list && wp->w_lcs_chars.eol != NUL
-						&& wlv.p_extra != at_end_str)
+						&& lcs_eol_one != -1)
```
zeertzjq, line 4117: This makes the `at_end_str` variable no longer necessary, as it is now only used once in `wlv.p_extra = at_end_str`, which can now be changed to just `wlv.p_extra = ""`.

src/drawline.c
```diff
@@ -4135,16 +4135,13 @@ win_line(
 
 	    // When not wrapping and finished diff lines, or when displayed
 	    // '$' and highlighting until last column, break here.
```
zeertzjq:
```suggestion
	    // When not wrapping and finished diff lines, break here.
```

chrisbra:
thanks, makes sense.

PR #14083, opened 2024-02-24
"C" doesn't include composing chars with 'virtualedit'

zeertzjq:
Problem:  "C" doesn't include composing chars with 'virtualedit'.
Solution: Use mb_head_off() and mb_ptr2len() instead of mb_tail_off().


chrisbra:

Issue #9181, opened 2021-11-21
Patch 8.2.3607 (commit 9459b8d46) is breaking guiligatures in gvim on cursor movement


dusanx:
### Steps to reproduce

1. Set guiligatures so something reasonable like `:set guiligatures=!#$&*+-./:<=>?@[]^\|~\"`
2. Type line with ligature characters, for instance `==>  >-- && || /= != !==`, if font supports it this char combinations will be shown like proper ligatures.
3. Move cursor over this characters, left and right, ligatures are not restored after cursor passes.

### Expected behaviour

Since we draw cursor as separate string, every underlying ligature combo was shown as ligature, while cursor char was being shown as non-ligaturized char, exactly as it should be. On cursor movement we did restore more than just a cursor char, which in turn redraws entire ligature combo properly. With patch 9459b8d46 cursor redrawing is optimized so we don't restore underlying ligature combo.

Solution:
9459b8d46 is going in good direction but when restoring screen on cursor movement we need to test and if it was ligature char on previous cursor position we need to restore (left and right) all other nested ligature chars, which will in turn restore ligature as whole.

I'll take a look at this in next day or two, of course if somebody wants to jump in just let me know.

### Operating system

Linux

### Version of Vim

9459b8d46 and onwards

dusanx:
@presuku after looking at code for 9459b8d46, maybe it is wiser if you have time to look at this. I have described possible solution above but I am not that great with GTK3 -- I can fix the problem but it can take couple of days. What do you think?

dusanx:
Update: 9459b8d46 is not directly responsible for this, it is well written, it just exposed underlying things that need fixing. Fix will be in gui.c, gui_undraw_cursor() so we can keep everything done so far. I'll have PR in couple of days. Two questions:

1. Working in gui.c and gui_undraw_cursor() I need to add GTK specific code. Is it Ok to do it in gui.c?
2. gui.cursor_col is holding cursor col position. How do I find out last visible column for that view? (I can look this up somewhere in code but I would appreciate help so I don't waste time digging).

brammool:
> Update: 9459b8d46 is not directly responsible for this, it is well
> written, it just exposed underlying things that need fixing. Fix will
> be in gui.c, gui_undraw_cursor() so we can keep everything done so
> far. I'll have PR in couple of days. Two questions:
> 
> 1. Working in gui.c and gui_undraw_cursor() I need to add GTK specific
> code. Is it Ok to do it in gui.c?

[...]

Issue #12901, opened 2023-08-23
[gVim] Font ligatures don't display correctly when directly placed under cursor


thunderbird-1990:
### Steps to reproduce

I disabled cursor blinking for the purpose of screenshots (`set guicursor+=a:blinkon0`), but the behaviours are the same with or without this line.

1. Enable ligatures (e.g., `set guiligatures=!\"#$%&()*+-./:<=>?@[]^_{\|~`)
2. Use a font that supports ligatures (e.g., `set guifont=JetBrainsMono\ 12`)
3. Place the cursor over a character that is a part of a ligature
4. The underlying (non-ligatured) character is shown, which is a bigger problem in insert mode because _both_ the non-ligatured character and the ligature are shown

![ligature_01](https://github.com/vim/vim/assets/88970199/8363ac54-a14f-4645-9503-58c820e0168f)
![ligature_02](https://github.com/vim/vim/assets/88970199/ff23dc59-e497-455e-8b00-c87c91d136e4)
![ligature_03](https://github.com/vim/vim/assets/88970199/f96dadee-8129-46b9-8a63-b6975c24c51d)
![ligature_04](https://github.com/vim/vim/assets/88970199/e499769c-6e74-4aa3-9b9a-b4bc2823d3bd)

### Expected behaviour

I expected only the ligature to be shown.

### Version of Vim

9.0.1736

### Environment

OS: Arch Linux
Typeface/font: [JetBrains Mono](https://www.jetbrains.com/lp/mono/)
gVim: 9.0.1736
Colour theme:  [PaperColor](https://github.com/NLKNguyen/papercolor-theme)

thunderbird-1990:
Can anyone replicate this? Or fail to replicate? That would also be helpful.

Just want to add that my desktop environment is KDE. This bug continues to exist on my machine with gVim 9.0.2143.

CaffeineViking:
I think I'm seeing something similar.

OS: Windows 10
Typeface/font: Iosevka
gVim: 9.1.0
Color theme: gruvbox

The ligature starts in a broken state, but eventually fixes itself (mostly) after a while. But whenever you bring the cursor on-top of a ligature again, it looks broken again. By pressing Ctrl-L (which IIRC redraws the screen) the problem is completely "fixed" until you start moving the cursor again or write more ligatures.

[...]

PR #14084, opened 2024-02-24
Fixed font ligatures not rendering correctly on cursor movement in Win32

CaffeineViking:
Related open issues: #12901 
Related closed issues: #9181

Problem: font ligatures don't render correctly in the Win32 GUI-version of `gvim` even when `set rop=type:directx` is used. Setting `guiligatures` also doesn't make any difference. This leads to broken font ligatures when the cursor passes through them. It does not recover from this, and they remain broken until you re-render the whole buffer (e.g. by using Ctrl+L). See GIF below before fix.

ligatures bug (broken)

Steps to reproduce/test:

1. Open any Win32 GUI-release of `gvim`
2. `set guifont=Iosevka`
3. `set rop=type:directx`
4. Notice font ligatures rendering correctly, until you pass the cursor through them.

Solution: the problem is that we only re-draw the current and previous character in `gui_undraw_cursor` and only have the special case for GTK when it comes to rendering ligatures. This PR enables `gui_adjust_undraw_cursor_for_ligatures` to also happen for Win32 GUI if `guiligatures` is setup correctly (all this does is expand the range of `gui_undraw_cursor` with ligature characters).

Here's a GIF of how it behaves after the fix:

ligatures bug (working)

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -4000,7 +4000,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 						*'guiligatures'* *'gli'* *E1243*
 'guiligatures' 'gli'	string	(default "")
 			global
-			{only for GTK GUI}
+			{only for GTK and Win32 GUI}
 	List of ASCII characters that, when combined together, can create more
 	complex shapes. Each character must be a printable ASCII character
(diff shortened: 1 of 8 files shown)


chrisbra:
ah, great, thanks!

PR #14086, opened 2024-02-24
Insufficient testing for 'delcombine'

zeertzjq:
Problem:  Insufficient testing for 'delcombine'.
Solution: Add test for both Normal and Insert modes without Arabic.


chrisbra:
thanks!

PR #14090, opened 2024-02-25
Getregion opts

Shougo:
Continues of https://github.com/vim/vim/pull/13998

* Add opts arguments

* `{pos1}` and `{pos2}` can be list

~~I don't understand what is the correct argument check.~~

Note: I want to add `winid` support, but I gave up.  current routine depends on current buffer...

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -263,7 +263,7 @@ getqflist({what})		Dict	get specific quickfix list properties
 getreg([{regname} [, 1 [, {list}]]])
 				String or List   contents of a register
 getreginfo([{regname}])		Dict	information about a register
-getregion({pos1}, {pos2}, {type} [, {opts}])
+getregion({pos1}, {pos2}[, {opts}])
 				List	get the text from {pos1} to {pos2}
 getregtype([{regname}])		String	type of a register
 getscriptinfo([{opts}])		List	list of sourced scripts
@@ -4271,9 +4271,8 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
-getregion({pos1}, {pos2}, {type} [, {opts}])			*getregion()*
-		Returns the list of strings from {pos1} to {pos2} as if it's
-		selected in visual mode of {type}.
+getregion({pos1}, {pos2}[, {opts}])			*getregion()*
+		Returns the list of strings from {pos1} to {pos2}.
 
 		For possible values of {pos1} and {pos2} are |List| or |string|.
 		When {pos1} or {pos2} is |string|, see |line()|.
@@ -4282,17 +4281,21 @@ getregion({pos1}, {pos2}, {type} [, {opts}])			*getregion()*
 			[{lnum}, {col}, {off}]
 
 		{type} is the selection type:
-			"v" for |characterwise| mode
-			"V" for |linewise| mode
-			"<CTRL-V>" for |blockwise-visual| mode
 
 		The optional argument {opts} is a Dict and supports the
 		following items:
 
-		exclusive	Specify whether the ending column is exclusive
-				instead of 'selection' value.
+			type		Specify the selection type instead of
+					|visualmode()| value.
 
-		You can get the last selection type by |visualmode()|.
+					"v" for |characterwise| mode
+					"V" for |linewise| mode
+					"<CTRL-V>" for |blockwise-visual| mode
+
+			exclusive	Specify whether the ending column is
+					exclusive instead of 'selection'
+					value.
+
 		If Visual mode is active, use |mode()| to get the Visual mode
 		(e.g., in a |:vmap|).
 		This function uses the line and column number from the
@@ -4318,10 +4321,10 @@ getregion({pos1}, {pos2}, {type} [, {opts}])			*getregion()*
 
 		Examples: >
 			:xnoremap <CR>
-			\ <Cmd>echow getregion('v', '.', mode())<CR>
+			\ <Cmd>echow getregion('v', '.')<CR>
 <
 		Can also be used as a |method|: >
-			'.'->getregion("'a", 'v')
+			'.'->getregion("'a")
 <
 getregtype([{regname}])					*getregtype()*
 		The result is a String, which is type of register {regname}.
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4275,12 +4275,16 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		Returns the list of strings from {pos1} to {pos2} in current
 		buffer.
 
-		For possible values of {pos1} and {pos2} are |List| or
-		|string|.
-		When {pos1} or {pos2} is |string|, see |line()|.
-		When {pos1} or {pos2} is |List|, with two or three items:
-			[{lnum}, {col}]
-			[{lnum}, {col}, {off}]
+		{pos1} and {pos2} must be a |List| with four numbers:
+			[bufnum, lnum, col, off]
+		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
+		is the buffer number of the mark.
+		"lnum" and "col" are the position in the buffer.  The first
+		column is 1.
+		The "off" number is zero, unless 'virtualedit' is used.  Then
+		it is the offset in screen columns from the start of the
+		character.  E.g., a position within a <Tab> or after the last
+		character.
 
 		{type} is the selection type:
 
@@ -4315,19 +4319,17 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the selection starts or ends in the middle of a multibyte
 		  character, it is not included but its selected part is
 		  substituted with spaces.
-		- If {pos1} or {pos2} equals "v" (see |line()|) and it is not in
-		  |visual-mode|, an empty list is returned.
-		- If {pos1}, {pos2} or {type} is an invalid string, an empty
-		  list is returned.
-		- If {pos1} or {pos2} is a mark in different buffer, an empty
-		  list is returned.
+		- If {type} is an invalid string, an empty list is returned.
+		- If {pos1} or {pos2} is not current buffer, an empty list is
+		  returned.
 
 		Examples: >
 			:xnoremap <CR>
-			\ <Cmd>echow getregion('v', '.', #{ type: mode() })<CR>
+			\ <Cmd>echow getregion(
+			\ getpos('v'), getpos('.'), #{ type: mode() })<CR>
 <
 		Can also be used as a |method|: >
-			'.'->getregion("'a")
+			'.'->getregion(getpos("'a"))
 <
 getregtype([{regname}])					*getregtype()*
 		The result is a String, which is type of register {regname}.
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -263,7 +263,7 @@ getqflist({what})		Dict	get specific quickfix list properties
 getreg([{regname} [, 1 [, {list}]]])
 				String or List   contents of a register
 getreginfo([{regname}])		Dict	information about a register
-getregion({pos1}, {pos2}[, {opts}])
+getregion({pos1}, {pos2} [, {opts}])
 				List	get the text from {pos1} to {pos2}
 getregtype([{regname}])		String	type of a register
 getscriptinfo([{opts}])		List	list of sourced scripts
@@ -4271,10 +4271,12 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
-getregion({pos1}, {pos2}[, {opts}])			*getregion()*
-		Returns the list of strings from {pos1} to {pos2}.
+getregion({pos1}, {pos2} [, {opts}])			*getregion()*
+		Returns the list of strings from {pos1} to {pos2} in current
+		buffer.
 
-		For possible values of {pos1} and {pos2} are |List| or |string|.
+		For possible values of {pos1} and {pos2} are |List| or
+		|string|.
 		When {pos1} or {pos2} is |string|, see |line()|.
 		When {pos1} or {pos2} is |List|, with two or three items:
 			[{lnum}, {col}]
@@ -4286,7 +4288,7 @@ getregion({pos1}, {pos2}[, {opts}])			*getregion()*
 		following items:
 
 			type		Specify the selection type instead of
-					|visualmode()| value.
+					"v" value.
 
 					"v" for |characterwise| mode
 					"V" for |linewise| mode
@@ -4296,6 +4298,7 @@ getregion({pos1}, {pos2}[, {opts}])			*getregion()*
 					exclusive instead of 'selection'
 					value.
 
+		You can get the last selection type by |visualmode()|.
 		If Visual mode is active, use |mode()| to get the Visual mode
 		(e.g., in a |:vmap|).
 		This function uses the line and column number from the
@@ -4321,7 +4324,7 @@ getregion({pos1}, {pos2}[, {opts}])			*getregion()*
 
 		Examples: >
 			:xnoremap <CR>
-			\ <Cmd>echow getregion('v', '.')<CR>
+			\ <Cmd>echow getregion('v', '.', #{ type: mode() })<CR>
 <
 		Can also be used as a |method|: >
 			'.'->getregion("'a")
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4281,10 +4281,7 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		is the buffer number of the mark.
 		"lnum" and "col" are the position in the buffer.  The first
 		column is 1.
-		The "off" number is zero, unless 'virtualedit' is used.  Then
-		it is the offset in screen columns from the start of the
-		character.  E.g., a position within a <Tab> or after the last
-		character.
+		The "off" number is ignored.
 
 		{type} is the selection type:
 
@@ -4319,7 +4316,6 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the selection starts or ends in the middle of a multibyte
 		  character, it is not included but its selected part is
 		  substituted with spaces.
-		- If {type} is an invalid string, an empty list is returned.
 		- If {pos1} or {pos2} is not current buffer, an empty list is
 		  returned.
 
@@ -4329,7 +4325,7 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 			\ getpos('v'), getpos('.'), #{ type: mode() })<CR>
 <
 		Can also be used as a |method|: >
-			'.'->getregion(getpos("'a"))
+			getpos('.')->getregion(getpos("'a"))
 <
 getregtype([{regname}])					*getregtype()*
 		The result is a String, which is type of register {regname}.
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4275,15 +4275,10 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		Returns the list of strings from {pos1} to {pos2} in current
 		buffer.
 
-		{pos1} and {pos2} must be a |List| with four numbers:
+		{pos1} and {pos2} must be a |List| with four numbers in the
+		same format as the result of |getpos()|:
 			[bufnum, lnum, col, off]
-		"bufnum" must be zero or current buffer number.
-		"lnum" and "col" are the position in the buffer.  The first
-		column is 1.
-		The "off" number is zero, unless 'virtualedit' is used.  Then
-		it is the offset in screen columns from the start of the
-		character.  E.g., a position within a <Tab> or after the last
-		character.
+		And "bufnum" must be zero or current buffer number.
 
 		{type} is the selection type:
 
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -263,7 +263,7 @@ getqflist({what})		Dict	get specific quickfix list properties
 getreg([{regname} [, 1 [, {list}]]])
 				String or List   contents of a register
 getreginfo([{regname}])		Dict	information about a register
-getregion({pos1}, {pos2}, {type})
+getregion({pos1}, {pos2}, {type} [, {opts}])
 				List	get the text from {pos1} to {pos2}
 getregtype([{regname}])		String	type of a register
 getscriptinfo([{opts}])		List	list of sourced scripts
@@ -4271,7 +4271,7 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
-getregion({pos1}, {pos2}, {type})			*getregion()*
+getregion({pos1}, {pos2}, {type} [, {opts}])			*getregion()*
 		Returns the list of strings from {pos1} to {pos2} as if it's
 		selected in visual mode of {type}.
 		For possible values of {pos1} and {pos2} see |line()|.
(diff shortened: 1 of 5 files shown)


src/evalfunc.c
```diff
@@ -5490,37 +5490,70 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
+    if (fnum >= 0 && fnum != curbuf->b_fnum)
 	return;
-    p1 = *fp;
 
-    fp = var2fpos(&argvars[1], TRUE, &fnum, FALSE);
-    if (fp == NULL || (fnum >= 0 && fnum != curbuf->b_fnum))
+    if (argvars[1].v_type == VAR_LIST)
+    {
+	if (!list2fpos(&argvars[1], &p2, &fnum, &curswant, FALSE))
+	    return;
```
zeertzjq: It seems that `var2fpos()` is more often used for functions that get some value (e.g. `col()`, `virtcol()`), while `list2fpos()` is used for functions that change cursor position (e.g. `cursor()`). Also, `curswant` isn't used in this function, so `var2fpos()` behavior is less confusing.

Shougo: Oh, I have not know `var2fpos()` support list...

chrisbra:
Thanks, It's a bit disappointing, that we first make it using string args and now we add in addition list args, plus an optional dict type. I thought we could avoid having different types for the same function. Also, initially when I suggested the `type` arg, I meant it to be using a dict object instead. Now we are adding yet another dict argument for additional settings. 

Let's see what others think.

Shougo:
I have decided the 3rd argument `{type}` is not optional.
Because it is not only for visual mode function.  It should be explicit.

runtime/doc/builtin.txt
```diff
@@ -4271,14 +4271,27 @@ getreginfo([{regname}])					*getreginfo()*
...
 
-getregion({pos1}, {pos2}, {type})			*getregion()*
+getregion({pos1}, {pos2}, {type} [, {opts}])			*getregion()*
 		Returns the list of strings from {pos1} to {pos2} as if it's
 		selected in visual mode of {type}.
-		For possible values of {pos1} and {pos2} see |line()|.
+
+		For possible values of {pos1} and {pos2} are |List| or |string|.
+		When {pos1} or {pos2} is |string|, see |line()|.
+		When {pos1} or {pos2} is |List|, with two, three or four item:
```
```suggestion
		When {pos1} or {pos2} is |List|, with two or three items:
```

yegappan:
> Thanks, It's a bit disappointing, that we first make it using string args and now we add in addition list args, plus an optional dict type. I thought we could avoid having different types for the same function. Also, initially when I suggested the `type` arg, I meant it to be using a dict object instead. Now we are adding yet another dict argument for additional settings.
> 
> Let's see what others think.

I agree.  In one of the comments for the previous PR, I mentioned this point about using a Dict
for the third argument.  The third argument should be a Dict.

yegappan:
> I have decided the 3rd argument `{type}` is not optional. Because it is not only for visual mode function. It should be explicit.

You can use a default behavior.  You can add a "visualmode" item in the Dict that specifies the desired visual mode.

src/evalfunc.c
```diff
@@ -2132,7 +2132,7 @@ static funcentry_T global_functions[] =
 			ret_getreg,	    f_getreg},
     {"getreginfo",	0, 1, FEARG_1,	    arg1_string,
 			ret_dict_any,	    f_getreginfo},
-    {"getregion",	3, 3, FEARG_1,	    arg3_string,
+    {"getregion",	3, 4, FEARG_1,	    arg3_string,
```
yegappan: The arg3_string type check will no longer work this.  You need to have a special function that checks whether
the first argument is a String or a List and the fourth optional argument is a Dict.  In Vim9, supporting different
types for the same argument is deprecated.  From a type checking perspective, it is better to support only
one type for the arguments.

src/testdir/test_visual.vim
```diff
@@ -1674,6 +1674,14 @@ func Test_visual_getregion()
   call assert_equal(['two'], "'a"->getregion("'a", 'V'))
   call assert_equal(['one', 'two'], "."->getregion("'a", "\<c-v>"))
 
+  " Using List
+  call cursor(1, 1)
+  call assert_equal(['one', 'two'], [2, 3]->getregion('.', 'v'))
+  call assert_equal(['one', 'two'], '.'->getregion([2, 3], 'v'))
+  call assert_equal(['one', 'two'], '.'->getregion([2, 3], 'V'))
+  call assert_equal(['two'], [2, 3]->getregion([2, 3], 'V'))
+  call assert_equal(['one', 'two'], '.'->getregion([2, 3], "\<c-v>"))
```
yegappan: Can you add a test for using `offset` in the List argument?

yegappan (changes requested):

chrisbra:
> I agree. In one of the comments for the previous PR, I mentioned this point about using a Dict
for the third argument. The third argument should be a Dict.

We added `getregion()` in `9.1.120` just last week. I suppose it is okay to change the `getregion()` behaviour and just use 

```
getregion({pos1}, {pos2}, {Dict})			*getregion()*
		Returns the list of strings from {pos1} to {pos2} as if it's
		selected in visual mode of {type}. {pos1} and {pos2} are list
                items. For possible values see |getpos()|
		{Dict} is a dictionary with additional properties and supports
                the following keys:
                'type':  specify the selection type, supports the following
                values: "v" for |characterwise| mode
			"V" for |linewise| mode
			"<CTRL-V>" for |blockwise-visual| mode
                'selection': specify the 'selection' mode using the values
                             from the 'selection' setting.
```

Something like this?

yegappan:
> > I agree. In one of the comments for the previous PR, I mentioned this point about using a Dict
> > for the third argument. The third argument should be a Dict.
> 
> We added `getregion()` in `9.1.120` just last week. I suppose it is okay to change the `getregion()` behaviour and just use
> 
> ```
> getregion({pos1}, {pos2}, {Dict})			*getregion()*
> 		Returns the list of strings from {pos1} to {pos2} as if it's
> 		selected in visual mode of {type}. {pos1} and {pos2} are list
>                 items. For possible values see |getpos()|
> 		{Dict} is a dictionary with additional properties and supports
>                 the following keys:
>                 'type':  specify the selection type, supports the following
>                 values: "v" for |characterwise| mode
> 			"V" for |linewise| mode
> 			"<CTRL-V>" for |blockwise-visual| mode
>                 'selection': specify the 'selection' mode using the values
>                              from the 'selection' setting.
> ```
> 
> Something like this?

Yes.  That looks good to me.

Shougo:
OK.  I have changed `{type}` to `{opts}`.

src/evalfunc.c

runtime/doc/builtin.txt
```diff
@@ -263,7 +263,7 @@ getqflist({what})		Dict	get specific quickfix list properties
 getreg([{regname} [, 1 [, {list}]]])
 				String or List   contents of a register
 getreginfo([{regname}])		Dict	information about a register
-getregion({pos1}, {pos2}, {type})
+getregion({pos1}, {pos2}[, {opts}])
```
```suggestion
getregion({pos1}, {pos2} [, {opts}])
```

runtime/doc/builtin.txt
```diff
@@ -4271,15 +4271,31 @@ getreginfo([{regname}])					*getreginfo()*
 		Can also be used as a |method|: >
 			GetRegname()->getreginfo()
 
-getregion({pos1}, {pos2}, {type})			*getregion()*
-		Returns the list of strings from {pos1} to {pos2} as if it's
-		selected in visual mode of {type}.
-		For possible values of {pos1} and {pos2} see |line()|.
+getregion({pos1}, {pos2}[, {opts}])			*getregion()*
```
```suggestion
getregion({pos1}, {pos2} [, {opts}])			*getregion()*
```

src/evalfunc.c
```diff
@@ -5517,16 +5522,35 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
+	pos2 = tv_get_string(&argvars[1]);
 
     is_visual = (pos1[0] == 'v' && pos1[1] == NUL)
 	|| (pos2[0] == 'v' && pos2[1] == NUL);
 
     if (is_visual && !VIsual_active)
 	return;
 
+    default_type[0] = VIsual_mode;
+    default_type[1] = NUL;
```
zeertzjq: I don't think it's a good idea to use `VIsual_mode` as the default type, as it seems that changing the motion force of an operator can also change `VIsual_mode`. I think it's better to have a deterministic default value "v".

Shougo:
I have changed `{type}` default value.

Shane-XB-Qian:
> In Vim9, supporting different types for the same argument is deprecated.
> From a type checking perspective, it is better to support only
> one type for the arguments.

i may suggest re-thinking this "claim", IMO:
1. user defined func perhaps should be only one type for an argument,
2. but to vim native function "supporting different types for the same argument" was a tricky/classic/good design.

-- 
shane.xb.qian

chrisbra:
Thanks, but I'd rather see either we change the first two position arguments to a list type and drop the string type or we do not change the type further. But let's not do both. It's also better for Vim9 type checking when compiling the function.

Shougo:
> Thanks, but I'd rather see either we change the first two position arguments to a list type and drop the string type or we do not change the type further. But let's not do both. It's also better for Vim9 type checking when compiling the function.

I have changed the interface.  `{string}` is not allowed.

chrisbra:
Can I get some comments before I merge? Everybody fine with this backwards-incompatible change?

habamax:
> Can I get some comments before I merge? Everybody fine with this backwards-incompatible change?

I think it is fine, I doubt anyone used `getregion()` in the plugins yet.

runtime/doc/builtin.txt
```diff
@@ -4315,19 +4319,17 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
...
 
 		Examples: >
 			:xnoremap <CR>
-			\ <Cmd>echow getregion('v', '.', #{ type: mode() })<CR>
+			\ <Cmd>echow getregion(
+			\ getpos('v'), getpos('.'), #{ type: mode() })<CR>
 <
 		Can also be used as a |method|: >
-			'.'->getregion("'a")
+			'.'->getregion(getpos("'a"))
```
habamax: In this example '.' is a string representing current position? But pos1 couldn't be a string, right?

runtime/doc/builtin.txt
```diff
@@ -4275,12 +4275,16 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
...
+		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
+		is the buffer number of the mark.
+		"lnum" and "col" are the position in the buffer.  The first
+		column is 1.
+		The "off" number is zero, unless 'virtualedit' is used.  Then
+		it is the offset in screen columns from the start of the
+		character.  E.g., a position within a <Tab> or after the last
+		character.
 
 		{type} is the selection type:
```
habamax, line 4283: Should this line be removed?

runtime/doc/builtin.txt
```diff
@@ -4315,19 +4319,17 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the selection starts or ends in the middle of a multibyte
 		  character, it is not included but its selected part is
 		  substituted with spaces.
-		- If {pos1} or {pos2} equals "v" (see |line()|) and it is not in
-		  |visual-mode|, an empty list is returned.
-		- If {pos1}, {pos2} or {type} is an invalid string, an empty
-		  list is returned.
-		- If {pos1} or {pos2} is a mark in different buffer, an empty
-		  list is returned.
+		- If {type} is an invalid string, an empty list is returned.
```
habamax: `{type}` is in the dict now, and here it is referenced as a standalone parameter of a function.

runtime/doc/builtin.txt

runtime/doc/builtin.txt
Shougo: Yes.  I have removed.

Shougo:
I have fixed the documentation.

runtime/doc/builtin.txt
```diff
@@ -4271,14 +4271,34 @@ getreginfo([{regname}])					*getreginfo()*
...
+		Returns the list of strings from {pos1} to {pos2} in current
+		buffer.
+
+		{pos1} and {pos2} must be a |List| with four numbers:
+			[bufnum, lnum, col, off]
+		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
+		is the buffer number of the mark.
+		"lnum" and "col" are the position in the buffer.  The first
+		column is 1.
+		The "off" number is ignored.
```
zeertzjq: I don't think "off" is ignored. `coladd` is used in the code.

runtime/doc/builtin.txt
```diff
@@ -4271,14 +4271,34 @@ getreginfo([{regname}])					*getreginfo()*
...
-		selected in visual mode of {type}.
-		For possible values of {pos1} and {pos2} see |line()|.
+getregion({pos1}, {pos2} [, {opts}])			*getregion()*
+		Returns the list of strings from {pos1} to {pos2} in current
+		buffer.
+
+		{pos1} and {pos2} must be a |List| with four numbers:
+			[bufnum, lnum, col, off]
+		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
+		is the buffer number of the mark.
```
zeertzjq: There is no mention of "mark" anywhere else in the docs of this function, and it doesn't allow non-current buffer.

runtime/doc/builtin.txt
Shougo: Oh, I forgot it.

zeertzjq:
Actually now that `pos1` and `pos2` only support lists, it should be possible to support non-current buffer as long as they have the same buffer number.

Shougo:
> Actually now that pos1 and pos2 only support lists, it should be possible to support non-current buffer as long as they have the same buffer number.

It is possible.  But I gave up.  The PR is welcome.

runtime/doc/builtin.txt
```diff
@@ -4271,14 +4271,36 @@ getreginfo([{regname}])					*getreginfo()*
...
+
+		{pos1} and {pos2} must be a |List| with four numbers:
+			[bufnum, lnum, col, off]
+		"bufnum" must be zero or current buffer number.
+		"lnum" and "col" are the position in the buffer.  The first
+		column is 1.
+		The "off" number is zero, unless 'virtualedit' is used.  Then
+		it is the offset in screen columns from the start of the
+		character.  E.g., a position within a <Tab> or after the last
+		character.
```
zeertzjq: I think it's better to simply reference `getpos()`
```suggestion
		{pos1} and {pos2} must be a |List| with four numbers in the
		same format as the result of |getpos()|:
			[bufnum, lnum, col, off]
		While "bufnum" must be zero or current buffer number.
```

src/evalfunc.c
```diff
@@ -5491,41 +5492,44 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
-    pos_T		*fp = NULL;
-    char_u		*pos1, *pos2, *type;
+    char_u		*type;
+    char_u		default_type[] = "v";
     int			save_virtual = -1;
     int			l;
     int			region_type = -1;
-    int			is_visual;
+    int			is_select_exclusive;
+    colnr_T		curswant = -1;
```
zeertzjq: `curswant` is not used and `list2fpos()` can accept `NULL` as `curswantp`
```suggestion
```

src/evalfunc.c
```diff
@@ -5491,41 +5492,44 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
-	    || check_for_string_arg(argvars, 2) == FAIL)
+    if (check_for_list_arg(argvars, 0) == FAIL
+	    || check_for_list_arg(argvars, 1) == FAIL
+	    || check_for_opt_dict_arg(argvars, 2) == FAIL)
 	return;
 
-    // NOTE: var2fpos() returns static pointer.
-    fp = var2fpos(&argvars[0], TRUE, &fnum, FALSE);
-    if (fp == NULL || (fnum >= 0 && fnum != curbuf->b_fnum))
+    if (list2fpos(&argvars[0], &p1, &fnum, &curswant, FALSE) != OK
```
zeertzjq:
```suggestion
    if (list2fpos(&argvars[0], &p1, &fnum, NULL, FALSE) != OK
```

src/evalfunc.c
```diff
@@ -5491,41 +5492,44 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
+	    || (fnum >= 0 && fnum != curbuf->b_fnum))
 	return;
-    p1 = *fp;
 
-    fp = var2fpos(&argvars[1], TRUE, &fnum, FALSE);
-    if (fp == NULL || (fnum >= 0 && fnum != curbuf->b_fnum))
+    if (list2fpos(&argvars[1], &p2, &fnum, &curswant, FALSE) != OK
```
zeertzjq:
```suggestion
    if (list2fpos(&argvars[1], &p2, &fnum, NULL, FALSE) != OK
```

zeertzjq (commented):
I think some of the tests should omit the `opts` argument or the `type` field to check for the default `type` behavior.

Shougo:
Fixed.

Shane-XB-Qian (commented):
- Order of {pos1} and {pos2} doesn't matter, it will always
		  return content from the upper left position to the lower
		  right position.

i doubt this design, tho it is useful, but similar vim native function e.g prop_remove was not playing like that, should keep consistent?

Shougo:
Well, `prop_remove()` arguments are line numbers.  It seems `getline()` like interface.

Of course, I can change the behavior.  If pos1 pos2 is reversed, it can returns empty.

Shougo:
I have checked visual mode region behavior.  It works if visual start and cursor position is swapped.

Shane-XB-Qian:
yes, please think if it should 'order' matters, 
and actually if it was a visual selection, really needed to pass pos1/2 ?

Shougo:
Because it is not only for visual mode.  It is just get region text from `pos1` to `pos2`.

Shane-XB-Qian:
> `getregion()` <- what?

maybe `getregiion(vs_type)` was enough?

habamax:
In case of getting visual text in vimscript I don't like the idea of checking if pos1 is less than pos2 before calling the function. I often use `o` to extend visual from both sides...

chrisbra:
> yes, please think if it should 'order' matters,

Order should not matter. Let's not add burden to plugin authors so everybody has to figure out how to determine which position comes first. That is not so trivial

> and actually if it was a visual selection, really needed to pass pos1/2 ?

Yes, let's please keep it consistent. It should be fine to always give two position as arguments. The function signature should be the same always. That's better for Vim9 scripting it can easier detect potential problems.

Shane-XB-Qian:
> > yes, please think if it should 'order' matters,
> 
> Order should not matter. Let's not add burden to plugin authors so everybody has to figure out how to determine which position comes first. That is not so trivial

yes, that's the why asking such "question", since now some native function was 'order' matter.

> > and actually if it was a visual selection, really needed to pass pos1/2 ?
> 
> Yes, let's please keep it consistent. It should be fine to always give two position as arguments. The function signature should be the same always. That's better for Vim9 scripting it can easier detect potential problems.

and further, actually to this 'getregion()' function, looked the pos1/2 not only not matter, but also seems maybe not required, or the signature should be `getregion(vs_type, [pos1, [pos2]])`, that maybe was the reducing plugin author or general vim user's burden.
// just my opinion, but your call.

chrisbra:
> yes, that's the why asking such "question", since now some native function was 'order' matter.

For the purpose of getting the region, we only need to know the beginning and end-points. Order shouldn't matter, Vim can figure it out on its own which one is the starting and which one is the end position, right? For all other examples, this is not the case and then order matters of course.

> and further, actually to this 'getregion()' function, looked the pos1/2 not only not matter, but also seems maybe not required, or the signature should be getregion(vs_type, [pos1, [pos2]]), that maybe was the reducing plugin author or general vim user's burden.

I would think the positions are the most important parameters, so they should come first, don't you think? Then we have added another parameter, that allows you to further specify additional properties but you don't have to specify them, it's an optional dictionary. And we can later add additional settings, without changing the function signature.

Shane-XB-Qian:
> I would think the positions are the most important parameters, so they should come first, don't you think? 

i donot know, but please check the signature of prop_remove and prop_clear, that maybe give your answer, 
to me, i felt getregion is similar like prop_remove.
// all in all, i think the usage of vim should keep consistent, specially those tricky/classic/good design.

chrisbra:
> i donot know, but please check prop_remove and prop_clear, that maybe give your answer, to me, i felt getregion is similar like prop_remove.

I don't think so, it is quite different.

yegappan:
> I would think the positions are the most important parameters, so they should come first, don't you think? Then we have added another parameter, that allows you to further specify additional properties but you don't have to specify them, it's an optional dictionary. And we can later add additional settings, without changing the function signature.
>

Yes.  If a builtin function needs optional arguments, then it should use a Dict as the last argument.
Instead of a Dict, if a specific type is used for the optional arguments, then when using multiple optional
arguments, to specify an optional argument all the preceding optional arguments must be specified.
For example, take the count() function as an example:

   count({comp}, {expr} [, {ic} [, {start}]])

This function accepts two optional arguments.  To specify the starting position, the 'ignorecase'
argument must be specified (even if it is not needed).  This is not a good design.

Shane-XB-Qian:
> For example, consider the count() function:
> 
> count({comp}, {expr} [, {ic} [, {start}]])
> 
> This function accepts two optional arguments. To specify the starting position, the 'ignorecase' argument must be specified (even if it is not needed). This is not a good design. The last argument should have been a Dict with "ignorecase" and "startidx" items.

not same case, here pos1/2 was some kind of burden:
`to this 'getregion()' function, looked the pos1/2 not only 'order' not matter, but also seems maybe not required`

chrisbra:
> to this 'getregion()' function, looked the pos1/2 not only 'order' not matter, but also seems maybe not required

it may only be not required, if you want to get the last visual selected region. But what if you want to get the region for various other marks?

Shane-XB-Qian:
> > to this 'getregion()' function, looked the pos1/2 not only 'order' not matter, but also seems maybe not required
> 
> it may only be not required, if you want to get the last visual selected region. But what if you want to get the region for various other marks?

then what that 3rd arg `type` default `v` for?
or was there really possible needing for `other marks`? or had `getbufline` there for them.

-- 
shane.xb.qian

chrisbra:
> then what that 3rd arg `type` default `v` for?

To specify what kind of a selection you want? That matters for all cases of selection of course. 

> or was there really possible needing for `other marks`? or had `getbufline` there for them.

Why not? Why are you trying to argue so much against it?

Shane-XB-Qian:
> > then what that 3rd arg `type` default `v` for?
> 
> To specify what kind of a selection you want? That matters for all cases of selection of course. 

ok, so here 'selection' meant selection shape, not visual selection.

> > or was there really possible needing for `other marks`? or had `getbufline` there for them.
> 
> Why not? Why are you trying to argue so much against it?

i am not against it, or what benefit i would get if against it? :smile:

just the usage was confusing, and doubt the design or if "against" the passed vim classic design;
and to 'visual selection' the pos1/pos2 was redundant.

there may should be two func:
1. getregion(pos1, pos2, {opts})
2. getselection({opts})

or just similar one:
a-1. getregion({opts}, [pos1, [pos2]])
a-2. getregion({opts}, [pos1, pos2])

-- 
shane.xb.qian

chrisbra:
> if "against" the passed vim classic design

So what exaclty was vim classic design? We have always added new features. 

> 1. getregion(pos1, pos2, {opts})
> 2. getselection({opts})

This adds too much clutter. Why having 2 functions performing almost the same thing?

> a-1. getregion({opts}, [pos1, [pos2]])
> a-2. getregion({opts}, [pos1, pos2])

It doesn't make sense to `{opts}` as the first variable, since it is optional.  We could have made one list of position lists, but I think then it wouldn't work as `-> method` call

Shane-XB-Qian:
> > a-1. getregion({opts}, [pos1, [pos2]])
> > a-2. getregion({opts}, [pos1, pos2])
> 
> It doesn't make sense to `{opts}` as the first variable, since it is optional.  We could have made one list of position lists, but I think then it wouldn't work as `-> method` call

please check 'prop_remove()', similar situation, it can be `-> method` call, if `a-1` not preferred, then perhaps `a-2` was good.
and {opts} was not optional, since had to indicate a 'selection shape' anyhow as above said so, just there was a default value.

-- 
shane.xb.qian

Shane-XB-Qian:
but it is truly Not worthy to refactor it again, 
it shoud be ok as long as if user knew that was not 'visual selection' but 'selection shape'.. :shrug:
let it be~ thanks arguing with me.. :smile:

PR #14093, opened 2024-02-25
runtime(vim): Update syntax file, improve :substitute matching

### LINKED ISSUE #13883: vim file displays `[s` `(s` syntax exception
### Steps to reproduce

1. cat 1.vim
```vim
vim9script

        str[s]
        str(s)

def Test()
        str[s]
        str(s)
enddef
```
2. `vim --clean 1.vim -c 'set nu'`
step

L3 and L4 is fine
L7 and L8 is wrong

### Expected behaviour

Syntax consistent

### Version of Vim

9.1.41

### Environment

os: ubuntu18.04
terminal: windows terminal
$TERM: xterm-256color
shell: bash4.4.20

### Logs and stack traces

_No response_

dkearns:
Improve `:substitute` matching.

- Differentiate between :substitute and `substitute(),` fixes https://github.com/vim/vim/issues/13883.
- Match all allowed `:substitute` delimiters.
- Remove leading context from `:substitute` matches.


chrisbra:
thanks!

PR #14095, opened 2024-02-26
Del in cmdline mode doesn't delete composing chars

zeertzjq:
Problem:  Del in cmdline mode doesn't delete composing chars.
Solution: Use mb_head_off() and mb_ptr2len().

### DIFF
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -4188,32 +4188,12 @@ mb_copy_char(char_u **fp, char_u **tp)
     int
 mb_off_next(char_u *base, char_u *p)
 {
-    int		i;
-    int		j;
+    int		head_off = (*mb_head_off)(base, p);
 
-    if (enc_utf8)
-    {
-	if (*p < 0x80)		// be quick for ASCII
-	    return 0;
+    if (head_off == 0)
+	return 0;
 
-	// Find the next character that isn't 10xx.xxxx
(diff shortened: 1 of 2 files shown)


zeertzjq:
The help for 'delcombine' only mentions Normal and Insert mode, so it indeed shouldn't affect Cmdline mode.

tonymec:
> The help for 'delcombine' only mentions Normal and Insert mode, so it indeed shouldn't affect Cmdline mode.

The help for 'delcombine' mentions `Normal-mode x` (`x` doesn't act as a Delete command in Insert and Command-line mode anyway) and it mentions `backspace` without specifying any mode so I would expect it to apply to `< Bsp >` wherever it deletes anything, and identically in both Insert and Command-line modes (in Normal mode it is a motion command, not a Delete-backward command).

Best regards,
Tony.

zeertzjq:
Hmm, at least there is no uncertainty about the behavior of `<Del>` in Cmdline mode, and this PR achieve the expected behavior.

tonymec:
> Hmm, at least there is no uncertainty about the behavior of `<Del>` in Cmdline mode, and this PR achieves the expected behavior.

OK by me if it does. :-)

Best regards,
Tony.

chrisbra:
seems fine now, so let me include it now. Thanks!

Issue #12028, opened 2023-02-20
If 'list' is on, 'below' virtual text which includes 1 or 2 characters are gone


rickhowe:
If text is a single character and draw the "below" virtual line.

![44](https://user-images.githubusercontent.com/3096934/218414052-9e1e6764-c2fb-4ce8-b65a-a6c7665976e5.png)

Then &list is set, all the virtual lines are gone.

![55](https://user-images.githubusercontent.com/3096934/218414347-c8039172-2d6e-4500-a64b-38c4aca66d53.png)

If "above", looks like no problem.

![66](https://user-images.githubusercontent.com/3096934/218415384-3dddf85f-c260-4c1c-b6f6-12f04f5eb7bc.png)

And if "below", the problem happens only if text is 1 or 2 characters.

![77](https://user-images.githubusercontent.com/3096934/218417020-3702afdb-855a-4464-972d-08cf98b8e3da.png)

rickhowe:
I found that I forgot to mention about the link to [related problem](https://github.com/vim/vim/issues/11959#issuecomment-1427589736).

dylanahsmith:
I opened a fix for the related problem https://github.com/vim/vim/pull/14096, but didn't notice this problem when testing that, where this issue seems to share the same test script.

Am I right that this has been fixed already?

rickhowe:
Not completely fixed.

I tested on 9.1.0139.
```
call setline(1, ['11111', '', '33333', '', '55555'])
set number
let vt = 'test'
call prop_type_add(vt, {'highlight': 'ToDo'})
for ln in range(1, line('$'))
  call prop_add(ln, 0, {'type': vt, 'text': '+', 'text_align': 'below'})
endfor
```
Then `:set list`. Screen looks OK, but if press `j` to move cursor  to the 2nd line:

![image](https://github.com/vim/vim/assets/3096934/ace2f535-52e7-4de4-b55e-213afcb3b1b6)

The cursor is incorrectly moved on to the virtual text. Same things happen if virtual text is 2 characters, but no  problem for 3 or more characters

If 'above', no problem for screen and cursor.

dylanahsmith:
That's the problem my fix (#14096) addresses, so it looks like this issue can be closed by that fix.

dylanahsmith:
> The cursor is incorrectly moved on to the virtual text. Same things happen if virtual text is 2 characters

[...]

PR #14096, opened 2024-02-26
Fix cursor row after 1 char 'below' virtual text when EOL is shown

### LINKED ISSUE #11959: Window screen gets wrong when a virtual text is placed on 'above' or 'below' on an empty line
### Steps to reproduce

On the following window:

bb

Run the following script to place `'---'` above and `'+++'` below on each line:
```
let vt = 'test'
call prop_type_delete(vt)
call prop_type_add(vt, {'highlight': 'ToDo'})
for ln in range(1, line('$'))
  call prop_add(ln, 0, {'type': vt, 'text': '---', 'text_align': 'above'})
  call prop_add(ln, 0, {'type': vt, 'text': '+++', 'text_align': 'below'})
endfor
```

aa

The empty lines on both line 2 and 4 are gone. Then the line number appears on the virtual text and the cursor appears out  of last line 5.

### Expected behaviour

The empty lines should be left displayed as well.

### Version of Vim

9.0.1291

### Environment

Windows 10

### Logs and stack traces

_No response_

dylanahsmith:
Fixes #11959
Fixes #14096

## Problem

The cell size of 2 was being used for the EOL character (as if it were going to display a NULL character with `^@`) instead of 1, which propagated to the calculation of space needed to put the 'below' virtual text on the next line.  This off-by-one error resulted in the 'below' virtual text with a single character to be treated as if it fit on the line with the text it is supposed to be below, resulting in the cursor to appear to appear on a row earlier for each of these lines it is after.

## Solution

Use 1 for the cell size of the EOL character in calculating the cursor's screen position.

Instead of adding a new test, I modified an existing one that seemed to come from an earlier partial fix for the issue.  When testing with the change to src/charset.c reverted, I get a dump file difference in Test_prop_above_below_empty_2.dump which shows a difference in the line the cursor is on.


chrisbra:
okay thanks. Seems fine. I suppose however this fixes #12028 and not #14096 which is this pr here.  I am not closing #12028, because of the related `gk`/`gj` issue.

PR #14100, opened 2024-02-27
Put in Visual mode wrong if it replaces fold marker

### LINKED ISSUE #14097: Visual mode put behaves incorrectly if it replaces fold marker
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
call setline(1, repeat(['{{{1', 'foo', 'bar', ''], 4))
set foldmethod=marker
```
3. Press `GkzoyyVp`
4. Buffer content becomes
```
{{{1
foo
bar

{{{1
{{{1
foo
bar

foo
bar

{{{1
foo
bar

```

### Expected behaviour

Buffer content is still
```
{{{1
foo
bar

{{{1
foo
bar

{{{1
foo
bar

{{{1
foo
bar

```

### Version of Vim

9.1.139

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Put in Visual mode wrong if it replaces fold marker.
Solution: Temporarily disable folding during put in Visual mode.

Fix #14097

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -7323,6 +7323,9 @@ nv_put_opt(cmdarg_T *cap, int fix_indent)
     int		dir;
     int		flags = 0;
     int		keep_registers = FALSE;
+#ifdef FEAT_FOLDING
+    int		save_fen = curwin->w_p_fen;
+#endif
 
     if (cap->oap->op_type != OP_NOP)
     {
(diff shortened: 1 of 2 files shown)


zeertzjq:
The CI runners seem to be encountering some network problems.

chrisbra:
Hm,
I suppose this will break the following current behaviour: putting something in visual mode over a folded section of code, replaces the whole fold with whatever you are putting. I suppose this is no longer possible then? E.g.
```
vim --clean
call setline(1, repeat(['{{{1', 'foo', 'bar', ''], 4))
set foldmethod=marker
let @a=getline(2)
call cursor(1,1)
norm! V"ap
```
Now we should have (with `:set nofoldenable`):
```
foo
{{{1
foo
bar

{{{1
foo
bar

{{{1
foo
bar

```

We also don't seem to have a test for it. Not sure if disabling folding is the correct way to handle it.

zeertzjq:
I think it still works, because the Visual selection already includes the whole fold.

PR #14102, opened 2024-02-27
runtime(sh): Update ftplugin, fix #14101

### LINKED ISSUE #14101: Change `comments` to `b:#` to avoid automatically adding a comment leader after shebang line in shell scripts
### Steps to reproduce

1. Open a `sh` file
2. Type `#!/bin/bash` in the first line
3. Type `Enter` in `Insert` mode or `o` in `Normal` mode
4. You will see a comment leader is automatically added in the second line, which I don't want

### Expected behaviour

No automatically added comment leader after the shebang line.

I suggest to change `comments` to `b:#` (file `runtime/ftplugin/sh.vim`) to fix this, just like it is in `runtime/ftplugin/python.vim`

### Version of Vim

9.0.1378

### Environment

OS: Debian 12
Terminal: Konsole
Shell: bash 5.2.15

### Logs and stack traces

_No response_

dkearns:
Add the 'b' flag to 'comments'.

Fixes #14101.

### DIFF
--- a/runtime/ftplugin/sh.vim
+++ b/runtime/ftplugin/sh.vim
@@ -4,7 +4,7 @@
 " Previous Maintainer:	Dan Sharp
 " Contributor:		Enno Nagel <ennonagel+vim@gmail.com>
 "			Eisuke Kawashima
-" Last Change:		2024 Jan 14
+" Last Change:		2024 Feb 27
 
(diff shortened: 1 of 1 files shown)


zeertzjq:
It should be runtime(sh)

chrisbra:
thanks.

PR #14103, opened 2024-02-27
runtime(vim): Update syntax file, improve :echo highlighting

dkearns:
Improve :echo highlighting.

- Normalise behaviour of :echo commands and improve expression matching.
- Allow continued argument lines.
- Refine string interpolation groups.
- Remove duplicated :menu and :map generated commands that are handled
  specially later in the file.


chrisbra:
thanks!

PR #14105, opened 2024-02-27
runtime(java): Recognise _when_ clauses in _switch_ blocks

zzzyxwvut:
Also:

- distinguish `yield` when used as a contextual keyword from  
  when used qualified as a method or a method reference (as  
  can be seen in `testdir/input/java_switch.java`, variables   
  and method declarations named `yield` will be recognised   
  as the namesake keyword--consider picking other names for  
  variables, and defining `g:java_highlight_functions` to have  
  method names painted; since `yield` statements can have  
  trailing parens, they must be recognised as statements,  
  for only qualified `yield` method calls are supported);

- recognise grouped `default` `case` labels;

- describe primitive types for `case` labels (JLS, §14.11,  
  §3.10.1);

- recognise some non-ASCII identifiers (see `javaLambdaDef`,  
  `javaUserLabel`) (further improvement for better recognition  
  of identifiers will be arranged in a separate PR).

Because the arrow `->` is used in two kinds of expressions,  
lambda (abstractions) and `switch`, necessary changes were  
made for the recognition of either (and further improvement  
touching lambda expressions will be separately arranged).

Because `default` is used for instance method declarations  
in interfaces and in `switch` labels, necessary changes were  
made for the recognition of either (and further improvement  
touching method declarations will be separately arranged).

Finally, it deemed appropriate to put `yield` in the syntax  
group of `javaOperator` rather than `javaStatement`, for its  
member `var` is also another contextual keyword (e.g., this  
is valid syntax: `var var = var(test.var);`).

References:
https://openjdk.org/jeps/361 (Switch Expressions)
https://openjdk.org/jeps/440 (Record Patterns)
https://openjdk.org/jeps/441 (Pattern Matching for switch)


zzzyxwvut:
During screen dump generation (`v9.1.0136`), I ran into  
a discontinuity between two screen dumps.  (There are four  
lines in the input file that exceed the 75 column limit and  
their spilled over tails are painted at the start of a next  
additional line for screen dumps: `00`, `01`; `03`.)

`testdir/failed/java_switch_03.dump` [BOTTOM 6 lines]:
```java
                        case Letters[] ll       -> "SwitchTests$1Letters[]"
;
                        default                 -> "java.lang.Object";
                });

                                                         55,3-17       46%
```

`testdir/failed/java_switch_04.dump` [TOP 6 lines]:
```java
                switch (ch) {
                case 'a':               { echo('a'); break; }
                case 'b':               { echo('b'); break; }
                default:                { echo('\u0000'); break; }
                };

```

`testdir/input/java_switch.java` [MIDDLE (6 * 2) + 2 lines]:
```java
		echo(switch (o) {
			case null		-> "null";
			case Letters[] ll	-> "SwitchTests$1Letters[]";
			default			-> "java.lang.Object";
		});

		char ch = 'c';

		switch (ch) {
		case 'a':		{ echo('a'); break; }
		case 'b':		{ echo('b'); break; }
		default:		{ echo('\u0000'); break; }
		};

```

The two lines missing at the top of `java_switch_04.dump`  
(commented out for clarity):
```java
//		char ch = 'c';
//
```

chrisbra:
Thanks. I see the problem. It's the many tabstops in the Java file. I'll include some additional changes that make use of ':set ts=4'.

I checked all the dumps manually and the vanished line is there now.

PR #14108, opened 2024-02-28
fix(echospace): don't subtract SHOWCMD_COLS if `showcmdloc` is statusline or tabline

Sam-programs:
The command is not visible in the message area when `showcmdloc` is statusline making echospace smaller than what it can be.

### DIFF
--- a/src/screen.c
+++ b/src/screen.c
@@ -4536,7 +4536,7 @@ comp_col(void)
 	if (!last_has_status)
 	    sc_col = ru_col;
     }
-    if (p_sc && STRCMP(p_sloc, "last") == 0)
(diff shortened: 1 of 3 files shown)


src/screen.c
```diff
@@ -4536,7 +4536,7 @@ comp_col(void)
 	if (!last_has_status)
 	    sc_col = ru_col;
     }
-    if (p_sc)
+    if (p_sc && STRCMP(p_sloc, "last") == 0)
```
zeertzjq:
```suggestion
    if (p_sc && *p_sloc == 'l')
```

chrisbra:
thank you!

Issue #13839, opened 2024-01-10
E340 & E315 by bufhidden=wipe and autocmd diffsplit


kawarimidoll:
### Steps to reproduce

I got E340 and E315 when I run the commands below (file names are not important):

```vim
edit a.txt
edit b.txt
setlocal bufhidden=wipe
autocmd BufLeave * diffsplit c.txt
bnext
quit
```

here is demo:

https://github.com/vim/vim/assets/8146876/41296bda-29dd-4e41-a9e5-a9518947f26e

when run `:bnext`, these two conflicting things seem to occur:

1. `bufhidden=wipe` wipes `b.txt`
1. `autocmd BufLeave * diffsplit c.txt` opens `c.txt` with keeping `b.txt` open

then buffers are mismatch and a strange buffer is displayed (you can see it at end of demo).

### Expected behaviour

handle these errors

### Version of Vim

Vim 9.1.0 (Compiled by Homebrew)

### Environment

OS: macOS Sonoma 14.2.1
Terminal: wezterm 20230712-072601-f4abf8fd
value of $TERM: xterm-256color
Shell: zsh 5.9 (arm-apple-darwin21.3.0)

### Logs and stack traces

```shell
E340: Internal error; if you can reproduce please report a bug
Press ENTER or type command to continue
E315: ml_get: Invalid lnum: 1
Press ENTER or type command to continue
```

chrisbra:
That is nasty. I am not sure why it doesn't crash for you, for me it crashes with a use-after-free for the buffer `b.txt`. Basically Vim doesn't expect that a another window is opened which contains a reference to the immediately later freed buffer `b.txt`. But since this is not the `curwin` it doesn't notice it when checking the side-effects of the BufLeave autocommand. 

I haven't yet figured out a way to prevent this.

PR #14109, opened 2024-02-28
Relax split restrictions when split-moving windows

seandewar:
Following up on [my comment here](https://github.com/vim/vim/pull/14042#discussion_r1493919216), it seems unnecessary to check `b_locked_split` when "split-moving" windows. It looks like the flag exists to stop new views into a buffer being created while it's closing (which can result in windows into a freed buffer remaining), but split-moving doesn't result in any new views into a buffer, unlike regular splitting.

The doc comment for `b_locked_split` also seems to support this:
https://github.com/vim/vim/blob/0c98a71236e6eca983fed5d903d07bd53190d16b/src/structs.h#L3002-L3003

So, continue to check `split_disallowed` when split-moving (it looks possibly relevant), but add a test to check that split-moving is OK when `b_locked_split` is set (I've made sure it crashes Vim if the `b_locked_split` check is reverted for regular splitting).

---

**Edit:** hmm, actually the crash I based the test on has the same backtrace as #13839, which wasn't what I was going for :laughing:.

I was hoping to base it on `Test_autocmd_normal_mess`, which is the test introduced with `b_locked_split`, but even that test doesn't seem to crash anymore when reverting the `b_locked_split` check for regular splitting...

**Edit 2:** actually, it seems to be a different issue (unrelated to 9.1.0143 or the issue it fixes); it seems `close_buffer` and its callers don't consider that autocmds can make another window view a buffer that's being wiped, without causing `b->b_nwindows > 1`.

The following crashes Vim with a heap UAF:

```vim
let w = win_getid()
sp
new
let b = bufnr()
au BufWipeout * ++once cal win_gotoid(w) | exe b 'b' | winc p
bw!
```

I'll try to address this at some point, unless someone beats me to it. :innocent:

### DIFF
--- a/src/evalwindow.c
+++ b/src/evalwindow.c
@@ -1001,8 +1001,8 @@ f_win_splitmove(typval_T *argvars, typval_T *rettv)
 	size = (int)dict_get_number(d, "size");
     }
 
-    // Check if we can split the target before we bother switching windows.
-    if (text_or_buf_locked() || check_split_disallowed(targetwin) == FAIL)
+    // Check if we can switch or move windows before we bother switching.
+    if (text_or_buf_locked() || check_split_disallowed(NULL) == FAIL)
 	return;
 
     if (curwin != targetwin)
--- a/src/testdir/test_window_cmd.vim
+++ b/src/testdir/test_window_cmd.vim
@@ -1181,18 +1181,9 @@ func Test_win_splitmove()
   call assert_equal(v:true, s:triggered)
   unlet! s:triggered
 
-  new
-  augroup WinSplitMove
-    au!
-    au BufHidden * ++once let s:triggered = v:true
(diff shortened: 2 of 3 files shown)


chrisbra:
> actually, it seems to be a different issue (unrelated to 9.1.0143 or the issue it fixes); it seems close_buffer and its callers don't consider that autocmds can make another window view a buffer that's being wiped, without causing b->b_nwindows > 1.

Yeah the patch I made only addresses BufLeave autocommands, however the same can probably happen for other autocommands. We should try to make this more central. 

Thanks

chrisbra:
this has stalled, so closing

seandewar:
No problem. I may revisit later as the `b_nwindows` bug that blocked this was likely fixed in #19728, but no ETA as I'm currently largely away from the computer due to an emergency. Best wishes in the meantime to all!

PR #14113, opened 2024-02-28
revert: vim@7f630e6 

pheiduck:


chrisbra:
Thanks for working on that. I have actually thought we should just revert 7f630e6f358275d0bce9fea2665a2 ? Would that work? It might be a bit simpler

pheiduck:
> Thanks for working on that. I have actually thought we should just revert [7f630e6](https://github.com/vim/vim/commit/7f630e6f358275d0bce9fea2665a2763edbf6a0e) ? Would that work? It might be a bit simpler

Yep, should be the easiest way :)

PR #14119, opened 2024-02-29
v:echospace wrong after setting invalid value to 'showcmdloc'

zeertzjq:
Problem:  v:echospace wrong after setting invalid value to 'showcmdloc'.
Solution: Only call comp_col() if value is valid.


chrisbra:
thanks!

PR #14123, opened 2024-03-01
STRLEN() changes part 2

basilisk0315:
Here is another batch of changes.


zeertzjq:
I think there is a problem. `ml_line_len` includes the size of text properties, so I think it may be necessary to rethink whether any of these changes (including ones in #14052) are correct.

chrisbra:
yes, that's a good point. However the tests looked good, are we missing something here?

zeertzjq:
If there are no tests that involve text properties when these places are reached, then no test fails, but the problen still exists.

chrisbra:
Yes, what I was trying to say, was that we are apparently missing a test that shows that the returned memline length should be counted without the attached text properties. 

Anyways, I'll have to reconsider to revert the previous strlen patch now

chrisbra:
I think it makes sense to store the actual line length excluding the text properties in addition to the text properties. Something like this:

```diff
diff --git a/src/memline.c b/src/memline.c
index bbc13d7ed..5af2ae67f 100644
--- a/src/memline.c
+++ b/src/memline.c
@@ -2703,7 +2703,7 @@ ml_get_buf_len(buf_T *buf, linenr_T lnum)
     if (*ml_get_buf(buf, lnum, FALSE) == NUL)
         return 0;

-    return buf->b_ml.ml_line_len - 1;
+    return buf->b_ml.ml_line_len2 - 1;
 }

 /*
@@ -2737,6 +2737,7 @@ ml_get_buf(
 errorret:
        STRCPY(questions, "???");
        buf->b_ml.ml_line_len = 4;
+       buf->b_ml.ml_line_len2 = buf->b_ml.ml_line_len;
        buf->b_ml.ml_line_lnum = lnum;
        return questions;
     }
@@ -2746,6 +2747,7 @@ errorret:
     if (buf->b_ml.ml_mfp == NULL)      // there are no lines
     {
        buf->b_ml.ml_line_len = 1;
+       buf->b_ml.ml_line_len2 = buf->b_ml.ml_line_len;
        return (char_u *)"";
     }

@@ -2796,6 +2798,7 @@ errorret:

        buf->b_ml.ml_line_ptr = (char_u *)dp + start;
        buf->b_ml.ml_line_len = end - start;
+       buf->b_ml.ml_line_len2 = buf->b_ml.ml_line_len;
        buf->b_ml.ml_line_lnum = lnum;
        buf->b_ml.ml_flags &= ~(ML_LINE_DIRTY | ML_ALLOCATED);
     }
@@ -3646,6 +3649,7 @@ ml_replace_len(

     curbuf->b_ml.ml_line_ptr = line;
     curbuf->b_ml.ml_line_len = len;
+    curbuf->b_ml.ml_line_len2 = len_arg + 1;
     curbuf->b_ml.ml_line_lnum = lnum;
     curbuf->b_ml.ml_flags = (curbuf->b_ml.ml_flags | ML_LINE_DIRTY) & ~ML_EMPTY;

diff --git a/src/structs.h b/src/structs.h
index 5b88260f3..e1852b854 100644
--- a/src/structs.h
+++ b/src/structs.h
@@ -800,7 +800,8 @@ typedef struct memline
 #define ML_ALLOCATED   0x10    // ml_line_ptr is an allocated copy
     int                ml_flags;

-    colnr_T    ml_line_len;    // length of the cached line, including NUL
+    colnr_T    ml_line_len;    // length of the cached line + textproperties, including NUL
+    colnr_T    ml_line_len2;   // length of the cached line, including NUL
     linenr_T   ml_line_lnum;   // line number of cached line, 0 if not valid
     char_u     *ml_line_ptr;   // pointer to cached line

```

What do you think?

zeertzjq:
I think that may still be wrong, because it assigns the same value to the two in `ml_get_buf()`. Meanwhile `ml_line_fulllen` and `ml_line_textlen` may be better names, and it may be better for the latter one to exclude the NUL to avoid the `- 1` at callsites.

chrisbra:
oh right. Should we revert it then? Or we will need to use this then https://github.com/chrisbra/vim/commit/84e32117e9bacb850a3216252158d1d2870eb824

basilisk0315:
I see 3 approaches here:
1: abandon the whole thing;
2: search backwards from the end of ml_line_ptr for the NUL to calculate the text length;
3: store text length (ex text property) separately in the buf_T struct.

Option 1 is easiest, but that is no fun ;-)
Option 2 has the advantage of being a small and concise change, but still requires scanning the text.
Option 3 means more surgery but will be quicker overall.

Cheers
John

zeertzjq:
> oh right. Should we revert it then? Or we will need to use this then [chrisbra@84e3211](https://github.com/chrisbra/vim/commit/84e32117e9bacb850a3216252158d1d2870eb824)

Since storing the text length in `struct data_block` is not backward-compatible, this may be the only solution. Although it's possible to compute `ml_line_textlen` on demand, i.e. set it to 0 when its value cannot be computed in O(1) time, and call `STRLEN()` when `ml_get_buf_len()` is called later.

zeertzjq:
I think this PR needs a rebase now.

basilisk0315:
Hi @zeertzjq how do I do a rebase? This is all new to me.

basilisk0315:
Thanks.
I think I have done it right.
How do I repush to the PR?

zeertzjq:
`git push --force`

basilisk0315:
Cool. I have done something wrong. I did git push --force and I got "everything up-to-date".

zeertzjq:
I guess merging master branch (as in this PR) is actually also fine.

Meanwhile there seems to be a mistake in `ml_replace_len()`: `len_arg` already includes the `NUL` if `has_props` is `TRUE`, so it should only add 1 to `len_arg` is `has_props` is `FALSE`:
```diff
diff --git a/src/memline.c b/src/memline.c
index 99b9e9f2e..522815526 100644
--- a/src/memline.c
+++ b/src/memline.c
@@ -3661,7 +3661,7 @@ ml_replace_len(
 
     curbuf->b_ml.ml_line_ptr = line;
     curbuf->b_ml.ml_line_len = len;
-    curbuf->b_ml.ml_line_textlen = len_arg + 1;
+    curbuf->b_ml.ml_line_textlen = len_arg + !has_props;
     curbuf->b_ml.ml_line_lnum = lnum;
     curbuf->b_ml.ml_flags = (curbuf->b_ml.ml_flags | ML_LINE_DIRTY) & ~ML_EMPTY;
 
```

chrisbra:
thanks. I think this is fine now. Let me include it

PR #14125, opened 2024-03-02
runtime(vim): Update base-syntax, improve :augroup highlighting

dkearns:
Improve :augroup highlighting
- Explicitly match the bang and group name in :aug! {name}.
- Allow any characters in a group name.
- Match default group switch marker, END.
- Match :aug without arguments (list command).

:autocmd improvements to come.


chrisbra:
thanks!

PR #14126, opened 2024-03-02
runtime(r,rhelp,rmd,rnoweb,rrst): Update ftplugin, browsefilter labels

dkearns:
Update browsefilter labels to use the standard format:
  `File Description (*.ext1, *.ext2, *.ext3)`

@chrisbra, I removed the list of changes attributed to "Vim Project".  They were primarily intended to notify temporarily absent maintainers and probably don't make much sense for unmaintained files.  I'll re-add them if you prefer.


chrisbra:
Thanks I think this is fine

PR #14127, opened 2024-03-02
ci.yml: downgrade codecov-action@v3

pheiduck:
added missing packages for codecov ci
```
warning - 2024-03-02 15:18:43,549 -- coverage.py is not installed or can't be found.
info - 2024-03-02 15:18:43,652 -- Found 0 coverage files to upload
Error: No coverage reports found. Please make sure you're generating reports successfully.
Warning: Codecov:
                      Failed to properly upload report: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -683,7 +683,7 @@ jobs:
       - name: Codecov
         timeout-minutes: 20
         if: matrix.coverage
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v3
         with:
           directory: src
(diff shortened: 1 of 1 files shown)


pheiduck:
Downgrade might be better as it’s not possible to solve all issues and it is not our problem as well. The action maintainers need to fix that.

chrisbra:
so why are we downgrading? What prolem does this fix?

pheiduck:
> so why are we downgrading? What prolem does this fix?

It can not generate any reports because a package is not installed on the action.
But we can still ignore this until upstream is fixing that.

PR #14129, opened 2024-03-03
Implement support for using builtin functions with vim9 objects

yegappan:
Add support for using builtin functions like string(), len() and empty() with Vim9 objects.
Allow object methods with the same name as the builtin functions. Compile the builtin method calls.


chrisbra:
Thanks. Is this the coninuation of #13238 ?

PR #14131, opened 2024-03-03
Getregion buffer support

Shougo:
Continues of https://github.com/vim/vim/pull/14090.

NOTE:  The buffer yank is executed on current window context.
NOTE: `{pos1}` and `{pos2}` must be same buffer.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4307,8 +4307,8 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the region is blockwise and it starts or ends in the
 		  middle of a multi-cell character, it is not included but
 		  its selected part is substituted with spaces.
-		- If and {pos1} and {pos2} are not in the same buffer, an
-		  empty list is returned.
+		- If {pos1} and {pos2} are not in the same buffer, an empty
+		  list is returned.
 		- {pos1} and {pos2} must be exists in tabs.
 
 		Examples: >
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4309,7 +4309,8 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		  its selected part is substituted with spaces.
 		- If {pos1} and {pos2} are not in the same buffer, an empty
 		  list is returned.
-		- {pos1} and {pos2} must be exists in tabs.
+		- {pos1} and {pos2} must belong to a buffer that is loaded in a
+		  window in any tab.
 
 		Examples: >
 			:xnoremap <CR>
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -4274,8 +4274,7 @@ getreginfo([{regname}])					*getreginfo()*
 			GetRegname()->getreginfo()
 
 getregion({pos1}, {pos2} [, {opts}])			*getregion()*
-		Returns the list of strings from {pos1} to {pos2} in current
-		buffer.
+		Returns the list of strings from {pos1} to {pos2} in buffer.
 
 		{pos1} and {pos2} must both be |List|s with four numbers.
 		See |getpos()| for the format of the list.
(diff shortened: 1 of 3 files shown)


runtime/doc/builtin.txt
```diff
@@ -4306,8 +4305,9 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the region is blockwise and it starts or ends in the
 		  middle of a multi-cell character, it is not included but
 		  its selected part is substituted with spaces.
-		- If {pos1} or {pos2} is not current in the buffer, an empty
-		  list is returned.
+		- If {pos1} or {pos2} is not same buffer, an empty list is
+		  returned.
```
```suggestion
		- If and {pos1} and {pos2} are not in the same buffer, an
		  empty list is returned.
```

runtime/doc/builtin.txt
```diff
@@ -4306,8 +4305,9 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the region is blockwise and it starts or ends in the
 		  middle of a multi-cell character, it is not included but
 		  its selected part is substituted with spaces.
-		- If {pos1} or {pos2} is not current in the buffer, an empty
-		  list is returned.
+		- If and {pos1} and {pos2} are not in the same buffer, an
+		  empty list is returned.
```
zeertzjq: Oops
```suggestion
		- If {pos1} and {pos2} are not in the same buffer, an empty
		  list is returned.
```

src/evalfunc.c
```diff
@@ -5507,12 +5510,25 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
+    buf = buflist_findnr(fnum1);
+    FOR_ALL_TABPAGES(tp)
+    {
+        FOR_ALL_WINDOWS(wp)
+        {
+            if (wp->w_buffer == buf)
+                break;
+        }
+    }
+    if (wp == NULL)
```
zeertzjq: It may be necessary to use current window if the specified bufnum is 0, as if the loop finds a non-current window, its 'list' or 'virtualedit' options may be different.

Shougo:
Hm...

I get the internal error when the code.

```vim
    #" using a mark in another window
    new
    VAR anotherbuf = bufnr()
    call setline(1, range(10))
    normal! GmA
    normal! GmB
    wincmd p
    call assert_equal([anotherbuf, 10, 1, 0], getpos("'A"))
    call assert_equal([], getregion(getpos("'B"), getpos("'A"), {'type': 'v' }))
    exe $':{anotherbuf}bwipe!'
```

Can you check the implementation?

src/evalfunc.c

runtime/doc/builtin.txt
```diff
@@ -4306,8 +4305,9 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		- If the region is blockwise and it starts or ends in the
 		  middle of a multi-cell character, it is not included but
 		  its selected part is substituted with spaces.
-		- If {pos1} or {pos2} is not current in the buffer, an empty
+		- If {pos1} and {pos2} are not in the same buffer, an empty
 		  list is returned.
+		- {pos1} and {pos2} must be exists in tabs.
```
chrisbra: `{pos1} and {pos2} must belong to a buffer that is loaded in a window in any tab`?

chrisbra:
Hm, not sure this an inprovement. This seems to work only with global/upper-case marks and if you have the buffer not loaded in  a window, it won't work.

Shougo:
> Hm, not sure this an inprovement. This seems to work only with global/upper-case marks and if you have the buffer not loaded in a window, it won't work.

Yes.  Because it is yank limitation.

runtime/doc/builtin.txt

chrisbra:
> Yes. Because it is yank limitation.

Hm, not sure. I think it might be simpler to just switch curbuf and possibly curwin similarly how the `getbufvar()` function does it. This should then also work for buffers not visible in any window. You then also don't have to iterate over all windows several times to find the buffer.

zeertzjq:
A problem here is that window options like 'list' and 'virtualedit' may affect the behavior.

Shougo:
OK.  I have fixed the errors.  The buffer yank is executed in current window context.

chrisbra:
thanks. I see another problem, if the buffer is not loaded, I get `E340: Internal error; if you can reproduce please report a bug`, I'll fix it while merging.

PR #14133, opened 2024-03-03
ml_get_buf_len() does not consider text properties

chrisbra:
Problem:  ml_get_buf_len() does not consider text properties
Solution: Store text length excluding text properties length
          in addition in the memline

related #14123

### DIFF
--- a/src/change.c
+++ b/src/change.c
@@ -1364,6 +1364,7 @@ del_bytes(
 	    mch_memmove(newp + newlen + 1, oldp + oldlen + 1,
 			       (size_t)curbuf->b_ml.ml_line_len - oldlen - 1);
 	curbuf->b_ml.ml_line_len -= count;
+	curbuf->b_ml.ml_line_textlen = (int)STRLEN(newp) + 1;
     }
 #endif
 
--- a/src/edit.c
+++ b/src/edit.c
@@ -5070,6 +5070,7 @@ ins_tab(void)
 			vim_free(curbuf->b_ml.ml_line_ptr);
 		    curbuf->b_ml.ml_line_ptr = newp;
 		    curbuf->b_ml.ml_line_len -= i;
+		    curbuf->b_ml.ml_line_textlen = (int)STRLEN(newp) + 1;
 		    curbuf->b_ml.ml_flags =
 			   (curbuf->b_ml.ml_flags | ML_LINE_DIRTY) & ~ML_EMPTY;
 		}
--- a/src/memline.c
+++ b/src/memline.c
@@ -2703,7 +2703,7 @@ ml_get_buf_len(buf_T *buf, linenr_T lnum)
     if (*ml_get_buf(buf, lnum, FALSE) == NUL)
         return 0;
 
-    return buf->b_ml.ml_line_len - 1;
+    return buf->b_ml.ml_line_textlen - 1;
 }
 
 /*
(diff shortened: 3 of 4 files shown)


chrisbra:
@zeertzjq  are you okay with this?

basilisk0315:
In ml_get_buf(), to make ml_line_textlen equal the text length, inc NUL, you should add 1 in the call to STRLEN() (which excludes the NUL).

The only time ml_line_textlen is actually used, in ml_get_buf_len(), 1 is subtracted. Might it be better to make ml_line_textlen just exclude the NUL?

zeertzjq:
Some changes seem also needed at places where `ml_line_len` is used in `change.c` and `edit.c`

basilisk0315:
I have been doing some research into the impact of this change (adding the call to STRLEN() to ml_get_buf()).
Throughout the source there are examples of calls to ml_get*() where the length of the line is not required. Some examples:
In change.c, see lines 1539 to 1544, 1562 to 1580 and 1638 to 1646.
In match.c, see lines 462 to 540.
In drawline.c see lines 1511, 1514, 1557.

In addition, I counted the number of times where ml_line_textlen is set in ml_get_buf() (inside the 'if (buf->b_ml.ml_line_lnum != lnum || mf_dont_release)') block. I did this my adding a static int to the block and incrementing and outputting it each time it is executed.

My results from my HP-UX build (note that the build has +textprop, but there are no actual text properties in play):
./vim memline.c
Startup gives 136 executions.
Type ‘/ml_line_textlen<CR>’ to get to line 2706, gives 28860 more.
Type 'n' to get to line 2740, gives 174 more.
Type 'n' to get to line 2750, gives 6314 more.
Type 'n' to get to line 2804, gives 359 more.
Type 'j' to get to line 2805, gives 128 more.
Type 'j' to get to line 2806, gives 128 more.
Type 'j' to get to line 2807, gives 128 more.

This is a total of 36227.

Doing some light editing quickly brings the total up to 50000+. The question as to why this function is called so many times would be interesting to answer.

If buf->b_has_textprop had been true, STRLEN() would have been called 50000+ times. Chances are that most of those calls would not have been needed.

Performing the same test on my windows gui build gives similar but higher numbers.

Given the goal was to reduce the number of STRLEN() calls, it seems to me like we have gone backwards with these changes.

Is this correct or am I missing something?

Thoughts?

Cheers
John

chrisbra:
yeah, sounds backwards. Perhaps we need to backout those changes then. I am not sure if the `buflen()` functions are then useful? Or should we do scanning using `strchr()` for the NULL byte?

zeertzjq:
I've mentioned one approach in https://github.com/vim/vim/pull/14123#issuecomment-1974937312

> set it to 0 when its value cannot be computed in O(1) time, and call STRLEN() when ml_get_buf_len() is called later.

zeertzjq:
> Or should we do scanning using `strchr()` for the NULL byte?

I don't think that'll work either from the start or the end. From the start it's no different than `STRLEN()`, and from the end it won't work because text properties can contain NUL bytes.

zeertzjq:
> I've mentioned one approach in [#14123 (comment)](https://github.com/vim/vim/pull/14123#issuecomment-1974937312)
> 
> > set it to 0 when its value cannot be computed in O(1) time, and call STRLEN() when ml_get_buf_len() is called later.

I've create a PR #14155 to do that.

PR #14134, opened 2024-03-03
translation(ru): Updated Russian translation for NSIS

RestorerZ:
translation(ru): Updated Russian translation for NSIS


chrisbra:
thanks!

Issue #14135, opened 2024-03-03
vim.vim syntax highlighting broken wrt system()


chdiza:
### Steps to reproduce

With 9.1.0148, have a vimscript file with just the following:
```
com Foo call system('ls')
```
and vim --clean to view it.

### Expected behaviour

I expect that the "system" part is not highlighted as an error.  But it is!  It's wrongly getting counted as `vimFunctionError`.

This wrong behavior is brand new to 9.1.0148.  9.1.0147 does not have this problem.

### Version of Vim

9.1.0148

### Environment

LinuxMint 21.1
Kitty

chdiza:
Sorry, I made a mistake. The bad commit is c4aef9a, not 9.1.0148.

9.1.0148 but before c4aef9a is fine.

PR #14136, opened 2024-03-03
runtime(vim): Update base-syntax, fix issue #14135

dkearns:
Fix incorrect error highlighting for function calls in :command definitions.
vimFunctionError should be restricted to :function header lines.

Fixes issue #14135.


chrisbra:
thanks!

PR #14137, opened 2024-03-04
runtime(vim): Update base-syntax, fix escaping :syn and :hi sub-groups

dkearns:
Fix escaping :syn and :hi sub-groups
    
- Remove contained :syntax and :highlight sub-groups from the function body cluster.  These should only match in the respective commands.
- Remove vimSynLine syntax group from several clusters.  The definition of vimSynLine was removed in Vim 5.3.

Use standard date format (YYYY MMM DD) in Last Change header comment.


runtime/syntax/generator/vim.vim.base
```diff
@@ -3,7 +3,7 @@
 " Maintainer:	Hirohito Higashi <h.east.727 ATMARK gmail.com>
 " 	Doug Kearns <dougkearns@gmail.com>
 " URL:	https://github.com/vim-jp/syntax-vim-ex
-" Last Change:	Mar 03, 2024
+" Last Change:	Mar 04, 2024
```
RestorerZ: Maybe we should use a standard date format for uniformity here, too?

dkearns: Done, thanks.

chrisbra:
thanks!

PR #14140, opened 2024-03-04
Fix 'winfixbuf' heap UAF/null pointer member access

seandewar:
Problem: qf_goto_win_with_qfl_file may check if prevwin has 'winfixbuf' set without checking if it's valid first.

Solution: fix the condition. Add a test, a modeline, and a missing CheckFeature.


chrisbra:
thanks

ColinKennedy:
Sorry for the `&&` mix-up. For future reference @seandewar, how did you discover / troubleshoot this? Is it findable with an optional sanitizer or did you find it the hard way?

seandewar:
I noticed this while skim-reading the patch, then wrote a test attempting to hit UBSAN which confirmed the issue.

(#14142 contains other things of interest after reviewing the patch closer, which included some classic trial and error)

Issue #12672, opened 2023-07-14
Bug in syntax coloring of Vim files


Aster89:
### Steps to reproduce

1. `vi -u NONE`,
2. `:syntax on`,
3. `:set syntax=vim`,
4. Type the following:
```vim
nnoremap <leader>foo :echo call(
  \ {x->x},
  \ ['foo'])<cr>

nnoremap <leader>bar :echo (
  \
  \ 'bar')<cr>
```
5. Move the cursor to either `)`,
6. `:echo map(synstack(line('.'), col('.')), 'synIDattr(v:val, "name")')[0]`.

It will print `vimOperError`.

### Expected behaviour

`vimMapRhsExtend`

### Version of Vim

9.0.1676

### Environment

Operating system: up-to-date ArchLinux
Terminal: urxvt
Value of $TERM: rxvt-unicode-256color
Shell: bash

PR #14141, opened 2024-03-04
runtime(vim): Update base-syntax, improve :map highlighting

dkearns:
,Improve :map command highlighting.

- Fix multiline RHS matching, allow continued lines and comments.
- Allow ^V-escaped whitespace in LHS.
- Handle map-bar properly and allow trailing commands.

Fixes issue #12672.


chrisbra:
thanks!

PR #13903, opened 2024-01-23
Add 'stickybuf' support

ColinKennedy:
Closes: https://github.com/vim/vim/issues/6445

# Summary
This PR adds a new "local to window" option called `'stickybuf'`. When a window has `:set stickybuf` applied, any command, mapping, etc like `:edit` that would normally change the window to look at another buffer will fail or have some other alternate behavior.

This PR impacts many Vim commands and mappings. To make reviewing easier, I've grouped affected commands and mappings into groups which I'll explain below.

## New Error
For any command / mapping that must fail, `E922` + a message is printed to the user.

# All `'stickybuf'` Cases
## Fail Or Allow With [!] (forceit)
Any command in this list may potentially change a window's current buffer. When `'stickybuf'` is present, these `:foo` commands will fail with an error message. Users can still call `:foo!` to force the command.

<details>
<summary>Click to expand</summary>

- `:Next`
- `:argdo`
- `:argedit`
- `:args`
    - Note: This jumps for the first args when called. Maybe `:args` should be
      in the "Allow Command But Don't Swap" section, instead?
- `:bNext`
- `:bfirst`
- `:blast`
- `:bmodified`
- `:bnext`
- `:bprevious`
- `:brewind`
- `:bufdo`
- `:buffer`
- `:djump`
- `:edit`
- `:enew`
- `:ex`
- `:find`
- `:first`
- `:ijump`
- `:lNext`
- `:lNfile`
- `:last`
- `:ldo`
- `:ll`
- `:llast`
- `:lnext`
- `:lnfile`
- `:lpfile`
- `:lprevious`
- `:lrewind`
- `:ltag`
- `:next`
- `:pop`
- `:previous`
- `:rewind`
- `:tNext`
- `:tag`
- `:tfirst`
- `:tjump`
- `:tlast`
- `:tnext`
- `:tprevious`
- `:trewind` - Note: No unittests
- `:view`
- `:visual`
- `:wNext`
- `:wnext`
- `:wprevious`
</details>

## Disabled Commands
These commands are completely disallowed and have no [!] alternative.

- `:drop`
    - @Maintainers - Would you ...
    - like to add [!] to this command so 'stickybuf' can be forced?
    - or split a new window?

## Disabled Mappings
These mappings are completely disallowed and have no alternative.

- `:normal \<C-]>`
- `:normal \<C-^>`
- `:normal \<C-t>`
- `:normal ]\<C-d>`
- `:normal ]\<C-i>`
- `:normal g\<C-]>`
[...]

PR #14142, opened 2024-03-05
Address some 'winfixbuf' things

seandewar:
Problem: some 'winfixbuf' issues exist, mostly relating to the quickfix list.

Solution: address them and adjust tests. Retab and reflow a few things too.

---

Things touched include:

- Replace the semsgs with gettext'd emsgs.

- Handle window switching in ex_listdo properly, so curbuf and curwin are kept in-sync and trigger autocommands; handle those properly.

- Don't change the list entry index in qf_jump_edit_buffer if we fail due to 'wfb' (achieved by returning FAIL; QF_ABORT should only be used if the list was changed).

- Make qf_jump_edit_buffer actually switch to prevwin when using `:cXX` commands **outside** of the list window if 'wfb' is set in curwin. Handle autocommands properly in case they mess with the list.

  NOTE: previously, it seemed to split if 'wfb' was set, but do nothing and fail if prevwin is *valid*. This behaviour seemed strange, and maybe unintentional? Now it aligns more with what's described for the `:cXX` commands in the original PR description when used outside a list window, I think.

- In both functions, only consider prevwin if 'wfb' isn't set for it; fallback to splitting otherwise.

- Use win_split to split. Not sure if there was a specific reason for using ex_splitview. win_split is simpler and respects modifiers like :vertical that may have been used. Plus, its return value can be checked for setting opened_window in qf code (technically win_split_ins autocmds could immediately close it or change windows, in which the qf code might close some other window on failure; it's already the case elsewhere, though).

### DIFF
--- a/src/ex_cmds2.c
+++ b/src/ex_cmds2.c
@@ -460,26 +460,33 @@ ex_listdo(exarg_T *eap)
 
     if (curwin->w_p_wfb)
     {
-        if ((eap->cmdidx == CMD_ldo || eap->cmdidx == CMD_lfdo) && !eap->forceit)
-        {
-            // Disallow :ldo if 'winfixbuf' is applied
-            semsg("%s", e_winfixbuf_cannot_go_to_buffer);
-            return;
-        }
+	if ((eap->cmdidx == CMD_ldo || eap->cmdidx == CMD_lfdo) &&
+		!eap->forceit)
+	{
+	    // Disallow :ldo if 'winfixbuf' is applied
+	    emsg(_(e_winfixbuf_cannot_go_to_buffer));
+	    return;
+	}
 
-        if (win_valid(prevwin))
-            // Change the current window to another because 'winfixbuf' is enabled
-            curwin = prevwin;
-        else
-        {
-            // Split the window, which will be 'nowinfixbuf', and set curwin to that
-            exarg_T new_eap;
-            CLEAR_FIELD(new_eap);
-            new_eap.cmdidx = CMD_split;
-            new_eap.cmd = (char_u *)"split";
-            new_eap.arg = (char_u *)"";
-            ex_splitview(&new_eap);
-        }
+	if (win_valid(prevwin) && !prevwin->w_p_wfb)
+	{
+	    // 'winfixbuf' is set; attempt to change to a window without it.
+	    win_goto(prevwin);
+	}
+	if (curwin->w_p_wfb)
+	{
+	    // Split the window, which will be 'nowinfixbuf', and set curwin to
+	    // that
+	    win_split(0, 0);
+
+	    if (curwin->w_p_wfb)
+	    {
+		// Autocommands set 'winfixbuf' or sent us to another window
+		// with it set.  Give up.
+		emsg(_(e_winfixbuf_cannot_go_to_buffer));
+		return;
+	    }
+	}
     }
 
 #if defined(FEAT_SYN_HL)
--- a/runtime/doc/message.txt
+++ b/runtime/doc/message.txt
@@ -122,7 +122,7 @@ wiped out a buffer which contains a mark or is referenced in another way.
 You cannot have two buffers with exactly the same name.  This includes the
 path leading to the file.
 
-							*E1513* >
+							*E1513*
   Cannot edit buffer. 'winfixbuf' is enabled ~
 
 If a window has 'winfixbuf' enabled, you cannot change that window's current
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -3146,7 +3146,8 @@ qf_goto_win_with_qfl_file(int qf_fnum)
 	    // Didn't find it, go to the window before the quickfix
 	    // window, unless 'switchbuf' contains 'uselast': in this case we
 	    // try to jump to the previously used window first.
-	    if ((swb_flags & SWB_USELAST) && win_valid(prevwin) && !prevwin->w_p_wfb)
+	    if ((swb_flags & SWB_USELAST) && win_valid(prevwin) &&
+		    !prevwin->w_p_wfb)
 		win = prevwin;
 	    else if (altwin != NULL)
 		win = altwin;
@@ -3158,7 +3159,8 @@ qf_goto_win_with_qfl_file(int qf_fnum)
 	}
 
 	// Remember a usable window.
-	if (altwin == NULL && !win->w_p_pvw && !win->w_p_wfb && bt_normal(win->w_buffer))
+	if (altwin == NULL && !win->w_p_pvw && !win->w_p_wfb &&
+		bt_normal(win->w_buffer))
 	    altwin = win;
     }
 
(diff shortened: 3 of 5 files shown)


src/ex_cmds2.c
```diff
@@ -460,26 +460,33 @@ ex_listdo(exarg_T *eap)
 
     if (curwin->w_p_wfb)
```
seandewar, line 461: Currently 'wfb' is checked for by `:windo` and such, where being able to switch buffers may not be 100% necessary. ~~Don't have much of an opinion on the matter, though.~~

Actually, it's a little unideal, as in the case of `:windo`, the split possibly created by the 'wfb' handling can cause a window to be skipped by `:windo` (as the command's range isn't updated). Not addressing this here, though.

(Also I'm pretty sure most of the stuff codecov is complaining about has test coverage...)

chrisbra:
Thanks for that! I think this looks good.

Shane-XB-Qian:
>       NOTE: previously, it seemed to split if 'wfb' was set, but do nothing
>       and fail if prevwin is *valid*. This behaviour seemed strange, and maybe
>       unintentional? Now it aligns more with what's described for the `:cXX`
>       commands in the original PR description when used outside a list window,
>       I think.
>     
>     - In both functions, only consider prevwin if 'wfb' isn't set for it;
>       fallback to splitting otherwise.

i opened a netrw somehow now it splitted to 5 windows,
i am not certainly sure this was the cause, and which part of my script was impacted,
but certainly it changed its previous workable script, could we not make up new amaze behavior?

-- 
shane.xb.qian

chrisbra:
netrw does not currently set 'winfixbuf', so you have set it yourself? It helps if you give precise reproduction steps.

Shane-XB-Qian:
> previously, it seemed to split if 'wfb' was set, but do nothing and fail if prevwin is *valid*.

could you try to keep that, kept do "nothing" and fail but Not failback to split?

// yes, I tried to set winfixbuf to varify if it seviced and not broken any script, but obviously this did.

--
shane.xb.qian

seandewar:
Without a repro there's nothing to go on here.

> i opened a netrw somehow now it splitted to 5 windows

This patch should only change behaviour if 'winfixbuf' is set, opening up at most **one split at any time**.
The behaviour also only applies to `:Xdo` commands and `:cX` commands for the quickfix list.
Unless they're being used, this patch is unlikely to be related.

> > previously, it seemed to split if 'wfb' was set, but do nothing and fail if prevwin is *valid*.
>
> could you try to keep that, kept do "nothing" and fail but Not failback to split?

The old behaviour would do nothing (and eventually fail with E1513) if there was no previous window for
the current window, but would still split to avoid 'winfixbuf' if there was.

That old behaviour seems wrong, and was likely unnoticed as no tests used `:cX` commands from **outside**
the quickfix list window (when used from inside, `qf_goto_win_with_qfl_file` is called which jumps to a different
window; likely why it wasn't noticed).

> could we not make up new amaze behavior?

Who is this directed to?

Shane-XB-Qian:
> > could we not make up new amaze behavior?
> 
> Who is this directed to?

you.
the patches of 'winbufbuf' before yesterday, it at least was not broken something in my view.
// as for this was correct or wrong, maybe others would complain (instead of me) if something broken their and previous classic plugins.

seandewar:
I would say the change is not "amaze" and instead fixed a bug. The previous behaviour would switch windows and fail with E1513 (without restoring the current window from before either).

Again, unless 'winfixbuf' is set and any of the commands previously mentioned were used, this patch should have no effect on you. Without a repro there isn't much else to be said.

Shane-XB-Qian:
> I would say the change is not "amaze" and instead fixed a bug. The previous behaviour would switch windows and fail with E1513 (without restoring the current window from before either).

"fail" was ok, and i supposed this 'winfixbuf' was intended to make it "fail" if set it.
// ~~seems you made the "fail" be showing nicer which only the msg but no stack, such change should be ok.~~
but "failback" was another topic.

> Again, unless 'winfixbuf' is set and any of the commands previously mentioned were used, this patch should have no effect on you.

this is interesting. 
"common sense": to vim plugins  (or vim itself) author maybe he would not use set winfixbuf such thing, but his users perhaps would, how/what you supposed he should handle such "broken"?
// to be clear, it did not effect on me, or IF i as one of official vim/plugins user, then it Do effect on me.

that's all.
// i would not use it "winfixbuf", so above comments please treat them as free check and some kind of alert in advance.

seandewar:
> "fail" was ok, and i supposed this 'winfixbuf' was intended to make it "fail" if set it.

For `:cX` commands, no; the original PR (#13903) stated those commands attempt to work even with 'wfb' set.

As for `:Xdo`, clearly the better behaviour would be to not change windows at all if prevwin had 'wfb' set, but that's
also not what the prior behaviour did.

> "common sense": to vim plugins (or vim itself) author maybe he would not use set winfixbuf such thing, but his users perhaps would, how/what you supposed he should handle such "broken"?

This is a discussion for #13903, which was already brought up previously there.

If you use 'winfixbuf', which is a lock-style option that _adds_ restrictions plugins may not have accounted for before,
then some breakage is inevitable; but this has little to do with _this_ patch.

There's still no clear indication that this patch is responsible for the change in behaviour you observed in netrw,
and neither have you provided any explicit confirmation that 'winfixbuf' was set and either `:cX` or `:Xdo` was used.

Unless you can provide a repro or some other confirmation that _this_ patch is responsible, there is nothing more to say.

Shane-XB-Qian:
> Also, the issue of possibly breaking existing plugins was brought up already in #13903.

yea, that's too, tho not sure exactly, but generally new win/buf flag would make previous plugins be harder.

i can confirm mostly this patch made the behavior change, 
// since this is the only merged change at yesterday for winfixbuf, so i thought this is the cause.
// and should be no related to qf (tho it maybe same situation), but i am gussing it was ':Xdo'.

if this cannot give your some hint, you can wait some days later, some others people may jump in.

seandewar:
> // and should be no related to qf (tho it maybe same situation), but i am gussing it was ':Xdo'.

Assuming 'wfb' _is indeed_ set, then this may be useful information.
There's three possible changes in this patch which could cause a change in behaviour for plugins:

1. if curwin has 'wfb' set, the switch to prevwin now acts normally; thus it may fire autocommands

2. previously, autocommands fired immediately after a switch to prevwin could observe `bufnr() != winbufnr(winnr())` and other inconsistent state, which was a bug; but maybe in some strange case fixing this caused a change in behaviour for plugins

3. if curwin has 'wfb' set, **at most one** split may happen to get a 'nowfb' window **only if** switching to the prevwin was impossible (e.g: there was none, or it also has 'wfb' set), _or_ if we switched to the prevwin, but autocommands set 'wfb' in it after we switched to it, or sent us to a different window with 'wfb' set in it

All possibilities _require 'wfb' to be set_, and involve autocommands, so I would look for a repro in that direction.

Number 2 is in general very unlikely to be related.
Number 3 is unlikely to be related if you previously were not seeing E1513 before this patch.

- If 1 is responsible, maybe blocking autocommands could be considered when switching to prevwin; not necessarily a fan of this, as it makes it act less like a normal window switch.

- If 2 is responsible, then not much can be changed here, as the previous behaviour was a bug.

- If 3 is responsible, we could consider not attempting a split, but instead attempt no switch at all to prevwin (unlike the previous behaviour) if it also has 'wfb' set. For the sake of the intended convenience of `:cX` from the original PR description, this change can be limited to just `:Xdo` stuff.

  If 3 is indeed responsible, implementing this solution seems like the most reasonable compared to the others, as splitting in general can create [unideal behaviour in `:windo`](https://github.com/vim/vim/pull/14142#discussion_r1511980450) (though I question whether checking 'wfb' ought to be done for `:Xdo` commands unrelated to `:bufdo`, `:argdo`, etc. at all, which it currently is).

chrisbra:
Sorry I find it hard to follow. Please create a new issue with *clear* reproduction steps and *clear*  summary what you expected instead. Guessing around here is not really helpful.

Thanks.

Shane-XB-Qian:
this was a some kind of alert, not a bug report, 
a.k.a it's on your thinking or re-think that you merged feat was ok or a broken.

Shane-XB-Qian:
--> https://github.com/vim/vim/issues/14301

seandewar:
> --> #14301

Thanks, but that's not a regression from this PR. It's also reproducible in [v9.1.0147](https://github.com/vim/vim/commit/215703563757a4464907ead6fb9edaeb7f430bea).

Note that I already documented that `:windo` behaviour [here](https://github.com/vim/vim/pull/14142#discussion_r1511980450) as not being ideal. It's something I'll consider addressing later when I have some free time, unless somebody beats me to it.

ColinKennedy:
@seandewar is right, the split behavior already existed before his PR was included. Let's concentrate replies in https://github.com/vim/vim/issues/14301 so we can keep chats about this all in one place please :)!

Shane-XB-Qian:
:shrug:

--
shane.xb.qian

PR #14144, opened 2024-03-05
Check SHM_FILEINFO when autoread

Shougo:
When autoread, the file info message is displayed.
But it is not disabled when `set shortmess+=F`.

I have fixed it.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -7253,7 +7253,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 	  q	use "recording" instead of "recording @a"		*shm-q*
 	  F	don't give the file info when editing a file, like	*shm-F*
 		`:silent` was used for the command; note that this also
-		affects messages from autocommands
+		affects messages from autocommands and 'autoread' reloading.
 	  S	do not show search count message when searching, e.g.	*shm-S*
 		"[1/5]"
 
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -4344,7 +4344,7 @@ buf_check_timestamp(
 	    }
 	    else
 	    {
-		if (!autocmd_busy)
+		if (!autocmd_busy && !shortmess(SHM_FILEINFO))
 		{
 		    msg_start();
 		    msg_puts_attr(tbuf, HL_ATTR(HLF_E) + MSG_HIST);
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -4294,7 +4294,7 @@ buf_check_timestamp(
 #endif
     }
 
-    if (mesg != NULL)
+    if (mesg != NULL && !shortmess(SHM_FILEINFO))
     {
 	path = home_replace_save(buf, buf->b_fname);
 	if (path != NULL)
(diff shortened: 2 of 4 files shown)


chrisbra:
Thanks. Can we have a test please?

Shougo:
I have added the tests.  But it is incomplete.  Can you fix it?

Shougo:
I have fixed the tests.

The reload dialog is disabled when `SHM_FILEINFO` is set.

chrisbra:
thanks!

zeertzjq:
I don't think this change is correct. After this change :checktime doesn't do anything of 'shortmess' contains 'F'.

Shougo:
Hm....

chrisbra:
@Shougo can you please check it and propose a follow up patch?

Shougo:
Yes.  I will do it.

Shane-XB-Qian:
@zeertzjq thanks, you helped the prod quality, and avoid unnecessary argue later.

Shougo:
Done.

zeertzjq:
Hmm, actually my previous comment isn't very accurate: this patches only causes :checktime to not do anything if the file didn't previously exist, and it is created while open in a buffer in Vim.

Nevertheless, I don't this think patch achieve the behavior it is described to achieve. It doesn't hide the file info message at all. Instead it hides the 'autoread' warning message and dialog, which isn't what 'shortmess' "F" flag is meat to hide.

Therefore, I think this patch should be reverted.

zeertzjq:
I think the actual change should be like this:
```diff
diff --git a/src/fileio.c b/src/fileio.c
index 344c819e7..ad09c5a22 100644
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -4294,7 +4294,7 @@ buf_check_timestamp(
 #endif
     }
 
-    if (mesg != NULL && !shortmess(SHM_FILEINFO))
+    if (mesg != NULL)
     {
 	path = home_replace_save(buf, buf->b_fname);
 	if (path != NULL)
@@ -4489,8 +4489,14 @@ buf_reload(buf_T *buf, int orig_mode, int reload_options)
 
 	if (saved == OK)
 	{
+	    int old_msg_silent = msg_silent;
+
 	    curbuf->b_flags |= BF_CHECK_RO;	// check for RO again
 	    keep_filetype = TRUE;		// don't detect 'filetype'
+
+	    if (shortmess(SHM_FILEINFO))
+		msg_silent = 1;
+
 	    if (readfile(buf->b_ffname, buf->b_fname, (linenr_T)0,
 			(linenr_T)0,
 			(linenr_T)MAXLNUM, &ea, flags) != OK)
@@ -4521,6 +4527,8 @@ buf_reload(buf_T *buf, int orig_mode, int reload_options)
 		    u_unchanged(curbuf);
 		}
 	    }
+
+	    msg_silent = old_msg_silent;
 	}
 	vim_free(ea.cmd);
 
```
which doesn't pass the test added here.

zeertzjq:
Using `Screenline()` seems to work. I've create a PR #14159.

PR #14145, opened 2024-03-05
runtime(c): Recognize "__inline"

adah1972:
`__inline` is recognized by both GCC and MSVC, so it is preferred in some projects and should probably be recognized.

(MSVC does not recognize `__inline__`.)


chrisbra:
thanks!

PR #14147, opened 2024-03-06
Text properties corrupted with fo+=aw and backspace

zeertzjq:
Problem:  Text properties corrupted with fo+=aw and backspace.
Solution: Allocate line and move text properties.


chrisbra:
thanks!

PR #14150, opened 2024-03-06
runtime(java): Recognise string templates

zzzyxwvut:
As this is encouraged in the referenced JEPs, "to visually  
distinguish a string template from a string literal, and  
a text block template from a text block", the default  
colours for `java\%[Debug]StrTempl` are made distinct from  
`java\%[Debug]String`.

According to §3.2 Lexical Translations (JLS, c. 1996 or any  
more recent version), line terminators, white space, and  
comments are discarded before tokens are accepted. Since  
a template expression comprises a template processor, a dot,  
and a template, it may be visually appealing to break up  
its head across a few lines whenever its tail already spans  
multiple lines.  Curiously, no allowance for it is made in  
the distributed tests for OpenJDK 21; the proposed regexp  
patterns take in consideration a line terminator and white  
space after a dot.

References:
https://openjdk.org/jeps/430 (Preview)
https://openjdk.org/jeps/459 (Second Preview)
https://openjdk.org/jeps/465


zzzyxwvut:
@chrisbra, what do think about [a fix](https://github.com/zzzyxwvut/vimvim/tree/runtime/syntax/dumps/wip) that should prevent the  
lines go missing in screen dumps generated from input test  
files with long lines?

chrisbra:
I had a slightly simpler approach in mind:
```
diff --git a/runtime/syntax/testdir/runtest.vim b/runtime/syntax/testdir/runtest.vim
index a2cdef372..57798a8de 100644
--- a/runtime/syntax/testdir/runtest.vim
+++ b/runtime/syntax/testdir/runtest.vim
@@ -181,7 +181,16 @@ func RunTest()
       let topline = 1
       let nr = 1
       while linecount - topline > 20
-       let topline += 18
+        " the terminal window has 20 lines - 1 for the command line
+        " line('w$') should be 19
+       "let topline += 18
+       call term_sendkeys(buf, printf(":echo line('w$')\<cr>"))
+        if (nr == 1)
+            let topline = term_getline(buf, 20) + 0
+        else
+            " account for scrolloff value
+            let topline = term_getline(buf, 20) + 5
+        endif
```

But this is slow as hell and it's still off sometimes. Perhaps we can use `line('w$')` instead of your long function and use your file IPC idea to make this work?

zzzyxwvut:
The thing with `line('w$')` is that it reports the position of  
the last visible _file line_ and not any "extra" line(s)  
produced by wrapping over (and it matters if there are more  
screenfuls to visit).  This is very much like walking lines  
with (`G`, `j`, `k`) vs (`gj`, `gk`).

For example, below are "two" 76 character lines; adjust Vim  
so that they wrap.

```
...........................................................................
1
...........................................................................
2
```

We are looking for `4`, but will get `2`:

```vim
:normal Ggj
:echo line('w$')
```

I've also run into off-by-some jumps with my own "solution":

```vim
call term_sendkeys(buf, '18gjzt')

```

when cursor travels from a line within a wrapped line, count  
can be of a greater or less value than `18`.

Let's continue exploring the problem in a dedicated issue.  
I'll file one in a day or two.

chrisbra:
okay, let's merge it now and keep the improvement for the syntax test scrolling for a separate PR then. Thanks

PR #14152, opened 2024-03-06
Make 'wfb' failing to split still report E1513, fix CI failures

seandewar:
Problem: may not be clear why failing to split causes an `:Xdo` command to abort if 'wfb' is set.

Solution: do not return immediately if win_split fails, so E1513 is still given. Expect both errors in the test.

A follow up to my comment [here](https://github.com/vim/vim/commit/af7ae8160041e2d17c56945381e9370e7178e596#r139468883); hopefully this is OK. :+1:

Also fix the failures from testing test_winfixbuf on the CI. (Don't mind the one flaky test_normal.vim failure :innocent:)

### DIFF
--- a/src/testdir/test_winfixbuf.vim
+++ b/src/testdir/test_winfixbuf.vim
@@ -479,6 +479,9 @@ endfunc
 
 " Fail :browse edit but :browse edit! is allowed
 func Test_browse_edit_fail()
+  " A GUI dialog may stall the test.
+  CheckNotGui
+
   call s:reset_all_buffers()
 
   let l:other = s:make_buffer_pairs()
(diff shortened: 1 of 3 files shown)


src/testdir/test_winfixbuf.vim
```diff
@@ -2490,8 +2508,8 @@ func Test_previous()
   call assert_equal(l:first, bufnr())
 endfunc
 
-" Fail pydo if it changes a window with 'winfixbuf' is set
-func Test_python_pydo()
+" Fail pyxdo if it changes a window with 'winfixbuf' is set
```
zeertzjq, line 2511:
```suggestion
" Fail pyxdo if it changes a window with 'winfixbuf' set
```

src/testdir/test_winfixbuf.vim
```diff
@@ -2523,8 +2541,8 @@ EOF
   unlet g:_previous_buffer
 endfunc
 
-" Fail pyfile if it changes a window with 'winfixbuf' is set
-func Test_python_pyfile()
+" Fail pyxfile if it changes a window with 'winfixbuf' is set
```
zeertzjq, line 2544:
```suggestion
" Fail pyxfile if it changes a window with 'winfixbuf' set
```

src/testdir/test_winfixbuf.vim
```diff
@@ -487,18 +490,31 @@ func Test_browse_edit_fail()
...
 
-  browse edit! other
-  call assert_equal(l:other, bufnr())
+  try
+    browse edit! other
+    call assert_equal(l:other, bufnr())
+  catch /E338:/
+    " Ignore E338, which occurs if console Vim is built with +browse.
+    " Console Vim without +browse will treat this as a regular :edit.
+  endtry
```
seandewar, line 499: Hoped to use a dummy FileExplorer for this instead, but for some reason that caused the 2 Windows VIMDLL jobs to fail... :shrug:

chrisbra:
thank you!

chrisbra:
oh yes, thanks for that

PR #14154, opened 2024-03-06
Duplicate assignment in f_getregion()

zeertzjq:
Problem:  Duplicate assignment in f_getregion().
Solution: Remove the duplicate assignment.  Also improve getregion()
          docs wording and fix an unrelated typo.


chrisbra:
thanks!

PR #14155, opened 2024-03-06
Calling STRLEN() to compute ml_line_textlen when not needed

zeertzjq:
Problem:  Calling STRLEN() to compute ml_line_textlen when not needed.
Solution: Use 0 when STRLEN() will be required and call STRLEN() later.

### DIFF
--- a/src/change.c
+++ b/src/change.c
@@ -1364,7 +1364,7 @@ del_bytes(
 	    mch_memmove(newp + newlen + 1, oldp + oldlen + 1,
 			       (size_t)curbuf->b_ml.ml_line_len - oldlen - 1);
 	curbuf->b_ml.ml_line_len -= count;
-	curbuf->b_ml.ml_line_textlen = (int)STRLEN(newp) + 1;
+	curbuf->b_ml.ml_line_textlen = 0;
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


chrisbra:
might be worth a try. I guess it depends how often `ml_get_buf_len()` is actually called and how often we have to re-calculate the `strlen()` then. Let's see if @basilisk0315 can check the call to strlen for this pr here.

basilisk0315:
@chrisbra Sorry for being a slow. What is it you want me to do?

chrisbra:
@basilisk0315 you have been doing some analysis on how often `strlen()` was performed. can you compare against this PR please?

basilisk0315:
I have had a look at this PR and I think it looks good. The strlen() in ml_get_buf_len() will only be called if the buffer actually has text properties which is then the only reliable way to determine the string's length. If the buffer has text properties the behaviour will be back to what it was before these changes (ie calling strlen()).

A small optimisation could be to change line 2806 in memline.c to this:
if (buf->b_has_textprop **&& buf->b_ml.ml_line_len >= sizeof(textprop_T)**)

This would mean we would only be calling strlen() on lines where there is a possibility of text props.

It's a shame the text props don't have a record of how many there on the line. I have a couple of ideas of how this info could be stored in the line. This would probably require some relatively invasive surgery which might not be palatable and so might not be worth pursuing. I don't know.

Overall I think the change is a good solution.

zeertzjq:
> A small optimisation could be to change line 2806 in memline.c to this: if (buf->b_has_textprop **&& buf->b_ml.ml_line_len >= sizeof(textprop_T)**)

It can be `>` actually, because there is always a NUL byte.

chrisbra:
thanks. Let me merge it then.

PR #14159, opened 2024-03-07
'shortmess' "F" flag doesn't work properly with 'autoread'

zeertzjq:
Problem:  'shortmess' "F" flag doesn't work properly with 'autoread'.
Solution: Hide the file info message instead of the warning dialog.

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -4294,7 +4294,7 @@ buf_check_timestamp(
 #endif
     }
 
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks.

It's interesting, `msg_silent` is usually set to `1` to enable but resetting it sometimes it's `0` and sometimes reset to `FALSE` ...

Anyhow, something for some later clean up.

PR #14161, opened 2024-03-08
expand() removes slash after env variable that ends with colon

zeertzjq:
Problem:  expand() removes a slash after an environment variable that
          ends with a colon on Windows.
Solution: Check the correct char for a colon.

### DIFF
--- a/src/testdir/test_expand.vim
+++ b/src/testdir/test_expand.vim
@@ -51,6 +51,19 @@ func Test_expand_tilde_filename()
   bwipe!
 endfunc
 
+func Test_expand_env_pathsep()
+  let $FOO = './foo'
+  call assert_equal('./foo/bar', expand('$FOO/bar'))
+  let $FOO = './foo/'
+  call assert_equal('./foo/bar', expand('$FOO/bar'))
+  let $FOO = 'C:'
+  call assert_equal('C:/bar', expand('$FOO/bar'))
+  let $FOO = 'C:/'
+  call assert_equal('C:/bar', expand('$FOO/bar'))
+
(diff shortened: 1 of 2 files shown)


src/testdir/test_expand.vim
```diff
@@ -51,6 +51,19 @@ func Test_expand_tilde_filename()
...
 
+func Test_expand_env_pathsep()
+  let $FOO = './foo'
+  call assert_equal('./foo/bar', expand('$FOO/bar'))
+  let $FOO = './foo/'
+  call assert_equal('./foo/bar', expand('$FOO/bar'))
+  let $FOO = 'C:'
+  call assert_equal('C:/bar', expand('$FOO/bar'))
+  let $FOO = 'C:/'
+  call assert_equal('C:/bar', expand('$FOO/bar'))
```
chrisbra, line 62: that is surprising. Shouldn't this return `C://bar`?

zeertzjq: Maybe, but it's always been like this.

chrisbra: On Unix like operating systems, it shouldn't really matter, as several repetitive directory separators are allowed and will be merged to the one single directory separator anyhow. Not sure what is the behavior  on Windows, e.g. does opening all this result in the same file:  `c:\foo.txt` and `c:\\foo.txt`  and `c:\\\foo.txt` (or with `:set shellslash` set, `c:/foo.txt`, `c://foo.txt`, `c:///foo.txt`).

so we are not changing behaviour here, then I guess it is fine to merge then.

PR #14162, opened 2024-03-08
runtime(vim): Update base-syntax, improve :menu{,translate} highlighting

dkearns:
Improve :menu and :menutranslate highlighting.

- Match args to :menutranslation and :popup.
- Only highlight special notation in {rhs} of :menu, like :map.
- Allow line continuations in {rhs} of :menu and between {english} and
  {mylang} of :menutranslation, matching common usage.
- Highlight :popup {name}.
- Bug fixes.


chrisbra:
thanks!

PR #14163, opened 2024-03-08
msgfmt ver. 0.22 forcibly converts text to UTF-8

RestorerZ:
Problem: Starting from version 0.22, msgfmt forcibly converts text to UTF-8 regardless of the value of the "charset" field
https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=NEWS;h=a525eb62b59095b6fb251009314146e127eb5d8d;hb=31f092d6b15c5b5086307245e43ea2b89b3bd410#l70
Solution: Check the version number and set the switch "--no-convert" for version 0.22


RestorerZ:
The same is required for "Make_cyg.mak" and "Make_ming.mak".
Unfortunately, I am not very familiar with these systems.

chrisbra:
Thanks, but the change to `src/po/Makefile` breaks for BSD builds (see Circle-CI [here](https://github.com/vim/vim/pull/14163/checks?check_run_id=22453125723)), because `ifeq` is not compatible with BSD Make.

I tried this patch instead:
```patch
diff --git a/src/po/Makefile b/src/po/Makefile
index b1144ba68..03aaa532c 100644
--- a/src/po/Makefile
+++ b/src/po/Makefile
@@ -20,7 +20,8 @@ SED = LANG=C sed
 # tools 0.10.37, which use a slightly different .po file format that is not
 # compatible with Solaris (and old gettext implementations) unless these are
 # set.  gettext 0.10.36 will not work!
-MSGFMTCMD = OLD_PO_FILE_INPUT=yes $(MSGFMT) -v
+MSGFMTC != if $(MSGFMT) --help |grep -q -- --no-convert ; then echo msgfmt -v --no-convert; else echo msgfmt -v; fi
+MSGFMTCMD = OLD_PO_FILE_INPUT=yes $(MSGFMTC)
 XGETTEXT = OLD_PO_FILE_INPUT=yes OLD_PO_FILE_OUTPUT=yes xgettext
 MSGMERGE = OLD_PO_FILE_INPUT=yes OLD_PO_FILE_OUTPUT=yes msgmerge

```

But now this breaks on the Apple Github Runners, because their Make version is only v3.81 (from 2006), which the `!=` assignment needs a Gnu Make 4.0 from 2013.

We could install Gnu Make on the Mac OS runners using this patch in addition:
```patch
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index d2148f7f1..953f07bec 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -361,11 +361,16 @@ jobs:
             sed -i.bak -f ci/config.mk.clang-12.sed ${SRCDIR}/auto/config.mk
           fi

+      - name: Install diffutils for testing, update GNU Make to v4, since we are using != assignment in src/po/Makefile
+        run: |
+          # Apple diff is broken. Use GNU diff instead. See #14032.
+          brew install diffutils make
+
       - name: Build
         env:
           LC_ALL: C
         run: |
-          make -j${NPROC}
+          gmake -d -j${NPROC}

       - name: Check version
         run: |
@@ -373,10 +378,6 @@ jobs:
           "${SRCDIR}"/vim -u NONE -i NONE --not-a-term -esNX -V1 -S ci/if_ver-1.vim -c quit
           "${SRCDIR}"/vim -u NONE -i NONE --not-a-term -esNX -V1 -S ci/if_ver-2.vim -c quit

-      - name: Install packages for testing
-        run: |
-          # Apple diff is broken. Use GNU diff instead. See #14032.
-          brew install diffutils

       - name: Test
         timeout-minutes: 20
```

Not sure. @ychin what's you opinion on this please?

RestorerZ:
@chrisbra , thank you for your comments!   

> but the change breaks for BSD builds

> But now this breaks on the Apple Github Runners

The wonderful, marvelous world of *nix. But, in principle, as far as Windows is concerned, it is no better. ;-)

If I remember the documentation correctly, then the assignment is "`!=`" this is the same as "`$(shell)`".
I'll try to rewrite your proposed fix later, let's see how it works.

RestorerZ:
@chrisbra , please take a look.
I checked in Ubuntu 22 with GNU make 3.81, GNU make 4.3, msgfmt 0.21 and FreeBSD 14, msgfmt 0.22.3.
It was done, of course, in an ugly way, but it didn't work out differently to come up with something to work everywhere.

chrisbra:
Thanks, yeah this is not nice. Let's wait a few days, if anybody knows some better ways around it.

RestorerZ:
Yes, of course. I really hope that the experts will suggest a good way.  

I have no doubt that can somehow get out of it, but you need to know both the makefile in BSD and GNU well. Unfortunately, I don't know them at all.

k-takata:
```
  70 * Portability:
  71   - On systems with musl libc, the *gettext() functions in libc now work
  72     with MO files generated from PO files with an encoding other than UTF-8.
  73     To this effect, the msgfmt program now converts the messages to UTF-8
  74     encoding before storing them in a MO file.  You can prevent this by
  75     using the msgfmt --no-convert option.
```
Just a question. Does msgfmt converts the messages to UTF-8 even on non-musl libc systems?

RestorerZ:
@k-takata , unfortunately, yes.

This is ja.euc-jp.mo
euc-jp mo

This is ru.cp1251.mo
cp1251-ru mo

RestorerZ:
@chrisbra , look at this.
In the current version, as I understand the principles of make, the following happens:
BSD make does not know "`$(shell)`" and, when it encounters it in the assignment of a variable, ignores it, the variable is not initialized. Then he sees the operator "`!=`", knows it and initializes the variable.
GNU make 3.81 knows "`$(shell)`", and initializes the variable. Then he sees the operator "`!=`", but does not know it and skips it, but the variable has already been assigned.
But GNU make 4.3 knows both the function "`$(shell)`" and the operator "`!=`" and we get double code execution. This, as you know, is bad. Very bad.
What if we do this then:
```
MSGFMTC != if $(MSGFMT) --help |grep -q -- --no-convert ; \
	then echo $(MSGFMT) -v --no-convert ; else echo $(MSGFMT) -v ; fi
MSGFMTC ?= $(shell if $(MSGFMT) --help |grep -q -- --no-convert ; \
	  then echo $(MSGFMT) -v --no-convert ; else echo $(MSGFMT) -v ; fi)
```
BSD make and GNU make 4.3 know the operator "`!=`" and perform variable assignment. Then they see the operator "`?=`", but the variable already has a value and they skip the assignment.
GNU make 3.81 does not know "`!=`" and skips variable assignment. Does he see further "`?=`" with the function "`$(shell`)" and performs the assignment of the variable, because it has not yet been assigned.
This is also not very good, but in any case, we will get rid of double code execution.

RestorerZ:
@chrisbra , what do you say about this:
```makefile
.po.mo:
	if $(MSGFMT) --help |grep -q -- --no-convert ; \
		then \
		$(MSGFMTCMD) --no-convert -o $@ $< ; \
		else \
		$(MSGFMTCMD) -o $@ $< ; \
	fi
```
Of course, it's better when the presets are set in advance, but it's not working yet.
If there is no normal way to define a variable so that it works both in BSD and with different versions of GNU make, then let's just make code that works both ways.
In my opinion, this is the best way so far.
Consider having separate makefiles for BSD, though.

RestorerZ:
> Hm, there may be another option. Let Configure take care of this:

Oh! Yes!
This is probably the most versatile and best solution.
Thank you, @chrisbra !

RestorerZ:
> ```diff
> +# MSGFMTCMD = OLD_PO_FILE_INPUT=yes $(MSGFMT) -v
> ```

Only this line should be removed.
We already have it in the configuration file.

ychin:
> We could install Gnu Make on the Mac OS runners using this patch in addition:
> Not sure. @ychin what's you opinion on this please?

Sorry for the late response, but I think that question is no long relevant from skimming the thread?

Either way, requiring a newer version of Make means everyone who builds Vim/MacVim locally on a Mac will now need to install Make manually. It's a rather large change in requirement IMO. Currently, in order to build Vim on macOS, all you need to do is to run `./configure && make` using system builtin tools.

RestorerZ:
> will now need to install a separate version Make manually.

> Currently, in order to build Vim on macOS, all you need to do is to run `./configure && make` using system builtin tools.

Yes, I agree with you completely. There is no need to create additional complications for no good reason.
But now, thanks to @chrisbra , it's not necessary.

PR #14165, opened 2024-03-09
Add support for using a class type of itself in an object method

### LINKED ISSUE #12369: Cannot use class type of itself in the method

yegappan:
Create the script-local variable for the class before parsing the contents of the class.  Fixes #12369.

Also, fix a regression issue introduced by 627c950e5b30503a084bd0092f9cb2a5fe73fe87 (9.1.0091)
in using valgrind for running the tests.

### DIFF
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -3830,7 +3830,7 @@ set_var(
  * If the variable already exists and "is_const" is FALSE the value is updated.
  * Otherwise the variable is created.
  */
-    void
+    int
 set_var_const(
     char_u	*name,
     scid_T	sid,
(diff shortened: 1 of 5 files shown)


chrisbra:
thanks!

PR #14167, opened 2024-03-09
Add XDG_CONFIG_HOME/vim/vimrc support

### LINKED ISSUE #2034: [RFC] .config/vim(UNIX), AppData/Roaming/vim(Windows) directory structure
Recently, many application store there configuration files into `~/.config/xxx`. This is very useful to take-out configuration files into USB-disk or dropbox. As far as I know, many users make symbolic links into `~/Dropbox/config/vim/vimrc` from `~/.vimrc` or `~/.vim/vimrc`.

*ADDITIONAL:*

## UNIX

`$XDG_CONFIG_HOME/vim`

## Windows

`%APPDATA%\Vim`

diegoviola:
fixes: #2034

### DIFF
--- a/src/os_unix.h
+++ b/src/os_unix.h
@@ -210,6 +210,9 @@ typedef struct dsc$descriptor   DESC;
 #ifndef SYS_VIMRC_FILE
 # define SYS_VIMRC_FILE "$VIM/vimrc"
 #endif
+#ifndef XDG_VIMRC_FILE
+# define XDG_VIMRC_FILE "~/.config/vim/vimrc"
+#endif
 #ifndef SYS_GVIMRC_FILE
 # define SYS_GVIMRC_FILE "$VIM/gvimrc"
 #endif
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -809,7 +809,8 @@ accordingly.  Vim proceeds in this order:
 	name.  Also see |vimrc-intro|.
 
 	Places for your personal initializations:
-		Unix		$HOME/.vimrc or $HOME/.vim/vimrc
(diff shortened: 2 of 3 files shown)


src/os_unix.h
```diff
@@ -210,6 +210,9 @@ typedef struct dsc$descriptor   DESC;
 #ifndef SYS_VIMRC_FILE
 # define SYS_VIMRC_FILE "$VIM/vimrc"
 #endif
+#ifndef XDG_VIMRC_FILE
+# define XDG_VIMRC_FILE "~/.config/vim/vimrc"
```
zeertzjq, line 214: This doesn't even use the `$XDG_CONFIG_HOME` environment variable.

diegoviola: @zeertzjq yes, I couldn't figure out how to do that yet, first time ever making a change in the vim codebase.

diegoviola: @zeertzjq what would you advice? I choose to hard code it for now.

diegoviola: @zeertzjq what would you advice?

diegoviola: @zeertzjq what would you suggest?

zeertzjq: First, as mentioned in the issue, this should come after `USR_VIMRC_FILE2` etc..

And you can use `mch_getenv()` to check if `XDG_CONFIG_HOME` environment variable is set. If not, use `~/.config/vim/vimrc`, otherwise use `$XDG_CONFIG_HOME/vim/vimrc`.

diegoviola: @zeertzjq Thanks, unfortunately I don't have a lot of time at the moment, so I will leave it to someone more experienced to continue.

PR #14169, opened 2024-03-09
Check listxattr's return value for failure

paultag:
The existing logic will return when the listxattr call returns with the errno set to ENOTSUP (or a size of 0 bytes), without checking to see if listxattr actually failed. listxattr can fail with at least E2BIG, ENOTSUP, ERANGE, or anything that `stat(2)` can fail with (in my case; ENOENT from stat).

The returned size is stored to a size_t, but the return type is a ssize_t. On failure, listxattr returns -1, which will get translated to size_t's MAX. If the listxattr call failed with anything other than ENOTSUP, this triggers a request for size_t MAX bytes.

This means that, if the listxattr call fails with anything other than ENOTSUP on save, vim will error with

`E342: Out of memory!  (allocating 18446744073709551615 bytes)`

(keen observers will note 18446744073709551615 is 0xffffffffffffffff)

In reality, this is likely masking a different (usually filesystem?) error -- but at least it's an error being pushed to the user now, and we don't try to allocate size_t MAX bytes.

I've opted to change the type that we store listxattr to from size_t to ssize_t, to match listxattr(2)'s signature, and to check for the -1 return value. Additionally, I've removed the errno check -- if we get a listxattr failure for any reason, we may as well bail without trying; it's not like we can even recover.

### DIFF
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -3048,7 +3048,7 @@ mch_copy_sec(char_u *from_file, char_u *to_file)
 
     size = listxattr((char *)from_file, NULL, 0);
     // not supported or no attributes to copy
-    if (errno == ENOTSUP || size == 0)
+    if (size <= 0)
 	return;
 
     for (index = 0 ; index < (int)(sizeof(smack_copied_attributes)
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, makes sense.

PR #14172, opened 2024-03-10
Internal error or crash when passing invalid position to getregion()

zeertzjq:
Problem:  Internal error or crash when passing invalid position to
          getregion().
Solution: Give an error for invalid position.


chrisbra:
thanks!

PR #14173, opened 2024-03-10
Add Vim function 'searchstr()'

rajeeshknambiar:
Add a Vim function `searchstr()` which returns matched string. Same as `search()` function in all other aspects.

There are no easy options to get a matched string in Vimscript otherwise, see [1] for some adhoc solutions.

[1] https://vi.stackexchange.com/questions/4305/is-it-possible-to-get-the-matched-string-after-calling-search

NB: This is my first contribution to Vim and I'm sure I've overlooked many things; and there are better ways to accomplish the functionality. Reviews are very welcome. Once the approach is finalized, I intend to add a `searchlist()` function too, which returns a list of all matches.

### DIFF
--- a/runtime/doc/version9.txt
+++ b/runtime/doc/version9.txt
@@ -195,6 +195,7 @@ New and extended functions: ~
 |readdirex()|		get a List of file information in a directory
 |reduce()|		reduce a List to a value
 |searchcount()|		get number of matches before/after the cursor
+|searchstr()|		get string that matches the pattern
 |setcellwidths()|	set character cell width overrides
 |setcharpos()|		set character position of cursor, mark, etc.
 |setcursorcharpos()|	set character position of the cursor
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -148,6 +148,7 @@ static void f_screencol(typval_T *argvars, typval_T *rettv);
 static void f_screenrow(typval_T *argvars, typval_T *rettv);
 static void f_screenstring(typval_T *argvars, typval_T *rettv);
 static void f_search(typval_T *argvars, typval_T *rettv);
+static void f_searchstr(typval_T *argvars, typval_T *rettv);
 static void f_searchdecl(typval_T *argvars, typval_T *rettv);
 static void f_searchpair(typval_T *argvars, typval_T *rettv);
 static void f_searchpairpos(typval_T *argvars, typval_T *rettv);
@@ -2568,6 +2569,8 @@ static funcentry_T global_functions[] =
 			ret_list_number,    f_searchpairpos},
     {"searchpos",	1, 5, FEARG_1,	    arg15_search,
 			ret_list_number,    f_searchpos},
+    {"searchstr",	1, 5, FEARG_1,	    arg15_search,
+			ret_string,	    f_searchstr},
     {"server2client",	2, 2, FEARG_1,	    arg2_string,
 			ret_number_bool,    f_server2client},
     {"serverlist",	0, 0, 0,	    NULL,
@@ -9503,7 +9506,7 @@ get_search_arg(typval_T *varp, int *flagsp)
 }
 
 /*
- * Shared by search() and searchpos() functions.
+ * Shared by search(), searchpos() and searchstr() functions.
  */
     static int
 search_cmn(typval_T *argvars, pos_T *match_pos, pos_T *end_pos, int *flagsp)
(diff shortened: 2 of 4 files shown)


src/evalfunc.c
```diff
@@ -148,6 +148,7 @@ static void f_screencol(typval_T *argvars, typval_T *rettv);
 static void f_screenrow(typval_T *argvars, typval_T *rettv);
 static void f_screenstring(typval_T *argvars, typval_T *rettv);
 static void f_search(typval_T *argvars, typval_T *rettv);
+static void f_searchstr(typval_T *argvars, typval_T *rettv);
```
rajeeshknambiar: The position needs to be alphabetically ordered, will fix in next revision.

src/evalfunc.c
```diff
@@ -9819,7 +9829,65 @@ f_search(typval_T *argvars, typval_T *rettv)
...
+    multi_line = 1;
+    }
+    // Copy first buffer line from match start location to a new allocated string
+    vim_strncpy(buf, ml_get_pos(&match_pos), matchlen);
+    if (multi_line)
+    {
+        // Copy entire lines before first & last
+	for (linenr_T l = start_lnum + 1; l < end_lnum; l++)
+	{
+	vim_strcat(buf, "\n", BUFSIZ);
```
rajeeshknambiar, line 9880: It is correct to join lines with newline? Is there a better way to return matching text over multiple lines?

chrisbra:
are you potentially looking for the [matchstrlist() function](https://github.com/vim/vim/blob/82e079df814f7372e9579450730062b205449efa/runtime/doc/builtin.txt#L6421-L6456) ?

runtime/doc/version9.txt
```diff
@@ -195,6 +195,7 @@ New and extended functions: ~
 |readdirex()|		get a List of file information in a directory
 |reduce()|		reduce a List to a value
 |searchcount()|		get number of matches before/after the cursor
+|searchstr()|		get string that matches the pattern
```
yegappan, line 198: Can you update the `function-list` section in usr_41.txt to include this function?

src/evalfunc.c
yegappan: You can use a List to return the matching strings across the lines.  This function can always return a List.  The List will contain either a single string item or multiple string items.

src/evalfunc.c
```diff
@@ -9819,7 +9829,65 @@ f_search(typval_T *argvars, typval_T *rettv)
...
+    colnr_T	end_col;
+    char_u	buf[BUFSIZ];
+    int		multi_line = 0;
+    int		matchlen = 0;
+    int		flags = 0;
+
+    rettv->vval.v_string = NULL;
+    rettv->v_type = VAR_STRING;
+
+    rettv->vval.v_number = search_cmn(argvars, &match_pos, &end_pos, &flags);
```
yegappan, line 9855: If invalid arguments are passed to the function, then search_cmn() will return FAIL.  You should handle that here.

yegappan, line 9855: The type of the return value is set to VAR_STRING, but here v_number is assigned.  You should store the return value of search_cmn() in some other variable and process it.

src/evalfunc.c
```diff
@@ -9819,7 +9829,65 @@ f_search(typval_T *argvars, typval_T *rettv)
...
+    {
+    match_pos.col -= 1; // search_cmn increments match start col by 1, adjust it for ml_get_pos.
+    start_lnum = match_pos.lnum;
+    start_col = match_pos.col;
+    end_lnum = end_pos.lnum;
+    end_col = end_pos.col;
+    if (end_lnum == start_lnum)
+    {
+    matchlen = end_col - start_col - 1;
+    }
```
yegappan, line 9867: The Vim coding style is to not use curly braces for a single statement block.  Remove the curly braces here.

src/testdir/test_search.vim
```diff
@@ -2194,5 +2194,13 @@ func Test_search_with_invalid_range()
   bwipe!
 endfunc
 
+func Test_searchstr()
+  new
+  call setline(1, ['  1', '  2 these', '  3 the', '  4 their', '  5 there', '  6 their', '  7 the', '  8 them', '  9 these', ' 10 foobar'])
+  call assert_equal("2 these\n  3", searchstr('\<2\s\+\_.*3\>'))
+
```
yegappan, line 2201: Can you add test cases for passing invalid type of argument (for each argument) to searchstr() and check the error code?
Can you add a test case to search for a string that is not present?  Can you add a test case for searching for a multi-byte string?  Can you add test cases for a few more complex search patterns?  Can you add a test case to check whether the cursor has moved?

yegappan (changes requested):

monkoose:
What about `matchbufline()`?

rajeeshknambiar:
> are you potentially looking for the [matchstrlist() function](https://github.com/vim/vim/blob/82e079df814f7372e9579450730062b205449efa/runtime/doc/builtin.txt#L6421-L6456) ?

More like [`matchstr()`](https://github.com/vim/vim/blob/82e079df814f7372e9579450730062b205449efa/runtime/doc/builtin.txt#L6457-L6471) but on the current buffer instead of an `{expr}`. 

I would also propose a `matchstrlist()` (or a  better name) that could return all the matched strings in the current buffer.

rajeeshknambiar:
> What about `matchbufline()`?

Thank you for pointing to this function (that's the one I'm missing :smile:). And it is added quite recently (just a couple of months ago), so it wasn't available when I was looking for.

Having gone through `matchbufline({buf}, {pat}, {lnum}, {end}, [, {dict}])`, I believe it would serve the purpose for me. That said, I believe there's value in adding another function (or two) as convenient wrapper(s) around `matchbufline` defaulting to the current buffer,  from starting to ending lines and returning the `text` (or list of `text` for the list returning variant) from resulting `{dict}`.

Does that sound like a viable approach? @yegappan @chrisbra ?

monkoose:
> That said, I believe there's value in adding another function (or two) as convenient wrapper(s) around matchbufline defaulting to the current buffer

I think instead of adding more functions extending matchbufline dict argument would be better, maybe adding some `stop` key to stop processing and return that much found matches. So 1 means return first one found. Default to 0 - all found matches.

chrisbra:
not sure, You can get the line that matches using the `search()` function and then you can use that one for the `lnum` and `end` position to only return a list of matches in that range.  I am not a big fan of adding a few more convenient functions around the existing functionality. Or you can directly use a combination of `search()`, `getline()` and `matchstr()` functions.

benknoble:
> > That said, I believe there's value in adding another function (or two) as convenient wrapper(s) around matchbufline defaulting to the current buffer
> 
> I think instead of adding more functions extending matchbufline dict argument would be better, maybe adding some `stop` key to stop processing and return that much found matches. So 1 means return first one found. Default to 0 - all found matches.

These kinds of extension are testing and documentation nightmares, in my experience: the function does too many things and remembering how to get it to do what you want is hard. A simplified interface via a wrapper function doesn't seem like nearly as big a maintenance or user burden.

monkoose:
> These kinds of extension are testing and documentation nightmares

How is this different to `search()` flags, except that it is more verbose, but at the same time more readable?

chrisbra:
I added a reference to `matchbufline()` to `:h search()`. 

However I am not yet convinced we need those additional wrapper functions.

rajeeshknambiar:
> I added a reference to `matchbufline()` to `:h search()`.

Thank you.

> However I am not yet convinced we need those additional wrapper functions.

Fair enough, I shall close this PR.

PR #14118, opened 2024-02-29
Added Lithuanian translation of Tutor v. 1.4 by Laurynas Stančikas

rimas-kudelis:
This translation has been available for free download from lietuvybe.org, then lietuvybe.akl.lt, then lietuvybė.lt since at least 2006[1], but nobody ever bothered to contribute it upstream.

I'm not sure if the original author's email still works, but I'm positive that he wouldn't mind his work being shared more widely. I don't think it makes much sense to distribute it only via the website dedicated to Lithuanian language and localization in IT, which lietuvybė.lt is.

[1] https://web.archive.org/web/20060222162117/http://lietuvybe.org/files/vim-tutor-lt.txt

PR #14176, opened 2024-03-10
Update Lithuanian translation of Vim tutor to v.1.7

rimas-kudelis:
Like I mentioned in my comment to #14118, I've actually updated the tutor to 1.7, so here it goes.

As part of the update, I made a lot of edits to the existing text (virtually nothing was left untouched), and added my own translator's notes at the end of the intro, which 1) explain that letter case matters for commands, and 2) list Lithuanian  key names for Esc, Ctrl and Enter.

Note I've removed the email address of the original translator, because that address seems to no longer exist (as mentioned in #14118, my email was bounced). I hope it's okay, but I can revert that change if not.


chrisbra:
thanks!

PR #14177, opened 2024-03-10
Changing buffer in another window causes it to show matchparen

zeertzjq:
Problem:  Changing buffer in another window using win_execute() causes
          it to show matchparen (after 9.0.0969).
Solution: Delay highlighting with SafeState in BufWinEnter.


zeertzjq:
In #13077 it was suggested that `:noautocmd` can be used, but that may not always be ideal, as `BufWinEnter` may also be used for purposes not limited to the current window, in which case it should not be blocked.

chrisbra:
thanks!

PR #14178, opened 2024-03-11
Fix wrong tablen with above virtual text

julio-b:
First tab is not calculated correctly with above virtual text. To reproduce the problem run this:
```sh
vim --clean - +so <<-END
	%d
	set bt=nowrite list listchars=tab:>\ 
	call append(0, ["\t\taaaa bb","\tccccccc cc"])
	call prop_type_add("type", {"highlight":"NonText"})
	call prop_add(1,0,{'text_padding_left': 16, 'type': 'type', 'text': '┌─above 1', 'text_align': 'above'})
	call prop_add(1,0,{'text_padding_left': 16, 'type': 'type', 'text': '┌─above 2', 'text_align': 'above'})
	call prop_add(2,0,{'text_padding_left': 8, 'type': 'type', 'text': '┌─above 1', 'text_align': 'above'})
	33vsp
END
```
and resize the left window.
[asciicast](https://asciinema.org/a/F9Ht3j10cuoODp4DJDpEv8jM3)


julio-b:
Hm, cursor position is broken now

julio-b:
Ok cursor position looks good, but this is probably not the right solution. Suggestions?

PR #14179, opened 2024-03-11
Internal error with blockwise getregion() in another buffer

zeertzjq:
Problem:  Internal error with blockwise getregion() in another buffer.
Solution: Also change curwin->w_buffer when changing curbuf.


chrisbra:
thanks!

PR #14183, opened 2024-03-12
More code can use ml_get_buf_len() instead of STRLEN()

zeertzjq:
Problem:  More code can use ml_get_buf_len() instead of STRLEN().
Solution: Change more STRLEN() calls to ml_get_buf_len().  Also do not
          set ml_line_textlen in ml_replace_len() if "has_props" is set,
          because "len_arg" also includes the size of text properties in
          that case.


chrisbra:
Thank you.

Issue #2467, opened 2017-12-18
heap-use-after-free in win_equal_rec


gy741:
Hello.

I found a heap-use-after-free bug in vim.

Please confirm.

Thanks.

Summary: heap-use-after-free
OS: CentOS 7 64bit
Version: b254af3
PoC Download:  [free_win_equal_rec.zip](https://github.com/vim/vim/files/1569825/free_win_equal_rec.zip)

Steps to reproduce:
1.Download the .POC files.
2.Compile the source code with ASan.
3.Execute the following command
: ./vim -u NONE -Z -X -e -s -S $POC -c :qa!

```
=================================================================
==21216==ERROR: AddressSanitizer: heap-use-after-free on address 0x60600000c2e0 at pc 0x0000014a6cf1 bp 0x7ffed15aaf70 sp 0x7ffed15aaf68
READ of size 8 at 0x60600000c2e0 thread T0
    #0 0x14a6cf0 in win_equal_rec /home/karas/vim/src/window.c:2046:14
    #1 0x147cad1 in win_equal /home/karas/vim/src/window.c:1767:5
    #2 0x1476bce in win_close /home/karas/vim/src/window.c:2454:2
    #3 0x53f6be in do_arg_all /home/karas/vim/src/buffer.c:5045:4
    #4 0x8727fa in ex_all /home/karas/vim/src/ex_docmd.c:7698:5
    #5 0x847da1 in do_one_cmd /home/karas/vim/src/ex_docmd.c:2908:2
    #6 0x82727d in do_cmdline /home/karas/vim/src/ex_docmd.c:1071:17
    #7 0x813fb7 in do_source /home/karas/vim/src/ex_cmds2.c:4411:5
    #8 0x810477 in cmd_source /home/karas/vim/src/ex_cmds2.c:4024:14
    #9 0x810596 in ex_source /home/karas/vim/src/ex_cmds2.c:3999:2
    #10 0x847da1 in do_one_cmd /home/karas/vim/src/ex_docmd.c:2908:2
    #11 0x82727d in do_cmdline /home/karas/vim/src/ex_docmd.c:1071:17
    #12 0x82c835 in do_cmdline_cmd /home/karas/vim/src/ex_docmd.c:671:12
    #13 0x1658084 in exe_commands /home/karas/vim/src/main.c:2953:2
    #14 0x1651bc6 in vim_main2 /home/karas/vim/src/main.c:800:2
[...]

PR #14185, opened 2024-03-12
Small split-move related improvements

seandewar:
Problem: small improvements can be made to split-move related functions.

Solution: apply them:

- Improve some doc comments (frame_flatten should still work for non-current tabpages, despite the topframe check, which looks benign, though I'm unsure if it's needed; see #2467).

- f_win_splitmove should check_split_disallowed on wp, not targetwin, as that's what win_splitmove checks (though it's probably unnecessary to check b_locked_split at all; see #14109, which I hope to get around to finishing at some point).

- Make winframe_restore restore window positions for the altframe, which winframe_remove changes. This doesn't affect the prior behaviour, as we called win_comp_pos after, but as win_comp_pos only works for curtab, and winframe_remove supports non-current tabpages, we should undo it. Regardless, this should mean we don't need win_comp_pos anymore; adjust tests to check that window positions remain unchanged.

  To be honest, I'm not sure win_comp_pos is needed anyway after last_status if it's not stealing rows from another frame to make room for a new statusline, which shouldn't be the case after winframe_remove? To be safe, I'll leave that as is.

### DIFF
--- a/src/evalwindow.c
+++ b/src/evalwindow.c
@@ -1001,8 +1001,8 @@ f_win_splitmove(typval_T *argvars, typval_T *rettv)
 	size = (int)dict_get_number(d, "size");
     }
 
-    // Check if we can split the target before we bother switching windows.
-    if (text_or_buf_locked() || check_split_disallowed(targetwin) == FAIL)
+    // Check if we're allowed to continue before we bother switching windows.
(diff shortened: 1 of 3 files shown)


chrisbra:
Thanks.

Issue #14047, opened 2024-02-17
Vim beeps when opening a file using the CtrlP plugin


sfsam:
### Steps to reproduce

1. Open Vim.
2. Use CtrlP to open a file.

### Expected behaviour

Vim beeps when opening the file. It should open the file without the beep. I checked `:messages` but saw nothing there.

I think the commit that triggers this behavior is bf44b69d1f91d9778ae1887128c63d35d9a3d19b. I used `git bisect`, but I've never built Vim before nor used `bisect` before so I might be mistaken.

### Version of Vim

9.1.76

### Environment

macOS 14.3.1

seandewar:
> I think the commit that triggers this behavior is https://github.com/vim/vim/commit/bf44b69d1f91d9778ae1887128c63d35d9a3d19b.

If that's indeed the commit, it _may_ not indicate a bug in Vim; `:wincmd p` beeping when there's no previous window is normal.

However, before that commit, there was a case (seemingly the only such possible) where both the current and previous windows could refer to the same window if closing a different window resulted in Vim choosing the previous as the new current window. That may have resulted in `:wincmd p` not beeping previously, but still having no effect.

sfsam:
Thanks for the reply. I've been using CtrlP to open files like this for a while and this is the first time I've ever experienced this beeping behavior. CtrlP opens a window at the bottom of the screen from which you can select or fuzzy-find a file to open. When you select your file, it opens and the CtrlP window at the bottom closes. Any idea how to proceed? Is this a Vim issue or a CtrlP issue? Of course, I can still use CtrlP, but I'd much prefer it not beep every single time I open a file.

seandewar:
> Any idea how to proceed? Is this a Vim issue or a CtrlP issue?

If things are happening as I assume, I'd consider the old behaviour a bug (and the lack of beeping prior to the patch). Plugins using `:wincmd p` should consider using `:silent!` or check `winnr('#') != 0` if they don't want potential beeps. Though if the patch causes too much fallout in plugins, maybe reverting it and documenting the old behaviour could be preferable.

[...]

PR #14186, opened 2024-03-12
Fix a win_execute issue, re-allow curwin == prevwin

seandewar:
Details are in the commit messages. Hopefully it's OK. :+1:

If the 2nd commit is merged, #14047 should be closed.

_**Note**: alternatively, I have a patch [here](https://github.com/seandewar/vim/commit/d243d6595f601f407f073ecbb0da9531af578723) that addresses the cases mentioned in the last bullet in the 2nd commit's message, if preferred._


chrisbra:
thank you. Yes I think it makes sense and the reasoning in the commit message it also good. So let's partly revert the 9.1.0001 patch.

PR #14188, opened 2024-03-12
Some comments in fold.c are outdated

zeertzjq:
Problem:  Some comments in fold.c are outdated.
Solution: Update comments.

### DIFF
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -6612,7 +6612,7 @@ ex_vimgrep(exarg_T *eap)
 	goto theend;
     }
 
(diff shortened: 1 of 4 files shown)


src/quickfix.c
```diff
@@ -6612,7 +6612,7 @@ ex_vimgrep(exarg_T *eap)
 	goto theend;
     }
 
-    // Jump to first match if the current window is not 'winfixbuf'
+    // Jump to first match.
```
zeertzjq, line 6615: This comment also seems wrong as 'winfixbuf' is checked at the top of this function.

zeertzjq:
Moved to #14189

PR #14189, opened 2024-03-13
Coverity reports dead code

zeertzjq:
Problem:  Coverity reports dead code.
Solution: Remove the dead code. Also fix a mistake and some comments.

### DIFF
--- a/src/memline.c
+++ b/src/memline.c
@@ -2686,7 +2686,7 @@ ml_get_len(linenr_T lnum)
     colnr_T
 ml_get_pos_len(pos_T *pos)
 {
-    return ml_get_buf_len(curbuf, curwin->w_cursor.lnum) - pos->col;
+    return ml_get_buf_len(curbuf, pos->lnum) - pos->col;
 }
 
 // return length (excluding the NUL) of the cursor line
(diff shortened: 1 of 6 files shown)


src/ops.c
```diff
@@ -2687,9 +2684,10 @@ do_addsub(
 
     curwin->w_cursor = *pos;
     ptr = ml_get(pos->lnum);
+    linelen = ml_get_len(pos->lnum);
     col = pos->col;
 
-    if (*ptr == NUL || col + !!save_coladd >= (int)STRLEN(ptr))
```
chrisbra, line 2692: why is the `*ptr == NUL` condition removed?

zeertzjq: It's not needed because `col >= 0`, and `ml_get_len()` returns 0 if `*ptr == NUL`.

src/memline.c
```diff
@@ -2686,7 +2686,7 @@ ml_get_len(linenr_T lnum)
     colnr_T
 ml_get_pos_len(pos_T *pos)
 {
-    return ml_get_buf_len(curbuf, curwin->w_cursor.lnum) - pos->col;
+    return ml_get_buf_len(curbuf, pos->lnum) - pos->col;
```
chrisbra, line 2689: oh, I should have caught this when reviewing yesterday :/

src/ops.c
chrisbra: ah, so when `*ptr == NUL` this means that the second condition will be also true: `col + !!save_coladd >= linelen` I see. thanks

chrisbra:
thanks!

PR #14190, opened 2024-03-13
Cursor column wrong with 'virtualedit' and conceal

zeertzjq:
Problem:  Cursor column wrong with 'virtualedit' and conceal.
Solution: Correct cursor column at end of line if never reached.


zeertzjq:
Before:
https://github.com/vim/vim/assets/35768171/2931af2e-0993-4158-8b3a-4a9403a9f5d6

After:
https://github.com/vim/vim/assets/35768171/f1b7595d-5b35-4909-8531-ad8b5a0cc918

chrisbra:
thanks!

PR #14192, opened 2024-03-13
'cursorline' and 'wincolor' hl missing with conceal and wrap

zeertzjq:
Problem:  'cursorline' and 'wincolor' hl missing with conceal and wrap.
Solution: Apply 'cursorline' and 'wincolor' highlight to boguscols.

Since 'cursorline' and 'wincolor' highlight apply after the end of the
line, it is more consistent to have them also apply to boguscols.

Assigning MAXCOL to values in ScreenCols[] make mouse click behave the
same with 'cursorline' and 'nocursorline', but such behavior may be
incorrect, as it puts the cursor on the next screen line.  That may be
fixed in a future PR.


chrisbra:
Mouse clicking with concealed items has been wrong since the concealing patch has been included. I never knew how to properly fix it. So it should be fine to keep it like this for now. Thanks!

PR #14194, opened 2024-03-13
runtime(doc): Recover some missed commas and periods in starting.txt

zzzyxwvut:
Also, insert some missing words and, in general, strive for  
consistency with capitalisation of words.

The `gvim` words were left alone for now, but they deserve to  
be treated like proper names, GVim or GUI Vim, unless these  
refer to executable filenames (on \*nix systems).

### DIFF
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -1,4 +1,4 @@
-*starting.txt*  For Vim version 9.1.  Last change: 2023 Oct 20
+*starting.txt*  For Vim version 9.1.  Last change: 2024 Mar 13
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -113,10 +113,10 @@ Additional characters may follow, they are ignored.  For example, you can have
 "gvim-8" to start the GUI.  You must have an executable by that name then, of
 course.
 
-On Unix, you would normally have one executable called Vim, and links from the
-different startup-names to that executable.  If your system does not support
-links and you do not want to have several copies of the executable, you could
-use an alias instead.  For example: >
+On Unix, you would normally have one executable called "vim", and links from
+the different startup-names to that executable.  If your system does not
+support links and you do not want to have several copies of the executable,
+you could use an alias instead.  For example, in a C shell descendant: >
 	alias view   vim -R
 	alias gvim   vim -g
 <
@@ -194,7 +194,7 @@ a slash.  Thus "-R" means recovery and "-/R" readonly.
 
 --cmd {command}						*--cmd*
 		{command} will be executed before processing any vimrc file.
-		Otherwise it acts like -c {command}.  You can use up to 10 of
+		Otherwise, it acts like -c {command}.  You can use up to 10 of
 		these commands, independently from "-c" commands.
 
 							*-S*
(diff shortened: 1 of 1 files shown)


zzzyxwvut:
This `csh` snippet made me look for its introduction:

https://github.com/vim/vim/blob/760f664213dea9a300454992ba1589f4601d622f/runtime/doc/starting.txt#L485-L487

So, I ended up here ([`v4.0`](https://github.com/vim/vim-history/blob/1a6a0426303bb1a14fee3eb93cb65f89c1d603f2/doc/vim_ref.txt#L375-L377)):

>		mappings and settings.  A shell alias can be used to make
>		this easy to use.  For example:
>			"alias vimc vim -u ~/.c_vimrc !*".

And lost its track in `v3.0`, whose [`readme`](https://github.com/vim/vim-history/blob/81d8b473260b31ea465268f9463b8cbf61c329d4/readme#L17-L27) says:

> Documentation:
> tutor/readme - one hour training course for beginners
> reference.doc - complete reference of all Vim commands
> difference.doc - summarizes the differences with UNIX vi
> windows.doc - reference for the multi windows and buffer commands
> index - alfabetical list of commands
> amiga.doc - remarks for Amiga
> unix.doc - remarks for unix
> msdos.doc - remarks for MSDOS
> winnt.doc - remarks for Windows NT
> archie.doc - remarks for Archimedes

There are no `*.doc` files in that repository under that tag.  
Were these files MS Word binaries?

chrisbra:
I don't think so. Probably just the extension for some text document. 

While you are at the csh alias thing, can you please add, that this is only valid for (t)csh ? and I believe it should probably be:
```
alias vimc 'vim -u ~/.c_vimrc !*'
```

zzzyxwvut:
Well, I'm not a C shell user, however, these aliases seem  
to do the job with `bsd-csh`, version `20230828-1` (Debian  
trixie/testing):

```csh
alias vimd vim -u /tmp/defaults.vim !\*
alias vimd 'vim -u /tmp/defaults.vim \!*'
```

whereas, these aliases report "0: Event not found.":


```csh
alias vimd vim -u /tmp/defaults.vim !*
alias vimd 'vim -u /tmp/defaults.vim !*'
```

Should we, instead, replace it with a Bash equivalent?

```bash
alias vimc='vim -u ~/.c_vimrc'
```

chrisbra:
I haven't used tcsh in 20 years, but yeah, it seems like one needs to escape the `!` instead. You can also add a bash equivalent, that would be more understandable than that `!*`

zzzyxwvut:
Done.

runtime/doc/starting.txt
```diff
@@ -1087,7 +1089,7 @@ Avoiding trojan horses ~
 							*trojan-horse*
 While reading the "vimrc" or the "exrc" file in the current directory, some
 commands can be disabled for security reasons by setting the 'secure' option.
-This is always done when executing the command from a tags file.  Otherwise it
+This is always done when executing the command from a tags file.  If not, it
```
chrisbra: here I find the Otherwise more clear

chrisbra:
mostly fine. For the comma rules, I'd like to get a native speaker to have a look if possible.

runtime/doc/starting.txt
zzzyxwvut: Sure, I'll put it back.  The reason I went for "If not," is 
that I lothed to reformat to width the lines after for the  
needed change "Otherwise,".

PR #14196, opened 2024-03-14
'wincolor' highlight missing with "below" virtual text

zeertzjq:
Problem:  'wincolor' highlight missing with "below" virtual text.
Solution: Subtract n_attr_skip from n_attr.  Combine 'wincolor' with
          other highlights when 'nowrap' is set.


zeertzjq:
This should also make the applying of `CursorLine`/`CursorColumn`/`ColorColumn` highlight consistent on line without text property vs. line with text property:
- `ColorColumn` is shown over `CursorLine`
- `CursorColumn` is not shown on current line

https://github.com/vim/vim/assets/35768171/11fb5961-d5b4-4179-8b40-e63a5ec2af3f

chrisbra:
thanks!

PR #14197, opened 2024-03-14
Cursor pos wrong when double-width chars are concealed

zeertzjq:
Problem:  Cursor pos wrong when double-width chars are concealed.
Solution: Advance one more virtual column for a double-width char.
          Run some tests with both 'wrap' and 'nowrap'.


chrisbra:
thanks!

PR #14198, opened 2024-03-14
Makefile for running indent tests on OS Windows.

RestorerZ:
Problem: On OS Windows there is no makefile for testing indents.
Solution: Add Make_mvc.mak to the "runtime/indent" directory.
Also:
"indent/testdir/runtest.vim" — bit optimized.
"indent/Makefile" — added rules .PHONY and .SUFFIXES
"Filelist" — added the file "runtime/indent/Make_mvc.mak". Some files have been moved to more logically appropriate variables (macros).

### DIFF
--- a/Filelist
+++ b/Filelist
@@ -1,8 +1,8 @@
 # List of distributed Vim files.
 # Used by Makefile.
 
-# Files of Git and GitHub.
-FILE_GIT =	\
+# Source files for all source archives.
+SRC_ALL =	\
 		.cirrus.yml \
 		.gitattributes \
 		.github/CODEOWNERS \
@@ -21,11 +21,6 @@ FILE_GIT =	\
 		ci/if_ver*.vim \
 		ci/setup-xvfb.sh \
 		ci/remove_snap.sh \
-		src/libvterm/.bzrignore \
-		src/libvterm/.gitignore \
-
-# The source code itself is for all source archives.
-BARE_SRC_ALL = \
 		src/Make_all.mak \
 		src/README.md \
 		src/alloc.c \
@@ -343,7 +338,10 @@ BARE_SRC_ALL = \
 		src/proto/viminfo.pro \
 		src/proto/winclip.pro \
 		src/proto/window.pro \
+		src/libvterm/.bzrignore \
+		src/libvterm/.gitignore \
 		src/libvterm/LICENSE \
+		src/libvterm/Makefile \
 		src/libvterm/README \
 		src/libvterm/CONTRIBUTING \
 		src/libvterm/tbl2inc_c.pl \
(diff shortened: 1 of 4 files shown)


chrisbra:
Can you please next time make sure, to not add un-related changes to it? I find it hard to review, if you change all the commens only slightly.

Filelist
```diff
@@ -429,8 +431,13 @@ SRC_ALL =	\
 		src/xdiff/xutils.c \
 		src/xdiff/xutils.h \
 
+# Source files for all source archives.
+SRC_ALL =	\
+		$(FILE_GIT) \
+		$(BARE_SRC_ALL) \
```
chrisbra: why split those?

Filelist
```diff
@@ -834,15 +833,24 @@ RT_SCRIPTS =	\
...
 		README_unix.txt \
+		runtime/doc/*.awk \
+		runtime/doc/Makefile \
+		runtime/doc/vim.1 \
+		runtime/doc/evim.1 \
+		runtime/doc/vimdiff.1 \
+		runtime/doc/vimtutor.1 \
+		runtime/doc/xxd.1 \
+		runtime/indent/Makefile \
+		runtime/pack/dist/opt/editorconfig/mkzip.sh \
```
chrisbra: I don't think those manpages belong to $RT_UNIX

Filelist
```diff
@@ -761,29 +762,28 @@ RT_ALL =	\
...
 		runtime/pack/dist/opt/editorconfig/autoload/*.vim \
 		runtime/pack/dist/opt/editorconfig/autoload/editorconfig_core/*.vim \
-		runtime/pack/dist/opt/editorconfig/doc/tags \
 		runtime/pack/dist/opt/editorconfig/doc/editorconfig.txt \
 		runtime/pack/dist/opt/editorconfig/ftdetect/editorconfig.vim \
 		runtime/pack/dist/opt/editorconfig/plugin/editorconfig.vim \
 		runtime/pack/dist/opt/justify/plugin/justify.vim \
 		runtime/pack/dist/opt/matchit/plugin/matchit.vim \
 		runtime/pack/dist/opt/matchit/doc/matchit.txt \
-		runtime/pack/dist/opt/matchit/doc/tags \
```
chrisbra: don't remove those tags, otherwise the help won't work

RestorerZ: This file is in the "RT_ALL_BIN" variable, where "runtime/doc/tags" is located.
I assumed that if "doc/tags" is originally defined in this variable, then it is logical that others should be defined in the same variable.
Moreover, the variable "RT_ALL_BIN" is added to all runtime archives.

Filelist
RestorerZ: Well, I assumed that they apply to all Unix-like systems. Although, of course, I may be wrong. But anyway, Windows doesn't have a built-in tool for viewing manpages.

Filelist
RestorerZ: These same files are not directly source code.
For an offline local archive, they are, in principle, not required to build the program.

RestorerZ:
> Can you please next time make sure, to not add un-related changes to it? I find it hard to review, if you change all the commens only slightly.

I understood. I'm sorry. I will roll back these made changes to the "Filelist" except for "Make_mvc.mak". And we will discuss these changes in another PR, if necessary. Ok?

chrisbra:
no problem. Leave the PR as is. Just for the next time please :)

Issue #9987, opened 2022-03-21
wrong syntax highlight of vim


Shane-XB-Qian:
### Steps to reproduce

```vim
let foo = {'end': 123}

if 123
	echo (foo.end)
else
	echo 'bar'
endif
```

that first `echo` is not correctly hightlight unless remove `()` from `foo.end`.

### Expected behaviour

highlight correctly

### Version of Vim

vim v8.2.4602

### Environment

linux

Shane-XB-Qian:
![image](https://user-images.githubusercontent.com/13550529/159282341-0f9f40dd-22f1-4b65-ad8f-50e692fd19c3.png)

Shane-XB-Qian:
it is fixed after last runtime updated.
// closing...

PR #14199, opened 2024-03-14
runtime(vim): Update base-syntax, improve :echo and :execute highlighting

dkearns:
Improve :echo and :execute highlighting.

- Add better line-continuation support for both commands.
- Improve the :execute command's expression argument matching.
- Remove the fix for issue #9987 as this is now handled by correctly
  matching the parens in :echo (...) as operator parens.


chrisbra:
thanks!

zeertzjq:
This seems to regress #9987.

Before:
before

After:
after

PR #14201, opened 2024-03-15
tools/rename.bat uses 'mv' and "move"

RestorerZ:
The file "tools/rename.bat" uses both "mv" (if available) and the native Windows "move" command.
To use this file on UNIX-like systems, you need to make it executable.

### DIFF
--- a/tools/rename.bat
+++ b/tools/rename.bat
@@ -1,6 +1,36 @@
-mv vim.exe vimw32.exe
-mv tee/tee.exe teew32.exe
-mv xxd/xxd.exe xxdw32.exe
-mv vim.pdb vimw32.pdb
-mv install.exe installw32.exe
-mv uninstall.exe uninstallw32.exe
+@ 2>/dev/null # 2>nul & goto:win32
+if test -e ../src/vim.exe ; then mv ../src/vim.exe ../src/vimw32.exe ; fi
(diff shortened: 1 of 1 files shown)


chrisbra:
I am surprised. Does this make it work on Windows as on Linux?

RestorerZ:
Well, it should, in theory.
Tested WinNT 6.1, Ubuntu 22, FreeBSD 14
But it's better to check somewhere else. I just don't have anything else at hand.

It's just that the original required an "mv", and I might not have one.
And the documentation implies that the Vim installer can be made on Unix-like systems, so that one tool can be used.

k-takata:
I think Bram used to use this on Windows with Unix-like tools before running NSIS.
Currently, our CI creates NSIS installers without using this file.

chrisbra:
okay thanks.

RestorerZ:
> I think Bram used to use this on Windows with Unix-like tools before running NSIS.

Well, yes, its use before creating the installer is directly described in the documentation. But due to the use of facilities in this tool that are not originally available in MSFT Windows, it is unfortunately not widely used.

> Currently, our CI creates NSIS installers without using this file.

Of course, you can do without this tool. And not only in CI (Continuous Integration), but also in "home conditions" when you create an installer for yourself and your friends.

Ultimately, whether this tool is needed or not is up to you to decide. But if it is described in the documentation and is primarily focused on the Windows platform, then I tried to make it "friendly" to this platform and, if possible, universal.

PR #14203, opened 2024-03-15
runtime(vim): Update base-syntax, improve function highlighting

dkearns:
Improve function highlighting.

- Match bang and function modifiers - abort etc.
- Only match valid scope modifiers.
- Match listing commands.
- Don't match ex commands in function names.
- Split function syntax groups into :func and :def subgroups.
- Match Vim9-script parameter and return types.
- Limit legacy-script and Vim9-script comments to :func and :def definitions, respectively.


chrisbra:
thanks!

PR #14206, opened 2024-03-15
Refactor to remove STRLEN() calls

basilisk0315:
I have added a new structure to store key (int)/value (string) pairs for use in static lookup tables (eg event names, regexp character classes, special keywords, highlighting). This structure embeds the length of the string which allows removal of calls to STRLEN() when performing the lookup comparisons.

In addition to factoring out calls to STRLEN(), I have added a simple circular FILO caching mechanism to some lookup functions with the goal to speed them up.
To test the effectiveness of the caching I added some counters to the function event_name2nr() in autocmd.c. Running ./vim autocmd.c on my HP-UX build without the cache the main lookup loop was executed 35182 times. With the cache, the cache loop was executed 5796 times and the main lookup loop was executed 282 times (total 6078). This is a reduction of 29104 executions.


zeertzjq:
I think I now know how you cam rebase this PR:
```sh
git remote add upstream https://github.com/vim/vim.git
git fetch upstream
git rebase upstream/master
```

PR #14208, opened 2024-03-16
runtime(doc): make :h tag-! more consistent

zeertzjq:
- Use "on" and "off" for 'winfixbuf' option values.
- Retab the table.


chrisbra:
thanks!

PR #14209, opened 2024-03-16
Can define function with invalid name inside 'formatexpr'

zeertzjq:
Problem:  Can define function with invalid name inside 'formatexpr'.
Solution: Use goto instead of checking for did_emsg later.


chrisbra:
thanks!

PR #14213, opened 2024-03-16
Fix Go `goPackageComment` highlighting too broad

matthewhughes934:
Previously this would highlight any comment before a line starting `package`, for example

    var (
      // This comment would be highlighted as a goPackageComment
      packages []string
    )

The package clause is only valid when followed by a space[1], so include this restriction

This is upstreaming a change from `go-vim`[2]

[1] https://go.dev/ref/spec#Package_clause
[2] https://github.com/fatih/vim-go/commit/d1c36cc4173093d0be78fd826978a024febe6233

### DIFF
--- a/runtime/syntax/go.vim
+++ b/runtime/syntax/go.vim
@@ -531,12 +531,12 @@ if s:HighlightBuildConstraints() || s:FoldEnable('package_comment')
   " matched as comments to avoid looking like working build constraints.
   " The he, me, and re options let the "package" itself be highlighted by
   " the usual rules.
-  exe 'syn region  goPackageComment    start=/\v(\/\/.*\n)+\s*package/'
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks, have you checked with the maintainer of the file?

matthewhughes934:
> Thanks, have you checked with the maintainer of the file?

I did not, I jumped the gun a bit (read: I did not carefully read `CONTRIBUTING.md` :sweat_smile:) before creating this PR. I'll leave a question on the original `vim-go` PR since that was approved by the maintainer of this file.

bhcleek:
@chrisbra yes, I'm ok with this change being merged.

chrisbra:
alright thanks all

chrisbra:
included as of 3d46de703cf645032fa4dbcd71259176db1b46b7

PR #14215, opened 2024-03-16
runtime(testing): Support Vim configuration for syntax tests

zzzyxwvut:
Currently, the very same syntax file for which a test can  
be written is the only place where global variables can be  
defined so that the file parts guarded with such variables  
can be read during screen dump generation.  This approach  
would lead to littering the syntax file with test-related  
queries.

Instead, we could borrow the idea of comment-based mechanism  
for test setup from the indent test runner.  With it, the  
first 20 lines of each test file would be ALWAYS scanned in  
search of the `VIM_TEST_SETUP` markers and, when found, the  
part between the end of the marker and the end of the line  
would be treated as a Vim Ex command.

Note that with these changes, `runtime/defaults.vim` is no  
longer sourced for screen dump generation; however, some of  
its functionality is reintroduced.

Also, we disable italic highlighting for all test files to  
avoid issues on some terminals.

Not all programming languages support comments; without such  
support, the `VIM_TEST_SETUP` functionality that offers  
embeddable Vim Ex commands may no longer be applicable.

We can achieve similar functionality by storing Ex commands  
in a Vim file with its basename matching the basename of the  
test file, and having this Vim file sourced, and then having  
the test file loaded and tested.  When such a Vim file would  
be used for a language that has comments and whose matching  
test file has embedded `VIM_TEST_SETUP` lines, we will  
accommodate it by letting the `VIM_TEST_SETUP` Ex commands  
augment and/or overrule sourced configuration.

Further details can be found in the discussion thread at  
https://github.com/vim/vim/discussions/14117.

Also:

- Retrofit input/c.c. with `VIM_TEST_SETUP`.

- Include a syntax test for Markdown that takes advantage of  
  a sourceable setup configuration.

- Update Filelist.


zzzyxwvut:
@chrisbra, @dkearns, please review.

runtime/syntax/testdir/runtest.vim
```diff
@@ -145,6 +147,23 @@ func RunTest()
...
+	func SetUpVim()
+	  call cursor(1, 1)
+	  " Defend against rogue TEST_SETUP commands.
+	  for _ in range(20)
+	    let lnum = search('\<TEST_SETUP\>', 'eW', 20)
+	    if lnum < 1
+	      break
+	    endif
+	    exe substitute(getline(lnum), '.*TEST_SETUP', '', '')
+	  endfor
```
cvwillegen, line 159: Is it possible to execute "dangerous" commands this way? Should this be in the sandbox, if it isn't already?

zzzyxwvut: It is sure possible to _try_ to execute a non-command and  
fall flat.  Does the sourcing of `syntax/foo.vim` happen  
in a sandbox?  If not, you may well sandbox the setting  
of an innocuous variable _here_ and defer arbitrary execution  
to the code that queries it _there_.  The author(s) of  
`indent/testdir/runtest.vim` did not make any allowance for  
"better securing" similar lines.

Let us seek more informed opinions on the severity of doing  
nothing.

chrisbra: that should be fine, since this is only run from the test-suite. So hopefully any dangerous commands are noticed before we introduce them into the repo.

runtime/syntax/testdir/README.txt
```diff
@@ -26,14 +26,33 @@ Creating a syntax plugin test
...
+
+Remember that no heuristic is set in place to discover whether the matched
+lines belong to a comment or any other valid or invalid construct of the
+language under test; therefore, only for Vim configuration can "TEST_SETUP"
+appear within the top 20 lines.  As an alternative, you can keep the same Vim
+Ex commands in a Vim file named after the base of the test file.  Create this
+file in the "setup" directory, whose parent directory is "input".  This file
+would then be "input/setup/java.vim".  Note that when both sources of
+configuration are available, the commands from the "TEST_SETUP" lines will be
+processed after the commands from "input/setup/java.vim".
```
dkearns: The language in this section seems a little bit elaborate to me so here's another take - just a suggestion.

> Optionally, pre-configure the testing environment by including setup commands at the top of the input file.  The format for these lines is:
>  
>         TEST_SETUP {command}
>  
> where {command} is any valid Ex command.  The first 20 lines of the input file are scanned for setup commands and these will be executed before the syntax highlighting is enabled.  Typically, these lines would be included as comments so as not to introduce any syntax errors in the input file but this is not required.
>  
> Continuing the Java example:
>  
>         // TEST_SETUP let g:java_space_errors = 1
>         // TEST_SETUP let g:java_minlines = 5
>         class Test { }
> 
> As an alternative, setup commands can be included in an external Vim script file in the "input/setup" directory.  This script file must have the same base name as the input file.
> 
> So, the equivalent example configuration using this method would be to create an "input/setup/java.vim" script file with the following lines:
> 
>         let g:java_space_errors = 1
>         let g:java_minlines = 5
> 
> Both inline setup commands and setup scripts may be used at the same time, the script file will be sourced before any TEST_SETUP commands are executed.

runtime/syntax/testdir/README.txt
```diff
@@ -26,14 +26,33 @@ Creating a syntax plugin test
...
+example:
+
+	// TEST_SETUP let g:java_space_errors = 1
+	// TEST_SETUP let g:java_minlines = 5
+	class Test { }
+
+Remember that no heuristic is set in place to discover whether the matched
+lines belong to a comment or any other valid or invalid construct of the
+language under test; therefore, only for Vim configuration can "TEST_SETUP"
+appear within the top 20 lines.  As an alternative, you can keep the same Vim
```
dkearns: > therefore, only for Vim configuration can "TEST_SETUP"  appear within the top 20 lines.

therefore, only the first 20 lines of the input file are scanned for Vim TEST_SETUP commands.

dkearns (commented):
Thanks @zzzyxwvut.

I added a comment about the same sentence @chrisbra mentioned the other day and offered some alternate documentation for that section.

Otherwise, it looks good to me.

runtime/syntax/testdir/README.txt
zzzyxwvut: It is considerably more lucid.  Thank you.  I'll take it.

runtime/syntax/testdir/README.txt
zzzyxwvut: Would you rather prefer this variant?

> Remember that no heuristic is set in place to discover whether the matched
> lines belong to a comment or any other valid or invalid construct of the
> language under test; therefore, only the first 20 lines of the input file are
> scanned for Vim TEST_SETUP commands.

Comments or not comments, any `TEST_SETUP` lines in the top  
20 lines of a test file are pre-empted on behalf of Vim  
configuration.  And we need to be honest about this fact.  
So that test authors would exercise care by not having  
within those lines `TEST_SETUP` labels in `switch-case`  
statements, `TEST_SETUP` macros, etc.

Can we rack our brains and come up with a better wording?

runtime/syntax/testdir/README.txt
zzzyxwvut: I also think we need to forewarn about dangling comment  
markers:

```html
<!-- TEST_SETUP let g:html_syntax_folding = 1 -->
```

```c
/* TEST_SETUP let g:c_minlines = 5 */
```

Both configurations are invalid!

So I would amend your variant:

> where {command} is any valid Ex command.  . . .

as follows:

> where {command} is any valid Ex command, which extends to the end of line.  . . .

runtime/syntax/testdir/README.txt
zzzyxwvut: Would changing this section:

> The first 20 lines of the input file are scanned for setup commands and these
> will be executed before the syntax highlighting is enabled. Typically, these
> lines would be included as comments so as not to introduce any syntax errors in
> the input file but this is not required.

to this variant read better:

> The first 20 lines of the input file are always scanned for setup commands and these
> will be executed before the syntax highlighting is enabled.  Typically, these
> lines would be included as comments so as not to introduce any syntax errors in
> the input file but this is not required. However, every TEST_SETUP line is
> treated as Vim configuration in the first 20 lines.  Please do not confuse Vim
> by naming TEST_SETUP your case labels in switch statements, your macros, etc.

and the whole sentence of "Remember that no heuristic etc."  
be expunged?

runtime/syntax/testdir/README.txt
dkearns: Good point.  Let's add that too.

runtime/syntax/testdir/README.txt
dkearns: If you like, but I don't think we need to be that defensive in the documentation.  I think it's all explicit enough from the initial description.

Ideally the search for the token would be tightened up to match at the start of the line or in the 'commentstring' option template.  Unfortunately, there's lots of filetypes that don't have a 'commentstring' defined.

The error reporting for bad commands isn't great but that doesn't have to be addressed at this time as far as I'm concerned.

zzzyxwvut: Okay, then ALWAYS in caps may suffice?

> The first 20 lines of the input file are ALWAYS scanned for setup commands and
> these will be executed before the syntax highlighting is enabled.  Typically,
> these lines would be included as comments so as not to introduce any syntax
> errors in the input file but this is not required.
>
> Continuing the Java example:

And no "Remember that no heuristic etc." sentence.

dkearns: OK.

If we're really worried about this sort of accidental execution of arbitrary code we should be using the sandbox and probably a more specific token like `VIM_SETUP`.

zzzyxwvut:
The changes have been pushed, please review.

chrisbra:
looks good except for the failing filelist test.

zzzyxwvut:
I'm not really familiar with the CI here.

Should I patch Filelist in the RT_SCRIPTS section for the  
HEAD commit like so:

```yaml
		runtime/syntax/testdir/input/setup/*.vim \
```

chrisbra:
Make sure you have  `runtime/syntax/testdir/input/setup/*.*` around here: https://github.com/vim/vim/blob/0e17f7e97290f81ff55743ab5cdc69ef118be080/Filelist#L826-L829

zzzyxwvut:
Okay, just give a minute or two.

zzzyxwvut:
How about now?

chrisbra:
Hm, the markdown syntax test fails on MacOS 12. I think it cannot handle the italic highlighting so the screen dump fails. We had similar issues with HTML before I believe. I am not sure if we can skip the MacOS 12 runner for this particular test.

grafik

zzzyxwvut:
I use `screen.xterm-256color` (`$TERM`) for generation.  Now  
I did `make clean test` with `xterm-256color` and there are  
no failures, `testdir/failed` is empty (Debian).

But they sure differ with

```shell
../../src/vim --clean '+call term_dumpload("testdir/dumps/markdown_conceal_00.dump")'
```

when run from `screen.xterm-256color`:

screen xterm-256color

and when run from `xterm-256color`:

xterm-256color

zzzyxwvut:
Let me know if you want me to remove all _italics_ from  
`input/markdown_conceal.markdown` and push again.

chrisbra:
oh great. Can you try this patch on top of it? Let's push it and see if CI is happy with it.
```patch
index a1585b18b..8c8d4fcce 100644
--- a/runtime/syntax/testdir/input/setup/markdown_conceal.vim
+++ b/runtime/syntax/testdir/input/setup/markdown_conceal.vim
@@ -1,3 +1,5 @@
+" disable italic highlighting, causes issues on some terminals
+set t_ZH= t_ZR=
```

zzzyxwvut:
Okay.

runtime/syntax/testdir/runtest.vim
dkearns: Is there a reason not to run these commands in the sandbox?  Anything requiring full capabilities would more naturally be included in a setup script file anyway.

What we really need to guard against is users adding random sample files to their forks and running the tests, rather than what gets included here.

There's about [8k matches](https://github.com/search?q=%2F%28%3F-i%3A%5CbTEST_SETUP%5Cb%29%2F&type=code) for `TEST_SETUP` on GitHub, 8 for `VIM_SETUP` and 0 for `VIM_TEST_SETUP` so changing the token would also help.

zzzyxwvut: A unique name is a boon.

I'll go and rebase commits then, replace `TEST_SETUP` with  
`VIM_TEST_SETUP`, reflow documentation, regenerate and peruse  
`dumps/c_*`.

@chrisbra, the CI jobs seem to no longer fail for `runtime/`  
tests.  I guess, the fix for _italics_ have worked, right?

chrisbra: yes looks like it. But I think we should move it into the `Xtestscript` instead?

zzzyxwvut: So for any filetype, right?

```vim
	  " BEGIN [runtime/defaults.vim]
	  " Also, disable italic highlighting; it causes issues on some terminals.
	  set display=truncate ruler scrolloff=5 t_ZH= t_ZR=
	  " Provide pre-VIM_TEST_SETUP support for input/*.c.
	  let g:c_comment_strings = 1
	  syntax on
	  " END [runtime/defaults.vim]
```

zzzyxwvut:
The changes are up, please review.

Note that I've tightened up the patterns matched against by  
`search()` and `substitute()` so that we don't embarrass  
ourselves with flukes like:

```vim
" vim:set ignorecase nosmartcase:
"
" ................................ Let's assign g:vim_test_setup
" to "NVIM_TEST_SETUP!".
" VIM_TEST_SETUP let g:vim_test_setup = "NVIM_TEST_SETUP!"
```

chrisbra:
thanks!

zzzyxwvut:
@chrisbra, there seems to be outdated file versions in the master. 

In my fork,

```shell
git switch master
git pull --ff-only --verbose
git diff a9bc195d6497a18b24428a0271a2839213444baf runtime/syntax/setup/20240316 -- runtime/syntax/testdir/{runtest.vim,README.txt}
```

```diff
diff --git a/runtime/syntax/testdir/README.txt b/runtime/syntax/testdir/README.txt
index 40757759b..9c1264831 100644
--- a/runtime/syntax/testdir/README.txt
+++ b/runtime/syntax/testdir/README.txt
@@ -34,7 +34,7 @@ Make sure to include some interesting constructs with plenty of complicated
 highlighting.  Optionally, pre-configure the testing environment by including
 setup commands at the top of the input file. The format for these lines is:
 
-	TEST_SETUP {command}
+	VIM_TEST_SETUP {command}
 
 where {command} is any valid Ex command, which extends to the end of the line.
 The first 20 lines of the input file are ALWAYS scanned for setup commands and
@@ -44,8 +44,8 @@ errors in the input file but this is not required.
 
 Continuing the Java example:
 
-	// TEST_SETUP let g:java_space_errors = 1
-	// TEST_SETUP let g:java_minlines = 5
+	// VIM_TEST_SETUP let g:java_space_errors = 1
+	// VIM_TEST_SETUP let g:java_minlines = 5
 	class Test { }
 
 As an alternative, setup commands can be included in an external Vim script
@@ -59,7 +59,7 @@ an "input/setup/java.vim" script file with the following lines:
 	let g:java_minlines = 5
 
 Both inline setup commands and setup scripts may be used at the same time, the
-script file will be sourced before any TEST_SETUP commands are executed.
+script file will be sourced before any VIM_TEST_SETUP commands are executed.
 
 If there is no further setup required, you can now run the tests:
 
diff --git a/runtime/syntax/testdir/runtest.vim b/runtime/syntax/testdir/runtest.vim
index 4802315ba..e1cfdcfe1 100644
--- a/runtime/syntax/testdir/runtest.vim
+++ b/runtime/syntax/testdir/runtest.vim
@@ -149,17 +149,18 @@ func RunTest()
 
 	func SetUpVim()
 	  call cursor(1, 1)
-	  " Defend against rogue TEST_SETUP commands.
+	  " Defend against rogue VIM_TEST_SETUP commands.
 	  for _ in range(20)
-	    let lnum = search('\<TEST_SETUP\>', 'eW', 20)
+	    let lnum = search('\C\<VIM_TEST_SETUP\>', 'eW', 20)
 	    if lnum < 1
 	      break
 	    endif
-	    exe substitute(getline(lnum), '.*TEST_SETUP', '', '')
+	    exe substitute(getline(lnum), '\C.*\<VIM_TEST_SETUP\>', '', '')
 	  endfor
 	  call cursor(1, 1)
 	  " BEGIN [runtime/defaults.vim]
-	  set display=truncate ruler scrolloff=5
+	  " Also, disable italic highlighting to avoid issues on some terminals.
+	  set display=truncate ruler scrolloff=5 t_ZH= t_ZR=
 	  syntax on
 	  " END [runtime/defaults.vim]
 	  redraw!
```

Please double check.

zzzyxwvut:
Almost there.  :)

Please generate new `dumps/` versions from `input/c.c` and  
`input/markdown_conceal.markdown`.

habamax:
It would be very handy if available without vim sources or without need of copying into vim src directory.

chrisbra:
that will be hard, since it requires the src/testdir setup as dependency (and it requires a similar directory structure between $VIMRUNTIME and the source directory.. It's probably easier to just clone the vim repository. (or use the zip download button from github directly)

PR #14216, opened 2024-03-17
'wincolor' hl missing with 'rightleft', "below" virtual text, 'nowrap'

zeertzjq:
Problem:  'wincolor' highlight missing with 'rightleft', "below" virtual
          text and 'nowrap'.
Solution: Handle 'rightleft' in draw_screen_line().


chrisbra:
thanks!

PR #14218, opened 2024-03-17
Wrong cursor pos when clicking after eol with 'rl', 've' and conceal

zeertzjq:
Problem:  Wrong cursor position when clicking after end of line with
          'rightleft', 'virtualedit' and conceal.
Solution: Set values in ScreenCols[] also with SLF_RIGHTLEFT.  Also fix
          off-by-one cursor position with 'colorcolumn'.


chrisbra:
thanks!

PR #14221, opened 2024-03-17
runtime(rust): Respect no_plugin_maps and no_rust_maps globals

myypo:
Makes it possible to opt-out of rust ft buffer mappings the same way it works for other ft plugins.


MyyPo:
@chris-morgan Hello, I hope it is ok. This problem has been bugging me for quite a while.

chrisbra:
Thanks. There isn't much activity over at https://github.com/rust-lang/rust.vim so let me include it now.

PR #14226, opened 2024-03-18
Dafny files are not recognized

zeertzjq:
Problem:  Dafny files are not recognized.
Solution: Recognize *.dfy files as filetype "dafny".

Ref: https://dafny.org/
Ref: https://github.com/mlr-msft/vim-loves-dafny


chrisbra:
thanks!

PR #14228, opened 2024-03-18
runtime(vim): Update base-syntax, improve :highlight command

dkearns:
Improve :highlight command highlighting

- Use the same highlight groups for "default link" with and without
  bang.
- Match some common line-continuation use.
- Match :hi clear variants.
- Highlight color-name values.

Resync vim.vim and generator/vim.vim.base.


chrisbra:
thanks!

PR #14229, opened 2024-03-18
filetype: add Vento *.vto

wrapperup:
Vento is a templating engine (https://vento.js.org/), this adds a filetype for `*.vto` files


chrisbra:
thanks!

PR #14232, opened 2024-03-19
runtime(vim): Update base-syntax, disallow '.' at start of menu item names

### LINKED ISSUE #14230: vim9script syn hl vimMenuName steals autoloaded function call hl
### Steps to reproduce

1. `vim -Nu NONE`
2. `:set ft=vim`
3. `:syntax on`
4. paste following snippet
```
export def HistoryJumpMenu()
    popup.FilterMenu("Jump history", dir_hist,
        (res, _) => {
            HistoryJump(res.text)
        })
enddef
```
image

### Expected behaviour

'string' should be highlighted as String 

ping @dkearns 

### Version of Vim

9.1.187

### Environment

debian, bash, st

### Logs and stack traces

_No response_

dkearns:
Disallow '.' at the start of a menu item name.

This is the menu path separator character and should be escaped with a
'\' in this case.

Partially fixes #14230.  "popup" is still incorrectly matched as the Ex
command.


chrisbra:
thanks!

PR #14235, opened 2024-03-20
runtime/colors/lists/default.vim: avoid unneeded renorm; don't re-parse additions

### LINKED ISSUE #14234: performance: colors/lists/default.vim: slow, unneeded repeated re-normalization
Currently `runtime/colors/lists/default.vim` adds some default colors, then re-normalizes `keys(v:colornames)` to lower-case every time the script is loaded (it gets loaded repeatedly, at least if `vimrc` uses `colo`). If the addition doesn't actually add new keys, no re-normalization is needed.

Also, the added colors are re-parsed on every reload.

I'm about to add a PR that addresses both issues. Sometimes even the first re-normalization isn't triggered (i.e. the added colors are already in `v:colornames`). Just to make sure, I've manually added invalid colornames, and then the re-normalization *does* trigger.

I've checked with `--startuptime`. Avoiding re-normalization halves the running time on my machine. Avoiding re-parsing (from the first re-load on) pushes the timings into 0.1 msec range.

mralusw:
Fixes #14234

### DIFF
--- a/runtime/colors/lists/default.vim
+++ b/runtime/colors/lists/default.vim
@@ -18,6 +18,32 @@ function! s:Cleanup()
   unlet s:keepcpo
 endfunction
 
+function! s:AddColors(cnames) abort
+  call extend(v:colornames, a:cnames, 'keep')
+
+  " all keys should be in lower case, convert keys that are not yet
+  let len_after = len(v:colornames)
+  if len_after == len(a:cnames)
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. I guess we need the same logic for `runtime/colors/lists/csscolors.vim` ? 

Also, we should probably document the fact, that if you want to make changes to a colorname, you need to add a dummy entry to make sure it is reloaded properly.

PR #14240, opened 2024-03-20
runtime(deb822sources): Add minimal ftplugin

jamessan:
Set comment related options and avoid automatic line wrapping.

### DIFF
--- a/runtime/ftplugin/deb822sources.vim
+++ b/runtime/ftplugin/deb822sources.vim
@@ -1,0 +1,16 @@
+" Language:     Debian sources.list
+" Maintainer:   Debian Vim Maintainers <team+vim@tracker.debian.org>
+" Last Change:  2024 Mar 20
+" License:      Vim License
+" URL:          https://salsa.debian.org/vim-team/vim-debian/blob/main/ftplugin/deb822sources.vim
+
+if exists('b:did_ftplugin')
+  finish
+endif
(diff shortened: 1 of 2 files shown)


dkearns (approved):

chrisbra:
thanks!

runtime/ftplugin/deb822sources.vim
```diff
@@ -0,0 +1,16 @@
...
+" Last Change:  2024 Mar 20
+" License:      Vim License
+" URL:          https://salsa.debian.org/vim-team/vim-debian/blob/main/ftplugin/deb822sources.vim
+
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin=1
+
+setlocal comments=:#
```
chrisbra, line 12: Wondering if you rather would like to use `comments=b:#` instead?

runtime/ftplugin/deb822sources.vim
```diff
@@ -0,0 +1,16 @@
...
+setlocal commentstring=#%s
```
chrisbra, line 13: Same here? commentstring=#\ %s

chrisbra (approved):

chrisbra:
I had 2 tiny comments, but I am fine with it and I'll merge it now. I guess you can consider this for later (or not at all :) )

PR #9588, opened 2022-01-21
feat(filetype): add detection for gnuplot config

nobodyatandnothing:

PR #14243, opened 2024-03-20
Update `filetype.vim`: Fixed Gnuplot detection

RobbiZ98:
Gnuplot files with the ".gnuplot" extension are now correctly recognised (added missing asterisk)

### DIFF
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -283,7 +283,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     glsl: ['file.glsl'],
     gn: ['file.gn', 'file.gni'],
     gnash: ['gnashrc', '.gnashrc', 'gnashpluginrc', '.gnashpluginrc'],
-    gnuplot: ['file.gpi', 'file.gnuplot'],
+    gnuplot: ['file.gpi', '.gnuplot', 'file.gnuplot'],
     go: ['file.go'],
     gomod: ['go.mod'],
     gosum: ['go.sum', 'go.work.sum'],
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -283,7 +283,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     glsl: ['file.glsl'],
     gn: ['file.gn', 'file.gni'],
     gnash: ['gnashrc', '.gnashrc', 'gnashpluginrc', '.gnashpluginrc'],
-    gnuplot: ['file.gpi', '.gnuplot'],
+    gnuplot: ['file.gpi', 'file.gnuplot'],
     go: ['file.go'],
     gomod: ['go.mod'],
     gosum: ['go.sum', 'go.work.sum'],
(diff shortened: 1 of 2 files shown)


zeertzjq:
It seems that this is no official standard extension for Gnuplot files: https://stackoverflow.com/questions/5497889/is-there-a-standard-file-extension-for-gnuplot-files

RobbiZ98:
He, one of the answers on that question re-refers to how vim does it via Wikipedia :-D
How interesting and annoying. `.gnuplot` is how I learned it in uni a few years ago.

`.gp` is already associated with GP (which seems to be a block-based language like Scratch) in vim, the other extensions mentioned on Stackoverflow are potentially free to be associated with Gnuplot (although I find `.gnu` a weird choice, given that Gnuplot is not associated with GNU, but that's beside the point). 
The other extensions mentioned are: `.gpl`, `gnu`, `.gplt`, `.gnu`, `.plt`, `.plot`

chrisbra:
so is `foobar.gnuplot` used within the gnuplot community?

RobbiZ98:
Originally I thought that I had encountered a bug, but it turns out the inclusion of `.gnuplot` (without asterisk) by #9588 refers to the Gnuplot configuration file, as per the manual: http://gnuplot.info/docs_5.5/loc4369.html

I will point out that I am not the first to make this mistake, this comment on the aforementioned Stackoverflow question also assumes that the `.gnuplot` in the current vim master is supposed to refer to the file extension, not the config file:
https://stackoverflow.com/questions/5497889/is-there-a-standard-file-extension-for-gnuplot-files#comment133670299_7133952

@dkearns wrote:
> It would appear so, there's [160 thousand hits](https://github.com/search?q=path%3A*.gnuplot&type=code) on Github.

Apparently I am not the only one who thinks `.gnuplot` is a valid file extension for Gnuplot files, so I am leaving the pull request open.

chrisbra:
So I assume the gnuplog configuration file requires different syntax highlighting than the actual gnuplot files? Is this correct? Then we would need a separate filetype for `.gnuplot` files (the configuration files, like `gnuplotrc`) and actual gnuplot files.

src/testdir/test_filetype.vim
```diff
@@ -283,7 +283,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     glsl: ['file.glsl'],
     gn: ['file.gn', 'file.gni'],
     gnash: ['gnashrc', '.gnashrc', 'gnashpluginrc', '.gnashpluginrc'],
-    gnuplot: ['file.gpi', '.gnuplot'],
+    gnuplot: ['file.gpi', 'file.gnuplot'],
```
chrisbra: does `*` no longer match `.gnuplot` files itself?

RobbiZ98:
According to the old doc, the .gnuplot config file should be standard gnuplot code and not require special highlighting rules: http://www.gnuplot.info/docs_4.2/gnuplot.html#x1-6900018
The newest version of the docs does not mention this anymore, but I find it unlikely that they changed it.

With the added asterisk, the .gnuplot config file is still correctly recognised. I can't speak for the test, as I honestly don't fully understand that and only followed the example of previous commits.

chrisbra:
okay thanks. In that case, please leave the `.gnuplot` file in the test and just add another `file.gnuplot`.

chrisbra:
Thanks!

PR #14244, opened 2024-03-20
problem: cannot use :e with 'winfixbuf' enabled

### LINKED ISSUE #14237: 'winfixbuf' should allow :edit without args
## Problem

Raised by @tpope at https://github.com/tpope/vim-fugitive/issues/2272#issuecomment-2004992176 : 

> This option may be too aggressive. It prevents reloading the window by calling `:e` with no arguments. That doesn't change what buffer the window points to, but it is nonetheless blocked.

## Expected behavior

'winfixbuf' should not show an error if `:edit` is used without args to reload the buffer (which may be application-defined for non-file buffers).

cc @ColinKennedy

chrisbra:
Problem:  cannot use :e with 'winfixbuf' enabled
Solution: allow :e without args

closes: #14237

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -7252,6 +7252,7 @@ ex_edit(exarg_T *eap)
     if (
 	    eap->cmdidx != CMD_badd
 	    && eap->cmdidx != CMD_balt
+	    && !(eap->cmdidx == CMD_edit && *eap->arg == NUL)
 	    // All other commands must obey 'winfixbuf' / ! rules
 	    && !check_can_set_curbuf_forceit(eap->forceit))
         return;
--- a/src/testdir/test_winfixbuf.vim
+++ b/src/testdir/test_winfixbuf.vim
@@ -3283,4 +3283,25 @@ func Test_bufdo_cnext_splitwin_fails()
   set winminheight&vim winheight&vim
 endfunc
 
+" Fdit without filename is allowed
+func Test_edit_noarg()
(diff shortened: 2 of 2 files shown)


ColinKennedy:
@chrisbra Thank you for the quick fix. One thing wanted to mention though - there's 2 more cases that we may need to consider

1. `:edit` (no args, exists your PR already)
2. `:edit foo` (file lives on disk, current buffer is already foo)
3. `:edit foo` (file is an in-memory buffer that does not live on disk, current buffer is already foo)

I started an exploratory branch here https://github.com/neovim/neovim/compare/master...ColinKennedy:neovim:issues/14237-neovim_edit_without_args

In short, `Test_edit_same_buffer_in_memory` is the only test that doesn't work. But I suspect this might be a neovim-only issue. The issue was because the "is this buffer name the same as the current" relies on a function that [expects the buffer name to exist on-disk](https://github.com/neovim/neovim/blob/734848dc1a9ebdf7af26cf54b92133cc6d94e19c/src/nvim/buffer.c#L3140C23-L3140C32). Vim on the other hand [does the same check differently](https://github.com/vim/vim/blob/70e566b0bd29189b1baf3dc915b6a98de36fbded/src/buffer.c#L3738-L3765) so I planned on trying that next.

Anyway in short, I think we should also account for (2) and (3), if not in this PR but in another follow-up

chrisbra:
yes makes sense. Then I await an updated PR.

PR #14249, opened 2024-03-21
May leak memory in completion when ga_grow() fails

### LINKED ISSUE #14248: There is a memory leak defect at line 656 in the file /vim/src/insexpand.c.
### Steps to reproduce

1.At line 549 in the file /vim/src/insexpand.c, a pointer variable named 'wca' is defined. This pointer variable is allocated a block of dynamic memory by the function ALLOC_MULT at line 559. When the if statement at line 560 returns false, it indicates that the dynamic memory allocation for the pointer 'wca' was successful. However, when the if statement at line 655 returns true, the program will return at line 656 without executing the vim_free(wca) operation at line 666. This results in a memory leak for the memory space pointed to by the 'wca' pointer. The detailed path of the defect is illustrated in the diagram below:
https://github.com/LuMingYinDetect/vim_defects/blob/main/vim_8.png

### Expected behaviour

If this memory leak defect is true, it is advisable to fix this defect.

### Version of Vim

9.1.0192

### Environment

Ubuntu 22.04

### Logs and stack traces

_No response_

zeertzjq:
Problem:  May leak memory in completion when ga_grow() fails.
Solution: Free "wca" when ga_grow() fails.

Fix #14248

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -653,7 +653,10 @@ ins_compl_infercase_gettext(
 	    // getting to six bytes from the edge of IObuff switch to using a
 	    // growarray.  Add the character in the next round.
 	    if (ga_grow(&gap, IOSIZE) == FAIL)
+	    {
+		vim_free(wca);
 		return (char_u *)"[failed]";
+	    }
 	    *p = NUL;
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

PR #14256, opened 2024-03-21
Improve the manageability of the installer compilation

RestorerZ:
Provides a single installer build management interface from the command line of
the program "makensis.exe".
Allows you to avoid unintended errors during direct manual editing or other ways
of changing the file "gvim.nsi".


nsis/README.txt
```diff
@@ -1,7 +1,7 @@
-This builds a one-click install for Vim for Win32 using the Nullsoft
+This builds a one-click install for Vim for OS Windows using the Nullsoft
```
k-takata: I've never seen the wording "OS Windows" in other part of Vim document.
"MS-Windows" might be better?

nsis/README.txt
```diff
@@ -16,10 +16,10 @@ To build the installable .exe:
 	install.exe,
 	uninstall.exe,
 	tee/tee.exe,
-	xxd/xxd.exe,
+	xxd/xxd.exe
 
-    Then execute tools/rename.bat to rename the executables. (mv command is
-    required.)
+    Then execute tools/rename.bat to rename the executables (on Unix-like,
+    it must first be made executable; chmod +x path/to/vimsrc/tools/rename.bat).
```
k-takata: How about setting the executable flag in this repository beforehand?

nsis/README.txt
```diff
@@ -61,25 +71,39 @@ To build the installable .exe:
...
+Download Unicode version the ShellExecAsUser plug-in for NSIS from:
+	https://nsis.sourceforge.io/ShellExecAsUser_plug-in
+and put ShellExecAsUser.dll to path\to\NSIS\Plugins\x86-unicode
 
 
 Unpack the images:
 	cd nsis
-	unzip icons.zip
+	unzip icons.zip (on Unix-like)
+	7z.exe x icons.zip or WinRar.exe x icons.zip (on OS Windows)
```
k-takata: There is also Windows version of unzip, and there is also Linux version of 7z (p7zip).

RestorerZ: I agree. But if in Unix-like systems "unzip" is practically part of the system, then in Windows it is usually only if Gnuwin32 or something similar is installed.
I proceeded from the fact that it is most common in Windows, and 7z is more likely to be installed by the user.
But you're right, it's better to clarify this in the documentation.

nsis/README.txt
RestorerZ: Oh! That would be great! Could you tell me how to do it? It's just that my main tool is Windows.
I'm going to try to log in to the Ubuntu via VM and set this flag in my fork. Will that be enough?

nsis/README.txt
RestorerZ: Yes, it would be better that way. MS Windows.

nsis/README.txt
k-takata: Changing the flag on a unix-like system should be the easiest way and that should be enough.
However, on Windows, you can use `git update-index --add --chmod=+x <filename>`.

RestorerZ: Thank you very much! Everything worked out! Thanks!

chrisbra:
Thanks!

PR #14257, opened 2024-03-21
add errorformat for dot compiler

Konfekt:
Left here as a suggestion for reference as the current maintainer could not be reached by mail


chrisbra:
thanks, let me include it for now.

Issue #14261, opened 2024-03-21
Ruler vcol inconsistently includes 'above' virtual text


dylanahsmith:
### Steps to reproduce

1. Save the following as test.vim
```vim
vim9script
:40vnew
setlocal number
setline(1, ['some text'])

prop_type_add('test', {highlight: 'ToDo'})
prop_add(1, 0, {type: 'test', text: 'Above', text_align: 'above'})
normal l
```
2. `vim --clean -S test.vim` and notice that the ruler shows `1,2-38`
3. Press `gk` and notice that the ruler now shows `1,2` without the cursor moving

### Expected behaviour

I didn't expect the ruler to show any difference as a result of `gk` on the first line.

I would expect the 'above' aligned virtual text to not affect the vcol in the ruler, so I would expect `1,2` even before `gk` is typed.  The ruler currently seems to normally include the 'above'  aligned virtual text, so I wanted to confirm that this is a bug before looking into how best to fix this.

### Version of Vim

9.1.0193

chrisbra:
Ah, I have seen this before with wrapped lines and virtual text. Currently the ruler change is the only indication that `gk`/`gj` does indeed do something, even so the cursor does not seem to move and one may have to type it several times until the cursor moves to the next screen line. However, I would agree, that this behaviour is rather annoying and I think the `gk` should directly skip to the next screenline. I guess if we make this work, it would also fix this ruler behaviour.

PR #14262, opened 2024-03-21
Fix `gj`/`gk`/`gM` to avoid being affected by non-inline virtual text

### LINKED ISSUE #12028: If 'list' is on, 'below' virtual text which includes 1 or 2 characters are gone

dylanahsmith:
Fixes #12028

## Problem

As mentioned in https://github.com/vim/vim/issues/12028#issuecomment-1966441523, `gj`/`gk` needed to be repeated for each virtual text line to skip over them.  It was actually moving the `w_curswant` position onto the outer virtual text, as if the outer virtual text (and their padding) was a part of the line, even though the actual cursor position would be shifted to stay in the actual text.

I found a similar issue with `gM` including the outer virtual text in the line length used to calculate where the middle of the line is to move the cursor there.

## Solution

I've added a `linetabsize_no_outer` function in order to get the line length with outer virtual text properties excluded.  We can't just exclude text properties entirely, since we do want the influence of inline virtual text in these cases.  I've then used that to get the line length in the functions to handle `gj`/`gk`/`gM` cursor movement.

Note that I had tested `gk` on the first line and `gj` on the last line to ensure the cursor doesn't move, however, I omitted that from the included regression test due to issue #14261 causing the ruler to change when I would otherwise verify that the screen didn't change.  Also, the regression tests screen dumps will need updating if we exclude the 'above' virtual text from the ruler's vcol as I suggested in that issue.


src/charset.c
```diff
@@ -798,6 +798,40 @@ linetabsize(win_T *wp, linenr_T lnum)
...
+    init_chartabsize_arg(&cts, wp, lnum, 0, line, line);
+
+    int write_idx = 0;
+    for (int read_idx = 0; read_idx < cts.cts_text_prop_count; read_idx++)
+    {
+	textprop_T *tp = &cts.cts_text_props[read_idx];
+	if (tp->tp_col != MAXCOL)
+	{
+	    if (read_idx != write_idx)
+		cts.cts_text_props[write_idx] = *tp;
```
chrisbra: I think you are leaking memory here.

dylanahsmith: Oh, not here, but if `cts.cts_text_prop_count` gets changed to `0` through this filtering, then `cts.cts_text_props` won't be freed by `clear_chartabsize_arg`.

I've pushed a fixup commit to fix the memory leak.  Would you like me to rebase and squash the fixup?

chrisbra: if you could squash, that would be prefered, thanks!

chrisbra:
thanks. I think this makes sense.

chrisbra:
Hi @dylanahsmith Coverity complains about 
grafik

I think this is a false positive, right?

dylanahsmith:
Yeah, that looks like a false positive, since `cts_text_prop_count` is 0 if `cts_text_props` is `NULL`, and has a branch ensuring that if the allocation of `cts_text_props` fails.

chrisbra:
Yes thanks that was also what I thought. Just wanted to get a second opinion about it. I have now marked it false positive in coverity.

PR #14267, opened 2024-03-22
Fix for Sodium version 1.0.19 and MinGW

RestorerZ:
It should, in theory, solve the problem with MinGW and libsodium ver. 1.0.19


chrisbra:
thank you!

PR #14268, opened 2024-03-22
Re-use 'smoothscroll' logic for page-wise scrolling

luukvbaal:
Problem:  Page-wise scrolling with Ctrl-F/Ctrl-B implements
          it's own logic to change the topline and cursor.
          In doing so, skipcol is not handled properly for
          'smoothscroll', and virtual text properties.
Solution: Re-use the logic from Ctrl-E/Ctrl-Y while staying
          backward compatible as much as possible.


luukvbaal:
Reported at https://github.com/neovim/neovim/issues/27967.

There are some slight numerical differences but scrolling works. In particular with wrapped lines and filler lines there are some differences, which onepage() handled on it's own. Thos now work slightly worse with 'smoothscroll' disabled and better with 'smoothscroll' enabled. Hope that it is backward compatible enough. Change seems worth it to me.

src/testdir/test_diffmode.vim
```diff
@@ -1664,11 +1664,11 @@ func Test_diff_scroll_many_filler()
...
-  call assert_equal(19, winsaveview().topfill)
+  call assert_equal(1, winsaveview().topline)
+  call assert_equal(0, winsaveview().topfill)
   exe "normal! \<C-B>"
   redraw
-  call assert_equal(2, winsaveview().topline)
+  call assert_equal(1, winsaveview().topline)
   call assert_equal(0, winsaveview().topfill)
   exe "normal! \<C-B>"
   redraw
```
zeertzjq: This test may need more lines at the top, otherwise the last two `<C-B>` have no room to scroll.

src/testdir/test_normal.vim
```diff
@@ -4172,20 +4170,30 @@ func Test_normal34_zet_large()
...
+  call prop_type_add(vt, {'highlight': 'IncSearch'})
+  call prop_add(1, 0, {'type': vt, 'text': '---', 'text_align': 'above'})
   1
   norm! 0}
 
   call assert_equal([0, 1, 20, 0], getpos('.'))
   norm! {
   call assert_equal([0, 1, 1, 0], getpos('.'))
+
+  " Ctrl-B scrolls up with hidden virt_above text property.
```
zeertzjq:
```suggestion
  " Ctrl-B scrolls up with hidden "above" virtual text.
```

zeertzjq:
`:h 'smoothscroll'` should also need change.

chrisbra:
Hi,
thanks. In general, I think this is a good change. But can you please give more details on this part?

> In particular with wrapped lines and filler lines there are some differences, which onepage() handled on it's own. Those now work slightly worse with 'smoothscroll' disabled and better with 'smoothscroll' enabled. Hope that it is backward compatible enough

What does it mean it works worse with 'smoothscroll' disabled?

luukvbaal:
> What does it mean it works worse with 'smoothscroll' disabled?

To be honest, I'm not even sure it's worse. I would have to inspect the old logic more closely and try to qualify the difference in behavior. I just replaced it outright because scrolling I thought using the logic that scrolls X lines would be similar enough to scrolling a page.

luukvbaal:
The current(old) behavior with filler lines seems strictly worse to me. What's more, with `:se sms` it can result in flip-flopping between two toplines, thus not being able to scroll up or down. The PR prevents that.

The new behavior with large wrapped lines and `:se sms` is unfortunately not great, we end up skipping much of the buffer. I think temporarily setting 'smoothscroll' in scrollup(), then restoring the old value and setting skipcol to zero if sms was unset would improve it. I'll test and push that.

chrisbra:
thanks!

PR #14269, opened 2024-03-22
Short circuit evaluation clears the first value

### LINKED ISSUE #14265: [vim9script] Unexpected results for `!false || !false` and `!true && !true` in `:def` style function
### Steps to reproduce

For these 2 statements (`!false || !false` and `!true && !true`):
- If each statement's second operator is enclosed in `()`, the results are as expected (returning true / false respectively);
- If executed at the vim9script level (outside of def functions), the results are also as expected;
- If executed within a legacy function, the results are as expected as well.

Let me express it in a vim9script:

```vim
vim9script

# Function F() serves as a demonstration of the bug;
# the other parts are provided for reference only (to illustrate that in other cases, the behavior is as expected)

def F()
    echo 'unexpected behavior:'
    echo 'echo !false || !false'
    echo !false || !false
    echo 'echo !true && !true'
    echo !true && !true
enddef
F()

echo '---'

echo 'expect true'
echo !false || !false
echo !false || (!false)
echo 'expect false'
echo !true && !true
echo !true && (!true)

def F2()
    echo 'expect true'
    echo !false || (!false)
    echo 'expect false'
    echo !true && (!true)
enddef
F2()

legacy function! s:F3()
    echo 'expect true'
    echo !v:false || !v:false
    echo !v:false || (!v:false)
    echo 'expect false'
    echo !v:true && !v:true
    echo !v:true && (!v:true)
endfunction
legacy call s:F3()
```

### Expected behaviour

`!false || !false` always returns true; `!true && !true` always returns false.

### Version of Vim

9.1.178

### Environment

Verified on linux (alpine linux) and macos.

### Logs and stack traces

_No response_

yegappan:
Fixes #14265.

### DIFF
--- a/src/testdir/test_vim9_expr.vim
+++ b/src/testdir/test_vim9_expr.vim
@@ -311,6 +311,33 @@ def Test_expr2()
       assert_equal([1], g:vals)
   END
   v9.CheckDefAndScriptSuccess(lines)
+
+  # test the short-circuit operation
+  lines =<< trim END
+    assert_equal(true, true && true)
+    assert_equal(false, true && !true)
+    assert_equal(false, !true && true)
+    assert_equal(false, !true && !true)
+
+    assert_equal(true, true || true)
+    assert_equal(true, true || !true)
+    assert_equal(true, !true || true)
+    assert_equal(false, !true || !true)
+
+    assert_equal(false, false && false)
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #14273, opened 2024-03-22
Refactor to avoid STRLEN() calls (take2)

basilisk0315:
I have added a new structure to store key (int)/value (string) pairs for use in static lookup tables (eg event names, regexp character classes, special keywords, highlighting). This structure embeds the length of the string which allows removal of calls to STRLEN() when performing the lookup comparisons.

In addition to factoring out calls to STRLEN() I have added a simple circular FILO caching mechanism to the event_nr2name() lookup functions with the goal to speed it up. Other lookup functions now use a binary search (bsearch()). This meant changing the sequencing of the various lookup compiled-in arrays (tables) and a test (test_highlight.vim). This is much simpler than my previous attempt with caching which I also found would be messy (and a little error prone) to use in a general case.

I also addressed a potential infinite loop in the test test_codestyle.vim.


basilisk0315:
Most (all?) of the failing checks look like this:

> Error detected while processing /home/runner/work/vim/vim/runtime/plugin/matchparen.vim:
> line   25:
> E216: No such group or event: WinScrolled * call s:Highlight_Matching_Pair()
> Error: The action 'Test' has timed out after 25 minutes.

I am not able to reproduce the E216 error.
I don't know why the test is timing out.

Any help would be appreciated.
Thanks
John

chrisbra:
Thanks for working on this. Can we please break this up into smaller logical units?

basilisk0315:
Will do.

basilisk0315:
Closing this in order to break the change into smaller pieces.

PR #14275, opened 2024-03-23
Not enough tests for the slice() function

zeertzjq:
Problem:  Not enough tests for the slice() function.
Solution: Test with multibyte chars, and in both Legacy and Vim9 script.
          Update docs to be clearer about how it treats composing chars.


chrisbra:
thanks!

Issue #14272, opened 2024-03-22
`<C-[>` still broken for certain Cyrillic keyboard layouts


wildekat:
### Steps to reproduce

1. Run `gvim -u NONE`
2. Enter Insert mode
3. Switch to Russian (I'm using the standard/most common [ЙЦУКЕН keyboard layout](https://en.wikipedia.org/wiki/JCUKEN))
4. Press `Ctrl+Cyrillic letter х` to attempt to exit Insert mode
5. Observe how nothing happens

### Expected behaviour

Before Vim-8.2.4807, `<C-х>` was fully equivalent to `<C-[>` / `<Esc>`, in any mode (`[` and `Х` are one and the same key).

### Version of Vim

9.1.0181

### Environment

OS: Windows 10
`$TERM`: N/A
On Windows, the issue is only present in gVim and does not seem to affect console Vim at all.

RestorerZ:
Please take a look [here](https://github.com/vim/vim/issues/12595) and [here](https://github.com/vim/vim/pull/12752).

wildekat:
@RestorerZ
Thanks, I've been through those threads and a number of other similar ones.

If I were to set `set_keycode_trans_strategy` to `experimental`, that would also require setting up *hundreds* of explicit `<C-*>` mappings: `<C-ц>` would have to be mapped to `<C-w>`, `<C-г>` to `<C-u>` and so on. Clearly that's not the way to go.

The problem with `set_keycode_trans_strategy == classic` is that it does not quite restore the old (pre-8.2.4807) behavior: in this mode, `<C-х>` basically does not exist, so you cannot even map it meaningfully (but thankfully all the other `Ctrl+Cyrillic letter` combinations seem to work out of the box).

wildekat:
Sorry, my bad. It turns out <C-х> does NOT work in Vim-8.

ant0sha:
Question is, if it is meaningful feature request then...

wildekat:
I for one would very much like to see this implemented.
One can actually map `<C-х>` to `<C-[>` under `set_keycode_trans_strategy = experimental`, but for that solution to be viable, we would need an equivalent of `langmap` for Ctrl key bindings.

ant0sha:
Thanks to @RestorerZ explanations in PR #14277, right way to
solve this issue seems to be to modify slightly the standard RU
keyboard layout.

Using MSFT "Keyboard Layout Creator" I created custom RU layout
(it is possible to load the existing layout as the starting
[...]

PR #14277, opened 2024-03-23
(feat) Ctrl+[ always sends ESC, even on russian keyboard (#14272)

ant0sha:
issue #14272 is marked as bug and is closed;
This PR is made from the assumption that it is useful feature request and tries to fix it.

According to my debugging, reason why with RU keyboard layout we do not get ``Ctrl+[`` seems to be that TranslateMessage() simply silently ignores this key combo, so need to intercept it earlier and inject "ESC" directly into vim input_buf.

This happens on the place where we do not have any idea about if we are on RU layout or on some else keyboard layout, therefore it does change handling logic for this key combo for _all_ keyboard layouts. We can think about restricting it effect to only RU layout by checking MapVirtualKey(vk, 2) == '?', which seems to be the result of that call for RU (EN: '[', DE: 'ü', ... probably another layouts will exist which do return '?' as well), right now such restriction is not made. Any thoughts?

Not 100% sure as well that we need ``MapVirtualKey(vk, 2) == '['`` comparison...

Anyway patch seems to work fine in my tests with following keyboard layouts: EN, RU, DE, FR

Would be nice as well if someone will test this PR on console windows vim, I made my tests only with gvim.

### DIFF
--- a/src/gui_w32.c
+++ b/src/gui_w32.c
@@ -2271,6 +2271,11 @@ static void process_message_usual_key_classic(UINT vk, const MSG *pmsg)
 	    string[0] = Ctrl_AT;
 	    add_to_input_buf(string, 1);
 	}
+	else if (vk == 219 || MapVirtualKey(vk, 2) == (UINT)'[')
+	{
+	    string[0] = 27; // ESC
+	    add_to_input_buf(string, 1);
+	}
 	else
 	    TranslateMessage(pmsg);
     }


RestorerZ:
> According to my debugging, reason why with RU keyboard layout we do not get `Ctrl+[` seems to be that TranslateMessage() simply silently ignores this key combo, so need to intercept it earlier and inject "ESC" directly into vim input_buf.

Well, it's not that she ignores this keyboard shortcut.

The fact is that in the Russian keyboard layout (file kbdru.dll ), most often
used in MS Windows, there is nothing assigned to the keyboard shortcut `CTRL+[`
(`CTRL+x`) at all.
kbdru dll
kbdru-vk_oem_4

If we take the Russian (typewritten) keyboard layout (file kbdru1.dll ), this is
the second Russian layout available in MS Windows, then the `ESC` code will be
transmitted from the keyboard shortcut `CTRL+]` (`CTRL+ъ`).
kbdru1 dll
kbdru1-vk_oem_6


In the American‐English keyboard layout (file kbdus.dll) the `ESC` code is
assigned, as expected, to the keyboard shortcut `CTRL+[`.
kbdus dll
kbdus-vk_oem_4


This is a fragment of the source code of the American‐English keyboard layout:
```c
static ALLOC_SECTION_LDATA VK_TO_WCHARS3 aVkToWch3[] = {
//                      |         |  Shift  |  Ctrl   |
//                      |=========|=========|=========|
  {VK_OEM_4     ,0      ,'['      ,'{'      ,0x001b   },
  {VK_OEM_6     ,0      ,']'      ,'}'      ,0x001d   },
```

This is a fragment of the Russian layout:
```c
SHIFTSTATE

0	// Column 4
1	// Column 5 : Shift
2	// Column 6 :       Ctrl
6	// Column 7 :       Ctrl Alt
3	// Column 8 : Shift Ctrl

LAYOUT		// an extra '@' at the end is a dead key

//SC	VK_		Cap	0	1	2	6	3
//--	---		----	----	----	----	----	----
1a	OEM_4		1	0445	0425	-1	-1		// CYRILLIC SMALL LETTER HA, CYRILLIC CAPITAL LETTER HA, <none>, <none>
1b	OEM_6		1	044a	042a	-1	-1		// CYRILLIC SMALL LETTER HARD SIGN, CYRILLIC CAPITAL LETTER HARD SIGN, <none>, <none>
```

And this is a Russian (typewritten) layout:
```c
SHIFTSTATE

0	// Column 4
1	// Column 5 : Shift
2	// Column 6 :       Ctrl
6	// Column 7 :       Ctrl Alt
3	// Column 8 : Shift Ctrl

LAYOUT		// an extra '@' at the end is a dead key

//SC	VK_		Cap	0	1	2	6	3
//--	---		----	----	----	----	----	----
1a	OEM_4		1	0445	0425	-1	-1		// CYRILLIC SMALL LETTER HA, CYRILLIC CAPITAL LETTER HA, <none>, <none>
1b	OEM_6		1	044a	042a	001b	-1		// CYRILLIC SMALL LETTER HARD SIGN, CYRILLIC CAPITAL LETTER HARD SIGN, <control>, <none>
```

There are a lot of unexpected and illogical things in MS Windows. So you
shouldn't rely on the "default" behavior there.

If I understand your code correctly, then you are emulating just the behavior
for `VK_OEM_4` (code `219`) as in the American‐English layout (adopted in most systems).
But in other layouts, this behavior may be somewhat unexpected.

Here is a good [resource](https://kbdlayout.info/) on different keyboard layouts.
Well, or you can use a tool from MSFT "Keyboard Layout Creator"

chrisbra:
so if this is the case, it sounds like it's an issue with the default russian keyboard layout and I don't think we should try to work-around this in Vim.

ant0sha:
Agree to both @chrisbra and @RestorerZ 

Many thanks by the way, @RestorerZ, for excellent resource
linked and explanations!

PS: Wounder if there is a way to work it around _without_
creating custom keyboard layout - for example tweaking
something in the registry...

PPS: As a last rescue (if not willing to mess with custom
keyboard layout), users can decide themself to apply this tiny
patch and build their custom gvim version.

RestorerZ:
> PS: Wounder if there is a way to work it around _without_
> creating custom keyboard layout - for example tweaking
> something in the registry...

Actually, there are. And there are several of them. I once started to write an article for Habr.com about this and Russian in general in Vim. I should look for that draft and maybe add a brief description in "doc/russian.txt".

LemonBoy:
The problem _is_ with Vim as all the other applications/toolkits correctly translate the `VK_OEM` scancodes into something usable.
This is what the "experimental" translation mechanism tried to do for every key, perhaps it was too radical of a change for some and not so backward compatible, but the idea could be applied only to the `VK_OEM` keys mentioned above.

The problem is not limited to users using the Russian keyboard layout (even if they are the most vocal ones, complaining on the ML) but also people using custom key layouts created using the MSKLC tool.

PR #14286, opened 2024-03-24
'winfixbuf' is now more flexible - You can :edit the current buffer

### LINKED ISSUE #14237: 'winfixbuf' should allow :edit without args

ColinKennedy:
A follow-up to https://github.com/vim/vim/pull/14244

Problem: cannot use :e with 'winfixbuf' enabled
Solution: allow :e without args

closes: https://github.com/vim/vim/issues/14237

## What Is In This PR
With `'winfixbuf'` enabled on a window, the following are now allowed:

- `:edit` (no arguments) 
- `:edit foo` (if foo is in-memory)
- `:edit /tmp/foo` (if /tmp/foo is a currently-saved file)
- Various permutations of "if looking at an in-memory buffer and attempting to switch to an on-disk buffer of the same name (relative or absolute paths / with or without `cd`-ing in advance) will detect the "other" buffer and fail

## What Is Not In This PR
- `:edit foo` (if `:cd`'ed to `/tmp` and `/tmp/foo` is on-disk)

I couldn't figure out a way to determine if a buffer name refers to a buffer that lives on-disk or not. @chrisbra do you have any ideas? I got this equivalent working in Neovim. To get a visual on exactly the problem, please have a look at https://github.com/neovim/neovim/compare/master...ColinKennedy:neovim:issues/14237-neovim_edit_without_args. In particular the `Test_edit_same_buffer_on_disk_relative_path` test. In Neovim the test works because I'm allowed to check `!curbuf->file_id_valid` which covers all cases that I could see. But the Vim equivalent, `buf_T.b_dev_valid`, is only available on UNIX builds and so will fail Windows unittests. So I excluded `Test_edit_same_buffer_on_disk_relative_path` from this PR, because it fails to pass and is still a known issue

Edit: It looks like `Test_edit_same_buffer_on_disk_absolute_path` is also failing on Windows, which I didn't expect. Possibly / probably because of `\` vs `\` and `fnamecmp` behavior.

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -462,7 +462,6 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
 
 /*
  * Check if ffname differs from fnum.
- *
  * fnum is a buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
  * ffname is a full path to where a buffer lives on-disk or would live on-disk.
  *
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -460,6 +460,37 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
 }
 #endif
 
+// @param fnum A buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
+// @param ffname The full path to where a buffer lives on-disk or would live on-disk.
+// @return if `ffname` is the same as `fnum` buffer, return true.
+//
+static int is_other_file(int fnum, char_u *ffname)
+{
+  if (fnum != 0)
+  {
+    if (fnum == curbuf->b_fnum)
+      return FALSE;
+
+    return TRUE;
+  }
+
+  if (ffname == NULL)
+    return TRUE;
+
+  if (*ffname == NUL)
+    return FALSE;
+
+  // TODO: Need a reliable way to know whether a buffer is meant to live on-disk
+  // !curbuf->b_dev_valid is not always available (example: missing on Windows)
+  if (curbuf->b_sfname != NULL
+      && *curbuf->b_sfname != NUL)
+    // This occurs with unsaved buffers. In which case `ffname`
+    // actually corresponds to curbuf->b_sfname
+    return fnamecmp(ffname, curbuf->b_sfname) != 0;
+
+  return otherfile(ffname);
+}
+
 /*
  * do_exmode(): Repeatedly get commands for the "Ex" mode, until the ":vi"
  * command is given.
@@ -7248,12 +7279,15 @@ ex_open(exarg_T *eap)
     static void
 ex_edit(exarg_T *eap)
 {
+    char_u *ffname = eap->cmdidx == CMD_enew ? NULL : eap->arg;
+
     // Exclude commands which keep the window's current buffer
     if (
 	    eap->cmdidx != CMD_badd
 	    && eap->cmdidx != CMD_balt
 	    // All other commands must obey 'winfixbuf' / ! rules
-	    && !check_can_set_curbuf_forceit(eap->forceit))
+	    && (is_other_file(0, ffname) && !check_can_set_curbuf_forceit(eap->forceit))
+    )
         return;
 
     do_exedit(eap, NULL);
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -460,11 +460,15 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
 }
 #endif
 
-// @param fnum A buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
-// @param ffname The full path to where a buffer lives on-disk or would live on-disk.
-// @return if `ffname` is the same as `fnum` buffer, return true.
-//
-static int is_other_file(int fnum, char_u *ffname)
+/*
+ * Check if ffname differs from fnum.
+ *
+ * fnum is a buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
+ * ffname is a full path to where a buffer lives on-disk or would live on-disk.
+ *
+ */
+    static int
+is_other_file(int fnum, char_u *ffname)
 {
   if (fnum != 0)
   {
--- a/src/testdir/test_winfixbuf.vim
+++ b/src/testdir/test_winfixbuf.vim
@@ -1125,6 +1125,146 @@ func Test_edit()
   call assert_equal(l:other, bufnr())
 endfunc
 
+" Fail :e when selecting a buffer from a relative path if in a different folder
+"
+" In this tests there's 2 buffers
+"
+" foo - lives on disk, in some folder. e.g. /tmp/foo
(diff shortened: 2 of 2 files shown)


src/ex_docmd.c
```diff
@@ -460,6 +460,37 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
 }
 #endif
 
+// @param fnum A buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
+// @param ffname The full path to where a buffer lives on-disk or would live on-disk.
+// @return if `ffname` is the same as `fnum` buffer, return true.
+//
+static int is_other_file(int fnum, char_u *ffname)
```
chrisbra: Style: please move `static int` to the previous line (with some indent) and start this line with `is_other_file`

src/ex_docmd.c
```diff
@@ -460,6 +460,37 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
 }
 #endif
 
+// @param fnum A buffer number. 0 == current buffer, 1-or-more must be a valid buffer ID.
```
chrisbra: we are not using doxygen here

src/ex_docmd.c
```diff
@@ -460,6 +460,37 @@ restore_dbg_stuff(struct dbg_stuff *dsp)
...
+      return FALSE;
+
+    return TRUE;
+  }
+
+  if (ffname == NULL)
+    return TRUE;
+
+  if (*ffname == NUL)
+    return FALSE;
```
chrisbra, line 484: I am slightly confused, why is `ffname == NULL` considered TRUE and `*fname == NUL` considered false?

ColinKennedy: From experimentation,

1. Currently `Test_enew` used this. You're right, if `fnum` is `0` and `ffname == NULL` then it would make sense to return `FALSE`, which works for all commands except `enew`.
2. `Test_edit_no_arguments` uses `*ffname == NUL`. When no argument is provided and `fnum` is 0 then that indicates that the user is refreshing the buffer, hence `FALSE`.

IMO the logic of (2) makes sense. (1) is a bit of a stretch though. How about this? We can `ffname == NULL` and return `TRUE` which wouldn't work with `enew` (`Test_enew`) but it will work if we add this check to `ex_edit`

```c
  if (eap->cmdidx == CMD_enew
      && curbuf->b_ffname != NULL
      && *curbuf->b_ffname != NUL
      && !check_can_set_curbuf_forceit(eap->forceit)
  )
    // If the user is calling `:enew` on a buffer that was named,
    // they're trying to switch to a different buffer.
    // Which triggers a 'winfixbuf' error unless [!] is included.
    return;
```

With the check above, an extra benefit is that

```
enew
set winfixbuf
enew
```

doesn't error anymore (currently it does). Another edge case that I just found and need to write a unittest for. Does that sound good to you?

src/ex_docmd.c
ColinKennedy: My bad, this is now done

src/ex_docmd.c
ColinKennedy: I think I got it this time, please let me know if not

src/ex_docmd.c
chrisbra: That should be fine, since the second `enew` re-uses the buffer. I suppose it would error out, if you make some changes, and then `:enew`?

src/ex_docmd.c
chrisbra: thanks!

src/ex_docmd.c
ColinKennedy: Oh I hadn't tried

```
enew
set winfixbuf
do changes
enew again
```

Another edge case. I can make another PR for that

chrisbra:
> Edit: It looks like Test_edit_same_buffer_on_disk_absolute_path is also failing on Windows, which I didn't expect. Possibly / probably because of \ vs \ and fnamecmp behavior.

Yeah, I still see the problem, but it works fine in the vim/vim-win32-installer repo. I can also **not reproduce** the failure in a local build. Not sure what's going on here. I'll disable this teset on Windows.

seandewar:
`tempname()` returns Windows 8.3-style paths on the CI, which also was a headache in #14152; I think this is the reason for the test failing here too.

The following patch fixes the test on the CI, as can be seen [here](https://github.com/seandewar/vim/actions/runs/8454308454/job/23159031604) (ignore the unrelated failures):

```diff
-  " This fails on CI (Windows builds), why?
-  CheckNotMSWindows
   call s:reset_all_buffers()
 
   let file = tempname()
+  call writefile([], file)  " file must exist for expansion of 8.3 paths to succeed
+  let file = fnamemodify(file, ':p')
```

Note that the file must exist for 8.3 expansion to work (without the `writefile()` line, [the test fails](https://github.com/seandewar/vim/actions/runs/8454304161/job/23159016942#step:12:9196), as [`GetLongPathNameW`](https://github.com/vim/vim/blob/194a912d49fb2abea43771ed9bc5eeb8f06d9db0/src/filepath.c#L374) requires the [file to exist](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew#return-value)).

Though this is maybe something `is_other_file` should handle; I'm not sure how best to approach that, though.

ColinKennedy:
I wasn't able to get `is_other_file` to handle this case because `otherfile` also relied on an API call that requires the file to exist on disk. In Neovim it's possible to completely solve the problem because I could check for existence with [file_id_valid](https://github.com/neovim/neovim/blob/141182d6c6c06ad56413b81a518ba9b777a0cbe0/src/nvim/buffer_defs.h#L385C8-L385C21) but the Vim equivalent [file_buffer->b_dev_valid](https://github.com/vim/vim/blob/82e079df814f7372e9579450730062b205449efa/src/structs.h#L3034-L3038) isn't available outside of UNIX

chrisbra:
thanks all, let me apply this patch here.

ColinKennedy:
@chrisbra I was wondering if you could check out [this comment](https://github.com/vim/vim/pull/14286#issuecomment-2023134377) and weigh in on it. Part of me wonders if it'd be possible to make `b_dev_valid` always available since Neovim was able to achieve that behavior or if there's another, simpler way to query "does this buffer represent something on disk or in memory". Ultimately that is the query I was trying to make but I couldn't find any other reliable method.

If we do find a way, that opens up some opportunities for more flexible `'winfixbuf'` behavior, at least for `:edit`

chrisbra:
Thanks, I have seen your comment, but I just don't know. I guess it's not just Unix, but potentially VMS as well were we would need a way to figure this out.

PR #14287, opened 2024-03-24
Refactor to remove STRLEN() part 1

basilisk0315:
I have added a new structure to store key (int)/value (string) pairs for use in static lookup tables (eg event names, regexp character classes, special keywords, highlighting). This structure embeds the length of the string (determined at compile time) which allows removal of calls to STRLEN() when performing the lookup comparisons.

In addition to factoring out calls to STRLEN() I have added a simple circular FILO caching mechanism to the event_nr2name() lookup functions with the goal to speed it up. Other lookup functions now use a binary search (bsearch()). This meant changing the sequencing of the various lookup compiled-in arrays (tables). This is much simpler than my previous attempt with caching which I also found would be messy (and a little error prone) to use in a general case.

I also addressed a potential infinite loop in the test test_codestyle.vim.


basilisk0315:
Hi Christian,

I have fixed my mysterious failures.
This leaves 2 outstanding: a Mac one which looks to be fairly normal and a codecov one.

How should I proceed with resolving these?
Cheers
John

chrisbra:
Sorry, I haven't had time to dig into this deeper. I'll look into it in more details after the Easter vacation.

PR #14288, opened 2024-03-24
Added the `SessionWritePost` autocommand

### LINKED ISSUE #14242: Add `SessionWritePost` event
The use-case described below may be solvable already but I wasn't sure how to achieve it. Another idea came to me instead.

Suppose you have 2 plugins, both of which extend the default `:mksession` with extra data. They implement this by appending extra data needed by the plugin to the end of the Session.vim, like so

```vim
function plugin_a#make_session()
    let l:file_path = "Session.vim"
    execute "mksession " . l:file_path
    " Add stuff to l:file_path
endfunction

command PluginAMakeSession call plugin_a#make_session()
```

```vim
function plugin_b#make_session()
    let l:file_path = "Session.vim"
    execute "mksession " . l:file_path
    " Add stuff to l:file_path
endfunction

command PluginBMakeSession call plugin_b#make_session()
```

If the user of both plugins wants to have a Session.vim that contains the contents of both plugin a and b, they can't because the plugins overwrite each other's data. The user could modify the plugins to play nice but I was thinking there could be an alternate, better way - expose an autocmd called `SessionWritePost`.

Neovim supports `SessionWritePost`. When triggered, `v:saved_session` (Edit: Update, forgot about `v:this_session`. So no new `v:saved_session` would be required) points to the absolute path on-disk here a session file was written. Then plugin a and b can choose to apply their changes by appending to the created file, like so:

```vim
function plugin_a#append_to_session_path(file_path)
    " Append extra data to a:file_path, below
endfunction

autocmd SessionWritePost * call plugin_a#append_to_session_path(v:saved_session)
```

```vim
function plugin_b#append_to_session_path(file_path)
    " Append extra data to a:file_path, below
endfunction

autocmd SessionWritePost * call plugin_b#append_to_session_path(v:saved_session)
```

With a unified autocmd to trigger from, both plugins can append their relevant data to to the session and then the user can later load the Session.vim file as normal and expect to see Vim's + plugin a's + plugin b's settings.

A possible issue with this idea would be that mksession has a footer at the end of the file e.g. `" vim: set ft=vim :`. If plugin a and b appended to the end, it would move the footer to the middle of the file. Not sure if that footer is important. Though plugin a and b could possibly append to the beginning of Session.vim instead.

n/a

**Additional context**
Originally posted in https://github.com/neovim/neovim/issues/27761

ColinKennedy:
Closes: https://github.com/vim/vim/issues/14242

### DIFF
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -429,7 +429,7 @@ Name			triggered by ~
 |SessionLoadPost|	after loading a session file
 
 |SessionWritePost|	After writing the session file using
-			the :mksession| command.
+			the |:mksession| command.
 
 |MenuPopup|		just before showing the popup menu
 |CompleteChanged|	after Insert mode completion menu changed
--- a/runtime/doc/autocmd.txt
+++ b/runtime/doc/autocmd.txt
@@ -428,6 +428,9 @@ Name			triggered by ~
 
 |SessionLoadPost|	after loading a session file
 
+|SessionWritePost|	After writing the session file using
+			the :mksession| command.
+
 |MenuPopup|		just before showing the popup menu
 |CompleteChanged|	after Insert mode completion menu changed
 |CompleteDonePre|	after Insert mode completion is done, before clearing
@@ -1114,6 +1117,9 @@ SafeStateAgain			Like SafeState but after processing any
 
 							*SessionLoadPost*
 SessionLoadPost			After loading the session file created using
+				the |:mksession| command.
+							*SessionWritePost*
+SessionWritePost		After writing a session file by calling
 				the |:mksession| command.
 							*ShellCmdPost*
 ShellCmdPost			After executing a shell command with |:!cmd|,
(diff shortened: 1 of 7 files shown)


ColinKennedy:
Most of the Windows tests are failing. Seems to be unrelated to this PR though and more of a CI issue.

Screenshot from 2024-03-24 14-20-01

If it is something in this PR, please let me know and I can fix it.

runtime/doc/autocmd.txt
```diff
@@ -428,6 +428,9 @@ Name			triggered by ~
 
 |SessionLoadPost|	after loading a session file
 
+|SessionWritePost|	After writing the session file using
+			the :mksession| command.
```
zeertzjq:
```suggestion
			the |:mksession| command.
```

chrisbra:
thanks. I assume a single `<Session>x.vim` is sufficient?

ColinKennedy:
> thanks. I assume a single `<Session>x.vim` is sufficient?

Yep! It does the job even as a single file. It's a bit unfortunate that Sessionx.vim is called before Session.vim since I would've liked the ability to override certain Vim behaviors or append X or Y behavior after the buffers that Session.vim creates but I don't have a specific use case at the moment. Not important.

The autocmd is important though, so I prioritized it first

chrisbra:
okay, thanks. I moved the tests back to `test_autocmd`

PR #14289, opened 2024-03-25
Cannot use modifiers before :-Ntabmove

zeertzjq:
Problem:  Cannot use modifiers before :-Ntabmove.
Solution: Check backwards from the command instead of checking from the
          start of the command line. Slighyly adjust docs to make them
          more consistent.


chrisbra:
thanks!

PR #14291, opened 2024-03-25
Recalculate scrollbar in `popup_getpos()`

### LINKED ISSUE #14290: `win_execute` should(?) update popup properties (`scrollbar`)
### Steps to reproduce

1. `vim -Nu NONE`
2. Run following script:

```
vim9script

hi PmenuThumb ctermbg=red guibg=red

var winid = popup_create([], {
    maxheight: 3,
    minheight: 3,
    minwidth: 19,
    maxwidth: 19,
    scrollbar: true,
    wrap: 1,
    filter: (id, key) => {
        if key == "\<esc>"
            popup_close(id, -1)
        endif
        return true
    }
})

popup_settext(winid, [
    "hello world",
    "hello world (10)",
    "hello world",
    ])

win_execute(winid, "setl nu")

# workaround
# popup_move(winid, {})

var pos = popup_getpos(winid)

echow "Is there a scrollbar?" pos.scrollbar

```

Result:

Function `popup_getpos()` returns `scrollbar: 0` even though popup window shows there is scrollbar.

image

### Expected behaviour

Function `popup_getpos()` should return `scrollbar: 1` if there is scrollbar.

As a workaround one can use `popup_move(winid, {})` before calling `popup_getpos` to ensure scrollbar value is updated.

The use case is the alignment of 2 popup windows where one window could have a scrollbar and you need to adjust width to accommodate scrollbar width. In some cases `popup_getpos` returns 0 for a scrollbar even though it is there.

image

The issue as I can see it is in `win_execute` that might add `setl number` effectively changing available horizontal space for a text, wrapping some lines.

### Version of Vim

9.1.202

### Environment

debian wsl, bash, simple terminal

### Logs and stack traces

_No response_

habamax:
Use `popup_adjust_position()` in `popup_getpos()` to update scrollbar information before returning it.

Fixes #14290

### DIFF
--- a/src/popupwin.c
+++ b/src/popupwin.c
@@ -3057,6 +3057,7 @@ f_popup_getpos(typval_T *argvars, typval_T *rettv)
     wp = find_popup_win(id);
     if (wp == NULL)
 	return;  // invalid {id}
+    popup_adjust_position(wp);
     top_extra = popup_top_extra(wp);
     left_extra = wp->w_popup_border[3] + wp->w_popup_padding[3];
 
--- a/src/popupwin.c
+++ b/src/popupwin.c
@@ -3046,7 +3046,6 @@ f_popup_getpos(typval_T *argvars, typval_T *rettv)
     win_T	*wp;
     int		top_extra;
     int		left_extra;
-    int		lnum;
 
     if (rettv_dict_alloc(rettv) == FAIL)
 	return;
--- a/src/popupwin.c
+++ b/src/popupwin.c
@@ -3061,8 +3061,6 @@ f_popup_getpos(typval_T *argvars, typval_T *rettv)
     top_extra = popup_top_extra(wp);
     left_extra = wp->w_popup_border[3] + wp->w_popup_padding[3];
 
-    popup_adjust_position(wp);
-
     // we know how much space we need, avoid resizing halfway
     dict = rettv->vval.v_dict;
     hash_lock_size(&dict->dv_hashtab, 11);
@@ -3079,11 +3077,13 @@ f_popup_getpos(typval_T *argvars, typval_T *rettv)
     dict_add_number(dict, "core_width", wp->w_width);
     dict_add_number(dict, "core_height", wp->w_height);
 
-    dict_add_number(dict, "scrollbar", wp->w_has_scrollbar);
     dict_add_number(dict, "firstline", wp->w_topline);
     dict_add_number(dict, "lastline", wp->w_botline - 1);
     dict_add_number(dict, "visible",
 	    win_valid(wp) && (wp->w_popup_flags & POPF_HIDDEN) == 0);
+
+    popup_adjust_position(wp);
+    dict_add_number(dict, "scrollbar", wp->w_has_scrollbar);
 
     hash_unlock(&dict->dv_hashtab);
 }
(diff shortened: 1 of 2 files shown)


habamax:
scrollbar recalc is incorrect.

I think calling `popup_adjust_position` is a better/right one

habamax:
hm, failing tests seems to be unrelated.

Anyway, I guess I am fine with a workaround I have:

- either set popup contents before `win_execute(winid, "setl number")`
- or call `popup_move(winid, {})` before getting `popup_getpos()`

Closing for now.

chrisbra:
Interestingly, the test seems to fail for me, if I run it interactively (even without your patch).

habamax:
> Interestingly, the test seems to fail for me, if I run it interactively (even without your patch).

Do you mean test fails with or without patch if you `:source` it?

chrisbra:
it fails for me, even if I run the test manually, starting from `vim --clean` build from master. I open the test file, enter the commands manually.

habamax:
Not sure I understand, to be frank :)

So this is extract from the test where I add `echo v:errors` and fixed expected vs actual in `assert_equal`:

```
let winid = popup_create([], {'maxheight': 3, 'minheight': 3, 'minwidth': 19, 'maxwidth': 19, 'scrollbar': v:true, 'wrap': 1})
call popup_settext(winid, ["hello world", "hello world (10)", "hello world"])
call win_execute(winid, "setl nu")
let pos = popup_getpos(winid)
call assert_equal(1, pos.scrollbar)
echo v:errors
call popup_clear(winid)
```

Shows after `:source`:
- in master vim: `Expected 1 but got 0`
- in this PR vim: []

chrisbra:
I mean, if I execute `Test_popup_firstline()` manually: 

https://github.com/vim/vim/blob/cb204e688e5c9d56a78b621ef27c35d91860cb09/src/testdir/test_popupwin.vim#L456-L471

(e.g. I execute each line manually) it fails with: `Expected 10 but got 11'`

I can also make the test fail by adding an explicit redraw:
```
diff --git a/src/testdir/test_popupwin.vim b/src/testdir/test_popupwin.vim
index a397f7075..1da738144 100644
--- a/src/testdir/test_popupwin.vim
+++ b/src/testdir/test_popupwin.vim
@@ -467,6 +467,7 @@ func Test_popup_firstline()

   " Normal command changes what is displayed but not "firstline"
   call win_execute(winid, "normal! \<c-y>")
+  redraw
   call assert_equal(11, popup_getoptions(winid).firstline)
   call assert_equal(10, popup_getpos(winid).firstline)
```

```
Found errors in Test_popup_firstline():
command line..script /home/chrisbra/code/vim-src/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_popup_firstline line 26: Expected 10 but got 11
```

Which was the same failure that your patch produced? https://github.com/vim/vim/pull/14291#issuecomment-2017410647

habamax:
Ah, yes indeed.

That testcase, I don't understand

habamax:
I would use (and first patch did it) `popup_adjust_position(wp);` just after:

```
    wp = find_popup_win(id);
    if (wp == NULL)
	return;  // invalid {id}

    popup_adjust_position(wp);
```

But it breaks that testcase.

chrisbra:
yeah, `popup_adjust_position` causes a redraw, that's why it breaks the test. We may need to figure out, why the firstline property changes after a redraw.

PR #14297, opened 2024-03-25
runtime(java): Update java[CR]_JavaLang type lists

zzzyxwvut:
- Add to the list of `java.lang` runtime exceptions two new  
  types: `MatchException` (Pattern Matching for `switch`) and  
 `WrongThreadException` (Virtual Threads).
- "Demote" `Compiler` (removed in JDK 21) from the list of  
  `java.lang` class types to a new list `javaLangDeprecated`.

References:
https://bugs.openjdk.org/browse/JDK-8205129
https://bugs.openjdk.org/browse/JDK-8282274
https://bugs.openjdk.org/browse/JDK-8284161


chrisbra:
Thanks!

PR #14300, opened 2024-03-26
Backspace with 'smarttab' doesn't respect 'revins'

zeertzjq:
Problem:  Backspace with 'smarttab' doesn't respect 'revins'.
Solution: Delete from the end of whitespace with 'smarttab'.  Check
          inindent() after calling inc_cursor().  Also remove several
          unnecessary multiplications, as "col / ts * ts" is the same
          as "col - col % ts".

It's not entirely clear what the expected behavior of 'smarttab' should
be when 'revins' is set, but this at least fixes the obviously incorrect
behavior at the start of a line:
- Before 9.1.0204, it inserted spaces if the line above is long enough.
- After 9.1.0204, it didn't do anything.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -4023,14 +4023,14 @@ ins_bs(
 
     if (stop_arrow() == FAIL)
 	return FALSE;
-    in_indent = inindent(0);
-    if (in_indent)
-	can_cindent = FALSE;
     end_comment_pending = NUL;	// After BS, don't auto-end comment
 #ifdef FEAT_RIGHTLEFT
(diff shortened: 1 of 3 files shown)


chrisbra:
So does it now behave like pre v9.1.204?

zeertzjq:
No, the behavior before 9.1.0204 made completely no sense at the start of line.

zeertzjq:
I'll later move the unrelated changes to a separate PR, so that this PR only contains changes for 'revins'.

zeertzjq:
I'm no longer interested in this.

PR #14305, opened 2024-03-26
Remove unused display_text_first boolean variable

dylanahsmith:
## Problem

The variable display_text_first in the win_line function doesn't actually have any effect, but its presence makes it harder to understand the code.

The only place it is used, uses its initial constant value, then the following conditionally values set to the variable are unused.

For context, the variable was initially introduced in commit 9d9a20ee8799bafe9caac616fef11b7a26db6a8d, where it had an effect, but then commit 234c3fab28c14846b962c90097496b27ee1b4df8 changed the code to the current state where it doesn't anything.  Those changes are both covered by regression tests.

## Solution

Unless someone sees why the variable is still needed, I think we should just remove it.  If we end up needing to add a variable like that back, it is better to do once we have a regression test showing the need for it.

### DIFF
--- a/src/drawline.c
+++ b/src/drawline.c
@@ -2072,8 +2072,6 @@ win_line(
 		    // not on the next char yet, don't start another prop
 		    --bcol;
 # endif
-		int display_text_first = FALSE;
-
 		// Add any text property that starts in this column.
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, makes sense

PR #14306, opened 2024-03-26
runtime(debcontrol): add Static-Built-Using field

guilherme-puida:
From `man 5 deb-control`:

```
 Static-Built-Using: package-list
           This dependency field lists  extra  source  packages  that
           were  used  during  the  build of this binary package, for
           static building  purposes  (for  example  linking  against
           static  libraries,  builds  for  source-centered languages
           such as Go or Rust, usage of header-only C/C++  libraries,
           injecting  data blobs into code, etc.).  This is useful to
           track whether this package might need to be  rebuilt  when
           source packages listed here have been updated, for example
           due  to  security  updates.   This  field must be a comma-
           separated list of source package  names  with  strict  ‘=’
           version relationships enclosed within parenthesis.

           Supported since dpkg 1.21.3.
```

### DIFF
--- a/runtime/syntax/debcontrol.vim
+++ b/runtime/syntax/debcontrol.vim
@@ -3,7 +3,7 @@
 " Maintainer:  Debian Vim Maintainers
 " Former Maintainers: Gerfried Fuchs <alfie@ist.org>
 "                     Wichert Akkerman <wakkerma@debian.org>
-" Last Change: 2023 Dec 22
+" Last Change: 2024 Mar 26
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks! Did you check with the Debian Team? ping @jamessan

jamessan (approved):

guilherme-puida:
My bad, didn't realize that the file is maintained by the Debian Team.

chrisbra:
thanks, I'll include it then.

PR #14307, opened 2024-03-26
Fix undisplayed lines below line truncated just before virtual text

### LINKED ISSUE #12213: Virtual text to the right of the line that fits exactly on screen causes that line and all lines below it not to be displayed
### Steps to reproduce

When vim-ale puts a bit of virtual text after the end of the line, and the length of the actual text exactly matches the window width, Vim gets confused and hides the entire line (and usually the lines below it):

paveikslas

This requires `:set nowrap` and is a bit easier to see with `:set display=`.

I've managed to reproduce this or something very like this (I don't see the `@` characters, but the text itself disappears):

1. Create a `repro.vim` with
  
```
40vsplit
0put ='here is some text more text even *more*:'
put ='  more text!'
put ='  more text!'
put ='  more text!'
put ='  more text!'
set nomodified " make :qa work for convenience
set nowrap
set display=
set list listchars=extends:>
call prop_type_add("bug", {"highlight": "Error"})
call prop_add(1, 0, {"type": "bug", "text": "# virtual text that is longer than the actual text", "text_padding_left": 1})
```

2. Run `vim --clean -S repro.vim`

Observe:

paveikslas

### Expected behaviour

This is what things look like when the window is one character wider than the text:

paveikslas

and in my synthetic example:

paveikslas

### Version of Vim

9.0.1392

### Environment

OS: Ubuntu 22.10
Terminal: gnome-terminal 3.46.2 using VTE 0.70.0
$TERM: xterm-256color
Shell: bash

### Logs and stack traces

_No response_

dylanahsmith:
Fixes #12213

## Problem

The if branch to set `text_prop_follows` was both checking if it was at the end of the buffer text line or if it was at the end of the screen line, but the former being true skipped a guard condition in the latter to only consider "below" virtual text to follow. `text_prop_follows` being improperly set caused it to skip a conditional block to break at the end as well as one to move `ptr` to the end of the text line, while repeated for each following line of the window.

As part of fixing this problem, I noticed that it was only checking the next virtual text property to see if is a "below" virtual text, even though it may follow "after"/"right" virtual text properties.

## Solution

Since I found the virtual "after"/"right"/"below" virtual text properties aren't ordered in the buffer, I figured a loop was going to be needed to find if there is a "below" virtual text property when at the end of the screen line.  So I extracted the text property filtering part of the condition into the added loop, so it will apply regardless of whether it is at the end of the text line.

As part of regression testing this, I found that "after"/"right" virtual text properties caused the while loop to make text properties active to break before reaching the "below" virtual text property.  This just involved moving part of the conditional from breaking the loop into the loop to only use for skipping properties. This caused a lot of duplication, which made the code more fragile and hard to read, so I did some refactoring, but did that in a separate commit so it can be reviewed as a separate atomic change.


chrisbra:
thanks!

PR #14308, opened 2024-03-26
Unnecessary multiplications in backspace code

zeertzjq:
Problem:  Unnecessary multiplications in backspace code, as
          "col / ts * ts" is the same as "col - col % ts".
Solution: Change "col / ts * ts" to "col - col % ts".  Adjust the loop
          and the comments ins_bs() to be easier to understand.  Update
          tests to reset 'smarttab' properly.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -4210,12 +4210,11 @@ ins_bs(
 				&& (!*inserted_space_p
 				    || arrow_used))))))
 	{
-	    int		ts;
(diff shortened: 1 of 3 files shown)


chrisbra:
That seems like unneccessary optimizations at the risk of breaking something. Does this really gain us something?

chrisbra:
okay thanks!

PR #14309, opened 2024-03-27
regexp: verymagic cannot match before/after a mark

julio-b:
Fix regexp parser for `\v%>'m` and `\v%<'m`
Currently `\v%'m` works fine, but it is unable to match before or after the position of mark `m`. The reason is that in line 1653, when the condition is true, `c` gets a negative value returned by `getchr()`

### DIFF
--- a/src/testdir/test_regexp_latin.vim
+++ b/src/testdir/test_regexp_latin.vim
@@ -890,6 +890,7 @@ func Test_matching_marks()
   new
   set regexpengine=1
   call Regex_Mark()
+  call Regex_Mark_Verymagic()
   set regexpengine=2
   call Regex_Mark()
   call Regex_Mark_Verymagic()
--- a/src/testdir/test_regexp_latin.vim
+++ b/src/testdir/test_regexp_latin.vim
@@ -874,12 +874,25 @@ func Regex_Mark()
   %d
 endfunc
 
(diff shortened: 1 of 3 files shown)


src/testdir/test_regexp_latin.vim
```diff
@@ -874,12 +874,25 @@ func Regex_Mark()
...
+  exe "normal jfSmsj0fEme:.-4,.+6s/\\v.%>'s\\_.*%<'e../again/\<CR>"
+  call assert_equal(['', '', '', 'Marks:', 'asdfhereasdf', 'asdfagainasdf',
+        \ '', '', '', '', '', ''], getline(1, '$'))
+  %d
+endfunc
+
 func Test_matching_marks()
   new
   set regexpengine=1
   call Regex_Mark()
```
chrisbra, line 892: the new test should also be executed for `regexpengine=1`

chrisbra:
thanks. Does it work correctly with regexpengine=1?

julio-b:
Oh you are right! It turns out that old engine has the same issue, i have pushed a fixup. Now it's working correctly for both engines.

src/testdir/test_regexp_latin.vim
julio-b: fixed by latest commit.

PR #14314, opened 2024-03-27
Error on exit with EXITFREE and 'winfixbuf'

zeertzjq:
Problem:  Error on exit with EXITFREE and 'winfixbuf'.
Solution: Handle DT_FREE before checking for 'winfixbuf'.


chrisbra:
thanks!

PR #14316, opened 2024-03-27
Half-page scrolling does not support smooth-scrolling

luukvbaal:
Problem:  Page-wise scrolling with Ctrl-D/Ctrl-U implements
          it's own logic to change the topline and cursor.
          More logic than necessary for scrolling with Ctrl-F/Ctrl-B
          was removed in patch 9.1.0211.
Solution: Re-use the logic from Ctrl-E/Ctrl-Y/Ctrl-F/Ctrl-B while
          staying backward compatible as much as possible.
          Restore some of the logic that determined how many lines will
          be scrolled.


luukvbaal:
I took some more time to look into the behavioral differences after running into more failing tests when porting the previous patch for Neovim. This gets us closer to the previous behavior.

luukvbaal:
Half page scrolling now uses the same logic to scroll X number of lines as well.

chrisbra:
thanks!

luukvbaal:
`git blame` actually points to https://github.com/vim/vim/commit/e6d8b4662ddf9356da53f56e363b67b524fd8825 breaking that test🤷🏻

luukvbaal:
This patch did change the cursor position in that test but it has been broken since patch 9.1.0061.

chrisbra:
Hm, seems unintentionally not run. But isn't the only difference the VISUAL highlighting now?

chrisbra:
I have just created https://github.com/vim/vim/pull/14322 to make matchparen test run in CI. I didn't see a change in the cursor position however.

luukvbaal:
After this patch the cursor ends up on line 8 instead of line 11 for me. The updated display dump from #14322 also still fails.

PR #14317, opened 2024-03-27
Fix cursor may move too many lines over "right" & "below" virt text

dylanahsmith:
Based on #14307, see https://github.com/dylanahsmith/vim/compare/fix-12213...dylanahsmith:vim:fix-cursor-move-over-below-right to exclude that branch's commits from the comparison

## Problem

The problem can be demonstrated by saving the following vim script as repro.vim, then loading it with `vim --clean -S repro.vim`

```vim
vim9script
set nowrap

setline(1, [
  'here is text long enough to fill the row',
  'second line',
])

prop_type_add("test", {"highlight": "Error"})
prop_add(1, 0, {type: 'test', text: 'right text', text_align: 'right'})
prop_add(1, 0, {type: 'test', text: 'below text', text_align: 'below'})
normal G$
```

which without this fix results in the cursor appearing on the screen line **below** the last character in `second line`.

My PR #14307 actually work around this issue that I originally uncovered from an earlier version of its regression test, which added the "right" virtual text property before the "below" property.  As noted in that PR description, those text properties aren't sorted, but instead got stored in the reverse order of how they were stored.

When calculating the line height in `prop_count_above_below`, `next_right_goes_below = TRUE` was for some reason set when encountering the "below" text property, leading to it incorrectly counting an "extra" line for the following "right" property even though it was the first "right" property.

## Solution

Only "right" virtual text properties should always shift other "right" properties on the same line to the next line, so I removed the unnecessary `next_right_goes_below = TRUE;` statements and simplified the code.  I then updated the regression test to cover this case, such that the test fails without this fix.


chrisbra:
thanks. If I see this correctly, after having merged #14307 we just need the latest commit here: 9adb353

I'll include that now as v9.1.224. Please check if there is anything missing from that.

PR #14318, opened 2024-03-27
runtime(vim): Update base-syntax, match empty blob and :abclear modifiers

dkearns:
- Match empty blob literals.
- Match modifier arguments to :abclear commands.


chrisbra:
thanks!

PR #14319, opened 2024-03-27
Fix Haskell TODO syntax

Aster89:
Over a year, and I forgot about this.

Anyway, the point is that  #8055 is not sufficient to allow TODO to have the correct syntax coloring.

This commit should fix it.


chrisbra:
thanks!

PR #14320, opened 2024-03-28
Duplicate condition in win_lbr_chartabsize()

zeertzjq:
Problem:  Duplicate condition in win_lbr_chartabsize().
Solution: Remove the duplicate condition, as it's already checked above.


chrisbra:
makes sense, thanks!

PR #14322, opened 2024-03-28
Problem: test_matchparen not run in CI

chrisbra:
Problem:  test_matchparen not run in CI
Solution: add test_matchparen.res to NEW_TESTS_RES in Makefile,
          update changed highlighting dump from v9.1.61


luukvbaal:
OK so we now have different cursor position based on compiled features. That doesn't seem right for this test, I'll try to look into it.

chrisbra:
yes, let's just update the test with the changed cursor positions.

PR #14323, opened 2024-03-28
Two unrelated things are tested by a single test

zeertzjq:
Problem:  Two unrelated things are tested by a single test.
Solution: Split it into two, restoring the old Test_brace_single_line().
          Add missing cleanup to some tests.


zeertzjq:
I also don't think 'smoothscroll' should be considered "partly implemented" now, as it now cover all applicable scrolling commands listed in `:h scroll-down` and `:h scroll-up`. It isn't applicable to `z+`  and `z^` as they move the cursor to the first non-blank of a line.

luukvbaal:
It still works poorly with `gj` and `gk`. I think there is an open issue/todo.txt item for that.

chrisbra:
Thanks, makes sense.

PR #14324, opened 2024-03-28
Improve doc for empty(), len(), and string()

mityu:
In the documentation for the use of empty(), len(), and string() for objects, there're links to |string()|, |len()|, and |string()| but these are misleading.  I think they should be |object-string()|, |object-len()| and |object-string()|.

This PR suggests a concrate changes of this.


chrisbra:
thanks , makes sense

PR #14327, opened 2024-03-29
Add MsgArea highlight support

Shougo:
neovim implements `MsgArea` highlight support.

So I have implemented it in Vim.

It is useful to implement input command line plugin.

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -5496,6 +5496,8 @@ MessageWindow	Messages popup window used by `:echowindow`.  If not defined
 		|hl-WarningMsg| is used.
 							*hl-ModeMsg*
 ModeMsg		'showmode' message (e.g., "-- INSERT --").
+							*hl-MsgArea*
+MsgArea		Area for messages and cmdline.
 							*hl-MoreMsg*
 MoreMsg		|more-prompt|
 							*hl-NonText*
--- a/src/message.c
+++ b/src/message.c
@@ -1637,6 +1637,9 @@ msg_outtrans_len_attr(char_u *msgstr, int len, int attr)
     // Only quit when got_int was set in here.
     got_int = FALSE;
 
+    if (attr == 0)
+	attr = HL_ATTR(HLF_MSG);
+
     // if MSG_HIST flag set, add message to history
     if (attr & MSG_HIST)
     {
@@ -2230,6 +2233,9 @@ msg_puts_attr_len(char *str, int maxlen, int attr)
     if (msg_silent != 0)
 	return;
 
+    if (attr == 0)
+	attr = HL_ATTR(HLF_MSG);
+
     // if MSG_HIST flag set, add message to history
     if ((attr & MSG_HIST) && maxlen < 0)
     {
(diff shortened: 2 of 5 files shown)


Shougo:
Note:  I want to add the support for `Cursor` highlight.  But it is hard.

bfrg:
> It is useful to implement input command line plugin.

How is this useful for input command-line plugins?

In the screenshot it looks like you're trying to hide the command-line and instead show the entered text in the popup window. Wouldn't it be more convenient to add a new function like a `popup_prompt()` (or `popup_input()`) that asks the user for input, similar to the already existing `input()` function?

Shougo:
> How is this useful for input command-line plugins?

Some plugin like `fall.vim` implements command line interface.
But it cannot hide original command line.  In neovim, it can hide the command line by `MsgArea`.

https://github.com/vim-fall/fall.vim



> In the screenshot it looks like you're trying to hide the command-line and instead show the entered text in the popup window. Wouldn't it be more convenient to add a new function like popup_prompt() [or popup_input()] that asks the user for input, similar to the already existing input() function?

The new function may be useful.  But it will be complex implementation.

chrisbra:
Hm, it seems `MsgArea` is just a fancy way to highlight the command line? But if you want to use a command line plugin, it seems you rather want something different? So how does it relate?

Shougo:
> Hm, it seems MsgArea is just a fancy way to highlight the command line?

Yes.  It is just the highlight of command line.
However, this can be applied to implementing your own command line window to hide original command line area.

> But if you want to use a command line plugin, it seems you rather want something different? So how does it relate?

The approach advantage is compatibility. 
The command line plugins simply mirror the contents of the command line.

chrisbra:
Hm, so what would your command line plugin from your initial post would look like if we do not try to style the command line? E.g. what does it look like currently?

Shougo:
This is the current Vim's behavior.

The below command line message cannot be hidden.

Shane-XB-Qian:
> Note: The command line cursor cannot be disabled in both.

`:sleep! 6` can hide the cursor, but just 'another' cursor donot know where it from.

Shougo:
It works but I cannot operate Vim.  Because it waits until sleep is finished...

Shane-XB-Qian:
> It works but I cannot operate Vim. Because it waits until sleep is finished...
> 
> I have research the code. `cursor_on` and `cursor_off` API are exists in Vim source code. I want to use it from Vim script...

yea, but i also had a concern, but if that came true, maybe should be sure that cursor on/off was under explicitly and safe control, otherwise it may be used to steal cursor/focus to do some fake things.

chrisbra:
so basically you are just doing `:hi MsgArea ctermfg=black ctermbg=black guifg=black guifg=black` to make it "disappear"? I am not sure if it is a good idea to allow this, but on the other hand, the current solution also does not look too nice.

Shougo:
> so basically you are just doing :hi MsgArea ctermfg=black ctermbg=black guifg=black guifg=black to make it "disappear"?

Yes.  The current code is:

```vim
  " NOTE: Disable cmdline area highlight
  const hl_normal = has('nvim') ?
        \ nvim_get_hl(0, #{ name: 'Normal'}) : hlget('Normal')
  const hl_msg = has('nvim') ?
        \ nvim_get_hl(0, #{ name: 'MsgArea'}) : hlget('MsgArea')
  const hl_cursor = has('nvim') ?
        \ nvim_get_hl(0, #{ name: 'Cursor'}) : hlget('Cursor')

  if has('nvim')
    call nvim_set_hl(0, 'MsgArea', #{ fg: hl_normal.bg, bg: hl_normal.bg })
    call nvim_set_hl(0, 'Cursor', #{ fg: hl_normal.bg, bg: hl_normal.bg })
  else
    call hlset([
          \   #{
          \     name: 'MsgArea',
          \     guifg: hl_normal[0].guibg,
          \     guibg: hl_normal[0].guibg,
          \   },
          \ ])
  endif
```

Shougo:
> Note: I want to add the support for Cursor highlight. But it is hard.

I have added `cursor_on()` and `cursor_off()` for it.

chrisbra:
There is already Cursor Highlight, which seems to work fine in the GUI. For the terminal, I guess you can use some kind terminal escape sequences to hide the cursor using e.g. `echoraw()`. But as mentioned, I would strongly recommend not to completely hide the cursor, instead make it a small horizontal bar if this is possible.

chrisbra:
also can you please add some tests here?

Shougo:
> But as mentioned, I would strongly recommend not to completely hide the cursor, instead make it a small horizontal bar if this is possible.

I have checked ... but Vim does not support the feature.
I use `t_SI`, `t_EI` but it is not for the cursor change.

PR #14336, opened 2024-03-29
runtime:compilers: ensure compiler! sets global options

Konfekt:
Previously some options were only set locally by
&l:makeprg/errorformat

This suffices for :compiler (without a trailing bang) but falls short for :compiler! that sets &g:makeprg/errorformat as well

### DIFF
--- a/runtime/compiler/bdf.vim
+++ b/runtime/compiler/bdf.vim
@@ -1,7 +1,7 @@
 " Vim compiler file
 " Compiler:             BDF to PCF Conversion
 " Previous Maintainer:  Nikolai Weibull <now@bitwi.se>
-" Latest Revision:      2006-04-19
+" Latest Revision:      2024 Mar 29
 
 if exists("current_compiler")
   finish
@@ -11,9 +11,12 @@ let current_compiler = "bdf"
 let s:cpo_save = &cpo
 set cpo-=C
 
-setlocal makeprg=bdftopcf\ $*
+if exists(":CompilerSet") != 2 # Older Vim always used :setlocal
+  command -nargs=* CompilerSet setlocal <args>
+endif
 
-setlocal errorformat=%ABDF\ %trror\ on\ line\ %l:\ %m,
+CompilerSet makeprg=bdftopcf\ $*
+CompilerSet errorformat=%ABDF\ %trror\ on\ line\ %l:\ %m,
       \%-Z%p^,
       \%Cbdftopcf:\ bdf\ input\\,\ %f\\,\ corrupt,
       \%-G%.%#
--- a/runtime/compiler/bdf.vim
+++ b/runtime/compiler/bdf.vim
@@ -1,7 +1,8 @@
 " Vim compiler file
 " Compiler:             BDF to PCF Conversion
 " Previous Maintainer:  Nikolai Weibull <now@bitwi.se>
-" Last Change:      2024 Mar 29 by Enno Nagel
+" Contributors:         Enno Nagel
+" Last Change:          2024 Mar 29
 
 if exists("current_compiler")
   finish
--- a/runtime/compiler/bdf.vim
+++ b/runtime/compiler/bdf.vim
@@ -12,7 +12,7 @@ let current_compiler = "bdf"
 let s:cpo_save = &cpo
 set cpo-=C
 
-if exists(":CompilerSet") != 2 # Older Vim always used :setlocal
+if exists(":CompilerSet") != 2 " Older Vim always used :setlocal
   command -nargs=* CompilerSet setlocal <args>
 endif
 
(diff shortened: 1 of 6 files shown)


runtime/compiler/context.vim
```diff
@@ -20,6 +20,7 @@ g:current_compiler = 'context'
 if get(b:, 'context_ignore_makefile', get(g:, 'context_ignore_makefile', 0)) ||
   (!filereadable('Makefile') && !filereadable('makefile'))
   &l:makeprg =  join(context.ConTeXtCmd(shellescape(expand('%:p:t'))), ' ')
+  execute 'CompilerSet makeprg='. escape(&l:makeprg, ' ')
```
dkearns: This file is Vim9-script so it has stricter syntax.  Can you please add whitespace before the concat operator and use `..`.

I think it would be better just to use a local variable `makeprg` for all of these rather than set the local option value just to read it again and call `CompilerSet`.

dkearns (changes requested):
Thanks @Konfekt.

Could you take this opportunity to change the "Latest Revision: " headers to "Last Change: " and also add a "by Enno Nagel" after the date or, if you're shy, "by The Vim Project"?

runtime/compiler/context.vim
Konfekt: Thank you @dkearns, I applied your suggestions.

Konfekt:
Since upstream https://github.com/PProvost/vim-ps1/blob/master/compiler/powershell.vim might also have other improvements, it could still be preferred to use their latest version instead.

Konfekt:
Just noting for reference that the maintainers of tex.vim and modelsim_vcom.vim could not be reached by email

runtime/compiler/context.vim
```diff
@@ -3,7 +3,7 @@ vim9script
 # Language:           ConTeXt typesetting engine
 # Maintainer:         Nicola Vitacolonna <nvitacolonna@gmail.com>
 # Former Maintainers: Nikolai Weibull <now@bitwi.se>
-# Latest Revision:    2023 Dec 26
+# Last Change:        2024 Mar 29 by Enno Nagel
```
lifepillar: I'd suggest this instead:

```
Contributors: Enno Nagel
Last Change: 2024 Mar 29
```

So that the contributor's name won't be lost at the next update.

Konfekt: Well, Peter Collingbourne would already be victim to this pull request

Konfekt: .. or, less drama, rather the written memory of his contribution in the compiler file

lifepillar: I remember Bram wanting former maintainers to remain in the files. I think it's fair to leave all the names that are already there.

Konfekt: Yes, seems so sensible I applied your suggestions in the latest iteration

runtime/compiler/bdf.vim
```diff
@@ -11,9 +12,12 @@ let current_compiler = "bdf"
 let s:cpo_save = &cpo
 set cpo-=C
 
-setlocal makeprg=bdftopcf\ $*
+if exists(":CompilerSet") != 2 # Older Vim always used :setlocal
```
dkearns: The comment char needs to be `"` not `#`.

runtime/compiler/context.vim
```diff
@@ -3,7 +3,8 @@ vim9script
 # Language:           ConTeXt typesetting engine
 # Maintainer:         Nicola Vitacolonna <nvitacolonna@gmail.com>
 # Former Maintainers: Nikolai Weibull <now@bitwi.se>
-# Latest Revision:    2023 Dec 26
+" Contributors:       Enno Nagel
```
dkearns: Comment char needs to be `#`.

runtime/compiler/mcs.vim
```diff
@@ -12,7 +12,12 @@ let current_compiler = "mcs"
 let s:cpo_save = &cpo
 set cpo-=C
 
-setlocal errorformat=
+if exists(":CompilerSet") != 2 # Older Vim always used :setlocal
```
dkearns: Comment char needs to be `"`

dkearns (changes requested):

runtime/compiler/mcs.vim
Konfekt: Fixed. Thank you very much, and please pardon my hastiness.

dkearns (approved):
Thanks @Konfekt

chrisbra:
thanks, I'll merge it. It looks like the `:CompilerSet` command was introduced with at least Vim 7, which was released around 2004. So not sure we need to do the `if exists(":CompilerSet") != 2 # Older Vim always used :setlocal` - dance anymore.

PR #14344, opened 2024-03-30
Filetype may be undetected when SwapExists sets ft in other buf

zeertzjq:
Problem:  Filetype may be undetected when a SwapExists autocommand sets
          filetype in another buffer.
Solution: Make filetype detection state buffer-specific.  Also fix a
          similar problem for 'modified'.


chrisbra:
thanks!

PR #14348, opened 2024-03-30
runtime(vim): Distinguish Vim9 builtin object methods from namesake builtin functions

zzzyxwvut:
Currently, the overriding object method definitions are  
matched as `vimFunctionError` (`:help builtin-object-methods`,  
`v9.1.0148`).

For example:
```
vim9script

class Test
	def string(): string
		return "Test"
	enddef
endclass

echo string(Test.new()) == Test.new().string()
```

Instead, let's introduce a new syntax group `vimMethodName`  
and make these methods its members.  In order to emphasise  
the link between the overriding methods and the overridden  
functions for highlighting, `vimMethodName` is linked by  
default to `vimFuncName`.


zzzyxwvut:
@dkearns, @h-east, please review.

runtime/syntax/generator/vim.vim.base
```diff
@@ -259,6 +259,7 @@ syn match	vimFuncSID	contained	"\c<sid>"
 syn match	vimFuncSID	contained	"\<[sg]:"
 syn keyword	vimFuncKey	contained	fu[nction]
 syn keyword	vimDefKey	contained	def
+syn match	vimMethodName	contained	"\<\%(empty\|len\|string\)\>"
```
dkearns: Is there a reason not to use syn-keyword rather than syn-match here?

dkearns (changes requested):
Thanks @zzzyxwvut this looks good to me.

I have a number of Vim9 fixes in the pipeline so it might be worth pinging me before working too hard on anything just to check it's not already incoming.  I'm just trying to save on duplicated work and am certainly not trying to discourage you.

runtime/syntax/generator/vim.vim.base
zzzyxwvut: Ugh!  That is an overlooked vestige from another iteration:
`"\<def\s\+\zs\%(empty\|len\|string\)\>"` (before it was  
included into `vimUserFunc` as well).

zzzyxwvut:
Understood.  Next time around for the problem at hand, I'll  
look for the old issue to comment on, otherwise file a new  
one.  :)

I have applied the suggested changes and pushed them.

chrisbra:
thanks!

PR #14349, opened 2024-03-30
runtime(doc): Fill in a few details regarding :enums

zzzyxwvut:
- Mention the support of `eval()` for enumeration values.
- Clarify the extent of immutability for enumeration values.
- Specify the requirements for class methods to meet for  
  class variable initialisation and their use in nested  
  functions and lambda expressions.
- Remove a duplicate sentence that describes how to access  
  parent class methods in derivative classes (see another  
"copy" two paragraphs below).

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -1,4 +1,4 @@
-*vim9class.txt*	For Vim version 9.1.  Last change: 2024 Mar 28
+*vim9class.txt*	For Vim version 9.1.  Last change: 2024 Apr 04
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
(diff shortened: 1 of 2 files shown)


runtime/doc/vim9class.txt
```diff
@@ -972,8 +972,8 @@ The following example shows an enum with object variables and methods: >
     echo Planet.Earth.has_rings
 <
 						*E1421* *E1423* *E1424* *E1425*
-Enums and their values are immutable. They cannot be modified after
-declaration and cannot be utilized as numerical or string types.
+Enums and their values are immutable; however, mutable components are allowed
+for enum values.  They cannot be utilized as numerical or string types.
```
chrisbra: I don't understand that sentence. First we are saying *... their values are immutable*  and just a few words further we are saying that *mutable components are allowed for enum values*? What does that mean and what exactly mutable for enum values?

zzzyxwvut: Sigh... I wish I could express it better.

Well, given these types:

```
vim9script

class Class
    var text: string = "Class"
endclass

enum Enum 
    INSTANCE

    var value: number = 0

    def new()
    enddef

    def SetValue(value: number)
	this.value = value
    enddef
endenum
```

> Enums and their values are immutable

Neither enumerations are extendable by a class (and so no  
new state can be acquired through class inheritance),

```
# E1416
enum Enum extends Class
    INSTANCE
endenum
```

nor any _new_ instances beyond those already declared can be  
created.

```
# E1325
Enum.new()
```

But the state of any enumeration value can be changed when  
the said state is mutable.

```
Enum.INSTANCE.SetValue(5)
echo Enum.INSTANCE
```

zzzyxwvut: I've rewritten vague parts.  Hopefully, the changed text is
more clear now.

yegappan (approved):

chrisbra:
Thanks, looks good

PR #14355, opened 2024-03-31
filetype: intel hex files are not recognized

Freed-Wu:
Problem: sometimes intel hex files are not recognized
Solution: Add a pattern for them
Refer: https://en.wikipedia.org/wiki/Intel_HEX


chrisbra:
thanks!

PR #14356, opened 2024-03-31
filetype: supertux files are not recognized

Freed-Wu:
Problem: Supertux use lisp to store hotkeys in config and game stage information
Solution: Add a pattern for supertux files
Refer: https://github.com/SuperTux/supertux/wiki/S-Expression


chrisbra:
thanks!

PR #14357, opened 2024-03-31
filetype: texlua files are not recognized

Freed-Wu:
Problem: texlua files are not recognized.
Solution: Add a pattern for texlua files.
Refer: https://github.com/TeX-Live/texdoc/tree/master/script


chrisbra:
thanks

PR #14359, opened 2024-03-31
filetype: jupyterlab config and sublime config are not recognized

Freed-Wu:
Problem: jupyterlab config and sublime config are not recognized
Solution: Add patterns for jupyterlab config and sublime config


chrisbra:
thanks

PR #14360, opened 2024-03-31
filetype: gnuplot history files are not recognised

Freed-Wu:
Problem: Gnuplot history files are not recognised.
Solution: Add a pattern for gnuplot history files.


chrisbra:
thanks!

PR #14361, opened 2024-03-31
filetype: some python tools config files are not recognized

Freed-Wu:
Problem: Some python tools config files are not recognized.
Solution: Add a pattern for some python tools config files.


chrisbra:
thanks!

PR #14362, opened 2024-03-31
filetype: mysql history files cannot be recognized

Freed-Wu:
Problem: Mysql history files cannot be recognized.
Solution: Add a pattern for mysql history files.


chrisbra:
thanks

PR #14363, opened 2024-03-31
filetype: octave history files are not recognized

Freed-Wu:
Problem: Octave history files are not recognized.
Solutiion: Add a pattern for octave history files.


chrisbra:
thanks!

PR #14364, opened 2024-03-31
filetype: nerdtree history files cannot be recognized

Freed-Wu:
Problem: Nerdtree history files cannot be recognized.
Solution: Add a pattern for nerdtree history files.
Refer: :help .netrwhist


chrisbra:
thanks.

PR #14365, opened 2024-03-31
filetype: bash history files are not recognized

Freed-Wu:
Problem: Bash history files are not recognized
Solution: Add a pattern for bash history files


chrisbra:
thanks!

PR #14366, opened 2024-03-31
filetype: zsh theme and history files are not recognized

Freed-Wu:
Problem: Zsh theme and history files are not recognized
Solution: Add a pattern for Zsh theme and history files


chrisbra:
thanks!

PR #14367, opened 2024-03-31
filetype: fontconfig config files are not recognized

Freed-Wu:
Problem: Fontconfig config files are not recognized.
Solution: Add a pattern for fontconfig config files.


chrisbra:
thanks!

PR #14368, opened 2024-03-31
filetype: bundle config files are not recognized

Freed-Wu:
Problem: Bundle config files are not recognized.
Solution: Add a pattern for bundle config files.


chrisbra:
thanks!

PR #14369, opened 2024-03-31
filetype: yarn lock files are not recognized

Freed-Wu:
Problem: Yarn lock files are not recognized.
Solution: Add a pattern for yarn lock files


chrisbra:
thanks!

PR #14370, opened 2024-03-31
filetype: rock_manifest and config.ld files cannot be recognized

Freed-Wu:
Problem: Rock_manifest and config.ld files cannot be recognized.
Solution: Add a pattern for rock_manifest and config.ld files


chrisbra:
thanks!

PR #14371, opened 2024-03-31
filetype: ldscripts cannot be recognized

Freed-Wu:
Problem: Ldscripts cannot be recognized
Solution: Add a pattern for ldscripts


chrisbra:
thanks.

PR #14373, opened 2024-03-31
Filetype test fails

zeertzjq:
Problem:  Filetype test fails.
Solution: Move detection by name before detection by extension.
          Improve TextChanged test and remove wrong test.


chrisbra:
thanks!

PR #14377, opened 2024-04-01
Improve scrolling behavior and backward compatibility

### LINKED ISSUE #14338: Documentation of CTRL-D/CTRL-U is no longer accurate
@chrisbra after patch 9.1.0215, the following from `:h CTRL-D` is no longer accurate:

    The cursor is moved the same number of lines down in the file

Also, CTRL-D was the only scroll key that would stop at the end of the buffer (https://github.com/neovim/neovim/pull/28100#discussion_r1544476467) instead of scrolling to reveal EOB lines.

We can either:
1. Update the documentation.
2. Try to restore the previous cursor/EOB behavior only for CTRL-D/U.
3. Try to apply the previous cursor/EOB behavior from CTRL-D/U to all page scroll keys if that behavior is considered desirable.

Personally I tend toward 1. or 3., I don't see why these keys should behave differently (other than backward compatibility).

luukvbaal:
Problem:  Support for 'smoothscroll' in (half-)page scrolling
          broke backward compatibility and can be made to work better.
Solution: Restore the previous cursor and end-of-buffer behavior for
          half-page scrolling and make 'smoothscroll' support work
          better in general.

Fix #14338


src/testdir/test_scroll_opt.vim
```diff
@@ -1017,28 +1017,71 @@ func Test_smoothscroll_page()
...
+  " Moving cursor to bottom works on line that doesn't fit with 'smoothscroll'.
+  " Skipcol was adjusted later for cursor being on not visible part of line.
+  setlocal smoothscroll
+  norm zb
+  redraw
+  call assert_equal(520, winsaveview().skipcol)
+
+  bwipe!
+endfunc
 " vim: shiftwidth=2 sts=2 expandtab
```
zeertzjq:
```suggestion

" vim: shiftwidth=2 sts=2 expandtab
```

luukvbaal:
Backward compatibility is largely restored, which can be checked by comparing to the affected test files from before patch 9.1.0211 like so:

    git diff 9ccc2972373c8310c20ae7621b9c634d0dc43e26 -- \*test_diffmode.vim \*test_normal.vim \*test_edit.vim \*test_scroll_opt.vim

The tests also include some changes due to what I believe to bugfixes/behavioral improvements for 'smoothscroll'. I'll review myself one more time tomorrow.

itchyny:
Previously, the mappings keep the column of the cursor. I tested your patch but it moves the cursor to the end of line.

luukvbaal:
Repro steps please.

itchyny:
Open eval.c with `vim src/eval.c`, `20|` to move to column 20, `5<C-d>` to scroll 5 lines. The cursor should be at column 20.

luukvbaal:
Thanks, steps seem to miss `:se nostartofline` but I'll fix it.

src/normal.c
```diff
@@ -5771,6 +5771,15 @@ nv_g_home_m_cmd(cmdarg_T *cap)
...
 	    i = (virtcol - width1) / width2 * width2 + width1;
+
+	// When ending up below 'smoothscroll' marker, move just beyond it so
+	// that skipcol is not adjusted later.
+	if (curwin->w_skipcol > 0 && curwin->w_cursor.lnum == curwin->w_topline)
+	{
+	    int overlap = sms_marker_overlap(curwin, -1);
+	    if (overlap > 0 && i == curwin->w_skipcol)
+		i += overlap;
+	}
```
luukvbaal, line 5782: I think this better matches the documentation for "g0":

    To the first character of the screen line.

And it allows us to use it to move to the beginning of the screen line for page scrolling with 'smoothscroll' and 'startofline' enabled. Which is what I think should happen. Without this change that would cause us to scroll a different amount of lines than requested.

src/move.c
```diff
@@ -3087,73 +3094,125 @@ static int get_scroll_overlap(int dir)
...
-#endif
-    linenr_T	prev_topline = curwin->w_topline;
-    colnr_T	prev_skipcol = curwin->w_skipcol;
+    int		nochange = TRUE;
+    colnr_T	prev_col = curwin->w_cursor.col;
+    colnr_T	prev_curswant = curwin->w_curswant;
+    linenr_T	prev_lnum = curwin->w_cursor.lnum;
+    oparg_T	oa = { 0 };
+    cmdarg_T	ca = { 0 };
+    ca.oap = &oa;
```
luukvbaal, line 3156: This is maybe not ideal but re-using the logic from the normal commands seemed desirable to me. Could refactor to expose some inner functions if that is preferred.

src/testdir/test_scroll_opt.vim
```diff
@@ -453,7 +453,7 @@ func Test_smoothscroll_long_line_showbreak()
       vim9script
       # a line that spans four screen lines
       setline(1, 'with lots of text in one line '->repeat(6))
-      set smoothscroll scrolloff=0 showbreak=+++\ 
+      set smoothscroll scrolloff=0 showbreak=+++\
```
zeertzjq: ?

luukvbaal: Auto remove trailing whitespace...

src/move.c
```diff
@@ -3087,73 +3094,126 @@ static int get_scroll_overlap(int dir)
...
-    {
-	// Scroll at least one full line without 'smoothscroll'.
-#ifdef FEAT_DIFF
-	count -= plines_nofill(curwin->w_topline);
-#else
-	count -= plines(curwin->w_topline);
-#endif
-	scroll_redraw(dir == FORWARD, 1);
+	// Don't scroll if we already know that it will reveal end of buffer lines.
+	int buflen = curbuf->b_ml.ml_line_count;
```
chrisbra: can you move buflen declaration to the beginning of the function?

src/move.c
```diff
@@ -3087,73 +3094,126 @@ static int get_scroll_overlap(int dir)
...
+	if (dir == BACKWARD
+	    || (curwin->w_botline - 1 < buflen)
+	    || (curwin->w_p_sms && curwin->w_botline - 1 == buflen
+		&& curwin->w_skipcol < linetabsize(curwin, buflen)))
+	{
+	    nochange = scroll_with_sms(dir, &count);
+	    validate_botline();
+	    // Hide any potentially revealed end of buffer lines.
+	    if (!nochange && curwin->w_botline - 1 == curbuf->b_ml.ml_line_count)
```
chrisbra: !nochange && curwin->w_botline - 1 == buflen

src/move.c
```diff
@@ -3087,73 +3094,126 @@ static int get_scroll_overlap(int dir)
...
+	    {
+		curwin->w_cursor.lnum = curbuf->b_ml.ml_line_count;
```
chrisbra: same here: `buflen`

src/move.c
```diff
@@ -3087,73 +3094,126 @@ static int get_scroll_overlap(int dir)
...
+	else if (dir == FORWARD)
+	    cursor_down_inner(curwin, count);
+	else
+	    cursor_up_inner(curwin, count);
+	curwin->w_curswant = prev_curswant;
+
+	if (get_scrolloff_value())
+	    cursor_correct();
 #ifdef FEAT_FOLDING
-	// Move cursor to first line of closed fold.
```
chrisbra, line 3136: please leave the comment

src/normal.c
chrisbra: thanks, yes, makse sense

chrisbra (commented):
looks mostly good, I left a few minor comments to be addressed.

PR #14378, opened 2024-04-02
runtime(jq): add support

Freed-Wu:
runtime/filetype.vim contains jq, however its syntax is missing.

### DIFF
--- a/runtime/compiler/jq.vim
+++ b/runtime/compiler/jq.vim
@@ -1,0 +1,28 @@
+" Vim compiler file
+" Language:	jq
+" Maintainer:	Vito C <vito.blog@gmail.com>
+" Upstream:	https://github.com/vito-c/jq.vim
+" Last Change:	2015 Nov 28
+
+if exists('b:current_compiler')
+  finish
+endif
+let b:current_compiler = 'jq'
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+if exists(':CompilerSet') != 2
+  command -nargs=* CompilerSet setlocal <args>
+endif
+if has('unix')
+  CompilerSet makeprg=jq\ -f\ %:S\ /dev/null
+else
+  CompilerSet makeprg=jq\ -f\ %:S\ nul
+endif
+CompilerSet errorformat=%E%m\ at\ \\<%o\\>\\,\ line\ %l:,
+            \%Z,
+            \%-G%.%#
+
+let &cpoptions = s:save_cpoptions
+unlet s:save_cpoptions
--- a/runtime/ftplugin/jq.vim
+++ b/runtime/ftplugin/jq.vim
@@ -1,0 +1,21 @@
+" Vim filetype plugin
+" Language:	jq
+" Maintainer:	Vito C <vito.blog@gmail.com>
+" Upstream:	https://github.com/vito-c/jq.vim
+" Last Change:	2015 Nov 28
+
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+let b:undo_ftplugin = 'setl commentstring<'
+
+setlocal commentstring=#%s
+compiler jq
+
+let &cpoptions = s:save_cpoptions
+unlet s:save_cpoptions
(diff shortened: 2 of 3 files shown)


zeertzjq:
The compiler and ftplugin files are missing headers, and I think you should also mention that this is taken from <https://github.com/vito-c/jq.vim>.

Freed-Wu:
Thanks, fixed.

clason:
It's not just about the license, you also need to attribute where you took it from in the file (look at other syntax files, e.g. `zig.vim`).

chrisbra:
@Freed-Wu you mention all those people as maintainers for the newly to be included runtime files. Please check with them first, they are actually confortable at  all  with this.

Shane-XB-Qian:
> @vito-c Thanks for you publishing your work in MIT license which allow it is re-distributed. 👍

ahhhhhhh... Western friends maybe nice to directly ask or talk, but @Freed-Wu i think they may wish be respected their work be handled smoothly too, besides those tens of runtime files rush updated now (from you) who you expected them be maintained by whom in the future? :smile:

Freed-Wu:
Firstly we should wait their responses. Then:

- If response is 'I don't allow it is included to vim runtime'. We have to close it.
- If respones is 'OK, I can update it in the future'. We continue.
- If response is 'I support it is included to vim runtime, however I have lose interest', I'll ask for the permission of upstream repo, then update it by mineself.

Shane-XB-Qian:
i meant:
1. it is their work/product, they may need some comfort, vs notify directly, it was not just legal issue, even was allowed.
2. runtime files in official vim usually maintained by those their authors, unless you were permitted and able & keep to maintain them all (if license allowed and they were really No Resp).
3. vim maintainer IF allowed or was happy you to do so? :smile:

Freed-Wu:
You are right. I have added acknowledgments for these PR to thank for their effort. :smile:

Shane-XB-Qian:
anyway, IMO: except those just wrote and maintained from yourself, 
others those left runtime PR tickets better be all marked as 'Draft' and trying to wait their resp.

Freed-Wu:
> be all marked as 'Draft'

Yes, they are drafts.

vito-c:
Hello. I think it would be great to include jq syntax files directly into vim. I have a  few questions about the process though. Who becomes the maintainer of these files? How does the maintainer get notified about issues? I remember when I made some fixes to json syntax awhile back the maintainer was MIA.

chrisbra:
In the current form, you are maintainer. If you want to address issues or have bugs fixed, I'd like you to create a PR with your (approved) changes here.

Shane-XB-Qian:
FYI: updated contribute md link page: https://github.com/vim/vim/discussions/13087#discussioncomment-8992229

vito-c:
@chrisbra @Shane-XB-Qian should I open another PR with the files? Also if no one is stepping up to take care of the json syntax files that seems like something that I could also put on my plate. What are your thoughts?

Shane-XB-Qian:
I think yes. but up to @chrisbra
// and thanks for your resp and taken.

--
shane.xb.qian

Shane-XB-Qian:
I meant:
you can link your new opened PR to this and marked `closes #14378` in its desc.
as for json seems there some dialect version of json, you can check with @chrisbra for that.

--
shane.xb.qian

vito-c:
OK I have a follow up question. If I have a plugin that has syntax files and vim also has syntax files for the same language: then my plugin files should override the builtin ones correct? I just wanted to clarify this so that I could then maintain changes in my local plugin, test them, then port them over. Is there anyway to write unit tests or these?
Thanks!

chrisbra:
Yes this should typically happen, because your local syntax plugin is usually in `~/.vim` runtimepath and comes before the system wide runtime folder from `$VIMRUNTIME`

For unit tests, have a look at the syntax screen tests here: https://github.com/vim/vim/tree/master/runtime/syntax/testdir

Freed-Wu:
> If you rather like some small changes,
> rather like to rework most of it, we can close it and await an updated PR from your side then

So should I make some adjustments or close it to wait the original author's PR?

Freed-Wu:
I create some PR in upstream, Let us wait the original author's PR after accept it :smile:

https://github.com/vito-c/jq.vim/pull/20/files

PR #14380, opened 2024-04-02
runtime(zathurarc): add support

Freed-Wu:


src/testdir/test_filetype.vim
```diff
@@ -823,6 +823,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     yaml: ['file.yaml', 'file.yml', 'file.eyaml', 'any/.bundle/config', '.clangd', '.clang-format', '.clang-tidy', 'file.mplstyle', 'yarn.lock'],
     yang: ['file.yang'],
     yuck: ['file.yuck'],
+    zathurarc: ['zathurarc'],
```
zeertzjq: These keys are sorted alphabetically, so this should be after `z8a`

Freed-Wu: Thanks, fixed.

chrisbra:
just being explicit here: @Freed-Wu you are volunteering to maintain those files?

runtime/ftplugin/zathurarc.vim
```diff
@@ -0,0 +1,22 @@
...
+let b:did_ftplugin = 1
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+let b:undo_ftplugin = 'setlocal comments< commentstring< include< iskeyword<'
+setlocal comments=:#
+setlocal commentstring=#\ %s
+setlocal include=^\sinclude
+setlocal iskeyword+=-
```
chrisbra: are you setting this because of the syntax keywords containing a `-`? In that case please use the `:syn iskeyword` command instead

chrisbra (changes requested):

runtime/ftplugin/zathurarc.vim
Freed-Wu: Fixed.

runtime/ftplugin/zathurarc.vim
```diff
@@ -0,0 +1,21 @@
...
+" Latest Revision:      2024-04-02
+
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
```
Konfekt, line 15: How about adding a [&keywordprg](https://github.com/Konfekt/vim-keyword-lookup/blob/46edeedadac4d30f8764dce1a7d53ceda66cec82/ftdetect/keywordprg.vim#L17)

chrisbra: I suppose you meant line 17 instead?

chdiza: Shouldn't the "URL" field point at some kind of upstream repo, rather than the zathura docs?

Freed-Wu: > some kind of upstream repo

Should be `Upstream:`?

chrisbra: URL should point to your upstream repo if you have one. You can rename the current URL to Documentation I guess.

PR #14381, opened 2024-04-02
runtime(typst): add support

Freed-Wu:


jiangyinzuo:
Seems that these codes come from https://github.com/kaarmu/typst.vim, should we contact the author @kaarmu  ?

chrisbra:
Yes, ideally this should come from the original maintainer (who offers to become maintainer of the runtime file in Vim). 
So @kaarmu please ACK if this is fine with you to distribute together with Vim.

@Freed-Wu I appreciate your effort to make Vim work with more filetypes, but please credit original authors in your contributions  and check with them if they are okay with including it here (also from a license perspective) before submitting PR here. Thanks.

Freed-Wu:
> check with them if they are okay with including it here

OK, will ping them

> also from a license perspective

Which license can indicate it? Many vim plugins use vim license, is it OK?

chrisbra:
Yes, Vim license allows and is prefered. For other licenses, it depends if they allow re-distribution.

Freed-Wu:
> it depends if they allow re-distribution.

As I know it looks like any all of open source license allow re-distribution? So just add a note about its license is MIT or vim or GPL or others is OK, right?

clason:
Even if it's not legally required, it's simply proper etiquette to ask before reusing someone else's work. They may also have good reason for not wishing these files to be included here (e.g., they consider them not ready yet or under active development).

clason:
> Agree. For those files which are updated some years ago, we can think they are stable.

Or abandoned. It's better to ask than to assume.

kaarmu:
Author here. The project is not abandoned and still under WIP since Typst is a relatively new project (only a year old). While I think that the core syntax is _perhaps_ stable enough, I do not think it appropriate to merge the rest of my plugin like this because much of it is experimental.

Licensing aside, I am also a bit confused about the intent here since there's multiple PRs opened, only in the last day, of which @Freed-Wu authored none. As was already mentioned, it would have been nice to have a discussion about it before this PR was made.

If it is okay for the vim maintainers, I would like to close this PR and open another for syntax alone.

Freed-Wu:
> If it is okay for the vim maintainers, I would like to close this PR and open another for syntax alone.

No serious, you are welcome :+1:

PR #14385, opened 2024-04-02
runtime(uci): add support

Freed-Wu:


Freed-Wu:
@cmcaine Thanks for your GPL3 work!

chrisbra:
@cmcaine please ACK for your work to be included and distributed with Vim.

runtime/filetype.vim
```diff
@@ -2897,6 +2897,9 @@ au BufNewFile,BufRead {.,}tmux*.conf*		setf tmux
 " Universal Scene Description
 au BufNewFile,BufRead *.usda,*.usd		setf usd
 
+" Uci
+au BufNewFile,BufReadPost */etc/config/*		setf uci
+
```
cmcaine:
```suggestion
```

I think this would lead to UCI being used for a lot of files it shouldn't be.

runtime/ftplugin/uci.vim
```diff
@@ -0,0 +1,13 @@
+" Vim ftplugin file
+" Language:	OpenWrt Unified Configuration Interface
+" Maintainer:	Colin Caine <colin.caine@manchester.ac.uk>
```
cmcaine:
```suggestion
" Maintainer:	Colin Caine <complaints@cmcaine.co.uk>
```

runtime/syntax/uci.vim
```diff
@@ -0,0 +1,33 @@
+" Vim syntax file
+" Language:	OpenWrt Unified Configuration Interface
+" Maintainer:	Colin Caine <colin.caine@manchester.ac.uk>
```
cmcaine:
```suggestion
" Maintainer:	Colin Caine <complaints@cmcaine.co.uk>
```

src/testdir/test_filetype.vim
```diff
@@ -758,6 +758,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     typescriptreact: ['file.tsx'],
     ungrammar: ['file.ungram'],
     uc: ['file.uc'],
+    uci: ['any/etc/config/file'],
```
cmcaine, line 767:
```suggestion
```

If removing the line in filetype.vim...

runtime/ftplugin/uci.vim
```diff
@@ -0,0 +1,13 @@
...
+" Last Change:	2021 Sep 19
+"
+" For more information on uci, see https://openwrt.org/docs/guide-user/base-system/uci
+
+" UCI files are indented with tabs.
+let b:did_uci = 1
+setl noet
+
+setl commentstring=#\ %s
+let b:undo_ftplugin = "setlocal et< cms<"
```
cmcaine: Bump date and put the comment back with the line that it comments.

```suggestion
" Last Change:	2024 Apr 4
"
" For more information on uci, see https://openwrt.org/docs/guide-user/base-system/uci

let b:did_uci = 1

" UCI files are indented with tabs.
setl noet
setl commentstring=#\ %s

let b:undo_ftplugin = "setlocal et< cms<"
```

cmcaine (changes requested):

runtime/filetype.vim
chrisbra: agree, that looks wrong. Is there a better pattern for uci files?

cmcaine: I don't think there's any better way of matching on their paths.

Paths to UCI files:
https://openwrt.org/docs/guide-user/base-system/uci#configuration_files

This pattern matches all the UCI files, and rarely matches other files in my use, which may be good enough, but I'm sure it will match files that it is not intended to.

Maybe that's fine. I don't know what the baseline is here or what the consequences are for plugins if vim starts assigning the wrong filetype here.

glepnir:
@chrisbra  add original author to co-author when you push would be better.

chrisbra:
okay, can do this before I merge

Freed-Wu:
Fixed

runtime/filetype.vim
cmcaine: I'm not sure this is resolved yet. Can someone confirm that it's okay for filetypes.vim to risk false positives, even for obscure filetypes like UCI?

cmcaine:
Can someone respond to this thread before this PR is merged, please?

https://github.com/vim/vim/pull/14385#discussion_r1550195092

runtime/filetype.vim
Freed-Wu: I think the best way is let upstream use some path like `/etc/uci-config/`?

dkearns: > I'm not sure this is resolved yet. Can someone confirm that it's okay for filetypes.vim to risk false positives, even for obscure filetypes like UCI?

It looks like all the files start with `^\s*config\>`, is that right?  If so, you could also test for that pattern in the matched files.

cmcaine: The files are all in the /etc/config directory, and I matched */etc so that accessing an openwrt filesystem mounted on e.g. sshfs will still work.

The filenames don't have a matchable pattern, imo.

> I think the best way is let upstream use some path like /etc/uci-config/?

I don't think openwrt are gonna change their config directory name to accommodate this plugin :)

----

Could someone respond to this question I asked?

> Can someone confirm that it's okay for filetypes.vim to risk false positives, even for obscure filetypes like UCI?

runtime/filetype.vim
```diff
@@ -2897,6 +2897,9 @@ au BufNewFile,BufRead {.,}tmux*.conf*		setf tmux
 " Universal Scene Description
 au BufNewFile,BufRead *.usda,*.usd		setf usd
 
+" Uci
+au BufNewFile,BufReadPost */etc/config/*		setf uci
```
dkearns: This needs to use `s:StarSetf()` and be moved to that section the file.

`BufReadPost` should be `BufRead`

dkearns (changes requested):

runtime/filetype.vim
dkearns: > The filenames don't have a matchable pattern, imo.

I was suggesting matching the file **content** with that pattern.

>  Can someone confirm that it's okay for filetypes.vim to risk false positives, even for obscure filetypes like UCI?

False positives are an unavoidable risk in filetype detection.  The path looks terribly general but a [quick search](https://github.com/search?q=path%3A*%2Fetc%2Fconfig%2F*&type=code&p=1) on GitHub suggests it may not be too much of a problem.

runtime/ftplugin/uci.vim
```diff
@@ -0,0 +1,15 @@
+" Vim ftplugin file
+" Language:	OpenWrt Unified Configuration Interface
+" Maintainer:	Colin Caine <complaints@cmcaine.co.uk>
+" Upstream:	https://github.com/cmcaine/vim-uci
+" Last Change:	2024 Apr 4
+"
+" For more information on uci, see https://openwrt.org/docs/guide-user/base-system/uci
+
+let b:did_uci = 1
```
dkearns: This line should be replaced with

```vim
if exists("b:did_ftplugin")      
  finish
endif
let b:did_ftplugin = 1
```

dkearns (changes requested):

Shane-XB-Qian:
> False positives are an unavoidable risk in filetype detection. The path looks terribly general but a quick search<https://github.com/search?q=path%3A*%2Fetc%2Fconfig%2F*&type=code&p=1> on GitHub suggests it may not be too much of a problem.

it maybe not worthy to take this risk, and count on the weight of searching in github seems was weird too, such general path looked easily to be used for something else at users side.

--
shane.xb.qian

runtime/filetype.vim
cmcaine: Right, sorry for misunderstanding you!

I didn't know filetype.vim was allowed to match on file contents :)

For what it's worth I'm okay with the filepath pattern I originally had, and I'm also okay if the file contents are also scanned, but I don't know how to do that :)

Anyway, I'm gonna peace out of this issue now. Tag me again if I'm needed!

runtime/filetype.vim
Freed-Wu: Fixed

runtime/ftplugin/uci.vim
Freed-Wu: Fixed.

runtime/filetype.vim
Shane-XB-Qian: @cmcaine please check dist/ft.vim they mostly are all there for content detection.

Freed-Wu:
Rebased.

Freed-Wu:
> content detection

@cmcaine Can upstream detect the content of file to ensure it is uci?

Shane-XB-Qian:
> > it maybe not worthy to take this risk, and count on the weight of searching in github seems was weird too, such general path looked easily to be used for something else at users side.
> 
> Well, a GitHub search is pretty much all we have to go on other than a finger in the wind.  If a search had returned 100,000 unrelated files we'd _know_ it was a problem.

um... tho github "is" (supposed it is) the most popular one for OSS, but it did not mean it is the standard answer to such question, besides some users preferred offline..
similar question like do you believe all answers which GPT gave you? ahahahahaha...
// no need to be serious on this, it was just a plain feeling of 'weird', no other meaning.. :smile:

anyway, `*/etc/config/*` should be a too general path, i suggested using 'content detection' if there was a way to do so, just a suggestion.

-- 
shane.xb.qian

Freed-Wu:
Perhaps close it and wait original author's PR will be better before upstream's context detect is updated?

cmcaine:
I don't really wanna do that work. I'm happy to accept a PR for it.

(I mean, maybe I'd get around to it eventually, but you shouldn't count on
that)

PR #14390, opened 2024-04-02
Prefix the class name with the script name in the stack trace.

### LINKED ISSUE #14376: In a stacktrace, `class` name doesn't have `<SNR>` information
### Steps to reproduce

Note in the following stack trace there is `C1.H[1]` and `C1.Create[1]`. The same class name, but in different files. That information should be visible.

```
imported_file: :source buffer=1[31]..function <SNR>36_F[1]..C1.H[1]..C1.Create[1]
```

Run the following (the temp file name might have to be changed) to see the problem

```
vim9script

# File that gets imported. Change this name if ...
var import_file_name = '/tmp/TempScript' .. string(rand()) .. '.vim'

var lines =<< trim END
    vim9script

    export class C1
        static def Create()
            echo 'imported_file:' expand('<stack>')
        enddef
    endclass
END

writefile(lines, import_file_name)

import import_file_name as imported_file

delete(import_file_name)

class C1
    static def H()
        imported_file.C1.Create()
    enddef
endclass

def F()
    C1.H()
enddef
F()
```

### Expected behaviour

Expect stack trace like
```
imported_file: :source buffer=1[31]..function <SNR>36_F[1]..<SNR>36_C1.H[1]..<SNR>##_C1.Create[1]
```

Amusingly, there's #12078, which I filed a little year ago. Guess I didn't think about it enough at that time. I'll take a look at the fix for that, and maybe I can come up with a fix. But first, a question:

@chrisbra @yegappan  Are you OK with fixing this by changing the existing output of `expand('<stack>')`? (Is is considered not backwards compatible in a bad way?)

### Version of Vim

9.1.252

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

errael:
Fix #14376

### DIFF
--- a/src/scriptfile.c
+++ b/src/scriptfile.c
@@ -215,6 +215,9 @@ estack_sfile(estack_arg_T which UNUSED)
 	    if (*class_name != NUL)
 	    {
 		// For class methods prepend "<class name>." to the function name.
+		ga_concat(&ga, (char_u *)"<SNR>");
+		ga.ga_len += vim_snprintf((char *)ga.ga_data + ga.ga_len, 23,
+		       "%d_", entry->es_info.ufunc->uf_script_ctx.sc_sid);
 		ga_concat(&ga, class_name);
 		ga_append(&ga, '.');
 	    }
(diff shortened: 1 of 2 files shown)


errael:
@yegappan Modified an existing test.

PR #14392, opened 2024-04-02
runtime(vim): Add typespec filetype

happenslol:
This adds a filetype for the [typespec language](https://typespec.io/), based on the `.tsp` extension.


chrisbra:
thanks!

PR #14394, opened 2024-04-03
Problems with "zb" and scrolling to new topline with 'smoothscroll'

luukvbaal:
Problem: "zb" does not reveal filler lines at the start of a buffer.
          Scrolled cursor position with 'smoothscroll' is unpredictable,
          and may reset skipcol later if it is not visible.
Solution: Replace confusing for loop that reaches final control value too
          early with while loop. Set "w_curswant" accordingly so cursor
          will be placed in visible part of topline.


chrisbra:
thanks!

PR #14396, opened 2024-04-03
Normal mode TextChanged isn't tested properly

zeertzjq:
Problem:  Normal mode TextChanged isn't tested properly.
Solution: Combine Test_Changed_ChangedI() and Test_Changed_ChangedI_2()
          and also run it on Windows. Fix a typo in main.c.


chrisbra:
thanks!

PR #14398, opened 2024-04-03
fix(dialog): save unamed buffer correctly

glepnir:
Problem: when bufname is empty it can not save correctly

Solution: set bufname before save.

see https://github.com/neovim/neovim/issues/21682

### DIFF
--- a/src/ex_cmds2.c
+++ b/src/ex_cmds2.c
@@ -164,6 +164,8 @@ dialog_changed(
     int		ret;
     buf_T	*buf2;
     exarg_T     ea;
+    int		empty_buf;
+    char_u	cwd[MAXPATHL];
 
     dialog_msg(buff, _("Save changes to \"%s\"?"), buf->b_fname);
     if (checkall)
@@ -181,10 +183,30 @@ dialog_changed(
 	// May get file name, when there is none
 	browse_save_fname(buf);
 #endif
-	if (buf->b_fname != NULL && check_overwrite(&ea, buf,
-				    buf->b_fname, buf->b_ffname, FALSE) == OK)
+	empty_buf = buf->b_fname == NULL ? TRUE : FALSE;
+	if (empty_buf == TRUE)
+	{
+	    buf->b_fname = vim_strsave((char_u *)"Untitled");
+	    if (mch_dirname(cwd, MAXPATHL) == OK)
+	    {
+		buf->b_ffname = concat_fnames(vim_strsave(cwd), buf->b_fname, TRUE);
+	    }
+	}
+	if (check_overwrite(&ea, buf, buf->b_fname, buf->b_ffname, FALSE) == OK)
+	{
 	    // didn't hit Cancel
-	    (void)buf_write_all(buf, FALSE);
+	    if (buf_write_all(buf, FALSE) == OK)
+		return;
+	}
+
+	// restore to empty when write failed
+	if (empty_buf == TRUE)
+	{
+	    vim_free(buf->b_fname);
+	    buf->b_fname = NULL;
+	    vim_free(buf->b_ffname);
+	    buf->b_ffname = NULL;
+	}
     }
     else if (ret == VIM_NO)
     {
--- a/src/ex_cmds2.c
+++ b/src/ex_cmds2.c
@@ -165,7 +165,6 @@ dialog_changed(
     buf_T	*buf2;
     exarg_T     ea;
     int		empty_buf;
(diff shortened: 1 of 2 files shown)


src/ex_cmds2.c
```diff
@@ -181,10 +181,30 @@ dialog_changed(
 	// May get file name, when there is none
 	browse_save_fname(buf);
 #endif
-	if (buf->b_fname != NULL && check_overwrite(&ea, buf,
-				    buf->b_fname, buf->b_ffname, FALSE) == OK)
+	int empty_buf = buf->b_fname == NULL ? TRUE : FALSE;
```
chrisbra: Please move the declaration to the beginning of the function.

src/ex_cmds2.c
```diff
@@ -181,10 +181,30 @@ dialog_changed(
...
+	if (empty_buf == TRUE)
+	{
+	    buf->b_fname = vim_strsave((char_u *)"Untitled");
+	    if (STRLEN(NameBuff) == 0 && mch_dirname(NameBuff, MAXPATHL) == OK)
```
chrisbra: I don't understand this. What is the reason for the `STRLEN()`? and why do you copy only the current directory into this buffer?

src/ex_cmds2.c
```diff
@@ -181,10 +181,30 @@ dialog_changed(
...
+	    {
+		buf->b_ffname = vim_strsave(NameBuff);
```
chrisbra: Now we have the directory in `buf->b_ffname`. But shouldn't this contain the Filename + the full path?

src/testdir/test_buffer.vim
```diff
@@ -270,6 +270,20 @@ func Test_goto_buf_with_confirm()
...
+  call writefile(['foo'], 'Xfoo', 'D')
+  enew
+  call setline(1, 'test')
+  call feedkeys('y', 'L')
+  confirm edit Xfoo
+  call assert_equal('Xfoo', @%)
+  bprevious
+  call assert_equal('Untitled', @%)
+  close!
+endfunc
```
chrisbra, line 269: Does this test actually work? For me it fails with:
```
Vim(edit):E338: Sorry, no file browser in console mode @ command line..script /home/chrisbra/code/vim-upstream/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_confirm_write_unamed, line 7
```

chrisbra (changes requested):

src/ex_cmds2.c
```diff
@@ -181,10 +181,30 @@ dialog_changed(
...
+	}
+	if (check_overwrite(&ea, buf, buf->b_fname, buf->b_ffname, FALSE) == OK)
+	{
 	    // didn't hit Cancel
-	    (void)buf_write_all(buf, FALSE);
+	    if (buf_write_all(buf, FALSE) == OK)
+		return;
+	}
+
+	// resotre to empty when write failed
```
chrisbra: typo

src/ex_cmds2.c
```diff
@@ -181,10 +181,30 @@ dialog_changed(
...
+	if (empty_buf)
```
chrisbra: Style: here you are using `if(empty_buf)`, above you were using `if(empty_buf == TRUE)`

src/testdir/test_buffer.vim
glepnir: also doesn't work so i wonder how i got the browser notify . seems like ran into  FEAT_BROWSE  . seems like compiled with browse feature.

src/ex_cmds2.c
glepnir: sry for mistake.

src/ex_cmds2.c
glepnir: sometimes  NameBuff is empty for me after run :pwd it has value ..

glepnir: without `STRLEN` check the `Test_goto_buf_with_confirm` will failed.  thought is it correct when press y to confirm bdelete it it should failed ? it should write to Untitled.

src/testdir/test_buffer.vim
glepnir: Setting fname has been simplified. Adjusted previous test cases. But still enter the browse macro branch

chrisbra:
Thanks, I found a few more problems with this. Hopefully I could fix them and CI runs fine now.

PR #14399, opened 2024-04-03
runtime: Remove fallback :CompilerSet definition from compiler plugins

dkearns:
The `:CompilerSet` command was added in version Vim 6.4 which was released twenty years ago.  Other runtime files do not support versions of that vintage so it is reasonable to remove this fallback command definition now.


chrisbra:
Thanks, that makes sense.

Konfekt:
Maybe this is also an opportunity to update `:help :CompilerSet` that still recommends the fallback

PR #14403, opened 2024-04-03
runtime(java): Improve the matching of contextual keywords

zzzyxwvut:
- Recognise a `record` contextual keyword.
- Recognise `non-sealed`, `sealed`, and `permits` contextual  
  keywords.
- Admit `$` to keyword characters.
- Group `abstract`, `final`, `default`, `(non-)sealed`  
  (apart from `(non-)sealed`, the incompossibility of these  
  modifiers calls for attention).
- Remove another `synchronized` keyword redefinition.

I have also replaced a function with an expression.  Before  
patch `8.1.0515`, it should have been declared `:function!` to  
work with repeatable script sourcing; there is less to worry  
about with an expression.

References:
https://openjdk.org/jeps/395 (Records)
https://openjdk.org/jeps/409 (Sealed Classes)
https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.8


runtime/syntax/java.vim
```diff
@@ -2,7 +2,7 @@
 " Language:	Java
 " Maintainer:	Claudio Fleiner <claudio@fleiner.com>
 " URL:          https://github.com/fleiner/vim/blob/master/runtime/syntax/java.vim
-" Last Change:	2024 Mar 25
+" Last Change:	2024 Mar 26
```
dkearns: I think it would be a good idea, if you're willing, to officially take over as maintainer.

runtime/syntax/java.vim
```diff
@@ -48,34 +52,43 @@ syn match   javaOperator	"\%(\%(::\|\.\)[[:space:]\n]*\)\@80<!\<yield\>"
...
-
-" Java Modules(Since Java 9, for "module-info.java" file)
-if s:isModuleInfoDeclarationCurrentBuffer()
-    syn keyword javaModuleStorageClass	module transitive
-    syn keyword javaModuleStmt		open requires exports opens uses provides
-    syn keyword javaModuleExternal	to with
-    syn cluster javaTop add=javaModuleStorageClass,javaModuleStmt,javaModuleExternal
+" Note that a "module-info" file will be recognised with an arbitrary
+" file extension (or no extension at all) so that more than one such
+" a declaration for the same Java module can be maintained for modular
```
dkearns: "such declaration"

runtime/syntax/java.vim
```diff
@@ -48,34 +52,43 @@ syn match   javaOperator	"\%(\%(::\|\.\)[[:space:]\n]*\)\@80<!\<yield\>"
...
+" testing in a project without attendant confusion for IDEs, with the
+" ".java\=" extension used for a production version and an arbitrary
+" extension used for a testing version.
+let s:module_info_cur_buf = fnamemodify(bufname("%"), ":t") =~ '^module-info\%(\.class\>\)\@!'
+lockvar s:module_info_cur_buf
```
dkearns, line 84: Use of `:lockvar` here seems a bit unusual but if it means something to you stylistically then each to their own. :)  There is `:const` too.

runtime/syntax/java.vim
zzzyxwvut: Oops!

runtime/syntax/java.vim
zzzyxwvut: I'm willing and am going to set up a public Git repository  
for the syntax file and `input/` test files for another PR.  
What else do I need to do?

chrisbra: Move Claudio to `Former Maintainer` and add yourself as Maintainer. I think that should be all

dkearns: Also add yourself to CODEOWNERS.

I'm not trying to railroad you into extra work but if you're interested the ftplugin is also lacking a maintainer.  It's often easier if they're maintained together.

Thanks!

zzzyxwvut: Okay, I'll read up on ftplugins and make necessary changes  
with the next Java PR then.

chrisbra:
thanks

PR #14404, opened 2024-04-03
Test for TextChanged is flaky with ASAN

zeertzjq:
Problem:  Test for TextChanged is flaky with ASAN.
Solution: Also wait for the file to be non-empty.


chrisbra:
thanks!

Issue #14350, opened 2024-03-30
Vim9: Allow lambda blocks in (static) initialisers for classes and enums


zzzyxwvut:
### Steps to reproduce

The support for lambda blocks is somewhat uneven.  These  
groups can be identified with respect to functions (methods)  
and variables (fields):

1) Supported in functions and methods, e.g.:
```
vim9script

def IdFun1(): func(number): number
    return (num: number): number => {
	return num
    }
enddef
IdFun1()(4)

def IdFun2(): func(number): number
    return (num: number): number => {
        # OK
	return num
    }
enddef
IdFun2()(4)

class IdClass5
    def IdFun1(): func(number): number
	return (num: number): number => {
	    return num
	}
    enddef

    def IdFun2(): func(number): number
	return (num: number): number => {
	    # OK
	    return num
	}
    enddef

    static def IdFun3(): func(number): number
	return (num: number): number => {
	    return num
	}
    enddef

    static def IdFun4(): func(number): number
	return (num: number): number => {
	    # OK
	    return num
	}
    enddef
endclass

enum IdEnum5
    INSTANCE

    def IdFun1(): func(number): number
	return (num: number): number => {
	    return num
	}
    enddef

    def IdFun2(): func(number): number
	return (num: number): number => {
	    # OK
	    return num
	}
    enddef

    static def IdFun3(): func(number): number
	return (num: number): number => {
	    return num
	}
    enddef

    static def IdFun4(): func(number): number
	return (num: number): number => {
	    # OK
	    return num
	}
    enddef
endenum
```

2) Supported in RHS of script variables and with  
instance variable initialisers for classes, e.g.:
```
vim9script

const IdVar1: func(number): number = (num: number): number => {
    return num
}
IdVar1(4)
const IdVar2: func(number): number = (num: number): number => {
    # OK
    return num
}
IdVar2(4)

class IdClass1
    const Id: func(number): number = (num: number): number => {
	return num
    }
endclass
```

3) Not supported (`E1171`) with instance variable  
initialisers for enums and with class variable  
initialisers for classes and enums, e.g.:
```
vim9script

enum IdEnum1
    ID

    const Id: func(number): number = (num: number): number => {
	return num
    } # E1171
endenum

class IdClass3
    static const Id: func(number): number = (num: number): number => {
	return num
    } # E1171
endclass

enum IdEnum3
    ID

    static const Id: func(number): number = (num: number): number => {
	return num
    } # E1171
endenum

```

[...]

PR #14405, opened 2024-04-04
Support evaluating lambda blocks from a string

yegappan:
Fixes the issues reported in #14350.


yegappan:
There is a memory leak with this PR.  I will look at this later (don't merge this yet).

yegappan:
> There is a memory leak with this PR. I will look at this later (don't merge this yet).

This memory leak is addressed now in the updated PR.

chrisbra:
thanks!

PR #14408, opened 2024-04-04
runtime(vim): add earthfile filetype

glehmann:
for https://earthly.dev/

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -665,6 +665,9 @@ au BufNewFile,BufRead *.dtd			setf dtd
 " DTS/DSTI/DTSO (device tree files)
 au BufNewFile,BufRead *.dts,*.dtsi,*.dtso,*.its	setf dts
 
+" Earthfile
+au BufNewFile,BufRead Earthfile			setf earthfile
+
 " EDIF (*.edf,*.edif,*.edn,*.edo) or edn
 au BufNewFile,BufRead *.ed\(f\|if\|o\)		setf edif
 au BufNewFile,BufRead *.edn
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -220,6 +220,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     dylan: ['file.dylan'],
     dylanintr: ['file.intr'],
     dylanlid: ['file.lid'],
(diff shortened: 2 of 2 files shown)


runtime/filetype.vim
```diff
@@ -665,6 +665,9 @@ au BufNewFile,BufRead *.dtd			setf dtd
 " DTS/DSTI/DTSO (device tree files)
 au BufNewFile,BufRead *.dts,*.dtsi,*.dtso,*.its	setf dts
 
+" Earthfile
+au BufNewFile,BufRead Earthfile setf earthfile
```
dkearns: I'm sorry to do this to you but could you tab out the `setf` command to align with those above and below?

dkearns (changes requested):

runtime/filetype.vim
glehmann: I’d be pleased to do any required change, but I’m not sure to understand how to align `setf`: the command above and below are not at the same position, they are not preceded by the same number of space either. I fail to see the pattern to apply here :-/

dkearns: You probably have a different `tabstop` option value.  The file formatting assumes the default of 8.

There should be 3 tabs before `setf` in this case, rather than the single space, or you can apply the following patch.

Thanks.

```diff
diff --git a/runtime/filetype.vim b/runtime/filetype.vim
index 361e23ad5..40b17a0ef 100644
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -666,7 +666,7 @@ au BufNewFile,BufRead *.dtd			setf dtd
 au BufNewFile,BufRead *.dts,*.dtsi,*.dtso,*.its	setf dts
 
 " Earthfile
-au BufNewFile,BufRead Earthfile setf earthfile
+au BufNewFile,BufRead Earthfile			setf earthfile
 
 " EDIF (*.edf,*.edif,*.edn,*.edo) or edn
 au BufNewFile,BufRead *.ed\(f\|if\|o\)		setf edif
```

glehmann: oh! Indeed my editor is configured with 4 spaces per tabs.

Sorry to have missed that both when editing the file and with you comment. It should be ok now :-)

dkearns (approved):
Thanks.

chrisbra:
thanks!

PR #14410, opened 2024-04-04
Collect and tweak 'smoothscroll' cursor logic

luukvbaal:
Problem:  Logic to make sure cursor is in visible part of the screen after
          scrolling the text with 'smoothscroll' is scattered, asymmetric
          and contains bugs.
Solution: Adjust and create helper function for 'smoothscroll' cursor logic.

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -59,7 +59,7 @@ adjust_plines_for_skipcol(win_T *wp)
  * the window height.
  */
     static int
-plines_correct_topline(win_T *wp, linenr_T lnum)
+plines_correct_topline(win_T *wp, linenr_T lnum, int limit_winheight)
 {
     int n;
 #ifdef FEAT_DIFF
@@ -70,7 +70,7 @@ plines_correct_topline(win_T *wp, linenr_T lnum)
 	n = plines_win(wp, lnum, FALSE);
     if (lnum == wp->w_topline)
 	n -= adjust_plines_for_skipcol(wp);
-    if (n > wp->w_height)
+    if (limit_winheight && n > wp->w_height)
 	n = wp->w_height;
     return n;
 }
@@ -119,7 +119,7 @@ comp_botline(win_T *wp)
 	else
 #endif
 	{
-	    n = plines_correct_topline(wp, lnum);
+	    n = plines_correct_topline(wp, lnum, TRUE);
 	}
 	if (
 #ifdef FEAT_FOLDING
@@ -921,7 +921,7 @@ curs_rows(win_T *wp)
 	    else
 #endif
 	    {
-		wp->w_cline_row += plines_correct_topline(wp, lnum);
+		wp->w_cline_row += plines_correct_topline(wp, lnum, TRUE);
 		++lnum;
 	    }
 	}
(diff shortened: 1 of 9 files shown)


luukvbaal:
In trying to resolve the bug for the added test I struggled a bit with following the logic for adjusting the cursor due to it being scattered around. I think this improves the situation and may allow to more easily tweak the logic in the future. 

This patch currently favors keeping the cursor in the same column when scrolling the text up changes the topline. As opposed to the current behavior which keeps it near the top of the screen, which seems to be untested. I can try to restore that behavior, although I am not sure it makes sense currently. Because the same cannot happen when scrolling the text down changes the botline, because the cursor cannot be on a partially visible botline in the first place, so it is a bit asymmetric.

luukvbaal:
I keep encountering more bugs, mostly in small windows and with 'scrolloff'.
Attempts to fix them result in other tests failing so I settled on this for now...

src/move.c
```diff
@@ -3161,33 +3145,39 @@ pagescroll(int dir, long count, int half)
...
+	if (count > 0)
+	{
+	    validate_cursor();
+	    int	prev_wrow = curwin->w_wrow;
+	    nochange = scroll_with_sms(dir, count);
+	    if (!nochange)
 	    {
-		curwin->w_cursor.lnum = buflen;
-		scroll_cursor_bot(0, TRUE);
+		validate_cursor();
```
luukvbaal, line 3226: Instead of messing with curswant here which is already set accordingly in scrollup/down(), validate the cursor and bring it to the row it was previously on. I think the cursor position is much more predictable now with 'nostartofline'. 

There are still some bugs with 'scrolloff' with large lines in a small window that prevent scrolling or corrupt the screen. Those were also present before patch 9.1.0211, but now also apply to (half)-page scrolling in addition to Ctrl-E/Y.

luukvbaal:
TBH I think it is ready. I spent some more time trying to make the small window, large lines and 'scrolloff' case work somewhat better but all I could come up with was ugly special casing.

I think what's necessary to make that work properly is to allow the cursor to be on a not fully visible botline.

src/testdir/test_scroll_opt.vim
```diff
@@ -1071,6 +1077,13 @@ func Test_smoothscroll_next_topline()
   redraw
   call assert_equal(2, line('w0'))
 
+  " Cursor does not end up above topline, adjusting topline later.
+  setlocal nu cpo+=n
```
zeertzjq, line 1081: 'cpoptions' is a global option and needs to be restored at the end of the test.

src/move.c
```diff
@@ -3161,35 +3203,46 @@ pagescroll(int dir, long count, int half)
...
+	if (dir == FORWARD)
 	{
-	    nochange = scroll_with_sms(dir, &count);
-	    validate_botline();
-	    // Hide any potentially revealed end of buffer lines.
-	    if (!nochange && curwin->w_botline - 1 == buflen)
+	    int n = plines_win(curwin, curwin->w_topline, FALSE);
+		n -= adjust_plines_for_skipcol(curwin);
+	    if (n - count < curwin->w_height && curwin->w_topline < buflen)
+		n += plines_m_win(curwin, curwin->w_topline + 1, buflen, TRUE);
```
luukvbaal, line 3213: This does not include filler lines below the last line in the buffer. Thus Ctrl-D near the end of a file does not reveal any potential filler lines (nor did it before the inital halfpage smoothscroll patch). I think it would make sense if pressing Ctrl-D near the end of a file would reveal filler lines. A low-level plines fix for that could be:
```diff
@@ -347,8 +347,11 @@ plines_win(
 #if defined(FEAT_DIFF) || defined(PROTO)
     // Check for filler lines above this buffer line.  When folded the result
     // is one line anyway.
-    return plines_win_nofill(wp, lnum, limit_winheight)
+    int n = plines_win_nofill(wp, lnum, limit_winheight)
                                                   + diff_check_fill(wp, lnum);
+    if (lnum == wp->w_buffer->b_ml.ml_line_count)
+       n += diff_check_fill(wp, lnum + 1);
+    return n;
 }

 /*
```

I think it makes sense because plines() can never be called with lnum == b_ml.ml_line_count + 1, which is where filler lines below the last line are attached to. Otherwise we could handle it in pagescroll() directly.

zeertzjq: I don't think the change should be in `plines_win()` as that's inconsistent and may lead to more mistakes.

luukvbaal: Fair enough. Do you think adding it to plines_m_win() is fine? Otherwise I may need to duplicate it's logic in pagescroll(). Or change plines_m_win "last" argument to a pointer whose final value can be checked at the callsite.

luukvbaal: I will assume yes and assume the behavior is wanted and update the PR, let me know if it is not.

zeertzjq: I think adding it to `plines_m_win()` makes sense, considering that it also has other special cases like using `w_topfill` for `w_topline`.

chrisbra:
thank you. Let me include this now.

PR #14413, opened 2024-04-05
runtime: Remove more fallback :CompilerSet definitions from compiler plugins

jiangyinzuo:
Continue with #14399


jiangyinzuo:
@dkearns  It seems that your PR #14399 misses some `:CompilerSet` definition removal changes, most of them use single-quoted string `exists(':CompilerSet')`

dkearns (approved):
Thanks for cleaning up.

chrisbra:
thanks!

PR #14414, opened 2024-04-05
runtime: fix :compiler leaving behind a g:makeprg variable

zeertzjq:
Problem:  :compiler may leave behind a g:makeprg variable after #14336.
Solution: Use a script local variable.


chrisbra:
thanks!

PR #14416, opened 2024-04-05
runtime: add Roc filetype

### LINKED ISSUE #14415: Add Roc language filetype
[Roc](https://www.roc-lang.org/) is a fast, friendly, functional language.
Roc uses the `.roc` extension for its files. I am attempting to help make
the experience of developing Roc better by adding support for it in
various tools, and I think having Roc be automatically detected in Vim
would be useful.

nat-418:
Closes https://github.com/vim/vim/issues/14415

### DIFF
--- a/runtime/ftplugin/roc.vim
+++ b/runtime/ftplugin/roc.vim
@@ -1,0 +1,14 @@
+" Roc filetype plugin file
+" Language: Roc
+" Maintainer: nat-418 <93013864+nat-418@users.noreply.github.com>
+" Latest Revision: 2024-04-5
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+setlocal comments=:##,:#
+setlocal commentstring=#\ %s
+
+let b:undo_ftplugin = "setl com< cms<"
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -595,6 +595,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     rpgle: ['file.rpgle', 'file.rpgleinc'],
     robot: ['file.robot', 'file.resource'],
     robots: ['robots.txt'],
+    roc: ['file.roc'],
     ron: ['file.ron'],
     routeros: ['file.rsc'],
     rpcgen: ['file.x'],
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -1880,6 +1880,9 @@ au BufNewFile,BufRead *.robot,*.resource	setf robot
 " Robots.txt
 au BufNewFile,BufRead robots.txt		setf robots
 
+" Roc
+au BufNewFile,BufRead *.roc			setf roc
+
 " RON (Rusty Object Notation)
 au BufNewFile,BufRead *.ron			setf ron
 


runtime/ftplugin/roc.vim
```diff
@@ -0,0 +1,20 @@
...
+" Maintainer: nat-418 <93013864+nat-418@users.noreply.github.com>
+" Latest Revision: 2024-04-5
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:cpo_save = &cpo
+set cpo&vim
```
chrisbra: I think there is nothing in your plugin requiring the 
```
let s:cpo_save = &cpo
set cpo&vim
```
dance and you can remove it.

runtime/ftplugin/roc.vim
```diff
@@ -0,0 +1,20 @@
...
+
+setlocal comments=:#
+setlocal commentstring=#%s
```
dkearns: I don't know the language's formatting norms but it's probably more common to specify `commenstring` as `#\ %s` for a line comment. 

This is what the Elm ftplugin does, for example.

https://github.com/ElmCast/elm-vim/blob/4b71facd77297cb33bbb3b14894676cff0a9bd1d/ftplugin/elm.vim#L41

clason: Sorry for hijacking this discussion, but it struck me as a good (and relevant) opportunity: Is there an official Vim policy on commentstrings? Most but not all include a space after (resp. before) the comment marker(s). Is there interest in homogenizing (some of) them?

jamessan: I can see value in `commentstring=#\ %s`. I've also seen mention of `comments=b:#`, which I think should only be used if the syntax _requires_ whitespace after the `#`.

nat-418: > I don't know the language's formatting norms but it's probably more common to specify `commenstring` as `#\ %s` for a line comment.
> 
> This is what the Elm ftplugin does, for example.
> 
> https://github.com/ElmCast/elm-vim/blob/4b71facd77297cb33bbb3b14894676cff0a9bd1d/ftplugin/elm.vim#L41

Roc and Elm differ: in Roc, a comment can be written as `# comment` or `## doc comment`, the latter form having some special meaning to the Roc CLI tool. Also, Roc has no multiline strings, as Elm does. What is the best way to represent this if not `#%s`?  I have changed to `#\ %s`, assuming that is the correct answer.

src/testdir/test_filetype.vim
```diff
@@ -595,6 +595,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     rpgle: ['file.rpgle', 'file.rpgleinc'],
     robot: ['file.robot', 'file.resource'],
     robots: ['robots.txt'],
+    roc: ['file.roc'],
```
zeertzjq, line 598: Why did you add a test for filetype detection without implementing it?

nat-418: Mistake on my part

runtime/ftplugin/roc.vim
dkearns: > Sorry for hijacking this discussion, but it struck me as a good (and relevant) opportunity: Is there an official Vim policy on commentstrings? Most but not all include a space after (resp. before) the comment marker(s). Is there interest in homogenizing (some of) them?

I think it would probably be a good idea.

I don't like the lack of consistency throughout the runtime files, in general, but in the past these decisions were intentionally left to the maintainers.

runtime/ftplugin/roc.vim
```diff
@@ -0,0 +1,14 @@
...
+" Language: Roc
+" Maintainer: nat-418 <93013864+nat-418@users.noreply.github.com>
+" Latest Revision: 2024-04-5
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+setlocal comments=:#
```
dkearns: You probably want to use the following so that doc comments are distinctly matched.

`setlocal comments=:##,:#`

Then the appropriate comment character(s) will be used when formatting comments etc.

runtime/ftplugin/roc.vim
```diff
@@ -0,0 +1,14 @@
+" Roc filetype plugin file
+" Language: Roc
+" Maintainer: nat-418 <93013864+nat-418@users.noreply.github.com>
```
dkearns, line 3: Using a 'real' name would be preferred but I'm aware there's plenty of good reasons not to do so.

@chrisbra, do you care about this?  It was somewhat frowned upon in the past.

chrisbra: Yeah, I suppose the email is invalid anyhow? I prefer real-names as well, but won't insist on it.

nat-418: I use the github provided email to prevent spam. I use the moniker to avoid weirdos.

chrisbra: okay thanks

PR #14417, opened 2024-04-05
File name entered in GUI dialog is ignored

zeertzjq:
Problem:  File name entered in GUI dialog is ignored.
Solution: Only set file name to "Untitled" if GUI dialog didn't set it.


chrisbra:
ah, I missed that. Thanks for cleaning it up!

PR #14421, opened 2024-04-05
Test for TextChanged is still flaky with ASAN

zeertzjq:
Problem:  Test for TextChanged is still flaky with ASAN.
Solution: Don't index the result of readfile().

It turns out that with ASAN the file may become empty during a write
even if it's non-empty both before and after the write, in which case
indexing the result of readfile() will error, so use join() instead.

Also don't delete the file halfway the test, just in case it may cause
errors on the next read.

### DIFF
--- a/src/testdir/test_autocmd.vim
+++ b/src/testdir/test_autocmd.vim
@@ -4495,39 +4495,40 @@ func Test_Changed_ChangedI()
         \ {'term_rows': 10})
   call assert_equal('running', term_getstatus(buf))
   call WaitForAssert({-> assert_true(filereadable('XTextChangedI3'))})
+  defer delete('XTextChangedI3')
   call WaitForAssert({-> assert_equal([''], readfile('XTextChangedI3'))})
 
   " TextChanged should trigger if a mapping enters and leaves Insert mode.
   call term_sendkeys(buf, "\<CR>")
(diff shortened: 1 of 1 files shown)


chrisbra:
unrelated to your PR, but what is wrong with choco install python2, it seems to start failing the windows CI builds 😢 ? https://github.com/vim/vim/actions/runs/8566827152/job/23477355255?pr=14421#step:2:73

src/testdir/test_autocmd.vim
```diff
@@ -4537,13 +4537,13 @@ func Test_Changed_ChangedI()
     call term_sendkeys(buf, "\<Esc>")
     call WaitForAssert({-> assert_notmatch('^-- INSERT --', term_getline(buf, 10))})
     call term_sendkeys(buf, ":let [g:autocmd_n, g:autocmd_i] = ['', '']\<CR>")
-    call delete('XTextChangedI3')
+    call writefile([], 'XTextChangedI3')
```
chrisbra, line 4541: call writefile([], 'XTextChangedI3', 'D') ?

zeertzjq: This will delete the file every time `s:validate_mixed_textchangedi()` returns, which isn't really necessary as it only needs to delete the file at the end of the whole test. I guess it's better to add an explicit `:defer delete()` above.

chrisbra:
thanks!

PR #14426, opened 2024-04-05
Refactor to remove STRLEN() part 2

basilisk0315:
Refactor highlight.c to reduce the number calls to STRLEN(). At the same time improve the efficiency of various lookups, some of which are done many times.

My measurements on startup with an empty buffer:
In function highlight_set_termgui_attr(), the number of comparisons changed from 701 (which includes a call to STRLEN() and STRNICMP()) down to 427 (includes a call to STRNICMP() only).
In function gui_get_color_cmn(), the number of comparisons changed from 763 down to 246 (both only include a call to STRICMP()).
In function highlight_set_cterm_color(), the number of comparisons changed from 897 down to 295 (both only include a call to STRICMP()).

Cheers
John


src/testdir/test_highlight.vim
```diff
@@ -1086,8 +1086,8 @@ func Test_hlget()
   let lines =<< trim END
     call assert_notequal([], filter(hlget(), 'v:val.name == "Visual"'))
     call assert_equal([], hlget('SomeHLGroup'))
-    highlight MyHLGroup term=standout cterm=reverse ctermfg=10 ctermbg=Black
-    call assert_equal([{'id': hlID('MyHLGroup'), 'ctermfg': '10', 'name': 'MyHLGroup', 'term': {'standout': v:true}, 'ctermbg': '0', 'cterm': {'reverse': v:true}}], hlget('MyHLGroup'))
+    highlight MyHLGroup term=standout cterm=inverse ctermfg=10 ctermbg=Black
+    call assert_equal([{'id': hlID('MyHLGroup'), 'ctermfg': '10', 'name': 'MyHLGroup', 'term': {'standout': v:true}, 'ctermbg': '0', 'cterm': {'inverse': v:true}}], hlget('MyHLGroup'))
```
zeertzjq: This behavior change seems backward-incompatible.

basilisk0315: This comes about because both "reverse" and "inverse" share the same #define value (HL_INVERSE) and when I resorted the highlight table for searching, the sequence of the two attributes was reversed. A fix is to use an index which restores the original sequence of highlight attributes for display purposes.

chrisbra:
Thanks, I think this is good now.

PR #14429, opened 2024-04-06
Change variable types to resolve compiler warnings

mrdubya:
Windows compiles were reporting a few size_t to signed integer conversion warnings that can be resolved by changing the variable declarations to be size_t to start with.


chrisbra:
thanks!

PR #14430, opened 2024-04-06
runtime(vim): Update syntax generator, autocmd event list parsing

dkearns:
Track changes to the autocommand event list introduced in commit 78d742a.


chrisbra:
thanks!

PR #14433, opened 2024-04-06
regexp: wrong match when 'ic' is set and comparing multi-byte with single byte char

### LINKED ISSUE #14294: Unicode character ſ is matched as itself and as 's.'
### Steps to reproduce

1. Open a utf8 encoded file in vim containing the line "Die Gleichheit fordert das Nachdenken heraus durch Fragen, die ſich daran knüpfen und nicht ganz leicht zu beantworten ſind."
2. use `:set ignorecase`
3. replace the character ſ by s using `:%s/\%u017F/s/g`

### Expected behaviour

The two occurrences of 'ſ' will be replaced by 's' resulting in "Die Gleichheit fordert das Nachdenken heraus durch Fragen, die sich daran knüpfen und nicht ganz leicht zu beantworten sind." However, the actual result is "Die Gleichheit fordert dasNachdenken herausdurch Fragen, die sich daran knüpfen und nicht ganz leicht zu beantworten sind.", i.e., the two original s characters **together with the following character** are also replaced by 's' as if I had used the command `:%s/s./s/g`. See also [this discussion](https://stackoverflow.com/questions/78214264).

### Version of Vim

9.1.151 but also older like 8.0

### Environment

system: x86_64 GNU/Linux
terminal: konsole, linux
$TERM: linux, xterm-256color
$LANG: de_DE.UTF-8, en_GB.UTF-8, C.UTF-8

### Logs and stack traces

_No response_

chrisbra:
When the regexp engine compares two utf-8 codepoints case insensitively it may match an adjacent character, because it assumes it can step over as many bytes as the pattern contains.

Let's consider the pattern 'ſ' and the string 's '. When comparing and ignoring case, the single character 's' matches, and since it matches Vim will try to step over the match (by the amount of bytes of the pattern), assuming that since it matches, the length of both strings is the same.

However in that case, it should only step over the single byte value 's' by 1 byte and try to start matching after it again. So for the backtracking engine we need to ensure:
- we try to match the correct length for the pattern and the text
- in case of a match, we step over it correctly

For the NFA engine, it basically already handles it already. Here we just need to make sure, that find_match_text is only called if prog->match_text is actually not only NULL but also doesn't point to 0x0 (NUL).

fixes: #14294

### DIFF
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -3801,6 +3801,15 @@ utf_strnicmp(
  * two characters otherwise.
  */
     int
+mb_strnicmp2(char_u *s1, char_u *s2, int n1, int n2)
+{
+    if (n1 == n2 || !enc_utf8)
+	return mb_strnicmp(s1, s2, n1);
+    else
+	return utf_strnicmp(s1, s2, n1, n2);
+}
+
+    int
 mb_strnicmp(char_u *s1, char_u *s2, size_t nn)
 {
     int		i, l;
--- a/src/proto/mbyte.pro
+++ b/src/proto/mbyte.pro
@@ -48,6 +48,7 @@ int utf_islower(int a);
 int utf_tolower(int a);
 int utf_isupper(int a);
 int mb_strnicmp(char_u *s1, char_u *s2, size_t nn);
+int mb_strnicmp2(char_u *s1, char_u *s2, int n1, int n2);
 void show_utf8(void);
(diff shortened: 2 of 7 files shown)


zeertzjq:
It seems that syntax tests are failing on FreeBSD and macOS starting from this PR for some reason.

chrisbra:
Hm, I'll probably back out that change and will invest some time to get a BSD image up and try to reproduce.

PR #14434, opened 2024-04-06
filetype: no support for keymap files

0xAdk:
zmk uses devicetree files for configuration with a `.keymap` extension 

https://zmk.dev/docs/features/keymaps

btw should the autocmd line be split up since it's a bit long?


chrisbra:
I think it is fine, we have other lines that are that long. For future improvements we should split it up however.

PR #14435, opened 2024-04-07
autocmd may change cwd after :tcd and :lcd

zeertzjq:
Problem:  Autocommand may change currect directory after :tcd and :lcd.
Solution: Also clear tp_localdir and w_localdir when using aucmd_win.


chrisbra:
thanks!

PR #14436, opened 2024-04-07
Add cursor_on(), cursor_off()

Shougo:
Related: #14327 

I have added two APIs to hide cursor.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -141,8 +141,8 @@ cscope_connection([{num}, {dbpath} [, {prepend}]])
 cursor({lnum}, {col} [, {off}])
 				Number	move cursor to {lnum}, {col}, {off}
 cursor({list})			Number	move cursor to position in {list}
-cursor_off()			none	disable cursor
-cursor_on()			none	enable cursor
+cursor_off()			none	hide cursor
+cursor_on()			none	unhide cursor
 debugbreak({pid})		Number	interrupt process being debugged
 deepcopy({expr} [, {noref}])	any	make a full copy of {expr}
 delete({fname} [, {flags}])	Number	delete the file or directory {fname}
@@ -1961,7 +1961,7 @@ cursor({list})
 			GetCursorPos()->cursor()
 
 cursor_off()						*cursor_off()*
-		Disable cursor like |:sleep!|.
+		Hide cursor like |:sleep!|.
 
 cursor_on()						*cursor_on()*
 		Enable cursor.
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -141,6 +141,8 @@ cscope_connection([{num}, {dbpath} [, {prepend}]])
 cursor({lnum}, {col} [, {off}])
 				Number	move cursor to {lnum}, {col}, {off}
 cursor({list})			Number	move cursor to position in {list}
+cursor_off()			none	disable cursor
+cursor_on()			none	enable cursor
 debugbreak({pid})		Number	interrupt process being debugged
 deepcopy({expr} [, {noref}])	any	make a full copy of {expr}
 delete({fname} [, {flags}])	Number	delete the file or directory {fname}
@@ -1957,6 +1959,12 @@ cursor({list})
 
 		Can also be used as a |method|: >
 			GetCursorPos()->cursor()
+
+cursor_off()						*cursor_off()*
+		Disable cursor like |:sleep!|.
+
+cursor_on()						*cursor_on()*
+		Enable cursor.
 
 debugbreak({pid})					*debugbreak()*
 		Specifically used to interrupt a program being debugged.  It
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -906,8 +906,8 @@ Cursor and mark position:		*cursor-functions* *mark-functions*
 	wincol()		window column number of the cursor
 	winline()		window line number of the cursor
 	cursor()		position the cursor at a line/column
-	cursor_off()		disable cursor
-	cursor_on()		enable cursor
+	cursor_off()		hide cursor
+	cursor_on()		unhide cursor
 	screencol()		get screen column of the cursor
 	screenrow()		get screen row of the cursor
 	screenpos()		screen row and col of a text character
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -906,6 +906,8 @@ Cursor and mark position:		*cursor-functions* *mark-functions*
 	wincol()		window column number of the cursor
 	winline()		window line number of the cursor
 	cursor()		position the cursor at a line/column
+	cursor_off()		disable cursor
+	cursor_on()		enable cursor
 	screencol()		get screen column of the cursor
 	screenrow()		get screen row of the cursor
 	screenpos()		screen row and col of a text character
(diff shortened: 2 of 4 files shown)


src/evalfunc.c
```diff
@@ -3862,11 +3868,33 @@ set_cursorpos(typval_T *argvars, typval_T *rettv, int charcol)
  * Returns 0 when the position could be set, -1 otherwise.
  */
     static void
-f_cursor(typval_T *argvars, typval_T *rettv)
+f_cursor(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
```
zeertzjq: ?

Shougo: It is my mistake...

runtime/doc/builtin.txt
```diff
@@ -1958,6 +1960,12 @@ cursor({list})
 		Can also be used as a |method|: >
 			GetCursorPos()->cursor()
 
+cursor_off()						*cursor_off()*
+		Disable cursor like |:sleep!|.
+
```
Shane-XB-Qian: i may suggest:
1. mentioned it do would recover cursor in 1 second anyhow, tho it may repeatly to be 'off'.
2. when 'off', perhaps the cmdline should indicate the status msg about the cursor was off.

Shane-XB-Qian (commented):
when 'off', perhaps the cmdline should indicate the status msg about the cursor was off.

runtime/doc/builtin.txt
chrisbra: Disable is a bit vague? I guess what is meant it *hide* cursor?

runtime/doc/builtin.txt
```diff
@@ -1958,6 +1960,12 @@ cursor({list})
...
+cursor_on()						*cursor_on()*
+		Enable cursor.
```
chrisbra, line 1967: *show* cursor again?

runtime/doc/usr_41.txt
```diff
@@ -906,6 +906,8 @@ Cursor and mark position:		*cursor-functions* *mark-functions*
 	wincol()		window column number of the cursor
 	winline()		window line number of the cursor
 	cursor()		position the cursor at a line/column
+	cursor_off()		disable cursor
+	cursor_on()		enable cursor
```
chrisbra: same here.

src/evalfunc.c
```diff
@@ -3867,6 +3873,28 @@ f_cursor(typval_T *argvars, typval_T *rettv)
...
 
+/*
+ * "cursor_off()" function
+ *
+ * Disable cursor.
+ */
+    static void
+f_cursor_off(typval_T *argvars UNUSED, typval_T *rettv UNUSED)
+{
+    cursor_sleep();
```
chrisbra, line 3884: does that work in the GUI?

chrisbra:
Hm, It seems in the GUI you can already do this using:

    :hi Cursor gui=NONE

However, hiding the cursor is completly confusing. You don't know where the cursor is currently, how would you know which one is the active window? How do you know, when editing the command line how many cells you need to move with the cursor? I don't really like this.

chrisbra:
oh and we would need tests of course.

Shane-XB-Qian:
> I don't really like this.

me too :smile:
if really required, I may wish cmdline msg or somewhere should indicate the cursor was off.

--
shane.xb.qian

wilhelmtell:
In addition to tests, can this PR also come with a use case?

habamax:
> In addition to tests, can this PR also come with a use case?

I could use this API here: https://github.com/vim/vim/discussions/14227

But I would prefer cursor not to pierce through popup window as an option of `popup_create()`.

src/evalfunc.c
Shougo: It works both CUI and GUI.

Shougo:
> Hm, It seems in the GUI you can already do this using:

Yes.   But it does not work in CUI.

NOTE:  neovim supports the feature though.

Shougo:
> oh and we would need tests of course.

Yes.  But I don't know what is the tests...

habamax:
I wonder what would happen here (some plugin that didn't properly handle possible exceptions):

```
def SomeFunc()
    cursor_off()

    # function throws exception
    AnotherFuncThatRaisesException()

    cursor_on()
enddef
```

And how easy for the user it would be to show cursor again?

Shougo:
> if really required, I may wish cmdline msg or somewhere should indicate the cursor was off.

I think it is plugin's work.  User should not call the API directly.

Shougo:
> Similar situation is with popup windows that might stay on the screen if plugin failed to close them, but there is way to close all of them with `:call popup_clear()`. Which I am not sure how easy it would be for `:call cursor_on()` if cursor is invisible. Should be fine I think?

If `:call popup_close()` is ok, `:call cursor_on()` is also ok.

Shane-XB-Qian:
1. yes, that's mainly was a concern of safety.
2. and I think hide the cursor by `highlight` was ugly too.
3. if your purpose was making the cursor was the only one when popup showup, then you maybe just make it be an option of popup creation.

--
shane.xb.qian

Shougo:
> if your purpose was making the cursor was the only one when popup showup, then you maybe just make it be an option of popup creation.

Hm.  It is ok to hide cursor when popup creation.

Shougo:
The new option `cursor` is false in `popup_create-arguments`, it is hide the cursor when the popup is visible.
The spec is OK?

Shane-XB-Qian:
when that option was set for that popup:
1. when that popup was end, then recover the cursor.
2. if popup was borken, it can be recovered by popup_clear by force.
3. the usage was limited only to when that popup get active and focus.

--
shane.xb.qian

Shane-XB-Qian:
> The default is true.

I think better not, some popup maybe stayed there forever or very long.

that should be just an option only for some one-time or current active popup.
--
shane.xb.qian

habamax:
What will happen if:

1. there is popup *A* that hides cursor
2. new popup *B* is created that also hides cursor is created (now there are 2 popups and no cursor is visible)
3. popup *B* is closed (and presumably cursor visibility is restored)

Will there be cursor visible along with popup *A*? Or it would be hidden as popup *A* was created with hidden cursor?

Shougo:
> Will there be cursor visible along with popup A? Or it would be hidden as popup A was created with hidden cursor?

It is stay hidden.  Because popup A is not closed.

Shougo:
> 3. the usage was limited only to when that popup get active and focus.

It may be better though I think Vim's popup feature does not have active/focus.

neovim's popup window has active/focus though.

Shane-XB-Qian:
> > 3. the usage was limited only to when that popup get active and focus.
> 
> It may be better though I think Vim's popup feature does not have active/focus.
> 
> neovim's popup window has active/focus though.

here `active` meant when that popup is eating the key typing, otherwise for a just ntf popup, i supposed user didnot like it steal the cursor.

Shane-XB-Qian:
anyway, here just a discussing, i am not very sure that's a right way, i still a little concern or worried the activities like 'hide' or 'disable' cursor.

chrisbra:
Sorry, I am not going to include this. I think this will cause too much irritation if the cursor unintentionally stays hidden.

Shougo:
> Sorry, I am not going to include this. I think this will cause too much irritation if the cursor unintentionally stays hidden.

Of course, there may be side effects depending on the plugin implementation, but that can happen with other features as well.

For example, the auto completion plugin may break your buffer.

The feature can solve the problem.

https://github.com/vim/vim/discussions/14227

Shougo:
I get the another solution.

PR #14438, opened 2024-04-07
runtime(doc): Normalise builtin-function optional parameter formatting

dkearns:
These should generally be formatted as `func([{arg}])` and referenced as `{arg}` in the description.


chrisbra:
thanks!

PR #14440, opened 2024-04-08
filetype: r history files cannot be recognized

Freed-Wu:
Problem: r history files cannot be recognized.
Solution: Add a pattern for r history files.


chrisbra:
thanks!

PR #14447, opened 2024-04-09
Don't compile abstract methods

### LINKED ISSUE #14443: Vim9: `E1027` with `defcompile` for abstract methods with non-`void` return types
### Steps to reproduce

These type hierarchies are fine:

```
vim9script

abstract class A1
    abstract def MethodA()
    abstract def MethodB()
endclass


abstract class A2
    abstract def MethodA(): void
    abstract def MethodB(): void
endclass


interface I3
    def MethodA()
    def MethodB()
endinterface

abstract class A3 implements I3
    var value: number

    abstract def MethodA()
    abstract def MethodB()
endclass

class C3 extends A3
    var condition: bool

    def MethodA()
    enddef

    def MethodB()
    enddef
endclass

const c3: C3 = C3.new()
c3.MethodA()
c3.MethodB()


interface I4
    def MethodA(): void
    def MethodB(): void
endinterface

abstract class A4 implements I4
    var value: number

    abstract def MethodA(): void
    abstract def MethodB(): void
endclass

class C4 extends A4
    var condition: bool

    def MethodA(): void
    enddef

    def MethodB(): void
    enddef
endclass

const c4: C4 = C4.new()
c4.MethodA()
c4.MethodB()

defcompile
```

These type hierarchies emit `E1027` with `:defcompile`:

```
vim9script

abstract class A5
    abstract def MethodA(): string
    abstract def MethodB(): string
endclass

# E1027
#defcompile
############################################################


interface I6
    def MethodA(): string
    def MethodB(): string
endinterface

abstract class A6 implements I6
    var value: number

    abstract def MethodA(): string
    abstract def MethodB(): string
endclass

# E1027
#defcompile
############################################################


abstract class A7
    var value: number

    abstract def MethodA(): string
    abstract def MethodB(): string
endclass

class C7 extends A7
    var condition: bool

    def MethodA(): string
	return "MethodA"
    enddef

    def MethodB(): string
	return "MethodB"
    enddef
endclass

const c7: C7 = C7.new()
echo c7.MethodA()
echo c7.MethodB()

# E1027
#defcompile
############################################################


interface I8
    def MethodA(): string
    def MethodB(): string
endinterface

abstract class A8 implements I8
    var value: number

    abstract def MethodA(): string
    abstract def MethodB(): string
endclass

class C8 extends A8
    var condition: bool

    def MethodA(): string
	return "MethodA"
    enddef

    def MethodB(): string
	return "MethodB"
    enddef
endclass

const c8: C8 = C8.new()
echo c8.MethodA()
echo c8.MethodB()

# E1027
defcompile
```

As can be seen, despite what `:defcompile` reports, objects  
are created and their methods are invoked just fine.

### Expected behaviour

`:defcompile` should fail _or_ succeed irrespective  
of what available types are used.

### Version of Vim

v9.1.0274

### Environment

### Logs and stack traces

_No response_

yegappan:
Fixes #14443.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -10583,4 +10583,46 @@ def Test_lambda_block_in_class()
   v9.CheckScriptSuccess(lines)
 enddef
 
+" Test for defcompiling an abstract method
+def Test_abstract_method_defcompile()
+  # Compile an abstract class with abstract object methods
+  var lines =<< trim END
+    vim9script
+    abstract class A
+      abstract def Foo(): string
+      abstract def Bar(): list<string>
+    endclass
+    defcompile
+  END
+  v9.CheckScriptSuccess(lines)
+
+  # Compile a concrete object method in an abstract class
+  lines =<< trim END
+    vim9script
+    abstract class A
+      abstract def Foo(): string
+      abstract def Bar(): list<string>
+      def Baz(): string
+        pass
+      enddef
+    endclass
+    defcompile
+  END
+  v9.CheckScriptFailure(lines, 'E476: Invalid command: pass', 1)
+
+  # Compile a concrete class method in an abstract class
+  lines =<< trim END
+    vim9script
+    abstract class A
+      abstract def Foo(): string
+      abstract def Bar(): list<string>
+      static def Baz(): string
+        pass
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #14448, opened 2024-04-09
filetype: pip configuration files cannot be recognized

Freed-Wu:
Problem: Pip configuration files cannot be recognized.
Solution: Add a pattern for pip configuration files.


chrisbra:
thanks!

PR #14449, opened 2024-04-09
Text height function does not respect it's argument

luukvbaal:
Problem:  plines_m_win() does not take into account it's "limit_winheight"
          argument for filler lines below the last line of the buffer.
Solution: Check window height when "limit_winheight" is TRUE.

### DIFF
--- a/src/misc1.c
+++ b/src/misc1.c
@@ -502,7 +502,7 @@ plines_m_win(win_T *wp, linenr_T first, linenr_T last, int limit_winheight)
 {
     int		count = 0;
 
-    while (first <= last)
+    while (first <= last && (!limit_winheight || count < wp->w_height))
     {
 #ifdef FEAT_FOLDING
 	int	x;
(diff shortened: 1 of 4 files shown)


src/misc1.c
```diff
@@ -519,17 +519,20 @@ plines_m_win(win_T *wp, linenr_T first, linenr_T last, int limit_winheight)
 #endif
 	{
 #ifdef FEAT_DIFF
-	    if (first == wp->w_buffer->b_ml.ml_line_count)
-		count += diff_check_fill(wp, first + 1);
 	    if (first == wp->w_topline)
-		count += plines_win_nofill(wp, first, limit_winheight)
-							       + wp->w_topfill;
+		count += plines_win_nofill(wp, first, FALSE) + wp->w_topfill;
```
luukvbaal, line 523: I don't think it makes sense to limit each plines call for winheight. Instead exit the loop early when "w_height" is reached and return it when "limit_winheight" is TRUE.

chrisbra: yes I think this makes sense. Thanks

chrisbra:
thanks!

PR #14450, opened 2024-04-09
filetype: lxqt configuration files cannot be recognized

Freed-Wu:
Problem: Lxqt configuration files cannot be recognized.
Solution: Add a pattern for lxqt configuration files.


chrisbra:
thanks!

PR #14451, opened 2024-04-09
filetype: XDG mimeapps.list files cannot be recognized

Freed-Wu:
Problem: XDG mimeapps.list files cannot be recognized
Solution: Add a pattern for
Refer: https://wiki.archlinux.org/title/XDG_MIME_Applications#Format


chrisbra:
thanks!

PR #14452, opened 2024-04-09
filetype: some json files cannot be recognized

Freed-Wu:
Problem: Some json files cannot be recognized.
Solution: Add a pattern for some json files

Refer: https://github.com/microsoft/PowerToys/blob/main/.vsconfig
Refer: https://jscs-dev.github.io/


chrisbra:
thanks. Can you please add a filetype test?

Freed-Wu:
Fixed.

PR #14455, opened 2024-04-09
Still problems with cursor position for CTRL-D/U

luukvbaal:
Problem:  Problems with cursor position when scrolling half a page.
Solution: Rework the cursor logic.

### DIFF
--- a/src/testdir/test_normal.vim
+++ b/src/testdir/test_normal.vim
@@ -4225,4 +4225,16 @@ func Test_single_line_filler_zb()
   bw!
 endfunc
 
+" Test for Ctrl-U not getting stuck at end of buffer with 'scrolloff'.
+func Test_halfpage_scrolloff_eob()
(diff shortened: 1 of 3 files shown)


src/testdir/test_normal.vim
```diff
@@ -4225,4 +4225,16 @@ func Test_single_line_filler_zb()
   bw!
 endfunc
 
+" Test for Ctrl-D not getting stuck at end of buffer with 'scrolloff'.
```
zeertzjq:
```suggestion
" Test for Ctrl-U not getting stuck at end of buffer with 'scrolloff'.
```

luukvbaal:
Simply restoring the previous cursor position before moving the cursor "count" lines for half-page scrolling seems to work as expected now(it didn't before...). This ignores some of the work done in scrollup/down() but special casing "half" in the inner functions will get messy.

PR #14456, opened 2024-04-09
filetype: some TeX files are not recognized

Freed-Wu:
Problem: Some TeX files are not recognized
Solution: Add patterns for some TeX files


runtime/filetype.vim
```diff
@@ -305,8 +305,9 @@ au BufNewFile,BufRead *.csdl			setf csdl
 " Cabal
 au BufNewFile,BufRead *.cabal			setf cabal
 
-" Cdrdao TOC
-au BufNewFile,BufRead *.toc			setf cdrtoc
+" Cdrdao TOC or LaTeX \tableofcontents files
+au BufNewFile,BufRead *.toc
+	\ if getline(1) =~ '\contentsline' |setf tex|else|setf cdrtoc|endif
```
zeertzjq:
```suggestion
	\ if getline(1) =~# '\\contentsline' |setf tex|else|setf cdrtoc|endif
```

runtime/filetype.vim
```diff
@@ -1258,7 +1259,11 @@ au BufNewFile,BufRead */etc/login.defs		setf logindefs
 au BufNewFile,BufRead *.lgt			setf logtalk
 
 " LOTOS
-au BufNewFile,BufRead *.lot,*.lotos		setf lotos
+au BufNewFile,BufRead *.lotos		setf lotos
+
+" LOTOS or LaTeX \listoftables files
+au BufNewFile,BufRead *.lot
+	\ if getline(1) =~ '\contentsline' |setf tex|else|setf lotos|endif
```
zeertzjq:
```suggestion
	\ if getline(1) =~# '\\contentsline' |setf tex|else|setf lotos|endif
```

Freed-Wu: Right:

```tex
\contentsline {table}{\numberline {1}{\ignorespaces caption}}{1}{table.1}%
```

chrisbra:
thanks!

PR #14457, opened 2024-04-09
Fix Windows warning for size_t to int conversion

mrdubya:
Another Windows size_t to int conversion warning was resolved by changing a declaration.


chrisbra:
thanks for fixing the Windows builds!

PR #14458, opened 2024-04-09
runtime(cgdbrc): add support

Freed-Wu:


runtime/syntax/cgdbrc.vim
```diff
@@ -0,0 +1,21 @@
...
+" Documentation:        https://cgdb.github.io/docs/Configuring-CGDB.html
+" Latest Revision:      2024-04-09
+
+if exists('b:current_syntax')
+  finish
+endif
+let b:current_syntax = 'cgdbrc'
+
+" Read the Vim syntax to start with
+if version < 600
```
chrisbra: I think we do not need the version test for vim < 6.0 :)

Freed-Wu: Fixed. It is copied from https://github.com/Freed-Wu/vim/blob/master/runtime/syntax/arduino.vim#L25

runtime/ftplugin/cgdbrc.vim
```diff
@@ -0,0 +1,20 @@
...
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+let b:undo_ftplugin = 'setl <cms'
+
+setlocal commentstring=#%s
```
dkearns, line 17: Can you please set `'comments'` as well?

dkearns (changes requested):

chrisbra:
thanks!

PR #14459, opened 2024-04-09
compiler: add vimdoc

Freed-Wu:
https://github.com/google/vimdoc generates vim help files from vimscript files

### DIFF
--- a/runtime/compiler/vimdoc.vim
+++ b/runtime/compiler/vimdoc.vim
@@ -1,0 +1,19 @@
+" Vim Compiler File
+" Language:             vimdoc
+" Maintainer:           Wu, Zhenyu <wuzhenyu@ustc.edu>
+" Latest Revision:      2024-04-09
+"
+" you can get it by `pip install vimdoc` or the package manager of your distribution.
+
+if exists('b:current_compiler')
+  finish
+endif
+let b:current_compiler = 'vimdoc'
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+CompilerSet makeprg=vimdoc
+
+let &cpoptions = s:save_cpoptions
+unlet s:save_cpoptions
--- a/runtime/ftplugin/vim.vim
+++ b/runtime/ftplugin/vim.vim
@@ -15,6 +15,8 @@ let b:did_ftplugin = 1
 let s:cpo_save = &cpo
 set cpo&vim
 
+compiler vimdoc
+
 if !exists('*VimFtpluginUndo')
   func VimFtpluginUndo()
     setl fo< isk< com< tw< commentstring< include< define<


runtime/compiler/vimdoc.vim
```diff
@@ -0,0 +1,20 @@
...
+
+if exists('b:current_compiler')
+  finish
+endif
+let b:current_compiler = 'vimdoc'
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+if exists(':CompilerSet') != 2
```
zeertzjq: This should be removed after #14399

Konfekt:
This looks useful. Maybe also https://github.com/FooSoft/md2vim for those that write a README in (Github) markdown and then add a Vim help file as an afterthought?

runtime/compiler/vimdoc.vim
```diff
@@ -0,0 +1,17 @@
...
+
+if exists('b:current_compiler')
+  finish
+endif
+let b:current_compiler = 'vimdoc'
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+CompilerSet makeprg=vimdoc\ .
```
chrisbra: Can you add a comment where to get vimdoc?

chrisbra: > vimdoc\ .

I suppose this is should run vimdoc with the currently edited directory? `.` might be different for that however because of `:cd/:tcd` etc. 

Also, does this work correctly, if there isn't a `doc/` directory?

Freed-Wu: > Can you add a comment where to get vimdoc?

Fixed.

Freed-Wu: > Also, does this work correctly, if there isn't a doc/ directory?

It will `mkdir doc`.

> run vimdoc with the currently edited directory

In my case, if my `$PWD` is `/the/directory/of/a/vim/plugin` and I am editing `/the/directory/of/a/vim/plugin/autoload/my_plugin/my_functions.vim` or `/the/directory/of/a/vim/plugin/plugin/my_plugin.vim`, Run `vimdoc .` will work.

Freed-Wu: More details about vimdoc you can ask the maintainer of vimdoc @dbarnett

chrisbra: Yeah, but what if it is  `/tmp` ? I guess `:makeprg` should rather be `makeprg=vimdoc\ %:h:S` (similar to what other compiler plugins set).

Freed-Wu: >  makeprg=vimdoc\ %:h:S

Oh, we can search `addon.json` to know the path `XXX` of `vimdoc XXX`

https://github.com/google/vim-maktaba/tree/HEAD/examples/helloworld

dkearns: We were, generally, trying to move away from compiler plugins that specify the target in `makeprg` because being able to specify it in the call to `:make` is more flexible.

chrisbra: not sure what `addon.json` is supposed to be doing and the URL doesn't explain it. I have never seen this with vimscripts so if we start searching for that file, it probably makes the compiler not useful anymore...

Freed-Wu: https://github.com/google/vim-maktaba/blob/fe95bb10f6bb250943a44632107f6a3d76ce5f28/examples/helloworld/addon-info.json

Sorry, should be addon-info.json

Freed-Wu: I think upstream should have a default path of `vimdoc PATH` :smile:

https://github.com/google/vimdoc/issues/120

Freed-Wu: > probably makes the compiler not useful anymore

`vimdoc XXX` must know the root directory `XXX` of the vim plugin. For `/the/path/autoload/a/b.vim`, it is `%:p:h:h:h`. For `/the/path/ftplugin/a.vim`, it is `%:p:h:h`. `vimdoc` expect a `addon-info.json` in `/the/path` contain the information of this vim plugin, such as name, version, ...

Usually `XXX` is `.`, so `vimdoc .` usually work. We can also search `addon-info.json` and use `.` as a fallback.

dbarnett: I'd expect the target path would be a vim plugin dir, and implicitly defaulting to CWD feels a little weird to me but I don't know what conventions are normal for vim compiler hooks.

I'm not an authority in what changes should get merged into vim runfiles though. @malcolmr FYI

chrisbra: so let's just remove the directory from the :make command as @dkearns mentions. Let's not overcomplicate things then.

Freed-Wu: Fixed.

Shane-XB-Qian:
i am curious to ask anyone really ever try this vimdoc, even regardless this 'compiler' which just the cmd vimdoc
// no offense if you were the author, but seems that project had been Not updated over 2 years;

i tried a bit, but seems would Not generate some useful doc,
besides now vim9 script it still recognized?
// again, no offense, but it maybe need to confirm it was useful before putting it into official vim code base...

-- 
shane.xb.qian

Freed-Wu:
> i am curious to ask anyone really ever try this vimdoc

I use it on my vim plugin.

> now vim9 script it still recognized?

I think the answer is no. @dbarnett Am I right?

Shane-XB-Qian:
> > i am curious to ask anyone really ever try this vimdoc
> 
> I use it on my vim plugin.

have you ever try any some others popular plugins?
what's that doc looked like?
and it mentioned 'pip install' or on its distr, but seems it only worked build from its source code

-- 
shane.xb.qian

Freed-Wu:
> and it mentioned 'pip install' or on its distr, but seems it only worked build from its source code

You can install it by [`pre-commit`](https://pre-commit.com/) or [AUR](https://aur.archlinux.org/packages/vimdoc).

> have you ever try any some others popular plugins?

You said md2vim? I haven't use it.

> what's that doc looked like?

Official example:

https://github.com/google/vim-maktaba/blob/fe95bb10f6bb250943a44632107f6a3d76ce5f28/examples/helloworld/doc/helloworld.txt

Shane-XB-Qian:
> > and it mentioned 'pip install' or on its distr, but seems it only worked build from its source code
> You can install it by [`pre-commit`](https://pre-commit.com/) or [AUR](https://aur.archlinux.org/packages/vimdoc).

it's not in 'pip' store, and not in most others distr regular stores i guess so.

> > have you ever try any some others popular plugins?
> You said md2vim? I haven't use it.

> > what's that doc looked like?
> Official example:
> https://github.com/google/vim-maktaba/blob/fe95bb10f6bb250943a44632107f6a3d76ce5f28/examples/helloworld/doc/helloworld.txt

no, i think it's been out-date long time, not even saying vim9 script, so this commit perhaps just some burden vs/but not useful.

-- 
shane.xb.qian

Shane-XB-Qian:
um, let me reword my expression, i meant:
it's ok to use it as an extra cmd when you dev a vim script project if you like and as its spec, and if you were able to get that cmd (by self-build or somehow), 
but as if it's going to be treat as a `compiler` of vim, i supposed it perhaps not much useful, seems not easy/direct usage, specially as if to be 'default' compiler of vim ft.

Shane-XB-Qian:
> The general policy has always been that _any_ submitted compiler plugin would be added.

donot understand what you mean, but 'vimdoc' now it is the _default_ compiler of vim ft.

-- 
shane.xb.qian

chrisbra:
> donot understand what you mean, but 'vimdoc' now it is the _default_ compiler of vim ft.

Yes and where before you would get 
```
make: *** No targets specified and no makefile found.  Stop.

```

you will now get:
```
bash: vimdoc: command not found

```

I fail to see why this is a problem now?

runtime/ftplugin/vim.vim
```diff
@@ -15,6 +15,8 @@ let b:did_ftplugin = 1
 let s:cpo_save = &cpo
 set cpo&vim
 
+compiler vimdoc
+
```
Shane-XB-Qian, line 19: remove this if really like to keep vimdoc as a compiler.

runtime/compiler/vimdoc.vim
```diff
@@ -0,0 +1,19 @@
...
+
+if exists('b:current_compiler')
+  finish
+endif
+let b:current_compiler = 'vimdoc'
+
+let s:save_cpoptions = &cpoptions
+set cpoptions&vim
+
+CompilerSet makeprg=vimdoc
```
Shane-XB-Qian, line 16: this `vimdoc` mostly would not be easy getting from most distr, and it req a plugin name as its arg, this is a problem.

Shane-XB-Qian:
> I'd probably expect a linter to be the default compiler for the vim filetype, if there was going to be one. It's rare to specify a compiler in an ftplugin.

so far, i think there probably no good option ones, supposed no one good fit vim9 script, so perhaps just back out.
// meanwhile, recently tens of runtime update, maybe need more care about the 'usability' before putting into vim code base.

runtime/ftplugin/vim.vim
Freed-Wu: OK, we can use linter.

runtime/compiler/vimdoc.vim
Freed-Wu: Yes, I also think upstream should add a default path as its arg.

Freed-Wu:
This is vint:

```sh
$ vint runtime/bugreport.vim
"runtime/bugreport.vim:20:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:41:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:47:47: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:47:55: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:48:38: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:49:38: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:50:38: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:53:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:60:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:64:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:68:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:72:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
"runtime/bugreport.vim:76:9: warning! " "Prefer single quoted strings(Google VimScript Style Guide (Strings))"
```

PR #14461, opened 2024-04-09
ftplugin(asm): add Matchit support

Freed-Wu:
Refer https://github.com/vim/vim/blob/master/runtime/ftplugin/masm.vim#L18-L29

### DIFF
--- a/runtime/ftplugin/asm.vim
+++ b/runtime/ftplugin/asm.vim
@@ -3,11 +3,21 @@
 " Maintainer:	Colin Caine <cmcaine at the common googlemail domain>
 " Last Change:  23 May 2020
 " 		2023 Aug 28 by Vim Project (undo_ftplugin)
+" 		2024 Apr 09 by Vim Project (add Matchit support)
 
 if exists("b:did_ftplugin") | finish | endif
 let b:did_ftplugin = 1
 
+setl include=^\\s*%\\s*include
 setl comments=:;,s1:/*,mb:*,ex:*/,://
 setl commentstring=;%s
 
-let b:undo_ftplugin = "setl commentstring< comments<"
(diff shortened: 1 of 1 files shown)


runtime/ftplugin/asm.vim
```diff
@@ -3,11 +3,21 @@
 " Maintainer:	Colin Caine <cmcaine at the common googlemail domain>
 " Last Change:  23 May 2020
 " 		2023 Aug 28 by Vim Project (undo_ftplugin)
+" 		2024-04-09
```
chrisbra: by Vim Project (add matchit support)

Freed-Wu: Fixed

runtime/ftplugin/asm.vim
```diff
@@ -3,11 +3,21 @@
 " Maintainer:	Colin Caine <cmcaine at the common googlemail domain>
 " Last Change:  23 May 2020
 " 		2023 Aug 28 by Vim Project (undo_ftplugin)
+" 		2024-04-09 by Vim Project (add Matchit support)
```
dkearns: And then there were three date formats.

Can we please use YYYY Mon DD for the "Vim Project" lines?

runtime/ftplugin/asm.vim
```diff
@@ -3,11 +3,21 @@
...
 
 if exists("b:did_ftplugin") | finish | endif
 let b:did_ftplugin = 1
 
+setl include=^\s*%\s*include
```
dkearns:
```suggestion
setl include=^\\s*%\\s*include
```

dkearns (changes requested):

PR #14464, opened 2024-04-09
refactor(argedit): filter duplicate item and insert in correct index

glepnir:
Problem: Currently, the argedit command does not work correctly. According to the help documentation, when a filename already exists in the argument list, that entry should be edited. However, argedit does not check for duplicates and instead inserts a new one. Additionally, it does not function like argadd. For instance, in the test case Test_argedit, after executing `argedit b` and `argedit a`, where the current file is a, then executing `argedit C D` should insert C and D after a. However, the expected result is a b C D, similar to the behavior observed with `argadd` where the current file is b: executing `:argadd x` results in the argument list `a b x c`.

Solution: Refactor argedit to avoid inserting duplicate items and to insert new items at the correct position.

(And fname_case is removed can remove test case for it ? )

### DIFF
--- a/src/arglist.c
+++ b/src/arglist.c
@@ -434,6 +434,25 @@ arglist_del_files(garray_T *alist_ga)
 }
 
 /*
+ * Expand the file names in the argument list based on wildcards.
+ */
+    static int
+expand_arglist_files(garray_T *ga, int *count, char_u ***files)
+{
+    int result;
+    result = expand_wildcards(ga->ga_len, ga->ga_data, count, files,
+	    EW_DIR|EW_FILE|EW_ADDSLASH|EW_NOTFOUND);
+    ga_clear(ga);
+
+    if (result == FAIL || count == 0)
+    {
+	emsg(_(e_no_match));
+	return FAIL;
+    }
+    return OK;
+}
+
+/*
  * "what" == AL_SET: Redefine the argument list to 'str'.
  * "what" == AL_ADD: add files in 'str' to the argument list after "after".
  * "what" == AL_DEL: remove files in 'str' from the argument list.
@@ -450,7 +469,6 @@ do_arglist(
     garray_T	new_ga;
     int		exp_count;
     char_u	**exp_files;
-    int		i;
     int		arg_escaped = TRUE;
 
     if (check_arglist_locked() == FAIL)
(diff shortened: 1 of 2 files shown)


glepnir:
aha I didn't look closely at the history. Repeating elements is just one of bug The insertion order of argedit is not as described in help. It does not have the same result as argadd. Not considered backwards compatible. Is it possible to change to the new arg command?

glepnir:
@chrisbra  Hello any thought on this ? Whether vim accepts this change or adds it to vim as a new command ?

chrisbra:
I mentioned already that removing duplicates would break existing scripts and was the reason for the `:argdedupe` command. So that would leave the refactoring? Or what is then the benefit of this change?

glepnir:
original issue https://github.com/neovim/neovim/issues/11705 and I agree with some of his points. The current way of working is different from the work of the argedit help documentation. Secondly, the current insertion order of argedit is also different from argadd.  I had mentioned in above problem section. 

```
argedit a 
argedit b
argedit a
argedit C D
echo argv()  ==> a, b C, D
```
document of argadd. b is current  add x  result is a b x .   argedit document say `This is like using |:argadd| and then |:edit|.`
```
			If the argument list is "a b c", and "b" is the
			current argument, then these commands result in:
				command		new argument list ~
				:argadd x	a b x c
```

chrisbra:
> echo argv()  ==> a, b C, D

For me this is: `['a', 'b', 'a', 'C', 'D']`

> document of argadd. b is current add x result is a b x . argedit document say This is like using |:argadd| and then |:edit|.

			If the argument list is "a b c", and "b" is the
			current argument, then these commands result in:
				command		new argument list ~
				:argadd x	a b x c

This seems almost correct. If I run `vim --clean a b c` -c ':next' 
- If I type `:argadd x` result will be: `a   [b] x   c`
- If I type `:argedit x` result will be: `a   b   [x] c`


I update the documentation.

glepnir:
> For me this is: ['a', 'b', 'a', 'C', 'D']

yup because duplicate items.   if exist entry will be edit it should be ` a C D b` .   if you think duplicate items make sense in there just close .

chrisbra:
yes, that's what the documentation mentions and why one should use the `:argdedupe` command. I'll close it then. Thanks anyhow!

zeertzjq:
`:argdedupe` is only mentioned in the documentation of `:argadd`. The documentation of `:argedit` still mentions the behavior before 8.0.0721
```
:[count]arge[dit][!] [++opt] [+cmd] {name} ..		*:arge* *:argedit*
			Add {name}s to the argument list and edit it.
			When {name} already exists in the argument list, this
			entry is edited.
```

chrisbra:
Thanks, I changed the documentation

Shane-XB-Qian:
> `:argdedupe`

an irrelevant comment: why it was `argdedupe` but not `argdeldup` or `argdedup`, that last letter `e` seems confusing?

-- 
shane.xb.qian

PR #14466, opened 2024-04-09
runtime(vim): Update base-syntax, no curly-brace names in Vim9 script

dkearns:
Remove curly-brace name matching for `:def` functions.  This is not supported in Vim9 script.


chrisbra:
thanks!

PR #14467, opened 2024-04-09
fix inaccuracies in pandoc compiler

Konfekt:
as kindly pointed out by @Freed-Wu the definition of `:CompilerSet` has been abandoned, but at some point slipped in


chrisbra:
thanks!

PR #14470, opened 2024-04-09
Missing test for patch 9.1.0285

luukvbaal:
Problem:  Not all of what was fixed by patch 9.1.0285 is tested.
Solution: Add a test for cursor movement at buffer boundaries.


chrisbra:
Thanks!

PR #14472, opened 2024-04-10
Vim9: heredoc start may be recognized in string

zeertzjq:
Problem:  Vim9: heredoc start may be recognized in string.
Solution: Don't skip to closing bracket for invalid list assignment.


chrisbra:
thanks!

PR #14473, opened 2024-04-10
Return type not set for a lambda assigned to a script variable

### LINKED ISSUE #14445: [vim9script] script level loses (or doesn't preserve) some `func` type information
### Steps to reproduce

**NOTE:** https://github.com/vim/vim/issues/14445#issuecomment-2045597829 is a simpler example of the issue, no `class`, no global `dict`.

```
vim9script

class C
    def CF0(a: string): string
        return printf("xx'%s'xx", a)
    enddef
endclass

def WrapInvocation(): func
    var x: list<string>

    var Q = () => {
         return C.new().CF0("FOO")
        }
    echo '(1)' typename(Q)                              ### func(): string
    #return Q

    x =<< trim [CODE]
        g:SomeRandomFunction = () => {
            return C.new().CF0("FOO")
            }
    [CODE]
    execute(x)
    echo '(2)' typename(g:SomeRandomFunction)           ### func()

    var U = g:SomeRandomFunction    
    unlet g:SomeRandomFunction

    return U
enddef
WrapInvocation()
```
The type of the lambda, as indicated by `typename()`, is different depending on how it is created.
```
(1) func(): string
(2) func()
```
This is a problem because of how the return can be used. When the return is actually correctly type, it gets an error. Note that executing `echo F0() F1() F2()` in the following works.
```
var F0 = WrapInvocation()                           ### OK
var F1: func = WrapInvocation()                     ### OK
var F2: func() = WrapInvocation()                   ### OK
var F3: func(): string = WrapInvocation()           ### ERROR
```

```
E1012: Type mismatch; expected func(): string but got func()
```
BTW, typing `def WrapInvocation(): any` fails in the same way. So the type information that is causing the problem is not because of the return type of the function, it is embedded in the returned value.

### Expected behaviour

No error

This shows how it's actually used, along with a workaround. The workaround allows lambda return type information to be supplied so it does not depend on inference to get the return type.
```
vim9script

var XXX: any
def WrapInvocation(fcall: string, signature: string = '()'): any
    var x =<< trim eval [CODE]
        XXX = {signature} => {{
            {fcall}
            }}
    [CODE]

    execute(x)
    return XXX
enddef

def F()
    var X: func(): string = WrapInvocation('return "FOO"', '(): string')
    echo X()
    X = WrapInvocation('return "FOO"')
    echo X()
enddef
F()
```
The first call to `WrapInvocation()` works, the 2nd fails
```
FOO
Error detected while processing :source buffer=1[21]..function <SNR>36_F:
line    3:
E1012: Type mismatch; expected func(): string but got func()
```

### Version of Vim

9.1.281

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

yegappan:
Fixes #14445.

Also refactor the code slightly.

### DIFF
--- a/src/testdir/test_vim9_assign.vim
+++ b/src/testdir/test_vim9_assign.vim
@@ -1104,6 +1104,27 @@ def Test_assignment_partial()
       Ref(0)
   END
   v9.CheckScriptFailure(lines, 'E1013: Argument 2: type mismatch, expected string but got number')
+
+  lines =<< trim END
+    var Fn1 = () => {
+        return 10
+      }
+    assert_equal('func(): number', typename(Fn1))
+    var Fn2 = () => {
+        return "a"
+      }
+    assert_equal('func(): string', typename(Fn2))
+    var Fn3 = () => {
+        return {a: [1]}
+      }
+    assert_equal('func(): dict<list<number>>', typename(Fn3))
+    var Fn4 = (...l: list<string>) => {
+        return []
+      }
+    assert_equal('func(...list<string>): list<any>', typename(Fn4))
+  END
+  v9.CheckSourceSuccess(['vim9script'] + lines)
+  v9.CheckSourceSuccess(['def Xfunc()'] + lines + ['enddef', 'defcompile'])
 enddef
 
 def Test_assignment_list_any_index()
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #14476, opened 2024-04-10
runtime: add PicoLisp filetype

nat-418:
This PR adds support for the [PicoLisp](https://picolisp.com/wiki/?home) programming language. ~~PicoLisp exclusively uses the `*.l` filename extension, which is currently also used by Lex among several other extensions, so I have removed `*.l` from the Lex filename extensions and transferred it to PicoLisp.~~

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1191,6 +1191,28 @@ export def FTdat()
   endif
 enddef
 
+# Determne if a *.l file is Lex or PicoLisp
+export def FTl()
+  " Default to Lex
+  set ft=lex
+  for i in range(1, line("$"))
+    let l:line = trim(getline(i))
+    " PicoLisp uses # comments
+    if l:line =~ "^#\.*$"
+      set ft=picolisp
+      return
+    " Lex uses // comments
+    elseif (l:line =~ "//" && l:line !~ "([^)]*//") || l:line =~ "/\\*" || l:line =~ "\\*/"
+      set ft=lex
+      return
+    " PicoLisp uses Lisp syntax
+    elseif l:line =~ "^("
+      set ft=picolisp
+      return
+    endif
+  endfor
+enddef
+
 export def FTlsl()
   if exists("g:filetype_lsl")
     exe "setf " .. g:filetype_lsl
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1191,27 +1191,31 @@ export def FTdat()
   endif
 enddef
 
-# Determine if a *.l file is Lex or PicoLisp
+# Determine if a *.l file is Lex or PicoLisp (default to Lex)
 export def FTl()
-  " Default to Lex
-  setf lex
+  if exists("g:filetype_l")
+    exe "setf " .. g:filetype_l
+    return
+  endif
   for i in range(1, line("$"))
-    let l:line = trim(getline(i))
-    " PicoLisp uses # comments
-    if l:line =~ "^#\.*$"
-      set ft=picolisp
+    var line = trim(getline(i))
+    if line =~ "^%{.*$"
+      setf lex
       return
-    " Lex uses // comments
-    elseif (l:line =~ "//" && l:line !~ "([^)]*//") || l:line =~ "/\\*" || l:line =~ "\\*/"
-      set ft=lex
+    elseif line =~ "^%%.*$"
+      setf lex
       return
-    " PicoLisp uses Lisp syntax
-    elseif l:line =~ "^("
-      set ft=picolisp
+    elseif line =~ "^#.*$"
+      setf picolisp
+      return
+    elseif line =~ "^\(.*$"
+      setf picolisp
       return
     endif
   endfor
-enddef
+  setf lex
+  return
+enddef 
 
 export def FTlsl()
   if exists("g:filetype_lsl")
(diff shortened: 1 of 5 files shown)


runtime/filetype.vim
```diff
@@ -1199,7 +1199,7 @@ au BufRead,BufNewFile *.ldg,*.ledger,*.journal			setf ledger
 au BufNewFile,BufRead *.less			setf less
 
 " Lex
-au BufNewFile,BufRead *.lex,*.l,*.lxx,*.l++	setf lex
+au BufNewFile,BufRead *.lex,*.lxx,*.l++		setf lex
```
dkearns, line 1202: You can't just steal the `*.l` extension for the new filetype, unless you're willing to fight off a horde of pitch-fork wielding Lex users.

You need to add a function to https://github.com/vim/vim/tree/master/runtime/autoload/dist/ft.vim that can determine the file type based on content heuristics and call that from the autocmd in filetype.vim.

dkearns (changes requested):

runtime/autoload/dist/ft.vim
```diff
@@ -1156,13 +1156,13 @@ export def FTfoam()
...
 
-# Determine if a *.tf file is TF mud client or terraform
-export def FTtf()
-  var numberOfLines = line('$')
-  for i in range(1, numberOfLines)
-    var currentLine = trim(getline(i))
-    var firstCharacter = currentLine[0]
-    if firstCharacter !=? ";" && firstCharacter !=? "/" && firstCharacter !=? ""
+# determine if a *.tf file is tf mud client or terraform
+export def fttf()
```
zeertzjq: This function name should not be changed
```suggestion
# Determine if a *.tf file is TF mud client or terraform
export def FTtf()
```

runtime/filetype.vim
nat-418: I confess that I have never written vim9script before and I have forgotten what vimscript I once did know, so if the function I have added seems poor to you I apologize in advance. Essentially, I am trying to express the notion that, as a Lisp, a PicoLisp file should always begin with a leading `(` somewhere (excluding shebangs, etc.).

runtime/autoload/dist/ft.vim
```diff
@@ -1191,6 +1191,20 @@ export def FTdat()
   endif
 enddef
 
+# Determne if a *.l file is Lex or PicoLisp
+export def FTl()
+  var numberoflines = line('$')
+  for i in range(1, numberoflines)
+    var currentline = trim(getline(i))
+    var firstcharacter = currentline[0]
+    if firstcharacter !=? "("
```
zeertzjq: `!=?` is inequality. `==?` is equality.
```suggestion
    if firstcharacter ==? "("
```

runtime/autoload/dist/ft.vim
nat-418: this was a mistake on my part, and I offer my apologies

runtime/autoload/dist/ft.vim
```diff
@@ -1191,6 +1191,28 @@ export def FTdat()
   endif
 enddef
 
+# Determne if a *.l file is Lex or PicoLisp
```
dkearns:
```suggestion
# Determine if a *.l file is Lex or PicoLisp
```

runtime/autoload/dist/ft.vim
```diff
@@ -1191,6 +1191,28 @@ export def FTdat()
...
+export def FTl()
+  " Default to Lex
+  set ft=lex
```
dkearns:
```suggestion
  setf lex
```

All `set ft=` lines need to use `setf`.

runtime/autoload/dist/ft.vim
```diff
@@ -1191,6 +1191,28 @@ export def FTdat()
   endif
 enddef
 
+# Determne if a *.l file is Lex or PicoLisp
+export def FTl()
```
dkearns, line 1195: There needs to be an option to override this behaviour with a global variable `g:filetype_l` like the `FTlsl` example.

This also needs to be documented at `:help filetype-overrule`.

dkearns (changes requested):

nat-418:
Thanks to @dkearns  and @zeertzjq for guiding me through this PR.

chrisbra:
thanks, but tests still seem to fail.

chrisbra:
test still fails :/

src/testdir/test_filetype.vim
```diff
@@ -528,6 +528,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     pf: ['pf.conf'],
     pfmain: ['main.cf', 'main.cf.proto'],
     php: ['file.php', 'file.php9', 'file.phtml', 'file.ctp', 'file.phpt', 'file.theme'],
+    picolisp: ['file.l'],
```
```suggestion
```

runtime/autoload/dist/ft.vim
```diff
@@ -1191,6 +1191,32 @@ export def FTdat()
   endif
 enddef
 
+# Determine if a *.l file is Lex or PicoLisp (default to Lex)
+export def FTl()
+  if exists("g:filetype_l")
+    exe "setf " .. g:filetype_l
+    return
+  endif
+  for i in range(1, line("$"))
```
zeertzjq, line 1200: Many filetype detection functions have a limit to the number of lines to check, like using `range(1, min([line("$"), 100]))`

chrisbra:
this has stalled, so closing

PR #14478, opened 2024-04-10
none: ignore runtime/doc/tags-??

Shane-XB-Qian:
e.g `tags-ru` was generating each time after make install, supposed it's not required since it is not pushed to git too.


k-takata:
Could you also update `.hgignore` in the same way?

Shane-XB-Qian:
> Could you also update `.hgignore` in the same way?

ok, opened at #14479

Shane-XB-Qian:
> Normally, you don't need to create another PR, though.
> Just pushing a new commit to your branch should be enough.

it was created at 'master' which i had trimmed it right after, so created a new branch.

-- 
shane.xb.qian

PR #14312, opened 2024-03-27
translation(ru): The main file README.txt and uganda.txt

RestorerZ:
Added "README.rux.txt" and "uganda.rux" files to be able to include them in the
Vim installer.

PR #14479, opened 2024-04-10
none: ignore runtime/doc/tags-??

Shane-XB-Qian:
e.g tags-ru was generating each time after make install, supposed it's not required since it is not pushed to git too.
// sry dup, i forgot created a branch, #14478


zzzyxwvut:
I am wondering about [the same](https://github.com/vim/vim/pull/14312#issuecomment-2041603791).

chrisbra:
thanks!

Shane-XB-Qian:
@chrisbra my patch was `tags-??`, but looks you modified it a bit to `tags-*`, not sure if had outage, but Makefile was `tags-??`, so you may need to modify it for them in Makefile all.

chrisbra:
I replied to your mail, but stupid groups decide to block my mails, because I am spamming the list :) 

I think it doesn't matter, I liked the `tags-*` slightly better :) 

Thanks both

mralusw:
This happened ever since 2ca7d5f and it seems to be due to `runtime/doc/uganda.rux`, which is the only translated `uganda`.

`./runtime/doc/tags-ru` seems to be solely tags from `uganda.rux`. Maybe this should be a separate bug, I'm not sure.

Shane-XB-Qian:
@mralusw that is a tag file, nothing else, this PR is not to del it, donot get it wrong.

RestorerZ:
> This happened ever since [2ca7d5f](https://github.com/vim/vim/commit/2ca7d5f4831ce3d86cdc3f0c844fc50ea0deb7e5) and it seems to be due to `runtime/doc/uganda.rux`, which is the only translated `uganda`.
> 
> `./runtime/doc/tags-ru` seems to be solely tags from `uganda.rux`. Maybe this should be a separate bug, I'm not sure.

To avoid creating "tags-??", you can probably use the "tags" target instead of "vimtags".
Or exclude "tags-??" from the list of "files" files.

https://github.com/vim/vim/blob/4b5cd7257ee99384940d5210cf50298ff925924e/src/Makefile#L2347C1-L2365C71
```
# Generate the help tags with ":helptags" to handle all languages.
# Move the distributed tags file aside and restore it, to avoid it being
# different from the repository.
        cd $(HELPSOURCE); if test -z "$(CROSS_COMPILING)" -a -f tags; then \
                mv -f tags tags.dist; fi
        @echo generating help tags
        cd $(HELPSOURCE); \
                files=`ls *.txt tags`; \
                $(INSTALL_DATA) $$files  $(DEST_HELP); \
                cd $(DEST_HELP); \
                chmod $(HELPMOD) $$files
        $(INSTALL_DATA)  $(HELPSOURCE)/*.pl $(DEST_HELP)
        chmod $(SCRIPTMOD) $(DEST_HELP)/*.pl
        cd $(HELPSOURCE); if test -f tags.dist; then mv -f tags.dist tags; fi
```

https://github.com/vim/vim/blob/4b5cd7257ee99384940d5210cf50298ff925924e/runtime/doc/Makefile#L25C1-L37C1

```
# Use Vim to generate the tags file.  Can only be used when Vim has been
# compiled and installed.  Supports multiple languages.
vimtags: $(DOCS)
        @$(VIMEXE) --clean -esX -V1 -u doctags.vim


# Use "doctags" to generate the tags file.  Only works for English!
tags: doctags $(DOCS)
        ./doctags $(DOCS) | LANG=C LC_ALL=C sort >$@
        uniq -d -2 tags


doctags: doctags.c
        $(CC) doctags.c -o doctags
```

PR #14480, opened 2024-04-10
Allow to define GETTEXT_PATH in environment

dertuxmalwieder:
It makes no sense to enforce modifying Vim source files just because your build stuff is not where it's expected. My change is supposed to add support for a locally-defined %GETTEXT_PATH%.

### DIFF
--- a/src/po/Make_mvc.mak
+++ b/src/po/Make_mvc.mak
@@ -44,7 +44,9 @@ VIM = ..\vim.exe
 
 # Correct the following line for the directory where gettext et al is
 # installed.  Please do not put the path in quotes.
+!IFNDEF GETTEXT_PATH
 GETTEXT_PATH = D:\Programs\GetText\bin
+!ENDIF
 
 # Starting from version 0.22, msgfmt forcibly converts text to UTF-8 regardless
 # of the value of the "charset" field.


k-takata:
Your change is okay, but you can specify it from the command line:
```
nmake -f Make_mvc.mak GETTEXT_PATH=C:\path\to\gettext
```

This doesn't work now:
```
set GETTEXT_PATH=C:\path\to\gettext
nmake -f Make_mvc.mak
```
but you can still work around like this:
```
set GETTEXT_PATH=C:\path\to\gettext
nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%
```

dertuxmalwieder:
Currently, GETTEXT_PATH is hardcoded unconditionally. Doesn't that just override whatever was specified on the command line?

RestorerZ:
> Currently, GETTEXT_PATH is hardcoded unconditionally. Doesn't that just override whatever was specified on the command line?

No. The priority is set on the command line.

> NMAKE then assigns macro definitions with the following precedence (highest
> to lowest):
> 1. Macros defined on the command line
> 2. Macros defined in a makefile or include file
> 3. Inherited macros
> 4. Macros defined in the TOOLS.INI file
> 5. Predefined macros (such as CC and RFLAGS)
> 
> Macro definitions are assigned first in order of priority and then in the order in
> which NMAKE encounters them. For example, a macro defined in an include
> file overrides a macro with the same name from the TOOLS.INI file. Note that
> a macro within a makefile can be redefined; a macro is valid from the point it is
> defined until it is redefined or undefined.

dertuxmalwieder:
Ah, so my patch "only" adds support for environment variables. Well, good enough. 😏
*(I did **not** know that, thank you.)*

dertuxmalwieder:
`nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%` does **NOT** work without my patch. It still wants `D:\Programs\GetText\bin\msgfmt.exe`.

chrisbra:
I think it makes sense, thanks!

k-takata:
> `nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%` does **NOT** work without my patch.

Interesting. It works on my side.
Does it work when you execute the commands from the command line?

Expansion of environment variables in batch files has some pitfalls.
E.g. This shows "AAA", not "BBB":
```batch
@echo off
set AAA=AAA
if 1==1 (
	set AAA=BBB
	echo %AAA%
)
```

RestorerZ:
> `nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%` does **NOT** work without my patch. It still wants `D:\Programs\GetText\bin\msgfmt.exe`.

And can you show me what you have in your environment variable `%GETTEXT_PATH%`?

RestorerZ:
> > > `nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%` does **NOT** work without my patch. It still wants `D:\Programs\GetText\bin\msgfmt.exe`.
> > 
> > 
> > And can you show me what you have in your environment variable `%GETTEXT_PATH%`?
> 
> The path to my installation of Gettext.

Yes, I understand that.
What does he look like? What's that path?

k-takata:
Adding `echo` on the top of the line might be helpful to see what is happening:
```
echo nmake -f Make_mvc.mak GETTEXT_PATH=%GETTEXT_PATH%
```

PR #14483, opened 2024-04-10
runtime(doc): Changes in documentation

RestorerZ:
A few adjustments to the documentation

### DIFF
--- a/runtime/doc/helphelp.txt
+++ b/runtime/doc/helphelp.txt
@@ -315,7 +315,7 @@ Hints for translators:
   using the "tag@en" notation.
 - Make a package with all the files and the tags file available for download.
   Users can drop it in one of the "doc" directories and start use it.
-  Report this to Bram, so that he can add a link on www.vim.org.
+  Report to the development team, so they can add a link on www.vim.org.
 - Use the |:helptags| command to generate the tags files.  It will find all
   languages in the specified directory.
 
--- a/runtime/doc/intro.txt
+++ b/runtime/doc/intro.txt
@@ -133,7 +133,7 @@ There are three ways to report bugs:
    This is a maillist, you need to become a member first and many people will
    see the message.  If you don't want that, e.g. because it is a security
    issue, please contact any of the current Vim maintainers
-   https://github.com/orgs/vim/people (but not Bram or the vim-dev ML).
+   https://github.com/orgs/vim/people (but not the vim-dev ML).
    In the future, a proper process for handling security issues will be setup.
 
 Please be brief; all the time that is spent on answering mail is subtracted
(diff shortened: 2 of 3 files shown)


runtime/doc/intro.txt
```diff
@@ -133,7 +133,7 @@ There are three ways to report bugs:
    This is a maillist, you need to become a member first and many people will
    see the message.  If you don't want that, e.g. because it is a security
    issue, please contact any of the current Vim maintainers
-   https://github.com/orgs/vim/people (but not Bram or the vim-dev ML).
+   https://github.com/orgs/vim/people (but not the vim-dev ML).
```
k-takata: Bram's account and the vim-dev ML account used to be listed in this URL.
However, they are now private and not visible for non-vim-org members.
So, the comment in the parentheses is not needed anymore?

runtime/doc/helphelp.txt
```diff
@@ -315,7 +315,7 @@ Hints for translators:
   using the "tag@en" notation.
 - Make a package with all the files and the tags file available for download.
   Users can drop it in one of the "doc" directories and start use it.
-  Report this to Bram, so that he can add a link on www.vim.org.
+  Report to the development team, so they can add a link on www.vim.org.
```
Shane-XB-Qian, line 318: it seems called `the vim project team`.
// and tho the info is out-of-date, but somewhere kept the information or markers or name of Bram it ok.
// e.g `the vim project team, successor of Bram.`

Shane-XB-Qian, line 318: it seems called `the vim project team`.
// and tho the info is out-of-date, but somewhere kept the information or markers or name of Bram is ok.
// e.g `the vim project team, successor of Bram.`

runtime/doc/intro.txt
RestorerZ: Then perhaps it's better to change it at all as stated in SECURITY.md? It's all about security messages.

runtime/doc/helphelp.txt
RestorerZ: > it seems called `the vim project team`. // and tho the info is out-of-date, but somewhere kept the information or markers or name of Bram is ok. // e.g `the vim project team, successor of Bram.`

Yes, it's probably better. But then I guess "The Vim Project" then, as it is specified in PO files, for example.

Shane-XB-Qian: then better uniform it, :smile:

chrisbra:
Thanks. I have made a few more changes, since the voting feature is no longer possible.

RestorerZ:
That's so much better! Thank you!

PR #14495, opened 2024-04-11
Optimize quickfix garbage collection

yegappan:
If garbage collection is already completed for a quickfix list then skip processing that quickfix list.


yegappan:
There is already a copyID check for the contents of a quickfix list (context, user data and callback function).  So there is no need to add it to the quickfix list also.

PR #14497, opened 2024-04-11
filetype: pacman related tools' config files are not recognized

Freed-Wu:
paru.conf use same format of pacman.conf

makepkg use bash syntax, https://man.archlinux.org/man/makepkg.conf.5.en Although `man makepkg.conf` doesn't mention it, e.g.,

```sh
PACKAGER="$DEBFULLNAME <$EMAIL>"
OPTIONS+=(strip zipman purge)
```

can work, add space before `=` will not.


Freed-Wu:
Included to #14505

PR #14499, opened 2024-04-11
filetype: translate shell config files are not recognized

Freed-Wu:
Problem: Translate shell config files are not recognized.
Solution: Add a pattern for translate shell config files.
Refer: https://github.com/soimort/translate-shell/wiki/Configuration

See its syntax highlight


Freed-Wu:
Rebased.

chrisbra:
thanks!

PR #14500, opened 2024-04-11
filetype: xdg-user-dirs files are not recognized

Freed-Wu:
Problem: Xdg-user-dirs files are not recognized.
Solution: Add a pattern for xdg-user-dirs files.

Refer: https://www.freedesktop.org/wiki/Software/xdg-user-dirs/

$(XDG_CONFIG_HOME)/user-dirs.dirs specifies the current set of directories for the user. This file is in a shell format


chrisbra:
please include as part of #14502

Freed-Wu:
Done.

PR #14501, opened 2024-04-11
filetype: matplotlibrc files are not recognized

Freed-Wu:
Problem: Matplotlibrc files are not recognized.
Solution: Add a pattern for matplotlibrc files.
Refer: https://matplotlib.org/stable/users/explain/customizing.html#the-matplotlibrc-file


Freed-Wu:
Rebased.

PR #14502, opened 2024-04-11
filetype: .xprofile files are not recognized

Freed-Wu:
Problem: .xprofile files are not recognized.
Solution: Add a pattern for .xprofile files.


chrisbra:
please include as part of #14505

Freed-Wu:
Done.

PR #14504, opened 2024-04-11
filetype: ctags config files are not recognized

Freed-Wu:
Problem: Ctags config files are not recognized
Solution: Add a pattern for ctags config files
Refer: https://docs.ctags.io/en/latest/option-file.html#order-of-loading-option-files


chrisbra:
please include as of #14509

Freed-Wu:
Done.

PR #14506, opened 2024-04-11
filetype: chktex config files are not recognized

Freed-Wu:
Problem: chktex config files are not recognized.
Solution: Add a pattern for chktex config files.
Refer: https://www.nongnu.org/chktex/


chrisbra:
is good now, thanks!

PR #14507, opened 2024-04-11
filetype: texdoc config files are not recognized

Freed-Wu:
Problem: texdoc config files are not recognized.
Solution: Add a pattern for texdoc config files.
Refer: https://github.com/TeX-Live/texdoc/blob/master/texdoc.cnf


chrisbra:
thanks!

PR #14508, opened 2024-04-11
filetype: flatpak config files are not recognized

Freed-Wu:
Problem: flatpak config files are not recognized.
Solution: Add a pattern for flatpak config files.


chrisbra:
please include into #14512

Freed-Wu:
Done.

PR #14509, opened 2024-04-11
filetype: reply config files are not recognized

Freed-Wu:
Problem: reply config files are not recognized.
Solution: Add a pattern for reply config files.
Refer: https://metacpan.org/dist/Reply/view/bin/reply#-cfg-~/.replyrc


chrisbra:
please include into #14512

Freed-Wu:
Done.

PR #14510, opened 2024-04-11
filetype: wakatime config files are not recognized

Freed-Wu:
Problem: wakatime config files are not recognized.
Solution: Add a pattern for wakatime config files.
Refer: https://github.com/wakatime/wakatime-cli?tab=readme-ov-file#usage


chrisbra:
please include into #14512

Freed-Wu:
Done.

PR #14511, opened 2024-04-11
filetype: x11vnc config files are not recognized

Freed-Wu:
Problem: x11vnc config files are not recognized.
Solution: Add a pattern for x11vnc config files.
Refer: https://linux.die.net/man/1/x11vnc


chrisbra:
thanks!

chrisbra:
thanks!

PR #14513, opened 2024-04-11
filetype: some history files are not recognized

Freed-Wu:
Problem: some history files are not recognized.
Solution: Add patterns for those history files.


chrisbra:
thanks!

PR #14519, opened 2024-04-12
ftplugin(bitbake): add include

Freed-Wu:


Freed-Wu:
Oh, I should ask upstream: https://github.com/openembedded/bitbake/tree/master/contrib/vim

PR #14521, opened 2024-04-12
ftplugin(kconfig): add include

Freed-Wu:


chrisbra:
please go via the upstream repo

PR #14523, opened 2024-04-12
ftplugin(xml): add include

Freed-Wu:


chrisbra:
please, for upstream maintained plugins, go via the upstream repo.

PR #14527, opened 2024-04-12
Filler lines not checked properly in get_scroll_overlap()

zeertzjq:
Problem:  Filler lines not checked properly in get_scroll_overlap().
Solution: Add missing parentheses.

The missing parentheses causes the second argument to diff_check_fill()
to always be 0 as it is the result of a comparison between a positive
integer and -1 (the value of BACKWARD), in which case diff_check_fill()
always returns 0 instead of the number of filler lines above a line.

It's very hard to add a test for this, because this mistake at most
leads to 2 screen lines of difference in scrolling behavior, and in
cases where it may indeed lead to a difference in behavior, neither
behavior achieves complete symmetry between CTRL-F and CTRL-B.

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -3102,7 +3102,7 @@ static int get_scroll_overlap(int dir)
 
     loff.lnum = dir == FORWARD ? curwin->w_botline : curwin->w_topline - 1;
 #ifdef FEAT_DIFF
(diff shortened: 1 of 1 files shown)


chrisbra:
okay thanks. Looks good.

Issue #14491, opened 2024-04-10
command-block limitations 


balki:
### Steps to reproduce

Found below limitations when translating functions to command-block

1. No return statement allowed. E.g.
```
command Foo {
if someCond() | return | endif
// do something 
}
```

2. let-heredoc are not allowed.
```
command Foo {             
    const abc =<< trim END
        hello world       
        blah blah         
    END                   
    echo abc              
}                         
Foo
# E488: Trailing characters:         hello world^@       ...
```

### Expected behaviour

Fix the limitations or add the limitations to docs. 

### Version of Vim

9.1.252

### Environment

ArchLinux

benknoble:
What should return do? It wouldn’t be valid in a normal command block.

It would probably be nice to support heredocs though.

chrisbra:
yeah, return is strange. commands are basically nice and simple ways to execute functions. Thus I would argue the desired logic belongs into a function.

balki:
return does not have to return anything to called. But useful as a control flow to stop the command execution to avoid nesting. Creating a wrapper function is not bad but it is unnecessary boiler plate if the function is called in only once which is in the command definition. I think the behavior should at-least be documented.

chrisbra:
`:return` does not make sense in an interactive session (which is basically the same what `:com` does). I also think there is nothing to document at least not for the `:return` case.

And btw: you can always negate the condition.

chrisbra:
Yes, let's leave this ticket for the missing heredoc functionality.

PR #14528, opened 2024-04-12
Support heredoc in a block used for :command

yegappan:
Addresses the limitation mentioned in #14491.


chrisbra:
Thanks Yegappan!

PR #14530, opened 2024-04-12
runtime(vim): Update base-syntax, add legacy header folding

dkearns:
Allow for syntax-based folding of Vim9 script legacy header regions.

This is enabled with the "H" flag of the g:vimsyn_folding config variable.


chrisbra:
thanks!

runtime/doc/syntax.txt
```diff
@@ -3764,12 +3764,15 @@ Some folding is now supported with syntax/vim.vim: >
...
+   g:vimsyn_folding =~ 'h' : fold heredocs
+   g:vimsyn_folding =~ 'H' : fold Vim9-script legacy headers
    g:vimsyn_folding =~ 'l' : fold lua      script
    g:vimsyn_folding =~ 'm' : fold mzscheme script
    g:vimsyn_folding =~ 'p' : fold perl     script
    g:vimsyn_folding =~ 'P' : fold python   script
    g:vimsyn_folding =~ 'r' : fold ruby     script
    g:vimsyn_folding =~ 't' : fold tcl      script
+let g:vimsyn_folding = 'acfhlmpPrt'
 <
```
h-east, line 3776: Is this line(L3775) intentional?
(Need an indentation?)

dkearns: I'm not sure but it looks like there should be a short explanation there about the default value so I'll add that to my next syntax file update today or tomorrow.

Thanks for noticing.

PR #14531, opened 2024-04-12
i3config: Line continuation is not detected for 'set' command

julio-b:
Problem:  Valid i3config syntax is highlighted as error.
Solution: Skip over line-breaks correctly.

### DIFF
--- a/runtime/syntax/i3config.vim
+++ b/runtime/syntax/i3config.vim
@@ -100,7 +100,7 @@ syn match i3ConfigKeyword /^no_focus .*$/ contains=i3ConfigCondition
 " 4.17 Variables
 syn match i3ConfigVariable /\$[0-9A-Za-z_:|[\]-]\+/
 syn keyword i3ConfigSetKeyword set contained
-syn match i3ConfigSet /^set \$.*$/ contains=i3ConfigSetKeyword,i3ConfigVariable,i3ConfigColor,i3ConfigString,i3ConfigNumber,i3ConfigShCommand,i3ConfigShDelim,i3ConfigShParam,i3ConfigShOper,i3ConfigBindModkey
+syn region i3ConfigSet start=/^set\s\+\$/ skip=/\\$/ end=/$/ contains=i3ConfigSetKeyword,i3ConfigVariable,i3ConfigColor,i3ConfigString,i3ConfigNumber,i3ConfigShCommand,i3ConfigShDelim,i3ConfigShParam,i3ConfigShOper,i3ConfigBindModkey keepend
 
 " 4.18 X resources
 syn keyword i3ConfigResourceKeyword set_from_resource contained


julio-b:
Before/After screeshot:
2024-04-12-205158_370x255_scrot
CC: @hiqua @JosefLitos

JosefLitos:
Never thought about that, looks good.

Though, could I ask you to base the changess off of my plugin, where I keep the syntax? - I made a couple other changes that were just regarding string inner command parsing, so I didn't bother updating it here, so if you could take those changes and add this one / I will add this change to the file right now. (It would also require the swayconfig file update)

litoj (changes requested):
`\$` cannot be part of the start match because then the `i3ConfigVariable` highlight won't work:
obrazek

I fixed it and pushed the commit to [i3config.vim](https://github.com/JosefLitos/i3config.vim/blob/master/syntax/i3config.vim#L108)

julio-b:
> Never thought about that, looks good.
> 
> Though, could I ask you to base the changess off of my plugin, where I keep the syntax? - I made a couple other changes that were just regarding string inner command parsing, so I didn't bother updating it here, so if you could take those changes and add this one / I will add this change to the file right now.

> (It would also require the swayconfig file update)

I don't use sway, but it seems that swayconfig by default  [includes i3config](https://github.com/vim/vim/blob/e74cad3321ce1dcefc1fc64f617511275b6cd930/runtime/syntax/swayconfig.vim#L19C1-L19C29). So it should recognize the new `set` syntax, right?


> `\$` cannot be part of the start match because then the `i3ConfigVariable` highlight won't work:

Hm, i can't reproduce this. Maybe this is because in your version you added `matchgroup=i3ConfigKeyword`?

> obrazek
> 
> I fixed it and pushed the commit to [i3config.vim](https://github.com/JosefLitos/i3config.vim/blob/master/syntax/i3config.vim#L108)

Nice! I see that you didn't include the `keepend` there. In my testings it was required. Otherwise the `set` region may be expanded to the next line incorrectly. E.g this syntax is not valid because the `\` on the second line is missing:
```
set $mycmd xterm \
       -e bash
       -c "program"
```
Thank you!

JosefLitos:
> E.g this syntax is not valid because the \ on the second line is missing...

When you do `:Inspect` on the extra string line, does it really show also being under the `i3ConfigSet` group? For me it did not, but that may be because of the refactoring I did, but haven't yet pushed/updated to vim.

In my case it also gets highlighted, but because of a fixup long-block-highlighting mechanism for swayconfig, but when set to i3config syntax I wasn't able to reproduce it.

Would you mind, if I made a separate PR with updating `i3config` (and `swayconfig`) with my overall refactor and included my version of the fix there?

litoj (approved):

chrisbra:
@JosefLitos You approved, so I'll merge. Please feel free to send an updated PR for `i3config` and `swayconfig` to further improve it.

PR #14532, opened 2024-04-12
fix: using 'vimdoc' as vim ft compiler was a bit aggressive

Shane-XB-Qian:
// trying to mdf if really to keep it as a 'compiler' 
// or maybe just back out #14459

### DIFF
--- a/runtime/compiler/vimdoc.vim
+++ b/runtime/compiler/vimdoc.vim
@@ -1,9 +1,10 @@
 " Vim Compiler File
 " Language:             vimdoc
 " Maintainer:           Wu, Zhenyu <wuzhenyu@ustc.edu>
-" Latest Revision:      2024-04-09
+" Latest Revision:      2024-04-13
 "
-" you can get it by `pip install vimdoc` or the package manager of your distribution.
+" If you can not find 'vimdoc' in the package manager of your distribution e.g
+" 'pip', then you may need to build it from its source.
 
 if exists('b:current_compiler')
   finish
--- a/runtime/ftplugin/vim.vim
+++ b/runtime/ftplugin/vim.vim
@@ -1,7 +1,7 @@
 " Vim filetype plugin
 " Language:		Vim
 " Maintainer:		Doug Kearns <dougkearns@gmail.com>
-" Last Change:		2024 Apr 08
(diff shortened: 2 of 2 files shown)


Freed-Wu:
Maybe we can use vint as default compiler of vim https://github.com/Vimjas/vint/issues/393

runtime/compiler/vimdoc.vim
```diff
@@ -1,9 +1,10 @@
 " Vim Compiler File
 " Language:             vimdoc
 " Maintainer:           Wu, Zhenyu <wuzhenyu@ustc.edu>
-" Latest Revision:      2024-04-09
+" Latest Revision:      2024-04-13
 "
-" you can get it by `pip install vimdoc` or the package manager of your distribution.
+" if can not find 'vimdoc' in the package manager of your distribution e.g 'pip',
```
zeertzjq:
```suggestion
" If you cannot find 'vimdoc' in the package manager of your distribution e.g 'pip',
```

runtime/compiler/vimdoc.vim
```diff
@@ -13,7 +14,7 @@ let b:current_compiler = 'vimdoc'
 let s:save_cpoptions = &cpoptions
 set cpoptions&vim
 
-CompilerSet makeprg=vimdoc
+CompilerSet makeprg=vimdoc\ .
```
chrisbra: Please remove the dot.

Shane-XB-Qian:
> -CompilerSet makeprg=vimdoc
> +CompilerSet makeprg=vimdoc\ .
> 
> Please remove the dot.

ah.. remove this `dot` maybe worse, it probably would raise err or err msg,
but i would not insist it, mostly just was for making it not impact vim ft,
so if you confirm (and not to back out) i will remove that `dot`.

-- 
shane.xb.qian

Shane-XB-Qian:
> -CompilerSet makeprg=vimdoc
> +CompilerSet makeprg=vimdoc\ .
> 
> Please remove the dot.

nevermind, i felt removed that `dot` should be a bit safer than had, since seems it would override `./doc` by force.

-- 
shane.xb.qian

chrisbra:
Thanks.

PR #14535, opened 2024-04-13
Crash when using heredoc with comment in command block

zeertzjq:
Problem:  Crash when using heredoc with comment in command block.
Solution: Handle a newline more like the end of the line.


zeertzjq:
Note: This fixes the coverity warning.

yegappan (approved):

chrisbra:
thanks!

PR #14537, opened 2024-04-13
Fail the Vim9 script compilation when a class is defined inside a function

### LINKED ISSUE #13184: public modifier allowed in static but not object method declarations

### LINKED ISSUE #13326: No compile error for classes definined in functions
### Steps to reproduce

```vim
vim9script

def Foo(): void
  class Bar
  endclass
enddef

defcompile
```

This compiles but a runtime error is output when `Foo()` is called: `E476: Invalid command: endclass:   endclass`

### Expected behaviour

A compile-time error message: `E476: Invalid command: class Bar`

The runtime message doesn't look correct either.

### Version of Vim

9.0.2004

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Class definition inside a function is not supported.  Fail the compilation and give a proper error message.  Fixes #13326.
Display a consistent error message when the public keyword is used with class and object methods.  Fixes #13184.

### DIFF
--- a/runtime/doc/vim9.txt
+++ b/runtime/doc/vim9.txt
@@ -1521,7 +1521,6 @@ Custom types can be defined with `:type`: >
 	:type MyList list<string>
 Custom types must start with a capital letter, to avoid name clashes with
 builtin types added later, similarly to user functions.
-{not implemented yet}
 
 And classes and interfaces can be used as types: >
 	:class MyClass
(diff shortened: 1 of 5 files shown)


chrisbra:
thanks!

PR #14538, opened 2024-04-13
runtime(go): fix highlighting import string followed by some comment

rhysd:
This patch is upstreamed from vim-go project.

- https://github.com/fatih/vim-go/pull/3657
- https://github.com/fatih/vim-go/commit/9049ae085ed2f822c377e8caaa96b0d02859470d

I confirmed the maintainer has no concern on upstreaming this patch by [this comment](https://github.com/fatih/vim-go/pull/3657#issuecomment-2052209898).

This PR fixes the issue described in https://github.com/fatih/vim-go/issues/3656. When some characters follow an import string, the import string loses its syntax highlight.


chrisbra:
thanks!

Issue #14529, opened 2024-04-12
`interpolated string` has inconsistent handling of containers


errael:
### Steps to reproduce

```
vim9script

var l = [ 'one', 'two' ]
echo '(1)' string(l)
echo '(2)' printf("%s", l)
echo '(3)' $'{l}'

echo '================='

var d = {a: 1, b: 2}
echo '(4)' string(d)
echo '(5)' printf("%s", d)
echo '(6)' $'{d}'
```
(3) - each list item is followed by a new line
(6) - fails on dictionary
```
(1) ['one', 'two']
(2) ['one', 'two']
(3) one
two
=================
(4) {'a': 1, 'b': 2}
(5) {'a': 1, 'b': 2}
(6)
Error detected while processing :source buffer=1:
line   13:
E731: Using a Dictionary as a String
```

### Expected behaviour

According to the docs
```
Any expression returning a value can be enclosed between curly braces.
The value is converted to a string.
```
Does this mean it should be like `string(expr)`?

### Version of Vim

9.1.301

### Environment

ubuntu/gtk

chrisbra:
I guess it should mean `string(expr)` then.

errael:
Closing as "won't fix" (`dict` is fixed, not `list`). See https://github.com/vim/vim/pull/14541#issuecomment-2054157620. Not sure what #7684 has to do with interpolated string.

Anyway, `${string(l)}` is a workaround.

> I guess it should mean string(expr) then.

Guess not.

zeertzjq:
Another way is to put it inside a List:
```vim
vim9script

var l = [ 'one', 'two' ]
echo '(1)' string(l)
echo '(2)' printf("%s", l)
echo '(3)' $'{[l]}'

echo '================='

var d = {a: 1, b: 2}
echo '(4)' string(d)
echo '(5)' printf("%s", d)
echo '(6)' $'{[d]}'
```
Uutput:
```
(1) ['one', 'two']
(2) ['one', 'two']
(3) ['one', 'two']
=================
(4) {'a': 1, 'b': 2}
(5) {'a': 1, 'b': 2}
(6) {'a': 1, 'b': 2}
```

errael:
My bad, it was completed.

Issue #7684, opened 2021-01-15
Vim9: cannot use ":put =expr" to append list of lines


lacygoill:
**Describe the bug**

In Vim9 script, we cannot use `:put` to append a list of lines in the buffer via the expression register.

**To Reproduce**

Run this shell command:

    vim -Nu NONE -S <(cat <<'EOF'
        vim9
        def Func()
            pu =['a', 'b', 'c']
        enddef
        Func()
    EOF
    )

This single line is appended into the buffer:

    ['a', 'b', 'c']

**Expected behavior**

These three lines are appended into the buffer:

    a
    b
    c

**Environment**

 - Vim version: 8.2 Included patches: 1-2352
 - OS: Ubuntu 16.04.7 LTS
 - Terminal: xterm(363)

**Additional context**

Regression introduced in [8.2.1637](https://github.com/vim/vim/releases/tag/v8.2.1637).

lacygoill:
The issue is limited to a `:def` function.  At the script level, `:put =expr` correctly appends 3 lines:
```vim
vim9
pu =['a', 'b', 'c']
```
    a
    b
    c

PR #14541, opened 2024-04-13
Support string interpolation for Dict

yegappan:
When a Dict is used in an interpolated string or heredoc, convert the Dict to a string.
Addresses the issue described in #14529.

This also enables the support for using a Dict with the `:put =` command.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -575,7 +575,8 @@ skip_expr_concatenate(
 
 /*
  * Convert "tv" to a string.
- * When "convert" is TRUE convert a List into a sequence of lines.
+ * When "convert" is TRUE convert a List into a sequence of lines and a Dict
+ * into a textual representation of the Dict.
  * Returns an allocated string (NULL when out of memory).
  */
     char_u *
@@ -596,6 +597,8 @@ typval2string(typval_T *tv, int convert)
 	ga_append(&ga, NUL);
 	retval = (char_u *)ga.ga_data;
     }
+    else if (convert && tv->v_type == VAR_DICT)
+	retval = dict2string(tv, get_copyID(), FALSE);
     else
 	retval = vim_strsave(tv_get_string(tv));
     return retval;
--- a/runtime/doc/change.txt
+++ b/runtime/doc/change.txt
@@ -1345,8 +1345,8 @@ expression (like with the "/" command).
 The expression must evaluate to a String.  A Number is always automatically
 converted to a String.  For the "p" and ":put" command, if the result is a
 Float it's converted into a String.  If the result is a List each element is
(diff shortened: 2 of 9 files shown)


src/eval.c
```diff
@@ -596,6 +596,8 @@ typval2string(typval_T *tv, int convert)
 	ga_append(&ga, NUL);
 	retval = (char_u *)ga.ga_data;
     }
+    else if (convert && tv->v_type == VAR_DICT)
```
zeertzjq, line 600: The comment above this function should need update.

yegappan: Thanks.  Updated the comment to mention that a Dict is converted into its textual representation.

errael:
```
vim9script
echo '==='
var l = [ 'one', 'two' ]
echo '(3)' $'{l}'
echo '==='
```
still has a problem, expect $'{l}' to be on a single line, not
```
===
(3) one
two
===
```

chrisbra:
this PR only includes Dict support, not lists.

yegappan:
> ```
> vim9script
> echo '==='
> var l = [ 'one', 'two' ]
> echo '(3)' $'{l}'
> echo '==='
> ```
> 
> still has a problem, expect $'{l}' to be on a single line, not
> 
> ```
> ===
> (3) one
> two
> ===
> ```

This is by design.  Refer to [patch 8.2.2356](https://github.com/vim/vim/commit/883cf97f109d2ff281cf77f7b2e3bb44aced7cb3) and issue https://github.com/vim/vim/issues/7684.

With the current implementation, you can use the `:put =list` command to add a list of lines.
In this case, returning string(list) is not useful.

errael:
> This is by design. Refer to [patch 8.2.2356](https://github.com/vim/vim/commit/883cf97f109d2ff281cf77f7b2e3bb44aced7cb3) and issue #7684.
> 
> With the current implementation, you can use the `:put =list` command to add a list of lines. In this case, returning string(list) is not useful.

Thanks for the reference. I don't see what that has to do with interpolated string: `${l}`.

But I see that this PR is about `Dict`.

chrisbra:
thanks!

yegappan:
> > This is by design. Refer to [patch 8.2.2356](https://github.com/vim/vim/commit/883cf97f109d2ff281cf77f7b2e3bb44aced7cb3) and issue #7684.
> > With the current implementation, you can use the `:put =list` command to add a list of lines. In this case, returning string(list) is not useful.
> 
> Thanks for the reference. I don't see what that has to do with interpolated string: `${l}`.
> 

They are related because both the `:put` command and string interpolation are using the same
code path to convert a List into a string.  I will see whether we can differentiate between the two.

PR #14542, opened 2024-04-13
Garbled output on serial terminals with XON/XOFF flow control

### LINKED ISSUE #12674: Garbled output on serial terminals with XON/XOFF flow control after patch 8.2.0852
### Steps to reproduce

1. Login over a serial terminal, in this case a DEC VT420 which only supports XON/XOFF flow control, and not CTS/RTS.
2. run vim.
3. encounter severe screen corruption, especially on large redraws, and even spurious inputs.

I tracked down the problem to a regression introduced with patch 8.2.0852, referencing bug report #6166. Removing IXON from the bitmask inversion at `src/os_unix.c:3670`, fixes the issue.

While it's not unreasonable to want the ability to bind ctrl-s to something, the solution cannot be to make vim unusable over serial lines relying on software flow control. At the very least this should be made an option.

### Expected behaviour

I expected vim to be usable on a common DEC serial terminal.

### Version of Vim

9.0.1672

### Environment

Operating System: Debian GNU/Linux
Terminal: DEC VT420
$TERM: vt420
shell: bash 5.2.15

The critical factor is that the terminal relies on XON/XOFF flow control. All other details are irrelevant.

### Logs and stack traces

_No response_

ant0sha:
Problem:  When used terminal with XON/XOFF flow control, vim tries to
          still make CTRL-S mapping available, which results in severe
          screen corruption, especially on large redraws, and even
          spurious inputs
Solution: Disallow CTRL-S mapping if such terminal is recognized.
          Don't remove IXON from the bitmask inversion.

closes: #12674 (well, at least CTRL-S mapping is disabled, cannot test vt420 corruption because I do not have such hardware)

**When started with TERM=vt420**:

    TERM=vt420 vim

subsequent

    :set termcap

shows ``t_xon=y``

mapping

    map <C-S> :echo "abc"<CR>

does nothing (after ``<C-S>`` output freezes and subsequent ``<C-Q>``
unfreezes it)

**When started with TERM=xterm**:

    TERM=xterm vim

subsequent

    :set termcap

shows ``t_xon=``

mapping

    map <C-S> :echo "abc"<CR>

works (after ``<C-S>`` one see ``abc`` string echo-ed)

### DIFF
--- a/src/optiondefs.h
+++ b/src/optiondefs.h
@@ -3014,6 +3014,7 @@ static struct vimoption options[] =
     p_term("t_8f", T_8F)
     p_term("t_8b", T_8B)
     p_term("t_8u", T_8U)
+    p_term("t_xon", T_XON)
 
 // terminal key codes are not in here
 
--- a/src/os_unix.c
+++ b/src/os_unix.c
@@ -3766,7 +3766,8 @@ mch_settmode(tmode_T tmode)
     {
 	// ~ICRNL enables typing ^V^M
 	// ~IXON disables CTRL-S stopping output, so that it can be mapped.
-	tnew.c_iflag &= ~(ICRNL | IXON);
+	tnew.c_iflag &= ~(ICRNL |
+		(T_XON == NULL || *T_XON == NUL ? IXON : 0));
 	tnew.c_lflag &= ~(ICANON | ECHO | ISIG | ECHOE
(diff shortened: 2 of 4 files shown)


chrisbra:
thanks, I have made a few minor changes, mostly documentation and tests.

PR #14543, opened 2024-04-13
runtime(java): Recognise non-ASCII identifiers

zzzyxwvut:
Also:

- Remove the already commented out and less general in its  
  definition `javaFuncDef` alternative.

- Stop recognising some bespoke `{p,trace}` debugging API.

Non-ASCII identifiers have been supported from the outset  
of the Java language.

> An _identifier_ is an unlimited-length sequence of _Java
> letters_ and _Java digits_, the first of which must be a
> Java letter.  An identifier cannot have the same spelling
> (Unicode character sequence) as a keyword . . . Boolean
> literal . . . or the null literal . . .
> . . . . . . . .
> Letters and digits may be drawn from the entire Unicode
> character set . . .
> . . . . . . . .
> A Java letter is a character for which the method
> `Character.isJavaLetter` . . . returns `true`.  A Java
> letter-or-digit is a character for which the method
> `Character.isJavaLetterOrDigit` . . . returns `true`.
> . . . . . . . .
> The Java letters include . . . for historical reasons, the
> ASCII underscore (`_`) . . . and dollar sign (`$`) . . .

(Separate syntax tests will be written when particular parts  
now touched will have been further improved.)

Reference:
https://javaalmanac.io/jdk/1.0/langspec.pdf [§3.8]

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -156,6 +156,7 @@ runtime/ftplugin/icon.vim		@dkearns
 runtime/ftplugin/indent.vim		@dkearns
 runtime/ftplugin/ishd.vim		@dkearns
 runtime/ftplugin/j.vim			@glts
+runtime/ftplugin/java.vim		@zzzyxwvut
 runtime/ftplugin/javascript.vim		@dkearns
 runtime/ftplugin/javascriptreact.vim	@dkearns
 runtime/ftplugin/json.vim		@dbarnett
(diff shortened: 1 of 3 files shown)


zzzyxwvut:
As previously [discussed](https://github.com/vim/vim/pull/14403#discussion_r1552039319), I'm willing to take on the  
maintenance of Java filetype and syntax files.

chrisbra:
Thanks!

PR #14544, opened 2024-04-13
perf(syn-i3/swayconfig): clusters, nextgroup and command variant separation

litoj:
Since the last update, I made another two refactorings and in order to use clusters, to reduce repetition, and nextgroup, to improve user experience when writing code.

Other changes include:
- support multiline `set` as discovered in #14531
- improve command-within-string handling
- fallback for lines from blocks that are out of parsing range (_"orphan"_ blocks)
- trying to remove as many extra explicit keyword groups as possible → using regions with appropriate matchgroup

I apologize for the awful diffs, but refactor of the entire structure to use a slightly different approach will probably always result in such a mess.

cc @hiqua @jamespeapen

### DIFF
--- a/runtime/syntax/i3config.vim
+++ b/runtime/syntax/i3config.vim
@@ -2,8 +2,8 @@
 " Language: i3 config file
 " Original Author: Josef Litos (JosefLitos/i3config.vim)
 " Maintainer: Quentin Hibon (github user hiqua)
-" Version: 1.0.2
-" Last Change: 2023-12-28
+" Version: 1.2.1
+" Last Change: 2024-04-14
 
 " References:
 " http://i3wm.org/docs/userguide.html#configuring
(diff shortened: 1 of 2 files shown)


jamespeapen:
Thanks! - Checking against my test file, the sway changes look like an improvement, especially catching syntax errors. However, there are a couple of keywords with broken support that were previously working:

image
https://github.com/swaywm/sway/blob/087226d997c15f4df30542778854999c632642a3/sway/sway.5.scd?plain=1#L117

image
https://github.com/swaywm/sway/blob/087226d997c15f4df30542778854999c632642a3/sway/sway.5.scd?plain=1#L672
This also affects i3 configs.

image
https://github.com/swaywm/sway/blob/087226d997c15f4df30542778854999c632642a3/sway/sway.5.scd?plain=1#L776
See https://github.com/JosefLitos/i3config.vim/pull/1/commits/9dbe0c5ce07be3cb3773449237774c3cfaca35df

JosefLitos:
> pic<floating>

this is intended - floating modifier is valid only as an action → `bindsym x floating toggle`

I have fixed the other two along with mostly completing the transition to using `nextgroup` to determine the syntax, now also for the sway side.

I have my own test file + my personal configs with which I check, but I have a quite simple config, so it is not impossible I've missed/broken yet something else.

chrisbra:
thanks!

PR #14546, opened 2024-04-14
Using heredoc in string not tested with :execute

zeertzjq:
Problem:  Using heredoc in string not tested with :execute.
Solution: Test with both :execute and execute().


chrisbra:
thanks!

PR #14548, opened 2024-04-14
macos: use explicit version tags for runners

pheiduck:
as we do on others like ubuntu
`macos-14` can become `macos-latest` soon if we want that anyway.


chrisbra:
thanks!

PR #14550, opened 2024-04-14
document pandoc compiler and enable configuring arguments

Konfekt:
This enables globally, or buffer, in particular filetype, locally, configuring the options passed to pandoc


chrisbra:
thanks!

PR #14554, opened 2024-04-15
No test for highlight behavior with 'ambiwidth'

zeertzjq:
Problem:  No test for highlight behavior with 'ambiwidth'.
Solution: Add a screendump test for 'ambiwidth' with 'cursorline'.


chrisbra:
thanks!

PR #14555, opened 2024-04-15
Some assert_equal() calls have wrong order of arguments

zeertzjq:
Problem:  Some assert_equal() calls have wrong order of arguments.
Solution: Correct the order.


chrisbra:
thanks, yeah this is confiusing. I always have to lookup the order.

PR #14556, opened 2024-04-15
When interpolating a string with a List variable, use the textual representation of the List

yegappan:


src/eval.c
```diff
@@ -597,7 +597,9 @@ typval2string(typval_T *tv, int convert)
 	ga_append(&ga, NUL);
 	retval = (char_u *)ga.ga_data;
     }
-    else if (convert && tv->v_type == VAR_DICT)
+    else if (tv->v_type == VAR_LIST)
+	retval = list2string(tv, get_copyID(), FALSE);
+    else if (tv->v_type == VAR_DICT)
 	retval = dict2string(tv, get_copyID(), FALSE);
```
zeertzjq: Maybe this should use `tv2string()` instead? That handles NULL list and NULL dict.

zeertzjq:
The argument name `convert` also seems no longer meaningful. Maybe it should be renamed `join_list` instead.

zeertzjq:
The argument name `convert` in `typval2string()` and `eval_to_string()` also seems no longer meaningful. Maybe it should be renamed to `join` instead.

src/eval.c
yegappan: Thanks.  Modified the code to use tv2string() for Dict and List.  Add additional tests for NULL List/Dict.

yegappan:
> The argument name `convert` in `typval2string()` and `eval_to_string()` also seems no longer meaningful. Maybe it should be renamed to `join` instead.

Renamed the argument from "convert" to "join_list"..

chrisbra:
thanks!

PR #14552, opened 2024-04-14
Makefile: install rust and cargo autoload files

mhammerly:
Fixes https://github.com/vim/vim/issues/14551, essentially https://github.com/vim/vim/commit/c8b6d4b37853b24c92d9166eb48636918b6e676e but for the Rust and Cargo folders.

[`/runtime/compiler/cargo.vim`](https://github.com/vim/vim/blob/f23193496589553d986e0f5c2a6170485c1bccea/runtime/compiler/cargo.vim#L27-L28) attempts to call `cargo#quickfix#CmdPre()` and `cargo#quickfix#CmdPost()` from [`runtime/autoload/cargo/quickfix.vim`](https://github.com/vim/vim/blob/f23193496589553d986e0f5c2a6170485c1bccea/runtime/autoload/cargo/quickfix.vim#L3-L27), but the `autoload/cargo` directory is not actually part of the distribution.

I tested with:
```
make unixall
cd dist/vim91
./configure --prefix=/throwaway
make
make install
```

Without the patch, the files were not in the resulting installation: 
[install-without-patch.tar.gz](https://github.com/vim/vim/files/14972380/install-without-patch.tar.gz)

With the patch, the files were there: 
[install-with-patch.tar.gz](https://github.com/vim/vim/files/14972381/install-with-patch.tar.gz)

PR #14557, opened 2024-04-15
fix: lost rust and cargo runtime dir

### LINKED ISSUE #14551: runtime directories autoload/cargo and autoload/rust are missing from Makefile output
### Steps to reproduce

[Lines 2423-2430 of src/Makefile](https://github.com/vim/vim/blob/f23193496589553d986e0f5c2a6170485c1bccea/src/Makefile#L2423-L2430) copy the source runtime files to the build destination runtime files. The runtime autoload sub-directories are listed in the Makefile: `autoload/dist/`, `autoload/xml/` and `autoload/zig/`. However, the `autoload/cargo/` and `autoload/rust/` directories are not included, so after vim is installed the runtime files in these subdirectories are missing.

This means that, for example, these [runtime/compiler/cargo.vim](https://github.com/vim/vim/blob/master/runtime/compiler/cargo.vim#L27-L28) lines don't work:

```vim
    autocmd QuickFixCmdPre make call cargo#quickfix#CmdPre()
    autocmd QuickFixCmdPost make call cargo#quickfix#CmdPost()
```

### Expected behaviour

The `runtime/autoload/cargo/*.vim` and `runtime/autoload/rust/*.vim` files should be included in vim make output.

I tried to make a PR, similar to [patch 9.0.1658](https://github.com/vim/vim/commit/c8b6d4b37853b24c92d9166eb48636918b6e676e) which added `runtime/autoload/zig/*.vim`, but it didn't work for me so I'll just open an issue for people with more experience with vim builds to look at.

### Version of Vim

9.1.0330

### Environment

linux and mac

### Logs and stack traces

_No response_

Shane-XB-Qian:
closes #14551

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2340,7 +2340,7 @@ installrtbase: $(HELPSOURCE)/vim.1 $(DEST_VIM) $(VIMTARGET) $(DEST_RT) \
 		$(DEST_HELP) $(DEST_PRINT) $(DEST_COL) \
 		$(DEST_SYN) $(DEST_SYN)/modula2 $(DEST_SYN)/modula2/opt $(DEST_SYN)/shared \
 		$(DEST_IND) $(DEST_FTP) \
-		$(DEST_AUTO) $(DEST_AUTO)/dist $(DEST_AUTO)/xml $(DEST_AUTO)/zig \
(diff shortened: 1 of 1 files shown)


k-takata:
It looks like duplicate of #14552.

Shane-XB-Qian:
> It looks like duplicate of #14552.

um... ok, didnot notice it, then you or chris can pick one to process.

chrisbra:
thanks!

PR #14563, opened 2024-04-15
Problem: Error with matchaddpos() and empty list

### LINKED ISSUE #14525: E340 + E341 Internal Error
### Steps to reproduce

`vim --clean`
`:echo matchaddpos('Error', [])`

Then:

```
E340: Internal error; if you can reproduce please report a bug
E341: Internal error; lalloc(0, )
```

### Expected behaviour

echo a match id.

### Version of Vim

9.1 and latest 9.1.0309, but works well on 9.0.

### Environment

Windows 10

### Logs and stack traces

_No response_

chrisbra:
Problem:  Error with matchaddpos() and empty list
          (@rickhow)
Solution: Return early for an empty list

fixes: #14525

### DIFF
--- a/src/testdir/test_match.vim
+++ b/src/testdir/test_match.vim
@@ -312,6 +312,8 @@ func Test_matchaddpos_error()
   call assert_fails("call matchaddpos('Search', [[{}]])", 'E728:')
   call assert_fails("call matchaddpos('Search', [[2, {}]])", 'E728:')
   call assert_fails("call matchaddpos('Search', [[3, 4, {}]])", 'E728:')
+  call assert_equal(-1, matchaddpos('Error', test_null_list()))
+  call assert_equal(-1, matchaddpos('Error', []))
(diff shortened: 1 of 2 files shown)


src/testdir/test_match.vim
```diff
@@ -312,6 +312,8 @@ func Test_matchaddpos_error()
   call assert_fails("call matchaddpos('Search', [[{}]])", 'E728:')
   call assert_fails("call matchaddpos('Search', [[2, {}]])", 'E728:')
   call assert_fails("call matchaddpos('Search', [[3, 4, {}]])", 'E728:')
+  call assert_equal(-1, matchaddpos('Error', test_null_list()))
```
zeertzjq, line 315: This is already tested 6 lines above.

chrisbra: one can never have enough tests! :)

PR #14564, opened 2024-04-16
Convert screen dump XDG tests to normal tests

yegappan:
Screen dump tests are flaky, slow to run and not flexible.  If possible, it is better to avoid them.
Convert the screen dump tests for the new XDG feature to normal tests.


yegappan:
For some reason, the last test is failing in the CI environment.  This test passes in my local Linux setup.

chrisbra:
it seems $XDG_CONFIG_HOME is defined in the runners, so you need to unset it. https://github.com/vim/vim/commit/65e237529e732bdae8b67008f596a9ac902f5b34

yegappan:
> it seems $XDG_CONFIG_HOME is defined in the runners, so you need to unset it. [65e2375](https://github.com/vim/vim/commit/65e237529e732bdae8b67008f596a9ac902f5b34)

Thanks .  I have updated the PR with this change.  Let us see whether the CI tests pass now.

yegappan:
> > it seems $XDG_CONFIG_HOME is defined in the runners, so you need to unset it. [65e2375](https://github.com/vim/vim/commit/65e237529e732bdae8b67008f596a9ac902f5b34)
> 
> Thanks . I have updated the PR with this change. Let us see whether the CI tests pass now.

Now the tests are passing with that change.  Thanks.  This PR is ready now.

chrisbra:
thanks!

PR #14565, opened 2024-04-16
[vim9script] handle compiled access to item imported through symbolic linked directory

### LINKED ISSUE #14536: [vim9script] unexpected error accessing items imported through symbolic linked directory
### Steps to reproduce

**NOTE:** for better example, with additional debug output, see https://github.com/vim/vim/issues/14536#issuecomment-2054136374

At the entry of `as_dir/f1.vim` there is

```
vim9script
echomsg 'EXECUTING f1'
echomsg expand('<stack>')
```

There's the following output, slightly edited for clarity.
The important point is that **the import file f1 is loaded twice**.
Which, IIUC, is impossible; `vim` shouldn't load the same import file twice.

```
lib_dir: /home/err/bugs/vim/import-symlink/syml_dir
EXECUTING f1
:source buffer=1[22]
    function <SNR>39_Fmain[1]
    script /home/err/bugs/vim/import-symlink/as_dir/f1.vim[3]
EXECUTING f1
:source buffer=1[22]
    function <SNR>39_Fmain[1]
    script /home/err/bugs/vim/import-symlink/as_dir/f1.vim[3]

Error detected while processing :source buffer=1[22]..function <SNR>39_Fmain:
line    1:
E117: Unknown function: <SNR>40_Ff1
```

Symlinks are involved in the problem.
Note in the source three ways to workaround the problem.
1. `import lib_dir .. '/f1.vim'`
2. `const Ff1 = f1.Ff1`
3. import from the directory, not `syml_dir`: `.. '/as_dir'`

(1) and (2) loads the import before the function `Fmain()` executes. So it
seems like the problem only occurs during a function invocation when it needs
to do an autoload and there is a symlink involved. My hand waving guess is
that when there's a synlink involved it's using the wrong thing to check if
the file has already been loaded. (3) avoids the symlink.

Set up the following files in this structure:
```
.
├── as_dir
│   └── f1.vim
├── main.vim
└── syml_dir -> as_dir
```

**main.vim**
```
vim9script

export var lib_dir: string = fnamemodify(
    getscriptinfo(
        {sid: str2nr(matchstr(expand('<SID>'), '\v\d+'))}
    )[0].name, ':p:h') .. '/syml_dir'

echomsg 'lib_dir:' lib_dir


# REMOVE autoload FROM FOLLOWING LINE AND IT WORKS
import autoload lib_dir .. '/f1.vim'

# UNCOMMENT THE FOLLOWING LINE AND IT WORKS
#const Ff1 = f1.Ff1

def Fmain()
    var xxx = f1.Ff1()
enddef
Fmain()
```

**syml_dir/f1.vim, as_dir/f1.vim**
```
vim9script
echomsg 'EXECUTING f1'
echomsg expand('<stack>')

export const v1val = 1

export def Ff1(): number
    return v1val
enddef
```

### Expected behaviour

No error.

I'm going to look at tracking this down, but I've never looked an imports before...

Any hints?

### Version of Vim

9.1.301

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

errael:
Fix #14536
@yegappan 
If an import is through a symbolic link then at first encounter during compilation create a script_items entry for the real file; and adjust to use the real file for "ISN_SOURCE" and var loading.

### DIFF
--- a/src/proto/scriptfile.pro
+++ b/src/proto/scriptfile.pro
@@ -9,6 +9,7 @@ void ex_runtime(exarg_T *eap);
 void set_context_in_runtime_cmd(expand_T *xp, char_u *arg);
 int find_script_by_name(char_u *name);
 int get_new_scriptitem_for_fname(int *error, char_u *fname);
+void check_script_symlink(int sid);
 int do_in_path(char_u *path, char *prefix, char_u *name, int flags, void (*callback)(char_u *fname, void *ck), void *cookie);
 int do_in_runtimepath(char_u *name, int flags, void (*callback)(char_u *fname, void *ck), void *cookie);
 int source_runtime(char_u *name, int flags);
--- a/src/scriptfile.c
+++ b/src/scriptfile.c
@@ -405,6 +405,43 @@ get_new_scriptitem_for_fname(int *error, char_u *fname)
     return sid;
 }
 
+/*
+ * If the script for "sid" is a symlink and "sn_source_sid" is not set
+ * then initialize it. A new script_item is created if needed.
+ */
+    void
+check_script_symlink(int sid)
+{
+    scriptitem_T *si = SCRIPT_ITEM(sid);
+    if (si->sn_syml_checked || si->sn_sourced_sid > 0)
+	return;
+    si->sn_syml_checked = TRUE;
+
+    // If fname is a symbolic link, create an script_item for the real file.
+
+    char_u *real_fname = fix_fname(si->sn_name);
+    if (real_fname != NULL && STRCMP(real_fname, si->sn_name) != 0)
+    {
+	int real_sid = find_script_by_name(real_fname);
+	int error2 = OK;
+	int new_sid = FALSE;
+	if (real_sid < 0)
+	{
+	    real_sid = get_new_scriptitem_for_fname(&error2, real_fname);
+	    new_sid = TRUE;
+	}
+	if (error2 == OK)
+	{
+	    si = SCRIPT_ITEM(sid);
+	    si->sn_sourced_sid = real_sid;
+	    if (new_sid)
+		SCRIPT_ITEM(real_sid)->sn_import_autoload
+						    = si->sn_import_autoload;
+	}
+    }
+    vim_free(real_fname);
+}
+
     static void
 find_script_callback(char_u *fname, void *cookie)
 {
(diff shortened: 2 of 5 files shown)


yegappan:
Looks good to me.

yegappan (approved):

errael:
@yegappan Minor change https://github.com/vim/vim/compare/39e005a38c48e83c3877c7ee8673bdcf014924b1..6364ac6b24255209f3e9d10059742b9eb4d537c5

Insure script is never checked more than once for sym link.
Adjust test so early return code is executed.

chrisbra:
thanks makes sense.

errael:
@yegappan There's another twist; before this PR, `v9.1.0338`, the following fails. Note, there are **no symbolic links** involved. I think this explains some problems I was having with `autoload ./xxx.vim` last year.

```
vim9script

var lines =<< trim END
    vim9script

    export def F1(): string
        return 'InFile.vim'
    enddef
END
writefile(lines, 'xfile.vim')

import autoload './xfile.vim'

def F(): string
    return xfile.F1()
enddef

try
    echo F()
catch
    echo "FAIL"
    echo v:exception
endtry

delete('xfile.vim')
```
```
FAIL
Vim(echo):E117: Unknown function: <SNR>40_F1
```

errael:
I wanted to put together a test for this. But if I put something like this in a test, it does not fail. Tried the following in test_vim9_import.vim
```
def Test_autoload_import_relative_compiled()
  # autoload relative, access from compiled function. #14565
  var lines =<< trim END
    vim9script

    export def F1(): string
        return 'InFile.vim'
    enddef
  END
  writefile(lines, 'xfile.vim', 'D')
  lines =<< trim END
    vim9script

    import autoload './xfile.vim'

    def F(): string
      return xfile.F1()
    enddef
    assert_equal('xInFile.vim', F())
  END
  v9.CheckScriptSuccess(lines)
enddef
```

errael:
BTW, a simple way to remove the effect of this PR is to comment out the two lines in `compile_load_scriptvar`
```
	check_script_symlink(import->imp_sid);
	import_check_sourced_sid(&import->imp_sid);
```

Shane-XB-Qian:
@yegappan after this, lsp hover failed:
```log
E1048: Item not found in script: ParseMarkdown
```
at line 60 of lspgfm.vim

yegappan:
> @yegappan after this, lsp hover failed:
> 
> ```
> E1048: Item not found in script: ParseMarkdown
> ```
> 
> at line 60 of lspgfm.vim

Thanks for reporting the failure.  I will take a look.

PR #14566, opened 2024-04-16
fix: count tag files on helplang due to tags-ru

Shane-XB-Qian:
fix: https://github.com/vim/vim/pull/14312#issuecomment-2058086070

### DIFF
--- a/src/testdir/test_taglist.vim
+++ b/src/testdir/test_taglist.vim
@@ -99,8 +99,11 @@ func Test_tagfiles()
   call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
 
   help
-  let tf = tagfiles()
-  call assert_equal(1, len(tf))
+  let hplang = &helplang
+  set helplang=en  " May other 'tags-xx' (e.g 'tags-ru') exist
+  let tf = tagfiles()->filter("v:val =~ 'tags\\%(-' .. &helplang .. '\\)\\?$'")
+  call assert_equal((&helplang != '' && &helplang != 'en') ? 2 : 1, len(tf))
+  let &helplang = hplang | unlet hplang
   call assert_equal(fnamemodify(expand('$VIMRUNTIME/doc/tags'), ':p:gs?\\?/?'),
 	\           fnamemodify(tf[0], ':p:gs?\\?/?'))
   helpclose


Shane-XB-Qian:
some system (or type) seems flaky to gen 'tags-ru' in CI, 
// and macos was fail due to something else.
not able or not existed, so/anyway back to 'en' to verify help tags.

src/testdir/test_taglist.vim
```diff
@@ -99,8 +99,11 @@ func Test_tagfiles()
   call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
 
   help
-  let tf = tagfiles()
-  call assert_equal(1, len(tf))
+  let hplang = &helplang
+  set helplang=en  " May other 'tags-xx' (e.g 'tags-ru') exist
+  let tf = tagfiles()->filter("v:val =~ 'tags\\%(-' .. &helplang .. '\\)\\?$'")
+  call assert_equal((&helplang != '' && &helplang != 'en') ? 2 : 1, len(tf))
```
zeertzjq, line 105: `&helplang` is always `'en'` here now.

Shane-XB-Qian: yes, i was trying to add 'ru', but failed, see comment here: https://github.com/vim/vim/pull/14566#issuecomment-2058458220

RestorerZ:
As I said in [this comment](https://github.com/vim/vim/pull/14479#issuecomment-2053995559), building and installing Vim will always create files of type "tags-???" if there are "*.??x" files in the "runtime/doc" directory. There's also a suggestion there as to how I think this can be eliminated.

As for the test, as I understand it, it's simply checking for the presence of a "tags" file.
```
  let tf = tagfiles()
  call assert_equal(1, len(tf))
```
Maybe it could be done like this (if files like "tags-??" are created):
```
if "tags" == findfile("tags", "doc") | echo "ok" | else | echo "not ok" | endif
```
or like this:
```
if 0 == len(tagfiles()) | echo "not ok" |else | echo "ok" | endif 
```

src/testdir/test_taglist.vim
```diff
@@ -99,8 +99,11 @@ func Test_tagfiles()
   call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
 
   help
-  let tf = tagfiles()
-  call assert_equal(1, len(tf))
+  let hplang = &helplang
+  set helplang=en  " May other 'tags-xx' (e.g 'tags-ru') exist
```
Shane-XB-Qian, line 103: @RestorerZ you may need to find out why here some system failed in CI if changed to 'ru'

Shane-XB-Qian (commented):
@RestorerZ you may need to find out why it was failed, or simply truly del tags-ru after make install?
anyway, here this PR is just to make that file not block the test (continue to fail)

src/testdir/test_taglist.vim
RestorerZ: @Shane-XB-Qian 
It's not just about the tags-ru file. If someone, for example, add a file "uganda.dex", then there will be three tag files: "tags" (for English), "tags-ru" (for Russian), "tags-de" (for German).
And this condition will no longer work.
```
call assert_equal((&helplang != '' && &helplang != 'en') ? 2 : 1, len(tf))
```

> @RestorerZ you may need to find out why here some system failed in CI if changed to 'ru'

As I remember, in order for "helplang" to work, it requires that Vim be built with "+multi_lang" support.
Perhaps this happens, for example, when building a "TINY" version.

Shane-XB-Qian: > @Shane-XB-Qian It's not just about the tags-ru file. If someone, for example, add a file "uganda.de", then there will be three tag files: "tags" (for English), "tags-ru" (for Russian), "tags-de" (for German).

interesting, i am cn, not sure why it gen 'ru' to me; and yes, now (this PR) even there 100 tag files, made it only knew 'en'.

> > @RestorerZ you may need to find out why here some system failed in CI if changed to 'ru'
> 
> As I remember, in order for "helplang" to work, it requires that Vim be built with "+multi_lang" support. Perhaps this happens, for example, when building a "TINY" version.

https://github.com/vim/vim/actions/runs/8701135087

---
again, i am not planning to look into why it failed, but just make that file not block the test.

chrisbra:
that doesn't seem right yet. Should it be like this:
```diff
diff --git a/src/testdir/test_taglist.vim b/src/testdir/test_taglist.vim
index 39e78bcb0..7b4556e98 100644
--- a/src/testdir/test_taglist.vim
+++ b/src/testdir/test_taglist.vim
@@ -98,12 +98,26 @@ func Test_tagfiles()
   set tags=Xtags1,Xtags2
   call assert_equal(['Xtags1', 'Xtags2'], tagfiles())

+  let _helplang = &helplang
+  set helplang=
   help
   let tf = tagfiles()
   call assert_equal(1, len(tf))
   call assert_equal(fnamemodify(expand('$VIMRUNTIME/doc/tags'), ':p:gs?\\?/?'),
        \           fnamemodify(tf[0], ':p:gs?\\?/?'))
   helpclose
+  if !empty(_helplang)
+    if _helplang =~ 'en'
+      let length = len(split(_helplang, ','))
+    else
+      " it should search all specified languages + EN
+      let length = len(split(_helplang, ',')) + 1
+    endif
+    let &helplang = _helplang
+    help
+    call assert_equal(length, len(tagfiles()))
+  endif
+  let &helplang = _helplang
   call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
   set tags&
   call assert_equal([], tagfiles())
```

This assumes there is a valid tagfile for each language specified in `'helplang'`. Not sure this is true. Or perhaps we should just verify, that `len >= 1`?

julio-b:
> This assumes there is a valid tagfile for each language specified in `'helplang'`. Not sure this is true. Or perhaps we should just verify, that `len >= 1`?

I was thinking if a patch like this is enough?
```diff
diff --git a/src/testdir/test_taglist.vim b/src/testdir/test_taglist.vim
index 39e78bcb0..f2fbc709f 100644
--- a/src/testdir/test_taglist.vim
+++ b/src/testdir/test_taglist.vim
@@ -100,7 +100,7 @@ func Test_tagfiles()
 
   help
   let tf = tagfiles()
-  call assert_equal(1, len(tf))
+  call assert_inrange(1, 2, len(tf))
   call assert_equal(fnamemodify(expand('$VIMRUNTIME/doc/tags'), ':p:gs?\\?/?'),
        \           fnamemodify(tf[0], ':p:gs?\\?/?'))
   helpclose
```

And then we could just increase the range if any additional translation is added.

chrisbra:
yeah, let me just include this for now.

Shane-XB-Qian:
> that doesn't seem right yet. Should it be like this:
> 
> ```diff
> diff --git a/src/testdir/test_taglist.vim b/src/testdir/test_taglist.vim
> index 39e78bcb0..7b4556e98 100644
> --- a/src/testdir/test_taglist.vim
> +++ b/src/testdir/test_taglist.vim
> @@ -98,12 +98,26 @@ func Test_tagfiles()
>    set tags=Xtags1,Xtags2
>    call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
> 
> +  let _helplang = &helplang
> +  set helplang=
>    help
>    let tf = tagfiles()
>    call assert_equal(1, len(tf))
>    call assert_equal(fnamemodify(expand('$VIMRUNTIME/doc/tags'), ':p:gs?\\?/?'),
>         \           fnamemodify(tf[0], ':p:gs?\\?/?'))
>    helpclose
> +  if !empty(_helplang)
> +    if _helplang =~ 'en'
> +      let length = len(split(_helplang, ','))
> +    else
> +      " it should search all specified languages + EN
> +      let length = len(split(_helplang, ',')) + 1
> +    endif
> +    let &helplang = _helplang
> +    help
> +    call assert_equal(length, len(tagfiles()))
> +  endif
> +  let &helplang = _helplang
>    call assert_equal(['Xtags1', 'Xtags2'], tagfiles())
>    set tags&
>    call assert_equal([], tagfiles())
> ```
> 
> This assumes there is a valid tagfile for each language specified in `'helplang'`. Not sure this is true. Or perhaps we should just verify, that `len >= 1`?

that assumption was not correct, helplang= or helplang=en is same thing, it was I bring it for filter for tagfiles.

Shane-XB-Qian:
> > This assumes there is a valid tagfile for each language specified in `'helplang'`. Not sure this is true. Or perhaps we should just verify, that `len >= 1`?
> 
> I was thinking if a patch like this is enough?
> 
> ```diff
> diff --git a/src/testdir/test_taglist.vim b/src/testdir/test_taglist.vim
> index 39e78bcb0..f2fbc709f 100644
> --- a/src/testdir/test_taglist.vim
> +++ b/src/testdir/test_taglist.vim
> @@ -100,7 +100,7 @@ func Test_tagfiles()
>  
>    help
>    let tf = tagfiles()
> -  call assert_equal(1, len(tf))
> +  call assert_inrange(1, 2, len(tf))
>    call assert_equal(fnamemodify(expand('$VIMRUNTIME/doc/tags'), ':p:gs?\\?/?'),
>         \           fnamemodify(tf[0], ':p:gs?\\?/?'))
>    helpclose
> ```
> 
> And then we could just increase the range if any additional translation is added.

this is a terrible temp workaround, if so, if there is tags-de then truly really there is 3 tags.

Shane-XB-Qian:
my patch (this PR) is letting it be back to original situation which it was only 'en' help doc.
// BTW: your merge seems ignored my patch (and credit) totally, sigh~~~

Shane-XB-Qian:
my patch (this PR) is letting it be back to original situation which it was only 'en' help doc.
// BTW: your merge seems ignored my patch (and credit) totally, sigh~~~

chrisbra:
> >    This assumes there is a valid tagfile for each language specified in 'helplang'. Not sure this is true. Or perhaps we should just verify, that len >= 1?

> that assumption was not correct, helplang= or helplang=en is same thing, it was I bring it for filter for tagfiles.

What assumption? Even with helplang being empty or  equal to `en`, len would be `1`. So this should work no?

> // BTW: your merge seems ignored my patch (and credit) totally, sigh~~~

Correct, because:
- It was not your patch that I merged to master
- as was pointed out, your patch has still [problems](https://github.com/vim/vim/pull/14566#discussion_r1566898669) but you [said, you have no interest in finding out why it failed](https://github.com/vim/vim/pull/14566#discussion_r1567182239)

> my patch (this PR) is letting it be back to original situation which it was only 'en' help doc.

So is the current test, it doesn't assume only en helpfile, but can also handle 1 additional language.

Yes we may need to adjust the test again, once we have more translation of other helpfiles included. But who knows when this will happen. However, I would think we should rather update the test and verify correctly that several tagfiles are loaded *when another language was set for 'helplang'*.

RestorerZ:
> I would think we should rather update the test and verify correctly that several tagfiles are loaded _when another language was set for 'helplang'_.

@chrisbra , @Shane-XB-Qian , the creation of "tags-??" files has nothing to do with the value of the "helplang" option at all. They will always be created if there are files with the extension "*.??x" in the "runtime/doc" directory.

PR #14568, opened 2024-04-16
source gvimrc if XDG_CONFIG_HOME is used

### LINKED ISSUE #14567: gvimrc is not sourced from `~/.config/vim/`
### Steps to reproduce

1. have ~/.config/vim/gvimrc with setting of guifont
2. run `gvim`

### Expected behaviour

Font settings should be applied.

Looks like `gvimrc` is not sourced at all.

image

### Version of Vim

9.1.336

### Environment

debian 12

### Logs and stack traces

_No response_

habamax:
Fixes #14567

Probably needs test case added?

### DIFF
--- a/src/gui.c
+++ b/src/gui.c
@@ -579,6 +579,10 @@ gui_init(void)
 		 && do_source((char_u *)USR_GVIMRC_FILE3, TRUE,
 						     DOSO_GVIMRC, NULL) == FAIL
 #endif
+#ifdef XDG_GVIMRC_FILE
+		 && do_source((char_u *)XDG_GVIMRC_FILE, TRUE,
+						     DOSO_GVIMRC, NULL) == FAIL
+#endif
 				)
 	    {
 #ifdef USR_GVIMRC_FILE4
--- a/src/os_unix.h
+++ b/src/os_unix.h
@@ -281,6 +281,12 @@ typedef struct dsc$descriptor   DESC;
 # endif
 #endif
 
+#ifndef XDG_GVIMRC_FILE
+# define XDG_GVIMRC_FILE mch_getenv("XDG_CONFIG_HOME") \
+	? (char_u *)"$XDG_CONFIG_HOME/vim/gvimrc" \
+	: (char_u *)"~/.config/vim/gvimrc"
+#endif
+
 #ifndef VIM_DEFAULTS_FILE
 # define VIM_DEFAULTS_FILE "$VIMRUNTIME/defaults.vim"
 #endif


chrisbra:
Thanks, yes a test would be helpful. Something similar to [Test_xdg_runtime_files()](https://github.com/vim/vim/blob/2da68c8b61286ed347076718b66028f7251c8cdf/src/testdir/test_xdg.vim#L42C6-L138) in [test_gui.vim](https://github.com/vim/vim/blob/master/src/testdir/test_gui.vim)/[setup_gui.vim](https://github.com/vim/vim/blob/master/src/testdir/setup_gui.vim)?

chrisbra:
I have the following patch now, If you could please have a look into how to check that the gvimrc file is loaded please?
```patch
diff --git a/runtime/doc/gui.txt b/runtime/doc/gui.txt
index 6e0875281..795f1faa2 100644
--- a/runtime/doc/gui.txt
+++ b/runtime/doc/gui.txt
@@ -70,6 +70,8 @@ When the GUI starts up initializations are carried out, in this order:
   - For Win32, $HOME is set by Vim if needed, see |$HOME-windows|.
   - When a "_gvimrc" file is not found, ".gvimrc" is tried too.  And vice
     versa.
+  - On Unix, if "~/.config/vim/gvimrc" or "$XDG_CONFIG_HOME/vim/gvimrc" is
+    exists, it is source.  You can check this with ":version".
   The name of the first file found is stored in $MYGVIMRC, unless it was
   already set.
 - If the 'exrc' option is set (which is NOT the default) the file ./.gvimrc
diff --git a/src/gui.c b/src/gui.c
index 195369102..4180a7b06 100644
--- a/src/gui.c
+++ b/src/gui.c
@@ -578,6 +578,10 @@ gui_init(void)
 #ifdef USR_GVIMRC_FILE3
                 && do_source((char_u *)USR_GVIMRC_FILE3, TRUE,
                                                     DOSO_GVIMRC, NULL) == FAIL
+#endif
+#ifdef XDG_GVIMRC_FILE
+                && do_source((char_u *)XDG_GVIMRC_FILE, TRUE,
+                                                    DOSO_GVIMRC, NULL) == FAIL
 #endif
                                )
            {
diff --git a/src/os_unix.h b/src/os_unix.h
index 67dad2e68..8f9d88e1f 100644
--- a/src/os_unix.h
+++ b/src/os_unix.h
@@ -281,6 +281,12 @@ typedef struct dsc$descriptor   DESC;
 # endif
 #endif

+#ifndef XDG_GVIMRC_FILE
+# define XDG_GVIMRC_FILE (mch_getenv("XDG_CONFIG_HOME") \
+       ? "$XDG_CONFIG_HOME/vim/gvimrc" \
+       : "~/.config/vim/gvimrc")
+#endif
+
 #ifndef VIM_DEFAULTS_FILE
 # define VIM_DEFAULTS_FILE "$VIMRUNTIME/defaults.vim"
 #endif
diff --git a/src/testdir/test_xdg.vim b/src/testdir/test_xdg.vim
index 557864e25..9dc981874 100644
--- a/src/testdir/test_xdg.vim
+++ b/src/testdir/test_xdg.vim
@@ -147,14 +147,22 @@ func Test_xdg_version()
   unlet $XDG_CONFIG_HOME
   let a = execute(':version')->split('\n')
   let a = filter(a, { _, val -> val =~ '\.config\|XDG_CONFIG_HOME' })
-  call assert_equal(1, len(a))
-  call assert_match('\~/.config/vim/vimrc', a[0])
+  " There should be 1 entry for gvimrc and 1 entry for vimrc,
+  " but only if Vim was compiled with gui support
+  call assert_equal(1 + has("gui"), len(a))
+  call assert_match('\~/\.config/vim/vimrc', a[0])
+  if has("gui")
+    call assert_match('\~/\.config/vim/gvimrc', a[1])
+  endif

   let $XDG_CONFIG_HOME = expand('~/.xdg')
   let a = execute(':version')->split('\n')
   let a = filter(a, { _, val -> val =~ '\.config\|XDG_CONFIG_HOME' })
-  call assert_equal(1, len(a))
+  call assert_equal(1 + has("gui"), len(a))
   call assert_match('XDG_CONFIG_HOME/vim/vimrc', a[0])
+  if has("gui")
+    call assert_match('XDG_CONFIG_HOME/vim/gvimrc', a[1])
+  endif
   unlet $XDG_CONFIG_HOME
 endfunc

diff --git a/src/version.c b/src/version.c
index 68b58b0b4..78523c338 100644
--- a/src/version.c
+++ b/src/version.c
@@ -1795,11 +1795,18 @@ list_version(void)
     version_msg(USR_GVIMRC_FILE2);
     version_msg("\"\n");
 # endif
+    // USR_GVIMRC_FILE3 is defined on Amiga,
+    // XDG_GVIMRC_FILE is defined on Unix
 # ifdef USR_GVIMRC_FILE3
     version_msg(_("3rd user gvimrc file: \""));
     version_msg(USR_GVIMRC_FILE3);
     version_msg("\"\n");
 # endif
+# ifdef XDG_GVIMRC_FILE
+    version_msg(_("3rd user gvimrc file: \""));
+    version_msg(XDG_GVIMRC_FILE);
+    version_msg("\"\n");
+# endif
 #endif
     version_msg(_("       defaults file: \""));
     version_msg(VIM_DEFAULTS_FILE);
```

chrisbra:
might not be perfect yet, the USR_GVIMRC_FILE3

habamax:
> I have the following patch now: https://github.com/chrisbra/vim/commit/ba89d66bc6fb6567c4784a62847cea7c4b061b3e, If you could please have a look into how to check that the gvimrc file is loaded please?

Looks fine, gvimrc is loaded

chrisbra:
Okay, let me try to find a way to test this change for GUI env.

PR #14569, opened 2024-04-16
Missing XDG vimrc entry in :version

diegoviola:
related: #14182

### DIFF
--- a/src/version.c
+++ b/src/version.c
@@ -1743,9 +1743,20 @@ list_version(void)
     version_msg(USR_VIMRC_FILE2);
     version_msg("\"\n");
 #endif
-#ifdef USR_VIMRC_FILE3
+#if defined(USR_VIMRC_FILE3) && defined(XDG_VIMRC_FILE)
+# ifdef USR_VIMRC_FILE3
     version_msg(_(" 3rd user vimrc file: \""));
     version_msg(USR_VIMRC_FILE3);
+    version_msg("\"\n");
+# endif
+# ifdef XDG_VIMRC_FILE
+    version_msg(_(" 4th user vimrc file: \""));
+    version_msg((char *)(XDG_VIMRC_FILE));
+    version_msg("\"\n");
+# endif
+#elif defined(XDG_VIMRC_FILE)
+    version_msg(_(" 3rd user vimrc file: \""));
+    version_msg((char *)(XDG_VIMRC_FILE));
     version_msg("\"\n");
 #endif
 #ifdef USR_EXRC_FILE


diegoviola:
Not sure why the tests fail here, it compiles fine locally for me.

src/version.c
```diff
@@ -1748,6 +1748,11 @@ list_version(void)
     version_msg(USR_VIMRC_FILE3);
     version_msg("\"\n");
 #endif
+#ifdef XDG_VIMRC_FILE
+    version_msg(_(" 4th user vimrc file: \""));
+    version_msg(XDG_VIMRC_FILE);
```
yegappan: You need a `(char *)` cast for XDG_VIMRC_FILE.

diegoviola: @yegappan Thanks!

diegoviola: Still failing, not sure.

diegoviola: Looks like it's passing now, had to compile with `CC="gcc -Werror -Wpointer-sign"` to be able to reproduce the errors locally.

chrisbra:
That looks strange IMHO:
```
   system vimrc file: "$VIM/vimrc"
     user vimrc file: "$HOME/.vimrc"
 2nd user vimrc file: "~/.vim/vimrc"
 4th user vimrc file: "$XDG_CONFIG_HOME/vim/vimrc"
      user exrc file: "$HOME/.exrc"
  system gvimrc file: "$VIM/gvimrc"
    user gvimrc file: "$HOME/.gvimrc"
2nd user gvimrc file: "~/.vim/gvimrc"
       defaults file: "$VIMRUNTIME/defaults.vim"
    system menu file: "$VIMRUNTIME/menu.vim"
  fall-back for $VIM: "/usr/local/share/vim"
```

It seems the `3rd` vimrc file is usually not defined on Unix. How about this patch instead:
```diff
diff --git a/src/version.c b/src/version.c
index f3c46e55e..4e5cfb960 100644
--- a/src/version.c
+++ b/src/version.c
@@ -1745,15 +1745,21 @@ list_version(void)
     version_msg(USR_VIMRC_FILE2);
     version_msg("\"\n");
 #endif
-#ifdef USR_VIMRC_FILE3
+#if defined(USR_VIMRC_FILE3) && defined(XDG_VIMRC_FILE)
+# ifdef USR_VIMRC_FILE3
     version_msg(_(" 3rd user vimrc file: \""));
     version_msg(USR_VIMRC_FILE3);
     version_msg("\"\n");
-#endif
-#ifdef XDG_VIMRC_FILE
+# endif
+# ifdef XDG_VIMRC_FILE
     version_msg(_(" 4th user vimrc file: \""));
     version_msg((char *)(XDG_VIMRC_FILE));
     version_msg("\"\n");
+# endif
+#elif defined(XDG_VIMRC_FILE)
+    version_msg(_(" 3rd user vimrc file: \""));
+    version_msg((char *)(XDG_VIMRC_FILE));
+    version_msg("\"\n");
 #endif
 #ifdef USR_EXRC_FILE
     version_msg(_("      user exrc file: \""));
```

chrisbra:
at one point, we may also think about replacing `~` by `$HOME` in the various `USR_VIMRC_FILE` and `XDG_VIMRC_FILE`. Currently it seems to be a strange mixture between both.

diegoviola:
> How about this patch instead:

Looks good, I pushed your change.

diegoviola:
@chrisbra how do you feel about this advice? I received it in #c (irc.libera.chat):

```
jp │ dviola: you should add parenthesis around the ternary in the #define and remove the pointless (char_u *) casts so you don't need a (char *) cast at all
jp │ dviola: the (char_u *) casts in the #define are not needed since there is already a cast here: https://github.com/vim/vim/blob/2da68c8b61286ed347076718b66028f7251c8cdf/src/main.c#L3280
jp │ which confusingly does the wrong thing since there are no parenthesis originally in that #define
jp │ it pointlessly casts the result of mch_getenv("XDG_CONFIG_HOME")
```

diegoviola:
@chrisbra your changes look great, thank you!

chrisbra:
just waiting for CI to complete and then push it out.

PR #14571, opened 2024-04-16
Fixes creation of tags-?? tag files and failure test

RestorerZ:
@Shane-XB-Qian  suggested a workaround for the solution in #14566  only for
cases where there are only two tag files, English and some other language. Here
propose a solution that is independent of the number of tag files and retains
the logic of checking if there are tag files for the help buffer in the
"test_tagfiles" test from the "test_taglist.vim" file.

A solution is also offered when installing compiled Vim.  
Tag files of languages other than English and files "uganda.??x" are excluded
from copying. This condition applies only if there are no other files with
translations of the documentation, so that the logic of work provided by the
developers of the program is not violated.


RestorerZ:
Ah, I see that there is already a fix.
Well, this is also one of the solutions.  But now we just need to remember to make the appropriate corrections.

src/testdir/test_taglist.vim
```diff
@@ -100,10 +100,7 @@ func Test_tagfiles()
 
   help
   let tf = tagfiles()
-  " if 'helplang includes another language, then we may find
-  " 2 tagfiles (e.g.: for EN and RU)
-  " we may need to adjust this, if further translated help files are included
-  call assert_inrange(1, 2, len(tf))
+  call assert_notequal(0, len(tf))
```
Shane-XB-Qian, line 103: this `notequal` seems meanlingless, it always at least 1.

src/Makefile
```diff
@@ -2355,9 +2355,13 @@ installrtbase: $(HELPSOURCE)/vim.1 $(DEST_VIM) $(VIMTARGET) $(DEST_RT) \
 	# thus use the executable in the current directory.
 	-@BUILD_DIR="`pwd`"; cd $(HELPSOURCE); if test -z "$(CROSS_COMPILING)"; then \
 		$(MAKE) VIMEXE="$$BUILD_DIR/$(VIMTARGET)" vimtags; fi
+	# TODO: Make a check for the language.
+	# TODO: Copy "uganda.??x" and "tags-??" depending on the language.
 	cd $(HELPSOURCE); \
 		files=`ls *.txt tags`; \
+		if test `find -maxdepth 1 -type f -name '*.??x' ! -name 'uganda.??x' | wc -l` -gt 0 ; then \
 		files="$$files `ls *.??x tags-?? 2>/dev/null || true`"; \
+		fi; \
```
Shane-XB-Qian, line 2364: now that *.rux at least was there, so this check perhaps is meaningless too.

Shane-XB-Qian:
you may need to find out why it was failed, or simply just remove those 'tags-??' after make install?
// anyway, i probably would be not taking part in this topic, you can think the solution or workaround by you own now.

chrisbra:
I don't think changing the Makefile is the right solution. I think the test should rather handle several tagfiles correctly, as I tried [here](https://github.com/vim/vim/pull/14566#issuecomment-2059839982)

RestorerZ:
@chrisbra , of course it's up to you to accept this pull request or not.  It was just another solution to the situation that had arisen. If the existing solution completely solves the problem that has arisen, then I'll just close this request.

RestorerZ:
Resolved in the [patch 9.1.0342](https://github.com/vim/vim/commit/ae7e61c928b4adaa220e59ecebc75ef630674207)

Shane-XB-Qian:
it resolved nothing, just a workaround, whatever better or wrong, and this PR was clear wrong, please check my review comment.

maybe really need to review that tags-ru gen in the makefile, it had made so much issues.

--
shane.xb.qian

RestorerZ:
I have read your comments very carefully. It is possible that to you, as a
specialist, this request seems to be erroneous, but look:
in the original test, when opening the help buffer, the presence of a tag file
was checked and an error was returned if the file was missing. I.e., the length
of the array was zero, not one. After the appearance of the "tags-??" files, the
error began to be issued because the length of the array is greater than one.
The solution I proposed will also give an error if there are no tag files
"Expected not equal to 0". Yes, it's probably better to add a more
understandable error message.

> this notequal seems meanlingless, it always at least 1.

Why do you say it will always return one? Run `:echo assert_notequal(0,
len(tagfiles()))`
on an empty buffer and it will return 1. And then the same command in the help buffer and 0 will be returned.
Or am I misunderstanding something?

> you may need to find out why it was failed, or simply just remove those 'tags-??' after make install?

Where do you suggest deleting the "tags-??" files?
They will not be copied to the destination directory, nor will the "uganda.??x"
files. If you accept my suggested solution, of course.

> maybe really need to review that tags-ru gen in the makefile, it had made so much issues.

I [suggested using](https://github.com/vim/vim/pull/14479#issuecomment-2053995559) the target "tags" instead of "vimtags" when calling make in
the "runtime/doc" directory. Then the "tags-??" files will not be created. But
Bram chose the "vimtags" target for some reason, so it's probably best not to
change it yet.

The solution proposed by @chrisbra , in principle, also solves the problem of
test failure. The main thing is not to forget to update it when adding new
translations. But, by the looks of the trend, this will not be necessary very
soon.

PR #14572, opened 2024-04-17
'showcmd' wrong for partial mapping with multibyte

zeertzjq:
Problem:  'showcmd' is wrong for partial mapping with multibyte char,
          and isn't very readable with modifyOtherKeys.
Solution: Decode multibyte char and merge modifiers into the char.

This improves the following situations:
- Multibyte chars whose individual bytes are considered unprintable are
  now shown properly in 'showcmd' area.
- Ctrl-W with modifyOtherKeys now shows ^W in 'showcmd' area.

The following situation may still need improvement:
- If the char is a special key or has modifiers that cannot be merged
  into it, internal keycodes are shown in 'showcmd' area like before.
  This applies to keys typed in Normal mode commands as well, and it's
  hard to decide how to make it more readable due to the limited space
  taken by 'showcmd', so I'll leave it for later.

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -1281,6 +1281,80 @@ del_typebuf(int len, int offset)
 	typebuf.tb_change_cnt = 1;
 }
 
+typedef struct
+{
+    int		prev_c;
+    char_u	buf[MB_MAXBYTES * 3 + 4];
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #14575, opened 2024-04-17
runtime(uci): add support

cmcaine:
Updated version of #14385

This version adds content detection for files that have a line starting with "c \S", "config \S" or "p \S" or "package \S" in the first 3 lines.

The detection only checks the second line if the first line looked like it was blank or a comment. Similarly for line 3.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1298,9 +1298,9 @@ export def FTvba()
   endif
 enddef
 
-export def Detect_UCI_statements()
-  " Match a config or package statement at the start of the line.
-  let pattern = '^\s*\(\(c\|config\)\|\(p\|package\)\)\s\+\S'
+export def Detect_UCI_statements(): bool
+  # Match a config or package statement at the start of the line.
+  const pattern = '^\s*\(\(c\|config\)\|\(p\|package\)\)\s\+\S'
   return getline(1) =~# pattern || getline(2) =~# pattern || getline(3) =~# pattern
 enddef
 
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1298,5 +1298,11 @@ export def FTvba()
   endif
 enddef
 
+export def Detect_UCI_statements()
+  " Match a config or package statement at the start of the line.
+  let pattern = '^\s*\(\(c\|config\)\|\(p\|package\)\)\s\+\S'
+  return getline(1) =~# pattern || getline(2) =~# pattern || getline(3) =~# pattern
+enddef
+
 # Uncomment this line to check for compilation errors early
 # defcompile
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1300,8 +1300,18 @@ enddef
 
 export def Detect_UCI_statements(): bool
   # Match a config or package statement at the start of the line.
-  const pattern = '^\s*\(\(c\|config\)\|\(p\|package\)\)\s\+\S'
-  return getline(1) =~# pattern || getline(2) =~# pattern || getline(3) =~# pattern
+  const config_or_package_statement = '^\s*\(\(c\|config\)\|\(p\|package\)\)\s\+\S'
+  # Match a line that is either all blank or blank followed by a comment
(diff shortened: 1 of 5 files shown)


runtime/ftplugin/uci.vim
```diff
@@ -0,0 +1,21 @@
...
+let b:did_ftplugin = 1
+
+" UCI files are indented with tabs.
+setl noexpandtab
+setl shiftwidth=0
+setl softtabstop=0
+
+setl commentstring=#\ %s
+
+let b:undo_ftplugin = "setlocal et< cms<"
```
chrisbra: missing shiftwidth and softatbstop

runtime/autoload/dist/ft.vim
```diff
@@ -1298,5 +1298,21 @@ export def FTvba()
...
+  # Match a line that is either all blank or blank followed by a comment
+  const comment_or_blank = '^\s*\(#.*\)\?$'
+
+  # Return true iff the file has a config or package statement near the
+  # top of the file and all preceding lines were comments or blank.
+  return getline(1) =~# config_or_package_statement
+  \   || getline(1) =~# comment_or_blank
+  \      && (   getline(2) =~# config_or_package_statement
+  \          || getline(2) =~# comment_or_blank
+  \             && getline(3) =~# config_or_package_statement
```
chrisbra, line 1313: is this okay to only check the first 3 lines? The logic is a bit ambigious: `expr || expr && expr`.

cmcaine: What's ambiguous? && has higher precedence than ||, so the meaning of the code is well defined.

> is this okay to only check the first 3 lines?

I don't understand this question either, sorry. Can you rephrase it, please?

runtime/ftplugin/uci.vim
cmcaine:
```suggestion
let b:undo_ftplugin = "setlocal et< cms< sts< sw<"
```

runtime/autoload/dist/ft.vim
chrisbra: > What's ambiguous? && has higher precedence than ||, so the meaning of the code is well defined.

It might be well-defined, but I prefer to be explict, so I don't have to remember it

chrisbra: > I don't understand this question either, sorry. Can you rephrase it, please?

Is it enough to only check the first 3 lines? Can't there be 20 line comments first, and only then a `config_or_package_statement`?

cmcaine: && has higher precedence in every programming language I remember using it in, but sure, I don't really care if there are extra parentheses in there or not. Are you requesting that I add them?

The heuristic can be wrong if it checks 0 or 1000 lines. 3 lines was chosen arbitrarily. E.g. it might be better if a blank file with path `sshfs/router1/etc/config/firewall` gets UCI syntax highlighting, but the current detection code would not assign it the uci filetype because it is trying to avoid false positives at the expense of having more false negatives.

Honestly, I regret mentioning that the path matches perhaps too many files 🥲

chrisbra: alright fine.

src/testdir/test_filetype.vim
```diff
@@ -764,6 +764,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     typespec: ['file.tsp'],
     ungrammar: ['file.ungram'],
     uc: ['file.uc'],
+    uci: ['any/etc/config/file'],
```
chrisbra: that Test fails. I think you need to create a file with specific content to make this work, see e.g. Test_def_file()

cmcaine: I think this is now resolved.

chrisbra:
Thanks, let me include it now

cmcaine:
Thanks!

Shane-XB-Qian:
> && has higher precedence in every programming language I remember using it in, but sure, I don't really care if there are extra parentheses in there or not.

@cmcaine unfortunately it is not always true if there was a bug :-) you check there is a bug seems last month tho that's for vim9.
just FYI, have fun anyway.

--
shane.xb.qian

cmcaine:
Lol, well, hopefully somebody has added a regression test for that now :)

PR #14580, opened 2024-04-18
No test that completing a partial mapping clears 'showcmd'

zeertzjq:
Problem:  No test that completing a partial mapping clears 'showcmd'.
Solution: Complete partial mappings in Test_showcmd_part_map() instead
          of using :echo.  Adjust some comments.


chrisbra:
thanks!

PR #14579, opened 2024-04-17
for `Test_autoload_import_relative_compiled()`, left out `:bw!` when copying example.

errael:
Add `:bw!` as in similar uses.
Disable test on windows; at least until the reason for the failure is discovered.

PR #14582, opened 2024-04-18
Refactor the typval2type() function and add a static type for typealias

yegappan:


errael:
@yegappan Could you clarify what the file naming convention with Windows?

yegappan:
> @yegappan Could you clarify what the file naming convention with Windows?

The test failure is caused by using import in a script which is sourced from a buffer (not stored in a file).
This Vim bug is specific to MS-Windows and needs to be fixed.

I changed the name of the file so that is it unique.  Bram spent considerable effort to make the
filenames used in tests unique so that the tests can be run in parallel.

errael:
> > @yegappan Could you clarify what the file naming convention with Windows?
> 
> The test failure is caused by using import in a script which is sourced from a buffer (not stored in a file). This Vim bug is specific to MS-Windows and needs to be fixed.
> 
> I changed the name of the file so that is it unique. Bram spent considerable effort to make the filenames used in tests unique so that the tests can be run in parallel.

I put this together the following  while your response was sent. So we agree that #14579 is the right fix for now?

See https://github.com/vim/vim/pull/14577 OP in the original PR; it is only the test in question. Note how to reproduce the test "manually", not part of `vim`'s test infrastructure. In that PR, the use of
```
new
setline(1, lines)
:source
```
is intentional. Using `v9.CheckScriptSuccess(lines)` does not reproduce the error.

Your proposed change to the test `Test_autoload_import_relative_compiled()` is what the the test looked like originally, but it doesn't fail; it is like using `:source fname`, but for the test to fail, you must do `:source` to execute the buffer.

The PR I put together to get rid the windows failure, https://github.com/vim/vim/pull/14579, only runs the test on unix; until the cause of the failure on windows can be sorted out. Someone else will have to investigate the windows failure, I no longer run windows. My thought is to open an issue showing how the test doesn't work on windows and hope someone will address it.

The easiest way to get vim into the state before #14565, is to modify compile_load_scriptvar() in vim9expr.c and comment out the lines
```
check_script_symlink(import->imp_sid);
import_check_sourced_sid(&import->imp_sid);
```
Then a test for the issue should fail; using `v9.CheckScriptSuccess(lines)` the **test does not fail**.


NOTE: the longer names you used certainly is the right thing to do.

yegappan:
> > > @yegappan Could you clarify what the file naming convention with Windows?
> > 
> > 
> > The test failure is caused by using import in a script which is sourced from a buffer (not stored in a file). This Vim bug is specific to MS-Windows and needs to be fixed.
> > I changed the name of the file so that is it unique. Bram spent considerable effort to make the filenames used in tests unique so that the tests can be run in parallel.
> 
> I put this together the following while your response was sent. So we agree that #14579 is the right fix for now?
>

Yes.  I will revert the changes to this test in my PR.

> 
> See #14577 OP in the original PR; it is only the test in question. Note how to reproduce the test "manually", not part of `vim`'s test infrastructure. In that PR, the use of
> 
> ```
> new
> setline(1, lines)
> :source
> ```
> 
> is intentional. Using `v9.CheckScriptSuccess(lines)` does not reproduce the error.
> 
> Your proposed change to the test `Test_autoload_import_relative_compiled()` is what the the test looked like originally, but it doesn't fail; it is like using `:source fname`, but for the test to fail, you must do `:source` to execute the buffer.
> 
> The PR I put together to get rid the windows failure, #14579, only runs the test on unix; until the cause of the failure on windows can be sorted out. Someone else will have to investigate the windows failure, I no longer run windows. My thought is to open an issue showing how the test doesn't work on windows and hope someone will address it.
> 
> The easiest way to get vim into the state before #14565, is to modify compile_load_scriptvar() in vim9expr.c and comment out the lines
> 
> ```
> check_script_symlink(import->imp_sid);
> import_check_sourced_sid(&import->imp_sid);
> ```
> 
> Then a test for the issue should fail; using `v9.CheckScriptSuccess(lines)` the **test does not fail**.
> 
> NOTE: the longer names you used certainly is the right thing to do.

Can you update your PR to include the updated file name?

errael:
> Can you update your PR to include the updated file name?

Right, in progress.

chrisbra:
thanks let me merge it. Regarding the failing Windows test, I'll will have a look later.

errael:
> thanks let me merge it. Regarding the failing Windows test, I'll will have a look later.

@chrisbra #14579 addresses the failing Windows test. (There's follow up work as we..)

PR #14586, opened 2024-04-18
Page scrolling should place cursor at window boundaries

### LINKED ISSUE #14585: Whole-page scrolling is off by one line
### Steps to reproduce

1. Open Vim 9.1.0346 to a buffer of 1000 lines: `git checkout v9.1.0346 && make -j && make install prefix=$HOME/myvim && seq 1000 | ~/myvim/bin/vim --clean -`
2. Press CTRL-F CTRL-B

### Expected behaviour

Same behavior as **v9.1.0210**: After pressing CTRL-F, the cursor is at first line of the screen, and after pressing CTRL-B, the cursor is at the last line of the screen (in my case, line 40 on a window that shows 40 lines)

Actual behavior since **v9.1.0211**: After pressing CTRL-F, the cursor is at first line of the screen, and after pressing CTRL-B, the cursor is at the **second-to-last line** of the screen (in my case, line 39 on a window that shows 40 lines)

### Version of Vim

9.1.0346

### Environment

Arch Linux, GNOME Terminal

### Logs and stack traces

_No response_

luukvbaal:
Problem:  Page scrolling does not always place the cursor at the top or
          bottom of the window.
Solution: Place the cursor at the top or bottom of the window.

Fix #14585

### DIFF
--- a/src/move.c
+++ b/src/move.c
@@ -3244,17 +3244,6 @@ pagescroll(int dir, long count, int half)
 	    cursor_down_inner(curwin, curscount);
 	else
 	    cursor_up_inner(curwin, curscount);
-
-	if (get_scrolloff_value() > 0)
-	    cursor_correct();
-#ifdef FEAT_FOLDING
-	// Move cursor to first line of closed fold.
-	foldAdjustCursor();
-#endif
-
-	nochange = nochange
-	    && prev_col == curwin->w_cursor.col
-	    && prev_lnum == curwin->w_cursor.lnum;
     }
     else
     {
@@ -3262,7 +3251,22 @@ pagescroll(int dir, long count, int half)
 	count *= ((ONE_WINDOW && p_window > 0 && p_window < Rows - 1) ?
 				MAX(1, p_window - 2) : get_scroll_overlap(dir));
 	nochange = scroll_with_sms(dir, count);
+
+	// Place cursor at top or bottom of window.
+	validate_botline();
+	curwin->w_cursor.lnum = (dir == FORWARD ? curwin->w_topline
+						    : curwin->w_botline - 1);
     }
+
+    if (get_scrolloff_value() > 0)
+	cursor_correct();
+#ifdef FEAT_FOLDING
+    // Move cursor to first line of closed fold.
+    foldAdjustCursor();
+#endif
+    nochange = nochange
+	&& prev_col == curwin->w_cursor.col
+	&& prev_lnum == curwin->w_cursor.lnum;
 
     // Error if both the viewport and cursor did not change.
     if (nochange)
--- a/src/testdir/test_normal.vim
+++ b/src/testdir/test_normal.vim
@@ -3813,8 +3813,8 @@ func Test_normal_vert_scroll_longline()
   call assert_equal(11, line('.'))
   call assert_equal(1, winline())
   exe "normal \<C-B>"
-  call assert_equal(10, line('.'))
-  call assert_equal(4, winline())
+  call assert_equal(11, line('.'))
+  call assert_equal(5, winline())
   exe "normal \<C-B>\<C-B>"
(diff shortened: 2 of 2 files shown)


luukvbaal:
Should it? Even on 9.1.0210 this did not always happen so this change seems a bit arbitrary. As was the behavior pre-9.1.0210. Checking the behavior from `Test_normal_vert_scroll_longline` on older versions shows that the cursor is not always at the botline and the scroll amount seemed arbitrary. 

This is one way to at least make it consistent in a way that seems expected by users(#14585)? But if it is not fully backwards compatible anyways I'm not sure what's the point. Perhaps we need two separate code paths for 'smoothscroll' and backward compatible behavior (which would also reintroduce some bugs). Or come to terms with a new behavior. Be it this or just applying the half-page behavior to all scroll keys.

chrisbra:
Thanks. I know I would prefer a slightly more intuitive behaviour, but it seems it may causes issues with people getting used to the old (buggy?) behaviour. However not sure if it is worth it, introducing two code-paths, since both have to be maintained somehow.

luukvbaal:
Perhaps I framed it a bit biased, I think this patch is fine assuming the expected behavior is correct. I can try to decypher where the old behavior not matching this expected behavior was coming from, and try to restore that as well. But if the old behavior was inconsistent in that sense that feels a bit counter productive.

Having said that I think at this point I would also prefer if we could just apply the half-page scrolling cursor logic to all scroll keys. I think Ctrl-D/U are the more popular keys anyways so perhaps users are more used to their behavior (hence why they were quick to comment on it breaking, or perhaps that was just not the case for full-page keys because it's behavior was closer to the old). Aditionally I think we could also remove all 'smoothscroll'cursor logic from the inner scrolling functions if we were to do that...

chrisbra:
thanks!

zeertzjq:
It seems that this wasn't always the case. When using Ctrl-B near the top of the buffer and there is less than a full page to scroll, the old behavior was moving cursor down by one line, not to the bottom of the window.

luukvbaal:
Well yes, that was already mentioned. I guess your feelings on "But if the old behavior was inconsistent in that sense that feels a bit counter productive." are then: yes but should be done anyways?

luukvbaal:
The old behavior when scrolling up near the start of the buffer just seems like another unintended behavior due to oversight rather than intentional. It works contrary to what happens when a full page can be scrolled, rather than the cursor moving closer to the bottom of the window it moves closer to the top.

I think this yields the old behavior but restoring that seems rather pointless to me:

```diff
@@ -3248,17 +3248,21 @@ pagescroll(int dir, long count, int half)
     }
     else
     {
+       linenr_T prev_topline = curwin->w_topline;
        // Scroll [count] times 'window' or current window height lines.
        count *= ((ONE_WINDOW && p_window > 0 && p_window < Rows - 1) ?
                                MAX(1, p_window - 2) : get_scroll_overlap(dir));
        nochange = scroll_with_sms(dir, count);

        // Place cursor at top or bottom of window.
-       if (curwin->w_topline != prev_topline)
+       if (dir == FORWARD)
+           curwin->w_cursor.lnum = curwin->w_topline;
+       else if (curwin->w_topline == 1)
+           curwin->w_cursor.lnum = prev_topline + 1;
+       else
        {
            validate_botline();
-           curwin->w_cursor.lnum = (dir == FORWARD ? curwin->w_topline
-                                                   : curwin->w_botline - 1);
+           curwin->w_cursor.lnum = curwin->w_botline - 1;
        }
     }
```

zeertzjq:
Hmm, the old behavior also seems inconsistent when the 'window' option is set a to smaller value than its default one. In that case, Ctrl-F moves cursor to the top of the new viewport, but Ctrl-B doesn't move cursor to the bottom of the new viewport.

PR #14587, opened 2024-04-18
Problem: MS-Windows: --remote may change working directory

### LINKED ISSUE #14549: 'gvim --remote' erroneously changes the current directory when the 'shellslash' option is enabled
### Steps to reproduce

1.  Run gvim process with 'shellslash' option enabled. (I use compiled program directly from vim\src directory)
```
> gvimd.exe --clean --servername GVIMD --cmd 'set ssl'
```
2. Check current directory using commands `:pwd<CR>` or `:echo getcwd()<CR>`. Note that the result uses forward slashes.
3. Open another terminal, change to some directory and then run a second gvim process with the file using `--remote` command line switch.
```
> cd C:\tmp
> full\path\to\vim\src\gvimd.exe --clean --servername GVIMD --remote test.txt
```
4.  Double-check current directory using commands `:pwd<CR>` or `:echo getcwd()<CR>`. Now the current directory has erroneously changed to C:/tmp, as if the 'autochdir' option had been enabled.

### Expected behaviour

The value of the current directory in step 4 should be exactly the same as in step 2. In other words, loading additional files into the editor via '--remote' should not automatically change the current directory (unless the 'autochdir' option is enabled).

### Version of Vim

9.1.318

### Environment

OS: Windows 11
Terminal: Windows Terminal
$TERM: unset
Shell: cmd (x64 Native Tools Command Prompt for VS 2022 is used for building/debugging purposes)

### Logs and stack traces

In the debugger, configure program arguments to '--clean --servername GVIMD --cmd "set ssl"' and set a breakpoint on the 'server_to_input_buf' function. When loading an additional file via '--remote' the value of 'str' parameter should look like this:

```
<C-\><C-N>:cd C:\tmp<CR><C-\><C-N>:let g:_wig=&wig|set wig=<CR>:if exists('*inputsave')|call inputsave()|endif|drop test.txt|if exists('*inputrestore')|call inputrestore()|endif<CR><C-\><C-N>:if !exists('+acd')||!&acd|if haslocaldir()|cd -|lcd -|elseif getcwd() ==# 'C:\tmp'|cd -|endif|endif<CR><C-\><C-N>:let &wig=g:_wig|unlet g:_wig<CR>:cal foreground()|if &im|star|en|redr|f<CR>
```

The problem appears to be a comparison of 'getcwd() ==# 'C:\tmp'' where the value on the right uses backslashes, while getcwd() under the effect of the 'shellslash' option uses forward slashes. Due to inaccurate checking, the reverse directory change is skipped.

See `build_drop_cmd` function at https://github.com/vim/vim/blob/4b5cd7257ee99384940d5210cf50298ff925924e/src/clientserver.c#L557

chrisbra:
Problem:  MS-Windows: --remote may change working directory with 'shellslash'
Windows:  normalize directory separators

Note: Tests does not run on Windows, so it is kind of useless :( 

fixes: #14549

### DIFF
--- a/src/clientserver.c
+++ b/src/clientserver.c
@@ -608,7 +608,7 @@ build_drop_cmd(
 
     // Call inputsave() so that a prompt for an encryption key works.
     ga_concat(&ga, (char_u *)
-		       "<CR>:if exists('*inputsave')|call inputsave()|endif|");
+	    "<CR><C-\\><C-N>:if exists('*inputsave')|call inputsave()|endif|");
     if (tabs)
 	ga_concat(&ga, (char_u *)"tab ");
     ga_concat(&ga, (char_u *)"drop");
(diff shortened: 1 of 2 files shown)


Jajauma:
The title probably should say 'Solution: normalize directory separators' not 'Windows: normalize directory separators'?

chrisbra:
Yes of course 🤦

PR #14593, opened 2024-04-19
runtime(doc): minor fixes to starting.txt

diegoviola:


chrisbra:
thanks!

chrisbra:
reading it again, is it:
```
Vim will only care of the $XDG_CONFIG_HOME directory
```
or is it rather *"care for the $XDG_CONFIG_HOME"*?

diegoviola:
> or is it rather _"care for the $XDG_CONFIG_HOME"_?

Good question, I'm really not a English native speaker.

I did ask though and suggestions were to use "use" instead, such as, "_Vim will only use $XDG_CONFIG_HOME_".

Personally, I think _for_ makes sense, but again, my English is probably not the best to judge the sentence.

diegoviola:
> or is it rather _"care for the $XDG_CONFIG_HOME"_?

Good question, I'm really not a native English speaker.

I did ask though and suggestions were to use "use" instead, such as, "_Vim will only use $XDG_CONFIG_HOME_".

Personally, I think _for_ makes sense, but again, my English is probably not the best to judge the sentence.

chrisbra:
Okay, let me merge it with `use` then.

PR #14598, opened 2024-04-19
Refactor to remove STRLEN() part 4

basilisk0315:
Refactor ops.c to remove STRLEN() calls.


src/ops.c
```diff
@@ -369,15 +372,17 @@ shift_block(oparg_T *oap, int amount)
...
 	mch_memmove(newp, oldp, (size_t)bd.textcol);
-	vim_memset(newp + bd.textcol, TAB, (size_t)tabs);
-	vim_memset(newp + bd.textcol + tabs, ' ', (size_t)spaces);
-	// Note that STRMOVE() copies the trailing NUL.
-	STRMOVE(newp + bd.textcol + tabs + spaces, bd.textstart);
+	newlen = bd.textcol;
+	vim_memset(newp + newlen, TAB, (size_t)tabs);
+	newlen += tabs;
+	vim_memset(newp + newlen, ' ', (size_t)spaces);
+	vim_strncpy(newp + newlen + spaces, bd.textstart, shiftedlen);
```
chrisbra: are we not missing the trailing NUL now?

src/ops.c
```diff
@@ -463,21 +471,28 @@ shift_block(oparg_T *oap, int amount)
...
 	if (newp == NULL)
 	    return;
-	mch_memmove(newp, oldp, (size_t)(verbatim_copy_end - oldp));
-	vim_memset(newp + (verbatim_copy_end - oldp), ' ', (size_t)fill);
-	// Note that STRMOVE() copies the trailing NUL.
-	STRMOVE(newp + (verbatim_copy_end - oldp) + fill, non_white);
+	mch_memmove(newp, oldp, fixedlen);
+	newlen = fixedlen;
+	vim_memset(newp + newlen, ' ', (size_t)fill);
+	vim_strncpy(newp + newlen + fill, non_white, shiftedlen);
```
chrisbra: same here

src/ops.c
```diff
@@ -809,8 +830,9 @@ op_delete(oparg_T *oap)
 	    vim_memset(newp + bd.textcol, ' ',
 				     (size_t)(bd.startspaces + bd.endspaces));
 	    // copy the part after the deleted part
-	    oldp += bd.textcol + bd.textlen;
-	    STRMOVE(newp + bd.textcol + bd.startspaces + bd.endspaces, oldp);
+	    vim_strncpy(newp + bd.textcol + bd.startspaces + bd.endspaces,
+				    oldp + bd.textcol + bd.textlen,
+				    oldlen - (bd.textcol + bd.textlen));
```
chrisbra: same here

chrisbra:
I am a bit worried about the `vim_strncpy` vs `strmove()`, otherwise looks good.

basilisk0315:
In vim_strncpy() (like strncpy()), the 'n' argument is the number of characters to copy. These functions expect the destination buffer will be large enough to accommodate at least 'n' characters plus 1 for the NUL.
Unlike strncpy(), vim_strncpy() guarantees the destination is terminated with a trailing NUL.

Having said all that I think I can simply things a little.

chrisbra:
I see thanks for clarification.

PR #14540, opened 2024-04-13
Fix CUI `setcellwidths` characters draw behavior to same GUI behavior.

mikoto2000:
Fixed #14539 issue.

Issue #14539, opened 2024-04-13
The behavior when specifying 2 for `setcellwidths` is different between the GUI version and the CUI version.


mikoto2000:
### Steps to reproduce

1. Create the following file as `test.txt`.
    ```txt
    0x279c:
    ➜➜➜➜➜
    ```
2. Open a file in vim
    ```sh
    vim -u NONE -N ./test.txt
    ```
3. Set character code and cell width
    ```
    set encoding=utf-8
    call setcellwidths([[0x279c, 0x279c, 2]])
    ```

### Expected behaviour

Same behavior as GUI version.

GUI:
![320875565-7a624485-fe37-411e-9ce5-a6563752c5ce](https://github.com/vim/vim/assets/345832/8f6dfc2e-f96e-48b4-8431-c655ddaef3cb)

CUI:
![320875299-229a17da-2cfb-45c3-9763-1c4d7b3863e8](https://github.com/vim/vim/assets/345832/513b6100-ed33-4260-9d47-fc6dd7bd2b86)
※ Text is drawn up to the yellow arrow, but the cursor moves to the green arrow.

### Version of Vim

9.1.309

### Environment

We have confirmed that this happens in the following three environments

Environment1:
- OS: Windows 11 Pro 64bit build 22631.3296
- Terminal: Windows Terminal
- $TERM: N/A

Environment2:
- OS: Ubuntu 22.04 on WSL2 on Windows 11 Pro 64bit build 22631.3296
- Terminal: mintty(wsltty)
- $TERM: /bin/bash

Environment3:
- OS Ubuntu 22.04
- Terminal: gnome-terminal
- $TERM: /bin/bash

PR #14600, opened 2024-04-20
fix: Regression setcellwidths with GUI

h-east:
A regression has occurred since version [9.1.0344](https://github.com/vim/vim/commit/e20fa59903525e15cecd680a2f32ece8a5d1bc0c).
(ID-Call as patch authors: @mikoto2000 @zeertzjq)

Related: #14539 #14540

### Step to repro.

1. Start GUI Vim with few parameters.
```bash
$ vim -g -Nu NONE +'call setcellwidths([[0x2502, 0x2502, 2]])' +'call setline(1, ["abc\u2502def"])'
```

2. Behavior of GUI Vim 9.1.0344
There is an extra space before `d`.
gui_ng_1

    After type `fd`, The cursor will move to the extra space and `d` will appear.
gui_ng

### Expected behavior
3. Behavior of GUI Vim before version 9.1.0343 or with this PR patch applied.
There is no extra space before d.
gui_ok_1

    After type `fd`.
gui_ok

### DIFF
--- a/src/screen.c
+++ b/src/screen.c
@@ -1981,22 +1981,20 @@ screen_char(unsigned off, int row, int col)
     {
 	char_u	    buf[MB_MAXBYTES + 1];
 
-	if (get_cellwidth(ScreenLinesUC[off]) > 1)
+	if (1
+#ifdef FEAT_GUI
+	    && !gui.in_use
+#endif
+	    && get_cellwidth(ScreenLinesUC[off]) > 1
+	    )
 	{
 	    // If the width is set to 2 with `setcellwidths`
 
-#ifdef FEAT_GUI
-	    if (!gui.in_use)
-	    {
-#endif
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. Can you please turn that into a regression test?

h-east:
@chrisbra This problem only occurs in the GUI, testing using `VerifyScreenDump()` (`term_dumpdiff()`) will not show the difference before and after the fix.
Is there any better way?

chrisbra:
no sure, okay let me include it then.

yegappan:
> @chrisbra This problem only occurs in the GUI, testing using `VerifyScreenDump()` (`term_dumpdiff()`) will not show the difference before and after the fix. Is there any better way?

Sometime ago I added the test_gui_event() function to test GUI specific events.  We should
add a similar function to test GUI specific rendering of characters.

PR #14603, opened 2024-04-20
Relative import in a script sourced from a buffer doesn't work

### LINKED ISSUE #14588: [vim9script] on Windows import in a script sourced from a buffer fails
### Steps to reproduce

The test `Test_autoload_import_relative_compiled_buffer()` in `test_vim9_import.vim` fails on windows. As @yegappan explains in https://github.com/vim/vim/pull/14582#issuecomment-2063953986

> The test failure is caused by using import in a script which is
> sourced from a buffer (not stored in a file). This Vim bug is
> specific to MS-Windows and needs to be fixed.

Running under windows, the test gives the error
```
Failures:
    From test_vim9_import.vim:
    Found errors in Test_autoload_import_relative_compiled():
    Caught exception in Test_autoload_import_relative_compiled():
Vim(import):E282: Cannot read from ":xfile.vim" @ command line..script
```
Note the test is currently disabled on Windows.

Under a recent vim, v9.1.353, on windows, follow the steps in #14577 OP and the failure should occur (I don't have windows, ymmv).

@yegappan, or anyone with windows, could you verify that the steps in #14577 actually fail.

### Expected behaviour

The test runs and passes on Windows.

### Version of Vim

9.1.353

### Environment

windows

### Logs and stack traces

_No response_

yegappan:
When a script is sourced from a buffer, the file name is set to ":source buffer=<number>". 
In MS-Windows, the ":" is a path separator character (used after a drive letter).  This results
in the code trying to use the ":" prefix to import the script on MS-Windows.  To fix this,
when importing a script from a script sourced from a buffer with nofile, don't try to use
a script relative path name.  Fixes #14588.

### DIFF
--- a/src/scriptfile.c
+++ b/src/scriptfile.c
@@ -1430,8 +1430,13 @@ do_source_buffer_init(source_cookie_T *sp, exarg_T *eap)
 	return NULL;
 
     // Use ":source buffer=<num>" as the script name
-    vim_snprintf((char *)IObuff, IOSIZE, ":source buffer=%d", curbuf->b_fnum);
-    fname = vim_strsave(IObuff);
+    if (curbuf->b_ffname != NULL)
+	fname = vim_strsave(curbuf->b_ffname);
+    else
+    {
+	vim_snprintf((char *)IObuff, IOSIZE, ":source buffer=%d", curbuf->b_fnum);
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #14604, opened 2024-04-20
filetype: recognize ondir configuration files

jparise:


runtime/ftplugin/ondir.vim
```diff
@@ -0,0 +1,19 @@
+" Vim filetype plugin file
+" Language:     ondir <https://github.com/alecthomas/ondir>
+" Maintainer:   Jon Parise <jon@indelible.org>
+
+if exists('b:did_ftplugin')
+  finish
+endif
+
+let s:cpo_save = &cpoptions
+set cpoptions&vim
```
chrisbra: those are needed, since you are not using line-contination

jparise: Sorry, but I'm afraid I'm not sure what you're suggesting. Should I _not_ be reseting/restoring `cpoptions` in this way, or at all, in this context? Or is your comment (ha!) about the comments options below?

chrisbra: typicall we need to save and restore the cpo settings, when using e.g. line-continuation (see :h line-continuation). You are not using it in the ftplugin, so you can save the whole save/restore of 'cpoptions' things.

jparise: Ah, thank you!

chrisbra:
thanks!

PR #14607, opened 2024-04-21
[vim9script] Better handle script-level references to variables exported from autoload files

### LINKED ISSUE #14591: [vim9script] Can not change the value of an exported variable under an autoload directory
### Steps to reproduce

Note: there is **no** `import autoload` or symbolic links.
Note: if the directory is not named `autoload` no error.

```
vim first.vim
:source
```
```
Error detected while processing :source buffer=1[3]../home/err/play/yyy/autoload/second.vim:
line    4:
E492: Not an editor command: val = 42
```

**first.vim**
```
vim9script

import './autoload/second.vim'

echo second.val
```
**second.vim**
```
vim9script

export var val = 11
val = 42
```

### Expected behaviour

No error

I ran into this in a largish vim9script application while I was cleaning up; I don't think it was a change I made, but it might have been. I can't say for sure it was a recent change.

I haven't looked into the vim code. Plenty of time getting to the simple example;) I took out the `symlink` code I recently added, still happens.

### Version of Vim

9.1.353

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

errael:
Fix #14591.
This is about script-level access to `autoload` variables.
There are a variety of issues; see #14591 for the basic setup.
Most of the problems are when the variable is on the left hand side.

The mentioned issue shows that the imported file can not modify its own variable that it exports. In the file that does the import, there are similar issues.

The basic issue is that when a file is imported from a directory that has an `autoload` component, the exported items are put into the global namespace with names like "f2#val". This isn't handled well at script-level.

There are 4 changes
- Create `find_var_autoload_prefix()` to check if a variable is an autoload export.
  It is extracted from `find_var()`. It is used in the following three items
- In `lookup_scriptitem()`, which is used when checking if a script command is an assignment, use it for identifying the assignment lhs.
- In `set_var_const()`, to check if asign target is an autoload imported name.
- In `set_var_const()`, to check if asign target is an autoload exported name.
-

### DIFF
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -3340,12 +3340,31 @@ find_var(char_u *name, hashtab_T **htp, int no_autoload)
 	}
     }
 
+    // and finally try
+    return find_var_autoload_prefix(name, 0, htp, NULL);
+}
+
+/*
+ * Find variable "name" with sn_autoload_prefix.
+ * Return a pointer to it if found, NULL if not found.
+ * When "sid" > 0, use it otherwise use "current_sctx.sc_sid".
+ * When "htp" is not NULL  set "htp" to the hashtab_T used.
+ * When "namep" is not NULL set "namep" to the generated name, and
+ * then the caller gets ownership and is responsible for freeing the name.
+ */
+    dictitem_T *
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #14610, opened 2024-04-21
Adopt GTK3 CSS cursor names

dvogel:
Problem: GTK4 will drop numeric cursor support.
Solution: Adopt GTK3 CSS cursor convention.

---

I've been working on GTK4 support. GTK3 began the transition to CSS styling for mouse cursors. GTK4 drops almost all X11-specific constructs, including the numeric X cursors. GTK3 began this transition by supporting CSS names for mouse cursors. This PR adopts that convention as a precursor to GTK4 support.

### DIFF
--- a/src/gui_gtk_x11.c
+++ b/src/gui_gtk_x11.c
@@ -2767,57 +2767,19 @@ mainwin_realize(GtkWidget *widget UNUSED, gpointer data UNUSED)
     static GdkCursor *
 create_blank_pointer(void)
 {
-    GdkWindow	*root_window = NULL;
-#if GTK_CHECK_VERSION(3,0,0)
-    GdkPixbuf   *blank_mask;
-#else
-    GdkPixmap	*blank_mask;
-#endif
     GdkCursor	*cursor;
+
 #if GTK_CHECK_VERSION(3,0,0)
-    GdkRGBA	color = { 0.0, 0.0, 0.0, 0.0 };
+    GdkDisplay  *disp = gtk_widget_get_display(gui.mainwin);
+    cursor = gdk_cursor_new_from_name(disp, "none");
 #else
-    GdkColor	color = { 0, 0, 0, 0 };
-    char	blank_data[] = { 0x0 };
-#endif
-
-#if GTK_CHECK_VERSION(3,12,0)
-    {
-	GdkWindow * const win = gtk_widget_get_window(gui.mainwin);
-	GdkScreen * const scrn = gdk_window_get_screen(win);
-	root_window = gdk_screen_get_root_window(scrn);
-    }
-#else
-    root_window = gtk_widget_get_root_window(gui.mainwin);
-#endif
-
     // Create a pseudo blank pointer, which is in fact one pixel by one pixel
     // in size.
-#if GTK_CHECK_VERSION(3,0,0)
-    {
-	cairo_surface_t *surf;
-	cairo_t		*cr;
-
-	surf = cairo_image_surface_create(CAIRO_FORMAT_A1, 1, 1);
-	cr = cairo_create(surf);
-
-	cairo_set_source_rgba(cr,
-			     color.red,
-			     color.green,
-			     color.blue,
-			     color.alpha);
-	cairo_rectangle(cr, 0, 0, 1, 1);
-	cairo_fill(cr);
-	cairo_destroy(cr);
-
-	blank_mask = gdk_pixbuf_get_from_surface(surf, 0, 0, 1, 1);
-	cairo_surface_destroy(surf);
-
-	cursor = gdk_cursor_new_from_pixbuf(gdk_window_get_display(root_window),
-					    blank_mask, 0, 0);
-	g_object_unref(blank_mask);
-    }
-#else
+    GdkWindow	*root_window = NULL;
+    GdkColor	color = { 0, 0, 0, 0 };
+    GdkPixmap	*blank_mask;
+    char	blank_data[] = { 0x0 };
+    root_window = gtk_widget_get_root_window(gui.mainwin);
     blank_mask = gdk_bitmap_create_from_data(root_window, blank_data, 1, 1);
     cursor = gdk_cursor_new_from_pixmap(blank_mask, blank_mask,
 					&color, &color, 0, 0);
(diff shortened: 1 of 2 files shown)


src/gui_gtk_x11.c
```diff
@@ -7291,11 +7253,38 @@ static const int mshape_ids[] =
...
+    "crosshair",                // GDK_CROSSHAIR
+    "pointer",                  // GDK_HAND1
+    "pointer",                  // GDK_HAND2
+    "pointer",                  // GDK_PENCIL (no good option for this one)
+    "help",                     // GDK_QUESTION_ARROW
+    "default",                  // GDK_RIGHT_PTR (no css analogue)
+    "default",                  // GDK_CENTER_PTR (no css analogue)
+    "default"                   // GDK_LEFT_PTR
+};
+# endif
```
dvogel: These choices are best considered relative to the `mshape_ids` array above. It seems we'll have to make some tough choices because GTK no longer supports all of the shapes documented under `mouseshape`. This PR is still in draft mode because I have not updated that documentation to explain which shapes are supported under GTK. I will do so once we agree on the choices in this list.

chrisbra:
Thanks for working on that. I think that makes sense. Do I understand correctly, that this should not break anything on GUI (GTK3) Vim?

dvogel:
> Thanks for working on that. I think that makes sense. Do I understand correctly, that this should not break anything on GUI (GTK3) Vim?

No errors or crashes will occur. However it is not backward compatible. Users who currently use numeric cursor shape IDs or any of the following vim shape names will see different icons. The underlying issue is that [`gdk_cursor_new_from_name`](https://docs.gtk.org/gdk3/ctor.Cursor.new_from_name.html) supports fewer cursor shapes than [`gdk_cursor_new`/`GdkCursorType`](https://docs.gtk.org/gdk3/enum.CursorType.html) does. The latter is based on X11 conventions that are being deprecated by GTK.

* `hand1`
* `hand2`
* `pencil`
* `rightup-arrow`
* `up-arrow`

chrisbra:
Okay, so it looks like we will have those 'mouseshapes': 

(System is `w` for Windows and `x` for X11, perhaps we should add a `G` for GTK). 


System | Vim Name | GDK/CSS Name | Desc
-------|----------|----------|------
w x	   |arrow		  |default   |  Normal mouse pointer
w x	 |blank     |blank		 |  no pointer at all (use with care!)
  x	   |hand1     |pointer	 |	black hand
  x	   |hand2     |pointer	 |  white hand
  x	   |pencil    |pointer	 |	what you write with
  x	   |question  |help	     |  big  ?
  x	   |crosshair |crosshair |  like a big thin +
w x	   |updown    | ns-resize|	up-down sizing arrows
w x	   |leftright | ew-resize|  left-right sizing arrows
w x	   |no        |not-allowed|	The system's usual 'no input' pointer
  x	   |udsizing|ns-resize|indicates up-down resizing
  x	   |lrsizing|ew-resize|indicates left-right resizing

Missing:
System | Vim Name | GDK Name | Desc
-------|----------|----------|------
w x	   |busy		| | The system's usual busy pointer
w x	   |beam		 | |  I-beam
  x	   |rightup-arrow | |	arrow pointing right-up
w x	   |up-arrow | |	arrow pointing up
  x	   |<number> | |	any X11 pointer number (see X11/cursorfont.h)


I am a bit surprised we don't have *beam* and *busy*. But when this is no longer supported by the underlying library :/ 


Currently the help states also this:
> 	Any modes not specified or shapes not available use the normal mouse pointer.

So it seems more or less okay? Can we keep the current cursor shapes when we still use GTK2/3 UI?

dvogel:
We will have support for the vim names `beam` and `busy`. They translate to the CSS names `text` and `progress`. Here's a couple of screenshots showing how those look on my screen with `vim -g --clean -U /dev/null`. Please ignore the cursor size. It appears comically large here because the (a) it is configured to be extra large by preference and (b) the default font size is very small.

vim_mouseshape=a:beam
vim_mouseshape=a:busy

I like the idea of adding a `g` for GTK to that table. I'll incorporate that into my next update.

Regarding when to adopt this, I don't have a firm preference. I wanted to start on this because it is coming eventually and would take some discussion. If it is best to delay this until GTK4 that is fine with me. I think one argument for adopting this in GTK3 sooner rather than later relates to GTK2 support. I don't yet have a firm grasp on how hard it will be to maintain support for all of GTK2/3/4. Whatever the absolute effort required, on a relative basis it would be considerably easier to support any two versions of GTK at once. This seems to have been the approach so far. If we decide that we will support only GTK3/4 in vim10 then I think there's two ways to approach that:

* We could decide that vim with GTK is meant to lean into all of the GTK-isms, including only supporting the X11-isms that have direct Wayland couterparts. This would make Motif the primary means of accessing X11-isms.
* We could instead decide that vim's GTK3 integration will opt into all of the X11-isms to allow people to maximize their use of X11 until GTK5 support is added. This would mean the only way for gvim users to maximize their use of Wayland would be to use GTK4.

The first option would argue for merging this now. The second option would argue for folding this into the much larger effort to support GTK4.

We currently have a kludge that shunts vim users into X11 even with GTK3 unless they explicitly opt into Wayland. I think that was good when it was added but eventually this will have to be removed. 

```
#if GTK_CHECK_VERSION(3,10,0)
    // Vim currently assumes that Gtk means X11, so it cannot use native Gtk
    // support for other backends such as Wayland.
    //
    // Use an environment variable to enable unfinished Wayland support.
    if (getenv("GVIM_ENABLE_WAYLAND") == NULL)
	gdk_set_allowed_backends("x11");
#endif
```

src/gui_gtk_x11.c
```diff
@@ -7287,12 +7276,26 @@ mch_set_mouse_shape(int shape)
 	    else
 		id &= ~1;	// they are always even (why?)
 	}
+# if GTK_CHECK_VERSION(3,0,0)
+	else if (shape < (int)ARRAY_LENGTH(mshape_css_names))
+	{
+	    css_name = mshape_css_names[shape];
+	}
```
chrisbra: style, you can get rid of the braces for single line statements.

src/gui_gtk_x11.c
```diff
@@ -7287,12 +7276,26 @@ mch_set_mouse_shape(int shape)
...
+# else
 	else if (shape < (int)ARRAY_LENGTH(mshape_ids))
+	{
```
chrisbra: also here.

chrisbra (commented):
Thanks. this looks like we can clean up the GTK code a bit. Didn't you also want to update the help table at `:h 'mouseshape'`?

dvogel:
> Thanks. this looks like we can clean up the GTK code a bit. Didn't you also want to update the help table at `:h 'mouseshape'`?

Yeah, thanks for the reminder. I'll try to update that later today.

chrisbra:
thanks

PR #14611, opened 2024-04-21
test_winfixbuf.vim is very slow

yegappan:
The test_winfixbuf.vim test takes 15 seconds to run.  After these changes it takes around 1.3 seconds.
The "hidden" option is not cleared at the end of some tests and "--not-a-term" option is not used when
using system() to run a Vim instance.


chrisbra:
wow, great improvement

PR #14613, opened 2024-04-21
Crash when typing many keys with D- modifier

zeertzjq:
Problem:  Crash when typing many keys with D- modifier (after 9.1.0227).
Solution: Don't treat a 0x80 byte inside a special sequence as the start
          of a special sequence.


chrisbra:
thanks!

PR #14614, opened 2024-04-22
Running test_vim9_builtin.vim tests is slow

yegappan:
The tests in test_vim9_builtin.vim use the CheckDefAndScriptFailure(), CheckDefExecFailure(), etc. functions from the vim9.vim file.
These functions store the supplied lines in a file and then source the file.  This takes time.  Instead of using a file, these lines can be stored in a buffer and then sourced.  This method is already used by the new Vim9 test scripts (e.g. test_vim9_class.vim, test_vim9_typealias.vim, test_vim9_enum.vim etc.).  Convert the tests in test_vim9_builtin.vim to use the new functions.

Before this change, it used to take around 8 seconds to run this test in my local setup.  After this change, it takes around 2.5 seconds.  Depending on the disk and IO method used, it may take varying amount of time to run this test.

### DIFF
--- a/src/testdir/test_vim9_builtin.vim
+++ b/src/testdir/test_vim9_builtin.vim
@@ -66,16 +66,16 @@ def Test_abs()
   assert_equal(0, abs(0))
   assert_equal(2, abs(-2))
   assert_equal(3, abs(3))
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #14615, opened 2024-04-22
In `:def` handle storing to vim9 autoload export.

### LINKED ISSUE #14606: [vim9script] Missing exception when modifying `import autoload`'s `const` variable from `:def`
A write to a modifiable export also fails. No write occurs (and there is no indication of a problem).

### Steps to reproduce

Run the following script and see the error
```
FAIL: missing exception for trying to modify f5.val
```
```
vim9script

mkdir('Xautoload_vars/autoload', 'p')
# Create script to import that exports a constant variable
var lines =<< trim END
    vim9script
    export const val = 11
END
writefile(lines, 'Xautoload_vars/autoload/Xauto_vars_f5.vim')

import autoload './Xautoload_vars/autoload/Xauto_vars_f5.vim' as f5

try
    def F()
        f5.val = 13
    enddef
    F()
    echo "FAIL: missing exception for trying to modify f5.val"
catch
    echo v:throwpoint
    echo v:exception
endtry


delete('./Xautoload_vars/autoload/Xauto_vars_f5.vim')
delete('./Xautoload_vars/autoload', 'd')
delete('./Xautoload_vars', 'd')
```

**NOTE:** If the imported script is loaded before attempting the write access, then an exception occurs as expected.

### Expected behaviour

Expect something like the following error
```
:source buffer=1[17]..function <SNR>39_F, line 1
Vim(eval):E741: Value is locked: val
```

### Version of Vim

9.1.353

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

errael:
Fix #14606.

Also fix issue with :def in legacy script accessing the wrong variables.

Problem occurs when `import autoload ./.../autoload/...`. The `autoload` in the specified path causes the use of an autoload_prefix which combines with the `import autoload` to create trouble.

In `generate_store_var()` `case dest_script` use `ISN_STOREEXPORT`, when needed, instead of `ISN_STORES`. When executing `ISN_STOREEXPORT`, check for autoload_prefix.

**Reviewers** - there may be better ways to handle this problem.

### DIFF
--- a/src/testdir/test_vim9_import.vim
+++ b/src/testdir/test_vim9_import.vim
@@ -1206,6 +1206,7 @@ def Test_autoload_export_variables()
   mkdir('Xautoload_vars/autoload', 'pR')
   var lines =<< trim END
     vim9script
+    g:Xautoload_vars_autoload = true
     export var val = 11
     val = 42
   END
@@ -1215,13 +1216,24 @@ def Test_autoload_export_variables()
   writefile(lines, 'Xautoload_vars/autoload/Xauto_vars_f2.vim', 'D')
   lines =<< trim END
     vim9script
+    g:Xautoload_vars_autoload = false
 
     import autoload './Xautoload_vars/autoload/Xauto_vars_f2.vim' as f2
+    # Verify that the import statement does not load the file.
+    assert_equal(false, g:Xautoload_vars_autoload)
 
     def F(): number
       return f2.val
     enddef
+    # Verify compile does not load the file.
+    defcompile F
+    assert_equal(false, g:Xautoload_vars_autoload)
+
+    # load the file by accessing the exported variable
     assert_equal(42, F())
+    assert_equal(true, g:Xautoload_vars_autoload)
+    unlet g:Xautoload_vars_autoload
+
     assert_equal(42, f2.val)
     f2.val = 17
     assert_equal(17, f2.val)
@@ -1264,6 +1276,44 @@ def Test_autoload_export_variables()
     f4.val = 13
   END
   v9.CheckScriptFailure(lines, 'E46:')
+
+  # Test const var is not modifiable from importing script from :def.
+  # Github issue: #14606
+  lines =<< trim END
+    vim9script
+    export const val = 11
+  END
+  writefile(lines, 'Xautoload_vars/autoload/Xauto_vars_f5.vim', 'D')
+  lines =<< trim END
+    vim9script
+
+    import autoload './Xautoload_vars/autoload/Xauto_vars_f5.vim' as f5
+
+    def F()
+      f5.val = 13
+    enddef
+    F()
+  END
+  v9.CheckScriptFailure(lines, 'E741:')
+
+  # Still part of Github issue: #14606
+  lines =<< trim END
+    vim9script
+    export var val = 11
+  END
+  writefile(lines, 'Xautoload_vars/autoload/Xauto_vars_f6.vim', 'D')
+  lines =<< trim END
+    vim9script
+
+    import autoload './Xautoload_vars/autoload/Xauto_vars_f6.vim' as f6
+
+    def F()
+      f6.val = 13
+    enddef
+    F()
+    assert_equal(13, f6.val)
+  END
+  v9.CheckScriptSuccess(lines)
 enddef
 
 def Test_autoload_import_relative_autoload_dir()
--- a/src/vim9instr.c
+++ b/src/vim9instr.c
@@ -2394,6 +2394,7 @@ generate_store_var(
 	case dest_vimvar:
 	    return generate_STORE(cctx, ISN_STOREV, vimvaridx, NULL);
 	case dest_script:
+	case dest_script_v9:
 	    {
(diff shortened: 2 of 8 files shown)


errael:
@yegappan I hope you look at this. The notes in the related issue may be of interest.

src/vim9instr.c
```diff
@@ -2401,10 +2401,15 @@ generate_store_var(
...
+		    // of ISN_STORES) if there is no sn_autoload_export or if
+		    // the name does not start with "s:". Note that ISN_STORES
+		    // assumes/requires the name to start with "s:".
 		    if (SCRIPT_ID_VALID(scriptvar_sid)
 			     && SCRIPT_ITEM(scriptvar_sid)->sn_import_autoload
-			     && SCRIPT_ITEM(scriptvar_sid)->sn_autoload_prefix
-								       == NULL)
+			     && ((SCRIPT_ITEM(scriptvar_sid)
+						  ->sn_autoload_prefix == NULL)
+				|| STRCMP("s:", name) != 0))
```
yegappan: The `s:` prefix for script-local variable is not supported for script-local variables in Vim9 scripts.
Under what condition the name will start with `s:` in a Vim9 script?

src/testdir/test_vim9_import.vim
```diff
@@ -1264,6 +1264,44 @@ def Test_autoload_export_variables()
     f4.val = 13
   END
   v9.CheckScriptFailure(lines, 'E46:')
+
+  # Test const var is not modifiable from importing script from :def.
```
yegappan, line 1280: Can you also add new tests to test_vim9_disassemble.vim to check the generated instructions for these cases (based on the examples in your note)?

src/vim9instr.c
errael: Looking at #14606 pure vim9script is generating a `ISN_STORES` which it should not do. But as to

> Under what condition the name will start with s: in a Vim9 script

I have virtually no experience with legacy script, and even less with using legacy and `vim9script` together but consider this legacy script 
```
let s:foo = 'init'
def F()
    s:foo = ':def assign to script local'
enddef
call F()
disa F
echo ' '
echo 'after F():' s:foo
```
which outputs/generates the `ISN_STORES` which is causing the trouble in pure `vim9script`
```
F
    s:foo = ':def assign to script local'
   0 PUSHS ":def assign to script local"
   1 STORES s:foo in /tmp/legacy.vim
   2 RETURN void

after F(): :def assign to script local
```
And look a few lines down in this function where it says
```
// "s:" may be included in the name.
```
Look in the original code in `vim9execute.c`
```
// store s: variable in old script or autoload import
case ISN_STORES:
case ISN_STOREEXPORT:
    {
        int		sid = iptr->isn_arg.loadstore.ls_sid;
        hashtab_T	*ht = &SCRIPT_VARS(sid);
        char_u	*name = iptr->isn_arg.loadstore.ls_name;
        dictitem_T	*di = find_var_in_ht(ht, 0,
                                iptr->isn_type == ISN_STORES
                                         ? name + 2 : name, TRUE);
```
When it sees the `ISN_STORES` instruction it skips the assumed `s:`, using `name + 2` for the `ht` lookup, but name was `val` in the test that causes the problem, so it looks up `l`.

src/testdir/test_vim9_import.vim
errael: I can try :)

src/vim9instr.c
yegappan: Thanks for the explanation.

yegappan:
> > Can you also add new tests to test_vim9_disassemble.vim
> 
> @yegappan Added test for the case that was failing. I think the rest of the cases are already tested.
> 
> Also tweaked the first test in `Test_autoload_export_variables()` to verify that neither the `import` nor `defcompile` load the script; it is not loaded until the function is executed.

Thanks for adding the test.

yegappan (approved):

errael:
@yegappan After saying that `val` would have been looked up as `l`, I realized I didn't explore this. Sure enough, both before and after this PR, things are screwy. And it looks like the `s:` prefix in the `:def` might be optional, something else to check.

I put this PR into draft and will take a look tomorrow. I'm fuzzy on what should happen. ???

The following has code to changes `xxfoo`, but `foo` gets changed
```
let s:foo = 'init'
let s:xxfoo = 'init'
def F()
    xxfoo = 'NEW VALUE'
enddef
call F()
echo ' '
echo printf('after F(): s:foo: %s, s:xxfoo: %s', s:foo, s:xxfoo)
```
outputs
```
after F(): s:foo: NEW VALUE, s:xxfoo: init
```

errael:
- Updated PR to distinguish `lhs->lhs_dest` between `dest_script` and `dest_script_v9` which simplifies the compile/execute logic.
- Add test for `:def` in legacy script accesses the correct script variables.

errael:
@yegappan Here's the "final" (we'll see) changes for review
https://github.com/vim/vim/compare/a8a9a9e71f06026b66962695f6e7fbcb85ba9eb5..bf526e0039daccdbeebed969e68116eb93569824

yegappan (approved):

chrisbra:
thanks all!

Issue #14609, opened 2024-04-21
getregionpos() returns a list of buffer segments for a given range


justinmk:
# Problem

A buffer region in Vim, is described by four parameters: `{pos1}, {pos2}, {regtype}, {inclusive}`

Any function that accepts only `{pos1}, {pos2}` must either be reworked to also accept `{regtype}, {inclusive}`, or the caller must manually figure out the buffer segments that map to the "region", and feed them into the function.

# Feature request

Extracted from discussion in https://github.com/vim/vim/pull/13998#issuecomment-2067800078

> (Main use case here: highlight-on-yank as in https://github.com/machakann/vim-highlightedyank.)

Introduce  `getregionpos({pos1}, {pos2}, {regtype}, {inclusive})` which returns a list of start/end positions in the buffer.

PR #14617, opened 2024-04-22
Add getregionpos()

Shougo:
For #14609


src/evalfunc.c
```diff
@@ -5675,11 +5700,78 @@ f_getregion(typval_T *argvars, typval_T *rettv)
...
     curbuf = save_curbuf;
     curwin->w_buffer = curbuf;
     virtual_op = save_virtual;
 }
 
+/*
+ * "getregionpos()" function
+ */
+    static void
+f_getregionpos(typval_T *argvars, typval_T *rettv)
```
yegappan, line 5779: Can you add tests for the new getregionpos() function?

Shougo: Yes.  It should.  But current tests are very many.  It is hard.

yegappan: Without tests, this cannot be merged.

Shougo: Yes...

Shougo:
I have added tests.

NOTE:  I cannot test for `list_alloc()` and `list_append_list()` though...

runtime/doc/builtin.txt
```diff
@@ -4327,6 +4329,27 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		Can also be used as a |method|: >
 			getpos('.')->getregion(getpos("'a"))
 <
+getregionpos({pos1}, {pos2} [, {opts}])            *getregionpos()*
+		Returns the list of position from {pos1} to {pos2} from a
+		buffer.
+		See |getregion()| for the arguments and notes.
+		Note: When the selection is blockwise, it returns a pair of
+		positions for every line.
```
justinmk: should it always return the same structure? For non-blockwise, the end pos is not technically "necessary", but have the same structure simplifies the logic needed in consumer logic.

runtime/doc/builtin.txt
```diff
@@ -4327,6 +4329,27 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
...
+
+		The position is a |List| with four numbers:
+		    [bufnum, lnum, col, off]
+		"bufnum" is zero, unless a mark like '0 or 'A is used, then it
+		is the buffer number of the mark.
```
justinmk: Is "bufnum is zero" referring to the input positions or the output? Can it support non-current buffer?

runtime/doc/builtin.txt
```diff
@@ -4327,6 +4329,27 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		Can also be used as a |method|: >
 			getpos('.')->getregion(getpos("'a"))
 <
+getregionpos({pos1}, {pos2} [, {opts}])            *getregionpos()*
+		Returns the list of position from {pos1} to {pos2} from a
```
justinmk:
```suggestion
		Returns a list of positions describing the buffer text segments bound by {pos1} to {pos2}.
```

runtime/doc/builtin.txt
Shougo: OK.  Fixed.

runtime/doc/builtin.txt
Shougo: I have changed it.

runtime/doc/builtin.txt
```diff
@@ -4327,6 +4329,27 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
 		Can also be used as a |method|: >
 			getpos('.')->getregion(getpos("'a"))
 <
+getregionpos({pos1}, {pos2} [, {opts}])            *getregionpos()*
+		Returns a list of positions describing the buffer text
+		segments bound by {pos1} to {pos2} buffer.
```
justinmk:
```suggestion
		segments bound by {pos1} and {pos2}.
```

runtime/doc/builtin.txt
```diff
@@ -4327,6 +4329,27 @@ getregion({pos1}, {pos2} [, {opts}])			*getregion()*
...
+		See |getregion()| for the arguments and notes.
+		Note: When the selection is blockwise, it returns a pair of
+		positions for every line:
+                    [[{pos_start}, {pos_end}], ...]
```
justinmk: this note is not needed now, right? because the same structure is always returned

Shougo: The same structure is returned.  But it may be hard to understand why the positions are list.

girishji: Can you say instead:
```
Same as |getregion()|, but returns a list of positions describing the buffer text
segments bound by {pos1} and {pos2}.
```

You can remove:
```
See |getregion()| for the arguments and notes.
```

src/evalfunc.c
girishji: Can you add one more test similar to `def Test_getregion()` in `test_vim9_builtin.vim`, say `Test_getregionpos()` and check for `E1209` and `E1211`?

runtime/doc/builtin.txt
Shougo: OK.  I have fixed it.

src/evalfunc.c
Shougo: OK.   I have added them.

chrisbra:
So, is this ready? Have a few people tried it out?

jessepav:
I've been running with this PR applied, and it works as well as `getregion()` for me.

The only issue I've faced with both of these functions is that they error out (`E964: Invalid column number: 2147483647`) when you make a linewise visual selection and use the `'<` and `'>` positions, since the column of `pos2` is set to `v:maxcol`. 

I don't know if that is unintended behavior, though, and is easy enough to check for in scripts.

girishji:
> So, is this ready? Have a few people tried it out?

1. I tried this test, and it gave error `E964: Invalid column number: 2147483647`

```
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            :echom getregionpos(getpos("'["), getpos("']"))
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :normal yy
EOF
)
```

Please check if my test itself is wrong. I wrote it in a hurry.

2. It seems it is not exactly like `getregion()`. If I were to select a visual region using `v` or `V`, I think it should return `{pos}` of all intermediate lines, not just first and last line. It does this for `ctrl-v`. I got this from reading the doc. I have not tested this.

Shougo:
I have fixed columns error.  It already exists when `getregion()` implementation though.

> It seems it is not exactly like getregion(). If I were to select a visual region using v or V, I think it should return {pos} of all intermediate lines, not just first and last line. It does this for ctrl-v. I understood this from reading the doc. I have not tested this.

I know the feature.  I can change the behavior.  But I think it is not useful.
`<c-v>` behavior is changed because of https://github.com/vim/vim/pull/14617#issuecomment-2068988099.
Please fix my documentation if it is not exact.

jessepav:
After the most recent commit, this PR works great for me – and now `getregion()` works for `'V'` mode too.

girishji:
The test I mentioned above still fails. Same error (`E964`).

Somewhere it is mentioned that the motivation for this API is [this](https://github.com/machakann/vim-highlightedyank), which roughly translates to:

```
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            var pos = getregionpos(getpos("'["), getpos("']"))->mapnew((_, v) => [v[0][1], v[0][2], v[1][2] - v[0][2] + 1])
            var m = matchaddpos('IncSearch', pos)
            timer_start(300, (_) => m->matchdelete())
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :normal yy
EOF
)
```

This test does not pass. I'm afraid this PR is not well thought through.

jessepav:
The test given above doesn't fail for me with the latest commit.

`pos` is set to `[[1, 1, 5]]` by the `autocmd` when I test it.

Shougo:
Yes.  It works for me.  I think you don't build the latest version.

girishji:
The second test fails though. There is no value to `getregionpos()` if you are just returning two `getpos()`, which are the exact arguments I am supplying!

Try the following. It does not highlight both lines, but only the first line. This is because getregionpos is returning `[[[1, 1, 1, 0], [1, 2, 7, 0]]]` instead of `[[[1, 1, 1, 0], [1, 1, 5, 0]], [[1, 2, 1, 0], [1, 2, 7, 0]]]`.

```
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            var pos = getregionpos(getpos("'["), getpos("']"))->mapnew((_, v) => [v[0][1], v[0][2], v[1][2] - v[0][2] + 1])
            var m = matchaddpos('IncSearch', pos)
            timer_start(300, (_) => m->matchdelete())
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :normal yj
EOF
)
```

girishji:
I found another bug. `<c-v>` does not work. The values returned are incorrect.

```
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            :echom getregionpos(getpos("'["), getpos("']"))
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :exec "normal \<c-v>"
    :normal 3ljy
EOF
)
```

This PR is not well tested.

Shougo:
> I found another bug. `<c-v>` does not work. The values returned are incorrect. It just returns first and last positions, instead of every line in the visual block.

It is not bug.  Please see `getregion()` documentation.

```
		The optional argument {opts} is a Dict and supports the
		following items:

			type		Specify the region's selection type
					(default: "v"):
			    "v"		for |characterwise| mode
			    "V"		for |linewise| mode
			    "<CTRL-V>"	for |blockwise-visual| mode

			exclusive	If |TRUE|, use exclusive selection
					for the end position
					(default: follow 'selection')
```

The default `type` is `v`.  You need to use `visualmode()`.

It works for me.

```vim
vim9script
:set shortmess=I
autocmd TextYankPost * {
    if v:event.operator ==? 'y'
      :echom getregionpos(getpos("'["), getpos("']"), { type: visualmode() })
    endif
}
setline(1, ['abcd', 'efghij'])
:exec "normal \<c-v>"
:normal 3ljy
```

girishji:
Thanks, I still get `E475: Invalid value for argument type:` when I do this:

```vim
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            var pos = getregionpos(getpos("'["), getpos("']"), { type: visualmode() })
            var m = matchaddpos('IncSearch', pos->mapnew((_, v) => [v[0][1], v[0][2], v[1][2] - v[0][2] + 1]))
            timer_start(300, (_) => m->matchdelete())
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :normal yj
EOF
)

```

jessepav:
Visual mode was never entered in your test script, so `visualmode()` is returning an empty string.

Shougo:
> Try the following. It does not highlight both lines, but only the first line. This is because getregionpos is returning [[[1, 1, 1, 0], [1, 2, 7, 0]]] instead of [[[1, 1, 1, 0], [1, 1, 5, 0]], [[1, 2, 1, 0], [1, 2, 7, 0]]].

OK.  I get the problem.  I will change the result value.

girishji:
> ```viml
> getregionpos(getpos("'["), getpos("']"), { type: visualmode() })
> ```

When I just do `echom` instead, it enters the visual mode though.

```
   :echom getregionpos(getpos("'["), getpos("']"), { type: visualmode() })
```

girishji:
> Visual mode was never entered in your test script, so `visualmode()` is returning an empty string.

Thanks. Modified the script.

I see 2 issues:

- `<c-v>` is reporting one additional character (to the right) than what was selected. Is this intended? How are the tests passing?
- intermediate lines are ignored in `v` and `V`

You can use the script, visual select, and verify.

```
/vim -Nu NONE -S <(cat <<EOF                                                                                                        [getregion_pos]
    vim9script
    :set shortmess=I
    autocmd TextYankPost * {
        if v:event.operator ==? 'y'
            var pos = visualmode() == null_string ? \
                 getregionpos(getpos("'["), getpos("']")) : getregionpos(getpos("'["), getpos("']"), { type: visualmode() })
            :echom pos
            var m = matchaddpos('IncSearch', pos->mapnew((_, v) => [v[0][1], v[0][2], v[1][2] - v[0][2] + 1]))
            timer_start(300, (_) => m->matchdelete())
        endif
    }
    setline(1, ['abcd', 'efghij'])
    :normal yj
EOF
)
```

Shougo:
> `<c-v>` is reporting one additional character (to the right) than what was selected. Is this intended? How are the tests passing?

Hm...  It seems broken.  I will fix.

Shougo:
I have fixed the problems.  Please test.

src/testdir/test_visual.vim
```diff
@@ -1955,4 +1992,31 @@ func Test_getregion_invalid_buf()
   bwipe!
 endfunc
 
+func Test_getregion_maxcol()
+  new
+  set shortmess=I
```
zeertzjq: Need to restore 'shortmess' value at end of test

chrisbra:
@justinmk does that do what you want?

src/testdir/test_visual.vim
Shougo: Fixed.

girishji:
> I have fixed the problems. Please test.

I found many problems:

- Do reverse selection in `v` (begin position after end position), and it gives wrong results.
- Have two words `foo bar`, and cursor on `b`, and then yank the word `ye`, it returns wrong results (highlights `foo` instead of `bar`).
- `v` does not work unless cursor is on first column

I updated the test to include the `offset` and `Exclusive`. If anyone has time, please try the script below. Yank some text (with and without visual selection) and verify the values printed.

```
./vim -Nu NONE -S <(cat <<EOF
    vim9script
    def HighlightYank()
        if v:event.operator ==? 'y'
            var [beg, end, vmode] = [getpos("'["), getpos("']"), visualmode()]
            var pos = getregionpos(beg, end, vmode != null_string ? {type: vmode} : {})
            :echom pos
            var m = matchaddpos('IncSearch', pos->mapnew((_, v) => {
                var col_beg = v[0][2] + v[0][3]
                var col_end = v[1][2] + v[1][3] + (!v:event.inclusive ? 1 : 0)
                return [v[0][1], col_beg, col_end - col_beg + 1]
            }))
            timer_start(300, (_) => m->matchdelete())
        endif
    enddef
    autocmd TextYankPost * HighlightYank()
    setline(1, ['abcd', 'efghij', 'foo bar', '	baz	 bza', 'ぬfooぼ barі'])
    :set ruler
    :normal 2lv2jy
EOF
)
```

Shougo:
Well, can you add the tests in `test_visual.vim` for the problems?

Shougo:
I have fixed https://github.com/vim/vim/pull/14617#issuecomment-2088142616 problems.

girishji:
> Well, can you add the tests in `test_visual.vim` for the problems?
> 
> The visual mode position routine is broken. I will fix it.

Sure, but has to wait till next week.

zeertzjq:
The problem with `'>` mark seems to apply to cursor() with 'virtualedit' as well. I've created another PR #14698 for that.

Shougo:
I have rebased it and fixed the conflict.

src/testdir/test_visual.vim
```diff
@@ -1784,10 +1823,8 @@ func Test_visual_getregion()
     call assert_fails('call getregion([g:buf, 10, 2, 0], [g:buf, 0, 1, 0])', 'E966:')
     call assert_fails('call getregion([g:buf, 1, 1, 0], [g:buf, 11, 2, 0])', 'E966:')
     call assert_fails('call getregion([g:buf, 11, 2, 0], [g:buf, 1, 1, 0])', 'E966:')
-    call assert_fails('call getregion([g:buf, 1, 1, 0], [g:buf, 10, 0, 0])', 'E964:')
-    call assert_fails('call getregion([g:buf, 10, 0, 0], [g:buf, 1, 1, 0])', 'E964:')
-    call assert_fails('call getregion([g:buf, 1, 1, 0], [g:buf, 10, 3, 0])', 'E964:')
-    call assert_fails('call getregion([g:buf, 10, 3, 0], [g:buf, 1, 1, 0])', 'E964:')
+    call assert_fails('call getregion([g:buf, 1, 0, 0], [g:buf, 1, 1, 0])', 'E964:')
+    call assert_fails('call getregion([g:buf, 1, 1, 0], [g:buf, 1, 0, 0])', 'E964:')
```
zeertzjq: I think this change can now be reverted.

Shougo: OK.  I have added the tests.

chrisbra:
Thanks all. I have made one minor change using `mb_ptr2len()` instead of `utfc_ptr2len()`.

PR #14620, opened 2024-04-22
runtime(java): Improve the recognition of the "style" method declarations

zzzyxwvut:
- Request the new regexp engine (`v7.3.970`) for `[:upper:]` and  
  `[:lower:]`.

- Recognise declarations of in-line annotated methods.

- Recognise declarations of `strictfp` methods.

- Establish partial order for method modifiers as shown in  
  the MethodModifier production; namely, `public` and  
  friends should be written the leftmost, possibly followed  
  by `abstract` or `default`, or possibly followed by other  
  modifiers.

- Stop looking for parameterisable primitive types (`void<?>`,  
  `int<Object>`, etc., are malformed).

- Stop looking for arrays of `void`.

- Acknowledge the prevailing convention for method names to  
  begin with a small letter and for class/interface names to  
  begin with a capital letter; and, therefore, desist from  
  claiming declarations of enum constants and constructors  
  with `javaFuncDef`.  
  Rationale:
    + Constructor is distinct from method:
      * its (overloaded) name is not arbitrary;
      * its return type is implicit;
      * its `throws` clause depends on indirect vagaries of  
        instance (variable) initialisers;
      * its invocation makes other constructors of its type  
        hierarchy invoked one by one, concluding with the  
        primordial constructor;
      * its explicit invocation, via `this` or `super`, can  
        only appear as the first statement in a constructor  
        (not anymore, see JEP 447); else, its `super` call  
        cannot appear in constructors of `record` or `enum`;  
        and neither invocation is allowed for the primordial  
        constructor;
      * it is not a member of its class, like initialisers,  
        and is never inherited;
      * it is never `abstract` or `native`.
    + Constructor declarations tend to be few in number and  
      merit visual recognition from method declarations.
    + Enum constants define a fixed set of type instances  
      and more resemble class variable initialisers.

Note that the code duplicated for `@javaFuncParams` is written  
keeping in mind for `g:java_highlight_functions` a pending 3rd  
variant, which would require none of the `:syn-cluster` added  
groups.


chrisbra:
> // VIM_TEST_SETUP set encoding=utf-8 termencoding=utf-8

> as they are only meaningful for visual inspection of created
screen dump files?

Yeah I believe so. But when trying to inspect debug the test, a nice visual representation is helpful instead of using `??`

But isn't `encoding=utf-8` not already default encoding? At least  that's what I thought for Linux.

chrisbra:
Thanks, I didn't notice that vim still uses latin1 encoding in this case. I guess this is fine. 

> libvterm: Unhandled CSI t

I have seen this a few times, especially with debug builds, but those appear to be harmless.

Thanks!

zzzyxwvut:
I have also looked at the output of `xxd` for an `input/` file  
and a `*_00*` screen dump.  The relevant multibyte sequences  
match.

We can proceed.

chrisbra:
okay thanks!

Shane-XB-Qian:
> Note that the code duplicated for ***@***.***` is written  
> keeping in mind for `g:java_highlight_functions` a pending 3rd  
> variant, which would require none of the `:syn-cluster` added  
> groups.

some weird on `g:java_highlight_functions`, not exactly what's that for..

for example following:

```java
    public static Date StrToDate(String date) {  // <== why this would not highlight whatever its value (or unlet)?
        try {
            return SIMPLE_DATE_FORMAT.parse(date);
        } catch (ParseException e) {
            e.printStackTrace();
        }
        return null;
    }

    public static String DateToString(Date date) {  // <== why this would not highlight whatever its value (or unlet)?
        return SIMPLE_DATE_FORMAT.format(date);  // <-- would highlight this if it set to 'indent' and why hl this?
    }

    public static java.sql.Date utilDateToSqlDate(Date date) {  // <~~ would highlight this if it set to 'style'
        return new java.sql.Date(date.getTime());  // <-- would highlight this if it set to 'indent' and why hl this?
    }
```

-- 
shane.xb.qian

zzzyxwvut:
Why, `:h ft-java-syntax` says what `g:java_highlight_functions`  
is for.

With `style` assigned to `g:java_highlight_functions`, all  
supplied examples are correctly highlighted: `StrToDate` and  
`DateToString` do not follow naming conventions for method  
names.  Please read `:h ft-java-syntax` and [code conventions](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html#200)  
from Sun.  Hint: try `strToDate` and `dateToString`.  This PR  
hasn't changed that.

With `indent` assigned to `g:java_highlight_functions`, all  
supplied examples are outrightly irrelevant: your code uses  
4-space indents (not supported), when that value is doubled,  
any code construct that looks like a method declaration is  
highlighted as such; here, `return ... ()`.  Try `:setl sw=2`  
or `:setl sw=8` and reformat all lines with `=`.  Again, this  
PR hasn't changed anything related to this behaviour.

(Maybe "consistently indented by either" instead of "always  
indented by either" would be a better wording.  Anyway, file  
an issue if that change in the documentation is needed.)

In general, I would suggest to instead consider using `style`  
(or `signature` that is planned for the next PR).  The simple  
syntax pattern of `indent` works when you can afford flouting  
conventions or when there are no nested types with their own  
method declarations.

Shane-XB-Qian:
Actually I was thinking why it not just highlight the methods in class which is a similar cfg like the one in c/cpp.
// I did not say your PR changed something to it, but just felt that cfg originally was weird, perhaps make one or default to make the method be highlighted was good.

--
shane.xb.qian

zzzyxwvut:
Eh, similar to this:

https://github.com/vim/vim/blob/3f821d6de2586d921fb23e2facb4764ef9eb3294/runtime/syntax/c.vim#L447-L449

and that:

https://github.com/vim/vim/blob/3f821d6de2586d921fb23e2facb4764ef9eb3294/runtime/syntax/cpp.vim#L16-L17

?

Since I'm not the original syntax file author, I can only  
speculate.  A 1998 [version](https://raw.githubusercontent.com/vim/vim-history/v5.0/syntax/java.vim) of `java.vim` has only `indent`  
disguised as `style`.  When in 1997 the support for inner  
classes in Java had been introduced, `style` was [added](https://raw.githubusercontent.com/vim/vim-history/v5.4/runtime/syntax/java.vim) in  
1999.

There is no method related keyword in Java.

There are no `class` or `interface` `:syn-region`s defined for  
`javaFuncDef` to be included in.  If we had them, I suspect,  
`:syn-sync` would demand a large value from `g:java_minlines`  
for humongous classes, e.g. [`java.lang.Character`](https://raw.githubusercontent.com/openjdk/jdk/jdk-23%2B20/src/java.base/share/classes/java/lang/Character.java).

As to why the method regexp patterns are complicated, it is  
because Java is very thrifty with its keywords and permits  
complex types that can contain methods that can contain  
types that can contain methods that can contain types and so  
on.  So, there are package-private methods (that is, neither  
`public`, nor `protected`, nor `private`); there are `abstract` or  
`final` or `strictfp` methods and `class`es.  In addition,  
`else if (` or `new [Ff]oo(` are not helping either.

Shane-XB-Qian:
yes, i meant `c_functions`, 
and as above example, regardless the keywords, but just hl following:
```
StrToDate
DateToString
utilDateToSqlDate
```
or if it is a var: `public static final SimpleDateFormat SIMPLE_DATE_FORMAT = new xxxxxx;` then hl `SIMPLE_DATE_FORMAT` as well // tho maybe different hl color.

i am not sure if that make sense, or make sense to you, perhaps useful, or let it be~ thanks anyway.
// ps. yes, i hear java changed a lot, i knew a bit java approx in 2003~2008, looked it changed a lot later, :-)

zzzyxwvut:
You do realise that `g:c_functions` highlights both function   
_call sites_ and _their declarations_ from about `v9.1.0047`,  
whereas `g:java_highlight_functions` only highlights _method  
declarations_ from about `v5.0`.  Could you let me in on this  
cunning plan of yours? :)

With methods that bear `camelCase` _and_ `PascalCase` names  
indifferently, I wonder how did you name your own Java types  
back in ‘2003-8’?

Steer clear from "indent", its _simultaneous_ support of 2-  
or 8-space or tab indented methods is a mistake: with just  
4x 2-space indents, there will be misidentified constructs  
highlighted as method declarations.  (Before `v5.2`, it was  
strictly 2 spaces or a tab.)

Follow naming conventions and use whatever indentation you  
like with "style".  Else, you are free to exercise the third    
option _yourself_:

https://github.com/vim/vim/blob/38b9f45253f582ab63174376e321092f8a9a7808/runtime/doc/syntax.txt#L2023-L2026

Shane-XB-Qian:
> You do realise that `g:c_functions` highlights both function   
> _call sites_ and _their declarations_ from about `v9.1.0047`,  
> whereas `g:java_highlight_functions` only highlights _method  
> declarations_ from about `v5.0`.  Could you let me in on this  
> cunning plan of yours? :)

donot understand what're you trying to say,
`c_functions` i recalled it was adding (by Luca) near after https://github.com/vim/vim/issues/13641 which we corrected something for c/cpp syntax at that time around.

> With methods that bear `camelCase` _and_ `PascalCase` names  
> indifferently, I wonder how did you name your own Java types  
> back in ‘2003-8’?

if you were arguing the naming or something of that example, ok, yes, i copied some piece somewhere just for AS IT IS an example, it definitely was not me wrote it at 2003~2008.

so or whatever, it's on you, you can think about this idea, or let it be, but for now to me i think this g:java_highlight_functions looks a bit difficult or confused to use, :-)

-- 
shane.xb.qian

PR #14622, opened 2024-04-23
compile_def_function() is too long.  Move out the code to compile the body of a function

yegappan:


chrisbra:
thanks!

chrisbra:
Hi Yegappan, 
Coverity complains about this one here however: https://scan5.scan.coverity.com/#/project-view/41242/10101?selectedIssue=1596610

I think this is a false positive?

vim-ml:
Hi Christian,

On Tue, Apr 23, 2024 at 11:46 PM Christian Brabandt <
***@***.***> wrote:

> Hi Yegappan,
> Coverity complains about this one here however:
> https://scan5.scan.coverity.com/#/project-view/41242/10101?selectedIssue=1596610
>
> I think this is a false positive?
>
>
> Yes.  This is a false positive.  I ran a few tests to compile a def
function with different types
of comments under valgrind to make sure that the vim9_bad_comment()
function doesn't
perform an out-of-bounds access.  I also compared the code before and after
the
refactoring to make sure that the logic is not changed.

Regards,
Yegappan

chrisbra:
Thanks. Yeah, I starred at it for a while, but it wasn't clear to me how the buffer could overrun. Thanks for double checking!

PR #14624, opened 2024-04-23
Improvements to the alloc

ghost:
Improvements made:

- Consolidated the conditional expression for determining the index in the allocation/free size arrays.
- Utilized ternary operators for concise conditional assignments.
- Removed redundant comments to emphasize readability through code clarity.
- Simplified the memory peak update using a ternary expression.

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -19,7 +19,7 @@
 
 #if defined(MEM_PROFILE) || defined(PROTO)
 
-# define MEM_SIZES  8200
+#define MEM_SIZES 8200
 static long_u mem_allocs[MEM_SIZES];
 static long_u mem_frees[MEM_SIZES];
 static long_u mem_allocated;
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, but please don't change the function definition style. A function definition always starts in column 0, it makes it easy to grep for the definition of a function.

k-takata:
Hmm, I don't see practical improvements in this PR.
Whether using conditional operators or if statements is mostly subjective issue. Recent compilers may generate the same code for both of them.

src/alloc.c
```diff
@@ -19,7 +19,7 @@
 
 #if defined(MEM_PROFILE) || defined(PROTO)
 
-# define MEM_SIZES  8200
+#define MEM_SIZES 8200
```
k-takata, line 22: Please don't remove the indentation of `define`.

src/alloc.c
```diff
@@ -28,106 +28,98 @@ static long_u mem_peak;
 static long_u num_alloc;
 static long_u num_freed;
 
-    static void
+static void
```
k-takata, line 31: Please don't remove the indentation.

src/alloc.c
```diff
@@ -28,106 +28,98 @@ static long_u mem_peak;
...
 
-    static void
-mem_post_alloc(
-    void **pp,
-    size_t size)
+static void mem_post_alloc(void **pp, size_t size)
 {
     if (*pp == NULL)
-	return;
+        return;
```
k-takata, line 45: Please don't change the indentation from tab to space.

PR #14627, opened 2024-04-24
Calling CLEAR_FIELD() on the same struct twice

zeertzjq:
Problem:  Calling CLEAR_FIELD() on the same struct twice.
Solution: Remove the second CLEAR_FIELD().  Move the assignment of
          cookie.sourceing_lnum.

### DIFF
--- a/src/scriptfile.c
+++ b/src/scriptfile.c
@@ -1273,7 +1273,7 @@ cmd_source(char_u *fname, exarg_T *eap)
 	    emsg(_(e_argument_required));
 	else
 	    // source ex commands from the current buffer
-	    do_source_ext(NULL, FALSE, FALSE, NULL, eap, clearvars);
+	    do_source_ext(NULL, FALSE, DOSO_NONE, NULL, eap, clearvars);
     }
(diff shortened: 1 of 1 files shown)


src/scriptfile.c
```diff
@@ -1273,7 +1273,7 @@ cmd_source(char_u *fname, exarg_T *eap)
 	    emsg(_(e_argument_required));
 	else
 	    // source ex commands from the current buffer
-	    do_source_ext(NULL, FALSE, FALSE, NULL, eap, clearvars);
+	    do_source_ext(NULL, FALSE, DOSO_NONE, NULL, eap, clearvars);
```
chrisbra, line 1276: This change should be testable

chrisbra:
Thanks, however I am not yet sure this is correct.

src/scriptfile.c
zeertzjq: `DOSO_NONE` and `FALSE` are the same value

chrisbra:
Thanks, looks good now, I was able to verify it now.

PR #14628, opened 2024-04-24
ci.yml: remove repo microsoft-prod.list

pheiduck:
This is added by default, and it is often broken, but we don't need anything from it.


chrisbra:
Thanks.

Issue #14626, opened 2024-04-24
Toggle comments in code


girishji:
This is a RFC. The proposal is to include a script in `autoload/` that provides a utility function to toggle comments in code. Something along the lines of https://github.com/habamax/.vim/blob/master/autoload/comment.vim (or tpope's vim-commentary, although I prefer habamax's vim9 solution).

chrisbra:
That sounds useful. @habamax would you be okay with this?

habamax:
sure, will prepare PR

PR #14634, opened 2024-04-25
Add comment toggling

habamax:
Refers to #14626


chrisbra:
Thanks. Can you add this:
```
	vim9script
	import autoload 'dist/comment.vim'
	nnoremap <silent> <expr> gc comment.Toggle()
	xnoremap <silent> <expr> gc comment.Toggle()
	nnoremap <silent> <expr> gcc comment.Toggle() .. '_'
```

into a new file `pack/dist/opt/comment/plugin/comment.vim`? 

Then we can add the following to the documentation:

```
 Load the plugin with this command: >
    packadd comment
<
This way you can use the plugin with the default key bindings `gc` and similar for commenting (which is a well-established mapping established in the Vim community). 

To have more control over the mappings used, add the following snippet to your .vimrc (assuming vim9 script): >
	vim9script
	import autoload 'dist/comment.vim'
	nnoremap <silent> <expr> gc comment.Toggle()
	xnoremap <silent> <expr> gc comment.Toggle()
	nnoremap <silent> <expr> gcc comment.Toggle() .. '_'
```

Something like this? 

Can you please also add yourself in the header of the autoload script as upstream author? And please also as [Codeowner](https://github.com/vim/vim/blob/master/.github/CODEOWNERS). We may also need to update the [Filelist](https://github.com/vim/vim/blob/master/Filelist)

clason:
Would it make sense to move the `autoload` to `pack/dist/opt/comment` as well then?

habamax:
It doesn't feel right to add help entry into filetype.txt though.

clason:
> It doesn't feel right to add help entry into filetype.txt though.

We put it in various.txt: https://github.com/neovim/neovim/blob/a736e845a48c5ccdcfeb4ea485aa859a04b35d59/runtime/doc/various.txt#L560-L597

chrisbra:
Thanks. I would move it to around here: https://github.com/vim/vim/blob/98b12ede31754071f36fb7a73324456c1ee7b89c/runtime/doc/usr_05.txt#L440 so it's close to the editorconfig and matchit install instructions.

We should also document the use of 'commentstring' option. Also that it won't do anything if this option isn't set for a particular filetype and how to set this to make `gc` work. The Neovim doc looks quite clear in this regard.

I am wondering, whether we should add help-tags for `gc`. That will certainly conflict with tpopes commentary plugin. So not sure if this is a particular good idea or not.

clason:
> I am wondering, whether we should add help-tags for gc. That will certainly conflict with tpopes commentary plugin. So not sure if this is a particular good idea or not.

We did, because otherwise `:h gc` won't work because there are tags with a higher score. And conflicts (E154) are only triggered by duplicate tags in the same directory? @zeertzjq

zeertzjq:
Since this is an optional plugin, maybe mapping tags like `gc` should be added to `pack/dist/opt/comment/doc/` instead, so that they are only found when the plugin is loaded.

Shane-XB-Qian:
> I am wondering, whether we should add help-tags for `gc`. That will certainly conflict with tpopes commentary plugin. So not sure if this is a particular good idea or not.

please let the choice back to users, there lots wonderful vim plugins, it should be not possible to add them all within vim,
besides some users specifically may just like tpopes' commentary (e.g me), so that conflict may not be liked by amount of people.

-- 
shane.xb.qian

habamax:
> We should also document the use of 'commentstring' option. Also that it won't do anything if this option isn't set for a particular filetype and how to set this to make gc work. The Neovim doc looks quite clear in this regard.

If it would wait another week or so, or anyone else would be able to add this documentation.

Shane-XB-Qian:
> > please let the choice back to users, there lots wonderful vim plugins, it should be not possible to add them all within vim,
> > besides some users specifically may just like tpopes' commentary (e.g me), so that conflict may not be liked by amount of people.
> 
> It's completly optional. You will need to use `packadd` to use it. Or simply use your favorite other plugin. And it makes sense, to add the plugin specific documentation into the package itself. 

yes, if really added this within vim, i think @zeertzjq 's suggestion should be acceptable; or leave it alone, this commentary plugin looked was not so essential, visual-selection operation can do it also.

-- 
shane.xb.qian

chrisbra:
Thanks all. I have included it and adjusted the documentation.

@habamax I made a simple change to the plugin, so that it doesn't error out for non-modifiable buffers: https://github.com/vim/vim/blob/5400a5d4269874fe4f1c35dfdd3c039ea17dfd62/runtime/pack/dist/opt/comment/autoload/comment.vim#L18

ColinKennedy:
Related: https://github.com/neovim/neovim/pull/28176

habamax:
Thanks, @chrisbra!

girishji:
Thank you, @habamax and @chrisbra :)

Issue #14629, opened 2024-04-24
Wrong DisplayVersion registry key written in Windows 11


jonathan-b-wiebe:
### Steps to reproduce

1. Open a terminal and install vim using the command 'winget install vim.vim'. As of today Vim 9.1.0312_x64.exe is downloaded and installed.
2. Type the command 'winget upgrade'. The results will show that version 9.1.0 is installed and 9.1.0312 is available.
3. Open vim. This will show that installed version is indeed 9.1.0312.
4. Open the registry editor and go to the key: Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Vim 9.1
5. Note that the DisplayVersion is set t0 9.1.0. This is also the source of the version shown in Settings -> Apps -> Installed Apps.
6. Manually installing the latest version from the vim-win32-installer site (9.1.0368) will update the registry key to 9.1.0000, which is still incorrect.

### Expected behaviour

1. Winget should show that version 9.1.0312 is installed and not show that an update is available.
2. Settings -> Apps -> Installed Apps should show version 9.1.312 is installed.
3. The registry key should be set to 9.1.0312.

### Version of Vim

9.1.0312

### Environment

Windows 11-23H2

chrisbra:
Isn't that a problem with winget then, if it works when manually installing?

jonathan-b-wiebe:
Manually installing the latest version from the vim-win32-installer site (9.1.0368) will update the registry key to 9.1.0000, which is still incorrect. The registry key should be set to 9.1.0312.

chrisbra:
I don't think this is true, here on my current system (Win10), it shows the correct version:
![grafik](https://github.com/vim/vim/assets/244927/20b17f15-28ca-4afb-872c-613a05b992ae)
![grafik](https://github.com/vim/vim/assets/244927/861b34e0-1358-405d-981e-a7e032e6d59d)

jonathan-b-wiebe:
I have re-installed using the same version you are running (9.1.0136) and it indeed works correctly.
I will uninstall and reinstall with newer versions to determine when this behavour started.
Thanks for your help!

jonathan-b-wiebe:
I can now confirm that this behaviour started somewhere between version 9.1.0282 and 9.1.0296. I cannot get move exact then this because these are two consecutive released on the vim-win32-installer site and I am not set up to compile vim in order to test the intermediate versions between these.
[...]

PR #14635, opened 2024-04-25
Correct version number in Windows Registry

RestorerZ:
Fixed display of version number with patch level in Windows register.
issue #14629  

Other:
changed contents of “Publisher” and “CompanyName” fields to “The Vim Project”
for uniformity.


chrisbra:
thanks!

PR #14642, opened 2024-04-26
fix: wrong botline in BufEnter

tomtomjhj:
Problem:
When `:edit`ing an existing buffer, `line('w$')` returns a wrong result.

`do_ecmd` reinitializes the current the current window (`curwin_init()`) whose `w_valid` may have `VALID_BOTLINE`. Resetting `w_botline` without marking it as invalid makes subsequent `validate_botline()` no-op, resulting in wrong `line('w$')` value.

Solution:
Reset `w_valid`.


chrisbra:
thanks. I think this makes sense.

PR #14646, opened 2024-04-26
runtime(jq): update tags for jq syntax

pheiduck:


chrisbra:
I thought I re-generated those tags, but apparently I forgot.  Thanks!

PR #14647, opened 2024-04-26
1-second delay after Test_BufEnter_botline()

zeertzjq:
Problem:  1-second delay after Test_BufEnter_botline().
Solution: Wipe the created buffers.


chrisbra:
thanks!

PR #14652, opened 2024-04-27
Count just enough lines for half page scrolling

### LINKED ISSUE #14650: High CPU usage on Fedora 40
### Steps to reproduce

1. Open one log file, here is an example of mine, the complete file size is about 250M
[log.txt](https://github.com/vim/vim/files/15136953/log.txt)
2. Press Ctrl-D to continue scrolling down.
3. CPU usage reaches 100%
Screenshot from 2024-04-27 11-02-12

flamegraph:
flamegraph

### Expected behaviour

CPU usage should be at a low level.

### Version of Vim

9.1.309

### Environment

OS: fedora 40
terminal: gnome-terminal-3.50.1-6.fc40.x86_64
$TERM: xterm-256color

### Logs and stack traces

_No response_

luukvbaal:
Problem:  Calculating line height for unnecessary amount of lines with
          half-page scrolling.
Solution: Replace "limit_winheight" argument with higher resolution
          "max" argument to which to limit the calculated line height
          in plines_m_win() to.

Fix #14650

### DIFF
--- a/src/drawscreen.c
+++ b/src/drawscreen.c
@@ -1781,7 +1781,7 @@ win_update(win_T *wp)
 	    if (j < wp->w_height - 2)		// not too far off
 	    {
 		i = plines_m_win(wp, wp->w_topline, wp->w_lines[0].wl_lnum - 1,
-									 TRUE);
+								wp->w_height);
 #ifdef FEAT_DIFF
 		// insert extra lines for previously invisible filler lines
 		if (wp->w_lines[0].wl_lnum != wp->w_topline)
--- a/src/misc1.c
+++ b/src/misc1.c
@@ -497,12 +497,17 @@ plines_win_col(win_T *wp, linenr_T lnum, long column)
     return lines;
 }
 
+/*
+ * Return number of window lines the physical line range from "first" until
+ * "last" will occupy in window "wp". Takes into account folding, 'wrap',
+ * topfill and filler lines beyond the end of the buffer. Limit to "max" lines.
+ */
     int
-plines_m_win(win_T *wp, linenr_T first, linenr_T last, int limit_winheight)
+plines_m_win(win_T *wp, linenr_T first, linenr_T last, int max)
 {
     int		count = 0;
 
-    while (first <= last && (!limit_winheight || count < wp->w_height))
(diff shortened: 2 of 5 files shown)


luukvbaal:
Could also (or instead of, but the "max" argument seems useful even for the other current `limit_winheight == FALSE` instances) add this cheap check that determines whether calling `plines_m_win()` is even necessary to begin with:
```diff
@@ -3219,7 +3219,8 @@ pagescroll(int dir, long count, int half)

        int curscount = count;
        // Adjust count so as to not reveal end of buffer lines.
-       if (dir == FORWARD)
+       if (dir == FORWARD
+                   && curwin->w_topline + count > buflen - curwin->w_height)
        {
            int n = plines_correct_topline(curwin, curwin->w_topline, FALSE);
            if (n - count < curwin->w_height && curwin->w_topline < buflen)
```

luukvbaal:
Could also add this cheap check that determines whether calling `plines_m_win()` is even necessary to begin with, although that only works if the window has no folds:
```diff
@@ -3219,7 +3219,8 @@ pagescroll(int dir, long count, int half)

        int curscount = count;
        // Adjust count so as to not reveal end of buffer lines.
-       if (dir == FORWARD)
+       if (dir == FORWARD
+                   && curwin->w_topline + count > buflen - curwin->w_height)
        {
            int n = plines_correct_topline(curwin, curwin->w_topline, FALSE);
            if (n - count < curwin->w_height && curwin->w_topline < buflen)
```
@@ -3219,7 +3219,8 @@ pagescroll(int dir, long count, int half)

        int curscount = count;
        // Adjust count so as to not reveal end of buffer lines.
-       if (dir == FORWARD)
+       if (dir == FORWARD && (hasAnyFolding(curwin)
+                   || curwin->w_topline + count >= buflen - curwin->w_height))
        {
            int n = plines_correct_topline(curwin, curwin->w_topline, FALSE);
            if (n - count < curwin->w_height && curwin->w_topline < buflen)

zhscn:
I just applied this PR and the CPU usage goes down to 3.5%. 👍

chrisbra:
thanks!

PR #14658, opened 2024-04-28
Update asm.vim

ghost:
E492: Not an editor command: indentexpr< indentkeys<


chrisbra:
weird, I though I fixed that previously. Thanks, obviously right.

PR #14659, opened 2024-04-28
runtime(java): Improve the recognition of the "indent" method declarations

zzzyxwvut:
There is a flaw in the current implementation that has been  
exacerbated around `v5.2`.  It lies in the recognition of all  
three indentation styles simultaneously: a tab, two space,  
and eight space character(s).  With it, it is not uncommon  
to misidentify various constructs as method declarations  
when they belong to two-space indented members and other  
blocks of a type and are offset at eight space characters or  
a tab from the start of the line.

For example,

```java
class Test
{
  static String hello() { return "hello"; }

  public static void main(String[] args)
  {
    try {
      if (args.length > 0) {
        // FIXME: eight spaces.
        System.out.println(args[0]);
      } else {
        // FIXME: a tab.
	System.out.println(hello());
      }
    } catch (Exception e) {
      throw new Error(e);
    }
  }
}
```

```vim
:let g:java_highlight_functions = 'indent'
:doautocmd Syntax
```

A better approach is to pick an only indentation style out  
of all supported styles (so either two spaces _or_ eight  
spaces _or_ a tab).  Note that tabs and spaces can still be  
mixed, only the leading tab or the leading run of spaces  
matters for the recognition.  And there is no reason to not  
complement the set of valid styles with any number of spaces  
from `1` to `8`, inclusively.

Please proceed with the necessary change as follows:

- rename from "indent" to "indent2" for a 2-space run;
- rename from "indent" to "indent8" for an 8-space run;
- continue to have "indent" for a tab run;
- define an "indent" variable with a suffix number denoting  
  the preferred amount of indentation for any other run of  
  spaces `[1-8]`.

As before, this alternative style of recognition of method  
declarations still does not prescribe naming conventions and  
still cannot recognise method declarations in nested types  
that are conventionally indented.

The proposed changes also follow suit of "style" in stopping  
the claiming of constructor and enum constant declarations.


chrisbra:
thanks, makes sense!

PR #14619, opened 2024-04-22
Adding JQ Syntax to vim!

vito-c:
I am adding jq syntax to vim as well as updating the maintainer for json.
@itchyny @dtolnay @Freed-Wu

PR #14661, opened 2024-04-29
runtime(jq): fix undefined variable s:save_cpoptions in ftplugin

itchyny:
This PR fixes an issue about undefined variable in the jq ftplugin file. It is recommended not to setting 'cpo' if unnecessary (https://github.com/vim/vim/pull/14619#discussion_r1575249285), I removed the resetting lines instead of adding `let s:save_cpooptions = &cpooptions` and `set cpo&vim`.


chrisbra:
thanks, see also #14663 I'll merge those two together :)

PR #14663, opened 2024-04-29
completely removing cpoptions and adding support for inlcude

vito-c:
@chrisbra There was a piece that I forgot to remove the previous PR that will break the jq syntax file. Here is the fix. I also added the includes.


chrisbra:
Thanks, see also #14661 I'll merge those two manually (and also add the include option to the undo_ftplugin).

PR #14670, opened 2024-04-30
fix: tex should not take '.out' ext

Shane-XB-Qian:
// '.out' should be a very common ext,
// i think and supposed 'tex' should not or was not taking it as belonging to it.

related: https://github.com/vim/vim/pull/14456

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2413,7 +2413,7 @@ au BufNewFile,BufRead texdoc.cnf		setf conf
 " LaTeX packages will generate some medium LaTeX files during compiling
 " They should be ignored by .gitignore https://github.com/github/gitignore/blob/main/TeX.gitignore
 " Sometime we need to view its content for debugging
-au BufNewFile,BufRead *.{pgf,nlo,nls,out,thm,eps_tex,pygtex,pygstyle,clo,aux,brf,ind,lof,loe,nav,vrb,ins,tikz,bbx,cbx,beamer}	setf tex
+au BufNewFile,BufRead *.{pgf,nlo,nls,thm,eps_tex,pygtex,pygstyle,clo,aux,brf,ind,lof,loe,nav,vrb,ins,tikz,bbx,cbx,beamer}	setf tex
 
 " ConTeXt
 au BufNewFile,BufRead *.mkii,*.mkiv,*.mkvi,*.mkxl,*.mklx   setf context
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -750,7 +750,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     teraterm: ['file.ttl'],
     terminfo: ['file.ti'],
     'terraform-vars': ['file.tfvars'],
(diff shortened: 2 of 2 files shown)


clason:
If it's such a common extension, why hasn't it been recognized then? If not TeX, what _should_ these files be? Please motivate your change by an actual problem instead of "previous change bad! revert!"

Shane-XB-Qian:
> If it's such a common extension, why hasn't it been recognized then? If not TeX, what _should_ these files be? Please motivate your change by an actual problem instead of "previous change bad! revert!"

interesting logic or mindset, have you ever do `gcc foo.c` before?

-- 
shane.xb.qian

clason:
> If it's such a common extension, why hasn't it been recognized then? If not TeX, what _should_ these files be? Please motivate your change by an actual problem instead of "previous change bad! revert!"
> interesting logic or mindset, have you ever do `gcc foo.c` before?
> […](#)
> -- shane.xb.qian

Yes, and have you ever done `vim a.out` before? And has this regressed now? (I'm not saying these files _should_ be recognized as TeX -- in fact, they're usually empty -- but you still should be able to give a better motivation than "**I** think this should not be done.")

Shane-XB-Qian:
> attempt to provide some content sanity check.

i guess it simply was a mistake of @Freed-Wu , '.out' should be nothing related to tex, i mean even the content.
// or if it did, then @Freed-Wu

-- 
shane.xb.qian

Shane-XB-Qian:
> > If it's such a common extension, why hasn't it been recognized then? If not TeX, what _should_ these files be? Please motivate your change by an actual problem instead of "previous change bad! revert!"
> > interesting logic or mindset, have you ever do `gcc foo.c` before?
> > […](#)
> > -- shane.xb.qian
> 
> Yes, and have you ever done `vim a.out` before? And has this regressed now? (I'm not saying these files _should_ be recognized as TeX -- in fact, they're usually empty -- but you still should be able to give a better motivation than "**I** think this should not be done.")

please find someone else to argue, and this is just a correction for official vim, you can do whatever you like.

clason:
We all want Vim's filetype detection to be as complete and correct as possible. Asking for motivation of changes so their effect can be gauged against the intent is part of that. (And, as a personal observation, your transparent attempts at gatekeeping "official vim" by adding friction to other contributors are getting hard to take. I wish you could be more constructive in your comments.)

Again, is there a better filetype for `.out` files that can be recognized based on contents? Can't we do better than simply not recognizing them at all?

Shane-XB-Qian:
> your transparent attempts at gatekeeping "official vim" by adding friction to other contributors are getting hard to take.

i donot think so, i weclome every nice and wonderful contribution to official vim and the past decision which Bram did.
and everyone is free and has right to talk, 
just please donot throw your mood to me.

chrisbra:
Yeah, using `.out` for `tex` was probably an error. Let's not set a filetype for this one, since it can be anything and probably for some users it might be more typically filetype a while others may encounter it as totally different filetype. 

I have used out files as a generic text file in the past, but that may as well be wrong for other users.

PR #14674, opened 2024-04-30
portuguese localisation

lesorubshayan:
close https://github.com/vim/vim/pull/10718
close https://github.com/vim/vim/pull/10717


ronin49:
@chrisbra

chrisbra:
Same comment as for #14685 
- Move the added translation to before the line resetting 'cpo' 
- Change the commit author

Thanks!

ronin49:
@chrisbra

chrisbra:
thanks!

ychin:
I just noticed this change when merging upstream for MacVim and seeing some merge conflicts since MacVim uses some Apple official translations for some items, and in this change the pt_br (Brazil) translations now just source the pt_pt (Portugual) one.

I don't speak Portuguese myself but I do have some questions @ronin49 . Don't the two locales (pt_pt and pt_br) use different words for the same concept? For example it seems that both the Apple official translations and the old Vim menu translations (and also Google Translate etc) use "Salvar" for "Save" in Brazil, but "Guardar" in Portugual.

Was there a reason why this merge was done? Is it because the Brazilian version is no longer maintained?

chrisbra:
please keep the existing translations in this case. I only noticed later, that the PRs from @ronin49 were mostly machine generated. 

You may even want to create a PR to re-use the existing Apple provided translations so that the merge conflict will be fixed for you in the future.

ychin:
> You may even want to create a PR to re-use the existing Apple provided translations so that the merge conflict will be fixed for you in the future.

It's a little hard to do that right now because MacVim uses some custom menu items that are MacVim specific (e.g. a link to MacVim release notes, or an "About MacVim" item) and I'm not sure if they make sense to be in Vim, and also Apple translations could be different from how other platforms translate things. I could think about that more as I do want to upstream more MacVim stuff to reduce merge conflicts though (there are a fair bit of MacVim-specific changes that were made a long time ago that seemed to be made mostly out of convenience of not having to go through Vim/Bram).

In this case the merge conflicts were easy to resolve anyway (I just have an externally generated file that I add a line to include them), and I was more confused why the Portuguese translations were merged.

> please keep the existing translations in this case. I only noticed later, that the PRs from @ronin49 were mostly machine generated.

I could submit a PR to do that.

ychin:
See #15557 for my revert and comments.

PR #14676, opened 2024-04-30
filetype: Kbuild files are not recognized

ambroisie:
Problem:  Kbuild files are not recognized.

Solution: Detect Kbuild files as make files.


chrisbra:
thanks!

PR #5566, opened 2020-02-02
look up keywords by hitting 'K' in batch and shell files

Konfekt:
Just left here for reference after a month without response from the respective maintainer's e-mail addresses.

PR #14679, opened 2024-05-01
ftplugin: sh: don't assume MANPAGER makes sense

benknoble:
Say you use Vim and set MANPAGER='vim -M +MANPAGER --not-a-term -'; then :ShKeywordPrg (or K) will crap out and spew terminal garbage into less when bash's "help" fails. This was introduced by 2f25e40b1 (runtime: configure keywordpg for some file types (#5566), 2023-08-23) and may be present in other files touched by that commit.

Make the "man" invocation sensible by unsetting MANPAGER in the environment.

### DIFF
--- a/runtime/ftplugin/sh.vim
+++ b/runtime/ftplugin/sh.vim
@@ -46,11 +46,11 @@ endif
 
 if get(b:, "is_bash", 0)
   if !has("gui_running") && executable("less")
-    command! -buffer -nargs=1 ShKeywordPrg silent exe '!bash -c "{ help "<args>" 2>/dev/null || man "<args>"; } | LESS= less"' | redraw!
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks. Does it make sense to use instead:
```
unlet $MANPAGER
```

benknoble:
> Thanks. Does it make sense to use instead:
> 
> ```
> unlet $MANPAGER
> ```

I doubt it: that would mess with the user's environment for that Vim, right? In other words, there's not a good way to scope that to just this command (without the save/unlet/restore block), right?

Konfekt:
Thank you @benknoble, grepping shows that this affects  zsh, udevrules, sshconfig, systemd, readline, sudoers, muttrc, modconf, gpg as well. Would you mind applying it there as well?

Ideally, one day one would employ [your solution](https://github.com/benknoble/Dotfiles/blob/c0c985866ad2cadbc7b9186d165bd2de20f09d71/links/vim/autoload/bk/gitconfig.vim#L3).

chrisbra:
Thanks, I include it

Issue #13035, opened 2023-09-05
Add support for files and directory fuzzy matching


mortezadadgar:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
it's not impossible there is a `matchfuzzy()` 

**Describe the solution you'd like**
commands like `:find` would benefits from having fuzzy completions

**Describe alternatives you've considered**
nothing I'm aware except using external plugins

**Additional context**
none

mortezadadgar:
cc @yegappan

chrisbra:
isn't that something that a plugin should provide?

mortezadadgar:
`matchfuzzy()` can be used anywhere right why not takes advantages of built-in vim features?

chrisbra:
Well, every feature comes with a cost. And correct me if I am wrong, A plugin can already receive the list of existing files and filter them according to the `matchfuzzy()` results, right? So I would think this is a good opportunity for a plugin.

mortezadadgar:
yea I assume a plugin can takes advantages of this api, what do you think about having a `fuzzy` option for completeopt? i think that make more sense to be implemented as a feature

glepnir:
@chrisbra  can close now.

zeertzjq:
I don't think there is any fuzzy matching that applies to `:find`

PR #14694, opened 2024-05-01
feat(completion): add two highlight groups PmenuMatch and PmenuMatchSel

glepnir:
Add two highlights group `PmenuMatch` and `PmenuMatchSel` for the leader matched position in candidate words. 

image


techntools:
Will it work with fuzzy matching ?

glepnir:
> Thanks, can you explain what those are for? Also should be documented at `:h highlight-groups`

yup but i think not finish . i need test it within mulitple bytes characters..

glepnir:
> Will it work with fuzzy matching ?

Hello . currently the `completeopt` doesn't have fuzzy option.  if vim maintainer thought we can add it. i think it can be support. 

old

glepnir:
@chrisbra  ptal . works fine for me now.

techntools:
Please vote yes.

Suggested [here](https://github.com/vim/vim/issues/13035#issuecomment-1709652208) as well.

glepnir:
I tried it briefly.  😄 

fuzzy

src/highlight.c
```diff
@@ -257,6 +257,8 @@ static char *(highlight_init_both[]) = {
     "default link CursorLineFold FoldColumn",
     "default link CurSearch Search",
     "default link PmenuKind Pmenu",
+    "default link PmenuMatch Pmenu",
+    "default link PmenuMatchSel PmenuSel",
```
chrisbra, line 262: keep that sorted please

src/popupmenu.c
```diff
@@ -416,6 +416,71 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+                ga.ga_len++;
+            }
+            match = mb_strstr(match + leader_len, match_leader);
+        }
+    }
+
+    while (*ptr != NUL && ptr < text + textlen)
+    {
+	char_len = mb_ptr2len(ptr);
+	cells = utf_ptr2cells(ptr);
```
chrisbra: mb_ptr2cells()

src/popupmenu.c
```diff
@@ -416,6 +416,71 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+	int new_attr = attr;
```
chrisbra: move to the beginning of the block please

src/popupmenu.c
```diff
@@ -596,7 +660,8 @@ pum_redraw(void)
 				    else
 					--cells;
 				}
-				screen_puts_len(st, size, row, col, attr);
+				pum_screen_put_with_attr(row, col, st, size, attr);
+				//screen_puts_len(st, size, row, col, attr);
```
chrisbra: remove commented out line please

src/popupmenu.c
```diff
@@ -416,6 +416,71 @@ pum_under_menu(int row, int col, int only_redrawing)
 	    && col < pum_col + pum_width + pum_scrollbar;
 }
 
+/*
+ * displays text on the popup menu with specific attributes.
+ */
+    static void
+pum_screen_put_with_attr(int row, int col, char_u *text, int textlen, int attr)
```
chrisbra, line 423: can you make this interface similar to `screen_puts_len()`? 

`pum_screen_put_with_attr(text, textlen, row, col, attr)`

src/testdir/test_popup.vim
```diff
@@ -1339,4 +1339,32 @@ func Test_pum_highlights_custom()
   call StopVimInTerminal(buf)
 endfunc
 
+"Test match relate highlight group in pmenu
```
chrisbra: Missing space. Also the comment is not clear

src/optiondefs.h
```diff
@@ -300,7 +300,7 @@ struct vimoption
...
+# define HIGHLIGHT_INIT "8:SpecialKey,~:EndOfBuffer,@:NonText,d:Directory,e:ErrorMsg,i:IncSearch,l:Search,y:CurSearch,m:MoreMsg,M:ModeMsg,n:LineNr,a:LineNrAbove,b:LineNrBelow,N:CursorLineNr,G:CursorLineSign,O:CursorLineFold,r:Question,s:StatusLine,S:StatusLineNC,c:VertSplit,t:Title,v:Visual,V:VisualNOS,w:WarningMsg,W:WildMenu,f:Folded,F:FoldColumn,A:DiffAdd,C:DiffChange,D:DiffDelete,T:DiffText,>:SignColumn,-:Conceal,B:SpellBad,P:SpellCap,R:SpellRare,L:SpellLocal,+:Pmenu,=:PmenuSel,k:PmenuMatch,<:PmenuMatchSel,[:PmenuKind,]:PmenuKindSel,{:PmenuExtra,}:PmenuExtraSel,x:PmenuSbar,X:PmenuThumb,*:TabLine,#:TabLineSel,_:TabLineFill,!:CursorColumn,.:CursorLine,o:ColorColumn,q:QuickFixLine,z:StatusLineTerm,Z:StatusLineTermNC,g:MsgArea"
```
chrisbra, line 304: Those added entries should probably move to the end of the old entries.

chrisbra: This also requires to update `:h 'hl'`

src/highlight.c
glepnir: no please then :)

glepnir:
I've thought about this carefully. currently we use strncmp to compare compl_leader which always matches from the beginning of the string. so highlighting this part doesn't make sense. adding completeopt+=fuzzy would make this PR more useful. so I think I'll implement completeopt fuzzy first and then reorganize this PR

romainl:
If this is merged, please ping:

- @habamax, @neutaaaaan and me (@romainl) so that we can update the built-in colorschemes at https://github.com/vim/colorschemes where and if necessary,
- @lifepillar so that he can update https://github.com/lifepillar/vim-colortemplate.

Thanks

src/mbyte.c
```diff
@@ -1760,6 +1760,32 @@ mb_string2cells(char_u *p, int len)
     return clen;
 }
 
+
+
+/*
+ * Finds the first occurrence of the substring 'needle' in the multibyte string 'haystack'.
+ *Returns a pointer to the beginning of the substring, or NULL if the substring is not found.
```
zeertzjq:
```suggestion
 * Returns a pointer to the beginning of the substring, or NULL if the substring is not found.
```

src/optiondefs.h
glepnir: > Those added entries should probably move to the end of the old entries.

should be continuous.  tail doesn't make sense. reference https://github.com/vim/vim/commit/6a7c7749204b256e779c245b1e999bf852ad7b64

glepnir:
after https://github.com/vim/vim/commit/a218cc6cdabae1113647b817c4eefc2b60a6902f  ..this should be more useful now and i have do some update to make pr can also handle match highlight when in fuzzy .

pum

src/popupmenu.c
glepnir: currently can't

src/popupmenu.c
```diff
@@ -416,6 +416,114 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+    int		leader_len;
+    int		char_len;
+    int		cells;
+    int		new_attr;
+    char_u	*rt_leader = NULL;
+    char_u	*match_leader = NULL;
+    char_u	*ptr = text;
+    garray_T	*ga = NULL;
+    char_u	*leader = ins_compl_leader();
+    int		cur_cot_flags = get_cot_flags();
```
zeertzjq:
```suggestion
    unsigned	cur_cot_flags = get_cot_flags();
```

techntools:
@chrisbra Please merge this

src/popupmenu.c
```diff
@@ -416,6 +416,114 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+    int		char_len;
+    int		cells;
+    int		new_attr;
+    char_u	*rt_leader = NULL;
+    char_u	*match_leader = NULL;
+    char_u	*ptr = text;
+    garray_T	*ga = NULL;
+    char_u	*leader = ins_compl_leader();
+    unsigned	cur_cot_flags = get_cot_flags();
+    int		in_fuzzy = (cur_cot_flags & COT_FUZZY) != 0;
```
chrisbra:
```suggestion
    int		in_fuzzy = (get_cot_flags() & COT_FUZZY) != 0;
```
and then get rid of the cur_cot_flags please

src/search.c
```diff
@@ -5091,6 +5091,129 @@ fuzzy_match_str(char_u *str, char_u *pat)
...
+
+    score_item = list_find(retlist, 2);
+    if (score_item != NULL && score_item->li_tv.v_type == VAR_LIST)
+    {
+        list_T *score_list = score_item->li_tv.vval.v_list;
+        if (score_list->lv_len > 0)
+        {
+            listitem_T *first_score_item = score_list->lv_first;
+            if (first_score_item != NULL && first_score_item->li_tv.v_type == VAR_NUMBER)
+            {
```
chrisbra: style: remove braces around single statements

src/search.c
```diff
@@ -5091,6 +5091,129 @@ fuzzy_match_str(char_u *str, char_u *pat)
...
+    list_free(match_str_list);
+    list_free(match_pos_list);
+    list_free(match_score_list);
+    list_free(retlist);
+    list_free(l);
+    ga_clear(match_positions);
+    return NULL;
+#else
+    (void)str;
+    (void)pat;
```
chrisbra: those `(void)str;` and `(void)pat;` seems strange. I guess we don't need those and we can just return NULL instead?

src/popupmenu.c
```diff
@@ -416,6 +416,114 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+                    p[0] = (int)(match - text);     // start position
+                    p[1] = p[0] + leader_len;       // end position
+                    ga->ga_len++;
+                }
+                match = mb_strstr(match + leader_len, match_leader);
+            }
+        }
+        else
+        {
+#ifdef FEAT_SEARCH_EXTRA
```
chrisbra: I don't think we need the `#ifdef FEAT_SEARCH_EXTRA` here, since you added those ifdefs in the function as well. So we can ge trid of those here.

src/search.c
glepnir: compiler warning the param not used if so.

zeertzjq: Usually the solution is to add `UNUSED` after the function parameter in the signature.

glepnir: thanks :)

glepnir:
emm lots of windows failed Test_ColonEight_MultiByte() ..is related ?

chrisbra:
yeah, those are new failures starting on the Windows builds recently. I have no idea why shortening pathnames does no longer work on Github CI Windows runners. But it only fails with multi-byte paths. No idea why, but I assume it's an issue with the Github CI runners, since there weren't any changes in the related code.

chrisbra:
I think this looks good now. I am just slightly worried this will cause a bit of slowdown, because we are looping over the original text several times, once to determine the match positions and then to set the highlighting attributes.

src/popupmenu.c
```diff
@@ -416,6 +416,109 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+    int		leader_len;
+    int		char_len;
+    int		cells;
+    int		new_attr;
+    char_u	*rt_leader = NULL;
+    char_u	*match_leader = NULL;
+    char_u	*ptr = text;
+    garray_T	*ga = NULL;
+    char_u	*leader = ins_compl_leader();
+    int		in_fuzzy = (get_cot_flags() & COT_FUZZY) != 0;
```
chrisbra, line 435: How about this, to not having to figure out extra highlighting when not required?

```suggestion
    if (highlight_attr[HLF_PMSI] == highlight_attr[HLF_PSI] &&
       highlight_attr[HLF_PMNI] == highlight_attr[HLF_PNI])
    {
       screen_puts_len(text, textlen, row, col, attr);
       return;
    }
```

src/testdir/test_popup.vim
```diff
@@ -1341,4 +1341,33 @@ func Test_pum_highlights_custom()
...
+  CheckScreendump
+  let lines =<< trim END
+    func Omni_test(findstart, base)
+      if a:findstart
+        return col(".")
+      endif
+      return [#{word: "five"}, #{word: "fitfit"}, #{word: "first"}, #{word:  "你好"}, #{word: "你好吗"}, #{word: "你不好吗"}]
+    endfunc
+    set omnifunc=Omni_test
+    set completeopt=menu,menuone,fuzzy
```
chrisbra: can we run this test with and without `fuzzy`?

src/testdir/test_popup.vim
```diff
@@ -1341,4 +1341,33 @@ func Test_pum_highlights_custom()
...
+  call writefile(lines, 'Xscript', 'D')
+  let buf = RunVimInTerminal('-S Xscript', {})
+  call TermWait(buf)
+  call term_sendkeys(buf, "i\<C-X>\<C-O>\<BS>\<BS>\<BS>fi")
+  call TermWait(buf, 50)
+  call VerifyScreenDump(buf, 'Test_pum_highlights_03', {})
+  call term_sendkeys(buf, "\<ESC>S你")
+  call VerifyScreenDump(buf, 'Test_pum_highlights_04', {})
+  call term_sendkeys(buf, "\<ESC>S你吗")
+  call VerifyScreenDump(buf, 'Test_pum_highlights_05', {})
```
chrisbra, line 1384: all those screen-dumps do not show the popup menu at all

src/testdir/test_popup.vim
```diff
@@ -1341,4 +1341,33 @@ func Test_pum_highlights_custom()
...
+    endfunc
+    set omnifunc=Omni_test
+    set completeopt=menu,menuone,fuzzy
+    hi PmenuMatchSel  ctermfg=6 ctermbg=225
+    hi PmenuMatch     ctermfg=4 ctermbg=225
+  END
+  call writefile(lines, 'Xscript', 'D')
+  let buf = RunVimInTerminal('-S Xscript', {})
+  call TermWait(buf)
+  call term_sendkeys(buf, "i\<C-X>\<C-O>\<BS>\<BS>\<BS>fi")
```
chrisbra: This closes the popup menu after the first backspace.

chrisbra (commented):
I was just about to merging this, when I noticed the test does not work as expected and doesn't show any popup menus. Can you please fix those and also make sure the screen-dump test run once in fuzzy mode and one in regular completion mode?

In addition to your changes, I had the following changes on top, please add them while fixing the test:
```diff
diff --git a/runtime/doc/options.txt b/runtime/doc/options.txt
index 655fd6c52..1e6f313eb 100644
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1,4 +1,4 @@
-*options.txt*  For Vim version 9.1.  Last change: 2024 Jun 05
+*options.txt*  For Vim version 9.1.  Last change: 2024 Jun 10


                  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -4275,6 +4275,7 @@ A jump table for the options with a short description can be found at |Q_op|.
                                     T:DiffText,>:SignColumn,-:Conceal,
                                     B:SpellBad,P:SpellCap,R:SpellRare,
                                     L:SpellLocal,+:Pmenu,=:PmenuSel,
+                                    k:PmenuMatch,<:PmenuMatchSel,
                                     [:PmenuKind,]:PmenuKindSel,
                                     {:PmenuExtra,}:PmenuExtraSel,
                                     x:PmenuSbar,X:PmenuThumb,*:TabLine,
@@ -4341,6 +4342,8 @@ A jump table for the options with a short description can be found at |Q_op|.
        |hl-PmenuExtraSel| }  popup menu "extra" selected line
        |hl-PmenuSbar|   x  popup menu scrollbar
        |hl-PmenuThumb|  X  popup menu scrollbar thumb
+       |hl-PmenuMatch|  k  popup menu matched text
+       |hl-PmenuMatchSel| <  popup menu matched text in selected line

        The display modes are:
                r       reverse         (termcap entry "mr" and "me")
diff --git a/runtime/doc/syntax.txt b/runtime/doc/syntax.txt
index 23de60704..3f684412c 100644
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -1,4 +1,4 @@
-*syntax.txt*   For Vim version 9.1.  Last change: 2024 Jun 09
+*syntax.txt*   For Vim version 9.1.  Last change: 2024 Jun 10


                  VIM REFERENCE MANUAL    by Bram Moolenaar
diff --git a/runtime/doc/tags b/runtime/doc/tags
index 96be83ab0..3085ccc4c 100644
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -8090,6 +8090,8 @@ hl-PmenuExtra     syntax.txt      /*hl-PmenuExtra*
 hl-PmenuExtraSel       syntax.txt      /*hl-PmenuExtraSel*
 hl-PmenuKind   syntax.txt      /*hl-PmenuKind*
 hl-PmenuKindSel        syntax.txt      /*hl-PmenuKindSel*
+hl-PmenuMatch  syntax.txt      /*hl-PmenuMatch*
+hl-PmenuMatchSel       syntax.txt      /*hl-PmenuMatchSel*
 hl-PmenuSbar   syntax.txt      /*hl-PmenuSbar*
 hl-PmenuSel    syntax.txt      /*hl-PmenuSel*
 hl-PmenuThumb  syntax.txt      /*hl-PmenuThumb*
diff --git a/runtime/doc/version9.txt b/runtime/doc/version9.txt
index 6edb8891b..71e95554e 100644
--- a/runtime/doc/version9.txt
+++ b/runtime/doc/version9.txt
@@ -1,4 +1,4 @@
-*version9.txt*  For Vim version 9.1.  Last change: 2024 Jun 05
+*version9.txt*  For Vim version 9.1.  Last change: 2024 Jun 10


                  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -41596,6 +41596,9 @@ Autocommands: ~
 Highlighting: ~

 |hl-MsgArea|           highlighting of the Command-line and messages area
+|hl-PmenuMatch|                Popup menu: highlighting of matched text
+|hl-PmenuMatchSel|     Popup menu: highlighting of matched text in selected
+                       line

 Commands: ~

diff --git a/src/popupmenu.c b/src/popupmenu.c
index 1c8c91c28..f3726f6ef 100644
--- a/src/popupmenu.c
+++ b/src/popupmenu.c
@@ -434,6 +434,12 @@ pum_screen_put_with_attr(int row, int col, char_u *text, int textlen, int attr)
     char_u     *leader = ins_compl_leader();
     int                in_fuzzy = (get_cot_flags() & COT_FUZZY) != 0;

+    if (highlight_attr[HLF_PMSI] == highlight_attr[HLF_PSI] &&
+       highlight_attr[HLF_PMNI] == highlight_attr[HLF_PNI])
+    {
+       screen_puts_len(text, textlen, row, col, attr);
+       return;
+    }
 #ifdef FEAT_RIGHTLEFT
     if (leader != NULL && curwin->w_p_rl)
         rt_leader = reverse_text(leader);
diff --git a/src/testdir/test_popup.vim b/src/testdir/test_popup.vim
index 434d86e2d..db7556d95 100644
--- a/src/testdir/test_popup.vim
+++ b/src/testdir/test_popup.vim
@@ -1357,7 +1357,7 @@ func Test_pum_highlights_match()
     hi PmenuMatch     ctermfg=4 ctermbg=225
   END
   call writefile(lines, 'Xscript', 'D')
-  let buf = RunVimInTerminal('-S Xscript', {})
+  let  buf = RunVimInTerminal('-S Xscript', {})
   call TermWait(buf)
   call term_sendkeys(buf, "i\<C-X>\<C-O>\<BS>\<BS>\<BS>fi")
   call TermWait(buf, 50)
diff --git a/src/vim.h b/src/vim.h
index d32d661d1..e70323937 100644
--- a/src/vim.h
+++ b/src/vim.h
@@ -1501,7 +1501,7 @@ typedef enum
     , HLF_PNI      // popup menu normal item
     , HLF_PSI      // popup menu selected item
     , HLF_PMNI     // popup menu matched text in normal item
-    , HLF_PMSI     // popup menu mathced text in selected item
+    , HLF_PMSI     // popup menu matched text in selected item
     , HLF_PNK      // popup menu normal item "kind"
     , HLF_PSK      // popup menu selected item "kind"
     , HLF_PNX      // popup menu normal item "menu" (extra text)
```

src/popupmenu.c
glepnir: this will cause Match highlight only work with fuzzy.

glepnir: I have do some cleanup and update the description of Match relate highlight. they are only make sense when in fuzzy.

src/popupmenu.c
```diff
@@ -416,6 +416,80 @@ pum_under_menu(int row, int col, int only_redrawing)
...
+    char_u	*rt_leader = NULL;
+    char_u	*match_leader = NULL;
+    char_u	*ptr = text;
+    garray_T	*ga = NULL;
+    char_u	*leader = ins_compl_leader();
+
+    if ((get_cot_flags() & COT_FUZZY) == 0)
+    {
+       screen_puts_len(text, textlen, row, col, attr);
+       return;
```
chrisbra: I think we should also skip this, if the HLF_PMSI and HLF_PMNI highlightings are not setup?

glepnir: yup also need compare attr if same as PSI also need run into this if.

chrisbra:
so *PmenuMatch*  and *PmenuMatchSel*  highlighting only works for when fuzzy completeopt is enabled? I think it should also work when `fuzzy` has not been enabled yet.

glepnir:
emm if you want . I can restore some codes of before commit to make it rework. honestly it not make sense when there is no fuzzy. because we only use prefix match (strncmp(item ,leader, leader_len) that mean it always highlight the first few letters of each word.

chrisbra:
True, but the expectation from those newly introduced highlighting groups is, that they highlight the matching pattern, even if it is only a substring highlight I would think. So yes please enable a simple substring highlighting for the non-fuzzy case then

glepnir:
this is picture is the first implementation .  the point is item2 have two `fi` position. they both should highlight then 

image


the before function on there is 


```C
   char_u *
mb_strstr(char_u *haystack, char_u *needle)
{
    char_u *h = haystack;
    size_t needle_len = STRLEN(needle);

    if (!needle_len)
	return haystack;  // Empty needle.

    while (*h != NUL)
    {
	// Check if the needle matches here
	if (STRNCMP(h, needle, needle_len) == 0)
	    return h;
	h += mb_ptr2len(h);  // Move to the next character
    }
    return NULL;
}
```

chrisbra:
I think only the first pattern match should be fine, so a simple subtring match is fine. I don't see the point of highlighting additional matches further down in the matches, because those are not considered when filtering the candidates. Default completion only works with single submatch filtering.

glepnir:
done @chrisbra  PTAL

neutaaaaan:
I'll try and test drive this during the week.
I agree with @glepnir that there is little to no point in highlighting partial matches outside of fuzzy matching, and I'm somewhat wary of how we will have to tweak the 8 and 16c versions of the colorschemes to handle this properly, if possible at all, but the feature itself is definitely pertinent.

chrisbra:
Nice, thanks for updating the test!

chrisbra:
@glepnir unfortunately, this has introduced a few issues with Coverity in `fuzzy_match_str_with_pos()`

can you please have a look?

```c
5199 cleanup:
5200    vim_free(tv_str.vval.v_string);
// CID 1603599: (#2 of 2): Dereference after null check (FORWARD_NULL)
// var_deref_model: Passing null pointer match_str_list to list_free, which dereferences it.["show details"]
5201    list_free(match_str_list);
// CID 1603601: (#2 of 2): Dereference after null check (FORWARD_NULL)
// var_deref_model: Passing null pointer match_pos_list to list_free, which dereferences it.["show details"] 
// CID 1603602: (#2 of 2): Dereference null return value (NULL_RETURNS)
// dereference: Dereferencing a pointer that might be NULL match_pos_list when calling list_free.
5202    list_free(match_pos_list);   
// CID 1603603: (#1 of 1): Explicit null dereferenced (FORWARD_NULL)
// var_deref_model: Passing null pointer match_score_list to list_free, which dereferences it
//  CID 1603602: (#2 of 2): Dereference null return value (NULL_RETURNS)
// dereference: Dereferencing a pointer that might be NULL match_pos_list when calling list_free.
5203    list_free(match_score_list);
//  CID 1603600: (#1 of 1): Explicit null dereferenced (FORWARD_NULL)
// var_deref_model: Passing null pointer retlist to list_free, which dereferences it.
5204    list_free(retlist);
5205    list_free(l);
5206    ga_clear(match_positions);
5207    return NULL;
5208#else
5209    return NULL;
5210#endif
5211}
5212
```

jclsn:
I am interested in this feature. Is this merged now or not? I am a bit confused that the PR is closed

habamax:
> I am interested in this feature. Is this merged now or not? I am a bit confused that the PR is closed

It is merged. However it has some issues discovered pointed out by Christian here.

jclsn:
Alright thanks, will take some time until it lands in the new release anyway, I guess.

PR #14698, opened 2024-05-01
cursor() and getregion() don't handle v:maxcol well

zeertzjq:
Problem:  cursor() and getregion() don't handle v:maxcol well.
Solution: Add special handling for v:maxcol like setpos() does.


chrisbra:
thanks!

PR #14699, opened 2024-05-02
Add file type detection for Inko

yorickpeterse:
This adds file type detection for [Inko](https://inko-lang.org/).

While there's also a Vim plugin that does this (https://github.com/inko-lang/inko.vim), it includes more logic (e.g. indent and syntax rules). Some of that (most notably the syntax) might change such that upstreaming and maintaining it in Vim itself is a bit painful, hence this PR only includes the file type detection logic.


chrisbra:
thanks!

PR #14703, opened 2024-05-02
danish localisation

lesorubshayan:


chrisbra:
ping @scootergrisen

scootergrisen:
The changes make no sense.

ronin49:
@chrisbra 
close this,https://github.com/vim/vim/pull/10703,https://github.com/vim/vim/issues/10161

chrisbra:
I am going to close this and the other PR from @ronin49.

You seem to be running this through some automatic machine translation, which I am afraid does not help to make translations better. I got a few complaints already about the bad quality.

I would rather prefer that the translations submitted make sense for someone speaking this language, so I don't think this helps much. It also takes away time and effort to review your PRs which we could spend on other things.

Also please don't just try to submit double PRs like #14710 . 

Thanks for understanding

k-takata:
@chrisbra You have already merged some of the PRs from @ronin49.
Are you going to revert them or keep them?

> You seem to be running this through some automatic machine translation, which I am afraid does not help to make translations better.

Agree. Recent machine translation is far better than before. But it is still not enough for the translation of the user interface.
Especially, the problem is that menus and messages don't have enough context. We often need to run the program and/or check the source code to understand the context.

k-takata:
Okay, let's keep them and wait feedback from native speakers.

Issue #14660, opened 2024-04-29
test fails


lesorubshayan:
### Steps to reproduce

make test

### Expected behaviour

test pass

### Version of Vim

latest

### Environment

arch

### Logs and stack traces

```shell
-------------------------------
Executed:  6154 Tests
 Skipped:    71 Tests
  FAILED:     1 Tests


Failures:
        From test_normal.vim:
        Found errors in Test_mouse_shape_after_cancelling_gr():
        command line..script /home/work/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_mouse_shape_after_cancelling_gr line 26: Expected ['beam', 'arrow'] but got ['arrow', 'arrow']

TEST FAILURE
make[2]: *** [Makefile:53: report] Error 1
make[2]: Leaving directory '/home/work/vim/src/testdir'
make[1]: *** [Makefile:2190: scripttests] Error 2
make[1]: Leaving directory '/home/work/vim/src'
make: *** [Makefile:39: test] Error 2
[work@archlinux vim]$
```

ronin49:
and question how to run specific test
tryed to run: make test <Test Name> but it failed

ronin49:
it is not possible to test cover memory allocation failures right?

zeertzjq:
It seems that #12157 will fix this.

ronin49:
is it possible to: make test -j4?

ronin49:
> It seems that #12157 will fix this.

yes

PR #14710, opened 2024-05-03
Fix mouseshape to be set correctly when using 'r' or 'gr' #12157

lesorubshayan:
https://github.com/vim/vim/pull/12157
https://github.com/vim/vim/issues/14660


chrisbra:
closing for the reasons mentioned here: https://github.com/vim/vim/pull/14703#issuecomment-2092337585

PR #14712, opened 2024-05-03
Support defcompile in test_override()

### LINKED ISSUE #14553: [vim9script] add `test_override('defcompile', 1)` option
I'm doing some refactoring on a vim9script project that uses several files. I'm thinking of adding `:defcompile` to all the files temporarily to maximize checking.

While looking at `import autoload` I came across `test_override('autoload', 1)`, which is good for testing script level references to possibly changed names/functions. But normal testing leaves some `:def` and `:class` unchecked.

If it's simple to do, with a `defcompile` option, the following
```
test_override('autoload', 1)
test_override('defcompile', 1)
```
would do a lot of checking, even for functions/methods that are only used in specialized circumstances, like error handling. This provides a manual way to get part of #13156.

yegappan:
Fixes #14553.

### DIFF
--- a/runtime/doc/testing.txt
+++ b/runtime/doc/testing.txt
@@ -369,6 +369,9 @@ test_override({name}, {val})				*test_override()*
 		autoload     `import autoload` will load the script right
 			     away, not postponed until an item is used
 		char_avail   disable the char_avail() function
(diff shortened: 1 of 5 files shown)


chrisbra:
thanks!

PR #14715, opened 2024-05-03
CI: Workaround for test_sound on macos-12

ichizok:
## Problem

CI: test_sound often fails on macos-12.

## Cause

Some of macos-12 runners (machines) have no audio device installed.

On my investigation; macos-12 runner runs on any one of at least 3 machines and 2 of them have no audio device installed.
Therefore, when tests run on such machine, test_sound fails.

## Solution

Install an audio device before testing.
 
This PR adds to the step of installing blackhole-2ch, which is macOS audio driver.


chrisbra:
thanks!

PR #14718, opened 2024-05-04
runtime(doc): correct getscriptinfo() example

zeertzjq:
When "sid" is specified, it returns a List with a single item.


chrisbra:
thanks!

PR #14724, opened 2024-05-07
runtime(cpp): Fix digit separator in octals and floats

adah1972:
and the incorrect rendering of floats that start with ".".


adah1972:
The original maintainer, vim-jp project, has been inactive for three years. But I have not changed the comment about the maintainer.

The current modes does not seem to require a dedicated maintainer, anyway....

adah1972:
This commit allows literals like `0'644` and `3.14159'26535'89793` to be correctly highlighted in Vim.

zeertzjq:
Related https://github.com/vim-jp/vim-cpp/pull/67

chrisbra:
thanks!

PR #14727, opened 2024-05-07
runtime(java): Stop handpicking syntax groups for @javaTop

zzzyxwvut:
Also:

- Remove the obsolete comment for `g:java_allow_cpp_keywords`.
- Remove the commented out groups `java\%[Debug\]StringError`.
- Infer and set the preferred formatting Vim options from  
  the modeline.
- Set `&encoding` and `&termencoding` to "utf-8" for test files.
- Limit non-ASCII charset to `[§ƒɐɘʬʭΑ-Τα-μ]` for test files.

Since `vim-6-0u`, non-contained syntax groups can be referred  
to by using the `contains=TOP...` argument.


runtime/syntax/java.vim
```diff
@@ -411,6 +385,10 @@ if exists("java_highlight_functions")
   syn match javaLambdaDef "\k\@4<!(\%(\k\|[[:space:]<>?\[\]@,.]\)*)\s*->"
 endif
 
+" The @javaTop cluster comprises non-contained Java syntax groups.
```
chrisbra, line 388: The @javaTop cluster was removed, so this comment needs to be updated.

zzzyxwvut: No, `@javaTop` is still defined.  Look below.

chrisbra: ah yes obviously. Sorry.

zzzyxwvut:
So, a syntax test for `input/java_methods_style.java` fails  
on Mac (Huge, 14).  Apparently for a multi-byte character `𝓔`  
(f0 9d 93 94).

zzzyxwvut:
Well, tests seem to pass now that `&enc` and `&tenc` are set  
to "utf-8".

zzzyxwvut:
I have locally run `make` with [`script`](https://man7.org/linux/man-pages/man1/script.1.html) in `xterm-256color`
(`$TERM`), with default `&tenc` and `&enc`, and with the options   
set to "utf-8".

```sh
cd runtime/syntax/
script --log-out /tmp/make.script --log-timing /tmp/make.timing --command 'make clean test'
scriptreplay --divisor 0.05 --log-out /tmp/make.script --log-timing /tmp/make.timing
```
(I wonder if it would be beneficial to instrument `make` with  
`script` for the CI so that its generated files and log files  
can be downloaded for post-mortem investigation for matching  
terminals.)

In either case, multi-byte characters are replaced on screen  
with what appears like `�` (ef bf bd) (Replacement Character)  
when the generated script file is replayed at a slower rate.
So, I'm not sure if `&enc` and `&tenc` matter here.

It would be helpful if someone with access to [Mac](https://github.com/vim/vim/blob/b4757e627e6c83d1c8e5535d4887a82d6a5efdd0/.github/workflows/ci.yml#L307) (12, 14)
and `make` have tried to locally reproduce [test failure](https://github.com/vim/vim/actions/runs/8988489000/job/24689448069#step:11:17029) and  
uploaded the generated `failed/` test files.

runtime/syntax/java.vim
```diff
@@ -341,8 +321,6 @@ if exists("java_highlight_debug")
   syn region  javaDebugStrTemplEmbExp	contained matchgroup=javaDebugStrTempl start="\\{" end="}" contains=javaComment,javaLineComment,javaDebug\%(Paren\)\@!.*
   syn region  javaDebugStrTempl		contained start=+\%(\.[[:space:]\n]*\)\@<="+ end=+"+ contains=javaDebugStrTemplEmbExp,javaDebugSpecial
   syn region  javaDebugStrTempl		contained start=+\%(\.[[:space:]\n]*\)\@<="""[ \t\x0c\r]*$+hs=e+1 end=+"""+he=s-1 contains=javaDebugStrTemplEmbExp,javaDebugSpecial,javaDebugTextBlockError
-  " The next line is commented out, it can cause a crash for a long line
```
dkearns, line 344: Is this still true?  I'm curious about the bug rather than the error highlighting feature.

chrisbra:
> uploaded the generated failed/ test files.

Yeah, that is currently hard to do. I wondered if we can just simply cat all files from the `failed/` directory, so one can at least copy/paste the content locally to be able to inspect it.

Somthing like we do for ASAN here: https://github.com/vim/vim/blob/b8ecedce79149ac6b994177e9a68979f86065cb1/.github/workflows/ci.yml#L299-L305

Something like this perhaps? 
```
bash -c "shopt -s globstar; tail -n +1  **/failed/*" 2>/dev/null || true
```

runtime/syntax/java.vim
zzzyxwvut: There was never a test suite for it so I wouldn't know.

Apparently, `javaStringError` became a burden [sometime](https://github.com/vim/vim-history/commit/c1525e06801a41cd7e1a0927b0cc0fa62ffa2682#diff-0fddd2a7f1b39c711eb9cf31da1cd6e9ca133ed96213fedf4ae46521f1796d37) around  
October 2000, and `javaDebugStringError` was [just recently](https://github.com/vim/vim/commit/100a665634a0a2edbc755daf453c4915800684dd)  
still its copy squirreled away.

Anyway, sooner or later any fledgeling Javaist would learn  
that:

> It is a compile-time error for a line terminator (§3.4) to  
> appear after the opening " and before the matching closing  
> ". 

and that no line splicing is supported for either `"a"` or  
`'a'`:

> The line continuation escape sequence can appear in a text  
> block, but cannot appear in a character literal or a string  
> literal because each disallows a LineTerminator. 

(JLS-21, [§3.10.5](https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.10.5), [§3.10.7](https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.10.7).)

Let's toss it out.

zzzyxwvut:
The failed test is not a regression.  I have found at least  
four other _earlier_ run failures for Mac (Huge, 14): [1](https://github.com/vim/vim/actions/runs/8824925004/job/24228298023#step:10:16370), [2](https://github.com/vim/vim/actions/runs/8840856619/job/24276935260#step:10:16347), [3](https://github.com/vim/vim/actions/runs/8854043458/job/24316210645#step:10:16362),  
[4](https://github.com/vim/vim/actions/runs/8856501136/job/24322848765#step:10:16354), since the introduction of `input/java_methods_style.java`.
(The links are anchored at the top of the file, scroll down  
a little to read the failure message.)

What is puzzling is that there are enough of successful test  
[runs](https://github.com/vim/vim/actions/runs/8827593571/job/24235935458#step:10:16353) in between the failures, and between the first failure  
and [the initial committing](https://github.com/vim/vim/actions/runs/8788552867) of the test file.

Maybe the fact that `𝓔` (Unicode category: Letter, Uppercase)  
does not qualify as a letter in Vim but a punctuation symbol
https://github.com/vim/vim/blob/d5c8c0920e1eee9ff7a9fa5168d8e85c01670630/src/mbyte.c#L2951

can be a clue.  (It is matchable with `\P` or `\F`, but not with  
`\k` or `\K`.)  Or is it a general aversion for characters made  
of three or more bytes (it is four: `f0 9d 93 94`)?

Should I replace such problematic characters in tests?

Should I drop the setting of `&enc` and `&tenc`?

chrisbra:
Yes, I would say keep it as simple as possible in the test, so `𝓔` should be replaced by something simpler. I think setting enc and tenc is fine, it's probably closer to what a user has set as well.

zzzyxwvut:
With `/[^[:alnum:][:blank:][:punct:]]` for aid, I have looked  
through all Java test files and replaced all characters made  
of more than two bytes with either ASCII characters or some  
two-byte characters (`[§ƒɐɘʬʭΑ-Τα-μ]`).

Let's see if these were the problem.

(By the way, testing on the linux console (`Ctrl-Alt-F1`, with  
`$TERM` saying `linux`) is also successful, with and without  
setting `&enc` and `&tenc` to "utf-8".)

chrisbra:
alright, thanks!

zzzyxwvut:
It is still [a problem](https://github.com/vim/vim/actions/runs/9038417238/job/24839482152#step:11:16980), now with `Τ` (GREEK CAPITAL LETTER TAU)  
of two bytes (ce a4) BUT for Mac (Huge, 12).

And there are two other candidates: `input/vim_keymap.vim` and  
`input/vim9_keymap.vim` with their `á` (c3 a1) and `“` (e2 80 9c).

Since it fails only so often, can it be time related (`sleep` 
or `TermWait()`)?  And without actual `failed/` files to look  
at, it is still a mystery who's at fault here: `readfile()` or  
`term_dumpwrite()`?

Another fun fact: when Vim is started with `LC_ALL=C` and it  
is needed to set `&enc` and `&tenc` to "utf-8" for the terminal  
buffer, it turns out that they must be set in a non-terminal  
buffer as well to have effect for the terminal buffer.  Huh?  
So, you would probably just set it once before `term_start()`  
is called.  (That should account for replacement characters
(`�`) filling in for other multi-byte characters that was  
mentioned [above](https://github.com/vim/vim/pull/14727#issuecomment-2099300251).)

zzzyxwvut:
@chrisbra, there appears to be a way to fetch `failed/*` files  
with the [upload-artifact](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) action.

> Storing artifacts uses storage space on GitHub.  GitHub
> Actions usage is free for standard GitHub-hosted runners
> in public repositories, and for self-hosted runners.

Just following [GitHub examples](https://github.com/actions/upload-artifact?tab=readme-ov-file#usage), we can give it a try with  
(please double check my YAML):

```yaml
jobs:

  . . . . . . .

  macos:
    steps:

      . . . . . . .

      - name: Test
        timeout-minutes: 20
        run: |
          make ${TEST}

      - name: Upload failed syntax tests
      - uses: actions/upload-artifact@v4
        with:
          # Name of the artifact to upload.
          # Optional. Default is 'artifact'
          # (Note that due to how artifacts are created in this v4
          # version, it is no longer possible to upload to the same
          # named artifact multiple times. You must either split the
          # uploads into multiple artifacts with different names, or
          # only upload once. Otherwise you will encounter an error.)
          name: ${{ matrix.runner }}-${{ matrix.features }}-${{ github.workflow_sha }}-${{ github.run_attempt }}-failed-syntax-tests

          # A file, directory or wildcard pattern that describes what
          # to upload.
          # Required.
          # (Note that within an individual job, there is a limit of
          # 500 artifacts that can be created for that job.)
          path: ${{ github.workspace }}/runtime/syntax/testdir/failed/

          # The desired behavior if no files are found using the
          # provided path.
          # Available Options:
          #   warn: Output a warning but do not fail the action
          #   error: Fail the action with an error message
          #   ignore: Do not output any warnings or errors, the action
          #     does not fail
          # Optional. Default is 'warn'
          if-no-files-found: ignore

          # Duration after which artifact will expire in days. 0 means
          # using default retention.
          # Minimum 1 day.
          # Maximum 90 days unless changed from the repository
          # settings page.
          # Optional. Defaults to repository settings.
          retention-days: 0

          # The level of compression for Zlib to be applied to the
          # artifact archive.
          # The value can range from 0 to 9.
          # For large files that are not easily compressed, a value of
          # 0 is recommended for significantly faster uploads.
          # Optional. Default is '6'
          compression-level: 9

          # If true, an artifact with a matching name will be deleted
          # before a new one is uploaded.
          # If false, the action will fail if an artifact for the
          # given name already exists.
          # Does not fail if the artifact does not exist.
          # Optional. Default is 'false'
          overwrite: true
```

Zip files should be available for [download](https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go):

> At the bottom of the workflow summary page, there is
> a dedicated section for artifacts.

and there are also [REST API endpoints](https://docs.github.com/en/rest/actions/artifacts) for Curl and friends.

chrisbra:
thanks, I created #14771 for that

PR #14728, opened 2024-05-08
Wrong display with 'smoothscroll' when emptying quickfix list

zeertzjq:
Problem:  Wrong display with 'smoothscroll' when emptying quickfix list.
Solution: Reset w_skipcol when w_topline is changed.


zeertzjq:
This is not the correct solution.

PR #14729, opened 2024-05-08
Check the imported variable type properly

yegappan:
When an imported dict or a class variable is set, check the dict item or the class variable type.


chrisbra:
thanks!

PR #14730, opened 2024-05-08
Wrong display with 'smoothscroll' when changing quickfix list

zeertzjq:
Problem:  Wrong display with 'smoothscroll' when changing quickfix list.
Solution: Reset w_skipcol when replacing quickfix list.


chrisbra:
thanks!

PR #14732, opened 2024-05-08
Remove some Win9x legacy from netrw built-in plugin

nir9:
Problem: The built-in file browsing plugin "netrw" contains a bunch of code that handles Windows 95/Win16 which are not supported by modern versions of Vim

Solution: Remove some legacy code/documentation from "netrw"

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -283,7 +283,7 @@ if !exists("g:netrw_scp_cmd")
  if executable("scp")
   call s:NetrwInit("g:netrw_scp_cmd" , "scp -q")
  elseif executable("pscp")
-  if (has("win32") || has("win95") || has("win64") || has("win16")) && filereadable('c:\private.ppk')
+  if (has("win32") || has("win64")) && filereadable('c:\private.ppk')
    call s:NetrwInit("g:netrw_scp_cmd", 'pscp -i c:\private.ppk')
   else
    call s:NetrwInit("g:netrw_scp_cmd", 'pscp -q')
@@ -296,7 +296,7 @@ endif
 call s:NetrwInit("g:netrw_sftp_cmd" , "sftp")
 call s:NetrwInit("g:netrw_ssh_cmd"  , "ssh")
 
-if (has("win32") || has("win95") || has("win64") || has("win16"))
+if (has("win32") || has("win64"))
   \ && exists("g:netrw_use_nt_rcp")
   \ && g:netrw_use_nt_rcp
   \ && executable( $SystemRoot .'/system32/rcp.exe')
@@ -311,7 +311,7 @@ endif
 " Default values for netrw's global variables {{{2
 " Cygwin Detection ------- {{{3
 if !exists("g:netrw_cygwin")
- if has("win32") || has("win95") || has("win64") || has("win16")
+ if has("win32") || has("win64")
   if  has("win32unix") && &shell =~ '\%(\<bash\>\|\<zsh\>\)\%(\.exe\)\=$'
    let g:netrw_cygwin= 1
   else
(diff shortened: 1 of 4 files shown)


runtime/autoload/netrw.vim
```diff
@@ -283,7 +283,7 @@ if !exists("g:netrw_scp_cmd")
  if executable("scp")
   call s:NetrwInit("g:netrw_scp_cmd" , "scp -q")
  elseif executable("pscp")
-  if (has("win32") || has("win95") || has("win64") || has("win16")) && filereadable('c:\private.ppk')
+  if (has("win32") || has("win64")) && filereadable('c:\private.ppk')
```
k-takata: Actually, this is still redundant.
`has("win32")` returns 1 on both 32-bit Windows and 64-bit Windows.
So, this can be:
```suggestion
  if has("win32") && filereadable('c:\private.ppk')
```

nir9:
One CI job fails since I did not rebuild the doc tags, I don't remember if that is necessary for PRs, if so let me know and I will fix

chrisbra:
no problem, I can fix it while merging it.

runtime/autoload/netrw.vim
```diff
@@ -311,7 +311,7 @@ endif
 " Default values for netrw's global variables {{{2
 " Cygwin Detection ------- {{{3
 if !exists("g:netrw_cygwin")
- if has("win32") || has("win95") || has("win64") || has("win16")
+ if has("win32") || has("win64")
   if  has("win32unix") && &shell =~ '\%(\<bash\>\|\<zsh\>\)\%(\.exe\)\=$'
    let g:netrw_cygwin= 1
   else
```
k-takata: (This comment is out of scope of this PR, but ...)
This Cygwin detection is completely broken.
`has("win32")` never returns 1 on Cygwin. So, `g:netrw_cygwin` will be always set to 0 (even on Cygwin).

k-takata:
Could you also update this part?
https://github.com/vim/vim/blob/fb745756db23c5500294ba3f8ef48ef18080f247/runtime/plugin/netrwPlugin.vim#L38

nir9:
@k-takata Fixed CR 👍

runtime/autoload/netrw.vim
```diff
@@ -283,7 +283,7 @@ if !exists("g:netrw_scp_cmd")
  if executable("scp")
   call s:NetrwInit("g:netrw_scp_cmd" , "scp -q")
  elseif executable("pscp")
-  if (has("win32") || has("win95") || has("win64") || has("win16")) && filereadable('c:\private.ppk')
+  if (has("win32")) && filereadable('c:\private.ppk')
```
k-takata: Could you change `(has("win32"))` to `has("win32")`?

nir9: fixed 👌

runtime/autoload/netrw.vim
```diff
@@ -12700,7 +12663,7 @@ endfun
 " ---------------------------------------------------------------------
 " s:ShellEscape: shellescape(), or special windows handling {{{2
 fun! s:ShellEscape(s, ...)
-  if (has('win32') || has('win64')) && $SHELL == '' && &shellslash
+  if (has('win32')) && $SHELL == '' && &shellslash
```
dkearns: Useless parens.

nir9:
@dkearns fixed thanks, missed that due to the single quotes :)

chrisbra:
Thanks. I'll merge it. However it's strange that is hardcodes  `'c:\private.ppk'` which doesn't seem to be documented and if you use pscp with a private key often, I would imagine you would typically use different keys for different servers.

nir9:
@chrisbra That's a good point, fixed that on another branch #14739 :)

PR #14733, opened 2024-05-08
Add runtime files and filetype detection for jj

gpanders:
[jj](https://github.com/martinvonz/jj) (Jujutsu) is a Git-compatible version control system.

Like git, `jj` allows the user to write a description of each change in a buffer. These are opened in the editor with a temporary file with a `.jjdescription` extension. Example:

```
Add runtime files and filetype detection for jj

JJ: This commit contains the following changes:
JJ:     M runtime/filetype.vim
JJ:     A runtime/ftplugin/jj.vim
JJ:     A runtime/syntax/jj.vim

JJ: Lines starting with "JJ: " (like this one) will be removed.
```

The format of the file is similar to a Git commit file. There is some plain text at the top of the file describing the change and some comments showing which files changed. The comment character is `JJ: ` (indicated by the text in the buffer).

The included syntax plugin highlights comments with the `Comment` highlight group, and highlights each of the individual changed files using the builtin `Added`, `Changed`, and `Removed` highlight groups.

The include filetype plugin is copied almost verbatim from the `gitcommit` filetype plugin, though I dropped the Git commands and adapted for the updated comment strings.

### DIFF
--- a/.github/MAINTAINERS
+++ b/.github/MAINTAINERS
@@ -161,6 +161,7 @@ runtime/ftplugin/j.vim			@glts
 runtime/ftplugin/java.vim		@zzzyxwvut
 runtime/ftplugin/javascript.vim		@dkearns
 runtime/ftplugin/javascriptreact.vim	@dkearns
+runtime/ftplugin/jj.vim			@gpanders
 runtime/ftplugin/json.vim		@dbarnett
 runtime/ftplugin/json5.vim		@dkearns
 runtime/ftplugin/jsonc.vim		@izhakjakov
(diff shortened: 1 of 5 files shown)


chrisbra:
Thanks, can you please add your email or github handle at the maintainer line? And perhaps update https://github.com/vim/vim/blob/master/.github/CODEOWNERS ?

clason:
Just a small comment: Users in `CODEOWNERS` actually have to be members of the org; otherwise this will not work properly and Github will complain that "This file has errors".

Of course, you can still use that file for your own tracking purposes, but I wonder if a different name and format (`MAINTAINERS.{md,txt}`) wouldn't be more useful then (and shut Github up)?

gpanders:
> Can you please add a filetype detection test?

Done, thanks for the reminder.

>Thanks, can you please add your email or github handle at the maintainer line

Done as well. I didn't update CODEOWNERS tho, per @clason's comment it sounds like that may not work as intended? But if you prefer I can update it regardless.

chrisbra:
Yeah, I noticed that warning. It's not particular useful that one cannot add non-org people there. I have just renamed it to MAINTAINERs (in 3ac83c7141dd60ae45c0346b7c0f4539587391aa) Not sure if it is actually useful (or it's just redundant).

chrisbra:
now you can add it to the MAINTAINERS file 😬

PR #14736, opened 2024-05-09
For an unknown type, display only the type name in the error message

### LINKED ISSUE #13153: Badly formatted error message when casting to unknown type
### Steps to reproduce

```vim
vim9script
var a = <T>b
```

The following error message is output: `E1010: Type not recognized: T>b`

### Expected behaviour

`E1010: Type not recognized: T`

### Version of Vim

9.0.1927

### Environment

Ubuntu 20.04

### Logs and stack traces

_No response_

yegappan:
Fixes #13153.


chrisbra:
thanks!

PR #14737, opened 2024-05-09
zsh modules' files are not recognized

Freed-Wu:
Problem: zsh modules' files are not recognized
Solution: add some patterns for zsh modules' files

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -406,14 +406,18 @@ export def FTidl()
   setf idl
 enddef
 
-# Distinguish between "default", Prolog and Cproto prototype file.
+# Distinguish between "default", Prolog, zsh module's C and Cproto prototype file.
 export def ProtoCheck(default: string)
+  # zsh modules use '#include "*.pro"'
+  # https://github.com/zsh-users/zsh/blob/63f086d167960a27ecdbcb762179e2c2bf8a29f5/Src/Modules/example.c#L31
+  if getline(1) =~ '/* Generated automatically */'
+    setf c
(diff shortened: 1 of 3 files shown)


runtime/autoload/dist/ft.vim
```diff
@@ -408,12 +408,16 @@ enddef
 
 # Distinguish between "default", Prolog and Cproto prototype file.
```
k-takata: This line should be also updated.

src/testdir/test_filetype.vim
```diff
@@ -855,7 +855,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     zsh: ['.zprofile', '/etc/zprofile', '.zfbfmarks', 'file.zsh', 'file.zsh-theme', 'file.zunit',
           '.zcompdump', '.zlogin', '.zlogout', '.zshenv', '.zshrc', '.zsh_history',
           '.zcompdump-file', '.zlog', '.zlog-file', '.zsh', '.zsh-file',
-          'any/etc/zprofile', 'zlog', 'zlog-file', 'zsh', 'zsh-file'],
+          'any/etc/zprofile', 'zlog', 'zlog-file', 'zsh', 'zsh-file', 'file.mmi'],
```
k-takata: Mistake of 'file.mdd'?

Freed-Wu: Fixed.

runtime/autoload/dist/ft.vim
Freed-Wu: Fixed.

chrisbra:
thanks!

PR #14738, opened 2024-05-09
xxd: buffer-overflow with specific flags

Ordoviz:
This is a continuation of #14003.

xxd writes each output line into a global buffer before printing. The maximum size of that buffer was not calculated correctly.

This command was crashing in AddressSanitizer:

    xxd -Ralways -g1 -c256 -d -o 9223372036854775808 /etc/passwd

This prints a line of 6680 bytes but the buffer only had room for 6549 bytes. If the output from "-b" was colored, the line could be even longer.

### DIFF
--- a/src/testdir/test_xxd.vim
+++ b/src/testdir/test_xxd.vim
@@ -416,9 +416,10 @@ endfunc
 func Test_xxd_buffer_overflow()
   new
   let input = repeat('A', 256)
-  let expected = ' -9223372036854775808: ' . repeat("\e[1;32m41\e[0m ", 256) . ' ' . repeat("\e[1;32mA\e[0m", 256)
-  exe 'r! printf ' . input . '| ' . s:xxd_cmd . ' -Ralways -g1 -c256 -d -o 9223372036854775808'
-  call assert_equal(expected, join(getline(1, '$')))
+  call writefile(['-9223372036854775808: ' . repeat("\e[1;32m41\e[0m ", 256) . ' ' . repeat("\e[1;32mA\e[0m", 256)], 'Xxdexpected', 'D')
+  exe 'r! printf ' . input . '| ' . s:xxd_cmd . ' -Ralways -g1 -c256 -d -o 9223372036854775808 > Xxdout'
+  call assert_equalfile('Xxdexpected', 'Xxdout')
+  call delete('Xxdout')
   bwipe!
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks, can you please add a test for this? See https://github.com/vim/vim/blob/master/src/testdir/test_xxd.vim

src/testdir/test_xxd.vim
```diff
@@ -411,6 +411,17 @@ func Test_xxd_max_cols()
   endfor
 endfunc
 
+
+" Try to trigger a buffer overflow (#14738)
+func Test_xxd_buffer_overflow()
+  new
+  let input = repeat('A', 256)
+  let expected = ' -9223372036854775808: ' . repeat("\e[1;32m41\e[0m ", 256) . ' ' . repeat("\e[1;32mA\e[0m", 256)
```
Ordoviz: I don't know why the expected result starts with a space (it does not seem to be printed by xxd). I also think the offset shouldn't be negative but it is what it is.

src/xxd/xxd.c
```diff
@@ -205,29 +205,16 @@ char osver[] = "";
...
-	     1 +                                     /* new line */    \
-	     1)                                      /* zero */
+#define LLEN \
+    39             /* addr: ⌈log10(ULONG_MAX)⌉ if "-d" flag given. We assume ULONG_MAX = 2**128 */ \
+    + 2            /* ": " */ \
+    + 13 * COLS    /* hex dump with colors */ \
+    + (COLS - 1)   /* whitespace between groups if "-g1" option given and "-c" maxed out */ \
+    + 2            /* whitespace */ \
+    + 12 * COLS    /* ASCII dump with colors */ \
+    + 2            /* "\n\0" */
```
k-takata:
```suggestion
    (39            /* addr: ⌈log10(ULONG_MAX)⌉ if "-d" flag given. We assume ULONG_MAX = 2**128 */ \
    + 2            /* ": " */ \
    + 13 * COLS    /* hex dump with colors */ \
    + (COLS - 1)   /* whitespace between groups if "-g1" option given and "-c" maxed out */ \
    + 2            /* whitespace */ \
    + 12 * COLS    /* ASCII dump with colors */ \
    + 2)           /* "\n\0" */
```
It is better to surround with parentheses.

PR #14739, opened 2024-05-09
Remove hardcoded private.ppk logic from netrw

nir9:
Problem: File browsing plugin "netrw" has weird logic assuming "C:\private.ppk" exists on the computer, and if so, sets that as the connection key

Solution: Remove this logic from the plugin, no reason to assume this file exists on the drive and handle this case

Sanity tested :nwrite scp://... command after this change with pscp and seems to work fine


chrisbra:
thanks!

lilydjwg:
Hi, is it intended to prefer `pscp` over `scp`? It breaks for me because I have a regular `scp` command but a `pscp` for "parallel scp", not the dropbear one.

chrisbra:
should now be fixed with 43f2edc09

PR #14741, opened 2024-05-10
filetype: add URLs for some zsh modules files, recognize mdd as sh

Freed-Wu:
Problem:  mdd files is a sh script not a zsh script
Solution: add URLs for some zsh modules files, recognize mdd as sh

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -3094,12 +3094,14 @@ au BufNewFile,BufRead .zsh*,.zlog*,.zcompdump*  call s:StarSetf('zsh')
 au BufNewFile,BufRead zsh*,zlog*		call s:StarSetf('zsh')
 
 " Zsh module
+" mdd: https://github.com/zsh-users/zsh/blob/57248b88830ce56adc243a40c7773fb3825cab34/Etc/zsh-development-guide#L285-L288
+" mdh, pro: https://github.com/zsh-users/zsh/blob/57248b88830ce56adc243a40c7773fb3825cab34/Etc/zsh-development-guide#L268-L271
 " *.mdd will generate *.mdh, *.pro and *.epro.
 " module's *.c will #include *.mdh containing module dependency information and
 " *.pro containing all static declarations of *.c
 " *.epro contains all external declarations of *.c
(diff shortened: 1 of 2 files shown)


Shane-XB-Qian:
just a friendly reminder: not sure how to determine the ext to some ft, but perhaps it was not so clear if you were not the files creator or usual users, so be careful, e.g https://github.com/vim/vim/pull/14532 and specially e.g last time https://github.com/vim/vim/pull/14456 :
`au BufNewFile,BufRead *.{pgf,nlo,nls,out,thm,eps_tex,pygtex,pygstyle,clo,aux,brf,ind,lof,loe,nav,vrb,ins,tikz,bbx,cbx,beamer}	setf tex`
i helped to correct '.out', but here those ext were you sure really it was for tex? e.g 'pygstyle' it naming like for 'python'.

Shane-XB-Qian:
and this one seem you were going to correct https://github.com/vim/vim/pull/14737 , i am not a zsh user, but if '.mdd' not a freq used file which just a temp made up file then perhaps ok to no ft, otherwise i were glad you correct it.

Freed-Wu:
Thanks for your suggestion. I have added the document links for them:

- mdd: https://github.com/zsh-users/zsh/blob/57248b88830ce56adc243a40c7773fb3825cab34/Etc/zsh-development-guide#L285-L288
- mdh, pro: https://github.com/zsh-users/zsh/blob/57248b88830ce56adc243a40c7773fb3825cab34/Etc/zsh-development-guide#L268-L271

chrisbra:
thanks!

chrisbra:
> au BufNewFile,BufRead *.mdd			setf sh

Is this correct? It's unclear to me, whether this will be sourced by a bourne compatible shell, or `zsh` itself.

Freed-Wu:
>Is this correct?

In the document:

> Each module is described by a file with a name ending in `.mdd' somewhere
> under the Src directory.  This file is actually a shell script that will
> sourced when zsh is built. To describe the module it can/should set the
> following shell variables:

`*.mdd` is used in autotools, just source it then generate `*.pro`, `*.epro`, `*.mdh`.

PR #14742, opened 2024-05-10
Recursively delete directory by default on netrw

nir9:
Problem: Currently the netrw delete command (triggered by pressing D on the folder) cannot delete directories which are not empty.

Solution: Delete directories recursively by default, there is a confirmation anyway, so the chance for mistake is slim

Sanity tested deletion on Windows 10 and Debian WSL and works


chrisbra:
thanks!

PR #14744, opened 2024-05-10
runtime(java): Strive to remain compatible for at least Vim 7.0

zzzyxwvut:
Also:

- Limit all look-behind regexp patterns.
- Cache regexp capabilities for `[:upper:]` and `[:lower:]`.


chrisbra:
Wow, that is some tricky stuff. Thanks!

PR #14747, opened 2024-05-11
Divide by zero with getmousepos() and 'smoothscroll'

zeertzjq:
Problem:  Divide by zero with getmousepos() and 'smoothscroll'.
Solution: Don't compute skip_lines when width1 is zero.


chrisbra:
thanks!

PR #14748, opened 2024-05-11
runtime(doc): fix invalid link to Open Publication License

zeertzjq:
The link in https://opencontent.org/openpub is actually invalid too:
```
Copyright (c) <year> by <author’s name or designee>.
This material may be distributed only subject to the terms and conditions set forth in the Open Publication License, vX.Y or later (the latest version is presently available at https://www.opencontent.org/openpub/).
```

But since the text here doesn't exactly match that anyway (doesn't have the parentheses), it's probably fine to fix it.


chrisbra:
thanks, I include this together with some more fixes of invalid links.

PR #14749, opened 2024-05-11
fix: make 'rm -rf foo/' be a bit harder

Shane-XB-Qian:
// if not back out https://github.com/vim/vim/pull/14742 
// then make it be a bit harder to make mistake to rm all 
//
// also seems recover its original logic of 'all' argument

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -11454,24 +11454,29 @@ fun! s:NetrwLocalRmFile(path,fname,all)
 
   else
    " attempt to remove directory
+   let rmfile= substitute(rmfile,'[\/]$','','e')
    if !all
     echohl Statement
     call inputsave()
-    let ok= input("Confirm *recursive* deletion of directory<".rmfile."> ","[{y(es)},n(o),a(ll),q(uit)] ")
+    " make it be a bit harder to make mistake by inputting upper case
+    let ok= input("Confirm *recursive* deletion of directory<".rmfile."> ","[{Y(ES)},n(o),A(LL),q(uit)] ")
     call inputrestore()
-    let ok= substitute(ok,'\[{y(es)},n(o),a(ll),q(uit)]\s*','','e')
+    let ok= substitute(ok,'\[{Y(ES)},n(o),A(LL),q(uit)]\s*','','e')
     if ok == ""
      let ok="no"
     endif
-    if ok =~# 'a\%[ll]'
+    if ok =~# 'a\%[ll]'->toupper()
      let all= 1
     endif
-   endif
-   let rmfile= substitute(rmfile,'[\/]$','','e')
-
-   if all || ok =~# 'y\%[es]' || ok == ""
-    if delete(rmfile,"rf")
-     call netrw#ErrorMsg(s:ERROR,"unable to delete directory <".rmfile.">!",103)
+    if all || ok =~# 'y\%[es]'->toupper()
+     if delete(rmfile,"rf")
+      call netrw#ErrorMsg(s:ERROR,"failed to *recursive* delete directory <".rmfile.">!",103)
+     endif
+    endif
+    let ok = ok->tolower()
+   else
+    if delete(rmfile,"d")
+     call netrw#ErrorMsg(s:ERROR,"unable to delete directory <".rmfile."> -- is it empty?",103)
     endif
    endif
   endif


runtime/autoload/netrw.vim
```diff
@@ -11454,24 +11454,29 @@ fun! s:NetrwLocalRmFile(path,fname,all)
 
   else
    " attempt to remove directory
+   let rmfile= substitute(rmfile,'[\/]$','','e')
```
chrisbra, line 11457: I would leave the directory indicator. That is actually useful.

runtime/autoload/netrw.vim
```diff
@@ -11454,24 +11454,29 @@ fun! s:NetrwLocalRmFile(path,fname,all)
...
    if !all
     echohl Statement
     call inputsave()
-    let ok= input("Confirm *recursive* deletion of directory<".rmfile."> ","[{y(es)},n(o),a(ll),q(uit)] ")
+    " make it be a bit harder to make mistake by inputting upper case
+    let ok= input("Confirm *recursive* deletion of directory<".rmfile."> ","[{Y(ES)},n(o),A(LL),q(uit)] ")
```
chrisbra, line 11462: Now you are unexpectedly changing the key-bindings. This will break peoples habits. I don't think this is appreciated by long-time users. (besides, it actually looks funny, that some keys need to be upper-cased and others not).

Shane-XB-Qian: well, i am not against just back out the https://github.com/vim/vim/pull/14742, then kept the appreciated long-time users' habits, or you had better idea not to so easy or high possibility to make mistake to rm all?

chrisbra: What mistake, there is still a confirmation dialog

runtime/autoload/netrw.vim
Shane-XB-Qian: it was there before this PR and https://github.com/vim/vim/pull/14742 
i guess it was remained there for judging if it was a dir or not, i just kept it.

chrisbra: No, this was not touched by #14742

Shane-XB-Qian: yes, so i didnot plan to touch it either.
// please notice that if/else

runtime/autoload/netrw.vim
Shane-XB-Qian: 1. lower case is easy to mis-type, like a 'hit-enter' happened.
2. it was designed to: if 'all=1' then no 'confirm', else req a confirm, and confirm to only delete('rm', 'd') not 'rf'.

chrisbra:
Okay, let's step back a bit for now. What problem are you trying to solve?

I usually don't use netrw, but I just tried: `D` on a directory. There is always a confirmation dialog, `Confirm *recursive* deletion of directory</tmp/c/> [{y(es)},n(o),a(ll),q(uit)]`, I need to confirm using `y` or `a` to delete. I think this is fine. Just pressing Enter, does not delete anything.

Conversely, when I go back to commit d3952e8cfe7baed1ff5c8111da5 (so before #14742) was merged, it errored out for a non-empty directory, even if I explicitly confirmed to delete. Which I do not find particular helpful.

Shane-XB-Qian:
there 2 rm operation in netrw actually, one is 'local', another is 'remote', there some cfg about 'rm' for 'remote', but 'local' was designed to only delete('rmfile', 'd') not 'rf', 
it knew it maybe a bit inconvenience, but please Keep In Mind:
lots or most official vim users would use vim (vs xxxvim) in production, (a.k.a 'what vim is'), such 'rm all' operation is dangerous which like last time the performance bug https://github.com/vim/vim/issues/14650 , which would make 'real or classic vim users' afraid to use vim. (or use 'nano'? :sweet_smile:)

clason:
Here you go again with the gatekeeping...

Shane-XB-Qian:
> Conversely, when I go back to commit [d3952e8](https://github.com/vim/vim/commit/d3952e8cfe7baed1ff5c8111da5272a038cb55e4) (so before #14742) was merged, it errored out for a non-empty directory, even if I explicitly confirmed to delete. Which I do not find particular helpful.

i think it meant to make 'rm all' be harder, so if you'd like to keep #14742 , then wish you accept this enhance (and fix) too.

chrisbra:
> there 2 rm operation in netrw actually, one is 'local', another is 'remote', there some cfg about 'rm' for 'remote', but 'local' was designed to only delete('rmfile', 'd') not 'rf',

Correct. But #14742 only changed the `NetrwLocalRmFile` function, which as the name indicated, is only used for local browsing. I tested with using `:e scp://localhost//tmp/` and then trying to delete anything. #14742 did not change anything here. In fact, I was not able to delete anything on the remote side (recursively). So this has nothing to do with local vs. remote. 

Sorry, I think the new behaviour makes a bit more sense to me.

Shane-XB-Qian:
> Here you go again with the gatekeeping...

funny boy

Shane-XB-Qian:
> > there 2 rm operation in netrw actually, one is 'local', another is 'remote', there some cfg about 'rm' for 'remote', but 'local' was designed to only delete('rmfile', 'd') not 'rf',
> 
> Correct. But #14742 only changed the `NetrwLocalRmFile` function, which as the name indicated, is only used for local browsing. I tested with using `:e scp://localhost//tmp/` and then trying to delete anything. #14742 did not change anything here. In fact, I was not able to delete anything on the remote side (recursively). So this has nothing to do with local vs. remote.

default for 'remote' is 'rmdir' which would not rm non-empty dir.

> #14742 did not change anything here.

it changed:
1.    lower case is easy to mis-type, like a 'hit-enter' happened.
 2.   it was designed to: if 'all=1' then no 'confirm', else req a confirm, and confirm to only delete('rmfile', 'd') not 'rf'.

chrisbra:
> lower case is easy to mis-type, like a 'hit-enter' happened.

This was not changed. It was like that before already.

> it was designed to: if 'all=1' then no 'confirm', else req a confirm, and confirm to only delete('rmfile', 'd') not 'rf'.

This seems to happen when either:
1) moving from [local directory to remote directory](https://github.com/vim/vim/blob/58448e09be497a8abb595ae309b6edfbc8e0e05a/runtime/autoload/netrw.vim#L7855-L7863). (in which case it makes sense I would say) 
2) When [deleteing several marked files/directories](https://github.com/vim/vim/blob/58448e09be497a8abb595ae309b6edfbc8e0e05a/runtime/autoload/netrw.vim#L11362-L11395) in which asking for confirmation several times is probably redundant and the changed behaviour also makes sense I would think.

Shane-XB-Qian:
> > lower case is easy to mis-type, like a 'hit-enter' happened.
> 
> This was not changed. It was like that before already.

i meant after https://github.com/vim/vim/pull/14742 when you mis-typed 'D' then lower case is easier to make mistake which compared to upper case you had to press 'shift' first, a.k.a 'SAFE'.

> > it was designed to: if 'all=1' then no 'confirm', else req a confirm, and confirm to only delete('rmfile', 'd') not 'rf'.
> 
> This seems to happen when either:
> 
>     1. moving from [local directory to remote directory](https://github.com/vim/vim/blob/58448e09be497a8abb595ae309b6edfbc8e0e05a/runtime/autoload/netrw.vim#L7855-L7863). (in which case it makes sense I would say)
> 
>     2. When [deleteing several marked files/directories](https://github.com/vim/vim/blob/58448e09be497a8abb595ae309b6edfbc8e0e05a/runtime/autoload/netrw.vim#L11362-L11395) in which asking for confirmation several times is probably redundant and the changed behaviour also makes sense I would think.

yes, seems those 2 places for now, but please note:
```vim
" s:NetrwLocalRmFile: remove file fname given the path {{{2
"                     Give confirmation prompt unless all==1
fun! s:NetrwLocalRmFile(path,fname,all)
```
when all == 1, it was designed to 'no confirm', in the future some places may call this fun too, supposed it is Not ok to rm all directly.

runtime/autoload/netrw.vim
```diff
@@ -11454,24 +11454,29 @@ fun! s:NetrwLocalRmFile(path,fname,all)
...
      let ok="no"
     endif
-    if ok =~# 'a\%[ll]'
+    if ok =~# 'a\%[ll]'->toupper()
      let all= 1
     endif
-   endif
-   let rmfile= substitute(rmfile,'[\/]$','','e')
-
-   if all || ok =~# 'y\%[es]' || ok == ""
```
Shane-XB-Qian, line 11472: also please note: i removed this `ok == ""` though not sure when it would happen (or mostly it would not happen), but it scared me if i just pressed 'enter' somehow.

chrisbra:
> i meant after https://github.com/vim/vim/pull/14742 when you mis-typed 'D' then lower case is easier to make mistake which compared to upper case you had to press 'shift' first, a.k.a 'SAFE'.

You already have to type capital `D`, which makes deleting already a conscious action (I would have expected `d`, like in `ncdu`, but this doesn't delete). So that is not an issue then. 

> in the future some places may call this fun too, supposed it is Not ok to rm all directly.

May be, or may be not. We don't have a netrw maintainer and I don't expect this to change soon. I think this needs to be taken care of, when new functions start calling it. But that is not something to worry about currently.

Okay, let's close it then. I think we have verified that the current new version is fine and there is no issue. Sorry, I am not convinced that this change is a good one.

Shane-XB-Qian:
> > i meant after #14742 when you mis-typed 'D' then lower case is easier to make mistake which compared to upper case you had to press 'shift' first, a.k.a 'SAFE'.
> 
> You already have to type capital `D`, which makes deleting already a conscious action (I would have expected `d`, like in `ncdu`, but this doesn't delete). So that is not an issue then.

ncdu is a gui-like diaglog for deletion, is it same?
which compared to vim is a cli input confirmation in the bottom, same like inputting any others keys, which somehow mis-typed in case.

> > in the future some places may call this fun too, supposed it is Not ok to rm all directly.
> 
> May be, or may be not. We don't have a netrw maintainer and I don't expect this to change soon. 

um, i supposed we should not wait a knew problem to happen, there perhaps __no regret to recover files once rm all__.
if you were not sure, i supposed you still can try to ask question/advice from @cecamp 
or if you were not sure, then why merged  #14742 so easy, ahahahah...

> I think this needs to be taken care of, when new functions start calling it. But that is not something to worry about currently.
> 
> Okay, let's close it then. I think we have verified that the current new version is fine and there is no issue. Sorry, I am not convinced that this change is a good one.

anyway, if you were not like this PR, i can maintain this local hack to myself, no problem.

chrisbra:
> ncdu is a gui-like diaglog for deletion, is it same?

It's a tui application. Like Vim is a tui application.

> we should not wait a knew problem to happen, 

Let me repeat: There is currently not a known problem.

Shane-XB-Qian:
> > ncdu is a gui-like diaglog for deletion, is it same?
> 
> It's a tui application. Like Vim is a tui application.

it is gui-_like_ __diaglog__, not a cli inputing yes/no/all/quit in vim, so to ncdu, users knew what he is doing, and mostly impossible to feedkey.

> > we should not wait a knew problem to happen,
> 
> Let me repeat: There is currently not a known problem.

i think i had repeated many times:
1. inputting by lower case to `rm -rf` is high possibility making mistake.
2. "                     Give confirmation prompt unless all==1
fun! s:NetrwLocalRmFile(path,fname,all)
when all == 1, it was designed to 'no confirm', in the future some places may call this fun too, supposed it is Not ok to rm all directly.

benknoble:
There is [some research that making deletes harder is not always the right idea](https://ux.stackexchange.com/a/74412). I appreciate that the confirmation makes it harder for my cat to accidentally delete files, but extra roadblocks (given the requirement to shift-d already) are probably not necessary.

Clarifying the interface of a public function (even one not intended for public use) may be worthwhile.

chrisbra:
> it is gui-like diaglog, not a cli inputing yes/no/all/quit in vim, so to ncdu, users knew what he is doing, and mostly impossible to feedkey.

I'd still call it TUI and it requires to confirm :) 
 

>    1.  inputting by lower case to rm -rf is high possibility making mistake.
>     2. " Give confirmation prompt unless all==1
>     fun! s:NetrwLocalRmFile(path,fname,all)
>     when all == 1, it was designed to 'no confirm', in the future some places may call this fun too, supposed it is Not ok to rm all directly.

I repeat one more time:  
1. That is not a problem per se and just changing key-bindings for this will probably annoy users.
2. You just [confirmed, that this is okay currently](https://github.com/vim/vim/pull/14749#issuecomment-2105687353). So no problem currently.

Shane-XB-Qian:
> > it is gui-like diaglog, not a cli inputing yes/no/all/quit in vim, so to ncdu, users knew what he is doing, and mostly impossible to feedkey.
> 
> I'd still call it TUI and it requires to confirm :)

gui-like, __-like__, or ok, if you like, it is popup window in tui.. um.. the point was not the like, vs/but how it interactive.

> > 1. inputting by lower case to rm -rf is high possibility making mistake.
> >    2. " Give confirmation prompt unless all==1
> >    fun! s:NetrwLocalRmFile(path,fname,all)
> >    when all == 1, it was designed to 'no confirm', in the future some places may call this fun too, supposed it is Not ok to rm all directly.
> 
> I repeat one more time:
> 
>     1. That is not a problem per se and just changing key-bindings for this will probably annoy users.

it was not having to change but the https://github.com/vim/vim/pull/14742 break the game (yes, this is a repeat too).
or changing it designed then required some way to save it, that change (to 'rf') also was an annoying or scare to users.
 
>     2. You just [confirmed, that this is okay currently](https://github.com/vim/vim/pull/14749#issuecomment-2105687353). So no problem currently.

i cofirmed there now seems 2 functions called `s:NetrwLocalRmFile`, but `knew` it was dangerous if there would be more and  if `all == 1`.

chrisbra:
let's just agree to disagree here and let's not waste more time here. :) 

Thanks all.

Shane-XB-Qian:
ok, fine, to pass-by vim users, if you scared the mis-typed 'D' `rm -rf` all your files, then try to patch this PR by yourself.

PR #14751, opened 2024-05-11
ftplugin(python): E16 fix, async keyword support for define

tpict:
Two changes:
- From @gdupras: Fix "E16: Invalid range" when using a count with jump to start/end of class/method
- From @craigmac: Update `define` with optional `async` keyword

I also updated my preferred contact email.

Thanks!

### DIFF
--- a/runtime/ftplugin/python.vim
+++ b/runtime/ftplugin/python.vim
@@ -1,10 +1,9 @@
 " Vim filetype plugin file
 " Language:	python
-" Maintainer:	Tom Picton <tom@tompicton.co.uk>
+" Maintainer:	Tom Picton <tom@tompicton.com>
 " Previous Maintainer: James Sully <sullyj3@gmail.com>
 " Previous Maintainer: Johannes Zellner <johannes@zellner.org>
-" Last Change:	Mon, 5 October 2020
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

runtime/ftplugin/python.vim
```diff
@@ -15,7 +14,7 @@ set cpo&vim
 setlocal cinkeys-=0#
 setlocal indentkeys-=0#
 setlocal include=^\\s*\\(from\\\|import\\)
-setlocal define=^\\s*\\(def\\\|class\\)
+setlocal define=^\\s*\\([async ]\\?def\\\|class\\)
```
zeertzjq, line 17: This line errors:
```
E518: Unknown option: ]\\?def\\|class\\)
```
And shouldn't it use `\(` and `\)` instead?
```vim
setlocal define=^\\s*\\(\\(async\\s\\+\\)\\?def\\\|class\\)
```
(And there may be multiple spaces between `async` and `def`/`class`.)

zeertzjq: Created a PR at https://github.com/tpict/vim-ftplugin-python/pull/22

Issue #14752, opened 2024-05-11
Broken syntax highlighting (swayconfig)


diegoviola:
### Steps to reproduce

Just open `/etc/sway/config` in vim and notice that some commands (e.g. `bindsym` ones) have broken syntax highlighting.

Here's a screenshot of what it currently looks like:

![20240511_15h35m47s_grim](https://github.com/vim/vim/assets/4371/8bc647b7-346b-403a-8a3e-48d988d35aca)

### Expected behaviour

Syntax highlighting is properly applied to all commands.

### Version of Vim

9.1.405

### Environment

Operating system: Arch Linux
Terminal: foot
Value of $TERM: foot
Shell: /usr/bin/bash

diegoviola:
I did some research on this and looks like commit https://github.com/vim/vim/commit/dd83b63eb5b7884c1c882eb984778492f6951ed7 replaced the following line:

    syn region i3ConfigBlockOrphan start=/^\s\+\S/ skip=/^\s\|^$/ end=/^}\?/ contains=swayConfigBindBlockCombo,i3ConfigCriteria,i3ConfigAction,i3ConfigSeparator,i3ConfigActionKeyword,i3ConfigOption,i3ConfigString,i3ConfigNumber,i3ConfigVariable,i3ConfigBoolean,i3ConfigComment,i3ConfigParen keepend extend

With this:

    syn region swayConfigBlockOrphan start=/^\s\+\(--[a-z-]\+ \)*[$a-zA-Z0-9_+]\+ [a-z[]/ skip=/\\$\|$\n^\s*}$/ end=/$/ contains=i3ConfigBindArgument,i3ConfigBindCombo,i3ConfigParen keepend extend

When the lines are reverted and changed to `swayConfigBlockOrphan` the problem goes away, so it must be some regex thing?

Cc @JosefLitos @chrisbra.

chrisbra:
ping @jamespeapen

jamespeapen:
Thanks for the report - I can replicate the issue and will open a PR with a fix. Its not just the `bindsym` that's affected - any keywords with leading spaces are considered errors if they're not surrounded by `{}` in a multiline block.

diegoviola:
@jamespeapen amazing, thanks a lot!

JosefLitos:
Interesting, I would've thought that `skipwhite` would skip whitespace. Apparently it doesn't work like that with match like `/^/`.

I will make a different PR, since the changes required to solve this for all commands would be quite different.

[...]

PR #14753, opened 2024-05-11
runtime(sway + i3): Allow leading spaces for sway keywords

jamespeapen:
Solution for #14752
It was not just `bindsym` that was affected - any keywords or keyword matches with leading spaces were considered errors if they were not surrounded by `{}` in a multiline block.

The only solution I could find was to revert the introduction of the `i3/swayConfigBlockOrphan` group. However, since that `swayConfigBlockOrphan` is defined multiple times it may be better to remove them and rewrite with less repetition if necessary. I haven't seen the use case where that is necessary in #14544 so @JosefLitos could you show where that's needed?

Taking out the `swayConfigBlockOrphan` does not fix the original issue. I think part of the error is that may be because the [`i3ConfigTopLevelDirective`](https://github.com/vim/vim/blob/86f6e2c2eed7df2bf5c60cc74d08d7a8d3c75f45/runtime/syntax/i3config.vim#L263) was not allowing spaces. I thought `skipwhite` would allow that, but apparently it doesn't.

Allowing leading spaces for keywords fixes the original issue, but causes the same issue for commented lines with leading spaces. This necessitated a fix for the `i3Comment` keyword, removing leading spaces from its definition as they are captured by the `i3ConfigTopLevelDirective`.

@hiqua for i3 review

### DIFF
--- a/runtime/syntax/i3config.vim
+++ b/runtime/syntax/i3config.vim
@@ -46,7 +46,7 @@ syn match i3ConfigIncludeCommand /`[^`]*`/ contained contains=@i3ConfigSh
 syn region i3ConfigParamLine matchgroup=i3ConfigKeyword start=/include / end=/$/ contained contains=@i3ConfigStrVar,i3ConfigIncludeCommand,i3ConfigShOper keepend
 
 " 4.2 Comments
-syn match i3ConfigComment /^\s*#.*$/ contained contains=i3ConfigTodo
+syn match i3ConfigComment /#.*$/ contained contains=i3ConfigTodo
 
 " 4.3 Fonts
 syn match i3ConfigFontSize / \d\+\(px\)\?$/ contained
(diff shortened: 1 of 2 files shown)


JosefLitos:
The issue with the Orphan fixes is that normal commands are considered a keybind name. I narrowed down the match even further to only modifiers or special keys (starting with capital), or simple chars (one letter, usually used in custom `mode`s).

jamespeapen:
Before #14757 is merged would you mind showing an example of the three orphan blocks in action? 
Do you mean the first line is a block?

hiqua:
I'll wait for you folks to agree on something before I start reviewing anything, if that's fine for you!

JosefLitos:
> ... would you mind showing an example of the three orphan blocks in action?

obrazek

> Do you mean the first line is a block?

I am not sure I'm using the right terms but as an example take the first picture of the `bindsym` orphans, this is the initial block part, that vim needs to recognize the entire block:
obrazek
As you can see, at this point the orphan highlighting isn't in use, because the `bindsym` block is correctly recognized. But without the keyword (the line being outside the displayed text when scrolling from the bottom up), it cannot match the block, therefore the Orphan match is used to guess and highlight the lines that look like the contents of a `bindsym` block.

---

It works the same way with `bindgesture` and `bindswitch`, though probably most people won't have such long blocks of bindings for these two, making them unnecessary, but just for the sake of completeness and uniformity I added Orphan matches for them too.

I know this is kind of hacky, but I couldn't think of a better way. But I would like to also try some other way at some point - at least to see and compare.

jamespeapen:
I see - not orphans in the file itself, but orphan based on the current buffer view. Thanks for explaining and for the fix!
Closing in favor of #14757.

Issue #14487, opened 2024-04-10
9.1.0296 completely breaks syntax highlighting of &ft=mail


chdiza:
### Steps to reproduce

Build 9.1.0296 from source.  Open a file with &ft=mail that has some header lines in it.  E.g., start composing a mail in `mutt`, or open a blank file, set the &ft, and write:
```
From: someaddress
To: someotheraddress
```

Observe that nothing in the buffer is highlighted except for email addresses and email-quoted text.  This is wrong; on 9.1.0295 and earlier, the header names are highlighted.

If I put the cursor on the "From:" and do this:
```
echo synIDattr(synID(line("."), col("."), 1), "name")
```
I get nothing at all in 9.1.0296, but it shows `mailHeaderKey` in 9.1.0295 and earlier.

Depending on which email message I'm looking at, I can also see that highlighting of quoted-printable blocks that are being "> "-quoting is also broken.

### Expected behaviour

I expect mail highlighting to not be such as to highlight nearly no text at all.

### Version of Vim

9.1.0296

### Environment

Linut Mint 21.x
Terminal: Kitty

Shane-XB-Qian:
um.. interesting, is it related to locale or encoding something? i replied some mail today, why i did not face same problem, :sweat_smile:

chdiza:
I have limited time to hunt down details today, but the problem at least partly disappears if I set `&re=2`.

I have also identified that with &re=1, the `:g` command is broken.  E.g., this type of thing `doesn't` work anymore even though it did one patchlevel earlier:
```
:g/\cSIZ/d
```

No unicode characters there, or in the file in which I try to make that deletion.

PR #14756, opened 2024-05-12
regexp: wrong match with 'ic' and comparing multi-byte with single byte char

### LINKED ISSUE #14294: Unicode character ſ is matched as itself and as 's.'

chrisbra:
This is v2 of v9.1.296, but still draft to see if this still breaks.

When the regexp engine compares two utf-8 codepoints case insensitively it may match an adjacent character, because it assumes it can step over as many bytes as the pattern contains.

This however is not necessarily true because of case-folding, a multi-byte UTF-8 character can be considered equal to some single-byte value.

Let's consider the pattern 'ſ' and the string 's'. When comparing and ignoring case, the single character 's' matches, and since it matches Vim will try to step over the match (by the amount of bytes of the pattern), assuming that since it matches, the length of both strings is the same.

However in that case, it should only step over the single byte value 's' by 1 byte and try to start matching after it again. So for the backtracking engine we need to ensure:
- we try to match the correct length for the pattern and the text
- in case of a match, we step over it correctly

There is one tricky thing for the backtracing engine. 
We also need to calculate correctly the number of bytes to compare the 2 different utf-8 strings s1 and s2. So we will count the number of characters in s1 that the byte len specified. Then we count the number of bytes to step over the same number of characters in string s2 and then we can correctly compare the 2 utf-8 strings. 

A similar thing can happen for the NFA engine, when skipping to the next character to test for a match. We are skipping over the regstart pointer, however we do not consider the case that because of case-folding we may need to adjust the number of bytes to skip over. So this needs to be adjusted in find_match_text() as well.

A related issue turned out, when prog->match_text is actually empty. In that case we should try to find the next match and skip this condition.

Note: this breaks Mail and CSS Syntax highlighting and CI on FreeBSD/MacOS https://github.com/vim/vim/issues/14487 and
https://groups.google.com/d/msgid/vim_dev/CAJkCKXtui%3DDTWx9eV8Dbs19XoFL9b63ObSNXWCRvLsEZCB6yfw%40mail.gmail.com.

fixes: #14294

### DIFF
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -3801,6 +3801,15 @@ utf_strnicmp(
  * two characters otherwise.
  */
     int
+mb_strnicmp2(char_u *s1, char_u *s2, size_t n1, size_t n2)
+{
+    if (n1 == n2 || !enc_utf8)
+	return mb_strnicmp(s1, s2, n1);
+    else
+	return utf_strnicmp(s1, s2, n1, n2);
+}
+
+    int
 mb_strnicmp(char_u *s1, char_u *s2, size_t nn)
 {
     int		i, l;
--- a/src/proto/mbyte.pro
+++ b/src/proto/mbyte.pro
@@ -48,6 +48,7 @@ int utf_islower(int a);
 int utf_tolower(int a);
 int utf_isupper(int a);
 int mb_strnicmp(char_u *s1, char_u *s2, size_t nn);
+int mb_strnicmp2(char_u *s1, char_u *s2, size_t n1, size_t n2);
 void show_utf8(void);
 int latin_head_off(char_u *base, char_u *p);
 int dbcs_screen_head_off(char_u *base, char_u *p);
--- a/src/regexp.c
+++ b/src/regexp.c
@@ -1812,7 +1812,7 @@ cstrncmp(char_u *s1, char_u *s2, int *n)
 cstrchr(char_u *s, int c)
 {
     char_u	*p;
-    int		cc;
+    int		cc, lc;
 
     if (!rex.reg_ic || (!enc_utf8 && mb_char2len(c) > 1))
 	return vim_strchr(s, c);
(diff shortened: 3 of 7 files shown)


chrisbra:
can someone test this please? It seems to work for me and does no longer have the previous issue.

zzzyxwvut:
> There is one tricky thing for the backtracing engine.

It still seems to be inconsistent.

For example,

```
Two Miſsiſsippis (Mississippis).
Διαίρεσις (diaeresis).
```

All Greek small sigmas are found with `/\%#=1\c[\u017f\u03c3]`  
or `/\%#=1\c[ſσ]`, but only Latin long s's.

According to this [Wiki entry](https://en.wikipedia.org/wiki/Long_s), both long and short s's are  
small (lowercase) letters, and, I think, should be likewise  
discoverable similar to Greek small sigmas `[σς]`.

chrisbra:
Hm, the backtracking engine doesn't handle case folding when handling collections/bracket expressions. `[...]`. Hm, that function `cstrchr()` is strange, it makes `cc` sometimes lower case and sometimes uppercase and then uses that to compare against the utf-folded version of the search string... 
https://github.com/vim/vim/blob/e4b991ed36f96dd01c6d75e46a04fd1a99180e58/src/regexp.c#L1786-L1832

I wonder if we should change it like this instead:
```c
      static char_u *
  cstrchr(char_u *s, int c)
  {
      char_u      *p;
      int         cc;

      if (!rex.reg_ic || (!enc_utf8 && mb_char2len(c) > 1))
          return vim_strchr(s, c);

      // tolower() and toupper() can be slow, comparing twice should be a lot
      // faster (esp. when using MS Visual C++!).
      // For UTF-8 need to use folded case.
      if (enc_utf8 && c > 0x80)
          cc = utf_fold(c);
      else
      {
           if (MB_ISUPPER(c))
              cc = MB_TOLOWER(c);
          else if (MB_ISLOWER(c))
              cc = c;
          else
              return vim_strchr(s, c);
      }

      if (has_mbyte)
      {
          for (p = s; *p != NUL; p += (*mb_ptr2len)(p))
          {
              int uc = utf_ptr2char(p);
              if (enc_utf8 && (c > 0x80 || uc > 0x80))
              {
                  // Do not match an illegal byte.  E.g. 0xff matches 0xc3 0xbf,
                  // not 0xff.
                  if ((uc < 0x80 || uc != *p) && utf_fold(uc) == cc)
                      return p;
              }
[...]
```

That seems to fix your failure.

PR #14763, opened 2024-05-13
ftplugin(python): Fix bad define regex

tpict:
Thanks to @zeertzjq


chrisbra:
thanks, I merge it now.

FTR: In the future you may want to consider using `:let` syntax, see [:h :let-&](https://vimhelp.org/eval.txt.html#%3Alet-%26) so you don't need to escape the `\` anymore:
```vim
let &l:include='^\s*\(from\|import\)'
let &l:define='^\s*\(\(async\s\+\)\?def\|class\)' 
```

PR #14771, opened 2024-05-14
CI: Upload failed screendump tests

chrisbra:
It's a bit of a pain to debug failing screendump tests without knowing exactly what went wrong. Therefore include actions/upload-artifact for the Github CI runners and have them uploaded those failing screen dump tests.

Let's add this step to each of the Linux/MacOS/Windows workflows..

Example:
https://github.com/chrisbra/vim/actions/runs/9085493619

### DIFF
--- a/.github/actions/screendump/action.yml
+++ b/.github/actions/screendump/action.yml
@@ -1,0 +1,26 @@
+name: 'screendump'
+description: "Upload failed syntax tests"
+runs:
+  using: "composite"
+  steps:
+    - name: Upload failed syntax tests
+      uses: actions/upload-artifact@v4
+      with:
+        # Name of the artifact to upload.
+        name: ${{ github.workflow }}-${{ github.job }}-${{ join(matrix.*, '-') }}-failed-syntax-tests
+
+        # A file, directory or wildcard pattern that describes what
+        # to upload.
+        path: |
+         ${{ github.workspace }}/runtime/syntax/testdir/failed/*
+         ${{ github.workspace }}/src/testdir/failed/*
+        # The desired behavior if no files are found using the
+        # provided path.
+        if-no-files-found: ignore
+
+        # Duration after which artifact will expire in days. 0 means
+        # using repository settings.
+        retention-days: 0
+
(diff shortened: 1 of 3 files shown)


dundargoc (commented):
You could use a local action to deduplicate the code if you wanna

zzzyxwvut:
A few suggestions:

- Since all kinds of failed tests will be collected, their  
  archive suffixes better be generic, e.g. `-failed-tests`;  
  also these descriptions: `- name: Upload failed tests`.

- `matrix.runner` is only defined for the `macos` jobs; but  
  e.g. `${{ matrix.toolchain }}-${{ matrix.features }}-` and  
  `linux-${{ matrix.compiler }}-${{ matrix.features }}-` can  
  be used elsewhere.

chrisbra:
> You could use a local action to deduplicate the code if you wanna

Thanks, even so it may not work because of the issue mentioned by @zzzyxwvut how would one do that?

chrisbra:
Ah thank you that is clever

zzzyxwvut:
Let me [contextualise it](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes):

> Uploading to the same named Artifact multiple times.
>
> Due to how Artifacts are created in this new version, it is no longer possible
> to upload to the same named Artifact multiple times. You must either split the
> uploads into multiple Artifacts with different names, or only upload once.
> Otherwise you will encounter an error.

dundargoc:
They are talking about same name in the same workflow, not the same name in the entire repository: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

zzzyxwvut:
All right, it seems that a dedicated namespace is allotted  
for each workflow  and `github.workflow_sha` would indeed be  
superfluous (unless you want provenance written in the name  
of the archive).  Is there a need for `github.run_attempt`to  
allow for manual `Re-run jobs` functionality?

Also, are there file name length limits to consider for any  
popular file systems?

dundargoc:
I don't know. I will leave that as an exercise for the reader.

PR #14778, opened 2024-05-15
Update vi_diff.txt: added default value for flash

dertuxmalwieder:
The "flash" option was added relatively late and seems to be exclusive to System V. (It's not in 4.4BSD and it's not in V8 UNIX, checked [on TUHS](https://www.tuhs.org/cgi-bin/utree.pl).) The oldest occurrence of `flash` in "a vi" I could find is in SysV R2 for the VAX, where it [defaults to 1](https://github.com/ryanwoodsmall/oldsysv/blob/master/sysvr2-vax/src/cmd/vi/vax/ex_data.c) = on. So, added.


chrisbra:
thanks!

PR #14780, opened 2024-05-15
Add ]r and [r motions to go to rare words.

### LINKED ISSUE #14773: Add `]r` / `[r` to go to the next / previous `SpellRare`
Currently there is `]s` and `[s` which go to the next and previous `SpellBad` or `SpellRare` in the buffer. This works well if you have a comprehensive dictionary that defines nearly all words, in which case `]s` and `[s` might just hop to instances of `SpellRare`. But what if your dictionary has missing words? Then you'd have to press `]s` repeatedly through 1+ `SpellBad` just to get to the `SpellRare` that you meant to jump to.

It'd be nice if `SpellRare` got its own mapping. Since it's currently open, `]r` and `[r` could do the trick.

Writing a function that computes the next jump location and defining the mapping manually. In truth that's what my plugin currently does but a native solution would be faster and less prone to errors.

**Additional context**
Like `[s` and `]s`, any new mappings should take into account `'wrapscan'`.

cvwillegen:
Add motion commands to go to the next/previous 'rare' word (as indicated by the spelling engine).

Closes #14773


src/testdir/Make_all.mak
```diff
@@ -278,6 +278,7 @@ NEW_TESTS = \
 	test_spell \
 	test_spell_utf8 \
 	test_spellfile \
+	test_spellrare \
```
chrisbra, line 281: You also need to add it below to NEW_TESTS_RES target

src/testdir/test_spellrare.vim
```diff
@@ -0,0 +1,73 @@
+" Test spell checking
+" Note: this file uses latin1 encoding, but is used with utf-8 encoding.
+
+source check.vim
+CheckFeature spell
+
+source screendump.vim
```
chrisbra: I think you are not using screendumps.

chrisbra:
Nice. And that was fast!

src/testdir/test_spellrare.vim
cvwillegen: Fixed

src/testdir/Make_all.mak
cvwillegen: Fixed

runtime/doc/spell.txt
```diff
@@ -62,6 +62,17 @@ To search for the next misspelled word:
 							*[S*
 [S			Like "]S" but search backwards.
 
+							*]r*
+]r			Move to next 'rare' word after the cursor.
```
zeertzjq:
```suggestion
]r			Move to next "rare" word after the cursor.
```

runtime/doc/spell.txt
```diff
@@ -62,6 +62,17 @@ To search for the next misspelled word:
...
+			A count before the command can be used to repeat.
+			'wrapscan' applies.
+
+							*[r*
+[r			Like "]r" but search backwards, find the 'rare'
```
zeertzjq:
```suggestion
[r			Like "]r" but search backwards, find the "rare"
```

src/testdir/test_spellrare.vim
```diff
@@ -0,0 +1,71 @@
+" Test spell checking
+" Note: this file uses latin1 encoding, but is used with utf-8 encoding.
```
zeertzjq: This file is ASCII-only, and is recognized as utf-8 by Vim, so this comment doesn't really make sense.

src/spell.c
```diff
@@ -1337,6 +1337,7 @@ spell_move_to(
     win_T	*wp,
     int		dir,		// FORWARD or BACKWARD
     int		allwords,	// TRUE for "[s"/"]s", FALSE for "[S"/"]S"
+    int		rare,		// TRUE for "[r"/"]r", FALSE otherwise
```
zeertzjq: Maybe you can combine these two arguments into a single enum?

runtime/doc/spell.txt
cvwillegen: Fixed.

runtime/doc/spell.txt
cvwillegen: Fixed.

src/testdir/test_spellrare.vim
cvwillegen: That'll teach me to copy existing test files :-/

src/testdir/test_spellrare.vim
```diff
@@ -0,0 +1,65 @@
...
+
+  " Create a small word list to test that spellbadword('...')
+  " can return ['...', 'rare'].
+  e Xwords
+  insert
+foo
+foobar/?
+foobara/?
+.
+   w!
```
yegappan: Instead of this, can you use the following?
```
   let lines =<< trim END
     foo
     foobar/?
     foobara/?
   END
   call writefile(lines, 'Xwords', 'D')
```

src/testdir/test_spellrare.vim
```diff
@@ -0,0 +1,65 @@
...
+  normal [r
+  call assert_equal('foobar', expand('<cword>'))
+  normal 2[r
+  call assert_equal('foobar', expand('<cword>'))
+
+  bwipe!
+  set nospell
+
+  call delete('Xwords.spl')
+  call delete('Xwords')
```
yegappan: If you use writefile() above with the 'D' flag, then the file will be automatically deleted at the end of this function.

src/testdir/test_spellrare.vim
```diff
@@ -0,0 +1,65 @@
+" Test spell checking
+
+source check.vim
+CheckFeature spell
+
+func TearDown()
+  set nospell
```
yegappan: At the end of the Test_spellrareword() function, the 'spell' option is reset.  So this is not needed in TearDown().

yegappan (changes requested):

ColinKennedy:
When I made the GitHub issue, I was expecting it to be addressed deep into the future. And I woke to find an already working PR in < 24 hours. Thank you @cvwillegen for taking this on! Vim's community is the best!

chrisbra:
Thanks, I think this looks good

chrisbra:
thanks all, good work!

PR #14781, opened 2024-05-15
Problem: find_module has been removed in Python 3.13

### LINKED ISSUE #14776: `find_module` is deleted in python3.13
### Steps to reproduce

https://github.com/vim/vim/blob/master/src/if_py_both.h#L7328-L7333

since python3.13, `find_module` is deleted https://github.com/python/cpython/pull/98059

maybe we should just delete L7328-L7333

### Expected behaviour

work with python 3.13

### Version of Vim

9.1

### Environment

Python 3.13.0b1

### Logs and stack traces

```shell
`:python3 import vim; vim.find_module`


PathFinder has no attribute 'find_module'
```
```

### PR DESCRIPTION
**Pull request by chrisbra**
Problem:  find_module has been removed in Python 3.13
Solution: Do not include find_module for Python >= 3.13

fixes: #14776

### DIFF
--- a/src/if_py_both.h
+++ b/src/if_py_both.h
@@ -7325,12 +7325,11 @@ populate_module(PyObject *m)
 	return -1;
     }
 
+# if PY_VERSION_HEX < 0x30c00a7
+    // find_module has been removed as of Python 3.12.0a7
     if ((py_find_module = PyObject_GetAttrString(cls, "find_module")))
-    {
-	// find_module() is deprecated, this may stop working in some later
-	// version.
 	ADD_OBJECT(m, "_find_module", py_find_module);
-    }
+# endif
 
     Py_DECREF(imp);
 
--- a/runtime/doc/if_pyth.txt
+++ b/runtime/doc/if_pyth.txt
@@ -1,4 +1,4 @@
-*if_pyth.txt*   For Vim version 9.1.  Last change: 2023 Oct 25
+*if_pyth.txt*   For Vim version 9.1.  Last change: 2024 May 16
 
 
 		  VIM REFERENCE MANUAL    by Paul Moore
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by puremourning** [comment]
Need to update the docs? Or do you update runtime files periodically like Bram did?

**Comment by chrisbra** [comment]
oh, good point. I didn't think it would the find_module is there, but it is actually mentioned. I'll just add a note, that `find_module` has been removed starting with Python3.13, but won't try to change the implementation mentioned at `:h python-special-path`.

Now that I think about it, I hope module loading still works even when `find_module` is no longer available. If not, we need a more involved patch, but I don't know the python interface very well.

File: src/if_py_both.h [hunk 1]
```diff
@@ -7325,12 +7325,15 @@ populate_module(PyObject *m)
 	return -1;
     }
 
+# if PY_VERSION_HEX < 0x30d00f0
+    // find_module has been removed as of Python 3.13
     if ((py_find_module = PyObject_GetAttrString(cls, "find_module")))
     {
 	// find_module() is deprecated, this may stop working in some later
 	// version.
```
k-takata: This should be removed (or updated).

src/if_py_both.h
```diff
@@ -7325,12 +7325,15 @@ populate_module(PyObject *m)
 	return -1;
     }
 
+# if PY_VERSION_HEX < 0x30d00f0
```
k-takata: 0x30d00f0 means 3.13.0 final, but the reported version is beta1.
Should be 0x30d00b0 (or 0x30d00a0 or 0x30d0000)?

chrisbra: Actually, it seems to have been removed already in 3.12: https://docs.python.org/3/whatsnew/3.12.html

chrisbra: It seems to have been merged as of this commit: https://github.com/python/cpython/blob/326997829d02458246dfd5b6d03297e2418bde52/Include/patchlevel.h

chrisbra: I changed it to 3.12.0a7+

chrisbra:
Regarding documentation, I am actually not sure. There is `vim.find_module()` but this seems this is internal to the vim module and has nothing todo with `imp.find_module`. However it looks like `vim.find_spec()` was never documented, so I mentioned it for now. 

Let me know if this looks good now.

src/if_py_both.h
chrisbra: done

PR #14783, opened 2024-05-16
Unable to leave long line with 'smoothscroll' and 'scrolloff'

### LINKED ISSUE #14726: smoothscroll down never leaves the line when scrolloff non zero
### Steps to reproduce

On a long line, continually doing `gj` never leaves the line. When you get close to the end of the long line, the first chars of the long line appear at the top of screen. Attached a file, but I'd guess any old file with a long line works.
```
gvim -u NONE -U NONE <attached_file>
set smoothscroll
set lines=22
set display=truncate
set scrolloff=3
map j gj
```

**NOTE:** change to `set scrolloff=0` and it works OK.

Additionally, being at the top line of the file, doing the first `gj` end up on the third display line. There is no way to easily place the cursor on the first display lines.

[long-line.txt](https://github.com/vim/vim/files/15237909/long-line.txt)

### Expected behaviour

Going down one line goes down one display line.

### Version of Vim

9.1.393

### Environment

ubuntu/gtk

### Logs and stack traces

_No response_

luukvbaal:
Problem:  Unable to leave long line with 'smoothscroll' and 'scrolloff'.
          Corrupted screen near the end of a long line with 'scrolloff'.
Solution: Only correct cursor in case scroll_cursor_bot() was not itself
          called to make the cursor visible. Avoid adjusting for
          'scrolloff' beyond the text line height.

Fix #14726


chrisbra:
Thanks. I move some of the variable declaration to the beginning of the function so (even so it currently compiles fine (here)), but so far we have been using C90 features only I believe.

luukvbaal:
>  I move some of the variable declaration to the beginning of the function so (even so it currently compiles fine (here)), but so far we have been using C90 features only I believe.

Hmm I seem to remember some discussion around lifting that restriction, and though it was lifted.

errael:
Oops, left out this important paragraph
```
Please don't make changes everywhere to use the C99 features, it causes merge
problems for existing patches.  Only use them for new and changed code.
```

chrisbra:
ah thanks. Then it shall be okay in the future. Didn't find this reference, apologize.

errael:
> Didn't find this reference

One of several things buried in `:help develop`. I haven't read it in detail and wonder how up to date it might be.

chrisbra:
I shall remember ;) (or someone will remind me of this thread) :)

PR #14770, opened 2024-05-14
two additional remarks in documentation

Konfekt:
These two points did not seem immediately clear to everyone

PR #14784, opened 2024-05-16
runtime(doc): Remove trailing spaces

k-takata:
Fix the issues introduced by #14770.


chrisbra:
oh I didn't notice. Thanks for catching it!

PR #14787, opened 2024-05-16
Adding register viminfo partial write warning

ElectricPulse:
The following often happens to me:
I cut ("c" command) 100+ lines from file1, close file1, paste into file2, realize I had lost 50+ lines of text, which can't be recovered from file1 because it was cut out and the file closed.
I realize that this problem has many solutions (use one vim session, `:set viminfo`, use system clipboard &c) yet this still happens to me, especially on machines where I forgot to `:set viminfo` in vimrc and where there is no system clipboard.
For the sake of noobs I decided to atleast give a warning when such a thing is about to happen.

This feature request tries to address this in the following ways:
1. Add a warning when exiting with register size bigger than viminfo
2. Increasing viminfo register max line count from default 50 lines to 500 lines

Note 1: Regarding the warning, it doesn't prevent the user from exiting, it just displays the messages and after pressing Enter, exits. This is because even though I changed it so that do_viminfo now returns a FAIL or OK status, write_viminfo doesn't return such an error because I have no idea of how about implementing the more intrusive feature of an getout failing, ie. ex_exit failing.

Note 2: What error number should be applied to the new warning (I set it to 2000)?

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -2296,8 +2296,6 @@ EXTERN char e_not_possible_to_change_sign_str[]
 #ifdef FEAT_VIMINFO
 EXTERN char e_cant_rename_viminfo_file_to_str[]
 	INIT(= N_("E886: Can't rename viminfo file to %s!"));
-EXTERN char e_warning_registers_partially_written_to_viminfo[]
-	INIT(= N_("E2000: Warning: registers only partially written to viminfo, increase viminfo size"));
 #endif
 #if defined(FEAT_PYTHON) || defined(FEAT_PYTHON3)
 EXTERN char e_sorry_this_command_is_disabled_python_side_module_could_not_be_loaded[]
--- a/src/errors.h
+++ b/src/errors.h
@@ -2296,6 +2296,8 @@ EXTERN char e_not_possible_to_change_sign_str[]
 #ifdef FEAT_VIMINFO
 EXTERN char e_cant_rename_viminfo_file_to_str[]
 	INIT(= N_("E886: Can't rename viminfo file to %s!"));
+EXTERN char e_warning_registers_partially_written_to_viminfo[]
+	INIT(= N_("E2000: Warning: registers only partially written to viminfo, increase viminfo size"));
 #endif
 #if defined(FEAT_PYTHON) || defined(FEAT_PYTHON3)
 EXTERN char e_sorry_this_command_is_disabled_python_side_module_could_not_be_loaded[]
--- a/src/optiondefs.h
+++ b/src/optiondefs.h
@@ -2764,7 +2764,7 @@ static struct vimoption options[] =
 #ifdef FEAT_VIMINFO
 			    (char_u *)&p_viminfo, PV_NONE, did_set_viminfo, NULL,
 #if defined(MSWIN)
-			    {(char_u *)"", (char_u *)"'100,<50,s10,h,rA:,rB:"}
+			    {(char_u *)"", (char_u *)"'100,<500,s100,h,rA:,rB:"}
 #elif defined(AMIGA)
 			    {(char_u *)"",
 				 (char_u *)"'100,<50,s10,h,rdf0:,rdf1:,rdf2:"}
--- a/src/optiondefs.h
+++ b/src/optiondefs.h
@@ -2769,7 +2769,7 @@ static struct vimoption options[] =
 			    {(char_u *)"",
 				 (char_u *)"'100,<50,s10,h,rdf0:,rdf1:,rdf2:"}
 #else
-			    {(char_u *)"", (char_u *)"'100,<50,s10,h"}
+			    {(char_u *)"", (char_u *)"'100,<500,s100,h"}
 #endif
 #else
 			    (char_u *)NULL, PV_NONE, NULL, NULL,
--- a/src/testdir/test_viminfo.vim
+++ b/src/testdir/test_viminfo.vim
@@ -315,6 +315,13 @@ func Test_cmdline_history_order()
   call delete('Xviminfo')
 endfunc
 
+func Force_write_viminfo()
+  redir => res
+  silent! wviminfo Xviminfo
+  redir END
+  call assert_match('W23:', res)
+endfunc
+
 func Test_viminfo_registers()
   call test_settime(8)
   call setreg('a', "eight", 'c')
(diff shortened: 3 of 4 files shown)


src/viminfo.c
```diff
@@ -1877,8 +1879,10 @@ write_viminfo_registers(FILE *fp)
 	    len = 0;
 	    for (j = 0; j < num_lines; j++)
 		len += (long)STRLEN(y_ptr->y_array[j]) + 1L;
-	    if (len > (long)max_kbyte * 1024L)
+	    if (len > (long)max_kbyte * 1024L) {
```
k-takata:
```suggestion
	    if (len > (long)max_kbyte * 1024L)
	    {
```

src/viminfo.c
```diff
@@ -1906,8 +1910,10 @@ write_viminfo_registers(FILE *fp)
 	fprintf(fp, "\t%s\t%d\n", type, (int)y_ptr->y_width);
 
 	// If max_num_lines < 0, then we save ALL the lines in the register
-	if (max_num_lines > 0 && num_lines > max_num_lines)
+	if (max_num_lines > 0 && num_lines > max_num_lines) {
```
k-takata:
```suggestion
	if (max_num_lines > 0 && num_lines > max_num_lines)
	{
```

src/viminfo.c
```diff
@@ -2949,7 +2958,12 @@ do_viminfo(FILE *fp_in, FILE *fp_out, int flags)
 	write_viminfo_search_pattern(fp_out);
 	write_viminfo_sub_string(fp_out);
 	write_viminfo_history(fp_out, merge);
-	write_viminfo_registers(fp_out);
+
+	if(write_viminfo_registers(fp_out) == FAIL) {
```
k-takata:
```suggestion
	if (write_viminfo_registers(fp_out) == FAIL)
	{
```

src/viminfo.c
```diff
@@ -2949,7 +2958,12 @@ do_viminfo(FILE *fp_in, FILE *fp_out, int flags)
...
+		emsg(_(e_warning_registers_partially_written_to_viminfo));
```
yegappan: Can you add tests for these changes to the testdir/test_viminfo.vim file?

src/errors.h
```diff
@@ -2296,6 +2296,8 @@ EXTERN char e_not_possible_to_change_sign_str[]
 #ifdef FEAT_VIMINFO
 EXTERN char e_cant_rename_viminfo_file_to_str[]
 	INIT(= N_("E886: Can't rename viminfo file to %s!"));
+EXTERN char e_warning_registers_partially_written_to_viminfo[]
+	INIT(= N_("E2000: Warning: registers only partially written to viminfo, increase viminfo size"));
```
RestorerZ: If I understood your description correctly, then this is not an error, but a warning.
Like this:
https://github.com/vim/vim/blob/68f8cfb7efa79510a01dd55d35cb5c7456a7261f/src/buffer.c#L2278
Accordingly, the number should not start with the letter "E", but with the letter "W". The last warning number was "W22".
> Note 2: What error number should be applied to the new warning (I set it to 2000)?
> 

>

RestorerZ:
It might be a good idea to add a description of this warning.
For example, here: https://vimhelp.org/options.txt.html#%27viminfo%27

src/errors.h
ElectricPulse: Oh okay and why are there Warnings under src/errors.h?

ElectricPulse:
Fixed the tests.

chrisbra:
> Note 1: Regarding the warning, it doesn't prevent the user from exiting, it just displays the messages and after pressing Enter, exits. This is because even though I changed it so that do_viminfo now returns a FAIL or OK status, write_viminfo doesn't return such an error because I have no idea of how about implementing the more intrusive feature of an getout failing, ie. ex_exit failing.

I checked this and this got me wondering. I see, that there is already some error handling, that simply shows an error message but does still quite Vim after the user presses Enter. So I am wondering how useful this is? It's certainly too late for the user to do anything against this and fix the warning. But on the other side, with the newly increased defaults, perhaps this is something not to worry about it for now?

src/optiondefs.h
```diff
@@ -2769,7 +2769,7 @@ static struct vimoption options[] =
 			    {(char_u *)"",
 				 (char_u *)"'100,<50,s10,h,rdf0:,rdf1:,rdf2:"}
 #else
-			    {(char_u *)"", (char_u *)"'100,<50,s10,h"}
+			    {(char_u *)"", (char_u *)"'100,<500,s100,h"}
```
chrisbra, line 2772: If we change the default, I think we should also increase it for at least Windows (see above).

src/viminfo.c
```diff
@@ -1827,6 +1827,8 @@ write_viminfo_registers(FILE *fp)
     long	len;
     yankreg_T	*y_ptr;
     yankreg_T	*y_regs_p = get_y_regs();;
+    // Signifies partial write
+    int		part_write = OK;
```
chrisbra:
```suggestion
    int		retval = OK;
```

src/testdir/test_viminfo.vim
```diff
@@ -315,6 +315,13 @@ func Test_cmdline_history_order()
   call delete('Xviminfo')
 endfunc
 
+func Force_write_viminfo()
+  redir => res
+  silent! wviminfo Xviminfo
+  redir END
+  call assert_match('W23:', res)
```
chrisbra, line 320: instead of :redir, you can use `execute()`:

```suggestion
  let res = execute('wviminfo Xviminfo')
  call assert_match('W23:', res)
```

ElectricPulse: I commited it along with the `silent!` since that allows the capturing of the warning

src/optiondefs.h
ElectricPulse: fixed that

k-takata: options.txt should be also updated accordingly:
https://github.com/vim/vim/blob/701ad50a9efcf0adfe6d787b606c4e4dbd31f26d/runtime/doc/options.txt#L9032-L9034

src/viminfo.c
```diff
@@ -2895,17 +2905,18 @@ read_viminfo_up_to_marks(
 /*
  * do_viminfo() -- Should only be called from read_viminfo() & write_viminfo().
  */
-    static void
+    static int
```
chrisbra, line 2908: What is the reason for returning a return value here? It seems like all callers of `do_viminfo()` do not check the return value, so I guess it is no needed?

chrisbra:
I did check the vim-history repo and the `s` flag (max size of registers) has been added to the viminfo option as of [patch 6.2.393](https://github.com/vim/vim-history/commit/98e997387c18473e1e49875bd8) (from around Mar 2004). So I think, increasing the limit by 10 times to 100 KBytes should be okay after more than 20 years. I didn't try to find when the `<`/`"`  item has been set initially. So all in all I think this is good change

chrisbra:
@ElectricPulse are you going to address those few points please?

ElectricPulse:
> @ElectricPulse are you going to address those few points please?
Yes, it is a macos test failing and my macos vm is taking forever to install.

chrisbra:
This one: 
```
Expected {'id': 15, 'jsonrpc': '2.0', 'result': 'delayed-payload'} but got {}
```
is a known flaky one. 

Not sure about `Test_popupwin_beval_3.dump` for macos 12, huge. Let me try to re-trigger this one.

benknoble:
A late suggestion, but one I hope will be considered: perhaps this is better hooked into `'confirm'`? I know current documentation says it's about fallible operations due to unsaved changes, but this feels like it's in that vein of problems (give me the opportunity to notice that something is wrong when quitting).

In fact, I suspect most of the warning + prompt code is already available in the confirm implementation; this would only need to perform the check from this PR in the right spot?

chrisbra:
closing for now. If you ever finish this up, please open a new PR. thanks

PR #14788, opened 2024-05-16
Introduce some ireggularity in screen states

luukvbaal:
Currently some of the states do not visibly change after key press


chrisbra:
thanks!

PR #14796, opened 2024-05-17
Refactor to remove STRLEN() part 6

basilisk0315:
This PR reduces the number of calls to STRLEN() in search.c. There is nothing tricky here but because all of the call sites needed to change, there are 15 source files involved.

Cheers
John

### DIFF
--- a/src/cmdhist.c
+++ b/src/cmdhist.c
@@ -297,11 +297,11 @@ static int	last_maptick = -1;	// last seen maptick
 add_to_history(
     int		histype,
     char_u	*new_entry,
+    size_t	new_entrylen,
     int		in_map,		// consider maptick when inside a mapping
     int		sep)		// separator character used (search hist)
 {
     histentry_T	*hisptr;
-    int		len;
 
     if (hislen == 0)		// no history
 	return;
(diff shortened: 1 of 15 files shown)


src/cmdhist.c
```diff
@@ -566,7 +565,7 @@ f_histadd(typval_T *argvars UNUSED, typval_T *rettv)
 	return;
 
     init_history();
-    add_to_history(histype, str, FALSE, NUL);
+    add_to_history(histype, str, STRLEN(str), FALSE, NUL);
```
chrisbra, line 568: Hm, that doesn't seem like a clear improvement here. Instead of calling `strlen()` inside the function, you'll have to add the `strlen()` when calling the function.

basilisk0315: Yes it looks weird. However, this function is called from five places (including this one) and in the other four places the string's length in known and passed. It's only this one where STRLEN() has to be called.

A similar thing happens for other functions as well.

chrisbra: okay thanks

src/tag.c
```diff
@@ -3936,14 +3938,14 @@ jumpto_tag(
...
+		    pbuflen = vim_snprintf((char *)pbuf, LSIZE, "^%s\\s\\*(", tagp.tagname);
+		    if (!do_search(NULL, '/', '/', pbuf, pbuflen, (long)1,
 							 search_options, NULL))
 		    {
 			// Guess again: "^char * \<func  ("
-			sprintf((char *)pbuf, "^\\[#a-zA-Z_]\\.\\*\\<%s\\s\\*(",
+			pbuflen = vim_snprintf((char *)pbuf, LSIZE, "^\\[#a-zA-Z_]\\.\\*\\<%s\\s\\*(",
 								tagp.tagname);
-			if (!do_search(NULL, '/', '/', pbuf, (long)1,
+			if (!do_search(NULL, '/', '/', pbuf, len, (long)1,
```
zeertzjq, line 3948: Why is `len` used here? Shouldn't this be `pbuflen`?

zeertzjq: I've created PR #14817 to correct this.

chrisbra: sorry, I missed that during review.

PR #14799, opened 2024-05-18
Add additional tests for eval.c

yegappan:


src/testdir/test_vim9_script.vim
```diff
@@ -2510,6 +2510,13 @@ def Test_for_loop()
         reslist->add('x')
       endfor
       assert_equal(['x', 'x', 'x'], reslist)
+
+      # Test for ignoring loop var assignment
```
chrisbra: is that comment correct?

src/testdir/test_vimscript.vim
```diff
@@ -7509,13 +7509,6 @@ func Test_for_over_string()
     let res ..= c .. '-'
   endfor
   call assert_equal('', res)
-
-  " Test for ignoring loop var assignment
-  let c = 0
-  for _ in 'abc'
-    let c += 1
-  endfor
-  call assert_equal(3, c)
```
chrisbra, line 7518: should we keep this legacy test?

yegappan: I added this test to test_vimscript.vim instead of test_vim9_script.vim in my previous PR by mistake.  Using `_` to ignore a value in a for loop is supported only in a Vim9 script.  In the legacy script, a variable named `_` is used.

For example, the following legacy script:

```
for _ in range(4)
    echo _
endfor
```
will output:

```
0
1
2
3
```

But the following Vim9 script:

```
vim9script
for _ in range(4)
    echo _
endfor
```

will output:

```
Error detected while processing :source buffer=1:
line    4:
E1181: Cannot use an underscore here
```

I will modify these tests to check for these.

src/testdir/test_vim9_script.vim
yegappan: Yes.  The `_` is used to ignore the for loop variable assignment.

chrisbra:
Ah, I missed that. Thanks!

PR #14800, opened 2024-05-19
runtime(sql): set commentstring for sql files

ribru17:
Defines comment strings (`--`) for SQL-family files


chrisbra:
Thanks! I slightly changed style, but is good to merge.

PR #14801, opened 2024-05-19
runtime(graphql): basic ftplugin file for graphql

ribru17:
Adds a minimal graphql ftplugin file, for comment strings


chrisbra:
thanks!

PR #14802, opened 2024-05-19
runtime(dart): add basic dart ftplugin file

### LINKED ISSUE #14793: Add commentstring for dart filetypes
Missing commentstring for dart filetypes

here is the documentation for comments in Dart https://dart.dev/language/comments

---

I am not familiar with contributing to Vim as I was directed here from NeoVim but I would assume it would be similar to this https://github.com/vim/vim/pull/13921/files and this is from the dart-vim-plugin https://github.com/dart-lang/dart-vim-plugin/blob/master/ftplugin/dart.vim

Here is the reference from NeoVim https://github.com/neovim/neovim/issues/28812

ribru17:
Closes #14793


chrisbra:
thanks!

PR #14805, opened 2024-05-20
getregionpos() wrong with blockwise mode and multibyte

zeertzjq:
Problem:  getregionpos() wrong with blockwise mode and multibyte.
Solution: Use textcol and textlen instead of start_vcol and end_vcol.
          Handle coladd properly.

Also remove unnecessary buflist_findnr() in add_regionpos_range(), as
getregionpos() has already switched buffer.


chrisbra:
thanks!

girishji:
Thanks!

PR #14807, opened 2024-05-20
runtime(prisma): basic prisma ftplugin file

ribru17:
Reference: https://www.prisma.io/docs/orm/prisma-schema/overview#comments


chrisbra:
thanks!

PR #14808, opened 2024-05-20
runtime(slint): slint ftplugin, filetype detection

ribru17:
Reference: https://releases.slint.dev/1.5.0/docs/slint/src/language/syntax/comments


chrisbra:
thanks!

PR #14810, opened 2024-05-20
runtime(verilog): set commentstring option

ribru17:


runtime/ftplugin/verilog.vim
```diff
@@ -31,6 +31,7 @@ setlocal fo-=t fo+=croqlm1
 
 " Set 'comments' to format dashed lists in comments.
 setlocal comments=sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,://
+setlocal commentstring=//\ %s
```
chrisbra, line 35: please add it to `b:undo_ftplugin` variable above

chrisbra:
Thanks, please also amend the "Last Change" header line at the top of the file.

ribru17:
Ah of course, thanks!

runtime/ftplugin/verilog.vim
```diff
@@ -2,7 +2,7 @@
 " Language:	Verilog HDL
 " Maintainer:	Chih-Tsun Huang <cthuang@cs.nthu.edu.tw>
 " Last Change:	2017 Aug 25 by Chih-Tsun Huang
-"		2024 Jan 14 by Vim Project (browsefilter)
+"		2024 May 20 by Vim Project (browsefilter)
```
chrisbra: please keep the existing line and add a new one for the commentstring option. Thanks!

ribru17: Ah sorry :sweat_smile:

PR #14811, opened 2024-05-20
runtime(arduino): Add Arduino ftplugin and indent files

k-takata:
We already have Arduino syntax file, but we didn't have ftplugin and indent files.

This commit adds a basic ftplugin file and a basic indent file.
Both of them are derived from {ftplugin,indent}/c.vim.


k-takata:
TODO:

Arduino IDE uses the following style for function definition:
```c
void functionname(void) {
    ...
}
```
Vim's function motions like `[[` and `]]` don't work with this style. This should be supported in the future.

chrisbra:
Thanks!

PR #14812, opened 2024-05-20
runtime(nim): basic nim ftplugin file for comments

ribru17:
Reference: https://nim-by-example.github.io/comments/#:~:text=Comments%20in%20Nim%20begin%20with%20the%20hash%20character.&text=Multiline%20or%20block%20comments%20begin,line%20comments%20can%20be%20nested.


chrisbra:
thanks!

PR #14813, opened 2024-05-20
runtime(purescript): purescript filetype, ftplugin

ribru17:
Reference: https://github.com/purescript/documentation/blob/master/language/Syntax.md#comments


chrisbra:
thanks!

PR #14817, opened 2024-05-21
Tag guessing leaves wrong search history with very short names

zeertzjq:
Problem:  Tag guessing leaves wrong search history with very short names
          (after 9.1.0426).
Solution: Use the correct variable for pattern length.


chrisbra:
thanks for being careful and providing a test!

PR #14818, opened 2024-05-21
Revise termdebug mapping tests

k-takata:
Expected and actual were reversed in `Test_termdebug_mapping()`.
Reverse them. Use `assert_{true,false}()` if suitable.


k-takata:
Related: 7fbbd7fdc6df9dc198b3735cfbe8dbe8afd646f9

Shane-XB-Qian:
@Takata thanks, looks I didnot notice the expected/actual arg order.

--
shane.xb.qian

PR #14822, opened 2024-05-21
runtime(rescript): basic rescript ftplugin file

ribru17:
Reference: https://rescript-lang.org/docs/manual/latest/overview#comments


chrisbra:
thanks!

PR #14824, opened 2024-05-22
Move the garbage collection code from eval.c into a separate gc.c file

yegappan:
Also move the f_slice() function to list.c from eval.c


chrisbra:
thanks!

PR #14825, opened 2024-05-22
getregionpos() doesn't handle selection contained in one char

zeertzjq:
Problem:  getregionpos() doesn't handle selection contained in one char.
Solution: Handle startspaces differently when is_oneChar is set.
          Also add a test for an exclusive charwise selection with
          multibyte chars.


chrisbra:
thanks!

PR #14827, opened 2024-05-22
fix(swayconfig): floating_modifier

### LINKED ISSUE #14826: Minor syntax highlighting issue with swayconfig files
### Steps to reproduce

This is a continuation of https://github.com/vim/vim/issues/14752, syntax highlighting improved a lot but there is still a minor issue with the following config:

    floating_modifier $mod normal

The "normal" string shows up as red.

Here's a screenshot of what it currently looks like:

20240522_15h22m56s_grim

### Expected behaviour

Syntax highlighting is properly applied to all commands.

### Version of Vim

9.1.429

### Environment

Operating system: Arch Linux
Terminal: foot
Value of $TERM: foot
Shell: /usr/bin/bash

### Logs and stack traces

_No response_

jamespeapen:
Restores correct highlighting of swayconfig's `floating_modifier` with the `normal` and `inverse` options. Both i3 and sway have this keyword, but i3 does not have the `normal|inverse`. When this match was set as `contained` in 679f5ab, it prevented the sway options from being recognized.

Added the `none` option which was not supported but is a [valid option](https://github.com/swaywm/sway/blob/970415241497ceccfb013b6f8cb2395abee74e5c/sway/sway.5.scd?plain=1#L680) in sway config.

Ideally `floating_modifier` would  be a keyword as it is in `i3config.vim`, but I'm still working that out.

Fixes #14826

### DIFF
--- a/runtime/syntax/i3config.vim
+++ b/runtime/syntax/i3config.vim
@@ -2,8 +2,8 @@
 " Language: i3 config file
 " Original Author: Josef Litos (JosefLitos/i3config.vim)
 " Maintainer: Quentin Hibon (github user hiqua)
-" Version: 1.2.2
-" Last Change: 2024-04-14
+" Version: 1.2.3
+" Last Change: 2024-05-23
 
 " References:
 " http://i3wm.org/docs/userguide.html#configuring
@@ -67,7 +67,7 @@ syn keyword i3ConfigBindKeyword bindsym bindcode contained skipwhite nextgroup=i
 syn region i3ConfigModeBlock matchgroup=i3ConfigKeyword start=/mode\ze\( --pango_markup\)\? \([^'" {]\+\|'[^']\+'\|".\+"\)\s\+{$/ end=/^}\zs$/ contained contains=i3ConfigShParam,@i3ConfigStrVar,i3ConfigBindKeyword,i3ConfigComment,i3ConfigParen fold keepend extend
 
 " 4.7 Floating modifier
-syn keyword i3ConfigKeyword floating_modifier contained skipwhite nextgroup=i3ConfigVariable,i3ConfigBindModkey
+syn match i3ConfigKeyword /^floating_modifier [$0-9A-Za-z]*$/ contains=i3ConfigVariable,i3ConfigBindModkey
 
 " 4.8 Floating window size
 syn keyword i3ConfigSizeSpecial x contained
--- a/runtime/syntax/swayconfig.vim
+++ b/runtime/syntax/swayconfig.vim
@@ -2,8 +2,8 @@
 " Language: sway config file
 " Original Author: Josef Litos (JosefLitos/i3config.vim)
 " Maintainer: James Eapen <james.eapen@vai.org>
-" Version: 1.2.2
-" Last Change: 2024-05-12
+" Version: 1.2.3
+" Last Change: 2024-05-23
 
 " References:
 " http://i3wm.org/docs/userguide.html#configuring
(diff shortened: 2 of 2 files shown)


jamespeapen:
setting as draft since some issues remain:
image

The first two lines should show `bad normal` as errors but do not.

Using 

```vim
syn match i3ConfigKeyword /floating_modifier \(none$\|$[a-zA-Z0-9+]\+ \(normal\|inverse\)\)$/ contains=i3ConfigVariable,i3ConfigBindModkey,swayConfigFloatingModifierOpts
```
with the `$` outside the alphanumeric match gets
image

Here it looks better except that Mod keys aren't allowed anything following them.

jamespeapen:
Made the capture groups a little more explicit:
image

Let me know if you have any suggestions @JosefLitos. Otherwise I'm happy with this. Since both i3 and sway don't share too many different options for the same keyword like `floating_modifier`, I'm okay with the current fix.

JosefLitos:
This is a workaround to the issue, not actually fixing the root of it. The reall source of the problem is a change in `i3config.vim floating_modifier` implementation. I went from a match to a keyword, which takes precedence.

To make them work both, I'd suggest leaving the `contained` keyword in the swayconfig part and replacing `i3config` part with:

```diff
" 4.7 Floating modifier
- syn keyword i3ConfigKeyword floating_modifier contained skipwhite nextgroup=i3ConfigVariable,i3ConfigBindModkey
+ syn match i3ConfigKeyword /floating_modifier [$a-zA-Z0-9+]\+$/ contained contains=i3ConfigVariable,i3ConfigBindModkey
```

jamespeapen:
Thanks! I've reverted the i3 change that caused the issue @hiqua. For the precise matches, I think its more consistent to be precise since other syntax elements are precisely matched, but you have a point that its not going to be used more than once and vim needn't act as a sway linter. I'll go with the simpler option for now but will keep trying to see if there's another way to get that precision.

For a start I did notice two of the errors are recognized if there are any leading spaces:
image

chrisbra:
Thanks. So is this good to be included? Or are you still working on it? @JosefLitos can you please ACK as well please?

jamespeapen:
Yes its good to go.

runtime/syntax/i3config.vim
```diff
@@ -67,7 +67,7 @@ syn keyword i3ConfigBindKeyword bindsym bindcode contained skipwhite nextgroup=i
 syn region i3ConfigModeBlock matchgroup=i3ConfigKeyword start=/mode\ze\( --pango_markup\)\? \([^'" {]\+\|'[^']\+'\|".\+"\)\s\+{$/ end=/^}\zs$/ contained contains=i3ConfigShParam,@i3ConfigStrVar,i3ConfigBindKeyword,i3ConfigComment,i3ConfigParen fold keepend extend
 
 " 4.7 Floating modifier
-syn keyword i3ConfigKeyword floating_modifier contained skipwhite nextgroup=i3ConfigVariable,i3ConfigBindModkey
+syn match i3ConfigKeyword /^floating_modifier [$0-9A-Za-z]*$/ contains=i3ConfigVariable,i3ConfigBindModkey
```
litoj, line 70: I may have miscommunicated the changes in i3. The reverted changes in i3 introduce back the issue for which they were made - leading spaces.
We didn't need to put back the start-of-line anchor, but we should have kept in the `contained` keyword. I wrote the exact changes in the original comment, but a _revert_ was probably not the best way to describe them.

Sorry for the late review.

JosefLitos:
> For a start I did notice two of the errors are recognized if there are any leading spaces...

The reason leading spaces play a role is the inclusion of `^` in the i3config regex. Otherwise it would be also recognized as valid syntax. That is because of the simplified highlighting. Since that could be quite a common mistake to write it that way, a small simple change could be made for detecting the first character ->

i3config:
```vim
syn match i3ConfigKeyword /floating_modifier [$A-Z][0-9A-Za-z]*$/ contained contains=i3ConfigVariable,i3ConfigBindModkey
```

swayconfig:
```vim
syn match i3ConfigKeyword /floating_modifier \(none\|[$A-Z][a-zA-Z0-9+]\+ \(normal\|inverse\)\)$/ contained contains=i3ConfigVariable,i3ConfigBindModkey,swayConfigFloatingModifierOpts
```

(added `[$A-Z]` before the rest of the key/var match)

jamespeapen:
Ah I did focus more on the revert and the sway suggestion than the i3 diff. Thanks @JosefLitos, and apologies @chrisbra the new pull request should supersede this one.

PR #14828, opened 2024-05-23
runtime(zig): Remove zig autoload install/uninstall

derekschrock:
d1d9316c6 removed autoload/zig builds fail since the dir doesn't exist and no .vim files are present.


derekschrock:
I believe the PR still applies however this appears to be an issue with BSD
make.  GNU make just reports the error but continues where BSD make stops
during install.

chrisbra:
Thanks, I was wondering, because I just did a fresh install and did not even see an error :/

PR #14830, opened 2024-05-23
Wrong yanking with exclusive selection and ve=all

zeertzjq:
Problem:  Wrong yanking with exclusive selection and virtualedit=all,
          and integer overflow when using getregion() on it.
Solution: Set coladd when decreasing column and 'virtualedit' is active.
          Add more tests for getregion() with 'virtualedit'.


chrisbra:
thanks!

PR #14832, opened 2024-05-23
Crash when using '?' as separator for :s

### LINKED ISSUE #28935: Neovim crashes on nightly
### Problem

I tried using a keymap which I use every day.
using the latest nightly build, it crashes nvim.

### Steps to reproduce

```sh
cat <<'EOF' > init.vim
let g:mapleader = ' '
vnoremap <leader>r "hy:.,$s?<C-r>h?<C-r>h?gc<left><left><left>
EOF

cat <<'EOF' > myfile
bla bla \
bla bla \
EOF

nvim --clean -u init.vim myfile
```

now visually select the space and the slash after it,
hit <kbd>Leader</kbd><kbd>r</kbd>
try to move, nvim crashes.

### Expected behavior

nvim doesn't crash

### Neovim version (nvim -v)

NVIM v0.11.0-dev-57+gcd48b72b6

no, VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 13 2024 12:27:48)

### Operating system/version

macOS 14.5

### Terminal name/version

iterm2

xterm-256color

### Installation

asdf.vm

zeertzjq:
Problem:  Crash when using '?' as separator for :s and pattern contains
          escaped '?'s (after 9.1.0409).
Solution: Always compute startplen.

Fix neovim/neovim#28935

### DIFF
--- a/src/regexp.c
+++ b/src/regexp.c
@@ -620,6 +620,7 @@ skip_regexp_ex(
 {
     magic_T	mymagic;
     char_u	*p = startp;
+    size_t	startplen = STRLEN(startp);
 
     if (magic)
 	mymagic = MAGIC_ON;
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks for this and watching out for problems. I was afraid that the strlen() refactor could cause issues like this.

chrisbra:
included as of 98ec4754e

PR #14833, opened 2024-05-23
filetype: Recognize Cygwin's *.cygport file

k-takata:
Recognize *.cygport as bash.
https://cygwin.github.io/cygport/cygport_in.html


chrisbra:
thanks!

PR #14834, opened 2024-05-23
Fix tiny and normal builds for Motif.

dvogel:
Problem: Motif requires non-const char pointer for XPM data shared with GTK.
Solution: Cast non-const to const char pointer for GTK.


chrisbra:
thanks, for fixing this. I just noticed your email address is invalid. I suppose this is done intentionally?

PR #14836, opened 2024-05-23
runtime(hare): update hare.vim to match upstream

selenebun:
This is a (long overdue) update of the hare.vim included within Vim to match the current state of the [upstream repository][1]. I also modified the username in the `MAINTAINERS` file as I have migrated to this account from my previous one (@rsaihe).

Outside of the various other changes, this also adds a new `haredoc` filetype for documentation files inside Hare modules. Per Hare's [documentation][2], these files use the name `README`. As this is a very common file name, automatic file detection must be opted into using `g:filetype_haredoc`.

When enabled, a recursive directory search is used to search for Hare source files. If any are found, the current directory is considered to be a Hare module and the `README` file is given the `haredoc` filetype. By default, this search recurses into only the first level of subdirectories, but this can be configured with `g:haredoc_search_depth`.

I did my best to ensure any potential performance impact for the automatic filetype detection remains minimal, especially given `README`'s prevalence as a file name. Hopefully `g:filetype_haredoc` being opt-in is sufficient, but I'm willing to pursue alternatives if not.

[1]: https://git.sr.ht/~sircmpwn/hare.vim
[2]: https://harelang.org/documentation/usage/documentation.html

### DIFF
--- a/runtime/autoload/hare.vim
+++ b/runtime/autoload/hare.vim
@@ -1,0 +1,26 @@
+" Vim autoload file.
+" Language:     Hare
+" Maintainer:   Amelia Clarke <selene@perilune.dev>
+" Last Updated: 2024-05-10
+" Upstream:     https://git.sr.ht/~sircmpwn/hare.vim
+
+" Attempt to find the directory for a given Hare module.
+function hare#FindModule(str)
+  let path = substitute(trim(a:str, ':', 2), '::', '/', 'g')
+  let dir = finddir(path)
+  while !empty(path) && empty(dir)
+    let path = substitute(path, '/\?\h\w*$', '', '')
+    let dir = finddir(path)
+  endwhile
+  return dir
+endfunction
+
+" Return the value of HAREPATH if it exists. Otherwise use a reasonable default.
+function hare#GetPath()
+  if empty($HAREPATH)
+    return '/usr/src/hare/stdlib,/usr/src/hare/third-party'
+  endif
+  return substitute($HAREPATH, ':', ',', 'g')
+endfunction
+
+" vim: et sts=2 sw=2 ts=8
--- a/runtime/indent/hare.vim
+++ b/runtime/indent/hare.vim
@@ -1,19 +1,16 @@
 " Vim indent file
-" Language: Hare
-" Maintainer: Amelia Clarke <me@rsaihe.dev>
-" Last Change: 2022 Sep 22
-"              2023 Aug 28 by Vim Project (undo_indent)
+" Language:    Hare
+" Maintainer:  Amelia Clarke <selene@perilune.dev>
+" Last Change: 2024-04-14
+" Upstream:    https://git.sr.ht/~sircmpwn/hare.vim
 
-if exists("b:did_indent")
+if exists('b:did_indent')
   finish
 endif
 let b:did_indent = 1
 
-if !has("cindent") || !has("eval")
-  finish
-endif
-
-setlocal cindent
+let s:cpo_save = &cpo
+set cpo&vim
 
 " L0 -> don't deindent labels
 " (s -> use one indent after a trailing (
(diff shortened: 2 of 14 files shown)


chrisbra:
thanks!

chrisbra:
Thanks. While merging I noticed a few minor problems so I added the following patch on top:
```diff
diff --git a/runtime/doc/ft_hare.txt b/runtime/doc/ft_hare.txt
index 7fe28a13e..937c5e096 100644
--- a/runtime/doc/ft_hare.txt
+++ b/runtime/doc/ft_hare.txt
@@ -43,7 +43,7 @@ The following options are set by default, in accordance with the official Hare
 style guide: >
        setlocal noexpandtab
        setlocal shiftwidth=0
-       setlocal softtabstop=8
+       setlocal softtabstop=0
        setlocal tabstop=8
        setlocal textwidth=80
 <
diff --git a/runtime/ftplugin/hare.vim b/runtime/ftplugin/hare.vim
index ef51c7141..422bb7b4e 100644
--- a/runtime/ftplugin/hare.vim
+++ b/runtime/ftplugin/hare.vim
@@ -31,7 +31,7 @@ setlocal suffixesadd=.ha
 setlocal path-=/usr/include,,
 let &l:path .= ',' .. hare#GetPath() .. ',,'

-let b:undo_ftplugin = 'setl cms< com< flp< fo< inc< inex< isf< pa< sua<'
+let b:undo_ftplugin = 'setl cms< com< flp< fo< inc< inex< isf< pa< sua< mp<'

 " Follow the Hare style guide by default.
 if get(g:, 'hare_recommended_style', 1)
```

You might want to take that change back upstream.

selenebun:
Oh, oops! Those are both very good finds, thank you!

selenebun:
Should `efm<` also be added to `b:undo_ftplugin`? I didn't think to reset the options set with `CompilerSet` before but I assume that one should also be reset.

chrisbra:
Oh yes, that should also be reset. Thanks for checking.

runtime/autoload/hare.vim
```diff
@@ -0,0 +1,26 @@
+" Vim autoload file.
+" Language:     Hare
+" Maintainer:   Amelia Clarke <selene@perilune.dev>
+" Last Updated: 2024-05-10
```
dkearns, line 4: Unless this header means something at your end could we change the format in all files to "Last Change: YYYY Mon DD" for the next update?

runtime/ftplugin/hare.vim
```diff
@@ -1,35 +1,59 @@
...
 
 if exists('b:did_ftplugin')
   finish
 endif
 let b:did_ftplugin = 1
 
-" Formatting settings.
-setlocal formatoptions-=t formatoptions+=croql/
+let s:cpo_save = &cpo
+set cpo&vim
```
dkearns, line 13: This is usually only needed when line continuations are used in the file. So unless I'm missing some other reason to do so the setting/resetting of 'cpo' can be removed from both ftplugins.

runtime/indent/hare.vim
```diff
@@ -41,7 +38,11 @@ setlocal cinwords=if,else,for,switch,match
 
 setlocal indentexpr=GetHareIndent()
```
dkearns, line 39: This and similar options can now use a script local function if you prefer.  See `:help indentexpr` for an example.

dkearns (commented):
I just added a couple of suggestions for the next update.

runtime/ftplugin/hare.vim
selenebun: Mmmm, I'm pretty sure I added those as a safety measure just in case any other contributors happen to use line continuations in the file. But as I'm the primary maintainer of the plugin, that situation feels pretty unlikely in practice. I'm more than happy to remove them in the places that are unnecessary.

runtime/indent/hare.vim
selenebun: I've been working on refactoring a lot of the indentation for a while and have done that for all the functions in the indent file, but the whole thing is still pretty WIP.

runtime/autoload/hare.vim
selenebun: Sure! I personally prefer using `YYYY-MM-DD` as a date format, but I am totally fine with changing it if that's the preferred format for runtime files.

PR #14837, opened 2024-05-23
Wrong Ex command executed when :g uses '?' as delimiter

zeertzjq:
Problem:  Wrong Ex command executed when :g uses '?' as delimiter and
          pattern contains escaped '?'.
Solution: Don't use "*newp" when it's not allocated.


chrisbra:
thanks!

PR #14838, opened 2024-05-24
getregionpos() can't properly indicate positions beyond eol

zeertzjq:
Problem:  getregionpos() can't properly indicate positions beyond eol.
Solution: Add an "eol" flag that enables handling positions beyond end
          of line like getpos() does.

Also fix the problem that a position still has the coladd beyond the end
of the line when its column has been clamped.  In the last test case
with TABs at the end of the line the old behavior is obviously wrong.

I decided to gate this behind a flag because returning positions that
don't correspond to actual characters in the line may lead to mistakes
for callers that want to calculate the length of the selected text, so
the behavior is only enabled if the caller wants it.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1,4 +1,4 @@
-*builtin.txt*	For Vim version 9.1.  Last change: 2024 May 22
+*builtin.txt*	For Vim version 9.1.  Last change: 2024 May 24
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

runtime/doc/builtin.txt
```diff
@@ -4348,6 +4348,19 @@ getregionpos({pos1}, {pos2} [, {opts}])            *getregionpos()*
 		the offset of the character's first cell not included in the
 		selection, otherwise all its cells are included.
 
+		Apart from the options supported by |getregion()|, {opts} also
+		supports the following:
+
+			eol		If |TRUE|, indicate positions beyond
+					the end of a line with "col" values
+					one more than the length of the line.
```
justinmk, line 4356: Was this intentionally chosen instead of `v:maxcol` ?

zeertzjq: This is consistent with the behavior `getpos()` etc. when 'virtualedit' is active.

PR #14839, opened 2024-05-24
Refactor the get_lval() function

yegappan:
The get_lval() function is too long.  Move the code for processing a list/dict/blob/object subitem into a separate function.


chrisbra:
thanks!

PR #14841, opened 2024-05-24
fixup(i3/swayconfig): `floating_modifier`

litoj:
Fixup of #14827 

Also removed the `+` from valid modkey sequence because `floating_modifier` doesn't allow multiple keys (whereas `bar.modifier` does, so that's why it was there originally).

(@jamespeapen)


jamespeapen:
LGTM and it fixes the issue I had previously seen. 
image

chrisbra:
Thanks!

PR #14843, opened 2024-05-24
feat(runtime)!: add padding to commentstrings

ribru17:
Changes the default 'commentstring' value from '/*%s*/' to '/* %s */', and changes many ftplugin files to add space padding in their comment strings, if possible.

---

I understand if this PR is too invasive, and if needed I can amend it to only include ftplugin changes, without changing the default value of `commentstring`. Using [this query](https://github.com/search?q=repo%3Avim%2Fvim+%2F%28commentstring%7Ccms%29.*%5B%5E+%5D%25s%2F+path%3Aruntime&type=code), we can see that about a fifth of the specified `commentstring`s in various `ftplugin`s are *not* given space padding, while the rest are. This PR is an attempt to standardize the space padding more (when possible in the language). The only one I didn't touch was COBOL, because it is already kind of wonky and column-specific. Of the ones I did change, I gave a quick spec-read and github code search to make sure that the space-padded comments are still valid syntax in that language. Sorry for the large review!

### DIFF
--- a/runtime/ftplugin/abaqus.vim
+++ b/runtime/ftplugin/abaqus.vim
@@ -3,6 +3,7 @@
 " Maintainer:   Carl Osterwisch <costerwi@gmail.com>
 " Last Change:  2022 Oct 08
 "               2024 Jan 14 by Vim Project (browsefilter)
+"               2024 May 23 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 
 " Only do this when not done yet for this buffer
 if exists("b:did_ftplugin") | finish | endif
@@ -27,7 +28,7 @@ setlocal isfname-=,
 
 " Define format of comment lines (see 'formatoptions' for uses)
 setlocal comments=:**
-setlocal commentstring=**%s
+setlocal commentstring=**\ %s
 
 " Definitions start with a * and assign a NAME, NSET, or ELSET
 " Used in [d ^wd and other commands
--- a/runtime/ftplugin/racket.vim
+++ b/runtime/ftplugin/racket.vim
@@ -3,8 +3,7 @@
 " Maintainer:           D. Ben Knoble <ben.knoble+github@gmail.com>
 " Previous Maintainer:  Will Langstroth <will@langstroth.com>
 " URL:                  https://github.com/benknoble/vim-racket
-" Last Change:          2022 Aug 29
-"                       2024 Jan 14 by Vim Project (browsefilter)
+" Last Change:          2024 May 28
 
 if exists("b:did_ftplugin")
   finish
@@ -21,7 +20,7 @@ setlocal iskeyword=@,!,#-',*-:,<-Z,a-z,~,_,94
 setlocal comments=:;;;;,:;;;,:;;,:;
 setlocal formatoptions+=r
 
-"setlocal commentstring=;;%s
+"setlocal commentstring=;;\ %s
 setlocal commentstring=#\|\ %s\ \|#
 
 setlocal formatprg=raco\ fmt
--- a/runtime/ftplugin/vim.vim
+++ b/runtime/ftplugin/vim.vim
@@ -2,6 +2,7 @@
 " Language:		Vim
 " Maintainer:		Doug Kearns <dougkearns@gmail.com>
 " Last Change:		2024 Apr 13
+" 			2024 May 23 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 " Former Maintainer:	Bram Moolenaar <Bram@vim.org>
 
 " Only do this when not done yet for this buffer
@@ -51,7 +52,7 @@ setlocal keywordprg=:help
 
 " Comments starts with # in Vim9 script.  We have to guess which one to use.
 if "\n" .. getline(1, 32)->join("\n") =~# '\n\s*vim9\%[script]\>'
-  setlocal commentstring=#%s
+  setlocal commentstring=#\ %s
 else
   setlocal commentstring=\"%s
 endif
--- a/runtime/ftplugin/xdefaults.vim
+++ b/runtime/ftplugin/xdefaults.vim
@@ -2,6 +2,7 @@
 " Language:             X resources files like ~/.Xdefaults (xrdb)
 " Previous Maintainer:  Nikolai Weibull <now@bitwi.se>
 " Latest Revision:      2008-07-09
+"                       2024-06-03 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 
 if exists("b:did_ftplugin")
   finish
@@ -13,7 +14,7 @@ set cpo&vim
 
 let b:undo_ftplugin = "setl com< cms< inc< fo<"
 
-setlocal comments=s1:/*,mb:*,ex:*/,:! commentstring& inc&
+setlocal comments=s1:/*,mb:*,ex:*/,:! commentstring=!\ %s inc&
 setlocal formatoptions-=t formatoptions+=croql
 
 let &cpo = s:cpo_save
--- a/.github/MAINTAINERS
+++ b/.github/MAINTAINERS
@@ -176,6 +176,7 @@ runtime/ftplugin/kotlin.vim		@udalov
 runtime/ftplugin/less.vim		@genoma
 runtime/ftplugin/liquid.vim		@tpope
 runtime/ftplugin/lua.vim		@dkearns
+runtime/ftplugin/lc.vim			@ribru17
 runtime/ftplugin/lynx.vim		@dkearns
 runtime/ftplugin/m3build.vim		@dkearns
 runtime/ftplugin/m3quake.vim		@dkearns
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1914,8 +1914,9 @@ A jump table for the options with a short description can be found at |Q_op|.
 			{not available when compiled without the |+folding|
 			feature}
 	A template for a comment.  The "%s" in the value is replaced with the
-	comment text.  Currently only used to add markers for folding, see
-	|fold-marker|.  Also used by comment plugins |comment-install|.
+	comment text, and should be padded with a space when possible.
+	Currently only used to add markers for folding, see |fold-marker|.
+	Also used by comment plugins |comment-install|.
 
 			*'compatible'* *'cp'* *'nocompatible'* *'nocp'*
 'compatible' 'cp'	boolean	(default on, off when a |vimrc| or |gvimrc|
--- a/runtime/ftplugin/arduino.vim
+++ b/runtime/ftplugin/arduino.vim
@@ -3,6 +3,7 @@
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
 "		Ken Takata <https://github.com/k-takata>
 " Last Change:	2024 Apr 12
+"		2024 Jun 02 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 "
 " Most of the part was copied from c.vim.
 
@@ -32,7 +33,7 @@ setlocal fo-=t fo+=croql
 
 " These options have the right value as default, but the user may have
 " overruled that.
-setlocal commentstring& define& include&
+setlocal commentstring=/*\ %s\ */ define& include&
 
 " Set completion with CTRL-X CTRL-O to autoloaded function.
 if exists('&ofu')
--- a/runtime/ftplugin/asm.vim
+++ b/runtime/ftplugin/asm.vim
@@ -4,13 +4,14 @@
 " Last Change:	2020 May 23
 " 		2023 Aug 28 by Vim Project (undo_ftplugin)
 " 		2024 Apr 09 by Vim Project (add Matchit support)
+" 		2024 May 23 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 
 if exists("b:did_ftplugin") | finish | endif
 let b:did_ftplugin = 1
 
 setl include=^\\s*%\\s*include
 setl comments=:;,s1:/*,mb:*,ex:*/,://
-setl commentstring=;%s
+setl commentstring=;\ %s
 
 let b:undo_ftplugin = "setl commentstring< comments< include<"
 
--- a/runtime/ftplugin/astro.vim
+++ b/runtime/ftplugin/astro.vim
@@ -2,6 +2,7 @@
 " Language:     Astro
 " Maintainer:   Romain Lafourcade <romainlafourcade@gmail.com>
 " Last Change:  2024 Apr 21
+"               2024 May 24 by Riley Bruins <ribru17@gmail.com> ('commentstring')
 
 if exists("b:did_ftplugin")
     finish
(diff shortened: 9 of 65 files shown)


chrisbra:
Thanks for doing that work. I am not sure, it is a bit intrusive and not sure if all runtime file maintainers appreciate it, if we change some values behind their back.

I also noticed that there are apparently 2 syntax script, that set the commentstring option. I think it sholdn't be set there. This clearly belongs into a ftplugin. 

We have done such a mass change before with @dkearns browsefilter change, but perhaps we should still try to contact at least the maintainers that appear to be active.

Ping @tpope  @romainl  @gpanders @Freed-Wu @jamessan  @hcs42 @dkearns @nickeb96 @justinmk @zzzyxwvut @vito-c @habamax @ObserverOfTime @benknoble @rhysd 
can you please indicate if you are okay with the suggested change?

Also can you please amend your commit and add the reason for your change in the last Change Header? So instead of :
```
"                2024 Jan 14 by Vim Project (browsefilter)
"                2024 May 24 by Riley Bruins <ribru17@gmail.com>
```
Make it like this:
```
"                2024 Jan 14 by Vim Project (browsefilter)
"                2024 May 24 by Riley Bruins <ribru17@gmail.com> ('commentstring')
```


@dkearns whats your opinion on such a mass change?

romainl:
I'm okay. Actually, I have been doing just that in my "after" ftplugins for years. +1 from me.

habamax:
Well, it might be overridden (and probably would be) when the next update for the plugin would be merged in. Like, gdscript or odin. I guess the same is for other maintainers having source code in separate repos.

What is the purpose of the change? Apart of aesthetic?

ObserverOfTime (approved):

justinmk:
LGTM but want to hear from @echasnovski too

echasnovski:
Yeah, looks good to me too. The only thing that might need an extra attention is whether 'commentstring' with a space does not represent a comment in a particular filetype. As in "#aaa" is comment but "# aaa" is not.

habamax:
There are many maintainers that do not follow on a regular basis vim_dev or github prs.

What is the plan to push these changes upstream to respected repositories? What should be done if new version of ft/plugin is sent to vim_dev to be included where commentstring is not changed?

What is the purpose of adding padding to commentstring?

runtime/ftplugin/abaqus.vim
```diff
@@ -3,6 +3,7 @@
 " Maintainer:   Carl Osterwisch <costerwi@gmail.com>
 " Last Change:  2022 Oct 08
 "               2024 Jan 14 by Vim Project (browsefilter)
+"               2024 May 23 by Riley Bruins <ribru17@gmail.com>
```
dkearns:
```suggestion
"               2024 May 23 by Riley Bruins <ribru17@gmail.com> (commentstring)
```

Can you please make this change to all files.  It might help grab the maintainer's attention.

Depending on your editing skills you might want to wait until a decision has been made to accept the PR.

dkearns (changes requested):

clason:
> I generally favour this style and would also prefer that we improved consistency across the runtime files where possible. However, we've traditionally left this entirely to the discretion of the maintainers. It might be a good time to rethink that policy more generally.

I think that is the sticking point here: how much consistency (and implied ownership) the Vim project wants in their runtime files. Right now, there's a sliding scale between "fully owned" and "we import an upstream plugin as-is", which is semi-indicated by the "maintainer" line (and word of mouth). Personally, I would favor more consistency and oversight, but that of course implies (significant) additional maintenance work. Obviously that's a decision the project maintainers must make (and document) and beyond the scope of this PR.

One option could be to farm this out into a separate "filetype files" repository that does the integration work and from which Vim syncs regularly (inspired by the -- very successful -- vim/colorschemes project, and what vim-polyglot should have been IMO); this is also how treesitter queries are handled by Neovim. One drawback would be that we're back at big "update runtime files" blobs, and I much prefer the current "fun-size" runtime file commits.

jamessan:
I've handled the Debian related files in https://github.com/vim/vim/pull/14849.

zzzyxwvut:
Unify if you're sure that 4/5 are in the right and `&cms` is  
relevant (i.e. `fdm=marker` vs `fdm=syntax`).

I'll merge these changes whenever I have my own to apply.

ribru17:
Added the `('commentstring')` note, and also moved the `syntax` changes to appropriate `ftplugin` ones (one of the syntax files had it set in both so I didn't need to create a new `ftplugin` file).

> The only thing that might need an extra attention is whether 'commentstring' with a space does not represent a comment in a particular filetype.

I tested each file I changed to make sure the syntax is still valid with a space, it looks to be the case for all of them. Though my testing was loose, just a quick read of the language spec (if available) and brief github search

> What is the purpose of the change? Apart of aesthetic?

Besides aesthetic, just consistency basically.

A question from myself: should this PR add a small documentation note saying that space-padding is prefered in `commentstring`s?

chrisbra:
I am going to answer to several diferent points mentioned, see below.

I think the *TLDR* is, this change seems appreciated by most maintainers and I am happy to merge those filetype changes, but please revert the changes to the new default value of the 'commentstring' option. 

> What is the purpose of the change? Apart of aesthetic?

I think it is mainly consistency and aesthtics for commenting plugins. The hope is to prevent overwriting on next runtime update is that either during review we notice the unwanted change or the author notices it directly when creating a PR with the updates. However there are no guarantees, that we won't un-intentionally revert it... :/  Are you okay with a change to your maintained runtime files?

> The only thing that might need an extra attention is whether 'commentstring' with a space does not represent a comment in a particular filetype. As in "#aaa" is comment but "# aaa" is not.

@ribru17 mentioned he checked that, at least briefly. So I hope this wouldn't happen...

> The value change does introduce some small issues with existing files. E.g., deleting a fold in a C file created with the old value only deletes the fold marker text and not the entire comment. This might eventually be categorised as a bug.

That is a good point. I tested it briefly with `fold.c` from this repository. Vim seems to handle it gracefully, even so my `'commentstring'` was set to `/*%s*/` and I could successfully delete foldmarkers created with  `// %s`. But still I think we shouldn't change the default 'commentstring'  option.

> One option could be to farm this out into a separate "filetype files" repository that does the integration work and from which Vim syncs regularly 

I am not so much a fan of this, because it won't solve the issue of us making neccessary(?) changes to the filetype plugins, but adds additional burden to merge those new repo every now and then (and we typically only get complaints once it has been merged into the main repo (or nowadays once you have merged it down to Neovim). So for now I am slightly negative to it, sorry.

> I've handled the Debian related files in https://github.com/vim/vim/pull/14849.

So please remove the Debian related changes from this PR.

> Unify if you're sure that 4/5 are in the right and &cms is relevant (i.e. fdm=marker vs fdm=syntax).

Sorry, I did not get this one.

Finally, to my own point:

> I also noticed that there are apparently 2 syntax script, that set the commentstring option. I think it sholdn't be set there. This clearly belongs into a ftplugin.

Can you please remove those (or add as a simple ftplugin instead)?

And finally, for my own maintained ftplugin (xml), I'll make a related change to my xml repository. But please keep the change here, I am not planning to PR my changes anytime soon anyhow.

I'll leave this open for a bit, so that others have a chance to comment (e.g. if they like to have it done as part of their regular runtime file updates).

chrisbra:
Ah, our comments crossed:

> A question from myself: should this PR add a small documentation note saying that space-padding is prefered in commentstrings?

Yes, that is a good idea, to mention the intent is to have consistency and therfore the value should be space-padded.

ribru17:
Ok, I have removed the default value change and also removed the changes in debian files :+1: And mentioned the space padding in the docs

clason:
> (or nowadays once you have merged it down to Neovim). 

Sorry about that 😔

> So for now I am slightly negative to it, sorry.

No worries, it's your project, and you do an admirable job maintaining it (if you don't mind me saying that). I just wanted to put the option on the table, so to speak.

vito-c:
LGTM :)

zzzyxwvut:
> > Unify if you're sure that 4/5 are in the right and `&cms` is  
> > relevant (i.e. `fdm=marker` vs `fdm=syntax`).

> Sorry, I did not get this one.

@chrisbra, the fact that a fifth of collected data reveals  
a different preference for `&cms` doesn't say much in the  
matter of idiosyncrasy (or policy), like tabs or spaces for  
indentation.  Besides,  there is no poll to look at for how  
many folks have their `&fdm` set to `marker` for this setting  
to matter at all.

habamax:
> Are you okay with a change to your maintained runtime files?

Sure. Hopefully it wouldn't be overridden in the future.

I mainly was asking about the purpose as it wasn't clear who/what would benefit from it. vim-commentary works(or at least worked) with current commentstring variations just fine, adding paddings where user expects it.

zeertzjq:
This option is also used for the addition of fold markers, which doesn't add whitespace padding automatically.

runtime/ftplugin/abaqus.vim
dkearns: Oops, missed the earlier comment on this.  Thanks.

tpope:
> I mainly was asking about the purpose as it wasn't clear who/what would benefit from it. vim-commentary works(or at least worked) with current commentstring variations just fine, adding paddings where user expects it.

I don't see it mentioned anywhere in the thread, so I'll call out the elephant in the room: The recently released Neovim 0.10 includes built-in `gc` map that was inspired by commentary.vim, but which lacks the whitespace normalization. I'm quite sure that's the underlying motivation for this change.

For markdown.vim, I only omitted spaces because that's what html.vim did. I'm happy so see a unilateral fix for `<!--%s-->` in particular get considered; HTML comments without a space are especially ugly.

For liquid.vim, the tag-like nature of it makes me think it's one of the few where the space makes it look worse? Not a hill I'm going to die on, but I think I'd recommend leaving it alone.

For pdf.vim, I don't even maintain my own repository for that. Change it, please, save me the trouble of sending an update.

runtime/ftplugin/vim.vim
```diff
@@ -51,9 +52,9 @@ setlocal keywordprg=:help
 
 " Comments starts with # in Vim9 script.  We have to guess which one to use.
 if "\n" .. getline(1, 32)->join("\n") =~# '\n\s*vim9\%[script]\>'
-  setlocal commentstring=#%s
+  setlocal commentstring=#\ %s
 else
-  setlocal commentstring=\"%s
+  setlocal commentstring=\"\ %s
```
zeertzjq: Using a space after the `"` in a Vimscript comment may not always work, as mentioned in `:h line-continuation-comment`. Although this may need special treatment anyway, as the line can be indented.

zeertzjq:
It may be necessary to first consider if having 'commentstring' as a string template is insufficient for its use cases. Some languages may have context-dependent comment formats (e.g. HTML, Vim script, JSX, TSX), but 'commentstring' can only specify one comment format.

hcs42:
I'm happy with this change. The folding markers look better this way.

(I am the maintainer of `erlang.vim`. I will update the [vim-erlang](https://github.com/vim-erlang/vim-erlang-runtime) repository after this PR gets merged.)

benknoble:
I will take a detailed look this week, just FYI.

ribru17:
@zeertzjq I agree. It would be nice (but maybe super breaking) if it could take a function as well maybe. The [astro ftplugin](https://github.com/vim/vim/blob/ef73374dc3e4bf8104ba31d5b22517f8028b467a/runtime/ftplugin/astro.vim#L181) uses a `CursorMoved` `autocmd` to get around this, but I think that is not an ideal solution.

ribru17:
> For liquid.vim, the tag-like nature of it makes me think it's one of the few where the space makes it look worse? Not a hill I'm going to die on, but I think I'd recommend leaving it alone.

I think I agree, but it seems like many inline comments for liquid have space padding (around 60% [of the Github examples](https://github.com/search?q=lang%3Aliquid+%2Fcomment+%25%5C%7D+.*+%5C%7B%25+endcomment%2F&type=code))

> For pdf.vim, I don't even maintain my own repository for that. Change it, please, save me the trouble of sending an update.

Sorry, do you mean to revert the PDF change in this PR? pdf.vim is one of the files I changed already :)

tpope:
> > For liquid.vim, the tag-like nature of it makes me think it's one of the few where the space makes it look worse? Not a hill I'm going to die on, but I think I'd recommend leaving it alone.
> 
> I think I agree, but it seems like many inline comments for liquid have space padding (around 60% [of the Github examples](https://github.com/search?q=lang%3Aliquid+%2Fcomment+%25%5C%7D+.*+%5C%7B%25+endcomment%2F&type=code))

60% at least. I retract my position.

> > For pdf.vim, I don't even maintain my own repository for that. Change it, please, save me the trouble of sending an update.
> 
> Sorry, do you mean to revert the PDF change in this PR? pdf.vim is one of the files I changed already :)

I mean keeping the change would be doing me a favor.

For the rest I will update my repos when this PR is merged. The changes won't get overwritten in a future update, and the  advisory comments in those files aren't necessary (but not hurting anything).

runtime/ftplugin/racket.vim
```diff
@@ -5,6 +5,7 @@
 " URL:                  https://github.com/benknoble/vim-racket
 " Last Change:          2022 Aug 29
 "                       2024 Jan 14 by Vim Project (browsefilter)
+"                       2024 May 23 by Riley Bruins <ribru17@gmail.com> ('commentstring')
```
benknoble: [I've made the requested change upstream](https://github.com/benknoble/vim-racket/compare/master~2..master); please correct the header to match:
```
" Vim filetype plugin
" Language:             Racket
" Maintainer:           D. Ben Knoble <ben.knoble+github@gmail.com>
" Previous Maintainer:  Will Langstroth <will@langstroth.com>
" URL:                  https://github.com/benknoble/vim-racket
" Last Change:          2024 May 28
```

(Neither commentstring nor browsefilter updates need mentioned here.)

benknoble:
I see some force-pushes: do you intend to squash the fixups into the original commit before this is merged? (It looks like you're using `commit --fixup`, so `rebase --autosquash` should do the trick.)

ribru17:
Yeah, I was doing fixups to make the intermediate diffs more transparent; I'll squash them at the end if that makes things easier for you all :+1:

nickeb96:
I am also fine with this change

runtime/ftplugin/vim.vim
```diff
@@ -51,7 +52,7 @@ setlocal keywordprg=:help
 
 " Comments starts with # in Vim9 script.  We have to guess which one to use.
 if "\n" .. getline(1, 32)->join("\n") =~# '\n\s*vim9\%[script]\>'
-  setlocal commentstring=#%s
+  setlocal commentstring=#\ %s
 else
   setlocal commentstring=\"%s
```
seblyng, line 57: Should this also have padding?

ribru17: I did add it initially, but @zeertzjq pointed out that sometimes a backslash must immediately follow the `"` character, and having a space after would break things

seblyng: Sorry, I missed that comment. Thought I read through all of them

chrisbra:
Seems we only got approvals here. If you could do me one more favor please?

> EDIT: Should I explicitly give space padding to commentstrings for files that use setl cms&? (Like C)

Yes, let's please also make this consistent here and add explicitly padding. Once this is done, please squash your changes into a single commit please. I am ready to include this then, thanks all!

runtime/ftplugin/xdefaults.vim
```diff
@@ -13,7 +14,7 @@ set cpo&vim
 
 let b:undo_ftplugin = "setl com< cms< inc< fo<"
 
-setlocal comments=s1:/*,mb:*,ex:*/,:! commentstring& inc&
+setlocal comments=s1:/*,mb:*,ex:*/,:! commentstring=/*\ %s\ */ inc&
```
clason: I think the default here was actually incorrect; the correct commentstring is `!\ %s`, see https://github.com/tpope/vim-commentary/issues/85

chrisbra:
alright then. I'll fix this up while merging. Thanks all

clason:
You forgot to include the fix to the `xs` commentstring (see last force-push).

telemachus:
@ribru17 @chrisbra This pull request left out the space [before old-style Vim comments in ftplugin/vim.vim](https://github.com/vim/vim/pull/14843/files#diff-a1538550ababe5d8b83dd63b22b3aa0a71a864270da9fd406350092754d2f726R55). Was that deliberate? (I'm happy to make a PR if it was a mistake.)

zeertzjq:
Yes, please read the view comments

telemachus:
@zeertzjq Thanks (and sorry for the noise). I scanned the main comments, but missed the review comments.

PR #14844, opened 2024-05-24
Not enough tests for getregion() with multibyte chars

zeertzjq:
Problem:  Not enough tests for getregion() with multibyte chars.
Solution: Add a few more tests.


chrisbra:
thanks!

PR #14845, opened 2024-05-25
Fix a Coverity warning, add a test and reduce the calls to clear_tv()

yegappan:


src/eval.c
```diff
@@ -1854,32 +1864,39 @@ get_lval_subscript(
...
+	clear_tv(&var2);
     }
 
+    rc = OK;
+
+done:
     clear_tv(&var1);
-    return p;
+    clear_tv(&var2);
+    return rc == OK ? p : NULL;
```
chrisbra, line 1901: This got me wondering. Is it fine to call `clear_tv()` for `var1` and  `var2` twice? Once on line 1890+1891 and then here on line 1897 and 1898 again when the loop finished?

yegappan: Only simple types like number, string, bool and float can be used to index a List, Dict, Blob, Object or a Class.
Other composite types are not supported as index values.  With the simple types, clear_tv() can be called twice
without causing any reference counting issues.  When encountering an unsupported composite type, this 
function will jump to the end of the function, where the type is cleared only once.

yegappan: Patch 8.0.0352 (https://github.com/vim/vim/commit/f06e5a549f42396be3478ccc1b5f03be64e1173e) is related to this.  In this patch, `var1` is cleared twice.

chrisbra: okay thanks

PR #14849, opened 2024-05-25
Update Debian runtime files

jamessan:
- Add space in template for 'commentstring'
- Add 'comments' and 'commentstring' support to debcontrol
- debversions: Move Ubuntu releases outside of standard support to unsupported


chrisbra:
Thanks!

PR #14851, opened 2024-05-26
getregionpos() inconsistent for partly-selected multibyte char

zeertzjq:
Problem:  getregionpos() behaves inconsistently for a partly-selected
          multibyte char.
Solution: Always use column of the first byte for a partly-selected
          multibyte char.


chrisbra:
thanks!

PR #14852, opened 2024-05-26
runtime(typescriptreact): fix highlighting nested and escaped quotes in string props

### LINKED ISSUE #279: Highlight is broken when a single quote is nested in double quote string
When single quotes are included in string literals with double quotes, they break syntax highlighting.

```tsx
export function Foo({children}: FooProps) {
    return (
        <Table
            title="This is 'test' title"
        >
            {children}
        </Table>
    );
}
```

Screenshot:

image

rhysd:
This PR is upstreamed from the patch https://github.com/HerringtonDarkholme/yats.vim/pull/280.

Fixes: https://github.com/HerringtonDarkholme/yats.vim/issues/279


chrisbra:
thanks!

PR #14854, opened 2024-05-26
fix(completion): reset shown to first when leader delete to empty

glepnir:
Problem:  when leader back to original leader the compl_shown_match need update.
Solution: restore to  first_match if not compl_no_select set to the first_match.cp_next

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1254,6 +1254,14 @@ ins_compl_build_pum(void)
     if (match_at_original_text(compl_shown_match))
 	shown_match_ok = TRUE;
 
+    if (compl_leader != NULL
+	    && STRCMP(compl_leader, compl_orig_text) == 0
+	    && shown_match_ok == FALSE)
+    {
+	compl_shown_match = compl_no_select ? compl_first_match
+					    : compl_first_match->cp_next;
+    }
+
     i = 0;
     compl = compl_first_match;
     do
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -630,14 +630,14 @@ func Test_pum_with_preview_win()
   CheckScreendump
 
   let lines =<< trim END
-      funct Omni_test(findstart, base)
-	if a:findstart
-	  return col(".") - 1
-	endif
-	return [#{word: "one", info: "1info"}, #{word: "two", info: "2info"}, #{word: "three", info: "3info"}]
-      endfunc
-      set omnifunc=Omni_test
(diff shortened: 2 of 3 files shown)


src/insexpand.c
```diff
@@ -1249,6 +1249,12 @@ ins_compl_build_pum(void)
     if (compl_match_array == NULL)
 	return -1;
 
+    if (compl_started && compl_leader != NULL && *compl_leader == NUL)
+    {
+	compl_shown_match = compl_no_select ? compl_first_match
+					    : compl_first_match->cp_next;
+    }
+
```
zeertzjq: It seems that the message at the bottom is not updated.

src/testdir/dumps/Test_pum_with_preview_win.dump
```diff
@@ -1,12 +1,12 @@
-|1+0&#ffffff0|i|n|f|o| @69
+|2+0&#ffffff0|i|n|f|o| @69
 |~+0#4040ff13&| @73
 |~| @73
 |[+1#0000000&|S|c|r|a|t|c|h|]| |[|P|r|e|v|i|e|w|]| @37|1|,|1| @11|A|l@1
-|o+0&&|n|e> @71
-|o+0#0000001#e0e0e08|n|e| @11| +0#4040ff13#ffffff0@59
-|t+0#0000001#ffd7ff255|w|o| @11| +0#4040ff13#ffffff0@59
+|t+0&&|w|o> @71
```
glepnir: emm seems like not correct.. wired i tried with minimal test.vim and build vim i can't reproduce it ?

src/insexpand.c
glepnir: okay .. thanks

pheiduck:
`test.vim` has to be under `testdir` and `Filelist` needed updated.

src/testdir/test_ins_complete.vim
```diff
@@ -630,14 +630,14 @@ func Test_pum_with_preview_win()
...
-      set omnifunc=Omni_test
-      set completeopt+=longest
+    funct Omni_test(findstart, base)
+      if a:findstart
+        return col(".") - 1
+      endif
+      return [#{word: "one", info: "1info"}, #{word: "two", info: "2info"}, #{word: "three", info: "3info"}]
+    endfunc
+    set omnifunc=Omni_test
+    set completeopt+=longest,noselect
```
chrisbra: why are you adding noselect here?

chrisbra:
thanks!

PR #14857, opened 2024-05-27
runtime(lua): optional one expr in unamed function conceal to lambda

A4-Tacks:
optional one expr in unamed function conceal to lambda

```lua
--print(function(x) return x+y end)
print(λx.x+y;)
--print(function(x) return x+y+print() end)
print(λx.x+y+print();)
--print(function(x) return (x+y)+print() end)
print(λx.(x+y)+print();)
--print(function(x) return function(m)return 2 end+print() end)
print(λx.λm.2;+print();)
--print(function(x) return function(m)return 2 end
--    +print() end)
print(λx.λm.2;
    +print();)
```

### DIFF
--- a/runtime/syntax/lua.vim
+++ b/runtime/syntax/lua.vim
@@ -7,6 +7,7 @@
 "               lua_subversion = 0 (for 4.0 or 5.0)
 "                               or 1, 2, 3 (for 5.1, 5.2 or 5.3)
 "               the default is 5.3
+"               lua_unamed_function_to_lambda = 0 or 1
 
 " quit when a syntax file was already loaded
 if exists("b:current_syntax")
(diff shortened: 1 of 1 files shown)


chrisbra:
have you tried to contact the maintainer of the syntax script?

A4-Tacks:
> have you tried to contact the maintainer of the syntax script?

chrisbra:
BTW, while looking at the change I noticed 2 things:
- you may need to describe the suggested change a bit better. Reading this PR description here, it's not entirely obvious what this change is all about
- the whole change seems to be driven by the new configuration variable `lua_unamed_function_to_lambda`. This should be documented at  `:h ft-lua-syntax`.

A4-Tacks:
I understand, thank you very much.

PR #14861, opened 2024-05-27
Fix syntax vim bug (Close #14858)

### LINKED ISSUE #14858: Vim syntax bug
### Steps to reproduce

```
:syntax enable
:edit newfile.vim
:put='echohl ErrorMsg'

" ErrorMsg text above gets concealed

```

### Expected behaviour

The problem is here
https://github.com/vim/vim/blob/ef73374dc3e4bf8104ba31d5b22517f8028b467a/runtime/syntax/vim.vim#L101

`Conceal` is an option and cannot be used as keyword (also mentioned at `:h :syn-keyword`). Therefore, we must switch to syn-match:

```VimL
syn match vimHLGroup contained /\<Conceal\>/
```

As `syntax/vim.vim` is a generated file, `runtime/syntax/generator/gen_syntax_vim.vim` should be fixed instead.

### Version of Vim

9.1.115

### Environment

Windows GVim

### Logs and stack traces

_No response_

h-east:
@chrisbra 
It appears that `runtime/syntax/vim.vim` has not been updated in over a month.

### DIFF
--- a/runtime/syntax/generator/gen_syntax_vim.vim
+++ b/runtime/syntax/generator/gen_syntax_vim.vim
@@ -464,6 +464,10 @@ function! s:parse_vim_hlgroup(li)
 		let item.type = 'both'
 		call add(a:li, copy(item))
 
+		" "Conceal" is an option and cannot be used as keyword, so remove it.
+		" (Separately specified as 'syn match' in vim.vim.base).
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

zeertzjq:
This still doesn't update `runtime/syntax/vim.vim`

chrisbra:
yeah sorry, fixed now

Issue #14853, opened 2024-05-26
Update defaults.vim


saccarosium:
I think is time to update defaults.vim to include some nice to have settings and mapping. After spending some times using neovim and toying around with [vim-sensible][1]. I do really appreciate the better defaults and the less line on my vimrc.

The most important settings to update are:
* Mapping `Y` by default to `y$`: This is simply better in every way because it behaves like the other capital letter commands.
* Enabling `hidden`: At this point I'm 100% sure that everyone wants to edit two separate file without saving first.
* Enabling `autoindent`: I think this make life easier when coding and usually is expected.
* Enabling `autoread`: I also think this is a no brainer since I'm almost certaint that you want to update your buffers if you modify the file externaly.
* Enabling `modeline`: This is used by a lot of file, including the entire vim source code, so having this by default is only helpful even to the user that doesn't know anything about it.

Some quality of life improvements, but optional, settings are:
* Removing `i` from `complete`: As an avid user of `<C-n>` and `<C-p>` I don't like how they hang on c projects to scan all my include files.
* Add `pum` to `wildoptions`: I find it a lot more readable than the default.
* Enabling `smarttab`: Really good while editing.

[1]: https://github.com/tpope/vim-sensible/blob/master/plugin/sensible.vim

chrisbra:
Well, out of those:
- mapping `Y` this is definitly a user preference, I don't think this belongs to `defaults.vim`
- `hidden`, I like it, but you may end up accumulating buffers, when you actually wanted to have them properly closed. A crash or `:qa!` could have unwanted consequences and `:set nohidden` forces you think if you want to write the current buffer and start editing a new one. I think this is definitly an option that a user should set intentionally, not have it set un-intentionally
- `autoindent` I don't think this makes sense because `defaults.vim` already loads filetype specific indent scripts, which is usually the better indent alternative
- `modline` is already the default for Vim, we don't need to touch this one
- `smarttab` I don't have it and I can imagine it may be surprising if a `<bs>` starts deleting more than one space suddenly.

Out of those suggested, the following may make sense
- `autoread` I don't have it and I don't remember have needed it ever, not sure on this one
- `complete` not sure, it may make sense
- `pum` may make sense, but don't know if this will work on Vim tiny (haven't tested it)

saccarosium:
[...]

PR #14862, opened 2024-05-27
Problem: "i" in 'complete' option may slow down Vim

chrisbra:
Problem:  "i" in 'complete' option may slow down Vim
Solution: Change default and remove the "i"

As mentioned in #14853 the "i" in the 'complete' option may cause slow down using <C-N>/<C-P>, because Vim has to scan all include files.

Let's remove it from the default option value and see if anybody complains.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1,4 +1,4 @@
-*options.txt*	For Vim version 9.1.  Last change: 2024 May 19
+*options.txt*	For Vim version 9.1.  Last change: 2024 May 27
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
@@ -2042,7 +2042,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 	'writebackup'	+ on or off	depends on the |+writebackup| feature
 
 						*'complete'* *'cpt'* *E535*
-'complete' 'cpt'	string	(default: ".,w,b,u,t,i")
+'complete' 'cpt'	string	(default: ".,w,b,u,t")
 			local to buffer
 	This option specifies how keyword completion |ins-completion| works
 	when CTRL-P or CTRL-N are used.  It is also used for whole-line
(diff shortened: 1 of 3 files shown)


benknoble:
I'm not sure how this will affect me yet, but I hope I don't end up missing it so much that I have to go back and add `i` myself via ftplugin or autocommand just to make sure it has effect everywhere. (I suppose with C-x C-i I can still get the matches if I need them.)

ghgary:
Shouldn't this be put in defaults.vim instead? I thought that was the place to put changes that "improve" the Vim experience for new users without breaking backwards compatibility.

chrisbra:
I have no idea if `i_CTRL-X_CTRL-N` is affected by the 'complete' option 🤷 

> Shouldn't this be put in defaults.vim instead?

I was thinking to change the "real" default, because:
- I think people affected by the annoying scan of included files are more likely to have a vimrc already (although that is just a guess) and they wouldn't benefit if we change `defaults.vim`
- I wanted everybody to benefit, whether they are using defaults.vim or not

ghgary:
> I was thinking to change the "real" default, because:
> 
>     * I think people affected by the annoying scan of included files are more likely to have a vimrc already (although that is just a guess) and they wouldn't benefit if we change `defaults.vim`
> 
>     * I wanted everybody to benefit, whether they are using defaults.vim or not

I think that argument applies to any change: we want the most people to benefit. As with most options, whether removing "i" is a benefit depends on the user. And one of the strengths of Vim is its attention to backwards compatibility. I think consistency is important, too. One of the things I value about Vim is that I can pull the latest release, build and install it, and it will work the same as the previous release, without me having to troubleshoot my configuration to try to find out why what worked last week isn't working this week.

The base Vim configuration is stable. New changes are made to defaults.vim. I think it's a mistake to add a third tier of changes that somehow don't fit in defaults.vim but are applied to the base Vim. It seems like this is trying to fix or work around the limitations of defaults.vim. It would be better to try to fix those, as is being discussed in the thread on Issue #14853.

Whether or not "i" slows Vim down depends a lot on the context of the file being searched. I used to be on a big project with a slow computer and had to remove "i". Since then, I've been working on smaller projects with a faster computer, have kept "i" and haven't had any problems.

Regards,
Gary

Shane-XB-Qian:
did you plan to remove 't' too? since huge tag was more possible slow down vim :rofl:

chrisbra:
okay, let me close this here. I have another patch that will always load defaults.vim. But of course this is controversial as well :/

PR #14874, opened 2024-05-29
Resolve size_t to int warnings on Windows.

mrdubya:
A couple of warnings in ex_docmd.c have been resolved by modifying their function argument types, followed by some changes in various function call sites.  This also allowed removal of some casts to cope with size_t/int conversion.


mrdubya:
Sorry. First messed up the commit content and then (obviously) missed something that can't cope with returned values not being negative it seems. Working on it ...

mrdubya:
Ok, well that is nice. Removing the unexpected changes in the commit has resolved the test failures. So should be good to be merged if acceptable.

chrisbra:
thanks!

PR #14876, opened 2024-05-30
No test for escaping '<' with shellescape()

zeertzjq:
Problem:  No test for escaping '<' with shellescape().
Solution: Add a test.  Use memcpy() in code to make it easier to
          understand.  Fix a typo.


chrisbra:
thanks!

PR #14878, opened 2024-05-30
feat(completion):  completeopt support fuzzy

glepnir:
Add fuzzy into completeopt

context #14694

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -113,6 +113,7 @@ struct compl_S
 				// cp_flags has CP_FREE_FNAME
     int		cp_flags;	// CP_ values
     int		cp_number;	// sequence number
+    int		cp_score;	// fuzzy match score
 };
 
 // values for cp_flags
@@ -153,6 +154,7 @@ static int	  compl_no_select = FALSE;	// FALSE: select & insert
 						// TRUE: noselect
 static int	  compl_longest = FALSE;	// FALSE: insert full match
 						// TRUE: insert longest prefix
+static int	  compl_fuzzy_match = FALSE;	// True: fuzzy match enabled
 
 // Selected one of the matches.  When FALSE the match was edited or using the
 // longest common string.
@@ -206,6 +208,8 @@ static int	  compl_cont_status = 0;
 
 static int	  compl_opt_refresh_always = FALSE;
 static int	  compl_opt_suppress_empty = FALSE;
+
+static int	  compl_selected_item = -1;
 
 static int ins_compl_add(char_u *str, int len, char_u *fname, char_u **cptext, typval_T *user_data, int cdir, int flags, int adup);
 static void ins_compl_longest_match(compl_T *match);
@@ -1059,12 +1063,15 @@ completeopt_was_set(void)
     compl_no_insert = FALSE;
     compl_no_select = FALSE;
     compl_longest = FALSE;
+    compl_fuzzy_match = FALSE;
     if (strstr((char *)p_cot, "noselect") != NULL)
 	compl_no_select = TRUE;
     if (strstr((char *)p_cot, "noinsert") != NULL)
 	compl_no_insert = TRUE;
     if (strstr((char *)p_cot, "longest") != NULL)
 	compl_longest = TRUE;
+    if (strstr((char *)p_cot, "fuzzy") != NULL)
+	compl_fuzzy_match = TRUE;
 }
 
 
@@ -1213,6 +1220,17 @@ trigger_complete_changed_event(int cur)
 #endif
 
 /*
+ * pumitem qsort compare func
+ */
+    static int
+ins_compl_fuzzy_sort(const void *a, const void *b)
+{
+    const int sa = (*(pumitem_T *)a).pum_score;
+    const int sb = (*(pumitem_T *)b).pum_score;
+    return sa == sb ? 0 : sa < sb ? 1 : -1;
+}
+
+/*
  * Build a popup menu to show the completion matches.
  * Returns the popup menu entry that should be selected. Returns -1 if nothing
  * should be selected.
@@ -1227,6 +1245,7 @@ ins_compl_build_pum(void)
     int		i;
     int		cur = -1;
     int		lead_len = 0;
+    int		max_fuzzy_score = 0;
 
     // Need to build the popup menu list.
     compl_match_arraysize = 0;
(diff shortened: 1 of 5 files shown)


src/insexpand.c
```diff
@@ -1212,6 +1219,17 @@ trigger_complete_changed_event(int cur)
...
 
+/*
+ * pumitem qsort compare func
+ */
+    static int
+ins_compl_fuzzy_sort(const void *a, const void *b)
+{
+  const int sa = (*(pumitem_T *)a).pum_score;
+  const int sb = (*(pumitem_T *)b).pum_score;
+  return sa == sb ? 0 : sa < sb ? 1 : -1;
```
yegappan: Is proper indentation (four spaces) used in this function?

src/insexpand.c
```diff
@@ -1236,9 +1255,17 @@ ins_compl_build_pum(void)
 
     do
     {
+	// when completeopt include fuzzy option and leader is not null or empty
+	// set the cp_score for after compare.
+	if (compl_fuzzy_match == TRUE && compl_leader != NULL && lead_len > 0)
+	{
```
yegappan: The coding convention is to not use curly braces if there is a single statement in the body of a if statement.

src/insexpand.c
```diff
@@ -1236,9 +1255,17 @@ ins_compl_build_pum(void)
 
     do
     {
+	// when completeopt include fuzzy option and leader is not null or empty
+	// set the cp_score for after compare.
+	if (compl_fuzzy_match == TRUE && compl_leader != NULL && lead_len > 0)
```
yegappan: As "compl_fuzzy_match" is a boolean variable, there is no need to compare it against "TRUE".

src/insexpand.c
```diff
@@ -1236,9 +1255,17 @@ ins_compl_build_pum(void)
...
+	{
+	    compl->cp_score = fuzzy_match_str(compl->cp_str, compl_leader);
+	}
+
 	if (!match_at_original_text(compl)
 		&& (compl_leader == NULL
-		    || ins_compl_equal(compl, compl_leader, lead_len)))
+		    || ins_compl_equal(compl, compl_leader, lead_len)
+		    || (compl_fuzzy_match == TRUE && compl->cp_score > 0)))
```
yegappan: No need to compare "compl_fuzzy_match" against TRUE.

src/insexpand.c
```diff
@@ -1267,9 +1294,10 @@ ins_compl_build_pum(void)
     {
 	if (!match_at_original_text(compl)
 		&& (compl_leader == NULL
-		    || ins_compl_equal(compl, compl_leader, lead_len)))
+		    || ins_compl_equal(compl, compl_leader, lead_len)
+		    || (compl_fuzzy_match == TRUE && compl->cp_score > 0)))
```
yegappan: Remove comparison to TRUE.

src/insexpand.c
```diff
@@ -1267,9 +1294,10 @@ ins_compl_build_pum(void)
...
 	{
-	    if (!shown_match_ok)
+	    if (!shown_match_ok && compl_fuzzy_match == FALSE)
```
yegappan: Can you use "!compl_fuzzy_match" instead of comparing against FALSE?

src/insexpand.c
```diff
@@ -4025,6 +4081,24 @@ ins_compl_show_filename(void)
     redraw_cmdline = FALSE;	    // don't overwrite!
 }
 
+    static compl_T *
+find_next_comp_when_fuzzy(void)
```
yegappan: Can you add a test for this function?

yegappan (changes requested):

glepnir:
I tried it briefly. but not sure yet because I don't fully understand some parts of the code ..

glepnir:
@chrisbra PTAL

src/insexpand.c
```diff
@@ -207,6 +209,8 @@ static int	  compl_cont_status = 0;
 static int	  compl_opt_refresh_always = FALSE;
 static int	  compl_opt_suppress_empty = FALSE;
 
+static int pum_selected_item = -1;
```
chrisbra: shouldn't that be rather `compl_selected_item` to align with the other compl items? Also, please align with the other static ints above.

src/testdir/test_ins_complete.vim
```diff
@@ -2451,4 +2451,56 @@ func Test_completefunc_first_call_complete_add()
   bwipe!
 endfunc
 
+func Test_complete_fuzzy_match()
+  function! OnPumChange()
+    let g:event = copy(v:event)
```
chrisbra: this seems unused?

chrisbra:
Thanks, I think it looks reasonable. Can you please also add some documentation at `:h 'completeopt'`?

src/insexpand.c
glepnir: because this variable take from neovim  :)

src/testdir/test_ins_complete.vim
```diff
@@ -2451,4 +2451,55 @@ func Test_completefunc_first_call_complete_add()
   bwipe!
 endfunc
 
+func Test_complete_fuzzy_match()
```
zeertzjq, line 2454:
```suggestion
func Test_complete_fuzzy_match()
  new

```

src/insexpand.c
```diff
@@ -4025,6 +4077,40 @@ ins_compl_show_filename(void)
...
+    if (compl_match_array == NULL ||
+	    (is_forward && compl_selected_item == compl_match_arraysize - 1)
+	    || (is_backward && compl_selected_item == 0))
+	return compl_first_match;
+
+    if (is_forward)
+	target_idx = compl_selected_item + 1;
+    else if (is_backward)
+      target_idx = compl_selected_item == -1 ? compl_match_arraysize - 1
+					    : compl_selected_item - 1;
```
yegappan, line 4099: Can you check the indentation for the lines in this function?  Vim source base uses four spaces for `shiftwidth`.

src/insexpand.c
```diff
@@ -1320,6 +1365,12 @@ ins_compl_build_pum(void)
 	compl = compl->cp_next;
     } while (compl != NULL && !is_first_match(compl));
 
+    if (compl_fuzzy_match && compl_leader != NULL && lead_len > 0)
+    {
+	// sort by the largest score of fuzzy match
+	qsort(compl_match_array, (size_t)compl_match_arraysize, sizeof(pumitem_T), ins_compl_fuzzy_sort);
+    }
```
yegappan, line 1372: Can you remove the curly braces for this single statement if block?

src/insexpand.c
```diff
@@ -4025,6 +4077,40 @@ ins_compl_show_filename(void)
     redraw_cmdline = FALSE;	    // don't overwrite!
 }
 
+    static compl_T *
```
yegappan, line 4080: Can you add a comment above describing the functionality in this function?

src/insexpand.c
```diff
@@ -1285,6 +1311,24 @@ ins_compl_build_pum(void)
 		    shown_compl = compl;
 		cur = i;
 	    }
+	    else if (compl_fuzzy_match)
+	    {
+		if (compl->cp_score > max_fuzzy_score)
```
yegappan, line 1316: Can you add a comment here describing the functionality in this block?

yegappan (changes requested):

chrisbra:
Opps. sorry, I didn't see Yegappans comments here when I started mergin down this. @glepnir can you make the suggested changes in a follow-up patch please?

src/testdir/test_ins_complete.vim
chrisbra: I fixed this locally when merging

src/insexpand.c
chrisbra: I'll fixed that already while merging

glepnir:
will do

PR #14879, opened 2024-05-30
remove the check for libelf in configure.ac and other parts of build system

omegacoleman:
After searching the whole code base, it seems that libelf is never used and bram himself left a TODO for its removal.
This patch removes the relevant part along with the TODO entry


chrisbra:
yeah, that makes sense. Thanks

PR #14880, opened 2024-05-30
runtime(java): Improve the matching of lambda expressions

zzzyxwvut:
- Distinguish some formal parameters.
- Support multi-line definitions.


zzzyxwvut:
The committed new `dumps/` files would not interfere with [the  
proposed solution](https://github.com/vim/vim/issues/14245#issuecomment-2131433317) that succeeds in recovering all missing  
lines in such files and aims to prevent from future mishaps  
of this kind, should that approach be found acceptable and  
PR'd later.

chrisbra:
thank you!

PR #14884, opened 2024-05-31
Resolve more size_t to in warnings on Windows

mrdubya:
Next round of warnings reduction on Windows.

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3383,7 +3383,7 @@ done:
 get_next_include_file_completion(int compl_type)
 {
     find_pattern_in_path(compl_pattern, compl_direction,
-	    compl_patternlen, FALSE, FALSE,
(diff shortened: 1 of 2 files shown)


mrdubya:
Investigating ...

chrisbra:
seems like you reverted back Patch 9.1.0399. It should be int, otherwise the value cannot get negative.

mrdubya:
Yup, you're right. It is that line that is throwing a size_t to int warning on Windows. I'll try something else.

mrdubya:
Rats. Not having a good day. The expression involves three different integer types, and different platforms/compilers complain about different combinations. Sometimes a cast has to be used ... :(

mrdubya:
Ok, that seems to have worked bar one flaky test failing.  Commit of all casts - I feel dirty. Still, Windows builds are warning free ... for now.

chrisbra:
thanks!

Issue #14864, opened 2024-05-27
Left shift is incorrect with vartabstop and shiftwidth=0


ghgary:
### Steps to reproduce

1. Start vim as "vim -N -u NONE -i NONE"
2. :set vts=11,7,5 sw=0 ruler
3. Insert some word on the first line, e.g. "aword<Esc>".
4. Shift right four times, e.g., ">>...".
5. The sequence of virtual column numbers shown by the rulter is 1, 12, 19. 24, 29.
6. Shift left four times, e.g., "<<...".
7. The sequence of virtual column numbers is 29, 24, 19, 14, 7.

### Expected behaviour

Left-shifting should position the first character of the word to the same tabstops as right-shifting does. In this example, the sequence of virtual column numbers should be 29, 24, 19, 12, 1.

### Version of Vim

9.1.445

### Environment

Operating system: Ubuntu 22.04
Terminal: XTerm(389)
TERM: xterm-256color
Shell: bash 5.1.16(1)

### Logs and stack traces

```shell
I have written a patch and a test for this. I'll submit a PR later in the week after I use it for a while.
```

ghgary:
There is also a bug in the way that shift-rounding is handled with 'vartabstop'. I think the solution may be independent of this one, but I'm still working on it. If they are, I'll post a separate issue and PR.

PR #14887, opened 2024-05-31
Fix left-shift with vartabstop issue #14864

ghgary:
"Left shift is incorrect with vartabstop and shiftwidth=0"

The problem was that with 'vartabstop' set and 'shiftwidth' equal 0, left shifts using << were shifting the line to the wrong column.  The tabstop to the right of the first character in the line was being used as the shift amount instead of the tabstop to the left of that first character.

The reason was that the tabstop_at() function always returned the value of the tabstop to the right of the given column and was not accounting for the direction of the shift.

The solution was to make tabstop_at() aware of the direction of the shift and to choose the tabtop accordingly.

A test was added to check this behavior.

### DIFF
--- a/src/indent.c
+++ b/src/indent.c
@@ -346,7 +346,7 @@ tabstop_first(int *ts)
     long
 get_sw_value(buf_T *buf)
 {
-    return get_sw_value_col(buf, 0, 0);
+    return get_sw_value_col(buf, 0, FALSE);
 }
 
 /*
--- a/src/indent.c
+++ b/src/indent.c
@@ -123,14 +123,22 @@ tabstop_padding(colnr_T col, int ts_arg, int *vts)
 
 /*
  * Find the size of the tab that covers a particular column.
+ *
+ * If this is being called as part of a shift operation, col is not the cursor
+ * column but is the column number to the left of the first non-whitespace
+ * character in the line.  If the shift is to the left (left = 1), then return
+ * the size of the tab interval to the left of the column.
  */
     int
-tabstop_at(colnr_T col, int ts, int *vts)
+tabstop_at(colnr_T col, int ts, int *vts, int left)
 {
-    int		tabcount;
-    colnr_T	tabcol = 0;
-    int		t;
-    int		tab_size = 0;
+    int		tabcount;	// Number of tab stops in the list of variable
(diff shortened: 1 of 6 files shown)


src/testdir/test_vartabs.vim
```diff
@@ -454,5 +454,62 @@ func Test_vartabstop_latin1()
...
+  " Shift back to the left margin.
+  norm! <<
+  let expect = "word"
+  call assert_equal(expect, getline(1))
+
+  " Shift again back to the left margin.
+  norm! <<
+  let expect = "word"
+  call assert_equal(expect, getline(1))
+endfunc
```
yegappan, line 514: Can you add a `bw!` at the end of this function?

ghgary: Done.

src/indent.c
```diff
@@ -123,14 +123,22 @@ tabstop_padding(colnr_T col, int ts_arg, int *vts)
 
 /*
  * Find the size of the tab that covers a particular column.
+ *
+ * If this is being called as part of a shift operation, col is not the cursor
+ * column but is the column number to the left of the first non-whitespace
+ * character in the line.  If the shift is to the left (left = 1), then return
```
zeertzjq: Should `left` use `TRUE` and `FALSE` instead of `0` and `1`? `TRUE` and `FALSE` are passed to the `left` argument of `shift_line()`, for example in indent.c:
```c
		shift_line(TRUE, FALSE, 1, TRUE);
```

ghgary: I think you're right. Fixed.

src/indent.c
```diff
@@ -123,14 +123,22 @@ tabstop_padding(colnr_T col, int ts_arg, int *vts)
...
+ * the size of the tab interval to the left of the column.
```
zeertzjq: It should also be `TRUE` instead of `1` in this comment
```suggestion
 * character in the line.  If the shift is to the left (left = TRUE), then
 * return the size of the tab interval to the left of the column.
```

ghgary: Yep. Thanks. Done.

chrisbra:
thanks Gary.

PR #14890, opened 2024-06-01
Refactor to remove STRLEN() calls part 7

basilisk0315:
Refactor drawline.c top reduce the number of calls to STRLEN().

Also, add a NULL check for alloc().


chrisbra:
Thanks. I think this looks fine

PR #14894, opened 2024-06-01
Add Faust filetype

PowerUser64:
This adds the `faust` filetype for [Faust](https://faust.grame.fr/) `.dsp` and `.lib` files.

The `.lib` file extension was marked as being for COBOL files, but I couldn't find any evidence to support `.lib` being an extension that's used by COBOL (see sources below), so I reassigned it to Faust, which does use the `.lib` extension ([example](https://github.com/grame-cncm/faustlibraries/blob/master/basics.lib)). If anyone can find a `.lib` file in COBOL, I'll gladly write a file type detector for it, but I would need to see it in order to know what I'm detecting.

---
Sources proving that COBOL doesn't use `.lib`:

1. [The Wikipedia article on COBOL](https://en.wikipedia.org/wiki/COBOL) does not list `.lib` as an extension
image
2. [Visual Studio's _COBOL Language Support_ extension](https://github.com/eclipse-che4z/che-che4z-lsp-for-cobol?tab=readme-ov-file#cobol-language-support) does not list `.lib` as an extension they detect as a COBOL file.
image
3. [The first 100 search results for `.lib` files on github](https://github.com/search?q=path%3A*.lib&type=code) are all in repos that do not contain any COBOL code 
(verified by hand)

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -1265,6 +1265,56 @@ export def FTtyp()
   setf typst
 enddef
 
+# Detect Microsoft Developer Studio Project files (Makefile) or Faust DSP
+# files.
+export def FTdsp()
+  if exists("g:filetype_dsp")
+    exe "setf " .. g:filetype_dsp
+    return
+  endif
+
+  # Test the filename
+  if expand('%:t') =~ '^[mM]akefile.*$'
+    setf make
+    return
+  endif
+
+  # Test the file contents
+  for line in getline(1, 200)
+    # Chech for comment style
+    if line =~ '^#.*'
+      setf make
+      return
+    endif
+
+    # Check for common lines
+    if line =~ '^.*Microsoft Developer Studio Project File.*$'
+      setf make
+      return
+    endif
+
+    if line =~ '^!MESSAGE This is not a valid makefile\..+$'
+      setf make
+      return
+    endif
+
+    # Check for keywords
+    if line =~ '^!(IF,ELSEIF,ENDIF).*$'
+      setf make
+      return
+    endif
+
+    # Check for common assignments
+    if line =~ '^SOURCE=.*$'
+      setf make
+      return
+    endif
+  endfor
+
+  # Otherwise, assume we have a Faust file
+  setf faust
+enddef
+
 # Set the filetype of a *.v file to Verilog, V or Cog based on the first 200
 # lines.
 export def FTv()
--- a/runtime/doc/filetype.txt
+++ b/runtime/doc/filetype.txt
@@ -149,6 +149,7 @@ variables can be used to overrule the filetype used for certain extensions:
 	*.csh		g:filetype_csh		|ft-csh-syntax|
 	*.dat		g:filetype_dat
 	*.def		g:filetype_def
+	*.dsp		g:filetype_dsp
 	*.f		g:filetype_f		|ft-forth-syntax|
 	*.frm		g:filetype_frm		|ft-form-syntax|
 	*.fs		g:filetype_fs		|ft-forth-syntax|
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -457,7 +457,7 @@ au BufNewFile,BufRead */.cmus/{autosave,rc,command-history,*.theme} setf cmusrc
 au BufNewFile,BufRead */cmus/{rc,*.theme}			setf cmusrc
 
 " Cobol
-au BufNewFile,BufRead *.cbl,*.cob,*.lib	setf cobol
+au BufNewFile,BufRead *.cbl,*.cob	setf cobol
 "   cobol or zope form controller python script? (heuristic)
 au BufNewFile,BufRead *.cpy
 	\ if getline(1) =~ '^##' |
@@ -531,6 +531,10 @@ au BufNewFile,BufRead s6-*                              setf execline
 
 " Fontconfig config files
 au BufNewFile,BufRead fonts.conf			setf xml
+
+" Faust
+au BufNewFile,BufRead *.lib				setf faust
+au BufNewFile,BufRead *.dsp				call dist#ft#FTdsp()
 
 " Libreoffice config files
 au BufNewFile,BufRead *.xcu,*.xlb,*.xlc,*.xba		setf xml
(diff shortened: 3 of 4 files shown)


chrisbra:
Thanks. Another issue: `*.dsp` is already used for Makefiles

PowerUser64:
Yes, I just noticed that - I'll write a detector for it. It seems like checking for `#` characters at the beginnings of lines, no semicolons `;` at the ends of lines, and looking for keywords like `ifeq` and `endif` could be a good way of doing this.

Example `.dsp` in Makefile: https://github.com/alberto-fc/csr-spi-ftdi/blob/21cda76812b916bb8e31097c389fe6b82fcb8512/obj-win32/Makefile.dsp
Example `.dsp` in Faust: https://github.com/magnetophon/lamb/blob/a2151cfd20c9827c92fb6fc092dc367b2b40d450/lamb.dsp

Also, I've never submitted a PR here before. Should this PR touch the same files as [#12363](https://github.com/vim/vim/pull/12363/files)?

ajatkj:
@PowerUser64 I think it is safe to remove .lib from COBOL filetype detector as it is not a standard COBOL extension. Standard types are .cbl and .cob and .cpy for copybooks.

chrisbra:
I think #12363 is a good example, which you can use as a template for you changes. Thanks for working on that.

PowerUser64:
Good news, it turns out that Make doesn't actually use the `.dsp` extension. Rather it's used by a discontinued program called Microsoft Developer Studio, hence the **D**eveloper **S**tudio **P**roject .dsp format. What's nice is these files are auto-generated, and all seem to have the same text at the top, ["# Microsoft Developer Studio Project File"](https://github.com/search?q="Microsoft+Developer+Studio+Project+File"+path%3A*.dsp&type=code). The format of the file is the same as that of Makefile, hence the association. I added in a few extra tests for things like keywords and comments for good measure.

PowerUser64:
Cool, tests pass. By the way, any idea why https://github.com/vim/vim/actions/runs/9557808149/job/26345595245?pr=14894#step:11:15186 failed?

chrisbra:
Thanks, those 2 failing ones are known flaky ones.

chrisbra:
thanks, looks good now.

PowerUser64:
Thanks for the review!

GitMensch:
Just needed to fixed a local installation that use the nonstandard "lib" suffix :-) [all editors and COBOL compilers I know don't use that extension by default].

The only way I could work around that was to use `--cmd` to reserve .lib for COBOL. (That way the faust definition, which does not use `!`, don't override it`. Would there have been a better/cleaner way?)

A file detector would indeed be nice!
_Most_ copybooks would start with several comment lines (1+), but those may look different depending on the dialect.

Side-note: "common" COBOL file types also include "pco" for [Oracle Pro*COBOL](https://docs.oracle.com/cd/A58617_01/server.804/a58232/ch07.htm#9384) and "sqb" for the [DB2 preprocessor](https://www.ibm.com/docs/en/db2/12.1.0?topic=commands-precompile).
Rarely used but also seen: `copy` (and of course: all of the extensions in upper-case as well).

Has anyone contacted Ankit Jain and got a response [the syntax files are kinda outdated]?

GitMensch:
Just needed to fixed a local installation that use the nonstandard "lib" suffix :-) [all editors and COBOL compilers I know don't use that extension by default].

The only way I could work around that was to use `--cmd` to reserve .lib for COBOL. (That way the faust definition, which does not use `!`, don't override it`. Would there have been a better/cleaner way?)

A file detector would indeed be nice!
_Most_ copybooks would start with several comment lines (1+), but those may look different depending on the dialect.

Side-note: "common" COBOL file types also include "pco" for [Oracle Pro*COBOL](https://docs.oracle.com/cd/A58617_01/server.804/a58232/ch07.htm#9384) and "sqb" for the [DB2 preprocessor](https://www.ibm.com/docs/en/db2/12.1.0?topic=commands-precompile).
Rarely used but also seen: `copy` and `cobol` (and of course: all of the extensions in upper-case as well).

Has anyone contacted Ankit Jain and got a response (the syntax files are kinda outdated)?

PowerUser64:
Good to know. Based on research, do you think searching for these words in the first 10 or so lines would detect most cobol files?

- `PERFORM`
- `PROCEDURE`
- `^[*]` (regex)

I've never seen a cobol .lib file before, so I don't know what I would expect to find in one. Feel to adjust.

GitMensch:
`PROCEDURE` would be less likely (there's only one place where you can use it), I'd drop that. ` DIVISION.` is similar (so I'd also not use that). They are likely to be found in COBOL programs, but we search for copybooks only (kind of `#include`, just with a possible complex text replacement (not positional like macros but leading/trailing/complete words).

For comments: the `*` at column 7 (standard, no matter what's in col 1-6) or column 1 (extension); less likely is `*>` (free format - if this is used as "complete line" comment (that's what we look for) it is prefixed by 0 to 11 spaces (I guess that's`^[ }+[*]>`).
For copybooks you will commonly see ` SECTION.` (non-regex), that's a good thing to add.
For "code" included there: ` PERFORM ` ... may be in, but ` IF ` and ` MOVE ` are much more likely.
For "declarations": it is likely best to search for ` PIC `, ` PICTURE `, ` USAGE `. ` COMP[ -.]` ` VALUE ` (anywhere in the line), other parts are position-dependent or "too common" (USAGE and VALUE may apply there, too, not sure).

For all of those words above: case-insensitive is important.


For some "modern style" copybooks you can have a look at
https://github.com/meyfa/CobolCraft/blob/main/src/_copybooks/state/DD-SERVER-PROPERTIES.cpy
https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/tools/cobjapi/AWT/src_cobol/cobjapifn.cpy
https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/tools/cobjapi/AWT/src_cobol/cobjapi.cpy
https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/samples/worldcities/checkfilestatus.cpy
https://sourceforge.net/p/gnucobol/contrib/HEAD/tree/trunk/tools/htm2cob/source/htm2cob8.cpy

Old style:
https://github.com/mridoni/gixsql/blob/main/copy/SQLCA.cpy
https://github.com/bmTas/JRecord/blob/master/examples/Cbl2Csv/src/test/resources/net/sf/JRecord/zData/DTAR020.cbl
https://github.com/Martinfx/Cobol/blob/master/AS400/QCBLLESRC/DUMP_FULL.CBLLE

PR #14906, opened 2024-06-03
runtime(vim): correctly handle ! attribute for augroup/doautocmd/doautoall

chrisbra:
I noticed that
```
:autocmd! FooBar
:augroup! FooBar
```

looks slightly different to how the ! is drawn. While the `!` for the autocmd case is shown in the same color as autocmd, for :augroup is is shown as VimEventList

In addition, the vim syntax seems to accept doautocmd! and doautoall! (with bang attribute) even so this is not valid.

So let's fix all of those and mark the ! as an error for the the :doautocmd/doautoall case.

cc @dkearns


chrisbra:
do you want to take over this one then? Or shall I squash your patch on top of mine and go ahead merging?

chrisbra:
closing as stalled

Issue #14904, opened 2024-06-03
GvimExt should check HKEY_CURRENT_USER before HKEY_LOCAL_MACHINE


dwagner0:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
Currently GVimExt looks in **HKEY_LOCAL_MACHINE** to locate the gvim.exe binary.  Generally, values in **HKEY_CURRENT_USER** can be used to "override" the corresponding setting in **HKEY_LOCAL_MACHINE** on a per-user basis when desired.  For example, **Explorer** loads the extension from **HKEY_LOCAL_MACHINE\SOFTWARE\Classes\\*\shellex\ContextMenuHandlers\gvim**, *unless* **HKEY_CURRENT_USER\SOFTWARE\Classes\\*\shellex\ContextMenuHandlers\gvim** exists, in which case it loads it from there.

**Describe the solution you'd like**
GVimExt should look at **HKEY_CURRENT_USER\Software\Vim\Gvim** to locate **gvim.exe** prior to looking at **HKEY_LOCAL_MACHINE**.

**Describe alternatives you've considered**
None.

**Additional context**
I believe this should be as easy as adding this code to `getGvimName` in  **vim/src/GvimExt/gvimext.cpp**, prior to the check for **HKEY_LOCAL_MACHINE**:

```
 if (RegOpenKeyEx(HKEY_CURRENT_USER, "Software\\Vim\\Gvim", 0,
				       KEY_READ, &keyhandle) == ERROR_SUCCESS)
    {
	hlen = BUFSIZE;
	if (RegQueryValueEx(keyhandle, "path", 0, NULL, (BYTE *)name, &hlen)
							     != ERROR_SUCCESS)
	    name[0] = 0;
	else
	    name[hlen] = 0;
	RegCloseKey(keyhandle);
    }
```

and changing `if (RegOpenKeyEx(HKEY_LOCAL_MACHINE, "Software\\Vim\\Gvim"` to an `else if`.

chrisbra:
can you please create a PR for this then?

dwagner0:
This is my first ever pull request and the first time I've done Windows code in ages, so feedback quite welcome.

PR #14917, opened 2024-06-04
Have GvimExt check HKEY_CURRENT_USER before HKEY_LOCAL_MACHINE

dwagner0:
Proposed fix for [#14904](https://github.com/vim/vim/issues/14904).


k-takata:
I haven't tried it, but the code looks okay.

BTW, it might be better to implement the user-only install mode in the installer to utilize this, though.

chrisbra:
Thanks, the change to gvim.nsi seems however much more involved, if I see this correctly.

PR #14922, opened 2024-06-05
Cannot have buffer-local value for 'completeopt'

### LINKED ISSUE #5487: Can 'completeopt' be made buffer-local?
**Is your feature request related something that is currently hard to do? Please describe.**
`'completeopt'` is a global list option. It contains 3 values describing how to display extra completion 'info': `preview`, `popup` and `popuphidden`. The `popuphidden` value overrides the other 2, and allows plugins to fetch extra completion 'info' asyncronously, and display it in the completion popup window when it is ready, by calling `popup_show()`.

However this poses a problem, as `popuphidden` can _only_ be used with a plugin which calls `popup_show()`. Without the `popup_show()` call, the completion 'info' is never displayed. Plugins in this category are generally filetype-specific, e.g. language server clients.

This means that if a user sets `set completetopt+=popuphidden` so that a plugin can make use of it with `popup_show()`, any completions _not_ provided by that plugin can no longer display 'info' documentation - the information may be there but never displayed with `popup_show()`.

So users who use different completion providers for different languages have to decide whether they _always_ want to use `popuphidden` (in which case some of their languages may no longer have documentation displayed), or _never_.

If `'completeopt'` becomes buffer-local, then it can simply be set in `FileType` autocmds or in `ftplugin` scripts.

The only workaround I can think of is constantly updating `'completeopt'` depending on the filetype in a `BufEnter` autocmd.

zeertzjq:
Problem:  Cannot have buffer-local value for 'completeopt'.
Solution: Make 'completeopt' global-local.

Close #5487

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -2101,7 +2101,7 @@ A jump table for the options with a short description can be found at |Q_op|.
 
 						*'completeopt'* *'cot'*
 'completeopt' 'cot'	string	(default: "menu,preview")
-			global
+			global or local to buffer |global-local|
 	A comma-separated list of options for Insert mode completion
 	|ins-completion|.  The supported values are:
 
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -2429,6 +2429,7 @@ free_buf_options(
     clear_string_option(&buf->b_p_lop);
     clear_string_option(&buf->b_p_cinsd);
     clear_string_option(&buf->b_p_cinw);
+    clear_string_option(&buf->b_p_cot);
     clear_string_option(&buf->b_p_cpt);
 #ifdef FEAT_COMPL_FUNC
     clear_string_option(&buf->b_p_cfu);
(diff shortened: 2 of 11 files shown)


chrisbra:
Hm, that seems like a new flaky test :/ 
```
From test_shortpathname.vim:
	Found errors in Test_ColonEight_MultiByte():
	command line..script D:/a/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_ColonEight_MultiByte line 10: Expected not equal to 'Xtest/日��本��語��の��フ��ァ��イ��ル��.txt'
	command line..script D:/a/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_ColonEight_MultiByte line 11: Pattern '\\~' does not match 'Xtest/日��本��語��の��フ��ァ��イ��ル��.txt'
```

chrisbra:
thanks!

Issue #14920, opened 2024-06-05
Vimscript colorization doesn't support non matching parenthesis in the pattern section of the substitute command


vds2212:
### Steps to reproduce

1. Create a Vimscript with the following content:
 ```
substitute /(/a/
substitute /a/b/
```
2. The first `substitute` is identified as a `vimSubst` where the second is identified as `vimVar`

### Expected behaviour

Both `substitute` call should be identified and colorized as `vimSubst`

Remark: If the pattern contains a closing parenthesis the problem is not present:
```
substitute /()/a/
substitute /a/b/
```

### Version of Vim

9.1.432

### Environment

OS: Windows 10
Vim Flavor: gVim

vds2212:
Thanks for your quick answer.

Indeed if I remove the space between `substitute` and the first `/` the highlighting is correct.

Thanks for your support.

vds2212:
I confirm that it solve the reported problem :-).
Thanks for the support!

PR #14923, opened 2024-06-05
runtime(vim): Update base-syntax, allow whitespace before :substitute pattern

dkearns:
Allow whitespace between the :substitute command and its pattern
argument.  Although unusual, it is supported and there are examples in
the wild.

Match Vi compatible :substitute commands like `:s\/{string}/.`  See :help
E1270.

Fixes issue #14920.


chrisbra:
thanks!

PR #14925, opened 2024-06-05
Inconsistencies between functions for option flags

zeertzjq:
Problem:  Inconsistencies between functions for option flags.
Solution: Consistently use "unsigned int" as return type and rename
          get_bkc_value() to get_bkc_flags().

### DIFF
--- a/src/bufwrite.c
+++ b/src/bufwrite.c
@@ -690,7 +690,7 @@ buf_write(
     int		    write_undo_file = FALSE;
     context_sha256_T sha_ctx;
 #endif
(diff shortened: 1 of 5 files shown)


chrisbra:
Hm, personally I don't like dropping the `int` from `unsigned int`, since the latter is more explicit. It seems this is not used consistently here in the code base (sometimes with the int, sometimes without).

zeertzjq:
Hmm, then I'll drop those changes.

chrisbra:
thanks!

PR #14926, opened 2024-06-05
Change 'cms' for C++ to '// %s'

### LINKED ISSUE #14911: comment package and C++ (and folds...)
When using the _comment_ package in C++, lines are commented with `'commentstring'` which defaults to `/*%s*/`. 

In a ideal world, they should be commented with `// %s`. But given `'commentstring'` is already used for another purpose (i.e. to create marker based folds), I'm not at ease to change the default used so far -- even if it seems to make more sense when I think about it.

**My feature request is:**

Ideally, it would be nice if _comment_ package had a way to read another option that would override `'commentstring'` value for those cases. It could be something like `b:commentstring` / `g:{&ft}_commentstring` (global filetype based overrides could be enough).

**Other open question:**

Shouldn't `'commentstring`' be `// %s` in C++?

**Context:** Vim [v9.1.0465](https://github.com/vim/vim/releases/tag/v9.1.0465)

LucHermitte:
Closes #14911

### DIFF
--- a/runtime/ftplugin/cpp.vim
+++ b/runtime/ftplugin/cpp.vim
@@ -1,7 +1,7 @@
 " Vim filetype plugin file
 " Language:	C++
 " Maintainer:	The Vim Project <https://github.com/vim/vim>
-" Last Change:	2023 Aug 10
+" Last Change:	2024 Jun 06
 " Former Maintainer:	Bram Moolenaar <Bram@vim.org>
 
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. I think it should rather be:

```
let b:undo_ftplugin ..= ' | setl commentstring<'
```

But I'll fix it while merging.

PR #14928, opened 2024-06-06
Fix issue #14927, type mismatch error in ccomplete

### LINKED ISSUE #14927: Type mismatch error in ccomplete#Complete
### Steps to reproduce

1. Start `vim --clean path/to/source.cpp` in a C/C++ project directory that contains a ctags file (e.g. `./tags`)
2. Enter the name of a `struct` variable followed by a dot (e.g. `myobject.`) and press `<c-x><c-o>`.

For me, this produces the error:
```
Error detected while processing function ccomplete#Complete:
line  200:
E1013: Argument 2: type mismatch, expected dict<string> but got dict<any>
```
I suspect that the line number is slightly misleading and that it has to do with the difference in `dict` types between [lines 211 and 213](https://github.com/vim/vim/blob/ae4641d68e2c01ee4847395bdb78a8c9c52394c7/runtime/autoload/ccomplete.vim#L211) in `runtime/autoload/ccomplete.vim`.

### Expected behaviour

Completion should succeed (or fail cleanly).

### Version of Vim

9.1.1-445

### Environment

* Operating system: Arch Linux on WSL
* Terminal: tmux in Windows Terminal
* Value of `$TERM`: `screen-256color`
* Shell: Zsh 5.9

### Logs and stack traces

_No response_

kyllingstad:
Fixes #14927

### DIFF
--- a/runtime/autoload/ccomplete.vim
+++ b/runtime/autoload/ccomplete.vim
@@ -210,7 +210,7 @@ export def Complete(findstart: bool, abase: string): any # {{{1
     # Find the variable in the tags file(s)
     var diclist: list<dict<any>> = taglist('^' .. items[0] .. '$')
       # Remove members, these can't appear without something in front.
-      ->filter((_, v: dict<string>): bool =>
+      ->filter((_, v: dict<any>): bool =>
                 v->has_key('kind') ? v.kind != 'm' : true)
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

PR #14931, opened 2024-06-06
runtime(vim): Update base-syntax, configurable comment string highlighting

dkearns:
Allow highlighting of strings within comments to be disabled by setting
g:vimsyn_comment_strings to false.


runtime/doc/syntax.txt
```diff
@@ -3834,7 +3833,11 @@ By default, g:vimsyn_folding is unset.  Concatenate the indicated characters
 to support folding of multiple syntax constructs; i.e.
```
zzzyxwvut: For instance, _e.g._ ?

dkearns: Yes, I noticed it too but decided to copy the `g:vimsyn_embed` entry rather than change both.  You've convinced me, I'll change it. :)

zzzyxwvut: There is a third one under `:help vimsyn_folding`.

zzzyxwvut: Scratch that, there are only two.

chrisbra:
can you please rebase?

chrisbra:
thanks!

zzzyxwvut:
It seems that the wrong set of the `dumps/` files have been  
committed _and_ merged: [CI fails](https://github.com/vim/vim/actions/runs/9470762069) (see Artifacts) and so does  
local `make test`.

The conflicting changes were introduced in #13936.

PR #14935, opened 2024-06-07
Added definitions of scripts and plugins.

ubaldot:
This is a proposal to enhance discussions around Vim scripting. 
Such proposal attempts at defining **_scripts_** and **_plugins_**.

### DIFF
--- a/runtime/doc/usr_41.txt
+++ b/runtime/doc/usr_41.txt
@@ -1897,7 +1897,7 @@ mapping.  For these commands, any character until the end-of-line or '|' is
 included.  As a consequence of this behavior, you don't always see that
 trailing whitespace is included: >
 
-	map <F4> o#include 
+	map <F4> o#include  
 
 Here it is intended, in other cases it might be accidental.  To spot these
 problems, you can highlight trailing spaces: >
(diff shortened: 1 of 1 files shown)


runtime/doc/usr_41.txt
```diff
@@ -1881,7 +1897,7 @@ mapping.  For these commands, any character until the end-of-line or '|' is
 included.  As a consequence of this behavior, you don't always see that
 trailing whitespace is included: >
 
-	map <F4> o#include  
+	map <F4> o#include
```
zeertzjq: The trailing whitespace here is required.

ubaldot: Auch! I have an autocommand that trim the whitespaces on save.

runtime/doc/usr_41.txt
```diff
@@ -1881,7 +1897,7 @@ mapping.  For these commands, any character until the end-of-line or '|' is
...
```
zeertzjq: Wrong number of trailing spaces. There should be two.

chrisbra:
thanks!

benknoble:
It would probably be a good idea to include something (here or elsewhere) to disambiguate from `:help plugin`, which says

```
Vim's functionality can be extended by adding plugins.  A plugin is nothing
more than a Vim script file that is loaded automatically when Vim starts.  You
can add a plugin very easily by dropping it in your plugin directory.
{not available when Vim was compiled without the |+eval| feature}

There are two types of plugins:

    global plugin: Used for all kinds of files
  filetype plugin: Only used for a specific type of file

```

ubaldot:
Good catch! I'll try fix it in the incoming days.

PR #14936, opened 2024-06-07
Make indentexpr evaluate from older vim scripts

### LINKED ISSUE #14111: vimindent variable is not defined
### Steps to reproduce

1. `vim --clean test.vim`
2. `:set indentexpr?`
this returns `vimindent.Expr()`
3. `:echo vimindent.Expr()`
this returns `E121: Undefined variable: vimindent`

### Expected behaviour

`vimindent` should be defined

This can be temporarily sidestepped by using the old(?) function
`:set indentexpr=GetVimIndent()`
once inside the session. Trying to do it with an autocommand Filetype or modeline will not work, because then `GetVimIndent` won't be defined.

### Version of Vim

9.1.16

### Environment

OS: Debian Sid
terminal: st
$TERM: tmux-256color
shell: 5.2.21

### Logs and stack traces

_No response_

smemsh:
Older scripts have a problem with recent changes to indentexpr for vim9, they will see it as undefined variable E121.  Patch avoids the error in vim8 script as mentioned by @monkoose in https://github.com/vim/vim/issues/14111#issuecomment-1986917946

Fixes #14111

### DIFF
--- a/runtime/indent/vim.vim
+++ b/runtime/indent/vim.vim
@@ -14,9 +14,7 @@ endif
 b:did_indent = true
 b:undo_indent = 'setlocal indentkeys< indentexpr<'
 
-import autoload '../autoload/dist/vimindent.vim'
-
-setlocal indentexpr=vimindent.Expr()
+setlocal indentexpr=dist#vimindent#Expr()
 setlocal indentkeys+==endif,=enddef,=endfu,=endfor,=endwh,=endtry,=endclass,=endinterface,=endenum,=},=else,=cat,=finall,=END,0\\
 execute('setlocal indentkeys+=0=\"\\\ ,0=#\\\ ')
 setlocal indentkeys-=0#


chrisbra:
Hm, I suppose if we make this change, we also don't need the `import` statement anymore. @yegappan @dkearns what is your opinion on this?

smemsh:
ok, removed the import line.

vim-ml:
On Sun, Jun 9, 2024 at 9:52 AM Christian Brabandt ***@***.***>
wrote:

> Hm, I suppose if we make this change, we also don't need the import
> statement anymore. @yegappan <https://github.com/yegappan> @dkearns
> <https://github.com/dkearns> what is your opinion on this?
>
>
>
If there is a need to support using "indentexpr" from a legacy script, then
I don't see an issue with this change.

Regards,
Yegappan

chrisbra:
so you want those changes because? Is this just because you need to manually call the `&indentexpr`?

smemsh:
> so you want those changes because? Is this just because you need to manually call the &indentexpr

the changes are so the plugin doesn't break with E121, it's the vim-stabs mentioned in the issue, which had been working ok until I upgraded a vim8 to 9.1.  I could carry a patch to vim-stabs, or I could carry a patch in vim but if it benefits anyone else and seems like better not to break vim8 scripts, then why should the patch not go in?

these reference exist:
```
 $ grep -sr indentexpr plugin
plugin/stabs.vim:       if &expandtab || !(&autoindent || &indentexpr || &cindent)
plugin/stabs.vim:               if &indentexpr != ''
plugin/stabs.vim:                       sandbox exe 'let inda='.&indentexpr
plugin/stabs.vim:       let checkalign = ( &expandtab || !(&autoindent || &indentexpr || &cindent))
```
don't know how to do anything more complicated with the patch, just cherry-picked the fix mentioned in the linked issue, because it fixed my problem.  hope that it can be fixed one way or another, so older scripts don't get undefined error.

jessepav:
> so you want those changes because? Is this just because you need to manually call the `&indentexpr`?

To chime in with my use case:

I've been running with a patch in my tree that exposes all vim9script `indentexpr`s at the global scope, because I have a mapping that toggles between the standard `indentexpr` and a custom, minimal indentation mode, and it needs to stash away the buffer's current `indentexpr` so it can restore it later.

chrisbra:
Hm, it seems like those autoloaded functions are automatically translated into autoloadable form, in this case like `dist#vimindent#Expr()`. One also cannot use `function()` to refer to those autoloaded vim9 functions, may be we should make this work instead, so that you can save and restore it using `function()`?

smemsh:
@chrisbra is this patch useful in the meantime to fix vim8 users, before a more complete change can be made, or should the PR be closed?

chrisbra:
I have another idea how to fix this, probably doesn't matter, but I wanted to keep the import and didn't want to have to use the autoloaded name. Let's see how this works for you.

smemsh:
> Let's see how this works for you

solves my issue afaict; thanks.

PR #14941, opened 2024-06-08
Fix errors in using the latest termdebug.vim

yegappan:
If you add a `defcompile` statement to the end of the latest termdebug.vim script, there are a few compilation errors.
This PR addresses a few of them.

I also noticed that the commit https://github.com/vim/vim/commit/ca48202b6f46cfb40a0d1d80033a2f3e8cb7b813 changed the
termdebug window layout.  I am used to having the GDB window and the client program window as horizontally split windows on
the left and the vertically split source window on the right.  This makes it easy to switch between the GDB window and the source
window.  After this commit, now all the windows are vertically split (with the GDB window on the left, the client program window
in the middle and the source window on the right).  This makes it difficult to switch between the GDB window and the source
window.  I don't know the reasoning for this change.

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -193,7 +193,7 @@ enddef
 
 def StartDebugCommand(bang: bool, ...args: list<string>)
   # First argument is the command to debug, rest are run arguments.
-  StartDebug_internal({'gdb_args': [args[0]], 'proc_args': args[1:], 'bang': bang})
+  StartDebug_internal({'gdb_args': [args[0]], 'proc_args': args[1 : ], 'bang': bang})
 enddef
 
 
@@ -618,6 +618,27 @@ def SendCommand(cmd: string)
   endif
 enddef
 
+# Interrupt or stop the program
+def StopCommand()
+  if way == 'prompt'
+    PromptInterrupt()
+  else
+    SendCommand('-exec-interrupt')
+  endif
+enddef
+
+# Continue the program
+def ContinueCommand()
+  if way == 'prompt'
+    SendCommand('continue')
+  else
+    # using -exec-continue results in CTRL-C in the gdb window not working,
+    # communicating via commbuf (= use of SendCommand) has the same result
+    SendCommand('-exec-continue')
+    # command Continue  term_sendkeys(gdbbuf, "continue\r")
+  endif
+enddef
+
 # This is global so that a user can create their mappings with this.
 def TermDebugSendCommand(cmd: string)
   if way == 'prompt'
@@ -626,13 +647,13 @@ def TermDebugSendCommand(cmd: string)
     var do_continue = 0
     if !stopped
       do_continue = 1
-      Stop
+      StopCommand()
       sleep 10m
     endif
     # TODO: should we prepend CTRL-U to clear the command?
     term_sendkeys(gdbbuf, cmd .. "\r")
     if do_continue
-      Continue
+      ContinueCommand()
     endif
   endif
 enddef
@@ -1037,17 +1058,8 @@ def InstallCommands()
   command Finish  SendResumingCommand('-exec-finish')
   command -nargs=* Run  Run(<q-args>)
   command -nargs=* Arguments  SendResumingCommand('-exec-arguments ' .. <q-args>)
-
-  if way == 'prompt'
-    command Stop  PromptInterrupt()
-    command Continue  SendCommand('continue')
-  else
-    command Stop  SendCommand('-exec-interrupt')
-    # using -exec-continue results in CTRL-C in the gdb window not working,
-    # communicating via commbuf (= use of SendCommand) has the same result
-    command Continue   SendCommand('-exec-continue')
-    # command Continue  term_sendkeys(gdbbuf, "continue\r")
-  endif
+  command Stop StopCommand()
+  command Continue ContinueCommand()
 
   command -nargs=* Frame  Frame(<q-args>)
   command -count=1 Up  Up(<count>)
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
thanks for your trying to patch-up that!
i've forsee that, without enough varify and experienced author such big change not sure would cost how much pain to fix all.

--
shane.xb.qian

Shane-XB-Qian:
i meant the vim9 convert, 
as for the layout https://github.com/vim/vim/pull/13474#issuecomment-1790687647 had explained it.

vim-ml:
On Sat, Jun 8, 2024 at 9:18 AM Shane-XB-Qian ***@***.***>
wrote:

> i meant the vim9 convert,
> as for the layout #13474 (comment)
> <https://github.com/vim/vim/pull/13474#issuecomment-1790687647> had
> explained it.
>
>
> What was the problem the layout change was trying to address?  The layout
was usable before
this change.  With this change now, it is difficult to switch between the
GDB window and the source
window (because there are other windows in the middle).  Also, the source
code window has
less than 80 characters leading to the wrapping of the source code.  This
change should be
reverted.

Regards,
Yegappan

Shane-XB-Qian:
> it is difficult to switch between the GDB window and the source window (because there are other windows in the middle).

i do not know how or what was leading you to such result, since the layout change looked was about 'Var' and 'Asm' window;
maybe that vim9 convert? i did not upgrade this convert yet (which i forsee there would be some broken since changed too much), so maybe you can varify that convert more if it was ok to do so? i may more trust you did that.

Shane-XB-Qian:
or revert that vim9 convert for termdebug (and actually i recalled somewhere bram or someone expressed wanted to keep termdebug as legacy vim script before, not sure why now it convert to vim9).
after that, if you still found it was `the auto-layout of Var and Asm` make you face that problem, then i am ok you to revert that one too.

vim-ml:
On Sat, Jun 8, 2024 at 9:33 AM Shane-XB-Qian ***@***.***>
wrote:

> it is difficult to switch between the GDB window and the source window
> (because there are other windows in the middle).
>
> i do not know how or what was leading you to such result, since the layout
> change looked was about 'Var' and 'Asm' window;
> maybe that vim9 convert? i did not upgrade this convert yet (which i
> forsee there would be some broken since changed too much), so maybe you can
> varify that convert more if it was ok to do so? i may more trust you did
> that.
>
>
> The change in window layout is not caused by the Vim9 script conversion.
This is caused by
the changes made in
https://github.com/vim/vim/commit/ca48202b6f46cfb40a0d1d80033a2f3e8cb7b813.

Regards,
Yegappan

Shane-XB-Qian:
what other window in the middle you were trying to say?
did you manually opened var/asm window?
if yes, you still also can move that window to bottom or somewhere by manually too.
// if say so, otherwise I did not know but wish you keep to review this vim9 convert of termdebug if not revert this vim9 convert.

--
shane.xb.qian

vim-ml:
On Sat, Jun 8, 2024 at 11:08 AM Shane-XB-Qian ***@***.***>
wrote:

> what other window in the middle you were trying to say?
> did you manually opened var/asm window?
>

No.  I am not using the var/asm windows.  I use the following
termdebug option to
set the window width:

let g:termdebug_config = {}
let g:termdebug_config['wide'] = 164

With the above setting, when I run the ":Termdebug ./vim" command from the
Vim src
directory, I used to get a window layout with two vertically split
windows.  The left window
is horizontally split into two windows with GDB in the top window and Vim
in the bottom
window.  The vertically split right window contains the source.

After this patch, now I get three vertically split windows.  GDB in the
leftmost window,
Vim in the middle window and the source in the rightmost window.

- Yegappan

if yes, you still also can move that window to bottom or somewhere by
> manually too.
> // if say so, otherwise I did not know but wish you keep to review this
> vim9 convert of termdebug if not revert this vim9 convert.
>
>

Shane-XB-Qian:
> > what other window in the middle you were trying to say?
> > did you manually opened var/asm window?
> >
> 
> No.  I am not using the var/asm windows.  I use the following
> termdebug option to
> set the window width:
> 
> let g:termdebug_config = {}
> let g:termdebug_config['wide'] = 164

then i do not know what happened to you, since the auto-layout is only about 'Var' and 'Asm' window,
if you did not manually open it (or enabled the option of auto, but i supposed you would not do so),
it should be not impact your other windows e.g source and gdb window.
// please check the code _before_ https://github.com/vim/vim/pull/14903 that vim9 convert,
// or after that convert, but i did not upgrade or tried to use that _yet_.

-- 
shane.xb.qian

ubaldot:
@yegappan Thanks for the changes, and sorry to have missed several spaces here and there. I like the introduction of `StopCommand()` and `ContinueCommand()`. I am planning to further exploit Vim9 import/export features to refactor the code and I will also try to use some funcrefs (example: switching funcref in a Init() function depending on the value of the variable `way` for example, but I identified several other places). My idea is to try to refactor the architecture in a more readable/easy-to-maintain/flexible way that shall easy e.g. the introduction of `lldb`(I read somewhere wish it) and perhaps DAP:s. Will I succeed? I dunno :)  I'll send a PR once done.

chrisbra:
thanks! May be we need to find out what causes the layout switch again. I though this only happens in the ASM or variable window when this is opened ?

nickspoons:
Christian asked twice in #13474 for screenshots to demonstrate what that PR should do. None were given. Now this discussion here involves a lot of confusion about which extra windows are displayed where with what options ... it would be much simpler if people could start including screenshots when discussing layout changes/issues. A picture paints a thousand words etc.

vim-ml:
On Sat, Jun 8, 2024 at 9:08 PM Shane-XB-Qian ***@***.***>
wrote:

> > > what other window in the middle you were trying to say?
> > > did you manually opened var/asm window?
> > >
> >
> > No. I am not using the var/asm windows. I use the following
> > termdebug option to
> > set the window width:
> >
> > let g:termdebug_config = {}
> > let g:termdebug_config['wide'] = 164
>
> then i do not know what happened to you, since the auto-layout is only
> about 'Var' and 'Asm' window,
> if you did not manually open it (or enabled the option of auto, but i
> supposed you would not do so),
> it should be not impact your other windows e.g source and gdb window.
> // please check the code _before_ https://github.com/vim/vim/pull/14903
> that vim9 convert,
> // or after that convert, but i did not upgrade or tried to use that _yet_.
>
>
>
This turned out to be a bug in the term_start() function.  This function
doesn't clear the vertical
split command modifier correctly.  I have opened a PR to address this.

Regards,
Yegappan

Shane-XB-Qian:
@chrisbra could we keep the legacy version of termdebug, which like coloschema, e.g `foo` vs `foo_256`?
// untill there some confident vim9 version can replace with legacy version, or just keep two version for good,
// if so, whatever vim9 version what changes (#14978 etc) going to be but kept the legacy version stable,
// same like some people like classic native coloschema vs not their refactor ones.

-- 
shane.xb.qian

PR #14945, opened 2024-06-08
runtime(java): Exclude lambda expressions from _when_ _switch-case_ label clauses

zzzyxwvut:
These guard clauses are always boolean expressions, whereas  
lambda expressions can only appear in either an assignment,  
a casting, or an invocation context.

References:
https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-14.11.1  
https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.27


chrisbra:
thanks

PR #14946, opened 2024-06-09
runtime(termdebug): Fix saved_mousemodel check

laburnumT:
Fix issue introduced by 23f29ffc64276dd790581f98b86a0a6435b7eb22 where saved_mousemodel is introduced as a variable, so the exists check will always be true.

Move to check the value of saved_mousemodel instead.


chrisbra:
thanks!

Shane-XB-Qian:
the logic of your assumption maybe incorrect, that var was equal to check `termdebug_popup` was set or not,
so seems using `exists()` was correct.

// but i donot know the logic after that vim9 convert, so just FYI.

-- 
shane.xb.qian

PR #14949, opened 2024-06-09
runtime(svelte): basic svelte ftplugin file

igorlfs:
Hello!

I'm a neovim user, and I was missing commenting support for svelte, so here we are :)

It's my first contributing, so I've used https://github.com/vim/vim/pull/14802 as an inspiration, let me know if I missed anything.


igorlfs:
Any thoughts @towry?

PR #14950, opened 2024-06-09
runtime: add .vim8 as vim filetype

mazunki:
Neovim does not support vim9script, and I am not sure they ever will. Because of this, I have started to use the `.vim8` extension on scripts which keep compatibility with old vimscript.

Would perhaps a more generic `*.vim[0-9]*` be better?

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2583,7 +2583,7 @@ au BufNewFile,BufRead *.tape			setf vhs
 au BufNewFile,BufRead *.hdl,*.vhd,*.vhdl,*.vbe,*.vst,*.vho  setf vhdl
 
 " Vim script
-au BufNewFile,BufRead *.vim,.exrc,_exrc,.netrwhist	setf vim
+au BufNewFile,BufRead *.vim,*.vim8,.exrc,_exrc,.netrwhist	setf vim
 
 " Viminfo file
 au BufNewFile,BufRead .viminfo,_viminfo		setf viminfo


msoyka2024:
I also tried using .vim8 and .vim9 file extensions for this purpose but found that, on Windows, both files would be sourced anyway.  You can verify this in either of two ways: put the files in the "plugin" directory or just invoke Vim's glob() function.

For example, create these files: test.vim8 and test.vim9.  Then :echo glob("*.vim") will return both files.  Apparently this happens because NTFS may record the filenames in 8.3 format as well as in long format.   It appears that the same code is used by Vim during startup when loading plugins.

My solution has been to create three files: one with the vim9script code, another with the traditional vimscript code, and a third, with a .vim extension, that decides which one to source (the first two files should not have a .vim-like extension).

Shane-XB-Qian:
> Neovim does not support vim9script, and I am not sure they ever will. Because of this, I have started to use the `.vim8` extension on scripts which keep compatibility with old vimscript.
> 
> Would perhaps a more generic `*.vim[0-9]*` be better?

um.. that called `legacy` vim script, not vim8 script.
and that's not like python2 vs python3, bram designed those both were supported by vim, tho/but vim9 have more performance and type advantage.
// vs legacy vim script maybe more easy and simple,
// vs some part of vim9 script maybe looked odd.

anyway, welcome to use vim9 script, and you can identify them by 'vim9script' at the top of vim fs.
// tho some classic vim script/plugin maybe better to keep using `legacy` version to avoid chaotic.

-- 
shane.xb.qian

chrisbra:
Perhaps we should use `vi8` and `vi9` then?

mazunki:
Personally, I'm not sure what the benefit of adding `.vi9` would be, unless there's some implication that vim10script is gonna be incompatible with vim9. I would assume a `*.vim` file is likely to use "the latest vim features".

I have no opinion on `.vim8` vs `.vi8` or even `.legacyvim`, as long as its identifiable. Should I rebase with `*.vi8` instead of `*.vim8` ?

Shane-XB-Qian:
> I would assume a `*.vim` file is likely to use "the latest vim features".

the legacy and vim9 script can be mixed to use in one *.vim, there no concept of `latest`,
so the *.vim8 (or *.vi8) looked odd, even you wrote it by yourself and limited it only for legacy.
// though mixture maybe not recommended, and some value/null/null_obj etc maybe odd too if mixed to use.

-- 
shane.xb.qian

chrisbra:
Can you explain, why you need a separate file extension just for legacy vim script?

mazunki:
In short, because Neovim supports legacy vim script, but not vim9script.

A trivial example:
```vim
vim9script
def g:Double(n: number): number
	return 2 * number
enddef
```
image

I use both vim and Neovim (for different purposes), and sourcing all the `.vim`-files of my `~/.config/vim/plugin` used to work fine. Now I need to be more careful about which files I wish to source from Neovim.

By renaming the files I know will work fine to `.vim8` (or similar), I can just glob for the files I know are compatible. I can think of a few other workarounds, but I feel this is the cleanest. I'm happy to hear about alternatives, though.

chrisbra:
I think you may want instead look into `:h vim9-mix`

nickspoons:
I wonder if it makes more sense to propose to the neovim project that `:vim9script` become a synonym for `:finish`? If they have no intention of porting vim9script then this would be a simple way to prevent vim9 scripts from executing past the point where they become unreadable to neovim.

chrisbra:
cc @clason ?

clason:
Honestly, I'm just as baffled as you about the point of this. We already have a way for plugins to check whether they run on Vim or Neovim: `has('nvim')` -- and I yet have to hear about a problem with this.

If Vim doesn't see the need for separate filetypes, we don't either.

chrisbra:
Okay, then let's just not do it. At least not until we really see a need for this. Thanks!

mazunki:
I see. Thanks for considering it though.

I think I may use `!has("vim9script")` option, even if this is mostly for local configration/scripts, and not really for plugins I'm publishing/sharing. I have seen it before, but never really gave it much thought, since it seems like a bit of unnecessary complexity for personal config files. Thanks for the suggestion though.

That said, I will probably still keep my `.vim8` extension around as a personal ftdetect, just because it allows me to avoid polluting the sources files. In terms of making the separation at a directory level, that's certainly an option too.

I made the PR mostly as a QoL request, not because it's really necessary for any functionality. I realize making it a supported extension can, to some degree, also make it an official extension, which may be undesired.

PR #14951, opened 2024-06-09
Fix the mousemodel restoration fix.

errael:
Follow on to #14946
@laburnumT @ubaldot @yegappan 
I'm somewhat obsessed with vim's null handling and saw this issue go by... I didn't test it, but the following shows the problem. You can apply the fix in this PR to the test and see that it behaves.

**Note**: One might expect that `if mm != null` would work, but the
string container is an exception and `isnot` is required. 
See `:help null-anomolies`

Run this, which is like the code without this PR, and see the unexpected results
```
vim9script

var mm = ''

def SaveRestoreMM()
    echom printf("save '%s'", &mousemodel)

    mm = &mousemodel
    &mousemodel = 'popup_setpos'

    if mm != ''
        echom printf("restore '%s'", mm)
        &mousemodel = mm
        mm = null_string
    endif
enddef

SaveRestoreMM()

&mousemodel = ''

SaveRestoreMM()

SaveRestoreMM()

&mousemodel = ''

SaveRestoreMM()
```
Output:
```
save 'extend'
restore 'extend'
save ''
save 'popup_setpos'
restore 'popup_setpos'
save ''
```

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -109,7 +109,7 @@ var winbar_winids = []
 var plus_map_saved = {}
 var minus_map_saved = {}
 var k_map_saved = {}
-var saved_mousemodel = ''
+var saved_mousemodel = null_string
 
 
 # Need either the +terminal feature or +channel and the prompt buffer.
@@ -1220,7 +1220,7 @@ def DeleteCommands()
     win_gotoid(curwinid)
     winbar_winids = []
 
-    if saved_mousemodel != ''
+    if saved_mousemodel isnot null_string
       &mousemodel = saved_mousemodel
       saved_mousemodel = null_string
       try


ubaldot:
Thanks! Could you please apply the same null_string change proposal where it applies? I think evalexpr and few other variables are initialized as '' and then tested against such a value and perhaps some more. 

Generally, when initializing variables I always prefer to be as explicit as possible as you can see, but it seems that in this case it's better to consider null_string.

errael:
> Thanks! Could you please apply the same null_string change proposal where it applies? I think evalexpr and few other variables are initialized as '' and then tested against such a value and that perhaps happens with some more variables.

I've never looked at the `termdebug` code, now that it's in vim9 I can probably make sense of it (never learned legacy).  The biggest issues around your request are familiarity with the code and testing. Where are the options tested?

This PR currently only changes two lines code, and has a test script that is adequately is a good conceptual test for showing the problem and why the trivial fix works; I'd be comfortable if the current PR got merged given that. But going beyond that real tests are required... Not to mention spending some time examining all the code. Some possibilities are

1. open an issue "termdebug problems with saving/restoring string options", close this PR.
   Find out what options need to be addressed. Find out where the current test are. Proceed as needed to a new PR.
2. Merge this as is.
3. Close this, let someone in the know worry about it.
4. Just let the open PR sit (This is my personal least favorite, included only for completeness. This will not happen)

I'll open an issue, and refer it back to this PR. @ubaldot, how extensive do you think the problems are?

And also determine if there are other issues related to string/null handling that should be addressed.

In addition, I'm compelled to ask if there's related `terminal`, not just `termdebug`, issues that should be considered?

errael:
Opened #14957 about the general issue of save/restore in `termdebug`

chrisbra:
Thanks. I am not sure if the other places are real issues or not. But we can make further changes when we see this as necessary. So let me merge this PR here first. 

BTW: I would have simply used `if !empty(saved_mousemodel)` 🤷

errael:
> BTW: I would have simply used `if !empty(saved_mousemodel)` 🤷

In what context? I'd agree that that's the right thing for the original code, but using `empty()` does not fix the problem

```
vim9script

var v1 = null_string
var v2 = ''

def T()
    echo 'v1:' empty(v1)
    echo 'v2:' empty(v2)
enddef
T()
```
They're both empty. The key to dealing with this null anomaly is using `is`/`isnot`.

Could have an additional flag that indicates if there is currently a saved value (I find this solution particularly  distasteful).

ubaldot:
I took a look at if the issue appears to other variables initialized as empty strings and apparently is does not. In the `saved_mousemodel` it make sense to test the instance through `is/isnot`, whereas for the others you have to check the variable values that, differently than `saved_mousemodel` which is initialized at "startup" and "checked" at "tear down", change at runtime. 

@chrisbra Although using `!empty(saved_mousemodel)` is actual nicer than what I did, in this case we really need `is/isnot`.

I agree with @chrisbra to make further changes if needed later on. When things don't work people don't save themselves to open issues ;) 
Good catch anyway! 

P.S. Perhaps it is only me, but I find the handling of `null` and `null_<type>` with the various exceptions (i.e. strings) very convoluted. Is there any way to simplify such a machinery?

errael:
> Perhaps it is only me, but I find the handling of null and null_<type> with the various exceptions (i.e. strings) very convoluted. Is there any way to simplify such a machinery?

#13458

ubaldot:
> #13458

Oh wow! Thanks for the eloquent discussion you linked to me. I also read the `:help null-anomolies` and my brain flipped upside-down. Hopefully such a "null" machinery will be simplified a bit. :(

errael:
Speaking of rabbit holes, read on

> In the saved_mousemodel it make sense to test the instance through is/isnot, whereas for the others you have to check the variable values that, differently than saved_mousemodel which is initialized at "startup" and "checked" at "tear down", change at runtime.

The statement 
> saved_mousemodel which is initialized at "startup"

 is not accurate. saved_mousemodel is **not always saved** at "startup". That's why there's a problem. But my reading might have a parsing error. If the statement should be read something like
```
 for the others you have to check the variable values that
(differently than saved_mousemodel) which is initialized at "startup"
```
then I'd have to think about it some more.

> In general, it seems that using is/isnot is always a good approach when you have to save and restore a variable value at startup/teardown of a script/plugin beacause you look at the very instance of the saving variable and not at the value it takes. Make sense?

I'm not sure what you're saying. Is this only about strings? There might be something in there that makes sense, but I'm not sure what it is ;)

If at startup you always do
```
var saved_value = some_option_or_other_value
```
then at teardown you can usually unconditionally do
```
some_option_or_other_value = saved_value
```
and no checking is required. Unless there are side effects or it is costly to do the restore, then ...

In the `&mousemodel` case, there's no indication of whether or not the variable was saved; so special care is required. In this case it is initialized to something that is not expected to be a valid value for `&mousemodel`. I just realized there's a potential bug in the code I did. What if the user had done
```
&mousemodel = null_string
```
It turns out  that after doing that, reading the value back does not return `null_string`, so it's ok _this time_.

Here's something bullet proof no matter what values and types are being dealt with. 
```
vim9script

class SavedValue
    var value: any
endclass

var user_value = 'whatever of any type or null_<type>'

var saved_value: SavedValue

echo "is saved:" saved_value != null

# maybe the following is done depending on code paths
saved_value = SavedValue.new(user_value)

echo "is saved:" saved_value != null

user_value = 'foo'

# then sometime later can reliably do
if saved_value != null
    user_value = saved_value.value
endif
```
Using a `var saved_value: list<any>`, instead of a class, is probably more efficient, then could tests `! l->isEmpty()`.

errael:
> EDIT: I just figured that something similar may happen with plus_map_saved, minus_map_saved and k_map_saved. In-spite mappings test pass, what if a user change e.g. 'K' mapping during a termdebug sessions? Also, I noticed a useless exists() in DeleteCommands() function.

Could you add this to the issue #14957 ? Otherwise it will get lost in this rabbit hole of a closed issue. (unless you're planning to fix something about that right now:) )

ubaldot:
> Could you add this to the issue #14957 ? Otherwise it will get lost in this rabbit hole of a closed issue. (unless you're planning to fix something about that right now:) )

Done. But I won't fix it right now but not even too late :)

Issue #14912, opened 2024-06-04
can't fuzzy match a single completion candidate


habamax:
### Steps to reproduce

1. run barebone vim with `vim -Nu NONE`
2. `:set completeopt+=fuzzy`
3. add line `xyz`
4. on the next line type `x` and press `CTRL-N`

I would expect it to complete `xyz`. It doesn't happen.

![image](https://github.com/vim/vim/assets/234774/e062dab2-6533-4698-97c4-83c59e583302)

### Expected behaviour

xyz should be completed as withoug `fuzzy`

### Version of Vim

9.1.465

### Environment

debian 12

habamax:
@glepnir fyi

glepnir:
Issues description are not as detailed as we discussed in the commit section. 

1. first does fuzzy should works like `menuone` when only have one match.
2. not work when use normal completion `<C-N>/<C-p>`  reproduce step in #14910 

for question 2 a littile trouble because of `searchit` function based on regex. may need some change..

chrisbra:
Wait, not sure I understand. We added `fuzzy` to `'completeopt'` option. That is about how vim handles the completion *menu* and *list of already existing candidates*. So my understand for `fuzzy` was, when enabled, the reduction of the list of candidates happens by fuzzy matching, but *how* the list of candidates is created is depends on what kind of completion mode was used (e.g. those of `i_CTRL-X` mappings). One may use fuzzy-completion there as well (using a custom omni-completion), but you may as well simply use regex based search of atoms inside the current buffer of make use of the dictionary. 

So given that, I would expect that `completeopt+=fuzzy` doesn't make a difference for
`x<C-N>` since the list of candidates is build using default keyword local completion (even when there is only 1 candidate).

glepnir:
yes . normal completion is based on keyword regex. but I tried to modify `get_next_default_completion`. If I force some conditions to be skipped, it seems feasible to get fuzzy scores for `ptr` and `compl_orig_text`. not sure do we want also fuzzy on this mode

chrisbra:
I made a small change to the doc 2a2c4fffd7e04f74b316209404767f128684a9e1 on how this should work. I hope we can make it work then as documented.

habamax:
It is actually a bit worse. It can't complete at all if `menu` is not in `completeopt`:

[...]

PR #14955, opened 2024-06-10
fix(completion): compl_shown_match not update when fuzzy with complete

glepnir:
Problem: when use fuzzy option and get completion from complete the compl_shown_match not update .

Solution: when no leader and shown match is first match we should set it to the first compl which is first item in compl_match_array.

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1220,6 +1220,7 @@ ins_compl_build_pum(void)
 {
     compl_T     *compl;
     compl_T     *shown_compl = NULL;
+    compl_T     *after_first_compl = NULL;
     int		did_find_shown_match = FALSE;
     int		shown_match_ok = FALSE;
     int		i;
@@ -1296,6 +1297,8 @@ ins_compl_build_pum(void)
 	    }
 	    else if (compl_fuzzy_match)
 	    {
+		if (i == 0)
+		    after_first_compl = compl;
 		// Update the maximum fuzzy score and the shown match
 		// if the current item's score is higher
 		if (compl->cp_score > max_fuzzy_score)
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks, does this fix #14912?

glepnir:
Nope. that pr has two questions: first he want the fuzzy can works like menuone when only one result is matched also show popupmenu event there is no menuone in completeopt.  two is the keyword completion matches is from searchit which use regex.

chrisbra:
Hm, there is a problem with the test. 
I tried to run this:
```
  set completopt=noinsert,noselect,menu,fuzzy,preview,menuone
  func Comp()
    call complete(col('.'), ["fooBaz", "foobar", "foobala"])
    return ''
  endfunc
  exe "normal! i\<C-R>=Comp()\<CR>\<C-Y>"
  call assert_equal('fooBaz', getline('.'))
```

But this does not resultfor me in `fooBaz`. What's worse is, the test does not fail even without the corresponding change in inxexpand.c

So how can we reproduce the problem please?

habamax:
> Nope. that pr has two questions: first he want the fuzzy can works like menuone when only one result is matched also show popupmenu event there is no menuone in completeopt.

I didn't ask to show popupmenu, the expectation is to have a single candidate completed. It shouldn't matter if menuone is enabled or not.

habamax:
Is this PR related to https://github.com/vim/vim/issues/14968 ?

glepnir:
> Hm, there is a problem with the test. I tried to run this:
> 
> ```
>   set completopt=noinsert,noselect,menu,fuzzy,preview,menuone
>   func Comp()
>     call complete(col('.'), ["fooBaz", "foobar", "foobala"])
>     return ''
>   endfunc
>   exe "normal! i\<C-R>=Comp()\<CR>\<C-Y>"
>   call assert_equal('fooBaz', getline('.'))
> ```
> 
> But this does not resultfor me in `fooBaz`. What's worse is, the test does not fail even without the corresponding change in inxexpand.c
> 
> So how can we reproduce the problem please?

because i paste it in a wrong place ..

chrisbra:
thanks. got it now.

PR #14959, opened 2024-06-10
fix: clarify documentation for "v" position

telemachus:
Problem: the previous documentation falsely states that "v" always refers to the start of a visual area.  In fact, the references of "v" and "." complement each other.  If the cursor is at the start of a (characterwise) visual area, then "v" refers to the end of the area.

Solution: be more verbose and explicit about the connection between "." and "v" and also refer to |v_o| which many vim users will be familiar with for visual areas.


chrisbra:
thanks!

chrisbra:
included as of 210b39c2d686

telemachus:
Thank you for being open to the change and for the quick inclusion.

PR #14961, opened 2024-06-11
Remove the line continuation character used in termdebug

yegappan:
In term_start(), clear the WSP_VERT flag only if it is not already set in cmdmod.cmod_split.


chrisbra:
Thanks, but let me split this into 2 seprate commits.

PR #14956, opened 2024-06-10
runtime(kdl): Add syntax, indent and ftplugin files

jiangyinzuo:

PR #14962, opened 2024-06-11
runtime(kdl): use shiftwidth() instead of &tabstop

jiangyinzuo:


jiangyinzuo:
@dkearns  thinks we should use `shiftwidth()` rather than `&tabstop`, so do I. https://github.com/vim/vim/pull/14956#discussion_r1633759322

chrisbra:
thanks!

PR #14963, opened 2024-06-11
filetype: djot files not recognized

arcxio:
https://djot.net
just yanked these from https://github.com/jgm/djot/tree/main/editors/vim

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -666,6 +666,9 @@ au BufNewFile,BufRead *.rul
 	\   setf diva |
 	\ endif
 
+" Djot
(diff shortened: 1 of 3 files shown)


arcxio:
> You've added the syntax file as a ftplugin, can you please move it to `runtime/syntax/djot.vim`?

sorry, fixed

jgm:
> @jgm, are you happy for this to be distributed with the Vim runtime?

Well, these are very crude syntax files that I wrote in the early stages of this project, and haven't maintained even as the syntax has changed.  So I don't think it would be appropriate yet to make it part of vim.

PR #14965, opened 2024-06-11
runtime(vim): Remove orphaned screen dump files

dkearns:
These are no longer generated by the current test input files.


chrisbra:
thanks!

PR #14970, opened 2024-06-12
runtime(doc): deduplicate getpos(), line(), col(), virtcol()

zeertzjq:
Move the main description to getpos() and link to that from other
functions.


chrisbra:
thanks!

PR #14971, opened 2024-06-12
runtime(vim): Update g:vimsyn_comment_string dump file tests

dkearns:
Update g:vimsyn_comment_string dump file tests.

Leading whitespace is no longer matched as part of the line comment as of #13936.


chrisbra:
thanks!

PR #14972, opened 2024-06-12
Use string interpolation instead of string concatenation in termdebug

yegappan:


ubaldot:
@yegappan Thanks! I was thinking to do the same change during the porting to vim9, along with the line continuation and the quoted keys in dictionaries but I didn't for the following reasons, and I wish you can provide answers so I can learn something new today:

1. **String interpolation vs string concatenation:** are there any benefits in using the former?  In my scripts I tend to use string concatenation because variables stand out more clearly (through spacing and syntax highlight). But perhaps there are hidden benefits that I am not aware of. Would you mind to elaborate more?
2. **Line continuation:** sometimes you need `\` and sometimes you don't. It is difficult to remember when they are needed and when they are not. But if you always use you are not wrong either. -> I always use. What I am missing?
3. **Quoted  dictionary keys:** you don't need them in general, but there are special cases when you need. But if you always quote keys you never have problems. -> I always quote dictionary keys. What I am missing?

ubaldot:
> String interpolation makes the code more readable. With string concatenation, there are too many dots and it makes it hard to see the resulting string. In the case of Vim9 script, the number of instructions generated for a string interpolation is less than that of a string concatenation. For example, look at the disassemble output for the following script:
> 
> ```
> vim9script
> 
> def Foo()
>     var s = 'xxx'
>     var t = $'--- {s} ---'
>     t = '---' .. s .. '---'
> enddef
> defcompile
> disassemble Foo
> ```
> 

OK. 


> In a Vim9 script, you don't need to use the line continuation. Bram spent a lot of time implementing the support for this. Line continuation makes the code harder to read.

I am referring to the following excerpt from `:h vim9.txt`:

```
In places where line continuation with a backslash is still needed, such as
splitting up a long Ex command, comments can start with '#\ ': >
	syn region Text
	      \ start='foo'
	      #\ comment
	      \ end='bar'
Like with legacy script '"\ ' is used.  This is also needed when line
continuation is used without a backslash and a line starts with a bar: >
	au CursorHold * echom 'BEFORE bar'
	      #\ some comment
	      | echom 'AFTER bar'
```
 
> In a Vim9 script, you don't need to quote the keys in a Dict. What special conditions are you referring to that needs quoting the keys?

I am referring in case the key contains non-alphanumeric characters. Still from `:h vim9.txt`:

```
This works for alphanumeric characters, underscore and dash.  If you want to
use another character, use a single or double quoted string: >
	var dict = {'key with space': value}
	var dict = {"key\twith\ttabs": value}
	var dict = {'': value}			# empty key
```

chrisbra:
thanks!

PR #14973, opened 2024-06-12
fix: check list pointer before free in fuzzy_match_str_with_pos

glepnir:
Problem:  there is missing pointer check may cause free twice.

Solution: check pointer is not null before free.

cc @chrisbra

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -5198,11 +5198,16 @@ fuzzy_match_str_with_pos(char_u *str UNUSED, char_u *pat UNUSED)
 
 cleanup:
     vim_free(tv_str.vval.v_string);
-    list_free(match_str_list);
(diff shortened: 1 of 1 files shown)


glepnir:
TODO: Subsequent updates should rewrite this function. There are too many list allocation issues. We should implement a new function to replace fuzzy_match_str and fuzzy_match_str_with_pos. The new function can be like this

int fuzzy_match_str(char_u *str, char_u *pat, list_T *match_positions)

chrisbra:
I agree, I thought the same when I saw the many hops you went through just to generate the list of fuzzy positions :) 

I guess it was initially more designed to work with Vim Script types rather than at the lower level directly.

Issue #14921, opened 2024-06-05
Fuzzy-matching support for insert-completion not working (as expected)


vbd:
### Steps to reproduce

Function of patch 9.1.0463

Buffer (not saved, no filetype) contains:
```
one
two
three
```

Typing `o` followed by `CTRL+N` only offers `one` for selection.

`:echo &completeopt` returns `longest,menuone,popup,noinsert,fuzzy`

### Expected behaviour

At least I expected to offer `one` and `two` in the selection.

### Version of Vim

9.1.0465

### Environment

OS: Windows 11
gVim

habamax:
It might be different to the linked issue

PR #14976, opened 2024-06-12
feat(completion): add fuzzycollect

### LINKED ISSUE #14912: can't fuzzy match a single completion candidate

glepnir:
initial attempt we can generate a fuzzy matching regular expression here and give it to searchit. then keyword completion CTRL-N CTRL-P can get a fuzzy result and set into match array.  

Fix #14912

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -4100,8 +4100,7 @@ find_comp_when_fuzzy(void)
     int		is_backward = compl_shows_dir_backward();
     compl_T	*comp = NULL;
 
-    if (compl_match_array == NULL ||
-	    (is_forward && compl_selected_item == compl_match_arraysize - 1)
+    if ((is_forward && compl_selected_item == compl_match_arraysize - 1)
 	    || (is_backward && compl_selected_item == 0))
 	return compl_first_match;
 
@@ -4508,6 +4507,11 @@ ins_compl_use_match(int c)
     static int
 get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 {
+    int		i;
+    int		char_len;
+    size_t	fuzzy_len;
+    char_u	*fuzzy_pattern;
+
     if ((compl_cont_status & CONT_SOL) || ctrl_x_mode_path_defines())
     {
 	if (!compl_status_adding())
@@ -4621,6 +4625,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
     }
 
     compl_patternlen = STRLEN(compl_pattern);
+
+    if ((get_cot_flags() & COT_FUZZYCOLLECT) != 0)
+    {
+	// Adjust size to avoid buffer overflow
+	fuzzy_len = (size_t)compl_length * 5 + 10;
+	// Allocate enough space
+	fuzzy_pattern = alloc(fuzzy_len);
+	if (fuzzy_pattern == NULL)
+	{
+	    compl_patternlen = 0;
+	    return FAIL;
+	}
+	// Use 'very magic' mode for simpler syntax
+	STRCPY(fuzzy_pattern, "\\v");
+	i = 2; // Start from 2 to skip "\\v"
+	while (i < compl_length + 2)
+	{
+	    // Append "\\k*" before each character
+	    STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
+	    // Get length of current multi-byte character
+	    char_len = mb_ptr2len(compl_pattern + i);
+	    // Concatenate the character safely
+	    STRNCAT(fuzzy_pattern, compl_pattern + i, char_len);
+	    // Move to the next character
+	    i += char_len;
+	}
+	// Append "\\k*" at the end to match any characters after the pattern
+	STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
+	vim_free(compl_pattern);
+	compl_pattern = fuzzy_pattern;
+	compl_patternlen = STRLEN(compl_pattern);
+    }
 
     return OK;
 }
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -2586,6 +2586,37 @@ func Test_complete_fuzzy_match()
   call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
   call assert_equal('hello help hero h', getline('.'))
 
+  set completeopt=fuzzycollect
+  call setline(1, ['xyz  yxz  x'])
+  call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('xyz  yxz  xyz', getline('.'))
+  " can fuzzy get yxz when use Ctrl-N twice
+  call setline(1, ['xyz  yxz  x'])
+  call feedkeys("A\<C-X>\<C-N>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('xyz  yxz  yxz', getline('.'))
+
+  call setline(1, ['one two o'])
+  call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('one two one', getline('.'))
+
+  call setline(1, ['你好 你'])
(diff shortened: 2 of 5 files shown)


chrisbra:
I am not sure we should go this way. I thought the idea is to generate the list of candidates normally (without considering fuzzy) and apply the fuzzying only once the completion list has been show to reduce the number of items shown in the popup? So after the completion menu has been shown with all matches initially?

Otherwiese I am not sure it makes sense for completion methods like thesaurus completion, whole line completion, dictionary, spelling or tags completion. In the end, that's why we have different completion methods for and if you want to gather a list of completions using fuzzy matching, one can omni-completion or user-defined completion.

habamax:
It actually makes sense to apply fuzzy initially when gathering completion candidates as well. At least I would expect it to happen. Also https://github.com/vim/vim/issues/14921

glepnir:
removing edit on `get_normal_compl_info` will fix the problem of keyword not being completed when completeopt set fuzzy but doesn't include menu/menuone. 

> I am not sure we should go this way. I thought the idea is to generate the list of candidates normally (without considering fuzzy) and apply the fuzzying only once the completion list has been show to reduce the number of items shown in the popup? So after the completion menu has been shown with all matches initially?

Internally we differentiate between completion behaviors and modes. But for users they think they are all the same, so that's why they want to see fuzzy results in some default completions as well.

glepnir:
I tried to make fuzzy support for original pattern at the end. It works fine. But I don't use some default completions very often. So not sure if it's broken or some patterns don't work. Might need some testing and feedback. Thanks

src/testdir/test_ins_complete.vim
```diff
@@ -2566,6 +2566,15 @@ func Test_complete_fuzzy_match()
   call feedkeys("S\<C-x>\<C-o>fb\<C-n>", 'tx')
   call assert_equal('fooBaz', g:word)
 
+  set completeopt=fuzzy
```
zeertzjq: The 'completeopt' here includes `fuzzy` but not `menu`. Is this intended?

glepnir: yup .when no menu it will complete the word directly before .so i just set it to `fuzzy` to check this behavior..

chrisbra:
alright, convinced :)

src/insexpand.c
```diff
@@ -4606,11 +4611,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 	    STRCPY((char *)compl_pattern, "\\<");
 	    (void)quote_meta(compl_pattern + 2, line + compl_col,
 		    compl_length);
-	}
+        }
+
```
chrisbra: hm, whitespace change?

src/insexpand.c
```diff
@@ -4606,11 +4611,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
     }
 
     compl_patternlen = STRLEN(compl_pattern);
 
+    if (compl_fuzzy_match)
+    {
+	// Adjust size to avoid buffer overflow
+	fuzzy_len = (size_t)compl_length * 4 + 5;
+	// Allocate enough space
+	fuzzy_pattern = alloc(fuzzy_len);
```
zeertzjq, line 4634: Need to check for NULL

src/insexpand.c
```diff
@@ -4606,11 +4611,42 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 	    STRCPY((char *)compl_pattern, "\\<");
 	    (void)quote_meta(compl_pattern + 2, line + compl_col,
 		    compl_length);
-	}
+        }
```
zeertzjq: Why is the Tab changed into spaces here?

techntools:
@chrisbra @glepnir 

Thanks for this feature.

My primary completion method is keyword completion. When I start typing I am fairly certain about first character of the word I want to be completed. I get the candidates starting with that character and fuzzy match the word I want.

Too many candidates will be generated if we generate them as well with fuzzy. Which may slow down the completion. I would suggest generate candidates with fuzzy under another option such as ```fuzzysearch```.

chrisbra:
so maybe we can use a separate sub-option 'fuzzysearch' for the 'completeopt' option then?

techntools:
Yes

Shane-XB-Qian:
> @chrisbra @glepnir 
> 
> Thanks for this feature.
> 
> My primary completion method is keyword completion. When I start typing I am fairly certain about first character of the word I want to be completed. I get the candidates starting with that character and fuzzy match the word I want.
> 
> Too many candidates will be generated if we generate them as well with fuzzy. Which may slow down the completion. I would suggest generate candidates with fuzzy under another option such as ```fuzzysearch```.

maybe just let the 'fuzzy' do the filter, but not the generate, as your (chris) initially said somewhere, where/when after the base inserted, i guess so.

-- 
shane.xb.qian

techntools:
Let the people have it if they want it

Shane-XB-Qian:
> Let the people have it if they want it

ok, if author felt impl it was easy and/or simple, or you would impl it.
// and if was less impact current usage.
// btw: lsp server may offer fuzzy compl items generation already, and knew it would be slow if fuzzy to e.g keyword compl in practise.
// and check my `https://github.com/Shane-XB-Qian/jj` there is a fuzzy filter for a simple and funny fs filter and jump, but only filter.

-- 
shane.xb.qian

glepnir:
Ping me when there is a result and I will implement it

chrisbra:
yeah, let's have another option value for 'completeopt' that determines how the candidates for the completion-menu are collected. It may be called 'fuzzysearch' or 'fuzzycollect' then (not good on names), but it seems good to give people the chance on how they want the completion to work. Thanks!

techntools:
Thanks.

I vote for ```fuzzycollect```.

Shane-XB-Qian:
> yeah, let's have another option value for 'completeopt' that determines how the candidates for the completion-menu are collected. It may be called 'fuzzysearch' or 'fuzzycollect' then (not good on names), but it seems good to give people the chance on how they want the completion to work. Thanks!

ok, so 'fuzzy' of completeopt do only filter, right? seems ok, if so.

src/testdir/test_ins_complete.vim
```diff
@@ -2599,7 +2630,7 @@ func Test_complete_fuzzy_match()
   unlet g:word
 endfunc
 
-" Check that tie breaking is stable for completeopt+=fuzzy (which should
+" Check that tie breaking is stable for completeopt+=fuzzycollect (which should
```
zeertzjq: ?

glepnir: emm this test can be safe remove right ? sry for typo. because now use a new option for keyword completion..

zeertzjq: Why remove? This test is for sorting, not keyword completion.

glepnir: fuzzy not work for this keyword compleiton mode.   It seems a bit redundant. but there is no harm in keeping it

Shane-XB-Qian: um.. you may mis-understand, `fuzzy` of completeopt is to support for (filter) all kinds compl included keyword, vs/but to e.g keyword compl mode maybe not recommend using `fuzzycollect`.

Shane-XB-Qian: um.. you may mis-understand, `fuzzy` of completeopt is to support for (filter) all kinds compl included keyword, vs/but to e.g keyword compl mode maybe not recommend using `fuzzycollect`.

chrisbra:
so this is ready?

glepnir:
Any other questions about the code? @zeertzjq   I'm not sure if this is a result of discussion,  `fuzzycollect` is used for some completion modes that use regular expressions to get matching results right ? feel like the `fuzzyclloect` is a bit long :)

chrisbra:
No complaints? I take that as a general agreement :) Thanks everybody!

Issue #14957, opened 2024-06-10
[termdebug vim9script] problems with saving/restoring options


errael:
### Steps to reproduce

Refer to #14951, in the comments to that PR, it is pointed out that the problem
fixed by the PR may be a general problem in the vim9script port of `termdebug`.

So this issue potentially covers more than that PR. Since the new code
is in operation, I gave it a brief try, there's a chance there may be only
a few isolated problems.

The following script mimics the `termdebug` code. Each call to SaveRestoreMM
should do both a **save** and a **restore**. Run the following and see that
- If the saved value is `''` it is not restored.
- The wrong value ends up in `&mousemodel`.

```vim
vim9script

var mm = ''

def SaveRestoreMM()
    echom printf("save '%s'", &mousemodel)

    mm = &mousemodel
    &mousemodel = 'popup_setpos'

    if mm != ''
        echom printf("restore '%s'", mm)
        &mousemodel = mm
        mm = null_string
    endif
enddef

SaveRestoreMM()

&mousemodel = ''

SaveRestoreMM()

SaveRestoreMM()

&mousemodel = ''

SaveRestoreMM()
```
Output
```
save 'extend'
restore 'extend'
save ''
save 'popup_setpos'
restore 'popup_setpos'
save ''
```

### Expected behaviour

This is a fix, from #14951, for `&mousemodel` option, see `:help null-anomalies` and the PR for more information
```diff
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -112,1 +112,1 @@
-var saved_mousemodel = ''
+var saved_mousemodel = null_string
@@ -1223,1 +1223,1 @@
-    if saved_mousemodel != ''
+    if saved_mousemodel isnot null_string
```

Here is the test code with the fix
```vim
vim9script

var mm = null_string

def SaveRestoreMM()
    echom printf("save '%s'", &mousemodel)

    mm = &mousemodel
    &mousemodel = 'popup_setpos'

    if mm isnot null_string
        echom printf("restore '%s'", mm)
        &mousemodel = mm
        mm = null_string
    endif
enddef

SaveRestoreMM()

&mousemodel = ''

SaveRestoreMM()
```

### Version of Vim

9.1.472

### Environment

linux/gtk

errael:
The PR is merged. @ubaldot  The important remaining stuff for this issue is

[...]

PR #14978, opened 2024-06-12
Added InitScriptVariables() and SanityCheck() functions, added tests and some refactoring.  

ubaldot:
**!!! This is a significant update.**

I apologize if this pull request is too large. I had difficulty breaking down the changes from a version of `termdebug.vim` that I was developing on my own in one of my personal repositories.

Before starting, I copied the changes proposed in https://github.com/vim/vim/pull/14972 that is not still merged. 

**List of Changes and the Reasoning Behind Them:**

1. Introduction of `InitScriptVariables()` Function:

- Reasoning: This function has been introduced to ensure that when you open and close Termdebug, and then open it again, there are no leftover script variable values from the previous session. Leftover values could potentially cause issues. The goal is for each Termdebug session to be independent of previous sessions. At startup, all script variables are initialized. The only exception are `g:termdebug_is_running`, which acts as a guard to prevent multiple sessions from running in parallel and serves other useful purposes and `g:termdebug_loaded` located at the very beginning of the script to prevent sourcing the script twice. The variables are declared at script level and defined in  `InitScriptVariables()`. 

2. Introduction of `SanityCheck()` Function:

- Reasoning: This function has been introduced to ensure that "everything" is in place before starting the debugger. It consolidates various checks that were previously scattered throughout the script, making the functions more readable and maintainable.

3. Enhancements to `CloseBuffer()`:

- Reasoning: The `CloseBuffer()` function has been made more robust to handle edge cases and ensure stability when closing buffers.

4. Refactoring of the "Wait for Startup" Mechanism in `StartDebug_internal()`:

- Reasoning: The mechanism now explicitly sets a timeout as a variable. A new timeout key can be used with the `g:termdebug_config` dictionary to configure this timeout value, enhancing the flexibility and control over the startup process. This addresses some issues that I read a while ago. 

5. More Descriptive Variable Names:

- Reasoning: The names of variables have been made more comprehensive. Almost every Termdebug buffer now has a variable to indicate its name and another variable to indicate its number, improving code readability and maintainability. Due to the latest discussion around the `&mousemodel` option save/restore mechanism, perhaps some other variables shall be prepended with `saved_`.  

6. Consistent Naming for GDB Terminal Buffers:

- Reasoning: The name of the GDB terminal buffer now matches the name of the GDB program being used, e.g., 'gdb', 'mygdb', 'arm-eabi-none-gdb', etc. This ensures clarity and consistency in identifying buffers.

7. Added more tests on top of the existing. 

- Reasoning: I wanted to be sure that nothing breaks.

**Help needed**

There is a test related to `termdebug` who fails. It is about helptags, and I don't know how to fix it. Any guidance is appreciated! 

The other two tests that fail are related to `FAILED: 4: ['vim9_ex_comment_strings', 'vim9_ex_no_comment_strings', 'vim_ex_comment_strings', 'vim_ex_no_comment_strings']` which seems independent of `termdebug`. 

I have few more PR:s to prepare but they should be way smaller.

### DIFF
--- a/runtime/doc/tags
+++ b/runtime/doc/tags
@@ -8007,7 +8007,6 @@ help-buffer-options	helphelp.txt	/*help-buffer-options*
 help-context	help.txt	/*help-context*
 help-curwin	tips.txt	/*help-curwin*
 help-summary	usr_02.txt	/*help-summary*
-help-tags	tags	1
 help-translated	helphelp.txt	/*help-translated*
 help-writing	helphelp.txt	/*help-writing*
 help-xterm-window	helphelp.txt	/*help-xterm-window*
@@ -10593,6 +10592,7 @@ termdebug-mappings	terminal.txt	/*termdebug-mappings*
 termdebug-prompt	terminal.txt	/*termdebug-prompt*
 termdebug-starting	terminal.txt	/*termdebug-starting*
 termdebug-stepping	terminal.txt	/*termdebug-stepping*
+termdebug-timeout	terminal.txt	/*termdebug-timeout*
 termdebug-variables	terminal.txt	/*termdebug-variables*
 termdebug_disasm_window	terminal.txt	/*termdebug_disasm_window*
 termdebug_map_K	terminal.txt	/*termdebug_map_K*
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1282,6 +1282,8 @@ When the debugger ends, typically by typing "quit" in the gdb window, the two
 opened windows are closed.
 
 Only one debugger can be active at a time.
+If your gdb program is slow at startup, you may want to increase the
+|termdebug-timeout| value.
 							*:TermdebugCommand*
 If you want to give specific commands to the command being debugged, you can
 use the `:TermdebugCommand` command followed by the command name and
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1503,6 +1503,21 @@ deprecated global variable names are mentioned for completeness.  If you are
 switching over to using g:termdebug_config you can find the old variable name
 and take over the value, then delete the deprecated variable.
 
+Startup time ~
+						*termdebug-timeout*
+The startup time of GDB programs depends on the number of arguments passed to
+them. To ensure the correct behavior of Termdebug, it is important that the
+startup process of your GDB program completes successfully. However, if the
+startup process of your GDB program gets stuck, it may cause Termdebug to
+freeze.
+
+To prevent this, there is a timeout after which the Termdebug startup process
+will be aborted. If the startup time of your GDB program is long, you may want
+to increase the timeout value. This value is a multiple of a fixed time period
+of 10 ms. For example: >
+	let g:termdebug_config['timeout'] = 500  " 500 * 10 ms = 5000 ms
+
+The default value is set to 300 (=3000ms).
 
 Prompt mode ~
 						*termdebug-prompt*
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -4,7 +4,7 @@ vim9script
 
 # Author: Bram Moolenaar
 # Copyright: Vim license applies, see ":help license"
-# Last Change: 2024 Jun 03
+# Last Change: 2024 Jun 12
 # Converted to Vim9: Ubaldo Tiberi <ubaldo.tiberi@gmail.com>
 
 # WORK IN PROGRESS - The basics works stable, more to come
@@ -48,11 +48,14 @@ if !has('vim9script') ||  v:version < 900
     finish
 endif
 
+# Variables to keep their status among multiple instanced of Termdebug
 # if exists('g:termdebug_loaded')
 #     Echoerr('Termdebug already loaded.')
 #     finish
 # endif
 # g:termdebug_loaded = true
+g:termdebug_is_running = false
+
 
 # The command that starts debugging, e.g. ":Termdebug vim".
 # To end type "quit" in the gdb window.
@@ -132,7 +135,7 @@ var plus_map_saved: dict<any>
 var minus_map_saved: dict<any>
 
 
-def InitScriptVars()
+def InitScriptVariables()
   if exists('g:termdebug_config') && has_key(g:termdebug_config, 'use_prompt')
     way = g:termdebug_config['use_prompt'] ? 'prompt' : 'terminal'
   elseif exists('g:termdebug_use_prompt')
@@ -215,6 +218,8 @@ def InitScriptVars()
 enddef
 
 def SanityCheck(): bool
+  # CHECKME: This is checked after InitScriptVariables(). Perhaps we need a
+  # check also before initialization?
   var gdb_cmd = GetCommand()[0]
   var is_check_ok = true
   # Need either the +terminal feature or +channel and the prompt buffer.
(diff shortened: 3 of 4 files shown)


chrisbra:
Don't worry about those failing test. They come from the syntax test of Vim script highlighting. Apparently we made an error when including updated vim syntax scripts yesterday (I am sure @dkearns is already working on a fix).

You can now rebase your work on top of #14972 since I have just included that.

But those are a lot of changes. Can we please split them into smaller pieces and included them one by one? That makes reviewing easier. I'd suggest to separate your changes into logical smaller units that built up on each other

ubaldot:
@chrisbra ok, I will try to break it down in smaller chunks. Shall I create one PR per change?

chrisbra:
yes, start with 1 PR, wait until it is merged, some time later push the next PR please.

ubaldot:
I close this then.

Shane-XB-Qian:
@chrisbra could we keep the legacy version of termdebug, which like coloschema, `foo` vs `foo_256`?
// untill there some confident vim9 version can replace with legacy version, or just keep two version for good,
// if so, whatever vim9 version what changes going to be but kept the legacy version stable, 
// same like some people like classic native coloschema vs not their refactor ones.

ubaldot:
@Shane-XB-Qian Out of curiosity: does vim9 implies different colorschema? Furthermore, what instabilities have you found in the vim9 version that are not present in the legacy version? Could you please open an issue in the issue tracker?

chrisbra:
> @chrisbra could we keep the legacy version of termdebug, which like coloschema, foo vs foo_256?

Why should we? That does only add burden to have to maintain two different versions. Typically as a user you shouldn't notice the difference when using the termplugin whether it is written in Vim Script or Vim9 Script shouldn't matter for the user experience. 

If for whatever reason people require a legacy version, they can get it from git (e.g. using the [Vim 9.1 release](https://github.com/vim/vim/tree/v9.1.0000/runtime/pack/dist/opt/termdebug/plugin) or the last release before the migration to Vim9 script [v9.1.0470](https://github.com/vim/vim/tree/v9.1.0470/runtime/pack/dist/opt/termdebug/plugin)).

Shane-XB-Qian:
> > @chrisbra could we keep the legacy version of termdebug, which like coloschema, foo vs foo_256?
> 
> Why should we? That does only add burden to have to maintain two different versions.

what the burden would be? this vim9 convert for now just some translation vs/but not adding some new feat.

> Typically as a user you shouldn't notice the difference when using the termplugin whether it is written in Vim Script or Vim9 Script shouldn't matter for the user experience.

it is the problem now, the experience was different or buggy (you should notice some tickets had been opened), so kept the legacy version (at least for now) can keep the game going.

> If for whatever reason people require a legacy version, they can get it from git (e.g. using the [Vim 9.1 release](https://github.com/vim/vim/tree/v9.1.0000/runtime/pack/dist/opt/termdebug/plugin) or the last release before the migration to Vim9 script [v9.1.0470](https://github.com/vim/vim/tree/v9.1.0470/runtime/pack/dist/opt/termdebug/plugin)). 

yea, i heard the same story for the native classic coloschema replacement approach, they were just disappeared in folders but long lived in git history,
but the problem is they are in different user experience, the behavior (or highlight if to coloschema) was different.
besides, if as it planed which termdebug was going to split into some files and folders which using 'vim9 import' methodology, it should be not a simple file 'copy/replace' operation.


anyway, just my hint and friendly reminder, this change did not bother me at all, i can replace it with previous legacy version whatever this vim9 version going.

-- 
shane.xb.qian

chrisbra:
> what the burden would be?

Essential maintain 2 versions of the termdebug plugin.

ubaldot:
> it is the problem now, the experience was different or buggy (you should notice some tickets had been opened), so kept the legacy version (at least for now) can keep the game going.

I already asked, but perhaps you missed my message. So far, there is only https://github.com/vim/vim/issues/14957 opened which, btw, is consequence of a discussion that were on-going. And the identified bug has been already fixed. What other bugs that are not present in the legacy version you have encountered? Could you please some issue in the issue tracker? I think that is the best way of proceeding.

Shane-XB-Qian:
> > what the burden would be?
> 
> Essential maintain 2 versions of the termdebug plugin. 

if the user experience is different or not sure when it to be stable, then IMO it deserved 2 versions, at least for now.
// anyway, just my opinion, since termdebug is an important native and essential vim plugin.

-- 
shane.xb.qian

chrisbra:
> if the user experience is different

But it isn't (or shouldn't).

Shane-XB-Qian:
> > if the user experience is different
> 
> But it isn't (or shouldn't).

ok, let's see
// now the native classic coloschema is different.
// as for this 'termdebug', then i suggest keeping it as a single file which not split to files/folders using 'vim9 import' methodology, if so, user can be a bit easier to replace it with previous legacy version when necessary.

-- 
shane.xb.qian

ubaldot:
> // as for this 'termdebug', then i suggest keeping it as a single file which not split to files/folders using 'vim9 import' methodology, if so, user can be a bit easier to replace it with previous legacy version when necessary.

Why that would be a problem? 
You have a folder called termdebug. Then, you just remove all its content (no matter if it is one or more files) and you copy the old legacy termdebug.vim script there. Then, you call `:Termdebug` exactly as it is was before. Nothing more than that. 

On the contrary, splitting it in multiple files would greatly help the maintenance, possible extentibility (I am thinking about lldb and/or  DAP that someone mentioned) and last, but not least, you don't load a huge script (~2000 LOC at the moment) where in-fact many function are often not used at all - all the many functions that use the `prompt_buffer` when you use a terminal buffer.  Note that termdebug is a plugin where the `autoload` feature could really shine.

Shane-XB-Qian:
> > // as for this 'termdebug', then i suggest keeping it as a single file which not split to files/folders using 'vim9 import' methodology, if so, user can be a bit easier to replace it with previous legacy version when necessary.
> 
> Why that would be a problem? 
> You have a folder called termdebug. Then, you just remove all its content (no matter if it is one or more files) and you copy the old legacy termdebug.vim script there. Then, you call `:Termdebug` exactly as it is was before. Nothing more than that. 
> 
> On the contrary, splitting it in multiple files would greatly help the maintenance, possible extentibility (I am thinking about lldb and/or  DAP that someone mentioned) and last, but not least, you don't load a huge script (~2000 LOC at the moment) where in-fact many function are often not used at all - all the many functions that use the `prompt_buffer` when you use a terminal buffer.  Note that termdebug is a plugin where the `autoload` feature could really shine. 

to termdebug the performance was not the matter, it should be nothing there a matter to performance, but stable was its matter,
so i have no idea why this vim9 convert was necessary to termdebug;
and i recalled there was a similar conversation for it, someone or bram said he'd like to keep it as legacy version, :sweat_smile:

-- 
shane.xb.qian

ubaldot:
I think that someone is working for a solution for us :) 

@Shane-XB-Qian If you want to benefit of the improvements that we are bringing to termdebug during these days but still using the legacy language, you may want to take a look here: https://github.com/jiangyinzuo/neovim/commit/e8e8b29b892fa3c26c21bf025a1ac950cfa774e0

Shane-XB-Qian:
not sure what you are trying to say about that link,
but since now it had been converted to vim9 version and seems now not possible to back out or two versions,
then please keep review your work had done, and specially the issues found but/if no one took it, then fix it timely.

ubaldot:
What I am saying is that the termdebug.vim included in neovim is keeping the old legacy language while including the changes that we are integrating in the Vim9 version, which it seems what you wanted. Take a look at the files included in the commit I linked.

PR #14979, opened 2024-06-12
runtime(java): Include element values in non-marker annotations

zzzyxwvut:
Make a formal definition for normal and single-element kinds  
of annotations that otherwise require for their containment  
to repeat each time all syntax groups that describe element  
values.

Reference:
https://docs.oracle.com/javase/specs/jls/se21/html/jls-9.html#jls-9.7


chrisbra:
thanks!

PR #14980, opened 2024-06-12
Added InitScriptVariables() function, refactor & bugfix. 

ubaldot:
List of Changes and the Reasoning Behind Them:

1. **Introduction of InitScriptVariables() Function:**
  
**Reasoning**: This function has been introduced to ensure that when you open and close Termdebug, and then open it again, there are no leftover script variable values from the previous session. Leftover values could potentially cause issues. The goal is for each Termdebug session to be independent of previous sessions. At startup, all script variables are initialized. The only exception is g:termdebug_loaded located at the very beginning of the script to prevent sourcing the script twice. The variables are declared at script level and defined in `InitScriptVariables()`.

2. **More Descriptive Variable Names:**

**Reasoning:** The names of variables have been made more comprehensive. Almost every Termdebug buffer now has a variable to indicate its name and another variable to indicate its number, improving code readability and maintainability. Due to the latest discussion around the `&mousemodel` option save/restore mechanism, perhaps some other variables shall be prepended with saved_.

3. **Consistent Naming for GDB Terminal Buffers:**

**Reasoning:** The name of the GDB terminal buffer now matches the name of the GDB program being used, e.g., 'gdb', 'mygdb', 'arm-eabi-none-gdb', etc. This ensures clarity and consistency in identifying buffers.

4. **Other:**
Moved `EchoErr()` on top, added another test, some refactoring, mainly changed several `0` and `1` to `true` and  `false`.

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -43,7 +43,7 @@ def Echoerr(msg: string)
 enddef
 
 
-# Variables to keep their status among multiple instanced of Termdebug
+# Variables to keep their status among multiple instances of Termdebug
 # Avoid to source the script twice.
 if exists('g:termdebug_loaded')
     Echoerr('Termdebug already loaded.')
@@ -437,8 +437,8 @@ def StartDebug_term(dict: dict<any>)
   var counter = 0
   var counter_max = 300
   var success = false
-  while success == false && counter < counter_max
-    if IsGdbStarted() == false
+  while !success && counter < counter_max
+    if !IsGdbStarted()
       CloseBuffers()
       return
     endif
@@ -454,7 +454,7 @@ def StartDebug_term(dict: dict<any>)
     sleep 10m
   endwhile
 
-  if success == false
+  if !success
     Echoerr('Failed to startup the gdb program.')
     CloseBuffers()
     return
@@ -475,8 +475,8 @@ def StartDebug_term(dict: dict<any>)
   counter = 0
   counter_max = 300
   success = false
-  while success == false && counter < counter_max
-    if IsGdbStarted() == false
+  while !success && counter < counter_max
+    if !IsGdbStarted()
       return
     endif
 
@@ -509,7 +509,7 @@ def StartDebug_term(dict: dict<any>)
     sleep 10m
   endwhile
 
-  if success == false
+  if !success
     Echoerr('Cannot check if your gdb works, continuing anyway')
     return
   endif
@@ -548,7 +548,7 @@ def StartDebug_prompt(dict: dict<any>)
   prompt_setcallback(promptbuf, function('PromptCallback'))
   prompt_setinterrupt(promptbuf, function('PromptInterrupt'))
 
-  if vvertical == true
+  if vvertical
     # Assuming the source code window will get a signcolumn, use two more
     # columns for that, thus one less for the terminal window.
     exe $":{(&columns / 2 - 1)}wincmd |"
@@ -704,7 +704,7 @@ def TermDebugSendCommand(cmd: string)
     ch_sendraw(gdb_channel, $"{cmd}\n")
   else
     var do_continue = 0
-    if stopped == false
+    if !stopped
       do_continue = 1
       StopCommand()
       sleep 10m
@@ -823,7 +823,7 @@ def DecodeMessage(quotedText: string, literal: bool): string
         #\ \ ->substitute('\\0x00', NullRepl, 'g')
         ->substitute('\\\\', '\', 'g')
         ->substitute(NullRepl, '\\000', 'g')
-  if literal == false
+  if !literal
     return msg
       ->substitute('\\t', "\t", 'g')
       ->substitute('\\n', '', 'g')
@@ -1533,7 +1533,7 @@ def TermDebugBalloonExpr(): string
   if v:beval_winid != sourcewin
     return ''
   endif
-  if stopped == false
+  if !stopped
     # Only evaluate when stopped, otherwise setting a breakpoint using the
     # mouse triggers a balloon.
     return ''
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -42,11 +42,6 @@ def Echoerr(msg: string)
   echohl ErrorMsg | echom $'[termdebug] {msg}' | echohl None
 enddef
 
-if !has('vim9script') ||  v:version < 900
-    # Needs Vim version 9.0 and above
-    Echoerr("You need at least Vim 9.0")
-    finish
-endif
 
 # Variables to keep their status among multiple instanced of Termdebug
 # Avoid to source the script twice.
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -342,7 +342,6 @@ enddef
 
 # IsGdbRunning(): bool may be a better name?
 def IsGdbStarted(): bool
-  # CHECKME: check this implementation
   var gdbproc_status = job_status(term_getjob(gdbbufnr))
   var cmd_name = string(GetCommand()[0])
   if gdbproc_status !=# 'run'
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -50,11 +50,11 @@ endif
 
 # Variables to keep their status among multiple instanced of Termdebug
 # Avoid to source the script twice.
-# if exists('g:termdebug_loaded')
-#     Echoerr('Termdebug already loaded.')
-#     finish
-# endif
-# g:termdebug_loaded = true
+if exists('g:termdebug_loaded')
+    Echoerr('Termdebug already loaded.')
+    finish
+endif
+g:termdebug_loaded = true
 
 # Script variables declaration. These variables are re-initialized at every
 # Termdebug instance
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -4,7 +4,7 @@ vim9script
 
 # Author: Bram Moolenaar
 # Copyright: Vim license applies, see ":help license"
-# Last Change: 2024 Jun 03
+# Last Change: 2024 Jun 12
 # Converted to Vim9: Ubaldo Tiberi <ubaldo.tiberi@gmail.com>
 
 # WORK IN PROGRESS - The basics works stable, more to come
@@ -38,44 +38,61 @@ vim9script
 # The communication with gdb uses GDB/MI.  See:
 # https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI.html
 
-# In case this gets sourced twice.
-if exists('g:termdebug_loaded')
-  finish
+def Echoerr(msg: string)
+  echohl ErrorMsg | echom $'[termdebug] {msg}' | echohl None
+enddef
+
+if !has('vim9script') ||  v:version < 900
+    # Needs Vim version 9.0 and above
+    Echoerr("You need at least Vim 9.0")
+    finish
 endif
-g:termdebug_loaded = true
 
-var way = 'terminal'
-var err = 'no errors'
+# Variables to keep their status among multiple instanced of Termdebug
+# Avoid to source the script twice.
+# if exists('g:termdebug_loaded')
+#     Echoerr('Termdebug already loaded.')
+#     finish
+# endif
+# g:termdebug_loaded = true
 
-var pc_id = 12
-var asm_id = 13
-var break_id = 14  # breakpoint number is added to this
-var stopped = 1
-var running = 0
+# Script variables declaration. These variables are re-initialized at every
+# Termdebug instance
+var way: string
+var err: string
 
-var parsing_disasm_msg = 0
-var asm_lines = []
-var asm_addr = ''
+var pc_id: number
+var asm_id: number
+var break_id: number
+var stopped: bool
+var running: bool
+
+var parsing_disasm_msg: number
+var asm_lines: list<string>
+var asm_addr: string
 
 # These shall be constants but cannot be initialized here
 # They indicate the buffer numbers of the main buffers used
-var gdbbuf = 0
-var varbuf = 0
-var asmbuf = 0
-var promptbuf = 0
+var gdbbufnr: number
+var gdbbufname: string
+var varbufnr: number
+var varbufname: string
+var asmbufnr: number
+var asmbufname: string
+var promptbuf: number
 # This is for the "debugged program" thing
-var ptybuf = 0
-var commbuf = 0
+var ptybufnr: number
+var commbufnr: number
 
-var gdbjob = null_job
-var gdb_channel = null_channel
+var gdbjob: job
+var gdb_channel: channel
 # These changes because they relate to windows
-var pid = 0
-var gdbwin = 0
-var varwin = 0
-var asmwin = 0
-var ptywin = 0
-var sourcewin = 0
+var pid: number
+var gdbwin: number
+var varwin: number
+var asmwin: number
+var ptywin: number
+var sourcewin: number
 
 # Contains breakpoints that have been placed, key is a string with the GDB
 # breakpoint number.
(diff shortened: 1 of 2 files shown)


errael:
Is it worth mentioning specifically what bug is fixed by this PR?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -38,44 +38,61 @@ vim9script
 # The communication with gdb uses GDB/MI.  See:
 # https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI.html
 
-# In case this gets sourced twice.
+def Echoerr(msg: string)
+  echohl ErrorMsg | echom $'[termdebug] {msg}' | echohl None
+enddef
+
+if !has('vim9script') ||  v:version < 900
```
errael:
```suggestion
if v:versionlong < 9010450
```
Given that termdebug ships with vim, could either tighten up the version check as shown, or take it out entirely. vim9script has evolved considerably since v9.0 was released; so unless you're planning to test against it...

BTW, if not `vim9script` this script would have already aborted.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -38,44 +38,61 @@ vim9script
...
+    # Needs Vim version 9.0 and above
+    Echoerr("You need at least Vim 9.0")
+    finish
+endif
+
+# Variables to keep their status among multiple instanced of Termdebug
+# Avoid to source the script twice.
 if exists('g:termdebug_loaded')
-  finish
+    Echoerr('Termdebug already loaded.')
```
chrisbra, line 49: typicaly we don't error out, if a script is sourced twice. We just silently ignore this.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -272,45 +333,43 @@ enddef
...
-def CheckGdbRunning(): string
-  var gdbproc = term_getjob(gdbbuf)
-  var gdbproc_status = 'unknown'
-  if type(gdbproc) == v:t_job
-    gdbproc_status = job_status(gdbproc)
-  endif
-  if gdbproc == v:null || gdbproc_status !=# 'run'
-    var cmd_name = string(GetCommand()[0])
+def IsGdbStarted(): bool
+  # CHECKME: check this implementation
```
chrisbra: CHECKME

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -38,44 +38,61 @@ vim9script
...
+
+if !has('vim9script') ||  v:version < 900
+    # Needs Vim version 9.0 and above
+    Echoerr("You need at least Vim 9.0")
+    finish
+endif
+
+# Variables to keep their status among multiple instanced of Termdebug
+# Avoid to source the script twice.
 if exists('g:termdebug_loaded')
```
errael, line 48:
```suggestion
if g:->get('termdebug_loaded', false)
```

errael (commented):
So much stuff I'm not familiar with... And looks so much better.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -272,45 +333,43 @@ enddef
...
-    gdbproc_status = job_status(gdbproc)
-  endif
-  if gdbproc == v:null || gdbproc_status !=# 'run'
-    var cmd_name = string(GetCommand()[0])
+def IsGdbStarted(): bool
+  # CHECKME: check this implementation
+  var gdbproc_status = job_status(term_getjob(gdbbufnr))
+  var cmd_name = string(GetCommand()[0])
+  if gdbproc_status !=# 'run'
     Echoerr($'{cmd_name} exited unexpectedly')
```
errael, line 348: Seems no reason to compute `cmd_name` outside of the `if` and might as well have it directly as interpolated string
```vim
Echoerr($'{string(GetCommand()[0])} exited unexpectedly')
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -272,45 +333,43 @@ enddef
...
-  var gdbproc = term_getjob(gdbbuf)
-  var gdbproc_status = 'unknown'
-  if type(gdbproc) == v:t_job
-    gdbproc_status = job_status(gdbproc)
-  endif
-  if gdbproc == v:null || gdbproc_status !=# 'run'
-    var cmd_name = string(GetCommand()[0])
+def IsGdbStarted(): bool
+  # CHECKME: check this implementation
+  var gdbproc_status = job_status(term_getjob(gdbbufnr))
```
errael, line 345: Is the job guaranteed to have been started? 
If not, need an extra test
```
var gdbproc = term_getjob(gdbbufnr)
var gdbproc_status = gdbproc == null ? '' : job_status(gdbproc)
if gdbproc_status !=# 'run'
```

because `job_status` does not accept a `null`, it does accept a `null_job`; another expression of problematic null handling.
```
job_status(null_job)   # WORKS
job_status(null)       # FAIL E1218: Job Required for argument 1
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: True! I added it because when manually debugging one has to source the script many times and it often happens that the updated script is not sourced due to this guard and the developer may get mad to try to understand why a certain bug never go away. Adding this check considerably helped me. However, if you really want to remove it, no problems in removing it!  Just let me know :)

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Ooooops! :D 

Shall we keep a check that always return `true`? Naaaaaa! 

Good catch!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Hum, could you please elaborate why that is different? To be the devil's advocate I would say that the former is more readable :p 

Note that I also added a Echoerr message inside the if exist block (see above).

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: It's just for a matter of readability.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I remember that I tested it manually and I ended up with that convoluted implementation. That is why I added the CHECKME comment. If you have time, could you double check, please? Four eyes are better than two sometimes. 

I saw that you opened: https://github.com/vim/vim/pull/14983 
Feel free to adjust this function based on such a change.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: ?

ubaldot:
> Is it worth mentioning specifically what bug is fixed by this PR?

Line 1770: there where a **le**f**t**over from legacy Vim that produce an error: `let v:swapchoice = 'o'` is now `v:swapchoice = 'o'`.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: This is sensitive to the value of `termdebug_loaded`, whereas `exists()` is not. Either way works as long as you don't care about the value. I'd say the latter makes more sense:)

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: Mostly the comment was about putting
```
var cmd_name = string(GetCommand()[0])
```
inside the `if !=# 'run'`. There's no reason to run the code unless there's a problem. Putting the expr in string interpolate is just a matter of style. (and saving a few cycles, but I tend to way over optimize)

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: > I saw that you opened: https://github.com/vim/vim/pull/14983

I closed that in favor of #14984 which is simpler and seems to make more sense.

With either of those fixes,
```vim
var gdbproc_status = job_status(term_getjob(gdbbufnr))
```
works whether or not `gdbbufnr` has a job associated with it.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -38,44 +38,56 @@ vim9script
 # The communication with gdb uses GDB/MI.  See:
 # https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI.html
 
-# In case this gets sourced twice.
+def Echoerr(msg: string)
+  echohl ErrorMsg | echom $'[termdebug] {msg}' | echohl None
+enddef
+
+
+# Variables to keep their status among multiple instanced of Termdebug
```
yegappan: Typo: "instanced" -> "instances"

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -491,7 +548,7 @@ def StartDebug_prompt(dict: dict<any>)
   prompt_setcallback(promptbuf, function('PromptCallback'))
   prompt_setinterrupt(promptbuf, function('PromptInterrupt'))
 
-  if vvertical
+  if vvertical == true
```
yegappan: When using a Boolean variable, there is no need to it compare against `true` or `false`.  The previous code is more readable.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -381,13 +438,13 @@ def StartDebug_term(dict: dict<any>)
   var counter_max = 300
   var success = false
   while success == false && counter < counter_max
-    if CheckGdbRunning() != 'ok'
-      # Failure. If NOK just return.
+    if IsGdbStarted() == false
```
yegappan: When using a Boolean variable, there is no need to it compare against true or false. Can you use `!IsGdbStarted()`?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -406,30 +463,30 @@ def StartDebug_term(dict: dict<any>)
...
+  term_sendkeys(gdbbufnr, $"server new-ui mi {commpty}\r")
 
   # Wait for the response to show up, users may not notice the error and wonder
   # why the debugger doesn't work.
   counter = 0
   counter_max = 300
   success = false
   while success == false && counter < counter_max
-    if CheckGdbRunning() != 'ok'
+    if IsGdbStarted() == false
```
yegappan: When using a Boolean variable, there is no need to it compare against true or false. Can you use `!IsGdbStarted()`?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -646,13 +704,13 @@ def TermDebugSendCommand(cmd: string)
     ch_sendraw(gdb_channel, $"{cmd}\n")
   else
     var do_continue = 0
-    if !stopped
+    if stopped == false
```
yegappan: When using a Boolean variable, there is no need to it compare against true or false. The previous code is more readable.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -751,20 +809,21 @@ def DecodeMessage(quotedText: string, literal: bool): string
...
         #\       but we keep them out for performance-reasons until we actually see
         #\       those in mi-returns
         #\ \ ->substitute('\\0x\(\x\x\)', {-> eval('"\x' .. submatch(1) .. '"')}, 'g')
         #\ \ ->substitute('\\0x00', NullRepl, 'g')
-        \ ->substitute('\\\\', '\', 'g')
-        \ ->substitute(NullRepl, '\\000', 'g')
-  if !literal
+        ->substitute('\\\\', '\', 'g')
+        ->substitute(NullRepl, '\\000', 'g')
+  if literal == false
```
yegappan: When using a Boolean variable, there is no need to it compare against true or false. Can you use `!literal`?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1475,14 +1533,14 @@ def TermDebugBalloonExpr(): string
   if v:beval_winid != sourcewin
     return ''
   endif
-  if !stopped
+  if stopped == false
```
yegappan: When using a Boolean variable, there is no need to it compare against true or false. The previous code is more readable.

yegappan (changes requested):

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Correct. I get burned once with Python due to the truthy/falsy thing that I now tend to be explicit on pretty much everything. However, being Boolean variables, I agree that the == true and ==false are not needed. I'll fix it asap.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
chrisbra: That checkme should be fixed before merged. So either the implementation is correct. Then we can merge it or it isn't then we shouldn't.

chrisbra:
Thanks. All in all I like the approach this takes and the cleanup is valuable. Please address all the mentioned items.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done! I extended the process also in non-cited places that suffer of the same issue.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done! I extended the process also in non-cited places that suffer of the same issue.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done! I extended the process also in non-cited places that suffer of the same issue.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done! I extended the process also in non-cited places that suffer of the same issue.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done! I extended the process also in non-cited places that suffer of the same issue.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: #14984 is merged.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I hear you :) 
But in my opinion you lose readability: one thing is to read that you exited unexpectedly because of `cmd_name`, one is to read that you exited because of `string(GetCommand()[0])`. If we all agree, I would keep it as is. There is more room for optimization elsewhere. :)

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Perfect! Thanks for the PR!

ubaldot:
@chrisbra Is there anything else that you wish me to do before merging this PR? I think the points raised by @errael shall be discussed and the removal o f` Echoerr('Termdebug already loaded.')` at the beginning.  But IMHO we can leave them as they are.

chrisbra:
Yeah, I think its fine, if you could please remove the CHECKME now?

ubaldot:
> Yeah, I think its fine, if you could please remove the CHECKME now?

Done!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: My suggestion is putting the calculation of `cmd_name` inside the if
```vim
  if gdbproc_status !=# 'run'
    var cmd_name = string(GetCommand()[0])
    Echoerr($'{cmd_name} exited unexpectedly')
```
I don't see how calculating a value that is almost never used improves things. If there is an issue with clarity, a comment should suffice.
(this is **not about** putting the expr in string interpolate, that's primarily a style issue)

chrisbra: I'll adjust that when merging. Thanks!

ubaldot:
@chrisbra the master branch of my fork is still 7 commits ahead and I cannot sync with the upstram repo (I made thesee changes directly on the master branch in my own fork). It looks like this PR has not been merged yet... Can you confirm?

errael:
It is merged. Take a look at https://github.com/vim/vim/commits/master/ there's [patch 9.1.0482: termdebug plugin needs more love](https://github.com/vim/vim/commit/ef8eab86e29091eaff0d3fb0bc3f7c90f468f6aa)

When I pulled from the main repo, I get it in my local clone, also when I update and pull from my fork.

I _never_ update my master directly, always push a branch. Do you have merge conflicts?

errael:
@ubaldot I did have an accident a while back; I can't remember all the details, but...

I have "Branch protection rules", "require a pull request before merging" in my fork set to prevent direct changes to master.

I vaguely remember that after the accident, when my fork was screwed up, and I couldn't update,
there was a button under the "Sync Fork" pulldown in my fork. I don't remember what it said, but
it didn't work to clean things up. I had to temporarily disable the "require a pull request before
merging" and then the button (whatever it was called) worked and things got back in sync. It may
be that I had to repush some stuff from my local clone; don't remember.

ubaldot:
@errael In-fact I did the "mistake" of developing directly on (my fork) master and then push onto (upstream) master. Now I have this situation:

image

Whereas I expected to be only 32 commits behind. But I have now seen that my changes are included in a commit named "termdebug needs more love" so I assume I can safely discard these 7 commits.

PR #14984, opened 2024-06-13
If vim9script, return null_job from term_getjob when there is no job.

errael:
This change gets in more in accord with the spec/help. (I think)

It also gets `job_status(term_getjob(0))` to work; returns `'fail'` rather than throwing an exception. This came up around the vim9 termdebug port. @ubaldot 

Note that the return values is adjusted acc'd to if vim9; I'm assuming legacy can not handle a `null_job`. @yegappan


chrisbra:
Thanks. I think this makes sense. @yegappan ?

yegappan (approved):

yegappan:
> Thanks. I think this makes sense. @yegappan ?

Yes. This makes sense.

chrisbra:
thanks all!

PR #14985, opened 2024-06-13
fuzzy_match_str_with_pos() may leak memory on failure

zeertzjq:
Problem:  fuzzy_match_str_with_pos() may leak memory on failure.
Solution: Also free match_positions itself.


zeertzjq:
Closing in favor of #14987

PR #14986, opened 2024-06-13
fix(completion): when no leader no need run fuzzy logic

glepnir:
Problem: when no leader or empty it also run into fuzzy logic.

Solution: this is unnecessary.


zeertzjq:
All the NULL check below can now be removed.

chrisbra:
thanks!

PR #14987, opened 2024-06-13
fuzzy_match_str_with_pos() does unnecessary list operations

zeertzjq:
Problem:  fuzzy_match_str_with_pos() does unnecessary list operations.
Solution: Use fuzzy_match() directly.


chrisbra:
thanks for the cleanup :)

PR #14988, opened 2024-06-13
Sorting of completeopt+=fuzzy is not stable

zeertzjq:
Problem:  Sorting of completeopt+=fuzzy is not stable.
Solution: Compare original indexes when scores are the same.

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1203,11 +1203,13 @@ trigger_complete_changed_event(int cur)
  * pumitem qsort compare func
  */
     static int
-ins_compl_fuzzy_sort(const void *a, const void *b)
+ins_compl_fuzzy_cmp(const void *a, const void *b)
 {
     const int sa = (*(pumitem_T *)a).pum_score;
     const int sb = (*(pumitem_T *)b).pum_score;
-    return sa == sb ? 0 : sa < sb ? 1 : -1;
+    const int ia = (*(pumitem_T *)a).pum_idx;
+    const int ib = (*(pumitem_T *)b).pum_idx;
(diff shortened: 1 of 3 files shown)


src/insexpand.c
```diff
@@ -1203,11 +1203,13 @@ trigger_complete_changed_event(int cur)
...
     static int
-ins_compl_fuzzy_sort(const void *a, const void *b)
+ins_compl_fuzzy_cmp(const void *a, const void *b)
 {
     const int sa = (*(pumitem_T *)a).pum_score;
     const int sb = (*(pumitem_T *)b).pum_score;
-    return sa == sb ? 0 : sa < sb ? 1 : -1;
+    const int ia = (*(pumitem_T *)a).pum_idx;
+    const int ib = (*(pumitem_T *)b).pum_idx;
+    return sa != sb ? (ia == ib ? 0 : (ia < ib ? -1 : 1)) : (sa < sb ? 1 : -1);
```
chrisbra: Is that a typo?
```suggestion
    return sa == sb ? (ia == ib ? 0 : (ia < ib ? -1 : 1)) : (sa < sb ? 1 : -1);
```

zeertzjq: oops

src/testdir/test_ins_complete.vim
```diff
@@ -2578,4 +2578,24 @@ func Test_complete_fuzzy_match()
   unlet g:word
 endfunc
 
+" Check that tie breaking is stable for completeopt+=fuzzy (which should
+" behave the same on different platforms).
+func Test_complete_fuzzy_match_tie()
```
zeertzjq, line 2583: Without this patch, this test fails on Windows.

chrisbra:
thanks!

PR #14989, opened 2024-06-13
runtime(vim): Update base-syntax, match :catch and :throw args

dkearns:
Match `:catch /{pattern}/` and `:throw {expr1}`.


runtime/syntax/generator/vim.vim.base
```diff
@@ -206,6 +206,11 @@ syn match   vimBehave	"\<be\%[have]\>"	nextgroup=vimBehaveBang,vimBehaveModel,vi
 syn match   vimBehaveBang	contained	"\a\@1<=!" nextgroup=vimBehaveModel skipwhite
 syn keyword vimBehaveModel	contained	mswin	xterm
 
+" Exception Handling {{{2
+syn keyword	vimThrow	th[row]	skipwhite nextgroup=@vimExprList
+syn keyword	vimCatch	cat[ch]	skipwhite nextgroup=vimCatchPattern
```
chrisbra, line 211: Should those be `cat\%[ch]` and `th\%[row]`

dkearns: No, they're `syn-keyword` definitions rather than `syn-match`.

chrisbra: ah, obviously 🤦

PR #14992, opened 2024-06-13
filetype: snakemake files not recognized

ribru17:
Adds detection for [Snakemake](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#distribution-and-reproducibility) files

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -2243,6 +2243,9 @@ au BufNewFile,BufRead *.smt,*.smith		setf smith
 " Smithy
 au BufNewFile,BufRead *.smithy			setf smithy
 
+" Snakemake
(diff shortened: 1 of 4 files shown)


runtime/filetype.vim
```diff
@@ -2243,6 +2243,9 @@ au BufNewFile,BufRead *.smt,*.smith		setf smith
 " Smithy
 au BufNewFile,BufRead *.smithy			setf smithy
 
+" Snakemake
+au BufNewFile,BufRead Snakemake,*.smk		setf snakemake
```
zeertzjq:
```suggestion
au BufNewFile,BufRead Snakefile,*.smk		setf snakemake
```

ribru17: Oops, thanks!

PR #14996, opened 2024-06-13
Matched text shouldn't be highlighted in "kind" and "menu"

zeertzjq:
Problem:  Matched text shouldn't be highlighted in "kind" and "menu".
Solution: Pass hlf_T instead of the attribute.  Fix indent.


chrisbra:
thanks!

PR #14999, opened 2024-06-14
runtime(java): Add a config variable for commonly used compiler options

dkearns:
The value of `g:javac_makeprg_params` is now added to the value of 'makeprg' as an option string.

@Konfekt, does that work for you?


chrisbra:
thanks!

PR #15004, opened 2024-06-15
Wrong padding for pum "kind" with 'rightleft'

zeertzjq:
Problem:  Wrong padding for pum "kind" with 'rightleft'.
Solution: Fix off-by-one error.

The screen_fill() above is end-exclusive, and
- With 'rightleft' it fills `pum_col - pum_base_width - n + 1` to `col`,
  so the next `col` should be `pum_col - pum_base_width - n`.
- With 'norightleft' it fills `col` to `pum_col - pum_base_width + n - 1`,
  so the next `col` should be `pum_col - pum_base_width + n`.


chrisbra:
thanks!

PR #15005, opened 2024-06-15
Cmdline pum doesn't work properly with 'rightleft'

zeertzjq:
Problem:  Cmdline pum doesn't work properly with 'rightleft'.
Solution: Don't use curwin->w_p_rl in cmdline mode in pum_redraw().  Use
          a static variable since pum_may_redraw() may be called in any
          mode.  Also correct position of other popups with 'rightleft'.


zeertzjq:
For right-click menu and 'balloonevalterm', using the value of 'rightleft' may not be ideal either, but I'll keep the old behavior of taking 'rightleft' into account in this PR and only correct the positioning. For cmdline completion using 'rightleft' doesn't make sense as the cmdline is always left-to-right.

chrisbra:
thanks!

PR #15008, opened 2024-06-15
fix(completion): use exist shown_compl instead after_first_compl

glepnir:
Sorry I couldn't update the PR in time because of computer problems yesterday. I ignored the existing shown_compl and didn't need anyone to add any new variables.


chrisbra:
thanks!

PR #15010, opened 2024-06-15
fix(completion): avoid breaking default completion mode

### LINKED ISSUE #15009: fuzzy completion acts as if 'noinsert' is enabled
### Steps to reproduce

1. `vim -Nu NONE`
2. add text `hello helion hero`
3. `:set completeopt=menu,fuzzy`
4. in a new line type `he<C-n>`

### Expected behaviour

`hello` should be inserted and selected in popup menu.

Instead `hello` is selected only.

image

Basically it acts as if `noinsert` is in `completeopt` for non-fuzzy completion.

### Version of Vim

9.1.487

### Environment

debian12, bash

### Logs and stack traces

_No response_

glepnir:
currently we fuzzy search item in match array. so if array is empty we should still find from list. avoid broken default completion behavior. 

it's part of https://github.com/vim/vim/pull/14976, seems like #14976 need some dicussions so split it to a single pr

Fix #15009

### DIFF
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -2573,6 +2573,15 @@ func Test_complete_fuzzy_match()
   call feedkeys("S\<C-x>\<C-o>fb\<C-n>", 'tx')
   call assert_equal('fooBaz', g:word)
 
+  " avoid break default completion behavior
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello help hero h'])
+  exe "normal A\<C-X>\<C-N>"
+  call assert_equal('hello help hero hello', getline('.'))
+  set completeopt+=noinsert
+  exe "normal Shello help hero h\<C-X>\<C-N>"
+  call assert_equal('hello help hero h', getline('.'))
+
   " clean up
   set omnifunc=
   bw!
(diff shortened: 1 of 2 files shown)


src/testdir/test_ins_complete.vim
```diff
@@ -2573,6 +2573,15 @@ func Test_complete_fuzzy_match()
   call feedkeys("S\<C-x>\<C-o>fb\<C-n>", 'tx')
   call assert_equal('fooBaz', g:word)
 
+  " avoid break default completion behavior
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello help hero h'])
+  exe "normal A\<C-N>"
```
zeertzjq: Will using `\<C-X>\<C-N>` here be better?

src/testdir/test_ins_complete.vim
```diff
@@ -2573,6 +2573,15 @@ func Test_complete_fuzzy_match()
...
+  call assert_equal('hello help hero hello', getline('.'))
+  set completeopt+=noinsert
+  exe "normal Shello help hero h\<C-N>"
```
zeertzjq: ditto

PR #15011, opened 2024-06-15
runtime(vim): Update base-syntax, match shebang lines

dkearns:
Match shebang lines in Vim9 and legacy script.

Mark these as an error if they appear anywhere other than the first line of a legacy-script file.  In Vim9 script these match as normal line comments rather than an error.


chrisbra:
oh does this actually work? can we use a shebang line and run a vimscript directly? Is this documented somewhere?

chrisbra:
Interesting, I wasn't aware. of this. I searched the help for `shebang` but didn't find anything, so let me add this help-tag

Shane-XB-Qian:
> oh does this actually work? can we use a shebang line and run a vimscript directly? Is this documented somewhere? 

interesting too, but how to run it directly? i tried the example in its help, then `./foo.vim` seems failed.
// via bash cli

-- 
shane.xb.qian

chrisbra:
You need to make it executable: `chmod +x foo.vim && ./foo.vim`

Shane-XB-Qian:
> You need to make it executable: `chmod +x foo.vim && ./foo.vim`

yes, it was executable; after look into a bit more, seems the example of help was not accurate, it maynot find vim without path.

-- 
shane.xb.qian

chrisbra:
The example worked for me correctly. Of course you need to remove the `:q`

Shane-XB-Qian:
no, just the shebang needed to specify where vim is as general shebang it is. :sweat_smile:
// but nevermind, it should be no common fixed answer where vim must be.

-- 
shane.xb.qian

chrisbra:
How about this:
```
#!/bin/vim -S

let mylogbook='$HOME/logbook.md'
exe ':e ' .. mylogbook
$
put ='## ' .. strftime('%m. %b %Y')
norm! o
```

Shane-XB-Qian:
no, first line was the matter.

PR #15012, opened 2024-06-15
filetype: Vim-script files not detected by shebang line

dkearns:
Problem:  Vim-script files may not be recognised
Solution: Add shebang line detection


chrisbra:
thanks!

PR #15013, opened 2024-06-15
fix: [termdebug] correct K/-/+ mappings check

Shane-XB-Qian:
// and did same like 'saved_mousemodel' chk

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -194,9 +194,9 @@ def InitScriptVariables()
 
   winbar_winids = []
 
-  k_map_saved = {}
-  plus_map_saved = {}
-  minus_map_saved = {}
+  k_map_saved = null_dict
+  plus_map_saved = null_dict
+  minus_map_saved = null_dict
 
   if has('menu')
     saved_mousemodel = null_string
@@ -1232,32 +1232,38 @@ def DeleteCommands()
   delcommand Var
   delcommand Winbar
 
-  if exists('k_map_saved')
-    if !empty(k_map_saved) && !k_map_saved.buffer
+  if k_map_saved isnot null_dict
+    if !empty(k_map_saved) && k_map_saved.buffer
+      # pass
+    elseif !empty(k_map_saved) && !k_map_saved.buffer
       nunmap K
       mapset(k_map_saved)
     elseif empty(k_map_saved)
-      nunmap K
+      silent! nunmap K
     endif
-    k_map_saved = {}
+    k_map_saved = null_dict
   endif
-  if exists('plus_map_saved')
-    if !empty(plus_map_saved) && !plus_map_saved.buffer
+  if plus_map_saved isnot null_dict
+    if !empty(plus_map_saved) && plus_map_saved.buffer
+      # pass
+    elseif !empty(plus_map_saved) && !plus_map_saved.buffer
       nunmap +
       mapset(plus_map_saved)
     elseif empty(plus_map_saved)
-      nunmap +
+      silent! nunmap +
     endif
-    plus_map_saved = {}
+    plus_map_saved = null_dict
   endif
-  if exists('minus_map_saved')
-    if !empty(minus_map_saved) && !minus_map_saved.buffer
+  if minus_map_saved isnot null_dict
+    if !empty(minus_map_saved) && minus_map_saved.buffer
+      # pass
+    elseif !empty(minus_map_saved) && !minus_map_saved.buffer
       nunmap -
       mapset(minus_map_saved)
     elseif empty(minus_map_saved)
-      nunmap -
+      silent! nunmap -
     endif
-    minus_map_saved = {}
+    minus_map_saved = null_dict
   endif
 
   if has('menu')


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,32 +1232,38 @@ def DeleteCommands()
...
     elseif empty(k_map_saved)
-      nunmap K
+      silent! nunmap K
     endif
-    k_map_saved = {}
+    k_map_saved = null_dict
   endif
-  if exists('plus_map_saved')
-    if !empty(plus_map_saved) && !plus_map_saved.buffer
+  if plus_map_saved isnot null_dict
```
errael, line 1246:
```suggestion
  if plus_map_saved != null
```
For reference
```
vim9script
var d1: dict<any> = null_dict
var d2: dict<any> = {}
echo 'is d1 null?' d1 == null
echo 'is d2 null?' d2 == null
```
outputs
```
is d1 null? true
is d2 null? false
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,32 +1232,38 @@ def DeleteCommands()
   delcommand Var
   delcommand Winbar
 
-  if exists('k_map_saved')
-    if !empty(k_map_saved) && !k_map_saved.buffer
+  if k_map_saved isnot null_dict
```
errael, line 1235:
```suggestion
  if k_map_saved != null
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,32 +1232,38 @@ def DeleteCommands()
...
     elseif empty(plus_map_saved)
-      nunmap +
+      silent! nunmap +
     endif
-    plus_map_saved = {}
+    plus_map_saved = null_dict
   endif
-  if exists('minus_map_saved')
-    if !empty(minus_map_saved) && !minus_map_saved.buffer
+  if minus_map_saved isnot null_dict
```
errael, line 1257:
```suggestion
  if minus_map_saved != null
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
Shane-XB-Qian: if you recalled some months ago what we argued in such null/null_obj topic which chris applied your PR,
i am confused now truly null or null_obj which was your taste? :sweat_smile:
anyway, `{} isnot null_dict` also is true, specially it is just keeping it same operation as `saved_mousemodel` did,
so i think ok.

errael: strings are the weirdest of them all when it comes to null handling. With strings you must use `isnot null_string`, as it says at  `:help null-anomalies`
```
Unlike the other containers, an uninitialized string is equal to null.
```

I left out some important cases from the example.
```vim
echo 'is d1 null_dict?' d1 == null_dict
echo 'is d2 nul_dictl?' d2 == null_dict
```
Both return `true`.

`!= null` is easier to understand than `isnot null_dict` (at least for me). It also looks like what you'd expect to see in most any other language; to me, that makes it easier to maintain.

As you say, what's currently there works; just not as maintainable (but that's only my opinion)

ubaldot:
@Shane-XB-Qian Thanks! Could you please add some tests?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,32 +1232,38 @@ def DeleteCommands()
...
-    if !empty(k_map_saved) && !k_map_saved.buffer
+  if k_map_saved isnot null_dict
+    if !empty(k_map_saved) && k_map_saved.buffer
+      # pass
+    elseif !empty(k_map_saved) && !k_map_saved.buffer
       nunmap K
       mapset(k_map_saved)
     elseif empty(k_map_saved)
-      nunmap K
+      silent! nunmap K
```
chrisbra, line 1242: why the silent!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,32 +1232,38 @@ def DeleteCommands()
...
-    if !empty(plus_map_saved) && !plus_map_saved.buffer
+  if plus_map_saved isnot null_dict
+    if !empty(plus_map_saved) && plus_map_saved.buffer
+      # pass
+    elseif !empty(plus_map_saved) && !plus_map_saved.buffer
       nunmap +
       mapset(plus_map_saved)
     elseif empty(plus_map_saved)
-      nunmap +
+      silent! nunmap +
```
chrisbra, line 1253: why the silent!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
Shane-XB-Qian: there maybe buf local mapping at source, so e.g gdb had no 'K' mapping actually.

Shane-XB-Qian: there maybe buf local mapping at source window, then e.g gdb window had no 'K' mapping actually.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
chrisbra: that is not resolved

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
chrisbra: alright then, thanks. Can you please make the suggested changes from @errael ?

Shane-XB-Qian: that's a confusion to himself, https://github.com/vim/vim/pull/15013#discussion_r1641334643 had explained.
// though you accepted his PR vs/over mine in the past about this null/null_obj topic, but just a code style.
// vs this termdebug you had accepted null_obj for `saved_mousemodel`, a similar flag, so i just kept the same.

Shane-XB-Qian: that's a confusion to himself, https://github.com/vim/vim/pull/15013#discussion_r1641334643 had explained.
// though you accepted his PR vs/over mine in the past about this null/null_obj topic, but just a code style.
// vs this termdebug you had accepted null_obj for `saved_mousemodel`, a similar flag, so i just kept the same.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
Shane-XB-Qian: https://github.com/vim/vim/pull/15013#discussion_r1641661003

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
chrisbra: ah, okay. So you initialize to null_dict and then testing against null_dict should be correct.

PR #15014, opened 2024-06-15
Implement lstat() for Windows

### LINKED ISSUE #14933: "all links" argument to glob() not working on Windows
### Steps to reproduce

Create a bad symbolic link (with missing target) using MkLink command (needs Administrator access, or needs the Local Policy to be set accordingly).

Use `glob()` with alllinks argument set to true, to list the new link

```Vim
:!If Not Exist BadLink (MkLink BadLink NoSuchFileOrDirectory)
:echo 'BadLink'->glob(v:false, v:true, v:true)
```

If I use `readdir()` on the directory containing the symlink, then I can list the bad link.

### Expected behaviour

`glob()` should return bad symlinks if "all links" argument is `v:true`

### Version of Vim

9.1 (patches 1-445)

### Environment

Windows GUI (gvim.exe)

### Logs and stack traces

_No response_

LemonBoy:
lstat() differs from stat() in how it handles symbolic links, the former
doesn't resolve the symlink while the latter does so.

Implement a simple yet effective fallback using Win32 APIs.

Fixes https://github.com/vim/vim/issues/14933

### DIFF
--- a/src/testdir/test_functions.vim
+++ b/src/testdir/test_functions.vim
@@ -3823,12 +3823,6 @@ func Test_glob2()
 endfunc
 
 func Test_glob_symlinks()
-  " Symlinks on Windows are emulated in a number of ways, including by copying the source
-  " file over. Since we cannot detect that skip the test when running on those platforms.
-  if has('win32') && executable('uname') && system('uname -s') =~ '^MSYS\|MINGW\|CYGWIN'
-    throw 'Skipped: does not work on MinGW/Cygwin/Msys'
-  endif
-
   call writefile([], 'Xglob1')
 
   if has("win32")
--- a/src/testdir/test_functions.vim
+++ b/src/testdir/test_functions.vim
@@ -3823,6 +3823,12 @@ func Test_glob2()
 endfunc
 
 func Test_glob_symlinks()
+  " Symlinks on Windows are emulated in a number of ways, including by copying the source
+  " file over. Since we cannot detect that skip the test when running on those platforms.
+  if has('win32') && executable('uname') && system('uname -s') =~ '^MSYS\|MINGW\|CYGWIN'
+    throw 'Skipped: does not work on MinGW/Cygwin/Msys'
+  endif
+
   call writefile([], 'Xglob1')
 
   if has("win32")
--- a/src/testdir/test_functions.vim
+++ b/src/testdir/test_functions.vim
@@ -3822,6 +3822,29 @@ func Test_glob2()
   endif
 endfunc
 
+func Test_glob_symlinks()
+  call writefile([], 'Xglob1')
+
+  if has("win32")
+    silent !mklink XglobBad DoesNotExist
+    if v:shell_error
+      throw 'Skipped: cannot create symlinks'
+    endif
+    silent !mklink XglobOk Xglob1
+  else
+    silent !ln -s XglobBad DoesNotExist
+    silent !ln -s XglobOk Xglob1
+  endif
+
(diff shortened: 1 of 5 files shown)


chrisbra:
thank you! can we get a test please?

src/testdir/test_functions.vim
```diff
@@ -3822,6 +3822,29 @@ func Test_glob2()
...
+
+  if has("win32")
+    silent !mklink XglobBad DoesNotExist
+    if v:shell_error
+      throw 'Skipped: cannot create symlinks'
+    endif
+    silent !mklink XglobOk Xglob1
+  else
+    silent !ln -s XglobBad DoesNotExist
+    silent !ln -s XglobOk Xglob1
```
k-takata: It seems reversed:
```suggestion
    silent !ln -s DoesNotExist XglobBad
    silent !ln -s Xglob1 XglobOk
```

k-takata:
MinGW is now failing.
Maybe we should not fall back to `_wstat()` in `wstat_symlink_aware()`?

LemonBoy:
Hmm, it's just MinGW being MinGW. `ln -s` performs a full copy (!!!) unless you specify some flag and enable the Windows' developer mode. What shall we do here? Perhaps skipping the test on MinGW is an option here.

chrisbra:
If that is the case, I am fine with skipping. But I don't know how to detect that Vim is running in a Mingw/Msys environment.

vim-ml:
On 2024-06-17, Christian Brabandt wrote:
> If that is the case, I am fine with skipping. But I don't know how to detect
> that Vim is running in a Mingw/Msys environment.

My MINGW64 installation has a couple of environment variables that
you could test.

    MSYSTEM=MINGW64
    OSTYPE=msys

Regards,
Gary

k-takata:
Moreover, how about merging `mswin_lstat()` and `wstat_symlink_aware()` into one function?
https://github.com/k-takata/vim/commit/b48ce1c121270c97cd404addd336bddd5c04a52c
No need to skip the symlink test on MinGW.
Here are the results: https://github.com/k-takata/vim/actions/runs/9556704996

> `ln -s` performs a full copy (!!!)

It is not the reason why it failed on MinGW. `has("win32")` is true on MinGW, so the symlinks are created by the mklink command.
If we want to test the MSYS version of Vim, we need to care about the `ln -s` issue, but currently, we don't have MSYS tests.

k-takata:
To skip the test on MSYS, the check should be like this:
```vim
  if has('win32unix') && executable('uname') && system('uname -s') =~ '^MSYS'
    throw 'Skipped: does not work on Msys'
  endif
```
Note that `has('win32unix')` is used instead of `has('win32')` because MSYS is a variant of Cygwin.
`has('win32')` returns 0 on Cygwin/MSYS.

LemonBoy:
> Moreover, how about merging mswin_lstat() and wstat_symlink_aware() into one function?

Good idea, merged!

> It is not the reason why it failed on MinGW. has("win32") is true on MinGW, so the symlinks are created by the mklink command.
If we want to test the MSYS version of Vim, we need to care about the ln -s issue, but currently, we don't have MSYS tests.

Oh well, it seems that every Unix layer tries do do its own thing with symlinks. Let's forget about MSYS for the time being then.

> To skip the test on MSYS, the check should be like this:

On my MSYS-compiled version (using `Make_cyg_ming.mak`) the `'win32unix'` is not defined.

chrisbra:
thanks, let me include it then!

Issue #14994, opened 2024-06-13
[termdebug vim9script] Problems with breakpoint handling


errael:
### Steps to reproduce

1. Build /src/tools/vim/src with debug info
2. open termdebug on the vim just built
3. in gdb window, `b f_term_getjob`
4. in gdb window, `run`
5. in debugged program window, edit/add
```
    vim9script
    term_getjob(0)
```
6. in debugged program window, `:source`

Observe: get breakpoint, source code window has terminal.c

7. in source code window, `:message`

Observe
```
different fname: '' vs '/src/tools/vim/src/terminal.c'
"/src/tools/vim/src/terminal.c" 7831L, 190621B
Error detected while processing function <SNR>41_CommOutput[21]
    ..<SNR>41_HandleCursor[66]
    ..BufRead Autocommands for "*"
    ..function <SNR>41_BufRead:
line    5:
E1013: Argument 1: type mismatch, expected number but got string
```

8. in gdb window - do `up`, then do `down`

Observe
- the `up` is quick, the `down` takes a few seconds for the source code window to update
- the cursor is left in the source code window
- the execute line in the source window is not highlighted
- the breakpoint is not highlighted

That's 4 observable problems, all of them probably due to the "Error detected".

### Expected behaviour

No error in `:messages`; speedy `down`; proper highlighting in source window after the `down`; 

### Version of Vim

9.1.482

### Environment

linux/gtk3

errael:
@ubaldot I played around a bit and
```diff
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1949,7 +1949,7 @@
   for [id, entries] in items(breakpoints)
     for [subid, entry] in items(entries)
       if entry['fname'] == fname
-        PlaceSign(id, subid, entry)
+        PlaceSign(str2nr(id), str2nr(subid), entry)
       endif
     endfor
   endfor
```
seems to take care of the immediate problem. But looking around I see that `id` is sometimes a `string` and sometimes a `number`. So I'll skip doing a PR since a more comprehensive look is in order.

errael:
Cleaning up my desktop, I noticed the following. This proposed change is not a bug fix.

The key is that legacy does automatic conversion of `string` to `number` in some cases.

```diff
@@ -1856,8 +1856,7 @@
 
     # If "nr" is 123 it becomes "123.0" and subid is "0".
[...]

PR #15015, opened 2024-06-15
fix: [termdebug] id/subid maybe mis-match type

Shane-XB-Qian:
// fix err msg of type mis-match: string -> number


errael:
Fixes #14994

chrisbra:
thanks!

PR #15016, opened 2024-06-15
runtime(java): Fold multi-line comments with the syntax kind of &fdm

zzzyxwvut:
Also:

- Restore the capability to mark as an error braces nested  
  in parens with `g:javaInParen`.
- Try not to fold top-level-type bodies.  (Defining multiple  
  package-private top level types in a single source file is  
  not recommended as it can impose order among compilation  
  units; so it is assumed that only one such top level type  
  is usually defined.)
- Compose ‘method header’ highlighting and block braces  
  folding.
- Do not highlight block braces whenever ‘method header’  
  highlighting is requested.

This bundling of ‘method headers’ and block braces for  
highlighting can be traced back to Vim `v5.0`; however, no  
comment or documentation entry conveys any justification.  
For example, it is hard to discover the connection between  
block braces for `while`, `if`, etc., statements and method  
body block braces.  The former behaviour can be attained in,  
e.g. `~/.vim/after/syntax/java.vim`:

```vim
if exists("g:java_highlight_functions")
    syn clear javaBlock javaInParen
    syn match javaBlockOther "[{}]"
    syn region javaBlock transparent matchgroup=javaBlockStart
	\ start="\%(^\|^\S[^:]\+\)\@120<!{" end="}" fold
    hi def link javaBlockStart javaFuncDef
    hi def link javaBlockOther javaBlockStart

    if exists("g:java_mark_braces_in_parens_as_errors")
	syn match javaInParen contained "[{}]"
    endif
endif
```

Note: Read ‘a method header omitting a _throws_ clause’ for  
every ‘method header’ appellation used above.


zzzyxwvut:
As I have already mentioned it [elsewhere](https://github.com/vim/vim/issues/14245#issuecomment-2131433317), folded lines are  
always treated as opened for a syntax test when calculating  
the line offsets and determining what next batch of lines to  
dump to a file; hence, the generated dump files for them  
exceed the required number because more lines are repeated.

chrisbra:
thanks!

PR #15018, opened 2024-06-15
Test for patch 9.1.0489 doesn't fail without the fix

zeertzjq:
Problem:  Test for patch 9.1.0489 doesn't fail without the fix.
Solution: Use "!" flag of feedkeys() so that ex_normal_busy is not set
          and ins_compl_check_keys() is not skipped.


chrisbra:
thanks!

PR #15020, opened 2024-06-16
Wrong matched text highlighted in pum with 'rightleft'

zeertzjq:
Problem:  Wrong matched text highlighted in pum with 'rightleft'.
Solution: Match using the original text instead of the reversed text.


chrisbra:
thanks!

PR #15023, opened 2024-06-16
fix(highlight): pum matched text respect icase

### LINKED ISSUE #15021: highlight groups PmenuMatch and PmenuMatchSel some buggy
### Steps to reproduce

1. did not expect icase of match:
image

2. did not reflect the change of base:
// os.if -> the `if` highlighted -> changed to os.i which backspace the `f` -> the `i` highlight missed.
image

### Expected behaviour

work.

### Version of Vim

v9.1.493

### Environment

linux

### Logs and stack traces

#14694

glepnir:
Problem: 
1. STRNCMP does not respect the upper case when icase is set of complete item.
2. ins_compl_stop will free all leader even bs just delete the one leader. 

Solution: 
1.use MB_STRNICMP instead 
2. when compl_leader is null use `compl_orig_text` for match.

Fix #15021


src/popupmenu.c
```diff
@@ -485,7 +485,7 @@ pum_screen_put_with_attr(int row, int col, char_u *text, int textlen, hlf_T hlf)
 	    }
 	}
 	else if (!in_fuzzy && (ptr - text < leader_len)
-			     && (STRNCMP(text, match_leader, leader_len) == 0))
+			     && (MB_STRNICMP(text, match_leader, leader_len) == 0))
```
zeertzjq: This function is called repeatedly when it only needs to be called once. Maybe you can rebase this onto #15020.

src/insexpand.c
```diff
@@ -1448,7 +1448,7 @@ ins_compl_show_pum(void)
     char_u *
 ins_compl_leader(void)
 {
-    return compl_leader;
+    return compl_leader != NULL ? compl_leader : compl_orig_text;
```
zeertzjq, line 1451: Is this change tested?

glepnir: missed. should add a bs step.

src/popupmenu.c
glepnir: oh nice. i will wait that.

glepnir: mark as draft then.

chrisbra: I have merged it.

zeertzjq:
I force-pushed to your branch to rebase this PR. It also seems that the two screen dumps add in this PR are the same, so I combined them into one.

glepnir:
it's fine :)

Issue #15006, opened 2024-06-15
The popup functions in Vim lack a function to change  the buffer in popup


Yggdroot:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
Neovim has a function `nvim_win_set_buf`, which can change the buffer in the specified window, but vim's popup functions do
not have a function like it. 
If I want to change the content in the popup, I have to use `popup_settext`, if a file is too large, reading from the file and `popup_settext` would be too costly. It would be better if vim's popup has a function like `nvim_win_set_buf`.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

chrisbra:
> If I want to change the content in the popup, I have to use popup_settext, if a file is too large, reading from the file and popup_settext would be too costly.

I don't understand that comment. If the file is large and hasn't been loaded as a buffer yet, it will be exactly as costly because you need to  load this as a buffer. It may even be more costly, since we have to create a buffer for that file. Using `popup_settext()` can even be better, since you can use `readfile()` directly and do not need to handle the overhead of opening that file in a buffer yet. 

If the buffer has been already loaded, you may be right, but the same overhead has happened eventually, but might have happened a bit earlier, when the file was read into a buffer.

So I don't think this benefits us anything here.

Yggdroot:
I meant the buffer has already been loaded. 
Yes, the overhead is inevitable when loading the buffer. however，this only happens once. 
Actually, the popup feature is used only when we write a plugin. For example, if there is a plugin to preview the opening buffers in a popup window, I may switch the buffer back and forth,  if I switch to a large buffer 3 times, I have to call popup_settext() 3 times.   
Notice that just `popup_settext()`  without readfile()  is also very costly.

![1](https://github.com/vim/vim/assets/2705841/635bce4d-a0d5-4043-bf7c-c5586c6ca177)

Shane-XB-Qian:
do you use `matchstr()` etc funcs for ptn search in your code, if yes, maybe that's the cause of slowness,
i faced some times, maybe that's a flaw of perf, if you (any of you) can reproduce it as well, wish can be fixed.

bfrg:
> For example, if there is a plugin to preview the opening buffers in a popup window, I may switch the buffer back and forth, if I switch to a large buffer 3 times, I have to call popup_settext() 3 times.

[...]

PR #15026, opened 2024-06-16
  create popup_setbuf() function

chrisbra:
related: #15006

Problem:  cannot change buffer in a popup
Solution: add popup_setbuf() function


src/popupwin.c
```diff
@@ -2844,6 +2844,51 @@ f_popup_settext(typval_T *argvars, typval_T *rettv UNUSED)
...
+ * popup_setbuf({id}, {bufnr})
+ */
+    void
+f_popup_setbuf(typval_T *argvars, typval_T *rettv UNUSED)
+{
+    int		id;
+    win_T	*wp;
+    buf_T	*buf;
+
+    if (in_vim9script()
```
yegappan: As this is a new function, this check for the Vim9 script can be removed.  In the existing functions this check was added to not break backward compatibility with legacy scripts.

src/popupwin.c
```diff
@@ -2844,6 +2844,51 @@ f_popup_settext(typval_T *argvars, typval_T *rettv UNUSED)
...
+	    && (check_for_number_arg(argvars, 0) == FAIL
+		|| check_for_buffer_arg(argvars, 1) == FAIL))
+	return;
+
+    if (check_for_buffer_arg(argvars, 1) == FAIL)
```
yegappan: If the previous comment is addressed, then this check can be removed.

src/popupwin.c
```diff
@@ -2844,6 +2844,51 @@ f_popup_settext(typval_T *argvars, typval_T *rettv UNUSED)
...
+	return;
+
+    id = (int)tv_get_number(&argvars[0]);
+    wp = find_popup_win(id);
+    if (wp == NULL)
+	return;
```
yegappan, line 2867: Should the return value be set to -1 in this case?

src/popupwin.c
```diff
@@ -2844,6 +2844,51 @@ f_popup_settext(typval_T *argvars, typval_T *rettv UNUSED)
...
+
+    buf = tv_get_buf_from_arg(&argvars[1]);
+
+    if (buf == NULL)
+	return;
+#ifdef FEAT_TERMINAL
+    else if (buf->b_term != NULL && popup_terminal_exists())
```
yegappan: Is the "else" needed here as the previous "if" check does a return?

yegappan (changes requested):

src/popupwin.c
chrisbra: Thanks, adjusted.

src/popupwin.c
chrisbra: fixed.

src/popupwin.c
chrisbra: fixed.

src/popupwin.c
chrisbra: I made the function return a boolean now. So should be resolved now.

chrisbra:
made a few minor adjustments:
- let it return true/false on success/failure
- checked if the buffer is already displayed in the popup
- add new tests for opening a second terminal window in a popup
- updated documentation slightly

errael:
I noticed https://github.com/vim/vim/issues/15006#issuecomment-2171997625, which seems relevant, was not commented on. Why introduce a new function if not needed?

bfrg:
Wouldn't it be more useful to generalize the function such that it can be called for any window and not just popups? I mean, it's just a matter of time until some requests something like `win_setbuf({winid}, {bufnr})`.

chrisbra:
> Wouldn't it be more useful to generalize the function such that it can be called for any window and not just popups? I mean, it's just a matter of time until someone requests something like win_setbuf({winid}, {bufnr}).

I don't like that :) we have `:b` for that already.

chrisbra:
Copying here my comment from [#15006:](https://github.com/vim/vim/issues/15006#issuecomment-2171997625)
<hr>
> I noticed the describtion of popup_settext() explicitly limit the second arg not the buffer number, while popup_create() allowed the first arg. Would it be more unified to remove this limitation instead of creating a new interface?

Honestly, I don't understand that remark from popup_settext(). It says:

{text} is the same as supplied to |popup_create()|, except that a buffer number is not allowed.

But at popup_create() we don't have the {text} argument. I suppose it refers to the {what} of popup_create? I guess we could change popup_settext() and make {text} also take a buffer number (however not a buffer name). That could work.
<hr>
Also, I am not sure, if we should relax the possible argument types for the second arg. Now it is already '`String` or `list<string>`. Not sure if it is good design to now also allow `Number`. That seems to defeat the purpose of checking argument types in Vim9.

@yegappan what do you think?

yegappan:
> 
>> But at popup_create() we don't have the {text} argument. I suppose it refers to the {what} of popup_create? I guess we could change popup_settext() and make {text} also take a buffer number (however not a buffer name). That could work.
> >
> Also, I am not sure, if we should relax the possible argument types for the second arg. Now it is already '`String` or `list`. Not sure if it is good design to now also allow `Number`. That seems to defeat the purpose of checking argument types in Vim9.
> @yegappan what do you think?

It is better to have strict type checking to detect errors (also it helps with Vim9 code compilation).
So instead of changing the second argument to accept a number, you can either add a third
optional buffer argument or preferably a Dict argument.  But using a separate dedicated function
for changing the buffer displayed in a popup window is more cleaner.

errael:
> > > But at popup_create() we don't have the {text} argument. I suppose it refers to the {what} of popup_create? I guess we could change popup_settext() and make {text} also take a buffer number (however not a buffer name). That could work.
> > 
> > 
> > Also, I am not sure, if we should relax the possible argument types for the second arg. Now it is already '`String` or `list`. Not sure if it is good design to now also allow `Number`. That seems to defeat the purpose of checking argument types in Vim9.
> > @yegappan what do you think?
> 
> It is better to have strict type checking to detect errors (also it helps with Vim9 code compilation). So instead of changing the second argument to accept a number, you can either add a third optional buffer argument or preferably a Dict argument. But using a separate dedicated function for changing the buffer displayed in a popup window is more cleaner.

errael:
> > > But at popup_create() we don't have the {text} argument. I suppose it refers to the {what} of popup_create? I guess we could change popup_settext() and make {text} also take a buffer number (however not a buffer name). That could work.
> > 
> > 
> > Also, I am not sure, if we should relax the possible argument types for the second arg. Now it is already '`String` or `list`. Not sure if it is good design to now also allow `Number`. That seems to defeat the purpose of checking argument types in Vim9.
> > @yegappan what do you think?
> 
> It is better to have strict type checking to detect errors (also it helps with Vim9 code compilation). So instead of changing the second argument to accept a number, you can either add a third optional buffer argument or preferably a Dict argument. But using a separate dedicated function for changing the buffer displayed in a popup window is more cleaner.

Does it matter that `{what}` is widely used throughout the popup API? And in fact `{text}` is the exception? Or that the proposed function is inconsistent with the API? Currently `popup_settext()` is the only popup function that has a parameter `{text}`. Changing `{text}` to `{what}` (and then renaming all the `{what}` to `{text}` ;) ) is consistent. I'd be surprised if there is any measurable function/compilation performance difference by including bufnr. What is less confusing for the developer?

I agree that using a dict for the popup API is good, but it's too late for that, isn't it? Maybe each of the 6 functions that take `{what}` should get a companion function that takes a bufnr, and using bufnr in `{what}` should be deprecated?

I don't care that much; I'm glad there's some discussion about it. I don't see the word "API" in `:help development`, maybe something in there about best practices for API development would be good. So function overloading is considered bad practice, use dictionaries and give an error if the same parameter is specified more that once. E.g. the `{what}` parameter would have 3 possibilities in the dictionary `what_string`, `what_list_string`, `what_number` and it's an error to specify more than one of them. Is that the idea?

errael:
> E.g. the `{what}` parameter would have 3 possibilities in the dictionary `what_string`, `what_list_string`, `what_number` and it's an error to specify more than one of them. Is that the idea?

Oh, sadly with that approach there's no compile time checking like there is now with overloaded functions.

chrisbra:
>  I don't see the word "API" in :help development, maybe something in there about best practices for API development would be good. So function overloading is considered bad practice, use dictionaries and give an error if the same parameter is specified more that once

It's not only API, we need to be consistent in all areas, autocommands, functions, options, etc. This includes naming, arguments and types, documentation, messages (also for internal non-user facing functions). It's hard to get this right and hard to document what are our guidelines are for new features.

bfrg:
> I don't like that :) we have :b for that already.

Yes, but it can only be set for the _current_ window. To set the buffer of another window we need to use `win_execute(winnr, $'buffer {bufnr}')` or something like that which looks weird. And now for popups we have `popup_setbuf()`. IMO it would have been cleaner if there was one single function for both type of windows.

PR #15027, opened 2024-06-16
Termdebug8

errael:
Copy from last legacy version before vim9 port.
Rename legacy termdebug commands to Termdebug8.
With this one can more easily explore differences between legacy and vim9 termdebug and for emergency use.

Use like
```
gvim -c "set co=220" -c "packadd termdebug8" -c "Termdebug8 $cmd $cor
```

There's no expectation that this can run concurrently in same-instance with vim9 termdebug.
It uses the same configuration setup.

It's expected that this will fade as Termdebug becomes more robust.

Note: this is two commits. The first is the copy, the 2nd renames the commands.

### DIFF
--- a/Filelist
+++ b/Filelist
@@ -785,6 +785,7 @@ RT_ALL =	\
 		runtime/pack/dist/opt/shellmenu/plugin/shellmenu.vim \
 		runtime/pack/dist/opt/swapmouse/plugin/swapmouse.vim \
 		runtime/pack/dist/opt/termdebug/plugin/termdebug.vim \
+		runtime/pack/dist/opt/termdebug8/plugin/termdebug8.vim \
 
 # Runtime files for all distributions without CR/LF translation.
 RT_ALL_BIN =	\
--- a/runtime/pack/dist/opt/termdebug8/plugin/termdebug8.vim
+++ b/runtime/pack/dist/opt/termdebug8/plugin/termdebug8.vim
@@ -1,0 +1,1826 @@
+" Debugger plugin using gdb.
+"
+" Author: Bram Moolenaar
+" Copyright: Vim license applies, see ":help license"
+" Last Change: 2023 Nov 02
+"
+" WORK IN PROGRESS - The basics works stable, more to come
+" Note: In general you need at least GDB 7.12 because this provides the
+" frame= response in MI thread-selected events we need to sync stack to file.
+" The one included with "old" MingW is too old (7.6.1), you may upgrade it or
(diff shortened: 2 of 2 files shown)


jiangyinzuo:
Should we add some documentation to explain
1. the difference between termdebug and termdebug8
2. when we will deprecate temdebug8?

chrisbra:
That is the thing. What can you do with the legacy plugin, that is not possible with the current termplugin? I rather spend time to get the vim9script plugin fixed rather than having to include 2 versions of the same plugin

vim-ml:
@errael perhaps, let it be~ if no further outstanding huge change/break, let's just turn to use vim9 version of termdebug now..
// this tmp switch operation was too late.. :sweat_smile:
// thank you anyway~

-- 
shane.xb.qian

errael:
> Should we add some documentation to explain
> 
>     1. the difference between termdebug and termdebug8

At this point, a short paragraph for `Termdebug8` that references the `Termdebug`
docs would be in order. Something that basically says that `termdebug` is under
construction and that `Termdebug8` will go away when things stabalize. If needed
at some point the docs associated with legacy script `termdebug` could be copied.

But since it looks like this won't be merged...

>     2. when we will deprecate temdebug8?

There's substantial refactoring, changes, and new features proposed for `termdebug`.
I haven't seen a plan or estimate. After things stabalize, I think it could be
deprecated; or even have the command `Termdebug8` point the `Termdebug`.

errael:
> That is the thing. What can you do with the legacy plugin, that is not possible with the current termplugin?

There's substantial refactoring, changes, and new features proposed for `termdebug`.
The testing for `termdebug` seems to be weak; consider that it makes heavy use of
BufRead and a few other autloads, but it seems they are not tested.

I don't use `termdebug` frequently; I just got lucky this time fixing something
in the vim "C" code used by the vim9 port of `termdebug`. (remember CHECKME?)

When I go to use `termdebug`, I'm generally in the middle of something. Having to
stop what I'm doing and shift to debugging the debugger is less than desirable. I
don't know how many people use `termdebug` and how often or how comfortable they
might be with digging around for a working version and using that. `termdebug`
uses advanced vim features and I'm lightweight when it comes to them.

Anyway, even if this PR isn't merged, I have a changeset I can rebase and include
with my local version so I'll always have `termdebug` available when needed.

https://www.scientificamerican.com/article/einstein-s-parable-of-quantum-insanity/

> I rather spend time to get the vim9script plugin fixed rather than having to include 2 versions of the same plugin

Are you saying, with "spend time", that including this PR is an equivelent effort
to fixing vim9 `termdebug` issues?

PR #15029, opened 2024-06-17
Matched text isn't highlighted in cmdline pum

zeertzjq:
Problem:  Matched text isn't highlighted in cmdline pum.
Solution: Use cmdline completion pattern in cmdline mode.


zeertzjq:
For non-fuzzy matching this isn't always accurate, as every kind of Cmdline completion has a different kind of non-fuzzy matching procedure. This PR only uses prefix matching like what is used Insert mode completion.

Shane-XB-Qian:
perhaps better be refer this ticket with #15021

and this fuzzy of cli I recalled was impl by @yegappan so perhaps he can input something?

or I am not sure cli pum should be something like insert-compl? vs fuzzy of cli was good but fuzzy of insert-compl so far I can't completely feel that's a good idea.

--
shane.xb.qian

Shane-XB-Qian:
um.. the cli compl seems was from @ychin
let's see if there some input, or just your said way.

--
shane.xb.qian

chrisbra:
Thanks for that!

PR #15032, opened 2024-06-17
Save/restore feature refactored, added tests, updated docs. 

ubaldot:
* Save/restore feature refactored,
* Tests added,
* Variable naming consistency improved,
* Docs updated,
* Additional refactoring performed.

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -204,7 +204,6 @@ def InitScriptVariables()
 
   if has('menu')
     saved_mousemodel = &mousemodel
-    echom "mousemodel to be saved:" .. saved_mousemodel
   endif
 enddef
 # The command that starts debugging, e.g. ":Termdebug vim".
@@ -1258,10 +1257,7 @@ def DeleteCommands()
     win_gotoid(curwinid)
     # winbar_winids = []
 
-    echom "restoring mousemodel:" .. saved_mousemodel
     &mousemodel = saved_mousemodel
-    echom "mousemodel before leaving:" .. saved_mousemodel
-    echom "&mousemodel before leaving:" .. &mousemodel
     try
       aunmenu PopUp.-SEP3-
       aunmenu PopUp.Set\ breakpoint
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -45,11 +45,11 @@ enddef
 
 # Variables to keep their status among multiple instances of Termdebug
 # Avoid to source the script twice.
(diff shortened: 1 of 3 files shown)


ubaldot:
@yegappan Could you please check with your setting of `g:termdebug_config['wide']` if you get the desired result? That feature is a bit difficult to test through unit testing.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1232,38 +1223,22 @@ def DeleteCommands()
...
-    if !empty(minus_map_saved) && minus_map_saved.buffer
-      # pass
-    elseif !empty(minus_map_saved) && !minus_map_saved.buffer
-      nunmap -
-      mapset(minus_map_saved)
-    elseif empty(minus_map_saved)
-      silent! nunmap -
-    endif
-    minus_map_saved = null_dict
+  if !empty(saved_k_map)
```
chrisbra: you dropped the `!k_map_saved.buffer` intentionally here?

ubaldot: Yes. The logic now is much easier and it goes like the following:

1. At termdebug startup user-defined mapping are always saved no matter of the user settings. Maps are always saved even if they are empty (that happens in the Init function).
2. Termdebug set the mapping depending on user settings.
3. At teadown, termdebug restore the saved mapping exactly as they were before staring gdb.

The empty condition is a guard in case the `saved_k_map` is empty. In that case `mapset` would return an error.

chrisbra: But the problem is, if the mapping was a buffer-local (the `buffer` key is set) mapping, we may restore it wrongly here (because it get's re-created in the wrong buffer).

ubaldot: Correct. The maparg only stores info if the mapping was buffer-local but **not** to what buffer it relates (btw, could that provide some food for thoughts for a more general perspective?  ;-) ) 
I'll fix it. Thanks!

ubaldot: Done! Please, check it now. I also updated the related variables names with capital `K` since the mapping is for  `K` and not for `k`.

chrisbra:
thanks. I'd like that you include a new test!

chrisbra:
😆  What I was trying to say: "Thanks for adding a test for the save/restore function". But I take more tests ;)

Shane-XB-Qian:
really? the test was there but just the err was from vim itself when K mapping was none at all.
i may suggest you stop burning on this topic anymore, thank you tidy up some var naming here this ticket, but It maybe no interested to argue it too much, but/and this refactor totally actually was no adding any funcy thing, but left much argument.

--
shane.xb.qian

PR #15036, opened 2024-06-18
getcmdcompltype() interferes with cmdline completion

zeertzjq:
Problem:  getcmdcompltype() interferes with cmdline completion.
Solution: Don't set expand context when it's already set.


chrisbra:
thanks!

PR #15040, opened 2024-06-18
Add CursorMovedC autocmd

Shougo:
It is useful to emulate cmdline.


chrisbra:
Thanks. The change to syntax/vim.vim is not needed, we will however need to re-generate the vim.vim syntax script. I wonder what happens, if the autocommand changes the cursor position? Does it recursively trigger the autocommand (I hope not)?

Shougo:
>  I wonder what happens, if the autocommand changes the cursor position? Does it recursively trigger the autocommand (I hope not)?

I have tested the behavior.

If `++nested` is not set, it works without nesting call.
If `++nested` is set, it will be infinite loop.

```
Found errors in Test_Cmdline():
Caught exception in Test_Cmdline(): Vim(call):E218: Autocommand nesting too deep @ command line..script /home/shougo/src/vim/src/testdir/runtest.vim[607]..func
tion RunTheTest[57]..Test_Cmdline[108]..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"..CmdlineMoved A
utocommands for ":"..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"
..CmdlineMoved Autocommands for ":"..CmdlineMoved Autocommands for ":"
```

I think it is expected behavior.

Shougo:
I have added the test for it.

chrisbra:
thanks. I am not a big fan of adding yet another auto command... Wouldn't be CursorMovedC be the better name, that way it would be in more sync to the existing CursorMoved and CursorMovedI autocommands I would think.

Can you please back out the change to syntax/vim.vim and also please rename the auto command?

Shougo:
OK.

Shougo:
> Can you please back out the change to syntax/vim.vim and also please rename the auto command?

Done.

chrisbra:
thanks!

src/testdir/test_autocmd.vim
```diff
@@ -2096,6 +2096,21 @@ func Test_Cmdline()
...
+  au! CursorMovedC : let g:pos = getcmdpos()
+  let g:pos = 0
+  call feedkeys(":hello\<Left>\<ESC>", 'xt')
+  call assert_equal(5, g:pos)
+  call feedkeys(":12345678\<C-R>=setcmdpos(3)\<CR>\<ESC>", 'xt')
+  call assert_equal(3, g:pos)
+  au! CursorMovedC
+
+  " CursorMovedC changes the cursor position.
+  au! CursorMovedC : let g:pos = getcmdpos() | call setcmdpos(getcmdpos()-1)
```
zeertzjq, line 2109: It seems that this test doesn't really make sense, because `setcmdpos()` doesn't actually work when called in an autocommand:
```
		Only works while editing the command line, thus you must use
		|c_CTRL-\_e|, |c_CTRL-R_=| or |c_CTRL-R_CTRL-R| with '='.  For
		|c_CTRL-\_e| and |c_CTRL-R_CTRL-R| with '=' the position is
		set after the command line is set to the expression.  For
		|c_CTRL-R_=| it is set after evaluating the expression but
		before inserting the resulting text.
```

Shougo: Hm....  If so, why it is nesting?

https://github.com/vim/vim/pull/15040#issuecomment-2177285810

zeertzjq: Because you are triggering `CursorMovedC` even if cursor didn't move.

zeertzjq: It seems that the initial value of `save_cmdspos` is also wrong, as it is using the position from the previous cmdline. I'll fix these problems in #15064.

Shougo: Oh...

PR #15044, opened 2024-06-19
win32: Clean up old codes

k-takata:
* Remove very old codes for Cygwin version of GCC.  
  Nowadays Cygwin GCC cannot be used for building Win32 Vim. (The `-mno-cygwin` option was removed in Cygwin GCC4.)
* Remove old codes for old versions of MinGW.  
  Remove `__MINGW32__` as much as possible.
* Adjust makefile.


chrisbra:
Thanks. Nice cleanup!

PR #15045, opened 2024-06-19
:help execute() incorrectly specifies the return type

mgedmin:
The description clearly explains that this function returns a string.  I think the 'Number' here is a copy/paste error.


chrisbra:
obviously correct, thanks!

PR #15046, opened 2024-06-19
Reinforced mapping tests & simplified mapping restore logic.  

ubaldot:
* Tests now check that local-buffer mappings are restored in the correct buffers. 
* Mapping restoring at termdebug shutdown logic simplified.

EDIT; I am thinking that an even leaner approach would be to move some logic inside the `Init` function to clean up a bit `InstallCommands()` and `DeleteCommands()` (think for example to the logic to set the variable `map` inside `InstallCommands()`). It needs more thinking....

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1224,32 +1224,24 @@ def DeleteCommands()
   delcommand Winbar
 
 
-  if !empty(saved_K_map) && saved_K_map.buffer
-    # pass
-  elseif !empty(saved_K_map) && !saved_K_map.buffer
-    nunmap K
+  if !empty(saved_K_map) && !saved_K_map.buffer
     mapset(saved_K_map)
   elseif empty(saved_K_map)
     silent! nunmap K
   endif
 
-  if !empty(saved_plus_map) && saved_plus_map.buffer
-    # pass
-  elseif !empty(saved_plus_map) && !saved_plus_map.buffer
-    nunmap +
+  if !empty(saved_plus_map) && !saved_plus_map.buffer
     mapset(saved_plus_map)
   elseif empty(saved_plus_map)
     silent! nunmap +
   endif
 
-  if !empty(saved_minus_map) && saved_minus_map.buffer
-    # pass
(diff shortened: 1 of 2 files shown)


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1224,33 +1224,25 @@ def DeleteCommands()
...
 
-  if !empty(saved_K_map) && saved_K_map.buffer
-    # pass
-  elseif !empty(saved_K_map) && !saved_K_map.buffer
-    nunmap K
+  if !empty(saved_K_map) && !saved_K_map.buffer
     mapset(saved_K_map)
   elseif empty(saved_K_map)
     silent! nunmap K
   endif
```
Shane-XB-Qian, line 1231: this was a "comment" or remark, that is a reminder to following new coming coder to notice that difference.

Shane-XB-Qian (commented):
man, as said https://github.com/vim/vim/pull/15032#issuecomment-2176813567 and many similar tickets on this termdebug topic someone else complained in recent weeks, you donot have to burn on this topic over and over, it actually didnot benefit anything.
as for the test, let me repeat again, it was not the original test problem, but global K maybe was not there (e.g gdb window) at all, so when you quite there, the unmap will raise err, for any similar case which not only termdebug.
// so i may suggest you donot have to burn those code over and over, speically the test code maybe didnot benefit anything but maybe worse.
// but yea, thank you tidy up those var naming and init, wish those no buggy.

chrisbra:
thanks!

Shane-XB-Qian:
> > man, as said [#15032 (comment)](https://github.com/vim/vim/pull/15032#issuecomment-2176813567) and many similar tickets on this termdebug topic someone else complained in recent weeks, you donot have to burn on this topic over and over, it actually didnot benefit anything.
> 
> IMHO, getting the code as clean as possible helps maintainability. Of course you need good tests/testing to back it up.

ok, i am the one fixed buf local K mapping bug of termdebug, and corrected the recent unmap K err;
as for the tuning the code be beautiful as you guys felt, i left it to you guys, as long as no err/broken. :sweat_smile:

-- 
shane.xb.qian

PR #15048, opened 2024-06-19
runtime(java): Remove the group exclusion list from @​javaTop

zzzyxwvut:
Instances of anonymous classes can be passed as method  
arguments and should be subject to line folding as well.


chrisbra:
thanks!

PR #15055, opened 2024-06-20
fix(completion): check compl_length and regex fix

### LINKED ISSUE #15051: fuzzycollect internal error lalloc
### Steps to reproduce

1. `vim -Nu NONE`
2. `set completeopt=menuone,popup,fuzzy,fuzzycollect,noselect completepopup=highlight:Pmenu`
3. `hi PmenuMatch guifg=#ffaf5f guibg=#3a3a3a gui=NONE cterm=NONE`
4. `hi PmenuMatchSel guifg=#af5f00 guibg=#afaf87 gui=NONE cterm=NONE`
5. `:h fuzzy`
6. switch to previous window
7. type `omp` and `c-p` `c-p`

### Expected behaviour

There should be no errors.

image

[asciicast](https://asciinema.org/a/ODnO5rMO0Z4bkIiDRSq6EnFUv)

### Version of Vim

9.1.505

### Environment

debian 12, bash

### Logs and stack traces

_No response_

### LINKED ISSUE #15052: fuzzycollect "hanging" scanning tags
### Steps to reproduce

1. `vim -Nu NONE`
2. `set completeopt=menuone,popup,fuzzy,fuzzycollect,noselect completepopup=highlight:Pmenu`
3. add 2 words, e.g. `hello world`
4. in the new line goto insert mode and press `Ctrl-n`

### Expected behaviour

Words `hello` and `world` should be suggested for completion.

I can see "Scanning tags." and vim is not responing unless `Ctrl-c` is pressed.

[asciicast](https://asciinema.org/a/mUcEfRfagz1m0a7q1Hk0oxr8q)

### Version of Vim

9.1.505

### Environment

debian12, bash

### Logs and stack traces

_No response_

glepnir:
Problem: 
1. when there is no word for search  fuzzycollect will hang at scan.
2. wrong regex generate

Solution: 
1. also check compl_length.
2. use S* for word

Fix #15052 
Fix #15051

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -4508,9 +4508,9 @@ ins_compl_use_match(int c)
 get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 {
     int		i;
+    int		start;
     int		char_len;
-    size_t	fuzzy_len;
-    char_u	*fuzzy_pattern;
+    garray_T	fuzzy_ga;
 
     if ((compl_cont_status & CONT_SOL) || ctrl_x_mode_path_defines())
     {
@@ -4626,35 +4626,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 
     compl_patternlen = STRLEN(compl_pattern);
 
-    if ((get_cot_flags() & COT_FUZZYCOLLECT) != 0)
+    if ((get_cot_flags() & COT_FUZZYCOLLECT) != 0 && compl_length > 0)
     {
-	// Adjust size to avoid buffer overflow
-	fuzzy_len = (size_t)compl_length * 5 + 10;
-	// Allocate enough space
-	fuzzy_pattern = alloc(fuzzy_len);
-	if (fuzzy_pattern == NULL)
+	// Initialize the growable array
+	ga_init2(&fuzzy_ga, sizeof(char), 100);
+	ga_concat(&fuzzy_ga, (char_u *)"\\v\\S*");
+	// Adjust the start index based on different compl_pattern generation
+	if (STRNCMP(compl_pattern, "\\<\\k\\k", 6) == 0)
+	    i = 6; // Skip "\\<\\k\\k"
+	else if (STRNCMP(compl_pattern, "\\<", 2) == 0)
+	    i = 2; // Skip "\\<"
+	else
+	    i = 0; // No special prefix
+	start = i;
+	if (has_mbyte)
 	{
-	    compl_patternlen = 0;
-	    return FAIL;
+	    while (i < compl_length + start)
+	    {
+		// Get length of current multi-byte character
+		char_len = mb_ptr2len(compl_pattern + i);
+		// Move to the next character
+		i += char_len;
+	    }
+	    // Concatenate the characters safely
+	    ga_concat_len(&fuzzy_ga, compl_pattern + start, i - start);
 	}
-	// Use 'very magic' mode for simpler syntax
-	STRCPY(fuzzy_pattern, "\\v");
-	i = 2; // Start from 2 to skip "\\v"
-	while (i < compl_length + 2)
-	{
-	    // Append "\\k*" before each character
-	    STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
-	    // Get length of current multi-byte character
-	    char_len = mb_ptr2len(compl_pattern + i);
-	    // Concatenate the character safely
-	    STRNCAT(fuzzy_pattern, compl_pattern + i, char_len);
-	    // Move to the next character
-	    i += char_len;
-	}
-	// Append "\\k*" at the end to match any characters after the pattern
-	STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
+	else
+	    ga_concat_len(&fuzzy_ga, compl_pattern + start, compl_length);
+	// Append "\\S*" at the end to match any characters after the pattern
+	ga_concat(&fuzzy_ga, (char_u *)"\\S*");
 	vim_free(compl_pattern);
-	compl_pattern = fuzzy_pattern;
+	ga_append(&fuzzy_ga, NUL); // Null-terminate the string
+	compl_pattern = fuzzy_ga.ga_data;
 	compl_patternlen = STRLEN(compl_pattern);
     }
 
--- a/src/testdir/test_ins_complete.vim
+++ b/src/testdir/test_ins_complete.vim
@@ -2617,6 +2617,12 @@ func Test_complete_fuzzy_match()
   call assert_equal('one', g:word)
   call assert_equal('one two o', getline('.'))
 
+  set completeopt=fuzzycollect,menu,menuone
+  call setline(1, ['hello world', ''])
+  call cursor(2, 1)
+  call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('hello', getline('.'))
+
   " clean up
   set omnifunc=
   bw!


src/insexpand.c
```diff
@@ -4626,7 +4626,7 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
 
     compl_patternlen = STRLEN(compl_pattern);
 
-    if ((get_cot_flags() & COT_FUZZYCOLLECT) != 0)
+    if ((get_cot_flags() & COT_FUZZYCOLLECT) != 0 && compl_length > 0)
     {
 	// Adjust size to avoid buffer overflow
 	fuzzy_len = (size_t)compl_length * 5 + 10;
```
LemonBoy: Nit: you can simplify the code and get rid of this guess by using a `garray_T` to build the string.

glepnir: Speaking of which, Neovim has Stringbuilder, maybe Vim can have a wrapper too :)

src/insexpand.c
```diff
@@ -4626,35 +4625,26 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
-	{
-	    compl_patternlen = 0;
-	    return FAIL;
-	}
-	// Use 'very magic' mode for simpler syntax
-	STRCPY(fuzzy_pattern, "\\v");
+	// Initialize the growable array
+	ga_init2(&fuzzy_ga, sizeof(char), 100);
+	ga_concat(&fuzzy_ga, (char_u *)"\\v\\S*");
 	i = 2; // Start from 2 to skip "\\v"
```
LemonBoy: Where's the `\\v` inserted into `compl_pattern`? Are you sure `compl_patternlen` can never be < 2?
If so please add a comment.

src/insexpand.c
```diff
@@ -4626,35 +4625,26 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
 	    char_len = mb_ptr2len(compl_pattern + i);
 	    // Concatenate the character safely
-	    STRNCAT(fuzzy_pattern, compl_pattern + i, char_len);
+	    ga_concat_len(&fuzzy_ga, compl_pattern + i, char_len);
 	    // Move to the next character
 	    i += char_len;
 	}
-	// Append "\\k*" at the end to match any characters after the pattern
-	STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
+	// Append "\\w*" at the end to match any characters after the pattern
```
LemonBoy: The comment is not in sync with the code, `\S*` is inserted.

src/insexpand.c
```diff
@@ -4626,35 +4625,26 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
 	i = 2; // Start from 2 to skip "\\v"
 	while (i < compl_length + 2)
 	{
-	    // Append "\\k*" before each character
-	    STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
 	    // Get length of current multi-byte character
 	    char_len = mb_ptr2len(compl_pattern + i);
 	    // Concatenate the character safely
-	    STRNCAT(fuzzy_pattern, compl_pattern + i, char_len);
+	    ga_concat_len(&fuzzy_ga, compl_pattern + i, char_len);
```
LemonBoy: You can compute the number of bytes to be copied in a tight loop and then use a single `ga_concat_len` to append it, it should be quite faster..

src/insexpand.c
```diff
@@ -4626,35 +4625,26 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
+	// Initialize the growable array
+	ga_init2(&fuzzy_ga, sizeof(char), 100);
+	ga_concat(&fuzzy_ga, (char_u *)"\\v\\S*");
 	i = 2; // Start from 2 to skip "\\v"
 	while (i < compl_length + 2)
 	{
-	    // Append "\\k*" before each character
-	    STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
 	    // Get length of current multi-byte character
 	    char_len = mb_ptr2len(compl_pattern + i);
```
LemonBoy: Should this be guarded with `has_mbyte` ?

src/insexpand.c
glepnir: I tried some operations but still failed to trigger some of the branch conditions that generate compl_pattern. I added some judgments here. But I still think some stable test case are needed here ...

src/insexpand.c
```diff
@@ -4626,35 +4626,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
-	    // Append "\\k*" before each character
-	    STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
-	    // Get length of current multi-byte character
-	    char_len = mb_ptr2len(compl_pattern + i);
-	    // Concatenate the character safely
-	    STRNCAT(fuzzy_pattern, compl_pattern + i, char_len);
-	    // Move to the next character
-	    i += char_len;
+	    start = i;
+	    while (i < compl_length + 2)
```
LemonBoy: The `2` should be `start`.
The loop will stop when you have a number of characters that's >= `compl_length`, you want it to be <= `compl_length`.

src/insexpand.c
```diff
@@ -4626,35 +4626,38 @@ get_normal_compl_info(char_u *line, int startcol, colnr_T curs_col)
...
+		// Move to the next character
+		i += char_len;
+	    }
+	    // Concatenate the characters safely
+	    ga_concat_len(&fuzzy_ga, compl_pattern + start, i - start);
 	}
-	// Append "\\k*" at the end to match any characters after the pattern
-	STRNCAT(fuzzy_pattern, "\\k*", fuzzy_len - STRLEN(fuzzy_pattern) - 1);
+	else
+	    ga_concat_len(&fuzzy_ga, compl_pattern + 2, compl_length);
```
LemonBoy: Ditto, the `2` should be `start`.

LemonBoy (commented):
This pattern unwrapping/rewrapping looks pretty fragile, I'd look into directly generating the correct pattern.

LemonBoy:
The quoting logic is broken, it's either not applied at all or is broken due to the use of verymagic (check `quote_meta`).
I'm also not sure what verymagic is simplifying here, if anything it creates even more problems when unquoted patterns contain special characters.
The new regexp is wildly different than the previous one and doesn't look so "fuzzy" to me.

I'd suggest to take some time to define what "fuzzy" is and how to achieve it using regexps, then some more to write a good number of test cases that include "special" characters.

glepnir:
Yes I reproduced some of the issues. Working on fixing it. It's a bit tricky.

glepnir:
@LemonBoy could you provide some test cases which you get problem. thanks.

techntools:
This is how I am using it. Working so far.

https://github.com/vim/vim/discussions/15056

Instead of reverting, just list the cases where it fails if any.

glepnir:
thanks for information :) then i will continue work on this pr....

LemonBoy:
> could you provide some test cases which you get problem

I'm not having any problem, I'm just reviewing the code.

This small snippet generates a regexp with an unescaped dot:
```vim
set completeopt=fuzzycollect
set isi+=.
call setline(1, ['foo.'])
call feedkeys("A\<C-X>\<C-D>\<Esc>0", 'tx!')
```

I think it's quite hard if not impossible to replicate the fuzzy behaviour using regexp, one should try to modify the regexp engine to support this fuzzy matching to get some good results.

glepnir:
Perhaps only a compromise solution can be found here, such as some common text fuzzy regularization.

mg979:
@glepnir What is fuzzycollect for?

chrisbra:
https://github.com/vim/vim/blob/f0e691442005c12a7c35746048f0942f44e23a4c/runtime/doc/options.txt#L2147-L2158

mg979:
@chrisbra thanks

Is it late to do something on these lines, if you guys think it makes sense?
```vim
" enable fuzzy matching for all characters after the first, for all completion methods
set completeopt=fuzzy:1
" enable fuzzy matching for all characters after the first two, for 'file' completion only
set completeopt=fuzzy:2f
" enable fuzzy matching for all characters after the first two, for 'file', 'buffer' and 'keyword' completions
set completeopt=fuzzy:2fbn
```

Because imho:
1. fuzzy matching is more useful for certain completions (like file), but on other completions it can be annoying, so it would be good to specify the completion mode where it would be enabled
2. too many matches aren't useful, and make the thing slow, having fuzzy matching kick in after a certain number of characters are matched exactly, would be more useful

For example:
```
fuzzy:1
type 'f' -> matches words starting with 'f', no fuzzy
type 'fo' -> would match 'fo', 'fao', but not 'afo', because the first character is not part of the fuzzy pattern
type 'foo' -> would match 'foo', 'faoo', 'faoao', etc

fuzzy:2f (would match first 2 characters exactly, and would only work for 'file' completion)
type 'foa' -> would match 'foa', 'fooa', but not 'faoa', because 'fo' is matched exactly
```

glepnir:
1. number  of excatlly match can pass and set into `matchseq` 

2. about the completion mode you mentioned `fbn` etc . they are different. some default completion modes use regex to get result that's why there has fuzzycollect.   I think it's hard to do this with the current design of insexpand. Unless we refactor and remove the regular expression to get the result in the future. That would make everything easier.

mg979:
> some default completion modes use regex to get result that's why there has fuzzycollect

Is it new that regex are used for completion results? If you have exact part of the match to filter out most matches, and no regex (seems weird to me), also default word completion shouldn't be problematic with fuzzy.

glepnir:
Not new . it is how CTRL_N CTRL_P works.  generates compl_pattern and returns the result through the searchit function.

mg979:
Then maybe, for default completion, fuzzy and regex matching should be mutually exclusive: if the user chooses fuzzy the old method is disabled and the same fuzzy logic is used (as for the other methods). Fuzzy matching is a sort of simplified regex in the end, no need for both to be enabled at the same time.

glepnir:
I thought about the problem here carefully and read more code. As @LemonBoy  said, regular expression can't handle fuzzy. since I didn't read enough code before, I didn't handle the problem here well. I'm sorry. fuzzycollect will be reverted. I submitted a new PR to let fuzzy handle the default keyword completion. see #15193

PR #15057, opened 2024-06-20
Added sanity-check function, timeout parameter, g:termdebug_is_running var and changed several 0,1 to false,true. 

ubaldot:
* A sanity check function is run at startup to check if the environment is correctly set. Such a function collects few tests that were scattered around into one place.
* A timeout parameter is added given that gdb may take different times to startup. 
* A global variable g:termdebug_is_running is added. It is set to true just after the  TermdebugStartPost evennt and it set to false just after the TermdebugStopPost events.


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -259,32 +290,42 @@ def GetCommand(): list<string>
...
 
 def StartDebugCommand(bang: bool, ...args: list<string>)
-  # First argument is the command to debug, rest are run arguments.
-  StartDebug_internal({gdb_args: [args[0]], proc_args: args[1 : ], bang: bang})
-enddef
-
-
-def StartDebug_internal(dict: dict<any>)
-  if gdbwin > 0
+  if g:termdebug_is_running == true
```
chrisbra:
```suggestion
  if g:termdebug_is_running
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -259,32 +290,42 @@ def GetCommand(): list<string>
...
 def StartDebug(bang: bool, ...gdb_args: list<string>)
+  if g:termdebug_is_running == true
+    Echoerr('Terminal debugger already running, cannot run two')
+    return
+  endif
+
   InitScriptVariables()
+  if !SanityCheck()
+    return
+  endif
```
chrisbra: Those checks here:
```
  if g:termdebug_is_running == true
    Echoerr('Terminal debugger already running, cannot run two')
    return
  endif

  InitScriptVariables()
  if !SanityCheck()
    return
  endif
```
are done in `StartDebug()` and `StartDebugCommand()`. Better move it directly into `StartDebug_internal()` so those 2 functions don't duplicate logic and stay those thin wrappers around `StartDebug_internal()`

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -50,6 +50,13 @@ if exists('g:termdebug_loaded')
     finish
 endif
 g:termdebug_loaded = true
+g:termdebug_is_running = false
```
chrisbra, line 53: do we still need `running` var, now that we have the `g:termdebug_is_running` var?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Usch! You are absolutely right!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I think it is still needed since it relates to the code being debugged, not termdebug itself. Look for example at `HandleCursors()` function.

chrisbra: alright

PR #15064, opened 2024-06-21
CursorMovedC triggered at wrong pos when using setcmdpos()

zeertzjq:
Problem:  CursorMovedC triggered at wrong pos when using setcmdpos().
Solution: Remove the premature triggering.  Also don't trigger when
          cursor didn't move.


chrisbra:
thanks!

PR #15065, opened 2024-06-21
Mode message for spell completion doesn't match allowed keys

### LINKED ISSUE #29431: Insert mode spelling completion help text is wrong
## Background

When using [insert mode completion](https://github.com/neovim/neovim/blob/0e3e1e6b6d8370f1fcc9887d5cb931b131450a1c/runtime/doc/insert.txt#L607), a useful message is displayed on the command line. For instance, pressing `<C-x><C-f>` will show the text

```
-- File name completion (^F^N^P) match 1 of 43
```

Here, the `^F^N^P` part is telling you that you can press either <kbd>Ctrl</kbd>-<kbd>f</kbd>, <kbd>Ctrl</kbd>-<kbd>n</kbd>, or <kbd>Ctrl</kbd>-<kbd>p</kbd>. Pressing <kbd>Ctrl</kbd>-<kbd>f</kbd> will kind of "repeat" the original `<C-x><C-f>` action, which will move to the next item in the completion list. Using the `n` and `p` actions are like "next" and "previous", which will go up and down the list.

When using another completion mode, the help text might say `^L^N^P` or `^D^N^P` for example.

## Issue

With the spelling completion, the text is wrong. It should say `^S^N^P`, but it instead says `s^N^P`.

I assume this is becuase the code is just inserting a string corresponding to what the user typed into the message to be displayed. But in this case, following the help text will lead you astray. To move down the list, you need to press <kbd>Ctrl</kbd>-<kbd>s</kbd>, not just <kbd>s</kbd>.

### Steps to reproduce

1. Open vim with `vim -u NONE`
2. Run `:set spell` to enable spell check
3. Type `iasd<C-x>s`
4. You will see that the suggestion is working, and you can press `<C-s>` to move down the list, but the help text on the command line says `s` instead of `^S`.

### Expected behavior

You should expect the help text to match the actual behavior. The help text should say `^S^N^P` instead of `s^N^P`.

### Neovim version (nvim -v)

NVIM v0.11.0-dev

unsure

### Operating system/version

WSL

### Terminal name/version

WSLTTY

xterm-256color

### Installation

PPA

zeertzjq:
Problem:  Mode message for spell completion doesn't match allowed keys.
Solution: Use "^S" instead of "s".

fixes: neovim/neovim#29431

This matches the code in vim_is_ctrl_x_key():

	case CTRL_X_SPELL:
	    return (c == Ctrl_S || c == Ctrl_P || c == Ctrl_N);

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -58,7 +58,7 @@ static char *ctrl_x_msgs[] =
     N_(" Command-line completion (^V^N^P)"),
     N_(" User defined completion (^U^N^P)"),
     N_(" Omni completion (^O^N^P)"),
-    N_(" Spelling suggestion (s^N^P)"),
+    N_(" Spelling suggestion (^S^N^P)"),
     N_(" Keyword Local completion (^N^P)"),
     NULL,   // CTRL_X_EVAL doesn't use msg.
     N_(" Command-line completion (^V^N^P)"),
--- a/src/testdir/test_spell.vim
+++ b/src/testdir/test_spell.vim
@@ -5,6 +5,7 @@ source check.vim
 CheckFeature spell
 
 source screendump.vim
+source view_util.vim
 
 func TearDown()
   set nospell
@@ -298,6 +299,20 @@ func Test_compl_with_CTRL_X_CTRL_K_using_spell()
 
   bwipe!
   set spell& spelllang& dictionary& ignorecase&
+endfunc
+
+func Test_compl_with_CTRL_X_s()
+  new
+  set spell spelllang=en_us showmode
+  inoremap <buffer><F2> <Cmd>let g:msg = Screenline(&lines)<CR>
+
+  call feedkeys("STheatre\<C-X>s\<F2>\<C-Y>\<Esc>", 'tx')
+  call assert_equal(['Theater'], getline(1, '$'))
+  call assert_match('(^S^N^P)', g:msg)
+
+  bwipe!
+  set spell& spelllang& showmode&
+  unlet g:msg
 endfunc
 
 func Test_spellrepall()


chrisbra:
thanks!

nullromo:
Thanks for the fix! However, I fear there may be a localization issue. For example, search in `vim/src/po/nl.po` for "spelling suggestion" and you will find these lines:

```
msgid " Spelling suggestion (s^N^P)"
msgstr " spellingsuggestie (s^N^P)"
```

Now that the message has changed, I think the translations might not work properly.

I'd be happy to fix it, but I don't know how this is normally handled. Would I make a PR against vim or neovim? Do vim-based neovim patches get added manually?

zeertzjq:
> I'd be happy to fix it, but I don't know how this is normally handled.

I checked the commit history of src/po and it seems that translations usually don't get updated when a message changes, so I'm not sure if this should be done.

> Do vim-based neovim patches get added manually?

You don't need to worry about that.

nullromo:
I see. Based on commits like [this one](https://github.com/vim/vim/commit/e1051924c7ca1f9f9346daef756619d4e7fb7a83) it looks like each language and message is kind of handled individually on a case-by-base basis.

So in this case, since a message changed, each individual language might eventually get its own pull request to update that message with the proper text. Until then, all of the languages will just have wrong translations for that message. Seems like it would be better to just batch update them all at once while we're here, no?

Is there someone in charge of each language, or is it a community effort to keep the `.po` files up to date?

chrisbra:
you could ping the translators, check the author of those po files (or the Last-Translator in the files headers). We could also re-generate the po files in a separate PR. I believe we have sometimes done it.

victorhck:
hello @nullromo I would like to help, but really don't know how to do it! :)

nullromo:
I went ahead and made [a PR for it](https://github.com/vim/vim/pull/15113), just to keep things cleaner.

PR #15066, opened 2024-06-21
CI: macos 14 install gettext for tests

pheiduck:


pheiduck:
gettext is already installed

Issue #15067, opened 2024-06-21
The execution result of enum variable in `==` is incorrect in vim9script


jiangyinzuo:
### Steps to reproduce

In x1.vim
```vim
vim9script

enum Way
  Prompt,
  Terminal
endenum

var way: Way

if false
  way = Way.Prompt
  echom 'way: prompt'
elseif true
  way = Way.Terminal
  echom 'way: terminal'
else
  way = Way.Prompt
  echom 'way: prompt'
endif

if way == Way.Prompt
  echom 'way: prompt'
else
  echom 'way: terminal'
endif
```

In x2.vim
```vim
vim9script

var way: number

if false
  way = 0
  echom 'way: 0'
elseif true
  way = 1
  echom 'way: 1'
else
  way = 0
  echom 'way: 0'
endif

if way == 0
  echom 'way: 0'
else
  echom 'way: 1'
endif
```

Run `vim -U NONE --noplugin -S x1.vim` , output
```
way: terminal
way: prompt
```

Run `vim -U NONE --noplugin -S x2.vim`, output
```
way: 1
way: 1
```

In x3.vim
```vim
vim9script

enum Way
  Prompt,
  Terminal
endenum

export var way: Way

def g:Foo()
  g:way = Way.Terminal
  if g:way == Way.Prompt
    echom 'way: prompt'
  else
    echom 'way: terminal'
  endif
enddef

g:Foo()
echo g:way
```

output
```
way: prompt
enum Way.Terminal {name: Terminal, ordinal: 1}
```

### Expected behaviour

For x1.vim, output
```
way: terminal
way: terminal
```

For x2.vim, output
```
way: 1
way: 1
```

For x3.vim, output
```
way: terminal
enum Way.Terminal {name: Terminal, ordinal: 1}
```

### Version of Vim

9.1.0512

### Environment

Ubuntu 22.04

jiangyinzuo:
`==` seems buggy, but `is` seems correct.

```vim
vim9script

enum Way
  Prompt,
  Terminal
endenum

export var way: Way

def g:Foo()
  g:way = Way.Terminal
  if g:way is Way.Prompt
    echom 'way: prompt'
  else
    echom 'way: terminal'
  endif
enddef

g:Foo()
echo g:way
```

output
```
way: terminal
enum Way.Terminal {name: Terminal, ordinal: 1}
```

zzzyxwvut:
Curiously, when the example files are opened and _then_  
[...]

PR #15068, opened 2024-06-21
Change the types of some script variables in Termdebug

jiangyinzuo:
Problem: The types of some script variables in Termdebug
         can be changed for readability
Solution: Make varbufname, asmbufname, ptybufname,
          commbufname const. Change the type of `way`
          from string to`enum Way` (Yinzuo Jiang)

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -4,7 +4,7 @@ vim9script
 
 # Author: Bram Moolenaar
 # Copyright: Vim license applies, see ":help license"
-# Last Change: 2024 Jun 20
+# Last Change: 2024 Jun 22
 # Converted to Vim9: Ubaldo Tiberi <ubaldo.tiberi@gmail.com>
 
 # WORK IN PROGRESS - The basics works stable, more to come
@@ -58,9 +58,14 @@ g:termdebug_is_running = false
 command -nargs=* -complete=file -bang Termdebug StartDebug(<bang>0, <f-args>)
 command -nargs=+ -complete=file -bang TermdebugCommand StartDebugCommand(<bang>0, <f-args>)
 
+enum Way
+  Prompt,
+  Terminal
+endenum
+
 # Script variables declaration. These variables are re-initialized at every
 # Termdebug instance
-var way: string
+var way: Way
 var err: string
 
 var pc_id: number
@@ -133,16 +138,15 @@ var saved_K_map: dict<any>
 var saved_plus_map: dict<any>
 var saved_minus_map: dict<any>
 
-
 def InitScriptVariables()
   if exists('g:termdebug_config') && has_key(g:termdebug_config, 'use_prompt')
-    way = g:termdebug_config['use_prompt'] ? 'prompt' : 'terminal'
+    way = g:termdebug_config['use_prompt'] ? Way.Prompt : Way.Terminal
   elseif exists('g:termdebug_use_prompt')
-    way = g:termdebug_use_prompt
+    way = g:termdebug_use_prompt ? Way.Prompt : Way.Terminal
   elseif has('terminal') && !has('win32')
-    way = 'terminal'
+    way = Way.Terminal
   else
-    way = 'prompt'
+    way = Way.Prompt
   endif
   err = ''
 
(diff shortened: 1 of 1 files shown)


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -133,16 +138,15 @@ var saved_K_map: dict<any>
 var saved_plus_map: dict<any>
 var saved_minus_map: dict<any>
 
-
 def InitScriptVariables()
   if exists('g:termdebug_config') && has_key(g:termdebug_config, 'use_prompt')
-    way = g:termdebug_config['use_prompt'] ? 'prompt' : 'terminal'
+    way = g:termdebug_config['use_prompt'] ? Way.Prompt : Way.Terminal
   elseif exists('g:termdebug_use_prompt')
-    way = g:termdebug_use_prompt
```
jiangyinzuo, line 141: The original version seems buggy. `way` is expected to be 'prompt' or 'terminal', while `g:termdebug_use_prompt` is `1` or `0` according to `:h termdebug_use_prompt`

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -219,11 +219,11 @@ def SanityCheck(): bool
   var is_check_ok = true
   # Need either the +terminal feature or +channel and the prompt buffer.
   # The terminal feature does not work with gdb on win32.
-  if (way ==# 'prompt') && !has('channel')
+  if (way is Way.Prompt) && !has('channel')
```
jiangyinzuo, line 226: By the way, I found `==`  for enum is incorrect, it may be an internal bug in vim9script. so I use `is` in this PR. See https://github.com/vim/vim/issues/15067

ubaldot:
Thanks for contributing! :) 
Regarding the const: I thought about that and I was in front of two choices: 1) keep consistency and only declare script variables outside the Init function and define all of them inside the init function; 2) declare them as const as you are proposing - but then you have some definitions outside the Init function and some inside, one would one more thing to remember, i.e. some variables are defined outside the Init function.

Also, note that the script variables outside the Init function are "touched" only once, at script loading time. If you open/close Termdebug, they won't change since Termdebug command calls StartDebug() function, it doesn't source the whole script. On the other hand, having an Init function secure that all the Termdebug sessions are initialized in the same way. But I agree that if you would use const it won't matter being those variables... const! 

I would like the other to decide. If we go for 2), then note that some other vars that could be declared const and you may want to update this PR.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I also got confused when I saw it for the first time. But it actually works.

chrisbra:
I would be fine with the enum changes. However for the const, does it still work correctly when starting termdebug several times? Otherwise I don't really mind, I think I slightly prefer 1) but would leave the decision to @ubaldot for now

jiangyinzuo:
This PR is aimed to improve the code readability, I prefer 2) because const is a stronger constraint than var. I will continue updating this PR if we go for 2)

ubaldot:
The problem I see with 2) is the following - and that is why perhaps I decided to implement the Init function as-is instead of using `const` at script level. Imagine that for some reason we decide to make `ptybufname` user-configurable. To implement such a change, a developer would instinctively add some logic in the line around the line `const ptybufname = "debugged-program"` which is wrong. The variable `ptybufname`shall be moved inside the Init function along with all the surrounding logic. It shall not be at script level. Are we sure that will happen? IMHO, we cannot rely that all the developers are super careful and that all the reviewers immediately would spot such an issue.

Where is the problem with `const`? Well,  if the plugin is already loaded and the user decide to set `g:termdebug_config['ptybufname'] = 'MyProgram'` and then close/restart Termdebug to appreciate the desired change, he/she will see the buffer called `debugged-program` and NOT `MyProgram`. The user may get confused. In-fact,  to have the desired result, the user must unload and reload the termdebug plugin entirely, for example by closing and restarting the whole Vim (closing and restarting only termdebug would not be enough). Is that what we want? 

@jiangyinzuo Do you have any proposal to solve the two mention issues (developers and users side)?

jiangyinzuo:
> The problem I see with 2) is the following - and that is why perhaps I decided to implement the Init function as-is instead of using `const` at script level. Imagine that for some reason we decide to make `ptybufname` user-configurable. To implement such a change, a developer would instinctively add some logic in the line around the line `const ptybufname = "debugged-program"` which is wrong. The variable `ptybufname`shall be moved inside the Init function along with all the surrounding logic. It shall not be at script level. Are we sure that devs will notice that? IMHO, we cannot rely that all the developers are super careful and that all the reviewers immediately would spot such an issue.
> 
> User side: where is the problem with `const`? Well, if the plugin is already loaded and the user decide to set `g:termdebug_config['ptybufname'] = 'MyProgram'` and then close/restart Termdebug to appreciate the desired change, he/she will see the buffer called `debugged-program` and NOT `MyProgram`. The user may get confused. In-fact, to have the desired result, the user must unload and reload the termdebug plugin entirely, for example by closing and restarting the whole Vim (closing and restarting only termdebug would not be enough). Is that what we want?
> 
> @jiangyinzuo Do you have any proposal to solve the two mention issues (developers and users side)?

I get the point. We leave `ptybufname` var because it may be variables and user-configurable in the future. I have reverted the related changes.

ubaldot:
I will review the changes later on today, I just want to highlight that the singularity is not only for 'ptybufname' but for all the potential const defined variables. :)

chrisbra:
I think the change looks good now, thanks!

errael:
(Ignoring the bug about using `==` on objects)

This is a matter of style and depends on the situation, but I thought I'd bring it up for consideration. In this case, I believe `enum Way` is private to the file, so using `Prompt` instead of `Way.Prompt` is double OK. Anyway, the following is an option.
```
vim9script

enum Way
    Prompt, Terminal
endenum
const Prompt = Way.Prompt
const Terminal = Way.Terminal

var way: Way
way = Prompt

def F()
    echo way == Prompt
    echo way == Terminal
enddef
F()
```

ubaldot:
Not sure if it is relevant, but I was testing this with the latest MacVim and I got E492 on line 62 (`Prompt`). @chrisbra Is it OK?

PR #15070, opened 2024-06-21
Remove `recursive` parameter in xxx_equal functions

jiangyinzuo:
Problem:  the `recursive` parameter in xxx_equal functions can be removed
              as a code cleanup
Solution: Remove the `recursive` parameter in `dict_equal()`, `list_equal()`
              `object_equal` and `tv_equal()`. Use static variable `recursive_cnt == 0`
              to judge whether it is the first time to call `tv_equal()` (Yinzuo Jiang)

### DIFF
--- a/src/dict.c
+++ b/src/dict.c
@@ -1222,8 +1222,7 @@ dict_lookup(hashitem_T *hi)
 dict_equal(
     dict_T	*d1,
     dict_T	*d2,
-    int		ic,	    // ignore case for strings
-    int		recursive)  // TRUE when used recursively
+    int		ic)	    // ignore case for strings
 {
     hashitem_T	*hi;
     dictitem_T	*item2;
@@ -1247,7 +1246,7 @@ dict_equal(
 	    item2 = dict_find(d2, hi->hi_key, -1);
 	    if (item2 == NULL)
 		return FALSE;
-	    if (!tv_equal(&HI2DI(hi)->di_tv, &item2->di_tv, ic, recursive))
+	    if (!tv_equal(&HI2DI(hi)->di_tv, &item2->di_tv, ic))
 		return FALSE;
 	    --todo;
 	}
@@ -1275,7 +1274,7 @@ dict_count(dict_T *d, typval_T *needle, int ic)
 	if (!HASHITEM_EMPTY(hi))
 	{
 	    --todo;
-	    if (tv_equal(&HI2DI(hi)->di_tv, needle, ic, FALSE))
+	    if (tv_equal(&HI2DI(hi)->di_tv, needle, ic))
 		++n;
 	}
     }
--- a/src/typval.c
+++ b/src/typval.c
@@ -1605,8 +1605,7 @@ typval_compare_list(
     }
     else
     {
-	val = list_equal(tv1->vval.v_list, tv2->vval.v_list,
-							ic, FALSE);
+	val = list_equal(tv1->vval.v_list, tv2->vval.v_list, ic);
 	if (type == EXPR_NEQUAL)
 	    val = !val;
     }
@@ -1750,7 +1749,7 @@ typval_compare_object(
 	return OK;
     }
 
-    *res = object_equal(obj1, obj2, ic, FALSE) ? res_match : !res_match;
+    *res = object_equal(obj1, obj2, ic) ? res_match : !res_match;
     return OK;
 }
 
@@ -1787,7 +1786,7 @@ typval_compare_dict(
     }
     else
     {
-	val = dict_equal(tv1->vval.v_dict, tv2->vval.v_dict, ic, FALSE);
+	val = dict_equal(tv1->vval.v_dict, tv2->vval.v_dict, ic);
 	if (type == EXPR_NEQUAL)
 	    val = !val;
     }
@@ -1826,14 +1825,14 @@ typval_compare_func(
 	if (tv1->v_type == VAR_FUNC && tv2->v_type == VAR_FUNC)
 	    // strings are considered the same if their value is
 	    // the same
-	    val = tv_equal(tv1, tv2, ic, FALSE);
+	    val = tv_equal(tv1, tv2, ic);
 	else if (tv1->v_type == VAR_PARTIAL && tv2->v_type == VAR_PARTIAL)
 	    val = (tv1->vval.v_partial == tv2->vval.v_partial);
 	else
 	    val = FALSE;
     }
     else
-	val = tv_equal(tv1, tv2, ic, FALSE);
+	val = tv_equal(tv1, tv2, ic);
     if (type == EXPR_NEQUAL || type == EXPR_ISNOT)
 	val = !val;
     *res = val;
@@ -1988,7 +1987,7 @@ func_equal(
 	if (d1 != d2)
 	    return FALSE;
     }
-    else if (!dict_equal(d1, d2, ic, TRUE))
+    else if (!dict_equal(d1, d2, ic))
 	return FALSE;
 
     // empty list and no list considered the same
@@ -1998,7 +1997,7 @@ func_equal(
 	return FALSE;
     for (i = 0; i < a1; ++i)
 	if (!tv_equal(tv1->vval.v_partial->pt_argv + i,
-		      tv2->vval.v_partial->pt_argv + i, ic, TRUE))
+		      tv2->vval.v_partial->pt_argv + i, ic))
 	    return FALSE;
 
     return TRUE;
@@ -2013,8 +2012,7 @@ func_equal(
 tv_equal(
     typval_T *tv1,
     typval_T *tv2,
-    int	     ic,	    // ignore case
-    int	     recursive)	    // TRUE when used recursively
+    int	     ic)	    // ignore case
 {
     char_u	buf1[NUMBUFLEN], buf2[NUMBUFLEN];
     char_u	*s1, *s2;
@@ -2028,7 +2026,7 @@ tv_equal(
     // Reduce the limit every time running into it. That should work fine for
     // deeply linked structures that are not recursively linked and catch
     // recursiveness quickly.
-    if (!recursive)
+    if (recursive_cnt == 0)
 	tv_equal_recurse_limit = 1000;
     if (recursive_cnt >= tv_equal_recurse_limit)
     {
@@ -2058,13 +2056,13 @@ tv_equal(
     {
 	case VAR_LIST:
 	    ++recursive_cnt;
-	    r = list_equal(tv1->vval.v_list, tv2->vval.v_list, ic, TRUE);
+	    r = list_equal(tv1->vval.v_list, tv2->vval.v_list, ic);
 	    --recursive_cnt;
 	    return r;
 
 	case VAR_DICT:
 	    ++recursive_cnt;
-	    r = dict_equal(tv1->vval.v_dict, tv2->vval.v_dict, ic, TRUE);
+	    r = dict_equal(tv1->vval.v_dict, tv2->vval.v_dict, ic);
 	    --recursive_cnt;
 	    return r;
 
@@ -2100,7 +2098,7 @@ tv_equal(
 
 	case VAR_OBJECT:
 	    ++recursive_cnt;
-	    r = object_equal(tv1->vval.v_object, tv2->vval.v_object, ic, TRUE);
+	    r = object_equal(tv1->vval.v_object, tv2->vval.v_object, ic);
 	    --recursive_cnt;
 	    return r;
 
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -7535,7 +7535,7 @@ f_index(typval_T *argvars, typval_T *rettv)
 	{
 	    tv.v_type = VAR_NUMBER;
 	    tv.vval.v_number = blob_get(b, idx);
-	    if (tv_equal(&tv, &argvars[1], ic, FALSE))
+	    if (tv_equal(&tv, &argvars[1], ic))
 	    {
 		rettv->vval.v_number = idx;
 		return;
@@ -7568,7 +7568,7 @@ f_index(typval_T *argvars, typval_T *rettv)
     }
 
     for ( ; item != NULL; item = item->li_next, ++idx)
-	if (tv_equal(&item->li_tv, &argvars[1], ic, FALSE))
+	if (tv_equal(&item->li_tv, &argvars[1], ic))
 	{
 	    rettv->vval.v_number = idx;
 	    break;
--- a/src/list.c
+++ b/src/list.c
@@ -365,8 +365,7 @@ list_len(list_T *l)
 list_equal(
     list_T	*l1,
     list_T	*l2,
-    int		ic,	// ignore case for strings
-    int		recursive)  // TRUE when used recursively
+    int		ic)	// ignore case for strings
 {
     listitem_T	*item1, *item2;
 
@@ -386,7 +385,7 @@ list_equal(
     for (item1 = l1->lv_first, item2 = l2->lv_first;
 	    item1 != NULL && item2 != NULL;
 			       item1 = item1->li_next, item2 = item2->li_next)
-	if (!tv_equal(&item1->li_tv, &item2->li_tv, ic, recursive))
+	if (!tv_equal(&item1->li_tv, &item2->li_tv, ic))
 	    return FALSE;
     return item1 == NULL && item2 == NULL;
 }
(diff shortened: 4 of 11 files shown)


src/typval.c
```diff
@@ -2045,9 +2045,15 @@ tv_equal(
     // recursiveness quickly.
     if (!recursive)
 	tv_equal_recurse_limit = 1000;
+    // the value of "recursive" should be FALSE when calling
+    // tv_equal() for the first time
+    else if (tv_equal_recurse_limit <= -1)
+	emsg(_(e_internal_error_please_report_a_bug));
```
zeertzjq: Use `internal_error()` or `iemsg()`?

src/typval.c
```diff
@@ -2045,9 +2045,13 @@ tv_equal(
     // recursiveness quickly.
     if (!recursive)
```
LemonBoy: The fix doesn't look right, you _do_ want to recursively check the class members for equality.
`tv_equal_recurse_limit` should be initialized to its default value when `recursive_cnt` is zero, that is whenever `tv_equal` is first called from outside.

You're also missing a test.

jiangyinzuo: I have changed all the return statement to `goto theend`, below the label `theend:`,  I set `tv_equal_recurse_limit = -1;`

src/typval.c
```diff
@@ -1760,7 +1760,7 @@ typval_compare_object(
 
     for (int i = 0; i < cl1->class_obj_member_count; ++i)
 	if (!tv_equal((typval_T *)(obj1 + 1) + i,
-				 (typval_T *)(obj2 + 1) + i, ic, TRUE))
```
jiangyinzuo: @LemonBoy  The current version already has tests about object(enum) equality.

https://github.com/vim/vim/blob/7002c055d560ae0b3bb1e24ad409390a5443daad/src/testdir/test_vim9_enum.vim#L705-L725

These tests will fail if we still assign TRUE to the variable `recursive`.

src/typval.c
LemonBoy: The patch is much more complicated than it should be, replacing `if (!recursive)` with `if (recursive_cnt)` is a one-line fix.

src/typval.c
LemonBoy: Not with the patch I've suggested, the test suite runs just fine here.

src/typval.c
jiangyinzuo: > The patch is much more complicated than it should be, replacing `if (!recursive)` with `if (recursive_cnt)` is a one-line fix.


When `tv_equal` is called for the first time:
- The `recursive` is expected to be FALSE,  `if (!recursive)` will be executed.
- Initial `recursive_cnt` value is 0, `if (recursive_cnt)` will not be executed.

`if (!recursive)`  and `if (recursive_cnt)`  are 2 different behaviors..?

If we replace `if (!recursive)` with `if (recursive_cnt == 0)`, `recursive` will be unused.

It seems that we can remove the `recursive` parameter?

src/dict.c
```diff
@@ -1222,8 +1222,7 @@ dict_lookup(hashitem_T *hi)
 dict_equal(
     dict_T	*d1,
     dict_T	*d2,
-    int		ic,	    // ignore case for strings
-    int		recursive)  // TRUE when used recursively
+    int		ic)	    // ignore case for strings
```
yegappan, line 1225: Why was the recursive flag removed?  Won't this break comparison deeply nested values?

jiangyinzuo:
@yegappan In the former version, `recursive = false` is the first time to call `tv_equal()`, while `recursive = true`  is the 2nd, 3rd, ... time to call `tv_equal()`. In this patch, `recursive_cnt == 0` is the first call.

src/dict.c
```diff
@@ -1275,7 +1274,7 @@ dict_count(dict_T *d, typval_T *needle, int ic)
 	if (!HASHITEM_EMPTY(hi))
 	{
 	    --todo;
-	    if (tv_equal(&HI2DI(hi)->di_tv, needle, ic, FALSE))
```
jiangyinzuo, line 1278: For example, `dict_count()` is the first time to call `tv_equal()`, so it pass FALSE to `tv_equal()`

src/typval.c
```diff
@@ -1760,7 +1759,7 @@ typval_compare_object(
 
     for (int i = 0; i < cl1->class_obj_member_count; ++i)
 	if (!tv_equal((typval_T *)(obj1 + 1) + i,
```
jiangyinzuo: The `recursive` in the former version should be FALSE instead of TRUE, because it is the first time to call `tv_equal()`

src/typval.c
```diff
@@ -2067,19 +2066,21 @@ tv_equal(
...
 	case VAR_LIST:
 	    ++recursive_cnt;
-	    r = list_equal(tv1->vval.v_list, tv2->vval.v_list, ic, TRUE);
+	    r = list_equal(tv1->vval.v_list, tv2->vval.v_list, ic);
 	    --recursive_cnt;
 	    return r;
 
 	case VAR_DICT:
 	    ++recursive_cnt;
-	    r = dict_equal(tv1->vval.v_dict, tv2->vval.v_dict, ic, TRUE);
```
jiangyinzuo, line 2067: Here `list_equal()` and `dict_equal()` will recursively call `tv_equal()`, so `recursive` variables in the former version are TRUE.

errael:
Any idea why the result changes depending on how you run the test x1.vim? This is shown in the recent comments of #15067.

I wonder if there's a startup/initialization issue.

jiangyinzuo:
> Any idea why the result changes depending on how you run the test x1.vim? This is shown in the recent comments of #15067.
> 
> I wonder if there's a startup/initialization issue.

I guess the use of static variable `tv_equal_recurse_limit` is the reason. 
If we call `tv_equal(recursive=TRUE)` for the first time, then the result is incorrect.
If we call `tv_equal(recursive=FALSE)` for the first time, call `tv_equal(recursive=TRUE)` the second time, the result may happen to be correct.

So different startup methods result in different outputs in #15067.
https://github.com/vim/vim/blob/7002c055d560ae0b3bb1e24ad409390a5443daad/src/typval.c#L2028-L2051

errael:
FYI, There's a SEGV issue in this area. Opened #15073 to fix it (no conflict with this PR). VAR_OBJECT needs `recursive_cnt` protection.

LemonBoy:
> It seems that we can remove the recursive parameter?

Nope, you still want to decide whether to do a shallow or deep comparison. You can change the condition to `!recursive || recursive_cnt == 0` to reset the upper limit to its default value. This way you get the correct upper limit whenever a deep `tv_equal` chain is started.

You can somehow test this with the following script, adding some logging to `tv_equal` to print the recursion counters.

```vim
vim9script

class A
    var x: any
endclass

var l0: list<any> = [1]
call insert(l0, l0)
var l1: list<any> = [2]
call insert(l1, l1)

var x1 = A.new(l0)
var x2 = A.new(l1)

# This equality exhausts the upper limit.
echo x1 == x2
# At this point the upper limit is zero and the comparison bails out without even trying.
echo x1 == x2
```

chrisbra:
so does #15076 fix this issue? or is that a different issue?

jiangyinzuo:
> so does #15076 fix this issue? or is that a different issue?

The problem with issue #15067 is that we wrongly pass `recursive=TRUE` to  `tv_equal()` in `typval_compare_object()`. #15076 rewrites the related code,  maybe it can also fix the issue #15067

src/typval.c
```diff
@@ -2028,7 +2026,7 @@ tv_equal(
     // Reduce the limit every time running into it. That should work fine for
     // deeply linked structures that are not recursively linked and catch
     // recursiveness quickly.
-    if (!recursive)
+    if (recursive_cnt == 0)
```
jiangyinzuo, line 2029: In the former version,  `recursive=TRUE` appears 5 times. They are all directly/indirectly called by `tv_equal()`, which means `tv_equal()` may call itself indirectly and recursively

- 3 times are `list_equal()`, `dict_equal()` and `object_equal()` called directly in `tv_equal()`:
https://github.com/vim/vim/blob/7b29cc97d6c1450865969d08c5538a40c304593c/src/typval.c#L2059-L2069
https://github.com/vim/vim/blob/7b29cc97d6c1450865969d08c5538a40c304593c/src/typval.c#L2101-L2105

- twice in `func_equal()`. Currently, `func_equal()` is only called in `tv_equal()`.
https://github.com/vim/vim/blob/7b29cc97d6c1450865969d08c5538a40c304593c/src/typval.c#L1991-L2001

`recursive=FALSE` means that the caller of `tv_equal()` is the entry, it is the first time to call `tv_equal()` directly/indirectly.

`recursive_cnt` is a static value in `tv_equal()` and its initial value is 0. We only change the value of `recursive_cnt` in the form of the following example:
https://github.com/vim/vim/blob/7b29cc97d6c1450865969d08c5538a40c304593c/src/typval.c#L2101-L2105

When the recursive call of `tv_equal()` ends, `recursive_cnt` is 0. So I think we can replace the condition `if (!recursive)` with `if(recursive_cnt == 0)`.

jiangyinzuo:
@chrisbra  I am ready for code review again. Now this PR is aimed at cleaning up the code. The `recursive` parameter may confuse the developers and be assigned a wrong value.

yegappan:
I am adding some tests for comparing deeply recursive List/Dict values.  Please wait for that to be merged, before making this change.

ubaldot:
I am not sure if I follow what is happening here, but when it turns into equality relationships I would add the following tests: 

1) symmetry: if a=b then b=a
2) transitivity: if a=b and b=c, then a=c
3) reflexivity: every element is equal to itself, i.e., a=a

@yegappan what do you think?

errael:
> I am adding some tests for comparing deeply recursive List/Dict values. Please wait for that to be merged, before making this change.

@yegappan I've got a couple of recursive tests for objects in progress; they may be superseded by your work and not make it to a PR. I was going to use the debugger to see how recursion behaves. Do you have a way to retrieve the `recursive_cnt` and `tv_equal_recurse_limit for a test? Is there a need to set them for a test?

errael:
> I am not sure if I follow what is happening here, but when it turns into equality relationships I would add the following tests:
> 
>     1. symmetry: if a=b then b=a
> 
>     2. transitivity: if a=b and b=c, then a=c
> 
>     3. reflexivity: every element is equal to itself, i.e., a=a


I'd be surprised if `1.` and `3.` aren't already tested. Did you check?

`2.` has been discussed ad naseum, violating `2.` is just the way it is.

There's some discussion, #13458, about how this might be addressed, perhaps you'd like to contribute some comments there.

yegappan:
> I am not sure if I follow what is happening here, but when it turns into equality relationships I would add the following tests:
> 
> 1. symmetry: if a=b then b=a
> 2. transitivity: if a=b and b=c, then a=c
> 3. reflexivity: every element is equal to itself, i.e., a=a
> 
> @yegappan what do you think?

We can add the above tests.  But I am not sure in the context of comparing deeply nested List/Dict items,
whether these tests add any additional value.

yegappan:
> > I am adding some tests for comparing deeply recursive List/Dict values. Please wait for that to be merged, before making this change.
> 
> @yegappan I've got a couple of recursive tests for objects in progress; they may be superseded by your work and not make it to a PR. I was going to use the debugger to see how recursion behaves. Do you have a way to retrieve the `recursive_cnt` and `tv_equal_recurse_limit for a test? Is there a need to set them for a test?

I am adding tests for comparing deeply nested List/Dict items.  I am not adding the tests for comparing
objects.  You can go with your PR for the object comparison.

The values of internal variables `recursive_cnt` and `tv_equal_recurse_limit` are not currently available to a test.
But you can start a new Vim instance from a test to start with the default values for these variables.

errael:
> `2.` has been discussed ad naseum, violating `2.` is just the way it is.

I should have been more clear. It is only violated in some cases when `null` and `null_<type>` are involved. AFAIK, it works in all other cases.

errael:
> > > I am adding some tests for comparing deeply recursive List/Dict values. Please wait for that to be merged, before making this change.
> > 
> > 
> > @yegappan I've got a couple of recursive tests for objects in progress; they may be superseded by your work and not make it to a PR. I was going to use the debugger to see how recursion behaves. Do you have a way to retrieve the `recursive_cnt` and `tv_equal_recurse_limit for a test? Is there a need to set them for a test?
> 
> I am adding tests for comparing deeply nested List/Dict items. I am not adding the tests for comparing objects. You can go with your PR for the object comparison.
> 
> The values of internal variables `recursive_cnt` and `tv_equal_recurse_limit` are not currently available to a test. But you can start a new Vim instance from a test to start with the default values for these variables.

I haven't thought it all the way through, but I believe that the value of `tv_equal_recurse_limit` after a compare is useful for how/why the compare result was computed. One reason is that I think I can put together a test where the actual value of `==` is `false`, but `true` is returned.

ubaldot:
> > I am not sure if I follow what is happening here, but when it turns into equality relationships I would add the following tests:
> > ```
> > 1. symmetry: if a=b then b=a
> > 
> > 2. transitivity: if a=b and b=c, then a=c
> > 
> > 3. reflexivity: every element is equal to itself, i.e., a=a
> > ```
> 
> I'd be surprised if `1.` and `3.` aren't already tested. Did you check?
> 
> `2.` has been discussed ad naseum, violating `2.` is just the way it is.
> 
> There's some discussion, #13458, about how this might be addressed, perhaps you'd like to contribute some comments there.

Regarding 1.: I got a bit shocked when the following code:

```
var a = 'foo'
echo a is a
```

returned `false`. 

All in all, I would feel comfy if we "step on the giant shoulders" and try to be as compliant as possible to standard algebraic structures in mathematics, such as rings, fields, etc. That comes for free, we don't need to rethink concepts twice, especially if they are widely proven to work. For example, I would feel more comfy in using typed nulls because the underlying sets are naturally different (list differs from dicts which differ from numbers, etc), and therefore it is natural for me to think to a different "zero" element depending on the underlying set (0 kilos is a different concept than 0 meters and each of them has its own use-case). What is more important instead, is that whatever the underlying set is, its zero element shall meet at the least the following basics properties:

1. Uniqueness
2. Neutrality (i.e. it does not alter other elements when used in operations). For example, a + 0 = 0 + a = a for all a that belongs to the set A. For example, what is the result of `add(A, null_list)` or `extend(B, null_dict`)?  Does it apply to **every** list/dict? What about other operations? And what about operations defined on the other datatypes?

Perhaps the equal operator could be also carefully reviewed. To be the devil's advocate, I could argue that the expression  `a == a` is ill-defined. Am I checking the value of `a`? Its identifier? Its datatype? Its instance? The occupied memory location? ...  Yet, whatever quantity you choose, it is more important that the criteria I wrote above are met.   

If there is the plan to also review other comparison operators like "greater than/smaller than", then I would gladly leave the user to define them, with the exception of [totally ordered sets](https://en.wikipedia.org/wiki/Total_order) like `numbers`.  I am 100% confident that we will never find a common agreement on comparison operators for non-totally ordered sets like dicts or list. 

In my experience, when you skip maths, it will unavoidably fire back at some point in time, and it will have no mercy. :) 

I'll write few lines in the https://github.com/vim/vim/discussions/13458,

jiangyinzuo:
> I am adding tests for comparing deeply nested List/Dict items. I am not adding the tests for comparing objects. You can go with your PR for the object comparison.
> 
> The values of internal variables `recursive_cnt` and `tv_equal_recurse_limit` are not currently available to a test. But you can start a new Vim instance from a test to start with the default values for these variables.

@errael  @yegappan 
Should we raise an "RecursionError" instead of returning TRUE or FALSE when `==` comparison reaches the `tv_equal_recurse_limit` ?

The following is how Python deals with the nested list/dict comparison issue.
```
Python 3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = [None, 0]
>>> a[0] = a
>>> a
[[...], 0]
>>> b = [None, 0]
>>> b[0] = b
>>> a ==  b
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RecursionError: maximum recursion depth exceeded in comparison
```

errael:
> Should we raise a "RecursionError" instead of returning TRUE or FALSE when `==` comparison reaches the `tv_equal_recurse_limit` in the following lines?

I can't say for certain, but I believe returning `true` when hit the limit is how it works for structured items, `list`/`dict`, for many years. I guess `obect` is in the structured item category.

chrisbra:
thanks!

PR #15071, opened 2024-06-21
Behavior of CursorMovedC is strange

### LINKED ISSUE #15069: cursormovedc didnot work when c-w or c-u
### Steps to reproduce

1. vim -clean
2. setup cursormovedc au
3. :foo bar wee 
4. c-w

### Expected behaviour

work

### Version of Vim

v9.1.512

### Environment

linux

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Behavior of CursorMovedC is strange.
Solution: Also trigger when the cmdline has changed.

Fix #15069


zeertzjq:
This PR should not be merged if it is preferable not to trigger CursorMovedC when the cmdline has changed. https://github.com/vim/vim/issues/15069#issuecomment-2183674298

chrisbra:
thanks!

Issue #15062, opened 2024-06-20
vim.vim: Outermost _foreign_ comments after :endfun _or_ :endef are not recognised


zzzyxwvut:
### Steps to reproduce

Write the following files and open them with the `--clean`  
option applied.

Example A:
```vim
" VIM_TEST_SETUP unlet! g:vimsyn_folding

fun s:Test1() abort		" fun
    return 1
endfun				" endfun

def s:Test2(): number		# def
    return 2
enddef				# enddef [!]

fun s:Test3() abort		" fun
    fun s:DoTest3() abort	" fun
        return 3
    endfun			" endfun
    return s:DoTest3()
endfun				" endfun

def s:Test4(): number		# def
    def DoTest4(): number	# def
        return 4
    enddef			# enddef
    return DoTest4()
enddef				# enddef [!]

def s:Test5(): number		# def
    fun DoTest5() abort		" fun
        return 5
    endfun			" endfun [!]
    return DoTest5()
enddef				# enddef [!]

fun s:Test6() abort		" fun
    def s:DoTest6(): number	# def
        return 6
    enddef			# enddef [!]
    return s:DoTest6()
endfun				" endfun

for d in range(1, 6)->reverse()
    exec $'echo s:Test{d}()'
    exec $'delfunction s:Test{d}'
endfor
```

Example B:
```
vim9script
# VIM_TEST_SETUP unlet! g:vimsyn_folding

fun Test1() abort		" fun
    return 1
endfun				" endfun [!]

def Test2(): number		# def
    return 2
enddef				# enddef

fun Test3() abort		" fun
    fun s:DoTest3() abort	" fun
        return 3
    endfun			" endfun
    return s:DoTest3()
endfun				" endfun [!]

def Test4(): number		# def
    def DoTest4(): number	# def
        return 4
    enddef			# enddef
    return DoTest4()
enddef				# enddef

def Test5(): number		# def
    fun DoTest5() abort		" fun
        return 5
    endfun			" endfun [!]
    return DoTest5()
enddef				# enddef

fun Test6() abort		" fun
    def s:DoTest6(): number	# def
        return 6
    enddef			# enddef [!]
    return s:DoTest6()
endfun				" endfun [!]

for d in range(1, 6)
    exec $'echo Test{d}()'
endfor
```

As examples A and B show, the outermost _foreign_ comments,  
i.e. Vim9 comments in a legacy script _or_ legacy comments in    
a Vim9 script, are currently not highlighted as such (marked  
[...]

PR #15072, opened 2024-06-21
runtime(vim): Update base-syntax, fix function tail comments

dkearns:
Match Vim9-script comments after `:def` and `:enddef` and legacy-script comments after `:func` and `:endfunc,` in any definition context.

Highlight incorrect comment types after these commands as errors.

Fixes issue #15062.


zzzyxwvut:
Thank you!

chrisbra:
thanks!

PR #15073, opened 2024-06-21
Avoid SEGV: for object compare, use recursive_cnt.

errael:
Fix the  SEGV that results from executing the following
```vim
vim9script

class C
public var nest: C
endclass

var o1 = C.new()
o1.nest = o1

var result = o1 == o1
echo result
```


errael:
This PR has nothing to do with the issue addressed by  #15070, at least not directly. Without this PR, #15073, running the test above manually, with or without #15070, you get the SEGV. 

But the testdir test included with this PR doesn't fail without the fix from #15070, which points out that the different startup conditions, fixed by #15070, affect testdir comparison tests.

chrisbra:
thanks!

PR #15076, opened 2024-06-22
Fix obj comparison

LemonBoy:
The recursive flag in tv_equal should be set only when the caller is
tv_equal, meaning that the comparison depth is > 1. The comparison
predicates for other object types are all following this rule, except
for the object one, and that may cause some weird issues like causing
the max depth limit not to be initialized in some cases.

PR stacked on #15073


chrisbra:
thanks, can you please rebase?

src/typval.c
```diff
@@ -1758,14 +1751,7 @@ typval_compare_object(
...
 
-    for (int i = 0; i < cl1->class_obj_member_count; ++i)
-	if (!tv_equal((typval_T *)(obj1 + 1) + i,
-				 (typval_T *)(obj2 + 1) + i, ic, TRUE))
-	{
-	    *res = !res_match;
-	    return OK;
-	}
-    *res = res_match;
+    *res = object_equal(obj1, obj2, ic, FALSE) ? res_match : !res_match;
```
chrisbra, line 1754: shouldn't we pass TRUE here for recursiveness? Since that is what was done in the lines above that were removed?

errael:
Cool. So #15070 can be closed and not merged I think.

I'm thinking about two tests I'll try tomorrow. One something like
```vim
class C
var nest1: C
var nest2: C
def Init(n1: C, n2: C)
endclass

var o1 = C.new()
var o2 = C.new()
o1.Init(o1, o2)
o2.Init(o2, o1)
echo o1 == o2
```
I think that will spin until it hits the limit.

And another test that is deeply nested using unique objects
```vim
class C
var nest: C
var n: number
endclass
```
Maybe with a static member `CreateNested(depth: number)` that creates something deeply nested but with all unique objects so there will be no short circuit of comparison (which is what this new PR does with identical objects, IIUC). I'm thinking they will all be the same/equal as determined by `n` except possibly for the last one(s) in the chain, so the results of the compare can be tweaked by modifying the end of the change.

I was wondering if I should close #15073 in favor of this, I wasn't sure if/how that would work. But too late, it was just merged and you've already rebased.

How did you mechanically set this up? Did you pull #15073 into your local repo first?

src/typval.c
LemonBoy: Nope, that's the source of the problem.
The `recursive` flag is only set from whithin `tv_equal`, that's the assumption that makes the code work. If you check the other cases such as `typval_compare_list` or the dict one you'll see this parameter is false.

LemonBoy:
Writing a test is so tricky I gave up. I think you need at least two equalities, one with recursive objects to deplete the `tv_equal_recurse_limit` and another one with equal-but-not-same objects to show that no comparison is performed as a side effect of the limit being zero.

> How did you mechanically set this up? Did you pull https://github.com/vim/vim/pull/15073 into your local repo first?

I did download your diff as a patch (append .patch to the gh PR URL to download it) and then applied to my branch with `git am`.

errael:
> I'm thinking about two tests I'll try tomorrow

I should have said, _after this is merged_.

If the tests seem interesting, I can do a follow on PR.

jiangyinzuo:
> Cool. So #15070 can be closed and not merged I think.
> 
> I'm thinking about two tests I'll try tomorrow. One something like
> 
> ```viml
> class C
> var nest1: C
> var nest2: C
> def Init(n1: C, n2: C)
> endclass
> 
> var o1 = C.new()
> var o2 = C.new()
> o1.Init(o1, o2)
> o2.Init(o2, o1)
> echo o1 == o2
> ```
> 
> I think that will spin until it hits the limit.
> 
> And another test that is deeply nested using unique objects
> 
> ```viml
> class C
> var nest: C
> var n: number
> endclass
> ```
> 
> Maybe with a static member `CreateNested(depth: number)` that creates something deeply nested but with all unique objects so there will be no short circuit of comparison (which is what this new PR does with identical objects, IIUC). I'm thinking they will all be the same/equal as determined by `n` except possibly for the last one(s) in the chain, so the results of the compare can be tweaked by modifying the end of the change.
> 
> I was wondering if I should close #15073 in favor of this, I wasn't sure if/how that would work. But too late, it was just merged and you've already rebased.
> 
> How did you mechanically set this up? Did you pull #15073 into your local repo first?

IMHO,  the parameter `recursive` of all the `xxx_equal()` could be removed, including `object_equal()` in this PR, and this is what  I did in PR #15070. I can wait for this PR to be merged and update PR #15070.

src/vim9class.c
```diff
@@ -3841,6 +3841,34 @@ object_len(object_T *obj)
     return tv_to_number(&rettv);
 }
 
+/*
+ * Return TRUE when two objects have exactly the same values.
+ */
+    int
+object_equal(
```
yegappan, line 3848: This function is very similar to the typval_compare_object() function.  Can you modify typval_compare_object()
to use object_equal()?

yegappan (changes requested):

src/vim9class.c
LemonBoy: That's already done, isn't it?

errael:
> IMHO, the parameter `recursive` of all the `xxx_equal()` could be removed, including `object_equal()` in this PR, and this is what I did in PR #15070. I can wait for this PR to be merged and update PR #15070.

I see, my mistake. In that case, after this PR, #15070 is a code clean up and not a bug fix; is that right?

jiangyinzuo:
> > IMHO, the parameter `recursive` of all the `xxx_equal()` could be removed, including `object_equal()` in this PR, and this is what I did in PR #15070. I can wait for this PR to be merged and update PR #15070.
> 
> I see, my mistake. In that case, after this PR, #15070 is a code clean up and not a bug fix; is that right?

Yes.

src/typval.c
chrisbra: Ah I see thanks.

src/vim9class.c
chrisbra: Yes, typval_compare_object() uses now `object_equal()`

chrisbra:
Okay, I am convinced this is good now. Let me merge it

chrisbra:
@jiangyinzuo I have merged it. You can base your changes on top of master now. Thanks!

chrisbra:
@errael 

> How did you mechanically set this up? Did you pull https://github.com/vim/vim/pull/15073 into your local repo first?

I have the following in my `.gitconfig` to merge PRs:
```
vim-pr = "!sh -c 'nr=\"$1\" ; shift ; git fetch origin refs/pull/$nr/head:pr-$nr \"$@\"' -"
```

Then I can use `git vim-pr 15076` and git will create a local branch `pr-<nr>` with all commits, which I often then rebase into a single commit before merging (actually I do `git cherry-pick --no-commit`, so I can add the changes to version.c on top of it).

errael:
@chrisbra 
> > How did you mechanically set this up? Did you pull #15073 into your local repo first?
> 
> I have the following in my `.gitconfig` to merge PRs:
> 

Thank for the info, I've made a note for my reference that references this PR. (The general idea is useful (but I use mercurial, not git, so I'll need to rework it for my `.hgrc`.))

BTW, I manually played around this morning with fetching from github and patching with mercurial; works great. I can't believe I've never tried this before.

PR #15079, opened 2024-06-22
Fix class downcasting

### LINKED ISSUE #13244: [vim9class] cannot cast an object
### Steps to reproduce

The following fails when compiling `F2`. 
Notice that `F1` work no matter what it is called with.
Using F2 is preferable because the argument is constrained and the ` type(o) == v:t_object &&` is not needed.
The following script produces
```
999999
3
999999
Error detected while compiling :source buffer=2[34]..function <SNR>38_F2:
line    2:
E1012: Type mismatch; expected object<B> but got object<A>
```

```
vim9script

class A
endclass

class B extends A
    this.mylen: number
endclass

def F1(o: any): number
    if type(o) == v:t_object && o->instanceof(B)
        return (<B>o).mylen     # This works
    endif
    return 999999
enddef

def F2(o: A): number
    if o->instanceof(B)
        return (<B>o).mylen     # Type mismatch **during compilation**
        # The following is a workaround
        #var xxx: any = o
        #return (<B>xxx).mylen
    endif
    return 999999
enddef

var o1 = A.new()
var o2 = B.new(3)

echo F1(o1)
echo F1(o2)
echo F1(13)

disa F2
```

### Expected behaviour

No error.

### Version of Vim

9.0.1958

### Environment

ubuntu

### Logs and stack traces

_No response_

LemonBoy:
When casting an object from one class to another the target type may be a subclass (downcast) or superclass (upcast) of the source one. Upcasts require a runtime type check to be emitted.

Fixes #13244

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -10762,4 +10762,21 @@ def Test_class_object_index()
   v9.CheckScriptFailure(lines, 'E689: Index not allowed after a object: a[10] = 1', 5)
 enddef
 
+def Test_class_cast()
+  var lines =<< trim END
+    vim9script
+    class A
+    endclass
+    class B extends A
+      var mylen: number
+    endclass
+    def F(o: A): number
+      return (<B>o).mylen
+    enddef
+
+    defcompile F
+  END
+  v9.CheckScriptSuccess(lines)
+enddef
+
 " vim: ts=8 sw=2 sts=2 expandtab tw=80 fdm=marker
--- a/src/structs.h
+++ b/src/structs.h
@@ -4902,7 +4902,8 @@ typedef enum {
     WT_MEMBER,
     WT_METHOD,		// object method
     WT_METHOD_ARG,	// object method argument type
-    WT_METHOD_RETURN	// object method return type
+    WT_METHOD_RETURN,	// object method return type
+    WT_CAST,		// type cast
 } wherekind_T;
 
 // Struct used to pass the location of a type check.  Used in error messages to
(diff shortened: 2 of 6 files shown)


chrisbra:
Thanks. @yegappan are you fine with this?

errael:
Downloaded and played with it. Tried a few things, no problems. Didn't really look at the implementation.

> the target type may be a subclass (downcast) or superclass (upcast) of the source one

About nomenclature; there's a typo in the title (I think). Shouldn't the title be about `downcasting`; casting to a subclass is what failed.

LemonBoy:
> About nomenclature; there's a typo in the title (I think). Shouldn't the title be about downcasting; casting to a subclass is what failed.

Indeed, please remember to fix the typo in the commit message before merging.

src/testdir/test_vim9_class.vim
```diff
@@ -10762,4 +10762,21 @@ def Test_class_object_index()
...
+      var mylen: number
+    endclass
+    def F(o: A): number
+      return (<B>o).mylen
+    enddef
+
+    defcompile F
+  END
+  v9.CheckScriptSuccess(lines)
+enddef
```
yegappan, line 10780: It will be good to add a disassemble test for this case to test_vim9_disassemble.vim.

yegappan (approved):

yegappan:
> Thanks. @yegappan are you fine with this?

The change looks good to me.  It will be good to add a disassemble test for this.

yegappan (approved):

chrisbra:
thanks!

PR #15082, opened 2024-06-22
Fix hang of `string(object)` when object has recursive references.

### LINKED ISSUE #15080: [vim9class] `echo obj` puts vim into infinite loop requiring force kill
### Steps to reproduce

```vim
vim9script

class C
    var nest1: C
    var nest2: C
    def Init(n1: C, n2: C)
        this.nest1 = n1
        this.nest2 = n2
    enddef
endclass

var o1 = C.new()
var o2 = C.new()
o1.Init(o1, o2)
o2.Init(o2, o1)

# The following outputs true
#var result = o1 == o2
#echo result

# the following put's vim into an infinite loop

echo o1
```

### Expected behaviour

If `o2.Init(o2, o1)` is removed, then the echo gets `E724: Variable nested too deep for displaying`.

Expect `E724:` also for the case that goes into an infinite loop.

### Version of Vim

9.1.514

### Environment

linux/gtk3

### Logs and stack traces

_No response_

errael:
Fix #15080.

And add a hang test and a compare test (they look very similar).
There's a significant whitespace difference in `vim9class.c`.

Similar to the recent `object_equal()` issue. `object2string` was called directly from
`echo_string_core`. Patterned this change from `dict_tv2string` usage.

@LemonBoy @yegappan object_tv2string is getting `numbuf`, `echo_style`,
`composite_val`, unlike `dict_tv2string`. That keeps the arguments to
`object2string` unchanged. Also in `object_tv2string` I didn't think the case
```
else if (copyID != 0 && obj->obj_copyID == copyID
```
would be doing anything, but now the failing case outputs
```
object of C {nest1: object of C {...}, nest2: object of
    C {nest1: object of C {...}, nest2: object of C {...}}}
```
This case was that last thing implemented.

If the case is commented out, then the test outputs
```
E724: Variable nested too deep for displaying
```
which is the situation that caused the hang.

(2nd crash in two days, might be a record for me :) )

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -6139,6 +6139,58 @@ class_tv2string(typval_T *tv, char_u **tofree)
 }
 
 /*
+ * Return a textual representation of an Object in "tv".
+ * If the memory is allocated "tofree" is set to it, otherwise NULL.
+ * When "copyID" is not zero replace recursive object with "...".
+ * When "restore_copyID" is FALSE, repeated items in the object are
+ * replaced with "...".  May return NULL.
+ */
+    static char_u *
+object_tv2string(
+    typval_T	*tv,
+    char_u	**tofree,
+    int		copyID,
+    int		restore_copyID,
+    char_u	*numbuf,
+    int		echo_style,
+    int		composite_val)
+{
+    char_u	*r = NULL;
+
+    object_T	*obj = tv->vval.v_object;
+    if (obj == NULL || obj->obj_class == NULL)
+    {
+	*tofree = NULL;
+	r = (char_u *)"object of [unknown]";
+    }
+    else if (copyID != 0 && obj->obj_copyID == copyID
+            && obj->obj_class->class_obj_member_count != 0)
+    {
+	int n = 25 + strlen((char*)obj->obj_class->class_name);
+	r = alloc(n);
+	if (r != NULL)
+	    (void)vim_snprintf((char*)r, n, "object of %s {...}",
+						obj->obj_class->class_name);
+	*tofree = r;
+    }
+    else
+    {
+	int old_copyID;
+	if (restore_copyID)
+	    old_copyID = obj->obj_copyID;
+
+	obj->obj_copyID = copyID;
+	*tofree = object2string(obj, numbuf, copyID, echo_style,
+				restore_copyID, composite_val);
+	if (restore_copyID)
+	    obj->obj_copyID = old_copyID;
+	r = *tofree;
+    }
+
+    return r;
+}
+
+/*
  * Return a string with the string representation of a variable.
  * If the memory is allocated "tofree" is set to it, otherwise NULL.
  * "numbuf" is used for a number.
@@ -6169,7 +6221,7 @@ echo_string_core(
 	{
 	    // Only give this message once for a recursive call to avoid
 	    // flooding the user with errors.  And stop iterating over lists
-	    // and dicts.
+	    // and dicts and objects.
 	    did_echo_string_emsg = TRUE;
 	    emsg(_(e_variable_nested_too_deep_for_displaying));
 	}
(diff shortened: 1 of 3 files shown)


errael:
Rebased, last vim crash I know of

@chrisbra @yegappan @LemonBoy  Hope someone can take a look and get this merged. 

Can do side by side comparisons of
```
object_tv2string dict_tv2string
   object2string dict2string
```

chrisbra:
Thanks. @yegappan are you fine with this?

src/vim9class.c
```diff
@@ -3889,47 +3891,58 @@ object2string(
...
-		ga_concat(&ga, echo_string_core(
-			    (typval_T *)(obj + 1) + i,
-			    &tf, numbuf, copyID, echo_style,
-			    restore_copyID, composite_val));
-		vim_free(tf);
+		ok = FAIL;
+		break;
 	    }
-	    ga_concat(&ga, (char_u *)"}");
+	    line_breakcheck();
```
yegappan, line 3936: Why is the line_breakcheck() needed here?  This may process other pending events that may change the value that is displayed here.

errael: Look at a side by side of this function, `object2string` with `dict2string`, that's why it is there. In fact, when doing my last check before opening the PR, I put this in, without knowing exactly why it's there.

I'm happy to take it out, LMK.

@yegappan Another thing I was unsure of, in dict_tv2string there is
```c
    else if (copyID != 0 && tv->vval.v_dict->dv_copyID == copyID
	    && tv->vval.v_dict->dv_hashtab.ht_used != 0)
```
which I converted as
```c
    else if (copyID != 0 && obj->obj_copyID == copyID
            && obj->obj_refcount != 0)
```
But I think the check against `refcount` is bogus and can/should just be taken out. Do you agree?

errael: Ahh, instead if that bogus `refcount` check, the following is probably the equivalent check
```
    else if (copyID != 0 && obj->obj_copyID == copyID
            && obj->obj_class->class_obj_member_count != 0)
```

errael:
BTW, when the bug this issue addresses is encountered, a force kill is required. I'm guess that is because there some kind of nested signal handling going on, IDK.

errael:
@yegappan  Changed `&& obj->obj_refcount != 0` to `&& obj->obj_class->class_obj_member_count != 0` as discussed above.

Any further comment about `line_breakcheck()`?

src/vim9class.c
yegappan: Yes.  The check for `obj->obj_class->class_obj_member_count` is the correct check.

yegappan:
> @yegappan Changed `&& obj->obj_refcount != 0` to `&& obj->obj_class->class_obj_member_count != 0` as discussed above.
> 
> Any further comment about `line_breakcheck()`?

We can keep the `line_breakcheck()` call to be consistent with `dict2string()`.

yegappan (approved):

errael:
Rebased, only docs have changed.

chrisbra:
thanks

src/eval.c
```diff
@@ -6138,6 +6138,58 @@ class_tv2string(typval_T *tv, char_u **tofree)
...
+    object_T	*obj = tv->vval.v_object;
+    if (obj == NULL || obj->obj_class == NULL)
+    {
+	*tofree = NULL;
+	r = (char_u *)"object of [unknown]";
+    }
+    else if (copyID != 0 && obj->obj_copyID == copyID
+            && obj->obj_class->class_obj_member_count != 0)
+    {
+	int n = 25 + strlen((char*)obj->obj_class->class_name);
```
LemonBoy, line 6169: This should be a `size_t` rather than an `int`, MSVC emits a warning otherwise.
Nit: `char*` should be `char *`, ditto for the other one in the `snprintf` below.

errael: > This should be a `size_t` rather than an `int`, MSVC emits a warning otherwise. Nit: `char*` should be `char *`, ditto for the other one in the `snprintf` below.

Thanks, #15187

PR #15083, opened 2024-06-22
runtime(java): Compose "g:java_highlight_signature" and "g:java_highlight_functions"

zzzyxwvut:
With the variables defined, distinctly highlight parts of  
a method declaration header: its name and parameter list  
parens, from its type parameters, return type, and formal  
parameters; and distinctly highlight parts of a lambda  
expression: its parameter list parens and the arrow, from  
its formal parameters and identifiers.


chrisbra:
thanks!

PR #15085, opened 2024-06-23
Fix SEGV in `object_equal()`.

errael:
@LemonBoy @yegappan Simple Fix 

@chrisbra This probably conflicts with #15082. Hopefully I'll quickly get to the one you don't check in for rebase/fixup...

(3 crashes in 2 days, got to be a new record :) ) And I've got a couple tests I'm still working on. 

If you feel like crashing vim run
```
vim9script

class C
    public var nest: C
endclass
var o1 = C.new()
var o2 = C.new(C.new())

var result = o1 == o2
```

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -10502,6 +10502,20 @@ def Test_Object_Compare_With_Recursive_Class_Ref()
     assert_equal(true, result)
   END
   v9.CheckScriptSuccess(lines)
+
+  lines =<< trim END
+    vim9script
+
+    class C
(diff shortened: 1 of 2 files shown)


yegappan (approved):

chrisbra:
thanks, makes sense

LemonBoy:
I guess it's `o1` that's NULL here, isn't it?
Either way LGTM too.

errael:
> I guess it's `o1` that's NULL here, isn't it?

`o1` is the short chain in the `vim9script`; not sure about the "C" code.

PR #15086, opened 2024-06-23
Refactored StartDebug_term and EndDebug functions.

ubaldot:
* Functions are way too long. Readability and maintainability should be slightly improved.
* Some variables are re-assigned to their initial value at teardown. This should not be needed since all internal variables are re-initialized at startup of every Termdebug session.


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,12 +1295,8 @@ def DeleteCommands()
   endif
 
   sign_unplace('TermDebug')
-  breakpoints = {}
-  breakpoint_locations = {}
-
```
chrisbra: Those should be kept, no?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1306,12 +1295,8 @@ def DeleteCommands()
...
   sign_undefine('debugPC')
   sign_undefine(BreakpointSigns->map("'debugBreakpoint' .. v:val"))
-  BreakpointSigns = []
```
chrisbra, line 1314: That should be kept, no?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -607,15 +621,16 @@ def StartDebug_prompt(dict: dict<any>)
 
   ch_log($'executing "{join(gdb_cmd)}"')
   gdbjob = job_start(gdb_cmd, {
-    exit_cb: function('EndPromptDebug'),
+    # exit_cb: function('EndPromptDebug'),
```
chrisbra: can be removed

chrisbra:
Seems fine for me

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I am not 100% sure, I may be missing something. :) What is the problem in leaving few script variable set when they will be reset anyway during the startup of the next Termdebug session? Do we really need to reset them to empty values at teardown? Why only these vars and not ALL the vars? (I am sure I am missing something though).

ubaldot: However, I re-added them.

chrisbra: True, I guess we don't neccessarily need this.

PR #15087, opened 2024-06-23
Fix freeze on macos during big text paste

yuraaka:
Environment
> VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Apr 13 2024 12:28:29)
macOS version - arm64
Included patches: 1-2136, 2142

Problem: When I paste more than 5890 chars text via Cmd+V in insert mode vim freezes indefinitely.

Steps to reproduce:
1. Run vim on new file
2. Copy somewhere >5890 chars text
3. Enter insert mode in vim
4. Paste via Cmd+V
5. Find vim frozen and a console not responsible

P.S. Don't know how to write UT for such situation.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -4447,7 +4447,7 @@ bracketed_paste(paste_mode_T mode, int drop, garray_T *gap)
     for (;;)
     {
 	// When the end is not defined read everything there is.
-	if (end == NULL && vpeekc() == NUL)
+	if (vpeekc() == NUL)
 	    break;
 	do
 	    c = vgetc();


chrisbra:
I suppose this is on the terminal? What is end set to in your case? what is `:set t_PE?`

YuraAka:
> I suppose this is on the terminal? What is end set to in your case? what is `:set t_PE?`

Afaik, end is "\033[201~"

chrisbra:
And on what terminal? Does that terminal support bracketed paste?

YuraAka:
> And on what terminal? Does that terminal support bracketed paste?

I've tested on builtin macos Terminal.app and in iTerm2. Don't know about bracketed paste, how can I check it?

chrisbra:
I don't know. The manual of your terminal should know it. Does it work, if you `:set t_PE=` ?

YuraAka:
> I don't know. The manual of your terminal should know it. Does it work, if you `:set t_PE=` ?

Yes, it works, with additional '^[[201~' at the end.

chrisbra:
> Yes, it works, with additional '^[[201~' at the end.

That's strange. That indicates bracketed paste should actually work for you.

chrisbra:
I don't think this patch is right. So I am not including this.

PR #15088, opened 2024-06-23
python: fix compatibility with Python 3.13

yselkowitz:
_PyObject_CallFunction_SizeT was dropped in Python 3.13:

https://github.com/python/cpython/pull/105051


k-takata:
Does this affect when the limited API is used?

yselkowitz:
PyObject_CallFunction is listed as part of the Limited API according to https://docs.python.org/3/c-api/stable.html#contents-of-limited-api but perhaps I didn't understand your question.

zdohnal:
Hi @k-takata ,

what do you mean by "VIm compiled with limited API" - do you mean "dynamic" configure option value for --enable-python3interp? If you do, I would say the patch preserves using the old API for older Pythons and uses the new one for 3.13.
Please let us know how we could change the patch - VIm is currently not building in Fedora Rawhide, because new Python 3.13 came into the compose, so we would like to help with the fix.

PR #15089, opened 2024-06-24
In builtin overview specify {buf} as param for appendbufline.

errael:
Get the overview and Detail to match. `{buf}` is used every where for this


zeertzjq:
This change seems needed for `setbufline()` as well.

zeertzjq:
Also, the "Last change" date at the top of the file should be updated.

errael:
> This change seems needed for `setbufline()` as well.

Thanks

chrisbra:
thanks

errael:
@chrisbra What are you doing here? **Go play and enjoy**.

PR #15091, opened 2024-06-24
Added Deprecation Warnings

ubaldot:
As discussed, here is a proposal for introducing Deprecation Warnings. 
Let's see if it could be a viable solution for handling old stuff without creating too much unhappy users. 
I took the freedom to set a deadline 1 Jan 2025, but we can discuss on possible time-windows of course.


chrisbra:
thanks!

PR #15094, opened 2024-06-24
Correctly assign the member declared types

### LINKED ISSUE #15090: [vim9class] incorrect type checking modifying list in class from outside class
### Steps to reproduce

Run
```
vim9script

class Conflict
    var l: list<string> = []
endclass

var o = Conflict.new()
echo o.l->add(123)
```
and see output
```
[123]
```

### Expected behaviour

Should get an error like
```
E1013: Argument 2: type mismatch, expected string but got number
```

### Version of Vim

9.1.516

### Environment

linux/gtk3

### Logs and stack traces

_No response_

LemonBoy:
Ensure the generated typevals get the declared type and generate typechecks to ensure that's the case.

Closes #15090

@errael Mind giving this a try?

### DIFF
--- a/src/vim9compile.c
+++ b/src/vim9compile.c
@@ -3315,7 +3315,7 @@ obj_constructor_prologue(ufunc_T *ufunc, cctx_T *cctx)
 		// the initialization expression type.
 		m->ocm_type = type;
 	    }
-	    else if (m->ocm_type->tt_type != type->tt_type)
+	    else
 	    {
 		// The type of the member initialization expression is
 		// determined at run time.  Add a runtime type check.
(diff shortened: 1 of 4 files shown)


src/vim9compile.c
```diff
@@ -3330,6 +3330,15 @@ obj_constructor_prologue(ufunc_T *ufunc, cctx_T *cctx)
...
 	    push_default_value(cctx, m->ocm_type->tt_type, FALSE, NULL);
 
+	if ((m->ocm_type->tt_type == VAR_DICT
+		    || m->ocm_type->tt_type == VAR_LIST)
+		&& m->ocm_type->tt_member != NULL
+		&& m->ocm_type->tt_member != &t_any
+		&& m->ocm_type->tt_member != &t_unknown)
+	    // Set the type in the list or dict, so that it can be checked,
+	    // also in legacy script.
+	    generate_SETTYPE(cctx, m->ocm_type);
```
yegappan, line 3340: Can you add a test to test_vim9_disassemble.vim for this instruction?

errael:
Tried it, no problems.

I got sucked into looking at the code trying to understand why it wasn't working before this change.
I guess the assumption was that only compiled code would need the type information.
I suppose the additional instruction  in the constructor prologue is worth it to avoid data corruption,
it looks like SETTYPE may traverse the data structure. I guess it's cheap in the normal/simple cases, an empty container.
Wonder if it could be done lazily; something for the future compiler optimization work ;-)

yegappan (approved):

yegappan:
Thanks for adding the Vim9 disassemble test.  Looks good to me.

chrisbra:
thanks!

PR #15096, opened 2024-06-24
runtime(vim): Update base-syntax, improve :match command highlighting

dkearns:
Match group and pattern arguments to :match commands.


chrisbra:
thanks!

PR #15102, opened 2024-06-24
Right release selects immediately when pum is truncated

### LINKED ISSUE #15101: mousemodel=popup not working properly with many items on PopUp menu
### Steps to reproduce

Create a script that turns mouse functionality on, sets `mousemodel` to `popup`, and adds many items to the `PopUp` menu.

```vim
" example.vim

set mouse=a
set mousemodel=popup

silent! unmenu PopUp
for i in range(1000)
  execute 'menu PopUp.' .. i .. ' <nop>'
endfor
```

Run Vim and source the script.

```vim
vim -clean
:source example.vim
```

Then right-click somewhere towards the top of the running Vim.

### Expected behaviour

The menu pops up briefly and then goes away. Holding down the right mouse button results in the menu staying open, but the menu doesn't expand down, even though there is more room for the menu in that direction.

I expected for the menu to popup the same way as it would if there were fewer items (staying open), perhaps with a scrollbar (similar to what the insert completion menu has when there are too many items to fit on screen), and/or expanding in the direction with more room (down in the running example).

### Version of Vim

9.1

### Environment

Operating system: Kubuntu 24.04
Terminal: Konsole 23.08.5
$TERM: xterm-256color
Shell: bash 5.2.21(1)

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Right release selects immediately when pum is truncated.
Solution: Use pum_height instead of pum_size when checking click row.
          Don't place it above mouse row when there is more space below.

Fix #15101

### DIFF
--- a/src/popupmenu.c
+++ b/src/popupmenu.c
@@ -1332,9 +1332,10 @@ pum_set_event_info(dict_T *dict)
     static void
 pum_position_at_mouse(int min_width)
 {
-    if (Rows - mouse_row > pum_size)
+    if (Rows - mouse_row > pum_size || Rows - mouse_row > mouse_row)
     {
-	// Enough space below the mouse row.
+	// Enough space below the mouse row,
+	// or there is more space below the mouse row than above.
 	pum_row = mouse_row + 1;
 	if (pum_height > Rows - pum_row)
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

PR #15105, opened 2024-06-25
fix(completion): wrong select when cp_score is zero

### LINKED ISSUE #15095: Fuzzy completion does not allow cycle back to original, wrong highlight, and Ctrl-P Ctrl-N issues
### Steps to reproduce

1. `vim -Nu NONE`
2. `:set completeopt+=fuzzy`
3.  type "hello helio hero" in the first line
4. type "he" in the second line, and then press `<C-X><C-P>`
This happens: 
image

### There are 3 separate issues:

1. the menu highlight is at "hello" but the completion is "hero"
2. repeatedly pressing `<C-P>` gets stuck at the first selection and does not cycle back
3. repeatedly pressing `<C-N>` does cycle back to the first option but does not cycle back to the original. In other word, this screen never happens:
image

### Expected behaviour

1, 2, 3 should work as if `fuzzy` is not set

### Version of Vim

Vim 9.1 patch 1-516

### Environment

MacOS, Iterm2, Tmux

### Logs and stack traces

_No response_

glepnir:
Problem: when match array build from keyword that mean leader is NULL and cpscore both zero the selected item is not correct

Solution:  when in fuzzy and shown_match_ok is false need find the compl then update cur.

FIx #15095


glepnir:
when i working on this issue fix .. i think i have find a better to do keyword fuzzy completion we don' need fuzzy collect. regex pattern can not do fuzzy. I will submit a pr recently.

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
...
   call assert_equal('hello help hero h', getline('.'))
 
+  " issue #15095 wrong select when leader is NULL and score both 0.
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<Esc>0", 'tx!')
+  call assert_equal('hello helio hero    hero', getline('.'))
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<C-P>\<C-P>\<Esc>0", 'tx!')
+  call setline(1, ['hello helio hero    hello'])
```
zeertzjq: Do you mean to use `assert_equal()` instead of `setline()` here?

glepnir: oops sry wrong paste..

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
...
   call assert_equal('hello help hero h', getline('.'))
 
+  " issue #15095 wrong select when leader is NULL and score both 0.
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<Esc>0", 'tx!')
+  call assert_equal('hello helio hero    hero', getline('.'))
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<C-P>\<C-P>\<Esc>0", 'tx!')
+  call assert_equal(1, ['hello helio hero    hello'], getline('.'))
```
zeertzjq:
```suggestion
  call assert_equal('hello helio hero    hello', getline('.'))
```

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
   call feedkeys("A\<C-X>\<C-N>\<Esc>0", 'tx!')
   call assert_equal('hello help hero h', getline('.'))
 
+  " issue #15095 wrong select when leader is NULL and score both 0.
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<Esc>0", 'tx!')
```
zeertzjq, line 2592:
```suggestion
  call feedkeys("A\<C-X>\<C-P>\<Esc>", 'tx')
```

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
...
+  call assert_equal('hello helio hero    hero', getline('.'))
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<C-P>\<C-P>\<Esc>0", 'tx!')
```
zeertzjq, line 2595:
```suggestion
  call feedkeys("A\<C-X>\<C-P>\<C-P>\<C-P>\<Esc>", 'tx')
```

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
...
+  call assert_equal(['hello helio hero    hello'], getline('.'))
```
zeertzjq:
```suggestion
  call assert_equal('hello helio hero    hello', getline('.'))
```

src/testdir/test_ins_complete.vim
```diff
@@ -2586,6 +2586,15 @@ func Test_complete_fuzzy_match()
...
 
+  " issue #15095 wrong select when leader is NULL and score both 0.
+  set completeopt=fuzzy,menu
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<Esc>0", 'tx!')
+  call assert_equal('hello helio hero    hero', getline('.'))
+  call setline(1, ['hello helio hero    h'])
+  call feedkeys("A\<C-X>\<C-P>\<C-P>\<C-P>\<Esc>0", 'tx!')
+  call assert_equal('hello helio hero    hello', getline('.'))
+
```
zeertzjq: It seems that this test passes even without the fix in this PR.

src/testdir/test_popup.vim
```diff
@@ -1482,6 +1482,17 @@ func Test_pum_highlights_match()
...
   call VerifyScreenDump(buf, 'Test_pum_highlights_09', {})
 
+  " issue #15095 wrong select
+  call term_sendkeys(buf, "\<ESC>:set completeopt=fuzzy,menu\<CR>")
+  call TermWait(buf)
+  call term_sendkeys(buf, "S hello helio hero h\<C-X>\<C-P>")
+  call TermWait(buf, 50)
+  call VerifyScreenDump(buf, 'Test_pum_highlights_10', {})
+
+  call term_sendkeys(buf, "S hello helio hero h\<C-X>\<C-P>\<C-P>")
```
zeertzjq: Shouldn't there be an `<ESC>` before this `S`?

glepnir: yup.. thanks

chrisbra:
thanks!

PR #15109, opened 2024-06-25
Support for the "arity" argument in get()

### LINKED ISSUE #15097: Add `get({func}, 'arity')` to determine the number of expected arguments to a funcref

There appears to be no way to get the expected arity a function (builtin or user-defined). I have some functional code that I want to add which would work better if I could change how it works based on the number of expected arguments. I’ve called it `arity` because that's what Ruby, Elixir, and other languages with functional behaviours call it that.

```vim
export def All(expr: any, fn: any): bool
  if expr->empty()
    return true
  endif

  var Fn = Ref(fn)

  if Fn->type() == v:t_func
    var arity = Fn->get('arity')->len()

    if arity == 1
      if expr->type() == v:t_dict
        for [k, v] in Items(expr)
          if !Fn([k, v]) | return false | endif
        endfor
      else
        for v in expr->deepcopy()
          if !Fn(v) | return false | endif
        endfor
      endif
    elseif arity == 2
      for [k, v] in Items(expr)
        if !Fn(k, v) | return false | endif
      endfor
    else
      throw 'Invalid argument count for ' .. string(Fn)
    endif

    return true
  elseif Fn->type() == v:t_string
    return None(expr, printf('!(%s)', Fn))
  endif

  throw 'Invalid type ' .. typename(Fn) .. ' for ' .. string(Fn)
enddef
```

An additional `{what}` for `get({func}, {what})`, called `"arity"`. The specification would be similar to that of Ruby:

> Returns an indication of the number of arguments accepted by a method. Returns a nonnegative integer for methods that take a fixed number of arguments. For Ruby methods that take a variable number of arguments, returns -n-1, where n is the number of required arguments. Keyword arguments will be considered as a single additional argument, that argument being mandatory if any keyword argument is mandatory. For methods written in C, returns -1 if the call takes a variable number of arguments.

Thus:

```vim
def Func() # => arity 0
def Func(a: any) #=> arity 1
def Func(a: any, b: any) # => arity 2
def Func(a: any, b: any = null) # => arity -2
def Func(a: any, ...b: list<any>) # => arity -2
```

If preferred, this could be replaced with multiple calls or return a dict with multiple values:

```vim
def Func() # => {required: 0, optional: 0, varargs: false}
def Func(a: any) # => {required: 1, optional: 0, varargs: false}
def Func(a: any, b: any) # => {required: 2, optional: 0, varargs: false}
def Func(a: any, b: any = null) # => {required: 1, optional: 1, varargs: false}
def Func(a: any, ...b: list<any>) # =. {required: 1, optional: 0, varargs: true}
```

I primarily care about *required* arguments for my use case, but optional and varargs may be useful for other cases.

For `vim9script` it may also be useful to be able to get the argument *types* on reflection, but that's probably a different feature altogether.

Unless this is implemented, I will be implementing a fixed argument list requirement, which I would rather not do, because the used code reads better without the extra argument if not required.

LemonBoy:
Add a way to extract the number of required and optional arguments for a given function or partial.

Closes #15097

### DIFF
--- a/src/testdir/test_getvar.vim
+++ b/src/testdir/test_getvar.vim
@@ -142,20 +142,29 @@ func Test_get_func()
   let l:F = function('tr')
   call assert_equal('tr', get(l:F, 'name'))
   call assert_equal(l:F, get(l:F, 'func'))
+  call assert_equal({'required': 3, 'optional': 0, 'varargs': v:false},
+      \ get(l:F, 'arity'))
 
   let Fb_func = function('s:FooBar')
   call assert_match('<SNR>\d\+_FooBar', get(Fb_func, 'name'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:false},
+      \ get(Fb_func, 'arity'))
   let Fb_ref = funcref('s:FooBar')
   call assert_match('<SNR>\d\+_FooBar', get(Fb_ref, 'name'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:false},
+      \ get(Fb_ref, 'arity'))
 
   call assert_equal({'func has': 'no dict'}, get(l:F, 'dict', {'func has': 'no dict'}))
   call assert_equal(0, get(l:F, 'dict'))
   call assert_equal([], get(l:F, 'args'))
+  call assert_equal({'required': 3, 'optional': 0, 'varargs': v:false}, get(l:F, 'arity'))
+
   let NF = test_null_function()
   call assert_equal('', get(NF, 'name'))
   call assert_equal(NF, get(NF, 'func'))
   call assert_equal(0, get(NF, 'dict'))
   call assert_equal([], get(NF, 'args'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:false}, get(NF, 'arity'))
 endfunc
 
 " get({partial}, {what} [, {default}]) - in test_partial.vim
--- a/src/testdir/test_partial.vim
+++ b/src/testdir/test_partial.vim
@@ -310,6 +310,9 @@ func Test_auto_partial_rebind()
   call assert_equal('dict1', dict2['f2']())
 endfunc
 
+func s:Qux(x, y, z=3,...)
+endfunc
+
 func Test_get_partial_items()
   let dict = {'name': 'hello'}
   let args = ["foo", "bar"]
@@ -331,6 +334,17 @@ func Test_get_partial_items()
   let dict = {'partial has': 'no dict'}
   call assert_equal(dict, get(P, 'dict', dict))
   call assert_equal(0, get(l:P, 'dict'))
+
+  call assert_equal({'required': 2, 'optional': 1, 'varargs': v:true},
+      \ get(funcref('s:Qux', []), 'arity'))
+  call assert_equal({'required': 1, 'optional': 1, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 1, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2, 3]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2, 3, 4]), 'arity'))
 endfunc
 
 func Test_compare_partials()
--- a/src/testdir/test_partial.vim
+++ b/src/testdir/test_partial.vim
@@ -310,7 +310,7 @@ func Test_auto_partial_rebind()
   call assert_equal('dict1', dict2['f2']())
 endfunc
 
-func s:Qux(x, y, z=3,...)
+func s:Qux(x, y, z=3, w=1,...)
 endfunc
 
 func Test_get_partial_items()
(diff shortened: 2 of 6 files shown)


chrisbra:
Thanks! Can we get a bit of documentation please?

Saleh5515:
عدد الشيكات الاصال تقريبا 55

LemonBoy:
> Thanks! Can we get a bit of documentation please?

Done.
I don't really like the behaviour wrt partials as one can get negative number of arguments if too many arguments are specified, but I can't think of a better way to handle this. Validating the partial is not something `get` should do IMO, so throwing an error is out.

zeertzjq:
I wonder if this should be named "nargs" instead. The word "arity" has never appeared in any Vim feature, while "nargs" has appeared in `:command`.

halostatue:
> I wonder if this should be named "nargs" instead. The word "arity" has never appeared in any Vim feature, while "nargs" has appeared in `:command`.

As the person who raised the linked feature request, I would support the name change with a caveat: I worry that its name is similar enough to `args` that the documentation is going to be doing a *lot* of heavy lifting to distinguish it from the total number of expected args rather than the same as `get(func, 'args')->len()`. `arity` is the word that I first learned with Ruby and have seen since in other functional programming languages, so there is precedent.

chrisbra:
I don't like `nargs` for two reasons:
- it's to close to `args` as mentioned already
- `-nargs` is a bit different (since it refers to `:commands` not a function) and a bit limited as it currently only allows to specify 0, 1 or many arguments.

It seems `arity` is well-known enough (even so I had to look this up). So let's keep `arity`.

src/testdir/test_getvar.vim
```diff
@@ -142,20 +142,29 @@ func Test_get_func()
...
+      \ get(Fb_func, 'arity'))
   let Fb_ref = funcref('s:FooBar')
   call assert_match('<SNR>\d\+_FooBar', get(Fb_ref, 'name'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:false},
+      \ get(Fb_ref, 'arity'))
 
   call assert_equal({'func has': 'no dict'}, get(l:F, 'dict', {'func has': 'no dict'}))
   call assert_equal(0, get(l:F, 'dict'))
   call assert_equal([], get(l:F, 'args'))
+  call assert_equal({'required': 3, 'optional': 0, 'varargs': v:false}, get(l:F, 'arity'))
```
chrisbra, line 160: that seems doubled with line 145

src/testdir/test_partial.vim
```diff
@@ -331,6 +334,17 @@ func Test_get_partial_items()
...
+  call assert_equal({'required': 2, 'optional': 2, 'varargs': v:true},
+      \ get(funcref('s:Qux', []), 'arity'))
+  call assert_equal({'required': 1, 'optional': 2, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 2, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 1, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2, 3]), 'arity'))
+  call assert_equal({'required': 0, 'optional': 0, 'varargs': v:true},
+      \ get(funcref('s:Qux', [1, 2, 3, 4]), 'arity'))
```
chrisbra, line 347: I don't understand that output. Shouldn't that read: `required: 2`, `optional: 2`, `varargs: v:true` (assuming that the arity returns the number of given arguments to the function)? Also can you add a test for giving even more arguments, than the function signature?

chrisbra:
> I don't really like the behaviour wrt partials as one can get negative number of arguments if too many arguments are specified, but I can't think of a better way to handle this. Validating the partial is not something `get` should do IMO, so throwing an error is out

Thanks, I wondered about this. It's also possible to do this:
```vim
func Qux(x, y, z=3, w=1)
endfunc
echo  get(funcref('Qux', [1,2,3,4,5,6]), 'arity')
{'optional': 0, 'required': 0, 'varargs': v:false}
:echo  Qux(1,2,3,4,5,6)
E118: Too many arguments for function: Qux
```

I think for both cases it should throw an error however, otherwise I fear this would not be too useful.

Also I am slightly confused by what exactly is returned. Does this mean:
```diff
+                       "arity" A dictionary with information about the number of
+                               arguments accepted by the function with the following
+                               fields:
+                                   required    The number of positional arguments.
+                                   optional    The number of optional arguments, in
+                                               addition to the required ones.
+                                   varargs     |TRUE| if the function accepts a variable
+                                               number of arguments. See |...|.
```

It returns the number of arguments the function accepts? Or the number of arguments the function expects minus the ones provided by the partial? Or why do the numbers decrease when adding arguments to the funcref() definition?

LemonBoy:
> I think for both cases it should throw an error however, otherwise I fear this would not be too useful.

Strictly speaking the `'optional': 0, 'required': 0` part is already telling the function doesn't accept more arguments, checking if there are too many of them is validation which IMO does not have place in a property getter such as this one. Not married to this opinion though, if you or anyone else have good use cases where an error would me more appropriated let me know.

> It returns the number of arguments the function accepts? Or the number of arguments the function expects minus the ones provided by the partial? Or why do the numbers decrease when adding arguments to the funcref() definition?

It returns the number of arguments the function accepts, which for a funcref is the number of arguments of the underlying function minus the number of arguments "baked" into the funcref itself.

chrisbra:
Alright, I have tried to make further clarification based on this. Let's see if people are okay with this.

Issue #14402, opened 2024-04-03
[vim9class] vim9 seems to get confused about what file is executing


errael:
### Steps to reproduce

(this may only happen during object creation)

`A import B`, `B import C` - failure when `B` accesses something in `C`
Add `A import C` in primary file - then  `B`'s access into `C` succeeds

Comment out the indicated line, line 36, in the following to see the error
```
vim9script

# Files that gets imported. Change this name if ...
var obj_key_file_name = '/tmp/TempScript' .. string(rand()) .. '.vim'
var generator_file_name = '/tmp/TempScript' .. string(rand()) .. '.vim'

var obj_key_lines =<< trim END
    vim9script
    ### import generator_file_name as generator ###
    echo "OBJ_KEY IMPORTED"

    export class ObjKey
        const unique_object_id = generator.GenerateKey()
    endclass

END

# NOTE: If workaround enabled, then comment out the following line
#       and it still works.
obj_key_lines[1] = $'import "{generator_file_name}" as generator'

var generator_lines =<< trim END
    vim9script
    echo "GENERATOR IMPORTED"

    export def GenerateKey(): string
        return "SomeKey"
    enddef

END

writefile(obj_key_lines, obj_key_file_name)
writefile(generator_lines, generator_file_name)

import obj_key_file_name as obj_key
import generator_file_name as generator           ### EXECUTE THIS AND NO ERROR

delete(obj_key_file_name)
delete(generator_file_name)

class C1 extends obj_key.ObjKey
endclass
echo C1.new()
```
The error
```
GENERATOR IMPORTED
OBJ_KEY IMPORTED
Error detected while compiling :source buffer=1[43]..function <SNR>39_C1.new:
line    1:
E1001: Variable not found: generator
```
With the workaround
```
GENERATOR IMPORTED
OBJ_KEY IMPORTED
object of C1 {unique_object_id: SomeKey}
```

### Expected behaviour

No error

For reference, the following is the original code before the above was discovered.
The following also has a workaround, the code that can't find the function is copied into
the primary file.

In this case
`A import B` - failure when `B` accesses something in `B`
```
vim9script

# File that gets imported. Change this name if ...
var import_file_name = '/tmp/TempScript' .. string(rand()) .. '.vim'

var lines =<< trim END
    vim9script

    export interface IObjKey
        var unique_object_id: string
    endinterface

    # helper sub-class.
    export class ObjKey implements IObjKey
        const unique_object_id = GenerateKey()
    endclass

[...]

PR #15112, opened 2024-06-26
Use correct script ctx when evaluating member init

### LINKED ISSUE #14011: Variable not found when extending class with member initialized from script variable
### Steps to reproduce

1. Create script `a.vim` with the following content:

```vim
vim9script

const DEFAULT = 'default'

export class Foo
  var _x = DEFAULT
endclass
```

2. Create script `b.vim` with the following content:

```vim
vim9script
import './a.vim' as lib

class Bar extends lib.Foo
endclass

var bar = Bar.new()
```

3. Source `b.vim`.

You should see `E1001: Variable not found: DEFAULT`.

### Expected behaviour

The script should compile and run just fine.

### Version of Vim

9.1.95

### Environment

macOS 14.3
Apple Terminal
xterm-256color
ZSH 5.9

### Logs and stack traces

_No response_

LemonBoy:
The initializer expression for class members is evaluated as-needed. Use the script context where the class is defined when evaluating it to avoid missing variable errors.

Closes #14011

### DIFF
--- a/src/structs.h
+++ b/src/structs.h
@@ -1551,6 +1551,7 @@ typedef struct {
     type_T	*ocm_type;
     int		ocm_flags;
     char_u	*ocm_init;	// allocated
+    sctx_T	ocm_init_sctx;	// script context of the initializer expression
 } ocmember_T;
 
 // used for the lookup table of a class member index and object method index
--- a/src/testdir/test_vim9_import.vim
+++ b/src/testdir/test_vim9_import.vim
@@ -3330,4 +3330,25 @@ def Test_import_locked_var()
   v9.CheckScriptFailure(lines, 'E741: Value is locked: Foo', 3)
 enddef
 
+def Test_import_member_initializer()
+  var lines =<< trim END
+    vim9script
+    export const DEFAULT = 'default'
+    export class Foo
+      public var x = DEFAULT
+    endclass
+  END
+  writefile(lines, 'Ximportclass.vim', 'D')
+  # The initializer for Foo.x is evaluated in the context of Ximportclass.vim.
+  lines =<< trim END
+    vim9script
+    import './Ximportclass.vim' as X
+    class Bar extends X.Foo
+    endclass
+    var o = Bar.new()
+    assert_equal(X.DEFAULT, o.x)
+  END
+  v9.CheckScriptSuccess(lines)
+enddef
+
 " vim: ts=8 sw=2 sts=2 expandtab tw=80 fdm=marker
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -204,7 +204,13 @@ add_member(
 	m->ocm_flags |= OCMFLAG_HAS_TYPE;
     m->ocm_type = type;
     if (init_expr != NULL)
+    {
 	m->ocm_init = init_expr;
+	// Save the script context, we need it when evaluating or compiling the
+	// initializer expression.
+	m->ocm_init_sctx = current_sctx;
+	m->ocm_init_sctx.sc_lnum += SOURCING_LNUM;
+    }
     ++gap->ga_len;
     return OK;
 }
@@ -1259,7 +1265,11 @@ add_class_members(class_T *cl, exarg_T *eap, garray_T *type_list_gap)
 	typval_T	*tv = &cl->class_members_tv[i];
 	if (m->ocm_init != NULL)
 	{
+	    sctx_T	save_current_sctx = current_sctx;
+
+	    current_sctx = m->ocm_init_sctx;
 	    typval_T *etv = eval_expr(m->ocm_init, eap);
+	    current_sctx = save_current_sctx;
 	    if (etv == NULL)
 		return FAIL;
 
--- a/src/vim9compile.c
+++ b/src/vim9compile.c
@@ -3295,9 +3295,13 @@ obj_constructor_prologue(ufunc_T *ufunc, cctx_T *cctx)
 
 	if (m->ocm_init != NULL)
 	{
-	    char_u *expr = m->ocm_init;
+	    char_u	*expr = m->ocm_init;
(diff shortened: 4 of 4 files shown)


LemonBoy:
I think this fixes #14402 too, @errael can you confirm that's the case here?

errael:
> I think this fixes #14402 too, @errael can you confirm that's the case here?

The example in #14402 does not fail with this patch; so you can add it as another issue that is fixed by this PR.

I'll play with it a little more, see if I can come up with a multi-step/nested test that also fails. Maybe I can couple my test to @lifepillar test (the tests superficially seem different).

errael:
Found another failing case which seems related.

Put the two files in the same directory.

Source `obj_key.vim` and it works OK with current and this PR.

**t.vim**
```
vim9script
import './obj_key.vim' as obj_key

class C1 extends obj_key.ObjKey
endclass
var o = C1.new()
echom string(o)
```

**obj_key**
```
vim9script
echom "OBJ_KEY IMPORTED"

const DEFAULT = 'default-obj_key'

export class ObjKey
    const unique_object_id3 = DEFAULT
endclass

var o = ObjKey.new()
echom string(o)
```

Source `t.vim` with current and it fails as expected.
Looks like the original problem
```
OBJ_KEY IMPORTED
object of ObjKey {unique_object_id3: 'default-obj_key'}
Error detected while compiling /home/err/play/tdir2/t.vim[6]..function <SNR>45_C1.new:
line    1:
E1001: Variable not found: DEFAULT
```

Source `t.vim` with patch and it fails differently.
Looks like a similar confusion, but about `export`
```
OBJ_KEY IMPORTED
object of ObjKey {unique_object_id3: 'default-obj_key'}
Error detected while processing /home/err/play/tdir2/t.vim[6]..function <SNR>45_C1.new:
line    1:
E1049: Item not exported in script: DEFAULT
```

LemonBoy:
> Looks like a similar confusion, but about export

That's expected, the initializer expression is taken by reference and evaluated in the `new` method, hence the need to make it visible from outside the script it was declared in. Note that the former error states `detected while compiling` while the new one `detected while processing`.

Is this right? I don't know. Braam modeled this behavior after Typescript (and I'm not really a fan) but have no idea of how it behaves in such situations.

errael:
> > Looks like a similar confusion, but about export
> 
> That's expected, the initializer expression is taken by reference and evaluated in the `new` method, hence the need to make it visible from outside the script it was declared in.

That sounds like an implementation issue. With that description, it could be argued that this PR as it stands is not needed.

In the OP of this PR it says

> Use the script context where the class is defined

If it doesn't use the context where it is defined for execution, then why is it needed for compilation?

> Note that the former error states `detected while compiling` while the new one `detected while processing`.

I thought that if it compiled, then it would run. Excluding when `any` is used and there's dynamic type checking involved. I looked around a little, and didn't any clarification on this. Under `help constructor` it says

> In the second step the object variables of the parent class are initialized first.

Anyway, this behavior is confusing and shouldn't be a runtime error.

@yegappan can you offer some clarification?

> 
> Is this right? I don't know. Braam modeled this behavior after Typescript (and I'm not really a fan) but have no idea of how it behaves in such situations.

I know nothing about Typescript. But if it must fail (and I don't agree that it should), better to fail in compilation.

LemonBoy:
> That sounds like an implementation issue. With that description, it could be argued that this PR as it stands is not needed.

Without this patch the compilation failed because the variable could not be found in the script context, fixing that means you can now worry about the variable being exported or not.

You can't do better than this if we wish to retain this behaviour, the `new` cannot be executed in the declaration context as it'll cause countless problems beside feeling wrong on many levels.

Personally speaking I don't really like this dynamic behaviour and would rather require all the initializers to be constant rvalues. We could then evaluate the initializers at compile time and use that value in the constructor, avoiding all the context-related issues.

errael:
> Without this patch the compilation failed

Code that compiles should not fail at at runtime (except for type errors, probably a few other reasons)

> if we wish to retain this behaviour

It's good to know that I can still be shocked an horrified by vim9script ;-)

BTW, this change is not backwards compatible (that doesn't bother me if it's a bug fix, but in this situation I'm lost). 

Add this line to t.vim
```
var DEFAULT = 'NOT FROM obj_key.vim'
```
Then run `t.vim`, without this patch, and you get **no error**
```
OBJ_KEY IMPORTED
object of ObjKey {unique_object_id3: 'default-obj_key'}
object of C1 {unique_object_id3: 'NOT FROM obj_key.vim'}
```
With this patch you get an error
```
OBJ_KEY IMPORTED
object of ObjKey {unique_object_id3: 'default-obj_key'}
Error detected while processing /home/err/play/tdir2b/t.vim[9]..function <SNR>45_C1.new:
line    1:
E1049: Item not exported in script: DEFAULT
```
@dkearns Were you aware of this? Do you have any words to bring me down?

LemonBoy:
> Code that compiles should not fail at at runtime

Well one could argue that the intializer expressions all must have the same privacy as the class itself. In this case the compilation would fail if the class is declared as `export` and the non-exported `DEFAULT` value is referenced as initializer.

> Add this line to t.vim

The previous behaviour look quite wrong to me.

errael:
> > Code that compiles should not fail at at runtime
> 
> Well one could argue that the intializer expressions all must have the same privacy as the class itself. In this case the compilation would fail if the class is declared as `export` and the non-exported `DEFAULT` value is referenced as initializer.

I don't understand how it is supposed to work.

> 
> > Add this line to t.vim
> 
> The previous behaviour look quite wrong to me.

What did you mean by the following?

> You can't do better than this if we wish to retain this behaviour, the new cannot be executed in the 
> declaration context as it'll cause countless problems beside feeling wrong on many levels.

chrisbra:
so, can we fix it, so that is behaves like for Typescript?

PR #15113, opened 2024-06-26
translation: update spelling suggestion message across all languages

nullromo:
I updated the translation files based on the discussion in [this thread](https://github.com/vim/vim/pull/15065#issuecomment-2183187747).

There was debate about whether this should be done or not, but I decided to submit a pull request for it anyway.

It doesn't _exactly_ follow the process in `src/po/README.txt`, but if this update is not made now, then messages will be incorrect until each individual language translation is updated. That could take a long time, and in the meantime the messages will not be up to date.


chrisbra:
thanks!

PR #15114, opened 2024-06-27
netrw: fix remaining case of register clobber

Konfekt:
complements 62f7b55c1a4564f8744af9446bc7af47fe16a245


chrisbra:
ah, thanks!

PR #15116, opened 2024-06-27
(Try to) gracefully handle out-of-memory in bindtextdomain().

cvwillegen:
The error return value of ``bindtextdomain()`` was not propagated to the original calling script. This implements returning ``FALSE`` when the call fails, and ``TRUE`` when it succeeds.

### DIFF
--- a/src/testdir/test_gettext.vim
+++ b/src/testdir/test_gettext.vim
@@ -1,18 +1,18 @@
 source check.vim
 
+CheckFeature gettext
+
 " Test for gettext()
 func Test_gettext()
-  if has("gettext")
-    call assert_fails('call bindtextdomain("test")', 'E119:')
-    call assert_fails('call bindtextdomain("vim", "test")', 'E475:')
+  call assert_fails('call bindtextdomain("test")', 'E119:')
+  call assert_fails('call bindtextdomain("vim", "test")', 'E475:')
 
-    call assert_fails('call gettext(1)', 'E1174:')
-    call assert_equal('xxxTESTxxx', gettext("xxxTESTxxx"))
+  call assert_fails('call gettext(1)', 'E1174:')
+  call assert_equal('xxxTESTxxx', gettext("xxxTESTxxx"))
 
-    call assert_equal('xxxTESTxxx', gettext("xxxTESTxxx", "vim"))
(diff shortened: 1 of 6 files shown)


zeertzjq:
Build failure on macos-14

vim-ml:
Op wo 3 jul. 2024 13:06 schreef zeertzjq ***@***.***>:

> Build failure on macos-14
>

Is the return type on OS-14 different? It gives me a ... weird error.
Expression expected??

The other uses of the function don't check the return value, so I can't see
how it "should" be done.

I propose we don't check the return value. Getting an out of memory error
here looks very unlikely...

Christ van Willegen

>

chrisbra:
Hm, not sure what is wrong on MacOS. Perhaps change this:
```c
if (bindtextdomain((const char *)argvars[0].vval.v_string, (const char *)argvars[1].vval.v_string) == NULL)
``` 

to 
```c
if (!bindtextdomain((const char *)argvars[0].vval.v_string, (const char *)argvars[1].vval.v_string))
```

LemonBoy:
In [this](https://github.com/vim/vim/blob/898b3740c7f9fa544c82c9ae6927617aad9a6463/src/vim.h#L601) case the check expands to `( == NULL)`, that's a problem. The macros should expand to some non-null value.

src/testdir/test_gettext.vim
```diff
@@ -2,15 +2,17 @@ source check.vim
 
 " Test for gettext()
 func Test_gettext()
-  call assert_fails('call bindtextdomain("test")', 'E119:')
-  call assert_fails('call bindtextdomain("vim", "test")', 'E475:')
+  if has("gettext")
```
k-takata: `CheckFeature gettext` might be better?

src/vim.h
```diff
@@ -598,7 +598,7 @@ extern int (*dyn_libintl_wputenv)(const wchar_t *envstring);
 # ifdef bindtextdomain
 #  undef bindtextdomain
 # endif
-# define bindtextdomain(x, y) // empty
+# define bindtextdomain(x, y) (y)
```
LemonBoy:
```suggestion
# define bindtextdomain(x, y) ""
```
This is better as `bindtextdomain(<whatever>, NULL)` does not return `NULL` (unless there's something wrong).

chrisbra:
thanks!

Issue #11244, opened 2022-09-28
MANPAGER, :Man does not follow the correct section of a man page


FueLee:
### Steps to reproduce

1. vim -c MANPAGER -u NORC -i NONE - (set MANPAGER environment variable, without user config file)
2. man 2 read
3. scroll down to ERRORS where there is a "link" to open(2). (See open(2) for further details on...)
4. Hit Shift + k = (K) on "open(2)"
5. xdg-open(1) is opened, not open(2)

### Expected behaviour

After "Shift+k" on "open(2)" the open(2) manual page should be opened.

Later, I modified the iskeyword value: set iskeyword+=(,)

in ...plugin/manpager.vim and finally moved this into ...ftplugin/man.vim and the correct man page was opened next time for those cases:
- in terminal/console at bash: man 2 read 
- in vim as :Man 2 read

### Version of Vim

VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Sep 13 2022 09:35:02), Included patches: 1-3995, 4563, 4646, 4774, 4895, 4899, 4901, 4919

### Environment

Ubuntu Linux 22.04.1 LTS
Virtual console (dev/tty) $TERM=linux, Virtual terminal $TERM=rxvt-unicode-256color
$SHELL=/bin/bash
$BASH_VERSION=5.1.16(1)-release

lacygoill:
I can't reproduce this issue on my machine:
![gif](https://user-images.githubusercontent.com/8505073/192819854-c735b7ee-b968-42b8-9078-7f37477c6ad6.gif)

When I press `K` on the `open(2)` token, the manpage for `open(2)` is opened, not the one for `xdg-open(1)`.

**Version of Vim**

9.0 Included patches: 1-612

**Environment**

Operating system: Ubuntu 20.04.5 LTS
Terminal: XTerm(353)
Value of $TERM: xterm-256color
Shell: GNU bash, version 5.0.17

brammool:
> ### Steps to reproduce
> 
> 1. vim -c MANPAGER -u NORC -i NONE - (set MANPAGER environment variable, without user config file)
> 2. man 2 read
> 3. scroll down to ERRORS where there is a "link" to open(2). (See open(2) for further details on...)
> 4. Hit Shift + k = (K) on "open(2)"
> 5. xdg-open(1) is opened, not open(2)
 
I used to have the same problem, which was caused by the K key being
mapped.  Do you have a mapping for the K key?  ":verbose map K" should
show.

[...]

PR #15117, opened 2024-06-28
runtime(man): UX fixes

intelfx:
Two minor improvements to manpage support that _seem_ to do the right thing for me.
Note that I'm fairly new to Vim{,script}, so please do not kick me too hard if any of these changes are bogus.

- Honor placement (vert/hor/tab) cmd modifiers to `:Man`, if any are present, before `g:ft_man_open_mode`.  
  This way, e.g. `:vert Man` does the expected thing even in presence of a contradicting `g:ft_man_open_mode`.

- Actually add parentheses to `iskeyword=` so that `K` on a man page reference considers its section (if present).  
  The comment suggests that the parentheses were added to the iskeyword set elsewhere, but I could not find it in the code (and, indeed, it did not work in practice).

### DIFF
--- a/runtime/autoload/dist/man.vim
+++ b/runtime/autoload/dist/man.vim
@@ -165,7 +165,9 @@ func dist#man#GetPage(cmdmods, ...)
       endwhile
     endif
     if &filetype != "man"
-      if exists("g:ft_man_open_mode")
+      if a:cmdmods =~ '\<\(tab\|vertical\|horizontal\)\>'
+	let open_cmd = a:cmdmods . ' split'
+      elseif exists("g:ft_man_open_mode")
         if g:ft_man_open_mode == 'vert'
 	  let open_cmd = 'vsplit'
         elseif g:ft_man_open_mode == 'tab'
@@ -174,7 +176,7 @@ func dist#man#GetPage(cmdmods, ...)
 	  let open_cmd = 'split'
         endif
       else
-	let open_cmd = a:cmdmods . ' split'
+	let open_cmd = 'split'
       endif
     endif
   endif
--- a/runtime/ftplugin/man.vim
+++ b/runtime/ftplugin/man.vim
@@ -23,9 +23,9 @@ let s:cpo_save = &cpo
 set cpo-=C
 
 if &filetype == "man"
-  " Allow hyphen, plus, colon, dot, and commercial at in manual page name.
-  " Parentheses are not here but in dist#man#PreGetPage()
-  setlocal iskeyword=48-57,_,a-z,A-Z,-,+,:,.,@-@
+  " Allow hyphen, plus, colon, dot, parentheses and commercial at
+  " in manual page names.
+  setlocal iskeyword=48-57,_,a-z,A-Z,-,+,:,.,(,),@-@
   let b:undo_ftplugin = "setlocal iskeyword<"
 
   " Add mappings, unless the user didn't want this.


zzzyxwvut:
What manual page issue are you currently having that can be  
solved by including `(,)` in `&iskeyword` script-wide?

The reason for chucking out parentheses at [`8cfe52`](https://github.com/vim/vim/commit/8cfe52e6fbf44032cd40d1561e93644786b15ee7) was that  
having them included _or_ not is immaterial for parsing `str`:

https://github.com/vim/vim/blob/88bbdb04c2776ba69b8e5da58051fd94f8842b03/runtime/autoload/dist/man.vim#L72-L77

zzzyxwvut:
I can reproduce the faulty behaviour (v9.1.0516).

For example, initially read `signal(2)` and later fail to  
read `signal(7)`with `K` (when `&kp` is set to `:Man`):
```shell
MANWIDTH=80 VIMRUNTIME=/path/to/fork/dir/runtime \
vim --clean --not-a-term \
  '+set vif=NONE' \
  '+sil! unmap K' \
  '+doautoall FileType setl ft=man kp=:Man nofen' \
  '+Man 2 signal' \
  '+/signal(7)' \
  '+norm f7BK'
```

However, finding a page should work whenever the sought page  
is locally installed in only one section directory and its  
name matches the referenced section number (e.g. `full(4)`).

It appears that parentheses had exclusively been used for  
fetching `<cword>` ever since [`vim-6-0k`](https://github.com/vim/vim-history/commit/2672c27009bf75a8252a100ddf5e79dd1150640b) (Oct. 2000) and until  
recently, when at [`f269ea`](https://github.com/vim/vim/commit/f269eabc6c4f5bdcef989cd5b4b95ba8ccaa4d8a) (see #11244) similar changes were  
applied, but summarily dismissed at [`71badf`](https://github.com/vim/vim/commit/71badf9547e8f89571b9a095183671cbb333d528) with a comment  
referencing `dist#man#PreGetPage()`.

And you don't get to see `signal(7)` anyway with the proposed  
changes for the above Vim command, because _any_  parenthesis  
now belongs to `\k`, so `K` would try looking for `signal(7))`  
(note the extra closing parenthesis).  Also, with `K` you  
can't follow a reference that terminates in a hyphen-minus  
or a hyphen; that is, split between two lines.

Anyway, you need to run the changes by the maintainers:  
@goweol, @lifecrisis.

benknoble:
I expect the modifiers support would be welcome, but the parentheses bit seems controversial.

zzzyxwvut:
Further observations.

The current execution of `K` (`setl kp=:Man`) is aligned with  
the `Man` command invocation when the latter does not take any    
section number, e.g. `:Man signal`, either would pick the same  
page; and `[count]K` is aligned with `:Man [section] [page]`,  
e.g. `7K` and `:Man 7 signal`.

Indefinitely adding `(,)` to `&isk` would break this symmetry  
by rendering `K` more capable than `:Man [page]`, and it would  
also neither make `K` work nor `\K` nor `C-]` to boot on some  
references that contain a `\k` character after the closing  
parenthesis, e.g. `signal(7)) `, `signal(7).)`.  If accepted,  
this line should also be adapted:


https://github.com/vim/vim/blob/88bbdb04c2776ba69b8e5da58051fd94f8842b03/runtime/autoload/dist/man.vim#L74

chrisbra:
thanks. As mentioned by others, can you please drop the iskeyword setting? That seems too controversial. I haven't heard from the maintainers in a long time, so I am fine with including the command modifier changes here.

PR #15118, opened 2024-06-28
win32: Enable batch inference for libvterm

k-takata:
Enable batch inference for libvterm by using $(OBJDIR)/libvterm as the object directory.
This should make the compilation process a bit faster.


chrisbra:
thanks!

PR #15119, opened 2024-06-28
xxd,msvc: Specify source-charset

k-takata:
xxd.c has non-ASCII-character comments. This causes the following warning on MSVC:
```
warning C4819: The file contains a character that cannot be represented
in the current code page (932). Save the file in Unicode format to
prevent data loss.
```

Add the `/source-charset:utf-8` option to avoid this.


chrisbra:
thanks!

PR #15120, opened 2024-06-28
ex: fix newline escaping

### LINKED ISSUE #6135: global/veto handling of backslashes in multi-line commands
## Description
The command(s) executed by the `global`/`veto` ex command
can be multi-line commands, e.g., `append`, `change`, or `insert`.  The newlines
in these multi-line commands are escaped by backslashes, but the
newlines themselves must remain in the `global` command list
to allow for appending/changing/inserting
multiple lines per match of the `global` command.

For example, if, in a C program, you wanted to insert an empty line before each
*enum* declaration, you would write (in ed or in traditional ex):
```
:g/enum/i\
\
.
```

That is, the backslashes -- *but not the newlines protected by those backslashes* --
are to be swallowed by `global` parsing.

However, what vim does (at least in ex mode as entered by visual-mode `Q`)
is to insert nothing, presumably because both the backslash *and* the newline
have been swallowed, leaving the `insert` command with nothing
(i.e., the empty string) to insert,

POSIX [says](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ex.html#tag_20_40_13_20)
> For the `append`, `change`, and `insert` commands, the input text shall be included as part of the command, and the terminating `<period>` can be omitted if the command ends the list of commands.

It also says that multiple commands can be separated by `<vertical-line>` or by `<backslash>` escaped newlines.

The POSIX [ed man](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ed.html#tag_20_38_13_10) page makes this clearer:
> A single command or the first of a list of commands shall appear on the same line as the `global` command. All lines of a multi-line list except the last line shall be ended with a backslash preceding the terminating `<newline>`; the `a`, `i`, and `c` commands and associated input are permitted. The '`.`' terminating input mode can be omitted if it would be the last line of the command list.

## To Reproduce
1. Run `vim --clean`
2. Edit `example`
3. Type `aenum^[Qg/enum/i\^M\^M.`, where `^[` is ASCII `ESC` (escape)
and `^M` is ASCII `CR` (carriage return),
backslashes exactly as specified (i.e., literally).
Or in ex mode, entered via `Q`:
```
:a
enum
.
:g/enum/i\
\
.
:%p
```
4. There should be an empty line in front of the line containing *enum*.

## Expected behavior
Multi-line `append`, `change`, and `insert` ex commands should be possible in ex `global` commands, at least in ex mode as entered via the visual `Q` command.:

## Environment:
 - Vim version: 8.2 (2019 Dec 12)
 - OS: Fedora 31
 - Terminal: urxvt

### LINKED ISSUE #7244: ex: global and ! commands don't handle multi-line commands correctly
**Describe the bug**
When vim is invoked as ex, it does not seem to handle several commands passed to the global command.

**To Reproduce**
Detailed steps to reproduce the behavior:
1. Run `printf '%s\n' i 1 2 . g/./p\\ p | ex`

**Expected behavior**
```
1
1
2
2
```

**Actual behavior**
```
1
2
2
```

 - Vim version 8.2.1943
 - OS: macOS 10.15.7
 - Terminal: Terminal.app

**Additional context**

It should also handle `!` commands correctly. `printf '%s\n' i 1 2 . 'g /./ !echo line \' p | ex` should output:

```
line
1
line
2
```

mohd-akram:
This fixes newline escaping to allow passing multiple commands to ":global", multiple lines to shell commands, and ending lines in append mode with backslashes. This reverts a previous incorrect attempt to fix append mode which removed half of trailing backslashes which lead to, eg. the following two commands being parsed as having a different number of backslashes:

```
!echo foo\\\
```

```
!echo foo\\ \
```

Fixes #6135
Fixes #7244

@konrad-schwarz

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -2336,12 +2336,7 @@ do_one_cmd(
     {
 	for (p = ea.arg; *p; ++p)
 	{
-	    // Remove one backslash before a newline, so that it's possible to
-	    // pass a newline to the shell and also a newline that is preceded
-	    // with a backslash.  This makes it impossible to end a shell
-	    // command in a backslash, but that doesn't appear useful.
-	    // Halving the number of backslashes is incompatible with previous
-	    // versions.
+	    // Remove one backslash before a newline
 	    if (*p == '\\' && p[1] == '\n')
 		STRMOVE(p, p + 1);
 	    else if (*p == '\n' && !(ea.argt & EX_EXPR_ARG))
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -3138,31 +3138,15 @@ redraw:
 	windgoto(msg_row, msg_col);
 	pend = (char_u *)(line_ga.ga_data) + line_ga.ga_len;
 
-	// We are done when a NL is entered, but not when it comes after an
-	// odd number of backslashes, that results in a NUL.
-	if (line_ga.ga_len > 0 && pend[-1] == '\n')
+	// We are done when a NL is entered, but not when it comes after a
+	// backslash in prompt mode.
+	if (line_ga.ga_len > 0 && pend[-1] == '\n'
+		&& (line_ga.ga_len <= 1 || pend[-2] != '\\' || !promptc))
 	{
-	    int bcount = 0;
-
-	    while (line_ga.ga_len - 2 >= bcount && pend[-2 - bcount] == '\\')
-		++bcount;
-
-	    if (bcount > 0)
-	    {
-		// Halve the number of backslashes: "\NL" -> "NUL", "\\NL" ->
-		// "\NL", etc.
-		line_ga.ga_len -= (bcount + 1) / 2;
-		pend -= (bcount + 1) / 2;
-		pend[-1] = '\n';
-	    }
-
-	    if ((bcount & 1) == 0)
-	    {
-		--line_ga.ga_len;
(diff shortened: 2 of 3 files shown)


mohd-akram:
The fix for this turned out to be much simpler. The culprit was 417f5e7f1173d63e9dcb77c4b80cb6bb317336c8, so ex mode has been broken for around 14 years...

chrisbra:
Thanks!

k-takata:
Does this just revert 7.3.014?
The issue that 7.3.014 tried to solve is unresolved?

mohd-akram:
It is resolved. We fix it correctly by not interpreting backslashes at all in insert mode.

chrisbra:
It's not completely reverted, there is an added check for `!promptc` var.

There were some discussions around 7.3.014: https://groups.google.com/g/vim_dev/c/F5-tDqoafz8/m/GqKF-uQsLD0J

The example mentioned in this thread: 
```
To recreate the problem, perform the following:
cat - > ex.input << EOF
a
X \\
Y
.
w ex.output
EOF
cat -tvse ex.input
ex - -u NONE < ex.input
cat -tvse ex.output

When working correctly (rhel4), the output from cat of ex.output is:
$
X \$
Y$ 
```

This works now correctly. Should probably turn this into a test as well.

PR #15124, opened 2024-06-29
Accept a default priority level for signs

### LINKED ISSUE #8334: Allow sign_define to set a default value for the sign priority

Plugins authors can currently tell users that they can override a sign definition to customize the text, or highlight groups. But if they want to allow users to customize the priority of signs they need to provide a separate custom setting.

It would be nice if users could set the default priority as part of a sign definition.

- Extend the `dict` argument of `sign_define` to take a `priority` option.
- Change `sign_place` to lookup the `priority` value from the definition if there is no `priority` property within the `dict` argument.

The plugin can expose a custom setting to configure the sign priority and uses it whenever the plugin calls `sign_place`.

---

(I could create a PR to make this change, just wanting to ensure the feature itself would be accepted)

LemonBoy:
Use the specified value instead of the default one (SIGN_DEF_PRIO) when no priority is explicitly specified in sign_place or :sign place.

Closes #8334


zeertzjq:
This is not documented

chrisbra:
thanks!

PR #15125, opened 2024-06-29
Refactor some code around f_srand()

LemonBoy:
Nothing too serious, a small refactoring and a bonus.

Move the machine-specific parts to os_mswin and os_unix.
Implement a simple fallback for Windows (works on 10 or later).

### DIFF
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -9267,69 +9267,47 @@ f_test_srand_seed(typval_T *argvars, typval_T *rettv UNUSED)
     static void
 init_srand(UINT32_T *x)
 {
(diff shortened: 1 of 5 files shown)


chrisbra:
Thanks. The change on MS-Windows, do we need to care for Windows 8 support?

LemonBoy:
> The change on MS-Windows, do we need to care for Windows 8 support?

On systems older than Windows 10 this change is a no-op and libsodium (if available) is used instead, adding a fallback using `RtlGenRandom` or similar takes too many lines for little or no gain.

chrisbra:
alright, thanks then.

PR #15127, opened 2024-06-29
Wrong spanish translation. Lost some meaning

rodoviario:
I was followng the vimtutor when spotted this.
The specific line looks to have been initially transliterated word by word from english, then rewrited, but part of the important keywords like command and operators remained in an awkward position.

Just to be sure, I've checked the main english version of the line in the tutor file.

Spanish is my mother language, anyway, but if there is a spanish speaker among the devs who cares to take a look at this to double check my criteria, please do :)

### DIFF
--- a/runtime/tutor/tutor.es
+++ b/runtime/tutor/tutor.es
@@ -236,7 +236,7 @@ Ahora continúe con la Lección 2.
 
   Muchos comandos que cambian texto están compuestos por un operador y un
   movimiento.
-  El formato para eliminar un comando con el operador de borrado  d  es el
+  El formato para comando eliminar con el operador de borrado  d  es el
   siguiente:
 
     d   movimiento
--- a/runtime/tutor/tutor.es.utf-8
+++ b/runtime/tutor/tutor.es.utf-8
@@ -236,7 +236,7 @@ Ahora continúe con la Lección 2.
 
   Muchos comandos que cambian texto están compuestos por un operador y un
   movimiento.
-  El formato para eliminar un comando con el operador de borrado  d  es el
+  El formato para comando eliminar con el operador de borrado  d  es el
   siguiente:
 
     d   movimiento


rodoviario:
I saw the 30 automatic tests roll. The one in macos 12 is faling. The freeBSD also fails (time out).
Looks to me too complex for just a grammar/translation fix.  Hope it gets a second chance with the failing tests since that the timeout cannot be related to the change in that text file.
Considering that this is a small edit in a specific language text file, I guess that it will be a lot harder for real code to get trough all the ci testing.

RestorerZ:
Probably, it needs to be fixed here too:
https://github.com/vim/vim/blob/f0837ba0b9dba9d30e3dc2626d4f690a2d51103e/runtime/tutor/tutor.es#L239

rodoviario:
I didn't fix it there because I've readed that the the file is created
automatically from the utf-8 one...
See here
https://github.com/vim/vim/blob/f0837ba0b9dba9d30e3dc2626d4f690a2d51103e/runtime/tutor/README.txt#L29
 but I was not sure about that. I can make the change there too if it is
not going to create conflict with that auto function. Just let me know

RestorerZ:
No, it is not created automatically neither when building, nor when adding to archives, nor when creating an installer.
```
make tutor.es
```

rodoviario:
> No, it is not created automatically neither when building, nor when adding to archives, nor when creating an installer.
> 
> ```
> make tutor.es
> ```

I've just edited in github directly, but it appear to have change the encoding so the whole file looks changed. I'm cloning the repo in my machine. I will try to recommit with *only* the intended change.

RestorerZ:
If you cloned the git repository, the directory is "tutor/make -f Makefile tutor.es" on UNIX-like systems, or "tutor\nmake.exe -f Make_mvc.mak totor.es". Just be sure to put your edited UTF-8 file in the "/runtime/tutor" directory of the repository before doing so.
Direct editing in modern browsers is not the best solution when there are encodings other than UTF-8.
Lost technologies...

https://github.com/vim/vim/blob/f0837ba0b9dba9d30e3dc2626d4f690a2d51103e/runtime/tutor/Make_mvc.mak#L103C1-L112C1

https://github.com/vim/vim/blob/f0837ba0b9dba9d30e3dc2626d4f690a2d51103e/runtime/tutor/Make_mvc.mak#L103C1-L112C1

rodoviario:
Nevermind, the solution was in front of my eyes. Instead trying to understando github / vscode automatic transcoding, the only (right) thing to do was editing the file in vim :) and use the git cli to force push the change. This way it didn't change anything else beyond the edited line :smile:

RestorerZ:
That's great!
Vim editor is the best!

chrisbra:
thanks!

PR #15128, opened 2024-06-29
Fix exception throwing after silent try

### LINKED ISSUE #538: Explicit throw in context of :silent! call breaks try/catch error handling
It looks like calling `throw` explicitly within the context of `:silent! call` somehow messes up either program execution flow or future error handling.

Minimal reproduction:

``` viml
function! ThrowBoom() abort
  " throw "Can't open file /tmp/notfound"
  call readfile('/tmp/notfound')
endfunction

function! Repro() abort
  call ThrowBoom()
  try
    call join(1)
  catch /E714:/
  endtry
endfunction
```

Calling `Repro()` calls `ThrowBoom()`, which will throw E484 (because `/tmp/notfound` doesn't exist). If execution had continued, the call to `join(1)` would throw E714, which would be ignored.

This works exactly as expected:

```
:call Repro()
Error detected while processing function Repro[1]..ThrowBoom:
line    2:
E484: Can't open file /tmp/notfound
```

And `:silent! call Repro()` correctly produces no output.

Now uncomment the explicit `throw` command in `ThrowBoom()`:

``` viml
function! ThrowBoom() abort
  throw "Can't open file /tmp/notfound"
  " call readfile('/tmp/notfound')
endfunction
```

`:call` functions in basically the same way as before:

```
:call Repro()
Error detected while processing function Repro[1]..ThrowBoom:
line    1:
E605: Exception not caught: Can't open file /tmp/notfound
```

Different exception, but fine. Now try with `:silent!`:

```
:silent! call Repro()
Error detected while processing function Repro:
line    3:
E714: List required
```

In this case, the exception thrown from `ThrowBoom()` is swallowed, program execution continues, and then the E714 exception thrown from `join(1)` is _not_ caught by the `catch` clause.

It seem like program flow under `:silent!` will differ depending on whether an exception occurs implicitly (from e.g. `readfile()` throwing E484) or explicitly (via `throw`), and in the latter case something also happens that breaks later error handling.

The documentation for `:silent!` isn't very clear on what effect it has, but I don't see any reason for it to differ depending on whether an exception was generated internally or via `:throw`. (Ideally, I'd like for program execution to be identical with or without `:silent!`, except that using it would suppress reporting of error messages to the user.)

LemonBoy:
The silent! flag causes the evaluation not to be aborted in case of uncaught exceptions, adjust handle_did_throw to take this detail into account.

Fixes the long-standing todo.txt item:
```
Problem that a previous silent ":throw" causes a following try/catch not
to work. (ZyX, 2013 Sep 28) With examples: (Malcolm Rowe, 2015 Dec 24)
Also see #8487 for an example.
```

Fixes #538

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -1460,8 +1460,13 @@ handle_did_throw(void)
     current_exception->throw_name = NULL;
 
     discard_current_exception();	// uses IObuff if 'verbose'
-    suppress_errthrow = TRUE;
-    force_abort = TRUE;
+
+    // If "silent!" is active the uncaught exception is not fatal.
+    if (emsg_silent == 0)
+    {
+	suppress_errthrow = TRUE;
+	force_abort = TRUE;
+    }
 
     if (messages != NULL)
     {
--- a/src/testdir/test_vimscript.vim
+++ b/src/testdir/test_vimscript.vim
@@ -7536,6 +7536,31 @@ func Test_deeply_nested_source()
   call system(cmd)
 endfunc
 
+func Test_exception_silent()
+  XpathINIT
+  let lines =<< trim END
+  func Throw()
+    Xpath 'a'
+    throw "Uncaught"
+    " This line is not executed.
+    Xpath 'b'
+  endfunc
+  " The exception is suppressed due to the presence of silent!.
+  silent! call Throw()
+  try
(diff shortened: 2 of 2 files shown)


chrisbra:
thanks!

PR #15130, opened 2024-06-30
runtime(man): use `nnoremap` to map to Ex commands

intelfx:
(Companion to #15117...)

If the user plays rebinding games such as

    nnoremap : ,
    nnoremap , :

(cf. https://konfekt.github.io/blog/2016/10/03/get-the-leader-right),
then the mappings defined by man.vim will become non-functional.


chrisbra:
thanks!

PR #15138, opened 2024-07-02
runtime(rust): use shiftwidth() in indent script

9ary:
This fixes a couple of hardcoded values and replaces an `&shiftwidth` with `shiftwidth()`.

Upstream seems unmaintained so I figured this was the best place to submit this, let me know if I should attempt to submit this there anyway.

This script defines an `s:shiftwidth()`, though it's not using it anywhere. I went for the builtin directly since this is the vim repo anyway.

I've also left one instance alone for `where` clauses, since it's explicitly doing horizontal alignment.

### DIFF
--- a/runtime/indent/rust.vim
+++ b/runtime/indent/rust.vim
@@ -1,7 +1,7 @@
 " Vim indent file
 " Language:         Rust
 " Author:           Chris Morgan <me@chrismorgan.info>
-" Last Change:      2023-09-11
+" Last Change:      2024-07-03
 " For bugs, patches and license go to https://github.com/rust-lang/rust.vim
 
 " Only load this indent file when no other was loaded.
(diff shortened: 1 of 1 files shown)


chrisbra:
> Upstream seems unmaintained so I figured this was the best place to submit this, let me know if I should attempt to submit this there anyway.

Is this really the case @da-x @chris-morgan ?

9ary:
For what it's worth: https://github.com/rust-lang/rust.vim/issues/502#issuecomment-1746166405 (october 2023)
>We lack active maintainers. I for one, too busy doing other stuff in the last two years.

Upstream has received no commits since september 2022, and during 2022, there were only 4 trivial commits.

Several changes were made to the Rust files in this repo that aren't present upstream.

chrisbra:
I see, thanks

PR #15139, opened 2024-07-03
Unwanted cursor movement with pagescroll at start of buffer

luukvbaal:
Problem:  Cursor is moved to bottom of window trying to pagescroll when
          already at the start of the buffer.
Solution: Don't move cursor when buffer content did not move.


Asheq:
Thanks @luukvbaal!

I created one more issue related to <C-b>:  https://github.com/vim/vim/issues/15146

chrisbra:
thanks!

PR #15099, opened 2024-06-24
Uniform naming in makefiles of macros pointing to a Vim executable file

RestorerZ:
Problem:
All makefiles in use have different names for macros pointing to the
location of the Vim executable. And when you need to specify a different
location of the Vim file via the make program command line, you either have to
remember the name of the macro in the given makefile, or you have to open the
makefile to see the name of the macro.
This, as you may realize, is not very good for productivity and also creates
some confusion in the makefiles themselves.

Solution:
Use a common name for such a macro in the makefiles.
This macro name is "VIMPROG", similar to the `v:progpath` variable.
(in addition, the same name is used in the syntax checker Makefile and testdir).

PR #15143, opened 2024-07-03
[vim9class] Fix `string()` to handle an object's function ref.

### LINKED ISSUE #15129: [vim9class] incorrect `string()` of funcref of object's function
### Steps to reproduce

```
vim9script

class C
    def C_F()
    enddef
endclass

var o = C.new()
var F = o.C_F
echo string(F)
```
outputs
```
function('g:C_F')
```

### Expected behaviour

I'm not sure what's correct, `function('object of C: C_F')`  seems weird but it does match a few patterns.

### Version of Vim

9.1.516

### Environment

ubuntu/gtk3

### Logs and stack traces

_No response_

errael:
Fix #15129.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -5923,6 +5923,37 @@ func_tv2string(typval_T *tv, char_u **tofree, int echo_style)
 }
 
 /*
+ * Return a textual representation of the object method in "tv", a VAR_PARTIAL.
+ * If the memory is allocated "tofree" is set to it, otherwise NULL.
+ * When "echo_style" is FALSE, put quotes around the function name as
+ * "function()", otherwise does not put quotes around function name.
+ * May return NULL.
+ */
+    static char_u *
+method_tv2string(typval_T *tv, char_u **tofree, int echo_style)
+{
+    char_u	buf[MAX_FUNC_NAME_LEN];
+    partial_T	*pt = tv->vval.v_partial;
+
+    size_t len = vim_snprintf((char *)buf, sizeof(buf), "<SNR>%d_%s.%s",
+			   pt->pt_func->uf_script_ctx.sc_sid,
+			   pt->pt_func->uf_class->class_name,
+			   pt->pt_func->uf_name);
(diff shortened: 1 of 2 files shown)


errael:
Rebased.

@yegappan Could you  take a look? A question is if the check for a partial being a class method is OK.

errael:
Huh, I just noticed that there's a comment I made that I can't find. Guess I didn't push the button.

This change brings the output of `string()` for an object function in line with what is looks like in a stack trace, #14390

errael:
There's an error in `vimtags`, doesn't look familiar; maybe from the rebase. Is there anything I need to do?

chrisbra:
> There's an error in vimtags, doesn't look familiar; maybe from the rebase. Is there anything I need to do?

that is strange, running `cd runtime/doc; make; git diff` is clean here. Looking at the diff in CI seems like it suddenly adds tags which are inside code samples in the help file. However, I cannot reproduce the failure on my side. It's clean here...

chrisbra:
this should have been fixed by #10813 (Patch v9.0.0110), not sure why this happens now in CI again.

errael:
> this should have been fixed by #10813 (Patch v9.0.0110), not sure why this happens now in CI again.

Fix for this found in https://github.com/vim/vim/pull/15099#issuecomment-2211425224

chrisbra:
Thanks, I think this makes sense. Not sure about the question for partials

errael:
1 char change to test in PR. Need to escape a `.` in `assert_match` pattern.

errael:
> > A question is if the check for a partial being a class method is OK.

> Not sure about the question for partials

We can say that it is not not OK since the test does work.

The `pt_obj` field was added with https://github.com/vim/vim/commit/29bb67f1beefc7fd393dbfd9ee77d92f1db3a3c0 specifically for "object funcref members" so it's probably a good check.

chrisbra:
thanks, let me include it now.

PR #15144, opened 2024-07-03
runtime(terraform): Adds filetype plugin for terraform

JannoTjarks:
Adds filetype plugin for terraform

- Sets commentstring property for terraform
- Sets comments property for terraform

### DIFF
--- a/runtime/ftplugin/terraform.vim
+++ b/runtime/ftplugin/terraform.vim
@@ -3,6 +3,11 @@
 " Maintainer: Janno Tjarks (janno.tjarks@mailbox.org)
 " Last Change: Jul 3, 2024
 
+if exists('b:did_ftplugin')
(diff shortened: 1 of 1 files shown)


runtime/ftplugin/terraform.vim
```diff
@@ -0,0 +1,9 @@
+" Vim filetype plugin
+" Language: terraform
+" Maintainer: Janno Tjarks (janno.tjarks@mailbox.org)
+" Last Change: Jul 3, 2024
+
```
zeertzjq, line 5: This is missing the `b:did_ftplugin` block.
```suggestion

if exists('b:did_ftplugin')
  finish
endif
let b:did_ftplugin = 1

```

chrisbra:
thanks!

Issue #14330, opened 2024-03-29
[vim9class] enum_value->string() issues


errael:
### Steps to reproduce

string(enum_value) no longer includes variable values.

### Expected behaviour

@zzzyxwvut @yegappan

A `class`', and initially an `enum`'s, default `string()` method provides useful
debug information by including instance variables in the string return.
If wanted it can be overridden to something more useful to the programmer.

In a recent change to `enum` classes, the default `string()` method now returns
something of the form "Planet.earth" rather than output something like a `class`.
I believe this change was made related to 

```
class variables leak into "values".
```

in comment https://github.com/vim/vim/pull/14224#issuecomment-2023939547,
if `values` refers to an `enum`'s static `values`, like `list<Planet>`,
then _the statement doesn't make sense_.
`string(obj)` has always included class variables.

The change to `string(enum_value)` removes some useful
functionality: the display of instance contents. It also diverges enum from class.

There is also, in https://github.com/vim/vim/pull/14224#issuecomment-2023172587
```
Suppose I want to store enumeration values as dictionary
keys, and use string() for them. How can I later undo
that conversion, as I get E121 with eval(string(Foo.BAR))?
```

I didn't fully understand this on first read. Looking again, it seems that
the idea is to be able to use both "Planet.earth" and "Count.ONE" as keys
in a dict and reconstruct the original enum value using `eval(key)`.

The change to `enum`'s `string()` also handles this, but breaks expected
usage of `string()` for a class.

Here's a suggestion to facilitate constructing a dict key for an enum.
Add an instance variable `class_name` to an enum value. Then the developer
can do
```
    var e = Planet.earth
    some_dict[e.class_name .. '.' .. e.name] = something
```
A similar possibility would be adding a builtin method/var `AsKey()`/`as_key`
to the enum instance; then `e.as_key->split('.')[0]` is `Planet`.

Note that any of these changes to string() for specific purposes, like generating
a dict key, seems to interfere/conflict with the idea of being able to override `string()`a.

If `class_name` isn't needed/wanted, then consider something specific to enums.

### Version of Vim

9.1.225

### Environment

ubuntu/gtk

yegappan:
[...]

PR #15145, opened 2024-07-03
Add builtin `id(:any)`, returns a string (the address) of the arg.

errael:
There is a discussion, #14374, about using a `class` or `enum` as a `dict` key.
Consider java's `HashMap` and `IdentityHashMap`.

This PR provides a way to have an identity `dict` using `dict` or `list` or
`object` as the key. See below for an example comparing the difference. Without
something like this, it is impossible to create an efficient, or even close to
efficient, identity map.

I went with a short name, `id`, as is my wont. Maybe `identity` is better.
No help/tests yet, looking for feedback first.

job and channel could be included or anything that has a refcount. LMK.

This should be considered an initial implementation and probably should be
extended, or supplanted, in the future. Some issues are
1. Must manually keep a reference to an item used as a key. The `id()` call
   does not increment the refcount.
2. Could add an option to `id()` to take a reference. But this requires
   considerable consideration. For example need a decref and avoid SEGV.

Returns empty string if not `VAR_OBJECT`/`VAR_DICT`/`VAR_LIST`.

```vim
vim9script

var d1: dict<any>
var d2: dict<any>

d1 = {a: 1}
d2 = {a: 2}

var d: dict<any>
d[string(d1)] = d1
d[string(d2)] = d2

echo d

d->filter((_, _) => false)
d2.a = 1
d[string(d1)] = d1
d[string(d2)] = d2
echo d

# Now the same thing with "id()"

d->filter((_, _) => false)
d1 = {a: 1}
d2 = {a: 2}
d[id(d1)] = d1
d[id(d2)] = d2
echo d

d->filter((_, _) => false)
d2.a = 1
d[id(d1)] = d1
d[id(d2)] = d2
echo d
```

**using `string()`**
Equality is used as the index.
```
{'{''a'': 2}': {'a': 2}, '{''a'': 1}': {'a': 1}}
{'{''a'': 1}': {'a': 1}}
```

**using `id()`**
Identity is used as the index.
```
{'0x5e9614976fc0': {'a': 2}, '0x5e961497ab60': {'a': 1}}
{'0x5e9614976fc0': {'a': 1}, '0x5e961497ab60': {'a': 1}}
```

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -309,6 +309,7 @@ hlget([{name} [, {resolve}]])	List	get highlight group attributes
 hlset({list})			Number	set highlight group attributes
 hostname()			String	name of the machine Vim is running on
 iconv({expr}, {from}, {to})	String	convert encoding of {expr}
+id({item})			String	get address of item as a string
 indent({lnum})			Number	indent of line {lnum}
 index({object}, {expr} [, {start} [, {ic}]])
 				Number	index in {object} where {expr} appears
@@ -5600,6 +5601,34 @@ iconv({string}, {from}, {to})				*iconv()*
 
 		Can also be used as a |method|: >
 			GetText()->iconv('latin1', 'utf-8')
+<
+		Return type: |String|
+
+
+id({item})							    *id()*
+		The result is a unique String associated with the {item} and
+		not with the {item}'s contents. It is only valid while the
+		{item} exists and is referenced. It is valid only in the
+		instance of vim that produces the result. The whole idea is
+		that `id({item})` does not change if the contents of {item}
+		changes. This is useful as a `key` for creating an identity
+		dictionary, rather than one based on equals.
+
+		This operation does not reference {item} and there is no
+		function to convert the `id` to the {item}. It may be useful to
+		have a map of `id` to {item}. The following >
+		    var referenceMap: dict<any>
+		    var id = item->id()
+		    referenceMap[id] = item
+<		prevents {item} from being garbage collected and provides a
+		way to get the {item} from the `id`.
+
+		{item} may be a List, Dictionary, Object, Job, Channel or
+		Blob. If the item is not a permitted type, or it is a null
+		value, then an empty String is returned.
+
+		Can also be used as a |method|: >
+			GetItem()->id()
 <
 		Return type: |String|
 
--- a/src/evalfunc.c
+++ b/src/evalfunc.c
@@ -82,6 +82,7 @@ static void f_haslocaldir(typval_T *argvars, typval_T *rettv);
 static void f_hlID(typval_T *argvars, typval_T *rettv);
 static void f_hlexists(typval_T *argvars, typval_T *rettv);
 static void f_hostname(typval_T *argvars, typval_T *rettv);
+static void f_id(typval_T *argvars, typval_T *rettv);
 static void f_index(typval_T *argvars, typval_T *rettv);
 static void f_indexof(typval_T *argvars, typval_T *rettv);
 static void f_input(typval_T *argvars, typval_T *rettv);
@@ -2207,6 +2208,8 @@ static funcentry_T global_functions[] =
 			ret_string,	    f_hostname},
     {"iconv",		3, 3, FEARG_1,	    arg3_string,
 			ret_string,	    f_iconv},
+    {"id",		1, 1, FEARG_1,	    NULL,
+			ret_string,	    f_id},
     {"indent",		1, 1, FEARG_1,	    arg1_lnum,
 			ret_number,	    f_indent},
     {"index",		2, 4, FEARG_1,	    arg24_index,
@@ -7484,6 +7487,40 @@ f_hostname(typval_T *argvars UNUSED, typval_T *rettv)
     mch_get_host_name(hostname, 256);
     rettv->v_type = VAR_STRING;
     rettv->vval.v_string = vim_strsave(hostname);
+}
+
+/*
+ * "id()" function
+ * Identity. Return address of item as a hex string, %p format.
+ * Currently only valid for object/container types.
+ * Return empty string if not an object.
+ */
+    void
+f_id(typval_T *argvars, typval_T *rettv)
+{
+    char_u	numbuf[NUMBUFLEN];
+
+    switch (argvars[0].v_type)
+    {
+	case VAR_LIST:
+	case VAR_DICT:
+	case VAR_OBJECT:
+	case VAR_JOB:
+	case VAR_CHANNEL:
+	case VAR_BLOB:
+	    // Assume pointer value in typval_T vval union at common location.
+	    if (argvars[0].vval.v_object != NULL)
+		vim_snprintf((char*)numbuf, sizeof(numbuf), "%p",
+					    (void *)argvars[0].vval.v_object);
+	    else
+		numbuf[0] = NUL;
+	    break;
+	default:
+	    numbuf[0] = NUL;
+    }
+
+    rettv->v_type = VAR_STRING;
+    rettv->vval.v_string = vim_strsave(numbuf);
 }
 
 /*
--- a/src/testdir/test_listdict.vim
+++ b/src/testdir/test_listdict.vim
@@ -1619,4 +1619,34 @@ func Test_deep_nested_listdict_compare()
   call v9.CheckLegacyAndVim9Success(lines)
 endfunc
 
+" Test for using id()
+def Test_id_with_dict()
+  # demonstate a way that "id(item)" differs from "string(item)"
+  var d1 = {one: 1}
+  var d2 = {one: 1}
+  var d3 = {one: 1}
+  var idDict: dict<any>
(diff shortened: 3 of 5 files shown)


errael:
Went ahead and added `blob`/`job`/`channel`.

src/evalfunc.c
```diff
@@ -7486,6 +7489,38 @@ f_hostname(typval_T *argvars UNUSED, typval_T *rettv)
...
+    {
+	case VAR_LIST:
+	case VAR_DICT:
+	case VAR_OBJECT:
+	case VAR_JOB:
+	case VAR_CHANNEL:
+	case VAR_BLOB:
+	    // Assume pointer value in typval_T vval union at common location.
+	    vim_snprintf((char*)numbuf, sizeof(numbuf), "%p",
+					    (void*)argvars[0].vval.v_object);
```
zeertzjq, line 7513: Maybe you can use `vim_vsnprintf_typval()` instead?

zeertzjq: Oh, that won't work.

errael:
@zeertzjq Do those messages mean you approve this PR ;-)

runtime/doc/builtin.txt
```diff
@@ -5604,6 +5605,27 @@ iconv({string}, {from}, {to})				*iconv()*
...
+		is only valid while the {item} exists and is referenced. The
+		returned value is useful for implementing an identity hash.
+		This operation does not reference to the item and there is no
+		function to convert the `id` to the item; it may be useful to
+		have a map of `id` to {item}, for example: >
+		    var referenceMap: dict<any>
+		    var id = item->id()
+		    referenceMap[id] = item
+<
+		{item} may be a List, Dictionary Object, Job, Channel or Blob.
```
```suggestion
		{item} may be a List, Dictionary, Object, Job, Channel or Blob.
```

errael:
Rebased.

errael:
> Thanks, I think this is useful. However I think we should emphasize that this is machine dependent and is not portable (e.g. when using sessions, etc).

Did some wordsmithing to the help text, and added a simple test.

I need some clarification. I don't understand `sessions` (I've never knowingly used them), I scanned `:he Session` but I don't see how this interacts with that.

Is there further text and/or changes required to address any concerns?

chrisbra:
> I need some clarification. I don't understand sessions (I've never knowingly used them), I scanned :he Session but I don't see how this interacts with that.

Sessions may save global variables, if 'sessionoptions' contains "global" and if the variable starts with an uppercase letter, see `:h 'sessionoptions'`. But according to the documentation this is only supported for string and number variables. I am pretty sure I contributed a patch to fix this several years ago.... Oh wait, it wasn't for session files, but for the viminfo file, see `:h 'viminfo'`. So if you include the flag `!` in your `viminfo` setting, Vim will store global variables (consisting only of upper case letters, grml, that is quite inconsistent...) in the viminfo file. It works for Dicts and Lists at least but not for deeply nested variables I think and I have actually no idea, how it behaves for the new types channels, jobs, blobs, etc. 

So long story short, I suppose this won't work properly if such a variable is stored in the viminfo file. But I see you addressed this already in the updated documentation, so this is probably good enough.

zeertzjq (commented):
Maybe you should also document how it treats `null_dict` etc.?

src/testdir/test_listdict.vim
```diff
@@ -1619,4 +1619,28 @@ func Test_deep_nested_listdict_compare()
...
+
+    var stringDict: dict<any>
+    stringDict[string(d1)] = d1
+    stringDict[string(d2)] = d2
+    stringDict[string(d3)] = d3
+    assert_equal(1, stringDict->len())
+
+    assert_equal('', id(3))
+  END
+  v9.CheckSourceSuccess(lines)
```
zeertzjq: Using `CheckSourceSuccess` here is a bit strange, as for this function it's not really different from putting these lines in the test function directly, and unlike `CheckLegacyAndVim9Success` it only runs these lines once.

errael:
> Maybe you should also document how it treats `null_dict` etc.?

Thanks, I missed that entirely. Currently it returns `"(nil)"` for any of the `null_*`.

@zeertzjq, @chrisbra What about returning `"null_list"`/`"null_dict"`/... or whatever?

src/testdir/test_listdict.vim
errael: I'm fuzzy about the test infrastructure. I guess I was thinking that using `v9.Check*` also gives you a context so you don't need to clean up. I can make that change. I guess `:unlet ...` is required then?

errael: Oh, in any event, it should have been `CheckScriptSuccess()` (copy/paste blunder).

zzzyxwvut:
A few anomalies with `b92408660`, I think:

```vim
echo printf('"%s"', id(null_string))
echo printf('"%s"', id(""))

try
    echo id(null_class)
catch
    echo v:exception
endtry
```

errael:
> A few anomalies with `b92408660`, I think:
> 
> ```viml
> echo printf('"%s"', id(null_string))
> echo printf('"%s"', id(""))
> 
> try
>     echo id(null_class)
> catch
>     echo v:exception
> endtry
> ```

The exception with `id(null_class)` is expected. Notice that `get(null_class, '')`, or `has(null_class)`, or `len(null_class)` all do the same thing.

zzzyxwvut:
My bad, I read ‘If the item is not a permitted type, then an  
empty String is returned’ from the documentation for `id()` 
_into_ the output of `id(null_class)`.

I would argue for `id()` to return either an empty string for  
invalid arguments (and _any_ `null*`s) or ‘an address’ string.

zzzyxwvut:
And let `id(null_class)` raise an exception.

chrisbra:
Agree with the empty string

src/testdir/test_listdict.vim
errael: > Oh, in any event, it should have been `CheckScriptSuccess()` (copy/paste blunder).

Ah, but wait... I just looked a little deeper into `v9.CheckSource` vs `v9.CheckScript`. I'm now thinking that `v9.CheckSource` is correct here; it starts with a new window, puts the code in it, sources the code from the buffer; `:bwipes` the buffer when done; whereas `v9CheckScript` creates a file.

errael:
Also return an empty string for a null type.

src/testdir/test_listdict.vim
zeertzjq: > I'm fuzzy about the test infrastructure. I guess I was thinking that using `v9.Check*` also gives you a context so you don't need to clean up. I can make that change. I guess `:unlet ...` is required then?

The variables here are all local variables, so no `:unlet` is required if you put the lines directly in the function.

runtime/doc/builtin.txt
```diff
@@ -5604,6 +5605,34 @@ iconv({string}, {from}, {to})				*iconv()*
...
+		have a map of `id` to {item}. The following >
+		    var referenceMap: dict<any>
+		    var id = item->id()
+		    referenceMap[id] = item
+<		prevents {item} from being garbage collected and provides a
+		way to get the {item} from the `id`.
+
+		{item} may be a List, Dictionary, Object, Job, Channel or
+		Blob. If the item is not a permitted type, or it is a null
+		type, then an empty String is returned.
```
zeertzjq: It's more often called "null value".
```suggestion
		value, then an empty String is returned.
```

zzzyxwvut:
I played with `5539ca` a little: took `adt.vim` from [here](https://github.com/vim/vim/issues/14330#issuecomment-2028938515) and  
`abc_gen.vim` from [here](https://github.com/vim/vim/issues/14330#issuecomment-2033163775) and rewritten the original `test.vim`:

```
vim9script

# Generate A-Z Letter implementations and declare "g:abc_src".
source abc_gen.vim

execute 'import "' .. g:abc_src .. '" as abc'
import './adt.vim'

# Create a Set (i.e. dict<number>) of abc.Letter elements.
const set: adt.Set = adt.Set.newFromList(
    # Populate this Set with elements from a list<abc.Letter>.
    abc.LETTERS,

    # Provide a lambda expression to map elements to dictionary keys.
    # (Available as set.ToStringer().)
    (o: abc.Letter): string => id(o),

    # Provide a lambda expression to map dictionary keys to elements.
    # (Here, partially apply a generated list that maps its indices to
    # object identities.)
    # (Available as set.FromStringer().)
    ((ii: list<string>) => (id: string): abc.Letter =>
			abc.LETTERS[index(ii, id)])(abc.LETTERS
					->mapnew((_, v) => id(v))))

# Recover all original objects.
for object in set.Elements()
    echo string(object) object is abc.ABC.Get(object.AsString())
endfor

echo set.FromStringer(set.ToStringer(abc.ABC.Get('a')))
echo set.ToStringer(abc.ABC.Get(tolower('A')))
echo set.Contains(abc.ABC.Get('Ω'))
```

The `id()` function behaves as advertised.

zeertzjq (approved):

chrisbra:
thanks all!

PR #15153, opened 2024-07-05
Tests for object equals with deeply nested objects.

errael:
Additional tests.
Follow on to #15081, #15085, #15070.


chrisbra:
Thanks!

PR #15159, opened 2024-07-06
runtime: CUDA ftplugin file

ribru17:
Filetype plugin for [CUDA](https://docs.nvidia.com/cuda/cuda-c-programming-guide/) files


chrisbra:
thanks!

PR #15163, opened 2024-07-06
runtime: update kivy syntax, add ftplugin support

ribru17:
Kivy uses "#:" for preprocessing commands (like "#:import ...") which were overridden by the comment syntax. This has been changed, and a commentstring has been added.


chrisbra:
thanks!

PR #15164, opened 2024-07-06
runtime(syntax-tests): Stop generating redundant `*_99.dump` files

zzzyxwvut:
- Remove the code handling their generation.
- Remove the code written for the shell filetype to clear up  
  the buffer screen before such a file can be generated.
- Remove all but `yaml.yaml_99.dump` of such generated files  
  (since there is no published `input/yaml.yaml.yaml`, let  
  its maintainer regenerate screen dump files and remove  
  redundant ones).

The new algorithm turns a `*_99.dump` file obsolete because  
all lines written in such a file will already be present in  
lower-numbered `*.dump` files.


chrisbra:
thanks!

chrisbra:
included as of b5a92d72eb494a4a8a330bbd266fb4c4c5bef366

PR #15166, opened 2024-07-06
runtime: add mysql ftplugin

ribru17:
This one just inherits from the SQL ftplugin. Without this, *.mysql files don't inherit the common SQL settings.


chrisbra:
thanks!

PR #15168, opened 2024-07-06
runtime: tf ftplugin file

ribru17:
Adds ftplugin support for tf (TinyFugue). Comment support taken from [here](https://github.com/kruton/tinyfugue/blob/1e8ac0bb014036c07bb3c679b0292ef20a6a0bb5/src/command.c#L568)


chrisbra:
thanks!

PR #15169, opened 2024-07-06
Unused assignment in sign_define_cmd()

zeertzjq:
Problem:  Unused assignment in sign_define_cmd().
Solution: Remove the assignment.  Also document the "priority" flag of
          sign_define().


chrisbra:
ah thanks, I noticed the same from Coverity.

PR #15170, opened 2024-07-06
runtime: objcpp ftplugin

ribru17:
Just inherits the base `objc` ftplugin


chrisbra:
I suppose this should source the corresonding ftplugin instead of the syntax file?

PR #15171, opened 2024-07-06
runtime: mojo ftplugin file

ribru17:
Take from excerpts of the Python ftplugin and adapted

### DIFF
--- a/runtime/ftplugin/mojo.vim
+++ b/runtime/ftplugin/mojo.vim
@@ -1,0 +1,41 @@
+" Vim filetype plugin
+" Language:	Mojo
+" Maintainer:	Riley Bruins <ribru17@gmail.com>
+" Last Change:	2024 Jul 07
+
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin = 1
+
+setlocal include=^\\s*\\(from\\\|import\\)
+setlocal define=^\\s*\\(\\(async\\s\\+\\)\\?def\\\|class\\)
+
+" For imports with leading .., append / and replace additional .s with ../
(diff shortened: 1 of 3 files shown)


runtime/ftplugin/mojo.vim
```diff
@@ -0,0 +1,45 @@
...
+" Maintainer:	Riley Bruins <ribru17@gmail.com>
+" Last Change:	2024 Jul 06
+
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin = 1
+
+setlocal cinkeys-=0#
+setlocal indentkeys-=0#
```
chrisbra: Those two belong into the corresponding indent plugin.

runtime/indent/mojo.vim
```diff
@@ -0,0 +1,15 @@
...
+" Last Change:	2024 Jul 07
+
+" Only load this indent file when no other was loaded.
+if exists("b:did_indent")
+   finish
+endif
+let b:did_indent = 1
+
+setlocal cinkeys-=0#
+setlocal indentkeys-=0#
```
chrisbra, line 13: I suppose you would also want `:setl cindent`?

ribru17: I'm not sure; these were taken verbatim from the python `ftplugin`, but python also has it's own conflicting indent settings in its indent plugin (which also sets `indentexpr` rather than `cindent`). Perhaps mojo should just inherit this indent file and call it a day? (Mojo syntax is a superset of python's)

chrisbra: Sure, you can alternatively just source the python indent script if that works.

ribru17:
Sorry for the delay - thank you for picking this up!!

PR #15172, opened 2024-07-06
runtime: squirrel ftplugin

ribru17:
Comment spec found [here](https://squirrel-lang.org/squirreldoc/reference/language/lexical_structure.html#comments)


chrisbra:
thanks!

PR #15174, opened 2024-07-06
runtime: lex ftplugin

ribru17:
Spec found [here](https://docs.oracle.com/cd/E19504-01/802-5880/lex-6/index.html)


chrisbra:
thanks!

PR #15175, opened 2024-07-07
runtime: cmakecache ftplugin file

ribru17:
[Specification here](https://cmake.org/cmake/help/book/mastering-cmake/chapter/CMake%20Cache.html)


chrisbra:
thanks!

PR #15176, opened 2024-07-07
runtime: ldapconf ftplugin file, file recognition

ribru17:
[Specification](https://www.openldap.org/software//man.cgi?query=ldap.conf&sektion=5&apropos=0&manpath=OpenLDAP+2.4-Release)


chrisbra:
thanks!

PR #15177, opened 2024-07-07
fix: failing test `Test_null_values` when vim configured without channel

dpelle:
Test `Test_null_values` was failing when vim is configured without channel because the test uses function `null_channel()` (which requires the `+channel` feature) and using `null_job()` which requires the `+job` feature. But was reproducible when vim is configured with:
```
./configure --with-features=huge --enable-gui=none --disable-channel
```

Only check `null_channel()` when the `+channel` feature is available. Only check `null_job()` when the `+job` feature is available.


chrisbra:
thanks!

chrisbra:
thanks!

PR #15180, opened 2024-07-08
doc example fix: autocmd accepts list not dict

EdwarDu:
doc example is missing square brackets for autocmd accepts list not dict.


chrisbra:
Thanks, but the same should be done a few lines below for the legacy script example

PR #15182, opened 2024-07-08
Add KeyInputPre autocmd

Shougo:
I have added `KeyInputPre` autocmd.

It works like neovim's on_key handler.

~~NOTE:  It is WIP.  Because no tests.~~


Shougo:
This is similar with `InsertCharPre` but

* It works all modes
* It can handle non character code

Shougo:
It is useful for example: https://github.com/tani/dmacro.nvim

Shougo:
I have added tests.

src/getchar.c
```diff
@@ -2138,6 +2145,62 @@ vgetc(void)
...
+    // Lock the text to avoid weird things from happening.
+    ++textlock;
+    set_vim_var_string(VV_CHAR, buf, -1);  // set v:char
+
+    if (apply_autocmds(EVENT_KEYINPUTPRE, curr_mode, curr_mode, FALSE, curbuf)
+	&& STRCMP(buf, get_vim_var_str(VV_CHAR)) != 0)
+    {
+	// Get the value of v:char.  It may be empty or more than one
+	// character.  Only use it when changed, otherwise continue with the
+	// original character to avoid breaking autoindent.
```
k-takata: The comment "to avoid breaking autoindent" might be irrelevant.

Shougo: Fixed.

Shougo:
I will add `v:event.typed` to check typed state.

runtime/doc/autocmd.txt
```diff
@@ -977,6 +979,15 @@ InsertLeavePre			Just before leaving Insert mode.  Also when
 							*InsertLeave*
 InsertLeave			Just after leaving Insert mode.  Also when
 				using CTRL-O |i_CTRL-O|.  But not for |i_CTRL-C|.
+							*KeyInputPre*
+KeyInputPre			When a key is typed. The pattern is
+				matched against `'current_mode'`.
```
zeertzjq: Why is `current_mode` quoted?

src/getchar.c
```diff
@@ -2138,6 +2145,62 @@ vgetc(void)
...
+    int		save_State = State;
+
+    res = c;
+
+    // Return quickly when there is nothing to do.
+    if (!has_keyinputpre())
+	return res;
+
+    buf[0] = c;
+    buf[1] = NUL;
```
zeertzjq: I don't think this is correct. This doesn't work with either multibyte chars or special keys.

src/getchar.c
```diff
@@ -2138,6 +2145,62 @@ vgetc(void)
...
+	// character.  Only use it when changed, otherwise continue with the
+	// original character.
+	char_u *v_char;
+
+	v_char = get_vim_var_str(VV_CHAR);
+
+	// Convert special bytes when it is special string.
+	if (STRLEN(v_char) == 3 && v_char[0] == K_SPECIAL)
+	    res = TERMCAP2KEY(v_char[1], v_char[2]);
+	else if (STRLEN(v_char) == 1)
```
zeertzjq: This doesn't work with multibyte chars either.

runtime/doc/autocmd.txt
```diff
@@ -977,6 +979,15 @@ InsertLeavePre			Just before leaving Insert mode.  Also when
 							*InsertLeave*
 InsertLeave			Just after leaving Insert mode.  Also when
 				using CTRL-O |i_CTRL-O|.  But not for |i_CTRL-C|.
+							*KeyInputPre*
+KeyInputPre			When a key is typed. The pattern is
```
zeertzjq: I don't think "typed" is an accurate description of when this event is triggered. This is triggered before a key is processed, and the key may come from a mapping (which is not a typed key).

src/getchar.c
Shougo: Can you fix it for multibytes?

src/getchar.c
Shougo: Yes.  It does not support multibytes.

runtime/doc/autocmd.txt
k-takata: Something like this?
```suggestion
				matched against a string that indicates the
				current mode, which is the same as that is
				returned by `mode(1)`.
```
(Sorry if my English is not correct.)

zeertzjq: It may be better to use `mode(1)` here to be more accurate, otherwise it seems OK.

src/getchar.c
zeertzjq:
```suggestion
    if (IS_SPECIAL(c))
    {
	buf[0] = K_SPECIAL;
	buf[1] = KEY2TERMCAP0(c);
	buf[2] = KEY2TERMCAP1(c);
	buf[3] = NUL;
    }
    else
	buf[(*mb_char2bytes)(c, buf)] = NUL;
```

src/getchar.c
```diff
@@ -2138,6 +2145,62 @@ vgetc(void)
...
+	    res = v_char[0];
```
k-takata: maybe?
```suggestion
	else if (MB_CHARLEN(v_char) == 1)
	    res = PTR2CHAR(v_char);
```

runtime/doc/autocmd.txt
k-takata: Updated.

src/getchar.c
Shougo: Thanks.  But I will use another method.

runtime/doc/autocmd.txt
Shougo: Fixed.

runtime/doc/autocmd.txt
Shougo: Fixed.

src/getchar.c
```diff
@@ -2138,6 +2145,74 @@ vgetc(void)
...
+	// character.  Only use it when changed, otherwise continue with the
+	// original character.
+	char_u *v_char;
+
+	v_char = get_vim_var_str(VV_CHAR);
+
+	// Convert special bytes when it is special string.
+	if (STRLEN(v_char) == 3 && v_char[0] == K_SPECIAL)
+	    res = TERMCAP2KEY(v_char[1], v_char[2]);
+	else
```
k-takata: The comment in L2190 says "It may be empty...", so this might be better:
```suggestion
	else if (STRLEN(v_char) > 0)
```

src/getchar.c
```diff
@@ -2197,7 +2197,7 @@ do_key_input_pre(int c)
 	// Convert special bytes when it is special string.
 	if (STRLEN(v_char) == 3 && v_char[0] == K_SPECIAL)
```
zeertzjq:
```suggestion
	if (STRLEN(v_char) >= 3 && v_char[0] == K_SPECIAL)
```

runtime/doc/autocmd.txt
```diff
@@ -977,6 +979,20 @@ InsertLeavePre			Just before leaving Insert mode.  Also when
...
+							*KeyInputPre*
+KeyInputPre			Just before a key is processed. The pattern is
+				matched against a string that indicates the
+				current mode, which is the same as that is
+				returned by `mode(1)`.
+				The |v:char| variable indicates the key typed
+				and can be changed during the event to input
+				a different key.  When |v:char| is not single
+				character or one special key, it is the fist
+				character is used.
```
k-takata:
```suggestion
				a different key.  When |v:char| is not a
				single character or one special key, the first
				character is used.
```

chrisbra:
That sounds dangerous. I hope it's not possible to change the mode?

Shougo:
> That sounds dangerous. I hope it's not possible to change the mode?

The event cannot change the mode and text.
I will update the documentation.

chrisbra:
thanks!

PR #15183, opened 2024-07-08
Termdebug help add timeout info

EdwarDu:
My setup is using a debugger in a NFS which is slow to launch/startup, default 3s timeout if not enough.

Adding info for timeout in help and error message, though error message is not directly visible, but does show up in log file with -V option.

### DIFF
--- a/runtime/doc/terminal.txt
+++ b/runtime/doc/terminal.txt
@@ -1535,6 +1535,14 @@ If there is no g:termdebug_config you can use: >
 <
 However, the latter form will be deprecated in future releases.
 
+Startup timeout ~
+						*termdebug-timeout*
+The termdebug plugin waits debugger to startup for 3 seconds by default,
+however, if the debugger takes longer to launch, e.g. in a slow NFS, consider
+to increase the timeout with: >
+	let g:termdebug_config['timeout'] = 600
+This will set the timeout to 600 x 10 milliseconds = 6 seconds.
+
 Mappings ~
 The termdebug plugin enables a few default mappings.  All those mappings
 are reset to their original values once the termdebug session concludes.
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -517,7 +517,8 @@ def CreateGdbConsole(dict: dict<any>, pty: string, commpty: string): string
   endwhile
 
   if !success
-    return 'Failed to startup the gdb program.'
+    return 'Failed to startup the gdb program within '
+      .. counter_max .. ' x 10ms, may consider to increase timeout'
   endif
 
   # ---- gdb started. Next, let's set the MI interface. ---
@@ -576,7 +577,6 @@ enddef
 
 # Open a terminal window without a job, to run the debugged program in.
 def StartDebug_term(dict: dict<any>)
-
   var programpty = CreateProgramPty()
   if programpty is null_string
     Echoerr('Failed to open the program terminal window')
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -517,7 +517,7 @@ def CreateGdbConsole(dict: dict<any>, pty: string, commpty: string): string
   endwhile
 
   if !success
-    return 'Failed to startup the gdb program within '
+    return $'Failed to startup the gdb program within {counter_max}ms, consider increasing the timeout'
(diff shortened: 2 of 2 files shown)


runtime/doc/terminal.txt
```diff
@@ -1535,6 +1535,14 @@ If there is no g:termdebug_config you can use: >
 <
 However, the latter form will be deprecated in future releases.
 
+Startup timeout ~
+						*termdebug-timeout*
+The termdebug plugin waits debugger to startup for 3 seconds by default,
+however, if the debugger takes longer to launch, e.g. in a slow NFS, consider
+to increase the timeout with: >
+	let g:termdebug_config['timeout'] = 600
+This will set the timeout to 600 x 10 milliseconds = 6 seconds.
```
LemonBoy, line 1544: Why tens of milliseconds and not milliseconds?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -517,7 +517,8 @@ def CreateGdbConsole(dict: dict<any>, pty: string, commpty: string): string
   endwhile
 
   if !success
-    return 'Failed to startup the gdb program.'
+    return 'Failed to startup the gdb program within '
```
LemonBoy:
```suggestion
    return $'Failed to startup the gdb program within {counter_max}ms, consider increasing the timeout'
```

runtime/doc/terminal.txt
EdwarDu: don't know. The original code counter is by 10ms. direct ms could be more straightforward.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -517,7 +517,8 @@ def CreateGdbConsole(dict: dict<any>, pty: string, commpty: string): string
   endwhile
 
   if !success
-    return 'Failed to startup the gdb program.'
+    return $'Failed to startup the gdb program within {counter_max}ms, consider increasing the timeout'
+      .. counter_max .. ' x 10ms, may consider to increase timeout'
```
errael, line 521: This seems to have some issues.
1. There is an extra/leftover line
2. The message is inaccurate (it assumes counter_max is in milliseconds)
3. Missing '.'

```suggestion
    return $'Failed to startup the gdb program within {counter_max} x 10ms, consider increasing the timeout.'
```

EdwarDu: Yep, I accepted the change without checking carefully.
Now, I saw that someone already added timeout info in the terminal.txt for termdebug. And @LemonBoy mentioned that counter by 10ms is a bit strange, now I'm thinking might as well change the counter to ms?

I will withdraw this PR, and file another one.

I'm not very familiar with github PR flow, sorry for the inconvenience.

LemonBoy: I guess the granularity is 10ms to avoid having a busy polling loop here. The docs should be slightly amended to mention `timeout` is a timeout in terms of `10ms` ticks (or units).

EdwarDu: Well, counter can be made to +10 instead of +1 to align with ms. I think this way is less confusing.

errael: > mentioned that counter by 10ms is a bit strange, now I'm thinking might as well change the counter to ms?

Is that an incompatible change? Something that breaks existing config files? I wouldn't couple that with a help change.

No reason to open a different PR. Changes are made as needed and usually commits are squashed and/or force updated as things proceed.

LemonBoy: Something like `min([1, g:termdebug_config['timeout'] / 10])` would also work, I doubt anyone would notice the timeout is rounded to 10ms.

EdwarDu: > > mentioned that counter by 10ms is a bit strange, now I'm thinking might as well change the counter to ms?
> 
> Is that an incompatible change? Something that breaks existing config files? I wouldn't couple that with a help change.
> 
> No reason to open a different PR. Changes are made as needed and usually commits are squashed and/or force updated as things proceed.

Yes, it will be existing configuration timeout /10 as intended, if timeout early, there is a problem, the "debugged program" terminal window will not show, menu will not show etc.

Hmm, if going this way, this PR is not really necessary any more, someone already added the timeout section in the doc, and the error message anyway is not visible without -V option and check the log afterwards.

PR #15184, opened 2024-07-08
Correct shell command to get $VIMRUNTIME in help doc for $VIMRUNTIME.

shaneharper:
The original shell command could set the VIMRUNTIME variable to an invalid path.
    - since v9.0.0592 trailing spaces were typically added making the path invalid. (I observed that spaces were added if the length of the path wasn't equal to one less than an integer multiple of the number of display columns.)
    - newline character/s would be added, making the path invalid, if the path was longer than the number of display columns.

The new command is shorter than the original command.


chrisbra:
thanks!

PR #15186, opened 2024-07-08
Add a way to run the test suite in random order

LemonBoy:
When $TEST_RANDOMIZE is set the whole set of test functions is shuffled and then executed.
This greatly helps to ensure all the test cases are self-contained and prevent accidental order dependencies between them.

Whether or not enable it on one of the many CI jobs it's up to the management crew.

### DIFF
--- a/src/testdir/runtest.vim
+++ b/src/testdir/runtest.vim
@@ -586,8 +586,20 @@ if $TEST_MAY_FAIL != ''
   let s:may_fail_list = split($TEST_MAY_FAIL, ',')->map({i, v -> v .. '()'})
 endif
 
-" Execute the tests in alphabetical order.
-for g:testfunc in sort(s:tests)
+if $TEST_RANDOMIZE != ''
+  " Execute the tests in random order, this helps to ensure tests are isolated from each
+  " other.
+  call add(s:messages, 'Running tests in random order')
+  for i in range(len(s:tests))
+    let j = rand() % len(s:tests)
+    let [s:tests[i], s:tests[j]] = [s:tests[j], s:tests[i]]
+  endfor
+else
+  " Execute the tests in alphabetical order.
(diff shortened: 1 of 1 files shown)


zeertzjq:
Some tests requires being run as the first (because they require a clean state) or the last (because they have irreversible effects) tests, so it's not possible to prevent order dependencies. These tests typically have names that start with `Test_00`, `Test_aa` or `Test_zz`.

chrisbra:
That is an interesting experiment, but as mentioned there are a few tests, that depend on the state, so I would be expecting some failures already. And I would hate a failing CI just because the order of the tests changed between runs.

I think a bit more beneficial for the test suite would be, if we could divide the list of test-files into a few buckets and then try to run vim for each bucket in parallel to speed up testing.

chrisbra:
BTW: I just stumbled over https://github.com/mxschmitt/action-tmate I wonder if this could make debugging failing scripts easier, because it allows to ssh into the runner directly and inspect it directly.

LemonBoy:
> Some tests require being run as the first (because they require a clean state) or the last (because they have irreversible effects) tests,

Such tests can be excluded from the shuffle if we decide on a consistent naming scheme and document it.

If there's no interest in getting this in let's save time and close the PR.

> I think a bit more beneficial for the test suite would be, if we could divide the list of test-files into a few buckets and then try to run vim for each bucket in parallel to speed up testing.

Well, if you can run them in random order you can surely run them in parallel.

errael:
> Well, if you can run them in random order you can surely run them in parallel.

If two tests use the same resource, they can not run in parallel.

I think a lot of work was done with the tests to insure unique file names for just this reason. I don't know if this is still an issue.

chrisbra:
Thanks for you work on this. But I am not going to merge this one. Sorry.

PR #15187, opened 2024-07-08
Fix MSVC warning. I can't test it, but now it matches the prototypes.

errael:
Change cast `(char*)` to `(char *)`. Doesn't seem required style, but the space is used over 98% of the time; so...


chrisbra:
thanks!

PR #15188, opened 2024-07-08
Add '%hidden' highlighting to tmux file type

lostl1ght:
Currently `%hidden` is not highlighted in tmux config files. This PR adds the missing highlighting.


ericpruitt:
I've filed https://github.com/vim/vim/pull/15195, so this can be closed.

PR #15190, opened 2024-07-09
runtime(filetype): Add detection for Angular HTML templates

dlvandenberg:
Since Angular 17, the new [Control Flow Syntax](https://angular.dev/guide/templates/control-flow) is not valid HTML. This PR adds a new filetype detection for the HTML templates of Angular. 

It first checks the filename. The Angular convention is to use `*.component.html` for the template. However, this is not mandatory.

If the filename does not match, it will check the contents of the file if it contains:
- One of the Control-Flow blocks: `@if`, `@for`, `@switch`, `@defer`
- A structural directive: `*ngIf`, `*ngFor`, `*ngSwitch`, `*ngTemplateOutlet`
- Builtin Angular elements: `ng-template` or `ng-content`
- String interpolation: {{ something }}

This enables the Angular LSP to attach only to `htmlangular` filetypes, as well as language parsers, such as tree-sitter.

I've added a simple ft test, but only on filename. I'd like to add a testcase to check the contents of the file, but I do not know how. Tips are welcome.


runtime/ftplugin/htmlangular.vim
```diff
@@ -0,0 +1,13 @@
...
+" Last Change: 2024 Jul 8
+
+" Only use this filetype plugin when no other was loaded.
+if exists("b:did_ftplugin")
+  finish
+endif
+
+" Use HTML and Angular template ftplugins
+runtime! ftplugin/html.vim
+runtime! ftplugin/htmlangular.vim
```
clason:
```suggestion
```

Or did you forget to commit a second `angular.vim` ftplugin (in which case you just need to set the filetype to `html.angular` to automate this)?

dlvandenberg: No, not sure what to put in such a file.. `angular.vim`

chrisbra:
> I've added a simple ft test, but only on filename. I'd like to add a testcase to check the contents of the file, but I do not know how. Tips are welcome.

Have a look https://github.com/vim/vim/blob/4a7a4a3675b6ad90a525524ba4684925df212325/src/testdir/test_filetype.vim#L2485-L2491

Nevermind, I'll add a test while merging.

zeertzjq:
As mentioned in https://github.com/vim/vim/pull/13594#issuecomment-1834465890 detecting angular files by filename is not desired. I think this is still the case.

runtime/ftplugin/htmlangular.vim
```diff
@@ -0,0 +1,12 @@
...
+" Language: Angular HTML Template
+" Maintainer: Dennis van den Berg <dennis@vdberg.dev>
+" Last Change: 2024 Jul 8
+
+" Only use this filetype plugin when no other was loaded.
+if exists("b:did_ftplugin")
+  finish
+endif
+
+" Use HTML and Angular template ftplugins
```
zeertzjq, line 11: This comment seems wrong now, as it only uses HTML ftplugin.

chrisbra:
Okay, we can revert looking at the filename and just make it inspect the file, when doing so, we can also fix the  comment. @dlvandenberg is that fine for you?

dlvandenberg:
> Okay, we can revert looking at the filename and just make it inspect the file, when doing so, we can also fix the comment. @dlvandenberg is that fine for you?

Yeah, sure.

PR #15192, opened 2024-07-09
revert: patch 9.1.0503

glepnir:
fuzzy can handle keyword completion as well. so fuzzycollect is no longer needed.

### DIFF
--- a/runtime/doc/version9.txt
+++ b/runtime/doc/version9.txt
@@ -41642,8 +41642,8 @@ Improved visual highlighting.
 
 Python3 support in OpenVMS.
 
-Support for |fuzzy-matching| during |ins-completion| with the "fuzzy" and
-"fuzzycollect" values of the 'completeopt' setting
+Support for |fuzzy-matching| during |ins-completion| with the "fuzzy"
+values of the 'completeopt' setting
 
 ==============================================================================
 COMPILE TIME CHANGES					*compile-changes-9.2*
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -2151,11 +2151,6 @@ A jump table for the options with a short description can be found at |Q_op|.
 		    difference how completion candidates are reduced from the
 		    list of alternatives, but not how the candidates are
 		    collected (using different completion types).
-	  fuzzycollect
-		    Enable fuzzy collection for default keyword completion.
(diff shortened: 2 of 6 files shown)


runtime/doc/version9.txt
```diff
@@ -41642,9 +41642,6 @@ Improved visual highlighting.
 
 Python3 support in OpenVMS.
 
-Support for |fuzzy-matching| during |ins-completion| with the "fuzzy" and
-"fuzzycollect" values of the 'completeopt' setting
-
```
zeertzjq: You should just remove "fuzzycollect" from the sentence instead of deleting the whole sentence.

glepnir: oops sry.

runtime/doc/version9.txt
```diff
@@ -41642,8 +41642,7 @@ Improved visual highlighting.
 
 Python3 support in OpenVMS.
 
-Support for |fuzzy-matching| during |ins-completion| with the "fuzzy" and
-"fuzzycollect" values of the 'completeopt' setting
+Support for |fuzzy-matching| during |ins-completion| with the "fuzzy".
```
zeertzjq, line 41645:
```suggestion
Support for |fuzzy-matching| during |ins-completion| with the "fuzzy"
value of the 'completeopt' option.
```

chrisbra:
thanks!

zeertzjq:
This change from patch 9.1.0503 hasn't been reverted:
```diff
diff --git a/src/insexpand.c b/src/insexpand.c
index c673df927..78fea5127 100644
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -4100,8 +4100,7 @@ find_comp_when_fuzzy(void)
     int		is_backward = compl_shows_dir_backward();
     compl_T	*comp = NULL;
 
-    if (compl_match_array == NULL ||
-	    (is_forward && compl_selected_item == compl_match_arraysize - 1)
+    if ((is_forward && compl_selected_item == compl_match_arraysize - 1)
 	    || (is_backward && compl_selected_item == 0))
 	return compl_first_match;
 
```
Does it need to be?

glepnir:
find_next_completion_match has checked the array is not NULL so there no need revert.

PR #15193, opened 2024-07-09
feat(completion): fuzzy on default completion

glepnir:
make `fuzzy` works on keyword completion.


glepnir:
need a rebase after revert pr merge.

chrisbra:
thanks. Can someone test this please?

techntools:
@chrisbra @glepnir 

Thanks for working on this.

 Its working but in the screencast below, is it updating the match message correctly ?


https://github.com/vim/vim/assets/71767293/d42c7bfc-3dcd-4294-a67d-5b4125fc5149

glepnir:
try remove fuzzy in completeopt there will still get a wrong status msg. `xyz xyb xbc` then x -><C-X><C-N> xy msg still is match x of 3 . because cp_matches not update.

techntools:
@chrisbra What do you think ?

src/insexpand.c
```diff
@@ -3716,6 +3721,85 @@ get_next_default_completion(ins_compl_next_state_T *st, pos_T *start_pos)
...
+		    circly_end.col = 0;
+		    circly_end.coladd = 0;
+		}
+
+		// Check if looped around and back to start position
+		if (looped_around && EQUAL_POS(current_pos, circly_end))
+		    break;
+
+		// Get the current line buffer
+		if (current_pos.lnum >= 1
```
LemonBoy: Can `current_pos` be invalid here?
Wrap-around is _not_ controlled by `p_ws`.
`p_ic` is _not_ respected.
This code should be wrapped in a function (e.g. `search_for_fuzzy_match`) similar to `search_for_exact_line`, you can re-use most of it.

src/insexpand.c
```diff
@@ -3716,6 +3721,85 @@ get_next_default_completion(ins_compl_next_state_T *st, pos_T *start_pos)
...
+		    && current_pos.lnum <= st->ins_buf->b_ml.ml_line_count
+		    && !(current_pos.col - 1 == ml_get_buf_len(st->ins_buf, current_pos.lnum)))
+		{
+		    ptr = ml_get_buf(st->ins_buf, current_pos.lnum, FALSE);
+		    if (ptr != NULL)
```
LemonBoy: This is never `NULL`.

src/insexpand.c
```diff
@@ -3716,6 +3721,85 @@ get_next_default_completion(ins_compl_next_state_T *st, pos_T *start_pos)
...
+			ptr += current_pos.col;
+		}
+
+		// If ptr is null or end of line is reached, move to next line or previous line based on direction
+		if (ptr == NULL || *ptr == NUL)
+		{
+		    if (compl_dir_forward())
```
LemonBoy: This block is identical to the one at the end of the loop.

chrisbra:
> @chrisbra What do you think ?

I find it hard to follow. What keys did you press? And does it also happen without fuzzy as @glepnir mentions?

techntools:
@chrisbra @glepnir 

It does also happen without fuzzy as @glepnir mentions

I started with ```vim -u NONE```. Used C-N.

To clear further, as seen in my ss:

1. I type x, press C-N, browse through candidates with C-N, message is ```match 1 of 4```
2. I type z after x while the completion menu still visible, completion menu shows total candidates are only 2, browse through candidates with C-N, message is ```match 1 of 4```. But I think message should be ```match 1 of 2```

But if I press ```ESC```, press C-N, completion menu opens and shows the right value for total matches

glepnir:
because `cp_matches` is the count of all candidate words. `statusmsg` uses it. But when the leader adds or deletes the pum array  regenerates , `cp_matches` is not updated. So it is still the initial count. It should be easy to fix. I just don't know why no one has raised an issue. The default completion has existed for a long time. Someone should have noticed it.

habamax:
Tried it, looks good to me.

habamax:
However the help topic now incorrect:


```
	   fuzzy    Enable |fuzzy-matching| for completion candidates. This
		    allows for more flexible and intuitive matching, where
		    characters can be skipped and matches can be found even
		    if the exact sequence is not typed.  Only makes a
		    difference how completion candidates are reduced from the
		    list of alternatives, but not how the candidates are
		    collected (using different completion types).

```

This part

```
                                                         Only makes a
		    difference how completion candidates are reduced from the
		    list of alternatives, but not how the candidates are
		    collected (using different completion types).
```

Currently it also affects how candidates are collected. At least for keywords. Let me check other methods.

habamax:
okay, there is a regression, at least for line completion, doesn't work at all when `fuzzy` is in `completeopt`.

@glepnir pls check following:

1. `vim -Nu NONE`
2. add 2 lines of text:
``` world is on fire
no one can save me but you```
3. `set completeopt+=fuzzy`
4. in the new line type `w<ctrl-x><ctrl-l>`

"world is on fire" should be completed. But it is not:

[![asciicast](https://asciinema.org/a/stdZ08FGmRwgK3glYskfw36Uc.svg)](https://asciinema.org/a/stdZ08FGmRwgK3glYskfw36Uc)

File: src/search.c [hunk 4]
```diff
@@ -5139,6 +5140,143 @@ fuzzy_match_str_with_pos(char_u *str UNUSED, char_u *pat UNUSED)
...
+	{
+	    *pos = current_pos;
+	    break;
+	}
+
+    move_line:
+	// Move to the next line or previous line based on direction
+	if (dir == FORWARD)
+	{
+	    if (++current_pos.lnum > buf->b_ml.ml_line_count)
```
LemonBoy, line 5275: Check for `p_ws` to know if wrapping is wanted or not & add test.

src/insexpand.c
```diff
@@ -3764,8 +3770,9 @@ get_next_default_completion(ins_compl_next_state_T *st, pos_T *start_pos)
 		&& start_pos->col  == st->cur_match_pos->col)
 	    continue;
 
-	ptr = ins_compl_get_next_word_or_line(st->ins_buf, st->cur_match_pos,
-							   &len, &cont_s_ipos);
+	if (len == 0)
```
LemonBoy: At first glance `len` may be uninitialized, let's use `in_fuzzy` .

src/search.c
```diff
@@ -5139,6 +5140,143 @@ fuzzy_match_str_with_pos(char_u *str UNUSED, char_u *pat UNUSED)
...
+	    circly_end.coladd = 0;
+	}
+
+	// Check if looped around and back to start position
+	if (looped_around && EQUAL_POS(current_pos, circly_end))
+	    break;
+
+	// Ensure current_pos is valid
+	if (current_pos.lnum < 1 || current_pos.lnum > buf->b_ml.ml_line_count)
+	    goto move_line;
```
LemonBoy: Again, can `current_pos` ever be invalid?
Can we avoid the `goto` ?

src/insexpand.c
glepnir: in process

src/search.c
glepnir: I ignored that comment sry.

Shane-XB-Qian:
> because `cp_matches` is the count of all candidate words. `statusmsg` uses it. But when the leader adds or deletes the pum array  regenerates , `cp_matches` is not updated. So it is still the initial count. It should be easy to fix. I just don't know why no one has raised an issue. The default completion has existed for a long time. Someone should have noticed it.

this maybe not a bug, here the second c-n after the first one was not same meaning, it still was the first compl.
// regardingless what is fuzzy or fuzzycollect going, wish not impact the original compl,
// though do welcome you to fix its original real bug if there, thx~

-- 
shane.xb.qian

mg979:
fuzz1

File completion doesn't use fuzzy when started.

fuzzy2

Line completion is broken with `completeopt+=fuzzy`.

started with `vim --clean -c 'so .testrc'`

.testrc
```vim
color habamax

hi link PmenuMatch PmenuKind
hi link PmenuMatchSel PmenuKindSel

set completeopt=menuone,popup,fuzzy,noinsert,noselect
inoremap <c-f> <c-x><c-f>
```

src/search.c
```diff
@@ -5139,6 +5140,167 @@ fuzzy_match_str_with_pos(char_u *str UNUSED, char_u *pat UNUSED)
...
+    while (*str != NUL)
+    {
+	// Skip non-word characters
+	start = find_word_start(str);
+	if (*start == NUL)
+	    break;
+	end = find_word_end(start);
+
+	// Extract the word from start to end
+	word = vim_strnsave(start, (int)(end - start));
```
mg979: I don't think it's needed to allocate a string here, this should work just the same (pls somebody correct me if I'm wrong)

```c
	// Store the character after word, then make it NUL
	// so that fuzzy_match_str stops right there
	char oldchar = *end;
	*end = NUL;

	// Perform fuzzy match
	result = fuzzy_match_str(start, pat);

	// Restore the character we made NUL before
	*end = oldchar;
```

At least it seems to work just fine for me. There's a hell lot of allocations being done there, otherwise.

mg979:
Just my thoughts. Without this PR fuzzy completion 'sort of works', that is:

1. it works with all methods (also default completion), but only if the popup is visible already
2. the completion leader is never matched in a fuzzy way (probably it should be)
3. fuzzy just doesn't work when completion is invoked on a pattern that you'd like to be fuzzy (because of 2.)

With this PR, points 2 and 3 above are not a problem for default completion, but they still are for other types of completion.
But it breaks line completion additionally, and overall I think there might be a much simpler solution to fix the problems above than doing all this. Of course I might be wrong.

glepnir:
sry I don't understand what do you mean.. file is not fuzzy because I haven't implemented it yet. will update later.

mg979:
> sry I don't understand what do you mean.. file is not fuzzy because I haven't implemented it yet. will update later.

This was your original comment, magically turned into

> sry I don't understand what do you mean. file fuzzy works for me

Why do you keep force-pushing? We'd like to see what changes between commits.

glepnir:
i edit comment :) because i can't reproduce on file completion before.  line has fixed .  it will always be squash merged in the end

mg979:
It doesn't matter if it's squashed or not, one can still see changes between commits after that, if you don't force-push. It's difficult to keep track of what you're doing this way.

src/search.c
glepnir: yup this is effective.

glepnir:
Aha okay. file completion I might try to fix that tomorrow.

src/Makefile
```diff
@@ -587,7 +587,7 @@ CClink = $(CC)
 # When using -g with some older versions of Linux you might get a
 # statically linked executable.
 # When not defined, configure will try to use -O2 for gcc and -O for others.
-#CFLAGS = -g
+CFLAGS = -g
```
habamax: is this necessary?

glepnir: Please only comment for bugs. These are just the changes I made and they are not in gitignore. I noticed this and I will fix it after everything is completed.

src/testdir/test_ins_complete.vim
```diff
@@ -2586,9 +2586,74 @@ func Test_complete_fuzzy_match()
...
+  call cursor(3, 1)
+  call feedkeys("Sy\<C-X>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('y', getline('.'))
+  set wrapscan
+  call feedkeys("Sy\<C-X>\<C-N>\<Esc>0", 'tx!')
+  call assert_equal('xyz', getline('.'))
+
+  " fuzzy on file
+  write fobar
+  write foobar
```
chrisbra:
```suggestion
  call writefile([''], 'foobar', 'D')
```

Same for the file  `fobar`. And then you can get rid of the `call delete()` call at the end of the function

glepnir: Okay then some other test case like this write can be clean

habamax:
Sometimes it hangs vim (having `completeopt=menu,fuzzy` and doing `<C-n>`) but I can't get any reproducible steps. And by "hangs" I mean, vim is completely unresponsive and the only way is to kill vim.

Here I tried to complete `ri`:
image

image

glepnir:
I have reproduced. wait for a while.

roccomao:
Thank you very much for your work. Regarding this patch, here are some of my superficial thoughts and issues:

1. As [habamax](https://github.com/habamax) said:

   > Currently it also affects how candidates are collected. At least for keywords.

   If there are `xyz` and `xzy`, when the `completeopt` option contains `fuzzy`, no matter I enter `y` or `z` and then press `<C-n>`, `xyz` and `xzy` will be collected as candidates. At least for me, I don't want this. Can you add an option as discussed by #14976 , such as `fuzzycollect=keyword`? (Currently `fuzzycollect` has been deleted, maybe more options can be added later, such as `fuzzycollect=keyword,tag`, etc.) Refer to the above https://github.com/vim/vim/pull/15193#issuecomment-2225148548, Can we just use the `fuzzy` option for fuzzy filtering as the documentation says, and then add an another option for users to choose which content also supports candidate fuzzy collection? I don't know how difficult it is to implement. @glepnir @chrisbra What do you think?

2. Just according to the current implementation logic of this patch, If the option also includes `longest`, the results are unexpected. Reproduce as follows:

https://github.com/user-attachments/assets/c005166f-f0d1-4478-825f-71d20d43952b

3. Based on the `longest` option mentioned above, there is another problem. I am not sure whether it is really caused by the `complete` option, at least on the surface. Reproduce as follows:

https://github.com/user-attachments/assets/399de653-4b98-41a8-b081-f0b393f95875

Summary of the reproduction of Video 3 :

```
hello helio hero test_h another_h

1. set completeopt+=fuzzy

After `fuzzy`, `test_h` and `another_h` is added now.

2. set completeopt+=longest

  After adding longest:
    1. <test_h> and <another_h> disappeared
    2. <h> was completed to the longest match <he> by default
  which is what I expected.

*** So far, everything seems to be ok. ***

3. set complete-=i

When I add `set complete-=i`, now only `hello` is matched,
which brings us back to the issue in the video 2.
```

glepnir:
If you have any questions, please raise an issue as it is not tracked here

roccomao:
OK, since the problem is caused by the patch of this PR, then I have commented here. Now I have split it into three issues, please refer to #15294, #15295, #15296.

src/search.c
```diff
@@ -5139,6 +5140,169 @@ fuzzy_match_str_with_pos(char_u *str UNUSED, char_u *pat UNUSED)
...
+    pos_T	*pos,
+    char_u	*pattern,
+    int		dir,
+    pos_T	*start_pos,
+    int		*len,
+    char_u	**ptr,
+    int		whole_line)
+{
+    pos_T	current_pos = *pos;
+    pos_T	circly_end;
```
zeertzjq, line 5218: "circly" isn't a word.

glepnir: already notice. will fix in follow up pr.

PR #15198, opened 2024-07-09
Implement buffer skipping in :bnext/:bprev

### LINKED ISSUE #4478: :bnext doesn't behave properly (and hasn't for many versions, probably)
**Description**
According to the documentation for :bnext (at [vimhelp.org](https://vimhelp.org/windows.txt.html#:bnext)), if you have multiple help buffers, you **should** be able to use :bnext to move *exclusively* to the next help buffer, and skip over any normal buffers (and probably other kinds of buffers).

*This doesn't happen*

Instead, you're sent to the next normal buffer (or perhaps another type of buffer --haven't checked), if available. If there are no normal buffers and only help buffers, the error ```E85 : There is no listed buffer``` is thrown.

**Expected behavior**
:bn (or :bnext) should move the user from the current help buffer to the next available help buffer, and skip over any normal buffers.

**Environment:**
 - Vim versions tested: 7.3, 7.4, and 8.1
 - OS: macOS 10.7
 - Terminal: Terminal.app, MacVim.app

<sub>Apparently almost no one has multiple help buffers open but me...</sub>

LemonBoy:
As explained in the help section for :bnext and :bprev the commands should jump from help buffers to help buffers (and from regular ones to regular ones).

Closes #4478

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -1358,9 +1358,8 @@ do_buffer_ext(
 	    // Don't count unlisted buffers.
 	    // Avoid non-help buffers if the starting point was a non-help buffer and
 	    // vice-versa.
-	    if (unload || (buf->b_p_bl &&
-			((flags & DOBUF_SKIPHELP) == 0 || (bp != buf &&
-			    buf->b_help == bp->b_help))))
+	    if (unload || (buf->b_p_bl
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

PR #15199, opened 2024-07-10
:bwipe doesn't remove file from jumplist of other tabpages

zeertzjq:
Problem:  :bwipe doesn't remove file from jumplist of other tabpages.
          Time complexity of mark_forget_file() is O(n^2) when removing
          all entries.
Solution: Use FOR_ALL_TAB_WINDOWS().  Start the loops over the arrays
          from the end instead of the start.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -750,13 +750,14 @@ aucmd_abort:
      */
     if (wipe_buf)
     {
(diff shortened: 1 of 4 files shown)


chrisbra:
thanks!

Issue #15200, opened 2024-07-10
Value of stopdir has no effect in 'tags'


jessepav:
### Steps to reproduce

1. `touch tags`

2. Create a nested directory: `mkdir -p foo/bar/baz`

3. `cd foo/bar/baz`

4. ```
   vim -Nu NONE -S <(cat <<EOF
   set tags=./tags;..
   EOF
   )
   ```

5. In Vim, `:echo tagfiles()`

### Expected behaviour

`tagfiles()` should not report the created `tags` file, because it's too far up the directory hierarchy. But Vim seems to disregard the value of the stopdir after the semicolon and always search up to the root.

### Version of Vim

9.1.0554

### Environment

Ubuntu 22.04

### Logs and stack traces

_No response_

zeertzjq:
The following patch will make stopdir with an absolute path with for 'tags':
```diff
diff --git a/src/tag.c b/src/tag.c
index 57f9fe016..2b7b30317 100644
--- a/src/tag.c
+++ b/src/tag.c
@@ -3409,6 +3409,8 @@ get_tagfname(
 	    (void)copy_option_part(&tnp->tn_np, buf, MAXPATHL - 1, " ,");
 
 	    r_ptr = vim_findfile_stopdir(buf);
+	    if (r_ptr != NULL)
+		r_ptr = vim_strsave(r_ptr);
 	    // move the filename one char forward and truncate the
 	    // filepath with a NUL
 	    filename = gettail(buf);
@@ -3420,6 +3422,7 @@ get_tagfname(
 		    FALSE,	   // don't free visited list
 		    FINDFILE_FILE, // we search for a file
 		    tnp->tn_search_ctx, TRUE, curbuf->b_ffname);
+	    vim_free(r_ptr);
 	    if (tnp->tn_search_ctx != NULL)
 		tnp->tn_did_filefind_init = TRUE;
 	}
```
With this patch stopdir works when using `:let &tags = './tags;' .. fnamemodify('./..', ':p')`.

zeertzjq:
This may be a better solution:
```diff
diff --git a/src/tag.c b/src/tag.c
index 57f9fe016..283343aae 100644
--- a/src/tag.c
+++ b/src/tag.c
@@ -3412,6 +3412,11 @@ get_tagfname(
 	    // move the filename one char forward and truncate the
 	    // filepath with a NUL
 	    filename = gettail(buf);
+	    if (r_ptr != NULL)
+	    {
[...]

PR #15202, opened 2024-07-10
Stop directory doesn't work properly in 'tags'

zeertzjq:
Problem:  Stop directory doesn't work properly in 'tags'.
Solution: Also move the stop directory forward by one byte.

related: #15200


jessepav:
Thanks for the quick turnaround! There's one issue with this patch:

The stopdir works properly only if it ends with a trailing slash `/` (which `fnamemodify()` happens to include).  If the stopdir doesn't end with a slash, searching will continue one directory too far up the tree.

chrisbra:
Ah, it seems there is yet another bug. I think we need this one on top of the path:
```diff
diff --git a/src/findfile.c b/src/findfile.c
index 263660925..59ab37c69 100644
--- a/src/findfile.c
+++ b/src/findfile.c
@@ -1078,11 +1078,13 @@ vim_findfile(void *search_ctx_arg)
                && search_ctx->ffsc_stopdirs_v != NULL && !got_int)
        {
            ff_stack_T  *sptr;
+           // path_end may point to the NUL or the previous path separator
+           int plen = (path_end - search_ctx->ffsc_start_dir +
+                   (*path_end != NUL));

            // is the last starting directory in the stop list?
            if (ff_path_in_stoplist(search_ctx->ffsc_start_dir,
-                      (int)(path_end - search_ctx->ffsc_start_dir),
-                      search_ctx->ffsc_stopdirs_v) == TRUE)
+                      plen, search_ctx->ffsc_stopdirs_v) == TRUE)
                break;

            // cut of last dir
```

zeertzjq:
That doesn't seems to be the correct solution, and `ff_path_in_stoplist()` already does that:
```c
    // eat up trailing path separators, except the first
    while (path_len > 1 && vim_ispathsep(path[path_len - 1]))
	path_len--;
```

zeertzjq:
I added a fix for the problem with trailing slash in another commit.

chrisbra:
> That doesn't seem to be the correct solution, and ff_path_in_stoplist() already does that:

That is correct, but it doesn't work correctly, because the length calculation in `vim_findfile()` is wrong. On the first iteration, path_end points to the NUL, so `path_end - ffsc_start_dir` gives the correct length of the directory, so that in `ff_path_in_stoplist()` `path[path_len-1]` correctly points to the trailing slash.

However in subsequent iterations,  `path_end` points to the last (, second last, etc) path delimiter of `ffsc_start_dir` and as such the length calculation `path_end - ffsc_start_dir` is off by one, so that one would need to check `path[path_len]` if this points to the directory delimiter

chrisbra:
> This doesn't support relative stop directories yet, as they are not
supported in other places like findfile() either.

BTW: can't we simply convert the relative path to an absolute path then?

zeertzjq:
> > That doesn't seem to be the correct solution, and ff_path_in_stoplist() already does that:
> 
> That is correct, but it doesn't work correctly, because the length calculation in `vim_findfile()` is wrong. On the first iteration, path_end points to the NUL, so `path_end - ffsc_start_dir` gives the correct length of the directory, so that in `ff_path_in_stoplist()` `path[path_len-1]` correctly points to the trailing slash.
> 
> However in subsequent iterations, `path_end` points to the last (, second last, etc) path delimiter of `ffsc_start_dir` and as such the length calculation `path_end - ffsc_start_dir` is off by one, so that one would need to check `path[path_len]` if this points to the directory delimiter

I don't think the length is the problem here. In `ff_path_in_stoplist()`, when `path_len` is no larger than the length of the stopdir, it uses `fnamecmp()` to compare, which may actually compare beyond `path_len` until the `NUL`.

chrisbra:
But in any case, removing the path separator doesn't work properly, because the length is wrong.

zeertzjq:
Oh, I see. The value of `path_len` isn't too large. It's too small.

zeertzjq:
I added another commit to expand relative paths. However, only `tag;../` works, `tag;..` doesn't.

chrisbra:
thanks!

PR #15205, opened 2024-07-10
Expand dir path in 'cdpath'

### LINKED ISSUE #374: Completion for ":cd" from 'cdpath'
```
The `cd` command should consider the `cdpath` setting for completion of 
relative paths.

Currently the completion appears to use only the current working directory when 
completing relative paths.

This means that `:cd foo<Tab>` should complete `foobar` if there exists 
`/path/in/cdpath/foobar` (and `/path/in/cdpath` is in `cdpath`).
```

Original issue reported on code.google.com by `dhahler@gmail.com` on 15 Jun 2015 at 6:55

LemonBoy:
Closes #374

### DIFF
--- a/src/usercmd.c
+++ b/src/usercmd.c
@@ -102,6 +102,7 @@ static struct
     {EXPAND_BREAKPOINT, "breakpoint"},
     {EXPAND_SCRIPTNAMES, "scriptnames"},
 #endif
+    {EXPAND_DIRS_IN_CDPATH, "dir_in_path"},
     {0, NULL}
 };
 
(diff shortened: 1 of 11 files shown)


src/usercmd.c
```diff
@@ -102,6 +102,7 @@ static struct
     {EXPAND_BREAKPOINT, "breakpoint"},
     {EXPAND_SCRIPTNAMES, "scriptnames"},
 #endif
+    {EXPAND_DIRS_IN_CDPATH, "dir_in_path"},
```
zeertzjq, line 105: This is not documented in `:h getcompletion()` or `:h :command-complete`

LemonBoy: Thanks, I had completely missed that.

gdupras:
This entry in `:h todo.txt` can be removed:

https://github.com/vim/vim/blob/50dc83cf9215aa787da54abbb0bd2ab4fb89e720/runtime/doc/todo.txt#L542

Issue #10835, opened 2022-08-02
prolog filetype detection


ignamartinoli:
### Steps to reproduce

1. `vim example.pro`
2. `:set ft?`

### Expected behaviour

The file to be recognized as a **Prolog** file

### Version of Vim

latest

### Environment

Operating system: Gentoo
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: zsh

chrisbra:
Hm,
currently, this is detect as IDL:

https://github.com/vim/vim/blob/e175dc6911948bcd0c854876b534fee62fb95b9f/runtime/filetype.vim#L887-L888

Is there a way to determine prolog files from the content?

brammool:
> ### Steps to reproduce
> 
> 1. `vim example.pro`
> 2. `:set ft?`
> 
> ### Expected behaviour
> 
> The file to be recognized as a **Prolog** file

Does Prolog uses the .pro file name extension?  I see .pl and .pdb.
Unfortunately .pl is also used for Perl.  Oh, it appears GNU Prolog
suggests using .pro.  But it is also used for idlang and C/C++ header
files.  Inspecting the file contents is required.

We can copy the check for Prolog from the .pl rule to also use for .pro
files:

  if getline(2) =~ '.;$'
    setf cpp
  else
    # recognize Prolog by specific text in the first non-empty line
    # require a blank after the '%' because Perl uses "%list" and "%translate"
    var l = getline(nextnonblank(1))
    if l =~ '\<prolog\>' || l =~ '^\s*\(%\+\(\s\|$\)\|/\*\)' || l =~ ':-'
      setf prolog
    else
      exe 'setf ' .. default
    endif
  endif


-- 
Get a life?  What is the URL where it can be downloaded?

 /// Bram Moolenaar -- ***@***.*** -- http://www.Moolenaar.net   \\\
///                                                                      \\\
\\\        sponsor Vim, vote for features -- http://www.Vim.org/sponsor/ ///
 \\\            help me help AIDS victims -- http://ICCF-Holland.org    ///

ignamartinoli:
Normally a prolog file can commence with a """function""" like syntax with a dot at the end as

```prolog
something(somethigs).
```

An edge case is that the dot can be on another line for intentation as

```pro
something(too_much_things,
  to_be_on,
  the_same_line).
```

benknoble:
[...]

PR #15206, opened 2024-07-10
Closes: https://github.com/vim/vim/issues/10835

ignamartinoli:
Improved the Prolog file detection regex and added tests for all cases.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -465,7 +465,7 @@ export def ProtoCheck(default: string)
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var lnum = getline(nextnonblank(1))
-    if lnum =~ '\<prolog\>' || lnum =~ '^\s*\(%\+\(\s\|$\)\|/\*\)' || lnum =~ ':-'
+    if lnum =~ '\<prolog\>' || lnum =~ '(^\s*(:-\|%\|\/\*))\|.\s*$'
       setf prolog
     else
       exe 'setf ' .. default
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -2576,4 +2576,31 @@ func Test_uci_file()
   filetype off
 endfunc
 
+func Test_pro_file()
+  filetype on
+
+  "Prolog
+  call writefile([':-module(test/1,'], 'Xfile.pro', 'D')
+  split Xfile.pro
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  call writefile(['% comment'], 'Xfile.pro', 'D')
+  split Xfile.pro
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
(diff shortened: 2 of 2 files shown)


chrisbra:
thanks!

runtime/autoload/dist/ft.vim
```diff
@@ -465,7 +465,7 @@ export def ProtoCheck(default: string)
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var lnum = getline(nextnonblank(1))
-    if lnum =~ '\<prolog\>' || lnum =~ '^\s*\(%\+\(\s\|$\)\|/\*\)' || lnum =~ ':-'
+    if lnum =~ '\<prolog\>' || lnum =~ '(^\s*(:-\|%\|\/\*))\|.\s*$'
```
zeertzjq, line 468: This pattern doesn't make sense. It will match any non-empty line, because `.\s*` can match any character at the end of a line.

ignamartinoli: I left it in case there were trailing whitespaces, just in case. Is that a bad idea?

zeertzjq: It will match any non-empty line:
```vim
:echo 'foobar' =~ '(^\s*(:-\|%\|\/\*))\|.\s*$'
```
prints 1.

zeertzjq: If you want to match a `.` at the end of the line. You should use `\.\s*` instead. But that doesn't pass the tests, which means the pattern is still wrong :(

ignamartinoli: Yea, you're right. I will work on fixing it right away.

ignamartinoli: Writing this:

```viml
let regex = '^\(:-\|%\|\/\*\)\|\.$'

let pattern1 = ':-module(test/1,'
let pattern2 = '% comment'
let pattern3 = '/* multiline comment'
let pattern4 = 'rule(test, 1.7).'
let patternX = 'a'

if pattern1 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern2 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern3 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern4 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if patternX =~ regex
  echo 'ERROR'
else
  echo 'OK'
endif
```

All tests are passing now. I forgot that in Vim regexes grouping is with `\(` and not just `(`. I'm pushing the fix now. Thanks @zeertzjq for the feedback!

ignamartinoli: @chrisbra how can I find out which patch number I should use?

k-takata: @ignamartinoli He will add a patch number when he commit a patch to the master branch. You don't need to add a patch number by yourself.

ignamartinoli: @k-takata thanks for the info. I think I was able to fix it now

PR #15207, opened 2024-07-10
Improve performance of `id()` builtin by over 50%.

errael:
I was concerned about using `printf`, forgot to check.Sure enough...

This also has the probably incredibly minor improvement that unequal strings generally fail faster since the string is reversed. Interestingly, tried `&& v != 0` in loop termination, may have been a nanosecond or so slower.

@zeertzjq @zzzyxwvut @chrisbra who says I'm obsessive about performance?

Here's some data.
**PR**
  0.069  0.068  0.069  0.069  0.068  0.068 : ret = type(blob1)  ###-k
  0.080  0.081  0.081  0.081  0.080  0.080 : ret = id(blob1)  ###-m
  0.094  0.092  0.095  0.093  0.095  0.094 : ret = typename(blob1)  ###-o
  0.128  0.119  0.127  0.120  0.122  0.120 : ret = getpid()  ###-l

**current**
  0.064  0.065  0.063  0.065  0.064  0.064 : ret = type(blob1)  ###-k
  0.184  0.194  0.191  0.192  0.188  0.185 : ret = id(blob1)  ###-m
  0.097  0.098  0.097  0.098  0.098  0.098 : ret = typename(blob1)  ###-o
  0.129  0.125  0.135  0.126  0.130  0.124 : ret = getpid()  ###-l


src/evalfunc.c
```diff
@@ -7540,17 +7541,18 @@ f_id(typval_T *argvars, typval_T *rettv)
...
 	    // Assume pointer value in typval_T vval union at common location.
 	    if (argvars[0].vval.v_object != NULL)
-		vim_snprintf((char*)numbuf, sizeof(numbuf), "%p",
-					    (void *)argvars[0].vval.v_object);
-	    else
-		numbuf[0] = NUL;
-	    break;
+	    {
+		vimlong_T v = (vimlong_T)(void *)argvars[0].vval.v_object;
+		for(; p < numbuf + sizeof(void *) * 2; ++p, v >>= 4)
```
zzzyxwvut: Substitute `<< 1` for `* 2` (and parenthesise)?

errael: Except for `p` they're all constants; the compiler figures it out.

(using `<<` would require parens)

zzzyxwvut:
So there will be no wasteful "0x" prefix to boot, right?

errael:
> So there will be no wasteful "0x" prefix to boot, right?

Right, another immeasurable improvement ;-)

LemonBoy:
IMO the readability is much worse, the gains negligible and the operation pretty uncommon.

> Here's some data.

No units, no script to reproduce the results.

errael:
> IMO the readability is much worse, the gains negligible

I think this is the first time I've heard a well over 50% performance gain as negligible.

It looks to me that `vim` goes to great pains, for better or worse, to be efficient.

> and the operation pretty uncommon.

That remains to be seen.

> 
> > Here's some data.
> 
> No units, no script to reproduce the results.

Oops, I left out the heading. See https://github.com/errael/vim-microprof for details/scripts for the data collection process and output generation which is based on the vim profiler. These particular results are generated from `examples/basic_prof.vim`.

The heading
```
=== basic micro-prof, 5 run/params ===
     10     30     65    100    200    300 : nKeys (usec/op)
```
So it's microseconds per operation. `nKeys` can be ignored for operations that don't involve dictionaries.

In this situation, performance goes from 180 nanoseconds per `id()` call to 80 nanoseconds.

LemonBoy:
> I think this is the first time I've heard a well over 50% performance gain as negligible.

50% of zero is still zero ;)

chrisbra:
Personally, I am with @LemonBoy  here and I prefer the current readable approach, but I see you added some comments. so I can as well just include it.

zzzyxwvut:
I guess, it'll be nice to have `id()` recognised as a built-in  
function for syntax highlighting as well, @errael:

```shell
cd runtime/syntax/generator
make clean all
git add ../vim.vim
git commit
```

errael:
> I guess, it'll be nice to have `id()` recognised as a built-in function for syntax highlighting as well, @errael:
> 
> ```shell
> cd runtime/syntax/generator
> make clean all
> git add ../vim.vim
> git commit
> ```

I've never done anything with syntax highlighting. Are those steps to do and and some file to this PR? (I'd replace the git commands with the VCS I use ;-) )

And sending a message to @lacygoill, for https://github.com/lacygoill/vim9-syntax (which I use).

errael:
> I guess, it'll be nice to have `id()` recognised as a built-in function for syntax highlighting 

Done. Yet another thing I didn't want to know about and I'm glad it exists.

zzzyxwvut:
Well, it could have been included in the PR that introduced  
`id()`.  I don't know if it merits a separate future PR or  
if we should wait for it to be automatically included the  
next time `vim.vim` is updated.

chrisbra:
I try to think about updating vim syntax script when including new functions, but often simply forget.

errael:
> Well, it could have been included in the PR that introduced `id()`. I don't know if it merits a separate future PR or if we should wait for it to be automatically included the next time `vim.vim` is updated.

It's in this PR now.

zzzyxwvut:
There is a decimal [`tltoa()`](https://github.com/vim/vim/blob/8252ef134f5a57e8a07b66ce79068c262a6386d8/src/term.c#L2624).

errael:
> There is a decimal [`tltoa()`](https://github.com/vim/vim/blob/8252ef134f5a57e8a07b66ce79068c262a6386d8/src/term.c#L2624).

Wow. That example has both a division and modulo in the loop; that might be closer to printf performance. 

I was comfortable doing this loop, because it's all bit operations and absolutely minimalist.

I wonder about basic `Intro to "C"` constructs being too difficult to understand, and it's better to use _much more expensive_ library functions. One of `vim`'s most appreciated features is it's performance. Are things changing?

PR #15209, opened 2024-07-11
Tests for findfile() & finddir() save and set 'shellslash' unnecessarily

zeertzjq:
Problem:  Tests for findfile() and finddir() save and set 'shellslash'
          unnecessarily, as it's already set by runtest.vim.
Solution: Remove these lines.

### DIFF
--- a/src/testdir/test_findfile.vim
+++ b/src/testdir/test_findfile.vim
@@ -30,9 +30,7 @@ endfunc
 " Test findfile({name} [, {path} [, {count}]])
 func Test_findfile()
   let save_path = &path
-  let save_shellslash = &shellslash
   let save_dir = getcwd()
-  set shellslash
   call CreateFiles()
   cd Xfinddir1
   e Xdir2/foo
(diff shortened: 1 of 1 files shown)


zeertzjq:
Actually the main inconsistency here is that `Test_finddir()` saves and restores the value of 'shellslash' without setting it, so an alternative is to actually set it in `Test_finddir()`:
```diff
diff --git a/src/testdir/test_findfile.vim b/src/testdir/test_findfile.vim
index 20d509613..311020e26 100644
--- a/src/testdir/test_findfile.vim
+++ b/src/testdir/test_findfile.vim
@@ -133,6 +133,7 @@ func Test_finddir()
   let save_shellslash = &shellslash
   let save_dir = getcwd()
   set path=,,
+  set shellslash
   call CreateFiles()
   cd Xfinddir1
 
```

chrisbra:
I'll rather keep the tests explicit, so let me include your alternative approach. Thank you Sir!

chrisbra:
included as of e7b98ab96e1f1bd12032c620615a2c69adbf018d

PR #15215, opened 2024-07-11
Keep track of container types in assignments

### LINKED ISSUE #15208: [vim9script] `dict` corruption using `items()` return value
### Steps to reproduce

```vim
vim9script

var d: dict<list<number>> = {a: [1], b: [2]}
var item = d->items()[0]
item[1][0] = 'abc'
echo typename(d)
echo d
```
outputs
```
dict<list<number>>
{'a': ['abc'], 'b': [2]}
```
Observe one of the values is a `list<string>`

### Expected behaviour

Should be an exception when executing `item[1][0] = 'abc'`.

### Version of Vim

9.1.554

### Environment

ubuntu/gtk3

### Logs and stack traces

_No response_

LemonBoy:
Introduce runtime checks to ensure any assignment to an existing item of a dictionary or list takes into account the member type.

Closes #15208

The `vim9execute.c` part can be reviewed by comparing the new functions with `list_append_tv` and `dict_add_tv` that are used in the else branch. The `eval.c` is sligtly more involved but is essentially the same but for the expression evaluator.

### DIFF
--- a/src/dict.c
+++ b/src/dict.c
@@ -1636,6 +1636,24 @@ dict_set_items_ro(dict_T *di)
 }
 
 /*
+ * Replaces the typval_T of dict item "di" to "tv".  "tv" is copied.
+ * Return FAIL when the type is wrong.
+ */
+    int
+dict_set_item_move(dict_T *d, dictitem_T *di, typval_T *tv)
+{
+    if (d->dv_type != NULL
+	    && d->dv_type->tt_member->tt_type != VAR_ANY
+	    && check_typval_arg_type(d->dv_type->tt_member, tv, NULL, 0) == FAIL)
+	return FAIL;
+
+    clear_tv(&di->di_tv);
+    di->di_tv = *tv;
+
+    return OK;
+}
+
+/*
  * "has_key()" function
  */
     void
--- a/src/eval.c
+++ b/src/eval.c
@@ -2313,9 +2313,9 @@ set_var_lval(
 		typ = lp->ll_valtype;
 	    // Use the type infos available at runtime.
 	    else if (lp->ll_dict != NULL)
-		typ = lp->ll_dict->dv_type;
+		typ = lp->ll_dict->dv_type->tt_member;
 	    else if (lp->ll_list != NULL)
-		typ = lp->ll_list->lv_type;
+		typ = lp->ll_list->lv_type->tt_member;
 
 	    if (typ != NULL && typ->tt_type != VAR_ANY
 		    && check_typval_arg_type(typ, rettv, NULL, 0) == FAIL)
--- a/src/eval.c
+++ b/src/eval.c
@@ -2312,9 +2312,9 @@ set_var_lval(
 	    if (lp->ll_valtype != NULL)
 		typ = lp->ll_valtype;
 	    // Use the type infos available at runtime.
-	    else if (lp->ll_dict != NULL)
+	    else if (lp->ll_dict != NULL && lp->ll_dict->dv_type != NULL)
 		typ = lp->ll_dict->dv_type->tt_member;
-	    else if (lp->ll_list != NULL)
+	    else if (lp->ll_list != NULL && lp->ll_list->lv_type != NULL)
 		typ = lp->ll_list->lv_type->tt_member;
 
 	    if (typ != NULL && typ->tt_type != VAR_ANY
--- a/src/eval.c
+++ b/src/eval.c
@@ -2304,10 +2304,23 @@ set_var_lval(
 	    return;
 	}
 
-	if (lp->ll_valtype != NULL
-		    && check_typval_arg_type(lp->ll_valtype, rettv,
-							      NULL, 0) == FAIL)
-	    return;
+	if (in_vim9script())
+	{
+	    type_T	*typ = NULL;
+
+	    // Use the declared type if available.
+	    if (lp->ll_valtype != NULL)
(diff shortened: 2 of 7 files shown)


errael:
I've been playing around with the patch.  Haven't been able to slip anything past it. Tried with both `list` and `dict`.

I took a quick look at the patch; don't claim to fully understand it. But here's a few things that caught my attention.

In eval.c there's `if (typ != NULL && typ->tt_type != VAR_ANY`, why the `VAR_ANY` check? Isn't
this handled by `check_typval_arg_type`.

About the names of the new methods, `*_set_item_move`.
Why `_move`? Should it be `_copy` or `_replace` or `_reference` or `_assign`?

I ran a performance test,  `xxx[1] = 3` is around 25% slower. STOREINDEX is in the ballpark of twice as slow. Correct behavior is most important, but I wonder if the `*_set_item_move` could be optimized. I attached the profiling script.

This is based on
```vim
var ln: list<number> = [1, 2, 3]
var la: list<any> = [ 'a', ln]
def ListAssign()
    var xxx: any = la[1]
    xxx[1] = 3                      ###-F
enddef
```
The statement in question disassembles as
```
    xxx[1] = 3
  14 PUSHNR 3
  15 PUSHNR 1
  16 LOAD $0
  17 STOREINDEX any
  ```

Don't have instuction level profiling (briefly talked to Bram about it, but never got there). Assuming that each of the four instructions takes the same amount of time without this patch (which is probably not valid, STOREINDEX is probably the slowest, but gives a feel), we can compute that STOREINDEX takes around twice as long.

#### NO CHECK
```
=== list assign check micro-prof, 5 run/params ===
    x50   x100   x200 : argXnLoop (usec/op)
  0.054  0.053  0.053 : xxx[1] = 3  ###-F
```

#### WITH CHECK
```
=== list assign check micro-prof, 5 run/params ===
    x50   x100   x200 : argXnLoop (usec/op)
  0.073  0.071  0.070 : xxx[1] = 3  ###-F
```
[list-assign-check.vim.txt](https://github.com/user-attachments/files/16199603/list-assign-check.vim.txt)

**Comment by errael** [comment]
Something I didn't think to examine is the following case. The compiler can determine that no type checking is needed. So no runtime check is needed and it shouldn't suffer performance degradation.

In this case there's `STOREINDEX list`, in the other there's `STOREINDEX any`. I wonder if the `list` vs `any` is useful.

```
vim9script

def F()
    var l: list<number> = [1, 2, 3]
    l[1] = 3
enddef
F()
```

**Comment by errael** [comment]
Wow, I didn't realize how simple it was to reproduce. Without this patch there's
```
vim9script

var ln: list<number> = [1, 2, 3]
def OUCH()
    var xxx: any = ln
    xxx[1] = 'foo'
enddef
OUCH()
echo 'OUCH' ln
```
with output showing a corrupted list
```
OUCH [1, 'foo', 3]
```

**Comment by errael** [comment]
> I didn't think to examine is the following case. The compiler can determine that no type checking is needed

This case
```
var ln: list<number> = [1, 2, 3]
def F()
    var xxx: list<number> = ln
    xxx[1] = 3
enddef
```
where the compiler knows that the types are OK also has the performance degradation.

I'm starting to think of this as a compiler, not runtime, bug.

Shouldn't/couldn't the compiler add a `CHECKTYPE` instruction if the target of the assignment
**might** have typeinfo that hasn't been checked?

@yegappan Any thoughts on compiler vs runtime and how to detect the situations vulnerable to corruption?

**Comment by errael** [comment]
> Shouldn't/couldn't the compiler add a `CHECKTYPE` instruction if the target of the assignment
> **might** have typeinfo that hasn't been checked?

I guess at runtime if the target is `any`, it might have been assigned something with a type, and so should be checked. I still think it's a compiler issue, but this might be a runtime workaround that avoids the corruption but doesn't degrade performance when the compiler has done type checking.

**Comment by errael** [comment]
At the entry to `execute_storeindex(isn_T *iptr, ectx_T *ectx)` there is `SOURCING_LNUM = iptr->isn_lnum;`.

The assignment appears two more times after this first time. It seems like they are not needed.

**Comment by errael** [comment]
@chrisbra @yegappan @LemonBoy 

I think the problem is a compiler bug; some missing `CHECKTYPE` instructions.
If there's disagreement on that, or a compiler fix is not forthcoming, I
haven't seen any problems with this PR. Here's some ways to look at things.

1. Wait for a compiler fix.
2. Go with the runtime check permanently.
   Compiler can probably remove some `CHECKTYPE` instructions.
   This slows down assignments fully checked by the compiler.
3. Apply this PR for now to avoid data corruption. Leave bug open,
   or open new bug for a compiler fix.
4. Let the compiler add a `CHECKTYPE` with every STOREINDEX. This still
   is a performance hit by checking when not needed, but it doesn't
   pollute the runtime. Could be fixed later. Might be easier to work
   with going forward than (3).

No matter how you look at it, the data corruption issue should be fixed.

```
vim9script
var ln: list<number> = [1, 2, 3]
def F()
    var l: any = ln
    l[1] = 'foo'
enddef
F()
echo typename(ln)
echo ln
```

```
list<number>
[1, 'foo', 3]
```

chrisbra:
You are right, it smells like a compiler bug. I am still slightly worried about the added performance overhead, which seems to happen at runtime. 

@LemonBoy what is the status of this patch, since it is marked as draft?

LemonBoy:
Been AFK for the past weeks, I'll get back to this ASAP.

PR #15218, opened 2024-07-11
fnamemodify() treats ".." and "../" differently

zeertzjq:
Problem:  fnamemodify() treats ".." and "../" differently.
Solution: Expand ".." properly like how "/.." is treated in 8.2.3388.


chrisbra:
thanks

PR #15221, opened 2024-07-12
Revise test_gettext_make

k-takata:
* Remove unnecessary comment.
* Check the gettext feature.
* Check whether %GETTEXT_PATH% is set.
* Check the return code of the make command.
* Use heredoc.


chrisbra:
thanks!

PR #15223, opened 2024-07-12
Mark Test_gui_lowlevel_keyevent as flaky

k-takata:
Test_gui_lowlevel_keyevent is failing on vim-win32-installer recently.
Mark it as flaky for now.

(I tried to reproduce this in my local environment, but I couldn't. So, I'm not sure what the root cause is.)


chrisbra:
thanks!

PR #15224, opened 2024-07-12
Work around Test_gui_lowlevel_keyevent (2nd try)

k-takata:
I found that v9.1.0571 was not enough.
Even with the 5 times retry, the test still often fails in some environments.

Ctrl-C interruption may occur before the preceding events are processed.
Exclude Ctrl-C to avoid the flakiness.

### DIFF
--- a/src/testdir/test_gui.vim
+++ b/src/testdir/test_gui.vim
@@ -1706,10 +1706,10 @@ endfunc
 func Test_gui_lowlevel_keyevent()
   CheckMSWindows
   new
-  let g:test_is_flaky = 1
 
   " Test for <Ctrl-A> to <Ctrl-Z> keys
-  for kc in range(65, 90)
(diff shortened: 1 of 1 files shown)


chrisbra:
Hm, I remember having fixed that some time ago already. Ah here: f6ebaa7ae64edbe2e56d65c2a3f2494301d11acb

Do we for some reason no longer get the correct Interrupt? Or is that caused by the timing problem you mentioned? Can you then please also revert the whole `try/endtry` part from here? https://github.com/vim/vim/commit/f6ebaa7ae64edbe2e56d65c2a3f2494301d11acb#diff-0ef14f691c02c5166c87d226ec2b3aa94ba03ae04f7843c6960ec574aad5ea06 

I guess we do no longer need it then.

k-takata:
Ctrl-C still generates the interrupt, but sometimes(?) it causes the timing issue. So I think `try` blocks are still needed.

chrisbra:
thanks!

Issue #15098, opened 2024-06-24
Termdebug: use conpty with Windows. 


ubaldot:
**Please describe the problem.**

At the moment it looks like that termdebug for Windows use only the `prompt` alternative.  However, newer Windows version include conpty that could enable the utilization of termdebug in `terminal` mode. 

**Describe the solution you'd like**

It would be nice to use termdebug in `terminal` mode on Windows. 

**Describe alternatives you've considered**
I have tried to run the following script:

```
vim9script

var commbufnr = term_start('NONE', {
    term_name: 'test',
    hidden: true
    })

var commpty = term_gettty(commbufnr)
if has('win32')
  commpty = substitute(commpty, '\\', '\\\\', 'g')
endif

echo commpty
execute $'!gdb -q -ex "new-ui mi {commpty}"'
```

but I get a `open terminal failed: Permission denied`. 

I think the problem (and I may be wrong) is that the pty is created by Vim and then I try to use it with gdb. 
TBH I don't know if that is the problem, nor how to solve it.
Any help would be appreciated.

LemonBoy:
gdb assumes the pty/pipe descriptor is bidirectional, as you can see [here](https://github.com/bminor/binutils-gdb/blob/218bb9dcdbb8ee5a15977dc1ffe7ed7422803368/gdb/ui.c#L183). Vim, on the other hand, opens two simplex pipes for the in/out channels, hence the failure you're seeing.

Changing the [two](https://github.com/vim/vim/blob/f0837ba0b9dba9d30e3dc2626d4f690a2d51103e/src/terminal.c#L7644) named pipes' mode to be duplex makes gdb happy, but seemingly breaks other things (I get a hang in `ClosePseudoConsole` when executing commands with `:!`).

cc @k-takata

ubaldot:
@LemonBoy I must say that I am not an expert of pty:s in general and such low level stuff, yet your explanation has been very clear. Many thanks for that. And also thanks for attempting to fix the problem! 

Do you think it would be possible to trace the effect of the change you proposed and eventually accommodate the broken parts?

LemonBoy:
I'm not an expert in ptys either, I've uploaded some code as proof of concept [here](https://github.com/LemonBoy/vim/commit/fed84233b085a8386eb467673d212deab24af3eb) to show that the gdb issue is fixed and that there's something wrong somewhere.

[...]

PR #15226, opened 2024-07-12
Make Variables window handling more flexible.

ubaldot:
The current implementation limits the way user can layout Termdebug-variables-listing window on the screen after it has been created. 

For example, by doing something like the following:

```
:Var
var currwin = win_getid()
var var_bufnr = bufnr()
split
exe $':buffer {var_bufnr}'
win_execute(currwin, 'close')
```

the variable window will not be updated because the ID of the window created by `:Var` has gone. 

The proposal here is to refresh the buffer containing Variables information, no matter in which window such a buffer is going to be displayed, without wiping it out.

Given that it is difficult to write automatic tests for assessing the robustness of this approach, I would really appreciate if someone could test the changes manually.  

**Note:** I tried to implement the same change for `:Asm` but that seems more involved. In-fact by performing the same change the `:Asm` buffer/windows management won't work. Any help is appreciated.


errael:
I know that a while ago I gave up trying to adjust the window layout, including `:Var, so I'm hopeful that I can do that one of these days.

> the variable window will not be updated because the ID of the window created by `:Var` has gone.

I don't know if this is relevant, but you can get the winid with
```
vim9cmd echo bufwinid('Termdebug-variables-listing')
```

> The proposal here is to refresh the buffer containing Variables information, no matter in which window such a buffer is going to be displayed, without wiping it out.

I didn't try to understand the patch in full, but I noticed the
```
setlocal bufhidden=hide
```
With this patch. If I delete/close the :Var window, will it continue to be updated while it's hidden? If so, that seems like a problem.

> Given that it is difficult to write automatic tests for assessing the robustness of this approach, I would really appreciate if someone could test the changes manually.

What's the problem doing automatic tests?

ubaldot:
> I know that a while ago I gave up trying to adjust the window layout, including `:Var, so I'm hopeful that I can do that one of these days.
> 

Good to know that I am not alone :) 

> > the variable window will not be updated because the ID of the window created by `:Var` has gone.
> 
> I don't know if this is relevant, but you can get the winid with
> 
> ```
> vim9cmd echo bufwinid('Termdebug-variables-listing')
> ```
> 

Assume that you moved the `Termdebug-variable-listing` to another window with another id.  To let the variable values to update in such a new window, you need to assign the new window id to  `varwin` variable.  However, this is not possible. The only way to reassign `varwin` is by closing the variables window and run  `:Var` again but then the newly created window will be at a fixed position that you cannot freely move after that has been created. 

The current way of handling the variables window is to slam the variables values in a scratch buffer that is wiped out when the window containing it is closed. It is a nice and very efficient idea but it limits the way you can move the variable window around. When you create a variable window will appear... there. 

> 
> I didn't try to understand the patch in full, but I noticed the
> 
> ```
> setlocal bufhidden=hide
> ```
>

Exactly. The idea is to not wipe out the variable buffer at window closing but to hide it. The buffer is updated when needed. By doing this, you can open a variable window in whatever position you want and slam that buffer in it. 

In reality I changed few lines, but given that I fixed the indentation it seems a huge change. 

 
> With this patch. If I delete/close the :Var window, will it continue to be updated while it's hidden? If so, that seems like a problem.

Why that could be a problem?

> 
> > Given that it is difficult to write automatic tests for assessing the robustness of this approach, I would really appreciate if someone could test the changes manually.
> 
> What's the problem doing automatic tests?

Because you should create a layout with different windows and check for example that every time you hit a breakpoint the variables values in that window change. Oh well, perhaps it is not too difficult.... 

All in all, attempting to make Termdebug flexible enough to let the user to arbitrarily chose a layout may be  nightmare and I think that every solution in that regard would make someone  unhappy. Hence, my thinking was to leave termdebug to give an initial, fixed layout, and then let users to shuffle the windows around in "post" in the way they want. All should be happy in this way. But with the current implementation this trick is not possible.

ubaldot:
BTW: there is something that I haven't checked: what if a `Termdebug-variables-listing` buffer already exists? What happens if I run `:Var` and then I close the window and run `:Var` again? How many buffer will I have? To be checked...

errael:
> Assume that you moved the `Termdebug-variable-listing` to another window with another id. To let the variable values to update in such a new window, you need to assign the new window id to `varwin` variable. However, this is not possible. The only way to reassign `varwin` is by closing the variables window and run `:Var` again but then the newly created window will be at a fixed position that you cannot freely move around.

Some things that come to mind; (1) seems simpler.
1. Get rid of `varwin` variable in favor of a function like
```
def GetVarWin(): number
    return bufwinid('Termdebug-variables-listing')
enddef
```
2. Use autocommands to keep `varwin` in sync as things shift around

> The current way of handling the variables window is to slam the variables values in a scratch buffer that is updated until it is visible and it is wiped out when the window containing it is closed. It is a nice and very efficient idea but it limits the way you can move the variable window around. When you create a variable window will appear... there.

So this means the variables buffer is around and active all the time?

> > I didn't try to understand the patch in full, but I noticed the
> > ```
> > setlocal bufhidden=hide
> > ```
> 
> Exactly. The idea is to not wipe out the variable buffer at window closing but to hide it. The buffer is updated when needed even if it is hidden. By doing this, you can open a variable window in whatever position you want and slam that buffer in it.
> 
> In reality I changed few lines, but given that I fixed the indentation it seems a huge change.

Got it. (should have tried ignore whitespace)

> > With this patch. If I delete/close the :Var window, will it continue to be updated while it's hidden? If so, that seems like a problem.
> 
> Yes, exactly, the buffer that was associated to that window will continue to be updated and reused. Why that could be a problem?

This question is based on my ignorance of the gdb/mi protocol and how variables window/buffer is handled. I assumed that the variables weren't transmitted unless requested; thinking about a low performance connection. But, IIUC, they are always transmitted and handled, so not an issue.

> All in all, attempting to make Termdebug flexible enough to let the user to arbitrarily chose a layout may be nightmare and I think that every solution in that regard would make someone unhappy. Hence, my thinking was to leave termdebug to give an initial, fixed layout, and then let users to shuffle the windows around in "post" in the way they want.

Sounds like a good approach.

> All should be happy in this way. But with the current implementation this trick is fairly limited. But of course, there is no free lunch. :)

I don't understand why, but that's the advantage of having no idea what's going on inside ;)

ubaldot:
> Some things that come to mind; (1) seems simpler.
> 
> 1. Get rid of `varwin` variable in favor of a function like
> 
> ```
> def GetVarWin(): number
>     return bufwinid('Termdebug-variables-listing')
> enddef
> ```
> 

Hum, I am failing to understand how this could be used? Could you elaborate a bit more? What happens when you have more windows showing `'Termdebug-variables-listing'` buffer? How to  use it when you have `'Termdebug-variables-listing'` in window A and you want to move its content to some window B that has been created with some split command? 

> 2. Use autocommands to keep `varwin` in sync as things shift around

This could be interesting but it is very scary!  

> 
> This question is based on my ignorance of the gdb/mi protocol and how variables window/buffer is handled. I assumed that the variables weren't transmitted unless requested; thinking about a low performance connection. But, IIUC, they are always transmitted and handled, so not an issue.
> 

In-fact the variables are transmitted from gdb to the pty 'gdb-communication' when they are requested from Vim to gdb  via the same pty. That happens through the command  `SendCommand('-stack-list-variables 2')`  which is called for example when a breakpoint is hit or when `:Var` is called.  When a new message from gdb to Vim via the pty happens the message is handled through the `CommOutput` callback function.

> > All in all, attempting to make Termdebug flexible enough to let the user to arbitrarily chose a layout may be nightmare and I think that every solution in that regard would make someone unhappy. Hence, my thinking was to leave termdebug to give an initial, fixed layout, and then let users to shuffle the windows around in "post" in the way they want.

> 
> 
> I don't understand why, but that's the advantage of having no idea what's going on inside ;)

Ahahaha! I think termdebug has a nice architecture, perhaps some function look very long and the parsing of the messages received is very elaborated. I think that not even Bram was very satisfied with the current communication, but it works so far. I would refactoring it a bit but unfortunately my life is going to again be very busy soon and I have to carefully select how to use my time. :( 

Still on the communication in a more general perspective: the prompt way is very ugly but apparently that is the only solution nowadays with Windows. I tried to use the new `conpty` but it does not work. I really hope that someone could look into it, see https://github.com/vim/vim/issues/15098#issuecomment-2190005953. 

BTW: the "no-name buffer leakage" seems to be solved with the latest commit.

errael:
> > Some things that come to mind; (1) seems simpler.
> > 
> > 1. Get rid of `varwin` variable in favor of a function like
> > 
> > ```
> > def GetVarWin(): number
> >     return bufwinid('Termdebug-variables-listing')
> > enddef
> > ```
> 
> Hum, I am failing to understand how this could be used?

The OP says
> > > the variable window will not be updated because the ID of the window created by :Var has gone.

and in the next post

> > Assume that you moved the Termdebug-variable-listing to another window with another id. To let the variable values to update in such a new window, you need to assign the new window id to varwin variable. However, this is not possible. The only way to reassign varwin is by closing the variables window and run :Var again but then the newly created window will be at a fixed position that you cannot freely move around.

> Could you elaborate a bit more?

The point is that the `varwin` variable is not needed as a global. Inside a `:def` you can do
```
var varwin = GetVarWin()
```

> What happens when you have more windows showing `'Termdebug-variables-listing'` buffer?

Are you talking about the same buffer in multiple windows or multiple buffers with similar names?

If the former, doesn't it work to update/change the buffer and all the windows show the changes?

> How to use it when you have `'Termdebug-variables-listing'` in window A and you want to move its content to some window B that has been created with some split command?

Use it? By "it" do you mean the windid where the `:Vars` buffer is? You don't really need it, if I understand. If I'm not missing something, you can do something like
```
var original_varwin = GetVarWin()
win_gotoid(split_target_windowB)
:buffer Termdebug-variables-listing
# Now you have 'Termdebug-variables-listing' in two windows.
win_gotoid(original_varwin)
# Do whatever is needed to get original_varwin to display some new content.
```
This might be cumbersome. There's also a variety of commands to move and/or exchange windows.
```
# exchange the vars window with the window after it
win_gotoid(GetVarWin())
:wincmd x
```
After any of the above, `GetVarWin()` is the new winid of the vars window, which is the originally stated goal.

But `vim` has so many ways to do so many things...

> In-fact the variables are transmitted from gdb to the pty 'gdb-communication' when they are requested from Vim to gdb via the same pty. That happens through the command `SendCommand('-stack-list-variables 2')` which is called for example when a breakpoint is hit or when `:Var` is called.

So when the vars window is hidden, it's straightforward to avoid this overhead.

errael:
> We should be ready to go.

It looks like the idea is to manipulate the buffer; don't care about the window. Is it OK to remove the global `varswin` variable?

With this patch I can move around the vars window? Close it and then re-open it later? No overhead while it's closed?

ubaldot:
> 
> It looks like the idea is to manipulate the buffer; don't care about the window.

Exactly. 

>  Is it OK to remove the global `varswin` variable?
> 

No. Because `:Var` either create a new window with the variable listing in it OR it move the cursor to the variable listing window if the variable listing is already displayed in a window which tbh is a handy feature (note the function name ´GotoVariableswinOrCreateIt()`).  

> With this patch I can move around the vars window? Close it and then re-open it later? 
 
 Exactly! I would appreciate if you give it a go and provide feedback. :) 

> No overhead while it's closed?

If you refer to the overhead due to running `SendCommand('-stack-list-variables 2')` when the `Termdebug-variables-listing` is hidden, then that is an easy fix: you send that command only when the buffer is visible in a window (cfr line 1722 of the latest commit's file):

```
  if running && !empty(win_findbuf(varbufnr))
    SendCommand('-stack-list-variables 2')
  endif
enddef
```

I can test it quickly and commit the above change.

ubaldot:
> Are you talking about the same buffer in multiple windows or multiple buffers with similar names?
> 
> If the former, doesn't it work to update/change the buffer and all the windows show the changes?
> 

This is exactly what I am doing. But to do that we need keep the buffer somewhere, hence I changed the `wipe` to `hide`. 

> Use it? By "it" do you mean the windid where the `:Vars` buffer is? You don't really need it, if I understand. If I'm not missing something, you can do something like
> 
> ```
> var original_varwin = GetVarWin()
> win_gotoid(split_target_windowB)
> :buffer Termdebug-variables-listing
> # Now you have 'Termdebug-variables-listing' in two windows.
> win_gotoid(original_varwin)
> # Do whatever is needed to get original_varwin to display some new content.
> ```
> 

But then `GetVarWin()` shall be available to the user. Correct?  

> This might be cumbersome. There's also a variety of commands to move and/or exchange windows.
> 
> ```
> # exchange the vars window with the window after it
> win_gotoid(GetVarWin())
> :wincmd x
> ```
> 
> After any of the above, `GetVarWin()` is the new winid of the vars window, which is the originally stated goal.

Ok, it looks like that  `GetVarWin()` shall be available to the user. If so, certainly this solution would work well. 

> So when the vars window is hidden, it's straightforward to avoid this overhead.

Yeah, I don't think the overhead is large, but if we can remove it easily why not? (See my latest commit).

errael:
> > Is it OK to remove the global `varswin` variable?
> 
> No. Because `:Var` either create a new window with the variable listing in it OR it move the cursor to the variable listing window if the variable listing is already displayed in a window which tbh is a handy feature (note the function name ´GotoVariableswinOrCreateIt()`).

Isn't it a bug to use `varswin` after the vars window has been moved? Seems correct to test if the buffer exists rather than some stale variable. Does it use `varswin` to "move the cursor to the variable listing window", that is clearly a problem if it is in a different window.

ubaldot:
> 
> Isn't it a bug to use `varswin` after the vars window has been moved? Seems correct to test if the buffer exists rather than some stale variable. Does it use `varswin` to "move the cursor to the variable listing window", that is clearly a problem if it is in a different window.

In-fact it is a bug. 
To be 100% correct we should replace `if !win_gotoid(varwin) ...` with ` if !empty(win_findbuf(varbufnr)) ... `. 
A new commit is on its way. Many thanks for the feedback!

errael:
> > Are you talking about the same buffer in multiple windows or multiple buffers with similar names?
> > If the former, doesn't it work to update/change the buffer and all the windows show the changes?
> 
> Yes, but that is not always the case. Try this: run `:Var` to open a variable window, split it and close the original window opened with `:Var` .

This is not the former "same buffer multiple windows" case. It's the later case that's the problem, which is part of this PR's fix.

> > Use it? By "it" do you mean the windid where the `:Vars` buffer is? You don't really need it, if I understand. If I'm not missing something, you can do something like
> > ```
> > var original_varwin = GetVarWin()
> > win_gotoid(split_target_windowB)
> > :buffer Termdebug-variables-listing
> > # Now you have 'Termdebug-variables-listing' in two windows.
> > win_gotoid(original_varwin)
> > # Do whatever is needed to get original_varwin to display some new content.
> > ```
> 
> But then `GetVarWin()` shall be available to the user. Correct?

The user can do
```
bufwinid('Termdebug-variables-listing')
```
Is that what you mean?

> Also, even if you get the id, when the variable window has gone, so is the buffer therein contained. What buffer shall we put in a newly created window that we want to use to monitor the variables?

I'm not sure I understand the question. One thing is that you can not save `varswin`, it changes.
Every time you want to use it, `GetVarWin()` must be used.

> 
> > This might be cumbersome. There's also a variety of commands to move and/or exchange windows.
> > ```
> > # exchange the vars window with the window after it
> > win_gotoid(GetVarWin())
> > :wincmd x
> > ```
> >   
> > After any of the above, `GetVarWin()` is the new winid of the vars window, which is the originally stated goal.
> 
> Ok, it looks like that `GetVarWin()` shall be available to the user. But still, I don't think it will work.

What is it that wouldn't work?

BTW, it looks like `win_findbuf(varbufnr)->empty()` is equivelent to `bufwinid(varbufnr) < 0`. The
first, with it's list creation, is probably just a bit more expensive.

errael:
Just to mention, there's also `bufwinnr(varbufnr) < 0` :)

ubaldot:
I just commit a change, where I used `win_findbuf`. I am going to replace it with `bufwinnr(varbufnr) < 0´ ;)

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1715,7 +1715,7 @@ def GotoVariableswinOrCreateIt()
       exe $'resize {GetVariablesWindowHeight()}'
     endif
   else
-    win_gotoid(varwins[0])
+    win_gotoid(varwin)
   endif
 
   if running && !empty(win_findbuf(varbufnr))
```
errael:
```suggestion
  if running && bufwinid(varbufnr) > 0
```

I didn't look for other uses of `win_findbuf`, but this one was just sitting there visible when I looked at the patch.

Ah there's both varbufnr and varbufname. Most of the functions will take either.

ubaldot: Done! I used `bufname `as per `:h` :)

errael: You might take a look at `:help bufname`, that string is file-pattern plus some extras and there's some weirdness. Using a number as the arg is unambiguous; but I doubt it matters here.

ubaldot: Oh ok! I never noticed that. Good to know!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1718,7 +1718,7 @@ def GotoVariableswinOrCreateIt()
     win_gotoid(varwin)
   endif
 
-  if running && bufwinid(varbufname) > 0
+  if running && varwin > 0
```
errael: If I understand the logic here, when this is executed, there is **always** a `:Vars` window.

And the `varwin` variable is stale if the window was created.

ubaldot: True. Once you get there a` :Var` window exists. We don't need the `&&` condition.

ubaldot:
I am late to the party, but I discovered such a `win_splitmove()` function only... now. 
By using such a function, maybe users can arbitrarily move the Var window around with the current termdebug version (not 100% though).

At this point, I think we should make a decision: either considering the buffer-based approach that I am proposing with this PR, or to leave everything as is and ditch this PR.

It follows pros and cons that I can see with any of the two approaches.

**buffer-based approach (this PR).**
**pros**: 
1. It should be easier from a user side to move windows around because I assume folks use more frequently the paradigm   "split and load buffer in the split window" than `win_splitmove()`. This PR would prevent user to remember that in this specific use-case they cannot use the paradigm "split and load buffer in the split window" - which is the most obvious. They must use `win_splitmove()`. To me, as a user, that would be yet another oddity of vim. This PR would fix this. 
2. Perhaps it is more robust and maintainable to have an unlisted, hidden buffer that contains Var and that you can place it whatever you want when you want. 

**cons**:
1. We have the implementation for `:Var` but not `:Asm` (which is window-based), hence we would not have consistency. I tried to use the mechanism for `:Asm` but it didn't work out of the box. `:Asm` requires more work and I am not sure if I have the skills/time, but maybe someone could look into it.  

**window-based approach (current termdebug version).**

**pros and cons** are reverted. 

@errael @chrisbra What do you think?

Shane-XB-Qian:
too long to read what you've chatted in above comments,

but per the code change here for now, donot understand what was the 'flexible' (as ticket title) here,
and donot understand why you changed 'new' -> 'split';
and you knew Var window mostly was for current var value of debugged source code, right?

so maybe suggested you leave it alone.

-- 
shane.xb.qian

errael:
In the OP
> Note: I tried to implement the same change for :Asm but that seems more involved. In-fact by performing the same change the :Asm buffer/windows management won't work.

Does your experience over the last few days change this?

errael:
> considering the buffer-based approach that I am proposing with this PR, or to leave everything as is and ditch this PR

In my mind, it is about manipulating `termdebug` information. Can the user define their own layouts and how? I would like to be able manipulate the layout using regular `vim` commands and not be concerned about the implementation (there might be restrictions, but best to keep them to a minimum). 

Some requirements I'd like to see
- close the data window, bring it back up later
- move windows around
- turn a window into a popup
- put info in a specified window
- ...

Considering this, and more, as the requirements
- What's a good implementation
- What needs to be documented (like the names of particular buffers, for manipulation)
- What support commands should be provided (like `PopupWin`/`PopupBuf`)
- Can I create an empty window and issue a command to fill it with particular info (like Asm or Var)
-  Are there any special considerations the user needs to be aware of

I'm not suggesting all this needs to be done right now. But the internal design should support this stuff and/or move in that direction.

errael:
> **buffer-based approach (this PR).** **pros**:
> 
> 1. It should be easier from a user side to move windows around because I
> assume folks use more frequently the paradigm   "split and load buffer in the
> split window" than `win_splitmove()` - at least that apply to me. This PR
> would prevent user to remember that in this specific use-case they cannot use
> the paradigm "split and load buffer in the split window". They must use
> `win_splitmove()`. I am not sure how obvious that would be. This PR would fix
> this.

I don't understand this. Why doesn't "split and load buffer in the split window" work? Maybe my confusion is that it's not clear to me (after reading this) what works with or without this PR. Why would some stuff not work?

> 2. Perhaps it is easier to maintain having an unlisted, hidden buffer that
> contains Var and that you can place it whatever you want when you want.

Is using `bufhidden` needed anymore? With some of the new code in this PR, since `varwin` is no more, it seems like it doesn't matter if the buffer goes away or not. I'm not saying it's a bad idea to do `&bufhidden = "hide"`, I don't know all the implications; just wondering.

ubaldot:
>>Note: I tried to implement the same change for :Asm but that seems more involved. In-fact by performing the same change the >>:Asm buffer/windows management won't work.

>Does your experience over the last few days change this?

Nope. That would require more investigation. `:Var` is OK, `:Asm` is NOK. 

 >Some requirements I'd like to see

> - lose the data window, bring it back up later
> - move windows around

At the moment this is possible with and without this PR. I wasn't aware of the `win_splitmove()` function that allows you to move a window while keeping its ID. 

>- turn a window into a popup

This is  super-cool idea. But it was long ago since I used any popups. 

> - put info in a specified window

This can be certainly done with this PR in a fairly easy and intuitive way. That is, you open a new window and you place the hidden variable buffer in it. 
Without this PR you could but the process is a bit more involved and you must use `win_splitmove()`

> Why doesn't "split and load buffer in the split window" work? 

This work with this PR but not without because of the way windows ID are handled and how the `GotoOrCreateVariablesWin()` is created. I can try to make you an example.

You run `:Var` and it creates a variable window with win_id = [1058] in a predefined position. You hit breakpoints, step, etc. The buffer contained in [1058] is updated. Perfect. 

Now say that you want to have the same buffer contained in window with id = [1058] in another window. You create a split and you get a window with win_id = [1059].  You display the buffer located in [1058] in [1059] and then you close [1058] because you don't need it anymore. You continue debugging but you won't see any change inside window [1059]. 

Why? Because this is the way termdebug works: it is designed to update **only**  [1058]. If [1058] is closed, then such an ID has gone along with the buffer therein contained. If you had a split window with the variable buffer in it you won't see updates any longer. 

What to do then? You must run again `:Var` again. A new variable window is open in the predefined position with ID = [1060]. Until such a window exists, termdebug updates the buffer therein contained. 

I hope it is a bit clear now. :) 

Regarding this PR  I think you got the idea. Instead of wiping away the variable buffer the idea is to keep it hidden. In this way you don't care about any win_id. If you want to display the variables listing just open any window and unhide the hidden variable buffer in it.

errael:
Been playing around with the patch manually from the command line. Nice.
The main things
1) Split the `:Var` window, `CTRL-W_s`, observe both windows update at breakpoint.
2) Create a new window and load it with the `var` info.
    - Closed the `:Var` window 
    - Moved the cursor to where I wanted the `var` window
    - `:below 7new`
    - `:buf 5`

Some questions/comments

1. Why is `buflisted` set to false?
    - Makes it harder to find the window to manipulate the layout.
    - The name of the buffer probably should be documented in the help.
2. Seems like modifiable should be false, turned on only when changing.
    The way it is now, it's easy to edit, that doesn't feel right. The user
    might be surprised if their edits suddenly disappear.
3. Add an autocommand specifically for the var buffer.
  The help: "when a hidden buffer is displayed in a window".
4. Clean up a leftover bug (the `call`) with `ch_logfile()`.

```diff
$ diff -U2 ../runtime/pack/dist/opt/termdebug/plugin/ /junk/install/vim/share/vim/vim91/pack/dist/opt/termdebug/plugin/termdebug.vim
--- ../runtime/pack/dist/opt/termdebug/plugin/termdebug.vim	2024-07-14 15:59:40.998254336 -0700
+++ /junk/install/vim/share/vim/vim91/pack/dist/opt/termdebug/plugin/termdebug.vim	2024-07-14 18:34:52.210954772 -0700
@@ -346,6 +346,6 @@
   endif
 
-  # Uncomment this line to write logging in "debuglog".
-  # call ch_logfile('debuglog', 'w')
+  # Uncomment this line to write logging in "termdebug.log".
+  # ch_logfile('termdebug.log', 'w')
 
   # Assume current window is the source code window
@@ -732,4 +732,5 @@
     au BufUnload * BufUnloaded()
     au OptionSet background Highlight(0, v:option_old, v:option_new)
+    execute $'au BufWinEnter {varbufname} BufWinEnterVar()'
   augroup END
 
@@ -1703,5 +1704,5 @@
       setbufvar(varbufname, "&bufhidden", 'hide')
       setbufvar(varbufname, "&buftype", 'nofile')
-      setbufvar(varbufname, "&buflisted", false)
+      #setbufvar(varbufname, "&buflisted", false)
       setbufvar(varbufname, "&signcolumn", 'no')
       setbufvar(varbufname, "&modifiable", true)
@@ -1718,4 +1719,10 @@
     SendCommand('-stack-list-variables 2')
   endif
+enddef
+
+def BufWinEnterVar()
+  if running && bufwinnr(varbufname) != -1
+    SendCommand('-stack-list-variables 2')
+  endif
 enddef
```

ubaldot:
> Been playing around with the patch manually from the command line. Nice.

Cool! :) 

> Split the :Var window, CTRL-W_s, observe both windows update at breakpoint.

That is normal because you are showing the same buffer in two windows. 

>Create a new window and load it with the var info.
>
>- Closed the :Var window
>- Moved the cursor to where I wanted the var window
>- :below 7new
>- :buf 5

The last line should be `:buf Termdebug-variables-listing`. Note that `Termdebug-variables-listing` is hard-coded in termdebug. 

> Why is buflisted set to false?

Because it is a sort of read-only buffer used internally by termdebug. Having it "on your way" in the buffer list that you are directly interacting with would be weird; listed buffers are supposed to be edited by the user and it would be weird to have a sort of "intruder" there. That is, setting "Termdebug-var-listing" unlisted would just make the UI more tidy. 

IMO `gdb-communication` and `debugged-program` shall also be unlisted. 

If user really needs to access these buffers, he can access them through `ls u` or by referring them by their names, which are hard-coded in termdebug. But yes, that shall be documented in the help, as you said. 

>Seems like modifiable should be false, turned on only when changing.
>The way it is now, it's easy to edit, that doesn't feel right. The user
>might be surprised if their edits suddenly disappear.

I thought about that, but always forgot to implement! :D 

> Add an autocommand specifically for the var buffer.
> The help: "when a hidden buffer is displayed in a window".

I would be a bit hesitant to add autocommands tbh. Isn't it OK now with the `GotoOrCreateVarWindow()` function? 

Caveat: with this PR adding support for popup windows should be quite easy. Without this PR I am not sure it is possible because the popup window should get the ID of the `Var` window and I don't think this could be possible - we would need a `win_popupmove()` analogous to `win_splitmove()`.

errael:
Forgot to mention a bug I saw. Related to `:Var` window, but only loosely related to this PR. Notice the typo in the following when using old style config `disasm`->`variables`
```vim
def GetVariablesWindow(): number ### <<<<<<<<<<<<< WRONG TYPE
  if exists('g:termdebug_config')
    return get(g:termdebug_config, 'variables_window', 0)
  endif
  if exists('g:termdebug_disasm_window') ###<<<<<<<<<<<<<<< TYPO
    return g:termdebug_variables_window
  endif
  return 0
enddef
```
The return type should be `bool`. Without that return type the user can get an exception `Using a number as a bool` in places that use `GetVariablesWindow`, such as `StartDebug_internal`. At least with the `: bool` the function name that shows up on the stack might help the user track down the problem without needing to look at the `termdebug` source.

ubaldot:
> Hmm, isn't the whole idea of this PR so the user can directly interact with it.

Bad wordings from my side: yes, the purpose is to interact with it, what I mean was that user can actively type into it. 

>  If there were an option to turn off modifiable in the source window, I'd use it.

I thought about that as well soooo many times! :)  

> I completely agree with that sentiment; I avoid them like the plague. But if
> you want normal vim commands to work in specialized environments, like
> termdebug, I've found that there's not much choice.

Note that in `HandleCursor()` you have: 

```
  if running && stopped && bufwinnr(varbufname) != -1
    SendCommand('-stack-list-variables 2')
  endif
```

perhaps it is enough to remove that `&& bufwinnr(varbufname) != -1`  and we don't need the autocommand? I am just elaborating because as you said:

> I completely agree with that sentiment; I avoid them like the plague.

>>The only stopper I see for this PR is the Asm buffer/window management. If
>>it is OK with you I could commit the same mechanism for the Asm
>>buffer/window as for the Var and point out where I think there is a
>>problem so you could take an eye on it, if you want. :)

>That sounds good. With both handled and flexible, it's more like a UI spec.

Tomorrow ;-)

errael:
> > I completely agree with that sentiment; I avoid them like the plague. But if
> > you want normal vim commands to work in specialized environments, like
> > termdebug, I've found that there's not much choice.
> 
> Note that in `HandleCursor()` you have:
> 
> ```
>   if running && stopped && bufwinnr(varbufname) != -1
>     SendCommand('-stack-list-variables 2')
>   endif
> ```
> 
> perhaps it is enough to remove that `&& bufwinnr(varbufname) != -1` and we don't need the autocommand? User typically take actions when the debugger is stopped somewhere.

Adding unneeded overhead, to a possibly slow communication channel, seems like
the wrong choice.

chrisbra:
I am sorry, I lost track a bit. What is the status of this one please? Did we reach an agreement here?

ubaldot:
More work is needed here, but I am on vacation right now, I'll keep working on it once back.

ubaldot:
@chrisbra Hello Christian, it looks like that I won't be able to go back to work on this anytime soon. I am going through a fairly hectic period that also involves a job change among other things, so my free time is reduced at minimum at the moment. 

I don't know what is the best to do, with this PR if close it, set it as Draft or just keep it opened. Everything is fine with me :)

chrisbra:
Sure no worries. Let's close it then for now. We can always revisit later. Thanks for your contributions.

PR #15227, opened 2024-07-12
filetype: Debian devscripts configuration files are not recognized

Freed-Wu:
Problem:  Debian devscripts configuration files are not recognized
Solution: recognize devscripts.conf and .devscripts files as sh filetype
          (sourced by /bin/sh)


chrisbra:
@jamessan do you approve?

jamessan:
These files are a small subset of sh syntax -- just comments and variable assignment. Using sh filetype will work, even if it's too liberal in what it accepts.

chrisbra:
> These files are a small subset of sh syntax -- just comments and variable assignment. Using sh filetype will work, even if it's too liberal in what it accepts.

I think we don't have anything other than the "shell" filetype for such a restricted "shell dialect". I leave that decision to include this to you @jamessan Whatever you prefer, I will commit :)

jamessan (approved):

chrisbra:
thanks!

PR #15229, opened 2024-07-12
ex: fix inserting text after trailing bar

mohd-akram:
This is the behavior in [traditional ex](https://github.com/n-t-roff/heirloom-ex-vi) and as specified by [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ex.html) (step 12.c):

> If the command was an **append**, **change**, or **insert** command, and the step 12.b. ended at a \<vertical-line> character, any subsequent characters, up to the next non- \<backslash>-escaped \<newline> shall be used as input text to the command.

### DIFF
--- a/runtime/doc/insert.txt
+++ b/runtime/doc/insert.txt
@@ -1992,8 +1992,8 @@ too long when appending characters a line break is automatically inserted.
 			name.
 
 These two commands will keep on asking for lines, until you type a line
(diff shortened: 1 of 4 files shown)


chrisbra:
Thanks. Can you add some documentation please?

mohd-akram:
Updated the documentation.

chrisbra:
thanks!

PR #15230, opened 2024-07-12
ex: fix implicit print for single address

mohd-akram:
This makes it more consistent rather than having eg. `1,1` behave differently than `1,2`, or `.` doing nothing. This is also how it works in [traditional ex](https://github.com/n-t-roff/heirloom-ex-vi), and there's no mention of differing addresses in the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ex.html) (step 6.b) for an implied print to take place.

### DIFF
--- a/runtime/doc/motion.txt
+++ b/runtime/doc/motion.txt
@@ -351,6 +351,7 @@ gg			Goto line [count], default first line, on the first
 
 							*:[range]*
 :[range]		Set the cursor on the last line number in [range].
+			In Ex mode, print the lines in [range].
 			[range] can also be just one line number, e.g., ":1"
 			or ":'m".
 			In contrast with |G| this command does not modify the
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -2730,7 +2730,7 @@ ex_range_without_command(exarg_T *eap)
 {
     char *errormsg = NULL;
 
-    if ((*eap->cmd == '|' || (exmode_active && eap->line1 != eap->line2))
+    if ((*eap->cmd == '|' || exmode_active)
 #ifdef FEAT_EVAL
 	    && !in_vim9script()
 #endif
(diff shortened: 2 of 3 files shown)


chrisbra:
Can you please give an example, how this is different currently? Also, I don't understand why you changed the text to do an `:s` command. I don't think this should be done

mohd-akram:
If you don't specify a command to ex, but only an address or address range, it's supposed to print the contents of the address range. Eg:

```
:.
```

is the same as

```
:.p
```

Likewise,

```
:1,2
```

is the same as

```
:1,2p
```

Currently, Vim's ex only does this implicit print when an address range is given where the first number is different than the second, so `:1,1` doesn't print that line, but `:1,2` does. Likewise, `:.` doesn't print because a single address is the same as an address range where `line1 == line2`.

> Also, I don't understand why you changed the text to do an :s command. I don't think this should be done.

Because Vim's ex ignored single addresses, that line was a no-op despite the address being invalid. As a consequence of this change, the address is now checked in order to print so that command is seen as invalid since the buffer is empty. I just changed it to a valid command, but it can be any other valid command. AFAICT the test doesn't rely on those specific contents, just the length based on the comment above it.

chrisbra:
> Because Vim's ex ignored single addresses, that line was a no-op despite the address being invalid. As a consequence of this change, the address is now checked in order to print so that command is seen as invalid since the buffer is empty. I just changed it to a valid command, but it can be any other valid command. AFAICT the test doesn't rely on those specific contents, just the length based on the comment above it.

Ah, so this is just there, because otherwise we couldn't test that the pasting actually succeeds. I see thanks then.

Can you please document the new behaviour?

mohd-akram:
I couldn't find the implicit print mentioned anywhere else so I added it to the `:[range]` doc.

chrisbra:
thanks. I also added a test for that behaviour.

mohd-akram:
Ah, that's how you test it :D Thanks!

PR #15231, opened 2024-07-12
Add v:event.typedchar for KeyInputPre autocmd

Shougo:
Continues of #15182 

I have added `v:event.typedchar` for KeyInputPre autocmd.
Because `v:char` is mapped keys.

It is useful to record user input keys.

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -42,6 +42,9 @@ static buffheader_T recordbuff = {{NULL, {NUL}}, NULL, 0, 0};
 
 static int typeahead_char = 0;		// typeahead char that's not flushed
 
+static char_u typedchars[MAXMAPLEN + 1] = { NUL };  // typed chars before map
+static int typedchars_pos = 0;
+
 /*
  * When block_redo is TRUE the redo buffer will not be changed.
  * Used by edit() to repeat insertions.
@@ -1371,7 +1374,15 @@ gotchars(char_u *chars, int len)
 
 	// Handle one byte at a time; no translation to be done.
 	for (i = 0; i < state.buflen; ++i)
+        {
 	    updatescript(state.buf[i]);
+
+	    if (typedchars_pos < MAXMAPLEN)
+	    {
+		typedchars[typedchars_pos] = state.buf[i];
+		typedchars_pos++;
+	    }
+        }
 
 	if (reg_recording != 0)
 	{
@@ -2137,6 +2148,9 @@ vgetc(void)
     c = do_key_input_pre(c);
 #endif
 
+    // Clear the next typedchars_pos
+    typedchars_pos = 0;
+
     // Need to process the character before we know it's safe to do something
     // else.
     if (c != K_IGNORE)
@@ -2175,6 +2189,8 @@ do_key_input_pre(int c)
     else
 	buf[(*mb_char2bytes)(c, buf)] = NUL;
 
+    typedchars[typedchars_pos] = NUL;
+
     get_mode(curr_mode);
 
     // Lock the text to avoid weird things from happening.
@@ -2183,6 +2199,7 @@ do_key_input_pre(int c)
 
     v_event = get_v_event(&save_v_event);
     (void)dict_add_bool(v_event, "typed", KeyTyped);
+    (void)dict_add_string(v_event, "typedchar", typedchars);
 
     if (apply_autocmds(EVENT_KEYINPUTPRE, curr_mode, curr_mode, FALSE, curbuf)
 	&& STRCMP(buf, get_vim_var_str(VV_CHAR)) != 0)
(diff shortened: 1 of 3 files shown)


Shane-XB-Qian:
maybe also add an option to user who can globally forbit such event,
 and only can switch that option at cli manually which cannot be set in plugins,
otherwise this was some kind of like a backdoor to evil.

--
shane.xb.qian

4513ECHO:
You can use `'eventignore'` option.

Shane-XB-Qian:
an option to user who can globally forbit such event,
 and __only__ can switch that option at cli manually which cannot be set in plugins

--
shane.xb.qian

Shougo:
> otherwise this was some kind of like a backdoor to evil.

I think all plugins can be evil.
I understand your concerns, but they are not specific to this feature.

Shane-XB-Qian:
I supposed no one like his input key be potentially monitored.
// like some evils on clipboard.

--
shane.xb.qian

tani:
You say that monitoring keystrokes leads to backdoors, but when you can run scripts in the first place, arbitrary code (really arbitrary!!!) by `system()` function can be executed, so your concern seems to be that of worrying about front door security in a house with constantly released windows. Nevertheless, I think it is worthwhile to list the risks, and the comment itself deserves respect.

Shane-XB-Qian:
yes. I meant keyinputpre _open_ a backdoor potentially if no limitation.
as well as system(), but that's why we have dist#xxx chk and workspace if you trusted.
nevertheless, it is free and you do not need to thank me or pay me _or_ you can just trust the privacy paper. :shrug: ahahaaaaaa...

--
shane.xb.qian

chrisbra:
Hm, we already have `-w` do something similar. Can't you re-use this, check the `scriptout` global variable? Also I think you may want to add a few ifdefs, since this is only required with `FEAT_EVAL`.

src/getchar.c
```diff
@@ -1371,8 +1374,16 @@ gotchars(char_u *chars, int len)
...
 	// Handle one byte at a time; no translation to be done.
 	for (i = 0; i < state.buflen; ++i)
+        {
 	    updatescript(state.buf[i]);
 
+	    if (typedchars_pos < MAXMAPLEN)
+	    {
+		typedchars[typedchars_pos] = state.buf[i];
+		typedchars_pos++;
+	    }
```
zeertzjq, line 1384: You need to call `vim_unescape_csi()` on `state.buf[]`, otherwise chars like "…" will have K_SPECIAL/CSI escaped.

src/getchar.c
```diff
@@ -2175,6 +2189,8 @@ do_key_input_pre(int c)
     else
 	buf[(*mb_char2bytes)(c, buf)] = NUL;
 
+    typedchars[typedchars_pos] = NUL;
```
zeertzjq, line 2193: You need to call `vim_unescape_csi()` on `typedchars[]` before calling the autocommand (like how that's done when finishing recording a register), otherwise chars like "…" will have K_SPECIAL/CSI escaped.
```suggestion
    typedchars[typedchars_pos] = NUL;
    vim_unescape_csi(typedchars);
```

zeertzjq:
```suggestion
    typedchars[typedchars_pos] = NUL;
    vim_unescape_csi(typedchars);
```

Shougo:
> Hm, we already have -w do something similar. Can't you re-use this, check the scriptout global variable? Also I think you may want to add a few ifdefs, since this is only required with FEAT_EVAL.

`scriptout` variable seems `FILE *`.  It is type-incompatible.

Shougo:
I have added `FEAT_EVAL` ifdef.

chrisbra:
> scriptout variable seems FILE *. It is type-incompatible.

True, but I wonder if we can fill `v:event.typedchar` at the same place where scriptout is handled (in `updatescript()`)

Shougo:
OK.  I have changed the code.

chrisbra:
thanks!

Shane-XB-Qian:
@chrisbra i thought you agreed to make this evt be protected, e.g re-use the '-w' option, but actually not, still no limit to use?

chrisbra:
No, I was merely suggesting to add the new code at the same position where `-w` is handled.

I don't understand the protected reasoning. You can use `:set eventignore` if you don't like it, but of course there is no way to prevent a plugin from disabling this. But I don't see what makes this so sensitive, considering that there already is `-w` and then a malicious plugin could also run `call system('rm -rf $HOME')` which is certainly worse.

Shane-XB-Qian:
i think made 'keyinputpre' work only when vim startup with '-w' option is clear and safe.
as for system() i have said above, as for 'rm -rf xxx' looks we had a disagreement (of one netrw PR) already.

ok, just for a (my) record, but whatever your choice.

chrisbra:
So why exactly do you think this is a security issue? I am missing the attack vector here.

PR #15234, opened 2024-07-12
Add typst ftplugin runtime files

gpanders:
Adapted with some minor modifications from https://github.com/kaarmu/typst.vim.

The `typst/emoji.vim` and `typst/symbols.vim` are quite large and are used to provide conceal characters for emoji and math symbols. The `g:typst_conceal` variable is disabled by default so these files won't be used unless the user opts in to them. I can remove them if they're too bulky.

### DIFF
--- a/runtime/autoload/typst.vim
+++ b/runtime/autoload/typst.vim
@@ -1,0 +1,50 @@
+" Language:    Typst
+" Maintainer:  Gregory Anders
+" Last Change: 2024-07-14
+" Based on:    https://github.com/kaarmu/typst.vim
+
+function! typst#indentexpr() abort
+    let l:lnum = v:lnum
+    let s:sw = shiftwidth()
+
+    let [l:plnum, l:pline] = s:get_prev_nonblank(l:lnum - 1)
+    if l:plnum == 0 | return 0 | endif
+
+    let l:line = getline(l:lnum)
+    let l:ind = indent(l:plnum)
+
+    let l:synname = synIDattr(synID(l:lnum, 1, 1), 'name')
+
+    " Use last indent for block comments
+    if l:synname == 'typstCommentBlock'
+        return l:ind
+    endif
+
+    if l:pline =~ '\v[{[(]\s*$'
+        let l:ind += s:sw
+    endif
+
+    if l:line =~ '\v^\s*[}\])]'
+        let l:ind -= s:sw
+    endif
+
+    return l:ind
+endfunction
+
+" Gets the previous non-blank line that is not a comment.
+function! s:get_prev_nonblank(lnum) abort
+    let l:lnum = prevnonblank(a:lnum)
+    let l:line = getline(l:lnum)
+
+    while l:lnum > 0 && l:line =~ '^\s*//'
+        let l:lnum = prevnonblank(l:lnum - 1)
+        let l:line = getline(l:lnum)
+    endwhile
+
+    return [l:lnum, s:remove_comments(l:line)]
+endfunction
+
+" Removes comments from the given line.
+function! s:remove_comments(line) abort
+    return substitute(a:line, '\s*//.*', '', '')
+endfunction
--- a/runtime/compiler/typst.vim
+++ b/runtime/compiler/typst.vim
@@ -1,0 +1,15 @@
+" Vim compiler file
+" Language:    Typst
+" Maintainer:  Gregory Anders
+" Last Change: 2024-07-14
+" Based on:    https://github.com/kaarmu/typst.vim
+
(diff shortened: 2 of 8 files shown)


chrisbra:
@ribru17 are you aware that you will be taken away maintainershipt for the ftplugin? 

> The g:typst_conceal variable is disabled by default so these files won't be used unless the user opts in to them. I can remove them if they're too bulky.

Yeah, let's remove it please. People who want the full expierence can still install the upstream repo.

runtime/autoload/typst.vim
```diff
@@ -0,0 +1,337 @@
+function! typst#TypstWatch(...)
```
chrisbra: Can you add a header please?

runtime/compiler/typst.vim
```diff
@@ -0,0 +1,34 @@
+" Vim compiler file
+" Compiler: typst
```
chrisbra: Can you please add the usual Header lines (Maintainer, Last Change?)

runtime/compiler/typst.vim
```diff
@@ -0,0 +1,34 @@
...
+    finish
+endif
+let current_compiler = get(g:, 'typst_cmd', 'typst')
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+if exists(":CompilerSet") != 2
+    command -nargs=* CompilerSet setlocal <args>
+endif
```
chrisbra: We got rid of the fallback for the `:CompilerSet` command back in 408281e16a36c15eed10fbf0406fa8ab159fc4bf

runtime/compiler/typst.vim
```diff
@@ -0,0 +1,34 @@
...
+
+" With `--diagnostic-format` we can use the default errorformat
+let s:makeprg = [current_compiler, 'compile',
+              \  '--diagnostic-format', 'short']
+
+if has('patch-7.4.191')
+    call add(s:makeprg, '%:S')
+else
+    call add(s:makeprg, '%')
+endif
```
chrisbra: Since we are including this for Vim 9.1, I think we can remove the test for patch-7.4.191 (after all, that's been 10 years).

ribru17:
> @ribru17 are you aware that you will be taken away maintainershipt for the ftplugin?

Fine with me :+1:

runtime/compiler/typst.vim
```diff
@@ -0,0 +1,34 @@
+" Vim compiler file
+" Compiler: typst
+
+if exists('current_compiler')
+    finish
+endif
+let current_compiler = get(g:, 'typst_cmd', 'typst')
```
chrisbra, line 10: This variable should be documented

runtime/ftplugin/typst.vim
```diff
@@ -1,14 +1,43 @@
...
 setlocal commentstring=//\ %s
+setlocal comments=s1:/*,mb:*,ex:*/,://
+setlocal formatoptions+=croq
+setlocal suffixesadd=.typ
+
+if has('win32')
+    setlocal iskeyword=a-z,A-Z,48-57,_,-,128-167,224-235
+else
+    setlocal iskeyword=a-z,A-Z,48-57,_,-,192-255
+endif
```
chrisbra: Why do we need different iskeyword settings for different systems? Is that for word motions or for the syntax keyword script in which case it should probably ba a :syn iskeyword command in the syntax script.

runtime/ftplugin/typst.vim
```diff
@@ -1,14 +1,43 @@
...
+
+let b:undo_ftplugin = 'setl cms< com< fo< isk< sua<'
+
+if get(g:, 'typst_conceal', 0)
```
chrisbra, line 21: missing documentation

runtime/ftplugin/typst.vim
gpanders: The setting itself was changed in https://github.com/kaarmu/typst.vim/issues/77 but it looks like they just set it to the default value. So I'll delete these lines.

gpanders:
* Added docs for `g:typst_cmd`, `g:typst_conceal`, and `g:typst_embedded_languages`. Removed the `typst/emoji.vim` and `typst/symbols.vim` variables.
* Removed the `CompilerSet` definition.
* Updated file headers
* Removed `'iskeyword'` modification
* I also removed most of the functionality in the `autoload/typst.vim` file (which were used for the custom user commands) and kept only the indentexpr function. If users want the `TypstWatch` or `Toc` commands they can use the full plugin.

chrisbra:
thanks!

PR #15235, opened 2024-07-13
Wrong comments in alt_tabpage()

zeertzjq:
Problem:  Wrong comments in alt_tabpage().
Solution: Correct the comments.


chrisbra:
thanks!

PR #15236, opened 2024-07-13
Fix test_gettext_make (again)

k-takata:
There was a mistake in v9.1.0570.
Correct how to check the existence of %GETTEXT_PATH%.


chrisbra:
thanks!

PR #15238, opened 2024-07-13
link javaConceptKind to javaStorageClass instead of NonText

### LINKED ISSUE #15237: javaConceptKind should not link to the NonText highlight group
### Steps to reproduce

Introduced by this commit: https://github.com/vim/vim/commit/9ecf02cd5f5feabe096f66cd5f503a822c78793f#diff-0fddd2a7f1b39c711eb9cf31da1cd6e9ca133ed96213fedf4ae46521f1796d37R353

Open a java source file with syntax highlighting on where the `final` keyword is used

### Expected behaviour

From `:help highlight`, the NonText highlight group is for "characters that do not really exist in the text (e.g., ">" displayed when a double-wide character doesn't fit at the end of the line).

Linking javaConceptKind to NonText seems very wrong given this definition, and in practice it also causes problems because NonText is often set to a fainter, less visible colour.

I would expect javaConceptKind to link to Type, or a similar highlight group as that seems far more appropriate.

### Version of Vim

9.1.0407

### Environment

Operating system: Void Linux, Terminal: xterm, Shell: bash

### Logs and stack traces

_No response_

dexgs:
From :help highlight, the NonText highlight group is for "characters that do not really exist in the text (e.g., ">" displayed when a double-wide character doesn't fit at the end of the line).

Because of this, linking the javaConceptKind highlight group (which consists of the keywords `abstract` and `final`) to NonText seems wrong.

This commit changes javaConceptKind to link to javaStorageClass instead. This seems more correct as javaStorageClass is the highlight group to which keywords `public` `static` `synchronized` (and others) are linked.

These keywords often occur together with `abstract` and `final`, e.g. `public abstract class ...` or `static final ...`

fixes #15237

### DIFF
--- a/runtime/syntax/java.vim
+++ b/runtime/syntax/java.vim
@@ -562,7 +562,7 @@ hi def link javaStorageClass		StorageClass
 hi def link javaMethodDecl		javaStorageClass
 hi def link javaClassDecl		javaStorageClass
 hi def link javaScopeDecl		javaStorageClass
-hi def link javaConceptKind		NonText
+hi def link javaConceptKind		javaStorageClass
 
 hi def link javaBoolean		Boolean
 hi def link javaSpecial		Special


zeertzjq:
cc @zzzyxwvut

zzzyxwvut:
I have left [a comment](https://github.com/vim/vim/issues/15237#issuecomment-2227039859) in the issue thread.

zzzyxwvut:
@dexgs, note that `javaConceptKind` items appear in many Java  
syntax test files, so make sure you review and _commit_ new  
screen dumps along with syntax file changes.  For guidance,  
see this [README.txt](https://github.com/vim/vim/blob/master/runtime/syntax/testdir/README.txt).

zzzyxwvut:
I'm fine with the proposed change.  However, there are some  
additional changes that I request:

1) Document the `javaConceptKind` highlighting group.  Feel  
free to improve upon the wording (note that there are no  
_methods_ only _functions_ under `:help ft-java-syntax`).

```diff
diff --git a/runtime/doc/syntax.txt b/runtime/doc/syntax.txt
index 3cb97f38f..c04954f07 100644
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -1,4 +1,4 @@
-*syntax.txt*   For Vim version 9.1.  Last change: 2024 Jul 08
+*syntax.txt*   For Vim version 9.1.  Last change: 2024 Jul 14


                  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -2063,6 +2063,10 @@ down some of their components may be of value.  Provided that such type names
 conform to the Java naming guidelines, you may arrange it with >
        :let java_highlight_generics=1

+Incompatible modifiers for types etc., e.g. `abstract` and `final`, can be
+differently highlighted than other modifiers with >
+       :hi link javaConceptKind NonText
+
 In Java 1.1 the functions System.out.println() and System.err.println() should
 only be used for debugging.  Therefore it is possible to highlight debugging
 statements differently.  To do this you must add the following definition in
```

2) Patch `java_contextual_keywords.java` as follows:

```diff
diff --git a/runtime/syntax/testdir/input/java_contextual_keywords.java b/runtime/syntax/testdir/input/java_contextual_keywords.java
index 094037964..2fbcddcdb 100644
--- a/runtime/syntax/testdir/input/java_contextual_keywords.java
+++ b/runtime/syntax/testdir/input/java_contextual_keywords.java
@@ -1,3 +1,7 @@
+// VIM_TEST_SETUP :hi link javaConceptKind NonText
+
+
+
 class ContextualKeywordsTests          // See JLS, §3.9 Keywords.
 {
        private ContextualKeywordsTests() { throw new Error(); }
```

3) Regenerate screen dump files, review them, make syntax  
tests pass locally, and commit the updated screen dump files.

(Note that `&ts` is automatically set to `4` for Java syntax  
tests, so indentation for `input/` files may expectedly differ  
from  `failed/` or `dumps/` screen dump files.)

benknoble:
@zzzyxwvut can you explain why the patch to relink to NonText in the test file is needed?

RE the documentation:
```
+Incompatible modifiers for types etc., e.g. `abstract` and `final`, can be
+differently highlighted than other modifiers with >
+       :hi link javaConceptKind NonText
+
```
what about these are "incompatible" modifiers? Other than I might suggest more like
```
Incompatible (?) modifiers for types, for example `abstract` and `final`, can be highlighted separately using `javaConceptKind`. Example: >
       :hi link javaConceptKind NonText

```

zzzyxwvut:
`java_contextual_keywords.java` is and _should remain_ a test  
file where `javaConceptKind` can be showcased; other test  
files will adopt the proposed change.

Well, `final` and `abstract` are incompatible with each other.

benknoble:
That makes sense. 

So the point of the original choice was to make two incompatible modifiers the same color (but different from the other modifiers)? I'm not sure I understand that: it seems like either using both should make the second an error highlight or it should just be left to the compiler to error when both are used. This is probably bikeshedding at this point, though. If I think about it long enough, maybe I understand that making these two stand out makes it easier to spot when both are used together…

zzzyxwvut:
The syntax group actually comprises five modifiers:

https://github.com/vim/vim/blob/581d4a7b356395bcb8606c1717ded65d47d26c68/runtime/syntax/java.vim#L119-L122

And their almost incompossibility serves for grouping.  Now,  
certain modifiers like `final` mean one thing when used with  
variables and mean another thing when used with methods or  
types.  There are `abstract` methods and types; and there  
are `default` methods for annotations and interfaces; and,  
finally, `(non-)sealed` types help to manage hierarchies.  
But the primary focus is on types and methods, with `native`  
(as opposed to `abstract`) and `volatile` (as opposed to `final`)  
considered and rejected: the former can only be used with  
methods _and_ it already enjoys a distinct colour, and the  
latter can only be used with field variables.

The reason for choosing a different colour is not related to  
linting.  For example, `abstract non-sealed` can be a valid  
combination of _two_ such modifiers for a `class` _permitted_ and  
derived from another `sealed` `class`.  This distinction in  
terms of colour can be useful as a hint of what follows (for  
reading) and what may follow (for writing).  Thus concepts  
about the immediate context can be formed and checked:
- for an `abstract` class, its API constructors need not be  
  `public`, and `abstract` methods can be declared;
- for a `final` class, `abstract` methods are invalid, `final`  
  methods are redundant, no subclasses can be derived;
- for an `abstract` method, there will be no method body  
  (and equally true for a `native` method) and the method  
  must belong to an `abstract` type and the modifier itself  
  will be redundant for an `abstract` interface method;
- for a `default` method, there will be a value (annotation)  
  or a method body (interface) or an error emitted;
- for a `sealed` type, the `permits` clause can be omitted when  
  all its members bear names and are defined in this very  
  compilation unit.

Naturally, long type/method bodies and line folding impair  
this usability.

zzzyxwvut:
@dexgs, you may find contributing to [the upstream repository](https://github.com/zzzyxwvut/java-vim.git)  
to be more approachable.  Its _raison d'être_ is to spare  
contributors the trouble of diving in CI targeted tasks,  
mandated by _this_ repository, that may raise formidable  
barriers for _some_ syntax code integration.  There is nothing  
over there to compile, generate, and collate with.  So no  
need to reach for Make, GCC and friends; let the maintainer  
wrangle with testing the accepted changes downstream.  It is  
also less of inconvenience for other upstream contributors  
and potentially yourself in the future when all Java related  
changes are introduced in the upstream repository, for it is  
imperative to ‘synchronise’ each to-be-further-modified file  
with its diverged copy from this repository, which spins its  
own Git yarn.

On the other hand, if you're looking for an opportunity to  
learn the ropes with a view to taking on the maintenance of  
some currently unmaintained syntax files:

```shell
grep is\ looking\ for -R runtime/syntax/
```

don't hesitate to follow up with questions when you run into  
difficulties about accomplishing the above requests.

chrisbra:
It seems there is no more activity here, so I am tempted to close this here and leave remaining improvements in @zzzyxwvut hands for future improvements.

zzzyxwvut:
@chrisbra, would you like me to open another PR, take the  
submitted changes, and complete the requested changes?

chrisbra:
Yes, if you think the changes here are worthwhile, but not complete, let's create a new PR with what you are satisfied. Thanks!

zzzyxwvut:
Okay.

zzzyxwvut:
The new PR is #15664.

PR #15239, opened 2024-07-13
Unnecessary checks for v:sizeoflong in test_put.vim

zeertzjq:
Problem:  Unnecessary checks for v:sizeoflong in test_put.vim.  They are
          no longer necessary as patch 8.2.3661 has changed the count to
          be within 32-bit integer limit.
Solution: Remove the checks.


chrisbra:
thanks!

PR #15241, opened 2024-07-13
Ex command is still executed after giving E1247

zeertzjq:
Problem:  Ex command is still executed after giving E1247.
Solution: Indicate the error properly.


chrisbra:
thanks!

PR #15245, opened 2024-07-13
:lmap mapping for keypad key not applied when typed in Select mode

zeertzjq:
Problem:  An :lmap mapping for a printable keypad key is not applied
          when typing it in Select mode.
Solution: Change keypad key to ASCII after setting vgetc_char.


chrisbra:
thanks!

PR #15247, opened 2024-07-13
runtime(macros): fix maze compilation error

mohd-akram:
Implicit int is an error in newer versions of clang.

### DIFF
--- a/runtime/macros/maze/mazeclean.c
+++ b/runtime/macros/maze/mazeclean.c
@@ -9,6 +9,7 @@
 char *M, A, Z, E = 40, line[80], T[3];
 int
 main (C)
+int C;
 {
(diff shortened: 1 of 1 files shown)


runtime/macros/maze/mazeclean.c
```diff
@@ -9,6 +9,7 @@
 char *M, A, Z, E = 40, line[80], T[3];
 int
 main (C)
+int C;
```
chrisbra, line 12: Err, I think we should use the normal ANSI C function definition here:
```suggestion
main (int C)
```

mohd-akram: I was keeping with the "work with all compilers" comment, but yeah I think in 2024 we're safe with using 1989 C :D

PR #15248, opened 2024-07-13
runtime(doc): document global insert behavior

mohd-akram:
This is a partial revert of 8f3f58f2c361f1b7241128d9821f88d8a30aa066 since this issue was fixed in #15120.

### DIFF
--- a/runtime/doc/insert.txt
+++ b/runtime/doc/insert.txt
@@ -2005,8 +2005,15 @@ command in ex mode: >
 	one
 	two
 <
-NOTE: These commands cannot be used with |:global| or |:vglobal|.
-":append" and ":insert" don't work properly in between ":if" and
+When these commands are used with |:global| or |:vglobal| then the lines are
+obtained from the text following the command.  Separate lines with a NL
+escaped with a backslash: >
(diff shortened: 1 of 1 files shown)


runtime/doc/insert.txt
```diff
@@ -2005,8 +2005,15 @@ command in ex mode: >
...
 <
-NOTE: These commands cannot be used with |:global| or |:vglobal|.
-":append" and ":insert" don't work properly in between ":if" and
+When these commands are used with |:global| or |:vglobal| then the lines are
+obtained from the text following the command.  Separate lines with a NL
+escaped with a backslash: >
+	:global/abc/insert\
+	one line\
+	another line
+The final "." is not needed then.
```
zeertzjq, line 2014: This only works in normal Ex mode. In cmdline mode or Vim Ex mode this doesn't work.

chrisbra: I'll amend this then. Thanks

mohd-akram:
It seems this was merged in 6d6dffa6113d44b34d1bb2d2945c7f17cfe0172a. Thanks!

PR #15249, opened 2024-07-13
fix: fuzzy completion

mg979:
Simpler alternative to #15193.

How it works: when using fuzzy completion, the completion leader is cut down to 1 character (if longer), the rest is stored in a "fuzzy completion leader" that is inserted later, as if the user typed it, so that the popup menu shows all matches that start with the leader and match (fuzzily) the rest of the characters.

Note that I'm using a trick to have the popup updated (simulating an additional inserted character, followed by a backslash).

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -145,6 +145,12 @@ static int	  compl_enter_selects = FALSE;
 // are used.
 static char_u	  *compl_leader = NULL;
 
+// When "compl_leader_fuzzy" is not NULL, it's the part of "compl_orig_text"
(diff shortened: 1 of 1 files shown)


mg979:
fuzzy3

@glepnir @chrisbra @habamax please test/have a look.

glepnir:
Once again, regex cannot handle fuzzy.

mg979:
> Once again, regex cannot handle fuzzy. Spend some time understanding what fuzzy is

Just in case you take it back again.

chrisbra:
I guess we can close this one then.

PR #15250, opened 2024-07-13
filetype: recognize LaTeX files generated by Inkscape

jdujava:
For details see `--export-latex` in https://inkscape.org/doc/inkscape-man.html.


chrisbra:
Thanks!

Issue #14902, opened 2024-06-03
modifyOtherKeys=2 key mapping discrepancy


gnachman:
### Steps to reproduce

In a terminal emulator that supports XTMODKEYS 4;2, such as xterm 392 or iTerm2 3.5.0:

1. Run `vim -i NONE -u NONE test.txt`. Here's a copy of [test.txt](https://github.com/user-attachments/files/15528566/test.txt)
2. `:source test.txt`
3. Press <kbd>Ctrl</kbd><kbd>G</kbd> and then a non-control key like <kbd>j</kbd>
4. `Single press!` is not echoed.

In a terminal emulator that does not support mode 2 of `modifyOtherKeys`, such as Apple Terminal, the bug does not reproduce and `Single press!` *is* echoed.

I believe it is related to this key reporting mode because the bug appeared in iTerm2 after I added support for `CSI >4;2m`. In this mode, `<C-G>` sends `CSI 27;5;103~`. I have confirmed that vim 9.1 sends `CSI > 4;2m` at startup when run as above.

This was [originally filed as a bug against iTerm2](https://gitlab.com/gnachman/iterm2/-/issues/11569).

### Expected behaviour

`echom 'Single press!'<CR>` should have run, showing `Single press!` on the bottom of the screen.

### Version of Vim

9.1

### Environment

OS: macOS 14.5
Terminal: iTerm2 3.5.0, xterm 392
$TERM: xterm-256color
Shell: tcsh

zeertzjq:
Related #11061

chrisbra:
can you please share a logfile created with `--log` ?

gnachman:
Here is the log you requested, demonstrating the bug:
[log.txt](https://github.com/user-attachments/files/15540100/log.txt)

I made a second log with the terminal in the standard key reporting mode, where it does the right thing:
[log2.txt](https://github.com/user-attachments/files/15540123/log2.txt)

PR #15251, opened 2024-07-13
Fix for mapping processing when capital letters are represented with terminal codes

disRecord:
Problem: there are two mappings and
1) the first mapping is substring of the second,
2) the first non-matching letter is capital,
3) capital letters are represented with termcodes "ESC[27;2;<ascii code>~" on given system 
then first mapping is applied instead of second.

Example:

    :map B b
    :map BBB blimp!

and then

    BBB -> bbb

instead of

    BBB -> blimp!

Solution: force termcodes check if capital letter does not match.

### DIFF
--- a/src/getchar.c
+++ b/src/getchar.c
@@ -2726,8 +2726,11 @@ handle_mapping(
 		    }
 		}
 		else
+		{
 		    // No match; may have to check for termcode at next
-		    // character.  If the first character that didn't match is
+		    // character.  
+
+		    // If the first character that didn't match is
 		    // K_SPECIAL then check for a termcode.  This isn't perfect
 		    // but should work in most cases.
 		    if (max_mlen < mlen)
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. Can you please add a test for this?

chrisbra:
I thought it would also fix #14902 but unfortunately it doesn't. 

I have added a test and fixed the whitespace errors. Thanks, I included it.

disRecord:
Thanks. 

I tried to write more thorough test (https://github.com/vim/vim/commit/f89b0c3903bb0407d84733cac95b8e9dfb257e92) which tests overlapping sequences with  different modifiers. But it result is the same.

chrisbra:
so the current test is okay?

disRecord:
My code variant tests overlapping sequences with different modifiers (ctrl, alt, shift) and with escapes provided by `GetEscCodeCSI27()` and `GetEscCodeCSIu()`.  But it modifies original `Test_mapping_works_with_*()`  tests behavior. And simple test above still detects problem. 

I do not know input subsystem of vim well enough to appraise if new patch is necessary.

chrisbra:
sorry, it's unclear to me what you did and why. Addtitional tests should be always nice to add, so please create a new PR with additional tests then. If possible, don't try to change existing tests, so we can be reasonably sure the existing functionality is being tested as it was before

PR #15252, opened 2024-07-13
runtime: asymptote

avidseeker:
This adds the support for [Asymptote](https://asymptote.sourceforge.io/) Vector Graphics Language, and avoids packaging issues such as https://github.com/vectorgraphics/asymptote/issues/180.

Imported from [base/asy.vim](https://github.com/vectorgraphics/asymptote/blob/master/base/asy.vim).


runtime/ftplugin/asy.vim
```diff
@@ -0,0 +1,16 @@
...
+"
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let g:did_ftplugin = 1
+
+setlocal commentstring=/*\ %s\ */
+setlocal makeprg=asy\ %
+setlocal errorformat=%f:\ %l.%c:\ %m
```
dkearns: `makeprg` and `errorformat` should be moved to a dedicated compiler plugin.

runtime/syntax/asy.vim
```diff
@@ -0,0 +1,260 @@
...
+  HiLink asyComment              Comment
+  HiLink asySpecial              SpecialChar
+  HiLink asyCSpecial             SpecialChar
+  HiLink asyTodo                 Todo
+  HiLink asyPathSpec             Statement
+
+  delcommand HiLink
+endif
+
+let b:current_syntax = "c"
```
zeertzjq: Wrong value

runtime/syntax/asy.vim
```diff
@@ -0,0 +1,260 @@
+" Vim syntax file
+" Language:     Asymptote
+" Maintainer:   Andy Hammerlindl
```
dkearns: Is Andy going to handle future updates to this repo?  If not could you add yourself as a co-maintainer, or even the sole maintainer if that's more appropriate?

runtime/syntax/asy.vim
```diff
@@ -0,0 +1,260 @@
...
+" Hacked together from Bram Moolenaar's C syntax file, and Claudio Fleiner's
+" Java syntax file.
+
+" For version 5.x: Clear all syntax items
+" For version 6.x: Quit when a syntax file was already loaded
+if version < 600
+  syn clear
+elseif exists("b:current_syntax")
+  finish
+endif
```
dkearns, line 12: This block can be reduced to the following.  Vim 6 was released over twenty years ago and the special handling for Vim 5 was removed from the distributed files some time ago.

```vim
if exists("b:current_syntax")
  finish
endif
```

runtime/syntax/asy.vim
dkearns:
```suggestion
let b:current_syntax = "asy"
```

runtime/syntax/asy.vim
```diff
@@ -0,0 +1,260 @@
...
+  HiLink asyString               String
+  HiLink asyCString              String
+  HiLink asyComment              Comment
+  HiLink asySpecial              SpecialChar
+  HiLink asyCSpecial             SpecialChar
+  HiLink asyTodo                 Todo
+  HiLink asyPathSpec             Statement
+
+  delcommand HiLink
+endif
```
dkearns: This conditional can be removed along with the use of the `HiLink` command and replaced with `hi def link`.   See other syntax files for an example.

runtime/syntax/asy.vim
```diff
@@ -0,0 +1,260 @@
...
+syn keyword     asyConstant      realEpsilon realDigits
+syn keyword     asyPathSpec      and cycle controls tension atleast curl
+syn keyword     asyStorageClass  static public restricted private explicit
+syn keyword     asyStructure     struct typedef
+syn keyword     asyType          void bool bool3 int real string file
+syn keyword     asyType          pair triple transform guide path pen frame
+syn keyword     asyType          picture
+
+" module specific keywords
+if exists("asy_syn_plain")
```
dkearns, line 35: It would be nice if these config variables were documented at `:help ft-asy-syntax`.

dkearns (changes requested):
Thanks.

Could you also add entries to https://github.com/vim/vim/blob/master/.github/MAINTAINERS?

chrisbra:
Thanks. Can you please also add a test to testdir/test_filetype.vim?

runtime/syntax/asy.vim
clason: Why? Asymptote is basically C.

runtime/syntax/asy.vim
avidseeker: Last change to that file is [3 years ago](https://github.com/vectorgraphics/asymptote/commits/master/base/asy.vim). I'll add myself as a maintainer.

runtime/filetype.vim
```diff
@@ -111,6 +111,9 @@ au BufNewFile,BufRead *.ino,*.pde		setf arduino
 " Ash of busybox
 au BufNewFile,BufRead .ash_history		setf sh
 
+" Asymptote
+au BufNewFile,BufRead *.asy		setf asy
```
zeertzjq, line 115: This should be `asymptote` now

runtime/syntax/asymptote.vim
```diff
@@ -0,0 +1,243 @@
...
+hi def link asyCommentSkip          asyComment
+hi def link asyString               String
+hi def link asyCString              String
+hi def link asyComment              Comment
+hi def link asySpecial              SpecialChar
+hi def link asyCSpecial             SpecialChar
+hi def link asyTodo                 Todo
+hi def link asyPathSpec             Statement
+
+let b:current_syntax = "asy"
```
zeertzjq: ditto

src/testdir/test_filetype.vim
```diff
@@ -107,6 +107,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     asn: ['file.asn', 'file.asn1'],
     asterisk: ['asterisk/file.conf', 'asterisk/file.conf-file', 'some-asterisk/file.conf', 'some-asterisk/file.conf-file'],
     astro: ['file.astro'],
+    asy: ['file.asy'],
```
zeertzjq, line 110: ditto

runtime/filetype.vim
avidseeker: You mean `file.asymptote`? The common extension is `.asy`. I changed *the name* of the filetype which is different from the extension. E.g: filetype name of `.md` is markdown, and no one use `file.markdown` as a filename.

zeertzjq: `setf asymptote`

src/testdir/test_filetype.vim
avidseeker:
```
    asy: ['file.asymptote', 'file.asy'],
```
Both should be included, no?

zeertzjq: No,
```
    asymptote: ['file.asy'],
```

runtime/doc/syntax.txt
```diff
@@ -958,6 +958,25 @@ For Visual Basic use: >
...
+By default, only basic Asymptote keywords are highlighted. To highlight
+extended geometry keywords: >
+
+	:let g:asy_syn_plain = 1
+
+and for highlighting keywords related to 3D constructions: >
+
+	:let g:asy_syn_three = 1
+
+By default, Asymptote-defined color (e.g: lightblue) are highlighted. To
```
dkearns:
```suggestion
By default, Asymptote-defined colors (e.g: lightblue) are highlighted. To
```

dkearns (changes requested):
Thanks for the changes.

Is the filetype name change going to cause problems for existing users of the upstream repo?  Will it be changed there also?

avidseeker:
I have in mind submitting a PR to asymptote and packagers (Fedora) with the changes after merging this. Asymptote suggests to user copying `asy.vim` to their user config, so this might be a breaking change.

chrisbra:
> If it's already out there in the wild I think it would be better not to change the name.

Yeah, agree. Let't not change it please.

runtime/filetype.vim
```diff
@@ -111,6 +111,9 @@ au BufNewFile,BufRead *.ino,*.pde		setf arduino
 " Ash of busybox
 au BufNewFile,BufRead .ash_history		setf sh
 
+" Asymptote
+au BufNewFile,BufRead *.asy		setf asymptote
```
dkearns:
```suggestion
au BufNewFile,BufRead *.asy		setf asy
```

runtime/doc/syntax.txt
```diff
@@ -958,6 +958,26 @@ For Visual Basic use: >
 	:let g:filetype_asa = "aspvbs"
 	:let g:filetype_asp = "aspvbs"
 
+ASY							*ft-asy-syntax*
+ASYMPTOTE						*ft-asymptote-syntax*
```
dkearns:
```suggestion
ASYMPTOTE					*asy.vim* *ft-asy-syntax*
```

I think it probably makes most sense to title the section with the proper name and use the actual filetype in the tags.  The PPWizard (ppwiz) section does this.

The filename is also usually included in the tag list.  I know it's not in the docs directly above this entry.

dkearns (changes requested):

dkearns (approved):
Looks good, thanks.

Sorry for the naming merry-go-round.

PR #15220, opened 2024-07-12
Fix Prolog file detection regex

ignamartinoli:
Problem: filetype: .pro file detection for Prolog is broken
Solution: fixed the regex to only match on the tested cases (igna_martinoli)

Fixes: https://github.com/vim/vim/issues/10835
Closes: https://github.com/vim/vim/pull/15206

PR #15253, opened 2024-07-13
runtime(filetype): Update FTpl() regex for Prolog detection

### LINKED ISSUE #10835: prolog filetype detection

ignamartinoli:
Problem: The regex in the FTpl() function was outdated
Solution: Updated the regex as per https://github.com/vim/vim/pull/15220/files

fixes: #10835
closes: #15206

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -465,7 +465,7 @@ export def ProtoCheck(default: string)
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var lnum = getline(nextnonblank(1))
-    if lnum =~ '\<prolog\>' || lnum =~ '^\(:-\|%\|\/\*\)\|\.$'
+    if lnum =~ '\<prolog\>' || lnum =~ '^\s*\(:-\|%\+\(\s\|$\)\|\/\*\)\|\.\s*$'
       setf prolog
     else
       exe 'setf ' .. default
@@ -644,7 +644,7 @@ export def FTpl()
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var line = getline(nextnonblank(1))
-    if line =~ '\<prolog\>' || line =~ '^\(:-\|%\|\/\*\)\|\.$'
+    if line =~ '\<prolog\>' || line =~ '^\s*\(:-\|%\+\(\s\|$\)\|\/\*\)\|\.\s*$'
       setf prolog
     else
       setf perl
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -644,7 +644,7 @@ export def FTpl()
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var line = getline(nextnonblank(1))
-    if line =~ '\<prolog\>' || line =~ '^\s*\(%\+\(\s\|$\)\|/\*\)' || line =~ ':-'
+    if line =~ '\<prolog\>' || line =~ '^\(:-\|%\|\/\*\)\|\.$'
       setf prolog
     else
       setf perl
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -2614,28 +2614,28 @@ func Test_pl_file()
 
   "Prolog
   call writefile([':-module(test/1,'], 'Xfile.pro', 'D')
-  split Xfile.pro
+  split Xfile.pl
   call assert_equal('prolog', &filetype)
   bwipe!
 
   call writefile(['% comment'], 'Xfile.pro', 'D')
-  split Xfile.pro
+  split Xfile.pl
   call assert_equal('prolog', &filetype)
   bwipe!
 
   call writefile(['/* multiline comment'], 'Xfile.pro', 'D')
-  split Xfile.pro
+  split Xfile.pl
   call assert_equal('prolog', &filetype)
   bwipe!
 
   call writefile(['rule(test, 1.7).'], 'Xfile.pro', 'D')
-  split Xfile.pro
+  split Xfile.pl
   call assert_equal('prolog', &filetype)
   bwipe!
 
   " Perl
   call writefile(['%data = (1, 2, 3);'], 'Xfile.pro', 'D')
-  split Xfile.pro
+  split Xfile.pl
   call assert_notequal('prolog', &filetype)
 
   filetype off
--- a/src/testdir/test_filetype.vim
+++ b/src/testdir/test_filetype.vim
@@ -2613,28 +2613,28 @@ func Test_pl_file()
   filetype on
 
   "Prolog
-  call writefile([':-module(test/1,'], 'Xfile.pro', 'D')
+  call writefile([':-module(test/1,'], 'Xfile.pl', 'D')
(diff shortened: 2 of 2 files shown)


zeertzjq:
The comments in `FTpl()` and `FTpro()` also no longer match the code, so they should be updated.

runtime/autoload/dist/ft.vim
```diff
@@ -644,7 +644,7 @@ export def FTpl()
     # recognize Prolog by specific text in the first non-empty line
     # require a blank after the '%' because Perl uses "%list" and "%translate"
     var line = getline(nextnonblank(1))
-    if line =~ '\<prolog\>' || line =~ '^\s*\(%\+\(\s\|$\)\|/\*\)' || line =~ ':-'
+    if line =~ '\<prolog\>' || line =~ '^\(:-\|%\|\/\*\)\|\.$'
```
zeertzjq: This doesn't match the comment above that requires the `%\+\(\s\|$\)` pattern:
```
    # require a blank after the '%' because Perl uses "%list" and "%translate"
```
Meanwhile, the comment in `ProtoCheck()` needs to be updated, as it no longer matches the code there and `ProtoCheck()` has nothing to do with Perl.

ignamartinoli: @zeertzjq Do you suggest removing that line (645)? Or should it be better to write there a comment describing the RegEx?

zeertzjq: No, I think the requirement in the comment needs to be respected.

ignamartinoli: Aaaaaah I think I get it.

A Perl file may begin with, for example:

```perl
%data = ('John Paul', 45, 'Lisa', 30, 'Kumar', 40);
```

And it may get confused for a Prolog comment, like:

```prolog
% comment
```

Did I understand correctly?

zeertzjq: Yes

ignamartinoli: Alright, thanks for the patience, I modified the Regex to this:

```vim
let regex = '^\(:-\|%\+\(\s\|$\)\|\/\*\)\|\.$'

let pattern1 = ':-module(test/1,'
let pattern2 = '% comment'
let pattern3 = '/* multiline comment'
let pattern4 = 'rule(test, 1.7).'
let patternX = 'a'
let patternY = '%data = ("John Paul", 45, "Lisa", 30, "Kumar", 40);'

if pattern1 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern2 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern3 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if pattern4 =~ regex
  echo 'OK'
else
  echo 'ERROR'
endif

if patternX =~ regex
  echo 'ERROR'
else
  echo 'OK'
endif

if patternY =~ regex
  echo 'ERROR'
else
  echo 'OK'
endif
```

All tests are passing.

The `patternX` and `patternY` aren't supposed to match (I'm adding them as tests too).

Do you happen to see any other weird thing?

zeertzjq: There may be leading whitespace before the text, so I think it should still be `^\s*` at the start. That may apply to `ProtoCheck()` as well.

ignamartinoli: Alright, I improved the Regex for leading and trailing whitespace.

I'm not that experienced with PRs. I did `git add .` and `git commit`.
Now I should do `git push fork update-filetype-FTpl`, right?

zeertzjq: Yes

ignamartinoli: Thanks!

zeertzjq:
It seems that there should be a `Test_pl_file()` test as well.

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+
+  filetype off
+endfunc
+
+
+func Test_pl_file()
+  filetype on
+
+  "Prolog
+  call writefile([':-module(test/1,'], 'Xfile.pro', 'D')
```
```suggestion
  call writefile([':-module(test/1,'], 'Xfile.pl', 'D')
```

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+  split Xfile.pl
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  call writefile(['% comment'], 'Xfile.pro', 'D')
```
```suggestion
  call writefile(['% comment'], 'Xfile.pl', 'D')
```

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+  split Xfile.pl
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  call writefile(['/* multiline comment'], 'Xfile.pro', 'D')
```
```suggestion
  call writefile(['/* multiline comment'], 'Xfile.pl', 'D')
```

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+  split Xfile.pl
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  call writefile(['rule(test, 1.7).'], 'Xfile.pro', 'D')
```
```suggestion
  call writefile(['rule(test, 1.7).'], 'Xfile.pl', 'D')
```

ignamartinoli:
Alright, tests added. If the CI/CD Workflow succeeds, we should be good to go, I guess

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+  split Xfile.pl
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  " Perl
+  call writefile(['%data = (1, 2, 3);'], 'Xfile.pro', 'D')
+  split Xfile.pl
+  call assert_notequal('prolog', &filetype)
```
```suggestion
  call writefile(['%data = (1, 2, 3);'], 'Xfile.pl', 'D')
  split Xfile.pl
  call assert_equal('perl', &filetype)
```

ignamartinoli:
Thanks for pointing out those typos @zeertzjq

src/testdir/test_filetype.vim
```diff
@@ -2600,6 +2600,44 @@ func Test_pro_file()
...
+
+  call writefile(['rule(test, 1.7).'], 'Xfile.pl', 'D')
+  split Xfile.pl
+  call assert_equal('prolog', &filetype)
+  bwipe!
+
+  " Perl
+  call writefile(['%data = (1, 2, 3);'], 'Xfile.pl', 'D')
+  split Xfile.pl
+  call assert_notequal('prolog', &filetype)
```
```suggestion
  call assert_equal('perl', &filetype)
```

ignamartinoli:
Ok, lesson learned. I won't copy-paste a test function again, hahaha :facepalm:

chrisbra:
I think we can further improve it, by using a single pattern variable for the prolog pattern. Otherwise we will have to change it (hopefully we won't ever ;)) in 2 distinct places. That's how I'll merge it at least. 

Thanks all!

ignamartinoli:
Thanks for the help everyone

PR #15258, opened 2024-07-14
Printed line doesn't overwrite colon when pressing Enter in Ex mode

zeertzjq:
Problem:  Printed line no longer overwrites colon when pressing Enter in
          Ex mode (after 9.1.0573).
Solution: Restore the behavior of pressing Enter in Ex mode.


chrisbra:
thanks!

PR #15259, opened 2024-07-14
Various lines are indented inconsistently

zeertzjq:
Problem:  Various lines are indented inconsistently.
Solution: Retab these lines and correct some comments.


chrisbra:
thanks!

Issue #15254, opened 2024-07-13
TermdebugCommand E1023


step-:
In version 9.1.0391 starting the debugger in the following way worked:

    :TermdebugCommand /path/to/program arg args </dev/null

but version 9.1.580 throws and error (then the debugger starts):

Error detected while processing function <SNR>257_StartDebugCommand[2]..<SNR>257_StartDebug_internal[43]..<SNR>257_StartDebug_term[16]..<SNR>257_CreateGdbConsole:
line    74:
E1023: Using a Number as a Bool: 3

chrisbra:
@ubaldot please

ubaldot:
> @ubaldot please

I'll look into it later on today. I'll also add the patch you suggested regarding the whitespaces in filenames.

step-:
@ubaldot, thanks for porting termdebug to vim9 script. I use the old version a lot, and it has its own issues. I'm looking forward to your additions to termdebug.

ubaldot:
It should be fixed with Implemented in https://github.com/vim/vim/pull/15261. Please, double check and confirm so we can close the issue.

step-:
Fix working, confirmed. Thanks @ubaldot

Issue #12357, opened 2023-05-07
Termdebug broken on windows (program window missing and `:Break` doesn't work)


JJCUBER:
### Steps to reproduce

1. install mingw-w64 (I'm using 13.1.0 from winlibs)
2. open vim (or gvim) with `--clean`
3. run `:packadd termdebug`
4. run `:Termdebug a` (a is an executable compiled using gcc 13.1.0 with the flag `-g`)
---
This loads fine and all gdb commands input directly into the gdb window work as expected, except:
- the program window is missing
- using `run` in the gdb window makes the executable show/pop up in a completely separate window/terminal instance
- `:Break` does not work; a message appears in the gdb window saying, `-break-insert: Garbage following <location>`
- trying to switch the gdb layout using `layout asm` does not work (this might be just be a limitation and not related to my issue?)

### Expected behaviour

- There should be a program window where the executable to be run should "live"
- `:Break` should set a breakpoint at the current line of the source file
- layout switching in gdb shouldn't give an error (unless this is just a limitation)

### Version of Vim

9.0.0000

### Environment

OS: Windows 10 (22H2)
Terminal: Command Prompt (also tried using gvim)
mingw-w64: 13.1.0
vim (x64): 9.0.0000 (from [vim/vim-win32-installer](https://github.com/vim/vim-win32-installer/releases))

### Logs and stack traces

```shell
gdb> file a
Reading symbols from a...
----
-break-insert: Garbage following <location>
----
gdb> layout asm
Cannot enable the TUI when the interpreter is 'mi2'
----
gdb> break 6
Breakpoint 1 at 0x7ff7b76f16dd: file test.cpp, line 6.
gdb> run
Starting program: C:\\Users\\JJCUBER\\Desktop\	esting clangd\\a.exe 
[New Thread 29876.0x7ab0]
[New Thread 29876.0x5018]
[New Thread 29876.0x6b90]

Thread 1 hit Breakpoint 1, main () at test.cpp:6
6	    std::cout << \
gdb> step
7	    return 0;
gdb> step
8	}
gdb> step
0x00007ff7b76f1340 in __tmainCRTStartup ()
gdb> step
Single stepping until exit from function __tmainCRTStartup,which has no line number information.
[Thread 29876.0x5018 exited with code 0]
[Thread 29876.0x7ab0 exited with code 0]
[Thread 29876.0x6b90 exited with code 0]
[Inferior 1 (process 29876) exited normally]
gdb>
```

[...]

Issue #10300, opened 2022-04-27
termdebug: switch to tokenized messages


GitMensch:
Currently everything is "anonymous and asynchronous" in termdebug; we send an instruction and do a best guess when evaluating the response what the response is about. This issue is about tracking a change to this (whenever it may be done)

[ ] remove message anonymity by adding tokens to the messages sent (which is in general the suggested and commonly documented way)
[ ] for evaluation: use the token to _know_ what the response is for

This also allows to fix the guessing about the stop state:

* sent out an instruction that will tell us if the inferior is running or not
* internally block _all_ further messages that potentially tinker with the run state (easiest: all messages, possibly also block interaction with the GDB buffer)
* that way wait for the answer to the message we sent out (has its token on return so can be distinguished)
* handle the answer - we now know the current state
* unblock everything
* do something with the information "stopped yes/no"

_Originally posted by @GitMensch in https://github.com/vim/vim/issues/9158#issuecomment-981981900_

brammool:
I don't know what you mean with "end an instruction".  Are you talking about the communication with gdb?  That uses a gdb API.

GitMensch:
Yes, we use the GDB API - but we don't do what the documentation says: when using async mode pass a token on each api-call and check it when reading responses to recognize the request it belongs to.

GitMensch:
The GDB manuals had that in somewhere, I didn't found it with a quick ckeck.

But the [Input-Syntax](https://sourceware.org/gdb//onlinedocs/gdb/GDB_002fMI-Input-Syntax.html#GDB_002fMI-Input-Syntax) and [Output-Syntax](https://sourceware.org/gdb//onlinedocs/gdb/GDB_002fMI-Output-Syntax.html#GDB_002fMI-Output-Syntax) has the doc for the optional token - if we pass it in, we also get it in the response.

> If an input command was prefixed with a token then the corresponding output for that command will also be prefixed by that same token. 

Don't be confused like I was from the note that says that `async-output` has its tokens omitted - we commonly don't react on those, but on the other responses that are also returned in async mode.

brammool:
> The GDB manuals had that in somewhere, I didn't found it with a quick ckeck.
> 
> But the
> [Input-Syntax](https://sourceware.org/gdb//onlinedocs/gdb/GDB_002fMI-Input-Syntax.html#GDB_002fMI-Input-Syntax)
> and
[...]

PR #15261, opened 2024-07-14
Filenames with whitespaces AND ':Termdebug args' issues fix. 

ubaldot:
1. Filenames with whitespaces issue should be fixed now  #12357 
2. ":Termdebug args" should work now #15254

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -522,7 +522,7 @@ def CreateGdbConsole(dict: dict<any>, pty: string, commpty: string): string
 
   # ---- gdb started. Next, let's set the MI interface. ---
   # Set arguments to be run.
-  if len(proc_args)
+  if !empty(proc_args)
     term_sendkeys(gdbbufnr, $"server set args {join(proc_args)}\r")
   endif
 
@@ -1321,12 +1321,8 @@ def DeleteCommands()
   endif
 
   sign_unplace('TermDebug')
-  breakpoints = {}
-  breakpoint_locations = {}
-
   sign_undefine('debugPC')
   sign_undefine(BreakpointSigns->map("'debugBreakpoint' .. v:val"))
-  BreakpointSigns = []
 enddef
 
 
@@ -1339,7 +1335,7 @@ def Until(at: string)
     ch_log('assume that program is running after this command')
 
     # Use the fname:lnum format
-    var AT = empty(at) ? $"{fnameescape(expand('%:p'))}:{line('.')}" : at
+    var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at
     SendCommand($'-exec-until {AT}')
   else
     ch_log('dropping command, program is running: exec-until')
@@ -1358,7 +1354,7 @@ def SetBreakpoint(at: string, tbreak=false)
   endif
 
   # Use the fname:lnum format, older gdb can't handle --source.
-  var AT = empty(at) ? $"{fnameescape(expand('%:p'))}:{line('.')}" : at
+  var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at
   var cmd = ''
   if tbreak
     cmd = $'-break-insert -t {AT}'


ubaldot:
I also remove the reset of few variables at shutoff as it is no longer needed given that we have an Init function.

chrisbra:
thanks!

ubaldot:
@chrisbra the changes `var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at` **shall be reverted**! 

I am experiencing issues  with gdb saying `no source file named C` when I try to set a breakpoint. I cannot root-cause the problem, but by restoring `var AT = empty(at) ? $"{fnameescape(expand('%:p'))}:{line('.')}" : at` things work well again.

errael:
> @chrisbra the changes `var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at` **shall be reverted**!
> 
> I am experiencing issues with gdb saying `no source file named C` when I try to set a breakpoint. I cannot root-cause the problem, but by restoring `var AT = empty(at) ? $"{fnameescape(expand('%:p'))}:{line('.')}" : at` things work well again.

What does the string with/without PR changes look like?

ubaldot:
> What does the string with/without PR changes look like?

Before PR: `C:\Users\yt75534\Documents\defrecs\stm32f4x\ADC_Readings\application\temperature_sensor\tempsens_LM335.c:60`
After PR: `"C:\Users\yt75534\Documents\defrecs\stm32f4x\ADC_Readings\application\temperature_sensor\tempsens_LM335.c:60"`

It is only the double quote that differs. But for some reason, in my case the double quotes make troubles,. However, I am using a special gdb (arm-eabi-none-gdb).

errael:
> It is only the double quote that differs. But for some reason, in my case the double quotes make troubles,. However, I am using a special gdb (arm-eabi-none-gdb on Windows 10).

I just tried this using current build plus #15226. 
```
588.519723005 : ch_log(): sending to gdb: -break-insert "/src/tools/vim/src/evalfunc.c:7543"
```

Things work OK. I'm on linux. If things work OK on  windows with regular gdb then I'd be inclined to keep the current behavior and add an option for "don't quote mi command options" or whatever.

Caveat, I didn't try anything with blanks in the file name.

ubaldot:
> Things work OK. I'm on linux. If things work OK on windows with regular gdb then I'd be inclined to keep the current behavior and add an option for "don't quote mi command options" or whatever.

I agree. I may add an "if" condition to check if users is running the regular gdb or another version. 

> Caveat, I didn't try anything with blanks in the file name.

The best would be to add a test with filenames with spaces and without. At that point we leave the github workflow to run the tests for us. I should have some time for doing it today.

ubaldot:
Update: tested regular gdb on Windows and WSL2 (Ubuntu) with both files with empty spaces and without empty spaces. They all work well. Hence, the problem relies on arm-eabi-none-gdb, but then the same symptom may appear on other gdb:s. 

I could add something like the following:

```
var weird_gdbs = ['arm-eabi-none-gdb', 'my-gdb', 'another-weird-gdb' ]

var AT =""
if index(weird_gdbs, GetCommand()[0]) != -1  
  AT = empty(at) ? $"{fnameescape(expand('%:p'))}:{line('.')}" : at
else 
   AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at
endif
```

and populate the list `weird_gdbs` as long as users ask for support for a specific gdb? 

What do you think?

errael:
In the spec for [mi3](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Input-Syntax.html#GDB_003fMI-Input-Syntax) there's
```
parameter →
    non-blank-sequence | c-string

non-blank-sequence →
    anything, provided it doesn’t contain special characters
    such as "-", nl, """ and of course " "

c-string →
    """ seven-bit-iso-c-string-content """
```

Handling this problem in terms of gdb/mi compliance seems easier to maintain
and document; having a configurable like `gdb_mi_parameter_can_not_quote` or
`gdb_mi_broken_c_string` or something. In any event, it seems like its time
for a function that sits near `SendCommand`, something like
`CreateMiParameter` or `CreateParameter` instead of putting work-around code
all over the place.

I've wondered about have more inteligence in `SendCommand()` to automatically
format gdb/mi messages. But that seems like a lot of work; and since
`termdebug` has been around for a while, probably not worth it. (at least not
at this time)

errael:
(I lost something from the previous post)

> populate the list non_standard_gdbs as long as users ask for support for a specific gdb?

It seems like using `GetCommand()[0]` is a problem. Couldn't some weird command
have different versions, some of which handle quotes?
(followed by discussion of phrasing this issue in terms of gdb/mi compliance
rather than specific/innumerable gdb versions)

LemonBoy:
> What do you think?

Christ.

The problem is solved (with `arm-none-eabi-gdb` on Windows 11) by escaping all the backslashes, if you look at the path you're passing you'll notice a number of backslash + letter combinations that are valid escape sequences. Please add a helper function (`QuoteArg`) to 1) add quotes if not present 2) escape the quotes and backslashes present in the string, do not repeat the logic everywhere.

You probably want to apply this when sending the `file <arg>` command to gdb.

One more issue, if you specify an absolute path for `gdb_cmd` the `$"You have a file/folder named '{gdb_cmd}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."` error is raised. The check should be refined to actually check the file exists in the CWD.

ubaldot:
> In the spec for [mi3](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Input-Syntax.html#GDB_003fMI-Input-Syntax) there's
> 
> ```
> parameter →
>     non-blank-sequence | c-string
> 
> non-blank-sequence →
>     anything, provided it doesn’t contain special characters
>     such as "-", nl, """ and of course " "
> 
> c-string →
>     """ seven-bit-iso-c-string-content """
> ```
> 
> Handling this problem in terms of gdb/mi compliance seems easier to maintain and document; having a configurable like `gdb_mi_parameter_can_not_quote` or `gdb_mi_broken_c_string` or something. In any event, it seems like its time for a function that sits near `SendCommand`, something like `CreateMiParameter` or `CreateParameter` instead of putting work-around code all over the place.
> 

In-fact, I admit that what I proposed is utterly bad! 
FYI: there is also a mi4 (not documented). The problem in upgrading mi interface also require the user to upgrade gdb. But yes, I wonder how many are still stuck to very old gdb versions... 

> I've wondered about have more inteligence in `SendCommand()` to automatically format gdb/mi messages. But that seems like a lot of work; and since `termdebug` has been around for a while, probably not worth it. (at least not at this time)

Agree. Also, take a look here #10300

ubaldot:
> One more issue, if you specify an absolute path for `gdb_cmd` the `$"You have a file/folder named '{gdb_cmd}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."` error is raised. The check should be refined to actually check the file exists in the CWD.

Ops! Sorry! -> https://github.com/vim/vim/pull/15268

ubaldot:
> The problem is solved (with `arm-none-eabi-gdb` on Windows 11) by escaping all the backslashes, if you look at the path you're passing you'll notice a number of backslash + letter combinations that are valid escape sequences. Please add a helper function (`QuoteArg`) to 1) add quotes if not present 2) escape the quotes and backslashes present in the string, do not repeat the logic everywhere.
> 

Do you mean something like this: https://github.com/vim/vim/pull/15270 ?

Issue #4390, opened 2019-05-18
[Bug]w in cw is same as ce


icyhat:
**`w in cw is same as ce`**
When `w` combine with `c`, `cw` behaves same as `ce`. 

They both delete from current cursor to end of current word and insert.

But `cw` should delete from current cursor to where until next word which contain the blanks between it and next word.

**To Reproduce**
```
alpha text:         hello world vim
target word:              world
de:                 hello  vim
dw:                 hello vim
ce:                 hello  vim
cw:                 hello  vim
```
**Expected behavior**
`cw` has two parts, delete and insert. its deletion should be like `dw` which delete until next word, including blanks.
**Environment**
 - Vim version [version 8.1.320] 
 - OS: [Ubuntu 19.04]
 - Terminal: [GNOME Terminal)

tonymec:
This is documented (under `:help WORD`):
```
Special case: "cw" and "cW" are treated like "ce" and "cE" if the cursor is
on a non-blank.  This is because "cw" is interpreted as change-word, and a
word does not include the following white space.
```

icyhat:
> This is documented (under `:help WORD`):
> 
> ```
> Special case: "cw" and "cW" are treated like "ce" and "cE" if the cursor is
> on a non-blank.  This is because "cw" is interpreted as change-word, and a
> word does not include the following white space.
> ```

Thanks. You got a wide range of knowledge.

but since `cw` could treat as `ce`, then why the `dw` is not treated as `de` due to keeping consistency.

Or, can we exchange `w` and `e` as if `e` is actually end of the word with blanks while `w` is end of the word, because `w` is interpreted as word, and a word does not include the following white space.

chrisbra:
It's actually a strange vi behaviour:

https://github.com/vim/vim/blob/master/src/normal.c#L8648-L8656

Some time ago I made a patch (and I think Neovim merged that one: 
neovim/neovim#6235 and https://github.com/chrisbra/vim-mq-patches/blob/master/cpo_changeword

icyhat:
> It's actually a strange vi behaviour:
> 
> https://github.com/vim/vim/blob/master/src/normal.c#L8648-L8656
> 
> Some time ago I made a patch (and I think Neovim merged that one:
> [neovim/neovim#6235](https://github.com/neovim/neovim/pull/6235) and https://github.com/chrisbra/vim-mq-patches/blob/master/cpo_changeword

[...]

PR #15263, opened 2024-07-14
make d{motion} and dw work more consistent

chrisbra:
Problem:  d{motion} may work different than expected (also cw
          is special cased by POSIX)
Solution: add new cpo-z flag to make the behaviour configurable

There are two special vi compatible behaviours:

1): cw behaves differently than dw. That is, because cw is special cased
    by Vim and is effectively aliased to ce.

    POSIX behaviour:
https://pubs.opengroup.org/onlinepubs/9699919799/utilities/vi.html#tag_20_152_13_81

2): d{motion} may make the whole delete operation linewise, if the start
    and end of the motion are on different lines and there are only
    blanks before the start and after the end of the motion.
    Did not find a related POSIX link that requires this behaviour.

Both behaviours can be considered inconsistent, but we cannot easily change it, because it would be a backward incompatible change and also incompatible to how classic vi behaved.

So let's add the new cpo flag "w", which when not included fixes both behaviours and make them more consistent to what users would expect.

This has been requested several times:
https://groups.google.com/d/msg/vim_use/aaBqT6ECkA4/ALf4odKzEDgJ https://groups.google.com/d/msg/vim_dev/Dpn3xtUF16I/T6JcOPKN6usJ http://www.reddit.com/r/vim/comments/26nut8/why_does_cw_work_like_ce/ https://groups.google.com/d/msg/vim_use/vunNWLFWfQg/MmJh_ZGaAgAJ https://github.com/vim/vim/issues/4390

So in summary, if you want to have the w motion work more consistent, remove the 'z' from the cpo settings.

related: #4390

### DIFF
--- a/runtime/doc/change.txt
+++ b/runtime/doc/change.txt
@@ -1,4 +1,4 @@
-*change.txt*    For Vim version 9.1.  Last change: 2024 Jun 23
+*change.txt*    For Vim version 9.1.  Last change: 2024 Jul 14
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -95,13 +95,14 @@ For inserting text see |insert.txt|.
 These commands delete text.  You can repeat them with the `.` command
 (except `:d`) and undo them.  Use Visual mode to delete blocks of text.  See
 |registers| for an explanation of registers.
-
+							*d-special*
 An exception for the d{motion} command: If the motion is not linewise, the
 start and end of the motion are not in the same line, and there are only
 blanks before the start and there are no non-blanks after the end of the
 motion, the delete becomes linewise.  This means that the delete also removes
 the line of blanks that you might expect to remain. Use the |o_v| operator to
-force the motion to be characterwise.
+force the motion to be characterwise or remove the "z" flag from 'cpoptions'
+(see |cpo-z|) to disable this peculiarity.
 
 Trying to delete an empty region of text (e.g., "d0" in the first column)
 is an error when 'cpoptions' includes the 'E' flag.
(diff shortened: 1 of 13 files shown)


zeertzjq:
> So let's add the new cpo flag "w", which when not included fixes both

This should be "z" in the commit message now.

src/testdir/test_cpoptions.vim
```diff
@@ -912,4 +912,50 @@ func Test_cpo_dot()
   let &cpo = save_cpo
 endfunc
 
+" Test for the 'z' flag in 'cpo' (make cw and dw work similar and avoid
+" inconsistencies, see :h cpo-z)
+" modified)
```
zeertzjq: This line looks strange

runtime/doc/change.txt
```diff
@@ -95,13 +95,13 @@ For inserting text see |insert.txt|.
...
 |registers| for an explanation of registers.
-
+							*d-special*
 An exception for the d{motion} command: If the motion is not linewise, the
 start and end of the motion are not in the same line, and there are only
 blanks before the start and there are no non-blanks after the end of the
 motion, the delete becomes linewise.  This means that the delete also removes
 the line of blanks that you might expect to remain. Use the |o_v| operator to
-force the motion to be characterwise.
+force the motion to be characterwise or set |cpo-z|.
```
zeertzjq: `remove |cpo-z|`?

yegappan (commented):
Can you mention this change in the version9.txt file under the version-9.2 section?

Shane-XB-Qian:
BTW: welcome back~ @yegappan :smile:

--
shane.xb.qian

yegappan:
> BTW: welcome back~ @yegappan 😄
> […](#)
> -- shane.xb.qian

Thanks.  I was out on vacation.

chrisbra:
> Can you mention this change in the version9.txt file under the version-9.2 section?

Yes planned for when I merge it.

PR #15264, opened 2024-07-14
runtime(filetype): recommended indent options for go and gdscript

markmacode:
Updates the ftplugin files of to use the officially recommended indentation styles for Golang and GDScript.

## go
Go officially recommends using tabs over spaces, and their formatter enforces tabs for indents https://go.dev/doc/effective_go
> We use tabs for indentation and gofmt emits them by default. Use spaces only if you must.

## gdscript
Godot also officially recommended using tabs over spaces https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#formatting
> Use Tabs instead of spaces for indentation. (editor default)

*EDIT: (ftplugin maintainers have been contacted)*

### DIFF
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -25,7 +25,7 @@ setlocal foldexpr=GDScriptFoldLevel()
 
 if get(g:, 'gdscript_recommended_style', 1)
     setlocal noexpandtab tabstop=4 softtabstop=0 shiftwidth=0
-    let b:undo_ftplugin ..= ' | setlocal expandtab< tabstop< softtabstop< shiftwidth<'
+    b:undo_ftplugin ..= ' | setlocal expandtab< tabstop< softtabstop< shiftwidth<'
 endif
 
 
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -24,8 +24,8 @@ setlocal foldignore=
 setlocal foldexpr=GDScriptFoldLevel()
 
 if get(g:, 'gdscript_recommended_style', 1)
-    setlocal noexpandtab softtabstop=0 shiftwidth=0
-    let b:undo_ftplugin .= ' | setl et< sts< sw<'
+    setlocal noexpandtab tabstop=4 softtabstop=0 shiftwidth=0
+    let b:undo_ftplugin .= ' | setl et< ts< sts< sw<'
 endif
 
 
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -25,7 +25,7 @@ setlocal foldexpr=GDScriptFoldLevel()
 
 if get(g:, 'gdscript_recommended_style', 1)
     setlocal noexpandtab tabstop=4 softtabstop=0 shiftwidth=0
-    let b:undo_ftplugin .= ' | setl et< ts< sts< sw<'
+    let b:undo_ftplugin ..= ' | setlocal expandtab< tabstop< softtabstop< shiftwidth<'
 endif
 
 
--- a/runtime/ftplugin/gdscript.vim
+++ b/runtime/ftplugin/gdscript.vim
@@ -4,6 +4,7 @@ vim9script
 # Language: gdscript (Godot game engine scripting language)
 # Maintainer: Maxim Kim <habamax@gmail.com>
 # Website: https://github.com/habamax/vim-gdscript
+# Last Change: 2024 Jul 14
 
 if exists("b:did_ftplugin") | finish | endif
 
@@ -21,6 +22,11 @@ setlocal suffixesadd=.gd
 setlocal commentstring=#\ %s
 setlocal foldignore=
 setlocal foldexpr=GDScriptFoldLevel()
+
+if get(g:, 'gdscript_recommended_style', 1)
+    setlocal noexpandtab softtabstop=0 shiftwidth=0
+    let b:undo_ftplugin .= ' | setl et< sts< sw<'
+endif
 
 
 def GDScriptFoldLevel(): string
--- a/runtime/ftplugin/gomod.vim
+++ b/runtime/ftplugin/gomod.vim
@@ -8,7 +8,8 @@ if exists('b:did_ftplugin')
 endif
 let b:did_ftplugin = 1
 
+setlocal noexpandtab
 setlocal formatoptions-=t formatoptions-=c
 setlocal commentstring=//\ %s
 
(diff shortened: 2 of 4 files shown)


zeertzjq:
Did you contact the maintainers listed at the top of these files?

habamax:
I wouldn't do it for gdscript even though I did it in vim-godot. It is not makefile where you have to use tabs, so I would leave it up to the end user to setup what they prefer: tabs or spaces.

markmacode:
> I wouldn't do it for gdscript even though I did it in vim-godot. It is not makefile where you have to use tabs, so I would leave it up to the end user to setup what they prefer: tabs or spaces.

I agree that users should have their choice, they can still configure gdscript to use spaces. I'm following the precedence set by other ftplugin files such as python, zig and others.

https://github.com/vim/vim/blob/700cf8cfa1e926e2ba676203b3ad90c2c2083f1d/runtime/ftplugin/python.vim#L121-L124
https://github.com/vim/vim/blob/700cf8cfa1e926e2ba676203b3ad90c2c2083f1d/runtime/ftplugin/zig.vim#L25-L31

If the official recommendation of a language is either tabs or spaces, I think the default of an editor should be what they recommend,  or do you disagree @habamax?

For GDScript and Go, I believe that almost all users would end up using these exact settings in their configuration. I personally don't have strong opinions on tabs vs spaces or 2 vs 4 spaces, I use what the formatter defaults to or what my colleagues use -- which is why it is painful to be forced to worry about indentation settings when Vim doesn't do the expected path for me.

markmacode:
@dbarnett as you are the maintainer listed on `go.vim`, what are your thoughts on this?

habamax:
> If the official recommendation of a language is either tabs or spaces, I think the default of an editor should be what they recommend, or do you disagree @habamax?

Then it also should be `tabstop=4`. I don't mind making it "the way godot recommends it", I have `setlocal noet ts=4 sw=0` in my `after/ftplugin/gdscript.vim`. What I am not very fond of is a recommended style variable that needs to be documented separately.

markmacode:
> Then it also should be `tabstop=4`. I don't mind making it "the way godot recommends it", I have `setlocal noet ts=4 sw=0` in my `after/ftplugin/gdscript.vim`. What I am not very fond of is a recommended style variable that needs to be documented separately.

Good point, I can add `tabstop=4`.  As for the variable, I can remove that too. If I add those changes would you approve this PR (for `gdscript.vim`)? Or are there any other issues you have with these changes?

habamax:
> Good point, I can add `tabstop=4`. As for the variable, I can remove that too. If I add those changes would you approve this PR (for `gdscript.vim`)? Or are there any other issues you have with these changes?

We can keep the variable I think, that is my personal dislike :).

Could you additionally create PR to https://github.com/habamax/vim-gdscript so next time I update it here it wouldn't override this change?

runtime/ftplugin/gdscript.vim
```diff
@@ -22,6 +23,11 @@ setlocal commentstring=#\ %s
 setlocal foldignore=
 setlocal foldexpr=GDScriptFoldLevel()
 
+if get(g:, 'gdscript_recommended_style', 1)
+    setlocal noexpandtab tabstop=4 softtabstop=0 shiftwidth=0
+    let b:undo_ftplugin .= ' | setl et< ts< sts< sw<'
```
dkearns:
```suggestion
    let b:undo_ftplugin ..= ' | setlocal expandtab< tabstop< softtabstop< shiftwidth<'
```

...to make the file's style uniform.

runtime/ftplugin/gdscript.vim
```diff
@@ -25,7 +25,7 @@ setlocal foldexpr=GDScriptFoldLevel()
 
 if get(g:, 'gdscript_recommended_style', 1)
     setlocal noexpandtab tabstop=4 softtabstop=0 shiftwidth=0
-    let b:undo_ftplugin .= ' | setl et< ts< sts< sw<'
+    let b:undo_ftplugin ..= ' | setlocal expandtab< tabstop< softtabstop< shiftwidth<'
```
habamax: I have merged this to upstream, removing `let` as well.

runtime/ftplugin/gomod.vim
yu-yk, line 1: LGTM for adding this change as the go official tool is also using tab

markmacode:
@zeertzjq maintainer of `gdscript.vim` https://github.com/vim/vim/pull/15264#issuecomment-2228127820 and `gomod.vim` https://github.com/vim/vim/pull/15264#discussion_r1679041055 approve. Given that the `go.vim` file was last updated 10 years ago, I'm not sure how likely it will be to get a reply from them. What could be done if the maintainer can't be reached?

habamax:
That is why it is better to have separate PRs for different things :).

markmacode:
> That is why it is better to have separate PRs for different things :).

Haha yep I started to realise that separate PRs for this might have been the smart decision.

I'll give it a few days, but might have to remove the go.vim changes from this PR. Still curious what can be done to update ftplugin files if the maintainer can't be reached, since I will likely open up a separate PR for the go.vim change if that happens.

chrisbra:
>  Still curious what can be done to update ftplugin files if the maintainer can't be reached, since I will likely open up a separate PR for the go.vim change if that happens.

Since we seem to have approval, I'll go ahead and merge it and add a note for go runtime file then.

Limero:
Great change, thank you!
I see `tabstop=4` was added to `gdscript`, can the same be added for `go` as well? Or was there a reason behind not adding it?

markmacode:
> Great change, thank you! I see `tabstop=4` was added to `gdscript`, can the same be added for `go` as well? Or was there a reason behind not adding it?

The Godot editor set a precedence by having a tab size of 4 by default, so I set the options that mimic exactly what you would see in the Godot editor.

With Go there isn't much to suggest they have a preference for tab size, the [official go playground](https://go.dev/play/) has a tab size of 8. But the [official 'tour of go' tutorial](https://go.dev/tour/welcome/1) has a tab size of 4.

Limero:
> > Great change, thank you! I see `tabstop=4` was added to `gdscript`, can the same be added for `go` as well? Or was there a reason behind not adding it?
> 
> The Godot editor set a precedence by having a tab size of 4 by default, so I set the options that mimic exactly what you would see in the Godot editor.
> 
> With Go there isn't much to suggest they have a preference for tab size, the [official go playground](https://go.dev/play/) has a tab size of 8. But the [official 'tour of go' tutorial](https://go.dev/tour/welcome/1) has a tab size of 4.

Makes sense. Although I assume most people would prefer a tab size of 4 by default, we can't be sure as long as the playground has it set to 8.

PR #15266, opened 2024-07-15
runtime: mediawiki

avidseeker:
Imported from: https://github.com/chikamichi/mediawiki.vim

See also: http://en.wikipedia.org/wiki/Wikipedia:Text_editor_support#Vim


zeertzjq:
There is a comment in filetype.vim:
```vim
" FlexWiki - disabled, because it has side effects when a .wiki file
" is not actually FlexWiki
"au BufNewFile,BufRead *.wiki			setf flexwiki
```

This apparently also applies to MediaWiki, so it seems that the filetype detection shouldn't be added.

avidseeker:
The plugin and the Wikipedia page suggests these file detection patterns:

```vim
if has("autocmd")
  au BufRead,BufNewFile *.mw             set filetype=mediawiki
  au BufRead,BufNewFile *.wiki           set filetype=mediawiki
  au BufRead,BufNewFile *.mediawiki      set filetype=mediawiki
  au BufRead,BufNewFile *.wikipedia.org* set filetype=mediawiki
  au BufRead,BufNewFile *.wikibooks.org* set filetype=mediawiki
  au BufRead,BufNewFile *.wikimedia.org* set filetype=mediawiki
endif
```
I'm not sure how common these are, but I think we can go with `*.mediawiki` and `*.mw`?

avidseeker:
I'd still argue for autodetecting `*.wiki` though. It's the most popular format and it is the format that proved itself. Flexwiki last update was [10 years ago](https://sourceforge.net/projects/flexwiki/).

> it has side effects when a .wiki file is not actually FlexWiki

There is no Wiki syntax than FlexWiki except for Mediawiki.

zeertzjq:
Oh, where seems to be a discussion about this in https://groups.google.com/g/vim_dev/c/jFJ_xRzY5aI/

avidseeker:
The discussion is from 2010, is it still relevant after 14 years? It seems to be a problem with the flexwiki ftplugin (setting bomb), not with having .wiki as mediawiki file format. In fact, it has a supporting sentiment to my suggestion: 

> Also, why is "flexwiki" the handler, when wikipedia is probably the most widely used wiki? 

I propose the same question.

> I think Flexwiki is dead or nearly so. It seems like mapping .wiki files to wikipedia (MediaWiki?)
> rather than Flexwiki is far more useful. Feel free to remove the flexwiki mapping from 
> ftplugin.vim.

If Flexwiki is dead or nearly so in 2010, then it's surely dead now.

> I can't find another file format that uses the .wiki extension.
> -- Bram 

> It's common to use the '.wiki' for any wiki text file; so making both it and '.mw' load MediaWiki syntax makes sense.

runtime/filetype.vim
```diff
@@ -1430,6 +1426,10 @@ au BufNewFile,BufRead .mbsyncrc			setf conf
 " mcmeta
 au BufNewFile,BufRead *.mcmeta			setf json
 
+" MediaWiki
+au BufNewFile,BufRead *.wiki		setf mediawiki
+au BufNewFile,BufRead *.mw		setf mediawiki
```
jiangyinzuo: We can define the au command in just one line.
`au BufNewFile,BufRead *.mw,*.wiki setf mediawiki`

src/testdir/test_filetype.vim
```diff
@@ -437,6 +437,7 @@ def s:GetFilenameChecks(): dict<list<string>>
     matlab: ['file.m'],
     maxima: ['file.demo', 'file.dmt', 'file.dm1', 'file.dm2', 'file.dm3',
                 'file.wxm', 'maxima-init.mac'],
+    mediawiki: ['file.wiki'],
```
jiangyinzuo: We should add both 'file.wiki' and 'file.mw' to the test

jiangyinzuo (changes requested):

runtime/ftplugin/mediawiki.vim
```diff
@@ -0,0 +1,45 @@
...
+
+" Many MediaWiki wikis prefer line breaks only at the end of paragraphs
+" (like in a text processor), which results in long, wrapping lines.
+setlocal wrap linebreak
+setlocal textwidth=0
+
+setlocal formatoptions-=tc formatoptions+=l
+
+" utf-8 should be set if not already done globally
+setlocal fileencoding=utf-8
```
chrisbra: I think fileencoding should not be set in a ftplugin.

runtime/ftplugin/mediawiki.vim
```diff
@@ -0,0 +1,45 @@
...
+setlocal matchpairs+=<:>
+
+" Treat lists, indented text and tables as comment lines and continue with the
+" same formatting in the next line (i.e. insert the comment leader) when hitting
+" <CR> or using "o".
+setlocal comments=n:#,n:*,n:\:,s:{\|,m:\|,ex:\|}
+setlocal formatoptions+=roq
```
chrisbra: formatoptions is already set a few lines above. Please move to that place

runtime/ftplugin/mediawiki.vim
```diff
@@ -0,0 +1,45 @@
...
+    \ '<\@<=[ou]l\>[^>]*\%(>\|$\):<\@<=li\>:<\@<=/[ou]l>,' .
+    \ '<\@<=dl\>[^>]*\%(>\|$\):<\@<=d[td]\>:<\@<=/dl>,' .
+    \ '<\@<=\([^/][^ \t>]*\)[^>]*\%(>\|$\):<\@<=/\1>'
+endif
+
+" Enable folding based on ==sections==
+setlocal foldexpr=getline(v:lnum)=~'^\\(=\\+\\)[^=]\\+\\1\\(\\s*<!--.*-->\\)\\=\\s*$'?\">\".(len(matchstr(getline(v:lnum),'^=\\+'))-1):\"=\"
+setlocal fdm=expr
+
+let b:undo_ftplugin += "setl commentstring< comments< formatoptions< foldexpr< fdm< fileencoding<"
```
chrisbra: b:undo_ftplugin is not initialized, I think this errors out with `E121: undefined variable`.

runtime/ftplugin/mediawiki.vim
```diff
@@ -0,0 +1,45 @@
...
+let b:undo_ftplugin += " matchpairs< formatoptions< linebreak< wrap< textwidth<"
```
chrisbra: 'formatoptions is already reset i nthe line above

runtime/ftplugin/mediawiki.vim
chrisbra: also remove fileencoding here

runtime/syntax/mediawiki.vim
```diff
@@ -0,0 +1,292 @@
...
+
+hi def link htmlTag            Function
+hi def link htmlEndTag         Identifier
+hi def link htmlArg            Type
+hi def link htmlTagName        htmlStatement
+hi def link htmlSpecialTagName Exception
+hi def link htmlValue          String
+hi def link htmlSpecialChar    Special
+
+if !exists("html_no_rendering")
```
chrisbra, line 199: this variable is not documented

runtime/syntax/mediawiki.vim
```diff
@@ -0,0 +1,292 @@
...
+  hi def link htmlItalicBoldUnderline htmlBoldUnderlineItalic
+  hi def link htmlItalicUnderlineBold htmlBoldUnderlineItalic
+
+  hi def link htmlItalicBold          htmlBoldItalic
+  hi def link htmlItalicUnderline     htmlUnderlineItalic
+  hi def link htmlUnderlineBold       htmlBoldUnderline
+
+  hi def link htmlLink Underlined
+
+  if !exists("html_my_rendering")
```
chrisbra: and so is this one

runtime/ftplugin/mediawiki.vim
```diff
@@ -41,5 +37,5 @@ endif
 setlocal foldexpr=getline(v:lnum)=~'^\\(=\\+\\)[^=]\\+\\1\\(\\s*<!--.*-->\\)\\=\\s*$'?\">\".(len(matchstr(getline(v:lnum),'^=\\+'))-1):\"=\"
 setlocal fdm=expr
 
-let b:undo_ftplugin += "setl commentstring< comments< formatoptions< foldexpr< fdm< fileencoding<"
-let b:undo_ftplugin += " matchpairs< formatoptions< linebreak< wrap< textwidth<"
+let b:undo_ftplugin = "setl commentstring< comments< formatoptions< foldexpr< fdm<"
```
chrisbra: Sorry I just noticed. We reset `'commentstring'` option, but never set it. I can remove this while merging, or do we want to explicitly also set the 'commentstring' option for mediawiki?

avidseeker:
According to https://mediawiki.org/wiki/Help:Formatting, mediawiki uses the same HTML commentstrings

runtime/ftplugin/mediawiki.vim
```diff
@@ -0,0 +1,41 @@
...
+    \ '<\@<=[ou]l\>[^>]*\%(>\|$\):<\@<=li\>:<\@<=/[ou]l>,' .
+    \ '<\@<=dl\>[^>]*\%(>\|$\):<\@<=d[td]\>:<\@<=/dl>,' .
+    \ '<\@<=\([^/][^ \t>]*\)[^>]*\%(>\|$\):<\@<=/\1>'
+endif
+
+" Enable folding based on ==sections==
+setlocal foldexpr=getline(v:lnum)=~'^\\(=\\+\\)[^=]\\+\\1\\(\\s*<!--.*-->\\)\\=\\s*$'?\">\".(len(matchstr(getline(v:lnum),'^=\\+'))-1):\"=\"
+setlocal fdm=expr
+
+let b:undo_ftplugin = "setl commentstring< comments< formatoptions< foldexpr< fdm<"
```
dkearns:
```suggestion
let b:undo_ftplugin = "setl commentstring< comments< formatoptions< foldexpr< foldmethod<"
```

chrisbra:
thanks!

lilydjwg:
For Chinese content, one needs to put `setl formatoptions< linebreak<` in `after/ftplugin/mediawiki.vim` to fix wrapping (assuming global `fo` is tuned for Chinese).

PR #15267, opened 2024-07-15
Colorschemes update

habamax:
- Add PmenuMatch and PmenuMatchSel to all colorschemes
- Add contrast to habamax Type, String, Constant and PreProc
- Change habamax PmenuSel to neutral gray to make PmenuMatchSel more visible
- Change habamax Tabline and VertSplit
- Make Conceal less visible for zellner, torte, shine, ron, peachpuff, pablo, morning, koehler, evening, delek, blue, darkblue, lunaperche, retrobox
- Add Added/Changed/Removed highlights
- Fix retrobox Terminal background
- Other minor fixes and improvements


chrisbra:
thanks!

mmrwoods:
Hi @habamax,

This change seems to, well IMO anyway, negatively impact the syntax highlighting for git commits in the habamax color scheme, mostly due to setting `ctermfg=NONE` for  the Title highlight group. Without this columns that exceed the maximum recommend summary length are very hard to identify, see screenshots.

Also, purely a personal opinion, but FWIW I think the lack of a solid background on the separating vertically split windows is visually less clear, and seems inconsistent with the solid background used in the status line which separates horizontally split windows.

**before:**
habamax_gitcommit_before

**after:**
habamax_gitcommit_after

habamax:
I can agree with vertsplit, it doesn't look good with default `fillchars`, will revert it -- I can always override it locally for myself.

As for Title, that is a bit more difficult to reconsider. Title is used for a number of split windows in a tabline:
image

Let me play around the colors that could work here.

mmrwoods:
Nice, thanks :-)

mmrwoods:
@habamax Just noticed that setting `ctermfg=NONE` for Title also affects markdown headers. This will also affect html as `markdownH1` is linked to `htmlH1` which is linked to `Title`. I'm not sure if this was a desired effect, but FWIW here are some screenshots...

**before:**
Screenshot 2024-08-20 at 07 29 39

**after:**
Screenshot 2024-08-20 at 07 30 32

habamax:
Yeah I know, I would prefer yellowish Title as well, but then it plays really bad with Tabline, so there were 2 options:

- target all markup/down/html/tex filetypes and set their headings directly
- get used to just bold Title

for now I picked up the latter, hopefully in the future Title would not be used in Tabline so colorschemes can use it without making Tabline parts unreadable.

mmrwoods:
Fair enough, makes sense, tbh bold is probably good enough for markup headings, and git commit summary can be dealt with separately, so all good, thanks 👍

PR #15268, opened 2024-07-15
gdb file/folder check is now performed only in CWD.

ubaldot:
As per comment in https://github.com/vim/vim/pull/15261#issuecomment-2229032515

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -230,11 +230,11 @@ def SanityCheck(): bool
     err = 'Cannot debug, +channel feature is not supported'
   elseif (way is Way.Prompt) && !exists('*prompt_setprompt')
     err = 'Cannot debug, missing prompt buffer support'
-  elseif (way is Way.Prompt) && !empty(glob(gdb_cmd))
+  elseif (way is Way.Prompt) && !empty(glob($'./{gdb_cmd}')) || !empty(glob($'./{gdb_cmd}/'))
     err = $"You have a file/folder named '{gdb_cmd}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(asmbufname))
+  elseif !empty(glob($'./{asmbufname}')) || !empty(glob($'./{asmbufname}/'))
     err = $"You have a file/folder named '{asmbufname}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(varbufname))
+  elseif !empty(glob($'./{varbufname}')) || !empty(glob($'./{varbufname}/'))
     err = $"You have a file/folder named '{varbufname}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
   elseif !executable(gdb_cmd)
     err = $"Cannot execute debugger program '{gdb_cmd}'"
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -223,6 +223,12 @@ enddef
 
 def SanityCheck(): bool
   var gdb_cmd = GetCommand()[0]
+  var cwd =
+  var cwd = $'{getcwd()}/'
+  if exists('+shellslash') && !&shellslash
+    # on windows, need to handle backslash
+    cwd->substitute('\\', '/', 'g')
+  endif
   var is_check_ok = true
   # Need either the +terminal feature or +channel and the prompt buffer.
   # The terminal feature does not work with gdb on win32.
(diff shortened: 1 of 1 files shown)


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -230,11 +230,11 @@ def SanityCheck(): bool
...
   elseif (way is Way.Prompt) && !exists('*prompt_setprompt')
     err = 'Cannot debug, missing prompt buffer support'
-  elseif (way is Way.Prompt) && !empty(glob(gdb_cmd))
+  elseif (way is Way.Prompt) && !empty(glob($'./{gdb_cmd}')) || !empty(glob($'./{gdb_cmd}/'))
     err = $"You have a file/folder named '{gdb_cmd}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(asmbufname))
+  elseif !empty(glob($'./{asmbufname}')) || !empty(glob($'./{asmbufname}/'))
     err = $"You have a file/folder named '{asmbufname}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(varbufname))
+  elseif !empty(glob($'./{varbufname}')) || !empty(glob($'./{varbufname}/'))
```
chrisbra: I don't think we need duplicating the `glob()` argument. If there exists a file, then you cannot have a directory with the same name (at least not usually). Even if it was possible, I think adding the explicit path separator at the end would not make a match, in other words I think `glob('foobar')` would match the same directory as `glob('foobar/')` would

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -230,11 +230,11 @@ def SanityCheck(): bool
     err = 'Cannot debug, +channel feature is not supported'
   elseif (way is Way.Prompt) && !exists('*prompt_setprompt')
     err = 'Cannot debug, missing prompt buffer support'
-  elseif (way is Way.Prompt) && !empty(glob(gdb_cmd))
+  elseif (way is Way.Prompt) && !empty(glob($'./{gdb_cmd}')) || !empty(glob($'./{gdb_cmd}/'))
```
chrisbra: should we use `getcwd()` instead of `.` here?

ubaldot: Yes.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: I'll try it! :) Thanks for the feedback!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done!

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: Done!

chrisbra:
Hm thanks, I think we can further improve it. This makes sure we only call `getcwd()` once,
and handles the 'shellslash' option (relevant on Windows).
How about this:
```vim
diff --git a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
index 58d526d6a..57bb89e97 100644
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -224,19 +224,25 @@ enddef
 def SanityCheck(): bool
   var gdb_cmd = GetCommand()[0]
   var is_check_ok = true
+  # Add trailing slash, because we will concatenate a filename
+  var cwd = getcwd() .. '/'
+  if exists('+shellslash') && !&shellslash
+    # on windows, need to handle backslash
+    cwd->substitute('\\', '/', 'g')
+  endif
   # Need either the +terminal feature or +channel and the prompt buffer.
   # The terminal feature does not work with gdb on win32.
   if (way is Way.Prompt) && !has('channel')
     err = 'Cannot debug, +channel feature is not supported'
   elseif (way is Way.Prompt) && !exists('*prompt_setprompt')
     err = 'Cannot debug, missing prompt buffer support'
-  elseif (way is Way.Prompt) && !empty(glob(gdb_cmd))
+  elseif (way is Way.Prompt) && !empty(glob(cwd .. gdb_cmd))
     err = $"You have a file/folder named '{gdb_cmd}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(asmbufname))
+  elseif !empty(glob(cwd .. asmbufname))
     err = $"You have a file/folder named '{asmbufname}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !empty(glob(varbufname))
+  elseif !empty(glob(cwd .. varbufname))
     err = $"You have a file/folder named '{varbufname}' in the current directory Termdebug may not work properly. Please exit and rename such a file/folder."
-  elseif !executable(gdb_cmd)
+  elseif !executable(cwd .. gdb_cmd)
     err = $"Cannot execute debugger program '{gdb_cmd}'"
   endif
```

ubaldot:
Ok, I used the interpolation string  method instead of the ..

chrisbra:
thanks!

PR #15270, opened 2024-07-15
Added QuoteArgs to fix quoted/non_quoted args to gdb.

ubaldot:
As per https://github.com/vim/vim/pull/15261#issuecomment-2229032515

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1325,6 +1325,17 @@ def DeleteCommands()
   sign_undefine(BreakpointSigns->map("'debugBreakpoint' .. v:val"))
 enddef
 
+def QuoteArg(x: string): string
+  var x_escaped = x ->substitute('\\', '\\\\', 'g')
+    ->substitute('"', '\\"', 'g')
+
+  var x_list = split(x_escaped, '\zs')
+  if x_list[0] == '"' && x_list[-1] == '"'
+    return x_escaped
+  else
+    return $"\"{x_escaped}\""
+  endif
+enddef
 
 # :Until - Execute until past a specified position or current line
 def Until(at: string)
@@ -1335,7 +1346,7 @@ def Until(at: string)
     ch_log('assume that program is running after this command')
 
     # Use the fname:lnum format
-    var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at
+    var AT = empty(at) ? QuoteArg($"{expand('%:p')}:{line('.')}") : at
     SendCommand($'-exec-until {AT}')
   else
     ch_log('dropping command, program is running: exec-until')
@@ -1354,7 +1365,7 @@ def SetBreakpoint(at: string, tbreak=false)
   endif
 
   # Use the fname:lnum format, older gdb can't handle --source.
-  var AT = empty(at) ? $"\"{expand('%:p')}:{line('.')}\"" : at
+  var AT = empty(at) ? QuoteArg($"{expand('%:p')}:{line('.')}") : at
   var cmd = ''
   if tbreak
     cmd = $'-break-insert -t {AT}'
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1326,7 +1326,7 @@ def DeleteCommands()
 enddef
 
 def QuoteArg(x: string): string
-  # Find all the occurrences of " and \ and escape them and it double quote
+  # Find all the occurrences of " and \ and escape them and double quote
   # the resulting string.
   return printf('"%s"', x ->substitute('[\\"]', '\\&', 'g'))
 enddef
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1329,12 +1329,7 @@ def QuoteArg(x: string): string
   var x_escaped = x ->substitute('\\', '\\\\', 'g')
     ->substitute('"', '\\"', 'g')
 
-  var x_list = split(x_escaped, '\zs')
-  if x_list[0] == '"' && x_list[-1] == '"'
-    return x_escaped
-  else
-    return $"\"{x_escaped}\""
-  endif
+  return  x_escaped[0] == '"' && x_escaped[-1 : ] == '"' ? x_escaped : $'"{x_escaped}"'
 enddef
 
 # :Until - Execute until past a specified position or current line
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1326,10 +1326,9 @@ def DeleteCommands()
 enddef
 
 def QuoteArg(x: string): string
-  var x_escaped = x ->substitute('\\', '\\\\', 'g')
-    ->substitute('"', '\\"', 'g')
-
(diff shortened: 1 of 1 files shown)


runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1325,6 +1325,17 @@ def DeleteCommands()
...
+def QuoteArg(x: string): string
+  var x_escaped = x ->substitute('\\', '\\\\', 'g')
+    ->substitute('"', '\\"', 'g')
+
+  var x_list = split(x_escaped, '\zs')
+  if x_list[0] == '"' && x_list[-1] == '"'
+    return x_escaped
+  else
+    return $"\"{x_escaped}\""
+  endif
```
errael: I guess this is supposed to quote the string unless it is already quoted. According to the gdb/mi spec, the command parameter `x` does not have **any** `"` characters.  I think `$"..."` is better as `$'...'` since escape char is interpreted in "" (but I think it's actually OK in this case), or even simply `'"' .. x_escaped .. '"'. In any event, I think something like the following can replace all that `split()` ... stuff.
```
return x_escaped[0] == '"' ? x_escaped : $'"{x_escaped}"'
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1325,6 +1325,17 @@ def DeleteCommands()
   sign_undefine(BreakpointSigns->map("'debugBreakpoint' .. v:val"))
 enddef
 
+def QuoteArg(x: string): string
+  var x_escaped = x ->substitute('\\', '\\\\', 'g')
+    ->substitute('"', '\\"', 'g')
```
errael: Why the quote escape? There can be no quotes in parameter.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
LemonBoy: > According to the gdb/mi spec, the command parameter x does not have any " characters

You keep repeating that but I don't see any evidence in the docs. What's stated is that the string may be either unquoted or a C-string with double quotes, meaning that you can have escape sequences embedded (see [here](https://github.com/bminor/binutils-gdb/blob/a81f4e591fdb8530d832addc39beb31353b0ef2d/gdb/mi/mi-parse.c#L38)) and quotes (that, according to the source, can be escaped using backslashes.

ubaldot: I was not aware that in Vim you can index a string like a list. Though, it does not seem to work for negative indices (e.g. `x_escaped[-1]`). Is it enough to check check only the value at index `0`? Or shall I also check if the last character is a string as well? Because in the latter case I think the `split` would be needed and the following should be correct:

```
  var x_list = split(x_escaped, '\zs')
  return  x_list[0] == '"' && x_list[-1] == '"' ? x_escaped : $'"{x_escaped}"'
```

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
ubaldot: TBH I sticked to this request https://github.com/vim/vim/pull/15261#issuecomment-2229032515 However, reading at the point made by @LemonBoy above it looks like we need `->substitute('"', '\\"', 'g')` as well.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
chrisbra: you would need to use `x_escaped[-1:]`. There are some subtle differences between legacy Vim script and Vim9 script, but I think the new Vim9 script behaviour is more reasonable (you will always get characters, not bytes, which matters for multi-byte characters).

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
LemonBoy: You can condense that into a single `->substitute('[\\"]', '\\\0', 'g')` or something similar.

ubaldot: > You can condense that into a single `->substitute('[\\"]', '\\\0', 'g')` or something similar.

Sorry, just read that. I hope we can keep as it is in the last commit. Let me know.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: > > According to the gdb/mi spec, the command parameter x does not have any " characters
> 
> You keep repeating that but I don't see any evidence in the docs. What's stated is that the string may be either unquoted or a C-string with double quotes, meaning that you can have escape sequences embedded (see [here](https://github.com/bminor/binutils-gdb/blob/a81f4e591fdb8530d832addc39beb31353b0ef2d/gdb/mi/mi-parse.c#L38)) and quotes (that, according to the source, can be escaped using backslashes.

@LemonBoy Certainly the code is the ultimate spec. https://github.com/vim/vim/pull/15261#issuecomment-2229399949 is copied from the gdb/mi docs. Am I misunderstanding `c-string`?

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
```diff
@@ -1325,6 +1325,12 @@ def DeleteCommands()
...
+
+  return  x_escaped[0] == '"' && x_escaped[-1 : ] == '"' ? x_escaped : $'"{x_escaped}"'
```
errael: Is it possible for the first character to be a `"` and the last character to not be a `"`?

ubaldot: That is what I was wondering also, but given that I am not sure, I preferred the approach "better safe than sorry".

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: > Am I misunderstanding `c-string`?

Doh! This is a normal "C" quoted string? I think I read `non-blank-sequence` and spun out from that.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: > I preferred the approach "better safe than sorry".

I'm not sure how turning one mal-formatted string into a different mal-formatted string is "safe".

errael: I'm confused (obvious to all at this point). Some comments about input assumptions? In any event, the first character will never be a `"` in the ternary test; it would have been converted to `\"` by the `substitute`.

First looking at this code I had thought the idea was to also handle something already quoted; but I misunderstood.

runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
errael: > you would need to use `x_escaped[-1:]`

?
```
vim9script
var x = "abc"
echo x[0]
echo x[-1]
echo x[-1 : ]
```
outputs
```
a
c
c
```

chrisbra: well, seems another subtle difference between Vim9 script and legacy script.

chrisbra:
Thanks. I think we can simplify the code to this then:
```vim
def g:QuoteArg(x: string): string
  return printf('"%s"', x ->substitute('[\\"]', '\\&', 'g'))
enddef
```

errael:
> Thanks. I think we can simplify the code to this then:
> 
> ```viml
> def g:QuoteArg(x: string): string
>   return printf('"%s"', x ->substitute('[\\"]', '\\&', 'g'))
> enddef
> ```

The `g:` in the `def` statement is an artifact and not for inclusion?

chrisbra:
yes, sorry

chrisbra:
thanks!

PR #15273, opened 2024-07-16
Vim9: crash when accessing getregionpos() return value

### LINKED ISSUE #15272: segfault accessing list item coming from getregionpos in TextYankPost
### Steps to reproduce

1. `vim -Nu NONE`
2. copy following vimscript to the buffer
```
vim9script

def HighlightedYank(hlgroup = 'Pmenu', duration = 250)
    var type = v:event.regtype ?? 'v'
    var pos = getregionpos(getpos("'["), getpos("']"), {type: type})
    for p in pos
        echo p[0]
    endfor
enddef

augroup hlyank | au!
    autocmd TextYankPost * HighlightedYank()
augroup END

```
3. `:so`
4. `yy` any line

### Expected behaviour

It shoud not segfault.

[asciicast](https://asciinema.org/a/BWJZnIwUpPCCNupteMxSTrg7U)

### Version of Vim

9.1.589

### Environment

windows, debian

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Vim9: crash when accessing getregionpos() return value.
Solution: Correct the return type.

Fix #15272

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1,4 +1,4 @@
-*builtin.txt*	For Vim version 9.1.  Last change: 2024 Jul 11
+*builtin.txt*	For Vim version 9.1.  Last change: 2024 Jul 16
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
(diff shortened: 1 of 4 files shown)


habamax:
just tested, no crash, thx!

chrisbra:
thanks!

PR #15280, opened 2024-07-17
runtime(doc): minor updates.

shaneharper:


runtime/doc/builtin.txt
```diff
@@ -6947,10 +6947,10 @@ matchadd({group}, {pattern} [, {priority} [, {id} [, {dict}]]])
 matchaddpos({group}, {pos} [, {priority} [, {id} [, {dict}]]])
 		Same as |matchadd()|, but requires a list of positions {pos}
 		instead of a pattern. This command is faster than |matchadd()|
-		because it does not require to handle regular expressions and
-		sets buffer line boundaries to redraw screen. It is supposed
-		to be used when fast match additions and deletions are
-		required, for example to highlight matching parentheses.
+		because it does not handle regular expressions and it sets
```
habamax, line 6950: I would change `it sets` to just `sets` as before

runtime/doc/eval.txt
```diff
@@ -3124,9 +3124,10 @@ text...
 				let lconst[0] = 2  " Error!
 				let lconst[1][0] = 'b'  " OK
 <							*E995*
-			|:const| does not allow to for changing a variable: >
+			It is an error to specify an existing variable with
+			:const.
 				:let x = 1
-				:const x = 2  " Error!
+				:const x = 1  " Error!
```
habamax, line 3130: Why `1` here? `2` was fine to show that const with the same name is not allowed.

runtime/doc/various.txt
```diff
@@ -19,9 +19,8 @@ CTRL-L			Clear and redraw the screen.  The redraw may happen
 							*:redr* *:redraw*
 :redr[aw][!]		Redraw the screen right now.  When ! is included it is
 			cleared first.
-			Useful to update the screen halfway executing a script
-			or function.  Also when halfway a mapping and
-			'lazyredraw' is set.
+			:redraw can be used to redraw the screen in a script
```
habamax, line 22: Previous text looks better to me

shaneharper: What did you prefer about the original text?

Initially I intended just to replace the word "halfway" in the original text.  "...to update the screen _halfway_ executing a script..." doesn't make sense. (Replacing "halfway" with "while" helps: "...to update the screen _while_ executing a script...".)

habamax: I feel that `:redraw` is redundant as we are right under it already.

runtime/doc/builtin.txt
shaneharper: Without adding "it" `sets buffer line boundaries to redraw screen.` doesn't have a subject.

habamax: I might be reading original wrong, for me first "it" is a subject: it doesn't handle ... and sets ...

runtime/doc/eval.txt
shaneharper: I changed it to show that the error has nothing to do with the value that was used.

runtime/doc/builtin.txt
```diff
@@ -10334,8 +10334,8 @@ spellsuggest({word} [, {max} [, {capital}]])		*spellsuggest()*
 
 split({string} [, {pattern} [, {keepempty}]])			*split()*
 		Make a |List| out of {string}.  When {pattern} is omitted or
-		empty each white-separated sequence of characters becomes an
-		item.
+		empty each whitespace-separated sequence of characters becomes
```
dkearns, line 10337: "white-separated" is also used at `:help job_start` and `:help vimgrep`.

"white space", with the whitespace, is used over three hundred times in the help.

runtime/doc/various.txt
dkearns, line 22: "halfway through executing a script"?

runtime/doc/builtin.txt
Shane-XB-Qian: in IT languages, 'white' is not (only) equal to 'space'.

zzzyxwvut:
By the way, [`assert_fails`](https://github.com/vim/vim/blob/d24aaa9983619264cb5b4178924a352bb3a0eb37/runtime/doc/testing.txt#L564-L576) and [`assert_match`](https://github.com/vim/vim/blob/d24aaa9983619264cb5b4178924a352bb3a0eb37/runtime/doc/testing.txt#L639) _non-method_  
examples of `testing.txt` are also written in Vim9 syntax.

chrisbra:
thanks!

PR #15281, opened 2024-07-17
runtime(sdc): update syntax to SDC-standard 2.1

daniel-s-w:
Looking into the current standard for Synopsis Design Constraints (SDC) from their [Technology Access
Program](https://www.synopsys.com/community/interoperability-programs/tap-in.html), one can see that the current state of the sdc-syntax file is very outdated as well as short in coverage of keywords.

This commit pursues to add all the missing keywords from the current standard (Rev. 2.1).


chrisbra:
thanks!

PR #15282, opened 2024-07-17
Fix for the pot_plugpackage_path variable

### LINKED ISSUE #15275: Missing separator
https://github.com/vim/vim/blob/22105fd1fe0dcfe993b5c04c6ebe017a626116e3/src/po/Makefile#L320

RestorerZ:
Fix for the pot_plugpackage_path variable
Fix #15275

### DIFF
--- a/src/po/Makefile
+++ b/src/po/Makefile
@@ -314,7 +314,7 @@ $(LANGUAGES):
 #######
 
 # Preparing the POT file of the plug-in package
-POT_PLUGPACKAGE_PATH != pwd
+POT_PLUGPACKAGE_PATH = $(PWD)
 $(PLUGPACKAGE).pot: $(PO_PLUG_INPUTLIST)
 	$(VIMPROG) -u NONE --not-a-term -S tojavascript.vim \
 		$(PLUGPACKAGE).pot $?
@@ -327,7 +327,7 @@ $(PLUGPACKAGE).pot: $(PO_PLUG_INPUTLIST)
 	rm -f *.js ./vim_to_js
 
 # Converting the PO file of the plug-in package to the binary format of the MO
-MO_PLUGPACKAGE_PATH != pwd
+MO_PLUGPACKAGE_PATH = $(PWD)
 $(PLUGPACKAGE).mo: $(PO_PLUGPACKAGE)
 	$(MSGFMTCMD) -o $(MO_PLUGPACKAGE_PATH)/$@ $<
 


src/po/Makefile
```diff
@@ -313,8 +313,9 @@ $(LANGUAGES):
 # For translations of plug-ins
 #######
 
+POT_PLUGPACKAGE_PATH = ${PWD}
```
k-takata: Any reason to use `${}` instead of `$()`? (Both are the same, though.)

src/po/Makefile
```diff
@@ -327,7 +328,6 @@ $(PLUGPACKAGE).pot: $(PO_PLUG_INPUTLIST)
 	rm -f *.js ./vim_to_js
 
 # Converting the PO file of the plug-in package to the binary format of the MO
-MO_PLUGPACKAGE_PATH != pwd
 $(PLUGPACKAGE).mo: $(PO_PLUGPACKAGE)
 	$(MSGFMTCMD) -o $(MO_PLUGPACKAGE_PATH)/$@ $<
```
k-takata, line 332: Should be `POT_PLUGPACKAGE_PATH` instead of `MO_PLUGPACKAGE_PATH`?

chrisbra:
Wondering if this would be simpler to get rid of the `...PLUGPACKAGE_PATH` variable completly (I assume xgettext and msgfmt will use the current directory by default anyhow?):
```diff
diff --git a/src/po/Makefile b/src/po/Makefile
index 0f7f85dd1..caec2fe62 100644
--- a/src/po/Makefile
+++ b/src/po/Makefile
@@ -314,21 +314,18 @@ $(LANGUAGES):
 #######

 # Preparing the POT file of the plug-in package
-POT_PLUGPACKAGE_PATH != pwd
 $(PLUGPACKAGE).pot: $(PO_PLUG_INPUTLIST)
        $(VIMPROG) -u NONE --not-a-term -S tojavascript.vim \
                $(PLUGPACKAGE).pot $?
        $(XGETTEXT) --from-code=UTF-8 --default-domain=$(PLUGPACKAGE) \
                --package-name=$(PLUGPACKAGE) \
-               --output-dir=$(POT_PLUGPACKAGE_PATH) \
                --output=$(PLUGPACKAGE).pot --files-from=./vim_to_js
        $(VIMPROG) -u NONE --not-a-term -S fixfilenames.vim \
-               $(POT_PLUGPACKAGE_PATH)/$(PLUGPACKAGE).pot $?
+               $(PLUGPACKAGE).pot $?
        rm -f *.js ./vim_to_js

 # Converting the PO file of the plug-in package to the binary format of the MO
-MO_PLUGPACKAGE_PATH != pwd
 $(PLUGPACKAGE).mo: $(PO_PLUGPACKAGE)
-       $(MSGFMTCMD) -o $(MO_PLUGPACKAGE_PATH)/$@ $<
+       $(MSGFMTCMD) -o $@ $<

 # vim: set noet sw=8 ts=8 sts=0 wm=0 tw=0 ft=make:
```

src/po/Makefile
RestorerZ: As I understand it, the *BSD world prefers curly braces. And the GNU world also, in principle, uses them for variables.
But in this makefile, parentheses are used everywhere, so I fixed it. A uniform style must be observed.

src/po/Makefile
RestorerZ: Yes, I hurried, inattentive.

RestorerZ:
> Wondering if this would be simpler to get rid of the `...PLUGPACKAGE_PATH` variable completly (I assume xgettext and msgfmt will use the current directory by default anyhow?):

Probably, yes. I haven't tested it, but I think you're right.
I'll do it now.

RestorerZ:
Although not. We have declared the variable "MO_PLUGPACKAGE_PATH" and "POT_PLUGPACKAGE_PATH" in the documentation. This is done in case you need to place the result not in the "src/po" directory (which is for translations of the Vim editor itself), but where it is needed by the one who translates the scripts.
If you don't mind, I'll just fix what Takata‐san pointed out and send the push here.

chrisbra:
sure

chrisbra:
thanks!

PR #15283, opened 2024-07-17
Unnecessary redraw when setting 'winfixbuf'

zeertzjq:
Problem:  Unnecessary redraw when setting 'winfixbuf'.
Solution: Remove P_RWIN flag.


src/testdir/test_goto.vim
```diff
@@ -296,6 +296,7 @@ func Test_gd_string()
       return x;
     }
   [CODE]
+
```
zeertzjq, line 299: This blank line is present in every other test that uses this pattern, so adding it here.

chrisbra:
thanks!

PR #15288, opened 2024-07-18
Termdebug: still get E1023 when specifying arguments

zeertzjq:
Problem:  Termdebug: still get E1023 when specifying arguments and using
          a prompt buffer.
Solution: Use empty() instead of len().  Add a test.  Fix wrong order of
          arguments to assert_equal() in Test_termdebug_basic().


ubaldot:
Ops! I should have added a test actually ... :/ 
Many thanks for the fix @zeertzjq !

chrisbra:
thanks!

PR #15290, opened 2024-07-18
Wrong cursor position with 'breakindent' when wide char doesn't fit

### LINKED ISSUE #15289: The cursor position after wrapping with breakindent may be incorrect.
### Steps to reproduce

1. vim.exe -u NONE -N
2. :set encoding=utf-8
3. :set breakindent
4. input '\t'
5. Type 'a' until there is one character left before wrapping
6. input 'あ' (double width character)
7. switch normal mord

cursor position is not 'あ'.

vim-001

### Expected behaviour

cursor position is same 'あ'.

### Version of Vim

9.1.598-9.0.1826

### Environment

OS: Windows 11 Pro
vim: gvim.exe
vim version: 9.1.598-9.0.1826(more older vim-win-32-installer version is no reproduce this issue, maybe cause commit is  6e55e85f92aff43c1b3cb564201440f3552d63f0?)
See: [Release v9.0.1826 · vim/vim-win32-installer](https://github.com/vim/vim-win32-installer/releases/tag/v9.0.1826)

Linux is same reproduced.

OS: Ubuntu 24.04(WSL2)
Terminal: mintty
vim version: 9.1.16

### Logs and stack traces

```shell
no logs and no stack trace.
```

zeertzjq:
Problem:  Wrong cursor position with 'breakindent' when a double-width
          character doesn't fit in a screen line.
Solution: Include the width of the 'breakindent' properly.

Fix #15289


chrisbra:
thanks!

PR #15291, opened 2024-07-18
fix(completion): wrong exit condition on search fuzzy

### LINKED ISSUE #15287: Internal error with file completion with fuzzy
### Steps to reproduce

1. Started vim with vim -u NONE
2. Pressed ```C-X``` ```C-F```

Screenshot from 2024-07-18 11-11-49

### Expected behaviour

There should not be error

### Version of Vim

Latest vim compiled from source

### Environment

OS: Ubuntu 20

Terminal: Kitty

$TERM: tmux-256color

Shell: bash

### Logs and stack traces

_No response_

glepnir:
Problem:
1. in file completion leader can have the path separator.
2. Infinite loop when the col of startpos is not 0. Each time the loop reaches a new row, the col of currentpos will be reset to 0, making it impossible to exit the loop.

Solution:
1. when leader include path separator don't fuzzy result.
2. when back to startpos lnum and this line no match break loop avoid infinite loop

Fix #15287

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3538,7 +3538,8 @@ get_next_filename_completion(void)
     char_u	**sorted_matches;
     int		*fuzzy_indices_data;
 
-    if (in_fuzzy)
+    if (in_fuzzy && (vim_strchr(leader, '/') == NULL
+		&& vim_strchr(leader, '\\') == NULL))
     {
 	vim_free(compl_pattern);
 	compl_pattern = vim_strsave((char_u *)"*");
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3537,12 +3537,41 @@ get_next_filename_completion(void)
     int		in_fuzzy = ((get_cot_flags() & COT_FUZZY) != 0 && leader_len > 0);
     char_u	**sorted_matches;
     int		*fuzzy_indices_data;
+    char_u	*last_sep = NULL;
+    size_t	path_with_wildcard_len;
+    char_u	*path_with_wildcard;
 
-    if (in_fuzzy && (vim_strchr(leader, '/') == NULL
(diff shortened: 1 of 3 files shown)


src/insexpand.c
```diff
@@ -3538,7 +3538,8 @@ get_next_filename_completion(void)
     char_u	**sorted_matches;
     int		*fuzzy_indices_data;
 
-    if (in_fuzzy)
+    if (in_fuzzy && (vim_strchr(leader, '/') == NULL
```
LemonBoy: 1) Add a comment
2) I'd expect the fuzzy matching to be applied to the trailing part of the path.

src/search.c
```diff
@@ -5255,6 +5256,8 @@ search_for_fuzzy_match(
 			*pos = current_pos;
 			break;
 		    }
+		    else if (looped_around && current_pos.lnum == circly_end.lnum)
```
LemonBoy, line 5259: Isn't this caught in line 5237?

src/insexpand.c
glepnir: I am also thinking about point 2. So here we should get the final pattern part more carefully then do fuzzy

src/search.c
glepnir: Nope. like when startpos is lnum 1, col 5, in first running the ptr will from col 5 ` *ptr += current_pos.col;` we search from the col 5 . then loop back to lnum 1 currentpos.col is 0 because it will reset to 0 . then use ptr do fuzzy if this line nothing matched. should break. but in this case currentpos is lnum 1, col 0. so 5237 can't caught it in EQUAL_POS.  Is there anything that needs improvement?

chrisbra:
thanks!

PR #15297, opened 2024-07-18
runtime(vim): Update base-syntax, improve :map highlighting

dkearns:
Match :map ( RHS properly.

Only match ! after :map, :noremap, :unmap and :mapclear.


runtime/syntax/generator/gen_syntax_vim.vim
```diff
@@ -2,7 +2,7 @@
 " Language: Vim script
 " Maintainer: Hirohito Higashi (h_east)
 " URL: https://github.com/vim-jp/syntax-vim-ex
-" Last Change: 2024 Apr 07
+" Last Change: 2024 Jul 19
```
chrisbra, line 5: Having a time machine ;)

runtime/syntax/testdir/input/vim_ex_match.vim
```diff
@@ -11,3 +11,22 @@ match none
 3match FooGroup /Foo/
 3match
 3match none
+
+
+" Differentiate map() from :map
```
zeertzjq, line 16: Typo: it should be "match" here, not "map".

PR #15299, opened 2024-07-18
Unused function and unused error constants

dpelle:
Problem: unused function typval_compare_class() and error constants
Solution: remove function typval_compare_class() and ifdef out
          unused error constants


chrisbra:
thanks!

PR #15301, opened 2024-07-19
Allow popup_filter_yesno to handle "Press ENTER ...".

### LINKED ISSUE #15300: `vim` appears to hang with "Press ENTER or type command to continue"
### Steps to reproduce

There's a problematic interaction between `popup_dialog`/`popup_filter_yesno` and `Press ENTER ...`

Start vim with
```
gvim -u NONE -U NONE
```
then source
```
vim9script
echo 'one'
echo 'two'
popup_dialog('y/n?', {filter: 'popup_filter_yesno'})
```
See the message
```
```
Press ENTER and nothing happens. `Ctlr-C` or a very few other keys get out, but **the `popup_dialog` is never seen**.

### Expected behaviour

The most obvious is to handle `ENTER` and then display the popup.

### Version of Vim

9.1.586

### Environment

ubuntu/gtk3

### Logs and stack traces

_No response_

errael:
Fix #15300

Looking around I came across a simple fix. But key handling is complex. It doesn't seem like this could cause problems since it only kicks in when popup_filter_yesno is active.

I looked around for a solution from script, but it eludes me. For starters I couldn't find a way to tell if `Press ENTER ...` is active.

### DIFF
--- a/src/popupwin.c
+++ b/src/popupwin.c
@@ -2651,8 +2651,11 @@ f_popup_filter_yesno(typval_T *argvars, typval_T *rettv)
 	return;
 
     c = *key;
+    if (c == CAR && need_wait_return)
+	return;
     if (c == K_SPECIAL && key[1] != NUL)
 	c = TO_SPECIAL(key[1], key[2]);
+
 
     // consume all keys until done
     rettv->v_type = VAR_BOOL;


chrisbra:
yes thanks. I agree, key handling is rather complex, but this seems to be okay.

PR #15303, opened 2024-07-19
fix(filetype): use correct extension detection for Dracula

echasnovski:
Problem: pattern detection for Dracula language uses "\*lvs" and "\*lpe".
  As there is no dot, those are not treated as extensions which they
  should (judging by 'runtime/syntax/dracula.vim' and common sense).

Solution: use "\*.lvs" and "\*.lpe" patterns.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -713,7 +713,7 @@ if has("fname_case")
 endif
 
 " Dracula
-au BufNewFile,BufRead *.drac,*.drc,*lvs,*lpe	setf dracula
+au BufNewFile,BufRead *.drac,*.drc,*.lvs,*.lpe	setf dracula
 
 " Datascript
 au BufNewFile,BufRead *.ds			setf datascript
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

chrisbra:
It was hard to find some reference to this language. For future googlers, I found some specification documents from around 2000 when searching for Cadence DRC. However it seems this language has largely vanished and is no longer used. In addition, the `.drc` file extension seems to be now used by https://github.com/google/draco which is unfortunately a 3d data compiled binary file. And then there is the Dirac Video encoding that also uses `.drc`....

echasnovski:
> It was hard to find some reference to this language.

Yeah... That's why I was relieved to see the information in [syntax file](https://github.com/vim/vim/blob/d2cedc2db32d513343d162e319e644855dc2b255/runtime/syntax/dracula.vim#L5). Because detecting a niche filetype based only on file ending and not extension doesn't feel right.

PR #15308, opened 2024-07-19
Remove old statically-linked executable notice

alyptik:
The -g compiler flag might have resulted in a statically-linked executable 20 years ago (the date this was added to the documentation,) but nowadays that is no longer true.

Remove incorrect section.


chrisbra:
thanks makes sense.

PR #15311, opened 2024-07-20
runtime(mysql): update syntax script

jiangyinzuo:
Problem:

- `syn region ...`s in syntax/mysql.vim match function names inaccurately.
- no syntax rules for mysql window function.
- coarse highlight definition in syntax/mysql.vim.

Solution:

- add `\<...\>` to match the function name accurately.
- add syntax rules for mysql window function.
- enhance the highlight definition.

### DIFF
--- a/runtime/syntax/mysql.vim
+++ b/runtime/syntax/mysql.vim
@@ -1,9 +1,9 @@
 " Vim syntax file
 " Language:     mysql
 " Maintainer:   Kenneth J. Pronovici <pronovic@ieee.org>
-" Last Change:  $LastChangedDate: 2016-04-11 10:31:04 -0500 (Mon, 11 Apr 2016) $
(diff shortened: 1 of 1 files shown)


jiangyinzuo:
@pronovic could you please review the PR?

chrisbra:
thanks, but I think adding `\>` between the function end name and the closing parenthesis does't really make a difference.

jiangyinzuo:
> thanks, but I think adding `\>` between the function end name and the closing parenthesis does't really make a difference.

I have removed all the '\>'s in this PR.

chrisbra:
thanks!

pronovic:
> @pronovic could you please review the PR?

Hi @jiangyinzuo, I guess this is already merged, so you don't need my review.  As you can see from the history, this syntax file is pretty old, and I haven't personally updated it in a really long time &mdash; I don't use MySQL much any more.  So, I'm generally ok with any changes to bring it up-to-date.  Thanks for helping out.  :)

PR #14354, opened 2024-03-31
runtime(yaml): improve syntax highlighting for YAML

itchyny:
This PR fixes various known issues of YAML syntax highlighting.

- Recognize block scalar style to avoid unexpected highlighting by `yamlFlowString` (fix #11517)
- Improve performance of `yamlFlowMappingKey` by allowing execution by the NFA engine (fix #10730)
  - It was intentionally disabled before patterns were optimized by `s:SimplifyToAssumeAllPrintable`.
- Fix detection of flow style mapping indicators (fix #8234).
- Enable highlighting of explicit mapping value indicators and node properties in flow style.

PR #15312, opened 2024-07-20
Remove orphaned YAML dump files

itchyny:
These dump files are unintentionally committed to the repository in #14354. These files should be just removed.


chrisbra:
thanks!

PR #15313, opened 2024-07-20
runtime(yaml): Fix flow mapping key detection

### LINKED ISSUE #15196: YAML syntax issue: everything before the last colon in a line seems to be treated as a key
### Steps to reproduce

To open a yaml file with the example content, the `value:` seems to be highlighted:

```yaml
some_key: value:value
some_list:
- value:value
```

[](url)

### Expected behaviour

To not highlight values containing colons.

### Version of Vim

9.1

### Environment

OS: Manjaro Linux
Terminal: Konsole 24.05.1
Shell: bash 5.2.26

### Logs and stack traces

_No response_

itchyny:
This change fixes #15196.


chrisbra:
thanks!

PR #15314, opened 2024-07-21
feat: add extrahl in complete function

glepnir:
LSP spec mentioned completion item `deprecated` field when a completion item mark as deprecate the item should has a special highlight `strikethrough`.  helix also have this.  currently implement is add an extra field `extrahl` for combine. in this PR it works like `{word: "foo", kind: "func", menu:"menu", extrahl: "Strike"} hi Strike cterm=strikethrough`

QQ_1721548134338

Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem


glepnir:
how we test strikethrough in Ci..

src/insexpand.c
```diff
@@ -114,6 +114,7 @@ struct compl_S
     int		cp_flags;	// CP_ values
     int		cp_number;	// sequence number
     int		cp_score;	// fuzzy match score
+    int		cp_extrahlattr;	// extra highlight group attr
```
glepnir, line 117: extrahlattr a little long maybe need a better name here ?

chrisbra:
Hm, this seems to break the test dumps: `Test_wildmenu_pum_24.dump` and `Test_wildmenu_pum_32.dump` but only on 2 runners (I checked the dumps briefly, but couldn't make sense of the diff, the term content seemed to be fine, just some highlighting seemed to be changed).

That is surprising, since this change here doesn't even contain a test case yet :/

glepnir:
> this seems to break the test dumps: Test_wildmenu_pum_24.dump and Test_wildmenu_pum_32.dump but only on 2 runners (I checked the dumps briefly, but couldn't make sense of the diff, the term content seemed to be fine, just some highlighting seemed to be changed)

also wired too. because  `make test_cmdline` passed  every times when i run it 

> That is surprising, since this change here doesn't even contain a test case yet :/

Still trying to figure out how to test this strikethrough which requires a terminal. GitHub CI doesn't seem to be able to predict screen output like this. Going through our test cases also simply checks the highlight property.

glepnir:
now is passed. but seems like there is no way to test cterm/gui attribute in github ci..

glepnir:
@chrisbra  PTAL

runtime/doc/insert.txt
```diff
@@ -1183,6 +1183,9 @@ items:
 	user_data	custom data which is associated with the item and
 			available in |v:completed_item|; it can be any type;
 			defaults to an empty string
+	extrahl		allows specifying extra highlight group names to apply
```
zeertzjq: The "extra" name is a bit confusing as there are also `PmenuExtra` and `PmenuExtraSel` highlight groups, which are unrelated to this.

glepnir: makes sense. any good suggestions?

chrisbra: `hl_name` or `hl_group` ?

glepnir: purpose is to add cterm/gui attributes on the original basis.  meaning of hl_name hl_group is unclear

chrisbra: Well, should it should provide attributes or it should be a group name, with the applied highlighting attributes applied on top of it? The documentation above talks about `extra highlight group names to apply`, so `hl-group` sounds like a reasonable choice for a highlighting group, no?

glepnir: Currently it is the name of the highlight group. originally thought it was an attribute, but it is strange to add a highlight attribute in completion. Define a strikethrough highlight group in advance and pass it to item as an addition. hlgroup is also fine, but it is necessary to clearly indicate that the highlight group is merged on the original

chrisbra: That is fine. I don't remember that we need to specify additional highlighting attributes directly. for all kind of functions we assume a proper highlighting group must be defined beforehand and then only this highlighting group is used.  Please adjust the documentation, that the *specified highlighting group*  is combined together with `PmenuSel`/`Pmenu`. Would also be good which highlighting group has priority.

chrisbra:
thanks!

src/popupmenu.c
```diff
@@ -628,6 +631,8 @@ pum_redraw(void)
 	{
 	    hlf = hlfs[round];
 	    attr = highlight_attr[hlf];
+	    if (pum_array[idx].pum_extrahlattr > 0)
+		attr = hl_combine_attr(attr, pum_array[idx].pum_extrahlattr);
```
zeertzjq, line 635: The priority of this highlight group is inconsistent. Here it has higher priority than the default ones, but in `pum_screen_puts_with_attrs()` it has lower priority. Also, I wonder if the combination should be done in `pum_compute_text_attrs()` instead.

glepnir: Oops, forgot to change this. Because  can't test ceterm gui properties in github ci. So I adjusted the priority. Use a fake color simulation.

Shane-XB-Qian:
maybe just strike the label vs/not menu etc would be looking more nice?
@glepnir how do you think?
otherwise maybe a bit mess, since the menu or others' info maybe still wanted to be clear displayed.

glepnir:
word always aligns with the longest word. If you just enable hl_group on word, you will see some weird  behavior.

QQ_1723369800860

Shane-XB-Qian:
> word always aligns with the longest word. If you just enable hl_group on word, you will see some weird  behavior.

um.. seems that coz of you didnot give menu to those left all items, otherwise it maybe Not so "weird".

-- 
shane.xb.qian

glepnir:
I don't think there is anything wrong with the current behavior .

Shane-XB-Qian:
each strike should strike the label only and stop at the menu's left border.
image
which the 'method' should Not be striked.

Shane-XB-Qian:
> I suggest that create an issue as a trick

that is a relevant comment to this merged PR, maybe you can just open a PR to link to that/my comment.
// as your wish, i am ok to open a new issue ticket also, 
// but perhaps you knew in cn it is terrible network to play at github.
// so doing a 'comment' is more easy for me than open a new ticket.. :smile:

-- 
shane.xb.qian

Issue #174, opened 2015-08-18
Detect *.comp files as Mason


GoogleCodeExporter:
```
The attached patch makes vim assign the 'mason' filetype to files with the 
.comp extension, which denote Mason components.

```

Original issue reported on code.google.com by `hinrik....@gmail.com` on 24 Oct 2013 at 4:03

Attachments:
- [mason_filetype_comp.patch](https://storage.googleapis.com/google-code-attachments/vim/issue-174/comment-0/mason_filetype_comp.patch)

PR #15317, opened 2024-07-21
Add runtime files for GLSL

gpanders:
Upstream the GLSL syntax files from https://github.com/tikhomirov/vim-glsl. I've added myself as maintainer with a link to the upstream repo.

The only differences from upstream are:

1. **Removed setting `'formatoptions'` in the indent file.** If anything, these should go into a `ftplugin` file, not an `indent` file, but since these particular settings are subjective and not required for the filetype, I just removed it.
2. **Omitted the `after/syntax` files**. The upstream repo contains some `after/syntax` files to modify HTML and Elm filetypes to include inline GLSL. We _could_ modify the html and elm syntax files in Vim to include this, but I opted to leave those alone and just omit them (I'll leave this to Vim maintainers on how they want to address this).

There was a conflict in the `.comp` extension between GLSL and the "Mason" filetype. The `.comp` extension for mason was added in [August 2014](https://github.com/vim/vim/issues/174). There are not many details about that proposed change: the submitter said only that .comp files denote Mason components. However, the [upstream Mason documentation](https://metacpan.org/dist/Mason/view/lib/Mason/Manual/Components.pod) claims that component files use only `.mc`, `.mi`, and `.mp` extensions. I could not find _any_ search results indicating that the `.comp` extension is used for anything related to Mason (or indeed, for anything at all other than GLSL), so I think it is safe to remove this conflict in favor of GLSL.


chrisbra:
thanks!

PR #15318, opened 2024-07-21
runtime(netrw): Fix endless recursion in netrw#Explore()

### LINKED ISSUE #5723: :E with more than 1 arg is rising E132
:E is redundant with e when it is used with 1 argument but if you add more than one, it rise an E132 error.

**To Reproduce**
vim +'E /tmp BOOM'

**Expected behavior**
probably -nargs=? instead of -nargs=* should do the job but i haven't read the code that much (i'll do later if no one handle it)

vim 8.1.1401 provided by debian buster

ghost:
Problem:  ':E /etc BOOM' give E132 error.
Solution: Avoid recursion call with same arguments.

closes: #5723


chrisbra:
thanks!

PR #15321, opened 2024-07-22
runtime(zip-plugin): Fix extract file from zip browser

ghost:
Problem:  Enter 'x' in zip browser fail with E121
Solution: Fix typo in zip#Extract()

### DIFF
--- a/runtime/autoload/zip.vim
+++ b/runtime/autoload/zip.vim
@@ -395,8 +395,8 @@ fun! zip#Extract()
   endif
 
   " extract the file mentioned under the cursor
-"  call Decho("system(".g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(shell).")")
-  call system(g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(shell))
+"  call Decho("system(".g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(fname).")")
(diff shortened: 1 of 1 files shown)


runtime/autoload/zip.vim
```diff
@@ -395,8 +395,8 @@ fun! zip#Extract()
   endif
 
   " extract the file mentioned under the cursor
-"  call Decho("system(".g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(shell).")")
-  call system(g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(shell))
+"  call Decho("system(".g:zip_extractcmd." ".shellescape(b:zipfile)." ".shellescape(fname).")")
+  call system($"{g:zip_extractcmd} {shellescape(b:zipfile)} {shellescape(fname)}")
```
zzzyxwvut, line 399: The support for string interpolation is fairly recent
`v8.2.4883` (May 2022).  Substituting `fname` should be enough?

ghost: As this plugin is now maintained by the Vim project, I thought there was no problem.
Any opinions?

nickspoons: I thought the same, it is more readable to my eye with the string interpolation

chrisbra: should be fine to use string interpolation. We do not make any guarantees, that our runtime files need to be executable for older Vims anyhow. I would also just remove the commented out `Decho` call.

chrisbra:
thanks!

PR #15329, opened 2024-07-23
runtime(vim): Update base-syntax, improve :set highlighting

dkearns:
- Match bang, "all" and "termcap" options, and trailing command separator "|".
- Highlight set assignment operators.
- Match multiline :set and multiline option values.
- Mention the newer "0o" octal prefix at :help :set=.


chrisbra:
thanks!

zeertzjq:
Some cmdwin dumps files need to be updated after this patch.

chrisbra:
I am with you. It's a pain to use on Firefox. Also lately, the logs have gotten too long (because it prints the whole message strings for all languages), so that one always has to go to the plain text logs to see the errors.

I haven't investigated yet, why the message strings are output

PR #15336, opened 2024-07-24
Problem: exists(':keepalt') returns 0

### LINKED ISSUE #15305: Regression: exists(':keepalt') (and 5 more) return 0 instead of 2 (since Vim 9.1.352)
### Steps to reproduce

`exists(':keepalt')` returns `0` instead of `2` even though `:keepalt` (even on its own) is a valid built-in command.

### Expected behaviour

Scripts need a reliable way to query whether a command exists.

My [ingo-library plugin](http://www.vim.org/scripts/script.php?script_id=4433) uses a [a check](https://github.com/inkarkat/vim-ingo-library/blob/5321e59c5a2628b4567cb4fc2b629dbc60a7a30a/autoload/ingo/compat/commands.vim?plain=1#L27) to return an emulation of the `:keeppatterns` command for older Vim versions that don't have it. That emulation had scoping side-effects that triggered a bug in one of my plugins. So this was a very subtle problem ;-)

### Version of Vim

9.1.352 .. 9.1.602

### Environment

Vim on Ubuntu 20.04 LTS (shouldn't matter)

I've bisected the problem to https://github.com/vim/vim/commit/ed908f760d9fe933690745bac91aedb698bbd48f / v9.1.0352

A total of 6 commands (ignoring the `:{`, see below) are affected.

```
:echo filter(getcompletion('', 'command'), 'exists(":" . v:val) == 0')
['*', '++', '--', 'keepalt', 'keepjumps', 'keeppatterns', '{']
```
---

Furthermore, `:{` isn't an actual command, so it's a bit problematic that both `:{` and `:}` are in the list of completions, and `:echo exists(':}')` actually returns `2`, but `:}` gives `E1128: } without {: }` But that's a separate issue.

### Logs and stack traces

_No response_

chrisbra:
Problem:  exists(':keepalt') returns 0
Solution: use correct length when finding command,
          also ignore :{ and :} when expanding
          command line items

fixes: #15305

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -19,10 +19,6 @@ static int	ex_pressedreturn = FALSE;
 # define ex_hardcopy	ex_ni
 #endif
 
-#if defined(FEAT_EVAL) || defined(PROTO)
-static int cmp_cmdmod_info(const void *a, const void *b);
-#endif
-
 #ifdef FEAT_EVAL
 static char_u	*do_one_cmd(char_u **, int, cstack_T *, char_u *(*fgetline)(int, void *, int, getline_opt_T), void *cookie);
 #else
@@ -4050,16 +4046,6 @@ static cmdmod_info_T cmdmod_info_tab[] = {
     {"vim9cmd", 4, FALSE}
 };	// cmdmod_info_tab
 
-// compare two cmdmod_info_T structs by case sensitive name with length
-    static int
-cmp_cmdmod_info(const void *a, const void *b)
-{
-    cmdmod_info_T *cm1 = (cmdmod_info_T *)a;
-    cmdmod_info_T *cm2 = (cmdmod_info_T *)b;
-
-    return STRNCMP(cm1->name, cm2->name, cm2->minlen);
-}
-
 /*
  * Return length of a command modifier (including optional count).
  * Return zero when it's not a modifier.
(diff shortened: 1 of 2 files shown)


zeertzjq:
I don't think the length is the problem here. The problem here is that `cmdmod_info_tab` is sorted by the full string length, but it should be sorted by the part of the string that is actually compared. For `:keepmarks` it is `kee`, which should come before `:keepalt` (`keepa`) and `:keepjumps` (`keepj`).

zeertzjq:
> I don't think the length is the problem here. The problem here is that `cmdmod_info_tab` is sorted by the full string length, but it should be sorted by the part of the string that is actually compared. For `:keepmarks` it is `kee`, which should come before `:keepalt` (`keepa`) and `:keepjumps` (`keepj`).

Hmm, but changing the sorting doesn't solve the problem :(

inkarkat (approved):

chrisbra:
@zeertzjq you are right, that is the wrong way to fix it. I think what we need to do is to compare against the max length of both entries, so that we will always get an unambigious entry returned. And then also use `strlen(name)` for the target length. Otherwise we will go into the binary search with e.g. the entry 'keepalt' but because of the min length of 3 for 'keepmarks', the bsearch() returns the `:keepmarks` entry and then because of the following logic: https://github.com/vim/vim/blob/2979cfc2627d76a9c09cad46a1647dcd4aa73f5f/src/ex_docmd.c#L4127-L4135
it aborts and returns 0.

So like this:
```diff
diff --git a/src/ex_docmd.c b/src/ex_docmd.c
index e6200ed70..1caed8b08 100644
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -4057,7 +4057,7 @@ cmp_cmdmod_info(const void *a, const void *b)
     cmdmod_info_T *cm1 = (cmdmod_info_T *)a;
     cmdmod_info_T *cm2 = (cmdmod_info_T *)b;

-    return STRNCMP(cm1->name, cm2->name, cm2->minlen);
+    return STRNCMP(cm1->name, cm2->name, MAX(cm1->minlen, cm2->minlen));
 }

 /*
@@ -4079,7 +4079,7 @@ modifier_len(char_u *cmd)
        return 0;

     target.name = (char *)p;
-    target.minlen = 0;         // not used, see cmp_cmdmod_info()
+    target.minlen = STRLEN(p);
     target.has_count = FALSE;

     entry = (cmdmod_info_T *)bsearch(&target, &cmdmod_info_tab, ARRAY_LENGTH(cmdmod_info_tab), sizeof(cmdmod_info_tab[0]), cmp_cmdmod_info);
@@ -4119,7 +4119,7 @@ cmd_exists(char_u *name)
        cmdmod_info_T   *entry;

        target.name = (char *)name;
-       target.minlen = 0;              // not used, see cmp_cmdmod_info()
+       target.minlen = STRLEN(name);
        target.has_count = FALSE;

        // Check command modifiers.
@@ -5993,6 +5993,10 @@ get_command_name(expand_T *xp UNUSED, int idx)
 {
     if (idx >= (int)CMD_SIZE)
        return expand_user_command_name(idx);
+    // the following are no real commands
+    if (STRNCMP(cmdnames[idx].cmd_name, "{", 1) == 0 ||
+        STRNCMP(cmdnames[idx].cmd_name, "}", 1) == 0)
+       return (char_u *)"";
     return cmdnames[idx].cmd_name;
 }

```

chrisbra:
hm, still fails on Windows.

chrisbra:
I am going to partly revert patch v9.1.0352

PR #15340, opened 2024-07-25
Unnecessary STRLEN() in make_percent_swname()

zeertzjq:
Problem:  Unnecessary STRLEN() in make_percent_swname().
Solution: Pass the end of "dir" to make_percent_swname().


chrisbra:
thanks!

PR #15341, opened 2024-07-25
runtime(make): Fix highlight of Microsoft Makefile

k-takata:
Highlighting of variable expansion in Microsoft Makefile can be broken. E.g.:
https://github.com/vim/vim/blob/2979cfc2627d76a9c09cad46a1647dcd4aa73f5f/src/Make_mvc.mak#L1331

Don't use backslash as escape characters if `make_microsoft` is set. Also fix that `make_no_comments` was not considered if `make_microsoft` was set.

Add description for `make_microsoft` and `make_no_comments` in the document.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -532,6 +532,27 @@ export def FTm()
   endif
 enddef
 
+export def FTmake()
+  # Check if it is a Microsoft Makefile
+  if exists("b:make_microsoft")
+    unlet b:make_microsoft
+  endif
+  var n = 1
+  while n < 1000 && n <= line('$')
+    var line = getline(n)
+    if line =~? '^\s*!\s*\(ifn\=\(def\)\=\|include\|message\|error\)\>'
+      b:make_microsoft = 1
+      break
+    elseif line =~ '^ *ifn\=\(eq\|def\)\>' || line =~ '^ *[-s]\=include\s'
+      break
+    elseif line =~ '^ *\w\+\s*[!?:+]='
+      break
+    endif
+    n += 1
+  endwhile
+  setf make
+enddef
+
 export def FTmms()
   var n = 1
   while n < 20
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -2278,7 +2278,7 @@ By default mail.vim synchronises syntax to 100 lines before the first
 displayed line.  If you have a slow machine, and generally deal with emails
 with short headers, you can change this to a smaller value: >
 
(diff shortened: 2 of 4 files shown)


k-takata:
@rohieb Could you review this?

runtime/doc/syntax.txt
```diff
@@ -2278,7 +2278,7 @@ By default mail.vim synchronises syntax to 100 lines before the first
 displayed line.  If you have a slow machine, and generally deal with emails
 with short headers, you can change this to a smaller value: >
 
-    :let mail_minlines = 30
```
rohieb, line 2281: This doesn't seem related to the rest of the patch, was it intentional?

runtime/autoload/dist/ft.vim
```diff
@@ -532,6 +532,27 @@ export def FTm()
...
+  if exists("b:make_microsoft")
+    unlet b:make_microsoft
+  endif
+  var n = 1
+  while n < 1000 && n <= line('$')
+    var line = getline(n)
+    if line =~? '^\s*!\s*\(ifn\=\(def\)\=\|include\|message\|error\)\>'
+      b:make_microsoft = 1
+      break
+    elseif line =~ '^ *ifn\=\(eq\|def\)\>' || line =~ '^ *[-s]\=include\s'
```
rohieb, line 546: It's not clear to me yet why you need these two `elseif`s… Other than that, I really have no clue of Microsoft Make, but from what I can tell your patch looks fine.

runtime/doc/syntax.txt
k-takata: It's not related to the makefile fixes, but it was wrongly aligned.
I think we can include this small change as a part of the fix to syntax.txt.

runtime/autoload/dist/ft.vim
k-takata: This `elseif` checks GNU Make-style preprocessor directives and the `elseif` in line 548 checks GNU Make-style assignment operators.
These don't happen in Microsoft Make, so we can exit the while loop earlier.

chrisbra:
thanks both!

runtime/autoload/dist/ft.vim
```diff
@@ -532,6 +532,27 @@ export def FTm()
   endif
 enddef
 
+export def FTmake()
+  # Check if it is a Microsoft Makefile
+  if exists("b:make_microsoft")
+    unlet b:make_microsoft
```
chrisbra, line 538: I think we can just use `:unlet!` instead of the `exists()` part. I'll include it with this tiny change.

PR #15342, opened 2024-07-25
tests: Regenerate test_cmd{win,line} dump files for latest Vim syntax file changes (PR #15329)

dkearns:
Problem:  Vim syntax file changes to :set command highlighting render some test dump files invalid. See commit ddbb6fe.
Solution: Regenerate the affected dump files.


chrisbra:
thanks!

PR #15344, opened 2024-07-25
Cursor moves beyond first line of folded end of buffer

luukvbaal:
Problem:  Cursor moves beyond start of a folded range at the end of a buffer. 
Solution: Move cursor to start of fold when going beyond end of buffer.
          Check that cursor moved to detect FAIL in outer cursor function.

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -2849,7 +2849,6 @@ cursor_down_inner(win_T *wp, long n)
 	// count each sequence of folded lines as one logical line
 	while (n--)
 	{
-	    // Move to last line of fold, will fail if it's the end-of-file.
 	    if (hasFoldingWin(wp, lnum, NULL, &last, TRUE, NULL))
 		lnum = last + 1;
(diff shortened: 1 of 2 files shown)


src/edit.c
```diff
@@ -2849,7 +2855,6 @@ cursor_down_inner(win_T *wp, long n)
 	// count each sequence of folded lines as one logical line
 	while (n--)
 	{
-	    // Move to last line of fold, will fail if it's the end-of-file.
```
luukvbaal, line 2852: This regressed in #12480. With this context, I think this comment was moved erroneously and is out of place here. New comment above explains what is done such that the outer `cursor_down()` fails again in that case.

luukvbaal: Hmm, I think this needs to be handled before entering the inner function. So I moved it there and added additional test cases. (But forgot to update the commit description.)

chrisbra:
thanks!

PR #15345, opened 2024-07-25
feat: add initial filetype plugin for autohotkey

telemachus:


telemachus:
This is a very minimal ftplugin, and I don't really want to be the maintainer (though I will if that's what it takes). But after looking at `runtime/ftplugin`, there seem to be many that are minimal. Also a search on GitHub turns up several examples of `autohotkey.vim` that include these comment settings.

runtime/ftplugin/autohotkey.vim
```diff
@@ -0,0 +1,22 @@
...
+" Last Changed: 2024-07-25
+
+if exists("b:did_ftplugin")
+  finish
+endif
+let b:did_ftplugin = 1
+
+let s:cpo_save = &cpo
+set cpo&vim
+
```
dkearns:
```suggestion
```

The save/restore is only needed when using continuation lines which do not appear in this file.

runtime/ftplugin/autohotkey.vim
```diff
@@ -0,0 +1,22 @@
...
+setlocal comments=:;
+setlocal commentstring=;\ %s
+
+let b:undo_ftplugin = "setlocal comments< commentstring<"
+
+let &cpo = s:cpo_save
+unlet s:cpo_save
+
```
dkearns:
```suggestion
```

runtime/ftplugin/autohotkey.vim
```diff
@@ -0,0 +1,22 @@
+" Vim filetype plugin file
+" Language:     Autohotkey
+" Maintainer:   <Someone's name goes here>
+" Last Changed: 2024-07-25
```
dkearns:
```suggestion
" Last Change: 2024 Jul 25
```

runtime/ftplugin/autohotkey.vim
```diff
@@ -0,0 +1,22 @@
+" Vim filetype plugin file
+" Language:     Autohotkey
+" Maintainer:   <Someone's name goes here>
```
dkearns: The general rule has always been, I believe, that new runtime files need a maintainer.  Could you also add yourself to https://github.com/vim/vim/blob/master/.github/MAINTAINERS?  

Maybe the syntax file maintainer, @mmikeww, would be willing to take it on?

runtime/ftplugin/autohotkey.vim
```diff
@@ -0,0 +1,22 @@
+" Vim filetype plugin file
+" Language:     Autohotkey
```
dkearns:
```suggestion
" Language:     AutoHotkey
```

dkearns (changes requested):

runtime/ftplugin/autohotkey.vim
telemachus: Cool. Thanks for saving me from cargo-culting my copy-pasta. :)

runtime/ftplugin/autohotkey.vim
telemachus: I'll take it for now then, and add my name to MAINTAINERS.

telemachus:
One last question for @dkearns: I was inclined to add `setlocal formatoptions+=croq`, but I could not decide if that was too opinionated. Let me know if you think I should add it (and the corresponding undo).

mmikeww:
i'm the current maintainer of the ahk syntax highlighting (https://github.com/mmikeww/autohotkey.vim), although all i've really done is just add new keywords onto the older maintainers work. and its mostly been just for ahk v1. i didnt even know that a filetype plugin was necessary

telemachus:
> i didnt even know that a filetype plugin was necessary

@mmikeww I could be wrong, but I believe you normally need (and have) both a `syntax/something.vim` and a `ftplugin/something.vim`. (And usually also `indent/something.vim` and others.) These files do different jobs.

> that syntax file should already be included with default vim, and commenting should work

The syntax file is there, and that handles detection of comments for highlighting. But more is required for commenting to work fully. A `ftplugin/something.vim` file normally sets both [`comments`](https://vimhelp.org/options.txt.html#%27comments%27) and [`commentstring`](https://vimhelp.org/options.txt.html#%27commentstring%27) (as this one does). Without those being set, vim falls back to defaults, and those defaults apply (I think) to C files. If those *are* set, then vim can do various things automatically (like correctly apply the comment marker to the next line) and plugins can interact with comments through scripting. At the moment, none of that applies to AutoHotkey files. (That's sort of how I ended up here: someone [filed a bug in neovim](https://github.com/neovim/neovim/issues/29853) because a built-in commenting plugin doesn't work with AutoHotkey files.

tl;dr the syntax file is only part of what AutoHotkey needs; this file will help add more capabilities for AutoHotkey files.

mmikeww:
thanks @telemachus 
if you want to submit the files to my repo, thats fine. or if you want to submit them here too. or if you want to take over maintainership of both this and also the syntax file, thats also fine by me. whatever you think is best, let me know

telemachus:
> i use the tcomment plugin with normal vim and i havent had a problem using gcc and the likes in order to comment ahk files

@mmikeww If I start vim without this added filetype plugin and run `:set commentstring?`, the result is `commentstring=/* %s */`. That's a default derived from C. So with tcomment installed, commenting will work in the sense that it gives me C-style comments. E.g., "This is a test" becomes "/* This is a test */". That (sort of? maybe?) works because AutoHotkey allows for multiline comments with `/*...*/` too. But (strictly speaking), AutoKey says that multiline comments "can be used to comment out an entire section, *but only if the symbols appear at the beginning of a line* (excluding whitespace)." With vim's defaults, that's not what you get as a result. Instead, you get a single-line comment with `/*` at the start and `*/` at the end. When you use tcomment, what kind of comments do you get? Also, I thought that the primary way to comment in AutoHotkey is single-line comments with `; `. That does not seem to be supported for me unless this filetype plugin is added. 

@dkearns I'm okay to maintain this for now. Let me know if you want me to make any other changes.

dkearns (approved):
> @dkearns I'm okay to maintain this for now. Let me know if you want me to make any other changes.

Thanks.  LGTM.

telemachus:
> Thanks.  LGTM.

Great. I will leave out the format options, so this is ready. Thanks again for review and guidance.

mmikeww:
@telemachus using `gcc` does use `;` for ahk comments for me, so maybe i've made other changes to my config somewhere to get that to happen, not sure

chrisbra:
thanks all!

PR #15349, opened 2024-07-26
Fix some SEGV related to null_object.

### LINKED ISSUE #15338: [vim9class] REGRESSION: SEGV using a null_object
### Steps to reproduce

```
vim9script

def X()
    var x = null_object
    echom x
enddef
X()
```
Bisecting  indicates
https://github.com/vim/vim/commit/3164cf8f12f14b725b918e3170bb0a9085af8298
```
date:        Thu Mar 28 10:36:42 2024 +0100
summary:     patch 9.1.0219: Vim9: No enum support
```

### Expected behaviour

No exception, error message like
```
E1324: Using an Object as a String
```
Not all uses of null_object have this problem

### Version of Vim

9.1.611

### Environment

ubuntu gtk3

### Logs and stack traces

_No response_

errael:
Fix #15338 plus 3 more SEGV.

The #15338 part of the fix is by @chrisbra .

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -661,6 +661,44 @@ def Test_object_not_set()
     Func()
   END
   v9.CheckSourceFailure(lines, 'E1363: Incomplete type', 1)
+
+  # Reference a object variable through a null class object which is stored in a
+  # variable of type "any".
+  lines =<< trim END
+    vim9script
+
+    def Z()
+      var o: any = null_object
+      o.v = 4
+    enddef
+    Z()
+  END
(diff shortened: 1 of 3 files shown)


errael:
The failure for the test `# Do "echom" of a null object variable.` displays the wrong line number.
One of the changes, typval.c around line 269 `tv_get_bool_or_number_chk(` has no test.

yegappan (approved):

errael:
Updated PR to fix the wrong line number for `# Do "echom" of a null object variable`

chrisbra:
thanks!

errael:
One more update on the way. Found a way to test the last untested change.

errael:
OK. All the changes have a test that would previously SEGV.

PR #15352, opened 2024-07-26
win32: Rework win32 startup code

k-takata:
* Revise the code and reduce #ifdefs.
* For VIMDLL, stop using the default CRT startup code to reduce the file size.
  The file size becomes ~130 KB -> ~34 KB on MSVC.
* Update comments. Make them consistent between os_w32dll.c and os_w32exe.c.


chrisbra:
thanks!

PR #15353, opened 2024-07-26
fix(completion): wrong priority and wrong test

glepnir:
1. wrong test case fiedl name not changed..
 2. take advice from @zeertzjq adjust the priority and move to pum_compute_text_attrs


chrisbra:
thanks!

PR #15361, opened 2024-07-26
MinGW: Fix regressions in v9.1.0621 and v9.1.0622

k-takata:
* Fix build error when COVERAGE=yes.
* Fix if_lua with USE_GC_SECTIONS=yes.

Additionally, remove coverage files on "make clean".


chrisbra:
thanks!

PR #15363, opened 2024-07-27
win32,if_mzsch: Fix if_mzsch with VIMDLL

k-takata:
`scheme_register_tls_space()` doesn't support a thread-local variable in a DLL:
https://docs.racket-lang.org/inside/im_memoryalloc.html#%28cpp._scheme_register_tls_space%29

Define it inside (g)vim.exe and refer to it from the DLL.

This should be applied after #15361.


chrisbra:
Thanks. It looks like the first two commits were also in #15361 so I only include the last mzscheme related commit

PR #15365, opened 2024-07-27
fix(completion): clear matches when have a folder leader but no fuzzy matched

### LINKED ISSUE #15357: Unmatched candidates are displayed on i_CTRL-X_CTRL-F & completeopt+=fuzzy
### Steps to reproduce

1. `vim --clean`
2. `set completeopt=menuone,noselect,fuzzy`
3. insert path like `/qqq` that does not exist
4. type `<c-x><c-f>` to start files completion
5. files or directories don't have 'qqq' in their name, but they are displayed

fuzzy-path-completion

### Expected behaviour

Only candidates with a match (in the example above, files containing 'q' three times) are displayed.
If there are no matching files, nothing is displayed. 

### Version of Vim

9.1.618

### Environment

OS: macOS Sonoma 14.5
Terminal: wezterm 20240203-110809-5046fc22
value of $TERM: xterm-256color
Shell: zsh 5.9 (arm-apple-darwin23.0.0)

### Logs and stack traces

_No response_

glepnir:
Problem: when have a pattern follow path separator and in fuzzy. if pattern folder
                 not exit there still show all matches.

Solution: if leader not fuzzy matched clear the matches.

Fix #15357


chrisbra:
thanks

Issue #15355, opened 2024-07-26
Add superhtml and ziggy filetype 


EliSauder:
**Is your feature request about something that is currently impossible or hard to do? Please describe the problem.**
I guess technically it is currently impossible to do? I am working on adding superhtml support to the Neovim ecosystem (will be followed by ziggy), one of the things that is needed is adding the file type support.

**Describe the solution you'd like**
Ideally, the `superhtml` file type would be added using the `.shtml` file extension as described [here](https://github.com/kristoff-it/superhtml/issues/20#issuecomment-2248214586) by the creator of superhtml and then the `ziggy` file type would be added using the `.ziggy` and `.ziggy-schema` file extensions as described [here](https://ziggy-lang.io/).

**Additional context**
As these are simple changes, I am willing to do the leg work of the changes (though I am unfamiliar with vimscript atm). However, since `.shtml` is currently mapped and this is my first time being involved with the project I wanted to open an issue for discussion.

chrisbra:
`.shtml` is typically used for server-side includes, that would mean you would need to detect the filetype programatically by inspecting the file content. Check the function `FThtml()` in the file `runtime/autoload/dist/ft.vim` 

Please don't forget to add a test into testdir/test_filetype.vim file.

You can also have a look at the various commits from the past, that add new filetype protection, you can check from git by running `git log -v -p --grep 'filetype:'`

PR #15367, opened 2024-07-27
Add filetype support for SuperHTML and Ziggy

EliSauder:
This is a PR to address https://github.com/vim/vim/issues/15355 to add SuperHTML and Ziggy filetype support into vim.

Some things to note which may require further discussion.
1. SuperHTML is an overload of the .shtml file extension. As such I updated the FThtml() function as mentioned in https://github.com/vim/vim/issues/15355. I also added tests for both of the tags that indicate something is a SuperHTML file.
2. ziggy-schema's filetype is currently set to `ziggy_schema` to match with the language id that the lsp and tree-sitter setups rely on.

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -421,7 +421,6 @@ export def FThtml()
       setf htmlangular
       return
     endif
-
     # Check for XHTML
     if getline(n) =~ '\<DTD\s\+XHTML\s'
       setf xhtml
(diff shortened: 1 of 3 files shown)


chrisbra:
Thanks. I have have a question however. Looking at https://github.com/kristoff-it/superhtml it seems superhtml is a Language server. How does that come into play with a new filetype `superhtml`? And isn't `superhtml` a bit long filetype?

EliSauder:
> Thanks. I have have a question however. Looking at https://github.com/kristoff-it/superhtml it seems superhtml is a Language server. How does that come into play with a new filetype `superhtml`? And isn't `superhtml` a bit long filetype?

superhtml is both an lsp for html as well as a new templating language along with an lsp for that template language

EliSauder:
Also a question I have about the updates, the CONTRIBUTE.md file mentions adding yourself to the MAINTAINERS file. Do I just add a line like: `runtime/ftplugin/<ft>.vim`?

chrisbra:
I see thank you!

chrisbra:
We don't need yet a maintainer line, since you only added code for detecting the new file types, but did not include an additional runtime file. So that should be fine.

PR #15368, opened 2024-07-28
Refactor syntax.c to reduce the number of calls to STRLEN()

basilisk0315:
Also add missing prototypes for static functions.

### DIFF
--- a/src/syntax.c
+++ b/src/syntax.c
@@ -4445,7 +4445,8 @@ add_keyword(
 
     if (curwin->w_s->b_syn_ic)
     {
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks, looks okay. Is there a reason, you forward-declared all those functions? I think this is not really necessary (unless they are called before they are defined) and is probably the reason, why Bram did not declare them.

basilisk0315:
Hi Christian, I added the prototypes because I thought it might be good idea. ;-)

See the latest push.

chrisbra:
thanks!

PR #15371, opened 2024-07-28
Tweak some documents

h-east:


runtime/doc/version9.txt
```diff
@@ -41614,7 +41614,7 @@ Functions: ~
 Autocommands: ~
 
 |CursorMovedC|		after the cursor was moved in the comamnd-line
-|KeyInputPre|		process any Key event in any mode
+|KeyInputPre|		before the process any key event in any mode
```
k-takata, line 41617:
```suggestion
|KeyInputPre|		before processing any key event in any mode
```

h-east: Thanks! "Commit Suggestion" failed because the file size was too large, so I fixed it locally.

chrisbra:
thanks!

PR #15375, opened 2024-07-28
runtime(vim): Update base-syntax, improve :import highlighting

dkearns:
- Match `autoload` as a keyword in `:import `commands.
- Match an expression argument for the filename.


chrisbra:
thanks

skywind3000:
Error encountered in `vim.vim` after update to this PR:

图片

PR #15382, opened 2024-07-29
Allow `call({string_name}, args)` to work with imported name

### LINKED ISSUE #15381: [vim9script] `call({string_name}, args)` doesn't work with imported function
### Steps to reproduce

With the following two files in the same directory, do
```
vim -c ':source call_string.vim'
```
and observe the output
```
Error detected while processing command line..script /src/tools/vim/src/call_string.vim:
line    6:
E117: Unknown function: ImportedFoo
```
**call_string.vim**
```
vim9script
import './imported_call_string.vim' as i_imp
const ImportedFoo = i_imp.ImportedFoo

call('ImportedFoo', ['call as string through local importedFoo'])
```

**imported_call_string.vim**
```
vim9script

export def ImportedFoo(arg: any)
    echo 'In Imported Foo:' arg
enddef
```

### Expected behaviour

Expect the function to execute.

I'm putting together a PR, but it's not clear to me that it's the best (or even a good) fix. So this issue is opened in case that PR should be closed without merge.

### Version of Vim

9.1.621

### Environment

ubuntu gtk3

### Logs and stack traces

_No response_

errael:
Fix #15381

### DIFF
--- a/src/testdir/test_vim9_builtin.vim
+++ b/src/testdir/test_vim9_builtin.vim
@@ -497,6 +497,56 @@ def Test_call_call()
   v9.CheckSourceDefAndScriptFailure(['call("reverse", [2], [1])'], ['E1013: Argument 3: type mismatch, expected dict<any> but got list<number>', 'E1206: Dictionary required for argument 3'])
 enddef
 
+def Test_call_imports()
+  # Use call with an imported function
+  var lines =<< trim END
+    vim9script
+
+    export const foo = 'foo'
+
+    export def Imported()
+    enddef
+
+    var count: number
+    export def ImportedListArg(l: list<number>)
+      count += 1
+      l[0] += count
+    enddef
+  END
+  writefile(lines, 'Test_call_imports_importme', 'D')
+  lines =<< trim END
+    vim9script
+    import './Test_call_imports_importme' as i_imp
+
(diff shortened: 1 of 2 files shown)


errael:
Back to `Draft`

Examining the call stack
```
    f_call
    func_call
    call_func
    find_func
```
The location, `find_func`, for this PR's additional functionality looks good.

Only used in `vim9script`. Since this deals with `import` that seems right and avoids unexpected side-effects.

It's important to handle `call("i_import.FuncName", [])` so the programmer can avoid processing the import file until it's needed. So I'm looking at that now (so back to `Draft`). This new code already has some scaffolding to support this; BTW, it looks like `find_exported()` will play a part in this.

errael:
Update the PR to incorporate `call('i_imp.FuncName', [])` into `call()` handling.

Partion call import parsing
- userfunc.c - find_func_imported()
  determine, based on whether or not '.' is in the name, how to handle
- eval.c - handle_import_func()
  this handles the `'i_imp.FuncName'` specifically

errael:
There was a minor test failure yesterday in `foldexpr` loading a little early. Tracking that down and taking another look, has greatly simplified this PR.

chrisbra:
alright, thanks!

Issue #15376, opened 2024-07-28
Help understanding strange mapping problem in vim with kitty


jrmd54:
### Steps to reproduce

Hello,

I first reported the issue on [kitty's github](https://github.com/kovidgoyal/kitty/issues/7675) but it seems the issue is from vim's side.

I noticed that some maps were not working anymore in vim when ran through kitty, notably these 2 remaps:
```
" See :h i_CTRL-R_CTRL-P
imap <c-r> <c-r><c-p>
" NUL corresponds to ctrl-space
nnoremap <C-k><NUL> :echo 'bar'<CR>
```
                                                                                                                                                                                                                 
For example with this minimal .vimrc,
```
" .vimrc file
imap <c-r> <c-r><c-p>
nnoremap <C-k><NUL> :echo 'bar'<CR>
" Store 'foo' in register a
let @a='foo'
```

running the following commands:
```
$kitty -c NONE
$vim
<c-k><c-space>
i<c-r>a          
```      
should echo 'bar' in the cmdline and insert 'foo' in the opened vim buffer.
But while it works with e.g. gnome-terminal, in kitty nothing is echoed and only 'a' is inserted (as if the mapping was not properly interpreted).

By trial and error i found that adding the line
```
# ^R was obtained by pressing <c-v><c-r> while in insert mode in vim, and corresponds to the byte 0x12.
map ctrl+r send_key ^R
```
to kitty.conf fixed the issue with <c-r>, but i have no idea why it works and what was the problem in the first place.

Any idea about what causes this behavior?


Thank you very much for your help.


### Expected behaviour

The following maps should have the same behavior in kitty than in other terminals:
```
imap <c-r> <c-r><c-p>
nnoremap <C-k><NUL> :echo 'bar'<CR>
```

### Version of Vim

My vim version is 9.0, kitty is 0.35.2

### Environment

OS: Ubuntu 23.10 mantic
$TERM: xterm-kitty

### Logs and stack traces

_No response_

zeertzjq:
`<NUL>` does not necessary correspond to ctrl-space. It corresponds to the 0 byte. Is there any reason you are using `<NUL>` instead of `<C-Space>` in the mapping?

zeertzjq:
For `<C-k><NUL>`, `<NUL>` isn't equivalent to Ctrl-Space in kitty. Instead you can use `<C-Space>`. I'm not sure if this is a bug.

[...]

PR #15384, opened 2024-07-29
Check that mapping rhs starts with lhs fails if not simplified

zeertzjq:
Problem:  Check that mapping rhs starts with lhs doesn't work if lhs is
          not simplified.
Solution: Keep track of the mapblock containing the alternative lhs and
          also compare with it.

related: #15376


chrisbra:
thanks!

PR #15386, opened 2024-07-29
MinGW: Revise makefile conditions

k-takata:
Reduce the nesting level.
Directly check if the `-Wl,--entry` option is required or not.


chrisbra:
thanks!

PR #15388, opened 2024-07-29
ci.yml: improve macos package installation

pheiduck:
All brew install steps moved into one.

Install audio device for playing sound since some of macos-12 machines have no audio device installed. Is returning false, shouldn't be the case anymore.

libtool is skipped on macos-12 due already installed.

diffutils installed for macos-14

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,6 +330,10 @@ jobs:
       - name: Install packages
         if: matrix.features == 'huge'
         run: |
+          # Install lua required for features huge.
+          # Apple diff is broken. Use GNU diff instead. See #14032.
+          brew install lua diffutils
+          echo "LUA_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
           if [ $(uname -r | grep ^21 | cut -c1-2) == 21 ]; then
           # Install audio device (blackhole-2ch) for playing sound since some of macos-12 machines have no audio device installed.
             brew install blackhole-2ch
@@ -338,10 +342,6 @@ jobs:
           # Install libtool required for features huge.
             brew install libtool
           fi
-          # Install lua required for features huge.
-          # Apple diff is broken. Use GNU diff instead. See #14032.
-          brew install lua diffutils
-          echo "LUA_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
 
       - name: Grant microphone access for macos-14
         if: matrix.features == 'huge' && matrix.runner == 'macos-14'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -330,15 +330,12 @@ jobs:
...
-          brew install lua libtool
-          echo "LUA_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
-
-      - name: Install blackhole-2ch for macos-12
-        if: matrix.features == 'huge' && matrix.runner == 'macos-12'
-        run: |
-          # Install audio device for playing sound since some of macos-12 machines have no audio device installed.
-          if system_profiler -json SPAudioDataType | jq -er '.SPAudioDataType[]._items == []'; then
-            brew install blackhole-2ch
+          if [ $(uname -r | grep ^21 | cut -c1-2) ]; then
```
chrisbra: what is the output of this uname command chain? Can you please add a comment for when we need to install lua and blackhole-2ch?

.github/workflows/ci.yml
```diff
@@ -330,15 +330,12 @@ jobs:
...
+            brew install lua blackhole-2ch
+            echo "LUA_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
+          else
+            brew install lua libtool diffutils
```
chrisbra: Please keep the comment, why we need GNU diffutils

.github/workflows/ci.yml
pheiduck: > what is the output of this uname command chain?

oh well I missed something:
`if [ $(uname -r | grep ^21 | cut -c1-2) == 21 ]`
Identifies macos-12

> Can you please add a comment for when we need to install lua and blackhole-2ch?

I will add the comments back was thing about it too.

chrisbra:
thanks, looks good now.

chrisbra:
Hm no wait, don't we need `diffutils` always being installed? If I see that correctly, it is now only installed for non-macos 12 versions?

pheiduck:
> Hm no wait, don't we need `diffutils` always being installed? If I see that correctly, it is now only installed for non-macos 12 versions?

All suggestions are sort out, should be fine now.

chrisbra:
Hm, honestly I am not sure this is an improvement. 
If I compare this:
```
if [ $(uname -r | grep ^21 | cut -c1-2) == 21 ]; then
```
to 
```
 if: matrix.features == 'huge' && matrix.runner == 'macos-12'
```
I find the latter to be easier to understand. 

Also, even so we would now have a single `install packages` section I find it easy to understand if we split this up like it is currently:
- install packages
- Install audio device for playing sound 
- install packages for testing

So after all, I would just keep the current logic, sorry.

PR #15391, opened 2024-07-29
E1510 may happen when formatting a message for smsg()

zeertzjq:
Problem:  E1510 may happen when formatting a message (after 9.1.0181).
Solution: Only give E1510 when using typval.


chrisbra:
thanks!

PR #15394, opened 2024-07-29
Unnecessary STRLEN() when applying mapping

zeertzjq:
Problem:  Unnecessary STRLEN() when applying mapping.
Solution: Use m_keylen and vim_strnsave().


chrisbra:
thanks!

PR #15396, opened 2024-07-30
runtime(zip): Opening a remote zipfile don't work

ghost:
Problem:  Open zipfile from HTTP give an empty buffer.
 Solution: Ensure that the magic bytes check does not
                skip protocol processing.

Also use readblob().


chrisbra:
thanks!

PR #15397, opened 2024-07-30
Add myself as the ondir maintainer

jparise:
I missed this step when I originally submitted #14604.


chrisbra:
thanks, and so did I :)

PR #15398, opened 2024-07-30
Problem: curswant not set on gm in folded line

### LINKED ISSUE #11596: `gm` and `g^` do not remember column for vertical motion in folds

chrisbra:
Problem:  curswant not set on gm in folded line
Solution: in a folded line, call update_curswant_force()

fixes: #11596
closes: #11994

### DIFF
--- a/src/normal.c
+++ b/src/normal.c
@@ -5785,6 +5785,14 @@ nv_g_home_m_cmd(cmdarg_T *cap)
 	curwin->w_valid &= ~VALID_WCOL;
     }
     curwin->w_set_curswant = TRUE;
+#ifdef FEAT_FOLDING
+    if (hasAnyFolding(curwin))
+    {
+	validate_cheight();
+	if (curwin->w_cline_folded)
+	    update_curswant_force();
+    }
+#endif
     adjust_skipcol();
 }
 
--- a/src/testdir/test_curswant.vim
+++ b/src/testdir/test_curswant.vim
@@ -1,4 +1,7 @@
-" Tests for curswant not changing when setting an option
+" Tests for not changing curswant
+
+source check.vim
+source term_util.vim
 
 func Test_curswant()
   new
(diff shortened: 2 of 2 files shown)


src/normal.c
```diff
@@ -5782,6 +5782,10 @@ nv_g_home_m_cmd(cmdarg_T *cap)
 	curwin->w_valid &= ~VALID_WCOL;
     }
     curwin->w_set_curswant = TRUE;
+#ifdef FEAT_FOLDING
+    if (curwin->w_cline_folded)
```
zeertzjq: `validate_cheight()` (and possibly `hasAnyFolding()`) may need to be called before using this value.

chrisbra: yeah right

chrisbra: done

PR #15400, opened 2024-07-31
Update habamax colorscheme: change diff/search/todo colors

habamax:
- Make diff colors more accessible, Green for added, Red for deleted, Blue for Changed
- Change Search to blue to be visible with Diff colors
- Change Todo to bright magenta


chrisbra:
thanks!

Issue #15039, opened 2024-06-18
auto nohlsearch


habamax:
'hlsearch' is very useful but at the same time too intrusive.

Help topic states specifically that `:nohlsearch` doesn't work in autocommand (actually it also doesn't work in timer callbacks) so there is no obvious way to make it auto turned off.

There is vim-cool plugin by @romainl and the essense of it might be the following snippet I am currently using:

```
# turn off hlsearch after:
# - doing nothing for 'updatetime'
# - getting into insert mode
augroup auto_nohlsearch | au!
    set updatetime=2000
    noremap <Plug>(nohlsearch) <cmd>nohlsearch<cr>
    noremap! <expr> <Plug>(nohlsearch) execute('nohlsearch')[-1]
    au CursorHold * call feedkeys("\<Plug>(nohlsearch)", 'm')
    au InsertEnter * call feedkeys("\<Plug>(nohlsearch)", 'm')
augroup END
```

Should we add this example, even though it is "hackish", to the help topic right after `:h :nohlsearch` or maybe even a simple plugin in the same way `comment` was added?

[![asciicast](https://asciinema.org/a/vBRE7DeuLuxGtQIIaq2WWuxu9.svg)](https://asciinema.org/a/vBRE7DeuLuxGtQIIaq2WWuxu9)

romainl:
The doc is riddled with "hackish" examples like this so why not one more? I find the idea of adding a small plugin better, though. MacVim releases are lagging a bit so I didn't have a chance to play with your commenting plugin yet but it definitely looks like the way to go.

FWIW, [vim-cool](https://github.com/romainl/vim-cool) was a pretty successful attempt to solve *one* of my pet peeves with `hlsearch`: that it lingers on after it stopped being relevant. Whether it is through a built-in plugin, a snippet in the doc, a third-party plugin, or in the core, the problem can be considered solved at this point and, as such, doesn't deserve much attention.

But that still leaves the other one: that the highlighting affects all windows, for which there is no workaround AFAIK. If some effort can be made to improve `hlsearch`, I think that's where it would be most useful.

chrisbra:
There is an entry in todo.txt to make 'hlsearch' global/local option. I remember I spent some time several years ago looking into it (but at least back then it wasn't that simple, but I forgot details).

[...]

PR #15402, opened 2024-07-31
runtime(nohlsearch): add g:loaded_nohlsearch guard

habamax:
Relates to https://github.com/vim/vim/issues/15039#issuecomment-2260305256

### DIFF
--- a/runtime/pack/dist/opt/nohlsearch/plugin/nohlsearch.vim
+++ b/runtime/pack/dist/opt/nohlsearch/plugin/nohlsearch.vim
@@ -1,15 +1,16 @@
 " nohlsearch.vim: Auto turn off hlsearch
 " Last Change: 2024-06-19
 " Maintainer: Maxim Kim <habamax@gmail.com>
+"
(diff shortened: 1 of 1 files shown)


runtime/pack/dist/opt/nohlsearch/plugin/nohlsearch.vim
```diff
@@ -5,6 +5,12 @@
 " turn off hlsearch after:
 " - doing nothing for 'updatetime'
 " - getting into insert mode
```
zeertzjq, line 7: These comments seem to fit better just above the autocommands, not here.

zeertzjq: Actually that's not necessarily true, as they can serve as the description of the behavior of the file here.

habamax: well :)

chrisbra:
thanks!

PR #15407, opened 2024-07-31
ex: fix global insert of a newline

mohd-akram:
This fixes a corner case where the last command in a global command list is insert/append/change and the last line of that is an empty line, and the terminating period is not used (it is optional if the insert/append/change is the last command in a global command list since the end is unambiguous). The example is:

```
g/foo/i\
<newline>
```

### DIFF
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -3370,7 +3370,7 @@ ex_append(exarg_T *eap)
 	{
 	    // No getline() function, use the lines that follow. This ends
 	    // when there is no more.
-	    if (eap->nextcmd == NULL || *eap->nextcmd == NUL)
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

mohd-akram:
nvi has a bug in this (lichray/nvi2#135). For traditional vi (heirloom-ex-vi), it should work:

```
$ cat file
enum
$ printf '%s\n' 'g/enum/i\' '' '%p' | ./ex file
"file" 1 line, 5 characters

enum
```

Double-checked this with OpenIndiana (which has traditional SVR4 ex at `/usr/xpg4/bin/ex`) and same thing.

chrisbra:
Ah thanks, I leraned something today !

PR #15410, opened 2024-08-01
Wrong comment for "len" argument of call_simple_func()

zeertzjq:
Problem:  Wrong comment for "len" argument of call_simple_func().
Solution: Remove the "or -1 to use strlen()".  Also change its type to
          size_t to remove one cast.


chrisbra:
thanks!

PR #15417, opened 2024-08-02
RFC: runtime(netrw): honor `g:netrw_alt{o,v}` for `:{S,H,V}explore`

intelfx:
Make `:Sexplore` / `:Hexplore` / `:Vexplore` commands honor the user `&split{right,below}` settings (or netrw-specific `g:netrw_alt{o,v}`) instead of hardcoding a split direction. Similarly, update banged variants of the two latter commands to follow the inverted preference.

---

I'm not sure that this change would be welcome as it changes existing behavior (which some people are probably used to); but I value consistency so I decided to still propose it.


chrisbra:
thanks. I tend to agree.

PR #15418, opened 2024-08-02
fix(completion): file completion respect completeslash when fuzzy

### LINKED ISSUE #15392: completeopt+=fuzzy does not honor completeslash=slash (Windows)
### Steps to reproduce

1. set completeopt+=fuzzy
2. set completeslash=slash
3. Type C:/ and trigger file completion, the match list is empty (non fuzzy complete would give matches).
4. Type C:\ and trigger file completion, matches are found, but text is completed with foward slashes, making any follow-up completion impossible.

### Expected behaviour

Expecting fuzzy file completion to behave the same as non-fuzzy file completion when it comes to matching slashes (backward or forward).
Thanks a lot!

### Version of Vim

9.1.636

### Environment

Operating System: Windows

### Logs and stack traces

_No response_

glepnir:
Problem: fuzzy not respect completeslash in windows

Solution: replace  the path separator in leader according the value of completeslash

Fix #15392

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -3552,9 +3552,34 @@ get_next_filename_completion(void)
     size_t	path_with_wildcard_len;
     char_u	*path_with_wildcard;
 
+#ifdef BACKSLASH_IN_FILENAME
+    char pathsep = (curbuf->b_p_csl[0] == 's') ?
+	'/' : (curbuf->b_p_csl[0] == 'b') ? '\\' : PATHSEP;
+#else
+    char pathsep = PATHSEP;
+#endif
+
     if (in_fuzzy)
     {
-	last_sep = vim_strrchr(leader, PATHSEP);
+#ifdef BACKSLASH_IN_FILENAME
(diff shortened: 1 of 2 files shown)


src/insexpand.c
```diff
@@ -3552,9 +3552,34 @@ get_next_filename_completion(void)
...
+	    }
+	}
+	else if (curbuf->b_p_csl[0] == 'b')
+	{
+	    for (i = 0; i < leader_len; i++)
+	    {
+		if (leader[i] == '/')
+		    leader[i] = '\\';
+	    }
+	}
```
glepnir, line 3580: there can use temporary but hard to read 

```C
leader[i] = (curbuf->b_p_csl[0] == 's' && leader[i] == '\\') ? '/'
                        : (curbuf->b_p_csl[0] == 'b' && leader[i] == '/') ? '\\'
                        : leader[i];
```

chrisbra: Thanks. I prefer the more readable approach

PR #15422, opened 2024-08-03
MSVC: Revise CPU handling

k-takata:
* The variable `ASSEMBLY_ARCHITECTURE` was originally used to generate the manifest file. However, this was removed in v7.0.198.  Remove `ASSEMBLY_ARCHITECTURE` and use `CPU` instead.
* `/arch:AVX512` is available starting from Visual C++ 2017.
* ARM64 was not considered when validating CPUNR. Skip the validation for ARM64 for now.
* Remove old code for VC6/7 and Win9x.


chrisbra:
thanks!

PR #15423, opened 2024-08-03
runtime(habamax.vim): update Todo highlight

habamax:
Magenta background Todo is too bright and might interfere with DiffText. Make it less strong, without background, bold.


chrisbra:
Thanks. Can we keep the *Last Change* header to `YYYY Mon DD`? That is what we typically use in most other runtime files.

chrisbra:
Thanks. I'll merge it with the changed date format then.

habamax:
@lifepillar, do you think we can make last change format consistent across different OSes?

PR #15424, opened 2024-08-03
fix(completion): completion next word when fuzzy enable

### LINKED ISSUE #15412: Adding keyword completion not working with fuzzy completion
### Steps to reproduce

Deactivate fuzzy completion if needed `:set completeopt-=fuzzy`.
In insert mode, type in `alpha bravo charlie`, then `<enter>`, and on the new line type `al<c-n><c-x><c-n>`.
The first `<c-n>` completes `al` into `alpha`, then the `<c-x><c-n>` enters the _adding keyword completion_ mode, and inserts `bravo` (the only match). Another `<c-x><c-n>` would insert `charlie`.

If you try to do this with `set completopt+=fuzzy`, `alpha` is inserted on `<c-n>` but the chained completion is not triggered on `<c-x><c-n>`, and `bravo` is not inserted.

### Expected behaviour

See the **Steps to reproduce**.
Thanks again for bringing fuzzy completion to Vim!

### Version of Vim

9.1.636

### Environment

Tested on Windows (but would suspect the issue is the same on other OSes)

### Logs and stack traces

_No response_

glepnir:
Problem: when cursor after a word which already in buffer no completion is given or shown.

Solution: if word before cursor is matched return next word as completion result.

Fix #15412

### DIFF
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -5197,6 +5197,7 @@ ins_compl_start(void)
     if (line_invalid)
 	line = ml_get(curwin->w_cursor.lnum);
 
+    int in_fuzzy = get_cot_flags() & COT_FUZZY;
     if (compl_status_adding())
     {
 	edit_submode_pre = (char_u *)_(" Adding");
@@ -5212,6 +5213,11 @@ ins_compl_start(void)
 	    curbuf->b_p_com = old;
 	    compl_length = 0;
 	    compl_col = curwin->w_cursor.col;
+	}
+	else if (ctrl_x_mode_normal() && in_fuzzy)
+	{
+	    compl_startpos = curwin->w_cursor;
+	    compl_cont_status &= CONT_S_IPOS;
 	}
     }
     else
--- a/src/search.c
+++ b/src/search.c
@@ -5219,6 +5219,8 @@ search_for_fuzzy_match(
     pos_T	circly_end;
     int		found_new_match = FAIL;
     int		looped_around = FALSE;
+    char_u	*next_word_end = NULL;
+    char_u	*match_word = NULL;
 
     if (whole_line)
 	current_pos.lnum += dir;
@@ -5254,6 +5256,35 @@ search_for_fuzzy_match(
 		    found_new_match = fuzzy_match_str_in_line(ptr, pattern, len, &current_pos);
 		    if (found_new_match)
 		    {
+			if (ctrl_x_mode_normal())
+			{
+			    match_word = vim_strnsave(*ptr, *len);
+			    if (STRCMP(match_word, pattern) == 0)
+			    {
+				next_word_end = find_word_start(*ptr + *len);
+				if (*next_word_end != NUL && *next_word_end != NL)
+				{
+				    // Find end of the word.
+				    if (has_mbyte)
+					while (*next_word_end != NUL)
+					{
+					    int l = (*mb_ptr2len)(next_word_end);
+
+					    if (l < 2 && !vim_iswordc(*next_word_end))
+						break;
+					    next_word_end += l;
+					}
+				    else
+				       next_word_end = find_word_end(next_word_end);
+				}
+				else if (looped_around)
+				    found_new_match = FALSE;
+
+				*len = next_word_end - *ptr;
+				current_pos.col = *len;
+			    }
+			    vim_free(match_word);
+			}
 			*pos = current_pos;
 			break;
 		    }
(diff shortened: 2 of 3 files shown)


glepnir:
cc @egesip

egesip:
Thanks for the fix. It works but I have to press `<c-x><c-n>`  _twice_ to "chain" the completions.

Using the same `alpha bravo charlie` example line, I type `alp`, press `<c-n>`, I get `alpha` inserted (the only match), then I press `<c-x><c-n>`, I immediately get `bravo` inserted when fuzzy is off, while I get a menu with `alpha` and `alpha bravo` and nothing inserted when fuzzy is on (this is the first difference in behavior). 

From there, I then press `<c-n>` to get `bravo` inserted, I then press `<c-x><c-n>` again, it doesn't insert anything and says "hit end of paragraph" (whereas it immediately inserts `charlie` when fuzzy if off), I have to press `<c-x><c-n>` again and now `charlie` is inserted (second difference in behavior).

glepnir:
thanks for your feedback. i am really appreaciate . about first different i have update patch. about 2 i am not sure i understand. could you check it again. and if two no works  could you give a clear reproduce step 1. xx 2. xx thanks

egesip:
Thanks! The first difference is fixed with the new patch!
The second difference is still there, let me try to highlight it:
* case fuzzy off:
    1. Input the line `alpha bravo charlie` as usual.
    2. On the next line, type `alp`, then press `<c-n>`, vim inserts `alpha` and the message line says `-- Keyword completion (^N^P) The only match`.  
    3. Now press `<c-x><c-n>`, vim **inserts** `bravo`, and the message line says `-- Adding keyword completion (^N^P) The only match`.
    4. Now press  `<c-x><c-n>` again, vim **inserts** `charlie` and the message line says  `-- Adding keyword completion (^N^P) The only match`.
* case fuzzy on:
    1.  Input the line `alpha bravo charlie` as usual.
    2. On the next line, type `alp`, then press `<c-n>`, vim inserts `alpha` and  the message line says `-- Keyword completion (^N^P) The only match`.  (up to here, this is the same behavior as the case fuzzy off).
    3. Now press `<c-x><c-n>`, vim **does not insert** `bravo` and the message line says `-- Adding keyword completion (^N^P) Hit end of paragraph`.
    5. Press `<c-x><c-n>` again, now vim **inserts** `bravo` and the message line says `-- Adding keyword completion (^N^P) The only match`
    6. Press `<c-x><c-n>` again, vim **does not insert** `charlie`, the message line says `-- Adding keyword completion (^N^P) Hit end of paragraph`.
    7. Press  `<c-x><c-n>` again, now vim **inserts** `charlie`, etc..

glepnir:
I can't reproduce it works fine with me `alp` insert alpha and then <C-x><C-N> i got bravo where am i missing 

test_case

egesip:
You're right! I had to bissect my vimrc to find the culprit: `set complete-=i`.
Very strange...
So in summary, to reproduce the issue, you need:
```
set completeopt+=fuzzy
set complete-=i
```

If put `i` back in (`set complete+=i`), then it works fine.

glepnir:
emm i just use default complete value it does not include i

egesip:
Weird, on my machine, the help says `'complete' 'cpt' string (default: ".,w,b,u,t,i")` (i.e. it _does_ include `i`...)
But anyway, if you cannot reproduce it, perhaps it is a Windows specific issue?
The fact that adding/removing the scanning of include files triggers the bug on Windows and not on Linux smells like a bad memory access, but I've got no real clue here.

glepnir:
sry it's my mistake i mixed up the nvim and vim in my env. I have update patch for problem 2.

egesip:
Your latest patch fixes the issue!
Thanks!

glepnir:
aha good to know did you still get any problem on this patch . if not i will mark as ready for review.

egesip:
Less coding than usual for me this week, but so far so good!
Thank you!

glepnir:
no worry . I just solved the conflict but i want continue waiting for you test for a while . because i am not sure with other edge cases if have.

glepnir:
@egesip  hello do you still get any problem? could i mark as ready ?

egesip:
Hello @glepnir, sure, let's go!
Thank you.

chrisbra:
So is this ready to be included?

glepnir:
@egesip  confirmed it works fine so i think can be merge.

chrisbra:
thanks!

PR #15427, opened 2024-08-04
runtime(zip): escape '[' on Unix as well

### LINKED ISSUE #29977: bug(zip): `[Content_Type].xml` zip member produces empty buffer
### Problem

**Summary**
The "edit-files-within-a-zip-archive-without-extracting-first" feature of Neovim doesn't work for a member named `[Content_Types].xml`, possibly because of the square brackets in the name.

**Additional Context**
Neovim allows members of ("files" within) a Zip archive to be opened with a `netrw` type of interface. `:e file.zip` displays a list of the archive members, moving the cursor over one and pressing Enter opens it.

The `.docx`, `.pptx`, and `.xlsx` file formats are all Zip archives. A required member for these formats is `[Content_Types].xml` which maps other members to their MIME-type.

This behavior works fine on v0.9.1. I noticed the problem on 0.11.0-dev-482 and it's still there on 0.11.0-dev-573. 

### Steps to reproduce

- grab any DOCX, PPTX, or XLSX file, this link has one but any will do: https://dynamiclearningmaps.org/sites/default/files/documents/ELA_EEs/math_EE_overviewby_grade_EXTERNAL_July_2014.xlsx
- `:e math_EE_overviewby_grade_EXTERNAL_July_2014.xlsx`
- Move cursor onto `[Content_Types].xml` and press Enter
- Buffer is empty, but `unzip -l math_...xlsx` reports size `2793` (and this member is never empty for these file types).
- Opening any other member in the archive produces content.

### Expected behavior

Should show content, probably two lines, the first an XML declaration like `<?xml version="1.0"...>` and then all the XML content on a very long second line.

### Neovim version (nvim -v)

v0.11.0-dev-573+g58406ab9f-Homebrew

I don't have that installed

### Operating system/version

macOS 14.5

### Terminal name/version

wezterm 20240203-110809-5046fc22, but same on iTerm2 v3.5.3

xterm-256color

### Installation

brew install neovim --HEAD

zeertzjq:
Problem:  After 6f1cbfc9ab483a09877e153ad130164875c40b1d fnameescape()
          is no longer called on the name of the file to be extracted.
          However, while spaces indeed don't need to be escaped, unzip
          treats '[' as a wildcard character, so it need to be escaped.
Solution: Escape '[' on both MS-Windows and Unix.

From the docs it seems '*' and '?' also need escaping, but they seem to
actually work without escaping.

fixes: neovim/neovim#29977

### DIFF
--- a/runtime/autoload/zip.vim
+++ b/runtime/autoload/zip.vim
@@ -9,6 +9,7 @@
 " 2024 Jul 23 by Vim Project: fix 'x' command
 " 2024 Jul 24 by Vim Project: use delete() function
 " 2024 Jul 20 by Vim Project: fix opening remote zipfile
+" 2024 Aug 04 by Vim Project: escape '[' in name of file to be extracted
 " License:	Vim License  (see vim's :help license)
 " Copyright:	Copyright (C) 2005-2019 Charles E. Campbell {{{1
 "		Permission is hereby granted to use and distribute this code,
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, makes sense

PR #15428, opened 2024-08-04
MSVC: Revise v9.1.0657

k-takata:
Adjust indentation for readability.


chrisbra:
thanks, that makes it more readable

Issue #15326, opened 2024-07-22
Shift+Insert does not paste from the clipboard in Windows console


nickspoons:
### Steps to reproduce

In Windows 10 or 11, launch the console with e.g. `Win+R` -> `cmd.exe`
Start vim with `vim --clean`, enter insert mode, and hit `Shift+Insert`. Instead of inserting the contents of the clipboard, the mode switches to "replace" mode, and back to "insert" after another `Shift+Insert`, as if `Insert` was being typed without `Shift`.

### Expected behaviour

The contents of the clipboard (visible next to `"*` in `:registers` output) should be inserted.

### Version of Vim

9.0.1146 and later

### Environment

Tested in Windows10 and Windows11. This only occurs in the default windows console (`cmd.exe`), not powershell terminals or Windows Terminal.

After bisecting by installing Windows releases from https://github.com/vim/vim-win32-installer, this behaviour begins in 9.0.1148 and did not occur in 9.0.1140 (the previous release). This release includes 8 patches, the obvious culprit is [9.0.1146 MS-Windows: various special keys and modifiers are not mappable](https://github.com/vim/vim/commit/c8b204952f0452fe1276d30ae6a5d38a3f421b95), introduced in #11768

chrisbra:
Hm, I didn't notice this, because for me `cmd.exe` always starts in the Windows terminal

nickspoons:
Tagging @zewpo, as the author of #11768

zewpo:
ahh, just saw this, will have a look

PR #15430, opened 2024-08-04
mswin console - Shift+Insert to paste from clipboard

zewpo:
Resolves Issue #15326

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -1265,7 +1265,7 @@ decode_key_event(
 			    }
 			}
 			else if (pker->wVirtualKeyCode == VK_INSERT
-					&& (nModifs & SHIFT) != 0 
+					&& (nModifs & SHIFT) != 0
 					&& (nModifs & ~SHIFT) == 0)
 			{
 			    *pmodifiers = 0;
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -1264,6 +1264,13 @@ decode_key_event(
 				}
 			    }
 			}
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks, is it possible to add a test for this?

zewpo:
I'm not sure how to write a test script for it.  The test needs to perform a copy into windows clipboard from outside of vim, (I'm not sure how to achieve that outside of vim at the moment.)  And also to make sure that the test is actually using the old windows console-host program, (not the new windows terminal), in that it passes the input into vim to be decoded.   The new windows terminal seems to paste correctly by default.     The test harness functions that I created (last year) was pretty good for testing most low level ms-windows input events, but I'm not sure how to do the copy from outside vim bit yet.   There might be a way to emulate what we think the console is doing.  To be honest, this code is so fragile, with all sorts of conditions for old windows ansi implementation stuff, its got a lot of legacy to deal with.

chrisbra:
yeah, I wasn't sure if those low-level key events could be used for that. accessing the windows clipboard should be possible using something like `system('echo foobar |clip')` but I am not sure, if `clip.exe` is always available and also not sure how to test that Vim is running in the old cmd.com context. 

Okay, then let me merge this once the CI is "green". Thanks!

zewpo:
@chrisbra   fyi, I'm not Christian, I think that was a typo from a while ago.  I'm Christopher.

chrisbra:
Yes sorry, I should know. Typo on my side, since I have to type my name too often 🙈

PR #15434, opened 2024-08-05
Refactor of zip plugin

chrisbra:


runtime/autoload/zip.vim
```diff
@@ -83,29 +83,22 @@ fun! zip#Browse(zipfile)
...
-   if &verbose > 1
-    echoerr "the zip plugin is not available (your vim doesn't support fnameescape())"
-   endif
-   return
-  endif
   if !executable(g:zip_unzipcmd)
    redraw!
-   echohl Error | echo "***error*** (zip#Browse) unzip not available on your system"
-   let &report= repkeep
+   echohl Error | echomsg "***error*** (zip#Browse) unzip not available on your system"
```
k-takata: How about defining functions to show error or warning messages?

chrisbra: good idea.

runtime/autoload/zip.vim
```diff
@@ -83,29 +83,22 @@ fun! zip#Browse(zipfile)
...
+   call s:RestoreOpts(dict)
```
ghost: This could be a good use case of the :defer feature?

chrisbra: good idea. Have amended by a new commit

PR #15435, opened 2024-08-06
Wrong comment about return value of vim_copyfile()

zeertzjq:
Problem:  Wrong comment about return value of vim_copyfile().
Solution: Correct the comment.


zeertzjq:
I may have found another problem. I'll open a new PR later.

PR #15438, opened 2024-08-06
filecopy() may return wrong value when readlink() fails

zeertzjq:
Problem:  filecopy() may return wrong value when readlink() fails.
Solution: Set ret to -1 so that 0 is returned when readlink() fails.

I don't think it's possible to add a test for this, as I can't think of
a case where lstat() succeeds while readlink() fails.


chrisbra:
thanks!

PR #15441, opened 2024-08-07
[vim9class] During compilation, don't look for a return statement for an `abstract` method.

### LINKED ISSUE #15432: [vim9class] Illegal lookup with upcasting for instance methods
### Steps to reproduce

Upcasting from derived types to their base types is common  
practice: a) library authors don't know and shouldn't care  
about derived type hierarchies based on their API, b) users  
of libraries may follow suit and narrow their derived types  
to some common base API types.

Given the type hierarchy:

```
abstract class A
    abstract def ToString(): string

    def string(): string
        return 'A'
    enddef
endclass

class B extends A
    def ToString(): string
        return this.string()
    enddef

##  def string(): string
##      return 'B'
##  enddef
endclass

class C extends B
    def ToString(): string
        return this.string()
    enddef

    def ToLowerCasedString(): string
        return tolower(this.string())
    enddef

    def string(): string
        return 'C'
    enddef
endclass
```

let's explore two issues.

Here, a `B` instance is instantiated and upcast to `A` at its  
assignment, and from now on `B`'s implementation is accessed  
through `A`'s API.  And the same applies for a `C` instance.  
As can be seen, method dispatch is correctly performed.

```
const b_: A = B.new()
echo 'A' == b_.ToString()
echo 'A' == b_.string()

const c_: A = C.new()
echo 'C' == c_.ToString()
echo 'C' == c_.string()
echo 'c' == (<C>c_).ToLowerCasedString()
```

However, it appears that the upcast type is not enforced.

These should fail with `E1325` because `ToLowerCasedString()`  
is not defined in either `A` or `B` and `c_` was narrowed to  
`A`; but they don't:

```
echo '.... 1 ....'
echo 'c' == c_.ToLowerCasedString()
echo 'c' == (<A>c_).ToLowerCasedString()
echo 'c' == (<B>c_).ToLowerCasedString()
const ca: A = <A>c_
const cb: B = <B>c_
echo 'c' == ca.ToLowerCasedString()
echo 'c' == cb.ToLowerCasedString()
```

These correctly fail for `B`:

```
echo b_.ToLowerCasedString()
echo B.new().ToLowerCasedString()
```
Another upcasting issue involves `abstract` methods of upcast  
types.

These `C` declarations are accepted (both `ToString()` and  
`string()` are defined in `C`):

```
echo 'CCCC' == repeat([C.new()], 4)
    ->reduce((prefix: string, c: C) => prefix .. c.string(), '')

echo 'CCCC' == repeat([C.new()], 4)
    ->reduce((prefix: string, c: C) => prefix .. c.ToString(), '')
```

These `A` declarations are rejected (`E1027`) (`ToString()` is  
`abstract` in `A`):

```
echo '.... 2 ....'

echo 'CCCC' == repeat([C.new()], 4)
    ->reduce((prefix: string, a: A) => prefix .. a.ToString(), '')

def TestA(a: A)
    echo a.ToString()
enddef

TestA(C.new())
```

### Expected behaviour

1) Upcast instances should only have access to members of  
their cast type (users are free to downcast instances to  
their runtime types when necessary).

2) The `abstract` methods of an upcast type should not  
interfere in method dispatch whenever their implementations  
are referenced through this upcast type.

### Version of Vim

v9.1.0659

### Environment

### Logs and stack traces

_No response_

errael:
This fixes part of #15432. In particular
```
vim9script

abstract class A
    abstract def String(): string
endclass

class B extends A
    def String(): string
        return 'B'
    enddef
endclass
def F(o: A)
    assert_equal('B', o.String())
enddef
F(B.new())
```
generates
```
Error detected while compiling /tmp/d.vim[15]..function <SNR>46_F[1]..<SNR>46_A.String:
line    1:
E1027: Missing return statement
```
The error arises because the compiler tries to compile
```
abstract def String(): string
```
Since this is an abstract method declaration it has no body; there is no return statement. This PR treats the compilation of an abstract method the same as compiling an interface function; it exits `compile_def_function()` early. So for an abstract method, just like an interface method, the compiler never tries to compile the non-existent body (it never looks for a return statement).

**Note:** this fails the same in the `vim-9.1` release.

@dkearns Notice that this showcases the similarity between interface and abstract class, which has been the topic of much discussion.

@chrisbra @yegappan @LemonBoy @zzzyxwvut 
UPDATE: the updated PR, treating it like a interface function, seems to fit much better.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -6129,6 +6129,27 @@ def Test_abstract_method()
     assert_equal('foo', A.Foo())
   END
   v9.CheckSourceSuccess(lines)
+
+  # Invoke method returning a value through the abstract class. See #15432.
+  lines =<< trim END
+    vim9script
+
+    abstract class A
+        abstract def String(): string
+    endclass
+
+    class B extends A
+        def String(): string
+            return 'B'
+        enddef
+    endclass
+
+    def F(o: A)
+        assert_equal('B', o.String())
+    enddef
+    F(B.new())
+  END
+  v9.CheckSourceSuccess(lines)
 enddef
 
 " Test for calling a class method from a subclass
--- a/src/vim9compile.c
+++ b/src/vim9compile.c
@@ -4120,8 +4120,9 @@ compile_def_function(
     ufunc->uf_args_visible = ufunc->uf_args.ga_len;
 
     // Compiling a function in an interface is done to get the function type.
-    // No code is actually compiled.
-    if (ufunc->uf_class != NULL && IS_INTERFACE(ufunc->uf_class))
+    // No code is actually compiled. Same goes for an abstract method.
+    if ((ufunc->uf_class != NULL && IS_INTERFACE(ufunc->uf_class))
+	|| IS_ABSTRACT_METHOD(ufunc))
     {
 	ufunc->uf_def_status = UF_NOT_COMPILED;
 	ret = OK;


chrisbra:
Thanks. Can you update the comment above? Does this also prevent an abstract method to return anything? 

Are we all okay with this change? @yegappan @dkearns ?

Can you add another failing test for using return inside the abstract def:
```diff
diff --git a/src/testdir/test_vim9_class.vim b/src/testdir/test_vim9_class.vim
index 4d860685d..f8a899e8f 100644
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -6150,6 +6150,29 @@ def Test_abstract_method()
     F(B.new())
   END
   v9.CheckSourceSuccess(lines)
+
+  # Invoke abstract method returning a value does not compile
+  lines =<< trim END
+    vim9script
+
+    abstract class A
+        abstract def String(): string
+           return 'X'
+        enddef
+    endclass
+
+    class B extends A
+        def String(): string
+            return 'B'
+        enddef
+    endclass
+
+    def F(o: A)
+        assert_equal('B', o.String())
+    enddef
+    F(B.new())
+  END
+  v9.CheckScriptFailure(lines,  'E1318: Not a valid command in a class: return ''X''')
 enddef

 " Test for calling a class method from a subclass
```

zzzyxwvut:
Please don't merge it yet.  I have tried the offered patch  
and there are still issues pertaining to the _second_ problem  
it attempts to resolve (see the issue [thread](https://github.com/vim/vim/issues/15432)). I'm in the  
middle of summarising it at the moment.

zzzyxwvut:
1) [Original examples](https://github.com/vim/vim/issues/15432#issue-2447282658) work (with respect to the _second_  
problem).

2) Some [modified examples (1)](https://github.com/vim/vim/issues/15432#issuecomment-2269664509) fail as they should if we treat  
the _first_ problem (upcast types should only allow access to  
their members NOT instance type members) as a bug and try to  
consistently address it:

```
vim9script

abstract class A
    abstract def string(): string
endclass

abstract class B extends A
endclass

#### COPY class C FROM ORIGINAL EXAMPLES.

# E1325
var Concat: func(string, A): string = (prefix: string, a: A) =>
    prefix .. a.ToString()

# Huh?
def TestA(a: A)
    echo a.ToString()
enddef

TestA(C.new())
echo Concat('', C.new())
```
BUT, please can this check be moved into the compilation  
stage for `def` functions as with lambda expressions?

3) This [modified example (2)](https://github.com/vim/vim/issues/15432#issuecomment-2269664509) now cannot be compiled:

```
vim9script

interface I
    def string(): string
endinterface

abstract class A implements I
    abstract def string(): string
endclass

abstract class B extends A
    abstract def string(): string
endclass

#### COPY class C FROM ORIGINAL EXAMPLES.

def TestI(i: I)
    echo i.string()
enddef

# E1091?
TestI(C.new())
```

Some overlooked side effects?

errael:
> Thanks. Can you update the comment above? 

I'm not sure to what comment you are referring. I'll reword the OP; it does look confusing.

> Does this also prevent an abstract method to return anything?

Abstract methods are only declarations; they have no body. They never return anything.  They can be declared such that an implementation of the method must return something.
 
> Are we all okay with this change? @yegappan @dkearns ?
> 
> Can you add another failing test for using return inside the abstract def:

I can add the test; it tests that random things after an abstract method declaration fail. For example, this is also an error

```vim
vim9script

abstract class A
    abstract def F()
    enddef
endclass
```
which causes:
```
E1318: Not a valid command in a class: enddef
```
> 
> ```diff
> diff --git a/src/testdir/test_vim9_class.vim b/src/testdir/test_vim9_class.vim
> index 4d860685d..f8a899e8f 100644
> --- a/src/testdir/test_vim9_class.vim
> +++ b/src/testdir/test_vim9_class.vim
> @@ -6150,6 +6150,29 @@ def Test_abstract_method()
>      F(B.new())
>    END
>    v9.CheckSourceSuccess(lines)
> +
> +  # Invoke abstract method returning a value does not compile
> +  lines =<< trim END
> +    vim9script
> +
> +    abstract class A
> +        abstract def String(): string
> +           return 'X'
> +        enddef
> +    endclass
> +
> +    class B extends A
> +        def String(): string
> +            return 'B'
> +        enddef
> +    endclass
> +
> +    def F(o: A)
> +        assert_equal('B', o.String())
> +    enddef
> +    F(B.new())
> +  END
> +  v9.CheckScriptFailure(lines,  'E1318: Not a valid command in a class: return ''X''')
>  enddef
> 
>  " Test for calling a class method from a subclass
> ```

errael:
> Please don't merge it yet. I have tried the offered patch and there are still issues pertaining to the _second_ problem it attempts to resolve 

I've updated the OP to clarify that the intention of this PR is only to fix the "missing return statement" issue. I'll take a look at the other issues and see if they relate.

> (see the issue [thread](https://github.com/vim/vim/issues/15432)). I'm in the middle of summarising it at the moment.

zzzyxwvut:
> I'll take a look at the other issues and see if they relate.

> > (see the issue thread). I'm in the middle of summarising it at the moment.

Sigh. Sorry for the confusion, [the summary](https://github.com/vim/vim/pull/15441#issuecomment-2274241838) is in this  
thread, but the problem description (i.e. what the second  
problem is) is in the issue thread.

errael:
> 1. [Original examples](https://github.com/vim/vim/issues/15432#issue-2447282658) work (with respect to the _second_
>        problem).

Hopefully, the "missing return" is handled. Discussion/examples can go back to the original (and new super) issue.

> 2. Some [modified examples (1)](https://github.com/vim/vim/issues/15432#issuecomment-2269664509) fail as they should if we treat
>        the _first_ problem (upcast types should only allow access to
>        their members NOT instance type members) as a bug and try to
>        consistently address it:

Both with and without this patch I see. Which seems correct.
```
Error detected while compiling :source buffer=1[28]..function <lambda>3:
line    1:
E1325: Method "ToString" not found in class "A"
```

> 
> ~BUT, please can this check be moved into the compilation stage for `def` functions as with lambda expressions?~ [UPDATE: `defcompile TestA` should suffice here.]

I'm not sure what this is about; resolved?

>     3. This [modified example (2)](https://github.com/vim/vim/issues/15432#issuecomment-2269664509) now cannot be compiled:vim9script
> ...
> 
> Some overlooked side effects?

When I try it here, it compiles, runs and outputs `0`. (this is with the updated PR)

> 
> [UPDATE: Note that both examples from [another bug report](https://github.com/vim/vim/issues/15449#issue-2454456919) also cannot be compiled.]

I'll go over to the original issue now.

zzzyxwvut:
> > ~BUT, please can this check be moved into the compilation  
> > stage for `def` functions as with lambda expressions?~  
> > [UPDATE: `defcompile TestA` should suffice here.]
>
> I'm not sure what this is about; resolved?

(Look at 2. [above](https://github.com/vim/vim/pull/15441#issuecomment-2274241838) for context.)

Yes.  Initially, I had removed `TestA(C.new())` that followed  
the `TestA()` definition and had factored out `Concat` out of  
`reduce()`'s `{func}`; then _sourced_ the file and wondered  
how come `E1325` is only thrown for the lambda expression and 
not for `TestA()` (when the expression is commented out); had  
come here and written the now struck-out comment.  Then it  
dawned on me that no compilation took place for `TestA()`;  
hence the updated note.

I went over the related examples again with a newer build  
(`4be245a76`).

(Look at 3. [above](https://github.com/vim/vim/pull/15441#issuecomment-2274241838) for context.)

> When I try it here, it compiles, runs and outputs `0`. (this is with the updated PR)

I'm seeing it too, where does it come from?  Where did "C"
go?  Luckily, it turns out that all it takes to have "C" is  
to change the declaration of `C` from:

```
class C extends B
```
to:

```
class C extends B implements I
```

(Note that there is already `abstract class A implements I`.)

(It also works with the earlier `f958a72ab` build when the  
`implements` clause is added.)  It feels kind of redundant  
since you do need to implement all `abstract` methods in order  
to have a concrete type so it is implied that some supertype  
has the `implements` clause.  On the other hand, you do need  
to redeclare all `abstract` methods of an `interface` in an  
`abstract` class (or classes), so it is consistent and yet  
_now_ it becomes optional for the clause (`E1091` was helpful  
before but, unfortunately, I failed to understand its source  
among 3 `abstract` `string()`s).  Documentation entries under  
`E1349` and `E1389` and `:interface` do not tie in with the  
requirement for every concerned `class` to repeat `implements`  
for its hierarchy members.

Other than that the related examples work as expected.

In conclusion, the `implements` clause is needed with either  
version of the proposed patch, and I'd rather have it failed  
hard with `E1091` than look at some unexpected default value.  

Sorry for the false start yesterday.

errael:
> > When I try it here, it compiles, runs and outputs `0`. (this is with the updated PR)
> 
> I'm seeing it too, where does it come from? Where did "C" go?

There's definitely something screwy about abstract methods. #15449 may hold the key.

> Luckily, it turns out that all it takes to have "C" is to change the declaration of `C` from:
> 
> ```
> class C extends B
> ```
> 
> to:
> 
> ```
> class C extends B implements I
> ```
> 
> (Note that there is already `abstract class A implements I`.)

I agree that this is a bug. The implements should not be needed.

I hope that fixing #15449 will give more insight into these problems. 

Do you see any problems with this patch?

zzzyxwvut:
Other than that the earlier version (`f958a72ab`) of this  
patch happens to be less intrusive when it comes to failure  
and emits `E1091` rather than producing some out-of-thin-air  
value, all related examples work (after adding `implements`).

(See my previous comment, near the end).

errael:
> Other than that the earlier version (`f958a72ab`) of this patch happens to be less intrusive when it comes to failure and emits `E1091` rather than producing some out-of-thin-air value, all related examples work (after adding `implements`).
> 
> (See my previous comment, near the end).

Thx, now I get it... fixing this bug exposes another. And it turns out the first try HACK fix gave more info when the newly exposed bug is encountered.

errael:
Marking this PR `Draft` because of its relationship to #15449 and #15432; this PR affects the way those bugs manifest. But note that the bug in the OP for this PR fails in `vim-9.1`. This PR can stand as is, but it makes sense to consider it in the context of the other issues.

This PR doesn't directly have anything to do with overriding a concrete method with an abstract method.

chrisbra:
alright, yeah makes sense

zzzyxwvut:
@chrisbra, please reconsider the *version* of the patch for  
inclusion.

A few comments above I have shared [my observations](https://github.com/vim/vim/pull/15441#issuecomment-2276563799):

> Other than that the earlier version (`f958a72ab`) of this
> patch happens to be less intrusive when it comes to failure
> and emits `E1091` rather than producing some out-of-thin-air
> value, all related examples work (after adding `implements`).

The following example is problematic with `v9.1.0674`:

```
vim9script

interface I
    def string(): string
endinterface

abstract class A implements I
    abstract def string(): string
endclass

# TODO: Uncomment "implements I"!
class B extends A #implements I
    def string(): string
	return 'B'
    enddef
endclass

# TODO: Uncomment "implements I"!
class C extends B #implements I
    def string(): string
	return 'C'
    enddef
endclass

def TestI(i: I): string
    return i.string()
enddef

def ReportErr(id: string)
    const msg: string = v:errmsg
    if empty(msg)
	echo 'WILL REPORT ' .. id .. ' LATER, OK?'
    else
	echo msg
    endif
enddef


try
    echo 'B' == TestI(B.new())
    echo 'B.string() is string: ' .. string(type('') == type(TestI(B.new())))
catch /\<E1072:/
    # FIXME: v:errmsg is empty?
    ReportErr('E1072')
    echo 'B.string() is number: ' .. string(type(0) == type(TestI(B.new())))
endtry


try
    echo 'C' == TestI(C.new())
    echo 'C.string() is string: ' .. string(type('') == type(TestI(C.new())))
catch /\<E1072:/
    # FIXME: v:errmsg is empty?
    ReportErr('E1072')
    echo 'C.string() is number: ' .. string(type(0) == type(TestI(C.new())))
endtry


# FIXME: WRONG ASSIGNMENT TYPE!
const notStringC: number = TestI(C.new())

# FIXME: WRONG ASSIGNMENT TYPE!
const notStringB: number = TestI(B.new())
```

On the other hand, after reverting the included patch and  
applying [its previous version](https://github.com/vim/vim/commit/f958a72abd6df51789efdac23cfc5558847785d7), you are *forced* to repeat the  
`implements` clause for the first non-`abstract` derived `class`.
I find it a better compromise before #15484 will have been  
resolved.  Ping @errael.

Admittedly, neither version can fix wrong method dispatch  
when `implements` is repeated for `B` but not for `C` (you are  
not forced here), but it should be solved in another PR.

errael:
> @chrisbra, please reconsider the _version_ of the patch for inclusion.

I left it in `draft` but it got checked in, not sure how/why that happened. Oh well. 

But, IMHO, under **no circumstances** should a bad/wrong patch be merged when a correct patch is available. You can remove that patch, but a bad patch should not be included.

chrisbra:
I thought this was a fix for a clear bug? I can revert it, if this causes too much trouble...

zzzyxwvut:
The previous version (not HEAD) works for me. But @errael  
may have reservations about its readiness.

errael:
> I thought this was a fix for a clear bug? I can revert it, if this causes too much trouble...

It is a fix for a clear bug.

The problem that @zzzyxwvut has is that fixing this bug exposes another. And it turns out the first try HACK fix gave more info when the newly exposed bug is encountered.

errael:
> The previous version (not HEAD) works for me. But @errael may have reservations about its readiness.

Not "readiness", **appropriateness**. If it's not possible to fix #15484 completely, and a message/error/warning (hmm, are there warnings?) must be output, then that fix goes somewhere else, not with this patch.

errael:
> I thought this was a fix for a clear bug? I can revert it, if this causes too much trouble...

BTW, leaving something in `draft` state is the commonly accepted way on github to indicate "not to be checked in" for some reason or another. 

It's been apparent that this project has it's own set of rules, guessing at what they might be certainly keeps things interesting ;)

chrisbra:
I'd prefer, if we fix this incrementally, if the current patch exposes a different problem. But you are working on it, if you want me to rollback this fix here, I will do.

> It's been apparent that this project has it's own set of rules, guessing at what they might be certainly keeps things interesting ;)

Sorry, trying to do too many things at the same time, checking in bug fixes timely so we got feedback early. 🤷

zzzyxwvut:
Then let's revert it and let it be for now.

errael:
> I'd prefer, if we fix this incrementally, if the current patch exposes a different problem.

This patch is good as is. It doesn't attempt to fix every known bug...

> we got feedback early.

Yeah, I get that. But when it's in a draft state?

> Then let's revert it and let it be for now.

Why? Don't you already run with a custom vim that includes the earlier/experimental patch? I don't like seeing history getting mucked with.

chrisbra:
> But when it's in a draft state?

Missed that. I pull the PR directly using git command, which does not care whether a PR is in draft or not.

zzzyxwvut:
> > Then let's revert it and let it be for now.

> Why?

Err... because "Internal error" can be reproduced with it?

chrisbra:
I'll revert it.

errael:
@chrisbra What a mess. I had updated my local clone while this was merged. Cleanup didn't go well. I won't touch it anymore, do with it what you will...

chrisbra:
sorry, for causing issues, but I am done with it. I'll promise to be more careful

errael:
Nothing to apologize for; I was cleaning up and rushing and should know better than to be hasty when working with repositories. (needed to get the patch back into my working directory to finish some other things up and I screwed up) Cobbled things back together and finished up the unrelated #15495.

zzzyxwvut:
The unmerged patches ([v1](https://github.com/vim/vim/commit/f958a72abd6df51789efdac23cfc5558847785d7), [v2](https://github.com/vim/vim/commit/4be245a767cd2460924c727cf7c343d30eba7c7e)) are still available from this  
repository in case they are needed.

errael:
Opened #15512 so this issue specifically shows up in open issues.

PR #15443, opened 2024-08-07
win32: Restore main screen buffer

k-takata:
`vim.exe --cmd quit` quitted without switching back to the main screen buffer and the alternate screen buffer was kept shown.

Restore the main screen buffer even when termcap mode has not been started.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -3561,6 +3561,10 @@ mch_exit_c(int r)
     vtp_exit();
 
     stoptermcap();
+    // Switch back to main screen buffer.
+    if (use_alternate_screen_buffer)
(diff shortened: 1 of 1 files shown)


k-takata:
I've confirmed that both `vim.exe --cmd q` and `vim.exe -c q` restore the main screen buffer correctly.

chrisbra:
thanks. I had the feeling, this was a somewhat recent regression, but I couldn't find a commit that touched this part of the code recently. Oh well 🤷

k-takata:
v9.0.1235 or v9.0.0868 might be related. This only happens when quitting in the initialization stage. So, no one might not notice this...

I noticed this when I tried to run the tests in runtime/indent.

PR #15446, opened 2024-08-07
runtime: Add indent for Protobuf

Limero:
This adds the indent file for Protobuf from https://github.com/uarun/vim-protobuf/blob/master/indent/proto.vim

I've contacted the original author (Johannes Zellner) and he is ok with it being upstreamed.

Protobuf will fail to build unless the files are indented with 2 spaces, so it makes sense to have this be the default for any vim user editing proto files. Indenting in another way will throw an error like: `Found an incorrect indentation style "	". "  " is correct.`

### DIFF
--- a/.github/MAINTAINERS
+++ b/.github/MAINTAINERS
@@ -341,6 +341,7 @@ runtime/indent/perl.vim			@petdance
 runtime/indent/php.vim			@2072
 runtime/indent/postscr.vim		@mrdubya
 runtime/indent/prolog.vim		@dkearns
+runtime/indent/proto.vim		@Limero
 runtime/indent/ps1.vim			@heaths
 runtime/indent/qb64.vim			@dkearns
 runtime/indent/qml.vim			@ChaseKnowlden
--- a/runtime/indent/proto.vim
+++ b/runtime/indent/proto.vim
@@ -1,6 +1,6 @@
 " Vim indent file
 " Language:	Protobuf
-" Maintainer:	Johannes Zellner <johannes@zellner.org>
+" Maintainer:	David Pedersen <limero@me.com>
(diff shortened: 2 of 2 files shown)


chrisbra:
So you are submitting a new runtime file on behalf of Johannes Zellner? I'd very much prefer, if whoever volunteers to maintain the file sends it directly here. Can you ask him to at least ACK here then? Please add an entry to the [MAINTAINERS file](https://github.com/vim/vim/blob/master/.github/MAINTAINERS) ? 

Also this line here:
```
let b:undo_indent = "setl cin<"
```
doesn't undo all options set.

Limero:
> So you are submitting a new runtime file on behalf of Johannes Zellner? I'd very much prefer, if whoever volunteers to maintain the file sends it directly here. Can you ask him to at least ACK here then? Please add an entry to the [MAINTAINERS file](https://github.com/vim/vim/blob/master/.github/MAINTAINERS) ?
> 
> Also this line here:
> 
> ```
> let b:undo_indent = "setl cin<"
> ```
> 
> doesn't undo all options set.

I contacted him now to see if he wants to remain as the maintainer, otherwise I'll change it to be myself. I've updated the `undo_indent` now.

think-nice-things:
Hi, this is Johannes and I'm happy if David takes over maintainership.
In fact I don't use it since more than 20 years any more ;-)

Limero:
> Hi, this is Johannes and I'm happy if David takes over maintainership. In fact I don't use it since more than 20 years any more ;-)

That settles it, I've updated the maintainer and added myself to the `MAINTAINERS` file, thank you!

chrisbra:
thanks both!

PR #15451, opened 2024-08-08
runtime(doc): clarify directory of Vim's executable vs CWD

zeertzjq:
According to :h win32-PATH, "the same directory as Vim" means the same
directory as the Vim executable, not Vim's current directory.  In patch
8.2.4860 these two concepts were mixed up.

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -1,4 +1,4 @@
-*builtin.txt*	For Vim version 9.1.  Last change: 2024 Jul 28
+*builtin.txt*	For Vim version 9.1.  Last change: 2024 Aug 08
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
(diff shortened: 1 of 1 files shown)


runtime/doc/builtin.txt
```diff
@@ -2532,6 +2532,7 @@ executable({expr})					*executable()*
 		This function checks if an executable with the name {expr}
 		exists.  {expr} must be the name of the program without any
 		arguments.
+
 		executable() uses the value of $PATH and/or the normal
 		searchpath for programs.		*PATHEXT*
 		On MS-Windows the ".exe", ".bat", etc. can optionally be
```
k-takata:
```suggestion
		executable() uses the value of $PATH and/or the normal
		searchpath for programs.
							*PATHEXT*
		On MS-Windows the ".exe", ".bat", etc. can optionally be
```
Inserting a linebreak before `*PATHEXT*` might be better?

zeertzjq: Yes, makes sense

chrisbra:
thanks!

Issue #12470, opened 2023-05-30
:for can change readonly variable


zeertzjq:
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
for v:maxcol in range(1)
endfor
echomsg v:maxcol
```
3. `v:maxcol` is changed to 0

### Expected behaviour

`E46: Cannot change read-only variable "v:maxcol"`

### Version of Vim

9.0.1572

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

PR #15452, opened 2024-08-08
Problem: locked var can be changed in for loop

### LINKED ISSUE #15450: :for can bypass lock on variable
### Steps to reproduce

1. Run `vim --clean`
2. Source the following file:
```vim
const g:constvar = 10
for g:constvar in range(5)
endfor
```
3. No error is given and the value of `g:constvar` changes to 4. This also happens for Vim9 script (adding a `vim9script` line at the top).

### Expected behaviour

A `E741: Value is locked: g:constvar` error is given and the value of `g:constvar` remains 10. This was the behavior before patch 8.2.3511.
For `v:` variables this was fixed in patch 9.0.1611, but the problem still exists for locked (const) variables.

### Version of Vim

9.1.648

### Environment

Operating system: Arch Linux
Terminal: kitty
Value of $TERM: xterm-kitty
Shell: fish

### Logs and stack traces

_No response_

chrisbra:
Problem:  locked var can be changed in for loop
Solution: Check for locked value when assigning from for loop

related: #12470
closes: #15450

### DIFF
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -190,6 +190,7 @@ static int do_unlet_var(lval_T *lp, char_u *name_end, exarg_T *eap, int deep, vo
 static int do_lock_var(lval_T *lp, char_u *name_end, exarg_T *eap, int deep, void *cookie);
 static void list_one_var(dictitem_T *v, char *prefix, int *first);
 static void list_one_var_a(char *prefix, char_u *name, int type, char_u *string, int *first);
+static int var_check_ro_or_locked(dictitem_T *di, char_u *name);
 
 /*
  * Initialize global and vim special variables
@@ -4104,7 +4105,7 @@ set_var_const(
 			&& ((flags & ASSIGN_COMPOUND_OP) == 0
 			    || !type_inplace_modifiable))
 				 ? var_check_permission(di, name) == FAIL
-				 : var_check_ro(di->di_flags, name, FALSE))
+				 : var_check_ro_or_locked(di, name) == FAIL)
 		goto failed;
 	}
 	else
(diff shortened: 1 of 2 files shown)


zeertzjq:
This may still be inconsistent if the variable is locked with `:lockvar 0`.
The following gives E1122 error:
```vim
vim9script
def Func()
  g:var = 10
  lockvar 0 g:var
  for g:var in range(5)
  endfor
enddef
Func()
```
But the following two don't:
```vim
let g:var = 10
lockvar 0 g:var
for g:var in range(5)
endfor
```
```vim
vim9script
g:var = 10
lockvar 0 g:var
for g:var in range(5)
endfor
```
Not sure how to solve this. Maybe a separate DI_FLAGS_ is needed for variables created with `ASSIGN_FOR_LOOP`?

zeertzjq:
I may have found a simple solution:
```diff
diff --git a/src/evalvars.c b/src/evalvars.c
index 6facbeb13..f18b51637 100644
--- a/src/evalvars.c
+++ b/src/evalvars.c
@@ -4100,7 +4100,7 @@ set_var_const(
 
 	    // Modifying a final variable with a List value using the "+="
 	    // operator is allowed.  For other types, it is not allowed.
-	    if (((flags & ASSIGN_FOR_LOOP) == 0
+	    if ((((flags & ASSIGN_FOR_LOOP) == 0 || (flags & ASSIGN_DECL) == 0)
 			&& ((flags & ASSIGN_COMPOUND_OP) == 0
 			    || !type_inplace_modifiable))
 				 ? var_check_permission(di, name) == FAIL
```

chrisbra:
thanks, closing this then for # 15454

PR #15453, opened 2024-08-08
runtime(dosbatch): Add syn-sync

k-takata:
Closing parentheses were often highlighted as errors. Add a syntax sync command to reduce the error.

Also fix that `defined` was not highlighted as an operator inside parentheses.  E.g.:
```
if defined foo (
    if defined bar (
        ...
    )
)
```
The first `defined` was highlighted but the second one was not.

Also fix that `%%i` was not shown as a variable. Inside batch files, `%%i` must be used instead of `%i`.


chrisbra:
ping @mrdubya

chrisbra:
thanks, let me include it now.

PR #15454, opened 2024-08-08
Locked variable can be changed in a :for loop

zeertzjq:
Problem:  Locked variable can be changed in a :for loop.
Solution: Always do a full permission check on the first loop iteration
          where ASSIGN_DECL is not set.

related: #12470
fixex: #15450


chrisbra:
thanks!

PR #15456, opened 2024-08-09
assert_equal() doesn't show multibyte string correctly

zeertzjq:
Problem:  assert_equal() doesn't show multibyte string correctly.
Solution: Properly advance over a multibyte char.


chrisbra:
thanks!

PR #15464, opened 2024-08-09
crash with WinNewPre autocommand

chrisbra:
Problem:  crash when lastwin is null and attempting to calculate cmdline_row
Solution: validate that lastwin is a valid window pointer.


zeertzjq:
It seems that the crash with `lastwin == NULL` and `firstwin == NULL` can still be triggered in other ways,
1. Using `:redraw`
```vim
set showmode
autocmd WinNewPre * redraw
tabnew
```
2. Using `:wincmd t`
```vim
autocmd WinNewPre * wincmd t
tabnew
```
3. Using `:wincmd b`
```vim
autocmd WinNewPre * wincmd b
tabnew
```
4. Setting 'cmdheight'
```vim
autocmd WinNewPre * set cmdheight+=1
tabnew
```

It may be safer to just move the triggering of `WinNewPre` to somewhere where `firstwin` and `lastwin` or not `NULL`, but on the other hand `firstwin == NULL` make `winnr('$')` returns 0 when opening a new tabpage, which may indeed be correct as there are no windows in the tabpage yet.

chrisbra:
Well, It seems `WinNewPre` is triggered, when the window structures are not readily setup. At least in the `:tabnew` event. Since the whole idea of calling `WinNewPre` is to be able to get notified before the window layout is changed, we cannot defer it to when the window layout has been initialized. Maybe we can trigger it in `tabpage_leave()` instead (or we don't trigger it for `:tabnew` events at all, this seems too risky).

src/ex_getln.c
```diff
@@ -3925,7 +3925,8 @@ redrawcmd(void)
 compute_cmdrow(void)
 {
     // ignore "msg_scrolled" in update_screen(), it will be reset soon.
-    if (exmode_active || (msg_scrolled != 0 && !updating_screen))
+    if (exmode_active || (msg_scrolled != 0 && !updating_screen)
+	    || !win_valid(lastwin))
```
zeertzjq: This change is not needed anymore

chrisbra:
Ah yes of course thanks Am 10.08.2024 um 10:01 schrieb zeertzjq ***@***.***>:﻿
@zeertzjq commented on this pull request.



In src/ex_getln.c:
> +    if (exmode_active || (msg_scrolled != 0 && !updating_screen)
+	    || !win_valid(lastwin))

This change is not needed anymore

chrisbra:
included as of https://github.com/vim/vim/commit/fb3f9699362f8d51c3b48fcaea1eb2ed16c81454

PR #15465, opened 2024-08-09
[vim9class] Only allow child class override method signatures that require no runtime checks.

### LINKED ISSUE #15442: [vim9class] return type of method in child class not properly enforced
### Steps to reproduce

In the following, in the child class, the method String() returns a different type. No error.

```vim
vim9script

class A
    def String(): string
        return 'S'
    enddef
endclass

class B extends A
    def String(): any
        return 2
    enddef
endclass

def F()
    echo A.new().String()
    echo B.new().String()
enddef
F()
```
output
```
S
2
```

### Expected behaviour

Note that 9.1 behaves the same.

There should be a CHECKTYPE instruction for the return value of `B`'s `String()`. The `any` return type for the method in the child class is probably the source of the problem; I wonder if it should be considered an error and exact match required; which is 

 In vim9class, `:help extends`, it says
```
Object methods of the base class can be overruled.  The signature (arguments,
argument types and return type) must be exactly the same.
```
But this may have changed during the discussions about covariance; but I thought that only applied to object hierarchies.

**Any objections/comments on requiring the same type, except for class objects?** @dkearns comments?

This reminds me a little of #15208/#15215, but I don't really think type checking is needed here; just don't allow any.

### Version of Vim

9.1.661

### Environment

ubuntu/gtk3

### Logs and stack traces

_No response_

errael:
Fix #15442; the issue has considerable discussion.

Enforce the spec, 4th paragraph after `:help extends`
```
*vim9class.txt*	For Vim version 9.1.  Last change: 2024 Apr 13
...
Object methods of the base class can be overruled.  The signature (arguments,
argument types and return type) must be exactly the same.
```

Note this is not related to recent issues around method dispatch, but it may be that enforcing the same signature simplifies some solutions to those issues.

### DIFF
--- a/runtime/doc/vim9class.txt
+++ b/runtime/doc/vim9class.txt
@@ -654,10 +654,11 @@ is not possible to override them (unlike some other languages).
 
 						*E1356* *E1357* *E1358*
 Object methods of the base class can be overruled.  The signature (arguments,
-argument types and return type) must be exactly the same.  If the return type
-of a method is a class, then that class or one of its subclasses can be used
-in the extended method.  The method of the base class can be called by
-prefixing "super.".
+argument types and return type) must be compile time compatible with the base
+class signature; no runtime checks are required.  If the return type of a
+method is a class, then that class or one of its subclasses can be used in the
+extended method.  The method of the base class can be called by prefixing
+"super.".
 
 						*E1377*
 The access level of a method (public or protected) in a child class should be
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -7554,6 +7554,24 @@ def Test_extended_obj_method_type_check()
     endclass
   END
   v9.CheckSourceFailure(lines, 'E1383: Method "F": type mismatch, expected func(...list<any>) but got func(list<any>)', 10)
+
+  lines =<< trim END
+    vim9script
+
+    class A
+        def String(): string
+            return 'S'
+        enddef
+    endclass
+    class B extends A
+        def String(): any
+            return '?'
+        enddef
+    endclass
+
+    var o = B.new().String()
+  END
+  v9.CheckSourceFailure(lines, 'E1383: Method "String": type mismatch, expected func(): string but got func(): any', 12)
 enddef
 
 " Test type checking for class variable in assignments
--- a/src/vim9class.c
+++ b/src/vim9class.c
@@ -1427,8 +1427,14 @@ add_classfuncs_objmethods(
 			where_T where = WHERE_INIT;
 			where.wt_func_name = (char *)pname;
 			where.wt_kind = WT_METHOD;
-			(void)check_type(pf->uf_func_type, cf->uf_func_type,
-								TRUE, where);
+			int ret = check_type_maybe(pf->uf_func_type,
+						   cf->uf_func_type,
+						   TRUE, where);
+			// Signature must be such that no runtime checks
+			// are required. Only this compile time check.
+			if (ret == MAYBE)
+			    type_mismatch_where(pf->uf_func_type,
+						cf->uf_func_type, where);
 		    }
 		}
 	    }


zzzyxwvut:
I'm wondering about the scope of this change.  The following  
example works as before:

```
vim9script

interface Evaluable
    def Eval(_: any): any
endinterface

abstract class Word implements Evaluable
    const _nonWord: string

    abstract def Eval(_: any): any
endclass

class TextSplitter extends Word implements Evaluable
    def new(this._nonWord)
    enddef

    def Eval(word: string): list<string>
        return word->split(super._nonWord)
    enddef
endclass

class TextJoiner extends Word implements Evaluable
    def new(this._nonWord)
    enddef

    def Eval(ws: list<string>): string
        return ws->join(super._nonWord)
    enddef
endclass

class Product implements Evaluable
    def Eval(xs: list<number>): number
        return xs->reduce((t: number, x: number) => t * x, 1)
    enddef
endclass

echo "5!:\t120" == TextJoiner.new("\t")
    .Eval(TextSplitter.new(' ')
            .Eval('5!: ')
                + [string(Product.new()
                               .Eval(range(1, 5)))])
```

errael:
Interesting; oops. The code that checks the override signature, `vim9class.c::add_classfuncs_objmethods()`, has the comment
```
// If the child class overrides a function from the parent
// the signature must be equal.
```
This code used to call `check_type()` which let through the data corruption. This PR makes the change to use `check_type_maybe()`, and considers the `MAYBE` case as an error. I guess I put to much faith in the spec/comments.

With this change a `TYPECHECK` instruction never needs to be generated. This satisfies the primary goal of no corrupted data. No spec change was also a goal, but this change is compatible with the spirit of covariance which was introduced late in the game. With this change it's still only compile time checks.

I'm OK with this PR and would be OK with changing the spec to something like
> the child class method must be compatible with the base class method
> so that no type checking is required.

Considering your example, it looks like no bad types would slip through, do you agree?

In the future, some restrictions could be relaxed and runtime checks introduced as discussed in #15442. But that's beyond the goal of simply insuring no bad data with only compile time checks.

FYI,
```
/*
 * As check_type() but return MAYBE when a runtime type check should be used
 * when compiling.
 */
    int
check_type_maybe(
```

errael:
Update the title. Don't require exact signature match; require resolution with compile time checks.

If this is OK, the spec can be addressed.

chrisbra:
> Enforce the spec, 4th paragraph after `:help extends`
> 
> ```
> *vim9class.txt*	For Vim version 9.1.  Last change: 2024 Apr 13
> ...
> Object methods of the base class can be overruled.  The signature (arguments,
> argument types and return type) must be exactly the same.
> ```

Thanks. I agree this makes sense, considering your example and the documentation quoted.

errael:
> Thanks. I agree this makes sense

OK, I'll update `:help extends` as mentioned.

(changing to draft until then...)

errael:
> > @dkearns are you okay here?
> 
> @chrisbra, no as discussed in #15442 I think the current behaviour is correct.

And, as discussed in https://github.com/vim/vim/issues/15442#issuecomment-2289188239, I disagree.

Is there an example of a strongly typed system, that doesn't have return type polymorphism, which has this?

zzzyxwvut:
I played a bit more with this patch.

In Java, it is possible to use bounded type arguments for  
covariant return types.  Pay close attention to `class V`, in  
whose declaration we declare type variable (parameter) `O` of  
type `B` _or_ any subtype of `B` (note that `extends` here also  
includes `B`).  Then we use `O` for the return type of  `getA()`  
and elsewhere.  Observe that `getA()` promises to return `B`  
_or_ any subtype of `B`.

```java
class CovariantReturnTypes
{
	static class A extends Object
	{
		@Override public String toString()
		{
			return this.getClass().getName();
		}
	}

	static class B extends A { }
	static class C extends B { }
	static class D extends C { }

	static class T extends Object
	{
		A getA() { return new A(); }
	}

	static class U extends T
	{
		B getA() { return new B(); }
	}

	static class V<O extends B> extends U
	{
		private final O o;

		V(O o) { this.o = o; }

		O getA() { return this.o; }
	}

	static class W extends V<C>
	{
		private static final C DUMMY = new C();

		W() { super(DUMMY); }

		D getA() { return new D(); }

		//// This definition is fine too.
////		C getA() { return new C(); }
	}

	public static void main(String[] args)
	{
		System.out.println(new T().getA());
		System.out.println(new U().getA());

		System.out.println(new V<>(new B()).getA());
		System.out.println(new V<>(new C()).getA());
		System.out.println(new V<>(new D()).getA());

		System.out.println(new W().getA());
	}
}
```

Note that there is no real need to make `V` generic as shown  
above.  Here is an alternative implementation and its use:

```java
	static class V extends U
	{
		private final B o;

		V(B o) { this.o = o; }

		B getA() { return this.o; }
	}

	static class W extends V { /* No change here. */ }

	public static void main(String[] args)
	{
		// Other examples.
		System.out.println(new V(new B()).getA());
		System.out.println(new V(new C()).getA());
		System.out.println(new V(new D()).getA());
	}
```

Below is a Vim9 port (simulating the type argument `O` with  
`any` and enforcing its bound in the constructor):

```
vim9script

class A
endclass
class B extends A
endclass
class C extends B
endclass
class D extends C
endclass

class T
    def GetA(): A
        return A.new()

        # Any of the following variants is also correct.
#       return B.new()
#       return C.new()
#       return D.new()
    enddef
endclass

class U extends T
    def GetA(): B
        return B.new()

        # Any of the following variants is also correct.
#       return C.new()
#       return D.new()
    enddef
endclass

class V extends U
    const _b: B

    def new(b: B)
        this._b = b
    enddef

    def GetA(): any
        return this._b
    enddef
endclass

class W extends V
    def GetA(): D
        return D.new()
    enddef
endclass

echo typename(T.new().GetA())
echo typename(U.new().GetA())
echo typename(V.new(B.new()).GetA())
echo typename(V.new(C.new()).GetA())
echo typename(V.new(D.new()).GetA())
echo typename(W.new().GetA())

try
    echo typename(V.new(A.new()).GetA())
catch
    echo 'ERROR: ' .. v:errmsg
endtry

try
    echo typename(V.new('.').GetA())
catch
    echo 'ERROR: ' .. v:errmsg
endtry




## This type is correctly rejected when SOURCED.
## TODO: Wrong return type and value! Must be at least B.
#class V2 extends U
#    def GetA(): A
#        return A.new()
#    enddef
#endclass

## This type is correctly rejected when SOURCED.
## TODO: Wrong return type! Must be at least B.
#class V3 extends U
#    def GetA(): A
#        return B.new()
#    enddef
#endclass

## This type is correctly rejected when COMPILED.
## TODO: Wrong return value! Must be at least B.
class V4 extends U
    def GetA(): B
        return A.new()
    enddef
endclass

# defcompile V4
```

With the patch, a valid `V` implementation is rejected.

errael:
> Below is a Vim9 port (simulating the type argument `O` with
> `any` and enforcing its bound in the constructor):

Would you be so kind to provide a small/minimal test case that shows the problem?

zzzyxwvut:
How about this pared down version?

```
vim9script

class A
endclass
class B extends A
endclass

class U
    def GetA(): A
        return A.new()
    enddef
endclass

class V extends U
    const _b: B

    def new(b: B)
        this._b = b
    enddef

    def GetA(): any
        return this._b
    enddef
endclass

echo typename(U.new().GetA())
echo typename(V.new(B.new()).GetA())

try
    echo typename(V.new(A.new()).GetA())
catch
    echo 'ERROR: ' .. v:errmsg
endtry

try
    echo typename(V.new('.').GetA())
catch
    echo 'ERROR: ' .. v:errmsg
endtry
```

All `catch` blocks should catch bad arguments.

errael:
Thanks for the "pared down" test; it is not clear what it is supposed to show. Running it on unmodified `v.9.1.678` it gets errors. I would have thought it was supposed to work on release vim, and then fail when run with this patch.

In the above, there's
```
class V extends U
    def new(b: B)
```
and unsurprisingly doing
```
V.new(A.new())
```
generates
```
/tmp/k.vim, line 30
Vim(eval):E1013: Argument 1: type mismatch, expected object<B> but got object<A>
ERROR: E325: ATTENTION
```
And the second one has `V.new('.')` which is expected to fail.

zzzyxwvut:
The patched version would summarily reject type `V` because  
`any`, returned by `GetA()`, need not be a type derived from `A`.

Non-patched versions would allow type `V` and let a user deal  
with its `any`; and a user can in fact enforce covariance as  
shown: define a constructor (or a method) with a parameter  
type that matches the covariant bound (only in the eye of  
the beholder unfortunately).  With this type check in place,  
contra-variant types, like `A`, and invariant types, like  
string, are rejected in the same way as if `GetA()` returned  
`B` not `any` (hence the `catch` blocks).

Yes, this is a roundabout approach, but still a valid one  
that is currently supported.

> `E325: ATTENTION`

Aside: I remember staring at `E484: Can't open file 5` after  
editing, saving, re-sourcing a Vim9 script (not this one)  
several times in succession.

errael:
If you want the type handling/checking to be done by code rather than the compiler, then the base class should be defined with `any`.

If the base class can't be changed, then changing the contract advertised by the base class and subverting the type system makes no sense and probably breaks the code where the base class is defined.

Composition rather than inheritance is probably a better option in this kind of situation.

> this is a roundabout approach, but still a valid one that is currently supported.

One persons "feature" is another's "bug". IMHO, this is simply a bug.

I've yet to see a real world example of where this type system violation is of use. Make the base class type `any` if that's what the situation requires.

---

I still don't understand what the pared down example is supposed to show. There's
```
    def GetA(): any
        return this._b
    enddef
```
But this could be defined
```
def GetA(): A
```
and that works since `B` is an `A`

PR #15467, opened 2024-08-10
Some other options reset curswant unnecessarily when set

### LINKED ISSUE #15462: Setting `'commentstring'` resets the "remembered" column
### Steps to reproduce

With the following buffer:

```
1 here
2   is
3     s\%#ome
4
5     more
6   indented
7 text
```

with `\%#` being your cursor, move it up and down works as expected, remembering the column.  Now move your cursor to line 4 and issue `:setlocal commentstring=#\ %s`.  The cursor is now glued to column 1.

Apologies if this is intended behaviour but I couldn't find an issue about it.

Here's a video:

https://github.com/user-attachments/assets/61798724-263d-4e54-b48d-144bfe632a3a

### Expected behaviour

It remembers the cursor column after setting `'commentstring'`.

Thank you all for your continued work on Vim.  It's much appreciated!

### Version of Vim

9.1, included patches 1-500

### Environment

MacOS
Kitty
xterm-kitty
zsh

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Some other options reset curswant unnecessarily when set.
Solution: Don't reset curswant when setting 'comments', 'commentstring'
          or 'define'.

fixes: #15462

### DIFF
--- a/src/optiondefs.h
+++ b/src/optiondefs.h
@@ -621,13 +621,12 @@ static struct vimoption options[] =
     {"columns",	    "co",   P_NUM|P_NODEFAULT|P_NO_MKRC|P_VI_DEF|P_RCLR,
 			    (char_u *)&Columns, PV_NONE, NULL, NULL,
 			    {(char_u *)80L, (char_u *)0L} SCTX_INIT},
-    {"comments",    "com",  P_STRING|P_ALLOCED|P_VI_DEF|P_ONECOMMA
-							  |P_NODUP|P_CURSWANT,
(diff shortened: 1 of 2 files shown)


zeertzjq:
The `P_CURSWANT` flag was added to these options in the initial `P_CURSWANT` patch (i.e. patch 7.3.487), which initially came from https://groups.google.com/g/vim_dev/c/rAr6MYuro-8/m/HPQadQa6nrQJ

In that patch there was a `/* TODO: List more safe options. */`, so I think the `P_CURSWANT` wasn't added intentionally.

chrisbra:
thanks makes sense.

Issue #14490, opened 2024-04-10
Vim clone left "dirty" after make. src/po/ru.cp1251.po is modified.


errael:
### Steps to reproduce

I occasionally see this, and I think I've seen mention of this on the dev list; it might always be the same file; this morning I happened to catch it just after it happened. I also built yesterday morning. It doesn't happen too often (the revert file I see is from January), I haven't looked at makefiles, I don't know what dependency is triggering.

If anyone knows something specific to check...

After pull, make reconfig, the console shows the make ending as:
```
...
OLD_PO_FILE_INPUT=yes msgfmt -v -o vi.mo vi.po
1410 translated messages.
OLD_PO_FILE_INPUT=yes msgfmt -v -o zh_CN.UTF-8.mo zh_CN.UTF-8.po
2800 translated messages.
OLD_PO_FILE_INPUT=yes msgfmt -v -o zh_TW.UTF-8.mo zh_TW.UTF-8.po
1410 translated messages.
make[2]: Leaving directory '/ref/vim/src/po'
make[2]: Entering directory '/ref/vim/src/po'
OLD_PO_FILE_INPUT=yes msgfmt -v -o cs.cp1250.mo cs.cp1250.po
1272 translated messages.
...
OLD_PO_FILE_INPUT=yes msgfmt -v -o pl.cp1250.mo pl.cp1250.po
1858 translated messages.
rm -f ru.cp1251.po
iconv -f UTF-8 -t CP1251 ru.po | \
	LANG=C sed -e 's/charset=[uU][tT][fF]-8/charset=CP1251/' \
		-e 's/# Original translations/# Generated from ru.po, DO NOT EDIT/' \
		> ru.cp1251.po
OLD_PO_FILE_INPUT=yes msgfmt -v -o ru.cp1251.mo ru.cp1251.po
2979 translated messages.
...
make[2]: Leaving directory '/ref/vim/src/po'
make[1]: Leaving directory '/ref/vim/src'
err@harmony:/ref/vim/src
$ hg st
M src/po/ru.cp1251.po
```

The diff seems to be a lot of formatting. Here's an excerpt, as seen in vim.
```
-"W17: Äëÿ ðàáîòû ñ àðàáñêîé ïèñüìåííîñòüþ ââåäèòå êîìàíäó `:set "
-"encoding=utf-8`"
[...]

PR #15468, opened 2024-08-10
translations: Update generated po files

k-takata:
Some auto-generated po files cause differences after building Vim. Update them.


k-takata:
`ru.cp1251.po` also causes differences, but I think it is discussed in some other issues...

k-takata:
It was discussed in #14490.

chrisbra:
thanks!

PR #15471, opened 2024-08-10
PO file conversion does not work on FreeBSD

RestorerZ:
Problem: PO file conversion does not work on FreeBSD.
Solution: In the file conversion rules, change the `$<` variable to `$?`.
This is a portable variable, unrecognizable by both GNU make and BSD make.

FreeBSD:
>.OODATE   The  list  of  sources for this target that were deemed out-of-date; also known as `?'.

GNU:
> $? The names of all the prerequisites that are newer than the target, with spaces between them. If the target does not exist, all prerequisites will be included.

### DIFF
--- a/src/po/Makefile
+++ b/src/po/Makefile
@@ -107,55 +107,55 @@ nb.po: no.po
 ja.sjis.po: ja.po
 	@$(MAKE) sjiscorr
 	rm -f $@
-	iconv -f UTF-8 -t CP932 $< | ./sjiscorr > $@
+	iconv -f UTF-8 -t CP932 $? | ./sjiscorr > $@
 
 sjiscorr: sjiscorr.c
 	$(CC) -o sjiscorr sjiscorr.c
 
 ja.euc-jp.po: ja.po
-	iconv -f UTF-8 -t EUC-JP $< | \
+	iconv -f UTF-8 -t EUC-JP $? | \
 		$(SED) -e 's/charset=[uU][tT][fF]-8/charset=EUC-JP/' \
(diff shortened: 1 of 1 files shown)


chrisbra:
Hm, semantically, it makes a difference if we explicitly want to use only the *first* pre-requisite or *all* pre-requisites. On a quick check however, it doesn't seem to matter, as we always seem to have only a single pre-requisite, so the change should be fine. However, I noticed in the documentation of BSD Make this section:

> The shorter forms (`>`, `!`, `<`, `%`, `?`, `*`, and `@`) are permitted for backward compatibility with historical makefiles and	 legacy POSIX make and are not recommended.

which seem to indicate that using `$<` should be okay?  so when did you notice this issue? Also we seem to have the `$<` in a few more Makefiles. So I would think those are also affected.

RestorerZ:
@chrisbra 
Christian, yes for these rules it would be acceptable as there is only one dependency and it should supposedly always be newer than the target file.

> which seem to indicate that using $< should be okay? 

This variable is used in *BSD for inference rules.

>  .IMPSRC   In suffix-transformation	rules, the  name/path  of  the
> 		       source  from which the target is	to be transformed (the
> 		       "implied" source); also known as	`<'.  It  is  not  de-
> 		       fined in	explicit rules.

> Also we seem to have the $< in a few more Makefiles.

Yes, there are a few more make files, particularly in the "doc" directory. I'm planning to fix it next week. That is, if everything goes well.

chrisbra:
okay, then let me merge this one. Thanks!

PR #15477, opened 2024-08-12
Function dispatch problems when using super.

### LINKED ISSUE #15448: [vim9class] Non-overridden methods with a `super` call recurse
### Steps to reproduce

The overridden methods that invoke `super` methods recurse  
when another derived type does not override the same methods  
in turn.

```
vim9script

class A
    def string(): string
        return 'A'
    enddef
endclass

class B extends A
    def string(): string
        return super.string()
    enddef
endclass

class C extends B
#### Fix it with an implementation without 'super'.
#   def string(): string
#       return 'C'
#   enddef
endclass

echo C.new().string()
```

### Expected behaviour

There is no reason for any recursion as written.

### Version of Vim

v9.1.0661

### Environment

### Logs and stack traces

_No response_

errael:
Fix #15448 and #15463: (2) super.method may not execute in context of defining class

When compiling "super.Func()" force class context to class that defines function that is doing "super.Func()".
ISN_METHODCALL arg "cmf_is_super" for specific ufunc.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -3252,6 +3252,141 @@ def Test_using_base_class()
   v9.CheckSourceSuccess(lines)
 enddef
 
+def Test_super_dispatch()
+  # See #15448 and #15463
+  var lines =<< trim END
+    vim9script
+
+    class A
+        def String(): string
+            return 'A'
+        enddef
+    endclass
+
+    class B extends A
+        def String(): string
+            return super.String()
+        enddef
+    endclass
+
+    class C extends B
+    endclass
+
+    assert_equal('A', C.new().String())
+  END
+  v9.CheckSourceSuccess(lines)
+
+  lines =<< trim END
+    vim9script
+
+    class A
+        def F(): string
+            return 'AA'
+        enddef
+    endclass
+
+    class B extends A
+        def F(): string
+            return 'BB'
+        enddef
+        def S(): string
+            return super.F()
+        enddef
+        def S0(): string
+            return this.S()
+        enddef
+    endclass
+
+    class C extends B
+        def F(): string
+            return 'CC'
+        enddef
+        def ToB(): string
+            return super.F()
+        enddef
+    endclass
+
+    assert_equal('AA', B.new().S())
+    assert_equal('AA', C.new().S())
+    assert_equal('AA', B.new().S0())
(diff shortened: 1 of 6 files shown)


errael:
@zzzyxwvut FYI

zzzyxwvut:
Thank you, all related tests now work!

If you feel like adding another test, here is a working test  
prototype; it chains `super.` to `super.` and _then_ looks for  
`this.`:

```
vim9script

abstract class A
    abstract def _G(): string

    def F(): string
        return this._G()
    enddef
    def _H(): string
        return this.F()
    enddef
endclass

class B extends A
    def _G(): string
        return 'BBB'
    enddef
    def SF(): string
        return super._H()
    enddef
endclass

class C extends B
endclass

class D extends C
    def SF(): string
        return super.SF()
    enddef
endclass

class E extends D
    def SF(): string
        return super.SF()
    enddef
endclass

class F extends E
    def _G(): string
        return 'EEE'
    enddef
endclass

# D.new() -> A.F() -> B._G()
echo 'BBB' == E.new().F()
# F.new() -> E.SF() -> D.SF() -> B.SF() -> A._H() -> A.F() -> F._G()
echo 'EEE' == F.new().SF()
```

errael:
> Thank you, all related tests now work!
> 
> If you feel like adding another test, here is a working test prototype; it chains `super.` to `super.` and _then_ looks for `this.`:

Good to hear. Thanks for testing. And thanks for the test, I like the chain; added some sugar to check that all the chain elements execute (no compilation weirdness or whatever). Test now in the PR.

src/testdir/test_vim9_class.vim
```diff
@@ -3252,6 +3252,140 @@ def Test_using_base_class()
...
+    class E extends D
+        def SF(): string
+            flag = or(flag, 8)
+            return super.SF()
+        enddef
+    endclass
+
+    class F extends E
+        def _G(): string
+            return 'EEE'
```
zzzyxwvut: I have forgotten to update this value to 'FFF' when I took  
this method out of class `E`.  Maybe you should fix that and  
update `assert_equals()` accordingly.

errael: Done

errael:
Adjust the test to verify call chain ordering.

chrisbra:
thanks!

PR #15479, opened 2024-08-12
runtime(mediawiki): fix typos

stasjok:
Fixed `did_ftplugin` variable from global to buffer.

Also I think the meaning of `g:html_style_rendering` variable is reversed. I think it was meant `if exists("html_style_rendering")`.

Also maybe variables should be named with mediawiki perfix like `g:mediawiki_no_html_rendering` and `g:mediawiki_style_rendering` (no html here because it affects `''italic''` too).

@avidseeker

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -2374,7 +2374,7 @@ have the following in your .vimrc: >
 
 MEDIAWIKI					*ft-mediawiki-syntax*
 
-Be default, syntax highlighting includes basic HTML tags like style and
+By default, syntax highlighting includes basic HTML tags like style and
 headers |html.vim|. For strict Mediawiki syntax highlighting: >
 
 	let g:html_no_rendering = 1
--- a/runtime/ftplugin/mediawiki.vim
+++ b/runtime/ftplugin/mediawiki.vim
@@ -8,7 +8,7 @@
 if exists("b:did_ftplugin")
   finish
 endif
-let g:did_ftplugin = 1
+let b:did_ftplugin = 1
 
 " Many MediaWiki wikis prefer line breaks only at the end of paragraphs
 " (like in a text processor), which results in long, wrapping lines.


chrisbra:
thanks. Let's await @avidseeker  ACK then.

avidseeker:
I don't mind renaming to `g:mediawiki_style_rendering`, but as for `g:html_no_rendering` it exists for HTML too. Check `:h html-syntax`. I don't see why disabling HTML variable should be a separate thing for mediawiki syntax.

I agree that a more straightforward variable like `g:html_style_rendering` would be better than a negated one `g:html_no_rendering`, but again, this is how it is defined in html syntax.

stasjok:
> I don't mind renaming to `g:mediawiki_style_rendering`, but as for `g:html_no_rendering` it exists for HTML too. Check `:h html-syntax`. I don't see why disabling HTML variable should be a separate thing for mediawiki syntax.
> 
> I agree that a more straightforward variable like `g:html_style_rendering` would be better than a negated one `g:html_no_rendering`, but again, this is how it is defined in html syntax.

I'm not suggesting changing this, I just pointed out, that current `g:html_style_rendering` variable is confusing. Description in the help suggests that this variable will enable bold and italics, but it's the other way around. So either description (with adding 'no' to the variable name to reflect that it disables something), or actual behavior could be changed. But I'm not planing to do it myself. I submitted here only a small bugfix for `b:did_ftplugin` variable.

avidseeker:
I see. This should fix it then

```diff
diff --git a/runtime/syntax/mediawiki.vim b/runtime/syntax/mediawiki.vim
index 8ac30b93d..0c37e9141 100644
--- a/runtime/syntax/mediawiki.vim
+++ b/runtime/syntax/mediawiki.vim
@@ -17,6 +17,7 @@
 if exists("b:current_syntax")
   finish
 endif
+let b:mediawiki_style_rendering = 1

 syntax case ignore
 syntax spell toplevel
@@ -247,7 +248,7 @@ if !exists("html_no_rendering")

   hi def link htmlLink Underlined

-  if !exists("html_style_rendering")
+  if (&mediawiki_style_rendering)
     hi def htmlBold                term=bold                  cterm=bold                  gui=bold
     hi def htmlBoldUnderline       term=bold,underline        cterm=bold,underline        gui=bold,underline
     hi def htmlBoldItalic          term=bold,italic           cterm=bold,italic           gui=bold,italic
```

chrisbra:
> +  if (&mediawiki_style_rendering)

This of course needs to be `+  if (b:mediawiki_style_rendering)` and requires changes in the documentation.

stasjok:
Can this be done in a separate PR by @avidseeker or do you want me to incorporate the changes here?

Also I don't think it should be buffer-scoped variable. I think if someone wants to disable italics, they would want to do it globally.

chrisbra:
I agree, let's keep it off for now and leave the PR as is for now

PR #15486, opened 2024-08-13
docs: add tags for json + markdown global variables

JJCUBER:
I added help tags for them in the syntax.txt file since this is the only place they are mentioned.

### DIFF
--- a/runtime/doc/syntax.txt
+++ b/runtime/doc/syntax.txt
@@ -2157,7 +2157,8 @@ displayed line.  The default value is 10.  The disadvantage of using a larger
 number is that redrawing can become slow.
 
 
-JSON						*json.vim* *ft-json-syntax*
+JSON			*json.vim* *ft-json-syntax* *g:vim_json_conceal*
+					       *g:vim_json_warnings*
 
 The json syntax file provides syntax highlighting with conceal support by
 default. To disable concealment: >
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks, but what is the reason, why you think this is helpful?

JJCUBER:
> Thanks, but what is the reason, why you think this is helpful?

It isn't easy to search for information on them in the docs otherwise.

I personally ran into an issue where I was trying to find more information on them when I found a fix for concealment pertaining to the `indentLine` plugin.  I was unable to find any actual information on what these variables do, and they were working despite me having a version of the plugin that was from before they were documented.  To confirm my suspicions, I searched through their source code, and I found no checks for said variables, so I went on to grep through the vim syntax folder (based on where I suspected it would be).  I finally found it there, after which I searched through the vim repo to find all mentions of it throughout the code and helpdocs.

Long story short, it is entirely possible to run into one of these variables when looking at plugins or other people's vimrc's, and it is not immediately clear that the corresponding helpdocs entry would be within the `syntax.txt` file.  I wouldn't be surprised if someone else down the line runs into a similar situation and ends up wasting a bunch of time when it could have just been tagged properly.

*At some point in the future, I might create a PR for other global variables which aren't currently properly tagged in the helpdocs. 
 Note that some are already properly documented, i.e. `g:syntax_on`.  (Of course, assuming you all are okay with that.)*

chrisbra:
There is always `:helpgrep`. I am just a bit worried, that if we start adding helptags for each and every vimscript variable, we may be littering the help pages  too much, so we need to be a bit careful. By itself this change however is fine.

JJCUBER:
> There is always `:helpgrep`.

That command completely slipped my mind, thanks for that.

> I am just a bit worried, that if we start adding helptags for each and every vimscript variable, we may be littering the help pages too much, so we need to be a bit careful. By itself this change however is fine.

That's a fair point, I can see it muddying things.  Right now, the docs seems to be in an in-between state (since a chunk do have tags and a chunk don't).  For reference, there are 245 entries starting with `g:` in `:h help-tags` for `vim 9.0`.

Issue #3811, opened 2019-01-16
Feature Request: Trigger balloon via keyboard at cursor position


marcotrosi:
Hi,

I want to trigger the balloon text via keyboard at cursor position. I read the documentation and it looks like that at the moment the balloon textbox can be triggered only with the mouse, which is in my humble opinion not very Vim-like. Mouse is okay of course, but keyboard should have higher priority, or am I wrong?

It could be a function or a command that I can map (nmap/imap/vmap/xmap/smap) or use in Vimscript. Just the standard Vim solutions.

What do you think?

Thanks for your feedback

Have a great day

--marco

PR #15487, opened 2024-08-13
Fix balloon evaluation.

### LINKED ISSUE #15359: [Termdebug] balloon display evaluation of variable under mouse doesn't work
### Steps to reproduce

- `gvim -c "packadd termdebug" -c "Termdebug" ...`
- set a breakpoint then run
- At breakpoint hover the mouse over a variable in the breakpointed function
- nothing happens

### Expected behaviour

Staring up with Termdebug8, using the same version of vim, and things work as expected.

### Version of Vim

9.1.618

### Environment

ubuntu gtk3

### Logs and stack traces

_No response_

ubaldot:
It should fix https://github.com/vim/vim/issues/15359

### DIFF
--- a/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
+++ b/runtime/pack/dist/opt/termdebug/plugin/termdebug.vim
@@ -1536,7 +1536,7 @@ def CleanupExpr(passed_expr: string): string
 enddef
 
 def Balloon_show(expr: string)
-  if has("+balloon_eval") || has("+balloon_eval_term")
+  if has("balloon_eval") || has("balloon_eval_term")
     balloon_show(expr)
   endif
 enddef


errael:
Test?

ubaldot:
Tbh I have no idea how to test if a balloon is displayed, but any hints are welcome.

errael:
I've never done it, but in testdir doing `grep balloon * | wc` there's 130 matches. One of them is
```
func Test_balloon_show_gui()
```

ubaldot:
I took a look at existing tests and I found some interesting functions. 
However, I am not sure I am doing good use of them and digging in the `:help` didn't help too much either.

The main problem is that I don't know how to move the mouse pointer from Vim. 

What I tried to do?
I used the following test file:

```
vim9script

def SimpleBalloon(): string
   return 'Cursor is at line/column: ' .. v:beval_lnum .. ', ' .. v:beval_col
   .. ' in file ' ..  bufname(v:beval_bufnr) ..
     '.. Word under cursor is: "' .. v:beval_text .. '"'
enddef

if has("balloon_eval") || has("balloon_eval_term")
  set balloonexpr=SimpleBalloon()
  if has("balloon_eval")
    set ballooneval
  endif
  if has("balloon_eval_term")
    set balloonevalterm
  endif
endif
```

and from there I source the following script:

```
vim9script

test_setmouse(9, 4)
feedkeys("\<MouseMove>\<Ignore>", "xt")
sleep 1
echom balloon_gettext()
```

I would expected to have the string that contains the word `"has"` in `:messages` but that is not the case. 
The mouse pointer didn't move at all.

chrisbra:
you would need a screen-dump test. Perhaps you could create a similar function like this:
https://github.com/vim/vim/blob/dc831db6ea3b550e3acb60a0859fcd07fd3f86e6/src/testdir/test_balloon.vim#L21-L48

But this seems overly complicated for such an simple change.

ubaldot:
Thanks! Yeah, it as interesting as complicated. And it may take several days of work. I agree, I am not sure if it's worth for such a tiny change.

louregni:
Hello,

I was hitting the issue, so I'm testing the master branch.
I have `set balloonevalterm` 

When I try to evaluate an expression using the termdebug plugin: hovering the expression with the mouse triggers the balloon.
Doing the same with the keyboard shortcut using `K` or `:Evaluate`, doesn't trigger the balloon, it shows the expression in the status bar.

This https://github.com/vim/vim/issues/3811 says that it should be possible to show the balloon from the cursor position and the keyboard shortcut.

This is a bug or this is not supported ?

ubaldot:
Good question. I am not 100% sure it is supposed to be supported. Could you please try with a pre Vim9 conversion version of Termdebug what happens?

PR #15489, opened 2024-08-13
runtime(doc): Updated man pages

RestorerZ:
"evim.1"
`/usr/local/lib/vim/evim.vim`
changed to
`/usr/local/share/vim/vim??/evim.vim`
Added description of "vim???".
Typographical corrections.

"vim.1"
Arranged command line arguments.
Added a note for the `-F` argument.
Typographical corrections.

"vimdiff.1"
Corrected the date of the last modification.

"vimtutor.1"
`/usr/local/lib/vim/tutor/tutor[.language]`
`/usr/local/lib/vim/tutor/tutor.vim`
changed to
`/usr/local/share/vim/vim???/tutor/tutor/tutor[.language]`
`/usr/local/share/vim/vim??/tutor/tutor.vim`
Added description of "vim???".

"xxd.1"
Corrected the date of the last modification.
Corrected examples, fixed a mistake.
Added date of last change to version number.

"test_xxd.vim"
Updated test data for the new values in xxd.1

### DIFF
--- a/runtime/doc/evim.1
+++ b/runtime/doc/evim.1
@@ -1,4 +1,4 @@
-.TH EVIM 1 "2002 February 16"
+.TH EVIM 1 "2024 August 12"
 .SH NAME
 evim \- easy Vim, edit a file with Vim and setup for modeless editing
 .SH SYNOPSIS
@@ -34,8 +34,12 @@ Use CTRL-Q to obtain the original meaning of CTRL-V.
 See vim(1).
 .SH FILES
 .TP 15
-/usr/local/lib/vim/evim.vim
+/usr/local/share/vim/vim??/evim.vim
 The script loaded to initialize eVim.
+.br
+.I vim??
+is short version number, like vim91 for
+.B Vim 9.1
 .SH AKA
 Also Known As "Vim for gumbies".
 When using evim you are expected to take a handkerchief,
--- a/runtime/doc/vim.1
+++ b/runtime/doc/vim.1
@@ -1,4 +1,4 @@
-.TH VIM 1 "2024 Aug 03"
+.TH VIM 1 "2024 Aug 12"
 .SH NAME
 vim \- Vi IMproved, a programmer's text editor
 .SH SYNOPSIS
@@ -150,18 +150,6 @@ Example: vim "+set si" main.c
 .br
 Note: You can use up to 10 "+" or "\-c" commands.
 .TP
-\-S {file}
-{file} will be sourced after the first file has been read.
-This is equivalent to \-c "source {file}".
-{file} cannot start with '\-'.
-If {file} is omitted "Session.vim" is used (only works when \-S is the last
-argument).
-.TP
-\-\-cmd {command}
-Like using "\-c", but the command is executed just before
-processing any vimrc file.
-You can use up to 10 of these commands, independently from "\-c" commands.
-.TP
 \-A
 If
 .B Vim
@@ -224,11 +212,6 @@ is executed by a program that will wait for the edit
 session to finish (e.g. mail).
 On the Amiga the ":sh" and ":!" commands will not work.
 .TP
-\-\-nofork
-Foreground.  For the GUI version,
-.B Vim
-will not fork and detach from the shell it was started in.
-.TP
 \-F
 If
 .B Vim
@@ -239,6 +222,8 @@ in Farsi mode, i.e. 'fkmap' and 'rightleft' are set.
 Otherwise an error message is given and
 .B Vim
 aborts.
+.br
+Note: Farsi support has been removed in patch 8.1.0932.
 .TP
 \-g
 If
@@ -247,18 +232,6 @@ has been compiled with GUI support, this option enables the GUI.
 If no GUI support was compiled in, an error message is given and
 .B Vim
 aborts.
-.TP
-\-\-gui-dialog-file {name}
-When using the GUI, instead of showing a dialog, write the title and message of
-the dialog to file {name}.  The file is created or appended to.  Only useful
-for testing, to avoid that the test gets stuck on a dialog that can't be seen.
-Without the GUI the argument is ignored.
-.TP
-\-\-help, \-h, \-?
-Give a bit of help about the command line arguments and options.
-After this
-.B Vim
-exits.
 .TP
 \-H
 If
@@ -277,12 +250,12 @@ instead of the default "~/.viminfo".
 This can also be used to skip the use of the .viminfo file, by giving the name
 "NONE".
 .TP
-\-L
-Same as \-r.
-.TP
 \-l
 Lisp mode.
 Sets the 'lisp' and 'showmatch' options on.
+.TP
+\-L
+Same as \-r.
 .TP
 \-m
 Modifying files is disabled.
@@ -294,19 +267,19 @@ Modifications not allowed.  The 'modifiable' and 'write' options will be unset,
 so that changes are not allowed and files can not be written.  Note that these
 options can be set to enable making modifications.
 .TP
-\-N
-No-compatible mode.  Resets the 'compatible' option.
-This will make
-.B Vim
-behave a bit better, but less Vi compatible, even though a .vimrc file does
-not exist.
-.TP
 \-n
 No swap file will be used.
 Recovery after a crash will be impossible.
 Handy if you want to edit a file on a very slow medium (e.g. floppy).
 Can also be done with ":set uc=0".
 Can be undone with ":set uc=200".
+.TP
+\-N
+No-compatible mode.  Resets the 'compatible' option.
+This will make
+.B Vim
+behave a bit better, but less Vi compatible, even though a .vimrc file does
+not exist.
 .TP
 \-nb
 Become an editor server for NetBeans.  See the docs for details.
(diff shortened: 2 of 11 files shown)


runtime/doc/evim.1
```diff
@@ -28,14 +28,18 @@ See vim(1) for details about Vim, options, etc.
 The 'insertmode' option is set to be able to type text directly.
 .br
 Mappings are setup to make Copy and Paste work with the MS-Windows keys.
-CTRL-X cuts text, CTRL-C copies text and CTRL-V pastes text.
-Use CTRL-Q to obtain the original meaning of CTRL-V.
+CTRL\-X cuts text, CTRL\-C copies text and CTRL\-V pastes text.
+Use CTRL\-Q to obtain the original meaning of CTRL\-V.
```
chrisbra: why those added backslashes?

runtime/doc/vimdiff.1
```diff
@@ -1,4 +1,4 @@
-.TH VIMDIFF 1 "2001 March 30"
+.TH VIMDIFF 1 "2021 June 13"
```
chrisbra, line 1: why just the date change here?

runtime/doc/evim.1
RestorerZ: @chrisbra 
Christian, I think the keyboard shortcut in the documentation is not the same as a phrase like "long-term". It's possible that it's not.

> The hyphen-minus	is a particularly  unfortunate	case  of  overloading.
>        Its awkward name	in ISO 8859 and	later standards	reflects the many dis-
>        tinguishable  purposes  to  which it had	already	been put by the	1980s,
>        including a hyphen, a minus sign, and (alone or in  repetition)	dashes
>        of varying widths.  For best results in roff systems, use the "-" char-
>        acter  in input outside an escape sequence only to mean a hyphen, as in
>        the phrase "long-term".	For a minus sign in running  text  or  a  Unix
>        command-line option dash, use \\-	(or \\[-] in groff if you find it helps
>        the  clarity  of	the source document).  (Another	minus sign, for	use in
>        mathematical equations, is available as \\[mi]).	AT&T  troff  supported
>        em-dashes as \\(em, as does groff.

https://man.archlinux.org/man/groff_char.7.en

runtime/doc/vimdiff.1
RestorerZ: I based it on the date of the last change ("commit"). It is possible that they just forgot to change the date.
https://github.com/vim/vim/commit/2346a6378483c9871016f9fc821ec5cbea638f13#diff-2682a2bf53e9ab4a0692e5c69970484396e3f1a06f0e2be4c615a845d47c1a60

runtime/doc/evim.1
chrisbra: Hm, it makes sense, but I would argue that the `-` in `Ctrl-C` is not meant to be a minus sign, but rather a hyphen.

runtime/doc/vim.1
```diff
@@ -323,13 +296,23 @@ When N is omitted, open one window for each file.
 Open N tab pages.
 When N is omitted, open one tab page for each file.
 .TP
-\-P {parent-title}
+\-P {parent\-title}
 Win32 GUI only: Specify the title of the parent application.  When possible, Vim
-will run in an MDI window inside the application. {parent-title} must appear in
+will run in an MDI window inside the application. {parent\-title} must appear in
```
chrisbra: same here, it should be a hyphen, not a minus sign.

runtime/doc/vim.1
```diff
@@ -362,6 +335,13 @@ The same can be done with the command ":source! {scriptin}".
 If the end of the file is reached before the editor exits, further characters
 are read from the keyboard.
 .TP
+\-S {file}
+{file} will be sourced after the first file has been read.
+This is equivalent to \-c "source {file}".
+{file} cannot start with '\-'.
+If {file} is omitted "Session.vim" is used (only works when \-S is the last
+argument).
+.TP
```
chrisbra, line 344: here it might be okay, at least it seems like it was like that before as well

runtime/doc/vim.1
```diff
@@ -500,24 +499,27 @@ Use {name} as the server name.  Used for the current Vim, unless used with a
...
+Win32 GUI only: Make gVim try to use the window {id} as a parent, so that it
 runs inside that window.
 .SH ON-LINE HELP
 Type ":help" in
 .B Vim
 to get started.
 Type ":help subject" to get help on a specific subject.
 For example: ":help ZZ" to get help for the "ZZ" command.
-Use <Tab> and CTRL-D to complete subjects (":help cmdline\-completion").
+Use <Tab> and CTRL\-D to complete subjects (":help cmdline\-completion").
```
chrisbra: I think is should be hyphen here.

runtime/doc/vimdiff.1
chrisbra: Okay I see

runtime/doc/xxd.1
```diff
@@ -244,50 +246,50 @@ Print 120 bytes as a continuous hex dump with 20 octets per line.
...
+64220a2e5c220a2e5c222032317374204d617920
 .br
-617920313939360a2e5c22204d616e2070616765
+313939360a2e5c22204d616e2070616765206175
 .br
-20617574686f723a0a2e5c2220202020546f6e79
+74686f723a0a2e5c2220202020546f6e79204e75
 .br
-204e7567656e74203c746f6e79407363746e7567
+67656e74203c746f6e79407363746e7567656e2e
```
chrisbra, line 262: Thanks. I think the above is just an example and it is clear, that it will be different once the manpage changes. I don't think we need to keep the hex result of the `xxd.1` manpage in sync with the example here. But thanks, let's keep it now.

runtime/doc/evim.1
RestorerZ: Yes... My poor knowledge of English typography let me down, of course. I meant
"Unix command-line option dash" rather than a minus sign.
Thanks for the pointer. I've fixed it.

runtime/doc/xxd.1
RestorerZ: Well, when I read the documentation, I try to check if I understood everything
correctly by following the examples given. And I would be surprised if what I
received did not match what was specified in the documentation.
But I agree with you, it's quite troublesome to change it every time.

runtime/doc/vim.man
```diff
@@ -36,11 +36,11 @@ DESCRIPTION
             vim [options] [filelist]
 
        If the filelist is missing, the editor will start with an empty buffer.
-       Otherwise  exactly  one out of the following four may be used to choose
+       Otherwise exactly one out of the following four may be used  to  choose
```
benknoble, line 39: The spacing changes look odd to me: I’m not terribly familiar with roff semantics, but it seems like we removed some instances of 2 spaces between words but then added new instances throughout this file?

If one space in roff means an inter-word space, let’s stick to consistently one space between words.

(I guess it’s also possible it’s a different kind of invisible character I can’t see; a remark on why would be helpful for me in that case.)

chrisbra: I noticed those to. However I think those `.man` files are generated automatically using the Makefile (and probably some vimscript vodoo. So I am not too worried about it.

runtime/doc/xxd.1
chrisbra: Understood. Let's add a note, that this is the result as of the current manpage as of 2024 or something similar.

runtime/doc/vim.man
RestorerZ: I was trying to figure this out too, why text with different spacing is generated each time.
The command that is written in Makefie is as follows:
`nroff -man $< | sed -e s/.//g > $@`.
The pattern contains a tab character.
Started reading the [documentation](https://heirloom.sourceforge.net/doctools/troff.pdf), realised that it is possible, probably can be configured, but I no longer have enough time and patience.

benknoble: Sorry, missed the auto-generation! Thanks both. (I think you meant a backspace, not tab, but close enough for these purposes.)

RestorerZ: Yes, of course. Thank you.

runtime/doc/xxd.1
RestorerZ: Christian, added such a note:
The results of the examples below are relevant to the xxd.1 man page as of May 2024

chrisbra:
thanks, looks good now

k-takata:
The address is now 8 digits. These parts should also be updated.

```diff
--- a/runtime/doc/xxd.1
+++ b/runtime/doc/xxd.1
@@ -311,7 +311,7 @@ Patch the date in the file xxd.1
 .br
 \fI% xxd \-s 0x33 \-l 13 \-c 13 xxd.1\fR
 .br
-0000033: 3235 7468 204d 6179 2031 3939 36  25th May 1996
+00000033: 3235 7468 204d 6179 2031 3939 36  25th May 1996
 .PP
 .br
 Create a 65537 byte file with all bytes 0x00,
@@ -324,11 +324,11 @@ Hex dump this file with autoskip.
 .br
 \fI% xxd \-a \-c 12 file\fR
 .br
-0000000: 0000 0000 0000 0000 0000 0000  ............
+00000000: 0000 0000 0000 0000 0000 0000  ............
 .br
 *
 .br
-000fffc: 0000 0000 41                   ....A
+0000fffc: 0000 0000 41                   ....A
 .PP
 Create a 1 byte file containing a single 'A' character.
 The number after '\-r \-s' adds to the line numbers found in the file;
```

PR #15493, opened 2024-08-14
runtime(doc): portable Makefile for GNU and BSD

RestorerZ:
Continuation: #15471 
In the Makefile from "runtime/doc", the variable `$<` has been replaced with `$?` in explicit rules.

### DIFF
--- a/runtime/doc/Makefile
+++ b/runtime/doc/Makefile
@@ -157,61 +157,61 @@ os_win32.txt:
 # Especially since there is only one dependency and it should presumably always
 # be newer than the target file.
 vim-da.UTF-8.1: vim-da.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vimdiff-da.UTF-8.1: vimdiff-da.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vimtutor-da.UTF-8.1: vimtutor-da.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vim-de.UTF-8.1: vim-de.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vim-fr.UTF-8.1: vim-fr.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 evim-fr.UTF-8.1: evim-fr.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vimdiff-fr.UTF-8.1: vimdiff-fr.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vimtutor-fr.UTF-8.1: vimtutor-fr.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 xxd-fr.UTF-8.1: xxd-fr.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vim-it.UTF-8.1: vim-it.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 evim-it.UTF-8.1: evim-it.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
 
 vimdiff-it.UTF-8.1: vimdiff-it.1
-	iconv -f ISO-8859-1 -t UTF-8 $? >$@
+	iconv -f latin1 -t UTF-8 $? >$@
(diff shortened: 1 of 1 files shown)


runtime/doc/Makefile
```diff
@@ -151,88 +151,91 @@ os_risc.txt:
 os_win32.txt:
 	touch $@
 
-# Note that $< works with GNU make while $> works for BSD make.
-# Is there a solution that works for both??
+# In *BSD, the variable ‘$<’ is used in suffix-transformation rules (in GNU this
```
chrisbra: Can you please use normal single apostrophes? Not `U+2018 LEFT SINGLE QUOTATION MARK`?

runtime/doc/Makefile
```diff
@@ -151,88 +151,91 @@ os_risc.txt:
...
+# is called "implicit rules", and in MS Windows it is called "inference rules").
+# For code portability, it is better to use the ‘$?’ variable for explicit rules.
+# Especially since there is only one dependency and it should presumably always
+# be newer than the target file.
 vim-da.UTF-8.1: vim-da.1
-	iconv -f latin1 -t utf-8 $< >$@
+	iconv -f ISO-8859-1 -t UTF-8 $? >$@
```
chrisbra: Please don't change the encoding names, unless there is a specific reason for.

runtime/doc/Makefile
RestorerZ: Fixed it. I'm sorry, please.

runtime/doc/Makefile
RestorerZ: Well, it'll work both ways.
But I've always thought that unless there are some special reasons, it's better to stick to some standards. Of course, standards can be bypassed, extended or ignored altogether, but it is better to do it in special cases.
And in this case, the name ISO-8859-1 is the preferred official name. Other names are also official, but they are only aliases.

> ISO-8859-1 is the IANA preferred name for this standard when supplemented with the C0 and C1 control codes from ISO/IEC 6429. The following other aliases are registered: iso-ir-100, csISOLatin1, latin1, l1, IBM819. Code page 28591 a.k.a. Windows-28591 is used for it in Windows.[8] IBM calls it code page 819 or CP819 (CCSID 819).[9][10][11][12] Oracle calls it WE8ISO8859P1.[13]

https://en.wikipedia.org/wiki/ISO/IEC_8859-1

> ISO 8859-1 is also known as Latin-1.

[iso_8859-1(7)](https://manpages.org/iso_8859-1/7)

> 8859-1 (Latin-1)

[charsets(7)](https://manpages.org/charsets/7)

chrisbra: I know, that those are aliases, but I doubt they will go away ever. The current encoding names are valid (even if they are just aliases).

RestorerZ: I am sure they will be used even after you and I are gone.

RestorerZ: That's the other thing I don't understand. Both you and I agree that these are official, acceptable and used names. You and I both know that it will work both ways. What's the petra scandali then?

ychin: I would imagine readability is at least one good reason to use "latin1". Sometimes these aliases exist because "latin1" is a much easier to understand and remember for most people. Even for a widely used encoding not a lot of people are going to remember what a random number like "8859" means. Code isn't just for computers. They are also for humans. Also, I feel like as a general philosophy, changes should be for the better. If it's a completely sideways (in terms of proposed value) change I think the burden should be to show how it improves things, as changes always introduce some chances of issues or bugs even if unforeseen.

Also, mixing intentions or drive-by changes in a PR is always non-ideal. The PR says this is to make Makefile compatible in BSD and GNU but changing the encoding name does not do that. It's better to create separate PRs for different types of cleanups. E.g. if this causes problems and need reverts, etc, it would be harder to isolate the issue.

Just my 2c.

RestorerZ: @ychin 
Thank you for your five kopeck.

> Also, mixing intentions or drive-by changes in a PR is always non-ideal. The PR says this is to make Makefile compatible in BSD and GNU but changing the encoding name does not do that. It's better to create separate PRs for different types of cleanups. E.g. if this causes problems and need reverts, etc, it would be harder to isolate the issue.

Yes, I agree with you completely. This is unfortunately my great misfortune. Sometimes when you fix one thing, you tend to fix something else. Even if it is not so significant, but nevertheless it improves the general code at least a little.
I'm honestly struggling with this. And I hope I will be able to cope with this misfortune! ;-)

> I would imagine readability is at least one good reason to use "latin1". Sometimes these aliases exist because "latin1" is a much easier to understand and remember for most people. Even for a widely used encoding not a lot of people are going to remember what a random number like "8859" means. Code isn't just for computers. They are also for humans. Also, I feel like as a general philosophy, changes should be for the better. If it's a completely sideways (in terms of proposed value) change I think the burden should be to show how it improves things, as changes always introduce some chances of issues or bugs even if unforeseen.

I agree with you that All Humans, myself included, are more accustomed to symbols than to numbers. And it is easier for people to operate with alphabetic information than with digital information. That's the way we like it...

But the thing is, “latin1” doesn't really reflect the ISO-8859-1 code table.
If you go to the primary sources, ISO-8859-1 has four parts:
[C0 and US-ASCII](http://www.unicode.org/charts/PDF/U0000.pdf) - control codes and basic Latin, plus [C1 and latin1](http://www.unicode.org/charts/PDF/U0080.pdf) - additional control codes and extended Latin and punctuation. These four parts are ISO-8859-1.
I'm not really a pedant or a nerd, I'm more in favor of unification and uniformity.
But if everyone uses “latin1”, then let it be “latin1”, as long as everyone understands what we are talking about.  

In the end, “Everyone uses latin1 because everyone else is too...”  

And I hope that the esteemed Christian will pardon me for some possible hotheadedness.
At the end of the day, we are all here trying to make our product, which we love and work with every day, a little bit better and more convenient. So let's just call it a day.

PR #15497, opened 2024-08-15
Preserve substitute pattern with :keeppatterns

gpanders:
**Problem**: The `:keeppatterns` modifier does not preserve the last substitute pattern, so `:s/foo/bar`, `:keeppatterns s/asdf/xyz` followed by `:&` will replace `foo` with `xyz` (not `bar`). This makes it impossible for plugins to use a substitute command without modifying the user's own substitute pattern history.

**Solution**: Make `:keeppatterns` also preserve the substitute pattern.

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -379,7 +379,7 @@ terminals)
 
 :keepp[atterns] {command}			*:keepp* *:keeppatterns*
 		Execute {command}, without adding anything to the search
-		history
+		history or modifying the last substitute pattern.
 
 ==============================================================================
 2. Command-line completion				*cmdline-completion*
--- a/src/ex_cmds.c
+++ b/src/ex_cmds.c
@@ -3777,6 +3777,7 @@ ex_substitute(exarg_T *eap)
     int		endcolumn = FALSE;	// cursor in last column when done
     pos_T	old_cursor = curwin->w_cursor;
     int		start_nsubs;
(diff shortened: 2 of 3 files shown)


zeertzjq:
Should it also preserve this?
```
  ~	  \~	  replaced with the {string} of the previous
		  substitute					     *s~*
```

chrisbra:
I don't think so. `:keeppatterns` is meant to not store anything in the search history. For `~` we don't even have a separate history. I think the PR is fine as is. Thanks.

zeertzjq:
Right, `~` is a replacement, not a pattern, and making `:keeppatterns` preserve `~` may break scripts executed under `:keeppatterns` with no easy way to restore the old behavior.

runtime/doc/cmdline.txt
```diff
@@ -379,7 +379,7 @@ terminals)
 
 :keepp[atterns] {command}			*:keepp* *:keeppatterns*
 		Execute {command}, without adding anything to the search
-		history
+		history or modifying the last substitute pattern.
```
zeertzjq, line 382: This should be "substitute string" (as mentioned in `:h :&`), not "substitute pattern", but then it's still not accurate as this only keeps the last substitute string for `:&`, not for `~`, so maybe it should be `last substitute string for |:&|`.

chrisbra: thanks, I adjusted it

Houl:
"substitute pattern" is really confusing when you mean the replacement part, aka "substitute string".

PR #15498, opened 2024-08-15
translation(ja): Update to the latest release

k-takata:
https://github.com/vim-jp/lang-ja/releases/tag/20240815


chrisbra:
thanks!

PR #15499, opened 2024-08-15
Changed pattern for comments '#.' in cleanup.vim script.

RestorerZ:
Problem: Comments like '#.' are not deleted if they are followed by an
end-of-line character.
Solution: Removed the space in the `global` command search pattern `#\. `.

Other. Added a register "black hole" for the `d` operator.


chrisbra:
thanks!

PR #15502, opened 2024-08-15
runtime(tutor): portable Makefile for GNU and BSD

RestorerZ:
Continuation: #15493 
In the Makefile from "runtime/tutor", the variable `$<` has been replaced with `$?` in explicit rules.


chrisbra:
thanks

PR #15503, opened 2024-08-15
runtime(lang): portable Makefile for GNU and BSD

RestorerZ:
Continuation: #15502 
In the Makefile from "runtime/lang", the variable `$<` has been replaced with `$?` in explicit rules.


chrisbra:
thanks!

PR #15504, opened 2024-08-16
Rename from w_closing to w_locked is incomplete

zeertzjq:
Problem:  Rename from w_closing to w_locked is incomplete
          (after 9.1.0678).
Solution: Rename remaining occurrences of w_closing to w_locked and
          update comments.


chrisbra:
thanks makes sense. I think it was initially used to mark windows to be closed, but apparently later changed the meaning to mark windows as being locked, so I renamed it to make it obvious.

PR #15511, opened 2024-08-16
runtime(vim): Update base-syntax, improve :let-heredoc highlighting

dkearns:
The end marker must appear on line of its own without any trailing whitespace.
    
Whitespace is incorrectly allowed before all end markers.  Limiting this only to heredocs where "trim" was specified, and with the correct indent, is currently an intractable problem given that contained syntax groups (in this case :let) cannot be limited to start patterns.

Highlight interpolated expressions when "eval" is specified.


chrisbra:
thanks!

chrisbra:
included as of d164f2a521f8e5

PR #15476, opened 2024-08-11
runtime(syntax-tests): Facilitate the viewing of rendered screendumps

zzzyxwvut:
With the submitted `viewdumps.vim` script, a few manual  
steps in typical workflows (see below) can be automated.  
The updated `README.txt` contains additional information.

------------------------------------------------------------

Reviewing LOCAL failed syntax tests can be arranged as  
follows:

1) Run tests and generate screendumps:

```shell
cd /path/to/fork/runtime/syntax
make clean test
```

2) Examine the screendumps from the `failed` directory:

```shell
../../src/vim --clean -S testdir/viewdumps.vim
```

------------------------------------------------------------

Reviewing UPLOADED failed syntax tests can be arranged as  
follows (it can be further locally scripted):

1) Fetch an artifact with failed screendumps from
`github.com/vim/vim/actions/runs/A_ID/artifacts/B_ID`.

2) Extract the archived files:

```shell
unzip /tmp/failed-tests.zip -d /tmp
```

3) Set up the `dumps` directory.  Create a symlink to  
`/path/to/fork/dirs/dumps` in the extracted directories so  
that `term_dumpdiff()` can be used.  (The lookup algorithm  
resolves `dumps` for every loaded filename.  So, with  
`/tmp/runtime/syntax/testdir/failed/*.dump` files passed  
as script arguments, the algorithm will make the files in  
`/tmp/runtime/syntax/testdir/dumps` queried.)

```shell
cd /path/to/fork
ln -s $(pwd)/runtime/syntax/testdir/dumps \
  /tmp/runtime/syntax/testdir/dumps
```

4) Examine the extracted screendumps:

```shell
./src/vim --clean -S runtime/syntax/testdir/viewdumps.vim \
  /tmp/runtime/syntax/testdir/failed/*.dump
```

5) Clean up:

```shell
unlink /tmp/runtime/syntax/testdir/dumps
rm -rf /tmp/runtime
```

------------------------------------------------------------

Reviewing SUBMITTED FOR PULL REQUEST syntax tests can be  
arranged as follows (it can be further locally scripted):

1) List the fetched changeset and write the changed `dumps`  
filenames to `/tmp/filelist`:

```shell
cd /path/to/fork
git switch prs/1234
git diff-index --relative=runtime/syntax/testdir/dumps/ \
  --name-only prs/1234~1 > /tmp/filelist
```

2) Reconcile relative filepaths, and copy next-to-be-updated  
`dumps` files in the `failed` directory (note the missing  
[...]

PR #15515, opened 2024-08-16
Facilitate the viewing of rendered screendumps under `src/`

zzzyxwvut:
With the submitted `viewdumps.vim` script, a few manual  
steps in typical workflows (see below) can be automated.  
The updated `README.txt` contains additional information.

------------------------------------------------------------

Reviewing LOCAL failed screendump tests can be arranged as  
follows:

1) Run tests and generate screendumps:

```shell
cd /path/to/fork/src/testdir
make
```

2) Examine the screendumps from the `failed` directory:

```shell
../vim -u NONE -S viewdumps.vim
```

------------------------------------------------------------

Reviewing UPLOADED failed screendump tests can be arranged  
as follows (it can be further locally scripted):

1) Fetch an artifact with failed screendumps from
`github.com/vim/vim/actions/runs/A_ID/artifacts/B_ID`.

2) Extract the archived files:

```shell
unzip /tmp/failed-tests.zip -d /tmp
```

3) Set up the `dumps` directory.  Create a symlink to  
`/path/to/fork/dirs/dumps` in the extracted directories so  
that `term_dumpdiff()` can be used.  (The lookup algorithm  
resolves `dumps` for every loaded filename.  So, with  
`/tmp/src/testdir/failed/*.dump` files passed as script  
arguments, the algorithm will make the files in  
`/tmp/src/testdir/dumps` queried.)

```shell
cd /path/to/fork
ln -s $(pwd)/src/testdir/dumps /tmp/src/testdir/dumps
```

4) Examine the extracted screendumps:

```shell
./src/vim -u NONE -S src/testdir/viewdumps.vim \
  /tmp/src/testdir/failed/*.dump
```

5) Clean up:

```shell
unlink /tmp/src/testdir/dumps
rm -rf /tmp/src
```

------------------------------------------------------------

Reviewing SUBMITTED FOR PULL REQUEST screendump tests can be  
arranged as follows (it can be further locally scripted):

1) List the fetched changeset and write the changed `dumps`  
filenames to `/tmp/filelist`:

```shell
cd /path/to/fork
git switch prs/1234
git diff-index --relative=src/testdir/dumps/ \
  --name-only prs/1234~1 > /tmp/filelist
```

2) Reconcile relative filepaths, and copy next-to-be-updated  
`dumps` files in the `failed` directory (note the missing
new screendumps, if any):

```shell
git switch master
cd src/testdir/dumps
test -d ../failed || mkdir ../failed
cp -t ../failed $(cat /tmp/filelist)
```

3) Remember about the introduced INVERTED relation between  
`dumps` and `failed`, i.e. the files to be committed are in  
`dumps` already and their previous versions are in `failed`;  
therefore, copy the missing new screendumps from `dumps` to  
`failed` (otherwise these won't be shown):

```shell
git switch prs/1234
cp -t ../failed foo_10.dump foo_11.dump foo_12.dump
```

4) Examine the screendumps from the `failed` directory (new  
screendumps will be shown with no difference between their  
versions):

```shell
cd ..
../vim -u NONE -S viewdumps.vim
```

### DIFF
--- a/Filelist
+++ b/Filelist
@@ -185,6 +185,7 @@ SRC_ALL =	\
 		src/testdir/sautest/autoload/*.vim \
 		src/testdir/testluaplugin/lua/testluaplugin/*.lua \
 		src/testdir/check.vim \
+		src/testdir/commondumps.vim \
 		src/testdir/gui_init.vim \
 		src/testdir/gui_preinit.vim \
 		src/testdir/mouse.vim \
--- a/Filelist
+++ b/Filelist
@@ -198,6 +198,7 @@ SRC_ALL =	\
 		src/testdir/summarize.vim \
 		src/testdir/term_util.vim \
 		src/testdir/view_util.vim \
+		src/testdir/viewdumps.vim \
 		src/testdir/test[0-9]*.ok \
 		src/testdir/test77a.ok \
 		src/testdir/test77a.com \
--- a/runtime/syntax/testdir/README.txt
+++ b/runtime/syntax/testdir/README.txt
@@ -124,8 +124,8 @@ verify that the tests fail.  Then you know your changes are covered by the
 test.
 
 
-Viewing generated screendumps
------------------------------
+Viewing generated screendumps (local)
+-------------------------------------
 
 You may also wish to look at the whole batch of failed screendumps after
 running "make test".  Source the "viewdumps.vim" script for this task:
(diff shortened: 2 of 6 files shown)


zzzyxwvut:
The limitation for both [the previously included script](https://github.com/vim/vim/pull/15476) and  
for the proposed script is that neither is capable, at the  
same time, of using `term_dumpdiff()` for arbitrary directory  
sub-trees _and_ enjoying the support of being sourceable from  
arbitrary directories that lie in the direct path from the  
project's root directory towards either script locations.  
In other words, `syntax/testdir/viewdumps.vim` can only use  
`term_dumpload()` for `src/` screendumps and, conversely,  
`src/testdir/viewdumps.vim` can only use `term_dumpload()`  
for `syntax/` screendumps.  Hence, we need a script available  
for each `dumps` directory.

As AN ALTERNATIVE, there can be a single script placed in  
the root directory, offering some configuration FLAG to tell  
where to look for directories.  But this remoteness from  
tracked `dumps` directories would make it more cumbersome to  
use for the common case of inspecting local `make` failures,  
i.e. instead of current (and proposed) paths:

```
cd /path/to/fork/runtime/syntax
make clean test
../../src/vim --clean -S testdir/viewdumps.vim

cd /path/to/fork/src/testdir
make
../vim -u NONE -S viewdumps.vim
```

paths to scripts would have to be amended and a FLAG added:

```
cd /path/to/fork/runtime/syntax
make clean test
../../src/vim --clean -S ../../viewdumps.vim FLAG

cd /path/to/fork/src/testdir
make
../vim -u NONE -S ../../viewdumps.vim FLAG
```

zzzyxwvut:
Note that the CI-related workflow (and the PR-related one)  
mentioned here:

https://github.com/vim/vim/blob/c6ed816761f44da92d8c61f5ea6cb7fdbb45fac5/src/testdir/README.txt#L123-L128

will be found in the commit message *not* in `README.txt`.

Do we need those workflow instructions (CI and PR) copied  
from commit message bodies and separately documented?

chrisbra:
I am confused. Do we need `runtime/syntax/testdir/viewdumps.vim` and `src/testdir/viewdumps.vim` now?  Wouldn't it make more sense to refactor the common code into a common script and source it?

zzzyxwvut:
Do you want me to refactor common code into a third script  
and then source it for each `viewdumps.vim`?  Where would it  
be better to put this common script?

chrisbra:
I would place it into `src/testdir/` since that is where most of the current test functions live. Not sure if it makes sense to go into vim9.vim 

At some time, we should probably start to move common test utility vim scripts to it's own directory and `test_*.vim` functions to another one. It feels a bit dirty currently and it makes it hard to find the correct files.

zzzyxwvut:
Okay. What about documenting CI and PR workflows?

chrisbra:
Yes, I think it makes sense to document them in a README. The commit message is not so easily visible (and findable).

zzzyxwvut:
Now that I think about sourcing...  The current support for  
any working directory would not likely survive this. :(

zzzyxwvut:
How about `:runtime`?

zzzyxwvut:
The requested changes have been made.

PR #15516, opened 2024-08-17
Refactor usercmd.c to remove calls to STRLEN()

basilisk0315:
Also use bsearch() to perform some lookups. No dramatic improvements, but every bit helps.

### DIFF
--- a/src/usercmd.c
+++ b/src/usercmd.c
@@ -468,7 +468,6 @@ get_user_cmd_complete(expand_T *xp UNUSED, int idx)
     return (char_u *)command_complete_tab[idx].value;
 }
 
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. I was a bit worried with the `produce_cmdmods()` function and we would eventually run into the same issue as #15305 which commit 70a11a6bf69f477844470ce5995 fixed. 

I tried to reproduce using the following `:command`:

```
:com Echo echo "<mods>"
:kee keep keepm keepma keepmar keepmarks keepa keepalt keepjumps keeppatterns Echo
keepalt keepjumps keepmarks keeppatterns
```
which seems fine. So this seems fine to include.

PR #15519, opened 2024-08-17
runtime(zip): small fixes for zip plugin

### LINKED ISSUE #15445: Zip plugin does not handle inner file name contain glob character
e.g. if you create a zip file that contains a file with name `[a].txt` then the unzip command will look like

```bash
unzip -p -- a.zip \[a\].txt
```

to which `unzip` command will see its argument as `[a].txt` and perform another level of glob matching, resulting in no match.

In my version, it suffices to escape the characters `\[?*`. Escaping `]` appears to be unnecessary but I see no harm either.

```diff
--- a/runtime/autoload/zip.vim
+++ b/runtime/autoload/zip.vim
@@ -234,7 +234,7 @@ fun! zip#Read(fname,mode)
   " but allows zipfile://... entries in quickfix lists
   let temp = tempname()
   let fn   = expand('%:p')
-  exe "sil! !".g:zip_unzipcmd." -p -- ".s:Escape(zipfile,1)." ".s:Escape(fname,1).' > '.temp
+  exe "sil! !".g:zip_unzipcmd." -p -- ".s:Escape(zipfile,1)." ".s:Escape(escape(fname, '\?*[]'),1).' > '.temp
   sil exe 'keepalt file '.temp
   sil keepj e!
   sil exe 'keepalt file '.fnameescape(fn)
```

Not sure if it works on Windows/other platforms/implementations of `unzip`.

A small Python script for testing  [a.zip](https://github.com/user-attachments/files/16530612/a.zip)

chrisbra:
This does the following:
- verify the unzip plugin is executable when loading the autoload plugin
- handle extracting file names with '[' in its name correctly by escaping '[' by '[[]' for the unzip command (and adding a test for it)
- verify, that the extract plugin is not overwriting a file (could cause a hang, because unzip asking for confirmation)

closes: #15445


runtime/autoload/zip.vim
```diff
@@ -331,9 +337,24 @@ fun! zip#Extract()
    call s:Mess('Error', "***error*** (zip#Extract) Please specify a file, not a directory")
    return
   endif
+  if filereadable(fname)
+   call s:Mess('Error', "***error*** (zip#Extract) <" .. fname .."> already exists in directory, not overwriting!")
+   return
+  endif
+  let target = fname->substitute('\[', '[[]', 'g')
+  if &shell =~ 'cmd' && (has("win32") || has("win64"))
```
k-takata, line 345: (trivial) This is redundant. This can be:
```vim
  if &shell =~ 'cmd' && has("win32")
```

chrisbra: I'll fix it up, thanks

PR #15520, opened 2024-08-17
VMS does not have defined uintptr_t

arpadffy:
There are systems where the uintptr_t is not defined.
This pull request fixes the issue on VMS systems, but I guess there will be added some other exotic architectures that has the very same problem.

Thanks.


chrisbra:
thanks!

PR #15521, opened 2024-08-18
Better typedchar documentation in KeyInputPre autocmd

Shougo:
I have improved `typedchar` documentation.  Because it may be empty.


chrisbra:
thanks!

PR #15522, opened 2024-08-18
runtime(vim): Update base-syntax, improve script-interface command highlighting

dkearns:
Update base-syntax, improve script-interface command highlighting

- Normalise interface heredoc highlighting with that used for `:let-heredocs`.
- Remove interface feature testing.  The Lua and Python interface command scripts are now highlighted by default.  Loading all syntax files incurs an undesirable load-time burden so highlighting of the less popular MzScheme, Perl, Ruby and Tcl interfaces is disabled by default.  `g:vimsyn_embed` can still be used to customise the supported interfaces.
- Always highlight interface ex-commands as valid commands, even when the corresponding command-script highlighting is disabled.
- Remove error highlighting of heredoc and statement command-script regions when an interface is disabled.  These are now highlighted as plain text.
- Highlight simple command-script statement strings as well as heredocs.
- Allow indented heredoc end tokens when `trim` is specified.
- Match interface heredocs in `:def` functions.
- Fix runaway `vimEmbedError` regions.  These regions have been removed.
- Use python2 syntax for `:python`, and `:pythonx` when `'pyxversion'` is appropriately set.


runtime/syntax/generator/vim.vim.base
```diff
@@ -862,218 +863,266 @@ syn region	vimHiLink	contained matchgroup=Type start="\%(\<hi\%[ghlight]!\=\s\+\
...
-"   g:vimsyn_embed =~# 'p' : embed Perl     (but only if vim supports it)
-"   g:vimsyn_embed =~# 'P' : embed Python   (but only if vim supports it)
-"   g:vimsyn_embed =~# 'r' : embed Ruby     (but only if vim supports it)
-"   g:vimsyn_embed =~# 't' : embed Tcl      (but only if vim supports it)
+"   g:vimsyn_embed =~# 'l' : embed Lua
+"   g:vimsyn_embed =~# 'm' : embed MzScheme
+"   g:vimsyn_embed =~# 'p' : embed Perl
+"   g:vimsyn_embed =~# 'P' : embed Python
+"   g:vimsyn_embed =~# 'r' : embed Ruby
+"   g:vimsyn_embed =~# 't' : embed Tcl
```
k-takata: I'm wondering if we need to check if vim support the script.
I sometimes want to enable the highlight even if the current vim doesn't support it.

dkearns: Yes, I like to highlight all script commands all the time so I changed it to only highlight supported scripts by default but any user-set value of g:vimsyn_embed overrides that.  I think this behaviour matches the current documentation.

I don't think validating that a value in a user-set g:vimsyn_embed is actually supported is particularly useful.  Do you?

k-takata: Ah, I misunderstood your change.
Looks good.

runtime/syntax/generator/vim.vim.base
```diff
@@ -862,218 +863,266 @@ syn region	vimHiLink	contained matchgroup=Type start="\%(\<hi\%[ghlight]!\=\s\+\
...
+
 " [-- tcl --] {{{3
 if has("win32") || has("win95") || has("win64") || has("win16")
  " apparently has("tcl") has been hanging vim on some windows systems with cygwin
- let s:trytcl= (&shell !~ '\<\%(bash\>\|4[nN][tT]\|\<zsh\)\>\%(\.exe\)\=$')
+ let s:trytcl = (&shell !~ '\<\%(bash\>\|4[nN][tT]\|\<zsh\)\>\%(\.exe\)\=$')
 else
- let s:trytcl= 1
+ let s:trytcl = 1
 endif
```
k-takata: I'm wondering if we still need this part.
Trying to use Cygwin version of Tcl from Win32 version of Vim might cause hanging long time ago. But nowadays, Cygwin DLLs have the cyg prefix, so I don't think it can happen.

dkearns: OK, I'll remove it.

k-takata:
We have "python_compiled" and "python3_compiled", but we don't have "ruby_compiled" or other variants.

I wonder if there are any benefits by disabling some of the scripts... performance?
How about enabling all of them by default and removing `has()`? `"lmpPrt"`

chrisbra:
If we restrict it to lua and Python, what is the load penalty then?

chrisbra:
unless you disagree, we seem to have found a good compromise then ?

chrisbra:
thanks

zzzyxwvut:
@dkearns, locally running:

```sh
cd runtime/syntax
time testdir/tools/regenerate_screendumps.sh
```

reports that there are no matching syntax tests for any of   
`vim_ex_python3_*.dump` and `vim_ex_pythonx_*.dump` files.

PR #15529, opened 2024-08-20
refine pandoc compiler settings

Konfekt:
Make some settings configurable, just in case, and be less aggressive setting output document title, in particular if provided in source buffer

### DIFF
--- a/runtime/compiler/pandoc.vim
+++ b/runtime/compiler/pandoc.vim
@@ -25,30 +25,32 @@ let s:supported_filetypes =
 
 silent! function s:PandocFiletype(filetype) abort
   let ft = a:filetype
-  if ft ==# 'pandoc'
-    return 'markdown'
-  elseif ft ==# 'tex'
-    return 'latex'
-  elseif ft ==# 'xml'
-    " Pandoc does not support XML as a generic input format, but it does support
-    " EndNote XML and Jats XML out of which the latter seems more universal.
-    return 'jats'
-  elseif ft ==# 'text' || empty(ft)
-    return 'markdown'
-  elseif index(s:supported_filetypes, &ft) >= 0
-      return ft
+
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

chrisbra:
Hm, there seems to be a few issues with those changes. The makeprg setting does not include the `--standalone` and the errorformat uses invalid syntax. I think I fixed those, please verify with what I merge down. Thanks

Konfekt:
Hm... the errorformat did not change; however, I found 

```
#  /usr/local/share/vim/vim91/doc/quickfix.txt (line 1599)
:set efm=%C\ %.%#,%A\ \ File\ \"%f\"\\,\ line\ %l%.%#,%Z%[%^\ ]%\\@=%m
```

So I suppose `\"%f\"` was intended. The makeprg still included `--standalone`? In any event, thank you for adjusting the change dates, which I forgot about.

chrisbra:
- Ah perhaps those double quotes should have been escaped? Without those escapes, the local errorformat was empty, I think because the `"` was detected as an comment sign.

- With your version of the patch, the `--standalone` went missing. I assume this needs to be there right? I had to add an extra pair of parenthesis around this part here:
```
empty(b:pandoc_compiler_lang) ?
      \ '' : ' --metadata lang='..b:pandoc_compiler_lang .
      \ ' --from='..b:pandoc_compiler_from .
      \ ' '..get(b:, 'pandoc_compiler_args', get(g:, 'pandoc_compiler_args', ''))
```
otherwise, the `--standalone` got lost

Konfekt:
Yes, exactly. I can reproduce and found the same fix (that I don't fully understand, but in general the more parentheses, the merrier for the computer, at least). Thank you for spotting this. So the only thing missing is escaping the double quotes in errorformat.

PR #15533, opened 2024-08-20
Fix `mode()` returns wrong value at cmdline in visual mode

kuuote:
Problem: `mode()` returns wrong value at directly called commandline mode in visual mode
Solution: Change decision priority of `VIsual_active`


chrisbra:
thanks, makes sense

PR #15535, opened 2024-08-20
runtime:compiler:pandoc: escape quotes in &errorformat

Konfekt:
as discussed at https://github.com/vim/vim/pull/15529


chrisbra:
thanks

PR #15539, opened 2024-08-21
runtime(doc): fix getbufinfo() signature

jparise:
getbufinfo() expects a single buffer or dict, not a list.


k-takata:
These [] mean that they can be omitted.

jparise:
> These [] mean that they can be omitted.

Ugh, that's embarrassing! I somehow didn't make that connection after all of this time writing Vimscript. And it's a good sign I should go to bed.

Thank you!

PR #15544, opened 2024-08-21
translation(it): Fix filemode

k-takata:


chrisbra:
oh, did I miss the filemode again? Sorry for that.

PR #15549, opened 2024-08-21
runtime(java): Make optional the use of the bundled `&foldtext` function

zzzyxwvut:
- Obtain and pass through translated messages with this
  function.
- If `g:java_foldtext_show_first_or_second_line` is defined,
  assign this function to `&foldtext`.


Konfekt:
Thank you, this also made the need of a file type specific &foldtext clearer to me

chrisbra:
thanks!

PR #15550, opened 2024-08-22
runtime(netrw): Change line on `mx` if command output exists

KSR-Yasuda:
On external command execution by `mx` shortcut,
the command output is printed right after user input.

netrw_mx_20240822_161912

Shouldn't it change line?

#### How to Repro

```vim
" Open NetRW
:e ./

" Select some file in NetRW
mf

" Execute a command for the selected files
mx
file
" Then, the output of external command `file` follows this user input without line change.
```


chrisbra:
thanks. Let me also get rid of the trailing newline

KSR-Yasuda:
Thank you for merging and fixing!

PR #15552, opened 2024-08-22
runtime(syntax): fix missing syntax highlighting for html in htmlangular

### LINKED ISSUE #15459: htmlangular filetype no longer has syntax highlighting
### Steps to reproduce

Open any angular html file.
`:set filetype?` returns `htmlangular` and not `html.angular`

### Expected behaviour

File should have syntax highlighting, but there are no colors at all.

Is `htmlangular` now `html.angular`? I applied this patch and now it works.

```
diff --git a/runtime/autoload/dist/ft.vim b/runtime/autoload/dist/ft.vim
index c8942eb3d..522492953 100644
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -418,7 +418,7 @@ export def FThtml()
   while n < 40 && n <= line("$")
     # Check for Angular
     if getline(n) =~ '@\(if\|for\|defer\|switch\)\|\*\(ngIf\|ngFor\|ngSwitch\|ngTemplateOutlet\)\|ng-template\|ng-content\|{{.*}}'
-      setf htmlangular
+      setf html.angular
       return
     endif
     # Check for XHTML
```

### Version of Vim

9.1.0611 or higher

### Environment

Artixlinux, alacritty

### Logs and stack traces

_No response_

dlvandenberg:
Fixes #15459 

I did not have time to figure out how to do the Angular specific syntax highlighting. Is there any documentation on that part? Any help is appreciated.

At least the HTML highlighting should be fixed with this.

### DIFF
--- a/runtime/syntax/htmlangular.vim
+++ b/runtime/syntax/htmlangular.vim
@@ -1,0 +1,18 @@
+" Vim syntax file
+" Language: Angular HTML template
+" Maintainer: Dennis van den Berg <dennis@vdberg.dev>
+" Last Change: 2024 Aug 22
+
+" quit when a syntax file was already loaded
+if exists("b:current_syntax")
+  finish
+endif
+
+if !exists("main_syntax")
+  let main_syntax = 'html'
+endif
+
+runtime! syntax/html.vim
+unlet b:current_syntax
+
+let b:current_syntax = "htmlangular"


chrisbra:
thanks. That should be fine for now

chrisbra:
As an improvement for the future, have a look at https://github.com/vim/vim/blob/master/runtime/syntax/htmldjango.vim

PR #15554, opened 2024-08-22
runtime(vim): Update base-syntax, match :loadkeymap after colon and bar

dkearns:
Match :loadkeymap after Ex colons and bars.

Don't generate :loadkeymap as it is handled with a custom syntax group.


chrisbra:
thanks!

Issue #15543, opened 2024-08-21
Segfault when invoking :python3 command


audriusk:
### Steps to reproduce

On Slackware64-current (development branch of Slackware Linux) recent Vim upgrade to version 9.1.0680 causes Vim to segfault when invoking `:python3` command. Running this simple command line is enough to reproduce:
```sh
$ vim -u NONE -c 'py3 print("foobar")'
Vim: Caught deadly signal SEGV
Segmentation fault
```
git bisect shows that the problem started from 9.1.0668 patch.

### Expected behaviour

The version before the upgrade (9.1.0645) was working without segfaults.

### Version of Vim

9.1.0680

### Environment

OS: Slackware64-current (dev branch)
glibc 2.40
gcc 14.2
python 3.11.9
Vim built with --enable-python3interp=dynamic
Terminal: KDE Konsole 23.08.5
$TERM: xterm-256color
Shell: bash 5.2.32

### Logs and stack traces

```shell
Reading symbols from /usr/bin/vim...
[New LWP 6621]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `vim -u NONE -c py3 print("foobar")'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa974265d5b in PyObject_GC_UnTrack () from /usr/lib64/libpython3.11.so.1.0
(gdb) bt
#0  0x00007fa974265d5b in PyObject_GC_UnTrack () from /usr/lib64/libpython3.11.so.1.0
#1  0x00007fa9742e81db in ?? () from /usr/lib64/libpython3.11.so.1.0
#2  0x00007fa974250fe5 in ?? () from /usr/lib64/libpython3.11.so.1.0
#3  0x00007fa97423a01a in ?? () from /usr/lib64/libpython3.11.so.1.0
#4  0x00007fa9741e062d in PyObject_VectorcallMethod () from /usr/lib64/libpython3.11.so.1.0
#5  0x00007fa9742c54aa in ?? () from /usr/lib64/libpython3.11.so.1.0
#6  0x00007fa9742c4fab in Py_FinalizeEx () from /usr/lib64/libpython3.11.so.1.0
#7  0x00000000006ea34b in python3_end () at if_python3.c:1260
#8  0x000000000070ad47 in getout (exitval=1) at main.c:1755
#9  <signal handler called>
[...]

PR #15555, opened 2024-08-22
Problem:  python3: stable-abi by default may cause segfault on Unix

chrisbra:
Problem:  python3: stable-abi by default may cause segfault on Unix
Solution: do not enable the stable Python ABI by default, only when used
          with --with-python3-stable-abi argument is given

related: #15543

cc @k-takata

### DIFF
--- a/src/auto/configure
+++ b/src/auto/configure
@@ -7092,21 +7092,16 @@ printf %s "checking --with-python3-stable-abi argument... " >&6; }
 if test ${with_python3_stable_abi+y}
 then :
   withval=$with_python3_stable_abi;
-	 if test "X$withval" = "Xyes"; then
-	   vi_cv_var_python3_stable_abi=$python3_stable_abi_default
-	 else
-	   vi_cv_var_python3_stable_abi="$withval"
-	 fi
-	 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: $vi_cv_var_python3_stable_abi" >&5
+     if test "X$withval" = "Xyes"; then
+       vi_cv_var_python3_stable_abi=$python3_stable_abi_default
+     else
+       vi_cv_var_python3_stable_abi="$withval"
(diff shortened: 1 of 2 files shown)


chrisbra:
included as of 3f7024cf8

k-takata:
The result of `--with-python3-stable-abi argument` is not shown:
https://github.com/vim/vim/actions/runs/10515717808/job/29136395987?pr=15555#step:15:127
```
checking --with-python3-stable-abi argument... checking Python's abiflags...
```

It should be shown as follows:
```
checking --with-python3-stable-abi argument... no
checking Python's abiflags...
```

chrisbra:
Hm, any idea how to fix this?

k-takata:
Restoring `AC_MSG_RESULT(no)` in the 4th parameter of `AC_ARG_WITH()` should fix it:

```diff
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1549,7 +1549,8 @@ if test "$enable_python3interp" = "yes" -o "$enable_python3interp" = "dynamic";
      else
        vi_cv_var_python3_stable_abi="$withval"
      fi
-     AC_MSG_RESULT($vi_cv_var_python3_stable_abi)])
+     AC_MSG_RESULT($vi_cv_var_python3_stable_abi)],
+     AC_MSG_RESULT(no))
       if test "X$vi_cv_var_python3_stable_abi" != "X"; then
         AC_CACHE_VAL(vi_cv_var_python3_stable_abi_hex,
         [
```
(not tested)

BTW, it looks like 9.1.0691 broke the indentation.

chrisbra:
Thanks. the previous patch used tabs instead of spaces in `configure.ac`, so I changed it back to blanks with `:set et`

chrisbra:
It seems we have a strange mix of spaces and tabs in `configure.ac` anyhow.

PR #15557, opened 2024-08-23
Revert and fix wrong Portuguese menu translation files

ychin:
In #14674, the Brazilian and Portuguese localization files were erroneously combined, even though those are separate locales. This reverts that change, but maintains some of the original typo fixes in
it.

### DIFF
--- a/runtime/lang/menu_pt_br.vim
+++ b/runtime/lang/menu_pt_br.vim
@@ -68,8 +68,8 @@ menutrans Startup\ &Settings		&Inicialização
 menutrans &Global\ Settings		Opções\ &Globais
 
 menutrans Toggle\ Pattern\ &Highlight<Tab>:set\ hls!	Ativar/Desativar\ &Realce\ de\ Padrões<Tab>:set\ hls!
-menutrans Toggle\ &Ignore-case<Tab>:set\ ic!		Ativar/Desativar\ &Ignorar\ maiúsculas<Tab>:set\ ic!
-menutrans Toggle\ &Showmatch<Tab>:set\ sm!		Ativar/Desativar\ &coincidências<Tab>:set\ sm!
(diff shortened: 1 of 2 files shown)


ychin:
The original PR tried to bring some translations over from the pt-br locale to pt-pt, such as from "Spelling" to "Ortografia". I decided to not keep them so they will remain untranslated. I don't speak Portuguese (nor did the original PR author) and I think such decisions should be left to a Portuguese speaker as I don't know if they may have different locale-specific way of translating such things. For example, the text "User Manual" is translated differently under the two locales.

The original PR also changed translation from "Make" to "Fazer" which I think was a mistake. "Make" (see [Portuguese Wikipedia](https://pt.wikipedia.org/wiki/Make)) is a specific program's name and most translations (e.g. Spanish) directly keeps the English "Make" term as it's more a technical term than a general one. Again, I feel like such decisions should be made by a language speaker.

Just a comment on https://github.com/vim/vim/pull/14674#issuecomment-2246986200 though:

> I only noticed later, that the PRs from @ronin49 were mostly machine generated.

I don't think this is completely accurate. He tried to close #10718 and #10717 which used machine generated templates but the PR itself was not machine generated, so I'm not sure why those two PRs were closed as a result of that. FWIW I think #10718 and #10717's approach is flawed. It uses a script to generate a template to parse out the untranslated texts which is good in theory, but it leads to a lot of lines that look like the following:

    " menutrans Scheme TRANSLATION\ MISSING

Scheme is a filetype and does not need translation whatsoever. The script needs to be much smarter if we don't want to pollute the menu translation files. I think such initiative (converting to machine generated template files for translators) should be coordinated with a maintainer first before being checked in. After all they are the ones who will need to use this and it should be their choice how to organize their file.

chrisbra:
Thanks and thanks for clarification of my comment.

PR #15558, opened 2024-08-23
Update pt_BR.po

JNylson:


chrisbra:
thanks, I'll also update the revision date at the top of the file

chrisbra:
Hm, it seems I somehow broke by when importing this file and regenerating it. Let me try it again

PR #15561, opened 2024-08-23
feat(completion): add kind_hlgroup in function complete

glepnir:
Problem: can not set special highlight or kind in popupmen.

Solution: add kind_hlgroup in complete function

lsp has completion kind for special highlight group of kind. we can also support it.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind

Relate https://github.com/neovim/neovim/issues/30112

QQ_1724402392982


chrisbra:
thanks!

Shane-XB-Qian:
Seems wrong: should not override pmenukindsel, which only PmenuKind as its doc said..
// BTW: there 'ctermfd' in its doc, looks a typo.

-- 
shane.xb.qian

flexagoon:
> Seems wrong: should not override pmenukindsel, which only PmenuKind as its doc said..

Why? Pmenukindsel is the default color for selected pmenu kind, and this one is for a specific item

Shane-XB-Qian:
1. pmenukindsel became un-useful if it can be overrode
2. it is 'kind_hlgroup' doc said, which only pmenukind
// i saw your/this following PR #15572 which only correct 'ctermfd' typo
3. it may help on https://github.com/vim/vim/pull/15314#issuecomment-2282696695 which made 'kind_hlgroup' be higher priority than 'hl_group' too
// if so, 'kind_hlgroup' would be still possible to indicate that item was selected and was not striked

-- 
shane.xb.qian

PR #15564, opened 2024-08-23
runtime(vim): Update base-syntax, improve :let-heredoc highlighting

dkearns:
The end marker is not required to match the indent of :let when "trim"
is specified, it may also appear without leading whitespace as normal.


chrisbra:
thanks!

PR #15565, opened 2024-08-23
Wrong patlen value in ex_substitute()

zeertzjq:
Problem:  Wrong patlen value in ex_substitute() (after 9.1.0426).
Solution: Compute patlen after finding end separator.

Add a more explicit test.  The test already passes as the only case
where a overlarge patlen value matters was fixed by patch 9.1.0689.


chrisbra:
thanks!

PR #15567, opened 2024-08-24
runtime(matchparen): don't do matching on very long lines

zeertzjq:
On very long lines matching may become slow, so don't do matching if the
cursor column is larger than 'synmaxcol'.

Example:

	call setline(1, repeat(' foobar', 100000))
	runtime plugin/matchparen.vim
	normal! $hhhhhhhh


zeertzjq:
Actually I found a way to make this faster so that the limit is unnecessary. I'll open a new PR later.

PR #15568, opened 2024-08-24
matchparen is slow on a long line

zeertzjq:
Problem:  The matchparen plugin is slow on a long line.
Solution: Don't use a regexp to get char at and before cursor.

Example:

	call setline(1, repeat(' foobar', 100000))
	runtime plugin/matchparen.vim
	normal! $hhhhhhhh


chrisbra:
ah thanks!

PR #15569, opened 2024-08-24
Recognize file triggers of spec files

fundawang:
https://rpm-software-management.github.io/rpm/manual/file_triggers.html

RPM 4.13[1] introduces file triggers support, the relavant scriptlets should be highlighted.

[1]: https://rpm.org/wiki/Releases/4.13.0


fundawang:
@ignatenkobrain

PR #15575, opened 2024-08-25
Restore setting of VIMPROG when making vimtags

jamessan:
Revert part of 4c3616d7a2c since runtime/doc/Makefile's default value for VIMPROG does not work if vim was built in a SHADOWDIR.


chrisbra:
thanks!

PR #15576, opened 2024-08-25
hard to know the users personal Vim Runtime Directory

chrisbra:
Problem:  hard to guess the Vim Runtime Directory
Solution: Set the $MYVIMRCDIR environment variable to the directory
          containing $MYVIMRC

### DIFF
--- a/runtime/doc/starting.txt
+++ b/runtime/doc/starting.txt
@@ -1,4 +1,4 @@
-*starting.txt*  For Vim version 9.1.  Last change: 2024 Aug 03
+*starting.txt*  For Vim version 9.1.  Last change: 2024 Aug 31
 
 
 		  VIM REFERENCE MANUAL    by Bram Moolenaar
@@ -828,8 +828,8 @@ accordingly.  Vim proceeds in this order:
 	easy to copy it to another system.
 
 	If Vim was started with "-u filename", the file "filename" is used.
-	All following initializations until 4. are skipped. $MYVIMRC is not
-	set.
+	All following initializations until 4. are skipped. $MYVIMRC and
+	$MYVIMDIR are not set.
 	"vim -u NORC" can be used to skip these initializations without
 	reading a file.  "vim -u NONE" also skips loading plugins.  |-u|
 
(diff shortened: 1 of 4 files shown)


runtime/doc/starting.txt
```diff
@@ -828,8 +828,8 @@ accordingly.  Vim proceeds in this order:
 	easy to copy it to another system.
 
 	If Vim was started with "-u filename", the file "filename" is used.
-	All following initializations until 4. are skipped. $MYVIMRC is not
-	set.
+	All following initializations until 4. are skipped. $MYVIMRC and
+	$MYVIMRCDIR is not set.
```
habamax: are not set?

chrisbra: yeah correct

k-takata:
Just curious, but what is the use case that $MYVIMRC is not enough?

habamax:
My guess is to (or at least I would have used it) to simplify following:

```
&directory = $'{fnamemodify($MYVIMRC, ":p:h")}/.data/swap/'
&backupdir = $'{fnamemodify($MYVIMRC, ":p:h")}/.data/backup//'
&undodir = $'{fnamemodify($MYVIMRC, ":p:h")}/.data/undo//'
if !isdirectory(&undodir)   | mkdir(&undodir, "p")   | endif
if !isdirectory(&backupdir) | mkdir(&backupdir, "p") | endif
if !isdirectory(&directory) | mkdir(&directory, "p") | endif
```

chrisbra:
> &directory = $'{fnamemodify($MYVIMRC, ":p:h")}/.data/swap/'

basically that, except for that this doesn't work properly, if you are using `~/.vimrc` file, in which case `$MYVIMRCDIR` should point to `~/.vim/`.

I need to update the tests for this but at the same time busy with a few more security related issues :/

jmdevin:
> basically that, except for that this doesn't work properly, if you are using ~/.vimrc file, in which case $MYVIMRCDIR should point to ~/.vim/.

If I understand that correctly, perhaps the new variable is better named `$MYVIMDIR` rather than `$MYVIMRCDIR`? Just from the name, I would expect `$MYVIMRCDIR` to resolve to `~` when the user's vimrc is at `~/.vimrc`.

chrisbra:
> If I understand that correctly, perhaps the new variable is better named $MYVIMDIR rather than $MYVIMRCDIR

:) I had it as `$MYVIMDIR` initially, but I thought `MYVIMRCDIR` is clearer. I'll change it back then.

PR #15577, opened 2024-08-26
Fix typo in test_crash.vim

zeertzjq:


zeertzjq:
Reopening this in case #15578 isn't merged

PR #15578, opened 2024-08-26
"Untitled" file not removed when running Test_crash1_3 alone

zeertzjq:
Problem:  "Untitled" file not removed when running Test_crash1_3 alone
          with TEST_FILTER.
Solution: Use a TearDown function instead of another test.

### DIFF
--- a/src/testdir/test_crash.vim
+++ b/src/testdir/test_crash.vim
@@ -237,9 +237,9 @@ func Test_crash2()
   exe buf .. "bw!"
 endfunc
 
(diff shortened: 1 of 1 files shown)


chrisbra:
I tried that before and it still left the file around. That's why I used the workaround with the latest test. Did this work reliably for you?

zeertzjq:
Yes

chrisbra:
okay, then let me merge this one and close https://github.com/vim/vim/pull/15577 again :) Thanks

PR #15579, opened 2024-08-26
runtime(java): Provide support for syntax preview features

zzzyxwvut:
Introduce a new API variable `g:java_syntax_previews` whose  
value must be a list of syntax preview feature numbers.

Enumerate the currently supported numbers in a table at the  
end of the documentation entry for `ft-java-syntax`.

Also, disable the recognition of String Templates.  Despite  
the withdrawal of this preview feature in its proposed form  
from the upcoming JDK 23 release and the fact that the JDK  
22 release is coming to EOL this September, an earlier  
iteration of this preview feature was included in JDK 21  
(LTS) whose EOL is projected to fall due in late 2028 and,  
therefore, retain the current implementation.

Define `g:java_syntax_previews` and include number `430` in  
its list to enable the recognition of String Templates:

```vim
	let g:java_syntax_previews = [430]
```

References:
https://openjdk.org/jeps/430 (Preview)
https://openjdk.org/jeps/459 (Second Preview)
https://openjdk.org/jeps/465 (Third Preview)
https://mail.openjdk.org/pipermail/amber-spec-experts/2024-April/004106.html


chrisbra:
thanks!

PR #15586, opened 2024-08-28
runtime(vim): Update syntax, improve user-command matching

dkearns:
- Match -addr and -keepscript attributes and generate -addr values.
- Match attribute errors where = is specified.
- Highlight attributes with Special like other Ex command options.
- Match :delcommand -buffer attribute.


runtime/syntax/generator/vim.vim.base
```diff
@@ -3,7 +3,7 @@
 " Maintainer:	   Hirohito Higashi <h.east.727 ATMARK gmail.com>
 "	   Doug Kearns <dougkearns@gmail.com>
 " URL:	   https://github.com/vim-jp/syntax-vim-ex
```
k-takata: Maybe, this URL is not needed now.

dkearns: Removed, thanks.

chrisbra:
thanks!

Issue #15574, opened 2024-08-25
Vim fails to recognize autocmd names containing a capital I in Turkish locale


jamessan:
### Steps to reproduce

1. `env LC_CTYPE=tr_TR.UTF-8 vim --clean`
2. `:autocmd GUIEnter`

Bisecting shows this is a regression introduced in v9.1.0256.

### Expected behaviour

Any defined GUIEnter autocmds are listed. Instead, we get the error `E216: No such group or event: GUIEnter`.

### Version of Vim

9.1.0496

### Environment

Debian unstable

chrisbra:
@basilisk0315

chrisbra:
I am not sure if we should revert that part of the patch, or rather switch locales instead:
```diff
diff --git a/src/autocmd.c b/src/autocmd.c
index 00f41bddd..e5ab320c2 100644
--- a/src/autocmd.c
+++ b/src/autocmd.c
@@ -699,6 +699,10 @@ event_name2nr(char_u *start, char_u **end)
     char_u     *p;
     keyvalue_T target;
     keyvalue_T *entry;
+#if defined(HAVE_LOCALE_H) || defined(X_LOCALE)
+    char_u     *locale = (char_u *)setlocale(LC_CTYPE, NULL);
+    setlocale(LC_CTYPE, "C");
+#endif
     static keyvalue_T *bufnewfile = &event_tab[BUFNEWFILE_INDEX];
     static keyvalue_T *bufread = &event_tab[BUFREAD_INDEX];

@@ -720,6 +724,9 @@ event_name2nr(char_u *start, char_u **end)
        entry = bufread;
     else
        entry = (keyvalue_T *)bsearch(&target, &event_tab, ARRAY_LENGTH(event_tab), sizeof(event_tab[0]), cmp_keyvalue_value_ni);
+#if defined(HAVE_LOCALE_H) || defined(X_LOCALE)
+    setlocale(LC_CTYPE, locale);
+#endif

     if (*p == ',')
        ++p;
```

This patch should fix it, but I was surprised I need to set `LC_CTYPE` instead of `LC_COLLATE`.

chrisbra:
So I checked it a bit more and changing the locale is probably just in-efficient. 
[...]

PR #15591, opened 2024-08-30
Fix test_gettext tests when run in a shadowdir

jamessan:
Install locales-all package so the tests actually run in CI (which then exposes the issue in the shadowdir-enabled job).

Symlink the ru_RU directory into the shadowdir so the translations are available when testing inside a shadowdir.


chrisbra:
Ah, thank you. I assume the `locales-all` package can alo be helpful for testing the turkish locale issue #15574 (I wondered how I can enable this in CI).

PR #15593, opened 2024-08-30
Sorting of fuzzy filename completion is not stable

zeertzjq:
Problem:  Sorting of fuzzy filename completion is not stable.
Solution: Compare indexes when scores are equal.  Fix some typos.


chrisbra:
thanks!

PR #15594, opened 2024-08-30
change hl_group to hlgroup of complete function

glepnir:
Problem: hl_group looks not great with kind_hlgroup.

Solution: change to hlgroup.

### DIFF
--- a/runtime/doc/insert.txt
+++ b/runtime/doc/insert.txt
@@ -1183,7 +1183,7 @@ items:
 	user_data	custom data which is associated with the item and
 			available in |v:completed_item|; it can be any type;
 			defaults to an empty string
-	hl_group	an additional highlight group whose attributes are
+	hlgroup		an additional highlight group whose attributes are
 			combined with |hl-PmenuSel| and |hl-Pmenu| or
 			|hl-PmenuMatchSel| and |hl-PmenuMatch| highlight
 			attributes in the popup menu to apply cterm and gui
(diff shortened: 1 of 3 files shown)


glepnir:
or change to abbr_hlgroup and then only abbr has stripethrough ?

Shane-XB-Qian:
> or change to abbr_hlgroup and then only abbr has stripethrough ? 

had said here: https://github.com/vim/vim/pull/15561#issuecomment-2308661282
or if you now really like to separate it, then maybe 'label_hlgroup' was a good name.

// BTW: you changed not only 'runtime(doc)', but also 'c' code, so the ticket title was confused.

-- 
shane.xb.qian

glepnir:
label is lsp side.  rename to abbr_hlgroup or other name_hlgroup  need modifi some c code so it's next follow-up PR.

Shane-XB-Qian:
> label is lsp side.  rename to abbr_hlgroup or other name_hlgroup  need modifi some c code so it's next follow-up PR.

'abbr' is another well know native cmd in vim, that maybe confused, so i suggested 'label';
but that is not the point, the point is how to design the hl and their belonging/priority;
if you would not make PmenuKindSel *not* be override by PmenuKind (as above my comment said), then this PR the name changing seems would not change or help something.
// BTW: it is not regarding to lsp always, it just some hl which users may like to set (or not).

-- 
shane.xb.qian

glepnir:
word abbr respect complete-items and please do not raise unrelated issues under PR. If you have a problem with how kind works, please create an issue instead of sending a comment under an unrelated PR. I handle hlgroup issues here

Shane-XB-Qian:
> word abbr respect complete-items and please do not raise unrelated issues under PR. If you have a problem with how kind works, please create an issue instead of sending a comment under an unrelated PR. I handle hlgroup issues here

i am saying the related issue, hl_group impacted kind_hlgroup, now hl_group striked 'kind' area too if set that.
// but well, if you are not like to listen, that's fine, i donot mind.. good luck for this topic~

-- 
shane.xb.qian

PR #15588, opened 2024-08-29
Optimize insertion (`:normal 10000ix`)

k-takata:
Editing a very long line in Vim is slow.
This patch optimizes the insertion with a large count (e.g. `:normal 10000ix`).

It seems that calculation of the cursor position for a long line is slow and it takes O(n^2).
Disable the calculation if not needed.

Before:
```
$ time ./vim --clean -c 'normal 10000ix' -cq!
real    0m1.879s
user    0m1.328s
sys     0m0.139s

$ time ./vim --clean -c 'normal 20000ix' -cq!
real    0m5.574s
user    0m5.421s
sys     0m0.093s

$ time ./vim --clean -c 'normal 40000ix' -cq!
real    0m23.588s
user    0m23.187s
sys     0m0.140s
```

After:
```
$ time ./vim --clean -c 'normal 10000ix' -cq!
real    0m0.715s
user    0m0.156s
sys     0m0.155s

$ time ./vim --clean -c 'normal 20000ix' -cq!
real    0m0.678s
user    0m0.515s
sys     0m0.109s

$ time ./vim --clean -c 'normal 40000ix' -cq!
real    0m2.475s
user    0m2.311s
sys     0m0.124s
```

PR #15598, opened 2024-08-31
Optimize memory allocation for insertion

k-takata:
This is an additional optimization for #15588.
Reduce the memory allocation in ins_str().

If this PR is combined with #15588, `:normal 10000ix` becomes O(n):
```
$ time ./vim --clean -c 'normal 160000ix' -cq!
real    0m0.684s
user    0m0.109s
sys     0m0.108s

$ time ./vim --clean -c 'normal 1280000ix' -cq!
real    0m0.969s
user    0m0.827s
sys     0m0.077s
```

### DIFF
--- a/src/change.c
+++ b/src/change.c
@@ -1181,6 +1181,7 @@ ins_char_bytes(char_u *buf, int charlen)
     void
 ins_str(char_u *s)
 {
+    static char_u	*prevp = NULL;
     char_u	*oldp, *newp;
     int		newlen = (int)STRLEN(s);
     int		oldlen;
@@ -1198,7 +1199,7 @@ ins_str(char_u *s)
 
     newtotlen = oldlen + newlen;
     new_size = newtotlen + 1;
-    if ((curbuf->b_ml.ml_line_size < new_size)
+    if ((curbuf->b_ml.ml_line_size < new_size) || (prevp != oldp)
 #ifdef FEAT_PROP_POPUP
 	    || curbuf->b_has_textprop
 #endif
@@ -1215,6 +1216,7 @@ ins_str(char_u *s)
     }
     else
 	newp = oldp;
+    prevp = newp;
 
     if (col > 0 && newp != oldp)
 	mch_memmove(newp, oldp, (size_t)col);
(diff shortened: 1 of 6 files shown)


src/memline.c
```diff
@@ -4236,6 +4237,7 @@ ml_flush_line(buf_T *buf)
 	    }
 	}
 	vim_free(new_line);
+	buf->b_ml.ml_line_size = 0;
```
chrisbra, line 4247: don't you ned this a few lines below, when freeing `buf->b_ml.ml_line_ptr`?

src/change.c
```diff
@@ -1194,14 +1197,32 @@ ins_str(char_u *s)
...
+	    )
+    {
+#ifdef FEAT_PROP_POPUP
+	if (!curbuf->b_has_textprop)
+#endif
+	    new_size = new_size * 3 / 2;
+	newp = alloc(new_size);
+	if (newp == NULL)
+	    return;
+	curbuf->b_ml.ml_line_size = new_size;
```
chrisbra, line 1215: That should potentially be encapsulated into a new memline()  function `ml_replace_len_size()` which takes the new size parameter and stores the new size there.

src/memline.c
k-takata: I don't think it is necessary, but it should be safer.
Also clearing it in textprop.c might be better.

src/change.c
k-takata: I tried to implement this, but it didn't work...

zeertzjq:
I think this is O(n log n), not O(n).

k-takata:
I forget how to calculate the theoretical order of this algorithm, but the result shows that it is O(n):
```
$ time ./vim --clean -c 'normal 1280000ix' -cq!
real    0m0.969s
user    0m0.827s
sys     0m0.077s

$ time ./vim --clean -c 'normal 5120000ix' -cq!
real    0m4.222s
user    0m3.296s
sys     0m0.046s

$ time ./vim --clean -c 'normal 20480000ix' -cq!
real    0m13.190s
user    0m12.890s
sys     0m0.156s
```

zeertzjq:
Oh, I made a mistake. Yes it's O(n).

PR #15601, opened 2024-08-31
Set commentstring for nu filetype

mrcjkb:


runtime/ftplugin/nu.vim
```diff
@@ -0,0 +1,13 @@
+" Vim filetype plugin
+" Language:	nu
```
dkearns:
```suggestion
" Language:	Nu
```

I do what I can...

dkearns (changes requested):

chrisbra:
Thanks. Please also add yourself to the Maintainers file.

chrisbra:
I read https://programming-nu.github.io/syntax 
Is  `#` prefered as leading comment character (instead of using `;`)?

mrcjkb:
> I read https://programming-nu.github.io/syntax Is `#` prefered as leading comment character (instead of using `;`)?

I based this on [what they did in tree-sitter-nu](https://github.com/nushell/tree-sitter-nu/blob/0bb9a602d9bc94b66fab96ce51d46a5a227ab76c/plugin/init.lua#L5).

chrisbra:
alright thanks

PR #15603, opened 2024-08-31
GUIEnter event not found in Turkish locale

### LINKED ISSUE #15574: Vim fails to recognize autocmd names containing a capital I in Turkish locale

chrisbra:
Problem:   GUIEnter not found in Turkish locale
           (James McCoy, after v9.1.0256, the issue was there before,
            but v9.1.0256 made it more apparent)
Solution:  explicitly compare autocommand events by ASCII value and
           ignoring locale, because according to the documentation,
           events are case insensitive (:h autocommand-events)

fixes: #15574

### DIFF
--- a/src/highlight.c
+++ b/src/highlight.c
@@ -1215,8 +1215,8 @@ highlight_set_cterm_color(
 
 	target.key = 0;
 	target.value = (char *)arg;
-	target.length = 0;	// not used, see cmp_keyvalue_value_i()
-	entry = (keyvalue_T *)bsearch(&target, &color_name_tab, ARRAY_LENGTH(color_name_tab), sizeof(color_name_tab[0]), cmp_keyvalue_value_i);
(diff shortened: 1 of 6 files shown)


zeertzjq:
Should `cmp_keyvalue_value_i` be changed too?

chrisbra:
Good idea. It seems we can just get rid of `cmp_keyvalue_value_i()` and instead use `cmp_keyvalue_value_ni()` instead. Let me add this here as well

PR #15604, opened 2024-08-31
Refactor option.c to remove calls to STRLEN()

basilisk0315:
I experimented with doing something with the lookups on the options array (primarily in findoption()), but decided against it for this PR.


chrisbra:
Thanks. I read over it and it looks fine. I hope this doesn't introduce a regression, we had some of this in the past by those optimizations.

PR #15605, opened 2024-09-01
Recursive window update does not account for reset skipcol

luukvbaal:
Problem:  Window is updated with potentially invalid skipcol in recursive
          window update path. I.e. cursor outside of visible range in
          large line that does not fit.
Solution: Make sure it is valid.


chrisbra:
thanks!

PR #15607, opened 2024-09-02
runtime(sudoers): syntax improvements

intelfx:
Two improvements:

1. Recognize all Tag_Spec items, not just `{,NO}PASSWD`
2. Recognize the optional colon-delimited second part of Runas_Spec specifying the permitted group list

The second one is relatively sloppy (see commit message), but I guess it's better than nothing at all. If it's not acceptable, then I'd like some hints as to how to approach this properly because I'm not experienced with Vim syntax engine and I don't see a clean way to do that.

### DIFF
--- a/runtime/syntax/sudoers.vim
+++ b/runtime/syntax/sudoers.vim
@@ -22,7 +22,7 @@ syn match   sudoersUserSpec '^' nextgroup=@sudoersUserInSpec skipwhite
 
 syn match   sudoersSpecEquals         contained '=' nextgroup=@sudoersCmndSpecList skipwhite
 
-syn cluster sudoersCmndSpecList       contains=sudoersUserRunasBegin,sudoersPASSWD,@sudoersCmndInSpec
+syn cluster sudoersCmndSpecList       contains=sudoersUserRunasBegin,sudoersTagSpec,@sudoersCmndInSpec
 
 syn keyword sudoersTodo               contained TODO FIXME XXX NOTE
 
@@ -94,7 +94,7 @@ syn cluster sudoersUserSpec         contains=sudoersUserSpecComma,@sudoersHostIn
 
 syn match   sudoersUserRunasBegin   contained '(' nextgroup=@sudoersUserInRunas skipwhite skipnl
 syn match   sudoersUserRunasComma   contained ',' nextgroup=@sudoersUserInRunas skipwhite skipnl
-syn match   sudoersUserRunasEnd     contained ')' nextgroup=sudoersPASSWD,@sudoersCmndInSpec skipwhite skipnl
+syn match   sudoersUserRunasEnd     contained ')' nextgroup=sudoersTagSpec,@sudoersCmndInSpec skipwhite skipnl
 syn cluster sudoersUserRunas        contains=sudoersUserRunasComma,@sudoersUserInRunas,sudoersUserRunasEnd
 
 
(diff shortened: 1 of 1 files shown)


runtime/syntax/sudoers.vim
```diff
@@ -291,7 +292,7 @@ syn region  sudoersStringValue  contained start=+"+ skip=+\\"+ end=+"+ nextgroup
 syn match   sudoersListValue    contained '[^[:space:],:=\\]*\%(\\[[:space:],:=\\][^[:space:],:=\\]*\)*' nextgroup=sudoersParameterListComma skipwhite skipnl
 syn region  sudoersListValue    contained start=+"+ skip=+\\"+ end=+"+ nextgroup=sudoersParameterListComma skipwhite skipnl
 
-syn match   sudoersPASSWD                   contained '\%(NO\)\=PASSWD:' nextgroup=@sudoersCmndInSpec skipwhite
+syn match   sudoersTagSpec      contained '\%(NO\)\=\%(EXEC\|FOLLOW\|LOG_\%(INPUT\|OUTPUT\)\|MAIL\|INTERCEPT\|PASSWD\|SETENV\):' nextgroup=sudoersTagSpec,@sudoersCmndInSpec skipwhite
```
e-kwsm, line 295: :+1:

runtime/syntax/sudoers.vim
```diff
@@ -92,10 +92,11 @@ syn cluster sudoersUserList         contains=sudoersUserListComma,sudoersUserLis
...
 syn cluster sudoersUserSpec         contains=sudoersUserSpecComma,@sudoersHostInSpec
 
-syn match   sudoersUserRunasBegin   contained '(' nextgroup=@sudoersUserInRunas skipwhite skipnl
+syn match   sudoersUserRunasBegin   contained '(' nextgroup=@sudoersUserInRunas,sudoersUserRunasColon skipwhite skipnl
 syn match   sudoersUserRunasComma   contained ',' nextgroup=@sudoersUserInRunas skipwhite skipnl
-syn match   sudoersUserRunasEnd     contained ')' nextgroup=sudoersPASSWD,@sudoersCmndInSpec skipwhite skipnl
-syn cluster sudoersUserRunas        contains=sudoersUserRunasComma,@sudoersUserInRunas,sudoersUserRunasEnd
+syn match   sudoersUserRunasColon   contained ':' nextgroup=@sudoersUserInRunas skipwhite skipnl
+syn match   sudoersUserRunasEnd     contained ')' nextgroup=sudoersTagSpec,@sudoersCmndInSpec skipwhite skipnl
+syn cluster sudoersUserRunas        contains=sudoersUserRunasComma,sudoersUserRunasColon,@sudoersUserInRunas,sudoersUserRunasEnd
```
e-kwsm, line 99: [Runas_Spec can be](https://github.com/sudo-project/sudo/blob/4751a4d2dd09b94e438208439902b36aec79767e/docs/sudoers.man.in#L1436-L1482)

1. `(user:group)`
2. `(user)`
3. `(:group)`
4. `()`

where user/group are comma-separated lists.

The current vim syntax works for 2 only; this PR implements support of 1 and 3 but not 4. Would you address empty spec if possible?

e-kwsm (approved):

chrisbra:
Thanks. Since @e-kwsm approves, I assume we can leave this empty spec for later?

chrisbra:
It was approved, so I merged it (although I forgot to give proper authorship, mea culpa)

e-kwsm:
> If both Runas_List are empty, the command may only be run as the invoking user and the
group, if specified, must be one that the invoking user is a member of.

I don’t know sudo mechanism in detail;
perhaps empty spec is designed to prohibit the user to run the commands as the different users/groups.

Approval was a bit hasty, but I don’t believe this introduces regression and you can undertake it afterward.

PR #15612, opened 2024-09-03
Fix some test issues

jamessan:
Debian has a CI system which runs tests for an existing package when other relevant packages change to ensure they aren't breaking the existing package.

This is the context in which I encountered these test failures. The first because /usr/bin/xxd is being tested and the second because I hadn't updated the test environment to include xgettext.

- **Make Test_xxd_color_0 agnostic to xxd's path, similar to Test_xxd_color_1**
- **Skip Test_gettext_makefile if xgettext is not available**


chrisbra:
thanks!

PR #15614, opened 2024-09-03
Newline causes E749 in Ex mode

### LINKED ISSUE #15613: writefile call fails with `Empty buffer` error
### Steps to reproduce

Repro script:
```
#!/bin/sh
vim -NEnR --cmd '
call writefile(["hello"], "/dev/stderr")
quit
' </dev/tty >& /tmp/out.txt
```

### Expected behaviour

With old vim versions this would write "hello" to stderr and quit. After vim 9.1.0573 this produces an `E749: Empty buffer` and waits for user input.

This breaks the vimpager script as I reported here: https://github.com/rkitover/vimpager/issues/276

### Version of Vim

9.1.0573

I used `git bisect` to identify this specific version to be the first to show this behavior.

### Environment

MacOS homebrew
```
COLORTERM=truecolor
TERM=xterm-256color
TERM_PROGRAM=WezTerm
TERM_PROGRAM_VERSION=20240203-110809-5046fc22
```

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Newline causes E749 in Ex mode (after 9.1.0573).
Solution: Don't execute empty command followed by a newline.

Fix #15613

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -2916,6 +2916,11 @@ parse_command_modifiers(
 	    }
 	    return FAIL;
 	}
+	if (eap->nextcmd == NULL && *eap->cmd == '\n')
+	{
+	    eap->nextcmd = eap->cmd + 1;
+	    return FAIL;
+	}
 	if (*eap->cmd == NUL)
 	{
 	    if (!skip_only)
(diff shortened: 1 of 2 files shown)


zeertzjq:
There is also another long-standing problem (probably dating back to df177f679e950a2ab2ad5fe7d45c1daface004d7) that sourcing a script containing any blank lines in Ex mode leads to E749 because of the `ex_pressedreturn = TRUE` below. I wonder if such behavior should be changed.

chrisbra:
thanks!

PR #15619, opened 2024-09-03
Not correctly parsing color names (after v9.1.0709)

### LINKED ISSUE #15617: 9.1.0709 breaks colors
### Steps to reproduce

9.1.0709 breaks coloring.

With `vim --clean`, make sure to set &tgc and then try `hi normal guifg=gray40`.

### Expected behaviour

The text should be a dark gray, but it is in fact a pretty light gray.

Reverting to the commit before that immediately fixes the issue.

It's not a terminal probelm because it also happens in gvim.  I can reproduce it in uxterm and kitty, too.

### Version of Vim

9.1.0709

### Environment

Debian 12.6 or Linux Mint 21.3 both show it.
Shell: bash

### Logs and stack traces

_No response_

chrisbra:
Problem:  Not correctly parsing color names (chdiza, after v9.1.0709)
Solution: Revert part of the patch that compares the color names

fixes: #15617

### DIFF
--- a/src/highlight.c
+++ b/src/highlight.c
@@ -1215,8 +1215,8 @@ highlight_set_cterm_color(
 
 	target.key = 0;
 	target.value = (char *)arg;
-	target.length = 0;	// not used, see cmp_keyvalue_value_ni()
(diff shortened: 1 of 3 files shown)


chrisbra:
no test included, as I am not sure how to test this.

PR #15620, opened 2024-09-03
Unnecessary nextcmd NULL checks in parse_command_modifiers()

zeertzjq:
Problem:  Unnecessary nextcmd NULL checks in parse_command_modifiers().
Solution: Remove them.

Every place parse_command_modifiers() is called, nextcmd is NULL, and
after it's set to non-NULL the function returns very soon.
Even if one day nextcmd may be non-NULL, the NULL checks may still be
wrong as the correct behavior may be overriding nextcmd.

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -2898,12 +2898,9 @@ parse_command_modifiers(
 	if (comment_start(eap->cmd, starts_with_colon))
 	{
 	    // a comment ends at a NL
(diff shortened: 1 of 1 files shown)


chrisbra:
Thanks. Using `assert()` is not used much in the code base, so I wonder if this may cause crashes

zeertzjq:
Ah, I'll remove it then

chrisbra:
thanks!

PR #15624, opened 2024-09-04
Optimize cursor computation

### LINKED ISSUE #15606: Calculation of the cursor position for a long line is slow
As I wrote in #15588, it seems that the calculation of the cursor position for a long line is slow and it takes O(n^2).

I've found that `getvcol()` and `win_linetabsize_cts()` in charset.c are slow.
If we can optimize these functions, editing very long lines will become more efficient.

Currently, I have no idea how to optimize them. So, I create this as a record.

Also related: #555

k-takata:
Fixes #15606

Cache the result of `getvcol()` and `win_linetabsize_cts()`.
Cache the vcol value at 80% position or at the last 4KB position.
Invalidate the cache when the content of the line is changed or the window options are changed.

To check the performance, e.g. create a line with 2 million characters:
```
:call setline(1, repeat("x", 2 * 1000 * 1000))
```
then move the cursor around.

### DIFF
--- a/src/charset.c
+++ b/src/charset.c
@@ -900,7 +900,11 @@ win_linetabsize_cts(chartabsize_T *cts, colnr_T len)
 	    && (old_line != NULL)
 	    && (old_line == cts->cts_line)
 	    && (saved_ptr >= cts->cts_line)
-	    && (saved_ptr < cts->cts_line + slen))
+	    && (saved_ptr < cts->cts_line + slen)
+#ifdef FEAT_PROP_POPUP
+	    && !cts->cts_has_prop_with_text
+#endif
+       )
     {
 	cts->cts_ptr = saved_ptr;
 	cts->cts_vcol = vcol = saved_vcol;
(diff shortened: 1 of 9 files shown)


src/charset.c
```diff
@@ -849,16 +849,87 @@ linetabsize_no_outer(win_T *wp, linenr_T lnum)
...
+    void
+enable_vcol_cache(void)
+{
+    s_use_vcol_cache = TRUE;
+}
+
+    void
+disable_vcol_cache(void)
+{
+    s_use_vcol_cache = TRUE;
```
h-east: Maybe it's like this?

```suggestion
    s_use_vcol_cache = FALSE;
```

chrisbra:
I would think we need to invalidate the cache on a few more occasions. Probably what is done for the breakindent value (see get_breakindent_win() in indent.c

src/charset.c
```diff
@@ -849,16 +849,87 @@ linetabsize_no_outer(win_T *wp, linenr_T lnum)
...
+	if (slen == MAXCOL)
+	    slen = STRLEN(cts->cts_line);
+	col_save = (colnr_T)((varnumber_T)slen * 4 / 5);
+	if (slen - col_save > 4096)
+	    col_save = slen - 4096;
+    }
+    else
+	s_vcol_cache_valid1 = FALSE;
+    if (s_vcol_cache_valid1
+	    && (old_win != NULL) && (old_win != cts->cts_win)
```
zeertzjq:
```suggestion
	    && (old_win != NULL) && (old_win == cts->cts_win)
```

src/charset.c
k-takata: thanks

src/charset.c
k-takata: thanks

zeertzjq:
I think a cache invalidation is needed in `win_new_width()` as well.

zeertzjq:
And with `signcolumn=auto` cache invalidation may be needed when the width of the sign column changes, so maybe the value of `win_col_off(wp)` and `win_col_off2(wp)` also needed to be saved so that when their values change the cache can be invalidated...

PR #15628, opened 2024-09-05
Fix setcellwidths([])

k-takata:
Fix that `setcellwidths([])` didn't update the display.
Redraw after clearing the cellwidth table.

### DIFF
--- a/src/mbyte.c
+++ b/src/mbyte.c
@@ -5644,7 +5644,7 @@ f_setcellwidths(typval_T *argvars, typval_T *rettv UNUSED)
 	// Clearing the table.
 	VIM_CLEAR(cw_table);
 	cw_table_size = 0;
-	return;
+	goto done;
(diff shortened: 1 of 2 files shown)


src/mbyte.c
```diff
@@ -5644,6 +5644,8 @@ f_setcellwidths(typval_T *argvars, typval_T *rettv UNUSED)
 	// Clearing the table.
 	VIM_CLEAR(cw_table);
 	cw_table_size = 0;
+	changed_window_setting_all();
+	redraw_all_later(UPD_CLEAR);
 	return;
```
zeertzjq: Will it be better to add a `done:` label at the bottom and use a `goto done;` here?

k-takata: OK, updated.

chrisbra:
thanks!

PR #15629, opened 2024-09-05
Resetting cell widths can make 'listchars' or 'fillchars' invalid

zeertzjq:
Problem:  Resetting cell widths can make 'listchars' or 'fillchars'
          invalid.
Solution: Check for conflicts when resetting cell widths.


chrisbra:
thanks!

PR #15635, opened 2024-09-06
Wrong breakindentopt=list:-1 with multibyte or TABs

### LINKED ISSUE #15634: breakindentopt=list:-1 breaks with characters that are not one byte per display cell
### Steps to reproduce

1. Do `vim --clean`
2. Paste this:
```vimscript
set nomodified
enew
set buftype=nofile
set linebreak
set showbreak=
set breakindent
set breakindentopt=list:-1
set formatlistpat=\\m^\\s*\\(•\\\|*\\)\\+\\s\\+
call setline(1, "• " .. repeat('asdf ', 300))
```
3. Do `:%source`
4. See that the indent of the wrapped lines should be two cells, but is instead four.

I took an asciicast of the problem but I don't know how to display it here:
[vim_indent_bug.cast.txt](https://github.com/user-attachments/files/16913145/vim_indent_bug.cast.txt)

### Expected behaviour

The wrapped lines should indent based on the displayed width of the characters. Right now, the indent seems to be based on the byte length of the characters. The snippet above uses bullet point `•` ([link](https://www.compart.com/en/unicode/U+2022)), which is three bytes in utf-8, to demonstrate this.

Indenting with a tab also causes issues.

### Version of Vim

9.1.0689

### Environment

OS: Ubuntu 24.04.1
Terminal: Wezterm 20240203-110809-5046fc22

### Logs and stack traces

_No response_

zeertzjq:
Problem:  Wrong breakindentopt=list:-1 with multibyte chars or TABs in
          text matched by 'formatlistpat'.
Solution: Use the width of the match text.

fixes: #15634

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -1,4 +1,4 @@
-*options.txt*	For Vim version 9.1.  Last change: 2024 Aug 12
+*options.txt*	For Vim version 9.1.  Last change: 2024 Sep 07
 
 
 		  VIM REFERENCE MANUAL	  by Bram Moolenaar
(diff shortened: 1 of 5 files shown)


zeertzjq:
Wait, the indent may also be wrong with TABs

jmdevin:
> Wait, the indent may also be wrong with TABs

Yes, I can confirm that. That's actually how I finally figured out the cause of the issue.

Thanks for being on this so fast by the way

chrisbra:
thanks!

PR #15638, opened 2024-09-07
Crash when id > INT_MAX in text_prop_add()

### LINKED ISSUE #15637: Segmentation Fault when using prop_add with a large number
### Steps to reproduce

**Description**:
I encountered a segmentation fault when using the `prop_add` function with a large number (e.g., `2147483648`) as the `id`. The following error occurs:

```
Vim: Caught deadly signal SEGV
Vim: Finished.
[1]    18417 segmentation fault  vim -u NONE test.txt
```

It seems that the large value for `id` causes Vim to crash.

**Steps to Reproduce**:
1. Open Vim with a minimal configuration: `vim -u NONE test.txt`
2. Execute the following commands:
   ```
   :call prop_type_add('my_test', {})
   :call prop_add(1, 1, {'id': 2147483648, 'type': 'my_test'})
   ```
3. Vim crashes with a segmentation fault after executing `prop_add`.

### Expected behaviour

Vim should handle large numbers gracefully without crashing, or at least provide a meaningful error message instead of causing a segmentation fault.

### Version of Vim

9.1

### Environment

macOS version - arm64

### Logs and stack traces

_No response_

chrisbra:
Problem:  Crash when id > INT_MAX in text_prop_add()
Solution: Error out if the id is > INT_MAX

fixes: #15637

### DIFF
--- a/src/textprop.c
+++ b/src/textprop.c
@@ -372,7 +372,16 @@ f_prop_add_list(typval_T *argvars, typval_T *rettv UNUSED)
     type_name = dict_get_string(dict, "type", FALSE);
 
     if (dict_has_key(dict, "id"))
-	id = dict_get_number(dict, "id");
+    {
+	vimlong_T x;
+	x = dict_get_number(dict, "id");
+	if (x > INT_MAX || x  <= INT_MIN)
+	{
+	    semsg(_(e_val_too_large), dict_get_string(dict, "id", FALSE));
+	    return;
+	}
+	id = (int)x;
+    }
 
     if (get_bufnr_from_arg(&argvars[0], &buf) == FAIL)
 	return;
@@ -500,7 +509,7 @@ prop_add_common(
     {
 	vimlong_T x;
 	x = dict_get_number(dict, "id");
-	if (x > INT_MAX)
+	if (x > INT_MAX || x  <= INT_MIN)
 	{
 	    semsg(_(e_val_too_large), dict_get_string(dict, "id", FALSE));
 	    goto theend;
--- a/src/textprop.c
+++ b/src/textprop.c
@@ -497,7 +497,16 @@ prop_add_common(
 	end_col = 1;
 
     if (dict_has_key(dict, "id"))
-	id = dict_get_number(dict, "id");
+    {
+	vimlong_T x;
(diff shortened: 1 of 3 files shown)


src/textprop.c
```diff
@@ -497,7 +497,16 @@ prop_add_common(
...
-	id = dict_get_number(dict, "id");
+    {
+	vimlong_T x;
+	x = dict_get_number(dict, "id");
+	if (x > INT_MAX)
+	{
+	    semsg(_(e_val_too_large), dict_get_string(dict, "id", FALSE));
+	    goto theend;
+	}
+	id = (int)x;
```
DominiquePelle-TomTom, line 517: Related to adding text properties, I see a missing closing parenthesis in an example in the doc:
```
diff --git a/runtime/doc/textprop.txt b/runtime/doc/textprop.txt
index 6b46e06df..012d7686b 100644
--- a/runtime/doc/textprop.txt
+++ b/runtime/doc/textprop.txt
@@ -271,7 +271,7 @@ prop_add_list({props}, [{item}, ...])                       *prop_add_list()*
                        call prop_add_list(#{type: 'MyProp', id: 2},
                                        \ [[1, 4, 1, 7],
                                        \  [1, 15, 1, 20],
-                                       \  [2, 30, 3, 30]]
+                                       \  [2, 30, 3, 30]])
 <
                        GetProp()->prop_add_list([[1, 1, 1, 2], [1, 4, 1, 8]])
```

DominiquePelle-TomTom: I just see that `prop_add_list()` also crashes after this patch (and before the patch too). E.g.

```
:help 
:call prop_type_add('my_test', {})
:call prop_add_list(#{type: 'my_test', id: 2147483648}, [[1, 1, 1, 1]])

Vim: Caught deadly signal SEGV
Vim: Finished.
Segmentation fault: 11
```
It requires the same kind of fix at textprop.c:375 in `f_prop_add_list()`:
```
   374     if (dict_has_key(dict, "id"))
   375         id = dict_get_number(dict, "id");
```

chrisbra: thanks, fixed both as well. Also tested for negative INT_MIN

PR #15639, opened 2024-09-07
Fix session test by checking XDG_CONFIG_HOME too

jmdevin:
I haven't written vim tests before (I just finally got sick of this one
failing), so if someone who actually knows what they're doing could look this
over, I would appreciate it.

### DIFF
--- a/src/testdir/setup.vim
+++ b/src/testdir/setup.vim
@@ -34,6 +34,10 @@ if 1
   " defaults before we get here, and test_mksession checks that.
   let $ORIGHOME = $HOME
 
+  if !exists('$XDG_CONFIG_HOME')
+    let $XDG_CONFIG_HOME = $HOME .. '/.config'
+  endif
+
   " Make sure $HOME does not get read or written.
   " It must exist, gnome tries to create $HOME/.gnome2
   let $HOME = getcwd() . '/XfakeHOME'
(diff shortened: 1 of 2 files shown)


chrisbra:
So it fails for you, because you are using ~/.config/vim directory in your system?

jmdevin:
Correct. I'm not quite sure why `make test` would care which I'm using, but it seems to.

chrisbra:
Thanks. However I think this is slightly simpler?
```diff
diff --git a/src/testdir/test_mksession.vim b/src/testdir/test_mksession.vim
index 914d1e3a4..cf4dcae71 100644
--- a/src/testdir/test_mksession.vim
+++ b/src/testdir/test_mksession.vim
@@ -1270,7 +1270,7 @@ func Test_mkview_default_home()
     " use escape() to handle backslash path separators
     call assert_match('^' .. escape($ORIGHOME, '\') .. '/vimfiles', &viewdir)
   elseif has('unix')
-    call assert_match('^' .. $ORIGHOME .. '/.vim', &viewdir)
+    call assert_match('^' .. $ORIGHOME .. '/' .. '\(\.config/\|\.\)' ..  'vim/', &viewdir)
   elseif has('amiga')
     call assert_match('^home:vimfiles', &viewdir)
   elseif has('mac')
```

jmdevin:
It's simpler, but I believe it would fail if I had $XDG_HOME_CONFIG set to something other than the default of `$HOME/.config`.

chrisbra:
okay fair enough

PR #15640, opened 2024-09-07
Several fixes for GetLatestVimScripts

MiguelBarro:
GetLatestVimScripts builtin plugin required several fixes:
+ Support for the new vimbal `.vmb` extension introduced in [patch 9.0.1797: Vimball/Visual Basic filetype detection conflict](https://github.com/vim/vim/commit/f97f6bbf56408c0c97b4ddbe81fba858d7455b0d).
+ Update the urls from the old `sourceforge.net` to `vim.org`. The download url was hardcoded and a new variable is introduced.
+ Fix `curl` command line option to set a filename (`-O` uses the remote filename and `-o` specifies a filename).
+ Replace windows' command to move files. `REN` can only rename files and the script actually moves them. My educated guess was that originally only renaming was necessary. When the script was modified to move files no change was required on linux because `mv` does both.
+ Fix Autoinstall support to check `ftplugins` and `pack-plugins` too (`pack-plugins` did not exist when the plugin was created).

### DIFF
--- a/runtime/autoload/getscript.vim
+++ b/runtime/autoload/getscript.vim
@@ -212,9 +212,10 @@ fun! getscript#GetLatestVimScripts()
 "  call Decho("searching plugins for GetLatestVimScripts dependencies")
   let lastline    = line("$")
 "  call Decho("lastline#".lastline)
-  let firstdir    = substitute(&rtp,',.*$','','')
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks. I am surprised people are still using GetLatestVimScripts. It's seems to be mostly obsolete now, that people tend to use plugin managers.

PR #15642, opened 2024-09-08
runtime:compiler:pandoc: use latest &spelllang/filetype for :compiler

Konfekt:
Previously these would be cached in buffer-local variables and would not change on :compiler pandoc

### DIFF
--- a/runtime/compiler/pandoc.vim
+++ b/runtime/compiler/pandoc.vim
@@ -50,12 +50,12 @@ silent! function s:PandocLang()
 endfunction
 
 execute 'CompilerSet makeprg=pandoc'..escape(
-    \ ' --standalone' .
+    \ ' --standalone'..
     \ (s:PandocFiletype(&filetype) ==# 'markdown' && (getline(1) =~# '^%\s\+\S\+' || (search('^title:\s+\S+', 'cnw') > 0)) ?
-    \ '' : ' --metadata title=%:t:r:S') .
-    \ ' '..s:PandocLang() .
-    \ ' --from='..s:PandocFiletype(&filetype) .
-    \ ' '..get(b:, 'pandoc_compiler_args', get(g:, 'pandoc_compiler_args', '')) .
+    \ '' : ' --metadata title=%:t:r:S')..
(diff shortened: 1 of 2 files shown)


Konfekt:
I suspect the build failure

> Failures: 
	From test_normal.vim:
	Found errors in Test_mouse_shape_after_cancelling_gr():
	command line..script D:/a/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_mouse_shape_after_cancelling_gr line 26: Expected ['beam', 'arrow'] but got ['arrow', 'arrow']

to be unrelated to this PR

runtime/compiler/pandoc.vim
```diff
@@ -40,18 +42,21 @@ silent! function s:PandocFiletype(filetype) abort
...
-      \ ' --from='..b:pandoc_compiler_from .
-      \ ' '..get(b:, 'pandoc_compiler_args', get(g:, 'pandoc_compiler_args', '')) .
-      \ ' --output %:r:S.$* -- %:S', ' ')
+    \ ' --standalone' .
+    \ (s:PandocFiletype(&filetype) ==# 'markdown' && (getline(1) =~# '^%\s\+\S\+' || (search('^title:\s+\S+', 'cnw') > 0)) ?
+    \ '' : ' --metadata title=%:t:r:S') .
+    \ ' '..s:PandocLang() .
+    \ ' --from='..s:PandocFiletype(&filetype) .
+    \ ' '..get(b:, 'pandoc_compiler_args', get(g:, 'pandoc_compiler_args', '')) .
+    \ ' --output %:r:S.$* -- %:S', ' ')
```
dkearns, line 59: You're mixing `.` and `..` concatenation operators, the latter is preferred.  You might also want to add whitespace around them but I think we generally leave minor style issues up to the authors.

dkearns (commented):
Thanks for the update.

runtime/compiler/pandoc.vim
Konfekt: Alright, went for twice the points

chrisbra:
thanks

PR #15646, opened 2024-09-08
runtime:compiler:groff: Compile groff MOM files to PDF, HTML, ...

Konfekt:
Groff MOM (Macros for Manuscripts) is a macro package for the GNU troff (groff) typesetting system, a light-weight alternative to LaTeX for professional-quality documents.

### DIFF
--- a/runtime/compiler/groff.vim
+++ b/runtime/compiler/groff.vim
@@ -1,0 +1,45 @@
+" Vim compiler file
+" Compiler:     Pandoc
+" Maintainer:   Konfekt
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

chrisbra:
I suppose this compiler plugin is actually for Groff and not Pandoc? Also the documentation refers to `b:pandoc_compiler_lang` both seem like copy-paste errors ;)

I'll fix it while merging (and add a Maintainers entry).

PR #15648, opened 2024-09-09
Properly fix Python 3.13 stable ABI linking

### LINKED ISSUE #15460: VIm doesn't compile with --with-python3-stable-abi=3.11 and newer
### Steps to reproduce

(the same happens with --with-python3-stable-abi=3.12 and --with-python3-stable-abi=3.13)
1. ./configure --with-features=huge --enable-python3interp=dynamic --enable-gui=no --with-python3-stable-abi=3.11
2. make

[error.txt](https://github.com/user-attachments/files/16560230/error.txt)

The errors can be fixed by converting the argument to PyObject*, however then linker errors shows up - 
[linker-error.txt](https://github.com/user-attachments/files/16560244/linker-error.txt)

### Expected behaviour

Vim compiles with the latest Python stable ABI

### Version of Vim

9.1.660

### Environment

OS Fedora 41, but IMO effective on any OS with the mentioned Python and VIm versions

### Logs and stack traces

_No response_

ychin:
This fixes #15460 properly. There was an attempt to fix it in v9.1.0668, but it did it by manually copy and pasting definitions from Python 3.13 headers, which is problematic as it makes Vim dependent on low-level implementation details which are subject to change. That change also pulls in dependencies to private APIs (`_Py_IncRef`) which is a very bad idea as the next version of Python could very well remove that.

The core issue was simply that `Py_INCREF` and similar functions are not part of the stable API. We are supposed to be using `Py_IncRef` instead which performs null-check (similar to `Py_XINCREF`) and is available as a linkable function. We simply need to call it instead of the macro. We simply remap `Py_INCREF` (and friends) to the function version in limited API similar to how we mapped other functions.

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -219,16 +219,9 @@ static HINSTANCE hinstPy3 = 0; // Instance of python.dll
 # define PyObject_GetItem py3_PyObject_GetItem
 # define PyObject_IsTrue py3_PyObject_IsTrue
 # define PyModule_GetDict py3_PyModule_GetDict
-# if defined(USE_LIMITED_API) \
-    && (Py_LIMITED_API+0 >= 0x030c0000 || defined(Py_REF_DEBUG))
-#  undef Py_INCREF
-#  if Py_LIMITED_API+0 >= 0x030a00A7
-#   define _Py_IncRef py3__Py_IncRef
-#   define Py_INCREF _Py_IncRef
-#  else
-#   define Py_IncRef py3_Py_IncRef
-#   define Py_INCREF Py_IncRef
-#  endif
+# ifdef USE_LIMITED_API
+#  define Py_IncRef py3_Py_IncRef
+#  define Py_DecRef py3_Py_DecRef
 # endif
 # ifdef USE_LIMITED_API
 #  define Py_CompileString py3_Py_CompileString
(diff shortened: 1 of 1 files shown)


ychin:
Just a comment: I have noticed a lot of "Python fixes" in recent years for new Python versions tend to be along the variety of "copy whatever we can from the latest Python header and paste it to Vim's code". This is very problematic as they tend to rely on specific versions' implementation details and hard to detect issues in the future (no one is going to diff each Python's release against our inlined copies). There are usually better ways to solve the problem, especially when dealing with limited API. The *entire* reason for the limited API existing is that it provides a limited set of API that are guaranteed to be stable, meaning we aren't supposed to need to constantly fuss with it.

In this case, it just happens that we weren't using limited API properly (since we were calling the macro `Py_INCREF` instead of the proper API `Py_IncRef`). We should try to just do KISS and call the functions as we are expected to instead of trying to be clever and pulls in all kinds of new dependencies to private APIs and so on.

I kind of ran into this as well when doing #13359 which was also trying to prevent Vim from copy/pasting more inlined headers from Python 3 which would simply add more potential sources of breakages in the future.

chrisbra:
Thanks for figuring this out. I really hope this makes our live in the future easier. But we have seen too many of those breakages with python or perl and those patches always look a bit like black magic :)

ychin:
Thanks for merging! My own opinion the ad-hoc nature of a lot of our fixes are contributing to the black magic feeling of things. If there are magical ifdef like convoluted version checks, or if there are blatantly copied code from the Python codebase, generally that's a bad smell, and if there are compelling reasons to really have to add them we should at least have a policy of mandatory comments to explain where these come from and why. Otherwise 2 years after a change goes in and they break it adds a lot of time when you have to sleuth through commit history to find out why.

PR #15649, opened 2024-09-09
Fix Python 3.13 dynamic linking (when not using stable ABI)

### LINKED ISSUE #15457: Vim Python3 plugin does not work with Python 3.13 unstable ABI
### Steps to reproduce

0. Install python 3.13 - https://www.python.org/downloads/release/python-3130rc1/
1. ./configure --with-features=huge --enable-python3interp=dynamic --enable-gui=no
2. make
3. ./src/vim -c 'py3 print(sys.version)'

Error message is shown:
Error detected while processing command line:
E448: Could not load library function _PyObject_NextNotImplemented
E263: Sorry, this command is disabled, the Python library could not be loaded.

The mentioned function symbol was part of earlier Python3 versions, but it is not present anymore in 3.13.

### Expected behaviour

Python 3 plugin works.

### Version of Vim

9.1.660

### Environment

OS: Discovered in Fedora 41, since it includes Python 3.13 already, but it can be reproduced on any Linux system if the provided steps are followed.

### Logs and stack traces

_No response_

ychin:
Vim is importing a private Python symbol `_PyObject_NextNotImplemented` because it used to be required as part of the `PyIter_Check()` macro in an abstraction breaking way. Python eventually fixed the issue and in 3.13 it removed the private symbol export, which broke Vim. Simply remove importing this private symbol in newer Python versions as it's no longer needed for PyIter_Check to work.

Fix #15457

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -267,7 +267,8 @@ static HINSTANCE hinstPy3 = 0; // Instance of python.dll
 # define _Py_NoneStruct (*py3__Py_NoneStruct)
 # define _Py_FalseStruct (*py3__Py_FalseStruct)
 # define _Py_TrueStruct (*py3__Py_TrueStruct)
-# ifndef USE_LIMITED_API
+# if !defined(USE_LIMITED_API) && PY_VERSION_HEX < 0x030D0000
+// Private symbol that used to be required as part of PyIter_Check.
 #  define _PyObject_NextNotImplemented (*py3__PyObject_NextNotImplemented)
 # endif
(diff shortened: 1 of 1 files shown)


ychin:
For reference, Python fixed the reliance on the private symbol in 3.10 in https://github.com/python/cpython/commit/cc54001c2eb3b14320c1667b22602d69c90d5865.

Then in 3.13 they removed the export of the private symbol (https://github.com/python/cpython/commit/1e501122876f7e6abf979fdb756d7edd0d924ba1) which was what's causing Vim to break.

PR #15652, opened 2024-09-09
runtime(java): Recognise all available standard doclet tags

zzzyxwvut:
* Complement the tag set with `@spec`, `{@systemProperty}`,  
  `{@summary}`, `@provides`, `@uses`, `@hidden`, and `{@index}`.
* Do not hoard all tags under a single highlighting group.
* Skip over nested balanced braces in inline tags.
* Observe that tag names are case sensitive: both `{@docRoot}`  
  and `{@inheritDoc}` are valid, whereas `{@inheritdoc}` and  
  `{@docroot}` are not.
* In the `@see` tag arguments, allow for:
  - module name prefixes (e.g. `java.base/java.lang.String`);
  - references to arbitrary URI fragments (e.g. `##foo`);
  - matching any tag variation arguments on the next line.
* Test directives and tags for Java module declarations.
* Enforce the word end for "module-info" candidates.

References:
https://bugs.openjdk.org/browse/JDK-8226279 (`@spec`)
https://bugs.openjdk.org/browse/JDK-8214559 (`{@systemProperty}`)
https://bugs.openjdk.org/browse/JDK-8173425 (`{@summary}`)
https://bugs.openjdk.org/browse/JDK-8160196 (`@provides` & `@uses`)
https://bugs.openjdk.org/browse/JDK-8073100 (`@hidden`)
https://bugs.openjdk.org/browse/JDK-8044243 (`{@index}`)
https://docs.oracle.com/en/java/javase/21/docs/specs/javadoc/doc-comment-spec.html
https://github.com/openjdk/jdk/blob/jdk-21-ga/src/jdk.compiler/share/classes/com/sun/source/doctree/DocTree.java


zzzyxwvut:
Thankfully, syntax tests can be configured in a Vim script;  
test file renaming would have been pretty awkward in the  
header comments.  Another unrelated, nice property is that  
the generated screen dumps get to keep their checked-in  
names, only the buffer is renamed.

chrisbra:
thanks!

PR #15653, opened 2024-09-09
Use stable IncRef / DecRef in Python 3 dynamic bindings

ychin:
Previously, we were using the Py_DECREF macros even when using dynamic linking of Python 3. This caused issues because Python's headers contain references to internal APIs like `_Py_Dealloc` and in v8.1.2201 and v8.2.1225 we simply hacked around the issue by manually copying the Python header implementation to Vim and linking in the private APIs. This is fragile and prone to break. In fact, the Py_DECREF implementation is different in newer versions of Python meaning that this could potentially cause memory issues.

Instead, simply use the API versions (`Py_DECREF` and `Py_INCREF`) which are functions exposed by the Python library. They could be slightly slower since they require a function call instead of a macro, but are much more reliable and we should only be calling these when the Python Vim plugins are crossing the language boundary anyway which are always going to be slow.

Note that this only affects dynamically linked Python builds that are not using stable ABI.

Also see #15648

### DIFF
--- a/src/if_python3.c
+++ b/src/if_python3.c
@@ -219,7 +219,7 @@ static HINSTANCE hinstPy3 = 0; // Instance of python.dll
 # define PyObject_GetItem py3_PyObject_GetItem
 # define PyObject_IsTrue py3_PyObject_IsTrue
 # define PyModule_GetDict py3_PyModule_GetDict
-# ifdef USE_LIMITED_API
+# if defined(USE_LIMITED_API) || PY_VERSION_HEX >= 0x03080000
 #  define Py_IncRef py3_Py_IncRef
 #  define Py_DecRef py3_Py_DecRef
 # endif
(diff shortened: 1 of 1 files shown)


ychin:
This is a continuation of https://github.com/vim/vim/pull/15648, which tries to simplify the Python 3 bindings and remove as much of the copy-pasted Python implementations as possible. As I mentioned above, we should basically never copy implementations from Python as-is like this, because Python's implementations drift over versions and we were essentially having a different DecRef implementation that why Python was expecting.

I do start to wonder (I think some other people raised this before too) whether it still makes sense to support the non-stable ABI version. The way the Python headers are designed assume you don't dynamically load a library like Vim, which is the source of a lot of our linking issues. They just assume that if you dynamically load a library, you should use publicly exposed APIs which are usually stable. But I don't see a real need to change that for now. I think we can just slowly migrate to using more stable APIs which are always exposed in the library and guaranteed to be stable across versions.

k-takata:
Thank you for fixing the Stable ABI issues.
Maybe we can revert v9.1.0693 and v9.1.0691 to use Stable ABI by default again?

ychin:
> Maybe we can revert v9.1.0693 and v9.1.0691 to use Stable ABI by default again?

I'm not sure. There was the seg fault when using stable ABI which I think might be fixed but it will require confirmation since it's on Slackware.

Another thing is even if we are defaulting to stable ABI still means we have to support non-stable ABI unless we explicitly decide to drop it. I guess if that's the long-term decision maybe we could slowly move towards it... The tricky thing is to choose a high enough version to default to since newer versions (e.g. 3.10) allow for better functionality.

chrisbra:
Thanks! My personal opinion is we should slowly move towards using the stable-abi by default if possible. I am hoping this will simplify maintaining the code even if we still have to support the unstable ABI as well.

ychin:
Right. It does make sense after having looked at more of these build failures. The Python header expects there to be a linked library as the header macros use private symbols pretty liberally. This just breaks when you have to manually load an external library and manually load in the symbols you want after at runtime rather than link time. So in a way the current way we are using for dynamic Python builds is kind of wrong, and stable ABI is really how they expect the library to be used in such situations.

Note that we will still need to support non-stable API when using a normal non-dynamic build. But then the build rarely fails in those situations anyway since you are linking against Python normally where the lib is available in link time.

chrisbra:
thanks!

PR #15656, opened 2024-09-10
Revert outdated comment in completeopt's fuzzy documentation

ychin:
Originally, `:set completeopt+=fuzzy` did not affect how the candidate list is collected in default keyword completion. A comment was added to documentation as part of #14912 to clarify it. #15193 later changed the fuzzy behavior to now change the candidate collection behavior as well so the clarification in docs is now wrong. Remove them here.


ychin:
There were quite a few follow-up comments/issues in https://github.com/vim/vim/pull/15193 about whether this is the desired behavior etc. Regardless, the documentation should be correct. I was a little confused when going through the docs trying to understand the behavior myself while trying to understand exactly what the current fuzzy insert mode completion behavior is.

chrisbra:
Thanks. Yes this fell through the cracks :/

PR #15658, opened 2024-09-10
fix(runtime): add .swiftinterface to recognized filetypes

LosFarmosCTL:
.swiftinterface files are distributed inside of binary Swift modules and include type/method definitions, similar to header files in C/C++/etc.
While you are never expected to write such a file yourself, things like the Swift LSP will open them when looking at declarations of symbols in a source-closed library (e.g. Apples UI frameworks).


chrisbra:
thanks!

PR #15664, opened 2024-09-10
runtime(java): Highlight javaConceptKind modifiers with StorageClass

zzzyxwvut:
Stop assigning by default the `NonText` highlighting group for
`javaConceptKind` modifiers since its colour is hardly
distinguishable from a background colour for a range of
colour schemes.


chrisbra:
thanks!

PR #15668, opened 2024-09-11
Update starting.txt

sol:


runtime/doc/starting.txt
```diff
@@ -1078,7 +1078,7 @@ recommended to add these lines somewhere near the top: >
 	unlet! skip_defaults_vim
 	source $VIMRUNTIME/defaults.vim
 Then Vim works like before you had a .vimrc. Copying $VIMRUNTIME/vimrc_example
-is way to do this.  Alternatively, you can copy defaults.vim to your .vimrc
+is one way to do this.  Alternatively, you can copy defaults.vim to your .vimrc
```
sol, line 1081: Or maybe "the preferred"?

chrisbra:
coincidentally, there is also #15663 which addresses the same issue. But thanks!

PR #15670, opened 2024-09-11
runtime(syntax-tests): Support running a subset of syntax tests

zzzyxwvut:
Two methods of assembling a subset of test filenames for  
selection are provided:

- Filename and filetype Make targets will be generated, and  
  multiple such targets can be passed before the mandated  
  trailing `test` target, e.g. `make html markdown test`.

- Filenames and their parts can be specified as a regular  
  expression that is assigned to a `VIM_SYNTAX_TEST_FILTER`  
  environment variable, and used with the `test` Make target,  
  e.g. `VIM_SYNTAX_TEST_FILTER=html\\\|markdown make test`.
  (This variable will be ignored and the whole suite will be  
  run when Make is GNU Make and a parent Makefile is used.)

Methods can be used alone or together, with the Make targets  
having the higher precedence.  Neither method will influence  
the order of test execution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,9 @@ indenttest:
 # Executable used for running the syntax tests.
 VIM_FOR_SYNTAXTEST = ../../src/vim
 
+# (For local testing only with GNU Make.)
+VIM_SYNTAX_TEST_FILTER =
+
 syntaxtest:
 	cd runtime/syntax && \
 		$(MAKE) clean && \
--- a/runtime/syntax/Makefile
+++ b/runtime/syntax/Makefile
@@ -31,10 +31,38 @@ test:
 	@echo "../$(VIMPROG)" > testdir/vimcmd
 	@echo "$(RUN_VIMTEST)" >> testdir/vimcmd
 	VIMRUNTIME=$(VIMRUNTIME) $(ENVVARS) $(VIMPROG) --clean --not-a-term $(DEBUGLOG) -u testdir/runtest.vim
+	@rm -f testdir/Xfilter
 	@# FIXME: Temporarily show the whole file to find out what goes wrong
 	@#if [ -f testdir/messages ]; then tail -n 6 testdir/messages; fi
 	@if [ -f testdir/messages ]; then cat testdir/messages; fi
 
 
 clean testclean:
-	rm -f testdir/failed/* testdir/done/* testdir/vimcmd testdir/messages testdir/Xtestscript
+	rm -f testdir/failed/* testdir/done/* testdir/vimcmd testdir/messages testdir/Xtestscript testdir/Xfilter
+
+# All initial phony targets; these names may clash with file extensions.
+phonies = clean test testclean
+
+# Collect all input filenames and their file extensions.
+testnames != set +f; \
+awk 'BEGIN { \
+    for (i = 1; i < ARGC; i++) { \
+        split(ARGV[i], names, /\//); \
+        split(names[3], parts, /\./); \
+        exts[parts[2]]; \
+        print names[3]; \
+    } \
+    split("$(phonies)", scratch); \
+    for (phony in scratch) \
+        phonies[scratch[phony]]; \
+    for (ext in exts) \
+        print ext ((ext in phonies) ? "_" : ""); \
+}' testdir/input/*.*
+
+.PHONY: self-testing $(testnames)
+
+$(testnames)::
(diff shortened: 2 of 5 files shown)


zzzyxwvut:
As a _local_ alternative to always running the whole syntax  
test suite, consider running a single test when trying out  
changes _or_ consider running a batch of failing tests when  
investigating CI failures.  This can become possible by  
introducing an environment variable (Make is not very good  
at non-target arguments) and interpreting it as a regexp  
filename filter.

@chrisbra, @dkearns, what do you think?

zzzyxwvut:
I'm a bit paranoid about CI environment variables (GitHub  
and Cirrus).  Having an unrelated prefix may prevent some  
future clashes, wouldn't it?

zzzyxwvut:
Let's try completable Make targets as an alternative to  
initially proposed environment variables.

Please consider:

```shell
make java_com<Tab> java_n<Tab> test
make ja<Tab> test
```

as opposed to:

```
VIM_SYNTAX_TEST_FILTER='^java_\%(comments\|numbers\)' make test
VIM_SYNTAX_TEST_FILTER='\.java$' make test
```

One apparent shortcoming is the lack of support for regexp  
like Make targets, so one would have to Tab through every  
selected `foo_bar.foo`, `foo_bar_A.foo`, etc. _or_ settle for  
the whole `foo` suite rather than picking a prefix pattern  
`VIM_SYNTAX_TEST_FILTER='^foo\>_\='`. (Admittedly, target  
names can be treated as patterns, but then it would preclude  
the running of the shortest prefix-name test on its own.)   
Despite that, I find targeted selection to be superior for  
running single tests and filetype test suites.

Another thing to remember is that filename and filetype Make  
targets require the explicit `test` target to follow as they  
only update `testdir/Xfilter` otherwise.

Hopefully, there are no valid file extensions containing `_`,  
else the proposed implementation is faulty.

zzzyxwvut:
Provided there is but one `.` in a filename, the distribution  
of `_`s in either basenames or extensions is now immaterial.

As before, failing syntax tests can be disabled on demand by  
renaming them, with non-`.`s replacing `.`s; e.g. from `c.c` to  
`c_c`.  Filenames without `.`s will not be tested and Make  
targets will not be generated with their names.

One thing to be aware is that there is no association among  
completion candidates for a filetype and its test files.  So  
for Modula, `def` wouldn't be offered as a candidate when  
`modula2_` is Tab'd.

zzzyxwvut:
ASIDE.

While locally testing my latest change, I run into another  
kind of syntax test failure: the new screen dump reported  
the cursor position disparity (`>` and `<`).  The asynchronous  
nature of this failure makes it hard to reproduce it.

The fortuitous race between parent and child Vim processes  
can take place around here:

https://github.com/vim/vim/blob/7cfe693f9bfa74690867e4d96c25f2205d0d13e4/runtime/syntax/testdir/runtest.vim#L350-L353

Maybe we should `sleep` a wee bit more?

https://github.com/vim/vim/blob/7cfe693f9bfa74690867e4d96c25f2205d0d13e4/src/testdir/screendump.vim#L81-L87

chrisbra:
yes, can you try to increase the sleep?

zzzyxwvut:
50 millis?

zzzyxwvut:
It is the product of shell's support for command completion  
along with a shell completion function written for Make, _and_  
Make's capability for providing necessary information for  
completion to be useful.

In order to make command completion work for a NetBSD port  
of Make that lacks the `--print-data-base` (or `-p`) flag  
provided by GMake and which is used by the completion shell  
function in Bash, we can just query GMake:

```shell
unset -f bmake
bmake cl<Tab>
bmake() { case "$1" in -*p* | -p) command -p gmake "$@" ;; *) command -p bmake "$@"; echo >&2 BMAKING... ;; esac; }
bmake cl<Tab>
```

zzzyxwvut:
The top comment have been updated to match the progress so  
far.

zzzyxwvut:
Is there anything else to be addressed?

chrisbra:
Thanks. I think this is fine. Can you move it to ready please?

zzzyxwvut:
Would you like me to rebase the branch and squash all WIP  
commits into the current HEAD?

chrisbra:
yes please

zzzyxwvut:
PO checking on FreeBSD fails.

chrisbra:
thanks!

PR #15672, opened 2024-09-12
runtime(filetype)!: unify case sensitive extension matching

echasnovski:
Problem: there are different approaches of how extensions are matched
  with respect to case sensitivity. In particular, '\c' flag is used in
  pattern whereas in most places case sensitive matching is guarded
  behind `has("fname_case")` condition.
Solution: replace all instances of '\c' with an explicit case sensitive
  pattern variants guarded by `has("fname_case")`.
  Strictly speaking, this is a breaking change because only two (most
  common and prevailingly tested) variants are now matched: upper first
  letter and upper all letters.

------

Notes:
- All seven case insensitive extension matches were added in first half of April 2022: https://github.com/vim/vim/commit/3ad2090316edc85e93094bba7af64f9991cc7f85 and https://github.com/vim/vim/commit/0bbf09ca41382302493e5db51b01d2fbdc778586. I didn't find any mention of why all 8 variants for extensions should be recognized, so opted for what is (mostly) tested: `xxx`, `Xxx`, and `XXX`.

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -237,7 +237,10 @@ au BufNewFile,BufRead *.bat			setf dosbatch
 au BufNewFile,BufRead *.cmd
 	\ if getline(1) =~ '^/\*' | setf rexx | else | setf dosbatch | endif
 " ABB RAPID or Batch file for MSDOS.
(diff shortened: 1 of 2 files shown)


chrisbra:

echasnovski:
Thanks, Christian!

Issue #15671, opened 2024-09-11
vim.vim: No recognition of :fun or :def bodies commencing with empty lines if g:vimsyn_folding contains "f"


zzzyxwvut:
### Steps to reproduce

After [`fc64ac`](https://github.com/vim/vim/commit/fc64ac0be2c4f9446798de4d66b029526d1afa85#diff-0e98f1b5468aa4ddd31ea819e57fdfc26251f992ee0aeca39a41f3c30b865e6cR288-R289), empty lines at the head of a function body  
inhibit the recognition of this body _if_ `g:vimsyn_folding` 
contains `f`.

For example,
```vim
" VIM_TEST_SETUP let g:vimsyn_folding = 'f'
" VIM_TEST_SETUP setl fdc=2 fdm=marker fmr=(,) fen

fun FA1()


    return
endfun

fun FA2()
    return
endfun

fun FB1() abort

    return
endfun

fun FB2() abort
    return
endfun

fun FC1(_)

    return
endfun

fun FC2(_)
    return
endfun

fun FD1(_) abort

    return
endfun

fun FD2(_) abort
    return
endfun




def MA1()


    return
enddef

def MA2()
    return
enddef

def MB1(): void

    return
enddef

def MB2(): void
    return
enddef

def MC1(_: any)

    return
enddef

def MC2(_: any)
    return
enddef

def MD1(_: any): void

    return
enddef

def MD2(_: any): void
    return
enddef
```

### Expected behaviour

Function bodies should be recognised with or without empty  
lines at the head when `g:vimsyn_folding` contains `f`.

### Version of Vim

v9.1.0727

### Environment

GNU/Linux Debian (testing/trixie)

zzzyxwvut:
One way to fix it is by replacing `^.` with `^\_.`:

```diff
diff --git a/runtime/syntax/vim.vim b/runtime/syntax/vim.vim
index 247663d4d..4052573ee 100644
--- a/runtime/syntax/vim.vim
+++ b/runtime/syntax/vim.vim
@@ -339,8 +339,8 @@ syn match	vimDefParam	contained	"\<\h\w*\>"		skipwhite nextgroup=vimParamType,vi
 syn match	vimFuncParamEquals contained	"="			skipwhite        nextgroup=@vimExprList
[...]

PR #15674, opened 2024-09-12
runtime(vim): Update base-syntax, improve folding function matches

dkearns:
- Allow function command modifiers.
- Match function bodies starting with empty lines.

Fixes issue #15671.

Command modifiers reported by @Konfekt.


chrisbra:
thanks

PR #15679, opened 2024-09-13
Wrong culopt=screenline when resizing window

zeertzjq:
Problem:  Wrong culopt=screenline when resizing window.
Solution: Invalidate saved left_col and right_col when width1 or width2
          change.


chrisbra:
thanks!

PR #15684, opened 2024-09-14
runtime(vim): Update base-syntax, match Vim9 bool/null literal args to :if/:while/:return

dkearns:
Match Vim9 boolean and null literals in expression arguments of :if,
:elseif, :while and :return.


chrisbra:
thanks!

PR #15686, opened 2024-09-15
add lua output to xxd

### LINKED ISSUE #15685: add lua output to xxd
Problem: currently xxd can only output C code
Soution: would like to add an option to output Lua code. other parameters: variable name, capitalization, number of columns etc. should be used in a similar way.

yuvalif:
Problem: currently xxd can only output C code
Soution: add "-L" an option to output Lua code.
other parameters: variable name, capitalization, number of columns etc. should be used in a similar way.

closes: #15685

### DIFF
--- a/src/testdir/test_xxd.vim
+++ b/src/testdir/test_xxd.vim
@@ -268,6 +268,76 @@ func Test_xxd()
     call assert_equal(expected, getline(1,'$'), s:Mess(s:test))
   endfor
 
+  " Test 19: Print Lua file
+  let s:test += 1
+  call writefile(['TESTabcd09'], 'XXDfile')
+  %d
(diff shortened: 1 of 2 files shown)


chrisbra:
Thanks, but I am a bit unsure. lua seems a bit niche. Is this really useful?

yuvalif:
> Thanks, but I am a bit unsure. lua seems a bit niche. Is this really useful?

asked a wide audience in the lua community: https://www.reddit.com/r/lua/comments/1fi9jus/embedding_binary_strings_in_code/

must admit that there is no clear indication that this is a "must have" feature for lua devs

yuvalif:
closing based on above comments

PR #15687, opened 2024-09-15
feat(quickfix): handle name display in qflist for hard links

austin880625:
It is a small enhancement on quickfix list. It currently shows entries with duplicate name if the file is opened with the path of hard links.

This behavior can be produced by creating a file like `main1.cpp` and make its hard link `main2.cpp` then issue the following commands in vim:

```
:call setqflist([{'filename':'main1.cpp', 'lnum':1}, {'filename':'main2.cpp', 'lnum':1}])
:copen
``` 

The major cause is that qflist assumed that the filename passed into `qf_add_entry` matches the filename opened with the buffer.

This patch handles this case by introduce a `qf_fname` into `qfline_S` structure. It stores the filename from `qf_add_entry` for each quickfix line.

I'm not really familiar with vim codebase and I'm redirected from [nvim](https://github.com/neovim/neovim/pull/30368). Please point out if this PR would introduce any issues, thanks!

### DIFF
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -36,6 +36,7 @@ struct qfline_S
     int		qf_end_col;	// column when the error has range or zero
     int		qf_nr;		// error number
     char_u	*qf_module;	// module name for this error
+    char_u	*qf_fname;	// different filename if there're hard links
     char_u	*qf_pattern;	// search pattern for the error
     char_u	*qf_text;	// description of the error
     char_u	qf_viscol;	// set to TRUE if qf_col and qf_end_col is
@@ -2173,14 +2174,17 @@ qf_add_entry(
     typval_T	*user_data,     // custom user data or NULL
     int		valid)		// valid entry
 {
+    buf_T	*buf;
     qfline_T	*qfp;
     qfline_T	**lastp;	// pointer to qf_last or NULL
+    char_u	*fullname = NULL;
+    char_u	*p = NULL;
 
     if ((qfp = ALLOC_ONE_ID(qfline_T, aid_qf_qfline)) == NULL)
 	return QF_FAIL;
     if (bufnum != 0)
     {
-	buf_T *buf = buflist_findnr(bufnum);
+	buf = buflist_findnr(bufnum);
 
 	qfp->qf_fnum = bufnum;
 	if (buf != NULL)
@@ -2188,7 +2192,24 @@ qf_add_entry(
 		IS_QF_LIST(qfl) ? BUF_HAS_QF_ENTRY : BUF_HAS_LL_ENTRY;
     }
     else
+    {
 	qfp->qf_fnum = qf_get_fnum(qfl, dir, fname);
+	buf = buflist_findnr(qfp->qf_fnum);
+    }
+    if (fname != NULL)
+	fullname = fix_fname(fname);
+    qfp->qf_fname = NULL;
+    if (buf != NULL &&
+	buf->b_ffname != NULL && fullname != NULL)
+    {
+        if (fnamecmp(fullname, buf->b_ffname) != 0)
+        {
+            p = shorten_fname1(fullname);
+            if (p != NULL)
+		qfp->qf_fname = vim_strsave(p);
+        }
+    }
+    vim_free(fullname);
     if ((qfp->qf_text = vim_strsave(mesg)) == NULL)
     {
 	vim_free(qfp);
@@ -3709,7 +3730,10 @@ qf_list_entry(qfline_T *qfp, int qf_idx, int cursel)
 	if (qfp->qf_fnum != 0
 		&& (buf = buflist_findnr(qfp->qf_fnum)) != NULL)
 	{
-	    fname = buf->b_fname;
+	    if (qfp->qf_fname == NULL)
+		fname = buf->b_fname;
+	    else
+		fname = qfp->qf_fname;
 	    if (qfp->qf_type == 1)	// :helpgrep
 		fname = gettail(fname);
 	}
@@ -4034,6 +4058,7 @@ qf_free_items(qf_list_T *qfl)
 	qfpnext = qfp->qf_next;
 	if (!stop)
 	{
+	    vim_free(qfp->qf_fname);
 	    vim_free(qfp->qf_module);
 	    vim_free(qfp->qf_text);
 	    vim_free(qfp->qf_pattern);
@@ -4761,7 +4786,10 @@ qf_buf_add_line(
 			mch_dirname(dirname, MAXPATHL);
 		    shorten_buf_fname(errbuf, dirname, FALSE);
 		}
-		ga_concat(gap, errbuf->b_fname);
+		if (qfp->qf_fname == NULL)
+		    ga_concat(gap, errbuf->b_fname);
+		else
+		    ga_concat(gap, qfp->qf_fname);
 	    }
 	}
 
--- a/src/testdir/test_quickfix.vim
+++ b/src/testdir/test_quickfix.vim
@@ -6578,4 +6578,37 @@ func Test_cbuffer_range()
   call XbufferTests_range('l')
 endfunc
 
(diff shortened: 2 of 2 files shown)


src/quickfix.c
```diff
@@ -36,6 +36,7 @@ struct qfline_S
     int		qf_end_col;	// column when the error has range or zero
     int		qf_nr;		// error number
     char_u	*qf_module;	// module name for this error
+    char_u	*qf_fname;	// filename for a line
```
yegappan: Creating a copy of the filename for each quickfix entry will significantly increase the memory used for the quickfix list.  Is this really needed?

src/quickfix.c
```diff
@@ -2175,6 +2176,8 @@ qf_add_entry(
 {
     qfline_T	*qfp;
     qfline_T	**lastp;	// pointer to qf_last or NULL
+    char_u	*fullname = fix_fname(fname);
```
yegappan: Both filenames refer to the same file.  So I am not sure whether this is a significant issue that should be addressed.

src/quickfix.c
austin880625: Thanks for the comment. My bad that I wasn't aware of the memory usage at the time. My thought was that the list is from LSP response. If the filenames already occupy the memory before `setqflist`, then it should be ok to stay there(copy & release the original one by caller) and hold by qflist.

Another approach would be doing a comparison with `buf->b_ffname` and only copy the filename if they are different, or add an option on `setqflist` to tell if caller want it to do so. This still leads to another doubt that if the time of comparison on each entry really worth.

src/quickfix.c
austin880625: I believe my use case(LSP + a build infrastructure that makes hard links) is not very common. I just saw it and try to see if I can make it better. This issue makes me seeing alot of duplicated entries when looking on references of a function. If they should be treated as same file, then the entries shouldn't be duplicated. But this also depends on which component we think should handle this case:

1. vim: This is what my patch tries to deals with.
2. the plugin or LSP: LSP shouldn't send two names in the responses if they are already the same files.
3. Users:  Hard links are OS specific and just make a lot of mess in general. Users shouldn't expect editors/LSPs/plugins to know well on how files are linked.

As I see that vim only shows one filename when the file's hard link is opened(instead of showing another name). I believe this implementation was thought(and tested) more extensively and the better ways to address this issue might be 2. or 3. If that's the case then we can just close this PR.

chrisbra:
This seems a bit obscure. First thing, if you are using hard-links for your source this will likely also confuse your build-system and you can (too easily) get rid of the hardlinks, just by moving it to `/tmp` or similar. 

I share the same opinion about this patch as @yegappan , we will add quite a bit of extra memory. If possible, if we can delay this decision and only do this, once it really matters, perhaps like you mention here:
> Another approach would be doing a comparison with buf->b_ffname 

then this might be acceptable.

austin880625:
I just pushed the version of comparing before copying approach. As of my use case, the example at the beginning is the minimal one. If a building system aims to build binaries for multiple architectures and goes like this:

1. The editor works on `/src/project`
2. Build scripts runs `ln -r /src/project/* /build_env_arch1/project/`. Then the scripts can build the source and generates binaries in different directories(for different archs).
3. LSP runs with the information in compile_commands.json, which is based on each `build_env_arch1/project`. The editor runs on `/src/project` and the files inside are hard linked. Thus they turned out to be duplicated entries in the quicklist when looking up references from LSP.

Above is how I discovered the issue but might not be a expected usage(like LSP and editors might better run at same place). Hope this helps you decide

chrisbra:
okay. Can you please think of a testcase for https://github.com/vim/vim/blob/master/src/testdir/test_quickfix.vim ? I think it shouldn't be too hard. Guard the test with `Check Unix` to make sure it only runs on Unix. Then you should be able to create a single file (`call writefile([], 'file1.c')`) and hard-link it to another file `system('ln file1.c file2.c')` and then using `Xsetlist()` I think. Check the existing test file please

src/testdir/test_quickfix.vim
```diff
@@ -6578,4 +6578,32 @@ func Test_cbuffer_range()
   call XbufferTests_range('l')
 endfunc
 
+" Test for displaying fname pass from setqflist when the name
+" are hard links to prevent seemly duplicate entries.
+func Xtest_hardlink_fname(cchar)
```
yegappan, line 6583: You need to add another function starting with the prefix "Test_" that calls this function with the argument 'c' and 'l'.

src/testdir/test_quickfix.vim
```diff
@@ -6578,4 +6578,32 @@ func Test_cbuffer_range()
...
+  CheckUnix
+  call s:setup_commands(a:cchar)
+  %bwipe
+  " Create a sample source file
+  let lines =<< trim END
+    void sample() {}
+    int main() { sample(); return 0; }
+  END
+  call writefile(lines, 'file1.c')
```
yegappan: Pass the flag 'D' as third argument  to this function.

yegappan: Use a more unique file name than "file1.c" (e.g. Xtesthardlink.c")

src/testdir/test_quickfix.vim
```diff
@@ -6578,4 +6578,32 @@ func Test_cbuffer_range()
...
+  call system('ln file1.c file2.c')
+  call g:Xsetlist([], 'f')
+  " Make a qflist that contains the file and it's hard link
+  " like how LSP plugins set response into qflist
+  call g:Xsetlist([{'filename' : 'file1.c', 'lnum' : 1},
+        \ {'filename' : 'file2.c', 'lnum' : 1}], ' ')
+  copen
```
yegappan: Use "Xopen" instead of "copen".

src/testdir/test_quickfix.vim
```diff
@@ -6578,4 +6578,32 @@ func Test_cbuffer_range()
...
+  " Ensure that two entries are displayed with different name
+  " so that they aren't seen as duplication.
+  call assert_equal(['file1.c|1| ',
+        \ 'file2.c|1| '], getline(1, '$'))
+  cclose
```
yegappan: Use "Xclose" instead of "cclose".

yegappan (changes requested):

austin880625:
just pushed the mentioned changes, thanks.

austin880625:
seems there's a possible leak detected by ASAN, still looking on it

austin880625:
I've fixed the bugs detected by CI workflow and the issues you mentioned in the review. @yegappan let me know if you notice other considerations!

resslr (commented):
accidentally reviewed @yegappan delete this comment please

austin880625:
@chrisbra Now I couldn't figure out how macos huge pipelines affected by my changes. the [previous pipeline](https://github.com/vim/vim/actions/runs/11146454935)(only differs on comment) succeeded and [another one](https://github.com/vim/vim/actions/runs/11056640707/job/30733868719) fails on macos14. I also see few PRs got the same failure on the same testcase. Do you have any clue or experience on the failure of this `Test_popup_beval` testcase?

src/testdir/test_quickfix.vim
```diff
@@ -6578,4 +6578,37 @@ func Test_cbuffer_range()
   call XbufferTests_range('l')
 endfunc
 
+" Test for displaying fname pass from setqflist when the name
+" are hard links to prevent seemly duplicate entries.
+func Xtest_hardlink_fname(cchar)
+  CheckUnix
```
zeertzjq: It may be better to move this line to the start of `Test_hardlink_fname()`

austin880625: updated, thanks

chrisbra:
That is a known flaky test on macos. Don't worry.

chrisbra:
I made a few further improvements to the test, while merging. Thanks all!

PR #15692, opened 2024-09-17
translation(ru): updated manual page xxd-ru

RestorerZ:
Updated manual page files xxd-ru.1 and xxd-ru.UTF-8.1 to Vim patch 9.1.0732


chrisbra:
thanks!

PR #15696, opened 2024-09-18
runtime:ftplugin: let &keywordprg use :term

Konfekt:
See https://github.com/neovim/neovim/pull/30409

### DIFF
--- a/runtime/ftplugin/gpg.vim
+++ b/runtime/ftplugin/gpg.vim
@@ -15,20 +15,12 @@ let b:undo_ftplugin = "setl com< cms< fo<"
 
 setlocal comments=:# commentstring=#\ %s formatoptions-=t formatoptions+=croql
 
-if has('unix') && executable('less')
-  if !has('gui_running') && !has('nvim')
-    command -buffer -nargs=1 GpgKeywordPrg
-          \ silent exe '!' . 'LESS= MANPAGER="less --pattern=''^\s+--' . <q-args> . '\b'' --hilite-search" man ' . 'gpg' |
-          \ redraw!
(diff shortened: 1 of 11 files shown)


habamax:
Not sure if I like the idea of nvim specific checks in vim runtime files.

On the other hand it is already in some of runtime files, including netrw: https://github.com/vim/vim/blob/4298c5f472a05486d6e26988cf10f4d04c80afd2/runtime/plugin/netrwPlugin.vim#L159

it is also in ruby and rust:
- https://github.com/vim/vim/blob/4298c5f472a05486d6e26988cf10f4d04c80afd2/runtime/autoload/rust.vim#L533
- https://github.com/vim/vim/blob/4298c5f472a05486d6e26988cf10f4d04c80afd2/runtime/autoload/rubycomplete.vim#L629

Konfekt:
I understand. To ease maintenance, I  removed the first variant using `:!less` which does not work on Neovim, and try `:!term less`, assuming that most Vim now supply a `:term` command.
This term command output incompatibility is a pain point

chrisbra:
Thanks. Again it would be nicer if the zsh filetype went directly via my upstream repo. But I have fixed that now.

Konfekt:
Yes, I thought of it but wasn't sure there's a preference in order; would've added a PR for your repo, but glad to hear that it's already sorted out

chrisbra:
yes, I want to make it easy for users to use runtime files in both projects and if feasible, I think it is beneficial for both projects to keep the logic more or less in one runtime file. But if upstream runtime file maintainers have a different opinion, then it is ultimatively their decision.

PR #15698, opened 2024-09-18
runtime(java): Optionally recognise all primitive constants in _switch-case_ labels

zzzyxwvut:
Define `g:java_syntax_previews` and include number `455` in  
its list to enable this recognition:

```vim
let g:java_syntax_previews = [455]
```

Reference:
https://openjdk.org/jeps/455


chrisbra:
thanks!

PR #15699, opened 2024-09-19
runtime:ftplugin:(z)sh: set and unset compiler

Konfekt:
Could be applied similarly to other ftplugins as well


chrisbra:
Thanks. 

Next time, please go via the proper upstream repo: https://github.com/chrisbra/vim-zsh

PR #15700, opened 2024-09-19
runtime(netrw): Fix E872 too much '(' in highlight pattern for `mf` selection (fixup for #15551)

KSR-Yasuda:
I introduced #15551, but it causes E872 error for many file selection by `mf`:
since that makes a highlight patten with too much `\(`.

To avoid E872, it should've been `\%(`, that is without submatch reference.

### How to Repro

* Select 10? or more files by `mf` in a netrw window.


chrisbra:
thanks!

KSR-Yasuda:
Thank you for merging!

PR #15709, opened 2024-09-20
runtime(java): Optionally recognise _module_ import declarations

zzzyxwvut:
Define `g:java_syntax_previews` and include number `476` in  
its list to enable this recognition:

```vim
let g:java_syntax_previews = [476]
```

Reference:
https://openjdk.org/jeps/476


chrisbra:
thanks!

Issue #13893, opened 2024-01-20
`:setlocal cmdheight=2` affects global value


Bekaboo:
### Steps to reproduce

1. `vim --clean`
2. `:setlocal cmdheight=2`
3. `:setglobal cmdheight?` --> output: `2`

### Expected behaviour

Global value of `cmdheight` should not be affected by `:setlocal cmdheight=2`, like other global-local options, e.g. `number`:

1. `vim --clean`
2. `:setlocal number`
3. `:setglobal number?` --> output: `nonumber` (not affected by `:setlocal`)

### Version of Vim

VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Jan 02 2024 18:36:25)

### Environment

OS:

```
Linux ThinkPad-P15v-Gen1 6.7.0-arch3-1 #1 SMP PREEMPT_DYNAMIC Sat, 13 Jan 2024 14:37:14 +0000 x86_64 GNU/Linux
```

$TERM: `tmux-256color`
shell: bash

### Logs and stack traces

```shell
N/A
```

chrisbra:
we don't really have tab-local options. It may be a bit of work, to make this work properly and it will make the current already confusing handling of local, global and local-global options even more confusing.

Bekaboo:
@chrisbra Thanks for your reply. According to the doc, `&cmdheight` is tab-local, so does it really mean if we don't really have tab-local options? It does not behave like a global option, nor buffer-local option or window-local option, it is confusing. It will be helpful if we can make it clear er in the doc.

chrisbra:
the option is `global-local`, see `:h global-local`. It's the only option, that is global-local to a tab-window. What I mean with we don't have tab-local options is (and I think it has been discussed in the context of diff mode), is we don't have proper setup for tab-local options at all. `cmdheight` is currently special and is remembered for each tabpage, but that is not how tab-local options should work (from a code perspective). So we would need to properly support tab-local options similar to how window-local options work. However, I fear that will make the option setting and getting code even more complex.

PR #15713, opened 2024-09-20
allow get/set tab-local 'cmdheight' via gettabvar()/settabvar()

### LINKED ISSUE #4324: tab-local 'cmdheight' value cannot be queried via gettab[win]var()
**Describe the bug**
The 'cmdheight' option effectively has tab-local values (the `:help` says _global_, but then clarifies _The value of this option is stored with the tab page, so that each tab page can have a different value._) As such, it should be possible to query the option's current value with `:echo gettabvar(2, '&cmdheight')`. That doesn't work at all; nothing is returned (supposedly the `&optionname` syntax isn't supported there (yet; I think it should)).

Okay, try `:echo gettabwinvar(2, 1, '&cmdheight')`. That one returns a value, but it's always the one for the _current_ tab page, so not good either.

Interestingly, it **is** possible to **set** the value with `:call settabwinvar(2, 1, '&cmdheight', 2)` (from another tab page).

**Expected behavior**
The implementation of the tab-local `'cmdheight'` values seems to be incomplete with regards to the offered Vimscript API functions:
* `:help 'cmdheight'` should clearly signify the scope as _tab-local_; the _global (but ...)_ is confusing.
* `gettabvar()` should take `&optionname` varname(s) and allow to query the (so far only) tab-local option.
* `gettabwinvar()` should allow to query the `&cmdheight` value for an existing window in another tab page, instead of returning the (wrong) value for the current tab page.

 - Vim version 8.1.1242 (latest dev snapshot).
 - OS: Ubuntu 16.04 LTS 64-bit.
 - both terminal and GUI

Milly:
Fixes #4324.
Related #13893.


src/option.h
```diff
@@ -1339,6 +1339,17 @@ enum
     , WV_COUNT	    // must be the last one
 };
 
+/*
+ * "indir" values for tab-local options.
+ * These need to be defined globally, so that the TV_COUNT can be used in the
```
zeertzjq, line 1344: `TV_COUNT` isn't used anywhere

Milly: Yes. Fix the comment or remove the definition?

zeertzjq: Actually I think that means the implementation is incomplete. `BV_COUNT` and `WV_COUNT` are used for arrays that store script contexts that set local options, so if 'cmdheight' needs to be a fully tab-local option it's necessary to store a script context with the tabpage as well.

Milly: I think that should be done in a separate PR about `cmdheight`.
Because there is only one tab-local option.

chrisbra: Yeah, but we should still remember when this option was (re-)set right?

Milly: I'll write another PR to make `cmdheight` as a true tab-local option.
And in the meantime I'd like to mark this PR as a WIP.

chrisbra:
this has stalled, so closing

Issue #15512, opened 2024-08-16
[vim9class] `E1027: Missing return statement` compiling abstract method declaration


errael:
### Steps to reproduce

```
vim9script

abstract class A
    abstract def String(): string
endclass

class B extends A
    def String(): string
        return 'B'
    enddef
endclass
def F(o: A)
    assert_equal('B', o.String())
enddef
F(B.new())
```
generates
```
Error detected while compiling /tmp/d.vim[15]..function <SNR>46_F[1]..<SNR>46_A.String:
line    1:
E1027: Missing return statement
```

### Expected behaviour

No error.

There is PR #15441which fixes this. But it was merged/reverted through miscommunication, and now since the branch is deleted, the PR can not be re-opened. (There was no original issue for it) 

I'm attaching the patch file from that PR in case...

### Version of Vim

9.1.678

### Environment

ubuntu/gtk3

errael:
I did wget, but the patch file from the PR won't attatch. Here's the text

```patch
From 4be245a767cd2460924c727cf7c343d30eba7c7e Mon Sep 17 00:00:00 2001
From: Ernie Rael <errael@raelity.com>
Date: Tue, 6 Aug 2024 18:23:17 -0700
Subject: [PATCH] Fix #15432

During compilation, don't look for a return statement for an abstract method.
---
 src/testdir/test_vim9_class.vim | 21 +++++++++++++++++++++
 src/vim9compile.c               |  5 +++--
 2 files changed, 24 insertions(+), 2 deletions(-)

diff --git a/src/testdir/test_vim9_class.vim b/src/testdir/test_vim9_class.vim
index 56c7b99d943e0..4d860685dc3fa 100644
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -6129,6 +6129,27 @@ def Test_abstract_method()
     assert_equal('foo', A.Foo())
   END
   v9.CheckSourceSuccess(lines)
+
+  # Invoke method returning a value through the abstract class. See #15432.
+  lines =<< trim END
+    vim9script
+
+    abstract class A
+        abstract def String(): string
+    endclass
+
+    class B extends A
+        def String(): string
+            return 'B'
+        enddef
+    endclass
+
+    def F(o: A)
+        assert_equal('B', o.String())
+    enddef
+    F(B.new())
+  END
+  v9.CheckSourceSuccess(lines)
 enddef
 
 " Test for calling a class method from a subclass
[...]

Issue #15484, opened 2024-08-13
[vim9class] Unexpected method dispatch when type is interface


errael:
### Steps to reproduce

Note this is `vim-9.1` behavior.

Came across this sorting out some recent issues. I was surprised, wrote the test in `Java` and got the expected result.
@dkearns what do you think? @zzzyxwvut @chrisbra @yegappan note.
```
vim9script

interface I
    def F(): string
endinterface

class A implements I
    def F(): string
        return 'A'
    enddef
endclass

class C extends A
    def F(): string
        return 'C'
    enddef
endclass

# Unexpected: following outputs 'A'. Expect 'C'
def TestI(i: I)
    echo i.F()
enddef
TestI(C.new())
```
outputs
```
A
```

Was looking at`Abstract method overrides concrete method`, but that's not related to this (maybe loosely related).

### Expected behaviour

Expect dispatch of function `F()` on an object to be "dynamic dispatch".  For example, wrote the above test in `Java` and it works as expected.

Another example, this works as expected.
```
# Expected: following outputs 'C'.
def TestA(a: A)
    echo a.F()
enddef
TestA(C.new())
```

### Version of Vim

9.1.661

### Environment

ubuntu/gtk3

errael:
And in the example if  `class C extends A` is changed to
```
class C extends A implements I
```
Then it dispatches correctly.

errael:
> And in the example if `class C extends A` is changed to
> 
> ```
> class C extends A implements I
> ```
> 
> Then it dispatches correctly.

And BTW, with the OP example, doing
```
echo 'instanceof(C.new(), I): ' instanceof(C.new(), I)
```
outputs
```
1
```
Aha, so there's some magic during parse that makes `class C` more than just an instance of `interface I` if `implments I` is present.

zzzyxwvut:
It is good to have this issue formally fleshed out.  I have  
already broached the subject of it [here](https://github.com/vim/vim/discussions/15463#discussioncomment-10300781):

> Then there is also a wrong-result success: comment out the
> stipulated by (3) `implements` clauses, BUT make A concrete,
> or keep `A` `abstract` and implement `Odd()`, then observe the
> uniform output of `TestOdd*()`.

Now it remains to be seen how this requirement for the  
repetitive use of `implements` can be addressed:
* Always repeat `implements` (enforce it during compilation of   
  derived `class`es by failing to compile any `class` that does  
  not repeat `implements`, and document it; because sooner or  
[...]

PR #15714, opened 2024-09-21
Update vim9compile.c

pr3m-kamble:
The patch essentially prevents the compiler from looking for a return statement in an abstract method, which is unnecessary.

### DIFF
--- a/src/vim9compile.c
+++ b/src/vim9compile.c
@@ -4121,7 +4121,12 @@ compile_def_function(
 
     // Compiling a function in an interface is done to get the function type.
     // No code is actually compiled.
-    if (ufunc->uf_class != NULL && IS_INTERFACE(ufunc->uf_class))
+    if ((ufunc->uf_class != NULL && IS_INTERFACE(ufunc->uf_class))
+   || IS_ABSTRACT_METHOD(ufunc))
+{
+   ufunc->uf_def_status = UF_NOT_COMPILED;
+   ret = OK;
+}
     {
 	ufunc->uf_def_status = UF_NOT_COMPILED;
 	ret = OK;


chrisbra:
thanks. Can you create a test for that please?

pr3m-kamble:
Yes I'm working on it

zzzyxwvut:
This has already been tried with #15441 (patch [`v2`](https://github.com/vim/vim/commit/4be245a767cd2460924c727cf7c343d30eba7c7e)).  Read my  
objection [there](https://github.com/vim/vim/pull/15441#issuecomment-2289768763).

It looks malformed, what's up with the dangling original  
block?

https://github.com/vim/vim/blob/84c3dd8887b8760d1b9578d435689ae12f8e88cc/src/vim9compile.c#L4122-L4135

Ping @errael.

errael:
Assuming the new code (misindented) and not the old code; this PR continues processing. That's a problem. This looks like a typo.

> prevents the compiler from looking for a return statement in an abstract method, which is unnecessary.

See #15512 and #15441. The PR, which was closed and never merged for no good reason, also has a test.

zzzyxwvut:
This PR has also a test but in #15715.  Compare [yours](https://github.com/vim/vim/commit/4be245a767cd2460924c727cf7c343d30eba7c7e) and  
[the author's](https://github.com/vim/vim/pull/15715/commits/dee12d100f098705e952fa7426e5d0984020e92a).

chrisbra:
> The PR, which was closed and never merged for no good reason, also has a test.

Since this is the second time this has come up, perhaps we should re-consider and have it included then? Seems to fit a need here.

zzzyxwvut:
Let me summarise [my previous findings](https://github.com/vim/vim/discussions/15463#discussioncomment-10300781).

During the #15441 PR round, two patches were offered ([`v1`](https://github.com/vim/vim/commit/f958a72abd6df51789efdac23cfc5558847785d7) and  
[`v2`](https://github.com/vim/vim/commit/4be245a767cd2460924c727cf7c343d30eba7c7e)), and either one works as advertised ***iff*** every derived  
`class` that implements an `interface`, either directly or via  
inheritance, explicitly has the `implements` clause.  Now, this  
requirement is not documented or enforced (see #15484).  
When this requirement is flouted (forgotten, unaware of --  
there is no documentation), with the original `v2` patch, you  
will run into a bogus (default?) not well-typed value and,  
after some poking, internal errors whenever `interface` types  
are used for access (e.g. method parameters).

The current submission follows `v2` but in some malformed
manner (see above).

If feasible, the bogus value should also be addressed; if  
not, please document the `implements` requirement.

(See the linked comment for examples and further details.)

zzzyxwvut:
To clarify it for anyone looking at this issue for the first  
time and wondering about what is meant by the necessity to  
use the `implements` clause: the problem is with the current  
tacit requirement for the ***repeated use*** of the `implements`  
clause in derived `class`es so that implemented `interface`  
types can be used with instances of the said `class`es for  
method invocation.  Failure to follow it has shown to lead  
to wrong method dispatch etc.

errael:
> ...Now, this requirement is not documented or enforced (see #15484).

It's unclear to me why this situations continues to be mischaracterized. It's not about a requirement it's about a bug; actually interacting bugs. Further, the workaround is only needed in a limited number of situations (it is not a requirement in general).

If the nightmare bug, see #15484, is fixed then #15441 could be merged without issue (I think we all agree with that). Note that #15484 has nothing to do with `abstract method declarations` which is the subject of this PR and #15441. But the nightmare bug is far reaching and interacts with fixes for #15512.

#15484 discusses the problem, and #15495 has some code that plays with the code related to the problem.

One solution to #15484, is to require supposedly redundant `implements` clauses, see https://github.com/vim/vim/issues/15484#issuecomment-2291419375. While undesirable, it at least produces correct code in the situations that tweak the bug. To get that, an error must be produced if there is not an `implements` for all interfaces in any superclass. The code in #15495 shows where such a check could be put. But note that generating an error probably breaks a bunch of code.

errael:
> To clarify it for anyone looking at this issue for the first time and wondering about what is meant by the necessity to use the `implements` clause: the problem is with the current tacit requirement for the _**repeated use**_ of the `implements` clause in derived `class`es so that implemented `interface` types can be used with instances of the said `class`es for method invocation. Failure to follow it has shown to lead to wrong method dispatch etc.

I posted my previous comment before I saw this post and the change to **tacit** requirement. 

The thing is, the `repeated use of implements` workaround for #15484 has nothing directly to do with fixing #15512.

zzzyxwvut:
True, but how is the not well-typed bogus value _introduced_  
with `v2` or this patch is any better than some value obtained  
through wrong method dispatch.  Right, always use `implements`.  
How is that not a requirement?  And, if so, should only we be  
privy to this workaround?

errael:
> True, but how is the not well-typed bogus value _introduced_ with `v2` or this patch is any better than some value obtained through wrong method dispatch. Right, always use `implements`. How is that not a requirement? And, if so, should only we be privy to this workaround?

I don't understand what this post means. You want a bug in the compiler why?

Look at the "how to reproduce" case for #15484. It produces a bogus value. Bad values have nothing to do with #15441.

zzzyxwvut:
> Bad values have nothing to do with #15441.

Oh, dear, we're going around in circles.

One may argue that the original examples (#15432) deal only  
with `abstract class`es and either patch solves that.  But `v2`  
introduces another, more subtle bug for `interface`s.  Pay  
attention to the results under `v9.1.0661 with v2` in the  
example comments below.

According to the result type of `TestI()`, `0` is `string`, but  
`type()` claims `number`.  Where did `B` go?

Do we want to let it loose and tell not about `implements`?

```
vim9script

interface I
    def string(): string
endinterface

abstract class A implements I
    abstract def string(): string
endclass

# class B extends A implements I
class B extends A
    def string(): string
        return 'B'
    enddef
endclass

def TestI(i: I): string
    return i.string()
enddef

echo type(TestI(B.new()))
echo TestI(B.new())

# v9.1.0661 with v1:
# E1091: Function is not compiled: string
#
# v9.1.0661 with v1 and "class B extends A implements I":
# "1" is echo'd.
# "B" is echo'd.
#
#
# v9.1.0661 with v2 [THIS IS THE BOGUS VALUE; NOT A STRING!]:
# "0" is echo'd (WRONG TYPE!).
# "0" is echo'd (DEFAULT VALUE!).
#
# v9.1.0661 with v2 and "class B extends A implements I":
# "1" is echo'd.
# "B" is echo'd.
#
#
# v9.1.0733 (no patches)
# E1027: Missing return statement
#
# v9.1.0733 (no patches) and "class B extends A implements I":
# "1" is echo'd.
# "B" is echo'd.
```

errael:
> > Bad values have nothing to do with #15441.
> 
> Oh, dear, we're going around in circles.
> 
> One may argue that the original examples (#15432) deal only with `abstract class`es and either patch solves that. But `v2` introduces another, more subtle bug for `interface`s. 

You seem to be assuming that fixing this bug introduces another bug. Do you have evidence of that? I'm talking debugger and/or logging. It may expose another bug.

> Pay attention to the results under `v9.1.0661 with v2` in the example comments below.
> 
> According to the result type of `TestI()`, `0` is `string`, but `type()` claims `number`. 

vim has a variety of data integrity problems related to the kinds of things you're seeing. #15442, #15208. What's your point.

> Where did `B` go?

There is a bug in dispatching (as you know). Without the line you've commented out
```
# class B extends A implements I
```
that bug is encountered. I have not carefully characterized that bug to fully understand the various ways it might manifest. There is no expectation for it to work given the current state, as you know. 

If you add `implements I` does that fix things?

A known bug, not fully characterized, is producing buggy behavior.

> Do we want to let it loose and tell not about `implements`?

I'm all in favor of alerting the problem.

zzzyxwvut:
The problem this patch attempts to address has been taken  
care of in #16469 (`v9.1.1034`).

PR #15715, opened 2024-09-21
Update test_vim9_class.vim

pr3m-kamble:
Fix test for abstract methods in Vim9 classes

- Added a new test `Test_abstract_method()` in `test_vim9_class.vim`.
- This test defines an abstract class `A` with an abstract method `String()`, and a subclass `B` that implements the method.
- Verifies that calling the method on an instance of `B` works as expected without raising `E1027: Missing return statement`.
- Ensures the abstract method is handled correctly during compilation.

### DIFF
--- a/src/testdir/test_vim9_class.vim
+++ b/src/testdir/test_vim9_class.vim
@@ -3,27 +3,6 @@
 source check.vim
 import './vim9.vim' as v9
 
-def Test_abstract_method()
-    # Define an abstract class with an abstract method
-    abstract class A
-        abstract def String(): string
-    endclass
-
-    # Define a subclass that implements the abstract method
-    class B extends A
-        def String(): string
-            return 'B'
-        enddef
-    endclass
-
-    # Test the subclass implementation
-    def F(o: A)
-        assert_equal('B', o.String())
-    enddef
-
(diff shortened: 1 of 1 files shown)


src/testdir/test_vim9_class.vim
```diff
@@ -3,6 +3,30 @@
 source check.vim
 import './vim9.vim' as v9
 
+def Test_abstract_method()
+    vim9script
+
```
zeertzjq: You don't need to use `vim9script` in a `:def` function

src/testdir/test_vim9_class.vim
```diff
@@ -3,6 +3,28 @@
 source check.vim
 import './vim9.vim' as v9
 
+def Test_abstract_method()
```
yegappan: There is already an existing Test_abstract_method() test function that tests these cases.  What new test is added in this function?  Also, this function should not be first function in this file.  Also this should be added to the existing Test_abstract_method() function.

chrisbra:
Hm, currently it only adds a single line to src/testdir/test_vim9_class.vim That doesn't look useful. Are you still working on that?

chrisbra:
closing for now.

PR #15719, opened 2024-09-22
Fix incorrect result of diff mode for empty file.

ynkdir:
When using internal diff, empty file is read as one empty line file. So result differ from external diff.

```
$ echo "0" > a.txt
$ touch b.txt
$ vim --clean -d a.txt b.txt
```

result (render as 1c1)

result

expected (render as 1d0)

expected


chrisbra:
thanks!

PR #15720, opened 2024-09-22
runtime(python): support multi-line case

### LINKED ISSUE #15631: Missing syntax highlight on multiline case xxx: statements in Python
### Steps to reproduce

The keyword "case" is not highlighted in this case (pun indented):

```py
                  match node.func:
                      case ast.Attribute(
                          value=ast.Attribute(value=ast.Name(id=first), attr=second),
                          attr=third,
                      ):
```

This, however, if OK:

```py
                  match node.func:
                      case ast.Attribute(value=ast.Attribute(value=ast.Name(id=first), attr=second), attr=third):
```

### Expected behaviour

keywords to be highlighted in both single line and multiline statements.

### Version of Vim

9.0 and 9.1

### Environment

macOS arm64, iTerm2

Tested with:
- VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Aug 28 2024 20:08:35)
- VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jun 29 2024 05:34:37)

### Logs and stack traces

_No response_

npip99:
Updates runtime(python) so that it can handle multi-line case statements.

Before:

Screenshot 2024-09-21 at 10 09 07 PM

After:

Screenshot 2024-09-21 at 10 10 19 PM

Useful for matching against Enums with long names, where you want one line per "or" condition. Also, [black](https://github.com/psf/black) will automatically split long case statements using parenthesis in this way.

Fixes #15631

### DIFF
--- a/runtime/syntax/python.vim
+++ b/runtime/syntax/python.vim
@@ -1,7 +1,7 @@
 " Vim syntax file
 " Language:	Python
 " Maintainer:	Zvezdan Petkovic <zpetkovic@acm.org>
-" Last Change:	2023 Feb 28
+" Last Change:	2024 Sep 21
 " Credits:	Neil Schemenauer <nas@python.ca>
 "		Dmitry Vasiliev
 "
@@ -108,7 +108,7 @@ syn keyword pythonAsync		async await
 " These keywords do not mean anything unless used in the right context.
 " See https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords
 " for more on this.
-syn match   pythonConditional   "^\s*\zscase\%(\s\+.*:.*$\)\@="
+syn match   pythonConditional   "^\s*\zscase\%(\s\+.*[:(].*$\)\@="
 syn match   pythonConditional   "^\s*\zsmatch\%(\s\+.*:\s*\%(#.*\)\=$\)\@="
 
 " Decorators


chrisbra:
Thanks. Does that fix #15631 ? Did you contact the maintainer? Ping @zvezdan

npip99:
@chrisbra Hello! Yes, this does fix https://github.com/vim/vim/issues/15631 , updated the PR body to auto-close the issue on merge.

I've emailed the maintainer, I'll let you know if/when I get an email back.

Thanks!

zvezdan:
This change fixes the simple case shown in the example, but unfortunately it seems that it will also gladly match this:
```python
case = (
    a +
    b +
    c
)
```
Being a **soft keyword** it is [allowed as an identifier](https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords) (probably in older code). _It should not be highlighted as a keyword in that case._
So, we need to look for a regex that confirms there's a `):` at the end of the overall multi-line expression, if it starts with `(\n`.
As usual, these use cases are far from trivial. 🙂
I'll try to find the right solution ...

Thank you for bringing this up @npip99!
I think that `match` can have the same problem if it's written like this:
```python
match (
    100,
    200
):
    case (100, y):
        # ...
```
So, I'll have to look for a _generic_ solution for _all soft keywords_.
There's also a new soft keyword `type` since Python 3.12 that I need to add.

chrisbra:
closing on behalf of maintainer.

npip99:
@zvezdan Thanks for the response! I would consider things like soft keywords to be the job of an LSP or Treesitter, since requires parsing the entire abstract syntax tree to handle these cases properly.

Given that regexes are famously [not powerful enough to match parenthesis](https://imada.sdu.dk/u/rolf/Edu/DM534/E19/beregningsModellerSlides.pdf#page=81), this problem is actually computationally impossible, because it would involve detecting the trailing `):`, meanwhile the content in-between the parenthesis is an arbitrarily nested python expression.

It seems like the correct decision is to just always mark it as a keyword syntax highlight regardless of the situation, and only adjust the highlight if there's proof from the LSP that it's actually being used as a variable.

The behavior that I just described above is actually exactly how VSCode works, i.e., it is purple until the LSP decides that it's a variable.
Screenshot 2024-11-20 at 10 32 41 PM
Screenshot 2024-11-20 at 10 34 13 PM

The current regex in this PR works with my LSP in vim in the same way as it works in VSCode.
Screenshot 2024-11-20 at 10 42 09 PM

---

We could open a new GH issue to fix the much rarer edge-case that this PR would introduce after merge, but as mentioned there is no real solution to this problem, so it wouldn't necessarily be wrong to close that new issue as "won't fix" (But the discussion can happen in the body of that new GH issue).

Also, if you want to reduce the probability of this edge-case occuring, we could use a stricter regex such as

```
autocmd FileType python syn match pythonConditional "^\s*\zscase\%(\s\+(\)\@="
```

Meaning that, it only works if it's `case<whitespace>(`, and therefore `case = (` won't be caught. The only (even rarer) edge-case, would be when "case" is a multiline function call. By doing \w+, it's even rarer, it would have to be a function call written with a whitespace in-between the "case" and the open parenthesis, which is a very-nonstandard way to write a function call. It's possible that this is so rare that such a false positive doesn't occur in all of github, but I'm not sure, it's probably safe enough to be worth improving the very common situation of a multiline case/match.

Issue #14872, opened 2024-05-29
Include common commands :(L)Make/Dispatch,Silent for async make, opening a terminal app


Konfekt:
The following common tasks are often reimplemented as they are lacking a default implementation in Vim.
If this observation is shared, maybe one could agree upon one such implementation:

-  an async version [`(L)Make`]( https://github.com/mg979/tasks.vim/blob/master/plugin/async.vim) of `:(l)make` (and at that occasion, including a `:Compiler` command to quickly switch the compiler might be convenient)
- a command, say [`:Silent(!)`]( https://gist.github.com/Konfekt/8e484af2955a0c7bfe82114df683ce0f) to spawn an external (GUI) executable in background, which is somewhat cumbersome under Windows, Git Bash and WSL (when the GUI app is a Win32 exe) whereas `:silent! ... | redraw!` mostly suffices under Unix
- a command, say [`:Open`]( https://gist.github.com/Konfekt/8e484af2955a0c7bfe82114df683ce0f) to open a URL or file, say to look up something in the browser or open a compiled (markup file), for example 
	- `:set keywordprg=:Open!\ https://devdocs.io/#q=`
	- `:Open! %:r.pdf` to open a PDF, HTML, ... file (say compiled with `:compiler pandoc`)

Konfekt:
On Neovim, an analog of `:Open` was implemented natively as [vim.ui.open()](https://github.com/neovim/neovim/commit/af6e6ccf3dee815850639ec5613dda3442caa7d6]

Konfekt:
> :Silent, if I'm not misunderstanding, might be better named something like :Run?

Yes, indeed, `:Silent!` now has become `:Launch` in https://github.com/vim/vim/pull/15721 leveraged by `gx` but I am fine with any similar name to launch/run a (GUI) app in background without clutter.

> Async :make is more of a slippery slope. I know there's several plugins that cover this territory but I'm not sure how similar they are in functionality.

Agreed, the implementations vary wildly and it seems more intricate. What made @mg979 's `:Make` stand out is its window-local version `:Lmake` that vim-dispatch didn't [dare to implement](https://github.com/tpope/vim-dispatch/issues/31).

Konfekt:
[...]

PR #15721, opened 2024-09-22
runtime:netrw: simplify gx file handling

Konfekt:
Take care of gx in Git Bash and WSL as maintenance stalled.

The whole logic was somewhat convoluted with some parts repeatedly invoking failed commands.

The file handling was outdated, for example, nowadays Netscape is rarely used, and also opinionated, for example mainly Microsoft Paint and Gimp for Image files.
Instead, let (xdg-)open/start handle this and respect the user's preferences.

Uses a public `:Launch/Open` command to launch applications / open URLs and files similar to Neovim's native [vim.ui.open()](https://github.com/neovim/neovim/commit/af6e6ccf3dee815850639ec5613dda3442caa7d6)

### DIFF
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -5432,7 +5432,8 @@ fun! netrw#BrowseX(fname,remote)
 "   call Decho("(netrw#BrowseX) g:netrw_browsex_viewer<".g:netrw_browsex_viewer.">",'~'.expand("<slnum>"))
     exe 'Launch' viewer viewopt shellescape(fname, 1)
   else
-    exe 'Open' fname
+     " though shellescape(..., 1) is used in Open, it's insufficient
+     exe 'Open' escape(fname, '#%')
   endif
 
   " cleanup: remove temporary file,
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -5430,9 +5430,9 @@ fun! netrw#BrowseX(fname,remote)
 
   if exists("g:netrw_browsex_viewer") && executable(viewer)
 "   call Decho("(netrw#BrowseX) g:netrw_browsex_viewer<".g:netrw_browsex_viewer.">",'~'.expand("<slnum>"))
-    exe 'Launch' viewer viewopt escape(fname, '#%')
+    exe 'Launch' viewer viewopt shellescape(fname, 1)
   else
-    exe 'Open' escape(fname, '#%')
+    exe 'Open' fname
   endif
 
   " cleanup: remove temporary file,
@@ -5490,11 +5490,11 @@ fun! s:GetURL() abort
   let file = expand("<cfile>")
   if filereadable(file) | return file | endif
   " ... or in that of the current buffer?
-  let path = expand('%')
+  let path = fnamemodify(expand('%'), ':p')
   if isdirectory(path)
-    let dir = fnamemodify(path, ':p')
+    let dir = path
   elseif filereadable(path)
-    let dir = fnamemodify(path, ':p:h')
+    let dir = fnamemodify(path, ':h')
   endif
   if exists('dir') && filereadable(dir..'/'..file) | return dir..'/'..file | endif
 
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -5478,6 +5478,7 @@ fun! netrw#GX()
 endfun
 
 fun! s:GetURL() abort
+   let URL = ''
    if exists('*Netrw_get_URL_' .. &filetype)
       let URL = call('Netrw_get_URL_' .. &filetype, [])
    endif
--- a/runtime/autoload/netrw.vim
+++ b/runtime/autoload/netrw.vim
@@ -27,6 +27,7 @@
 "   2024 Sep 19 by Vim Project: mf-selection highlight uses wrong pattern (#15700)
 "   2024 Sep 21 by Vim Project: remove extraneous closing bracket (#15718)
 "   2024 Oct 21 by Vim Project: remove netrwFileHandlers (#15895)
+"   2024 Oct 21 by Vim Project: clean up gx mapping
 "   }}}
 " Former Maintainer:	Charles E Campbell
 " GetLatestVimScripts: 1075 1 :AutoInstall: netrw.vim
@@ -538,7 +539,6 @@ if !exists("g:netrw_sort_sequence")
 endif
 call s:NetrwInit("g:netrw_special_syntax"   , 0)
 call s:NetrwInit("g:netrw_ssh_browse_reject", '^total\s\+\d\+$')
-call s:NetrwInit("g:netrw_suppress_gx_mesg",  1)
 call s:NetrwInit("g:netrw_use_noswf"        , 1)
 call s:NetrwInit("g:netrw_sizestyle"        ,"b")
 " Default values - t-w ---------- {{{3
@@ -5406,25 +5406,6 @@ fun! netrw#BrowseX(fname,remote)
 "  call Decho("fname<".fname.">",'~'.expand("<slnum>"))
 "  call Decho("exten<".exten."> "."netrwFileHandlers#NFH_".exten."():exists=".exists("*netrwFileHandlers#NFH_".exten),'~'.expand("<slnum>"))
 
-  " set up redirection (avoids browser messages)
-  " by default, g:netrw_suppress_gx_mesg is true
-  if g:netrw_suppress_gx_mesg
-   if &srr =~ "%s"
-    if has("win32")
-     let redir= substitute(&srr,"%s","nul","")
-    else
-     let redir= substitute(&srr,"%s","/dev/null","")
-    endif
-   elseif has("win32")
-    let redir= &srr . "nul"
-   else
-    let redir= &srr . "/dev/null"
-   endif
-  else
-   let redir= ""
-  endif
-"  call Decho("set up redirection: redir{".redir."} srr{".&srr."}",'~'.expand("<slnum>"))
-
   " extract any viewing options.  Assumes that they're set apart by spaces.
   if exists("g:netrw_browsex_viewer")
 "   call Decho("extract any viewing options from g:netrw_browsex_viewer<".g:netrw_browsex_viewer.">",'~'.expand("<slnum>"))
@@ -5447,85 +5428,12 @@ fun! netrw#BrowseX(fname,remote)
 "   call Decho("viewer<".viewer.">  viewopt<".viewopt.">",'~'.expand("<slnum>"))
   endif
 
-  " execute the file handler
-"  call Decho("execute the file handler (if any)",'~'.expand("<slnum>"))
   if exists("g:netrw_browsex_viewer") && executable(viewer)
 "   call Decho("(netrw#BrowseX) g:netrw_browsex_viewer<".g:netrw_browsex_viewer.">",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !".viewer." ".viewopt.s:ShellEscape(fname,1).redir)
-   let ret= v:shell_error
-
-  elseif has("win32")
-"   call Decho("(netrw#BrowseX) win".(has("win32")? "32" : "64"),'~'.expand("<slnum>"))
-   if executable("start")
-    call s:NetrwExe('sil! !start rundll32 url.dll,FileProtocolHandler '.s:ShellEscape(fname,1))
-   elseif executable("rundll32")
-    call s:NetrwExe('sil! !rundll32 url.dll,FileProtocolHandler '.s:ShellEscape(fname,1))
-   else
-    call netrw#ErrorMsg(s:WARNING,"rundll32 not on path",74)
-   endif
-   let ret= v:shell_error
-
-  elseif has("win32unix")
-   let winfname= 'c:\cygwin'.substitute(fname,'/','\\','g')
-"   call Decho("(netrw#BrowseX) cygwin: winfname<".s:ShellEscape(winfname,1).">",'~'.expand("<slnum>"))
-   if executable("start")
-"    call Decho("(netrw#BrowseX) win32unix+start",'~'.expand("<slnum>"))
-    call s:NetrwExe('sil !start rundll32 url.dll,FileProtocolHandler '.s:ShellEscape(winfname,1))
-   elseif executable("rundll32")
-"    call Decho("(netrw#BrowseX) win32unix+rundll32",'~'.expand("<slnum>"))
-    call s:NetrwExe('sil !rundll32 url.dll,FileProtocolHandler '.s:ShellEscape(winfname,1))
-   elseif executable("cygstart")
-"    call Decho("(netrw#BrowseX) win32unix+cygstart",'~'.expand("<slnum>"))
-    call s:NetrwExe('sil !cygstart '.s:ShellEscape(fname,1))
-   else
-    call netrw#ErrorMsg(s:WARNING,"rundll32 not on path",74)
-   endif
-   let ret= v:shell_error
-
-  elseif has("unix") && $DESKTOP_SESSION == "mate" && executable("atril")
-"   call Decho("(netrw#BrowseX) unix and atril",'~'.expand("<slnum>"))
-   if a:fname =~ '^https\=://'
-    " atril does not appear to understand how to handle html -- so use gvim to edit the document
-    let use_ctrlo= 0
-"    call Decho("fname<".fname.">")
-"    call Decho("a:fname<".a:fname.">")
-    call s:NetrwExe("sil! !gvim ".fname.' -c "keepj keepalt file '.fnameescape(a:fname).'"')
-
-   else
-    call s:NetrwExe("sil !atril ".s:ShellEscape(fname,1).redir)
-   endif
-   let ret= v:shell_error
-
-  elseif has("unix") && executable("kfmclient") && s:CheckIfKde()
-"   call Decho("(netrw#BrowseX) unix and kfmclient",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !kfmclient exec ".s:ShellEscape(fname,1)." ".redir)
-   let ret= v:shell_error
-
-  elseif has("unix") && executable("exo-open") && executable("xdg-open") && executable("setsid")
-"   call Decho("(netrw#BrowseX) unix, exo-open, xdg-open",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !setsid xdg-open ".s:ShellEscape(fname,1).redir.'&')
-   let ret= v:shell_error
-
-  elseif has("unix") && executable("xdg-open")
-"   call Decho("(netrw#BrowseX) unix and xdg-open",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !xdg-open ".s:ShellEscape(fname,1).redir.'&')
-   let ret= v:shell_error
-
-  elseif has("macunix") && executable("open")
-"   call Decho("(netrw#BrowseX) macunix and open",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !open ".s:ShellEscape(fname,1)." ".redir)
-   let ret= v:shell_error
+    exe 'Launch' viewer viewopt escape(fname, '#%')
   else
-   call netrw#ErrorMsg(s:ERROR, "Couldn't find a program to open '".a:fname."'", 0)
-   let ret=0
+    exe 'Open' escape(fname, '#%')
   endif
-
-  if ret
-   call netrw#ErrorMsg(s:ERROR, "Failed to open '".a:fname."'", 0)
-  endif
-
-  " restoring redraw! after external file handlers
-  redraw!
 
   " cleanup: remove temporary file,
   "          delete current buffer if success with handler,
@@ -5563,11 +5471,35 @@ fun! netrw#GX()
   if &ft == "netrw"
    let fname= s:NetrwGetWord()
   else
-   let fname= expand((exists("g:netrw_gx")? g:netrw_gx : '<cfile>'))
+   let fname= exists("g:netrw_gx")? expand(g:netrw_gx) : s:GetURL()
   endif
 "  call Dret("netrw#GX <".fname.">")
   return fname
 endfun
+
+fun! s:GetURL() abort
+   if exists('*Netrw_get_URL_' .. &filetype)
+      let URL = call('Netrw_get_URL_' .. &filetype, [])
+   endif
+   if !empty(URL) | return URL | endif
+  " URLs end in letter, digit or forward slash
+  let URL = matchstr(expand("<cWORD>"), '\<' .. g:netrw_regex_url .. '\ze[^A-Za-z0-9/]*$')
+  if !empty(URL) | return URL | endif
+
+  " Is it a file in the current work dir ...
+  let file = expand("<cfile>")
+  if filereadable(file) | return file | endif
+  " ... or in that of the current buffer?
+  let path = expand('%')
+  if isdirectory(path)
+    let dir = fnamemodify(path, ':p')
+  elseif filereadable(path)
+    let dir = fnamemodify(path, ':p:h')
+  endif
+  if exists('dir') && filereadable(dir..'/'..file) | return dir..'/'..file | endif
+
+  return ''
+endf
 
 " ---------------------------------------------------------------------
 " netrw#BrowseXVis: used by gx in visual mode to select a file for browsing {{{2
@@ -6733,6 +6665,7 @@ fun! s:NetrwMaps(islocal)
    nnoremap <buffer> <silent> <nowait> U	:<c-u>call <SID>NetrwBookHistHandler(5,b:netrw_curdir)<cr>
    nnoremap <buffer> <silent> <nowait> v	:call <SID>NetrwSplit(2)<cr>
    nnoremap <buffer> <silent> <nowait> x	:<c-u>call netrw#BrowseX(<SID>NetrwBrowseChgDir(0,<SID>NetrwGetWord()),1)<cr>
+   nmap     <buffer>          <nowait> gx	x
    if !hasmapto('<Plug>NetrwHideEdit')
     nmap <buffer> <c-h> <Plug>NetrwHideEdit
    endif
@@ -12156,13 +12089,16 @@ endfun
 " s:NetrwExe: executes a string using "!" {{{2
 fun! s:NetrwExe(cmd)
 "  call Dfunc("s:NetrwExe(a:cmd<".a:cmd.">)")
-  if has("win32") && &shell !~? 'cmd\|pwsh\|powershell' && !g:netrw_cygwin
+  if has("win32")
 "    call Decho("using win32:",expand("<slnum>"))
     let savedShell=[&shell,&shellcmdflag,&shellxquote,&shellxescape,&shellquote,&shellpipe,&shellredir,&shellslash]
     set shell& shellcmdflag& shellxquote& shellxescape&
     set shellquote& shellpipe& shellredir& shellslash&
-    exe a:cmd
-    let [&shell,&shellcmdflag,&shellxquote,&shellxescape,&shellquote,&shellpipe,&shellredir,&shellslash] = savedShell
+    try
+     exe a:cmd
+    finally
+      let [&shell,&shellcmdflag,&shellxquote,&shellxescape,&shellquote,&shellpipe,&shellredir,&shellslash] = savedShell
+    endtry
   else
 "   call Decho("exe ".a:cmd,'~'.expand("<slnum>"))
    exe a:cmd
--- a/runtime/doc/pi_netrw.txt
+++ b/runtime/doc/pi_netrw.txt
@@ -1529,7 +1529,7 @@ Other than a file path, the text under the cursor may be a URL.  Netrw uses
 by default the following regular expression to determine if the text under the
 cursor is a URL:
 >
-        g:netrw_regex_url = '\%(\%(http\|ftp\|irc\)s\?\|file\)://\S\{-}'
+	g:netrw_regex_url = '\%(\%(http\|ftp\|irc\)s\?\|file\)://\S\{-}'
 <
 
 Netrw determines which special handler by the following method:
@@ -1541,7 +1541,7 @@ Netrw determines which special handler by the following method:
   * otherwise:
 
     * for Windows                     : explorer.exe is used
-  * for Mac OS X               : open is used.
+    * for Mac OS X               : open is used.
     * for Linux               : xdg-open is used.
 
 To launch a specific application <app>, type :Launch <app>.
@@ -2573,7 +2573,7 @@ your browsing preferences.  (see also: |netrw-settings|)
   *g:netrw_browsex_viewer*	specify user's preference for a viewer: >
 					"kfmclient exec"
 					"gnome-open"
-
+<
   *g:netrw_browsex_support_remote*
 				specify if the specified viewer supports a
 				remote URL.  (see |netrw-handler|).
--- a/runtime/doc/pi_netrw.txt
+++ b/runtime/doc/pi_netrw.txt
@@ -8,7 +8,7 @@ Author:  Charles E. Campbell  <NcampObell@SdrPchip.AorgM-NOSPAM>
 
 Copyright: Copyright (C) 2017 Charles E Campbell    *netrw-copyright*
 	The VIM LICENSE applies to the files in this package, including
-	netrw.vim, pi_netrw.txt, netrwFileHandlers.vim, netrwSettings.vim, and
+        netrw.vim, pi_netrw.txt, netrwSettings.vim, and
 	syntax/netrw.vim.  Like anything else that's free, netrw.vim and its
 	associated files are provided *as is* and comes with no warranty of
 	any kind, either expressed or implied.  No guarantees of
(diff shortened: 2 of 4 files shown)


Konfekt:
The file `runtime/autoload/netrw.vim` takes a while to load; could it not better be split up to make it more mangageable?

chrisbra:
oh, you cleaned up a lot. Can a few people try this out please?

runtime/autoload/netrw.vim
```diff
@@ -5570,12 +5469,52 @@ fun! netrw#GX()
...
    let fname= s:NetrwGetWord()
   else
-   let fname= expand((exists("g:netrw_gx")? g:netrw_gx : '<cfile>'))
+   let fname= exists("g:netrw_gx")? expand(g:netrw_gx) : s:GetURL()
   endif
 "  call Dret("netrw#GX <".fname.">")
   return fname
 endfun
 
+fun! s:GetURL() abort
```
Konfekt, line 5478: This part adapts https://github.com/habamax/.vim/blob/d5087779fee4a7372c1dd2c407674997fdb1268d/autoload/os.vim

runtime/autoload/netrw.vim
```diff
@@ -5570,12 +5469,52 @@ fun! netrw#GX()
...
+  if isdirectory(path)
+    let dir = fnamemodify(path, ':p')
+  elseif filereadable(path)
+    let dir = fnamemodify(path, ':p:h')
+  endif
+  if exists('dir') && filereadable(dir..'/'..file) | return dir..'/'..file | endif
+
+  return ''
+endf
+
```
saccarosium: I don't think it is a good idea to support specifically markdown and html urls. I think keeping the current behavior is better (only working if the cursor is on an actuall url)

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,67 @@ set cpo&vim
...
+  endif
+else
+  let s:redir= ""
+endif
+"  call Decho("set up redirection: redir{".redir."} srr{".&srr."}",'~'.expand("<slnum>"))
+
+if has('unix')
+  if has('win32unix') " Git Bash provides /usr/bin/start script calling cmd.exe //c
+    " use start //b "" to set void title and avoid ambiguity with passed argument
+    silent! command -complete=shellcmd -nargs=1 -bang Launch
```
saccarosium: I don't think this is a good user command. It would be probably be better as a function in `autoload/dist`. I don't know if the users wants to launch a program directly.

saccarosium (commented):
I've tested this branch and I found that it tries to open the program directly in the CWD.

Step to repro:
1. `sudo make install`
2. `vim`
3. `:Launch <program>`

I get the following error:
```
nohup: failed to run command '/home/sacca/Repos/github.com/vim/vim/xdg-open /home/sacca/Repos/github.com/vim/vim/README.rux.txt': No such file or directory
```

runtime/plugin/netrwPlugin.vim
Konfekt: See https://github.com/vim/vim/issues/14872 for examples; to read any output of input in Vim, say binary, or even text, like a web/html page. One could also imagine a background process like browser-sync for web developers, or more generally a file watcher

runtime/autoload/netrw.vim
Konfekt: Yes, a matter of taste. I deemed markdown links common enough, but these can be deferred to a suitable ftplugin

Konfekt: Mind you that you can already override the selection to be opened, and this option is preserved

Konfekt:
Thank you @saccarosium, I hopefully have fixed the command by the latest commit

runtime/plugin/netrwPlugin.vim
saccarosium: Yes, I do think that this functionality is useful. But I find difficult imagine a situation where a user wants the command handy to launch an app. I think this is more useful as a function when you are scripting.

runtime/autoload/netrw.vim
saccarosium: > Mind you that you can already override the selection to be opened, and this option is preserved

What do you mean?

runtime/plugin/netrwPlugin.vim
Konfekt: I think this is current practice. At the same time, there are so many plug-ins whose principal raison d'etre is launching some command, say https://github.com/romainl/vim-devdocs, https://github.com/wolandark/vim-live-server/, various markdown live preview [plug-ins](https://github.com/arminveres/md-pdf.nvim/issues/19) vs. launching [Okular](https://gist.github.com/Konfekt/c2066abc31b41430417310bd6e6d795e). Similar to using `:compiler` instead of installing linter or compiler plug-ins, one could opt for `:Launch` instead of installing application specific plug-ins.

runtime/autoload/netrw.vim
Konfekt: I meant `:help g:netrw_gx` though I find it rather coarse and in general maybe some options are no longer that useful. I dared to remove `g:netrw_browsex_viewer` as it relied by default on ancient applications such as Netscape and seemed overengineered for its user base (that I suspected to be small since by now somebody would have replaced Netscape Navigator by something more modern).

saccarosium:
@Konfekt the problem persist. Now it interprets the entire string as a command:

```
nohup: redirecting stderr to stdout
nohup: failed to run command 'xdg-open /home/sacca/Repos/github.com/vim/vim/src/ascii.h': No such file or directory
```

Konfekt:
I beg your pardon, there was too much escaping between Vim command-line and shell. Could you try again?

saccarosium:
@Konfekt I've tried and doesn't function with filenames containing spaces. 

I really like what your are trying to do but at this stage there are too many sharp edges. I suggest to close this, cook the changes a little more and than do another PR.

I think would be beneficial if you split this in two PRs:
1. Refactor the netrw filehandling
2. Remap the `gx` key in other buffers

Also I would suggest to do something in the lines of:
- Define a generic function to execute a program (in the lines of what you are doing with `Start`) in `autoload/dist`, preferably in vimscript9

```vim
# Execute a program and forks it
def ExecProgram(cmd: list<string>)
```

- Define command `Launch` and `Open` that call that function. Mainly because escaping directly in command as you are doing is a bit weird to predict and manage

Konfekt:
> I've tried and doesn't function with filenames containing spaces.

But this is intentional for most flexibility. Your command after `:Launch` can contain spaces, and then you can pass more than one file name to it, usually separated by spaces. Of course expectations might differ, but rather than restricting to a single scenario at this point (I suppose your preference being a single command with no spaces followed by a single argument possibly including spaces?) due to its wide applicability I'd prefer to keep it flexible.

Another question is how to go about Vim's special command-line characters like `%` and `#`; again I opted not to escape them to allow a user to  `:Open %:r.pdf` if she happens to have compiled a text file to a PDF, say.

If you think that `:Open` should be able to paste any file name as is, then we can agree on shell escaping and escaping special characters, but since this command may have many uses, I've opted to keep it flexible. 

This escaping can then (optionally) be done by, say mappings, calling `:Open` like `gx`.

Konfekt:
> I suppose your preference being a single command with no spaces followed by a single argument possibly including spaces?

So at the moment, say `:Launch zathura "path with spaces.pdf"` works; I suppose you prefer something like `:Launch zathura path with spaces.pdf` ? Problem would be that it breaks the first variant and makes passing more than one (file) parameter impossible.

It does work for `:Open path with spaces.pdf` though, as I just allowed it to have a single argument. Again, Vim's special characters are intentionally not escaped to allow for the above `:Open %:...` scenario

saccarosium:
> But this is intentional for most flexibility. Your command after :Launch can contain spaces, and then you can pass more than one file name to it, usually separated by spaces. Of course expectations might differ, but rather than restricting to a single scenario at this point (I suppose your preference being a single command with no spaces followed by a single argument possibly including spaces?) due to its wide applicability I'd prefer to keep it flexible.

If I press `x`, in the netrw buffer, on a file with spaces it will not work. That's what I'm saying.

I think a PR is not the right place to discuss a possible API for some user commands. Also this PR is too big is hard to follow or review. I think this should be closed and split in more manageable changes.

You need to keep in mind that if we expose user facing functionality it should be thought through more carefully. We don't really remove features, since we are really committed to backwards compatability and stability, so if you are figuring stuff as you go on the Pull Request is not really promessing for something good. 

General Tips:
- PRs reffering to runtime files should be prefixed with `runtime(<file>): <msg>`
- You probably need to parse in some way the `<q-args>` to get a reliable result with multiple files.

Konfekt:
> If I press x, in the netrw buffer, on a file with spaces it will not work. That's what I'm saying.

Can't reproduce on Linux. Would you mind providing an example?

Mind you that x in Netrw buffers is a differen mapping from the global mapping `gx` :

```vim
" runtime/autoload/netrw.vim (line 6665)
nnoremap <buffer> <silent> <nowait> x	:<c-u>call netrw#BrowseX(<SID>NetrwBrowseChgDir(0,<SID>NetrwGetWord()),1)<cr>
```

Therefore interesting would be the value of `<SID>NetrwBrowseChgDir(0,<SID>NetrwGetWord())` which this PR does not touch. If that's wrong, that would be a separate issue.

Konfekt:
> You probably need to parse in some way the <q-args> to get a reliable result with multiple files.

Please be more concrete. As I tried to explain above, I'd like to leave proper shell escaping up to the user for `:Launch` as these commands may be arbitrary complex; 
while for the more user facing `:Open` it is understood that a single file path is allowed; therefore proper shell escaping can be done and is for sake of user convenience not required.

saccarosium:
> Can't reproduce on Linux. Would you mind providing an example?

**Master**

[Screencast From 2024-10-14 22-24-49.webm](https://github.com/user-attachments/assets/3ef2a9cc-3278-4094-b580-b91a6c4fbd68)

**PR**

[Screencast From 2024-10-14 22-26-38.webm](https://github.com/user-attachments/assets/c50f86b6-53f3-4745-a33a-447d38571276)

Konfekt:
I can't reproduce with the current repo (on Linux). I created a PDF named `introduction to algorithms (Thomas H. Cormen, Charles E. Leiserson etc.).pdf` and it opened fine by `:Open introduction to algorithms (Thomas H. Cormen, Charles E. Leiserson etc.).pdf`. Could you check if `:Open introduction to algorithms (Thomas H. Cormen, Charles E. Leiserson etc.).pdf` works?

Konfekt:
Thank you for your patience!

Konfekt:
Looking at the last screenshot, I wonder why there's an initial backtick ``` with a final single quote `'`

runtime/doc/pi_netrw.txt
```diff
@@ -1,29 +1,29 @@
 *pi_netrw.txt*  For Vim version 9.1.  Last change: 2024 Jul 13
 
-	    ------------------------------------------------
-	    NETRW REFERENCE MANUAL    by Charles E. Campbell
-	    ------------------------------------------------
+            ------------------------------------------------
+            NETRW REFERENCE MANUAL    by Charles E. Campbell
+            ------------------------------------------------
```
k-takata: Did you intentionally expand the tabs?

runtime/doc/pi_netrw.txt
```diff
@@ -1469,119 +1469,79 @@ With either form of the command, netrw will first ask for confirmation
...
 Certain files, such as html, gif, jpeg, (word/office) doc, etc, files, are
 best seen with a special handler (ie. a tool provided with your computer's
-operating system).  Netrw allows one to invoke such special handlers by: >
+operating system).  Netrw allows one to invoke such special handlers by:
 
-	* when Exploring, hit the "x" key
-	* when editing, hit gx with the cursor atop the special filename
-<	  (latter not available if the |g:netrw_nogx| variable exists)
+        * hitting gx with the cursor atop the file path or alternatively x
+          in in a netrw buffer; the former can be disabled by defining the
```
k-takata:
```suggestion
          in a netrw buffer; the former can be disabled by defining the
```

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,100 @@ set cpo&vim
 " ---------------------------------------------------------------------
 " Public Interface: {{{1
 
+" Commands Launch/URL {{{2
+" surpress output of command; use bang for GUI applications
+
+" set up redirection (avoids browser messages)
+" by default, g:netrw_suppress_gx_mesg is true
+if get (g:, ':netrw_suppress_gx_mesg', 1)
```
k-takata:
```suggestion
if get(g:, ':netrw_suppress_gx_mesg', 1)
```

runtime/autoload/netrw.vim
```diff
@@ -24,8 +24,8 @@
 "   2024 Aug 22 by Vim Project: fix mf-selection highlight (#15551)
 "   2024 Aug 22 by Vim Project: adjust echo output of mx command (#15550)
 "   2024 Sep 15 by Vim Project: more strict confirmation dialog (#15680)
-"   2024 Sep 19 by Vim Project: mf-selection highlight uses wrong pattern (#15700)
```
k-takata, line 27: Mistake? (Or, needs to be rebased?)

Konfekt: Somehow got lost in between, readded. Thank you for your input!

runtime/doc/pi_netrw.txt
Konfekt: No, I let git diff suppress them and sometimes stumble over it. Today is the day

k-takata: The tabs in all the other parts have also been expanded. So, this file has conflicts.

chrisbra:
Now that #15895 has been merged, can you please rebase and squash?

runtime/plugin/netrwPlugin.vim
chrisbra: Why the `:silent!`?

Konfekt: Thank you. Since there's the `g:loaded_netrwPlugin` load guard, I'd say it's dispensable and dropped it.

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,100 @@ set cpo&vim
...
+" surpress output of command; use bang for GUI applications
+
+" set up redirection (avoids browser messages)
+" by default, g:netrw_suppress_gx_mesg is true
+if get(g:, ':netrw_suppress_gx_mesg', 1)
+  if &srr =~# "%s"
+    if has("win32")
+      let s:redir= substitute(&srr,"%s","nul","")
+    else
+      let s:redir= substitute(&srr,"%s","/dev/null","")
```
saccarosium: Here you can use `printf` like so:
```vim
printf(&srr, null)
```

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,100 @@ set cpo&vim
...
+    let s:cmd = 'open'
+endif
+silent! command -complete=file -nargs=1 Open exe 'Launch' s:cmd <q-args>
+" endif
+
+if !exists('g:netrw_regex_url')
+  let g:netrw_regex_url = '\%(\%(http\|ftp\|irc\)s\?\|file\)://\S\{-}'
+endif
+
+silent! function Netrw_get_URL_markdown()
```
saccarosium: Can this be removed? I think this should be remain general. The user can always just go over the url.

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,100 @@ set cpo&vim
...
+    if searchpair('\[.\{-}\](', '', ')\zs', 'cbW', '', line('.')) > 0
+      return matchstr(getline('.')[col('.')-1:], '\[.\{-}\](\zs' .. g:netrw_regex_url .. '\ze\(\s\+.\{-}\)\?)')
+    endif
+  finally
+    call winrestview(save_view)
+    return ''
+  endtry
+endfunction
+
+silent! function Netrw_get_URL_html()
```
saccarosium: Can this be removed? I think this should be remain general. The user can always just go over the url.

saccarosium (commented):
I think this can be simplified a little bit in the coming days I will try to give more feedback. Thank you for your patience.

runtime/plugin/netrwPlugin.vim
Konfekt: I made them opt-in

Konfekt: Going over the URL is not so easy if concealing is used, for example.

runtime/doc/pi_netrw.txt
```diff
@@ -1469,42 +1469,82 @@ With either form of the command, netrw will first ask for confirmation
...
     If the viewer you wish to use does not support handling of a remote URL
     directory, set |g:netrw_browsex_support_remote| to 0.
-  * for Windows 32 or 64, the URL and FileProtocolHandler dlls are used.
-  * for Gnome (with gnome-open): gnome-open is used.
-  * for KDE (with kfmclient)   : kfmclient is used
+  * otherwise:
+
+    * for Windows                     : explorer.exe is used
   * for Mac OS X               : open is used.
+    * for Linux               : xdg-open is used.
```
k-takata: not well aligned.

runtime/doc/pi_netrw.txt
```diff
@@ -2533,7 +2573,7 @@ your browsing preferences.  (see also: |netrw-settings|)
   *g:netrw_browsex_viewer*	specify user's preference for a viewer: >
 					"kfmclient exec"
 					"gnome-open"
-<
+
```
k-takata: It seems `<` is needed.

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,72 @@ set cpo&vim
...
+" surpress output of command; use bang for GUI applications
+
+" set up redirection (avoids browser messages)
+" by default, g:netrw_suppress_gx_mesg is true
+if get(g:, ':netrw_suppress_gx_mesg', 1)
+  if &srr =~# "%s"
+    if has("win32")
+      let s:redir= printf(&srr, null)
+    else
+      let s:redir= printf(&srr, null)
```
k-takata:
```suggestion
      let s:redir= printf(&srr, "nul")
    else
      let s:redir= printf(&srr, "/dev/null")
```

Konfekt: Dang, thank you!

runtime/doc/pi_netrw.txt
Konfekt: True, accidentally removed as help files conceal by default these

runtime/doc/pi_netrw.txt
Konfekt: Thank you, corrected

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,72 @@ set cpo&vim
...
 
+" Commands Launch/URL {{{2
+" surpress output of command; use bang for GUI applications
+
+" set up redirection (avoids browser messages)
+" by default, g:netrw_suppress_gx_mesg is true
+if get(g:, ':netrw_suppress_gx_mesg', 1)
+  if &srr =~# "%s"
+    if has("win32")
+      let s:redir= printf(&srr, "null")
```
k-takata:
```suggestion
      let s:redir= printf(&srr, "nul")
```

runtime/autoload/netrw.vim
```diff
@@ -5563,12 +5471,36 @@ fun! netrw#GX()
...
+   if !empty(URL) | return URL | endif
+  " URLs end in letter, digit or forward slash
+  let URL = matchstr(expand("<cWORD>"), '\<' .. g:netrw_regex_url .. '\ze[^A-Za-z0-9/]*$')
+  if !empty(URL) | return URL | endif
+
+  " Is it a file in the current work dir ...
+  let file = expand("<cfile>")
+  if filereadable(file) | return file | endif
+  " ... or in that of the current buffer?
+  let path = expand('%')
```
chrisbra:
```suggestion
  let path = fnamemodify(expand('%'), ':p')
```

chrisbra: and then later on: `let dir = path`

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,72 @@ set cpo&vim
...
 
+" Commands Launch/URL {{{2
+" surpress output of command; use bang for GUI applications
+
+" set up redirection (avoids browser messages)
+" by default, g:netrw_suppress_gx_mesg is true
+if get(g:, ':netrw_suppress_gx_mesg', 1)
+  if &srr =~# "%s"
+    if has("win32")
+      let s:redir= printf(&srr, "nul")
```
chrisbra:
```suggestion
      let s:redir= printf(&srr, has("win32") ? "nul" : "/dev/null")
```

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,72 @@ set cpo&vim
...
+    if has("win32")
+      let s:redir= &srr . "nul"
+    else
+      let s:redir= &srr . "/dev/null"
+    endif
+  endif
+else
+  let s:redir= ""
+endif
+"  call Decho("set up redirection: redir{".redir."} srr{".&srr."}",'~'.expand("<slnum>"))
```
chrisbra: remove this commented line please

runtime/autoload/netrw.vim
```diff
@@ -5563,12 +5471,36 @@ fun! netrw#GX()
...
   return fname
 endfun
 
+fun! s:GetURL() abort
+   if exists('*Netrw_get_URL_' .. &filetype)
+      let URL = call('Netrw_get_URL_' .. &filetype, [])
+   endif
+   if !empty(URL) | return URL | endif
+  " URLs end in letter, digit or forward slash
+  let URL = matchstr(expand("<cWORD>"), '\<' .. g:netrw_regex_url .. '\ze[^A-Za-z0-9/]*$')
```
chrisbra, line 5488: `g:netrw_regex_url` uses `\S`, so that `\ze[^A-Za-z0-9/]*` seems redundant?

runtime/autoload/netrw.vim
```diff
@@ -5447,86 +5428,13 @@ fun! netrw#BrowseX(fname,remote)
...
-  elseif has("unix") && executable("xdg-open")
-"   call Decho("(netrw#BrowseX) unix and xdg-open",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !xdg-open ".s:ShellEscape(fname,1).redir.'&')
-   let ret= v:shell_error
-
-  elseif has("macunix") && executable("open")
-"   call Decho("(netrw#BrowseX) macunix and open",'~'.expand("<slnum>"))
-   call s:NetrwExe("sil !open ".s:ShellEscape(fname,1)." ".redir)
-   let ret= v:shell_error
+    exe 'Launch' viewer viewopt escape(fname, '#%')
```
chrisbra:
```suggestion
    exe 'Launch' viewer viewopt shellescape(fname,1)
```

runtime/autoload/netrw.vim
```diff
@@ -5447,86 +5428,13 @@ fun! netrw#BrowseX(fname,remote)
...
   else
-   call netrw#ErrorMsg(s:ERROR, "Couldn't find a program to open '".a:fname."'", 0)
-   let ret=0
-  endif
-
-  if ret
-   call netrw#ErrorMsg(s:ERROR, "Failed to open '".a:fname."'", 0)
+    exe 'Open' escape(fname, '#%')
```
chrisbra: same here

runtime/autoload/netrw.vim
Konfekt: Seemingly, though uses non-greedy `{-}` so that we assume that a URL ends in an alphanumeric letter or forward slash.

runtime/autoload/netrw.vim
Konfekt: For user's convenience, I opted to shellescape directly in `:Open` so that  command-line completed paths open

runtime/autoload/netrw.vim
Konfekt: This keeps it working on URLs with trailing punctuation, parentheses, ...

runtime/autoload/netrw.vim
mgedmin: Why `fnamemodify(expand('%'), ':p')` instead of just `expand('%:p')`?

Konfekt: Good point, seems the same

runtime/autoload/netrw.vim
chrisbra: So shouldn't be anything up to the first white-space then considered an URL? so something like this: `[^ ]*` ?

runtime/autoload/netrw.vim
chrisbra: I prefer the former over the latter, because the intention is more clear and therefore not so obscure

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,63 @@ set cpo&vim
...
+  endif
+else
+  let s:redir= ""
+endif
+
+if has('unix')
+  if has('win32unix') " Git Bash provides /usr/bin/start script calling cmd.exe //c
+    " use start //b "" to set void title and avoid ambiguity with passed argument
+    command -complete=shellcmd -nargs=1 -bang Launch
+          \ exe 'silent ! start "" //b ' . trim(<q-args>)  s:redir | redraw!
```
chrisbra: is this `//b` correct?

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,63 @@ set cpo&vim
...
+  elseif exists('$WSL_DISTRO_NAME') " use cmd.exe to start GUI apps in WSL
+    command -complete=shellcmd -nargs=1 -bang Launch execute ':silent !'..
+          \ ((<q-args> =~? '\v<\f+\.(exe|com|bat|cmd)>') ?
+            \ 'cmd.exe /c start "" /b ' . trim(<q-args>) :
+            \ 'nohup ' trim(<q-args>) s:redir '&')
```
chrisbra, line 73: this uses a weird mix of `..`, `.` or no concatenation operators

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,63 @@ set cpo&vim
...
+          \ | redraw!
+  else
+    command -complete=shellcmd -nargs=1 -bang Launch execute ':silent ! nohup' trim(<q-args>) s:redir '&' | redraw!
+  endif
+elseif has('win32')
+  command -complete=shellcmd -nargs=1 -bang Launch
+        \ exe 'silent !'.. (&shell =~? '\<cmd\.exe\>' ? '' : 'cmd.exe /c')
+        \ 'start /b ' trim(<q-args>) s:redir | redraw!
+endif
+" if exists(':Launch') == 2
```
chrisbra: Can you enable this condition please?

runtime/autoload/netrw.vim
Konfekt: Likely a misunderstanding, but the idea is that, say, the trailing comma `,` in `.. to https://vim.org, but ...` is not part of the URL `https://vim.org`.

runtime/doc/pi_netrw.txt
```diff
@@ -1469,42 +1469,82 @@ With either form of the command, netrw will first ask for confirmation
...
-text under the cursor; one may change what |expand()| uses via the
-|g:netrw_gx| variable (options include "<cword>", "<cWORD>").  Note that
-expand("<cfile>") depends on the |'isfname'| setting.  Alternatively, one may
-select the text to be used by gx by making a visual selection (see
-|visual-block|) and then pressing gx.
+    * for Windows                     : explorer.exe is used
+    * for Mac OS X               : open is used.
+    * for Linux               : xdg-open is used.
+
+To launch a specific application <app>, type :Launch <app>.
```
chrisbra:
```suggestion
To launch a specific application <app>, type :Launch <app> <filename>
```

chrisbra (commented):
almost there, just a few minor things I noticed.

runtime/autoload/netrw.vim
chrisbra: I see. sounds fragile, but we can keep if for now.

runtime/plugin/netrwPlugin.vim
Konfekt: [I dug](https://github.com/junegunn/fzf/pull/3798/commits) into the [MSYS2 Path conversion magic docs](https://www.msys2.org/docs/filesystem-paths/) to ensure that `/b` is not misinterpreted as a path `\b`.
However, this assumes the `start` script coming with MSYS2 / Git Bash; I am not sure about Cygwin and maybe an additional case distinction is needed (as it only knows `cygstart`!?). I assumed it essentially superseded by MSYS2, Git Bash and WSL and would await feedback.

chrisbra: okay fine. I think you may want to adjust the comment above however. Also the comment mentions `start //b ""` while the code uses `start "" //b`

runtime/doc/pi_netrw.txt
Konfekt: I forgot about `:Open`. Launch is the general background app launcher. It's all documented now.

runtime/autoload/netrw.vim
Konfekt: Maybe in theory not completely waterproof but [in practice](https://github.com/qadzek/link.vim/issues/12#issuecomment-2343590178)

saccarosium:
I think Open should give an error when none of the programs are installed

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,86 @@ set cpo&vim
...
+          \ | redraw!
+  else
+    command -complete=shellcmd -nargs=1 -bang Launch execute ':silent ! nohup' trim(<q-args>) s:redir '&' | redraw!
+  endif
+elseif has('win32')
+  command -complete=shellcmd -nargs=1 -bang Launch
+        \ exe 'silent !'.. (&shell =~? '\<cmd\.exe\>' ? '' : 'cmd.exe /c')
+        \ 'start /b ' trim(<q-args>) s:redir | redraw!
+endif
+if exists(':Launch') == 2
```
saccarosium, line 83: This can be rewritten as:

```vim
if exists(':Launch') == 2
  function s:open(file)
    " Default / Git Bash
    let cmd = ''
    " Windows / WSL
    if executable('explorer.exe')
      let cmd = 'explorer.exe'
    " Linux / BSD
    elseif executable('xdg-open')
      let cmd = 'xdg-open'
    " MacOS
    elseif executable('open')
      let cmd = 'open'
    endif

    if empty(cmd) && !exists('g:netrw_browsex_viewer')
      echoerr "No program found to open this file. For more information see :help Open."
    else
      Launch cmd shellescape(a:file, 1)
    endif
  endfunction
  command -complete=file -nargs=1 Open call s:open(<q-args>)
endif
```

Konfekt: Yeah, I thought about it but found a check on every function call out of proportion for the unlikely event that one of these missing handlers without must be installed while Vim cannot be restarted

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,86 @@ set cpo&vim
...
+  elseif executable('open')
+      let s:cmd = 'open'
+  else
+    s:cmd = ''
+  endif
+  function s:Open(cmd, file)
+    if empty(a:cmd) && !exists('g:netrw_browsex_viewer')
+      echoerr "No executable handler found. See :help Open for default handlers. To customize it, see :help netrw_browsex_viewer"
+    else
+      Launch cmd shellescape(file, 1)
```
saccarosium:
```suggestion
      Launch cmd shellescape(a:file, 1)
```

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,86 @@ set cpo&vim
...
+    endif
+  endfunction
+  command -complete=file -nargs=1 Open call s:Open(<q-args>, s:cmd)
```
saccarosium:
```suggestion
  command -complete=file -nargs=1 Open call s:Open(s:cmd, <q-args>)
```

runtime/plugin/netrwPlugin.vim
```diff
@@ -31,6 +32,86 @@ set cpo&vim
...
+      let s:cmd = 'xdg-open'
+  " MacOS
+  elseif executable('open')
+      let s:cmd = 'open'
+  else
+    s:cmd = ''
+  endif
+  function s:Open(cmd, file)
+    if empty(a:cmd) && !exists('g:netrw_browsex_viewer')
+      echoerr "No executable handler found. See :help Open for default handlers. To customize it, see :help netrw_browsex_viewer"
```
saccarosium: How about "No program to open this file found. See :help Open for more information."? I think using 'executable handler' is a bit confusing.

runtime/plugin/netrwPlugin.vim
Konfekt: Oops, thanks a ton

runtime/plugin/netrwPlugin.vim
Konfekt: Oh my, I must have been rather absent-minded, thank you

runtime/plugin/netrwPlugin.vim
Konfekt: I openend for path (instead of say, locator) as a common term for file and URLs

zzzyxwvut:
I left [some notes](https://github.com/vim/vim/commit/3d7e567ea7392e43a90a6ffb3cd49b71a7b59d1a#diff-04bdb3e603079bce69af82dde6018a473e9ec1da595229d3f7bfc777885a169aR39) for the updated `netrwPlugin.vim` (lines 39,  
99).

runtime/doc/pi_netrw.txt
```diff
@@ -1469,42 +1469,93 @@ With either form of the command, netrw will first ask for confirmation
...
+by default the following regular expression to determine if the text under the
+cursor is a URL:
+>
+	g:netrw_regex_url = '\%(\%(http\|ftp\|irc\)s\?\|file\)://\S\{-}'
 <
+
+Netrw determines which special handler by the following method:
+
+  * if |g:netrw_browsex_viewer| exists, then it will be used to attempt to
+    view files.  Examples of useful settings (place into your <.vimrc>):
```
cpplearner, line 1538:
```suggestion
    view files.
```

PR #15722, opened 2024-09-22
More fixes for GetLatestVimScripts

MiguelBarro:
- A fix for two errors introduced in #15640:
  -  The wrong vim runtime directory name was selected if using powershell or bash on windows. The runtime directory is independent from the shell (only depends on OS windows: _vimfiles_ linux: ._vim_)
  - A faulty pattern for the search of autoload plugins. It has been fixed and the logic updated to contemplate the *package feature* (script predates the feature).
- Allow the user to customize the commands used to decompress the scripts downloaded from `www.vim.org`. This way windows users can delegate into wsl. For example:
```vim
    let g:GetLatestVimScripts_bunzip2= "wsl -e bunzip2"
    let g:GetLatestVimScripts_gunzip= "wsl -e gunzip"
    let g:GetLatestVimScripts_unxz= "wsl -e unxz"
    let g:GetLatestVimScripts_unzip= "wsl -e unzip"
```
will allow the downloaded scripts to be decompress using wsl tools (no need for those tools to be available on windows).

### DIFF
--- a/runtime/autoload/getscript.vim
+++ b/runtime/autoload/getscript.vim
@@ -7,7 +7,9 @@
 "  Installing:	:help glvs-install
 "  Usage:	:help glvs
 "  Last Change:	{{{1
-"   2024 Sep 08 by Vim Project: several small fixes
+"   2024 Sep 08 by Vim Project: several small fixes (#15640)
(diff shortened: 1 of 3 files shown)


chrisbra:
I was worried about changing the `firstdir` pattern, but it seems this is correct. Thanks.

PR #15725, opened 2024-09-23
Remove legacy Amiga code/files

nir9:
Problem: The Vim source tree still contains many Amiga related source code/files even though it is not supported anymore

Solution: Remove the legacy files and information


zeertzjq:
> it is not supported anymore

Source?

nir9:
@zeertzjq README.md states that Amiga is among the systems not supported anymore

zeertzjq:
No it doesn't. It states Amiga DOS is not supported.

nir9:
@zeertzjq AmigaDOS it the name of the operating system that came with the Amiga computer, it's synonymous with just saying Amiga (Amiga is technically the name of the computer itself).

The original version of Vim from 1991 was an Amiga exclusive, since then support has been dropped.

chrisbra:
we had a few PRs fix build errors for Amiga, like #12106 @sodero can you comment, whether Vim still builds and is supported on Amiga?

k-takata:
We dropped support for Win9x a few years ago. So, we can remove the Win9x-related texts.
I think creating a separate PR for the Wn9x parts is better.

nir9:
@k-takata 👍 Opened a new PR for the Win9x stuff (with additional related changes) #15730 
@chrisbra @zeertzjq Looks like indeed after a quick look at the history of patches quite a lot of work is still being done related to Amiga support of more modern Amiga-likes such as AROS or MorphOS, so I'll close this one since it looks less relevant.

What do you think about an alternative PR that will list these as officially supported in the README ?

sodero:
> we had a few PRs fix build errors for Amiga, like #12106 @sodero can you comment, whether Vim still builds and is supported on Amiga?

It still builds and works well on AmigaOS4, MorphOS and AROS.

PR #15730, opened 2024-09-23
Remove some Win9x legacy

nir9:
Problem: The source tree contains some Win9x legacy.

Solution: Remove some of the legacy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,10 +285,10 @@ dist/$(COMMENT_RT): dist/comment
 	echo "Vim - Vi IMproved - v$(VDOT) runtime files for MS-DOS and MS-Windows" > dist/$(COMMENT_RT)
 
 dist/$(COMMENT_W32): dist/comment
-	echo "Vim - Vi IMproved - v$(VDOT) binaries for MS-Windows NT/95" > dist/$(COMMENT_W32)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -285,10 +285,10 @@ dist/$(COMMENT_RT): dist/comment
 	echo "Vim - Vi IMproved - v$(VDOT) runtime files for MS-DOS and MS-Windows" > dist/$(COMMENT_RT)
 
 dist/$(COMMENT_W32): dist/comment
-	echo "Vim - Vi IMproved - v$(VDOT) binaries for MS-Windows NT/95" > dist/$(COMMENT_W32)
+	echo "Vim - Vi IMproved - v$(VDOT) binaries for MS-Windows NT" > dist/$(COMMENT_W32)
```
k-takata:
```suggestion
	echo "Vim - Vi IMproved - v$(VDOT) binaries for MS-Windows" > dist/$(COMMENT_W32)
```
(Same for other part.)

nir9: fixed

chrisbra:
I am slightly worried that removing the edge-case handling of `mch_rename` regresses unneccessarily. Can you please keep this, if this is still an issue that occurs on recent Windows versions?

nir9:
@chrisbra Fixed and adjusted the comment to only mention the edge case that is relevant for modern Windows versions (and is handled by the function logic)

chrisbra:
thanks!

PR #15731, opened 2024-09-23
getcmdprompt() may return the previous prompt

zeertzjq:
Problem:  getcmdprompt() may return the previous prompt.
Solution: Clear the prompt text when a prompt has ended.


Shougo:
@zeertzjq can you check https://github.com/vim/vim/pull/15743 implementation?

Issue #14437, opened 2024-04-07
vimdiff does not work correctly when using &diffexpr


rickhowe:
### Steps to reproduce

I am trying to use diff mode and compare line-by-line by setting a following simple script to `&diffexpr`. This script skips an empty line and continues to compare the next line.

```
set diffexpr=MyDiff()
function! MyDiff()
  let [in, new] = [readfile(v:fname_in), readfile(v:fname_new)]
  let [len_in, len_new] = [len(in), len(new)]
  let out = []
  let [ln_in, ln_new] = [0, 0]
  while ln_in < len_in && ln_new < len_new
    let [text_in, text_new] = [in[ln_in], new[ln_new]]
    let [empty_in, empty_new] = [empty(text_in), empty(text_new)]
    if empty_in == empty_new
      if text_in != text_new
        let out += [(ln_in + 1) . 'c' . (ln_new + 1)]
      endif
      let [ln_in, ln_new] += [1, 1]
    else
      if empty_in | let ln_in += 1 | endif
      if empty_new | let ln_new += 1 | endif
    endif
  endwhile
  call writefile(out, v:fname_out)
endfunction
```
Case 1:

![image](https://github.com/vim/vim/assets/3096934/93f41d5b-6a26-4953-998d-1ca719524b1a)
`:windo diffthis`
![image](https://github.com/vim/vim/assets/3096934/cf208758-ca14-4b54-a895-99b98d5a3fe5)
Both lines should be highlighted as changed lines, but  the 2nd one in the left window is highlighted as added line.
In this case,` ['1c1', '2c2']` is written to `v:fname_out`.

Case 2:

![image](https://github.com/vim/vim/assets/3096934/532d0f8b-a0a4-4d80-9edb-37c369e70b0f)
`:windo diffthis`
![image](https://github.com/vim/vim/assets/3096934/c73d35cf-d5bc-470c-b5d2-c6a600e0af7f)
Both lines should be highlighted as changed lines, but  the 2nd one in the right window is not highlighted.
In this case,` ['1c1', '3c2']` is written to `v:fname_out`.

### Expected behaviour

Both lines should be highlighted as changed lines.

### Version of Vim

9.1.0271 but same thing happens in 8.2 as well.

### Environment

Windows 10

ynkdir:
This is related to https://github.com/vim/vim/commit/d4b96bc637e22785ad956b22ff0de97a89a869e3

ynkdir:
[...]

PR #15735, opened 2024-09-24
Fix vim/vim#14437 and some other problems in diff mode.

ynkdir:
In diff mode Vim merges overlapped diff blocks.  But it doesn't work expectedly in some situation.

Here is diagram of added test.

https://viewer.diagrams.net/?tags=%7B%7D&lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=vim_diff_read.drawio#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1t9z_dy3Q3zZBHx5wG_mtEekAnI89ANfn%26export%3Ddownload


chrisbra:
Thanks. Can you give an example what vimdiff currently does and how it should look like?

ynkdir:
For example.

```
$ cat a.txt
a
b
c
$ cat b.txt
a
x
c
$ cat c.txt
y
b
c
$ vim --clean -d a.txt b.txt c.txt
```

result

result

expected

expected

chrisbra:
okay thanks. That is clearly wrong

ynkdir:
Thank you.

PR #15732, opened 2024-09-24
Added am.po

gagikh:
Created po file for Armenian lang tranlation

PR #15736, opened 2024-09-24
Update MAINTAINERS

gagikh:


zeertzjq:
Shouldn't this be included in #15732?

chrisbra:
let's close it here now. When you create a new translation file, please also add a new entry to the Maintainers file then. Thanks!

Issue #15727, opened 2024-09-23
comment on non-linewise text objects


Konfekt:
### Problem

Inline commenting is not uncommon, for example, `grep -E '\S\s+"\s*[a-zA-Z ]+$'  $VIMRUNTIME/*/*.vim | wc -l` yields 625 matches. Neither `gc$` nor a visual selection from the cursor to the end of the line and typing `gc` comments the intended part but the complete line

### Expected behavior

`gc` comments the given text-object

zeertzjq:
This will require an option separate from 'commentstring', since inline comments may have different syntax than linewise comments, and not all languages support inline comments.

Konfekt:
So why not set gC in ftplugin whenever commentstring is applicable in-line? That's not too rare.

chrisbra:
Honestly, I don't think we should support this. A non-linewise comment, may have other semantics, you may need an end comment marker as well. What about when it is already in a commented line, sometimes those are not valid inside commented out regions (thinking of xml `<!--` comment marker). 

It might be a nice VimScript side-project, but I would prefer if we keep the builtin commenting feature simple.

Konfekt:
@chrisbra I am sorry, I read this while finishing my PR; but it does not seem such a big story if restricted to those `&commentstrings` supporting inline comments which are quite common.
If yours do not, say by choice `/* ... */` in C instead of `// ...`, then `gC` is synonym to `gcc`. 
Let me not map it at all and just document it instead, so that no harm should be done.

PR #15737, opened 2024-09-24
runtime:comment: add gC mapping to (un)comment rest of line

Konfekt:
As discussed in https://github.com/vim/vim/issues/15727#issuecomment-2371636002 only works differently from `gcc` whenever the comment marker precedes code as happens for many languages such as Python, Lua, Vim, Sh, ...
To illustrate for C: `set cms =//\ %s` lets `gC` operate on the end of the line, whereas with `set cms=/*\ %s\ */` it behaves like `gcc`.

### DIFF
--- a/runtime/pack/dist/opt/comment/autoload/comment.vim
+++ b/runtime/pack/dist/opt/comment/autoload/comment.vim
@@ -1,7 +1,7 @@
 vim9script
 
 # Maintainer: Maxim Kim <habamax@gmail.com>
-# Last Update: 2024-04-26
+# Last Update: 2024-09-24
 #
 # Toggle comments
 # Usage:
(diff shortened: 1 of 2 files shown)


chrisbra:
@habamax  ?

habamax:
I can’t test it now. But provided it works for most of file types am not against it.

chrisbra:
alright, let me include it then.

b0ae989c:
There may be a bug in this commit. Testing with:
```sh
echo set number > test.vim
vim -U NORC test.vim
```
Reproduce:
```
:echom has('patch-9.1.0749')
:packadd comment
0
gcc
```
The commented line becomes
```vim
s" set number
```
which is different from the expected result
```vim
" set number
```

Konfekt:
Strange, that was just what was supposed to be solved

Konfekt:
Oh, this is an old commit. The remedy is https://github.com/vim/vim/pull/15774

PR #15739, opened 2024-09-24
Change default backupcopy to yes to avoid confusing file watchers

### LINKED ISSUE #15733: Vim file-save strategy breaks file-watching applications
### Steps to reproduce

I'm not a vim user so this is only an indirect report (see original: https://github.com/timholy/Revise.jl/pull/841).

@acertain, @rafaqz feel free to chime in with additional detail.

### Expected behaviour

My understanding is that the sequence of operations when vim saves a file is:
1. rename file to backup
2. write & fsync new file
3. remove backup

Unfortunately, for any application that performs file-watching, it may be triggered by 1 before 2 is complete. I'm sure I'm naive about the potential problems it might cause, but it would seem that a more robust approach might be:
1. write & fsync to a temporary file
2. rename the temporary file to the original filename, thus "displacing" the original file

### Version of Vim

?

### Environment

?

### Logs and stack traces

_No response_

nir9:
Problem: Current default saving behavior confuses file watching programs, for example ```tail -f asdf.txt``` doesn't refresh when Vim saves changes to asdf.txt

Solution: Change the default value to true, which causes Vim to copy a backup of the edited file rather than renaming, and as a result watchers don't get confused since the original stays at the same place

closes #15733


ghgary:
Don't change default values. Find some other solution.

nir9:
@ghgary Recently addressed this concern in the linked issue #15733

chrisbra:
I am not convinced that changing the default is the way to go. I see file-watcher use-case as some kind of special. If you have that use case, you should set the 'backupcopy' option value. But I don't want to change a value, that users expect and know and has been working well since at least Vim 7.

nir9:
👍 closing

PR #15740, opened 2024-09-24
runtime(java): Recognise the CommonMark form (///) of Javadoc comments

zzzyxwvut:
Complement `g:java_ignore_javadoc` with `g:java_ignore_html`  
and `g:java_ignore_markdown` to allow selectively disabling  
the recognition of HTML and CommonMark respectively.

(Note that this is not a preview feature.)

###### ======================== LIMITATION ========================

According to the syntactical details of JEP 467 :

> Any leading whitespace and the three initial / characters
> are removed from each line.
>
> The lines are shifted left, by removing leading whitespace
> characters, until the non-blank line with the least
> leading whitespace has no remaining leading whitespace.
>
> Additional leading whitespace and any trailing whitespace
> in each line is preserved, because it may be significant.

the following example:

```java
///    A summary sentence.
///     A list:
///      - Item A.
///     - Item B.
///
///     Some code span, starting here `
///      1 + 2 ` and ending at the previous \`.
```

should be interpreted as if it were written thus:

```java
///A summary sentence.
/// A list:
///  - Item A.
/// - Item B.
///
/// Some code span, starting here `
///  1 + 2 ` and ending at the previous \`.
```

Since automatic line rewriting will not be pursued, parts of  
such comments having significant whitespace may be ‘wrongly’  
highlighted.  For convenience, a `&fex` function is defined to  
‘correct’ it: `g:javaformat#RemoveCommonMarkdownWhitespace()`
(`:help ft-java-plugin`).

References:
https://openjdk.org/jeps/467
https://spec.commonmark.org/0.31.2


zzzyxwvut:
@dkearns, recalling [our previous discussion](https://github.com/vim/vim/discussions/14117#discussioncomment-8709402) about the quirks  
of `defaults.vim` and a way to get by without sourcing this  
file, please have a look at `:match` and its dependency on  
`:syntax on` for _some_ highlighting groups.

```shell
vim --clean -u NONE '+match Todo /foo/'
```

`E28: No such highlight group name: Todo`

With `:defer`, `:match` can still be used for a syntax test:

```
VIM_TEST_SETUP defer execute('match Todo /foo/')
```

https://github.com/vim/vim/blob/eb6bd720433c54dfc4526806ca79e790ae6b7364/runtime/syntax/testdir/input/java_comments_markdown.java#L10

chrisbra:
thanks!

serban:
FYI, this commit introduces circular syntax includes between the Java and Markdown syntax files that ship with Vim.

When `java` is in the list of `g:markdown_fenced_languages`, the Markdown syntax file includes the Java syntax file ｢1｣ which in turn includes the Markdown syntax file ｢2｣. This blows up with a lovely `E847: Too many syntax includes` error.

I worked around this by setting `g:java_ignore_markdown` in my `vimrc` ｢3｣.

｢1｣ https://github.com/vim/vim/blob/85f054aa3f0fb9530712d0897e3c8ba29946fad4/runtime/syntax/markdown.vim#L36
｢2｣ https://github.com/vim/vim/blob/85f054aa3f0fb9530712d0897e3c8ba29946fad4/runtime/syntax/java.vim#L374
｢3｣ https://github.com/serban/dotfiles/commit/84ae1b0127278205b8a472d558e957d20d0ddc0b

zzzyxwvut:
My apologies for the inconvenience, I have followed the  
_opt-out_ precedent established in `g:java_ignore_javadoc`  
when `g:java_ignore_html` and `g:java_ignore_markdown` were  
considered for introduction.  It makes sense for Java source  
files: you may want for the documentation comments to stand  
out next to the other comments, you may want for some parts  
of documentation to stand out, especially when you read many  
‘pages’ of API description at a sitting; should you want to  
suppress it, `g:java_ignore_javadoc` etc. are there.

Because `g:main_syntax` is not ‘re-entrant’, to correctly  
support circular inclusion, it would require a good deal of  
cooperative special casing in the candidate syntax files  
from all concerned maintainers, a strategy to manage the  
doubling of syntax items owing to `:syn-include`, and then  
arranging for a simultaneous push of all updated and *TESTED*  
files into `$VIMRUNTIME`.  Is it worth the effort?  I'd rather  
reject circularity from the outset.

Please see a new PR (#15796) with the necessary adjustment  
made for the inclusion of Markdown (and HTML).

With the proposed change and assuming that there is, say in  
`vimrc`:

```vim
let g:markdown_fenced_languages = ['java', 'html']
```

add to your `~/.after/ftplugin/java.vim` (note `ftplugin` NOT  
`syntax`):

```vim
if exists("g:markdown_fenced_languages") &&
	\ !(exists("g:java_ignore_javadoc") ||
	\ exists("g:java_ignore_markdown"))
    let s:idx = index(g:markdown_fenced_languages, 'java')

    if s:idx > -1
	call remove(g:markdown_fenced_languages, s:idx)
    endif

    unlet s:idx
endif
```
also, optionally add to your `~/.after/ftplugin/markdown.vim`  
(again, note `ftplugin` NOT `syntax`):

```vim
if exists("g:markdown_fenced_languages") &&
	\ index(g:markdown_fenced_languages, 'java') < 0
    call add(g:markdown_fenced_languages, 'java')
endif
```

Now you can enjoy the support of Markdown documentation  
comments in Java source files _and_ still be able to see Java  
code snippets highlighted in Markdown fenced blocks but  
without the recognition of Markdown documentation comments.  
(Unless the new version of the syntax file is made available  
from `$VIMRUNTIME`, and from `~/.vim/syntax` if necessary,  
OTHER discoverable file versions will be used whose  
behaviour may interfere with this fix.)

In general, the setting of `g:markdown_fenced_languages` is  
_useless for Java source files_, regardless of context and  
circularity, since neither `{@snippet}` nor `{@code}` nor  
`<code>` tags in the HTML documentation comments support the  
recognition and highlighting of their contents; and the same  
tags, in addition to code spans and code blocks, follow suit  
in the Markdown documentation comments.

PR #15742, opened 2024-09-25
:keep* commmands are sometimes misidentified as :k

dkearns:
Problem:  The :keep{alt,jumps,marks,patterns} commmands are sometimes
          misidentified as :k.
Solution: Make sure one_letter_cmd() only returns true for :k and not
          other :keep* commands.

This currently manifests as missing completion for :keep* commands and incorrect results from fullcommand().

E.g., fullcommand("keepmarks") returns "k" rather than "keepmarks".

The correct command, however, is executed as command modifiers are handled specially in do_one_cmd() rather than using find_ex_command().

Fix exists(':k') so that it returns 2 for a full match.

### DIFF
--- a/src/ex_docmd.c
+++ b/src/ex_docmd.c
@@ -3584,7 +3584,9 @@ skip_option_env_lead(char_u *start)
 /*
  * Return TRUE and set "*idx" if "p" points to a one letter command.
  * If not in Vim9 script:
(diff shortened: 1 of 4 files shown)


chrisbra:
I think this makes a lot of sense. Thanks!

PR #15743, opened 2024-09-25
Fix getcmdprompt()

h-east:
The size limit on the string retrieved by `getcmdprompt()` has been removed.
Since there is no size limit on the {msg} of `confirm()` and the {prompt} of `input()`, it makes sense that there should also be no size limit on `getcmdprompt()`.

### DIFF
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -30,7 +30,6 @@ static cmdline_info_T ccline;
 
 #ifdef FEAT_EVAL
 static int	new_cmdpos;	// position set by set_cmdline_pos()
-static char_u	*current_prompt;
 #endif
 
 static int	extra_char = NUL;  // extra character to display when redrawing
@@ -50,9 +49,6 @@ static void	alloc_cmdbuff(int len);
 static void	draw_cmdline(int start, int len);
 static void	save_cmdline(cmdline_info_T *ccp);
 static void	restore_cmdline(cmdline_info_T *ccp);
-#ifdef FEAT_EVAL
-static char_u	*get_prompt(void);
-#endif
 static int	cmdline_paste(int regname, int literally, int remcr);
 static void	redrawcmdprompt(void);
 static int	ccheck_abbr(int);
@@ -4236,24 +4232,6 @@ f_getcmdline(typval_T *argvars UNUSED, typval_T *rettv)
 }
 
 /*
- * Get current command line prompt.
- */
-    static char_u *
-get_prompt(void)
-{
-    return current_prompt;
-}
-
-/*
- * Set current command line prompt.
- */
-    void
-set_prompt(char_u* str)
-{
-    current_prompt = str;
-}
-
-/*
  * "getcmdpos()" function
  */
     void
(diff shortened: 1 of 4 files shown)


h-east:
Hmm, AddressSanitizer error has occurred.

```
addr2line: DWARF error: invalid or unhandled FORM value: 0x23
==11021==ERROR: AddressSanitizer: heap-use-after-free on address 0x5020000727d0 at pc 0x5650ff738b03 bp 0x7ffcfa91cec0 sp 0x7ffcfa91c688
READ of size 3 at 0x5020000727d0 thread T0
    #0 0x5650ff738b02 in ___interceptor_strlen ??:?
    #1 0x565100111771 in vim_strsave /home/runner/work/vim/vim/src/strings.c:26:11
    #2 0x5650ffb2a542 in f_getcmdprompt /home/runner/work/vim/vim/src/ex_getln.c:4275:40
...
```

src/ex_getln.c
```diff
@@ -30,7 +30,7 @@ static cmdline_info_T ccline;
 
 #ifdef FEAT_EVAL
 static int	new_cmdpos;	// position set by set_cmdline_pos()
-static char_u	current_prompt[CMDBUFFSIZE + 1] = "";
+static char_u	*current_prompt;
```
Shougo:
```suggestion
static char_u	*current_prompt = NULL;
```

src/ex_getln.c
```diff
@@ -4250,7 +4250,7 @@ get_prompt(void)
     void
 set_prompt(char_u* str)
 {
-    vim_strncpy(current_prompt, str, sizeof(current_prompt) - 1);
+    current_prompt = str;
```
Shougo: Thank you for the improvements.

I think this code is better.

```suggestion
    VIM_CLEAR(current_prompt);
    current_prompt = vim_strsave(str);
```

h-east:
I investigated further the implementation of "getcmdprompt()".

First of all, since there are existing recursive command line implementations [^1], I thought it would be a bad idea to have information about the current prompt outside the `cmdline_info_T` structure.

Since ccline has a cmdprompt member, I think it would make sense to refer to that.
The implementation is also simpler.

However, as shown in the implementation, if the prompt message contains a newline character, the message is output up to the final newline, and the part after the final newline is treated as the prompt. [^2]
The corrected PR now behaves as described above.
(I've also added tests)
https://github.com/vim/vim/blob/ac8a10ab7672aa5fd359f76d34d8b0187728b92d/src/ex_getln.c#L4903-L4919

Is there any problem if `getcmdprompt()` does not return the string passed as the first argument, `input()` or `confirm()`?
In that case, please close this PR and deal with the issue separately.

[^1]: ccline, prev_ccline, save_cmdline(), restore_cmdline()
[^2]: {msg} of confirm() and the {prompt} of input()

Shougo:
> However, as shown in the implementation, if the prompt message contains a newline character, the message is output up to the final newline, and the part after the final newline is treated as the prompt. [2](https://github.com/vim/vim/pull/15743#user-content-fn-2-96b6d569e097cd63c434424785f3d7f8)

> Is there any problem if getcmdprompt() does not return the string passed as the first argument, input() or confirm()?
In that case, please close this PR and deal with the issue separately.

The first argument contains new lines, it can be thought of as containing prompt string(the last line) and regular message.
I think if the feature is documented, the simpler implementation is better.
Because `ccline.cmdprompt` is the real prompt in Vim.

chrisbra:
That implementation looks cleaner. Thanks!

PR #15744, opened 2024-09-25
filetype: Some notmuch configs are not recognised

julio-b:
Problem:  Some notmuch configs are not recognised
Solution: Detect notmuch profile configuration files

Reference: https://notmuchmail.org/doc/latest/man1/notmuch-config.html#configuration


chrisbra:
thanks!

PR #15746, opened 2024-09-26
docs(options.txt): improve `:set` command behavior table

Milly:
`:set` command behavior table is inaccurate.

`:set option?` behaves differently depending on whether the local option is set or not-set.


Milly:
In other words `:set option?` will show the currently active values.

PR #15748, opened 2024-09-26
runtime(glvs): Correct the tag name of glvs-autoinstal

zzzyxwvut:


chrisbra:
Ah thanks, I just also noticed.

PR #15751, opened 2024-09-27
filetype: libtool files are not recognized

Freed-Wu:
Problem: Libtool files are not recognized
Solution: recognize libtool files as sh filetype

### DIFF
--- a/runtime/filetype.vim
+++ b/runtime/filetype.vim
@@ -567,6 +567,9 @@ au BufNewFile,BufRead *.dsp				call dist#ft#FTdsp()
 au BufNewFile,BufRead *.xcu,*.xlb,*.xlc,*.xba		setf xml
 au BufNewFile,BufRead psprint.conf,sofficerc		setf dosini
 
+" Libtool files
+au BufNewFile,BufRead *.lo,*.la,*.lai		setf sh
+
 " Lynx config files
 au BufNewFile,BufRead lynx.cfg			setf lynx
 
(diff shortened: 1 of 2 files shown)


chrisbra:
thanks!

chrisbra:
Hm, I have to wonder: Currently those are detected as `conf`  files, and for the few examples I have seen, it seems the `conf` filetype is more appropriate, no?

Freed-Wu:
`libtool` is a program written in sh, which read this files as sh.

However, these files are generated by libtool, which only use simple shell syntax (comment, `var='value'`), so think there are conf is also OK.

chrisbra:
can you then please change it then to make the `conf` filetype explicit?

Freed-Wu:
@ileanadumitrescu95 (libtool maintainer) What is your point?

ileanadumitrescu95:
These are control files that set variables for GNU Libtool to read from. GNU Libtool sources these files with sh for configuration, so either filetype is correct to me.

Freed-Wu:
> GNU Libtool sources these files with sh for configuration

that's why I choose sh :smile:

chrisbra:
yeah, but if I would be editing such a file and see the filetype is detected as `shell` I would assume all shell constructs would be valid and I might be tempted to add those there (after all the syntax supports it). Therefore I'd prefer if we keep it as `conf`

Freed-Wu:
@ileanadumitrescu95 so do you prefer conf, or sh?

ileanadumitrescu95:
> ... either filetype is correct to me.

I do not prefer one or the other

Freed-Wu:
> I would assume all shell constructs would be valid

IMO, all shell syntaxes are valid in `*.la`, ... And treat them as shell can make it easy to use those developer tools for shell. Just my 2c.

PR #15754, opened 2024-09-27
Fix negative numbers for gnu indent config file

jmdevin:
Some gnu indent options take negative numbers (e.g. --indent-label).
Add matching for an optional single '-' before the number.


chrisbra:
Thanks. I think this is fine @dkearns ?

dkearns (approved):

chrisbra:
okay will do

PR #15756, opened 2024-09-27
tests: Test_halfpage_longline() fails on large terminals

### LINKED ISSUE #15755: Test halfpage_longline() fails on large terminals
### Steps to reproduce

1. Open a terminal window with large dimensions (e.g.  `stty size: 112 478`)
2. `make`
3. `make test`

### Expected behaviour

The test `halfpage_longline()` shouldn't fail if executed in large terminal windows.
Works fine for me if the terminal is smaller (e.g. `stty size 112 214`).

I looked a bit into the test. Looks like the repeat is not large enough to fill terminals with many columns.

I tried to patch the test:

```diff
diff --git a/src/testdir/test_normal.vim b/src/testdir/test_normal.vim
index adb5ce725..b4509ab5d 100644
--- a/src/testdir/test_normal.vim
+++ b/src/testdir/test_normal.vim
@@ -4265,6 +4265,8 @@ endfunc

 " Test for Ctrl-D with long line
 func Test_halfpage_longline()
+  set lines=24
+  set columns=80
   10new
   call setline(1, ['long'->repeat(1000), 'short'])
   exe "norm! \<C-D>"
```

I set the lines and columns to the defaults instead of using the actual terminal dimensions. After patching the code, the test seems to work.
But I'm not sure whether it's sensible to do it that way.

### Version of Vim

9.1.0744

### Environment

OS: Gentoo
Terminal: st
TERM=xtern (set for the tests)
Shell: dash

### Logs and stack traces

```shell
Failures:
        From test_normal.vim:
        Found errors in Test_halfpage_longline():
        command line..script /home/kai/build/vim-normal/vim-9.1.0744/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_halfpage_longline line 4: Expected 2 but got 1
```

zeertzjq:
Problem:  Test_halfpage_longline() fails on large terminals.
Solution: Use a window with known width.

fixes: #15755

### DIFF
--- a/src/testdir/test_normal.vim
+++ b/src/testdir/test_normal.vim
@@ -4266,6 +4266,7 @@ endfunc
 " Test for Ctrl-D with long line
 func Test_halfpage_longline()
   10new
+  40vsplit
   call setline(1, ['long'->repeat(1000), 'short'])
   exe "norm! \<C-D>"
   call assert_equal(2, line('.'))


lazypingu:
Works fine for me! Thanks @zeertzjq for fixing!

chrisbra:
thanks!

PR #15760, opened 2024-09-28
feat: add completeitemalign

glepnir:
Adding an new option `completeitemalign` abbr is `cia` to custom the complete-item order in popupmenu.

close #14006


chrisbra:
Hm, isn't that the same as #14006 ?

glepnir:
only purpose same.  see code. should close that pr.

chrisbra:
Hm, the code seems to assume, that there will always be 3 items in 'completealignitems'. But what about:
```
set cia=abbr,abbr
set cia=
set cia=abbr,kind,menu,abbr
set cia=invalid
```

can you add test cases for those cases as well please?

glepnir:
update

src/testdir/test_popup.vim
```diff
@@ -1581,4 +1581,64 @@ func Test_pum_user_kind_hlgroup()
   call StopVimInTerminal(buf)
 endfunc
 
+function Test_pum_completeitemalign()
```
zeertzjq:
```suggestion
func Test_pum_completeitemalign()
```

runtime/doc/options.txt
```diff
@@ -2100,6 +2100,15 @@ A jump table for the options with a short description can be found at |Q_op|.
 	This option cannot be set from a |modeline| or in the |sandbox|, for
 	security reasons.
 
+						*'completeitemalgin'* *'cia'*
```
```suggestion
						*'completeitemalign'* *'cia'*
```

src/option.h
```diff
@@ -514,6 +514,8 @@ EXTERN int	p_confirm;	// 'confirm'
 #endif
 EXTERN int	p_cp;		// 'compatible'
 EXTERN char_u	*p_cot;		// 'completeopt'
+EXTERN char_u	*p_cia;		// 'completeitemalign'
+EXTERN unsigned cia_flags;	// order flags of 'completeitemalign'
 EXTERN unsigned	cot_flags;	// flags from 'completeopt'
```
zeertzjq: The order of variables here may be a bit strange: `p_cot` comes before `p_cia` but `cot_flags` comes after `cia_flags`.

glepnir: right. move to top of p_cot

src/optionstr.c
```diff
@@ -1635,6 +1635,62 @@ expand_set_completeopt(optexpand_T *args, int *numMatches, char_u ***matches)
 	    matches);
 }
 
+/*
+ * The 'completeitemalgin' option is changed.
```
zeertzjq:
```suggestion
 * The 'completeitemalign' option is changed.
```

glepnir: emm i need open spell check 😭

src/vim.h
```diff
@@ -2372,6 +2372,17 @@ typedef enum {
...
 
+/*
+ * Array indexes used for cp_text[].
+ */
+typedef enum {
+     CPT_ABBR,		// "abbr"
+     CPT_KIND,		// "kind"
+     CPT_MENU,		// "menu"
+     CPT_INFO,		// "info"
+     CPT_COUNT,		// Number of entries
```
zeertzjq:
```suggestion
    CPT_ABBR,		// "abbr"
    CPT_KIND,		// "kind"
    CPT_MENU,		// "menu"
    CPT_INFO,		// "info"
    CPT_COUNT,		// Number of entries
```

src/optionstr.c
```diff
@@ -1635,6 +1635,62 @@ expand_set_completeopt(optexpand_T *args, int *numMatches, char_u ***matches)
...
+	    count++;
+	}
+	else
+	    return e_invalid_argument;
+
+        p = vim_strchr(p, ',');
+	if (p == NULL)
+	    break;
+        if (*p == ',')
+            p++;
```
zeertzjq: This seems to allow invalid option values like `abbr1234,kind,menu`

glepnir: `  if (STRNCMP(p, "abbr", 4) == 0 && (p[4] == ',' || p[4] == NUL))` something like this will avoid invalid option ..any better idea ?

glepnir: oh copy_option_part

glepnir:
option tag is complaining about the location of cia.. maybe it should be a different abbreviation?

zeertzjq:
`cd runtime/doc; make tags`

src/popupmenu.c
```diff
@@ -621,28 +656,25 @@ pum_redraw(void)
...
 	    width = 0;
 	    s = NULL;
-	    switch (round)
-	    {
-		case 0: p = pum_array[idx].pum_text; break;
-		case 1: p = pum_array[idx].pum_kind; break;
-		case 2: p = pum_array[idx].pum_extra; break;
-	    }
+	    p = pum_order_item_isabbr(order, j) ? pum_array[idx].pum_text
+						: pum_get_item(idx, item_type);
```
zeertzjq: Why not use `pum_get_item` directly here?

src/popupmenu.c
```diff
@@ -774,33 +805,40 @@ pum_redraw(void)
...
 	    else
-		n = 1;
+		n = order[j] == CPT_ABBR ? 1 : 0;
+
+	    if (j + 1 < 3)
+	    {
+		if (pum_order_item_isabbr(order, j + 1))
+		    next_isempty = pum_array[idx].pum_text == NULL;
+		else
+		    next_isempty = pum_get_item(idx, order[j + 1]) == NULL;
```
zeertzjq: Same here, why not use `pum_get_item()` directly?

glepnir: function is abstracted later. So I forgot some adjustments.

src/popupmenu.c
```diff
@@ -549,19 +571,25 @@ pum_redraw(void)
...
     char_u	*s;
     char_u	*p = NULL;
-    int		totwidth, width, w;
+    int		totwidth, width, w;  // total-width item-width char-width
     int		thumb_pos = 0;
     int		thumb_height = 1;
-    int		round;
+    int		item_type;
+    int		order[3];
+    int		next_isempty = 0;
```
zeertzjq:
```suggestion
    int		next_isempty = FALSE;
```

src/popupmenu.c
```diff
@@ -10,6 +10,7 @@
 /*
  * popupmenu.c: Popup menu (PUM)
  */
+#include "libvterm/include/vterm.h"
```
zeertzjq: ?

glepnir: idk.. seems like lsp completion additontextedit?

chrisbra:
thanks all!

PR #15761, opened 2024-09-28
filetype: bun and deno history files are not recognized

Freed-Wu:
Problem: Bun and deno history files are not recognized
Solution: Recognize bun and deno history files as javascript filetype


chrisbra:
thanks!

PR #15762, opened 2024-09-28
runtime: recognize .http files

ribru17:
Based on the specification found
[here](https://learn.microsoft.com/en-us/aspnet/core/test/http-files?view=aspnetcore-8.0#comments)

### DIFF
--- a/.github/MAINTAINERS
+++ b/.github/MAINTAINERS
@@ -178,6 +178,7 @@ runtime/ftplugin/hgcommit.vim		@k-takata
 runtime/ftplugin/hlsplaylist.vim		@avidseeker
 runtime/ftplugin/hog.vim		@wtfbbqhax
 runtime/ftplugin/html.vim		@dkearns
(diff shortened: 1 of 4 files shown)


chrisbra:
thanks!

chrisbra:
Hm, looking at it, it seems http might be a bit too generic? Since this is is apparently related to asp, should this be `asphttp` instead?

ribru17:
Personally, I think it should still be just `http`; I provided a poor link, looks like [this one](https://github.com/JetBrains/http-request-in-editor-spec/blob/master/spec.md) is better. It is provided by JetBrains, but still looks like it aims to be a general spec for all editors. So I think the file format is agnostic in this way. Also, on GitHub, a [search for HTTP files](https://github.com/search?q=lang%3Ahttp&type=code) recognizes them as just the `http` language which again seems to be standardized (not just for ASP stuff or VS)

chrisbra:
okay, fair enough

PR #15764, opened 2024-09-28
ci.yml: macos-14 --> macos-15

pheiduck:


chrisbra:
Thanks, but there are still some failures. Let me re-trigger CI.

chrisbra:
Hm, this test failure looks suspicious:
```
Found errors in Test_normal_8g8():
	command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_normal_8g8 line 16: Expected [0, 1, 4, 0, 1] but got [0, 1, 5, 0, 1]
```

I haven't seen this failure before, did something change in MacOS 15?

ychin:
I can take a look at this, since I need to upgrade my CI to use macOS-15 on MacVim anyway.

My guess is Apple silently ~~broke~~ changed something under the hood without telling people again.

ychin:
> We use already gnu diff for tests so it might not be a hustle to add iconv: https://formulae.brew.sh/formula/cconv at least for the macos-15 runner

There's no need to do that because Apple already fixed iconv in macOS 14.2 or so (see https://github.com/vim/vim/pull/13943#issuecomment-1928167550). Note that using GNU versions of diff in the macOS tests should itself be considered an undesirable hack (but necessary here due to a bug in Apple's implementation), because by default macOS users are using Apple diff rather than GNU diff, so the tests are testing a version that is *not* what a normal user would see.

ychin:
I played around with this, and the culprit is Xcode 16, rather than macOS itself. It appears that the version of clang used in Xcode 16 has some odd issues when building optimized code (`-O2` or `-O3`). In the function `utf_find_illegal()`, it tries to inline both `utf_ptr2len` and `utf_ptr2char` and combine the logic together and the resulting code doesn't seem to work at all. Looking through the disassembly I couldn't quite understand why the compiler generated the code that it did.

This only happens using Xcode 16's clang and using optimized code. Interestingly if I turn on UBSAN (`-fsanitize=undefined`) just to see if this was due to any undefined behaviors, the generated code now works correctly…

I played around with other compilers or versions of clang (https://godbolt.org/z/9ndxb31Mo) and they don't seem to exhibit this behavior and they all work correctly (if you compile the same code snippet on a Mac running Xcode 16 it will print the wrong results).

May have to file a bug to Apple. Probably need to look up how to do that though as there are a gazillion ways to do it and none of them easy.

Note: GitHub Actions support using Xcode 16.1 beta, and that didn't fix the issue.

ychin:
I have filed the bug to Apple. Will see if they respond or fix the issue. Meanwhile, I don't think anyone should be using this version of Xcode to compile code (unless this is a completely obscure bug that somehow only Vim hits)…

clason:
You can try adding `-Wl,-no_deduplicate` to the compiler flags. (Apple Clang 16 enables a new deduplication algorithm by default, which... has issues.)

ychin:
> You can try adding `-Wl,-no_deduplicate` to the compiler flags. (Apple Clang 16 enables a new deduplication algorithm by default on release builds (only), which... has issues.)

Oh interesting, thanks for the suggestion. I tried the flags and unfortunately it did not work (using `clang -o program_opt -g3 -Wall -Wl,-no_deduplicate -O2 test.c`).

When you said "has issues", are there other threads that you are aware of? I tried looking around but couldn't find it but maybe I missed them. Maybe they would shed some light.

clason:
Ah, then it's likely a different issue. I only know about the linker step (which broke LuaJIT).

ychin:
I just filed a [bug](https://github.com/Homebrew/homebrew-core/issues/195325) at Homebrew. It seems like they are using Xcode 16 to compile bottles (pre-compiled binary) which makes sense, but that means they are distributing a version of Vim that has miscompiled behaviors due to the codegen bug described above. I'm honestly not sure what a good solution here is as I don't think they are usually open to compiling with older toolchain as Homebrew has a "always use latest" philosophy. Will see what they say.

ychin:
An update from that Homebrew thread is that they identified that when using Xcode 16, you can use `-mllvm -enable-constraint-elimination=0` to disable the offending optimization pass. I am recommending Homebrew to use this flag when building Vim and Neovim. We could technically add this flag and then add macos-15 to the runners, but to me I would rather just wait for Apple to fix the bug.

clason:
> Neovim

Not needed for Neovim, btw.

clason:
> When you said this is not needed for Neovim did you mean you guys are adding this flag yourself to your build scripts to alleviate the issue?

No, I mean Neovim works fine without that flag. (I've been building locally with Xcode 16 for months.)

(Which is how I caught the [above-mentioned, different, issue with fruit clang 16] -- not their most polished release...)

ychin:
> No, I mean Neovim works fine without that flag. (I've been building locally with Xcode 16 for months.)

What does "works fine" mean though? This is a subtle behavior bug. Unlike the other issue you mentioned, this bug does not generate a build error. It only generates slightly incorrect code that affects very specific parts of the program. Right now the `8g8` command is the only known instance. Neovim's `8g8` code (`utf_find_illegal`) is similar enough to Vim's and is therefore subject to the same miscompilation. I tried the Homebrew version and also tried building Neovim locally and in both instances Neovim shows the buggy behavior.

Here's an asciicinema video to demonstrate:

[asciicast](https://asciinema.org/a/683518)

Just in case it wasn't clear, in the video above, the cursor is supposed to land on the `<e2>` part, *not* `<82>`. You can compare it with older versions of Neovim or a Linux build etc and see that the behaviors are different.

ychin:
Just a random tidbit to add and one thing that confused me at first and took me a second to figure out was that Neovim is indeed running the same tests as Vim, in the "oldtest" part of matrix ([example](https://github.com/neovim/neovim/actions/runs/11495032754/job/31993695600)). I was wondering why all the tests were passing for Neovim even though you have upgraded to macos-15 runner while it fails for Vim. The reason turned out to be that Neovim CI [uses Debug instead of Release for building](https://github.com/neovim/neovim/blob/008782208d826a03ab046bb2cb1bdda27c6554d0/.github/workflows/test.yml#L155-L156) and therefore didn't trigger the clang miscompilation bug (which requires `-O2` or `-O3`).

ychin:
From https://github.com/Homebrew/homebrew-core/issues/195325#issuecomment-2492636277, it seems that the Xcode clang compiler bug has been fixed in Xcode 16.2 beta 3. Once this gets released and GitHub Actions adds support for Xcode 16.2 I can submit a PR to use that so we can migrate to macos-15 for CI.

PR #15773, opened 2024-09-30
Minor grammar change

TheProdigyLeague:
please consider revise

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,10 @@ or one of the [Maillists](https://www.vim.org/community.php).
 ## What is Vim?
 
 Vim is a greatly improved version of the good old UNIX editor
-[Vi](https://en.wikipedia.org/wiki/Vi_(text_editor)).  Many new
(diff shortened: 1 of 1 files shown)


chrisbra:
It is quite common in the help files to have 2 spaces after a sentence. It doesn't seem to make a difference for the rendered Markdown file. So let's just keep it how it is.

TheProdigyLeague (commented):
Line: 90 should at the very least be "The best thing to use is" instead of the literal "The best is to use is" Bro didn't even read my PR.

chrisbra:
Bro, I did read your PR. And I think, when looking for documentation, "the best is to use `:help`" may sound a bit strange at first but is quite right.

TheProdigyLeague:
> Bro, I did read your PR. And I think, when looking for documentation, "the best is to use `:help`" may sound a bit strange at first but is quite right. 

> The best is to use :help
> The best is to use
> The best is to
> The best is

**The best thing to use is :help**

PR #15776, opened 2024-10-01
Wrong display when typing in diff mode with 'smoothscroll'

zeertzjq:
Problem:  Wrong display when typing in diff mode with 'smoothscroll'.
Solution: Use adjust_plines_for_skipcol().


zeertzjq:
This change is similar to the change in [patch 9.0.1512](https://github.com/vim/vim/commit/c8502f9b880b6d23baa4f9d28b60e1ceb442e35f)

chrisbra:
thanks!

PR #15778, opened 2024-10-01
`cedit` raises error for invalid `<xxx>` notation and fix docs

Milly:
## Problem

- `:set cedit=<xxx>` does not raise a error.
- The help for `cedit` is inaccurate.
  - `:set cedit=<Esc>` works, but the example says to use `:exe "set cedit=\<Esc>"`.
  - `:set cedit=^F` works, but the help does not mention that.

## Expected behaviour

When executing `:set cedit=<xxx>`, the error `E474: Invalid argument` occurs.


chrisbra:
thanks!

PR #15779, opened 2024-10-01
docs: add `usr` tag to usr_toc.txt

dundargoc:
When typing `:h usr` it redirects to usr_01.txt, but I'd argue
usr_toc.txt is more useful as you can see an overview of all manuals.
When I usr `:h usr` I personally always intend to go to `usr_toc`.


chrisbra:
Thanks, makes sense. I'll update the tags file as well :)

PR #15780, opened 2024-10-01
ci.yml: macos-13 as macos-12 will be dropped

pheiduck:
See: https://github.com/actions/runner-images/issues/10721


chrisbra:
thanks, for noticing those changes!

PR #15784, opened 2024-10-02
runtime:ftplugin:systemd: small fixes to &keywordprg

Konfekt:
Vim's  `:term` does not understand initial env. var's and slashes need to be escaped.
Question: Isn't `:term ++close` preferable?


chrisbra:
Yeah, I think `++close` is preferable. Let me add this while merging. Thanks!

PR #15786, opened 2024-10-03
fix: missing replace of basic_width

glepnir:


chrisbra:
thanks!

glepnir:
seems like a typo in commit ? should be `from patch v9.1.0755` ?

PR #15787, opened 2024-10-03
set invalid key to 'wildcharm' will result in 0 without error

Milly:
## Problem

`:set wildcharm=<xxx>` does not throw an error, and set to `0`.
There are not enough test patterns for `wildchar` or `wildcharm`.

## Detail

`:set wildcharm=0` initializes it, I think it should not ignore errors.
This removed condition was implemented around Vim 7.0. I don't know the details.


chrisbra:
thanks!

PR #15790, opened 2024-10-03
Constrain gitcommitHash length to 40 or 64

fionn:
Git commit hashes are either 40 or 64 characters long (for SHA-1 or SHA-256 respectively). The previous implementation matches all hexadecimal strings of 40 or more characters, which is too eager.


zeertzjq:
cc @tpope

tpope:
Applied upstream at https://github.com/tpope/vim-git/commit/485e38f29494e46a9998100e707c755d114155f1.

chrisbra:
closing here then. @tpope please send an update when you feel is necessary. Thanks all

GGOemea (commented):
Ok

PR #15796, opened 2024-10-03
runtime(java): Manage circularity for every :syn-included syntax file

zzzyxwvut:
With `g:markdown_fenced_languages` defined and `java` added  
to its list, a circular dependency between the Markdown and  
Java syntax files will be made.  To break it, no Markdown  
documentation comments will be recognised in fenced blocks  
in Markdown files; in order to view Java source files,  
`java` must be removed from `g:markdown_fenced_languages`,  
and this task can be automated as follows.

1) Add to `~/.after/ftplugin/java.vim`:

```vim
if exists("g:markdown_fenced_languages") &&
	\ !(exists("g:java_ignore_javadoc") ||
	\ exists("g:java_ignore_markdown"))
    let s:idx = index(g:markdown_fenced_languages, 'java')
    if s:idx > -1
	call remove(g:markdown_fenced_languages, s:idx)
    endif
    unlet s:idx
endif
```

2) Optionally add to `~/.after/ftplugin/markdown.vim`:

```vim
if exists("g:markdown_fenced_languages") &&
	\ index(g:markdown_fenced_languages, 'java') < 0
    call add(g:markdown_fenced_languages, 'java')
endif
```

(Make sure that the above snippets appear in the files under  
the `ftplugin` NOT `syntax` directory.)

Finally, unless the new version of the syntax file is made  
available from `$VIMRUNTIME` (and from `~/.vim/syntax` if  
necessary), OTHER discoverable file versions will be used  
whose behaviour may interfere with this fix.


chrisbra:
thanks!

Issue #7536, opened 2020-12-23
Vim's `:compiler gcc` has an issue with `make: ***`


refaelsh:
It seems to be that Vim's `:compiler gcc` has a bug. It treats `make: ***` as an error and therefor opens an empty buffer named build.make when opening quickfix. Here is a [screenshot](https://i.stack.imgur.com/9ZDVh.png). And [here](https://pastebin.com/VudqY3qg) it is in text form.
It looks like a bug to me, but maybe I a m wrong.

bfrg:
Can you post the entire output of your build command in text form?

refaelsh:
Thank you for replying @bfrg. I've updated my original post with a link to the raw text.

bfrg:
These are the problematic lines:
```
make[2]: *** [giter/CMakeFiles/giter.dir/build.make:284: giter/giter] Error 1
make[1]: *** [CMakeFiles/Makefile2:161: giter/CMakeFiles/giter.dir/all] Error 2
make: *** [Makefile:103: all] Error 2
```

You need to add the following to the `gcc` compiler file:
```vim
CompilerSet errorformat^=
        \%-Eg%\\=make[%\\d%\\+]:\ ***\ %#[%f:%l:\ %.%\\+]\ %m,
        \%-Eg%\\=make:\ ***\ [%f:%l:\ %.%\\+]\ %m,
```
Notice the `^=` which will _prepend_ the items to `errorformat`. You can copy the shipped `gcc` compiler file into your `~/.vim/compiler/gcc.vim` and then just add the three lines below.

It is import to prepend, or otherwise one of the other formats will match first. For example, with your current `errorformat`, Vim matches the line `make: *** [Makefile:103: all] Error 2` with `%f:%l: %m`. This results in a file name `make: *** [Makefile` which doesn't make sense. 

Notice the `%-E` which will remove the lines from the quickfix list. If you want to include them, just remove the minus signs in `%-E`:
```vim
CompilerSet errorformat^=
        \%Eg%\\=make[%\\d%\\+]:\ ***\ %#[%f:%l:\ %.%\\+]\ %m,
        \%Eg%\\=make:\ ***\ [%f:%l:\ %.%\\+]\ %m,
```
But since you're using a CMake-generated Makefile, you probably don't care about the Makefiles.

refaelsh:
Thank you for the reply. The workaround you suggest will indeed work (i tried something very similar from a stack overflow post).
[...]

PR #15798, opened 2024-10-04
unset compiler

Konfekt:
just fostering (hopefully) best practices

### DIFF
--- a/runtime/ftplugin/hare.vim
+++ b/runtime/ftplugin/hare.vim
@@ -51,7 +51,7 @@ augroup hare.vim
 augroup END
 
 if !exists('current_compiler')
-  let b:undo_ftplugin ..= "| compiler make"
+  let b:undo_ftplugin .= "| compiler make"
   compiler hare
 endif
 
--- a/runtime/ftplugin/hare.vim
+++ b/runtime/ftplugin/hare.vim
@@ -1,7 +1,7 @@
 " Vim filetype plugin.
 " Language:     Hare
 " Maintainer:   Amelia Clarke <selene@perilune.dev>
-" Last Updated: 2024-05-10
+" Last Updated: 2024-10-04
 " Upstream:     https://git.sr.ht/~sircmpwn/hare.vim
 
 if exists('b:did_ftplugin')
@@ -11,9 +11,6 @@ let b:did_ftplugin = 1
 
 let s:cpo_save = &cpo
 set cpo&vim
-
-" Set the default compiler.
-compiler hare
 
 " Formatting settings.
 setlocal comments=://
@@ -52,6 +49,11 @@ augroup hare.vim
     autocmd InsertLeave * hi link hareSpaceError Error
   endif
 augroup END
+
+if !exists('current_compiler')
+  let b:undo_ftplugin ..= "| compiler make"
+  compiler hare
+endif
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
--- a/runtime/ftplugin/zig.vim
+++ b/runtime/ftplugin/zig.vim
@@ -1,7 +1,7 @@
 " Vim filetype plugin file
 " Language:     Zig
 " Maintainer:   Mathias Lindgren <math.lindgren@gmail.com>
-" Last Change:  2024 May 21
+" Last Change:  2024 Oct 04
 " Based on:     https://github.com/ziglang/zig.vim
 
 if exists("b:did_ftplugin")
@@ -13,14 +13,12 @@ let b:did_ftplugin = 1
 let s:cpo_save = &cpo
 set cpo&vim
 
-compiler zig_build
-
 " Match Zig builtin fns
 setlocal iskeyword+=@-@
 setlocal formatoptions-=t formatoptions+=croql
 setlocal suffixesadd=.zig,.zir,.zon
 let &l:define='\v(<fn>|<const>|<var>|^\s*\#\s*define)'
-let b:undo_ftplugin = 'setl isk< fo< sua< mp< def<'
+let b:undo_ftplugin ..= 'setl isk< fo< sua< mp< def<'
 
 if get(g:, 'zig_recommended_style', 1)
     setlocal expandtab
@@ -45,6 +43,11 @@ endif
 if exists('g:zig_std_dir')
     let &l:path .= ',' . g:zig_std_dir
     let b:undo_ftplugin .= ' | setl pa<'
+endif
+
+if !exists('current_compiler')
+    compiler zig_build
+    let b:undo_ftplugin .= "| compiler make"
 endif
 
 let &cpo = s:cpo_save
--- a/runtime/ftplugin/zig.vim
+++ b/runtime/ftplugin/zig.vim
@@ -18,7 +18,7 @@ setlocal iskeyword+=@-@
 setlocal formatoptions-=t formatoptions+=croql
 setlocal suffixesadd=.zig,.zir,.zon
 let &l:define='\v(<fn>|<const>|<var>|^\s*\#\s*define)'
-let b:undo_ftplugin ..= 'setl isk< fo< sua< mp< def<'
+let b:undo_ftplugin = 'setl isk< fo< sua< mp< def<'
 
 if get(g:, 'zig_recommended_style', 1)
     setlocal expandtab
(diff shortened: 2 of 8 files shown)


bfrg:
Shouldn't the `make.vim` compiler reset the options with a `<`?

```vim
CompilerSet makeprg<
CompilerSet errorformat<
```

All other buffer-local options are reset in `b:undo_ftplugin` like that.

The `set {option}<` notation is explained under `:h :setlocal`:
https://github.com/vim/vim/blob/075ab5ab3b8abc55baddce7105e22b6fa1b2e032/runtime/doc/options.txt#L370-L384

Konfekt:
The `:CompilerSet` is somewhat mysterious to me, as it looks user defined but rather originates from src/ex_cmds2.c, src/usercmd.c, src/ex_cmds.h.

However `:help current_compiler` says that

```
...
To support older Vim versions, the plugins always use "current_compiler" and
not "b:current_compiler".  What the command actually does is the following:

- Delete the "current_compiler" and "b:current_compiler" variables.
- Define the "CompilerSet" user command.  With "!" it does ":set", without "!"
  it does ":setlocal".
```

So unbanged `:CompilerSet` is setlocal.

Konfekt:
Oh, I think that was not actually the question. Rather if one expects `:compiler make` to reset to the user's global value or Vim's predefined value? I'd guess rather the latter, just by Vim's predefined value seemingly fit to Gnu make.

Konfekt:
But then again you have a point that here `:compiler make` does not achieve unsetting the local to the global values, as happens for the rest of the settings... that's an inconsistency.
So you propose something akin to `:setlocal makeprg< errorformat< | unlet b:compiler` instead?

bfrg:
All other options like `commentstring` are reset in `b:undo_ftplugin` with a `<`, so why not `makeprg` and `errorformat`?

bfrg:
I assumed `compiler/make.vim` was added just for resetting the `makeprg` and `errorformat` options in `b:undo_ftplugin`.

Konfekt:
How about a 

```
command UnCompilerSet exe <bang>0 ?  
	'setlocal makeprg& errorformat& | silent! unlet b:compiler g:compiler' : 
	'setlocal makeprg< errorformat< | silent! unlet b:compiler'
```

to unset the compiler?

Indeed an erroneous assumption was that `:compiler make` could also serve to unset `:compiler`, but as you noticed that's inconsistent. But initially I added it to reset to `makeprg&` and `errorformat&` gracefully.

runtime/ftplugin/zig.vim
```diff
@@ -13,14 +13,12 @@ let b:did_ftplugin = 1
...
 
-compiler zig_build
-
 " Match Zig builtin fns
 setlocal iskeyword+=@-@
 setlocal formatoptions-=t formatoptions+=croql
 setlocal suffixesadd=.zig,.zir,.zon
 let &l:define='\v(<fn>|<const>|<var>|^\s*\#\s*define)'
-let b:undo_ftplugin = 'setl isk< fo< sua< mp< def<'
+let b:undo_ftplugin ..= 'setl isk< fo< sua< mp< def<'
```
chrisbra: that looks wrong

runtime/ftplugin/hare.vim
```diff
@@ -53,6 +50,11 @@ augroup hare.vim
   endif
 augroup END
 
+if !exists('current_compiler')
+  let b:undo_ftplugin ..= "| compiler make"
```
chrisbra:
```suggestion
  let b:undo_ftplugin .= "| compiler make"
```
to be consistent with the style a few lines above

chrisbra (changes requested):

runtime/ftplugin/zig.vim
Konfekt: Thank you

runtime/ftplugin/hare.vim
Konfekt: Will do. (A few lines further above it's `..` again.)

Konfekt:
@chrisbra Mind that `:compiler make` uses `&` instead of `<` for `errorformat` and `makeprg` different from what `b:undo _ftplugin` does for all other options, as @bfrg remarked. But I think an additional `:compiler global` would be an abuse of `:compiler`

chrisbra:
Thanks. Yes I agree, `compiler global` is not needed

Konfekt:
@bfrg (This discussion on the default makeprg](https://old.reddit.com/r/vim/comments/1g5hf2l/whenever_the_internal_make_command_raises_an/) had me have a look again. I know see that `&errorformat` is exactly that of `compiler/gcc.vim`, but somehow `&makprg` is `make` instead of `gcc`.
More coherent would be either `&makeprg=gcc`, or `&errorformat` that of `make`, for which I don't know of any `compiler/ file.vim` setting it suitably, though.

In any event, this made it apparent that `make` does not really set the `&errorformat` of make, so I feel more inclined to abusing it to let `:compiler make` set `makeprg<` and `errorformat<` instead, and only `:compiler! make` then the default values `makeprg/errorformat&`. However, how to achieve the latter, that is, distinguish inside a `compiler/file.vim` if it has been sourced with a bang? Here `:help write-compiler-plugin` did not make it evident at least to me. 

On a sidenote, @chrisbra, should `compiler/gcc.vim` not then rather `set errorformat&` as it stands? If gcc's output changes, one could still document so in `compiler/gcc.vim`.

Konfekt:
Oh, I see that @bfrg  already suggested the same [four years ago](https://github.com/vim/vim/issues/7536#issuecomment-751519132)

Konfekt:
I stand corrected, @gpanders included filtering of make errors on Linux/Mac OS? since [9.0.1613](https://github.com/vim/vim/commit/d1911a8e2b1498f0cb0275a98f63dd212204a1a8). 

Still, maybe this mix could be documented (`&errorformat` by default parsing gcc with some parts of Gnu make on Unix systems.) I could not find anything in `:help 'errorformat'` or `:help error-file-format`

bfrg:
> I stand corrected, @gpanders included filtering of make errors on Linux/Mac OS? since [9.0.1613](https://github.com/vim/vim/commit/d1911a8e2b1498f0cb0275a98f63dd212204a1a8).

And now `make` errors are completely ignored. Not sure if that's a good idea.

Anyway, the issue is that `make` can call _any_ program. We don't know what someone uses `make` for, could be C, C++, Python, Java etc. So technically, you have to prepend the errorformat for catching Makefile errors to every compiler file.

Konfekt:
> And now make errors are completely ignored. Not sure if that's a good idea.

Yes, why completley ignore them? I thought `&errorformat` of make serves to populate the quickfix/location list with the locations of the errors spit out by `make`.

Instead, those of `make` are now forcibly ignored, and only those of `gcc` shown.
I find this rather confusing.

I'd propose to instead set `&errorformat` so that it populates the location list with the locations of make.

Then to append errorformats of compilers like `gcc` to that.  

Now either

- for sake of backwards compatibility, one could leave &errorformat set to that of `gcc`,
- or have a breaking change to set it to that of Gnu Make, as people likely expect when they read that `&makeprg` is make by default

gpanders:
`'errorformat'` dictates which lines are considered errors for use with quickfix commands like `:cnext`, `:cc`, etc. The line in the Makefile in which a source file containing an error is compiled is not itself an error, so it should not be included in the error list. If the Makefile *itself* contains an error (eg a syntax error) then that should be included in the error list, but those errors have a different format.

And the Makefile messages are not “completely ignored”, they are still present in the quickfix list after using `:make`, they are just not highlighted as errors or included in the error list.

The problem description in https://github.com/vim/vim/pull/12481 has additional context on why including the Makefile errors is a bad default.

bfrg:
> The line in the Makefile in which a source file containing an error is compiled is not itself an error, so it should not be included in the error list.

I can understand why you (and probably others) don't want that error in the quickfix list. But there are occasions where it's actually helpful. Consider the following simple Makefile:
```make
app: main.c
	$(CC) main.c -o app -Wall -Wextra -pedantic -std=c9
```
Instead of `-std=c99` I  wrote `-std=c9`. Running `make` will give you the following error:

```
cc main.c -o app -Wall -Wextra -pedantic -std=c9
cc: error: unrecognized command-line option ‘-std=c9’; did you mean ‘-std=c89’?
make: *** [Makefile:2: app] Error 1
```

While `gcc` gives you a clear message, it doesn't help you to find the line in the Makefile where you made the typo. Having the above `make` error in the quickfix list can very helpful, especially for larger Makefiles.

Konfekt:
I wonder if a compiler could be set similar to `&spell` by a `&compiler` option by comma separated values, where the first one defines the `&makeprg`, for example `set compiler=make,gcc`.
As it stands, one ends up writing one compiler file for each pair of compiler, say `javac`, and build tool, say `make`.

PR #15800, opened 2024-10-04
runtime(doc,syntax): help formatting and add syntax

Milly:
- Apply `helpExample`.
- Remove `helpHeader` from URL, because `helpURL` takes priority.
- Fix indentation.
- Add `N` emphasis to `helpSpecial`.

### DIFF
--- a/runtime/doc/cmdline.txt
+++ b/runtime/doc/cmdline.txt
@@ -481,13 +481,13 @@ The 'wildignorecase' option can be set to ignore case in filenames.  For
 completing other texts (e.g. command names), the 'ignorecase' option is used
 instead (fuzzy matching always ignores case, however).
 
-If you like tcsh's autolist completion, you can use this mapping:
+If you like tcsh's autolist completion, you can use this mapping: >
 	:cnoremap X <C-L><C-D>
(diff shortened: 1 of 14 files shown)


chrisbra:
Thanks. I had to guess what this is for:
```
syn match helpSpecial		"^\t[enfLlbghNEpticCuUwWkmMjJP]N "ms=s+2,me=e-1
```
It seems this is for `:h cinoptions-values`? If that is the case, it seems there are a few flags missing and it rather should be this?
```
syn match helpSpecial           "^\t[enfLlbghNEpticCuUwWkmMjJP>}{^:=+/)#]N "ms=s+2,me=e-1
```

Milly:
> ```
> syn match helpSpecial           "^\t[enfLlbghNEpticCuUwWkmMjJP>}{^:=+/)#]N "ms=s+2,me=e-1
> ```

`{non-alphanumeric-symbol}N` are not needed as they are already matched:
```vim
syn match helpSpecial		"\<N\>"
```

chrisbra:
Well, the highlighting is off for those characters

Milly:
I understand that the results vary depending on `'iskeyword'`.

PR #15803, opened 2024-10-04
Problem:  screenpos() may return invalid position

### LINKED ISSUE #15792: `screenpos()` returns all zeros under specific circumstances
### Steps to reproduce

Originally reported in the Neovim repo here:

- https://github.com/neovim/neovim/issues/30639

but I think it's actually common to both Neovim and Vim.

### Problem

`screenpos()` _sometimes_ returns all zeros; ie:

```
{'col': 0, 'row': 0, 'endcol': 0, 'curscol': 0}
```

instead of the actual screen coordinates, causing plug-ins like [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) to draw their autocomplete pop-up menu in the top-left of the screen instead of at the cursor location:

autocomplete menu appearing in the wrong place

This was originally reported in:

- https://github.com/neovim/neovim/issues/19606

but without a consistent repro. I've been seeing this happen intermittently for years and have never been able to find a reliable repro recipe, until now.

### Steps

This one is incredibly sensitive to initial conditions (screen size, font size, path lengths), and rather fiddly to reproduce. Nevertheless, I've found that I can trigger the bug from a clean Neovim install by interacting with the [vim-dirvish](https://github.com/justinmk/vim-dirvish) plug-in and following a very precise set of steps. (I don't think this is a bug with vim-dirvish or with nvim-cmp, but they are useful to illustrate the bug.)

1. Size your terminal window such that it is 137 columns wide. Verify this with:
   ```
   tput cols
   ```
2. Create a new scratch directory in which to reproduce the bug:
   ```
   mkdir /tmp/scratch
   ```
   Note that the name of the path matters here because the path must be of a certain length to trigger the bug.
3. Change into the scratch directory:
   ```
   cd /tmp/scratch
   ```
4. Create directories in which to put the repro files:
   ```
   mkdir -p aspects/nvim/files/.config/nvim/plugin/mappings
   ```
   There's nothing special about the _contents_ of this path (this is just a path from the project I was in when I discovered how to repro the bug) but the _length_ of the path does matter.
5. Create repro files (again, contents don't matter, so `touch` will suffice and, again, the names are just take from the project I was in and its the length that matters):
   ```
   touch aspects/nvim/files/.config/nvim/plugin/mappings/normal.lua
   touch aspects/nvim/files/.config/nvim/plugin/mappings/visual.lua
   ```
6. Grab a copy of vim-dirvish:
   ```
   mkdir -p pack/bundle/opt
   git clone https://github.com/justinmk/vim-dirvish.git pack/bundle/opt/vim-dirvish
   ```
7. Launch Neovim with no config:
   ```
   nvim -u NONE
   ```
8. Load vim-dirvish:
   ```
   set packpath=.
   packadd vim-dirvish
   packloadall
   ```
9. Create a mapping so that you can inspect the result of `screenpos()`:
   ```
   :nmap <c-s> :echo screenpos(0, 1, 1)<cr>
   ```
10. Open the first of the repro files:
    ```
    :e aspects/nvim/files/.config/nvim/plugin/mappings/normal.lua
    ```
11. Create a vertical split:
    ```
    :vs
    ```
12. Move focus to window on right side of the split:
    ```
    <C-w l>
    ```
13. Invoke vim-dirvish to show the contents of the containing directory:
    ```
    :Dirvish
    ```
14. Hit `j` to move down to the sibling file (`visual.lua`).
15. Hit `Enter` to open the file.
16. Hit the `<C-s>` binding you created previously to display the `screenpos()` result; note that it's all zeros:
    ```
    {'col': 0, 'row': 0, 'endcol': 0, 'curscol': 0}
    ```

Additional notes:

- This repros both inside tmux (`TERM=tmux-256color`) and outside (`TERM=xterm-kitty`).
- As noted, it also affects Neovim.
- As I mentioned above, this whole thing is very sensitive to path lengths and cursor position:
  - In the vim-dirvish listing, move the cursor to the end (`$`) before opening the second file, and the bug repros.
  - Move it to the beginning (`^`) before opening the second file, and it does not repro.
- In common with [the original report](https://github.com/neovim/neovim/issues/19606), after you've seen the bug, running `:set wrap` (even if it was already set) temporarily fixes the problem and hitting the `<C-s>` mapping shows that `screenpos()` will return sensible results again.
- Font size matters, because it affects the number of characters that appear before the line is too wide to show on the screen. For example, if I make my font small enough so that everything fits, this no longer repros. For reference, this screenshot shows my font size when reproing the issue: font size reference

### Expected behaviour

`screenpos()` should return non-zero values, such as:

```
{'col': 5, 'row': 1, 'endcol': 5, 'curscol': 5}
```

### Version of Vim

9.1.0750

### Environment

- Operating system: macOS 14.7
- kitty terminal: v0.36.4
- `TERM=tmux-256color` (inside tmux) and `TERM=xterm-kitty` (outside tmux)

### Logs and stack traces

_No response_

chrisbra:
Problem:  screenpos() may return invalid position
Solution: reset w_leftcol if wrapping has been set

fixes: #15792

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -6736,6 +6736,11 @@ win_copy_options(win_T *wp_from, win_T *wp_to)
     void
 after_copy_winopt(win_T *wp)
 {
+    // Set w_leftcol or w_skipcol to zero.
+    if (wp->w_p_wrap)
+	wp->w_leftcol = 0;
+    else
+	wp->w_skipcol = 0;
 #ifdef FEAT_LINEBREAK
     briopt_check(wp);
 #endif
--- a/src/testdir/test_cursor_func.vim
+++ b/src/testdir/test_cursor_func.vim
@@ -279,6 +279,21 @@ func Test_screenpos_number()
   bwipe!
 endfunc
 
+func Test_screenpos_edit_newfile()
+  new
+  20vsp
+  setl nowrap
+  call setline(1, 'abcdefghijklmnopqrstuvwxyz')
+  call cursor(1, 10)
+  norm! 5zl
+  call assert_equal(#{col: 5, row: 1, endcol: 5, curscol: 5}, screenpos(win_getid(), 1, 10))
+  enew!
+  call assert_equal(1, &l:wrap)
+  call assert_equal(#{col: 1, row: 1, endcol: 1, curscol: 1}, screenpos(win_getid(), 1, 1))
+
+  bwipe!
+endfunc
+
 " Save the visual start character position
 func SaveVisualStartCharPos()
   call add(g:VisualStartPos, getcharpos('v'))


zeertzjq (commented):
Setting `w_leftcol` in these two places is a bit random. How about doing this in `after_copy_winopt()` instead:
```diff
diff --git a/src/option.c b/src/option.c
index ee9502a2e..2dd1afdbf 100644
--- a/src/option.c
+++ b/src/option.c
@@ -6736,6 +6736,10 @@ win_copy_options(win_T *wp_from, win_T *wp_to)
     void
 after_copy_winopt(win_T *wp)
 {
+    if (wp->w_p_wrap)
+	wp->w_leftcol = 0;
+    else
+	wp->w_skipcol = 0;
 #ifdef FEAT_LINEBREAK
     briopt_check(wp);
 #endif
```

chrisbra:
yeah, you are right. I went with `curs_columns()` as the comment mentions this updates `w_leftcol` but for the test, I believe `curs_columns()` wasn't yet called. Let me change it like this (I think we can just use `did_set_wrap()` instead).

src/option.c
```diff
@@ -6736,6 +6736,8 @@ win_copy_options(win_T *wp_from, win_T *wp_to)
     void
 after_copy_winopt(win_T *wp)
 {
+    // make sure w_leftcol is correct
+    (void)did_set_wrap(NULL);
```
zeertzjq: This doesn't use the `wp` passed in (`did_set_wrap()` uses `curwin`) while the calls below all use it.

chrisbra: ah thanks, obviously right

PR #15804, opened 2024-10-05
runtime:compiler:add cppcheck C(++) linter

Konfekt:
Seems well maintained and on first sight there are not so many already included

### DIFF
--- a/runtime/doc/quickfix.txt
+++ b/runtime/doc/quickfix.txt
@@ -1294,10 +1294,11 @@ settings by default include
 For C++ files (`filetype == 'cpp'`), the `--language=c++` option is added to
 ensure Cppcheck treats the file as C++.
 
-By default the directories in &path are passed as include directories.
-These can be set by g/b:c_cppcheck_includes as a list of '-I' flags.
-Tim Pope's vim-apathy plug-in [0] can expand &path. To also append the folders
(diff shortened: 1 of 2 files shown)


runtime/doc/quickfix.txt
```diff
@@ -1278,6 +1278,31 @@ For writing a compiler plugin, see |write-compiler-plugin|.
...
+- `--enable=...`: Enables specific checks like warnings, style, performance,
+  portability, information, and missing includes.
+- `-j`: Utilizes multiple processors if available, determined by the
+  `getconf` command if available (requires omitting the unusedFunction check)
+
+For C++ files (`filetype == 'cpp'`), the `--language=c++` option is added to
+ensure Cppcheck treats the file as C++.
+
+By default the directories in &path are passed as include directories.
+These can be set by g/b:c_cppcheck_includes as a list of '-I' flags.
```
DominiquePelle-TomTom: Finding includes this way is almost unusable for large projects with many include paths.

It would be better to support include paths from `compile_commands.json` (already used by many other tools such as clangd, clang-tidy, include-what-you-use).  I see that cppcheck can use this too with `cppcheck --project=compile_commands.json`

chrisbra:
thanks

Issue #15797, opened 2024-10-04
Comment toggling does not work correctly in files that use mixed tab and space indentation


julio-b:
### Steps to reproduce

1. `vim --clean -c 'packadd comment'`
2. Edit the following c file:
```c
int main(void) {
    if (1) {
	int q;
	printf("");
	return q;
    }
}
```
3. Jump to line 4 : `4G`
4. `gca{`

Result:
```c
int main(void) {
    /* if (1) { */
    /*  q; */
    /* ntf(""); */
    /* urn q; */
    /* } */
}
```

Notice how lines 3,5 are now messed up.

### Expected behaviour

Comment toggling should not change the indentation or modify the code in any way.
```c
int main(void) {
    /* if (1) { */
	/* int q; */
	/* printf(""); */
	/* return q; */
    /* } */
}
```

### Version of Vim

v9.1.0757

### Environment

OS: Archlinux
Shell: `bash 5.2.32`
Terminal: `kitty`
$TERM: `xterm-kitty`

chrisbra:
oh, that is nasty. @habamax ?

habamax:
Will have a look next week. 

This is quite strange, I have used it quite a lot with vim’s source code which has mixed indentation and it never did this.

habamax:
While I am away, could anyone check with the original comment pack before it was changed by @Konfekt ?

Konfekt:
Since these are surrounding comment markers, the added if-block is not entered:

https://github.com/vim/vim/blob/40c6babc1789aceb241b23bab76eea16da37e33d/runtime/pack/dist/opt/comment/autoload/comment.vim#L26

However, `strlen()` in

```vim
# pack/dist/opt/comment/autoload/comment.vim (lines 62-63)
line = printf(indent_start .. substitute(cms, '%s\@!', '%%', 'g'),
        strpart(getline(lnum), strlen(indent_start)))
```

is too aggressive here as for`2gcc` on line 2  we've `strlen(indent_start) = 4` by the initial indent of 4 spaces, but in the following line there are only two tab characters.
Here's a try to remove only one tab for every multiple of `&tabstop` many spaces:

```vim
                line = getline(lnum)
                var ind_len = strlen(matchstr(line, '^\s*'))
                line = printf(indent_start .. substitute(cms, '%s\@!', '%%', 'g'),
                        strpart(line, (line[0 : strlen(indent_start) - 1] =~ '\t' ? ind_len / &tabstop : ind_len)))
```

[...]

PR #15805, opened 2024-10-05
runtime:comment: consider &tabstop in lines after whitespace indent

Konfekt:
The count `strlen()` in

```vim
line = printf(indent_start .. substitute(cms, '%s\@!', '%%', 'g'),
                       strpart(getline(lnum), strlen(indent_start)))
```

is too high here as for`2gcc` on line 2  we've `strlen(indent_start) = 4` for an initial indent of 4 spaces, but in the following line there is only a single tab character if they happen to be tabs with &tabstop = 8.

The check is still too simple because it assumes that as soon as there's a tab the whole indent is made of tabs; it's a start of entering the mixed tab and whitespace indent rabbit hole.

Addresses [0]

Links:

[0]: https://github.com/vim/vim/issues/15797


chrisbra:
thanks

PR #15809, opened 2024-10-05
Refactor ex_getln.c to remove calls to STRLEN()

basilisk0315:
Also add dealloc_cmdbuf() to keep ccline.cmdlen and ccline.cmdbuflen in sync with ccline.cmdbuff.

### DIFF
--- a/src/ex_getln.c
+++ b/src/ex_getln.c
@@ -45,6 +45,7 @@ static int	cmdline_charsize(int idx);
 static void	set_cmdspos(void);
 static void	set_cmdspos_cursor(void);
 static void	correct_cmdspos(int idx, int cells);
+static void	dealloc_cmdbuff(void);
 static void	alloc_cmdbuff(int len);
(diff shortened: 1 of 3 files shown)


chrisbra:
thanks!

src/ex_getln.c
```diff
@@ -4684,21 +4729,30 @@ open_cmdwin(void)
...
-		ccline.cmdbuff = vim_strsave((char_u *)p);
+		ccline.cmdbuff = vim_strnsave(p, plen);
+		ccline.cmdlen = plen;
+		ccline.cmdbufflen = plen + 1;
 		cmdwin_result = CAR;
 	    }
 	    else
 	    {
 		// First need to cancel what we were doing.
-		ccline.cmdbuff = NULL;
```
zeertzjq, line 4701: Why is this line removed?

zeertzjq: Oh I see. `dealloc_cmdbuff()` was already called above.

PR #15811, opened 2024-10-06
`cedit`, `termwinkey` and `wildchar` are not parsed correctly

Milly:
- Problem: `^@` raises an error.
  Solution: Store as `<Nul>`.
- Problem: `<t_xx` does not raise an error.
  Solution: Raises an error if closing `>` is missing.
- Problem: Single `<` or `^` raises an error. It is inconvenient for users.
  Solution: They are stored as a single character.


src/option.c
```diff
@@ -5622,7 +5627,10 @@ find_key_option(char_u *arg_arg, int has_lt)
     // Don't use get_special_key_code() for t_xx, we don't want it to call
     // add_termcap_entry().
     if (arg[0] == 't' && arg[1] == '_' && arg[2] && arg[3])
-	key = TERMCAP2KEY(arg[2], arg[3]);
+    {
+	if (!has_lt || arg[4] == '>')
```
chrisbra, line 5631: Hm, I think this should be: 
```suggestion
	if (!has_lt || (arg[4] && arg[4] == '>'))
```

zeertzjq: Why check the value of `arg[4]` twice?

chrisbra: ah you are right, I wanted to be careful and only access it if it is not null (as in the if condition above). But it seems may not be necessary 🤷

PR #15814, opened 2024-10-06
correct `vi` registers 1-9 documentation error

fkabell3:
When using heirloom vi originally written by Bill Joy (`:version` gives me "Version 4.0 (gritter) 12/25/06"), I am able to store text into registers 1-9 and subsequently use the `:edit` or `:next` commands to change files and paste the contents of those numbered registers, contrary to what Vim documentation states.


chrisbra:
Hm, perhaps this was a bug in one of the original vis. I checked Posix documentation for [buffers](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/ex.html#tag_20_40_13_10) and it does not mention such a restriction, so should be fine.

PR #15815, opened 2024-10-07
No test for patches 6.2.418 and 7.3.489

zeertzjq:
Problem:  No test for patches 6.2.418 and 7.3.489.
Solution: Add a test.  Fix some whitespace problems in test_mapping.vim.
          Document the behavior.


chrisbra:
thanks

PR #15818, opened 2024-10-07
refactor(completion): using abbr_hlgroup replace hlgroup in complete

glepnir:
Problem:
Currently hlgroup combines on all items, which is redundant and will cause some unnecessary calculations. And hlgroup and kind_hlgroup may cause confusion in usage.

Solution:
Renamed to abbr_hlgroup, only perform superposition on abbr.


zeertzjq:
Supersedes #15994?

glepnir:
I forgot about that draft. closed

chrisbra:
Thanks. This is mainly about renaming the name from `hlgroup` to `abbr_group`?

glepnir:
Previously hlgroup combined all item attrs. This was unnecessary for menu kind etc. Changed to `abbr_hlgroup` to be consistent with `kind_hlgroup` and only combine on abbr text. Still uses original attr for space padding.

src/popupmenu.c
```diff
@@ -823,7 +833,7 @@ pum_redraw(void)
 #endif
 	    {
 		screen_fill(row, row + 1, col, pum_col + basic_width + n,
-							      ' ', ' ', attr);
+							      ' ', ' ', orig_attr);
```
zeertzjq, line 836: Missing `attr -> orig_attr` change in the `pum_rl` branch above

src/popupmenu.c
```diff
@@ -781,13 +790,14 @@ pum_redraw(void)
...
-									attr);
+								    orig_attr);
 			    col -= 2;
 			}
 			else
 #endif
 			{
-			    screen_puts_len((char_u *)"  ", 2, row, col, attr);
+			    screen_puts_len((char_u *)"  ", 2, row, col,
+								    orig_attr);
```
zeertzjq, line 800: This change seems wrong. `attr` should be used here, not `orig_attr`

src/structs.h
```diff
@@ -4468,14 +4468,14 @@ typedef struct
...
-    int		pum_idx;	  // index of item before sorting by score
-    int		pum_user_hlattr;  // highlight attribute to combine with
-    int		pum_user_kind_hlattr; // highlight attribute for kind
+    char_u	*pum_text;		// main menu text
+    char_u	*pum_kind;		// extra kind text (may be truncated)
+    char_u	*pum_extra;		// extra menu text (may be truncated)
+    char_u	*pum_info;		// extra info
+    int		pum_score;		// fuzzy match score
+    int		pum_idx;		// index of item before sorting by score
+    int		pum_user_abbr_hlattr;	// highlight attribute to combine with
```
zeertzjq, line 4477: This comment may need change

src/popupmenu.c
```diff
@@ -771,7 +776,11 @@ pum_redraw(void)
 			    col += width;
 			}
 
-			vim_free(attrs);
+			if (attrs != NULL)
+			{
+			    vim_free(attrs);
+			    attrs = NULL;
+			}
```
zeertzjq, line 783: `VIM_CLEAR()` can be used here

src/insexpand.c
```diff
@@ -100,13 +100,14 @@ struct compl_S
...
-    int		cp_user_hlattr;	// highlight attribute to combine with
-    int		cp_user_kind_hlattr; // highlight attribute for kind
+    char_u	*cp_fname;		// file containing the match, allocated when
+					// cp_flags has CP_FREE_FNAME
+    int		cp_flags;		// CP_ values
+    int		cp_number;		// sequence number
+    int		cp_score;		// fuzzy match score
+    int		cp_user_abbr_hlattr;	// highlight attribute to combine with
+					// for abbr.
+    int		cp_user_kind_hlattr;	// highlight attribute for kind
```
zeertzjq, line 110: There two comments are a bit inconsistent

src/popupmenu.c
glepnir: purpose of abbr_hlgroup is to overlay some additional gui attributes such as strikethrough.  the space padding no need combine with new attr like `strikethrough` will looks wired

using orig_attr it will looks like this. 

QQ_1728450230523

zeertzjq: It's not space padding.

zeertzjq: I don't think this is space padding. A TAB is a part of the text.

src/popupmenu.c
glepnir: thanks. will change

src/popupmenu.c
glepnir: yes you are right. i mixed up it

PR #15823, opened 2024-10-07
Improve completion API

kekulta:
I added two features:
1. New completion type for user defined commands `shellcmdline`. It behaves exactly as `:!` which expects it first argument to be `shellcmd` and any subsequent argument to be a file.
2. New function `getcmdcomplpat()` which works in the manner of the `getcmdcomptype()` and exposes current completion pattern from `xpc->xp_pattern`.

Both features are useful for completion plugins and I can find [question about the first one as far as from 2011](https://stackoverflow.com/questions/5685005/shell-command-completion).
I already added some tests and updated docs.

My first attempt was to contribute into nvim, so there is some discussion and explanation in [the issue there](https://github.com/neovim/neovim/issues/30699#issuecomment-2396281502).

All the tests except for the seven are ok. The same seven tests failing in the master branch and they are not related to anything I commited, so it's either broken tests or some problems with my local configuration. Would be really thankful for any advice on this matter.


tomtomjhj:
I think it's fine to change the `shellcmd` completion to behave like the proposed `shellcmdline` completion. It's very unlikely for someone to rely on the old behavior. See https://github.com/search?q=%22complete%3Dshellcmd%22+language%3A%22Vim+Script%22&type=code&l=Vim+Script.

chrisbra:
Hm, there are 1.1k results. Calling this unlikely is a bit brave. Let's keep the shellcmdline completion type

chrisbra:
Note that there are also a few test failures

kekulta:
Yes, I'm trying to setup Windows environment to investigate this. For some reason tab completion works without problems on my Linux machine but breaks on Windows and macos. And only on medium and huge builds. I currently have zero ideas why that happened but I hope I will figure that out soon.

Shougo:
```diff
diff --git a/src/testdir/test_cmdline.vim b/src/testdir/test_cmdline.vim
index 7534e081a..47f74237c 100644
--- a/src/testdir/test_cmdline.vim
+++ b/src/testdir/test_cmdline.vim
@@ -3647,9 +3647,11 @@ func Test_cmdline_complete_shellcmdline_argument()
 
   set wildoptions=fuzzy
   call feedkeys(":MyCmd vim test_cmdline.\<Tab>\<C-B>\"\<CR>", 'xt')
+  sleep 100m
   call assert_equal('"MyCmd vim test_cmdline.vim', @:)
   set wildoptions&
   call feedkeys(":MyCmd vim test_cmdline.\<Tab>\<C-B>\"\<CR>", 'xt')
+  sleep 100m
   call assert_equal('"MyCmd vim test_cmdline.vim', @:)
 
   delcommand MyCmd
```

`:sleep` may help you.
I think the completion seems slow when macOS and Windows.  Because it is from shell command line.

Shougo:
```
	From test_cmdline.vim:
	Found errors in Test_cmdline_complete_shellcmdline_argument():
	command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_cmdline_complete_shellcmdline_argument line 5: Expected '"MyCmd vim test_cmdline.vim' but got '"MyCmd vim test_cmdline.'
	command line..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[607]..function RunTheTest[57]..Test_cmdline_complete_shellcmdline_argument line 8: Expected '"MyCmd vim test_cmdline.vim' but got '"MyCmd vim test_cmdline.'
```

kekulta:
After some fight with MSVC I was able to build it on windows. It seems to be the problem that I put `-nargs=1` into my test function. For some reason MS and apparently MacOS version has this behavior that it won't try to complete another argument if it exceeds `nargs`. Yet this behavior ignored if completion type is set to be `shellcmd`. I think if I change it to `-nargs=+` or something it should pass as expected. I wasn't able to run tests locally so hope they will rerun and pass on another push.

I think it is possible to hack a bit more to achieve the same behavior on completion as `shellcmd` but for me it looks more like bug on the part of the `shellcmd`. We do not provide subsequent completions for any completion type if we explicitly set agruments number to be one. Why would we do that for `shellcmd`?

Also interesting part is why this behavior exists on windows and macos but not on linux builds.

kekulta:
Everything except for one macos test passed. It looks like it flacked. It is not about completion and last time it was ok.

Also is that codecov test mandatory or not? Should I add more tests? For me it looks like I already did all the main scenarios.

kekulta:
It passed but I'm still not sure about codecov.

chrisbra:
I am not so worried about codecov, but can you add a few more examples, like some pattern that does not result in any completion result or without any pattern?

src/testdir/test_cmdline.vim
```diff
@@ -3642,30 +3666,32 @@ func Test_cmdline_complete_bang_cmd_argument()
...
+  call assert_equal("\"let a[6, 7, 'var', 'a']", @:)
   call feedkeys(":quit \<F2>\<C-B>\"\<CR>", "xt")
-  call assert_equal("\"quit [6, 7, '']", @:)
+  call assert_equal("\"quit [6, 7, '', '']", @:)
   call feedkeys(":nosuchcommand \<F2>\<C-B>\"\<CR>", "xt")
-  call assert_equal("\"nosuchcommand [15, 16, '']", @:)
+  call assert_equal("\"nosuchcommand [15, 16, '', '']", @:)
 
-  " Check that getcmdcompltype() doesn't interfere with cmdline completion.
+  " Check that getcmdcompltype() and getcmdcomplpat() doesn't interfere with
```
zeertzjq:
```suggestion
  " Check that getcmdcompltype() and getcmdcomplpat() don't interfere with
```

src/testdir/test_cmdline.vim
```diff
@@ -3631,6 +3641,52 @@ func Test_cmdline_complete_substitute_short()
   endfor
 endfunc
 
+" Test for shellcmdline command argument completion
+func Test_cmdline_complete_shellcmdline_argument()
```
zeertzjq, line 3645: Is a `CheckExecutable vim` needed here?

kekulta: We don't check actual type of the argument we don't complete right now. If we are past the first argument it will suggest only files. So even if there are no vim in the system everything gonna be fine.

I can add it but I can't really see the point here.

zeertzjq: Ah OK

kekulta:
Everything passed. I hope it's ok now.

chrisbra:
thanks

kekulta:
Thanks you guys! That is my first contribution to vim/nvim and that was a pleasure to work with you.

PR #15824, opened 2024-10-07
filetype: LLVM IR files are not recognized

Freed-Wu:
Problem: LLVM IR files are not recognized
Solution: recognize LLVM IR files as llvm filetype

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -169,6 +169,14 @@ export def FTcls()
   endif
 enddef
 
+export def FTll()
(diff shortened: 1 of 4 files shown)


chrisbra:
thanks. How about a test? I had some troubles finding some original lifelines files, but if this is for the lifelines genealogy, then there are some examples here: https://github.com/lifelines/lifelines/blob/master/reports

chrisbra:
Thanks. Please also change the Maintainer line here.

zeertzjq (commented):
Why is the same autocommand repeated twice?

chrisbra:
thanks!

PR #15825, opened 2024-10-07
filetype: openCL files are not recognized

Freed-Wu:
Problem: OpenCL files are not recognized
Solution: recognize OpenCL files as opencl filetype

### DIFF
--- a/runtime/autoload/dist/ft.vim
+++ b/runtime/autoload/dist/ft.vim
@@ -144,6 +144,14 @@ export def FTcfg()
   endif
 enddef
 
+export def FTcl()
+  if getline(1) .. getline(2) .. getline(3) .. getline(4) =~ '/\*'
+    setf opencl
+  else
+    setf lisp
+  endif
+enddef
+
 export def FTcls()
   if exists("g:filetype_cls")
     exe "setf " .. g:filetype_cls
--- a/runtime/ftplugin/opencl.vim
+++ b/runtime/ftplugin/opencl.vim
@@ -1,0 +1,13 @@
+" Vim filetype plugin file
+" Language:	OpenCL
+" Maintainer:	The Vim Project <https://github.com/vim/vim>
+" Last Change:	2024 Oct 22
+" Maintainer:	Wu, Zhenyu <wuzhenyu@ustc.edu>
+
+if exists("b:did_ftplugin") | finish | endif
+let b:did_ftplugin = 1
+
+setlocal comments=sO:*\ -,mO:*\ \ ,exO:*/,s1:/*,mb:*,ex:*/,:///,://
+setlocal commentstring=/*\ %s\ */ define& include&
+
+let b:undo_ftplugin = "setl commentstring< comments<"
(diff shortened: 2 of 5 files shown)


runtime/syntax/opencl.vim
```diff
@@ -0,0 +1,11 @@
+" Vim syntax file
+" Language:	C
+
+if exists("b:current_syntax")
+  finish
+endif
+
+" TODO: support openCL specific keywords
+setlocal syntax=c
```
chrisbra:
```suggestion
runtime! syntax/c.vim
```

runtime/syntax/opencl.vim
```diff
@@ -0,0 +1,11 @@
+" Vim syntax file
+" Language:	C
```
chrisbra: please add the usual Last Change and Maintainer header

runtime/ftplugin/opencl.vim
```diff
@@ -0,0 +1,10 @@
+" Vim filetype plugin file
+" Language:	OpenCL
```
chrisbra, line 2: please add the usual Last Change and Maintainer header

chrisbra:
again, can you please add a test?

chrisbra:
thanks, but please change the Maintainer line.

zeertzjq:
The test doesn't work

chrisbra:
thanks!

chrisbra:
Actually, it seems the tests still don't work

src/testdir/test_filetype.vim
```diff
@@ -1196,6 +1196,22 @@ func Test_cfg_file()
   filetype off
 endfunc
 
+func Test_cl_file()
+  filetype on
+
+  call writefile(['int f() {}'], 'Xfile.cl')
```
chrisbra: that doesn't match the test in `FTcl()`

runtime/autoload/dist/ft.vim
```diff
@@ -144,6 +144,14 @@ export def FTcfg()
   endif
 enddef
 
+export def FTcl()
+  if getline(1) ~= '/*'
```
chrisbra:
```suggestion
  if getline(1) =~ '/\*'
```

runtime/autoload/dist/ft.vim
```diff
@@ -144,6 +144,14 @@ export def FTcfg()
   endif
 enddef
 
+export def FTcl()
+  if getline(1) =~ '\/*'
```
chrisbra:
```suggestion
  if getline(1) =~ '/\*'
```

chrisbra:
Test still fails, probably because the `FTcl()` still uses the wrong pattern.

runtime/autoload/dist/ft.vim
```diff
@@ -144,6 +144,14 @@ export def FTcfg()
   endif
 enddef
 
+export def FTcl()
+  if getline(1) =~ '/\*'
```
zeertzjq: Is checking only the first line enough?

Freed-Wu: the file shoule have a header:

```
/*
 * XXX blahblah
 */
```

It should only occur in first line?

chrisbra:
thanks!

Issue #15775, opened 2024-09-30
Console Vim on Windows 11 sometimes fails to properly restore the original console screen contents


thegavaguy:
### Steps to reproduce

In order to make this easily reproducible I suggest grabbing Vim source code and extracting it into a directory of one's choice.  Then ...

```

C:\home\build> cd vim-9.1.0748\src
C:\home\build\vim-9.1.0748\src> cls
C:\home\build\vim-9.1.0748\src> type popupmenu.c
C:\home\build\vim-9.1.0748\src> vim -R popupmenu.c

```

From within Vim, to simulate editing activity, hit `Ctrl-F` 3 times (or `PgDn`).  Then type `:q<Enter>` to exit Vim.  Instead of restoring the the original screen contents with a new prompt right after the command line invoking Vim, the new prompt is in the middle of text much further back in the scrollback region of the console screen.  Here are some screen shots of what happens.

1.  Initial screen after `type`ing the file and entering the Vim command.

![001-Initial-screen-with-vim-command](https://github.com/user-attachments/assets/5f7cf378-bc19-4e28-a472-25b27c0da49d)

2. Vim screen after paging down in the file.

![002-vim-screen-after-paging](https://github.com/user-attachments/assets/30590a65-9038-4622-9ccf-4067a1331298)

3. The screen after exiting Vim.

![003-after-exiting-vim](https://github.com/user-attachments/assets/f21d6ac0-aa7f-46b6-a6f5-8f51db07d42e)

### Expected behaviour

Here is a screen shot of the expected result:

![006-vim-with-correct-result](https://github.com/user-attachments/assets/6c386b34-3269-48d4-a52d-5b2a1d1442dd)

I obtained it by using an earlier version of Vim.

### Version of Vim

9.1.748

### Environment

Windows 11 Home 23H2 (OS build 22631.4169)

The problem occurs with both the legacy console, implemented by conhost.exe, and the new console provided by Windows Terminal, implemented by openconsole.exe.  I am using cmd.exe as the shell in both cases.

Since the problem ends up being dependent on the use of Virtual Terminal Processing (VTP) in the Windows console, this probably only affects Windows 10 and Windows 11.  There are various checks in the code to determine whether the host system has a useable version of VTP.  Since I do not have access to earlier versions of Windows anymore, I was not able to investigate what happens on Windows 7, 8, 8.1, and 10 prior to the update that included VTP for the first time.

I note that this is a rather new issue.  Vim used to do the right thing until about two months ago.

[...]

PR #15829, opened 2024-10-07
Issue 15775 - Fix incorrect cursor position when restoring main scree…

thegavaguy:
Issue #15775 - Fix incorrect cursor position when restoring main screen buffer.

Patch 9.1.0664 moved the VTP command for switching back to the main screen buffer from termcap_mode_end() to mch_exit_c().  However, the saved cursor position from the main screen continued to be restored in termcap_mode_end().  This failed if the cursor position was beyond the console window height, since the alternate screen buffer is always the same size as the console window.

This patch restores the VTP command for switching back to the main screen buffer to termcap_mode_end().  In order to preserve the effect of patch 9.1.0664, the VTP command for switching back to the main screen buffer in mch_exit_c() is issued only if termcap mode was not active while exiting Vim.

See issue 15775 for a fuller description, with screen shots of the problem.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -3556,13 +3556,15 @@ mch_init_c(void)
     static void
 mch_exit_c(int r)
 {
+    int fTermcapMode = g_fTermcapMode;  // Copy flag since stoptermcap() will clear the flag.
+
     exiting = TRUE;
 
     vtp_exit();
 
     stoptermcap();
-    // Switch back to main screen buffer.
-    if (use_alternate_screen_buffer)
+    // Switch back to main screen buffer if TermcapMode was not active.
+    if (!fTermcapMode &&  use_alternate_screen_buffer)
 	vtp_printf("\033[?1049l");
 
     if (g_fWindInitCalled)
@@ -6337,6 +6339,10 @@ termcap_mode_end(void)
 # endif
     RestoreConsoleBuffer(cb, p_rs);
     restore_console_color_rgb();
+
+    // Switch back to main screen buffer.
+    if (exiting && use_alternate_screen_buffer)
+        vtp_printf("\033[?1049l");
 
     if (!USE_WT && (p_rs || exiting))
     {


chrisbra:
@k-takata ?

k-takata (approved):

src/os_win32.c
```diff
@@ -3556,13 +3556,15 @@ mch_init_c(void)
     static void
 mch_exit_c(int r)
 {
+    int fTermcapMode = g_fTermcapMode;  // Copy flag since stoptermcap() will clear the flag.
```
k-takata, line 3559: This line exceeds 80 columns.

chrisbra: I can fix this :)

chrisbra:
thanks!

src/os_win32.c
thegavaguy: Sorry for the long line.  Is this a Vim coding style requirement?  I read develop.txt and did not see any mention of a maximum line length.  Did I miss something?

I do agree that keeping code line lengths to a predictable maximum aids development, and a line length of 80 is certainly a time-honored standard.  If it is not documented in develop.txt it probably should be added there.

This is probably not the place for this observation, but at the end of develop.txt, in section 4.  Assumptions, it says "long is  32 or 64 bits signed, can hold a pointer".  Compilers targeting 64-bit Windows unfortunately do not all agree on the size of long, and long double.  On some long is 32-bits, on others 64-bits.  On some long double is 64-bits, on others 128 bits.  This means that a long cannot necessarily hold a pointer, which is always 64-bits.

Finally, a quick check of all the .c files shows:

```
    46 have a maximum line length < 80
    13 have a maximum line length = 80
   104 have a maximum line length > 80
```

The longest line has length 681, line 1997 of charset.c

I realize that retroactively imposing updated coding standards on a large codebase can often do more harm than good.  Requiring new code to adhere to them is just good practice.  I always make it a point to read and follow any and all documented coding styles and standards for a project, and if none are documented, do my best to follow the patterns of the existing code.  As this is my first direct submission to Vim I apologize if I missed something.

Note to self:  use a maximum line length of 80 in all future submissions to Vim.  😊

chrisbra: Thanks for contributing and for even reading through develop.txt. We should probably update it for the things you mention (PR welcome ;) ). Note the current codestyle is currently under discussion, see #15789 since it is hard to enforce using a code-formatter. But changing it is not a nice process (reviewing it is an exhaustive excercise) and breaks e.g. git blame. Still I am thinking to make a cut and have it changed to make ours and contributors life in the future easier.

chrisbra: Oh and most things mentions in develop.txt are *shoulds*  and not hard requirements, I think.

benknoble: "Breaks git blame" is a bit harsh: whitespace changes are ignored with `git blame -w`, for example, and some refactorings can be seen through with `-CCC -M` (what I've started calling "blame harder"). It's also possible to collect a list of "ignore-able blame revisions" for use with `--ignore-rev`.

PR #15830, opened 2024-10-07
A condition is always true

zeertzjq:
Problem:  A cmdlen == 0 condition is always true as ccline.cmdlen == 0
          was already checked above.
Solution: Remove the condition and the variable.


chrisbra:
thanks!

PR #15832, opened 2024-10-08
runtime(skill): Update syntax file to fix string escapes

simaoafonso-pwt:
This caught the common case:

```
"string\""
```

But not the uncommon case:

```
"<key>\\"
```


chrisbra:
Thanks. I had some troubles understanding your message, not sure if your change broke the un-common case. After testing your change, I believe what you were trying to say was:

The existing backslash escape in the syntax script allowed for the common case:
```
"string\""
```
But did not accomodate for the uncommon case:
```
"<key>\\"
```

Let's improve this in the syntax script and also skip a double backslash `\\`

PR #15834, opened 2024-10-09
"shellcmdline" doesn't work with getcompletion()

zeertzjq:
Problem:  "shellcmdline" doesn't work with getcompletion().
Solution: Use set_context_for_wildcard_arg().


chrisbra:
thanks!

PR #15836, opened 2024-10-09
fix: adjust comment and missing replace of orig_attr

glepnir:
missing orig_attr in rl branch. and adjust comments.

cc @zeertzjq


chrisbra:
thanks

PR #15837, opened 2024-10-09
Experiment: fuzzy completion based on simd

glepnir:
some experimental attempts. I might try some new algorithms.  will make a benchmark when I am satisfied with it.  It's pretty loose at the moment. but no relevant fuzzy completion tests are broken.  I'm guessing there are still edge cases that aren't caught. more actual behavioral testing is needed. It will still fall back to the original fuzzy_match for unsupported platforms to avoid incompatibilities.

(don't expect too much and pay too much attention. It may not be accepted 😊)

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -12,6 +12,19 @@
 
 #include "vim.h"
 
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+    #define PLATFORM_X86 1
+#elif defined(__ARM_NEON) || defined(__ARM_NEON__)
+    #define PLATFORM_ARM_NEON 1
+#endif
+
+#ifdef PLATFORM_X86
+    #include <immintrin.h>  // For x86/x64 SIMD intrinsics
+#elif defined(PLATFORM_ARM_NEON)
+    #include <arm_neon.h>   // For ARM NEON intrinsics
+#endif
+
+
 #ifdef FEAT_EVAL
 static void set_vv_searchforward(void);
 static int first_submatch(regmmatch_T *rp);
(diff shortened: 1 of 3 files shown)


src/search.c
```diff
@@ -4362,6 +4375,146 @@ typedef struct
...
+{
+    uint8x16_t t = vsubq_u8(r, vdupq_n_u8(63));
+    t = vcltq_u8(t, vdupq_n_u8(26));
+    t = vandq_u8(t, vdupq_n_u8(32));
+    return vaddq_u8(r, t);
+}
+#endif
+
+    int
+fuzzy_match_simd(const char_u *pattern, const char_u *text)
```
yegappan, line 4398: Can you add some comments describing this function and also describe the difference between this function and fuzzy_match()?

yegappan (changes requested):
Do you have any performance data running fuzzy match over a large number of strings using this new function?

src/search.c
glepnir: Experimental draft. So it may look irregular. Please don't mind. I will adjust it when it is ready.

glepnir:
> Do you have any performance data running fuzzy match over a large number of strings using this new function?

In preparation. I will update here once the results are available. :)

src/search.c
```diff
@@ -4362,6 +4375,146 @@ typedef struct
...
+        text_chunk = simd_tolower(text_chunk);
+
+        __m128i pattern_char = _mm_set1_epi8(tolower(pattern[p_idx]));
+        __m128i match = _mm_cmpeq_epi8(pattern_char, text_chunk);
+        int match_mask = _mm_movemask_epi8(match);
+
+	if (match_mask)
+	{
+	    // find position
+	    int shift = __builtin_ctz(match_mask);
```
ychin, line 4428: What's the compatibility of this function `__builtin_ctz`?

src/search.c
```diff
@@ -12,6 +12,19 @@
...
 #include "vim.h"
 
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+    #define PLATFORM_X86 1
+#elif defined(__ARM_NEON) || defined(__ARM_NEON__)
+    #define PLATFORM_ARM_NEON 1
+#endif
+
+#ifdef PLATFORM_X86
+    #include <immintrin.h>  // For x86/x64 SIMD intrinsics
```
ychin, line 22: I think using `immintrin.h` is a bad idea, because it pulls in all intrinsics including those from SSE3 / SSE4 / AVX instruction sets. I'm assuming you are targeting SSE2 (that's the one guaranteed by x64), but the CPU you are targeting isn't guaranteed to support say SSE4 or AVX. The MSVC compiler allows you to use all kinds of instructions even if the compiler flag doesn't specify it (e.g. with `-mavx2`). On your machine it will work and then on some random person's old machine this code will just crash, which is not great.

I think it's better to just include the headers you need to be explicit what instruction sets you are pulling in. So in this case you probably want `emmintrin.h` (which only contains SSE2 instructions)

See https://stackoverflow.com/questions/11228855/header-files-for-x86-simd-intrinsics

src/search.c
```diff
@@ -4362,6 +4375,146 @@ typedef struct
...
+#ifdef PLATFORM_X86
+static inline __m128i simd_tolower(__m128i r)
+{
+    __m128i offset = _mm_add_epi8(r, _mm_set1_epi8(63)); //  A = SCHAR_MIN
+    __m128i cmp = _mm_cmpgt_epi8(_mm_set1_epi8((char)(-102)), offset);
+    __m128i mask = _mm_and_si128(cmp, _mm_set1_epi8(32));
+    return _mm_add_epi8(r, mask);
+}
+#elif defined(PLATFORM_ARM_NEON)
+static inline uint8x16_t simd_tolower(uint8x16_t r UNUSED)
```
ychin, line 4388: Why is `r` tagged with `UNUSED`?

ychin:
Haven't taken a close look at the actual SIMD implementations yet, but just some initial comments looking at this:

1. Can you specify what hardware targets you are targeting here? Seems like for x86 you are targeting any x64 chips (which support SSE2 at a minimum) but what's the ARM target? Is this supposed to work on all ARM chips? Are we going to have situations where the program could crash because it's running on some older hardware that don't support certain instructions? Also see this comment from mine: https://github.com/vim/vim/pull/15837#discussion_r1806984841. If we use newer instructions than what older chips support we may need to switch dynamically instead of just using compiler flags.
2. Just to be clear, this works with ASCII only right? I don't think this code works with anything other than that. To be fair, ASCII is the most common things we use but I just want to be explicit about this. The utilities etc are named in such a way (e.g. `simd_tolower`) that could imply it works across the board. I would imagine you need some kind of switch to first detect if the string is ASCII only and then dynamically pick which version to use. Or is the plan to extend this to arbitrary text in Unicode or other encodings (seems hard)?
3. I wonder how this would be tested. IMO it's probably a good idea to set up a C unit test that compile both versions and feed a list of test data to make sure they perform identically on the same hardware. I guess Vim script tests could work but it's a somewhat indirect way to test this since you are just testing end-to-end.
4. As the other comment mentioned, having a consistent way to evaluate the performance of SIMD code is critical here, since adding SIMD specializations add a lot more testing and maintenance burdens and it would be nice to have concrete benchmarking from realistic scenarios. What's the motivation for this change btw? Is fuzzy slow in real usage? Genuine question (since I don't use the feature much).

src/search.c
glepnir: Yes I've been on arm for a while. So a bit sloppy. I'll test on an x64

glepnir:
> Just to be clear, this works with ASCII only right?

This is how the initial work is planned, because other encodings are somewhat troublesome.

about 4..I am also looking forward to whether the benchmark here will be improved. wait for a while

src/search.c
ychin: FWIW I looked at all the intrinsics you were using and they did seem like they are SSE2 only (from referencing Intel's [docs](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#)), so I think any x64 would work.

My comment is just that ideally you should have a target hardware spec (just saying baseline SSE2 is fine too since it works on all x64 devices so your `__x86_64__` ifdef check suffices) and make sure you stick to it. Including the all-inclusive `immintrin.h` makes it harder to stick to it since it allows newer intrinsics to be called whereas you want to the smaller headers that only include the specific instructions you want since it's easier to tell at a glance what instruction sets you are using.

I haven't done enough ARM NEON programming though. Are all instructions available to all `__ARM_NEON` platforms?

Edit: Nevermind the ARM NEON questions. I forgot it's all one instruction set so this isn't really an issue. I guess I'm traumatized having to write x86 SIMD code and needing to be very careful which instruction set version to pull from because some CPUs support some instruction sets but not others 😅. It's kind of a MSVC/x86 specific issue where you need to be quite careful with this.

glepnir: I really appreciate your advice. I will check carefully when i back here and can i ping you for any improve review?

PR #15840, opened 2024-10-10
Add missing filetype for PHP config files 

nisbet-hubbard:
This PR covers the following files distributed in the upstream tarball
```bash
/usr/local/php/etc/
├── php-fpm.conf
├── php-fpm.conf.default
└── php-fpm.d
    ├── www.conf
    └── www.conf.default
```
They are written in dosini, but currently don’t get parsed as such because they don’t use the `.ini` extension.


chrisbra:
Tests seem to fail however.

Issue #15839, opened 2024-10-09
Feature request: preserve current location when replacing the quickfix/loclist


jfly:
I explain this issue in terms of the quickfix list below, but I believe everything I say applies equally to the loclist as well.

## Problem

If I'm partway through a quickfix list, and then I do something to refresh the quickfix list, I lose my location in the list.

Here's a simple example of the issue:

1. `vim qf.vim`: You can find `qf.vim` below.
2. `source %`: This will populate (and open) the quickfix list with every line in the current file.
3. Select the Nth entry in the quickfix list (in the video below, I select entry 9, corresponding to line 9 of the buffer).
4. `source %`: This replaces the quickfix list with every line in the current file.
5. Observe how my location in the quickfix list has changed: index 1 is now selected.

<details>
<summary>qf.vim</summary>

```vimscript
" Dead simple script to convert every line in the current buffer into
" an entry in the quickfix list.
"
" This replaces the current quickfix list.
" Run with `:source %`

let qflist = []

let line_count = line('$')
for lnum in range(1, line_count)
    let line_text = getline(lnum)
    let bufnr = bufnr('%')
    call add(qflist, {'bufnr': bufnr, 'lnum': lnum, 'text': line_text})
endfor

call setqflist(qflist, 'r')

" Open the quickix list.
copen
```
</details>

And here it is in action:

[vim-2024-10-09_11.47.24.webm](https://github.com/user-attachments/assets/815d5481-f45f-4d20-b987-b1764eb2cca8)

**Describe the solution you'd like**

I think that `setqflist` should strive to preserve the current "location" in the quickfix list when the quickfix list is replaced. I say "location" rather than "index" because oftentimes you're legitimately changing/fixing things, and when you replace the list, the set of quickfix entries will be different. In other words: we know the line and column number of the currently selected entry, so when we refresh the quickfix list, I'd like for us to select the entry nearest that selected `(line, col)` rather than just selecting the very first entry in the list.

**Describe alternatives you've considered**

I could write a wrapper for `setqflist` that does the "preserve location" logic described above, but it's only useful for code *I* control that calls `setqflist`. If I'm using a plugin or some other builtin that calls `setqflist` directly, then I think the only fix is to implement this in Vim core.

**Additional context**

I originally filed this issue with Neovim: https://github.com/neovim/neovim/issues/30724, but they advised me to ask over here.

bfrg:
[...]

PR #15841, opened 2024-10-10
feature: add new `'u'` ("update") action for setqflist/setloclist

jfly:
This implements https://github.com/vim/vim/issues/15839

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -9765,6 +9765,8 @@ setqflist({list} [, {action} [, {what}]])		*setqflist()*
 			clear the list: >
 				:call setqflist([], 'r')
 <
+		'u'	Like 'r', but tries to preserve the current selection
+			in the quickfix list.
 		'f'	All the quickfix lists in the quickfix stack are
 			freed.
 
@@ -9787,8 +9789,6 @@ setqflist({list} [, {action} [, {what}]])		*setqflist()*
 				current entry.  See |quickfix-index|
 		    items	list of quickfix entries. Same as the {list}
 				argument.
-		    keep_idx	select the entry nearest the currently
-				selected entry
 		    lines	use 'errorformat' to parse a list of lines and
 				add the resulting entries to the quickfix list
 				{nr} or {id}.  Only a |List| value is supported.
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -9787,6 +9787,8 @@ setqflist({list} [, {action} [, {what}]])		*setqflist()*
 				current entry.  See |quickfix-index|
 		    items	list of quickfix entries. Same as the {list}
 				argument.
+		    keep_idx	select the entry nearest the currently
+				selected entry
 		    lines	use 'errorformat' to parse a list of lines and
 				add the resulting entries to the quickfix list
 				{nr} or {id}.  Only a |List| value is supported.
--- a/src/quickfix.c
+++ b/src/quickfix.c
@@ -7564,8 +7564,7 @@ qf_add_entries(
 	int		qf_idx,
 	list_T		*list,
 	char_u		*title,
-	int		action,
-	int		keep_idx)
+	int		action)
 {
     qf_list_T	*qfl = qf_get_list(qi, qf_idx);
     listitem_T	*li;
@@ -7588,6 +7587,7 @@ qf_add_entries(
     }
 
     int select_first_entry = FALSE;
+    int select_nearest_entry = FALSE;
 
     if (action == ' ' || qf_idx == qi->qf_listcount)
     {
@@ -7612,6 +7612,12 @@ qf_add_entries(
 	qf_free_items(qfl);
 	qf_store_title(qfl, title);
     }
+    else if (action == 'u')
+    {
+	select_nearest_entry = TRUE;
+	qf_free_items(qfl);
+	qf_store_title(qfl, title);
+    }
 
     qfline_T *entry_to_select = NULL;
     int entry_to_select_index = 0;
@@ -7633,7 +7639,7 @@ qf_add_entries(
 	qfline_T *entry = qfl->qf_last;
 	if (
 	    (select_first_entry && entry_to_select == NULL) ||
-	    (keep_idx &&
+	    (select_nearest_entry &&
 		(entry_to_select == NULL ||
 		 entry_is_closer_to_target(entry, entry_to_select, prev_fnum,
 					   prev_lnum, prev_col))))
@@ -7748,7 +7754,7 @@ qf_setprop_title(qf_info_T *qi, int qf_idx, dict_T *what, dictitem_T *di)
  * Set quickfix list items/entries.
  */
     static int
-qf_setprop_items(qf_info_T *qi, int qf_idx, dictitem_T *di, int action, int keep_idx)
+qf_setprop_items(qf_info_T *qi, int qf_idx, dictitem_T *di, int action)
 {
     int		retval = FAIL;
     char_u	*title_save;
@@ -7758,7 +7764,7 @@ qf_setprop_items(qf_info_T *qi, int qf_idx, dictitem_T *di, int action, int keep
 
     title_save = vim_strsave(qi->qf_lists[qf_idx].qf_title);
     retval = qf_add_entries(qi, qf_idx, di->di_tv.vval.v_list,
-	    title_save, action == ' ' ? 'a' : action, keep_idx);
+	    title_save, action == ' ' ? 'a' : action);
     vim_free(title_save);
 
     return retval;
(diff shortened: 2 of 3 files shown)


src/testdir/test_quickfix.vim
```diff
@@ -6370,7 +6370,7 @@ func Test_loclist_replace_autocmd()
   call setloclist(0, [], 'f')
   let s:bufnr = bufnr()
   cal setloclist(0, [{'0': 0, '': ''}])
-  au BufEnter * cal setloclist(1, [{'t': ''}, {'bufnr': s:bufnr}], 'r')
+  au BufEnter * cal setloclist(1, [{'t': ''}, {'bufnr': s:bufnr}, {'t': ''}], 'r')
```
jfly: This change is very non-obvious:

- This test replaced the loclist every time BufEnter occurs.
- Now that we preserve the loclist location when the loclist is being replaced, we *stay* on the selected item when selected it. Previously, we'd select the second item, which would switch buffers and trigger replacement of the loclist, which would then reset our selected loclist index to 1.
- Now, we stay on entry 2, and then we call `lnext`. This fails because there isn't a 3rd entry to navigate to.

The easiest fix to all this is just to add another entry to the loclist, so it's actually possible to navigate forwards in the loclist.

src/quickfix.c
```diff
@@ -12,6 +12,7 @@
  */
 
 #include "vim.h"
+#include "limits.h"
```
chrisbra: this should not be needed as it is in vim.h

jfly:
> e.g. when first running `:make` and later then later running a `:helpgrep` command

In your specific example, none of the code in this PR matters, as the `:helpgrep` always [produces a new quickfix list](https://github.com/vim/vim/blob/v9.1.0772/src/quickfix.c#L8558-L8559), and then it [explicitly jumps to the first entry](https://github.com/vim/vim/blob/v9.1.0772/src/quickfix.c#L8603-L8605).

I was careful to write the code here to only preserve location when using the `'r'` action to replace an existing quickfix list.

> Shouldn't this be an optional parameter to `setqflist()`/`setloclist()` instead?

I am of the opinion that specifically when using `'r'` to replace an existing quickfix list, this is sane default behavior. I've checked, and both `:vimgrep` and `:helpgrep` produce new quickfix lists, so they're unaffected by this change. Anyone who is replacing a list but wants to jump to the top could solve this in other ways:

- Stop replacing! Just create a new list. If you care about the limit of 10 quickfix lists, delete the old one first.
- Do the replace, and then explicitly jump to the top with `:cc 1`.

That said, I totally understand for a piece of software as ubiquitous as Vim, changing a default is pretty much guaranteed to break someone's workflow ([obligatory XKCD](https://xkcd.com/1172/)). So, I will totally understand if you tell me to hide this behind a new optional parameter. I'd like to see this feature in Vim regardless. I just think the world would benefit from minimizing the number of knobs here.

src/quickfix.c
jfly: Ah, so it is! Removed.

jfly:
> So let's please place it behind a new optional flag in this case

Can do. Do you like [@bfrq's `preserve_idx` option](https://github.com/vim/vim/issues/15839#issuecomment-2405022521)?

```
call setqflist([], 'r', {'items': my_new_list, 'preserve_idx': 1, 'id': my_qf_id})
```

Another idea: create a new `'u'` ("update") action:

```
call setqflist([], 'u', {'items': my_new_list, 'id': my_qf_id})
```

chrisbra:
I think `preserve_idx` or shorter `keep_idx` makes sense, since:
- it directly relates to the `idx` value
- unknown keys in `{what}` are explicitly ignored, thus this is backward compatible

runtime/doc/builtin.txt
```diff
@@ -9787,6 +9787,8 @@ setqflist({list} [, {action} [, {what}]])		*setqflist()*
 				current entry.  See |quickfix-index|
 		    items	list of quickfix entries. Same as the {list}
 				argument.
+		    keep_idx	select the entry nearest the currently
+		    		selected entry
```
jfly: I don't have much experience writing these. I fear this doesn't make a lot of sense. Feedback appreciated. Maybe I should say something like "useful when replacing the quickfix list and you want to keep your position in the list"?

jfly (commented):
@chrisbra, I've hidden this new feature behind a new `keep_idx` option.

src/quickfix.c
```diff
@@ -7822,7 +7914,19 @@ qf_set_properties(qf_info_T *qi, dict_T *what, int action, char_u *title)
...
     if ((di = dict_find(what, (char_u *)"items", -1)) != NULL)
-	retval = qf_setprop_items(qi, qf_idx, di, action);
+    {
+	int keep_idx = FALSE;
+	dictitem_T *keep_idx_di;
+
+	if ((keep_idx_di = dict_find(what, (char_u *)"keep_idx", -1)) != NULL)
+	{
+	    if (keep_idx_di->di_tv.v_type != VAR_NUMBER)
+		return FAIL;
```
jfly: I'm not sure if it's OK to bail out out `qf_set_properties` here. The other code just sets `retval = FAIL` and then goes on to possibly do stuff after this block of `if` statements.

Feedback appreciated!

src/testdir/test_quickfix.vim
jfly: Resolving. [We decided not to change the default behavior](https://github.com/vim/vim/pull/15841#pullrequestreview-2361372539), which means this test does not need to change.

src/quickfix.c
chrisbra: Yeah,  I think following what has been done for the other properties is correct and should be done here as well.

chrisbra:
Thanks. There are a few warnings, that are not covered by tests. Can you try to add a few more tests then?

src/quickfix.c
jfly: Ok, updated. This got kind of messy, I had to duplicate the call to `qf_setprop_items`.

I'm going to mark this resolved, but do let me know if you can think of a less bad way of doing this.

jfly:
> There are a few warnings, that are not covered by tests. Can you try to add a few more tests then?

Sure. Which warnings? Things I'm aware of that are not exhaustively tested:

1. `entry_is_closer_to_target` has a lot of careful handling around missing values. I've expanded to test for missing line numbers. I could do more here: we aren't testing how it handles column numbers, for instance.
2. I was not testing what happens if `keep_idx` is not a number. Fortunately, that's not an issue anymore now that we've dropped `keep_idx` in favor of a new `'u'` action.

runtime/doc/builtin.txt
jfly: Marking this resolved. I think the updated docs for the new `'u'` action are ok. Feedback appreciated!

src/quickfix.c
jfly: For the record, this is irrelevant now that we're doing a `'u'` action instead of a `keep_idx` option.

src/testdir/test_quickfix.vim
```diff
@@ -6462,6 +6462,32 @@ func Test_quickfix_buffer_contents()
   call setqflist([], 'f')
 endfunc
 
+func Test_quickfix_update()
+  " Setup: populate a couple buffers
+  call setline(1, range(1,5))
```
yegappan: You should create a new buffer before changing the contents of the buffer.

src/testdir/test_quickfix.vim
```diff
@@ -6462,6 +6462,32 @@ func Test_quickfix_buffer_contents()
...
+  call assert_equal(3, getqflist({'idx' : 0}).idx)
+
+  " Update the quickfix list. Make sure the third entry is still selected.
+  call setqflist([], 'u', { 'items': items })
+  call assert_equal(3, getqflist({'idx' : 0}).idx)
+
+  " Update the quickfix list again, but this time with missing line number
+  " information. Confirm that we keep the current buffer selected.
+  call setqflist([{'bufnr': b1}, {'bufnr': b2}], 'u')
+  call assert_equal(2, getqflist({'idx' : 0}).idx)
```
yegappan, line 6489: The test should clean up the newly created buffers.

yegappan (changes requested):

src/testdir/test_quickfix.vim
jfly: How do I ensure that happens even if the test fails?

src/testdir/test_quickfix.vim
jfly: Done!

src/testdir/test_quickfix.vim
jfly: I've added a couple of `%bwipe!` commands at the end of the test, so I'm going to mark this done.

`<rant>`I skimmed the other tests in this file, and I'm not convinced they're being consistent about cleaning up the buffers they allocate. My inclination would be to try to solve this structurally rather than carefully: the test infrastructure/framework should make sure tests start with a clean slate, and furthermore, clean up any buffers/other stuff allocated during the test.`</rant>`

jfly:
> There are a few warnings, that are not covered by tests

Oh, I see the codecoverage warnings now. I'll work on the code coverage now.

src/testdir/test_quickfix.vim
chrisbra: I think one `%bwipe!` should be fine :)

chrisbra:
thanks!

src/testdir/test_quickfix.vim
jfly: Oooh does that close all buffers? Ha! Got it.

Fixed.

jfly:
Let me know if you'd like me to rebase/squash the changes in this PR.

chrisbra:
no worries, I already fixed this up locally

jfly:
Thank you, @chrisbra!

PR #15845, opened 2024-10-10
Fix incorrect Win32 error checking & minor refactors

nir9:
Problem: In the function that adjusts the process privileges there is a mistake in which ```GetLastError``` is called after ```CloseHandle```, though clearly the last error check is meant for the privileges related call before hand and the current state appears like a mistake.

Solution: Fix the above problem, and also various other fixes:
- Remove the static variable done since the ```PlatformId``` is only called during initialization
- Fix incorrect parameter passed to the Win32 API privileges function

Ideally, I would also call an error handling function in case there is a failure, but it seems this is not customary in the early initialization code.

### DIFF
--- a/src/os_win32.c
+++ b/src/os_win32.c
@@ -918,9 +918,8 @@ null_libintl_wputenv(const wchar_t *envstring UNUSED)
  * Enables or disables the specified privilege.
  */
     static BOOL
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks

Issue #7567, opened 2020-12-29
Ctrl-L doesn't work with complete() 


idbrii:
**Describe the bug**
When using `complete()` to create a completion menu, CTRL-L fails to insert a character from the current match. Instead, a literal ^L is inserted.

> CTRL-L		  Add one character from the current match, may reduce the number of matches.

**To Reproduce**

minimal.vim:
```vim
" From :help complete()
inoremap <C-j> <C-R>=ListMonths()<CR>
func! ListMonths()
    call complete(col('.'), ['January', 'February', 'March',
                \ 'April', 'May', 'June', 'July', 'August', 'September',
                \ 'October', 'November', 'December'])
    return ''
endfunc

" Set options that avoid filling first completion.
set completeopt=menuone,noinsert,noselect

```

Detailed steps to reproduce the behavior:
1. Run `gvim --clean minimal.vim` (see above)
1. `source %` to load minimal.vim
1. Start inserting
1. type C-n to open ins-completion menu
1. type M to limit to March, May
1. type C-L to insert 'a'
1. works!
1. type C-u to clear
1. type C-j to open complete() completion menu 
1. type M to limit to March, May
1. type C-L to insert 'a'
1. fails!

**Expected behavior**
The letter a (the next character in the first match "March") is inserted just like it is for ins-completion.

**Actual behavior**
A literal ^L is inserted.

**Environment (please complete the following information):**
 - Vim version

	VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Dec  7 2020 23:02:04)
	MS-Windows 64-bit GUI version with OLE support
	Included patches: 1-2107
	Compiled by appveyor@APPVYR-WIN
 - OS: Windows 10 OS Build 18363.1256
 - GUI

**Additional context**
Encountered this using asyncomplete which uses `complete()` in prabirshrestha/asyncomplete.vim#177.

chrisbra:
>     Run gvim --clean minimal.vim (see above)
>     source % to load minimal.vim
>     Start inserting
>     type C-n to open ins-completion menu
>     type M to limit to March, May
>     type C-L to insert 'a'
>     works!

Typing Ctrl-N starts keyword completion mode. In that mode, after limiting the popup menu, typing Ctrl-L does add one character from the popup menu. Note: this does not use the completion function `ListMonths()`.

>     type C-u to clear
>     type C-j to open complete() completion menu
>     type M to limit to March, May
>     type C-L to insert 'a'

[...]

PR #15846, opened 2024-10-10
fix: do not close pum on <BS> using `complete()`

### LINKED ISSUE #15833: complete: keep pum visible when pressing <BS> for complete()
First of all, I am not sure if this should be considered a bug or a feature request, so forgive me if I opened a wrong type of issue😅

When using `complete()` and showing the pum, it is removed when pressing `<BS>`. This is not the same behaviour as using omnifunc and other CTRL-X submodes. Then the pum will still be visible even after pressing `<BS>`

I would like it to behave the same when completing via `complete()`, and have the pum still be visible after a `<BS>`

**Additional context**
I have opened an issue over in neovim (https://github.com/neovim/neovim/issues/30723) and have added a patch there that seems to fix the problem. As this code is shared between neovim and vim, I thought I would open an issue here as well to see what you think. Is it reasonable to unify the behaviour, and make `complete()` work in the same way as the other CTRL-X submodes? If so, does the following patch seem reasonable?

```diff
diff --git a/src/insexpand.c b/src/insexpand.c
index eb920a751..24258ec72 100644
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1917,7 +1917,6 @@ ins_compl_bs(void)
     // Respect the 'backspace' option.
     if ((int)(p - line) - (int)compl_col < 0
 	    || ((int)(p - line) - (int)compl_col == 0 && !ctrl_x_mode_omni())
-	    || ctrl_x_mode_eval()
 	    || (!can_bs(BS_START) && (int)(p - line) - (int)compl_col
 							- compl_length < 0))
 	return K_BS;
```

Using a vimrc with the snippet from `:h complete()`

Before patch:

https://github.com/user-attachments/assets/56c4b332-6a60-4493-8315-709f4b86df57

After patch:

https://github.com/user-attachments/assets/56e88952-fc6b-4a4d-ad13-02036da72e8e

seblyng:
Closes #15833

### DIFF
--- a/src/testdir/test_popup.vim
+++ b/src/testdir/test_popup.vim
@@ -51,16 +51,15 @@ func Test_popup_complete()
   call assert_equal(["August"], getline(1,2))
   %d
 
-  " <BS> - Delete one character from the inserted text (state: 1)
-  " TODO: This should not end the completion, but it does.
-  " This should according to the documentation:
-  " January
-  " but instead, this does
-  " Januar
-  " (idea is, C-L inserts the match from the popup menu
-  " but if the menu is closed, it will insert the character <c-l>
+  " C-L inserts the match from the popup menu
   call feedkeys("aJ\<f5>\<bs>\<c-l>\<esc>", 'tx')
-  call assert_equal(["Januar"], getline(1,2))
+  call assert_equal(["January"], getline(1,2))
+  %d
+
+  " C-L and a <bs> back to `J`, and then select next item from completion
+  " menu which should still be open
+  call feedkeys("aJ\<f5>\<bs>\<bs>\<bs>\<bs>\<bs>\<bs>\<c-n>\<c-y>\<esc>", 'tx')
+  call assert_equal(["June"], getline(1,2))
   %d
 
   " any-non special character: Stop completion without changing the match
--- a/src/insexpand.c
+++ b/src/insexpand.c
@@ -1916,7 +1916,6 @@ ins_compl_bs(void)
     // Respect the 'backspace' option.
     if ((int)(p - line) - (int)compl_col < 0
 	    || ((int)(p - line) - (int)compl_col == 0 && !ctrl_x_mode_omni())
-	    || ctrl_x_mode_eval()
 	    || (!can_bs(BS_START) && (int)(p - line) - (int)compl_col
 							- compl_length < 0))
 	return K_BS;


seblj:
I should probably update the test?

glepnir:
update the test case in test_popup which i already mentioned in issue and add some test case.

seblj:
Will do later today

seblj:
Let's see what CI says.

Let me know if I should it needs any more tests. I added a test that does a bunch of backspaces, and then do `<C-n>` to select the next item in the menu that should now be open.

src/testdir/test_popup.vim
```diff
@@ -51,16 +51,15 @@ func Test_popup_complete()
...
+  " C-L inserts the match from the popup menu
   call feedkeys("aJ\<f5>\<bs>\<c-l>\<esc>", 'tx')
-  call assert_equal(["Januar
"], getline(1,2))
+  call assert_equal(["January"], getline(1,2))
+  %d
+
+  " C-L and a <bs> back to `J`, and then select next item from completion
+  " menu which should still be open
+  call feedkeys("aJ\<f5>\<bs>\<bs>\<bs>\<bs>\<bs>\<bs>\<c-n>\<c-y>\<esc>", 'tx')
```
glepnir, line 61: Is it possible to add some more test cases to check for edge cases?

seblyng: Sure, but do you have any edge cases in mind?

seblj:
Failed CI seems unrelated to my changes?

glepnir:
Nope. random build error

src/testdir/test_popup.vim
glepnir: Single and multiple bs behaviors. otherwise LGTM

seblyng: Isn't that what I already have? I changed the failing test to a single bs, and then `<c-l>` to accept it, and then this with multiple bs

glepnir: Yes, it's just that the few case always feels a little loose. but it's okay. It's reasonable.

bstaletic:
After ~~receiving the bat signal~~ ping from puremourning, I did test this pull request with YouCompleteMe.
It does subtly break things.

Having vim hold onto the menu while the user hits backspace means a more appropriate menu now has to be explicitly requested by the user.

Here's how it used to work:

https://asciinema.org/a/xuCNFKE0LuJbqSrBpJOYIPLF3

And here's how it works after this patch:

https://asciinema.org/a/IkQ7OLa0uVYY6rjOJLPQ4PB1h

Notice that, at the end, where the line is `std::common_reference_`, in the "after" recording, the only candidate remained `common_reference_with`. To get all the candidates I, as a user, would have had to hit `<C-Space>` and explicitly make YouCompleteMe give me all the completions.

bstaletic:
A more problematic breaking change that this introduces is holding onto `<C-x><C-l>` menu when we shouldn't have.
Unfortunately, asciinema does not let vim receive `<C-l>`, so I'll have to describe the steps:

1. Open a new python file.
2. Write `aaaaaaaaaa = 5555555555` on the first line.
3. Open a new line.
4. Use `<C-x><C-l>` to complete the whole line.
5. Kill `=` and `555` with `<C-w><C-w>`
6. Start backspacing into `aaaaa`.

Expectation: YouCompleteMe takes over word completion automatically.
Actual: YouCompleteMe calls `complete()` once, the vim recalls... the full line completion and starts doing that instead.

chrisbra:
oh wow, that's a bummer. May be we can change the behaviour only for `CTRL-H` and not `BS` instead?

bstaletic:
@chrisbra YouCompleteMe used to have `s:OnCharacterDelete`, to implement the current behaviour back before `complete()` was a thing and we had to feed `<C-x><C-u>`.
We could resurrect that part of code again and make it work with the newly fixed `complete()`. It might be ugly, because of possible vim version checks, but it's doable.

The second problem that I've posted, the one where vim seems to be holding onto `<C-x><C-l>` for far too long - that one is a concern.
Can that part be fixed on the vim side? I have no clue right now how could that one be solved by plugins.

glepnir:
does relate https://github.com/vim/vim/issues/7567 ?  The failure you mentioned is not caught by our test case. strange. I tried the minimal reproducible steps. Unfortunately I can't find it. It seems that ycm is doing something special here

bstaletic:
> does relate https://github.com/vim/vim/issues/7567 ?

I don't think so... That's talking about `:h i_CTRL-l`, while I mentioned `:h i_CTRL-x_CTRL-l`.

> I tried the minimal reproducible steps. Unfortunately I can't find it. It seems that ycm is doing something special here

Hmm... YCM is, effectively, calling `complete()` on every `TextChangedI`, but through a timer, so that calculation of new menu candidates happens without blocking things. I guess that does count as "something special".

glepnir:
yes it is the usual approach for completion plugins with a debounce timer. I might try this again later. I hope to find a minimal test case and update our tests.

glepnir:
@bstaletic  Hi I want to know how the matches list is generated when ycm calls complete here? When the timer is triggered, I see that ycm in the gif provides the entire line of text after receiving the patch with few times bs. So how is the list of the complete function generated? This will help me complete a test case :)

bstaletic:
> When the timer is triggered, I see that ycm in the gif provides the entire line

That's exactly the problem. YCM never completes the full line. YCM is only trying to complete identifiers in this case - the `aaaaaaaa` identifier. You can recognize YCM completion by the `[ID]` which ends up in the `menu` property of the completion candidates.

This is where YCM takes the generated identifier candidates and converts them to vim data:
https://github.com/ycm-core/YouCompleteMe/blob/master/python/ycm/client/completion_request.py#L186

And this class is responsible for producing identifier candidates:
https://github.com/ycm-core/ycmd/blob/master/ycmd/completers/all/identifier_completer.py

I hope that helps! I'll gladly answer any further questions.

chrisbra:
converting to draft, until we have found out how to best handle this breaking change.

chrisbra:
this has stalled, so closing

PR #15848, opened 2024-10-11
test_strftime: Use GMT offsets to validate tzset

jamessan:
Some systems only provide timezones that follow the geographical region naming (e.g. `America/New_York`) by default and require an additional install for other names (like `EST`).

The `GMT+<offset>` format must always be honored, so use that to generate distinct data for the test.

### DIFF
--- a/src/testdir/test_functions.vim
+++ b/src/testdir/test_functions.vim
@@ -277,17 +277,17 @@ func Test_strftime()
     let tz = $TZ
   endif
 
-  " Force EST and then UTC, save the current hour (24-hour clock) for each
-  let $TZ = 'EST' | let est = strftime('%H')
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks. What kind of systems do not support those names? I tend to prefer using `TZ=EST` instead of guessing the region, so wondering

jamessan:
> What kind of systems do not support those names?

The latest update of the tzdata package in Debian moved a lot of those into the tzdata-legacy package.

>   * Upstream obsoleted the System V names CET, CST6CDT, EET, EST*, HST, MET,
    MST*, PST8PDT, and WET. They are symlinks now. Move those zones to
    tzdata-legacy and update /etc/localtime on package update to the new names.
    Please use Etc/GMT* in case you want to avoid DST changes.

PR #15850, opened 2024-10-11
runtime(vimtutor): Fix command in section 4.4 in spanish language

abenito-licu:
Added 's' character to complete sentence ':/laas/la/g'


chrisbra:
thanks!

PR #15855, opened 2024-10-11
More Windows size_t warning fixes 

mrdubya:


mrdubya:
I'm missing something fundamental that is showing up in the tests appearing on a different platform. I'll crawl back into my hole.

PR #15835, opened 2024-10-09
Tests for set and option are not comprehensive

Milly:
- Moved tests for `:set` to test_set.vim.
- Added comprehensive test patterns.
- Split up long errors test into separate tests for each options.
- Reordering option tests alphabetically.

PR #15856, opened 2024-10-11
tests: tests for option settings are not comprehensive

Milly:
Problem: tests for option settings are not comprehensive
Solution: add help-based black box testing

This PR is split part of #15835.


chrisbra:
thanks, that is much better to review!

PR #15857, opened 2024-10-11
`writefile`: optionally update the corresponding buffer's `mtime`

bobrippling:
This permits plugins which use `writefile()` to write to a buffer, to also update the buffer's `mtime`, so future `:write`s won't then think there's been a change to the file outside of vim

### DIFF
--- a/runtime/doc/builtin.txt
+++ b/runtime/doc/builtin.txt
@@ -12417,6 +12417,10 @@ writefile({object}, {fname} [, {flags}])			*writefile()*
 		     When {flags} does not contain "S" or "s" then fsync() is
 		     called if the 'fsync' option is set.
 
+		't'  if the {fname} corresponds to a buffer, update that buffer's
+		     modified time so future writes to the buffer don't think there's
+		     changes outside of Vim.
+
 		An existing file is overwritten, if possible.
 
 		When the write fails -1 is returned, otherwise 0.  There is an
--- a/src/proto/buffer.pro
+++ b/src/proto/buffer.pro
@@ -23,6 +23,9 @@ void free_buf_options(buf_T *buf, int free_p_ff);
 int buflist_getfile(int n, linenr_T lnum, int options, int forceit);
 buf_T *buflist_findname_exp(char_u *fname);
 buf_T *buflist_findname(char_u *ffname);
+#ifdef UNIX
+buf_T *buflist_findname_stat(char_u *ffname, stat_T *stp);
+#endif
 int buflist_findpat(char_u *pattern, char_u *pattern_end, int unlisted, int diffmode, int curtab_only);
 int ExpandBufnames(char_u *pat, int *num_file, char_u ***file, int options);
 buf_T *buflist_findnr(int nr);
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -38,7 +38,6 @@ static void	buflist_getfpos(void);
 static char_u	*buflist_match(regmatch_T *rmp, buf_T *buf, int ignore_case);
 static char_u	*fname_match(regmatch_T *rmp, char_u *name, int ignore_case);
 #ifdef UNIX
-static buf_T	*buflist_findname_stat(char_u *ffname, stat_T *st);
 static int	otherfile_buf(buf_T *buf, char_u *ffname, stat_T *stp);
 static int	buf_same_ino(buf_T *buf, stat_T *stp);
 #else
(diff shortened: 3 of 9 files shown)


bobrippling:
Test failure appears to be unrelated:

```
From test_popupwin.vim:
Found errors in Test_popup_beval():
Run 1, 18:00:10 - 18:00:23:
command line
  ..script /Users/runner/work/vim/vim/src/testdir/runtest.vim[607]
  ..function RunTheTest[57]
  ..Test_popup_beval[39]
  ..VerifyScreenDump

line 67: See dump file difference: call term_dumpdiff("testdir/failed/Test_popupwin_beval_3.dump", "testdir/dumps/Test_popupwin_beval_3.dump");
difference in line 4: "|4| @10|t+0#0000001#ffd7ff255|e|x|t| +0#0000000#ffffff0@58"
```

runtime/doc/builtin.txt
```diff
@@ -12417,6 +12417,10 @@ writefile({object}, {fname} [, {flags}])			*writefile()*
 		     When {flags} does not contain "S" or "s" then fsync() is
 		     called if the 'fsync' option is set.
 
+		't'  if the {fname} corresponds to a buffer, update that buffer's
+		     modified time so future writes to the buffer don't think there's
+		     changes outside of vim.
```
```suggestion
		     changes outside of Vim.
```

bobrippling: Thanks - sorted

vim-ml:
On 2024-10-12, Rob Pilling wrote:
> @bobrippling commented on this pull request.
> 
> -------------------------------------------------------------------------------
> 
> In runtime/doc/builtin.txt:
> 
> > @@ -12417,6 +12417,10 @@ writefile({object}, {fname} [, {flags}])                     *writefile()*
>                      When {flags} does not contain "S" or "s" then fsync() is
>                      called if the 'fsync' option is set.
> 
> +               't'  if the {fname} corresponds to a buffer, update that buffer's
> +                    modified time so future writes to the buffer don't think there's
> +                    changes outside of vim.

there's -> there have been

Regards,
Gary

chrisbra:
Thanks. Can you plesae add a test? I suppose you could use `:let mess=execute(':w')` after using `writefile()` and then check, that `mess` does not contain the warning. Or perhaps use `:checktime` and check that `W11` isn't returned.

bobrippling:
On testing this, I'm not convinced it's that useful - my initial usecase was to do a background write to a slow filesystem, without locking up Vim's UI, but `writefile` is still called synchronously (even from a timer).

I need to find a different approach to write the file and then update vim's `mtime`, I'll put this into draft for now

chrisbra:
you may want to consider disabling the 'fsync' option.

bobrippling:
Ah thank you, I'll see how well that helps

src/proto/buffer.pro
```diff
@@ -23,6 +23,9 @@ void free_buf_options(buf_T *buf, int free_p_ff);
 int buflist_getfile(int n, linenr_T lnum, int options, int forceit);
 buf_T *buflist_findname_exp(char_u *fname);
 buf_T *buflist_findname(char_u *ffname);
+#ifdef UNIX
+buf_T *buflist_findname_stat(char_u *ffname, stat_T *stp);
+#endif
```
k-takata, line 28: This file is generated by the cproto command. The `#ifdef` and `#endif` lines are not needed:
```suggestion
buf_T *buflist_findname_stat(char_u *ffname, stat_T *stp);
```

However, the definitions of `buflist_findname()` and `buflist_findname_stat()` need to be adjusted to make cproto work. E.g.:
```C
    buf_T *
buflist_findname(char_u *ffname)
{
#ifdef UNIX
    stat_T	st;

    if (mch_stat((char *)ffname, &st) < 0)
	st.st_dev = (dev_T)-1;
    return buflist_findname_stat(ffname, &st);
#else
    return buflist_findname_stat(ffname, NULL);
#endif
}

/*
 * Same as buflist_findname(), but pass the stat structure to avoid getting it
 * twice for the same file.
 * Returns NULL if not found.
 */
    buf_T *
buflist_findname_stat(
    char_u	*ffname,
    stat_T	*stp)
{
    buf_T	*buf;

    // Start at the last buffer, expect to find a match sooner.
    FOR_ALL_BUFS_FROM_LAST(buf)
	if ((buf->b_flags & BF_DUMMY) == 0 && !otherfile_buf(buf, ffname
#ifdef UNIX
		    , stp
#endif
		    ))
	    return buf;
    return NULL;
}
```

bobrippling:
Thanks for the review and comments - `'fsync'` seems to have alleviated this so I'll close the PR (don't think there's further use in providing the ability to update a buffer's `mtime` for now)

PR #15863, opened 2024-10-13
Fix correct name to "Vim script"

h-east:
The late Mr. Bram stated this clearly:
https://groups.google.com/g/vim_dev/c/3Z5yM8KER2w/m/wAqws0QSEAAJ

> "Vim script" is the short form.

The language label on GitHub is also wrong, so I created a new discussion about it.
https://github.com/orgs/community/discussions/141314

P.S.
I'll make a pull request for EditorConfig later.

### DIFF
--- a/runtime/autoload/getscript.vim
+++ b/runtime/autoload/getscript.vim
@@ -365,7 +365,7 @@ fun! getscript#GetLatestVimScripts()
 endfun
 
 " ---------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


zeertzjq:
Should "script" still be capitalized in a title?

h-east:
I think we should always write "Vim script".

chrisbra:
thanks

chrisbra:
thanks

PR #15865, opened 2024-10-13
Disambiguate `ctermfg=Brown` to show same colour across terminals

nisbet-hubbard:
As per in-line comment

> " Many terminals can only use six different colors (plus black and white).
" Therefore the number of colors used is kept low.

All colours not expressed as hex strings are from the [16 cterm-colors](https://vimhelp.org/syntax.txt.html#cterm-colors). Brown here means colour No. 3 (Brown, DarkYellow) and is shown as yellow on 8-colour and 16-colour terminals 

IMG_2035

but as brown on 256-colour terminals

IMG_2036

This PR uses the alternative, unambiguous term ‘DarkYellow’ so Statement can exhibit the same colour in all terminals.

Pro: cross-terminal consistency 

Con: users on 256-colour terminals will see a slight colour change in the default palette 

I’m not entirely sure whether this is worthwhile to do.


chrisbra:
Thanks, but I think most terminals nowadays support either 256colors or even true-colors. Meaning this will likely annoy people of 256 color terminals while not having noticable advantages for other users.

PR #15871, opened 2024-10-14
Add info on default colorscheme in syntax.txt

nisbet-hubbard:
One often reads on support forums that Vim’s `:colo default` is peachpuff. It’s less widely known that this is a slightly altered version of legacy peachpuff (and desert for dark background) and is loaded from "syntax/syncolor.vim", giving rise to confusion when users try to run `:colo peachpuff` after reading such advice and get a different look.

This PR documents the colorscheme of `:colo default`.


chrisbra:
Thanks, but I am not sure this is correct. The default colorscheme comes from `$VIMRUNTIME/colors/default.vim` at least that is where `g:colors_name` is defined.

chrisbra:
okay thanks.

PR #15873, opened 2024-10-14
fix: 'spellfile' does not check correctly for comma-separated entries

Milly:
'spellfile' does not check correctly for comma-separated entries.

- The path contains commas causes an error.
- Invalid file names separated by commas (e.g. empty or without a ".add" suffix) do not cause an error.


src/spell.c
```diff
@@ -2250,7 +2250,7 @@ parse_spelllang(win_T *wp)
 	else
 	{
 	    // One entry in 'spellfile'.
-	    copy_option_part(&spf, spf_name, MAXPATHL - 5, ",");
+	    copy_option_part(&spf, spf_name, MAXPATHL - 4, ",");
```
Milly, line 2253: 📝 The `maxlen` of `copy_option_part()` includes the trailing NUL, so `MAXPATHL - 4` is appropriate.

chrisbra:
thanks

PR #15874, opened 2024-10-14
runtime(doc): add preview flag to statusline example

msaher:
Problem:
The standard statusline example is missing the preview flag "%w"

Solution:
Add the preview flag "%w"


chrisbra:
thanks

msaher:
@chrisbra, seems like I have accidentally missed the second example as well.

```
	Similar, but add ASCII value of char under the cursor (like "ga") >vim
	  set statusline=%<%f%h%w%m%r%=%b\ 0x%B\ \ %l,%c%V\ %P

```

PR #15881, opened 2024-10-15
Fix Kitty protocol decoding of the Escape key

### LINKED ISSUE #15868: Ctrl-v<Esc> in kitty produces trailing garbage
### Steps to reproduce

In the kitty terminal, do `vim --clean` and then do Ctrl-v<Esc>.

### Expected behaviour

I expect to see "^[" (a literal escape character).

What I actually see is "^[[27u".

I bisected and this has been there since 9.0.0930.

### Version of Vim

9.1.0781

### Environment

Debian 12.6, and I also see this on Linux Mint 21.1.

### Logs and stack traces

_No response_

cfillion:
Problem:  `<CSI>27[;<mod>]u` is not decoded to a literal Escape character in kitty and foot
Solution: disable XTerm modifyOtherKeys form 1 (`<CSI>27;<mod>;<key>~`) when the kitty protocol is enabled

XTerm's modifyOtherKeys uses 27 (Escape) to denote a special alternate form (as per the author: "noticing that 27 was unused"). The kitty keyboard protocol doesn't special-case 27. Escape is sent just like other printable keys. Some keys are sent with a different terminator than `u` however in these cases the key number isn't an Unicode codepoint and cannot be simply used as a printable character in vim's Ctrl+V, so not handling them in this function seems OK (on par with other terminals).

References:
- https://invisible-island.net/xterm/modified-keys.html
- https://sw.kovidgoyal.net/kitty/keyboard-protocol/
- https://codeberg.org/dnkl/foot/src/commit/e891abdd6a6652bd46b28c1988700a7f30931210/kitty-keymap.h
- https://github.com/kovidgoyal/kitty/blob/d31459b0926f2afddc317d76314e4afd0d07d473/kitty/key_encoding.c#L193

fixes: #15868

### DIFF
--- a/src/edit.c
+++ b/src/edit.c
@@ -1625,7 +1625,8 @@ decodeModifyOtherKeys(int c)
     if (typebuf.tb_len >= 4 && (c == CSI || (c == ESC && *p == '[')))
     {
 	idx = (*p == '[');
-	if (p[idx] == '2' && p[idx + 1] == '7' && p[idx + 2] == ';')
+	if (p[idx] == '2' && p[idx + 1] == '7' && p[idx + 2] == ';' &&
+		kitty_protocol_state != KKPS_ENABLED)
 	{
 	    form = 1;
 	    idx += 3;
@@ -1640,9 +1641,10 @@ decodeModifyOtherKeys(int c)
 		break;
 	    ++idx;
 	}
+	int kitty_no_mods = argidx == 0 && kitty_protocol_state == KKPS_ENABLED;
 	if (idx < typebuf.tb_len
 		&& p[idx] == (form == 1 ? '~' : 'u')
-		&& argidx == 1)
+		&& (argidx == 1 || kitty_no_mods))
 	{
 	    // Match, consume the code.
 	    typebuf.tb_off += idx + 1;
@@ -1652,7 +1654,7 @@ decodeModifyOtherKeys(int c)
 		typebuf_was_filled = FALSE;
 #endif
 
-	    mod_mask = decode_modifiers(arg[!form]);
+	    mod_mask = kitty_no_mods ? 0 : decode_modifiers(arg[!form]);
 	    c = merge_modifyOtherKeys(arg[form], &mod_mask);
 	}
     }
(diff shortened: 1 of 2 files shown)


cfillion:
> But this doesn't seem to make any difference for me? How to verify this? 

Just pressing Ctrl+V then Escape in either Kitty or Foot. Outputs `^[[27;129u` or `^[[27u` depending on the sate of numlock. `^[` after the patch.

Screencap

> We should probably also add a test for this in testdir/test_termcodes.vim

That would be great. What is the correct way to simulate interraction with the terminal? Looks like I'd need to make vim believe the terminal sent a specific CSI response for `kitty_protocol_state` to be set by term.c:5671.

```diff
diff --git a/src/testdir/test_termcodes.vim b/src/testdir/test_termcodes.vim
index 507753c21..e58e06f67 100644
--- a/src/testdir/test_termcodes.vim
+++ b/src/testdir/test_termcodes.vim
@@ -2610,6 +2610,10 @@ func Test_cmdline_literal()
   call feedkeys(':' .. GetEscCodeCSI27('V', '6') .. GetEscCodeCSI27('Y', '5') .. "\<C-B>\"\<CR>", 'Lx!')
   call assert_equal("\"\<Esc>[27;5;89~", @:)
 
+  " FIXME: some magic command to enable kitty_protocol_state here?
+  call feedkeys('a' .. GetEscCodeCSIu('V', '5') .. GetEscCodeCSIu("\<Esc>", '129') .. "\<Esc>", 'Lx!')
+  call assert_equal("\<Esc>", getline(1))
+
   set timeoutlen&
 endfunc
 
```

chrisbra:
Hm, thanks. I see it now, but only when Num Lock is off. With NumLock on (which was my default), it only ever shows: `^[[27u` (patched or unpatched). kitty v0.36.4 (debian testing).

cfillion:
My bad, fixed. I've also added tests for both numlock and non-numlock (looks like just `feedkeys` with the "protocol is enabled" reply is enough).

chrisbra:
thanks for quick fixing. I think this is fine to be merged.

PR #15884, opened 2024-10-16
tests: 'equalalways' test is not correct

Milly:
The test preprocessing is not running correctly, but the results remain the same.


chrisbra:
thanks, makes sense

PR #15885, opened 2024-10-16
Support jinja syntax as secondary filetype

### LINKED ISSUE #15880: Allow jinja syntax to be used as a secondary filetype
Jinja syntax doesn't work for dotted filetypes like `conf.jinja`, `sshconfig.jinja`, `nginx.jinja` etc. Probably because of this piece of code:

```vim
if exists('b:current_syntax')
    finish
endif
```

Is there a workaround for this case?

Allow jinja syntax to be sourced as a secondary filetype (even when `b:current_syntax` is already defined).

**Additional context**
Jinja usually is used as a template engine, so it's natural to set dotted filetypes for it. Jinja syntax was added recently in #15689.

gpanders:
Resolves #15880

### DIFF
--- a/runtime/syntax/jinja.vim
+++ b/runtime/syntax/jinja.vim
@@ -3,7 +3,7 @@
 " Maintainer: Gregory Anders
 " Upstream: https://gitlab.com/HiPhish/jinja.vim
 
-if exists('b:current_syntax')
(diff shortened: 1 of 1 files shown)


chrisbra:
thanks!

PR #15888, opened 2024-10-17
Refactor cmdhist.c to remove calls to STRLEN()

basilisk0315:


basilisk0315:
I am getting a heap-buffer-overflow error and I am having difficulty tracking down where the problem is.

Does anyone know how I can find exactly which test or tests are triggering the error? I have been trawling the logs but no joy.

Any help is appreciated.

Cheers

vim-ml:
Op za 19 okt 2024 00:52 schreef John Marriott ***@***.***>:

> I am getting a heap-buffer-overflow error and I am having difficulty
> tracking down where the problem is.
>
> Does anyone know how I can find exactly which test or tests are triggering
> the error? I have been trawling the logs but no joy.
>
> Any help is appreciated.
>

One thing I can think of is running your new code alongside the old one,
and having a runtime assert on all calculations and function calls. That
way, you'd get the information at the instant where the code differs in
action?

Christ van Willegen

>

chrisbra:
Can you reproduce the heap buffer overflow locally? I typically use this: https://github.com/google/sanitizers/wiki/AddressSanitizerAndDebugger to have the debugger stop at the place where the buffer overflow is detected and then from there check the frames around it

basilisk0315:
Thanks for the comments.

I probably didn't make myself clear. The heap buffer error is reported as part of github's asan tests, but I have not been able to reproduce it locally. I downloaded the test log files and have going through them to identify the specific test or tests that are triggering the error, but no luck so far. The logs do report the call stack and the line of code which had the error, but I was hoping the test logs would mention which test triggered the error.

jacobmealey:
Hello, I was able to reproduce the heap overflow locally by turning on the address sanitizer around line 700 of `src/makefile`. I ran the tests and the failing one seems to be `Test_viminfo_jumplist()`. It provided the same stack trace from the sanitizer. When I ran the tests it wrote to src/testdir/messages to following: 
```


From test_viminfo.vim:
Executed Test_cmdline_history()                    in   0.067899 seconds
Executed Test_cmdline_history_order()              in   0.032296 seconds
Executed Test_global_vars()                        in   0.021711 seconds
Executed Test_global_vars_with_circular_reference() in   0.016178 seconds
Executed Test_viminfo_bad_syntax()                 in   0.017160 seconds
Executed Test_viminfo_bad_syntax2()                in   0.014733 seconds
Executed Test_viminfo_bad_syntax3()                in   0.178377 seconds
Executed Test_viminfo_bufferlist()                 in   0.110349 seconds
Executed Test_viminfo_encoding()                   in   0.016602 seconds
Executed Test_viminfo_error()                      in   0.028616 seconds
Executed Test_viminfo_file_mark_tabclose()         in   0.025528 seconds
Executed Test_viminfo_file_mark_unloaded_buf()     in   0.025827 seconds
Executed Test_viminfo_file_mark_zero_time()        in   0.008175 seconds
Executed Test_viminfo_file_marks()                 in   0.045957 seconds
Executed Test_viminfo_hlsearch()                   in   0.020748 seconds
Executed 15 tests                        in   1.303385 seconds
1 FAILED:
Found errors in Test_viminfo_jumplist():
E94: No matching buffer for Xviminfo
Test caused Vim to exit: Test_viminfo_jumplist()
```

I tried to take a swing at fixing it, but no luck.  Hope this helps narrow down the memory issue!

chrisbra:
> The logs do report the call stack and the line of code which had the error, but I was hoping the test logs would mention which test triggered the error.

Yeah, that is a bit unfortunate and I have wished for that as well. Hm, I wonder if we can set `ASAN_OPTIONS` to something like `log_path=${LOG_DIR}/asan_<test_name>` so it is easier to figure out which test caused the failure

basilisk0315:
Hi @jacobmealey thanks for the heads up. Big help.

I initially could not reproduce the error at my end, but then I switched from clang to gcc the heap error popped up in the same test file as you (text_viminfo.vim) but at a different place (Test_viminfo_last_spat_magic).

basilisk0315:
> Yeah, that is a bit unfortunate and I have wished for that as well. Hm, I wonder if we can set ASAN_OPTIONS to something like log_path=${LOG_DIR}/asan_<test_name> so it is easier to figure out which test caused the failure

That would be very useful. I tried to get that to work but gave up. I don't have much experience with ASAN.

In the end I just redirect stderr (ASAN writes to stderr) to a different file named after the test file.

basilisk0315:
Phew! I have resolved the issues. This just leaves a couple of test failures regarding values being returned in the unexpected ranges. These types of errors seem to come and go.

Thanks for your help all.

Cheers
John

chrisbra:
thanks, looks good

PR #15889, opened 2024-10-17
runtime:compiler:cppcheck: check for compile_commands in build subdirs

Konfekt:
As it is usually not in root dir


chrisbra:
thanks

PR #15890, opened 2024-10-17
Problem: AmigaOS4 build should use default runtime (newlib)

sodero:
Solution: Remove clib2 from compiler / linker flags


chrisbra:
thanks!

PR #15894, opened 2024-10-18
tests: set, setlocal and setglobal tests for each options

Milly:
- For each option, `set`, `setlocal`, and `setglobal` are all tested.
- Generates conditions such as `global-local` and `local-noglobal` from `runtime/doc/options.txt`.
- Create a `Test_opt_set_{fullname}` for each option instead of running them all through `Test_set_values`.
- Prevent unnecessary messages when running the generation script. (Use `call search(...)` instead of `/searchterm`.)

This test revealed some options that were missing error checking in `setglobal`, which I have already fixed and will be making a PR.

Problematic examples:
```vim
setglobal cc=foo    " No error
new
setlocal cc+=80    " E474: Invalid argument: cc+=80
```
User: "`cc+=80` is legal, so why...?"

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1695,7 +1695,7 @@ getout(int exitval)
 
 #ifdef FEAT_VIMINFO
     if (
-# if defined(EXITFREE) || defined(PROTO)
+# ifdef EXITFREE
 	    entered_free_all_mem == FALSE &&
 # endif
 	    *p_viminfo != NUL)
--- a/src/main.c
+++ b/src/main.c
@@ -1694,7 +1694,11 @@ getout(int exitval)
     }
 
 #ifdef FEAT_VIMINFO
-    if (*p_viminfo != NUL)
+    if (
+# if defined(EXITFREE) || defined(PROTO)
+	    entered_free_all_mem == FALSE &&
+# endif
+	    *p_viminfo != NUL)
 	// Write out the registers, history, marks etc, to the viminfo file
 	write_viminfo(NULL, FALSE);
 #endif
--- a/src/testdir/Make_ming.mak
+++ b/src/testdir/Make_ming.mak
@@ -22,7 +22,7 @@ default: nongui
 include Make_all.mak
 
 # Explicit dependencies.
-test_options.res test_alot.res: opt_test.vim
+test_options_all.res: opt_test.vim
 
 TEST_OUTFILES = $(SCRIPTS_TINY_OUT)
 DOSTMP = dostmp
--- a/src/testdir/Make_ming.mak
+++ b/src/testdir/Make_ming.mak
@@ -157,7 +157,7 @@ test_gui_init.res: test_gui_init.vim
 	$(VIMPROG) -u gui_preinit.vim -U gui_init.vim $(NO_PLUGINS) -S runtest.vim $<
 	@$(DEL) vimcmd
 
-opt_test.vim: gen_opt_test.vim ../optiondefs.h
+opt_test.vim: gen_opt_test.vim ../optiondefs.h ../../runtime/doc/options.txt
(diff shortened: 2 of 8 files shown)


zeertzjq:
It seems that this is causing problems with the CI logs. In the CI logs for the `testgui` job, a lot of logs for `test_options.vim` is missing. Also, there is now no `Skipped:` message for missing `opt_test.vim`. Maybe it's worth sourcing `opt_test.vim` from a new separate test file?

chrisbra:
Hm, there seems to be an ASAN failure: https://github.com/vim/vim/actions/runs/11414481680/job/31763444968?pr=15894

src/main.c
```diff
@@ -1694,7 +1694,11 @@ getout(int exitval)
     }
 
 #ifdef FEAT_VIMINFO
-    if (*p_viminfo != NUL)
+    if (
+# if defined(EXITFREE) || defined(PROTO)
+	    entered_free_all_mem == FALSE &&
+# endif
+	    *p_viminfo != NUL)
```
Milly: :memo: When leak detect is enabled, `p_viminfo` is accessed after freed.

src/testdir/Make_ming.mak
```diff
@@ -22,7 +22,7 @@ default: nongui
 include Make_all.mak
 
 # Explicit dependencies.
-test_options.res test_alot.res: opt_test.vim
```
chrisbra, line 25: you dropped one dependency here: `test_alot.res`. Is that intentionally?

src/main.c
chrisbra: ah thanks

src/testdir/gen_opt_test.vim
```diff
@@ -11,19 +12,65 @@ set nomore
...
+      \ scrolloff:	'TODO: fix missing error handling for setglobal',
+      \ shiftwidth:	'TODO: fix missing error handling for setglobal',
+      \ sidescrolloff:	'TODO: fix missing error handling for setglobal',
+      \ tabstop:	'TODO: fix missing error handling for setglobal',
+      \ termwinkey:	'TODO: fix missing error handling for setglobal',
+      \ termwinsize:	'TODO: fix missing error handling for setglobal',
+      \ textwidth:	'TODO: fix missing error handling for setglobal',
+      \}
+
 " The terminal size is restored at the end.
```
zeertzjq, line 64: It seems that this line should be removed.

src/testdir/gen_opt_test.vim
```diff
@@ -11,19 +12,65 @@ set nomore
...
+call cursor(1, 1)
+let global_locals = {}
+while search("^'[^']*'.*\\n.*|global-local", 'W')
+  let fullname = getline('.')->matchstr("^'\\zs[^']*")
+  let global_locals[fullname] = ''
+endwhile
+call extend(global_locals, #{
+      \ scrolloff: -1,
+      \ sidescrolloff: -1,
+      \ undolevels: -12345,
```
zeertzjq, line 28: Why is this -12345 instead of -123456?

src/testdir/gen_opt_test.vim
zeertzjq, line 64: It seems that this line should be removed as well.

Milly: Outdated?

zeertzjq: I mean, the terminal size is no longer restored at the end, as they are now restored in `test_prepost`.

Milly: Ah, I understand.

src/testdir/Make_ming.mak
Milly: *test_options.vim* is no longer sourced by *test_alot.vim* at bb962262.

PR #15892, opened 2024-10-17
tests: options valid or invalid values are not comprehensive

Milly:

PR #15896, opened 2024-10-19
Options with non-default values aren't restored after Test_set_values()

zeertzjq:
Problem:  Options with non-default values aren't restored after running
          Test_set_values().
Solution: Save the non-default values and restore them.

### DIFF
--- a/src/testdir/test_options.vim
+++ b/src/testdir/test_options.vim
@@ -1039,11 +1039,28 @@ endfunc
 
 func Test_set_values()
   " opt_test.vim is generated from ../optiondefs.h using gen_opt_test.vim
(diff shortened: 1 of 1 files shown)


zeertzjq:
Hmm, this doesn't fix the testgui failure from #15892. Maybe it's worth doing this anyway?

zeertzjq:
Actually this conflict with #15894. And there will be no problem if `opt_test.vim` is run from a separate test file.

chrisbra:
thanks, I wonder if we should include this until #15894 is ready ? Even so this is just a workaround. Wondering why writedelay does not seem to work in test_gui....

PR #15900, opened 2024-10-19
tests: tests fails depending on Windows environment.

Milly:
Tests failure on some windows environment.
This doesn't apply to GitHub CI, but it might help in some user environments.

```
> cmd /?
Starts a new instance of the Windows command interpreter

CMD [/A | /U] [/Q] [/D] [/E:ON | /E:OFF] [/F:ON | /F:OFF] [/V:ON | /V:OFF]
    [[/S] [/C | /K] string]

/C      Carries out the command specified by string and then terminates
/K      Carries out the command specified by string but remains
...
/D      Disable execution of AutoRun commands from registry (see below)
```

### DIFF
--- a/src/testdir/dos.vim
+++ b/src/testdir/dos.vim
@@ -3,7 +3,7 @@
 set shell=c:\COMMAND.COM shellquote= shellxquote= shellcmdflag=/c shellredir=>
 " This is used only when the +eval feature is available.
 if executable("cmd.exe")
-   set shell=cmd.exe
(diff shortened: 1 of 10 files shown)


chrisbra:
thanks. Do you know if `/D` always available?

Milly:
`/D` existed in cmd.exe in Windows XP, so I think it work in all currently available Windows environments.

chrisbra:
Thanks. I guess then it should be safe to include

PR #15901, opened 2024-10-19
Add support for the 'findprg' option

yegappan:
Add a new 'findprg' option to specify the external command to use for the `:find` command.

### DIFF
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -3556,13 +3556,16 @@ A jump table for the options with a short description can be found at |Q_op|.
 'findprg' 'fprg'	string	(default "")
 			global or local to buffer |global-local|
 	External program to use for the |:find| command.  When this option is
-	empty the internal file finding mechanism is used.  When executing the
-	external command, "$*" in 'findprg' is replaced with the argument
-	passed to the |:find| command.
-	If a matching file is found, the command should return one or more
-	file names separated by the newline character.  Example:
+	empty, the internal |file-searching| mechanism is used.
+	The external command is executed for each directory in 'path' until a
+	match is found.
+	When executing the external command, "$*" in 'findprg' is replaced
+	with the argument passed to the |:find| command.   The placeholder
+	"$@" in 'findprg' is replaced with a directory name from |'path'|.
+	If a match is found, the command should return one or more file names
+	separated by a newline character.  Example:
 >
-	    :set findprg=find\ .\ -name\ '$*'
+	    :set findprg=find\ $@\ -name\ '$*'
 <
 		*'fixendofline'* *'fixeol'* *'nofixendofline'* *'nofixeol'*
 'fixendofline' 'fixeol'	boolean	(default on)
--- a/runtime/doc/options.txt
+++ b/runtime/doc/options.txt
@@ -3552,6 +3552,18 @@ A jump table for the options with a short description can be found at |Q_op|.
 	  eob		EndOfBuffer		|hl-EndOfBuffer|
 	  lastline	NonText			|hl-NonText|
 
+						*'findprg'* *'fprg'*
+'findprg' 'fprg'	string	(default "")
+			global or local to buffer |global-local|
+	External program to use for the |:find| command.  When this option is
+	empty the internal file finding mechanism is used.  When executing the
+	external command, "$*" in 'findprg' is replaced with the argument
+	passed to the |:find| command.
+	If a matching file is found, the command should return one or more
+	file names separated by the newline character.  Example:
+>
+	    :set findprg=find\ .\ -name\ '$*'
+<
 		*'fixendofline'* *'fixeol'* *'nofixendofline'* *'nofixeol'*
 'fixendofline' 'fixeol'	boolean	(default on)
 			local to buffer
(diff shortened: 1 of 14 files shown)


runtime/doc/options.txt
```diff
@@ -3552,6 +3552,18 @@ A jump table for the options with a short description can be found at |Q_op|.
...
+'findprg' 'fprg'	string	(default "")
+			global or local to buffer |global-local|
+	External program to use for the |:find| command.  When this option is
+	empty the internal file finding mechanism is used.  When executing the
+	external command, "%" in 'findprg' is replaced with the current file
+	name.  To avoid this, escape "%" using "%%".
+	If a matching file is found, the command should return one or more
+	file names separated by the newline character.  Example:
+>
+	    :set findprg=find\ .\ -name\ '%'
```
bfrg: I don't quite understand the example. Why would anyone add the _current_ file name to `findprg`?

bfrg:
Thanks for doing this! Unfortunately, it doesn't always open the correct file.

1. To reproduce, open Vim in its src directory.
2. Run `set findprg=find` and then `:find beval<Tab>`. Select `beval.c` and press Enter. Vim will open a buffer named `.` and complain that `"." is a directory`.
3. Similarly, if you open the file with `:e beval.c` and then run `:find %:r<Tab>`, the wildmenu will display `beval.c` and `beval.h`. When I select `beval.h` and press Enter, Vim tries to open `.` again.

The above works fine with the internal find algorithm.

runtime/doc/options.txt
yegappan: The external find command needs a name to search for.  The "%" in 'findprg' is replaced with the argument passed to the ":find" command.

yegappan:
> Thanks for doing this! Unfortunately, it doesn't always open the correct file.
> 
> 1. To reproduce, open Vim in its src directory.
> 2. Run `set findprg=find` and then `:find beval<Tab>`. Select `beval.c` and press Enter. Vim will open a buffer named `.` and complain that `"." is a directory`.

The Linux/Unix find command needs the directory and a name to search for.  So 'findprg' should be set to `find . -name '%'`.

>
> 3. Similarly, if you open the file with `:e beval.c` and then run `:find %:r<Tab>`, the wildmenu will display `beval.c` and `beval.h`. When I select `beval.h` and press Enter, Vim tries to open `.` again.
> 
> The above works fine with the internal find algorithm.

bfrg:
> The Linux/Unix find command needs the directory and a name to search for. So 'findprg' should be set to find . -name '%'.

Oops, I completely forgot the `-name`. But the starting point is optional, at least in GNU find.

Well, I assumed that every file that is listed in the wildmenu after pressing TAB can be opened, but that's not the case. It seems like the command-completion still uses the internal find implementation?

And what is the purpose of `%`, is it a placeholder? If yes, shouldn't it be `$*` like in `grepprg` and `makeprg`. The help says that `%` is expanded to the _current_ file name.

runtime/doc/options.txt
bfrg: So it's a placeholder like the `$*` in `grepprg` and `makeprg`. Shouldn't we use the same here to be consistent?

benknoble:
The utility of `:find` is that it uses `'path'`; is there a design that doesn't give that up unnecessarily?

runtime/doc/options.txt
yegappan: Yes.  Updated the PR to use "$*" for the placeholder.

yegappan:
> The utility of `:find` is that it uses `'path'`; is there a design that doesn't give that up unnecessarily?

Yes.  We can support calling the `findprg` external program for every directory in `path`.  But the directories
with the wildcard "*" and "**" will need to be ignored though as the external programs may not support
those.

chrisbra:
Thanks. I wonder however if `findprg` is the best option. I remember a few years ago we had a discussion about `*prg` vs `*expr` options and I wonder if a `'findexpr'` option would be more useful and flexible?

bfrg:
> The utility of :find is that it uses 'path'; is there a design that doesn't give that up unnecessarily?

Personally I wouldn't expect `:find` to search in `path` after I've specifically set the find command with `findprg`.

Would it even work with something like `set findprg=git\ ls-files`?

yegappan:
> > The utility of :find is that it uses 'path'; is there a design that doesn't give that up unnecessarily?
> 
> Personally I wouldn't expect `:find` to search in `path` after I've specifically set the find command with `findprg`.
> 
> Would it even work with something like `set findprg=git\ ls-files`?

If the `findprg` value doesn't have the placeholder "$@", then the command will be executed only once and will not be executed for every path item.  Only if the `findprg` value contains "$@", the command will be executed for every path item until a match is found.

yegappan:
> Thanks. I wonder however if `findprg` is the best option. I remember a few years ago we had a discussion about `*prg` vs `*expr` options and I wonder if a `'findexpr'` option would be more useful and flexible?

I have created the PR https://github.com/vim/vim/pull/15905 to add the 'findexpr' option.  One of these two PRs can be merged based on the feedback.

chrisbra:
thanks for working on that. I am very much in favor of #15905 instead of this one. Any objections here?

bfrg:
So far `findprg` is not used for command-completion after `:find`. I think this is unexpected.

